PR #51404, opened 2026-07-15
fix: avoid repeated grantee scans in ListPolicy

shaoting-huang:
Related: #50236

- Reuse the initially loaded grantee keys/values when ListPolicy checks legacy grantee ID ownership.
- Avoid reloading the full grantee-privileges tree once per legacy grant.
- Add a regression test that verifies ListPolicy performs only one full grantee-prefix load for legacy grants.


congqixia (approved):
/lgtm

PR #51403, opened 2026-07-15
fix: restore dropped load transient budget fixes

sparknack:
issue: #49499
related: #50478

## Summary

Restore and complete the load transient-memory admission changes dropped by the final force-push of #50478.

The PR applies one process-wide transient-memory budget to Storage V2/V3 field-data loading and packed scalar V3 entry streaming, and aligns MCL loading-overhead reservations with the same concurrency model.

### Runtime transient budget

- `common.loadTransientBudgetBytes` limits aggregate in-flight transient bytes across concurrent load tasks.
- `0` disables runtime throttling. A non-zero value limits concurrency, while a request larger than the configured capacity may run exclusively to guarantee progress.
- Field-data batches acquire budget before load-pool submission, so budget waits do not occupy load workers. Budget is released incrementally as cells are finalized.
- Entry streams reserve budget per slice and release it after the consumer finishes with that slice.
- Cancellation wakes blocked budget admission and propagates through returned futures.

### MCL loading-overhead accounting

- Final resident resources and temporary loading overhead are estimated separately.
- Storage V2/V3 field translators use the shared `LoadTransientOverhead` group. With a configured budget, the group upper bound is at least both the budget and the largest single task. With budget `0`, the upper bound uses effective HIGH + LOW load-pool concurrency, including workers still alive during pool shrink.
- Memory and file overhead are tracked independently. Non-mmap field loads do not register file overhead.
- Packed scalar V3 indexes join the shared group only when the runtime budget is enabled. With budget `0`, each concurrently loading scalar slot reserves its own estimated overhead instead of using a process-wide shared cap.
- The scalar shared-group upper bound is process-wide and is not truncated by one file's total transient bytes. Per-cell overhead remains bounded by the concrete file estimate.
- Re-registering the same group takes the maximum upper bound per resource dimension, preventing a smaller translator from lowering the process-wide cap.

### Packed scalar V3 estimation

- Inspect only the packed-file footer and directory during admission; do not prefetch or decrypt `__meta__`.
- Detect encryption from each file's persisted directory instead of global plugin state.
- For encrypted files, derive total and per-task transient bytes from persisted ciphertext sizes and plaintext slice sizes.
- For plaintext file streams, account for both the read buffer and the aligned-write copy.
- Carry file-aware stream metadata into scalar resource estimation and CacheSlot construction. Hybrid indexes resolve their internal scalar type before estimating resources.
[...]

Issue #49499, opened 2026-05-04
[Enhancement]:Partial download capabilities


gland1:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

We need the ability for downloading parts of parquet files and enhance the concurrency while downloading such parts. 
For example, we needed such capability in:
1)	Index creation of large segment - to reduce RAM peak consumption during index creation, partial download is required to stream the parquet file by parts into disk instead of loading all of it first to memory and then save it to disk. (See also Milvus issue  - Reduce RAM consumption during index creation)

2)	During clustering compaction by vector field, we need to download part of a segment insert log file for gathering a uniform training sample from each segment.

liliu-z:
@sparknack is focusing on this. Plz take a check

sparknack:
We are developing a globally unified budget for transient memory usage to resolve this issue in #50267.
#50267 applies exclusively to Scalar Index V3. PR for field data and vector index are currently in progress.

PR #51405, opened 2026-07-15
fix: [3.0] restore dropped load transient budget fixes

sparknack:
pr: #51403
issue: #49499

## Summary

- Rebuild the cherry-pick from the current squash-merged form of #51403 on the latest `origin/3.0`.
- Restore process-wide transient-memory admission for Storage V2/V3 field loading and packed scalar V3 entry streams.
- Align MCL loading-overhead accounting with the runtime budget and effective load-pool concurrency.
- Correct file-aware scalar V3 estimates for plaintext, encrypted, bitmap, and hybrid index paths.
- Preserve cancellation, error-draining, and thread-pool submission safety fixes from the master PR.
- Use knowhere `v3.0.7`, which includes zilliztech/knowhere#1745 and the milvus-common group-policy API used by #51403.

### DIFF
--- a/internal/core/src/storage/LoadOverheadController.cpp
+++ b/internal/core/src/storage/LoadOverheadController.cpp
@@ -1,0 +1,163 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "storage/LoadOverheadController.h"
+
+#include <limits>
+
+#include "cachinglayer/Manager.h"
+#include "common/EasyAssert.h"
+#include "log/Log.h"
+
+namespace milvus::storage {
+
+namespace {
+
+bool
+UpdateGroupPolicy(const cachinglayer::LoadingOverheadGroupHandle& group_handle,
+                  const cachinglayer::LoadingOverheadPolicy& policy,
+                  const char* resource_name) {
+    if (group_handle == nullptr) {
+        return true;
+    }
+    auto result =
+        cachinglayer::Manager::UpdateLoadingOverheadGroup(group_handle, policy);
+    if (result != cachinglayer::LoadingOverheadUpdateResult::kApplied) {
+        LOG_ERROR("Failed to update load {} loading overhead group",
+                  resource_name);
+        return false;
+    }
+    return true;
+}
+
+template <cachinglayer::LoadingOverheadDimension Dimension>
+constexpr const char*
+ResourceName() {
+    if constexpr (Dimension ==
+                  cachinglayer::LoadingOverheadDimension::kMemory) {
+        return "memory";
+    }
+    return "file";
+}
+
+}  // namespace
+
+template <cachinglayer::LoadingOverheadDimension Dimension>
+LoadOverheadController<Dimension>&
+LoadOverheadController<Dimension>::GetInstance() {
+    static LoadOverheadController instance;
+    return instance;
+}
+
+template <cachinglayer::LoadingOverheadDimension Dimension>
+cachinglayer::LoadingOverheadPolicy
+LoadOverheadController<Dimension>::CurrentPolicy() const {
+    if constexpr (Dimension ==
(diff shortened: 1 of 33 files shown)


internal/core/thirdparty/knowhere/CMakeLists.txt
```diff
@@ -14,9 +14,9 @@
 # Update KNOWHERE_VERSION for the first occurrence
 milvus_add_pkg_config("knowhere")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-# v3.0.6 includes the knowhere iterator error-code interface
-# (zilliztech/knowhere#1699) that this segcore adaptation depends on.
-set( KNOWHERE_VERSION v3.0.6 )
+# v3.0.7 includes zilliztech/knowhere#1745 with the Cardinal and
+# milvus-common revisions required by the loading-overhead group API.
+set( KNOWHERE_VERSION v3.0.7 )
```
liliu-z, line 19: Could we pin the exact `16a71a3d` / zilliztech/knowhere#1745 commit used by master instead of `v3.0.7`? This tag also exposes `inverted_index_codec=block_adaptive`; its BM25 singleton path inserts from an empty `ip_vals_buf` and reproducibly crashes under ASan/UBSan (fixed after the tag in zilliztech/knowhere#1760). It also writes encoding id 4 without bumping the advertised engine version, so v3.0.6 nodes cannot load it during a rolling upgrade.

internal/core/src/storage/LoadOverheadController.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+bool
+UpdateGroupPolicy(const cachinglayer::LoadingOverheadGroupHandle& group_handle,
+                  const cachinglayer::LoadingOverheadPolicy& policy,
+                  const char* resource_name) {
+    if (group_handle == nullptr) {
+        return true;
+    }
+    auto result =
+        cachinglayer::Manager::UpdateLoadingOverheadGroup(group_handle, policy);
```
liliu-z, line 37: `kApplied` here only swaps the MCL policy; it does not reserve the larger target. Already-admitted loads can be waiting in `AcquireUntil()` or have work queued in HIGH/LOW, and the callers then expand and wake the runtime limit without another CacheSlot reserve. MCL can keep booking the old bound while actual transient usage reaches the new one, causing over-admission or OOM near the DList limit. Please reconcile the positive delta before expanding, or wait for the group to drain.

internal/core/src/storage/ThreadPools.cpp
```diff
@@ -109,19 +138,73 @@ ThreadPools::GetThreadPool(milvus::ThreadPoolPriority priority) {
...
+        LOG_ERROR(
+            "Failed to expand threadPool because the load overhead group "
+            "update failed, priority:{}, size:{}",
+            priority,
+            size);
         return;
     }
-    iter->second->Resize(size);
+
+    pool->Resize(size);
```
liliu-z, line 176: `Resize()` only lowers `max_threads_size_`; extra workers retire only after a 2s idle timeout, so with a busy queue they can keep running indefinitely. The policy update below then lets MCL reconcile to the smaller configured count while those workers are still executing. With the default `loadTransientBudgetBytes=0` there is no runtime memory cap, and file overhead is always executor-based. Can we keep the policy at the live-worker count until the extra workers actually exit?

internal/core/thirdparty/knowhere/CMakeLists.txt
sparknack: Done. The v3.0.7 tag has been updated and now points to 6d0f7fe.

internal/core/src/storage/LoadOverheadController.cpp
sparknack: This is an expected limitation for now.

The HIGH/LOW pool sizes are startup-oriented capacity settings rather than active runtime load-control knobs. In practice, they are almost never changed after startup, especially while loads are in progress.

Since we plan to replace this thread-pool-derived model with explicit async admission control in the future, we prefer not to introduce that complexity in this cherry-pick.

internal/core/src/storage/ThreadPools.cpp
sparknack: This is also an expected limitation under the same runtime-reconfiguration assumption.

Tracking live workers strictly would require changes to the thread-pool lifecycle. Workers retire independently after becoming idle, so the pool would need to notify the controller whenever an extra worker actually exits. The controller would then need to aggregate the live counts of both HIGH and LOW pools and reconcile the MCL policy. This also has to handle concurrent resizes, worker exits racing with policy updates, pool shutdown, and lock ordering between the worker lifecycle, resize path, and MCL. Since the live count may change again while an update is in progress, a simple one-shot callback is insufficient; it would require generation tracking or a reconciliation loop.

HIGH/LOW pool sizes are rarely changed after startup and are not intended to be adjusted as an active load-control mechanism while loads are running. Considering that future async admission control will replace this thread-pool-derived concurrency model, we prefer not to add this lifecycle and synchronization complexity here.

liliu-z (commented):
/lgtm
/approve

Issue #51366, opened 2026-07-14
[Bug]: BM25 field added by add_function_field remains unsearchable when base vec index is created after collection


lyyyuna:
## Environment

- **Milvus Version**: `harbor.tencent-stage.zilliz.cn/milvusdb/milvus:master-20260712-c8848b12`
- **Deployment Mode**: cluster
- **MQ**: woodpecker
- **SDK**: PyMilvus `3.1.0rc61`
- **OS**: Kubernetes

### Script

The script assumes Milvus is reachable at `http://127.0.0.1:19530`.

The important trigger in this script is that the base `vec` index is created **after** `create_collection` by calling `create_index(...)`.

```python
from __future__ import annotations

import time
import uuid

from pymilvus import DataType, FieldSchema, Function, FunctionType, MilvusClient


URI = "http://127.0.0.1:19530"
DIM = 4
QUERY = "bm25_repro_token"
ROW_COUNT = 500


def vector_for_id(row_id: int) -> list[float]:
    return [float(row_id), float(row_id % 7), float(row_id % 5), float(row_id % 3)]


def wait_index_ready(client: MilvusClient, collection: str, index_name: str, timeout: int = 180):
    deadline = time.time() + timeout
    last = None
    while time.time() < deadline:
        last = client.describe_index(collection, index_name=index_name)
        items = last if isinstance(last, list) else [last]
        if items and all(
            item.get("state") == "Finished"
            and int(item.get("pending_index_rows") or 0) == 0
            for item in items
        ):
            return last
        time.sleep(1)
    raise TimeoutError(f"index {index_name} not ready within {timeout}s; last={last}")


def main() -> None:
    client = MilvusClient(uri=URI)
    collection = f"bm25_post_create_index_repro_{uuid.uuid4().hex[:8]}"
    start = time.time()
    try:
        schema = client.create_schema(enable_dynamic_field=False, auto_id=False)
        schema.add_field("id", DataType.INT64, is_primary=True)
        schema.add_field("vec", DataType.FLOAT_VECTOR, dim=DIM)
        schema.add_field("text", DataType.VARCHAR, max_length=1024, enable_analyzer=True)

        dense_index = client.prepare_index_params()
        dense_index.add_index(field_name="vec", index_type="AUTOINDEX", metric_type="L2")
        client.create_collection(
            collection_name=collection,
            schema=schema,
            consistency_level="Strong",
        )
        client.create_index(collection_name=collection, index_params=dense_index)

        client.insert(
[...]

PR #51105, opened 2026-07-07
enhance: bind index meta to add function field DDL

MrPresent-Han:
issue: #51104
companion SDK PR (merge first): milvus-io/pymilvus#3670

## Summary

`add_function_field` previously accepted a vector output field without any index meta; bump-schema-version compaction then stalled forever (no segment index task can be created for a field with no index defined, and `GetQueryVChanPositions` never hands off unindexed compacted-to segments — the fail-closed chain is silent). This PR makes the index meta a mandatory, atomically-bound part of the DDL:

- **Mandatory at every layer**: SDK requires `index_params`; proxy and rootcoord reject vector function-output fields without an explicit `index_type` strictly before the broadcast (no silent AUTOINDEX).
- **create_index-aligned materialization**: proxy normalizes the params with the same logic as `create_index` (name rules via `validateIndexName`, checker existence, `CheckTrain` incl. dimension filling, function-type defaults such as `bm25_k1/bm25_b/bm25_avgdl`) and writes them back into the request; rootcoord prepare allocates index id/name, rejects name conflicts and unknown index types, and serializes a complete `indexpb.FieldIndex` into the new `AlterCollectionMessageUpdates.bound_field_indexes` WAL field.
- **Atomic apply**: the alter-collection ack callback commits the schema and then applies the bound index by dispatching a synthetic `CreateIndexMessage` to datacoord's existing `createIndexV2AckCallback` (same pattern as `cascadeDropFieldIndexesInline`). The callback is replayed until success across crashes and is fully idempotent (all ids come from the message body), so `DDL success ⇒ index meta exists` holds with no partial terminal state.
- **Reuse promotion**: `ValidateIndexParams`/`CheckDuplidateKey` moved from datacoord to `internal/util/indexparamcheck`; shared `ExpandIndexParams` / `FillFunctionOutputIndexParams` / `PrepareFunctionOutputIndexParams` helpers now back both the create_index path and this DDL.
- **Dead fan-out removed**: the ordinary `CreateIndex` broadcast is reverted to control-channel-only; the vchannel fan-out had no consumer and wrote one dead WAL entry per vchannel per index creation.
- **No milvus-proto change**: the request already carried `FieldInfo.index_name`/`extra_params` (previously unused); this feature activates them.
[...]

Issue #47882, opened 2026-02-26
[Bug] Compaction forward delete replay blocks tsafe advancement, causing Strong consistency query timeout


wangting0128:
## Environment

- **Milvus**: `2.6-20260225-cac3f784-amd64` (standalone, 8c16m)
- **PyMilvus**: `2.6.10rc2`
- **Deploy mode**: standalone with streaming enabled
- **Config**:
  ```yaml
  quotaAndLimits:
    dml:
      enabled: true
      upsertRate:
        max: 0.5
      insertRate:
        max: 0.5
      deleteRate:
        max: 0.5
    quotaCenterCollectInterval: 1
  ```
- **Collection**: 2 shards, 16 partitions (partition key), HNSW index, 2M vectors

## Problem

During a 5-hour concurrent upsert + `count(*)` (Strong consistency) stress test, a query request timed out with `DEADLINE_EXCEEDED` after 60 seconds. **The root cause is NOT slow segment scanning (which only took 67ms), but compaction forward delete replay blocking tsafe advancement for ~57-62 seconds.**

When compaction produces new sealed segments, the delegator must replay all historical WAL delete records from the segment's `startPosition` to catch up. In a long-running upsert workload, this replay grows linearly with test duration and can take 30-40+ seconds per segment.

### Timeline (requestId: `73297c19ab634f35a0851fa01caaa9da`)

```
00:00:23.920  Compaction triggers: 11 new L1 sealed segments need loading on channel 0
00:00:25~26   Segments loaded from storage

00:00:26.638  seg...015404 starts forward delete replay
              └─ startPosition = 2026/02/25 19:04:09  (test start time, ~5 hours ago!)
              └─ Must replay ~5 hours of delete WAL records

00:00:30.247  Query arrives (guaranteeTimestamp=464535953144283137)
              ├─ Channel 0: delegator waiting for tsafe (blocked by forward delete)
              └─ Channel 1: delegator waiting for tsafe (blocked by forward delete)

00:01:05.219  seg...015404 forward delete done (38.6 seconds!)
00:01:05.545  seg...035438 starts forward delete (startPosition=21:55:11, ~2h history)
00:01:05.548  seg...145439 starts forward delete (startPosition=21:55:45, ~2h history)

00:01:26.885  seg...035438 forward delete done (21.3s)
00:01:27.362  seg...145439 forward delete done (21.8s)
              ════ Channel 0 tsafe finally advances ════
[...]

PR #51409, opened 2026-07-15
fix: sync add-field bound index meta to sealed and growing segments (#51366)

MrPresent-Han:
## What

Fixes the bug in #51366: a BM25 function field added by `add_function_field` stays unsearchable when the base vector index was created **after** the collection. Searching the new field throws segcore `field index meta not found for field N` (`IndexMeta.cpp`) and never recovers within the backfill / index-build window.

## Root cause

Each segment's per-segment `col_index_meta_` is a **construction-time snapshot** and is never refreshed. `add_function_field` already carries the field's index meta on the DDL (`bound_field_indexes`, #51105), but only rootcoord consumes it — already-loaded segments never learn the new field. The only backstop is an index-checker reload that does **not** fire for below-threshold (fake-finished) segments, so the field can stay unsearchable for a long window.

## Fix

Implement the querynode `UpdateIndex` path so the add-field DDL's bound index meta is fanned to every loaded segment:

pipeline reads `bound_field_indexes` off the same `AlterCollection` DDL → `delegator.UpdateIndex` → `worker.UpdateIndex` → `QueryNode.UpdateIndex` → `collectionManager.UpdateIndex` (merge delta, monotonic by `index_barrier_ts`) → `segment.UpdateIndexMeta`, which atomically swaps `col_index_meta_` on **sealed** (`ChunkedSegmentSealedImpl`) and **growing** (`SegmentGrowingImpl` / `IndexingRecord`) segments. Field searchability is thereby decoupled from per-segment index build, so the brute-force `GetFieldIndexMeta` lookup no longer throws.

## Relation to #51004

Revives the approach of #51004, which relied on a **separate** `CreateIndex` WAL message + pipeline consumer. This is **simplified**: the index meta rides the same add-field DDL (it must, since #51105), so schema and index apply atomically under one barrier (`index_barrier_ts` = DDL `BeginTs`, mirroring `schema_barrier_ts`). No separate `CreateIndexMessage`.

Covers sealed + growing segments; `DropIndex` / `AlterIndex` propagation is left as a follow-up.

## Tests

- **Go UTs**: `mergeIndexAction` (all branches); `collectionManager.UpdateIndex` (real cgo — monotonic guard, merge+advance, idempotent stale re-fan, not-found); `QueryNode.UpdateIndex` and `delegator.UpdateIndex` fan-out; remote + local `worker.UpdateIndex`; per-segment sealed + growing `UpdateIndexMeta` (real cgo) and the L0 no-op; the pipeline reading `bound_field_indexes` off the `AlterCollection` DDL.
- **C++ gtests**: `Sealed.UpdateIndexMeta` and `Growing.UpdateIndexMeta` cover the segcore setters and the `UpdateSegmentIndexMeta` cgo boundary (both PASS locally).

issue: #51366


internal/querynodev2/segments/collection.go
```diff
@@ -128,10 +132,17 @@ func (m *collectionManager) PutOrRef(collectionID int64, schema *schemapb.Collec
...
+		// The load meta is the authoritative full index list, so replace. NOTE: a
+		// load snapshot that predates a very recent add-field DDL could momentarily
+		// regress a WAL-synced index (a narrow TOCTOU). The correct fix is a monotonic
+		// version guard here, which needs an index barrier threaded through the load
+		// path — see follow-up.
 		if meta != nil {
 			if err := collection.updateIndexMeta(meta); err != nil {
 				return err
 			}
+			collection.indexMeta.Store(meta)
```
liliu-z: PutOrRef stores the load's index meta into the new Go-side `collection.indexMeta` without version-gating or updating `indexMetaVersion`, so a load snapshot that predates a recently-applied add-field DDL rolls the WAL-synced meta back to a stale value while `indexMetaVersion` stays high. Because `collectionManager.UpdateIndex` then merges the next DDL onto `indexMeta.Load()` and fans the result to every segment, the rolled-back base drops the previously-synced field's index and re-propagates that loss to segments, re-introducing the unsearchable-field throw this PR fixes. The code comment acknowledges the TOCTOU, but the merge-base poisoning means it does not self-heal until a full reload. Version-gate the store so only strictly-newer load meta is applied. (Scope note: the ccollection push at line 142 is pre-existing; the new risk is the un-gated Go-side store plus the merge/fan-out machinery.)

MrPresent-Han: Thanks @liliu-z — confirmed and agreed, this is introduced by this PR: the new go-side `indexMeta` store on the load path (`PutOrRef`) is un-gated, and the merge/fan machinery then propagates a rolled-back base to segments.

Making `UpdateIndex` fully idempotent on the querynode against out-of-order load-path writes is costly — it needs a real monotonic index version threaded through the load path (synced with the WAL message sequence) and sourced from the coordinator index model; `schema_barrier_ts` can't stand in for it (it advances on non-index schema changes, so it decouples from index freshness).

For now I've switched `PutOrRef` to **union-merge** (never drop), so a stale / out-of-order load can no longer roll back a field already synced by a newer add-field DDL — covering both the go merge base and the ccollection push. Trade-off: a real `DropIndex` no longer propagates via the load path (a lingering entry is safe — brute-force still works, no throw; DropIndex is already deferred).

The proper idempotent fix (sync the indexMeta version with the WAL msg sequence) is tracked in #51451.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1233,6 +1235,63 @@ func (sd *shardDelegator) CatchingUpStreamingData() bool {
...
+			Op: &querypb.UpdateIndexRequest_Action_AddIndexRequest{
+				AddIndexRequest: &querypb.UpdateIndexRequest_AddIndex{
+					IndexInfo: fieldIndex.GetIndexInfo(),
+				},
+			},
+		},
+		IndexBarrierTs: indexBarrierTs,
+	}
+
+	sealed, growing, version := sd.distribution.PinOnlineSegments()
```
liliu-z, line 1349: UpdateIndex fans out via distribution.PinOnlineSegments(), a point-in-time snapshot that excludes segments still mid-load (a segment enters d.current only after worker.LoadSegments completes, via AddDistributions), and the receiver-side node.manager.Segment.GetBy() in QueryNode.UpdateIndex has the same limitation. A segment that reaches online after both snapshots — constructed on a worker whose collection index meta was never advanced by UpdateIndex (it got no fan-out RPC because it had no online segments) from a pre-DDL load-info union — comes online with a stale col_index_meta_ missing the added field. Because col_index_meta_ (unlike the schema, which self-heals via LazyCheckSchema on the search path) has no query-time reconcile, it keeps throwing 'field index meta not found' until a later reload carries post-DDL meta, and the index-checker skips exactly the below-threshold segments this PR targets; the PutOrRef union backstop only helps when the reload's load-info already contains the field. Consider advancing the worker's collection index meta / seeding the segment from it at publication, or ensuring post-DDL loads always carry the bound field index.

internal/querynodev2/pipeline/message.go
```diff
@@ -67,6 +79,16 @@ func (msg *insertNodeMsg) append(taskMsg msgstream.TsMsg) error {
...
 			msg.schema = body.GetUpdates().GetSchema()
 			msg.schemaBarrierTs = taskMsg.BeginTs()
+			// An add-field DDL carries the index meta bound to the new fields on the
+			// SAME message; fan it to segments' col_index_meta_ so the backfilled
+			// field's brute-force search stops throwing. Same BeginTs as the schema,
+			// so schema + index apply atomically and monotonically.
+			if boundFieldIndexes := body.GetUpdates().GetBoundFieldIndexes(); len(boundFieldIndexes) > 0 {
+				msg.indexUpdates = append(msg.indexUpdates, indexUpdate{
+					fieldIndexes: boundFieldIndexes,
+					barrierTs:    taskMsg.BeginTs(),
```
liliu-z, line 89: `BeginTs` isn't collection-wide: TimeTick is only comparable within one PChannel, but this version is later stored on a Collection shared by multiple shards. After a shard move, a newer DDL can look stale. Can we use a persisted collection-wide index epoch instead?

liliu-z (commented):
A few inline notes.

internal/querynodev2/delegator/delegator.go
```diff
@@ -285,7 +295,55 @@ func (sd *shardDelegator) GetSegmentInfo(readable bool) ([]SnapshotItem, []Segme
...
 func (sd *shardDelegator) SyncDistribution(ctx context.Context, entries ...SegmentEntry) {
+	sd.publishDistributions(ctx, entries...)
+}
+
+// publishDistributions makes segments readable and, under the index-meta fence, catches up
+// segments that missed a WAL-consumed index update while they were still loading. Every
+// AddDistributions call site routes through here.
+func (sd *shardDelegator) publishDistributions(ctx context.Context, entries ...SegmentEntry) {
+	sd.indexChangeMutex.RLock()
 	sd.distribution.AddDistributions(entries...)
```
liliu-z, line 306: `AddDistributions` only schedules the snapshot refresh, so the fence still has a gap: `UpdateIndex` can read the old `current` while this path captured the old index list. `AddGrowing` also bypasses this. Can we catch up before visibility and recheck the barrier?

internal/querynodev2/delegator/delegator.go
```diff
@@ -285,7 +295,55 @@ func (sd *shardDelegator) GetSegmentInfo(readable bool) ([]SnapshotItem, []Segme
...
+		}
+		req := &querypb.UpdateIndexRequest{
+			Base:              commonpbutil.NewMsgBase(commonpbutil.WithSourceID(paramtable.GetNodeID())),
+			CollectionID:      sd.collectionID,
+			CatchupIndexInfos: infos,
+			CatchupSegmentIds: segmentIDs,
+			IndexBarrierTs:    barrierTs,
+		}
+		req.GetBase().TargetID = nodeID
+		if _, err := worker.UpdateIndex(ctx, req); err != nil {
```
liliu-z, line 343: `UpdateIndex` puts apply failures in `Status` with a nil gRPC error, so this can silently publish a stale segment. Please check both via `merr.CheckRPCCall`; catch-up can't be best-effort here.

internal/querynodev2/services.go
```diff
@@ -1801,10 +1801,93 @@ func (node *QueryNode) DropIndex(ctx context.Context, req *querypb.DropIndexRequ
...
+// monotonic version guard makes it idempotent.
+func (node *QueryNode) catchUpSegmentIndexMeta(ctx context.Context, req *querypb.UpdateIndexRequest) *commonpb.Status {
+	collection := node.manager.Collection.Get(req.GetCollectionID())
+	if collection == nil {
+		return merr.Success() // released; nothing to catch up
+	}
+	// Union the caught-up DDL fields into the worker's current index meta (which already
+	// carries the base indexes each segment loaded) so applying it never drops the base.
+	addMeta := segments.ComposeIndexMeta(ctx, req.GetCatchupIndexInfos(), collection.Schema())
+	meta := collection.UnionIndexMeta(addMeta)
```
liliu-z, line 1868: `UnionIndexMeta` only builds a temporary value; it doesn't update `Collection.indexMeta` or CCollection. Search plan creation reads CCollection before the segment, so this can still fail with `field index ... is not loaded`, and the next DDL merges from a stale base.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -770,15 +770,15 @@ func (sd *shardDelegator) withPostLoadLimit(ctx context.Context, fn func() error
...
 	sd.schemaChangeMutex.RLock()
 	defer sd.schemaChangeMutex.RUnlock()
 	if schemaBarrierTs < sd.schemaBarrierTs {
 		return merr.WrapErrServiceInternal("schema barrier changed")
 	}
 
-	// alter distribution
-	sd.distribution.AddDistributions(entries...)
+	// alter distribution (+ index-meta catch-up for segments that missed the DDL fan-out)
+	sd.publishDistributions(ctx, entries...)
```
liliu-z, line 781: `loadStreamDelete` still holds `deleteMut.RLock` plus this schema lock here, so `publishDistributions` may do a worker RPC inside the delete critical section. A slow catch-up can block `ProcessDelete` and freeze tSafe (same shape as #47882). Can we move the RPC out of the lock?

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -105,6 +105,23 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
 
+	// Fan the add-field DDL's bound index meta to segments, after the schema apply
+	// above (both come from the same DDL). A failure is logged and swallowed here;
+	// the load-path PutOrRef reconcile is the backstop when a segment next (re)loads.
+	for _, iu := range nodeMsg.indexUpdates {
+		if len(iu.fieldIndexes) == 0 {
+			continue
+		}
+		ctx := context.TODO()
+		if err := dNode.delegator.UpdateIndex(ctx, iu.fieldIndexes, iu.barrierTs); err != nil {
```
liliu-z, line 116: If this fails we only log it, then advance tSafe with no durable retry/reconcile for already-online segments. Reload is the missing backstop for the fake-finished segments this PR targets, so this can leave the field broken indefinitely. Can we keep desired/applied state and retry?

internal/querynodev2/services.go
```diff
@@ -1801,10 +1801,93 @@ func (node *QueryNode) DropIndex(ctx context.Context, req *querypb.DropIndexRequ
...
+	}
+	// Marshal once and fan the same blob to every segment (each applies it under its own
+	// monotonic guard), instead of re-marshalling the meta per segment.
+	blob, err := proto.Marshal(newMeta)
+	if err != nil {
+		log.Warn(ctx, "failed to marshal collection index meta", mlog.Err(err))
+		return merr.Status(err), nil
+	}
+	var firstErr error
+	for _, segment := range node.manager.Segment.GetBy(segments.SegmentFilterFunc(func(s segments.Segment) bool {
```
liliu-z, line 1841: Each vchannel handles this DDL, but every RPC scans every segment of the collection. On a multi-shard node this becomes repeated full fan-out plus per-segment C++ parsing. Can we pass shard segment IDs or dedupe once per worker/broadcast?

Issue #48983, opened 2026-04-13
feat: support drop field, drop function field and disable dynamic field


sijie-ni-0214:
## Summary

Add support for dropping fields from a collection schema after creation, including:

- **Drop regular fields** by name or field ID via `AlterCollectionSchema` with `DropRequest`
- **Drop function fields** by function name, with cascade deletion of output fields and indexes
- **Disable dynamic field** via `AlterCollection` properties (`dynamicfield.enabled=false`)

## Motivation

Currently, fields added to a collection schema cannot be removed. This prevents users from iterating on their schema design without recreating the collection. Drop field support enables schema evolution by allowing safe removal of unused fields.

## Design

All drop operations are unified under the existing `AlterCollectionSchema` RPC using an action-based dispatch (`AddRequest` / `DropRequest`), consistent with the upstream add-function-field framework (PR #48810).

Key components:
- **Proxy layer**: validates drop safety (cannot drop PK, partition key, clustering key, last vector field, function-referenced fields, dynamic field)
- **RootCoord**: broadcasts schema changes via WAL with `DroppedFieldIds` for cascade index deletion
- **C++ segcore**: schema-driven filtering in `ComputeDiff*` functions and defense-in-depth checks during segment data loading
- **FieldID reuse prevention**: `max_field_id` property ensures new fields never reuse dropped field IDs

PR #50394, opened 2026-06-09
feat: support external collection schema evolution

weiliu1031:
issue: #45881

design doc: docs/design-docs/design_docs/20260605-external-table-schema-evolution.md

### What changed

This PR supports schema evolution for external collections.

It allows `AlterCollectionSchema` to update external collection schemas,
validates source-backed fields and function-output fields, adds SDK helpers,
and records refresh schema versions.

Refresh jobs recheck the captured schema version against DataCoord's current
schema cache before committing. Once DataCoord observes a newer schema, stale
refresh results are rejected. Cross-component serialization for the narrower
RootCoord-to-DataCoord cache propagation window remains out of scope.

It also ensures schema-visible external fields are not served until their
backing manifests are materialized. Query/search/index paths reject
unmaterialized fields, QueryCoord schedules loaded segment reopen when
manifests are stale, and QueryNode refreshes BM25/function state after schema
changes.

External collection schema-mutation WAL messages are unreplicable, including
the legacy add-field compatibility path. CDC therefore does not apply an
external schema change without the refresh and manifest updates needed to
materialize it. Internal collection schema evolution remains replicable.

### Validation

- `git diff --check`
- Targeted RootCoord tests for external and internal replication flags
- Targeted Go SDK external add-column refresh test

PR #51411, opened 2026-07-15
feat: add Go SDK and REST API support for drop field

sijie-ni-0214:
## Summary

- Add Go SDK APIs for adding a function-backed field and dropping a collection field or function-backed field through `AlterCollectionSchema`.
- Route Go SDK `AddCollectionField` through `AlterCollectionSchema`, with a fallback to the legacy `AddCollectionField` RPC when the new RPC is unavailable. The fallback covers both raw gRPC `Unimplemented` responses from older proxies and code-10 service-unimplemented responses from mixed-version proxy/coordinator deployments.
- Add REST v2 `POST /v2/vectordb/collections/fields/drop` support by field name or field ID, and route the existing `POST /v2/vectordb/collections/fields/add` handler through `AlterCollectionSchema` without changing its HTTP request contract.
- Validate bound-index parameters for function output fields, invalidate the Go client schema cache after successful mutations, and propagate `AlterCollectionSchemaResponse.AlterStatus` through REST responses, hooks, metrics, and logs.
- Add focused Go SDK, requestutil, REST handler, and end-to-end coverage.

issue: #48983

Design doc: docs/design-docs/design_docs/20260413-drop-collection-field-design.md

## Compatibility

- Go SDK: a new client falls back to the legacy `AddCollectionField` RPC when `AlterCollectionSchema` is unavailable, whether the signal is a raw gRPC `Unimplemented` status from an older proxy or a code-10 service-unimplemented status from a new proxy connected to an older coordinator. Other errors are returned without fallback.
- REST API: clients continue to call the same `/v2/vectordb/collections/fields/add` HTTP endpoint. An older Milvus server runs its existing legacy handler, while a newer server performs the `AlterCollectionSchema` conversion internally, so no REST client fallback is required.
- The new REST drop-field route is available only on servers that include this feature; there is no equivalent legacy drop RPC.

## Authorization

- On servers that support `AlterCollectionSchema`, Go SDK schema-mutation APIs and the REST add/drop field handlers are authorized with the collection-scoped `AlterCollectionSchema` privilege.
- When Go `AddCollectionField` falls back against an older server, the legacy `AddCollectionField` privilege applies.
- Existing REST function-field routes are unchanged by this PR.

## Test Plan

- [x] `make static-check`
- [x] `make build-go`
- [x] Go client `internal/merr` and `milvusclient` tests, including transport- and business-level AddField fallback plus non-fallback error cases
- [x] Milvus 2.6.17 compatibility E2E: `AlterCollectionSchema` returned gRPC `Unimplemented`, then legacy `AddCollectionField` succeeded
- [x] `pkg/util/requestutil` tests
- [x] Targeted REST handler and request-statistics interceptor package tests
- [x] Ruff check and format check for changed REST client files

### DIFF
--- a/client/internal/merr/merr.go
+++ b/client/internal/merr/merr.go
@@ -89,13 +89,15 @@ func (e *RPCError) Is(target error) bool {
 }
 
 var (
-	ErrServiceNotReady = newRPCError("service not ready", 1, commonpb.ErrorCode_NotReadyServe, true, false)
-	ErrServiceInternal = newRPCError("service internal error", 5, commonpb.ErrorCode_UnexpectedError, false, false)
+	ErrServiceNotReady      = newRPCError("service not ready", 1, commonpb.ErrorCode_NotReadyServe, true, false)
+	ErrServiceInternal      = newRPCError("service internal error", 5, commonpb.ErrorCode_UnexpectedError, false, false)
+	ErrServiceUnimplemented = newRPCError("service unimplemented", 10, commonpb.ErrorCode_UnexpectedError, false, false)
 
 	ErrCollectionNotFound       = newRPCError("collection not found", 100, commonpb.ErrorCode_CollectionNotExists, false, false)
 	ErrCollectionSchemaMismatch = newRPCError("collection schema mismatch", 109, commonpb.ErrorCode_SchemaMismatch, false, true)
 	ErrIndexNotFound            = newRPCError("index not found", 700, commonpb.ErrorCode_IndexNotExist, false, false)
 	ErrParameterInvalid         = newRPCError("invalid parameter", 1100, commonpb.ErrorCode_IllegalArgument, false, true)
+	ErrParameterMissing         = newRPCError("missing required parameters", 1101, commonpb.ErrorCode_IllegalArgument, false, true)
 
 	errUnexpected = newRPCError("unexpected error", unexpectedCode, commonpb.ErrorCode_UnexpectedError, false, false)
 )
@@ -326,4 +328,18 @@ func WrapErrParameterInvalid[T any](expected, actual T, messages ...string) erro
 		message = strings.Join(messages, "->") + ": " + message
 	}
 	return cloneWithMessage(ErrParameterInvalid, message)
+}
+
+func WrapErrParameterInvalidMsg(format string, args ...any) error {
+	return cloneWithMessage(ErrParameterInvalid, fmt.Sprintf(format, args...))
+}
+
+// WrapErrParameterInvalidErr reclassifies err as ErrParameterInvalid and only
+// retains the original error text, not its identity or code.
+func WrapErrParameterInvalidErr(err error, message string) error {
+	return cloneWithMessage(ErrParameterInvalid, message+": "+err.Error())
+}
+
+func WrapErrParameterMissingMsg(format string, args ...any) error {
+	return cloneWithMessage(ErrParameterMissing, fmt.Sprintf(format, args...))
 }
--- a/client/internal/merr/merr_test.go
+++ b/client/internal/merr/merr_test.go
@@ -127,6 +127,28 @@ func TestContextCodes(t *testing.T) {
 	require.Equal(t, TimeoutCode, Code(context.DeadlineExceeded))
 }
 
+func TestParameterMissing(t *testing.T) {
+	err := WrapErrParameterMissingMsg("collection name is required")
+	require.ErrorIs(t, err, ErrParameterMissing)
+	require.Equal(t, int32(1101), Code(err))
+	require.Equal(t, commonpb.ErrorCode_IllegalArgument, Status(err).GetErrorCode())
+}
+
+func TestParameterInvalidWrappers(t *testing.T) {
+	t.Run("message", func(t *testing.T) {
+		err := WrapErrParameterInvalidMsg("invalid field %q", "vector")
+		require.ErrorIs(t, err, ErrParameterInvalid)
+		require.Equal(t, int32(1100), Code(err))
+		require.Contains(t, err.Error(), `invalid field "vector"`)
+	})
+
+	t.Run("cause text", func(t *testing.T) {
+		err := WrapErrParameterInvalidErr(errors.New("invalid character"), "invalid legacy params")
+		require.ErrorIs(t, err, ErrParameterInvalid)
+		require.Contains(t, err.Error(), "invalid legacy params: invalid character")
+	})
+}
+
 func TestCheckRPCCall(t *testing.T) {
 	transportErr := errors.New("transport")
 	require.ErrorIs(t, CheckRPCCall(nil, transportErr), transportErr)
--- a/client/milvusclient/collection.go
+++ b/client/milvusclient/collection.go
@@ -21,8 +21,12 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/client/v3/entity"
 	"github.com/milvus-io/milvus/client/v3/internal/merr"
 )
@@ -197,6 +201,30 @@ func (c *Client) GetCollectionStats(ctx context.Context, opt GetCollectionOption
 	return stats, nil
 }
 
+func (c *Client) alterCollectionSchemaRequest(ctx context.Context, req *milvuspb.AlterCollectionSchemaRequest, callOpts ...grpc.CallOption) error {
+	return c.callService(func(milvusService milvuspb.MilvusServiceClient) error {
+		resp, err := milvusService.AlterCollectionSchema(ctx, req, callOpts...)
+		if resp == nil || resp.GetAlterStatus() == nil {
+			return merr.CheckRPCCall(nil, err)
+		}
+		if err := merr.CheckRPCCall(resp.GetAlterStatus(), err); err != nil {
+			return err
+		}
+		c.collCache.Evict(req.GetCollectionName())
+		return nil
+	})
+}
+
+func (c *Client) alterCollectionSchema(ctx context.Context, opt alterCollectionSchemaOption, callOpts ...grpc.CallOption) error {
+	if opt == nil {
+		return merr.WrapErrParameterMissingMsg("alter collection schema option is nil")
+	}
+	if err := opt.Validate(); err != nil {
+		return err
+	}
+	return c.alterCollectionSchemaRequest(ctx, opt.Request(), callOpts...)
+}
+
 // AddCollectionField adds a field to a collection.
 func (c *Client) AddCollectionField(ctx context.Context, opt AddCollectionFieldOption, callOpts ...grpc.CallOption) error {
 	if err := opt.Validate(); err != nil {
@@ -204,12 +232,42 @@ func (c *Client) AddCollectionField(ctx context.Context, opt AddCollectionFieldO
 	}
 
 	req := opt.Request()
+	fieldSchema := &schemapb.FieldSchema{}
+	if err := proto.Unmarshal(req.GetSchema(), fieldSchema); err != nil {
+		return merr.WrapErrParameterInvalidErr(err, "invalid field schema")
+	}
 
-	err := c.callService(func(milvusService milvuspb.MilvusServiceClient) error {
+	alterReq := newAlterCollectionSchemaAddRequest(req.GetCollectionName(), fieldSchema, nil)
+	alterReq.DbName = req.GetDbName()
+	alterReq.CollectionID = req.GetCollectionID()
+	err := c.alterCollectionSchemaRequest(ctx, alterReq, callOpts...)
+	if grpcstatus.Code(err) != codes.Unimplemented && !errors.Is(err, merr.ErrServiceUnimplemented) {
+		return err
+	}
+
+	return c.callService(func(milvusService milvuspb.MilvusServiceClient) error {
 		resp, err := milvusService.AddCollectionField(ctx, req, callOpts...)
-		return merr.CheckRPCCall(resp, err)
+		if err := merr.CheckRPCCall(resp, err); err != nil {
+			return err
+		}
+		c.collCache.Evict(req.GetCollectionName())
+		return nil
 	})
-	return err
+}
+
+// AddFunctionField adds a function and its output field to a collection.
+func (c *Client) AddFunctionField(ctx context.Context, opt AddFunctionFieldOption, callOpts ...grpc.CallOption) error {
+	return c.alterCollectionSchema(ctx, opt, callOpts...)
+}
+
+// DropCollectionField drops a field from a collection.
+func (c *Client) DropCollectionField(ctx context.Context, opt DropCollectionFieldOption, callOpts ...grpc.CallOption) error {
+	return c.alterCollectionSchema(ctx, opt, callOpts...)
+}
+
+// DropFunctionField drops a function and its output field from a collection.
+func (c *Client) DropFunctionField(ctx context.Context, opt DropFunctionFieldOption, callOpts ...grpc.CallOption) error {
+	return c.alterCollectionSchema(ctx, opt, callOpts...)
 }
 
 // AddCollectionStructField adds a struct array field to a collection.
--- a/client/milvusclient/collection_options.go
+++ b/client/milvusclient/collection_options.go
@@ -17,14 +17,18 @@
 package milvusclient
 
 import (
+	"encoding/json"
 	"fmt"
 
+	"github.com/samber/lo"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/client/v3/entity"
 	"github.com/milvus-io/milvus/client/v3/index"
+	"github.com/milvus-io/milvus/client/v3/internal/merr"
 )
 
 // CreateCollectionOption is the interface builds CreateCollectionRequest.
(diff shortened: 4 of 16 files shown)


czs007:
## Overview

Rerouting `drop_function` and the new field endpoints through `AlterCollectionSchema` changes three observable behaviors at the REST layer — request idempotency, the required RBAC privilege, and the metric label — none of which are covered by the new tests; since the affected endpoint is master-only and unreleased, these are semantics decisions to make deliberately rather than upgrade breaks.

## Medium

**`internal/distributed/proxy/httpserver/handler_v2.go:1068` — `drop_function` loses idempotency**

`dropCollectionFunction` now builds an `AlterCollectionSchemaRequest_DropRequest` with `FunctionName` set and `DropFunctionOutputFields` left `false`, replacing the `DropCollectionFunction` RPC. The two paths disagree when the function does not exist: the old task treated it as a success no-op (`internal/proxy/function_task.go:321-324`, with `PreExecute` leaving `fSchema` nil), while `validateDropFunction` now returns `function not found: %s` (`internal/proxy/task.go:1416-1418`). A client retrying after a timeout, or running idempotent cleanup twice, previously got success and now gets a user-visible error even though the desired end state already holds.

The endpoint is master-only (introduced by #49531, `39cd786ebb`, 2026-05-07; contained in no release tag or branch), so no shipped client depends on the old behavior — this is a semantics choice, not a compatibility break. It is also unguarded: the updated `TestDropCollectionFunctionNormal` only asserts that the mocked proxy receives the right `AlterCollectionSchemaRequest`, so `validateDropFunction` never runs.

*Suggestion:* decide explicitly whether `drop_function` remains idempotent. If yes, make the detach-only path (`DropFunctionOutputFields=false`) a no-op when the function is absent. If no, document the semantics change in the endpoint docs and add a test that exercises the real task.

*(raised by yezh, wzy)*

## Low

**`internal/distributed/proxy/httpserver/handler_v2.go:1068` — reroute silently changes the required RBAC privilege**

REST v2 authorization is derived from the request proto type, not the URL: `wrapperProxyWithLimit` → `checkAuthorizationV2` (`handler_v2.go:530`) → `proxy.PrivilegeInterceptor` → `funcutil.GetPrivilegeExtObj(req)`, which reads the `privilege_ext_obj` proto option. Passing `*milvuspb.AlterCollectionSchemaRequest` instead of `*milvuspb.DropCollectionFunctionRequest` therefore changes the required grant from `Global` / `PrivilegeCreateCollection` / `object_name_index:-1` to `Collection` / `PrivilegeAlterCollectionSchema` (enum 84) / `object_name_index:3`. With `authorizationEnabled=true`, a role holding global `CreateCollection` but not collection-level `AlterCollectionSchema` loses `drop_function`; a role holding only the latter gains it.

The new mapping is arguably more correct — it matches the RPC actually invoked and is collection-scoped rather than requiring a global create privilege — and the endpoint is unreleased, so no deployed grant set breaks on upgrade.

*Suggestion:* confirm the privilege change is intentional, and add coverage — the new tests call `initHTTPServerV2(mp, false)`, which disables auth entirely (`handler_v2_test.go:2468-2475`).

*(raised by wzy)*

**`internal/distributed/proxy/httpserver/handler_v2.go:92` — four routes collapse onto one metric label**

`routeToMethod` supplies the `methodTag` used as the `function_name` label of `metrics.ProxyFunctionCall` (`handler_v2.go:405-420`, the sole non-test consumer). This PR maps four distinct routes — the pre-existing `collections/drop_function` plus the new `collections/fields/drop`, `collections/function_fields/add` and `collections/function_fields/drop` — to the same `"AlterCollectionSchema"` value. The only other labels are `nodeID`, `status`, `dbName` and `collectionName`, so the four endpoints become indistinguishable: an add-function-field error rate cannot be separated from a drop-field error rate. `drop_function`'s label additionally flips from `DropCollectionFunction`.

Note that the REST access log is unaffected — it logs the URL path (`internal/proxy/accesslog/info/restful_info.go:101-103`), not `routeToMethod` — and no dashboard is keyed on the old label yet, given the endpoint is unreleased. The label collapse across the four new/changed routes is the real cost.

*Suggestion:* give each route a distinct `methodTag` (e.g. `DropCollectionField` / `AddFunctionField` / `DropFunctionField`) rather than reusing the RPC name.

*(raised by wzy)*

czs007:
Two issues survive; here's the comment body.

---

Adversarial review surfaced two medium-severity concerns, both stemming from the same reroute of `POST /v2/vectordb/collections/drop_function` onto the `AlterCollectionSchema` RPC.

## Medium

**`internal/distributed/proxy/httpserver/handler_v2.go:1075` — Rerouting `drop_function` silently changes the required RBAC privilege**

`wrapperProxyWithLimit` passes the proto request straight into `checkAuthorizationV2` → `proxy.PrivilegeInterceptor`, which derives the required privilege purely from the request message's `privilege_ext` proto option (`funcutil.GetPrivilegeExtObj` reads the descriptor option; it is not route-based). Because `dropCollectionFunction` now builds a `*milvuspb.AlterCollectionSchemaRequest` instead of a `*milvuspb.DropCollectionFunctionRequest`, the endpoint switches from requiring `{ObjectType: Global, Privilege: PrivilegeCreateCollection, ObjectNameIndex: -1}` to `{ObjectType: Collection, Privilege: PrivilegeAlterCollectionSchema, ObjectNameIndex: 3}`. Any existing role granted only global `CreateCollection` will start receiving HTTP 403 on an endpoint it could previously call; conversely, the check narrows from a global grant to a per-collection one. `PrivilegeAlterCollectionSchema` is reachable only via the `ReadWrite` privilege group and the collection-level list (`pkg/util/constant.go:359`, `pkg/util/constant.go:116`), so it is not implied by a bare global `CreateCollection` grant — the deny is real.

Suggestion: either keep `DropCollectionFunctionRequest` for authorization/dispatch, or make the privilege change explicit — add a `handler_v2_test.go` case asserting 403 for a `CreateCollection`-only role, and call it out in the release note as an RBAC migration. Whichever direction is intended, this is currently a silent breaking auth change on a public endpoint with no test coverage and no release note. *(raised by auditor)*

**`internal/distributed/proxy/httpserver/handler_v2.go:92` — Breaking metric relabel: `drop_function`'s `function_name` becomes `AlterCollectionSchema`**

`routeToMethod`'s only consumer is `handler_v2.go:405`, which feeds the `function_name` label of `metrics.ProxyFunctionCall`, so the blast radius is exactly `proxy_function_call{function_name}`. The `drop_function` route changes from `DropCollectionFunction` to `AlterCollectionSchema`, silently emptying existing dashboards and alerts that select `function_name="DropCollectionFunction"`. The three new endpoints (`fields/drop`, `function_fields/add`, `function_fields/drop`) have no prior label but become mutually indistinguishable, and also indistinguishable from the gRPC `AlterCollectionSchema` call, which the gRPC interceptor labels with the same `FullMethod`-derived tag (`request_interceptor.go:53`, `internal/proxy/impl.go:1058`).

Suggestion: keep per-operation `methodTag` values — the map value is a free-form string and need not match the underlying RPC, so the fix is local — or at minimum document the relabel in the release note before merge. Note that access-log `method` and trace `request_name` collapsing is inherent to routing through the `AlterCollectionSchema` RPC and cannot be addressed via `routeToMethod`. *(raised by wzy, czs007)*

sijie-ni-0214:
Addressed the REST metric-label concern in b0f540616a:

- REST metrics now retain operation-level labels: DropCollectionFunction, DropCollectionField, AddFunctionField, and DropFunctionField. The handlers still invoke AlterCollectionSchema; only the REST metric dimension is kept distinct.

For the other two observations, the AlterCollectionSchema behavior is intentional:

- Missing functions return an error by design. The drop-field design document specifies "function not found", proxy task tests cover it, and pymilvus has routed drop_collection_function through AlterCollectionSchema since milvus-io/pymilvus@a7b3166b. Retaining the legacy no-op only for REST would make the public SDKs disagree.
- The RBAC requirement follows the request sent by the handler. AlterCollectionSchemaRequest is explicitly declared as collection-scoped PrivilegeAlterCollectionSchema in the proto, which is also the privilege required by the pymilvus path. Keeping the legacy global CreateCollection check for this REST route would make authorization inconsistent with the actual RPC contract.

Local formatting and diff checks pass. The targeted httpserver test command is blocked in this worktree before test execution because pkg-config cannot locate milvus_core.

czs007:
This PR review comment addresses three verified findings; here it is:

---

This PR routes several HTTP endpoints onto `AlterCollectionSchema`. The change is largely sound, but one pre-existing endpoint's authorization tier shifts as a side effect, and two smaller consistency points are worth noting.

## Medium

**`internal/distributed/proxy/httpserver/handler_v2.go:1067` — rerouting `/collections/drop_function` onto `AlterCollectionSchema` silently changes its required privilege**

`dropCollectionFunction` is a pre-existing endpoint that now emits `AlterCollectionSchemaRequest` instead of `DropCollectionFunctionRequest`. `PrivilegeInterceptor` resolves authorization from the request message's proto option via `funcutil.GetPrivilegeExtObj(req)` (`internal/proxy/privilege_interceptor.go:62`), not from the `fullMethod` string — so swapping the request type swaps the enforced privilege:

- before: `DropCollectionFunctionRequest` → ObjectType=Global, Privilege=`PrivilegeCreateCollection`, ObjectNameIndex=-1
- after: `AlterCollectionSchemaRequest` → ObjectType=Collection, Privilege=`PrivilegeAlterCollectionSchema`, ObjectNameIndex=3

This has two opposite consequences. First, a widening: `PrivilegeCreateCollection` is in `DatabaseAdminPrivileges` (`pkg/util/constant.go:387`) while `PrivilegeAlterCollectionSchema` is in `CollectionReadWritePrivileges` (`pkg/util/constant.go:359`), so a non-admin role holding only the CollectionReadWrite group can now drop functions off a collection — an operation that previously required a database-admin-tier grant. Second, a break: a role granted exactly `CreateCollection` but not `AlterCollectionSchema` loses access to the endpoint after upgrade. It also leaves an asymmetry, since `/collections/add_function` still enforces Global/`PrivilegeCreateCollection` — dropping a function is now strictly easier than adding one.

The per-collection scoping is a genuine granularity improvement and may well be intended, so this is not necessarily wrong — but it is an unannounced auth change on a shipped public endpoint.

*Suggestion:* if intended, add a release note documenting the privilege migration for `drop_function` and consider aligning `add_function`. Either way, please add a test with `checkAuth=true` asserting the enforced privilege; the new tests in `handler_v2_test.go` (e.g. `TestDropCollectionFieldRoutesToAlterCollectionSchema:4817`) construct the suite without `checkAuth` and only assert request routing.

*(raised by wzy, czs007)*

## Nitpick

**`internal/distributed/proxy/httpserver/handler_v2.go:1315` — hardcoded `merr.Code(merr.ErrParameterInvalid)` in the `genFunctionSchema` failure body**

The abort body uses `merr.Code(merr.ErrParameterInvalid)` while `return nil, err` propagates the raw error. There is no live divergence today — `genFunctionSchema` (`utils.go:3619-3646`) has exactly two error returns, both `merr.WrapErrParameterInvalidMsg`, so the codes agree — and the hardcoded form is the dominant local idiom here, including in the sibling handlers `addCollectionFunction` (:1025) and `alterCollectionFunction` (:1051). Raising it purely as a latent-consistency nit.

*Suggestion:* use `merr.Code(err)` so the body stays correct if `genFunctionSchema` later returns a more specific code.

*(raised by wzy)*

**`internal/distributed/proxy/httpserver/handler_v2.go:994` — `checkAlterCollectionSchemaRPCCall`'s error branch is shadowed by the wrapper**

This PR adds an `AlterStatusGetter` branch to `GetStatusFromResponse` (`pkg/util/requestutil/getter.go:185-189`), and `AlterCollectionSchemaResponse` exposes only `GetAlterStatus()`. So `wrapperProxyWithLimit` (:611-616) already converts a populated business `AlterStatus` into `err` and aborts at :624, and `alterCollectionSchema`'s `if err == nil` guard (:1002) then skips `checkAlterCollectionSchemaRPCCall` entirely. Line 994 still runs on every success (where `merr.Error` returns nil), so it is not dead code — but it can never produce a non-nil error via this path; the only live error producer left is the nil/absent-`AlterStatus` branch at :991-992. The PR's own `TestAlterCollectionSchemaHookSeesRawBusinessResponse` (`handler_v2_test.go:5015`) demonstrates this: the injected business error is absorbed by the wrapper, never by line 994.

Harmless today, but the duplicated `error_type`/`Warn`/`HTTPAbort` block at :1008-1010 mirrors the wrapper's :621-624 and could mislead a future refactor into treating it as the primary error path.

*Suggestion:* drop the shadowed branch, or add a one-line comment noting that business-status errors are handled by the wrapper.

*(raised by wzy)*

sijie-ni-0214:
Follow-up on the latest review:

1. The RBAC migration is intentional. The REST handler must authorize the same AlterCollectionSchema request that Go SDK and pymilvus use; retaining a DropCollectionFunction authorization exception only in REST would make equivalent SDK operations require different grants. The PR description now documents the migration and the required role update. AddCollectionFunction intentionally remains on its existing RPC and is outside this change.

2. The hardcoded parameter-invalid code in the genFunctionSchema failure path is unchanged. genFunctionSchema currently returns parameter-invalid errors only, and the form matches the adjacent existing function handlers.

3. The wrapper is the primary handler for non-success AlterStatus values. checkAlterCollectionSchemaRPCCall is retained as a defensive check for a nil response or nil AlterStatus before returning the success body; no behavior changes from the duplicate successful-status conversion.

No further code change is planned for these observations.

czs007:
Both issues survived verification (as rewrites), so here's the comment body:

---

Adversarial review surfaced two low-severity consistency nits; neither is a functional bug, and both are optional cleanups.

## Low

**`internal/distributed/proxy/httpserver/handler_v2.go:1317`** — *error-handling*

`addFunctionField` hardcodes `merr.Code(merr.ErrParameterInvalid)` in the HTTP response body while returning the original `err` to `wrapperPost`, which derives its metric/log classification via `requestutil.ParseMetricLabel(resp, err)`. Today these two paths agree: both error returns in `genFunctionSchema` (`utils.go:3623`, `utils.go:3632`) are `merr.WrapErrParameterInvalidMsg`, so the hardcoded value is always `1100`. This is therefore a defensive/latent concern rather than a live divergence — it would only bite if `genFunctionSchema` later grows a non-parameter error return. Note also that this is not a deviation introduced by this PR: every existing `genFunctionSchema` call site uses the same hardcoded form (`handler_v2.go:1027`, `1053`, `2482`). The only real inconsistency is intra-function: sibling branches in `addFunctionField` (e.g. `Schema.GetProto`, `convertToExtraParams` at lines 1279 and 1300) use `merr.Code(err)`.

*Suggestion:* switch to `merr.Code(err)` to remove the stylistic split inside `addFunctionField` and eliminate future drift risk. If you take this, consider updating `1027`, `1053`, and `2482` in the same pass — changing only one site would leave the codebase more fragmented than it is now. Fine to skip entirely if you'd rather keep the change surface minimal.

*(raised by wzy)*

**`client/milvusclient/collection_options.go:650`** — *api-consistency*

`addFunctionFieldOption.Validate()` client-side rejects a bound index whose index type resolves to empty (`ErrParameterMissing`), while the REST path applies no such check: `FunctionFieldIndexParam.IndexType` carries no `binding:"required"` (`request_v2.go:127-132`), `addFunctionField` never checks it for emptiness, and `convertToExtraParams` (`utils.go:3104-3116`) simply omits the `index_type` key when it's empty, letting the server decide. So the same semantic input is rejected by the client on one path and handled by the server on the other.

The scope here is narrower than it may first appear: standard constructors (`NewAutoIndex`, `NewSparseInvertedIndex`, `NewMinHashLSHIndex`) all carry their own index type and pass validation — the new `autoindex_is_forwarded_to_server` test in `client/milvusclient/collection_test.go` confirms AUTOINDEX is forwarded with `index_type == AUTOINDEX`. Only an explicitly type-less `index.NewGenericIndex("", ...)` trips the check.

*Suggestion:* no behavior change recommended — server-authoritative index type resolution is not a bug. Worth a short godoc note on the SDK option stating that an explicit index type is required on the `GenericIndex` path, whereas REST permits omitting `indexType` and defers to the server.

*(raised by wzy)*

sijie-ni-0214:
Thanks for the verification. Both are intentional scope decisions, so no further code change is planned:

1. `genFunctionSchema` currently returns parameter-invalid errors only, and the hardcoded `ErrParameterInvalid` response code is the established form in the sibling function handlers. Changing only `addFunctionField` would introduce a new local inconsistency; changing all call sites is unrelated cleanup.

2. The Go SDK requires a concrete index type when a bound index is supplied. The only affected case is an explicitly type-less `NewGenericIndex("", ...)`; the standard index constructors provide their type and preserve the expected request. REST intentionally leaves an omitted `indexType` to server-side resolution. We will keep that SDK validation contract rather than make the two surfaces behave identically for this malformed generic-index input.

czs007:
Adversarial review surfaced two issues — a concurrency race that lets a stale schema re-populate the cache after invalidation, and a metrics-label collision that makes the new drop-field route unobservable.

## Medium

**`client/milvusclient/collection.go:209` — cache eviction cannot stop an in-flight fetch from re-inserting a stale schema**
After a successful mutation, `alterCollectionSchemaRequest` calls `c.collCache.Evict(collName)` (collection.go:209) to invalidate the cache, but the singleflight callback in `GetCollection` unconditionally runs `c.collections.Insert(collName, coll)` (common.go:34) with no check for an intervening `Evict` and no generation/epoch guard. Race: goroutine A enters `sf.Do` and begins fetching schema vN *before* a field drop; the drop succeeds and goroutine B runs `Evict` (map now empty); A then returns and re-inserts the stale vN. Additionally, requests issued after the drop but while A is still in flight join the same singleflight by key (common.go:28) and receive A's stale vN directly. The result is that an old schema can persist in the cache long after a successful mutation, breaking this PR's guarantee that the cache is invalidated on success. `retryIfSchemaError` (common.go:60) self-heals write paths that surface `ErrCollectionSchemaMismatch`, but does not cover the direct `getCollection` read path, and existing tests are all serial. Suggestion: introduce a generation counter on the cache entry — increment it on `Evict`, have the singleflight callback `Insert` only when the generation is unchanged, and prevent stale-generation results from being reused by later requests. *(raised by gzh)*

## Low

**`internal/distributed/proxy/httpserver/handler_v2.go:102` — drop-field REST route uses a generic metric label and cannot be observed on its own**
The new `/v2/vectordb/collections/fields/drop` route maps to the generic `"AlterCollectionSchema"` in `routeToMethod` (handler_v2.go:102), sharing one `methodTag` with `/collections/add_function_field` and `/collections/drop_function_field` (handler_v2.go:93-94) and with native gRPC `AlterCollectionSchema`. Since `methodTag` is the `function_name` label of `metrics.ProxyFunctionCall` (handler_v2.go:411-427), these four distinct REST/gRPC operations collapse under `function_name="AlterCollectionSchema"`, so drop-field cannot be measured, rate-limited, or alerted on per operation. The sibling route `/collections/fields/add` uses the operation-level label `"AddCollectionField"` (handler_v2.go:101), so the granularity is inconsistent. Suggestion: map `fields/drop` to an operation-level label (e.g. `"DropCollectionField"`) to match `fields/add`. *(raised by zack)*

sijie-ni-0214:
Thanks for the adversarial review. I rechecked both points against the current master implementation.

1. The underlying cache race predates this PR. `CollectionCache.GetCollection` has unconditionally inserted the fetched value after `singleflight.Do` since the cache was introduced, and master already calls `Evict` from `retryIfSchemaError` without coordinating it with in-flight fetches. This PR adds another eviction call site after successful schema mutations, so it exposes the existing cache limitation in that scenario, but it does not introduce the underlying cache implementation. A complete fix requires shared `CollectionCache` generation/singleflight coordination and should be handled as a separate follow-up rather than only inside these three APIs. No cache-concurrency change is planned in this PR.

2. `routeToMethod` is used for the `ProxyFunctionCall`/log method tag, but it is not used to select the rate-limit bucket. `CheckLimiter` derives the rate type from the protobuf request through `proxy.GetRequestInfo` and calls `limiter.Check` with that rate type, so sharing the `AlterCollectionSchema` metric label does not collapse or change rate limiting. The label intentionally represents the actual backend RPC and is consistent with the existing `add_function_field` and `drop_function_field` routes. The REST operation is still distinguishable by its HTTP path/access logs. Introducing synthetic operation-level labels for unified-RPC routes would be a broader metrics-convention change, so no change is planned here either.

czs007:
## Overview

Adversarial review surfaced four issues, three of them backward-compatibility regressions caused by routing field-add operations through `AlterCollectionSchema`, plus one silent metrics-label break.

---

### High

**`client/milvusclient/collection.go:244` — `AddCollectionField` regresses for external collections; the `Unimplemented`-only fallback never fires**

The SDK now calls `AlterCollectionSchema` first and only falls back to the legacy `AddCollectionField` RPC when `grpcstatus.Code(err) == codes.Unimplemented` (`collection.go:243-247`). However, the server rejects external collections at the business level: `AlterCollectionSchema` explicitly blocks them (`internal/proxy/impl.go:1040-1046`), while legacy `AddCollectionField` (`impl.go:900-957`) has no such check — the two paths are not capability-equivalent. That rejection arrives as an `AlterStatus`, is converted by `alterCollectionSchemaRequest` (`collection.go:204-216`) into a client-side `*merr.RPCError`, and `RPCError` does not implement `GRPCStatus()` (`client/internal/merr/merr.go:41-86`), so `grpcstatus.Code(err)` yields `codes.Unknown` and the fallback is skipped. The same call therefore succeeds on an old server and fails on a new one — a public API capability regression. The PR's own test already encodes it: `tests/go_client/testcases/external_table_refresh_test.go:1575` asserts `ErrorContains("alter collection schema operation is not supported for external collection")`, after which the test must bypass the SDK entirely via the raw RPC `mc.GetService().AddCollectionField(...)` (same file, line 1582).

*Suggestion:* route external collections to the legacy RPC branch (or allow the add-field sub-operation of `AlterCollectionSchema` for external collections). At minimum, document the removal explicitly in the release note. Broadening the fallback condition beyond `Unimplemented` would also help here.

*(raised by tinswzy)*

**`internal/distributed/proxy/httpserver/handler_v2.go:1287` — REST `/fields/add` switched to `AlterCollectionSchema`, unconditionally removing external-collection field addition**

`addCollectionField` now builds an `AlterCollectionSchemaRequest` and calls `h.proxy.AlterCollectionSchema` (`handler_v2.go:1272-1293`); it no longer calls `AddCollectionField`, and there is no legacy fallback on the REST side. Since `proxy.AlterCollectionSchema` returns `ParameterInvalid` for external collections (`internal/proxy/impl.go:1040-1046`) while the old path had no such restriction (`impl.go:900-957`, and the e2e test confirms it succeeds for external collections), adding a field to an external collection over REST breaks unconditionally on new servers. Unlike the SDK case, REST clients have no workaround at all — legacy `AddCollectionField` has no REST route — so the only escape is switching to gRPC.

*Suggestion:* keep a legacy RPC branch (e.g. dispatch to `AddCollectionField` when the target is an external collection), or state the capability removal explicitly in the PR's Compatibility section. The PR body currently only says "no REST client fallback is required" and does not mention this loss.

*(raised by tinswzy)*

---

### Medium

**`internal/distributed/proxy/httpserver/handler_v2.go:1287` — required privilege changes from `AddCollectionField` to `AlterCollectionSchema`; roles granted only the fine-grained legacy privilege will be denied**

Privileges are resolved from the proto extension on the request message type (`PrivilegeInterceptor` → `funcutil.GetPrivilegeExtObj(req)`, `internal/proxy/privilege_interceptor.go:62`). Verified behavior: `AlterCollectionSchemaRequest` → `PrivilegeAlterCollectionSchema`, `AddCollectionFieldRequest` → `PrivilegeAddCollectionField` (both collection-level). This PR makes both REST `/fields/add` (`handler_v2.go:1287` — `checkAuth` receives the `AlterCollectionSchemaRequest`) and the Go SDK `AddCollectionField` (`collection.go:243`) send `AlterCollectionSchemaRequest`. On existing clusters, any role granted only the fine-grained `AddCollectionField` privilege (without `AlterCollectionSchema`, and not using the built-in `CollectionReadWrite` group) will now get permission denied when adding a field. The built-in privilege groups contain both (`pkg/util/constant.go:115-116`, `358-359`), so only fine-grained grants are affected — and because the failure is not `Unimplemented`, the SDK fallback cannot catch it either (see the first issue).

*Suggestion:* either fall back to the legacy RPC on `PermissionDenied` as well in the SDK/REST paths, or call out the re-grant requirement as an explicit breaking change in the PR's Authorization section and in the release note.

---

### Low

**`internal/distributed/proxy/httpserver/handler_v2.go:101` — `routeToMethod` changes the `/fields/add` metric `function_name` from `AddCollectionField` to `AlterCollectionSchema`, breaking existing monitoring series**

`routeToMethod` is the sole source of the `function_name` label for REST-side `metrics.ProxyFunctionCall` (`handler_v2.go:411-421`; label defined at `pkg/metrics/metrics.go:130`). Changing the mapping for `/v2/vectordb/collections/fields/add` (`handler_v2.go:101`) means existing dashboards and alerts matching `function_name="AddCollectionField"` on REST traffic will drop to zero after upgrade and reappear under the new value — a silent monitoring break worth a release-note line.

Note on the related concern that add and drop now share a label: this matches the map's existing convention (the label is the invoked gRPC method, not the REST route — e.g. `add_function_field` / `drop_function_field` have long both mapped to `AlterCollectionSchema`, and `alter_properties` / `drop_properties` both to `AlterCollection`), so it is not introduced by this PR; distinguishing them would be a separate, map-wide change. Also, the earlier reference to a "previous labels-split fix (b0f540616a)" does not hold — that commit is not on master; it is an earlier revision of this PR's own branch.

*Suggestion:* mention the label change in the release note so operators can update dashboards and alerts.

*(raised by tinswzy)*

czs007:
## Adversarial review summary

Adversarial verification kept 3 findings: two high-severity regressions where routing add-field through `AlterCollectionSchema` breaks field addition on external collections (Go SDK and REST), plus one low-severity observability change.

### High

**1. `client/milvusclient/collection.go:244` — external-collection fallback in `AddCollectionField` never triggers**

`AddCollectionField` now calls `AlterCollectionSchema` first and falls back to the legacy `AddCollectionField` RPC only when `grpcstatus.Code(err) == codes.Unimplemented`. For external collections the server rejects `AlterCollectionSchema` with a business error carried in `AlterStatus` (`internal/proxy/impl.go:1040`); the client converts it via `merr.CheckRPCCall` → `Error(status)` into a client-side `milvusError`, not a gRPC status error, so `grpcstatus.Code(err)` yields `codes.Unknown` and the fallback branch is never taken. The legacy `Proxy.AddCollectionField` (`internal/proxy/impl.go:900-920`) has no external-collection guard and has always supported this, so a previously working public SDK operation now fails. The PR's own test change confirms the behavior shift: `tests/go_client/testcases/external_table_refresh_test.go:1575` asserts the error and then bypasses the SDK via `mc.GetService().AddCollectionField(...)`.

*Suggestion:* allow "pure add plain field" through `AlterCollectionSchema` server-side for external collections, or broaden the client fallback to also match the returned `milvusError` (not just transport-level `Unimplemented`). Please restore the original e2e assertion rather than routing the test around the public API. *(raised by @chyezh, @tinswzy)*

**2. `internal/distributed/proxy/httpserver/handler_v2.go:1272` — REST `POST /v2/vectordb/collections/fields/add` breaks on external collections, with no fallback at all**

Same root cause, different public surface. The handler no longer calls `h.proxy.AddCollectionField`; it builds an `AlterCollectionSchemaRequest` and calls `h.proxy.AlterCollectionSchema`. Since that path unconditionally rejects external collections, REST clients adding a field to an external collection go from success to `1100 alter collection schema operation is not supported for external collection`. Unlike the Go SDK, the REST handler contains no external-collection check and no fallback code, and the PR adds no REST test for this path. The Compatibility section of the PR description covers old/new server compatibility but not this narrowing of external-collection semantics.

*Suggestion:* fix server-side as in (1), or branch in the handler on `schemaProto.GetExternalField()` / the collection's external flag and dispatch to the legacy RPC. Add a REST test covering add-field on an external collection. *(raised by adversarial-audit)*

### Low

**3. `internal/distributed/proxy/httpserver/handler_v2.go:101` — REST add/drop field metrics collapse into one `AlterCollectionSchema` label**

`routeToMethod` changes `/v2/vectordb/collections/fields/add` from `AddCollectionField` to `AlterCollectionSchema`, and the new `/fields/drop` route reuses the same label. That value feeds the `function_name` label of `metrics.ProxyFunctionCall` and the `method` field of failure logs (`handler_v2.go:411-432`, `:445`). Two consequences: existing dashboards/alerts filtering on `function_name="AddCollectionField"` silently go to zero after upgrade, and add-field vs. drop-field become indistinguishable in metrics and logs. For fairness: collapsing is not unprecedented — `add_function_field` and `drop_function_field` already share the `AlterCollectionSchema` label on master — so the issue here is the unannounced rename of an existing label, not the pattern itself.

*Suggestion:* give add and drop distinct labels reflecting the REST semantics, or at minimum call out the label rename in the PR description so operators can update dashboards. *(raised by @chyezh, @tinswzy)*

sijie-ni-0214:
Thanks for the follow-up review. I rechecked the three findings against the current master.

1. The Go SDK external-collection regression is valid against the current master. This PR depends on https://github.com/milvus-io/milvus/pull/50394, which removes the blanket external-collection rejection from `AlterCollectionSchema` and implements the complete validation, refresh, and serving semantics for external schema evolution. This PR should not merge before #50394. After #50394 lands, I will rebase and restore the external AddField E2E to verify that `Client.AddCollectionField` succeeds directly through `AlterCollectionSchema`, instead of bypassing the public API through the legacy RPC. I do not plan to add a fallback for this business-level error, since that would duplicate and potentially bypass the unified RPC's server-side capability policy.

2. The REST `/fields/add` issue has the same root cause and is addressed by the same dependency. Once #50394 lands, the REST handler can continue routing through `AlterCollectionSchema` while preserving external AddField support. I will retain/add focused coverage for this path after the rebase rather than introduce a REST-side legacy-RPC fallback.

3. Agreed that the existing `/fields/add` metric/log method label changes from `AddCollectionField` to `AlterCollectionSchema`. The current label follows the actual backend RPC, consistent with the existing function-field routes. I will call out this observability migration in the PR compatibility notes so dashboards and alerts using the old label can be updated, rather than introduce synthetic operation-level labels only for these routes.

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/f8256ecded858642f36115843750a915b0dcdff2...czs007:team2-fix/T5663751d31?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/f8256ecded858642f36115843750a915b0dcdff2...czs007:team2-fix/T6252962085?expand=1

czs007:
## Adversarial review summary

Two issues surfaced during post-verification review: one medium-severity gap where an untested rolling-upgrade fallback path is also functionally broken, and one nitpick about a wrapper that silently drops an inner error code.

### 🟠 Medium

- **`client/milvusclient/collection.go:244` — Rolling-upgrade fallback (business `ErrServiceUnimplemented`) is untested and broken.** `AddCollectionField` pivots the legacy fallback on `grpcstatus.Code(err) != codes.Unimplemented`, but only the raw-transport `Unimplemented` path (nil resp + grpc status error) is covered by tests. In the new-Proxy → old-Coord scenario, the proxy's mixcoord client converts the coord's transport `Unimplemented` into a code-10 business error (`internal/util/grpcclient/client.go:436`), which `alterCollectionSchemaTask.Execute` serializes into a non-nil `AlterCollectionSchemaResponse.AlterStatus` (err=nil). The SDK reconstructs this as a plain `*RPCError` with code 10 (no `GRPCStatus()`), so `grpcstatus.Code()` returns `codes.Unknown`, the `!= codes.Unimplemented` check is true, and the legacy fallback is skipped — surfacing a hard error to users during the mixed-version window. **Suggestion:** add a test case where `AlterCollectionSchema` returns `&AlterCollectionSchemaResponse{AlterStatus: merr.Status(<code-10 error>)}` (resp non-nil, err nil) and assert the legacy `AddCollectionField` RPC is invoked; the assertion will currently fail, exposing the gap, and the fallback condition should be widened to also treat the code-10 business `ErrServiceUnimplemented` as a fallback trigger. *(raised by chyezh, tinswzy)*

### 🔵 Nitpick

- **`client/internal/merr/merr.go:336` — `WrapErrParameterInvalidErr` discards the inner error code.** The helper folds `err.Error()` into the message and clones `ErrParameterInvalid`, dropping any code the inner error carries. This is harmless for both current callers (`collection.go:237` wraps a `proto.Unmarshal` error, `collection_options.go:499` wraps a `json.Unmarshal` error — both client-side stdlib syntax errors with no meaningful merr code). **Suggestion:** add a doc comment noting the inner code is intentionally dropped, so it isn't reused where the inner code matters. *(raised by tinswzy)*

sijie-ni-0214:
Thanks for the post-verification review. Fixed in a0b8a91239.

1. AddCollectionField now falls back for both forms of service unavailability:
   - transport-level gRPC codes.Unimplemented from an older proxy;
   - Milvus ErrServiceUnimplemented (code 10) returned in AlterCollectionSchemaResponse.AlterStatus when a new proxy talks to an older coordinator.

   The new unit test returns a non-nil AlterCollectionSchemaResponse with AlterStatus code 10 and verifies that the legacy AddCollectionField RPC is invoked. Other transport and business errors still return without fallback.

2. Added a doc comment to WrapErrParameterInvalidErr clarifying that it deliberately reclassifies the error as ErrParameterInvalid and retains only the original text, not the inner identity or code. Its current proto/json parsing callers are unchanged.

Validation: make static-check passed with 0 issues, and the complete client/internal/merr and client/milvusclient package tests passed.

liliu-z (commented):
/lgtm
/approve

PR #51323, opened 2026-07-13
enhance: move JSON indexes into sealed segment runtime state

congqixia:
Related to #51068

- manage JSON index load, replace, and drop through reopen COW snapshots
- remove LoadScope_Index and the legacy QueryCoord-to-segcore update path
- drop JSON indexes by field and nested path to preserve sibling indexes
- regenerate query proto and QueryNode mocks

Issue #51068, opened 2026-07-06
[Enhancement]: Evolve `ChunkedSegmentSealedImpl` toward atomic COW runtime publication


congqixia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Introduce a long-term enhancement plan for `ChunkedSegmentSealedImpl` so that reader-visible mutable runtime state is updated through a unified **copy-on-write (COW) runtime publication model** instead of direct live-member mutation.

The target model is:

- readers capture one published snapshot and read both metadata and runtime ownership from it
- updaters prepare new runtime state on a cloned runtime graph
- publish atomically switches `(schema + load_info + readiness + runtime)`
- finalize only performs post-publication cleanup or follow-up work
- old readers keep the old runtime graph alive through shared ownership until they naturally retire

This enhancement should be treated as the architectural default for future `ChunkedSegmentSealedImpl` reopen/load/update work, not as a one-off fix for a specific path.

Concretely, the enhancement direction includes:

1. Converge reader-visible runtime state onto `PublishedSegmentState::runtime`
2. Avoid new mixed-state patterns where readers combine published metadata with live runtime members
3. Ensure readiness and runtime visibility describe the same published world
4. Continue migrating deferred runtime resources such as:
   - `vector_indexings_`
   - `text_indexes_`
   - json-related runtime state
   - any future reader-visible mutable runtime helper state
5. Require future `ChunkedSegmentSealedImpl` feature work to justify any state that remains outside the COW runtime bundle

### Why is this needed?

`ChunkedSegmentSealedImpl` handles many update paths while the segment is still queryable, including field loads, column-group loads, reopen/load-diff transitions, and multiple kinds of index/runtime state updates.

Without a unified COW publication model, the implementation is vulnerable to the same class of correctness problems repeating across different resources:

- published schema/load info switches before runtime state is ready
- runtime object is replaced before published readiness catches up
- prepare-time failures or cancellation leave uncommitted live-state residue
- readers observe mixed state composed from snapshot metadata and live mutable members
- old/new reader lifetime is hard to reason about across reopen publication boundaries

These are not isolated bugs tied to one field type or one helper. They are symptoms of the same structural problem: metadata publication and runtime ownership are not fully unified.

The enhancement is needed to establish one stable rule for sealed-segment evolution:

> if runtime state is reader-visible and mutable during reopen/load/update, it should be prepared on cloned state and published atomically with metadata whenever possible.

[...]

PR #51413, opened 2026-07-15
enhance: [3.0] move JSON indexes into sealed segment runtime state (#51323)

congqixia:
Cherry pick from master
pr: #51323
Related to #51068

- manage JSON index load, replace, and drop through reopen COW snapshots
- remove LoadScope_Index and the legacy QueryCoord-to-segcore update path
- drop JSON indexes by field and nested path to preserve sibling indexes
- regenerate query proto and QueryNode mocks

---------


liliu-z (commented):
/lgtm
/approve

Issue #51381, opened 2026-07-15
[Bug]: Struct-array element match conflates NULL rows with empty arrays and races schema evolution


xiaofan-luan:
## Is there an existing issue for this?
- [x] I have searched the existing issues

## Environment
master

## Current Behavior
Under pg-aligned three-valued logic a NULL array row and a real empty array [] are different inputs: MATCH_ALL over [] is a vacuous match, over NULL it is UNKNOWN (excluded, including under NOT). The struct-array machinery conflates them:
1. Array offsets record no row validity; schema-evolution backfill builds all-ZEROS offsets, so backfilled (NULL) rows vacuously match MATCH_ALL.
2. MATCH_* does not mask NULL rows: `not MATCH_ANY(...)` wrongly includes NULL rows.
3. The row-level nested-struct-index fold cannot recover row nullity from element-level index validity: `not array_contains(nullable_sub, x)` wrongly includes NULL rows.
4. `json_contains_all(array_field, [])` returns blanket TRUE including NULL rows (pg: NULL @> '{}' yields NULL).
5. `array_offsets_map_` / `struct_representative_fields_` are mutated by schema-evolution Reopen with no lock while query/insert paths read them concurrently (data race, UB).
6. `BitmapIndex::HasRawData()` returns true for non-nested ARRAY indexes, permitting index-only loads that leave struct MATCH without array offsets (query-time abort).

## Expected Behavior
NULL vs [] distinguished end-to-end (offsets row validity + MATCH NULL masking + index-fold validity recovery); empty-set ContainsAll keeps IS NOT NULL semantics; map access synchronized; ARRAY indexes never claim raw data.

## Anything else?
Behavior change needs a release note (NOT-of-MATCH and empty-list ContainsAll results change for NULL rows).

yanliang567:
@zhuwenxing please keep an eye on this issue and check if we need more coverage on this scenario.

/assign @zhuwenxing

zhuwenxing:
Validated against the latest master (`b9621d7d325d`) with PyMilvus `3.1.0rc64`. The fix from #51415 is working as expected.

The following E2E scenarios now pass:

- MATCH NULL-versus-empty semantics for sealed and growing rows.
- Nullable Struct Array validity before and after nested scalar index creation and reload.
- Schema-evolution backfill semantics after release/reload.

The tests confirm that NULL rows are excluded from MATCH evaluation while real empty arrays retain zero-element semantics.

PR #51415, opened 2026-07-15
fix: fix three-values nullbility and race in growing segmentArrayOffets

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/51381

### DIFF
--- a/internal/core/src/segcore/SegmentGrowingImpl.cpp
+++ b/internal/core/src/segcore/SegmentGrowingImpl.cpp
@@ -344,6 +344,8 @@ SchemaHasTextField(const Schema& schema) {
 
 void
 SegmentGrowingImpl::InitializeArrayOffsets() {
+    std::unique_lock lock(array_offsets_map_mutex_);
+
     // Group fields by struct_name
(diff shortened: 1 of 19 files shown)


internal/core/src/segcore/SegmentGrowingImpl.cpp
```diff
@@ -283,31 +283,81 @@ ExtractArrayLengths(const proto::schema::FieldData& field_data,
         // ARRAY: extract from scalars().array_data().data(i)
         const auto& array_data = field_data.scalars().array_data();
         auto element_type = field_meta.get_element_type();
+        bool compact_nullable = false;
+        if (field_meta.is_nullable()) {
+            int64_t valid_count = 0;
```
zhengbuqian: what does it mean if `field_data.valid_data_size() != num_rows`? and does valid_count left uninitialized(as 0)?

zhengbuqian: also does `field_data.valid_data_size() != num_rows` mean corrupt input? if so we should error

zhengbuqian (approved):

Issue #51419, opened 2026-07-15
[Enhancement]: Decouple the Go client from milvus/pkg and restore Go 1.24 compatibility`


congqixia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Make the standalone Go client module (`github.com/milvus-io/milvus/client/v2`) independent of the server-side `github.com/milvus-io/milvus/pkg/v3` module.

The client currently imports several helpers and constants from `milvus/pkg`, including:

- RPC error conversion and classification helpers
- retry utilities
- generic concurrent map, set, and vector conversion helpers
- generic singleflight wrappers
- authentication encoding helpers
- common collection, field, and index property keys

These imports require `client/go.mod` to depend on `github.com/milvus-io/milvus/pkg/v3` and use a local `replace` directive. As a result, the standalone client inherits the server package's dependency graph and Go toolchain requirements.

The client should instead contain a minimal client-local implementation of the required helpers:

- Add client-local packages under `client/internal` for error handling, retry, type utilities, singleflight, and authentication encoding.
- Add the required public property keys to `client/common`.
- Replace all client imports of `milvus/pkg` with the client-local packages.
- Expose structured client RPC errors and commonly used error sentinels from `milvusclient`.
- Remove the `github.com/milvus-io/milvus/pkg/v3` requirement and local replacement from `client/go.mod`.
- Lower the client module's Go version from Go 1.26.4 to Go 1.24.9 and select dependency versions compatible with Go 1.24.

The client-local RPC error implementation must preserve the existing wire contract and client behavior:

- Modern Milvus numeric error codes
- Deprecated `commonpb.ErrorCode` mappings
- Error reason and detail
- Retriable metadata
- Input-error metadata
- `errors.Is` matching by Milvus error code
- Context cancellation and timeout codes
- Transport errors and nil-response handling

### Why is this needed?

The Go SDK is a standalone client library and should not require server-only packages.

Depending on `milvus/pkg` causes several problems:

1. **Go toolchain coupling**

   The client inherits the server module's minimum Go version. After the server-side toolchain was upgraded to Go 1.26.4, users could no longer build the client module with Go 1.24 even though the client code itself did not require Go 1.26 features.

2. **Excessive transitive dependencies**

   Importing a small number of utility functions from `milvus/pkg` pulls in server-side observability, storage, process-management, Kubernetes, and other dependencies that are unrelated to an SDK consumer.

3. **Standalone module portability**

[...]

PR #51420, opened 2026-07-15
enhance: [GoSDK] decouple Go client from milvus pkg module

congqixia:
Related to #51419

- migrate required error and utility helpers into the client module
- preserve RPC error codes, retry metadata, and errors.Is semantics
- expose client-local RPC errors and common error sentinels
- remove github.com/milvus-io/milvus/pkg/v3 and server-only dependencies
- lower the client Go version to 1.24.9


client/internal/crypto/crypto.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crypto
+
+import "encoding/base64"
+
+func Base64Encode(value string) string {
```
zhengbuqian, line 21: looks redundant to me

zhengbuqian (approved):
/lgtm

Issue #51407, opened 2026-07-15
[Bug]: RefManager leaks file resource references when downloads fail


bigsheeper:
## Environment

- **Milvus Version**: master at `6ca035094475865e72e90526de2aff4736f9b1b8`
- **Deployment Mode**: cluster
- **MQ**: any
- **SDK**: not applicable
- **OS**: any

### Option B: Steps

1. Configure a DataNode to use file resources in `ref` mode.
2. Start a compaction or stats task that builds a text index with two file resources.
3. Let the first resource download succeed and make the second object-storage read fail, for example with a transient storage error.
4. Retry the task, then allow the file-resource GC loop to run.
5. Inspect the `RefManager` reference map and local file-resource directory.

### Trigger Conditions

- **Frequency**: always when `RefManager.Download` returns an error after reserving the batch
- **First observed after**: the first failed download
- **Does NOT happen when**: all resources download successfully and the caller executes `Release`

## Expected Behavior

A failed `Download` call should undo the reference reservations acquired by that call. Partially downloaded resources should become eligible for the existing GC cleanup.

## Actual Behavior

`RefManager.Download` increments every requested resource reference before I/O and returns immediately on an error without decrementing them. The compaction and stats callers install `defer Release` only after `Download` succeeds, so failed calls leave references permanently positive. Retries increment the counts again, preventing `CleanResource` from deleting stale local files and allowing disk usage to accumulate.

## Error Logs

The original object-storage error is returned. No additional error indicates that the file-resource references were leaked.

## Non-default Configuration

```yaml
common:
  fileResource:
    mode:
      dataNode: ref
```

## Analysis Hints

- Suspect code: `internal/util/fileresource/manager.go`, `RefManager.Download`
- Production callers:
  - `internal/datanode/compactor/compactor_common.go`
  - `internal/datanode/index/task_stats.go`
- `Download` reserves the full resource batch before I/O.
- Both callers call `Release` only after a successful `Download`.
- `CleanResource` removes only entries whose reference count is less than or equal to zero.
- A deterministic unit test can download one resource, fail the next, and assert that the complete batch reservation is rolled back.

PR #51424, opened 2026-07-15
fix: roll back file resource refs on download failure

bigsheeper:
issue: #51407

## What this PR does

- Roll back the full batch of file resource reference reservations when `RefManager.Download` fails.
- Keep successful download ownership unchanged so callers release references after use.
- Add a regression test covering a multi-resource batch that fails after a partial download.

## Verification

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/util/fileresource`
- `make static-check`


czs007:

czs007 (approved):
/lgtm

Issue #51425, opened 2026-07-15
[Bug]: TiKV metastore: transient region errors (epoch_not_match) escape txn commit as hard failures and crash streamingnode


shaoting-huang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: master (9e11bdae4e)
- Deployment mode(standalone or cluster): cluster (5 components, separate processes)
- MQ type(rocksmq, pulsar or kafka): pulsar
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus 3.1.0rc62
- OS(Ubuntu or CentOS): Ubuntu (kernel 6.8)
- CPU/Memory: 12 cores / 31 GB
- GPU: none
- Others: metastore.type=tikv, single-replica TiKV v8.5.2 (etcd kept for service discovery)

### Current Behavior

When Milvus runs with the TiKV metastore, a transient TiKV region error can escape `txn.Commit()` and get treated as a hard, non-retriable IO failure. Under a full e2e-suite write load, a region split bumped the region epoch (54 -> 55) while the single store was briefly unreachable for a few seconds. client-go exhausted its internal backoff budget and `Commit()` returned the raw region error string. `txnTiKV.MultiSave()` wrapped it as `ErrIoFailed` and propagated it:

```
[WARN] [broker/datacoord.go:135] ["failed to SaveBinlogPaths"] [error="epoch_not_match:<> : IO failed: Failed to commit for MultiSave()"]
panic: epoch_not_match:<> : IO failed: Failed to commit for MultiSave() [recovered, repanicked]
  github.com/milvus-io/milvus/internal/flushcommon/writebuffer.defaultWBOption.func1
  github.com/milvus-io/milvus/internal/flushcommon/syncmgr.(*SyncTask).HandleError
```

The sync path treats the failed meta save as fatal by design, so streamingnode panics and the whole write path goes down. TSO allocation (whose counter also lives in the metastore) degraded in the same window (`syncTimeStamp Failed: can not get timestamp`).

`epoch_not_match` (and similar region errors) are retriable by TiKV semantics: refreshing the region cache and re-running the transaction succeeds. The etcd metastore has no equivalent exposure since etcd has no region concept, so this failure mode is specific to the TiKV backend and more likely on small/single-replica TiKV deployments.

### Expected Behavior

[...]

PR #51426, opened 2026-07-15
fix: [Catalog] retry transient TiKV write transaction failures

shaoting-huang:
- Rebuild TiKV write transactions on retriable begin, build-stage IO, and classified-transient commit failures while aborting deterministic build failures and unsafe commit outcomes.
- Preserve original errors for TiKV undetermined commit results and caller cancellation before string-wrapping; keep private `retry.Unrecoverable` markers inside the retry helper so they never leak to callers.
- Stop `ReliableWriteMetaKv` from replaying undetermined TiKV write results, correct the TiKV begin/requestTimeout comment, and add regression tests for commit aborts, retry exhaustion, deadline stopping, and outer reliable-write behavior.

related: #51425

### DIFF
--- a/internal/kv/tikv/txn_tikv.go
+++ b/internal/kv/tikv/txn_tikv.go
@@ -62,6 +62,40 @@ const (
 
 var Params *paramtable.ComponentParam = paramtable.Get()
 
+var (
+	// errPredicateNotMet marks deterministic predicate mismatches inside write transaction builds.
+	errPredicateNotMet = errors.New("predicate not met")
+	errBuildAbort      = errors.New("deterministic build error")
+)
+
+func wrapWriteBuildErr(reason string, err error) error {
+	wrapped := merr.WrapErrIoFailedReason(reason, err.Error())
+	if isLocalMutationValidationErr(err) {
+		return markBuildAbort(wrapped)
+	}
+	return wrapped
+}
+
+func markPredicateNotMet(err error) error {
+	return markBuildAbort(errors.Mark(err, errPredicateNotMet))
+}
+
+func markBuildAbort(err error) error {
+	return errors.Mark(err, errBuildAbort)
+}
+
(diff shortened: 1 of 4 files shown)


internal/kv/tikv/txn_tikv.go
```diff
@@ -635,6 +593,43 @@ func (kv *txnTiKV) WalkWithPrefix(ctx context.Context, prefix string, pagination
...
+		txn, err := beginTxn(kv.txn)
+		if err != nil {
+			return merr.WrapErrIoFailedReason(fmt.Sprintf("Failed to create txn for %s", op), err.Error())
+		}
+		// Defer a rollback only if the transaction hasn't been committed
+		defer rollbackOnFailure(&attemptErr, txn)
+
+		if buildErr = build(txn); buildErr != nil {
+			attemptErr = buildErr
+			return retry.Unrecoverable(buildErr)
```
liliu-z: runWriteTxnWithRetry wraps every non-nil `build` error in `retry.Unrecoverable`, but the `build` closure for MultiSaveAndRemoveWithPrefix performs real read RPCs — `txn.Iter`/`iter.Next` over the prefix range (a live production path: datacoord/rootcoord call it with no predicates) and, when predicates are passed, `txn.Get` — that can fail transiently on TiKV region/leader changes. Those reads return WrapErrIoFailedReason (ErrIoFailed = SystemError), which retry.Do would normally retry, so the blanket Unrecoverable is exactly what forces them to abort on the first failure, re-opening the fail-fast gap this PR closes for commits. Scope the marker to the genuinely deterministic build errors (the unmet-predicate check; the reserved-value error is already non-retriable by type) so transient prefix-scan reads stay retriable. The function's doc comment claiming build errors are all 'deterministic (bad input, unmet predicate)' is likewise inaccurate.

PR #50849, opened 2026-06-27
fix: [cp 2.6] ignore stale replicated txn body

chyezh:
issue: #50846
pr: #50848

- streaming replication: treat non-current txn messages covered by the replicate checkpoint as stale duplicates
- replicate tests: cover stale txn body retry while another txn is in progress

PR #50544, opened 2026-06-15
fix: [2.6] describe_user returns ghost empty role names after repeated grant/revoke (#50058)

shaoting-huang:
Cherry-pick from master
pr: https://github.com/milvus-io/milvus/pull/50058
Related to https://github.com/milvus-io/milvus/issues/49989

PR #50518, opened 2026-06-13
fix: [cp2.6]make SalvageCheckpoint reachable via GetReplicateInfo after force-promote

bigsheeper:
Cherry-pick from master

pr: #50350
issue: #50344

## Summary

Cherry-picked from master PR #50350 (merged)

## Verification

- [x] File count matches original PR
- [x] Code changes verified
- [x] No conflict markers
- [ ] make static-check (skipped by cherry-pick workflow)

PR #50387, opened 2026-06-08
fix: remove stale collection schema metadata on alter

sijie-ni-0214:
issue: #50388

This PR removes stale per-schema metadata keys when AlterCollection updates collection fields, struct array fields, or functions. It keeps collection schema metadata and per-field/function metadata consistent after DropField-style schema updates, so RootCoord reload and BirdWatcher do not observe dropped fields from stale `root-coord/fields` keys.

Validation:
- `gofumpt -l internal/metastore/kv/rootcoord/kv_catalog.go internal/metastore/kv/rootcoord/kv_catalog_test.go`
- `git diff --check`
- `source scripts/check_build.sh`
- `make SKIP_3RDPARTY=1`
- `make build-go`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/metastore/kv/rootcoord -timeout 300s`
- `make static-check`
- Local standalone DropField regression with pymilvus: verified API output/filter errors for dropped field, raw etcd dropped field key removal, BirdWatcher output, and restart persistence.

PR #51429, opened 2026-07-15
fix: [poc_2.6.17] backport 10 CDC + stability/correctness fixes

czs007:
Backport a curated set of **10 CDC + stability/correctness fixes** from 2.6/master onto `poc_2.6.17` (Apple).

Code-only: the CDC force-promote pytest (#49423) was intentionally left off this poc branch (test coverage lives on canonical 2.6; keeps the poc branch ruff-clean).

## Included (10) — cherry-picked with `-x`, backporter-signed-off, build-verified (`make build-cpp` + `make build-go` pass)

| fix | source |
|---|---|
| [2.6] bump Woodpecker to v0.1.13-hotfix to prevent preempted writers from skipping segment IDs | #49670 |
| [cp2.6]recover import reader from premature EOF | #49867 |
| [2.6] avoid panic for non-point WKT validation | #50205 |
| remove stale collection schema metadata on alter | #50387 |
| [2.6] describe_user returns ghost empty role names after repeated grant/revoke | #50544 |
| [cp2.6]make SalvageCheckpoint reachable via GetReplicateInfo after force-promote | #50518 |
| [2.6] avoid access log trace id panic | #50565 |
| avoid panic in tombstoneScheduler.AddPending during shutdown | #50550 |
| [2.6] preserve nullable expr rewrite semantics | #50627 |
| [cp 2.6] ignore stale replicated txn body | #50849 |

## Verification
- `make build-cpp` — 100% (incl. segcore expression + WKT geometry)
- `make build-go` — produces `bin/milvus`

## Scope note (intentional exclusions)
Other post-branch CDC/important fixes were evaluated and **excluded** because `poc_2.6.17` lacks the infra they depend on (nullable-validity / mapping-storage, storage-v2, emb-list, index-codec-v2, CDC txn-produce, snapshot, alter-drop-function), or because they target master's `pkg/v3`+`mlog` logging which poc has not migrated to. Notably #50722/#50723/#51135 (null/unknown expression semantics) need the `ApplyFieldValidData`/ProcessedResult infra absent in poc — one (#50722) even applied cleanly but failed to compile, caught by the build.


tinswzy:
## Overview
This backport of #50627 introduces `canFoldBoolDomainToConstant` / `hasMissingPathSemantics` but never wires them in, so AND-contradiction folds guard on nullability alone and wrongly collapse to a constant for non-nullable JSON/array-path columns.

## High

- **`internal/parser/planparserv2/rewriter/term_in.go:383`** — Empty-result / contradiction folds guard only on `hasNullableFieldSemantics(g.col)`, letting non-nullable but missing-path columns (fixed JSON keys, array subscripts, struct-array subscripts, which carry `NestedPath` with `Nullable=false`) slip through. When the path/index is absent the inner predicate is UNKNOWN, but the fold treats it as a valid false and lets an outer `NOT` become `AlwaysTrue`, wrongly returning the row (e.g. `not (JSONField["age"] > 100 and JSONField["age"] < 50)`, `not (ArrayInt[0] in [1] and ArrayInt[0] > 2)`, `not (ArrayInt[0] in [1] and ArrayInt[0] in [2])`, `not (ArrayInt[0] in [1] and ArrayInt[0] != 1)`, `not (ArrayInt[0] in [1] and ArrayInt[0] == 2)`). Master guards these same sites with `!canFoldBoolDomainToConstant(g.col)` via #50979 ("propagate unknown for missing nested values"); this PR omits it, and `canFoldBoolDomainToConstant` is left as dead code. Affected sites: term_in.go:188 (multi-equals contradiction), term_in.go:210 (single-equal-not-in-set), term_in.go:383 (IN+range), term_in.go:511 (IN+IN), term_in.go:590 (IN+!=), range.go:209 (`combineAndRangePredicates`), range.go:611 (`combineAndBinaryRanges`). Suggestion: guard every AND-contradiction / empty-intersection fold with `hasNullableFieldSemantics(col) || hasMissingPathSemantics(col)` (i.e. `!canFoldBoolDomainToConstant`), ideally by cherry-picking #50979 (which also fixes the C++/segcore execution side). Add non-nullable JSON/array-path tests under `NOT` for the IN+range, IN+IN, IN+!=, and JSON `>`/`<` and `==`/`==` cases; the current suite only exercises the `Nullable*` variants. _(raised by yezh, wzy)_

czs007:
Thanks @tinswzy — confirmed and fixed in 9b3accfab4.

You're right: this branch took the `#50627` backport (which added `canFoldBoolDomainToConstant`/`hasMissingPathSemantics`) but not `#50979`, so the AND-contradiction / empty-intersection folds still guarded on `hasNullableFieldSemantics` alone and let non-nullable missing-path columns (fixed JSON keys, array / struct-array subscripts with `NestedPath, Nullable=false`) fold to a constant, so an outer `NOT` wrongly collapsed to `AlwaysTrue`.

The new commit backports the **Go/rewriter portion of #50979**: every fold site you listed — term_in.go:188/210/383/511/590, range.go:209/611, plus combineOrInWithNotEqual — now guards on `!canFoldBoolDomainToConstant(g.col)` (nullable OR missing-path), and the redundant `canFoldInNotEqualTautologyToTrue` is removed. Rewriter build + unit tests pass.

One caveat I want to be explicit about: the **C++/segcore execution side of #50979 is not backported here** — `poc_2.6.17` lacks the `ApplyFieldValidData` / value-helper infra it depends on (the same reason #50722/#50723/#51135 were left off). This commit fixes the rewriter over-folding (the wrong-result path you reported); segcore UNKNOWN evaluation for these predicates remains a known gap on this branch, which I'll flag to the POC owners. I'll add the non-nullable `NOT` test cases (IN+range, IN+IN, IN+!=, JSON `>`/`<`, `==`/`==`) next.

czs007:
Tests added in 5a2f099dbc, as promised.

`TestRewrite_JSON_NestedPath_NonNullable_ContradictionsKeepPredicate` and `TestRewrite_NonNullableArrayIndex_ContradictionsKeepPredicate` cover every fixed fold site — IN+equal, IN+IN, IN+!=, IN+range (term_in.go) and range+range (range.go) — for both JSON missing-key (`JSONField["age"]`) and out-of-range array-index (`ArrayInt[0]`) columns, under both the bare AND form (must not fold to valid false) and the negated form (must not fold to valid true), including your exact repros.

Verified they're real regression tests: they **fail on the pre-fix guards** (e.g. `JSONField["age"] > 100 and JSONField["age"] < 50` folds to valid-false) and pass after. Full `./internal/parser/planparserv2/rewriter/` suite is green.

tinswzy:

PR #51430, opened 2026-07-15
doc: add resource-group balance epoch MEP

xiaofan-luan:
## Summary

Add a design document for a Resource-Group-scoped balance epoch in QueryCoord. The proposed control loop is:

`Observe -> Plan -> Admit -> Execute -> Reconcile -> Observe again`

Each Resource Group owns at most one current normal-balance generation, while different RGs remain independent.

## Motivation

Issue #51244 reports long-running ScoreBasedBalancer ping-pong and availability impact from repeated segment releases. Distribution-aware task deltas help, but they do not provide one immutable planning snapshot, hard RG-wide admission budgets, or a reconciliation boundary between generations.

## Design highlights

- immutable RG placement snapshots and pending-task revision
- typed, generation-aware scheduler admission
- hard per-wave kind/node/collection/object budgets
- snapshot-only ScoreBased and ChannelLevel planning with strict-decrease potential
- authoritative distribution reconciliation for completed, failed, and ambiguous outcomes
- active/shadow/legacy rollout with active mode disabled by default
- conservative ambiguous RPC carry-over until placement settles

## Implementation

Draft implementation PR: #51431

Final implementation revision: `8c7b55f32cdf031ead70e99afdace281434761bb`

## Validation evidence captured by the MEP

- 9/9 focused QueryCoord selectors passed
- focused race selector passed with no race report
- complete `internal/querycoordv2/...` suite passed
- final implementation re-review: 0 Critical, 0 Important, 0 Minor
- scoped ruleguard: `rawmerrerror: 0`

Repository-wide `make static-check` is not claimed green: it stops before ruleguard because the checkout lacks the unchanged generated package `cmd/tools/migration/legacy/legacypb`. The MEP records this limitation explicitly.

## Related

issue: #51244

implementation: #51431

prior fixes: #49861, #50774

Issue #51244, opened 2026-07-10
[Bug]: ScoreBasedBalancer never converges — 80-89% of balance Moves net to zero (segment ping-pong), each release cancels in-flight lazy-load searches (LoadCellBatchAsync cancelled)


rajshivage:
## Environment

- **Milvus Version**: v2.6.18 (behavior identical on v2.6.17; search-cancellation impact absent on v2.6.7 because vector data was fully memory-resident there)
- **Deployment Mode**: cluster (Kubernetes/EKS; 25 QueryNode pods, 9 StreamingNodes, 13 Proxies, 1 MixCoord)
- **MQ**: woodpecker
- **SDK**: pymilvus 2.x (bug is server-internal; reproduces regardless of SDK)
- **OS**: Amazon Linux 2023 nodes (EKS)

### Option B: Steps (systemic / stability issue)

1. **Deploy**: cluster with 25 QueryNodes, StorageV2 storage format, tiered storage with default `warmup.vectorField=disabled`, default QueryCoord balancer (ScoreBasedBalancer, zero non-default queryCoord settings).
2. **Setup**: one collection, 57.7M rows, ~449 GB, 706 sealed L1 segments, HNSW/COSINE vector index, 16 vchannels, loaded with `replica_number=2`. Key precondition: sealed-segment sizes are heterogeneous across nodes — per-node average varies 39,850–194,603 rows/segment (5×), likely accumulated via `sealProportion: 0.25` + compaction history.
3. **Workload**: steady search traffic 13–28 QPS with `output_fields=["*"]` (includes the vector field → every search performs on-demand StorageV2 cell loads via `LoadCellBatchAsync`); light inserts (~55 req/min).
4. **Trigger**: none required — steady state. `balance_checker` fires `reason=segment unbalanced` Move tasks at ~440/hour indefinitely.
5. **Observed via**: coordinator logs (`task/scheduler.go:532` "task added" Move lines), QueryNode logs (`LoadCellBatchAsync cancelled for segment -1`), proxy error logs (`context canceled`, clientRole=querynode), Prometheus (`milvus_querynode_segment_num`, `milvus_querynode_entity_num`, `milvus_querynode_entity_size`, `milvus_querycoord_task_latency_count{task_type="segment_move"}`).

### Trigger Conditions

- **Frequency**: always — continuous for 14+ days; Move rate flat ~440/hr over any 6h window (496/434/411/499/435/406 per hour measured), no decay.
[...]

PR #50774, opened 2026-06-25
enhance: [cp2.6] make balance workload delta dist aware

sijie-ni-0214:
Cherry-pick from master
pr: #49861

issue: #49860

## Test Plan

- `gofumpt -l $(git diff --name-only upstream/2.6..HEAD -- '*.go')`
- `git diff --check upstream/2.6..HEAD`
- `rm -rf cmake_build/thirdparty/milvus-proto && make generated-proto-without-cpp`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/querycoordv2/assign -run "Test(ScoreBasedAssignPolicy|RowCountBasedAssignPolicy|RoundRobinAssignPolicy)" -timeout 300s`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/querycoordv2/task -run "TestTask/(TestCalculateTaskDelta|TestSegmentTaskDeltaWithDistFilter|TestChannelTaskDeltaCache)" -timeout 300s`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/querycoordv2/balance ./internal/querycoordv2/checkers -timeout 300s`
- `go test -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/querycoordv2 -run "Test(OpsService|Service)/(TestTransferSegment|TestTransferChannel|TestLoadBalance|TestLoadBalanceWithNoDstNode|TestLoadBalanceWithEmptySegmentList)$" -timeout 600s`

Local build note:

- `make build-go` could not complete on this macOS worktree because the local C++ output is stale and misses `SetStreamBudgetRatio`.
- Refreshing the C++ output with `CONAN_CMD=conan-1 make` is blocked by the local Homebrew clang/SDK environment during Conan's CMake compiler check: `ld: library 'System' not found`.

PR #49861, opened 2026-05-17
enhance: make balance workload delta dist aware

sijie-ni-0214:
#### What this PR does / why we need it:

Improve QueryCoord balance stability by making scheduler segment task deltas aware of current segment distribution.

When pending task deltas are not aligned with actual segment distribution, balance policies can repeatedly shift plans based on stale workload information. This PR tracks segment task delta effects with action records and filters effects that have already been reflected in distribution.

This also caches collection row count and delegator count in score workload status to avoid repeated distribution scans, and applies the updated segment delta snapshot to score and row-count balance paths. Round-robin segment assignment keeps the lightweight segment-count only ordering.

#### Which issue(s) this PR fixes:

Fixes #49860

#### Special notes for your reviewer:

MultiTargetBalancer is intentionally unchanged because it does not consume scheduler segment task deltas in the original design.

#### Tests

- `gofumpt -l internal/querycoordv2/task/scheduler.go internal/querycoordv2/assign/assign_policy_score.go internal/querycoordv2/assign/assign_policy_rowcount.go internal/querycoordv2/assign/assign_policy_roundrobin.go`
- `git diff --check`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/querycoordv2/assign -run "Test(ScoreBasedAssignPolicy|RowCountBasedAssignPolicy|RoundRobinAssignPolicy)" -timeout 300s`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${MILVUS_WORK_DIR}/cmake_build/lib -r ${MILVUS_WORK_DIR}/internal/core/output/lib" ./internal/querycoordv2/task -run "TestTask/(TestCalculateTaskDelta|TestSegmentTaskDeltaWithDistFilter|TestChannelTaskDeltaCache)" -timeout 300s`

PR #51431, opened 2026-07-15
enhance: add resource-group balance epochs

xiaofan-luan:
## What changed

Add an optional Resource-Group-scoped balance epoch for QueryCoord normal balancing.

- capture one immutable placement snapshot per RG generation
- plan ScoreBased/ChannelLevel waves from snapshot-only state
- enforce hard kind, node, collection, and object budgets
- admit tasks through typed generation-aware scheduler gates
- reconcile completion/failure/ambiguity from authoritative distribution
- keep ambiguous RPC objects and node reservations locked until placement settles
- integrate active, shadow, drain, and legacy fallback behavior in `BalanceChecker`
- add rollout configuration, metrics, retained retry/quarantine observation, and #51244 convergence fixtures

The control loop is:

`Observe -> Plan -> Admit -> Execute -> Reconcile -> Observe again`

Each RG has at most one current normal-balance generation. Different RGs remain independent.

## Why

Issue #51244 reports repeated zero-net segment moves and A→B→A ping-pong over long periods. Existing pending-task deltas reduce stale planning, but they do not provide a consistent generation boundary, one shared RG projection, hard wave admission, or authoritative reconciliation before replanning.

The retained static-target fixture converges across four generations:

`68550 -> 48750 -> 16550 -> 5550`

with admitted counts `2, 2, 1, 0`, no reverse move, no concurrent duplicate object, and no hidden zero-net suffix cycle.

## Safety and rollout

- active mode defaults to disabled
- shadow mode plans/evaluates without admission or runtime state
- disabling active mode drains existing epochs before legacy fallback
- unsupported balancers remain on the legacy path
- stopping and recovery balance retain higher priority and legacy admission in this MVP
- Grow-before-Reduce and source retention after failed Grow remain unchanged
- scheduler/RPC success never substitutes for authoritative placement

This MVP intentionally does not add memory/disk byte vectors, migration-byte cost, move cooldown, persistent epochs, or recovery/stopping epoch classes.

## Ambiguous RPC semantics

- failure before QueryNode callback invocation: definitive `RPCNotSentError`
- raw post-dispatch transport/gRPC error: `AmbiguousExecutionError`
- explicit non-OK QueryNode protobuf status: definitive
- `Task.Err()` is read only after `Done()` closes
- desired target-only placement always wins
- ambiguous object/source/target reservations persist until authoritative placement settles

## Validation

Final implementation revision: `8c7b55f32cdf031ead70e99afdace281434761bb`

- 9/9 focused selectors passed
- focused race selector passed with no race report
- complete `internal/querycoordv2/...` suite passed
- 22/22 commits passed DCO audit
- final whole-branch re-review: 0 Critical, 0 Important, 0 Minor
- branch-added raw constructors: 0
- scoped ruleguard: `rawmerrerror: 0`
- isolated mini process, ports 2379/2380, and temporary data cleanup all passed

Repository-wide `make static-check` is not claimed green. It stops in core-package typechecking, before ruleguard, because the checkout lacks the unchanged generated package `cmd/tools/migration/legacy/legacypb`. The scoped balance-package verifier reached the changed code and reported zero `rawmerrerror` findings.

## Links

issue: #51244

design: #51430

related: #49861, #50774


czs007:
Adversarial review surfaced five issues in the query-coordinator balance epoch code — one medium-severity correctness bug that can permanently pin a resource group in a degraded state, plus two low-severity and two nitpick-level robustness/consistency concerns.

## Medium

**`internal/querycoordv2/balance/epoch_manager.go:1477` — `classifyTerminalWork` never definitively resolves reverted terminal tasks**
`classifyTerminalWork` maps every `Canceled` task and every non-desired `Succeeded` task to `terminalWorkAmbiguous`. Ambiguous work is re-added to `nextCarry`, and the only write to the per-resource-group, epoch-persistent `runtime.carryOver` is `runtime.carryOver = nextCarry`. A preempted balance task that reverted to its source (`sourcePresent=true`, `targetPresent=false`, `done=true`) re-observes identically every cycle, so it never leaves `carryOver`: it pins the RG in `EpochDegraded` indefinitely and re-charges a `ReservationAmbiguousCapacity` on `From`/`To` on every plan, permanently reserving node capacity and blocking convergence. Only reachable when `balanceEpoch.enabled=true` (default `false`).
*Suggestion:* when `done && status ∈ {Canceled, Succeeded} && sourcePresent && !targetPresent`, treat it as a definitive revert (drop the carry, release the reservation); keep `terminalWorkAmbiguous` only when `!done` or `targetPresent` (genuine split placement).
*(raised by tinswzy)*

## Low

**`internal/querycoordv2/balance/epoch_manager.go:791` — terminal-state precedence overrides `Superseded` and mislabels `TimedOut` failures**
In `reconcileLocked`, `if hadLostPlacement { state = EpochDegraded }` runs before the terminal-intent branch, so an `EpochSuperseded` intent is overwritten by `EpochDegraded` whenever any object is lost. Separately, a `TimedOut` epoch carrying a definitive Grow/Reduce failure (`hadFailure=true`, nothing lost, empty `nextCarry`) skips the completion downgrade (guarded by `!hadFailure`) and finalizes as `timed_out` despite a definitive failure. Both are observability/result-label nuances — the states feed only `terminalEpochResultLabel` and `result.State`, with no distinct control-flow consumer — rather than a hard correctness bug.
*Suggestion:* let a `Superseded` terminal intent take precedence over the lost-placement downgrade, and distinguish the `TimedOut`-with-failure case so the emitted metric label reflects the actual outcome.
*(raised by tinswzy)*

**`internal/querycoordv2/task/scheduler.go:1077` — dead retry loop plus a wide RLock across two deep-copy scans**
`GetPendingBalanceTasks` samples `pendingRevision` before and after within a single continuous `pendingMu.RLock` section. The sole writer, `incrementPendingRevisionsLocked`, only runs under the exclusive `pendingMu.Lock`, so no writer can interleave; `before == after` always holds and the `if before != after { continue }` retry is unreachable dead code. The snapshot is internally consistent (not a correctness bug), but the RLock is held across both the `segmentTasks` and `channelTasks` `Range` scans, each deep-copying every task — O(all-tasks) work that blocks all pending add/admit/remove writers on the balance hot path.
*Suggestion:* drop the dead retry loop, and narrow the critical section — e.g. collect task references under the lock and perform the deep copies outside it — so writers aren't blocked for the full copy duration.
*(raised by tinswzy)*

## Nitpick

**`internal/querycoordv2/balance/epoch_snapshot.go:121` — `Validate` uses `context.Background()` instead of threading a `ctx`**
`Validate(token)` takes no `ctx` and calls `GetResourceGroup(context.Background())` and `captureToken(context.Background(), ...)`, unlike `Build()` which threads a real `ctx`. There is no behavioral impact today: `GetResourceGroup` ignores its `ctx` (pure in-memory `RLock` read), and `captureToken`/`buildFromCapture` only forward it to further in-memory reads that likewise ignore it — no blocking or cancellable work exists. This is a pure consistency nit, not a lost-cancellation bug.
*Suggestion:* have `Validate` accept and thread a `ctx` for uniformity with `Build`.
*(raised by chyezh, tinswzy)*

**`internal/querycoordv2/balance/epoch_snapshot.go:482` — non-deterministic tie-break for anonymous pending tasks in `mergePendingTasks`**
`mergePendingTasks` sorts by `(TaskID, CollectionID, ReplicaID)` with a non-stable `sort.Slice`. Two anonymous tasks (`TaskID==0`, same `CollectionID`, `NilReplica`) that differ only in `Actions` are distinct map entries (the anonymous merge key includes `Actions`) yet tie completely in the comparator, yielding non-deterministic ordering. Harmless today — the only consumer, `projectPendingWork`, sums into node-keyed maps (order-independent), and `SnapshotToken` hashes revisions rather than the ordered slice — but fragile.
*Suggestion:* add a final `Actions`-based tie-break to keep ordering deterministic in case a future planner iterates in order.
*(raised by tinswzy)*

PR #51097, opened 2026-07-07
enhance: add per-collection deltalog observability metrics

czs007:
issue: #51094

### What

Today no metric can answer "how close are segments to the delete-driven single-compaction thresholds":

- `milvus_datacoord_segment_binlog_file_count` sums binlogs + statslogs + **deltalogs** into one collection-wide number, so deltalog growth is drowned out by per-field data binlogs, and there is no per-segment view.
- `l0_delete_entries_num` / `store_level0_segment_size` cover L0 delete **rows/size** but not **file counts**.

In the incident behind #51094, ~1,400 same-batch segments accumulated deltalog files at nearly the same rate and crossed `single.deltalog.maxnum` together — nothing on any dashboard hinted at it until the rewrite avalanche started. (This PR is the observability companion of #51096, which adds the smoothing/rate-limiting itself.)

### How

Per-collection summary-style quantile gauges over healthy non-L0 segments, computed exactly in the existing `GetQuotaInfo` refresh loop (no new iteration; same pass that already feeds `segment_binlog_file_count`):

| Metric | Labels | Compare against |
|---|---|---|
| `segment_deltalog_file_count` | `quantile` = 0.5 / 0.9 / 0.99 / 1.0 | `dataCoord.compaction.single.deltalog.maxnum` (default 200) |
| `segment_deleted_rows_ratio` | `quantile` = 0.5 / 0.9 / 0.99 / 1.0 | `single.ratio.threshold` (default 0.2) |
| `l0_deltalog_file_num` | — | total L0 deltalog files (delete-stream spray; complements existing L0 rows/size metrics) |

Quantiles instead of a max-only gauge because the max cannot distinguish one lone dirty segment from a synchronized cohort: in the incident shape, p50 marching toward the threshold **is** the avalanche warning (half the segments are at ~180 files), weeks before anything trips. The 1.0 quantile still tracks the dirtiest segment for threshold alerts. Quantiles are recomputed exactly (nearest-rank over the actual per-segment values) on each refresh round, so no decaying-window summary state is needed.

Cardinality: `collections × 9` series, same order as the existing `stored_binlog_size` (collections × states).

### Metric lifecycle

Follows the existing double cleanup pattern:
- full `Reset()` + republish on every refresh round (dropped collections disappear on the next tick, gauges rebuild cleanly after DataCoord failover);
- eager `DeletePartialMatch` in `CleanupDataCoordWithCollectionID` on `DropCollection`, covering the case where the quota-driven refresh is not running.

### Test plan

[...]

PR #51096, opened 2026-07-06
enhance: de-synchronize and rate-limit delete-triggered single compaction

czs007:
issue: #51094

### What

Segments created in the same batch (bulk import, DR initial sync, restore) accumulate deltalogs at nearly the same rate, so they cross the single-compaction hard thresholds (`single.deltalog.maxnum` etc.) almost simultaneously. The result is a synchronized rewrite avalanche: in a production incident, ~1,400 segments became eligible within ~36 hours and the cluster rewrote multiple TB against a rate-limited object store, degrading queries for days. The same shock shape is produced by a threshold config change taking effect, a TTL cliff, or flipping `autoUpgradeSegmentIndex`.

This PR de-synchronizes and bounds delete/expiry-triggered single compaction with two independent mechanisms (each individually disableable, both refreshable):

1. **Deterministic per-segment threshold jitter** (`dataCoord.compaction.single.thresholdJitter`, default 0.25). Each segment gets a stable multiplier in `[1, 1+J]` (splitmix64 hash of segment ID — no persisted state, consistent across restarts/nodes) applied to the *accumulation* thresholds: deltalog count, deltalog size, deleted-rows ratio, and expired-entities ratio/size. A cohort's crossing times spread over `J ×` (accumulation period) — for the incident shape, ~12 days instead of ~36 hours.
2. **Token-bucket admission at plan generation** (`dataCoord.compaction.single.rateLimitTokens` / `rateLimitInterval`, default 256 per 60s). Jitter cannot help when eligibility jumps in one step (config change, TTL cliff); the bucket converts any mass eligibility into a paced stream. Deferred segments carry no state and are simply re-evaluated on later trigger rounds. Admission is dirtiest-first (highest deleted-rows ratio), the same rate-limit parameter shape as the existing `dataCoord.compaction.storageVersion.rateLimitTokens/rateLimitInterval`.

Guardrails:
- **Bounded deferral**: a segment whose deltalog count reaches 4× `single.deltalog.maxnum` is always admitted without consuming tokens, so the limiter can never starve a pathological segment.
- **Misconfiguration warning**: sustained throttling (30+ consecutive throttled rounds) emits a rate-limited WARN — an undersized bucket means backlog is growing and the operator must know.
- `signal.isForce` (manual compaction) bypasses admission entirely — explicit operator intent.
[...]

Issue #51094, opened 2026-07-06
[Enhancement]: Trigger-side rate limiting for delete-triggered compaction (avoid synchronized rewrite avalanches)


czs007:
### Is your feature request related to a problem? Please describe.

Trigger-side rate limiting exists for two of the three compaction trigger families, but not for the one most prone to synchronized bursts:

| Trigger | Trigger-side pacing |
|---|---|
| storage-version conversion | ✅ token bucket: `dataCoord.compaction.storageVersion.rateLimitTokens` / `rateLimitInterval` |
| sort compaction | ✅ per-round cap: `dataCoord.sortCompactionTriggerCount` |
| **delete-triggered rewrites** (`hasTooManyDeletions`: deltalog count / size / deleted-rows ratio) | ❌ none — every eligible segment is planned in the same trigger round |

When a large number of segments become eligible at the same time, all of them are submitted at once. Downstream slot-based execution only drains the queue; it does not prevent the avalanche.

**Production case (2.6.17-based build), fully log-evidenced:**

- A collection had ~1,400 segments (0.5–1 GB each) created together on one day by a bulk initial sync.
- A steady delete stream, applied via frequent small L0 compactions, appended deltalog files to these same-aged segments at a near-uniform rate.
- ~48 days later, the whole cohort crossed `dataCoord.compaction.single.deltalog.maxnum` (200) within ~36 hours. Trigger census in a sampled 17-minute DataCoord window: **128× "delta logs file count exceeds threshold"** (caught at `delta log count=202`, i.e. freshly crossed) vs 4× ratio, 0× size, 0× TTL; all 128 victims decode (TSO of segmentID) to the same birth date.
- Consequences of the resulting ~1,400 rewrites in waves:
  - object storage: each rewrite costs ~3–4× the segment size (read + write + index-rebuild read + query-node reload read); the store's **daily GET-bytes quota was exhausted**;
  - query availability: each completed wave swapped the delegator query view (`SyncTargetVersion` resets loadedRatio to 0), producing minutes-long unserviceable windows; requests requiring full data (Query / complex Delete, hard-coded ratio 1.0) failed in bursts;
  - other compaction types were starved during the waves (repeated `Failed to execute compaction task` for Sort submissions).
- Most of these rewrites were also *barely dirty*: the file-count trigger fired long before the 20% deleted-rows ratio, so ~0.5–1 GB segments were fully rewritten to purge a small number of deleted rows.

Adjacent mechanisms do not cover this:
[...]

PR #51432, opened 2026-07-15
enhance: [poc_2.6.17] add per-collection deltalog observability metrics

czs007:
Cherry-pick of #51097 onto `poc_2.6.17` (squashed; upstream commits 691db63 + 0a54c39 + 36a917d, both adversarial-review rounds folded in).

### What

Per-collection summary-quantile gauges over healthy **flushed non-L0** segments, split by `segment_level`, computed in the existing `GetQuotaInfo` refresh loop (no new iteration):

| Metric | Labels | Compare against |
|---|---|---|
| `segment_deltalog_file_count` | `segment_level`, `quantile`=0.5/0.9/0.99/1.0 | `single.deltalog.maxnum` (200) |
| `segment_deltalog_size` | `segment_level`, `quantile` | `single.deltalog.maxsize` |
| `segment_deleted_rows_ratio` | `segment_level`, `quantile` | `single.ratio.threshold` (0.2) |
| `l0_deltalog_file_num` | — | total L0 deltalog files |

Surfaces the delete-driven single-compaction **cohort signal** behind #51094 (p50 marching toward the threshold *is* the avalanche warning) before the rewrite avalanche starts. Observability companion of #51096.

### Backport adaptation

- `pkg/v3` → `pkg/v2` and `go-api/v3` → `go-api/v2` import paths.
- `meta.go`: kept the `poc_2.6.17`-only "not found database name" else-branch alongside the new `aggregate.observe()` call.

### Test

- Build passes; `TestDeltalogQuantile`, `TestDeltalogAggregateObserve`, `TestReportDeltalogMetrics` all pass locally on `poc_2.6.17` (aa866d0c22).

issue: #51094
pr: #51097

### DIFF
--- a/internal/datacoord/metrics_deltalog.go
+++ b/internal/datacoord/metrics_deltalog.go
@@ -1,0 +1,182 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package datacoord
+
+import (
+	"math"
+	"sort"
+	"sync"
+
+	"github.com/milvus-io/milvus/pkg/v2/metrics"
+	"github.com/milvus-io/milvus/pkg/v2/proto/datapb"
+)
+
+// deltalogQuantiles are the summary quantiles published for the per-segment
+// deltalog distributions. The 1.0 quantile (max) tracks the dirtiest segment
+// against the single-compaction thresholds; the lower quantiles rising toward
+// a threshold reveal a same-batch cohort accumulating toward a synchronized
+// compaction wave, which the max alone cannot distinguish from a lone dirty
+// segment.
+var deltalogQuantiles = []struct {
+	label string
+	q     float64
+}{
+	{"0.5", 0.5},
+	{"0.9", 0.9},
+	{"0.99", 0.99},
+	{"1.0", 1.0},
+}
+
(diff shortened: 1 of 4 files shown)


internal/datacoord/metrics_deltalog.go
```diff
@@ -0,0 +1,182 @@
...
+// quantilesInPlace sorts values once and returns the exact nearest-rank
+// (rank = ceil(q*n)) value for each requested quantile; all-zeros for an empty
+// slice. Sorting once (vs once per quantile) shortens the read-lock hold in
+// reportDeltalogMetrics.
+func quantilesInPlace(values []float64, qs []float64) []float64 {
+	out := make([]float64, len(qs))
+	if len(values) == 0 {
+		return out
+	}
+	sort.Float64s(values)
```
czs007, line 123: nit (non-blocking): `sort.Float64s` here runs inside `GetQuotaInfo`'s `segMu.RLock` (via the `reportDeltalogMetrics` call in `meta.go`), so the sort cost is paid while holding the segment read lock. The comment above already frames sorting-once as a lock-hold optimization, so this is a deliberate trade-off and fine at current scale. If a collection ever holds a very large number of segments, consider snapshotting the aggregated slices out of the lock and sorting/publishing after RUnlock. Not a blocker.

pkg/metrics/datacoord_metrics.go
```diff
@@ -86,6 +87,76 @@ var (
...
+	// manual filters to L1), so the distributions are kept per level rather than
+	// blended. It approximates — not exactly equals — the candidate population:
+	// transient/stable candidacy predicates (isCompacting, IsSorted, IsInvisible,
+	// compaction-protected) are not applied here.
+	// Compare against dataCoord.compaction.single.deltalog.maxnum: the 1.0
+	// quantile shows how close the dirtiest segment is to triggering a
+	// delete-driven compaction, while p50/p90 rising toward the threshold
+	// reveals a same-batch cohort building up toward a synchronized
+	// compaction wave (a lone dirty segment leaves them unmoved).
+	DataCoordSegmentDeltalogFileCount = prometheus.NewGaugeVec(
```
czs007, line 115: (non-blocking) Cardinality note: each collection can add up to ~25 new series here — `segment_deltalog_file_count` / `_size` / `_deleted_rows_ratio` are each `segment_level` (realistically <=2: L1/L2) x `quantile` (4) = 8, so 24 across the three, plus 1 `l0_deltalog_file_num`. The per-round prune covers dropped collections so there is no leak, but on deployments with many collections this multiplies — worth keeping in mind when wiring these into Prometheus.

PR #51336, opened 2026-07-14
fix: release segments from the delegator's reported serving set, not from target-version arithmetic

czs007:
## Summary

`SegmentChecker` decides whether a redundant sealed segment may be released by **inferring**, from target versions, whether the shard delegator still serves it:

```go
readableVersionNotUpdate := view.TargetVersion != initialTargetVersion &&
                            view.TargetVersion < currentTargetVersion
```

That inference is sound only while the delegator's readable view equals the current target. It breaks whenever a delegator is left on a target version coord no longer holds a snapshot of — the strict straddle `current < view < next`, which forms when the next target is replaced while a delegator is still synced to the previous one. The segment is then absent from **both** current and next, yet still inside the delegator's readable set: the checker releases it, `loadedRatio` drops below `1.0`, the shard becomes unserviceable, and cross-shard queries and non-PK-expression deletes fail with a retriable `ChannelNotAvailable` (`code=106`). The shard staying unserviceable keeps the current target frozen, so the state **sustains itself** instead of clearing.

**The delegator already knows the answer, per segment.** Its readable set is `queryView.sealedSegmentRowCount`, so `loaded sealed segments − readable set` is exactly the set it no longer serves (`distribution.go`, `SyncTargetVersion`). That fact is simply never reported to QueryCoord — which is why coord has to guess.

This PR reports it and releases from it.

issue: #51260

## What changes

**QueryNode** — the delegator reports the sealed segments it is loaded with but no longer serving (`LeaderView.not_serving_segmentIDs`), plus a capability bit (`LeaderViewStatus.reports_serving_set`). One set difference, no new bookkeeping.

**QueryCoord** — `filterOutSegmentInUse` releases only what the delegator says it no longer serves. Two properties:

* **Soundness of the fallback.** When the delegator does not report (older QueryNode, or a view coord has not synced yet), coord falls back to the target versions — but releases only when the readable set is *provably* free of the segment:

  | delegator view | readable set | release candidate (∉current ∧ ∉next) in it? |
  |---|---|---|
  | `view == current` | current target's segments | no, by construction → **release** |
  | `view == next` | next target's segments | no, by construction → **release** |
  | `view == initial` | nothing synced yet | → **release** (unchanged) |
  | anything else | a target snapshot coord **no longer holds** | **cannot be proven** → **keep** |

  Where the next-target pin from #50686 holds, `view` is always `current` or `next`, so this fallback **never withholds anything and behaviour is unchanged**. It only engages once that invariant is broken.

[...]

Issue #51260, opened 2026-07-11
[Bug]: QueryCoord may release a sealed segment a shard delegator is still serving, making the shard unserviceable


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (the same guard is present on the release-2.5 / release-2.6 branches)
- Deployment mode(standalone or cluster): both
- MQ type(rocksmq, pulsar or kafka): any
- SDK version(e.g. pymilvus v2.0.0rc2): any
- OS(Ubuntu or CentOS): any
- CPU/Memory: any
- GPU: any
- Others:
```

### Current Behavior

During heavy compaction churn combined with throttled segment loading (a low `queryCoord.queryNodeTaskParallelismFactor`), QueryCoord's `SegmentChecker` can release a sealed segment that a shard delegator is **still serving**. Losing that segment drops the shard's `loadedRatio` below `1.0`, and because a shard only serves reads when it is fully loaded (the readiness gate is a hard `loadedRatio == 1.0`), the whole shard becomes unserviceable and returns `ChannelNotAvailable`. This fails cross-shard queries and complex (non-PK expression) deletes — which run an internal cross-shard retrieve before deleting — until the delegator finishes loading the new segments and advances its readable view.

The root cause is in `filterOutSegmentInUse` (`internal/querycoordv2/checkers/segment_checker.go`). It treats a sealed segment as "still in use" **only** when the delegator's readable view *lags* the current target version:

```go
readableVersionNotUpdate := delegator.View.TargetVersion != initialTargetVersion && delegator.View.TargetVersion < currentTargetVersion
```

A delegator serves from a snapshot of the target that is normally **at or ahead of** current (delegators are synced to next-target versions before current promotes). During a compaction burst a short-lived compacted segment can be loaded and served by the delegator while a later compaction advances the plan past it, so it ends up in **neither** the current nor the next target. `getSealedSegmentDiff` then marks it for release, and because the delegator's view is not lagging, the in-use guard does not fire and the still-served segment is released.

### Expected Behavior

A sealed segment that is still present in a delegator's actually-served set must not be released, regardless of the target-version comparison direction.

### Steps To Reproduce

```markdown
1. Load a collection and keep it serving reads.
2. Drive heavy upsert with small segments so compaction runs frequently and produces short-lived segments.
[...]

PR #51433, opened 2026-07-15
fix: [poc_2.6.17] release segments from the delegator's reported serving set

czs007:
## Summary

Backport of #51336 to `poc_2.6.17`.

`SegmentChecker` decides whether a redundant sealed segment may be released by **inferring**, from target versions, whether the shard delegator still serves it. That inference is sound only while the delegator's readable view equals the current target. It breaks in the strict straddle `current < view < next`: the segment is absent from **both** current and next yet still inside the delegator's readable set, so the checker releases it, `loadedRatio` drops below `1.0`, the shard becomes unserviceable, and cross-shard queries and non-PK-expression deletes fail with a retriable `ChannelNotAvailable` (`code=106`).

**The delegator already knows the answer, per segment.** Its readable set is `queryView.sealedSegmentRowCount`, so `loaded sealed − readable` is exactly the set it no longer serves. It now reports that (`LeaderView.not_serving_segmentIDs`) plus a capability bit (`reports_serving_set`), and the checker releases only what the delegator says it no longer serves.

pr: #51336
issue: #51260

## What changes

**QueryNode** — the delegator reports the sealed segments it is loaded with but no longer serving, plus a capability bit.

**QueryCoord** — `filterOutSegmentInUse` releases only what the delegator says it no longer serves, with three properties:

* **Sound fallback.** When the delegator does not report (older QueryNode, or a view coord has not synced yet), coord falls back to the target versions but releases only when the readable set is *provably* free of the segment (`view` is initial / current / next); an orphan view is kept.
* **Known-loaded requirement.** `not_serving` is a negative list, so a candidate not loaded on this leader at all is also absent from it; a segment is in use only when it is **known loaded here** (present in `view.Segments`) **and** not on the not-serving list — otherwise its reclamation would be blocked indefinitely (review by @yhmo on the master PR).
* **Holder-side veto.** As a backstop the delegator refuses to release a segment still in its readable set; `force` releases (collection drop, node cleanup) bypass it.

## Test Plan

- [x] QueryCoord unit: `TestServingSetReleasesWhatDelegatorDropped`, `TestServingSetProtectsSegmentStillServed`, `TestServingSetReleasesCandidateNotLoadedHere` (the known-loaded case), `TestServingSetFallsBackToTargetVersionsOnOldQueryNode`, `TestServingSetNotTrustedBeforeFirstSync`.
- [x] Existing `checkers` package green; the two failing `delegator` tests (`TestGetHighlight`, `TestRunAnalyzer`) are pre-existing on the `poc_2.6.17` base and unrelated to this change.
- [x] **End-to-end on a 2.6 cluster** (4 shards, small segments, segment loading throttled to one task at a time, non-PK-expression delete probes), same binary and workload, 600s per arm:

  | | buggy guard | **this change** |
  |---|---|---|
  | delete failures | **22** | **0** |
  | query failures | **20** | **0** |
  | `code=106` | **42** | **0** |
  | delegator unserviceable events | **1663** | **1** (startup) |
  | in-use segments wrongly released | **3612** | **0** |
  | segments reclaimed | 761 | **719** (reclamation continues) |


tinswzy:
Adversarial review confirmed several issues. Two (critical build-break, low `Clone()` gap) are kept as-is, and the medium node-blind lookup finding needs rewriting per its verdict. Here's the rendered comment:

---

This review surfaced one build-breaking mock regeneration miss, one node-awareness gap in the coord-side redundancy check, and two test/latent-correctness gaps worth tightening before merge.

## 🔴 Critical

- **`internal/querynodev2/delegator/delegator.go:76` — `MockShardDelegator` not regenerated for the extended `ShardDelegator` interface.** This PR adds `NotServingSealedSegments()` and `ServedFromNode(...)` to the `ShardDelegator` interface, but `internal/querynodev2/delegator/mock_delegator.go` is not updated. `*MockShardDelegator` is consumed as a `ShardDelegator` across the querynode test packages (e.g. `pipeline/filter_node_test.go` wires a `*MockShardDelegator` into `newFilterNode`, whose field is typed `delegator.ShardDelegator`), so it no longer satisfies the interface and the pipeline/handlers/services test packages fail to compile. Regenerate the mock (`make generate-mockery-querynode`) before merge. _(raised by yezh, wzy)_

## 🟡 Medium

- **`internal/querycoordv2/checkers/segment_checker.go:54` — `knownLoaded` lookup is node-blind.** `_, knownLoaded := view.Segments[s.GetID()]` reads the one-entry-per-segment-ID map, so it answers "loaded somewhere in this delegator's view," not "loaded on `s.Node`." The sibling helper at `segment_checker.go:440-441` gates on `segInView.NodeID == s.Node`; this path does not, so a wrong-node duplicate can be marked in-use and retained. The impact is bounded — the holder-side veto in `ReleaseSegments` (`ServedFromNode` with `targetNodeID`) is node-aware and provides the real safety net, so the observable effect is delayed cleanup of a redundant copy rather than a permanent leak. For consistency, tighten this coord-side check to compare `NodeID` against `s.Node` and/or add a node-specific test that a wrong-node copy is not treated as loaded. _(raised by yezh, wzy)_

## 🟢 Low

- **`internal/querycoordv2/checkers/segment_checker_test.go:193` — `nextTargetVersion` release branch is never exercised with a non-zero value.** `setupServingSetCase` calls only `UpdateCollectionCurrentTarget`, so no next target exists and `GetCollectionTargetVersion(..., meta.NextTarget)` returns 0. Since `initialTargetVersion == 0`, `run(newTestLeaderView(nextTargetVersion))` is effectively `run(newTestLeaderView(0))`, which hits the `initialTargetVersion` arm of `provablyNotInReadableSet` rather than the `case nextTargetVersion:` arm this PR threads through. A regression dropping that arm would leave tests green. Publish a real next target so `nextTargetVersion != 0` and assert release at `view == next`. _(raised by wzy)_
- **`internal/querycoordv2/meta/channel_dist_manager.go:49` — `LeaderView.Clone()` drops the two new fields.** This PR adds `ReportsServingSet` and `NotServingSegments` to `LeaderView` but does not update `Clone()` to copy them. It is safe today because the checker read path uses `GetByFilter`, which returns the stored pointers rather than clones, so the fields survive; but this is a latent trap that would silently zero both fields for any future path switching to cloned views. Copy both fields in `Clone()`. _(raised by wzy)_

czs007:
Thanks — all four confirmed against the code and fixed in the latest push (also rebased onto the updated `poc_2.6.17`).

**🔴 Critical — mock not regenerated.** Confirmed: `go vet ./internal/querynodev2/pipeline/...` failed with `*MockShardDelegator does not implement ShardDelegator (missing method NotServingSealedSegments)`. I had only exercised the `checkers` and `delegator` packages, so I missed it — my fault. Root cause was that I also put `ServedFromNode` on the `ShardDelegator` interface, which it needn't be: the veto calls `sd.distribution.ServedFromNode(...)` directly. Removed `ServedFromNode` from the interface (matching master #51336) and added the `NotServingSealedSegments` mock. `go vet` on `pipeline` / `handlers` / `querynodev2` now compiles.

**🟡 Medium — node-blind `knownLoaded`.** Confirmed against the sibling check at `segment_checker.go:418` (`segInView.NodeID == s.Node`). Tightened to `segInView, loadedHere := view.Segments[s.GetID()]; knownLoaded := loadedHere && segInView.GetNodeID() == s.Node`, and added `TestServingSetReleasesWrongNodeCopy` (a copy on node 2 while the candidate is on node 1 must be released) — verified to fail on the node-blind logic.

**🟢 Low — next-target arm exercised with 0.** Confirmed. `setupServingSetCase` now publishes a current *and* a fresh next target, so `nextTargetVersion` is non-zero and distinct; the test asserts `NotZero(nextTargetVersion)` and `NotEqual(current, next)` before exercising the `view == next` release.

**🟢 Low — `Clone()` drops the two fields.** Confirmed. `LeaderView.Clone()` now copies `ReportsServingSet` and `NotServingSegments`.

Note: the Medium and both Low findings apply equally to master #51336 (same code); I'll fold the same three fixes into that PR.

tinswzy:
Adversarial review confirmed 3 issues (all with verdict `rewrite`, none dropped). Here is the PR review comment:

---

This review surfaces one test-coverage gap that leaves the PR's core release path unguarded, plus two low-severity latent robustness concerns worth addressing.

## Medium

- **`internal/querycoordv2/checkers/segment_checker.go:474`** — The known-loaded + not-serving release path (`inUse := knownLoaded && !view.NotServingSegments.Contain(s.GetID())`), which is the behavior this PR exists to add, is never exercised. `TestServingSetReleasesWhatDelegatorDropped` leaves `view.Segments` empty, so its release is driven by `knownLoaded == false`, not by the not-serving list; no test constructs the decisive combination `view.Segments == {seg.ID: seg.Node}` together with `NotServingSegments == {seg.ID}`. As a result, mutating the logic to `inUse := knownLoaded` — which disables the entire delegator-reported release — still passes all six new tests. Add a case with the segment loaded here on `s.Node` and present in `NotServingSegments`, asserting it is released. (raised by wzy)

## Low

- **`internal/querynodev2/delegator/delegator_data.go:942`** — The holder-side veto is batch-granular rather than segment-granular: `stillServing := lo.Filter(req.GetSegmentIDs(), ...)` followed by `if len(stillServing) > 0 { return err }` rejects the entire `ReleaseSegmentsRequest` when any single segment is still in the readable set. A redundant or wrong-node-dedup segment batched into the same request is therefore transiently deferred until its still-serving sibling advances out of the readable set and the checker retries. This is transient and self-healing (the veto keys correctly per-segment on `ServedFromNode`), but consider making the veto segment-granular to avoid deferring unrelated releases. (raised by wzy)

- **`internal/querycoordv2/meta/channel_dist_manager.go:173`** — `DmChannel.Clone()` rebuilds its `View` with only `{ID, CollectionID, Channel, Version, Status}` and copies neither `ReportsServingSet` nor `NotServingSegments`, while `LeaderView.Clone()` copies `NotServingSegments` by reference, sharing the underlying `UniqueSet` map. Both are latent only: `dist_handler.updateChannelsDistribution` builds a fresh `DmChannel` + `LeaderView` every heartbeat and the checker reads the stored pointer without cloning, so no aliasing occurs on the live path. Harmless today but a future footgun — deep-copy the set and carry the new fields through `Clone()` for consistency. (raised by wzy)

czs007:
Thanks — second round confirmed and addressed.

**🟡 Medium — decisive release path untested.** Confirmed: with `view.Segments` empty in `TestServingSetReleasesWhatDelegatorDropped`, release was driven by `knownLoaded == false`, so `inUse := knownLoaded` (dropping the not-serving check) passed all six tests. Added `TestServingSetReleasesReportedNotServing`: segment loaded here on `s.Node` **and** in `NotServingSegments` ⇒ released — verified to fail if the `!NotServingSegments.Contain` check is removed.

**🟢 Low — `DmChannel.Clone()` / deep-copy.** Confirmed. `DmChannel.Clone()` now carries `ReportsServingSet` and `NotServingSegments` into the rebuilt `View`, and both `LeaderView.Clone()` and `DmChannel.Clone()` deep-copy the `NotServingSegments` set (`.Clone()`) instead of sharing the map.

**🟢 Low — batch-granular veto.** The veto rejects the whole `ReleaseSegmentsRequest` when any segment is still served. On the live path this is already segment-granular: `createSegmentReduceTasks` builds one `SegmentAction` per segment, so each `ReleaseSegmentsRequest` carries a single segment ID and the whole-request reject only ever defers that one. It is also self-healing (the checker retries and the veto re-evaluates per segment on the next round). I've left it batch-shaped for that reason; happy to make it filter-and-defer if you'd prefer robustness against a future batched caller.

The same three (test + the two Clone fixes) are folded into master #51336.

tinswzy:

PR #51434, opened 2026-07-16
enhance: [poc_2.6.17] de-synchronize and rate-limit delete-triggered single compaction

czs007:
issue: #51094 (ported from #51096, targeting the poc_2.6.17 production branch where the incident occurred)

### What

Segments created in the same batch (bulk import, DR initial sync, restore) accumulate deltalogs at nearly the same rate, so they cross the single-compaction hard thresholds almost simultaneously. The result is a synchronized rewrite avalanche: in a production incident, ~1,400 segments became eligible within ~36 hours and the cluster rewrote multiple TB against a rate-limited object store, degrading queries for days.

Two independent, individually-disableable, refreshable mechanisms:

1. **Deterministic per-segment threshold jitter** (`dataCoord.compaction.single.thresholdJitter`, default 0.25). Each segment gets a stable multiplier in `[1, 1+J]` (splitmix64 hash of segment ID — no persisted state, consistent across restarts/nodes) applied to every accumulation dimension (deltalog count/size, deleted-rows ratio, expired-entities ratio/size). Applies to **all** single-compaction trigger axes and spreads a cohort's crossing times.
2. **Token-bucket admission at plan generation** (`dataCoord.compaction.single.rateLimitTokens` / `rateLimitInterval`, default 256 per 60s) for **delete-accumulation candidates only**, dirtiest-first, with a jittered bounded-deferral escape.

### Scope of each mechanism (precise guarantees)

- The **token-bucket rate ceiling applies to delete accumulation only.** Its dirtiest-first ordering and its bounded-deferral escape both use a *normalized delete pressure* = `max` over the three delete axes (deltalog count, deleted-rows ratio, delete-log size) of `actual / threshold`, so a segment triggered by any single axis (e.g. large varchar-PK deletes concentrated in a few files) is ordered and bounded on that axis — not only on file count.
- **Expiry accumulation, strict age-based TTL, and index rebuild bypass admission** (they are retention / correctness obligations that must not be paced behind a delete backlog). They are **de-synchronized by jitter only**, not rate-capped. A TTL cliff is therefore smeared by jitter, not converted into a token-bucket-paced stream.
- The bounded-deferral **escape is itself jittered** (`pressure >= 4x` scaled by the segment's jitter multiplier) and does not consume a token, so a deferred count-triggered cohort does not re-synchronize at the `4x` threshold and avalanche through the bypass together. The same jitter means a threshold config *decrease* raises pressure into a per-segment spread of escape points rather than releasing the cohort in one round.
- An axis whose threshold is non-positive is skipped, so a misconfigured (`<= 0`) threshold disables that axis rather than turning the escape into a free bypass.
- `thresholdJitter=0` plus `rateLimitTokens=0` restore exact legacy behavior; force/manual compaction bypasses admission.

### Adversarial-review fixes folded in (three rounds on #51096)

- **Starvation of non-delete reasons** — only delete accumulation is admission-gated; expiry accumulation, strict age-TTL and index rebuild bypass pacing.
- **Bounded deferral on every delete axis** (not just file count) via the normalized-pressure escape; dirtiest-first ordering likewise covers all axes.
- **Jittered escape** so a deferred count-triggered cohort cannot re-synchronize and avalanche through the bypass.
- **Sub-one token budget soft-deadlock** — a positive `rateLimitTokens` below 1 clamps to 1.
- **Non-positive threshold** no longer disables pacing (per-axis skip).
- **Throttle-warning masking** — the sustained-throttle counter is cleared only when the bucket has genuinely refilled.
- **Tokens vs backpressure** — both the L1 trigger and the L2 policy refund tokens for admission-gated segments whose plans are dropped before enqueue.

The single shared bucket is intentional: it bounds *aggregate* rewrite load against the object store (which is source-agnostic), and refills every round; per-caller buckets would multiply the aggregate ceiling by the caller count and defeat the protection. Sustained one-sided saturation is surfaced by the throttle warning.

### Test plan

- New `compaction_admission_test.go`: multiplier determinism / range / zero-jitter disable; disabled-bucket passthrough, dirtiest-first limiting, jittered escape bypass, token refill with injected clock; plus dedicated cases for each review fix (sub-one budget, delete-size-axis reaches the escape, non-positive threshold still paces, refund, throttle-counter survival on a drained bucket, reason classification).
- Existing `Test_compactionTrigger_*`, `TestSingleCompactionPolicySuite`, `TestCompactionTriggerManager` pass; two cases asserting the exact legacy `201 vs 200` deltalog boundary pin `thresholdJitter=0`.

### DIFF
--- a/internal/datacoord/compaction_admission.go
+++ b/internal/datacoord/compaction_admission.go
@@ -1,0 +1,286 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package datacoord
+
+import (
+	"context"
+	"math"
+	"sort"
+	"sync"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/milvus-io/milvus/pkg/v2/log"
+)
+
+// Admission smoothing for single (delete/expiry-accumulation-triggered) compaction.
+//
+// Segments created in the same batch accumulate deltalogs at nearly the same
+// rate, so they cross the hard trigger thresholds nearly simultaneously and
+// produce a rewrite avalanche (see issue #51094). Two mechanisms de-synchronize
+// and bound this:
+//
+//  1. Per-segment threshold jitter: each segment gets a deterministic
+//     multiplier in [1, 1+J] applied to the accumulation thresholds, spreading
+//     a cohort's crossing times over J x (accumulation period).
+//  2. A token bucket at candidate admission: no matter how many delete-triggered
+//     segments become eligible in one trigger round (e.g. after a threshold
+//     config change), at most `rateLimitTokens` are admitted per
+//     `rateLimitInterval`. A segment whose delete pressure has reached
+//     `deferralHardCap` times a trigger threshold (on any delete axis) is always
+//     admitted so deferral stays bounded; that escape is itself jittered so a
+//     deferred cohort does not re-synchronize and avalanche through it at once.
+//
+// Only delete-accumulation candidates are routed through the admitter; both the
+// dirtiest-first ordering and the bounded-deferral escape use a normalized delete
+// pressure across all three delete axes (deltalog count, deleted-rows ratio,
+// delete-log size), so a segment triggered by any single axis is ordered and
+// bounded on that axis. Expiry accumulation, index rebuild and strict age-based
+// TTL are retention / correctness obligations that must not be paced behind a
+// delete backlog, so their callers bypass admission entirely and are only
+// de-synchronized by jitter (see compaction_trigger.go) — the token-bucket rate
+// ceiling applies to delete accumulation only.
+//
+// Both knobs are refreshable; jitter=0 and tokens=0 restore legacy behavior.
+
+// deferralHardCap bounds how far the admission limiter may defer a segment:
+// once its normalized delete pressure reaches deferralHardCap (4x a trigger
+// threshold on some delete axis, scaled by the segment's jitter multiplier) it
+// is admitted regardless of remaining tokens.
+const deferralHardCap = 4.0
+
+// consecutiveThrottledRoundsToWarn controls how many consecutive throttled
+// admission rounds are tolerated before emitting a warning; sustained
+// throttling means the bucket is sized below steady-state demand.
+const consecutiveThrottledRoundsToWarn = 30
+
+// reliefTokenFraction is the post-refill token level (as a fraction of the
+// budget) at which the bucket is considered genuinely recovered. A deferral-free
+// call only resets the sustained-throttle counter when the bucket has refilled
+// to at least this level, so a lightly loaded caller cannot mask a warning while
+// another caller is being throttled every round against a still-drained bucket.
+const reliefTokenFraction = 0.5
+
+// singleCompactionThresholdMultiplier returns the deterministic per-segment
+// jitter multiplier in [1, 1+J]. It is a pure function of the segment ID:
+// stable across restarts and nodes, re-drawn naturally when a compaction
+// produces a segment with a new ID.
+func singleCompactionThresholdMultiplier(segmentID int64) float64 {
+	jitter := Params.DataCoordCfg.SingleCompactionThresholdJitter.GetAsFloat()
+	if jitter <= 0 {
+		return 1.0
+	}
+	// splitmix64 finalizer as a cheap uniform hash.
+	x := uint64(segmentID)
+	x ^= x >> 30
+	x *= 0xbf58476d1ce4e5b9
+	x ^= x >> 27
+	x *= 0x94d049bb133111eb
+	x ^= x >> 31
+	hash01 := float64(x>>11) / float64(uint64(1)<<53)
+	return 1.0 + jitter*hash01
+}
+
+// segmentDeleteCompactionPressure returns a segment's normalized dirtiness across
+// all three delete trigger axes: the max of deltalog-count / maxnum,
+// deleted-rows-ratio / ratioThreshold, and delete-log-size / sizeThreshold. A
+// value of 1.0 means the segment just crossed a 1x trigger threshold on some
+// axis; deferralHardCap (4.0) means it has reached the bounded-deferral escape.
+// Using the normalized max means a segment triggered by any single axis (e.g.
+// large varchar-PK deletes concentrated in a few files, with a low row ratio) is
+// both ordered dirtiest-first and bounded on that axis, not only on the
+// file-count axis. Axes whose threshold is non-positive are skipped, so a
+// misconfigured (<= 0) threshold cannot spuriously drive pressure to the escape.
+func segmentDeleteCompactionPressure(segment *SegmentInfo) float64 {
+	var deltalogCount int
+	var deletedRows, deleteLogSize int64
+	for _, deltaLogs := range segment.GetDeltalogs() {
+		for _, l := range deltaLogs.GetBinlogs() {
+			deletedRows += l.GetEntriesNum()
+			deleteLogSize += l.GetMemorySize()
+		}
+		deltalogCount += len(deltaLogs.GetBinlogs())
+	}
+	pressure := 0.0
+	if maxNum := Params.DataCoordCfg.SingleCompactionDeltalogMaxNum.GetAsFloat(); maxNum > 0 {
+		pressure = math.Max(pressure, float64(deltalogCount)/maxNum)
+	}
+	if rows := segment.GetNumOfRows(); rows > 0 {
+		if ratioThreshold := Params.DataCoordCfg.SingleCompactionRatioThreshold.GetAsFloat(); ratioThreshold > 0 {
+			pressure = math.Max(pressure, (float64(deletedRows)/float64(rows))/ratioThreshold)
+		}
+	}
+	if sizeThreshold := float64(Params.DataCoordCfg.SingleCompactionDeltaLogMaxSize.GetAsInt64()); sizeThreshold > 0 {
+		pressure = math.Max(pressure, float64(deleteLogSize)/sizeThreshold)
+	}
+	return pressure
+}
+
+// singleCompactionAdmitter is a token bucket shared by every single-compaction
+// candidate producer (the legacy trigger and the single compaction policy), so
+// the whole DataCoord observes one admission budget. The shared budget is
+// intentional: the constraint being protected is the aggregate rewrite load
+// against a rate-limited object store, which does not care whether a rewrite
+// originated from the L1 trigger or the L2 policy. Cross-caller fairness within
+// a single round is first-come, but the bucket refills every round, so no caller
+// is starved across rounds.
+type singleCompactionAdmitter struct {
+	mu              sync.Mutex
+	tokens          float64
+	lastRefill      time.Time
+	throttledRounds int
+	nowFn           func() time.Time // injectable for tests
+}
+
+var (
+	globalSingleCompactionAdmitter     *singleCompactionAdmitter
+	globalSingleCompactionAdmitterOnce sync.Once
+)
+
+func getSingleCompactionAdmitter() *singleCompactionAdmitter {
+	globalSingleCompactionAdmitterOnce.Do(func() {
+		globalSingleCompactionAdmitter = &singleCompactionAdmitter{nowFn: time.Now}
+	})
+	return globalSingleCompactionAdmitter
+}
+
+// admit selects which eligible segments may be submitted this round.
+// Ordering: segments past the deferral hard cap are always admitted; the rest
+// are admitted dirtiest-first while tokens last. A non-positive token config
+// disables limiting entirely (legacy behavior). The returned gatedIDs are the
+// admitted segments that consumed a token (i.e. excluding the hard-cap bypass),
+// so the caller can refund tokens for plans it ends up dropping.
+func (a *singleCompactionAdmitter) admit(eligible []*SegmentInfo) (admitted []*SegmentInfo, gatedIDs map[int64]struct{}, deferred int) {
+	if len(eligible) == 0 {
+		return nil, nil, 0
+	}
+	budget := Params.DataCoordCfg.SingleCompactionRateLimitTokens.GetAsFloat()
+	if budget <= 0 {
+		return eligible, nil, 0
+	}
+	// A sub-one positive budget can never let tokens reach 1, which would defer
+	// every non-hard-cap candidate forever. Treat any positive budget as at
+	// least one admission per interval.
+	if budget < 1 {
+		budget = 1
+	}
+	interval := Params.DataCoordCfg.SingleCompactionRateLimitInterval.GetAsDuration(time.Second)
+	if interval <= 0 {
+		return eligible, nil, 0
+	}
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	now := a.nowFn()
+	if a.lastRefill.IsZero() {
+		a.tokens = budget
+	} else {
+		a.tokens += budget * now.Sub(a.lastRefill).Seconds() / interval.Seconds()
+		if a.tokens > budget {
+			a.tokens = budget
+		}
+	}
+	a.lastRefill = now
+	tokensAfterRefill := a.tokens
+
+	// Precompute each segment's normalized delete pressure once; it drives both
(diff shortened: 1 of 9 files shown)


tinswzy:
## Adversarial Review

Verification surfaced one high-severity starvation bug and four medium-severity paced-admission issues; one fabricated finding was dropped.

### High

- **`internal/datacoord/compaction_admission.go:200` — expiry-accumulation segments are throttled with no bounded fallback, so expired-data reclamation can be starved indefinitely by delete backlog.** `admissionGated()` gates `reasonExpiryAccumulation`, but both the ordering metric (`segmentDeletedRowsRatio`) and the only bounded escape hatch (`segmentDeltalogCount(segment) >= hardCapCount`) count *delete* deltalogs only. Pure expiry accumulation lives in insert binlogs and produces no delete deltalogs, so such segments get ratio ≈ 0, sort last, and can never reach the hard-cap fallback. Delete-gated segments have a bounded deferral (admitted once they exceed the cap); expiry-gated segments have none. Under sustained delete pressure (≥ `rateLimitTokens` candidates per round) they are deferred without bound — contradicting the PR's own stated principle that TTL/retention obligations must never be throttled. *Suggestion:* let expiry-accumulation bypass gating like strict-expiry, or give it an age/expiry-based hard cap independent of delete-deltalog count, plus a regression test that a stale expiry candidate is admitted within bounded rounds under persistent delete backlog. _(raised by wzy)_

### Medium

- **`internal/datacoord/compaction_trigger.go:778` — first-match-wins classification lets a never-gated rebuild-index reason be absorbed by a co-occurring gated reason and thus paced.** `singleCompactionReasonFor` returns strict-expiry → expiry-accumulation → too-many-deletions → rebuild-index. A segment that both needs an index rebuild and matches a gated reason hits the gated reason first, so `reasonRebuildIndex` is paced despite the docs promising it is never gated. This is a documented, intentional trade-off (the paced rewrite rebuilds the index as a side effect) and index-version lag is not itself a correctness issue — the harm only materializes under the starvation in the finding above. *Suggestion:* adopting the expiry-accumulation bypass / age hard cap resolves this too; otherwise check `ShouldRebuildSegmentIndex` before returning a gated reason to honor the documented contract. _(raised by yezh, wzy)_

- **`internal/datacoord/compaction_admission.go:131` — the global token bucket is filled first-come per round, so one collection can drain it and no cross-collection fairness exists.** The admitter is a process-wide singleton, but `admit()` is invoked per collection (L1 in `handleSignal`, L2 in `triggerOneCollection`), and the dirtiest-first sort only orders within a single call. A collection with ≥ `rateLimitTokens` delete candidates can exhaust the bucket for the round, forcing other (possibly dirtier) collections to wait for refill. The PR notes this as intentional and no caller is starved across rounds, but under sustained load a single collection can dominate throughput. *Suggestion:* aggregate candidates and sort globally before admitting, or shard the bucket per collection. _(raised by yezh, wzy)_

- **`internal/datacoord/compaction_policy_single.go:252` — the L2 path consumes tokens but never refunds them, leaking tokens and depressing effective throughput under backpressure.** The L1 path captures `gatedIDs` and refunds unused tokens when the inspector is full or enqueue/AllocN fails; the L2 path discards `gatedIDs` (`admitted, _, deferred := ...`) and performs no refund anywhere in the file. Since `admit()` decrements `a.tokens` for every gated segment, any L2 plan dropped during later build/submit permanently occupies its token, throttling L2 (and the shared bucket's L1) below the configured rate. *Suggestion:* capture `gatedIDs` on the L2 path and refund when a plan is not successfully enqueued. _(raised by wzy)_

- **`internal/datacoord/compaction_admission.go:200` — `hardCapCount ≤ 0` (when `SingleCompactionDeltalogMaxNum ≤ 0`) makes every candidate bypass pacing, silently disabling the token bucket.** `currentHardCapCount` computes `int(deferralHardCap * SingleCompactionDeltalogMaxNum.GetAsFloat())` and is monotonic-increasing from an initial 0; a misconfigured value ≤ 0 keeps `hardCapCount` at 0. The fallback `segmentDeltalogCount(segment) >= hardCapCount` is then always true (delta count ≥ 0), so all eligible segments take the "bounded deferral, always admitted, no token consumed" branch — the entire pacing mechanism is bypassed, not merely the fallback. The default (200) is safe; this requires a manual misconfiguration, hence medium. *Suggestion:* clamp `currentHardCapCount` to a lower bound (treat < 1 as disabled/very-large), or require `hardCapCount > 0` before taking the bypass branch. _(raised by wzy)_

czs007:
Thanks for the thorough second pass — addressed in `5231f3f2fe`. Point by point:

**[High] Expiry-accumulation starved with no bounded fallback.** Correct, and it was a real blind spot: both the ordering metric (`segmentDeletedRowsRatio`) and the hard-cap escape (`segmentDeltalogCount >= hardCapCount`) only count *delete* deltalogs, so a pure-expiry segment sorts last and never reaches the cap. Fixed by narrowing the gate: `admissionGated()` now returns true **only for `reasonTooManyDeletions`**. Expiry accumulation joins strict age-TTL and index rebuild as a bypass reason, so it is never throttled behind a delete backlog. Its bounded per-segment jitter still de-synchronizes a cohort. New assertion in `TestSingleCompactionReasonAdmissionGated`.

**[Med] First-match classification pacing a rebuild-index reason.** Resolved by the same change. Delete accumulation is now the only gated reason, and it carries the bounded-deferral hard cap, so a delete-heavy segment that is also index-old is bounded-deferred (its eventual rewrite rebuilds the index). A segment whose *only* reason is index rebuild / expiry / age-TTL bypasses pacing.

**[Med] Global bucket first-come, no cross-collection fairness.** This one is intentional and I've kept it, with the rationale now spelled out in the code: the constraint being protected is *aggregate* rewrite load against a rate-limited object store, which is source-agnostic (L1 vs L2, collection A vs B). The bucket refills every round, so no caller is starved across rounds, and within a round admission is dirtiest-first. Global cross-collection ordering would require buffering every caller's candidates behind a coordinator for a marginal fairness gain over a 60s window; I'd rather not add that coupling here. Open to revisiting as a follow-up if it proves to matter in practice.

**[Med] L2 path consumes tokens but never refunds.** Correct — `SubmitSingleViewToScheduler` returns on alloc/handler/enqueue failure, leaking the token. Fixed: `MixSegmentView` now carries an `admissionGated` flag (set from `gatedIDs` when the L2 view is built), and the submitter refunds one token on any non-success exit via a `submitted`-flag defer. Symmetric with the L1 refund path.

**[Med] `hardCapCount <= 0` disables pacing.** Correct edge case (misconfigured `deltalog.maxnum <= 0`). The bypass branch now requires `hardCapCount > 0`, so a non-positive cap sends everything through the token bucket instead of bypassing it. New `non-positive hard cap does not disable pacing` test.

The fabricated finding being dropped is appreciated. All datacoord compaction suites (`Test_compactionTrigger_*`, `TestSingleCompactionPolicySuite`, `TestCompactionTriggerManager`) plus the new `compaction_admission_test.go` pass locally.

tinswzy:
## Adversarial Review

This PR de-synchronizes and rate-limits single compaction, but the bounded-deferral and de-synchronization guarantees only hold for a subset of the trigger axes, and the PR description overstates the coverage; the issues below range from a correctness gap in the deferral hard cap to documentation accuracy.

### High

**`compaction_admission.go:204` — delete-size / delete-ratio triggered segments still lack a bounded-deferral guarantee**
`hasTooManyDeletions` fires on three independent metrics (deltalog file count, deleted-rows ratio, delete-log byte size), but `admit`'s two safeguards only cover two of those axes: the bounded-deferral hard cap looks solely at `segmentDeltalogCount` (`>= 4×SingleCompactionDeltalogMaxNum`), and the sort key looks solely at `segmentDeletedRowsRatio`. A segment triggered only by delete-log size (e.g. varchar PK, large deletes concentrated in few files with a low row ratio) keeps losing the ordering to dirtier segments (low ratio → sorted last) while never reaching the file-count hard cap. Under sustained token pressure it is deferred every round while its delete-log volume grows unbounded — exactly the read-amplification condition the size trigger is meant to relieve. *Suggestion:* add `totalDeleteLogSize >= 4×SingleCompactionDeltaLogMaxSize` to the hard-cap escape, or change the sort key to a normalized max of size/count/ratio. *(raised by wzy, xiaocai)*

### Medium

**`compaction_trigger.go:785` — expiry-accumulation avalanche is not bounded by the token bucket; PR description overclaims**
`reasonExpiryAccumulation` has `admissionGated() == false`, so it bypasses the token bucket and is only spread by per-segment jitter (default 0.25×), with no 256/60s hard rate ceiling. Yet the PR description states this PR "de-synchronizes and bounds delete/expiry-accumulation-triggered single compaction with two mechanisms," and mechanism ② claims "Any mass eligibility (config change, TTL cliff) becomes a paced stream." A TTL cliff produces expiry-accumulation rewrites, which are only smeared across a ~25% window by jitter — they never become a token-bucket-paced stream, and this contradicts the design note that expiry is "not admission-gated." *Suggestion:* narrow the description to the actual guarantee (delete-accumulation is rate-capped; expiry-accumulation is only jitter-desynchronized), or wire expiry to its own token bucket. *(raised by xiaocai)*

**`compaction_admission.go:204` — hard-cap escape has no jitter; count-triggered cohorts re-synchronize at the 4× threshold and avalanche**
Jitter only desynchronizes the 1× threshold crossing in `hasTooManyDeletions` (`threshold × mult`, `mult ∈ [1, 1.25]`); the hard-cap bypass uses `segmentDeltalogCount >= 4×SingleCompactionDeltalogMaxNum` with no per-segment jitter. A cohort accumulating at a common rate crosses the 1× threshold at spread-out times but crosses the 4× threshold at the same instant. If the bucket defers the cohort from t1 until they all reach 4× (the sustained-saturation scenario this PR targets), the whole batch hits the hard cap simultaneously and passes through the escape valve together — and the escape does not consume tokens — reproducing the synchronized rewrite avalanche via the bypass, against which jitter has no desynchronizing effect. *Suggestion:* apply bounded jitter to the hard-cap multiplier as well (e.g. `4×[1, 1+J]`). *(raised by wzy)*

### Low

**`compaction_admission.go:132` — global bucket + per-group/per-collection `admit()` starves later candidates within a single sweep**
`getSingleCompactionAdmitter()` is a process-wide singleton, but `admit()` is called per candidate group: the L2 path calls `triggerOneCollection → admit` once per collection, and the L1 path calls `generatePlans → admit` once per group. Within a single sweep the bucket barely refills, so the first collection/group drains the shared bucket and every later one sees it empty and is fully deferred regardless of its dirtiness. If `GetCollections()` returns a stable order, the last collection is starved every sweep; even with an unstable order this is an order-dependent cross-collection starvation within the sweep. *Suggestion:* admit over the merged candidate set once, or implement per-collection fair-share. *(raised by wzy)*

**`compaction_admission.go:127` — monotonic `hardCapCount` is lost across process restart**
`currentHardCapCount` is monotonic only within the process (`hardCapCount` is a struct field that prevents releasing a whole cohort in one round when the config is lowered); after a restart the global singleton is rebuilt and recomputes it from the current config. If an operator lowers `SingleCompactionDeltalogMaxNum` and then restarts, the post-restart hard cap becomes `4×` the new lower value, and a batch of already-large segments all take the bypass in the first round after restart — the very avalanche the monotonic cap is meant to prevent, reset by the restart. This is a corner case (restarts are rare). *Suggestion:* note in the `currentHardCapCount` comment or config docs that the monotonicity does not survive a restart. *(raised by xiaocai)*

czs007:
Addressed in `b276724c7c`. This round drove a small redesign of the admitter's ordering/escape, plus a description fix.

**[High] Delete-size / delete-ratio triggers had no bounded deferral.** Correct — the escape only looked at file count and the sort only at row ratio, so a size-triggered segment (varchar PK, large deletes in few files, low ratio) sorted last and never hit the cap. Reworked both to a single **normalized delete pressure** = `max` over the three delete axes (`count/maxnum`, `deletedRatio/ratioThreshold`, `deleteSize/sizeThreshold`) of `actual/threshold`. The escape now fires at `pressure >= deferralHardCap`, and the dirtiest-first sort uses the same value, so a segment triggered by *any* single axis is both ordered and bounded on that axis. New test `delete-size axis reaches the bounded-deferral escape` builds a 1-file, ~0-ratio, large-byte segment and asserts it bypasses.

**[Med] Hard-cap escape had no jitter — cohorts re-synchronize at 4x and avalanche through the bypass.** Correct and subtle. The escape is now `pressure >= deferralHardCap × singleCompactionThresholdMultiplier(segID)`, i.e. the same per-segment jitter that de-synchronizes the 1x trigger also de-synchronizes the 4x escape, so a deferred cohort reaches the escape at spread-out times instead of all at once. As a bonus this subsumes the previous monotonic-hard-cap logic (now removed): a threshold config *decrease* raises pressure into a per-segment spread of escape points rather than releasing the cohort in one round — which also makes the earlier *[Low] monotonic-cap-lost-across-restart* observation moot (there is no monotonic state to lose).

**[Med] PR description overclaimed expiry coverage.** Fixed. The description now states the actual guarantee: the token-bucket rate ceiling applies to **delete accumulation only**; expiry accumulation, strict age-TTL and index rebuild bypass admission and are **de-synchronized by jitter only**, not rate-capped. The \"any mass eligibility becomes a paced stream\" wording is gone.

**[Low] Global bucket, per-group/per-collection `admit()`, within-sweep ordering.** Kept intentionally, now stated plainly in the description and code: the bucket bounds *aggregate* rewrite load against a source-agnostic object store, so a single global budget is the point — per-collection sharding would multiply the aggregate ceiling by the collection count and defeat the protection. It refills every round, and sustained one-sided saturation trips the throttle WARN. Global cross-collection fair-share (buffer all callers' candidates behind a coordinator, then admit once) is a reasonable follow-up if fairness proves to matter in practice, but I'd rather not add that coupling to a production hotfix.

**[Low] `hardCapCount <= 0` / non-positive threshold.** Also resolved by the pressure rework: an axis whose threshold is `<= 0` is skipped entirely, so a misconfigured threshold disables that axis rather than turning the escape into a free bypass. Test renamed to `non-positive deltalog-max does not disable pacing`.

Dropped `segmentDeletedRowsRatio` / `segmentDeltalogCount` / the monotonic `hardCapCount` field in favor of `segmentDeleteCompactionPressure`. All datacoord compaction suites plus `compaction_admission_test.go` pass locally.

tinswzy:
Adversarial review confirmed 2 documentation gaps and 1 test-coverage gap, all low-severity and non-blocking.

### Low severity

- **`pkg/util/paramtable/component_param.go:5453`** — The doc for `SingleCompactionRateLimitTokens` says "Max delete/expiry-accumulation single-compaction candidates admitted per rateLimitInterval," but expiry accumulation is never admitted through the bucket (`admissionGated()` returns true only for `reasonTooManyDeletions`); it also describes the escape as "deltalog count exceeds 4x the configured maximum," when the real escape uses normalized max pressure across all three delete axes (deltalog count, deleted-rows ratio, delete-log size), jittered per segment. *Suggestion:* say "delete-accumulation candidates only," add expiry accumulation to the never-gated list, and describe the escape as normalized-pressure ≥ 4x on any delete axis (jittered). *(raised by wzy)*

- **`internal/datacoord/compaction_trigger.go:830`** — The comment at lines 788–792 notes that a delete-heavy segment that is also index-old is still paced (rebuild is checked after deletions), but omits that a delete-heavy segment that also trips expiry-accumulation is *not* paced, since expiry is evaluated before deletions in `singleCompactionReasonFor` and bypasses admission gating. The behavior itself is the PR's intentional design (retention must not be starved behind a delete backlog); only the comment is misleading. *Suggestion:* extend the comment to call out the expiry co-occurrence explicitly. *(raised by wzy)*

- **`internal/datacoord/compaction_trigger_v2.go:478`** — The refund drop-path wiring (the `submitted`-flag defer at `compaction_trigger_v2.go:477-484` and the `refundUnusedTokens` closure at `compaction_trigger.go:389-453`) has no end-to-end test; current tests exercise `refund` only in isolation (`compaction_admission_test.go:238-263`). *Suggestion:* add an integration test that drives a gated view/plan through the drop path — e.g. force `inspector.isFull()` or an enqueue failure with an `admissionGated` segment and assert the bucket recovers the token. *(raised by wzy)*

czs007:
All three addressed in `7490c80b4a` (docs + a test).

**[Low] `SingleCompactionRateLimitTokens` doc inaccurate.** Rewritten: it now says "delete-accumulation candidates only," lists expiry accumulation alongside strict age-TTL and index rebuild as never-gated (de-synchronized by jitter only), and describes the escape as normalized delete pressure (max over deltalog count / deleted-rows ratio / delete-log size vs threshold) reaching 4x on any axis, scaled by the segment's jitter multiplier.

**[Low] `singleCompactionReasonFor` comment omitted the expiry co-occurrence.** Extended: the comment now spells out all three first-match interactions — delete+index-old is paced (deletions checked before rebuild); delete + strict-age-TTL / expiry-accumulation bypasses (those precede deletions, and retention must not be starved behind a delete backlog); expiry/age-TTL/index-only bypasses.

**[Low] Refund drop-path had no end-to-end test.** Added `TestSubmitSingleViewRefundsGatedTokenOnDrop` to the trigger-manager suite: it drives an `admissionGated` `MixSegmentView` through `SubmitSingleViewToScheduler` with an injected AllocN failure (the drop path) and asserts the shared bucket recovers the token, plus a non-gated companion that asserts the bucket is left untouched. The `refund`-in-isolation unit test stays.

Thanks for the reviews across all four rounds — the escape/ordering ended up materially more correct for it. All datacoord compaction suites plus `compaction_admission_test.go` pass locally.

tinswzy:

Issue #51440, opened 2026-07-16
enhance: make the stats-based chunk skip index correct and usable on storage v2


xiaofan-luan:
## Summary

The chunk-level `SkipIndex` built from parquet footer statistics (gated by `common.parquetStatsSkipIndex.enabled`, default **off**) is **latently incorrect** on storage v2, and even when it does prune it still materializes the pruned data. This makes the flag unsafe to enable and largely ineffective. This tracks making it correct and useful.

### 1. Cell / row-group misalignment (the original report)

On storage v2 (packed parquet), a cache **cell** packs *multiple* parquet **row groups** (`queryNode.segcore.storageV2.cellTargetSizeBytes`, ~4 row groups by default). But the footer-stats path builds **one skip metric per row group** and indexes it per **cell** (`CellIdMappingMode::IDENTICAL`). When `rgs_per_cell > 1`, the metrics are misaligned with cells:

- **False negative → wrong results.** A cell can be pruned using only its *first* row group's min/max, even though a *later* row group in the same cell still holds matching rows. Those rows are silently dropped from query results. This is why the flag cannot be safely enabled today.
- **No IO saved on prune.** Even when a cell is correctly skipped, the executor still fetches its data/validity to evaluate the (now all-false) predicate.

### 2. Further silent-drop bugs found while fixing the above

Each of these can prune a chunk that actually holds matching rows:

- `BooleanFieldChunkMetrics::CanSkipIn` decodes its input as `{contains_true, contains_false}` flags while the producer emits a value list → `bool_col IN (true, false)` can skip an all-false chunk.
- `StringFieldChunkMetrics::CanSkipIn` seeds its IN-list min/max hull with `string_view::empty()` as an "unset" sentinel, but `""` is a legitimate value → `s IN ("", "zzz")` prunes chunks holding empty strings.
- `SkipIndexStatsBuilder::Build` reads `min()/max()` from statistics reporting `!HasMinMax()` (an all-null row group, or a float row group containing NaN) → garbage bounds.
- `ProcessFieldMetrics` dereferences a failed `dynamic_pointer_cast` → crash when the footer's physical type does not match the field's data type.
- **VARCHAR use-after-free.** Arrow's BYTE_ARRAY `parquet::Statistics` expose min/max as `string_view`s into the reader's thrift memory, but the metrics are built *after* the metadata loader closes its reader → VARCHAR skip bounds are read from freed memory (silent wrong results on Linux, crash on macOS).
[...]

PR #49168, opened 2026-04-20
enhance: lazy offset mapping for nullable vector columns

marcelo-cjl:
## Summary

- `BuildValidRowIds` currently pins every chunk of a nullable vector column at segment load, which triggers a full-column S3 fetch for external tables and defeats `warmup=disable` for internal tables.
- Split `OffsetMapping` into `Init` (skeleton) + per-chunk `BuildChunk`, and route the load path through `InitValidRowIds` when per-row-group Parquet `null_count` metadata is available (storage v2 column groups). The skeleton populates per-chunk cumulative counts without pinning any chunk; l2p/p2l entries are filled lazily via `EnsureChunkOffsetMapping` on first touch.
- Decouple Parquet stats collection from `ENABLE_PARQUET_STATS_SKIP_INDEX`: stats for nullable vector fields are always collected (required for the lazy path), while the flag continues to gate skip-index usage and stats collection for other fields. v1 binlog path is unchanged and keeps the eager `BuildValidRowIds` fallback.
- Search-time consumers (`SearchOnSealedColumn`, `FilterVectorValidOffsets`, `InterimSealedIndexTranslator`) only ensure the chunks they actually touch; brute-force search pins all chunks as before.

## Test plan

- [x] New unit tests in `OffsetMappingTest.cpp` cover init + per-chunk build, vec/map mode, out-of-order fills, idempotency, concurrent builds on distinct chunks
- [x] New `ChunkedColumnInterfaceTest.cpp` uses `TYPED_TEST` to run the lazy-path invariants against both `ChunkedColumn` and `ProxyChunkColumn` (init without pin, ensure-only-touched, no-op on eager path, non-nullable no-op)
- [x] Full regression sweep on `OffsetMapping*`, `test_chunked_column*`, `ChunkedColumnInterfaceTest*`, `*Sealed*`, `*Interim*`, `*Filter*`, `*SkipIndex*` — 450+ tests, 0 failures
- [ ] CI integration suites

issue: #49110
related: https://github.com/milvus-io/milvus/issues/45993

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #51441, opened 2026-07-16
enhance: force one row group per cell for the stats skip index (storage v2)

xiaofan-luan:
## What & why

`common.parquetStatsSkipIndex.enabled` builds chunk-level skip metrics from Parquet footer statistics for Storage V2 sealed segments.

The previous implementation was unsafe and did not reliably save IO:

- Parquet statistics were positional per row group, while one cache cell could contain multiple row groups. Applying one row-group metric to a differently packed cell could produce false-negative pruning and silently drop matching rows.
- VARCHAR min/max values could outlive the Parquet metadata that backed their non-owning views.
- Missing, incompatible, or stale metrics could be associated with the wrong cell after a column replacement.
- Some scan and prefetch paths still pinned or materialized cells after the skip index had already ruled them out.

This PR makes the Storage V2 footer-statistics skip index conservative and generation-safe, moves pruning ahead of cell IO on the execution paths that queries use, retires the Storage V1 data-derived skip index that this seam supersedes, and enables the feature by default.

Issue: #51440
Design doc: `docs/design-docs/design_docs/20260903-skip-filter-interface.md`

## Removed: the Storage V1 chunk skip index

This PR also **removes** the Storage V1 data-derived chunk skip index. It is the change
with the widest blast radius, so it is stated here rather than left to be inferred from
the diff. The rationale and the full consequence analysis are in the design doc,
section *Storage V1: no chunk pruning under this interface*.

Before this PR, `load_field_data_common` bound every non-proxy scalar/string sealed column
to a `FieldChunkMetricsTranslator` that derived per-chunk min/max lazily from the chunk
data. That branch ran unconditionally, independent of `common.parquetStatsSkipIndex.enabled`.
After this PR the only producer of real metrics is `LoadGroupChunkMetadata`, which requires
the flag **and** Storage V2. Metrics are consumed through the column itself: every column is
a `FieldChunkMetricsProvider`, `FieldSkipMetricsView::FromProvider(column)` resolves the
view, and ordinary V1 columns expose a null metrics list, so they fail open and never prune.
There is no segment-level field-to-metrics map to keep in sync with column replacements.

| Segment | Before | After |
| --- | --- | --- |
| Storage V1 sealed | data-derived pruning, always on | no pruning |
| Storage V2 sealed, flag off | no pruning | no pruning |
| Storage V2 sealed, flag on (now the default) | footer-stats pruning, unsafe as described above | footer-stats pruning, conservative and generation-safe |
| Storage V3 / manifest | no pruning | no pruning |

The removal is deliberate: Storage V1 is being phased out and no longer warrants a second,
separately-shaped skip-metrics lifecycle whose bookkeeping the new seam cannot keep in sync.
Queries stay correct — the skip index only ever removed work — but clusters still holding V1
segments read more chunks for selective range and `IN` predicates. Storage-version compaction
(`dataCoord.compaction.storageVersion.enabled`, on by default) converts V1 segments to V2, so
that exposure shrinks over time. Restoring V1 pruning through the same
`FieldChunkMetricsProvider` seam is a follow-up, kept out of this PR so its memory and
load-time cost can be measured on their own.

Two consequences worth flagging for review:

- The V1-only regressions that drove pruning through the lazy-compute path
  (`Sealed.SkipIndexSkip{Unary,Binary}Range{,Nullable}`, `Sealed.SkipIndexSkipStringRange`,
  `TaskTest.SkipIndexWithBitmapInputAlignment`, and
  `OffsetsEvalProductionRegressionTest.ConjunctOffsetInputKeepsMatchAfterRealSkipIndexPruning`)
  are re-expressed against the Storage V2 footer path.
- The arrow/`RecordBatch` half of `SkipIndexStatsBuilder` (`LoadMetrics`, `WithNullState`,
  the bloom-filter and n-gram construction, `BooleanFieldChunkMetrics`) no longer has a
  production caller. It is retained for the Storage V3 producer tracked in Follow-ups; if that
  slips it should be deleted rather than left as dead weight.

## Correctness

- Force exactly one Parquet row group per cache cell for **every** column group while the flag is on. Packing is a function of the startup flag alone — never of schema content or footer content — so a segment's cache layout is deterministic and identical for vector-only, pk/ts, and scalar groups alike. Skip metrics are still only collected for the footer-statistics-skippable types INT8, INT16, INT32, INT64, FLOAT, DOUBLE, and VARCHAR; groups and fields without usable statistics simply get no filter. This gives skip metrics and cache cells a 1:1 positional mapping wherever metrics exist.
- Build VARCHAR metrics while the Parquet reader and metadata are alive, then retain deep-owning immutable bounds.
- Preserve one metric entry per row group, across all files of the group in load order. Missing or unusable statistics become `NoneFieldChunkMetrics`, which never prune, so holes cannot shift later metrics onto another cell.
- Validate each field's metric list twice: `PlanGroupChunkPacking` drops a field whose list length differs from the total row-group count or that has neither usable bounds nor a known all-null chunk (such a list can never prune; dropping it avoids binding a useless filter), and `GroupCTMeta::InstallSkipMetrics` re-checks list length against the loaded cell count as a producer-independent backstop. Any mismatch disables footer pruning for that field and reads the data.
- A `FieldSkipMetricsView` retains a `shared_ptr` to the column generation it resolved from. Replacing or dropping a field updates `runtime.fields` under copy-on-write publish; a view resolved before the replacement keeps the old column (and its metrics) alive, a view resolved after sees the new generation. A Storage V1/V2 remap cannot reuse stale metrics.
- Fix conservative behavior for BOOL `IN`, empty-string VARCHAR `IN`, missing or one-sided min/max values, and Parquet physical-type mismatches.
- Disable skip-index pruning for all arithmetic predicates. Arithmetic expressions still execute normally, but they cannot introduce overflow, truncation, rounding, or comparator-boundary false negatives through the skip index.
- Preserve SQL three-valued logic by gating the filter itself (`CanUseSkipFilter`): an expression binds a skip filter only for a non-nullable field or a null-rejecting consumer. Null-rejection propagates through `AND`/`OR` and stops at `NOT`; every top-level filter consumer (FilterBits, iterative filter, rescore) is null-rejecting, so a standalone predicate on a nullable field still prunes. A nullable field whose consumer distinguishes UNKNOWN from FALSE — anywhere below a `NOT` — binds no filter and evaluates the data normally, so `NOT` never turns a skipped NULL into a match by construction.

Unsupported or invalid statistics always fail open: they may lose a pruning opportunity, but they do not remove matching rows.

## IO avoidance

Pruning now happens before a sealed cell is pinned or materialized on both sequential scans and offset-input or iterative-filter scans.

The scan path, driver prefetch, and fallback prefetch share the same skip decision. A pruned cell is therefore not fetched later by another stage.

VARCHAR `IN` uses an owned `std::string` value list on both scan and prefetch paths, avoiding the previous `std::string` and `std::string_view` mismatch.

Known boundary: an `IN` list whose literals all fall outside the column type's range (`int8_col IN (300, 400)`) is now treated like a genuinely empty `IN` — `CanSkipIn` fails open on an empty value set, so the column is scanned and the (correct) empty result is computed from the data. The earlier always-prune shortcut for this case was dropped together with the `CanUseSkipFilter` simplification; it saved IO only for provably-empty `IN` lists and carried the subtlest machinery in the change (an element-level carve-out and a `type_info` guard). Restoring it is a small, separable follow-up.

## Reopen consistency

The current base includes #51623, which provides request-level `SegmentReadLease` protection for sealed segments:

- Production Search, Retrieve, and RetrieveByOffsets acquire the lease before expression execution.
- Search retains the lease in `SearchResult` through reduce, export, and output-field materialization. Retrieve holds it through result serialization.
- `PublishStateOnline` waits for active readers to drain before atomically publishing a Reopen state.
- Reopen stages changes in cloned runtime state instead of mutating the currently published snapshot.

The skip view is resolved from the same runtime snapshot as the data it describes: `GetFieldSkipMetrics` and `GetDataScanResources` read `runtime.fields` once and return the column and its metrics view together, and the bound filter retains that generation for the cursor's lifetime. Consequently, Reopen cannot publish runtime B between skip-metric resolution and chunk-layout or data access in the same production request. Direct internal C++ callers that combine skip metrics with chunk layout or data access must hold `SegmentReadLease`, as documented in `SegmentInterface.h`; all production Search and Retrieve entry points do so.

## Configuration and scope

- `common.parquetStatsSkipIndex.enabled` is **enabled by default**.
- The setting is startup-only and not refreshable because it determines row-group-to-cell packing. Changing it requires a restart and segment reload. Enforcement is structural: the value is pushed to the core once during QueryNode/StreamingNode init and no runtime update path exists.
- Footer-statistics pruning in this PR is Storage V2 only. Manifest-based Storage V3 segments do not build this skip index.
- The nullable-vector footer-statistics collection added in #49168 is fully retired, not merely narrowed. `CollectSkipStatsFields` gathers only types `SkipIndexStatsBuilder::Build` produces real metrics for, and vector types fell to its `default` arm yielding `NoneFieldChunkMetrics`, so those statistics could never prune. The `null_count`-driven offset mapping they fed has been superseded by `BuildValidRowIds()`, which reads real validity. A backport should not reinstate the dependency.
- Enabling it has a real cache cost, now borne by every column group: cells are as fine as one row group, so a group of 64 KiB row groups against the 4 MiB default `cellTargetSizeBytes` goes from roughly one cell to roughly 64, multiplying cache-slot metadata, pin/unpin calls and eviction bookkeeping. Two mitigating factors: finer cells reduce the bytes a point lookup or `Take` must pull in, and the per-cell overhead is proportional to row-group count, which the writer already controls. One known accounting artifact: `GroupChunkTranslator::cells_storage_bytes` floors each cell at 1 MiB for scanned-bytes and cache-byte metrics, so segments with small row groups will report inflated scanned bytes; keying that floor to actual cell size is a follow-up. The footer metrics themselves live in `GroupCTMeta` for the column group's lifetime and are not yet counted in the tiered-cache budget (one `FieldChunkMetrics` per row group per skippable field; VARCHAR entries own two `std::string` bounds) — bounding that is also a follow-up. Disabling the flag restores target-size packing everywhere and disables footer pruning.

## Tests

Storage V2 query correctness and executed-path coverage:

- `SkipIndexPr51441.StorageV2SkipQueryResultsCorrect`
- `SkipIndexPr51441.StorageV2CellPruneByFlag`
- `SkipIndexPr51441.StorageV2PackingFollowsFlagRegardlessOfFooterMetrics` — packing follows the flag for every group, including groups with statistics disabled or all-null columns; fields without usable bounds and without all-null chunks get no filter
- `SkipIndexPr51441.StorageV2VarcharInPrunesOnExecutedPath`
- `SkipIndexPr51441.StorageV2OneSidedVarcharFooterStatsFailOpen`
- `SkipIndexPr51441.PrunedCellsAreNotPrefetchedOrPinned`
- `SkipIndexPr51441.ArithmeticPredicatesDoNotUseSkipIndex`
- `SkipIndexPr51441.NullableSkippedCellsPreserveNotSemantics` — `NOT` over a nullable field never turns a NULL into a match
- `SkipIndexPr51441.OutOfRangeBinaryRangePrefetchMatchesScan`

`SkipIndexStatsBuilderTest` also covers missing statistics, physical-type mismatches, BOOL and empty-string `IN`, integral bounds, and VARCHAR bound ownership. `ConjunctExprTest.SkipFilterEligibility` pins the `CanUseSkipFilter` truth table; `FieldSkipMetricsViewTest` pins the provider tri-state (list / null list / per-chunk fallback) and generation retention across rebind and erase.

## Review follow-ups

The branch is squashed to one commit; these are the review fixes folded into it:

- `PhyUnaryRangeFilterExpr::PrefetchRawData` read its literal with `GetValueFromProto`, which truncates an out-of-range value to `T{}` without reporting the overflow, so `int8_col < 300` pruned every cell whose minimum was `>= 0`. It now checks `query::out_of_range` and prefetches exactly what the scan will touch — `PreCheckOverflow` answers such a predicate from the literal alone and reaches the column only to materialize validity, which `ApplyFieldValidData` skips for a non-nullable field. `PhyBinaryRangeFilterExpr` already had this guard; the unary sibling did not.
- `GetSkipIndexValues` used `GetValueFromProtoWithOverflow` for float/double, which asserts `kFloatVal`, while the `GetValueWithCastNumber` it replaced accepted an int64 literal — the assert would have surfaced as an opaque query failure from inside the prefetch future.
- The prepared-value cache hangs off a single `std::once_flag` but is built from a `T`-dependent `ValueType`. It records the type and fails open if the `PrefetchRawData` and `Eval` dispatch switches ever disagree, instead of reinterpreting the cached variants. The `type_info` comparison is by value, not address.
- `ProcessDataChunksForChunkedSegment` called `GetSkipIndex()` inside the per-chunk scan loop, and `GetSkipIndexSnapshot` allocated a fresh empty `SkipIndex` whenever the segment had none. The snapshot is hoisted, the empty instance is shared, and the scan reuses the prefetch loop's verdicts so a Term `IN` hull is not recomputed per cell. `ProcessDataChunksForElementLevel` and `ProcessDataChunksForSingleChunk` had the identical pattern — the latter is the growing-segment path, where the view is always empty and a throwaway was allocated per chunk per batch. All five call sites now capture the view once, which also keeps every chunk of a batch judged against a single resolved generation.
- `PhyBinaryRangeFilterExpr::PrefetchRawData` abandoned pruning whenever a literal did not fit the column's physical type and fell back to the whole-field prefetch, while the scan's `PreCheckOverflow<T>` resolves the same literals — a bound past the far end of `T` makes the range provably empty, a bound past the near end constrains nothing — and keeps pruning with the clamped bounds. With driver prefetch on, `3000000000 <= int32_col <= 3000000001` matched nothing and the scan read no cell, yet the prefetch pulled in the whole column (83 MB in the test fixture). The prefetch now mirrors `PreCheckOverflow`. Over-fetch only, but it broke the shared-skip-decision property claimed above.
- The metrics/cell alignment rule (a field's metric list must have exactly one entry per cache cell, or be dropped) lives in `GroupCTMeta::InstallSkipMetrics` with `skip_metrics_by_field_` private behind it, so every producer — including the test fixture — inherits the check by construction. `FindSkipMetric` absorbs the bounds guard `ChunkedColumnGroup` used to open-code. `PlanGroupChunkPacking` adds the prune-worthiness cleanup on top (misaligned lists and lists with neither usable bounds nor all-null chunks are dropped before install).
- `PhyTermFilterExpr::CanSkipSegment` value-initialises the `min`/`max` hull (an empty `IN` on a primary key previously reached `CanSkipBinaryRange` with indeterminate bounds) and fails open on an empty literal list.
- The per-cell skip check used to perform two field lookups (`SkipIndex`'s field → provider map, then the column group's field → list map) before indexing by `chunk_id`, and that cost scaled with the row-group count once cells are 1:1 with row groups. The segment-level map is now gone entirely: columns are their own `FieldChunkMetricsProvider`, `FieldSkipMetricsView::FromProvider` resolves the optional list once, and every scan and prefetch path indexes the view by `chunk_id`. The view owns a reference to the column generation it resolved, so a rebind or erase cannot change what an in-flight expression sees (`FieldSkipMetricsViewTest.RetainsResolvedGenerationAcrossRebindAndErase`). `skip_func` is `bool(const FieldSkipMetricsView&, int64_t)`, which also removes the `int` narrowing of chunk ids. `GetSkipMetricsList()` is tri-state so a producer that implements only the original per-chunk contract keeps pruning through a per-cell callback, while list-capable providers with no list for the field — Storage V1 columns included — never pay a per-cell call. The generic `SkipIndex` helper remains only for the `JsonKeyStats` callback interface and standalone tests.
- The nullable-validity handling was simplified from *skip the cell but still fetch its validity bitmap* (`SkippedChunkNeedsValidity`) to *bind no skip filter at all* (`CanUseSkipFilter`) for a nullable field with a validity-observing consumer. The old design required every scan, take, prefetch, and element-level path to thread a validity-only read for pruned cells, and one such path (the element-level skip branch, which hands `evaluate_batch` an empty `ValidityView`) was a review-found `NOT`-semantics bug. The new design makes the incorrect configuration unrepresentable, at the cost of losing pruning for nullable fields below `NOT`.
- `HasMetrics()` no longer reports a bound view for a known null or empty metrics list, so such fields bind no `ColumnFilter` at all instead of paying a per-cell callback that always fails open.
- `PlanGroupChunkPacking()` collects the force-one-row-group decision and its rationale in one place instead of copy-pasting it between the two storage-v2 column-group load overloads. The decision is the startup flag alone; it no longer depends on whether the group contains a skippable-typed field.

## Follow-ups

- Storage V3 footer-statistics skipping remains a follow-up on milvus-storage#583.
- Restoring chunk pruning for Storage V1 is a follow-up that plugs into the same `FieldChunkMetricsProvider` seam rather than reviving the removed path, kept separate so its memory and load-time cost can be measured on their own.
- Arithmetic skip-index pruning can be reconsidered only with an exact proof for overflow, conversion, rounding, and comparator boundaries.
- The element-level filter `ExprSet`s consume only data bits and could mark their roots null-rejecting, which would restore chunk pruning for nullable array fields.
- Restoring the always-prune verdict for `IN` lists whose literals all truncate out of the column type's range (dropped with the `CanUseSkipFilter` simplification; see *IO avoidance*).
- Footer metrics memory accounting, and keying the `cells_storage_bytes` 1 MiB floor to actual cell size (see *Configuration and scope*).


sparknack (commented):
**Blocking: Storage V2 skip metrics should be owned by the column group**

This change fixes the mutable segment-level map race and publishes `runtime.fields` and `runtime.skip_index_metrics` together. However, it still does not guarantee that a reader consumes the matching column and metrics snapshot.

`GetSkipIndex()` captures runtime A, while `chunk_size()`, `num_rows_until_chunk()`, and `get_batch_views()`/`chunk_data()` each capture the published state again. If Reopen publishes runtime B between these calls, metrics from A may be applied to B cell layout and data. Making the flag non-refreshable does not prevent replacement parquet files from having different row-group boundaries or counts. This can still cause false-negative pruning or out-of-range chunk access.

The new test only calls `CanSkip()` on an old view after dropping the field. It does not read data through the column associated with that view, so it does not cover this torn-read scenario.

Suggested ownership:

```text
Storage V1
  Segment::SkipIndex                // keep as-is

Storage V2/V3
  ProxyChunkColumn
    -> ChunkedColumnGroup
      -> GroupCTMeta::skip_metrics  // column-owned
```

For V2/V3:

- Store immutable per-field/per-cell metrics in `GroupCTMeta`.
- Let `ProxyChunkColumn` expose the metrics for its own field and chunk.
- Capture one column handle in the expression and use it for chunk count, row offsets, skip metrics, and `Span`/views.
- Missing statistics should return `NoneFieldChunkMetrics`; they must not fall back to the V1 segment index.
- Preserve missing row-group statistics as `NONE` entries instead of deleting them, so other aligned cells can still be pruned safely.

This makes the invariant structural: data, cell layout, and skip metrics have the same owner and lifetime. It removes parallel segment state, prevents torn reads during Reopen, lets old readers safely retain the old column group, and keeps skip lookup independent of materializing `GroupChunk` data. The footer-statistics path already builds all metrics eagerly, so it also removes the extra `CacheSlot<FieldChunkMetrics>` and per-cell cloning layer.

xiaofan-luan:
Good point — agreed the RuntimeResourceState-level pairing was not sufficient, since each segment accessor re-captures the published state. Adopted the column-group ownership in ff3fc13:

- `GroupChunkTranslator` now builds per-field/per-cell `FieldChunkMetrics` into `GroupCTMeta` at construction, from the same footer read that fixes the cell layout — metrics, layout, and data share one owner and lifetime. Skip lookups never materialize a `GroupChunk`, and the `CacheSlot<FieldChunkMetrics>` layer + per-cell cloning are gone for the footer path (`FieldChunkMetricsTranslatorFromStatistics` deleted).
- Missing row-group statistics are preserved as positional `NONE` entries instead of being dropped, so aligned cells still prune and holes can never shift metrics onto the wrong cell; the segment-level alignment guard is removed.
- `ProxyChunkColumn` exposes its field's metrics via `ChunkedColumnInterface::GetSkipMetrics` (default nullptr → never skip, no fallback to the v1 segment index). The `SkipIndex` view resolves v2/v3 fields through the captured `fields` map, so the metrics always come from the very column object the snapshot holds; storage v1 keeps its lazily-built slot staged in the runtime state.

One piece deliberately deferred to a separate PR: capturing one column handle in the expression and routing `chunk_size` / `num_rows_until_chunk` / views through it. That touches all data-access dispatch in `SegmentExpr`; noted as an explicit follow-up in the PR description. Until then the expression memoizes all skip decisions from a single snapshot per expression, so decisions are internally consistent; the residual window is data reads re-capturing a newer state, which the follow-up closes.

xiaofan-luan:
Note: after re-ordering, this PR is now **stacked on top of #51602 (per-operation snapshot pinning)** — the first commit here is the pinning commit. Review only the top two skip-index commits. Once #51602 merges, this PR rebases to skip-index-only. (Pinning was split out so it can merge first.)

xiaofan-luan:
The Add/Sub integer-overflow case is a pre-existing issue and is intentionally out of scope for #51441. This PR does not change that inversion path; its scope is the Storage V2 cell alignment, IO pruning, and observability work. We will not fix the Add/Sub overflow behavior in this PR. The PR description should also stop calling Add/Sub pruning unconditionally “exact”, since that wording is too broad in the presence of integer wraparound.

xiaofan-luan:
Addressing the blocking review above against the current minimal head `efeff1d`.

The review was valid for the old branch. Since then, #51623 (`a0fd88b573`) has merged into the base and established a request-level `SegmentReadLease` for sealed-segment reads:

- `AsyncSearch`, `AsyncRetrieve`, and `AsyncRetrieveByOffsets` acquire the lease before expression execution (`segment_c.cpp`).
- Search transfers the lease into `SearchResult`, so reduce, export, and output-field fill remain protected; Retrieve holds it through result serialization.
- `PublishStateOnline` marks the writer pending, waits for all active readers to drain, and only then atomically publishes the Reopen state. Reopen staging clones the runtime and `SkipIndex`; it does not mutate the currently published snapshot.

Therefore, while a production request is alive, Reopen cannot publish runtime B between `GetSkipIndex()` reading runtime A and later `chunk_size()`, `num_rows_until_chunk()`, or data/view access. All of those captures observe one published generation. QueryNode retries the narrow gate-busy admission failure.

The current patch also keeps footer metrics positional with `NONE` entries for missing statistics, forces one row group per cell when stats pruning is enabled, and fails open if the metrics count does not match the cell count.

This deliberately does not add the proposed compound column handle. Direct internal C++ callers that combine skip metrics and column access must hold `SegmentReadLease`, as documented in `SegmentInterface.h`; all production Search/Retrieve entry points do so.

Could you please re-review the current minimal head and confirm that this alternative generation-stability invariant addresses the blocking concern?

czs007:

Verified:
- `internal/core/src/segcore/storagev2translator/GroupChunkTranslator.cpp:170` — forcing one row group per cell establishes the required 1:1 positional mapping, while groups without skippable fields keep normal packing.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2398` — metrics are built while the Parquet metadata is still alive and each row group contributes exactly one owning metric, so VARCHAR bounds cannot dangle and no positional holes appear.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:7037` — replacement erases both metric sources before installing the new one, and the metric-count check fails open on layout mismatch.
- `internal/core/src/index/SkipIndex.cpp:32` — missing fields and out-of-range chunk IDs resolve to `NoneFieldChunkMetrics`, so stale or malformed lookups cannot prune data.
- `internal/core/src/index/skipindex_stats/SkipIndexStats.cpp:38` — null statistics, one-sided or missing bounds, and physical-type mismatches conservatively produce non-pruning metrics.
- `internal/core/src/exec/expression/Expr.h:1981` — nullable validity is still materialized whenever three-valued logic remains observable, while null-rejecting consumers can safely skip the cell read.

czs007:

**Verified:**

- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2435` — metrics are built inside the reader lambda while the parquet `FileMetaData` is still alive and `Build()` deep-copies into owned `std::string`, closing the BYTE_ARRAY `string_view` aliasing hazard for VARCHAR bounds.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:7094` — metric count is validated against `column->num_chunks()` and pruning is disabled (log + fail open) on any mismatch, so a misaligned metric vector can never prune a cell.
- `internal/core/src/segcore/storagev2translator/GroupChunkTranslator.cpp:177` — one row group per cell is forced only when footer metrics are actually usable, which is what preserves the 1:1 row-group→cell positional mapping the metric vector depends on.
- `internal/core/src/index/SkipIndex.h:187` — clones share the immutable `shared_ptr<const>` metric set under the read lock, while `:195` and `:393` install generation-local replacements and each load path erases the other metric source, so a v2→v1 replace cannot prune a new column with stale bounds.
- `internal/core/src/exec/expression/Expr.h:1942` — the `SkippedChunkNeedsValidity` gate is applied consistently across the sequential, by-offsets batch, by-offsets scalar and fallback-prefetch paths, so a leaf under `NOT` still materializes validity and three-valued semantics are preserved.
- `internal/core/src/index/skipindex_stats/SkipIndexStats.cpp:38` — missing, one-sided, or physically incompatible statistics degrade to `NoneFieldChunkMetrics` before any `min()`/`max()` is read, so no path consumes an unset bound.

## Non-blocking observations (low / nitpick)

These do not require a change in this PR; they are recorded for the author's awareness.

**low — `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2367`: silent behavior change — nullable vector fields no longer get parquet statistics collected.**
`CollectSkipStatsFields` collects only `IsStatsSkippableType` fields (INT8/16/32/64, FLOAT, DOUBLE, VARCHAR), replacing the previous `else` branch that unconditionally collected footer statistics for `is_nullable() && IsVectorDataType(...)` fields. This is not a regression: vector types fall to the `default` arm of `SkipIndexStatsBuilder::Build` and yield `NoneFieldChunkMetrics` (`SkipIndexStats.cpp:129-130`), so those metrics could never prune anything; every non-test `GetSkipIndex()` consumer is a scalar expression; and the `null_count`-driven offset mapping from #49168 has been superseded by `BuildValidRowIds()`, which reads real validity via `chunk->isValid(j)` (`ChunkedColumnInterface.h:257`) and does not touch parquet statistics. Storage V2 column groups also pass `is_proxy_column=true`, so removing the statistics does not fall back to `LoadSkip`. Suggestion: add one line to the PR description confirming the #49168 nullable-vector statistics path is fully retired, so a future backport does not reinstate a now-dead dependency. (surfaced during verification)

**nitpick — `internal/core/src/segcore/storagev2translator/GroupChunkTranslator.cpp:177`: `force_one_row_group_per_cell` also fragments non-skippable fields in the same column group.**
The flag is computed per column group (`skip_index_enabled && !fields_for_stats.empty()`), so a single skippable field pins `rgs_per_cell` to 1 for the whole group; co-located JSON/ARRAY/BOOL fields lose `cell_target_size_bytes` packing for zero pruning benefit. Pure vector groups are unaffected and the flag defaults to false, so the impact is limited to explicitly enabling the flag with mixed scalar + JSON/ARRAY column groups. The trade-off is already acknowledged in the comments and correctness is unaffected. Suggestion: for wide mixed groups, consider noting the cache-slot/IO-batch cost in the flag documentation, or revisit a per-field granularity if such layouts become common. (surfaced during verification)

czs007:

One low-severity finding from this round: the new narrow-integer prefetch path in `PhyBinaryRangeFilterExpr` abandons skip-index pruning on out-of-range literals where the scan path still prunes, costing avoidable IO but not correctness.

## Low

**`internal/core/src/exec/expression/BinaryRangeExpr.cpp:1358`** — `PrefetchRawData<T>()` early-returns to `SegmentExpr::PrefetchRawData(field_id_)` whenever `query::out_of_range<T>(lower_val) || query::out_of_range<T>(upper_val)`, and that overload (`Expr.h:2736-2738`) takes no chunk list, so it prefetches *every* cell of the field. The scan path handles the same literals differently: `PreCheckOverflow<T>` (`BinaryRangeExpr.cpp:386-400`) either short-circuits to an empty result (`gt_ub<T>(val1)` / `lt_lb<T>(val2)`) or clamps the bound into `T` (`lt_lb<T>(val1) → T::min` with `lower_inclusive = true`, `gt_ub<T>(val2) → T::max` with `upper_inclusive = true`) and keeps pruning with the clamped values via `skip_index_func` (`:598-604`). On a storage-v2 sealed segment with `common.parquetStatsSkipIndex.enabled=true`, an INT8/INT16/INT32 binary range on the RawData path with an out-of-range literal — e.g. `1 <= int8_field <= 200`, or `200 <= int8_field <= 300`, which the scan resolves to an empty set without reading a single cell → the driver prefetch (`PrefetchAsync → PrefetchRawData`) pulls in every row group of that column, including ones the scan has already proven cannot match, producing unnecessary object-storage reads and cache pressure. Results stay correct (over-fetch only), and this is not a regression: before this PR the call deduced `T = int64_t` from `H`, so `std::holds_alternative<T>` in `IntFieldChunkMetrics<int8_t>::CanSkipBinaryRange` (`SkipIndexStats.h:528-531`) was always false and narrow-integer prefetch pruning never fired at all. It does, however, mean the PR's stated property that "the scan path, driver prefetch, and fallback prefetch share the same skip decision" does not hold for this query shape. Suggestion: reuse the `PreCheckOverflow` rules on the prefetch side instead of returning early — emit an empty prefetch list when the range is provably empty, and clamp (`lt_lb → T::min`/`lower_inclusive = true`, `gt_ub → T::max`/`upper_inclusive = true`) before pruning otherwise, so the two paths agree by construction rather than incidentally. (raised by czs007)

internal/core/src/index/SkipIndex.cpp
```diff
@@ -11,48 +11,22 @@
...
-    const std::vector<milvus::cachinglayer::cid_t>& cids) {
-    std::vector<std::pair<milvus::cachinglayer::cid_t,
-                          std::unique_ptr<index::FieldChunkMetrics>>>
-        cells;
-    cells.reserve(cids.size());
-    for (auto chunk_id : cids) {
-        auto pw = column_->GetChunk(ctx, chunk_id);
-        auto chunk_metrics = builder_.Build(data_type_, pw.get());
-        cells.emplace_back(chunk_id, std::move(chunk_metrics));
+    auto source = fieldMetricSources_.find(field_id);
```
sparknack: **Suggested follow-up: field-scoped skip metrics**

Each cell-level `CanSkip*()` call performs two field lookups: `fieldMetricSources_.find(field_id)` here (`unordered_map`), followed by `skip_metrics_by_field_.find(field_id)` in [ChunkedColumnGroup::GetSkipMetrics()](https://github.com/milvus-io/milvus/blob/fabe037bd5bdd05212e724e4a3bc701caa54fa45/internal/core/src/mmap/ChunkedColumnGroup.h#L186-L199) (`std::map`), before indexing by `chunk_id`. Hoisting `GetSkipIndex()` out of the loop retains both lookups in the hot path. With one row group per cell, this work scales with the row-group count.

Bind the column once when the expression selects its execution path, and resolve the field's metrics once when constructing `ProxyChunkColumn`, retaining their owner for the lifetime of the reference. Prefetch and scan can then access metrics directly by `chunk_id`.

Longer term, let each column own its immutable field metrics and remove the segment-level `SkipIndex` field map. This eliminates the need for Reopen to update both `runtime.fields` and `fieldMetricSources_`, and allows a dropped field's metrics to be reclaimed with its old column even while sibling fields retain the shared column group. Preserve `SegmentReadLease` so statistics, cell layout, and data remain from the same column generation.

xiaofan-luan: Thanks — addressed the per-cell lookup in `a978e7a6bf`.

`SkipIndex::ResolveField()` now performs both field lookups once and returns a `FieldSkipMetricsView` that holds the provider and the field's `SkipMetricsList*`. `SegmentExpr` resolves it in `InitSegmentExpr` (the same construction-time snapshot `num_data_chunk_` comes from) and every scan / prefetch path indexes the view by `chunk_id`; `skip_func` is now `bool(const FieldSkipMetricsView&, int64_t)`. The view owns a reference to the column generation it resolved, so a rebind or erase in the meantime cannot change what an in-flight expression sees — `FieldSkipMetricsViewTest.RetainsResolvedGenerationAcrossRebindAndErase` pins that down. `GetSkipMetricsList()` is tri-state so a producer implementing only the per-chunk contract keeps pruning, while list-capable providers with no list for the field (Storage V1 columns included) never pay a per-cell call. `SegmentReadLease` is preserved as before.

The longer-term step — columns owning their metrics outright and dropping the segment-level `fieldMetricSources_` map — touches Reopen, Erase and Clone together, so I'd like to take it in a follow-up PR rather than widen this one.

czs007:

**Verified:**
- `internal/core/src/segcore/storagev2translator/GroupChunkTranslator.cpp:179` — with `force_one_row_group_per_cell` each cell holds exactly one row group and cells never span files, so cells and skip metrics stay 1:1 in file order.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2623` — a file whose footer row-group count disagrees with `ROW_GROUP_META_KEY` contributes no metrics at all, so the aggregate counts cannot coincidentally match and yield a misaligned bound.
- `internal/core/src/segcore/storagev2translator/GroupCTMeta.h:148` — `InstallSkipMetrics` is the single install point and discards (rather than truncates) any field whose metric list length differs from `num_cells`, with `FindSkipMetric` bounds-checked.
- `internal/core/src/segcore/SkipIndexStats.cpp:56` — absent min/max, a one-sided VARCHAR bound, a physical-type mismatch, and an INT8/INT16 bound outside `T` all degrade to `NoneFieldChunkMetrics`, so every statistics defect fails open instead of pruning.
- `internal/core/src/exec/expression/Expr.h:1222` — for a skipped, unpinned chunk the default `ValidityView`'s `Subview` only advances the offset and leaves `data_ == nullptr`, so the following `ApplyValidData` is a no-op rather than a null dereference.
- `internal/core/src/exec/expression/TermExpr.cpp:1369` — `GetSkipIndexValues` and the scan's `arg_set_` drop out-of-range literals through the same `GetValueFromProtoWithOverflow<T>` rule, keeping prefetch pruning and row-level evaluation in agreement, and the `call_once` initialization closes the prefetch/Eval race.
- `internal/util/initcore/query_node.go:185` — the removed Go wrapper had no callers and the flag is still pushed into segcore at query-node init, so the `refreshable:"false"` change matches the actual startup-only wiring.

*(raised by wzy, yihao)*

czs007:

Verified:
- `internal/core/src/exec/expression/Expr.h:75` — `CanUseSkipFilter(is_nullable, null_rejecting)` is the single gate for "a skipped cell needs no validity", and `null_rejecting_` reaches a leaf only when its whole consumer chain up to the filter root folds UNKNOWN into FALSE; propagation stops at `NOT`.
- `internal/core/src/exec/expression/NullExpr.cpp:148` — `IS NULL` / `IS NOT NULL` compile to `PhyNullExpr`, which reaches data only via `ProcessChunksForValid*` and passes no skip function, so the new `AllNulls → prune` rule cannot invert them.
- `internal/core/src/segcore/storagev2translator/GroupCTMeta.h:144` — `InstallSkipMetrics` drops any field whose metric-list length `!= num_cells` instead of applying it partially; with the aggregate-count check at `ChunkedSegmentSealedImpl.cpp:2747` and the per-file guard at `:2623`, a metric cannot land on a cell it does not describe.
- `SkipIndexStats.cpp:56` — `!statistic->HasMinMax()` falls open to `NoneFieldChunkMetrics`, which is the correct handling for all-null and NaN-float row groups where reading `min()/max()` would have produced a false prune.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2639` — VARCHAR min/max are built while the Parquet `reader` is alive and deep-copied via `std::string(info->min_)`, so the use-after-free the PR description calls out is genuinely closed.
- `ChunkedColumnFilter.h:53` — the new third constructor argument defaults to `true`, so every existing `ColumnFilter` site keeps the old pin-for-validity behaviour; only `SegmentExpr::BindColumnFilter` opts into the relaxation.

## Low — non-blocking

**`internal/core/src/segcore/storagev2translator/GroupChunkTranslator.cpp:179` — forced one-row-group-per-cell packing also fragments column groups that can never be pruned**
`PlanGroupChunkPacking` sets `plan.force_one_row_group_per_cell = selection.skip_index_enabled`, so the startup flag alone decides packing: it does not depend on the group containing an `IsStatsSkippableType` field, and it is not re-checked after the `plan.skip_metrics_by_field.erase(it)` prune-worthiness loop below it. With `DEFAULT_ENABLE_PARQUET_STATS_SKIP_INDEX` flipped to `true` in this PR (`internal/core/src/common/Consts.h:122`), a vector-only group — or a pk/ts group, or one whose metric lists were all dropped — bypasses `ComputeRowGroupsPerCell` entirely. Failure: any Storage V2 sealed segment loaded with the now-default `common.parquetStatsSkipIndex.enabled: true`, in a collection whose Parquet row groups are far below `cellTargetSizeBytes` (4 MiB default) — e.g. a vector-only group written with ~64 KiB row groups → that group goes from roughly one cache cell to roughly one cell per row group (~64x), multiplying cache-slot metadata, pin/unpin calls and eviction bookkeeping, and a brute-force vector scan pins ~64x as many cells for the same bytes with zero pruning benefit.
Suggestion: derive the flag from the group's actual prune-worthiness rather than from the startup flag — set `force_one_row_group_per_cell` after the erase loop, from `!plan.skip_metrics_by_field.empty()` — so groups that provably cannot be pruned retain target-size packing (`GroupCTMeta.h:62-79`).
Note on history: this was raised once before on this PR, as a nitpick against `GroupChunkTranslator.cpp:177`. There is no author response to that round in the review record. The trade-off is now documented on the parameter (`GroupChunkTranslator.h:52-56`, "On by default, even when statistics are absent") and in the PR body's configuration-and-scope section, but that documents the cost rather than narrowing it; re-raising at low rather than nitpick because the flag is now the default and the cost falls on groups that can never benefit. Reasonable to close as an explicit accepted trade-off rather than a code change — but please make that call explicitly. *(raised by xiaocai2333, tinswzy)*

**`internal/core/src/segcore/storagev2translator/GroupChunkTranslator.h:93` — the 1 MiB per-cell floor in `cells_storage_bytes` now inflates client-visible `scanned_total_bytes`**
`cells_storage_bytes` floors every cell at 1 MiB (`std::max(meta_.chunk_memory_size_[cid], MIN_STORAGE_BYTES)`). That line is untouched by this diff, but it was written against target-size packing, where a cell was ~`cellTargetSizeBytes` (4 MiB) and the floor almost never bound; with `force_one_row_group_per_cell` on by default a cell is one row group, so for typical sub-MiB row groups the floor binds on every cell and the reported figure becomes `num_cells * 1 MiB` rather than real bytes. This is not merely a Prometheus counter — it flows through `OpContext::storage_usage.scanned_total_bytes` (`CacheSlot.h:428-430`) into `SearchResult::search_storage_cost_`, out over cgo, and into `Status.ExtraInfo["scanned_total_bytes"]`. Failure: a search or query over a Storage V2 sealed segment with storage-usage tracking enabled, on a collection with 64 KiB row groups (the default configuration after this PR) → `scanned_total_bytes` / `scanned_cold_bytes` returned to the client and exported as `ProxyScannedTotalMB` / `ProxyScannedRemoteMB` over-report by up to ~16x per cell, and the `cacheHitRatio` derived from them at `internal/proxy/util.go:2989-2990` is computed from those inflated totals, so cost dashboards and consumption accounting read high for exactly the workloads this PR targets.
Suggestion: report the true `chunk_memory_size_[cid]` on the accounting path and confine the 1 MiB floor to whatever it was originally protecting (or scale it with the configured cell size), so the client-visible figures and the `cache_cell_access_hit/miss_bytes_total` metrics (`CacheSlot.h:242-246`) track bytes actually read. The PR body already lists this as a known artifact and a follow-up; raising it here because defaulting the flag on makes it live for every Storage V2 deployment with storage-usage tracking enabled. *(surfaced during verification)*

internal/core/src/index/skipindex_stats/SkipIndexStats.h
```diff
@@ -618,23 +656,32 @@ class StringFieldChunkMetrics : public FieldChunkMetrics {
...
-            string_values.push_back(*sv);
-        }
-        if (!bloom_filter_) {
-            std::string_view min, max;
-            for (auto v : string_values) {
-                if (min.empty() || v < min) {
-                    min = v;
+            std::string_view min = *first;
+            std::string_view max = *first;
+            for (size_t i = 1; i < values.size(); ++i) {
```
sparknack, line 732: Non-blocking performance follow-up: `GetSkipIndexValues()` caches the converted literals, but this loop recomputes their min/max for every cell. These bounds depend only on the query, so the range-only skip checks still do O(cells * IN values) work.

Precompute the bounds once per expression and reuse them in both prefetch and evaluation. For the V2 footer-statistics path, which has no Bloom filter, the existing inclusive `CanSkipBinaryRange()` can consume those bounds; keep the full-list `CanSkipIn()` path for Bloom-filter checks.

czs007:

**Verified:**
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2631` — footer metrics are built only when the footer's `num_row_groups` matches the `ROW_GROUP_META` vector, and inside the reader lambda, with VARCHAR bounds copied into owning `std::string`, so no statistics view outlives the `FileMetaData`.
- `internal/core/src/segcore/storagev2translator/GroupCTMeta.h:148` — a metrics list whose length differs from the cell count is dropped, and the packing flag and metrics come from a single read of `ENABLE_PARQUET_STATS_SKIP_INDEX` per load, so metrics only exist when cells are 1:1 with row groups.
- `internal/core/src/index/skipindex_stats/SkipIndexStats.h:896` — INT8/INT16 footer bounds stored as physical INT32 that fall outside the narrow type's range degrade to `NoneFieldChunkMetrics` instead of wrapping; a failed typed-statistics cast degrades the same way.
- `internal/core/src/index/skipindex_stats/SkipIndexStats.cpp:117` — a VARCHAR footer carrying only one bound fails open, matching Arrow's `HasMinMax()` reporting true when either bound is present.
- `internal/core/src/exec/expression/Expr.h:75` — all four skip-function builders are gated by `CanUseSkipFilter = !is_nullable || null_rejecting`, `null_rejecting_` is only set through `ExprSet(..., true)` from consumers that exclude UNKNOWN rows, and the scan/take fast paths omit validity only when `SkippedValidityRequired()` is false.
- `internal/core/src/exec/expression/BinaryRangeExpr.cpp:1365` — the narrow-integer prefetch now clamps out-of-range literals and short-circuits the provably-empty range exactly as `PreCheckOverflow` does on the scan path, which resolves the earlier-round finding at line 1358.

## Low / non-blocking

These items do not require a change in this PR.

- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2742` — `PlanGroupChunkPacking` sets `force_one_row_group_per_cell` purely from the startup flag before it knows whether any field retained usable metrics. The early return on an empty `skip_metrics_by_field` (line 2745, e.g. a vector-only column group) and the erase loop (lines 2752-2768, footer/meta count mismatch or no row group with usable stats) can leave a group with zero prunable fields while the flag stays true, so `GroupChunkTranslator.cpp:179` still emits one cache cell per row group. Failure: `common.parquetStatsSkipIndex.enabled=true` (the new default) with a storage-v2 column group where no field keeps metrics → the translator builds `num_cells == total row groups` instead of target-size packing, so the segment carries more and smaller cache cells and pays more pin/load round-trips per scan while no cell in that group can ever be pruned. Suggestion: consider clearing the force flag when the plan ends with no prunable fields so the translator falls back to `ComputeRowGroupsPerCell` for that group; if the unconditional behavior is intentional, no change is needed, since the design doc (lines 42-45, 139-141) and the `milvus.yaml` comment already state it. Note on provenance: this is the same fragmentation concern raised in an earlier round against `GroupChunkTranslator.cpp:179`. The author responded by documenting the trade-off in the design doc and the config comment rather than changing packing. Documentation records the cost but does not remove it, so the item stays open here as a design choice, not a defect. Separately, the earlier-round claim that this inflates `scanned_total_bytes` (`GroupChunkTranslator.h:93`) is withdrawn: `CacheSlot.h:242-430` accounts the sum of pinned cells' storage bytes, and an un-pruned pass pins every cell, so the byte total is independent of packing. (raised by xiaocai2333)

internal/core/src/exec/expression/TermExpr.cpp
```diff
@@ -1193,16 +1196,14 @@ PhyTermFilterExpr::ExecVisitorImplForData(EvalCtx& context) {
...
-            const SkipIndex& skip_index, FieldId field_id, int64_t chunk_id) {
-            auto* elements = std::any_cast<std::vector<T>>(&cached_elements);
-            if (elements == nullptr) {
-                return false;
-            }
-            return skip_index.CanSkipInQuery<T>(
-                op_ctx, field_id, chunk_id, *elements);
+    SkipChunkFn skip_index_func;
+    if (CanUseSkipFilter(is_nullable_, null_rejecting_)) {
+        const auto& skip_query = GetSkipIndexQuery<T>();
```
sparknack, line 1202: Non-blocking optimization: skip preparing `PreparedInQuery` when `skip_view_.HasMetrics()` is false. On raw-data IN paths such as Growing, V1/V3, or V2 with no usable footer metrics, this still converts and caches the whole IN list and computes its bounds once per segment expression, although it cannot prune anything. The normal IN evaluator already prepares `arg_set_`; this is additional skip-only state.

Check metrics availability before `GetSkipIndexQuery<T>()` both here and in `PrefetchRawData<T>()`. Keep the skip callback empty and use ordinary field prefetch when metrics are absent; leave `arg_set_` unchanged.

PR #49157, opened 2026-04-20
enhance: record and expose DataNode task cost_time / cost_cpu_num via QueryTask

xiaocai2333:
issue: #49156

## Summary

- Record `cost_time` / `cost_cpu_num` on DataNode index build tasks and surface them through `QueryTask` response `Properties`.
- Reserve the unified `cost_time` / `cost_cpu_num` keys under `pkg/taskcommon` so future PRs can wire up the other `CreateTask`-dispatched task types.
- Introduce small `internal/datanode/taskcost` helper (`NowMs`, `CalcCostTimeMs`, `EstimateIndexBuildCPUNum`) for index build task cost accounting.

## Notes

This PR is producer-side only. DataCoord / proxy / metrics exporter do not consume these properties yet; downstream consumption should land in a follow-up linked from #49156.

## Test plan

- [x] `go test ./taskcommon/...` from `pkg/`
- [x] `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datanode/taskcost`
- [ ] `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datanode/index -run 'TestIndexTaskSchedulerRecordsIndexTaskCost|Test_statsTaskInfoSuite/Test_IndexTaskCostMethods'`
- [ ] Manual sanity: `QueryTask` returns `cost_time` / `cost_cpu_num` for DataNode index build task success and failure paths.

Issue #49156, opened 2026-04-20
[Enhancement]: Expose DataNode task execution cost_time and cost_cpu_num via QueryTask properties


xiaocai2333:
**Is there an existing issue for this?**

- [x] I have searched the existing issues

## Motivation

DataCoord schedules 8 kinds of tasks on DataNode through `CreateTask`:
`PreImport`, `Import`, `Compaction`, `Index`, `Stats`, `Analyze`, `ExternalCollection`, `CopySegment`.

Currently DataCoord has no uniform way to read back how long a task actually executed on the worker or how much compute it consumed. Without these signals, scheduler decisions (backpressure, balancing, capacity planning) have to rely on coarse slot counting only.

## Proposal

Record two cost signals per DataNode task and surface them through the existing `QueryTask` response `Properties`:

- `cost_time` — actual execution duration in milliseconds (`end_ms - start_ms`).
- `cost_cpu_num` — estimated number of concurrent workers used by the task, computed from each task's own execution strategy (pool capacity, task slot, parallel futures, etc.).

Approach:

1. Add two key constants and getter/setter helpers to `taskcommon.Properties`:
   - `cost_time`
   - `cost_cpu_num`
2. Add a tiny `internal/datanode/taskcost` helper package with `NowMs`, `CalcCostTimeMs`, `EstimateConcurrentWorkers`.
3. Instrument task lifecycles so each task records start/end and cost fields:
   - Import family (PreImport / Import / L0PreImport / L0Import / CopySegment)
   - Index / Stats / Analyze (via index `TaskScheduler.processTask` + `TaskManager`)
   - Compaction (via `compactor.executor` with `GetTaskCost` API)
   - ExternalCollection (via `external.ExternalCollectionManager` pool path)
4. In `DataNode.QueryTask`, for every task type branch, append `cost_time` / `cost_cpu_num` to the response properties (default `0` when task info is not found).

## Notes

- No breaking change to existing proto payloads. Everything rides on `QueryTaskResponse.Properties`.
- `cost_cpu_num` is a strategy-based estimate, not OS-level CPU profiling. Naming and logs make this explicit.
- Historical / absent tasks return `cost_time=0` / `cost_cpu_num=0` so DataCoord can safely parse.

PR #51443, opened 2026-07-16
enhance: record and expose DataNode task cost_time / cost_cpu_num via QueryTask

xiaocai2333:
issue: #49156

## Summary

- Record `cost_time` / `cost_cpu_num` on DataNode index build tasks and surface them through `QueryTask` response `Properties`.
- Reserve the unified `cost_time` / `cost_cpu_num` keys under `pkg/taskcommon` so future PRs can wire up the other `CreateTask`-dispatched task types.
- Introduce small `internal/datanode/taskcost` helper (`NowMs`, `ElapsedMs`, `EstimateIndexBuildCPUNum`) for index build task cost accounting.
- Final state/failReason and execution-end cost are written in a single `TaskManager` critical section, and `QueryTask` reads them from a single snapshot, so state and cost in one response are always consistent.

## Notes

This PR is producer-side only. DataCoord / proxy / metrics exporter do not consume these properties yet; downstream consumption should land in a follow-up linked from #49156.

`cost_cpu_num` is a coarse, statistics-only estimate for observability (vector index ≈ knowhere build pool size = NumCPU in cluster mode; scalar index = 1). It is not used for coordinator-side scheduling, so the standalone pool-ratio deviation (NumCPU × buildIndexThreadPoolRatio) and pool sharing across concurrent builds are intentionally not modeled.

Supersedes #49157 (that PR's head branch was rebased onto latest master and force-pushed, so GitHub no longer allows reopening it). Content is identical, rebased to a single commit on top of current master.

## Test plan

- [x] `go test ./taskcommon/...` from `pkg/`
- [x] `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datanode/taskcost`
- [ ] `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datanode/index -run 'TestIndexTaskSchedulerRecordsIndexTaskCost|Test_statsTaskInfoSuite/Test_IndexTaskCostMethods'`
- [ ] Manual sanity: `QueryTask` returns `cost_time` / `cost_cpu_num` for DataNode index build task success and failure paths.

### DIFF
--- a/internal/datanode/index/scheduler.go
+++ b/internal/datanode/index/scheduler.go
@@ -262,20 +262,22 @@ func (sched *TaskScheduler) processTask(t Task) {
 			if indexTask != nil {
 				endMs := taskcost.NowMs()
 				costTimeMs := taskcost.CalcCostTimeMs(startMs, endMs)
-				indexTask.manager.StoreIndexTaskExecutionEnd(indexTask.req.GetClusterID(), indexTask.req.GetBuildID(), endMs, costTimeMs)
+				// End bookkeeping and final state must land in one critical
+				// section, so a concurrent QueryTask never sees a final cost
+				// paired with an in-progress state.
+				indexTask.SetStateWithCost(getStateFromError(err), err.Error(), endMs, costTimeMs)
 				mlog.Warn(t.Ctx(), "process task failed", mlog.Err(err), mlog.Int64("costTimeMs", costTimeMs), mlog.Int64("costCPUNum", costCPUNum))
 			} else {
+				t.SetState(getStateFromError(err), err.Error())
 				mlog.Warn(t.Ctx(), "process task failed", mlog.Err(err))
 			}
-			t.SetState(getStateFromError(err), err.Error())
 			return
 		}
 	}
 	if indexTask != nil {
 		endMs := taskcost.NowMs()
 		costTimeMs := taskcost.CalcCostTimeMs(startMs, endMs)
-		indexTask.manager.StoreIndexTaskExecutionEnd(indexTask.req.GetClusterID(), indexTask.req.GetBuildID(), endMs, costTimeMs)
-		t.SetState(indexpb.JobState_JobStateFinished, "")
+		indexTask.SetStateWithCost(indexpb.JobState_JobStateFinished, "", endMs, costTimeMs)
 		mlog.Debug(t.Ctx(), "process task completed", mlog.String("task", t.Name()), mlog.Int64("costTimeMs", costTimeMs), mlog.Int64("costCPUNum", costCPUNum))
 	} else {
 		t.SetState(indexpb.JobState_JobStateFinished, "")
--- a/internal/datanode/index/scheduler.go
+++ b/internal/datanode/index/scheduler.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"go.uber.org/atomic"
 
+	"github.com/milvus-io/milvus/internal/datanode/taskcost"
 	"github.com/milvus-io/milvus/internal/storagev2"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/indexpb"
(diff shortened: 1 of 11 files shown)


czs007:

Wait — let me re-read the task. Two issues have verdict `keep`/`rewrite`, not `drop`. Let me produce the actual comment.

This adds cost/CPU accounting to the datanode index build path and threads final `cost_time` / `cost_cpu_num` into the `QueryTask` response as a producer-only contract, but the state/cost snapshot is assembled non-atomically and the CPU estimate diverges from the standalone build pool.

## Medium

- **`internal/datanode/services.go:827` — non-atomic state + cost snapshot.** On the write side, `StoreIndexTaskExecutionEnd` (writes `CostTimeMs` under `stateLock`, `task_manager.go:153`) and `SetState` → `StoreIndexTaskState` (writes `State`/`FailReason` under a *separate* `stateLock` acquisition, `task_manager.go:129`) run in two lock windows on both the success and failure branches. On the read side, `QueryTask` reads state via `queryIndexTask` → `ForeachIndexTaskInfo` (`index_services.go:448`) and cost via a second, independent `GetIndexTaskInfo` call (`services.go:827`). A concurrent `QueryTask` can therefore observe `task_state=InProgress` alongside a non-zero final `cost_time`/`cost_cpu_num` — an inconsistent contract for the downstream consumer landing in #49156. Suggestion: thread cost out of the single `queryIndexTask` `info.Clone()` (dropping the extra `GetIndexTaskInfo` call), and add one `TaskManager` method that writes execution-end plus final state/reason under a single lock. *(raised by yezh, wzy)*

## Low

- **`internal/datanode/taskcost/cost.go:49` — CPU estimate ignores the standalone pool ratio and concurrency.** For vector indexes, `EstimateIndexBuildCPUNum(true)` returns `int64(hardware.GetCPUNum())`, which matches cluster mode (ratio 1) but over-counts in standalone, where the build pool is `GetCPUNum() * BuildIndexThreadPoolRatio` (0.75 default), and it ignores that multiple builds share one pool. Acceptable for a producer-only PR. Suggestion: document the approximation and its limits (standalone 0.75, concurrency) in the PR body, and correct it when a consumer is wired up. *(raised by yezh, wzy)*

- **`internal/datanode/index/scheduler_test.go:238` — vector-index CPU branch untested end-to-end.** `TestIndexTaskSchedulerRecordsIndexTaskCost` builds its task with `IndexType=STL_SORT` (scalar, `IsVectorIndex()==false`) and asserts `CostCPUNum == 1`; the vector branch (`EstimateIndexBuildCPUNum(true)` → `GetCPUNum()`) is only covered by the unit test in `taskcost/cost_test.go`, never driven through `processTask`. Non-blocking. Suggestion: add a scheduler subtest with a vector `IndexType` asserting the recorded `CostCPUNum` reflects `GetCPUNum()`. *(raised by wzy)*

xiaocai2333:
@czs007 Thanks for the review — addressed in 055842e1c9:

- **Medium (`services.go:827` non-atomic state + cost snapshot)** — Fixed on both sides.
  - *Write side*: replaced the `StoreIndexTaskExecutionEnd` + `SetState` pair with `TaskManager.StoreIndexTaskExecutionEndWithState`, which records execution-end cost together with the final state/failReason in a **single** `stateLock` critical section (`indexBuildTask.SetStateWithCost` wraps it and keeps the finish-latency metrics).
  - *Read side*: `QueryTask`'s Index branch now derives the response payload and the state/reason/cost properties from **one** `GetIndexTaskInfo` clone, dropping the `queryIndexTask` walk plus the second lookup. This also removes the double-clone/double-lock flagged earlier — the direct map hit replaces the O(n) `ForeachIndexTaskInfo` scan.
  - A concurrent `QueryTask` can no longer observe `task_state=InProgress` paired with a non-zero final `cost_time`/`cost_cpu_num`. Covered by the updated `Test_IndexTaskCostMethods` and `TestQueryTask/query index task with cost`.

- **Low (vector branch untested through `processTask`)** — Added subtest `vector index records build pool cpu num` in `TestIndexTaskSchedulerRecordsIndexTaskCost`: drives `processTask` with a real `*indexBuildTask` (`IsVectorIndex` mocked to true) and asserts the recorded `CostCPUNum == hardware.GetCPUNum()`.

- **Low (standalone pool ratio in `EstimateIndexBuildCPUNum`)** — Won't fix, documented instead: these two fields are task-level statistics for observability only and are **not** consumed for coordinator-side scheduling, so the standalone `NumCPU * buildIndexThreadPoolRatio` deviation (and pool sharing across concurrent builds) is acceptable. This is now stated in the `EstimateIndexBuildCPUNum` doc comment and in the PR body.

czs007:
Here's the PR review comment:

---

Adversarial review surfaced three candidate issues; one was dropped on verification, leaving two minor suggestions — one correctness nit around wall-clock timing and one logging cleanup.

## Low

**`internal/datanode/taskcost/cost.go:29` — Elapsed time is measured with the wall clock, so a clock step back can under-report or zero out a task's cost**

`NowMs()` returns the wall-clock `time.Now().UnixMilli()` (cost.go:25-27), and `processTask` samples it before and after execution to derive `CostTimeMs` (scheduler.go:254, 277-279). If the wall clock steps backward during a build (e.g. an NTP correction), the duration is under-reported; and when `endMs < startMs`, `CalcCostTimeMs` returns `0` (cost.go:29-34), which erases the cost of exactly the long-running tasks that are most worth diagnosing.

Suggestion: record `start := time.Now()` at the top of `processTask` and compute `CostTimeMs` with `time.Since(start).Milliseconds()` — Go's `time.Since` uses the monotonic clock and is immune to wall-clock steps. The externally exposed `ExecStartMs`/`ExecEndMs` can keep their wall-clock values; the two are not in conflict.

*(raised by yezh, wzy)*

## Nitpick

**`internal/datanode/index/scheduler.go:258` — Non-index tasks log a `costCPUNum` that is always 0**

The top-level debug line `mlog.Debug(t.Ctx(), "process task", mlog.String("task", t.Name()), mlog.Int64("costCPUNum", costCPUNum))` sits outside the type assertion and therefore fires for every task type. `costCPUNum` is only assigned inside the `if ibt, ok := t.(*indexBuildTask); ok` branch (scheduler.go:251-253), so it stays at its zero value otherwise. This scheduler also runs `analyzeTask` and `statsTask` — all three implement `Task` (task_analyze.go:80, task_index.go:149, task_stats.go:146) and are enqueued through `taskScheduler.TaskQueue.Enqueue` in `index_services.go` — so those two task types emit a constant, information-free `costCPUNum=0` on every run.

Suggestion: move the `costCPUNum`-bearing log into the `indexTask != nil` branch, consistent with the completion logs at lines 269 and 281, and let non-index tasks log the original field set.

*(raised by wzy)*

xiaocai2333:
@czs007 Second round addressed in 7b1208418d:

- **Low (`cost.go:29` wall-clock elapsed time)** — Fixed as suggested. `CalcCostTimeMs(startMs, endMs)` is removed; `processTask` now records `execStart := time.Now()` and derives `CostTimeMs` via the new `taskcost.ElapsedMs(start)` (`time.Since`, monotonic clock), so a wall-clock step can no longer under-report or zero out a long task's cost. `ExecStartMs`/`ExecEndMs` keep their wall-clock values for external exposure, and the doc comments on `NowMs`/`ElapsedMs` now state the contract (never subtract two `NowMs` readings).

- **Nitpick (`scheduler.go:258` non-index tasks log `costCPUNum=0`)** — Fixed. The `costCPUNum`-bearing "process task" debug line moved into the `*indexBuildTask` branch; analyze/stats tasks log the original field set again, consistent with the completion/failure logs.

czs007:

liliu-z (commented):
/lgtm
/approve

PR #51444, opened 2026-07-16
feat: add SQL query parser

zhagnlu:


zhengbuqian:
Ran the parser against a set of PG-valid statements on this branch. Syntactically it is a PG subset (libpg_query), but semantically several PG-valid statements are accepted and silently return different results. Grouping the findings:

**1. Please reject these instead of silently dropping / accepting them (list is not exhaustive)**

```sql
-- HAVING: stmt.HavingClause is never read; groups are returned unfiltered
SELECT data->>'category' AS c, count(*) FROM t GROUP BY c HAVING count(*) > 1;

-- DISTINCT: DistinctClause dropped, duplicates returned
SELECT DISTINCT name FROM t;

-- OFFSET: LimitOffset dropped, returns rows 1-10 instead of 6-15
SELECT id FROM t ORDER BY id LIMIT 10 OFFSET 5;

-- Locking clause dropped
SELECT id FROM t FOR UPDATE;

-- Type casts dropped (where_converter.go:85, :280): becomes data["age"] > 30 / x == 5
SELECT id FROM t WHERE (data->>'age')::int > 30;
SELECT id FROM t WHERE x = 5::text;

-- Ungrouped column next to an aggregate: PG errors, we accept
SELECT name, count(*) FROM t;

-- Vector query: ORDER BY is ignored entirely (task_sql_query.go:558).
-- DESC returns the same rows in the same order as ASC; ORDER BY id is also ignored.
-- Order is decided by the index metric (see 4), so reject explicit ASC/DESC and
-- any ORDER BY on something other than the distance expression.
SELECT id, embedding <-> $1 AS d FROM t ORDER BY d DESC LIMIT 5;
SELECT id, embedding <-> $1 AS d FROM t ORDER BY id LIMIT 5;

-- Vector query without LIMIT: PG returns all rows, we silently use topK=10 (extract.go:584)
SELECT id, embedding <-> $1 AS d FROM t;

-- Table-qualified column: json_path.go:88 takes Fields[0], so t.x becomes field "t"
SELECT id FROM t WHERE t.x = 1;
```

Each of these should fail at extraction with a clear "not supported" error. The design doc says "Unsupported syntax should fail early with a clear error rather than falling back to an incomplete execution"; these are the cases where that currently doesn't hold.

**2. Path access syntax — needs a sync meeting**

Milvus already has its own path syntax in the filter language: `data["a"]["b"]` / `data["arr"][0]` for JSON, `arr[0]` for arrays (0-based), `struct_field[sub]` / `struct_field[0][sub]` for struct sub-columns, `$meta["k"]` for the dynamic field. PG uses `data->'a'->>'b'` / `data#>>'{a,b}'` / PG14 `data['a']['b']` for jsonb, `arr[1]` (1-based) for arrays, and `(s).sub` for composites.

This PR maps `->`/`->>` onto `data["a"]["b"]` and drops the jsonb-vs-text distinction. That already produces observable differences: `WHERE data->>'age' > 30` is a PG type error but is accepted here; `WHERE data->'info'->>'level' = '0'` matches in PG (text compare) but goes through Milvus's typed JSON compare, which does not match a numeric 0 against `"0"` (the e2e test at `tests/e2e/sql_query_test.go:240` encodes the PG expectation with numeric `level`).

If we align the SQL surface with PG we diverge from the existing Milvus expression language; if we keep the Milvus `[]` syntax inside SQL we are no longer a PG subset (and array indexing base differs). Struct sub-columns have no PG equivalent at all. I'd like to schedule a meeting to decide this before the JSON path handling is finalized — it affects the WHERE converter, the planner, and the JSONBench queries.

**3. Expression evaluation — please reuse the filter engine's semantics**

WHERE arithmetic goes through the existing filter engine, which already matches PG on the points that matter: integer division for integer fields, error on division by zero. The SELECT-list computed expressions go through the new `project_executor.go` (and `ComputeProjectNode.cpp`), which evaluate everything in double, return 0 on division by zero (`project_executor.go:460`), return 0 for unrecognized `extract()` fields (`:411`), treat `to_timestamp` as identity, and pin the timezone to UTC. So `SELECT x / 2` gives 3.5 while `WHERE x / 2 = 3` is integer division — two different semantics inside one statement.

Suggest: SELECT-list evaluation should follow the same rules as the filter engine (type-preserving arithmetic, error instead of 0 on `/0`, error on unsupported `extract` fields). Where Milvus deliberately differs from PG (typed JSON comparison, no implicit string→int coercion) we keep the Milvus behavior; where it already agrees with PG we shouldn't introduce a third behavior.

**4. Vector distance: drop the pgvector operators, use one metric-agnostic function**

`<->`, `<=>`, `<#>`, `<~>`, `<%>` (and `l2_distance` / `cosine_distance` / `ip_distance` / ...) each encode a metric type. pgvector needs that because a table can carry several indexes with different opclasses, or none at all. Milvus doesn't: the metric is fixed when the index is built, and at search time it is optional-but-must-match — `plan_c.cpp:64` overwrites the plan's metric with the index's, and `internal/util/segcore/plan.go:108` rejects a request whose metric differs (`metric type not match`). So the operator → `MetricTypeKey` mapping in `task_sql_query.go:912` can only be redundant or wrong: an index built with COSINE plus a query written with `<->` fails outright.

Proposal: replace the five operators and five functions with a single Milvus function that takes the vector field and the query vector, e.g.

```sql
SELECT id, milvus_ann_distance(embedding, $1) AS d
FROM t
WHERE category = 'tech'
ORDER BY d
LIMIT 10;
```

- The metric comes from the loaded index; nothing is sent in `MetricTypeKey`.
- `ORDER BY` on the function (or its alias) + `LIMIT` is the KNN trigger, same model as pgvector; `LIMIT` is required and becomes topK; `WHERE` is the filter; `nprobe`/`ef`/`radius` keep coming from `searchParams`.
- No explicit `ASC`/`DESC` — nearest-first is decided by the index metric (`PositivelyRelated`). The output column is Milvus's raw score, same as the Search API, so IP/COSINE are "higher is closer" and L2/HAMMING/JACCARD are "lower is closer"; document that rather than negating scores the way pgvector's `<#>` does.
- Exactly one call per statement for now; multi-vector (hybrid) is out of scope.
- Still a PG subset syntactically: a function call in `ORDER BY` is ordinary PG grammar; the function simply doesn't exist in PG, same status as pgvector's `l2_distance()`.

This is a small change to the current code: `vectorDistFuncs` in `extract.go:39` collapses to one entry, `vectorDistOps` goes away, and `sqlComponentsToSearchRequest` stops emitting `MetricTypeKey`. It also removes the `<#>` sign mismatch and the non-pgvector `ip_distance` name from the surface entirely.

zycgit:
@zhagnlu 
Here’s a tool that lets you interact with Milvus using SQL through a JDBC driver. It supports common statements such as INSERT, UPDATE, DELETE, and SELECT, along with data import and many administrative operations. It looks like it could already serve as an alternative to the SDK.

https://www.dbvisitor.net/en/docs/features/milvus/about
https://github.com/zycgit/dbvisitor/tree/dev/dbvisitor-adapter/jdbc-milvus

PR #51377, opened 2026-07-14
fix: CAS the schema-bump in-place manifest commit to stop dropping a concurrent index (#51376)

MrPresent-Han:
issue: #51376

## What this fixes

The in-place partial-materialization path of `bump_schema_version_compaction` commits onto a live segment's manifest, and DataCoord adopted the result on a version-newer check only — it never verified the result was built on the segment's *current* manifest. A bump built on a plan-pinned base could therefore be adopted after a concurrent stats/index task advanced the manifest pointer, silently dropping that index (correct version number, wrong content lineage). The loss is permanent and non-self-healing, because the segment meta stats placeholder stops the stats inspector from re-triggering.

## Fix — application-layer adoption CAS (mirrors the stats path)

- The schema-bump result carries the plan-pinned base manifest (`CompactionSegment.base_manifest`, like `StatsResult.base_manifest`); the DataNode fills it with the manifest it materialized on.
- DataCoord adopts the in-place result only if `result.base_manifest == segment.ManifestPath` (the current pointer); a missing or stale base is rejected (`ErrIllegalCompactionPlan`) so the task re-triggers and rebuilds on the current manifest instead of overwriting the concurrent commit. This mirrors the stats path's `errStatsResultStale`.
- The storage commit stays on `OverwriteResolver` (unchanged). Because the CAS is at the adoption layer (against the meta pointer, not storage-latest), a result lost after a DataNode crash self-heals on retry: the pointer stays put, so the retry re-pins the same base and is adopted — no wedge.
- The bump policy is restricted to internal collections (matching every sibling compaction policy), so a rejected attempt's orphan is reclaimed by internal dropped-segment GC and bump never churns on external collections.

## Verification
Both interleavings converge with no data loss: stats-first → bump's base no longer matches the advanced pointer → reject + retrigger on the current manifest; bump-first → stats' existing base-check rejects and retries. Added meta-level adoption tests (matching/stale base) and updated the in-place adoption tests to carry the base.

Issue #51376, opened 2026-07-14
schema-bump in-place manifest commit can silently drop a concurrently-built index


MrPresent-Han:
**Is there an existing issue for this?**
- Searched, none found.

**Environment**
- Milvus master (StorageV3 / manifest path).

**Describe the bug**

The partial-materialization path of `bump_schema_version_compaction` commits in-place onto a live segment's manifest using `OverwriteResolver`, and the DataCoord acceptance only checks that the result manifest version is *newer* — it does not check that the result was *based on* the current manifest. There is no conflict detection at either the storage or the meta layer.

When a stats/index task (text / JSON / BM25 index, or bloom filter) commits into the same live segment's manifest concurrently:

1. The bump DataNode pins its base manifest at plan-build time (say `v10`).
2. A stats task commits the index into the manifest → `v11`, and DataCoord adopts it.
3. The bump, still building on the pinned `v10`, commits via `OverwriteResolver` which rebuilds from the stale `v10` (it ignores latest) → `v12`, whose content is `v10 + bump columns` and **omits the `v11` index**.
4. DataCoord accepts `v12` because `v12 > v11` (version-newer only).

Result: correct version number, wrong content lineage. The concurrently-built index silently disappears from the manifest that query nodes read. The loss is **permanent and non-self-healing** — the segment meta still carries the stats placeholder, so the stats inspector never retriggers.

Note the same codebase's stats path is safe only because it adds an application-layer base-lineage check at acceptance (`internal/datacoord/task_stats.go`, `errStatsResultStale`); the bump path copies neither that check nor the storage-layer `FailResolver`.

**Impact**

The concurrently-built text/JSON/BM25 index (or bloom filter) is lost for query nodes, degrading or breaking queries that depend on it, with no automatic recovery.

**Expected behavior**

The in-place bump commit must detect that the manifest advanced past its pinned base and retry on the current manifest instead of overwriting it.

PR #51450, opened 2026-07-16
fix: CAS the schema-bump in-place manifest commit to stop dropping a concurrent index (#51376)

MrPresent-Han:
issue: #51376
pr: #51377

## What this fixes

The in-place partial-materialization path of `bump_schema_version_compaction` commits onto a live segment's manifest, and DataCoord adopted the result on a version-newer check only — it never verified the result was built on the segment's *current* manifest. A bump built on a plan-pinned base could therefore be adopted after a concurrent stats/index task advanced the manifest pointer, silently dropping that index (correct version number, wrong content lineage). The loss is permanent and non-self-healing, because the segment meta stats placeholder stops the stats inspector from re-triggering.

## Fix — application-layer adoption CAS (mirrors the stats path)

- The schema-bump result carries the plan-pinned base manifest (`CompactionSegment.base_manifest`, like `StatsResult.base_manifest`); the DataNode fills it with the manifest it materialized on.
- DataCoord adopts the in-place result only if `result.base_manifest == segment.ManifestPath` (the current pointer); a missing or stale base is rejected (`ErrIllegalCompactionPlan`) so the task re-triggers and rebuilds on the current manifest instead of overwriting the concurrent commit. This mirrors the stats path's `errStatsResultStale`.
- The storage commit stays on `OverwriteResolver` (unchanged). Because the CAS is at the adoption layer (against the meta pointer, not storage-latest), a result lost after a DataNode crash self-heals on retry: the pointer stays put, so the retry re-pins the same base and is adopted — no wedge.
- The bump policy is restricted to internal collections (matching every sibling compaction policy), so a rejected attempt's orphan is reclaimed by internal dropped-segment GC and bump never churns on external collections.

## Verification
Both interleavings converge with no data loss: stats-first → bump's base no longer matches the advanced pointer → reject + retrigger on the current manifest; bump-first → stats' existing base-check rejects and retries. Added meta-level adoption tests (matching/stale base) and updated the in-place adoption tests to carry the base.


liliu-z (commented):
/lgtm
/approve

Issue #49195, opened 2026-04-21
[Enhancement]: Introduce unified gRPC metrics and gRPC logging across all Milvus components


chyezh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Introduce a unified, configurable gRPC observability layer (metrics + structured logging) applied to **every** Milvus component's gRPC server and client (Proxy, RootCoord / QueryCoord / DataCoord / MixCoord, QueryNode, DataNode, StreamingNode), covering both unary and stream RPCs.

**Proposed scope:**

1. **Unified gRPC interceptor framework**
   - A shared package `pkg/util/grpcutil/interceptor` (or similar) that provides:
     - `UnaryServerInterceptor` / `StreamServerInterceptor`
     - `UnaryClientInterceptor` / `StreamClientInterceptor`
   - All components register this interceptor chain in their `service.go`, replacing scattered ad-hoc implementations.

2. **gRPC metrics** (Prometheus)
   - Per-method RPC count (labeled by `component`, `method`, `grpc_code`, `role`: server/client, `stream_kind`: unary/server_stream/client_stream/bidi).
   - Per-method RPC latency histogram (with configurable buckets).
   - In-flight RPC gauge.
   - Streaming-specific metrics:
     - Stream duration histogram
     - Messages sent / received per stream
     - Stream keepalive / idle stats
   - Payload size metrics (reuse `metrics.NewGRPCSizeStatsHandler`, extend to all components).

3. **gRPC structured logging**
   - Start-of-RPC log (optional, sampled): method, peer, deadline, cluster/server IDs.
   - End-of-RPC log: method, duration, grpc code, error, bytes in/out, streaming message counts.
   - Log level and sampling controlled by config:
     - `grpc.log.enabled` (bool)
     - `grpc.log.level` (info/debug)
     - `grpc.log.sampleRate` (1.0 = all, 0.01 = 1%)
     - `grpc.log.slowThreshold` (always log slow RPCs regardless of sample rate)
     - Per-method allow/deny list for noisy high-QPS methods (e.g. `GetMetrics`, time-tick).
   - Reuse Milvus log context (trace ID, request ID) propagated via context.

4. **Config schema** (paramtable)
   - New section `grpc.observability.*` with metrics enable/disable, histogram buckets, logging controls.
   - All controls runtime-hot-reloadable (via config watcher) where feasible.

5. **Integration with existing infrastructure**
[...]

PR #51452, opened 2026-07-16
enhance: add unified gRPC metrics and logging interceptors

chyezh:
issue: #49195

- gRPC observability: add unified client and server interceptors for Prometheus metrics and configurable access logging
- component wiring: install the observability interceptors across Milvus gRPC servers and internal clients
- configuration: add hot-reloadable gRPC client/server log level and method filters with literal and regex matching
- agent docs: add the gRPC observability guide and link it from observability references


chyezh:
Automated review, not independently verified (Team II review)

Overall conclusion: I would not merge this yet. The review found one high-risk behavioral regression and one test coverage gap. The reviewer did not modify files or run builds/tests. GitHub API access was unavailable in the reviewer environment, so this is based on the local checkout and local PR diff only.

## High

`internal/distributed/proxy/service.go`

The external Proxy gRPC unary chain is currently ordered as:

```text
GrpcAuthInterceptor -> UnaryServerHookInterceptor -> PrivilegeInterceptor
```

The previous order was:

```text
GrpcAuthInterceptor -> PrivilegeInterceptor -> UnaryServerHookInterceptor
```

Failure scenario: when authorization is enabled and hook mock/before behavior is configured, an external user without the required RBAC permission can enter the hook before the privilege check. More importantly, `hoo.Mock(...)` in `internal/proxy/hook_interceptor.go` can directly `return mockResp, err` without invoking the later `PrivilegeInterceptor`, which can bypass the original authorization path. This ordering change is not required for observability and is a security/authorization semantic regression.

Suggested fix: preserve the original security order by moving `proxy.UnaryServerInterceptor(proxy.PrivilegeInterceptor)` back before `proxy.UnaryServerHookInterceptor()`. `mlog` / observability can stay after auth and before privilege if we want to record authorization failures, but the hook must not run before RBAC. Please also add a focused external proxy chain ordering test, at least covering that a hook mock cannot bypass privilege.

## Medium

`pkg/util/interceptor/observability_interceptor_test.go`

The access-log path tests only assert that the handler/invoker is called and the return value is normal. They do not assert the metrics sample, the effective log behavior after level/filter changes, or the real Proxy interceptor chain ordering.

Failure scenario: the hook/RBAC ordering regression above is exactly the kind of issue the current tests miss. In addition, the regex allowlist, dynamic config, and mlog level gating tests mostly cover parsing/calling behavior; they do not prove whether `grpc.server.call` / `grpc.client.call` are actually emitted or suppressed according to config.

Suggested focused tests:

- Build a unary interceptor chain and verify auth / privilege / hook / observability ordering.
- Use an observable logger sink to verify exact match / regex / invalid regex / level gating for access logs.
- Use `prometheus.Registry.Gather()` to verify expected metrics labels and status code.

## Notes

- Server metrics labels are mainly `grpc_type` / `grpc_service` / `grpc_method` / `grpc_code` / `node_id`. Method and service names come from proto definitions, so global cardinality is bounded; no obvious high-cardinality label was found.
- Client metrics currently do not include `node_id` / `dstServerID`, which avoids extra cardinality but reduces client-side RPC debugging by Milvus node. This is not blocking.
- Module dependencies do not show direct build breakage evidence. However, root uses `go-grpc-middleware/v2 v2.3.2`, while `pkg` / `tests/go_client` indirectly record `v2.1.0`. Please confirm this is stable after `go mod tidy` for each module.

Full Team II report: https://czs007.com/cluster/r/R39cfc51bb5f7?exp=1784632946364&sig=cbc8062acc2fdf8ae96a99a90f5837851f4476b36db443a91154f9a92d59acfd

pkg/util/interceptor/observability_interceptor.go
```diff
@@ -0,0 +1,355 @@
...
+	)
+}
+
+// NewObservabilityServerUnaryInterceptor records Prometheus metrics and — when
+// the full method is in the server log allowlist — emits a structured access
+// log. When no methods are allow-listed (the default), this interceptor is
+// equivalent to the bare metrics interceptor plus a single branch.
+func NewObservabilityServerUnaryInterceptor() grpc.UnaryServerInterceptor {
+	logCfg := getServerDynamicLogConfig()
+	metricsIntercept := metrics.GRPCServerMetric.UnaryServerInterceptor(
```
liliu-z, line 522: Blocking: grpcprom only sees the Go/gRPC error here. A lot of Milvus handlers return a failed `commonpb.Status` with a nil error, so both the metric and access log show `OK` while the request actually failed (and may be retried later). Please keep `grpc_code` as the transport status, but also report the Milvus response outcome and add a `merr.Status(err), nil` test.

internal/distributed/proxy/service.go
```diff
@@ -292,20 +292,7 @@ func (s *Server) startExternalGrpc(errChan chan error) {
...
-			proxy.GrpcAuthInterceptor(proxy.AuthenticationInterceptor),
-			proxy.UnaryServerInterceptor(proxy.PrivilegeInterceptor),
-			proxy.UnaryServerHookInterceptor(),
-			mlog.UnaryServerInterceptor(typeutil.ProxyRole),
-			proxy.RateLimitInterceptor(limiter),
-			accesslog.UnaryUpdateAccessInfoInterceptor,
-			proxy.TraceLogInterceptor,
-			connection.KeepActiveInterceptor,
-		))
+		unaryServerOption = grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(buildExternalProxyUnaryInterceptors(limiter)...))
```
liliu-z, line 296: Blocking for the linked CDC case: this external server only installs unary observability. `CreateReplicateStream` and `DumpMessages` run on this server, and the CDC Go client has no stream interceptor either, so the original RST/idle-timeout path is invisible on both ends. Please wire both sides, or narrow the PR/issue scope.

pkg/util/interceptor/observability_interceptor.go
```diff
@@ -0,0 +1,355 @@
...
+
+		start := time.Now()
+		var cs grpc.ClientStream
+		var callErr error
+		wrapped := func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+			cs, callErr = streamer(ctx, desc, cc, method, opts...)
+			return cs, callErr
+		}
+		cs, callErr = metricsIntercept(ctx, desc, cc, method, wrapped, opts...)
+		emitClientAccessLog(ctx, lvl, method, callErr, time.Since(start))
```
liliu-z: Blocking for stream diagnostics: this logs when the stream handle is created, not when the stream ends. RST, `Unavailable`, timeout, and later Send/Recv failures will still look like a successful short call. Please emit the client log on terminal stream completion, e.g. with a `ClientStream` wrapper and `sync.Once`.

internal/distributed/proxy/service.go
```diff
@@ -394,6 +381,24 @@ func (s *Server) startExternalGrpc(errChan chan error) {
 	mlog.Info(context.TODO(), "Proxy external grpc server exited")
 }
 
+func buildExternalProxyUnaryInterceptors(limiter types.Limiter) []grpc.UnaryServerInterceptor {
+	return []grpc.UnaryServerInterceptor{
+		streaming.ForwardLegacyProxyUnaryServerInterceptor(),
```
liliu-z, line 390: Blocking for rolling upgrades: this interceptor can short-circuit after forwarding, so the observability interceptor below never sees the request. The forwarding client also does not install the new client interceptor. Please cover both ingress and egress forwarding paths, otherwise upgrade-time retry traffic remains a blind spot.

pkg/util/interceptor/observability_interceptor.go
chyezh: No additional custom ClientStream wrapper will be added. The pending implementation delegates stream lifecycle reporting to go-grpc-middleware v2.3.2. It does not emit finished call when the stream handle is created; it emits finished call when stream creation fails or RecvMsg reaches EOF or another terminal error. The CDC path continuously calls Recv, so RST, Unavailable, and idle-timeout termination are observed there. The upstream middleware does not use sync.Once, but generated clients and the CDC loop stop reading after a terminal Recv result, so adding a second lifecycle wrapper would duplicate the middleware behavior. A regression test verifies that creation emits no finish log and terminal RecvMsg does.

internal/distributed/proxy/service.go
chyezh: Addressed locally and this will be included in the next PR update. The external Proxy now installs the unified stream server observability interceptor, covering CreateReplicateStream and DumpMessages. The CDC Milvus client also appends the unified client stream interceptor after its authority and TLS dial options, so the replication stream is observable on both ends. I added regression coverage for the external server stream metric and the CDC client dial-option wiring.

internal/distributed/proxy/service.go
chyezh: Keeping this unchanged. The legacy Proxy rolling-upgrade forwarding ingress and egress paths are outside the scope of this update, so this PR will not add observability wiring to that forwarding path.

Issue #51299, opened 2026-07-13
[Bug]: [Benchmark][Cluster] 100M DISKANN load_collection fails with QueryNode disk resource exhausted


wangting0128:
- [x] I have searched the existing issues

### Environment

- Milvus version: `3.0-20260711-56460826` (failed) / `2.6-20260710-65db4eaa` (passed) / `master-20260627-2310c93b` (passed)
- Deployment mode: cluster, streaming mode
- MQ type: pulsarv3
- SDK version: pymilvus `3.1.0rc63`
- OS: Linux (4AM cluster)
- CPU/Memory: queryNode 8C/32Gi, dataNode 8C/16Gi requests and 12C/32Gi limits, queryNode ephemeral-storage limit 100Gi
- Deploy config: `cluster_2c8m`, queryNode 1 replica, dataNode 2 replicas, indexNode disabled
- Extra config: `common.storage.useLoonFFI=true`, `dataNode.storage.format=vortex`, `dataCoord.compaction.storageVersion.enabled=true`, `dataCoord.compaction.storageFormat.enabled=true`

### Current Behavior

100M DISKANN benchmark failed during `load_collection`. QueryNode rejected segment loading because disk resource prediction exceeded the disk threshold, and the collection finally remained unloaded.

```text
[setup_method] Test case: test_concurrent_locust_100m_diskann_ddl_dql_filter_cluster, Test run_id: 2026071192527121
```

**Error examples**:

```text
[2026-07-11 20:31:12,535 - ERROR - fouram]: (api_response) : [MilvusClient.load_collection] <MilvusException: (code=101, message=segment request resource failed[resourceType=Disk]: collection not loaded[collection=467609666297792103])>, [requestId: 3f1830b74a4241d2ae7f0bdf7422036d]
```

**Error breakdown:**

| Error Type | Count | Note |
|-----------|-------|------|
| `segment request resource failed[resourceType=Disk]` | 1 | unexpected, load_collection failed |
| flush RateLimit | 0 | not involved |

### Expected Behavior

The same 100M DISKANN benchmark should load successfully under the configured 100Gi QueryNode disk limit, or Milvus should avoid planning stale/compacting segments that make the load plan exceed the logical collection size.

Recent successful comparisons:

| Run | Image | Load time | QueryNode disk usage |
|-----|-------|-----------|----------------------|
[...]

PR #51453, opened 2026-07-16
fix: skip raw vector fielddata on storage-v3 load when index has raw data

xiaofan-luan:
issue: #51299

## Problem

On the 100M / 128-dim DiskANN benchmark (`test_concurrent_locust_100m_diskann_ddl_dql_filter_cluster`, 1 replica, 100Gi querynode ephemeral-storage), `load_collection` fails on storage-v3 with:

```
segment request resource failed[resourceType=Disk]
```

Root cause: **storage-v3 (loon/vortex) segments load the raw vector column resident *on top of* the vector index, even when the index already carries the raw data.**

The v2 binlog load path skips the raw fielddata when the index has raw data (`ChunkedSegmentSealedImpl`: *"skip fielddata because index has raw data"*). The v3 manifest path in `SegmentLoadInfo::ComputeDiffColumnGroups` instead marked the column **lazy**, but `LoadColumnGroup(eager_load=false)` still materialized a `ProxyChunkColumn` and set `field_data_ready` — so the raw column stayed resident on disk anyway, ~doubling the per-segment footprint.

Evidence (from the failed run's logs + a segcore reproduction):

| | per-node disk | `num_chunk_data(vec)` | `"index has raw data"` skip |
|---|---|---|---|
| **2.6 (v2)** | ~71 GiB predicted (passes) | 0 | logged 99×/node |
| **3.0 (v3)** | ~101 GiB predicted (over 100Gi → fail) | 1 | logged 0×/node |

The extra ~47 GiB ≈ the raw vectors (100M × 128 × 4B).

## Fix

In `SegmentLoadInfo::ComputeDiffColumnGroups`, when a **user field's index carries raw data** and `prefer_field_data_when_index_has_raw_data == false`, drop the raw column from the load set entirely (and free any stale resident copy on the no-raw-index → raw-index reopen transition) instead of lazy-loading it. This mirrors the v2 binlog path. `prefer_field_data` remains the explicit opt-in to keep both resident.

## Test

Adds `internal/core/unittest/test_v2v3_raw_residency.cpp`: builds a V2 (binlog) and a V3 (loon manifest) float-vector segment with a **real IVF_FLAT index**, loads each through the **production load path** (`SegmentLoadInfo → SetLoadInfo → Load`, not a force-load helper), and asserts:
- `num_chunk_data(vec) == 0` and `HasFieldData(vec) == false` (raw column skipped) — for both v2 and v3;
- **correctness gate**: `get_vector` still returns the correct vectors (reconstructed from the index) after the raw column is skipped.

Before the fix this test reproduces the bug (v3 = 1 chunk resident); after the fix v3 should match v2 (0).

## Status

> ⚠️ Draft — the remote unittest build/run to confirm the retrieve-correctness gate + the segcore regression sweep was still in progress when this was pushed. Relying on CI to confirm; will flip to ready once green.

### DIFF
--- a/internal/core/src/segcore/SegmentLoadInfo.cpp
+++ b/internal/core/src/segcore/SegmentLoadInfo.cpp
@@ -471,6 +471,12 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
     auto prefer_field_data =
         SegcoreConfig::default_config()
             .get_prefer_field_data_when_index_has_raw_data();
+    // The primary key column is never skipped, even when its index carries raw
+    // data: sorted segments navigate rows through the pk column
+    // (num_rows_until_chunk / get_chunk_by_offset), and a pk scalar index is not
+    // registered in scalar_indexings for expression index-reads, so the pk raw
+    // column must stay resident.
+    auto pk_field_id = new_info.schema_->get_primary_field_id();
     auto cur_column_group = GetColumnGroups();
     auto new_column_group = new_info.GetColumnGroups();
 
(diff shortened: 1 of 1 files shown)


xiaofan-luan:
Superseded by #51541 — recreated on a slash-free branch (`fix-v3-diskann-raw-vector-double-load`). Same fix plus the primary-key exclusion.

Issue #51446, opened 2026-07-16
[Bug]: Skip unsupported snapshot/backfill/external-refresh DDLs in primary-secondary replication


chyezh:
## Environment

- **Milvus Version**: current master/topic branch around primary-secondary replication support
- **Deployment Mode**: cluster with primary-secondary replication
- **MQ**: any WAL-backed deployment
- **SDK**: N/A
- **OS**: N/A

### Option B: Steps

1. Deploy a primary-secondary Milvus topology.
2. Execute one of the following DDL/control operations on the primary:
   - `CreateSnapshot`
   - `RestoreSnapshot`
   - `DropSnapshot`
   - `DropSnapshotsByCollection`
   - `BatchUpdateManifest` / backfill commit
   - `RefreshExternalCollection`
3. Let the secondary consume/replay the replicated WAL/control-channel messages.
4. Compare the primary and secondary metadata/resulting state for snapshots, restored data, manifest updates, external collection refresh jobs, and segment metadata.

### Trigger Conditions

- **Frequency**: always for operations whose payload only carries the command and requires local recomputation on the secondary.
- **First observed after**: review of the current primary-secondary DDL callback paths.
- **Does NOT happen when**: the DDL is not replayed on the secondary, or when the WAL payload contains a deterministic primary-side result plus a valid primary-to-secondary logical segment mapping.

## Expected Behavior

Until these DDLs have a well-defined primary-secondary consistency contract, the replication/replay path should not process them on the secondary. The secondary should either skip them explicitly or reject/fence them with a clear unsupported-operation error, so it does not silently produce divergent metadata.

## Actual Behavior

The current callbacks can execute these operations on the secondary as if replaying the same command were equivalent to replaying the primary result. That is not true for these DDLs:

- `CreateSnapshot`: the WAL message only contains the snapshot command (`collectionID`, name, description, protection TTL). The secondary runs `GenSnapshot` locally, deriving local per-channel seek positions and local segment lists. WAL order is pchannel-scoped, and primary/secondary segment IDs are not the same, so the generated snapshot cannot be assumed to match the primary snapshot.
- `RestoreSnapshot`: depends on snapshot content and segment IDs. It also pins the source snapshot only during the primary RPC path; the secondary callback only receives a primary pin ID and does not create an equivalent local pin, so source snapshot protection is not equivalent.
[...]

PR #51454, opened 2026-07-16
fix: skip unreplicable replicated ddl

chyezh:
issue: #51446

- message properties: add an explicit `Unreplicable` marker for concrete WAL messages and expose builder opt-in through `WithUnreplicable`
- replication: skip unreplicable messages in the CDC sender and secondary replicate interceptor before callback replay
- DDL producers: mark unsupported snapshot, manifest backfill, and external collection refresh broadcasts as unreplicable
- streaming docs: document the message-level skip marker and current unsupported DDL replication behavior


liliu-z (commented):
/lgtm
/approve

PR #51455, opened 2026-07-16
fix: [cp 3.0] skip unreplicable replicated ddl

chyezh:
issue: #51446
pr: #51454

- message properties: add an explicit `Unreplicable` marker for concrete WAL messages and expose builder opt-in through `WithUnreplicable`
- replication: skip unreplicable messages in the CDC sender and secondary replicate interceptor before callback replay
- DDL producers: mark unsupported snapshot, manifest backfill, and external collection refresh broadcasts as unreplicable
- streaming docs: document the message-level skip marker and current unsupported DDL replication behavior

### DIFF
--- a/docs/agent_guides/streaming-system/message/message-semantic-collection.md
+++ b/docs/agent_guides/streaming-system/message/message-semantic-collection.md
@@ -18,6 +18,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 | CreateSnapshot | Broadcast: CChannel | No | SharedDBName + ExclusiveCollectionName + ExclusiveSnapshotName |
 | DropSnapshot | Broadcast: CChannel | No | ExclusiveSnapshotName |
 | RestoreSnapshot | Broadcast: CChannel | No | SharedDBName + ExclusiveCollectionName + ExclusiveSnapshotName |
+| DropSnapshotsByCollection | Broadcast: CChannel | No | SharedDBName + SharedCollectionName |
 | Import | Broadcast: VChannels (no CChannel) | No | — |
 | Insert | Single VChannel | No | — |
 | Delete | Single VChannel | No | — |
@@ -27,6 +28,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 | AlterLoadConfig | Broadcast: CChannel | No | SharedDBName + ExclusiveCollectionName |
 | DropLoadConfig | Broadcast: CChannel | No | SharedDBName + ExclusiveCollectionName (or ExclusiveCluster) |
 | BatchUpdateManifest | Broadcast: CChannel | No | SharedDBName + SharedCollectionName |
+| RefreshExternalCollection | Broadcast: CChannel | No | — |
 
 ## Message Descriptions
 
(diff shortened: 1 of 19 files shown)


tinswzy:

### Review: `fix: [cp 3.0] skip unreplicable replicated ddl`

Overall: **mergeable.** This is a faithful cherry-pick of the master change onto 3.0. The new `_ur` message property, the CDC-sender skip, and the secondary interceptor skip are all consistent with existing semantics, and I could not find a correctness or concurrency regression. Findings below are non-blocking (one compatibility caveat + a doc nit).

**Completeness check (passes).** I audited every producer of the affected message types (`CreateSnapshot`, `DropSnapshot`, `RestoreSnapshot`, `BatchUpdateManifest`, `RefreshExternalCollection`, `DropSnapshotsByCollection`). None of these types are `SelfControlled` (`pkg/streaming/util/message/message_type.go:27`), so the pre-existing self-controlled skip does **not** cover them — the `_ur` marker is the only thing keeping them out of replication. All six real WAL-broadcast sites are now marked:
- `internal/datacoord/services.go:2231/2288/2392/2751`, `internal/datacoord/snapshot_manager.go:762`, `internal/datacoord/services_commit_backfill.go:158`.
- `internal/rootcoord/ddl_callbacks_drop_collection.go:112` (`DropSnapshotsByCollection`) is correctly **not** marked: it is dispatched in-process via `registry.CallMessageAckCallback` (`internal/streamingcoord/server/broadcaster/registry/ack_message_callback.go:65`), never appended to the WAL, so the CDC replicator never sees it. Good.

**Skip-path correctness (passes).** `IsUnreplicable()` reuses the exact branch structure of `IsSelfControlled`: on the primary sender (`internal/cdc/replication/replicatestream/replicate_stream_client_impl.go:173`) it returns `ErrReplicateIgnored` and advances the last-replicated tick only when nothing is pending; on the secondary interceptor (`internal/streamingnode/server/wal/interceptors/replicate/replicates/impl.go:114`) `rh != nil` → `IgnoreOperation`, `rh == nil` → `ErrNotHandledByReplicateManager`, which the interceptor turns into a direct local WAL append (`replicate_interceptor.go:62-64`). Both branches are correct and race-free (the check runs before `impl.mu` is taken, same as the self-controlled check).

---

**Low — rolling-upgrade backlog is not retroactively protected.**
`replicate_stream_client_impl.go:173` / `impl.go:114`: skipping is keyed purely on the `_ur` property. Any snapshot/manifest DDL of these types that was already written to the primary WAL **before** the upgrade (and sits behind the secondary's `ReplicateCheckpoint`) carries no `_ur`, so after upgrading it will still be forwarded to the secondary and may fail replay — exactly the case this fix targets. This is inherent to the property-based design and is acceptable for a forward fix, but for a 3.0 cluster that already uses snapshots/manifest with active replication, the fix only protects **newly produced** messages. Worth calling out in the backport notes so operators don't assume an in-flight replication backlog is covered.

**Nit — docs could preempt a "missing entry" question.**
`docs/agent_guides/streaming-system/message/message-semantic-collection.md`: the message table now adds `DropSnapshotsByCollection` to the snapshot family, but the new "Replication Compatibility" paragraph lists only the other five types as `_ur`. A one-line note that `DropSnapshotsByCollection` is dispatched locally via the drop-collection ack callback (not a WAL broadcast) and therefore never enters the replication path would prevent a future reader from reading its absence as an oversight.

Tests are appropriate: the CDC unit test asserts `ErrReplicateIgnored` with an empty pending queue, and the manager test covers both the `rh == nil` (not-handled) and `rh != nil` (ignored) branches. Mocks are regenerated consistently across all four message mock files.

liliu-z (commented):
/lgtm
/approve

PR #51460, opened 2026-07-16
test: Add REST query order by coverage

lyyyuna:
## What this PR does

- Adds REST v2 query coverage for orderByFields across ascending, descending, multi-field, filtering, pagination, nullable fields, and invalid parameters.
- Uses class-scoped REST clients and shared collection setup consistent with the existing REST v2 test pattern.
- Verifies that omitting limit applies the REST default of 100 rows.
- Covers duplicate-key pagination behavior and exact validation error code 1100.

## Test

Against schema-evolution-v3-latest through port-forward:

    cd tests/restful_client_v2
    ../python_client/.venv/bin/python -m pytest testcases/test_query_order_by.py -v --tb=short -p no:rerunfailures --endpoint http://127.0.0.1:19530 --token root:Milvus

Result: 16 passed, 16 warnings in 40.87s.

### DIFF
--- a/tests/python_client/milvus_client/test_milvus_client_query_order.py
+++ b/tests/python_client/milvus_client/test_milvus_client_query_order.py
@@ -22,10 +22,26 @@ CUSTOM_PARTITION_NAME = "odd_partition"
 PRICE_MIN = 10.0
 RATING_MIN = 0.0
 CATEGORIES = [
-    "electronics", "tools", "books", "sports", "movies",
-    "games", "furniture", "health", "toys", "jewelry",
-    "cosmetics", "clothing", "garden", "food", "travel",
-    "pet_supplies", "beauty", "appliances", "automotive", "music"
+    "electronics",
+    "tools",
+    "books",
+    "sports",
+    "movies",
+    "games",
+    "furniture",
+    "health",
+    "toys",
+    "jewelry",
+    "cosmetics",
+    "clothing",
+    "garden",
+    "food",
+    "travel",
+    "pet_supplies",
+    "beauty",
+    "appliances",
+    "automotive",
+    "music",
 ]
 
 
@@ -82,7 +98,6 @@ def gen_query_order_data(nb, start_id=0):
     return rows
 
 
-
 # =====================================================================
 # Valid test cases — shared Collection A
 # =====================================================================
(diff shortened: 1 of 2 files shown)


yanliang567 (commented):
I found two verification gaps that should be addressed:

1. **[P1] The filter test can pass with the wrong PKs at the limit boundary.**
   In `test_query_order_by_with_filter`, `_assert_expected_values(rows, expected, ["price"])` only checks the sort-key values, while `_assert_pk_tie_breaker` only checks that the PKs returned within each price group are locally ascending. With the current data, the correct boundary rows for `price=14` are IDs `[4, 14, ..., 94]`, but returning `[104, 114, ..., 194]` would still satisfy both assertions. Please also compare the exact IDs, for example:

   ```python
   self._assert_expected_ids(rows, expected)
   ```

2. **[P2] The offset test does not exercise pagination across duplicate sort keys.**
   `score = (i * 17 + 11) % 200` is a permutation of `0..199`, so every sort key is unique. This means the test cannot detect unstable tie ordering, overlaps, or gaps between pages—the scenario the implicit PK tie-breaker is intended to handle. Please use a duplicated key such as `price` and compare the page's exact IDs with the corresponding baseline slice (ideally repeat the page request as well).

Non-blocking follow-up: the shared setup inserts and flushes only once, so this suite primarily exercises a single segment. A multi-insert/multi-flush L1 case would add useful coverage for the global merge path.

lyyyuna:
Thanks for pointing these out. I checked the Query ORDER BY design contract, and Section 8.2 explicitly states that tie order is non-deterministic, users should add an explicit final key (typically PK) for deterministic ordering, and implicit PK tie-breaking is a future option that is not implemented:

https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260203-query-orderby.md#L235-L240

Therefore, an exact PK-subset assertion without an explicit final key encodes behavior stronger than the approved design. I will leave the first filter-test suggestion unchanged for now. The second pagination suggestion is already addressed with a duplicated price key, an exact baseline slice, and a repeated page request. I will leave the test code and exact-ID assertions unchanged for now pending contract clarification.

lyyyuna:
@yanliang567  An implicit PK tie-breaker is a future option that is not implemented: [https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260203-query-orderby.md#L235-L240](https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260203-query-orderby.md#L235-L240). I updated the REST test helper/comments and the existing PyMilvus Query ORDER BY E2E comments in a09342c28f to remove the implicit-PK-contract wording and clarify that monotonic checks only describe the returned subset, not exact tied-row selection. Request parameters and assertion behavior are unchanged.

Issue #51418, opened 2026-07-15
[Bug]: StreamingNode panics when growing-source SaveBinlogPaths races with channel reassignment


zhuwenxing:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Environment

- Milvus version: `master-20260714-1acbe20c`
- Deployment mode(standalone or cluster): cluster
- MQ type(rocksmq, pulsar or kafka): kafka
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus `3.1.0rc62`
- OS(Ubuntu or CentOS): N/A; the environment was removed after the test
- CPU/Memory: StreamingNode request: 2 CPU / 8Gi; limit: 8 CPU / 16Gi
- GPU: N/A
- Others:
  - Environment: tc stage
  - Namespace: `chaos-testing`
  - Instance: `streamingnode-container-kill-494`
  - Helm chart: `milvus-5.0.24`
  - StreamingNode replicas: 2
  - Chaos type: `container-kill`
  - Chaos target: `streamingnode`
  - Jenkins build: [chaos-test-kafka-cron build 494](http://jenkins.milvus-cell.tc-op.zilliz.cn/job/chaos-test-kafka-cron/494/)

### K8s Pod List

Pod status captured by Jenkins at `2026-07-14 10:21:46 UTC`, before environment cleanup:

```
NAME                                                               READY   STATUS    RESTARTS   AGE
streamingnode-container-kill-494-etcd-0                            1/1     Running   0          63m
streamingnode-container-kill-494-etcd-1                            1/1     Running   0          63m
streamingnode-container-kill-494-etcd-2                            1/1     Running   0          63m
streamingnode-container-kill-494-kafka-0                           2/2     Running   0          63m
streamingnode-container-kill-494-kafka-1                           2/2     Running   0          63m
streamingnode-container-kill-494-kafka-2                           2/2     Running   1          63m
streamingnode-container-kill-494-kafka-exporter-8555d69fb-v5f84   1/1     Running   4          63m
streamingnode-container-kill-494-milvus-datanode-5bddc8fbd2xn9l   1/1     Running   2          63m
streamingnode-container-kill-494-milvus-datanode-5bddc8fbdj5mf5   1/1     Running   2          63m
streamingnode-container-kill-494-milvus-mixcoord-74bc95d87lwrs5   1/1     Running   2          63m
[...]

PR #51464, opened 2026-07-16
fix: avoid growing source panic on stale channel

chyezh:
issue: #51418

- flushcommon syncmgr: skip the write-buffer panic callback for stale growing-source meta errors while preserving failure metrics
- growing-source tests: cover stale channel and missing segment errors without invoking the failure callback

### DIFF
--- a/internal/flushcommon/syncmgr/growing_source.go
+++ b/internal/flushcommon/syncmgr/growing_source.go
@@ -564,6 +564,10 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
+	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
+		metrics.DataNodeFlushBufferCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel, t.level.String()).Inc()
+		return
+	}
 	if t.failureCallback != nil {
 		t.failureCallback(err)
 	}
--- a/internal/flushcommon/syncmgr/growing_source.go
+++ b/internal/flushcommon/syncmgr/growing_source.go
@@ -564,7 +564,7 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
-	if errors.Is(err, merr.ErrChannelNotFound) {
+	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
 		return
 	}
 	if t.failureCallback != nil {
--- a/internal/flushcommon/syncmgr/growing_source.go
+++ b/internal/flushcommon/syncmgr/growing_source.go
@@ -564,7 +564,7 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
-	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
+	if errors.Is(err, merr.ErrChannelNotFound) {
 		return
 	}
 	if t.failureCallback != nil {
--- a/internal/flushcommon/syncmgr/growing_source.go
+++ b/internal/flushcommon/syncmgr/growing_source.go
@@ -565,7 +565,6 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
 	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
-		metrics.DataNodeFlushBufferCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel, t.level.String()).Inc()
 		return
 	}
 	if t.failureCallback != nil {
--- a/internal/flushcommon/syncmgr/growing_source_test.go
+++ b/internal/flushcommon/syncmgr/growing_source_test.go
@@ -87,37 +87,42 @@ func (s *fakeCommitGrowingFlushSource) CommitGrowingFlush(targetOffset int64) {
 func TestGrowingSourceSyncTaskHandleErrorSkipsFailureCallbackForStaleMetaErrors(t *testing.T) {
 	paramtable.Get().Init(paramtable.NewBaseTable())
 
-	metrics.DataNodeFlushBufferCount.Reset()
-	callbackCalled := false
-	task := NewGrowingSourceSyncTask().
-		WithFailureCallback(func(error) {
-			callbackCalled = true
-			panic("failure callback should not be called")
+	testCases := []struct {
+		name string
+		err  error
+	}{
+		{
+			name: "channel_not_found",
+			err:  merr.WrapErrChannelNotFound("ch"),
+		},
+		{
+			name: "segment_not_found",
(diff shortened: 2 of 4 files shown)


internal/flushcommon/syncmgr/growing_source.go
```diff
@@ -564,6 +564,10 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
+	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
+		metrics.DataNodeFlushBufferCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel, t.level.String()).Inc()
```
liliu-z: The new stale-channel branch increments DataNodeFlushBufferCount{Fail} and returns without panicking, but HandleError runs twice per errored task: once from Run's deferred handler (growing_source.go:600) and once from the sync-manager `handler` callback (sync_manager.go:164, prepended in submit). Before this PR the first invocation panicked inside the write-buffer failureCallback (default errorHandler is panic(err)), so the second call was never reached; the graceful early-return removes that guard, so a single stale-channel error — a routine event during channel handoff — now increments the failure counter twice (2x inflation). Increment exactly once by deduping the two HandleError paths or moving the Inc out of the branch both callers hit.

internal/flushcommon/syncmgr/growing_source.go
```diff
@@ -564,6 +564,10 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
+	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
```
liliu-z: Can we keep this to ErrChannelNotFound? SaveBinlogPaths checks ownership first, so this reassignment race cannot return ErrSegmentNotFound. A missing segment on the current owner will now skip the panic and retry every 100ms, potentially forever while holding the checkpoint. That needs separate cleanup semantics.

internal/flushcommon/syncmgr/growing_source.go
```diff
@@ -564,6 +564,10 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
...
+		metrics.DataNodeFlushBufferCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel, t.level.String()).Inc()
+		return
```
liliu-z, line 568: After channel removal, this lets the WB completion callback record growing_source_sync_failure_count even though the WB is already closed. That gauge is not removed by metrics cleanup, so it can stay non-zero for a dead channel. Could closed + ChannelNotFound be treated as cancellation instead?

internal/flushcommon/syncmgr/growing_source_test.go
```diff
@@ -82,6 +82,40 @@ func (s *fakeCommitGrowingFlushSource) CommitGrowingFlush(targetOffset int64) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			callbackCalled := false
+			task := NewGrowingSourceSyncTask().
+				WithFailureCallback(func(error) {
+					callbackCalled = true
+					panic("failure callback should not be called")
+				})
+
+			require.NotPanics(t, func() {
+				task.HandleError(tc.err)
```
liliu-z: This only calls HandleError directly. The production path is Run defer -> SyncManager handler -> WriteBuffer callback. Can we cover that path with a real SyncManager and assert one metric increment, no fatal callback, and that the error still reaches the WB callback?

internal/flushcommon/syncmgr/growing_source.go
chyezh: Fixed in 8d5121a7df. HandleError now only treats ErrChannelNotFound as stale ownership; ErrSegmentNotFound goes through the normal failure callback and fail accounting path.

internal/flushcommon/syncmgr/growing_source.go
chyezh: Fixed in 4d07150791. The stale-channel HandleError path no longer increments DataNodeFlushBufferCount, so the Run defer + SyncManager callback double invocation does not double-count. Added the repeated HandleError assertion for this case.

internal/flushcommon/syncmgr/growing_source_test.go
chyezh: Added in 8d5121a7df. The new writebuffer-level regression test uses a real SyncManager and covers Run defer -> SyncManager handler -> WriteBuffer callback. For the closed-WB stale-channel case it asserts no fatal callback, nil future error, progress cleanup, and zero stale failure metrics. The direct HandleError tests now cover the ErrChannelNotFound / ErrSegmentNotFound split.

internal/flushcommon/syncmgr/growing_source.go
chyezh: Fixed in 8d5121a7df. When the write buffer is already closed and the growing-source task returns ErrChannelNotFound, the completion callback now treats it as cancellation: it rolls back task bookkeeping, clears the growing-source failure gauge, removes the progress entry, and returns nil.

liliu-z (commented):
/lgtm
/approve

xiaofan-luan (commented):
@yezhen I took another pass over this fix and traced the channel-removal and flush-task lifecycle end to end.

The immediate guard is directionally correct for the reported race: a stale owner rejected by DataCoord must not panic the whole StreamingNode. However, I think we still need a lifecycle fix, ideally before carrying this as-is into the 3.0 backport (#51465):

1. Drain/cancel every flush task owned by the PChannel during revoke, including normal SyncTask, GrowingSourceSyncTask, queued tasks, and in-flight metadata commits. The channel should first stop accepting new tasks, then cancel or await the existing tasks, and only then clean up the WriteBuffers. Today RemoveChannel -> Close(false) only marks the WriteBuffer closed and stops its retry timer; it does not drain tasks already submitted to SyncManager.

2. A confirmed fence should make the local node relinquish the channel. It should not panic or retry: transition the channel to fenced/revoked, cancel all channel tasks, stop checkpoint reporting/consumption, and clean up local WriteBuffer state. This needs an explicit fence/term-mismatch signal, though. SaveBinlogPaths currently returns ErrChannelNotFound both for a real owner mismatch and for failure to query the latest assignment, so treating every ErrChannelNotFound as ownership loss could incorrectly revoke a healthy owner. A dedicated ErrChannelFenced carrying requested/current owner and assignment term would make this safe.

3. ErrSegmentNotFound should not be grouped with the stale-owner case. SaveBinlogPaths performs the owner check before the segment lookup, so the reassignment race reports ErrChannelNotFound, not ErrSegmentNotFound. With the current final code, SegmentNotFound skips the fatal callback while the WriteBuffer can retain pendingCommitted/progress and retry at the 100 ms interval, or leave the growing-source checkpoint pinned if local segment metadata is gone. This needs separate terminal cleanup semantics, or should remain on the normal failure path.

4. Please add a production-path regression test covering GrowingSourceSyncTask.Run -> SyncManager callback -> WriteBuffer completion while the channel is closing/fenced. The final test only calls HandleError directly; the full-path test added in 8d5121a was removed by 9c2cd21. Re-running the original StreamingNode container-kill/channel-reassignment chaos case would also validate that flush resumes after the new owner recovers.

So I agree with avoiding the panic, but I view it as a crash guard rather than the complete ownership-transition fix. The robust model is graceful revoke: ACTIVE -> DRAINING -> REVOKED, plus a failure path where an explicit term fence immediately forces local cancellation and relinquishment.

PR #51465, opened 2026-07-16
fix: [cp 3.0] avoid growing source panic on stale channel

chyezh:
issue: #51418
pr: #51464

- flushcommon syncmgr: skip the write-buffer panic callback for stale growing-source meta errors while preserving failure metrics
- growing-source tests: cover stale channel and missing segment errors without invoking the failure callback

### DIFF
--- a/internal/flushcommon/syncmgr/growing_source.go
+++ b/internal/flushcommon/syncmgr/growing_source.go
@@ -564,6 +564,10 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
+	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
+		metrics.DataNodeFlushBufferCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel, t.level.String()).Inc()
+		return
+	}
 	if t.failureCallback != nil {
 		t.failureCallback(err)
 	}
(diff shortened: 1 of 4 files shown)


internal/flushcommon/syncmgr/growing_source.go
```diff
@@ -564,6 +564,10 @@ func (t *GrowingSourceSyncTask) TargetOffset() int64 {
 }
 
 func (t *GrowingSourceSyncTask) HandleError(err error) {
+	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
+		metrics.DataNodeFlushBufferCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel, t.level.String()).Inc()
```
liliu-z: For a stale-meta failure, HandleError runs twice for one task — from Run's defer (growing_source.go:600) and from the sync manager's submit-handler callback (sync_manager.go:164) — and the new early-return branch increments DataNodeFlushBufferCount{fail} each time, so a single ErrSegmentNotFound/ErrChannelNotFound flush is counted as two failures. Before this PR the default failureCallback panicked on the first HandleError call (before any increment, crashing the datanode), so the double-invocation never surfaced; now that the panic is skipped both increments fire. Move the fail-metric accounting to a single owner (the Run defer or the submit handler, not both) so one failure equals one count.

chyezh: Fixed in a54d7eae8f. The stale-channel HandleError path no longer increments DataNodeFlushBufferCount, so the Run defer + SyncManager callback double invocation does not double-count. The follow-up 7fbe586e53 keeps only ErrChannelNotFound on this cancellation path and adds the closed-WB cleanup/test coverage.

liliu-z (commented):
/lgtm
/approve

Issue #51466, opened 2026-07-16
[Bug]: index should be preferred for json related filtering


zhengbuqian:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version:
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

even though the filter operator may not benefit from the index algo, but using the inde still avoids the overhead of json parsing so we can run the `like` or other filtering directly on the deduped values , thus we should still use index for json path `like` operations

### Steps To Reproduce

```markdown

```

Issue #51193, opened 2026-07-09
JSON flat index should preserve UNKNOWN for scalar predicates on array values


yanliang567:
## Problem

JSON flat index can return rows that the raw filtering path treats as `UNKNOWN` for scalar predicates on non-root JSON paths.

A representative failing case is:

```sql
not (json_field["a"] > 2)
```

with a JSON payload:

```json
{"a": [2]}
```

The raw filtering path treats array-vs-number comparison as `UNKNOWN`; applying `not` to `UNKNOWN` should still exclude the row. With a JSON flat index (`json_cast_type=json`), array elements are indexed and this row can be incorrectly returned.

## Reproduction Outline

1. Create a collection with a nullable JSON field and rows including:
   - `{ "a": 2 }`
   - `{ "a": 3 }`
   - `{ "a": [2] }`
2. Create an inverted JSON flat index on the JSON field with `json_cast_type=json`.
3. Query with:

```sql
not (json_field["a"] > 2)
```

## Expected Behavior

Only comparable known values should participate in scalar comparison predicates. The row `{ "a": [2] }` should be treated as `UNKNOWN` and excluded.

Expected IDs for the representative dataset: only the row where `a == 2`.

## Actual Behavior

The JSON flat index path can include the row where `a` is an array, for example `{ "a": [2] }`, while the raw filtering path excludes it.

## Related

- Found while adding Python e2e JSON filtering coverage in PR #51171.
- Related to #50976, #50699, and #50698.

zhengbuqian:
not planning to fix, will deprecate flat index soon. user can use path index or shredding as alternative

PR #51467, opened 2026-07-16
fix: honor scalar index policy for JSON string predicates

zhengbuqian:
issue: https://github.com/milvus-io/milvus/issues/51466

## What changed

`PhyUnaryRangeFilterExpr::DetermineExecPath()` previously used a
JSON-specific hard-coded denylist for string predicates:

- `Match`
- `PostfixMatch`
- `InnerMatch`

This bypassed the concrete scalar index's `ShouldUseOp()` policy. As a result:

- indexes capable of evaluating these predicates, such as `STL_SORT`, were
  forced to scan and parse raw JSON;
- JSON string predicates could use a different execution policy from ordinary
  VARCHAR predicates;
- `RegexMatch` remained on the index path even for indexes whose planner policy
  explicitly preferred raw-data execution.

### DIFF
--- a/internal/core/src/exec/expression/UnaryExpr.cpp
+++ b/internal/core/src/exec/expression/UnaryExpr.cpp
@@ -2043,10 +2043,8 @@ PhyUnaryRangeFilterExpr::DetermineExecPath() {
             switch (val_type) {
                 case DataType::STRING:
                 case DataType::VARCHAR:
-                    can_use =
-                        expr_->op_type_ != proto::plan::OpType::Match &&
-                        expr_->op_type_ != proto::plan::OpType::PostfixMatch &&
-                        expr_->op_type_ != proto::plan::OpType::InnerMatch;
+                    can_use = SegmentExpr::CanUseIndexForOp<std::string>(
+                        expr_->op_type_);
                     break;
                 default:
                     can_use = true;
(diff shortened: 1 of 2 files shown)


internal/core/src/exec/expression/UnaryExpr.cpp
```diff
@@ -2043,10 +2043,8 @@ PhyUnaryRangeFilterExpr::DetermineExecPath() {
             switch (val_type) {
                 case DataType::STRING:
                 case DataType::VARCHAR:
-                    can_use =
-                        expr_->op_type_ != proto::plan::OpType::Match &&
-                        expr_->op_type_ != proto::plan::OpType::PostfixMatch &&
-                        expr_->op_type_ != proto::plan::OpType::InnerMatch;
+                    can_use = SegmentExpr::CanUseIndexForOp<std::string>(
```
liliu-z, line 2046: `CanUseIndexForOp<std::string>` at line 2046 routes JSON string predicates so a generic `Match` (e.g. `LIKE "r_d"`) is sent to the `JsonFlatIndex`, which matches flattened array string elements and marks the row valid via `ExactPathExists(String)` — whereas the raw-data path treats an array sitting at a string path as a 3VL type-mismatch and excludes it. The two paths therefore disagree: a JSON doc whose queried path holds a string array is counted as a match under the index but excluded under raw evaluation, a pre-existing flat-index-vs-raw divergence this PR newly surfaces on `Match` (it already fixes the analogous `RegexMatch` case). Consensus is that this needs a semantics decision plus a root-cause fix and a flat-index array-valued test, tracked separately rather than blocking this PR.

zhengbuqian: same as #51193, not planning to fix

PR #51473, opened 2026-07-16
enhance: [3.0] move sealed segment PK state into runtime snapshot (#51395)

congqixia:
Cherry-pick from master
pr: #51395
Related to #51068

Store PK index slots and virtual PK offset maps in RuntimeResourceState. Build Storage V1 and V2 PK indexes through the unified translator and stage PK replacements until the final state publication.

Make PK lookup, range search, bulk subscript, and delete filtering use a single published snapshot. Remove the sealed InsertRecord fallback and clear stale PK resources when external segments become empty.

Add regression coverage for atomic PK replacement and zero-row external segment cleanup.


liliu-z (commented):
/lgtm
/approve

Issue #51472, opened 2026-07-16
[Bug]: QueryNode delegator panics on nil worker in ReleaseSegments when target worker is offline, leaking growingSegmentLock


bigsheeper:
## Environment

- **Milvus Version**: master (7b05f993f3)
- **Deployment Mode**: cluster
- **MQ**: any
- **SDK**: N/A (internal path)
- **OS**: any

### Option B: Steps (systemic / timing issue)

1. Deploy: cluster with multiple QueryNodes, a loaded collection whose shard leader (delegator) is on node A and a sealed/growing segment replica on worker node B.
2. Trigger: make node B unavailable (kill the pod or make it unhealthy) so that `workerManager.GetWorker(ctx, targetNodeID)` fails.
3. While node B is down, QueryCoord issues `ReleaseSegments` with `NeedTransfer=true` for segments on node B (routine during rebalancing / node-down recovery). The shard leader forwards it via `shardDelegator.ReleaseSegments(ctx, req, false)` (`internal/querynodev2/services.go:682`).
4. Observed via: QueryNode panic (nil pointer dereference) in the delegator; if the release scope includes growing segments, subsequent inserts on that channel hang.

### Trigger Conditions

- **Frequency**: always, once `GetWorker` returns an error on this path (worker offline/unhealthy)
- **Does NOT happen when**: the target worker is reachable and healthy, or `force=true`

## Expected Behavior

`ReleaseSegments` should return the `GetWorker` error to QueryCoord so the release can be retried, without panicking and without leaking locks — the same way `LoadSegments` handles the identical failure (`internal/querynodev2/delegator/delegator_data.go:644-648` returns early on `GetWorker` error).

## Actual Behavior

In `internal/querynodev2/delegator/delegator_data.go:1179-1189`, when `GetWorker` fails the error is recorded into `releaseErr` but execution falls through and calls `worker.ReleaseSegments(ctx, req)` on a nil `cluster.Worker` interface, causing a nil-pointer panic:

```go
worker, err := sd.workerManager.GetWorker(ctx, targetNodeID)
if err != nil {
    log.Warn(ctx, "delegator failed to find worker", mlog.Err(err))
    releaseErr = err
}
req.Base.TargetID = targetNodeID
err = worker.ReleaseSegments(ctx, req) // worker is nil here
```

[...]

PR #51474, opened 2026-07-16
fix: avoid nil worker dereference in delegator ReleaseSegments

bigsheeper:
issue: #51472

## Summary

When the target worker is offline or unhealthy, `workerManager.GetWorker` returns `(nil, err)`. `shardDelegator.ReleaseSegments` recorded the error into `releaseErr` but still fell through and invoked `worker.ReleaseSegments(ctx, req)` on the nil `cluster.Worker` interface, panicking the QueryNode. When the release scope included growing segments (`DataScope_All` / `DataScope_Streaming`), the panic unwound past the `growingSegmentLock.Unlock()`, leaking the mutex and permanently blocking the channel's insert pipeline (`ProcessInsert` blocks forever).

This is reachable from the normal `needTransfer` release path (`QueryNode.ReleaseSegments` → `shardDelegator.ReleaseSegments(ctx, req, false)`) whenever QueryCoord releases segments on a node that has just gone offline — routine during rebalancing / node-failure recovery.

## Fix

Guard the worker call in an `else` branch so a failed `GetWorker` returns the error to QueryCoord for retry, matching how `LoadSegments` handles the same failure. The single `growingSegmentLock` unlock path and the existing `releaseErr` semantics (distribution already removed, segments already excluded, release retried by QueryCoord) are preserved.

## Test

- New regression test `TestReleaseSegmentsWorkerNotAvailable`: mocks `GetWorker` returning `(nil, err)`, asserts `ReleaseSegments` returns the error without panicking and that `growingSegmentLock` is released afterward. The test reproduces the nil-pointer panic at `delegator_data.go:1185` before the fix.
- Full `./internal/querynodev2/delegator/...` package tests pass.
- `make static-check` passes.

### DIFF
--- a/internal/querynodev2/delegator/delegator_data.go
+++ b/internal/querynodev2/delegator/delegator_data.go
@@ -1180,12 +1180,12 @@ func (sd *shardDelegator) ReleaseSegments(ctx context.Context, req *querypb.Rele
 		if err != nil {
 			log.Warn(ctx, "delegator failed to find worker", mlog.Err(err))
 			releaseErr = err
-		}
-		req.Base.TargetID = targetNodeID
-		err = worker.ReleaseSegments(ctx, req)
-		if err != nil {
-			log.Warn(ctx, "worker failed to release segments", mlog.Err(err))
-			releaseErr = err
+		} else {
+			req.Base.TargetID = targetNodeID
+			if err = worker.ReleaseSegments(ctx, req); err != nil {
+				log.Warn(ctx, "worker failed to release segments", mlog.Err(err))
+				releaseErr = err
+			}
 		}
 	}
 	if len(growing) > 0 {
--- a/internal/querynodev2/delegator/delegator_data_test.go
+++ b/internal/querynodev2/delegator/delegator_data_test.go
@@ -1733,6 +1733,31 @@ func (s *DelegatorDataSuite) TestReleaseSegment() {
 	s.NoError(err)
 }
 
+func (s *DelegatorDataSuite) TestReleaseSegmentsWorkerNotAvailable() {
+	ctx := context.Background()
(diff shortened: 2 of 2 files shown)


czs007:
Adversarial review surfaced one non-blocking nitpick:

---

The adversarial review pass surfaced one non-blocking nitpick regarding test assertion strength; everything else came back clean.

## Nitpick

**`internal/querynodev2/delegator/delegator_data_test.go:1748`** — The regression test `TestReleaseSegmentsWorkerNotAvailable` only asserts `s.Error(err)`, so any error at all will satisfy it, including one whose inner error code has been masked by an outer wrapper.

Consider adding `s.ErrorIs(err, merr.ErrNodeNotAvailable)` alongside the existing assertion to pin down the contract that `GetWorker`'s error code propagates through unchanged. Since `merr`'s `Is` compares by `errCode` (`pkg/util/merr/errors.go:413-419`), this assertion would fail if someone later wraps the return in a way that hides the inner code. The same suite already uses this pattern at lines 896 and 1058, and `merr` is already imported at line 62 — so this is zero-cost to add. *(raised by wzy)*

bigsheeper:
Good catch — adopted in eca560b.

I checked that the assertion actually pins something rather than just restating `s.Error`: `releaseErr` is assigned straight from `GetWorker` and returned unwrapped, and `milvusError.Is` compares by `errCode` (`pkg/util/merr/errors.go:413-419`), so a future wrapper that masks the inner code would now fail this test instead of passing on a bare `s.Error(err)`.

Full package is green: `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator` → ok 29.8s.

czs007:

czs007 (approved):
/lgtm

Issue #51475, opened 2026-07-16
[Bug]: NewManifestRecordReader passes raw encryption key instead of base64, breaking encrypted StorageV3 manifest reads


bigsheeper:
## Environment

- **Milvus Version**: master (6c7300f15c)
- **Deployment Mode**: any (requires cluster encryption enabled via cipher plugin)
- **MQ**: any
- **SDK**: N/A (internal path)
- **OS**: any

### Option B: Steps

1. Enable cluster encryption (`IsClusterEncryptionEnabled() == true`) with a cipher plugin, and create a collection with an encryption zone (EZ).
2. Write segments in StorageV3 manifest (loon) format for that collection.
3. Trigger any read that goes through `storage.NewManifestRecordReader` — e.g. a compaction on a V3 segment (`internal/datanode/compactor/compactor_common.go:301`) or the external function executor (`internal/datanode/external/function_executor.go:225`).
4. Observed via: decryption failure (or wrong key material) when the storage plugin consumes the key.

### Trigger Conditions

- **Frequency**: always on this path when encryption is enabled and the caller did not pre-supply `rwOptions.pluginContext`
- **Does NOT happen when**: reading via `NewBinlogRecordReader` (correct base64 encoding), or encryption disabled

## Expected Behavior

`StoragePluginContext.EncryptionKey` must carry the base64 encoding of the raw key, as produced by every other construction site:

- `internal/storage/rw.go:320` (`NewBinlogRecordReader`): `base64.StdEncoding.EncodeToString(unsafe)`
- `internal/storage/rw.go:438` (`NewBinlogRecordWriter`): `base64.StdEncoding.EncodeToString(unsafe)`
- `internal/flushcommon/syncmgr/pack_writer_v2.go:136`: `base64.StdEncoding.EncodeToString(unsafe)`
- `internal/util/hookutil/cipher.go:384`: `base64.StdEncoding.EncodeToString(key)`

## Actual Behavior

`internal/storage/rw.go:385` (`NewManifestRecordReader`) sets the field to the raw key bytes reinterpreted as a string:

```go
pluginContext = &indexcgopb.StoragePluginContext{
    EncryptionZoneId: ez.EzID,
    CollectionId:     ez.CollectionID,
    EncryptionKey:    string(unsafe), // raw bytes, NOT base64
}
```

[...]

PR #51480, opened 2026-07-16
fix: base64-encode encryption key in NewManifestRecordReader

bigsheeper:
issue: #51475

## Summary

`StoragePluginContext.EncryptionKey` is a base64-carrying field: every producer encodes the raw key with `base64.StdEncoding.EncodeToString` — `NewBinlogRecordReader` (`rw.go:320`), `NewBinlogRecordWriter` (`rw.go:438`), `pack_writer_v2.go:136`, and `hookutil.GetCPluginContext` (`cipher.go:384`). The one exception was `NewManifestRecordReader` (`rw.go:385`), which passed the raw key bytes as a string.

For an encryption-enabled collection stored in StorageV3 manifest (loon) format, reads going through `NewManifestRecordReader` — compaction (`compactor_common.go:301`) and the external function executor (`function_executor.go:225`) — therefore handed the cipher plugin differently-encoded key material than the writer used.

## Fix

One line: encode the key with base64, matching the four sibling construction sites. Also fixes `TestNewManifestRecordReaderBranches`, which had asserted the raw key (codifying the bug), to assert the base64-encoded key.

## Test

- `TestNewManifestRecordReaderBranches` now asserts the base64 contract and passes.
- Full `./internal/storage/` package passes locally, excluding the six Azure/GCP-endpoint tests (`TestAzureObjectStorage`, `TestReadFile`, `TestAzureChunkManager`, `TestGcpNativeObjectStorage`, `TestGcpNativeReadFile`, `TestInitRemoteChunkManager`) which require the cloud emulators available only in CI; none of them touch the changed code path.
- `make static-check` passes.

### DIFF
--- a/internal/storage/rw.go
+++ b/internal/storage/rw.go
@@ -382,7 +382,7 @@ func NewManifestRecordReader(ctx context.Context, manifestPath string, schema *s
 					pluginContext = &indexcgopb.StoragePluginContext{
 						EncryptionZoneId: ez.EzID,
 						CollectionId:     ez.CollectionID,
-						EncryptionKey:    string(unsafe),
+						EncryptionKey:    base64.StdEncoding.EncodeToString(unsafe),
 					}
 				}
 			}
--- a/internal/storage/serde_events_test.go
+++ b/internal/storage/serde_events_test.go
@@ -19,6 +19,7 @@ package storage
 import (
 	"bytes"
 	"context"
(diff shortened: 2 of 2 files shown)


czs007:

bigsheeper:
rerun cpp-unit-test

czs007 (approved):
/lgtm

PR #51483, opened 2026-07-16
fix: prevent schema updates from racing insert payload conversion

### LINKED ISSUE #51436: [Bug]: StreamingNode panics with vchannel not found and unexpected new field during schema evolution workload
## Environment

- **Milvus image**: `harbor.tencent-stage.zilliz.cn/milvusdb/milvus:master-20260715-25ac6b4c-amd64`
- **Deployment mode**: cluster, Storage V3 with embedded Woodpecker
- **Topology**: 3 Proxy, 3 QueryNode, 3 StreamingNode, 3 DataNode, 1 MixCoord
- **Runner image**: `harbor.tencent-stage.zilliz.cn/qa/schema-evolution-runner:20260710-rc61`
- **Workflow**: `schema-evolution-qn-fault-mt2tr`
- **Namespace**: `argo-schema-evolution-qn-fault-mt2tr` (deleted by workflow cleanup)

## Reproduction

1. Deploy the cluster topology above.
2. Run schema evolution tests that repeatedly create collections with three shards and multiple partitions.
3. Run concurrent operations on one collection:
   - add/drop nullable VARCHAR field `stress_extra`
   - add/drop BM25 function field `stress_bm25`
   - insert/delete/query/vector search
4. During the concurrent workload, periodically force-delete one StreamingNode pod, wait for the replacement pods to become Ready, then wait 60 seconds before the next deletion.

### Trigger Conditions

- **Frequency**: both panic signatures were observed once in this workflow.
- **Workload duration**: approximately 600 seconds for the concurrent DDL/DML phase.

## Expected Behavior

StreamingNode should not panic while collections are created or while DDL/DML operations and expected StreamingNode restarts are running.

## Actual Behavior

Two unexpected StreamingNode panics were observed in the same workflow.

### Panic 1: vchannel not found during collection creation

- **Time**: `2026-07-15T23:38:17Z`
- **Pod**: `schema-evolution-milvus-streamingnode-5d84c6cb7b-4cmg7`
- **Collection**: `test_add_function_and_drop_field_after_querynode_r_3qidg7`
- **Collection ID**: `467706411145214115`
- **Exit code**: `134`

```text
panic: code: STREAMING_CODE_INNER, cause: critical error: vchannel not found,
vchannel: by-dev-rootcoord-dml_1_467706411145214115v1, timetick: 0

github.com/milvus-io/milvus/internal/streamingnode/server/flusher/flusherimpl.(*versionedSchemaManager).GetSchema
github.com/milvus-io/milvus/internal/streamingnode/server/flusher/flusherimpl.RecoverWALFlusher
```

The panic happened immediately after the CreateCollection messages for the new collection were appended and acknowledged.

### Panic 2: unexpected new field in growing segment

- **Time**: `2026-07-16T00:13:30Z`
- **Pod**: `schema-evolution-milvus-streamingnode-5d84c6cb7b-6592j`
- **Collection**: `test_concurrent_ddl_dml_with_periodic_streamingnod_gpdgaj`
- **Collection ID**: `467706411158908666`
- **Segment ID**: `467706411145215091`
- **Field ID**: `580`
- **Field name**: `stress_extra`
- **Field type**: nullable VARCHAR
- **Exit code**: `134`

```text
panic: Assert "insert_record_.is_data_exist(field_id)" =>
unexpected new field in growing segment 467706411145215091, field id 580
at ../internal/core/src/segcore/SegmentGrowingImpl.cpp:658

github.com/milvus-io/milvus/internal/querynodev2/delegator.(*shardDelegator).ProcessInsert
```

Immediately before the panic, the logs contained:

```text
insertMessage schema version mismatch
schemaVersion=950
collectionSchemaVersion=951
```

The affected StreamingNode restarted after the panic. No OOMKilled event was observed.

## Cleanup

The Argo onExit cleanup succeeded and deleted the test namespace.

sijie-ni-0214:
## Problem

During schema evolution, SQN can convert an insert payload with the old schema and then race with a schema update before ProcessInsert creates and writes the growing segment. If the old payload still contains a dropped field, the native collection may already use the new schema and segcore rejects the write with unexpected new field.

## Solution

- Add a collection-local schema-transition RW lock.
  - Insert holds the reader lock from schema snapshot use through payload conversion and delegator.ProcessInsert, so a growing segment is created and written in the same schema epoch as its payload.
  - Schema update and existing-collection load refresh paths take the writer lock around native schema mutation and Go schema snapshot publication.
- Acquire a temporary collection lease before releasing collectionManager.mut. This keeps the collection alive while the writer waits, without serializing unrelated collection-manager operations behind a schema transition.
- Preserve dropped-field replay handling: when DDL has already completed, payload conversion filters fields absent from the current schema before inserting into a new-schema growing segment.

## Tests

- Add a deterministic native regression: pause after a v950 payload containing field 580 is converted, queue the v951 drop-field update, then resume a real NewSegment plus growing.Insert and verify both insert and DDL succeed.
- Assert the transition reader covers both payload conversion and ProcessInsert.
- Cover the DDL-first replay path, where field 580 is filtered before native insertion.
- Cover both schema writer entry points: UpdateSchema and existing-collection PutOrRef.

## Validation

- make static-check
- make build-go
- Native pipeline regression tests repeated 20 times with dynamic,test
- Collection schema-transition and lease tests repeated 20 times
- internal/querynodev2/pipeline and internal/querynodev2/segments package tests
- gofumpt -l and git diff --check

Fixes #51436

### DIFF
--- a/internal/querynodev2/pipeline/insert_node.go
+++ b/internal/querynodev2/pipeline/insert_node.go
@@ -44,9 +44,8 @@ type insertNode struct {
 	functionStore *function.FunctionRunnerLocalStore
 }
 
-func (iNode *insertNode) addInsertData(insertDatas map[UniqueID]*delegator.InsertData, msg *InsertMsg, collection *Collection) {
+func (iNode *insertNode) addInsertData(insertDatas map[UniqueID]*delegator.InsertData, msg *InsertMsg, schema *schemapb.CollectionSchema) {
 	ctx := msg.TraceCtx()
-	schema := collection.Schema()
 	insertRecord, skippedFields, err := storage.TransferInsertMsgToInsertRecord(schema, msg)
 	if err != nil {
 		err = merr.Wrap(err, "failed to get primary keys")
@@ -132,25 +131,26 @@ func (iNode *insertNode) Operate(in Msg) Msg {
 			mlog.Error(ctx, "insertNode with collection not exist", mlog.Int64("collection", iNode.collectionID))
 			panic("insertNode with collection not exist")
 		}
-		schema := collection.Schema()
-		functionOutputFieldIDs, err := iNode.functionStore.OutputFieldIDs(schema)
-		if err != nil {
-			mlog.Error(ctx, "failed to get embedding output fields", mlog.String("channel", iNode.channel), mlog.Err(err))
-			panic(err)
-		}
-
-		insertDatas := make(map[UniqueID]*delegator.InsertData)
-		for _, msg := range nodeMsg.insertMsgs {
-			if len(functionOutputFieldIDs) > 0 && !function.HasAllFieldDataByID(msg.GetFieldsData(), functionOutputFieldIDs) {
-				if err := iNode.functionStore.FillEmbeddingData(iNode.collectionID, schema, msg.InsertRequest); err != nil {
-					mlog.Error(msg.TraceCtx(), "failed to fill embedding data for insert message", mlog.String("channel", iNode.channel), mlog.Err(err))
-					panic(err)
-				}
+		collection.WithInsertSchemaTransition(func(schema *schemapb.CollectionSchema) {
+			functionOutputFieldIDs, err := iNode.functionStore.OutputFieldIDs(schema)
+			if err != nil {
+				mlog.Error(ctx, "failed to get embedding output fields", mlog.String("channel", iNode.channel), mlog.Err(err))
+				panic(err)
 			}
-			iNode.addInsertData(insertDatas, msg, collection)
-		}
 
-		iNode.delegator.ProcessInsert(insertDatas)
+			insertDatas := make(map[UniqueID]*delegator.InsertData)
+			for _, msg := range nodeMsg.insertMsgs {
+				if len(functionOutputFieldIDs) > 0 && !function.HasAllFieldDataByID(msg.GetFieldsData(), functionOutputFieldIDs) {
+					if err := iNode.functionStore.FillEmbeddingData(iNode.collectionID, schema, msg.InsertRequest); err != nil {
+						mlog.Error(msg.TraceCtx(), "failed to fill embedding data for insert message", mlog.String("channel", iNode.channel), mlog.Err(err))
+						panic(err)
+					}
+				}
+				iNode.addInsertData(insertDatas, msg, schema)
+			}
+
+			iNode.delegator.ProcessInsert(insertDatas)
+		})
 	}
 	metrics.QueryNodeWaitProcessingMsgCount.WithLabelValues(paramtable.GetStringNodeID(), metrics.DeleteLabel).Inc()
 
--- a/internal/querynodev2/pipeline/insert_node_test.go
+++ b/internal/querynodev2/pipeline/insert_node_test.go
@@ -17,18 +17,29 @@
 package pipeline
 
 import (
+	"context"
+	"sync"
 	"testing"
+	"time"
 
+	"github.com/bytedance/mockey"
 	"github.com/samber/lo"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/mocks/util/mock_segcore"
 	"github.com/milvus-io/milvus/internal/querynodev2/delegator"
 	"github.com/milvus-io/milvus/internal/querynodev2/segments"
+	"github.com/milvus-io/milvus/internal/storage"
+	"github.com/milvus-io/milvus/internal/util/initcore"
+	"github.com/milvus-io/milvus/pkg/v3/mq/msgstream"
+	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
 	"github.com/milvus-io/milvus/pkg/v3/proto/querypb"
+	"github.com/milvus-io/milvus/pkg/v3/proto/segcorepb"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 )
 
@@ -80,8 +91,16 @@ func (suite *InsertNodeSuite) TestBasic() {
 		Segment:    mockSegmentManager,
 	}
 
+	var transferOrigin func(*schemapb.CollectionSchema, *msgstream.InsertMsg) (*segcorepb.InsertRecord, []int64, error)
+	transferMock := mockey.Mock(storage.TransferInsertMsgToInsertRecord).To(func(schema *schemapb.CollectionSchema, msg *msgstream.InsertMsg) (*segcorepb.InsertRecord, []int64, error) {
+		suite.True(collection.HasInsertSchemaTransitionReaderForTest())
+		return transferOrigin(schema, msg)
+	}).Origin(&transferOrigin).Build()
+	defer transferMock.UnPatch()
+
 	suite.delegator = delegator.NewMockShardDelegator(suite.T())
 	suite.delegator.EXPECT().ProcessInsert(mock.Anything).Run(func(insertRecords map[int64]*delegator.InsertData) {
+		suite.True(collection.HasInsertSchemaTransitionReaderForTest())
 		for segID := range insertRecords {
 			suite.True(lo.Contains(suite.insertSegmentIDs, segID))
 		}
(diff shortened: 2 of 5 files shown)


czs007:
## Review Summary

Three non-blocking observations from the adversarial review of the schema-transition locking change; all are acceptable trade-offs or documentation/naming suggestions, and none require changes before merge.

---

### Low

**`internal/querynodev2/pipeline/insert_node.go:135` — Read-lock critical section now spans the whole `ProcessInsert`, so a pending DDL can block subsequent inserts**

The `RLock` taken by `WithInsertSchemaTransition` now covers the entire closure (`insert_node.go:135-154`), not just payload conversion and the native insert. It also covers `segmentManager.Put`, `addGrowing`, distribution updates, and IDF-oracle bookkeeping in the tail of `ProcessInsert` (`delegator/delegator_data.go:166`, `:191`, `:222-228`). Because Go's `sync.RWMutex` blocks new readers once a writer is queued, one pending schema update will block all subsequent inserts on that collection — and this does cross shards: every shard's `insertNode` and `shardDelegator` resolve the same `*Collection` from `collectionManager` by `collectionID` (`insert_node.go:130`, `delegator.go:1500` → `:1533`), hence the same `schemaTransitionMu`. The impact is bounded to a single batch and isolated per collection, which is acceptable.

*Suggestion:* releasing the lock after `growing.Insert` but before the distribution/oracle bookkeeping is the right direction, but it is not directly actionable here — `ProcessInsert` is a single call on the `ShardDelegator` interface, so the pipeline side cannot release the lock partway through. Doing this would require a delegator API change (e.g. an epoch-boundary callback). Worth tracking as a follow-up rather than blocking this PR. *(raised by wzy)*

---

### Nitpick

**`internal/querynodev2/pipeline/insert_node.go:135` — Read lock has no context/timeout cancellation path; the DDL ↔ native-insert coupling is newly introduced here**

`WithInsertSchemaTransition` uses a bare `RLock()` with no context and no timeout (`segments/collection.go:482`), and `ProcessInsert` passes `context.Background()` to both `NewSegment` and `growing.Insert` (`delegator/delegator_data.go:166-167`, `:191`). If `NewSegment` (native mmap allocation) stalls, a schema update on that collection stalls for the same duration with no way to cancel. To be precise about attribution: `NewSegment` being uncancellable is pre-existing, but the coupling — DDL blocked by the insert path's native slow path — is new in this PR. Previously `UpdateSchema` only took `collectionManager.mut` and never interacted with `ProcessInsert`, so no lock coupling existed between the two; now it goes through `collection.applySchemaUpdate` → `lockSchemaTransitionForUpdate()` → `schemaTransitionMu.Lock()`, contending directly with the `RLock` at `insert_node.go:135`.

*Suggestion:* this is a deliberate cost paid for correctness (payload and growing segment sharing one schema epoch) and is an acceptable design trade-off. No change requested, but it is worth recording that DDL worst-case wait is now determined by the tail latency of the native insert path and is unbounded. *(raised by wzy)*

**`internal/querynodev2/segments/schema_transition_testonly.go:47` — `HasInsertSchemaTransitionReaderForTest` uses `TryLock` and cannot distinguish readers from writers**

*(Line correction: the function lives at `schema_transition_testonly.go:47-53`; the file is only 53 lines long, so the originally reported line 838 is not a valid anchor.)*

The function infers "a reader holds the lock" from a failed `TryLock`, but `TryLock` can equally fail because a *writer* holds it — the return value cannot tell the two apart. The doc comment (`:44-46`) already acknowledges this and delegates disambiguation to the caller ("Callers must invoke it before starting a schema writer"), which is a verbal contract rather than a compile-time guarantee: someone calling it while a writer is active will silently get a false-positive `true`. Both current call sites satisfy the precondition, so there is no live defect. Note also that the "briefly blocks real readers on the hot path" concern does not apply: a successful `TryLock` does momentarily take the write lock, but the file is gated by `//go:build test` (`:17`) and the symbol does not exist in production builds.

*Suggestion (optional):* promote the precondition from a comment to an assertion, or rename to `HasSchemaTransitionLockHolderForTest` so the name reflects the actual semantics. *(raised by wzy)*

sijie-ni-0214:
Thanks for the detailed adversarial review.

We agree that these are non-blocking trade-offs and will keep this targeted bug-fix PR unchanged.

1. The transition reader intentionally extends through ProcessInsert. The old payload must be converted, used to create the growing segment, and inserted under one native schema epoch. Narrowing the scope before the delegator tail would require splitting the ShardDelegator API or introducing an epoch-boundary callback, which is a broader follow-up than this fix.

2. A schema update can now wait for the native insert path. This is the deliberate correctness cost of preventing the old-payload/new-schema race. The native operations are already uncancellable; adding cancellation or a bounded wait needs a broader lifecycle/API design.

3. HasInsertSchemaTransitionReaderForTest is only called before a writer is started in both current tests. The documented precondition makes the TryLock result unambiguous there. We will retain the reader-specific name because the assertion is intentionally about the insert reader scope, not merely any lock holder.

We will track possible critical-section narrowing separately as a performance follow-up.

czs007:
## Overview

Two minor findings from the adversarial review — both are low-impact and neither blocks merging; they are recorded for the author's consideration.

### Low

**`internal/querynodev2/pipeline/insert_node.go:145` — Read lock spans `FillEmbeddingData`, widening the window in which writers are blocked**

Both `functionStore.OutputFieldIDs` (line 136) and `FillEmbeddingData` (line 145) now execute inside the `RLock` held by `WithInsertSchemaTransition` (`segments/collection.go:481-486`, where `schemaTransitionMu.RLock()` covers the entire callback). Go's `sync.RWMutex` is writer-preferring: once `schemaTransitionMu.Lock()` on the `UpdateSchema`/`PutOrRef` path (`collection.go:470-472`) queues up, insert readers for *other* shards of the same collection also block until the current reader releases, since the `Collection` object is shared across all shards of that collection within the querynode. On the cache-miss path, `FillEmbeddingData` calls `BuildEmbeddingRunners` and `FillFunctionFields` (`util/function/local_store.go:94-103`), so in the (uncommon) case where embeddings must actually be generated, one slow fill stalls both the schema transition and inserts on the collection's other shards.

*Suggestion:* No change requested. This is an accepted trade-off given how rarely schema transitions occur; logged here only for the record. If embedding-fill latency ever becomes a concern, consider computing the output field IDs and embedding data outside the read lock and re-validating the schema version afterwards. *(raised by wzy)*

### Nitpick

**`internal/querynodev2/pipeline/insert_node.go:49` — `addInsertData` redundantly re-reads the schema**

`addInsertData` calls `collection.Schema()` again at line 49 instead of using the `schema` argument already supplied by the `WithInsertSchemaTransition` callback, which is in scope in the closure at lines 135-150. There is no actual hazard: the read lock is held, and the only post-publication `setSchema` mutation (`segments/collection.go:466`, inside `applySchemaUpdateLocked`, whose callers hold the write lock at `collection.go:432-437` and `439-446`) cannot interleave, so both reads necessarily return the same snapshot. The remaining `setSchema` calls (`collection.go:638/651/664`) all occur during `NewCollection` construction, before publication.

*Suggestion:* Pass the closure's `schema` parameter through to `addInsertData` (line 150) and drop the re-read at line 49. This makes it structurally evident that the transfer uses the epoch-pinned schema rather than an independent — if coincidentally equal — read. *(raised by wzy, yezh)*

sijie-ni-0214:
Addressed in 04829c0166.

addInsertData now receives the schema captured by WithInsertSchemaTransition, instead of independently calling collection.Schema. This makes payload conversion, BM25 statistics, and primary-key extraction structurally use the same epoch-pinned schema.

Validated with make static-check and the full internal/querynodev2/pipeline test package.

czs007:
Reviewed the PR head at `04829c0` and confirmed the finding. Here is the comment body:

---

One non-blocking performance nit surfaced during adversarial review; no correctness issues were found.

## Low severity

**`internal/querynodev2/segments/collection.go:439`** — Write lock is acquired before the `shouldUpdate` check, so no-op load refreshes still block insert batches on that collection.

`applyLoadUpdate` (438-456) and `applySchemaUpdate` (431-436) take the `schemaTransitionMu` write lock before deciding whether anything needs to change — the decision itself happens inside the lock, in `applySchemaUpdateLocked` (458-462), and `prepareCollectionSchemaUpdate` only reads the `c.schema` atomic snapshot, so it could be pre-checked cheaply outside the lock. Since `PutOrRef` sits on the load/ref path (`services.go:231` WatchDmChannels, `services.go:534` LoadSegments, `delegator_data.go:606`), every LoadSegments RPC runs `applyLoadUpdate`. When `shouldUpdate` is false and the index meta is unchanged (`updateIndexMeta` returns early at 405 for `meta == nil`, and at 415 when `proto.Equal` holds), the call changes nothing yet still takes the write lock.

The impact is slightly stronger than a brief stall: Go's `RWMutex` blocks subsequent `RLock` calls once a writer is waiting, and the insert-side reader (`WithInsertSchemaTransition`, 481-486) spans the entire insert batch — payload conversion, `FillEmbeddingData`, and the native growing write in `ProcessInsert` (`insert_node.go` `Operate`). So a no-op refresh waits out the current batch and blocks the next one behind it.

This is correctness-safe and a reasonable trade-off from the migration off `m.mut` (the old code held a broader manager-level lock here, but did not block inserts).

Suggestion: if this shows up in an insert-latency profile, run `prepareCollectionSchemaUpdate` as a cheap pre-check before acquiring the write lock (re-validating once inside it), so genuine no-ops skip the lock acquisition that stalls readers. Note that the fast path in `applyLoadUpdate` must still perform `c.Ref(1)` (454) — its `putOrUpdateStorageContext` depends on the current schema properties — and preserve `updateIndexMeta` semantics.

Non-blocking.

*(raised by wzy)*

sijie-ni-0214:
Thanks @czs007, agreed.

This is a safe performance optimization, but it is not required for the correctness fix in this PR. In particular, a schema no-op on the existing PutOrRef path can still need to refresh index metadata and acquire the caller-visible collection reference, so it cannot simply bypass the transition path based only on shouldUpdate.

A useful fast path would need an outside pre-check plus in-lock revalidation, while preserving those index-meta and Ref semantics. We will keep this PR focused on the schema/insert race and track that optimization separately.

Issue #51485, opened 2026-07-16
[Bug]: GIS predicate returns misaligned results on the iterative-filter / rescore offset-input path


xiaocai2333:
**Is there an existing issue for this?**

- [x] I have searched the existing issues

**Environment**

- Milvus version: master
- Deployment mode: N/A (segcore query execution)
- MQ type: N/A
- SDK version: N/A
- OS: N/A
- CPU: N/A
- GPU: N/A
- Others: N/A

**Current Behavior**

`PhyGISFunctionFilterExpr` (the physical GIS filter expression) slices its output
solely by its own batch cursor (`GetNextBatchSize()`) and never reads the
offset-input list (`context.get_offset_input()`). However it inherits
`SegmentExpr::SupportOffsetInput()`, which defaults to `true`.

When a GIS predicate (`st_intersects` / `st_within` / `st_contains` / …) is
reached through the native offset-input path — i.e. `IterativeFilterNode` /
rescore, where an upstream operator supplies a sparse list of candidate offsets
and expects the expression to evaluate only those offsets and return a bitmap
aligned to that list — the GIS filter ignores the offset list and instead
returns a bitmap sliced by its own sequential batch cursor. The two are
misaligned, so the rows selected do not correspond to the requested offsets:
a silent wrong-results bug for GIS predicates evaluated under iterative
filtering.

**Expected Behavior**

A GIS predicate reached through the iterative-filter / rescore path must produce
results aligned to the query, not to an unrelated internal cursor. Since the GIS
filter cannot consume an external offset list, it should report
`SupportOffsetInput() == false` so the consumer takes its non-native fallback
(evaluate over the batch and intersect), which is already correct.

**Steps To Reproduce**

Construct a query where a GIS predicate is evaluated on the native offset-input
path (an iterative filter / rescore that supplies a candidate offset list to the
GIS expression). The returned selection is misaligned relative to the supplied
offsets.

**Root cause**

`PhyGISFunctionFilterExpr` does not override `SupportOffsetInput()`, so it
advertises offset-input support it does not implement.

**Fix**

Override `PhyGISFunctionFilterExpr::SupportOffsetInput()` to return `false`.

xiaocai2333:
has been fixed
/close

PR #50675, opened 2026-06-22
enhance: split GIS geometry filter into coarse/refine and fuse same-column predicates

xiaocai2333:
issue: #50674
design doc: `docs/design_docs/gis_filter_coarse_refine_split_fusion.md`

## What

Adds the GIS geometry filter **coarse/refine split + same-column fusion** optimization, gated by `queryNode.segcore.enableGISSplitFusion` (default `false`).

- Split each geometry column's predicates into two cooperating nodes sharing a per-segment `GISGroupState`:
  - **`PhyGISCoarseConjunctExpr`** — runs each predicate's R-Tree query once per segment, combines per AND/OR into `B_coarse`; scheduled **early** (`indexed_expr` bucket) so its bitmap prunes other predicates via `bitmap_input`.
  - **`PhyGISRefineConjunctExpr`** — scheduled **last** (`heavy` bucket); consumes `bitmap_input` (== scalars ∧ `B_coarse`) and, for each surviving row, constructs the row geometry **once** and evaluates **all** same-column predicates against it (K→1).
- Optimizer hook `ReorderConjunctExpr` → `SplitFuseGISConjunct` detects same-column geometry predicates under an AND conjunction (direct GIS leaves, or a pure same-field GIS sub-conjunction), builds the two nodes, and rewrites the conjunction inputs.
- New flag wired through `SegcoreConfig` + `segcore_init_c` + `initcore` + `paramtable`.

## Why

Filtered vector search with geometry predicates spends ~20% of segcore CPU **constructing/destructing GEOS objects** (per predicate, per row); the actual geometric computation is <1%. The current `PhyGISFunctionFilterExpr` also (1) buckets the expensive exact refine early, and (2) refines the *entire* coarse candidate set without consuming `bitmap_input`. See the issue/design doc for the profile and algebra (`coarse ⊇ exact` ⇒ `scalars ∧ B ≡ scalars ∧ B_coarse ∧ B_refine`).

## Tests

- `GISCoarseRefineExprTest.EquivalenceFusionOnVsOff` — data-segment path.
- `RTreeIndexTest.GIS_SplitFusion_Equivalence_Indexed` — R-Tree indexed path (exercises `RunRTreeQuery`).

Both assert the bitset with the flag **ON is identical to OFF** across: scalar + single GIS, scalar + (GIS OR GIS) (Shape B), same-field intersects + within, and a lone GIS predicate (fusion no-op). Full `RTreeIndexTest` suite and 113 GIS-related tests pass with no regressions.

[...]

PR #51486, opened 2026-07-16
fix: gate GIS filters out of the offset-input path

xiaocai2333:
issue: #51485, #51501

## What this PR does

**1. Gate GIS filters out of the offset-input path (`GISFunctionFilterExpr.h`).**
`PhyGISFunctionFilterExpr` slices its output only by its own batch cursor
(`GetNextBatchSize()`) and never reads `context.get_offset_input()`, yet it
inherited `SegmentExpr::SupportOffsetInput()`'s default of `true`. On the native
offset-input path (`IterativeFilterNode` / boost rescore) the consumer feeds a
sparse offset list and expects a bitmap aligned to it; the GIS filter ignores
the list and returns a cursor-sliced bitmap, so the returned bitmap is sized by
the expression batch (8192 rows by default) rather than by the offset list.

**What that actually did before this fix**, since it matters for backport
triage: every consumer asserts the two lengths match (`Scorer.cpp:39`,
`Scorer.cpp:98`, `IterativeFilterNode.cpp:337-338`) and `Assert` carries no
`NDEBUG` guard, so any normal topk — which is far below 8192 — failed **loudly
with an internal error**, in release builds too. Silent row misalignment was
the narrow case where `active_count == offsets.size()` made the lengths
coincide, not the common one. The bug is a hard query failure first and a
wrong-results bug only in that corner. The override makes the consumer take its
non-native fallback instead.

**2. Make the non-native rescore fallback cover every expression batch
(`BoostScoreRunner.cpp`).** Adversarial review of this PR surfaced that the
fallback it routes GIS into had a pre-existing defect: it called
`ExprSet::Eval` exactly once, but each `Eval` only advances the expression by
one internal batch (`DEFAULT_EXEC_EVAL_EXPR_BATCH_SIZE` = 8192 rows), while the
topk offsets being scored can reference any row of the segment. Offsets beyond
the first batch silently lost their boost (`WeightScorer`) or read past the end
of the bitset (`RandomScorer`). This is reachable on master today via
text/phrase-match filters in a boost scorer; this PR would have widened the
exposure to GIS, so the fallback now accumulates batches until the bitset
covers all active rows (mirroring `PhyIterativeFilterNode`) and folds the valid
(null) bitmap into the result.

**3. Add the missing bounds check in `RandomScorer::batch_score`
(`Scorer.cpp`).** The `TargetBitmap` overload indexed `bitmap[offset]` with no
guard; `WeightScorer`'s sibling overload already treats out-of-range offsets as
no-match. Defense-in-depth for bitmaps that legitimately lag the segment.

**4. Evaluate the non-native boost filter once per segment, not once per
offset chunk (`BoostScoreRunner.cpp` / `boost_score.cpp`).** Second-round
review caught that fix 2, as first written, made
`ComputeScorerScoresOnPreparedChunks` re-run the whole-segment evaluation once
per offset chunk (each call built a fresh `ExprSet` whose cursor restarts at
row 0) — `O(num_chunks × active_count)` predicate evaluations. The
whole-segment bitset is now hoisted into `ComputeNonNativeFilterBitset`,
computed once per (scorer, segment) before the chunk loop, and passed into
every per-chunk `ComputeScorerScores` call. Single-shot callers
(`PhyRescoresNode`) are unchanged: without a precomputed bitset the call still
evaluates the filter itself.

**5. Advance GIS filter cursors on growing segments during conjunct skip
(`GISFunctionFilterExpr.h`).** `PhyGISFunctionFilterExpr::MoveCursor()`
forwarded to the base implementation only for sealed segments and was a no-op
on growing segments. When a preceding conjunct input left a batch with no
active rows, `PhyConjunctFilterExpr::SkipFollowingExprs` advanced every
skipped expression solely through `MoveCursor()`, so on a growing segment with
more than one expression batch the GIS cursor fell behind its siblings and
later batches silently ANDed the wrong rows (or tripped the conjunct's result
assert once the cursor exhausted early). The sealed-only guard existed because
`SegmentExpr::MoveCursorForIndex()` asserts sealed-only; growing segments now
route through `MoveCursorForData()`, and the growing interim-index path also
advances the global index position, mirroring `EvalForIndexSegment`.

## Behavior change

GIS predicates on the offset-input path flip from the native path (fast, but
failing the query in the common case — see fix 1) to the non-native fallback
(correct), and the fallback itself now evaluates the whole segment's active
rows — exactly once per (scorer, segment) regardless of how many offset chunks
the caller splits the topk into — instead of only the first 8192 rows.

Element-level searches (emb-list / struct-array vector fields) are covered too.
`SearchResult::element_level_` is a property of the placeholder, not of the
filter expression, so routing GIS into the non-native fallback made
`search(<struct-array vector field>, filter="st_within(geo, ...)")` reach a
branch that rejected element-level outright — which would have swapped one
thrown `SegcoreError` for another rather than fixing anything for those
queries. That branch now maps each element back to its doc through the
segment-wide bitset and returns element-level results.

All fixes are unconditional correctness fixes, not gated behind any flag, and
are standalone backport candidates for 2.6 / 3.0.

## Performance note

Routing GIS off the offset-input path is a correctness/latency trade, stated
here explicitly rather than discovered in production: with
`SupportOffsetInput() == false`, `IterativeFilterNode` evaluates the GIS
predicate over the whole segment before intersecting with the search result,
where the native path would have evaluated only the topk candidate offsets.
For a 1M-row sealed segment with `topk=100` that is roughly four orders of
magnitude more GEOS predicate evaluations per query (fix 4 caps this at one
whole-segment scan per scorer per segment — the `num_chunks` multiplier is
gone). This is still the right trade, but for a narrower reason than "we were
returning wrong answers anyway": in the common case the pre-fix native path did
not return at all, it failed the query on a length assertion, so there was no
correct-and-fast baseline to give up. Queries that did return — the
`active_count == offsets.size()` corner — were the ones returning misaligned
rows. Either way GIS + iterative-filter / boost-rescore queries that previously
failed will now succeed and be measurably slower until the follow-up below
lands.

## Follow-up (tracked in #51582, not in this PR)

Native offset-input support for `PhyGISFunctionFilterExpr`: implement
offset-list evaluation the way sibling exprs do (`UnaryExpr`, `TermExpr`,
`BinaryRangeExpr`), then flip `SupportOffsetInput()` back to `true` to restore
topk-sized evaluation on the iterative-filter path.

## Test

- `ExprTest.GISFilterDoesNotSupportOffsetInput` compiles a bare GIS leaf into
  its physical expr and asserts `SupportOffsetInput() == false`, locking the
  contract. Verified red/green: reverting the override turns it red.
- `BoostScoreRunnerTest.ComputeScorerScoresNonNativeFilterCoversAllBatches`
  (text match) and `BoostScoreRunnerTest.ComputeScorerScoresGISFilterCoversAllBatches`
  (GIS) drive the rescore fallback with N=10000 rows (> one expression batch)
  and assert offsets past row 8192 still receive their boost. Verified
  red/green: with the batch-accumulation fix reverted, both fail exactly at the
  offsets beyond the first batch.
- `RandomScorerTest.BatchScoreTargetBitmapOutOfBoundsOffsets` documents the
  bounds-check contract (out-of-range offsets are skipped, no OOB access).
- `BoostScoreCTest.NonNativeFilterScoresAllChunks` drives the C-API path with
  multiple offset chunks and asserts every chunk scores against the same
  full-segment bitset; `BoostScoreRunnerTest.ComputeNonNativeFilterBitsetNulloptWithoutFilter`,
  `…NulloptForNativeFilter`, and `…PrecomputedFilterBitsetScoresChunksConsistently`
  pin the hoisted-bitset contract of fix 4.
- `ExprTest.GISFilterGrowingConjunctSkipKeepsCursorInSync` and
  `ExprTest.GISFilterGrowingIndexConjunctSkipKeepsCursorInSync` reproduce the
  conjunct-skip cursor desync on growing segments (> one batch, data path and
  interim-index path) and assert the GIS cursor stays aligned after fix 5.

## Why a standalone PR

Split out of #50675 (GIS coarse/refine split + fusion) at review request: that
PR is a default-off, flag-gated optimization, while everything here takes
effect unconditionally. Landing it on its own keeps the changelog, `git
bisect`, revert, and backport all granular. #50675 will rebase on top once this
merges.

### DIFF
--- a/internal/core/src/exec/expression/Expr.cpp
+++ b/internal/core/src/exec/expression/Expr.cpp
@@ -675,5 +675,49 @@ OptimizeCompiledExprs(ExecContext* context, const std::vector<ExprPtr>& exprs) {
     milvus::monitor::internal_core_optimize_expr_latency.Observe(cost / 1000);
 }
 
+TargetBitmap
+EvalExprSetOverAllBatches(ExprSet& expr_set,
+                          EvalCtx& eval_ctx,
+                          int64_t total_rows,
+                          const char* what) {
+    std::vector<VectorPtr> results;
+    TargetBitmap bitset;
+    TargetBitmap valid_bitset;
+    while (static_cast<int64_t>(bitset.size()) < total_rows) {
+        expr_set.Eval(0, 1, true, eval_ctx, results);
+
+        AssertInfo(results.size() == 1 && results[0] != nullptr,
+                   "{}: filter expr returned null result",
+                   what);
+        auto col_vec = std::dynamic_pointer_cast<ColumnVector>(results[0]);
+        if (col_vec == nullptr) {
+            ThrowInfo(
+                UnexpectedError, "{}: result should be ColumnVector", what);
+        }
+        if (!col_vec->IsBitmap()) {
+            ThrowInfo(UnexpectedError, "{}: result should be bitmap", what);
+        }
+        auto col_vec_size = col_vec->size();
+        AssertInfo(col_vec_size > 0,
+                   "{}: filter expr returned empty batch after {} of {} rows",
+                   what,
+                   bitset.size(),
+                   total_rows);
+        TargetBitmapView view(col_vec->GetRawData(), col_vec_size);
+        bitset.append(view);
+        TargetBitmapView valid_view(col_vec->GetValidRawData(), col_vec_size);
+        valid_bitset.append(valid_view);
+    }
+    AssertInfo(static_cast<int64_t>(bitset.size()) == total_rows,
+               "{}: filter bitset size {} must match total rows {}",
+               what,
+               bitset.size(),
+               total_rows);
+    // Fold UNKNOWN (NULL) into FALSE explicitly (data &= valid) instead of
+    // relying on the convention that UNKNOWN rows carry data=0.
+    bitset.inplace_and(valid_bitset, bitset.size());
+    return bitset;
+}
+
 }  // namespace exec
 }  // namespace milvus
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -2714,5 +2714,17 @@ class ExprSet {
 expr::TypedExprPtr
 CreateTTLFieldFilterExpression(QueryContext* query_context);
 
+// Evaluate expr_set batch by batch until the accumulated result covers
+// total_rows, then fold UNKNOWN (NULL) into FALSE (data &= valid) so a
+// consumer that reads only the data bits is null-rejecting by construction.
+// Shared by the whole-range consumers of expressions that cannot take offset
+// input: PhyIterativeFilterNode's non-native branch and boost scoring's
+// non-native filter fallback. `what` names the caller in error messages.
+TargetBitmap
+EvalExprSetOverAllBatches(ExprSet& expr_set,
+                          EvalCtx& eval_ctx,
+                          int64_t total_rows,
+                          const char* what);
+
 }  //namespace exec
 }  // namespace milvus
--- a/internal/core/src/exec/expression/ExprGISMiscTest.cpp
+++ b/internal/core/src/exec/expression/ExprGISMiscTest.cpp
@@ -921,3 +921,122 @@ TEST(ExprTest, GISFilterDoesNotSupportOffsetInput) {
     ASSERT_EQ(expr_set.exprs().size(), 1u);
     EXPECT_FALSE(expr_set.exprs()[0]->SupportOffsetInput());
 }
+
+namespace {
+
+// Build a growing segment where `age` equals the row index and the geometry
+// sits inside the query polygon only for rows [8192, 16384), then evaluate
+// `age >= 8192 AND st_within(geo, ...)`. The first expression batch [0, 8192)
+// leaves the AND with no active rows, so the conjunct skips the GIS
+// expression through MoveCursor(); a skipped batch that does not advance the
+// GIS cursor makes every later batch evaluate the wrong rows.
+void
+RunGrowingConjunctSkipTest(bool with_rtree_index) {
+    using namespace milvus;
+    using namespace milvus::segcore;
+
+    auto schema = std::make_shared<Schema>();
+    auto pk_fid = schema->AddDebugField("pk", DataType::INT64);
+    auto age_fid = schema->AddDebugField("age", DataType::INT64);
+    auto geo_fid = schema->AddDebugField("geo", DataType::GEOMETRY);
+    schema->AddDebugField(
+        "vec", DataType::VECTOR_FLOAT, 16, knowhere::metric::L2);
+    schema->set_primary_field_id(pk_fid);
+
+    IndexMetaPtr index_meta = nullptr;
+    if (with_rtree_index) {
+        std::map<std::string, std::string> index_params = {
+            {"index_type", "RTREE"}};
+        std::map<std::string, std::string> type_params;
+        FieldIndexMeta geo_index_meta(
+            geo_fid, std::move(index_params), std::move(type_params));
+        std::map<FieldId, FieldIndexMeta> field_map = {
+            {geo_fid, geo_index_meta}};
+        index_meta =
+            std::make_shared<CollectionIndexMeta>(100000, std::move(field_map));
+    }
+
+    const int64_t N = 20000;  // three expression batches of 8192
+    const int64_t kBatch = 8192;
+    auto raw_data = DataGen(schema, N);
+
+    proto::schema::FieldData* age_field_data = nullptr;
+    proto::schema::FieldData* geo_field_data = nullptr;
+    for (auto& fd : *raw_data.raw_->mutable_fields_data()) {
+        if (fd.field_id() == age_fid.get()) {
+            age_field_data = &fd;
+        } else if (fd.field_id() == geo_fid.get()) {
+            geo_field_data = &fd;
+        }
+    }
+    ASSERT_NE(age_field_data, nullptr);
+    ASSERT_NE(geo_field_data, nullptr);
+
+    auto* age_col =
+        age_field_data->mutable_scalars()->mutable_long_data()->mutable_data();
+    for (int64_t i = 0; i < N; ++i) {
+        age_col->at(i) = i;
+    }
+
+    auto* geo_col = geo_field_data->mutable_scalars()->mutable_geometry_data();
+    geo_col->clear_data();
+    auto ctx = GEOS_init_r();
+    for (int64_t i = 0; i < N; ++i) {
+        const char* wkt = (i >= kBatch && i < 2 * kBatch)
+                              ? "POINT (0.5 0.5)"
+                              : "POINT (100.0 100.0)";
+        Geometry geom(ctx, wkt);
+        geo_col->add_data(geom.to_wkb_string());
+    }
+    GEOS_finish_r(ctx);
+
+    auto config = SegcoreConfig::default_config();
+    config.set_enable_interim_segment_index(with_rtree_index);
+    auto seg = CreateGrowingSegment(schema, index_meta, 1, config);
+    seg->PreInsert(N);
+    seg->Insert(0,
+                N,
+                raw_data.row_ids_.data(),
+                raw_data.timestamps_.data(),
+                raw_data.raw_);
+    ASSERT_EQ(seg->HasIndex(geo_fid), with_rtree_index);
+
+    proto::plan::GenericValue val;
+    val.set_int64_val(kBatch);
+    auto age_expr = std::make_shared<milvus::expr::UnaryRangeFilterExpr>(
+        milvus::expr::ColumnInfo(age_fid, DataType::INT64),
+        proto::plan::OpType::GreaterEqual,
+        val);
+    auto gis_expr = std::make_shared<milvus::expr::GISFunctionFilterExpr>(
+        milvus::expr::ColumnInfo(geo_fid, DataType::GEOMETRY),
+        proto::plan::GISFunctionFilterExpr_GISOp_Within,
+        "POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))");
+    auto filter = std::make_shared<milvus::expr::LogicalBinaryExpr>(
+        milvus::expr::LogicalBinaryExpr::OpType::And, age_expr, gis_expr);
+    auto plan = std::make_shared<milvus::plan::FilterBitsNode>(
+        DEFAULT_PLANNODE_ID, filter);
+
+    auto final =
+        milvus::query::ExecuteQueryExpr(plan, seg.get(), N, MAX_TIMESTAMP);
+    ASSERT_EQ(final.size(), N);
+    for (int64_t i = 0; i < N; ++i) {
+        bool expected = (i >= kBatch && i < 2 * kBatch);
+        ASSERT_EQ(final[i], expected)
+            << "cursor desync at row " << i
+            << " (with_rtree_index: " << with_rtree_index << ")";
+    }
+}
+
+}  // namespace
+
+// Raw-data path: growing segment without any geometry index.
+TEST(ExprTest, GISFilterGrowingConjunctSkipKeepsCursorInSync) {
+    RunGrowingConjunctSkipTest(false);
+}
+
+// Interim-index path: growing segment with an R-Tree interim index, where
+// EvalForIndexSegment advances the global index position together with the
+// data cursor and MoveCursor() must do the same on a skipped batch.
+TEST(ExprTest, GISFilterGrowingIndexConjunctSkipKeepsCursorInSync) {
+    RunGrowingConjunctSkipTest(true);
+}
--- a/internal/core/src/exec/expression/ExprGISMiscTest.cpp
+++ b/internal/core/src/exec/expression/ExprGISMiscTest.cpp
@@ -1001,6 +1001,32 @@ RunGrowingConjunctSkipTest(bool with_rtree_index) {
                 raw_data.raw_);
     ASSERT_EQ(seg->HasIndex(geo_fid), with_rtree_index);
 
+    {
+        // HasIndex() alone does not guarantee the ScalarIndex exec path —
+        // DetermineExecPath() silently falls back to RawData when the pin
+        // comes up empty (Expr.h). Assert the resolved path directly so the
+        // with_rtree_index variant fails loudly instead of silently
+        // degrading into a duplicate of the raw-data variant.
+        milvus::expr::TypedExprPtr gis_leaf =
+            std::make_shared<milvus::expr::GISFunctionFilterExpr>(
+                milvus::expr::ColumnInfo(geo_fid, DataType::GEOMETRY),
+                proto::plan::GISFunctionFilterExpr_GISOp_Within,
+                "POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))");
+        std::vector<milvus::expr::TypedExprPtr> leaf_filters{gis_leaf};
+        auto leaf_query_context = std::make_shared<milvus::exec::QueryContext>(
+            DEAFULT_QUERY_ID, seg.get(), N, MAX_TIMESTAMP);
+        milvus::exec::ExecContext leaf_exec_context(leaf_query_context.get());
+        milvus::exec::ExprSet leaf_expr_set(leaf_filters, &leaf_exec_context);
+        ASSERT_EQ(leaf_expr_set.exprs().size(), 1u);
+        auto segment_expr =
+            std::dynamic_pointer_cast<milvus::exec::SegmentExpr>(
+                leaf_expr_set.exprs()[0]);
+        ASSERT_NE(segment_expr, nullptr);
+        ASSERT_EQ(segment_expr->UseIndexCursor(), with_rtree_index)
+            << "GIS expr resolved to the wrong exec path (with_rtree_index: "
+            << with_rtree_index << ")";
+    }
+
     proto::plan::GenericValue val;
     val.set_int64_val(kBatch);
     auto age_expr = std::make_shared<milvus::expr::UnaryRangeFilterExpr>(
--- a/internal/core/src/exec/expression/ExprGISMiscTest.cpp
+++ b/internal/core/src/exec/expression/ExprGISMiscTest.cpp
@@ -41,6 +41,7 @@
 #include "common/Types.h"
 #include "common/protobuf_utils.h"
 #include "exec/QueryContext.h"
+#include "exec/expression/Expr.h"
 #include "expr/ITypeExpr.h"
 #include "geos_c.h"
 #include "gtest/gtest.h"
@@ -881,4 +882,42 @@ TEST_P(ExprTest, TestFloatingPointModulo) {
         }
         ASSERT_EQ(final.count(), expected_count);
     }
+}
+
+// Regression for the GIS offset-input contract. PhyGISFunctionFilterExpr slices
+// only by its own batch cursor (GetNextBatchSize) and never reads the
+// offset-input list, but the SegmentExpr base defaults SupportOffsetInput() to
+// true. Left at the default, the native offset-input path (IterativeFilterNode
+// / rescore) would feed a sparse offset list into an Eval that ignores it and
+// return misaligned rows -- a silent wrong-results bug on any GIS predicate
+// reached through iterative filtering. The override must report false so the
+// consumer takes its non-native fallback. This pins the contract so a future
+// change cannot regress it.
+TEST(ExprTest, GISFilterDoesNotSupportOffsetInput) {
+    auto schema = std::make_shared<Schema>();
+    auto pk_fid = schema->AddDebugField("pk", DataType::INT64);
+    auto geo_fid = schema->AddDebugField("geo", DataType::GEOMETRY);
+    schema->AddDebugField(
+        "vec", DataType::VECTOR_FLOAT, 16, knowhere::metric::L2);
+    schema->set_primary_field_id(pk_fid);
+
+    const int64_t N = 256;
+    auto dataset = DataGen(schema, N);
+    auto seg = CreateSealedWithFieldDataLoaded(schema, dataset);
+
+    // Compile a bare GIS leaf into its physical expr and inspect the contract.
+    milvus::expr::TypedExprPtr leaf =
+        std::make_shared<milvus::expr::GISFunctionFilterExpr>(
+            milvus::expr::ColumnInfo(geo_fid, DataType::GEOMETRY),
+            proto::plan::GISFunctionFilterExpr_GISOp_Intersects,
+            "POLYGON((-5 -5, 5 -5, 5 5, -5 5, -5 -5))");
+    std::vector<milvus::expr::TypedExprPtr> filters{leaf};
+
+    auto query_context = std::make_shared<milvus::exec::QueryContext>(
+        DEAFULT_QUERY_ID, seg.get(), N, MAX_TIMESTAMP);
+    milvus::exec::ExecContext exec_context(query_context.get());
+    milvus::exec::ExprSet expr_set(filters, &exec_context);
+
+    ASSERT_EQ(expr_set.exprs().size(), 1u);
+    EXPECT_FALSE(expr_set.exprs()[0]->SupportOffsetInput());
 }
--- a/internal/core/src/exec/expression/ExprGISMiscTest.cpp
+++ b/internal/core/src/exec/expression/ExprGISMiscTest.cpp
@@ -61,6 +61,7 @@
 #include "segcore/SegmentInterface.h"
 #include "segcore/SegmentSealed.h"
 #include "test_utils/DataGen.h"
+#include "test_utils/SegcoreConfigUtils.h"
 #include "test_utils/storage_test_utils.h"
 
 EXPR_TEST_INSTANTIATE();
@@ -990,6 +991,10 @@ RunGrowingConjunctSkipTest(bool with_rtree_index) {
     }
     GEOS_finish_r(ctx);
 
+    // SegcoreConfig members are process-global (inline static): the copy
+    // below does not isolate the interim-index toggle, so restore on scope
+    // exit instead of leaking it into every later growing-segment test.
+    ScopedSegcoreConfigRestore config_restore;
     auto config = SegcoreConfig::default_config();
     config.set_enable_interim_segment_index(with_rtree_index);
     auto seg = CreateGrowingSegment(schema, index_meta, 1, config);
--- a/internal/core/src/exec/expression/GISFunctionFilterExpr.h
+++ b/internal/core/src/exec/expression/GISFunctionFilterExpr.h
@@ -13,6 +13,7 @@
 
 #include <fmt/core.h>
 #include <stdint.h>
+#include <algorithm>
 #include <memory>
 #include <optional>
 #include <string>
@@ -83,11 +84,25 @@ class PhyGISFunctionFilterExpr : public SegmentExpr {
         return false;
     }
 
+    // A skipped batch (conjunct short-circuit via SkipFollowingExprs) must
+    // still advance this expression's cursors, otherwise it desynchronizes
+    // from its sibling expressions and later batches evaluate the wrong rows.
+    // The base MoveCursor() covers every case except the growing interim-index
+    // path: MoveCursorForIndex() asserts sealed-only, while
+    // EvalForIndexSegment() on a growing segment advances the global index
+    // position together with the data cursor -- mirror that here.
     void
     MoveCursor() override {
-        if (segment_->type() == SegmentType::Sealed) {
-            SegmentExpr::MoveCursor();
+        if (has_offset_input_ || execute_all_at_once_) {
+            return;
         }
+        if (UseIndexCursor() && segment_->type() != SegmentType::Sealed) {
+            current_index_chunk_pos_ +=
+                std::min(active_count_ - current_index_chunk_pos_, batch_size_);
+            MoveCursorForData();
+            return;
+        }
+        SegmentExpr::MoveCursor();
     }
 
  private:
--- a/internal/core/src/exec/expression/GISFunctionFilterExpr.h
+++ b/internal/core/src/exec/expression/GISFunctionFilterExpr.h
@@ -74,6 +74,15 @@ class PhyGISFunctionFilterExpr : public SegmentExpr {
         return fmt::format("{}", expr_->ToString());
     }
 
+    // The GIS filter slices by its own batch cursor (GetNextBatchSize) and never
+    // reads the offset-input list, so it cannot serve the offset-input
+    // (iterative-filter / rescore) path. Report false so IterativeFilterNode
+    // takes its non-native fallback instead of feeding offsets into Eval.
+    bool
+    SupportOffsetInput() override {
+        return false;
+    }
+
     void
     MoveCursor() override {
         if (segment_->type() == SegmentType::Sealed) {
--- a/internal/core/src/exec/expression/GISFunctionFilterExpr.h
+++ b/internal/core/src/exec/expression/GISFunctionFilterExpr.h
@@ -91,6 +91,11 @@ class PhyGISFunctionFilterExpr : public SegmentExpr {
     // path: MoveCursorForIndex() asserts sealed-only, while
     // EvalForIndexSegment() on a growing segment advances the global index
     // position together with the data cursor -- mirror that here.
+    // Unlike the base implementation, MoveCursorForData() is called without a
+    // HasFieldData() guard: this exec path already walks data chunks
+    // unconditionally (EvalForIndexSegment), and with no field data
+    // num_data_chunk_ is 0, making the call a no-op -- the omission is
+    // intentional, not an oversight.
     void
     MoveCursor() override {
         if (has_offset_input_ || execute_all_at_once_) {
--- a/internal/core/src/exec/operator/IterativeFilterNode.cpp
+++ b/internal/core/src/exec/operator/IterativeFilterNode.cpp
@@ -235,12 +235,22 @@ PhyIterativeFilterNode::GetOutput() {
                         auto [doc_id, elem_idx] =
                             array_offsets->ElementIDToRowID(element_id);
                         element_to_doc_mapping.push_back({doc_id, elem_idx});
-                        unique_doc_ids.insert(doc_id);
                     }
 
-                    doc_offsets.reserve(unique_doc_ids.size());
-                    for (auto doc_id : unique_doc_ids) {
-                        doc_offsets.emplace_back(static_cast<int32_t>(doc_id));
+                    // The deduplicated doc offsets feed only the native
+                    // offset-input Eval below; the non-native fallback
+                    // indexes the segment-wide bitset by doc id directly
+                    // and never reads eval_offsets.
+                    if (is_native_supported_) {
+                        for (const auto& [doc_id, elem_idx] :
+                             element_to_doc_mapping) {
+                            unique_doc_ids.insert(doc_id);
+                        }
+                        doc_offsets.reserve(unique_doc_ids.size());
+                        for (auto doc_id : unique_doc_ids) {
+                            doc_offsets.emplace_back(
+                                static_cast<int32_t>(doc_id));
+                        }
                     }
                     eval_offsets = &doc_offsets;
                 } else {
--- a/internal/core/src/exec/operator/IterativeFilterNode.cpp
+++ b/internal/core/src/exec/operator/IterativeFilterNode.cpp
@@ -308,8 +308,30 @@ PhyIterativeFilterNode::GetOutput() {
                             }
                         }
                     }
+                } else if (element_level) {
+                    // element_level_ is a property of the placeholder, not of
+                    // the filter expression, so an element-level search whose
+                    // filter cannot consume offset input (text match, GIS)
+                    // legitimately reaches this fallback. The segment-wide
+                    // bitset is indexed by doc offset and already holds the
+                    // verdict for every doc, so no doc_eval_cache is needed --
+                    // map each element back to its doc and test that bit.
+                    for (size_t i = 0; i < offsets.size(); ++i) {
+                        auto [doc_id, elem_idx] = element_to_doc_mapping[i];
+                        if (bitset[doc_id] > 0) {
+                            topk_binsert(search_result,
+                                         nq_index * unity_topk,
+                                         topk,
+                                         large_is_better,
+                                         distances[i],
+                                         doc_id,
+                                         elem_idx);
+                            if (topk == unity_topk) {
+                                break;
+                            }
+                        }
+                    }
                 } else {
-                    Assert(!element_level);
                     for (auto i = 0; i < offsets.size(); ++i) {
                         if (bitset[offsets[i]] > 0) {
                             topk_binsert(search_result,
--- a/internal/core/src/exec/operator/IterativeFilterNode.cpp
+++ b/internal/core/src/exec/operator/IterativeFilterNode.cpp
@@ -117,43 +117,12 @@ PhyIterativeFilterNode::GetOutput() {
     // get bitset of whole segment first
     if (!is_native_supported_) {
         EvalCtx eval_ctx(operator_context_->get_exec_context());
-
-        TargetBitmap valid_bitset;
-        while (num_processed_rows_ < need_process_rows_) {
-            exprs_->Eval(0, 1, true, eval_ctx, results_);
-
-            AssertInfo(
-                results_.size() == 1 && results_[0] != nullptr,
-                "PhyIterativeFilterNode result size should be size one and not "
-                "be nullptr");
-
-            if (auto col_vec =
-                    std::dynamic_pointer_cast<ColumnVector>(results_[0])) {
-                if (col_vec->IsBitmap()) {
-                    auto col_vec_size = col_vec->size();
-                    TargetBitmapView view(col_vec->GetRawData(), col_vec_size);
-                    bitset.append(view);
-                    TargetBitmapView valid_view(col_vec->GetValidRawData(),
-                                                col_vec_size);
-                    valid_bitset.append(valid_view);
-                    num_processed_rows_ += col_vec_size;
-                } else {
-                    ThrowInfo(UnexpectedError,
-                              "PhyIterativeFilterNode result should be bitmap");
-                }
-            } else {
-                ThrowInfo(
-                    UnexpectedError,
-                    "PhyIterativeFilterNode result should be ColumnVector");
-            }
-        }
-        Assert(bitset.size() == need_process_rows_);
-        Assert(valid_bitset.size() == need_process_rows_);
-        // Rows are included below on the data bit alone, so fold UNKNOWN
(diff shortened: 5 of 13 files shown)


xiaocai2333:
rerun ut

czs007:
## Summary

This PR correctly fixes the `SupportOffsetInput()` contract for GIS expressions, but routing GIS into the non-native fallback exposes a pre-existing batching defect in `PhyRescoresNode` and an unguarded bitmap index in `RandomScorer` that should be addressed alongside it.

---

### High

**`internal/core/src/exec/operator/RescoresNode.cpp:171` — Rescore non-native fallback only evaluates the first batch**

The non-native branch calls `expr_set->Eval(0, 1, true, eval_ctx, results)` exactly once and builds the bitmap from that single result. Compare `IterativeFilterNode.cpp:159-186`, which loops `while (num_processed_rows_ < need_process_rows_)` and then asserts `bitset.size() == need_process_rows_`. Since GIS sizes its result from its own batch cursor (`GISFunctionFilterExpr.cpp:220,224-225`; the index path does the same at `:403`), and `EXEC_EVAL_EXPR_BATCH_SIZE` defaults to 8192 (`Consts.h:87`), one `Eval` only covers rows `[0, 8192)`. The scorer then indexes that bitmap by segment offset (`bitmap[offsets[i]]`), and search-result offsets can point anywhere in the segment. For a segment with `active_count > 8192` and a candidate offset past the first batch, `WeightScorer::batch_score` (`Scorer.cpp:54-66`) bounds-checks and silently treats a matching row as no-match — losing its boost and corrupting ranking — while `RandomScorer::batch_score` (`Scorer.cpp:131-132`) reads out of range.

To be clear about scope: this defect is **not** specific to GIS and is **not** introduced by this PR. It is pre-existing and already affects the other non-native exprs. TextMatch/PhraseMatch are not safe here either — `ExecTextMatch` returns the whole cached bitmap only under `execute_all_at_once_` (`UnaryExpr.cpp:2068`), and the only caller of `SetExecuteAllAtOnce` is `FilterBitsNode.cpp:148-150`. `RescoresNode` never sets it, so `ExecTextMatch` falls through to the sliced path at `UnaryExpr.cpp:2075-2084` and is equally truncated to one batch.

**Suggestion:** This PR does not create the bug, but it routes a second expr family (GIS) into the same broken fallback, so it is worth fixing here: accumulate all batches in the `RescoresNode` non-native branch (mirroring `IterativeFilterNode.cpp:159-186`) before indexing by segment offset.

*(raised by wzy, yezh)*

---

**`internal/core/src/rescores/Scorer.cpp:131` — `RandomScorer::batch_score` (TargetBitmap overload) indexes the bitmap without a bounds check**

`if (bitmap[offsets[i]] > 0)` reads `bitmap[offset]` with no guard, so any offset `>= bitmap.size()` (e.g. `>= 8192` when the bitmap came from a single cursor-sliced `Eval` in the `RescoresNode` non-native branch) is an out-of-range read on the `TargetBitmap` — undefined behavior and a potential crash. The sibling `WeightScorer` TargetBitmap overload at `Scorer.cpp:54-66` already performs exactly this check (`if (offset >= 0 && static_cast<size_t>(offset) < bitmap_size)`, treating out-of-range as no-match), so the asymmetry is unambiguous.

**Suggestion:** Guard the index in `RandomScorer::batch_score`, mirroring the `WeightScorer` overload. This is worth doing regardless of how the `RescoresNode` batching issue is resolved.

*(raised by wzy)*

---

### Medium

**`internal/core/src/exec/expression/ExprGISMiscTest.cpp:971` — Test only pins the `SupportOffsetInput()` contract bit; no end-to-end regression exercises the rescore path**

The test compiles a bare GIS leaf via `ExprSet` and asserts `SupportOffsetInput() == false`, which usefully prevents a silent re-enable. However, it uses `N = 256` (a single batch) and never constructs a `PhyIterativeFilterNode` or `PhyRescoresNode`, so it exercises neither consumer of the contract bit it pins. It therefore cannot catch the wrong-result / out-of-bounds behaviour of the fallback the PR now routes GIS into.

**Suggestion:** Add a rescore regression with `N > 8192` — or lower `EXEC_EVAL_EXPR_BATCH_SIZE`, as `LikeConjunctExprTest.cpp:89` does with `EXEC_EVAL_EXPR_BATCH_SIZE.store(3)` — and a GIS scorer filter whose selected offset lands past the first evaluation batch.

*(raised by wzy, yezh)*

---

### Low

**`internal/core/src/exec/operator/IterativeFilterNode.cpp:155` — Gating GIS off the offset-input path makes every iterative-filter GIS query evaluate GEOS predicates over the whole segment**

The contract fix is correct, but the cost is worth stating explicitly in the PR rather than discovering in production. With `SupportOffsetInput() == false`, `is_native_supported_` is false and `IterativeFilterNode` takes the loop at `:155-188`, evaluating the GIS predicate over the entire segment (`need_process_rows_` rows) before doing anything with the search result. The native path at `:282-285` would have fed only the candidate offsets (topk-sized) into `Eval`. For a 1M-row sealed segment with `topk=100`, that is roughly four orders of magnitude more GEOS predicate evaluations per query, and GIS predicates are not cheap (`PreparedGeometry` construction plus a per-row op at `GISFunctionFilterExpr.cpp:428-459`).

This is the right trade — the pre-fix native path was silently wrong, so no correct-and-fast baseline is being given up.

**Suggestion:** A native offset-input implementation for `PhyGISFunctionFilterExpr` is the proper follow-up and should be tracked. The expr currently never calls `SetHasOffsetInput` at all, unlike every sibling (`UnaryExpr.cpp:184`, `TermExpr.cpp:65`, `BinaryRangeExpr.cpp:57`, …).

*(raised by zack)*

czs007:
Adversarial review surfaced two medium-severity issues: a performance regression in the non-native fallback path, and a pre-existing cursor-desync defect in the same GIS expression file this PR touches.

---

Both findings below are medium severity; one concerns the per-chunk cost of the new fallback loop, the other a latent cursor bug in the GIS expression class this PR edits.

## Medium

**`internal/core/src/rescores/BoostScoreRunner.cpp:108` — Non-native fallback re-scans the whole segment once per offset chunk**

The new batch-accumulation loop lives in `ComputeScorerScores`, which `ComputeScorerScoresOnPreparedChunks` invokes once per offset chunk (`segcore/boost_score.cpp:161-183`). Each invocation builds a fresh `ExprSet` (`BoostScoreRunner.cpp:66`), which resets the batch cursor to row 0, so the `while (bitset.size() < active_count)` loop re-evaluates the entire segment on every chunk. The cost is `O(num_chunks × active_count)` predicate evaluations rather than `O(active_count)`. Since `internal/util/segcore/boost_score.go:88` fans the call out over `len(offsets.Chunks())`, `num_chunks > 1` is the expected case, not an edge case. Before this PR the fallback only evaluated the first 8192 rows per chunk, so the per-chunk work grows by roughly `active_count/8192`.

Text match is shielded here because `ExecTextMatch` caches its full-segment bitset via `exec::ExprCacheHelper::GetOrCompute` under the expression signature (`UnaryExpr.cpp:2137-2172`). `PhyGISFunctionFilterExpr::EvalForDataSegment` (`GISFunctionFilterExpr.cpp:220-228`) has no equivalent cache, so every chunk pays a full GEOS pass over the segment. Concretely: a boost-scorer query with an `ST_Within` filter over a 1M-row sealed segment split into K offset chunks evaluates the GEOS predicate K × 1M times per segment per scorer.

Suggestion: hoist the whole-segment bitset out of the per-chunk loop — compute it once per (scorer, segment) in `ComputeScorerScoresOnPreparedChunks` and pass it into `ComputeScorerScores` — or route GIS through `ExprCacheHelper` the way text match already does. Worth noting that the PR description's perf estimate ("roughly four orders of magnitude more GEOS predicate evaluations") accounts for a single whole-segment scan and omits the `num_chunks` factor. *(raised by N1)*

**`internal/core/src/exec/expression/GISFunctionFilterExpr.h:87` — `MoveCursor()` is a no-op on growing segments, desynchronizing the cursor after a conjunct short-circuit**

This is pre-existing rather than introduced here, but it sits directly below the new `SupportOffsetInput()` override and belongs to the same family as the issue above: the GIS batch cursor not honoring its consumer's contract.

`PhyGISFunctionFilterExpr::MoveCursor()` forwards to `SegmentExpr::MoveCursor()` only for sealed segments (`GISFunctionFilterExpr.h:86-92`). When a preceding input leaves a batch with no active rows, `PhyConjunctFilterExpr::Eval` calls `SkipFollowingExprs(i + 1)` (`ConjunctExpr.cpp:202-205`), and that path advances each skipped input solely through `MoveCursor()` (`ConjunctExpr.cpp:83-88`). On a growing segment the GIS cursor therefore stays put while every sibling expression advances.

Failure scenario: a growing segment with more than one batch (> 8192 rows) and a filter such as `int_field > X AND ST_Within(geo, POLYGON(...))`. On a batch where `int_field > X` matches nothing, the conjunct takes the `active_rows.none()` branch and skips the GIS expression without advancing it. On the next batch the GIS expression re-evaluates rows `[0, 8192)` while the conjunct ANDs that bitmap against the sibling's rows `[8192, 16384)` — silently wrong rows. Once the GIS cursor falls far enough behind to exhaust itself, `EvalForDataSegment` returns `nullptr` on a batch the conjunct still expects, tripping its result `AssertInfo`.

The sealed-only guard is presumably there because `SegmentExpr::MoveCursor()` routes to `MoveCursorForIndex()`, which asserts `index mode only for sealed segment` (`Expr.h:405-407`). The fix is to select the data cursor on growing segments (`MoveCursorForData`, `Expr.h:426-428`, is valid there) rather than to skip cursor movement entirely. *(raised by N2)*

xiaocai2333:
All findings from both review rounds are now addressed in this PR (head `5ce25d0`, CI green). Resolution map:

**Round 1**

- **High — `RescoresNode` non-native fallback only evaluates the first batch** → fixed in f4a713b858: the fallback now accumulates `ExprSet::Eval` batches until the bitset covers all active rows (mirroring `PhyIterativeFilterNode`) and folds the valid bitmap into the result. Covered by `BoostScoreRunnerTest.ComputeScorerScoresNonNativeFilterCoversAllBatches` (text match) and `…GISFilterCoversAllBatches` (GIS), both red/green verified against the reverted fix.
- **High — `RandomScorer::batch_score` missing bounds check** → fixed in f4a713b858: the `TargetBitmap` overload now guards `offset >= 0 && offset < bitmap.size()`, mirroring the `WeightScorer` overload (out-of-range = no-match). Pinned by `RandomScorerTest.BatchScoreTargetBitmapOutOfBoundsOffsets`.
- **Medium — contract-bit test doesn't exercise either consumer** → addressed in 3942f86bf0: the rescore-fallback regressions above use N=10000 rows (> one 8192-row expression batch) with boosted offsets past the first batch, exercising the actual consumer path rather than only the contract bit.
- **Low — state the perf cost, track native offset-input follow-up** → the PR body now has an explicit "Performance note" (≈4 orders of magnitude more GEOS evaluations on a 1M-row segment at topk=100), and the native offset-input implementation for `PhyGISFunctionFilterExpr` is tracked as #51582.

**Round 2**

- **Medium — non-native fallback re-scans the whole segment once per offset chunk** → fixed in 8c53c9595f: the whole-segment bitset is hoisted into `ComputeNonNativeFilterBitset`, computed once per (scorer, segment) before the chunk loop in `ComputeScorerScoresOnPreparedChunks`, and passed into every per-chunk call — `O(active_count)` regardless of `num_chunks`. Single-shot callers (`PhyRescoresNode`) are unchanged. Covered by `BoostScoreCTest.NonNativeFilterScoresAllChunks` plus three `ComputeNonNativeFilterBitset` contract tests. The PR body's perf estimate has been corrected to note the `num_chunks` multiplier is gone.
- **Medium — `MoveCursor()` no-op on growing segments desyncs the cursor after conjunct skip** → fixed in 5ce25d0e74: growing segments now route through `MoveCursorForData()` (valid there; the sealed-only guard existed because `MoveCursorForIndex()` asserts sealed), and the growing interim-index path also advances the global index position, mirroring `EvalForIndexSegment`. Covered by `ExprTest.GISFilterGrowingConjunctSkipKeepsCursorInSync` and `…GrowingIndexConjunctSkipKeepsCursorInSync`.

PR description has been updated to document fixes 4 and 5 and the new tests.

czs007:

czs007:
## Overview

Adversarial review surfaced three issues in this PR: one residual correctness gap in fix 1's fallback path, plus two accuracy/efficiency observations worth addressing before backport.

---

### Medium

**`internal/core/src/exec/operator/IterativeFilterNode.cpp:357`**

**Problem.** Fix 1 sets `PhyGISFunctionFilterExpr::SupportOffsetInput()` to `false`, which routes GIS predicates into `IterativeFilterNode`'s non-native fallback. That fallback is only implemented for row-level searches — its first statement is `Assert(!element_level)`. But `element_level_` is a property of the *placeholder* (emb-list / struct-array vector search — see `VectorSearchNode.cpp:168`, `SearchOnSealed.cpp:95`), not of the filter expression, so `search(<struct-array vector field>, filter="st_within(geo, ...)")` on the iterative-filter path is a legal plan that now lands in the throwing branch. Note that `Assert` in `common/EasyAssert.h` has no `NDEBUG` guard, so this throws `SegcoreError` in release builds too. For element-level searches the PR does not fix anything — it swaps one thrown `SegcoreError` for another. Neither the "Behavior change" nor the "Performance note" section discloses this residual hole, and there is no test covering `element_level` + GIS.

**Suggestion.** Either implement the element-level case in the non-native fallback, or explicitly reject GIS + element-level earlier with a clear user-facing error; and document the limitation in the PR body. Please add a test covering `element_level` + GIS either way.

---

### Low

**`internal/core/src/rescores/Scorer.cpp:39`**

**Problem.** The PR justifies fix 1 as "a silent wrong-results bug for GIS predicates reached through iterative filtering", and the Performance note builds on that ("the pre-fix native path returned silently wrong results, so no correct-and-fast baseline is given up"). That is not the dominant pre-fix behaviour. `PhyGISFunctionFilterExpr` never calls `SetHasOffsetInput()` and sizes its output with `GetNextBatchSize()` (`GISFunctionFilterExpr.cpp:221`, `:404`), so on the offset-input path it returns a bitmap of one *expression* batch (8192 rows by default), while every consumer asserts the bitmap length equals the offset-list length (`Scorer.cpp:39`, `Scorer.cpp:98`, `IterativeFilterNode.cpp:337-338`). Since `Assert` is unconditional, any normal topk (≪ 8192) makes the pre-fix query fail loudly with an internal error rather than return misaligned rows. Silent misalignment requires `active_count == offsets.size()` — a coincidence, not the common case. The fix itself is correct; only the stated failure mode and severity are off, which matters for the 2.6/3.0 backport triage this PR explicitly requests.

**Suggestion.** Reword the PR description (and the Performance note) to describe the pre-fix behaviour as a hard `SegcoreError` in the common case, with silent misalignment as the narrow `active_count == offsets.size()` corner, so backport reviewers triage the severity correctly.

**`internal/core/src/rescores/BoostScoreRunner.cpp:118`**

**Problem.** `ComputeNonNativeFilterBitset()` builds a full `ExprSet` — which compiles the expression tree and, via `DetermineExecPath`/`EnsurePinnedIndex`, can pin scalar indexes — purely to answer "does every expr support offset input?", then discards it whenever the answer is yes. For a native filter, the per-chunk `ComputeScorerScores()` call (`BoostScoreRunner.cpp:156`) then rebuilds its own `ExprSet` for each chunk, so an N-chunk request goes from N to N+1 `ExprSet` constructions. Fix 4's premise is that per-chunk repetition of segment-level work is a defect worth a dedicated commit; the same argument applies to the `ExprSet` itself.

**Suggestion.** Return the built `ExprSet` (or at least the native/non-native verdict) from `ComputeNonNativeFilterBitset` and thread it into `ComputeScorerScores`, making the native path N−1 constructions cheaper instead of one more expensive.

xiaocai2333:
All findings from the 2026-07-21 review round are addressed. The branch has also been rebased onto latest master (`bd3ea0493b`), so the SHAs below are post-rebase; CI is re-running on the new head.

**Medium — element-level searches land in `Assert(!element_level)` in the non-native fallback** → fixed in c5f2893da5. The non-native fallback now implements the element-level case: each element is mapped back to its parent doc through the segment-wide bitset and element-level results are returned instead of throwing. Tests in 6b5a35c3fa drive `search(<struct-array vector field>, filter="st_within(...)")` through the iterative-filter path end to end (`test_element_filter.cpp`). The "Behavior change" section of the PR body now documents element-level coverage.

**Low — PR description mischaracterized the pre-fix failure mode** → PR body reworded. It now states the dominant pre-fix behaviour was a hard `SegcoreError` from the unconditional length `Assert` (release builds included), with silent misalignment only in the narrow `active_count == offsets.size()` corner, and the Performance note was rebuilt on that basis for backport triage.

**Low — `ComputeNonNativeFilterBitset` builds an `ExprSet` and discards it on the native path** → fixed in 068ab7b51e. The compiled `ExprSet` is handed back through an out-param and reused by every per-chunk `ComputeScorerScores` call, making the native path N−1 constructions cheaper instead of one more expensive. Reuse safety (offset-input is reset from arguments on every Eval; `MoveCursor` is a no-op while offset input is set) is documented in `BoostScoreRunner.h`.

---

On the earlier `ci-v2/e2e-default` failure: `test_struct_array_compaction_preserves_null_empty_offsets_and_indexes` failed at `assert compact_id > 0` with `-1`. `-1` means DataCoord's force trigger synchronously produced **zero compaction tasks** (`services.go` sets `CompactionID = -1` when `getCompactionTasksNumBySignalID(id) == 0`; manual candidates require `IsSorted`, i.e. sort compaction to have completed). That is a Go-side timing condition this PR — which touches only the C++ query path — does not reach. The struct-array search/element-search assertions *before* the compact call, which do exercise this PR's code, all passed.

czs007:
## Overview

Adversarial review surfaced three issues in this PR: one medium-severity consistency gap in the new NULL-folding logic, plus two low-severity items covering redundant per-batch work and a regression test that can silently stop testing its target branch.

---

### Medium

**`internal/core/src/rescores/BoostScoreRunner.cpp:192` — native boost branch skips the NULL fold added on the sibling branch**

*Problem:* This PR adds `bitset.inplace_and(valid_bitset, bitset.size())` to the non-native branch of `ComputeScorerScores` (line 105), citing parity with `PhyIterativeFilterNode`. The native offset-input branch of the same function passes `col_vec->GetRawData()` directly to `batch_score()` without ANDing `GetValidRawData()`, whereas the operator it mirrors folds valid on *both* branches (`IterativeFilterNode.cpp:264-269`). NULL policy in boost scoring therefore depends on whether the filter expression happens to implement `SupportOffsetInput()`: `weight(filter="a > 5")` (native) would boost an UNKNOWN-but-data-set row, while `weight(filter="text_match(t,'x')")` (non-native) would not. This is currently unreachable thanks to the leaf convention (`Expr.h:96-120`, `ValueOp.h:24-30`), which makes it a latent asymmetry rather than an observable wrong answer — but that is precisely the argument the PR rejected when adding the fold on the other branch.

*Suggestion:* Apply the same fold on the native branch — construct a `TargetBitmapView` over `col_vec->GetValidRawData()` and `inplace_and` it into `bitsetview` before calling `batch_score()`. Separately, consider whether `BoostScoreRunner.cpp:44-46` should build its `ExprSet` with `/*null_rejecting=*/true` as `IterativeFilterNode.cpp:70-71` does.

*(raised by adversarial reviewer — correctness/consistency)*

---

### Low

**`internal/core/src/exec/operator/IterativeFilterNode.cpp:238` — new element-level fallback still pays for the unused doc dedup**

*Problem:* The `unique_doc_ids` hash-set and `doc_offsets` vector (lines 232-247) exist only to build `eval_offsets` for the native offset-input `Eval`. The new `else if (element_level)` fallback (lines 311-322) never reads `eval_offsets`, `doc_offsets`, or `doc_eval_cache` — it indexes the segment-wide `bitset` by `doc_id` from `element_to_doc_mapping` — yet the block runs unconditionally on every iterator batch. For a struct-array vector search with `hints=iterative_filter` and a non-native predicate, each batch of up to `unity_topk` elements does one `unordered_set::insert` per element plus a full set walk, then discards the result. Before this PR the branch was `Assert(!element_level)`, so the overhead is new rather than pre-existing.

*Suggestion:* Guard the dedup block so it only runs when the native offset-input path will actually consume `eval_offsets` (e.g. hoist it under the same condition that selects that branch), leaving the element-level fallback to build only `element_to_doc_mapping`.

*(raised by adversarial reviewer — efficiency)*

---

**`internal/core/src/exec/expression/ExprGISMiscTest.cpp:1077` — index-path regression test can silently degrade to the raw-data path**

*Problem:* `GISFilterGrowingIndexConjunctSkipKeepsCursorInSync` targets exactly one new branch: `MoveCursor()`'s `UseIndexCursor() && type() != Sealed` path, reachable only when `exec_path_ == ExprExecPath::ScalarIndex`. Its only guard is `ASSERT_EQ(seg->HasIndex(geo_fid), with_rtree_index)`, but `SegmentExpr::DetermineExecPath()` documents in-code (`Expr.h:2459-2467`) that `HasIndex()==true` does not imply the ScalarIndex path — an empty `PinIndex()` silently falls back to `RawData`. Since the final-bitmap assertion is satisfied identically by the raw-data path already covered by the sibling test, any future change that empties the geometry interim-index pin (mid-load state, interim-index build policy, a `DetermineExecPath` tweak) would leave both `RunGrowingConjunctSkipTest(false)` and `(true)` green while the branch under test goes uncovered.

*Suggestion:* Assert the executed path directly rather than inferring it from `HasIndex()` — expose or query the expression's resolved `exec_path_` (or add a counter/flag on the index-cursor branch) and assert `ScalarIndex` in the `with_rtree_index=true` case.

*(raised by adversarial reviewer — test coverage)*

xiaocai2333:
All three findings from the 2026-07-22 review round are addressed on top of `6b5a35c3fa` (whose CI loop was fully green). Point by point:

> **`internal/core/src/rescores/BoostScoreRunner.cpp:192` — native boost branch skips the NULL fold added on the sibling branch**
>
> The native offset-input branch of the same function passes `col_vec->GetRawData()` directly to `batch_score()` without ANDing `GetValidRawData()`, whereas the operator it mirrors folds valid on *both* branches […] NULL policy in boost scoring therefore depends on whether the filter expression happens to implement `SupportOffsetInput()`.

Fixed in d622954382. The native branch now ANDs `GetValidRawData()` into the bitmap before `batch_score()`, matching the non-native branch and both branches of `PhyIterativeFilterNode`; `BuildFilterExprSet` also builds with `/*null_rejecting=*/true` for the same reason `IterativeFilterNode` does. As the review notes, today's leaf convention makes this a latent asymmetry rather than an observable wrong answer, so the new `BoostScoreRunnerTest.NativeFilterGivesNullRowsNoBoost` (nullable field, native filter, NULL rows interleaved) pins the contract rather than reproducing a live bug — stated as such in the commit message.

> **`internal/core/src/exec/operator/IterativeFilterNode.cpp:238` — new element-level fallback still pays for the unused doc dedup**
>
> The `unique_doc_ids` hash-set and `doc_offsets` vector […] exist only to build `eval_offsets` for the native offset-input `Eval`. The new `else if (element_level)` fallback […] never reads `eval_offsets`, `doc_offsets`, or `doc_eval_cache` — yet the block runs unconditionally on every iterator batch.

Fixed in 74cc86f183. `unique_doc_ids` / `doc_offsets` are now populated only when `is_native_supported_`; the non-native element-level fallback builds only `element_to_doc_mapping` and indexes the segment-wide bitset by doc id directly. Behaviour-preserving; the existing `ElementFilter*` suites (incl. the element-level GIS coverage) stay green.

> **`internal/core/src/exec/expression/ExprGISMiscTest.cpp:1077` — index-path regression test can silently degrade to the raw-data path**
>
> Its only guard is `ASSERT_EQ(seg->HasIndex(geo_fid), with_rtree_index)`, but […] `HasIndex()==true` does not imply the ScalarIndex path — an empty `PinIndex()` silently falls back to `RawData`. […] any future change that empties the geometry interim-index pin would leave both `RunGrowingConjunctSkipTest(false)` and `(true)` green while the branch under test goes uncovered.

Fixed in 315526dc83. `RunGrowingConjunctSkipTest` now compiles a bare GIS leaf against the same segment and asserts `UseIndexCursor() == with_rtree_index`, so an emptied interim-index pin fails the `with_rtree_index=true` variant loudly instead of leaving it a green duplicate of the raw-data variant.

---

Verified locally: `BoostScoreRunnerTest.* : RandomScorerTest.* : WeightScorerTest.* : BoostScoreCTest.* : ExprTest.GISFilter* : ElementFilter*` — 94/94 passed under ASAN.

czs007:
Adversarial review surfaced two non-blocking observations — one observability gap around silently-skipped out-of-bounds offsets, and one comment/clarity nit on the new growing-index cursor override; neither blocks merge.

## Low

**`internal/core/src/rescores/Scorer.cpp:137` — out-of-bounds offsets are silently dropped in the `TargetBitmap` overload**

The new bounds check in `RandomScorer::batch_score(TargetBitmap)` mirrors the pre-existing one in `WeightScorer` (`Scorer.cpp:60`, already on master), so the direction is right. Worth noting, though, that this silent skip is exactly the mechanism that masked the bug this PR fixes: before the fix the non-native branch did a single `Eval` + single `append`, so the bitset covered only one batch (8192 rows), and any `offset >= 8192` made `offset < bitmap_size` false — the boost was dropped silently instead of failing loudly. The new test (`test_boost_score_c.cpp`, `NonNativeFilterScoresAllChunks`, N=10000, offsets 9000/9001/9998) exercises precisely that path.

After the fix, `EvalFilterOverAllBatches` hard-asserts `bitset.size() == active_count` (`BoostScoreRunner.cpp:101`), so the bitmap side can no longer be short. The offset side is still unchecked: the cgo entry point `BuildScorerOffsets` only asserts `offset >= 0` and `offset <= INT32_MAX` (`boost_score.cpp:59-64`) and never compares against `active_count`, so a stale or cross-snapshot offset from the Go side still reaches this branch — and is still swallowed.

*Suggestion:* add a debug-only `Assert` (or a counter/log) on the out-of-bounds path so this class of desync — a future bitset-size regression, or a genuinely out-of-range offset from Go — becomes observable rather than a silently lost score.

*(raised by tinswzy)*

## Nitpick

**`internal/core/src/exec/expression/GISFunctionFilterExpr.h:102` — growing-index `MoveCursor` calls `MoveCursorForData()` unconditionally, unlike the base class**

The base `SegmentExpr::MoveCursor()` guards `MoveCursorForData()` with `if (segment_->HasFieldData(field_id_))` (`Expr.h:423-425`), while the new growing-index branch here calls it unconditionally. The behavior is safe: (1) the `EvalForIndexSegment()` growing branch it mirrors already walks data chunks unconditionally and reads `segment_->chunk_size(field_id_, i)` (`GISFunctionFilterExpr.cpp:601-624`), so that exec path already assumes raw data is present; and (2) if `HasFieldData` were false, `num_data_chunk_` stays 0 (`Expr.h:311-317`) and the cursor loop iterates zero times, making the call a no-op. So this is not a bug.

The existing comment above the function only explains why the base `MoveCursorForIndex()` can't be used (it asserts sealed-only) — it doesn't explain why dropping the `HasFieldData` guard is intentional, which reads like an accidental omission to a later reader.

*Suggestion:* add one line of comment noting that the guard is unnecessary here because `num_data_chunk_` is 0 when no field data exists, making `MoveCursorForData()` a no-op.

*(raised by tinswzy)*

xiaocai2333:
Both observations from this round are addressed:

> **`internal/core/src/rescores/Scorer.cpp:137` — out-of-bounds offsets are silently dropped in the `TargetBitmap` overload**
>
> […] this silent skip is exactly the mechanism that masked the bug this PR fixes […] The offset side is still unchecked: the cgo entry point `BuildScorerOffsets` only asserts `offset >= 0` and `offset <= INT32_MAX` and never compares against `active_count`, so a stale or cross-snapshot offset from the Go side still reaches this branch — and is still swallowed.

Fixed in 7bb2a4d1b8. Both `TargetBitmap` overloads (`WeightScorer` and `RandomScorer`) now count out-of-bounds offsets and emit one `LOG_WARN` per `batch_score` call — carrying the skipped count, the offsets size, and the bitmap size — while keeping the skip semantics (out of bounds ⇒ no boost) unchanged. A future bitset-size regression or a genuinely out-of-range offset from the Go side now leaves a signal instead of a silently lost score.

> **`internal/core/src/exec/expression/GISFunctionFilterExpr.h:102` — growing-index `MoveCursor` calls `MoveCursorForData()` unconditionally, unlike the base class** […] it doesn't explain why dropping the `HasFieldData` guard is intentional, which reads like an accidental omission to a later reader.

Fixed in 33066c38f2 — the comment now states explicitly that the guard is unnecessary here (the mirrored exec path already walks data chunks unconditionally, and with no field data `num_data_chunk_` is 0 so the call is a no-op), so the omission reads as intentional.

Verified locally: same 94-test filter (`BoostScoreRunnerTest.* : RandomScorerTest.* : WeightScorerTest.* : BoostScoreCTest.* : ExprTest.GISFilter* : ElementFilter*`) — 94/94 passed under ASAN after the change.

czs007:
## Adversarial Review Summary

One low-severity test-coverage gap was confirmed; no correctness defects survived verification.

### Low

**`internal/core/src/segcore/boost_score.cpp:181` — cross-chunk `ExprSet` reuse is never exercised on a `ScalarIndex`-backed filter**

*Problem.* The change hoists a single compiled `ExprSet` out of the per-chunk loop (`boost_score.cpp:181-183`) and reuses it for every offset chunk (`:208`). Reuse is safe by construction: offset-input evaluation never touches the stateful `ScalarIndex` cursor, because the index branch is gated on `!has_offset_input_` (`UnaryExpr.cpp:256`, `:1516`) and `SegmentExpr::MoveCursor()` is a whole-body no-op while offset input is set (`Expr.h:414-431`). That invariant is not pinned by any test. The new `NativeFilterHandsBackReusableExprSet` (`test_scorer.cpp:658`) filters `FieldId(100)` with `GreaterEqual`, but `FieldId(100)` is the schema primary key (`test_scorer.cpp:308-314`), so `DetermineExecPath()` resolves to `ExprExecPath::PkIndex` (`UnaryExpr.cpp:1988-1992`), not `ScalarIndex`. `PkIndex` has no separate index cursor (`UseIndexCursor()` is true only for `ScalarIndex`, `Expr.h:2430-2433`) and is itself skipped under offset input (`UnaryExpr.cpp:1508`), so the test never reaches the one exec path where cross-chunk cursor state could desync. A sweep of `internal/core/unittest` found no test combining a `ScalarIndex`-path expression with offset-input evaluation — `test_iterative_filter.cpp`'s `SealedIndex` case loads only a *vector* index, leaving its scalar fields on the raw-data path.

*Suggestion.* Add a variant of `NativeFilterHandsBackReusableExprSet` that loads a real scalar index on the filtered field, asserts `UseIndexCursor() == true`, and then reuses the `ExprSet` across chunks.

*Why low.* The same reuse pattern already ships in production: `PhyIterativeFilterNode` builds `exprs_` once (`IterativeFilterNode.cpp:70`) and re-`Eval`s it with offset input on every iterator batch (`:261-263`). This PR adds a new caller of an existing pattern rather than introducing new risk, so this is a coverage gap, not a defect.

*(raised by tinswzy)*

xiaocai2333:
Addressed:

> **`internal/core/src/segcore/boost_score.cpp:181` — cross-chunk `ExprSet` reuse is never exercised on a `ScalarIndex`-backed filter**
>
> The new `NativeFilterHandsBackReusableExprSet` filters `FieldId(100)`, but `FieldId(100)` is the schema primary key, so `DetermineExecPath()` resolves to `ExprExecPath::PkIndex` […] not `ScalarIndex`. […] A sweep of `internal/core/unittest` found no test combining a `ScalarIndex`-path expression with offset-input evaluation.

Fixed in e7e12c4df3 — `BoostScoreRunnerTest.NativeFilterExprSetReuseOnScalarIndexPath` loads a real STL_SORT index on a non-pk int64 field, asserts the compiled expression actually resolves to the ScalarIndex path via `UseIndexCursor()` (so the variant fails loudly if it ever silently degrades to RawData/PkIndex, same guard style the earlier round asked of the GIS growing-index test), then reuses one `ExprSet` across offset chunks straddling the 8192 expression-batch boundary — checking both the semantic expectations (rows below the threshold unboosted, above boosted) and exact agreement with freshly compiled `ExprSet`s.

The branch was also rebased onto latest master (`49535465f1`) before the commit, per our workflow; prior commits are unchanged in content, only re-based (new SHAs).

Verified locally: the affected filter (`BoostScoreRunnerTest.* : RandomScorerTest.* : WeightScorerTest.* : BoostScoreCTest.* : ExprTest.GISFilter* : ElementFilter*`) — 95/95 passed under ASAN.

czs007:
Adversarial review surfaced two non-blocking low-severity issues in the boost-scoring path — one around cancellation ordering and per-chunk pointer validation, and one around warning-log amplification during expected index lag.

### Low severity

- **`internal/core/src/segcore/boost_score.cpp:182` — Filter bitset built before cancellation and per-chunk output-pointer checks.** In `ComputeScorerScoresOnPreparedChunks`, the whole-segment `ComputeNonNativeFilterBitset` build is hoisted above the loop's first `throwIfCancelled` (line 187) and above the per-chunk `AssertInfo(output_score_chunks[chunk_idx] != nullptr)` / `output_has_score_chunks[chunk_idx] != nullptr` checks (lines 192/195). The `Eval` path used by `ComputeNonNativeFilterBitset` has no in-loop cancellation check, so a pre-cancelled request runs the full non-native segment scan before cancellation is observed, and a null per-element output pointer on the first non-empty chunk is only caught after that scan. (Array-level pointers are already validated at the C boundary in `ValidateScorerScoreInputs`.) *Suggestion:* add a single preflight pass — `throwIfCancelled` plus a per-chunk output-pointer check over the non-empty chunks — before building the hoisted bitset, which closes both gaps. (raised by tinswzy, chyezh)

- **`internal/core/src/rescores/Scorer.cpp:73` — `LOG_WARN` amplification during growing-index lag.** `WeightScorer::batch_score` (warning at line 74) and `RandomScorer::batch_score` (warning at line 165) already aggregate all out-of-bounds offsets within a call into one warning, but `batch_score` is invoked once per offset chunk via `ComputeScorerScores`. The out-of-bounds condition is the text-index-lags-vector-index race that the surrounding comment marks as expected/transient, so during sustained lag the warning fires once per chunk × segment × query. *Suggestion:* aggregate to one warning per segment/request, or emit a rate-limited metric that carries the segment id, instead of a per-chunk `LOG_WARN`. (raised by tinswzy, chyezh)

xiaocai2333:
Both findings addressed:

> **`internal/core/src/segcore/boost_score.cpp:182` — Filter bitset built before cancellation and per-chunk output-pointer checks.** […] a pre-cancelled request runs the full non-native segment scan before cancellation is observed, and a null per-element output pointer on the first non-empty chunk is only caught after that scan.

Fixed in 9b47988d58, exactly as suggested: a single preflight — `throwIfCancelled` plus an output-pointer sweep over the non-empty chunks — now runs before `ComputeNonNativeFilterBitset`, and the now-redundant in-loop asserts are dropped (the in-loop `throwIfCancelled` stays for long chunk loops). Traced by hand: a pre-cancelled token throws `FutureCancellation` at the preflight, before any `QueryContext`/`ExprSet` construction or segment scan; the existing C-API tests (`NonNativeFilterScoresAllChunks`, `AllEmptyChunksSkipNonNativeFilter`, `ZeroChunksSucceedsWithoutOutputBuffers`) pin that behavior is otherwise unchanged.

> **`internal/core/src/rescores/Scorer.cpp:73` — `LOG_WARN` amplification during growing-index lag.** […] during sustained lag the warning fires once per chunk × segment × query. *Suggestion:* aggregate to one warning per segment/request, or emit a rate-limited metric that carries the segment id.

Fixed in c9bb452c9c: the warning is now throttled to at most one per 10s process-wide (atomic CAS on a monotonic timestamp, no locks on the hot path) and carries the segment id. Per-call aggregation of the skipped count and the skip semantics (out of bounds ⇒ no boost) are unchanged, so the existing out-of-bounds contract tests still pass as-is.

Verified locally: the affected filter — 95/95 passed under ASAN.

czs007:
Adversarial review surfaced two low-severity, non-blocking observations in the scoring path — both are behavior-preservation/observability nits rather than functional defects.

### Low severity

- **`internal/core/src/rescores/Scorer.cpp:46` — Out-of-bounds warning throttle is process-wide and drops suppressed counts.** The `static std::atomic<int64_t> last_log_us` in `ShouldLogOutOfBoundsOffsets()` is shared across every `WeightScorer::batch_score` and `RandomScorer::batch_score` call (all segments, queries, and both scorer types), so at most one line prints process-wide per 10s. Because `out_of_bounds` is a per-call local that is discarded whenever the throttle returns false, the surviving log line reports only the passing call's count, not the aggregate — contradicting the comment at Scorer.cpp:37-42. As a result, a sustained systemic short-bitset (text-index desync) looks like one stray line instead of a storm. *Suggestion:* accumulate suppressed counts into the next emitted message, or expose a metric, so transient lag is distinguishable from real desync. *(raised by tinswzy, chyezh)*

- **`internal/core/src/segcore/boost_score.cpp:151` — Cancellation check skipped on the all-empty fast path.** The new `if (!has_offsets) return;` fast path (boost_score.cpp:151-153) runs before the preflight `throwIfCancelled(cancel_token)` (boost_score.cpp:160). Previously, an already-cancelled request with a non-zero set of all-empty chunks fell into the per-chunk loop and threw on the first iteration's cancellation check; with this PR the same request returns success. Practical impact is negligible (the work is a no-op and the caller discards the result), so this is a semantics-preservation nit. *Suggestion:* move `throwIfCancelled(cancel_token)` above the fast-path return to restore prior cancellation behavior at no cost. *(raised by tinswzy, chyezh)*

xiaocai2333:
Both observations addressed:

> **`internal/core/src/rescores/Scorer.cpp:46` — Out-of-bounds warning throttle is process-wide and drops suppressed counts.** […] Because `out_of_bounds` is a per-call local that is discarded whenever the throttle returns false, the surviving log line reports only the passing call's count, not the aggregate — contradicting the comment at Scorer.cpp:37-42.

Fixed in ec8a83f5c9. `ShouldLogOutOfBoundsOffsets()` is now `AccumulateOutOfBoundsForLog()`: suppressed calls `fetch_add` their count into a process-wide accumulator, and the next emitted warning drains it with `exchange(0)`, reporting the total skipped since the last report alongside the emitting call's own numbers (count/size/bitmap/segment). A sustained systemic desync now shows up as a large accumulated total instead of one stray line, and the comment matches the implementation.

> **`internal/core/src/segcore/boost_score.cpp:151` — Cancellation check skipped on the all-empty fast path.** […] *Suggestion:* move `throwIfCancelled(cancel_token)` above the fast-path return to restore prior cancellation behavior at no cost.

Fixed in f2f2f7e86c, exactly as suggested: `throwIfCancelled` now runs before the `has_offsets` fast-path return; the output-pointer preflight stays after it (all-empty inputs have no pointers to validate).

Verified locally: the affected filter — 95/95 passed under ASAN.

czs007:
Adversarial review surfaced two low-severity issues — one behavioral edge case around cancellation and one code-duplication cleanup — neither blocking but both worth addressing.

### Low

- **`internal/core/src/segcore/boost_score.cpp:151` — All-empty request returns Success on a pre-canceled token.** The new all-empty early return (`if (!has_offsets) return;`) sits before the preflight `throwIfCancelled(cancel_token)` at line 160. On master, `throwIfCancelled` ran at the top of every chunk-loop iteration *before* the `if (scorer_offsets.empty()) continue;` check, so an all-empty chunk set with an already-canceled token threw `Cancelled`; now the early return fires first and such a request returns `Success`. Impact is low (cancellation is best-effort, the caller discards results on cancel, and an all-empty request does no work regardless). Suggestion: move `throwIfCancelled(cancel_token)` above the `!has_offsets` early return to preserve the prior cancellation semantics. *(raised by chyezh, tinswzy)*

- **`internal/core/src/rescores/BoostScoreRunner.cpp:68` — Duplicated whole-segment null-folding loop.** `EvalFilterOverAllBatches` (loop at line 77) and the `!is_native_supported_` loop in `IterativeFilterNode.cpp:118-156` are independent copies of the same accumulate-then-`inplace_and` null-rejecting contract: loop calling `Eval(0, 1, true, ...)`, cast to `ColumnVector`, `bitset.append(view)` + `valid_bitset.append(valid_view)`, then `bitset.inplace_and(valid_bitset, ...)` to fold `UNKNOWN → FALSE`. They differ only in the loop bound and the error-check shape. This is a maintainability concern only. Suggestion: consolidate the shared logic into a helper in a follow-up. *(raised by tinswzy)*

xiaocai2333:
Latest adversarial round surfaced two low-severity items; disposition:

**1. "All-empty request returns Success on a pre-canceled token" (`boost_score.cpp:151`)** — already fixed before this round in f2f2f7e86c (the previous round raised the same finding): `throwIfCancelled` runs before the `!has_offsets` fast-path return on the current head. The round evidently reviewed the prior head; no further change needed.

**2. Duplicated whole-segment null-folding loop (`BoostScoreRunner.cpp` vs `IterativeFilterNode.cpp`)** — consolidated in 22e09940da rather than deferred: the accumulate-then-`inplace_and` contract now lives in one helper, `exec::EvalExprSetOverAllBatches` (Expr.h/Expr.cpp), and both `PhyIterativeFilterNode`'s non-native branch and boost scoring's fallback are thin wrappers over it. Error shapes are the union of the two former copies (non-ColumnVector / non-bitmap keep `ThrowInfo(ExprInvalid)`; null result / empty batch / coverage keep `AssertInfo`), so no consumer-visible classification changes. Existing tests pin both callers (`IterativeFilter*`, `BoostScoreRunnerTest.*`, `BoostScoreCTest.*`); CI on this head will provide the full verification pass.

CI note: the previous head `ec8a83f` (cancellation-preflight + warning-accumulation fixes) passed the full CI loop green.

czs007:

xiaocai2333:
Root cause of the two `ci-v2/build-ut-cov` failures (19208/19278, "shard1 completed with 0 tests captured") identified and fixed — it is **not** a defect in this PR's changes:

- `SegcoreConfig`'s members are `inline static` (process-global). A test added on master in #51540 sets `chunk_rows = 2` via a seeming copy and never restores it; every later growing segment in the process is then built with 2-row chunks, and `ExprTestSuite/ExprTest.TestTermJsonNullable` (100k-row JSON growing segment; the nullable path walks chunks O(n) under a lock per access) goes from ~1.5s to ~120s under coverage.
- gtest sharding is deterministic per test set. This PR adds ~10 C++ tests, which re-rolls the shard layout; in the resulting layout the poison lands in shard1 before nine heavy Nullable instances with no `chunk_rows`-restoring test in between → the shard deterministically exceeds the 1800s budget on every run. Master and PRs that do not change the C++ test set inherit master's currently-safe layout (its shard happens to run a restoring test right after the poison), which is why they pass. Minimal reproduction with two unmodified master tests, verified on a binary containing none of this PR's changes: see #51799.

Fix ships as a standalone PR #51800 (wraps the three `set_chunk_rows(2)` sites in the existing `ScopedSegcoreConfigRestore` guard; red/green verified). **Stacked on #51800: until it merges, this branch carries its commit** (d8ba04fa41) so CI here is unblocked; the commit will be dropped on the post-merge rebase. This branch also scopes its own GIS tests' interim-index toggle with the same guard (8c032ccaad).

PR #51487, opened 2026-07-16
fix: [2.6] gate GIS filters out of the offset-input path

xiaocai2333:
issue: #51485, #51501
pr: #51486

Backport of #51486, adapted to the 2.6 architecture. Three commits:

**1. Gate GIS filters out of the offset-input path.**
`PhyGISFunctionFilterExpr` slices only by its own batch cursor and never reads
the offset-input list, but inherited `SegmentExpr::SupportOffsetInput()`'s
default of `true` — so a GIS predicate on the native offset-input path
(iterative filter / rescore) returns rows misaligned to the requested offsets
(silent wrong results). Override it to `false` so the consumer takes its
correct non-native fallback. Carries the same standalone regression test
(`ExprTest.GISFilterDoesNotSupportOffsetInput`, red/green verified on master).

**2. Cover all expression batches in the boost rescore non-native fallback
(+ scorer bounds checks).**
2.6 has no `BoostScoreRunner` / chunked `boost_score` C-API — the fallback
lives inline in `PhyRescoresNode::GetOutput`, and it called `ExprSet::Eval`
exactly once. Each `Eval` only advances one internal batch (8192 rows), while
the topk offsets can reference any row of the segment: offsets past the first
batch silently lost their boost (`WeightScorer`) or read out of bounds
(`RandomScorer`). The loop is extracted into
`EvalNonNativeBoostFilterAllBatches` (unit-testable), accumulating batches
until the bitset covers all active rows and folding NULL rows to no-boost.
On 2.6 **both** `WeightScorer` and `RandomScorer` `TargetBitmap` overloads
lacked the bounds check (master's `WeightScorer` already had it) — both are
guarded now.

**3. Advance GIS filter cursors on growing segments during conjunct skip.**
`MoveCursor()` was a no-op on growing segments, so a conjunct short-circuit
(`SkipFollowingExprs`) desynchronized the GIS cursor from its siblings and
later batches ANDed the wrong rows. 2.6 adaptation: gate on
`SegmentExpr::CanUseIndex()` (2.6 has no `UseIndexCursor` helper); growing
interim R-Tree segments advance the global index position plus the data
cursor, mirroring `EvalForIndexSegment`.

The per-offset-chunk re-scan fix from master (#51486 commit 4) does not apply:
2.6 has no chunked `boost_score` C-API, the rescore fallback is single-shot.

## Test

- `ExprTest.GISFilterDoesNotSupportOffsetInput` — contract pin.
- `RescoresNode.NonNativeBoostFilterBitsetCoversAllBatches` — GIS filter over
  N=10000 rows (> one expression batch), asserts filter bits past row 8192.
- `WeightScorerTest` / `RandomScorerTest`
  `BatchScoreTargetBitmapOutOfBoundsOffsets` — out-of-range offsets are
  skipped, no OOB access.
- `ExprTest.GISFilterGrowingConjunctSkipKeepsCursorInSync` (+`Index` variant) —
  20000-row growing segment, `age >= 8192 AND st_within(...)` whose first
  batch matches nothing; asserts no cursor desync on the raw-data and interim
  R-Tree index paths.

### DIFF
--- a/internal/core/src/exec/expression/Expr.cpp
+++ b/internal/core/src/exec/expression/Expr.cpp
@@ -494,5 +494,48 @@ OptimizeCompiledExprs(ExecContext* context, const std::vector<ExprPtr>& exprs) {
     milvus::monitor::internal_core_optimize_expr_latency.Observe(cost / 1000);
 }
 
+TargetBitmap
+EvalExprSetOverAllBatches(ExprSet& expr_set,
+                          EvalCtx& eval_ctx,
+                          int64_t total_rows,
+                          const char* what) {
+    std::vector<VectorPtr> results;
+    TargetBitmap bitset;
+    TargetBitmap valid_bitset;
+    while (static_cast<int64_t>(bitset.size()) < total_rows) {
+        expr_set.Eval(0, 1, true, eval_ctx, results);
+
+        AssertInfo(results.size() == 1 && results[0] != nullptr,
+                   "{}: filter expr returned null result",
+                   what);
+        auto col_vec = std::dynamic_pointer_cast<ColumnVector>(results[0]);
+        if (col_vec == nullptr) {
+            ThrowInfo(ExprInvalid, "{}: result should be ColumnVector", what);
+        }
+        if (!col_vec->IsBitmap()) {
+            ThrowInfo(ExprInvalid, "{}: result should be bitmap", what);
+        }
+        auto col_vec_size = col_vec->size();
+        AssertInfo(col_vec_size > 0,
+                   "{}: filter expr returned empty batch after {} of {} rows",
+                   what,
+                   bitset.size(),
+                   total_rows);
+        TargetBitmapView view(col_vec->GetRawData(), col_vec_size);
+        bitset.append(view);
+        TargetBitmapView valid_view(col_vec->GetValidRawData(), col_vec_size);
+        valid_bitset.append(valid_view);
+    }
+    AssertInfo(static_cast<int64_t>(bitset.size()) == total_rows,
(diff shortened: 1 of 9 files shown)


xiaocai2333:
Completeness audit against the final master PR (#51486), applicable-on-2.6 items only:

**Present** — offset-input gate; full-batch non-native fallback (`EvalNonNativeBoostFilterAllBatches`); RandomScorer bounds check; growing-segment `MoveCursor` fix; native-branch NULL fold; final out-of-bounds warning shape (10s rate limit + suppressed-count folding + segment id); `HasFieldData` comment; the consolidated `exec::EvalExprSetOverAllBatches` helper with both call sites; and (just pushed, af6f4913ec) the `ScopedSegcoreConfigRestore` guard for the GIS growing tests' interim-index toggle, which landed on master after this branch was cut.

**Not applicable on 2.6** (already stated in the squash commit): element-level searches, the chunked C-API paths (per-chunk hoist / ExprSet reuse / cancellation preflight / all-empty fast path), the `null_rejecting` ExprSet flag, and the `UseIndexCursor()` exec-path test assertions.

**One declared omission**: the `NativeFilterGivesNullRowsNoBoost` contract test is not ported. The fix it pins (the native-branch valid-bitmap fold) *is* ported, but on 2.6 the fold lives inside `PhyRescoresNode::Execute` — there is no standalone `ComputeScorerScores` entry point to drive from a unit test, and replicating the full operator harness for one assertion is not worth the weight on a release branch. The fold is exercised indirectly by the existing `RescoresNodeTest` coverage.

czs007 (approved):
/lgtm

PR #51488, opened 2026-07-16
fix: [3.0] gate GIS filters out of the offset-input path

xiaocai2333:
issue: #51485, #51501
pr: #51486

Backport of #51486 to 3.0. The 3.0 `rescores` module matches master's
architecture, so all five commits cherry-pick cleanly with identical content:

1. **Gate GIS filters out of the offset-input path** — `SupportOffsetInput()`
   override so the iterative-filter / rescore consumer takes its non-native
   fallback instead of receiving a misaligned bitmap (silent wrong results).
2. **Cover all expression batches in the boost non-native fallback** — the
   fallback called `ExprSet::Eval` once (one 8192-row batch) while topk
   offsets can reference any row; offsets past the first batch lost their
   boost (`WeightScorer`) or read out of bounds (`RandomScorer`).
3. **Bounds check in `RandomScorer::batch_score`** — mirror `WeightScorer`'s
   guard; out-of-range offsets are treated as no-match.
4. **Evaluate the non-native boost filter once per segment, not per offset
   chunk** — hoist the whole-segment bitset into
   `ComputeNonNativeFilterBitset`, computed once per (scorer, segment) before
   the chunk loop; `O(active_count)` regardless of `num_chunks`.
5. **Advance GIS filter cursors on growing segments during conjunct skip** —
   `MoveCursor()` was a no-op on growing segments, desynchronizing the GIS
   cursor after `SkipFollowingExprs`; growing segments now route through
   `MoveCursorForData()`, and the growing interim-index path also advances
   the global index position.

## Test

Same suite as master #51486:
`ExprTest.GISFilterDoesNotSupportOffsetInput`,
`BoostScoreRunnerTest.ComputeScorerScoresNonNativeFilterCoversAllBatches` /
`…GISFilterCoversAllBatches`,
`RandomScorerTest.BatchScoreTargetBitmapOutOfBoundsOffsets`,
`BoostScoreCTest.NonNativeFilterScoresAllChunks` + three
`ComputeNonNativeFilterBitset` contract tests,
`ExprTest.GISFilterGrowingConjunctSkipKeepsCursorInSync` (+`Index` variant).

### DIFF
--- a/internal/core/src/exec/expression/Expr.cpp
+++ b/internal/core/src/exec/expression/Expr.cpp
@@ -675,5 +675,49 @@ OptimizeCompiledExprs(ExecContext* context, const std::vector<ExprPtr>& exprs) {
     milvus::monitor::internal_core_optimize_expr_latency.Observe(cost / 1000);
 }
 
+TargetBitmap
+EvalExprSetOverAllBatches(ExprSet& expr_set,
+                          EvalCtx& eval_ctx,
+                          int64_t total_rows,
+                          const char* what) {
+    std::vector<VectorPtr> results;
+    TargetBitmap bitset;
+    TargetBitmap valid_bitset;
+    while (static_cast<int64_t>(bitset.size()) < total_rows) {
+        expr_set.Eval(0, 1, true, eval_ctx, results);
+
+        AssertInfo(results.size() == 1 && results[0] != nullptr,
+                   "{}: filter expr returned null result",
+                   what);
+        auto col_vec = std::dynamic_pointer_cast<ColumnVector>(results[0]);
+        if (col_vec == nullptr) {
+            ThrowInfo(
+                UnexpectedError, "{}: result should be ColumnVector", what);
+        }
+        if (!col_vec->IsBitmap()) {
+            ThrowInfo(UnexpectedError, "{}: result should be bitmap", what);
+        }
+        auto col_vec_size = col_vec->size();
+        AssertInfo(col_vec_size > 0,
+                   "{}: filter expr returned empty batch after {} of {} rows",
+                   what,
+                   bitset.size(),
+                   total_rows);
+        TargetBitmapView view(col_vec->GetRawData(), col_vec_size);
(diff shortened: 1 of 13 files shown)


xiaocai2333:
The UT-CPP-Cov failure in run 19425 ("6445 total, process exit non-zero") is the same deterministic shard-timeout as on the master PR — not a defect in this PR's changes. The #51540 test that leaks the process-global `chunk_rows = 2` (SegcoreConfig members are `inline static`) was backported onto 3.0, and this PR's rebase onto that base plus its added tests re-rolled the gtest shard layout into a losing hand (poison before the heavy `TestTermJsonNullable` instances with no restoring test in between). The earlier fully-green run 19280 was on a pre-poison 3.0 base. Full analysis: #51799.

Fix ships as #51801 (clean cherry-pick of #51800 to 3.0). **Stacked on #51801: this branch carries its commit** (74ed14b943) until it merges, so CI here is unblocked; it will be dropped on the post-merge rebase.

xiaocai2333:
Rebased onto latest 3.0 (`02f7295031`), which now contains #51798 — that PR raised `CurrentScalarIndexEngineVersion` from 3 to 4 on 3.0, so `ResolveScalarIndexVersion()` satisfies `MinScalarIndexVersionForJsonPathMultiType` and `create_index` accepts STL_SORT/BITMAP JSON path indexes again.

The previous e2e-amd failure here was that branch-wide issue, not a defect in this PR: run 5053 (07-25 04:22Z) predates #51798 (merged 07-26 16:16Z), and every 3.0 e2e failure with the `requires scalar index engine version >= 4, current resolved version: 3` signature falls before that timestamp — including the 3.0 tip commit's own run. Runs whose merge-set includes the fix pass with those tests exercised (e.g. #51868: 103/103, covering `test_milvus_client_index.py` and `test_milvus_client_json_path_index.py`).

New head `c9269b6c87`; CI re-running. UT-CPP-Cov was already green on the previous head (8241/8241).

Issue #51493, opened 2026-07-17
[Bug]: Proxy metrics status label enum values were rewritten in place (fail -> fail_input/fail_system), silently breaking existing dashboards and alerts


czs007:
**Is there an existing issue for this?**

- [x] I have searched the existing issues

**Environment**

```
- Milvus version: v2.6.19, v2.6.20, master
- Deployment mode(standalone or cluster): both
- MQ type(rocksmq, pulsar or kafka): n/a
- SDK version(e.g. pymilvus v2.0.0rc2): n/a
- OS(Ubuntu or CentOS): n/a
- CPU/Memory: n/a
- GPU: n/a
- Others: Prometheus / Grafana
```

**Current Behavior**

#50221 (master) / #50545 (cherry-pick into 2.6, `39c9b039d2`, first released in **v2.6.19**) split the coarse `status` label values of `milvus_proxy_req_count` and `milvus_proxy_grpc_latency` into fine-grained ones:

| before (<= v2.6.18) | after (v2.6.19+) |
|---|---|
| `status="fail"` | `status="fail_input"` / `status="fail_system"` |
| `status="rejected"` | `status="rejected_user"` / `status="rejected_system"` |
| `status="fail"` / `status="rejected"` (client-cancelled requests) | `status="cancel"` (new value) |

The classification itself is useful — it tells "the caller sent a bad request" apart from "Milvus has an internal problem", so operators can alert only on the latter. The problem is **how** it was delivered: the enum values of an existing label were rewritten in place.

This is a silent breaking change for every existing Prometheus consumer:

- Any dashboard panel or alert rule matching `status="fail"` or `status="rejected"` returns **no data** after upgrading from <= v2.6.18 to v2.6.19+. It does not error — it silently goes to zero, which for an alert rule means **the alert stops firing**.
- The repo's own dashboard was migrated in the same PR (`deployments/monitor/grafana/milvus-dashboard.json` now uses `status=~"fail_.*"`), but users' own dashboards were not, and there is no way for us to migrate them.
- Users cannot opt out or stage the migration: the change is unconditional, with no config switch.

**Expected Behavior**

Adding a classification should not redefine the meaning of values that already exist. The standard Prometheus way to express "this failure, broken down by who caused it" is a **separate label**, not new values on the existing one:

```
status: success | fail | rejected | retry | total | abandon     (unchanged, exactly as <= v2.6.18)
cause:  user | system | cancel | na                             (new dimension)
```

[...]

PR #50221, opened 2026-06-01
enhance: standardize error handling on merr + Sys/Input classification

czs007:
issue: #47420

## What this PR does

Project-wide migration of raw `fmt.Errorf` / `errors.New` in function bodies onto
the `merr` framework, plus the Sys-vs-Input error classification and the
machinery it drives (retriability, fine-grained metrics, segcore unification),
plus the convention docs and a linter that keeps it from regressing.

Scope: storage, proxy, coordinators (root/data/query), query node, data node,
`pkg/util` & `internal/util`, expression parser, message queue, streaming, and
misc packages. Bare raw-error usages went from ~3000 to a ~340 allowlist
(package-level sentinels / build-tag / test sites).

---

## How to review this PR

It is large but the vast majority is mechanical. Changes fall into three tiers;
spend review budget on Part 2 and Part 3.

### Part 1 — Mechanical standardization (low risk, verify by rule)

Each converted call follows one of a small fixed set of rules. To review, check
that each site obeys the matching rule rather than reading every line:

| Pattern | Rule |
|---|---|
| `fmt.Errorf("...")` originating a new error | → `merr.WrapErrXxxMsg("...")` with a code matching the failure's meaning |
| Adding context to an existing typed error | → `merr.Wrap(err, "...")` / `merr.Wrapf(...)` — **preserves** the inner code (never `WrapErr*Err`, which overwrites it) |
| Errors inside the streaming subsystem | → `status.New*` factories (StreamingError), **not** merr — this is the component-internal dialect (see `docs/dev/error_handling_guide.md`) |
| Low-level / control-flow signal caught by `errors.Is` | → kept as a package-level `errors.New` sentinel (lowercase, same-package) |

Conventions are documented in `docs/dev/error_handling_guide.md` (how-to) and
`docs/dev/error_sentinel_convention.md` (rules + audit). A `gocritic`/`ruleguard`
rule (`rawmerrerror`, in `rules.go`) enforces "no raw `return errors.New/fmt.Errorf`"
under `make verifiers`.

### Part 2 — Behavior changes (review these closely)

These are the sites where the wire contract or runtime behavior changes, not just
the source text. Listed by category; representative locations given, full set in
the diff.

**A. gRPC wire-code shifts: `UnexpectedError(1)/Code 65535` → typed code.**
Where a handler previously returned a raw error (collapsed to `Code=65535` on the
wire), it now returns a typed merr, so the client sees a real code. The most
[...]

PR #51495, opened 2026-07-17
fix: keep proxy metric status label stable, split cause into its own label

czs007:
issue: #51493

### What

#50221 split the `status` label values of `milvus_proxy_req_count` / `milvus_proxy_grpc_latency` in place. Released in **v2.6.19**, that silently zeroes every existing dashboard panel and alert rule matching `status="fail"` / `status="rejected"` — an alert that stops firing rather than erroring.

```
cause:  user | system | cancel | na                            (new)
```

|---|---|
| `success` / `retry` / `total` / `abandon` | unchanged, `cause="na"` |

### Why a label instead of new status values

- **Old queries work unchanged and count each request once.** `status="fail"` is again every hard failure; Prometheus aggregates over `cause` for free. That rollup is what a label dimension *is* — the split forces every consumer who just wants "how many failures" to hand-write `status=~"fail_.*"`.
- **Cardinality is unchanged.** `cause` is functionally dependent on the outcome, so the realized `(status, cause)` pairs are exactly the series the split already produces. Additive, not multiplicative.
- **New consumers lose nothing**: alert on `status="fail", cause="system"`; route `cause="user"` to the owning application team.

The alternative — emitting old and new `status` values side by side during a transition — was rejected: it defers the break rather than removing it (the old values must still be dropped eventually, forcing the same migration later), and meanwhile double-counts every request in unfiltered aggregations.

`cancel` is folded into `cause` for the same reason: before v2.6.19 client cancellations were counted as `fail` (cancel in the response status) or `rejected` (cancel at the interceptor), so promoting it to a `status` value quietly makes `status="fail"` under-count versus older releases. As a cause it stays excludable via `cause!="cancel"` without redefining `status`.

### Also in this PR

- The 7 `snapshot_impl` sites that emitted a bare `fail` with no classification now report their real cause via `failMetricLabel(err)` (e.g. `snapshot_metadata_uri is required` is `cause="user"`, not a system fault). Their `status` is unchanged.
- `deployments/monitor/grafana/milvus-dashboard.json`: the "Faild Request Rate" panel goes back to `status="fail"` — a verbatim revert of what #50221 changed, which is the compatibility claim demonstrated.
- Dev docs and stale comments that named the old label values.

### Verification

- `TestParseMetricLabelStatusDomainIsStable` pins the status domain, so re-splitting it fails CI rather than shipping.
- Adding a label makes every `WithLabelValues` site arity-sensitive **at runtime, not compile time** — a missed site panics in production. Unit tests do not reach all of them (coverage shows `GetRestoreSnapshotState`, `ListRestoreSnapshotJobs`, `PinSnapshotData`, `UnpinSnapshotData` at 0%), so all **58 emit sites were verified statically via AST**, not only the ones tests happen to hit.
- Passing: `pkg/metrics`, `pkg/util/requestutil`, `pkg/util/merr`, `internal/distributed/proxy` (incl. `httpserver`, which covers the REST emit path), and the `internal/proxy` snapshot / metric-label tests. `golangci-lint` clean on every touched package.
- Pre-existing and unrelated: `service_test.go` bind failures (`listen tcp :19529: address already in use`) reproduce identically on unmodified master — a local process holds the port.

### Rollout

Should be cherry-picked to 2.6 so the window in which the fine-grained `status` values exist stays confined to v2.6.19–v2.6.20. Users who already adopted the new values on those two releases need a one-time query change (`status="fail_system"` -> `status="fail", cause="system"`); that is a known, bounded set, and preferable to leaving every pre-2.6.19 dashboard silently broken.

### DIFF
--- a/deployments/monitor/grafana/milvus-dashboard.json
+++ b/deployments/monitor/grafana/milvus-dashboard.json
@@ -4794,7 +4794,7 @@
                 "uid": "${datasource}"
               },
               "exemplar": true,
-              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=~\"fail_.*\"}[2m])/120) by(function_name, pod, node_id)",
+              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=\"fail\"}[2m])/120) by(function_name, pod, node_id)",
               "interval": "",
               "legendFormat": "{{function_name}}-{{pod}}-{{node_id}}",
               "queryType": "randomWalk",
--- a/pkg/util/requestutil/getter.go
+++ b/pkg/util/requestutil/getter.go
@@ -23,7 +23,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
+	grpcstatus "google.golang.org/grpc/status"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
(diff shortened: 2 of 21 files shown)


tinswzy:

deployments/monitor/grafana/milvus-dashboard.json
```diff
@@ -4794,7 +4794,7 @@
                 "uid": "${datasource}"
               },
               "exemplar": true,
-              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=~\"fail_.*\"}[2m])/120) by(function_name, pod, node_id)",
+              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=\"fail\"}[2m])/120) by(function_name, pod, node_id)",
```
liliu-z, line 4797: The verbatim revert to `status="fail"` silently under-counts failures during a rolling upgrade *from* v2.6.19/v2.6.20 (whose un-upgraded nodes still emit `fail_input`/`fail_system`) and for historical data recorded on those two releases; pre-2.6.19 deployments are unaffected since they already emitted `fail`. If transition/historical robustness matters, `status=~"fail.*"` matches `fail` identically for the common case while also covering the `fail_*` variants; otherwise this is an accepted, bounded trade-off for the clean revert.

pkg/util/requestutil/getter.go
```diff
@@ -251,48 +250,48 @@ func ParseMetricLabel(resp any, err error) string {
...
 		if errors.Is(err, context.Canceled) {
-			return metrics.CancelLabel
+			return metrics.RejectedLabel, metrics.CauseCancel
 		}
 		if merr.GetErrorType(err) == merr.InputError {
-			return metrics.RejectedUserLabel
+			return metrics.RejectedLabel, metrics.CauseUser
 		}
-		switch status.Code(err) {
+		switch grpcstatus.Code(err) {
```
liliu-z, line 289: Here a raw gRPC `codes.Canceled` error (reachable from `internal/proxy/impl.go:7305`) falls through to `cause="system"`. A client-cancelled request is not a Milvus/system fault, so bucketing it as `system` mis-attributes blame in the new cause dimension; this is pre-existing behavior and status-label compatible, hence low impact. Add an explicit `codes.Canceled` case; leave `DeadlineExceeded` as-is unless timeout-blame attribution is deliberately decided.

liliu-z:
Took another pass — I don’t see anything blocking this PR. The main fix looks good, and the remaining points are rollout edge cases or pre-existing metric debt. Fine to merge and follow up separately. LGTM.

liliu-z (commented):
/lgtm
/approve

PR #51496, opened 2026-07-17
fix: [hotfix-2.6.20] keep proxy metric status label stable, split cause into its own label (#51495)

czs007:
Cherry-pick from master
pr: https://github.com/milvus-io/milvus/pull/51495
issue: #51493

### What

```
cause:  user | system | cancel | na                            (new)
```

|---|---|
| `success` / `retry` / `total` / `abandon` | unchanged, `cause="na"` |

Old queries work unchanged and still count each request once: `status="fail"` is again every hard failure, because Prometheus aggregates over `cause` for free — that rollup is what a label dimension is for. `cause` is functionally dependent on the outcome, so the realized `(status, cause)` pairs are the same set of series the split already produces (10 vs 9 per node/function/db/collection), not the product of both domains.

Landing this on 2.6 confines the window in which the fine-grained `status` values ever existed to v2.6.19–v2.6.20.

### Cherry-pick notes

Not a clean pick — the differences from the master PR are:

- `internal/proxy/snapshot_impl.go` and `internal/util/queryutil/helpers.go` do not exist on this branch; those hunks are dropped (47 emit sites + 1 comment on master, 0 here).
- `pkg/util/requestutil/getter_test.go` import block resolved to this branch's `pkg/v2` / `go-api/v2` module paths.

Because the emit surface differs from master, **the 12 `WithLabelValues` sites on this branch were re-verified for label arity by AST on this branch**, not assumed from the master run — adding a label makes arity a runtime panic, not a compile error.

- Builds: `internal/proxy/...`, `internal/distributed/proxy/...`, `pkg/metrics`, `pkg/util/requestutil`.
- Passing: `pkg/metrics`, `pkg/util/requestutil` (incl. `TestParseMetricLabelStatusDomainIsStable`, which pins the status domain so a future re-split fails CI), `pkg/util/merr`, and `TestUnaryRequestStatsInterceptor`.
- AST arity check: 12/12 emit sites correct.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -34,21 +34,23 @@ const (
 	RetryLabel    = "retry"
 	RejectedLabel = "rejected"
 
-	// Fine-grained hard-failure labels (composite status values, not a new
-	// dimension label, to keep Prometheus cardinality additive rather than
-	// multiplicative). They split the coarse "fail"/"rejected" into the
-	// responsible party so monitoring/alerting can tell a user-input error
-	// (the caller must fix the request) apart from an internal system error
-	// (operators must intervene). MIGRATION: old queries on status="fail" must
-	// move to status=~"fail_.*" (fail_input/fail_system), and old queries on
-	// status="rejected" must move to status=~"rejected_.*" (rejected_user for
-	// caller-side auth/privilege/bad-arg rejections, rejected_system otherwise).
-	// Dashboards still matching the bare "fail"/"rejected" values return nothing
-	// after this split.
-	FailInputLabel      = "fail_input"
-	FailSystemLabel     = "fail_system"
-	RejectedSystemLabel = "rejected_system"
-	RejectedUserLabel   = "rejected_user"
+	// Values of the "cause" label, a dimension orthogonal to "status" that names
+	// the responsible party for a failed request, so monitoring can tell a
+	// user-input error (the caller must fix the request) apart from an internal
+	// system error (operators must intervene). It is a separate label rather
+	// than extra "status" values on purpose: the coarse status stays a valid
+	// query on its own ("fail" is still every hard failure, aggregated over
+	// cause by Prometheus), so dashboards written against it keep working.
+	// Cause is functionally dependent on the outcome, so the realized
+	// (status, cause) pairs are additive, not the product of both domains.
+	CauseUser   = "user"   // the request itself is at fault: bad arguments, missing auth, no such collection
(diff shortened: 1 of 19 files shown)


pkg/metrics/proxy_metrics.go
```diff
@@ -233,7 +233,7 @@ var (
 			Subsystem: typeutil.ProxyRole,
 			Name:      "req_count",
 			Help:      "count of operation executed",
-		}, []string{nodeIDLabelName, functionLabelName, statusLabelName, databaseLabelName, collectionName})
+		}, []string{nodeIDLabelName, functionLabelName, statusLabelName, causeLabelName, databaseLabelName, collectionName})
```
liliu-z, line 236: Promoting `cause` to its own label on `milvus_proxy_req_count` / `milvus_proxy_grpc_latency` subdivides each existing (node, function, db, collection, status) series one level finer by responsible party. Aggregation queries such as `sum(...) by (...)` are unaffected, and no in-repo asset breaks: all four bundled dashboard panels aggregate and the repo ships no alert rules. But a downstream raw per-series threshold like `rate(milvus_proxy_req_count{status="fail"}[5m]) > 10` now evaluates per cause — a tuple at 12/s that splits into user=6/s + system=6/s no longer crosses the trigger, so such an alert can silently stop firing. Narrow the PR's 'old queries work unchanged' claim in the release notes to aggregation queries; raw per-series thresholds and `on/ignoring(status)` vector matches must account for `cause`.

pkg/metrics/metrics.go
```diff
@@ -34,21 +34,19 @@ const (
...
+	// system error (operators must intervene). It is a separate label rather
+	// than extra "status" values on purpose: the coarse status stays a valid
+	// query on its own ("fail" is still every hard failure, aggregated over
+	// cause by Prometheus), so dashboards written against it keep working.
+	// Cause is functionally dependent on the outcome, so the realized
+	// (status, cause) pairs are additive, not the product of both domains.
+	CauseUser   = "user"   // the request itself is at fault: bad arguments, missing auth, no such collection
+	CauseSystem = "system" // Milvus is at fault: component failure, IO error, internal bug
+	CauseCancel = "cancel" // neither party: the client gave up before the request completed
+	CauseNA     = "na"     // no cause applies: the request did not hard-fail
```
LoveEachDay: set an empty string would be more clear, other than "na"

liliu-z (commented):
/lgtm
/approve

Issue #51501, opened 2026-07-17
[Bug]: Boost scoring with a non-offset-input filter only evaluates the first 8192 rows; RandomScorer can read out of bounds


xiaocai2333:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

- Milvus version: master
- Deployment mode: any

### Current Behavior

When a boost scorer's filter contains an expression that does not support offset input (text match / phrase match today; GIS filters once #51486 lands), `ComputeScorerScores` (`internal/core/src/rescores/BoostScoreRunner.cpp`) takes the non-native fallback branch. That branch calls `ExprSet::Eval` exactly once, but each Eval only advances the expression by one internal batch (`DEFAULT_EXEC_EVAL_EXPR_BATCH_SIZE` = 8192 rows). On any segment larger than 8192 rows the filter bitset therefore only covers the first batch, while the topk offsets being scored can reference any row of the segment:

1. `WeightScorer::batch_score(TargetBitmap)` bounds-checks the offset and **silently drops the boost** for every offset beyond the first batch (wrong scores, no error).
2. `RandomScorer::batch_score(TargetBitmap)` (`internal/core/src/rescores/Scorer.cpp`) has **no bounds check** and indexes `bitmap[offsets[i]]` past the end of the bitset — an out-of-bounds read.

Both defects were surfaced by the adversarial review of #50675 (see the review summary comment there); they pre-exist on master and are reachable today via a text-match filter in a boost scorer.

### Expected Behavior

The non-native fallback accumulates Eval batches until the filter bitset covers all active rows of the segment (mirroring `PhyIterativeFilterNode`), and `RandomScorer`'s `TargetBitmap` overload bounds-checks offsets like `WeightScorer` does.

### Steps To Reproduce

Boost (function score) search with a `text_match` filter in the scorer on a segment with more than 8192 rows; request topk results whose segment offsets exceed 8192. Matched rows beyond offset 8192 receive no boost; with a random scorer the read is out of bounds.

### Anything else?

Fix incoming.

xiaocai2333:
has been fixed.
/close

PR #51502, opened 2026-07-17
fix: cover all expression batches in boost scoring non-native filter path

xiaocai2333:
issue: #51501

### What this fixes

When a boost scorer's filter contains an expression that does not support offset input (text match / phrase match today; GIS once #51486 lands), `ComputeScorerScores` takes the non-native fallback branch in `BoostScoreRunner.cpp`. That branch called `ExprSet::Eval` exactly once, but each Eval only advances the expression by one internal batch (`DEFAULT_EXEC_EVAL_EXPR_BATCH_SIZE` = 8192 rows), so on segments larger than one batch the filter bitset only covered the first 8192 rows while the topk offsets can reference any row:

- `WeightScorer::batch_score(TargetBitmap)` silently dropped the boost for every offset beyond the first batch.
- `RandomScorer::batch_score(TargetBitmap)` had no bounds check and read past the end of the bitset (OOB read).

### Changes

1. **`BoostScoreRunner.cpp` — non-native fallback covers every expression batch.** The branch now accumulates Eval batches until the bitset covers all active rows (mirroring `PhyIterativeFilterNode`'s accumulation loop) and folds the valid (null) bitmap into the result.
2. **`Scorer.cpp` — missing bounds check in `RandomScorer::batch_score(TargetBitmap)`.** Now bounds-checks offsets exactly like `WeightScorer`, as defense-in-depth for bitmaps that legitimately lag the segment. The bounds check alone is not the fix — (1) is; it only converts an OOB read into a skipped boost.
3. **`BoostScoreRunner.cpp` — symmetric null handling on both paths** *(follow-up from the adversarial review of this PR)*. The native (offset-input) path now folds the valid bitmap exactly like the non-native path, so a null row never receives a boost whichever path evaluates the filter, instead of relying on the convention that UNKNOWN rows carry data=0 (`PhyIterativeFilterNode` folds on both paths for the same reason). With both consumers null-rejecting, the filter `ExprSet` is constructed with `null_rejecting=true`, letting conjunctions drop UNKNOWN rows from their active sets early.

### Performance note

The non-native fallback previously evaluated a single 8192-row batch per boost-scored query — which is precisely the bug: the bitset covered only those rows. It now evaluates **all active rows of the segment**, so its cost scales with segment size instead of staying constant. This applies to every filter expression without offset-input support: text/phrase-match today, GIS once #51486 lands. Full-segment evaluation is the only correct option while these expressions cannot consume an offset list — a deliberate correctness-over-performance trade (the pre-fix behavior was silently wrong or an OOB read, so no correct-and-fast baseline is given up). The native path stays O(topk).

**Follow-up (tracked, not in this PR):** implement offset-input support for the expressions stuck on the fallback (text-index ops in `PhyUnaryRangeFilterExpr`, GIS in `PhyGISFunctionFilterExpr`) the way sibling exprs do, then the fallback becomes cold again.

### Tests

- `BoostScoreRunnerTest.ComputeScorerScoresNonNativeFilterCoversAllBatches` — end-to-end through the non-native branch on a 10000-row segment (> one expression batch) with a text-match filter; offsets beyond row 8192 must still receive their boost.
- `RandomScorerTest.BatchScoreTargetBitmapOutOfBoundsOffsets` — the previously uncovered overload without the bounds check; out-of-range offsets are skipped with no OOB access.
- `BoostScoreRunnerTest.ComputeScorerScoresNativeFilterSkipsNullRows` — pins the null semantics on the native path: a nullable-field filter over a sealed segment with NULL rows never boosts a null row, independent of the UNKNOWN data-bit convention.

### Verification

All scorer tests (8/8, including the three above) pass locally under ASAN against the fixed code. Red-side verification of the batch-accumulation defect was established during review (single-batch bitset reproduces exactly at offsets past 8192).

### Scope / provenance

Both original defects pre-exist on master and are reachable today via a text-match filter in a boost scorer. They were surfaced by the adversarial review of #50675 (which widens exposure to GIS filters via `SupportOffsetInput() → false`, split out as #51486). Per the packaging feedback on #50675, this unconditional behavior fix lands as its own `fix:` PR instead of riding the optimization. Note: #51486 currently carries an equivalent copy of these fixes; whichever merges first, the other rebases to drop the overlap.

### DIFF
--- a/internal/core/src/rescores/BoostScoreRunner.cpp
+++ b/internal/core/src/rescores/BoostScoreRunner.cpp
@@ -97,21 +97,48 @@ ComputeScorerScores(exec::ExecContext* exec_context,
                             bitsetview,
                             output_scores);
     } else {
-        expr_set->Eval(0, 1, true, eval_ctx, results);
-
-        AssertInfo(!results.empty() && results[0] != nullptr,
-                   "ComputeScorerScores: filter expr returned null result, "
-                   "filter: {}",
-                   filter->ToString());
-        auto col_vec = std::dynamic_pointer_cast<ColumnVector>(results[0]);
-        AssertInfo(col_vec != nullptr,
-                   "ComputeScorerScores: failed to cast result to "
-                   "ColumnVector, filter: {}",
-                   filter->ToString());
+        // Without offset-input support each Eval only advances the expression
+        // by one internal batch (DEFAULT_EXEC_EVAL_EXPR_BATCH_SIZE rows), while
+        // `offsets` may reference any row of the segment. Accumulate every
+        // batch so the bitset covers all active rows.
+        auto active_count =
+            exec_context->get_query_context()->get_active_count();
         TargetBitmap bitset;
-        auto col_vec_size = col_vec->size();
-        TargetBitmapView view(col_vec->GetRawData(), col_vec_size);
-        bitset.append(view);
+        TargetBitmap valid_bitset;
+        while (static_cast<int64_t>(bitset.size()) < active_count) {
+            expr_set->Eval(0, 1, true, eval_ctx, results);
+
+            AssertInfo(!results.empty() && results[0] != nullptr,
+                       "ComputeScorerScores: filter expr returned null result, "
+                       "filter: {}",
+                       filter->ToString());
+            auto col_vec = std::dynamic_pointer_cast<ColumnVector>(results[0]);
+            AssertInfo(col_vec != nullptr,
+                       "ComputeScorerScores: failed to cast result to "
+                       "ColumnVector, filter: {}",
+                       filter->ToString());
+            auto col_vec_size = col_vec->size();
+            AssertInfo(col_vec_size > 0,
+                       "ComputeScorerScores: filter expr returned empty batch "
+                       "after {} of {} rows, filter: {}",
+                       bitset.size(),
+                       active_count,
+                       filter->ToString());
+            TargetBitmapView view(col_vec->GetRawData(), col_vec_size);
+            bitset.append(view);
+            TargetBitmapView valid_view(col_vec->GetValidRawData(),
+                                        col_vec_size);
+            valid_bitset.append(valid_view);
+        }
+        AssertInfo(static_cast<int64_t>(bitset.size()) == active_count,
+                   "ComputeScorerScores: filter bitset size {} must match "
+                   "active count {}, filter: {}",
+                   bitset.size(),
+                   active_count,
+                   filter->ToString());
+        // Fold UNKNOWN (NULL) into FALSE explicitly so null rows never
+        // receive a boost, matching PhyIterativeFilterNode's handling.
+        bitset.inplace_and(valid_bitset, bitset.size());
         scorer->batch_score(
             op_context, segment, function_mode, offsets, bitset, output_scores);
     }
--- a/internal/core/src/rescores/Scorer.cpp
+++ b/internal/core/src/rescores/Scorer.cpp
@@ -128,11 +128,18 @@ RandomScorer::batch_score(milvus::OpContext* op_ctx,
     target_offsets.reserve(offsets.size());
     idx.reserve(offsets.size());
 
+    auto bitmap_size = bitmap.size();
     for (auto i = 0; i < offsets.size(); i++) {
-        if (bitmap[offsets[i]] > 0) {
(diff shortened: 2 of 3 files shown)


czs007:
This PR fixes null handling in the non-native path of `BoostScoreRunner`, and the review surfaced three follow-ups: an asymmetry with the native path (which still doesn't fold `valid`), a missing `null_rejecting=true` flag, and a performance trade-off worth documenting.

---

Adversarial review of this change surfaced three follow-ups — all non-blocking — centered on the asymmetry between the native and non-native paths in `BoostScoreRunner::Rescore`.

## Low

**`internal/core/src/rescores/BoostScoreRunner.cpp:92` — native path does not fold `valid`, so null rows can still receive a boost**

The non-native path now folds UNKNOWN into FALSE via `bitset.inplace_and(valid_bitset, bitset.size())` (line 141), with the stated rationale that null rows should not receive a boost. The native path (lines 92-98) still wraps `col_vec->GetRawData()` directly into a `TargetBitmapView` and hands it to `batch_score`, never reading `GetValidRawData()`. For a nullable field (e.g. `nullable_field > 5`) on the native path, an UNKNOWN row whose data bit happens to be 1 will incorrectly receive a boost.

Notably, `PhyIterativeFilterNode` — the comparison this PR cites — folds on *both* paths: `IterativeFilterNode.cpp:197` (non-native) and `:308-310` (native/offset-input), and its comment explicitly states this is done "instead of relying on the convention that UNKNOWN rows carry data=0", i.e. that convention is deliberately treated as unreliable.

*Suggestion:* fold `valid` on the native path as well (mirroring `IterativeFilterNode.cpp:305-310`), or add a comment explaining why the native path is exempt. This is a pre-existing issue rather than a regression from this diff, but the PR currently fixes only one of the two paths and leaves the null semantics asymmetric. *(raised by auditor)*

**`internal/core/src/rescores/BoostScoreRunner.cpp:108` — non-native path now evaluates the whole segment, which is costly when `offsets` is small**

The non-native path loops `Eval` until the bitset covers `active_count` (`while (static_cast<int64_t>(bitset.size()) < active_count)`), so it evaluates the entire segment even when `offsets` holds only topk entries. Before the fix it evaluated a single batch of 8192 rows (`Consts.h:90`, `DEFAULT_EXEC_EVAL_EXPR_BATCH_SIZE = 8192`), so on a large sealed segment the cost grows from 8192 rows to `active_count` rows. The native path stays at O(topk) via offset input (`BoostScoreRunner.cpp:78`).

This is an inherent trade-off of the fix rather than a defect: text/phrase-match expressions currently return `false` from `SupportOffsetInput()` (`UnaryExpr.h:1032-1037`), so no offset input is available and full-segment evaluation is the only correct option. Correctness over performance is the right call here.

*Suggestion:* call out the added cost explicitly in the PR description, and follow up by extending offset-input support to these expressions. *(raised by wzy)*

## Nitpick

**`internal/core/src/rescores/BoostScoreRunner.cpp:66` — `ExprSet` is constructed without `null_rejecting=true`, diverging from the `PhyIterativeFilterNode` this PR cites**

The `ExprSet` here uses the default `null_rejecting=false`. The comment at line 141 says the explicit `valid` fold matches "PhyIterativeFilterNode's handling", but that node does both things: it passes `/*null_rejecting=*/true` (`IterativeFilterNode.cpp:70-71`) *and* folds `valid` on both paths. This PR adopts only the second half.

The flag is a pure optimization — the base `MarkNullRejecting()` is a no-op (`Expr.h:201-202`), and only `ConjunctExpr` (`ConjunctExpr.h:170-173`) acts on it and propagates downward, letting AND/OR drop UNKNOWN rows from the active set early. So a single text-match filter is unaffected; only filters containing AND/OR benefit.

*Suggestion:* either pass `null_rejecting=true` (the semantic precondition holds, since `valid` is explicitly folded here), or add a one-line comment noting the divergence is intentional. *(raised by wzy)*

xiaocai2333:
Superseded by #51486: the adversarial review of #51486 (which gates GIS into this same non-native fallback) asked for these fixes to land together with the GIS routing change, so this commit (21202ef3cf) has been absorbed into #51486 as-is, plus a GIS-variant regression test that requires both changes. Closing to avoid a duplicate; review follow-ups from this PR's adversarial review (native-path valid fold, null_rejecting flag, comment corrections) are being addressed on #51486.

PR #51504, opened 2026-07-17
feat: [Vortex1] add scalar local format data scan foundation

aoiasd:
relate: #50304

design doc: docs/design-docs/design_docs/20260305-local_format.md

## Summary
Introduce the scalar local-format metadata contract and column-oriented DataScan foundation described in the [local format design document](https://github.com/milvus-io/milvus/blob/buqian/aoiasd/aoiasd-vortex-data-scan/docs/design-docs/design_docs/20260305-local_format.md). This includes write-time grouping, field validation, raw-compatible scan interfaces, and focused coverage for the new contract.

Issue #50304, opened 2026-06-05
feat: support vortex local format


aoiasd:
## Motivation

Support Vortex as a local field format for sealed segment data. The goal is to load data locally, either from memory or disk, through Vortex without fully decompressing it into the existing row format first.

This can reduce memory and disk resource usage for scalar fields, and gives the query path an opportunity to benefit from Vortex compression and predicate/filter pushdown.

## Scope

- Add metadata/configuration to mark fields that should use Vortex local format.
- Write/load Vortex local-format data for supported scalar fields.
- Read data through Vortex reader paths without eagerly decompressing the full field into row format.
- Support scan/filter paths needed by query execution.

## Notes

Initial support is focused on scalar fields in sealed segments. Vector fields and external table end-to-end support can be considered separately.

xiaofan-luan:
@aoiasd 
i'm assuming this is still under progress.

PR #51505, opened 2026-07-17
feat: [Vortex2] evaluate compare expressions with column scans

aoiasd:
relate: #50304
pr: https://github.com/milvus-io/milvus/pull/51504
design doc: docs/design-docs/design_docs/20260305-local_format.md

## Summary
Evaluate compare expressions through batched column scans while preserving the existing raw execution behavior. Add focused coverage for scalar, string, nullable, offset-input, and reverse-comparison cases.

### DIFF
--- a/internal/core/src/exec/expression/CompareExpr.h
+++ b/internal/core/src/exec/expression/CompareExpr.h
@@ -17,11 +17,14 @@
 #pragma once
 
 #include <fmt/core.h>
+#include <algorithm>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
 #include <optional>
 #include <string>
+#include <string_view>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
@@ -40,12 +43,47 @@
 #include "exec/expression/Expr.h"
 #include "expr/ITypeExpr.h"
 #include "index/Index.h"
+#include "mmap/ChunkedColumnInterface.h"
 #include "pb/plan.pb.h"
 #include "segcore/SegmentChunkReader.h"
 #include "segcore/SegmentInterface.h"
 
 namespace milvus {
 namespace exec {
+
+template <typename T>
+inline constexpr bool IsCompareStringViewType =
+    std::is_same_v<std::remove_cv_t<T>, std::string_view>;
+
+template <typename T, typename U, proto::plan::OpType op>
+inline bool
+CompareColumnValues(const T& left, const U& right) {
+    if constexpr (op == proto::plan::OpType::Equal) {
+        return left == right;
+    } else if constexpr (op == proto::plan::OpType::NotEqual) {
+        return left != right;
+    } else if constexpr (op == proto::plan::OpType::GreaterThan) {
+        return left > right;
+    } else if constexpr (op == proto::plan::OpType::LessThan) {
+        return left < right;
+    } else if constexpr (op == proto::plan::OpType::GreaterEqual) {
+        return left >= right;
+    } else if constexpr (op == proto::plan::OpType::LessEqual) {
+        return left <= right;
+    } else if constexpr (op == proto::plan::OpType::PrefixMatch) {
+        if constexpr (IsCompareStringViewType<T> &&
+                      IsCompareStringViewType<U>) {
+            return PrefixMatch(left, right);
+        } else {
+            ThrowInfo(OpTypeInvalid,
+                      "PrefixMatch only supports string compare expr");
+        }
+    } else {
+        ThrowInfo(OpTypeInvalid,
+                  fmt::format("unsupported op_type:{} for compare expr", op));
+    }
+    return false;
+}
 
 template <typename T,
           typename U,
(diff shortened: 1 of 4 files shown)


internal/core/src/exec/expression/CompareExpr.h
```diff
@@ -17,11 +17,14 @@
 #pragma once
 
 #include <fmt/core.h>
+#include <algorithm>
```
zhengbuqian: test

internal/core/src/exec/expression/CompareExpr.h
```diff
@@ -322,19 +356,104 @@ class PhyCompareFilterExpr : public Expr {
...
+                                                             right_data + i,
+                                                             nullptr,
+                                                             1,
+                                                             res + i,
+                                                             values...);
+            }
         } else {
-            return ProcessBothDataChunksForSingleChunk<T, U, FUNC, ValTypes...>(
-                func, res, valid_res, values...);
+            const auto left_items = left_take->template Access<T>();
```
buqian-zilliz, line 431: **[P2] Avoid routing raw offset comparisons through generic Take**

`ChunkedColumnInterface::Take()` constructs a fresh non-owned `RawTakeResult` for ordinary field columns, so this branch is the normal offset-input path. Each of the two constructors first validates every offset, and each subsequent item performs virtual `PrepareItem`/value access plus a planner `Locate`; the previous path used cached typed chunk pointers and a single direct-pointer loop for one-chunk columns. At production batch sizes this turns one comparison pass into multiple `O(N)` passes with four virtual calls per row. Preserve the existing chunk path for raw columns and select `Take` only for a backend that can return already-owned dense values, exposing a cheap capability if necessary.

buqian-zilliz: This issue still exists: ordinary raw columns still enter the generic non-owned `Take` path for every offset batch, retaining the extra full validation passes and per-row virtual planner lookups instead of using the direct typed chunk path.

aoiasd: Good catch on the duplicate constructor pass. I removed RawTakeResult’s eager full-offset validation; each consumed offset is still range-checked by ColumnPlanner::Locate.

One correction after tracing the pre-change path: every Column-backed Raw segment here is ChunkedSegmentSealedImpl (`is_chunked() == true`), so the old offset-compare path also resolves both fields through GetChunkByOffset per row. Non-chunked segments do not expose a Column and therefore continue through the unchanged direct Chunk fallback. ColumnPlanner now also preserves the old physical single-Cell `{0, offset}` fast path.

I corrected the benchmark so all offsets hit one Cell while both sides retain their production lookup flow. Across nullable/non-nullable dense, ordered-25%, and shuffled-25% inputs (1024-row windows, Release, fixed CPU, 7 repetitions), Interface/Legacy CPU medians are 0.419x–0.485x. The full C++/gtest build and 114 focused Planner/Raw Take/offset tests pass. Therefore the redundant O(N) validation is gone, but a Raw-only execution fallback is not needed.

internal/core/src/mmap/ChunkedColumnTakeCursors.cpp
```diff
@@ -89,13 +89,6 @@ class RawTakeResult final : public ChunkedColumnInterface::TakeResult {
         AssertInfo(static_cast<bool>(pin_cell_),
                    "raw take cell accessor is null");
         AssertInfo(planner_ != nullptr, "raw take planner is null");
-        for (int64_t i = 0; i < offsets_.size; ++i) {
```
buqian-zilliz, line 92: **[P2] Preserve range validation for non-nullable validity access**

Removing this constructor validation leaves `RawTakeResult::IsValid()` unable to reject an out-of-range segment offset on non-nullable columns. `TakeResult::IsValid()` invokes `PrepareItem(index, false)`, and the non-nullable fast path returns `{true, false}` before `ResolveLocation()` reaches `ColumnPlanner::Locate()`, so invalid positional input is reported as valid instead of failing like the previous implementation and other `Take` backends. Keep the lazy behavior by resolving and validating the requested offset inside `PrepareItem()` before this fast return rather than restoring an eager full-array pass.

internal/core/src/exec/expression/CompareExpr.h
```diff
@@ -528,67 +665,253 @@ class PhyCompareFilterExpr : public Expr {
...
+        const auto window_start = GetCurrentRows();
+        if (left_data_scan_cursor_ == nullptr) {
+            const auto pin_policy =
+                segcore::SegcoreConfig::default_config()
+                        .get_scan_cursor_owns_pin()
+                    ? ChunkedColumnInterface::ScanPinPolicy::CursorOwned
+                    : ChunkedColumnInterface::ScanPinPolicy::ResultOwned;
+            auto left_cursor = left_data_column_->Scan(
+                op_ctx_,
+                ChunkedColumnInterface::ScanOptions::ForData(
```
buqian-zilliz, line 774: **[P2] Prefetch both scan columns before consuming cold cells**

`ScanOptions::ForData()` defaults `prefetch` to false, so both newly created cursors skip `RawScanCursor::PrefetchRemainingCells()`. `PhyCompareFilterExpr` has no raw-data `PrefetchAsync()` override, leaving each subsequent `Next()` to pin one cell at a time, first on the left and then on the right. For two cold multi-cell columns, this serializes up to twice the remaining cell count of tiered-storage loads on the query critical path. The generic `SegmentExpr` scan path enables cursor prefetch on its first data scan. Enable prefetch for both cursor opens, or issue equivalent batched prefetches before the first `Next()`, so cold-cell I/O is submitted in bulk.

aoiasd: Thanks. I checked the pre-change path: PhyCompareFilterExpr did not override PrefetchAsync or call prefetch_chunks, and its sequential raw-data path pinned the left and right chunks synchronously as they were consumed. The new Scan path intentionally preserves that I/O behavior while replacing the access interface. Batched prefetch may improve cold multi-cell comparisons, but it changes the existing scheduling behavior and is outside this PR’s compatibility-focused scope; we will consider it as a follow-up optimization.

PR #51506, opened 2026-07-17
feat: [Vortex3] support Vortex local format readers

aoiasd:
relate: #50304
pr: https://github.com/milvus-io/milvus/pull/51505

## Summary
Load Storage V3 scalar fields through reusable Vortex local-format readers backed by sparse local files, following the [local format design document](https://github.com/milvus-io/milvus/blob/buqian/aoiasd/aoiasd-vortex-local-format-reader/docs/design-docs/design_docs/20260305-local_format.md). Integrate reader and cache lifecycles with sealed segments, preserve raw fallback and storage error categories, and add focused Vortex reader coverage.

### DIFF
--- a/internal/core/src/mmap/VortexColumn.cpp
+++ b/internal/core/src/mmap/VortexColumn.cpp
@@ -182,6 +182,31 @@ struct VortexReaderRange {
     int64_t range_end;
 };
 
+struct VortexDataScanSource {
+    VortexReaderRange range;
+    PinWrapper<std::shared_ptr<arrow::RecordBatchReader>> reader;
+};
+
+struct VortexRowIdScanSource {
+    VortexReaderRange range;
+    PinWrapper<std::shared_ptr<arrow::RecordBatchReader>> matched_reader;
+    std::optional<PinWrapper<std::shared_ptr<arrow::RecordBatchReader>>>
+        validity_reader;
+};
+
+std::vector<uint64_t>
+MergeCellIds(const std::vector<uint64_t>& left,
+             const std::vector<uint64_t>& right) {
+    std::vector<uint64_t> cell_ids;
+    cell_ids.reserve(left.size() + right.size());
+    cell_ids.insert(cell_ids.end(), left.begin(), left.end());
+    cell_ids.insert(cell_ids.end(), right.begin(), right.end());
+    std::sort(cell_ids.begin(), cell_ids.end());
+    cell_ids.erase(std::unique(cell_ids.begin(), cell_ids.end()),
+                   cell_ids.end());
+    return cell_ids;
+}
+
 std::optional<VortexReaderRange>
 NextVortexReaderRange(const VortexColumn* column,
                       int64_t* scan_pos,
@@ -391,15 +416,11 @@ class VortexColumn::ArrowStringLikeColumn {
 class VortexRowIdScanCursor final : public ChunkedColumnInterface::ScanCursor {
  public:
     VortexRowIdScanCursor(const VortexColumn* column,
-                          milvus::OpContext* op_ctx,
                           int64_t start_offset,
                           int64_t length,
-                          std::string predicate)
-        : column_(column),
-          op_ctx_(op_ctx),
-          predicate_(std::move(predicate)),
-          scan_pos_(start_offset),
-          scan_end_(start_offset + length) {
+                          std::vector<VortexRowIdScanSource>&& sources)
+        : scan_pos_(start_offset),
+          sources_(std::move(sources)) {
         AssertInfo(start_offset >= 0 && length >= 0 &&
                        start_offset + length <=
                            static_cast<int64_t>(column_->NumRows()),
@@ -466,40 +487,31 @@ class VortexRowIdScanCursor final : public ChunkedColumnInterface::ScanCursor {
         if (reader_active_) {
             return true;
         }
-        auto range = NextVortexReaderRange(column_, &scan_pos_, scan_end_);
-        if (!range.has_value()) {
+        if (source_index_ >= sources_.size()) {
             return false;
         }
 
+        const auto& source = sources_[source_index_++];
+        const auto& range = source.range;
+        AssertInfo(range.range_start == scan_pos_,
+                   "vortex row id source starts at {}, expected {}",
+                   range.range_start,
+                   scan_pos_);
+
         reader_active_ = true;
-        reader_chunk_start_ = range->chunk_start;
-        reader_range_start_ = range->range_start;
-        reader_range_end_ = range->range_end;
+        reader_chunk_start_ = range.chunk_start;
+        reader_range_start_ = range.range_start;
+        reader_range_end_ = range.range_end;
         invalid_reader_next_row_id_ = reader_range_start_;
         matched_row_ids_.clear();
         invalid_row_ids_.clear();
         matched_pos_ = 0;
         invalid_pos_ = 0;
-        reader_may_contain_invalids_ = ReaderNeedsValidityStream();
-        matched_reader_ = column_->OpenRowIdScanForFile(op_ctx_,
-                                                        range->chunk_id,
-                                                        range->local_offset,
-                                                        range->length,
-                                                        predicate_);
-        if (reader_may_contain_invalids_) {
-            invalid_reader_ = column_->OpenDataScanForFile(
-                op_ctx_, range->chunk_id, range->local_offset, range->length);
-        }
+        reader_may_contain_invalids_ = source.validity_reader.has_value();
+        matched_reader_ = source.matched_reader;
+        invalid_reader_ = source.validity_reader;
         scan_pos_ = reader_range_end_;
         return true;
-    }
-
-    bool
-    ReaderNeedsValidityStream() const {
-        // VortexColumn does not build ChunkedColumnInterface's materialized
-        // valid-count mapping. For nullable fields, read the Arrow validity
-        // side-stream and merge it with pushed-down row ids.
-        return column_->IsNullable();
     }
 
     void
@@ -718,11 +730,9 @@ class VortexRowIdScanCursor final : public ChunkedColumnInterface::ScanCursor {
         row_ids->emplace_back(row_id);
     }
 
-    const VortexColumn* column_;
-    milvus::OpContext* op_ctx_;
-    std::string predicate_;
     int64_t scan_pos_;
-    int64_t scan_end_;
+    std::vector<VortexRowIdScanSource> sources_;
+    size_t source_index_{0};
     bool reader_active_{false};
     bool reader_may_contain_invalids_{false};
     int64_t reader_chunk_start_{0};
@@ -742,17 +752,17 @@ class VortexRowIdScanCursor final : public ChunkedColumnInterface::ScanCursor {
 class VortexDataScanCursor final : public ChunkedColumnInterface::ScanCursor {
  public:
     VortexDataScanCursor(const VortexColumn* column,
-                         milvus::OpContext* op_ctx,
                          int64_t start_offset,
                          int64_t length,
                          ChunkedColumnInterface::ScanProjection projection,
-                         ChunkedColumnInterface::ScanValueKind value_kind)
+                         ChunkedColumnInterface::ScanValueKind value_kind,
+                         std::vector<VortexDataScanSource>&& sources)
         : column_(column),
-          op_ctx_(op_ctx),
           projection_(projection),
           value_kind_(value_kind),
           scan_pos_(start_offset),
-          scan_end_(start_offset + length) {
+          scan_end_(start_offset + length),
+          sources_(std::move(sources)) {
         AssertInfo(start_offset >= 0 && length >= 0 &&
                        start_offset + length <=
                            static_cast<int64_t>(column_->NumRows()),
@@ -824,7 +834,7 @@ class VortexDataScanCursor final : public ChunkedColumnInterface::ScanCursor {
                 continue;
             }
 
-            if (!OpenReaderForScanPos()) {
+            if (!OpenNextReader()) {
                 break;
             }
         }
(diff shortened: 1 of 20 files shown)


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -8566,6 +8967,192 @@ ChunkedSegmentSealedImpl::LoadColumnGroup(
...
+    const auto base_data_byte_size =
+        vortex_group_file_size / vortex_group_field_count;
+    const auto data_byte_size_remainder =
+        vortex_group_file_size % vortex_group_field_count;
+
+    for (size_t field_index = 0; field_index < milvus_field_ids.size();
+         ++field_index) {
+        const auto field_id = milvus_field_ids[field_index];
+        const auto& field_meta = field_metas.at(field_id);
+        const auto data_byte_size =
```
buqian-zilliz, line 9122: **[P1] Keep physical accounting separate from retrieve-size estimates**

`data_byte_size` is only this logical field's share of the Vortex group's physical, compressed files, but `load_field_data_common` immediately feeds `column->DataByteSize()` into `variable_fields_avg_size`. `Retrieve` enforces `limit_size` as that average times the result row count, so a compressible or unevenly sized `STRING`/`JSON`/`ARRAY` field—and especially an aliased field whose share is divided again—can be underestimated enough to bypass the response limit and exhaust memory. Keep additive physical-file accounting separate from a decoded per-field output-size estimate used by retrieve guards.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -8566,6 +8967,192 @@ ChunkedSegmentSealedImpl::LoadColumnGroup(
...
+        VortexColumnGroupLocalFormat::Vortex) {
+        return false;
+    }
+
+    for (const auto& field_id : milvus_field_ids) {
+        const auto& field_meta = field_metas.at(field_id);
+        AssertInfo(!SystemProperty::Instance().IsSystem(field_id),
+                   "vortex local_format is not supported for system field {}",
+                   field_id.get());
+        AssertInfo(!IsVectorDataType(field_meta.get_data_type()),
```
buqian-zilliz, line 9051: **[P1] Reject recursive struct arrays before selecting Vortex**

`TryLoadVortexColumnGroup` admits every non-vector scalar, including a recursive `ARRAY` subfield in a struct. During `load_field_data_common`, bracket-named struct fields require `ArrayOffsetsSealed::BuildFromColumn`; its recursive-array branch calls `GetChunk`, but `VortexColumn::GetChunk` always throws `Unsupported`. Consequently, a valid Storage V3 segment using `local_format=vortex` for such a field cannot load. Route this field shape through Raw until offset construction uses `ArrayValueViews` or `Scan`, or implement that path before admitting it here.

internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.h
```diff
@@ -188,27 +189,64 @@ class SealedDataGetter : public DataGetter<OutputType> {
...
+        // lifetime. Replace it with candidate batching or a bounded page cache
+        // only as a separate, measured group-by optimization.
+        auto cached = vortex_file_data_.find(chunk_id);
+        if (cached != vortex_file_data_.end()) {
+            return cached->second;
+        }
+
+        const auto start = column_->GetNumRowsUntilChunk(chunk_id);
+        const auto rows = column_->chunk_row_nums(chunk_id);
+        std::vector<int64_t> offsets(rows);
```
buqian-zilliz, line 218: **[P2] Avoid decoding a whole file for one group-by candidate**

The first group-by candidate from a Vortex file allocates offsets for every row, decodes the complete field with `Take`, and retains the resulting `OwnedTakeData` for the getter lifetime. A query that needs only a bounded candidate set can therefore read and decode millions of unrelated rows, and candidates spanning several files retain all of those decoded values—especially expensive for `JSON`. Batch only the actual candidate offsets or use a bounded cell/page cache instead of materializing each visited file.

internal/core/src/mmap/VortexColumnGroup.cpp
```diff
@@ -0,0 +1,379 @@
...
+                file.file_size,
+                file.footer_size);
+        // TODO(local-format): Thread OpContext into footer and later remote
+        // range reads after milvus-storage exposes context-aware APIs. The
+        // current API only permits cancellation checks around, not during, IO.
+        auto open_status = state.footer_reader->Open(source_fs);
+        if (!open_status.ok()) {
+            ThrowVortexStatus(
+                open_status,
+                ErrorCode::DataFormatBroken,
```
buqian-zilliz, line 143: **[P2] Preserve retryable footer-read failures**

`VortexFooterReader::Open(source_fs)` is the remote footer-read boundary, but its status is passed with `DataFormatBroken` as the fallback. `ThrowVortexStatus` only preserves out-of-memory and cancellation, so a transient Arrow I/O error from object storage or the network is reported as corrupt input and may bypass retry handling. Preserve the structured storage error or distinguish source I/O from footer decoding before assigning `DataFormatBroken`.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -8566,6 +8967,192 @@ ChunkedSegmentSealedImpl::LoadColumnGroup(
...
+            file.Get<uint64_t>(milvus_storage::api::kPropertyFileSize, 0),
+            file.Get<uint64_t>(milvus_storage::api::kPropertyFooterSize, 0)});
+    }
+
+    auto group_cache_warmup_policy =
+        getCacheWarmupPolicy(aggregated_warmup_policy,
+                             /*is_vector=*/false,
+                             /*is_index=*/false,
+                             /*in_load_list=*/eager_load);
+    auto vortex_column_group =
```
buqian-zilliz, line 9086: **[P2] Honor the configured Vortex backing policy**

The Vortex branch constructs `VortexColumnGroup` before resolving `Schema::MmapEnabled` or the global scalar mmap policy. The group then always creates `SparseVortexMmapFile`, while this loader later passes `false` as `enable_mmap`; `mmap.populate` is never applied either. Thus disabling mmap still consumes a memfd and VMA per source file, `is_mmap_field` and memory accounting report the wrong backing, and the documented populate setting is inert. Resolve the physical-group policy once and pass it consistently into backing creation and `load_field_data_common`.

internal/core/src/mmap/VortexColumn.cpp
```diff
@@ -0,0 +1,3953 @@
...
+VortexColumn::SupportsScanPushdown(const ScanOptions& options) const {
+    switch (data_type_) {
+        case DataType::STRING:
+        case DataType::VARCHAR:
+            break;
+        default:
+            return false;
+    }
+
+    return options.output == ScanOutput::RowIds &&
```
buqian-zilliz, line 2642: **[P3] Remove or wire the unreachable row-id pushdown stack**

`ScanOptions::ForUnary`, `ForBinaryRange`, `SupportsScanPushdown`, and `ScanOutput::RowIds` have no production execution-layer caller; every active expression scan still constructs `ForData`. The changed design document nevertheless exposes `queryNode.segcore.enableVortexScanPushdown` as an effective setting, so the setting cannot alter a query and the substantial row-id cursor/predicate stack is maintained only for hypothetical use. Either wire configuration and expression execution to this API in this PR, or remove the row-id surface until a real caller exists.

sparknack (commented):
Static review of production code at 508805d546. Existing inline comments already cover whole-file group-by materialization, footer error classification, and mmap policy. Adding one failure-cleanup finding and one ownership question below. Tests and generated files were excluded; no builds or tests were run.

internal/core/src/mmap/VortexColumnGroup.cpp
```diff
@@ -0,0 +1,379 @@
...
+                ErrorCode::FileOpenFailed,
+                fmt::format(
+                    "failed to create vortex group translator for file {}",
+                    file.path));
+        }
+        std::unique_ptr<
+            cachinglayer::Translator<milvus_storage::vortex::VortexCellGuard>>
+            translater = std::move(translater_result).ValueOrDie();
+        state.slot = cachinglayer::Manager::GetInstance().CreateCacheSlot(
+            std::move(translater), op_ctx);
```
sparknack, line 265: **[P1 · Fix in this PR] Roll back sparse-file writes when Cell loading fails**

The newly integrated translator writes all requested ranges before constructing any `VortexCellGuard`. In the pinned dependency, [`get_cells`](https://github.com/milvus-io/milvus-storage/blob/15ab3d7d3f2c222a9ae70bbdae7520b8eb2ca810/cpp/src/format/vortex/vortex_translater.cpp#L164-L185) throws immediately if a later range fails, so no guard punches the ranges already written. During an on-demand read, the segment keeps the memfd alive, leaving resident pages with no successfully loaded Cell to evict. Failures across different Cells can accumulate this memory until retry or unload. Add exception-safe cleanup of the attempted ranges on failure, including partial writes, before integrating this translator.

internal/core/src/mmap/VortexColumn.cpp
```diff
@@ -0,0 +1,3953 @@
...
+                   file_id,
+                   file.start_index,
+                   file.end_index);
+        file_offsets.emplace_back(offset - file.start_index);
+    }
+
+    auto plan = PlanOffsets(file_id, file_offsets);
+    // Take fully materializes decoded Arrow buffers. The cell pin only needs
+    // to protect the reader through import and is released on return.
+    auto pin = PinPlanCells(op_ctx, file_id, plan.cell_ids);
```
sparknack, line 3072: **[Needs confirmation] Does `take()` guarantee buffers independent of the sparse backing?**

This function releases the Cell pin on return, while callers retain the returned Arrow table and string views. The pinned [`VortexFormatReader::take`](https://github.com/milvus-io/milvus-storage/blob/15ab3d7d3f2c222a9ae70bbdae7520b8eb2ca810/cpp/src/format/vortex/vortex_format_reader.cpp#L930-L968) imports an Arrow stream and wraps its batches without an explicit deep copy. Can the Rust implementation guarantee that every returned buffer, including string-view backing buffers, is independent of the sparse mapping for every supported encoding? Otherwise eviction can punch bytes still referenced by the result. Please establish that ownership contract; if it permits borrowed buffers, retain the Cell pin with the result or copy before releasing it.

internal/core/src/mmap/VortexColumnGroup.cpp
aoiasd: Confirmed. The failure cleanup belongs in the pinned milvus-storage VortexTranslater, which owns both the attempted byte ranges and the sparse range file. We will track exception-safe rollback, including the partially written failing range, as a separate milvus-storage follow-up and update the dependency there; this PR will not add a Milvus-side workaround.

internal/core/src/mmap/VortexColumn.cpp
aoiasd: Confirmed: the returned buffers are independent of the sparse mapping. The pinned Rust filesystem bridge allocates a ByteBufferMut and passes its output buffer to loon_filesystem_reader_readat; the C++ bridge calls RandomAccessFile::ReadAt(offset, nbytes, out_data), and SparseVortexMmapFile implements that overload with memcpy. The synchronous take then imports Arrow-owned buffers, and OrderedTakeOwner retains the Arrow tables backing string views. Releasing the Cell pin on return is therefore safe.

internal/core/src/segcore/ChunkedSegmentSealedImpl.h
```diff
@@ -1236,12 +1243,14 @@ class ChunkedSegmentSealedImpl : public SegmentSealed {
...
-                        T* dst_raw,
-                        bool small_int_raw_type = false);
+    bulk_subscript_impl(
+        milvus::OpContext* op_ctx,
+        ChunkedColumnInterface* field,
+        const int64_t* seg_offsets,
+        int64_t count,
+        T* dst_raw,
+        bool small_int_raw_type = false,
+        const ChunkedColumnInterface::OwnedTakeData* taken_data = nullptr);
```
sparknack, line 1253: **Fix in this PR: separate reading from copying an existing Take result.** Adding `taken_data` gives these helpers two alternative data sources: `field + seg_offsets + count`, or an already materialized result. When `taken_data` is present, the read parameters are unused, while the primitive branch iterates `taken_data.size` and the string/array branches iterate `count`. The API leaves callers responsible for keeping those redundant descriptions consistent. Keep the existing read helpers and use separate conversion helpers accepting only `const OwnedTakeData&` and the destination for the pre-read path. This removes the nullable mode switch from each overload without changing the Raw read strategy.

internal/core/src/segcore/ChunkedSegmentSealedImpl.h
```diff
@@ -2024,6 +2055,27 @@ class ChunkedSegmentSealedImpl : public SegmentSealed {
...
+        const std::vector<FieldId>& milvus_field_ids,
+        const std::unordered_map<FieldId, FieldMeta>& field_metas,
+        const SegmentLoadInfo& segment_load_info,
+        const SchemaPtr& schema_snapshot,
+        bool eager_load,
+        const std::string& aggregated_warmup_policy,
+        milvus::OpContext* op_ctx,
+        bool is_replace,
+        RuntimeResourceState* runtime,
+        StagedStateCommitter* committer);
```
sparknack, line 2072: **Non-blocking follow-up: separate Vortex construction from segment publication.** This 13-parameter helper selects the backend, opens files, constructs columns, attributes shared storage bytes, and installs fields. It also accepts two mutually exclusive publication targets (`runtime` and `committer`), enforced only by a runtime assertion. Have the Vortex construction step return the columns to install, and leave installation to the existing segment load/staging flow. That would remove publication-mode knowledge from the backend builder and reduce the overlapping schema/field metadata passed through it; merely wrapping all parameters in one options struct would preserve the same coupling.

internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.h
```diff
@@ -188,27 +189,64 @@ class SealedDataGetter : public DataGetter<OutputType> {
     const segcore::SegmentSealed& segment_;
     const FieldId field_id_;
     bool from_data_;
+    bool from_vortex_column_ = false;
```
sparknack, line 192: **Non-blocking follow-up: keep backend dispatch inside the column access layer.** Beyond the full-file materialization concern already raised, this adds Vortex-specific state and caching to group-by, while expression prefetch and `SegmentChunkReader` also branch on `GetLocalFormat()`. The common column interface still exposes `GetChunk()`, which Vortex rejects, so consumers must know the concrete backend to choose a valid API. As the legacy Chunk consumers are migrated, route these operations through the common Scan/Take contract and keep backend-specific access policy below that boundary. Otherwise each additional local format requires changes across execution operators.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
sparknack: From the interface-design pass, this ambiguity remains at 508805d: `VortexColumn::DataByteSize()` explicitly returns attributed physical file bytes, while `load_field_data_common()` still uses it to update `variable_fields_avg_size`. The interface needs separate contracts for physical storage accounting and decoded logical-size estimates; a caller should not have to inspect `GetLocalFormat()` to interpret the unit/meaning of the same method. If decoded-size metadata is unavailable, represent that explicitly and handle it at the estimate consumer instead of substituting compressed physical bytes. This is the same issue as the original finding, not an additional review item.

Issue #51503, opened 2026-07-17
[Bug]: RESTful v2 add_function/alter_function still forward DDL with nil FunctionSchema after parameter validation fails


bigsheeper:
## Environment

- **Milvus Version**: master (09c44f2dbb)
- **Deployment Mode**: any
- **SDK**: RESTful v2 API

## Background: what `HTTPAbortReturn` actually does

`HTTPAbortReturn` (`internal/distributed/proxy/httpserver/utils.go:78`) is a thin wrapper over gin's `Context.AbortWithStatusJSON`:

```go
func HTTPAbortReturn(c *gin.Context, code int, result gin.H) {
    c.Set(HTTPReturnCode, result[HTTPReturnCode])
    ...
    c.AbortWithStatusJSON(code, result)
}
```

`AbortWithStatusJSON` writes the JSON response immediately and calls `Abort()`, which only sets gin's internal handler index to `abortIndex` so that **pending handlers in the middleware chain** are skipped. It does NOT unwind the Go call stack of the handler that is currently executing — gin's own doc comment states: *"Abort prevents pending handlers from being called. Note that this will not stop the current handler."* (https://pkg.go.dev/github.com/gin-gonic/gin#Context.Abort)

Therefore every call site must be followed by an explicit `return`, which is the pattern used throughout `handler_v2.go` (e.g. `createCollection`, `addCollectionField`).

### Steps

1. Create a collection.
2. `POST /v2/vectordb/collections/add_function` (or `alter_function`) with a function whose `type` is an invalid string (any value `genFunctionSchema` rejects).
3. Observe proxy logs and the HTTP response body.

### Trigger Conditions

- **Frequency**: always, with any request whose function schema fails validation

## Expected Behavior

The handler rejects the request immediately: exactly one error response is written, and no `AddCollectionFunction`/`AlterCollectionFunction` RPC is issued.

## Actual Behavior

In `addCollectionFunction` and `alterCollectionFunction` (`handler_v2.go`), the error branch calls `HTTPAbortReturn` but is missing the `return nil, err`. Since `HTTPAbortReturn` cannot stop the current handler (see Background above), execution falls through after the error response has been written:

- The DDL RPC is still sent with `FunctionSchema = nil` (rejected later by the task's `PreExecute` nil-check — a wasted dd-queue round trip, and the real validation error is replaced by a misleading "function schema is nil" warning in proxy logs).
[...]

PR #51511, opened 2026-07-17
fix: Return early after parameter validation fails in RESTful v2 function DDL handlers

bigsheeper:
issue: #51503

`addCollectionFunction` and `alterCollectionFunction` in the RESTful v2 handler call `HTTPAbortReturn` when `genFunctionSchema` rejects the request, but are missing the early `return`. gin's `Abort` only skips pending handlers in the middleware chain — it does not stop the currently executing handler (see [gin docs](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Abort)) — so execution falls through:

- the DDL RPC is still issued with a nil `FunctionSchema` (rejected later by the task's `PreExecute` nil-check — a wasted dd-queue round trip, with the real validation error replaced by a misleading "function schema is nil" warning);
- a second JSON response is written onto the already-committed connection; under the REST timeout middleware this surfaces to the client as a 408 "request timeout" after the full timeout instead of the immediate validation error (observed in the regression test before the fix).

This PR adds the missing `return nil, err` after `HTTPAbortReturn` in both handlers, matching every sibling handler in the file, and adds regression tests asserting that an invalid function type is rejected immediately without issuing any RPC (strict mock: an unexpected `AddCollectionFunction`/`AlterCollectionFunction` call fails the test).

These two handlers are the only defective call sites — verified by sweeping all 101 `HTTPAbortReturn` call sites in the package.

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -961,6 +961,7 @@ func (h *HandlersV2) addCollectionFunction(ctx context.Context, c *gin.Context, 
 			HTTPReturnCode:    merr.Code(merr.ErrParameterInvalid),
 			HTTPReturnMessage: err.Error(),
 		})
+		return nil, err
 	}
 
 	req.FunctionSchema = fSchema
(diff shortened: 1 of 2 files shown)


czs007:

bigsheeper:
rerun cpp-unit-test

czs007 (approved):
/lgtm

Issue #51513, opened 2026-07-17
[Bug]: TiKV kv layer: prefix scans ignore caller context, MultiLoad mutates caller's keys slice, txn latency metric always ~0


shaoting-huang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (85c3aeb8c0)
- Deployment mode(standalone or cluster): any deployment with metastore.type=tikv
- MQ type(rocksmq, pulsar or kafka): any
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): any
- CPU/Memory: N/A
- GPU: N/A
- Others: all three defects are in internal/kv/tikv/txn_tikv.go
```

### Current Behavior

Three latent defects in the TiKV metastore kv layer, found during a hardening audit:

**1. Prefix scans cannot be canceled.** `HasPrefix`, `LoadWithPrefix` and `WalkWithPrefix` accept a `ctx` but never consult it. client-go snapshot iterators structurally cannot take a context — their internal backoffer is created with `context.Background()`, and the ServerBusy backoff class is exempted from the usual cap, allowing up to ~10 minutes of backoff sleep. A caller that has timed out, disconnected, or is shutting down leaves an unkillable scan running against TiKV. The write-path prefix-delete scan inside `MultiSaveAndRemoveWithPrefix` (txn.Iter over each removal prefix) has the same property. On large prefixes (e.g. segment meta during recovery) this turns one canceled request into minutes of wasted scan work.

**2. MultiLoad mutates the caller's keys slice.** `batchConvertFromString` writes the rootPath-prefixed path back into the caller's slice (`keys[i] = util.GetPath(prefix, keys[i])`), and MultiLoad's result lookup silently depends on that mutation. Any caller that legally reuses its slice across calls (e.g. a retry loop) gets a double-prefixed miss on the second call, reported as key-not-found for keys that exist.

**3. Transaction latency metric is always ~0.** `executeTxn` reads the stopwatch (`elapsed := start.ElapseSpan()`) BEFORE calling `commitTxn`, so the `MetaTxnLabel` latency histogram observes the pre-commit elapsed (~0) instead of the commit duration. Every Multi* write transaction's real latency — the dominant meta write cost on TiKV (2PC + TSO round trips) — is invisible: the histogram count grows while the sum stays near zero. This systematically hides the exact cost that capacity planning for the TiKV backend needs to see.

[...]

PR #51514, opened 2026-07-17
fix: [Catalog] harden TiKV KV scans and write metrics

shaoting-huang:
- Honor caller cancellation before and during TiKV prefix scans. Read-path scans check `ctx.Err()` before opening snapshot iterators and while iterating; write-path prefix deletes now also check immediately before the contextless `txn.Iter`, including deadlines that expire during non-ctx-bound `beginTxn`.
- Avoid mutating `MultiLoad` callers' key slices by building prefixed byte keys separately, while still reporting missing keys under the callers' original names.
- Measure TiKV transaction latency after commit completes so `MetaTxnLabel` captures the commit path instead of the pre-commit elapsed time.
- Add hardening tests for `MultiLoad` slice reuse, canceled prefix scans/deletes, begin-time cancellation before prefix-delete scans, and transaction latency metrics.

related: #51513

### DIFF
--- a/internal/kv/tikv/txn_tikv.go
+++ b/internal/kv/tikv/txn_tikv.go
@@ -227,6 +227,12 @@ func (kv *txnTiKV) HasPrefix(ctx context.Context, prefix string) (bool, error) {
 	var loggingErr error
 	defer logWarnOnFailure(&loggingErr, "txnTiKV HasPrefix() error", mlog.String("prefix", prefix))
 
+	// The snapshot iterator below does not accept a context, so honor
+	// caller cancellation explicitly before starting the scan.
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return false, ctxErr
+	}
+
 	ss := getSnapshot(kv.txn, SnapshotScanSize)
 
 	// Retrieve bounding keys for prefix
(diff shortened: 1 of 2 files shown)


internal/kv/tikv/txn_tikv.go
```diff
@@ -559,8 +578,14 @@ func (kv *txnTiKV) MultiSaveAndRemoveWithPrefix(ctx context.Context, saves map[s
 				}
 				defer iter.Close()
 
-				// Iterate over keys and delete them
+				// Iterate over keys and delete them. The transaction
+				// iterator does not accept a context either, so poll
+				// caller cancellation each step; the bare context error
+				// surfaces to the caller through the build-error path.
 				for iter.Valid() {
+					if ctxErr := ctx.Err(); ctxErr != nil {
```
liliu-z, line 592: The read prefix scans each added an entry `ctx.Err()` check immediately before creating the iterator, but the write path only polls inside the loop (here), after `txn.Iter` at line 575 has already launched the removal scan. `retry.Do` catches the already-canceled case and the predicate `txn.Get` is itself ctx-bound, but `beginTxn`'s startTS fetch is not (the comment at 112-116 notes a ~15s TSO budget that can outlive the 10s requestTimeout), so a deadline expiring during Begin still starts the scan on a dead context — up to ~10min under sustained ServerBusy. Add a `ctx.Err()` check before `txn.Iter` to match the read paths.

shaoting-huang: Fixed in 7b93afd27d by adding a `ctx.Err()` check immediately before `txn.Iter` in `MultiSaveAndRemoveWithPrefix`. Added `TestMultiSaveAndRemoveWithPrefixStopsPrefixScanWhenContextCanceledDuringBegin`, which cancels from the injected `beginTxn` wrapper and verifies the prefix-removal path returns `context.Canceled` before `commitTxn` is reached.

Issue #51512, opened 2026-07-17
[Bug]: datacoord GC recycleUnusedAnalyzeFiles builds a wrong remove prefix, stale analyze stats versions are never reclaimed


bigsheeper:
## Environment

- **Milvus Version**: master (85c3aeb8c0)
- **Deployment Mode**: any (requires clustering compaction / analyze tasks)

## Background: on-disk layout of analyze stats files

Analyze (clustering) result files are written by the worker with taskID as the first path segment after the root (`internal/core/src/clustering/KmeansClustering.h`, `build_id` = taskID, `index_version` = version; same layout used by `meta.CleanPartitionStatsInfo`):

```
{rootPath}/analyze_stats/{taskID}/{version}/{collectionID}/{partitionID}/{fieldID}/centroids
{rootPath}/analyze_stats/{taskID}/{version}/{collectionID}/{partitionID}/{fieldID}/{segmentID}/offset_mapping
```

A task's `Version` increments every time the task is assigned to a worker (`analyzeTask.UpdateVersion`), so a retried task leaves files under multiple version prefixes; only the latest is live.

## Actual Behavior

In `garbage_collector.go` `recycleUnusedAnalyzeFiles`, the loop that is supposed to "remove analyze stats files which version is less than current task" builds the prefix like this:

```go
var i int64
for i = 0; i < task.Version; i++ {
    removePrefix := prefix + fmt.Sprintf("%d/", task.Version)
    if err := gc.option.cli.RemoveWithPrefix(ctx, removePrefix); err != nil {
```

Two defects in one line:

1. **The taskID is missing** — the resulting prefix is `{rootPath}/analyze_stats/{version}/`, but the first segment on disk is the taskID (a large allocated ID). The prefix matches no objects, so `RemoveWithPrefix` is a no-op.
2. **The loop variable `i` is unused** — even with the taskID present, it would target `task.Version` (the live version that must be kept) rather than each stale version `i`, repeated `task.Version` times.

The sibling loops in the same file for text-index and JSON-stats files correctly build `{buildID}/{i}/`-style prefixes, confirming the intent.

## Expected Behavior

For a finished analyze task still present in meta, all files under `analyze_stats/{taskID}/{v}/` with `v < task.Version` are removed, e.g.:

```go
removePrefix := prefix + fmt.Sprintf("%d/%d/", taskID, i)
```

## Impact

[...]

PR #51516, opened 2026-07-17
fix: recycle stale analyze stats versions and persist analyze terminal states

bigsheeper:
issue: #51512

`recycleUnusedAnalyzeFiles` built the remove prefix as `analyze_stats/{task.Version}/`: the taskID was missing and the loop variable was unused, so the prefix never matched any object and the loop removed the same nonexistent prefix `Version` times.

Analyze stats files are written as `analyze_stats/{taskID}/{version}/...` (see `KmeansClustering.h` `GetRemoteCentroidsObjectPrefix` / `GetRemoteCentroidIdMappingObjectPrefix`; `build_id` = taskID), and a task's `Version` increments on every worker (re)assignment — so every stale-version centroids/offset_mapping file from retried analyze tasks leaked in object storage until the task was dropped from meta entirely (only the `task == nil` full-prefix removal branch actually deleted anything).

This PR builds the prefix as `analyze_stats/{taskID}/{i}/` to match the on-disk layout (and the sibling text-index/JSON-stats cleanup loops), and adds a regression test asserting that a finished task at `Version=2` removes exactly versions 0 and 1 under its own task prefix — before the fix, the recorded removes were `analyze_stats/2/` twice.

## Persisting analyze terminal states

Adversarial review surfaced that the cleanup above was unreachable for one class of analyze task. `CreateTaskOnWorker` can drive a task straight to a terminal state before dispatch — the input segment is gone, or `numClusters < ClusteringCompactionMinCentroidsNum` — and both branches used `SetState`, which mutates only the `proto.Clone`d, scheduler-owned copy. The only persisting path is `UpdateStateWithMeta` → `analyzeMeta.UpdateState` → `saveTask`; `UpdateVersion` persists `Version`/`NodeID` only. The scheduler's terminal branch just drops the task, so meta stayed at `JobStateInit`:

- `processAnalyzing` keeps reading `Init`, falls through to `default:` and never transitions, so the clustering compaction stays in `analyzing` forever with its input segments pinned as `compacting`; there is no `analyzing` timeout.
- `CheckCleanAnalyzeTask` only returns `canRecycle=true` for `Finished`, so the stale-version cleanup added by this PR never ran for those tasks.

Both branches now use `UpdateStateWithMeta`. If persistence fails the in-memory state is left untouched, so the scheduler re-enqueues the task instead of dropping it.

Not addressed here: even with the state persisted, `CheckCleanAnalyzeTask` still returns `canRecycle=false` for `Failed`, so recycling the files of failed analyze tasks needs a separate change.

## Verification

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord/...` — ok (all five packages)
- The three terminal-state assertions fail against the previous `SetState` implementation.

### DIFF
--- a/internal/datacoord/garbage_collector.go
+++ b/internal/datacoord/garbage_collector.go
@@ -1651,7 +1651,8 @@ func (gc *garbageCollector) recycleUnusedAnalyzeFiles(ctx context.Context, signa
 				// process canceled.
 				return
 			}
-			removePrefix := prefix + fmt.Sprintf("%d/", task.Version)
+			// analyze stats files are laid out as analyze_stats/{taskID}/{version}/...
+			removePrefix := prefix + fmt.Sprintf("%d/%d/", taskID, i)
 			if err := gc.option.cli.RemoveWithPrefix(ctx, removePrefix); err != nil {
 				mlog.Warn(ctx, "garbageCollector recycleUnusedAnalyzeFiles remove files with prefix failed",
 					mlog.Int64("taskID", taskID), mlog.String("removePrefix", removePrefix))
--- a/internal/datacoord/garbage_collector_test.go
+++ b/internal/datacoord/garbage_collector_test.go
@@ -56,6 +56,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/objectstorage"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
+	"github.com/milvus-io/milvus/pkg/v3/proto/indexpb"
 	"github.com/milvus-io/milvus/pkg/v3/proto/workerpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/lock"
@@ -3432,6 +3433,66 @@ func TestGarbageCollector_recycleUnusedTextIndexFiles_SnapshotReference(t *testi
 	// Verify - files should NOT be removed
 	assert.Empty(t, removedFiles,
 		"text index files should not be removed when segment is referenced by snapshot")
+}
+
+func TestGarbageCollector_recycleUnusedAnalyzeFiles_StaleVersions(t *testing.T) {
+	ctx := context.Background()
+
+	meta := &meta{
+		catalog: &datacoord.Catalog{},
+		analyzeMeta: &analyzeMeta{
+			ctx: ctx,
+			tasks: map[int64]*indexpb.AnalyzeTask{
+				401: {
+					TaskID:  401,
+					Version: 2,
+					State:   indexpb.JobState_JobStateFinished,
+				},
+			},
+		},
+	}
+
+	cli := storage.NewLocalChunkManager(objectstorage.RootPath("/tmp/test"))
+
+	gc := newGarbageCollector(meta, &ServerHandler{}, GcOption{
+		cli:              cli,
+		enabled:          true,
+		checkInterval:    time.Millisecond * 10,
+		scanInterval:     time.Hour * 7 * 24,
+		missingTolerance: 0,
+		dropTolerance:    time.Hour * 24,
+	})
+
+	// The non-recursive walk lists the per-task top-level dirs under analyze_stats.
+	mockWalk := mockey.Mock((*storage.LocalChunkManager).WalkWithPrefix).To(
+		func(cm *storage.LocalChunkManager, ctx context.Context, prefix string,
+			recursive bool, fn storage.ChunkObjectWalkFunc,
+		) error {
+			if strings.Contains(prefix, common.AnalyzeStatsPath) {
+				fn(&storage.ChunkObjectInfo{
+					FilePath:   path.Join(prefix, "401") + "/",
+					ModifyTime: time.Now().Add(-time.Hour),
+				})
+			}
+			return nil
+		}).Build()
+	defer mockWalk.UnPatch()
+
+	removedPrefixes := []string{}
+	mockRemove := mockey.Mock((*storage.LocalChunkManager).RemoveWithPrefix).To(
+		func(cm *storage.LocalChunkManager, ctx context.Context, prefix string) error {
+			removedPrefixes = append(removedPrefixes, prefix)
+			return nil
+		}).Build()
+	defer mockRemove.UnPatch()
+
+	gc.recycleUnusedAnalyzeFiles(ctx, nil)
+
+	// Analyze files are written under analyze_stats/{taskID}/{version}/... — for a
+	// finished task at Version=2, exactly the stale versions 0 and 1 under the
+	// task's own prefix must be removed, and version 2 must be kept.
+	root := path.Join(cli.RootPath(), common.AnalyzeStatsPath) + "/"
+	assert.ElementsMatch(t, []string{root + "401/0/", root + "401/1/"}, removedPrefixes)
 }
 
 func TestGarbageCollector_recycleUnusedJSONIndexFiles_SnapshotReference(t *testing.T) {
--- a/internal/datacoord/task_analyze.go
+++ b/internal/datacoord/task_analyze.go
@@ -171,7 +171,11 @@ func (at *analyzeTask) CreateTaskOnWorker(nodeID int64, cluster session.Cluster)
 		if info == nil {
 			log.Warn(context.TODO(), "analyze task is processing, but segment is nil, fail the task",
 				mlog.FieldSegmentID(segID))
-			at.SetState(indexpb.JobState_JobStateFailed, fmt.Sprintf("segmentInfo with ID: %d is nil", segID))
+			if err := at.UpdateStateWithMeta(indexpb.JobState_JobStateFailed,
+				fmt.Sprintf("segmentInfo with ID: %d is nil", segID)); err != nil {
+				// State is left untouched, so the scheduler re-enqueues the task and retries.
+				log.Warn(context.TODO(), "failed to persist the failed state of the analyze task", mlog.Err(err))
+			}
 			return
 		}
 		totalSegmentsRows += info.GetNumOfRows()
@@ -202,7 +206,10 @@ func (at *analyzeTask) CreateTaskOnWorker(nodeID int64, cluster session.Cluster)
 						mlog.Float64("raw data size", totalSegmentsRawDataSize),
 						mlog.Int64("num clusters", numClusters),
 						mlog.Int64("minimum num clusters required", Params.DataCoordCfg.ClusteringCompactionMinCentroidsNum.GetAsInt64()))
-					at.SetState(indexpb.JobState_JobStateFinished, "")
+					if err := at.UpdateStateWithMeta(indexpb.JobState_JobStateFinished, ""); err != nil {
+						// State is left untouched, so the scheduler re-enqueues the task and retries.
+						log.Warn(context.TODO(), "failed to persist the finished state of the analyze task", mlog.Err(err))
+					}
 					return
 				}
 				if numClusters > Params.DataCoordCfg.ClusteringCompactionMaxCentroidsNum.GetAsInt64() {
--- a/internal/datacoord/task_analyze_test.go
+++ b/internal/datacoord/task_analyze_test.go
@@ -203,6 +203,12 @@ func (s *analyzeTaskSuite) newTask() *analyzeTask {
 	}, s.mt)
 }
 
+// restoreMetaTask puts back the task the suite was set up with, so a test that
+// persists a state transition does not leak it into the following tests.
+func (s *analyzeTaskSuite) restoreMetaTask(task *indexpb.AnalyzeTask) {
+	s.mt.analyzeMeta.tasks[s.taskID] = task
+}
+
 func (s *analyzeTaskSuite) TestCreateTaskOnWorker_SegmentNil() {
 	// Replace segment 102 with a dropped segment so it's filtered out by isSegmentHealthy
 	s.mt.segments.SetSegment(102, &SegmentInfo{
@@ -230,10 +236,14 @@ func (s *analyzeTaskSuite) TestCreateTaskOnWorker_SegmentNil() {
 	catalog := catalogmocks.NewDataCoordCatalog(s.T())
 	catalog.On("SaveAnalyzeTask", mock.Anything, mock.Anything).Return(nil)
 	s.mt.analyzeMeta.catalog = catalog
+	defer s.restoreMetaTask(s.mt.analyzeMeta.tasks[s.taskID])
 
 	at.CreateTaskOnWorker(1, session.NewMockCluster(s.T()))
 	s.Equal(indexpb.JobState_JobStateFailed, at.GetState())
 	s.Contains(at.GetFailReason(), "102")
+	// The terminal state must be persisted, not only set on the scheduler-owned copy.
+	s.Equal(indexpb.JobState_JobStateFailed, s.mt.analyzeMeta.GetTask(s.taskID).GetState())
+	s.Contains(s.mt.analyzeMeta.GetTask(s.taskID).GetFailReason(), "102")
 }
 
 func (s *analyzeTaskSuite) TestCreateTaskOnWorker_DimExtractionError() {
@@ -274,10 +284,36 @@ func (s *analyzeTaskSuite) TestCreateTaskOnWorker_DataTooSmall() {
 	catalog := catalogmocks.NewDataCoordCatalog(s.T())
 	catalog.On("SaveAnalyzeTask", mock.Anything, mock.Anything).Return(nil)
 	s.mt.analyzeMeta.catalog = catalog
+	defer s.restoreMetaTask(s.mt.analyzeMeta.tasks[s.taskID])
 
 	at.CreateTaskOnWorker(1, session.NewMockCluster(s.T()))
 	// data too small → skip → mark as finished
 	s.Equal(indexpb.JobState_JobStateFinished, at.GetState())
+	// Persisting Finished is what lets the GC recycle the task's analyze stats files.
+	s.Equal(indexpb.JobState_JobStateFinished, s.mt.analyzeMeta.GetTask(s.taskID).GetState())
+}
+
+func (s *analyzeTaskSuite) TestCreateTaskOnWorker_TerminalStateNotPersisted() {
+	// Set MinCentroidsNum very high so data is considered too small
+	origMin := Params.DataCoordCfg.ClusteringCompactionMinCentroidsNum.SwapTempValue("999999999")
+	defer Params.DataCoordCfg.ClusteringCompactionMinCentroidsNum.SwapTempValue(origMin)
+
+	at := s.newTask()
+	catalog := catalogmocks.NewDataCoordCatalog(s.T())
+	// The first save is UpdateVersion, the second one is the terminal state transition.
+	catalog.On("SaveAnalyzeTask", mock.Anything, mock.Anything).Return(nil).Once()
+	catalog.On("SaveAnalyzeTask", mock.Anything, mock.Anything).
+		Return(merr.WrapErrServiceInternalMsg("mock save error")).Once()
+	catalog.On("SaveAnalyzeTask", mock.Anything, mock.Anything).Return(nil)
+	s.mt.analyzeMeta.catalog = catalog
+	defer s.restoreMetaTask(s.mt.analyzeMeta.tasks[s.taskID])
+	stateBefore := s.mt.analyzeMeta.GetTask(s.taskID).GetState()
+
+	at.CreateTaskOnWorker(1, session.NewMockCluster(s.T()))
+	// A failed persistence leaves the task at Init so the scheduler re-enqueues it,
+	// rather than dropping it on an in-memory-only terminal state.
+	s.Equal(indexpb.JobState_JobStateInit, at.GetState())
+	s.Equal(stateBefore, s.mt.analyzeMeta.GetTask(s.taskID).GetState())
 }
 
 func (s *analyzeTaskSuite) TestCreateTaskOnWorker_NumClustersCapped() {


czs007:
Analyze thoroughly. Reflect deeply. Explore fully.

Adversarial review found one non-blocking efficiency nit; the loop over stale analyze-stats versions starts at a version that is never written to disk.

### Low

**`internal/datacoord/garbage_collector.go:1653`** — The stale-version cleanup loop `for i = 0; i < task.Version; i++` starts at version 0, but analyze stats are never persisted under version 0: an `AnalyzeTask` is created without a `Version` field (defaulting to 0), and `UpdateVersion` bumps the persisted version to 1 before the first worker dispatch, so the earliest on-disk directory is `analyze_stats/{taskID}/1/`. The `i=0` iteration therefore always targets a nonexistent prefix, costing one extra list+delete round trip to object storage per task per GC scan. This is harmless for correctness (`RemoveWithPrefix` on a missing prefix is a no-op), but worth noting because this change turns these calls from universally-no-op into real object-storage traffic. Suggestion: start the loop at `i = 1` to drop the dead call; the new test's expected set would then become just `401/1/`. *(raised by adversarial review)*

bigsheeper:
Re: the Low finding on the loop starting at `i = 0` — verified and intentionally keeping it as is.

Confirmed the premise: version 0 is indeed never written. Across master, 2.5, and 2.4, `UpdateVersion` always runs before dispatch and the worker request version is always ≥ 1, so the earliest on-disk directory is `analyze_stats/{taskID}/1/` (writing version 0 would require a task version of -1, which is impossible since versions only increment and taskIDs are never reused).

Choosing to keep `i = 0` anyway:

- The cost is one no-op `RemoveWithPrefix` (a LIST on a nonexistent prefix) per retained finished task per GC scan — noise-level object-storage traffic given how few analyze tasks stay in meta.
- `for i = 0; i < task.Version` states "all versions below current" with no preconditions, while `i = 1` bakes in a cross-component invariant (scheduler → worker → C++ writer) that readers must know to trust the loop. The bet is also asymmetric: if that invariant ever changed, `i = 0` wastes one call, `i = 1` silently leaks — the very bug this PR fixes.

Happy to revisit if maintainers prefer consistency with the text-index loop (which starts at 1).

czs007:
Adversarial review surfaced one pre-existing issue adjacent to this PR's GC fix: two terminal-state paths in the analyze task never persist their state, which keeps the affected tasks permanently invisible to the stale-version cleanup this PR improves.

## Medium

**`internal/datacoord/task_analyze.go:174` (and `:205`) — terminal state set only in memory, meta stays at `Init`**

*Problem:* Both terminal paths in `CreateTaskOnWorker` — segment is nil → `Failed` (`:174`), and `numClusters` below `ClusteringCompactionMinCentroidsNum` → `Finished` (`:205`) — call `at.SetState(...)` and return. `SetState` (`task_analyze.go:88-91`) only mutates in-memory fields; persistence requires `UpdateStateWithMeta` → `analyzeMeta.UpdateState` → `saveTask` (`task_analyze.go:93-99`, `analyze_meta.go:144-160`), which is exactly what the `InProgress` path at `:240` does. Worse, the `analyzeTask` holds a `proto.Clone` of the meta object (`analyze_inspector.go:59-62`, `compaction_task_clustering.go:762`) and `saveTask` replaces the map entry with yet another clone (`analyze_meta.go:73-79`), so `SetState` does not even reach the object in `analyzeMeta.tasks`. Meanwhile `UpdateVersion` (`task_analyze.go:142`) has already persisted the version bump, so the record on disk ends up with a bumped version and `State == Init`. The scheduler then drops the task based on the in-memory terminal state (`task/global_scheduler.go:301-310` — neither `runningTasks` nor `pendingTasks`), so nothing ever advances it again.

Two consequences:
1. `CheckCleanAnalyzeTask` only reports recyclable when `State == Finished` (`analyze_meta.go:190-200`), so `recycleUnusedAnalyzeFiles` always takes the `canRecycle == false` branch for these tasks (`garbage_collector.go:1624-1628`) — the stale-version cleanup fixed by this PR never applies to them.
2. More seriously, for the `:205` path, `clusteringCompactionTask.processAnalyzing` reads the persisted meta (`compaction_task_clustering.go:617-637`); `State == Init` falls into the `default` branch and returns `nil`, leaving the clustering compaction stuck in `analyzing` until a timeout or cleanup path intervenes.

*Scope note:* This is pre-existing, not introduced here — the PR diff touches only `garbage_collector.go` (+2/-1) and `garbage_collector_test.go` (+61/-0). Neither path writes stats files, so it does not weaken the leak fix in this PR. It is raised because it bounds the effectiveness of the cleanup this PR is improving.

*Suggestion:* Replace both `SetState` calls with `UpdateStateWithMeta` and handle the returned error (log and return without dropping the task, consistent with the `:240` path). Can reasonably be split into a follow-up PR.

(raised by xiaocai2333)

bigsheeper:
Thanks — verified the whole chain against the current head (`49bc433`). The finding holds. Not changing it in this PR; rationale at the end.

**Mechanism confirmed, every step:**

- `SetState` (`task_analyze.go:88-91`) only mutates the in-memory fields; persistence goes through `UpdateStateWithMeta` (`:93-99`) → `analyzeMeta.UpdateState` → `saveTask` → `catalog.SaveAnalyzeTask`. The success path at `:240` uses exactly that, so the contrast is real.
- Both terminal paths — segment nil → `Failed` (`:174`), and `numClusters` below `ClusteringCompactionMinCentroidsNum` → `Finished` (`:205`) — call `SetState` and return.
- The object-identity point is right too: `analyzeTask` holds a `proto.Clone` (`analyze_inspector.go:60`, `compaction_task_clustering.go:762`) and `saveTask` swaps the map entry for another clone (`analyze_meta.go:74-78`), so `SetState` never reaches `analyzeMeta.tasks`.
- `UpdateVersion` (`:142`) has already persisted the bump, so the record ends up version-bumped with `State == Init`, and the scheduler then drops the task based on the in-memory terminal state (`task/global_scheduler.go:301-310`).
- Both consequences check out: `CheckCleanAnalyzeTask` gates on `State == Finished` (`analyze_meta.go:190-200`), so GC always takes the `canRecycle == false` branch (`garbage_collector.go:1624-1628`); and `processAnalyzing` reads the persisted meta, where `Init` falls into `default:` and returns nil (`compaction_task_clustering.go:625-638`).

**Two refinements to the writeup:**

1. "nothing ever advances it again" holds within a single datacoord lifetime. Across a restart, `analyzeInspector.reloadFromMeta` (`analyze_inspector.go:52-64`) re-enqueues anything persisted as `Init`; it then deterministically re-hits the same branch and is dropped again. So the effect is a permanent livelock rather than a one-shot wedge — same conclusion, more precise mechanism.
2. On "until a timeout or cleanup path intervenes": `QueryTaskOnWorker` early-returns for `analyzing` (`compaction_task_clustering.go:136-142`), and I did not find a datacoord-side path that drives `analyzing` to `timeout`. I'm not claiming none exists — I just couldn't trace one, so I'd treat the compaction as stuck until shown otherwise.

Also noting a third `SetState` site that is **not** a defect: `:192` (`GetDimFromParams` failure) sets `Init`, which is non-terminal, so the scheduler pushes the task back to `pendingTasks` and the persisted state already matches. Only `failReason` is lost. No change needed there.

**Why not in this PR** — this PR is a +2/-1 fix for building the wrong remove prefix, with a regression test scoped to exactly that. The finding fixes a different thing: whether a task can *reach* a recyclable state at all. Changing the persistence timing at `:174`/`:205` moves the clustering-compaction state machine (consequence 2), which needs its own tests, its own issue, and its own cherry-pick risk assessment. That's a separate change, and folding it in here would widen the diff and the review surface with no coupling benefit.

One addition to your scope note, which is otherwise accurate: neither path writes stats files itself, so this doesn't weaken the leak fix here. But a task that *did* run on a worker at earlier versions and only later hits `:174` leaks those earlier versions' files permanently — which is exactly the bound you're describing.

czs007:
## Re-review of `49bc433ea267`

- **Confirmed** — `internal/datacoord/task_analyze.go:174` (terminal states set via `SetState` instead of `UpdateStateWithMeta`, leaving persisted state at `Init`): still stands; both reviewers re-traced the full chain locally and every link holds, so it is reported again below with the second terminal path (`:205`) now included.

---

One previously reported medium-severity correctness issue in the analyze task terminal paths remains confirmed; the stale-version prefix fix in this PR is itself sound, with no new critical or high-severity findings.

### Medium

**`internal/datacoord/task_analyze.go:174`** (also `:205`) — Both terminal paths in `CreateTaskOnWorker` — missing input segment (`:174`) and too few clusters (`:205`) — call `SetState` (`task_analyze.go:88-91`), which only mutates `at.State`/`at.FailReason` on the scheduler-owned clone (`analyze_inspector.go:59-62`, `compaction_task_clustering.go:762`). The only persisting path is `UpdateStateWithMeta` (`task_analyze.go:93-99`) → `analyzeMeta.UpdateState` → `saveTask` (`analyze_meta.go:146-162`); the earlier `UpdateVersion` call persists only `Version`/`NodeID` and leaves `State` untouched (`analyze_meta.go:113-128`), so metadata stays at the `JobStateInit` value written at task creation (`compaction_task_clustering.go:754`). The scheduler reads that in-memory clone after `CreateTaskOnWorker` returns and, on a terminal state, only calls `backoffs.Remove` — no persistence, no re-enqueue (`task/global_scheduler.go:286-310`) — and `analyzeInspector` reloads from meta only once at `Start()` (`analyze_inspector.go:44-46`), so nothing ever repairs the record.

Failure scenario: a clustering compaction's analyze task hits either terminal short-circuit — an input segment is dropped/compacted away before scheduling, or `numClusters < ClusteringCompactionMinCentroidsNum` → the terminal state exists only in the discarded clone, so `processAnalyzing` keeps reading `Init` and falls through to `default:` (`compaction_task_clustering.go:617,624-637`), `Process` never returns true and `analyzing` has no timeout branch (`:293-296,321-330`), leaving the clustering compaction stuck in `analyzing` forever with its input segments pinned as `compacting`; the same stale `Init` makes `CheckCleanAnalyzeTask` always return `canRecycle=false` (`analyze_meta.go:194-198`), so the stale-version cleanup fixed by this PR never runs for those tasks and their `centroids`/`offset_mapping` files linger in object storage, and a datacoord restart re-enqueues from `Init` straight back into the same short-circuit.

Suggestion: use `UpdateStateWithMeta` for both terminal transitions and handle persistence failures so the task stays retryable, and assert catalog/meta state rather than only the in-memory state in the corresponding tests. Note that `SetState(JobStateInit, ...)` at `task_analyze.go:192` is not affected — `Init` matches the persisted state and the scheduler's `Init` branch re-enqueues the task.

Note: this is a pre-existing defect outside this PR's diff (which only touches `garbage_collector.go`), but it interacts directly with the analyze GC path this PR fixes.

*(raised by tinswzy, chyezh)*

bigsheeper:
Thanks — this round closed the one gap I had explicitly left open, and it surfaced a consequence neither of us had. Both independently re-verified below. My position on scope is unchanged, and I'd like to settle it explicitly rather than re-litigate it each round.

**You answered my open question.** In my previous reply I said I could not find a datacoord-side path driving an `analyzing` compaction to `timeout`, and deliberately did not claim none existed. Your `:293-296` / `:321-330` pointers let me close it, and I re-traced it myself:

- `retryableProcess`'s switch routes `analyzing` only to `processAnalyzing()`, which returns nil on `Init` — no timeout arm.
- `Process` returns true only for `completed` / `cleaned` / `failed` / `timeout`, so `analyzing` never terminates.
- The only site that *writes* `CompactionTaskState_timeout` for this task type is `compaction_task_clustering.go:215`, inside the switch over the result state returned by the DataNode — and `QueryTaskOnWorker` early-returns for `analyzing` (`:136-142`), so the DataNode is never queried and that switch is unreachable.
- `compaction_inspector` only counts `timeout` for metrics (`:104-156`) and moves already-timed-out tasks to cleaning (`:650`); it does not originate a timeout.

So "stuck until something intervenes" is now "stuck, full stop." That is a real strengthening of the finding, and I'm recording it as such.

**The pinned-segments consequence is correct, and it matters more than the file leak.** I verified it: `resetSegmentCompacting` has exactly one caller, the last step of `doClean` (`:738-740`), which only runs once the task reaches a terminal state via `Clean()`. A task wedged in `analyzing` never gets there, so its input segments keep `compacting = true` indefinitely and are excluded from any future compaction. That is a compounding scheduling defect, not just orphaned `centroids` / `offset_mapping` objects in object storage.

On severity: given a permanently wedged clustering compaction plus permanently pinned input segments, reachable through two ordinary triggers (an input segment dropped before scheduling, or `numClusters < ClusteringCompactionMinCentroidsNum`), Medium reads low to me. That is worth reflecting wherever this gets tracked.

**My position, stated plainly and unchanged:** this belongs in a separate change, not here.

This PR is +2/-1 in `garbage_collector.go` plus a regression test asserting one thing — that a finished task at `Version=2` removes exactly versions 0 and 1 under its own task prefix. The finding is about something categorically different: whether an analyze task can *reach* a recyclable state at all. Fixing it means changing persistence timing at `:174`/`:205`, which moves the clustering-compaction state machine — the very consequence chain you just extended. That needs its own tests, its own issue, and its own cherry-pick risk assessment against 2.5/2.6/3.0. Nothing in this round argues for coupling the two; if anything, the longer consequence chain argues harder for keeping them apart.

**A request on process.** Both rounds have re-reported this same pre-existing defect against identical code — `49bc433` has not changed between them — and both rounds concluded the fix in this PR is itself sound. A third pass over unchanged code will reach the same place.

So: if you agree that (a) this PR's change is correct and mergeable as it stands, and (b) the `task_analyze.go` defect is out of scope for it, please say so explicitly — a plain "no issues" verdict on this PR is fine. If you *don't* agree and think it must be fixed here before merge, say that instead and I'll take it as a blocker and act on it. What I'd like to avoid is a third identical report that leaves the scope question unresolved.

czs007:
## Re-review of `49bc433ea267`

- **`internal/datacoord/task_analyze.go:174` — terminal path uses `SetState` instead of `UpdateStateWithMeta`, so the terminal state exists only in the in-memory copy and meta stays at `Init`: Confirmed.** Both terminal branches still call `SetState`, which mutates only the scheduler-owned clone, and the sole persistence path (`UpdateStateWithMeta` → `analyzeMeta.UpdateState`) is never reached.
- **`internal/datacoord/task_analyze.go:174` — the in-memory-only terminal transition stalls clustering compaction and causes the analyze task to be skipped by GC: Confirmed.** The scheduler drops the task on Finished/Failed without persisting anything, `processAnalyzing` keeps reading `Init` and makes no transition, and `CheckCleanAnalyzeTask` only allows recycling for a persisted `Finished` state.

---

Two medium-severity persistence issues remain in the analyze terminal-state paths; the changed diff itself is correct and adequately tested.

### Medium

**`internal/datacoord/task_analyze.go:174` — terminal branches in `CreateTaskOnWorker` do not persist state, stalling clustering compaction**

Both terminal branches — `:174` (nil segment → `Failed`) and `:205` (`numClusters < MinCentroidsNum` → `Finished`) — call `SetState` (`task_analyze.go:88-91`), which only mutates `at.State`/`at.FailReason` on a `proto.Clone`d, scheduler-private copy (`analyze_inspector.go:59-62`, `compaction_task_clustering.go:762`). The only persistence path is `UpdateStateWithMeta` (`task_analyze.go:93-99`) → `analyzeMeta.UpdateState` → `saveTask` (`analyze_meta.go:146-162`). The scheduler's terminal branch (`task/global_scheduler.go:301-310`) only calls `backoffs.Remove` for `None/Finished/Failed` — no re-queue, no persistence — so the task leaves scheduling with meta still at `Init`.

Failure: an input segment becomes non-healthy before dispatch (dropped, or absorbed by another compaction) so `CreateTaskOnWorker` finds `nil` in `segmentsMap`, or the raw data size makes `numClusters < ClusteringCompactionMinCentroidsNum` → the persisted state stays `JobStateInit`, the task is dropped from both pending and running sets with no retry, `processAnalyzing` (`compaction_task_clustering.go:616-639`) falls through to `default:` and returns `nil` every round, and the clustering compaction task is stuck in `analyzing` forever; its input segments keep `isCompacting = true` because `doClean`/`resetSegmentCompacting` is never reached, excluding them from all subsequent compactions. Only a datacoord restart re-enqueues the task via `reloadFromMeta`, which then repeats the same memory-only transition. There is no timeout fallback for `analyzing` in `retryableProcess`/`Process` (`compaction_task_clustering.go:261-330`); the timeout is driven solely by worker results in `QueryTaskOnWorker`.

Suggestion: use `UpdateStateWithMeta` in both terminal branches and handle the persistence failure explicitly. *(raised by chyezh, xiaocai2333)*

---

**`internal/datacoord/task_analyze.go:205` — the `numClusters`-too-small shortcut leaves state unpersisted, so this PR's stale-version cleanup never runs for such tasks and their files leak**

Following from the above: on the `task_analyze.go:200-206` shortcut, the persisted state stays `Init`. `CheckCleanAnalyzeTask` (`analyze_meta.go:190-200`) returns `canRecycle=true` only for `JobState_JobStateFinished`, so here it returns `(false, t)` and `recycleUnusedAnalyzeFiles` (`garbage_collector.go:1624-1630`) hits `continue` — the stale-version cleanup loop added by this PR (`garbage_collector.go:1648-1661`) is unreachable for that task. Because the same task also pins the clustering compaction in `analyzing`, `CleanPartitionStatsInfo` → `DropAnalyzeTask` (`meta.go:3148`) is never called either, so the `task == nil` whole-prefix fallback (`garbage_collector.go:1631-1643`) never fires.

Failure: analyze task `401` is dispatched once (Version 0→1), the worker writes `analyze_stats/401/1/` (centroids, offset_mapping) and then the process dies, so `QueryTaskOnWorker` resets the task to `Init`; on retry the input segments have shrunk and `numClusters < ClusteringCompactionMinCentroidsNum`, taking the memory-only `Finished` shortcut → `analyze_stats/401/1/` is never recycled by either the new stale-version loop or the whole-prefix fallback, and such orphaned directories accumulate in object storage indefinitely.

Two scoping notes: the cleanup logic added by this PR is *not* dead in general — the normal completion path (`setJobInfo` → `FinishTask`, `analyze_meta.go:164-181`) persists `Finished` and the loop works as intended; only this shortcut path is affected. And switching to `UpdateStateWithMeta` fixes only the `Finished` shortcut: even with correct persistence, the `:174` `Failed` path still yields `canRecycle=false` from `CheckCleanAnalyzeTask`, so recycling for failed analyze tasks needs a separate design. This is a pre-existing defect outside this PR's diff, but it directly gates the behavior of the code this PR changes. *(raised by chyezh, xiaocai2333)*

---

### Verified (changed diff)

- `internal/datacoord/garbage_collector.go:1648-1656` — the loop removes only versions strictly below `task.Version`, so the current result version is retained.
- `internal/datacoord/garbage_collector.go:1655` — the new `analyze_stats/{taskID}/{i}/` prefix matches the real on-disk layout produced by `KmeansClustering.h:70-83` (`build_id`/`index_version` as the first two levels), and the trailing `/` prevents `401/0/` from matching sibling directories like `401/10/`.
- `internal/datacoord/task_analyze.go:142,156` — `UpdateVersion` runs before the worker request is built from the pre-increment snapshot, so the persisted result version equals `meta.Version`; the `i = 0..Version-1` range is off-by-one free.
- `internal/datacoord/garbage_collector.go:1656-1661` — a failure removing one stale version does not abort cleanup of the remaining ones.
- `internal/datacoord/garbage_collector_test.go:3438-3495` — the regression test distinguishes task ID from version and fails against the previous implementation, which recorded `analyze_stats/2/` twice instead of `analyze_stats/401/0/` and `analyze_stats/401/1/`.

bigsheeper:
Thanks — both findings confirmed again, and I've changed course from my earlier reply: instead of deferring them, they're fixed in this PR (`eafa4cac34`, which also rebases onto latest master).

**Change.** Both terminal branches in `CreateTaskOnWorker` now use `UpdateStateWithMeta`. On a persistence failure the in-memory state is deliberately left untouched, so the scheduler's `Init` branch re-enqueues the task rather than dropping it on a state that exists nowhere.

**Verified.** `./internal/datacoord/...` is green, and the three new/extended assertions — persisted `Failed` on the nil-segment path, persisted `Finished` on the too-few-clusters path, and meta unchanged + task retriable when the save fails — all fail against the previous `SetState` implementation. I also fixed a state leak between the tests in that suite, which shares one `meta` across cases.

**Still open, as you flagged.** `CheckCleanAnalyzeTask` allows recycling only for `Finished`, so the files of a `Failed` analyze task are still not recycled. That needs its own design and I'm not doing it here.

One scoping note on your report: `SetState(JobStateInit, ...)` on the dim-extraction path is unaffected — `Init` matches the persisted state and the scheduler re-enqueues from it.

czs007:

**Verified:**

- `internal/datacoord/garbage_collector.go:1597-1656` — the removal prefix now includes both the parsed task ID and each stale version (`{taskID}/{i}/`), matching the on-disk `analyze_stats/{taskID}/{version}/` layout, and the trailing separators prevent prefix collisions.
- `internal/datacoord/garbage_collector.go:1649` — the loop bound `i < task.Version` removes only `[0, Version)` and preserves the current version's centroids/offset_mapping; version arithmetic in `UpdateVersion` confirms meta `Version` equals the on-disk current, so there is no off-by-one.
- `internal/datacoord/task_analyze.go:171-176, 200-207` — both converted branches persist their terminal state (`Failed`, `Finished`) via `UpdateStateWithMeta` before returning, and both are reached before the `tryDropTaskOnWorker` defer, so early return does not spuriously drop the task on the worker.
- `internal/datacoord/task/global_scheduler.go:281-310` — a failed catalog write leaves the task at `Init`, which the scheduler re-enqueues with backoff, while successfully persisted terminal states are removed from scheduling; no unpersisted terminal state is ever exposed.
- `internal/datacoord/analyze_meta.go:140-155` — `UpdateState` holds the metadata write lock, clones the latest task, and persists before replacing the in-memory entry, so there are no partial in-memory transitions or stale-field overwrites; the `analyzeMeta` write lock does not conflict with the GC path's read lock.
- `internal/datacoord/garbage_collector_test.go:3438-3495` and `internal/datacoord/task_analyze_test.go:212-317` — the regression tests assert the concrete task-scoped prefixes for versions 0 and 1 with version 2 retained, and cover persisted `Failed`, persisted `Finished`, and a catalog-save failure leaving both scheduler and metadata state at `Init`; both genuinely fail pre-fix.

czs007 (approved):
/lgtm

Issue #51237, opened 2026-07-10
[Bug]: geometry_field is not null fails on growing segments after RTREE index is loaded


zhuwenxing:
## Environment
- Milvus version: `master-20260709-b542464f30`
- Milvus image: `harbor.milvus.io/milvusdb/milvus:master-20260709-b542464f`
- Deployment mode: cluster
- MQ type: woodpecker
- SDK version: `pymilvus 3.1.0rc63`
- OS: Linux `4.18.0-372.13.1.el8_6.x86_64`
- Python: `3.12.13`
- CPU/Memory: 16 CU test deployment; querynode and datanode each `4` replicas, `2 CPU / 25Gi` request and `8 CPU / 32Gi` limit
- GPU: N/A
- Others: Kubernetes deployment with Milvus Operator, in-cluster etcd and MinIO

### K8s Pod List
```text
milvus-stable-16cu-24h-g72fm-1417291493                         2/2     Running     0               152m
milvus-stable-16cu-24h-g72fm-2307405793                         2/2     Running     0               152m
milvus-stable-16cu-24h-g72fm-325849973                          2/2     Running     0               152m
milvus-stable-16cu-24h-g72fm-3615251855                         0/2     Completed   0               155m
milvus-stable-16cu-24h-g72fm-46241993                           2/2     Running     0               152m
stable-16cu-workload-etcd-0                                     1/1     Running     0               155m
stable-16cu-workload-etcd-1                                     1/1     Running     0               155m
stable-16cu-workload-etcd-2                                     1/1     Running     0               155m
stable-16cu-workload-milvus-datanode-76ff5464f6-hxvxl           1/1     Running     0               154m
stable-16cu-workload-milvus-datanode-76ff5464f6-kdv6h           1/1     Running     0               154m
stable-16cu-workload-milvus-datanode-76ff5464f6-kqf77           1/1     Running     0               154m
stable-16cu-workload-milvus-datanode-76ff5464f6-zpf8m           1/1     Running     0               154m
stable-16cu-workload-milvus-mixcoord-57cc9c65df-jxzv6           1/1     Running     0               154m
stable-16cu-workload-milvus-proxy-d85dd57f9-mx7hn               1/1     Running     0               154m
stable-16cu-workload-milvus-querynode-0-77899cd587-5xd4v        1/1     Running     0               154m
stable-16cu-workload-milvus-querynode-0-77899cd587-82rpj        1/1     Running     0               154m
[...]

PR #50951, opened 2026-06-30
fix: harden geometry RTree index and cache against memory-safety issues

xiaocai2333:
issue: #50885

## What & why

StreamingNode crashed with a native SIGSEGV — heap corruption that surfaced
in folly's thread-local destructor (`SingletonThreadLocal::LocalLifetime::destroy`,
tracking maps filled with `0xff`) — during chaos testing on collections with
GEOMETRY fields + RTREE index, under concurrent growing-segment incremental
indexing, sealed-segment load, search, and segment drop. This PR hardens the
geometry RTree index and geometry cache against the memory-safety defects found
in that subsystem.

> Honest caveat: the crashing environment was torn down and the available
> coredump only held the secondary-victim folly TLS frames, so none of these is
> *proven* to be the single first domino. They are all genuine,
> independently-reproducible defects in the implicated code, fixed defensively.
> Each is covered by a unit test run under AddressSanitizer.

## Behavior change (beyond memory-safety hardening)

Rows whose WKB fails to parse (and rows with no computable envelope) used to be
**dropped from the RTree index**; they are now **indexed with a deterministic
placeholder MBR** (`POINT(0 0)`) and therefore counted by `Count()` /
`IsNotNull()`. Rationale: a corrupt row is not NULL, and dropping it permanently
desynchronizes the index row count from the segment row count, which then trips
the growing coarse-bitmap bounds guard on every subsequent geometry query
against that segment. Correctness is preserved by exact refinement (a
placeholder row is always refined out unless it truly matches); the tradeoff is
that queries whose bbox covers the origin pay refinement cost for placeholder
rows. This is what the `Count()==2` → `Count()==3` assertion change in
`RTreeIndexTest.cpp` reflects.

Related: for `ST_Equals` with an empty query geometry (no envelope), the index
now falls back to the full candidate set (an intentional full scan) instead of
returning zero candidates, so `ST_Equals(field, 'POLYGON EMPTY')` matches empty
field geometries exactly like the un-indexed data path does. Other operators
keep the no-candidate fast path.

## Fixes

- **Geometry value semantics**: `Geometry::operator=` shallow-copied the raw
  `GEOSGeometry*`, so two instances owned one pointer and the second destructor
  double-freed it. Now releases the existing geometry then deep-clones, and adds
  a noexcept move ctor/assignment. Copying drives GEOS through the *source*
  instance's context and the copy keeps that context, so copying a cache-owned
  Geometry from a query thread is documented as forbidden (implicit copies
[...]

PR #51523, opened 2026-07-17
fix: size the scalar-index valid bitmap by active count on growing segments

xiaocai2333:
issue: #51237

## Root cause

A growing segment carries a segment-level interim R-Tree index for a geometry field as soon as the collection has an `RTREE` index on it (`SegmentGrowingImpl::HasIndex`/`PinIndex`). That makes geometry the only field type whose expressions take the **ScalarIndex** path on a growing segment.

On that path, `ProcessIndexChunksForValid()` sliced the segment-global `IsNotNull()` bitmap with `size_per_chunk_` as the upper bound. On a sealed segment `size_per_chunk()` equals the row count, but on a growing segment it is the raw-data chunk granularity (`segcore.chunkRows`, 128 by default) and is unrelated to the index bitmap. A `geometry_field is not null` query over more than one chunk of growing rows therefore returned a truncated bitmap and crashed `FilterBitsNode`:

```
Assert "col_vec_size == need_process_rows_" => bitset size: 128, need_process_rows_: 5000
```

## Fix

- `ProcessIndexChunksForValid()`: slice the cached valid bitmap by the rows visible to the query (`active_count_`) instead of `size_per_chunk_`, and assert the bitmap covers the requested range. The invariant holds because `SegmentGrowingImpl::Insert` appends to the interim index **before** acking the rows into the active count, and `HasIndex` is gated on the index being synced. For sealed segments `active_count_ == size_per_chunk_ == row count`, so behavior is unchanged.
- `MoveCursorForIndex()`: drop the sealed-only assert. The index cursor is a global row position for growing segments with a segment-level index too; a conjunction whose cheaper side is all-false short-circuits the geometry expression via `MoveCursor()` and would otherwise trip this assert.

## Verification

Regression test `TestGeometryNullExprOnGrowingWithRTreeIndex` reproduces the exact crash before the fix (`bitset size: 128, need_process_rows_: 1000`) and covers:

- `is not null` / `is null` on the execute-all-at-once path (the reported crash)
- a conjunction with a raw-data expression that runs the batched loop across multiple growing chunks
- a conjunction whose numeric side is all-false, exercising the `MoveCursor()` short-circuit on the growing index cursor

Expression regression suite (`*Expr*:*Null*:*GIS*:*RTree*:*Geometry*`, 4984 tests) passes with 0 failures.

Note: `ProcessIndexChunks()` (the result-bitmap path) has the same `size_per_chunk_` bound in its batched branch, but no expression currently reaches it on a growing segment (GIS spatial predicates have their own growing-aware branch in `EvalForIndexSegment`); left untouched to keep this change minimal.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -1947,8 +1947,19 @@ class SegmentExpr : public Expr {
                 *cached_index_chunk_valid_res_, elem_start, elem_count);
 
             current_index_chunk_pos_ = data_pos + batch_rows;
-        } else if (execute_all_at_once_) {
-            // Fast path: move cached bitmap directly, no copy
+        } else if (execute_all_at_once_ &&
+                   int64_t(cached_index_chunk_res_->size()) == active_count_ &&
+                   int64_t(cached_index_chunk_valid_res_->size()) ==
+                       active_count_) {
+            // Fast path: the cached bitmap lines up exactly with the rows
+            // this query emits, so move it out with no copy. The size guard
+            // keeps this branch under the same invariant as the slicing
+            // branch below: on a growing segment the interim index bitmap
+            // may run ahead of active_count_ under concurrent inserts (see
+            // issue #51237), and moving an oversized bitmap wholesale would
+            // hand downstream more rows than the batch. That case falls
+            // through to the slicing branch, which bounds by active_count_
+            // and asserts coverage.
             current_index_chunk_pos_ += cached_index_chunk_res_->size();
             return std::make_shared<ColumnVector>(
                 std::move(*cached_index_chunk_res_),
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -1954,11 +1954,43 @@ class SegmentExpr : public Expr {
                 std::move(*cached_index_chunk_res_),
                 std::move(*cached_index_chunk_valid_res_));
         } else {
-            // Normal index or row-level result: batch by rows directly
+            // Normal index or row-level result: batch by rows directly.
+            //
+            // Slice by active_count_ for the same reason as
+            // ProcessIndexChunksForValid(): the cached bitmap is
+            // segment-global (a scalar index always has exactly one chunk),
+            // while size_per_chunk_ is the raw-data chunk granularity
+            // (segcore.chunkRows) and is unrelated to it. On a sealed segment
+            // the two agree -- size_per_chunk() == get_row_count() -- which is
+            // why bounding by size_per_chunk_ has not broken yet; today only
+            // sealed segments reach here, because on growing segments
+            // HasIndex() is true only for vector/geometry fields and real
+            // geometry predicates take the dedicated branch in
+            // GISFunctionFilterExpr::EvalForIndexSegment(). The moment a
+            // scalar field gains an interim index on growing, or geometry is
+            // routed through ProcessIndexChunks, size_per_chunk_ would
+            // over-run the bitmap exactly as in issue #51237.
+            //
+            // The old third min term was also subtly wrong on its own: it
+            // clamped against the bitmap's FULL length rather than the length
+            // remaining after data_pos, so a bitmap shorter than the row count
+            // combined with data_pos > 0 made append() read past its end.
+            // Assert coverage instead, mirroring the sibling function.
             auto data_pos = current_index_chunk_pos_;
-            auto size =
-                std::min(std::min(size_per_chunk_ - data_pos, batch_size_),
-                         int64_t(cached_index_chunk_res_->size()));
+            auto size = std::min(active_count_ - data_pos, batch_size_);
+            AssertInfo(
+                int64_t(cached_index_chunk_res_->size()) >= data_pos + size,
+                "index bitmap covers {} rows, batch needs rows [{}, {})",
+                cached_index_chunk_res_->size(),
+                data_pos,
+                data_pos + size);
+            AssertInfo(
+                int64_t(cached_index_chunk_valid_res_->size()) >=
+                    data_pos + size,
+                "index valid bitmap covers {} rows, batch needs rows [{}, {})",
+                cached_index_chunk_valid_res_->size(),
+                data_pos,
+                data_pos + size);
 
             result.append(*cached_index_chunk_res_, data_pos, size);
             valid_result.append(*cached_index_chunk_valid_res_, data_pos, size);
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -403,8 +403,9 @@ class SegmentExpr : public Expr {
 
     void
     MoveCursorForIndex() {
-        AssertInfo(segment_->type() == SegmentType::Sealed,
-                   "index mode only for sealed segment");
+        // The index cursor is a global row position. This holds for sealed
+        // segments and for growing segments with a segment-level scalar
+        // index (the geometry interim R-Tree, see issue #51237).
         auto size =
             std::min(active_count_ - current_index_chunk_pos_, batch_size_);
 
@@ -2205,13 +2206,28 @@ class SegmentExpr : public Expr {
             cached_index_chunk_id_ = 0;
         }
 
-        // Process current batch
+        // Process current batch.
+        //
+        // The cached bitmap is segment-global (a scalar index always has
+        // exactly one chunk), so slice it by the rows visible to this query
+        // (active_count_). Bounding by size_per_chunk_ would be wrong on a
+        // growing segment, where size_per_chunk_ is the raw-data chunk
+        // granularity (segcore.chunkRows) and unrelated to the index bitmap.
+        // A growing segment reaches this path through the geometry interim
+        // R-Tree index; its bitmap may also run ahead of active_count_ under
+        // concurrent inserts, so active_count_ -- not the bitmap size --
+        // decides how many rows to emit. See issue #51237.
         TargetBitmap valid_result;
         valid_result.set();
 
         auto data_pos = current_index_chunk_pos_;
-        auto size = std::min(std::min(size_per_chunk_ - data_pos, batch_size_),
-                             int64_t(cached_index_chunk_valid_res_->size()));
+        auto size = std::min(active_count_ - data_pos, batch_size_);
+        AssertInfo(
+            int64_t(cached_index_chunk_valid_res_->size()) >= data_pos + size,
+            "index valid bitmap covers {} rows, batch needs rows [{}, {})",
+            cached_index_chunk_valid_res_->size(),
+            data_pos,
+            data_pos + size);
 
         valid_result.append(*cached_index_chunk_valid_res_, data_pos, size);
 
--- a/internal/core/src/exec/expression/ExprGISMiscTest.cpp
+++ b/internal/core/src/exec/expression/ExprGISMiscTest.cpp
@@ -30,6 +30,7 @@
 #include "ExprTestBase.h"
 #include "NamedType/named_type_impl.hpp"
 #include "bitset/bitset.h"
+#include "common/Common.h"
 #include "common/Consts.h"
 #include "common/Exception.h"
 #include "common/Geometry.h"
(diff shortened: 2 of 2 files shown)


czs007:
Adversarial review surfaced one non-blocking test-coverage gap.

## Nitpicks

**`internal/core/src/exec/expression/ExprGISMiscTest.cpp:1043`** — Conjunction case never exercises multi-batch index-cursor advancement

No case in `TestGeometryNullExprOnGrowingWithRTreeIndex` (starting at line 969) runs more than one batch, so cross-batch index-cursor advancement — the exact path the `size_per_chunk_` bug corrupted — is never covered.

Cases 1 and 2 (single `NullExpr`) take the all-at-once path: `FilterBitsNode.cpp:176-178` calls `SetExecuteAllAtOnce()`, which sets `batch_size_ = active_count_` (`Expr.h:2384`), collapsing execution into a single batch by construction. Cases 3 and 4 (the conjunctions) do enter the batched loop, but `batch_size_` resolves to the 8192 default: `ExecuteQueryExpr` builds a `QueryContext` with a default `QueryConfig`, so `get_expr_batch_size()` falls back to `EXEC_EVAL_EXPR_BATCH_SIZE.load()` (`QueryContext.h:138-141`) == `DEFAULT_EXEC_EVAL_EXPR_BATCH_SIZE` == 8192 (`Consts.h:90`), and the `ExprTest` fixture (`ExprTestBase.h`) never overrides it. With `N = 1000` (line 1003), 1000 < 8192 yields a single iteration and `current_index_chunk_pos_` never advances past 0. As a result, neither the new `size = std::min(active_count_ - data_pos, batch_size_)` slicing at a nonzero `data_pos` nor the `MoveCursorForIndex()` assertion relaxation that case 4 targets is covered at any cursor position other than 0.

Suggestion: shrink the batch size for this test rather than inflating `N`, reusing the save/restore guard pattern already present in this codebase at `MatchExprTest.cpp:441-444` (`EXEC_EVAL_EXPR_BATCH_SIZE.store(...)` with a scope guard restoring the prior value). A batch size of, say, 128 with `N = 1000` yields roughly 8 batches and directly pins cross-batch cursor correctness. Non-blocking — case 1 already reproduces the original crash, so the regression itself is pinned; this only concerns the incremental cursor path.

(raised by wzy)

xiaocai2333:
Addressed the adversarial-review coverage gap (raised by wzy): the previous test cases all collapsed into a single batch — the all-at-once `NullExpr` cases by construction (`SetExecuteAllAtOnce` sets `batch_size_ = active_count_`), and the conjunction cases because `batch_size_` fell back to the 8192 default while `N = 1000`. So `current_index_chunk_pos_` never advanced past 0 and the non-zero `data_pos` slicing / `MoveCursorForIndex` relaxation were never exercised.

Fixed by pinning `EXEC_EVAL_EXPR_BATCH_SIZE = 128` for the test (scope-guarded restore, same pattern as `MatchExprTest.cpp:441-444`), so the conjunction cases now iterate ~8 batches and advance the cursor across chunk boundaries.

Verified the added coverage actually catches the bug: with the batch size pinned to 128, temporarily reverting the `active_count_` slicing back to the `size_per_chunk_` bound makes the multi-batch cases fail with the exact original assert (`bitset size: 128, need_process_rows_: 1000`); restoring the fix turns them green again.

czs007:

czs007:
Adversarial review surfaced two low-severity findings — one latent fragility in `Expr.h`'s batched index path and one coverage gap in the new regression test — neither of which blocks the fix.

## Low

**`internal/core/src/exec/expression/Expr.h:1957` — batched `else` branch in `ProcessIndexChunksImpl` keeps the `size_per_chunk_` slicing pattern this PR fixes elsewhere**

The batched `else` branch still computes `size = std::min(std::min(size_per_chunk_ - data_pos, batch_size_), cached_index_chunk_res_->size())` — exactly the construction this PR identifies as wrong for growing segments in `ProcessIndexChunksForValid()` (the crash shape from issue #51237). It is not an active bug today: on growing segments `HasIndex` only returns true for vector/geometry fields (`SegmentGrowingImpl.h:530-533`), non-geometry scalar expressions fall through to raw data, and real geometry spatial predicates take the dedicated growing branch in `GISFunctionFilterExpr::EvalForIndexSegment()` (`GISFunctionFilterExpr.cpp:601-624`). So this branch is only reached on sealed segments, where `size_per_chunk() == get_row_count()` (`ChunkedSegmentSealedImpl.cpp:3148-3150`). It is fragile, though: the moment any scalar field gains an interim index on growing, or geometry is routed through `ProcessIndexChunks`, it will crash in `FilterBitsNode` the same way. A separate minor hazard is worth folding in: the third `min` term uses the **full bitmap length** `cached_index_chunk_res_->size()` rather than `size() - data_pos`, so when the bitmap is shorter than the row count and `data_pos > 0`, `result.append(*cached_index_chunk_res_, data_pos, size)` reads out of bounds — precisely the class of problem the PR guards with `AssertInfo` in the sibling function (`Expr.h:2222-2227`).

*Suggestion:* slice against `active_count_` and add the same coverage assertion as the sibling function, or at minimum add a comment documenting that this branch assumes a sealed segment (single chunk == row count). *(raised by tinswzy)*

---

**`internal/core/src/exec/expression/ExprGISMiscTest.cpp:970` — regression test covers only `NullExpr`; the growing + RTREE spatial-predicate path has zero coverage**

All four sub-blocks of the new `TestGeometryNullExprOnGrowingWithRTreeIndex` construct only `expr::NullExpr` (IsNotNull / IsNull / two `And` combinations), which matches the `geometry_field is not null` report in issue #51237 — the fix direction is correct. However, spatial predicates (`ST_Contains`, `ST_Intersects`, etc.) on a multi-chunk growing segment take the separate growing branch in `GISFunctionFilterExpr::EvalForIndexSegment()` (`GISFunctionFilterExpr.cpp:601-624`), which this PR neither changes nor tests. Additionally, every pre-existing growing GIS case in this file builds its segment with `empty_index_meta`, so `HasIndex` is false and they all take the raw-data path — meaning "geometry index on a growing segment" remains covered by the `NullExpr` path alone, and the index branch for spatial predicates has zero coverage in the C++ unit tests.

*Suggestion:* add a follow-up test exercising a spatial predicate on a growing segment with an RTREE index. (On inspection that branch appears correct today — its slicing is bounded by `real_batch_size` and no overread was found — so this can reasonably be tracked as follow-up rather than blocking.) *(raised by tinswzy)*

czs007:
Adversarial review surfaced two non-blocking, follow-up-track invariant/concurrency concerns in the index-result execution path that are worth tracking separately from this PR.

### Medium

- **`internal/core/src/exec/expression/Expr.h:2238` — Validity bitmap coverage invariant is not synchronized with concurrent inserts (data race).** `ProcessIndexChunksForValid()` assumes `index_ptr->IsNotNull()` returns a consistent bitmap covering `active_count_` and enforces this with `AssertInfo` (Expr.h:2258), but insert-before-ack only guarantees acked rows are appended before `active_count_` advances — it does not synchronize the bitmap snapshot with the next concurrent insert. `RTreeIndex::IsNotNull()` (RTreeIndex.cpp:447) calls `Count()` *before* taking `mutex_` (:450), and `Count()` reads `wrapper_->count()` and `null_offset_.size()` lock-free (RTreeIndex.h:82-88); `RTreeIndexWrapper::count()` reads `rtree_.size()` without `rtree_mutex_` (RTreeIndexWrapper.cpp:299-302) while `add_geometry()` mutates the tree under that lock (RTreeIndexWrapper.cpp:61, 96). A query racing the next insert is a C++ data race that can yield an inconsistent/invalid count and fail the new coverage assertion (or worse). *Suggestion:* have `Count()`/`IsNotNull()` take a consistent synchronized snapshot across null and non-null rows (or derive growing-segment validity from the raw validity column), and add a query-vs-insert concurrency regression test. This race predates and is orthogonal to this PR, so it should not block merge — track it as an independent follow-up. *(raised by chyezh)*

### Low

- **`internal/core/src/exec/expression/Expr.h:1951` — `execute_all_at_once_` result branch lacks symmetric hardening.** This branch moves the entire cached bitmap and advances the cursor by `cached_index_chunk_res_->size()` with no `active_count_` clamp or assertion, unlike the adjacent normal-index `else` branch (Expr.h:1980-1994), which this PR already updated to slice by `active_count_` with an `AssertInfo` coverage check. It is unreachable on growing today (geometry results take the dedicated `GISFunctionFilterExpr::EvalForIndexSegment` path; only sealed segments reach here, consistent with the author's note at Expr.h:1966-1973), so this is not an active bug — but it is the same latent shape as #51237 and would break if a growing scalar-index field ever routed index results here. *Suggestion:* add a `bitmap.size() == active_count_` assertion (or clamp) so all three branches share the same invariant; can be handled as a follow-up. *(raised by tinswzy)*

xiaocai2333:
Addressed both findings from the latest adversarial review round (raised by chyezh / tinswzy), pushed as `9759cba`..`794a2da`:

**Medium — `RTreeIndex` count/null-bitmap reads race concurrent growing inserts** (`6142adefde`)
`mutex_` is now the snapshot lock for the growing write path: `AddGeometry()` holds the unique lock across the whole append (tree or `null_offset_` mutation plus the `total_num_rows_` update), and `Count()`/`IsNull()`/`IsNotNull()` compute the count and scan null offsets inside a single shared-lock scope via a new `CountLocked()` — one scope also avoids re-acquiring the shared lock recursively, which can deadlock against a pending writer under write priority. `RTreeIndexWrapper::count()` now takes `rtree_mutex_` shared instead of reading `rtree_.size()` lock-free. Lock order is `mutex_` → `rtree_mutex_` everywhere; I audited all call sites and no path takes the reverse order.

Regression test (`794a2da6f8`): a writer thread appends 20k rows in offset order with a deterministic null pattern while a reader loops over `Count()`/`IsNull()`/`IsNotNull()`, asserting the count is monotonic and every returned bitmap matches the pattern over its whole prefix — before the fix the count could include a row whose `null_offset_` entry was not yet visible, reporting a null row as valid.

**Low — `Expr.h` `execute_all_at_once_` branch lacks the active-count invariant** (`9759cba0a3`)
The zero-copy move fast path is now guarded: it only fires when both cached bitmaps line up exactly with `active_count_`; otherwise execution falls through to the slicing branch, which bounds by `active_count_` and asserts coverage — so all three branches of `ProcessIndexChunksImpl` now share the same invariant.

Verified locally under ASAN: `RTreeIndexTest` 16/16 (incl. the new concurrency test) and all 232 GIS/Geometry expression tests pass. Honest caveats: the data-race elimination rests on the lock audit plus the semantic regression test (no local TSAN run), and no existing test constructs the bitmap-size-mismatch shape, so the new guard's fallback arm is exercised only via the already-covered slicing branch.

czs007:
Adversarial review surfaced two high-severity concurrency defects in the growing R-Tree index path that can cause incorrect null-filtering results, out-of-bounds writes, and silently dropped rows under concurrent inserts.

## 🔴 High

**`internal/core/src/index/RTreeIndex.cpp:442` — `lower_bound` over `null_offset_` assumes a sorted, all-in-range vector that concurrent inserts violate**
`IsNull()`/`IsNotNull()` call `std::lower_bound(null_offset_.begin(), null_offset_.end(), count)` and mutate every entry before the returned end, which is only correct when `null_offset_` is sorted *and* every entry is `< count`. Both invariants hold only for a single in-order writer — the one case the new unit test exercises. Under the growing segment's concurrent-insert contract (`PreInsert` reserves offsets via atomic `fetch_add`, `Insert` takes only a shared lock, and `ack_responder_` explicitly merges out-of-order ranges), two threads can append offsets interleaved, leaving the vector non-monotonic. On an unsorted range the partition is meaningless: a null offset `< count` can land after `end` and be skipped (IS NOT NULL false-positives an invalid row), or an offset `>= count` can land before `end`, so `bitset.set(offset)` writes past the `count`-sized `TargetBitmap` (heap corruption/crash). The same pattern repeats in `IsNotNull()` at line 457.
*Suggestion:* keep `null_offset_` sorted under `mutex_`, or replace `lower_bound` with a bounds-checked scan that applies only offsets `< count`; add a multi-writer, offset-interleaved regression test, since the current single-writer unit test cannot surface this. *(raised by tinswzy)*

**`internal/core/src/index/RTreeIndex.cpp:615` — first-insert wrapper init runs outside `mutex_` behind a non-atomic check-then-set, racing concurrent first inserts**
The growing R-Tree's wrapper-init path is not serialized against concurrent first inserts, and `InitForBuildIndex` unconditionally reassigns `wrapper_` (line 90). In `AddGeometry`, `if (!wrapper_) InitForBuildIndex(true);` (lines 615–618) runs *before* the `unique_lock` at line 625, so two threads can both observe `wrapper_ == nullptr`, both construct, and race on the `shared_ptr` assignment — a data race in which the loser's added rows go to a wrapper that is overwritten and lost. The production entry `ScalarFieldIndexing::process_geometry_data` (`FieldIndexing.cpp:743-758`) compounds this with `if (!built_) { InitForBuildIndex(true); built_ = true; }`, a non-atomic check-then-set on an atomic `built_` called with no lock under `Insert`'s shared lock; two concurrent first batches can both enter the branch, and the loser's rows are silently dropped, later tripping the coverage assertion or returning missing rows. The `built_` gate guarantees a non-null wrapper only *after* the first batch settles — it does not cover two racing first batches.
*Suggestion:* centralize growing-index init behind `mutex_` or `std::call_once`, make `InitForBuildIndex` idempotent, and remove the racy pre-lock check in `AddGeometry`. *(raised by tinswzy)*

xiaocai2333:
Correction to my earlier comment: the R-Tree concurrency commits it referenced have been **reverted from this PR** (branch force-pushed back to `9759cba0a3`).

While addressing the latest adversarial-review round (the two High findings: unsorted `null_offset_` vs `lower_bound`, and the first-insert wrapper-init race), I found that **#50951 already fixes all of them** — the bounds-checked null scan is the identical fix, the growing init is idempotent under the write lock there, and it goes further (locked `RTreeIndexWrapper::count()`/`ByteSize()`, torn-`shared_ptr` publication guards, row-space `Count()` semantics, plus a full set of `GrowingConcurrent*` multi-writer regression tests). Carrying a second, divergent copy of the same fixes here would guarantee conflicts and double review.

So this PR is scoped back to issue #51237 only:

- `caa5a2f5e5` valid-bitmap sizing by active count (the original fix)
- `06899f6f01` batched index-result slicing by active count
- `0f0618d3a5` growing + RTREE spatial-predicate coverage
- `9759cba0a3` all-at-once fast-path guarded by the same active-count invariant (this round's Low)

All R-Tree concurrency findings from the review rounds (the earlier Medium on `Count()`/`IsNotNull()` racing inserts, and both Highs) are addressed by and should be reviewed in **#50951**. Note #50951 also documents that production currently serializes growing-segment inserts per vchannel, so the multi-writer hazards are contract-level (AppendingIndex claims "concurrent, reentrant"), not an active production crash path.

czs007:
Adversarial review surfaced two related issues around the growing-segment GIS cursor: a high-severity correctness bug where `MoveCursor()` no-ops on growing segments, and a medium-severity test gap that leaves that path uncovered (with the proposed repro being vacuous).

## High

**`internal/core/src/exec/expression/GISFunctionFilterExpr.h:78` — `MoveCursor()` no-ops on growing segments, stranding both cursors on short-circuit**
`PhyGISFunctionFilterExpr::MoveCursor()` only delegates to `SegmentExpr::MoveCursor()` when `segment_->type() == SegmentType::Sealed`, so on a growing segment it advances neither the index nor the data cursor. When a conjunction short-circuits (`PhyConjunctFilterExpr::Eval` builds an active-row bitmap and, on `active_rows.none()`, calls `SkipFollowingExprs(i+1)` → `MoveCursor()` rather than `Eval`), the growing GIS expr's `current_index_chunk_pos_` / `current_data_chunk_pos_` stay put. For a query like `int_expr AND ST_Intersects(geo, …)` over a growing indexed segment — where the predicate is all-false on early batches (GIS skipped, cursor stale) but active on a later batch (GIS finally evaluated) — `GetNextBatchSize()` recomputes the batch from the stale `current_index_chunk_pos_`, so `EvalForIndexSegment` re-appends the index bitmap slice from row 0 while the conjunction is at rows `[k*128, (k+1)*128)`, ANDing mismatched ranges: silent false positives/negatives, and the tail can trip the `processed_rows == real_batch_size` asserts. This PR relaxed the base `MoveCursorForIndex()` (dropping its sealed-only assert) precisely so growing works, but the GIS override still short-circuits growing, so the relaxed base is never reached and the bug remains.
*Suggestion:* Drop the sealed-only guard and always delegate to `SegmentExpr::MoveCursor()` (the relaxed base now advances both index and data cursors correctly for growing), or explicitly advance both cursors in the override. *(raised by tinswzy)*

## Medium

**`internal/core/src/exec/expression/ExprGISMiscTest.cpp:1123` — Growing R-Tree spatial test never exercises `MoveCursor`; the High-severity path is uncovered and the suggested repro would pass anyway**
`TestGeometrySpatialExprOnGrowingWithRTreeIndex` runs each spatial case as a standalone `GISFunctionFilterExpr` under `FilterBitsNode`. A standalone geometry-index expr has `exec_path_ == ScalarIndex`, so `CanExecuteAllAtOnce()` returns true and `FilterBitsNode` calls `SetExecuteAllAtOnce()`, which sets `batch_size_ = active_count_` and overrides the `EXEC_EVAL_EXPR_BATCH_SIZE.store(128)` the test installs. All 1000 rows are evaluated in a single `Eval`, `current_index_chunk_pos_` goes 0→1000 in one call, and `MoveCursor()` is never invoked — so the "batched slicing path runs more than once" comment is inaccurate here and the exact High shape (a spatial predicate skipped in a short-circuiting conjunction on a growing segment) has zero regression coverage. Note also that the suggested repro `int_expr(<0, all-false) AND ST_Intersects` would not catch the bug: `DataGen` INT64 values are always `>= 0`, so `< 0` is all-false on every batch, GIS is skipped (via `MoveCursor`) every batch and never evaluated, the result equals the raw-data segment, and the test passes even with the broken override. The `NullExpr` conjunction cases iterate multiple batches but drive `PhyNullExpr`'s `MoveCursor`, not the GIS override.
*Suggestion:* Add a case with a predicate that is false on early batches but active on a later one — e.g. `int64 >= N/2` (int_expr first so it short-circuits) `AND ST_Intersects` — over the growing indexed segment across multiple batches, asserting the results equal the raw-data segment's. That skips GIS early, then evaluates it with a stale cursor and diverges, exercising the override under a genuine batch loop. *(raised by tinswzy)*

czs007:
The issues array is empty — everything was dropped, so per the output rule:

czs007:

liliu-z (commented):
/lgtm
/approve

PR #51526, opened 2026-07-17
fix: [3.0] keep proxy metric status label stable, split cause into its own label (#51495)

czs007:
Cherry-pick from master
pr: https://github.com/milvus-io/milvus/pull/51495
issue: #51493

### What

```
```

|---|---|

Outcomes that have no responsible party emit `cause=""`. Prometheus treats an empty label value as equivalent to the label being absent, so those series stay byte-identical to what pre-2.6.19 emitted — the new label materializes only where it carries information.

Old queries work unchanged and still count each request once: `status="fail"` is again every hard failure, because Prometheus aggregates over `cause` for free — that rollup is what a label dimension is for. `cause` is functionally dependent on the outcome, so the realized `(status, cause)` pairs are the same set of series the split already produces, not the product of both domains.

### Cherry-pick notes

Rebased onto current 3.0. Three files conflicted with upstream's refactor that extracted the response-status switch into `requestutil.GetStatusFromResponse()` (which also handles `AlterStatusGetter`); the resolution keeps upstream's helper and applies the cause classification on top, in `pkg/util/requestutil/getter.go`, `internal/distributed/proxy/request_interceptor.go` and `internal/distributed/proxy/httpserver/handler_v2.go`.

The emit surface was **re-verified on this branch rather than assumed**: all 62 `WithLabelValues` sites for the two metrics were checked for label arity by a paren-balanced scan (61 explicit, 1 variadic in a test), because adding a label makes arity a runtime panic, not a compile error.

`master` currently emits `cause="na"` for the no-cause case; #51907 brings it in line with the empty value used here.

- Green: `internal/proxy`, `internal/distributed/proxy`, `internal/distributed/proxy/httpserver`, `internal/distributed/proxy/client`, `pkg/metrics`, `pkg/util/requestutil` (incl. `TestParseMetricLabelStatusDomainIsStable`, which pins the status domain so a future re-split fails CI), `pkg/util/merr`.
- `make verifiers`: 0 issues.
- Realized `(status, cause)` pairs per node/function/db/collection, enumerated from the emit sites: `total`/`success`/`retry`/`abandon` with no cause, plus `fail` and `rejected` each crossed with `user`/`system`/`cancel` — 10 series, versus 9 for the v2.6.19 split. The `|status| x |cause|` product never materializes because cause is functionally dependent on the outcome.

### DIFF
--- a/deployments/monitor/grafana/milvus-dashboard.json
+++ b/deployments/monitor/grafana/milvus-dashboard.json
@@ -4794,7 +4794,7 @@
                 "uid": "${datasource}"
               },
               "exemplar": true,
-              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=~\"fail_.*\"}[2m])/120) by(function_name, pod, node_id)",
+              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=\"fail\"}[2m])/120) by(function_name, pod, node_id)",
               "interval": "",
               "legendFormat": "{{function_name}}-{{pod}}-{{node_id}}",
               "queryType": "randomWalk",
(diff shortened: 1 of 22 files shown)


liliu-z (commented):
/lgtm
/approve

liliu-z (commented):
/lgtm
/approve

Issue #51191, opened 2026-07-09
[Bug]: copy-segment restore pipeline: stuck Pending job on partial task creation, restore failed by transient query RPC error, UpdateTask mutates cache before persist


xiaocai2333:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: master (observed at commit e203c56f0a, "fix: clean up mlog call formatting (#51027)")
- Deployment mode: cluster
- Component: DataCoord (snapshot restore / copy-segment pipeline)

### Current Behavior

Three robustness defects in the copy-segment restore pipeline (`internal/datacoord/copy_segment_*.go`). We hit them while integrating the snapshot feature downstream; all three reproduce against master.

**1. `checkPendingJob` cannot resume a partially created job — job stuck `Pending` forever (high)**

`copy_segment_checker.go`, `checkPendingJob`:

```go
tasks := c.copyMeta.GetTasksByJobID(c.ctx, job.GetJobId())
if len(tasks) > 0 {
    return
}
```

Task creation is a multi-step sequence: per-group `AllocID` + `AddTask` (each persisted individually), then `UpdateJob -> Executing`. If any step fails mid-way (etcd hiccup, DataCoord restart between steps), the job remains `Pending` with a subset of tasks persisted. On the next check round, the early return above fires because *some* tasks exist, so the job never gets its remaining tasks and never reaches `Executing` — `checkCopyingJob` only handles `Executing` jobs, and the restore hangs until the job timeout with no way to make progress.

**2. A transient `QueryCopySegment` RPC error fails the whole restore (medium)**

`copy_segment_task.go`, `QueryTaskOnWorker`:

```go
resp, err := cluster.QueryCopySegment(nodeID, req)
if err != nil {
    t.markTaskAndJobFailed(fmt.Sprintf("query copy segment RPC failed: %v", err))
    return
}
```

A single network blip or DataNode restart during a status poll permanently fails the task and its parent restore job. This is inconsistent with the import pipeline, where a `QueryImport` RPC error resets the task to `Pending` for retry instead of failing the job.

**3. `copySegmentMeta.UpdateTask` mutates shared state before persisting (medium)**

`copySegmentTask.Clone` copies the `atomic.Pointer` value, not the protobuf:

```go
cloned.task.Store(t.task.Load())
```

[...]

PR #51527, opened 2026-07-17
fix: harden copy-segment restore pipeline against partial failures and transient errors

xiaocai2333:
issue: #51191

Fixes three robustness defects in the copy-segment restore pipeline (`internal/datacoord/copy_segment_*.go`), each with a unit test covering the failure mode.

### 1. `checkPendingJob` cannot resume a partially created job (high)

Task creation is a multi-step sequence (per-group `AllocID` + `AddTask`, each persisted individually, then the `Pending → Executing` transition). If any step fails mid-way (etcd hiccup, DataCoord restart), the job stayed `Pending` with only a subset of tasks persisted, and the old early-return `if len(tasks) > 0 { return }` fired on every subsequent round — the restore hung until the job timeout with no way to progress.

Now the transition is resume-safe: it computes the source segments already covered by persisted tasks, creates tasks only for the uncovered mappings, and always (re-)applies the idempotent `Executing` transition (so a round that persisted all tasks but failed the job update also recovers).

### 2. Transient `QueryCopySegment` RPC error failed the whole restore (medium)

A single network blip or DataNode restart during a status poll permanently failed the task and its parent restore job. This was inconsistent with the import pipeline, where a `QueryImport` RPC error resets the task for retry instead of failing the job.

Now a query RPC error is logged and the task is kept `InProgress` for re-query on the next tick. The job-level timeout still bounds a permanently unreachable node; worker-reported task failures keep the fail-fast behavior.

### 3. `copySegmentTask.Clone` mutated shared state before persisting (medium)

`Clone` copied the `atomic.Pointer` value, not the protobuf, so the "clone" used by `UpdateTask` shared the same `*datapb.CopySegmentTask` as the cached entry. Update actions mutate the proto in place, so the in-memory task was mutated *before* `SaveCopySegmentTask` — a failed catalog save left the coordinator cache reflecting the new state while etcd kept the old one. Now `Clone` deep-copies with `proto.Clone`, matching `copySegmentJob.Clone`.

### Testing

- `TestCheckPendingJob_ResumesPartialTaskCreation`, `TestCheckPendingJob_AllTasksExistTransitionsToExecuting`
- `TestQueryTaskOnWorker_RPCErrorKeepsTaskInProgress`
- `TestClone_DeepCopiesProto`, `TestUpdateTask_SaveFailureLeavesCacheUnchanged`
- Full `./internal/datacoord/` package: PASS

### DIFF
--- a/internal/datacoord/copy_segment_checker.go
+++ b/internal/datacoord/copy_segment_checker.go
@@ -297,6 +297,23 @@ func (c *copySegmentChecker) LogTaskStats() {
 func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 	log := mlog.With(mlog.FieldJobID(job.GetJobId()))
 
+	// Step 0: Re-read the cached job. The `job` argument is a snapshot taken
+	// before this function ran, but tasks of a Pending job can already be
+	// dispatched (the inspector does not filter by job state) and a concurrent
+	// failure (markTaskAndJobFailed) may have moved the job to a terminal
+	// state and released its snapshot pin. Creating more tasks for such a job
+	// would be wasted work at best.
+	current := c.copyMeta.GetJob(c.ctx, job.GetJobId())
+	if current == nil {
+		log.Info(c.ctx, "job no longer exists, skip pending check")
+		return
+	}
+	if current.GetState() != datapb.CopySegmentJobState_CopySegmentJobPending {
+		log.Info(c.ctx, "job is no longer pending, skip pending check",
+			mlog.String("currentState", current.GetState().String()))
+		return
+	}
+
 	// Step 1: Validate job has segment mappings
 	idMappings := job.GetIdMappings()
 	if len(idMappings) == 0 {
@@ -373,11 +390,20 @@ func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 	// Step 5: Update job state to Executing. This also runs when all segments
 	// were already covered (groups is empty), retrying a transition that a
 	// previous round failed to persist.
-	err := c.copyMeta.UpdateJob(c.ctx, job.GetJobId(),
+	// The transition is state-guarded (Pending -> Executing only): a task
+	// dispatched during Step 4 can fail concurrently, and markTaskAndJobFailed
+	// then moves the job to Failed and releases its snapshot pin. An
+	// unconditional update here would resurrect that Failed job as Executing.
+	updated, err := c.copyMeta.UpdateJobInState(c.ctx, job.GetJobId(),
+		datapb.CopySegmentJobState_CopySegmentJobPending,
 		UpdateCopyJobState(datapb.CopySegmentJobState_CopySegmentJobExecuting),
 		UpdateCopyJobTotalSegments(int64(len(idMappings))))
 	if err != nil {
 		log.Warn(c.ctx, "failed to update job state to Executing", mlog.Err(err))
+		return
+	}
+	if !updated {
+		log.Info(c.ctx, "job left Pending state concurrently, skip transition to Executing")
 		return
 	}
 	log.Info(c.ctx, "copy segment job started",
--- a/internal/datacoord/copy_segment_checker.go
+++ b/internal/datacoord/copy_segment_checker.go
@@ -286,19 +286,18 @@ func (c *copySegmentChecker) LogTaskStats() {
 //   - Full segment metadata (binlogs, indexes) is fetched by DataNode when executing
 //   - Keeps task metadata small and efficient to persist
 //
-// Idempotency:
-//   - Safe to call multiple times - only creates tasks on first call
-//   - Subsequent calls return early if tasks already exist
+// Idempotency and crash recovery:
+//   - Task creation is a multi-step sequence (per-group AllocID + AddTask, then
+//     the Executing transition), each step persisted individually. A failure
+//     mid-way (etcd hiccup, DataCoord restart) leaves the job Pending with only
+//     a subset of tasks persisted.
+//   - To be resume-safe, each round creates tasks only for source segments not
+//     yet covered by persisted tasks, then (re-)applies the idempotent
+//     Pending → Executing transition.
 func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 	log := mlog.With(mlog.FieldJobID(job.GetJobId()))
 
-	// Step 1: Check if tasks already created (idempotent operation)
-	tasks := c.copyMeta.GetTasksByJobID(c.ctx, job.GetJobId())
-	if len(tasks) > 0 {
-		return
-	}
-
-	// Step 2: Validate job has segment mappings
+	// Step 1: Validate job has segment mappings
 	idMappings := job.GetIdMappings()
 	if len(idMappings) == 0 {
 		log.Warn(c.ctx, "no id mappings to copy, mark job as completed")
@@ -310,9 +309,23 @@ func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 		return
 	}
 
-	// Step 3: Split mappings into groups (max segments per task)
+	// Step 2: Compute source segments already covered by persisted tasks,
+	// so a partially created job resumes instead of duplicating tasks.
+	tasks := c.copyMeta.GetTasksByJobID(c.ctx, job.GetJobId())
+	coveredSourceIDs := make(map[int64]struct{})
+	for _, task := range tasks {
+		for _, mapping := range task.GetIdMappings() {
+			coveredSourceIDs[mapping.GetSourceSegmentId()] = struct{}{}
+		}
+	}
+	pendingMappings := lo.Filter(idMappings, func(mapping *datapb.CopySegmentIDMapping, _ int) bool {
+		_, covered := coveredSourceIDs[mapping.GetSourceSegmentId()]
+		return !covered
+	})
+
+	// Step 3: Split uncovered mappings into groups (max segments per task)
 	maxSegmentsPerTask := Params.DataCoordCfg.MaxSegmentsPerCopyTask.GetAsInt()
-	groups := lo.Chunk(idMappings, maxSegmentsPerTask)
+	groups := lo.Chunk(pendingMappings, maxSegmentsPerTask)
 
 	// Step 4: Create task for each group
 	for i, group := range groups {
@@ -357,7 +370,9 @@ func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 			mlog.Int("segmentCount", len(group)))
 	}
 
-	// Step 5: Update job state to Executing
+	// Step 5: Update job state to Executing. This also runs when all segments
+	// were already covered (groups is empty), retrying a transition that a
+	// previous round failed to persist.
 	err := c.copyMeta.UpdateJob(c.ctx, job.GetJobId(),
 		UpdateCopyJobState(datapb.CopySegmentJobState_CopySegmentJobExecuting),
 		UpdateCopyJobProgress(0, int64(len(idMappings))))
@@ -366,7 +381,8 @@ func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 		return
 	}
 	log.Info(c.ctx, "copy segment job started",
-		mlog.Int("taskCount", len(groups)),
+		mlog.Int("newTaskCount", len(groups)),
+		mlog.Int("resumedTaskCount", len(tasks)),
 		mlog.Int("totalSegments", len(idMappings)))
 }
 
--- a/internal/datacoord/copy_segment_checker.go
+++ b/internal/datacoord/copy_segment_checker.go
@@ -375,7 +375,7 @@ func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 	// previous round failed to persist.
 	err := c.copyMeta.UpdateJob(c.ctx, job.GetJobId(),
 		UpdateCopyJobState(datapb.CopySegmentJobState_CopySegmentJobExecuting),
-		UpdateCopyJobProgress(0, int64(len(idMappings))))
+		UpdateCopyJobTotalSegments(int64(len(idMappings))))
 	if err != nil {
 		log.Warn(c.ctx, "failed to update job state to Executing", mlog.Err(err))
 		return
--- a/internal/datacoord/copy_segment_task.go
+++ b/internal/datacoord/copy_segment_task.go
@@ -22,6 +22,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/cockroachdb/errors"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
@@ -377,6 +378,26 @@ func (t *copySegmentTask) markTaskAndJobFailed(reason string) {
 		WrapCopySegmentTaskLog(t, mlog.String("reason", reason))...)
 }
 
+// isCopyTaskLostOnWorker reports whether a QueryCopySegment error means the
+// worker-side task is confirmed lost, as opposed to a transient transport error.
+//
+// Confirmed-loss signals (audited against every construction site on the
+// QueryCopySegment path):
+//   - merr.ErrNodeNotFound: the node manager no longer knows the assigned
+//     DataNode (its session was removed after a restart/replacement), so its
+//     in-memory task manager — and the task with it — is gone.
+//   - merr.ErrImportSysFailed: on this RPC the code is produced only by the
+//     DataNode's task-not-found branch (importv2.WrapTaskNotFoundError when the
+//     queried task is absent from its task manager), i.e. the DataNode is alive
+//     but restarted and lost the task.
+//
+// Everything else (gRPC transport errors, node briefly not serving/not ready,
+// response decode failures) may coexist with a still-running worker task and
+// must be retried by polling, not by re-dispatching.
+func isCopyTaskLostOnWorker(err error) bool {
+	return errors.Is(err, merr.ErrNodeNotFound) || errors.Is(err, merr.ErrImportSysFailed)
+}
+
 // QueryTaskOnWorker polls the DataNode for task execution status.
 //
 // Process flow:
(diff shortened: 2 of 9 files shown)


internal/datacoord/copy_segment_task.go
```diff
@@ -401,9 +410,13 @@ func (t *copySegmentTask) QueryTaskOnWorker(cluster session.Cluster) {
 		TaskID: t.GetTaskId(),
 	}
 	resp, err := cluster.QueryCopySegment(nodeID, req)
-	// Handle RPC error separately to avoid nil resp dereference
+	// Handle RPC error separately to avoid nil resp dereference.
+	// A transient RPC failure (network blip, DataNode restart) must not fail
+	// the whole restore: keep the task InProgress and re-query on the next
+	// tick. The job-level timeout still bounds permanently dead nodes.
```
weiliu1031: This only handles transient query failures; it does not recover from a DataNode restart.

Copy-segment tasks are stored only in the DataNode in-memory task manager. If the assigned DataNode restarts while a task is running, or after it completes but before DataCoord retrieves the result, the worker-side task is permanently lost. Subsequent queries will return either `node not found` or `task not found`.

Since this branch leaves the DataCoord task in `InProgress`, the global scheduler keeps it in `runningTasks` and never requeues it. As a result, the restore job remains `Executing` until the job timeout, which is 24 hours by default.

Could we distinguish transient transport errors from confirmed worker or task loss? For `SessionDel`, `ErrNodeNotFound`, or an explicit task-not-found response, the task should be recovered explicitly, for example by resetting it to `Pending` with `NullNodeID` and rescheduling it, or by failing it with a clear reason instead of silently waiting for timeout.

Please also add a regression test covering DataNode restart or task-not-found. If rescheduling is introduced, partial copied files and allocated build IDs must be handled idempotently.

xiaocai2333:
# Adversarial Review

Reviewed the three fixes with their full surrounding context: the `UpdateTask`/`Clone` chain in `copy_segment_meta.go`, the global scheduler's `schedule()`/`check()` loops, the inspector, the DataNode side (`services.go`, `importv2/task_manager.go`, `copy_segment_utils.go` path generation), and the import pipeline as the reference implementation.

**TL;DR**: all three fixes are logically sound, and fix 3 incidentally fixes a pre-existing data race (update actions used to mutate the shared proto in place while lock-free readers could observe intermediate state). However, **the safety-net claim behind fix 2 does not hold**: the job-level timeout it relies on can never fire due to a pre-existing timestamp-unit mismatch, so the worst case of "retry on transient error" becomes "hang forever".

---

## High

### H1 — The job-level timeout backstop never fires; with query errors now retried, a permanently erroring restore hangs forever

- **Where**: `internal/datacoord/snapshot_manager.go:1252` (write) + `internal/datacoord/copy_segment_checker.go:651` (read); the behavior change is `internal/datacoord/copy_segment_task.go:421-431`
- **Problem**: the copy-segment job is created with `TimeoutTs: uint64(time.Now().Add(jobTimeout).UnixNano())`, but `tryTimeoutJob` interprets it via `tsoutil.PhysicalTime()`, which expects a **hybrid TSO timestamp** (physical ms << 18). UnixNano ≈ 1.78e18; shifted right by 18 and read as milliseconds it lands around **year 2185**, so `time.Now().Before(timeoutTime)` is always true and the timeout check is skipped forever. Compare with the import pipeline: `importutilv2.GetTimeoutTs` correctly uses `tsoutil.ComposeTSByTime` — the copy-segment write site is a unit-mismatch bug.
  This bug is pre-existing, not introduced here. But before this PR a query RPC error failed the job immediately, which masked it. After this PR, the timeout is the *only* backstop — and it is broken. Consequence: if all DataNodes are permanently unreachable, or `QueryCopySegment` keeps erroring for any persistent reason, the task bounces Pending ↔ InProgress indefinitely (the scheduler's exponential backoff bounds the dispatch *rate*, not the *duration*), the job stays `Executing` forever, and the user's restore never reports failure. The snapshot pin self-heals via its TTL, but the job itself is stuck.
- **PR description**: the body states "The job-level timeout still bounds a permanently unreachable node" — this failure mode was not traced end-to-end and the claim is currently false.
- **Suggested fix**: in this PR (or as a prerequisite) compose `TimeoutTs` with `tsoutil.AddPhysicalDurationOnTs(tsoutil.ComposeTSByTime(time.Now()), jobTimeout)` to match import, or parse it symmetrically as UnixNano in `tryTimeoutJob`. At minimum, correct the PR description and add a unit test that the timeout actually fires (writing that test today would immediately expose the bug).

## Medium

### M1 — Transient-blip scenario (node still alive): the old node's task is never cleaned up — slot leak + orphan index files + concurrent same-path writes

- **Where**: `internal/datacoord/copy_segment_task.go:422-424`
- **Problem**: the reset overwrites `NodeId` with `NullNodeID`, and any later `DropTaskOnWorker` goes to the **new** NodeId. If the query error was just a network blip and the old DataNode is still executing the task, and the re-dispatch lands on a **different** node:
  1. The old node's task stays in its `importTaskMgr` until it finishes and is never dropped (there is no orphan-task reaper on the DataNode side — checked `importv2/`). Its task slots keep being reported via `QuerySlot`, reducing that node's capacity until the DataNode restarts.
  2. Each dispatch re-runs `AssembleCopySegmentRequest`, which allocates fresh `newBuildIDs`, so attempt-1's index files sit under stale buildID paths as orphans (`CleanupCopiedFiles` only triggers on a Failed-state drop, which never happens here); only generic GC can reclaim them.
  3. Both nodes concurrently write binlogs to the **same deterministic target paths** (paths are a source-ID → target-ID substitution). Content is identical so this is benign in practice, but it relies on the object store's semantics for concurrent PUTs to the same key.
  Mitigating factors: re-dispatch to the **same** node is a clean resume (`taskManager.Add` dedupes by taskID, `task_manager.go:45-53`), which is the common single-DataNode case; and the import pipeline has the same stale-task gap, so "mirroring import" is a defensible position.
- **Suggested fix**: before resetting, issue a best-effort `cluster.DropCopySegment(nodeID, taskID)` and ignore the error (same pattern as `AbortAndRemoveTask`). If the node is truly gone the call fails harmlessly; if it was a blip, this aborts the old execution and avoids the double-write and slot leak.

## Low

### L1 — New test inserted in the wrong spot, orphaning another function's doc comment

- **Where**: `internal/datacoord/copy_segment_task_test.go:722` (insertion point of `TestClone_DeepCopiesProto`)
- The new test landed between the comment `// createTestCopyTask creates a minimal CopySegmentTask for testing syncVectorScalarIndexes.` and `createTestCopyTask` itself, so the comment now documents the wrong function. Move the test above the comment or next to the other tests.

### L2 — `TestQueryTaskOnWorker_NodeGoneResetsToPending` covers the same code path as the RPC-error test

- **Where**: `internal/datacoord/copy_segment_task_test.go:524-546`
- The implementation does not special-case `NodeNotFound` — both tests exercise the identical `err != nil` branch with a different error value, adding no coverage. Fine to keep as intent documentation, but worth noting there is no distinct behavior behind the name.

### L3 — Progress transiently resets to zero on resume

- **Where**: `internal/datacoord/copy_segment_checker.go:376-378`
- Tasks persisted while the job was stuck in Pending are dispatched (and can even complete) regardless of job state, since the inspector doesn't filter by job state. On resume, `UpdateCopyJobProgress(0, total)` overwrites any real progress with 0 until the next `checkCopyingJob` tick (~2s) corrects it. Display-only; consider writing only the total.

---

## Dimension checklist

1. **Functional correctness** — Fix 1: source-ID coverage is sound (each source segment appears exactly once in `IdMappings`, `snapshot_manager.go:1173`); when fully covered, `groups` is empty and only the Executing transition is retried — idempotent. Fix 2: the scheduler's `check()` sees the task back in Init, records backoff, removes it from `runningTasks` and re-queues it (`global_scheduler.go:339-342`) — the mechanism described in the PR matches the code. Fix 3: `UpdateTask` stores the clone's proto back into the cached task after a successful save (`copy_segment_meta.go:629`), so the deep copy does not break state propagation. **Finding: H1** (the backstop claim).
2. **Concurrency safety** — the deep copy fixes a pre-existing race (in-place proto mutation under `UpdateTask` vs. lock-free getters). The transient window between `check()`'s remove/re-push and the inspector's `Enqueue` is covered by Enqueue's dedup (pre-existing pattern). No new issues.
3. **Error handling** — a failed reset leaves the task InProgress and retries next tick; logging is structured with ctx. No issues.
4. **Resource leaks** — **Finding: M1**.
5. **Performance** — one `proto.Clone` per `UpdateTask` over lightweight ID mappings; negligible. No issues.
6. **Test coverage** — the 5 new tests target the right scenarios. Gaps: the `resetErr` branch is untested, and **there is no test that the job timeout fires (adding one would expose H1)**; plus L2.
7. **API compatibility** — no proto/wire/config changes. The behavior change (query RPC errors no longer fail the job) is the PR's intent and aligns with import semantics.
8. **Maintainability** — comments match the implementation; log fields are clear. **Finding: L1**.

## Overall

**Needs work** — the fixes themselves are well built, but **H1 must be addressed**: either fix the `TimeoutTs` unit (recommended; one line + one test) or restate the PR description to say the timeout backstop is currently non-functional and tracked as a follow-up. M1 is a one-line best-effort drop that meaningfully reduces the stale-task blast radius. L1–L3 are trivial cleanups.

*(Note: `ci-v2/build-ut-cov` is currently red with 1 C++ test failure — this PR is Go-only, so likely unrelated/flaky, but it needs a rerun before merge.)*

internal/datacoord/copy_segment_checker.go
```diff
@@ -357,7 +370,9 @@ func (c *copySegmentChecker) checkPendingJob(job CopySegmentJob) {
 			mlog.Int("segmentCount", len(group)))
 	}
 
-	// Step 5: Update job state to Executing
+	// Step 5: Update job state to Executing. This also runs when all segments
```
xiaocai2333, line 390: **[P1] Guard the Pending -> Executing transition against stale job snapshots**

`checkPendingJob` is invoked with a job object obtained before this function starts, but partial tasks can already be dispatched while the parent job is still Pending: the inspector enqueues Pending tasks without filtering the job state. If one of those tasks fails concurrently, `markTaskAndJobFailed` replaces the current job with a Failed clone and releases the snapshot pin. This function still holds the old Pending object and then calls `UpdateJob(...Executing)`, which updates the current cached job without checking its previous state. The result is a Failed job being resurrected as Executing after its snapshot protection has already been released.

Please re-read/validate the current job state under the metadata lock and make this an expected-state transition (Pending -> Executing only); return without creating more tasks if the current job is already terminal. A regression test can pass a stale Pending job to this method after updating the cached job to Failed and assert that it remains Failed.

internal/datacoord/copy_segment_task.go
```diff
@@ -401,9 +411,23 @@ func (t *copySegmentTask) QueryTaskOnWorker(cluster session.Cluster) {
...
+	// Handle RPC error separately to avoid nil resp dereference.
+	// A query error is transient (network blip) or means the assigned node is
+	// gone (DataNode restart/replacement -> NodeNotFound). Either way, failing
+	// the whole restore is wrong: reset the task to Pending so the scheduler
+	// re-dispatches it to a live node, matching the import pipeline. Leaving it
+	// InProgress would make the scheduler poll the same (possibly dead) node
+	// forever, since only Pending tasks are re-dispatched.
 	if err != nil {
-		t.markTaskAndJobFailed(fmt.Sprintf("query copy segment RPC failed: %v", err))
+		if resetErr := t.copyMeta.UpdateTask(context.TODO(), t.GetTaskId(),
```
xiaocai2333, line 460: **[P1] Do not re-dispatch on every QueryCopySegment error**

A query error does not prove that the worker-side task is gone. This call path also returns transient gRPC timeouts/unavailability, temporary service-not-ready errors, and response parsing/unmarshal errors. Resetting all of them to Pending makes the global scheduler remove the task from `runningTasks` and dispatch it again, potentially to another DataNode, while the original task may still be copying. Clearing `NodeID` also loses the only handle needed to drop that original task. Each re-dispatch rebuilds the request and allocates fresh build IDs, so a single polling blip can produce concurrent copies plus orphan worker state/files.

Please distinguish confirmed worker/task loss (for example node-not-found or an explicit task-not-found result) from ambiguous transport failures. Keep the task InProgress and query the same node again for transient failures; handle deterministic decode/protocol errors explicitly instead of re-dispatching. The regression test should exercise the global scheduler and verify that a one-off transport error does not invoke `CreateCopySegment` a second time.

xiaocai2333 (commented):
Adversarial review found two correctness issues in the recovery paths.

internal/datacoord/copy_segment_task.go
xiaocai2333: Addressed in b6333f05ac.

`QueryTaskOnWorker` now distinguishes the two cases via `isCopyTaskLostOnWorker` (every error construction site on this RPC path was audited):

- **Confirmed loss** — `merr.ErrNodeNotFound` (node manager dropped the session after a DataNode restart/replacement) or the DataNode's explicit task-not-found response (surfaced as `ErrImportSysFailed`, the only producer of that code on this path): the task is reset to `Pending` with `NullNodeID`, so the global scheduler removes it from `runningTasks` and re-dispatches it to a live node.
- **Transient transport errors** (RPC timeout, node briefly not ready, decode failures): the task stays `InProgress` and is simply polled again next round — re-dispatching here could start a concurrent duplicate copy.

Regression tests: `TestQueryTaskOnWorker_NodeGoneResetsToPending` (DataNode restart / NodeNotFound), `TestQueryTaskOnWorker_TaskNotFoundOnWorkerResetsToPending` (task-not-found, with a `Status -> Error` wire round-trip), and `TestQueryTaskOnWorker_TransientRPCErrorKeepsInProgress`.

Idempotency on re-dispatch: target binlog paths are deterministic transforms of the source paths (same content on overwrite), and every dispatch allocates fresh buildIDs — files from a partial earlier attempt are never referenced by meta and are reclaimed by GC. Covered by `TestAssembleCopySegmentRequest_RedispatchAllocatesFreshBuildIDs`.

Related: the job-level timeout backstop this relies on had a pre-existing unit bug (TimeoutTs stored as UnixNano but decoded as a hybrid TSO, so it could never fire) — fixed in 225375a6f4 with tests that the timeout actually fires.

internal/datacoord/copy_segment_checker.go
xiaocai2333: Fixed in 10e04ed0b0.

- `checkPendingJob` now re-reads the cached job at entry and returns without creating tasks if the job is gone or no longer Pending.
- The `Pending -> Executing` transition goes through the new `CopySegmentMeta.UpdateJobInState`, which validates the expected state and applies the update under the same metadata write lock — a job that concurrently moved to a terminal state is skipped, so a Failed job can no longer be resurrected as Executing after its snapshot pin was released.

Regression tests: `TestUpdateJobInState_AppliesOnlyWhenStateMatches` (meta-level guard), and `TestCheckPendingJob_StaleSnapshotDoesNotResurrectFailedJob` — exactly the scenario described here: the cached job is moved to Failed, then `checkPendingJob` is called with the stale Pending snapshot, and the job must remain Failed with zero tasks created (the allocator mock would fail the test on any AllocID call).

internal/datacoord/copy_segment_task.go
xiaocai2333: Fixed in b6333f05ac (behavior) + 10e04ed0b0 (scheduler-level test).

Only confirmed worker/task loss triggers a re-dispatch: `merr.ErrNodeNotFound` (session gone) or the DataNode's explicit task-not-found response (`ErrImportSysFailed`, the only producer of that code on this RPC path). Everything ambiguous — gRPC transport errors, service-not-ready, response decode failures — keeps the task `InProgress` on the same node and is retried by polling; a persistently erroring path is now bounded by the job-level timeout fixed in 225375a6f4.

Scheduler-level regression test as requested: `TestScheduler_OneOffTransientErrorDoesNotRedispatch` runs the real `globalTaskScheduler` with a mock cluster that fails exactly one `QueryCopySegment` call, and verifies the task stays in `runningTasks`, keeps being polled on the same node across ticks, `CreateCopySegment` is never invoked a second time, and `NodeID` is unchanged.

weiliu1031:
nit (non-blocking): `job` is the snapshot captured before `checkCopyingJob` runs. In the narrow case where the deadline has elapsed and the final tasks complete in the same checker round, `finishJob` may persist `Completed` first, while `tryTimeoutJob` still sees the old `Executing` snapshot and changes the cached job to `Failed`.

This is a very small boundary window, but we could avoid overwriting a terminal transition by re-reading the current job state or making the timeout update conditional on the expected state. Fine to handle as a follow-up if preferred.

weiliu1031:
nit (non-blocking): This reset does not check the current state of the parent job. If another task moves the job to `Failed` while this query is in flight, a delayed node/task-not-found response can reset this task to `Pending`, allowing the scheduler to issue an additional dispatch before `checkFailedJob` converges its state back to `Failed`.

This should normally result in at most a short-lived extra attempt rather than a persistent retry loop. As a defensive hardening, we could atomically verify that the task is still `InProgress` and the parent job is still active before resetting it, or re-check the job immediately before dispatch.

xiaocai2333:
@weiliu1031 Thanks — both nits were valid, and I've addressed them in a follow-up: #51641 (master) and #51642 (3.0 cherry-pick, which also carries this PR since 3.0 didn't have it yet).

**[nit 1](https://github.com/milvus-io/milvus/pull/51527#issuecomment-5023256048) — `tryTimeoutJob` overwriting a terminal transition.** Confirmed exactly as you described: `Start()` pushes one snapshot through `checkCopyingJob` (which may `finishJob` -> `Completed`) and then `tryTimeoutJob` in the same round, so the latter passes its state switch on the stale `Executing` snapshot and, with the deadline elapsed, calls `UpdateJobStateAndReleaseRef(...Failed)` — which applied the action unconditionally. The existing `wasTerminal` logic only guarded the *unpin* side effect, not the state mutation, so it didn't prevent this. A restore that had just completed was reported as `Failed`.

Fixed by making the transition conditional on the current state: `UpdateJobStateAndReleaseRef` now skips the update when the **current cached** job is already terminal. There's no legitimate terminal -> terminal transition (every caller drives a non-terminal job to `Completed`/`Failed`), so observing one always means a concurrent path won and this caller is stale. I put the guard in the meta layer rather than in `tryTimeoutJob` because the check has to be atomic with the mutate and the lock lives there — a re-read inside `tryTimeoutJob` would still be TOCTOU. It also hardens `markTaskAndJobFailed` and `finishJob` against the same race.

**[nit 2](https://github.com/milvus-io/milvus/pull/51527#issuecomment-5023256435) — reset not checking the parent job state.** Also confirmed. Added `CopySegmentMeta.UpdateTaskInStateIfJobActive`, which applies the actions only if the task is still `InProgress` **and** its parent job is still active, with both checks and the update under the same write lock — i.e. the atomic verify-before-reset you suggested. The worker-loss path now goes through it, so a delayed confirmed-loss response can no longer earn the scheduler an extra dispatch for an already-dead job.

Both are the same defect class as the P1 fix in this PR (`UpdateJobInState`) — a transition applied from a stale snapshot — so the follow-up reuses that pattern.

Tests: `TestTryTimeoutJob_DoesNotOverwriteTerminalJob`, `TestQueryTaskOnWorker_WorkerLossDoesNotResetWhenJobTerminal`, plus meta-level coverage of both guards including the positive cases (the guard must not block the legitimate `Executing -> Failed` timeout transition). Each regression test was mutation-checked — neutering the corresponding guard makes it fail — so they exercise the defect rather than passing vacuously.

Issue #50915, opened 2026-06-30
[Bug]: StreamingNode panics when Vortex growing flush close/upload fails with XAmzContentChecksumMismatch


yanliang567:
## Environment

- **Milvus Version**: `harbor.milvus.io/milvusdb/milvus:master-20260628-4018b294`
- **Deployment Mode**: cluster, Milvus Operator
- **Instance / Namespace**: `qa-milvus/laion1b-test-2`
- **MQ / WAL**: streaming mode, Pulsar WAL
- **Object Storage**: MinIO/S3-compatible backend, bucket `bucket-laion2`
- **Workload**: mixed read/write LAION 1B benchmark
- **Workflow**: https://argo-workflows.zilliz.cc/workflows/qa/laion1b-test-cron-sn-nullable-1782741600?tab=workflow
- **Failed pod**: `laion1b-test-2-milvus-streamingnode-7f796c8cfc-2h8rq`

Kubernetes termination status:

```text
restartCount=1
lastReason=Error
lastExitCode=134
lastStartedAt=2026-06-29T08:50:41Z
lastFinishedAt=2026-06-29T21:09:23Z
```

### Steps

1. Deploy Milvus cluster with StreamingNode enabled, using image `master-20260628-4018b294`.
2. Enable StorageV3 / Vortex:

```yaml
common:
  enabledJSONShredding: false
  storage:
    useLoonFFI: true
dataNode:
  storage:
    format: vortex
```

3. Use a large LAION-style collection (`laion_stable_10_nullable`) with continuous mixed traffic.
4. Run the mixed read/write workflow for 7h:
   - 20 concurrent users
   - task weights `load:query:search:upsert = 1:4:13:2`
   - search: `nq=1`, `top_k=100`, timeout 10s
   - query: `count(*)`
   - upsert: `numRows=500`, random id/vector, timeout 60s
5. During the run, StreamingNode flushes growing segment data through the Vortex writer path.
6. A Vortex object close/upload fails with `XAmzContentChecksumMismatch`.
7. StreamingNode panics and restarts with exit code 134. Search/query then see channel unavailable.

### Trigger Conditions

- **Frequency**: observed once in this latest laion1b workflow after several hours of mixed DQL/DML traffic.
- **First observed in this run**: `2026-06-29T21:09:22Z`
- **Direct trigger**: `FlushGrowingSegmentData` fails while closing/uploading a Vortex file.
[...]

PR #51530, opened 2026-07-17
enhance: preserve storage error classification across segcore boundaries

jiaqizho:
issue: #50915

Storage failures were being flattened into generic unexpected, read, or write errors as they crossed milvus-storage, Arrow, segcore, and cgo boundaries. This hid the difference between retryable network or service failures and permanent storage, permission, configuration, or data errors, leaving upper layers without reliable information for retry and failure handling.

This change routes storage-backed Arrow statuses through ToSegcoreError across filesystem initialization, packed readers and writers, segment loading and flushing, manifest translation, memory planning, and text or LOB reads. The resulting SegcoreError is preserved through ThrowInfo or FailureCStatus, while the StorageError and StorageTransientError codes are registered in merr so Go keeps the intended retriable classification.

The packed Loon FFI handler now reads the structured error code instead of treating every failure as transient. Errors explicitly marked retryable by milvus-storage continue to carry ErrLoonTransient, while permanent failures become typed storage errors. Returned messages also include the original Loon code, and the affected sync and external refresh paths now handle the narrowed transient signal and permanent storage failures consistently. Packed reader shutdown keeps its previous behavior by releasing the reader without changing how close-status failures are surfaced. The remaining nested milvus-table FFI code-loss path is documented with a FIXME because it should be addressed through a consistent boundary design instead of introducing a one-off exception type.

### DIFF
--- a/internal/storagev2/packed/ffi_common.go
+++ b/internal/storagev2/packed/ffi_common.go
@@ -24,16 +24,9 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 )
 
-// ErrLoonTransient marks any failure surfaced by the loon FFI layer. Today
-// milvus-storage does not expose structured error codes, so callers cannot
-// distinguish a recoverable concurrent-transaction conflict from a hard IO
-// error. We treat all loon failures as retryable for now and rely on a
-// bounded retry budget plus outer error handling to keep the worst case
-// finite.
-//
-// TODO(storage v3): once milvus-storage exposes explicit error codes, narrow
-// this sentinel to only the concurrent-transaction case (FailResolver) and
-// let other errors propagate immediately as retry.Unrecoverable.
+// ErrLoonTransient marks failures that milvus-storage explicitly classifies as
+// retryable through loon_ffi_is_retryable_errcode. Permanent Loon FFI failures
+// use the typed merr storage error instead.
 var ErrLoonTransient = errors.New("loon FFI transient error")
 
 // Property keys exported by milvus-storage/ffi_c.h.
@@ -315,13 +308,16 @@ func injectExternalSpecProperties(properties *C.LoonProperties, collectionID int
 func HandleLoonFFIResult(ffiResult C.LoonFFIResult) error {
 	defer C.loon_ffi_free_result(&ffiResult)
 	if C.loon_ffi_is_success(&ffiResult) == 0 {
+		errCode := int(ffiResult.err_code)
 		errMsg := C.loon_ffi_get_errmsg(&ffiResult)
 		errStr := "Unknown error"
 		if errMsg != nil {
 			errStr = C.GoString(errMsg)
 		}
-
-		return merr.Wrapf(ErrLoonTransient, "FFI operation failed: %s", errStr)
+		if C.loon_ffi_is_retryable_errcode(ffiResult.err_code) != 0 {
+			return merr.Wrapf(ErrLoonTransient, "FFI operation failed (code %d): %s", errCode, errStr)
+		}
+		return merr.WrapErrStorageMsg("FFI operation failed (code %d): %s", errCode, errStr)
 	}
 	return nil
 }
--- a/internal/storagev2/packed/ffi_common.go
+++ b/internal/storagev2/packed/ffi_common.go
@@ -24,9 +24,16 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 )
 
-// ErrLoonTransient marks failures that milvus-storage explicitly classifies as
-// retryable through loon_ffi_is_retryable_errcode. Permanent Loon FFI failures
-// use the typed merr storage error instead.
+// ErrLoonTransient marks any failure surfaced by the loon FFI layer. Some
+// milvus-storage paths can still lose their structured error detail and fall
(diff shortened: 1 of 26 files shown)


internal/storagev2/packed/ffi_common.go
```diff
@@ -315,13 +308,16 @@ func injectExternalSpecProperties(properties *C.LoonProperties, collectionID int
...
 		errStr := "Unknown error"
 		if errMsg != nil {
 			errStr = C.GoString(errMsg)
 		}
-
-		return merr.Wrapf(ErrLoonTransient, "FFI operation failed: %s", errStr)
+		if C.loon_ffi_is_retryable_errcode(ffiResult.err_code) != 0 {
+			return merr.Wrapf(ErrLoonTransient, "FFI operation failed (code %d): %s", errCode, errStr)
+		}
+		return merr.WrapErrStorageMsg("FFI operation failed (code %d): %s", errCode, errStr)
```
liliu-z: A transient manifest-commit failure now crashes the datanode instead of retrying. milvus-storage's Transaction::Commit re-wraps any non-conflict write failure — including S3 throttle/network errors that carried a StorageTransientNetwork detail — into a plain arrow::Status::IOError with no ExtendStatusDetail (transaction.cpp:438), so loon_transaction_commit falls back to LOON_LOGICAL_ERROR (code 4). Since 4 is not an ExtendStatusCode, loon_ffi_is_retryable_errcode(4) is false and this line returns permanent ErrStorage; classifyLoonErr then marks it retry.Unrecoverable, retry.Do returns without retrying, and the sync task's deferred HandleError hits the default panic(err) handler. Pre-PR every loon error was ErrLoonTransient and was retried within the bounded budget, so a transient blip recovered rather than crashing. Keep code-4 / detail-less commit failures retryable (or plumb the transient detail through the FFI) until milvus-storage stops flattening it.

weiliu1031 (commented):
### Findings

- [P1][Blocker] The growing flush path does not consume the retriable classification and still reaches the default panic handler
  Location: `internal/core/src/segcore/segment_c.cpp:2363`, `internal/flushcommon/syncmgr/growing_source.go:685`, `internal/flushcommon/writebuffer/options.go:49`
    - Impact: Even when C++ correctly returns `2045`, `FlushGrowingData` still returns the error directly, after which `HandleError` invokes the default `panic(err)`. The StreamingNode crash described in [issue 50915](https://github.com/milvus-io/milvus/issues/50915) is therefore not fixed.
    - Evidence: `GrowingSourceSyncTask.writeRetryOpts` is only stored and is never applied to `FlushGrowingData`. Code `2045` was already registered as retriable in the base branch, so the `merr` changes in this PR add no new behavior that consumes the retriable classification.
    - Recommendation: Consume `2045` at the growing-source task layer using the existing retry configuration, or safely reschedule the task. Until the entire flush retry path has been proven idempotent, transient storage errors should at least be prevented from reaching the default panic handler.
    - Source: Correctness, Code quality.

- [P2][Blocker] Real MinIO `UNKNOWN 400` errors are still downgraded to permanent `2044`
  Location: `internal/core/src/segcore/segment_c.cpp:2363`, `internal/core/src/segcore/packed_reader_c.cpp:222`
    - Impact: `XAmzContentChecksumMismatch` and `IncompleteBody` still do not receive code `2045`. On packed paths that previously returned retriable `FileReadFailed` or `FileWriteFailed` errors, this also changes the existing retry semantics.
    - Evidence: The pinned `milvus-storage@f9f5bd1` revision returns unrecognized `UNKNOWN 400` failures as plain `IOError` statuses without `ExtendStatusDetail`. The Vortex FFI also preserves them as plain I/O errors, so `ToSegcoreError` ultimately maps them to the non-retriable code `2044`.
    - Recommendation: Classify these real failures and attach structured details at their producers. Ensure that FFI and transaction wrappers do not reconstruct them as plain `IOError` statuses, and add end-to-end tests covering the path from the real producer to the resulting Go status.
    - Source: Correctness, Code quality.

- [P2][Blocker] The newly activated `PackedInvalidArgs` mapping blames internal invariant violations on user input
  Location: `internal/core/src/segcore/packed_writer_c.cpp:134`, `pkg/util/merr/segcore.go:108`
    - Impact: Mismatches involving internally generated schemas, paths, column groups, or record batches become `2042 InputError`, causing the default `retry.Do` behavior to terminate immediately and exposing an incorrect blame classification to upper layers.
    - Evidence: Producers of `PackedInvalidArgs` include null filesystems or schemas, mismatched path and group counts, missing field IDs in schemas, and internal column-index overflows. All of these values are constructed internally by Milvus. A comment in the pinned dependency also explicitly states that this mapping was previously dormant; this PR turns it into an active path.
    - Recommendation: Map `PackedInvalidArgs` produced by internal Milvus calls to a permanent system error, or split genuine caller-input failures from internal invariant violations in `milvus-storage`.
    - Source: Correctness, Code quality.

- [P3][Non-blocker] The PR description contradicts the current Loon handler behavior
  Location: `internal/storagev2/packed/ffi_common.go:315`
    - Impact: The PR description claims that permanent Loon failures are now converted into typed storage errors, but the current handler does not read `err_code` at all and still wraps every failure with `ErrLoonTransient`.
    - Recommendation: Update the PR body and commit description to state explicitly that the implementation still uses a bounded all-transient fallback.
    - Source: Code quality, G3 over-claim review.

PR #51531, opened 2026-07-17
enhance: migrate sealed segment misc state to runtime snapshots

congqixia:
Related to #51068

- move row count, mmap markers, average field sizes, and skip index into runtime state
- remove duplicated live resource containers from chunked sealed segments
- preserve old and new resource lifetimes through COW publication
- reuse one published snapshot capture within each internal method
- add runtime lifetime and skip index regression coverage

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -242,60 +242,11 @@ cancel_and_erase_scalar_index(
     }
 }
 
-static inline void
-cancel_and_clear_scalar_indexings(
-    std::unordered_map<FieldId, index::CacheIndexBasePtr>& scalar_indexings) {
-    for (auto& [_, index] : scalar_indexings) {
-        cancel_warmup(index);
-    }
-    scalar_indexings.clear();
-}
-
-static inline void
-cancel_and_erase_ngram_index(
-    std::unordered_map<
-        FieldId,
-        std::unordered_map<std::string, index::CacheIndexBasePtr>>&
-        ngram_indexings,
-    FieldId field_id,
-    const std::string& nested_path) {
-    auto field_it = ngram_indexings.find(field_id);
-    if (field_it == ngram_indexings.end()) {
-        return;
-    }
-
-    auto& path_indexings = field_it->second;
-    if (auto path_it = path_indexings.find(nested_path);
-        path_it != path_indexings.end()) {
-        cancel_warmup(path_it->second);
-        path_indexings.erase(path_it);
-    }
-
-    if (path_indexings.empty()) {
-        ngram_indexings.erase(field_it);
-    }
-}
-
-static inline void
-cancel_and_clear_ngram_indexings(
-    std::unordered_map<
-        FieldId,
-        std::unordered_map<std::string, index::CacheIndexBasePtr>>&
-        ngram_indexings) {
-    for (auto& [_, path_indexings] : ngram_indexings) {
-        for (auto& [__, index] : path_indexings) {
-            cancel_warmup(index);
-        }
-    }
-    ngram_indexings.clear();
-}
-
 PinWrapper<const storagev2translator::TimestampIndexCell*>
 ChunkedSegmentSealedImpl::PinTimestampIndex(
     const std::shared_ptr<const RuntimeResourceState>& runtime,
     milvus::OpContext* op_ctx) const {
-    auto slot = runtime != nullptr ? runtime->timestamp_index_slot
-                                   : *timestamp_index_slot_.rlock();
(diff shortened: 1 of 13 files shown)


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -3801,17 +3848,35 @@ ChunkedSegmentSealedImpl::DropFieldData(const FieldId field_id,
...
+        old_column = get_column(snapshot->runtime, field_id);
+    }
     if (old_column) {
         old_column->CancelWarmup();
     }
     if (runtime != nullptr) {
         runtime->fields.erase(field_id);
         runtime->array_offsets_map.erase(field_id);
+        runtime->mmap_field_ids.erase(field_id);
+        runtime->variable_fields_avg_size.erase(field_id);
```
liliu-z: DropFieldData now erases variable_fields_avg_size for the dropped field (added at 3867 for the reopen/runtime path and 3876 for the direct path; the pre-PR DropFieldData had no such erase). On the reopen path that releases a variable-length field's raw binlogs while the field stays retrievable through an index that carries raw data (ComputeDiffBinlogs/ComputeDiffColumnGroups add it to field_data_to_drop when prefer_field_data is false and field_index_has_raw_data is set), get_field_avg_size then returns 0 for that field. The retrieve size guard at SegmentInterface.cpp:233 (`output_data_size += get_field_avg_size(field_id) * result_rows`) therefore stops counting that still-returned field, so an oversized retrieve of it escapes the limit_size check. Erase the avg-size entry only when the field is truly removed from schema, not when it remains index-retrievable.

xiaofan-luan:
@congqixia heads-up — this overlaps two PRs I have open in the same sealed-segment runtime-snapshot area, so let's coordinate to avoid duplicated/conflicting work:

- **#51602** — per-**operation** published-state pinning: `PinOpSnapshot(op_ctx)` captures one `PublishedSegmentState` at Search/Retrieve entry, stores it in the operation's `OpContext` (type-erased `pinned_segment_state` + owner guard, already merged into milvus-common as `1.0.0-6b16d93`), and every read resolves it via `CapturePublishedState(op_ctx)`. Independent commit on master.
- **#51441** — stats skip-index correctness; already migrates `skip_index_metrics` into `RuntimeResourceState` (v1) + column-group ownership (v2/v3), plus skip regression tests. Stacked on #51602.

### How #51531 overlaps

| #51531 does | overlap |
| --- | --- |
| move **skip index** into runtime state | duplicates #51441's `skip_index_metrics` migration |
| **reuse one published snapshot capture within each internal method** | subsumed by #51602 — see below |
| skip-index regression coverage | overlaps #51441 / #51602 tests |
| move **row count / mmap markers / average field sizes** into runtime state | **unique to #51531** — nothing else covers this |

### Per-method capture vs per-operation pin

The "one capture per internal method" here is a real perf win, but it only makes a **single method** internally consistent — it re-captures across method boundaries. The window that matters is **between** method calls, which is the common case in a real query:

- a multi-batch scan (`ProcessDataChunks` calls `chunk_size` / `get_batch_views` per chunk — a separate capture each batch),
- Retrieve's two phases (`visitor.get_retrieve_result` filter → `FillTargetEntry` fetch — separate captures),
- chunk count cached at `InitSegmentExpr` vs data read per batch.

If a concurrent `Reopen` (raw-data reclaim after an index build, drop-field, compaction) removes/replaces the queried field's column between those captures, the later capture observes the new snapshot and hits `AssertInfo(field_data_ready)` / a null `get_column`. Putting the snapshot in `OpContext` (one per operation, threaded through every method) is exactly what closes that cross-method window — per-method reuse can't, since the second phase simply captures the *newer* snapshot.

So **per-operation pin is a strict superset of per-method reuse**: once #51602 lands, an internal method just resolves the OpContext-pinned snapshot and gets per-op consistency for free.

### Proposed split (no one loses work)

1. **#51602 (pinning)** lands first — the general snapshot mechanism.
2. **#51441 (skip)** stacked on it — skip correctness + `skip_index_metrics` migration.
3. **#51531** slims to its **unique** part — migrating **row count / mmap markers / average field sizes** into `RuntimeResourceState` — and rebases on top of #51602 + #51441. Drop the skip-index migration (already in #51441) and replace the per-method captures with resolving the OpContext pin (from #51602). Bonus: state you migrate into the runtime snapshot is then automatically covered by the pin.

One reconciliation to note: #51602 currently documents `num_rows_` / `col_index_meta_` / `is_sorted_by_pk_` as immutable members ("must not gain a Reopen write path"). #51531 moves row count into the runtime snapshot, so when it lands on top, please update those comments (row count then lives in the snapshot and is pinned).

Happy to help rebase/split — let me know how you'd like to divide it.

liliu-z (commented):
/lgtm
/approve

Issue #51533, opened 2026-07-17
[Enhancement]: Proxy meta cache invalidations are O(N) scans under the global lock


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Rebuild the proxy meta cache so that every invalidation is O(1) instead of a linear scan under the global write lock.

An audit of `internal/proxy/meta_cache.go` found every id-, alias- or database-driven invalidation scanning under `m.mu`:

| Scan | Scope | Trigger |
|---|---|---|
| find an id's name keys | O(all cached collections) | every Load/Release/Drop/Rename/Alter broadcast |
| partition sweep (`StaleIf` prefix) | O(all cached partition entries) | nested in every eviction |
| alias cleanup | O(the db's aliases) | every eviction |
| partition-DDL name expansion | O(db's collections + aliases) | partition DDL |
| `RemoveDatabase` | O(db entries) + O(all partition entries) | database DDL |

Proposed design (doc: `docs/design-docs/design_docs/20260716-proxy-metacache-id-inverted-index.md`):
1. Primary store keyed by the cluster-unique collection id as the single source of truth; name/alias resolution become hints validated against the primary on read; partition caches keyed by id; database invalidation via a generation counter; lazy GC reclamation.
2. Fill/invalidation ordering via a fill RWMutex (drain in-flight describes before evicting) replacing the per-collection timestamp floor and its gaps.
3. Rootcoord forwards the pre-alter alias target id in the AlterAlias expiration so both targets are evicted by id, closing the concurrent-AlterAlias stale-alias race at the source.

### Why is this needed?

At a large cached working set (1M-collections target), each invalidation broadcast stalls all queries on the proxy for a full-cache sweep under the write lock. The timestamp floor also left known staleness gaps (unversioned alias write-backs, RequestTime==0 compat bypass, RemoveDatabase resurrection race).

### Anything else?

Follow-ups tracked in the design doc: LRU/TTL bound for the primary store, optional concurrent-map reads, WAL-subscription meta replication as the endgame.

PR #51313, opened 2026-07-13
fix: fill collection_name and keep request id in DescribeCollection queried by id

xiaofan-luan:
issue: #51253

> Rebased on master after #51254 merged. The `db_name`/`db_id` part of the original PR is now covered by #51254; this PR is reduced to the two remaining gaps on the same query-by-id path.

## Problem

When `DescribeCollection` is called with **only a `collectionID`** — no collection name, no db name (e.g. the HTTP management API on `:9091`) — two gaps remain after #51254:

1. **Top-level `collection_name` is empty** on both the cached provider path and the `describeCollectionTask` path. The response is initialized from the (empty) `request.CollectionName` before the name is resolved, and never backfilled.
2. **The cached provider discards the caller-provided collection id.** After deriving the name from the id, it re-resolves the id from that name via `GetCollectionID(db, name)`. Id-only lookups are cached under the request db name — empty here — so with same-name collections in different databases, a concurrent cache refresh can rebind the entry under the `""` key and the request would silently describe the wrong collection (the derived name matches, the id does not).

## Fix

- `service_provider.go`:
  - Backfill `resp.CollectionName` from the resolved cached schema when the request carried no name. Requests that pass a name (including aliases) still echo it unchanged.
  - Skip the name→id re-resolution when the caller already supplied the id. `GetCollectionInfo` already validates the cache entry against the id (`collInfo.collID != collectionID` → refresh by id), so the caller-provided id is authoritative end to end. Name+id requests keep the existing name-precedence behavior.
  - Resolve identifiers on local copies instead of rewriting `request.CollectionName`/`CollectionID` in place — the access log interceptor and the success metric labels serialize the request after the handler returns, and previously recorded the resolved values instead of what the client sent.
- `task.go`: backfill `t.result.CollectionName` from the coordinator result on the non-cached path.
- `meta_cache.go`: fix a stale comment on `ResolveCollectionAlias` — `update()` now always keys `collInfo` by the real collection name (aliases live in the separate alias map), the old comment described pre-refactor behavior.
- Tests:
  - `TestCachedProxyServiceProvider_DescribeCollection_ByIDFillsNameAndUsesRequestID`: drives the id-only path; `GetCollectionID` is deliberately not mocked, so any regression back to re-resolving the id panics the test.
  - `TestDescribeCollectionTask_FillsNameFromResultWhenQueriedByID`: same assertion for the non-cached path.
[...]

PR #51369, opened 2026-07-14
fix: [3.0] return actual db and collection identity in DescribeCollection queried by id

xiaofan-luan:
issue: #51253
pr: #51313

Combined cherry-pick of the two master fixes for `DescribeCollection` queried by collection id only (e.g. the HTTP management API on `:9091`):

- #51254 — return the actual `db_name`/`db_id` from the proxy meta cache instead of echoing the (empty) request db name when `proxy.enableCachedServiceProvider` is on.
- #51313 — two commits:
  1. Backfill the top-level `collection_name` on both the cached provider path and the `describeCollectionTask` path; stop discarding the caller-provided collection id on the cached path; resolve identifiers on local copies instead of rewriting the request in place (access logs / metric labels read it after the handler returns).
  2. Rework the proxy meta cache to be entity-keyed: fills land under the collection's actual database (the bogus `""` bucket for id-only lookups no longer exists, ending duplicate entries and same-name cross-database eviction), by-id lookups are served from a cluster-wide `collectionID → entry` index in O(1) (matching rootcoord, which ignores the db name for by-id describes), and empty-db name lookups normalize to `default` mirroring rootcoord's backward-compat normalization.

All commits cherry-picked clean; no conflicts, no 3.0-specific adjustments. Verified on this branch: `TestMetaCache*`, `TestCachedProxyServiceProvider*`, `TestDescribeCollectionTask*`, alias and partition test sets all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Update: cherry-picked the id-primary meta cache rebuild from the master PR

Two additional commits cherry-picked from #51313 (clean picks, same base state): the AlterAlias pre-alter target id forwarding (additive `old_collection_id` proto field) and the id-primary cache rebuild with O(1) invalidation. See #51313's description and `docs/design-docs/design_docs/20260716-proxy-metacache-id-inverted-index.md` for the full design. issue: #51533

---

## Update: simplification series cherry-picked from #51313 (final form)

Same commits as the master PR's id-primary simplification series (see #51313's description and the design doc for details, including the §6 WONT-FIX accepted gaps).

PR #51534, opened 2026-07-17
enhance: rebuild the proxy meta cache around an id-primary store with O(1) invalidation

xiaofan-luan:
issue: #51533

**Based on #51313** — this branch stacks on it; the first commits are #51313's and will disappear from the diff once it merges (this PR will be rebased then).

Design doc: [`docs/design-docs/design_docs/20260716-proxy-metacache-id-inverted-index.md`](https://github.com/xiaofan-luan/milvus/blob/enhance/proxy-metacache-id-primary/docs/design-docs/design_docs/20260716-proxy-metacache-id-inverted-index.md) (included in this PR).

## Problem

Every id-, alias- or database-driven invalidation in the proxy meta cache was a linear scan under the global write lock — O(all cached collections) to find an id's name keys, O(all cached partition entries) for the nested partition sweep, O(db's aliases) for alias cleanup — on **every** Load/Release/Drop/Rename/Alter broadcast. At a 1M-collection cached working set each broadcast stalls every query on the proxy. Separately, cache fills raced invalidation broadcasts on two unordered channels, guarded by a per-collection timestamp floor with known gaps (unversioned alias write-backs, the `RequestTime==0` compat bypass, `RemoveDatabase` never raising it).

## Design

**P1 — id-primary store.** `collections map[id]*entry` is the single source of truth; an entry is live iff present AND its `dbGen` matches the database's current generation. Name/alias resolution become **hints** validated against the primary on every read — a stale hint costs one extra describe, never a stale read — and are reclaimed lazily by the background GC. Partition caches are keyed by the collection id (ids are never reused, so drop+recreate cannot alias into old entries). Every invalidation becomes **O(1)**: by-id = one primary delete; partition DDL = exact id keys; Drop/AlterDatabase = one generation bump. Rename needs no eager cleanup (the old-name hint fails validation), and an old-rootcoord id-describe with no DbName becomes cacheable while preserving the cross-database mis-hit protection (no name hint).

**P2 — pessimistic fill ordering.** A global `fillMu` RWMutex: fills (describe RPC + write-back) hold the read side for their whole duration; invalidations take the write side, draining in-flight fills before evicting — a pre-DDL snapshot always lands *before* the eviction that cleans it. Cache hits touch neither `fillMu` nor any invalidation state. This replaces the `collectionCacheVersion` floor entirely and closes the `RemoveDatabase` resurrection race, the unversioned alias write-back race, and the `RequestTime==0` bypass. Deliberate relaxation: duplicate/late broadcasts now over-invalidate (one extra re-describe) instead of being version-filtered.

**P3 — alias DDL invalidation info comes from the source.** rootcoord resolves the alias's pre-alter target under its database lock and forwards it in the new `AlterAliasMessageHeader.old_collection_id` (additive field, computed at broadcast time so replay stays deterministic); the ack callback emits a second expiration entry, so proxies evict BOTH AlterAlias targets by id — immune to the race where a concurrent describe re-points the proxy's alias resolution before the expiration arrives. Older rootcoords (field 0) fall back to the proxy's hint resolution (same rolling-upgrade caveat as the existing forwarded new-target id).

Also fixes a latent stale read: partition entries cached under an alias name survived DropCollection and could serve stale partition ids after a same-name recreate — id-keying removes the class.

## Verification

- Targeted cache suites green: rename, alias re-point under concurrent describe, drop+recreate under a reused name, database generation invalidation, cross-db isolation, fill/invalidation ordering (gated-mock concurrency tests), rootcoord `TestDDLCallbacksAliasDDL`.
- Three negative controls verified the tests bite: disabling hint name-validation, the dbGen check, or the fill drain each fails exactly the test guarding it.
- Full-package `-race` run delegated to CI.

## Rolling upgrade

Proxy changes are in-memory and proxy-local. The one wire change is the additive `old_collection_id` field; older rootcoords leave it 0 with the documented fallback ("upgrade rootcoord before proxies; avoid alias DDL during mixed deployment" — same caveat as the existing new-target id forwarding).


xiaofan-luan:
Folding these commits into #51313 (and #51369 for 3.0) per the author's decision — this branch stacked on #51313 anyway. Tracking issue stays: #51533.

PR #51536, opened 2026-07-17
fix: [3.0] skip raw vector fielddata on storage-v3 load when index has raw data (#51541)

xiaofan-luan:
pr: #51541
issue: #51299

Backport of #51541 to the 3.0 branch.

## Problem

storage-v3 (loon/vortex) sealed segments load the raw vector column resident **on top of** a vector index that already carries the raw data (`HasRawData=true`), roughly **doubling** the per-segment disk footprint of DiskANN / IVF_FLAT / HNSW(with-raw) segments. On the 100M / 128-dim DiskANN benchmark this pushes a querynode over its 100Gi ephemeral-storage cap and `load_collection` fails with `segment request resource failed[resourceType=Disk]`.

## Fix

In `ComputeDiffColumnGroups`, when a field's index carries raw data and `prefer_field_data_when_index_has_raw_data == false`, drop the raw column from the load set entirely (mirroring the v2 binlog path) instead of lazy-loading it. Covers vector fields and other indexed scalars (retrieve/filter read the raw value from the index).

The **primary key is excluded**: sorted segments navigate rows through the pk column (`num_rows_until_chunk` / `get_chunk_by_offset`) and a pk scalar index is not registered in `scalar_indexings`, so the pk raw column must stay resident (otherwise a `ColumnExpr` on the pk falls back to a column scan and crashes — `TestCreateSortedScalarIndex`).

Identical to the master fix in #51541 (cherry-picked clean; clang-format-15 verified).

### DIFF
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -215,8 +215,17 @@ ResolveHybridInternalIndexType(
 bool
 IndexFactory::CanUseIndexRawDataForField(DataType field_type,
                                          bool has_raw_data) {
+    // ARRAY and JSON indexes only index a projection of the value (array
+    // elements / a JSON path) and cannot reconstruct the whole raw value, so
+    // their index is never a stand-in for the raw column. This mirrors the
+    // segment runtime contract (HasRawDataFromState returns column-based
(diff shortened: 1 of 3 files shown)


liliu-z (commented):
/lgtm

Issue #51537, opened 2026-07-17
[Bug]: ProcessDataByOffsets skip-index branches desync cursor-tracking callbacks, silently dropping matches under iterative filtering


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

- Milvus version: master (c4fa9ea356)

### Current Behavior

The offset-input evaluation path \`SegmentExpr::ProcessDataByOffsets\` (internal/core/src/exec/expression/Expr.h) has six skip-index branches (chunked-sealed view/scalar per-row loops, non-chunked sealed whole batch, growing loop, VectorArrayView loop, \`ProcessDataByOffsetsForSealedSeg\`) that apply the valid mask and skip the row **without invoking the callback**.

Cursor-tracking callbacks (UnaryExpr / TermExpr / BinaryRange / BinaryArith ForData lambdas) maintain a \`processed_cursor\` that indexes \`bitmap_input\` (the already-evaluated left operand of an AND conjunct) **by batch position**. A skipped row that never reaches the callback fails to advance that cursor, so every later candidate in the batch reads \`bitmap_input\` at a too-low index. A row whose conjunct bit is 1 can read a 0 bit, its evaluation is skipped, and a genuinely matching row is silently dropped from the results — **row-level false negatives under \`A and B\` with iterative-filter offsets whenever the skip index prunes some candidates' chunks**.

The sequential path already handles this correctly: \`ProcessDataChunksForMultipleChunk\`'s skip branch calls \`func(nullptr, ...)\` with an explicit comment that this keeps \`processed_cursor\` aligned. The offsets path never adopted that convention.

### Expected Behavior

Every candidate offset reaches the callback exactly once — skipped rows as a null batch (\`data == nullptr\`) — so cursor-tracking callbacks stay aligned with \`bitmap_input\`, matching the sequential path's established contract.

### Steps To Reproduce

Regression tests demonstrating the desync (red on master, green with the fix) are included in the fix PR: \`OffsetsEvalCorrectnessTest.SkipBranchDrivesCallbackPerCandidate\` and \`SkipBranchKeepsBitmapCursorAligned\`.

### Anything else?

Found by an adversarial multi-agent review during #51140 (unrelated to that PR's changes; pre-existing on master). Fix PR incoming.

PR #51540, opened 2026-07-17
fix: keep bitmap cursors aligned on skipped offset paths

xiaofan-luan:
issue: #51537

## What changed

- Preserve the evaluator callback contract when SkipIndex prunes rows or array elements: apply the validity mask, then invoke the evaluator with `data == nullptr` so callback-local bitmap cursors advance by every logical candidate.
- Cover row offsets, element offsets, and element-level full scans for both growing and sealed segments.
- Treat nullable scalar-index reverse-lookup misses as null callback batches so later offset candidates remain aligned.
- Audit every production offset evaluator and make the remaining TIMESTAMPTZ, unary ARRAY, and unary JSON evaluators null-batch safe.
- Do not apply chunk-level SkipIndex to scalar-index-only offset lookup. The scalar index is segment-wide, while SkipIndex statistics follow raw-data chunks whose boundaries are unavailable after field data is dropped.
- Use `IArrayOffsets` as the source of truth for the flattened logical ARRAY element space. Nullable rows contribute zero logical elements even if storage retains a physical payload.
- Fix growing-segment ARRAY length recovery to read the actual dense `Array` layout and to assign zero elements to invalid rows.
- Remove unreachable non-chunked sealed helpers and the remaining sealed size branch; sealed storage is always chunked.

## Root cause

Offset and element evaluators process a candidate list while expression callbacks index `bitmap_input` with a callback-local `processed_cursor`. Several SkipIndex branches updated result validity but bypassed the callback. After a skipped candidate, later candidates read an earlier conjunct bit and could produce false negatives or false positives.

The scalar-index-only path had a second issue: it treated the single scalar-index chunk as raw chunk 0 and could apply chunk-0 SkipIndex to the entire candidate batch. Once raw field data is dropped, that mapping cannot be reconstructed safely, so candidates must be evaluated individually through reverse lookup.

For nullable ARRAY fields, result allocation follows the logical element count in `IArrayOffsets`, but the element full-scan path previously advanced by physical payload length even for invalid rows. A NULL row can retain payload bytes while contributing zero logical elements, so physical counting could desynchronize the evaluator and write beyond the logical result range.

## Regression coverage

Fifteen focused tests cover:

1. growing fixed-width row-offset callback count
2. growing fixed-width row-offset bitmap alignment
3. growing element-offset alignment
4. sealed fixed-width row-offset alignment
5. a real `ExprSet` / conjunct regression with two binlog chunks and real SkipIndex pruning
6. sealed view (`std::string_view`) row-offset alignment
7. `VectorArrayView` row-offset alignment
8. sealed element-offset alignment
9. growing element full-scan alignment
10. sealed element full-scan alignment
11. growing nullable ARRAY logical element counting
12. growing loaded-field nullable ARRAY offset recovery
13. sealed nullable ARRAY logical element counting
14. nullable scalar-index-only reverse lookup and ignored chunk-level SkipIndex
15. nullable TIMESTAMPTZ scalar-index-only offset evaluation through a real physical expression

## Validation

- clang-format 15 dry-run passed for all changed C++ files.
- `git diff --check` passed.
- Focused cursor/nullable/index-only regression suites passed: 15/15.
- Related ElementFilter and TIMESTAMPTZ suites passed: 71/71.
- The branch is rebased onto the current master; CI is rerunning on the updated head.

## Scope

This PR fixes logical-candidate accounting when a batch contains elements. The existing framework behavior for a whole row batch whose logical element count is zero (all rows are NULL or empty arrays) is not changed here and should be handled separately.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -614,6 +614,16 @@ class SegmentExpr : public Expr {
                                               std::move(valid_res));
     }
 
+    // Candidate evaluator contract:
+    // - every logical candidate position advances the evaluator exactly once;
+    // - data == nullptr means the reader already decided the candidate result
+    //   (for example SkipIndex pruning or a NULL reverse lookup), so the
+    //   evaluator must only advance position-indexed state by size and return;
+    // - result and validity for a null batch are owned by the reader.
+    //
+    // This contract keeps callback-local bitmap_input cursors aligned across
+    // raw-data, element-level, and scalar-index-only candidate paths.
+
     // when we have scalar index and index contains raw data, could go with index chunk by offsets
     template <typename T, typename BatchEvaluator, typename... ValTypes>
     int64_t
@@ -1063,6 +1073,11 @@ class SegmentExpr : public Expr {
                       "Json element type is not supported for "
                       "element-level filtering");
 
+        auto array_offsets = segment_->GetArrayOffsets(field_id_);
+        AssertInfo(array_offsets != nullptr,
+                   "ArrayOffsets not found for field {}",
+                   field_id_.get());
+
         int64_t processed_rows = 0;
         int64_t processed_elems = 0;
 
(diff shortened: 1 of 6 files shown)


xiaofan-luan:
rerun go-sdk

czs007:
Adversarial verification confirmed two documentation/tradeoff items on the index-only evaluation path; both concern accuracy of the PR narrative rather than defects in the shipped code, and neither blocks merge.

## 🟡 Low

- **`internal/core/src/exec/expression/Expr.h:633`** — The index-only path (`ProcessIndexLookupByOffsets`) drops the previous batch fast-path (`res.set(0, batch_size)` / `valid_res.set(0, batch_size)` under `skip_func`-pruned + `all_valid`) and now unconditionally does a per-offset `Reverse_Lookup` + individual evaluate. This is not a regression from a valid optimization: the removed fast-path was itself incorrect, since scalar indexes are a single segment-level logical chunk while SkipIndex statistics are partitioned by the original data-chunk, so applying chunk-0's SkipIndex to the whole candidate batch mispredicts a full-batch hit. The non-skip path was already per-offset `Reverse_Lookup` in both old and new code, so the only real wall-clock cost lands in the (incorrect) "SkipIndex could short-circuit the whole batch" case. **Suggestion:** no code change needed — the tradeoff is already recorded in the PR body's Root cause / Scope sections; consider downgrading this to a nitpick. *(raised by tinswzy)*

## 🔵 Nitpick

- **`internal/core/src/exec/expression/Expr.h`** (PR description) — The PR body's "What changed" section groups the three new `data == nullptr` guards (TIMESTAMPTZ, unary ARRAY, unary JSON) under the "when SkipIndex prunes rows or array elements" context. That attribution is imprecise: all three call `ProcessDataByOffsets` with `skip_func = std::nullptr_t{}`, and `if (!skip_func || !skip_func(...))` is always true for a null `skip_func`, so the SkipIndex-prune branch can never fire for them. Their only source of a null batch is the index-only `ProcessIndexLookupByOffsets` on a reverse-lookup miss (`raw.has_value() == false`). The guards themselves are correct and complete. **Suggestion:** reword the PR body to distinguish these three guards from the SkipIndex-prune scenario and attribute them to the reverse-lookup-miss path. *(raised by tinswzy)*

liliu-z (commented):
/lgtm
/approve

PR #51541, opened 2026-07-17
fix: skip raw vector fielddata on storage-v3 load when index has raw data

xiaofan-luan:
issue: #51299

## Problem

storage-v3 (loon/vortex) sealed segments load the raw vector column resident **on top of** a vector index that already carries the raw data (`HasRawData=true`), roughly **doubling** the per-segment disk footprint of DiskANN / IVF_FLAT / HNSW(with-raw) segments. On the 100M / 128-dim DiskANN benchmark (1 replica, 100Gi querynode ephemeral-storage) this pushes a querynode over the cap and `load_collection` fails with `segment request resource failed[resourceType=Disk]`.

The v2 binlog path already skips the raw fielddata when the index has raw data (`ChunkedSegmentSealedImpl`: *"skip fielddata because index has raw data"*); the v3 manifest path in `SegmentLoadInfo::ComputeDiffColumnGroups` only marked the column **lazy**, and `LoadColumnGroup(eager_load=false)` still materialized a `ProxyChunkColumn` and set `field_data_ready`, so the raw column stayed resident anyway.

## Fix

In `ComputeDiffColumnGroups`, when a field's index carries raw data and `prefer_field_data_when_index_has_raw_data == false`, drop the raw column from the load set entirely (mirroring the v2 binlog path) instead of lazy-loading it. This covers vector fields and other indexed scalars (retrieve/filter read the raw value from the index).

The **primary key is excluded**: sorted segments navigate rows through the pk column (`num_rows_until_chunk` / `get_chunk_by_offset`) and a pk scalar index is not registered in `scalar_indexings`, so the pk raw column must stay resident — otherwise a `ColumnExpr` on the pk falls back to a column scan (`use_index_data_ == false`) and crashes (`field 100 must exist when getting rows until chunk`, regressed go-sdk `TestCreateSortedScalarIndex`).

The stale-copy drop is scoped to the real no-raw-index -> raw-index reopen transition. `prefer_field_data` remains the explicit opt-in to keep both resident.

### DIFF
--- a/internal/core/src/segcore/SegmentLoadInfo.cpp
+++ b/internal/core/src/segcore/SegmentLoadInfo.cpp
@@ -471,6 +471,12 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
     auto prefer_field_data =
         SegcoreConfig::default_config()
             .get_prefer_field_data_when_index_has_raw_data();
+    // The primary key column is never skipped, even when its index carries raw
+    // data: sorted segments navigate rows through the pk column
+    // (num_rows_until_chunk / get_chunk_by_offset), and a pk scalar index is not
+    // registered in scalar_indexings for expression index-reads, so the pk raw
+    // column must stay resident.
+    auto pk_field_id = new_info.schema_->get_primary_field_id();
     auto cur_column_group = GetColumnGroups();
     auto new_column_group = new_info.GetColumnGroups();
 
@@ -548,16 +554,22 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
             bool is_replace_field = was_default_filled || files_changed;
             bool index_has_raw_data =
                 new_info.field_index_has_raw_data_.count(fid) > 0;
-            // When the vector index already carries the raw data, the separate
+            bool is_pk_field =
+                pk_field_id.has_value() && pk_field_id.value() == fid;
+            // When a field's index already carries the raw data, the separate
             // raw column is redundant and must NOT be made resident — neither
             // eager nor "lazy" (the loon lazy path still materializes a
             // ProxyChunkColumn and sets field_data_ready, so it stays on disk on
             // top of the index, doubling the footprint). Mirror the binlog path
             // (ChunkedSegmentSealedImpl: "skip fielddata because index has raw
-            // data"): drop it from the load set entirely. prefer_field_data is
-            // the explicit opt-in to keep both resident; system fields always load.
-            if (field_id >= START_USER_FIELDID && index_has_raw_data &&
-                !prefer_field_data) {
+            // data"): drop it from the load set entirely. This covers vector
+            // fields and other indexed scalars (retrieve/filter read the raw
+            // value from the index), but NOT the primary key — its raw column
+            // is still needed resident for row navigation, so the pk always
+            // loads regardless of its index. prefer_field_data keeps both
+            // resident; system fields always load.
+            if (field_id >= START_USER_FIELDID && !is_pk_field &&
+                index_has_raw_data && !prefer_field_data) {
                 // Only the no-raw-index -> raw-index reopen transition can
                 // leave a stale resident copy: current had no raw-data index,
                 // so the raw column was loaded, and now the index carries it.
--- a/internal/core/src/segcore/SegmentLoadInfo.cpp
+++ b/internal/core/src/segcore/SegmentLoadInfo.cpp
@@ -570,14 +570,21 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
             // resident; system fields always load.
             if (field_id >= START_USER_FIELDID && !is_pk_field &&
                 index_has_raw_data && !prefer_field_data) {
-                // Only the no-raw-index -> raw-index reopen transition can
-                // leave a stale resident copy: current had no raw-data index,
-                // so the raw column was loaded, and now the index carries it.
-                // Drop it in that case. In steady state the column was already
-                // skipped (never resident), so there is nothing to free and no
-                // diff to emit.
-                if (cur_iter != cur_field_to_files.end() &&
-                    field_index_has_raw_data_.count(fid) == 0) {
+                // Free any stale resident copy of the raw column before the
+                // early continue (which bypasses the replace path below):
+                //  - default-filled: FillDefaultValueFields put a resident
+                //    column in runtime that is not in the current manifest
+                //    (cur_iter == end); the new manifest now backs it with a
+                //    raw-data index, so the default column must be dropped or it
+                //    leaks (and the default-filled flag is cleared on reopen, so
+                //    no later pass can clean it).
+                //  - no-raw-index -> raw-index transition: the raw column was
+                //    resident because the current index had no raw data; drop it
+                //    now that the index carries it.
+                // Steady state (already skipped, never resident) drops nothing.
+                if (was_default_filled ||
+                    (cur_iter != cur_field_to_files.end() &&
+                     field_index_has_raw_data_.count(fid) == 0)) {
                     diff.field_data_to_drop.emplace(field_id);
                 }
                 continue;
--- a/internal/core/src/segcore/SegmentLoadInfo.cpp
+++ b/internal/core/src/segcore/SegmentLoadInfo.cpp
@@ -548,6 +548,28 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
             bool is_replace_field = was_default_filled || files_changed;
             bool index_has_raw_data =
                 new_info.field_index_has_raw_data_.count(fid) > 0;
+            // When the vector index already carries the raw data, the separate
+            // raw column is redundant and must NOT be made resident — neither
+            // eager nor "lazy" (the loon lazy path still materializes a
+            // ProxyChunkColumn and sets field_data_ready, so it stays on disk on
+            // top of the index, doubling the footprint). Mirror the binlog path
+            // (ChunkedSegmentSealedImpl: "skip fielddata because index has raw
+            // data"): drop it from the load set entirely. prefer_field_data is
+            // the explicit opt-in to keep both resident; system fields always load.
+            if (field_id >= START_USER_FIELDID && index_has_raw_data &&
+                !prefer_field_data) {
+                // Only the no-raw-index -> raw-index reopen transition can
+                // leave a stale resident copy: current had no raw-data index,
+                // so the raw column was loaded, and now the index carries it.
+                // Drop it in that case. In steady state the column was already
+                // skipped (never resident), so there is nothing to free and no
+                // diff to emit.
+                if (cur_iter != cur_field_to_files.end() &&
+                    field_index_has_raw_data_.count(fid) == 0) {
+                    diff.field_data_to_drop.emplace(field_id);
+                }
+                continue;
+            }
             // Eager-load when: system field, OR schema says load AND
             // (we want to keep field data alongside index, OR index can't serve raw).
             bool should_eager_load =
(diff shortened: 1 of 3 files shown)


internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -548,6 +554,34 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
...
+            // ProxyChunkColumn and sets field_data_ready, so it stays on disk on
+            // top of the index, doubling the footprint). Mirror the binlog path
+            // (ChunkedSegmentSealedImpl: "skip fielddata because index has raw
+            // data"): drop it from the load set entirely. This covers vector
+            // fields and other indexed scalars (retrieve/filter read the raw
+            // value from the index), but NOT the primary key — its raw column
+            // is still needed resident for row navigation, so the pk always
+            // loads regardless of its index. prefer_field_data keeps both
+            // resident; system fields always load.
+            if (field_id >= START_USER_FIELDID && !is_pk_field &&
```
liliu-z: Can we keep this vector-only for now? `field_index_has_raw_data_` is also true for JSON + STL_SORT, but that index only contains one JSON path. With this branch, the full JSON column is never created, so full/dynamic JSON output will fall back to `get_raw_data()` and fail.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -548,6 +554,34 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
...
+                // leave a stale resident copy: current had no raw-data index,
+                // so the raw column was loaded, and now the index carries it.
+                // Drop it in that case. In steady state the column was already
+                // skipped (never resident), so there is nothing to free and no
+                // diff to emit.
+                if (cur_iter != cur_field_to_files.end() &&
+                    field_index_has_raw_data_.count(fid) == 0) {
+                    diff.field_data_to_drop.emplace(field_id);
+                }
+                continue;
```
liliu-z, line 575: Skipping scalar columns also breaks some nullable paths. Text index rebuild calls `raw.value()` after `AddNullSealed()` without a `continue`, and search group-by assumes `Reverse_Lookup()` is never null. These worked before because the lazy column was still available. Either keep this vector-only or fix and test those paths here.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -548,6 +554,34 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
...
+            }
```
liliu-z, line 576: Can we add a focused Storage V3 test for this? At least verify that a real raw-data vector index loads without a field column and retrieve still returns the correct vectors. The PK regression from the first revision shows this path is not covered today.

internal/core/src/segcore/SegmentLoadInfo.cpp
liliu-z: @zhengbuqian

liliu-z: Fixed in `635ee413`: JSON is now excluded by `CanUseIndexRawDataForField`, which matches the runtime contract. Thanks.

internal/core/src/segcore/SegmentLoadInfo.cpp
liliu-z: Still blocking on the latest head. The skip still applies to raw-data scalar indexes, but there is no fix or test for nullable text-index rebuild or nullable search group-by. The new INT64 diff test does not cover either consumer.

internal/core/src/segcore/SegmentLoadInfo.cpp
liliu-z: Thanks for adding the diff tests. They cover the LoadDiff classification, but they do not run a real manifest, `LoadColumnGroup`, a raw-data vector index, or retrieve. I would still like one production-path V3 vector test here.

internal/core/src/segcore/SegmentLoadInfo.cpp
liliu-z: Index-only offset filtering also loses NULL validity. `ProcessIndexLookupByOffsets()` sets `res=false` for `nullopt` but leaves `valid_res=true`, so `NOT (...)` can incorrectly include NULL rows. Please clear `valid_res` too, or keep scalar columns.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -548,6 +554,34 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
...
+                index_has_raw_data && !prefer_field_data) {
+                // Only the no-raw-index -> raw-index reopen transition can
+                // leave a stale resident copy: current had no raw-data index,
+                // so the raw column was loaded, and now the index carries it.
+                // Drop it in that case. In steady state the column was already
+                // skipped (never resident), so there is nothing to free and no
+                // diff to emit.
+                if (cur_iter != cur_field_to_files.end() &&
+                    field_index_has_raw_data_.count(fid) == 0) {
+                    diff.field_data_to_drop.emplace(field_id);
```
liliu-z, line 573: This drop can race with an in-flight retrieve. `bulk_subscript()` reads source selection, index, and column from different published snapshots. A query can choose the old column path, then reopen publishes the raw index without a column, and the query fails on the new state. Please use one snapshot for selection and access, or otherwise keep the old source alive for readers.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -548,6 +554,34 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
...
+            bool is_pk_field =
+                pk_field_id.has_value() && pk_field_id.value() == fid;
+            // When a field's index already carries the raw data, the separate
+            // raw column is redundant and must NOT be made resident — neither
+            // eager nor "lazy" (the loon lazy path still materializes a
+            // ProxyChunkColumn and sets field_data_ready, so it stays on disk on
+            // top of the index, doubling the footprint). Mirror the binlog path
+            // (ChunkedSegmentSealedImpl: "skip fielddata because index has raw
+            // data"): drop it from the load set entirely. This covers vector
+            // fields and other indexed scalars (retrieve/filter read the raw
```
liliu-z: This is not true for every scalar path. `TIMESTAMPTZ +/- INTERVAL` is still column-scan-only. With an STL_SORT index, this branch removes the column, leaves `num_data_chunk_ == 0`, and the filter fails at its processed-size assertion. Please keep this vector-only for now, or keep TIMESTAMPTZ field data.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -548,6 +554,34 @@ SegmentLoadInfo::ComputeDiffColumnGroups(LoadDiff& diff,
...
+            // value from the index), but NOT the primary key — its raw column
+            // is still needed resident for row navigation, so the pk always
+            // loads regardless of its index. prefer_field_data keeps both
+            // resident; system fields always load.
+            if (field_id >= START_USER_FIELDID && !is_pk_field &&
+                index_has_raw_data && !prefer_field_data) {
```
liliu-z, line 566: Nullable search group-by still breaks when this skips the scalar column. `Reverse_Lookup()` returns `nullopt` for a NULL row, but `SealedDataGetter::Get()` asserts `raw.has_value()`. Please return the optional directly, or keep scalar columns.

internal/core/src/segcore/SegmentLoadInfo.cpp
liliu-z, line 575: Nullable text-index rebuild has the same exposure. After `AddNullSealed(i)`, it falls through and calls `raw.value()`, so the first NULL throws once the VARCHAR column is skipped. Please add `continue` there, or keep scalar columns.

internal/core/src/segcore/SegmentLoadInfo.cpp
xiaofan-luan: This exact path is fixed in #51597: `SealedDataGetter::Get()` no longer asserts `raw.has_value()` on the `from_data_ == false` (Reverse_Lookup) branch — a NULL row returns `std::nullopt` and forms a distinct null group. It ships with a group-by null regression test (test_search_group_by.cpp).

internal/core/src/segcore/SegmentLoadInfo.cpp
xiaofan-luan: Fixed in #51595: `ProcessIndexLookupByOffsets` now clears validity for a `Reverse_Lookup == nullopt` row (`res[i] = valid_res[i] = false`) instead of leaving `valid_res` true, so NOT no longer includes NULL rows. Covered by test_offsets_eval_correctness / ExprArithOpTest.

internal/core/src/segcore/SegmentLoadInfo.cpp
xiaofan-luan: Filed as #51599 (assigned @zhengbuqian): add the missing `continue` after `AddNullSealed(i)` in the text-index rebuild + a nullable regression test. Tracking separately.

internal/core/src/segcore/SegmentLoadInfo.cpp
xiaofan-luan: Filed as #51600 (assigned @yanliang567): `PhyTimestamptzArithCompareExpr` is column-scan-only, so it needs either an index-reverse-lookup eval path or the column kept resident. Tracking separately.

internal/core/src/segcore/SegmentLoadInfo.cpp
xiaofan-luan: This is exactly the per-operation published-state pin work: zilliztech/milvus-common#106 (merged) adds the OpContext carrier, and #51594 pins one immutable snapshot per operation so bulk_subscript's selection/index/column accessors all read a consistent state across a concurrent Reopen. Tracked there.

internal/core/src/segcore/SegmentLoadInfo.cpp
xiaofan-luan: Will add a production-path V3 vector test: a float-vector segment with a real IVF_FLAT index, loaded through SegmentLoadInfo -> SetLoadInfo -> Load (real loon manifest + LoadColumnGroup + LoadBatchIndexes, no force-load helper), asserting num_chunk_data(vec)==0 / HasFieldData(vec)==false and that get_vector returns the correct vectors reconstructed from the index.

xiaofan-luan: Update: @wangting has already run the storage-v3 load tests and verified the raw-data vector index loads without a resident field column and retrieve returns correct vectors. We are also narrowing this PR to vector fields only (see summary), so scalar paths are untouched here.

xiaofan-luan:
## Summary — scope narrowed to vector fields only

Per @liliu-z's review, this PR now skips the raw column **only for vector fields** (`c8f69e7`). A vector index (IVF_FLAT / DiskANN / HNSW-with-raw) fully reconstructs the raw vector at retrieve (`get_vector`), so its separate raw column is redundant — and that is exactly where the ~47 GiB / doubled per-segment footprint (100M×128 DiskANN load OOM) comes from.

**Scalar / JSON / ARRAY / primary-key raw columns are no longer skipped** — they stay resident as before. This makes the change safe on its own and removes the per-type carve-outs (pk exclusion, default-filled drop) since a vector field is never the pk and has no default value.

### Why scalars are deferred

Skipping a scalar raw column forces every read onto the index path, which today does not cover all consumers:
- nullable `Reverse_Lookup` returns `nullopt` → asserts/throws in group-by, offset-filtering, and text-index rebuild;
- some expressions are column-scan-only (`TIMESTAMPTZ +/- INTERVAL`);
- `bulk_subscript` reads source-selection / index / column from different published snapshots (TOCTOU with a concurrent Reopen drop).

### Scalar raw-column skip — series of follow-ups

Extending the skip to indexed scalars needs these to land first:

| area | tracking |
|---|---|
| group-by NULL on index-only nullable scalar | #51597 |
| offset-filtering 3VL NULL validity (`ProcessIndexLookupByOffsets`) | #51595 |
| retrieve snapshot race — per-operation published-state pin | #51594 + zilliztech/milvus-common#106 |
| text-index rebuild NULL (missing `continue` after `AddNullSealed`) | #51599 |
| `TIMESTAMPTZ +/- INTERVAL` column-scan-only | #51600 |
| JSON exists NULL across raw/index/stats | #51596 |
| `BulkIsValid` non-nullable fall-through | #51598 |

Once these are in, a follow-up can flip the gate from vector-only back to "vector + indexed scalar".

### Verification
- Vector skip + retrieve-from-index correctness: storage-v3 e2e / go-sdk (run by @wangting).
- Unit: LoadDiff classification tests — scalars (incl. pk) with a raw-data index are NOT skipped; `CanUseIndexRawDataForField` excludes JSON/ARRAY.

3.0 backport: #51536 (identical).

liliu-z (commented):
/lgtm
/approve

Issue #51497, opened 2026-07-17
[Bug]: /management/config/get returns stale mq.type (RuntimeSource) after WAL switch completes — startup runtime overlay permanently shadows the etcd-persisted value


tinswzy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```
- Milvus version: master (after #45286, WAL switching)
- Deployment mode(standalone or cluster): cluster
- MQ type(rocksmq, pulsar or kafka): pulsar -> woodpecker (WAL switch)
- SDK version(e.g. pymilvus v2.0.0rc2): N/A (management HTTP API)
- OS(Ubuntu or CentOS): any
- CPU/Memory: any
- GPU: N/A
- Others: N/A
```

### Current Behavior

After a successful WAL switch via `POST /management/wal/alter` (e.g. pulsar → woodpecker):

- all vchannels ACK the AlterWAL broadcast,
- the ack callback writes `mq.type=woodpecker` into etcd (`alterWALV2AckCallback` → `UpdateConfigInEtcd`, `internal/coordinator/wal_callbacks.go`),
- the WAL is reopened with the target implementation.

But `GET /management/config/get?keys=mq.type` keeps returning the **old** value until the coordinator process restarts:

```json
{"configs":[{"key":"mq.type","value":"pulsar","source":"RuntimeSource"}]}
```

An external operator that polls this endpoint to confirm switch completion never observes the new value (a real control-plane workflow exhausted 360/360 retries against a cluster whose switch had actually completed successfully).

#### Root cause

1. At startup the coordinator calls `InitAndSelectWALName()` (`cmd/roles/roles.go`), which saves the **resolved** WAL name into the runtime overlay: `paramtable.Get().Save("mq.type", walName)` → `Manager.SetConfig` → `m.overlays` (`internal/util/streamingutil/util/wal_selector.go`, `pkg/config/manager.go`).
2. `Manager.GetConfig()` checks `m.overlays` **first** and returns `(RuntimeSource, value)` unconditionally — the EtcdSource value is permanently shadowed (`pkg/config/manager.go`).
3. The AlterWAL ack callback only updates etcd (and linearizably refreshes the local EtcdSource), but never updates/clears the runtime overlay. So the overlay keeps the pre-switch name until restart.

Note: this is **not** related to `Immutable: true` or the `refreshable` tag. The etcd write and the in-process EtcdSource refresh both succeed; only the startup-time runtime overlay shadows them on the read path.

[...]

PR #36822, opened 2024-10-12
feat: management WebUI homepage

jaime0815:
issue: #36621
1. Implement an embedded web server for WebUI access.  
2. Complete the homepage development.

Home page demo:
iShot_2024-10-10_17 57 34

PR #51543, opened 2026-07-17
fix: stop runtime overlay from shadowing etcd-persisted mq.type after WAL switch

tinswzy:
issue: #51497

After a successful WAL switch via `POST /management/wal/alter`, `GET /management/config/get?keys=mq.type` kept returning the pre-switch value with `"source":"RuntimeSource"` until the coordinator restarted, even though the ack callback had already persisted the new mq.type into etcd (and linearizably refreshed the local EtcdSource). External operators polling that endpoint to confirm switch completion never saw the new value.

### Root cause

`InitAndSelectWALName()` saved the startup-resolved WAL name into the config manager's **runtime overlay**, and `Manager.GetConfig` serves the overlay with top priority — permanently shadowing the EtcdSource for every reader in the process. The overlay write existed only to feed `ProcessImmutableConfigs`, which pins `GetConfig`'s value into etcd create-if-absent at first boot (without it, the literal `"default"` would be persisted).

### Fix

Replace the overlay bridge with an explicit per-key **renderer** on the persistence path, so mq.type reads fall through to the etcd source and naturally follow a WAL switch:

- `ProcessImmutableConfigs(renderers map[string]func(raw string) string)`: a renderer converts a placeholder raw value (mq.type's literal `"default"`) into the concrete value at create-if-absent persist time. It never runs when etcd already holds the key (a completed switch survives restart), and it also covers the key-absent-from-all-sources case (raw is passed as `""`). Generic: any future immutable config with a placeholder value can plug in its own renderer.
- `InitAndSelectWALName()` no longer writes the runtime overlay and returns the resolved `message.WALName`; the coordinator startup passes a mq.type renderer backed by that name.
- `dependency.HealthCheck` resolves the literal `"default"` to the actually-enabled MQ before probing (this gap predates the overlay Save: since #36822, deployments running with `mq.type: default` probed nothing and always reported unhealthy).

This also fixes two other readers of the shadowed value:
- the `HandleAlterWAL` same-type short-circuit: re-posting the completed target no longer re-broadcasts, and a rollback to the original MQ is no longer silently swallowed with "already configured";
- the proxy MQ health check (`/_cluster/dependencies`) no longer keeps probing the old MQ forever after a switch.

### Behavior note

`MQCfg.Type.GetValue()` now follows etcd at runtime instead of being frozen at process start: the coordinator sees a switch immediately (linearizable refresh in the ack callback); other nodes converge within one EtcdSource poll cycle (default `refreshInterval` 5s). Audited readers: `mustSelectMQType` / `MustSelectWALName` resolve `"default"` themselves and are startup-scoped; the AlterWAL idempotency check and the proxy health check are exactly the readers this change fixes.

### Verification

- New tests (all watched failing first): renderer converts placeholder before first persist / existing etcd value never overwritten nor re-rendered / no-renderer keys unchanged / key absent from all sources still pinned via renderer (`pkg/config`); `InitAndSelectWALName` leaves mq.type served from its original source, not `RuntimeSource` (`streamingutil/util`); `HealthCheck("default")` resolves to the enabled MQ (`dependency`).
- Full package runs green locally: `pkg/config`, `internal/util/streamingutil/util`, `internal/util/dependency` (with `-tags dynamic,test -gcflags="all=-N -l"`).
- `cmd/roles`, `internal/coordinator`, `internal/distributed/streaming` type-checked via `go vet` locally (no current-master segcore build on this machine); relying on CI for full compilation. A live-cluster alter → config/get end-to-end pass is still pending — the switch-visibility chain (etcd write → `GetConfig` returns the new value) is covered at unit level by the existing `TestAlterConfigsInEtcd`.

### DIFF
--- a/cmd/roles/roles.go
+++ b/cmd/roles/roles.go
@@ -416,10 +416,14 @@ func (mr *MilvusRoles) Run() {
 
 	// Persist immutable configurations at startup, such as mqType paramItem
 	if (mr.EnableRootCoord && mr.EnableDataCoord && mr.EnableQueryCoord) || mr.EnableMixCoord {
-		// Initialize the actual walName instead of default
-		util.InitAndSelectWALName()
-		// persist immutable configs if necessary
-		if err := paramtable.GetBaseTable().Manager().ProcessImmutableConfigs(); err != nil {
+		// Resolve the actual walName instead of default
+		walName := util.InitAndSelectWALName()
+		// persist immutable configs if necessary; mq.type's literal "default" is
+		// rendered to the resolved walName so the value pinned in etcd is always
+		// a concrete WAL name (issue #51497)
+		if err := paramtable.GetBaseTable().Manager().ProcessImmutableConfigs(map[string]func(string) string{
+			paramtable.Get().MQCfg.Type.Key: func(string) string { return walName.String() },
+		}); err != nil {
 			mlog.Error(context.TODO(), "failed to process immutable configs", mlog.Err(err))
 			return
 		}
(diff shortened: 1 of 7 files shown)


czs007:

PR #51307, opened 2026-07-13
fix: refresh expired next target for unavailable delegator

chyezh:
issue: #48903

- querycoord target observer: require a runtime-synced NextTarget view to be serviceable before it can pin an expired NextTarget refresh
- target observer tests: cover expired refresh when a delegator has the next target version but is not serviceable

PR #50686, opened 2026-06-22
fix: sync target after delegators ready

chyezh:
issue: #48903

- querycoord target observer: split NextTarget handling into prepare and commit phases so readable views are synced only after all required delegators are ready
- next target expiry: keep an in-progress commit version from being replaced by the survival-time refresh path
- target observer tests: cover partial delegator readiness and commit-start expiry blocking

Issue #48903, opened 2026-04-10
[Bug]: [Benchmark][Cluster] HybridSearch/Search/Query 503 'channel distribution is not serviceable' during streaming DML+DQL with StreamingNode segment rebalance


wangting0128:
- [x] I have searched the existing issues

### Environment

- Milvus version: `3.0-20260409-08bec0a3-amd64` (commit: `08bec0a3`)
- Deployment mode: cluster (streaming mode enabled, with StreamingNode)
- MQ type: pulsarv3
- OS: Linux (4AM cluster)
- CPU/Memory:
  - queryNode: 8C/8Gi (limits), 5C/5Gi (requests), 1 replica
  - dataNode: 10C/8Gi (limits), 6C/5Gi (requests), 1 replica
  - streamingNode: 2C/8Gi (limits), 2C/5Gi (requests), 1 replica
- Deploy config: `cluster_2c8m`

### Current Behavior

During the `test_hybrid_search_locust_dml_dql_streaming_cluster` benchmark test, hybrid_search/search/query operations intermittently fail with HTTP 503 error: `channel distribution is not serviceable`. The failures occur when the **StreamingNode** (registered as QueryNode-2, ServerID=2) is performing segment rebalance — its `currentLoadRatio` drops to ~62% while `requiredLoadRatio=1.0`, causing the channel to be marked as not serviceable.

```
[setup_method] Test case: test_hybrid_search_locust_dml_dql_streaming_cluster, Test run_id: 2026040933416400 (base.py:76)
```

**Error examples** (requestId = Milvus traceID, can be used to trace in Loki):

```
[2026-04-09 19:01:31,375 - ERROR - fouram]: (api_response) : [MilvusClient.hybrid_search] <MilvusException: (code=503, message=fail to search on QueryNode 2: channel distribution is not serviceable: channel not available[channel=by-dev-rootcoord-dml_0_465500508362639217v0])>, [requestId: 60a0a805d671447e8177403eec7ee9dd] (api_request.py:64)

[2026-04-09 20:01:06,256 - ERROR - fouram]: (api_response) : [MilvusClient.query] <MilvusException: (code=503, message=fail to Query on QueryNode 2: channel distribution is not serviceable: channel not available[channel=by-dev-rootcoord-dml_1_465500508362639217v1])>, [requestId: bdcf008dc41c492fa65c706bdb41f5f4] (api_request.py:64)

[...]

Issue #51332, opened 2026-07-14
[Bug]: Multi-replica rolling update stalls when blocked StreamingNode channel plans starve QueryNode segment drain


zhuwenxing:
## Environment

- Milvus version: master
- Milvus commit: `944b614a8ce763f01c3729140b095367838f5908`
- Image: `harbor.milvus.io/manta/milvus:master-20260713-944b614`
- Target image: `harbor.milvus.io/manta/milvus@sha256:1155044644e6143747a3f87e596441527f652ba55f380fd5d6d38f7573759d6b`
- Deployment mode: Kubernetes cluster, Milvus Operator `rollingUpgrade`
- MQ type: Woodpecker
- SDK version: PyMilvus 3.1.0rc64, commit `2b42a8552855332dad5cc020f721bc6dc44f496d`
- Python / XGBoost: 3.12.12 / 3.3.0
- Kubernetes: v1.24.2
- OS: Rocky Linux 8.6, amd64, containerd 1.6.6
- Topology: 2 Proxy, 2 legacy QueryNode, 2 StreamingNode, 1 MixCoord, 1 DataNode
- Collection replicas: 2
- QueryNode / StreamingNode resources: 2 CPU / 8 GiB each
- GPU: N/A
- Namespace: `chaos-testing`
- Instance: `xgboost-chain-cluster-jeremy-0713`
- Test window: `2026-07-13T11:19:57Z`–`2026-07-13T11:29:58Z`

### K8s Pod List

The following is the historical Prometheus Ready/restart snapshot at `2026-07-13T11:19:57Z`, immediately before the rolling update. The dedicated test instance has since been cleaned up.

```text
NAME                                                                    READY   STATUS    RESTARTS
xgboost-chain-cluster-jeremy-0713-etcd-0                                1/1     Running   0
xgboost-chain-cluster-jeremy-0713-minio-798fb7f555-kljr4                1/1     Running   0
xgboost-chain-cluster-jeremy-0713-milvus-mixcoord-58c94cfdgc9bc         1/1     Running   0
xgboost-chain-cluster-jeremy-0713-milvus-proxy-6d9fbf98f4-7p95p         1/1     Running   0
xgboost-chain-cluster-jeremy-0713-milvus-proxy-6d9fbf98f4-q448t         1/1     Running   0
xgboost-chain-cluster-jeremy-0713-milvus-querynode-0-6755b24wx7         1/1     Running   0
[...]

PR #51547, opened 2026-07-17
fix: revert target sync readiness change

chyezh:
issue: #48903

- QueryCoord target observer: revert the stricter all-delegator readiness gate introduced by #50686
- target observer tests: restore the previous partial-ready delegator sync coverage
- release job tests: remove the target-version mock added for the reverted target observer path

### DIFF
--- a/internal/querycoordv2/job/job_release_test.go
+++ b/internal/querycoordv2/job/job_release_test.go
@@ -147,7 +147,6 @@ func newReleaseJobDeps(t *testing.T, m *meta.Meta, dist *meta.DistributionManage
 	targetMgr.EXPECT().IsNextTargetExist(mock.Anything, mock.Anything).Return(true).Maybe()
 	targetMgr.EXPECT().IsCurrentTargetExist(mock.Anything, mock.Anything, mock.Anything).Return(true).Maybe()
 	targetMgr.EXPECT().GetDmChannelsByCollection(mock.Anything, mock.Anything, mock.Anything).Return(map[string]*meta.DmChannel{}).Maybe()
-	targetMgr.EXPECT().GetCollectionTargetVersion(mock.Anything, mock.Anything, mock.Anything).Return(int64(0)).Maybe()
 	targetMgr.EXPECT().UpdateCollectionNextTarget(mock.Anything, mock.Anything).Return(nil).Maybe()
 	targetMgr.EXPECT().RemoveCollection(mock.Anything, mock.Anything).Return().Maybe()
 	nodeMgr := session.NewNodeManager()
--- a/internal/querycoordv2/observers/target_observer.go
+++ b/internal/querycoordv2/observers/target_observer.go
@@ -86,12 +86,9 @@ type TargetObserver struct {
 	initChan chan initRequest
 	// nextTargetLastUpdate map[int64]time.Time
 	nextTargetLastUpdate *typeutil.ConcurrentMap[int64, time.Time]
-	// nextTargetSyncStarted records the NextTarget version after the observer
-	// enters SyncTargetVersion commit phase for a collection.
-	nextTargetSyncStarted *typeutil.ConcurrentMap[int64, int64]
-	updateChan            chan targetUpdateRequest
-	mut                   sync.Mutex                // Guard readyNotifiers
-	readyNotifiers        map[int64][]chan struct{} // CollectionID -> Notifiers
+	updateChan           chan targetUpdateRequest
+	mut                  sync.Mutex                // Guard readyNotifiers
+	readyNotifiers       map[int64][]chan struct{} // CollectionID -> Notifiers
 
 	// loadingDispatcher updates targets for collections that are loading (also collections without a current target).
 	loadingDispatcher *taskDispatcher[int64]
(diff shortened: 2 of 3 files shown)


czs007:

Reviewed at head `720e5fc328` against `master`. **The revert itself is clean and I have no blocking objection** — it is a byte-exact inverse of `4bfdd7f3e5` (#50686): `git diff 4bfdd7f3e5^ 720e5fc328 -- internal/querycoordv2/{observers/target_observer.go,observers/target_observer_test.go,job/job_release_test.go}` is empty, no other commit has touched these three files since #50686, and `grep` finds no residual references to the removed `nextTargetSyncStarted` / `isNextTargetSyncStarted` symbols anywhere in the tree. Reverting both sub-changes together (the prepare/commit gate *and* the expiry pinning in `shouldUpdateNextTarget`) is the right scope — the pinning only existed to serve the commit phase, so keeping one without the other would not be coherent.

The mechanism also checks out. Under #50686, `shouldUpdateCurrentTarget` returned `false` from the prepare loop on the *first* not-ready delegator, **before** `syncNextTargetToDelegator` was ever called. During a rolling update the dist holds both the old (stopping/RO) and the new delegator for the same `(replica, channel)` pair, so the stale one aborts the whole prepare phase → no `SyncTargetVersion` is sent to anyone → the new delegator never becomes serviceable → `ChannelDistManager.GetShardLeader` (`internal/querycoordv2/meta/channel_dist_manager.go:322`, which prefers serviceable candidates) never picks it → `scheduler.go:1160-1183` blocks the channel Reduce forever → `StoppingBalancer` never emits segment plans → the old QueryNode never drains. That is a genuine self-sustaining livelock and matches #51332's log evidence exactly. For a loaded collection the other sync path at `target_observer.go:321` is unreachable (it is gated on the current target *not* existing), so there was no escape hatch. The restored code calls `syncNextTargetToDelegator` unconditionally with whatever delegators are ready, which allows the partial progress that breaks the cycle. Good call.

Findings below are all non-blocking follow-ups.

### medium — #51307 is still open against master and patches the code this PR deletes
PR #51307 (`fix: refresh expired next target for unavailable delegator`, issue #48903, branch `48903-next_target_serviceable_refresh`) modifies `internal/querycoordv2/observers/target_observer.go` by hardening exactly the `isNextTargetSyncStarted` path removed here. Once this merges, #51307 will conflict and is moot. Its own 2.6 cherry-pick #51264 was already closed with the note "this branch does not have `nextTargetSyncStarted` runtime-view pinning" — which is precisely the state master returns to here.

**Suggestion:** close #51307 as part of landing this, referencing this PR, so it does not get rebased back on top of deleted code.

### medium — the PR body drops the #48903 thread entirely
The body links only `issue: #51332`. #48903 was closed on 2026-04-30, i.e. *before* #50686 (2026-06-26), so #50686 was follow-up hardening rather than the closing fix — reverting it does not directly reopen the original 503 report. But the hardening it provided (not exposing a partially-synced readable view that can later be discarded as an orphan NextTarget) is now gone with no tracking issue pointing at it.

**Suggestion:** add a line to the body naming what is being given back up (`reverts the all-delegator readiness gate from #50686; the partial-sync exposure it guarded against is re-accepted, tracked in #48903`) and reopen #48903 or file a fresh issue. Otherwise the next person to hit this rediscovers it from scratch.

### medium — restored code: the per-replica loop does not actually enforce per-replica readiness (pre-existing, re-accepted here)
`internal/querycoordv2/observers/target_observer.go:489-514`:

```go
// Iterate through each replica to check if all its delegators are ready.
// this approach ensures each replica has at least one ready delegator for every channel.
// This prevents the issue where some replicas may lack nodes during dynamic replica scaling,
// while the total count still meets the threshold.
readyDelegatorsInCollection := make([]*meta.DmChannel, 0)
    readyDelegatorsInReplica := make([]*meta.DmChannel, 0)
    for channel := range channelNames { ... }
    readyDelegatorsInCollection = append(readyDelegatorsInCollection, readyDelegatorsInReplica...)
}
syncSuccess := ob.syncNextTargetToDelegator(ctx, collectionID, readyDelegatorsInCollection, newVersion)
syncedChannelNames := lo.Uniq(lo.Map(readyDelegatorsInCollection, func(ch *meta.DmChannel, _ int) string { return ch.ChannelName }))
if !syncSuccess || !lo.Every(syncedChannelNames, lo.Keys(channelNames)) {
    return false
}
```

The per-replica accumulation is decorative: `readyDelegatorsInReplica` is flattened straight into `readyDelegatorsInCollection` and never checked per replica, so the final `lo.Every` is a **collection-wide** channel-coverage check. **Failure scenario:** collection with channels `{ch0, ch1}` and `replica_number=2`. Replica A is ready on `ch0` only, replica B is ready on `ch1` only. `syncedChannelNames = [ch0, ch1]`, `lo.Every` passes, and the current target is promoted even though *neither* replica is fully ready — the exact "some replicas may lack nodes while the total count still meets the threshold" case the comment claims to prevent. This is pre-existing on master pre-#50686 and not introduced by this PR, but it is the real gap #50686 was reaching for, so it should be captured rather than silently restored.

**Suggestion:** track it on the follow-up issue, and either fix the check to be per-replica (`lo.Every(uniq(channels(readyDelegatorsInReplica)), lo.Keys(channelNames))` inside the loop, `return false` on miss) or correct the comment to describe what the code actually does. Any re-landed version of #50686 must additionally exclude stopping/RO delegators from the readiness set, otherwise it reintroduces #51332.

### low — no evidence the rolling-update stall was re-verified
This reverts a behavioral change to fix a convergence failure that only reproduces under the Operator rolling sequence with transient old/new node coexistence (#51332 explicitly says it is not reducible to an API-only script). The diff restores unit coverage but adds nothing that exercises the stall, so CI green here is only evidence the happy path is intact, not that #51332 converges.

**Suggestion:** state in the body that the `xgboost-chain-cluster` rolling-update scenario was rerun on a build with this revert and converged within the 600s window. If that rerun has not happened yet, say so — a revert landed on the strength of code reasoning alone is fine, but should be labelled as such.

### nit — `shouldUpdateCurrentTarget` performs RPCs
`target_observer.go:449` — a `should*` predicate that issues `ListIndexes` and `SyncTargetVersion` to every delegator as a side effect is surprising, and it is what makes the "gate the sync on readiness that the sync itself produces" mistake easy to make. Pre-existing and out of scope for a revert; worth renaming (`trySyncAndCheckCurrentTarget`) whenever this function is next touched.

chyezh:
may block the replica management or rolling upgrading

liliu-z (commented):
/lgtm
/approve

PR #51548, opened 2026-07-17
fix: [cp 3.0] revert target sync readiness change

chyezh:
issue: #48903
pr: #51547

- QueryCoord target observer: revert the stricter all-delegator readiness gate introduced by #50686
- target observer tests: restore the previous partial-ready delegator sync coverage
- release job tests: remove the target-version mock added for the reverted target observer path

### DIFF
--- a/internal/querycoordv2/job/job_release_test.go
+++ b/internal/querycoordv2/job/job_release_test.go
@@ -147,7 +147,6 @@ func newReleaseJobDeps(t *testing.T, m *meta.Meta, dist *meta.DistributionManage
 	targetMgr.EXPECT().IsNextTargetExist(mock.Anything, mock.Anything).Return(true).Maybe()
 	targetMgr.EXPECT().IsCurrentTargetExist(mock.Anything, mock.Anything, mock.Anything).Return(true).Maybe()
 	targetMgr.EXPECT().GetDmChannelsByCollection(mock.Anything, mock.Anything, mock.Anything).Return(map[string]*meta.DmChannel{}).Maybe()
-	targetMgr.EXPECT().GetCollectionTargetVersion(mock.Anything, mock.Anything, mock.Anything).Return(int64(0)).Maybe()
 	targetMgr.EXPECT().UpdateCollectionNextTarget(mock.Anything, mock.Anything).Return(nil).Maybe()
 	targetMgr.EXPECT().RemoveCollection(mock.Anything, mock.Anything).Return().Maybe()
 	nodeMgr := session.NewNodeManager()
--- a/internal/querycoordv2/observers/target_observer.go
+++ b/internal/querycoordv2/observers/target_observer.go
@@ -86,12 +86,9 @@ type TargetObserver struct {
 	initChan chan initRequest
 	// nextTargetLastUpdate map[int64]time.Time
 	nextTargetLastUpdate *typeutil.ConcurrentMap[int64, time.Time]
-	// nextTargetSyncStarted records the NextTarget version after the observer
-	// enters SyncTargetVersion commit phase for a collection.
-	nextTargetSyncStarted *typeutil.ConcurrentMap[int64, int64]
-	updateChan            chan targetUpdateRequest
-	mut                   sync.Mutex                // Guard readyNotifiers
-	readyNotifiers        map[int64][]chan struct{} // CollectionID -> Notifiers
+	updateChan           chan targetUpdateRequest
+	mut                  sync.Mutex                // Guard readyNotifiers
+	readyNotifiers       map[int64][]chan struct{} // CollectionID -> Notifiers
 
 	// loadingDispatcher updates targets for collections that are loading (also collections without a current target).
 	loadingDispatcher *taskDispatcher[int64]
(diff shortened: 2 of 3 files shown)


czs007:

Reviewed at head `13889655ae` against base `3.0`. **No blocking code issues** — this is a faithful, complete revert. My comments are about scope/justification and the known weaknesses of the code being restored, not about defects introduced by this PR.

### Verification of the revert itself

- The three files are restored **byte-for-byte** to their pre-#50686 state (`b07c62d53c`): `git diff b07c62d53c HEAD -- internal/querycoordv2/observers/target_observer.go internal/querycoordv2/observers/target_observer_test.go internal/querycoordv2/job/job_release_test.go` is empty.
- The revert scope matches the reverted commit exactly: `4bfdd7f3e5` ("fix: sync target after delegators ready (#50686)") touched exactly these three files, and no other commit on `3.0` since then depends on the removed symbols (`rg 'nextTargetSyncStarted|isNextTargetSyncStarted'` returns nothing outside unrelated `writebuffer` test locals). No orphaned references, no stranded follow-ups.
- The diff is identical to master PR #51547 (modulo `index` lines). Good cherry-pick parity.
- Removing the `GetCollectionTargetVersion` expectation in `job_release_test.go:150` is correct and safe: it was only needed because #50686's `isNextTargetSyncStarted` called it from `shouldUpdateNextTarget`. In the reverted code, `shouldUpdateNextTarget` (`target_observer.go:419-421`) only calls `IsNextTargetExist` + `isNextTargetExpired`, and `shouldUpdateCurrentTarget` returns early at `target_observer.go:458-462` because the test mocks `GetDmChannelsByCollection` to an empty map — so the call site is unreachable in that suite.
- Release exposure is low: no release tag contains `4bfdd7f3e5`, so the stricter gate never shipped in a `3.0` release. That materially lowers the risk of this revert.

### Findings

**medium — the PR does not acknowledge that this reopens #48903.**
#50686 was the fix for #48903 (intermittent `code=503 ... channel distribution is not serviceable` during StreamingNode segment rebalance, ~0.67% of hybrid_search/search/query in the locust benchmark). Reverting restores the partial-sync behavior, and the restored test asserts exactly that regression surface — `target_observer_test.go:729-736`:

```go
result := observer.shouldUpdateCurrentTarget(ctx, collectionID)
assert.True(t, result)   // node 2 is NOT ready, yet current target is promoted
```

Trading a transient-503 regression for a rollout that never converges is the right call for a release branch, but the PR body only lists the mechanical changes. Please state in the body that #48903 regresses, and either reopen it or link the follow-up that will re-land a correct version of the gate. Otherwise the next person to touch this file has no record of why the stricter gate was removed and will re-introduce it.

**medium — no stated validation that the revert actually resolves #51332.**
Issue #51332's own root-cause section attributes the stall to a cross-controller cycle (StreamingCoord keeps `rootcoord-dml_0` on node 14 → QueryCoord channel Reduce blocks in `task/scheduler.go` waiting for node 15 to become shard leader → `StoppingBalancer` emits no segment plans while channel plans exist → QueryNode 12 never drains). The target observer readiness gate is a plausible contributor — under #50686, `syncNextTargetToDelegator` is reached only after every delegator in every replica is ready, so a fresh delegator on node 15 could never receive `SyncTargetVersion` and never become serviceable, which is precisely the link the blocked Reduce waits on. But that chain is not asserted anywhere in the PR, and the issue names a different primary mechanism. Please add the repro evidence (rolling-update run converging on a build with the revert) to the body, or note explicitly that this is a mitigation of one link in the cycle rather than a full fix for #51332.

**low — merge ordering vs master.**
Master PR #51547 is still `OPEN`. Per the branch convention the cherry-pick should land after master, otherwise `3.0` diverges if #51547 changes during review. The `pr: #51547` link is correctly present; just hold this until master merges.

**low — restored code does not enforce what its own comment claims (pre-existing, not introduced here).**
`target_observer.go:489-514`:

```go
// this approach ensures each replica has at least one ready delegator for every channel.
readyDelegatorsInCollection := make([]*meta.DmChannel, 0)
    ...
    readyDelegatorsInCollection = append(readyDelegatorsInCollection, readyDelegatorsInReplica...)
}
syncedChannelNames := lo.Uniq(lo.Map(readyDelegatorsInCollection, func(ch *meta.DmChannel, _ int) string { return ch.ChannelName }))
if !syncSuccess || !lo.Every(syncedChannelNames, lo.Keys(channelNames)) {
    return false
}
```

`syncedChannelNames` is the **union across all replicas**, so the coverage check is collection-wide, not per-replica. Failure scenario: with `replica_number=2` and channels `{c1, c2}`, if replica A has only `c1` ready and replica B has only `c2` ready, the union is `{c1, c2}`, `lo.Every` passes, and the current target is promoted while neither replica is individually complete. The comment asserts the opposite invariant. This is the gap #50686 was aiming at, so it is expected to come back with the revert — worth capturing in the follow-up issue rather than fixing here.

**nit — stale test comment.**
`target_observer_test.go:606-608`: "This test specifically validates the fix for the bug where all delegators (including non-ready ones) were being added to `readyDelegatorsInReplica`." After the revert this reads as if it guards a fix that is no longer the current concern. Harmless (it is a verbatim restore), but a one-line note that partial-readiness sync is intentional would help the next reader.

**nit — dropped nil guards.**
#50686 added `if delegator == nil || delegator.View == nil { return false }`; the revert removes them, and `checkDelegatorDataReady` dereferences `channel.View.TargetVersion` at `target_observer.go:479`/`486` unguarded. I checked and this is not a live risk: every `DmChannel` entering `ChannelDistManager` is built at `dist/dist_handler.go:275` with a non-nil `View`, and `DmChannel.Clone()` (`meta/channel_dist_manager.go:167`) would already panic on a nil `View`, so the invariant holds. Noting only so it is a conscious decision rather than an oversight.

### Suggested actions

1. Expand the PR body: state that #48903 regresses, link the follow-up, and add the #51332 repro evidence (or scope the claim to "mitigation").
2. Hold the merge until master #51547 lands, to keep `3.0` from diverging.
3. File the follow-up capturing both the per-replica coverage gap (`lo.Every` over the cross-replica union) and the #48903 transient-503 window, so the correct gate can be re-landed with the rolling-update drain path covered by a test.

Code-wise this is clean and I have no objection to the revert landing on `3.0` once the above is addressed.

liliu-z (commented):
/lgtm
/approve

PR #51302, opened 2026-07-13
fix: avoid structured binding capture in OpenMP lambdas

sijie-ni-0214:
## Summary

Avoid direct captures of structured bindings in `ChunkedSegmentSealedImpl.cpp`, which is compiled with OpenMP enabled.

- Use an init-capture for the field ID in `LoadBatchTextIndexes`.
- Use an init-capture for the field ID in `FinalizeLoadDiffForReopen` when dropping JSON indexes.
- Keep the inner text-index commit lambda as a normal value capture because it captures the ordinary init-capture from the outer lambda, not a structured binding.

issue: #51231

## Test Plan

- [x] `make`
- [x] `ninja -C cmake_build src/segcore/CMakeFiles/milvus_segcore.dir/ChunkedSegmentSealedImpl.cpp.o`
- [x] `git diff --check`

Issue #51231, opened 2026-07-10
[Bug]: C++ build fails with OpenMP when a lambda captures a structured binding


sijie-ni-0214:
## Environment

- **Milvus Version**: master at df36780667
- **Deployment Mode**: N/A
- **MQ**: N/A
- **SDK**: N/A
- **OS**: macOS arm64
- **Compiler**: Homebrew Clang 18.1.8 with OpenMP enabled via libomp

### Option B: Steps

1. Build Milvus from current master on macOS arm64 with Clang/OpenMP enabled.
2. Compile `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp`.
3. Observe the compiler failure around `ChunkedSegmentSealedImpl::LoadBatchJsonKeyIndexes`.

### Trigger Conditions

- **Frequency**: always in the affected Clang/OpenMP environment
- **First observed after**: commit `df36780667` on master
- **Does NOT happen when**: compiling a reduced sample without OpenMP, or when the lambda uses an init-capture instead of directly capturing the structured binding

## Expected Behavior

The C++ native build should compile successfully with OpenMP enabled.

## Actual Behavior

The build fails because Clang does not support directly capturing a structured binding in an OpenMP-enabled compilation unit.

## Error Logs

```text
capturing a structured binding is not yet supported in OpenMP
```

## Non-default Configuration

```yaml
N/A
```

## Analysis Hints

The failing code is in `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp` inside `ChunkedSegmentSealedImpl::LoadBatchJsonKeyIndexes`.

The loop uses a structured binding:

```cpp
for (const auto& [field_id, info_proto] : infos) {
```

The lambda then directly captures `field_id`:

```cpp
[field_id, index = std::move(index)](...)
```

With Clang/OpenMP, that direct structured-binding capture is rejected. Using a generalized init-capture keeps the same value-capture semantics while avoiding the unsupported capture form:

```cpp
[field_id = field_id, index = std::move(index)](...)
```

sijie-ni-0214:
Proposed fix: https://github.com/milvus-io/milvus/pull/51232

PR #51549, opened 2026-07-17
fix: [cp3.0] avoid structured binding capture in OpenMP lambdas

sijie-ni-0214:
Cherry-pick from master
pr: #51302

issue: #51231

Use init-captures for structured-binding field IDs in OpenMP lambdas. The 3.0 adaptation also fixes the matching JSON stats callback introduced by the 3.0 runtime-state backport.

## Test Plan

- `/opt/homebrew/opt/llvm@18/bin/clang-format -i --style=file internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp`
- `git diff --check`
- `ninja -C cmake_build -j1 -v src/segcore/CMakeFiles/milvus_segcore.dir/ChunkedSegmentSealedImpl.cpp.o`


czs007:

Issue #51553, opened 2026-07-17
[Bug]: json path index should not cause json raw data to be released since it does not include the full JSON data


zhengbuqian:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version:
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

as the title says

### Steps To Reproduce

```markdown

```

yanliang567:
@zhengbuqian any repo code for testing?

zhengbuqian:
fixed by pr of #51299

PR #51554, opened 2026-07-17
fix: keep JSON field data for path indexes

zhengbuqian:
issue: https://github.com/milvus-io/milvus/issues/51553


zhengbuqian:
fixed in https://github.com/milvus-io/milvus/pull/51541

Issue #51557, opened 2026-07-17
[Bug]: newly added BM25 sparse field is not searchable after segment reopen


MrPresent-Han:
### Current Behavior

After `addFunctionField` adds a BM25 function that outputs a new sparse vector field and the backfill completes, existing **sealed** segments are reloaded through the Reopen path (`SegmentChecker` → `LoadScope_Reopen` → `QueryNode.reopenSegments` → `LocalSegment.Reopen` → `csegment.Reopen`). A BM25 search on the newly added sparse field then **fails**.

### Root Cause

The segcore segment's collection-level index meta (`col_index_meta_` in `ChunkedSegmentSealedImpl`) is captured **once at segment creation** (from `col->get_index_meta()`) and is **never refreshed on reopen** — `ReopenRequest` carried only `LoadInfo` + `Schema`, no index meta.

The field added after creation is therefore absent from the stale `col_index_meta_`, so:

- A sealed BM25 sparse field with no vector index falls to the brute-force path and reads its k1/b via `col_index_meta_->GetFieldIndexMeta(field_id)`, which hits `CollectionIndexMeta::GetFieldIndexMeta` → `AssertInfo` (throws even in release builds) → the search fails.
- `generate_interim_index` returns `false` on `!HasField`, so no interim index is ever built for the field, forcing it onto the crashing brute-force path.

### Expected Behavior

Reopen should refresh the segment's collection-level index meta so that a field added after the segment was created (e.g. a backfilled BM25 sparse field) becomes searchable without waiting for a full reload.

### Environment

- Milvus branch: master (targeting 3.0)
- Component: querynode / segcore

PR #51558, opened 2026-07-17
fix: sync collection index meta into segment on reopen

MrPresent-Han:
issue: #51557 #51584
related: https://github.com/milvus-io/milvus/issues/51366

## Problem

After `addFunctionField` adds a BM25 function that outputs a new sparse vector field and the backfill completes, existing **sealed** segments are reloaded through the Reopen path (`SegmentChecker` → `LoadScope_Reopen` → `QueryNode.reopenSegments` → `LocalSegment.Reopen` → `csegment.Reopen`). A BM25 search on the newly added sparse field then fails.

**Root cause:** the segcore segment's collection-level index meta (`col_index_meta_` in `ChunkedSegmentSealedImpl`) is captured once at segment creation and never refreshed on reopen (`ReopenRequest` carried only `LoadInfo` + `Schema`). The new field is absent from the stale meta, so:

- brute-force BM25 search reads k1/b via `col_index_meta_->GetFieldIndexMeta(field_id)` → `AssertInfo` (throws even in release) → search fails;
- `generate_interim_index` returns `false` on `!HasField`, so no interim index is built for the field.

## Fix

Fold `col_index_meta` into the published-state snapshot and thread the latest `CollectionIndexMeta` through the reopen path. Only the reopen path is touched — full load already snapshots the up-to-date meta at segment creation.

- **C++ segcore:** `col_index_meta` moves into `PublishedSegmentState`/`StateDelta`; the bare `col_index_meta_` member is removed. Readers (brute-force search, `generate_interim_index`) now read it from the published/staged snapshot, so reopen can refresh it without racing readers.
- **C++ + cgo:** the 3-arg `Reopen` and `AsyncReopenSegment` gain an `IndexMetaPtr` / `index_meta_blob` (nullptr/empty keeps the current meta). Reopen sets the staged + delta `col_index_meta` and publishes it atomically.
- **Stale-schema reopen:** a reopen that races behind a concurrent `LazyCheckSchema` schema advance clamps to the current schema instead of returning early — it still applies the reopen's load info + index meta, rather than silently reporting success and stranding the segment on stale data/meta with no retrigger (index meta needs no clamp: `Reopen` is its only, serialized writer that `LazyCheckSchema` never races).
- **Go:** `ReopenRequest.IndexMeta`; `cSegmentImpl.Reopen` marshals it to a blob; `LocalSegment.Reopen` fills it from `s.collection.SchemaVersionAndIndexMeta()` (one transition-lock snapshot of schema + version + meta, so a reopen never reads a torn `(schema, meta)` pair).

## Verification

- [x] Adversarial self-review: all `Reopen` overrides consistent across interface/growing/sealed/cgo/Go; end-to-end trace confirms the new field's meta reaches the published snapshot and search no longer asserts.
- [x] Local Debug compile — built and deployed locally (StorageV3 + `bumpSchemaVersion` enabled).
- [x] Unit tests — segcore reopen tests (`test_sealed.cpp`: `SchemaAwareReopenDiscardsOlderSchema`, `ReopenWithNonNullIndexMetaPublishesNewFieldIndex`) + Go collection/reopen tests; run under CI `build-ut-cov`.
- [x] Reproduction — `repro_51366.py`: 4 below-threshold sealed segments, `addFunctionField` → backfill → auto reopen → BM25 search becomes sustained-searchable in ~15s with **0** `field index meta not found` over a 120s window (run on the built binary).

---

## Commit 2 — index-meta version gate (#51584)

`CollectionIndexMeta` had no version, so QueryNode refreshed collection index meta unconditionally on every load: a delayed pre-DDL `LoadSegments` whose stale schema is correctly rejected could still clobber fresher index meta (schema/meta inconsistent until the next in-order load), and a `createIndex`-only change could not be ordered against concurrent loads at all. Both races were left open by commit 1 and tracked in #51584.

Give index meta its own ordering domain, orthogonal to `schema.Version` and `schema_barrier_ts`:

- **DataCoord** stamps every index-meta mutation (`CreateIndex` / `AlterIndex` / `MarkIndexAsDeleted`) with a TSO timestamp drawn inside `fieldIndexLock`, so version order matches state order. `reloadFromKV` re-stamps the reloaded snapshot with a fresh base version; the TSO's global monotonicity guarantees it exceeds every pre-restart version, keeping the order crash-safe with no new persistence.
- **QueryCoord** `broker.ListIndexes` returns the `(snapshot, version)` pair read under one lock hold and threads it through `LoadMetaInfo.index_meta_version` on LoadSegments / WatchDmChannels / SyncDistribution.
- **QueryNode** applies an index-meta payload only when its version is newer than the applied one. Version 0 (legacy coordinator / DescribeIndex fallback) keeps the old unconditional refresh, so rolling upgrade behaves as before.

### Verification (commit 2)
- [x] DataCoord UT: monotonic stamping across the three mutations; TSO-alloc failure aborts before the catalog write; catalog failure does not bump; reload stamps the base version; nil allocator keeps legacy version-0 behavior.
- [x] QueryNode UT: stale version rejected (race 1), same-schema newer index version applied (race 2), equal version dropped, version-0 payload keeps unconditional refresh without regressing the stored version.
- [x] QueryCoord UT: version passthrough from `ListIndexesResponse`; `DescribeIndex` fallback reports 0.
- [x] Regression: querycoordv2 (root/task/checkers/observers/meta), datacoord index-meta tests, `make build-go` — all green locally.


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -6397,12 +6403,22 @@ ChunkedSegmentSealedImpl::generate_interim_index(
...
+    if (committer == nullptr) {
+        published_snapshot = CapturePublishedState();
+    }
+    const IndexMetaPtr& col_index_meta =
+        committer != nullptr ? committer->staged_state()->col_index_meta
+                             : published_snapshot->col_index_meta;
+    if (col_index_meta == nullptr || !col_index_meta->HasField(field_id)) {
         return false;
     }
     auto schema_snapshot = CaptureSchemaSnapshot();
```
liliu-z: On the reopen path generate_interim_index reads the refreshed meta from the staged snapshot (committer->staged_state()->col_index_meta) but still takes the schema from CaptureSchemaSnapshot(), which returns the published snapshot's schema and is not swapped to the new schema until committer.Publish at line 7244. For a newly added field, HasField passes on the staged meta, then schema_snapshot->operator[](field_id) at line 6420 hits the AssertInfo in Schema::operator[] and throws a SegcoreError — outside the try at line 6466. WaitAllFutures rethrows it out of the column-group load and Reopen has no surrounding catch, so the whole reopen fails and the meta is never published, defeating the fix. Source the schema from the same state as the meta (committer->staged_state()->schema, else published_snapshot->schema).

internal/core/src/segcore/ChunkedSegmentSealedImpl.h
```diff
@@ -309,7 +309,8 @@ class ChunkedSegmentSealedImpl : public SegmentSealed {
     void
     Reopen(milvus::OpContext* op_ctx,
            const milvus::proto::segcore::SegmentLoadInfo& new_load_info,
-           SchemaPtr new_schema) override;
+           SchemaPtr new_schema,
```
liliu-z, line 296: The `Reopen` override now takes a fourth parameter `const IndexMetaPtr& new_index_meta` with no default and no forwarding overload. Existing callers still pass three arguments — `test_sealed.cpp:3845` invokes `sealed->Reopen(&op_ctx, stale_proto, stale_schema)` — so that test target fails to compile with a deterministic 'no matching function' error. Add a defaulted/forwarding overload or update the test callers.

internal/querynodev2/segments/segment.go
```diff
@@ -1308,6 +1308,10 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
 		LoadInfo:      newLoadInfo,
 		Schema:        schema,
 		SchemaVersion: schemaVersion,
+		// Refresh the collection-level index meta so a field added after this
+		// segment was created (e.g. a backfilled BM25 sparse field) is
+		// searchable without waiting for a full reload.
+		IndexMeta: s.collection.GetCCollection().IndexMeta(),
```
liliu-z: `IndexMeta: s.collection.GetCCollection().IndexMeta()` unconditionally dereferences the native CCollection, but `TestLocalSegmentReopenUsesSegcoreSchemaVersion` constructs `&Collection{}` with a mock csegment and `setSchema` (collection.go:488) only sets `c.schema`, never `c.ccollection`. `GetCCollection()` therefore returns nil and `IndexMeta()` nil-derefs, panicking deterministically before the mocked Reopen runs. Guard the nil CCollection or populate it in the test.

MrPresent-Han: Fixed — added a nil-safe, lock-guarded accessor `Collection.GetIndexMeta()` that returns nil when the collection is released (mirroring the existing `c.ccollection == nil` guards on the other accessors), and `Reopen` now uses it instead of `GetCCollection().IndexMeta()`. `TestLocalSegmentReopenUsesSegcoreSchemaVersion` passes locally.

internal/core/src/segcore/ChunkedSegmentSealedImpl.h
MrPresent-Han: Fixed by updating the test caller: `test_sealed.cpp:3845` now calls the 4-arg `Reopen(&op_ctx, stale_proto, stale_schema, nullptr)` (the impl falls back to `current->col_index_meta` when `new_index_meta` is null, matching the 2-arg overload which itself forwards `nullptr, nullptr`). The `no matching function` compile error is resolved.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
MrPresent-Han: Good catch — fixed. `generate_interim_index` now reads the schema and `col_index_meta` from the **same** snapshot: during reopen (`committer != nullptr`) both come from `committer->staged_state()`, otherwise both from the published snapshot. Previously it read the staged `col_index_meta` but the published (pre-`Publish`) schema, so a newly added field passed `HasField` on the staged meta and then threw in `schema_snapshot->operator[](field_id)` — outside the `try` — failing the whole reopen. They are consistent now.

internal/querynodev2/segments/segment.go
```diff
@@ -1308,6 +1308,10 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
 		LoadInfo:      newLoadInfo,
 		Schema:        schema,
 		SchemaVersion: schemaVersion,
+		// Refresh the collection-level index meta so a field added after this
+		// segment was created (e.g. a backfilled BM25 sparse field) is
+		// searchable without waiting for a full reload.
+		IndexMeta: s.collection.GetIndexMeta(),
```
liliu-z: On reopen, LocalSegment.Reopen reads the collection's index meta via s.collection.GetIndexMeta(), but applyLoadUpdate version-gates only the schema — updateIndexMeta runs unconditionally and short-circuits solely on proto.Equal, with no version check. A delayed pre-DDL LoadSegments whose stale schema is correctly rejected still overwrites the fresh index meta with its old value, so every subsequent reopen publishes the new schema paired with stale meta and the brute-force BM25 path asserts in GetFieldIndexMeta on the added field until a newer load arrives. Version-gate updateIndexMeta the same way the schema is gated, or thread the accepted request's own meta through ReopenSegments/LocalSegment.Reopen.

internal/core/unittest/test_sealed.cpp
```diff
@@ -3842,7 +3842,8 @@ TEST(SealedSegmentReopen, SchemaAwareReopenDiscardsOlderSchema) {
     proto::segcore::SegmentLoadInfo stale_proto =
         sealed->TestGetSegmentLoadInfo()->GetProto();
     milvus::OpContext op_ctx;
-    EXPECT_NO_THROW(sealed->Reopen(&op_ctx, stale_proto, stale_schema));
+    EXPECT_NO_THROW(
+        sealed->Reopen(&op_ctx, stale_proto, stale_schema, nullptr));
```
liliu-z: The only 4-arg Reopen added at test_sealed.cpp:3846 passes nullptr for the replacement meta, so the non-null col_index_meta refresh — the PR's core behavior — is never exercised, and no BM25 brute-force search after reopen exists anywhere in the file. A regression where the refreshed meta lacks the added field would still compile and pass CI. Add a test that reopens with a non-null meta introducing a field, then runs a BM25 brute-force search on that field.

internal/querynodev2/segments/segment.go
MrPresent-Han: Fixed. `applyLoadUpdate` now fences the index meta with the same load barrier that fences the schema: `updateIndexMeta` is applied only when the incoming `schema_barrier_ts >=` the collection's current barrier (captured before the schema update advances it). A delayed pre-DDL load (older barrier) can therefore no longer overwrite the fresher meta, while a same-barrier index-only update (createIndex does not bump the barrier) is not strictly older and still applies, so newly indexed fields stay visible. I used `schema_barrier_ts` (the purpose-built load fence = `collection.UpdateTimestamp`, set at RootCoord DDL time), not the logical schema version. Added `TestPutOrRefStaleBarrierDoesNotClobberIndexMeta`.

Scope note: this fences the cross-schema-version staleness (the add-field case here, which bumps the barrier). A same-barrier index-only race (createIndex on an existing field, no DDL) cannot be ordered by any current signal — `CollectionIndexMeta`/`FieldIndexMeta` carry no version — so it is pre-existing and would need an index-meta version to close. Tracking that separately rather than widening this PR.

internal/core/unittest/test_sealed.cpp
MrPresent-Han: Added `SealedSegmentReopen.ReopenWithNonNullIndexMetaPublishesNewFieldIndex` (test_sealed.cpp): a sealed segment is reopened through the 4-arg path with a **non-null** `CollectionIndexMeta` that introduces a new field's index, and the test asserts the published `col_index_meta->HasField(new_field)` is false before and true after the reopen — so a regression where the refreshed meta drops the added field now fails here. It also traverses `generate_interim_index`'s schema+meta-same-snapshot path for the new field.

On the BM25 brute-force search specifically: there is no existing BM25 sealed-search harness in the C++ unit tests to build on, and an empty `index_info` would throw `std::out_of_range` at `SearchBruteForce` (`index_info.at(BM25_K1)`) rather than degrade — so the meaningful, reliable guard for the regression you flagged ("refreshed meta lacks the added field") is the HasField-on-published-snapshot assertion above. I can add a full BM25-search test as a follow-up if you'd prefer that too.

internal/querynodev2/segments/collection.go
```diff
@@ -439,14 +451,25 @@ func (c *Collection) applyLoadUpdate(schema *schemapb.CollectionSchema, meta *se
...
-	// for search plan creation (CollectionIndexMeta::HasField check).
-	if err := c.updateIndexMeta(meta); err != nil {
-		return collectionSchemaUpdatePlan{}, false, err
+	// Fence stale index meta with the same load barrier that fences the schema:
+	// barrierTs advances on the collection DDL that adds the field, so a load
+	// whose barrier predates the applied state is stale and must not clobber the
+	// fresher meta. A same-barrier index-only update (createIndex does not bump
+	// barrierTs) is not stale and still applies, keeping newly indexed fields
+	// visible for search plan creation (CollectionIndexMeta::HasField check).
+	if schemaBarrierTs >= currentBarrierTs {
```
liliu-z: The index-meta fence added here gates the meta write on `schemaBarrierTs` (the DDL barrier), but the barrier is also advanced by schema-only property updates (`UpdateSchema` -> `applySchemaUpdate` -> `setSchema`) that never carry index meta, while the schema payload itself is gated by the independent `schema.Version` domain. A load can therefore have its schema accepted yet its meta fenced out (`schemaBarrierTs < currentBarrierTs`) once an unrelated schema-only DDL has pushed the barrier past it, leaving the collection with a schema that has the field but an index meta that lacks it. The consequence is not the `vector_search` assert and not a node crash: a BM25 search fails gracefully with `FieldNotLoaded` at plan creation (`plan_c.cpp:49`) until a later load with a high-enough barrier refreshes the meta. Consider co-versioning the meta with the schema version (or a dedicated meta generation) rather than the shared barrier.

MrPresent-Han: Agreed — the barrier is the wrong domain to fence index meta, exactly as you describe: `schema_barrier_ts` is advanced by schema-only property DDLs that carry no index meta, while the schema payload is gated by the independent `schema.Version` domain, so a load can have its schema accepted yet its meta fenced out and end up with a schema that has the field but an index meta that lacks it (graceful `FieldNotLoaded` at plan creation, not a crash). And it cannot order the same-barrier index-only case (`createIndex` bumps neither) anyway.

I've **reverted** the barrier fence and restored the unconditional `updateIndexMeta` (removed `TestPutOrRefStaleBarrierDoesNotClobberIndexMeta`). Unconditional is strictly safer than the barrier fence here: a good meta always lands, and only a later out-of-order stale load can transiently overwrite it (last in-order load wins), rather than a good meta being permanently fenced out by an unrelated schema-only DDL.

Properly closing both the out-of-order stale clobber and the same-version index-only race needs `CollectionIndexMeta` to carry its own monotonic version/generation (co-versioning the meta, as you suggest), which is a proto change out of scope for this reopen-sync fix. Tracked as a follow-up: #51584.

internal/querynodev2/segments/segment.go
liliu-z: Blocking: this can turn a temporary stale collection meta into a persistent bad segment. A fresh reopen for segment A can finish PutOrRef(S2, M2), then a delayed load for segment B overwrites the collection meta with M1 (its stale schema is rejected, but its meta is still applied), and A reopens with current D2/S2 plus stale M1. QueryCoord sees A at the latest manifest/DataVersion, so later loads restore only the collection meta and do not reopen A. Please avoid applying/propagating stale meta, or pass the accepted request snapshot into Reopen.

internal/querynodev2/segments/collection.go
```diff
@@ -362,6 +362,18 @@ func (c *Collection) GetCCollection() *segcore.CCollection {
...
+// GetIndexMeta returns the collection-level index meta, or nil if the
+// collection has been released.
+func (c *Collection) GetIndexMeta() *segcorepb.CollectionIndexMeta {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.ccollection == nil {
+		return nil
+	}
+	return c.ccollection.IndexMeta()
```
liliu-z: Blocking: this accessor is locked, but Reopen reads schema first and meta separately without schemaTransitionMu.RLock(). Since applyLoadUpdate stores the new schema before updating meta, Reopen can observe S2/M1 during a normal S1/M1 -> S2/M2 update. Please return schema/version/meta from one snapshot, or hold the transition read lock across both reads.

internal/querynodev2/segments/segment.go
MrPresent-Han: Agreed — you're right that this is persistent, not transient (my earlier "last in-order load wins / self-heals" framing was wrong: once a segment reopens with the clobbered meta, QueryCoord sees it at the latest manifest/DataVersion and never re-reopens it, and later loads only restore the collection meta).

Fixed in 15c8c44 by co-versioning the index meta with the schema version, i.e. fencing the meta write in `applyLoadUpdate` on the same domain that gates the schema payload:

```go
_, currentVersion, _, _ := c.schemaSnapshotWithSegcoreSchemaVersion() // captured before the update advances it
...
if logicalSchemaVersion >= currentVersion {
    if err := c.updateIndexMeta(meta); err != nil { ... }
}
```

`logicalSchemaVersion` is `schema.GetVersion()` — the exact domain `prepareCollectionSchemaUpdate` uses to reject a stale schema (`logical < current`). So a strictly-older pre-DDL load whose schema is rejected now has its meta rejected too → it can no longer clobber the fresher collection meta → no persistent "schema has field / meta lacks field" segment. Because both gates live in the same version domain, this is *not* the barrier cross-domain fence you flagged on 07-18 (schema-accept ⟹ meta-accept; no "schema accepted but meta fenced" hole).

Added `TestPutOrRefStaleSchemaVersionDoesNotClobberIndexMeta` (v2-with-field load, then stale v1 load, asserts the field survives).

Scope note: this closes the cross-schema-version (add-field) clobber. A same-`schema.Version` out-of-order index-only race (`createIndex` bumps neither version nor barrier) still can't be ordered by any current signal — tracked in #51584 (needs a dedicated monotonic meta version).

internal/querynodev2/segments/collection.go
MrPresent-Han: Fixed in 15c8c44. Replaced `GetIndexMeta` with a single-snapshot accessor `SchemaVersionAndIndexMeta()` that reads schema, segcore version, and index meta under one `schemaTransitionMu.RLock()`:

```go
func (c *Collection) SchemaVersionAndIndexMeta() (*schemapb.CollectionSchema, uint64, *segcorepb.CollectionIndexMeta) {
    c.schemaTransitionMu.RLock()
    defer c.schemaTransitionMu.RUnlock()
    schema, _, _, segcoreSchemaVersion := c.schemaSnapshotWithSegcoreSchemaVersion()
    var indexMeta *segcorepb.CollectionIndexMeta
    c.mu.RLock()
    if c.ccollection != nil { indexMeta = c.ccollection.IndexMeta() }
    c.mu.RUnlock()
    return schema, segcoreSchemaVersion, indexMeta
}
```

`applyLoadUpdate` holds `schemaTransitionMu.Lock()` across both writes (`setSchema` then `updateIndexMeta`), so a reader under `RLock` can only observe an all-old or all-new pair — the `S2/M1` torn read is gone. `Reopen` (segment.go) now uses this accessor instead of the previous two separate reads (`SchemaAndSegcoreVersion` + `GetIndexMeta`). Lock order is consistent (`schemaTransitionMu → c.mu`), no reversal.

internal/querynodev2/segments/collection.go
```diff
@@ -439,14 +460,22 @@ func (c *Collection) applyLoadUpdate(schema *schemapb.CollectionSchema, meta *se
...
-	// Always update index meta to ensure newly indexed fields are visible
-	// for search plan creation (CollectionIndexMeta::HasField check).
-	if err := c.updateIndexMeta(meta); err != nil {
-		return collectionSchemaUpdatePlan{}, false, err
+	// Co-version the index meta with the same schema-version domain that gates the
+	// schema payload: a strictly-older load (logicalSchemaVersion < current) is
+	// stale and must not clobber the fresher meta, while a same-version load
+	// (e.g. createIndex, which does not bump the schema version) still refreshes
+	// newly indexed fields for search plan creation (CollectionIndexMeta::HasField).
+	if logicalSchemaVersion >= currentVersion {
```
liliu-z: Schema version cannot safely order index meta here. QueryCoord fetches the schema and ListIndexes separately, so a request can be S2/M1 or S1/M2. This condition still accepts stale M1 when the schema versions match, and can drop fresh M2 when the request carries an older schema. Please use an index-meta generation, or keep the old behavior and handle meta ordering in #51584.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7177,6 +7198,8 @@ ChunkedSegmentSealedImpl::Reopen(
     }
 
     auto target_schema = new_schema ? std::move(new_schema) : current_schema;
+    auto target_index_meta =
```
liliu-z, line 7414: Blocking: a search can move this segment to a newer schema via `LazyCheckSchema` while this reopen is queued. The stale-schema check above then returns success without applying the new load info/index meta, but Go still records the new `DataVersion`. Can we keep the current schema and apply D/M, or return a retryable error instead?

internal/querynodev2/segments/collection.go
MrPresent-Han: You're right — reverted the version gate in a6f376c835. `applyLoadUpdate` is back to unconditional `updateIndexMeta` (with a comment pointing at #51584); the `SchemaVersionAndIndexMeta` atomic accessor (the torn-read fix) stays. Removed the now-moot `TestPutOrRefStaleSchemaVersionDoesNotClobberIndexMeta`.

Confirmed by a code trace that no schema-derived signal can order the meta here, and QueryCoord provides no fallback:
- executor fetches schema (`getMetaInfo`) and `ListIndexes` (`getLoadInfo`) as two separate RPCs at run time (`executor.go:239`/`:244`) -> a request can be S2/M1 or S1/M2, exactly as you said; the gate accepted stale M1 and dropped fresh M2.
- scheduler dedup is per `(ReplicaID, SegmentID, IsGrowing)` (`scheduler.go:70-83`) -> cross-segment collection-meta writes aren't deduped; `checkStale` (`:1398`) is topology-only, no version check.

So meta ordering (S2/M1 + S1/M1 clobber, and the same-version index-only race) is deferred to #51584 (dedicated meta generation carried end-to-end). This PR now only closes the reopen-refresh path (#51557) and the torn read. Updated #51584 with this detail.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
MrPresent-Han: Agreed — this is the stale-schema reopen race, and a real Blocking bug: a reopen queued before `LazyCheckSchema` advances the segment would hit the stale-schema branch, return success without applying the new load info / index meta, while Go still records the new `DataVersion`, stranding the segment with no retrigger.

Fixed with your first option — the branch no longer returns early: it clamps to the current schema (never rolls it back) and still applies the reopen's load info + index meta. Index meta needs no clamp of its own (`Reopen` is its only serialized writer, which `LazyCheckSchema` never races), so only the schema is clamped.

```cpp
auto target_schema =
    (new_schema && new_schema->get_schema_version() >=
                       current_schema->get_schema_version())
        ? std::move(new_schema)
        : current_schema;
// ... falls through and applies load info + index meta
```

The `Skip stale reopen` early return is gone. Folded into the reopen-sync commit (current head `930fc2c2`).

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7361,30 +7381,34 @@ void
...
+    // them while reporting success, stranding the segment on stale data/meta
+    // with no retrigger. Only the schema needs clamping: index meta has a single
+    // serialized writer (Reopen) that LazyCheckSchema never races, so it cannot
+    // roll back and needs no guard of its own.
+    auto target_schema =
+        (new_schema && new_schema->get_schema_version() >=
+                           current_schema->get_schema_version())
+            ? std::move(new_schema)
+            : current_schema;
+    auto target_index_meta =
```
liliu-z, line 7414: Correction after checking the QC scheduler: the normal QC path deduplicates active tasks by (ReplicaID, SegmentID, IsGrowing), so two reopens for the same segment should not overtake each other. The remaining question is only cross-segment convergence: segment A may finish with V10 after another segment advances the shared CCollection to V20. If the data/index checker guarantees A will always get another reopen for every relevant meta change, no segment-level generation is needed here.

internal/querynodev2/segments/segment.go
```diff
@@ -1303,11 +1303,15 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
...
-	schema, schemaVersion := s.collection.SchemaAndSegcoreVersion()
+	schema, schemaVersion, indexMeta := s.collection.SchemaVersionAndIndexMeta()
 	err := s.csegment.Reopen(ctx, &segcore.ReopenRequest{
 		LoadInfo:      newLoadInfo,
 		Schema:        schema,
 		SchemaVersion: schemaVersion,
+		// Schema and index meta come from one transition-lock snapshot so a field
+		// added after this segment was created (e.g. a backfilled BM25 sparse
+		// field) is searchable after reopen without observing a torn pair.
+		IndexMeta: indexMeta,
```
liliu-z, line 1314: Correction: QC already deduplicates one active task per replica/segment, so this QN-side coalescing is not a blocker on the normal QC path. Please ignore this comment unless another producer can send concurrent reopens for the same segment.

internal/querynodev2/segments/collection.go
```diff
@@ -435,18 +460,35 @@ func (c *Collection) applySchemaUpdate(schema *schemapb.CollectionSchema, logica
...
-	// Always update index meta to ensure newly indexed fields are visible
-	// for search plan creation (CollectionIndexMeta::HasField check).
-	if err := c.updateIndexMeta(meta); err != nil {
-		return collectionSchemaUpdatePlan{}, false, err
+	// Index meta carries its own coordinator-stamped version (#51584),
+	// independent of schema.Version and schemaBarrierTs, so a stale load cannot
+	// clobber fresher index meta. Version 0 marks a legacy unstamped payload
+	// and keeps the unconditional refresh. Equal versions skip silently: every
+	// load re-sends the current version, so that is the steady state.
+	if indexMetaVersion == 0 || indexMetaVersion > c.indexMetaVersion {
```
liliu-z: Blocking: v0 can leave the content and watermark inconsistent. forwardL0RemoteLoad rebuilds LoadSegmentsRequest without LoadMeta, so a delayed M10 arrives as v0, overwrites M20, but c.indexMetaVersion stays V20. The next M20/V20 is then skipped as equal, so this sticks until V21. Please forward LoadMeta on every path and do not let v0 roll back a collection once it has seen a stamped version.

internal/querycoordv2/observers/target_observer.go
```diff
@@ -578,10 +579,11 @@ func (ob *TargetObserver) syncToDelegator(ctx context.Context, replica *meta.Rep
...
 		LoadMeta: &querypb.LoadMetaInfo{
-			LoadType:      ob.meta.GetLoadType(ctx, LeaderView.CollectionID),
-			CollectionID:  LeaderView.CollectionID,
-			PartitionIDs:  partitions,
-			ResourceGroup: replica.GetResourceGroup(),
+			LoadType:         ob.meta.GetLoadType(ctx, LeaderView.CollectionID),
+			CollectionID:     LeaderView.CollectionID,
+			PartitionIDs:     partitions,
+			ResourceGroup:    replica.GetResourceGroup(),
+			IndexMetaVersion: indexMetaVersion,
```
liliu-z, line 586: Blocking: this version is sent but not consumed for SyncType_UpdateVersion; QueryNode only calls SyncTargetVersion and ignores req.IndexInfoList/LoadMeta. Also, delegator sync returns early when IndexInfoList is empty, so deleting the last index is not applied there. Please make this a real reconcile path and treat version > 0 + empty list as an authoritative empty snapshot.

internal/core/unittest/test_sealed.cpp
```diff
@@ -3974,13 +3974,73 @@ TEST(SealedSegmentReopen, SchemaAwareReopenDiscardsOlderSchema) {
...
+                 before->col_index_meta->HasField(new_field));
+
+    proto::segcore::SegmentLoadInfo proto =
+        sealed->TestGetSegmentLoadInfo()->GetProto();
+    milvus::OpContext op_ctx;
+    EXPECT_NO_THROW(sealed->Reopen(&op_ctx, proto, new_schema, new_index_meta));
+
+    auto after = sealed->TestGetPublishedStateSnapshot();
+    ASSERT_NE(after->col_index_meta, nullptr);
+    EXPECT_TRUE(after->col_index_meta->HasField(new_field))
```
liliu-z, line 4101: This only checks HasField on a dense FLAT meta. It does not exercise the actual failure: BM25 brute-force reading k1/b, interim-index generation, or the Go -> cgo -> C++ path. Please add a real sparse BM25 reopen/search test and an L0/version-0 regression test.

liliu-z (commented):
A few blocking correctness issues I found:

internal/querynodev2/segments/collection.go
MrPresent-Han: fixed

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
MrPresent-Han: Agreed with the correction. Verified the same: QC `preAdd` dedups active tasks by (ReplicaID, SegmentID, IsGrowing), and `LocalSegment.Reopen` fresh-reads `SchemaVersionAndIndexMeta()` at execution time, so a serialized reopen always publishes the current collection meta — a segcore-level index-meta generation is defense-in-depth, not required on the normal path. Tracking a dedicated segcore generation as follow-up.

internal/querynodev2/segments/segment.go
MrPresent-Han: Agreed, ignoring per your correction. Reopen is dispatched as a SegmentTask (segment_checker → NewSegmentAction(ActionTypeReopen) → NewSegmentTask → scheduler.Add → preAdd), so QC never runs two concurrent reopens for the same segment; the QN-side isLoading skip / delegator singleflight are therefore not hit by same-segment reopens on the normal path. Deferred.

internal/querynodev2/segments/collection.go
MrPresent-Han: Fixed in 6ba9077754. Two parts: (a) `forwardL0RemoteLoad` now forwards `LoadMeta`, so the L0 delta-forward carries the coordinator-stamped `index_meta_version` instead of arriving as v0; (b) `applyLoadUpdate` no longer lets a v0 payload overwrite the meta content once a stamped version has been applied — v0 refreshes unconditionally only on an all-legacy cluster (c == 0). Together the collection meta can no longer be rolled back to a stale v0 snapshot while the watermark stays ahead.

internal/querycoordv2/observers/target_observer.go
MrPresent-Han: Fixed in 6ba9077754. (1) `SyncDistribution` `SyncType_UpdateVersion` now reconciles the collection index meta (`PutOrRef` with the node schema + `req.IndexInfoList` + `LoadMeta`) before acking the sync, so the watermark advances and altered params land even when no segment reopen is triggered. (2) A version>0 empty `IndexInfoList` is now treated as an authoritative empty snapshot on `LoadSegments`, `WatchDmChannels`, and the delegator `syncCollectionIndexMeta`, so dropping the last index (and channel resubscribe on recovery) is applied instead of rejected with ErrIndexNotFound.

internal/core/unittest/test_sealed.cpp
MrPresent-Han: Acknowledged. The current UT asserts the published `col_index_meta` has the added field after reopen (guarding the meta-refresh regression), but it does not run a BM25 sparse brute-force search through the Go → cgo → C++ path — that needs new sparse-search UT infra with no existing model, tracked as follow-up. The end-to-end behavior is covered by a live repro (below-threshold sealed segments become BM25-searchable after add_function_field backfill; `field index meta not found` never observed over the full window).

PR #51559, opened 2026-07-17
fix: make release manual flush local only

chyezh:
issue: #51332

- streaming local WAL: expose release manual flush through Local().PrepareReleaseManualFlushIfLocal and remove the remote handler retry path
- QueryNode unsubscribe: treat local-only prepare errors as unavailable so channel release can continue

### DIFF
--- a/internal/distributed/streaming/local.go
+++ b/internal/distributed/streaming/local.go
@@ -20,6 +20,15 @@ func (w localServiceImpl) GetLatestMVCCTimestampIfLocal(ctx context.Context, vch
 	return w.handlerClient.GetLatestMVCCTimestampIfLocal(ctx, vchannel)
 }
 
+func (w localServiceImpl) PrepareReleaseManualFlushIfLocal(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error) {
+	if !w.lifetime.Add(typeutil.LifetimeStateWorking) {
+		return false, ErrWALAccesserClosed
+	}
+	defer w.lifetime.Done()
+
+	return w.handlerClient.PrepareReleaseManualFlushIfLocal(ctx, collectionID, vchannel, releaseSegmentIDs)
+}
+
 // GetMetrics gets the metrics of the wal.
 func (w localServiceImpl) GetMetricsIfLocal(ctx context.Context) (*types.StreamingNodeMetrics, error) {
 	if !w.lifetime.Add(typeutil.LifetimeStateWorking) {
--- a/internal/distributed/streaming/streaming.go
+++ b/internal/distributed/streaming/streaming.go
@@ -173,11 +173,6 @@ type WALAccesser interface {
 	// Local returns the local services.
 	Local() Local
 
-	// PrepareReleaseManualFlush prepares process-local release handoff.
-	// Returns false when the current process is not the local flush owner or
-	// the channel does not need growing-source retention.
-	PrepareReleaseManualFlush(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error)
-
 	// RawAppend writes a records to the log.
 	RawAppend(ctx context.Context, msgs message.MutableMessage, opts ...AppendOption) (*types.AppendResult, error)
 
@@ -203,6 +198,10 @@ type Local interface {
 	// GetLatestMVCCTimestampIfLocal gets the latest mvcc timestamp of the vchannel.
 	// If the wal is located at remote, it will return 0, error.
 	GetLatestMVCCTimestampIfLocal(ctx context.Context, vchannel string) (uint64, error)
+
+	// PrepareReleaseManualFlushIfLocal prepares process-local release handoff.
+	// If the wal is located at remote, it will return false, error.
+	PrepareReleaseManualFlushIfLocal(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error)
 
 	// GetMetricsIfLocal gets the metrics of the local wal.
 	// It will only return the metrics of the local wal but not the remote wal.
--- a/internal/distributed/streaming/test_streaming.go
+++ b/internal/distributed/streaming/test_streaming.go
@@ -117,6 +117,10 @@ func (n *noopLocal) GetLatestMVCCTimestampIfLocal(ctx context.Context, vchannel 
 	return 0, errors.New("not implemented")
 }
 
+func (n *noopLocal) PrepareReleaseManualFlushIfLocal(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error) {
+	return false, getExpectErr()
+}
+
 func (n *noopLocal) GetMetricsIfLocal(ctx context.Context) (*types.StreamingNodeMetrics, error) {
 	return &types.StreamingNodeMetrics{}, nil
 }
@@ -186,10 +190,6 @@ func (n *noopWALAccesser) WALName() string {
 
 func (n *noopWALAccesser) Local() Local {
 	return &noopLocal{}
-}
-
-func (n *noopWALAccesser) PrepareReleaseManualFlush(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error) {
-	return false, getExpectErr()
 }
 
 func (n *noopWALAccesser) RawAppend(ctx context.Context, msgs message.MutableMessage, opts ...AppendOption) (*types.AppendResult, error) {
--- a/internal/distributed/streaming/wal.go
+++ b/internal/distributed/streaming/wal.go
@@ -83,10 +83,6 @@ func (w *walAccesserImpl) Local() Local {
 	return localServiceImpl{w}
 }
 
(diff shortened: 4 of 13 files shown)


czs007:

Reviewed at head `8d8708cc10` against `master`. The direction is right — making the prepare strictly local, guarding it with the `Local()` accessor lifetime (`ErrWALAccesserClosed` is an on-shutdown `StreamingError`, so it is already covered by the QueryNode classifier), and dropping the remote `createHandlerAfterStreamingNodeReady` retry loop all look good. However, the rewrite drops one error class that the old implementation explicitly tolerated, and the second commit introduces an unbounded wait inside a release RPC. I'd consider the first one blocking.

## High — wrong-streaming-node / term-mismatch errors now fail channel release

`internal/streamingnode/client/handler/handler_client_impl.go:167`

```go
w, err := registry.GetLocalAvailableWAL(assign.Channel)
if err != nil {
    return false, err
}
```

The removed code handled this case deliberately:

```go
if errors.Is(err, registry.ErrNoStreamingNodeDeployed) ||
    status.AsStreamingError(err).IsWrongStreamingNode() {
    logger.Info(ctx, "skip release manual flush prepare because channel is not owned by local streaming node", ...)
    return false, nil
}
```

`GetLocalAvailableWAL` reaches `walmanager.managerImpl.GetAvailableWAL` (`internal/streamingnode/server/walmanager/manager_impl.go:105-112`), which returns `status.NewChannelNotExist(...)` when the WAL lifetime is not open yet, and `status.NewUnmatchedChannelTerm(...)` when the watcher's assignment term differs from the locally opened WAL term. Both are `IsWrongStreamingNode()` per `internal/util/streamingutil/status/streaming_error.go:34-37`, and **neither** is matched by `isReleaseManualFlushPrepareUnavailable` (`internal/querynodev2/services.go:730-739`), which only checks the three sentinels, the two merr codes, and `IsOnShutdown()`.

Failure scenario (standalone / embedded, i.e. whenever the StreamingNode role is local — in a pure cluster QueryNode the role is not local, so `GetLocalAvailableWAL` short-circuits to `ErrNoStreamingNodeDeployed`, which is handled): a channel is released during the window where the assignment watcher already holds the pchannel assignment but the local WAL lifetime has not finished opening (startup), or right after a term bump when the watcher still serves the previous term. `PrepareReleaseManualFlushIfLocal` returns `ChannelNotExist` / `UnmatchedChannelTerm` → `isReleaseManualFlushPrepareUnavailable` returns false → `UnsubDmChannel` returns `merr.Status(err)` (`internal/querynodev2/services.go:393-395`) → QueryCoord retries the channel release, and it stalls until the term/WAL state converges. Before this PR the same window returned `(false, nil)` and the unsubscribe proceeded.

Fix: restore the tolerance at one of the two ends — either in the handler client

```go
w, err := registry.GetLocalAvailableWAL(assign.Channel)
if err != nil {
    if status.AsStreamingError(err).IsWrongStreamingNode() {
        return false, ErrReadOnlyWAL // or a dedicated ErrNotLocalWAL sentinel
    }
    return false, err
}
```

or in the classifier by adding `streamingstatus.AsStreamingError(err).IsWrongStreamingNode()` alongside the existing `IsOnShutdown()` check. The classifier end is preferable if you want `GetLatestMVCCTimestampIfLocal` to keep surfacing these errors unchanged. Please add a test covering it.

## Medium — unbounded wait for assignment inside the release path, and it blocks handler-client Close

`internal/streamingnode/client/handler/handler_client_impl.go:161-166` (commit `8d8708cc10`)

```go
assign := hc.watcher.Get(ctx, pchannel)
for assign == nil {
    if err := hc.watcher.Watch(ctx, pchannel, nil); err != nil {
        return false, err
    }
    assign = hc.watcher.Get(ctx, pchannel)
}
```

`watcherImpl.Watch` (`internal/streamingnode/client/handler/assignment/watcher_impl.go:81-101`) blocks on `cond.Wait(ctx)` and only returns when the assignment term advances or the **caller's** ctx is done. `watcher.Close()` cancels the background discovery but does not wake cond waiters. Meanwhile this function holds `hc.lifetime`, and `handlerClientImpl.Close()` (`:371-378`) does `lifetime.Wait()` *before* `hc.watcher.Close()`.

Failure scenario: a `UnsubDmChannel` arrives while assignment discovery has not yet produced the pchannel (early startup, or the owning streaming node is down). The RPC blocks for its full deadline instead of proceeding with the release, and if the process is shutting down at that moment, `handlerClient.Close()` blocks until that RPC's ctx expires. It is also inconsistent with the sibling `GetLatestMVCCTimestampIfLocal` (`:55-57`), which returns `ErrClientAssignmentNotReady` immediately.

Fix: bound the wait, e.g. derive a short ctx (`context.WithTimeout`) for the loop and fall back to `ErrClientAssignmentNotReady` on expiry. The QueryNode already classifies that sentinel as unavailable, so the release continues instead of stalling — which is exactly the intent of the classifier.

## Low — `ErrClientAssignmentNotReady` is now unreachable from this path, but is still asserted as handled

`internal/querynodev2/services.go:733` and `internal/querynodev2/services_test.go:594-598`

Commit `94e7f52eb0` returned `ErrClientAssignmentNotReady` when `assign == nil`; commit `8d8708cc10` replaced that return with the wait loop. The classifier entry and its table-test case survive, so `TestIsReleaseManualFlushPrepareUnavailable` now asserts a case that `PrepareReleaseManualFlushIfLocal` can never produce. The test reads as "assignment-not-ready lets the release continue", while the real behavior is "the release blocks". Adopting the bounded-wait fallback above makes the branch live again; otherwise the entry and the case should be dropped.

## Low — the read-only branch is not actually covered

`internal/streamingnode/client/handler/handler_client_test.go:261-293`

`TestHandlerClient_PrepareReleaseManualFlushIfLocal` builds an `AccessModeRO` assignment but only asserts `assert.Error(t, err)`. At that point no local WAL manager is registered and the StreamingNode role is not enabled, so the call fails at `GetLocalAvailableWAL` with `ErrNoStreamingNodeDeployed` and never reaches the `w.Channel().AccessMode != types.AccessModeRW` branch at `handler_client_impl.go:172`. The result is that `ErrReadOnlyWAL` — newly added to the QueryNode classifier — has no test proving this function ever returns it. Suggest registering a WAL manager that returns an RO WAL and asserting `assert.ErrorIs(t, err, ErrReadOnlyWAL)`; at minimum assert the concrete error rather than "any error".

## Low — global test state is mutated without cleanup

`internal/streamingnode/client/handler/handler_client_test.go:297-299`

```go
paramtable.SetLocalComponentEnabled(typeutil.StreamingNodeRole)
registry.RegisterLocalWALManager(shutdownWALManager{})
```

Both are process-global and never reset. `registry` is a `syncutil.Future`; `Set` closes its channel (a second `Set` panics with "close of closed channel"), and `Get()` blocks forever when the role is enabled but nothing is registered. Consequences: (a) every later test in package `handler` now sees a shutting-down local WAL manager instead of "no streaming node deployed" — `TestHandlerClient_GetReplicateCheckpointReplicateViolation` currently survives only because `shouldUseRemoteWAL` (`:397-408`) treats `IsOnShutdown()` the same as `ErrNoStreamingNodeDeployed`; (b) the suite becomes order-dependent — under `-shuffle=on`, or if any future test enabling the role lands before a WAL manager is registered, `registry.Get()` deadlocks. `registry.ResetRegisterLocalWALManager()` already exists in `registry/test_utility.go`; please call it (plus reset the role) from a `t.Cleanup`.

## Nit — the local prepare path lost all logging

The old implementation logged pchannel / vchannel / collectionID / releaseSegmentIDs and the skip reason. The new `PrepareReleaseManualFlushIfLocal` logs nothing, so a release that skips the prepare is now only visible via the QueryNode-side warn, without the WAL-side context (access mode, term, owning node). Consider keeping a debug/info line on the skip paths.

chyezh:
Thanks for the detailed review. I rechecked these points against the latest head.

The blocking part has been addressed in follow-up commits:

- `isReleaseManualFlushPrepareUnavailable` now treats `StreamingError.IsWrongStreamingNode()` as unavailable, so `CHANNEL_NOT_EXIST` and `UNMATCHED_CHANNEL_TERM` from the local WAL manager no longer fail QueryNode channel release.
- `ErrClientAssignmentNotReady` was removed from the release-prepare classifier, because `PrepareReleaseManualFlushIfLocal` no longer returns it after the assignment-ready wait was added.
- The classifier test now covers both `NewChannelNotExist` and `NewUnmatchedChannelTerm`.

For the assignment wait: this is intentional for this API. The desired semantic is local-only, but it should wait for the assignment watcher to become ready before deciding whether the WAL is local. The wait is still bounded by the caller's RPC context/deadline, and the path does not re-enter the remote handler/retry flow.

The remaining points are test/logging quality nits rather than correctness blockers:

- the RO branch test can be made more precise by registering a local RO WAL and asserting `ErrReadOnlyWAL`;
- the shutdown-WAL test mutates local registry/global role state and can be made cleaner with registry cleanup;
- the logging point is optional, since QueryNode logs the unavailable prepare result and the local preparer logs actual prepare/skip results.

Given the current fix scope, I don't think more production-code changes are needed for this PR.

liliu-z (commented):
/lgtm
/approve

PR #51560, opened 2026-07-17
fix: [cp 3.0] make release manual flush local only

chyezh:
issue: #51332
pr: #51559

- streaming local WAL: expose release manual flush through Local().PrepareReleaseManualFlushIfLocal and remove the remote handler retry path
- QueryNode unsubscribe: treat local-only prepare errors as unavailable so channel release can continue

### DIFF
--- a/internal/distributed/streaming/local.go
+++ b/internal/distributed/streaming/local.go
@@ -20,6 +20,15 @@ func (w localServiceImpl) GetLatestMVCCTimestampIfLocal(ctx context.Context, vch
 	return w.handlerClient.GetLatestMVCCTimestampIfLocal(ctx, vchannel)
 }
 
+func (w localServiceImpl) PrepareReleaseManualFlushIfLocal(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error) {
+	if !w.lifetime.Add(typeutil.LifetimeStateWorking) {
+		return false, ErrWALAccesserClosed
+	}
+	defer w.lifetime.Done()
+
+	return w.handlerClient.PrepareReleaseManualFlushIfLocal(ctx, collectionID, vchannel, releaseSegmentIDs)
+}
+
 // GetMetrics gets the metrics of the wal.
 func (w localServiceImpl) GetMetricsIfLocal(ctx context.Context) (*types.StreamingNodeMetrics, error) {
 	if !w.lifetime.Add(typeutil.LifetimeStateWorking) {
--- a/internal/distributed/streaming/streaming.go
+++ b/internal/distributed/streaming/streaming.go
@@ -173,11 +173,6 @@ type WALAccesser interface {
 	// Local returns the local services.
 	Local() Local
 
-	// PrepareReleaseManualFlush prepares process-local release handoff.
-	// Returns false when the current process is not the local flush owner or
-	// the channel does not need growing-source retention.
-	PrepareReleaseManualFlush(ctx context.Context, collectionID int64, vchannel string, releaseSegmentIDs []int64) (bool, error)
-
 	// RawAppend writes a records to the log.
 	RawAppend(ctx context.Context, msgs message.MutableMessage, opts ...AppendOption) (*types.AppendResult, error)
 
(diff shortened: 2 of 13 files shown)


tinswzy:

Reviewed the two picked commits against base 3.0 (local-only release manual flush + wait-for-assignment-ready). The refactor is clean and the mock regeneration is consistent, but there is one real behavior regression in the error taxonomy that the caller does not compensate for. Non-blocking-otherwise; I'd fix the High before merge.

### High — `IsWrongStreamingNode` errors from the local WAL are no longer treated as "skip and continue release"

`internal/streamingnode/client/handler/handler_client_impl.go:168-174`
```go
w, err := registry.GetLocalAvailableWAL(assign.Channel)
if err != nil {
    return false, err          // <- raw streaming error propagated
}
if w.Channel().AccessMode != types.AccessModeRW {
    return false, ErrReadOnlyWAL
}
```

The removed old code explicitly swallowed the wrong-node class:
```go
if errors.Is(err, registry.ErrNoStreamingNodeDeployed) ||
    status.AsStreamingError(err).IsWrongStreamingNode() {
    return false, nil   // skip prepare, release proceeds
}
```

`GetLocalAvailableWAL` -> `managerImpl.GetAvailableWAL` (`internal/streamingnode/server/walmanager/manager_impl.go:106,111`) returns `NewChannelNotExist` and `NewUnmatchedChannelTerm`, and both are `IsWrongStreamingNode() == true` (`internal/util/streamingutil/status/streaming_error.go:34-37`). These are exactly the transient states that occur during channel handoff/rebalance — the watcher already holds the new RW assignment (term N) but the local WAL manager has not yet opened/promoted the WAL to term N, so `GetAvailableWAL` returns `CHANNEL_NOT_EXIST` (wal nil / not available) or `UNMATCHED_CHANNEL_TERM`.

Now trace the caller: `isReleaseManualFlushPrepareUnavailable` (`internal/querynodev2/services.go:730-739`) catches `ErrServiceUnavailable / ErrChannelNotAvailable / ErrClientClosed / ErrClientAssignmentNotReady / ErrReadOnlyWAL / ErrNoStreamingNodeDeployed / ErrNoReleaseManualFlushPreparer` and `IsOnShutdown()` — but NOT `IsWrongStreamingNode()`. So these two codes fall through to the `else` branch in `UnsubDmChannel` (`internal/querynodev2/services.go:394-398`) and the unsubscribe RPC returns an error:
```go
} else {
    log.Warn(ctx, "failed to prepare release manual flush before unsubscribing channel", ...)
    return merr.Status(err), nil   // unsub fails -> querycoord retries, release/balance stalls
}
```

Failure scenario: querycoord balances/releases a channel; on the streaming node the WAL for that pchannel is mid-handoff (term just bumped or briefly removed). `UnsubDmChannel` now returns `channel ... term ... is expected, but current term is ...` / `channel not exist` instead of continuing the release as it did before this PR. The release is retried until the local WAL settles — a regression from the prior "skip prepare, continue unsubscribe" semantics.

Note the asymmetry that makes this a clear miss: the author DID handle `IsOnShutdown` (line 739, with a dedicated test), which is a sibling error returned by the very same `GetAvailableWAL`, but left the other two codes that function returns uncovered.

Fix: extend the unavailable classifier to also cover the wrong-node class, e.g.
```go
se := streamingstatus.AsStreamingError(err)
return se.IsOnShutdown() || se.IsWrongStreamingNode()
```
or restore the explicit `IsWrongStreamingNode()` skip inside `PrepareReleaseManualFlushIfLocal` before returning. Please also add a table case to `TestIsReleaseManualFlushPrepareUnavailable` for `status.NewChannelNotExist(...)` / `status.NewUnmatchedChannelTerm(...)`.

### Low — `ErrClientAssignmentNotReady` in the unavailable list is dead for this path

`internal/querynodev2/services.go:733`

`PrepareReleaseManualFlushIfLocal` no longer returns `ErrClientAssignmentNotReady`; the new watch loop (`handler_client_impl.go:160-166`) blocks until an assignment appears instead of returning that sentinel (unlike `GetLatestMVCCTimestampIfLocal`, which still returns it). Harmless, but the added `errors.Is(err, handler.ErrClientAssignmentNotReady)` branch is unreachable for the prepare path now — either drop it or add a comment noting it's defensive.

### Low — unbounded wait on assignment readiness inside an RPC handler

`internal/streamingnode/client/handler/handler_client_impl.go:160-166`
```go
assign := hc.watcher.Get(ctx, pchannel)
for assign == nil {
    if err := hc.watcher.Watch(ctx, pchannel, nil); err != nil {
        return false, err
    }
    assign = hc.watcher.Get(ctx, pchannel)
}
```

This changes prepare from "return immediately if not ready" to "block until ready or ctx cancel." It correctly respects `ctx` (the `context.Canceled` test covers it), so it's safe, but if the assignment never arrives this now holds `UnsubDmChannel` for the full inbound RPC deadline where it previously returned fast. Given `UnsubDmChannel` is on the release critical path, consider a bounded wait so a stuck assignment can't pin the handler. Non-blocking.

### Nit
- `handler_client_impl.go:152` doc: "If the wal is located at remote, it will return false, error." In cluster mode the remote case actually surfaces as `ErrNoStreamingNodeDeployed`, and the RW check surfaces as `ErrReadOnlyWAL`; the comment is fine but slightly understates the concrete error set callers must classify.

Everything else (interface move `WALAccesser` -> `Local`, `test_streaming.go` noop, and all three regenerated mock files) is consistent and correct.

liliu-z (commented):
/lgtm
/approve

chyezh:
all ci has been passed

Issue #51564, opened 2026-07-17
[Bug]: Premature segment release and next-target churn can block multi-shard convergence


weiliu1031:
## Environment

- Milvus version: master
- Deployment mode: cluster
- Component: QueryCoord
- Collection layout: multiple shards

## Current Behavior

QueryCoord can observe different readable target versions across shards
while the current and next targets continue to advance.

For example:

- current target: V1
- shard ch1 delegator: V2
- shard ch2 delegator: V3
- next target: V4

Three behaviors interact in this state.

### 1. Segments may be released before all delegators leave them

The segment release check only considers a subset of delegator versions
as potentially using an old sealed segment.

A delegator whose target version is ahead of current may still expose a
readable snapshot that references the segment, but the segment can be
treated as releasable.

This can release data that is still referenced by one of the shards.

### 2. Periodic next-target replacement can prevent convergence

TargetObserver force-refreshes an existing next target after
`NextTargetSurviveTime`, even when segment loading is still progressing.

If next changes from V4 to V5 before all delegators reach V4, and this
continues periodically, no target generation can fully synchronize and
be promoted to current.

After applying a conservative release barrier, this target churn would
also prevent old segments from ever becoming releasable.

### 3. ErrSegmentNotFound bypasses TargetObserver

When a segment load task returns `ErrSegmentNotFound`, the task scheduler
directly updates TargetManager.

This bypasses TargetObserver's target progress state, stale-target state,
and collection-level target update sequencing.

## Expected Behavior

1. A sealed segment must remain in use while any delegator target version
   differs from the current target version.

2. A healthy next target must remain stable while the number of loaded
   target segments is increasing.

3. TargetObserver should force-refresh next only when:
   - no loading progress is observed for `NextTargetSurviveTime`; or
   - a load task explicitly reports that a target segment no longer
     exists.

4. A next target known to be stale must not be promoted to current.

5. Runtime next-target refreshes should be coordinated by TargetObserver
   instead of directly mutating TargetManager from the task scheduler.

## Steps To Reproduce

1. Load a collection with at least two shards.
2. Let the shard delegators report different target versions.
3. Keep current behind both delegator versions.
4. Create a newer next target that has not fully converged.
5. Allow the periodic next-target timeout to replace next repeatedly.
6. Observe that:
   - an old segment may be released while a delegator still references
     it; or
   - conservative release remains blocked because current never catches
     up.

A load task returning `ErrSegmentNotFound` also demonstrates the direct
[...]

PR #51565, opened 2026-07-17
fix: stabilize multi-shard target convergence

weiliu1031:
issue: [#51564](https://github.com/milvus-io/milvus/issues/51564)

## What changed

This PR stabilizes multi-shard target convergence while keeping sealed
segment release conservative.

### 1. Preserve partial target synchronization

The removed all-delegator
prepare gate and sync-start pin are not restored, so ready delegators
continue to synchronize independently.

### 2. Track replica-scoped next-target progress

Periodic refresh now treats the next target as progressing when either
of these high-water counters increases:

- `readySegmentReplicas`: unique `(replicaID, segmentID)` pairs whose
  observed distribution copies satisfy the target manifest and
  `DataVersion`;
- `readyReplicaChannels`: unique `(replicaID, channelName)` pairs with a
  delegator that is data-ready or already synchronized to the target
  version.

The counters are high-water marks within one target version, so readiness
oscillation cannot repeatedly reset the timeout. A new target version
rebuilds the baseline.

These counters are only liveness signals for deciding whether automatic
refresh should wait. Current-target promotion continues to use the
existing readiness path.

### 3. Route unloadable targets through TargetObserver

When a `SegmentTask` or segment-based `LeaderTask` finishes with
`ErrSegmentNotFound`, the scheduler reports both the collection and
segment to TargetObserver. Channel tasks do not trigger this path.

The observer marks the target stale only when that segment is still in
the current next target. This avoids coupling tasks to a target generation;
a concurrent replacement can cause at most one conservative extra refresh.

Promotion keeps commit-wins semantics. Once readiness and
`SyncDistribution` have started, a late stale event does not veto the
in-flight promotion. The stale revision remains available to drive the
subsequent next-target refresh.

### 4. Preserve stale events across recovery refreshes

The stale boolean is replaced by a per-collection revision. Before a
recovery RPC, the observer snapshots the next-target version and stale
revision.

- RPC failures keep the stale revision and progress baseline.
- Empty or no-op recovery results do not clear stale or reset progress.
- Progress is reset only after a newer next-target version is installed.
- Only the captured stale revision is cleared; an event arriving during
  the RPC increments the revision and survives the compare-and-delete.

### 5. Keep explicit manual refresh semantics

Automatic progress-aware refresh keeps a healthy next target stable while
loading advances. A stale target or an explicit `UpdateNextTarget` /
`UpdatePartition` request may intentionally replace it and restart the
transition. The SegmentChecker comment now states this narrower invariant.

## Why

A delegator on any target version other than current may still serve a
readable snapshot that references an obsolete sealed segment. Releasing
that segment early can break reads. At the same time, repeatedly replacing
a healthy next target can prevent convergence and keep old segments pinned
forever.

The new behavior keeps release safe, gives healthy transitions time to
make measurable progress, and refreshes targets that are stale or no longer
advancing.

## Risks

The conservative release barrier can temporarily increase QueryNode memory
usage while target convergence is in progress. Explicit manual refreshes
can restart an in-progress transition by design.

## Test coverage added

- delegators on non-current versions retaining sealed segments;
- manifest- and `DataVersion`-only segment progress;
- identical segment IDs and channel names progressing across replicas;
- high-water progress resisting readiness oscillation;
- target-version changes rebuilding the progress baseline;
- failed, empty/no-op, and concurrent stale refresh lifecycles;
- stale events ignored when the segment is no longer in next;
- `SegmentTask` and `LeaderTask` stale callbacks, with channel tasks
  excluded.

## Validation

- `git diff --check`
- `GOTOOLCHAIN=auto make milvus`
- `GOTOOLCHAIN=auto go test -tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${RPATH}" github.com/milvus-io/milvus/internal/querycoordv2/checkers github.com/milvus-io/milvus/internal/querycoordv2/job -count=1 -timeout 300s`
- `GOTOOLCHAIN=auto make lint-fix`
- `GOTOOLCHAIN=auto CLANG_FORMAT=/opt/homebrew/opt/llvm@15/bin/clang-format make check-proto-product`
- `GOTOOLCHAIN=auto CLANG_FORMAT=/opt/homebrew/opt/llvm@15/bin/clang-format make verifiers`

### DIFF
--- a/internal/querycoordv2/checkers/segment_checker.go
+++ b/internal/querycoordv2/checkers/segment_checker.go
@@ -33,7 +33,6 @@ import (
 	"github.com/milvus-io/milvus/internal/querycoordv2/session"
 	"github.com/milvus-io/milvus/internal/querycoordv2/task"
 	"github.com/milvus-io/milvus/internal/querycoordv2/utils"
-	"github.com/milvus-io/milvus/internal/storagev2/packed"
 	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
@@ -333,19 +332,12 @@ func (c *SegmentChecker) getSealedSegmentDiff(
 		if !existInDist {
 			return false
 		}
-		// Trigger reopen when storage v2 data version is behind the target.
-		// DataVersion bumps on storage v2 binlog changes that don't necessarily
-		// move the manifest version.
-		// Skip when the QueryNode did not report DataVersion (nil pointer from
-		// proto3 optional): during a mixed-version rollout an old QueryNode has
-		// no way to advance DataVersion, so triggering Reopen would loop forever.
-		if segInDist.DataVersion != nil && *segInDist.DataVersion < segment.GetDataVersion() {
-			return true
-		}
-		// Trigger reopen when dist manifest is older than target manifest.
-		// If dist manifest is same or newer (e.g., loaded after L0 compaction updated DataCoord),
-		// the data is already up-to-date and no reopen is needed.
-		cmp, err := packed.CompareManifestPath(segInDist.ManifestPath, segment.GetManifestPath())
+		// Trigger reopen when the loaded segment is behind the target. The helper
+		// checks DataVersion before manifest readiness, so storage v2 binlog changes
+		// still trigger Reopen when manifest paths are not comparable. A nil
+		// DataVersion from an old QueryNode is treated as compatible to avoid an
+		// endless Reopen loop, while Storage v3 relies on manifest readiness only.
+		ready, err := utils.IsSegmentDataReadyForTarget(segInDist, segment)
 		if err != nil {
 			mlog.RatedWarn(ctx, rate.Limit(10), "manifest path not comparable, skip reopen",
 				mlog.FieldSegmentID(segment.GetID()),
@@ -354,7 +346,7 @@ func (c *SegmentChecker) getSealedSegmentDiff(
 				mlog.Err(err))
 			return false
 		}
-		return cmp < 0
+		return !ready
 	}
 
 	nextTargetExist := c.targetMgr.IsNextTargetExist(ctx, collectionID)
@@ -463,17 +455,39 @@ func (c *SegmentChecker) filterOutSegmentInUse(ctx context.Context, replica *met
 			continue
 		}
 
-		stillInUseByDelegator := false
-		// if delegator has valid target version, and before it update to latest readable version, skip release it's sealed segment
-		for _, delegator := range delegatorList {
-			// Notice: if syncTargetVersion stuck, segment on delegator won't be released
-			readableVersionNotUpdate := delegator.View.TargetVersion != initialTargetVersion && delegator.View.TargetVersion < currentTargetVersion
-			if partition != nil && readableVersionNotUpdate {
-				// leader view version hasn't been updated, segment maybe still in use
-				stillInUseByDelegator = true
-				break
-			}
-		}
+		// Release a sealed segment only after every delegator with a synced target
+		// has converged to the current target version. A delegator on another target
+		// may still serve an older or orphaned readable snapshot that references the
+		// segment. The initial target version means no target has been synced yet and
+		// does not block release.
+		//
+		// This conservative safety barrier has two liveness risks:
+		//
+		//  1. Obsolete segments are retained while target convergence is in progress,
+		//     increasing QueryNode memory usage. If the retained data leaves
+		//     insufficient capacity to load the next target, current-target promotion
+		//     and segment release can stall each other. Increase QueryNode capacity to
+		//     allow the next target to finish loading.
+		//
+		//  2. Replacing the next target while a target transition is in progress can
+		//     cause permanent convergence starvation. For example, current may remain
+		//     at V1 while different delegators are serving V2 and V3. If next is
+		//     replaced from V4 to V5 before all delegators reach V4, and this keeps
+		//     happening, no target generation can be fully synchronized and promoted
+		//     to current. Because release requires every synced delegator to match
+		//     current, obsolete segments will then remain blocked indefinitely.
+		//
+		//     Progress-aware automatic refreshes therefore retain a healthy next target
+		//     while replica-scoped segment and channel readiness keeps advancing. A
+		//     stale target or an explicit manual refresh may intentionally replace the
+		//     selected target and restart the transition.
+		//
+		// Do not bypass this barrier to recover from either condition, since doing so
+		// may release segments still referenced by a readable snapshot.
+		stillInUseByDelegator := partition != nil && lo.ContainsBy(delegatorList, func(delegator *meta.DmChannel) bool {
+			return delegator.View.TargetVersion != initialTargetVersion &&
+				delegator.View.TargetVersion != currentTargetVersion
+		})
 
 		if !stillInUseByDelegator {
 			notUsed = append(notUsed, s)
--- a/internal/querycoordv2/observers/target_observer.go
+++ b/internal/querycoordv2/observers/target_observer.go
@@ -73,6 +73,15 @@ type targetUpdateRequest struct {
 
 type initRequest struct{}
 
+type nextTargetProgress struct {
+	targetVersion        int64
+	readySegmentReplicas int
+	readyReplicaChannels int
+	// lastUpdated records when this target version was installed or when its
+	// readiness last advanced.
+	lastUpdated time.Time
+}
+
 type TargetObserver struct {
 	cancel    context.CancelFunc
 	wg        sync.WaitGroup
@@ -84,11 +93,16 @@ type TargetObserver struct {
 	nodeMgr   *session.NodeManager
 
 	initChan chan initRequest
-	// nextTargetLastUpdate map[int64]time.Time
-	nextTargetLastUpdate *typeutil.ConcurrentMap[int64, time.Time]
-	updateChan           chan targetUpdateRequest
-	mut                  sync.Mutex                // Guard readyNotifiers
-	readyNotifiers       map[int64][]chan struct{} // CollectionID -> Notifiers
+	// nextTargetProgresses records the loading progress checkpoint for each
+	// collection's next target.
+	nextTargetProgresses *typeutil.ConcurrentMap[int64, nextTargetProgress]
+	// nextTargetStale records collections whose next target must be refreshed.
+	// A refresh consumes the marker before the RPC and restores it on failure or
+	// no-op, so a notification arriving during the RPC remains pending.
+	nextTargetStale *typeutil.ConcurrentSet[int64]
+	updateChan      chan targetUpdateRequest
+	mut             sync.Mutex                // Guard readyNotifiers
+	readyNotifiers  map[int64][]chan struct{} // CollectionID -> Notifiers
 
 	// loadingDispatcher updates targets for collections that are loading (also collections without a current target).
 	loadingDispatcher *taskDispatcher[int64]
@@ -116,7 +130,8 @@ func NewTargetObserver(
 		broker:               broker,
 		cluster:              cluster,
 		nodeMgr:              nodeMgr,
-		nextTargetLastUpdate: typeutil.NewConcurrentMap[int64, time.Time](),
+		nextTargetProgresses: typeutil.NewConcurrentMap[int64, nextTargetProgress](),
+		nextTargetStale:      typeutil.NewConcurrentSet[int64](),
 		updateChan:           make(chan targetUpdateRequest, 10),
 		readyNotifiers:       make(map[int64][]chan struct{}),
 		initChan:             make(chan initRequest),
@@ -298,6 +313,19 @@ func (ob *TargetObserver) TriggerUpdateCurrentTarget(collectionID int64) {
 	ob.loadingDispatcher.AddTask(collectionID)
 }
 
+// MarkNextTargetStale marks the collection's next target as stale when the
+// failed segment still belongs to that target.
+func (ob *TargetObserver) MarkNextTargetStale(collectionID, segmentID int64) {
+	if ob.targetMgr.GetSealedSegment(context.TODO(), collectionID, segmentID, meta.NextTarget) == nil {
+		return
+	}
+
+	// A concurrent target replacement can turn this into a conservative false
+	// positive. That only causes an extra refresh and avoids coupling tasks to a
+	// target generation.
+	ob.nextTargetStale.Upsert(collectionID)
+}
+
 func (ob *TargetObserver) check(ctx context.Context, collectionID int64) {
 	ob.keylocks.Lock(collectionID)
 	defer ob.keylocks.Unlock(collectionID)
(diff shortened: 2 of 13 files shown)


internal/querycoordv2/observers/target_observer.go
```diff
@@ -421,15 +449,64 @@ func (ob *TargetObserver) clean() {
...
+	progress, hasProgress := ob.nextTargetProgresses.Get(collectionID)
+	if hasProgress && progress.targetVersion != nextVersion {
+		ob.updateNextTargetProgress(ctx, collectionID)
+		return false
+	}
+
 	if !ob.isNextTargetExpired(collectionID) {
 		return false
 	}
+	if ob.isNextTargetSyncStarted(collectionID, nextVersion) {
```
liliu-z: This still depends on the all-delegator prepare gate and sync-start pin that were removed in #51547. Since the gate also sees RO/stopping delegators, one stale delegator that will no longer receive loads can prevent the ready ones from ever getting `SyncTargetVersion`. Could we rebase this on current master and keep the partial-sync behavior?

internal/querycoordv2/observers/target_observer.go
```diff
@@ -450,23 +527,44 @@ func (ob *TargetObserver) updateNextTarget(ctx context.Context, collectionID int
...
 	ob.nextTargetLastUpdate.Insert(collectionID, time.Now())
 }
 
-func (ob *TargetObserver) isNextTargetSyncStarted(ctx context.Context, collectionID int64) bool {
+func (ob *TargetObserver) countLoadedNextTargetSegments(ctx context.Context, collectionID int64) int {
+	targetSegments := ob.targetMgr.GetSealedSegmentsByCollection(ctx, collectionID, meta.NextTarget)
+	loadedSegmentIDs := typeutil.NewUniqueSet()
+	for _, segment := range ob.distMgr.SegmentDistManager.GetByFilter(meta.WithCollectionID(collectionID)) {
+		if _, ok := targetSegments[segment.GetID()]; ok {
+			loadedSegmentIDs.Insert(segment.GetID())
```
liliu-z: This count collapses every replica and node into one set of segment IDs, while readiness is checked per replica/channel and also considers manifest/data version. For example, a new replica can load the same IDs one by one without moving this count at all, so an otherwise healthy transition can still be refreshed as stalled. Could progress use the same dimensions as readiness?

internal/querycoordv2/observers/target_observer.go
```diff
@@ -482,6 +580,10 @@ func (ob *TargetObserver) isNextTargetSyncStarted(ctx context.Context, collectio
 }
 
 func (ob *TargetObserver) shouldUpdateCurrentTarget(ctx context.Context, collectionID int64) bool {
+	if ob.nextTargetStale.Contain(collectionID) {
```
liliu-z: The stale bit is only checked once here. `MarkNextTargetStale` can run from `scheduler.remove` while this function is in the readiness/`SyncDistribution` RPCs, and a mark that lands before `UpdateCollectionCurrentTarget` will still be ignored. We need to revalidate the same target generation atomically at commit time.

internal/querycoordv2/observers/target_observer.go
```diff
@@ -450,23 +527,44 @@ func (ob *TargetObserver) updateNextTarget(ctx context.Context, collectionID int
 			mlog.Err(err))
 		return err
 	}
-	ob.updateNextTargetTimestamp(collectionID)
+	ob.updateNextTargetProgress(ctx, collectionID)
 	ob.nextTargetSyncStarted.Remove(collectionID)
+	ob.nextTargetStale.Remove(collectionID)
```
liliu-z: Clearing the flag unconditionally here can lose a newer stale event: the recovery snapshot may be taken, the segment can disappear and get marked stale, and then this delete removes that mark even if the installed snapshot still contains it. Also, `UpdateCollectionNextTarget` can return nil without replacing `next` when recovery info is empty. Can we clear only the generation that was actually replaced and verify the version advanced?

internal/querycoordv2/checkers/segment_checker.go
```diff
@@ -464,11 +464,36 @@ func (c *SegmentChecker) filterOutSegmentInUse(ctx context.Context, replica *met
...
+		//  2. Replacing the next target while a target transition is in progress can
+		//     cause permanent convergence starvation. For example, current may remain
+		//     at V1 while different delegators are serving V2 and V3. If next is
+		//     replaced from V4 to V5 before all delegators reach V4, and this keeps
+		//     happening, no target generation can be fully synchronized and promoted
+		//     to current. Because release requires every delegator to match current,
+		//     obsolete segments will then remain blocked indefinitely.
+		//
+		//     Once any delegator reports a version different from current, periodic
+		//     and manual refreshes must keep the selected next target unchanged until
```
liliu-z: The manual paths don't currently enforce this invariant. `UpdateNextTarget` and `UpdatePartition` can replace an in-progress next target and clear its sync marker without retaining or reconciling the old generation; the collection lock only serializes the update call, not the whole transition. Either those paths need the same protection, or this comment and the PR scope should be narrowed.

weiliu1031: Narrowed the invariant instead of restricting the manual paths. Progress-aware automatic refresh keeps a healthy next target stable while replica-scoped segment or channel readiness advances. `UpdateNextTarget` and `UpdatePartition` remain explicit, low-frequency force-refresh requests and may intentionally replace the selected target and restart the transition.

Those manual updates share the per-collection lock with the observer check, including readiness RPCs and promotion, so they do not interrupt a check already in progress; they can replace next between checks by design. The SegmentChecker comment and PR description now state this scope explicitly.

internal/querycoordv2/observers/target_observer.go
weiliu1031: Fixed by replacing the stale boolean with a per-collection stale revision. Before fetching recovery info, the observer snapshots both the current next-target version and the stale revision.

RPC failures and empty/no-op refreshes leave the stale revision and progress baseline unchanged. Progress is reset and the captured revision is cleared only after a newer next-target version is actually installed. If another `ErrSegmentNotFound` arrives while the recovery RPC or progress sampling is in flight, it increments the revision, so the compare-and-delete cannot remove that newer event. Manual refreshes use the same path and are serialized with observer checks and promotion by the collection lock.

internal/querycoordv2/observers/target_observer.go
weiliu1031: After tracing the transition, I kept commit-wins semantics instead of adding a commit-time stale veto. Once readiness has passed and `SyncDistribution` RPCs have started, aborting promotion can leave delegators serving generation N while current remains on an older generation.

The stale callback now carries both collection and segment ID, covers `SegmentTask` and segment-based `LeaderTask`, and marks stale only if the failed segment is still in the current next target. A mark that arrives during the in-flight promotion is retained as a stale revision and drives the subsequent next-target refresh instead of invalidating the already-started commit.

The membership check intentionally does not attach a target generation to each task. A concurrent replacement can therefore produce one conservative false-positive refresh, but it cannot drop a real stale event.

internal/querycoordv2/observers/target_observer.go
weiliu1031: Updated the liveness metric to use replica-scoped readiness dimensions:

- `readySegmentReplicas` counts unique `(replicaID, segmentID)` pairs whose observed distribution copies satisfy the target manifest and `DataVersion`.
- `readyReplicaChannels` counts unique `(replicaID, channelName)` pairs that are data-ready or already synchronized to the target version.

Both counters are high-water marks within one target version, so a decrease in one dimension cannot be replayed as new progress when readiness oscillates. A target-version change rebuilds the baseline. This covers a new replica loading existing segment IDs, manifest/DataVersion-only reopen progress, and per-replica channel catch-up. These counters remain liveness signals only; promotion still uses the existing readiness path.

internal/querycoordv2/observers/target_observer.go
weiliu1031: Rebased onto current master, including [#51547](https://github.com/milvus-io/milvus/pull/51547). The removed all-delegator prepare gate and sync-start pin are not restored. `shouldUpdateCurrentTarget` continues to collect and synchronize ready delegators independently, preserving partial-sync behavior when an RO or stopping delegator cannot receive new loads. The timeout path now relies on replica-scoped loading progress instead of reintroducing that gate.

internal/querycoordv2/observers/target_observer.go
liliu-z: The latest code is still a ConcurrentSet, not a revision. More importantly, newVersion > oldVersion does not prove the refresh changed anything: the real TargetManager creates a new CollectionTarget with time.Now().UnixNano() for every non-empty recovery response, even when the content is identical and the failed segment is still there. So this can clear stale and reset progress on a semantic no-op, then churn through the same bad target again. Can the update return a real changed result, or keep enough segment/revision information to verify the stale cause was actually replaced? The test should use the real TargetManager for this case.

internal/querycoordv2/observers/target_observer.go
```diff
@@ -416,37 +455,182 @@ func (ob *TargetObserver) clean() {
...
+			mlog.String("reason", "next target progress is missing"))
 		return true
 	}
-	return time.Since(lastUpdated) > params.Params.QueryCoordCfg.NextTargetSurviveTime.GetAsDuration(time.Second)
+	if progress.targetVersion != nextVersion {
+		ob.recordNextTargetProgress(collectionID, ob.sampleNextTargetProgress(ctx, collectionID, nextVersion))
+		return false
+	}
+
+	if time.Since(progress.lastUpdated) <= params.Params.QueryCoordCfg.NextTargetSurviveTime.GetAsDuration(time.Second) {
```
liliu-z, line 506: This only samples readiness after the timeout has already expired. If progress happened shortly after the last sample and then stopped, we notice it at T and start another full T, so refresh can take almost 2x the configured no-progress interval. Can we sample every observer tick and keep lastUpdated as the time of the last observed increase?

internal/querycoordv2/checkers/segment_checker.go
```diff
@@ -463,17 +455,39 @@ func (c *SegmentChecker) filterOutSegmentInUse(ctx context.Context, replica *met
...
+		//     current, obsolete segments will then remain blocked indefinitely.
+		//
+		//     Progress-aware automatic refreshes therefore retain a healthy next target
+		//     while replica-scoped segment and channel readiness keeps advancing. A
+		//     stale target or an explicit manual refresh may intentionally replace the
+		//     selected target and restart the transition.
+		//
+		// Do not bypass this barrier to recover from either condition, since doing so
+		// may release segments still referenced by a readable snapshot.
+		stillInUseByDelegator := partition != nil && lo.ContainsBy(delegatorList, func(delegator *meta.DmChannel) bool {
```
liliu-z, line 487: This also blocks release when the delegator is on the current next target. For a release candidate that is absent from next, that case is already known to be safe. Keeping everything until current advances means one permanently stuck shard can stop reclamation on healthy shards and create the old-segments/no-memory/no-promotion loop. I think this needs a more precise serving-set or holder-side check, or another way to reclaim segments that are provably outside the readable view.

internal/querycoordv2/checkers/segment_checker.go
liliu-z: These are not only manual or necessarily low-frequency paths. Load/partition sync, refreshCollection/import, and replica-number or RG updates also call UpdateNextTarget or UpdatePartition. They can still replace a healthy in-progress target and recreate the churn this PR is trying to stop. Either those callers need the same stability rule, or we need an interaction test and a much narrower guarantee here.

internal/querycoordv2/observers/target_observer_test.go
```diff
@@ -733,13 +836,349 @@ func TestShouldUpdateCurrentTarget_OnlyReadyDelegatorsSynced(t *testing.T) {
...
+	}
+	mockUpdateNextTarget := mockey.Mock((*mockeyTargetManager).UpdateCollectionNextTarget).
+		To(func(_ *mockeyTargetManager, _ context.Context, _ int64) error {
+			if updateNextTargetErr != nil {
+				return updateNextTargetErr
+			}
+			if markDuringUpdate {
+				observer.MarkNextTargetStale(collectionID, segment.GetID())
+			}
+			if advanceTargetVersion {
```
liliu-z, line 1120: This test can model a no-op only because the mock chooses not to advance nextTargetVersion. Production TargetManager advances it for every non-empty recovery result, so this misses the real failing case. Please add coverage with the real TargetManager and identical recovery data, plus one fault-injection test for the full ErrSegmentNotFound -> scheduler callback -> observer refresh path.

internal/querycoordv2/observers/target_observer.go
weiliu1031: Sampling on every observer tick would significantly increase the cost of this path in high-cardinality deployments. A progress sample walks target and distribution segments, replicas, and channels, performs manifest/DataVersion readiness checks, and repeatedly takes distribution-manager read locks. With the defaults, changing the cadence from 300 seconds to 10 seconds makes this part roughly 30x more frequent. In a bulk-load or failure scenario, thousands of collections may have pending next targets at the same time, while shouldUpdateCurrentTarget already performs similar readiness scans in the same observer check.

This progress value is only a refresh-throttling liveness heuristic; it is not a promotion or readability condition. The explicit ErrSegmentNotFound stale signal still forces a refresh on the next observer tick. We therefore prefer to keep the bounded-cost observation-window behavior here: if a T-spaced sample observes progress, establish a new baseline and wait for the next window. This accepts a worst-case delay close to 2T, rather than scanning every pending collection every 10 seconds.

A strict last-progress + T guarantee would be better implemented with collection-scoped distribution change notifications or event-driven progress accounting, so only collections whose relevant distribution changed are resampled. That is a larger change and is outside this PR.

internal/querycoordv2/checkers/segment_checker.go
weiliu1031: This is a valid scope limitation, but these callers represent explicit state changes rather than the periodic automatic refresh controlled by the progress heuristic. refreshCollection/import may need newly published segments, UpdatePartition must include a newly created partition, and replica-number or resource-group updates change the desired placement. Retaining the old next target in those cases can omit the state the caller explicitly requested.

All of these updates go through the same per-collection key lock as readiness checks, SyncDistribution, and current-target promotion. They therefore cannot replace next in the middle of a transition check; they may replace it only between checks, which is intentional.

The guarantee in this PR is deliberately narrower: automatic observer refresh keeps a healthy next target stable while loading progresses, while explicit business-state updates may replace it and restart convergence. If those callers are shown to trigger frequently enough to cause churn, request coalescing or semantic target-diff handling should be addressed separately rather than suppressing required target changes here.

internal/querycoordv2/observers/target_observer_test.go
weiliu1031: Implemented this with the production TargetManager path.

- TestNextTargetStale_IdenticalRecoveryKeepsCause uses a real TargetManager with identical recovery data. It verifies that the target version advances while the failed segment remains selected and the stale cause stays pending.
- TestNextTargetStale_ErrSegmentNotFoundRefreshesRealTarget injects ErrSegmentNotFound through the real executor and scheduler removal callback into TargetObserver. It verifies that identical recovery retains the cause and that a later recovery without the failed segment clears it.

The observer now tracks failed segment IDs per collection and clears each cause only when that segment is absent from the installed next target. A timestamp-only target-version advance is therefore no longer treated as a semantic replacement.

The three targeted stale-lifecycle tests pass with -race and -cover.

Issue #51574, opened 2026-07-17
[Bug]: SemanticHighlighter misaligns nullable VARCHAR field rows


bigsheeper:
## Environment
- Milvus version: current master (`e8c233b69d9f0a177baf5aa239329f8d523b29db`)
- Component: Proxy search highlighter pipeline
- Highlighter: SemanticHighlighter

## Current Behavior
When `SemanticHighlighter` highlights a schema VARCHAR field that contains nullable rows, the proxy passes the compact string payload directly to `SemanticHighlight.Process`.

For a result set with three logical hits and `ValidData = [true, false, true]`, the string payload may contain only the two non-null values:

```go
StringData.Data = []string{"text 1", "text 3"}
```

The semantic highlighter currently receives:

```go
[]string{"text 1", "text 3"}
```

This loses the null row and shifts later text values to the wrong result row, so highlight output can be attached to the wrong hit or have the wrong cardinality.

## Expected Behavior
Semantic highlighter input should be aligned to logical search result rows. Nullable rows should be represented by an empty string placeholder, matching the existing lexical highlighter row-alignment behavior.

For the same result set, `SemanticHighlight.Process` should receive:

```go
[]string{"text 1", "", "text 3"}
```

## Reproduction Evidence
A focused Go test against the old code fails before the fix:

```text
FAIL: TestSearchPipeline/TestSemanticHighlightOpNullableStringAlignsRows
expected: []string{"text 1", "", "text 3"}
actual  : []string{"text 1", "text 3"}
```

Command used:

```bash
GOWORK=off bash -lc 'source scripts/setenv.sh && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy -run "TestSearchPipeline/TestSemanticHighlightOpNullableStringAlignsRows"'
```

## Suspected Cause
`semanticHighlightOperator.run` uses `fieldDatas.GetScalars().GetStringData().GetData()` for schema fields, while the lexical highlighter path already uses `rowAlignedStringFieldData(...)` to expand nullable compact payloads by `ValidData`.

## Impact
High correctness issue: valid semantic highlight results can be associated with the wrong search hit whenever a nullable highlighted VARCHAR field has null rows before non-null rows in the returned result set.

PR #51575, opened 2026-07-18
enhance: enforce row-count invariant for semantic highlight inputs

bigsheeper:
issue: #51574

## What

`SemanticHighlight.Process` slices its `documents` argument by `topks` (`semantic_highlight.go:195`) without checking that they sum to `len(documents)` — the only thing standing behind that is a `//nolint:gosec // bounds are guaranteed by topks summing to len(documents)` comment. A shorter slice panics in the proxy; verified directly: `Process(ctx, []int64{3}, []string{"a","b"})` gives `slice bounds out of range [:3] with capacity 2`.

This PR makes that invariant enforced rather than assumed:

- Schema fields go through `rowAlignedStringFieldData(fieldData, sum(topks))` — the helper the lexical path has used since #50572 — so `texts` always carries exactly one slot per search-result row, NULL rows included.
- The dynamic-field (`$meta`) branch gets the matching length check, so both branches return the same error instead of one of them panicking.

## Scope correction (after review)

An earlier version of this description claimed nullable scalar payloads can arrive compacted and shift later rows left. **I could not substantiate that for the search path and am no longer claiming it**: segcore writes one scalar slot per row unconditionally (`Utils.cpp:651-659`; `MergeDataArray`'s scalar branch likewise, only its vector branch skips invalid rows), `AppendFieldData` indexes string data by logical row (`schema.go:1118-1129`), `queryutil` slices scalars logically (`helpers.go:890`), and compaction is scoped to vector types by `IsCompactNullableVectorFieldData` (`schema.go:2101-2102`). Treat this as hardening of an unchecked invariant, not a fix for observed data corruption. Issue #51574 needs the same correction.

`typeutil.GetDataIterator` (`schema.go:3488-3515`) does branch on both layouts, so `rowAlignedStringFieldData` remains correct either way.

## Follow-ups (not in this PR)

- `pipeline_trace.go:209-212` asserts the **opposite** invariant ("StringData uses compact encoding") and `perNqCardinalityCompact` miscounts cardinality against full-size data with nulls (`topks=[3]`, `ValidData=[t,f,t]`, `Data=["a","","a"]` reports 3, logical answer 2). Observability only; filing separately.
- NULL rows still reach the highlight provider as empty documents.

## Verification

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy -run "TestSearchPipeline|TestHighlighter"`
- `golangci-lint run ./internal/proxy/...` — 0 issues

### DIFF
--- a/internal/proxy/highlighter.go
+++ b/internal/proxy/highlighter.go
@@ -576,7 +576,10 @@ func (op *semanticHighlightOperator) run(ctx context.Context, span trace.Span, i
 		if !ok {
 			return nil, merr.WrapErrParameterInvalidMsg("get highlight failed, text field not in output field %d", fieldID)
 		}
-		texts := fieldDatas.GetScalars().GetStringData().GetData()
+		texts, err := rowAlignedStringFieldData(fieldDatas, searchResultHitCount(resultData))
+		if err != nil {
+			return nil, err
+		}
 		fieldName := op.highlight.GetFieldName(fieldID)
 
 		highlightResult, err := op.processFieldHighlight(ctx, fieldName, texts, topks)
--- a/internal/proxy/highlighter.go
+++ b/internal/proxy/highlighter.go
@@ -569,6 +569,7 @@ func (op *semanticHighlightOperator) run(ctx context.Context, span trace.Span, i
 	}
 	highlightResults := []*commonpb.HighlightResult{}
 	topks := result.Results.GetTopks()
+	rowNum := searchResultHitCount(resultData)
 
 	// Process schema fields
 	for _, fieldID := range op.highlight.FieldIDs() {
@@ -576,7 +577,7 @@ func (op *semanticHighlightOperator) run(ctx context.Context, span trace.Span, i
 		if !ok {
 			return nil, merr.WrapErrParameterInvalidMsg("get highlight failed, text field not in output field %d", fieldID)
 		}
-		texts, err := rowAlignedStringFieldData(fieldDatas, searchResultHitCount(resultData))
+		texts, err := rowAlignedStringFieldData(fieldDatas, rowNum)
 		if err != nil {
 			return nil, err
 		}
@@ -620,6 +621,14 @@ func (op *semanticHighlightOperator) run(ctx context.Context, span trace.Span, i
 
 		for _, dynFieldName := range dynFieldNames {
 			texts := allDynFieldTexts[dynFieldName]
+			// Process slices documents by topks without checking that they sum to
+			// len(documents), so an undersized $meta payload would go out of range
+			// there. Fail with the same error the schema branch returns instead.
+			if len(texts) != rowNum {
+				return nil, merr.WrapErrServiceInternalMsg(
+					"get highlight failed, dynamic field %s has %d rows, expected %d",
+					dynFieldName, len(texts), rowNum)
+			}
 
 			highlightResult, err := op.processFieldHighlight(ctx, dynFieldName, texts, topks)
 			if err != nil {
--- a/internal/proxy/search_pipeline_test.go
+++ b/internal/proxy/search_pipeline_test.go
@@ -1609,6 +1609,73 @@ func (s *SearchPipelineSuite) TestSemanticHighlightOp() {
 	s.Equal([]string{"<em>highlighted</em> text 3"}, highlightResult.Datas[2].Fragments)
 }
 
+func (s *SearchPipelineSuite) TestSemanticHighlightOpNullableStringAlignsRows() {
+	ctx := context.Background()
+
+	mockProcess := mockey.Mock((*highlight.SemanticHighlight).Process).To(
+		func(h *highlight.SemanticHighlight, ctx context.Context, topks []int64, texts []string) ([][]string, [][]float32, error) {
+			s.Equal([]int64{3}, topks)
+			s.Equal([]string{"text 1", "", "text 3"}, texts)
+			return [][]string{
+					{"highlighted text 1"},
+					{},
+					{"highlighted text 3"},
+				}, [][]float32{
+					{0.9},
+					{},
+					{0.7},
+				}, nil
+		}).Build()
+	defer mockProcess.UnPatch()
+
+	mockFieldIDs := mockey.Mock((*highlight.SemanticHighlight).FieldIDs).Return([]int64{101}).Build()
+	defer mockFieldIDs.UnPatch()
+
+	mockGetFieldName := mockey.Mock((*highlight.SemanticHighlight).GetFieldName).Return(testVarCharField).Build()
+	defer mockGetFieldName.UnPatch()
+
+	op := &semanticHighlightOperator{
+		highlight: &highlight.SemanticHighlight{},
+	}
+
+	searchResults := &milvuspb.SearchResults{
+		Results: &schemapb.SearchResultData{
+			NumQueries: 1,
+			TopK:       3,
+			Topks:      []int64{3},
+			Ids:        testSearchResultIDs(1, 2, 3),
+			FieldsData: []*schemapb.FieldData{
+				{
+					FieldId:   101,
+					FieldName: testVarCharField,
+					Type:      schemapb.DataType_VarChar,
+					ValidData: []bool{true, false, true},
+					Field: &schemapb.FieldData_Scalars{
+						Scalars: &schemapb.ScalarField{
+							Data: &schemapb.ScalarField_StringData{
+								StringData: &schemapb.StringArray{
+									Data: []string{"text 1", "text 3"},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	results, err := op.run(ctx, s.span, searchResults)
+	s.NoError(err)
+	s.Require().Len(results, 1)
+
+	result := results[0].(*milvuspb.SearchResults)
+	s.Require().Len(result.GetResults().GetHighlightResults(), 1)
+	highlightResult := result.GetResults().GetHighlightResults()[0]
+	s.Require().Len(highlightResult.GetDatas(), 3)
+	s.Equal([]string{}, highlightResult.GetDatas()[1].GetFragments())
+	s.Equal([]float32{}, highlightResult.GetDatas()[1].GetScores())
+}
+
 func (s *SearchPipelineSuite) TestSemanticHighlightOpMissingField() {
 	ctx := context.Background()
 
--- a/internal/proxy/search_pipeline_test.go
+++ b/internal/proxy/search_pipeline_test.go
@@ -1671,9 +1671,15 @@ func (s *SearchPipelineSuite) TestSemanticHighlightOpNullableStringAlignsRows() 
 	result := results[0].(*milvuspb.SearchResults)
 	s.Require().Len(result.GetResults().GetHighlightResults(), 1)
 	highlightResult := result.GetResults().GetHighlightResults()[0]
+	// The fix under test is the row alignment feeding Process: the NULL row
+	// materializes as "" at its own index, asserted on `texts` inside the mock
+	// above. Here we only verify the per-row results survive alignment 1:1 —
+	// the row-1 payload is whatever the mock returned, not a NULL-semantics
+	// guarantee, so we assert positional pass-through across all three rows.
 	s.Require().Len(highlightResult.GetDatas(), 3)
+	s.Equal([]string{"highlighted text 1"}, highlightResult.GetDatas()[0].GetFragments())
 	s.Equal([]string{}, highlightResult.GetDatas()[1].GetFragments())
-	s.Equal([]float32{}, highlightResult.GetDatas()[1].GetScores())
+	s.Equal([]string{"highlighted text 3"}, highlightResult.GetDatas()[2].GetFragments())
 }
 
 func (s *SearchPipelineSuite) TestSemanticHighlightOpMissingField() {
--- a/internal/proxy/search_pipeline_test.go
+++ b/internal/proxy/search_pipeline_test.go
@@ -2110,6 +2110,72 @@ func (s *SearchPipelineSuite) TestSemanticHighlightOpMixedFields() {
 	s.Equal([]string{"<em>dynamic</em> text"}, result.Results.HighlightResults[1].Datas[0].Fragments)
 }
 
+// Process slices documents by topks without checking that they sum to
+// len(documents), so a $meta payload shorter than the hit count would go out of
+// range there. The dynamic-field branch must reject it up front, the same way
+// the schema branch does.
+func (s *SearchPipelineSuite) TestSemanticHighlightOpDynamicFieldRowCountMismatch() {
+	ctx := context.Background()
+
+	mockProcess := mockey.Mock((*highlight.SemanticHighlight).Process).To(
+		func(h *highlight.SemanticHighlight, ctx context.Context, topks []int64, texts []string) ([][]string, [][]float32, error) {
+			s.Fail("Process must not be called with a row-count mismatch")
+			return nil, nil, nil
+		}).Build()
+	defer mockProcess.UnPatch()
+
+	mockFieldIDs := mockey.Mock((*highlight.SemanticHighlight).FieldIDs).Return([]int64{}).Build()
+	defer mockFieldIDs.UnPatch()
+
+	mockHasDynamicFields := mockey.Mock((*highlight.SemanticHighlight).HasDynamicFields).Return(true).Build()
+	defer mockHasDynamicFields.UnPatch()
+
+	mockDynamicFieldNames := mockey.Mock((*highlight.SemanticHighlight).DynamicFieldNames).Return([]string{"dyn_content"}).Build()
+	defer mockDynamicFieldNames.UnPatch()
+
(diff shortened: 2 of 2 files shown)


czs007:
Adversarial review surfaced three issues — one medium correctness gap in NULL-row handling and two low-severity items (a missing length guard and a tautological test assertion) — none of which block, but all worth addressing.

## 🟠 Medium

**`internal/proxy/highlighter.go:579` — NULL rows materialized as `""` are sent to the external semantic-highlight provider**
`rowAlignedStringFieldData` initializes `texts := make([]string, rowNum)` and only overwrites non-nil rows, so NULL rows remain the zero-value `""` (highlighter.go:388–399). Those `""` documents flow through `Process → processOneQuery → provider.highlight`, a remote Zilliz model call (zilliz_highlight_provider.go:76–77). The only enforced contract is `len(highlights) == len(documents)` (semantic_highlight.go:173); nothing guarantees a `""` document maps back to empty fragments/zero scores, so a NULL row can return a non-empty highlight, and every NULL row also consumes a billed/latency document slot. Unlike the lexical path — where `""` is tokenized locally and deterministically yields no fragments (highlighter.go:428, 500–517) — the "NULL ⇒ no highlight" guarantee is not enforced on the semantic path. (Note this mirrors the pre-existing dynamic-field path, which already uses `""` as a graceful-degradation sentinel at highlighter.go:667,678, so the fix inherits the pattern rather than introducing it.)
*Suggestion:* in `processFieldHighlight`, overwrite `result.Datas[i]` with an empty `HighlightData` for NULL rows; or drop NULL rows via `ValidData` before `Process` and re-expand afterward. *(raised by xiaocai)*

## 🟡 Low

**`internal/proxy/highlighter.go:659` — `extractMultipleDynamicFieldTexts` lacks a length guard against `searchResultHitCount`**
The schema-field path now validates row count in `rowAlignedStringFieldData`, returning a `merr` error on length mismatch (highlighter.go:358–400, 374–378). The dynamic-field path, however, sizes its output solely by `len(jsonDataBytes)` (highlighter.go:659) with no check against `searchResultHitCount(resultData)`. If `len(jsonDataBytes)` is ever smaller than `sum(topks)`, the resulting `texts` slice reaches `Process`, where `documents[start:start+size]` indexes out of range and panics (semantic_highlight.go:195). In practice `$meta` is a per-row, generally non-nullable JSON field, so a mismatch is unlikely — hence low, not medium — but the guard asymmetry is real.
*Suggestion:* add a one-line length check (or reuse the schema-path validation) so both paths fail consistently and safely. *(raised by xiaocai)*

**`internal/proxy/search_pipeline_test.go:1675` — test asserts a NULL-highlight guarantee the production code does not make**
The mock at search_pipeline_test.go:1615–1628 hardcodes `Process` to return `{}` / `{}` for row index 1, and `processFieldHighlight` merely copies `highlights[i]`/`scores[i]` into `Datas[i]` (highlighter.go:651–653). The assertions `s.Equal([]string{}, ...GetDatas()[1].GetFragments())` and `s.Equal([]float32{}, ...GetScores())` at lines 1675–1676 are therefore tautological — they prove the operator copies mock output, not that a NULL row yields an empty highlight. The assertion carrying real weight is line 1618, `s.Equal([]string{"text 1", "", "text 3"}, texts)`, which verifies `rowAlignedStringFieldData` produced `""` at the NULL row (the actual fix).
*Suggestion:* the row-1 output assertions only become meaningful once the code forces NULL rows to empty (see the line-579 finding); until then, don't count them as coverage of NULL semantics — drive the mock from real operator output or assert against the enforced behavior. *(raised by xiaocai)*

bigsheeper:
Thanks — verified all three against the current head.

**Medium (NULL rows sent to the semantic provider):** mechanism confirmed — `rowAlignedStringFieldData` leaves NULL rows as `""` and they reach `Process`. Keeping as-is for this PR: this mirrors the pre-existing dynamic-field path (`""` graceful-degradation sentinel at highlighter.go:667/678), so it is not newly introduced here, and this PR's scope is row **alignment**, not provider-input filtering. A clean "empty out NULL rows' `HighlightData`" is not reliably doable in `processFieldHighlight`: after alignment a NULL row and a genuinely-empty string are both `""` and indistinguishable without threading `ValidData` through. The proper fix — drop NULL rows before `Process` and re-expand after (which also saves the billed slot) — is a separate hardening; tracking it as a follow-up rather than folding it in here.

**Low (dynamic-path length guard, highlighter.go:659):** real asymmetry — the schema path validates length, the dynamic path sizes by `len(jsonDataBytes)` with no check against `searchResultHitCount`, so an undersized `$meta` would panic in `documents[start:start+size]`. It is on the dynamic-field path this PR does not touch, so leaving it out of this PR's scope and addressing the guard as a follow-up so both paths fail consistently.

**Low (tautological test assertion):** fixed in c4c278b59a. The row-1 `[]`/`[]` assertions only proved the operator copies the mock output, not NULL semantics. Replaced them with an honest positional pass-through check across all three rows plus a clarifying comment; the assertion that actually pins the fix is the aligned `texts` (`["text 1","","text 3"]`) asserted inside the mock.

czs007:
Adversarial review of this PR surfaced six findings — one high-severity concern about the stated premise, three medium-severity gaps in hardening and invariant consistency, and two low-severity design/consistency notes.

---

### High

**`internal/proxy/highlighter.go:579` — the claimed live compaction bug has no producer in the search path**

*Problem:* The PR body and issue #51574 assert a live correctness bug ("nullable scalar field payloads can be compacted... reading `StringData.Data` directly can shift later rows left"). Every producer and consumer of nullable **scalar** `FieldData` in the search path uses the full-size layout (one data slot per logical row, placeholder at null positions); compaction is a documented convention for nullable **vector** payloads only:

- `internal/core/src/segcore/Utils.cpp:650-658` — `CreateScalarDataArrayFrom` writes `count` entries unconditionally for VARCHAR/TEXT; `valid_data` is emitted separately.
- `internal/core/src/segcore/Utils.cpp:978-983` — `MergeDataArray`'s scalar branch always appends one value per row; only the vector branch (900-913) skips data for invalid rows.
- `pkg/util/typeutil/schema.go:1118-1129` — `AppendFieldData` indexes the source by *logical* row and emits one value per row, so proxy reduce both requires and produces the full-size layout.
- `pkg/util/typeutil/schema.go:2101-2103` and `pkg/util/typeutil/nullable_vector.go:22` — `IsCompactNullableVectorFieldData` scopes compaction to vector types explicitly.
- `internal/proxy/search_pipeline.go:2757-2767` and `internal/distributed/proxy/httpserver/utils.go:2642` — the pipeline reorder and the REST row projection also index string data by logical row.

The issue's "Reproduction Evidence" is this PR's own unit test over a hand-built `FieldData`, which is circular. Verified end-to-end: applying only the test file to the PR base fails (`expected []string{"text 1","","text 3"} / actual []string{"text 1","text 3"}`), and applying the fix makes the highlight suite pass — so the test is a genuine guard for the guarded *shape*, but nothing demonstrates that shape is reachable from a real search. Secondarily: if compact scalar payloads were real, the highlighter would not be the first victim — reduce, reorder, and REST projection would already have corrupted the result beforehand, so a highlighter-only patch could not restore correctness.

*Suggestion:* Either name the producing code path (and fix it there), or downgrade the description from a high-severity correctness fix to defensive hardening. *(raised by adversarial-audit)*

---

### Medium

**`internal/util/function/highlight/semantic_highlight.go:195` — stated impact does not match the code path**

*Problem:* The issue and PR describe the pre-fix impact as "highlight output can be attached to the wrong hit or have the wrong cardinality". For the exact scenario given (nq=1, `topks=[3]`, 2 non-null texts), `Process` executes `documents[0:3]` on a 2-element slice: that panics when `cap == len` — the case in this PR's own fixture, `Data: []string{"text 1","text 3"}` — or silently reads zero-valued slots past the logical end when the proto decoder left spare capacity. Lines 191-202 carry `//nolint:gosec // bounds are guaranteed by topks summing to len(documents)`, but nothing in `Process` checks that invariant (only `len(highlight.queries) != nq` at line 182), and `processOneQuery` (173-175) validates provider output against the *slice* length rather than the logical row count. Because the test mocks `Process`, this failure mode is never exercised.

*Suggestion:* Correct the impact wording to reflect a panic / out-of-range read rather than mis-attribution; if a panic is genuinely reachable, that raises the priority of the question above about who produces the short slice. *(raised by adversarial-audit)*

**`internal/proxy/highlighter.go:616` — sibling dynamic-field branch still feeds `Process` an unvalidated row count**

*Problem:* `semanticHighlightOperator.run` now validates row alignment for schema fields via `rowAlignedStringFieldData(fieldDatas, searchResultHitCount(resultData))` (highlighter.go:366-385, 579), but the dynamic-field branch a few lines below builds `texts` with length `len(jsonDataBytes)` (612-624, output length derived purely from the `$meta` payload at 660-663) and never compares it to `searchResultHitCount(resultData)` before handing it to the same `Process`. If the `$meta` payload ever carries fewer rows than `sum(topks)`, the failure is a slice-out-of-range panic in the proxy rather than the clean `merr` the schema branch now returns.

*Suggestion:* Add the same length check to the dynamic-field branch — one line, and it makes the two branches of the same function consistent with the PR's stated goal of "preserving one text slot per search result row". *(raised by adversarial-audit)*

**`internal/proxy/pipeline_trace.go:210` — repo contains the opposite invariant for the same data, unreconciled**

*Problem:* `pipeline_trace.go` states and relies on the inverse of what segcore, `AppendFieldData`, and `reorderFieldData` do — line 209-212 comments "StringData uses compact encoding: only non-null values are stored" and dispatches to `perNqCardinalityCompact`, which advances `compactIdx` only on valid rows (245-273); lines 180-182 make the same assumption. Exactly one of the two beliefs can be true, and this PR silently picks the `pipeline_trace` side for one call site without resolving the contradiction. If full-size is the truth, `perNqCardinalityCompact` mis-splits NQ boundaries whenever nulls are present: with `topks=[3]`, `ValidData=[true,false,true]`, `Data=["a","","a"]`, the loop collects `nqVals=["a", compact[1]=""]` and adds 1 for the null → cardinality 3, while the true logical answer is 2, and it never reads row 2's real value.

*Suggestion:* Whichever way the premise above is resolved, one of these two files needs a fix — and the invariant should be documented in one place rather than restated inconsistently. *(raised by adversarial-audit)*

---

### Low

**`internal/proxy/highlighter.go:397` — NULL rows are sent to the external highlight provider as empty documents**

*Problem:* After alignment a NULL row materializes as `""` (387-399: `texts := make([]string, rowNum)`, null rows left at the zero value) and is forwarded to the remote provider like any other document — `processOneQuery` (semantic_highlight.go:165-172) passes the whole slice through, with `len(documents) == 0` the only short-circuit, and `zilliz_highlight_provider.go:76-82` hands `texts` straight to the client. Null rows therefore consume model quota and latency, and whatever the provider returns for an empty document is surfaced to the client as a normal `HighlightData` for a row with no text. The PR's own test comment concedes this ("the row-1 payload is whatever the mock returned, not a NULL-semantics guarantee"), and since `Process` is mocked the test cannot detect it. This is a pre-existing design gap rather than a defect introduced here, but it is the part of "nullable support" left undefined.

*Suggestion:* Short-circuit locally — emit empty fragments/scores for `!ValidData` rows and drop them from the provider batch. *(raised by adversarial-audit)*

**`internal/proxy/highlighter.go:579` — row count recomputed inside the per-field loop**

*Problem:* The lexical operator hoists `rowNum := searchResultHitCount(resultData)` above the task loop and reuses it (highlighter.go:417, 428, 447); the new semantic call site recomputes it on every highlight field inside `for _, fieldID := range op.highlight.FieldIDs()`.

*Suggestion:* Hoist it. The cost is negligible, but the asymmetry is gratuitous given the PR is explicitly reusing the lexical path's helper — and it makes the value available for the dynamic-field guard requested above. *(raised by adversarial-audit)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/c4c278b59a3f0412145aa433046e677ee563dc18...czs007:team2-fix/T69c0f52b45?expand=1

bigsheeper:
Thanks — all six verified against the current head. The High finding holds, and I've corrected the PR description accordingly rather than defending the premise; two of the mediums are fixed in `24ef3981aa`.

**High (`highlighter.go:579`) — confirmed. Description corrected, change kept as hardening.**

I went looking for the producer and could not find one. Every producer and sibling consumer in the search path uses the full-size scalar layout:

- `Utils.cpp:651-659` — `CreateScalarDataArrayFrom`'s VARCHAR/TEXT branch loops `for (i = 0; i < count; i++)` with no validity check.
- `MergeDataArray` — the vector branch has `if (!is_valid) { …; continue; }`, the scalar branch that follows takes `data[src_offset]` by logical row and emits `valid_data` separately.
- `schema.go:1118-1129` — `AppendFieldData`'s string branch indexes `srcScalar.StringData.Data[idx]` by logical row; it doesn't just tolerate full-size, it *requires* it.
- `IsCompactNullableVectorFieldData` (`schema.go:2101-2102`) gates on `IsSupportedNullableVectorType`, which lists only the six vector types (`nullable_vector.go:25-36`).
- One more you didn't list, pointing the same way: `queryutil/helpers.go:890` slices scalars logically by `[start:end]`, and `:1368,1394` build compact indices only for vector fields.

So the "compacted payload shifts later rows left" mechanism is not substantiated for this path, and your point about the test being circular is fair. I've rewritten the PR body to say so explicitly and retitled it `enhance:`; issue #51574 needs the same correction and I'll update it.

I do think the change is still worth landing, on a narrower and checkable claim: `searchResultHitCount` is exactly `sum(topks)`, so `rowAlignedStringFieldData(fd, sum(topks))` *guarantees* `len(texts) == sum(topks)` — which is precisely what `Process`'s `//nolint:gosec // bounds are guaranteed by topks summing to len(documents)` asserts and never checks. It also puts the semantic branch on the same helper the lexical branch has used since #50572. Note `GetDataIterator` (`schema.go:3488-3515`) itself branches on `dataLen == len(validData)` vs. a compact index map, so the helper is not inventing a shape.

**Medium (`semantic_highlight.go:195`) — confirmed, wording fixed.** You're right that the pre-fix failure mode is a panic / out-of-range read, not mis-attribution. Verified directly rather than by reading: `Process(ctx, []int64{3}, []string{"a","b"})` panics with `slice bounds out of range [:3] with capacity 2`. The PR body now describes it that way.

**Medium (`highlighter.go:616`, dynamic-field branch) — fixed.** I had deferred this as a follow-up last round; you're right that it belongs here, since leaving it means the two branches of one function fail differently. Added the same length check, returning the schema branch's error shape.

One honesty note on the test I added (`TestSemanticHighlightOpDynamicFieldRowCountMismatch`): it mocks `Process`, so what it actually pins is that `Process` is never entered with a mismatched row count — it does not itself execute the out-of-range slice. The slicing panic is the separate direct check quoted above.

**Medium (`pipeline_trace.go:210`) — confirmed, filing separately.** Real contradiction, and I reproduced your arithmetic: `topks=[3]`, `ValidData=[true,false,true]`, `Data=["a","","a"]` gives cardinality 3 where the logical answer is 2, because `perNqCardinalityCompact` advances `compactIdx` only on valid rows. Given the evidence above, `pipeline_trace.go` is the side that's wrong. It's observability-only, pre-existing, and outside this diff, so I'm not folding it in — filing an issue and noting it in the PR body as a follow-up.

**Low (`highlighter.go:397`, NULL rows to the provider)** — unchanged from my previous reply: real, pre-existing, and a proper fix needs drop-before-`Process` plus re-expand-after, which is its own change. Listed as a follow-up.

**Low (`highlighter.go:579`, row count recomputed per field) — fixed.** Hoisted `rowNum` above the field loop, matching the lexical operator, and reused it for the dynamic-field guard.

czs007:

Verified:

- `internal/proxy/highlighter.go:572-586` — schema-field documents are derived from `searchResultHitCount` (the exact sum of `topks`), which is precisely the precondition `SemanticHighlight.Process` relies on when slicing `documents[start:start+size]`.
- `internal/proxy/highlighter.go:358-401` — `rowAlignedStringFieldData` rejects both undersized and oversized payloads up front, and handles the no-`validData`, full-size zero-filled, and compact nullable layouts, returning typed internal errors instead of panicking.
- `pkg/util/typeutil/schema.go:3476-3519` — the data iterator distinguishes full-size scalar payloads from compact nullable ones and never shifts values across NULL rows, so NULL rows stay `""` at their logical positions.
- `internal/proxy/highlighter.go:622-633` — dynamic-field arrays are checked for exact row-count equality before provider invocation, since `extractMultipleDynamicFieldTexts` always returns `len(jsonDataBytes)` entries; this closes both the short-slice panic and the silently-ignored-extra-rows path.
- `internal/proxy/highlighter.go:628` — row-count violations originate from internal search-result data rather than user request content, so `WrapErrServiceInternalMsg` is the correct classification and matches the schema branch.
- `internal/proxy/search_pipeline_test.go:1612-1683, 2117-2177` — the nullable test asserts the exact row-aligned `texts` passed to `Process` plus positional propagation, and the dynamic-mismatch test asserts `Process` is never invoked and the expected error is returned; both genuinely reach the new branches.

czs007 (approved):
/lgtm

PR #48953, opened 2026-04-12
feat: support Sort/Bitmap/Hybrid index types for JSON Path Index

zhengbuqian:
this PR allows to create Sort/Bitmap/Hybrid index for json path index, and also removed the `JsonInvertedIndex` class, to use `InvertedIndexTantivy` directly for Inverted index type.

C++ changes:
- Add ConvertJsonToTypedFieldData<T>() to extract typed values from JSON field data with separate tracking of non_exist_offsets for EXISTS semantics
- Add JsonScalarIndexWrapper<T, BaseIndex> template for Sort/Bitmap with dual file-manager pattern (original JSON schema for reading, cast-type schema for base index dispatching)
- Add JsonHybridScalarIndex<T> with validity-aware cardinality counting
- Add IndexBase::Exists() virtual method; override in Sort/Bitmap/Hybrid wrappers using non_exist_offsets (serialized via WriteEntries/LoadEntries)
- Simplify ExistsExpr to use index->Exists() uniformly
- Extend IndexFactory::CreateJsonIndex() to route STL_SORT/BITMAP/HYBRID

Go changes:
- Update STL_SORT/Bitmap/Hybrid checkers to accept JSON with cast_type and json_path validation
- Change AUTOINDEX default for JSON from INVERTED to HYBRID
- Bump ScalarIndexEngineVersion to 4
- Add version gate in DataCoord CreateIndex and snapshot RestoreIndexes
- Update test fixtures to cover new (index_type, cast_type) combinations

image

issue: https://github.com/milvus-io/milvus/issues/48954
design-doc: https://github.com/milvus-io/milvus-design-docs/blob/main/design_docs/20260410-json_path_index_multi_type.md

PR #51580, opened 2026-07-18
enhance: keep milvus-proto checkout pinned

FantasqueX:
Always select the milvus-proto revision required by go.mod, even when a cached checkout already exists. Otherwise, switching Milvus branches can generate sources from an incompatible proto revision.

This backports the version-selection behavior introduced on master by commit 99097abea8f5eacb275b29cc8042b7836cf10425 (#48953). Use a detached checkout so pinning the dependency does not move its local branch.

backport pr: #48953

Assisted by gpt-5.6-sol.

### DIFF
--- a/scripts/download_milvus_proto.sh
+++ b/scripts/download_milvus_proto.sh
@@ -8,18 +8,22 @@ if [ ! -d "$THIRD_PARTY_DIR/milvus-proto" ]; then
   mkdir -p $THIRD_PARTY_DIR
   pushd $THIRD_PARTY_DIR
   git clone https://github.com/milvus-io/milvus-proto.git
-  cd milvus-proto
-  # try tagged version first
-  COMMIT_ID=$(git ls-remote https://github.com/milvus-io/milvus-proto.git refs/tags/${API_VERSION} | cut -f 1)
-  if [[ -z $COMMIT_ID ]]; then
-    # parse commit from pseudo version (eg v0.0.0-20230608062631-c453ef1b870a => c453ef1b870a)  
-    COMMIT_ID=$(echo $API_VERSION | awk -F'-' '{print $3}')
-  fi
-  echo "version: $API_VERSION, commitID: $COMMIT_ID"
-  if [ -z $COMMIT_ID ]; then
-      git checkout -b $API_VERSION $API_VERSION
-  else
-      git reset --hard $COMMIT_ID
-  fi
   popd
 fi
+
+pushd "$THIRD_PARTY_DIR/milvus-proto"
+# Always enforce the version selected by go.mod. A previously created checkout
+# may otherwise remain on a newer branch and generate incompatible Go sources.
+git fetch --tags origin
+COMMIT_ID=$(git ls-remote https://github.com/milvus-io/milvus-proto.git refs/tags/${API_VERSION} | cut -f 1)
+if [[ -z $COMMIT_ID ]]; then
+  # parse commit from pseudo version (eg v0.0.0-20230608062631-c453ef1b870a => c453ef1b870a)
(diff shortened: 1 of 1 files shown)


scripts/download_milvus_proto.sh
```diff
@@ -8,18 +8,22 @@ if [ ! -d "$THIRD_PARTY_DIR/milvus-proto" ]; then
...
-  echo "version: $API_VERSION, commitID: $COMMIT_ID"
-  if [ -z $COMMIT_ID ]; then
-      git checkout -b $API_VERSION $API_VERSION
-  else
-      git reset --hard $COMMIT_ID
-  fi
   popd
 fi
+
+pushd "$THIRD_PARTY_DIR/milvus-proto"
```
buqian-zilliz, line 14: **[P1] Validate the dependency checkout before running Git commands**

`-d` does not guarantee that this path is the milvus-proto repository. If a stale or partial cache leaves a plain directory here, `pushd` succeeds and the following Git commands discover the parent Milvus worktree instead. When `ls-remote` is unavailable, `COMMIT_ID` is empty and line 25 can create/reset and check out the Milvus release tag named by `API_VERSION`, switching the caller's main checkout. Verify that this directory is a Git worktree whose top level is exactly this path (or reclone it), and abort before running any Git command otherwise.

scripts/download_milvus_proto.sh
```diff
@@ -8,18 +8,22 @@ if [ ! -d "$THIRD_PARTY_DIR/milvus-proto" ]; then
...
+COMMIT_ID=$(git ls-remote https://github.com/milvus-io/milvus-proto.git refs/tags/${API_VERSION} | cut -f 1)
+if [[ -z $COMMIT_ID ]]; then
+  # parse commit from pseudo version (eg v0.0.0-20230608062631-c453ef1b870a => c453ef1b870a)
+  COMMIT_ID=$(echo $API_VERSION | awk -F'-' '{print $3}')
+fi
+echo "version: $API_VERSION, commitID: $COMMIT_ID"
+if [ -z $COMMIT_ID ]; then
+    git checkout -B "$API_VERSION" "$API_VERSION"
+else
+    git checkout --detach --force "$COMMIT_ID"
```
buqian-zilliz, line 27: **[P2] Propagate failures when pinning the requested revision**

`git checkout` failures are ignored because this script does not exit on errors and the following `popd` becomes its successful exit status. If a cached clone is shallow or the preceding fetch fails before the selected object is present, this checkout fails but proto generation continues from the stale revision—the exact incompatibility this change is intended to prevent. Propagate failures from the fetch, lookup, and checkout sequence and verify that the selected object is checked out before returning success.

scripts/download_milvus_proto.sh
```diff
@@ -8,18 +8,22 @@ if [ ! -d "$THIRD_PARTY_DIR/milvus-proto" ]; then
...
-  else
-      git reset --hard $COMMIT_ID
-  fi
   popd
 fi
+
+pushd "$THIRD_PARTY_DIR/milvus-proto"
+# Always enforce the version selected by go.mod. A previously created checkout
+# may otherwise remain on a newer branch and generate incompatible Go sources.
+git fetch --tags origin
```
buqian-zilliz, line 17: **[P2] Avoid unconditional remote lookups on every build**

`download-milvus-proto` is on every `generated-proto`/C++ build path, yet a warm cache still performs this fetch and the next line performs a separate `ls-remote` even when `HEAD` already matches `go.mod`. The code-checker also invokes the dependency through two separate Make processes, so these round trips are repeated in one job; a fresh clone redundantly fetches again as well. Resolve the requested tag/hash locally first and contact the remote only when the object is missing.

zhengbuqian (approved):

Issue #51344, opened 2026-07-14
[Bug]: StorageV3 write-buffer sync appends 7 groups to a recovered 6-group manifest after AddField and blocks flush


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

Related issues:

- #50697 was fixed by #50738. It covered growing-source flush writing an incompatible SINGLE/schema-based layout and its retry loop.
- #51275 covers loading manifests containing dropped-field column groups.
- This failure uses a different path: growing-source successfully commits a six-group layout, the new StreamingNode recovers that layout, then a normal write-buffer `SyncTask` tries to append seven groups.

### Environment

- Milvus image: `harbor.tencent-stage.zilliz.cn/milvusdb/milvus:master-20260713-4dbaba00`
- Commit: `4dbaba0042d3952fa75bbb5d2fb9606c6b67f44d`
- Deployment mode: Cluster
- Helm chart: `milvus-5.0.24`
- MQ: Pulsar v3
- Storage:
  - Storage V3 enabled
  - Loon FFI enabled
  - Growing-source flush enabled
- Streaming service: enabled
- Python: `3.12.13`
- PyMilvus: `3.1.0rc62`
- Kubernetes namespace: `chaos-testing`
- Instance: `etcd-leader-pod-kill-364`
- Milvus replicas:
  - StreamingNode: 2
  - DataNode: 2
  - QueryNode: 3
  - Proxy: 1
  - MixCoord: 1
- Milvus component resources:
  - Most components: request `2 CPU / 8Gi`, limit `8 CPU / 16Gi`
  - MixCoord: request `2 CPU / 8Gi`, limit `16 CPU / 32Gi`
- GPU: N/A
- OS: Linux containers on Kubernetes. The exact node distribution was unavailable because the instance had already been torn down.

Jenkins build:

http://jenkins.milvus-cell.tc-op.zilliz.cn/blue/organizations/jenkins/chaos-test-cron/detail/chaos-test-cron/364/pipeline

### K8s Pod List

Snapshot at `2026-07-14 06:25:12 UTC`, before teardown:

```text
NAME                                                                      READY   STATUS      RESTARTS
etcd-leader-pod-kill-364-0                                                1/1     Running     0
etcd-leader-pod-kill-364-1                                                1/1     Running     0
etcd-leader-pod-kill-364-2                                                1/1     Running     0
[...]

PR #51581, opened 2026-07-18
fix: flush all-null nullable column on growing-source flush

MrPresent-Han:
issue: #51344

## Problem

A growing segment after `AddField` can hold an all-null nullable column: `valid_data` present, vector chunk empty. The flush-layout predicate keyed on `entry->empty()`, treated the column as non-materialized and dropped it, so growing-source flush committed a manifest with a missing column group. The later write-buffer `SyncTask` emits that column, Loon rejects the append (`Column group size mismatch`), and the deterministic error is retried forever — pinning the channel checkpoint until `flush()` times out. `empty()`/`valid_count` are also mutable chunk state and race with a concurrent first-row insert (thanks @liliu-z), which could drop a normal column the same way.

## Fix

Flush-layout membership now comes from a monotonic per-column `materialized` marker:

- set at the first `set_data_raw` (every materialization path funnels there), never cleared;
- `get_data_field_ids` = `has_field(f) && materialized(f)`; `FlushGrowingSegmentData` gates on the same marker;
- immune to the first-insert race (the marker only flips false→true) and to interim-index raw-vector `clear()`.

Also in this PR:

- `FlushGrowingSegmentData` takes one runtime-schema snapshot (`get_schema_snapshot`) instead of bare `get_schema()` references — closes a UAF window against concurrent `Reopen` and pins all dropped-field checks to a single schema version.
- Free the strdup'd `status.error_msg` in `FlushRejectsEndOffsetBeyondRowCount` — pre-existing leak from #51532 that failed LSan `ut-cpp` for every C++ PR.

## Test

New segcore UTs: all-null nullable column kept in the layout; dropped / non-materialized function-output fields skipped; marker stays set across `clear()`.

## Known limitations

- Prevention-only: a manifest already committed with a missing group is not self-healed.
- DropField after the field entered the committed manifest still ends in a stuck (never silently lossy) sync — broken in all pre-PR variants too; whether Loon tolerates the resulting group-count shrink on append is storage-side and unverified here. Proper fix (all-null synthesis or seal-on-drop) is a follow-up.
- Retry classification for the deterministic Loon group mismatch is a separate issue.

### DIFF
--- a/internal/core/src/segcore/InsertRecord.h
+++ b/internal/core/src/segcore/InsertRecord.h
@@ -1733,21 +1733,30 @@ class InsertRecordGrowing {
         return data_.find(field_id) != data_.end();
     }
 
-    // Field ids that have materialized columns. The exact set the flush
-    // layout must be trimmed to: non-materialized function outputs are
-    // absent here.
+    // True once set_data_raw has run on the field's column (see
+    // VectorBase::materialized).
+    bool
+    materialized(FieldId field_id) const {
+        std::shared_lock<std::shared_mutex> lck(field_map_mutex_);
+        auto it = data_.find(field_id);
+        return it != data_.end() && it->second && it->second->materialized();
+    }
+
+    // Field ids in the flush layout: columns the segment has materialized
+    // (see VectorBase::materialized) that are still present in `schema`.
+    // `schema` MUST be the segment's own runtime schema, NOT the staler flush
+    // snapshot — else a dropped field would be wrongly kept. Dropped fields and
+    // never-materialized function outputs (backfilled later by bump-schema
+    // compaction) stay out.
     std::vector<int64_t>
-    get_data_field_ids() const {
+    get_data_field_ids(const Schema& schema) const {
         std::shared_lock<std::shared_mutex> lck(field_map_mutex_);
         std::vector<int64_t> ids;
         ids.reserve(data_.size());
         for (const auto& [field_id, entry] : data_) {
-            // An allocated-but-empty column (e.g. a function output the
-            // replayed older-era inserts never filled) is not materialized.
-            if (!entry || entry->empty()) {
-                continue;
+            if (entry && schema.has_field(field_id) && entry->materialized()) {
+                ids.push_back(field_id.get());
             }
-            ids.push_back(field_id.get());
         }
         return ids;
     }
--- a/internal/core/src/segcore/ConcurrentVector.h
+++ b/internal/core/src/segcore/ConcurrentVector.h
@@ -203,8 +203,24 @@ class VectorBase {
         return empty;
     }
 
+    // Monotonic materialization marker, the canonical flush-layout presence
+    // test. Set on the first set_data_raw (any path, incl. all-null bulk fill),
+    // never cleared. Avoids the race-prone chunk state empty()/valid_count,
+    // which a concurrent insert flips mid set_data_raw and interim-index
+    // raw-vector removal empties for a column that still exists.
+    bool
+    materialized() const {
+        return materialized_.load(std::memory_order_relaxed);
+    }
+
  protected:
+    void
+    mark_materialized() {
+        materialized_.store(true, std::memory_order_relaxed);
+    }
+
     const int64_t size_per_chunk_;
+    std::atomic<bool> materialized_{false};
 };
 
 template <typename Type, bool is_type_entire_row = false>
@@ -288,6 +304,9 @@ class ConcurrentVectorImpl : public VectorBase {
     set_data_raw(ssize_t element_offset,
                  const void* source,
                  ssize_t element_count) override {
+        // Entry, not post-chunk: an all-null column reaches here with
+        // valid_count==0 (no chunk emplaced), yet is still materialized.
+        mark_materialized();
         ssize_t valid_count = 0;
         ssize_t storage_offset = 0;
         if (use_mapping_storage_) {
--- a/internal/core/src/segcore/SegmentGrowingImpl.h
+++ b/internal/core/src/segcore/SegmentGrowingImpl.h
@@ -218,6 +218,15 @@ class SegmentGrowingImpl : public SegmentGrowing {
         return *schema_;
     }
 
+    // Reopen swaps schema_ under sch_mutex_; a bare get_schema() reference can
+    // dangle if the swap frees the old Schema mid-use. Snapshot under the lock
+    // (shared_ptr copy keeps the referent alive). Mirrors Collection::get_schema.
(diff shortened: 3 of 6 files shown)


internal/core/src/segcore/InsertRecord.h
```diff
@@ -1733,18 +1734,34 @@ class InsertRecordGrowing {
...
+            }
+            // An empty-chunk column is kept only for a TRUE all-null nullable
+            // ordinary column. IsAllNullNullable (valid offset mapping enabled
+            // with zero valid rows), not empty(), is the robust all-null test:
+            // interim-index raw-vector removal empties the chunk of a column
+            // that still HAS data, and the valid offset mapping is unaffected.
+            // Such an emptied-but-non-null column stays out here (its data is
+            // served from the index / flushed via the index path), not misread
+            // as all-null.
+            if (entry->empty() && (!schema.has_field(field_id) ||
```
liliu-z: get_data_field_ids gates a field's inclusion on two non-atomic live reads — entry->empty() then IsAllNullNullable(entry->get_offset_mapping()) — but a concurrent insert's set_data_raw updates offset_mapping_ (valid_count) before it emplaces the chunk, so a nullable field receiving its first valid rows can be observed as empty()==true with valid_count>0 and dropped from MaterializedFieldIDs. Because the trimmed layout is persisted as the segment's sticky currentSplit (growing_source.go:694, reused verbatim at task.go:254) and trimColumnGroupsToMaterialized only ever removes fields, that transient drop is permanent and silently loses the column. This is a pre-existing empty()-gated race the predicate rewrite narrows but does not close (the race is not about sch_mutex_ sharing — get_data_field_ids never takes sch_mutex_); a robust fix needs a consistent snapshot excluding concurrent set_data_raw rather than two independent reads of live chunk state.

MrPresent-Han: Confirmed, and as you say it's pre-existing: `empty()`/`valid_count` are mutable chunk state a concurrent insert flips mid-`set_data_raw` (`offset_mapping_.Append` runs before the chunk emplace), and the sticky `currentSplit` makes the transient drop permanent. My earlier predicate narrowed it but didn't close it.

Instead of snapshotting/fencing the two reads, I removed the dependency on mutable chunk state entirely. Flush-layout membership is now decided by a monotonic per-column `materialized` marker, set at the first `set_data_raw` (any path — insert, all-null bulk fill, Reopen) and never cleared. A concurrent first-row insert can only flip it `false`→`true` (fail-safe keep), and interim-index raw-vector removal (`clear()`) can't drop a still-present column. `get_data_field_ids` collapses to `has_field && materialized()`, and the `segment_c.cpp` flush gate uses the same marker — which also aligns growing-source flush with the write-buffer `SyncTask`, which derives its layout from schema/buffered data and never reads live segment state.

Pushed as an amend. Thanks for catching this.

internal/core/src/segcore/InsertRecord.h
```diff
@@ -1733,21 +1733,30 @@ class InsertRecordGrowing {
...
+    get_data_field_ids(const Schema& schema) const {
         std::shared_lock<std::shared_mutex> lck(field_map_mutex_);
         std::vector<int64_t> ids;
         ids.reserve(data_.size());
         for (const auto& [field_id, entry] : data_) {
-            // An allocated-but-empty column (e.g. a function output the
-            // replayed older-era inserts never filled) is not materialized.
-            if (!entry || entry->empty()) {
-                continue;
+            if (entry && schema.has_field(field_id) && entry->materialized()) {
```
liliu-z, line 1757: get_data_field_ids now gates membership on schema.has_field(field_id), so a column that was materialized and already flushed into the segment's manifest before its field was dropped is excluded from every later growing-source flush: Reopen leaves the column in data_ but has_field returns false once schema_ advances past the drop. Growing-source flush appends to the existing manifest (growingSourceReadVersion reuses the prior version), so the post-drop append commits fewer column groups than the committed manifest — the same cross-append group-count inconsistency this PR fixes for the all-null case, only as a shrink. Whether this re-triggers the checkpoint-pinning failure depends on milvus-storage tolerating a dropped-field group-count shrink on append, which is external to this repo (the PR's own 'Known limitations' flags it as an unverified follow-up). Confirm that tolerance against pin e658197, or retain fields already present in the committed manifest/currentSplit, before relying on the dropped-field path.

MrPresent-Han: Confirmed end to end — your read is right. With a committed manifest, `has_field` excludes a materialized-then-dropped column from the re-sync layout, so the growing-source append commits fewer groups than the committed split (a shrink), symmetric to the all-null grow case this PR fixes.

Findings from tracing it:

- **Reachability**: real. DropField advances the growing segment's `schema_` *lazily* via query `LazyCheckSchema → Reopen`. Once advanced, `has_field(c)=false`, and the Insert loop (iterates `schema_`, SegmentGrowingImpl.cpp:694) stops writing c while row_ids/timestamps/ack keep growing — so c's `ConcurrentVector` freezes *shorter than the segment row count*. Reopen does **not** erase c (`drop_field_data` has no callers). On a recovered segment, c is absent from `data_` entirely (the ctor builds columns from the current schema, which no longer has c).

- **Failure mode**: **stuck, not silent data loss**. Growing-source buffering is progress-only (no Go-side dual-buffer, l0_write_buffer.go:68-74), so on the mismatch `markNonRetryableFailure` strands the segment: it's filtered from every later sync selection, the retained-pin is only released on a flush commit that never happens, and the checkpoint stays pinned. Data remains in WAL + segcore memory — never acked durable, never discarded. Deterministic on restart.

- **Shrink tolerance**: `Column group size mismatch` is thrown in the Loon/packed writer (external to this repo), so I can't confirm shrink-vs-grow tolerance in-tree — matches the Known-limitations note.

The invariant this needs: once a manifest with c is committed, every later sync's columnGroups must equal the committed `currentSplit` (no shrink/grow), and a `currentSplit` field absent from current data/schema is flushed **all-null without reading the segment's c** (live: misaligned/OOB read; recovered: absent). It has to hold on **both** the growing-source (C++) and the post-restart old-sync (Go) serializers, and the dropped field's arrow meta has to come from the committed/versioned schema, not `schema_`.

Taking your second suggestion ("retain fields already present in the committed manifest/currentSplit") as the basis. Since the complete fix spans both serializers + historical-schema meta and is orthogonal to the all-null materialization + the concurrency/schema races fixed here, I'll land it as a **separate PR/issue** and keep this one scoped to those. The dropped-field path is stuck-not-lost, so it never silently corrupts. Thanks for the precise catch.

PR #51583, opened 2026-07-18
fix: [3.0] size the scalar-index valid bitmap by active count on growing segments (#51523)

xiaocai2333:
issue: #51237
pr: #51523

Backport of #51523 to 3.0 (clean cherry-pick — the net diff is identical to #51523's).

## Root cause

A growing segment carries a segment-level interim R-Tree index for a geometry field as soon as the collection has an `RTREE` index on it, making geometry the only field type whose expressions take the **ScalarIndex** path on a growing segment. The index bitmap is segment-global (a scalar index always has exactly one chunk), but the batching code sliced it with `size_per_chunk_` as the upper bound — on a growing segment that is the raw-data chunk granularity (`segcore.chunkRows`, 128 by default), not the row count. A `geometry_field is not null` query over more than one chunk of growing rows returned a truncated bitmap and crashed `FilterBitsNode`:

```
Assert "col_vec_size == need_process_rows_" => bitset size: 128, need_process_rows_: 5000
```

## Fix

- `ProcessIndexChunksForValid()`: slice the cached valid bitmap by `active_count_` instead of `size_per_chunk_`, and assert the bitmap covers the requested range.
- `ProcessIndexChunks()`: same slicing for the result bitmap. The old third `min` term was also wrong on its own — it clamped against the bitmap's *full* length rather than the length remaining after `data_pos`, so a short bitmap plus `data_pos > 0` made `append()` read past the end.
- The `execute_all_at_once_` fast path: only move the cached bitmap wholesale when it lines up exactly with `active_count_`; otherwise fall through to the slicing branch, which bounds by `active_count_` and asserts coverage.
- `MoveCursorForIndex()`: drop the sealed-only assert — the index cursor is a global row position for growing segments with a segment-level index too.

## Note on the GIS cursor fix

An earlier revision of this branch also carried a `PhyGISFunctionFilterExpr::MoveCursor()` change, because 3.0 lacked master's equivalent and the short-circuit regression test below failed without it. #51488 ("[3.0] gate GIS filters out of the offset-input path") has since landed on 3.0 and provides it, so that commit was dropped on rebase. This PR is now exactly #51523.

## Tests

- `TestGeometryNullExprOnGrowingWithRTreeIndex` — `is not null` / `is null` on a growing segment with the R-Tree index, plus a batched conjunction and the short-circuit path.
- `TestGeometrySpatialExprOnGrowingWithRTreeIndex` — real spatial predicates, indexed vs raw growing segment over the same rows, must select identical rows.
- `TestGeometryConjunctionShortCircuitOnGrowingWithRTreeIndex` — conjunction whose numeric side short-circuits the early batches, anchored on a standalone spatial run rather than an indexed-vs-raw comparison alone.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -392,8 +392,9 @@ class SegmentExpr : public Expr {
 
     void
     MoveCursorForIndex() {
-        AssertInfo(segment_->type() == SegmentType::Sealed,
-                   "index mode only for sealed segment");
+        // The index cursor is a global row position. This holds for sealed
+        // segments and for growing segments with a segment-level scalar
+        // index (the geometry interim R-Tree, see issue #51237).
         auto size =
             std::min(active_count_ - current_index_chunk_pos_, batch_size_);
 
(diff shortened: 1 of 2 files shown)


liliu-z (commented):
/lgtm
/approve

PR #47466, opened 2026-02-01
fix: Fix macOS 15+ build compatibility for Apple Silicon

xiaofan-luan:
issue: #47460

PR #51585, opened 2026-07-18
enhance: make Tantivy Cargo discovery portable

FantasqueX:
Find Cargo through CMake with Rustup installation directories as hints instead of requiring $HOME/.cargo/bin to exist. Use the actual built and installed libraries as custom-command outputs so the generated build graph tracks Tantivy artifacts correctly.

This fixes issue when installing rustup using apt.

This adapts the Tantivy build-rule changes introduced on master by commit 4f902f1f216bbe3795a303c65addccfdeefb6cf0 (#47466) for the 2.6 branch.

pr: #47466

Assisted by gpt-5.6-sol

### DIFF
--- a/internal/core/thirdparty/tantivy/CMakeLists.txt
+++ b/internal/core/thirdparty/tantivy/CMakeLists.txt
@@ -1,11 +1,20 @@
 set(TANTIVY_FEATURES_LIST "" CACHE STRING "List of Cargo features to enable")
 string(REPLACE ";" "," TANTIVY_FEATURES "${TANTIVY_FEATURES_LIST}")
 
+set(CARGO_SEARCH_PATHS "$ENV{HOME}/.cargo/bin")
+if (NOT "$ENV{CARGO_HOME}" STREQUAL "")
+    list(PREPEND CARGO_SEARCH_PATHS "$ENV{CARGO_HOME}/bin")
+endif ()
+find_program(CARGO_EXECUTABLE
+        NAMES cargo
+        HINTS ${CARGO_SEARCH_PATHS}
+        REQUIRED)
+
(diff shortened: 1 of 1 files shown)


internal/core/thirdparty/tantivy/CMakeLists.txt
```diff
@@ -23,34 +32,38 @@ set(TANTIVY_NAME "libtantivy_binding${CMAKE_STATIC_LIBRARY_SUFFIX}")
...
-
-add_custom_command(OUTPUT ls_cargo
-        COMMENT "ls cargo"
-        COMMAND ls ${HOME_VAR}/.cargo/bin/
-        )
-add_custom_target(ls_cargo_target DEPENDS ls_cargo)
-
-add_custom_command(OUTPUT compile_tantivy
+add_custom_command(
+        OUTPUT ${LIB_FILE}
```
buqian-zilliz: **[P1] Track Cargo inputs before using the archive as an output**

The new rule declares `${LIB_FILE}` as the output but does not declare any Cargo inputs as dependencies. `scripts/core_build.sh` reuses its build directory with the Unix Makefiles generator, so after the first successful build, changes to `Cargo.toml`, `Cargo.lock`, `build.rs`, Rust sources, `${TOKENIZER_PROTO}`, or `TANTIVY_FEATURES_LIST` leave the archive up to date and Cargo is not invoked. Milvus can therefore silently install and link a stale Tantivy library. Please keep the Cargo step phony with the archive as a byproduct, or model all file and configuration inputs with a dependency/stamp scheme.

FantasqueX: Fixed. I just backported from master. Master branch seems to be affected as well.

zhengbuqian (approved):

Issue #51586, opened 2026-07-18
Iterative-filter multi-query top-k ordering bug and several segcore NULL 3VL correctness bugs


xiaofan-luan:
## Summary

While adversarially reviewing the iterative-filter / expression-eval **offset path**, several correctness bugs surfaced. They fall into two groups.

### 1. Multi-query iterative-filter top-k ordering

`SearchGroupByOperator`-adjacent `insert_helper` in `IterativeFilterNode` indexes the per-query result window by the **relative** count instead of the **absolute** window end (`base_idx + count`). For `nq_index >= 1` the shift guard `count > pos` is always false, so an out-of-order candidate (a closer distance arriving after farther ones — which approximate vector iterators do emit) overwrites an existing top-k entry instead of shifting it. Result: the top-k of every query vector **after the first** in a multi-query filtered search is corrupted. The sibling `IterativeElementFilterNode` already uses the correct absolute-window math, confirming the intended behavior.

### 2. NULL three-valued-logic / null-batch-callback correctness on the offset path

- **`ExistsExpr`**: a column-null row sets `res=false` but leaves `valid_res=true`, so `NOT exists(json[path])` wrongly matches null-JSON rows.
- **`ProcessIndexLookupByOffsets`**: (a) a null/absent index row (`Reverse_Lookup == nullopt`) leaves `valid_res=true` (same `NOT` leak) and never drives the callback, desyncing the cursor-tracking `processed_cursor` under an iterative-filter `AND`; (b) the SkipIndex skip branch sets `res=true`, but a pruned chunk cannot match — it should be `false`.
- **`ProcessElementLevelByOffsets` / `ProcessDataChunksForElementLevel`** skip branches don't drive the null-batch callback, so cursor-tracking callbacks desync (parity gap with the doc-level skip branches).
- **`SearchGroupByOperator`**: group-by on an index-only nullable scalar field throws `AssertInfo("field data not found")` on a null row instead of forming a distinct null group like the from-data path does.
- **`ChunkedColumn` / `ProxyChunkColumn` `BulkIsValid`**: the non-nullable branch falls through into the nullable branch, invoking the callback twice per row (latent — currently guarded by callers).

## Impact

Wrong / mis-ordered search results on multi-query filtered search (group 1); wrong `NOT` results on nullable columns, cursor desync under iterative-filter `AND`, and a group-by query crash on index-only nullable fields (group 2).

[...]

PR #51587, opened 2026-07-18
fix: iterative-filter top-k ordering and segcore NULL 3VL correctness

xiaofan-luan:
Two groups of correctness fixes found while adversarially reviewing the iterative-filter / expression-eval **offset path**. Filed together per request; see the sections below for independent review.

## 1. Multi-query iterative-filter top-k insertion

`insert_helper` indexed the per-query result window by the **relative** `count` instead of the **absolute** window end (`base_idx + count`). For `nq_index >= 1` (`base_idx > 0`) the shift guard `count > pos` is always false, so an out-of-order candidate (a closer distance arriving after farther ones — which approximate vector iterators do emit) overwrote an existing top-k entry instead of shifting it, corrupting the top-k of **every query vector after the first** in a multi-query filtered search.

Fix: extracted the insertion into a single shared `Utils.h::topk_binsert` with the correct absolute-window math, used by **both** iterative-filter nodes so the two paths cannot diverge again. The sibling `IterativeElementFilterNode` already had the correct math — this removes the divergent copy that caused the bug.

Regression test: `IterativeFilterInsert.MultiQueryWindowOrdering` (deterministic; inserts an out-of-order stream into the second query window and asserts the correct sorted result — the previous relative-count math corrupts it).

## 2. NULL 3VL / null-batch-callback correctness on the offset path

- **`ExistsExpr`**: a column-null row set `res=false` but left `valid_res=true`, so `NOT exists(json[path])` wrongly matched null-JSON rows → `res=valid=false`.
- **`ProcessIndexLookupByOffsets`**: (a) null/absent index row (`Reverse_Lookup == nullopt`) left `valid_res=true` and never drove the callback, desyncing `processed_cursor` under iterative-filter `AND` → now `res=valid=false` + null-batch callback; (b) the SkipIndex skip branch set `res=true`, but a pruned chunk cannot match → `res=false` with validity from the index + null-batch callback, matching every other skip branch.
- **`ProcessElementLevelByOffsets` / `ProcessDataChunksForElementLevel`** skip branches now drive the null-batch callback, keeping cursor-tracking callbacks aligned with the doc-level skip branches.
- **`SearchGroupByOperator`**: group-by on an index-only nullable scalar field threw `AssertInfo("field data not found")` on a null row → returns `std::nullopt` to form a distinct null group, consistent with the from-data getters.
- **`ChunkedColumn` / `ProxyChunkColumn` `BulkIsValid`**: the non-nullable branch fell through into the nullable branch (invoking the callback twice per row) → added the missing `return`.

## Verification

Group 1 has a deterministic unit test (green with the fix). Group 2 applies the established null-rejecting / null-batch-callback contract and is covered here by compilation + no-regression against the existing expression suite; dedicated null regression tests (NOT exists on nullable JSON, NOT over a nullable indexed field via offset input, index-only nullable group-by) are a planned follow-up.

Deferred (documented, low priority): `LogicalBinaryExpr` 3VL merge (dead code — all AND/OR route to `ConjunctExpr`), `BitmapIndex` empty-array-on-legacy-file (narrow, safe direction), `UnaryExpr` array `is_same_array` valid-bit (unverified).

issue: #51586

### DIFF
--- a/internal/core/src/exec/expression/ExistsExpr.cpp
+++ b/internal/core/src/exec/expression/ExistsExpr.cpp
@@ -181,7 +181,10 @@ PhyExistsFilterExpr::EvalJsonExistsForDataSegment(EvalCtx& context) {
                 offset = (offsets) ? offsets[i] : i;
             }
             if (valid_data != nullptr && !valid_data[offset]) {
-                res[i] = false;
+                // Column-null row: three-valued logic requires marking it
+                // invalid too, else a wrapping NOT flips res=false to true and
+                // wrongly matches the null row.
+                res[i] = valid_res[i] = false;
                 continue;
             }
             if (has_bitmap_input && !bitmap_input[processed_cursor + i]) {
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -735,7 +735,20 @@ class SegmentExpr : public Expr {
                 auto offset = (*input)[i];
                 auto raw = index_ptr->Reverse_Lookup(offset);
                 if (!raw.has_value()) {
-                    res[i] = false;
+                    // Null/absent row: three-valued logic requires res=false
+                    // AND valid=false (else a wrapping NOT matches the null
+                    // row). Still drive the callback on a null batch so
+                    // cursor-tracking callbacks keep processed_cursor aligned
+                    // — otherwise later candidates read the wrong bitmap_input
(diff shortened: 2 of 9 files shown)


xiaofan-luan:
Folded into #51540 (combined into one PR per request). The commit is now on #51540's branch (`fix/expr-offsets-eval-correctness`, fast-forward). Closing this in favor of the combined #51540.

Issue #51588, opened 2026-07-18
[Bug]: HandleCommitVchannel can make import segments visible after rejecting commit timestamp


bigsheeper:
## Bug Report

`HandleCommitVchannel` can return success and clear `is_importing` for an import segment even when `UpdateCommitTimestamp` rejects the commit timestamp because it is lower than the segment's max `binlog.TimestampTo`.

### Impact

This can make an import/CDC segment visible without a valid `commit_timestamp`. The StreamingNode WAL flusher treats a successful `HandleCommitVchannel` response as the commit fence being accepted and then observes the WAL message, so it will not retry the failed visibility update.

### Root cause

`UpdateCommitTimestamp(segID, commitTs)` logs and returns `false` when `commitTs < max(binlog.TimestampTo)`, but it does not set `updateSegmentPack.err`. `meta.UpdateSegmentsInfo` continues to the next operator, so `UpdateIsImporting(segID, false)` still mutates and persists the segment. `importMeta.HandleCommitVchannel` then records the vchannel as committed.

Production path:

`StreamingNode WAL flusher dispatchCommitImport` -> `MixCoord.HandleCommitVchannel` -> `DataCoord Server.HandleCommitVchannel` -> `meta.UpdateSegmentsInfo(UpdateCommitTimestamp(segID, commitTs), UpdateIsImporting(segID, false))`.

### Repro evidence

I added this failing test before the fix:

```bash
GOWORK=off PKG_CONFIG_PATH=$PWD/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=$PWD/internal/core/output/lib go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord -run TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp
```

Observed on current master:

```text
ERROR datacoord/meta.go:1748 meta update: update commit timestamp rejected - commit_ts < max(binlog.TimestampTo) segmentID=10 commitTs=300 maxBinlogTimestampTo=500
--- FAIL: TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp
    services_test.go:5829: Should be false
    services_test.go:5834: Should be true
```

The failed assertions show that the RPC status is successful and the segment no longer has `is_importing=true` after the rejected timestamp update.

### Expected behavior

[...]

PR #51589, opened 2026-07-18
fix: reject invalid import commit timestamp

bigsheeper:
issue: #51588

## What

`UpdateCommitTimestamp` detects the `commit_ts < max(binlog.TimestampTo)`
invariant violation but returned `false` without setting an error on the update
pack. `UpdateSegmentsInfo` ignores an operator's bool return and aborts only on
`updatePack.err`, so the batch issued by `HandleCommitVchannel` continued into
`UpdateIsImporting(false)`, persisted the segment and returned success — the
segment was published with `commit_timestamp` still 0, which is precisely the
state the invariant exists to prevent.

This makes the rejection abort the batch, so `is_importing` stays set and the
commit fence is not acknowledged.

## Why this blocks instead of failing the job

The error is deliberately left retriable (`ErrImportSysFailed`, SystemError).
`retry.Do` aborts by default only on `InputError`, and the StreamingNode flusher
wraps `HandleCommitVchannel` in `retry.Do(..., retry.AttemptAlways())`, so a
rejected fence blocks that pchannel's WAL replay until an operator intervenes.
That is the intent, not an oversight:

- The violation is expected to be **unreachable** for a normal import: its rows
  are stamped with the Import message's timetick (`DataTs` =
  `BroadcastResult.GetMaxTimeTick()`) and the commit fence is a later timetick on
  the same WAL. Only a binlog/backup import, whose rows keep the source cluster's
  original timestamps, can carry `max(binlog.TimestampTo)` above the local fence.
- Reaching it means an assumption has broken, so stopping beats proceeding on it.
- Blocking is observable: a stalled pchannel shows up in WAL/replication lag.
- Failing the import job instead would be worse under replication. The predicate
  is evaluated independently on each cluster against its own local TSO, so the
  source can reject a fence the replica accepts. Once the replica's job is
  `Committing`/`Completed`, `rollbackImportV2AckCallback` no-ops, so that
  divergence cannot be rolled back by any existing path. A wedged pchannel stops
  the cluster loudly; a diverged replica corrupts it quietly.

### What "an operator intervenes" means here

Recovery is deliberately manual. Once `HandleCommitVchannel` has flipped the job
to `Committing` — persisted before the callback runs (`import_meta.go:365-374`) —
every automatic path out is closed, and that is accepted rather than overlooked:

- `AbortImport` rejects `Committing` (`services.go:3028-3031`).
- The `RollbackImport` ack callback no-ops on `Committing`
  (`ddl_callbacks_import.go:304-310`).
- `tryTimeoutJob` does cover `Committing`, but compares against `TimeoutTs`,
  which `importutilv2.GetTimeoutTs` leaves at `math.MaxUint64` when the request
  carries no `timeout` option (`option.go:79`) — so by default it never fires.
  An import submitted **with** a `timeout` option does fail automatically, and
  the fence then clears via the `Failed` branch (`import_meta.go:351-352`).
- `checkCommittingJob` has only a success branch (`import_checker.go:522-541`),
  and `checkGC` handles `Completed`/`Failed` only (`:609-611`).
- The flusher retries forever, so a restart replays the same WAL message into the
  same rejection.

Detection is therefore by monitoring rather than by job state: the pchannel stops
advancing (WAL / replication lag) and the rejection is logged at `Error` on every
attempt (~1 line / 3 s per stuck vchannel). Resolution is an operator diagnosing
why the invariant broke — clock skew between the source of a binlog/backup import
and this cluster being the realistic cause — and then clearing the job out of
band. That cost is accepted in exchange for never publishing a segment with
`commit_timestamp = 0`.

## Tests

- `GOWORK=off PKG_CONFIG_PATH=$PWD/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=$PWD/internal/core/output/lib go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord -run 'Test(UpdateCommitTimestamp|HandleCommitVchannelRPC)'`
- `GOWORK=off PKG_CONFIG_PATH=$PWD/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=$PWD/internal/core/output/lib go build -tags dynamic ./internal/datacoord/`

## Not verified

- No end-to-end reproduction of the violation; the unreachability argument above
  is derived from code, not measured.
- Import in a replicating cluster is gated by
  `dataCoord.import.enableInReplicatingCluster` (default `false`), so the
  cross-cluster case only applies when that switch is enabled.
- `make static-check` currently fails in my local checkout with
  `no go files to analyze` on an unmodified tree as well, so it is an
  environment issue rather than a result of this change; CI covers it.

### DIFF
--- a/internal/datacoord/commit_timestamp_test.go
+++ b/internal/datacoord/commit_timestamp_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/msgpb"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/tsoutil"
 )
 
@@ -122,9 +123,7 @@ func TestUpdateCommitTimestamp_RejectBelowMaxTimestampTo(t *testing.T) {
 	assert.NoError(t, meta.AddSegment(context.Background(), segWithBinlogs(1, 5000)))
 
 	err = meta.UpdateSegmentsInfo(context.Background(), UpdateCommitTimestamp(1, 3000))
-	// UpdateSegmentsInfo returns nil when the operator returns false (no-op);
-	// the assertion is on the persisted state, not the error.
-	assert.NoError(t, err)
+	assert.ErrorIs(t, err, merr.ErrImportSysFailed)
 	assert.Equal(t, uint64(0), meta.GetSegment(context.Background(), 1).GetCommitTimestamp(),
 		"rejected update must leave commit_timestamp unchanged")
 }
--- a/internal/datacoord/meta.go
+++ b/internal/datacoord/meta.go
@@ -1846,7 +1846,15 @@ func UpdateCommitTimestamp(segmentID int64, ts uint64) UpdateOperator {
 					mlog.Int64("segmentID", segmentID),
 					mlog.Uint64("commitTs", ts),
 					mlog.Uint64("maxBinlogTimestampTo", maxTsTo))
-				return false
+				// Fail-stop. Unreachable for a normal import: its rows carry the
+				// Import message's timetick and the commit fence is a later
+				// timetick on the same WAL. Keep the error retriable so the
+				// flusher blocks on the fence instead of failing the job — a
+				// blocked pchannel surfaces as WAL lag, whereas a replica that
+				// commits what the source rejected can no longer be rolled back.
+				return modPack.fail(merr.WrapErrImportSysFailedMsg(
+					"commit timestamp %d is less than max binlog timestamp %d for import segment %d",
+					ts, maxTsTo, segmentID))
 			}
 		}
 		segment.CommitTimestamp = ts
--- a/internal/datacoord/meta.go
+++ b/internal/datacoord/meta.go
@@ -1813,6 +1813,27 @@ func UpdateIsImporting(segmentID int64, isImporting bool) UpdateOperator {
 	}
 }
 
+// maxBinlogTimestampTo returns the highest TimestampTo across a segment's insert
+// binlogs, or 0 when the arrays are absent.
+//
+// Absent is not the same as "no rows": a V3 (manifest-backed) segment never
+// persists these arrays -- buildAlterSegmentsKvs skips the per-FieldBinlog KVs
+// for it (kv_catalog.go:357) and the SegmentInfo is written without them -- so a
+// V3 segment reloaded after a DataCoord restart reports 0 here regardless of the
+// row timestamps it actually holds. Callers therefore get a bound that is safe
+// to compare against but that does not fire for reloaded V3 segments.
+func maxBinlogTimestampTo(fieldBinlogs []*datapb.FieldBinlog) uint64 {
+	var maxTsTo uint64
+	for _, fb := range fieldBinlogs {
+		for _, l := range fb.GetBinlogs() {
+			if l.GetTimestampTo() > maxTsTo {
+				maxTsTo = l.GetTimestampTo()
+			}
+		}
+	}
+	return maxTsTo
+}
+
 // UpdateCommitTimestamp sets the commit_timestamp on an import/CDC segment.
 // Non-zero marks it as committed at that transaction time, overriding
 // start_position.Timestamp for all temporal decisions.
@@ -1833,14 +1854,7 @@ func UpdateCommitTimestamp(segmentID int64, ts uint64) UpdateOperator {
 			return false
 		}
 		if ts != 0 {
-			var maxTsTo uint64
-			for _, fieldBinlogs := range segment.GetBinlogs() {
-				for _, l := range fieldBinlogs.GetBinlogs() {
-					if l.GetTimestampTo() > maxTsTo {
-						maxTsTo = l.GetTimestampTo()
-					}
-				}
-			}
+			maxTsTo := maxBinlogTimestampTo(segment.GetBinlogs())
 			if ts < maxTsTo {
 				mlog.Error(context.TODO(), "meta update: update commit timestamp rejected - commit_ts < max(binlog.TimestampTo)",
 					mlog.Int64("segmentID", segmentID),
@@ -1852,6 +1866,14 @@ func UpdateCommitTimestamp(segmentID int64, ts uint64) UpdateOperator {
 				// flusher blocks on the fence instead of failing the job — a
 				// blocked pchannel surfaces as WAL lag, whereas a replica that
 				// commits what the source rejected can no longer be rolled back.
+				//
+				// Recovery from here is out of band. The job is already
+				// Committing by the time this runs -- commitImportV2AckCallback
+				// persists that state on the broadcast FastAck, independent of
+				// this fence -- and Committing is terminal for AbortImport while
+				// tryTimeoutJob never reaches it (TimeoutTs defaults to
+				// math.MaxUint64). Validating earlier does not change that: the
+				// ack path flips the state regardless of what this check says.
 				return modPack.fail(merr.WrapErrImportSysFailedMsg(
 					"commit timestamp %d is less than max binlog timestamp %d for import segment %d",
 					ts, maxTsTo, segmentID))
--- a/internal/datacoord/services_test.go
+++ b/internal/datacoord/services_test.go
@@ -5979,3 +5979,67 @@ func TestAbortImport_CommittingRejected(t *testing.T) {
 	assert.NoError(t, err)
 	assert.False(t, merr.Ok(resp))
 }
+
+// TestHandleCommitVchannelRPC_V3SegmentIsNotFencedYet pins a pre-existing gap
+// that this PR does not close: the fence derives its bound from the segment's
+// binlog arrays, and a V3 (manifest-backed) segment never persists those --
+// buildAlterSegmentsKvs skips the per-FieldBinlog KVs for it and the SegmentInfo
+// is written without them -- so a V3 segment reloaded after a DataCoord restart
+// compares against 0 and admits any commit timestamp, however low.
+//
+// Import segments become V3 as soon as UpdateManifest runs, so this is the main
+// import path, not a corner. The follow-up that reads Stats.TimestampTo instead
+// should flip these assertions; until then the current behavior is recorded
+// rather than left to be discovered.
+func TestHandleCommitVchannelRPC_V3SegmentIsNotFencedYet(t *testing.T) {
+	ctx := context.Background()
+
+	importMetaMock := NewMockImportMeta(t)
+	importMetaMock.EXPECT().HandleCommitVchannel(mock.Anything, int64(3002), "vchan-0", mock.AnythingOfType("func() error")).
+		RunAndReturn(func(ctx context.Context, jobID int64, vchannel string, callback func() error) error {
+			return callback()
+		})
+
+	segIDs := []int64{11}
+	getSegIDsMock := mockey.Mock((*Server).getImportSegmentIDsByVchannel).
+		Return(segIDs).Build()
+	defer getSegIDsMock.UnPatch()
+
+	segments := NewSegmentsInfo()
+	// Exactly the shape a reloaded V3 import segment has: a manifest, no binlog
+	// arrays, and Stats carrying the row timestamps that did survive the restart.
+	segments.SetSegment(11, &SegmentInfo{SegmentInfo: &datapb.SegmentInfo{
+		ID:            11,
+		CollectionID:  100,
+		PartitionID:   10,
+		InsertChannel: "vchan-0",
+		State:         commonpb.SegmentState_Flushed,
+		IsImporting:   true,
+		ManifestPath:  "files/insert_log/100/10/11/manifest",
+		Stats:         &datapb.Statistics{TimestampTo: 500},
+	}})
+
+	server := &Server{
+		importMeta: importMetaMock,
+		meta: &meta{
+			catalog:  &datacoordkv.Catalog{MetaKv: NewMetaMemoryKV()},
+			segments: segments,
+		},
+	}
+	server.stateCode.Store(commonpb.StateCode_Healthy)
+
+	resp, err := server.HandleCommitVchannel(ctx, &datapb.HandleCommitVchannelRequest{
+		JobId:           3002,
+		Vchannel:        "vchan-0",
+		CommitTimestamp: 300, // below Stats.TimestampTo=500, yet admitted
+	})
+	assert.NoError(t, err)
+	assert.True(t, merr.Ok(resp), "the fence does not fire for a reloaded V3 segment")
+
+	seg := server.meta.GetSegment(ctx, 11)
+	require.NotNil(t, seg)
+	assert.EqualValues(t, 300, seg.GetCommitTimestamp())
+	assert.False(t, seg.GetIsImporting())
+	assert.EqualValues(t, 0, maxBinlogTimestampTo(seg.GetBinlogs()),
+		"the bound the fence compares against is 0 despite Stats.TimestampTo=500")
+}
--- a/internal/datacoord/services_test.go
+++ b/internal/datacoord/services_test.go
@@ -5827,6 +5827,61 @@ func TestHandleCommitVchannelRPC_StoresCommitTimestamp(t *testing.T) {
 	}
 }
 
+func TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp(t *testing.T) {
+	ctx := context.Background()
+
+	importMetaMock := NewMockImportMeta(t)
+	importMetaMock.EXPECT().HandleCommitVchannel(mock.Anything, int64(3001), "vchan-0", mock.AnythingOfType("func() error")).
+		RunAndReturn(func(ctx context.Context, jobID int64, vchannel string, callback func() error) error {
+			return callback()
+		})
+
+	segIDs := []int64{10}
+	getSegIDsMock := mockey.Mock((*Server).getImportSegmentIDsByVchannel).
+		Return(segIDs).Build()
+	defer getSegIDsMock.UnPatch()
+
+	segments := NewSegmentsInfo()
+	segments.SetSegment(10, &SegmentInfo{SegmentInfo: &datapb.SegmentInfo{
+		ID:            10,
+		CollectionID:  100,
+		PartitionID:   10,
+		InsertChannel: "vchan-0",
+		State:         commonpb.SegmentState_Flushed,
+		IsImporting:   true,
+		Binlogs: []*datapb.FieldBinlog{{
+			FieldID: 100,
+			Binlogs: []*datapb.Binlog{{
+				LogID:       10,
+				TimestampTo: 500,
+			}},
+		}},
+	}})
+
+	server := &Server{
+		importMeta: importMetaMock,
+		meta: &meta{
+			catalog:  &datacoordkv.Catalog{MetaKv: NewMetaMemoryKV()},
+			segments: segments,
+		},
+	}
+	server.stateCode.Store(commonpb.StateCode_Healthy)
+
+	resp, err := server.HandleCommitVchannel(ctx, &datapb.HandleCommitVchannelRequest{
+		JobId:           3001,
+		Vchannel:        "vchan-0",
+		CommitTimestamp: 300,
+	})
+	assert.NoError(t, err)
+	assert.False(t, merr.Ok(resp))
+	assert.ErrorIs(t, merr.Error(resp), merr.ErrImportSysFailed)
+
+	seg := server.meta.GetSegment(ctx, 10)
+	require.NotNil(t, seg)
+	assert.EqualValues(t, 0, seg.GetCommitTimestamp())
+	assert.True(t, seg.GetIsImporting())
+}
+
 func TestHandleCommitVchannelRPC_MissingJobReturnsError(t *testing.T) {
 	ctx := context.Background()
 


czs007:
This review surfaced one critical error-handling defect that can permanently wedge the pchannel flusher, plus one lower-priority test-coverage gap around the same code path.

### 🔴 Critical

**`internal/datacoord/meta.go:1752` — Commit-timestamp rejection can permanently wedge the pchannel flusher**
When the commit-timestamp invariant is violated (`commitTs < max(binlog.TimestampTo)`), `UpdateCommitTimestamp` calls `modPack.fail(WrapErrImportSysFailedMsg(...))`, aborting the whole batch before `AlterSegments`, so nothing is persisted (`is_importing` stays `true`) and the RPC returns `merr.Status(err)`. The sole consumer (`wal_flusher.go`) wraps the call in `retry.Do(..., retry.AttemptAlways())`, and `ErrImportSysFailed` (2101) is an un-wrapped `SystemError` — it matches neither the `IsRecoverable` nor the `InputError` gate, so the flusher retries the identical, deterministically-failing request forever, blocking WAL replay for the entire pchannel. There is no escape hatch: `AbortImport` rejects the `Committing` state, and `tryTimeoutJob` is inert by default (`GetTimeoutTs` defaults to `math.MaxUint64`). This replaces the pre-PR silent no-op (commit proceeds with `commitTs` unset) with a permanent stall.
*Suggestion:* fail the job terminally and ack the fence so `HandleCommitVchannel`'s `Failed` branch returns `nil` and the flusher unblocks, rather than retrying an unrecoverable error indefinitely.
*(raised by xiaocai)*

### 🟢 Low

**`internal/datacoord/services_test.go:5786` — New RPC test cannot catch the wedge (coverage gap, not a test defect)**
`TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp` stubs `importMeta.HandleCommitVchannel` with `RunAndReturn(func(...) error { return callback() })`, which elides the real transition that persists `State = Committing` before the callback runs. The test therefore verifies only that the services layer surfaces `ErrImportSysFailed` in the response `Status`; it exercises neither the pre-persisted `Committing` state nor the `AttemptAlways` retry loop, so it stays green whether or not the critical wedge is fixed. This follows the same mocking convention as its sibling tests (`services_test.go:5698`, `:5731`), so it is a legitimately-scoped RPC unit test — the gap is systemic, not introduced here.
*Suggestion:* add a separate integration-style regression test using a real `importMeta` that asserts post-rejection job state and flusher recoverability end-to-end.
*(raised by xiaocai)*

czs007:
Adversarial review verified 3 issues across critical, medium, and low severity, spanning a permanent commit-ack retry wedge, a missing zero-timestamp guard, and a mis-encoded copy-job timeout.

## 🔴 Critical

**`internal/streamingnode/server/flusher/flusherimpl/wal_flusher.go:347`** — Rejected `HandleCommitVchannel` wedges the entire pchannel

When `commitTs < max(binlog.TimestampTo)` (a real CDC scenario where the source cluster's TSO exceeds the target's), this PR turns a silent no-op into `modPack.fail(WrapErrImportSysFailedMsg(...))`. The batch aborts before `UpdateIsImporting(false)` runs and the job stays persisted in `Committing`. The StreamingNode flusher then retries the ack via `retry.Do(..., retry.AttemptAlways())` with a fixed `msg.TimeTick()`. Because `ErrImportSysFailed` carries the default `SystemError` type (not `InputError`), `retry.Do` never auto-aborts, so it spins forever on an immutable condition — and per the flusher's own comment this retry blocks the whole pchannel flusher, stalling WAL replay for every other collection on that pchannel. `tryTimeoutJob` does not rescue the default path, since `TimeoutTs` defaults to `math.MaxUint64` (~centuries out) absent an explicit `timeout` option.

*Suggestion:* On invariant violation, fail the import job through the `import_checker` `Failed` transition and return a successful/acknowledged status to the flusher so the pchannel drains, rather than rejecting the ack forever. *(raised by yezh, wzy)*

## 🟡 Medium

**`internal/datacoord/services.go:2996`** — `HandleCommitVchannel` accepts `CommitTimestamp == 0`, bypassing the invariant check

`commitTs := req.GetCommitTimestamp()` is passed unvalidated into `UpdateCommitTimestamp` + `UpdateIsImporting(false)`. `UpdateCommitTimestamp` guards the whole max-binlog check behind `if ts != 0`, so `ts == 0` skips validation, writes `segment.CommitTimestamp = 0`, and clears `is_importing` — exposing the import/CDC segment with no commit timestamp. Downstream, `commit_timestamp == 0` makes delete/MVCC handling fall back to `StartPosition.Timestamp`, exactly the temporal-correctness behavior this PR aims to prevent. The in-version flusher always sends a non-zero `msg.TimeTick()`, so this is defense-in-depth against a mixed-version StreamingNode or malformed internal caller rather than an in-version trigger.

*Suggestion:* Reject `commitTs == 0` in `HandleCommitVchannel` before building the update operators, leaving the `ts == 0` reset path exclusively for the compaction caller. *(raised by yezh)*

## 🟢 Low

**`internal/datacoord/snapshot_manager.go:1252`** — `CopySegmentJob` `TimeoutTs` written as UnixNano but read as a composed TSO

The sole production writer sets `TimeoutTs` to `uint64(time.Now().Add(jobTimeout).UnixNano())` (raw nanoseconds), but the checker reads it via `tsoutil.PhysicalTime(...)`, which decodes a composed TSO (`ts >> 18` as milliseconds). A 2026 UnixNano decodes to a physical time centuries in the future, so `time.Now().Before(timeoutTime)` is always true and the job never times out — making `CopySegmentJobTimeout` a no-op on the snapshot-initiated copy path. Contrast the import path, which stores a proper TSO. Out-of-diff and pre-existing, but adjacent to the same recovery/timeout machinery this PR touches.

*Suggestion:* Write `tsoutil.ComposeTSByTime(time.Now().Add(jobTimeout))` instead of `UnixNano()`. *(raised by wzy)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/2908c99a595ecb92cb584a50ea68e5cbb664f9ec...czs007:team2-fix/Tec19f94f6e?expand=1

czs007:
This review surfaced one critical correctness/availability regression that can wedge an entire pchannel flusher and must be addressed before merge.

## 🔴 Critical

**`internal/datacoord/meta.go:1752` — Rejecting `commit_ts < max(binlog.TimestampTo)` with a System-class error causes the pchannel flusher to retry forever** *(raised by wzy, yezh)*

**Problem:** This PR changes the `commit_ts < max(binlog.TimestampTo)` branch in `UpdateCommitTimestamp` from `return false` (a no-op that let `UpdateIsImporting` clear `is_importing`, ack the fence, and let the pchannel advance) to `return modPack.fail(merr.WrapErrImportSysFailedMsg(...))`. Because `fail` sets `updatePack.err`, the operator loop early-returns before `UpdateIsImporting`/`AlterSegments` (`meta.go:1822-1827`), so neither `commit_timestamp` nor `is_importing` is persisted (`is_importing` stays `true`). `HandleCommitVchannel` returns this error to the flusher (`services.go:3070-3072`). `ErrImportSysFailed` (code 2101) is deliberately a `SystemError` — not tagged `InputError` (`errors.go:261-266`) — so it satisfies none of `retry.Do`'s abort conditions (`retry.go:86-93`). The flusher wraps this call in `retry.Do(..., retry.AttemptAlways())` and only calls `ObserveMessage` after success (`wal_flusher.go:347-367`). Since `commitTs` and the persisted `TimestampTo` are invariant across retries, the rejection is deterministic, not transient, so the flusher thread blocks forever (per its own comment at `wal_flusher.go:341-342`) — halting **all** subsequent WAL messages on that pchannel, including unrelated healthy DML. This escalates a single import job's bad data into a full write-path outage for the pchannel. The two new tests (`services_test.go`, `commit_timestamp_test.go`) freeze this stuck state as the expected behavior and add no coverage for the flusher escaping the retry or the job reaching a terminal state.

**Suggestion:** Route the rejection through a terminal path instead of an infinite retry. Either return an `InputError`-class error (or add a flusher-side allowlist that breaks out of `retry.Do`) and transition the offending import job to `Failed` so the fence can be acked and the pchannel can advance; or, on the datacoord side, detect the invariant violation and fail the whole import job while returning success for the fence. Update the new tests to assert the job reaches a terminal state and the flusher unblocks, rather than asserting the wedged state.

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/1be47b3a40206bac8694dd73b89b9b3f76261459...czs007:team2-fix/T6e525db337?expand=1

czs007:
## Overview

Adversarial review surfaced one critical reliability bug where a rejected import commit timestamp returns a non-terminal system error, wedging the entire pchannel flusher in an infinite retry loop.

### 🔴 Critical

**`internal/datacoord/meta.go:1752`** — Rejected commit timestamp wedges the whole pchannel flusher
- **Problem:** When `commitTs < max binlog TimestampTo`, this path calls `modPack.fail(merr.WrapErrImportSysFailedMsg(...))`. `ErrImportSysFailed` (2101) is a `SystemError`, not `InputError`, so `retry.Do` treats it as recoverable. The flusher wraps `HandleCommitVchannel` in `retry.Do(..., retry.AttemptAlways())` with no `RetryErr` predicate, and for a fixed `msg.TimeTick()` the fence is deterministically rejected on every attempt → infinite loop that blocks WAL replay for all collections on that pchannel. Because the failing `UpdateSegmentsInfo` returns before `AlterSegments`/`SetSegment`, `is_importing` stays `true` and `commit_timestamp` is never persisted, leaving a stuck invisible segment. The job cannot recover: it stays in `Committing` (which `AbortImport` rejects), and `tryTimeoutJob` never fires because the default `TimeoutTs` is `math.MaxUint64`.
- **Suggestion:** Make this path terminal — classify the error as `InputError` (via `WrapErrImportFailedMsg`, sentinel 2100) so `retry.Do` aborts, and add a flusher/handler branch that fails the import job and acks the fence so the flusher advances while keeping the segment invisible.
- _(raised by wzy, yezh)_

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/1be47b3a40206bac8694dd73b89b9b3f76261459...czs007:team2-fix/T6561cf9cd5?expand=1

bigsheeper:
Thanks — the mechanism in the Critical finding is accurate, and I verified every step:

- `ErrImportSysFailed` (2101) is a SystemError (`errors.go:266`, no `WithErrorType(InputError)`).
- `retry.Do` aborts by default only on `InputError` (`retry.go:86`).
- The flusher wraps the call in `retry.Do(..., retry.AttemptAlways())` with no `RetryErr` (`wal_flusher.go:347`).
- `msg.TimeTick()` is fixed for a given message, so the fence is rejected identically on every attempt, and the retry blocks the whole pchannel — the call site says so itself.
- `AbortImport` rejects `Committing` (`services.go:3024`), so the job cannot be aborted by the user.

The description is right. The blocking is nevertheless intentional and I'd like to keep it.

**1. The suggested reclassification doesn't fit this error.** `errors.go:255-266` reserves `ErrImportFailed` (2100, InputError) for the import *data* layer — malformed user files — and states that server-side orchestration failures stay `ErrImportSysFailed` (2101) and "must not be bucketed as a user-caused failure". This rejection is a cross-cluster/TSO condition in the 2PC commit path, not a property of the caller's files.

**2. More importantly, failing the job terminally is unsafe under replication.** The predicate compares a replicated, identical `max(binlog.TimestampTo)` against each cluster's *own local* fence timetick. Those differ, and the replica evaluates strictly later, so the source can reject a fence the replica accepts. If we then fail the job on the source and ack the fence, the replica holds committed data the source does not — and it cannot be undone: `rollbackImportV2AckCallback` no-ops once the peer job is `Committing`/`Completed` (`ddl_callbacks_import.go:299-310`), and the `import_checker.go` rollback path is explicitly designed to release an *Uncommitted* peer only. A wedged pchannel is recoverable; that divergence is not.

The violation is also expected to be unreachable for a normal import: rows are stamped with the Import message's timetick and the fence is a later timetick on the same WAL. Only a binlog/backup import, which keeps the source cluster's row timestamps, can exceed the local fence. Hitting it means an assumption broke, so fail-stop is the behaviour I want; detection is via WAL/replication lag.

I've added a comment at the rejection site so it doesn't read as an accidental retry loop, and rewritten the PR description.

Two caveats I'd welcome pushback on: I have no end-to-end reproduction, and the flusher's retry log is at Debug (`wal_flusher.go:355`), so the lag metric is the detection signal rather than the log — raising that line may be worth a follow-up.

czs007:
## Overview

Adversarial verification surfaced six issues in this change — two high-severity concerns around the new fail-stop's convergence and blast radius, two medium-severity gaps in the reachability argument and the zero-timestamp path, and two low-severity items covering log volume and test depth.

---

### High

**`internal/datacoord/meta.go:1758` — Fail-stop makes the commit fence non-convergent and closes every recovery path**

The new `modPack.fail(...)` converts the rejection into an error that the flusher retries forever (`retry.AttemptAlways`), but both operands of the predicate are immutable across retries: `maxTsTo` derives from already-persisted segment binlogs, and `commitTs` is `msg.TimeTick()` of the same WAL message, replayed unchanged. The retry can therefore never succeed on its own. Meanwhile `importMeta.HandleCommitVchannel` persists the job as `Committing` *before* invoking the callback (`import_meta.go:369-370`, callback at `:393`), and every escape hatch is closed: `AbortImport` explicitly rejects `Committing` as terminal (`services.go:3024`), `checkGC` only handles `Completed`/`Failed` (`import_checker.go:608-611`), and `tryTimeoutJob` compares against `TimeoutTs`, which defaults to `math.MaxUint64` (≈ year 4200) when the user passes no `timeout` option (`importutilv2/option.go:79`). The PR body's justification that "a wedged pchannel is recoverable" has no in-tree recovery path under the default configuration — the only escape is dropping the collection.

*Suggestion:* validate the invariant *before* the job is flipped to `Committing` (i.e. at the CommitImport broadcast / fence-emit site), or permit `AbortImport` on a `Committing` job whose vchannel ack was rejected. (surfaced during verification)

**`internal/streamingnode/server/flusher/flusherimpl/wal_flusher.go:257` — The block stalls the entire pchannel dispatch loop, not just the importing collection**

`dispatchCommitImport` is called inline from `WALFlusherImpl.dispatch` (`:234`), the single per-pchannel message loop, so the `retry.AttemptAlways` loop holds up WAL consumption for *all* vchannels multiplexed on that pchannel. The PR body describes the cost as "blocks that pchannel's WAL replay", which understates it: one bad import segment in one collection halts flush/sync for every unrelated collection on the same pchannel — growing segments never sync and WAL retention grows unbounded. Combined with the non-convergence above, a single invariant violation becomes a cluster-wide, operator-only-recoverable stall.

*Suggestion:* state this trade-off explicitly in the PR description at minimum; preferably make the fence rejection per-vchannel non-blocking (park the job and ack the fence as failed) rather than parking the shared loop. (surfaced during verification)

---

### Medium

**`internal/datanode/importv2/util.go:214` — The one acknowledged reachable path (binlog/backup import) is not gated by `enableInReplicatingCluster`**

The PR body argues the violation is unreachable for a normal import and confines residual risk to replication, which is gated by `dataCoord.import.enableInReplicatingCluster` (default false). But binlog/backup import — an ordinary, ungated, user-facing restore — deliberately preserves the *source cluster's* row timestamps: `if _, ok := data.Data[common.TimeStampField]; !ok` stamps `req.GetTs()` only when the reader did not supply timestamps, as the comment at `:211` states, and the copy-segment path preserves source `TimestampTo` verbatim (`copy_segment_utils.go:557`). All import jobs, auto-commit included, traverse the same `Uncommitted → CommitImport` fence (`import_checker.go:497-503`), so a restore whose source cluster's wall clock ran ahead of the target's produces `max(binlog.TimestampTo) > commitTs` and lands squarely in the two high-severity failure modes above.

*Suggestion:* amend the "Not verified" section — the un-gated backup-restore case, not the gated replication case, is what determines the failure-mode budget. (surfaced during verification)

**`internal/datacoord/meta.go:1738` — `commitTs == 0` is still silently accepted on the commit path**

The guard sits inside `if ts != 0`, documented as the compaction reset path, and `HandleCommitVchannel` forwards `req.GetCommitTimestamp()` to `UpdateCommitTimestamp` with no validation (`services.go:3054,3061`). A commit fence carrying 0 therefore proceeds to `segment.CommitTimestamp = ts` and then `UpdateIsImporting(false)`, publishing the segment with `commit_timestamp = 0` — precisely the state the PR's What section says the invariant exists to prevent. In production `msg.TimeTick()` is never 0, so this is a hardening gap rather than an active bug, but the RPC boundary trusts a value it never checks, and the existing sibling test `TestHandleCommitVchannelRPC` (`services_test.go:5718-5723`) already exercises that exact shape by omitting `CommitTimestamp`.

*Suggestion:* reject `commitTs == 0` in `HandleCommitVchannel`; the reset path reaches the operator directly from compaction rather than over RPC, so it is unaffected. (surfaced during verification)

---

### Low

**`internal/datacoord/meta.go:1748` — Pre-existing `mlog.Error` is now inside an infinite retry loop**

Previously the rejection fired once per commit attempt and the batch completed, so the Error line was emitted once. With the fail-stop, the flusher retries the same RPC forever with backoff capped at `maxSleepTime = 3s` (`retry/options.go:52-58`), so the same unrate-limited Error is logged roughly every 3 seconds — about 29k lines/day per wedged segment — plus a `retry func failed` Warn every 4th attempt (`retry/retry.go:51-58`). Because the condition is deterministic and permanent, this is pure log flood on exactly the path an operator will be reading to diagnose the stall.

*Suggestion:* downgrade to a rate-limited log, or emit once per `(segmentID, commitTs)` pair. (surfaced during verification)

**`internal/datacoord/services_test.go:5789` — New test uses a single segment, so it does not cover batch atomicity**

The regression being fixed is that `UpdateSegmentsInfo` kept running the remaining operators of the batch after a rejection. A real batch is `len(segIDs)*2` operators across many segments (`services.go:3057-3065`), and the interesting guarantee is that a *sibling* segment which passed the check is also not persisted when a later one is rejected (`meta.go:1830-1835`). `TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp` mocks `getImportSegmentIDsByVchannel` to return `[]int64{10}`, so it only proves the failing segment's own `UpdateIsImporting` did not run.

*Suggestion:* add a second, valid segment (e.g. `TimestampTo: 100` alongside the violating `TimestampTo: 500` with `CommitTimestamp: 300`) and assert the valid one also retains `commit_timestamp == 0` and `is_importing == true`, pinning the batch-abort semantics the PR introduces. (surfaced during verification)

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3e9f26b0e726f1dac80e561b9525c6dec4d79df4...czs007:team2-fix/Tafb1d197ef?expand=1

czs007:
## Overview

Adversarial review surfaced one critical recoverability gap in the new fail-stop commit-timestamp path, plus a supporting test-coverage gap and a log-flooding side effect; the core validation logic itself is sound.

---

### 🔴 Critical

**`internal/datacoord/import_meta.go:367` — a rejected commit timestamp durably wedges the job in `Committing` with no reachable recovery path**

**Problem.** `HandleCommitVchannel` persists the `Uncommitted → Committing` transition *before* invoking the callback (`import_meta.go:366-374`). When the new validation at `meta.go:1757` rejects the fence via `modPack.fail(merr.WrapErrImportSysFailedMsg(...))`, `UpdateSegmentsInfo` aborts the whole operator batch (`meta.go:1830-1835`), the callback returns an error (`import_meta.go:393`), and `committed_vchannels` is never appended — yet the job is already durably in `Committing`. In-memory/persistent atomicity is fine (`modPack.Get` clones, so the aborted batch leaves no partial state), but every bounded recovery path is closed:

1. `AbortImport` explicitly rejects `Committing` — `services.go:3023-3027`, `"terminal/committed state %s, abort not allowed"`.
2. The `RollbackImport` WAL ack callback is a no-op for `Committing` — `ddl_callbacks_import.go:304-310`.
3. `tryTimeoutJob` would rescue it (`import_checker.go:565-575`; once `Failed`, `HandleCommitVchannel` returns `nil` at `import_meta.go:351-352` and the fence clears), but `importutilv2.GetTimeoutTs` defaults `timeoutTs = math.MaxUint64` when the caller passes no timeout option (`option.go:78-89`, stored via `services.go:1972/2031`), so by default it never fires.
4. `checkCommittingJob` has only a success branch (`import_checker.go:522-541`) — no failure transition.
5. Retry is unbounded: `ErrImportSysFailed` is code 2101 `SystemError` and is not tagged `InputError` (`merr/errors.go:266`), while `retry.Do` only short-circuits on `InputError` (`retry/retry.go:86`) and the flusher uses `retry.AttemptAlways()` (`wal_flusher.go:347-363`, whose own comment notes it "blocks the whole pchannel flusher").

**Failure scenario.** A binlog/backup import preserves the source cluster's `TimestampTo` (`copy_segment_utils.go:557`); if a segment's `max(binlog.TimestampTo)` exceeds the local fence timetick, the first rejection makes the flusher retry the same fence forever, blocking WAL replay for the entire pchannel — impacting every collection sharing it. A restart simply replays the same WAL message into the same rejection. For a multi-vchannel job, vchannels that committed earlier already have `is_importing=false` and are externally visible (`import_meta.go:385-391`), so that partial data stays visible permanently while the job never reaches `Completed`. The only remaining remedies are dropping the collection (`import_checker.go:594-600`) or hand-editing etcd — so the PR description's "until an operator intervenes" does not hold: the wedge is observable (WAL lag) but not actionable.

**Suggestion.** Either (a) roll the job state back to `Uncommitted` when the callback fails, or (b) relax `AbortImport` (and the rollback callback) to accept this class of `Committing` job. Note that a follow-up commit `566a5c0be4` ("fix: allow aborting an import job whose commit fence was rejected") already implements option (b) — marking the error with `errCommitTimestampBelowBinlog`, recording `importJobReasonCommitFenceRejected`, and letting abort/rollback through — but it is not in this PR's head (`3e9f26b0e7`), so the defect stands as of this PR.

*(raised by @xiaocai2333, @chyezh)*

---

### 🟡 Medium

**`internal/datacoord/services_test.go:5783` — new tests cover only the segment-side post-state, not job recoverability after rejection**

**Problem.** The added coverage (`commit_timestamp_test.go:126`'s `assert.ErrorIs` switch and `TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp`) asserts the right segment post-conditions — `ErrImportSysFailed`, `commit_timestamp == 0`, `is_importing == true` — but nothing asserts the *job* state after a rejection. The new test stubs `HandleCommitVchannel` through `MockImportMeta` down to `return callback()` (`services_test.go:5786-5790`), so the real `Uncommitted → Committing` persistence and `committed_vchannels` logic never execute. The "job stuck in `Committing` and un-abortable" behaviour is therefore structurally undetectable by this case.

**Suggestion.** Add a case asserting that after a rejection the job remains recoverable (abortable, or returned to `Uncommitted`) — the natural companion to the fix above. Follow-up commit `566a5c0be4` adds exactly this in `services_test.go` and `ddl_callbacks_import_test.go` (+76 test lines), confirming the gap is real.

*(raised by @xiaocai2333)*

---

### 🔵 Low

**`internal/datacoord/meta.go:1751` — the rejection-path `mlog.Error` turns from one-shot into an unbounded flood (~1 line / 3s, forever)**

**Problem.** The rejection branch logs at `Error` level (`meta.go:1751-1755`). Previously the operator returned `false`, the batch continued, the RPC succeeded, and the flusher did not retry — one log line total. With the new error return, the flusher's `retry.Do(..., retry.AttemptAlways())` retries the same fence indefinitely at a 3s max backoff (`retry/options.go:53-68`), re-entering the operator each time and producing roughly 20 `Error`-level lines per minute per stuck vchannel, forever (plus one `Debug` per attempt on the flusher side). This is not an objection to fail-stop semantics — it's the operability cost of them: deployments that alert on `Error` will see continuous noise, and the message carries no information that varies between attempts.

**Suggestion.** Downgrade or de-duplicate this log — e.g. `Error` on the first occurrence and `Warn` thereafter, or log once per (job, vchannel) — or suppress repeated rejections of the same fence at the caller.

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3e9f26b0e726f1dac80e561b9525c6dec4d79df4...czs007:team2-fix/T486447230f?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3e9f26b0e726f1dac80e561b9525c6dec4d79df4...czs007:team2-fix/Tf1b6693050?expand=1

bigsheeper:
Thanks, and sorry for the delay — this sat for two weeks. I verified every step
of both rounds against the head you reviewed (`3e9f26b0e7`; the branch is now
rebased onto latest master as `ad43c567d0`, patch-id identical). The Critical's
mechanism is correct in full, and I am **not** changing the behaviour. What was
wrong was the PR description, which I have rewritten.

### Critical / High 1 — the wedge is real, and it is the intended failure mode

Every link holds; I traced each rather than taking the description:

- `HandleCommitVchannel` persists `Uncommitted → Committing` before the callback
  runs (`import_meta.go:365-374`, callback at `:393`).
- `UpdateSegmentsInfo` aborts the whole batch on `updatePack.err` (`meta.go:1831-1834`).
- `AbortImport` rejects `Committing` (`services.go:3028-3031`) — and this is
  unchanged on current master, so a rebase does not alter it.
- The `RollbackImport` ack callback no-ops on `Committing` (`ddl_callbacks_import.go:304-310`).
- `tryTimeoutJob` covers `Committing` but compares `TimeoutTs`, which
  `GetTimeoutTs` leaves at `math.MaxUint64` absent a `timeout` option (`option.go:79`).
- `checkCommittingJob` has only a success branch (`import_checker.go:522-541`);
  `checkGC` handles `Completed`/`Failed` only (`:609-611`).
- `ErrImportSysFailed` carries no `WithErrorType(InputError)` (`merr/errors.go:266`),
  so `GetErrorType` returns `SystemError`, `retry.Do` does not short-circuit
  (`retry/retry.go:86`), and `AttemptAlways` retries forever.

High 2 is also correct and understated in the original text: `dispatchCommitImport`
runs inline from `WALFlusherImpl.dispatch` (`wal_flusher.go:234`, sole call site
`:129`), the single per-pchannel message loop, so every vchannel on that pchannel
stalls — as the code's own comment at `:341-342` already says.

Where I disagree is the conclusion. This check is a last-resort guard against a
condition that should be unreachable; if it ever fires, an assumption has broken
and stopping is the point. A stalled pchannel is loud — WAL/replication lag plus
a per-attempt `Error` line — and an operator investigates and clears the job out
of band. Publishing a segment with `commit_timestamp = 0` is silent and
unrecoverable. I am keeping fail-stop and accepting the manual recovery.

What was not acceptable was the PR body claiming "a wedged pchannel is
recoverable". That is false, and you were right to attack it. The description now
has a **What "an operator intervenes" means here** section enumerating each closed
path above, stating that detection is via monitoring rather than job state, and
noting the one automatic escape that does exist: an import submitted with a
`timeout` option is failed by `tryTimeoutJob`, after which the fence clears
through the `Failed` branch (`import_meta.go:351-352`).

One note on the record: the review cites "follow-up commit `566a5c0be4`" twice as
already implementing option (b). That SHA does not resolve in `milvus-io/milvus`
or in `czs007/milvus` (GitHub returns 422 for both), and the three
`team2-fix/*` branches linked from this PR are now 404. Nothing has been fixed
anywhere; please do not treat that as corroboration.

### Medium — test asserts segment post-state, not job recoverability

Accurate as an observation, but there is no recoverable post-state to assert:
remaining un-abortable in `Committing` is the specified behaviour, not a defect,
and it is now documented as such. Not adding that case.

### Low — the rejection log becomes an unbounded flood

Confirmed: ~20 `Error` lines per minute per stuck vchannel, indefinitely. Keeping
it at `Error` and un-deduplicated on purpose — with no job-state signal and no
automatic recovery, this log plus WAL lag is the entire detection surface, and a
rate-limited or downgraded line is one an alert can miss. The noise is the alarm.

czs007:

Adversarial review surfaced one behavioral concern and one test-coverage gap in this PR, plus one real but pre-existing hazard tracked separately below.

## Medium

**`internal/datacoord/meta.go:1825` — the new deterministic failure branch lands after the job has already been persisted as `Committing`, turning what used to be a transient-only callback failure into a permanent trap.**
`importMeta.HandleCommitVchannel` persists the job state as `Committing` (`import_meta.go:367-375`) *before* running the callback, which is `UpdateSegmentsInfo(UpdateCommitTimestamp, UpdateIsImporting)` (`services.go:3059-3073`). Before this PR the only way that callback could fail was a catalog/etcd write error — transient, so retry converges. The new branch is input-determined: if the segment's binlog timestamp is above the fence, every retry returns the same `ErrImportSysFailed`, which is a `SystemError` (`merr/errors.go:266`), so the flusher's `retry.Do(..., retry.AttemptAlways())` retries forever and, per its own comment, blocks the whole pchannel flusher — not just this import job. Failure scenario: a binlog/backup import (`importutilv2.IsBackup`, which keeps the source cluster's original ts — `importv2/util.go:453-462`) restored into a target cluster whose wall clock lags the source, with no `timeout` option set → the commit is rejected on every retry, WAL replay for the entire pchannel stalls so every collection sharing it stops ingesting, and the job is stuck in `Committing` where `AbortImport` is explicitly refused (`services.go:3027-3031`) and `tryTimeoutJob` never fires because `TimeoutTs` defaults to `math.MaxUint64` (`importutilv2/option.go:78-79`), leaving job and task meta leaked and only recoverable out of band.
Suggestion: perform this validation (or a pre-check) *before* the job flips to `Committing`, so a rejection lands in a window where `AbortImport` is still available. The PR description enumerates and accepts these costs, but the blast radius — all collections on the pchannel, manual-only recovery — is worth confirming explicitly. *(surfaced during verification)*

## Low

**`internal/datacoord/services_test.go:53` — the new regression test only constructs a V2-shaped segment, so it structurally cannot cover the V3 path.**
`TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp` builds a segment with in-memory `Binlogs{TimestampTo: 500}` and an empty `ManifestPath`, so `isV3Segment` (`metastore/kv/datacoord/util.go:116-118`) reports false and only the "segment still in memory" branch is exercised. The test is correct for what it claims to assert, but a green run does not demonstrate the invariant holds for reloaded V3 segments (see the pre-existing hazard below).
Suggestion: if the value source is changed to `Stats.TimestampTo`, add a companion case with empty `Binlogs` and only `Stats` + `ManifestPath` set. Non-blocking. *(raised by tinswzy)*

- **`internal/datacoord/meta.go:1813` — `maxTsTo` is derived from `segment.GetBinlogs()`, which is always empty for V3 segments after a datacoord restart, so the guard is silently bypassed on the main import path.** `resetBinlogFields` clears `Binlogs` before the SegmentInfo KV write (`metastore/kv/datacoord/util.go:146-151`), and `buildAlterSegmentsKvs` skips per-field binlog KVs for V3 segments outright (`kv_catalog.go:357-359`), so `applyBinlogInfo` (`kv_catalog.go:231-234`) has nothing to rebuild from on reload → `maxTsTo = 0` → `ts < 0` is never true. Import segments become V3 as soon as `UpdateManifest` runs (`import_task_import.go:266-275`), while the `Stats` written in the same batch survive restart. Failure scenario: a V3 backup segment with row timestamps up to 5000 and a local fence `commitTs=3000`, with datacoord restarting between the binlog persist and the fence being processed (the flusher's `AttemptAlways` replays the same WAL message) → after reload the guard passes, the segment is published with `CommitTimestamp=3000` and `IsImporting` cleared, i.e. a commit_ts below the real row timestamps — the exact state the invariant exists to prevent, and one that the same fence correctly rejects when no restart intervenes. This is unrelated to this PR because the diff only replaces `return false` with `modPack.fail(...)` and leaves the `maxTsTo` derivation at `meta.go:1812-1818` untouched, while V3 segments never persisting their binlog arrays is long-standing storage design (see the "arrays were never persisted" comment at `util.go:109-115`). Suggested follow-up: read `segment.EnsureStats().GetTimestampTo()` (`segment_info.go:59-77`) instead, without falling back to `BuildStatsFromFieldBinlogs`, which is equally empty for reloaded V3 segments. Recommend merging that separate fix before this PR. *(raised by xaxys, tinswzy)*

bigsheeper:
Thanks — the Medium and the Low are both taken in `bcacc84e71` (the branch is also rebased onto latest master, so the earlier commit carries a new SHA with unchanged content). I verified every step against the head you reviewed rather than reasoning from the description, and I found one thing about the ordering between your Medium and your pre-existing hazard that I do not think either of us said out loud.

## Medium — rejection lands after the job is already `Committing`. Confirmed; fixed.

Your chain holds at every link:

- `import_meta.go:367-375` persists the job as `Committing` **before** running the callback, and the comment there says the ordering is deliberate.
- `services.go:3028-3031` treats `Committing` as terminal and refuses `AbortImport`.
- `option.go:78` defaults `TimeoutTs` to `math.MaxUint64`, so `tryTimeoutJob` never fires without an explicit timeout.
- The error is retriable **by design** — the comment this PR added says so — so an input-determined rejection there retries forever rather than converging.

So a rejection inside that callback is a permanent trap with out-of-band recovery only, exactly as you described.

The same check now runs read-only in `HandleCommitVchannel` before `s.importMeta.HandleCommitVchannel`, where the job is still `Uncommitted` and an operator can abort it. `maxBinlogTimestampTo` is extracted so the pre-check and `UpdateCommitTimestamp` cannot drift apart.

One thing I did **not** change, to be explicit: the rejection stays a retriable `SystemError`, so the flusher still blocks on the fence. That trade — a blocked pchannel surfacing as WAL lag, versus a replica committing what the source rejected — is the one the PR body already argues for, and your finding is about *where the rejection lands*, not about that trade. This fixes recoverability, not the blocking.

`TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp` now registers no `EXPECT` on the import-meta mock at all, so the test fails if the commit phase is entered. Reverting the pre-check makes it fail with `mock: I don't know what to return because the method call was unexpected`.

## Low — the regression test could not cover the V3 path. Confirmed; added.

Right on the mechanism: the old test built in-memory `Binlogs` with an empty `ManifestPath`, so `isV3Segment` was false and only the in-memory branch ran.

`TestHandleCommitVchannelRPC_V3SegmentIsNotFencedYet` adds the other shape — `ManifestPath` set, no binlog arrays, `Stats{TimestampTo: 500}` — i.e. exactly what a reloaded V3 import segment looks like. It asserts the **current** behavior: `commitTs=300` is admitted despite `Stats.TimestampTo=500`, and the bound the fence compares against is 0.

## Pre-existing hazard — confirmed, not fixed here, and the ordering matters

I checked whether it is really pre-existing rather than taking your word: this PR's entire diff to `meta.go` was replacing `return false` with `modPack.fail(...)` plus a comment. The `maxTsTo` derivation from `GetBinlogs()` was already on master. So yes, pre-existing.

I also confirmed the mechanism end to end: `buildAlterSegmentsKvs` skips the per-FieldBinlog KVs for V3 (`kv_catalog.go:357`), `buildSegmentAndBinlogsKvs` persists the SegmentInfo without the arrays, and `applyBinlogInfo` has no KVs to rebuild from — so a reloaded V3 segment reports `maxTsTo = 0`. Your caveat about the fix is precise, too: `EnsureStats()` does fall back to `BuildStatsFromFieldBinlogs`, which is equally empty for a reloaded V3 segment, so the follow-up has to read `Stats.TimestampTo` without that fallback.

**The part I want to flag.** You recommend merging that fix *before* this PR. I would put it the other way round, or at least insist they land together. While the V3 gap stands, the fence barely fires on the main import path — import segments become V3 as soon as `UpdateManifest` runs — which is what keeps your Medium rare today. Close the V3 gap first and the fence starts firing on V3 segments, in the window where the job is already `Committing` and unabortable. That would turn the trap from a corner case into the common one. With this commit in, that ordering is safe either way.

Happy to open the V3 follow-up as its own issue naming both halves; say the word if you would rather it be filed against a different component.

## Verification

- `./internal/datacoord/` full package — pass (50.7s)
- Both behavioral tests verified to fail with the change reverted
- `make static-check` — 0 issues in core, pkg, client and go_client

czs007:

Adversarial review of this PR surfaced one high-severity issue: the new pre-commit fence validation does not actually deliver the abortability guarantee its comments and tests claim.

## High

- **`internal/datacoord/services.go:3063`** — `ValidateImportCommitTimestamp` is justified (in `services.go:3059-3062` and `meta.go:1856-1858`) as keeping a fence-rejected job `Uncommitted` and therefore still abortable, but the `Uncommitted → Committing` transition is not gated on the fence, so that guarantee does not hold. Two independent orderings break it: (1) `broadcastCommitImportMessage` builds the CommitImport message with `WithBroadcast(vchannels)` and no `OptBuildBroadcastAckSyncUp` (`services.go:2893-2900`), so the broadcaster fast-acks right after WAL append and `commitImportV2AckCallback` unconditionally persists the job as `Committing` (`ddl_callbacks_import.go:274-278`), independent of any flusher-side fence rejection; (2) with multiple vchannels, the first vchannel that passes validation flips the job to `Committing` inside `importMeta.HandleCommitVchannel` (`import_meta.go:367-374`) before a later vchannel's rejection arrives. Failure scenario: an import whose binlog rows carry `max(TimestampTo)` above the local commit fence timetick is committed and the WAL append succeeds → the FastAck callback persists `Committing` regardless of the fence, the flusher retries `HandleCommitVchannel` forever (`wal_flusher.go:347-363`, `retry.AttemptAlways()`), WAL replay on that pchannel stalls indefinitely, and `AbortImport` rejects `Committing` as terminal (`services.go:3028-3031`) — no API path to recover, contrary to what the new comments and test rationale (`services_test.go:5830-5834`) advertise. As written, `ValidateImportCommitTimestamp` is dead code for its stated purpose; the only effective rejection remains the in-operator `modPack.fail` (`meta.go:1905`), and the new test mocks `importMeta` with no job state, so neither ordering is exercised. Suggested fix: either make the code, comments, and tests state honestly that recovery after a fence rejection is out-of-band only (and drop or re-justify the pre-check), or provide a real recovery path — for example, allow abort/rollback of a `Committing` job whose commit fence was rejected before any vchannel was recorded committed. Note that simply gating `Committing` on validation across all vchannels conflicts with the FastAck broadcast design and would also require reworking the ack callback; either way, add a test covering broadcast → FastAck callback → fence rejection, and one where one vchannel commits before another rejects. *(raised by chyezh, sijie-ni-0214)*

bigsheeper:
You are right, and I am withdrawing the pre-check. Head is now `63bc5d997b`.

## Retracting what I claimed last round

I wrote that rejecting before `HandleCommitVchannel` "keeps the job `Uncommitted`, where an operator can still abort it". **That is false**, and I should not have asserted it. I verified the consumer chain thoroughly and never audited who *writes* the state — one `grep -rn "ImportJobState_Committing"` would have shown two writers:

- `ddl_callbacks_import.go:275` — `commitImportV2AckCallback`, which flips `Uncommitted → Committing` unconditionally
- `import_meta.go:369` — `HandleCommitVchannel`, the first vchannel

My pre-check gated neither. Worse than a race: `broadcastCommitImportMessage` builds CommitImport without `OptBuildBroadcastAckSyncUp`, so `FastAck` fires on WAL append completion (`broadcast_task.go:449-470`), while the flusher can only reach `HandleCommitVchannel` *after* consuming that same append. The ack path is structurally ahead, not merely likely to win. The pre-check's window opens only when the ack scheduler is backed up or DataCoord restarts between append and ack — and `callMessageAckCallbackUntilDone` closes even that.

So it was dead code for its stated purpose, and its comments and test rationale advertised a guarantee it could not make. Removed rather than re-justified, per your first suggested option.

**One note for the record, not a deflection:** the pre-check placement came from the previous round's suggestion — *"perform this validation (or a pre-check) before the job flips to `Committing`, so a rejection lands in a window where `AbortImport` is still available."* That suggestion does not hold for the same reason. The mistake that matters is mine: I promoted a review suggestion to a verified guarantee without auditing the write side.

## What the branch is now

Back to what the previous round judged `must-fix 0 / mergeable as-is`, plus honest wording and the V3 test:

- The fence still fails via `modPack.fail`. That change is the point of the PR: the old `return false` was silent, because `UpdateSegmentsInfo` discards an operator's boolean and checks only `updatePack.err` — so a violating input skipped setting `CommitTimestamp` while the sibling `UpdateIsImporting(false)` in the same call still applied, publishing the segment with `CommitTimestamp 0` and reporting success.
- The comment at the rejection now states plainly that recovery is out of band, and why: the job is already `Committing` via the FastAck path regardless of this fence, `AbortImport` treats that as terminal, and `tryTimeoutJob` cannot reach it. No implication that validating earlier would change it.
- `maxBinlogTimestampTo` is kept as a pure extraction — no behavior change — because the V3 test asserts against it.
- `TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp` is back to exercising the callback path.
- `TestHandleCommitVchannelRPC_V3SegmentIsNotFencedYet` (your Low) stays: `ManifestPath` set, no binlog arrays, `Stats{TimestampTo: 500}`, and `commitTs=300` admitted. It pins current behavior so the V3 follow-up has assertions to flip.

I did not add the broadcast→FastAck and multi-vchannel tests you suggested. With the pre-check gone they would pin the trap rather than verify a fix, and the trade is already stated in the PR body and now in the code. Happy to add them if you would rather have it executable than written down.

## CI

`code-check` is failing on something outside this branch. Master HEAD `95e3770aff` (#52006) calls `getSegmentBinlogFields` at `meta_test.go:2362` and the function is not defined anywhere in the tree, so `internal/datacoord` does not compile. I confirmed it on a clean master checkout, not just here — any PR rebased onto current master fails code-check identically. Local runs used a temporary stub for that symbol, which is not part of the commit; with it, `./internal/datacoord/` passes apart from the one master test that needs the real function, and `make static-check` reports 0 issues across core, pkg, client and go_client.

czs007:

Verified:
- `meta.go:1895-1900` — `UpdateSegmentsInfo` checks `updatePack.err` after each operator and returns before `catalog.AlterSegments` (`meta.go:1932`) and the in-memory `SetSegment`, so a rejected commit timestamp aborts the whole batch with nothing persisted and retries stay idempotent.
- `services.go:3058-3068` — `UpdateCommitTimestamp` and `UpdateIsImporting(false)` run in the same `UpdateSegmentsInfo` batch, so on rejection neither is applied and the original bug (publishing a segment with `commitTs=0`) is genuinely closed.
- `meta.go:1828` — a full-repo audit of `CommitTimestamp` write sites found no path that bypasses the fence: the remaining writers are compaction zero-resets (`ts=0` always passes) and snapshot restore (invariant preserved by pairwise copy), and the CDC copy-segment path never writes the field.
- `errors.go:266` and `utils.go:1485-1487` — `ErrImportSysFailed` (code 2101) carries no `InputError` marking and survives the RPC boundary intact, so the flusher's `retry.Do(..., AttemptAlways())` (`wal_flusher.go:347-363`) blocks fail-stop as intended, and marking the job Failed unblocks it out of band via `services.go:352-353`.
- `kv_catalog.go:354-359` — the V3 (manifest-backed) segment fence gap is pre-existing (the old `return false` branch was equally unreachable), is not regressed by this PR, and is explicitly pinned by `TestHandleCommitVchannelRPC_V3SegmentIsNotFencedYet` with a noted follow-up.
- `TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp` — the new RPC-level test asserts a non-ok response with the correct error code, `commitTs=0`, and `is_importing=true`, exactly the failure path this PR fixes, and the operator unit test was tightened from `assert.NoError` to `assert.ErrorIs`.

czs007 (approved):
/lgtm

Issue #51385, opened 2026-07-15
[Tracking]: Array element-match stack — correctness, performance, scalar MATCH and nested array indexes


xiaofan-luan:
Umbrella for a stacked series of PRs restructured out of #50669 (plus independent fixes found during its review). Merge order is top to bottom; each layer is independently correct and testable.

## Stack (merge in order)

| # | Issue | Scope | Size | Risk profile |
|---|---|---|---|---|
| 1 | #51379 | fix: bitset read() range check capped multi-bit reads at 8 bits | ~60 | trivial |
| 2 | #51380 | enhance: element-match fold/lookup performance (reserve, batched range lookups, word-wise fold, run-based element resolution) | ~1k | zero semantic change |
| 3 | #51381 | fix: three-valued NULL semantics (NULL row vs empty array) + schema-evolution races in struct-array element match | ~1.3k | user-visible behavior change, release note |
| 4 | #51382 | enhance: lock-free readers for ArrayOffsetsGrowing (chunked storage + atomic watermark) | ~0.9k | concurrency; memory-ordering proofs in code comments |
| 5 | #46517 | feat: quantified element-level MATCH_* / element_filter over scalar arrays, brute-force execution (reworked #50669) | ~6.5k | feature core; wire-compat via dual-written MatchColumnInfo |
| 6 | #46517 | feat: nested element-level indexes for plain scalar arrays + per-segment is_nested_index marker + index-served array equality | ~8k | rolling-upgrade surface; see design doc upgrade/rollback matrix |
| 7 | #51378 | enhance: serve array_length predicates purely from array offsets | ~0.6k | independent optimization |

## Independent (mergeable anytime)

- #51383 — fix: binlog import does not validate struct-array sub-field consistency
- #51384 — fix: Driver destroys segcore error codes (ExecOperatorException) — PR to follow

### Offset-path / NULL-3VL fixes found during this review

Independent expression-offset (iterative-filter) correctness fixes surfaced while reviewing the element-match stack. Each is its own PR, independently reviewable and tested (index-family tests are CI-verified — the scalar-sort-index / conjunct harnesses fail locally on macOS/arm64):

- [ ] #51540 — null-batch callback contract on skip-index `ProcessDataByOffsets` paths + timestamptz NULL 3VL
- [ ] #51591 — multi-query iterative-filter top-k insertion (`insert_helper` absolute-window fix)
[...]

PR #51591, opened 2026-07-18
fix: correct multi-query top-k insertion in iterative filter

xiaofan-luan:
`insert_helper` in `PhyIterativeFilterNode` indexed the per-query result window by the **relative** count instead of the **absolute** window end (`base_idx + count`). For `nq_index >= 1` the shift guard `count > pos` is always false, so an out-of-order candidate (a closer distance arriving after farther ones — which approximate vector iterators do emit) overwrote an existing top-k entry instead of shifting it, corrupting the top-k of **every query vector after the first** in a multi-query filtered search.

Extracted the insertion into `Utils.h::topk_binsert` with the correct absolute-window math, matching the already-correct math in `PhyIterativeElementFilterNode::CollectResults`.

Deterministic regression test `IterativeFilterInsert.MultiQueryWindowOrdering` inserts an out-of-order stream into the second query window and asserts the correct sorted result.

issue: #51385

### DIFF
--- a/internal/core/src/exec/operator/IterativeFilterNode.cpp
+++ b/internal/core/src/exec/operator/IterativeFilterNode.cpp
@@ -88,47 +88,6 @@ PhyIterativeFilterNode::IsFinished() {
     return is_finished_;
 }
 
-inline void
-insert_helper(milvus::SearchResult& search_result,
-              int& topk,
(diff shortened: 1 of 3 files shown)


liliu-z (commented):
/lgtm
/approve

PR #51592, opened 2026-07-18
fix: NULL 3VL on index-reverse-lookup and JSON-exists offset paths

xiaofan-luan:
Two sibling NULL-3VL fixes on the expression **offset (iterative-filter) path**, found by the same review that produced #51540 (which fixed the raw `ProcessDataByOffsets` skip branches). These are the sibling paths that PR missed.

## 1. `ProcessIndexLookupByOffsets` (scalar-index reverse-lookup)
- A null/absent row (`Reverse_Lookup == nullopt`) set `res=false` but left `valid_res=true`, so a wrapping `NOT` flipped it to a wrong match; it also never drove the callback, desyncing the cursor-tracking `processed_cursor` under an iterative-filter `AND`. Now `res=valid=false` + null-batch callback.
- The SkipIndex skip branch set `res=true`, but a pruned chunk cannot match → `res=false` with validity from the index + null-batch callback, matching every other skip branch.

## 2. `ExistsExpr`
A column-null row set `res=false` but left `valid_res=true`, so `NOT exists(json[path])` wrongly matched null-JSON rows → `res=valid=false`.

## Tests
Adds `not (...)` cases over a nullable scalar-sort-indexed field (`TestBinaryArithOpEvalRangeWithScalarSortIndexNullable`) and over `exists` on a nullable JSON column (`TestExistsWithJSONNullable`), asserting NULL rows are excluded under NOT on both the sequential and offset-input paths.

issue: #51385

### DIFF
--- a/internal/core/src/exec/expression/ExistsExpr.cpp
+++ b/internal/core/src/exec/expression/ExistsExpr.cpp
@@ -181,7 +181,10 @@ PhyExistsFilterExpr::EvalJsonExistsForDataSegment(EvalCtx& context) {
                 offset = (offsets) ? offsets[i] : i;
             }
             if (valid_data != nullptr && !valid_data[offset]) {
-                res[i] = false;
+                // Column-null row: three-valued logic requires marking it
+                // invalid too, else a wrapping NOT flips res=false to true and
+                // wrongly matches the null row.
+                res[i] = valid_res[i] = false;
                 continue;
             }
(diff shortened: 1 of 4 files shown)


xiaofan-luan:
Split for one-problem-per-PR: #51595 (ProcessIndexLookupByOffsets index reverse-lookup NULL 3VL) + #51596 (JSON exists column-NULL across raw/index/stats). Both tracked under #51385.

PR #51593, opened 2026-07-18
fix: NULL in group-by on index-only fields and BulkIsValid fall-through

xiaofan-luan:
Two independent NULL-robustness fixes found during the offset-path review.

## 1. `SealedDataGetter::Get` — group-by on an index-only nullable field
The indexed branch (`from_data_ == false`, a scalar field loaded index-only) did `AssertInfo(raw.has_value(), "field data not found")` and **threw on a NULL row**, failing the whole group-by query — while all four from-data getter branches correctly return `std::nullopt` for a null. Now returns `std::nullopt` so the NULL forms a distinct null group.

## 2. `ChunkedColumn` / `ProxyChunkColumn` `BulkIsValid`
The `if (!nullable_)` branch had no `return`, so a non-nullable column fell through into the nullable block and invoked the callback a second time per row (latent — currently masked by callers that guard with `if (!IsNullable()) return`). Added the missing `return`.

## Tests
`GroupBY.SealedNullableGroupByField` asserts a nullable group-by field does not crash and preserves null groups. The index-only `Reverse_Lookup==nullopt` branch (fix 1) is additionally exercised on CI; BulkIsValid (fix 2) is a latent-path guard.

issue: #51385

### DIFF
--- a/internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.h
+++ b/internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.h
@@ -282,7 +282,6 @@ class SealedDataGetter : public DataGetter<OutputType> {
                 return raw;
             }
         } else {
-            // null is not supported for indexed fields
             AssertInfo(index_ptr_.get() != nullptr,
                        "indexed field {} has no valid index pointer",
                        field_id_.get());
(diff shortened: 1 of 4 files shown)


xiaofan-luan:
Split for one-problem-per-PR: #51597 (NULL group-by on index-only nullable field) + #51598 (BulkIsValid non-nullable fall-through). Both tracked under #51385.

PR #51596, opened 2026-07-18
fix: honor column NULL for JSON exists across raw/index/stats paths

xiaofan-luan:
`exists(json[path])` on a column-null row left `valid_res=true` on all three paths (raw callback, JSON index, JSON stats), so `NOT exists()` wrongly matched null-JSON rows. Raw path → `res=valid=false`; index & stats cached paths apply the field null bitmap via `ApplyFieldValidDataByOffsets`. Adds `not(exists ...)` cases over a nullable JSON column.

issue: #51385

### DIFF
--- a/internal/core/src/exec/expression/ExistsExpr.cpp
+++ b/internal/core/src/exec/expression/ExistsExpr.cpp
@@ -16,6 +16,7 @@
 
 #include "ExistsExpr.h"
 
+#include <algorithm>
 #include <set>
 
 #include "bitset/bitset.h"
@@ -44,6 +45,82 @@
 namespace milvus {
 namespace exec {
 
+namespace {
+
+TargetBitmap
+GetJsonFieldValidity(const segcore::SegmentInternalInterface* segment,
+                     index::IndexBase* index,
+                     milvus::OpContext* op_ctx,
+                     FieldId field_id,
+                     int64_t active_count) {
+    if (auto index_valid = index->FieldIsNotNull(op_ctx);
+        index_valid.has_value()) {
+        AssertInfo(index_valid->size() == static_cast<size_t>(active_count),
+                   "JSON field validity size {} not equal to row count {}",
+                   index_valid->size(),
+                   active_count);
+        return std::move(*index_valid);
+    }
+
+    TargetBitmap valid(active_count, true);
+    if (!segment->is_nullable(field_id)) {
+        return valid;
+    }
+
+    // Backward compatibility for JSON indexes built before field-level
+    // validity was persisted.  New indexes are self-contained; old indexes
+    // can still execute while the raw column is loaded.
+    AssertInfo(segment->HasFieldData(field_id),
+               "nullable JSON index does not contain field validity and raw "
+               "field data is unavailable");
+    TargetBitmapView valid_view(valid);
+    int64_t processed = 0;
+    for (int64_t chunk_id = 0; processed < active_count; ++chunk_id) {
+        auto chunk_size = std::min(segment->chunk_size(field_id, chunk_id),
+                                   active_count - processed);
+        segment->ApplyFieldValidData(
+            op_ctx, field_id, chunk_id, 0, chunk_size, valid_view + processed);
+        processed += chunk_size;
+    }
+    return valid;
+}
+
+VectorPtr
+GatherJsonExistsByOffsets(const TargetBitmap& cached_res,
+                          const TargetBitmap& cached_valid_res,
+                          const OffsetVector& offsets,
+                          const TargetBitmap& bitmap_input) {
+    AssertInfo(bitmap_input.empty() || bitmap_input.size() == offsets.size(),
+               "bitmap input size {} not equal to offset count {}",
+               bitmap_input.size(),
+               offsets.size());
+
+    auto result =
+        std::make_shared<ColumnVector>(TargetBitmap(offsets.size(), false),
+                                       TargetBitmap(offsets.size(), true));
+    TargetBitmapView result_view(result->GetRawData(), offsets.size());
+    TargetBitmapView valid_view(result->GetValidRawData(), offsets.size());
+    for (size_t i = 0; i < offsets.size(); ++i) {
+        auto offset = offsets[i];
+        AssertInfo(
+            offset >= 0 && static_cast<size_t>(offset) < cached_res.size(),
+            "JSON exists offset {} out of range [0, {})",
+            offset,
+            cached_res.size());
+        valid_view[i] = cached_valid_res[offset];
+        if (!valid_view[i]) {
+            result_view[i] = false;
+            continue;
+        }
+        if (bitmap_input.empty() || bitmap_input[i]) {
+            result_view[i] = cached_res[offset];
+        }
+    }
+    return result;
+}
+
+}  // namespace
+
 void
 PhyExistsFilterExpr::DetermineExecPath() {
     if (CanUseJsonStatsAtInit()) {
@@ -70,8 +147,8 @@ PhyExistsFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
     switch (data_type) {
         case DataType::JSON: {
             span.GetSpan()->SetAttribute("json_filter_expr_type", "exists");
-            if (exec_path_ == ExprExecPath::ScalarIndex && !has_offset_input_) {
-                result = EvalJsonExistsForIndex();
+            if (exec_path_ == ExprExecPath::ScalarIndex) {
+                result = EvalJsonExistsForIndex(context);
             } else {
                 result = EvalJsonExistsForDataSegment(context);
             }
@@ -83,8 +160,10 @@ PhyExistsFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
 }
 
 VectorPtr
-PhyExistsFilterExpr::EvalJsonExistsForIndex() {
-    auto real_batch_size = GetNextBatchSize();
+PhyExistsFilterExpr::EvalJsonExistsForIndex(EvalCtx& context) {
+    auto* input = context.get_offset_input();
+    auto real_batch_size =
+        input != nullptr ? input->size() : GetNextBatchSize();
     if (real_batch_size == 0) {
         return nullptr;
     }
@@ -103,6 +182,7 @@ PhyExistsFilterExpr::EvalJsonExistsForIndex() {
                 auto* index = pinned_index_[cached_index_chunk_id_].get();
                 AssertInfo(index != nullptr,
                            "Cannot find json index with path: " + pointer);
+                auto* mutable_index = const_cast<index::IndexBase*>(index);
                 TargetBitmap res;
                 if (index->GetCastType().data_type() ==
                     JsonCastType::DataType::JSON) {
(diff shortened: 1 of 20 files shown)


czs007:

internal/core/src/exec/expression/ExistsExpr.cpp
```diff
@@ -44,6 +45,82 @@
...
+
+    TargetBitmap valid(active_count, true);
+    if (!segment->is_nullable(field_id)) {
+        return valid;
+    }
+
+    // Backward compatibility for JSON indexes built before field-level
+    // validity was persisted.  New indexes are self-contained; old indexes
+    // can still execute while the raw column is loaded.
+    AssertInfo(segment->HasFieldData(field_id),
```
liliu-z, line 73: Could we avoid failing here for legacy nullable JSON indexes? SORT/BITMAP/HYBRID normally skip loading raw field data because `HasRawData()` is true, so an existing v4 index without this sidecar can hit this assert after upgrade. Maybe force raw loading when the sidecar is missing, or provide an explicit rebuild path.

internal/core/src/index/JsonScalarIndexWrapper.h
```diff
@@ -172,6 +185,10 @@ class JsonScalarIndexWrapper : public BaseIndex {
                                non_exist_offsets_.data(),
                                non_exist_offsets_.size() * sizeof(size_t));
         }
+        auto field_null_data = EncodeFieldNullOffsets();
+        writer->WriteEntry(INDEX_FIELD_NULL_OFFSET_FILE_NAME,
```
liliu-z, line 189: This changes the JSON scalar-index capability, but the entry is written unconditionally while old and new nodes still report scalar engine v4. In a mixed cluster old QNs ignore it, and the coordinator cannot identify legacy v4 artifacts for rebuild. Should this be gated by a new scalar engine version?

pkg/common/common.go
```diff
@@ -190,8 +190,8 @@ const (
 )
 
 const (
-	// Version 3: metadata moved to separate meta.json file (instead of parquet metadata)
-	JSONStatsDataFormatVersion = 3
+	// Version 4: the shared column carries the original JSON column validity.
+	JSONStatsDataFormatVersion = 4
```
liliu-z, line 194: Bumping this format exposes a mixed-version task issue: DataCoord sends its local version, while a DataNode with a different version currently warns and returns nil, so the job is marked Finished with no artifact and the inspector retries. Can version mismatch be retriable or scheduled by capability, with a rolling-upgrade test?

zhengbuqian:
it seems to me the only problematic scenario is: 

the json column at this row is not set, thus `NULL`(not `{}`, not empty string `""`, not null literal `"null"`). then json stats evaluates `exists(json["a"])` as `result=false, valid=true`, thus breaking the 3 value logic.

this pr is somewhat correct in fixing this specific issue, but updates the json stats format thus requires a version update. but the json stats versioning system is broken, thus I think this fix is not optimal.

another way of doing so is: 

note that the only error case is when the entire json row is null, we can use the validity bitmap of this field to update the result setting the valid bit correctly, avoiding changing the stats format.

to avoid loading the json field just for this validity bitmap, we can store an extra in the `JsonKeyStats` object? that means 1 extra bitmap mem usage per json field, but may be acceptable until we fix the json stats versioning system.

zhengbuqian:
> it seems to me the only problematic scenario is:
> 
> the json column at this row is not set, thus `NULL`(not `{}`, not empty string `""`, not null literal `"null"`). then json stats evaluates `exists(json["a"])` as `result=false, valid=true`, thus breaking the 3 value logic.
> 
> this pr is somewhat correct in fixing this specific issue, but updates the json stats format thus requires a version update. but the json stats versioning system is broken, thus I think this fix is not optimal.
> 
> another way of doing so is:
> 
> note that the only error case is when the entire json row is null, we can use the validity bitmap of this field to update the result setting the valid bit correctly, avoiding changing the stats format.
> 
> to avoid loading the json field just for this validity bitmap, we can store an extra in the `JsonKeyStats` object? that means 1 extra bitmap mem usage per json field, but may be acceptable until we fix the json stats versioning system.

did some investigation, now it is very hard to only read the validity bitmap from storage, may not be an option

zhengbuqian:
/hold
looking closer

internal/core/src/index/Index.h
```diff
@@ -113,6 +114,19 @@ class IndexBase {
         ThrowInfo(NotImplemented, "Exists() not supported for this index type");
     }
 
+    // Returns the validity of the original field row, independently of any
+    // path extraction or cast performed by a JSON index.  JSON predicates need
+    // this distinction for SQL three-valued logic: a missing path is a valid
+    // FALSE result, while a NULL JSON column is UNKNOWN.
```
zhengbuqian, line 120: > `a missing path is a valid FALSE result` 

this statement is incorrect. in PG this is a null, not valid FALSE.

> The field/element/path extraction operators return NULL, rather than failing, if the JSON input does not have the right structure to match the request; for example if no such key or array element exists.

citing https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-OP-TABLE

zhengbuqian: but this is only a comment error.

internal/core/src/index/JsonScalarIndexWrapper.h
```diff
@@ -393,8 +470,164 @@ class JsonScalarIndexWrapper : public BaseIndex {
...
+        auto slice_meta = std::move(datas.at(INDEX_FILE_SLICE_META));
+        auto codecs = CompactIndexDatasByKey(
+            INDEX_FIELD_NULL_OFFSET_FILE_NAME, std::move(slice_meta), datas);
+        auto assembled = AssembleIndexDataCodec(codecs);
+        DecodeFieldNullOffsets(assembled->PayloadData(),
+                               assembled->PayloadSize());
+    }
+
     std::vector<size_t> non_exist_offsets_;
+    std::vector<size_t> field_null_offsets_;
```
zhengbuqian, line 627: `field_null_offsets_` is used only for serialization/deserialization, no need to store it as member.

internal/core/src/index/Index.h
```diff
@@ -113,6 +114,19 @@ class IndexBase {
...
+    //
+    // nullopt means that this index format does not carry field-level
+    // validity (for example, an index built by an older Milvus version).
+    virtual std::optional<TargetBitmap>
+    FieldIsNotNull(milvus::OpContext* op_ctx = nullptr) {
```
zhengbuqian, line 125: instead of adding a per path index null bitmap storage, maybe store it in field level, so in the case of multiple path indexes or multiple shredded keys, we don't store this duplicately multiple times.

internal/core/src/index/JsonHybridScalarIndex.h
```diff
@@ -198,13 +297,145 @@ class JsonHybridScalarIndex : public HybridScalarIndex<T> {
...
+        DecodeFieldNullOffsets(data->PayloadData(), data->PayloadSize());
+    }
+
     JsonCastType cast_type_;
     std::string nested_path_;
     JsonCastFunction cast_function_;
     proto::schema::FieldSchema json_schema_;
     storage::MemFileManagerImplPtr json_file_manager_;
     std::vector<size_t> non_exist_offsets_;
+    std::vector<size_t> field_null_offsets_;
```
zhengbuqian, line 435: same, no need to store this `field_null_offsets_` as a member.

internal/core/src/index/JsonScalarIndexWrapper.h
```diff
@@ -393,8 +470,164 @@ class JsonScalarIndexWrapper : public BaseIndex {
...
     TargetBitmap exists_bitset_;
+    TargetBitmap field_validity_bitset_;
```
zhengbuqian, line 629: I think we needed `field_validity_bitset_` only to distinguish this case:

| data | `exists(/a)` | `NOT exists(/a)` |
|---|---|---|
| SQL NULL | UNKNOWN | UNKNOWN, NOT Match |
| `{}` | FALSE | TRUE, Match |

If we accept `exists(/a)` to be `False` and `NOT exists(/a)` to be `True` when a row is `null` at this json field, we don't need `field_validity_bitset_` anymore? 

Actually in PG, `JSON null` and `SQL NULL` are distinguished, do we distinguish that as well? 



I think we can just treat them the same: treating our `null` as `JSON null` instead of `SQL NULL`?

Following PG's 3 value logic strictly related to JSON will have tons of gotchas and I don't think we want them all.

1 example: in PG comparing values with different types have this rule: Object > Array > Boolean > Number > String > JSON null, so we can get lots of unexpected results:

| Data | `doc['a'] >= '1'::jsonb` | `NOT (...)` |
|---|---:|---:|
| `{"a":1}` | `true` | `false` |
| `{"a":0}` | `false` | `true` |
| `{"a":"string"}` | `false` | `true` |
| `{"a":[1]}` | `true` | `false` |
| `{"a":"1"}` | `false` | `true` |
| `{"a":"0"}` | `false` | `true` |
| `{"a":true}` | `true` | `false` |
| `{"b":1}` | SQL `NULL` | SQL `NULL` |

I think in the case of Milvus, we'd better treat type mismatch as `UNKNOWN` instead of `TRUE/FALSE`.

xiaofan-luan: Your table is exactly the crux — and it's why I'd keep the field validity (3VL) rather than fold SQL NULL into the `{}` / JSON-null case. The deciding column is `NOT exists`:

- `{}` (doc present, path missing): `exists` = FALSE (valid) → `NOT exists` = TRUE → **matches**. ✅
- SQL NULL (whole column unset): `exists` = UNKNOWN → `NOT exists` = UNKNOWN → **does not match**.

If we treat SQL NULL as JSON null (2VL), the null row folds into the `{}` row and `NOT exists` starts matching it — i.e. `WHERE NOT exists(json['a'])` would return rows that have *no JSON value at all*. That's exactly the #51385 behavior this PR fixes, so dropping `field_validity_bitset_` would reintroduce it. It also matches PG's `?` operator (`'{}'::jsonb ? 'a'` → false, `NULL::jsonb ? 'a'` → NULL) and keeps JSON consistent with the array NULL-vs-empty 3VL in #51381.

I fully agree with your PR-level point that the weak spot here isn't the semantics but that this fix rides on the JSON-stats format version bump — that's the real thing to settle, addressing it separately.

internal/core/src/index/Index.h
xiaofan-luan: Will fix the wording. The `missing → NULL` rule you quoted is the *extraction* family (`->`, `->>`, `#>`); `exists` / `?` is a separate operator where a missing key is a definite `false`, not null (`'{}'::jsonb ? 'a'` → false). So “a missing path is a valid FALSE” holds for the exists operator — only the *column* being SQL NULL is UNKNOWN. I'll reword the comment to scope it to the exists operator instead of an unqualified “missing path”.

internal/core/src/index/Index.h
xiaofan-luan: Agree this is the right altitude — column-nullness is a field property, not a per-path-index one, so it shouldn't be duplicated across multiple path indexes / shredded keys. The catch (the same one you hit in your PR-level investigation) is persistence: reading just the field validity from storage without a format change is hard today. So rather than bolt another per-index sidecar on, I'd like to land the field-level validity together with the stats-versioning fix. Will follow up here with a concrete shape.

internal/core/src/index/JsonScalarIndexWrapper.h
xiaofan-luan: Agree — `field_null_offsets_` is only needed during serialize/deserialize, I'll keep it local instead of a member.

internal/core/src/index/JsonHybridScalarIndex.h
xiaofan-luan: Same as `JsonScalarIndexWrapper` — will keep `field_null_offsets_` local to (de)serialize rather than storing it as a member.

internal/core/src/index/JsonScalarIndexWrapper.h
zhengbuqian: but I think `WHERE NOT exists(json['a']) would return rows that have no JSON value at all` is somewhat acceptable.

and this is the semantic as used by clickhouse(verified through real clickhouse instance):

image

note that clickhouse explicitly does not allow `Nullable(JSON)`, so it does not have the `SQL NULL` case of PG, only `JSON null`.

PR #51597, opened 2026-07-18
fix: NULL group-by on an index-only nullable scalar field

xiaofan-luan:
`SealedDataGetter::Get`'s indexed branch (`from_data_==false`) threw `AssertInfo("field data not found")` on a null row instead of returning `std::nullopt` like the from-data branches, failing the whole group-by query. Now returns nullopt so the NULL forms a distinct null group. Adds a nullable group-by regression test (the index-only branch is CI-verified).

issue: #51385


liliu-z (commented):
/lgtm
/approve

PR #51598, opened 2026-07-18
fix: BulkIsValid non-nullable fall-through in ChunkedColumn/ProxyChunkColumn

xiaofan-luan:
The `if (!nullable_)` branch had no `return`, so a non-nullable column fell through into the nullable block and invoked the callback a second time per row (latent — currently masked by callers that guard with `if (!IsNullable()) return`). Added the missing `return`.

issue: #51385


liliu-z (commented):
/lgtm
/approve

Issue #51594, opened 2026-07-18
enhance: segcore per-operation published-state pin to keep in-flight sealed reads consistent across a concurrent Reopen (raw-data reclaim / column replace)


xiaofan-luan:
## Summary

A sealed segment publishes all query-visible resources as one immutable `PublishedSegmentState` (schema, `RuntimeResourceState` with the column map, indexes, skip-index metrics, readiness bitsets), swapped atomically (`std::atomic_store`) by load / `Reopen` / staged-commit publishers. Readers are lock-free: **each accessor independently `atomic_load`s the current state** and works on whatever snapshot is current *at that call* (design note at `ChunkedSegmentSealedImpl.h`: *"Serializes all published_state_ writers … Does NOT block readers — reader paths access published_state_ via atomic_load"*).

Because a single in-flight `Search`/`Retrieve` re-captures the published state **per accessor call** (not once per operation), a concurrent `Reopen` that removes or replaces a field's column mid-operation is observable to that operation's *next* accessor. If the field's raw column data was dropped, the next data read hits `AssertInfo(field_data_ready_bitset[field])` (or a null `get_column`) → **the query node crashes** instead of returning results.

This issue proposes a **segcore-level, per-operation published-state pin**: once a read starts, it holds one `shared_ptr<const PublishedSegmentState>` for the whole operation, and **every** read routes through it, so the operation observes exactly one consistent snapshot and the columns it reads stay alive until it finishes.

## Why the existing protections don't fully cover this

| Layer | Covers | Gap |
| --- | --- | --- |
| **#51396 DataViewGate** (RootCoord read-gate + complex-delete drain) | Schema **DROP FIELD**: gates the dropped field invisible to *new* reads, drains in-flight complex-deletes, WAL-orders before the DDL | Keyed on schema drop; does **not** cover raw-data reclaim (field stays in schema). Drains complex-deletes, not in-flight **searches** already executing in segcore. |
| **Proxy schema versioning / guarantee-ts** | Orders schema changes vs newly-admitted queries | An already-admitted, still-executing segcore read vs a concurrent in-place `Reopen` on the same segment. |
| **querynode Go (`LoadStateLock`)** | `Release` drains (`waitOrPanic(canReleaseAll)` → `refCnt==0`) | `Reopen` uses `PinIfNotReleased()` — does **not** wait for in-flight read `refCnt` to drain. |
[...]

PR #51602, opened 2026-07-19
enhance: pin and consolidate sealed-segment runtime snapshots

xiaofan-luan:
## What & why

Sealed-segment readers previously captured `published_state_` independently at multiple accessor calls. A concurrent `Reopen` could make one logical read derive routing/layout from snapshot A and fetch data from snapshot B.

The concrete #51594 failure is a no-raw-index -> raw-data-capable-index transition: query execution chooses the raw column from the old state, `Reopen` publishes the new index state and reclaims that column, then later Retrieve/Search materialization consults the new state and asserts.

This PR makes each sealed-segment Search/Retrieve operation observe one immutable published snapshot from query entry through reduce and output materialization.

## Design

- `ExecPlanNodeVisitor` pins before deriving `active_count`.
- `QueryContext` is the exec/query layer's single source of `OpContext`; operators call `query_context_->get_op_context()`, and `CompileExpression` resolves that same pointer into the physical expression tree's existing `op_ctx_`.
- Segcore accessors accept the resulting `OpContext*`; segcore does not depend on `QueryContext`.
- `RetrieveResult` and `SearchResult` retain the type-erased segment pin after `QueryContext` is destroyed.
- Reduce/export reconstruct one temporary context per result/segment, so parallel segments never share a mutable pin slot.
- Offset-driven Retrieve pins independently.
- Deletes remain live; sealed delete-mask row count advances only with the final runtime-state publication.

## Completeness: absorb #51531 into #51602

This PR supersedes #51531 and keeps its useful sealed runtime-state consolidation:

- row count, mmap-field markers, and variable-field average-size metadata live in `RuntimeResourceState`;
- fields, indexes, array offsets, timestamp state, and misc runtime metadata follow one clone/freeze/publish COW lifecycle;
- aborted Reopen staging cannot leak a future row count into `deleted_record_`;
- average-size metadata survives raw-column reclamation while the schema still retains the field.

It intentionally does not adopt a mutable `shared_ptr<SkipIndex>`. Skip-index ownership remains immutable and column/runtime-snapshot aligned.

## #51594 coverage

The routing and data reads in the reported chain now resolve the same pin:

`IndexHasRawData / HasFieldData -> bulk_subscript -> get_raw_data / get_vector / get_emb_list`

Regression coverage includes:

- `PinnedVectorRetrieveSurvivesRawIndexTransition`;
- `SearchResultPinSurvivesRawIndexTransitionDuringOutputFill`;
- `PinIdentityAcrossConcurrentPublish`;
- `OwnerGuardRejectsForeignPin`;
- `StagedRowCountDoesNotMutateDeleteStateBeforePublish`.

## Validation

- `cmake --build cmake_build --target milvus_core all_tests -j4`;
- 41/41 `SealedSegmentCowState.*`, `SealedSegmentOpSnapshotPin.*`, and `CommitTimestamp.*`;
- 121/121 external-take, Match expression, JSON flat-index, and element-filter tests;
- clang-format 15 idempotence and `git diff --check`;
- five unrelated broader-test failures were reproduced identically on committed baseline `4fa0143`.

issue: #51594

### DIFF
--- a/docs/design-docs/design_docs/20260717-sealed-segment-op-snapshot-pinning.md
+++ b/docs/design-docs/design_docs/20260717-sealed-segment-op-snapshot-pinning.md
@@ -54,11 +54,11 @@ Non-goals:
 
 ### Carrier: a type-erased slot on `milvus::OpContext`
 
-`OpContext` (milvus-common, `common/OpContext.h`) is already created one-per-operation
-per segment at the segcore C API boundary (`segment_c.cpp`: `AsyncSearch`,
-`AsyncRetrieve`, `SegmentLoad`, `AsyncReopenSegment`) and threaded through roughly half
-of the read accessors. It is the natural carrier for the pin. milvus-common must not
-know segcore types, so the slot is type-erased:
+`OpContext` (milvus-common, `common/OpContext.h`) is the existing operation-scoped
+carrier for cancellation, load priority, tracing, and storage accounting. Query
+execution creates its per-segment context in `ExecPlanNodeVisitor`; load/reopen paths
+create their own contexts at the segcore C API boundary. It is the natural carrier for
+the pin. milvus-common must not know segcore types, so the slot is type-erased:
 
 ```cpp
 // milvus-common: common/OpContext.h
@@ -94,9 +94,11 @@ The query visitor creates the per-segment `OpContext` and calls `PinOpSnapshot()
 before deriving `active_count` or constructing the plan fragment. This is after schema
 preparation at the C API / segment entry and before any layout-derived expression
 state is created. Once the `QueryContext` exists, it is the exec/query layer's single
-source of the operation context: operators and expressions obtain
-`query_context->get_op_context()` and pass only the resulting `OpContext*` across the
-lower-level segcore accessor boundary.
+source of the operation context: operators obtain
+`query_context->get_op_context()` at the point of use, while `CompileExpression`
+resolves the same pointer from `QueryContext` into the physical expression tree's
+pre-existing `op_ctx_` slot. Only the resulting `OpContext*` crosses the lower-level
+segcore accessor boundary; there is no second query-side propagation chain.
 Offset-driven `RetrieveByOffsets` is a separate sealed-segment read operation and pins
 at its own entry.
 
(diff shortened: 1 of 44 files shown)


xiaofan-luan:
Updated this PR on top of latest master and absorbed the useful runtime-state consolidation from #51531. The exec/query propagation now goes through QueryContext, Retrieve reuses one pin through filter/size/fill/ORDER BY, and the new PinnedVectorRetrieveSurvivesRawIndexTransition regression directly covers #51594. Local post-rebase build passed with 39 focused tests. @czs007 @godchen0212 please take another look when CI finishes.

czs007:
Adversarial review surfaced three low-severity, pre-existing snapshot/cache consistency gaps that are out of scope for this PR but worth tracking as follow-ups, as none introduce a regression or use-after-free.

## Low

**`internal/core/src/segcore/boost_score.cpp:142`** — `ComputeScorerScoresOnPreparedChunks` reads live segment state (`get_active_count`) and creates a bare `OpContext` that is set on the `QueryContext` without pinning a snapshot or inheriting the producing search's pin. Since the offsets originate from a prior sealed-segment search, a concurrent field-mutating `Reopen` between search and scoring can apply old-snapshot offsets to new-snapshot columns, producing wrong boost scores or a field-not-found/OOB assertion. This is pre-existing behavior not touched by this PR, and true cross-operation snapshot handoff is an explicit non-goal of the pinning design. Suggestion: track `ComputeScorerScoresOnOffsetChunks` as an uncovered sealed-segment read entry point in a dedicated follow-up rather than patching it with a self-pin at scoring entry, which would capture the new state and not fix the handoff. _(raised by chyezh)_

**`internal/core/src/segcore/SegmentInterface.cpp:152`** — `Search()` calls `check_search(plan)` before constructing the visitor that performs `PinOpSnapshot`, so admission checking and pinned execution can straddle a `Reopen`. The shared lock does not block `Reopen`, so the admission check may observe a different readiness view than execution; worst case is a spurious early `FieldNotLoaded`. This is not a use-after-free, since the subsequent pin keeps the old state alive. Suggestion: pin in `Search()` before `check_search` and thread `op_ctx` into `check_search` so admission and execution share one snapshot. _(raised by chyezh, tinswzy)_

**`internal/core/src/exec/expression/ExprCache.h:114`** — `ExprResCacheManager::Key` is `{segment_id, signature}` and a cache hit is validated only by `active_count`, relying on the invariant that the same `active_count` always produces the same bitmaps. An external-manifest-advance `Reopen` that changes row values while row count stays constant breaks this invariant and could reuse a stale bitset. This PR only plumbs `op_ctx` into schema reads and does not alter the cache key or validation, and the cache is cross-operation while the pin is per-operation, so pinning does not address it. Suggestion: add a published-state generation/version to the cache key in a dedicated issue. _(raised by chyezh, tinswzy)_

Issue #51606, opened 2026-07-19
[Bug]: DataCoord GC pause reports success after rolling back on caller timeout


bigsheeper:
### What happened?

`garbageCollector.pause` bounds the handoff to the meta GC worker by the caller's
context and rolls the pause ticket back when that context expires — but it then
falls through to `return nil`.

`Server.GcControl` maps a nil error to `Success`, so the caller is told that
garbage collection was paused when in fact the ticket had already been rolled
back and GC kept running. This affects the management path
`Proxy.PauseDatacoordGC` -> `MixCoord.GcControl` -> `Server.GcControl` ->
`garbageCollector.Pause`.

Reporting success for a safety interlock that did not take effect is the actual
defect: a caller that pauses GC in order to safely read segment files will
proceed on that false assurance.

### Where the wait actually is

An earlier revision of this issue described the risk as "the worker received the
signal but has not acknowledged it yet". That framing was wrong, and is corrected
here.

That post-send window is preemption-sized: every receive site closes `done` in
the same statement (`ackSignal`, and the recycle worker's own
`case cmd := <-signal`), so the acknowledgement is effectively immediate.

The reachable wait is the **send side**. `controlChannels["meta"]` is unbuffered,
and the meta worker only takes the signal at its top-of-loop select or at one of
the `ackSignal` polls scattered through a recycle round — so a slow object-storage
call inside that round delays the rendezvous. That path was already bounded by
`cmd.timeout`; what was wrong with it was the return value.

A second, narrower gap: the post-send ack wait had no timeout at all, so if it
ever did block, `Pause` returned success with the ticket still active.

### Expected behavior

When the caller's context expires, `Pause` must roll the pause ticket back **and
return the context error**, so `GcControl` surfaces a failed status instead of
`Success`. This should hold for both the send-side and the ack-side wait.

### Reproduction

Unit test in `internal/datacoord/garbage_collector_test.go`:

```bash
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 \
  ./internal/datacoord -run 'TestGarbageCollector$/TestPauseResume'
```

(`PKG_CONFIG_PATH` / `LD_LIBRARY_PATH` must point at your local
`internal/core/output/lib` build output.)

Before the fix, the timeout path returns `nil` and leaves the caller believing the
pause succeeded.

### Environment

Upstream master at time of report: `149903e0fd`.

zeweihan:
[...]

PR #51607, opened 2026-07-19
fix: honor gc pause timeout

bigsheeper:
issue: #51606

## What

`garbageCollector.pause` rolled the pause ticket back when the caller's context
expired, but then fell through to `return nil`. `GcControl` therefore reported
**Success** for a pause that had already been undone — the caller was told
garbage collection was paused while it was in fact running.

This returns `cmd.ctx.Err()` from the timeout path instead, and bounds the
post-send ack wait by the same context.

## Where the wait actually is

The original description (and #51606) framed the risk as "the worker received the
signal but has not acknowledged it yet". That window is in practice
preemption-sized: every receive site closes `done` in the same statement
(`ackSignal`, and the recycle worker's own `case cmd := <-signal`).

The reachable wait is the *send* side. `controlChannels["meta"]` is unbuffered,
and the meta worker only takes the signal at its top-of-loop select or at one of
the `ackSignal` polls scattered through a recycle round — so a slow object-storage
call inside that round delays the rendezvous. That path was already bounded by
`cmd.timeout`; what was wrong with it was the return value.

## Changes

- Carry the caller context in `gcCmd` so the timeout path can return its error.
- Return `cmd.ctx.Err()` from both timeout arms instead of `nil` (**the material
  fix**).
- Select on the caller timeout while waiting for the worker ack, so the ticket is
  rolled back there too.
- Bound the outer send by `gc.ctx.Done()`. Pre-existing hardening, raised in
  review: on `close()` the meta worker can return before receiving, and with a
  caller context that has no deadline the control-loop goroutine would park on the
  unbuffered send forever, hanging `gc.wg.Wait()`. Only the outer select needs
  this; the inner ack wait is immune for the reason above.

## Tests

- `go test -tags dynamic,test -gcflags="all=-N -l" -race -count=30 ./internal/datacoord -run 'TestGarbageCollector$/TestPauseResume'` → ok
- `go vet -tags dynamic,test ./internal/datacoord/`

The regression test no longer races a fixed 10 ms deadline against signal
delivery: it waits until the worker has taken the signal, then cancels, so the
inner timeout arm is exercised deterministically.

## Not verified locally

- `GcControlServiceSuite` (which holds the new status-projection test) needs a
  real etcd at `localhost:2379`; the whole suite panics in my checkout on an
  unmodified tree as well, so that test is covered by CI only.
- `make static-check` fails locally with `no go files to analyze` on an
  unmodified tree too — environment, not this change.

### DIFF
--- a/internal/datacoord/garbage_collector.go
+++ b/internal/datacoord/garbage_collector.go
@@ -89,6 +89,7 @@ type gcCmd struct {
 	collectionID int64
 	ticket       string
 	done         chan error
+	ctx          context.Context
 	timeout      <-chan struct{}
 }
 
@@ -294,6 +295,7 @@ func (gc *garbageCollector) Pause(ctx context.Context, collectionID int64, ticke
 		collectionID: collectionID,
 		ticket:       ticket,
 		done:         done,
+		ctx:          ctx,
 		timeout:      ctx.Done(),
 	}:
 		return <-done
@@ -433,10 +435,23 @@ func (gc *garbageCollector) pause(cmd gcCmd) error {
 	}
 	select {
 	case signalCh <- signal:
-		<-signal.done
+		select {
+		case <-signal.done:
+		case <-cmd.timeout:
+			gc.resume(cmd)
+			return cmd.ctx.Err()
+		}
 	case <-cmd.timeout:
 		// timeout, resume the pause
 		gc.resume(cmd)
+		return cmd.ctx.Err()
+	case <-gc.ctx.Done():
+		// The collector is closing. The meta worker may already have returned on
+		// its own ctx.Done(), and signalCh is unbuffered, so the send above would
+		// never be received: bound the wait by the collector's lifetime instead of
+		// parking this control-loop goroutine and hanging gc.wg.Wait() in close().
+		gc.resume(cmd)
+		return merr.WrapErrServiceUnavailable("garbage collector is closing")
 	}
 	return nil
 }
--- a/internal/datacoord/garbage_collector.go
+++ b/internal/datacoord/garbage_collector.go
@@ -94,6 +94,10 @@ type gcCmd struct {
 }
 
 type gcPauseRecord struct {
+	// id uniquely identifies this record within its gcPauseRecords. Tickets are
+	// not unique -- the REST route in restful_mgr_routes.go issues every pause
+	// with an empty ticket -- so rollback must delete by id, not by ticket.
+	id         int64
 	ticket     string
 	pauseUntil time.Time
 }
@@ -101,6 +105,7 @@ type gcPauseRecord struct {
 type gcPauseRecords struct {
 	mut     sync.RWMutex
 	maxLen  int
+	nextID  int64
 	records typeutil.Heap[gcPauseRecord]
 }
 
@@ -119,17 +124,15 @@ func (gc *gcPauseRecords) PauseUntil() time.Time {
 	return gc.records.Peek().pauseUntil
 }
 
-func (gc *gcPauseRecords) Insert(ticket string, pauseUntil time.Time) error {
+// Insert records a pause ticket and returns the id of the record it created, so
+// that a failed pause can roll back exactly its own record via DeleteByID.
+func (gc *gcPauseRecords) Insert(ticket string, pauseUntil time.Time) (int64, error) {
 	gc.mut.Lock()
 	defer gc.mut.Unlock()
 
 	// heap small enough, short path
 	if gc.records.Len() < gc.maxLen {
-		gc.records.Push(gcPauseRecord{
-			ticket:     ticket,
-			pauseUntil: pauseUntil,
-		})
-		return nil
+		return gc.pushLocked(ticket, pauseUntil), nil
 	}
 
 	records := make([]gcPauseRecord, 0, gc.records.Len())
@@ -145,25 +148,46 @@ func (gc *gcPauseRecords) Insert(ticket string, pauseUntil time.Time) error {
 	})
 
 	if gc.records.Len() < gc.maxLen {
-		gc.records.Push(gcPauseRecord{
-			ticket:     ticket,
-			pauseUntil: pauseUntil,
-		})
-		return nil
+		return gc.pushLocked(ticket, pauseUntil), nil
 	}
 
 	// too many pause records, refresh heap
-	return merr.WrapErrTooManyRequests(64, "too many pause records")
+	return 0, merr.WrapErrTooManyRequests(64, "too many pause records")
 }
 
+// pushLocked appends a new record with a freshly allocated id. Caller must hold mut.
+func (gc *gcPauseRecords) pushLocked(ticket string, pauseUntil time.Time) int64 {
+	gc.nextID++
+	gc.records.Push(gcPauseRecord{
+		id:         gc.nextID,
+		ticket:     ticket,
+		pauseUntil: pauseUntil,
+	})
+	return gc.nextID
+}
+
+// Delete drops every record holding the given ticket. This is the user-facing
+// resume semantic ("release my ticket"); use DeleteByID to undo a single record.
 func (gc *gcPauseRecords) Delete(ticket string) {
+	gc.deleteMatching(func(r gcPauseRecord) bool { return r.ticket == ticket })
+}
+
+// DeleteByID drops the single record with the given id, leaving records that
+// merely share its ticket untouched.
+func (gc *gcPauseRecords) DeleteByID(id int64) {
+	gc.deleteMatching(func(r gcPauseRecord) bool { return r.id == id })
+}
+
+// deleteMatching rebuilds the heap without the matching records, dropping
+// already-expired records along the way.
+func (gc *gcPauseRecords) deleteMatching(match func(gcPauseRecord) bool) {
 	gc.mut.Lock()
 	defer gc.mut.Unlock()
 	now := time.Now()
 	records := make([]gcPauseRecord, 0, gc.records.Len())
 	for gc.records.Len() > 0 {
 		record := gc.records.Pop()
-		if now.Before(record.pauseUntil) && record.ticket != ticket {
+		if now.Before(record.pauseUntil) && !match(record) {
 			records = append(records, record)
 		}
 	}
@@ -301,6 +325,11 @@ func (gc *garbageCollector) Pause(ctx context.Context, collectionID int64, ticke
 		return <-done
 	case <-ctx.Done():
 		return ctx.Err()
+	case <-gc.ctx.Done():
+		// cmdCh is unbuffered and startControlLoop has already returned, so the
+		// send above can never be received; without this arm the caller parks
+		// until its own ctx is canceled (forever for a Done()-less ctx).
+		return merr.WrapErrServiceUnavailable("garbage collector is closing")
 	}
 }
 
@@ -322,6 +351,9 @@ func (gc *garbageCollector) Resume(ctx context.Context, collectionID int64, tick
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
+	case <-gc.ctx.Done():
+		// see Pause: the control loop is gone, nothing will receive this send.
+		return merr.WrapErrServiceUnavailable("garbage collector is closing")
 	}
 }
 
@@ -411,8 +443,9 @@ func (gc *garbageCollector) pause(cmd gcCmd) error {
 		mlog.Duration("duration", cmd.duration),
 	)
 	var err error
+	var recordID int64
 	if cmd.collectionID <= 0 { // legacy pause all
-		err = gc.pauseUntil.Insert(cmd.ticket, reqPauseUntil)
+		recordID, err = gc.pauseUntil.Insert(cmd.ticket, reqPauseUntil)
 		log.Info(gc.ctx, "global pause ticket recorded")
 	} else {
 		curr, has := gc.pausedCollection.Get(cmd.collectionID)
@@ -420,7 +453,7 @@ func (gc *garbageCollector) pause(cmd gcCmd) error {
 			curr = NewGCPauseRecords()
 			gc.pausedCollection.Insert(cmd.collectionID, curr)
 		}
-		err = curr.Insert(cmd.ticket, reqPauseUntil)
+		recordID, err = curr.Insert(cmd.ticket, reqPauseUntil)
 		log.Info(gc.ctx, "collection new pause ticket recorded")
 	}
 	if err != nil {
@@ -438,22 +471,41 @@ func (gc *garbageCollector) pause(cmd gcCmd) error {
 		select {
 		case <-signal.done:
 		case <-cmd.timeout:
-			gc.resume(cmd)
+			gc.rollbackPause(cmd, recordID)
 			return cmd.ctx.Err()
 		}
 	case <-cmd.timeout:
 		// timeout, resume the pause
-		gc.resume(cmd)
+		gc.rollbackPause(cmd, recordID)
 		return cmd.ctx.Err()
 	case <-gc.ctx.Done():
 		// The collector is closing. The meta worker may already have returned on
 		// its own ctx.Done(), and signalCh is unbuffered, so the send above would
 		// never be received: bound the wait by the collector's lifetime instead of
 		// parking this control-loop goroutine and hanging gc.wg.Wait() in close().
-		gc.resume(cmd)
+		gc.rollbackPause(cmd, recordID)
 		return merr.WrapErrServiceUnavailable("garbage collector is closing")
 	}
 	return nil
+}
+
+// rollbackPause undoes exactly the record this pause() call inserted. It must not
+// go through resume(), whose delete is ticket-scoped: tickets are not unique --
+// every pause issued by the REST route in restful_mgr_routes.go carries an empty
+// ticket -- so a ticket-scoped delete would also drop a concurrent caller's
+// still-valid pause record and resume GC while that caller believes it is paused.
+func (gc *garbageCollector) rollbackPause(cmd gcCmd, recordID int64) {
+	if cmd.collectionID <= 0 {
+		gc.pauseUntil.DeleteByID(recordID)
+	} else if curr, has := gc.pausedCollection.Get(cmd.collectionID); has {
+		curr.DeleteByID(recordID)
+		if curr.Len() == 0 || time.Now().After(curr.PauseUntil()) {
+			gc.pausedCollection.Remove(cmd.collectionID)
+		}
+	}
+	mlog.Info(gc.ctx, "pause rolled back",
+		mlog.Int64("collectionID", cmd.collectionID),
+		mlog.String("ticket", cmd.ticket))
 }
 
 func (gc *garbageCollector) resume(cmd gcCmd) {
--- a/internal/datacoord/garbage_collector_test.go
+++ b/internal/datacoord/garbage_collector_test.go
@@ -2277,6 +2277,56 @@ func (s *GarbageCollectorSuite) TestPauseResume() {
 		s.Zero(gc.pauseUntil.PauseUntil())
 	})
 
+	s.Run("pause_timeout_after_signal_received", func() {
+		gc := newGarbageCollector(s.meta, newMockHandler(), GcOption{
+			cli:              s.cli,
+			enabled:          true,
+			checkInterval:    time.Hour,
+			scanInterval:     time.Hour * 7 * 24,
+			missingTolerance: time.Hour * 24,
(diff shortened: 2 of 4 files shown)


czs007:
Adversarial review surfaced no blocking defects, but a few robustness and test-quality improvements are worth addressing before merge.

## Medium

- **`internal/datacoord/garbage_collector.go:436` — Bound the pause path by collector lifetime.** Neither the outer nor the inner select in `pause()` waits on `gc.ctx.Done()`. The real exposure is the *outer* unbuffered send (`signalCh <- signal`): if `GcControl(Pause)` is called with no client deadline while the meta worker is mid-round, `close()`→`gc.cancel()` can let the worker exit before it receives, leaving the control-loop goroutine parked on the send forever so `gc.wg.Wait()` hangs shutdown. This is a pre-existing latent liveness gap that this PR touches rather than introduces (the new inner `<-signal.done` select is largely immune, since every receive site closes `done` in the same statement). Suggestion: as a defensive hardening, add `case <-gc.ctx.Done()` to both selects, rolling back and returning `gc.ctx.Err()` / service-unavailable so the wait is bounded by the collector's lifetime. *(raised by xiaocai)*

## Low

- **`internal/datacoord/garbage_collector.go:437` — Correct the PR-description framing of the fix.** The genuinely long wait in `pause()` is the send-side rendezvous, not the post-send/pre-ack window: all receive sites close `done` in the same statement, so the acknowledgement is preemption-sized and the added inner-select timeout guards a near-nonexistent window. The material behavior change this PR actually delivers is the corrected return value on timeout (`return nil` → `return cmd.ctx.Err()`). Suggestion: reword the PR body so the described risk is that the worker has not yet *received* the unbuffered send (a path already bounded by the outer `cmd.timeout`), rather than that it "received the signal but has not acknowledged it yet." *(raised by yezh, xiaocai)*

- **`internal/datacoord/services.go:1827` — Cover the nil→error behavior change in `GcControl`.** Before this PR, the timeout/rollback path returned `nil`, so `GcControl(Pause)` reported `Success`; it now returns `cmd.ctx.Err()`, which `GcControl` surfaces as `UnexpectedError` with a wrapped reason. The new `pause_timeout_after_signal_received` test asserts only `Pause()` directly and never exercises the `GcControl` status projection. Suggestion: add a test asserting the caller-visible `Success` → `UnexpectedError` flip (correct behavior, just currently unverified). *(raised by xiaocai)*

- **`internal/datacoord/garbage_collector_test.go:2320` — Remove the timing race in the rollback test.** The rollback assertion itself is sound (`collectionID=-1` is a global pause, so a correct rollback zeroes `pauseUntil`), but the fixed 10ms caller deadline races signal delivery: under loaded CI / `-race`, if the control loop doesn't rendezvous within 10ms, `pause()` takes the outer `cmd.timeout` arm, nothing is sent to the meta channel, the first `select { case signal = <-signalCh … case <-time.After(time.Second): Fatal(…) }` times out and Fatals, and the forwarding goroutine leaks. Suggestion: widen/remove the fixed deadline or drive the outer-vs-inner ordering deterministically. *(raised by yezh, xiaocai)*

bigsheeper:
Thanks — all four verified, three applied.

**Medium (`garbage_collector.go:436`)** — applied, outer select only. `controlChannels["meta"]` is unbuffered and the meta worker returns on its own `ctx.Done()`, so on `close()` a caller context without a deadline would park the control-loop goroutine on the send and hang `gc.wg.Wait()`. I skipped the inner select for the reason you gave yourself — every receive site closes `done` in the same statement, so `<-signal.done` cannot block meaningfully and the arm would be dead code. Returned `merr.WrapErrServiceUnavailable` rather than `gc.ctx.Err()`: a bare `context.Canceled` reads as "the caller cancelled" when it is actually the server shutting down, and `Resume()` already uses that sentinel.

**Low (`garbage_collector.go:437`, description framing)** — you're right, PR description rewritten. I traced it: the ack window is preemption-sized (`ackSignal` and the worker's `case cmd := <-signal` both close in the same statement), and the reachable wait is the send-side rendezvous, delayed by whatever the recycle round is doing between `ackSignal` polls. The material fix is `return nil` → `return cmd.ctx.Err()`. I'll update #51606 too.

**Low (`garbage_collector_test.go:2320`, timing race)** — applied. The fixed 10ms deadline could also make the test take the *outer* arm and Fatal on the undelivered signal, leaking the forwarding goroutine. Replaced with `WithCancel`: wait until the worker has taken the signal, then cancel, so the inner arm is exercised deterministically. Green under `-race -count=30`.

**Low (`services.go:1827`, untested status flip)** — added `TestPauseErrorSurfacesAsFailedStatus`. Note `TestTimeoutCtx` already covers a cancelled context, but it goes through `Pause()`'s *outer* select, which returned an error before this PR too — so it does not pin the flip. The new test mocks `(*garbageCollector).Pause` to return an error, which covers the projection deterministically.

Caveat: `GcControlServiceSuite` needs a real etcd, and the whole suite panics in my checkout on an unmodified tree, so that last test is CI-verified only — flag it if CI disagrees.

czs007:
## Overview

Adversarial review surfaced three issues in the GC pause/resume control path: two `medium` correctness gaps around shutdown-time blocking and error-status flattening, plus one `low` test-assertion weakness.

---

### 🟠 Medium

**1. `internal/datacoord/garbage_collector.go:292` — `Pause()`/`Resume()` send on `cmdCh` is not guarded by `gc.ctx`**

**Problem:** This PR adds `case <-gc.ctx.Done()` to the inner select that signals the meta worker (`garbage_collector.go:448`), but the public entry points are not symmetrically guarded. `Pause()` (`:291-305`) and `Resume()` (`:313-321`) select only on `case gc.cmdCh <- gcCmd{...}` and `case <-ctx.Done()`. `cmdCh` is unbuffered (`:240`), and its only receiver — `startControlLoop` — returns as soon as `gc.ctx` is cancelled (`:385-400`). After `close()` runs `gc.cancel(); gc.wg.Wait()` (`:513-521`), the control loop is gone, so any `Pause`/`Resume` entering afterwards blocks on the send until the *caller's* ctx is cancelled.

Scope, precisely: this does **not** hang `gc.wg.Wait()` — `Pause`/`Resume` are invoked from gRPC handler goroutines, which are not part of `gc.wg` (it only tracks the 4 workers added at `:332-367`). The blocked goroutine is the handler, so the symptom is a `GcControl` RPC that never returns. Nor is the caller ctx a nil-`Done()` channel on the production path: `services.go:1827` receives the gRPC handler ctx (`proxy/management.go:151` and `coordinator/restful_mgr_routes.go:154` pass `req.Context()`), which is closed on client disconnect or RPC completion. So the real-world impact is an in-flight `GcControl` hanging until client timeout/disconnect during shutdown — a permanent block only occurs with a `Done()`-less ctx such as `context.Background()`, which is exactly what the datacoord test passes (`services_test.go:1969`).

**Suggestion:** Mirror the inner fix at the outer selects — add `case <-gc.ctx.Done(): return merr.WrapErrServiceUnavailable("garbage collector is closing")` to both `Pause` and `Resume`. `merr` is already used at `:310`, so no new import is needed.

*(raised by chyezh, tinswzy)*

**2. `internal/datacoord/services.go:1828` — `GcControl` flattens `Pause`/`Resume` errors into `UnexpectedError`, dropping `Code` and `Retriable`**

**Problem:** `services.go:1827-1831` hand-builds the status via `status.ErrorCode = commonpb.ErrorCode_UnexpectedError; status.Reason = fmt.Sprintf(...)`, leaving `Code` at `0` and `Retriable` at `false`. This PR changes `pause()`'s timeout/shutdown paths from `return nil` to `return cmd.ctx.Err()` (`:442`, `:447`) and `merr.WrapErrServiceUnavailable(...)` (`:454`) — precisely the errors clients most need to distinguish. `merr.Code` maps `context.Canceled → CanceledCode` and `context.DeadlineExceeded → TimeoutCode` (`pkg/util/merr/utils.go:50-56`), and `ServiceUnavailable` is retriable. Through the flattening path the client sees only `ErrorCode=UnexpectedError, Code=0, Retriable=false`, unable to tell timeout from cancellation from unavailability, and with no retry semantics.

To be accurate about attribution: the flattening is **pre-existing code** (not in this PR's diff) — `pause()` could already return an error from `gc.pauseUntil.Insert` (`:426-428`). What this PR changes is that these three error classes now become routine return values, which is what brings the latent defect to the surface.

**Suggestion:** Replace both sites (`Pause` at `:1827-1831`, `Resume` at `:1838-1842`) with `return merr.Status(err), nil`, which populates `Code`/`ErrorCode`/`Retriable`/`Detail` (`pkg/util/merr/utils.go:106-126`). The same function already uses this correct form at `:1823` and `:1835`.

*(raised by chyezh, tinswzy)*

---

### 🟡 Low

**3. `internal/datacoord/services_test.go:1976` — `TestPauseErrorSurfacesAsFailedStatus` asserts only `!merr.Ok`, so it cannot catch code flattening**

**Problem:** The test mocks `(*garbageCollector).Pause` to return `context.DeadlineExceeded` but asserts only `s.False(merr.Ok(resp))`. `merr.Ok` is `status.GetErrorCode() == Success && status.GetCode() == 0` (`pkg/util/merr/utils.go:281-283`), and `services.go:1828` already sets `ErrorCode` to `UnexpectedError` — that single field alone satisfies the assertion, leaving `Code == 0` and `Retriable == false` completely unguarded. The test therefore proves only "it didn't report success", and provides no regression coverage for the code/retriable flattening described above.

**Suggestion:** Add `s.Equal(int32(merr.TimeoutCode), resp.GetCode())` (alongside the `merr.Status` change in `services.go`), or at minimum `s.ErrorIs(merr.Error(resp), context.DeadlineExceeded)`. Worth noting that the PR description states this suite requires a live etcd on `localhost:2379` and was not run locally — with CI as the only coverage, a weakened assertion costs more than usual.

*(raised by chyezh, tinswzy)*

czs007:
Adversarial review of this PR surfaced four issues: one correctness risk around shared-ticket rollback, plus three lower-severity concerns on error typing, invariant fragility, and test coverage.

## Medium

**`internal/datacoord/garbage_collector.go:441` — shared-ticket rollback can delete another caller's pause record**

The rollback paths (`:441` inner ack-wait branch, `:446` outer send-timeout branch, `:453` close branch) all call `gc.resume(cmd)`, which goes through `gcPauseRecords.Delete(cmd.ticket)` (`:159-173`). The retention predicate there is `now.Before(record.pauseUntil) && record.ticket != ticket`, so it drops **every** record with a matching ticket, not just the one this call inserted.

Empty tickets are a real entry point, not a hypothetical: `ControlDataCoordGC` in `internal/coordinator/restful_mgr_routes.go:154` builds params containing only `"duration"` — no `"ticket"`, no `"collection_id"`. In `services.go:1821`, `common.GetInt64Value` returns `(0, nil, false)` for a missing key (`pkg/common/common.go:902-904`), so `collectionID=0` takes the `cmd.collectionID <= 0` global branch in `pause()`, and `GetStringValue` at `services.go:1825` returns `""` (`common.go:917-919`). Every pause from this REST route is therefore a global pause with `ticket=""`. (The proxy path at `internal/proxy/management.go:139` generates a UUID ticket and is unaffected.)

`startControlLoop` (`:387-394`) processes `cmdCh` serially, so two `""` records can't be inside `pause()` at once — but the failing sequence doesn't require that: pause A completes successfully and returns nil (its `""` record stays in the heap; the caller believes GC is paused until A's deadline) → pause B inserts a second `""` record and blocks on the `signalCh` send or the ack wait behind a slow recycle round → B's caller ctx times out → `resume(cmd)` → `Delete("")` wipes both A's and B's records → GC resumes while A still considers itself paused.

*Suggestion:* the outer-timeout collateral delete predates this PR, but the two new rollback points at `:441` and `:453` reuse the same shared-ticket `resume`, widening the trigger surface. Assign each insertion a unique internal record id and delete by id, or force a non-empty generated ticket at the empty-ticket entry point.

*(raised by chyezh, tinswzy)*

## Low

**`internal/datacoord/services.go:1827` — `GcControl` flattens the PR's new typed errors into `UnexpectedError`, losing retriable semantics**

The point of this PR is to surface real pause failures to the caller, but `GcControl` projects them all onto `status.ErrorCode = commonpb.ErrorCode_UnexpectedError` plus a `Reason` string, leaving `status.Code` and `status.Retriable` unset (`services.go:1826-1830`), rather than using `merr.Status(err)`. On the client side, `merr.Error` sees `status.GetCode() == 0` and falls into the `OldCodeToMerr(ErrorCode)` branch; `UnexpectedError` has no mapping and hits `default → errUnexpected` with `retriable=false` (`pkg/util/merr/utils.go:296-300, 276-278`). As a result, pause timeouts (`DeadlineExceeded`) and the newly added `merr.WrapErrServiceUnavailable("garbage collector is closing")` (`garbage_collector.go:454`) — a genuinely transient, retriable condition — both arrive at `merr.CheckRPCCall` in `restful_mgr_routes.go:159` as non-retriable `UnexpectedError`, leaving callers unable to distinguish "temporarily unavailable, retry" from "actually broken". The new `TestPauseErrorSurfacesAsFailedStatus` only asserts `s.False(merr.Ok(resp))`, so this flattening goes undetected.

*Suggestion:* return `merr.Status(err), nil` so the merr code survives into the wire contract.

**`internal/datacoord/garbage_collector.go:442` — timeout branch's non-nil return rests on an implicit `cmd.timeout == cmd.ctx.Done()` invariant**

The timeout branches at lines 442 and 447 return `cmd.ctx.Err()`. That value is only guaranteed non-nil because `cmd.timeout` and `cmd.ctx.Done()` come from the same context — an invariant maintained solely at `Pause:298-299`, coupling this PR's new `ctx` field to the pre-existing `timeout` field. The two fields are demonstrably separable: the signal `gcCmd` that `pause()` builds internally (`:432-435`) sets `timeout` but not `ctx`. If `timeout` is ever decoupled into a standalone timer, or a `gcCmd` is constructed elsewhere, the timeout branch would silently return nil and resurrect the "timeout reported as Success" bug this PR fixes.

*Suggestion:* write it defensively — `if err := cmd.ctx.Err(); err != nil { return err }; return merr.WrapErrTimeout(...)` — or document the invariant on the `gcCmd` fields (`:86-94`).

*(raised by tinswzy)*

**`internal/datacoord/garbage_collector_test.go:2280` — new tests don't cover the shared empty-ticket rollback conflict**

The new subtest `pause_timeout_after_signal_received` (`:2280`) uses a unique ticket `"timeout-ticket"` against a freshly built `gc` with an empty heap; the pre-existing `pause_resume_timeout` (`:2257`) does use an empty ticket, but likewise times out on an empty heap. Both therefore only cover "timeout returns an error + `pauseUntil` resets to zero", and neither exercises the conflict path where an in-effect empty-ticket global pause is wiped by a second empty-ticket pause's rollback (the medium finding above).

*Suggestion:* add a regression test that first performs a successful `Pause(ctx, -1, "", time.Minute)`, then lets a second empty-ticket `Pause` time out, and asserts `gc.pauseUntil.PauseUntil()` still holds the first call's deadline.

*(raised by tinswzy)*

xiaofan-luan:
@bigsheeper Friendly ping — could you take a look at the review feedback above when you get a chance? The medium one (shared empty-ticket rollback in `resume(cmd)` wiping another caller's pause record) probably needs to be addressed before merge; the `merr.Status(err)` one should be a one-liner. Thanks!

bigsheeper:
@xiaofan-luan thanks for the ping, and thanks to both reviewers — all of it verified and addressed in `de3ce2e1e`. Details, including one correction to the review itself.

**1. `merr.Status(err)` (services.go:1827, :1838) — fixed.**

Confirmed as the most consequential of the findings, and it's this PR's own gap rather than a pre-existing one: the three error classes this PR introduces — `context.Canceled`, `context.DeadlineExceeded`, and the transient `"garbage collector is closing"` — were *exactly* the ones being flattened. The PR's stated benefit was not actually reaching the wire. Both sites now use `merr.Status(err)`, matching `:1822` / `:1834` in the same function.

One thing I checked before making the change, since it could have been a regression: `proxy/management.go:161` branches on the **deprecated** `ErrorCode` field, not on `Code`. `merr.Status` sets `ErrorCode: oldCode(code)` (`merr/utils.go:123`), and `oldCode` has no mapping for `TimeoutCode` (10001) or `ErrServiceUnavailable` (2), so both fall through to `default → UnexpectedError` (`utils.go:233-234`) — non-Success. The proxy path is unchanged. There's now an assertion pinning that.

**2. Shared empty-ticket rollback (garbage_collector.go:441/446/453) — fixed.**

Mechanism confirmed end to end: `Delete`'s predicate drops *every* record matching the ticket, and `ControlDataCoordGC` (`restful_mgr_routes.go:140-145`) sends only `"duration"`, so every pause from that route is a global pause with `ticket=""`. Your attribution is also right — the outer-timeout `resume(cmd)` predates this PR, but the two rollback points added here reuse it, and the inner ack wait is precisely where a slow recycle round parks.

Worth adding: the existing subtest at `garbage_collector_test.go:2238` already performs two successful empty-ticket pauses and asserts the heap keeps both, so "two `""` records coexist" is established behaviour, not a corner case.

Fix follows your first suggestion. `gcPauseRecord` gains an `id`, `Insert` returns it, and the three rollback sites go through a new `rollbackPause(cmd, recordID)` that deletes by id. The user-facing `Resume` keeps ticket-scoped `Delete` — that's the correct semantic for "release my ticket".

**Scope I did not change, deliberately:** two callers holding the *same non-empty* ticket still clobber each other on an explicit `Resume(ticket)`. That is the pre-existing "one ticket, one pause" semantic and I did not want to redefine it here. This fix covers the **rollback** path only.

**3. `Pause`/`Resume` not guarded by `gc.ctx` — fixed.**

Verified including your scoping, which was accurate: `gc.wg.Add(4)` appears only at `:332`, so the handler goroutines aren't tracked and `gc.wg.Wait()` doesn't hang; the production symptom is a `GcControl` RPC parked until client disconnect. Both outer selects now carry a `gc.ctx.Done()` arm, mirroring what this PR added inside `pause()`.

**4. Test assertion — fixed, and one correction.**

Your primary suggestion works and is in: `s.Equal(merr.TimeoutCode, resp.GetCode())`.

Your fallback, `s.ErrorIs(merr.Error(resp), context.DeadlineExceeded)`, **does not hold** — I tried it and it failed. `merr.Error` rebuilds a fresh `milvusError` from the status code (`utils.go:287-303`); it does not restore the original sentinel, so `errors.Is` against `context.DeadlineExceeded` can never succeed across the wire projection. Flagging it in case the same suggestion goes to other PRs.

Also added `TestPauseUnavailableStaysRetriable`, which covers the half of finding 1 that the timeout case does not: that `"garbage collector is closing"` arrives with `Retriable=true`.

**5. `cmd.ctx` / `cmd.timeout` invariant (garbage_collector.go:442) — not changed.**

Confirmed the fields are separable, and there's a second example beyond the one you cited: `Resume` (`:315-321`) also builds a `gcCmd` with `timeout` but no `ctx`. Current code is correct — `pause()` only ever receives a cmd built by `Pause`, which sets both — so I left it rather than grow this PR further.

**On verification:** the `GcControlServiceSuite` needs a live etcd, so I ran it against a local one rather than relying on CI. The new rollback test was checked against the pre-fix `resume(cmd)` and fails there (first caller's deadline wiped to zero), so it isn't a no-op. Full `./internal/datacoord/...` is green. I could not complete a repo-wide `make static-check` locally — an unrelated cgo typecheck error in `internal/util/segcore` from stale local C++ artifacts — so I ran the same lint config scoped to `./internal/datacoord/...`, which is clean; CI covers the rest.

czs007:

**Verified:**

- `internal/datacoord/garbage_collector.go:129` — pause-record IDs are allocated and inserted under the same mutex, and heap rebuilds preserve IDs, so rollback identifies exactly the record created by the failed command.
- `internal/datacoord/garbage_collector.go:417` — control commands are consumed serially by a single `startControlLoop` goroutine, so the insert and its subsequent rollback cannot interleave with another pause/resume mutation.
- `internal/datacoord/garbage_collector.go:469` — both the pre-send and post-send timeout paths roll back the inserted record and return `cmd.ctx.Err()`, so a timed-out pause can no longer report success.
- `internal/datacoord/garbage_collector.go:497` — rollback deletes by record ID rather than by ticket, so a failed empty-ticket pause cannot remove another caller's still-valid pause for the same collection.
- `internal/datacoord/garbage_collector.go:328` — the public `Pause`/`Resume` sends now select on collector shutdown as well as the caller context, so an unbuffered `cmdCh` cannot strand callers after the control loop exits.
- `internal/datacoord/services.go:1827` — `merr.Status` preserves the typed error code and retriability while keeping the deprecated `ErrorCode` non-success, so both the RESTful and proxy consumers still see failure.

czs007 (approved):
/lgtm

Issue #51538, opened 2026-07-17
[Bug]: TimestamptzArithCompareExpr evaluates NULL rows with the storage placeholder on the month/year-interval path


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

- Milvus version: master (c4fa9ea356)

### Current Behavior

\`PhyTimestamptzArithCompareExpr::ExecCompareVisitorImplForAll\` (internal/core/src/exec/expression/TimestamptzArithCompareExpr.cpp) — the manual gmtime/timegm path used for year/month intervals — never checks \`valid_data\` and never clears \`valid_res\`. NULL rows of a nullable TIMESTAMPTZ column are computed using the dense storage placeholder (epoch 0):

- \`ts_col + INTERVAL 1 month < '2000-01-01'\` returns every NULL row (1970-02-01 < 2000-01-01 evaluates true).
- \`not (ts_col + INTERVAL 1 month < '1900-01-01')\` also returns every NULL row, because \`valid_res\` is left true and the negation flips the placeholder result.

The day/hour-interval path delegates to \`PhyBinaryArithOpEvalRangeExpr\`, which handles NULLs correctly — so the same predicate returns different rows depending on the interval unit.

Two adjacent gaps in the same lambda: it also lacks the \`data == nullptr\` skip-batch guard and never resolves the \`offsets\` array for \`FilterType::random\` (latent only — sealed segments are always chunked today, so the whole-batch random path is currently unreachable).

### Expected Behavior

NULL rows never match, under either polarity (three-valued logic), regardless of interval unit — consistent with every other comparison expression.

### Steps To Reproduce

Regression tests (red on master, green with the fix) in the fix PR: \`OffsetsEvalCorrectnessTest.TimestamptzNullRowsNeverMatch\` and \`TimestamptzOffsetInputPerCandidate\`.

### Anything else?

Found by an adversarial multi-agent review during #51140 (pre-existing on master, unrelated to that PR). Fix PR incoming.

PR #51608, opened 2026-07-19
fix: honor NULL semantics in timestamptz arithmetic compare

xiaofan-luan:
issue: #51538

## What changed

- Treat nullable TIMESTAMPTZ rows as SQL UNKNOWN on the manual year/month interval arithmetic path.
- Add focused regression coverage for growing and sealed segments, both sequential scans and offset-input evaluation.

## Root cause

`PhyTimestamptzArithCompareExpr` evaluated the storage placeholder for NULL rows without consulting `valid_data`. This could return NULL rows, especially after logical negation, and disagreed with the fixed-duration arithmetic path.

## Validation

- `clang++ -fsyntax-only` passed for the new test translation unit using the repository-generated compile flags.
- `git diff --check` passed.
- The full C++ test binary could not be linked locally because the existing Tantivy Rust dependency build cannot find macOS system headers (`string.h`, `stdio.h`); this failure occurs before compiling this change.

### DIFF
--- a/internal/core/src/exec/expression/TimestamptzArithCompareExpr.cpp
+++ b/internal/core/src/exec/expression/TimestamptzArithCompareExpr.cpp
@@ -108,6 +108,12 @@ PhyTimestamptzArithCompareExpr::ExecCompareVisitorImplForAll(
             proto::plan::Interval interval) {
         const int64_t compare_us = compare_value;
         for (int i = 0; i < size; ++i) {
+            if (valid_data != nullptr && !valid_data[i]) {
+                // NULL never matches, under either polarity (three-valued
+                // logic); do not evaluate the storage placeholder value.
+                res[i] = valid_res[i] = false;
+                continue;
+            }
             const int64_t current_ts_us = data[i];
             const int op_sign =
                 (arith_op == proto::plan::ArithOpType::Add) ? 1 : -1;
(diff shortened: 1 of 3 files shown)


czs007:
Adversarial review surfaced two low-severity maintainability and test-coverage items in the Timestamptz arithmetic-compare kernel and its tests; neither is a live bug, but both are cheap to close.

### Low

- **`internal/core/src/exec/expression/TimestamptzArithCompareExpr.cpp:110`** — The kernel indexes `data[i]`/`valid_data[i]`/`res[i]` sequentially and never dereferences the `offsets` parameter, diverging from the gather-via-`offsets[i]` convention every other kernel follows (e.g. `BinaryArithOpEvalRangeExpr.cpp:205`). This is safe today because the only `SegmentSealed` implementation is `ChunkedSegmentSealedImpl` (always `is_chunked() == true`), leaving the non-chunked sealed branch at `Expr.h:898` — which passes base pointers plus `offsets` expecting gather semantics — dead. Add a one-line comment noting the kernel assumes per-row/sequential dispatch, so a future non-chunked sealed path does not silently read wrong rows. *(raised by tinswzy)*

- **`internal/core/unittest/test_timestamptz_arith_compare.cpp:141`** — `AssertOffsetInputNullSemantics` exercises only `kFarFutureUs` with `negated=false`. Negation is orthogonal (applied to the whole result vector post-kernel, so the sequential `AssertNullSemantics` path already covers it), making this low-value, but adding a negated offset-input case would cheaply complete the coverage matrix. *(raised by tinswzy)*

liliu-z (commented):
/lgtm
/approve

Issue #51617, opened 2026-07-20
[Bug]: Empty list expression template parameters fail across filter operations


zhuwenxing:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Environment

- Milvus version: master-based build, commit `316023629315febad293794533208bbc8703f46e` (base master: `b9621d7d325dea78fc8fd33c9d2ba0dcbda24c04`)
- Deployment mode: standalone
- MQ type: woodpecker
- SDK version: PyMilvus `3.1.0rc64`
- OS: Ubuntu 22.04.4 LTS in the Milvus container; Rocky Linux 8.10 on the Kubernetes node
- CPU/Memory: Milvus pod request and limit: 2 CPU / 8 GiB; node capacity: 64 CPU / approximately 503 GiB
- GPU: N/A
- Others: Python 3.12.12, Kubernetes v1.24.2

### K8s Pod List

```text
NAME                                                            READY   STATUS    RESTARTS   AGE
issue-51416-array-template-etcd-0                               1/1     Running   0          4m13s
issue-51416-array-template-milvus-standalone-65b977489f-n28w9   1/1     Running   0          3m40s
issue-51416-array-template-minio-74b46bc978-tl64n               1/1     Running   0          4m12s
```

## Current Behavior

An ordinary `ARRAY<VARCHAR>` field cannot be used with `array_contains_all` when its filter template parameter is an empty list:

```python
filter="array_contains_all(tags, {tags})"
filter_params={"tags": []}
```

Both query and search fail with:

```text
failed to create query plan: unknown template variable value type: <nil>: query plan failed: invalid parameter
```

The following controls succeed:

```python
filter="array_contains_all(tags, [])"
```

```python
filter="array_contains_all(tags, {tags})"
filter_params={"tags": ["blue"]}
```

This confirms that ordinary Array fields are affected and that the failure is specifically triggered by an empty-list template value.

## Expected Behavior

The parameterized expression should have the same behavior as its inline equivalent:

```python
array_contains_all(tags, {tags})
# filter_params={"tags": []}
```

```python
array_contains_all(tags, [])
```

For the reproduction data below, both forms should return IDs `[1, 2]`.

## Steps To Reproduce

1. Deploy the master-based Milvus build described above.
2. Install PyMilvus `3.1.0rc64`.
3. Set `MILVUS_URI` and, if authentication is enabled, `MILVUS_TOKEN`.
4. Run the script below.
[...]

Issue #51416, opened 2026-07-15
[Bug]: Struct MATCH and element_filter do not bind filter template values


zhuwenxing:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Environment

- Milvus version: master, commit `6ca035094475865e72e90526de2aff4736f9b1b8`
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): woodpecker
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus `3.1.0rc62`
- OS(Ubuntu or CentOS): Ubuntu 22.04.4 LTS in the Milvus container; Ubuntu 20.04.6 LTS on the Kubernetes node
- CPU/Memory: Milvus pod request and limit: 2 CPU / 8 GiB; node capacity: 64 CPU / approximately 503 GiB
- GPU: N/A
- Others: Python 3.12.12

### K8s Pod List

```text
NAME                                                              READY   STATUS    RESTARTS   AGE
issue-51381-e2e-0715-etcd-0                                       1/1     Running   0          4h15m
issue-51381-e2e-0715-milvus-standalone-58b8b4b4f8-n57f8           1/1     Running   0          4h15m
issue-51381-e2e-0715-minio-64fd7fb7fd-w8wh9                       1/1     Running   0          4h15m
```

## Current Behavior

Filter template values inside Struct `MATCH_*` and `element_filter` expressions are not bound.

The equivalent inline expressions work:

```text
MATCH_LEAST(profile, $[p_int] >= 10, threshold=1)
element_filter(profile, $[p_tag] == "blue")
```

The template forms fail:

```text
MATCH_LEAST(profile, $[p_int] >= {min_score}, threshold=1)
element_filter(profile, $[p_tag] == {tag})
```

The `MATCH_LEAST` template fails because segcore expects an Int64 `GenericValue`:

```text
Operator::GetOutput failed for [Operator:PhyFilterBitsNode] :
Assert "(value_proto.val_case() ==
milvus::proto::plan::GenericValue::kInt64Val)"
at /workspace/milvus/internal/core/src/exec/expression/Utils.h:239
```

The `element_filter` template fails because segcore expects a String `GenericValue`:

```text
Operator::GetOutput failed for [Operator:PhyElementFilterBitsNode] :
Assert "(value_proto.val_case() ==
milvus::proto::plan::GenericValue::kStringVal)"
[...]

PR #51621, opened 2026-07-20
fix: support empty array expression templates

zhuwenxing:
issue: #51617

## What this PR does

- Preserve empty array expression template parameters instead of treating them as unknown values.
- Keep empty and nested arrays intact through REST and Go SDK conversion paths, and return parameter errors for unsupported nil/object values instead of panicking.
- Report empty and nested array parameter lengths correctly in access logs.
- Reject unsupported nested array-valued `IN` operands during planning.
- Add unit and Python E2E coverage for query, search, hybrid search, and delete.

## How this was tested

- Affected Go unit test packages passed with `milvus-dev-cli`.
- Full `internal/parser/planparserv2` tests passed on commit `f29510b`.
- `tests/python_client/milvus_client/test_milvus_client_empty_array_template.py`: `1 passed in 6.18s`.
- Deployed commit `f29510b` and confirmed the server-reported `git_commit`; teardown left no collections.

### DIFF
--- a/client/milvusclient/read_options.go
+++ b/client/milvusclient/read_options.go
@@ -264,11 +264,18 @@ func slice2TmplArrayValue(rv reflect.Value) (*schemapb.TemplateArrayValue, error
 	}
 
 	result := &schemapb.TemplateArrayValue{}
-	if rv.Len() == 0 {
-		return result, nil
-	}
 
-	_, elementKind, err := templateArrayElement(rv.Index(0))
+	var elementKind reflect.Kind
+	if rv.Len() == 0 {
+		// Preserve the data oneof for typed empty slices so older servers can
+		// decode them. An empty interface slice has no encodable element type.
+		if rv.Type().Elem().Kind() == reflect.Interface {
+			return result, nil
+		}
+		elementKind, err = normalizeTemplateArrayElementKind(rv.Type().Elem().Kind())
+	} else {
+		_, elementKind, err = templateArrayElement(rv.Index(0))
+	}
 	if err != nil {
 		return nil, err
 	}
@@ -365,21 +372,29 @@ func templateArrayElement(rv reflect.Value) (reflect.Value, reflect.Kind, error)
 		return reflect.Value{}, reflect.Invalid, err
 	}
 
-	switch rv.Kind() {
+	kind, err := normalizeTemplateArrayElementKind(rv.Kind())
+	if err != nil {
+		return reflect.Value{}, reflect.Invalid, err
+	}
+	return rv, kind, nil
+}
+
+func normalizeTemplateArrayElementKind(kind reflect.Kind) (reflect.Kind, error) {
+	switch kind {
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
-		return rv, reflect.Int64, nil
+		return reflect.Int64, nil
 	case reflect.Bool:
-		return rv, reflect.Bool, nil
+		return reflect.Bool, nil
 	case reflect.Float32, reflect.Float64:
-		return rv, reflect.Float64, nil
+		return reflect.Float64, nil
 	case reflect.String:
-		return rv, reflect.String, nil
+		return reflect.String, nil
 	case reflect.Slice:
-		return rv, reflect.Slice, nil
+		return reflect.Slice, nil
 	default:
-		return reflect.Value{}, reflect.Invalid, fmt.Errorf(
+		return reflect.Invalid, fmt.Errorf(
 			"unsupported template type: slice of %v",
-			rv.Kind(),
+			kind,
 		)
 	}
 }
--- a/client/milvusclient/read_options.go
+++ b/client/milvusclient/read_options.go
@@ -242,55 +242,159 @@ func slice2TmplValue(val any) (*schemapb.TemplateValue, error) {
 		return &schemapb.TemplateValue{Val: arrVal}, nil
 	}
 
-	rv := reflect.ValueOf(val)
-	switch t := reflect.TypeOf(val).Elem().Kind(); t {
-	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
-		data := make([]int64, 0, rv.Len())
-		for i := 0; i < rv.Len(); i++ {
-			data = append(data, rv.Index(i).Int())
-		}
-		arrVal.ArrayVal.Data = &schemapb.TemplateArrayValue_LongData{
-			LongData: &schemapb.LongArray{
-				Data: data,
-			},
-		}
-	case reflect.Bool:
-		data := make([]bool, 0, rv.Len())
-		for i := 0; i < rv.Len(); i++ {
-			data = append(data, rv.Index(i).Bool())
-		}
-		arrVal.ArrayVal.Data = &schemapb.TemplateArrayValue_BoolData{
-			BoolData: &schemapb.BoolArray{
-				Data: data,
-			},
-		}
-	case reflect.Float32, reflect.Float64:
-		data := make([]float64, 0, rv.Len())
-		for i := 0; i < rv.Len(); i++ {
-			data = append(data, rv.Index(i).Float())
-		}
-		arrVal.ArrayVal.Data = &schemapb.TemplateArrayValue_DoubleData{
-			DoubleData: &schemapb.DoubleArray{
-				Data: data,
-			},
-		}
-	case reflect.String:
-		data := make([]string, 0, rv.Len())
-		for i := 0; i < rv.Len(); i++ {
-			data = append(data, rv.Index(i).String())
-		}
-		arrVal.ArrayVal.Data = &schemapb.TemplateArrayValue_StringData{
-			StringData: &schemapb.StringArray{
-				Data: data,
-			},
-		}
-	default:
-		return nil, fmt.Errorf("unsupported template type: slice of %v", t)
+	arrayVal, err := slice2TmplArrayValue(reflect.ValueOf(val))
+	if err != nil {
+		return nil, err
 	}
 
 	return &schemapb.TemplateValue{
-		Val: arrVal,
+		Val: &schemapb.TemplateValue_ArrayVal{
+			ArrayVal: arrayVal,
+		},
 	}, nil
+}
+
+func slice2TmplArrayValue(rv reflect.Value) (*schemapb.TemplateArrayValue, error) {
+	rv, err := unwrapTemplateArrayElement(rv)
+	if err != nil {
+		return nil, err
+	}
+	if rv.Kind() != reflect.Slice {
+		return nil, fmt.Errorf("unsupported template array value type: %v", rv.Kind())
+	}
+
+	result := &schemapb.TemplateArrayValue{}
+	if rv.Len() == 0 {
+		return result, nil
+	}
+
+	_, elementKind, err := templateArrayElement(rv.Index(0))
+	if err != nil {
+		return nil, err
+	}
+
+	elementAt := func(index int) (reflect.Value, error) {
+		element, kind, err := templateArrayElement(rv.Index(index))
+		if err != nil {
+			return reflect.Value{}, err
+		}
+		if kind != elementKind {
+			return reflect.Value{}, fmt.Errorf(
+				"template array elements have inconsistent types: %v and %v",
+				elementKind,
+				kind,
+			)
+		}
+		return element, nil
+	}
+
+	switch elementKind {
+	case reflect.Int64:
+		data := make([]int64, rv.Len())
+		for i := range data {
+			element, err := elementAt(i)
+			if err != nil {
+				return nil, err
+			}
+			data[i] = element.Int()
+		}
+		result.Data = &schemapb.TemplateArrayValue_LongData{
+			LongData: &schemapb.LongArray{Data: data},
+		}
+	case reflect.Bool:
+		data := make([]bool, rv.Len())
+		for i := range data {
+			element, err := elementAt(i)
+			if err != nil {
+				return nil, err
+			}
+			data[i] = element.Bool()
+		}
+		result.Data = &schemapb.TemplateArrayValue_BoolData{
+			BoolData: &schemapb.BoolArray{Data: data},
+		}
+	case reflect.Float64:
+		data := make([]float64, rv.Len())
+		for i := range data {
+			element, err := elementAt(i)
+			if err != nil {
+				return nil, err
+			}
+			data[i] = element.Float()
+		}
+		result.Data = &schemapb.TemplateArrayValue_DoubleData{
+			DoubleData: &schemapb.DoubleArray{Data: data},
+		}
+	case reflect.String:
+		data := make([]string, rv.Len())
+		for i := range data {
+			element, err := elementAt(i)
+			if err != nil {
+				return nil, err
+			}
+			data[i] = element.String()
+		}
+		result.Data = &schemapb.TemplateArrayValue_StringData{
+			StringData: &schemapb.StringArray{Data: data},
+		}
+	case reflect.Slice:
+		data := make([]*schemapb.TemplateArrayValue, rv.Len())
+		for i := range data {
+			element, err := elementAt(i)
+			if err != nil {
+				return nil, err
+			}
+			data[i], err = slice2TmplArrayValue(element)
+			if err != nil {
+				return nil, err
+			}
+		}
+		result.Data = &schemapb.TemplateArrayValue_ArrayData{
+			ArrayData: &schemapb.TemplateArrayValueArray{Data: data},
+		}
+	default:
+		return nil, fmt.Errorf("unsupported template type: slice of %v", elementKind)
+	}
+
+	return result, nil
+}
+
+func templateArrayElement(rv reflect.Value) (reflect.Value, reflect.Kind, error) {
+	rv, err := unwrapTemplateArrayElement(rv)
+	if err != nil {
+		return reflect.Value{}, reflect.Invalid, err
+	}
+
+	switch rv.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return rv, reflect.Int64, nil
+	case reflect.Bool:
+		return rv, reflect.Bool, nil
+	case reflect.Float32, reflect.Float64:
+		return rv, reflect.Float64, nil
+	case reflect.String:
+		return rv, reflect.String, nil
+	case reflect.Slice:
+		return rv, reflect.Slice, nil
+	default:
+		return reflect.Value{}, reflect.Invalid, fmt.Errorf(
+			"unsupported template type: slice of %v",
+			rv.Kind(),
+		)
+	}
+}
+
+func unwrapTemplateArrayElement(rv reflect.Value) (reflect.Value, error) {
+	for rv.IsValid() && rv.Kind() == reflect.Interface {
+		if rv.IsNil() {
+			return reflect.Value{}, fmt.Errorf("unsupported template type: nil array element")
+		}
+		rv = rv.Elem()
+	}
+	if !rv.IsValid() {
+		return reflect.Value{}, fmt.Errorf("unsupported template type: invalid array element")
+	}
+	return rv, nil
 }
 
 func (r *AnnRequest) WithANNSField(annsField string) *AnnRequest {
--- a/client/milvusclient/read_options.go
+++ b/client/milvusclient/read_options.go
@@ -204,7 +204,8 @@ func any2TmplValue(val any) (*schemapb.TemplateValue, error) {
 		// base64 inflation.
 		result.Val = &schemapb.TemplateValue_BytesVal{BytesVal: v}
 	default:
-		if reflect.TypeOf(val).Kind() == reflect.Slice {
+		valueType := reflect.TypeOf(val)
+		if valueType != nil && valueType.Kind() == reflect.Slice {
 			return slice2TmplValue(val)
 		}
 		return nil, fmt.Errorf("unsupported template value type: %T", val)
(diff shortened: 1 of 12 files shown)


xiaocai2333 (commented):
Reviewed the full diff plus the surrounding code, and triaged all four red checks. Overall: **needs work** — the direction is right and the scope is genuinely broader than the empty-list bug itself (REST, Go SDK, access log, e2e all covered), and turning three `panic` sites into classified errors is a real improvement. Two things need attention before this can land, one of them a hard merge blocker.

## CI triage — 3 of the 4 failures are not this PR

| Check | Verdict |
|---|---|
| `ci-v2/e2e-default` | **Real blocker, caused by this PR** — see below |
| `ci-v2/ut-cpp` | Unrelated to this PR |
| `codecov/project`, `codecov/project/Go` | Noise — `codecov/patch` passes |
| `ci-v2/e2e-pulsar` | Self-declared non-blocking ("Tencent E2E gray only; failure does not affect ci-passed") |

### e2e-default — merge blocker, needs a one-line change in this PR

Job 43700: `1 failed, 6124 passed`. The single failure is

```
FAILED milvus_client/test_milvus_client_struct_array_nullable.py::
  TestMilvusClientStructArraySchemaEvolution::
  test_struct_array_filter_template_matches_inline_expression[contains_all_empty_list]
[XPASS(strict)]
```

That case carries `pytest.mark.xfail(reason="https://github.com/milvus-io/milvus/issues/51416", strict=True)` at `tests/python_client/milvus_client/test_milvus_client_struct_array_nullable.py:1285-1286`. It was marked xfail precisely because of the empty-list bug this PR fixes. Now that the bug is gone the case passes, and `strict=True` turns an unexpected pass into a failure.

So this failure is actually end-to-end proof that the fix works against a real cluster — but it does mean **the marker has to be removed in this PR**. If this merges with the marker in place, master's e2e goes red immediately. Suggested change: drop the two `marks=pytest.mark.xfail(...)` lines, keep `id="contains_all_empty_list"`.

(I hit the identical failure independently while validating the same root cause on a separate branch, so this reproduces across two independent e2e runs.)

### ut-cpp — unrelated

Job 13679: `7806 tests ran, 7796 PASSED, 10 SKIPPED`, zero test failures. The stage fails on LeakSanitizer:

```
Direct leak of 64 byte(s) in 1 object(s):
  #1 milvus::FailureCStatus(...) EasyAssert.h:130
  #2 FlushGrowingSegmentData  segcore/segment_c.cpp:1972
  #3 FlushGrowingSegmentTest_FlushRejectsEndOffsetBeyondRowCount_Test::TestBody()
     segcore/flush_growing_segment_test.cpp:1681
```

This PR changes 16 files, all Go plus one Python test, and zero C++ — so it cannot be the cause. Looks like a pre-existing leak on the `FailureCStatus` error path in growing-segment flush. (I did not confirm it reproduces on a clean master; that judgement is based only on this PR having no C++ changes.)

## Review findings

Inline comments cover the rest. Summary by severity:

- **High** — the array-valued `IN` rejection removes previously supported, explicitly tested behavior (2 inline comments)
- **High** — the strict-xfail blocker above (no diff line to anchor to)
- **Medium** — unreachable branch in `getTemplateArrayType`
- **Low** — Python test mixes raw client with framework wrappers; no collection cleanup; access-log length sentinel changes from `-1` to `0`

Dimensions checked with no findings: concurrency safety (no shared state, no goroutines, all pure functions), error handling (the `panic` → `merr` conversion is correct, and `ErrParameterInvalid` is the right classification since the request content itself forces the branch), performance (no change in algorithmic shape or allocation behavior), resource leaks on the Go side.

Worth calling out as a fix beyond the stated scope: before this change, `exprParams: {"ids": null}` or a nested object reached `panic(...)` in `getTemplateType` on the REST path. That is now a proper 400-style error, and `TestInvalidExprParamsRejectBeforeProxyDispatch` locks it in.

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1167,6 +1167,9 @@ func (v *ParserVisitor) VisitTerm(ctx *parser.TermContext) interface{} {
 		array := elementValue.GetArrayVal().GetArray()
 		values = make([]*planpb.GenericValue, len(array))
 		for i, e := range array {
+			if IsArray(e) {
```
xiaocai2333: **High — this removes previously supported, explicitly tested behavior.**

This guard rejects every array element unconditionally, without looking at the column type. But `castValue` (`internal/parser/planparserv2/utils.go:254`) deliberately supports two array cases:

```go
if typeutil.IsJSONType(dataType) { return value, nil }                    // :255 — JSON column accepts any value, arrays included
if typeutil.IsArrayType(dataType) && IsArray(value) { return value, nil } // :258 — Array column vs array value
```

Since the new check runs *before* `castValue`, both paths are now dead. After upgrading, queries like

```
json_col["arr"] in [[1,2],[3,4]]
ArrayField in [[1,2,3]]
```

fail to parse, where they previously planned fine.

The diff's own test changes confirm this is a behavior change rather than a bug fix: `range_json_test.go` rewrites `TestRewrite_JSON_NestedPath_ArrayNotEqualRewriteBlocked`, which used to assert that `JSONField["arr"] not in [[1, 2]]` parses and yields a `Not` unary, into a test asserting it errors. And `fill_expression_value_test.go` moves `ArrayField in {list}` (array-of-arrays) from the valid group to the invalid group.

The PR body describes this only as "Reject unsupported nested array-valued `IN` operands during planning", without saying why these operands are unsupported, and it is not flagged as a breaking change.

**Suggestion:** if the motivation is that segcore misbehaves or crashes on array terms, please link that evidence and narrow the guard to the column types that genuinely cannot handle array operands — let `castValue` keep deciding for JSON and Array columns. If it is defensive hardening rather than a known failure, consider dropping it. Either way this is orthogonal to empty-array template support and would be easier to review (and to revert) as its own PR with an explicit breaking-change note.

internal/parser/planparserv2/fill_expression_value.go
```diff
@@ -74,6 +74,9 @@ func FillTermExpressionValue(expr *planpb.TermExpr, templateValues map[string]*p
 	array := value.GetArrayVal().GetArray()
 	values := make([]*planpb.GenericValue, len(array))
 	for i, e := range array {
+		if IsArray(e) {
```
xiaocai2333: **High — same issue as the inline path, see the comment on `parser_visitor.go:1170`.**

This is the template-value counterpart of the same unconditional rejection, and it has the same consequence: `dataType` here has already been resolved to the column's element type where relevant (lines 66-72), so `castValue` was in a position to accept legitimate array operands for JSON and Array columns. This guard short-circuits that.

Concretely, this is what moves `ArrayField in {list}` from the valid to the invalid group in `fill_expression_value_test.go` in this same diff.

**Suggestion:** keep this consistent with whatever is decided for the inline path — narrow to the column types that actually cannot accept arrays, or drop it.

internal/parser/planparserv2/rewriter/range_json_test.go
```diff
@@ -105,16 +105,25 @@ func TestRewrite_JSON_NestedPath_ScalarNotEqualRewriteBlocked(t *testing.T) {
...
+	require.NotNil(t, unary, "array-valued JSON missing-path NOT must remain explicit: %s", exprStr)
+	require.Equal(t, planpb.UnaryExpr_Not, unary.GetOp(), exprStr)
+}
+
+func TestRewrite_JSON_NestedPath_ArrayTermRejected(t *testing.T) {
+	helper := buildSchemaHelperWithJSON(t)
+
 	for _, exprStr := range []string{
-		`not (JSONField["arr"] == [1, 2])`,
+		`JSONField["arr"] in [[1, 2]]`,
```
xiaocai2333: Flagging this as the clearest evidence of the behavior change discussed on `parser_visitor.go:1170`.

`JSONField["arr"] not in [[1, 2]]` moved from "parses successfully, rewrite blocked, returns an explicit `Not` unary" to "must return an error". That was deliberate, asserted behavior before this PR — the old test name (`...ArrayNotEqualRewriteBlocked`) and its assertion on `planpb.UnaryExpr_Not` show the array-valued JSON term path was intentionally supported.

If the rejection is kept, this is a user-visible breaking change for JSON array predicates and should be called out in the PR description and release notes.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -3340,52 +3350,65 @@ func generateTemplateArrayData(list []interface{}) *schemapb.TemplateArrayValue
...
+		return nil, merr.WrapErrParameterInvalidMsg("unsupported exprParams array value type %s", dtype.String())
 	}
-	return data
+	return data, nil
 }
 
-func getTemplateArrayType(value []interface{}) schemapb.DataType {
-	dtype := getTemplateType(value[0])
+func getTemplateArrayType(value []interface{}) (schemapb.DataType, error) {
+	if len(value) == 0 {
```
xiaocai2333: **Medium — unreachable branch.**

`getTemplateArrayType` has exactly one caller, `generateTemplateArrayData` at line 3267, and that caller already returns early on line 3263:

```go
if len(list) == 0 {
    return &schemapb.TemplateArrayValue{}, nil
}
```

So `len(value) == 0` here can never be true, and the `DataType_None` return value never reaches any switch. Leaving it in suggests callers must handle a `None` result when they don't.

**Suggestion:** pick one of the two places to handle empty. Either delete this branch, or drop the early return in `generateTemplateArrayData` and let `getTemplateArrayType` return `None` with the switch handling it — but not both.

internal/proxy/accesslog/info/util.go
```diff
@@ -119,6 +119,10 @@ func getLengthFromTemplateValue(tv *schemapb.TemplateValue) int {
 		return len(arrayValues.GetStringData().GetData())
 	case *schemapb.TemplateArrayValue_JsonData:
 		return len(arrayValues.GetJsonData().GetData())
+	case *schemapb.TemplateArrayValue_ArrayData:
+		return len(arrayValues.GetArrayData().GetData())
+	case nil:
```
xiaocai2333, line 124: **Low — heads-up on a sentinel-value change, not a defect.**

`-1` was the "undefined" sentinel for this function; empty arrays now report a real length of `0`. That is the more correct semantics and the accompanying `util_test.go` update reflects it. Just noting that any downstream access-log tooling filtering on `-1` to spot malformed template values will stop seeing this case.

tests/python_client/milvus_client/test_milvus_client_empty_array_template.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+        self.create_collection(
+            client,
+            collection_name,
+            schema=schema,
+            index_params=index_params,
+            consistency_level="Strong",
+        )
+        client.insert(
```
xiaocai2333: **Low — mixes the framework wrapper with the raw client.**

`self.create_collection(...)` above goes through `TestMilvusClientV2Base`, which adds response checking and logging, but `client.insert(...)` here and the subsequent `client.query(...)` / `client.search(...)` / `client.hybrid_search(...)` / `client.delete(...)` calls bypass it. When this test fails in CI the framework context that normally makes triage quick won't be in the output.

**Suggestion:** use `self.insert` / `self.query` / `self.search` / `self.hybrid_search` / `self.delete` throughout for consistency with the rest of the suite.

tests/python_client/milvus_client/test_milvus_client_empty_array_template.py
```diff
@@ -0,0 +1,145 @@
...
+            filter=empty_filter,
+            filter_params=empty_params,
+        )
+        remaining_rows = client.query(
+            collection_name,
+            filter="id >= 1",
+            output_fields=["id"],
+            limit=10,
+        )
+        assert sorted(row["id"] for row in remaining_rows) == [1, 2]
```
xiaocai2333, line 153: **Low — no explicit cleanup.**

The test ends on a query assertion and never drops the collection it created. If `TestMilvusClientV2Base`'s teardown already handles this, ignore me — otherwise this leaves a collection behind on the shared CI cluster on every run.

zhuwenxing:
Thanks for the detailed review. I rebased the branch onto the latest master and addressed the actionable feedback in the latest revision:

- Removed the strict xfail marker from `contains_all_empty_list`, since the fixed behavior now passes.
- Restored the existing JSON/Array array-valued `IN` behavior.
- Removed the unreachable empty-array type branch.
- Switched the new E2E test to the framework DML wrappers.

The collection is already registered for cleanup by `TestMilvusClientV2Base.create_collection`, and reporting `0` for a valid empty array remains intentional.

Validation passed: incremental Go UT for all five affected packages with four shards, plus Ruff check and format check for the Python tests.

internal/parser/planparserv2/parser_visitor.go
zhuwenxing: Fixed in `e96213050e`. I removed the unconditional array-element rejection, so array-valued operands continue through `castValue` and preserve the existing JSON and Array behavior.

internal/parser/planparserv2/fill_expression_value.go
zhuwenxing: Fixed in `e96213050e`. The template-value path now also delegates array-valued operands to `castValue`, consistent with the inline path.

internal/parser/planparserv2/rewriter/range_json_test.go
zhuwenxing: Fixed in `e96213050e`. I restored the successful JSON array-valued `not in` coverage and removed the rejection-only test.

internal/distributed/proxy/httpserver/utils.go
zhuwenxing: Fixed in `7c75667366`. I removed the unreachable empty-array branch from `getTemplateArrayType`; empty arrays remain handled by `generateTemplateArrayData` before calling the helper.

internal/proxy/accesslog/info/util.go
zhuwenxing: Acknowledged. Reporting `0` is intentional because the value is a valid empty array; `-1` remains reserved for unsupported or undefined array layouts.

tests/python_client/milvus_client/test_milvus_client_empty_array_template.py
zhuwenxing: Fixed in `40f9efc35f`. The test now uses the framework wrappers for insert, flush, load, query, search, hybrid search, and delete.

tests/python_client/milvus_client/test_milvus_client_empty_array_template.py
zhuwenxing: Confirmed that `TestMilvusClientV2Base.create_collection` registers the collection in `tear_down_collection_names` by default, and the base teardown drops it. No explicit cleanup is needed in this test.

client/milvusclient/read_options.go
```diff
@@ -199,7 +199,8 @@ func any2TmplValue(val any) (*schemapb.TemplateValue, error) {
 	case string:
 		result.Val = &schemapb.TemplateValue_StringVal{StringVal: v}
 	default:
-		if reflect.TypeOf(val).Kind() == reflect.Slice {
+		valueType := reflect.TypeOf(val)
+		if valueType != nil && valueType.Kind() == reflect.Slice {
```
xiaocai2333, line 208: **Medium — the Go SDK still rejects generic empty slices and nested slices.**

The nil check prevents `reflect.TypeOf(nil).Kind()` from panicking, but every non-nil slice is still delegated to `slice2TmplValue`, whose switch only supports primitive element kinds. I reproduced both failures:

```go
any2TmplValue([]any{})       // unsupported template type: slice of interface
any2TmplValue([][]int64{{}}) // unsupported template type: slice of slice
```

A typed primitive empty slice such as `[]int64{}` works, but dynamically constructed empty parameters commonly arrive as `[]any{}`, and nested arrays cannot be encoded through the public Go SDK at all. This makes the new empty/nested-array support in the REST and planner paths unavailable through the Go SDK, contrary to the PR description.

Please consider encoding a zero-length slice with an unset `TemplateArrayValue` data oneof, recursively handling slice/interface elements for nested arrays, and adding SDK request tests for `[]any{}`, `[][]int64`, and nested empty arrays.

xiaocai2333 (commented):
I found one remaining issue in the Go SDK conversion path.

client/milvusclient/read_options.go
zhuwenxing: Fixed in `b9f02e2ef4`. The Go SDK now encodes zero-length slices with an unset `TemplateArrayValue` data oneof and recursively handles interface-backed and nested slices. Added request-level coverage for `[]any{}`, `[][]int64`, and nested empty arrays, plus conversion-level coverage for dynamic interface slices. The targeted Go unit tests pass.

xiaocai2333: Clarification: this is **not intended to block this PR from merging**. The core empty-array fix is valid, and typed primitive empty slices such as `[]int64{}` already work through the Go SDK. Support for generic `[]any{}` values and nested Go slices is a pre-existing SDK limitation rather than a regression introduced here. It is worth tracking as a follow-up (or narrowing the PR description accordingly), but it is not required for this fix.

client/milvusclient/read_options.go
```diff
@@ -208,59 +209,159 @@ func any2TmplValue(val any) (*schemapb.TemplateValue, error) {
...
+	rv, err := unwrapTemplateArrayElement(rv)
+	if err != nil {
+		return nil, err
+	}
+	if rv.Kind() != reflect.Slice {
+		return nil, fmt.Errorf("unsupported template array value type: %v", rv.Kind())
+	}
+
+	result := &schemapb.TemplateArrayValue{}
+	if rv.Len() == 0 {
```
xiaocai2333, line 269: **Medium — encoding every zero-length slice as an unset data oneof regresses typed empty slices against older servers.**

`slice2TmplArrayValue` now early-returns a `TemplateArrayValue{}` with no data oneof whenever `rv.Len() == 0`, including for typed slices such as `[]int64{}`. Before this commit, `[]int64{}` was encoded as `LongData{Data: []}`, which servers **without** this PR already accept: `convertArrayValue`'s `LongData` branch handles empty lists and fills an empty-values `TermExpr`, same as inline `in []`.

The unset-oneof form, however, is only understood by servers that carry this PR's new `case nil` in `convertArrayValue` — any server without it answers `unknown template variable value type: <nil>`. So a new Go SDK talking to an existing server turns a previously working `[]int64{}` template parameter into an error. That is a client-side compatibility regression independent of the empty-array feature itself.

Suggestion: keep the typed empty encoding whenever the element type is statically known — typed slices, including the inner `[]int64{}` of `[][]int64{{}}`, where the recursion can read `rv.Type().Elem()` — and reserve the unset oneof for `[]any{}` where no element type is available. That preserves old-server compatibility while still enabling the untyped empty case this PR fixes.

Two minor notes, no action strictly required:
- `[]any{1, "a"}` now fails with "template array elements have inconsistent types", while the REST path degrades mixed-type arrays to JSON encoding. The divergence between the two entry points is pre-existing (the SDK never supported mixed types), but worth a code comment.
- Unsigned integer kinds (`uint`, `uint8`, …) still fall through to the default error; also pre-existing.

zhuwenxing: Fixed in `36241e5140`. Typed empty slices now retain their statically known data oneof, including nested empty slices, so they remain compatible with servers that predate unset-array support. Only empty interface slices use an unset oneof. Added request- and conversion-level coverage for typed, generic, and nested empty slices. The targeted Go unit tests pass.

I left the mixed-type and unsigned-integer slice behavior unchanged because both predate this PR.

xiaocai2333:
Verified the fix end-to-end against the repro in #51617: pymilvus sends `TemplateValue{array_val: TemplateArrayValue{}}` with the data oneof unset → the new `case nil` in `convertArrayValue` yields an empty `planpb.Array` → `checkContainsElement` and `FillJSONContainsExpressionValue` both iterate zero elements → segcore takes the `vals_.empty()` early-return in `JsonContainsExpr.cpp:138` and returns an all-true bitmap for `ContainsAll`. Since the template path converges on exactly the same C++ code as the inline form, "template == inline" holds structurally, not just by test coincidence. Nice.

Two follow-up test cases worth adding later (no implementation change needed — just coverage):

**1. Non-PK scalar field with an empty template list.** The Python test uses `id in {values}`, and `id` is the primary key, so it takes `exec_path_ == ExprExecPath::PkIndex` → `ExecPkTermImpl` (`TermExpr.cpp:240`). An ordinary INT64/VARCHAR field instead goes through `ExecVisitorImplForData` / `ExecVisitorImplForIndex` (`TermExpr.cpp:927`), and neither of those has the `vals_.size() == 0` early-return that the JSON/ARRAY branches have. I traced both by hand and they do produce an all-false bitmap (empty `SimdBatchElement`/`FlatVectorElement` → every lookup misses; the index path calls `TermIndexFunc(index, 0, ...)` → zero iterations), and inline `Int64Field in []` already exercises the same path — so this is low risk, but it is currently reasoning rather than verification. Adding a non-PK scalar field to the existing `query_cases` list would cover it cheaply, ideally one with a scalar index so the `ForIndex` branch is hit too.

**2. Nullable ARRAY field with `array_contains_all`/`array_contains_any` + empty template.** `JsonContainsExpr.cpp:147` has a dedicated `data_type == ARRAY && nullable_` branch that ANDs the result with the validity bitmap, so null rows do *not* match — semantics differ from the non-nullable all-true case. The `tags` field in the new test is not nullable, so that branch is untested here. Since the empty-list `array_contains_all` on an ARRAY field is precisely what #51617 is about, the nullable variant seems worth one more parametrize entry.

Unrelated to this PR, but noticed while tracing: `PhyTermFilterExpr::CanSkipSegment<T>()` (`TermExpr.cpp:175-181`) declares `T min, max;` and passes them to `CanSkipBinaryRange<T>` without initialization when `vals_` is empty — UB on numeric types. It is only reachable on the PK path, and inline `id in []` already triggers it today, so it predates this PR; the outcome happens to be correct either way (both branches end up all-false, which is what `in []` should return). Might be worth a separate issue with a `if (expr_->vals_.empty()) return false;` guard.

PR #51623, opened 2026-07-20
enhance: add request-level read leases for sealed segment reopen

congqixia:
Related to #51068

- retain sealed read leases throughout the SearchResult lifecycle
- drain active readers before publishing reopened segment snapshots
- keep growing reads ungated and validate schemas after lease acquisition
- reuse existing leases during reduce, export, and output-field fill
- add concurrency tests, regression coverage, and design documentation


internal/core/src/segcore/segment_c.cpp
```diff
@@ -488,6 +521,7 @@ AsyncSearch(CTraceContext c_trace,
                                                 enable_expr_cache,
                                                 span);
             }
+            search_result->read_lease_ = std::move(read_lease);
```
liliu-z, line 524: Holding each segment lease in its SearchResult creates a cross-segment hold-and-wait. searchSegments fans out in parallel and keeps completed results until errgroup.Wait(): R1 can hold S1 while blocked by the S2 writer, while R2 holds S2 and is blocked by the S1 writer; both writers wait for those held leases. Please add deadlock-free request-level admission or release-and-retry before keeping per-segment leases in SearchResult.

internal/core/src/segcore/SegmentReadLease.h
```diff
@@ -0,0 +1,242 @@
...
+
+    PublishLease
+    AcquirePublish(milvus::OpContext* op_ctx, int64_t segment_id) const {
+        std::unique_lock<std::mutex> lock(state_->mutex);
+        AssertInfo(!state_->writer_pending && !state_->writer_active,
+                   "concurrent publisher entered segment {} read gate",
+                   segment_id);
+        state_->writer_pending = true;
+
+        while (state_->active_readers != 0) {
```
liliu-z, line 226: op_ctx can be null here; production paths such as DropSealedSegmentIndex -> DropIndex do that. If a sealed SearchResult misses DeleteSearchResult, active_readers never reaches zero, this writer cannot be cancelled, and writer_pending blocks every later search on the segment. Please make this wait bounded or cancellable so one leaked result cannot permanently take the segment offline.

internal/core/src/segcore/SegmentReadLease.h
```diff
@@ -0,0 +1,242 @@
...
+                lock.unlock();
+                state_->cv.notify_all();
+                ThrowInfo(ErrorCode::FollyCancel,
+                          "publication drain cancelled for segment {}",
+                          segment_id);
+            }
+            state_->cv.wait_for(lock, std::chrono::milliseconds(10));
+        }
+
+        state_->writer_active = true;
```
liliu-z, line 254: Cancellation is only checked inside the drain loop. A pre-cancelled publisher with zero readers skips the check, and cancellation racing the last reader release can be missed too, so publication still commits. Please re-check cancellation after the drain and cover both cases.

internal/core/src/segcore/segment_c.cpp
```diff
@@ -562,6 +596,8 @@ AsyncRetrieve(CTraceContext c_trace,
 
             milvus::OpContext op_ctx(cancel_token);
             segment->LazyCheckSchema(plan->schema_, &op_ctx);
+            auto read_lease = AcquireSegmentReadLease(segment, cancel_token);
```
liliu-z, line 599: This gate also fails fast for AsyncRetrieve and AsyncRetrieveByOffsets, but only search retries gate-busy. Unary Query can return 2037 during publication, and QueryStream may emit partial segment results before the error. Please keep scoped retrieve admission blocking or retry each retrieve before sending results.

internal/core/src/segcore/segment_c.cpp
```diff
@@ -562,6 +611,9 @@ AsyncRetrieve(CTraceContext c_trace,
 
             milvus::OpContext op_ctx(cancel_token);
             segment->LazyCheckSchema(plan->schema_, &op_ctx);
+            auto read_lease =
+                AcquireScopedSegmentReadLease(segment, cancel_token);
```
liliu-z: This blocking wait runs on the Search CPU executor. A completed Search still owns its lease until its consume callback runs on the same pool. Queued Retrieves can fill the pool here, preventing those callbacks from releasing leases, so publication only escapes via the 30s timeout. Please use fail-fast per-retrieve retry or move this wait off the pool.

internal/core/src/segcore/SegmentReadLease.h
```diff
@@ -0,0 +1,288 @@
...
+            if (lock.owns_lock()) {
+                lock.unlock();
+            }
+            state_->cv.notify_all();
+        });
+
+        const auto deadline =
+            std::chrono::steady_clock::now() + publish_drain_timeout_;
+
+        while (state_->active_readers != 0) {
```
liliu-z, line 226: LazyCheckSchema reaches this wait from AsyncSearch/Retrieve on the Search CPU pool, before gate-busy retry can help. A completed SearchResult may still hold a lease while its consume callback is queued on that same pool; enough newer-schema requests can fill every worker with drain/reopen waiters, so the callback cannot run and publication only escapes via the 30s timeout. Please make lazy reopen fail fast or move it off this pool.

liliu-z (commented):
/lgtm
/approve

Issue #51351, opened 2026-07-14
[Enhancement]: Support composite atomic operations in the metadata catalog


tedxu:
Motivation

Several coordinator metadata mutations are logically atomic but are
persisted as multiple independent catalog calls, each its own etcd
transaction. With no composite primitive, callers hold an in-memory
lock across the calls and hand-order the writes so that a crash in
between is recoverable.

The canonical example is mix-compaction completion, in
completeMixCompactionMutation (internal/datacoord/meta.go). It issues
two separate AlterSegments calls, ordered "alter compactTo before
compactFrom segments to avoid data lost if service crash". The
ordering comment is the smell: the sequence is not atomic, so the
caller re-implements crash-safety by hand.

This pattern recurs across DataCoord, RootCoord, QueryCoord, and the
streaming catalogs -- at least a dozen sites, two of which already
carry a literal "should be a transaction" TODO in the code. The
ordering discipline is duplicated, easy to get wrong, and blocks
further refactoring of the catalog layer.

Where it shows up

A scan of the coordinator metastores groups these composites three
ways:

- Pure-metadata composites: two or more etcd catalog writes with no
  external I/O between them (e.g. compaction completion, channel drop,
  create/drop collection, clustering-compaction completion, stats and
  import task completion, QueryCoord collection+partition load and
  release).
- Catalog methods that are internally non-atomic: one method that
  saves and then separately removes, or batches across the txn limit
  (e.g. AlterSegments, AlterIndexes, streaming SaveVChannels).
- Two-phase flows with external I/O (object store / worker RPC)
  between catalog writes: these must keep their staged design and are
  out of scope.

Proposal

Introduce a composite catalog operation that commits a set of related
metadata changes as one unit -- atomically within the store's
transaction limits, and in a single well-defined crash-safe order
otherwise -- optionally guarded by preconditions (for example "create
only if it does not already exist") checked as part of the same
commit. The detailed design is tracked separately.

Benefits

- Fewer etcd operations and round trips: per-model calls collapse
  into a single transaction instead of two or more.
- Crash-safe ordering lives in one place, instead of being re-derived
  and re-commented at every call site.
- Atomic check-and-mutate via preconditions, removing read-then-write
  races (for example concurrent create-collection).
- Prepares for the planned catalog-service refactor: a batched
  mutation API maps cleanly to a single server-side transaction.

Initial targets

DataCoord mix-compaction completion and channel drop; RootCoord
[...]

PR #51624, opened 2026-07-20
enhance: add composite Update to coordinator metadata catalogs

tedxu:
See #51351

Coordinator metadata mutations that are logically one change were
persisted as several independent catalog calls, each its own etcd/TiKV
transaction. This adds a composite primitive so those collapse into a
single guarded transaction.

Update(actions ...UpdateAction) is added to the DataCoord, RootCoord and
QueryCoord catalogs, plus a txn.Builder-backed SaveRecoverySnapshot on
StreamingNode. A shared kv/txn.Builder accumulates ordered Save/Remove
ops plus commit-marker ops; txn.Commit applies them atomically in one
etcd transaction when the op set fits MaxEtcdTxnNum, otherwise via a
caller-ordered chunked fallback that writes the commit marker last, so a
crash leaves a recoverable partial state. Each action's (Type, Entry)
reuses the legacy per-model key/value encoding, so persisted keys and
values are byte-for-byte unchanged.

Two goals: fewer KV round-trips (a large win on TiKV), and free
atomicity whenever the mutation fits one etcd transaction (128 ops),
removing the hand-rolled crash-safety ordering callers used to maintain.

Migrated call sites (in-memory state applied only after the write):
- DataCoord: mix-compaction completion, channel drop, external
  collection refresh job drop and task creation, clustering-compaction
  partition-stats save, partition-stats + analyze-task cleanup.
- RootCoord: create-collection, drop-collection.
- QueryCoord: replica spawn + redundant release.
- StreamingNode: WAL recovery snapshot.

Notes:
- Compaction retires compactFrom inputs via AlterSegment, reproducing
  the legacy AlterSegments encoding byte-for-byte, including the
  handleDroppedSegment GC-compat write; channel drop keeps the
  record-only UpdateSegment encoding to avoid a per-segment prefix read
  on an unbounded batch.
- Foreign or unsupported (Entry, Type) combinations are rejected as a
  ServiceInternal programming error with no write.

Deferred as follow-ups (not in this PR):
- QueryCoord ResourceManager per-node assignment loops. Orthogonal to
  this primitive: SaveResourceGroup is already variadic, so the fix is a
  caller-side staging-overlay refactor with no interface change.
- RootCoord dropRole (DropRole + DropGrant). Blocked until the txn
  engine can emit exact and prefix removals in one transaction.

Testing: unit tests for all four catalogs (encoding equivalence,
ordering, partial failure, idempotent retry) and the migrated callers;
lint clean with the pinned golangci-lint.


internal/metastore/kv/txn/commit.go
```diff
@@ -0,0 +1,198 @@
...
+//
+// Otherwise Commit falls back to a caller-ordered, chunked flush: every
+// non-commit op is flushed first, in the order it was recorded. Consecutive
+// ops of the same kind are coalesced into a run and chunked into contiguous
+// slices of up to limit entries in recorded index order, so BOTH cross-kind
+// ordering (a Remove that must be visible before a later Save of the same
+// key) AND within-kind ordering across batches (an earlier put group must
+// persist before a later one, e.g. compactTo before compactFrom) are
+// preserved. Finally, the commit ops (CommitSave/CommitRemove) are applied
+// together as the last guarded txn; this final txn is the sole visibility
```
weiliu1031, line 46: [P1] Do not treat commit ops as a visibility barrier for the composite write

On the fallback path, non-commit chunks are written directly into the final keyspace before the commit marker. After a crash and restart, readers do not consult that marker: external refresh reloads every task record and enqueues active tasks without checking `job.TaskIds`, while a retry allocates fresh task IDs, so partially persisted tasks can run alongside their replacements. Streaming recovery similarly loads every segment and vchannel record with the old consume checkpoint and then replays the WAL, which can double-apply the boundary insert or schema change to a partially advanced record.

The final commit op orders writes, but it does not make the earlier keys invisible. Either stage records by generation and make readers resolve only the committed generation, or add domain-specific recovery that filters, reuses, or removes partial refresh tasks and prevents streaming recovery from combining newer entity state with an older checkpoint. If neither guarantee is available, marker-dependent operations must not enter the non-atomic fallback.

internal/metastore/streamingnode_catalog.go
```diff
@@ -38,4 +38,30 @@ type StreamingNodeCataLog interface {
...
+	// with the consume checkpoint always the last/commit-marker op, so a
+	// crash before it lands makes the whole snapshot invisible and the next
+	// retry re-persists everything (every part is an idempotent put on a
+	// deterministic key).
+	SaveRecoverySnapshot(ctx context.Context, pChannelName string, snapshot *WALRecoverySnapshot) error
+}
+
+// WALRecoverySnapshot is the compound payload of
+// StreamingNodeCataLog.SaveRecoverySnapshot.
+type WALRecoverySnapshot struct {
```
weiliu1031, line 59: [P2] WALRecoverySnapshot is a delta, not a replaceable snapshot

`SaveRecoverySnapshot` skips nil or empty sections and only changes entries present in the payload. The background caller supplies dirty segment and vchannel entries, while deletion requires explicit `FLUSHED` or `DROPPED` state; absent segment, vchannel, and schema keys remain unchanged. The initialization caller saves its current vchannels and checkpoint, but it does not prune keys left by an earlier markerless partial attempt. The API therefore cannot express “replace the persisted recovery state with this set,” especially when the replacement set is empty.

Rename this to `WALRecoveryUpdate`/`ApplyRecoveryUpdate` and document that missing entries mean unchanged and deletion requires a tombstone. Initialization should use a separate replace/prune operation or a committed generation switch. If this remains a snapshot API, it should represent a self-contained state, including empty-state replacement.

internal/metastore/update_action.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/milvus-io/milvus/pkg/v3/proto/querypb"
+)
+
+// ActionType classifies the intent of an UpdateAction.
+type ActionType int
+
+const (
+	// ActionAdd creates a new entry; the whole object is persisted.
+	ActionAdd ActionType = iota + 1
+	// ActionUpdate changes an existing entry. The change is expressed as a
```
weiliu1031: [P2] ActionUpdate conflates patching, replacement, upsert, and domain commands

The `ActionUpdate` documentation says the change is a mutation over the current value and that callers do not need to carry untouched fields. The implementations do not provide that behavior: `SegmentEntry` rewrites the complete persisted segment record derived from the supplied object, and `RefreshJobEntry` and `ReplicaEntry` marshal the supplied protobuf and overwrite the stored record. A caller following the documented partial-update contract would therefore lose omitted fields.

`batchSaveDropSegments` exposes the same ambiguity at the domain level: it performs a segment-drop operation through the generic `UpdateSegment` action without validating that the supplied segment is `Dropped`, while the channel in the same action list uses the explicit `MarkChannelDropped` command.

Either define and name these actions as full-value replace/upsert operations and correct the contract, or introduce separate Patch, Replace, and domain actions. The drop path should use an explicit `MarkSegmentDropped`/`SaveDroppedSegment` action that validates its state invariant.

internal/metastore/kv/streamingnode/update.go
```diff
@@ -0,0 +1,99 @@
...
+			b.Remove(key)
+			continue
+		}
+		data, err := proto.Marshal(info)
+		if err != nil {
+			return errors.Wrapf(err, "marshal segment %d at pchannel %s failed", info.GetSegmentId(), pChannelName)
+		}
+		b.Save(key, string(data))
+	}
+	for _, info := range snapshot.VChannels {
```
weiliu1031, line 74: [P2] Normalize vchannel operations before entering the fallback

This loop appends removals followed by saves for each vchannel, while `flushNonCommitOps` only coalesces consecutive operations of the same kind. If many vchannels contain both expired-schema removals and current-value saves, the operation stream repeatedly alternates between deletes and puts, producing at least two flush runs per vchannel, plus any additional chunking, and therefore O(N) etcd round trips.

Aggregate disjoint vchannel removals and saves into grouped phases before chunking while preserving required per-key ordering, or make the builder normalize same-kind operations when doing so cannot change ordering semantics.

weiliu1031 (changes requested):
This static review found one P1, three P2s, and one P3. The P1 is a protocol-level correctness issue in the chunked fallback: commit ops order writes but do not hide non-commit chunks, while the current readers do not consult the marker or a committed generation after restart. This can produce duplicate external refresh tasks and combine newer streaming entity state with an older consume checkpoint, so it should be fixed before merge.

The remaining findings cover mismatches between the RecoverySnapshot and ActionUpdate contracts and their actual behavior, request amplification in the streaming vchannel fallback, and unreachable legacy catalog APIs left behind by the migration.

Validation: tests were not run; the findings are based on static source and call-chain review of head 6c4db88a62.

internal/metastore/datacoord_catalog.go
```diff
@@ -73,6 +73,11 @@ type DataCoordCatalog interface {
 	AddSegment(ctx context.Context, segment *datapb.SegmentInfo) error
 	// TODO Remove this later, we should update flush segments info for each segment separately, so far we still need transaction
 	AlterSegments(ctx context.Context, newSegments []*datapb.SegmentInfo, binlogs ...BinlogsIncrement) error
+	// Update applies a composite set of UpdateActions as a single atomic
+	// write (or, when the op count exceeds the txn size limit, via a
+	// caller-ordered chunked fallback). See kv/txn.Commit for the exact
+	// atomicity contract.
+	Update(ctx context.Context, actions ...UpdateAction) error
```
weiliu1031, line 80: [P3] Remove legacy catalog APIs left unreachable by the migration

A production call-site scan found no consumers of the following methods outside private wrappers that are themselves no longer called:

- `DataCoordCatalog.MarkChannelDeleted`
- `DataCoordCatalog.DropAnalyzeTask`
- `DataCoordCatalog.SavePartitionStatsInfo`
- `DataCoordCatalog.DropPartitionStatsInfo`
- `DataCoordCatalog.SaveCurrentPartitionStatsVersion`
- `DataCoordCatalog.DropExternalCollectionRefreshJob`
- `DataCoordCatalog.DropExternalCollectionRefreshTask`
- `StreamingNodeCataLog.SaveVChannels`
- `StreamingNodeCataLog.SaveSegmentAssignments`
- `StreamingNodeCataLog.SaveSalvageCheckpoint`

Keeping both entry points expands the supported API surface, duplicates encoding paths, and allows future callers to bypass the composite-write semantics accidentally. Remove the unreachable interface methods, KV implementations, private wrappers, and generated mock methods. Rewrite legacy encoding-parity tests to exercise the shared encoders or expected KVs directly. If cleanup is intentionally deferred, mark these methods as deprecated and document the remaining migration plan.

internal/metastore/kv/txn/commit.go
tedxu: Thanks. The chunked fallback is not introduced by this PR: before
it, these mutations were persisted as several independent, non-atomic
etcd txns, so the same partial-crash window (and readers not
consulting the marker) already existed. This PR strictly narrows it -
the write is atomic whenever it fits one etcd txn (MaxEtcdTxnNum,
default 64), and even the fallback keeps the commit-marker last, so it
never regresses versus today.

The external-refresh reader gap is real on its own merit, though:
reloadFromKV loads every task by state without checking job.TaskIds,
so a partial write can leave orphan tasks that dispatch alongside the
fresh-ID retry. I will close that by reconciling loaded tasks against
the owning job TaskIds on reload and GC-ing orphans. The streaming
half is bounded by the per-record CheckpointTimeTick guards plus the
checkpoint-last ordering; the durable fix is the CAS commit already
noted in the recovery TODO, tracked as a follow-up.

internal/metastore/update_action.go
tedxu: ActionUpdate is a placeholder for a future partial-update (mutator)
API - see the SegmentEntry note on the deferred
func(*datapb.SegmentInfo) mutator that would be applied to a clone at
the apply site. Today every caller holds the full authoritative object
under the meta lock and supplies it whole, so the implementations are
full-value replace/upsert. I will fix the doc to describe the current
replace semantics and mark the partial-update contract as the future
direction, so the comment no longer over-promises. The drop path will
also get an explicit state check.

Issue #51568, opened 2026-07-17
[Bug]: OR-to-IN optimizer rewrite with mixed-type JSON predicates causes assertion failure and incorrect results


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: v2.6.20 (also reproduced on v2.6.19)
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus latest
- OS(Ubuntu or CentOS): macOS Darwin 25.3.0 (Apple Silicon) / Docker
- CPU/Memory: Apple M-series / 16GB
- GPU: N/A
- Others: Docker Compose deployment, tested with FLAT and HNSW index
```

### Current Behavior

The query optimizer merges multiple `field == value` branches of an `or` expression on the same JSON field into an internal `in` list for efficiency. When the branch values have **mixed types**, the merged list inherits inconsistent types, causing either a **segcore assertion crash** (code=2000) or **silent data loss** (wrong results, no error).

Each individual `or` branch is a valid, single-type expression that works correctly in isolation. The bug is introduced solely by the optimizer's branch-merging logic — hand-written `in [1.0, 2, 3, 4]` returns correct results; the optimizer-merged equivalent does not.

**Three sub-patterns share this root cause:**

| Sub-pattern | Type combination | Merge threshold | Behavior |
|-------------|-----------------|----------------|----------|
| **A. float + int** | `1.0` mixed with `2, 3, ...` | 4 branches | Crash (even count) or silent data loss (odd count) |
| **B. str/bool/... + int** | `"1"` or `true` mixed with `2, 3` | 3 branches | Always crash |
| **C. float + int (large)** | 1 float + N ints, N >= 4 | 5+ branches | Silent data loss scaling to 99%+ |

---

#### Sub-pattern A: float + int `$or` (4+ branches) — crash and silent data loss

**Crash (4 branches, float first):**
```python
col.query(expr='(meta["p"] == 1.0) or (meta["p"] == 2) or (meta["p"] == 3) or (meta["p"] == 4)')
# -> MilvusException code=2000, segcore assertion failure at Utils.h:228
```

**Silent data loss (5 branches, float first):**
```python
[...]

Issue #51567, opened 2026-07-17
[Bug]: Mixed bool and non-bool values in JSON IN/OR filters cause QueryNode crash or silent data loss


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
Milvus version: v2.6.20
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus latest
- OS(Ubuntu or CentOS): macOS Darwin 25.3.0 (Apple Silicon) / Docker
- CPU/Memory: Apple M-series / 16GB
- GPU: N/A
- Others: Docker Compose deployment, FLAT index
```

### Current Behavior

When a JSON field filter expression mixes **bool** values with **any other type** (int, float, or string), both `$in` and `$or` paths produce **crashes** or **silent data loss** depending on the specific values and their order.

This manifests through two paths:

1. **Direct `$in` path:** `meta["b"] in [true, 1]` — mixed-type list reaches segcore directly
2. **`$or` optimizer merge path:** `(meta["b"] == true) or (meta["b"] == 1) or (meta["b"] == 0)` — optimizer merges 3+ branches into a mixed-type internal `in` list

Both paths trigger the same root cause: the segcore C++ executor determines the expected value type from the first element in the `in` list and either asserts (crash) or silently ignores mismatched values (data loss).

**Unique bool characteristics:**
- **`true` (internal value 1) crashes; `false` (internal value 0) causes silent data loss** — an asymmetry caused by the zero-value code path
- **`$or` merge threshold is 3 branches** for bool (same as str+int), lower than float+int's 4 branches
- **Value ordering changes results:** same set of values in different order produces different result sets

---

#### Path 1: Bool mixed-type `$in`

```python
# Data: d0,d1: b=True(bool)  d2,d3: b=False(bool)  d4: b=0(int)  d5: b=1(int)

# CRASH -- true + int
col.query(expr='meta["b"] in [true, 1]')
# -> segcore assertion crash at Utils.h:228

# SILENT DATA LOSS -- false + int
col.query(expr='meta["b"] in [false, 1]')
# Expected: d2,d3,d5 (false=d2,d3 + int1=d5)
# Actual:   d2,d3 -- int match d5 silently dropped!

# ORDER MATTERS
col.query(expr='meta["b"] in [true, false, 0, 1]')
[...]

Issue #51489, opened 2026-07-16
[Bug]: Mixed-type values in JSON field 'in' operator triggers segcore assertion failure


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
-Milvus version: v2.6.20 (also reproduced on v2.6.19)
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus latest
- OS(Ubuntu or CentOS): macOS Darwin 25.3.0 (Apple Silicon) / Docker
- CPU/Memory: Apple M-series / 16GB
- GPU: N/A
- Others: Docker Compose deployment, tested with FLAT and HNSW index
```

### Current Behavior

When using in operator on a JSON field with a value list containing mixed types (e.g. int + string, int + bool), the QueryNode's segcore layer triggers a C++ assertion failure and returns error code 2000 (segcore error).

The same mixed-type in on a schema-defined scalar field correctly returns a parse error (code=1100) at the planning stage. Only JSON fields crash.

Error message:

MilvusException: (code=2000, message=failed to search: failed to search/query 
delegator 1 for channel by-dev-rootcoord-dml_...: fail to search on QueryNode 1: 
worker(1) query failed: N6milvus21ExecOperatorExceptionE :
Operator::GetOutput failed for [Operator:PhyFilterBitsNode, plan node id: 25936] : 
Assert "(value_proto.val_case() == milvus::proto::plan::GenericValue::kStringVal)" 
at ../../../internal/core/src/exec/expression/Utils.h:243
Stack trace: : segcore error[segcoreCode=2001])

### Expected Behavior

Mixed-type in on a JSON field should return a clear error at the expression parsing/planning stage (code=1100), consistent with how schema-defined scalar fields handle it:

MilvusException: (code=1100, message=failed to create query plan: cannot parse 
expression: priority in [1, "2"]: value type mismatch)
It should NOT reach the segcore C++ execution layer and trigger an assertion failure.

### Steps To Reproduce

```markdown
from pymilvus import (
    Collection, CollectionSchema, DataType, FieldSchema,
    connections, utility
)
import time

connections.connect("default", host="localhost", port="19530")

# 1. Create collection with JSON field
if utility.has_collection("repro_mixed_in"):
    Collection("repro_mixed_in").drop()

fields = [
[...]

PR #51389, opened 2026-07-15
fix: preserve segcore error codes through the exec driver

xiaofan-luan:
issue: #51384
Independent fix (listed in #51385); mergeable anytime.

## What
`CALL_OPERATOR` in exec/Driver.cpp rethrew every operator exception as `ExecOperatorException`, discarding the SegcoreError code chosen at the throw site — every classified error raised during expression evaluation reached the CGO boundary as generic `UnexpectedError`.

`ExecOperatorException` now derives from `SegcoreError`; `CALL_OPERATOR` catches, in order, `folly::FutureCancellation` (mapping it to `ErrorCode::FollyCancel`), then `SegcoreError` ahead of `std::exception`, rethrowing with the original code plus the same operator context. Both CGO boundary translators — `FailureCStatus` dynamic_cast and the futures consume callback's `thenError<SegcoreError>` arm — already dispatch on the base class, so consumers are unchanged: a code is only ever copied from the throw site, never invented.

## Wire-visible effects (release note)
These are attribution/classification changes on the query & search path. Nothing that was *correct* becomes wrong, but the code that reaches the client changes for failures that were previously flattened to `UnexpectedError`:
- **Cancellation**: a canceled / timed-out query now surfaces as `FollyCancel` (C++ 2038 → `merr.ErrSegcoreFollyCancel`) instead of a generic internal error, so the scheduler / proxy treats it as a real cancellation. Previously it collapsed to `UnexpectedError`. (Timeouts are converted to a cancellation upstream via the promise interrupt handler, so they surface here as `FutureCancellation` too; `CALL_OPERATOR` has no separate `FutureException`/`FutureTimeout` arm.) **Not covered** (follow-up): cancellation raised during a *cold* cache-cell load — `CacheSlot` (milvus-common) throws a plain `std::runtime_error`, not a `FutureCancellation`/coded `SegcoreError` — still surfaces as `UnexpectedError`.
[...]

PR #51625, opened 2026-07-20
fix: handle mixed-type JSON term expressions

zhengbuqian:
issue: #51489
issue: #51567
issue: #51568

## What changed

- Canonicalize mixed-type JSON `IN` expressions into homogeneous predicates.
- Partition JSON values by their concrete type: bool, int64, float, string, and array.
- Keep int64 and float values in separate predicates to avoid precision loss around `2^53`.
- Make the following expression optimizations type-aware for JSON:
  - `== OR` to `IN`
  - `!= AND` to `NOT IN`
  - combinations involving `IN`, equality, inequality, and range predicates
- Lower JSON array-valued membership to equality branches because segcore does not support array-valued `TermExpr`.
- Apply JSON term canonicalization even when expression optimization is disabled.
- Allow scalar JSON `NOT(x == value)` to be rewritten as `x != value` while preserving UNKNOWN for null, missing paths, and type mismatches.
- Add a segcore validation guard that rejects heterogeneous JSON `TermExpr` values before selecting raw data, JSON stats, or JSON path index execution paths.

## Root cause

A JSON `TermExpr` could contain values with different protobuf value types. Segcore selected its executor from the first value, causing later values to be interpreted using the wrong type.

The optimizer could also combine different JSON literal types back into a single `TermExpr`. Additionally, converting mixed integer and floating-point values to float could lose int64 precision.

## Behavior

For example:

```text
json["key"] in [1, 2, "3"]
```

is normalized to:

```text
json["key"] in [1, 2] OR json["key"] == "3"
```

Mixed-type `NOT IN` remains a negation of the complete split membership expression, preserving three-valued logic:

```text
NOT(json["key"] in [1, 2] OR json["key"] == "3")
```

Non-JSON columns continue to use their statically determined schema type.

## Verification

- Go parser and rewriter tests:

```bash
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2/...
```

- Segcore mixed JSON Term validation test.
- Raw JSON filtering three-valued semantics tests.
- JSON path index precision and `!=` semantics tests.
- JSON stats/shredding null, missing-path, and large-int64 precision tests.

### DIFF
--- a/internal/core/src/exec/expression/TermExpr.cpp
+++ b/internal/core/src/exec/expression/TermExpr.cpp
@@ -1229,6 +1229,23 @@ PhyTermFilterExpr::ExecVisitorImplForData(EvalCtx& context) {
 
 void
 PhyTermFilterExpr::DetermineExecPath() {
+    // Validate before execution-path selection or asynchronous prefetch can
+    // consume values using the type selected from vals_[0].
+    if (expr_->column_.data_type_ == DataType::JSON &&
+        expr_->vals_.size() > 1) {
+        const auto expected_type = expr_->vals_[0].val_case();
+        for (size_t i = 1; i < expr_->vals_.size(); ++i) {
+            if (expr_->vals_[i].val_case() != expected_type) {
+                ThrowInfo(DataTypeInvalid,
+                          "TermExpr values must have the same type, value 0 "
+                          "has type {} but value {} has type {}",
+                          static_cast<int>(expected_type),
+                          i,
+                          static_cast<int>(expr_->vals_[i].val_case()));
+            }
+        }
+    }
+
     // PkIndex
     if (is_pk_field_) {
         exec_path_ = ExprExecPath::PkIndex;
--- a/internal/core/src/exec/expression/UnaryExpr.cpp
+++ b/internal/core/src/exec/expression/UnaryExpr.cpp
@@ -253,7 +253,9 @@ PhyUnaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                 }
             }
 
-            if (exec_path_ == ExprExecPath::ScalarIndex && !has_offset_input_) {
+            if (exec_path_ == ExprExecPath::ScalarIndex &&
+                val_type != proto::plan::GenericValue::ValCase::kArrayVal &&
+                !has_offset_input_) {
                 switch (val_type) {
                     case proto::plan::GenericValue::ValCase::kBoolVal:
                         result = ExecRangeVisitorImplForIndex<bool>();
@@ -2041,6 +2043,9 @@ PhyUnaryRangeFilterExpr::DetermineExecPath() {
         case DataType::JSON: {
             auto val_type = FromValCase(expr_->val_.val_case());
             switch (val_type) {
+                case DataType::ARRAY:
+                    can_use = false;
+                    break;
                 case DataType::STRING:
                 case DataType::VARCHAR:
                     can_use = SegmentExpr::CanUseIndexForOp<std::string>(
--- a/internal/core/src/exec/expression/UnaryExpr.cpp
+++ b/internal/core/src/exec/expression/UnaryExpr.cpp
@@ -292,7 +292,15 @@ PhyUnaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                         result = ExecRangeVisitorImplJson<bool>(context);
                         break;
                     case proto::plan::GenericValue::ValCase::kInt64Val:
-                        result = ExecRangeVisitorImplJson<int64_t>(context);
+                        if ((has_offset_input_ ||
+                             exec_path_ != ExprExecPath::JsonStats) &&
+                            !IsInt64SafeForJsonDoubleIndex(
+                                expr_->val_.int64_val())) {
+                            result =
+                                ExecRangeVisitorImplJsonPreciseNumeric(context);
+                        } else {
+                            result = ExecRangeVisitorImplJson<int64_t>(context);
+                        }
                         break;
                     case proto::plan::GenericValue::ValCase::kFloatVal:
                         result = ExecRangeVisitorImplJson<double>(context);
@@ -1996,6 +2004,16 @@ PhyUnaryRangeFilterExpr::DetermineExecPath() {
     // JsonStats: use JSON statistics to skip segments when possible.
     if (CanUseJsonStatsAtInit()) {
         exec_path_ = ExprExecPath::JsonStats;
+        return;
+    }
+
+    // JSON array literals are only executable from raw data. Decide this
+    // before the base selector pins a JSON index so a cold or failed index
+    // fetch cannot affect an otherwise executable array comparison.
+    if (expr_->column_.data_type_ == DataType::JSON &&
+        expr_->val_.val_case() ==
+            proto::plan::GenericValue::ValCase::kArrayVal) {
+        exec_path_ = ExprExecPath::RawData;
         return;
     }
 
--- a/internal/core/src/exec/expression/UnaryExpr.cpp
+++ b/internal/core/src/exec/expression/UnaryExpr.cpp
@@ -253,9 +253,7 @@ PhyUnaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                 }
             }
 
-            if (exec_path_ == ExprExecPath::ScalarIndex &&
-                val_type != proto::plan::GenericValue::ValCase::kArrayVal &&
-                !has_offset_input_) {
+            if (exec_path_ == ExprExecPath::ScalarIndex && !has_offset_input_) {
                 switch (val_type) {
                     case proto::plan::GenericValue::ValCase::kBoolVal:
                         result = ExecRangeVisitorImplForIndex<bool>();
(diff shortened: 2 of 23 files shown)


internal/parser/planparserv2/rewriter/json_term.go
```diff
@@ -0,0 +1,153 @@
...
+
+	parts := make([]*planpb.Expr, 0, len(buckets))
+	for _, kind := range jsonTermKindOrder {
+		values := buckets[kind]
+		if len(values) == 0 {
+			continue
+		}
+		if kind == "array" {
+			for _, value := range values {
+				parts = append(parts, newUnaryRangeExpr(
```
liliu-z, line 83: Lowering JSON array membership to `col == [array]` makes it executable on the RawData path (raw `==` handles arrays via `CompareTwoJsonArray`), but the same predicate still hard-errors when the field has a `JsonFlatIndex` covering the path: `DetermineExecPath` commits to ScalarIndex without inspecting the value type, UnaryExpr's JSON refine switch treats `ARRAY` (`FromValCase(kArrayVal)`) as index-usable via `default: can_use = true` (UnaryExpr.cpp:2049-2050) so it is never downgraded to RawData, and the ScalarIndex eval branch has no `kArrayVal` case and throws `DataTypeInvalid` (UnaryExpr.cpp:283-285). So `json["k"] in [[...],[...]]` returns results on an unindexed segment but fails the entire query once a JSON flat index exists — a segment-state-dependent inconsistency in the PR's new array-IN support (pre-PR it threw uniformly via TermExpr.cpp:139, so this is an incomplete fix rather than a fresh uniform regression). Fix by returning `can_use = false` for `kArrayVal` in the JSON refine switch so array equality always uses RawData.

internal/parser/planparserv2/rewriter/term_in.go
```diff
@@ -327,7 +343,11 @@ func (v *visitor) combineAndInWithRange(parts []*planpb.Expr) []*planpb.Expr {
 			continue
 		}
 		if ue := e.GetUnaryRangeExpr(); ue != nil && isRange(ue.GetOp()) && ue.GetValue() != nil && ue.GetColumnInfo() != nil {
-			k := columnKey(ue.GetColumnInfo())
+			k, ok := valueGroupKey(ue.GetColumnInfo(), ue.GetValue())
```
liliu-z, line 398: For a JSON column, combineAndInWithRange folds an IN against a range via filterValuesByRange(effectiveDataType(g.col), ...) at line 402, but effectiveDataType returns DataType_JSON, which cmpGeneric does not handle and returns 0 for (range.go:427), so with inclusive bounds satisfiesLower/satisfiesUpper treat every term value as in-range (0>=0 / 0<=0) and the range clause is silently dropped — JSONField["v"] in [1,2] and JSONField["v"] >= 2 rewrites to IN [1,2] and matches v==1 (verified by running ParseExpr under default config; exclusive bounds happen to be safe because the empty-filter result hits the canFoldBoolDomainToConstant(JSON)==false guard). The faulty fold at line 402 predates this PR, but line 346's new kind-partitioned grouping routes JSON term+range into it and type-aware JSON range handling is a stated PR goal; fix by comparing with the resolved literal kind or skipping the fold when the effective type is JSON.

internal/parser/planparserv2/rewriter/json_term.go
```diff
@@ -0,0 +1,153 @@
...
+					term.GetColumnInfo(), planpb.OpType_Equal, value))
+			}
+			continue
+		}
+		if len(values) == 1 {
```
liliu-z, line 88: json_term.go:88 lowers a single-value int64 bucket of a mixed JSON IN to a unary Equal (newUnaryRangeExpr(..., OpType_Equal, values[0])), and the unary JSON compare macro's double branch (UnaryExpr.cpp:866-870, used by Equal at :1007) evaluates value == val where a stored double is compared against the int64 literal promoted to double, losing precision at 2^53 — so {"v":9007199254740992.0} falsely matches == 9007199254740993. The term executor (TermExpr.cpp:842-848, floor + int64) and the JSON-stats path (CompareInt64ToDouble, JsonNumberComparison.h) both compare exactly, so the same value kept in a multi-value int64 bucket (a TermExpr) does not match — the outcome depends on bucket size and execution path, undercutting the PR's stated 2^53-precision goal. The lossy double branch is pre-existing (it also affects a direct == bigint), so the fix belongs there: align it with the term path's floor+int64 conversion or CompareJsonNumberToBound, not merely with the lowering.

internal/core/src/exec/expression/TermExpr.cpp
```diff
@@ -1229,6 +1229,23 @@ PhyTermFilterExpr::ExecVisitorImplForData(EvalCtx& context) {
...
 void
 PhyTermFilterExpr::DetermineExecPath() {
+    // Validate before execution-path selection or asynchronous prefetch can
+    // consume values using the type selected from vals_[0].
+    if (expr_->column_.data_type_ == DataType::JSON &&
+        expr_->vals_.size() > 1) {
+        const auto expected_type = expr_->vals_[0].val_case();
+        for (size_t i = 1; i < expr_->vals_.size(); ++i) {
+            if (expr_->vals_[i].val_case() != expected_type) {
+                ThrowInfo(DataTypeInvalid,
```
liliu-z, line 1239: The new mixed-JSON guard ThrowInfo(DataTypeInvalid, ...) at TermExpr.cpp:1239 runs inside DetermineExecPath, which executes under std::call_once (Expr.h:2433); when it throws the once_flag stays unset, and cleanup via Driver::Close() calls op->WaitPrefetch() (Driver.cpp:244) — which re-invokes EnsureExecPathDetermined() and re-throws — before Task::RemoveDriver() (Driver.cpp:250). Because a Driver holds a shared_ptr<Task> (Driver.h:140) and ~Task() is empty (Task.h:75), skipping RemoveDriver never breaks the Task<->Driver cycle, leaking one task graph per rejected query (e.g. an old proxy sending mixed JSON terms during a rolling upgrade); this reproduces even under the default EnableDriverPrefetch=false, since the Eval-time throw leaves the once_flag unset for Close()'s WaitPrefetch to re-hit. The exception-unsafety of Close() (WaitPrefetch/op->Close before RemoveDriver) is the pre-existing root — transient prefetch failures such as index cold-fetch throttling or chunk I/O leak identically — so the durable fix is to validate the term types at plan build before the Task exists, or make Close() exception-safe.

internal/core/src/exec/expression/UnaryExpr.cpp
```diff
@@ -2041,6 +2043,9 @@ PhyUnaryRangeFilterExpr::DetermineExecPath() {
         case DataType::JSON: {
             auto val_type = FromValCase(expr_->val_.val_case());
             switch (val_type) {
+                case DataType::ARRAY:
```
liliu-z: At line 2046 `case DataType::ARRAY: can_use = false;` downgrades array-literal equality to RawData only after `SegmentExpr::DetermineExecPath()` has already run at line 2002, so the base selector may cold-fetch and pin the JSON flat index the query will never use — unlike `JsonContainsExpr`, which short-circuits arrays before the base call. The always-present cost is wasted pinning and cache pressure; additionally, if the cold fetch inside `EnsurePinnedIndex` throws (throttled or unavailable index) the array-equality query can fail even though it would ultimately select RawData. Short-circuit array literals to RawData before invoking the base selector.

xiaofan-luan:
The red `UT-CPP-Cov` on `bab18f0` (ciloop-18636) is the new `ExprJsonTermTest.MixedValueTypesReturnError` test — and it's failing on exec-driver error-code wrapping, **not** on the guard or the Task/Driver cycle fix (`DetermineExecPathFailureReleasesTaskDriverCycle` passes in the same run).

Root cause from the log:

```
=> TermExpr values must have the same type, value 0 has type 2 but value 1 has type 4
   at TermExpr.cpp:1239
unknown file: Failure
C++ exception with description "Operator::GetOutput failed for [PhyFilterBitsNode ...] => TermExpr values must have the same type ..."
```

The guard fires correctly, but `CALL_OPERATOR` in `exec/Driver.cpp` rethrows it as `ExecOperatorException`, which today is **not** a `SegcoreError` and drops the `DataTypeInvalid` code. So the test's `catch (const SegcoreError& error)` never catches it, the exception escapes the test body, and gtest reports an uncaught C++ exception → FAIL.

This is exactly what #51389 ("preserve segcore error codes through the exec driver") fixes: it makes `ExecOperatorException` derive from `SegcoreError` and has `CALL_OPERATOR` rethrow with the original code. Once #51389 lands:

- `catch (const SegcoreError&)` catches the wrapped exception and `get_error_code() == DataTypeInvalid` holds → `MixedValueTypesReturnError` goes green;
- the client also receives a clean `DataTypeInvalid` instead of a generic `UnexpectedError` / segcore `2000`, which is closer to the clean error code #51489 asks for.

Suggestion: land #51389 first (or rebase this PR on top of it) so the new test is green. If you'd rather merge this independently before #51389, the interim option is to assert on the wrapped exception (e.g. `EXPECT_ANY_THROW` + check `what()` contains `"must have the same type"`), then tighten it back to `SegcoreError` + `DataTypeInvalid` once #51389 merges.

PR #51632, opened 2026-07-20
fix: [cp3.0] correct multi-query top-k insertion in iterative filter

liliu-z:
Backport of #51591 to the 3.0 branch.

pr: #51591
issue: #51385

## What changed

Fix the per-query top-k insertion window in iterative filtered search. The old implementation mixed an absolute insertion position with a relative result count, so out-of-order candidates could overwrite existing results for query vectors after the first.

This is an exact cherry-pick of the master fix and includes the deterministic multi-query regression test.

## Verification

- Exact patch-id match with master commit f6bffd8de5eecd7410356d1fbe984e41a17514b8
- git diff --check
- clang-format 15 dry-run on all changed C++ files
- C++ regression test will run in PR CI

### DIFF
--- a/internal/core/src/exec/operator/IterativeFilterNode.cpp
+++ b/internal/core/src/exec/operator/IterativeFilterNode.cpp
@@ -84,47 +84,6 @@ PhyIterativeFilterNode::IsFinished() {
     return is_finished_;
 }
 
-inline void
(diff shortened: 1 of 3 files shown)


congqixia (approved):
/lgtm

PR #51633, opened 2026-07-20
fix: [cp3.0] prevent BulkIsValid fall-through for non-nullable columns

liliu-z:
Backport of #51598 to the 3.0 branch.

pr: #51598
issue: #51385

## What changed

Prevent non-nullable ChunkedColumn/ProxyChunkColumn BulkIsValid paths from falling through into nullable handling and invoking the callback twice per row.

This is an exact cherry-pick of the master fix and includes the regression coverage.

## Verification

- Exact patch-id match with master commit bf674af67b835bd031898fe18d6dfb601dff9a10
- git diff --check
- clang-format 15 dry-run on all changed C++ files
- CMake-generated compile command passed for ChunkedColumnGroupTest.cpp
- Upstream #51598 passed build-ut-cov and e2e-default CI
- Full local all_tests execution was not completed to avoid exhausting shared disk during concurrent cold C++ builds; 3.0 PR CI provides the monolithic runtime run

### DIFF
--- a/internal/core/src/mmap/ChunkedColumn.h
+++ b/internal/core/src/mmap/ChunkedColumn.h
@@ -175,6 +175,10 @@ class ChunkedColumnBase : public ChunkedColumnInterface {
                     fn(true, i);
                 }
             }
+            // Non-nullable is fully handled above; without this return the
+            // control falls through into the nullable block and invokes fn a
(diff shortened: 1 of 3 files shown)


congqixia (approved):
/lgtm

PR #51634, opened 2026-07-20
fix: [cp3.0] NULL group-by on an index-only nullable scalar field

liliu-z:
Backport of #51597 to the 3.0 branch.

pr: #51597
issue: #51385

## What changed

Fix NULL group-by on sealed nullable scalar fields when field data is not
loaded and values are read through the scalar index. A NULL reverse lookup
now returns std::nullopt instead of failing the whole query.

This is an exact backport of the master fix and includes the nullable
group-by regression test.

## Verification

- Exact stable patch-id match with master commit fd46aa32d1a4662903dfea5331ab41ef552ef3f4
- git diff --check
- clang-format-15 dry-run on changed C++ files
- make build-cpp-with-unittest
- make run-test-cpp filter=GroupBY.SealedIndexOnlyNullableGroupBy (1/1 passed)
- make run-test-cpp filter=GroupBY.* (7/7 passed)
- Independent AI review found no blocking issues

### DIFF
--- a/internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.h
+++ b/internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.h
@@ -282,7 +282,6 @@ class SealedDataGetter : public DataGetter<OutputType> {
                 return raw;
             }
         } else {
-            // null is not supported for indexed fields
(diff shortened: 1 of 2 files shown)


congqixia (approved):
/lgtm

PR #51636, opened 2026-07-20
feat: support Decimal field type for exact fixed-point numerics

### LINKED ISSUE #50956: feat: add Decimal/Numeric field type for exact fixed-point values
### Is your feature request related to a problem? Please describe.

Milvus currently offers `Float` and `Double` for real numbers, but no exact fixed-point numeric type. For domains like e-commerce, finance, billing and precise measurements, binary floating point introduces representation error: filters such as `price == 19.99` or aggregations/sorts over money can behave unexpectedly, and users are forced to work around it (store integer cents, or a `VarChar` and compare as strings). Both workarounds lose type safety and make range/equality filtering awkward.

### Describe the solution you'd like

Add a fixed-point **`Decimal` / `Numeric`** scalar field type.

- Schema: `DataType.Decimal` with `precision` (total digits) and `scale` (digits after the point), e.g. `Decimal(18, 4)`.
- Storage: fixed-width backing (e.g. int64/int128 unscaled value + scale metadata) rather than a string, so it is compact and index-friendly.
- Filtering: exact equality (`==`, `!=`, `IN`) and range (`<, <=, >, >=`) with correct decimal semantics (no float rounding).
- Arithmetic: `+ - * /` and `%` following decimal rules (scale propagation), consistent between compile-time constant folding and runtime execution.
- Index: support the scalar sort/inverted index paths used by other numeric types.
- SDK: expose as a decimal/string-safe value in pymilvus and other SDKs to avoid re-introducing float error at the client boundary.

### Describe alternatives you've considered

- Storing integer "minor units" (cents): works but pushes scaling logic into every client and breaks natural range queries.
- Storing as `VarChar`: loses numeric ordering and range filtering.
- `Double`: not exact; unacceptable for money/precise values.

### Additional context

Part of a broader scalar type-system expansion (companion requests: native `UUID` type, `Map<K,V>` storage). Prior art: PostgreSQL `numeric`, MySQL `DECIMAL`, Arrow `Decimal128/256` (Milvus already depends on Arrow, which has a decimal representation to build on).

AyushKashyapII:
## Summary

Closes: https://github.com/milvus-io/milvus/issues/50956

Adds a new scalar field type, `Decimal`, for fixed-point exact numeric values
(currency amounts, financial quantities) that don't suffer the binary
floating-point rounding error inherent to `FLOAT`/`DOUBLE`.

- Wire format: decimal values travel as UTF-8 literal text (e.g. `"19.99"`)
  over the existing `bytes_data` oneof slot — no proto wire-format change
  beyond the `Decimal = 30` enum value (already merged in
  [milvus-proto#630](https://github.com/milvus-io/milvus-proto/pull/630)).
- Internally, values are decoded once into a fixed-width unscaled `int64`
  (e.g. scale=4: `"19.99"` → `199900`), which reuses essentially all of
  `Int64`'s existing storage, comparison, and indexing machinery.
- Indexing: `STL_SORT`, wired into AUTOINDEX.
- Filter expressions: comparisons, ranges, `IN`, null checks, and `+`/`-`
  arithmetic (e.g. `price - 5 > 10`) are supported, with literals rescaled
  from exact source text (never through `float`) to avoid reintroducing the
  rounding error Decimal exists to eliminate.

## Verification status

- **Go**: compiled and test-verified natively (`go test -tags dynamic,test
  ./internal/parser/planparserv2/...`, plus storage/typeutil packages).
- **C++**: implemented (segcore storage, query execution, indexing
  dispatch), but **not yet build-verified locally** — the dev environment
  used for this PR is memory-constrained for a from-scratch
  `make build-cpp`. Opening as a **draft PR** to get real signal from CI
  rather than blocking on that locally. Will convert to ready-for-review
  once CI is green.

## Explicitly out of scope (follow-ups)

- `Multiply`/`Divide`/`Modulo` arithmetic on `Decimal` (needs fixed-point
  rescaling logic distinct from `Add`/`Sub`'s constant-rescale approach).
- `BITMAP`/`HYBRID` indexing (cardinality-based auto-selection).
- Field-vs-field arithmetic (e.g. `price - other_price`), including two
  `Decimal` columns with different scales.

## Links

- issue: #50956
- design doc: `docs/design-docs/design_docs/20260718-decimal_type.md`

## Test plan

- [x] `internal/parser/planparserv2` unit tests (comparisons, ranges, IN,
      null checks, Add/Sub arithmetic, Mul/Div/Mod rejection)
- [ ] C++ unit tests (blocked on build environment / CI)
- [ ] End-to-end insert/query against a live standalone instance


AyushKashyapII:
heads up: one commit here (parser_visitor.go reorder) fixes a UT failure that actually reproduces on clean master (56c3b79dcd) — TestExpr_BloomMatch_Errors/bloom_match_rejected_inside_MATCH_* broke when #51725  MATCH validation started rejecting all function calls before the bloom_match-specific guard could fire. I moved the hasBloomFilterExpr check ahead of validateMatchPredicateElementLevel so the specific error wins. happy to split it into its own PR if that's preferred!

AyushKashyapII:
hello @xiaofan-luan , i dont know if u have gotten time to look into it but it seems to have passed all the tests and if u could .so that i this PR could be resolved. thanks !
i will be ready to make the necessary changes

xiaofan-luan:
Thanks for following up. I reviewed the current head (`66371dd5f0666527241ba4ae665e0ae93e75fe2d`). The main CI jobs are green, but I do not think this is ready to merge yet. I found several blockers that are not covered by the current tests:

1. **The C++ enum value does not match the proto value.** `internal/core/src/common/Types.h` defines `DataType::DECIMAL = 27`, while `milvus-proto` defines `Mol = 27` and `Decimal = 30`. `FieldMeta::ParseFrom` directly casts `schema_proto.data_type()` to the internal enum, so a real Decimal schema becomes `DataType(30)` and none of the `case DataType::DECIMAL` branches are reachable.

2. **The wire representation conflicts with the proto contract.** The design and implementation treat `bytes_data` as UTF-8 text such as `"19.99"`, while `milvus-proto` documents Decimal as a big-endian unscaled integer (8 or 16 bytes per element). Please settle and document the intended wire contract first; clients and server code must use the same representation.

3. **Nullable Decimal rows fail during decoding.** Proxy null expansion places an empty byte slice at invalid positions, but both `ConcurrentVector.cpp` and `internal/storage/utils.go` decode every element unconditionally. A valid nullable insert therefore reaches `DecodeDecimalUnscaled("")` / `EncodeUnscaledInt64("")` and fails. The decode paths need to honor `valid_data`.

4. **Decimal binlogs cannot be read back.** A Decimal payload writer was added, but `PayloadReader.GetDataFromPayload` has no Decimal case, so reload/compaction/deserialization returns `unknown type` before the Decimal branch in `data_codec.go` can run.

5. **Query output and upsert helpers do not copy Decimal `BytesData`.** `PrepareResultFieldData`, `AppendFieldData`, `AppendFieldDataByColumn`, `getScalarDataLen`, and `getData` do not handle Decimal/`ScalarField_BytesData`. Query/search reduction can therefore return an empty Decimal output field, and upsert row merging can drop the Decimal column.

6. **`DECIMAL(p,s)` precision validation is incorrect.** The current check counts only the fractional digits present in the input, not the digits after padding to the field scale. For example, `DECIMAL(5,2)` incorrectly accepts `"12345"` (encoded as `1234500`), while `DECIMAL(3,3)` incorrectly rejects `"0.001"`. The invariant should include `integer_digits <= precision - scale`.

Please also add Decimal-specific coverage for C++ schema parsing, nullable growing/flush paths, flush-and-reload/compaction, sealed and growing query output, upsert, and filter templates. The existing green C++ and E2E suites do not exercise these paths.

xiaofan-luan:
One clarification on the wire format: **the user-facing SDK API may accept a string or a language-native Decimal value, but `ScalarField.bytes_data` should not carry the UTF-8 text itself.** The SDK should hide the binary representation from ordinary users.

For the current `precision <= 18` design, please use one canonical fixed-width representation on the wire: the signed unscaled value as exactly 8 bytes (`int64`, two's complement). Milvus uses little-endian for its internal scalar/vector binary buffers by default, so little-endian would be the natural choice here as well.

This must be treated as a public contract rather than an implementation detail:

- update `milvus-proto`, which currently documents Decimal as big-endian and says 8 or 16 bytes;
- keep Go, C++, and every SDK encoder/decoder consistent;
- use `valid_data` for null rows instead of trying to parse their placeholder bytes;
- add cross-language golden tests for positive, negative, zero, min/max precision, and nullable values.

In short: users can pass `"19.99"`; the SDK converts it using the schema scale, but the protobuf payload should contain the canonical binary unscaled `int64`, not the string `"19.99"`.

AyushKashyapII:
hey @xiaofan-luan ,long time, i have enacted on the suggestions u made and the obvious mistakes that were made in the previous build ( just to like get the complete picture took a little more time than expected ) ,when u get time would love to get ur review on it and will be ready to make the changes as required 
thanks a lot for ur time and help !!!

xiaofan-luan:
Thanks for addressing the previous feedback. Before continuing with implementation fixes, I think we should step back and complete the Decimal design first.

I would prefer the public `Decimal(p, s)` type to be more flexible and closer to PostgreSQL `NUMERIC`, rather than permanently coupling the logical type and its public wire contract to an 8-byte `int64` with `precision <= 18`. The initial implementation may still support only Decimal64, but the design and wire contract should leave a clear path for Decimal128/256 or higher precision without requiring a breaking change.

Please update the design document to cover:

- supported precision/scale ranges and whether negative scale or scale greater than precision is allowed;
- physical representations for Decimal64/128/256 or arbitrary precision;
- wire encoding and backward compatibility;
- rounding, overflow, and cross-scale arithmetic semantics;
- growing-segment memory layout;
- Storage V1/V2 binlogs and Storage V3 Arrow representation;
- flush, reload, mmap, compaction, and schema-evolution behavior;
- supported query operations and aggregations;
- supported index types and raw-data retrieval behavior;
- SDK, REST, and bulk-import responsibilities.

After the design and support matrix are agreed upon, please add end-to-end coverage for:

- insert and immediate growing-segment query/search;
- nullable/default values;
- literal and template filters;
- multi-segment search output;
- flush and sealed-segment reload;
- `STL_SORT` indexed retrieval;
- Storage V3 growing-source flush;
- compaction/restart;
- add-field schema evolution;
- precision, scale, rounding, and overflow boundaries.

The same logical Decimal values should round-trip identically through SDK input, wire encoding, growing storage, flush, sealed storage, indexing, compaction/restart, and SDK output.

AyushKashyapII:
hey @xiaofan-luan ,
Updated the design doc to cover the full support matrix you asked for (precision/scale bounds, Decimal64→128/256 growth path, rounding/overflow, storage/compaction/schema-evolution, query/index scope, SDK/REST responsibilities) plus worked examples of what's allowed. Also incorporated the new valid_data location (ScalarField/VectorField instead of FieldData) into the nullable-Decimal paths, and merged in latest master. Finishing full E2E test coverage now and will push shortly.
thanks for review again !! i will try to be more careful in adding better docs for the people keeping compatablity in mind.

AyushKashyapII:
the last push was just for the review of doc ?i have made some changes but the merge has not completely landed 
looking to fix the branch

AyushKashyapII:
hey @xiaofan-luan ,
Update: all tests are passing now (including the C++ segcore suite). the branch is merged up to current master and ready for another look whenever you have time. Also updated the design doc per your last review ,it's there for reference if useful.
thanks for ur time and will be happy to follow up as per suggestions!!

AyushKashyapII:
hey @xiaofan-luan @czs007 ,
Update: all tests are passing now (including the C++ segcore suite). the branch is merged up to current master and ready for another look whenever you have time. Also updated the design doc per your last review ,it's there for reference if useful.
thanks for ur time and will be happy to follow up as per suggestions!!

AyushKashyapII:
hey james ,their are multiple data types which have been included in proto with no implementation in the master branch like for decimal and date time and for now i think one thing that can be done is that until the mods find time to review we can add that the implementation is not complete and it is proto only in the doc preventing any confusion as talked in this issue #52647 
if this is ok then we can probably change the docs it wont be a really hard change to milvus-proto and as the branches get merged then the warning can be cleared form proto repo
cc: @xiaofan-luan

AyushKashyapII:
hey @xiaofan-luan , the doc needs to be reviewed and this time only if u could give me the green light then i m thinking abt correcting the build issues ,but it has been many cycles of correction and still the doc is up for review 
could u spare some time to take a look into it ?

Issue #51635, opened 2026-07-20
[Bug]: growing-source provider stays in handoff-only mode after release handoff clears


bigsheeper:
## Bug Report

A querynode growing-source provider can remain in provider-wide `handoffOnly` mode after a release manual-flush handoff has completed and `ClearReleasePrepared` is called for the released segment.

### Impact

On a still-active channel, later growing-source flush resolution for unrelated segments is rejected before the provider even checks `SegmentManager.GetGrowing`. For segments whose source decision is already sticky as `FlushSourceGrowing`, this can make subsequent flush attempts see `GrowingSourceUnavailable` and retry instead of serving the local growing source.

Severity: High. This is a normal release/manual-flush path and can block correctness/liveness of later growing-source flushes on the same channel.

### Root Cause

`PrepareGrowingSourceReleaseHandoff` enters provider-wide handoff-only mode by setting `handoffOnly = true` and adding allowed segment IDs. `ClearReleasePrepared` deletes the released segment from `releasePrepared`, `releaseAllowed`, and `handoffAllowed`, but it never disables `handoffOnly` once there are no retained or allowed handoff segments left.

After that, `acquireLease` rejects any segment that is neither in `handoffAllowed` nor `retained`.

### Reproduction

A minimal failing test was added locally before filing this issue:

```bash
GOWORK=off \
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/internal/core/output/lib64/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator -run TestDelegatorGrowingSourceProviderClearsHandoffOnlyAfterReleasePrepared
```

Failure evidence:

```text
expected: 2
actual  : 0
Test: TestDelegatorGrowingSourceProviderClearsHandoffOnlyAfterReleasePrepared
...
FAIL: GetGrowing(int64)
0 out of 1 expectation(s) were met.
```

[...]

PR #51637, opened 2026-07-20
fix: reset growing source handoff mode after release

bigsheeper:
issue: #51635

## What this PR changes

After a release manual-flush handoff clears the last prepared segment, the querynode growing-source provider now exits `handoffOnly` mode when there are no retained sources and no allowed handoff segment IDs left.

This prevents a still-active provider from rejecting later growing-source flush requests for unrelated segments before checking `SegmentManager.GetGrowing`.

## Verification

- Added `TestDelegatorGrowingSourceProviderClearsHandoffOnlyAfterReleasePrepared`.
- Repro before fix failed with `GrowingSourceUnavailable` and an unmet `GetGrowing` mock expectation.
- After fix:
  - `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator -run TestDelegatorGrowingSourceProviderClearsHandoffOnlyAfterReleasePrepared`
  - `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator`
  - `make static-check`

## Notes

The static-check worktree required regenerating ignored local `cmd/tools/migration/legacy/legacypb` output for analysis only; it is not part of this PR.

### DIFF
--- a/internal/querynodev2/delegator/delegator_data.go
+++ b/internal/querynodev2/delegator/delegator_data.go
@@ -1282,13 +1282,21 @@ func (sd *shardDelegator) ReleaseSegments(ctx context.Context, req *querypb.Rele
 		sd.growingSegmentLock.Unlock()
 	}
 
-	if releaseErr != nil {
-		return releaseErr
-	}
+	// Clear the release-handoff bookkeeping even when the worker release failed.
+	// RemoveDistributions and AddExcludedSegments above already ran
+	// unconditionally, so the delegator has committed to dropping these segments
+	// either way. ClearReleasePrepared is the growing-source counterpart of that
+	// cleanup and is the only place that removes IDs from handoffAllowed on an
+	// active provider: gating it on releaseErr would pin handoffOnly forever and
+	// keep rejecting every unrelated growing segment. A querycoord retry
+	// re-prepares, since prepareReleaseManualFlush filters on IsReleasePrepared.
 	if len(growing) > 0 && sd.growingSourceProvider != nil {
 		for _, entry := range growing {
 			sd.growingSourceProvider.ClearReleasePrepared(entry.SegmentID)
 		}
+	}
+	if releaseErr != nil {
+		return releaseErr
 	}
 	return nil
 }
--- a/internal/querynodev2/delegator/delegator_data_test.go
+++ b/internal/querynodev2/delegator/delegator_data_test.go
@@ -1870,6 +1870,95 @@ func (s *DelegatorDataSuite) TestReleaseGrowingSourceAfterPreparedHandoff() {
 	s.Require().NoError(err)
 }
 
+// TestReleaseGrowingSourceClearsHandoffOnWorkerReleaseFailure pins that a failed
+// worker release still clears the release-handoff bookkeeping. RemoveDistributions
+// and AddExcludedSegments already ran unconditionally by that point, so leaving
+// handoffAllowed populated would keep handoffOnly armed forever and reject every
+// unrelated growing segment -- the same symptom as issue #51635.
+func (s *DelegatorDataSuite) TestReleaseGrowingSourceClearsHandoffOnWorkerReleaseFailure() {
+	ctx := context.Background()
+	s.workerManager = &cluster.MockManager{}
+	s.manager = segments.NewManager()
+	s.loader = &segments.MockLoader{}
+	s.genTextCollection()
+	s.enableGrowingSourceFlush()
+
+	delegator, err := NewShardDelegator(
+		ctx,
+		s.collectionID,
+		s.replicaID,
+		s.vchannelName,
+		s.version,
+		s.workerManager,
+		s.manager,
+		s.loader,
+		10000,
+		nil,
+		s.chunkManager,
+		NewChannelQueryView(nil, nil, nil, initialTargetVersion),
+		nil)
+	s.Require().NoError(err)
+	sd := delegator.(*shardDelegator)
+	defer sd.Close()
+
+	const (
+		segmentID    = int64(1001)
+		partitionID  = int64(500)
+		targetOffset = int64(10)
+	)
+
+	segment := segments.NewMockSegment(s.T())
+	segment.EXPECT().ID().Return(segmentID).Maybe()
+	segment.EXPECT().Version().Return(int64(1)).Maybe()
+	segment.EXPECT().Shard().Return(s.channel).Maybe()
+	segment.EXPECT().Collection().Return(s.collectionID).Maybe()
+	segment.EXPECT().Partition().Return(partitionID).Maybe()
+	segment.EXPECT().Type().Return(segments.SegmentTypeGrowing).Maybe()
+	segment.EXPECT().Level().Return(datapb.SegmentLevel_Legacy).Maybe()
+	segment.EXPECT().PinIfNotReleased().Return(nil).Once()
+	segment.EXPECT().InsertCount().Return(targetOffset).Once()
+	segment.EXPECT().MemSize().Return(int64(1024)).Once()
+	segment.EXPECT().Unpin().Maybe()
+
+	s.manager.Segment.Put(ctx, segments.SegmentTypeGrowing, segment)
+	sd.distribution.AddGrowing(SegmentEntry{
+		NodeID:      paramtable.GetNodeID(),
+		SegmentID:   segmentID,
+		PartitionID: partitionID,
+		Version:     1,
+	})
+
+	releaseErr := merr.WrapErrServiceInternal("mock worker release failure")
+	worker := &cluster.MockWorker{}
+	worker.EXPECT().ReleaseSegments(mock.Anything, mock.AnythingOfType("*querypb.ReleaseSegmentsRequest")).Return(releaseErr).Once()
+	s.workerManager.EXPECT().GetWorker(mock.Anything, paramtable.GetNodeID()).Return(worker, nil).Once()
+
+	err = sd.growingSourceProvider.PrepareGrowingSourceReleaseHandoff(ctx, 10000, []syncmgr.GrowingSourceReleaseHandoffSegment{
+		{
+			SegmentID:    segmentID,
+			TargetOffset: targetOffset,
+		},
+	})
+	s.Require().NoError(err)
+
+	err = sd.ReleaseSegments(ctx, &querypb.ReleaseSegmentsRequest{
+		Base:         commonpbutil.NewMsgBase(),
+		NodeID:       paramtable.GetNodeID(),
+		CollectionID: s.collectionID,
+		SegmentIDs:   []int64{segmentID},
+		Scope:        querypb.DataScope_Streaming,
+		Shard:        s.vchannelName,
+	}, false)
+	// The release error must still surface to the caller.
+	s.Require().Error(err)
+
+	// ...but the handoff fence must not survive it.
+	sd.growingSourceProvider.mu.Lock()
+	handoffAllowed := len(sd.growingSourceProvider.handoffAllowed)
+	sd.growingSourceProvider.mu.Unlock()
+	s.Require().Zero(handoffAllowed, "handoffAllowed must be cleared even when the worker release fails")
+}
+
 func (s *DelegatorDataSuite) TestReleaseGrowingSourceAfterFencePreparedHandoff() {
 	ctx := context.Background()
 	s.workerManager = &cluster.MockManager{}
--- a/internal/querynodev2/delegator/growing_flush_source.go
+++ b/internal/querynodev2/delegator/growing_flush_source.go
@@ -399,6 +399,7 @@ func (p *delegatorGrowingSourceProvider) ClearReleasePrepared(segmentID int64) {
 	delete(p.releasePrepared, segmentID)
 	delete(p.releaseAllowed, segmentID)
 	delete(p.handoffAllowed, segmentID)
+	p.exitHandoffOnlyIfDrainedLocked()
 }
 
 func (p *delegatorGrowingSourceProvider) ReleasePreparedSegments() []int64 {
@@ -458,7 +459,14 @@ func (p *delegatorGrowingSourceProvider) tryReleaseRetainedLocked(segmentID int6
 		return nil, nil
 	}
 	delete(p.retained, segmentID)
+	p.exitHandoffOnlyIfDrainedLocked()
 	return p.unregisterIfInactiveLocked(), retained
+}
+
+func (p *delegatorGrowingSourceProvider) exitHandoffOnlyIfDrainedLocked() {
+	if p.handoffOnly && len(p.handoffAllowed) == 0 && len(p.retained) == 0 {
+		p.handoffOnly = false
+	}
 }
 
 func (p *delegatorGrowingSourceProvider) releaseRetained(registration *syncmgr.GrowingSourceRegistration, retained *retainedGrowingFlushSource) {
--- a/internal/querynodev2/delegator/growing_flush_source.go
+++ b/internal/querynodev2/delegator/growing_flush_source.go
@@ -51,6 +51,7 @@ type delegatorGrowingSourceProvider struct {
 	releasePrepared map[int64]int64
 	handoffOnly     bool
 	handoffAllowed  map[int64]struct{}
+	handoffInflight int
 }
 
 func newDelegatorGrowingSourceProvider(segmentManager segments.SegmentManager, waitFence func(context.Context, uint64) error, getTSafe ...func() uint64) *delegatorGrowingSourceProvider {
@@ -153,6 +154,11 @@ func (p *delegatorGrowingSourceProvider) PrepareGrowingSourceReleaseHandoff(ctx 
 		return p.prepareDeactivatedGrowingSourceReleaseHandoff(fenceTs, segments)
 	}
 	handoffSnapshot := p.enterHandoffOnly(segments)
+	// Keep the handoff-only fence armed until this in-flight handoff finishes.
+	// enterHandoffOnly releases p.mu before waitFence, so a concurrent
+	// ClearReleasePrepared could otherwise drain handoffAllowed and drop the
+	// fence while we are still parked in waitFence.
+	defer p.leaveHandoffOnlyInflight()
 	if p.waitFence != nil && fenceTs > 0 {
 		if err := p.waitFence(ctx, fenceTs); err != nil {
 			p.rollbackHandoffOnly(handoffSnapshot)
@@ -333,6 +339,7 @@ func (p *delegatorGrowingSourceProvider) enterHandoffOnly(segments []syncmgr.Gro
 	}
 
 	p.handoffOnly = true
+	p.handoffInflight++
 	for _, segment := range segments {
 		p.handoffAllowed[segment.SegmentID] = struct{}{}
 	}
@@ -464,9 +471,22 @@ func (p *delegatorGrowingSourceProvider) tryReleaseRetainedLocked(segmentID int6
 }
 
 func (p *delegatorGrowingSourceProvider) exitHandoffOnlyIfDrainedLocked() {
-	if p.handoffOnly && len(p.handoffAllowed) == 0 && len(p.retained) == 0 {
+	if p.handoffOnly && len(p.handoffAllowed) == 0 && len(p.retained) == 0 && p.handoffInflight == 0 {
 		p.handoffOnly = false
 	}
+}
+
+// leaveHandoffOnlyInflight marks one in-flight PrepareGrowingSourceReleaseHandoff
+// as finished. It re-checks the drain condition so that if the last allowed/
+// retained entry was cleared while this handoff was still in flight (drain was
+// blocked by handoffInflight), the fence is lowered now that it reaches zero.
+func (p *delegatorGrowingSourceProvider) leaveHandoffOnlyInflight() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.handoffInflight > 0 {
+		p.handoffInflight--
+	}
+	p.exitHandoffOnlyIfDrainedLocked()
 }
 
 func (p *delegatorGrowingSourceProvider) releaseRetained(registration *syncmgr.GrowingSourceRegistration, retained *retainedGrowingFlushSource) {
--- a/internal/querynodev2/delegator/growing_flush_source.go
+++ b/internal/querynodev2/delegator/growing_flush_source.go
@@ -50,7 +50,7 @@ type delegatorGrowingSourceProvider struct {
 	releaseAllowed  map[int64]uint64
 	releasePrepared map[int64]int64
 	handoffOnly     bool
-	handoffAllowed  map[int64]struct{}
+	handoffAllowed  map[int64]int
 	handoffInflight int
 }
 
@@ -62,7 +62,7 @@ func newDelegatorGrowingSourceProvider(segmentManager segments.SegmentManager, w
 		retained:        make(map[int64]*retainedGrowingFlushSource),
 		releaseAllowed:  make(map[int64]uint64),
 		releasePrepared: make(map[int64]int64),
-		handoffAllowed:  make(map[int64]struct{}),
+		handoffAllowed:  make(map[int64]int),
 	}
 	if len(getTSafe) > 0 {
 		provider.getTSafe = getTSafe[0]
@@ -358,11 +358,9 @@ func (p *delegatorGrowingSourceProvider) enterHandoffOnlyLocked(segments []syncm
 
 	p.handoffOnly = true
 	for _, segment := range segments {
-		if _, ok := p.handoffAllowed[segment.SegmentID]; ok {
-			// Already allowed by a concurrent handoff: not ours to undo.
-			continue
-		}
-		p.handoffAllowed[segment.SegmentID] = struct{}{}
+		// handoffAllowed is ref-counted: overlapping handoffs each hold one
+		// reference on the same segment, so a rollback gives back only its own.
+		p.handoffAllowed[segment.SegmentID]++
 		snapshot.added[segment.SegmentID] = struct{}{}
 	}
 	return snapshot
@@ -373,6 +371,10 @@ func (p *delegatorGrowingSourceProvider) rollbackHandoffOnly(snapshot handoffSna
 	defer p.mu.Unlock()
 
 	for segmentID := range snapshot.added {
+		if p.handoffAllowed[segmentID] > 1 {
+			p.handoffAllowed[segmentID]--
+			continue
+		}
 		delete(p.handoffAllowed, segmentID)
 	}
 	// Never clear handoffOnly directly. Another handoff may still be parked in
@@ -588,7 +590,7 @@ func (p *delegatorGrowingSourceProvider) Close() {
 	p.releaseAllowed = make(map[int64]uint64)
 	p.releasePrepared = make(map[int64]int64)
 	p.handoffOnly = false
-	p.handoffAllowed = make(map[int64]struct{})
+	p.handoffAllowed = make(map[int64]int)
 	registration := p.registration
 	p.registration = nil
 	p.deleteRetainedMetricsLocked()
@@ -608,7 +610,7 @@ func (p *delegatorGrowingSourceProvider) unregisterIfInactiveLocked() *syncmgr.G
 	p.releaseAllowed = make(map[int64]uint64)
 	p.releasePrepared = make(map[int64]int64)
 	p.handoffOnly = false
-	p.handoffAllowed = make(map[int64]struct{})
+	p.handoffAllowed = make(map[int64]int)
 	p.deleteRetainedMetricsLocked()
 	return registration
 }
--- a/internal/querynodev2/delegator/growing_flush_source.go
+++ b/internal/querynodev2/delegator/growing_flush_source.go
@@ -165,6 +165,23 @@ func (p *delegatorGrowingSourceProvider) PrepareGrowingSourceReleaseHandoff(ctx 
 			return err
 		}
 	}
+	// waitFence parks without holding p.mu, so Deactivate may have landed while we
+	// were there. Resuming into the active path would register a retained entry --
+	// and a segment pin -- on a deactivated provider, which nothing detaches and
+	// which keeps unregisterIfInactiveLocked from ever firing. Take the same route
+	// a handoff arriving after Deactivate takes.
+	if p.isDeactivated() {
+		p.rollbackHandoffOnly(handoffSnapshot)
+		return p.prepareDeactivatedGrowingSourceReleaseHandoff(fenceTs, segments)
+	}
+	// Drop segments retired while we were parked. ClearReleasePrepared wipes a
+	// segment's handoffAllowed entry to mean "this handoff is over, forget it";
+	// republishing releaseAllowed/releasePrepared -- or re-pinning via
+	// registerRetained -- would resurrect state nothing will clear again.
+	segments = p.stillFenced(segments)
+	if len(segments) == 0 {
+		return nil
+	}
 	snapshot := p.snapshotRetained(segments)
 	allowedSegments := make([]syncmgr.GrowingSourceReleaseHandoffSegment, 0, len(segments))
 	preparedSegments := make([]syncmgr.GrowingSourceReleaseHandoffSegment, 0, len(segments))
@@ -186,6 +203,21 @@ func (p *delegatorGrowingSourceProvider) PrepareGrowingSourceReleaseHandoff(ctx 
 	p.markReleaseAllowed(fenceTs, allowedSegments)
 	p.markReleasePrepared(fenceTs, preparedSegments)
 	return nil
+}
+
+// stillFenced keeps only the segments whose handoffAllowed reference this
+// provider still holds. A reference disappears when ClearReleasePrepared retires
+// the segment, which can happen while a handoff is parked in waitFence.
+func (p *delegatorGrowingSourceProvider) stillFenced(segments []syncmgr.GrowingSourceReleaseHandoffSegment) []syncmgr.GrowingSourceReleaseHandoffSegment {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	kept := make([]syncmgr.GrowingSourceReleaseHandoffSegment, 0, len(segments))
+	for _, segment := range segments {
+		if p.handoffAllowed[segment.SegmentID] > 0 {
+			kept = append(kept, segment)
+		}
+	}
+	return kept
 }
 
 func (p *delegatorGrowingSourceProvider) prepareDeactivatedGrowingSourceReleaseHandoff(fenceTs uint64, segments []syncmgr.GrowingSourceReleaseHandoffSegment) error {
@@ -218,6 +250,13 @@ func (p *delegatorGrowingSourceProvider) registerRetained(segmentID int64, targe
 	if p.closing {
 		p.mu.Unlock()
 		return errGrowingSourceProviderClosed
+	}
+	if p.deactivated {
+		p.mu.Unlock()
+		// Classified as channel-not-available on purpose: the provider is
+		// deactivated because the delegator is closing, and the unsubscribe path
+		// treats that as "prepare unavailable, carry on" rather than a failure.
+		return merr.WrapErrChannelNotAvailable(p.channelName, "growing source provider is deactivated")
 	}
 	if retained, ok := p.retained[segmentID]; ok {
 		if retained.targetOffset < targetOffset {
@@ -328,7 +367,10 @@ func (p *delegatorGrowingSourceProvider) rollbackRetained(snapshot map[int64]ret
 // lower handoffOnly without consulting handoffInflight -- while another handoff
 // is still parked in waitFence.
 type handoffSnapshot struct {
-	added map[int64]struct{}
+	// added counts the references this call took per segment, so a rollback
+	// gives back exactly as many as it took even if the caller passed the same
+	// segment twice.
+	added map[int64]int
 }
 
 // enterHandoffOnly arms the handoff-only fence without registering an in-flight
@@ -354,14 +396,14 @@ func (p *delegatorGrowingSourceProvider) enterHandoffOnlyInflight(segments []syn
 }
 
 func (p *delegatorGrowingSourceProvider) enterHandoffOnlyLocked(segments []syncmgr.GrowingSourceReleaseHandoffSegment) handoffSnapshot {
-	snapshot := handoffSnapshot{added: make(map[int64]struct{}, len(segments))}
+	snapshot := handoffSnapshot{added: make(map[int64]int, len(segments))}
 
 	p.handoffOnly = true
 	for _, segment := range segments {
 		// handoffAllowed is ref-counted: overlapping handoffs each hold one
 		// reference on the same segment, so a rollback gives back only its own.
 		p.handoffAllowed[segment.SegmentID]++
-		snapshot.added[segment.SegmentID] = struct{}{}
+		snapshot.added[segment.SegmentID]++
 	}
 	return snapshot
 }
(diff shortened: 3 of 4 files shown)


czs007:
## Overview

Adversarial verification surfaced two issues: one medium-severity concurrency defect in the handoff-only fence lifecycle, and one gap in test coverage for a newly added fence-exit path.

---

### Medium

**`internal/querynodev2/delegator/growing_flush_source.go:402` — `ClearReleasePrepared` can disarm the handoff-only fence while another handoff is still blocked in `waitFence`, and the fence is never re-armed**

**Problem.** The new exit hook in `ClearReleasePrepared` drops the channel-level handoff-only fence based solely on the *current* contents of `handoffAllowed`/`retained`, with no accounting of how many handoffs are in flight. Any `ClearReleasePrepared(S)` that removes the last `handoffAllowed` entry while another `PrepareGrowingSourceReleaseHandoff` is still parked in `waitFence` turns the fence off for the remainder of that handoff — and nothing re-arms it: `enterHandoffOnly` runs only *before* `waitFence` (`growing_flush_source.go:155-161`), and after the fence returns, `Prepare` touches only `releaseAllowed`/`releasePrepared` (`growing_flush_source.go:180-181`), never `handoffOnly`/`handoffAllowed`. From that point on, `acquireLease` (`growing_flush_source.go:494-500`) hands out growing flush sources for brand-new segments on a channel that is mid-release — precisely what `TestDelegatorGrowingSourceProviderHandoffOnlyRejectsNewSegmentsBeforeFence` asserts must not happen.

The reachable trigger is not "the same segment released twice" (`ReleaseSegments` does not do that). It is a cross-RPC interleaving on a single channel:

1. `UnsubDmChannel` → `prepareReleaseManualFlush` → `BeginGrowingSourceReleaseHandoff`/`Prepare` places `S` in `handoffAllowed` and blocks in `waitFence`, which for the delegator provider is `sd.waitTSafe` (`delegator.go:1576-1579`) — a real window, unbounded until tsafe advances.
2. Concurrently, a querycoord-issued `ReleaseSegments(DataScope_Streaming, S)` (created at `internal/querycoordv2/checkers/segment_checker.go:249` for redundant growing segments) reaches `shardDelegator.ReleaseSegments`; worker release succeeds and `delegator_data.go:1198-1201` calls `ClearReleasePrepared(S)` — after `sd.growingSegmentLock.Unlock()` at `:1192`, so it is not serialized against `Prepare`.
3. `handoffAllowed` becomes empty and `retained` is still empty (`Prepare` has not reached `registerRetained` yet, and `TargetOffset == 0` segments are never retained per `:167-169`), so `handoffOnly` flips to `false` mid-handoff.

Single-growing-segment channels are the common case; with several segments in the batch the fence survives until the last one is cleared. Note that the same shared-state weakness also lets `rollbackHandoffOnly` (`growing_flush_source.go:342-348`) restore `handoffAllowed` wholesale and re-arm the fence for an already-released segment, after which no call site can ever clear it again.

**Suggestion.** Introduce an in-flight-handoff refcount or generation token so that only the handoff that armed the fence may disarm it, and add a test that clears a segment while a second handoff is blocked in `waitFence`, asserting that an unrelated `GetGrowingFlushSource` remains `Unavailable`. *(raised by chyezh, tinswzy)*

---

### Low

**`internal/querynodev2/delegator/growing_flush_source_test.go:267` — `tryReleaseRetainedLocked` exit path is untested**

**Problem.** The new test exercises only the `ClearReleasePrepared` call site: it prepares `SegmentID 1001` with no `TargetOffset`, so `Prepare` hits `if segment.TargetOffset <= 0 { continue }` (`growing_flush_source.go:167-169`) and the segment is never retained — the exit therefore fires from `ClearReleasePrepared` with `len(p.retained) == 0` already true. The second new call site, `growing_flush_source.go:462` in `tryReleaseRetainedLocked` — where a retained source drains last and is what finally releases the fence — has no coverage here or elsewhere. `TestDelegatorGrowingSourceProviderReleasesWhenDetachedBeforeCommit` (test file `:336-357`) does drive `registerRetained` → `MarkReleaseDetached` → `releaseRetainedIfComplete`, but asserts nothing about the fence, so deleting line 462 leaves the suite green.

**Suggestion.** Add a case with `TargetOffset > 0` that registers a retained source, calls `ClearReleasePrepared` (fence must still hold), then drives `MarkReleaseDetached` + `releaseRetainedIfComplete` to completion and asserts that a subsequent `GetGrowingFlushSource` for an unrelated segment returns `GrowingSourceUsable`. *(raised by tinswzy)*

bigsheeper:
Good catch — confirmed reachable and fixed in cd38032d70.

`enterHandoffOnly` arms the fence and adds the segment to `handoffAllowed` before releasing `p.mu` for `waitFence`; a concurrent `ClearReleasePrepared` of that same in-flight segment (e.g. an independent release landing during a slow checkpoint) could then drain `handoffAllowed` and, via the new `exitHandoffOnlyIfDrainedLocked`, lower `handoffOnly` while the handoff was still parked in `waitFence`, letting a brand-new segment acquire a lease mid-handoff.

Traced the two call sides to confirm reachability: `ClearReleasePrepared` runs on the querynode `ReleaseSegments` path (`delegator_data.go`, after `growingSegmentLock` is released) while `PrepareGrowingSourceReleaseHandoff` runs on the writebuffer flush path (`write_buffer.go`) — different goroutines sharing only `p.mu`, which is dropped during `waitFence`, with no higher-level per-segment serialization.

Fix: track in-flight handoffs (`handoffInflight`, incremented in `enterHandoffOnly`, decremented via a `defer` covering all Prepare exits) and refuse to lower the fence until it reaches zero; the decrement re-checks the drain so the fence still lowers correctly once the last in-flight handoff finishes. Added `TestDelegatorGrowingSourceProviderKeepsHandoffOnlyWhileHandoffInflight` — a deterministic concurrent test (Clear fires while Prepare is parked in `waitFence`); verified it fails without the guard. This also closes the test-coverage gap you flagged.

czs007:
## Review Summary

Adversarial review found one critical resource-leak defect that prevents this PR from actually lowering the handoff fence on the production code path, plus four medium issues covering concurrent-rollback correctness, a missing regression test, and an unfixed fence-stuck path on release failure.

---

### 🔴 Critical

**1. `internal/querynodev2/delegator/growing_flush_source.go:342` — `BeginGrowingSourceReleaseHandoff` leaks `handoffInflight`, so the drain never fires in production**

*Problem:* `enterHandoffOnly` (:342) unconditionally does `handoffInflight++`, but the two callers are asymmetric. `PrepareGrowingSourceReleaseHandoff` (:156) balances it with `defer leaveHandoffOnlyInflight()` (:161); the closure returned by `BeginGrowingSourceReleaseHandoff` (:146) only calls `rollbackHandoffOnly` (:349–355), which restores `handoffOnly`/`handoffAllowed` but never touches `handoffInflight` — and on the committed path the closure is never invoked at all. Neither outcome balances the counter, and `Close` (:550–570) / `unregisterIfInactiveLocked` (:572–584) do not reset it either.

The only production release-handoff entry point, `releaseManualFlushPreparer.PrepareReleaseManualFlush`, goes through `Begin` at `release_manual_flush_preparer.go:47` (leaving `handoffInflight == 1` permanently after commit) and then through `Prepare` at :104 (self-balancing), so after a *successful* release the counter rests at 1 instead of 0. As a result the `handoffInflight == 0` condition in `exitHandoffOnlyIfDrainedLocked` (:474) is unsatisfiable on this path, `handoffOnly` stays `true` forever, and `acquireLease` (:514–520) keeps returning `GrowingSourceUnavailable` for every unrelated growing segment — effectively reintroducing bug #51635, now on the success path rather than only on failure.

*Reproduction:* Following the production order `Begin(1001)` → `Prepare(1001)` → `ClearReleasePrepared(1001)` yields `handoffOnly=true handoffInflight=1 handoffAllowed=0 retained=0`, and `acquireLease`/`GetGrowingFlushSource` reject unrelated segment 1002. The rollback path (invoking the closure returned by `Begin`) leaves `handoffInflight=1` as well.

*Suggestion:* `Begin` does not hold parked state, so it should not participate in `handoffInflight` at all — move the `++` into `Prepare`. Alternatively, give `Begin` a completion lifecycle that decrements on both the commit and the rollback side.

*(raised by tinswzy, chyezh)*

---

### 🟠 Medium

**2. `internal/querynodev2/delegator/growing_flush_source.go:353` — whole-snapshot rollback in `rollbackHandoffOnly` discards concurrent handoff state and bypasses the new inflight invariant**

*Problem:* `rollbackHandoffOnly` (:353–354) overwrites `p.handoffAllowed` and `p.handoffOnly` wholesale from the snapshot captured by this operation's `enterHandoffOnly`, never going through `exitHandoffOnlyIfDrainedLocked`. It can therefore set `handoffOnly = false` while `handoffInflight > 0` — precisely what this PR's new invariant (and its own test `TestDelegatorGrowingSourceProviderKeepsHandoffOnlyWhileHandoffInflight`) forbids. With two overlapping prepares — A enters first (snapshot `{enabled:false, allowed:{}}`), B then adds `segB` — a failure of A in `waitFence` (:163–165) clears `handoffAllowed` to `{}` and resets `handoffOnly` to `false`, dropping `segB` and lowering the fence while B is still parked; `acquireLease` (:514) then admits unrelated segments mid-handoff. Reachability: overlapping/retried `UnsubDmChannel` on the same vchannel (`services.go:385` → :723 → preparer); `services.go:380–381` only Insert/Remove a set, with no dedup or per-channel serialization, so a coord retry is enough to overlap.

*Reproduction:* With A (`fenceTs=100`, fails) and B (`fenceTs=200`, still parked): both parked gives `handoffOnly=true inflight=2 allowed=map[1001:{} 2001:{}]`; after A rolls back, `handoffOnly=false inflight=1 bAllowed=false` and `acquireLease(3003)` returns `true`.

*Note:* the function body itself predates this PR (untouched by the diff), but the newly introduced `handoffInflight` invariant makes it a live breach.

*Suggestion:* Switch to reference-counted / per-segment-contribution rollback, and route rollback through `exitHandoffOnlyIfDrainedLocked` so the invariant holds on every exit path.

*(raised by chyezh, tinswzy)*

---

**3. `internal/querynodev2/delegator/growing_flush_source.go:297` — `rollbackRetained` has no ownership/version check and can delete + unpin another operation's retained source (pre-existing)**

*Problem:* `rollbackRetained` (:297–322) restores or deletes retained entries purely from a pre-operation snapshot, with no ownership or version check: when `entry.existed == false` but the entry currently exists, :312–314 unconditionally deletes and unpins it. If two prepares both snapshot segment 1001 as absent before either calls `registerRetained`, A can create and commit the retained entry, B reuses it (`registerRetained` :222–229 hits the already-exists branch), and a later failure of B on a *different* segment (:176–183, e.g. the behind-target-offset error at :242) deletes and unpins A's retained source. A still reports release-prepared and returns success, but its committed handoff has no retained flush source once the active growing segment detaches, and eventually resolves as unavailable.

*Reproduction:* Driving the real functions in the real order (B `snapshotRetained` → A `registerRetained` + `markReleasePrepared` → B `rollbackRetained` with the stale snapshot) gives `after B rollback: retained(1001)=false, A still reports prepared=true`, with A's pinned segment unpinned.

*Note:* both the function and its call sites predate this PR (diff untouched) — pre-existing, same file. Reachability is identical to issue #2 (overlapping `UnsubDmChannel` on one vchannel, no serialization).

*Suggestion:* Serialize the prepare flow (which also fixes the snapshot-rollback problem above), or make retained rollback ownership-scoped / versioned.

*(raised by chyezh)*

---

**4. `internal/querynodev2/delegator/growing_flush_source_test.go:271` — new tests only cover the self-balancing `Prepare` path; no `Begin → commit → drain` regression test**

*Problem:* Both new tests (:271, :309) drive `PrepareGrowingSourceReleaseHandoff` directly — the one caller that *does* balance `handoffInflight` — so the real production `Begin` path is entirely uncovered. No test in the `delegator` package calls `provider.BeginGrowingSourceReleaseHandoff`; the only matches repo-wide are fake-provider implementations in `writebuffer/l0_write_buffer_test.go:117` and `streamingnode/server/service/handler_test.go:40`, neither of which touches this provider. This is why issue #1 can land with a green test suite — adding `Begin` to the sequence (`Begin` → `Prepare` → `ClearReleasePrepared`) turns the test red immediately (`handoffInflight=1`, `handoffOnly=true`).

*Suggestion:* Add a regression test that calls `provider.BeginGrowingSourceReleaseHandoff`, then `Prepare` + `ClearReleasePrepared`, and asserts `handoffOnly == false && handoffInflight == 0` (and that an unrelated segment can acquire a lease). Also add overlapping success/failure scenarios asserting that a successful operation's retained pin and allow-set survive a concurrent operation's rollback.

*(raised by tinswzy, chyezh)*

---

**5. `internal/querynodev2/delegator/delegator_data.go:1195` — `ClearReleasePrepared` is skipped when release fails, leaving `handoffAllowed` populated and the fence stuck (still present after fixing #1)**

*Problem:* The drain has only two trigger points: `ClearReleasePrepared` (:409) and `tryReleaseRetainedLocked` (:469). But `shardDelegator.ReleaseSegments` returns early at `delegator_data.go:1195–1197` when `releaseErr != nil`, never reaching the `ClearReleasePrepared` loop at :1198–1202. Since `ClearReleasePrepared` (:408) is the only place that deletes from `handoffAllowed` on an active provider, any `worker.ReleaseSegments` failure (worker lookup failure, RPC error — see :1178–1189) leaves those segment IDs in `handoffAllowed` permanently. The `len(handoffAllowed) == 0` condition in `exitHandoffOnlyIfDrainedLocked` (:473–477) then never holds, `handoffOnly` stays `true`, and `acquireLease` keeps rejecting all unrelated growing segments — the same symptom as #51635 even with #1 fixed. Recovery requires querycoord retrying with the exact same segment IDs against the same delegator, or a `Close`/`Deactivate`.

*Suggestion:* Move `ClearReleasePrepared` onto a path unaffected by `releaseErr` (a `defer`, or clear it on the failure branch too), and/or add a timeout / fence-ts fallback for `handoffAllowed`.

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/cd38032d7063a8377b23022a8149432bcfd118a3...czs007:team2-fix/T48f8dfb418?expand=1

bigsheeper:
Thanks — the Critical is confirmed, and it's the important one. Fixed in `fcd6c78562`, which also rebases the branch onto latest master (it was 47 commits behind). Per-finding below, including two I'm not fixing here and one thing I could not verify.

**#1 Critical — confirmed, fixed using your candidate patch.**

I enumerated every `handoffInflight` site in the repo: one `++` (`enterHandoffOnly:342`), one `--` (`leaveHandoffOnlyInflight:487`), whose sole caller is the `defer` in `Prepare`. `Begin`'s closure calls only `rollbackHandoffOnly`, which doesn't touch the counter, and neither `Close` nor `unregisterIfInactiveLocked` resets it — they reset `handoffOnly`/`handoffAllowed` and stop there. So the arithmetic has no escape: `Begin` → `Prepare` → `Clear` rests at 1 and the `handoffInflight == 0` arm is unsatisfiable. Your reading of the production path holds too — `PrepareReleaseManualFlush` sets `handoffCommitted = true`, so the rollback closure is dropped.

One more argument for your fix that isn't in the review: `write_buffer.go:482` guards the `Prepare` call with `if len(releaseSegments) > 0`, so there is a production path where `Begin` runs and `Prepare` never does. Pre-fix that leaked a count with nothing downstream able to return it. `Begin` genuinely must not participate in the counter.

I took the candidate patch as-is. The three-way split (`enterHandoffOnly` / `enterHandoffOnlyInflight` / `enterHandoffOnlyLocked`) is cleaner than what I'd have written.

**#4 — confirmed, and it's the reason #1 shipped green.** The patch's two tests are in. No delegator test called `Begin` before; the only repo-wide matches were fake providers in `writebuffer` and `streamingnode` tests that never touch this provider.

**#5 — confirmed, fixed.** `if releaseErr != nil { return }` sat ahead of the `ClearReleasePrepared` loop. The clear now runs on both paths and the error still propagates. What made this an easy call: `RemoveDistributions` and `AddExcludedSegments` already run unconditionally above it, so the delegator has committed to dropping those segments regardless — leaving `handoffAllowed` populated contradicts state the same function already changed. A querycoord retry re-prepares, since `prepareReleaseManualFlush` filters on `IsReleasePrepared`. Added a regression test with a failing mock worker.

**#2 and #3 — mechanisms confirmed, not fixed here.**

I verified both, including the reachability claim: `services.go:380-381` only `Insert`/`Remove` a `ConcurrentSet`, and the only reader of `Contain` is `services.go:219` in `WatchDmChannels` — so concurrent `UnsubDmChannel` on one vchannel is genuinely unserialized. `rollbackHandoffOnly` does overwrite wholesale without going through the drain check, and `rollbackRetained:308-315` does delete-and-unpin with no ownership check.

I'm leaving both out because their real fix is the same thing: serializing the prepare flow. Patching them individually inside a fence-lifecycle PR means touching the concurrency model of the whole handoff state machine with no test coverage for the overlapping-operation scenarios — which is how #1 happened. They deserve their own change with tests that actually drive two overlapping prepares.

**One flag on the fix for #1:** after it, nothing holds the fence across the `Begin` → `Prepare` gap (`handoffOnly = true`, `handoffInflight = 0`). A concurrent `ClearReleasePrepared` in that window would drain and lower the fence early. The pre-fix leaked count masked this by accident. In the normal flow both calls happen inside one `PrepareReleaseManualFlush` while `ClearReleasePrepared` is driven by a later `ReleaseSegments`, so I could not construct a reachable interleaving — but I did not prove it unreachable either. Flagging rather than claiming it safe.

**What I did not verify.** I could not build the C++ core in this environment, so **no test in this package was executed locally** — not the new ones, not the existing suite. I also could not do the check that matters most here: temporarily reverting each fix to confirm the new tests go red. Given that #4 was precisely "tests that don't exercise the production path," I'd rather say plainly that these three tests are unproven as regression guards than let CI green imply otherwise. If you want that evidence before merge, say so and I'll produce it.

czs007:
## Adversarial review summary

Two medium-severity concurrency defects in the growing-source release/handoff path, both triggered by overlapping `Prepare` calls on the same vchannel; one is pre-existing, the other undermines a guard introduced in this PR.

### Medium

**`internal/querynodev2/delegator/growing_flush_source.go:367` — `rollbackHandoffOnly` wipes the fence that `handoffInflight` is meant to hold**

This PR adds `handoffInflight` so the handoff-only fence stays armed while any `Prepare` is still parked in `waitFence` (see the comment at `growing_flush_source.go:157-161` and `TestDelegatorGrowingSourceProviderKeepsHandoffOnlyWhileHandoffInflight`), but the invariant is enforced only in `exitHandoffOnlyIfDrainedLocked` (`:491-495`). `rollbackHandoffOnly` (`:367-373`) unconditionally restores the whole snapshot — `p.handoffOnly = snapshot.enabled; p.handoffAllowed = snapshot.allowed` — without consulting `handoffInflight` or merging entries added concurrently, so the error path at `:163-166` bypasses the new guard. Failure: two `Prepare` calls overlap (A arms the fence with `allowed={S}` and blocks in `waitFence`, B extends it to `allowed={S,T}` with `handoffInflight=2` and also blocks) and A's `waitFence` returns an error on ctx cancel/timeout → A restores `handoffOnly=false` and replaces `handoffAllowed` with its stale empty map while B is still in flight, so `acquireLease` (`:526-541`) admits arbitrary unrelated growing segments — exactly the case the new test asserts must be rejected — and B's `T` is silently dropped from `handoffAllowed`. Suggested fix: have `rollbackHandoffOnly` undo only the entries this operation added and then re-evaluate `exitHandoffOnlyIfDrainedLocked` (which already checks `handoffInflight`), instead of restoring the snapshot wholesale. *(surfaced during verification)*

**`internal/querynodev2/delegator/growing_flush_source.go:297` — `rollbackRetained` rolls back from a stale snapshot and deletes retained sources published by a concurrent `Prepare`**

`PrepareGrowingSourceReleaseHandoff` takes `snapshotRetained(segments)` once before the per-segment loop (`:168`), but `registerRetained` acquires and releases `p.mu` per segment, so the snapshot goes stale. A concurrent `Prepare` can reuse an entry the first one just created (`:251-258` only raises `targetOffset` when the entry exists) and successfully `markReleasePrepared(S)`; if the first `Prepare` then fails on a later segment `T` — e.g. the recoverable `currentOffset < targetOffset` path at `:240-243` — its `rollbackRetained` deletes and unpins `retained[S]` based on its own outdated snapshot (`:312-315`). Failure: overlapping release-handoff preparations on one vchannel (e.g. a querycoord `UnsubDmChannel` retry) over segments `S` and `T`, where B publishes `S` and A later fails on `T` → `releasePrepared[S]` survives without a retained owner, so `DetachStreaming(S)` moves the segment into `growingOnReleasingSegments` without calling `release()`, `MarkReleaseDetached(S)` returns early because `p.retained[S]` is gone (`:442-446`), `ReleaseDetached` never fires, and the C++ growing segment leaks until querynode restart (`QueryNodeNumSegments` never decrements, `Exist(S)` stays true) while growing flush for `S` returns `GrowingSourceUnavailable`. The `existed=true` branch has the same shape and can overwrite a concurrently advanced `committedOffset`/`detached` with older values. Suggested fix: make retained reservations operation-owned or reference-counted so a failed operation only unwinds state it still owns, plus a regression test with overlapping successful and failing preparations. Note: `snapshotRetained`/`rollbackRetained` are byte-identical to master and untouched by this PR (which only changes `ReleaseSegments` cleanup ordering and `handoffInflight`), so this is a pre-existing defect — worth a separate issue/PR rather than blocking this one. *(raised by chyezh)*

bigsheeper:
Both confirmed. The first is fixed in `85c034a5b3`; the second I'm leaving where you put it.

**`rollbackHandoffOnly` bypassing the guard — confirmed, and your framing is sharper than mine was.**

I deferred this last round as "same root cause as the retained one, both want prepare serialisation, take them together elsewhere". That was the wrong call. The retained defect is pre-existing; this one is a hole in a guard *this PR introduces*, and the PR ships a test asserting the very invariant the hole breaks. Shipping a guard whose own test cannot see past it is worse than shipping neither.

Verified the mechanism end to end: `rollbackHandoffOnly` (`:367-373`) assigned `handoffOnly` and replaced `handoffAllowed` wholesale with no reference to `handoffInflight` and no call to `exitHandoffOnlyIfDrainedLocked` (`:491`), and `enterHandoffOnlyLocked` (`:351`) captured the full prior map before adding its own entries. So with A on `{S}` and B extending to `{S,T}`, A's `waitFence` failure at `:163-166` restored `handoffOnly=false` plus A's stale empty map — dropping T and, since `acquireLease` (`:526`) only consults the allow-list while `handoffOnly` is true, admitting arbitrary unrelated segments.

Your suggested fix is what I implemented, and it turned out smaller than the "concurrency model rework" I predicted last round. The snapshot now records only the entries this call actually inserted — skipping any a concurrent handoff had already allowed, since those are not ours to undo — and the rollback deletes just those before re-evaluating the drain check rather than writing `handoffOnly` itself. The ordering works out: on the `Prepare` path the drain check is a no-op at rollback time because the caller still holds its own in-flight count, and its deferred `leaveHandoffOnlyInflight` lowers the fence once the count hits zero; `Begin` holds no count, so its rollback lowers immediately when nothing else remains.

Added a regression test driving two overlapping prepares where the first fails in `waitFence`. It asserts the fence stays armed, B's segment survives, A's own entry is undone, and an unrelated segment still cannot get a lease. I checked it isn't vacuous by restoring the wholesale restore — it fails with `fence must stay armed while B is still in flight`.

**`rollbackRetained` — confirmed, not fixed here, agreeing with your own note.**

The stale-snapshot shape is real: `snapshotRetained` is taken once at `:168` while `registerRetained` releases `p.mu` per segment, so a concurrent prepare can publish an entry that the first prepare's rollback then deletes and unpins from its outdated view — leaving `releasePrepared[S]` without a retained owner, so `MarkReleaseDetached` returns early and the C++ growing segment leaks until restart. But the function is byte-identical to master and this PR does not touch it, and fixing it properly means making retained reservations operation-owned — a different change with its own tests.

Worth noting the two findings look alike but are not the same call: this PR created the invariant the first one breaks, and created none of the second one.

czs007:
Adversarial review surfaced one medium-severity concurrency issue in the new handoff rollback path; everything else in the diff checked out.

## Medium

**`internal/querynodev2/delegator/growing_flush_source.go:361` — `handoffAllowed` ownership is assigned to the first inserter, so one caller's rollback can revoke an entry another caller still depends on.**

`enterHandoffOnlyLocked` skips a segment already present in `handoffAllowed` and does not record it in `snapshot.added` (:361-364), so when two handoffs cover the same segment the entry belongs solely to the first inserter. If that first caller fails, `rollbackHandoffOnly` deletes the entry (:375-377) even though the second handoff is still in progress, and `exitHandoffOnlyIfDrainedLocked` (:503-506) then clears `handoffOnly` because both maps are empty and `handoffInflight == 0`. Nothing serializes the two paths: `release_manual_flush_preparer.go:47` → `GrowingSourceRegistry.BeginGrowingSourceReleaseHandoff` (`growing_source.go:216`, no per-channel lock) → `growing_flush_source.go:139`, and `UnsubDmChannel`'s `unsubscribingChannels` set (`services.go:380-381`) is only read by `WatchDmChannels` (`services.go:219`), so concurrent `UnsubDmChannel` calls on the same vchannel are not mutually exclusive. The second half compounds it: `Prepare` writes only `retained`/`releaseAllowed`/`releasePrepared` after `waitFence` (:168-188) and never restores `handoffAllowed`, so a deleted entry is never re-added.

*Failure:* two concurrent `UnsubDmChannel` calls on one vchannel resolve the same growing segment set `{S}` — A's `Begin` takes ownership (`added={S}`), B's `Begin` records nothing, then A fails at `wal.Append` / `CheckReleaseManualFlushNeed` / `GetGrowingFlushProgress` and its deferred `rollbackHandoff()` fires → A's rollback removes `handoffAllowed[S]` and tears down the fence while B's handoff is still running, so for the duration of B's `ManualFlush` append and progress query `acquireLease` returns true for arbitrary unrelated growing segments and `GetGrowingFlushSource` hands out a usable source — exactly what this PR's own `growing_flush_source_test.go:411` asserts must not happen; and if `retained` is non-empty the fence stays armed instead, in which case `GetGrowingFlushSource(S)` is rejected with `GrowingSourceUnavailable` until B's `registerRetained` publishes, reproducing the original symptom of #51635.

*Suggestion:* make `handoffAllowed` a per-segment refcount rather than a set, so an entry is removed only when the last handoff holding it rolls back or completes; alternatively serialize `Begin`/`Prepare` per channel. (raised by chyezh, xiaocai2333)

bigsheeper:
Confirmed, and fixed in `da166ec946` (the branch is also rebased onto latest master, so the commits you reviewed have new SHAs; their content is unchanged apart from this one). Your framing is right: the `handoffSnapshot` fix from last round solved the "rollback revokes someone else's entry" case and introduced its mirror image — the entry now belongs to whoever inserted it *first*.

**Verified, step by step, against the head you reviewed:**

- `enterHandoffOnlyLocked` (`growing_flush_source.go:359-368`) `continue`d on an existing entry and did not record it in `snapshot.added`, so B's `added` was empty for a segment A already held.
- `rollbackHandoffOnly` (`:375-377`) deleted A's entries, then called `exitHandoffOnlyIfDrainedLocked` (`:503-506`), whose four conditions — `handoffOnly`, empty `handoffAllowed`, empty `retained`, `handoffInflight == 0` — were all satisfiable.
- `Prepare`'s success path (`:167-188`) writes only `retained` / `releaseAllowed` / `releasePrepared` and never restores `handoffAllowed`, so the deletion was permanent.
- No serialization anywhere on that path: `BeginGrowingSourceReleaseHandoff` in the registry (`internal/flushcommon/syncmgr/growing_source.go:216`) holds `r.mu` only inside `getProviders`, and `unsubscribingChannels` (`internal/querynodev2/services.go:382-383`) is a `ConcurrentSet` whose only reader is `WatchDmChannels` at `services.go:219` — it does not exclude a second `UnsubDmChannel` on the same vchannel.

**Two things I'd add to your write-up, because they change how bad this is.**

*Overlap is the normal case, not a coincidence.* Both `UnsubDmChannel` calls derive `releaseSegmentIDs` from `node.localGrowingSegmentIDs(channel, nil)` (`services.go:387`), so racing calls on one vchannel compute the *same* list. A and B don't merely happen to share a segment — they share all of them.

*The severity splits by entry point.* Through `Prepare`/`Prepare` the second handoff holds `handoffInflight`, so the drain check is blocked and the fence stays armed — the damage is limited to `GetGrowingFlushSource(S)` returning `GrowingSourceUnavailable` until B's `registerRetained` publishes, i.e. the original #51635 symptom. Through `Begin`/`Begin` — the `UnsubDmChannel` path — neither side holds an in-flight count (deliberately, per the comment at `:334-338`), so A's rollback deletes the entry *and* lowers the fence, and `acquireLease` (`:543-550`) then admits arbitrary unrelated growing segments. That second variant is the one worth fixing.

**Fix: ref-count, as you suggested.** `handoffAllowed` is now `map[int64]int`; `enterHandoffOnlyLocked` increments unconditionally and always records the segment in `snapshot.added`, and `rollbackHandoffOnly` decrements, deleting only at zero. Each handoff holds exactly one reference and gives back exactly that one.

I left the three wholesale-clear sites as full deletes rather than decrements — `ClearReleasePrepared` (`:435-441`), `Close` (`:590-591`) and `unregisterIfInactiveLocked` (`:610-611`). They are segment-level and provider-level *terminal* cleanups ("this handoff is over, forget the segment"), not one holder releasing its share, so a decrement there would be wrong. Note this leaves the pre-existing hazard untouched: a `ClearReleasePrepared` landing mid-handoff still wipes the entry, which for the `Prepare` path is covered by the `handoffInflight` guard and for the `Begin` path is not. That is not introduced or worsened here, and I have not addressed it.

I did not take the serialization option: it needs a new per-channel lock ordered against `p.mu`, and it would not help anyway, since `Begin` (from `UnsubDmChannel`) and `Prepare` (from `write_buffer.go:483`, on the `ManualFlush` message) are *designed* to straddle the WAL round-trip and overlap.

**Test.** Added `TestDelegatorGrowingSourceProviderOverlappingBeginRollbackKeepsFence`. You were right that `TestDelegatorGrowingSourceProviderRollbackKeepsConcurrentHandoff` misses this — it misses it on two axes at once: its A and B use *disjoint* segments (1001 vs 2001) and both enter through `Prepare`, so the in-flight count masks the fence drop. The new case uses two `Begin`s over the same segment and asserts the fence stays armed, B's reference survives, an unrelated segment (3003) still cannot lease while the handed-off one can, and that the fence drops only after the last rollback.

Verified rather than assumed: reverting `enterHandoffOnlyLocked`/`rollbackHandoffOnly` to the first-inserter-owns semantics (keeping the `int` type so it still compiles) makes the new test fail at `growing_flush_source_test.go:439` — `fence must stay armed while B still holds the handoff`. Full `./internal/querynodev2/delegator/...` passes on the rebased head, and `make static-check` reports 0 issues across all four scopes.

czs007:
## Adversarial Review Summary

The core fix (ref-counted `handoffAllowed` plus the `handoffInflight` guard) is sound and mergeable; verification confirmed 2 medium and 2 low issues, all suitable as follow-ups or test hardening rather than merge blockers.

### Medium

- **`internal/querynodev2/delegator/growing_flush_source.go:168`** — `PrepareGrowingSourceReleaseHandoff` checks `isDeactivated()` only at entry; after parking in `waitFence`, the resume block (`snapshotRetained`/`registerRetained`/`markReleaseAllowed`/`markReleasePrepared`) never re-checks deactivation, and `registerRetained` guards only `p.closing`, which production never sets. A `Prepare` resuming after `UnsubDmChannel` has run `delegator.Close()` → `Deactivate()` but before `RemoveBy` can re-create a `retained` entry (with a segment pin) on a deactivated provider that nothing will ever detach, leaking the pin and keeping `unregisterIfInactiveLocked` from firing. This structure predates the PR, so it warrants a follow-up rather than blocking: add a `deactivated` re-check (and/or a generation token) so a `Prepare` cannot commit after its handoff was invalidated. *(raised by tinswzy)*

- **`internal/querynodev2/delegator/growing_flush_source_test.go:505`** — The new race tests assert only the fence boolean and inflight count, not the map-level post-conditions where state resurrection would surface. `TestDelegatorGrowingSourceProviderKeepsHandoffOnlyWhileHandoffInflight` passes even though the resuming `Prepare`'s `markReleaseAllowed` re-creates a stale `releaseAllowed[1001]` entry after `ClearReleasePrepared` deleted it, and `TestReleaseGrowingSourceClearsHandoffOnWorkerReleaseFailure` asserts only `len(handoffAllowed)==0`. Strengthen at least one `TargetOffset>0` and one `TargetOffset=0` race test to assert the fully drained state — `handoffAllowed`, `releaseAllowed`, `releasePrepared`, and `retained` all empty, plus an unrelated `GetGrowingFlushSource` succeeding. *(raised by tinswzy)*

### Low

- **`internal/querynodev2/delegator/growing_flush_source.go:440`** — `ClearReleasePrepared` does a plain `delete(p.handoffAllowed, segmentID)`, dropping all ref-counted references at once, then runs `exitHandoffOnlyIfDrainedLocked`. Because `BeginGrowingSourceReleaseHandoff` intentionally holds no `handoffInflight` token, a concurrent `ReleaseSegments` → `ClearReleasePrepared` on the only fenced segment can lower `handoffOnly` while a handoff is still parked between `Begin` and `Prepare` — the very invariant the PR's own tests pin against. Narrow window; consider decrementing one reference per `Begin`, or having `Begin` hold an inflight-style token. *(surfaced during verification)*

- **`internal/querynodev2/delegator/growing_flush_source.go:356`** — `enterHandoffOnlyLocked` increments `handoffAllowed[id]` once per occurrence while the rollback snapshot's `added` is a set, so duplicate segment IDs in one call would take N references but roll back only one, pinning `handoffOnly` until an incidental `ClearReleasePrepared`. Currently unreachable since both production callers dedup via `lo.Uniq`, but adding an assert or `lo.Uniq` inside the function would make the invariant enforced rather than implicit. *(raised by tinswzy)*

bigsheeper:
Thanks — all four verified against `da166ec946`. Three are fixed in `41f2643f76` (also rebased onto latest master, so the earlier commits have new SHAs; content unchanged), and the fourth I am deliberately leaving, with reasons below.

**One of these turned out to be larger than the finding described, so I went past what was asked. Flagging that up front rather than burying it.**

---

**Medium — `growing_flush_source_test.go:505`, race tests assert only the fence. Confirmed, and the gap was hiding a real defect.**

I wrote the map-level assertions you asked for, and they failed: `releaseAllowed must not be resurrected`, `map[1001:100]`. So this was not only a coverage gap.

The mechanism: `ClearReleasePrepared(S)` retires a segment by wiping its `handoffAllowed` entry — "this handoff is over, forget it" — but a `Prepare` still parked in `waitFence` resumes and republishes `releaseAllowed`/`releasePrepared` for S anyway. With `TargetOffset > 0` it goes further and calls `registerRetained`, creating a retained entry **and taking a segment pin** that nothing detaches, since the concurrent release path has already moved on.

Your write-up scoped this as "strengthen the assertions"; the assertions could not be written truthfully without also fixing it. `Prepare` now filters its segment list after the fence down to those whose `handoffAllowed` reference it still holds (`stillFenced`), and returns early if none remain. That is a production change beyond the finding — say the word if you would rather see it split out, and I will carry the weaker assertion instead.

**Medium — `growing_flush_source.go:168`, no deactivation re-check after `waitFence`. Confirmed, and worse than "pre-existing".**

I checked the claim that `p.closing` is never set in production, and it is stronger than that: `provider.Close()` has **no production caller at all** — `delegator.go:1394`'s `Deactivate()` is the only one, and `p.closing = true` appears solely inside `Close()`. So `registerRetained`'s guard is not merely guarding the wrong field, it has never executed in production. This is the function's first live lifecycle guard.

Fixed in two layers. After `waitFence` returns, `Prepare` re-checks `isDeactivated()` and routes to `prepareDeactivatedGrowingSourceReleaseHandoff` — the same path a handoff arriving after `Deactivate` already takes, so no new semantics. `registerRetained` gets the guard too, as a backstop for the remaining window.

**A detail worth recording, because your suggested one-line fix would have caused a regression.** Adding `p.deactivated` to the existing guard and returning `errGrowingSourceProviderClosed` looks right but is not: that sentinel is a bare `errors.New`, and `isReleaseManualFlushPrepareUnavailable` (`services.go:740-754`) only recognizes merr sentinels. Making that branch live in production would have turned a `Prepare` racing `Deactivate` into a **failed `UnsubDmChannel`** instead of the intended "prepare unavailable, carry on". The new guard returns `merr.WrapErrChannelNotAvailable`, which that classifier does recognize. The reason this was invisible before is exactly the deadness above — the error had no live path to escape through.

**Low — `growing_flush_source.go:356`, duplicate ids take N references and roll back one. Confirmed, unreachable, fixed anyway.**

Both production callers dedup (`write_buffer.go:454`, `services.go:730`), so this is not reachable today. I did not add `lo.Uniq` inside the function, though — I made `handoffSnapshot.added` a `map[int64]int` recording how many references the call actually took, so the rollback returns exactly that many. That holds whether or not callers dedup, which makes the invariant structural instead of dependent on a convention two other files happen to follow.

**Low — `growing_flush_source.go:440`, `ClearReleasePrepared` drops all references at once. Confirmed, not fixed, and both suggested fixes break something.**

You are right about the window, and it is the same gap I flagged as unaddressed in my previous reply — good to have it independently pinned with a concrete path. But I traced both remedies and neither works as stated:

- *"decrement one reference per `Begin`"* — on the committed path `Begin`'s rollback closure is never invoked (the caller sets `handoffCommitted` and skips it), so its reference would never be returned. `handoffOnly` stays pinned forever, which is the original #51635 symptom this PR exists to remove.
- *"have `Begin` hold an inflight-style token"* — the comment at `:334-338` was written for precisely this: a token taken in `Begin` has no matching release on the committed path.

The real fix needs an explicit "handoff finished" signal on the committed path, so that `Begin`'s and `Prepare`'s references have symmetric lifetimes. That is a lifecycle change with its own design questions, and I would rather not smuggle it into this PR. Recorded as a follow-up.

---

Tests: the strengthened race test now asserts all four maps empty plus an unrelated segment becoming servable; new cases cover the `TargetOffset > 0` retirement race, `Deactivate` landing mid-fence, and the duplicate-id rollback. Each fix was checked in both directions — reverting `stillFenced` fails with `releaseAllowed must not be resurrected`, reverting the deactivation re-check fails with an unexpected error, reverting the reference count fails with `map[1001:1]` / `both references must be returned`. My first version of the retirement test hung for 600s instead of failing (an unexpected mock call on a background goroutine deadlocks), so it now sets a `.Maybe()` expectation and the defect surfaces as an assertion. Full `./internal/querynodev2/delegator/...` passes on the rebased head and `make static-check` reports 0 issues across all four scopes.

czs007:
Adversarial review found no issues requiring changes in this round.

Final-round reviewer consensus:

- yezh concurred with the previously reported high-severity findings and raised no additional issues.
- gzh confirmed agreement with no additional issues.

Note: the final-round reviewer notes did not include specific file:line verification evidence, so a `Verified:` list is omitted rather than fabricated. The reviewers' sign-off in this round consists of concurrence with the prior findings and confirmation that no new issues surfaced.

bigsheeper:
Closing. The bug is real and the fix works, but the implementation is more machinery than the problem warrants.

Six commits, and five of them exist to protect one stored boolean. `handoffOnly` is derived state kept as a field, so every mutation site has to remember to recompute it and no rollback may clobber a concurrent contribution. That is what produced the ref-counted `handoffAllowed`, the `handoffInflight` counter, the split between `enterHandoffOnly` and `enterHandoffOnlyInflight`, the delta-based `handoffSnapshot`, and `stillFenced` — each closing an interleaving in that scheme rather than in the original problem.

It does not need to be stored. All three sources are populated only by the handoff path: `handoffAllowed` only in `enterHandoffOnlyLocked`, `handoffInflight` only in the enter/leave pair, and `registerRetained` has exactly one caller, `PrepareGrowingSourceReleaseHandoff`. So

```go
func (p *delegatorGrowingSourceProvider) handoffFencedLocked() bool {
	return len(p.handoffAllowed) > 0 || len(p.retained) > 0 || p.handoffInflight > 0
}
```

is equivalent to the field. Deleting the field removes `exitHandoffOnlyIfDrainedLocked` and its four call sites, and with them the whole class of "changed state, forgot to recompute the fence" defect that generated the follow-ups. Ref-counting, the in-flight counter, the enter split and `stillFenced` stay — those answer the actual problem.

To be clear about what this is not: the defect itself holds. After a handoff completes, `ClearReleasePrepared` empties `handoffAllowed` while the fence stays up, and `acquireLease` then rejects every growing segment that is neither allowed nor retained — deterministic, not a race. Reachability is analysed on #51635. What is absent is urgency: `common.storage.useLoonFFI` and `common.storage.enableGrowingSourceFlush` both default to false and are ANDed, so the provider is not constructed at all in a default deployment.

Redoing this with the derived fence when the feature moves toward default-on. #51635 stays open.

PR #51638, opened 2026-07-20
fix: honor alter-field schema in growing flush

zhagnlu:
issue: #51344
Signed-off-by: luzhang <luzhang@zilliz.com>

### DIFF
--- a/internal/core/src/segcore/ConcurrentVector.h
+++ b/internal/core/src/segcore/ConcurrentVector.h
@@ -203,8 +203,24 @@ class VectorBase {
         return empty;
     }
 
+    // Monotonic materialization marker, the canonical flush-layout presence
+    // test. Set on the first set_data_raw (any path, incl. all-null bulk fill),
+    // never cleared. Avoids the race-prone chunk state empty()/valid_count,
+    // which a concurrent insert flips mid set_data_raw and interim-index
+    // raw-vector removal empties for a column that still exists.
+    bool
+    materialized() const {
+        return materialized_.load(std::memory_order_relaxed);
+    }
+
  protected:
+    void
+    mark_materialized() {
+        materialized_.store(true, std::memory_order_relaxed);
+    }
+
     const int64_t size_per_chunk_;
+    std::atomic<bool> materialized_{false};
 };
 
 template <typename Type, bool is_type_entire_row = false>
@@ -288,6 +304,9 @@ class ConcurrentVectorImpl : public VectorBase {
     set_data_raw(ssize_t element_offset,
                  const void* source,
                  ssize_t element_count) override {
+        // Entry, not post-chunk: an all-null column reaches here with
+        // valid_count==0 (no chunk emplaced), yet is still materialized.
+        mark_materialized();
         ssize_t valid_count = 0;
         ssize_t storage_offset = 0;
         if (use_mapping_storage_) {
--- a/internal/core/src/segcore/SegmentGrowingImpl.h
+++ b/internal/core/src/segcore/SegmentGrowingImpl.h
@@ -405,6 +405,12 @@ class SegmentGrowingImpl : public SegmentGrowing {
                    bool small_int_raw_type = false) const override;
 
     std::unique_ptr<DataArray>
+    bulk_subscript_with_field_meta(milvus::OpContext* op_ctx,
+                                   const FieldMeta& field_meta,
+                                   const int64_t* seg_offsets,
+                                   int64_t count) const;
+
+    std::unique_ptr<DataArray>
     bulk_subscript(milvus::OpContext* op_ctx,
                    FieldId field_id,
                    const int64_t* seg_offsets,
--- a/internal/core/src/segcore/segment_c.cpp
+++ b/internal/core/src/segcore/segment_c.cpp
@@ -954,12 +954,7 @@ namespace {
 
 // struct to hold field info for building Arrow arrays
 struct FieldInfo {
-    milvus::FieldId field_id;
-    std::string field_name;
-    milvus::DataType data_type;
-    milvus::DataType element_type;
-    bool nullable;
-    int64_t dim;  // for vector types
+    const milvus::FieldMeta* field_meta = nullptr;
     const milvus::segcore::VectorBase* vec_base;
     milvus::segcore::ThreadSafeValidDataPtr valid_data;
     // For TEXT fields with spillover: reader for temp LOB file
@@ -1231,10 +1226,13 @@ BuildFixedWidthVectorArrayFromBulkSubscript(
     const milvus::proto::schema::FieldData& data_array,
     int64_t num_rows,
     int64_t byte_width) {
+    const auto& field_meta = *field_info.field_meta;
+    auto data_type = field_meta.get_data_type();
+    auto dim = field_meta.get_dim();
     const void* data = nullptr;
     int64_t data_size = 0;
     const auto& vectors = data_array.vectors();
-    switch (field_info.data_type) {
+    switch (data_type) {
         case milvus::DataType::VECTOR_FLOAT: {
             const auto& values = vectors.float_vector().data();
             data = values.data();
@@ -1271,13 +1269,13 @@ BuildFixedWidthVectorArrayFromBulkSubscript(
         return arrow::Status::Invalid(fmt::format(
             "bulk_subscript vector payload size mismatch, field={}, bytes={}, "
             "validRows={}, byteWidth={}",
-            field_info.field_id.get(),
+            field_meta.get_id().get(),
             data_size,
             valid_count,
             byte_width));
     }
 
-    if (field_info.nullable || has_valid_data) {
+    if (field_meta.is_nullable() || has_valid_data) {
         arrow::BinaryBuilder builder;
         ARROW_RETURN_NOT_OK(builder.Reserve(num_rows));
         auto bytes = static_cast<const uint8_t*>(data);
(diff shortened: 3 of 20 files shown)


internal/core/unittest/test_storage.cpp
```diff
@@ -945,3 +1070,181 @@ TEST(MinioChecksumConfig, NeedChecksumOverrideDispatch) {
...
+    reduced_proto.clear_fields();
+    for (const auto& field : full_proto.fields()) {
+        if (field.fieldid() != 101) {
+            *reduced_proto.add_fields() = field;
+        }
+    }
+    auto reduced_schema = Schema::ParseFrom(reduced_proto);
+    auto* growing = dynamic_cast<SegmentGrowingImpl*>(segment.get());
+    ASSERT_NE(growing, nullptr);
+    growing->Reopen(reduced_schema);
```
yhmo, line 1225: internal/core/unittest/test_storage.cpp line:1218
Medium ---- This `Reopen` does not actually change the segment schema: both `full_proto` and `reduced_proto` leave `schema_version` at the default 0, while `SegmentGrowingImpl::Reopen` only swaps when the new version is strictly greater. The test therefore flushes with field 101 still present in the runtime schema and can pass without exercising the dropped-field behavior it claims to cover. Please give the schemas increasing versions and assert the reopened schema no longer contains 101 before flushing.

internal/core/unittest/test_storage.cpp
```diff
@@ -489,6 +489,131 @@ TEST_F(StorageTest, FlushGrowingSegmentSkipsEmptyFunctionOutputColumn) {
...
+    ts_field->set_data_type(milvus::proto::schema::DataType::Int64);
+    auto* pk_field = schema_proto.add_fields();
+    pk_field->set_fieldid(100);
+    pk_field->set_name("pk");
+    pk_field->set_data_type(milvus::proto::schema::DataType::Int64);
+    pk_field->set_is_primary_key(true);
+    // Ordinary nullable vector field in the segment's own schema: the ctor
+    // allocates its column, and the consume path bulk-fills it all-null
+    // (valid_data all-false, vector chunk empty) because inserts never carry it.
+    auto* vec_field = schema_proto.add_fields();
```
yhmo, line 534: internal/core/unittest/test_storage.cpp line:530
Medium ---- This regression constructs field 101 in the segment's initial schema, so it exercises `Insert`'s missing-field bulk fill rather than the AddField/`Reopen` path from #51344. `Reopen` must first append the new column metadata and then call `fill_empty_field`; a regression in that setup would not fail this test. Please create the segment with only the PK, insert rows, reopen with a higher-version schema that adds nullable field 101, and then flush/assert that 101 is materialized and retained.

internal/core/src/segcore/SegmentGrowingImpl.h
```diff
@@ -218,6 +218,15 @@ class SegmentGrowingImpl : public SegmentGrowing {
         return *schema_;
     }
 
+    // Reopen swaps schema_ under sch_mutex_; a bare get_schema() reference can
+    // dangle if the swap frees the old Schema mid-use. Snapshot under the lock
+    // (shared_ptr copy keeps the referent alive). Mirrors Collection::get_schema.
+    SchemaPtr
+    get_schema_snapshot() {
```
yhmo: internal/core/src/segcore/SegmentGrowingImpl.h line:225
Low ---- `get_schema_snapshot()` is added as a new public accessor but has no caller in this PR (the flush path parses `config->schema_blob` and `bulk_subscript_with_field_meta` receives the metadata explicitly). Keeping an unused synchronization API widens the class surface and leaves a second schema-access pattern to maintain; could this be removed until a caller needs it?

internal/flushcommon/syncmgr/growing_source.go
```diff
@@ -637,12 +630,11 @@ func (t *GrowingSourceSyncTask) Run(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
-	// Unification point: from here on the intended layout and the layout the
-	// flush actually writes are one. Every consumer below (writer config,
-	// binlog meta, metacache current split) sees the same trimmed groups.
-	columnGroups, err = t.trimColumnGroupsToMaterialized(ctx, columnGroups)
-	if err != nil {
-		return err
+	if hasCommittedGrowingSourceLayout(segment) {
```
buqian-zilliz: **[P1] Keep the persisted split aligned with the actual flushed fields**

This removes the only pre-flush normalization of `columnGroups`, but the downstream representations are still not unified. `FlushGrowingData` can omit a non-materialized function output and return the actual set in `FlushedFieldIDs`; `buildGrowingSourceInsertBinlogs` trims only its local copy, while `schemaBasedPattern` was built from the original groups and `Run` later stores those original groups in `UpdateCurrentSplit`. After a successful first flush or committed-flush retry, metacache can therefore advertise a column/group that is absent from the manifest and binlog `ChildFields`, so the next append can hit the same column-group mismatch and stall. Derive the writer pattern, binlog summaries, and persisted `currentSplit` from one authoritative field set.

internal/flushcommon/syncmgr/task.go
```diff
@@ -237,23 +239,31 @@ func resolveColumnGroups(segmentInfo *metacache.SegmentInfo, schema *schemapb.Co
...
 			field2idx[field.GetFieldID()] = idx
 		}
 		for idx, cg := range currentSplit {
-			cg.Columns = lo.Map(cg.Fields, func(fieldID int64, _ int) int {
-				return field2idx[fieldID]
-			})
+			columns := make([]int, 0, len(cg.Fields))
+			for _, fieldID := range cg.Fields {
+				columnIdx, ok := field2idx[fieldID]
+				if !ok {
```
buqian-zilliz, line 253: **[P1] Allow zero-row finalization with a historical split**

`currentSplit` may legally retain a field after DropField until compaction removes it. When a flush races the last in-flight growing-source batch, the follow-up finalization task has no batches, so `getGrowingSourceSyncTask` uses `GetSchema(0)` and this new lookup rejects the retained field before `Run` reaches its `expectedRows == 0` branch. The error text is not recognized by `isGrowingSourceLayoutMismatch`, so the deterministic failure is rescheduled indefinitely and the segment never reaches Flushed. Reuse the existing split without remapping for zero-row flush/drop tasks, or carry the historical schema and make this mismatch non-retryable.

internal/core/src/segcore/ConcurrentVector.h
```diff
@@ -288,6 +304,9 @@ class ConcurrentVectorImpl : public VectorBase {
     set_data_raw(ssize_t element_offset,
                  const void* source,
                  ssize_t element_count) override {
+        // Entry, not post-chunk: an all-null column reaches here with
+        // valid_count==0 (no chunk emplaced), yet is still materialized.
+        mark_materialized();
```
buqian-zilliz, line 309: **[P1] Publish materialization only after population completes**

Publish the materialization marker only after the column is fully populated. The marker is currently stored before validity mapping, chunk allocation, and data copy, while `FlushGrowingSegmentData` treats it as permission to read the column without synchronizing with `Reopen`/`fill_empty_field`. A concurrent schema backfill can therefore expose a missing chunk, partially initialized data, or an incomplete offset map, causing a flush failure or corrupted output. Store the marker at successful function exit, including the all-null path, and use release/acquire ordering for publication.

internal/core/src/segcore/segment_c.cpp
```diff
@@ -2021,36 +1969,20 @@ FlushGrowingSegmentData(CSegmentInterface c_segment,
...
-                            "segment {}",
-                            field_id.get(),
-                            growing_segment->get_segment_id());
-                        skipped_columns.insert(field_id.get());
-                        continue;
-                    }
+                // In the layout iff materialized (see VectorBase::materialized);
+                // this covers interim-index vector fields too, since their raw
+                // data lands in the index only after set_data_raw marked them.
+                if (!insert_record.materialized(field_id)) {
```
liliu-z, line 2067: The flush gate now returns UnexpectedError for any non-function-output field named by the flush schema (and allowed_field_ids/currentSplit) that is not materialized, replacing the pre-PR branch that skipped a field absent from the segment's own runtime schema. On the manifest-reload path, load_field_data_common skips a field the reloaded schema_ no longer has without calling set_data_raw, so a dropped-but-in-manifest field reloads non-materialized and would hit this error. The new FlushUsesTaskSchemaForDroppedFieldAfterReopen test only covers Insert-then-Reopen (the field is materialized before the drop) and never exercises the load path. In practice the Go-side resolveColumnGroups ErrDataIntegrity usually preempts this before segcore runs, and WAL-replay of pre-drop inserts materializes the field, so the concrete risk is a latent load-path/flush-gate inconsistency plus a test-coverage gap rather than a confirmed boot crash-loop.

internal/flushcommon/syncmgr/task.go
zhagnlu: This should be addressed now. Zero-row growing-source finalization no longer calls `getColumnGroups()` because `Run()` only resolves columnGroups when `expectedRows > 0`, so retained historical split fields are not remapped against the current sync schema.

  Also, no-batch tasks preserve the historical schema timestamp through `progress.schemaTimestamp` / segment start position fallback instead of always using `GetSchema(0)`. The `"missing from sync schema"` error is now classified by `isGrowingSourceLayoutMismatch()` as
  well, so it will not be retried indefinitely if reached elsewhere.

PR #51639, opened 2026-07-20
fix: accept empty-list expression template parameters

xiaocai2333:
## What & why

An empty-list expression template parameter — e.g.

```python
filter="array_contains_all(tags, {tags})"
filter_params={"tags": []}
```

— failed plan creation with `unknown template variable value type: <nil>`, while the equivalent inline literal `array_contains_all(tags, [])` succeeded.

Root cause: an empty list is encoded on the wire as a `TemplateArrayValue` whose typed `data` oneof is left unset, so `GetData()` is `nil`. `convertArrayValue` had no case for that and fell through to the `default` error branch.

Per the reporter this affects `IN` / `NOT IN`, `array_contains_all` / `array_contains_any` and `json_contains_all` / `json_contains_any`, across query, search, hybrid search and delete, on both gRPC and REST. All of these funnel through this single conversion point.

## Fix

Handle the unset (empty-list) case explicitly, mirroring the inline `[]` literal produced by `ParserVisitor.VisitEmptyArray`: an empty, same-type array with an unknown (`None`) element type. Parameterized empty lists now build the same plan as their inline equivalents.

## Tests

- `Test_ConvertToGenericValue_EmptyArray`: the conversion output for an empty template list matches the inline `[]` `GenericValue` exactly.
- `TestEmptyListTemplate`: for all six operations, the empty-list template now parses without error and fills to zero values, agreeing with the inline form.
- `TestEmptyListTemplateEqualsInline`: filled template plans are structurally identical to their inline `[]` plans (ignoring the template-only bookkeeping fields, which are inert for an empty list).
- Moved the previously "expected to fail" `Int64Field in {empty_list}` case into the valid set.

## Verification scope

Verified at the plan-parser (Go) level via the unit tests above; end-to-end query/search on a running cluster was **not** exercised in this environment.

issue: #51617

### DIFF
--- a/internal/parser/planparserv2/convert_field_data_to_generic_value.go
+++ b/internal/parser/planparserv2/convert_field_data_to_generic_value.go
@@ -90,6 +90,15 @@ func convertArrayValue(templateName string, templateValue *schemapb.TemplateArra
 			arrayValues[i] = parsedValue
 		}
 		elementType = schemapb.DataType_JSON
+	case nil:
+		// An empty list carries no typed elements, so the TemplateArrayValue's
+		// data oneof is left unset and GetData() is nil. Mirror the inline `[]`
+		// literal (ParserVisitor.VisitEmptyArray): an empty, same-type array
+		// with an unknown (None) element type. This keeps parameterized empty
+		// lists behaving identically to inline ones across IN / NOT IN,
+		// array_contains_*, and json_contains_*. See issue #51617.
+		arrayValues = nil
+		elementType = schemapb.DataType_None
 	default:
 		return nil, merr.WrapErrQueryPlanMsg("unknown template variable value type: %v", templateValue.GetData())
 	}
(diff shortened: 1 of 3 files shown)


xiaocai2333:
Closing as a duplicate of #51621, which fixes the same root cause more comprehensively.

Both PRs patch the same spot in `internal/parser/planparserv2/convert_field_data_to_generic_value.go`, but #51621 additionally covers the REST handler, the Go SDK client, access-log parameter lengths, and nested-array rejection, and ships Python e2e coverage. This PR is a strict subset of it.

This PR was opened by an automated nightly job that failed to notice #51621 was already open against the same issue. Apologies for the review noise.

One finding from this PR's CI that may be useful for #51621: the only e2e failure here was `test_milvus_client_struct_array_nullable.py::TestMilvusClientStructArraySchemaEvolution::test_struct_array_filter_template_matches_inline_expression[contains_all_empty_list]`, reported as `XPASS(strict)` — that case carries `pytest.mark.xfail(..., strict=True)` (lines 1285-1286) because the empty-list bug used to make it fail. Once the bug is fixed the case passes, and strict xfail turns that into a failure. #51621's e2e run (job 43700) fails on that same single case for the same reason, so removing the marker there should be the last thing blocking it.

PR #51644, opened 2026-07-21
fix: fail stale schema reopen instead of silent success

congqixia:
Related to #51068

Stale schema reopen used to return success from segcore, which let the Go caller update LocalSegment loadInfo even though the core state was not reopened. Return a retriable schema-version error instead so QueryCoord can retry with a fresh schema.

Map the new segcore code to ErrCollectionSchemaVersionNotReady and add tests for both the C++ stale reopen path and the Go-side loadInfo guard.

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -150,8 +150,7 @@
 namespace milvus::segcore {
 using namespace milvus::cachinglayer;
 
-constexpr auto kCollectionSchemaVersionNotReady =
-    static_cast<ErrorCode>(2046);
+constexpr auto kCollectionSchemaVersionNotReady = static_cast<ErrorCode>(2046);
 
 static std::string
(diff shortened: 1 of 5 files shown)


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -150,6 +150,8 @@
 namespace milvus::segcore {
 using namespace milvus::cachinglayer;
 
+constexpr auto kCollectionSchemaVersionNotReady = static_cast<ErrorCode>(2046);
```
liliu-z, line 153: `2046` isn’t registered in milvus-common yet. Can we add a named `ErrorCode` there and bump the dependency instead of hard-coding it here? Otherwise a future allocation could silently collide with this mapping.

liliu-z: Tracking this in #51704 and assigned it to @congqixia. I’m fine with handling it as a follow-up so this PR can land the stale-reopen correctness fix first.

liliu-z (commented):
/lgtm
/approve

PR #51648, opened 2026-07-21
enhance: use inverted index for LIKE match

aoiasd:
pr: https://github.com/milvus-io/milvus/pull/51477

## Summary
Route generic LIKE Match operations through the varchar inverted index again instead of falling back to raw-data scan.


zhengbuqian (approved):
/lgtm

PR #51651, opened 2026-07-21
enhance: [3.0] parallelize external-table index build reads and stream raw data to disk

xiaocai2333:
issue: #51707
pr: #51708

## Problem

For external-collection index build, downloading the raw vector data dominates wall time. Profiling a
worker in UAT (AWS us-west-2, laion 1B external parquet, 8C/32Gi worker) showed the download phase
taking ~63% of task time while CPU sat near idle and only **one** S3 connection was ever open.

Four independent causes were found by tracing the read path end to end:

1. `common.arrow.reader.holeSizeLimitBytes` / `rangeSizeLimitBytes` were applied only to
   `LoonFFIPropertiesSingleton`'s cached map. Index build and the FFI readers build their
   milvus_storage properties per task through `MakeInternalPropertiesFromStorageConfig`, bypassing
   that cache, so the configuration never reached those reads.
2. milvus-storage's global reader thread pool is never initialized, so
   `ThreadPoolHolder::GetParallelism()` returns 1 and `Reader::get_chunks()` reads chunks (row groups)
   strictly one at a time.
3. The per-round prefetch window (`reader.record_batch_max_size`, default 32MB) is smaller than a
   single 64MB-class parquet row group, so each round admits exactly one row group — i.e. one S3
   range read at a time, regardless of how large the arrow IO thread pool is.
4. Batch decode (external-type normalization + `FieldData` materialization) ran single-threaded on the
   consumer loop and strictly alternated with fetch rounds; additionally
   `cache_raw_data_to_disk_storage_v2` materialized the entire column in memory before writing it out.

## Changes

| Commit | What |
| --- | --- |
| `fix: carry arrow reader prebuffer config into per-task loon properties` | Stamp the configured prebuffer limits into every freshly built properties map. The pair is published and read as one coherent snapshot under its own lock (the parquet reader validates `range > hole`, so a mix of two generations would be rejected), separate from the singleton mutex so it stays callable from inside `Init`. |
| `feat: add file-level and row-group-level read concurrency for index build` | Two opt-in knobs (details below) plus the C API, DataNode wiring and hot-reload watchers, mirroring the existing arrow IO pool wiring. |
| `enhance: parallelize manifest batch decode and stream index-build raw data to disk` | `IterateFieldDataFromManifest` streams batches in order: `ReadNext` stays on the calling thread (the reader is not thread-safe) while per-batch decode is submitted to the MIDDLE pool with an in-flight window bounded by bytes (not batch count, which would scale with pool size and batch size). Decode now runs N-way parallel *and* overlaps with the reader's next prefetch round. `cache_raw_data_to_disk_storage_v2` consumes batches streaming, so this task's retention drops from the full raw column to the prefetch window plus the decode window. |
| `test: cover IterateFieldDataFromManifest contract in non-vortex manifest tests` | Moves the streaming-variant parity assertions into `FlushGrowingSegmentTest::ReadFlushedFieldData` so every flush round-trip case exercises them on plain parquet builds. |
| `fix: address review feedback on index-build read concurrency` | Publishes the two prebuffer limits as one snapshot (they are validated against each other downstream), bounds the decode window by bytes rather than batch count, reports the effective reader-pool size instead of the requested one, and corrects the configuration guidance below. |
| `test: add end-to-end coverage for CacheRawDataToDisk` | manifest -> `CacheRawDataToDisk` -> local file, checking the file exists and that its size, header and payload are right. This path had no effective test — the only case covering it is skipped wholesale — which is how a file-creation regression reached UAT undetected. |

`GetFieldDatasFromManifest` becomes a thin wrapper over the streaming variant, so mem-index and other
manifest readers inherit the parallel decode unchanged.

## New configuration

Both default to `0`. Note this disables the two new read-concurrency knobs but is **not** a full
no-op for the PR: `IterateFieldDataFromManifest` always moves batch decode to the MIDDLE pool, the
manifest `DiskFileManager` path always writes streaming, and `GetFieldDatasFromManifest` (used by
mem-index and other manifest readers) inherits both. Only the knobs are gated.

`readerThreadPoolSize` resizes in both directions at runtime for values > 0. Only `0` is special:
the pool cannot be destroyed, so rolling back to `0` is a no-op until restart — the watcher logs the
effective size alongside the requested one and warns when they diverge. Readers also latch
parallelism when they open, so a change only affects tasks started afterwards.

| Key | Default | Purpose |
| --- | --- | --- |
| `common.storage.readerThreadPoolSize` | `0` (off) | Initializes milvus-storage's global reader thread pool. Chunk (row-group) reads *may* be fanned out across it — see the caveat below on contiguous-chunk merging. Resizable at runtime; only disabling it (`0`) needs a restart. |
| `common.storage.indexBuildReadWindowBytes` | `0` (loon's 32MB) | Per-round read window stamped only into index-build manifest read properties. Set to N × row-group size (e.g. 512MB or 1GB) so one round spans multiple row groups whose column chunks prefetch in parallel on the arrow IO thread pool. Bounded to 4GB. Costs up to this many bytes buffered per running build task, on top of the decode window. Query-node load paths keep the default window. |

Suggested starting point. Note the window must span **more** chunks than the pool size to get
per-chunk fan-out: milvus-storage merges contiguous chunks without limit when the chunk count does
not exceed the pool size, so a 1GB window over 64MB row groups (16 chunks) with a 16-thread pool is
submitted as a single task. The range-level parallelism from arrow's parquet prebuffer is
independent of that pool and is what the measurements below primarily exercise.

```yaml
common:
  arrow:
    ioThreadPoolCoefficient: 8
  storage:
    readerThreadPoolSize: 16
    indexBuildReadWindowBytes: 1073741824   # 1GB
```

## Measured results

UAT, AWS us-west-2, same instance and collection, 8C/32Gi worker, external parquet
(64MB row groups, 768-dim fp32), comparing the base image against the same image with these commits
and the configuration above:

| | before | after |
| --- | --- | --- |
| download phase | 275s | 145s |
| task total (4.68M rows) | 437s (93.4 s per million rows) | 193s (51.5 s per million rows) |
| concurrent S3 connections during download | 1 | 19 |
| memory during download/write phase | full raw column materialized | 1.5–3 GiB |

A later same-environment A/B (2026-07-22, stock image vs this branch, both on the same disk and
collection) measured 283s avg (134 tasks) vs 151s avg — **~2x from the code+config alone**, which is
the honest software-only claim. The remaining historical gap traced to a deployment issue: the local
spill path (`localStorage.path`) sat on the node's EBS root volume (gp3, 125MB/s) instead of the
instance-store NVMe. With that fixed and `readerThreadPoolSize: 16`, tasks reach **71-78s per
3.75M-row task (~19-21 s per million rows)**, verified on this branch's image (74-75s). The
`[ManifestReadStats]` phase split added here is what attributed each of these steps. A 50-worker run
over the same dataset completed 1.17B rows / 251 index tasks in 35 minutes (effective parallelism 45x).

The knobs must be set to reproduce these numbers. The contribution of the knobs versus the
unconditional pipeline changes (parallel decode, streaming writes) has not been separated by an
ablation; that is worth doing before the guidance above is treated as tuned.

## Testing

- `StorageTest`: new cases for the pool C API (create/resize/validate/release) and the read-window API
  (stamping, reset, bounds), plus a case asserting the prebuffer config reaches freshly built
  properties.
- `StorageTest.ArrowReaderConfigSnapshotIsAtomicUnderHotReload`: hot-reloads between two valid
  generations while four threads build fresh properties, asserting no thread ever observes a
  crosswise mix.
- `FlushGrowingSegmentTest.IterateFieldDataFromManifestMultiBatchOrdering`: inserts 20k rows to force
  several record batches (loon's per-batch row limit is 8192) and compares the concatenated result
  against the values originally inserted, in order — independent ground truth rather than a
  comparison against the same implementation.
- `FlushGrowingSegmentTest`: every flush round-trip case (scalar / vector / string / nullable /
  partial range) additionally asserts the streaming variant delivers batches in order on the calling
  thread.
- Storage + flush tests and `pkg/util/paramtable` tests pass locally.

## Follow-ups (not in this PR)

- (Resolved during measurement, no code needed) A long-suspected "writeback bottleneck" — dirty
  pages parked at 4.5-4.8GB, drain capped at 130-150MB/s, a 3.3GB/s NVMe swap showing no gain, and
  an O_DIRECT prototype running 4x slower — turned out to be a deployment issue: `localStorage.path`
  pointed at the node's EBS root volume (gp3, 125MB/s baseline) while the instance-store NVMe sat
  idle. Diagnosed via per-write latency histograms + the `[ManifestReadStats]` phase split included
  here (three independent nodes converged to identical 26.8ms per 4MB write — the gp3 pace — while
  in-pod `dd` to the NVMe did 3.2GB/s). With the path fixed, default buffered writes are optimal
  (`file_dirty` = 0, io.pressure = 0); the O_DIRECT/sync_file_range prototypes are parked on wip
  branches as diagnostic history.
- An ablation separating the read-concurrency knobs from the unconditional pipeline changes, and a
  metric/test for the actual number of concurrent loon read tasks. Note that measurements on this
  workload must discard the first task of each freshly started worker; cold start inflates it enough
  to invert conclusions.
- Decoupling the batch consumer from the reader thread in `IterateFieldDataFromManifest` — no longer
  a performance need (buffered consume is 2-2.6s per task), kept as a robustness improvement for any
  future blocking consumer.
- The in-flight byte budget uses `arrow::util::TotalBufferSize`, which over-counts by ~10x on
  external-table columns (shared dictionary buffers counted per batch), making the decode window
  admit only 1-2 batches. Harmless today (decode-wait is ~6s) but worth a tighter estimator.
- `RemoteOutputStream::Write(int fd, size_t)` reuses a single buffer across loop iterations while
  handing it to the output stream. If the underlying stream ever writes asynchronously without
  copying, the buffer can be overwritten in flight. Worth hardening independently of this PR.


internal/core/src/storage/loon_ffi/property_singleton.h
```diff
@@ -62,9 +63,11 @@ class LoonFFIPropertiesSingleton {
     void
     SetArrowReaderConfig(int64_t hole_size_limit_bytes,
                          int64_t range_size_limit_bytes) {
+        arrow_reader_hole_size_limit_bytes_.store(hole_size_limit_bytes,
```
xiaocai2333: Could we publish `holeSizeLimitBytes` and `rangeSizeLimitBytes` as one coherent snapshot? These two relaxed stores, together with the two independent loads in `ApplyArrowReaderConfig`, allow a task creating fresh properties to observe values from different generations. For example, an update from 1 MiB/4 MiB to 8 MiB/16 MiB can transiently produce 8 MiB/4 MiB, and the Parquet reader will reject it because `range_size_limit <= hole_size_limit`. Please use a single immutable snapshot or a separate config lock, and add a hot-reload-versus-property-creation concurrency test.

pkg/util/paramtable/component_param.go
```diff
@@ -781,6 +783,37 @@ This configuration is only used by querynode and indexnode, it selects CPU instr
 	}
 	p.ArrowReaderRangeSizeLimitBytes.Init(base.mgr)
 
+	p.StorageReaderThreadPoolSize = ParamItem{
+		Key:          "common.storage.readerThreadPoolSize",
+		Version:      "3.0.0",
+		DefaultValue: "0",
+		Doc: `Size of milvus-storage's global reader thread pool. When > 0, chunk ` +
```
xiaocai2333, line 790: This fan-out guarantee does not hold for the recommended `window / row_group_size ~= pool_size` configuration with milvus-storage `e658197`. Its `split_chunks()` returns one unlimited continuous block when `chunk_count <= pool_threads`; therefore 16 contiguous row groups with a 16-thread pool submit only one Loon future. The observed S3 concurrency may be coming primarily from Arrow prebuffer rather than this pool. Please add a test or metric that verifies the actual number of concurrent Loon read tasks, and either fix the dependency batching logic or adjust the configuration guidance based on an ablation test.

internal/core/src/storage/storage_c.cpp
```diff
@@ -199,6 +200,52 @@ InitArrowReaderConfig(CArrowReaderConfig c_arrow_reader_config) {
...
+    try {
+        if (num_threads < 0) {
+            return milvus::FailureCStatus(
+                milvus::ConfigInvalid,
+                "loon reader thread pool size must be non-negative");
+        }
+        // 0 = leave the pool uninitialized (loon reads stay sequential,
+        // the pre-existing behavior). Once created the pool cannot be
+        // destroyed at runtime; later calls only resize it.
+        if (num_threads == 0) {
```
xiaocai2333, line 218: `0` is documented as off and the parameter is refreshable, but once the singleton exists this path returns success without changing the effective pool size. The watcher then logs `readerThreadPoolSize=0` as successfully reconfigured while the old pool is still active, so a dynamic rollback silently does nothing. Please either make this setting restart-only/non-refreshable, or report/reject the no-op using the effective pool size. It should also be documented that readers cache parallelism when they open, so enabling the pool only affects later tasks.

internal/util/initcore/init_core.go
```diff
@@ -501,6 +501,22 @@ func InitDiskFileWriterConfig(params *paramtable.ComponentParam) error {
 	return HandleCStatus(&status, "InitDiskFileWriterConfig failed")
 }
 
+// InitLoonReaderConfig applies the milvus-storage (loon) reader concurrency
+// settings: the global reader thread pool size (chunk/file-level read
+// fan-out) and the index-build read window (how many bytes one prefetch
+// round may span, which bounds how many row groups download in parallel per
+// round). Both default to 0 = pre-existing sequential behavior.
```
xiaocai2333, line 534: At the PR level, defaults of zero do not preserve the previous runtime behavior exactly. They only disable the two new read-concurrency knobs; `IterateFieldDataFromManifest` still unconditionally moves decoding to the MIDDLE pool, and the manifest DiskFileManager path still switches to streaming writes. `GetFieldDatasFromManifest` also changes behavior for its other callers. Please correct the PR description/no-op claim, or gate these changes as well if a true default-off rollout is required. The benchmark should also separate the contribution of the knobs from the unconditional pipeline changes.

internal/core/src/storage/Util.cpp
```diff
@@ -1634,12 +1640,39 @@ GetFieldDatasFromManifest(
...
+    // (the reader is not thread-safe), but everything after it — external
+    // normalization plus FieldData materialization — is pure per-batch work
+    // and is the dominant cost. Offloading it keeps the reader thread free
+    // to trigger the next prefetch round, so network fetch and decode
+    // overlap instead of strictly alternating. Results are delivered to
+    // `consumer` on this thread in batch order; the bounded in-flight
+    // window provides backpressure so decoded-but-undelivered batches
+    // cannot pile up without limit.
+    auto& pool = ThreadPools::GetThreadPool(ThreadPoolPriority::MIDDLE);
+    const size_t max_inflight = std::max<size_t>(2, pool.GetMaxThreadNum() * 2);
```
xiaocai2333: This is bounded by batch count, not by bytes. With the default 16-thread cap this permits roughly 32 decoded batches per build; at 64 MiB per batch that is already about 2 GiB, in addition to the reader window, Arrow buffers, the batch being written, and other concurrent build tasks. If the thread cap is disabled, this grows with CPU count. Please consider a byte-based budget or a global semaphore, and add a concurrent-build peak-memory test. At minimum, the comments and PR body should avoid claiming that peak memory is capped by the reader window.

internal/core/src/segcore/flush_growing_segment_test.cpp
```diff
@@ -81,12 +82,43 @@ class FlushGrowingSegmentTest : public ::testing::Test {
...
+            field_meta,
+            data_type,
+            dim,
+            element_type,
+            std::nullopt,
+            [&](FieldDataPtr fd) {
+                EXPECT_EQ(std::this_thread::get_id(), caller_tid);
+                streamed.push_back(std::move(fd));
+            });
+        EXPECT_EQ(streamed.size(), field_datas.size());
```
xiaocai2333, line 110: This does not independently verify the claimed streaming contract. Both `field_datas` and `streamed` are produced by the same new `IterateFieldDataFromManifest` implementation, and the comparison only checks row counts and validity, not values. Most callers of this helper use only a few rows, so they are also unlikely to create multiple batches; the existing large-data test does not read through this helper. Please add a real multi-batch test that forces decode futures to finish out of order and compares actual values, plus an end-to-end manifest -> `CacheRawDataToDisk` test that validates the raw header/payload and nullable sidecar.

xiaocai2333:
Review summary for `0e2b9f0101`:

The overall direction is reasonable and the benchmark is useful, but I think the following should be addressed before merge:

1. Publish the Arrow hole/range settings as one coherent snapshot; the current independent atomics can expose a mixed and invalid pair during hot reload.
2. Verify that `readerThreadPoolSize` produces real Loon read fan-out. In milvus-storage `e658197`, the recommended chunk-count/pool-size alignment can collapse contiguous chunks into a single future.
3. Clarify or fix hot-reload semantics: setting the pool size back to zero reports success but does not disable an existing pool, and already-open readers retain their original parallelism.
4. Correct the default-no-op and bounded-memory claims. Parallel decode and streaming writes are unconditional, and the in-flight limit is batch-count-based rather than byte-based.
5. Add a true multi-batch ordering/value test and an end-to-end manifest-to-local-raw-file/validity-sidecar test.

I left inline comments with the concrete code paths and suggested verification.

Issue #51652, opened 2026-07-21
[Bug]: Storage V2 projections use full chunk size estimates


sparknack:
## Environment

- Milvus Version: master
- Deployment Mode: storage V2 or external collection
- OS: Linux

## Reproduction

1. Load a storage V2 column group containing multiple physical columns.
2. Use a projected chunk reader that selects only a subset of those columns, including lazy sealed fields or JSON stats columns.
3. Create the ManifestGroupTranslator for the projected reader.

### Trigger Conditions

- Frequency: always when the projected columns are smaller than the full column group.
- Does NOT happen when the reader projects the complete column group.

## Expected Behavior

Cell sizing and transient-memory reservation should use decoded-memory estimates for the projected physical columns only.

## Actual Behavior

The translator uses the full chunk estimate. A narrow projection can therefore receive an inflated cell size and transient-memory reservation, which may unnecessarily constrain or block loading.

## Error Logs

No explicit error is required; the resource estimate is larger than the data loaded by the projection.

## Analysis Hints

- Suspect code: internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
- Lazy sealed fields and JSON stats need to pass their exact projected physical column names.
- milvus-storage PR https://github.com/milvus-io/milvus-storage/pull/586 adds per-column decoded-memory estimates and distinguishes unavailable estimates from valid zero estimates.

sparknack:
has been fixed.

PR #51653, opened 2026-07-21
fix: size storage v2 cells by projected columns

sparknack:
issue: #51652

- Bump milvus-storage for per-column decoded-memory estimates.
- Size Storage V2 cells from the exact projected physical columns for sealed fields and JSON stats.
- Use sampled bytes-per-row, total chunk estimates, then 4KB/row only when more precise estimates are unavailable.
- Add one decision log per translator without logging inside row-group or cell loops.


internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -179,18 +270,11 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
-        // Non-external: use format metadata; only if it reports zero
-        // (e.g. Vortex without size stats) fall back to a 4KB/row
-        // last-resort estimate.
-        if (fallback_bytes_per_row > 0 && cell_rows > 0) {
-            cell_size = cell_rows * fallback_bytes_per_row;
-        } else if (cell_size == 0 && cell_rows > 0) {
+        // A successful metadata estimate, including a known zero, takes
+        // precedence over fallbacks. The 4KB/row last resort is used only
+        // when projected, sampled, and total estimates are all unavailable.
+        if (size_estimate_source == SizeEstimateSource::LAST_RESORT &&
```
buqian-zilliz: **[P1] Keep live-row cells positive for the memory planner**

`get_chunk_column_estimated_size()` permits a successful zero for live rows; the pinned Lance implementation does this when a projected field is absent from an older fragment after schema evolution. This branch preserves that zero in `chunk_memory_size_`, and `get_cells()` forwards it as `CellSpec::memory_size`, but `LoadCellBatchAsync` rejects every non-positive cell before reading. A lazy projection of the new field can therefore fail on its first pin or warmup. Use a conservative positive fallback whenever `cell_rows > 0 && cell_size == 0`, while handling true zero-row chunks separately.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +131,111 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+            break;
+        }
+        for (size_t i = 0; i < projected_row_group_sizes.size(); ++i) {
+            projected_row_group_sizes[i] += column_sizes[i];
+        }
+    }
+
+    SizeEstimateSource size_estimate_source;
+    std::string total_estimate_error;
+    if (projected_estimate_available) {
```
buqian-zilliz, line 213: **[P1] Do not treat Lance's lower bound as a safe reservation**

This precedence treats every successful projected estimate as a safe decoded-memory reservation. At the pinned revision, Lance's single-page variable-width estimator returns encoded page bytes and explicitly documents them as only a lower bound on decoded Arrow memory. Because this branch bypasses `fallback_bytes_per_row`, highly compressed external data can make cell grouping and `CellSpec::memory_size` orders of magnitude too small, allowing concurrent loads to exceed the transient-memory budget or OOM. Keep the sampled estimate as a safety floor, or only prefer projected estimates that are guaranteed to be conservative.

internal/core/thirdparty/milvus-storage/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( milvus-storage_VERSION e658197)
-set( GIT_REPOSITORY  "https://github.com/milvus-io/milvus-storage.git")
+set( milvus-storage_VERSION eb25d608d9e7baf33dbdd51601eb5a126883a89e)
```
buqian-zilliz: **[P1] Pin a revision that normalizes per-file column order**

The pinned storage revision resolves a field index against the first file's schema and applies that position to every chunk, while each file's `column_memory_sizes` remains in its own physical order. Its append validation accepts equal column-name sets in different orders, so a valid group containing `[a, b]` followed by `[b, a]` returns `b`'s estimate for `a` in the second file. The translator then under-reserves and packs cells using the wrong size, which can exceed the memory budget at load time. Pin a revision that normalizes every file's estimates by field name or stable field ID.

internal/core/thirdparty/milvus-storage/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
...
+set( GIT_REPOSITORY  "https://github.com/jiaqizho/milvus-storage.git")
```
buqian-zilliz: **[P1] Do not ship a dependency from a personal fork**

This makes clean Milvus builds fetch a production dependency from a contributor-owned fork. The pinned SHA is still the head of an open upstream PR whose review decision is `CHANGES_REQUESTED`, and it remains two commits ahead of the official branch. The full SHA protects content integrity but not repository availability, so deleting or privatizing the fork would break uncached builds and bypass organization-controlled provenance. Merge the storage change first and pin it from `milvus-io/milvus-storage`, or use an organization-owned immutable mirror.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +131,111 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+    std::vector<uint64_t> row_group_sizes(row_group_rows.size(), 0);
+    std::vector<uint64_t> projected_row_group_sizes(row_group_rows.size(), 0);
+    std::string projected_estimate_error;
+    bool projected_estimate_available = !projected_columns.empty();
+    if (!projected_estimate_available) {
+        projected_estimate_error = "projection contains no columns";
+    }
+    for (const auto& column_name : projected_columns) {
+        auto column_size_result =
+            chunk_reader_->get_chunk_column_estimated_size(column_name);
```
buqian-zilliz: **[P2] Avoid quadratic projected-column lookup**

This call runs once per projected column, but the pinned reader performs a linear name lookup plus a second duplicate scan over the full column-group list on every invocation. JSON shredding places all columns in one group and supports up to 10,000 columns, so eager segment load or warmup can perform roughly 100 million string comparisons per translator before row-group sizing. Use `get_chunk_estimated_size()` for full projections, or fetch the bulk per-column estimates once and index the projected names.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
buqian-zilliz: This issue still exists: f9f5bd1 still returns a known zero for a field absent from a live fragment, and the translator forwards it to `LoadCellBatchAsync`, which rejects non-positive `CellSpec::memory_size` before reading.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
buqian-zilliz: This issue still exists: the pinned Lance estimator still uses encoded bytes for single-page variable-width columns, while the translator still selects that projected estimate ahead of the sampled bytes-per-row fallback.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -179,31 +299,32 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
-        // (e.g. Vortex without size stats) fall back to a 4KB/row
-        // last-resort estimate.
-        if (fallback_bytes_per_row > 0 && cell_rows > 0) {
-            cell_size = cell_rows * fallback_bytes_per_row;
-        } else if (cell_size == 0 && cell_rows > 0) {
+        // A zero estimate for live rows can mean that a projected field is
+        // absent from an older fragment after schema evolution. The memory
+        // planner requires a positive reservation, so use the sampled size
+        // when available and otherwise fall back to 4KB/row.
+        if (cell_rows > 0 && cell_size == 0) {
```
buqian-zilliz: **[P1] Apply zero-size fallback before cell aggregation**

The pinned reader reports zero when a projected field is absent from an older physical row group, but reading that row group still materializes typed null data. This fallback runs only after row groups have been packed and only when the whole cell sum is zero. A global cell range can mix live zero-estimate row groups with a positive row group, so the positive value masks the missing-row allocation in both `ComputeRowGroupsPerCell` and `CellSpec::memory_size`, allowing schema-evolved loads to exceed the transient-memory budget or OOM. Replace each live row group's zero projected estimate with a conservative fallback before computing `rgs_per_cell` and aggregating cells.

internal/core/thirdparty/milvus-storage/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( milvus-storage_VERSION e658197)
+set( milvus-storage_VERSION f9f5bd1)
```
buqian-zilliz, line 17: **[P1] Avoid quadratic lazy-reader metadata replication**

At f9f5bd1, every projected `ChunkReader` copies the complete per-row-group column-size vectors and builds a fresh retained `unordered_map<string, uint64_t>` containing every physical field, regardless of projection. The lazy JSON path retains one reader per shredded column, so N columns and R row groups now retain O(N^2 * R) metadata; the default 1,024-column limit already creates more than one million hash entries per row group, and the supported 10,000-column cap creates 100 million before any cell is pinned. `MetadataCache` shares only the source footer metadata, not these derived copies. Pin a revision that shares immutable indexed size metadata across projected readers or stores only requested-column estimates.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
buqian-zilliz: This issue still exists: successful projected estimates still bypass `fallback_bytes_per_row`, while f9f5bd1's Lance single-page variable-width estimator still returns encoded page bytes as only a lower bound on decoded Arrow memory.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
buqian-zilliz: This issue still exists for partial projections: the new aggregate fast path covers only exact full projections, while partial eager groups and per-column lazy readers still perform a linear full-group name lookup for every projected column, so wide groups remain quadratic.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
sparknack: This is to be expected. Estimates are not required to be fully accurate.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
sparknack: Fixed in `9aae9906e3`. Live row groups with a zero estimate now receive the sampled or 4 KiB/row fallback before `ComputeRowGroupsPerCell`, so a positive sibling row group cannot mask the reservation. Added `TestMixedZeroProjectedEstimateUsesPerRowGroupFallback`; the translator suite passes 28/28.

internal/core/thirdparty/milvus-storage/CMakeLists.txt
sparknack: Confirmed. Fixing this requires milvus-storage to share indexed immutable size metadata across projected readers instead of rebuilding group-wide maps. That cross-repo redesign is outside this PR. This patch records the `O(N^2 * R)` limitation with a FIXME and accepts it under the current assumption that partial/lazy column groups remain small.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
sparknack: Confirmed for partial projections. The current path is `O(P * N)`, and N single-column lazy readers make aggregate lookup `O(N^2)`; full projections use the aggregate fast path. Added a FIXME documenting this. The complete fix should be combined with shared indexed metadata in milvus-storage and is deferred from this PR.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +132,199 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+                projected_estimate_error = fmt::format(
+                    "column {} row group count mismatched, expected {}, "
+                    "actual {}",
+                    column_name,
+                    projected_row_group_sizes.size(),
+                    column_sizes.size());
+                break;
+            }
+            for (size_t i = 0; i < projected_row_group_sizes.size(); ++i) {
+                projected_row_group_sizes[i] += column_sizes[i];
```
buqian-zilliz, line 206: **[P1] Apply missing-field fallback before projection aggregation**

The projected estimates are summed across fields before any zero fallback is applied. At `f9f5bd1`, a field absent from an older physical file contributes zero, while the read path still materializes that field as a typed-null Arrow array. In a multi-field projection, any present sibling keeps the aggregate positive, so no reservation is added for the missing field; the full-projection aggregate path loses the same per-field information. Eager schema-evolved loads can therefore understate both cache usage and transient loading memory. Apply a type-aware positive fallback per projected field before aggregation, including the full-projection path.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +132,199 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+    // missing reservation or distort ComputeRowGroupsPerCell.
+    constexpr uint64_t kLastResortBytesPerRow = 4096;
+    const auto positive_fallback_bytes_per_row =
+        fallback_bytes_per_row > 0
+            ? static_cast<uint64_t>(fallback_bytes_per_row)
+            : kLastResortBytesPerRow;
+    size_t live_row_groups = 0;
+    size_t positive_fallback_row_groups = 0;
+    for (size_t i = 0; i < row_group_sizes.size(); ++i) {
+        if (row_group_rows[i] == 0) {
```
buqian-zilliz, line 288: **[P1] Do not preserve zero-byte cells for fully deleted Iceberg row groups**

The new storage pin reports a fully position-deleted Iceberg row group as zero rows with a zero estimated size. This branch skips the positive fallback but still leaves that row group in the cell ranges, so when it forms its own cell `chunk_memory_size_` remains zero. `get_cells()` forwards that value to `LoadCellBatchAsync`, which rejects non-positive `CellSpec::memory_size` before the Iceberg reader can return its valid empty batch; eager warmup or `PinAllCells` can therefore fail the segment load. Remove zero-row groups from the cell mapping or assign them a positive sentinel loading reservation.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
sparknack: Confirmed that a missing physical field can contribute zero before projection aggregation. This affects reservation accuracy rather than read correctness, and this PR will not duplicate Arrow's typed-null memory-layout accounting in the translator. A type-aware estimate should be produced by milvus-storage, which owns the actual read schema and materializes the null array; the existing per-field safety-floor TODO tracks that follow-up. Deferring this from this PR.

zhengbuqian (approved):

PR #51657, opened 2026-07-21
test: Add focused filtering expression coverage matrix

yanliang567:
### What problem does this PR solve?

This PR adds focused Python client E2E coverage for Milvus expression filtering gaps around JSON mixed-type expressions, NULL/UNKNOWN semantics, scalar/JSON index consistency, special scalar types, StructArray filtering, and historical `expressions` issues.

issue: #51568

Related issues: #51489, #51567, #51538, #51617, #50882, #50964, #48440

### What is changed?

- Use `query` with exact-ID assertions as the default oracle for filtering correctness.
- Keep `search`/`hybrid_search` only where the tracked issue is specific to that API or execution path (#51489, #51567, #51568, #51617, and #48440).
- Add deterministic scalar, ARRAY, JSON, NULL, arithmetic, composition, and negative-message matrices.
- Add materialized scalar, JSON, StructArray, TIMESTAMPTZ, and GEOMETRY index coverage.
- Add special-type, template-param, StructArray, mixed-segment, and RANDOM_SAMPLE coverage.
- Maintain the directory overview and coverage-family mappings.

### Review fixes

- Require 3000-row index materialization with exact identity, state, and row accounting.
- Reject missing index counters, fail fast on terminal states, and propagate cleanup errors.
- Add arithmetic threshold sentinels so removing `+` or `*` changes expected IDs.
- Add middle-position native/JSON ARRAY witnesses plus fractional ARRAY_DOUBLE cast/truncation and exact-array order/length decoys.
- Add TRIE prefix witnesses and a substring-only decoy.
- Split one-doc #51568 positive and negative checks into independent nodes.
- Gate issue-specific ANN known-loss xfails with scalar query baselines.
- Add active numeric, BOOL, STRING, and JSON array-subscript `IN` query controls plus same-path `meta["b"]` INT-only/STRING-only issue-specific search controls.
- Add a growing JSON-integer witness for float-literal cross-type evaluation.
- Add TIMESTAMPTZ day/year outer-`NOT` UNKNOWN cases.
- Add RTREE refinement witnesses for bounding-box false positives and paired INTERSECTS-true/CONTAINS-false boundary-point cases.
- Add active non-empty hybrid/delete template controls with a three-row delete fixture.
- Rebalance L0 toward active filtering smoke coverage.
- Keep #51193 skipped per the maintainer wontfix/deprecation decision.

### Validation

Latest rebased commit: `812b213fb5`

Latest focused-suite validation: `eb1cfc1058` (identical filtering patch before rebase)

Server under test:

```text
image: harbor.milvus.io/manta/milvus:master-20260721-efbcd35
version: unknown-20260721-efbcd352f9
git commit: efbcd352f9
build time: Tue Jul 21 09:40:32 UTC 2026
deploy mode: STANDALONE
```

```text
py_compile: passed
ruff check and format: passed
git diff --check: passed
mutation oracle audit: passed
index helper simulation: passed
```

```text
focused collect: 323 tests
L0=10, L1=33, L2=280
coverage mappings: 107
missing nodes: 0
level mismatches: 0
duplicate matrix IDs: 0
```

Remote focused suites with `-n 4 --dist loadgroup`:

```text
276 passed, 47 xfailed in 167.26s
```

There were no ordinary failures. All xfails matched exact signatures or result deltas of tracked open issues.

### Fixture strategy

- Shared collections consolidate equal schema/index requirements.
- Plain/indexed twins use identical logical values with decorrelated sibling fields.
- Destructive, mixed-segment, and special-schema cases remain isolated.
- `--dist loadgroup` keeps shared fixtures on one worker while independent groups run in parallel.

### Known behavior

- General filtering correctness is validated with `query` to avoid ANN recall and distance-ordering noise.
- Issue-specific `search` and `hybrid_search` reproductions retain their original API paths.
- `element_filter` is element-level; `MATCH_ANY` is row-level.
- Normal-vector search rejects element-level `element_filter` filters.
- Constant-power composition is supported; field-power expressions are rejected.
- Empty-template hybrid behavior remains a #51617 dynamic xfail; non-empty hybrid controls are active.
- The dynamic two-column crash reproduction remains pending-manual because it can terminate QueryNode.


wangting0128 (approved):
/lgtm

Issue #50562, opened 2026-06-16
[Enhancement]: [TEXT field] Create E2E test case


FeilongHou:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Create E2E test case

yanliang567:
Review notes for PR #50563 (`test: Add TEXT LOB Python client coverage`):

1. `test_text_lob_not_loaded_flush_and_reload` does not actually cover the unloaded path.

   In `tests/python_client/milvus_client/test_milvus_client_text_lob.py`, the test creates the collection with `index_params=index_params`. In the MilvusClient schema path, creating a collection with index params also loads the collection after index creation. That means the following insert/flush sequence is running on a loaded collection, while the test target says it verifies unloaded TEXT insert/flush.

   Suggested fix: create the collection with `load=False` and avoid passing index params through `create_collection`, or reuse the helper path that explicitly creates indexes separately and only loads when requested. The test should assert or otherwise ensure the collection is unloaded before insert/flush.

2. The PR adds permanent skipped placeholder tests.

   `test_text_lob_config_gated_storage_paths` and `test_text_lob_compaction_reuse_rewrite_lob_files` are collected but always skipped. Since LOB storage config and compaction reuse/rewrite behavior are core TEXT LOB scenarios, these placeholders can make the coverage look more complete than it is.

   Suggested fix: either remove the placeholder tests, or convert them into real gated tests / strict xfails with linked follow-up issues and clear unblock conditions.

Verification I ran locally:

- `python3 -m py_compile tests/python_client/milvus_client/test_milvus_client_text_lob.py tests/python_client/bulk_insert/test_bulk_insert_api.py tests/python_client/milvus_client/test_milvus_client_external_table.py` passed.
- `pytest --collect-only` could not be completed in my local environment: system Python is 3.9 and fails on `datetime.UTC`; Python 3.12 is present but does not have pytest installed.

PR #51658, opened 2026-07-21
test: add TEXT collection field coverage

FeilongHou:
## Summary

- Add L1 coverage for adding a nullable TEXT field through MilvusClient.add_collection_field.
- Verify pre-existing and omitted values read back as null after schema evolution.
- Verify a 65,537-byte TEXT LOB payload by checksum and exact text_match membership.
- Keep the collection unloaded during schema evolution to avoid unnecessary growing-segment setup and keep CI runtime low.

issue: #50562

## Test Plan

- [x] python3 -m pytest milvus_client/test_milvus_client_text_lob.py::TestMilvusClientTextLOBIndependent::test_text_lob_add_collection_field -q --tb=short
- [x] python3 -m ruff check tests/python_client/milvus_client/test_milvus_client_text_lob.py
- [x] python3 -m ruff format --check tests/python_client/milvus_client/test_milvus_client_text_lob.py
- [x] python3 -m py_compile tests/python_client/milvus_client/test_milvus_client_text_lob.py

### DIFF
--- a/tests/python_client/milvus_client/test_milvus_client_text_lob.py
+++ b/tests/python_client/milvus_client/test_milvus_client_text_lob.py
@@ -1836,6 +1836,102 @@ class TestMilvusClientTextLOBIndependent(TestMilvusClientV2Base):
         assert_rows_payload(rows_by_id, expected, [CONTENT_FIELD, CONTENT_ALT_FIELD])
 
     @pytest.mark.tags(CaseLabel.L1)
+    def test_text_lob_add_collection_field(self):
+        """
+        target: verify MilvusClient.add_collection_field supports adding a TEXT field
+        method: add an analyzer-enabled nullable TEXT field after one row exists, then insert null/LOB values
+        expected: old and omitted values are null, the LOB payload is exact, and text_match finds only that row
+        """
+        client = self._client()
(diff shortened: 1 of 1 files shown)


yanliang567 (commented):
Found one assertion gap that should be fixed before approval:

The nullable-value verification can pass when the requested field is missing from the query response entirely.

At the call to `assert_rows_payload(rows_by_id, expected, [added_text_field])`, the helper eventually evaluates `row.get(field)`. For rows 0 and 1, a response that omits `added_text` therefore produces the same `None` as a correctly returned null value. This means the test does not fully verify the stated expectation that pre-existing and omitted values "read back as null."

Please assert field presence and null explicitly, for example:

```python
assert added_text_field in rows_by_id[0]
assert added_text_field in rows_by_id[1]
assert rows_by_id[0][added_text_field] is None
assert rows_by_id[1][added_text_field] is None
```

Alternatively, update the payload assertion helper to require each requested output field to be present before validating its value.

FeilongHou:
Fixed in 8a217ca8da. The test now explicitly asserts that added_text is present in both the pre-existing row and the post-add omitted-field row before asserting each value is None, so a missing output field can no longer satisfy the nullable-value verification.

Issue #51655, opened 2026-07-21
[Enhancement]: Align AlterCollection broadcasts with TruncateCollection AckSyncUp behavior


chyezh:
## Goal

Align every `AlterCollection` broadcast path with `TruncateCollection` by enabling `AckSyncUp` on the broadcast message.

Today `TruncateCollection` disables fast DDL ack by building the broadcast with:

```go
WithBroadcast(channels, message.OptBuildBroadcastAckSyncUp())
```

Most `AlterCollection` paths currently build broadcast messages with only:

```go
WithBroadcast(channels)
```

This means RootCoord may receive fast ack after WAL append without waiting for StreamingNode consumers to sync up, while `TruncateCollection` waits for consumer-side ack. `AlterCollection` operations that affect collection metadata/schema visibility should have the same acknowledgement semantics as `TruncateCollection`.

## Current code paths to audit

Known `AlterCollection` builders that should be checked and aligned include:

- `internal/rootcoord/ddl_callbacks_alter_collection_name.go`
- `internal/rootcoord/ddl_callbacks_alter_collection_add_field.go`
- `internal/rootcoord/ddl_callbacks_alter_collection_add_struct_field.go`
- `internal/rootcoord/ddl_callbacks_alter_collection_schema.go`
- `internal/rootcoord/ddl_callbacks_alter_collection_properties.go`
- `internal/rootcoord/ddl_callbacks_alter_collection_field.go`
- `internal/rootcoord/ddl_callbacks_collection_function.go`

Reference behavior:

- `internal/rootcoord/ddl_callbacks_truncate_collection.go` uses `message.OptBuildBroadcastAckSyncUp()`.
- `internal/rootcoord/ddl_callbacks_drop_collection.go` also uses `message.OptBuildBroadcastAckSyncUp()` for the collection broadcast.
- Streaming broadcaster fast-ack behavior is controlled by `BroadcastHeader.AckSyncUp`; when it is not set, the broadcaster self-acks after append instead of waiting for consumer-side ack.

## Expected work

- Audit all `AlterCollection` message construction paths.
- Add `message.OptBuildBroadcastAckSyncUp()` to every collection-channel `AlterCollection` broadcast where consumers must observe the message before the operation returns success.
- Keep control-channel-only follow-up messages unchanged unless they also require sync-up semantics.
- Update/add unit tests to assert the generated `AlterCollection` broadcast header has `AckSyncUp == true` for the affected paths.
[...]

PR #51660, opened 2026-07-21
fix: flush schema-change alter collection

chyezh:
issue: #51655

- rootcoord: broadcast AlterCollection with AckSyncUp and wait for schema-change vchannel checkpoints before applying metadata
- streamingnode: flush channels for schema-change AlterCollection so pending L0 deletes and sealed L1 segments reach the checkpoint
- datacoord: expose an internal checkpoint wait path backed by WatchChannelCheckpoint

### DIFF
--- a/docs/agent_guides/streaming-system/message/message-semantic-collection.md
+++ b/docs/agent_guides/streaming-system/message/message-semantic-collection.md
@@ -8,7 +8,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 |---------|----------|-------------------|-------------|
 | CreateCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
-| AlterCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
+| AlterCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | TruncateCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | CreatePartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropPartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
@@ -34,7 +34,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 
 - **CreateCollection**: Creates a new collection with its partitions and VChannels.
 - **DropCollection**: Drops a collection and all its data, indexes, and load config. Implicitly flushes all growing segments.
-- **AlterCollection**: Alters collection properties, description, consistency level, or schema. Schema changes implicitly flush growing segments. When used for **RenameCollection**, the ResourceKey changes to `ExclusiveDBName(srcDB) + ExclusiveDBName(dstDB)` (deduplicated if same DB), blocking all collection DDL in both databases.
+- **AlterCollection**: Alters collection properties, description, consistency level, or schema. Uses AckSyncUp. Schema changes implicitly flush growing segments and wait for channel checkpoints to reach the schema-change timestamp before RootCoord applies the metadata update. When used for **RenameCollection**, the ResourceKey changes to `ExclusiveDBName(srcDB) + ExclusiveDBName(dstDB)` (deduplicated if same DB), blocking all collection DDL in both databases.
 - **TruncateCollection**: Logically truncates by sealing and dropping all segments before the truncation timestamp. Implicitly flushes all growing segments. Uses AckSyncUp.
 - **CreatePartition** / **DropPartition**: Creates or drops a partition. DropPartition implicitly flushes the partition's growing segments.
 - **CreateIndex** / **AlterIndex** / **DropIndex**: Manages indexes on a collection's field. CChannel-only.
--- a/docs/agent_guides/streaming-system/message/message-semantic-collection.md
+++ b/docs/agent_guides/streaming-system/message/message-semantic-collection.md
@@ -8,7 +8,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 |---------|----------|-------------------|-------------|
 | CreateCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
-| AlterCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
+| AlterCollection | Broadcast: VChannels + CChannel (schema-change AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | TruncateCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | CreatePartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropPartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
@@ -34,7 +34,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 
 - **CreateCollection**: Creates a new collection with its partitions and VChannels.
 - **DropCollection**: Drops a collection and all its data, indexes, and load config. Implicitly flushes all growing segments.
-- **AlterCollection**: Alters collection properties, description, consistency level, or schema. Uses AckSyncUp. Schema changes implicitly flush growing segments and wait for channel checkpoints to reach the schema-change timestamp before RootCoord applies the metadata update. When used for **RenameCollection**, the ResourceKey changes to `ExclusiveDBName(srcDB) + ExclusiveDBName(dstDB)` (deduplicated if same DB), blocking all collection DDL in both databases.
+- **AlterCollection**: Alters collection properties, description, consistency level, or schema. Schema-change alters use AckSyncUp, implicitly flush growing segments, and wait for channel checkpoints to reach the schema-change timestamp before RootCoord applies the metadata update. Non-schema alters, including property-only alters and renames, keep the fast-ack path. When used for **RenameCollection**, the ResourceKey changes to `ExclusiveDBName(srcDB) + ExclusiveDBName(dstDB)` (deduplicated if same DB), blocking all collection DDL in both databases.
 - **TruncateCollection**: Logically truncates by sealing and dropping all segments before the truncation timestamp. Implicitly flushes all growing segments. Uses AckSyncUp.
 - **CreatePartition** / **DropPartition**: Creates or drops a partition. DropPartition implicitly flushes the partition's growing segments.
 - **CreateIndex** / **AlterIndex** / **DropIndex**: Manages indexes on a collection's field. CChannel-only.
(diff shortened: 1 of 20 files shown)


czs007:

Reviewed at head `d76d674e3f`. The core mechanism is sound: `FlushChannel` sets `flushTs` on the write buffer, `GetFlushTsPolicy` (`writebuffer/sync_policy.go:80`) then syncs every buffer whose `MinTimestamp < flushTs` — and since L0 delete buffers live in `wb.buffers` via `bufferDelete` (`write_buffer.go:1416`), they are checkpoint candidates in `GetCheckpoint` (`write_buffer.go:571`), so `cp >= flushTs` really does imply the pending L0 deletes and sealed L1 segments have been synced and registered. The `ttNode` fast path (`flow_graph_time_tick_node.go:132`, `needUpdate=true` → `AddTask(..., flush=true)`) means the checkpoint is pushed to DataCoord promptly rather than on the 60s tick, so the added DDL latency is sync-bound. All 10 `NewAlterCollectionMessageBuilderV2` sites got `OptBuildBroadcastAckSyncUp()` — none missed. No blocking correctness defect found; the findings below are about liveness, blast radius, and over-triggering.

---

### 1. medium — unbounded, un-timeoutable wait: `WaitForChannelCheckpoint` can block forever, and a stuck ack callback wedges DDL for the whole DB

`internal/datacoord/services.go:2176`
```go
for channelName, flushTs := range flushTsList {
    if err := s.meta.WatchChannelCheckpoint(ctx, channelName, flushTs); err != nil {
```
`WatchChannelCheckpoint` (`internal/datacoord/meta.go:3615`) loops on `m.channelCPs.cond.Wait(ctx)` with no deadline, and that cond is broadcast from exactly **one** place — `UpdateChannelCheckpoints` at `meta.go:2859`. The other three mutation sites do not broadcast:

* `UpdateChannelCheckpoint` (singular, `meta.go:2730`)
* `MarkChannelCheckpointDropped` (`meta.go:2775`)
* `DropChannelCheckpoint` (`meta.go:2883`) — which **deletes** the map entry, so `ok == false` and the loop can never be satisfied for that channel.

The caller side is `s.notifier.Context()` (background, never cancelled) inside `callMessageAckCallbackUntilDone` (`ack_callback_scheduler.go:296`), which retries forever, while `doAckCallback` holds the task's resource-key guards for the whole duration.

Failure scenario: RootCoord restarts with an `AlterCollection` broadcast task in `WAIT_ACK` for a collection that has since been dropped and GC'd (`garbage_collector.go:1102` → `DropChannelCheckpoint`). On recovery the ack callback runs, `checkpoints[vchannel]` is gone, the wait never returns, `MarkAckCallbackDone` is never reached, and the `SharedDBName + ExclusiveCollectionName` resource keys are held permanently — every subsequent DDL in that database blocks with no error and no log after the initial "start to execute ack callback".

Suggested fixes: (a) bound the wait — `context.WithTimeout` in `WaitForChannelCheckpoint` returning a retriable `merr` so the ack-callback backoff loop makes the stall observable; (b) treat a missing entry or `funcutil.IsDroppedChannelCheckpoint(cp)` as already satisfied instead of waiting; (c) add `m.channelCPs.cond.UnsafeBroadcast()` under the lock in `UpdateChannelCheckpoint`, `MarkChannelCheckpointDropped` and `DropChannelCheckpoint`. This risk pre-existed for `DropSegmentsByTime`/TruncateCollection, but this PR moves it onto the common online-DDL path, so it is worth hardening here.

### 2. medium — `AckSyncUp` is now enabled for *every* AlterCollection, but the checkpoint wait it enables is only used for schema changes

`internal/rootcoord/ddl_callbacks_alter_collection_properties.go:213`, `:341`, `:403`, plus `..._name.go:97`, `..._field.go:169`, `..._add_field.go:118`, `..._add_struct_field.go:108`, `..._schema.go:181/424`, `..._collection_function.go:84`.

With `AckSyncUp`, `FastAck` returns early (`broadcast_task.go:457`), so the task is only acked once every StreamingNode has actually consumed the message from the WAL; `Broadcast` blocks on `BlockUntilDone` (`broadcast_scheduler.go:58`) until the ack callback finishes. Every property-only alter — `mmap.enabled`, TTL seconds, replica number, `RenameCollection` — now pays full WAL-consumption latency on all vchannels, and fails/hangs the client RPC while any one vchannel's StreamingNode is restarting or its flusher is still recovering. Previously these returned as soon as the append landed.

Note the checkpoint wait does not appear to require `AckSyncUp`: with fast ack, `FastAck` still populates `AppendResult.TimeTick` per vchannel, and `cp >= alterTimeTick` by itself already implies the flusher consumed past the alter message. So unless `AckSyncUp` is needed for a separate ordering guarantee (WAL-side apply before RootCoord meta apply), consider gating it the same way the wait is gated — the update mask is fully computed before `WithBroadcast(...)` at all 10 sites, so `IsSchemaChange(header)` is available there. If it *is* needed for all alters, please state which guarantee in the PR description.

### 3. medium — TTL-property alters are misclassified as schema changes, forcing a pointless full-channel flush

`internal/rootcoord/ddl_callbacks_alter_collection_properties.go:178`
```go
// Ensure schema update mask exists so QueryNode pipeline treats this as a schema update event.
if !funcutil.SliceContain(header.UpdateMask.Paths, message.FieldMaskCollectionSchema) {
}
// Build schema snapshot with updated properties (schema version should NOT be changed for properties-only alter).
```
The mask is added purely so QueryNode refreshes runtime schema properties, and the comment explicitly says the schema version is not bumped. But `messageutil.IsSchemaChange` (`pkg/streaming/util/message/messageutil/header.go:22`) only looks at that mask, so both new code paths fire: `HandleAlterCollection` force-flushes every growing + L0 buffer on every vchannel, and `alterCollectionV2AckCallback` blocks the DDL until all checkpoints catch up.

Failure scenario: a user toggling `collection.ttl.field` on a large collection with many vchannels triggers a cluster-wide flush of all growing segments and blocks the alter for the duration of the sync — a bulk write-amplification event for a metadata-only change.

Suggestion: distinguish "schema actually changed" from "schema snapshot re-broadcast". Either use a dedicated header flag / schema-version delta for the flush decision, or skip appending `FieldMaskCollectionSchema` here and give QueryNode a separate signal. Whatever predicate is chosen must be used on **both** sides (RootCoord wait and StreamingNode flush) or they will disagree — if the flusher skips `FlushChannel` while RootCoord still waits, the wait degrades to "wait until the buffer ages out via the stale policy", i.e. a multi-minute DDL.

### 4. low — `HandleAlterCollection` discards the passed-in `ctx`

`internal/streamingnode/server/flusher/flusherimpl/msg_handler_impl.go:140` and `:144`
```go
if err := impl.wbMgr.SealSegments(context.Background(), vchannel, ...)
...
if err := impl.wbMgr.FlushChannel(context.Background(), vchannel, putCollectionMsg.TimeTick()); err != nil {
```
The function takes `ctx context.Context` and `HandleAlterWAL` right below (`:173`, `:180`) correctly threads it through. Using `context.Background()` on the new line means the call is not cancellable on flusher shutdown / channel drop, and the log context (trace ID, channel fields) is lost. Please pass `ctx`.

### 5. low — missing negative test for the new branch

`internal/streamingnode/server/flusher/flusherimpl/msg_handler_impl_test.go:230` covers the schema-change case, but there is no test asserting that a **non**-schema-change `AlterCollection` does *not* call `FlushChannel`. Since mockery's `NewMockBufferManager(t)` fails on unexpected calls, a second case with only `SealSegments` expected (and no `FlushChannel` expectation) is a two-line addition and pins the gating that finding #3 is about.

Also, `TestAlterCollectionNonSchemaAckCallbackSkipsChannelCheckpoint` sets `.Return(nil).Maybe()` on `WaitForChannelCheckpoint` and then relies on `AssertNotCalled`; dropping the `.Maybe()` expectation entirely would make mockery itself catch a regression at call time rather than at assert time.

### 6. nit — helper placement

`waitForAlterCollectionSchemaChangeFlush` (`ddl_callbacks_alter_collection_properties.go:458`) is shared by all alter paths (add field, add struct field, alter field, rename, function, schema add/drop), not just the properties path. Consider moving it next to `alterCollectionV2AckCallback`'s shared helpers, or into a `ddl_callbacks_alter_collection_common.go`, so it is discoverable from the other seven call sites.

---

Nothing here blocks the fix's correctness for the reported bug; #1 and #3 are the ones I would want addressed (or explicitly deferred with an issue) before this lands on a release branch, since both convert a rare hang / a metadata-only change into cluster-wide DDL stalls.

chyezh:
Thanks for the detailed review.

Follow-up commit `f6d01cd3457c` addresses the actionable items here:

- Checkpoint wait hardening: `WatchChannelCheckpoint` now treats missing, nil, and dropped channel checkpoints as already satisfied, and `UpdateChannelCheckpoint`, `MarkChannelCheckpointDropped`, and `DropChannelCheckpoint` now broadcast the condition variable so waiters are woken up. I did not add a new timeout in this PR to avoid changing the shared ack-callback retry policy, but the drop/recovery deadlock case called out in the review is covered.
- AckSyncUp blast radius: AlterCollection broadcast options are now gated by `messageutil.IsSchemaChange(header)`, so non-schema alters such as rename and normal property-only changes keep the fast-ack path.
- Handler context: `HandleAlterCollection` now passes the incoming `ctx` to both `SealSegments` and `FlushChannel`.
- Tests: added coverage for schema-change vs non-schema-change behavior, checkpoint waiter wakeups, and context propagation. The non-schema ack-callback test no longer relies on `.Maybe()` for `WaitForChannelCheckpoint`.

For the TTL-field property alter: I kept it under the existing `messageutil.IsSchemaChange` contract. That path intentionally appends the schema update mask so QueryNode refreshes runtime schema properties without release/load. Splitting “schema snapshot refresh” from “physical schema mutation” would require a new predicate/header and coordinated RootCoord/StreamingNode/QueryNode changes, which is out of scope for this bugfix. Both RootCoord wait and StreamingNode flush now use the same predicate, so they stay consistent.

For helper placement: I kept the helper local for now and avoided a pure placement-only move. All AlterCollection broadcast callsites now route through the same helper, so the behavior is centralized without introducing a non-functional file move in this PR.

xaxys:

Reviewed at head b31dd270b8bd.

Key issues:

- High — `internal/datacoord/meta.go:3789`: `!ok || cp == nil` returns success, conflating an unknown checkpoint with a completed or dropped channel. `internal/datacoord/services.go:1452-1454` explicitly ignores checkpoint-initialization failures, so a live channel can legitimately be absent from this map. A schema-changing AlterCollection can then apply RootCoord metadata before the flusher checkpoint reaches the alter timestamp. The same helper can let TruncateCollection drop segment metadata prematurely, risking mixed-schema data or data loss.

- Low — `docs/agent_guides/streaming-system/message/message-semantic-collection.md:11,37`: the documentation says AlterCollection generally uses AckSyncUp, while `internal/rootcoord/ddl_callbacks_alter_collection_properties.go:460-464` enables it only for schema changes. Property-only alterations and renames intentionally retain fast ACK behavior, so the documentation contradicts the implementation.

Fix suggestions:

1. Treat only a checkpoint at or beyond the target, or the explicit dropped-channel sentinel, as successful. For missing/nil checkpoints, return a retriable system error or wait using authoritative channel-lifecycle state. Add coverage showing that a live channel with a missing checkpoint cannot satisfy the fence and that a subsequent checkpoint update wakes the waiter.
2. Document AckSyncUp as schema-change-only for AlterCollection, explicitly noting that non-schema alterations continue using fast ACK.

The latest import-format-only follow-up looks correct.

chyezh:
I checked the channel-checkpoint initialization path.

For the normal live-channel path, a collection vchannel is not expected to be missing from DataCoord channel checkpoints. RootCoord's CreateCollection ack callback calls MixCoord/DataCoord `WatchChannels` with per-vchannel start positions, DataCoord initializes the channel checkpoint through `UpdateChannelCheckpoint`, and DataCoord reloads persisted checkpoints from catalog during startup. So the `!ok || cp == nil` branch is not expected for a normally initialized live channel.

There is a separate pre-existing hardening gap in `WatchChannels`: if the initial checkpoint persistence fails, the current code logs a warning and continues. I do not think this PR should fix that by making `WatchChannelCheckpoint` wait or fail on a missing map entry, because that would bring back the dropped-channel/GC liveness problem this helper was hardened for and would also affect the existing TruncateCollection path. If we want to close the initialization failure gap, the right place is the `WatchChannels` initialization path: make the initial checkpoint persistence fail the watch or retry until it is durable, instead of changing the schema-change fence to wait on an unknown checkpoint entry.

The LOW documentation mismatch is fixed in `3300f259c7`: AlterCollection is now documented as schema-change-only for AckSyncUp, while non-schema alters keep the fast-ack path.

PR #51661, opened 2026-07-21
fix: [cp 3.0] flush schema-change alter collection

chyezh:
issue: #51655
pr: #51660

- rootcoord: broadcast AlterCollection with AckSyncUp and wait for schema-change vchannel checkpoints before applying metadata
- streamingnode: flush channels for schema-change AlterCollection so pending L0 deletes and sealed L1 segments reach the checkpoint
- datacoord: expose an internal checkpoint wait path backed by WatchChannelCheckpoint

### DIFF
--- a/docs/agent_guides/streaming-system/message/message-semantic-collection.md
+++ b/docs/agent_guides/streaming-system/message/message-semantic-collection.md
@@ -8,7 +8,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 |---------|----------|-------------------|-------------|
 | CreateCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
-| AlterCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
+| AlterCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | TruncateCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | CreatePartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropPartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
@@ -34,7 +34,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 
 - **CreateCollection**: Creates a new collection with its partitions and VChannels.
 - **DropCollection**: Drops a collection and all its data, indexes, and load config. Implicitly flushes all growing segments.
-- **AlterCollection**: Alters collection properties, description, consistency level, or schema. Schema changes implicitly flush growing segments. When used for **RenameCollection**, the ResourceKey changes to `ExclusiveDBName(srcDB) + ExclusiveDBName(dstDB)` (deduplicated if same DB), blocking all collection DDL in both databases.
+- **AlterCollection**: Alters collection properties, description, consistency level, or schema. Uses AckSyncUp. Schema changes implicitly flush growing segments and wait for channel checkpoints to reach the schema-change timestamp before RootCoord applies the metadata update. When used for **RenameCollection**, the ResourceKey changes to `ExclusiveDBName(srcDB) + ExclusiveDBName(dstDB)` (deduplicated if same DB), blocking all collection DDL in both databases.
 - **TruncateCollection**: Logically truncates by sealing and dropping all segments before the truncation timestamp. Implicitly flushes all growing segments. Uses AckSyncUp.
 - **CreatePartition** / **DropPartition**: Creates or drops a partition. DropPartition implicitly flushes the partition's growing segments.
 - **CreateIndex** / **AlterIndex** / **DropIndex**: Manages indexes on a collection's field. CChannel-only.
--- a/docs/agent_guides/streaming-system/message/message-semantic-collection.md
+++ b/docs/agent_guides/streaming-system/message/message-semantic-collection.md
@@ -8,7 +8,7 @@ All broadcast messages implicitly carry **SharedCluster** via the Broadcaster.
 |---------|----------|-------------------|-------------|
 | CreateCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropCollection | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
-| AlterCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
+| AlterCollection | Broadcast: VChannels + CChannel (schema-change AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | TruncateCollection | Broadcast: VChannels + CChannel (AckSyncUp) | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | CreatePartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
 | DropPartition | Broadcast: VChannels + CChannel | Yes (VChannel) | SharedDBName + ExclusiveCollectionName |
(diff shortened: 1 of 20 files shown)


tinswzy:

**Overall:** Mergeable. This is a faithful cherry-pick of master #51660 (commit `d76d674`) onto 3.0 — no conflict artifacts, and every symbol it depends on already exists on 3.0 (`OptBuildBroadcastAckSyncUp`, `messageutil.IsSchemaChange`, `funcutil.IsControlChannel`, `wbMgr.FlushChannel`, `meta.WatchChannelCheckpoint`). The `types.MixCoord` interface gained `WaitForChannelCheckpoint` and all implementers/mocks are updated consistently (`mix_coord.go`, `internal/mocks/mock_mixcoord.go`, `datacoord/mock_test.go`, `rootcoord/mock_test.go`), so no hidden compile gap. No blocking issues found — the notes below are non-blocking and are inherited from the original change.

**Correctness of the core semantics — looks right:**
- `AckSyncUp` is now added to every AlterCollection broadcast variant (properties/dynamic-field/add-field/add-struct-field/field/rename/schema-add/schema-drop/function), so the ack callback runs with full per-vchannel results.
- The checkpoint wait in `alterCollectionV2AckCallback` (`internal/rootcoord/ddl_callbacks_alter_collection_properties.go:479`) only fires when `IsSchemaChange(header)` is true. I verified the schema-mutating paths (`add_field.go:108`, `schema.go:170/413`) set `FieldMaskCollectionSchema` in the `UpdateMask`, so they correctly trigger the flush wait, while pure rename/properties alters pay only the AckSyncUp cost without an unnecessary flush wait. Ordering is correct: the wait happens *before* `c.meta.AlterCollection`, so pending L0 deletes / sealed L1 segments under the old schema reach the checkpoint before metadata switches.

**low — unbounded checkpoint wait if a vchannel checkpoint never advances**
`internal/datacoord/meta.go:3578` `WatchChannelCheckpoint` loops on `cond.Wait(ctx)` until `cp.GetTimestamp() >= targetTs` or `ctx` is cancelled. `WaitForChannelCheckpoint` (`internal/datacoord/services.go:2176`) walks channels sequentially with no internal deadline, and `waitForAlterCollectionSchemaChangeFlush` passes the callback `ctx` straight through. If a target vchannel stops advancing mid-alter (streamingnode not consuming, channel being dropped, or the FlushChannel never completing), the AlterCollection ack callback blocks until that `ctx` is cancelled. Because AckSyncUp callbacks can run under a long-lived broadcast/recovery context, a single stuck channel can stall the ack indefinitely. This matches master and the wait is cancelable (it honors `ctx`), so it is not a regression — flagging for awareness. A bounded timeout on the wait path would harden it.

**nit — `context.Background()` drops handler cancellation/deadline**
`internal/streamingnode/server/flusher/flusherimpl/msg_handler_impl.go` `HandleAlterCollection` now calls both `SealSegments(context.Background(), ...)` and the new `FlushChannel(context.Background(), vchannel, TimeTick())`. The `SealSegments` `context.Background()` is preexisting; the new `FlushChannel` copies it, so cancellation/deadline from the handler `ctx` is not propagated. Consider threading the real handler `ctx` for both while touching this line.

**Test coverage:** the added `ddl_callbacks_alter_collection_ack_sync_test.go` exercises the schema-change vs non-schema-change branch and the wait path, and `msg_handler_impl_test.go` covers the new FlushChannel-on-schema-change behavior — adequate for the change.

No further concerns; approve on cherry-pick correctness.

chyezh:
Thanks for reviewing the cherry-pick.

The follow-up master fix has now been cherry-picked to this 3.0 PR as `29047773803d`.

This includes the actionable hardening from the master review:

- `HandleAlterCollection` now threads the incoming `ctx` through `SealSegments` and `FlushChannel`, so the cancellation/deadline concern is addressed here as well.
- `WatchChannelCheckpoint` now treats missing, nil, and dropped checkpoints as satisfied, and checkpoint mutation paths now broadcast waiters. I did not add a separate timeout in this PR because that would change the shared ack-callback retry behavior; the recovery/drop deadlock case is handled by the satisfied-on-missing/dropped rule.
- AlterCollection `AckSyncUp` is now gated by `messageutil.IsSchemaChange(header)`, matching the checkpoint-wait predicate and avoiding the extra sync-up cost for non-schema alters.
- The same rootcoord/datacoord/flusher test coverage from master was carried over.

The TTL-field property alter remains classified through the existing `messageutil.IsSchemaChange` contract because that path intentionally marks the schema mask for QueryNode runtime schema-property refresh. Splitting that from physical schema mutation needs a separate coordinated predicate/header and is outside this bugfix.

xiaocai2333:
🤖 **team2 automated review** (single agent · not cross-verified, for reference) · by `xiaocai`

Reviewed at head `b4c35fb` (base `3.0`). **Mergeable** — this is a faithful backport of the master change (`#51655` + the `#51660` follow-up) with a 3.0-only gci import fix on top. Concurrency of the checkpoint wait path is sound, symbol surface exists on 3.0, and the full CI loop (Build / Code-Check / UT-GO / UT-Integration / UT-CPP-Cov) is green on the head commit. Findings below are non-blocking (low / nit / informational). I've tried not to repeat the earlier cherry-pick/symbol audit already posted on this PR.

### Verified correct (called out because they're the risky parts)
- **Cond signalling is race-free.** `channelCPs.cond` is built over the *same* `RWMutex` (`meta.go:150` `NewContextCond(&cp.RWMutex)`), and all four new `cond.UnsafeBroadcast()` sites (`meta.go:2754/2787/2836/2858`) run under `channelCPs.Lock()`, matching `UnsafeBroadcast`'s "must hold the lock" contract. `WatchChannelCheckpoint` re-checks the predicate under the lock in a loop, so no lost-wakeup.
- **No double-unlock.** `ContextCond.Wait` unlocks `L` and, on the `ctx.Done()` path, returns the error *without* re-locking; `WatchChannelCheckpoint` (`meta.go:3591`) correctly returns straight through without a second `Unlock()`.
- **Schema-change gating logic** in `alterCollectionV2AckCallback` (`ddl_callbacks_alter_collection_field.go:886+`) waits *before* `meta.AlterCollection`, control channel is filtered out of `flushTsList` via `funcutil.IsControlChannel`, and the error path aborts the metadata update — the `PropagatesChannelCheckpointError` test confirms `AlterCollection` is not called on wait failure.

### low — shared `WatchChannelCheckpoint` semantics also change for TruncateCollection
`internal/datacoord/meta.go:3585` — the predicate changed from `ok && cp != nil && cp.GetTimestamp() >= targetTs` to `!ok || cp == nil || IsDroppedChannelCheckpoint(cp) || cp.GetTimestamp() >= targetTs`. This function is still shared with the pre-existing `DropSegmentsByTime` / TruncateCollection consumer (`services.go:2159`). Previously a **missing** checkpoint blocked; now it returns immediately as "satisfied". For the new schema-change flush wait this is intended, but it also silently relaxes TruncateCollection's wait. In practice a live collection always has a checkpoint (reloaded from catalog on restart), so the missing-checkpoint window is narrow — but please confirm TruncateCollection is happy with "no checkpoint ⇒ proceed" too. This matches master, so it's a heads-up rather than a regression.

Note also that the `IsDroppedChannelCheckpoint(cp)` clause is effectively redundant with the existing `>= targetTs` check: the dropped sentinel is `math.MaxUint64` (`funcutil/checkpoint.go:36`), so `cp.GetTimestamp() >= targetTs` is already true for any real `targetTs`. It's harmless and makes intent explicit — no change requested, just noting the real behavioral delta is the `!ok || cp == nil` early return, not the dropped case.

### low — unbounded wait can wedge the DDL ack callback
`ddl_callbacks_alter_collection_field.go:869-484` → `datacoord/services.go:2182` iterate channels and call `WatchChannelCheckpoint` with no internal deadline; if a streamingnode never advances the checkpoint (stuck flush) the AlterCollection ack callback blocks for as long as `ctx` lives. The `ctx` threading added in the follow-up is the right mitigation (cancellation/shutdown propagates), so this is acceptable, but worth confirming the ack-callback `ctx` actually carries a bound in production rather than being effectively `context.Background()`. Channels are also waited sequentially, so total latency is the sum across vchannels — fine for correctness, just a note for large fan-out collections.

### nit — backport is ahead of the still-open master PR
The PR body links `pr: #51660`, but `#51660` is currently **OPEN / not merged** (the label bot already flags this), and the cherry-pick source hashes (`d76d674`, `f6d01cd`) don't correspond to `#51660`'s current commits (`481decb`/`22ba8e0`/`b31dd27`). Content-wise the backport looks complete, but please land master first (or in lockstep) so the two don't diverge — especially since commit 3 (`b4c35fb`, gci import ordering) is 3.0-only and won't have a master counterpart.

### nit — test coverage is good
The added `ddl_callbacks_alter_collection_ack_sync_test.go` and the flusher `HandleAlterCollection` schema-change/non-schema-change split tests exercise both the AckSyncUp-marking and the flush-vs-skip branches, and the new `meta_test.go` cases cover missing / single-update / drop / mark-dropped wakeups. `TestServer_DropSegmentsByTime` was correctly updated to seed a checkpoint so the new missing-checkpoint early-return doesn't mask the intended wait. Nice.

Overall: no blocking issues — approve once the master PR merge ordering is sorted.

PR #51317, opened 2026-07-13
feat: [cp 3.0] support xgboost function chain expr

junjiejiangjjj:
pr: https://github.com/milvus-io/milvus/pull/51195
issue: https://github.com/milvus-io/milvus/issues/51192

PR #51664, opened 2026-07-21
test: [cp 3.0] add XGBoost FunctionChain coverage

zhuwenxing:
## What changed

- Backport the complete MilvusClient XGBoost FunctionChain test suite from #51545 to the `3.0` branch.
- Preserve the functional L0/L1/L2 layering and the L3 performance, recovery, multi-replica, and Chaos coverage.
- Keep the same realistic 3,000-row-per-segment fixtures and data-correctness assertions used by the master test PR.

## Why

The XGBoost L0 FunctionChain server implementation was backported to `3.0` by #51317, but the complete MilvusClient regression suite is only present in the master test PR.

This PR depends on #51614 for TestPyPI `pymilvus==3.0.1rc64`, which provides `pymilvus.function_chain.fn.xgboost`. The current `3.0` requirement, `3.1.0rc62`, does not expose that helper. This PR intentionally does not duplicate the requirements change.

pr: #51545

## Validation

- Verified the test file is byte-identical to the file in #51545.
- Verified `fn.xgboost` import with TestPyPI `pymilvus==3.0.1rc64`.
- `python -m py_compile tests/python_client/milvus_client/test_milvus_client_function_chain.py`
- `ruff format --check tests/python_client/milvus_client/test_milvus_client_function_chain.py`
- `ruff check tests/python_client/milvus_client/test_milvus_client_function_chain.py`
- `pytest --collect-only`: 37 nodes total; L0/L1/L2/L3 = 1/8/4/24.

Runtime E2E has not yet been rerun against a `3.0` deployment during this backport preparation; the previous master/cluster results are not claimed as `3.0` evidence.

### DIFF
--- a/tests/python_client/milvus_client/test_milvus_client_function_chain.py
+++ b/tests/python_client/milvus_client/test_milvus_client_function_chain.py
@@ -1,0 +1,3957 @@
+import io
+import json
+import os
+import re
+import subprocess
+import threading
+import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from datetime import UTC, datetime
+
+import numpy as np
+import pytest
+import xgboost as xgb
+from base.client_v2_base import TestMilvusClientV2Base
+from common import common_func as cf
+from common import common_type as ct
+from common.common_type import CaseLabel, CheckTasks
+from pymilvus import AnnSearchRequest, DataType, FunctionChain, FunctionChainStage, RRFRanker
+from pymilvus.client.types import LoadState
+from pymilvus.function_chain import col, fn
+from pymilvus.grpc_gen.common_pb2 import SegmentState
+from utils.util_log import test_log as log
+
+
+class TestMilvusClientFunctionChainXGBoost(TestMilvusClientV2Base):
+    """MilvusClient coverage for XGBoost L0 FunctionChain reranking."""
+
+    nb = ct.default_nb
+    sealed_segment_count = 2
+    segment_count = 3
+    total_nb = nb * segment_count
+    dim = ct.default_dim
+    nq = ct.default_nq
+    limit = ct.default_limit
+    vector_field = "embedding"
+    feature_names = [
+        "user_level",
+        "item_category_id",
+        "recent_click_count",
+        "inventory_count",
+        "ctr_7d",
+        "quality_score",
+    ]
+    feature_types = [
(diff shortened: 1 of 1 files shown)


yanliang567 (changes requested):
Requesting changes because two L3 cases can leave persistent mutations/faults in the shared Kubernetes environment.

1. **[P1] The rolling-image test never restores the original Milvus CR image.** The test patches `spec.components.image` to `MILVUS_ROLLING_IMAGE`, but the `finally` block only cleans up the FileResource. On either success or an intermediate assertion failure, the cluster remains on the target image and affects subsequent tests. Please capture the original CR value before patching, restore it in `finally`, and wait for the rollback to become Healthy.

2. **[P1] Chaos cleanup failures are silently ignored.** `_delete_chaos()` uses `check=False` and discards stdout/stderr. If deletion of the `NetworkChaos` resource fails, the QueryNode-to-MinIO partition can remain active while the test exits successfully. Cleanup should retry, verify that the resource is absent, and surface teardown failure with diagnostics.

3. **[P2] The warm-cache MinIO pod-kill test does not exercise search while MinIO is unavailable.** It waits for the replacement MinIO Pod to become Ready before issuing the 20 searches, and only requires `success_count > 0`. The test therefore passes even if QueryNode does not use the warm model cache and simply downloads the model again after MinIO recovers. Run a search worker continuously from before injection through recovery and assert an explicit availability/error budget for the outage window.

4. **[P2] The PR dependency/validation description is stale.** #51614 was closed without merge. This PR's actual base already uses `pymilvus==3.0.1rc66`, while the current `3.0` branch uses `3.0.1rc74` via #51942. Please update the description and rerun against the current 3.0 base.

Verification performed on head `646f41f935928e2666b1f973e1c5a473e232fa5f`:
- The test file is blob-identical to the merged master PR #51545.
- Ruff, format, `git diff --check`, and `py_compile` passed.
- `pymilvus==3.0.1rc74` exposes `fn.xgboost`.
- 37 tests collect: L0/L1/L2/L3 = 1/8/4/24.
- Current e2e-amd, Go SDK, and main-trigger checks are green.

The L3 Chaos/rolling scenarios were not executed locally because they require the dedicated Kubernetes topology. Since the same file is already on master, please apply the environment-safety fixes there as well and backport them consistently.

PR #51666, opened 2026-07-21
test: cover sealed vector index cache pin lifetime

chasingegg:
## What this PR does

- Add regression coverage for the sealed vector index cache pin lifetime used by GroupBy iterators.
- Verify that cache eviction remains blocked while `SearchResult` owns the iterators and succeeds after `SearchResult` is destroyed.
- Keep this PR test-only because the production behavior is already provided by #51623 through `SearchResult::resource_pins_`.

## How was it tested?

- `cmake --build cmake_build --target all_tests -j 32`
- `all_tests --gtest_filter='SearchOnSealedIndexCachePinLifetime.*:SearchOnSealedIndexBitsetLifetime.*'` (2 tests passed)
- Red/green check: temporarily removing the `resource_pins_` retention from #51623 makes the new test fail because `ManualEvictAll()` succeeds while `SearchResult` still owns iterators; restoring it makes the test pass.

### DIFF
--- a/internal/core/src/query/SearchOnSealedIndexBitsetLifetimeTest.cpp
+++ b/internal/core/src/query/SearchOnSealedIndexBitsetLifetimeTest.cpp
@@ -268,6 +268,58 @@ TEST(SearchOnSealedIndexBitsetLifetime,
     AssertVectorIteratorUsableAfterSearchReturns(search_result, valid_count);
 }
 
+TEST(SearchOnSealedIndexCachePinLifetime,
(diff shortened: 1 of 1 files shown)


chasingegg:
/hold

maybe has been fixed by other pr, checking

chasingegg:
/unhold

make this pr only modify UT as the bug has been fixed by other pr

liliu-z (commented):
/lgtm
/approve

Issue #50425, opened 2026-06-10
[Enhancement]: refactor flush mode from writerbuffer to growing source


zhagnlu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Milvus supports flushing growing segment data either from DataNode WriteBuffer payload or from QueryNode growing-source data. This mechanism is used by TEXT fields, and can also apply to ordinary fields when growing-source flush is enabled.

PR #51668, opened 2026-07-21
fix: disable growing source flush by default

zhagnlu:
issue: #50425 
Disable growing-source flush by default so TEXT collections use the writer-buffer path unless the feature is explicitly enabled.

Also write TEXT V3 bloom filter and BM25 stats into the manifest instead of returning legacy stats logs.

### DIFF
--- a/internal/storage/binlog_record_writer.go
+++ b/internal/storage/binlog_record_writer.go
@@ -517,7 +517,7 @@ func (pw *PackedManifestRecordWriter) Close() error {
 // storage, and appends StatEntry records onto updates so the surrounding
 // commit registers inserts + stats atomically. Leaves pw.statsLog and
 // pw.bm25StatsLog nil so callers know stats are embedded in the manifest.
-func (pw *PackedManifestRecordWriter) appendV3Stats(updates *packed.ManifestUpdates) error {
+func (pw *packedBinlogRecordWriterBase) appendV3Stats(updates *packed.ManifestUpdates) error {
 	statsBlob, pkFieldID, err := pw.pkCollector.SerializeBlob(pw.rowNum)
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


internal/storage/binlog_record_writer.go
```diff
@@ -720,16 +720,20 @@ func (pw *PackedTextManifestRecordWriter) Close() error {
 	}
 	pw.finalizeBinlogs()
 	if out == nil {
-		return pw.writeStats()
+		return nil
 	}
 
+	updates := &packed.ManifestUpdates{NewFiles: out}
+	if err := pw.appendV3Stats(updates); err != nil {
```
buqian-zilliz, line 727: **[P2] Preserve retried batched stats uploads for TEXT compaction**

Switching TEXT compaction to `appendV3Stats` also bypasses the configured `BlobsWriter`. In compaction that uploader is `BinlogIO.Upload`, which retries object-store writes under the task context and uploads BM25 blobs concurrently; `appendV3Stats` instead issues one context-free synchronous `packed.WriteFile` per PK/BM25 blob. A transient PUT after the segment writer has closed now forces the whole compaction to rerun, and schemas with multiple BM25 functions serialize every remote round trip. Keep the single manifest commit, but upload the serialized stat blobs through a context-aware retrying/batched path before registering them.

zhagnlu: Agreed. The current change keeps TEXT aligned with the existing PackedManifestRecordWriter behavior and only fixes the missing manifest stats registration. Switching V3 stats uploads to the configured BlobsWriter is a broader behavior change that should apply to both
  PackedManifestRecordWriter and PackedTextManifestRecordWriter, so I will leave this PR scoped and address that in a follow-up.

zhengbuqian (approved):

PR #51672, opened 2026-07-21
enhance: batch binary chunk payload writes

sparknack:
issue: #51673

## What changed

- Write STRING, JSON, and GEOMETRY payloads once per non-empty Arrow batch instead of once per row.
- Reuse precomputed offsets and preserve nullable, sliced-array, and padding layouts.
- Check the uint32 chunk-size limit once after the monotonic offset scan instead of on every row.
- Add sliced multi-batch coverage for all three field types.

## Local microbenchmark

The benchmark harness is not included in this PR. Five-repetition CPU-time means using an in-memory target:

- 100k rows × 16 B, 1/16/256 Arrow batches: 2.03× / 1.94× / 1.84× faster; write calls drop from 100,002 to 3/18/258.
- 100k rows × 128 B, 16 batches: 1.25× faster.
- 10k rows × 1024 B, 16 batches: 1.21× faster.

The host had a high load average and CPU scaling enabled, so these are relative microbenchmark results only; mmap and disk I/O are excluded.

### DIFF
--- a/internal/core/src/common/ChunkWriter.cpp
+++ b/internal/core/src/common/ChunkWriter.cpp
@@ -83,10 +83,6 @@ CalculateBinaryLikeChunkSize(
             const auto relative_offset =
                 static_cast<size_t>(array->value_offset(i) - payload_begin);
             const auto absolute_offset = cursor + relative_offset;
-            AssertInfo(absolute_offset <= std::numeric_limits<uint32_t>::max(),
-                       "{} chunk size {} exceeds uint32 offset limit",
-                       chunk_name,
-                       absolute_offset);
             offsets.push_back(static_cast<uint32_t>(absolute_offset));
         }
 
--- a/internal/core/src/common/ChunkWriter.cpp
+++ b/internal/core/src/common/ChunkWriter.cpp
@@ -786,6 +786,9 @@ create_chunk_buffer(const FieldMeta& field_meta,
             file_path, mmap_populate, aligned_size, io_prio);
     }
     cw->write_to_target(array_vec, target);
+    // The writer is one-shot. Release its scratch buffers before a
+    // file-backed target populates the resulting mmap in release().
+    cw.reset();
     auto data = target->release();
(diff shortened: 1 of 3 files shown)


internal/core/src/common/ChunkWriter.cpp
```diff
@@ -89,61 +152,25 @@ StringChunkWriter::write_to_target(const arrow::ArrayVector& array_vec,
...
-    }
-
-    char padding[MMAP_STRING_PADDING] = {};
-    target->write(padding, MMAP_STRING_PADDING);
-
+    WriteBinaryLikePayload(
+        offsets_, payload_segments_, target, MMAP_STRING_PADDING);
     offsets_.clear();
-    offsets_.shrink_to_fit();
+    payload_segments_.clear();
```
buqian-zilliz, line 158: **[P2] Free one-shot writer buffers before mmap population**

`clear()` leaves the `offsets_` allocation at `row_count + 1` entries. `create_group_chunk()` retains all writers in `cws` until after `target->release()`, so in file-backed mode with `mmap_populate` enabled the full combined mapping is populated while every completed STRING/JSON/GEOMETRY offset table is still resident. Removing the previous capacity release can add roughly 4 bytes per row per variable-length field to peak RSS and can OOM large group loads. Reset each writer after its field is written, or otherwise release these one-shot capacities before releasing the target.

sparknack: Fixed in 7d8c58721e. `create_chunk_buffer()` now resets its one-shot writer before `target->release()`, and `create_group_chunk()` resets each writer immediately after its field is written. This releases offset and payload-segment scratch buffers before file-backed `MAP_POPULATE`. Verified with the relevant 9 chunk tests and `make verifiers`.

zhengbuqian (approved):

xiaofan-luan:
@sparknack maybe we should cherrypick this to 3.0 as well

sparknack:
> @sparknack maybe we should cherrypick this to 3.0 as well

has been cherry-picked.

PR #51674, opened 2026-07-21
test: add pinyin filter end-to-end coverage

zhuwenxing:
## Summary

- Add Pinyin analyzer end-to-end coverage for Python MilvusClient, RESTful v2, and the Go SDK.
- Cover full Pinyin, joined Pinyin, first-letter, and original Chinese matching through Python.
- Add L0 coverage for joined Pinyin and original Chinese matching through RESTful v2 and the Go SDK.
- Cover both `text_match` filtering and BM25 search.
- Construct 4,000 deterministic rows by inserting and flushing 3,000 rows, inserting and flushing 500 rows, creating the index and loading the collection, and finally inserting 500 growing rows without flushing.
- Assert the exact target IDs `{0, 3000, 3500}` without inspecting segment metadata.

issue: #45811

## Test Plan

- [x] Python static checks and collection: 4 cases
- [x] Python MilvusClient E2E: `4 passed in 25.67s`
- [x] RESTful v2 L0 E2E: `1 passed in 28.79s`
- [x] Go SDK compile check with `dynamic,test` tags and disabled optimizations
- [x] Go SDK L0 E2E: `TestPinyinFilterTextMatchAcrossDataPaths` passed in `17.56s`

### DIFF
--- a/tests/go_client/testcases/pinyin_filter_test.go
+++ b/tests/go_client/testcases/pinyin_filter_test.go
@@ -85,18 +85,21 @@ func flushPinyinRows(t *testing.T, ctx CtxT, mc MC, collectionName string) {
 	require.NoError(t, lastErr)
 }
 
-func queryPinyinIDs(t *testing.T, ctx CtxT, mc MC, collectionName, queryText string) []int64 {
-	result, err := mc.Query(ctx, client.NewQueryOption(collectionName).
+func searchPinyinIDs(t *testing.T, ctx CtxT, mc MC, collectionName, queryText string) []int64 {
+	result, err := mc.Search(ctx, client.NewSearchOption(collectionName, pinyinTotalCount,
+		[]entity.Vector{entity.FloatVector{0, 0}}).
+		WithANNSField("vector").
+		WithSearchParam("nprobe", "64").
 		WithFilter(`text_match(text, "`+queryText+`")`).
 		WithOutputFields("id", "text").
-		WithLimit(pinyinTotalCount).
 		WithConsistencyLevel(entity.ClStrong))
 	require.NoError(t, err)
+	require.Len(t, result, 1)
 
-	idColumn := result.GetColumn("id")
+	idColumn := result[0].GetColumn("id")
 	require.NotNil(t, idColumn)
-	ids := make([]int64, result.ResultCount)
-	for i := 0; i < result.ResultCount; i++ {
+	ids := make([]int64, result[0].ResultCount)
+	for i := 0; i < result[0].ResultCount; i++ {
 		value, err := idColumn.Get(i)
 		require.NoError(t, err)
 		ids[i] = value.(int64)
@@ -105,11 +108,11 @@ func queryPinyinIDs(t *testing.T, ctx CtxT, mc MC, collectionName, queryText str
 	return ids
 }
 
-func queryPinyinIDsUntilExpected(t *testing.T, ctx CtxT, mc MC, collectionName, queryText string) []int64 {
+func searchPinyinIDsUntilExpected(t *testing.T, ctx CtxT, mc MC, collectionName, queryText string) []int64 {
 	deadline := time.Now().Add(30 * time.Second)
 	var ids []int64
 	for time.Now().Before(deadline) {
-		ids = queryPinyinIDs(t, ctx, mc, collectionName, queryText)
+		ids = searchPinyinIDs(t, ctx, mc, collectionName, queryText)
 		if len(ids) == len(pinyinTargetIDs) {
 			matched := true
 			for i := range ids {
@@ -180,7 +183,7 @@ func TestPinyinFilterTextMatchAcrossDataPaths(t *testing.T) {
 		pinyinIndexedSealedCount+pinyinUnindexedSealedCount, pinyinGrowingCount)
 
 	for _, queryText := range []string{"zhongwen", "中文"} {
-		ids := queryPinyinIDsUntilExpected(t, ctx, mc, collectionName, queryText)
+		ids := searchPinyinIDsUntilExpected(t, ctx, mc, collectionName, queryText)
 		require.Equal(t, pinyinTargetIDs, ids)
 	}
 }
--- a/tests/go_client/testcases/pinyin_filter_test.go
+++ b/tests/go_client/testcases/pinyin_filter_test.go
@@ -1,0 +1,186 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


yanliang567 (commented):
Review findings:

1. **[P1] The output-mode matrix only verifies positive matches and cannot detect ignored flags.**

   Each parametrized Python case queries only the token for the enabled mode. An implementation that ignores all `false` values and always emits full, joined, and first-letter Pinyin tokens would still pass all three cases.

   The Chinese query does not independently verify `keep_original`, because the query text is analyzed by the same Pinyin analyzer and can match through generated Pinyin tokens even if the original Chinese token is missing.

   Suggested fix: add negative queries for every disabled mode and use `run_analyzer` (or an equivalent analyzer API) to verify the exact token set, including original-token retention.

2. **[P1] The `text_match` cases do not actually exercise the claimed indexed/unindexed search paths.**

   Python, REST, and Go create an IVF index on the vector field, but all assertions use the scalar Query/`text_match` API. Those calls never use the vector index, so the 3,000/500-row threshold does not prove indexed versus brute-force filtering behavior.

   Suggested fix: use vector search with a `text_match` filter so the first sealed segment uses the IVF index while the small sealed and growing segments use their fallback paths. Otherwise, remove the indexed/unindexed coverage claim and simplify the setup.

3. **[P2] Set equality hides duplicate results.**

   The Python and REST helpers accept results such as `[0, 0, 3000, 3500]` because converting to a set still equals `{0, 3000, 3500}`. The Python BM25 assertion has the same gap.

   Suggested fix: assert result cardinality and ID uniqueness in addition to exact set membership.

Static checks, formatting, collection, and Go compile checks passed, and the current CI is green.

zhuwenxing:
Addressed all review feedback:

- Added exact `run_analyzer` token assertions and negative coverage for disabled output modes, including `keep_original=false`.
- Switched the Python, REST, and Go Text Match cases from scalar Query to vector Search with a `text_match` filter, exercising indexed sealed, unindexed sealed, and growing paths.
- Strengthened Python, REST, and BM25 assertions to enforce exact cardinality, uniqueness, and ID sets.

Validation:

- Python Pinyin filter suite: 5 passed
- RESTful v2 L0 case: passed
- Go SDK targeted E2E: passed

tests/go_client/testcases/pinyin_filter_test.go
```diff
@@ -0,0 +1,189 @@
...
+	require.NoError(t, indexTask.Await(ctx))
+
+	loadTask, err := mc.LoadCollection(ctx, client.NewLoadCollectionOption(collectionName))
+	require.NoError(t, err)
+	require.NoError(t, loadTask.Await(ctx))
+
+	insertPinyinRows(t, ctx, mc, collectionName,
+		pinyinIndexedSealedCount+pinyinUnindexedSealedCount, pinyinGrowingCount)
+
+	for _, queryText := range []string{"zhongwen", "中文"} {
```
yanliang567: [P1] The Chinese-query branch still does not prove that `keep_original=true` is honored. `text_match` analyzes the query with the same field analyzer, so `"中文"` is also converted to the joined `zhongwen` token. This test therefore still passes if the original Chinese token is omitted. Please add an exact `RunAnalyzer` token assertion (and ideally a `keep_original=false` negative case) so the Go SDK path verifies original-token retention independently.

tests/restful_client_v2/testcases/test_pinyin_filter.py
```diff
@@ -0,0 +1,187 @@
...
+                "data": build_rows(
+                    INDEXED_SEALED_COUNT + UNINDEXED_SEALED_COUNT,
+                    GROWING_COUNT,
+                ),
+            }
+        )
+        assert rsp["code"] == 0, rsp
+        assert rsp["data"]["insertCount"] == GROWING_COUNT, rsp
+
+        for query_text in ["zhongwen", "中文"]:
```
yanliang567: [P1] The `"中文"` search cannot independently verify `keep_original=true`. The query is processed by the same Pinyin analyzer as the indexed text, so both sides can match through the generated `zhongwen` token even when the original Chinese token was not retained. Please call the REST `/v2/vectordb/common/run_analyzer` endpoint and assert the exact tokens (plus a `keep_original=false` negative if possible).

yanliang567 (commented):
One validation gap remains in the Go and REST L0 cases; details are in the inline comments.

tests/go_client/testcases/pinyin_filter_test.go
zhuwenxing: Addressed in `1744dfa8fc`. The Go L0 case now calls `RunAnalyzer` and asserts the exact token sequence for both configurations:

- `keep_original=true`: `["中文", "zhongwen", "测试", "ceshi"]`
- `keep_original=false`: `["zhongwen", "ceshi"]`

The targeted Go SDK E2E passed, including the existing indexed sealed, unindexed sealed, and growing Search checks.

tests/restful_client_v2/testcases/test_pinyin_filter.py
zhuwenxing: Addressed in `1744dfa8fc`. The REST L0 case now calls `/v2/vectordb/common/run_analyzer` and asserts the exact token sequence for both `keep_original=true` and `keep_original=false`.

The targeted RESTful E2E passed, including the existing indexed sealed, unindexed sealed, and growing Search checks.

yanliang567:
Review follow-up on `e8e620d`:

1. **[P1] `keep_original` is still not verified through the collection field path.**

   The current `run_analyzer` calls pass standalone analyzer parameters instead of reading the analyzer configured on the created field. Also, `text_match(text, "中文")` analyzes the query with the same Pinyin analyzer, so it can match through generated Pinyin tokens even if the original Chinese token was not indexed.

   Please run the analyzer against the created field:
   - Go: `WithField(collectionName, "text")`
   - Python: `analyzer_params=None, collection_name=..., field_name="text"`
   - REST: pass `collectionName` and `fieldName` without standalone `analyzerParams`

   For the Chinese `text_match` assertion, set `minimum_should_match` to the number of tokens produced for that query so every token, including the original Chinese token, must be present.

2. **[P2] Go and REST do not verify disabled output modes through the indexed field.**

   Positive joined-Pinyin searches still pass if the field path ignores `keep_full_pinyin=false`. Please add negative searches for disabled tokens such as `"zhong"` and `"zw"`, expecting no IDs across indexed sealed, unindexed sealed, and growing data.

3. **[P2] The Python BM25 case verifies recall but not BM25 ranking.**

   All three target documents use identical text, and the test only checks the returned ID set. Incorrect score direction or ordering would still pass. Please use deterministic documents with different term frequencies and assert descending BM25 distances and the expected ranking.

The previous vector-filter-path and exact cardinality/uniqueness issues are addressed. Ruff, Python compilation, Go formatting, and the Go package compile check pass on the current head.

zhuwenxing:
Addressed the latest review feedback:

- Updated the Python, Go, and REST analyzer checks to use the analyzer configured on the created field via `collectionName` and `fieldName`.
- Added `minimum_should_match` to the original Chinese `text_match` queries based on the analyzer token count, ensuring that the original Chinese token is required.
- Added negative Go and REST searches for the disabled `zhong` and `zw` tokens across indexed sealed, unindexed sealed, and growing data paths.
- Strengthened the Python BM25 case with different term frequencies and assertions for the exact ranking, strictly descending scores, and returned text.

Validation:

- Python Pinyin output-mode and BM25 cases passed; the updated `keep_original=false` field-analyzer case also passed.
- RESTful v2 targeted E2E passed.
- Go SDK targeted E2E and all four search subtests passed.

PR #51675, opened 2026-07-21
fix: resolve Apple build failures in Conan and go-m1cpu

### LINKED ISSUE #50615: Milvus core fails to build on macOS (Apple Silicon) with LLVM/clang 19

### Environment
- OS: macOS (Apple Silicon, arm64)
- Compiler: Homebrew LLVM / clang 19 (libc++ 19)
- Build: from source (`make`), Conan 2

### Current Behavior
Building the C++ core from source on macOS arm64 with clang 19 fails at several points:

1. **LLVM selection** — `scripts/setenv.sh` only probes up to LLVM 18. libc++ 17/18 has a `std::chrono` `operator<<` that clashes with Arrow's vendored `date` library; the build needs libc++ 19.
2. **folly F14 link error** — Apple Silicon enables `__ARM_FEATURE_CRC32` by default, so core/knowhere compile folly's F14 hash table in SimdAndCrc mode, while 3rdparty folly is built without `+crc`, producing undefined `folly::f14::detail::F14LinkCheck<...>` at link time.
3. **clang 19 stricter template parsing** — `InvertedIndexTantivy.cpp` uses `->template` on non-template members, rejected by clang 19 (`-Wmissing-template-arg-list-after-template-kw`).
4. **bsoncxx → libmongoc → utf8proc** — the BSON layer depends on the bsoncxx C++ driver, whose CMake unconditionally builds mongo-cxx-driver → libmongoc → bundled utf8proc, which fails to configure under newer CMake/clang on macOS. Milvus only needs BSON (de)serialization, not the MongoDB client.
5. **Arrow from-source rapidjson** — with `arrow:encryption=True`, Arrow requires rapidjson; its default (`cci.20220822`) has `GenericStringRef::operator=` assigning a `const` member, which clang 19 rejects. (CI uses a prebuilt Arrow binary so it never hits this; macOS builds Arrow from source and does.)

### Expected Behavior
`make` builds `bin/milvus` cleanly on macOS (Apple Silicon) with clang 19.

### Proposed fix
- Prefer LLVM 19+, add `+crc` for arm64 3rdparty, drop the spurious `->template`.
- Force a clang-19-compatible rapidjson for Arrow's from-source build.
- Replace the bsoncxx C++ driver with **libbson** (C) via a Conan package plus a thin `milvus::bson` shim, dropping libmongoc/mongocxx/utf8proc entirely.

PR incoming.

hicder-rddt:
* Align the Conan build profile with the host profile so protobuf/abseil are built with C++20 (required for std::weak_ordering). 
* Bump go-m1cpu to v0.2.1 for Apple Silicon compatibility

fix #50615

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -242,7 +242,7 @@ require (
 	github.com/sean-/seed v0.0.0-20170313163322-e2103e2c3529 // indirect
 	github.com/shirou/gopsutil v3.20.11+incompatible // indirect
 	github.com/shirou/gopsutil/v3 v3.23.12 // indirect
-	github.com/shoenig/go-m1cpu v0.1.6 // indirect
+	github.com/shoenig/go-m1cpu v0.2.1 // indirect
 	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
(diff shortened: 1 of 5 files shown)


hicder-rddt:
Hi @czs007 , would you mind taking a look at this PR? let me know if there is any improvements that need to be made. thank you!

Issue #51667, opened 2026-07-21
[Bug]: autoID primary keys diverge between primary and secondary when an import is replicated via CDC


bigsheeper:
## Environment

- **Milvus Version**: master (analyzed at commit bf674af67b)
- **Deployment Mode**: cluster, primary-secondary CDC replication (2PC import, `dataCoord.import.enableInReplicatingCluster=true`)
- **MQ**: any (WAL-level replication)
- **SDK**: any

### Option B: Steps

1. Deploy: two clusters with CDC primary -> secondary topology; import files present in both clusters' object storage (2PC import setup, see #51028).
2. Setup: create a collection with an **autoID=true** primary key.
3. Workload: run a 2PC bulk import (`auto_commit=false`) of the same file on the primary; commit after both sides reach `Uncommitted`; wait for both jobs to complete.
4. Trigger: query the same logical rows on both clusters and compare the PK column; then delete some imported rows on the primary (by PK or by filter expression) and wait for replication.
5. Observed via: query results — PK values differ between clusters; deleted rows remain visible on the secondary.

### Trigger Conditions

- **Frequency**: always, for autoID collections imported under CDC replication
- **Does NOT happen when**: the collection has a user-provided (non-autoID) primary key; or rows are written via `insert` instead of import

## Expected Behavior

The secondary holds an identical copy: same logical rows with the **same primary keys**, and subsequent replicated deletes/upserts on imported rows apply on the secondary.

## Actual Behavior

The same logical row gets PK=A on the primary and PK=B on the secondary. Consequences:

1. **Replicated deletes on imported rows silently no-op on the secondary.** A delete for PK=A is routed by hash(A) to a vchannel where the secondary's copy (PK=B, living in vchannel hash(B)) never sees it. Rows deleted on the primary stay visible on the secondary forever.
2. **Monotonic, silent divergence.** Upsert (= delete + insert) leaves both old and new versions on the secondary. No error is raised anywhere; only a full-content diff can detect it.
3. **Failover breaks identity.** PKs fetched from the primary dangle after failover; id fields in query/search results differ between clusters, breaking client-side caching/joins/dedup keyed on PK.
4. Cluster consistency verification (row count / checksum comparison) fails.

## Error Logs

```
No error in logs. Results are silently inconsistent between clusters.
```

## Analysis Hints (Optional)

Root cause: autoID assignment is a **nondeterministic side effect executed independently by both clusters**, because import replicates the *instruction* rather than the *result*:

[...]

PR #51676, opened 2026-07-21
fix: reject autoID import in replicating cluster

bigsheeper:
issue: #51667

This PR rejects non-backup import into autoID primary-key collections when the cluster is currently in CDC replication mode and import in a replicating cluster is enabled.

Why:
- The replicated import path broadcasts the import instruction, including schema/files/options, but not a materialized autoID assignment result.
- Non-backup DataNode import assigns autoID primary keys locally while processing the files, so primary and secondary clusters can assign different PKs for the same logical rows.
- Backup/binlog import remains allowed because it clears autoID on the import schema and preserves primary keys from the imported binlogs.

Verification:
- go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord -run 'TestImportCallbacksSuite/TestValidateImportRequest_(ReplicatingClusterRejectsAutoIDSchema|ReplicatingClusterEnabledRequiresManualCommit)'
- make static-check

### DIFF
--- a/internal/datacoord/ddl_callbacks_import.go
+++ b/internal/datacoord/ddl_callbacks_import.go
@@ -34,6 +34,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
+	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
 
 // importV1AckCallback handles the ack callback for import messages.
@@ -88,7 +89,7 @@ func (c *DDLCallbacks) importV1AckCallback(ctx context.Context, result message.B
 
 // validateImportRequest validates the import request before broadcasting.
 // This includes all validation logic previously done in CheckCallback and Proxy.
-func (s *Server) validateImportRequest(ctx context.Context, files []*msgpb.ImportFile, options []*commonpb.KeyValuePair) error {
+func (s *Server) validateImportRequest(ctx context.Context, files []*msgpb.ImportFile, options []*commonpb.KeyValuePair, schema *schemapb.CollectionSchema) error {
 	// Validate timeout
 	_, err := importutilv2.GetTimeoutTs(options)
 	if err != nil {
@@ -109,14 +110,14 @@ func (s *Server) validateImportRequest(ctx context.Context, files []*msgpb.Impor
 		return err
 	}
 
-	if err := s.validateImportReplication(ctx, options); err != nil {
+	if err := s.validateImportReplication(ctx, options, schema); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (s *Server) validateImportReplication(ctx context.Context, options []*commonpb.KeyValuePair) error {
+func (s *Server) validateImportReplication(ctx context.Context, options []*commonpb.KeyValuePair, schema *schemapb.CollectionSchema) error {
 	balancer, err := balance.GetWithContext(ctx)
 	if err != nil {
 		return err
@@ -137,6 +138,14 @@ func (s *Server) validateImportReplication(ctx context.Context, options []*commo
 	}
 	if importutilv2.IsAutoCommit(options) {
 		return merr.WrapErrOperationNotSupportedMsg("auto_commit=true import in replicating cluster is not supported")
+	}
+	if !importutilv2.IsBackup(options) {
+		for _, field := range schema.GetFields() {
+			if typeutil.IsAutoPKField(field) {
+				return merr.WrapErrOperationNotSupportedMsg(
+					"autoID primary key import in replicating cluster is not supported; primary keys would be assigned independently on each cluster")
+			}
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


czs007:
Adversarial review surfaced three issues worth addressing before merge: a TOCTOU window around the new replication guard, an over-broad rejection that blocks imports which never allocate primary keys locally, and a couple of missing positive test cases.

## High

**`internal/datacoord/ddl_callbacks_import.go:199` — replication-state check is performed outside the broadcaster resource lock (TOCTOU)**

*Problem:* `validateImportRequest` (including the new autoID rejection branch) runs at line 199, before `startBroadcastWithCollectionID` acquires the broadcaster resource keys at line 205. `StartBroadcastWithResourceKeys` implicitly appends a **shared** Cluster resource key (`broadcast_manager.go:201-209`), while `UpdateReplicateConfiguration` holds an **exclusive** Cluster resource key (`assignment.go:97`); the two are mutually exclusive on `resourceKeyLocker` (`resource_key_locker.go:88-100`). So the following interleaving is possible: (t0) `GetLatestChannelAssignment` reports a non-replicating cluster and the guard passes; (t1) a concurrent `AlterReplicateConfig` flips the cluster to replicating and completes its broadcast; (t2) this autoID import is finally broadcast — reproducing exactly the cross-cluster PK divergence this PR is meant to prevent.

*Suggestion:* Move the replication check inside the broadcaster lock (i.e. validate after `startBroadcastWithCollectionID` returns), which is cheap and closes the "alter is mid-broadcast" window. Two things worth noting so the fix isn't over-claimed: (a) the balancer's replicate configuration is back-filled asynchronously via a WAL ack callback (`assignment.go:359` → `balancer.UpdateReplicateConfiguration`), so a residual window remains where the alter has been broadcast but its ack has not yet landed in the local balancer snapshot; (b) this race equally affects the two pre-existing checks (`ImportInReplicatingCluster` / `auto_commit`) — it is inherited structure, not newly introduced by this PR.

*(raised by @chyezh, @tinswzy)*

## Medium

**`internal/datacoord/ddl_callbacks_import.go:142` — over-rejection: `allowInsertAutoID` collections and L0 imports do not allocate PKs locally**

*Problem:* The new branch decides purely on `!IsBackup(options)` plus field-level `IsAutoPKField`, which rejects two legitimate cases that cannot cause cross-cluster PK divergence:

1. **`allowInsertAutoID` collections** (the more material case). When the collection schema property `common.AllowInsertAutoIDKey=true`, import preserves the user-supplied PKs from the files rather than allocating them: `internal/datanode/importv2/util.go:199-200` only generates IDs when `pkField.GetAutoID() && (!ok || pkData == nil || pkData.RowNum() == 0 || !allowInsertAutoID)`, and the parsers keep the user PK as well (`json/row_parser.go:253`, `csv/row_parser.go:288`, `parquet/util.go:143`). PKs come entirely from the files and are identical on source and target clusters, yet this PR rejects them.
2. **L0 imports.** An L0 task only replays delete binlogs; PKs are read out of the binlog by `L0Reader` (`task_l0_import.go:174` → `:213 HashDeleteData` → `:217 syncDelete`) and `AppendSystemFieldsData` is never called, so no autoID is ever allocated. Practical impact is limited: L0 import is gated off by default via `dataCoord.import.enableL0Import=false` (`services.go:1898`), so only clusters that explicitly enable it hit the false rejection.

*Suggestion:* For (1), after `IsAutoPKField` matches, also consult `common.IsAllowInsertAutoID(schema.GetProperties()...)` — the same schema is carried along the proxy → datacoord → datanode chain, so the properties are available here. For (2), either keep the rejection but change the message to state plainly that "L0 import is not supported in a replicating cluster" (the surrounding comment already notes L0 is incompatible with `commit_timestamp` / 2PC), or drop it — but do not justify it with "autoID is allocated independently per cluster", since that reason does not match the actual behavior and would mislead anyone debugging this later.

*(raised by @tinswzy)*

## Low

**`internal/datacoord/ddl_callbacks_import_test.go:244` — missing the backup-exemption and non-autoID pass-through cases**

*Problem:* `TestValidateImportRequest_ReplicatingClusterRejectsAutoIDSchema` (lines 244-301) covers only the rejection path (replicating + flag on + non-backup + autoID). Two positive cases are absent:

1. **Backup exemption.** The PR's central argument is that backup/binlog import preserves the original PKs and is therefore allowed (implemented by `UnsetAutoID` at `internal/datanode/importv2/task_import.go:64-67`), but no `{backup=true} + autoID schema → NoError` test pins that down. If someone later reorders the branches or changes the `IsBackup` predicate, backup restore would be silently rejected with no test failure.
2. **Non-autoID schema passes.** The only existing "allowed" assertion is in `TestValidateImportRequest_ReplicatingClusterEnabledRequiresManualCommit` (lines 237-241), which passes `schema=nil` and therefore exercises the degenerate empty-`GetFields()` path — it never proves that a real schema with a non-autoID PK field is accepted.

*Suggestion:* Add two short cases: `{backup=true} + autoID schema → NoError`, and a schema with `IsPrimaryKey=true, AutoID=false` → `NoError`.

*(raised by @tinswzy)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/fefa2c241e3f65bd893da6438a5b2e93d429d280...czs007:team2-fix/Ted13243f08?expand=1

PR #50992, opened 2026-07-02
enhance: amortize per-row chunk resolution in ProxyChunkColumn bulk access

xiaofan-luan:
issue: #50991

### What this PR does

Every `ProxyChunkColumn` `Bulk*At` accessor resolved the field chunk per row: an `unordered_map` lookup plus a `shared_ptr<Chunk>` copy (two atomic refcount ops) per element, with rows visited in random offset order. These accessors are the search-reduce hot path (`FillPrimaryKeys` / `FillTargetEntry`), and under concurrent search the per-row refcount bumps hit the same chunk control blocks from every core.

- Reorder-safe accessors (`BulkPrimitiveValueAt`, `BulkVectorValueAt`, `BulkRawStringAt`, `BulkRawJsonAt`, `BulkArrayAt`, `BulkVectorArrayAt`) now visit rows grouped by chunk via a new `ForEachRowGroupedByChunk` helper — an O(count) counting pass over chunk ids; the chunk is resolved once per chunk. Callbacks receive the original row index, and all call sites address output by that index (audited all 22 call sites).
- `BulkValueAt` (its sparse-vector caller appends in invocation order) and the nullable `BulkIsValid` path (its caller builds `valid_offsets` in invocation order) keep caller order and only hoist the chunk resolution behind a last-chunk cache.
- Fix `BulkIsValid` missing an early return for non-nullable fields, which invoked the callback twice per row (covered by a new regression test that fails on master).

A comparison sort by offset was evaluated and rejected: it measured ~4x slower than the whole access loop, and with count ≪ rows-per-chunk sorted offsets share no cache lines anyway, so counting-bucket grouping keeps all of the win at O(count) cost.

### Benchmarks

`ProxyChunkColumnBulkAccessTest.cpp` (INT64 column, 10k random offsets per call, fresh offsets per iteration; Apple M-series):

| scenario | master | this PR |
|---|---|---|
| 4.2M-row group (64 chunks × 65536 × 8 fields), single thread | 18.8 ns/row | 12.7 ns/row (−33%) |
| same column, 8 concurrent threads | 10.5 ns/row agg | 2.6 ns/row agg (~4x) |
| small cache-resident configs (4–128 chunks) | 6.4–10.2 ns/row | 6.2–8.7 ns/row |

[...]

Issue #50991, opened 2026-07-02
[Enhancement]: ProxyChunkColumn bulk accessors pay a hash lookup and shared_ptr refcount per row


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Amortize per-row chunk resolution in `ProxyChunkColumn` (storage v2) bulk accessors: group rows by chunk with an O(count) counting pass so the field chunk is resolved once per chunk instead of once per row.

### Why is this needed?

Every `Bulk*At` accessor in `internal/core/src/mmap/ChunkedColumnGroup.h` resolves the field chunk per row: one `unordered_map` lookup plus a `shared_ptr<Chunk>` copy (two atomic refcount ops) per element. These accessors sit on the search reduce hot path (`FillPrimaryKeys` touches nq × topk × num_segments candidate rows per request, `FillTargetEntry` fills output fields), plus retrieve and delete-by-filter PK resolution.

Under concurrent search load the per-row refcount bumps hit the same chunk control blocks from every core, causing cache line ping-pong. Production flame graphs on a binary-vector workload (Milvus 2.6, querynode) show `ProxyChunkColumn` frames dominating `ReduceSearchResultsAndFillData`, wider than the vector distance computation itself.

Microbenchmark (INT64 column, 10k random offsets per call, 4.2M-row column group): single-threaded 18.8 → 12.7 ns/row; 8 threads on the same column 10.5 → 2.6 ns/row aggregate (~4x).

### Anything else?

Follow-ups after this lands: make the sparse-vector `BulkValueAt` caller and `FilterVectorValidOffsetsFromColumn` order-independent so those paths can also be grouped; share the grouping across output fields in `FillTargetEntry`; backport to 2.6.

PR #51678, opened 2026-07-21
enhance: amortize per-row chunk resolution in storage-v2 bulk access and expr offset paths

xiaofan-luan:
## Summary

issue: #50991

Storage-v2 hot paths resolved the chunk that owns a row **once per row** even when
thousands of score-ordered offsets map to only a few distinct chunks. Two layers had
the problem:

1. **Column layer** — `ProxyChunkColumn` bulk accessors: `GroupChunk::GetChunk(field_id_)`
   returns a `shared_ptr<Chunk>` by value, so every row paid a field-map lookup plus an
   atomic refcount bump. Under concurrent search the per-row atomic RMW on shared chunk
   control blocks contends across cores.
2. **Expression layer** — the by-offsets eval paths re-pinned a fresh chunk *and*
   re-resolved it per row (the string/json/array branch heap-allocated a one-element view
   vector per row); the offset-input paths in `PhyColumnExpr` / `PhyCompareFilterExpr`
   rebuilt a `ChunkDataAccessor` (re-pinning the chunk) every row.

## Changes (single commit)

**Column layer (`ChunkedColumnGroup.h`, `GroupChunk.h`)**

- New `GroupChunk::GetChunkRaw(field_id) -> Chunk*` **borrows** the chunk (no shared
  ownership, no atomic; valid while the pinned `CellAccessor` keeps the GroupChunk
  resident).
- All per-row bulk accessors route through one helper, `ForEachResolvedRow`, which pins
  the involved chunks once, then **dedups resolution per distinct chunk id** via a
  last-cid fast path + a hash map reserved to `min(count, num_chunks)` — allocation is
  proportional to the *request*, never to the segment's total chunk count, so a small
  top-K read on a large segment stays cheap (addresses the earlier review comment).
- `BulkIsValid` now honors the interface contract's `offsets == nullptr` full-scan mode
  on the nullable path (previously silently returned; storage-v1 and the non-nullable
  branch already honored it).

**Expression layer (`Expr.h`, `CompareExpr.h`, `ColumnExpr.cpp`, `CompareExpr.cpp`)**

- `SegmentExpr::ProcessDataByOffsets` fixed-width branch (sealed **and** growing): keep
  the pinned chunk across iterations, re-pin only when the chunk id changes; the
  SkipIndex decision is evaluated once per chunk instead of once per row.
- `SegmentExpr::ProcessDataByOffsets` string/json/array branch: group runs of
  consecutive same-chunk offsets and fetch each run with one `get_views_by_offsets`
  call (views bound by reference — no per-run copies), eliminating the per-row pin +
  per-row one-element vector allocation. Per-row callback invocation and row order
  unchanged.
- `PhyCompareFilterExpr::ProcessBothDataByOffsets`: cache both left and right pinned
  chunks, re-pin each only on its chunk-id change.
- `PhyColumnExpr::DoEval` and the `PhyCompareFilterExpr` offset-input path: cache the
  `ChunkDataAccessor` per chunk id instead of rebuilding (and re-pinning) it every row;
  the chunk-independent pinned-index view is resolved once.

**Segcore (`ChunkedSegmentSealedImpl.cpp`)**

- `bulk_subscript` nullable validity uses the batched `BulkIsValid` (pins the involved
  chunks once) instead of per-row `IsValid` (which pins a cell on every call).
- Documented the `ReadTimestamp` StorageV2 fallback's per-row-pin access pattern (see
  Follow-ups).

**Growing validity lookups (`ConcurrentVector.h`, `SegmentGrowingImpl.cpp`)**

- New `ThreadSafeValidData::bulk_is_valid(offsets, count, out)`: answers `count`
  offsets under one lock acquisition (out-of-range yields `false`, matching the
  loop it replaces). `FilterVectorValidOffsets` (growing `bulk_subscript`,
  nullable vector fields) no longer copies the whole validity bitmap per query —
  previously an O(segment rows) `get_valid_data()` copy used only for `count`
  lookups (and, on the mapping-storage branch, an emptiness check now served by
  `empty()`). `get_data()` is documented as a whole-bitmap copy to avoid on hot
  paths.

**Enabling change for growing segments (`ConcurrentVector.h`, `InsertRecord.h`)**

`ThreadSafeValidData` (the growing nullable validity bitmap) previously stored one flat
`FixedVector<bool>` that could **reallocate on a concurrent insert**, so a validity
pointer cached across rows could dangle (a pre-existing one-row race that cross-row
caching would have widened). It now stores fixed-size per-chunk buffers in a deque
(chunk size = `size_per_chunk`, threaded through `append_valid_data`): appends only add
new buffers and never move existing ones, so a borrowed validity pointer stays valid for
reads within its chunk. All raw-pointer readers were audited to read within a single
chunk (`ApplyFieldValidData` callers, `InsertRecord::get_span_base`, expr span
consumers). This is what makes the cross-row caching above safe on growing segments —
the root fix requested in review, instead of gating growing out of the optimization.

Storage v1 `ChunkedColumn` needed no change: its `get_cell_of` already returns a raw
`Chunk*` with no shared_ptr or per-row field-map lookup.

## Performance

Qualitative, by mechanism:
- **Concurrent search** (primary target): the per-row shared_ptr refcount atomic on
  shared chunk control blocks is gone (`GetChunk` → borrowed `GetChunkRaw`), removing
  cross-core contention on hot segments.
- **Single-thread**: chunk resolution (a cache-cold map lookup on large segments) runs
  once per distinct chunk instead of once per row; the variable-length expr path
  additionally saves a per-row heap allocation.
- Microbenchmarks on an earlier prototype of this change (same borrow + once-per-distinct-
  chunk dedup semantics; cache structure since reworked per review) showed ~1.5–1.7×
  single-thread and ~5× 8-thread concurrent on 10k random-offset reads, with no
  regression at count=1. Not re-measured on the final form; treat as indicative only.

## Testing

- Covered by the existing mmap / expression / StorageV2 regression suites and CI; no new
  test files in this PR.
- The `ThreadSafeValidData` rewrite preserves the public interface (`set_data_raw` /
  `is_valid` / `get_chunk_data` / `get_data`, plus a new `empty()`), so existing nullable
  growing-segment suites exercise it directly.

## Notes for reviewers

- `folly::F14FastMap` was tried for the dedup cache and reverted: its hash requires
  SSE4.2 crc32 (`_mm_crc32_u64`) and fails to compile under the core build's runtime
  SIMD dispatch (no `-msse4.2` on these TUs); `std::unordered_map` reserved to
  `min(count, num_chunks)` is used instead.

## Follow-ups (separate PRs)

- `ReadTimestamp`'s StorageV2 fallback re-pins per row, but is effectively unreachable
  in production (every v2 load path materializes a fully-pinned zero-copy
  `TimestampData`); a NOTE in code documents the batched design it needs if the
  timestamp column ever becomes evictable.
- The `VectorArrayView` by-offsets branch still resolves per row — needs a batch
  by-offsets column API first.
- Cleanups: lazy-create the dedup map in `ForEachResolvedRow` (skip the allocation for
  single-chunk requests); unify the sealed/growing fixed-width loops and the duplicated
  offset→(chunk, offset) resolution lambdas behind one helper.

Supersedes #50992 (closed; predated master's `ToChunkIdAndOffset`/`GetGroupChunks`
batch-resolve, and its bundled `BulkIsValid` fix already merged as #51598).

https://claude.ai/code/session_01Md98mTWZSM9xDmLjeVYdvh

### DIFF
--- a/internal/core/src/common/GroupChunk.h
+++ b/internal/core/src/common/GroupChunk.h
@@ -44,6 +44,22 @@ class GroupChunk {
         return it->second;
     }
 
+    // Borrow the chunk for a field WITHOUT sharing ownership. The returned
+    // pointer is valid only while this GroupChunk stays pinned/resident (e.g.
+    // for the lifetime of the CellAccessor that pinned it). Bulk accessors use
+    // this on the hot path to avoid the per-row shared_ptr refcount bump that
+    // GetChunk() incurs — an atomic RMW on a shared control block that contends
+    // across concurrent search threads. For anything that must outlive the pin,
+    // use GetChunk() instead.
+    Chunk*
+    GetChunkRaw(FieldId field_id) const {
+        auto it = chunks_.find(field_id);
+        if (it == chunks_.end()) {
+            return nullptr;
+        }
+        return it->second.get();
+    }
+
     // Add a chunk for a specific field
     void
     AddChunk(FieldId field_id, std::shared_ptr<Chunk> chunk) {
--- a/internal/core/src/exec/expression/ColumnExpr.cpp
+++ b/internal/core/src/exec/expression/ColumnExpr.cpp
@@ -99,6 +99,15 @@ PhyColumnExpr::DoEval(OffsetVector* input) {
         valid_res.set();
 
         int64_t processed_rows = 0;
+        // Keep the chunk's data accessor (which pins the chunk) across
+        // iterations and rebuild only when the chunk id changes, avoiding a
+        // per-row chunk pin + accessor construction. Safe on both sealed
+        // (CellAccessor keeps the chunk resident) and growing (data and the
+        // chunked validity storage have stable per-chunk buffers). The pinned
+        // index view is chunk-independent, so it is resolved once.
+        const auto pinned_index = PinnedIndexForRawLookup();
+        int64_t cached_chunk_id = -1;
+        segcore::ChunkDataAccessor cda;
         for (auto i = 0; i < real_batch_size; ++i) {
             auto offset = (*input)[i];
             auto [chunk_id,
@@ -117,11 +126,14 @@ PhyColumnExpr::DoEval(OffsetVector* input) {
                     return {0, offset};
                 }
             }();
-            auto cda = segment_chunk_reader_.GetChunkDataAccessor(
-                expr_->GetColumn().data_type_,
-                expr_->GetColumn().field_id_,
-                chunk_id,
-                PinnedIndexForRawLookup());
+            if (chunk_id != cached_chunk_id) {
+                cda = segment_chunk_reader_.GetChunkDataAccessor(
+                    expr_->GetColumn().data_type_,
+                    expr_->GetColumn().field_id_,
+                    chunk_id,
+                    pinned_index);
+                cached_chunk_id = chunk_id;
+            }
             auto chunk_data_by_offset = cda(chunk_offset);
             if (!chunk_data_by_offset.has_value()) {
                 valid_res[processed_rows] = false;
--- a/internal/core/src/exec/expression/CompareExpr.cpp
+++ b/internal/core/src/exec/expression/CompareExpr.cpp
@@ -111,37 +111,55 @@ PhyCompareFilterExpr::ExecCompareExprDispatcher(OpType op, EvalCtx& context) {
 
         int64_t processed_rows = 0;
         const auto size_per_chunk = segment_chunk_reader_.SizePerChunk();
+        auto get_chunk_id_and_offset =
+            [&](const FieldId field,
+                const int64_t raw_data_chunk_count,
+                int64_t offset) -> std::pair<int64_t, int64_t> {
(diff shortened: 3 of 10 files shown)


xiaofan-luan:
ci-v2/e2e-default failures on this PR are a master-level CI blocker, not introduced here: `test_struct_array_compaction_preserves_null_empty_offsets_and_indexes` (added in #51524, first exposure on the distributed-pulsar profile) deterministically fails there with `compact_id == -1` — reproduced 3/3 across this PR and unrelated #51540 on independent clusters. Details and evidence: #51686. This PR touches only read paths (bulk access / expr eval); all struct-array read e2e cases in the same runs passed.

liliu-z (commented):
/lgtm
/approve

liliu-z:
Follow-up for the pre-existing `reserved_offset` issue: #51842, fixed by #51843.

The follow-up preserves this PRs fixed-chunk, pointer-stable validity storage and adds offset-aware writes for out-of-order growing insert completion.

Issue #51681, opened 2026-07-22
[Enhancement]: Support nullable StructArray in Go SDK and REST v2


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Milvus StructArray fields can be nullable, and pymilvus supports `None` values at the parent-row level. Add equivalent support to the Go SDK and REST v2 so clients can insert and read null StructArray rows without losing row alignment.

The enhancement should:

- add `Nullable`, `SetNullable`, and `AppendNull` behavior for Go SDK StructArray columns
- support both scalar and vector StructArray sub-fields
- preserve a shared validity mask through append, slice, protobuf encoding and decoding, query, and search
- handle compact nullable payloads and zero-row query/search responses
- accept `null` parent StructArray values through REST v2 insert
- return `null` parent values through REST v2 query and search
- reject null individual sub-field values inside a non-null struct row

### Why is this needed?

The server schema supports nullable StructArray fields, but the Go SDK currently rejects nullable operations and REST v2 does not provide complete nullable-row handling. This is inconsistent with pymilvus and prevents Go SDK and REST users from using the server capability.

Incorrect or missing validity-mask handling can also shift compact payload rows after a null value, causing subsequent rows to be decoded incorrectly.

### Anything else?

A null StructArray row must remain distinct from an empty StructArray value (`[]`) across insert, query, and search.

Acceptance criteria:

- Go SDK nullable StructArray values round-trip through insert, query, and search
- REST v2 nullable StructArray values round-trip through insert, query, and search
- null and empty StructArray values remain distinguishable
- requested StructArray columns are retained in zero-row responses
- unit and end-to-end coverage includes scalar and vector sub-fields

zhuwenxing:
fixed

PR #51682, opened 2026-07-22
enhance: support nullable struct arrays in Go SDK and REST v2

zhuwenxing:
issue: #51681

## What this PR does

- adds nullable row support to Go SDK StructArray columns, including scalar and vector sub-fields
- preserves null masks across append, slice, compaction, protobuf parsing, query, and search results
- restores requested StructArray columns for zero-row query and search responses
- preserves empty dynamic-field query behavior
- adds REST v2 nullable StructArray insert, query, and search support
- rejects per-sub-field null values in non-null struct rows
- isolates sliced scalar and vector array storage from source-column mutation while preserving nil versus empty-slice semantics
- validates FieldData slice ranges and decodes batched search fields once before per-query slicing

## Why

StructArray fields already support nullable parent rows on the server and in pymilvus. This change makes the Go SDK and REST v2 behavior consistent with that capability.

## Tests

- Go SDK column and Milvus client unit tests
- REST v2 HTTP server unit tests
- Go SDK nullable StructArray end-to-end test
- REST v2 nullable StructArray end-to-end test without FLAT-index dependency

### DIFF
--- a/client/column/columns.go
+++ b/client/column/columns.go
@@ -180,18 +180,8 @@ func parseStructArrayData(fieldName string, structArray *schemapb.StructArrayFie
 
 // parseVectorArrayData converts schemapb.VectorArray (per-row list of vectors) into the
 // matching ColumnXxxVectorArray. Used for ArrayOfVector sub-fields of struct arrays.
-func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end int) (Column, error) {
+func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, validData []bool, begin, end int) (Column, error) {
 	rows := va.GetData()
-	if end < 0 || end > len(rows) {
-		end = len(rows)
-	}
-	if begin < 0 {
-		begin = 0
-	}
-	if begin > end {
-		begin = end
-	}
-	rows = rows[begin:end]
 	// VectorArray.Dim may be 0 in server search responses; fall back to inner VectorField.Dim.
 	dim := int(va.GetDim())
 	if dim == 0 {
@@ -206,6 +196,64 @@ func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end
 		return nil, fmt.Errorf("vector array %q has unknown dim", fieldName)
 	}
 
+	nullable := validData != nil
+	sparseMode := false
+	logicalLen := len(rows)
+	if nullable {
+		logicalLen = len(validData)
+		switch len(rows) {
+		case logicalLen:
+			// Query results are row-dense and keep an empty placeholder for null rows.
+			sparseMode = true
+		case countValid(validData):
+			// Insert FieldData keeps only valid payload rows.
+			sparseMode = false
+		default:
+			return nil, fmt.Errorf("vector array %q payload row count %d does not match logical row count %d or valid count %d",
+				fieldName, len(rows), logicalLen, countValid(validData))
+		}
+	}
+	if begin < 0 {
+		begin = 0
+	}
+	if begin > logicalLen {
+		begin = logicalLen
+	}
+	if end < 0 || end > logicalLen {
+		end = logicalLen
+	}
+	if begin > end {
+		begin = end
+	}
+	selectedValidData := validData
+	if nullable {
+		selectedValidData = validData[begin:end]
+		if sparseMode {
+			rows = rows[begin:end]
+		} else {
+			valueBegin := countValid(validData[:begin])
+			valueEnd := valueBegin + countValid(selectedValidData)
+			rows = rows[valueBegin:valueEnd]
+		}
+	} else {
+		rows = rows[begin:end]
+	}
+	finish := func(column Column) (Column, error) {
+		if !nullable {
+			return column, nil
+		}
+		vectorColumn, ok := column.(interface {
+			setNullableData([]bool, bool) error
+		})
+		if !ok {
+			return nil, fmt.Errorf("vector array %q column does not support nullable data", fieldName)
+		}
+		if err := vectorColumn.setNullableData(selectedValidData, sparseMode); err != nil {
+			return nil, errors.Wrapf(err, "invalid vector array %q nullable data", fieldName)
+		}
+		return column, nil
+	}
+
 	switch va.GetElementType() {
 	case schemapb.DataType_FloatVector:
 		out, err := splitVectorArrayRows(fieldName, rows, dim, func(vf *schemapb.VectorField) []float32 {
@@ -218,7 +266,7 @@ func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end
 		if err != nil {
 			return nil, err
 		}
-		return NewColumnFloatVectorArray(fieldName, dim, out), nil
+		return finish(NewColumnFloatVectorArray(fieldName, dim, out))
 
 	case schemapb.DataType_Float16Vector:
 		out, err := splitVectorArrayRows(fieldName, rows, dim*2, func(vf *schemapb.VectorField) []byte {
@@ -227,7 +275,7 @@ func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end
 		if err != nil {
 			return nil, err
 		}
-		return NewColumnFloat16VectorArray(fieldName, dim, out), nil
+		return finish(NewColumnFloat16VectorArray(fieldName, dim, out))
 
 	case schemapb.DataType_BFloat16Vector:
 		out, err := splitVectorArrayRows(fieldName, rows, dim*2, func(vf *schemapb.VectorField) []byte {
@@ -236,7 +284,7 @@ func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end
 		if err != nil {
 			return nil, err
 		}
-		return NewColumnBFloat16VectorArray(fieldName, dim, out), nil
+		return finish(NewColumnBFloat16VectorArray(fieldName, dim, out))
 
 	case schemapb.DataType_BinaryVector:
 		if dim%8 != 0 {
@@ -248,7 +296,7 @@ func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end
 		if err != nil {
 			return nil, err
 		}
-		return NewColumnBinaryVectorArray(fieldName, dim, out), nil
+		return finish(NewColumnBinaryVectorArray(fieldName, dim, out))
 
 	case schemapb.DataType_Int8Vector:
 		out, err := splitVectorArrayRows(fieldName, rows, dim, func(vf *schemapb.VectorField) []byte {
@@ -263,7 +311,7 @@ func parseVectorArrayData(fieldName string, va *schemapb.VectorArray, begin, end
 		if err != nil {
 			return nil, err
 		}
-		return NewColumnInt8VectorArray(fieldName, dim, out), nil
+		return finish(NewColumnInt8VectorArray(fieldName, dim, out))
 
 	default:
 		return nil, fmt.Errorf("unsupported vector array element type %s", va.GetElementType())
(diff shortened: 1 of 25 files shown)


client/column/vector_array.go
```diff
@@ -85,46 +101,131 @@ func (c *columnVectorArrayBase[T]) GetAsBool(_ int) (bool, error) {
...
+}
 
-func (c *columnVectorArrayBase[T]) ValidateNullable() error { return nil }
+func (c *columnVectorArrayBase[T]) CompactNullableValues() {
+	if !c.nullable || !c.sparseMode {
+		return
+	}
 
-func (c *columnVectorArrayBase[T]) CompactNullableValues() {}
+	values := c.values[:0]
```
buqian-zilliz, line 165: **[P2] Preserve source columns when compacting sliced vector arrays**

`Slice` keeps `c.values` as a view of the original outer slice, but the write path calls `CompactNullableValues` in place before building an Insert or Upsert request. For a nullable slice whose backing values are `[nil, A, B]` and validity is `[false, true, true]`, compaction rewrites that shared array to `[A, B, B]`; the original column still interprets it with the sparse mask, so its second logical value changes from `A` to `B`. Merely using a sliced column in a write request can therefore silently mutate the caller's source column. Allocate independent outer storage before compaction, or make `Slice` copy the outer values slice.

client/column/columns.go
```diff
@@ -94,16 +94,54 @@ func parseScalarData[T any, COL Column, NCOL Column](
...
-		validData = validData[start:end]
-		ncol, err := nullableCreator(name, data, validData, WithSparseNullableMode[T](true))
-		return ncol, err
+	if start < 0 {
+		start = 0
+	}
+	if start > logicalLen {
+		start = logicalLen
+	}
+	if end < 0 || end > logicalLen {
```
buqian-zilliz: **[P2] Reject truncated fields instead of clamping bounded ranges**

`parseFieldData` clamps every `end > logicalLen` to `logicalLen` instead of rejecting a bounded range. `handleSearchResult` retains `ResultCount` from `Topks` and requests `[offset, offset+rc)`, so truncated FieldData can produce a ResultSet whose columns are shorter than its advertised `rc` rows. Iterating the advertised hits then loses values or encounters an out-of-range read even though the malformed response was accepted. Only the `end == -1` sentinel should mean "to the end"; reject other out-of-bounds ranges, including in the vector-array path.

client/column/columns.go
```diff
@@ -94,16 +94,54 @@ func parseScalarData[T any, COL Column, NCOL Column](
...
+		end = logicalLen
+	}
+	if start > end {
+		start = end
+	}
+
+	if validData != nil {
+		validCount := countValid(validData)
```
buqian-zilliz: **[P2] Avoid quadratic validity scans in batched search decoding**

`parseScalarData` scans the complete `validData` bitmap to compute `validCount` before testing the common row-dense representation (`dataLen == logicalLen`). Search decoding invokes this parser once per query over the same complete FieldData, so each nullable scalar or array field rescans all `nq * topK` validity bits `nq` times. At fixed top-K, this becomes O(nq² * topK) work and can dominate large batch searches. Detect dense payloads before scanning, and for compact payloads compute validity prefixes once or parse the complete column once and slice it.

client/column/generic_base.go
```diff
@@ -94,28 +94,58 @@ func (c *genericColumnBase[T]) Slice(start, end int) Column {
...
 	if end == -1 || end > l {
 		end = l
 	}
+	if start > end {
+		start = end
+	}
+
+	valueStart, valueEnd := start, end
+	if c.nullable && !c.sparseMode {
+		valueStart, valueEnd = compactValueRange(c.indexMapping, start, end)
```
buqian-zilliz, line 116: **[P1] Initialize compact indexes before slicing nullable vectors**

The new compact-slice path assumes `indexMapping` is populated, but the public `NewNullableColumnFloatVector`, `NewNullableColumnBinaryVector`, `NewNullableColumnFloat16Vector`, `NewNullableColumnBFloat16Vector`, `NewNullableColumnInt8Vector`, and `NewNullableColumnSparseFloatVector` constructors only set `validData` and `nullable`. A non-empty column returned successfully by any of these constructors now panics on `Slice(0, col.Len())` at `indexMapping[start:end]`. Build the compact mapping in those constructors or validate lazily before accessing it here.

client/milvusclient/read.go
```diff
@@ -112,17 +137,41 @@ func (c *Client) handleSearchResult(schema *entity.Schema, outputFields []string
...
+			}
 		}()
 	}
 	return sr, nil
 }
 
+func sliceSearchResultColumns(columns []column.Column, start, end int) []column.Column {
+	sliced := make([]column.Column, len(columns))
+	for idx, field := range columns {
+		sliced[idx] = field.Slice(start, end)
```
buqian-zilliz: **[P1] Preserve nullable Geometry rows in batched slicing**

The batched path now slices every decoded column, but `ColumnGeometryWKT.Slice` clamps the requested logical range to `len(c.values)`, which is only the compact physical value count for a nullable column. With validity `[false, true]`, slicing the full two-row result is reduced to `[0,1)` and drops the valid geometry while `ResultCount` remains 2. Make Geometry's `Len` and `Slice` nullable-aware, or exclude it from this whole-result slicing path.

client/column/columns.go
```diff
@@ -366,7 +473,7 @@ func FieldDataColumn(fd *schemapb.FieldData, begin, end int) (Column, error) {
 		if va == nil {
 			return nil, errFieldDataTypeNotMatch
 		}
-		return parseVectorArrayData(fd.GetFieldName(), va, begin, end)
+		return parseVectorArrayData(fd.GetFieldName(), va, validData, begin, end)
```
buqian-zilliz: **[P1] Use the outer dimension for all-null vector arrays**

`parseVectorArrayData` rejects `VectorArray.Dim == 0` before considering the validity mask, and this call discards `vectors.GetDim()`. A valid compact nullable ArrayOfVector payload with every row null has no inner rows from which to recover a dimension, so it fails with `unknown dim` even when the outer `VectorField` carries the schema dimension. Pass the outer dimension into the parser as a fallback.

client/milvusclient/read.go
```diff
@@ -170,6 +254,32 @@ func (c *Client) parseSearchResult(sch *entity.Schema, outputFields []string, fi
...
+			if field == nil || field.DataType != entity.FieldTypeArray || field.ElementType != entity.FieldTypeStruct {
+				continue
+			}
+			col, err := newEmptyStructArrayColumn(field, structOutputSelections[parentName])
+			if err != nil {
+				return nil, err
+			}
+			columns = append(columns, col)
+		}
+		if sch.EnableDynamicField && len(dynamicNames) > 0 {
```
buqian-zilliz: **[P2] Restore `$meta` in zero-row dynamic results**

`$meta` is injected into `sch.Fields`, so it is excluded from `dynamicNames`; this branch therefore synthesizes nothing for an explicit `$meta` or wildcard request when `fieldDataList` is empty. Non-empty results do include that column, so zero-hit queries and searches unexpectedly change shape and `GetColumn("$meta")` returns nil. Build the empty dynamic column from the schema's `IsDynamic` field, preserve its name and nullable state, append it when requested, and reuse it for dynamic-key wrappers.

client/column/dynamic.go
```diff
@@ -38,6 +38,13 @@ func (c *ColumnDynamic) Name() string {
 	return c.outputField
 }
 
+func (c *ColumnDynamic) Slice(start, end int) Column {
+	return NewColumnDynamic(
+		c.ColumnJSONBytes.Slice(start, end).(*ColumnJSONBytes),
```
buqian-zilliz, line 43: **[P2] Avoid re-copying the dynamic JSON base per projection**

All projected dynamic columns share one `ColumnJSONBytes`, but the batched search path calls this method independently for every projection. Each call copies the same `[][]byte` row headers and validity mask through `genericColumnBase.slice`, making memory and copy work O(dynamic fields × result rows) instead of slicing one shared base per query. Slice the backing JSON column once and reuse it for all wrappers with the same source and bounds.

zhengbuqian (approved):

client/milvusclient/read.go
```diff
@@ -213,7 +364,11 @@ func (c *Client) Query(ctx context.Context, option QueryOption, callOptions ...g
 			return err
 		}
 
-		columns, err := c.parseSearchResult(collection.Schema, resp.GetOutputFields(), resp.GetFieldsData(), 0, 0, -1)
+		outputFields := resp.GetOutputFields()
+		if len(outputFields) == 0 {
```
buqian-zilliz, line 368: **[P1] Preserve requested dynamic projections after response-field reconstruction**

`resp.OutputFields` is not a reliable copy of the user's projection once the schema contains a StructArray field: the proxy reconstructs it from returned `FieldData` names, so a requested dynamic key such as `color` becomes the backing `$meta` field. Because this fallback consults `req.GetOutputFields()` only when the response list is empty, `parseSearchResult` returns the raw JSON column instead of `ColumnDynamic("color")`, including for zero-row queries this change intends to restore. Preserve or merge the request's dynamic projections instead of treating every non-empty response list as authoritative.

client/column/struct.go
```diff
@@ -153,25 +181,84 @@ func (c *columnStructArray) GetAsBool(idx int) (bool, error) {
...
 
 func (c *columnStructArray) IsNull(idx int) (bool, error) {
-	return false, nil
+	if idx < 0 || idx >= c.Len() {
+		return false, errors.Newf("index %d out of range[0, %d)", idx, c.Len())
+	}
+	if !c.nullable {
+		return false, nil
+	}
+	return c.fields[0].IsNull(idx)
```
buqian-zilliz, line 190: **[P1] Reject mismatched StructArray null masks before reading rows**

`NewColumnStructArray` now marks the parent nullable when any child is nullable, but it does not verify that all children share the same validity mask, while `IsNull` trusts only `fields[0]`. Two individually valid vector-array children with masks `[true, false]` and `[false, true]` therefore construct successfully: row 0 is returned with one sub-field unexpectedly nil, and row 1 is returned as an entirely null struct, hiding the second child's valid value. `ValidateNullable` detects this, but neither construction nor the write path invokes it. Validate the cross-child mask invariant before exposing or serializing the column.

client/column/dynamic.go
```diff
@@ -38,6 +38,39 @@ func (c *ColumnDynamic) Name() string {
...
```
buqian-zilliz, line 43: **[P2] Use sharing-aware dynamic slicing in ResultSet.Slice**

`ResultSet.Slice` still invokes `Column.Slice` independently for every field. When multiple projected dynamic columns share one `ColumnJSONBytes`, this method clones the same row headers and validity mask once per projection and returns wrappers that no longer share their base, even though the batched search path now uses `SliceColumns`. Route `ResultSet.Slice` through `SliceColumns` as well so public and iterator slicing preserve the shared-base invariant and avoid O(projections × rows) copying.

client/column/columns.go
```diff
@@ -94,19 +94,81 @@ func parseScalarData[T any, COL Column, NCOL Column](
...
+			data = data[valueStart:valueEnd]
+		}
+		ncol, err := nullableCreator(name, data, selectedValidData, WithSparseNullableMode[T](sparseMode))
+		if err != nil {
+			return nil, err
+		}
+		// An empty nullable slice has no validity bits, but it must retain the
+		// nullable schema state for its parent struct array.
+		ncol.SetNullable(true)
+		return ncol, ncol.ValidateNullable()
```
buqian-zilliz, line 131: **[P3] Avoid repeatedly rebuilding compact nullable indexes**

`nullableCreator` already calls `ValidateNullable`, which allocates and builds `indexMapping` for compact payloads. For a non-empty validity mask, `SetNullable(true)` is a no-op and this second `ValidateNullable` immediately reallocates and rescans the same mask; `parseStructArrayData` then validates each child again. Large compact StructArray results can therefore allocate and scan every validity mask up to three times during decoding. Only force nullable state for an empty selected mask, or make subsequent validation reuse an already-current mapping.

PR #51121, opened 2026-07-07
fix: avoid DataNode panic when a binlog is missing during sort compaction

bigsheeper:
## What this fixes

When sort compaction reads a segment whose binlog references an object that is missing in object storage, the DataNode panics with a nil-pointer dereference and enters `CrashLoopBackOff`, instead of failing the compaction task cleanly.

## Root cause

`IterativeRecordReader.Next()` assigned `ir.cur` **before** checking the error returned by `iterate()`. When opening the next binlog chunk fails, `newPackedRecordReader` returns a nil `*packedRecordReader` boxed into a non-nil `RecordReader` interface (a typed-nil). The deferred `rr.Close()` in `sortSegment` then called `Close()` on a nil receiver and panicked — the existing `if pr.reader != nil` / `if ir.cur != nil` guards don't help because the receiver itself is nil.

## Fix

- Only publish the reader once `iterate()` succeeds (assign to a temp, check the error, then set `ir.cur`).
- Make `packedRecordReader.Close` / `ffiPackedRecordReader.Close` nil-receiver safe as defense-in-depth.
- The missing-object error is still surfaced from `Next()` (it is **not** converted to `io.EOF`), so the sort task fails cleanly and is retried rather than silently dropping data.

## Test

Added `TestIterativeRecordReader_MissingChunkDoesNotPanic`, which reproduces the missing-object chunk: it asserts the error is surfaced from `Next()` and that the deferred `Close()` does not panic.

issue: #50927

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #50927, opened 2026-06-30
[Bug]: StorageV2 flush can persist a missing binlog after MinIO drive full, causing DataNode CrashLoopBackOff


zhuwenxing:
## Environment
- Milvus version: `2.6-20260626-159a1a59`
- Deployment mode(standalone or cluster): cluster
- MQ type(rocksmq, pulsar or kafka): pulsar
- SDK version(e.g. pymilvus v2.0.0rc2): N/A, this was observed from Milvus background flush/compaction, not an SDK API repro.
- OS(Ubuntu or CentOS): Ubuntu 22.04.5 LTS
- CPU/Memory: DataNode limit 16 CPU / 32Gi memory; request 2 CPU / 8Gi memory
- GPU: N/A
- Others:
  - Environment: tc stage
  - Namespace: `chaos-testing`
  - Instance: `longrun-multi-replicas-chaos`
  - MinIO image: `minio/minio:RELEASE.2024-12-18T13-15-44Z`
  - Pulsar image: `apachepulsar/pulsar:3.0.7`
  - StorageV2 enabled for affected segment

### K8s Pod List
```
NAME                                                               READY   STATUS             RESTARTS         AGE
longrun-multi-replicas-chaos-etcd-0                                1/1     Running            2 (11h ago)      11h
longrun-multi-replicas-chaos-etcd-1                                1/1     Running            0                11h
longrun-multi-replicas-chaos-etcd-2                                1/1     Running            2 (11h ago)      11h
longrun-multi-replicas-chaos-milvus-datanode-6bc7d8d7c5-hk9x2      0/1     CrashLoopBackOff   72               11h
longrun-multi-replicas-chaos-milvus-datanode-6bc7d8d7c5-t8twz      0/1     CrashLoopBackOff   71               11h
longrun-multi-replicas-chaos-milvus-mixcoord-77d78db88c-nmxr4      1/1     Running            5                11h
longrun-multi-replicas-chaos-milvus-proxy-8494586c4b-dksnn         1/1     Running            5                11h
longrun-multi-replicas-chaos-milvus-querynode-rg1-5cd5695b87jxc    1/1     Running            7                7h55m
longrun-multi-replicas-chaos-milvus-querynode-rg1-5cd5695bplzn8    1/1     Running            2                7h15m
longrun-multi-replicas-chaos-milvus-querynode-rg2-579cb8b65tdkr    1/1     Running            0                81m
[...]

PR #51684, opened 2026-07-22
fix: [cp3.0]avoid DataNode panic when a binlog is missing during sort compaction

bigsheeper:
Cherry-pick from master

pr: #51121
issue: #50927

## Summary

Cherry-picked from master PR #51121 (merged).

When sort compaction reads a segment whose binlog references an object missing in object storage, the DataNode panicked with a nil-pointer dereference (typed-nil `RecordReader` from a failed `newPackedRecordReader`, dereferenced by the deferred `Close()` in `sortSegment`) and entered CrashLoopBackOff. The fix only publishes the reader once `iterate()` succeeds, makes `Close` nil-receiver safe, and keeps surfacing the missing-object error so the sort task fails cleanly and retries.

## Verification

- [x] File count matches original PR (2 files)
- [x] Code changes verified identical to master commit (clean cherry-pick)
- [x] No conflict markers

### DIFF
--- a/internal/storage/record_reader.go
+++ b/internal/storage/record_reader.go
@@ -44,10 +44,10 @@ func (pr *packedRecordReader) Next() (Record, error) {
 }
 
 func (pr *packedRecordReader) Close() error {
-	if pr.reader != nil {
-		return pr.reader.Close()
+	if pr == nil || pr.reader == nil {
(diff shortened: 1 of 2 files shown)


czs007 (approved):
/lgtm

PR #51685, opened 2026-07-22
fix: [cp2.6]avoid DataNode panic when a binlog is missing during sort compaction

bigsheeper:
Cherry-pick from master

pr: #51121
issue: #50927

## Summary

Cherry-picked from master PR #51121 (merged).

When sort compaction reads a segment whose binlog references an object missing in object storage, the DataNode panicked with a nil-pointer dereference (typed-nil `RecordReader` from a failed `newPackedRecordReader`, dereferenced by the deferred `Close()` in `sortSegment`) and entered CrashLoopBackOff. The fix only publishes the reader once `iterate()` succeeds, makes `Close` nil-receiver safe, and keeps surfacing the missing-object error so the sort task fails cleanly and retries.

## Backport adaptation

The master PR also added nil-receiver guards to `ffiPackedRecordReader` (Next/Close) and a matching test assertion. That type **does not exist on 2.6** (the FFI packed reader was never backported), so those parts are dropped. The core panic fix (`IterativeRecordReader.Next()`), the `packedRecordReader.Close` guard, and the `packedRecordReader` half of the test are preserved.

## Verification

- [x] Core fix + packedRecordReader guard applied
- [x] ffiPackedRecordReader parts dropped (type absent on 2.6); no dangling references
- [x] No conflict markers; gofmt clean

### DIFF
--- a/internal/storage/record_reader.go
+++ b/internal/storage/record_reader.go
@@ -37,10 +37,10 @@ func (pr *packedRecordReader) Next() (Record, error) {
 }
 
 func (pr *packedRecordReader) Close() error {
-	if pr.reader != nil {
-		return pr.reader.Close()
+	if pr == nil || pr.reader == nil {
+		return nil
 	}
-	return nil
+	return pr.reader.Close()
 }
 
 func newPackedRecordReader(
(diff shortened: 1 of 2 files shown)


czs007 (approved):
/lgtm

Issue #51687, opened 2026-07-22
[Bug]: TransferSegment/TransferChannel always fail on clusters with more than one collection


czs007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- Milvus version: 2.6.19 (also present on master)
- Deployment mode: cluster
- Reproduced with: birdwatcher `transfer-segment` / `transfer-channel`
```

### Current Behavior

`TransferSegment` and `TransferChannel` in `internal/querycoordv2/ops_services.go` iterate over every replica the source node belongs to. When the requested segment/channel is not found in the replica currently being visited, they `return` an error instead of continuing to the next replica:

```go
replicas := s.meta.GetByNode(ctx, req.GetSourceNodeID())
for _, replica := range replicas {
    segments := s.dist.SegmentDistManager.GetByFilter(
        meta.WithCollectionID(replica.GetCollectionID()), meta.WithNodeID(srcNode))
    segment, ok := lo.Find(segments, func(s *meta.Segment) bool { return s.GetID() == req.GetSegmentID() })
    if !ok {
        return merr.Status(merr.WrapErrSegmentNotFound(...)), nil   // aborts the whole request
    }
    ...
}
```

A query node normally serves many collections and therefore belongs to many replicas, while the requested segment lives in exactly one of them. `meta.GetByNode` returns replicas in map order, so the call only succeeds when the right replica happens to be visited first. In practice both APIs are unusable on any cluster with more than one collection.

`TransferChannel` has the identical pattern.

### Expected Behavior

The request should succeed as long as the segment/channel exists on the source node in any of its replicas, and only report NotFound after all replicas have been searched.

### Steps To Reproduce

```
1. Start a Milvus cluster and load 2+ collections so that one query node belongs to several replicas.
2. Pick a segment that is on that node:
   birdwatcher> show segment-loaded-grpc
3. Try to move it:
   birdwatcher> visit querycoord
   birdwatcher> transfer-segment --sourceNode <N> --targetNode <M> --segmentID <S>

Observed:
   transfer segment failed: segment not found in source node: segment not found[segment=<S>]

even though `get-querynode-distribution --nodeID <N>` lists that exact segment.
```

Same for `transfer-channel --sourceNode <N> --targetNode <M> --channelName <C>`.

### Milvus Log

```
transfer segment failed: segment not found in source node: segment not found[segment=467844892568376490]
[...]

PR #51688, opened 2026-07-22
fix: keep searching replicas in TransferSegment/TransferChannel

czs007:
issue: #51687

## Problem

`TransferSegment` and `TransferChannel` iterate over every replica the source node belongs to. When the requested segment/channel is not in the replica currently being visited, they abort the whole request instead of moving on:

```go
replicas := s.meta.GetByNode(ctx, req.GetSourceNodeID())
    segments := s.dist.SegmentDistManager.GetByFilter(
        meta.WithCollectionID(replica.GetCollectionID()), meta.WithNodeID(srcNode))
    segment, ok := lo.Find(segments, func(s *meta.Segment) bool { return s.GetID() == req.GetSegmentID() })
    if !ok {
        return merr.Status(merr.WrapErrSegmentNotFound(...)), nil   // <- aborts
    }
```

A query node normally serves many collections and so belongs to many replicas, while the requested segment lives in exactly one of them. `meta.GetByNode` returns replicas in map order, so the call only succeeds when the right replica is visited first. On any cluster with more than one collection both APIs are effectively unusable.

Reproduced on a local 2.6.19 cluster (3 collections, 3 query nodes):

```
$ transfer-segment --sourceNode 10 --targetNode 11 --segmentID 467844892568376490
transfer segment failed: segment not found in source node: segment not found[segment=467844892568376490]

$ get-querynode-distribution --nodeID 10
Sealed Segments (2): [467844892568376490 467844892568817028]      # it is right there
```

`transfer-channel` fails the same way.

## Fix

`continue` to the next replica instead of returning, and report NotFound only after all replicas have been searched. `transferAll` mode is unaffected.

## Test

Adds `TestTransferSegmentMultiReplica`, which puts the source node in 6 replicas where only one holds the requested segment.

- Without the fix the test fails (`Should be true` on the transfer response) on most runs — it only passes when the right replica happens to be first.
- With the fix it passes on every run (`-count=5`).
- It also asserts that a segment present in no replica is still reported as NotFound.

Build, vet and the `TestOpsService` suite (with `-tags dynamic,test`) all pass locally.

### DIFF
--- a/internal/querycoordv2/ops_service_test.go
+++ b/internal/querycoordv2/ops_service_test.go
@@ -981,3 +981,75 @@ func (suite *OpsServiceSuite) TestTransferChannel() {
 func TestOpsService(t *testing.T) {
 	suite.Run(t, new(OpsServiceSuite))
 }
+
+// TestTransferSegmentMultiReplica covers the case where the source node serves several
+// collections and therefore belongs to several replicas. The requested segment lives in
+// exactly one of them, and the other replicas must not abort the whole request.
+func (suite *OpsServiceSuite) TestTransferSegmentMultiReplica() {
+	ctx := context.Background()
+	suite.server.UpdateStateCode(commonpb.StateCode_Healthy)
+
+	const (
+		targetCollection = int64(100)
+		partitionID      = int64(1)
+		segmentID        = int64(1000)
+		channelName      = "channel-multi-replica"
+	)
+	nodes := []int64{1, 2}
+
+	// the source node also serves 5 other collections that hold no segment on it
+	for i, collectionID := range []int64{101, 102, 103, 104, 105, targetCollection} {
+		suite.meta.Put(ctx, utils.CreateTestReplica(int64(200+i), collectionID, nodes))
+		suite.meta.PutCollection(ctx,
+			utils.CreateTestCollection(collectionID, 1),
+			utils.CreateTestPartition(partitionID, collectionID))
+	}
+
+	segments := []*datapb.SegmentInfo{{
+		ID:            segmentID,
+		CollectionID:  targetCollection,
+		PartitionID:   partitionID,
+		InsertChannel: channelName,
+		NumOfRows:     1,
+	}}
+	channels := []*datapb.VchannelInfo{{CollectionID: targetCollection, ChannelName: channelName}}
+	suite.broker.EXPECT().GetRecoveryInfoV2(mock.Anything, targetCollection).Return(channels, segments, nil)
+	suite.targetMgr.UpdateCollectionNextTarget(ctx, targetCollection)
+	suite.targetMgr.UpdateCollectionCurrentTarget(ctx, targetCollection)
+	suite.dist.SegmentDistManager.Update(nodes[0], &meta.Segment{SegmentInfo: segments[0], Node: nodes[0]})
+
+	for _, node := range nodes {
+		suite.nodeMgr.Add(session.NewNodeInfo(session.ImmutableNodeInfo{
+			NodeID:   node,
+			Address:  "localhost",
+			Hostname: "localhost",
+		}))
+		suite.meta.HandleNodeUp(ctx, node)
+	}
+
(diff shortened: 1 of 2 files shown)


czs007:
Closing for now — the segment-moving use case this was meant to unblock is fully covered by `LoadBalance` (`birdwatcher balance-segment`), which takes an explicit `collectionID` and therefore never needs the replica search that is broken here.

Leaving #51687 open as a bug report: `transfer-segment --segmentID` and `transfer-channel --channelName` remain unusable on any multi-collection cluster. The `transferAll` mode of both is unaffected. Happy to revive this or take a different approach if maintainers want the API fixed.

tinswzy:
Adversarial review surfaced four low-severity issues — one behavior change worth documenting, plus test-strength and comment-accuracy nits; none are blocking.

## Low

**`internal/querycoordv2/ops_services.go:452` — Behavior change: source node in no replica now returns `SegmentNotFound`/`ChannelNotFound` instead of `Success`**
This is the only old-`Success` → new-`Error` flip in the PR, and it occurs only when `replicas` is empty (the source node belongs to no replica) and `TransferAll=false`: the old code never entered the loop body and fell through to `return merr.Success()`, while the new code leaves `segmentFound`/`channelFound` false and hits the fallback at L452 (and L537 for channels). When `replicas` is non-empty there is no flip, since the old code already returned `NotFound` at the first replica not holding the segment. The new behavior is more correct, but it is externally visible: this RPC is exposed only through the ops REST surface (`internal/proxy/management.go:550`, `internal/coordinator/restful_mgr_routes.go:1059`), so the blast radius is ops tooling and scripts.
*Suggestion:* call this out explicitly in the PR description, and note that the channel path (`TransferChannel`) carries the same flip. *(raised by tinswzy)*

**`internal/querycoordv2/ops_service_test.go:1035` — Multi-replica regression test asserts only the status code, not that a transfer task was produced**
The test uses `suite.taskScheduler.EXPECT().Add(mock.Anything).Return(nil).Maybe()` and asserts only `suite.NoError(err)` / `suite.True(merr.Ok(resp))`. That proves "no `SegmentNotFound`", but not that a task with `src=node1 → dst=node2, segment=1000` was actually generated — if `balanceSegments` regressed to producing no plan at all, this test would still pass.
*Suggestion:* align with the existing `TestTransferSegment` style in the same file (L622-627, L640-645): use `.Once()` with `RunAndReturn` to assert `t.Actions()` length and target node. `internal/querycoordv2/handlers.go:151` (`s.taskScheduler.Add(t)`) is the only observable side effect on this path, so this is straightforward. *(raised by chyezh, tinswzy)*

**`internal/querycoordv2/ops_services.go:517` — `TransferChannel` multi-replica path has no regression test**
The channel-side changes (the `continue` at L513-518, `channelFound = true` at L519, and the fallback `NotFound` at L537-541) are symmetric to but independent of the segment path, yet the new test covers segments only. None of the nine existing `OpsServiceSuite` cases exercises a multi-replica channel scenario — `TestTransferChannel` (L727) builds only a single replica for `collectionID=1` (L749-757).
*Suggestion:* add a `TestTransferChannelMultiReplica` mirroring the new segment case. *(raised by chyezh, tinswzy)*

**`internal/querycoordv2/ops_service_test.go:1037` — Test comment mischaracterizes the bug as replica-iteration-order dependent**
The comment reads "before the fix this returned SegmentNotFound whenever a replica of another collection happened to be visited before the one holding the segment", implying the bug depends on `GetByNode` iteration order and that the test catches it only probabilistically. In fact the removed early return fired unconditionally for *any* replica not holding the segment — even if the holding replica was visited first, a later decoy replica would still fail the whole request. With ≥1 decoy replica the old code failed 100% of the time, so the test is deterministic.
*Suggestion:* reword to "returned SegmentNotFound as soon as it visited any replica that does not hold the segment", so a future reader doesn't misdiagnose this test as flaky and "fix" it. *(surfaced during verification)*

PR #51692, opened 2026-07-22
fix: keep BM25/MinHash searchable after add_function_field with stale segment index meta

MrPresent-Han:
issue: #51366

## Problem

A field added by `add_function_field` (a BM25 or MinHash function output field) can stay unsearchable: the search keeps failing with `field index meta not found for field N` (an `AssertInfo` in `IndexMeta.cpp`) even though `describe_index` reports the index `Finished`, until the collection is released and loaded again.

Root cause: a segment captures its index meta at construction time — the per-segment `ChunkedSegmentSealedImpl::col_index_meta_`, and the growing `IndexingRecord::index_meta_`. A schema-only reopen refreshes the schema but not this snapshot, so a field added by `add_function_field` is missing from it. On segments without a built sparse index (sealed segments below the index-build threshold, and growing segments), BM25/MinHash brute force fetched the field's index params via `GetFieldIndexMeta`, which asserts on the missing field.

## Fix

- Guard the brute-force index-param fetch at both `vector_search` entry points — `ChunkedSegmentSealedImpl::vector_search` and `SegmentGrowingImpl::vector_search` (via `SearchOnGrowing`) — with `HasField`, so a field missing from the stale snapshot no longer trips the assert.
- Deliver BM25 `k1`/`b` through the search plan: the delegator reads them from the up-to-date collection-level index meta (`GetBM25Params`) and injects them into the search params alongside `avgdl`, so brute force sources them from the plan instead of the stale per-segment meta. `PrepareBFSearchParams` now reads `k1`/`b` tolerantly, keeping the plan-delivered values when the per-segment meta lacks the field. When the collection-level index meta already carries the field, the real (possibly user-customized) `k1`/`b` are passed down.
- MinHash falls back to knowhere defaults for the brief window before the segment is reloaded.

The interim-index build paths on both sides already guard with `HasField`, so a missing field is skipped there and the search falls through to the fixed brute-force path.

### DIFF
--- a/internal/core/src/common/QueryInfo.h
+++ b/internal/core/src/common/QueryInfo.h
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <memory>
+#include <optional>
 
 #include "ArrayOffsets.h"
 #include "common/Tracer.h"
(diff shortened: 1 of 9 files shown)


liliu-z (commented):
/lgtm
/approve

Issue #51693, opened 2026-07-22
[Bug]: Segment-derived resources can diverge from published segment state


xiaofan-luan:
## Current behavior

`ChunkedSegmentSealedImpl` publishes schema, load metadata, readiness, and most query-visible runtime objects through an immutable `PublishedSegmentState`. However, several resources derived from that data have used separate ownership or cache lifetimes:

- GEOMETRY decode caches are managed outside the published runtime state and depend on explicit segment-level cleanup.
- TEXT LOB readers are managed as a process-wide capacity cache instead of being owned by the loaded segment states that reference each LOB path.
- expression-result cache identity does not distinguish segment data/index generations, so a reopen or replacement with the same segment ID and row count can reuse results from an older generation.
- growing-segment schema, GEOMETRY cache, and TEXT representation state can be captured independently.

During reopen, field replacement, or drop, an in-flight reader may still own the old column/index generation while a separately managed derived resource is removed, reused, or observed from a newer generation. This makes cleanup ordering part of correctness and can produce stale expression results or mixed-generation interpretation.

## Expected behavior

Resources needed to interpret query-visible segment data should follow the same ownership and publication boundary as that data:

- sealed-segment runtime snapshots strongly own their GEOMETRY caches and TEXT LOB reader handles;
- old snapshots keep those resources alive until the final in-flight owner releases them;
- TEXT readers may be deduplicated by LOB path, but the registry should be weak and have no capacity/LRU eviction semantics;
- a segment should acquire a TEXT reader only when the internal V3 TEXT column exists in the storage manifest and is loaded;
- expression-result cache entries should include a query-state generation and should not be read or published after that generation changes;
- growing-segment schema and derived field resources should be captured with lifetime-safe ownership;
- dropping a field or segment should release its derived resources after the last snapshot/reference retires.

## Acceptance criteria

- Reopen/replace/drop tests demonstrate that captured derived resources remain usable for the lifetime of the captured state.
- Same-row-count field/index replacement cannot hit expression results from an older generation.
- Segments that reference the same TEXT LOB path share one reader handle, and the reader is released after the last segment/query owner.
- There is no `max_file_readers`/LRU ownership limit; loaded columns determine live reader count.
- Schema-only and external TEXT fields do not acquire partition LOB readers.
- GEOMETRY and TEXT behavior is covered for sealed and growing segments.

## Related

- #51068 tracks the broader COW runtime-publication direction.
[...]

PR #47567, opened 2026-02-05
feat: support TEXT column with LOB storage for large text fields

zhagnlu:
issue: #48783
design doc: https://github.com/milvus-io/milvus-design-docs/blob/main/design_docs/20260407-text_lob_storage.md

PR #51694, opened 2026-07-22
fix: bind segment-derived resources to published state

xiaofan-luan:
issue: #51693

## What changed

- Bind sealed-segment GEOMETRY caches and TEXT LOB reader handles to `PublishedSegmentState::RuntimeResourceState`, so old published snapshots retain the resources needed to interpret their data.
- Replace the capacity-based `TextColumnCache` with reference-counted `SharedTextLobReader` handles. A global weak registry deduplicates readers by LOB path but owns no reader and applies no capacity/LRU eviction.
- Acquire partition TEXT readers only for loaded internal V3 TEXT columns; schema-only and external TEXT fields do not create readers.
- Keep growing-segment TEXT spillover/remote-reader representation boundaries and GEOMETRY caches in lifetime-safe shared resource state, including field-drop cleanup.
- Return schema snapshots through `SchemaPtr` and have each expression capture the published schema snapshot at construction, so an in-flight expression keeps the schema of the segment state it started on.
- Preserve structured storage error categories across the TEXT reader boundary: storage-transient code `2045` remains retriable, while permanent storage code `2044` remains non-retriable.
- Add regression tests for sealed/growing resource lifetime, reader sharing and release, GEOMETRY ownership, TEXT representation boundaries, and storage error propagation.

## Why

Several query-visible resources were derived from segment data but lived in independent global caches or mutable members. During reopen, replacement, or drop, the published data and those resources could therefore have different ownership lifetimes.

The fix makes the published/runtime state the strong owner of derived resources. Shared readers are deduplicated for reuse, but their lifetime is controlled only by segment/query references rather than an unrelated process-wide capacity limit.

## Scope

This PR fixes ownership and lifetime for resources derived from segment data — GEOMETRY caches, TEXT LOB readers, growing-segment representation state, and each expression's schema snapshot.

The expression-result cache key stays generation-free (`{segment_id, predicate}`, validated by row_count). Instead of adding a generation, a reopen that **replaces existing row data** — compaction rewriting a column's parquet, or an external-manifest refresh, both of which can change a value while keeping the segment id and row count — drops that segment's cached filter bitsets (`LoadDiff::ReplacesExistingFieldData` → `InvalidateExprCacheAfterReopen`) **as `committer.Publish`'s `while_locked` hook**: after the new state is stored but before the publish lease reopens the read gate. The gate has already drained in-flight readers (their stale writes landed) and blocks new readers until the erase completes, so no query can read the freshly published data and still hit the pre-reopen bitmap — erasing *after* the lease released would leave exactly that window (P1 caught in review). Add-only reopens (new fields / indexes over unchanged values) keep the still-correct cache. This satisfies the same-row-count acceptance in #51693 without a generation in the key.

It also does not implement the complete per-operation sealed-segment state pin tracked by #51594; access paths not covered by an explicit snapshot may still observe a later published generation between separate accessor calls.

Related architectural direction: #51068. TEXT LOB origin: #47567 / #48783.

## Review fixes (2026-07-23)

Follow-up fixes from a deep multi-agent review, squashed into this commit:

- **Schema-only TEXT dispatch**: a TEXT field added by schema change has no manifest column and therefore no LOB reader by design; `get_raw_data` and `CreateTextIndexWithSchema` now dispatch on reader presence (no reader = plain string column) instead of hard-asserting, so adding a nullable / `enable_match` TEXT field no longer fails queries or segment reopen. Regression tests added for both paths.
- **Recursive `sch_mutex_` acquisition**: `SegmentGrowingImpl::Insert()` holds `sch_mutex_` shared and reached `get_schema()` again through `set_field_avg_size`, which is UB and can deadlock against a queued `Reopen` writer; the base `set_field_avg_size` no longer re-resolves the schema (callers already gate on `IsVariableDataType`).
- **Construction-time schema snapshots**: `PhyMatchFilterExpr`, `PhyElementFilterBitsNode`, and `PhyIterativeElementFilterNode` captured the schema per Eval/GetOutput batch; they now capture once at construction, matching `SegmentExpr::CaptureSchemaSnapshot()` so one query cannot mix schema generations.
- Null-check on Loon FFI properties in `InitTextLobReaders`.
- **Expr-cache invalidation on data-replacing reopen** (second + third review rounds): the process-level filter-bitset cache is keyed by `{segment_id, predicate}` with no data generation and was only cleared on full segment `Release`. A reopen that rewrites an existing column's backing files (compaction) or refreshes an external manifest keeps the segment id and row count but changes values, so a later same-predicate query could hit a stale bitmap. Reopen paths drop the segment's cache when `LoadDiff::ReplacesExistingFieldData()` (see Scope). The invalidation runs as the `committer.Publish` `while_locked` hook — after the state swap, before the publish lease reopens the read gate — so no query can read the new data and still hit the stale entry in the gap (a P1 found in review: erasing *after* `committer.Publish` returned left that window open). The erase is also unconditional (not gated on the cache enable switch — `SetEnabled(false)` only flips a flag without dropping entries, so a reopen during a disabled window must still clear entries that would otherwise resurface on re-enable; a second review-round P1). Corrects the earlier "not reachable" claim in this PR and in #51693.

Growing-segment per-operation resource capture (in-flight query vs. field-drop race, fail-loud only) is tracked separately in #51781.

## Validation

- Built the complete C++ `all_tests` target on the remote dev host at this commit.
- `SealedSegmentCowState.*` (36), `LoadDiffTest.*`, `SegmentLoadInfoTest.ComputeDiffMixedFormats`, `SchemaReopenTest.*`: all passed. Expr-cache tests: `ReopenReplacingDataInvalidatesExprCache` (data-replacing reopen drops the cache, add-only keeps it); `ExprCacheInvalidationOrderedBeforeReaderAdmission` (erase runs while a new read lease is refused — pins erase-before-reader-admission); `ExprCacheInvalidatedEvenWhenDisabledDuringReopen` (disabled reopen still erases, so nothing resurfaces on re-enable); `LoadDiffTest.ReplacesExistingFieldData` truth table. Plus schema-only TEXT read + match-index paths.
- CI code-check (gofmt + golangci-lint + clang-format) confirmed green on the prior commit.
- `ClearPublishedStateDropsRuntimeSnapshot` was asserting the wrong representation (`runtime == nullptr`); the publish boundary normalizes a cleared runtime to a fresh empty `RuntimeResourceState`. The test now asserts the semantics (no derived resources survive the clear).
- `pkg/util/merr` Go tests passed; `git diff --check` clean.


MrPresent-Han:
I take this

MrPresent-Han:
https://github.com/milvus-io/milvus/pull/51985
I set a new pr for this, WIP

Issue #51695, opened 2026-07-22
[Enhancement]: Reuse V3 packed-index tail reads


xaxys:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### What would you like to be added?

Reuse bytes that `IndexEntryReader` has already downloaded from the tail of a V3 packed scalar-index object, instead of issuing additional object-storage range requests for the same bytes.

Milvus scalar index engine version 3 made the single-file V3 packed layout the default, and engine version 4 continues to use the same format. A scalar-index build normally creates one packed object for each segment and indexed scalar field, for example:

```text
index_v1/.../milvus_packed_inverted_index.v3
index_v1/.../milvus_packed_stlsort_index.v3
index_v1/.../milvus_packed_bitmap_index.v3
```

Each object contains a magic header, one or more index entries, a `__meta__` entry, a directory table, and a footer. `IndexEntryReader` must already read up to the final 64 KiB of the object to locate and parse the footer and directory. Later operations can nevertheless issue separate `ReadAt` calls for the magic value, metadata, and entries whose complete ranges are already present in that downloaded tail.

The reader should:

- read the initial tail first and validate the V3 magic before interpreting the footer or directory;
- use the initial tail for magic validation when it contains the whole file, while retaining the required separate header read for larger files;
- retain at most the final 64 KiB for the lifetime of one reader;
- serve a later read from retained bytes only when the complete requested range is covered;
- fall back to the original `InputStream::ReadAt` behavior for misses and partial overlaps;
- apply the same behavior to plain, encrypted, in-memory, file-output, and streaming entry reads;
- preserve CRC verification, decryption, cancellation, short-read handling, provider error propagation, and output ordering;
- avoid a process-wide or cross-load cache, eviction policy, and stale-object semantics.

For a packed object no larger than 64 KiB, the initial tail is the complete file. Magic validation, metadata parsing, directory access, and tail-resident entry loading can therefore be satisfied without downloading overlapping ranges again. Larger objects still benefit when metadata or complete smaller entries reside in the retained final tail; their main entry payload continues to use the normal storage path.

### Why is this needed?

The current read shape creates request amplification during scalar-index load. Because packed objects are produced per segment and indexed scalar field, a collection with several scalar indexes or many segments can open many such objects at once. A small packed object may be downloaded as the initial tail and then partially refetched several times through S3 Range GETs.

[...]

PR #47690, opened 2026-02-09
feat: update scalar index serialized format to V3

zhengbuqian:
This PR added support for scalar index format V3, packing each scalar index into a single file, reducing the amount of metadata in ETCD, and number of files in S3.

~All C++ UTs are updated to run with V3 format. Production code will still run on V2.~

issue: https://github.com/milvus-io/milvus/issues/47417
design doc: https://github.com/milvus-io/milvus-design-docs/blob/main/design_docs/20260209-scalar-index-unified-format.md

PR #51696, opened 2026-07-22
enhance: reuse packed index tail reads

xaxys:
issue: #51695

## What does this PR do?

This PR removes redundant object-storage Range GETs while opening and loading V3 packed scalar indexes.

Milvus scalar index engine version 3 made the V3 single-file packed layout the default, and engine version 4 continues to use the same format. A scalar-index build normally creates one packed object for each segment and indexed scalar field. Each object contains a magic header, index entries, a `__meta__` entry, a directory table, and a footer.

`IndexEntryReader::Open()` downloads up to the final 256 KiB of the object to parse the footer and directory. Before this change, subsequent operations could issue separate reads for the magic value, metadata, and entries even when their complete ranges were already present in that downloaded tail. Small packed indexes were particularly inefficient: when the whole object fit in the tail read, portions of the same object could still be fetched repeatedly.

This is request amplification rather than useful index I/O. It is most visible with remote S3-compatible storage, multiple scalar indexes, many segments, and operations that reopen indexes, such as collection load, QueryNode restart, replica scale-out, rebalance, release/load, and loading replacement segments after compaction.

## How it works

- `Open` reads the initial file tail, validates the V3 magic before interpreting the footer or directory, and then parses and retains the tail.
- When the file is no larger than 256 KiB, magic validation uses the already-downloaded full-file buffer. Larger files keep the required separate header read.
- The reader retains at most the final 256 KiB. If directory or metadata parsing temporarily needs a larger buffer, only a fresh bounded final-tail allocation survives for the reader lifetime; the larger parse allocation is released.
- `InspectStreamLoadInfo()` keeps its 64 KiB inspection window because resource estimation only needs the footer and directory and does not retain a reader for subsequent entry loads.

Current master performs this inspection as a separate storage read while estimating resources for non-HYBRID V3 packed scalar-index loads, and as a fallback for HYBRID inspection. In the common small-directory case it adds one 64 KiB tail Range GET before the later `Open()` used by actual loading. The inspection reader is destroyed after returning sizing metadata, so those bytes cannot be reused by this reader-local tail cache; this PR does not eliminate that request.
- A coverage-aware `ReadAt` serves a request from the retained tail only when the complete requested range is covered.
- Cache misses and partial overlaps use the original `InputStream::ReadAt` path unchanged. The implementation intentionally does not stitch cached and provider bytes, so existing short-read and provider-error semantics remain intact.
- Plain, encrypted, in-memory, file-output, and streaming entry paths use the same coverage-aware read source.
- Invalid negative, unaddressable, or smaller-than-header-and-footer file sizes are rejected before issuing a storage read.
- The existing public `IndexEntryReader::Open` API is unchanged. No cipher-plugin injection overload or other test-only production API is added.

The legacy `small_entry_cache_` is removed. That cache copied every entry up to 1 MiB into an unbounded reader-owned map, but the production call graph reads each named auxiliary entry at most once per reader; `__meta__` is parsed during `Open` and is not fetched through `ReadEntry` again. It therefore added allocation and copying without production hits. For encrypted indexes it also retained a second plaintext copy until reader destruction. This was not a security vulnerability, but removing the unused copy gives decrypted data a tighter lifetime and lowers peak memory.

The trade-off is explicit: a future caller that repeatedly reads the same small entry outside the retained final tail would now issue another provider read instead of hitting `small_entry_cache_`. No current production caller has that access pattern. Entries fully covered by the retained tail still avoid provider I/O through the shared `TailCachedInputStream`.

The change does not add a process-wide cache, cross-load reuse, an eviction policy, or stale-object semantics. Existing CRC verification, encryption/decryption, cancellation, and entry-output behavior remain in place.

## Scope and user impact

This directly affects V3 packed scalar and text indexes loaded through `IndexEntryReader`, including STL_SORT, INVERTED, BITMAP, HYBRID, MARISA, JSON/ARRAY/nested scalar indexes, NGRAM, RTREE, and packed text-index variants. Encrypted packed indexes use the same read path.

It does not affect:

- vector-index artifacts such as IVF, HNSW, or DiskANN files;
- insert/field-data Parquet, Vortex/Loon, raw-vector, or LOB objects;
- delta logs, Bloom/PK/JSON/BM25 stats, manifests, WAL, snapshots, or backups;
- legacy pre-V3 scalar-index layouts;
- PUT, HEAD, or LIST requests;
- steady-state search after an index is loaded.

Collections with multiple scalar indexes and many relatively small segments benefit most. Pure-vector collections should see little or no change. For large packed indexes, the main entry data still follows the normal storage path, so the proportional benefit is smaller.

The expected effects are lower object-store request count, request cost, provider pressure, duplicated response bytes, and RTT sensitivity. End-to-end collection-load latency can still be dominated by QueryCoord scheduling, concurrent loads, and large vector-index downloads.

## Latest observed impact

Provider-level A/B testing confirmed a substantial reduction in packed-index GET amplification and showed lower warm-load latency under controlled object-store RTT, with no observed correctness, provider-error, or representative-workload regression.

The full dataset schema, data distribution, vector and scalar index configuration, query mix, run ordering, and results are documented separately in the [provider-level validation comment](https://github.com/milvus-io/milvus/pull/51696#issuecomment-5043387104).

## Validation

`IndexEntryReaderTailCacheTest.*` contains nine focused cases and was executed three times against this change (27/27 passed), covering:

- exact 256 KiB cached-tail boundary reads;
- plaintext full-coverage hits and partial-overlap provider fallback;
- oversized directory and metadata parsing while retaining only the final bounded tail;
- corrupt magic and arbitrary non-V3 input, including magic diagnostics before footer errors;
- provider short-read failure;
- pre-cancelled open;
- negative, unaddressable, and too-small file sizes.

`TailCachedInputStreamTest.ServesOnlyFullyCoveredRanges` directly verifies cache hits, misses, partial overlaps, zero-length reads, and range boundaries independently of entry encoding.

This PR does not include an encrypted-reader GET-count/decryption integration test. Encrypted read methods are structurally routed through the same `TailCachedInputStream`, but the existing mock cipher tests exercise encrypted writing rather than reader-side cache reuse. End-to-end encrypted reader coverage should use a real loadable cipher-plugin fixture instead of adding a test-only `IndexEntryReader::Open` overload.

### DIFF
--- a/internal/core/src/storage/IndexEntryReader.cpp
+++ b/internal/core/src/storage/IndexEntryReader.cpp
@@ -38,6 +38,7 @@
 #include "storage/EntryStreamUtils.h"
 #include "storage/Crc32cUtil.h"
 #include "storage/PluginLoader.h"
+#include "storage/TailCachedInputStream.h"
 
 namespace milvus::storage {
 namespace {
@@ -129,6 +130,8 @@ SaturatingAdd(size_t lhs, size_t rhs) {
 }
 
 constexpr size_t kEntryDownloadRangeSize = 16 * 1024 * 1024;
+constexpr size_t kInspectionTailSize = 64 * 1024;
+constexpr size_t kCachedTailMaxSize = 256 * 1024;
 
 void
 DrainFutures(std::vector<std::future<void>>& futures,
@@ -352,7 +355,8 @@ IndexEntryReader::InspectStreamLoadInfo(
     reader->CheckCancelled("IndexEntryReader::InspectStreamLoadInfo");
     // The caller has already selected the V3 path. Actual loading validates
     // the magic; inspection avoids a separate range read at offset zero.
-    reader->ReadFooterAndDirectory();
+    reader->ReadFooterAndDirectory(kInspectionTailSize,
+                                   /*validate_magic=*/false);
     reader->CheckCancelled("IndexEntryReader::InspectStreamLoadInfo");
     return reader->stream_load_info_;
 }
@@ -363,15 +367,32 @@ IndexEntryReader::Open(std::shared_ptr<milvus::InputStream> input,
                        int64_t collection_id,
                        ThreadPoolPriority priority,
                        folly::CancellationToken cancellation_token) {
+    ThrowIfCancelled(cancellation_token, "IndexEntryReader::Open");
+
+    constexpr int64_t kMinFileSize =
+        MILVUS_V3_MAGIC_SIZE + MILVUS_V3_FOOTER_SIZE;
+    AssertInfo(
+        file_size >= kMinFileSize,
+        "V3 index file size {} is smaller than the minimum header and footer "
+        "size {}",
+        file_size,
+        kMinFileSize);
+    if constexpr (sizeof(size_t) < sizeof(uint64_t)) {
+        AssertInfo(static_cast<uint64_t>(file_size) <=
+                       std::numeric_limits<size_t>::max(),
+                   "V3 index file size {} exceeds the addressable range",
+                   file_size);
+    }
+
     auto reader = std::unique_ptr<IndexEntryReader>(new IndexEntryReader());
     reader->input_ = std::move(input);
-    reader->file_size_ = file_size;
+    reader->file_size_ = static_cast<size_t>(file_size);
     reader->collection_id_ = collection_id;
     reader->priority_ = priority;
     reader->cancellation_token_ = cancellation_token;
     reader->CheckCancelled("IndexEntryReader::Open");
-    reader->ValidateMagic();
-    reader->ReadFooterAndDirectory();
+    reader->ReadFooterAndDirectory(kCachedTailMaxSize,
+                                   /*validate_magic=*/true);
     reader->CheckCancelled("IndexEntryReader::Open");
 
     if (reader->is_encrypted_) {
@@ -401,31 +422,36 @@ IndexEntryReader::CheckCancelled(const std::string& operation) const {
 }
 
 void
-IndexEntryReader::ValidateMagic() {
-    CheckCancelled("IndexEntryReader::ValidateMagic");
-    char magic_buf[MILVUS_V3_MAGIC_SIZE];
-    size_t bytes_read = input_->ReadAt(magic_buf, 0, MILVUS_V3_MAGIC_SIZE);
-    CheckCancelled("IndexEntryReader::ValidateMagic");
-    AssertInfo(bytes_read == MILVUS_V3_MAGIC_SIZE,
-               "Failed to read V3 magic number");
-    AssertInfo(
-        std::memcmp(magic_buf, MILVUS_V3_MAGIC, MILVUS_V3_MAGIC_SIZE) == 0,
-        "Invalid V3 magic number");
+IndexEntryReader::ValidateMagic(const uint8_t* magic) const {
+    AssertInfo(std::memcmp(magic, MILVUS_V3_MAGIC, MILVUS_V3_MAGIC_SIZE) == 0,
+               "Invalid V3 magic number");
 }
 
 void
-IndexEntryReader::ReadFooterAndDirectory() {
+IndexEntryReader::ReadFooterAndDirectory(size_t initial_tail_size,
+                                         bool validate_magic) {
     CheckCancelled("IndexEntryReader::ReadFooterAndDirectory");
-    constexpr size_t kTailBufferSize = 64 * 1024UL;
-    size_t tail_size =
-        std::min(static_cast<size_t>(file_size_), kTailBufferSize);
+    size_t tail_size = std::min(file_size_, initial_tail_size);
     size_t tail_offset = file_size_ - tail_size;
+
+    if (validate_magic && tail_offset != 0) {
+        uint8_t magic[MILVUS_V3_MAGIC_SIZE];
+        size_t magic_bytes = input_->ReadAt(magic, 0, MILVUS_V3_MAGIC_SIZE);
+        CheckCancelled("IndexEntryReader::ReadFooterAndDirectory");
+        AssertInfo(magic_bytes == MILVUS_V3_MAGIC_SIZE,
+                   "Failed to read V3 magic number");
+        ValidateMagic(magic);
+    }
 
     std::vector<uint8_t> tail_data(tail_size);
     size_t bytes_read =
         input_->ReadAt(tail_data.data(), tail_offset, tail_size);
     CheckCancelled("IndexEntryReader::ReadFooterAndDirectory");
     AssertInfo(bytes_read == tail_size, "Failed to read file tail");
+
+    if (validate_magic && tail_offset == 0) {
+        ValidateMagic(tail_data.data());
+    }
 
     // Parse 32-byte Footer from the last 32 bytes
     AssertInfo(tail_size >= MILVUS_V3_FOOTER_SIZE,
@@ -448,7 +474,7 @@ IndexEntryReader::ReadFooterAndDirectory() {
     AssertInfo(dir_size > 0, "Directory table size is zero");
     AssertInfo(static_cast<size_t>(dir_size) + meta_entry_size +
                        MILVUS_V3_FOOTER_SIZE + MILVUS_V3_MAGIC_SIZE <=
-                   static_cast<size_t>(file_size_),
+                   file_size_,
                "Directory table + meta entry + footer size exceeds file size");
 
     // Check if the directory itself needs a second read. The meta entry is
@@ -552,6 +578,21 @@ IndexEntryReader::ReadFooterAndDirectory() {
             entry_index_.emplace(std::move(name), std::move(meta));
         }
     }
+
+    std::vector<uint8_t> retained_tail;
+    if (tail_data.size() <= kCachedTailMaxSize) {
+        retained_tail = std::move(tail_data);
+    } else {
+        // vector::erase does not reduce capacity. Copy the final tail into a
+        // fresh bounded allocation so an unusually large directory or meta
+        // parse buffer is not retained for the reader lifetime.
+        retained_tail.assign(tail_data.end() - kCachedTailMaxSize,
+                             tail_data.end());
+    }
+    input_ = std::make_shared<detail::TailCachedInputStream>(
+        std::move(input_),
+        file_size_ - retained_tail.size(),
+        std::move(retained_tail));
 }
 
 std::vector<std::string>
@@ -600,11 +641,6 @@ IndexEntryReader::StreamDownloadTaskCount(const EntryMeta& meta) {
 Entry
 IndexEntryReader::ReadEntry(const std::string& name) {
     CheckCancelled("IndexEntryReader::ReadEntry");
-    auto cache_it = small_entry_cache_.find(name);
-    if (cache_it != small_entry_cache_.end()) {
-        return cache_it->second;
-    }
-
     auto it = entry_index_.find(name);
     AssertInfo(it != entry_index_.end(), "Entry not found: {}", name);
     const auto& meta = it->second;
@@ -614,10 +650,6 @@ IndexEntryReader::ReadEntry(const std::string& name) {
         result = ReadEncryptedEntry(meta);
     } else {
         result = ReadPlainEntry(meta);
-    }
-
-    if (result.data.size() <= kSmallEntryCacheThreshold) {
-        small_entry_cache_[name] = result;
     }
 
     return result;
@@ -640,6 +672,7 @@ IndexEntryReader::ReadPlainEntry(const EntryMeta& meta) {
     }
 
     auto& pool = ThreadPools::GetThreadPool(priority_);
+    auto input = input_;
     auto cancellation_token = cancellation_token_;
     uint8_t* dest = result.data.data();
 
@@ -655,10 +688,10 @@ IndexEntryReader::ReadPlainEntry(const EntryMeta& meta) {
             size_t this_offset = offset;
 
             futures.push_back(pool.Submit(
-                [this, dest, this_offset, len, &pm, cancellation_token]() {
+                [input, dest, this_offset, len, &pm, cancellation_token]() {
                     ThrowIfCancelled(cancellation_token,
                                      "IndexEntryReader::ReadPlainEntry");
-                    size_t n = input_->ReadAt(
+                    size_t n = input->ReadAt(
                         dest + this_offset,
                         MILVUS_V3_MAGIC_SIZE + pm.offset + this_offset,
                         len);
@@ -693,6 +726,11 @@ IndexEntryReader::ReadEncryptedEntry(const EntryMeta& meta) {
     result.data.resize(em.original_size);
 
     auto& pool = ThreadPools::GetThreadPool(priority_);
+    auto input = input_;
+    auto cipher_plugin = cipher_plugin_;
+    auto edek = edek_;
+    int64_t ez_id = ez_id_;
+    int64_t collection_id = collection_id_;
     auto cancellation_token = cancellation_token_;
     uint8_t* dest = result.data.data();
 
@@ -708,7 +746,11 @@ IndexEntryReader::ReadEncryptedEntry(const EntryMeta& meta) {
             size_t plain_len = std::min(remaining, slice_size_);
             cur_output_offset += plain_len;
 
-            futures.push_back(pool.Submit([this,
+            futures.push_back(pool.Submit([input,
+                                           cipher_plugin,
+                                           ez_id,
+                                           collection_id,
+                                           edek,
                                            slice,
                                            dest,
                                            this_output_offset,
@@ -717,15 +759,15 @@ IndexEntryReader::ReadEncryptedEntry(const EntryMeta& meta) {
                 ThrowIfCancelled(cancellation_token,
                                  "IndexEntryReader::ReadEncryptedEntry");
                 std::vector<uint8_t> cipher(slice.size);
-                size_t n = input_->ReadAt(cipher.data(),
-                                          MILVUS_V3_MAGIC_SIZE + slice.offset,
-                                          slice.size);
+                size_t n = input->ReadAt(cipher.data(),
+                                         MILVUS_V3_MAGIC_SIZE + slice.offset,
+                                         slice.size);
                 ThrowIfCancelled(cancellation_token,
                                  "IndexEntryReader::ReadEncryptedEntry");
                 AssertInfo(n == slice.size, "Failed to read encrypted slice");
 
                 auto dec =
-                    cipher_plugin_->GetDecryptor(ez_id_, collection_id_, edek_);
+                    cipher_plugin->GetDecryptor(ez_id, collection_id, edek);
                 auto plain = dec->Decrypt(cipher.data(), cipher.size());
 
                 AssertInfo(plain.size() == plain_len,
@@ -792,11 +834,16 @@ IndexEntryReader::SubmitEntryDownloadTasks(
     EntryDownloadState& state,
     std::vector<std::future<void>>& futures) {
     auto& pool = ThreadPools::GetThreadPool(priority_);
+    auto input = input_;
     auto cancellation_token = cancellation_token_;
     futures.reserve(futures.size() + DownloadTaskCount(meta));
 
     if (meta.encrypted) {
         const auto& em = meta.enc;
+        auto cipher_plugin = cipher_plugin_;
+        auto edek = edek_;
+        int64_t ez_id = ez_id_;
+        int64_t collection_id = collection_id_;
         size_t output_offset = 0;
 
         for (size_t i = 0; i < em.slices.size(); i++) {
@@ -806,7 +853,11 @@ IndexEntryReader::SubmitEntryDownloadTasks(
             size_t plain_len = std::min(remaining, slice_size_);
             output_offset += plain_len;
 
-            futures.push_back(pool.Submit([this,
+            futures.push_back(pool.Submit([input,
+                                           cipher_plugin,
+                                           ez_id,
+                                           collection_id,
+                                           edek,
                                            slice,
                                            fd = state.fd,
                                            this_output_offset,
@@ -817,15 +868,15 @@ IndexEntryReader::SubmitEntryDownloadTasks(
                 ThrowIfCancelled(cancellation_token,
                                  "IndexEntryReader::ReadEntriesToFiles");
                 std::vector<uint8_t> cipher(slice.size);
-                size_t n = input_->ReadAt(cipher.data(),
-                                          MILVUS_V3_MAGIC_SIZE + slice.offset,
-                                          slice.size);
+                size_t n = input->ReadAt(cipher.data(),
+                                         MILVUS_V3_MAGIC_SIZE + slice.offset,
+                                         slice.size);
                 ThrowIfCancelled(cancellation_token,
                                  "IndexEntryReader::ReadEntriesToFiles");
                 AssertInfo(n == slice.size, "Failed to read encrypted slice");
 
                 auto dec =
-                    cipher_plugin_->GetDecryptor(ez_id_, collection_id_, edek_);
+                    cipher_plugin->GetDecryptor(ez_id, collection_id, edek);
                 auto plain = dec->Decrypt(cipher.data(), cipher.size());
 
                 AssertInfo(plain.size() == plain_len,
@@ -857,18 +908,18 @@ IndexEntryReader::SubmitEntryDownloadTasks(
             size_t this_src_offset = src_offset;
             size_t this_range_idx = range_idx;
 
-            futures.push_back(pool.Submit([this,
-                                           this_src_offset,
+            futures.push_back(pool.Submit([this_src_offset,
                                            len,
                                            fd = state.fd,
                                            this_file_offset,
                                            this_range_idx,
                                            &state,
-                                           cancellation_token]() {
+                                           cancellation_token,
+                                           input]() {
                 ThrowIfCancelled(cancellation_token,
                                  "IndexEntryReader::ReadEntriesToFiles");
                 std::vector<uint8_t> buf(len);
-                size_t n = input_->ReadAt(
+                size_t n = input->ReadAt(
                     buf.data(), MILVUS_V3_MAGIC_SIZE + this_src_offset, len);
                 ThrowIfCancelled(cancellation_token,
                                  "IndexEntryReader::ReadEntriesToFiles");
--- a/internal/core/src/storage/IndexEntryReader.h
+++ b/internal/core/src/storage/IndexEntryReader.h
@@ -169,9 +169,9 @@ class IndexEntryReader {
     IndexEntryReader() = default;
 
     void
-    ReadFooterAndDirectory();
+    ReadFooterAndDirectory(size_t initial_tail_size, bool validate_magic);
     void
-    ValidateMagic();
+    ValidateMagic(const uint8_t* magic) const;
     void
     CheckCancelled(const std::string& operation) const;
 
@@ -258,7 +258,7 @@ class IndexEntryReader {
     FinalizeEntryStreamDownload(EntryStreamDownloadState& state);
 
     std::shared_ptr<milvus::InputStream> input_;
-    int64_t file_size_ = 0;
+    size_t file_size_ = 0;
     int64_t collection_id_ = 0;
     ThreadPoolPriority priority_ = ThreadPoolPriority::HIGH;
     folly::CancellationToken cancellation_token_;
@@ -274,8 +274,6 @@ class IndexEntryReader {
     EntryStreamLoadInfo stream_load_info_;
     std::vector<std::string> entry_names_;
 
-    static constexpr size_t kSmallEntryCacheThreshold = 1 * 1024 * 1024;
-    std::unordered_map<std::string, Entry> small_entry_cache_;
     nlohmann::json meta_json_;
 };
 
--- a/internal/core/src/storage/IndexEntryWriterTest.cpp
+++ b/internal/core/src/storage/IndexEntryWriterTest.cpp
@@ -23,9 +23,11 @@
 #include <future>
 #include <limits>
 #include <memory>
+#include <mutex>
 #include <stdexcept>
 #include <string>
 #include <thread>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
@@ -40,14 +42,18 @@
 #include "storage/IndexEntryEncryptedLocalWriter.h"
 #include "storage/IndexEntryReader.h"
 #include "storage/EntryStreamUtils.h"
+#include "storage/Crc32cUtil.h"
 #include "storage/PluginLoader.h"
 #include "storage/RemoteInputStream.h"
 #include "storage/RemoteOutputStream.h"
+#include "storage/TailCachedInputStream.h"
 #include "storage/ThreadPools.h"
 
 using namespace milvus::storage;
 
 namespace {
+
+constexpr size_t kTestCachedTailMaxSize = 256 * 1024;
 
 class IndexEntryStreamConfigGuard {
  public:
@@ -408,6 +414,126 @@ class TrackingDelayedInputStream : public milvus::InputStream {
     std::atomic<size_t> max_active_reads_{0};
 };
 
+class CountingInputStream : public milvus::InputStream {
+ public:
+    struct ReadCall {
+        size_t offset;
+        size_t size;
+    };
+
+    explicit CountingInputStream(std::vector<uint8_t> data)
+        : data_(std::move(data)) {
+    }
+
+    size_t
+    Size() const override {
+        return data_.size();
+    }
+
+    bool
+    Seek(int64_t offset) override {
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (offset < 0 || static_cast<uint64_t>(offset) > data_.size()) {
+            return false;
+        }
+        position_ = static_cast<size_t>(offset);
+        return true;
+    }
+
+    size_t
+    Tell() const override {
+        std::lock_guard<std::mutex> lock(mutex_);
+        return position_;
+    }
+
+    bool
+    Eof() const override {
+        std::lock_guard<std::mutex> lock(mutex_);
+        return position_ >= data_.size();
+    }
+
+    size_t
+    Read(void* ptr, size_t size) override {
+        std::lock_guard<std::mutex> lock(mutex_);
+        size_t bytes_read = CopyAt(ptr, position_, size);
+        position_ += bytes_read;
+        return bytes_read;
+    }
+
+    size_t
+    ReadAt(void* ptr, size_t offset, size_t size) override {
+        std::lock_guard<std::mutex> lock(mutex_);
+        read_calls_.push_back({offset, size});
+        return CopyAt(ptr, offset, size);
+    }
+
+    size_t
+    Read(int, size_t) override {
+        return 0;
+    }
+
+    size_t
+    ReadCount() const {
+        std::lock_guard<std::mutex> lock(mutex_);
+        return read_calls_.size();
+    }
+
+    size_t
+    ReadCount(size_t offset, size_t size) const {
+        std::lock_guard<std::mutex> lock(mutex_);
+        return std::count_if(
+            read_calls_.begin(), read_calls_.end(), [&](const ReadCall& call) {
+                return call.offset == offset && call.size == size;
+            });
+    }
+
+ private:
+    size_t
+    CopyAt(void* ptr, size_t offset, size_t size) const {
+        if (offset > data_.size()) {
+            return 0;
+        }
+        size_t bytes_read = std::min(size, data_.size() - offset);
+        if (bytes_read > 0) {
+            std::memcpy(ptr, data_.data() + offset, bytes_read);
+        }
+        return bytes_read;
+    }
+
+    std::vector<uint8_t> data_;
+    mutable std::mutex mutex_;
+    size_t position_{0};
+    std::vector<ReadCall> read_calls_;
+};
+
+TEST(TailCachedInputStreamTest, ServesOnlyFullyCoveredRanges) {
+    std::vector<uint8_t> data(128);
+    for (size_t i = 0; i < data.size(); ++i) {
+        data[i] = static_cast<uint8_t>(i);
+    }
+    auto inner = std::make_shared<CountingInputStream>(data);
+    std::vector<uint8_t> tail(data.begin() + 96, data.end());
+    detail::TailCachedInputStream stream(inner, 96, std::move(tail));
+
+    std::vector<uint8_t> output(32);
+    EXPECT_EQ(stream.ReadAt(output.data(), 96, output.size()), output.size());
+    EXPECT_EQ(std::memcmp(output.data(), data.data() + 96, output.size()), 0);
+    EXPECT_EQ(stream.ReadAt(nullptr, 128, 0), 0);
+    EXPECT_EQ(inner->ReadCount(), 0);
+
+    EXPECT_EQ(stream.ReadAt(output.data(), 95, 2), 2);
+    EXPECT_EQ(inner->ReadCount(95, 2), 1);
+
+    EXPECT_EQ(stream.ReadAt(output.data(), 120, 16), 8);
+    EXPECT_EQ(inner->ReadCount(120, 16), 1);
+
+    EXPECT_EQ(stream.ReadAt(output.data(), 8, 4), 4);
+    EXPECT_EQ(inner->ReadCount(8, 4), 1);
+
+    EXPECT_EQ(stream.ReadAt(output.data(), 129, 1), 0);
+    EXPECT_EQ(inner->ReadCount(129, 1), 1);
+}
+
 }  // namespace
 
 namespace {
@@ -434,6 +560,93 @@ VerifyPattern(const std::vector<uint8_t>& data, size_t expected_size) {
         ASSERT_EQ(data[i], static_cast<uint8_t>(i % 256))
             << "Mismatch at byte " << i;
     }
+}
+
+struct PackedFileImage {
+    std::vector<uint8_t> bytes;
+    std::unordered_map<std::string, std::pair<size_t, size_t>> entry_ranges;
+    size_t meta_offset = 0;
+    size_t meta_size = 0;
+    size_t directory_size = 0;
+};
+
+PackedFileImage
+BuildPackedFileImage(
+    const std::vector<std::pair<std::string, std::vector<uint8_t>>>& entries,
+    const std::string& meta,
+    size_t prefix_padding = 0) {
+    PackedFileImage image;
+    image.bytes.insert(image.bytes.end(),
+                       MILVUS_V3_MAGIC,
+                       MILVUS_V3_MAGIC + MILVUS_V3_MAGIC_SIZE);
+    image.bytes.insert(image.bytes.end(), prefix_padding, uint8_t{0});
+
+    nlohmann::json directory;
+    directory["entries"] = nlohmann::json::array();
+    for (const auto& [name, data] : entries) {
+        size_t absolute_offset = image.bytes.size();
+        size_t data_region_offset = absolute_offset - MILVUS_V3_MAGIC_SIZE;
+        image.bytes.insert(image.bytes.end(), data.begin(), data.end());
+        image.entry_ranges.emplace(
+            name, std::make_pair(absolute_offset, data.size()));
+        directory["entries"].push_back(
+            {{"name", name},
+             {"offset", data_region_offset},
+             {"size", data.size()},
+             {"crc32", Crc32cToHex(Crc32cValue(data.data(), data.size()))}});
+    }
+
+    image.meta_offset = image.bytes.size();
+    image.meta_size = meta.size();
+    image.bytes.insert(image.bytes.end(), meta.begin(), meta.end());
+    directory["entries"].push_back(
+        {{"name", MILVUS_V3_META_ENTRY_NAME},
+         {"offset", image.meta_offset - MILVUS_V3_MAGIC_SIZE},
+         {"size", meta.size()},
+         {"crc32", Crc32cToHex(Crc32cValue(meta.data(), meta.size()))}});
+
+    std::string directory_data = directory.dump();
+    image.directory_size = directory_data.size();
+    image.bytes.insert(
+        image.bytes.end(), directory_data.begin(), directory_data.end());
+
+    if (meta.size() > std::numeric_limits<uint32_t>::max() ||
+        directory_data.size() > std::numeric_limits<uint32_t>::max()) {
+        throw std::runtime_error("Packed test image footer size overflow");
+    }
+    uint8_t footer[MILVUS_V3_FOOTER_SIZE] = {};
+    uint16_t version = MILVUS_V3_FORMAT_VERSION;
+    uint32_t meta_size = static_cast<uint32_t>(meta.size());
+    uint32_t directory_size = static_cast<uint32_t>(directory_data.size());
+    std::memcpy(footer, &version, sizeof(version));
+    std::memcpy(footer + 24, &meta_size, sizeof(meta_size));
+    std::memcpy(footer + 28, &directory_size, sizeof(directory_size));
+    image.bytes.insert(
+        image.bytes.end(), footer, footer + MILVUS_V3_FOOTER_SIZE);
+    return image;
+}
+
+PackedFileImage
+BuildExactSizePackedFile(size_t target_size) {
+    size_t prefix_padding = 0;
+    for (size_t attempt = 0; attempt < 32; ++attempt) {
+        auto image = BuildPackedFileImage(
+            {{"payload", std::vector<uint8_t>{0x5a}}}, "{}", prefix_padding);
+        if (image.bytes.size() == target_size) {
+            return image;
+        }
+        if (image.bytes.size() < target_size) {
+            prefix_padding += target_size - image.bytes.size();
+        } else {
+            size_t excess = image.bytes.size() - target_size;
+            if (prefix_padding < excess) {
+                throw std::runtime_error(
+                    "Packed test image is larger than target size");
+            }
+            prefix_padding -= excess;
+        }
+    }
+    throw std::runtime_error("Failed to build exact-size packed index image");
 }
 
 }  // namespace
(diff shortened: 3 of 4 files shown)


xaxys:
## Provider-level A/B validation

This comment documents the workload separately from the implementation description. The comparison used `upstream/master` at `efbcd352f9edc48f916c51365fc48f485d325347` as the baseline and commit `da6ae6a3824814f7a5b8951d6b0ae015d8af5c10` as the candidate. Both variants used the same Milvus Go binary, runtime dependencies, Docker services, configuration, workload inputs, and reset service volumes. Experimental persistent small-file caching and one-shot prefetching in the research runtime were disabled.

### Dataset

Each row contained an `INT64` primary key; `tenant`, `category`, `price`, `rating`, and `active` scalar fields; a 256-dimensional L2 `semantic_vector`; and a 64-dimensional inner-product `behavior_vector`. The vectors contain 1,280 raw bytes per row.

- Focused workload: two collections, 200,000 rows, approximately 244 MiB of raw vector data.
- Representative workload: three collections, 300,000 rows, approximately 366 MiB of raw vector data.
- Each collection had four approximately equal 25,000-row partitions.
- Rows were inserted in 1,000-row batches over five waves, with a flush after each wave.
- Scalar values deterministically covered 32 tenants, seven categories, prices from 0 to 999.9, ratings from 0 to 4.99, and an approximately 80/20 active/inactive split.

Both vector fields used `IVF_FLAT`: semantic used L2 with `nlist=256`; behavior used inner product with `nlist=128`. The workloads created four scalar inverted indexes per collection.

### Focused RTT20 procedure and result

After background index work quiesced and a 30-second settling period, a fixed 20 ms delay with zero jitter was injected into downstream MinIO requests. Each collection performed one cold load and eight release/load/search rounds. Every load was followed by a one-partition, 256-dimensional semantic search using L2, `nprobe=32`, and top 50, plus a combined filter over tenant, category, price, rating, and `active == true`.

The stable comparison excluded the first warm sample for each collection, leaving 14 warm load/search samples. A matched repeated burst opened the same 32 packed scalar-index objects:

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| Physical GETs | 352 | 72 | **-79.5%** |
| Response bytes | about 2.153 MB | about 1.205 MB | **-44.0%** |
| Warm load mean | 2.788 s | 2.253 s | **-19.2%** |
| HTTP errors | 0 | 0 | unchanged |

This corresponds to 11 GETs per packed object in the baseline and 2.25 GETs per object on average in the candidate. Remaining candidate requests were ranges not fully covered by the retained final tail.

### Three-pair 600-second representative workload

No artificial object-store delay was used. Each run first executed 90 filtered semantic searches, 36 two-vector hybrid searches with RRF reranking, 60 scalar queries, and three compactions. It then repeatedly selected collections round-robin, released and reloaded one collection, and executed a filtered one-partition semantic search using L2, `nprobe=48`, and top 50 until at least 600 seconds elapsed.

Mean provider totals across three baseline/candidate pairs were:

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| All physical GETs | 64,512 | 24,912 | **-61.4%** |
| Packed-index GETs | 49,940 | 10,025 | **-79.9%** |
| Packed-index response bytes | 236.2 MB | 139.9 MB | **-40.8%** |
| HTTP errors | 0 | 0 | unchanged |

Baseline runs completed 121, 124, and 125 steady iterations; candidate runs completed 128, 129, and 130. Candidate mean seconds per iteration was 2.7% lower, but this is not claimed as an isolated throughput improvement because background compaction and index activity varied between repetitions. The representative conclusion is fewer provider requests with no observed correctness, provider-error, or workload regression.

The controlled RTT20 result demonstrates the latency mechanism; the 600-second runs demonstrate that the request-count reduction persists in a mixed workload without relying on injected latency.

czs007:
Adversarial review surfaced one low-severity issue: an error-path regression in `ReadFooterAndDirectory()` that delays magic validation behind the 64 KiB tail read for large files.

### Low

**`internal/core/src/storage/IndexEntryReader.cpp:424` — magic validation no longer fails fast on large corrupt files**

- **Problem:** Magic validation was moved out of `Open()` (where `ValidateMagic()` ran first and needed only an 8-byte read at offset 0) into `ReadFooterAndDirectory()`, where it now runs *after* the tail read. When `file_size_ > 64 KiB` (the `tail_offset != 0` branch), the 64 KiB tail Range GET at line 422-424 is issued before the byte-0 magic read and check at line 431-436. A malformed or corrupt large object therefore costs a 64 KiB GET plus an 8-byte GET before it errors out, instead of failing on 8 bytes as before. The new `CorruptMagicReportsMagicError` test bakes this in by asserting `ReadCount() == 2` for the 70 KiB-padding case.
- **Scope:** Error path only. Valid V3 files are unaffected or improved — `>64 KiB` files still take 2 reads, and `<=64 KiB` files drop from 2 to 1 (the intended win, covered by `ExactTailBoundaryReadCounts`).
- **Suggestion:** In the `tail_offset != 0` branch, hoist the 8-byte magic read and `ValidateMagic()` above the tail Range GET. `tail_offset` is derived from `file_size_` and is known before any read, so reordering costs nothing on the valid path while restoring 8-byte fail-fast for corrupt large files. Update `CorruptMagicReportsMagicError` to expect 1 read for the 70 KiB case.

*(raised by chyezh, tinswzy)*

xaxys:
Accepted and fixed in `0db6a71af5`.

For files larger than 64 KiB, `ReadFooterAndDirectory()` now reads and validates the 8-byte magic before allocating or fetching the 64 KiB tail. Valid files still perform the same two reads and transfer the same bytes; corrupt large files now fail after the header read without issuing the tail GET. Files no larger than 64 KiB keep the intended single full-tail read and validate magic from that buffer.

`CorruptMagicReportsMagicError` now expects one read for both size classes and explicitly verifies that the large corrupt case reads the header once and never reads the final 64 KiB tail. The PR remains a single signed commit, and CI has been retriggered for the updated head.

czs007:
**Overview:** Adversarial review surfaced one low-severity issue — a cancellation-error-code regression in the new `IndexEntryReader::Open()` size validation.

### Low

**`internal/core/src/storage/IndexEntryReader.cpp:343`** — Minimum-size assertion runs before the first cancellation check, masking `FollyCancel`

- **Problem:** The new `AssertInfo(file_size >= kMinFileSize, ...)` (`kMinFileSize` = `MILVUS_V3_MAGIC_SIZE` 8 + `MILVUS_V3_FOOTER_SIZE` 32 = 40) sits at the very top of `Open()`, while the first `reader->CheckCancelled("IndexEntryReader::Open")` is 22 lines later at line 365. In the parent commit (`0d28ea3`) `Open()` had no size validation at all — `CheckCancelled` → `ValidateMagic()` (which itself contains two more cancellation checks) were the first throwing statements, so an already-cancelled token always produced `ErrorCode::FollyCancel`. After this change, when the token is already cancelled *and* the object is truncated (`file_size < 40`), the caller now receives a `SegcoreError` from the size assertion instead of `FollyCancel`. `FollyCancel` is the dedicated code segcore uses to distinguish a deliberate cancellation from a genuine load failure (`EntryStreamUtils.h:51-56` `ThrowInfo(ErrorCode::FollyCancel, ...)`; `LoadCancellationTest.cpp` asserts on it in several places), so a cancellation would be misreported as index corruption. The trigger window is narrow (both conditions must hold, and both are terminal load-failure paths), hence low severity.
- **Suggestion:** Keep the assertion — it is valuable hardening (the parent commit would underflow on `file_size_ - tail_size` after `static_cast<size_t>` of a negative `file_size` inside `ReadFooterAndDirectory`). Instead, move the cancellation check ahead of it by calling `ThrowIfCancelled(cancellation_token, "IndexEntryReader::Open")` directly before the size assertion. This is safe for the new tests: `IndexEntryReaderTailCacheTest.InvalidSignedAndSmallSizesDoNotRead` uses a default (non-cancelled) token and still trips the size assertion for `int64_min`/`-1`/`0`/`39`, while `PreCancelledOpenDoesNotRead` uses a valid-size mirror and continues to surface `FollyCancel`.

*(raised by tinswzy, chyezh)*

xaxys:
Accepted and fixed in `33f09ca1d7`.

`IndexEntryReader::Open()` now calls `ThrowIfCancelled(cancellation_token, "IndexEntryReader::Open")` before the minimum-size and addressability checks. This restores the previous error precedence when cancellation and malformed input are both present: an already-cancelled operation reports `FollyCancel`, while a non-cancelled invalid size still fails before allocation or storage I/O.

The existing cancellation check after reader initialization is intentionally retained so cancellation requested after the early check but before the first read is still observed. The only normal-path cost is one additional cancellation-token state check.

`PreCancelledOpenDoesNotRead` now covers both a valid packed-file size and `file_size = 39`; both cases explicitly assert `ErrorCode::FollyCancel` and zero provider reads. The PR remains a single signed commit, and CI has been retriggered for the updated head.

czs007:

xiaofan-luan:
Good catch — the read amplification is real, the diagnosis is right, and the provider-level A/B numbers are convincing. Two asks before this goes in.

### 1. Remove `small_entry_cache_` rather than adding a second cache beside it

`small_entry_cache_` came in with the V3 format in #47690 and is never hit on any production path today:

- `GetMeta` / `HasMeta` read `meta_json_`, not `ReadEntry`, so `__meta__` is read exactly once (in `Open`).
- Every `ReadEntry` caller reads each name once per reader: `ScalarIndexSort` / `StringIndexSort` (`valid_bitset`), `BitmapIndex` (`BITMAP_INDEX_VALID_BITSET`), `InvertedIndexTantivy` (`INDEX_NULL_OFFSET_FILE_NAME`), `NgramInvertedIndex`, `RTreeIndex`, `JsonHybridScalarIndex` / `JsonScalarIndexWrapper`. The last two share an entry name but cannot nest — `HybridScalarIndex::GetInternalIndex()` only produces BITMAP / STLSORT / MARISA / INVERTED, and `JsonScalarIndexWrapper<T, BaseIndex>` is only instantiated over `ScalarIndexSort` / `BitmapIndex` / `InvertedIndexTantivy`.
- `ReadEntryStream` / `ReadEntryToFile` / `ReadEntriesToFiles` / `ReadEntriesStreamToFiles` neither consult nor populate it — and that is where the large entries went after they were migrated off `ReadEntry`.

So it is write-only: every `ReadEntry` result under 1 MiB gets a full copy that lives until the reader is destroyed, and nothing ever reads it back. The only two `ReadEntry` calls on the same name in the tree are in its own unit test.

With the tail cache added, tail-resident entries would be held three times over — tail buffer, `small_entry_cache_`, and the returned `Entry`. Please drop it here instead of leaving two caches side by side.

### 2. Make the retained tail 256 KiB, and change the write path to match

64 KiB is sized for footer + directory, not for the thing we actually want to eliminate. Your own numbers show 2.25 GETs per object remaining — those are small entries falling outside the retained tail.

The layout today is `magic | entries (in write order) | __meta__ | directory | footer`, so whether `valid_bitset` or a null-offset array lands in the tail is an accident of write order and total file size. If `IndexEntryWriter` buffers entries below a size threshold and emits them last, immediately before `__meta__`, then one 256 KiB tail read covers footer + directory + `__meta__` + every small entry, at any file size. That puts the common case at **1 GET per packed object** instead of leaving it dependent on the packed-file size distribution.

Two notes:

- Do this in the writer, not by asking each of the eight index implementations to order their `WriteEntry` calls — the invariant should have one owner.
- It is a writer-side change, so existing files keep working; they just get the smaller reader-side benefit.

With that layout change, 256 KiB has a rationale (it is sized to hold the small-entry region). Without it, a bigger number is just a bigger arbitrary number.

Memory is not a concern here: the buffer is already allocated today and the change is only to retention, and removing `small_entry_cache_` gives back more than the extra 192 KiB costs.

xiaofan-luan:
One more, on where this lives. Not a correctness objection — the coverage predicate and the retention policy are both written correctly. It is about which layer owns the invariant.

### The change is mostly plumbing

`IndexEntryReader.cpp` is +137/-62. Roughly:

| | lines |
|---|---:|
| `ReadContext::ReadAt` coverage predicate | ~19 |
| tail retention policy at the end of `ReadFooterAndDirectory` | ~12 |
| magic validation reorder | ~15 |
| `Open` argument validation | ~15 |
| two structs in the header | ~13 |
| **actual change** | **~74** |
| 12 × `input_->ReadAt` → `read_ctx_->ReadAt` | |
| 8 × rewritten lambda capture lists | |
| `auto read_ctx = read_ctx_;` plus re-capturing `cipher_plugin` / `edek` / `ez_id` / `collection_id` | |
| **plumbing** | **~95** |

More than half of what a reviewer has to read is unrelated to "do not download bytes we already have".

Four of those eight capture-list rewrites — both lambdas in `SubmitEntryStreamDownloadTasks`, and the `load_slice` in `ReadPlainEntryStream` and `ReadEncryptedEntryStream` — were already `auto input = input_;` by-value captures. They only got renamed to `auto read_ctx = read_ctx_;`. No semantics changed there.

### Why the layer looks wrong

`ReadContext::ReadAt` decides using only `(offset, size)` and a retained byte range. It needs to know nothing about magic, footer, directory, entries, CRC, encryption or slices — the entire knowledge of `IndexEntryReader`, none of which participates in the decision. Logic that touches none of its host's state usually does not belong to that host.

The practical consequence: the reader now has two ways to reach bytes, and nothing structural stops a future contributor from adding a 13th call site that goes around the cache. It relies on everyone remembering to use `read_ctx_`. A decorator makes the uncached stream unreachable instead — `input_` *is* the cached stream.

### What it would look like

`milvus::InputStream` has six pure virtuals. `ReadAt` is the thread-safe random-access primitive; `Seek`/`Tell`/`Read` are the stateful cursor API. Only `ReadAt` needs intercepting, and since `ReadAt` never moves the cursor, forwarding the rest is safe.

```cpp
// Serves fully covered ranges from a retained suffix; everything else falls through.
class TailCachedInputStream : public milvus::InputStream {
 public:
    TailCachedInputStream(std::shared_ptr<milvus::InputStream> inner,
                          size_t tail_offset,
                          std::vector<uint8_t> tail)
        : inner_(std::move(inner)),
          tail_offset_(tail_offset),
          tail_(std::move(tail)) {
    }

    using milvus::InputStream::Read;  // keep the base Read<T> template visible

    size_t
    ReadAt(void* ptr, size_t offset, size_t size) override {
        if (offset >= tail_offset_) {
            const size_t rel = offset - tail_offset_;
            if (rel <= tail_.size() && size <= tail_.size() - rel) {
                if (size > 0) {
                    fastmem::FastMemcpy(ptr, tail_.data() + rel, size);
                }
                return size;
            }
        }
        return inner_->ReadAt(ptr, offset, size);
    }

    size_t Size() const override { return inner_->Size(); }
    size_t Tell() const override { return inner_->Tell(); }
    bool   Eof()  const override { return inner_->Eof(); }
    bool   Seek(int64_t o) override { return inner_->Seek(o); }
    size_t Read(void* p, size_t n) override { return inner_->Read(p, n); }
    size_t Read(int fd, size_t n)  override { return inner_->Read(fd, n); }

 private:
    const std::shared_ptr<milvus::InputStream> inner_;
    const size_t tail_offset_;
    const std::vector<uint8_t> tail_;
};
```

Installed at the end of `ReadFooterAndDirectory`, replacing the `read_ctx_->cached_tail = ...` block:

```cpp
    std::vector<uint8_t> retained;
    if (tail_data.size() <= kCachedTailMaxSize) {
        retained = std::move(tail_data);
    } else {
        // vector::erase does not reduce capacity, so copy into a fresh bounded
        // allocation rather than shrinking in place.
        retained.assign(tail_data.end() - kCachedTailMaxSize, tail_data.end());
    }
    input_ = std::make_shared<TailCachedInputStream>(
        std::move(input_), file_size_ - retained.size(), std::move(retained));
```

Your reasoning for avoiding `vector::erase` is right and is preserved here. Note the `assign` must target a *different* vector — assigning from a container's own iterators back into itself is undefined; the current code already gets this right and it should stay that way through any refactor.

The `input_` member stays, so **all 12 call sites are untouched**. The four lambdas that capture `this` become `auto input = input_;` by-value captures — which is what four other lambdas in this same file already do, so it is more internally consistent, not less.

### What it buys

1. **Review surface goes from ~170 lines to ~45** — one 40-line class plus a 6-line install site.
2. **The coverage predicate becomes unit-testable on its own.** Much of `BuildPackedFileImage` / `BuildExactSizePackedFile` exists because the only way to reach the predicate today is through a full V3 `Open`. With a decorator it is a table test: boundary hit, off-by-one, `size == 0`, front overlap, back overlap, entirely before, entirely after, empty tail. No footer, no CRC, no JSON. The packed-image tests then only need to assert the *policy* — what gets retained — rather than the mechanism.
3. **The retention policy becomes a construction parameter.** Whether to keep the last 64 KiB, the last 256 KiB, or the whole object for small files turns into one line at the install site instead of a change to the cache implementation.
4. **It closes the bypass structurally** rather than by convention.

### Costs, honestly

- Six forwarding methods, roughly 15 lines of boilerplate, plus the `using Read;` needed to avoid hiding the base template. Real, but cheaper than 12 call sites and 8 capture lists.
- One extra virtual indirection per `ReadAt`. Unmeasurable against a 16 MiB range read or a cache-hit memcpy.
- `input_` no longer refers to the object the caller passed in after `Open`. Nothing depends on that today; renaming the member to `source_` would make it explicit if you prefer.
- If the motivation for `shared_ptr<ReadContext>` was lifetime rather than caching: by-value capture of `input_` gives identical lifetime properties, and four lambdas in this file already rely on exactly that.

Happy to be talked out of it if there is a constraint I am missing.

xaxys:
Thanks for the detailed review. I audited the production call graph and updated the implementation in `8ab3cd0eef`.

### Removed `small_entry_cache_`

I confirmed that production populates `small_entry_cache_` but does not hit it: metadata is served from `meta_json_`, production `ReadEntry()` callers read each entry name at most once per reader, and the streaming/file-download paths do not use this cache.

I therefore removed `small_entry_cache_` and its 1 MiB threshold. This avoids retaining and copying an unused second copy of small entries.

This also improves the lifecycle of encrypted data. `ReadEncryptedEntry()` returns plaintext, and the old cache copied that plaintext into reader-owned storage until reader destruction. This was not a security vulnerability, but it unnecessarily extended the plaintext lifetime and increased copying and peak-memory pressure.

### Moved caching below `IndexEntryReader`

The retained tail is now implemented as a `TailCachedInputStream` decorator:

- `input_` remains the reader's only byte source;
- fully covered `ReadAt()` calls are served from the retained suffix;
- all other operations are forwarded to the original stream;
- partial overlaps fall through without stitching, preserving existing short-read and provider-error semantics;
- future reader call sites cannot accidentally bypass the cache.

Only the final bounded 256 KiB suffix is retained. If directory or metadata parsing temporarily requires a larger buffer, the retained suffix is copied into a fresh bounded allocation rather than keeping the oversized allocation for the reader lifetime.

Asynchronous tasks continue to capture `input`, cipher state, EDEK, and IDs by value. Although the public operations drain their futures before returning, the tasks themselves execute asynchronously. Value capture avoids access through `this` and gives each queued task stable ownership of the resources it needs.

### Where 256 KiB helps

The useful boundary is more specific than simply "small versus large files." Increasing the retained tail to 256 KiB eliminates an additional request when the tail-resident working set is larger than 64 KiB but no larger than 256 KiB.

For common V3 side entries, the approximate coverage boundaries are:

| Entry | Serialized size | 64 KiB coverage | 256 KiB coverage |
|---|---:|---:|---:|
| validity bitset | `rows / 8` | about 524K rows | about 2.1M rows |
| `idx_to_offsets` | `rows * 4` | about 16K rows | about 65K rows |
| null offsets | `null_count * 8` | about 8K nulls | about 32K nulls |

Directory and metadata reduce these limits slightly.

As a concrete example, a 50K-row numeric STLSORT has about 200 KiB of `idx_to_offsets`, so 256 KiB can cover that tail entry while 64 KiB cannot. At 100K rows, the same entry is about 400 KiB, so neither limit fully covers it and 256 KiB is unlikely to remove another request for that entry.

The focused workload represents the other important case: its 100K-row low-cardinality inverted indexes produced 11-90 KiB packed objects. Twenty-four of the 32 objects already fit in 64 KiB, while the remaining eight were between 64 and 90 KiB. Increasing the limit to 256 KiB therefore moved all 32 objects to one GET.

High-cardinality strings, UUID-like values, text/ngram indexes, large Tantivy files, and geometry indexes will generally exceed 256 KiB much earlier. For those files, the incremental benefit depends on whether a tail side entry falls into the 64-256 KiB window; 256 KiB is not expected to eliminate an additional request for every packed index.

### Configuration

As for whether make the 256 KiB limit be exposed as a configuration, I prefer not. It is a transitional reader-side heuristic, not a stable product or file-format contract. Once the writer records the hot-cache region as `(offset, length)` in file metadata, the reader should follow the per-file declaration and this global threshold will lose its purpose.

Publishing a configuration now would create compatibility, documentation, operational, and test obligations for a setting that we already expect to deprecate. Removing a released configuration is substantially harder than replacing an internal constant, so I prefer to keep 256 KiB internal unless there is a concrete operational requirement for runtime tuning.

The updated A/B and correctness results are posted separately. The longer-term writer-declared hot-region design remains tracked by #51847.

xaxys:
### Updated A/B results: 64 KiB vs. 256 KiB retained tail

The 64 KiB and 256 KiB results come from two separate paired A/B campaigns. Each campaign had its own contemporaneous upstream control:

- the **64 KiB baseline** is the upstream run paired with the earlier 64 KiB candidate;
- the **256 KiB baseline** is the upstream run paired with the current 256 KiB candidate.

The focused baselines reproduced the same 352-GET burst and nearly identical latency and bytes. The representative baselines differ slightly because the later campaign completed more steady iterations, so the paired percentage changes are more meaningful than directly comparing its raw totals.

#### Focused RTT20 workload

The workload used two collections, four partitions per collection, 100,000 rows per collection, two vector fields, and four scalar inverted indexes per collection. Each run performed an initial load followed by repeated release/load/search cycles with a fixed 20 ms MinIO downstream delay and zero jitter.

For the stable 32-file packed-index burst:

| Metric | 64 KiB baseline | 64 KiB | 256 KiB baseline | 256 KiB |
|---|---:|---:|---:|---:|
| Physical GETs | 352 | 72 | 352 | 32 |
| GETs per file | 11 | 2.25 | 11 | 1 |
| Response bytes | approximately 2.153 MB | approximately 1.205 MB | approximately 2.153 MB | approximately 1.183 MB |
| Warm-load mean | 2.788 s | 2.253 s | 2.804 s | 2.271 s |
| HTTP errors | 0 | 0 | 0 | 0 |

| Paired comparison | Physical GETs | Response bytes | Warm-load mean |
|---|---:|---:|---:|
| 64 KiB baseline -> 64 KiB | -79.5% | approximately -44.0% | -19.2% |
| 256 KiB baseline -> 256 KiB | -90.9% | approximately -45.1% | -19.0% |

The two focused baselines are effectively identical: both produced 352 GETs for the same 32 packed objects, while warm-load means differed by only 0.6%. This makes the candidate-side request-count difference directly interpretable:

| Candidate comparison | Physical GETs | Response bytes | Warm-load mean |
|---|---:|---:|---:|
| 64 KiB -> 256 KiB | -55.6% | approximately -1.85% | +0.8% |

The packed-file size distribution explains the result:

- minimum: 11,624 bytes;
- median: 17,777 bytes;
- p75: 30,586 bytes;
- maximum: 89,507 bytes;
- 24 of 32 files were at most 64 KiB;
- the remaining 8 files were between 64 and 90 KiB.

With 64 KiB, the resulting request count was:

`24 * 1 + 8 * 6 = 72 GETs`

With 256 KiB, the initial read covered all 32 files:

`32 * 1 = 32 GETs`

The incremental benefit is therefore request elimination rather than reduced transfer volume. The 256 KiB version removed 55.6% of the GETs remaining with 64 KiB, while transferred bytes decreased by only about 1.85%.

Warm-load time changed from 2.253 s to 2.271 s. This 0.8% difference is within run-to-run noise, so I do not claim an additional latency improvement from 64 KiB to 256 KiB. The supported conclusion is that 256 KiB brings this observed normal packed-file distribution to one GET per object.

#### Representative 600-second workload

Each run used three collections, four partitions per collection, and 100,000 rows per collection. It executed 90 filtered semantic searches, 36 hybrid searches, 60 scalar queries, three compactions, and then repeated release/load/filtered-search cycles for at least 600 seconds without injected S3 latency.

Mean provider totals across three repetitions:

| Metric | 64 KiB baseline | 64 KiB | 256 KiB baseline | 256 KiB |
|---|---:|---:|---:|---:|
| All physical GETs | 64,512 | 24,912 | 66,749 | 18,613 |
| Packed-index GETs | 49,940 | 10,025 | 53,049 | 4,897 |
| Packed-index response bytes | approximately 236.2 MB | approximately 139.9 MB | approximately 246.2 MB | approximately 138.7 MB |
| HTTP errors | 0 | 0 | 0 | 0 |

| Paired comparison | All GETs | Packed-index GETs | Packed-index bytes |
|---|---:|---:|---:|
| 64 KiB baseline -> 64 KiB | -61.4% | -79.9% | -40.8% |
| 256 KiB baseline -> 256 KiB | -72.1% | -90.8% | -43.7% |

The later representative baseline completed more steady iterations and consequently recorded approximately 3-6% higher provider totals. For that reason, the paired reductions above are the primary comparison.

The candidate raw totals changed as follows:

| Candidate comparison | All GETs | Packed-index GETs | Packed-index bytes |
|---|---:|---:|---:|
| 64 KiB -> 256 KiB | -25.3% | -51.2% | approximately -0.9% |

Despite the later campaign executing more steady work, the 256 KiB candidate still recorded substantially fewer packed-index GETs. This is consistent with the exact focused result: increasing the retained range eliminates about half of the requests remaining with 64 KiB, while transferred bytes remain nearly unchanged.

Steady iteration time remained within normal run-to-run variance, and no RSS regression was observed. I therefore do not attribute a throughput or memory improvement to the 64-to-256 KiB change.

#### Correctness coverage

`IndexEntryReaderTailCacheTest.*` was run three times: 9 tests per run, 27/27 passing. Coverage includes:

- 256 KiB `-1`, exact, and `+1` boundaries;
- full cache hits and partial-overlap fallback;
- oversized directory and metadata;
- corrupt-magic fail-fast behavior;
- short reads;
- pre-cancelled open;
- negative, zero, and undersized file sizes.

czs007:
Adversarial review surfaced one medium-severity lifetime-safety gap in the async read paths of `IndexEntryReader`, where the by-value capture hardening described in the PR body is only partially applied.

## Medium

**`internal/core/src/storage/IndexEntryReader.cpp:639` — By-value capture hardening is incomplete: orphaned tasks can still write to a destroyed `dest`/`state` and a closed `fd`.**

The PR body states that `input`/`cipher_plugin` were changed to by-value captures because "ThreadPool::Submit can enqueue work before returning its future … if submission throws after enqueueing." However, the same lambdas still capture raw pointers and references: `ReadPlainEntry` captures `dest` (pointing into the local `Entry result.data`) and `&pm`; `ReadEncryptedEntry` captures `dest`; `SubmitEntryDownloadTasks` / `SubmitEntryStreamDownloadTasks` capture `&state` and `fd = state.fd` (lines 814/864/984/1035). In `ThreadPool::Submit`, `work_queue_.enqueue(wrap_func)` (ThreadPool.h:163) happens before the `std::thread` construction at line 179, which can throw `std::system_error`, and before `return task_ptr->get_future()` at line 185.

Failure scenario: loading a >16 MiB packed index entry (parallel-slice path in `ReadPlainEntry`) or taking the `ReadEntryToFile` path while the process is at its thread limit → the task is already enqueued but its future never reaches `futures`, so `DrainFutures` (639/657/732/1093) does not wait for it; the exception propagates, `Entry result` is destroyed and `ReadEntryToFile`'s catch block runs `::close(state.fd)` (1101-1104), while the orphaned worker still writes decrypted bytes into `dest`, updates `state.range_crcs`, and `pwrite`s to a closed (possibly recycled) fd — a heap use-after-free and index bytes landing in an unrelated file.

Suggestion: give the buffers and download state shared ownership (e.g. a `shared_ptr` backing buffer and a `shared_ptr<EntryDownloadState>` captured by value), or add a pool-level drain that runs when `Submit` throws, so the invariant the PR body claims actually holds for every captured resource rather than only for `input_`. Note this predates the PR, but the PR body describes it as already solved. *(surfaced during verification)*

xiaofan-luan:
Thanks for the updates. The tail-cache implementation, layering, and provider-level A/B evidence now look solid. I still have two merge-blocking asks.

1. Please complete the async lifetime fix raised in [this review](https://github.com/milvus-io/milvus/pull/51696#issuecomment-5093108055). Capturing `input` by value is not sufficient: orphaned tasks can still access `dest`, `state`, raw writers/fds, and the encrypted stream path still carries borrowed `this`/`em` state. Preferably establish the invariant at the `ThreadPool::Submit` layer so a task cannot be enqueued without the caller receiving a future, and add a deterministic regression test for the post-enqueue failure path.

2. Please refresh the PR description to match the current implementation. The code and latest validation use a 256 KiB retained tail, while the main body still describes a 64 KiB limit and 64 KiB boundary tests. Please also make the supported performance claim explicit: 256 KiB further reduces GET count, but does not demonstrate an additional latency improvement over 64 KiB.

The writer-declared hot-region work tracked by #51847 does not need to block this PR. Once these two items are addressed and the required CI/approval gates pass, I am comfortable with merging.

xaxys:
Thanks for surfacing this. The failure sequence is technically valid, but after auditing the shared thread-pool call graph we are intentionally not adding an `IndexEntryReader`-local submission gate in this PR.

The trigger is exceptionally narrow: a task must already have been enqueued, and then `ThreadPool::Submit()` must throw before returning its future. The only realistic path is dynamic `std::thread` construction failing with a recoverable `EAGAIN`/`ENOMEM` while the pool is expanding and the process remains alive. The potential impact remains real, but the occurrence condition is an infrastructure-level edge case rather than behavior introduced by the tail-cache change.

The audit found 31 direct production users of this `ThreadPool::Submit()` in current `upstream/master`: 7 in `IndexEntryReader` and 24 outside it. None implements a commit/cancel submission gate or can track a task for which `Submit()` did not return a future. The other users are:

- storage: `DiskFileManagerImpl`, `IndexEntryEncryptedLocalWriter`, and `Util` (`GetObjectData`, `PutIndexData`, and Storage V2 load paths);
- index/json stats: `JsonKeyStats`;
- segcore: `ChunkedSegmentSealedImpl`, `SegmentGrowingImpl`, `memory_planner`, `Utils`, `reduce/Reduce`, and `search_result_export_c`.

Several of those paths already use `WaitAllFutures`, `DrainFutures`, or scope guards. Those protections drain futures that were successfully returned; they do not cover the post-enqueue/no-future window. Some other paths are naturally lower risk because their tasks capture only values or shared ownership, while others capture `this`, stack references, raw buffers, raw readers, or result pointers and therefore share the same underlying contract gap.

A reader-local gate was prototyped, but keeping it here would duplicate an invariant that belongs to `ThreadPool::Submit()`, add an extra gate allocation/atomic synchronization to only one consumer, and require removal once the shared thread pool is fixed. The local gate has therefore been removed in `587c80c2d5`.

The shared follow-up is now tracked by #51919 and implemented in #51920. The pool-level change provisions any required expansion worker before publishing the task, so worker-creation failure rejects an unaccepted submission instead of creating a queued task with no returned future. Queue publication is then the acceptance boundary. The final implementation stays directly in the existing `Submit()` template and does not add a test peer, injectable worker factory, or reader-local ownership wrapper. Detailed validation and the fault-injection trade-off are documented in https://github.com/milvus-io/milvus/pull/51920#issuecomment-5100116968. This PR description has also been updated to avoid claiming that by-value input capture solves the separate submission-lifetime window.

czs007:
Adversarial review of the tail-cache change surfaced one pre-existing concurrency hazard and three smaller documentation/test-coverage gaps; none block the merge, but the PR body needs two corrections and follow-ups should be filed.

### Medium

- **`IndexEntryReader.cpp:638` — Async tasks still capture `dest`/`&pm`/`&state` raw, allowing use-after-free if thread creation fails.** `ThreadPool::Submit` enqueues the task (`ThreadPool.h:163`) before the dynamic-scaling branch constructs a `std::thread` (`ThreadPool.h:179`), which can throw `std::system_error` under thread/resource exhaustion. In that window the task is queued but its future never reaches the caller, so `DrainFutures` (`IndexEntryReader.cpp:129-143`) won't wait on it; after the driver unwinds, an existing worker can execute the orphaned lambda and dereference the freed output buffer (`dest`, plain path :638-649, encrypted :697-727) or the dangling local `&state` (file paths :804-845/:859-881, stream paths :974-985/:1029-1036). Note this is pre-existing — the old lambdas captured `dest`/`&pm`/`&state` the same way, and this PR's by-value captures (`input`, `cipher_plugin`, `edek`, …) actually shrink the dangling surface. The trigger window is extremely narrow, so this need not block the PR; recommend a dedicated follow-up that either makes `Submit` atomic (enqueue only after thread creation succeeds) or has tasks own their output buffers. *(raised by chyezh, tinswzy)*

### Low

- **`IndexEntryWriterTest.cpp:1028` — Read-count assertions cover plain entries only, and the PR body over-claims encrypted coverage.** All nine `IndexEntryReaderTailCacheTest` cases (:1011-1182) assert read counts on plaintext entries; no encrypted case exercises `ReadCount`/`CountingInputStream`, even though `ReadEncryptedEntry` was rewritten identically and reads through the same `TailCachedInputStream` wrapper (`IndexEntryReader.cpp:697-711`). Meanwhile the PR body's "Validation" section claims coverage of "encrypted tail reuse for in-memory, file-output, and streaming paths, including decrypted-content verification" — that coverage does not exist at the pushed head. Either add encrypted variants (e.g. an encrypted `FullCoverageHitsButPartialOverlapFallsBack`) or correct the PR body. *(raised by tinswzy)*

- **`IndexEntryReader.h:254` — `small_entry_cache_` (the ≤1MiB in-memory entry cache) is removed wholesale, and the PR body does not disclose this behavior change.** The `ReadEntry` cache lookup/insert, `kSmallEntryCacheThreshold`, the member map, and the `SmallEntryCache` test are all deleted. The replacement tail cache only covers the final 256KiB of the file, so repeat `ReadEntry` calls for small entries outside the tail now go to object storage every time instead of hitting memory. This may be a reasonable trade-off (the old cache had no eviction; A/B shows a large total-GET reduction), but the PR body describes only the added tail cache and never mentions the removal, leaving reviewers unable to evaluate it. Please disclose the removal in the PR body with the rationale (e.g. entries are read only once per load, so the cache never paid off). *(surfaced during verification)*

### Nitpick

- **`IndexEntryReader.cpp:126` — The 256KiB tail-size constant is duplicated between production code and tests.** `kCachedTailMaxSize` lives in an anonymous namespace, so the tests redefine it as `kTestCachedTailMaxSize` (`IndexEntryWriterTest.cpp:56`), and the boundary cases in `ExactTailBoundaryReadCounts` (:1011-1025) depend on the test-side copy. If the production value is ever retuned, the boundary tests would silently keep building ±1 files against the old value. Suggest promoting the constant to a `static constexpr` member of `IndexEntryReader.h` (or a dedicated header) and referencing it directly from tests. *(raised by tinswzy)*

xaxys:
Thanks — I re-audited these points against the current head.

- The `ThreadPool::Submit` lifetime hazard is valid and pre-existing. It is now tracked in #51919 and fixed at the shared pool layer in #51920. The final pool design provisions any required expansion worker before queue publication, so worker-creation failure rejects an unaccepted task instead of leaving an orphaned queued lambda. I am keeping #51696 free of a reader-local ownership gate.
- The encrypted-validation observation is correct. The nine tail-cache integration cases use plaintext images, and the existing mock-cipher cases test encrypted writing rather than encrypted reader GET counts/decryption. I corrected the PR body and filed #51921 for a normal-public-API encrypted reader fixture backed by a loadable test cipher plugin.
- The `small_entry_cache_` disclosure gap is also correct. The PR body now documents its removal, the audited no-hit production access pattern, the repeated-read trade-off, the unbounded allocation/copy cost, and the shorter plaintext lifetime for encrypted entries.
- I am not promoting the 256 KiB policy constant into `IndexEntryReader.h`. The test-side copy cannot silently mask a retune: if production decreases the value, the old `max-1` case changes from one read to two; if production increases it, the old `max+1` case changes from two reads to one. Either direction fails the boundary expectations. Exposing an internal heuristic through the reader header would add API/test coupling without improving detection.

No source change is needed for the last point; the two PR-body corrections and both follow-ups are now in place.

czs007:
## Re-review of `587c80c2d5a4`

This round re-verified all four previously posted findings against the unchanged head; three are confirmed and one is retracted.

- **Confirmed** — `IndexEntryReader.cpp:638` (async tasks capture raw `dest`/`&pm`/`&state`; use-after-free if thread creation fails): the full failure sequence still holds on this head, and the shared `ThreadPool::Submit` fix is not an ancestor of it.
- **Confirmed** — `IndexEntryWriterTest.cpp:1028` (read-count assertions cover only plain entries): all new counting tests build plaintext entries; the encrypted path still has no hit/fallback regression guard.
- **Confirmed** — `IndexEntryReader.cpp:126` (256 KiB tail cap duplicated between production and tests): both constants remain independently defined, so the ±1 boundary tests can silently drift from the real boundary.
- **Retracted** — `IndexEntryReader.h:254` (`small_entry_cache_` removal "not disclosed in the PR body"): our earlier comment was wrong. The code removal itself is real, but reviewers verified it was explicitly requested by the maintainer and openly discussed and evaluated in the PR conversation — an audit showed production populated the cache but never hit it, since metadata is served from `meta_json_`. The "undisclosed tradeoff" premise does not hold.

---

Adversarial re-review of this head confirms one open concurrency-lifetime hazard plus two non-blocking test/maintainability issues; no new critical or high-severity problems were found.

### Medium

- **`internal/core/src/storage/IndexEntryReader.cpp:638` — Orphaned async tasks can use-after-free when `ThreadPool::Submit` fails during worker creation.** `ThreadPool::Submit` enqueues the task (`ThreadPool.h:163`) before creating a worker thread (`ThreadPool.h:179`), which can still throw `std::system_error`. If it throws, the task is already accepted but no future is returned, so `DrainFutures` (`IndexEntryReader.cpp:657`) cannot wait for it. The lambdas in `ReadPlainEntry` (lines 638–649) capture raw `dest` (pointing into the local `result.data`) and `&pm`, the encrypted path (line 697) captures raw `dest`, and `SubmitEntryDownloadTasks` (lines 804, 856–863) captures `&state`. After stack unwinding destroys the local buffers, an idle worker can still execute the orphaned lambda, writing to freed memory or touching a closed/reused fd. This is a pre-existing defect; the shared-pool fix (#51920, commit `86568ae067`) is still open and not an ancestor of this head. **Suggestion:** ensure that fix, or equivalent atomic-submission semantics, lands together with this change before merging. *(raised by tinswzy, chyezh)*

### Low

- **`internal/core/src/storage/IndexEntryWriterTest.cpp:1028` — Tail-cache read-count assertions cover only plaintext entries; the encrypted path has no hit/fallback regression guard.** All new counting tests (`TailCachedInputStreamTest`, `IndexEntryReaderTailCacheTest`, including `FullCoverageHitsButPartialOverlapFallsBack`) are built on plaintext entries, while `ReadEncryptedEntry` goes through the same `TailCachedInputStream` wrapper with no `CountingInputStream`/`ReadCount` assertion anywhere in the diff (the existing EncryptedV3 tests use file streams and do not count reads). A regression causing encrypted slices fully inside the retained tail to still issue provider reads, or mishandling partial overlap, would go undetected. **Suggestion:** add encrypted full-hit and partial-overlap request-count test cases through `IndexEntryReader`. *(raised by tinswzy, chyezh)*

### Nitpick

- **`internal/core/src/storage/IndexEntryReader.cpp:126` — The 256 KiB tail cap is defined twice, in production code and in tests.** `kCachedTailMaxSize` lives in an anonymous namespace (invisible to tests), and `IndexEntryWriterTest.cpp:56` independently defines `kTestCachedTailMaxSize` with the same value. Boundary tests such as `ExactTailBoundaryReadCounts` (lines 1013–1016) depend on the test-side constant, so changing only the production value would leave these ±1 tests green while no longer exercising the real boundary. **Suggestion:** expose a single constant shared by production code and tests. *(raised by tinswzy, chyezh)*

xaxys:
Thanks for the re-review. I agree with the encrypted-path coverage observation, but I do not think it requires a change in this PR. I disagree with the duplicated-constant finding and am not adopting that suggestion.

### Encrypted tail-cache request-count coverage

The current counting tests use plaintext packed images. They verify full tail hits, partial-overlap fallback, the 256 KiB `-1`/exact/`+1` boundaries, oversized directory/meta handling, corrupt magic, cancellation, and invalid sizes. The existing `IndexEntryEncryptedV3Test` cases exercise encrypted writing, multi-slice layout, multiple entries, fd input, and large metadata; they do not open the encrypted object through `IndexEntryReader` with a `CountingInputStream`.

This is a real Low-severity end-to-end regression-coverage gap, not evidence that the encrypted implementation bypasses the cache. `TailCachedInputStream` is installed below `IndexEntryReader`, and `ReadEncryptedEntry` captures the already-decorated `input_` and calls the same virtual `ReadAt()` path:

```text
ReadEncryptedEntry
  -> captured input
  -> TailCachedInputStream::ReadAt
  -> retained-tail hit or wrapped-provider fallback
```

The wrapper's hit/fallback behavior has direct unit coverage, the plaintext reader cases verify that the reader installs and uses it, and the encrypted path consumes that same wrapper. What is missing is encrypted public-API integration coverage that also verifies decrypted content and provider-read counts.

Adding that correctly requires a loadable test cipher plugin because the public `IndexEntryReader::Open()` resolves encrypted readers through `PluginLoader`. The current `MockCipherPlugin` is passed directly to the encrypted writer and is not registered as a loadable reader plugin. I do not want to reintroduce an `Open(..., cipher_plugin)` overload or another production injection seam solely for a test. The normal-public-API fixture is tracked by #51921, where it can cover encrypted full-hit and partial-overlap behavior without changing the reader API. This gap is therefore acknowledged but non-blocking for #51696.

### Independent 256 KiB test constant

The two constants are independently defined, but the claim that they can silently drift while the boundary test remains green is not correct. `ExactTailBoundaryReadCounts` asserts:

```text
256 KiB - 1 -> 1 read
256 KiB     -> 1 read
256 KiB + 1 -> 2 reads
```

Any production-threshold change is detected:

- if production decreases to 128 KiB, the fixed 256 KiB case performs two reads instead of the expected one;
- if production increases to 512 KiB, the fixed `256 KiB + 1` case performs one read instead of the expected two;
- even a one-byte decrease breaks the exact case, and a one-byte increase breaks the `+1` case.

Keeping the test value independent is intentional: it acts as a policy-change sentinel and forces a maintainer to update the test explicitly when retuning the production heuristic. If the test imported the production constant, both the test inputs and implementation boundary would move together, so a policy change would no longer require that explicit review.

Exposing this temporary reader heuristic through `IndexEntryReader.h` would also couple the production header to a test policy and create a symbol that must later be removed when the writer-declared hot region `(offset, length)` replaces the fixed threshold. I am therefore keeping the constants independent and consider this nitpick resolved without a source change.

czs007:
Adversarial re-review of the unchanged head `587c80c2d5a4` re-confirms one pre-existing medium lifetime hazard and one low test-coverage gap, and retracts two earlier findings (plus their duplicates) that did not survive verification.

## Re-review of `587c80c2d5a4`

- **Confirmed** — `IndexEntryReader.cpp:638` (async tasks capture raw `dest`/`&pm`/`&state`): both reviewers re-verified the orphan-task use-after-free window in `ThreadPool::Submit`; the finding still stands (details below).
- **Confirmed** — `IndexEntryWriterTest.cpp:1028` (encrypted path lacks tail-cache read-count coverage): all provider-read-count assertions still use plain entries only; the gap stands and is restated below as a follow-up.
- **Retracted** — `IndexEntryReader.cpp:126` (duplicated 256 KiB tail constant): our earlier comment was wrong. `ExactTailBoundaryReadCounts` pins expected reads to 1/1/2 at 256 KiB−1 / 256 KiB / 256 KiB+1, so any one-sided change to the production threshold fails the test loudly rather than drifting silently — the independently defined test constant acts as a contract pin, not a hazard.
- **Retracted** — `IndexEntryReader.h:254` (`small_entry_cache_` removal undisclosed): our earlier comment was wrong. An audit of every production `ReadEntry` caller shows each entry is read at most once per reader instance (the `__meta__` entry is read once in `Open` and served from `meta_json_` thereafter), so the cache was populated but never hit in production and its removal changes no production I/O; the removal was also explicitly requested by a reviewer and audited in-thread, so it was not a silent behavior change.
- **Confirmed** — `internal/core/src/storage/IndexEntryReader.cpp:638` (full-path duplicate of the first item): the same lifetime defect was re-verified, including on the encrypted (`:697-727`) and file-download (`:857-880`) paths.
- **Confirmed** — `internal/core/src/storage/IndexEntryWriterTest.cpp:1028` (duplicate of the second item): no encrypted test combines `ReadEntry` with `CountingInputStream` or asserts cache-hit/fallback read counts.
- **Retracted** — `internal/core/src/storage/IndexEntryReader.cpp:126` (duplicate of the third item): same reason — the fixed 1/1/2 boundary expectations fail on any one-sided threshold change instead of staying silently green.

## This round's findings

### Medium

- `internal/core/src/storage/IndexEntryReader.cpp:638` — **Async tasks capture raw pointers, and `ThreadPool::Submit`'s publication ordering can orphan an accepted task, leading to use-after-free.** The plain (`:638-649`), encrypted (`:697-727`), and file-download (`:857-880`) lambdas capture raw `dest`, `&pm`, `&state`, and file descriptors. `Submit` enqueues the task (`ThreadPool.h:163`) before constructing an expansion worker (`:179`), which can throw `std::system_error` before the future is returned (`:185`); the caller's `futures.push_back` then never runs, so `DrainFutures` (`IndexEntryReader.cpp:129-143`) cannot await the already-enqueued task. Stack unwinding destroys `result`/`state` or closes the descriptor while an idle worker may still dereference the dangling captures or `pwrite` through a closed/reused fd. This is pre-existing — this PR's by-value captures already fixed the stream/cipher lifetime dependencies — and is tracked in #51919 with a pool-layer fix in #51920, but it is present in this tree. Suggestion: provision the worker before publishing the task, or guarantee every accepted task returns an awaitable handle. *(raised by tinswzy, chyezh)*

### Low

- `internal/core/src/storage/IndexEntryWriterTest.cpp:1297` — **The encrypted path has no provider-read-count assertions; this is disclosed in the PR body and best handled as a follow-up.** All `CountingInputStream` usages cover plain entries only, while the encrypted suite (`IndexEntryEncryptedV3Test`, `:1297-1466`) writes files and asserts sizes but never combines an encrypted `ReadEntry` with read counting — even though encrypted reads go through the same `TailCachedInputStream` wrapper (`IndexEntryReader.cpp:697-712`), so a regression in ciphertext full-hit or partial-overlap fallback behavior would go undetected. The PR body explicitly discloses this gap and the rationale (proper coverage needs a real loadable cipher-plugin fixture rather than a test-only `Open` overload), so this is a declared limitation rather than a defect of this PR. Suggestion: in a follow-up, add encrypted full-hit and partial-overlap read-count cases built on a real cipher-plugin fixture. *(raised by tinswzy, chyezh)*

xaxys:
Based on the latest adversarial re-review, the risk assessment has converged. All remaining findings are known risks and have already been clarified above, and the review acknowledges their corresponding handling: the pre-existing lifetime hazard is tracked in #51919/#51920, and the encrypted-path coverage gap is disclosed and tracked in #51921 as a non-blocking follow-up. The other previously raised concerns have been retracted. No new merge-blocking issue remains, and this PR is ready to merge.

xaxys:
Resolved the current master conflict at `aefb0be5f7`.

Master now has a separate `InspectStreamLoadInfo()` path for scalar-index load resource estimation. For non-HYBRID V3 packed scalar indexes, and for the HYBRID fallback path, that inspection normally adds one 64 KiB tail Range GET before the later `IndexEntryReader::Open()` used by actual loading. The inspection reader is destroyed after returning sizing metadata, so its downloaded bytes cannot be reused by the reader-local tail cache in this PR; this change does not remove that additional estimation request.

The integration keeps the two paths deliberately separate:

- `InspectStreamLoadInfo()` retains the upstream 64 KiB inspection window and continues to skip an independent magic read because its caller has already selected the V3 path. The large-meta inspection test now directly verifies one tail `ReadAt` rather than allowing an extra 8-byte magic read.
- Normal `IndexEntryReader::Open()` still validates the magic and reads/retains up to 256 KiB for subsequent metadata and entry reuse.

This preserves the lightweight upstream resource-estimation behavior while applying the 256 KiB reuse optimization only to the reader that survives into actual index loading.

ThreadPool context: current master also merged an alternative fix through [#51403](https://github.com/milvus-io/milvus/pull/51403). It now keeps an already-published task queued and returns its future when dynamic worker expansion fails. #51920 uses the different reject-before-publication policy and is currently being held open while that policy choice is considered. The current production call graph has no same-pool nested synchronous submission; such submit-and-wait usage should remain disallowed. This tail-cache PR therefore relies on the current master behavior for ordinary worker-spawn failure and does not add local ThreadPool lifetime workarounds.

czs007:

## Summary

This review reports one low-severity test-coverage gap introduced by this PR, plus one real but pre-existing hazard tracked separately.

## Low

- **`internal/core/src/storage/IndexEntryWriterTest.cpp:1113` — the encrypted tail-cache hit path has no read-count coverage.** All nine new `IndexEntryReaderTailCacheTest` cases build their images with `BuildPackedFileImage`, which writes no `__edek__` and therefore exercises only the plaintext (`is_encrypted_ = false`) branch; they are the only tests that assert provider read counts. This PR also routes the encrypted read paths (`ReadEncryptedEntry`, `SubmitEntryDownloadTasks`, `SubmitEntryStreamDownloadTasks`) through `input_` → `TailCachedInputStream` (`IndexEntryReader.cpp:729-733,762-764`), and the coverage predicate in `TailCachedInputStream.h:70-85` is format-agnostic, so an encrypted small entry fully inside the retained tail should also incur zero provider reads — but nothing pins that. The existing `IndexEntryEncryptedV3Test` uses `CreateInputStream` for a write/read roundtrip and never counts reads, so a regression on the encrypted path would be silent. Suggestion: add one counting test that combines `CountingInputStream` with a loadable cipher-plugin fixture and an encrypted small entry, asserting zero provider reads on a full tail hit. *(raised by xiaocai2333)*

- **`internal/core/src/storage/IndexEntryReader.cpp:690` — `ThreadPool::Submit` can enqueue a task and then throw before returning its future, leaving an orphaned task holding raw references to `dest` / `&pm` / `&state` / `fd`.** Every `Submit` loop in this file (`683-712`, `740-787`, `1147-1166`, `1192-1224`, `1253-1285`) is already wrapped in `try` with a `catch` that runs `DrainFutures(futures, first_error)` before closing fds or rethrowing, and `futures` is reserved up front — so the caller-side unwind path is covered. The residual window is inside `Submit` itself: the task is live in the queue at `ThreadPool.h:163`, but the future is only returned at ~`ThreadPool.h:185`, after metric updates, locking, the dynamic `std::thread` construction at `ThreadPool.h:179`, and the map insert at `ThreadPool.h:181`. Failure scenario: under high-concurrency load near `RLIMIT_NPROC`, the `std::thread` construction throws `std::system_error` after the task is enqueued → the task is absent from `futures`, so `DrainFutures` cannot wait on it; `ReadEntriesToFiles` closes all fds and unwinds, and the orphaned task then writes `state.range_crcs[i]` into a destroyed `EntryDownloadState` (heap use-after-free) and `pwrite`s to a closed — possibly recycled — fd, producing random segfaults, heap corruption, or silent cross-file data corruption. This is unrelated to the PR: the `Submit` code path is untouched here, and the diff only replaces `this` in the capture lists with by-value `input`/`cipher_plugin`/`edek`/`ez_id` — `dest`, `&pm`, `&state`, and `fd = state.fd` are carried over verbatim, as is the `try`/`catch` + `DrainFutures` structure in every caller. The PR narrows this window rather than widening it. Reviewer notes indicate this is tracked in #51919/#51920; the durable fix is to give the task ownership of the lifetimes (shared_ptr for the state and result buffer) or to make enqueued tasks reclaimable when `Submit` throws. Recommend landing that separate fix on its own schedule rather than folding it into this PR. *(raised by xiaocai2333)*

PR #51701, opened 2026-07-22
fix: [cp2.6]avoid nil worker dereference in delegator ReleaseSegments

bigsheeper:

pr: #51474
issue: #51472

## Summary

Cherry-picked from master PR #51474 (open — preemptive backport). 2.6 has the same nil-worker dereference: on `GetWorker` error the old code fell through and called `worker.ReleaseSegments` on a nil worker. Fix wraps it in an `else` branch.

## Divergence from master PR (intentional, 2.6 adaptations)

1. **Logging**: 2.6 predates the mlog migration, so the release-failure log keeps the 2.6 signature `log.Warn("...", zap.Error(err))` instead of master's `log.Warn(ctx, "...", mlog.Err(err))`. Logic is identical.
2. **Test**: the master diff's hunk pulled in 6 suite tests; only `TestReleaseSegmentsWorkerNotAvailable` belongs to this PR — the other 5 (`TestReleaseGrowingSource*`) are unrelated master-only features not on 2.6. Only the relevant regression test was ported. Net diff shape matches master (+31/-6).

## Verification

- [x] Fix logic matches master (else-branch guards nil worker)
- [x] Go-level compile clean (`go vet`); full test run needs cgo C++ core (validated by CI)
- [x] No conflict markers

### DIFF
--- a/internal/querynodev2/delegator/delegator_data.go
+++ b/internal/querynodev2/delegator/delegator_data.go
@@ -1005,12 +1005,12 @@ func (sd *shardDelegator) ReleaseSegments(ctx context.Context, req *querypb.Rele
 		if err != nil {
 			log.Warn("delegator failed to find worker", zap.Error(err))
 			releaseErr = err
-		}
-		req.Base.TargetID = targetNodeID
-		err = worker.ReleaseSegments(ctx, req)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


czs007 (approved):
/lgtm

Issue #51686, opened 2026-07-22
[Bug]: e2e test_struct_array_compaction_preserves_null_empty_offsets_and_indexes deterministically fails on distributed-pulsar (compact() returns -1), blocking ci-v2/e2e-default


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

CI e2e, deployment mode: **distributed-pulsar** (milvus-e2e-pipeline-gcp)

### Current Behavior

`milvus_client/test_milvus_client_struct_array_nullable.py::TestMilvusClientStructArraySchemaEvolution::test_struct_array_compaction_preserves_null_empty_offsets_and_indexes` fails deterministically on the **distributed-pulsar** e2e deployment, blocking `ci-v2/e2e-default` for any PR whose checkout includes the test.

Failure point (3 of 3 runs, two unrelated PRs, independent clusters):

```
assert compact_id > 0   # test line ~1252
E   assert -1 > 0
```

i.e. manual `compact()` on the struct-array collection returns **-1** on distributed-pulsar. (One run additionally surfaced an `insert_count == len(batch_ids)` mismatch during setup on an unhealthy cluster.)

Evidence:

| dispatcher run | PR | sub-job | mode | result |
|---|---|---|---|---|
| [15277](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-dispatcher/15277/) | #51678 | [44066](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-gcp/44066/) | distributed-pulsar | FAILED (this test) |
| [15285](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-dispatcher/15285/) | #51678 (rerun) | [44078](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-gcp/44078/) | distributed-pulsar | FAILED (this test, `assert -1 > 0`) |
[...]

PR #51470, opened 2026-07-16
enhance: [cp 3.0] add FileResource APIs to Go SDK and REST v2

zhuwenxing:
issue: #51241
pr: #51339
pr: #51703

## Summary

Cherry-pick of master PR #51339 to the `3.0` branch.

- Add `AddFileResource`, `ListFileResources`, and `RemoveFileResource` to the Go SDK.
- Add REST v2 FileResource endpoints under `/v2/vectordb/file_resources`.
- Keep add/remove under the cluster-level collection DDL limiter while leaving the read-only list operation zero-cost.
- Reject empty FileResource names at the server boundary for consistent SDK and REST behavior.
- Include the Go SDK, REST handler, E2E, and RBAC coverage from the master PR.

## Branch notes

This is a clean cherry-pick of all three commits from #51339. No 3.0-specific conflict resolution or code changes were required. `git range-diff` and stable patch IDs match all three source commits.

## Validation

- 3.0 remote Go UT: `client/milvusclient` (`TestFileResource`)
- 3.0 remote Go UT: `internal/distributed/proxy/httpserver` (`TestFileResourceHandlerV2`)
- 3.0 remote Go UT: `internal/proxy` (`TestRateLimitInterceptor|TestGetInfo`)
- 3.0 remote Go UT: `internal/rootcoord` (`TestRootCoord_(Add|Remove)FileResource`)
- `go test -tags dynamic,test -gcflags='all=-N -l' -count=1 ./util/merr/...` from `pkg/`
- Ruff check and format check for the changed REST client files
- `git diff --check upstream/3.0...HEAD`
- Master PR CI: build/UT, default E2E, Go SDK, and Go SDK ARM checks passed

PR #51703, opened 2026-07-22
test: [3.0] isolate struct array manual compaction

zhuwenxing:
pr: #51689
issue: #51686

### What changed

- Disable collection auto-compaction when creating the collection in `test_struct_array_compaction_preserves_null_empty_offsets_and_indexes`.
- Keep explicit manual compaction as the only compaction trigger.

### Why

Automatic sort/mix compaction can claim all flushed segments before explicit manual compaction on distributed Pulsar. DataCoord then creates no task and `compact()` returns `-1`, causing flaky `ci-v2/e2e-amd` failures.

This is the 3.0 backport of #51689.

### Verification

- `ruff check tests/python_client/milvus_client/test_milvus_client_struct_array_nullable.py`
- `git diff --check`


zhuwenxing:
covered by https://github.com/milvus-io/milvus/pull/51470

PR #51705, opened 2026-07-22
fix: [3.0] fail stale schema reopen instead of silent success (#51644)

congqixia:
Cherry-pick from master
pr: #51644
Related to #51068

Stale schema reopen used to return success from segcore, which let the Go caller update LocalSegment loadInfo even though the core state was not reopened. Return a retriable schema-version error instead so QueryCoord can retry with a fresh schema.

Map the new segcore code to ErrCollectionSchemaVersionNotReady and add tests for both the C++ stale reopen path and the Go-side loadInfo guard.

---------


liliu-z (commented):
/lgtm
/approve

Issue #51707, opened 2026-07-22
[Enhancement]: External-collection index build reads raw data serially (single S3 connection for the whole download phase)


xiaocai2333:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Parallelize the raw-data read path for external-collection index build, and stream the downloaded column to local disk instead of materializing it in memory.

Profiling a disk-index build worker (AWS, laion 1B external parquet, 768-dim fp32, 8C worker) shows the download phase takes ~63% of task wall time while CPU sits near idle and exactly **one** S3 connection is ever open. Tracing the read path end to end found four independent causes, each sufficient to serialize it:

1. `common.arrow.reader.holeSizeLimitBytes` / `rangeSizeLimitBytes` are applied only to `LoonFFIPropertiesSingleton`'s cached map. Index build creates its milvus-storage properties per task via `MakeInternalPropertiesFromStorageConfig`, bypassing the cache, so prebuffer configuration never reaches those reads.
2. milvus-storage's global reader thread pool is never initialized, so `ThreadPoolHolder::GetParallelism()` returns 1 and chunks (row groups) are read strictly one at a time.
3. The per-round prefetch window (`reader.record_batch_max_size`, default 32MB) is smaller than one 64MB-class parquet row group, so each round admits exactly one row group — i.e. one S3 range read in flight regardless of the arrow IO pool size. With column projection to a single vector column, a row group is a single ~62MB column chunk, and arrow's `CoalesceReadRanges` merges but never splits ranges — a single GET on a single connection, ~40MB/s.
4. Batch decode (external-type normalization + `FieldData` materialization) runs single-threaded on the consumer loop, strictly alternating with fetch rounds, and `cache_raw_data_to_disk_storage_v2` materializes the entire raw column in memory before writing it out.

### Why is this needed?

Measured on the same instance/collection (per-million-rows normalized, same environment A/B): the fix lifts a 4.68M-row disk-index task from 283s to ~151s (~2x software gain); combined with correct local-disk placement the task reaches 71-78s (~19-21 s per million rows). S3 concurrency goes from 1 to 19+ connections and the download-phase memory footprint drops from the full raw column (>11GB) to a bounded prefetch + decode window.

### Anything else?

Fix PRs: 3.0 branch and master (linked below). The same code exists unchanged on master.

PR #51708, opened 2026-07-22
enhance: parallelize external-collection index-build raw data reads

xiaocai2333:
issue: #51707

Master port of #51651 (3.0 branch). The touched read-path code is identical on master. See #51651 for the full measurement history.

## Problem

For external-collection index build, downloading the raw vector data dominates wall time: ~63% of task time with CPU near idle and exactly **one** S3 connection open. Four independent causes can serialize the read path:

1. Arrow reader prebuffer config (`holeSizeLimitBytes`/`rangeSizeLimitBytes`) only reaches the cached singleton properties; index build creates per-task properties and bypasses that map.
2. The milvus-storage global reader thread pool is never initialized (`GetParallelism() == 1`).
3. The per-round prefetch window (32MB default) is smaller than one 64MB-class row group.
4. Batch decode runs single-threaded and alternates with fetch; the raw-data spill materializes the whole raw column before writing.

## Changes

- Stamp prebuffer limits into every freshly built properties map as one coherent snapshot.
- Add two opt-in knobs with C API, DataNode wiring and hot-reload watchers: `common.storage.readerThreadPoolSize` and `common.storage.indexBuildReadWindowBytes`.
- Stream manifest batches through `IterateFieldDataFromManifest`: `ReadNext` remains single-threaded, per-batch decode runs on the LOW background pool under byte/count backpressure, and delivery remains ordered on the caller thread.
- Account in-flight bytes with Arrow referenced-buffer ranges so slices from one row group are charged only for the ranges they reference.
- Stream the raw-data spill so download-phase retention is bounded by the prefetch and decode windows instead of the full raw column.
- Propagate `minio.maxConnections` through DataCoord and all DataNode storage-config conversion paths into the C++ clients. The default remains 100.
- Emit one `[ManifestReadStats]` line per streamed read with fetch, decode-wait and consume phase timing.

## Measured (historical MIDDLE-pool iteration; not the current head)

The earlier 3.0 A/B run used the previous MIDDLE-pool decode implementation: laion 1B external parquet, 768-dim fp32, 8C/32Gi worker, 283s average to 151s average per task, with observed S3 connections increasing from 1 to 19+.

The final head uses the LOW pool, which changes worker count and the derived in-flight batch limit. That variant has not yet been re-benchmarked in the same environment, so the historical numbers do not validate a current-head ~2x claim. A same-environment LOW-pool A/B benchmark is follow-up work.

### DIFF
--- a/internal/core/src/storage/RecordBatchSize.h
+++ b/internal/core/src/storage/RecordBatchSize.h
@@ -23,16 +23,26 @@ namespace milvus::storage {
 
 // Returns the bytes referenced by this record-batch slice. Unlike
 // TotalBufferSize, this accounts for array offsets and does not charge every
-// slice for its entire shared backing buffer. Falls back to the row count if
-// Arrow cannot compute the referenced ranges.
+// slice for its entire shared backing buffer.
+//
+// ReferencedBufferSize cannot walk every layout. On the pinned arrow 17 its
+// byte-range visitor has no overload for STRING_VIEW / BINARY_VIEW /
+// LIST_VIEW and falls through to a TypeError catch-all, which
+// ReferencedBufferSize(RecordBatch) propagates for the whole batch -- and
+// external (vortex schemaless) readers do produce those layouts, on the
+// pre-normalization batch this is called with. Fall back to the plain buffer
+// sum, which works for any type. It over-charges slices that share a backing
+// buffer, but for a memory cap over-charging is the safe direction: the row
+// count would charge a 64MB batch as 8192 bytes and silently disable byte
+// backpressure altogether, leaving only the batch-count limit.
 inline int64_t
 EstimateRecordBatchBytes(const arrow::RecordBatch& batch) {
     auto size_result = arrow::util::ReferencedBufferSize(batch);
-    if (!size_result.ok()) {
-        return batch.num_rows();
+    if (size_result.ok() && size_result.ValueOrDie() > 0) {
+        return size_result.ValueOrDie();
     }
-    auto referenced_bytes = size_result.ValueOrDie();
-    return referenced_bytes > 0 ? referenced_bytes : batch.num_rows();
+    auto total_bytes = arrow::util::TotalBufferSize(batch);
+    return total_bytes > 0 ? total_bytes : batch.num_rows();
 }
 
 }  // namespace milvus::storage
--- a/internal/core/src/storage/RecordBatchSize.h
+++ b/internal/core/src/storage/RecordBatchSize.h
@@ -1,0 +1,38 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "arrow/record_batch.h"
+#include "arrow/util/byte_size.h"
+
+namespace milvus::storage {
+
+// Returns the bytes referenced by this record-batch slice. Unlike
+// TotalBufferSize, this accounts for array offsets and does not charge every
+// slice for its entire shared backing buffer. Falls back to the row count if
+// Arrow cannot compute the referenced ranges.
+inline int64_t
+EstimateRecordBatchBytes(const arrow::RecordBatch& batch) {
+    auto size_result = arrow::util::ReferencedBufferSize(batch);
+    if (!size_result.ok()) {
+        return batch.num_rows();
+    }
+    auto referenced_bytes = size_result.ValueOrDie();
+    return referenced_bytes > 0 ? referenced_bytes : batch.num_rows();
+}
+
+}  // namespace milvus::storage
--- a/internal/core/src/storage/Util.cpp
+++ b/internal/core/src/storage/Util.cpp
@@ -24,7 +24,6 @@
 
 #include "arrow/array/builder_binary.h"
 #include "arrow/array/builder_nested.h"
-#include "arrow/util/byte_size.h"
 #include "arrow/array/builder_primitive.h"
 #include "arrow/array/concatenate.h"
 #include "arrow/buffer_builder.h"
@@ -56,6 +55,7 @@
 #include "storage/LocalChunkManager.h"
 #include "storage/MemFileManagerImpl.h"
 #include "storage/minio/MinioChunkManager.h"
+#include "storage/RecordBatchSize.h"
 #include "storage/Types.h"
 #include "storage/Util.h"
 #include "common/Common.h"
@@ -1761,15 +1761,10 @@ IterateFieldDataFromManifest(
             continue;
         }
 
-        // Estimate the decoded footprint of this batch for the byte budget.
-        // The arrow buffers backing the batch are the best available proxy
-        // for the FieldData it will materialize into; fall back to the row
-        // count when the size cannot be computed.
-        int64_t batch_bytes = 0;
-        {
-            auto size_result = arrow::util::TotalBufferSize(*batch);
-            batch_bytes = size_result > 0 ? size_result : num_rows;
-        }
+        // Estimate the decoded footprint of this slice for the byte budget.
+        // A record batch may share a large row-group backing buffer with
+        // adjacent slices, so count only the ranges this slice references.
+        auto batch_bytes = EstimateRecordBatchBytes(*batch);
 
         auto decode_future = pool.Submit([batch,
                                           column_name,
--- a/internal/core/src/storage/Util.cpp
+++ b/internal/core/src/storage/Util.cpp
@@ -14,12 +14,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <chrono>
+#include <deque>
 #include <filesystem>
+#include <functional>
+#include <future>
 #include <memory>
 #include "common/FastMem.h"
 
 #include "arrow/array/builder_binary.h"
 #include "arrow/array/builder_nested.h"
+#include "arrow/util/byte_size.h"
 #include "arrow/array/builder_primitive.h"
 #include "arrow/array/concatenate.h"
 #include "arrow/buffer_builder.h"
@@ -1550,15 +1555,16 @@ GetFieldDatasFromStorageV2(std::vector<std::vector<std::string>>& remote_files,
     return field_data_list;
 }
 
-std::vector<FieldDataPtr>
-GetFieldDatasFromManifest(
+void
+IterateFieldDataFromManifest(
     const std::string& manifest_path,
     const std::shared_ptr<milvus_storage::api::Properties>& loon_ffi_properties,
     const FieldDataMeta& field_meta,
     std::optional<DataType> data_type,
     int64_t dim,
     std::optional<DataType> element_type,
-    std::optional<StorageColumnMapping> storage_column_mapping) {
+    std::optional<StorageColumnMapping> storage_column_mapping,
+    const std::function<void(FieldDataPtr)>& consumer) {
     auto loon_manifest = GetLoonManifest(manifest_path, loon_ffi_properties);
     auto column_groups = std::make_shared<milvus_storage::api::ColumnGroups>(
         loon_manifest->columnGroups());
@@ -1595,19 +1601,20 @@ GetFieldDatasFromManifest(
     };
     bool field_exists = column_exists(column_name);
     if (!field_exists) {
-        return {};
+        return;
     }
 
     std::vector<std::string> needed_columns = {column_name};
 
     bool nullable = field_meta.field_schema.nullable();
-    std::optional<FieldMeta> normalize_field_meta;
+    std::shared_ptr<const FieldMeta> normalize_field_meta;
     if (is_external) {
         auto schema = field_meta.field_schema;
         if (schema.fieldid() == 0) {
             schema.set_fieldid(field_meta.field_id);
         }
-        normalize_field_meta.emplace(FieldMeta::ParseFrom(schema));
+        normalize_field_meta =
+            std::make_shared<const FieldMeta>(FieldMeta::ParseFrom(schema));
     }
 
     // External tables: schemaless reader - let the reader derive types from
@@ -1655,11 +1662,78 @@ GetFieldDatasFromManifest(
 
     auto record_batch_reader = reader_result.ValueOrDie();
 
-    std::vector<FieldDataPtr> field_datas;
+    // Decode batches on the MIDDLE thread pool while this thread keeps
+    // draining the record batch reader. ReadNext must stay single-threaded
+    // (the reader is not thread-safe), but everything after it — external
+    // normalization plus FieldData materialization — is pure per-batch work
+    // and is the dominant cost. Offloading it keeps the reader thread free
+    // to trigger the next prefetch round, so network fetch and decode
+    // overlap instead of strictly alternating. Results are delivered to
+    // `consumer` on this thread in batch order; the bounded in-flight
+    // window provides backpressure so decoded-but-undelivered batches
+    // cannot pile up without limit.
+    //
+    // The window is bounded by bytes, not by batch count: a count-based cap
+    // scales with the pool size and the per-batch size, so on a large pool
+    // with 64MB batches it would admit gigabytes of decoded data per build
+    // (and several builds may run concurrently). The byte budget below caps
+    // the decoded-but-undelivered bytes regardless of batch size and thread
+    // count; at least two batches are always admitted so decode can still
+    // overlap with fetch when a single batch exceeds the budget.
+    auto& pool = ThreadPools::GetThreadPool(ThreadPoolPriority::MIDDLE);
+    constexpr int64_t kMaxInflightBytes = 512LL << 20;
+    const size_t max_inflight_batches =
+        std::max<size_t>(2, pool.GetMaxThreadNum() * 2);
+    std::deque<std::pair<std::future<FieldDataPtr>, int64_t>> pending;
+    int64_t pending_bytes = 0;
+
+    // Phase accounting for the streaming loop, reported once at the end.
+    // fetch = ReadNext (network/prefetch wait), decode_wait = blocking on
+    // the decode future, consume = the consumer callback (for index build:
+    // the local disk write). The three phases run on this thread and are
+    // mutually exclusive, so comparing their totals against the wall time
+    // shows where the pipeline actually spends its time — in particular
+    // whether a blocking consumer is starving ReadNext (fetch_max grows)
+    // or the writes themselves are slow (consume dominates).
+    int64_t fetch_ns = 0, fetch_max_ns = 0;
+    int64_t decode_wait_ns = 0;
+    int64_t consume_ns = 0, consume_max_ns = 0;
+    int64_t total_batch_bytes = 0;
+    size_t batch_count = 0;
+    auto wall_start = std::chrono::steady_clock::now();
+    auto now_ns = []() {
+        return std::chrono::duration_cast<std::chrono::nanoseconds>(
+                   std::chrono::steady_clock::now().time_since_epoch())
+            .count();
+    };
+
+    auto deliver_front = [&]() {
+        auto t0 = now_ns();
+        auto field_data = pending.front().first.get();
+        auto t1 = now_ns();
+        pending_bytes -= pending.front().second;
+        pending.pop_front();
+        consumer(std::move(field_data));
+        auto t2 = now_ns();
+        decode_wait_ns += t1 - t0;
+        consume_ns += t2 - t1;
+        consume_max_ns = std::max(consume_max_ns, t2 - t1);
+    };
+
+    auto data_type_v = data_type.value();
+    auto element_type_v = element_type.value();
     while (true) {
         std::shared_ptr<arrow::RecordBatch> batch;
+        auto fetch_start = now_ns();
         auto status = record_batch_reader->ReadNext(&batch);
+        auto fetch_elapsed = now_ns() - fetch_start;
+        fetch_ns += fetch_elapsed;
+        fetch_max_ns = std::max(fetch_max_ns, fetch_elapsed);
         if (!status.ok()) {
+            // Drain workers before throwing so no task outlives this scope.
+            for (auto& f : pending) {
+                f.first.wait();
+            }
             auto error = milvus_storage::ToSegcoreError(status);
             ThrowInfo(error.get_error_code(),
                       "Failed to read record batch: {}",
@@ -1674,23 +1748,108 @@ GetFieldDatasFromManifest(
             continue;
         }
 
-        auto raw_column = batch->GetColumnByName(column_name);
-        if (is_external) {
-            raw_column =
-                NormalizeExternalArrowByType(raw_column,
-                                             data_type.value(),
-                                             dim,
-                                             nullable,
-                                             element_type.value(),
-                                             normalize_field_meta.value());
+        // Estimate the decoded footprint of this batch for the byte budget.
+        // The arrow buffers backing the batch are the best available proxy
+        // for the FieldData it will materialize into; fall back to the row
+        // count when the size cannot be computed.
+        int64_t batch_bytes = 0;
+        {
+            auto size_result = arrow::util::TotalBufferSize(*batch);
+            batch_bytes = size_result > 0 ? size_result : num_rows;
         }
-        auto chunked_array = std::make_shared<arrow::ChunkedArray>(raw_column);
-        auto field_data = CreateFieldData(
-            data_type.value(), element_type.value(), nullable, dim, num_rows);
-        field_data->FillFieldData(chunked_array);
-        field_datas.push_back(field_data);
+
+        auto decode_future = pool.Submit([batch,
+                                          column_name,
+                                          is_external,
+                                          data_type_v,
+                                          element_type_v,
+                                          dim,
+                                          nullable,
+                                          normalize_field_meta,
+                                          num_rows]() -> FieldDataPtr {
+            auto raw_column = batch->GetColumnByName(column_name);
+            if (is_external) {
+                raw_column =
+                    NormalizeExternalArrowByType(raw_column,
+                                                 data_type_v,
+                                                 dim,
+                                                 nullable,
+                                                 element_type_v,
+                                                 *normalize_field_meta);
+            }
+            auto chunked_array =
+                std::make_shared<arrow::ChunkedArray>(raw_column);
+            auto field_data = CreateFieldData(
+                data_type_v, element_type_v, nullable, dim, num_rows);
+            field_data->FillFieldData(chunked_array);
+            return field_data;
+        });
+        pending.emplace_back(std::move(decode_future), batch_bytes);
+        pending_bytes += batch_bytes;
+        total_batch_bytes += batch_bytes;
+        ++batch_count;
+
+        // Backpressure: block on the oldest batch once the window is full
+        // by bytes or by count. Always keep at least one in flight so a
+        // single oversized batch cannot deadlock the loop.
+        while (pending.size() > 1 && (pending_bytes > kMaxInflightBytes ||
+                                      pending.size() >= max_inflight_batches)) {
+            deliver_front();
+        }
+        // Opportunistically deliver whatever is already done, keeping
+        // consumer-side work (e.g. disk writes) interleaved with fetching.
+        while (!pending.empty() &&
+               pending.front().first.wait_for(std::chrono::seconds(0)) ==
+                   std::future_status::ready) {
+            deliver_front();
+        }
     }
 
+    while (!pending.empty()) {
+        deliver_front();
+    }
+
+    if (batch_count > 0) {
+        auto wall_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                           std::chrono::steady_clock::now() - wall_start)
+                           .count();
+        LOG_INFO(
+            "[ManifestReadStats] column={} batches={} MB={} wall_ms={} "
+            "fetch_ms={} fetch_max_ms={} decode_wait_ms={} consume_ms={} "
+            "consume_max_ms={}",
+            column_name,
+            batch_count,
+            total_batch_bytes >> 20,
+            wall_ms,
+            fetch_ns / 1'000'000,
+            fetch_max_ns / 1'000'000,
+            decode_wait_ns / 1'000'000,
+            consume_ns / 1'000'000,
+            consume_max_ns / 1'000'000);
+    }
+}
+
+std::vector<FieldDataPtr>
+GetFieldDatasFromManifest(
+    const std::string& manifest_path,
+    const std::shared_ptr<milvus_storage::api::Properties>& loon_ffi_properties,
+    const FieldDataMeta& field_meta,
+    std::optional<DataType> data_type,
+    int64_t dim,
+    std::optional<DataType> element_type,
+    std::optional<StorageColumnMapping> storage_column_mapping) {
+    std::vector<FieldDataPtr> field_datas;
+    IterateFieldDataFromManifest(
+        manifest_path,
+        loon_ffi_properties,
+        field_meta,
+        data_type,
+        dim,
+        element_type,
+        std::move(storage_column_mapping),
+        [&](FieldDataPtr field_data) {
+            field_datas.push_back(std::move(field_data));
+        });
     return field_datas;
 }
 
--- a/internal/core/src/storage/Util.cpp
+++ b/internal/core/src/storage/Util.cpp
@@ -1676,10 +1676,22 @@ IterateFieldDataFromManifest(
     // The window is bounded by bytes, not by batch count: a count-based cap
     // scales with the pool size and the per-batch size, so on a large pool
     // with 64MB batches it would admit gigabytes of decoded data per build
-    // (and several builds may run concurrently). The byte budget below caps
-    // the decoded-but-undelivered bytes regardless of batch size and thread
-    // count; at least two batches are always admitted so decode can still
-    // overlap with fetch when a single batch exceeds the budget.
+    // (and several builds may run concurrently). The byte budget below bounds
+    // the window regardless of batch size and thread count; at least two
+    // batches are always admitted so decode can still overlap with fetch when
+    // a single batch exceeds the budget.
+    //
+    // What is charged is each batch's *input* arrow bytes
+    // (EstimateRecordBatchBytes on the source slice), not the decoded
+    // FieldData that actually accumulates in `pending`. For internal native
+    // types the two are close, but on the external path
+    // NormalizeExternalArrowByType can inflate the decoded footprint by a
+    // type-dependent factor, so peak retained bytes can exceed
+    // kMaxInflightBytes by roughly that factor. Charging post-decode would
+    // need the decode to finish before admission, which is exactly the
+    // serialization this pipeline exists to avoid; the accounting is
+    // symmetric (deliver_front discharges what emplace_back charged), so this
+    // is a looser bound, not a leak.
     //
     // The decode tasks go to the LOW pool, not MIDDLE. Every production
     // caller of this function is an index build (DiskFileManagerImpl /
@@ -1779,7 +1791,8 @@ IterateFieldDataFromManifest(
             continue;
         }
 
-        // Estimate the decoded footprint of this slice for the byte budget.
+        // Charge this slice's arrow input bytes against the byte budget (see
+        // the note on the budget above for why input, not decoded, bytes).
         // A record batch may share a large row-group backing buffer with
         // adjacent slices, so count only the ranges this slice references.
         auto batch_bytes = EstimateRecordBatchBytes(*batch);
--- a/internal/core/src/storage/Util.cpp
+++ b/internal/core/src/storage/Util.cpp
@@ -1564,7 +1564,11 @@ IterateFieldDataFromManifest(
     int64_t dim,
     std::optional<DataType> element_type,
     std::optional<StorageColumnMapping> storage_column_mapping,
-    const std::function<void(FieldDataPtr)>& consumer) {
+    const std::function<void(FieldDataPtr)>& consumer,
+    int64_t max_inflight_bytes) {
+    AssertInfo(max_inflight_bytes > 0,
+               "max_inflight_bytes must be positive, got {}",
+               max_inflight_bytes);
     auto loon_manifest = GetLoonManifest(manifest_path, loon_ffi_properties);
     auto column_groups = std::make_shared<milvus_storage::api::ColumnGroups>(
         loon_manifest->columnGroups());
@@ -1687,11 +1691,22 @@ IterateFieldDataFromManifest(
     // types the two are close, but on the external path
     // NormalizeExternalArrowByType can inflate the decoded footprint by a
     // type-dependent factor, so peak retained bytes can exceed
-    // kMaxInflightBytes by roughly that factor. Charging post-decode would
+    // max_inflight_bytes by roughly that factor. Charging post-decode would
     // need the decode to finish before admission, which is exactly the
     // serialization this pipeline exists to avoid; the accounting is
     // symmetric (deliver_front discharges what emplace_back charged), so this
     // is a looser bound, not a leak.
+    //
+    // The budget is a parameter because what it buys depends on the caller.
+    // For a streaming consumer (DiskFileManagerImpl spilling to local disk)
+    // the window *replaces* full-column retention, so a large one is a net
+    // reduction. For an accumulating caller — GetFieldDatasFromManifest, and
+    // through it every storage-v3 index build that goes via
+    // MemFileManagerImpl — the whole column is retained regardless, so the
+    // window only adds the source arrow batches that in-flight futures keep
+    // alive: pure extra peak RSS, on top of a column the Go-side build memory
+    // estimate already sized without it. Those callers pass
+    // kAccumulatingInflightBytes.
     //
     // The decode tasks go to the LOW pool, not MIDDLE. Every production
     // caller of this function is an index build (DiskFileManagerImpl /
@@ -1706,7 +1721,6 @@ IterateFieldDataFromManifest(
     // so this function must not be called from a LOW-pool thread. Index
     // build tasks enter segcore from Go, so no caller does today.
     auto& pool = ThreadPools::GetThreadPool(ThreadPoolPriority::LOW);
-    constexpr int64_t kMaxInflightBytes = 512LL << 20;
     const size_t max_inflight_batches =
         std::max<size_t>(2, pool.GetMaxThreadNum() * 2);
     std::deque<std::pair<std::future<FieldDataPtr>, int64_t>> pending;
@@ -1831,7 +1845,7 @@ IterateFieldDataFromManifest(
         // Backpressure: block on the oldest batch once the window is full
         // by bytes or by count. Always keep at least one in flight so a
         // single oversized batch cannot deadlock the loop.
-        while (pending.size() > 1 && (pending_bytes > kMaxInflightBytes ||
+        while (pending.size() > 1 && (pending_bytes > max_inflight_bytes ||
                                       pending.size() >= max_inflight_batches)) {
             deliver_front();
         }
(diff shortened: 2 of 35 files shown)


xiaocai2333:
Review summary for 86e506d978:

The overall direction and benchmark results look good, and I did not find any P0/P1 issue. I think the following P2 issues should be addressed before merge:

1. Avoid a false warning when the reader pool is disabled. ThreadPoolHolder::GetParallelism() returns 1 when the singleton pool does not exist. Therefore, with readerThreadPoolSize=0, hot-reloading only indexBuildReadWindowBytes produces requested=0/effective=1 and incorrectly warns that the pool could not be destroyed and a restart is required. Please warn only when requested is 0 and the effective parallelism is greater than 1, or expose the singleton existence explicitly.

2. Correct the hot-reload semantics documented for active readers. The new parameter documentation says readers latch parallelism and pool changes only affect tasks started afterwards. At milvus-storage e658197, a reader does latch parallelism_ when opened, but get_chunks() obtains the current singleton pool and splits work using its current numThreads(). Thus a non-zero resize such as 16 to 8 also changes subsequent rounds of already-open readers. Please either update the documentation/logging to describe the actual behavior or change the implementation to provide the promised isolation.

3. Validate the pool size before narrowing it to int32. InitLoonReaderConfig currently calls GetAsInt32(), which directly converts the underlying int64. Out-of-range positive values can wrap; for example, 4294967296 becomes 0 and is silently treated as disabled instead of being rejected. Please read it as int64, validate the range (and preferably a reasonable operational upper bound), then cast.

Verification performed:
- Existing PR CI is green for build/UT, 8240 C++ tests, E2E, and Go SDK.
- The focused paramtable test passed locally with the required dynamic,test tags and disabled optimizations/inlining.
- git diff --check passed.

The PR is currently conflicting with the latest master and will also need a rebase.

czs007:
Adversarial review of this PR surfaced four issues — one medium-severity thread-pool contention concern and three low-severity config-wiring/observability problems — none of which are blocking.

## Medium

**`internal/core/src/storage/Util.cpp:1679` — manifest decode is submitted unconditionally to the shared MIDDLE thread pool, contending with query reduce**

`IterateFieldDataFromManifest` changes manifest decoding from inline-on-the-calling-thread to `pool.Submit(...)` on `ThreadPoolPriority::MIDDLE`, with an in-flight cap of `max(2, GetMaxThreadNum() * 2)`. This path is not gated by either of the two new knobs (`common.storage.readerThreadPoolSize` / `common.storage.indexBuildReadWindowBytes`, both defaulting to 0) — `DiskFileManagerImpl.cpp:1020` and the `GetFieldDatasFromManifest` wrapper always take it. In standalone, querynode and datanode share the same `ThreadPools` singleton, and `ReduceHelper::FillPrimaryKey` (`Reduce.cpp:209`) plus the global refine step (`Reduce.cpp:457`) submit to MIDDLE and then block on `future.get()`.

Failure: in a standalone deployment with CPU ≥ 4 (MIDDLE pool saturated at the default 16-thread cap), an index build on an external collection running alongside multi-segment search traffic → each build task queues up to 32 CPU-bound decode tasks (multiplied by concurrent builds) ahead of the reduce tasks, so search P99 stays elevated for the whole build download phase until the decode queue drains.

Suggestion: gate this path behind an admission check, or route manifest decoding to a dedicated pool rather than the shared MIDDLE pool. Note this is a widening of an existing pattern rather than a brand-new one — the storage-v2 path at `Util.cpp:1471` already submits reads to MIDDLE — so at minimum this deserves an explicit note in the config docs. *(raised by chyezh, tinswzy)*

## Low

**`internal/util/initcore/init_core.go:509` — the global thread pool is mutated before the read window is validated, so a failed reconfigure is partially applied**

`InitLoonReaderConfig` calls `InitLoonReaderThreadPool` first (which resizes the global folly pool on success and cannot be undone — `storage_c.cpp:220-223` returns early for `num_threads == 0`, and the header notes the pool cannot be destroyed), then calls `InitIndexBuildReadWindow`, which rejects windows above 4 GiB with `ConfigInvalid` (`storage_c.cpp:245-249`).

Failure: pushing `common.storage.readerThreadPoolSize=32` and `common.storage.indexBuildReadWindowBytes=8589934592` (8 GiB) together to a running DataNode → the watcher resizes the global reader pool to 32 (not revertible via config; requires a restart), then fails validation and logs only `"failed to reconfigure loon reader params"` (`util.go:203-206`), leaving operators believing nothing changed while all subsequent loon reads run at concurrency 32.

Suggestion: validate both values up front and only mutate after both pass. *(raised by chyezh, tinswzy)*

**`internal/util/initcore/util.go:215` — the "thread pool not fully applied" warning fires unconditionally under the default configuration**

The hot-reload handler compares `requested != effective`, but `EffectiveLoonReaderThreadPoolSize()` bottoms out in `ThreadPoolHolder::GetParallelism()`, which returns `1` — not `0` — when no pool has been created (`thread_pool.h:59-64`). Since `common.storage.readerThreadPoolSize` defaults to `0` and both keys share one handler (`util.go:222-226`), the comparison is always unequal.

Failure: a DataNode leaves `readerThreadPoolSize` at its default `0` (no pool ever created) and an operator hot-updates only `common.storage.indexBuildReadWindowBytes=536870912` → the change applies correctly, yet the log immediately emits `"loon reader thread pool size not fully applied; the pool cannot be destroyed at runtime, restart to disable it"` (requested=0 / effective=1), sending the operator to restart the DataNode chasing a residual thread pool that does not exist.

Suggestion: guard with `requested > 0 && requested != effective`, or treat `0` and "no pool" as equivalent when computing the effective size. *(surfaced during verification)*

**`internal/datanode/index/init_segcore.go:108` — `common.storage.readerThreadPoolSize` is wired only on DataNode and silently has no effect on QueryNode**

`InitLoonReaderConfig` and `RegisterLoonReaderConfigWatchers` are called only from DataNode's segcore init (these are their only call sites in the repo), but `ThreadPoolHolder` is a process-wide singleton in milvus-storage, and QueryNode's loon readers (`ChunkedSegmentSealedImpl`, `SegmentGrowingImpl`) latch `ThreadPoolHolder::GetParallelism()` at `Reader::create` time. The key is named `common.*` and its doc string ("Size of milvus-storage's global reader thread pool") carries no component scope, unlike the adjacent `indexBuildReadWindowBytes`, which explicitly states "Only affects index build; query-node loads keep the default window."

Failure: in a distributed deployment an operator sets `common.storage.readerThreadPoolSize=16` globally, expecting chunk-level concurrency when QueryNode loads storage-v3 segments → QueryNode never calls `InitLoonReaderThreadPool`, its reader parallelism stays pinned at 1, and nothing is logged; the discrepancy is only discoverable by comparing DataNode and QueryNode behavior.

Suggestion: either wire the same init/watcher path into QueryNode's segcore initialization, or amend the doc string to state that the key applies to DataNode only. *(surfaced during verification)*

xiaocai2333:
All review comments are addressed in `5a5c1dda06` (rebased onto current master, so the conflict noted earlier is gone). Point-by-point:

## czs007 — Medium: manifest decode contends with query reduce on the shared MIDDLE pool

Fixed by routing per-batch decode to the **LOW** segcore pool (`Util.cpp`). Every production caller of `IterateFieldDataFromManifest` is an index build (`DiskFileManagerImpl` raw-data spill + opt-field IVF, `MemFileManagerImpl` build paths — verified by grep over `internal/core/src`), so nothing on the query path loses parallelism, while `ReduceHelper`'s MIDDLE submissions can no longer be queued behind a full decode window. LOW is `CPU_NUM`-sized and decode is CPU-bound, so the cores are still saturated. The code comment records the invariant this relies on: the calling thread blocks on decode futures, so the function must not be called from a LOW-pool thread (no caller does — index build enters segcore from Go).

Also documented under `common.storage.indexBuildReadWindowBytes` that decode runs on the LOW pool, deliberately not the pool serving search reduce.

**Caveat, stated explicitly rather than glossed over:** the 2x measurement in the PR body was taken with decode on MIDDLE. The LOW variant has not been re-benchmarked. Decode is CPU-bound so per-core throughput should be unchanged, but the in-flight batch cap drops from `max(2, 16*2)` to `max(2, CPU_NUM*2)` (the 512MB byte budget usually binds first). Treat the number as "measured on the MIDDLE variant" until a re-run says otherwise.

## czs007 — Low: partial apply when the window is rejected after the pool is resized

Fixed. `InitLoonReaderConfig` now range-checks **both** values before applying either, so an out-of-range window can no longer leave a resized (non-destroyable) reader pool behind. The 4GiB limit is mirrored on the Go side with a comment pointing at `storage_c.cpp` as the authority. The hot-reload failure log now names both values and says the previous settings stay in effect.

## czs007 — Low / @xiaocai2333 #1: the "not fully applied" warning fires under the default config

Fixed. The condition is now `requested == 0 && effective > 1`. `GetParallelism()` reporting 1 with no pool created is the pool being absent, not a failure to destroy it. Verified from the test log — hot-updating only the window with `readerThreadPoolSize=0` now emits just:
```
["loon reader params reconfigured"] [readerThreadPoolSizeRequested=0] [readerThreadPoolSizeEffective=1] [indexBuildReadWindowBytes=536870912]
```
with no warning.

## czs007 — Low: readerThreadPoolSize is wired only on DataNode

Documented rather than widened: the doc string now states the key applies to DataNode only and that query-node readers stay at parallelism 1 regardless. Wiring the same init into QueryNode would change the segment-load path in a way this PR has not measured; that belongs in its own change.

## @xiaocai2333 #2: hot-reload semantics for active readers

Documentation corrected in both the param doc and the watcher comment: a reader latches the parallelism it saw at open only as an on/off gate — opened with <= 1 it stays sequential for its lifetime, opened with > 1 it follows the pool's current size on each subsequent round, so a non-zero resize does affect later rounds of already-open readers.

## @xiaocai2333 #3: int32 narrowing

Fixed. The pool size is read as int64 and range-checked to `[0, 1024]`, returning `merr.ErrParameterInvalid`; `4294967296` is now rejected instead of wrapping to 0 and silently disabling the pool. (Note a pre-existing paramtable-wide behavior remains: a value that does not parse as int64 at all still falls back to 0. Out of scope here.)

## Verification

- `make build-cpp` / `ninja all_tests`: 0 errors (a cmake reconfigure was needed locally to pick up master's new GIS sources).
- C++: 88 passed / 4 skipped across `StorageTest.*`, `StorageV2*`, `*RawData*`, `FlushGrowingSegmentTest.*` (skips are Vortex/text-config gated). `FlushGrowingSegmentTest.IterateFieldDataFromManifestMultiBatchOrdering` passes on the LOW pool — multi-batch delivery order and values unchanged.
- Go: `internal/util/initcore` package green, including two new tests (`TestInitLoonReaderConfig`, `TestRegisterLoonReaderConfigWatchers`) covering the range rejection, the "pool must not be resized when the window is rejected" guard, and the warning condition. `pkg/util/paramtable` green.
- `gofmt` / `git diff --check` clean.

czs007:

Verified:
- `internal/core/src/storage/Util.cpp:1693-1804` — the bounded decode window and its backpressure behavior were examined directly against the 512 MiB budget.
- `internal/core/src/storage/Util.cpp:1719-1725` — the pending-worker drain path on reader failure was traced to confirm how outstanding decode futures are awaited.
- `internal/core/src/storage/Util.cpp:1802-1816` — the consumer invocation and stack-unwind paths were checked for outstanding prefetched work.
- `internal/core/src/storage/DiskFileManagerImpl.cpp:999-1005` — the local raw-data write consumer was inspected as the concrete throw site feeding the streaming pipeline.

xiaocai2333:
/ci-rerun-e2e-default

The e2e-default failure on `5a5c1dda06` is an infrastructure failure, not a code one. Only job-6 (`standalone-kafka-mmap`) actually failed; the other five were aborted by fast-fail.

Root cause chain in [job-6](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-4am/36720/pipeline-overview/): the 4AM cluster could not schedule the pods — `0/36 nodes are available: 14 node(s) didn't find available persistent volumes to bind, 3 Too many pods`. kafka-0/kafka-2 and milvus-standalone sat in `FailedScheduling` for ~17-20 minutes; kafka-2 only came up three minutes before the deadline, kafka-exporter hit `CrashLoopBackOff` eight times with `Error Init Kafka Client: kafka: client has run out of available brokers`, and `helm install --wait` then blew its 20m timeout. `deploy-helm` failed, so `check-status` and `test-sh` never ran a single test.

No panic, SIGSEGV or segcore error appears anywhere in the log — the only container log collected is kafka-exporter's connection failure. The same commit is green on `ci-v2/build-ut-cov` (8447 C++ tests, plus Code-Check / UT-GO / UT-Integration) and `ci-v2/go-sdk`.

internal/core/src/storage/Util.cpp
```diff
@@ -1666,23 +1755,108 @@ GetFieldDatasFromManifest(
...
-                                             nullable,
-                                             element_type.value(),
-                                             normalize_field_meta.value());
+        // Estimate the decoded footprint of this batch for the byte budget.
+        // The arrow buffers backing the batch are the best available proxy
+        // for the FieldData it will materialize into; fall back to the row
+        // count when the size cannot be computed.
+        int64_t batch_bytes = 0;
+        {
+            auto size_result = arrow::util::TotalBufferSize(*batch);
```
weiliu1031: [P2] Account for the referenced slice instead of the entire backing buffer

milvus-storage returns large row groups in `reader.record_batch_max_rows` slices (8192 rows by default), while Arrow's `TotalBufferSize` explicitly counts the entire backing buffer for a sliced array. As a result, multiple slices from the same row group can each be charged the full row-group size.

This can fill `kMaxInflightBytes` much earlier than the actual decoded footprint, reducing the pipeline to nearly serial decode for scalar or low-dimensional data. It also inflates the `MB` value reported by `[ManifestReadStats]`.

Please use `arrow::util::ReferencedBufferSize(*batch)` and handle its `Result`, or use another slice-aware estimate. Please also add a regression test where one row group produces multiple record-batch slices.

internal/core/src/storage/Util.cpp
```diff
@@ -1651,12 +1658,94 @@ GetFieldDatasFromManifest(
...
+    // ReduceHelper submits search reduce tasks and then blocks on
+    // future.get(). In standalone both share one ThreadPools singleton, so
+    // submitting a whole decode window to MIDDLE would queue CPU-bound
+    // tasks ahead of reduce and hold search latency up for the entire
+    // download phase. LOW carries only background segment loads and, being
+    // CPU_NUM-sized, still saturates the cores for this CPU-bound decode.
+    // Invariant this relies on: the calling thread blocks on decode futures,
+    // so this function must not be called from a LOW-pool thread. Index
+    // build tasks enter segcore from Go, so no caller does today.
+    auto& pool = ThreadPools::GetThreadPool(ThreadPoolPriority::LOW);
```
weiliu1031, line 1723: [P2] Re-benchmark the final LOW-pool implementation before attributing the performance result to this head

The PR description still says that batch decode runs on the MIDDLE pool and describes the 283s -> 151s result as measured with the "same code", but the current head submits decode tasks to the LOW pool.

This changes both the available worker count and `max_inflight_batches`, which is derived from `pool.GetMaxThreadNum() * 2`. The author also confirmed in the follow-up comment that the LOW-pool variant has not been benchmarked:
https://github.com/milvus-io/milvus/pull/51708#issuecomment-5088586530

Therefore, the existing measurement does not validate the final implementation. Please either repeat the same A/B benchmark on the current head, or update the PR description to state that the numbers apply only to the previous MIDDLE-pool variant and remove the "same code" and current-head ~2x claims.

xiaocai2333: Updated the PR description. It now states that the final implementation uses the LOW pool and labels 283s to 151s as historical data from the previous MIDDLE-pool iteration. It explicitly says those numbers do not validate a current-head ~2x claim and that a same-environment LOW-pool A/B benchmark remains follow-up work.

internal/core/src/storage/Util.cpp
xiaocai2333: Fixed in 7a7b21d76b. The in-flight byte accounting now uses arrow::util::ReferencedBufferSize through a slice-aware helper and falls back to row count if Arrow returns an error. Added a regression test with two record-batch slices sharing one backing allocation; it verifies TotalBufferSize double-charges the backing buffers while the production estimate partitions the referenced bytes. The focused C++ regression and existing multi-batch ordering test both pass.

internal/storagev2/packed/packed_writer_ffi.go
```diff
@@ -66,6 +66,7 @@ func CreateStorageConfig() *indexpb.StorageConfig {
 			UseVirtualHost:    paramtable.Get().MinioCfg.UseVirtualHost.GetAsBool(),
 			CloudProvider:     paramtable.Get().MinioCfg.CloudProvider.GetValue(),
 			RequestTimeoutMs:  paramtable.Get().MinioCfg.RequestTimeoutMs.GetAsInt64(),
+			MaxConnections:    uint32(paramtable.Get().MinioCfg.MaxConnections.GetAsInt()),
```
liliu-z, line 72: The value set at `packed_writer_ffi.go:69` is dropped because the Go FFI marshaller `MakePropertiesFromStorageConfig` (ffi_common.go) never appends `PropertyFSMaxConnections`. As a result the setting is dead for the packed-writer/manifest/transaction/LOB-GC paths, so their FS max-connections config silently has no effect (the C++ index-build path is fine). Append `PropertyFSMaxConnections` in the FFI properties marshaller.

czs007:
Adversarial review confirmed 3 issues (2 medium, 1 low) after verification; none are blockers, but the two medium items should be addressed before merge.

### Medium

- **`storage_c.cpp:221` — The reader thread pool is a process-global singleton, so the documented DataNode/QueryNode isolation does not hold in standalone mode.** `InitLoonReaderThreadPool` calls `milvus_storage::ThreadPoolHolder::WithSingleton(num_threads)`, which mutates a process-wide static singleton. The docs at `component_param.go:794-796` claim "query-node readers are unaffected and keep parallelism 1 no matter what this is set to" — true only in cluster mode. In standalone, DataNode and QueryNode share a process: after DataNode's `InitSegcore` (`data_node.go:221`) creates the pool, any QueryNode reader opened afterwards picks up the global parallelism at open time (`reader.cpp:108`) and, per the PR's own docs at `component_param.go:808-813`, follows the pool's current size on each round — fanning out on the shared pool and adding S3/CPU contention to the load path. Since the default (0) is opt-in, the fix is documentation: state that the setting is process-global and, in standalone, also affects QueryNode readers opened after DataNode init, rather than asserting isolation. *(raised by xaxys, tinswzy)*

- **`internal/util/initcore/util.go:207` — Hot-reload callbacks are not serialized, so pool resizes can apply in the reverse order of config writes.** The config watcher handler reads the live paramtable value (`init_core.go:539`) and then calls `C.InitLoonReaderThreadPool`, but event dispatch runs handlers inline on the caller's goroutine under a shared RLock (`event_dispatcher.go:42-57`; `base_table.go:301-306`), with no cross-call ordering guarantee. Two concurrent updates (set→16 then set→8) can interleave so the later write's resize lands first, leaving the pool at 16 while paramtable reports 8 — and the `requested == 0 && effective > 1` warning at `util.go:229` does not fire in that case. Real-world likelihood is low, but the resize is non-idempotent; adding a single-flight or mutex around `InitLoonReaderConfig` would close it. *(raised by xaxys, tinswzy)*

### Low

- **`internal/core/src/storage/Util.cpp:1761` — The in-flight byte budget is charged on pre-decode Arrow input, so decoded memory can exceed the 512MB cap on the external-collection path.** `batch_bytes = EstimateRecordBatchBytes(*batch)` measures the input Arrow slice, yet the comment at `Util.cpp:1675-1677` claims the budget "caps the decoded-but-undelivered bytes." What actually accumulates in `pending` is the decoded `FieldDataPtr` output (`Util.cpp:1784-1786`), plus per-batch `NormalizeExternalArrowByType` intermediates and the captured source batch. For internal native types input ≈ output, but on the external path normalization can inflate the decoded footprint by a constant factor, so peak retained memory can exceed `kMaxInflightBytes` (512MB) by that factor times the number of concurrently-decoding batches. Bookkeeping is symmetric (`deliver_front` discharges the same amount at `Util.cpp:1723`), so this is not a leak — but either charge a post-decode estimate or correct the comment (and the matching wording at `Util.h:326-333`) to say the bound is on input Arrow bytes, not decoded footprint. *(raised by tinswzy)*

czs007:
Adversarial review of this PR surfaced one medium-severity configuration gap and two low-severity, non-blocking issues; no blocking defects were found.

### Medium

- **`internal/core/src/storage/loon_ffi/util.cpp:307` — Configured `max_connections` never reaches the external-collection (extfs) read path.** The PR plumbs `minio.maxConnections` into the `fs.*` scope (`loon_ffi/util.cpp:244-245`, `ffi_common.go:169-170`), but external-collection raw-data reads go through the `extfs.<collectionID>.*` scope, and neither `kExtfsFields` (`util.cpp:305-320`) nor `kAllowedExtfsSpecKeys` (`util.cpp:285-302`) includes `max_connections`. Per `util.h:100-103`, Layer 0 is zero-initialized with no `fs.*` inheritance, and the pinned milvus-storage source (f9f5bd1) confirms there is no fallback: `ExtractExternalFsProperties` collects only `extfs.<name>.*` keys, and a missing `PROPERTY_FS_MAX_CONNECTIONS` falls back to the registry's built-in default of 100 — never to `fs.*`. So for the external-read path this PR aims to speed up, `minio.maxConnections` is a confirmed no-op (masked today only because the default happens to also be 100). **Suggestion:** add `max_connections` to the extfs field contract (`kExtfsFields` and the Go `extfsFields`), or document that this setting does not apply to external tables. Note the core performance fix is unaffected: `PROPERTY_READER_RECORD_BATCH_MAX_SIZE` lives in the `reader.*` namespace and does reach external reads. *(raised by xaxys, tinswzy)*

### Low

- **`internal/core/src/storage/storage_c.cpp:221` — Process-global reader thread pool also affects QueryNode in standalone mode, contradicting the config docs.** `InitLoonReaderThreadPool` creates the pool via `ThreadPoolHolder::WithSingleton`, a process-level singleton (`static inline` members in milvus-storage `thread_pool.h:100-101`), and only DataNode's segcore init calls it (`internal/datanode/index/init_segcore.go:108`). In standalone deployments all roles share one process, so QueryNode-side readers latch `ThreadPoolHolder::GetParallelism()` (`reader.cpp:105`) and become parallelized once DataNode has created the pool. The claim in `component_param.go:794-796` that "query-node readers are unaffected and keep parallelism 1 no matter what this is set to" therefore does not hold in standalone. **Suggestion:** correct the documentation to state that the setting applies only to DataNode in cluster deployments, while in standalone mode co-located QueryNode readers also use this pool. Non-blocking. *(raised by xaxys, tinswzy)*

- **`internal/core/src/storage/Util.cpp:1719` — Non-`ReadNext` throw sites propagate exceptions without draining the pending decode queue.** In `IterateFieldDataFromManifest`, only the `ReadNext` failure path drains `pending` before throwing (`Util.cpp:1741-1747`). If `pending.front().first.get()` (`:1721`, rethrowing a decode-task exception) or `consumer(std::move(field_data))` (`:1725`, which can throw during index build's local disk write) throws, the exception propagates without waiting for in-flight decode tasks. This is safe today only because the decode lambda (`Util.cpp:1763-1771`) captures everything by value, so no task references the unwinding stack scope — but a future switch to reference captures or shared mutable state would introduce dangling-reference risk. **Suggestion:** consolidate the drain into a single scope guard covering all exit paths. *(raised by tinswzy)*

internal/storagev2/packed/packed_writer_ffi.go
xiaocai2333: Good catch — this was already fixed in 6930b6ffe (pushed shortly after this comment). `MakePropertiesFromStorageConfig` now appends the property alongside `PropertyFSRequestTimeoutMS`:

```go
// internal/storagev2/packed/ffi_common.go:169
keys = append(keys, PropertyFSMaxConnections)
values = append(values, strconv.FormatUint(uint64(storageConfig.GetMaxConnections()), 10))
```

The constant is declared at `ffi_common.go:58` (`C.loon_properties_fs_max_connections`), and `exttable_test.go` asserts it round-trips (`assert.Equal(t, "237", loonPropertyString(props, PropertyFSMaxConnections))`). So the packed-writer / manifest / transaction / LOB-GC paths now do pick the setting up.

czs007:
## Summary

Adversarial review surfaced four issues: an unconditional `max_connections="0"` emit that regresses the S3 connection cap and splits the filesystem cache, two `StorageConfig` producers that still leave the field unset, an unbudgeted parallel-decode window on the index-build read path, and one inert-but-trappy config entry.

---

### High

**`internal/storagev2/packed/ffi_common.go:169` — emitting `max_connections="0"` regresses the S3 connection cap and splits the filesystem cache**

The producer now unconditionally emits `fs.max_connections`, so a `StorageConfig` whose Go-side `MaxConnections` is unset is written as the string `"0"` where the key used to be absent. milvus-storage's registry returns the registered default only when the key is *missing* (`properties.cpp:617-630`), and `PROPERTY_FS_MAX_CONNECTIONS` is registered with default `uint32_t(100)` (`properties.cpp:392-396`). An explicit `"0"` therefore clobbers that default: `s3_client_builder.cpp:167-172` computes `max(max(io_capacity, 25), options_.max_connections)`, so the effective cap drops from ≥100 to ≥25 (or the io-context capacity). Nothing clamps 0 on the way in (`fs.cpp:137` → `s3_filesystem_producer.cpp:173`). The external scope inherits the same zero — `ExtractExternalFsProperties` maps `extfs.<cid>.max_connections` → `fs.max_connections` via `SetValue(..., allow_undefined_key=false)` (`fs.cpp:198-202`), and `"0"` is a type-valid `UINT32`, so it is accepted rather than rejected. `ArrowFileSystemConfig::GetCacheKey()` hash-combines `max_connections` (`fs.h:255`, `fs.h:278`) and that key drives the filesystem cache (`fs.cpp:227`), so differing values also split the cache.

This is reachable on this branch: QueryNode's `segment_loader.go` and RootCoord's `createMilvusTableSnapshotStorageConfig` still leave the field zero (see the companion finding), and during a rolling upgrade a new DataNode receiving a `StorageConfig` from a not-yet-upgraded DataCoord also gets 0. The cache split has a visible second effect inside a single QueryNode process: `services.go:590` publishes metrics for a config built with `MaxConnections=100` (`filesystem_metrics.go:289`) while the loader reads through a filesystem keyed on 0, so published storage metrics no longer describe the filesystem actually in use. Same emit pattern at `internal/storagev2/filesystem_metrics.go:165-166`.

*Suggestion:* skip the key when the value is 0, or default 0 → 100 at the producer. *(raised by czs007)*

---

### Medium

**`internal/querynodev2/segments/segment_loader.go:1576` — two `StorageConfig` producers still leave `MaxConnections` unset, which is what makes the `"0"` emit reachable**

An exhaustive sweep of non-test `indexpb.StorageConfig{` producers finds 7 sites; this PR populates 5 (`compaction/params.go`, `datacoord/util.go`, `datacoord/garbage_collector_lob.go`, `storagev2/filesystem_metrics.go`, `storagev2/packed/packed_writer_ffi.go`) and leaves 2 unpopulated: `querynodev2/segments/segment_loader.go` `createStorageConfig` (func at `:1568`, remote branch returns at `:1576-1594`) and `rootcoord/create_collection_task.go` `createMilvusTableSnapshotStorageConfig` (func at `:569`, remote branch returns at `:577-595`). Both feed `MakePropertiesFromStorageConfig` — the QueryNode one from six call sites on the segment-load path (`:1246`, `:1388`, `:1414`, `:1430`, `:1443`, `:1448`), the RootCoord one from the milvus-table snapshot read at `:510` — so these are exactly the paths that emit `max_connections="0"` and take the connection-cap regression and cache split described above.

*Suggestion:* populate both from `paramtable.MinioCfg.MaxConnections` (default 100) so the emit is consistent across producers. *(raised by czs007)*

**`internal/core/src/storage/Util.cpp:1704` — parallel decode window is unconditional and unbudgeted, adding up to 512MB per in-flight build on full-column paths**

`IterateFieldDataFromManifest` always submits per-batch decode to the LOW thread pool and always admits up to `kMaxInflightBytes` (512MB) with `max_inflight_batches = max(2, pool.GetMaxThreadNum() * 2)`. Neither knob added by this PR gates it: `common.storage.readerThreadPoolSize` and `common.storage.indexBuildReadWindowBytes` both default to `"0"` and are consumed only by the milvus-storage reader pool and the per-round read window, never by this decode path. (The param doc's "0 keeps the pre-existing sequential behavior" at `component_param.go:798` is scoped to the storage reader pool and is not itself wrong — the gap is that the decode fan-out is a separate, always-on change with no off switch.)

On the streaming caller `DiskFileManagerImpl.cpp:1020` the window replaces full-column retention and is a net reduction. But `GetFieldDatasFromManifest` (`Util.cpp:1866-1887`) still accumulates every `FieldDataPtr` into a vector, so its callers now hold the whole column *plus* the in-flight window (up to 512MB of arrow input bytes, whose futures also keep the source `RecordBatch`es alive, plus decoded-but-undelivered `FieldData`): `MemFileManagerImpl.cpp:258` (`cache_raw_data_to_memory_storage_v2`, reached from `VectorMemIndex.cpp:514` and `RTreeIndex.cpp:285` via `CacheRawDataToMemory`, and from every scalar build via `storage::CacheRawDataAndFillMissing`) and `MemFileManagerImpl.cpp:468` / `DiskFileManagerImpl.cpp:1446` (opt-field caching). This affects every storage-v3 index build, not just external collections. Vector builds are capped at `dataNode.index.maxVecIndexBuildConcurrency` (default 4), so peak RSS can rise by up to ~4× the window; scalar builds bypass that pool entirely (`scheduler.go:313-315` runs non-vector tasks directly) and are unbounded in count. The Go-side index-build memory estimate (`MergeResourceParams` at `task_index.go:272`) is derived from `fieldDataSize` only and does not account for this window, so tasks that previously fit can now OOM.

*Suggestion:* derive/gate the window from `indexBuildReadWindowBytes` (or make `kMaxInflightBytes` configurable) and charge it against the build memory budget; at minimum use a much smaller window in `GetFieldDatasFromManifest`, which retains the full column anyway. *(raised by czs007)*

---

### Low

**`internal/core/src/storage/loon_ffi/util.cpp:329` — `max_connections` entry added to `kExtfsFields` is inert and becomes a failure trap if the Layer-0 loop is generalized**

`kExtfsFields` has exactly one consumer, the Layer-0 loop at `util.cpp:526-530`, which only writes entries with `is_bool == true`, so the new `{"max_connections", false}` entry is dead — actual propagation happens entirely through `kExtfsInheritedFsFields` (`util.cpp:344-348`) and the Layer-0b loop (`util.cpp:537-543`). The list's header comment (`util.cpp:308-309`) claims lockstep with a Go `extfsFields` symbol in `internal/storagev2/packed/ffi_common.go`; that symbol does not exist anywhere in the repo on this branch and is not added by this PR, so the entry is being kept in sync with nothing. The latent trap: if the Layer-0 loop is ever extended to zero-init non-bool slots, it would write `extfs.<cid>.max_connections = ""` directly into the map, which `ExtractExternalFsProperties` turns into `SetValue("fs.max_connections", "")` with `allow_undefined_key=false` (`fs.cpp:198-202`) — an empty string is not a valid `UINT32`, so it returns `arrow::Status::Invalid` and every external read fails.

*Suggestion:* drop the entry — behavior stays byte-identical and the trap goes away. *(raised by czs007)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/d8d7ab361d9d6375c1fd1065034b3108a478bd90...czs007:team2-fix/Te0e9b09257?expand=1

czs007:

Adversarial review found one config-propagation regression that partially undoes this PR's own `max_connections` widening, one gap in the new task-slot accounting, and one thread-pool placement concern on the newly-async decode path.

## Medium

**`internal/core/src/storage/loon_ffi/util.cpp:535` — integral `0` is mirrored into `extfs.<cid>.max_connections`, capping the external read path at 25 instead of 100**

The new Layer 0b copies `fs.max_connections` verbatim into `extfs.<collectionID>.max_connections`, but `PropertyValueAsString` (`util.cpp:364,367`) only filters *empty strings* — an integral `0` renders as `"0"`. On the C++ index-build path the key is always present and unguarded: `MakeInternalPropertiesFromStorageConfig` writes it unconditionally (`util.cpp:245-249`) from `storage_config.max_connections`, which `indexbuilder/index_c.cpp:129` assigns straight from the proto, overwriting the struct default of 100 in `storage/Types.h:129`.

Failure: during a rolling upgrade an older DataCoord that predates `MaxConnections` in `createStorageConfig` (`datacoord/util.go:398`) sends a zero-valued `indexpb.StorageConfig`, which `datanode/index/task_index.go:289` passes through verbatim → `extfs.<cid>.max_connections="0"` overrides milvus-storage's registry default of 100, and `s3_client_builder.cpp:169-172` (`maxConnections = max(max(io_capacity, 25), options_.max_connections)`) degrades the external S3 client to 25 connections — on exactly the read path this PR widens. A second reachable route: with `common.storageType: local`, `datacoord/util.go:377-381` populates only `RootPath`/`StorageType`, leaving `MaxConnections` at 0 while the collection's external data still lives on S3. Before this PR the key was absent, so `ExtractExternalFsProperties` → `GetValue<uint32_t>` fell back to 100. The explicit `"0"` also changes `ArrowFileSystemConfig::GetCacheKey()` (`fs.h:278` hash_combines `max_connections`), so producers that omit the key resolve to a different cached filesystem in the same process.

Suggestion: apply the same `> 0` convention the Go producer already uses (`storagev2/packed/ffi_common.go:178-181`, whose comment spells out the clobber reasoning) — either skip the mirror when the rendered value is `"0"`, or add the `> 0` guard in `MakeInternalPropertiesFromStorageConfig` to match `ChunkManager.cpp`/`MinioChunkManager.cpp`. Please also add a present-but-zero case to `InjectExtfsInheritedFields` in `test_external_take.cpp`; the four existing cases cover `"237"`, UINT32 64, absent key, and spec-not-overridable, but not zero. *(raised by xaxys, czs007)*

**`internal/datacoord/util.go:435` — task slot charges the 64MiB decode window but not the new `indexBuildReadWindowBytes` (up to 4GiB) per-task retention**

`calculateIndexTaskSlot` adds only `min(fieldSize, indexBuildDecodeWindowBytes)` (`util.go:422`, 64MiB, mirroring `Util.h:351`'s `kAccumulatingInflightBytes`). The `common.storage.indexBuildReadWindowBytes` knob this PR also introduces (`component_param.go:830`, max 4GiB) is not charged at all, yet it is genuine per-task resident memory: `ApplyIndexBuildReadWindow` (`index_c.cpp:225-228`) writes it into each build task's loon properties as `reader.record_batch_max_size`, and the reader fills a whole round up to `memory_usage_limit_` before delivering any batch (`reader.cpp:101-102`, `:420-437`), holding it in `current_rbs_`. The param's own doc says "up to this many bytes buffered per running build task, on top of the decode window" (`component_param.go:844`) and recommends N × row-group-size — its example is a 1GiB window over 64MB row groups, an order of magnitude above what is charged. For accumulating callers (`GetFieldDatasFromManifest` → `MemFileManagerImpl`, i.e. all in-memory storage-v3 builds, `Util.cpp:1875-1900`) this is purely additive: the full column is retained *plus* the reader round *plus* the decode window.

Failure: an operator follows the doc and sets `indexBuildReadWindowBytes=1GiB` while a DataNode runs several concurrent manifest-backed in-memory builds → each task additionally retains ~1GiB of arrow chunks but DataCoord charges only 64MiB extra per task, so concurrency is admitted against an underestimate and the index pool OOMs precisely when this tuning knob is enabled.

Suggestion: either charge the configured window alongside the decode window (`fieldSize += min(fieldSize, Params.CommonCfg.IndexBuildReadWindowBytes.GetAsInt64())` under `manifestBacked`), or state in the param doc that raising this knob requires raising `dataCoord.slot.indexTaskSlotUsage` correspondingly. *(raised by xaxys, czs007)*

## Low

**`internal/core/src/storage/Util.cpp:1723` — batch decode is submitted to the LOW pool and awaited on the calling thread, which in standalone queues behind blocking segment-load I/O**

Each batch decode goes to `ThreadPoolPriority::LOW`, and the reader thread then blocks on `pending.front().first.get()` in `deliver_front`. The MIDDLE-vs-LOW comment weighs only CPU contention with search reduce, but LOW is also where `GetObjectData` (`Util.cpp:1032`) queues one *blocking* object-storage read per binlog file via `DownloadAndDeserialize`, and `PriorityForLoad` maps `LoadPriority_LOW` straight to `ThreadPoolPriority::LOW` (`ThreadPools.h:33-36`) — the routine priority for balance, index-update, and stats-update loads (`index_checker.go:243,298`, `balance_checker.go:368`). `ThreadPools::thread_pool_map` is a process-level static (`ThreadPools.cpp:30-31`), so in standalone QueryNode and DataNode share it.

Failure: in standalone, an index build completes and `index_checker`/`balance_checker` issue `LoadPriority_LOW` segment loads that enqueue thousands of blocking S3 reads; a concurrent manifest-backed build's decode tasks queue behind them and the reader thread stalls in `deliver_front`, degrading the fetch/decode pipeline to serial waits and potentially making the download phase slower than pre-PR, where decode ran inline on the caller thread. This is not a deadlock — none of the four production callers in `DiskFileManagerImpl`/`MemFileManagerImpl` run on a LOW thread — and `[ManifestReadStats]`'s `decode_wait_ms` makes it diagnosable.

Suggestion: move decode to a pool that does not carry blocking I/O (or a dedicated one); if LOW is kept deliberately, correct the comment to state that the pool is shared with blocking load I/O in standalone, so the failure mode is head-of-line blocking rather than CPU contention. *(raised by czs007)*

czs007:

Adversarial review of this PR surfaced one Medium issue in the new byte-budget accounting, plus two real-but-pre-existing hazards tracked separately below.

## Medium

- **`internal/core/src/storage/RecordBatchSize.h:32` — the byte budget silently degrades to a row count for arrow view types, defeating the inflight-bytes cap this PR adds.** `EstimateRecordBatchBytes` falls back to `batch.num_rows()` when `arrow::util::ReferencedBufferSize` fails, and on the pinned arrow 17.0.0 (`internal/core/conanfile.py:15`) it always fails for `BinaryViewType` / `StringViewType` / `ListViewType`: the `GetByteRangesArray` visitor has no overload for them (`arrow/type.h:783`, `:926`, `:1177` — `ListViewType` is a sibling of `ListType`, not derived from it), so they hit the `Status::TypeError` catch-all at `byte_size.cc:318`, and `ReferencedBufferSize(const RecordBatch&)` propagates the first failing column via `ARROW_ASSIGN_OR_RAISE` (`byte_size.cc:385-392`). That failing column is the target column, since the reader projects exactly one (`Util.cpp:1611`), and the estimate runs on the *pre-normalization* batch (`Util.cpp:1812`) while `CanonicalizeArrowVariants` — which exists precisely to eliminate `STRING_VIEW`/`BINARY_VIEW`/`LIST_VIEW` (`Util.cpp:2792-2796`, `2830`, `2872`) — is only reached later, inside the decode lambda via `NormalizeExternalArrowByType` (`Util.cpp:1826` → `3319`). The vortex schemaless external reader is a real source of view layouts (`Util.cpp:2540`, `2780`; `Consts.h:159`). Failure: building an index over an external collection whose target column is VARCHAR/JSON/ARRAY read by the vortex schemaless reader, on the streaming path with `max_inflight_bytes = kStreamingInflightBytes` (512MB) → each batch is charged only its row count, `pending_bytes` never approaches 512MB, byte backpressure never triggers, and inflight batches are bounded solely by `max_inflight_batches = max(2, GetMaxThreadNum() * 2)` (`Util.cpp:1724`) — 16 concurrent 64MB-class arrow batches on 8 cores, 64 on 32 cores — pushing peak RSS to several times the intended ceiling and OOM-ing the build task, since the Go-side `MergeResourceParams` estimate does not account for this memory. Suggestion: for types the visitor cannot handle, fall back to `TotalBufferSize` (a plain buffer sum that works for any type) rather than to `num_rows`. *(raised by czs007)*

- **`internal/core/src/storage/Util.cpp:2980` — the GEOS context leaks when WKT parsing throws.** `ConvertWKTStringArrayToWKBBinary` calls `GEOS_init_r()` at `Util.cpp:2980` and `GEOS_finish_r(ctx)` only on the normal return path (`Util.cpp:3026`), while `Geometry geom(ctx, wkt.c_str())` (`Util.cpp:3013`) throws `SegcoreError` via `AssertInfo` on invalid WKT (`common/Geometry.h:81-82`). Failure: a GEOMETRY column in an external collection containing one unparseable WKT string (truncated, invalid geometry, empty) → the construction throws, `GEOS_finish_r` is skipped, the handle leaks permanently, and the error falls through to the default branch at `internal/datanode/index/scheduler.go:228` (`JobState_JobStateRetry`), so datacoord re-submits the task and leaks another handle on each retry, monotonically growing RSS in the long-lived datanode. Fix direction for the follow-up: manage the context with RAII, or reuse the existing `GetThreadLocalGEOSContext()` (`common/Geometry.h:31-43`), whose thread-local destructor already calls `GEOS_finish_r`. Unrelated to this PR: the function is untouched by this diff and `milvus/master` already has the same non-RAII `GEOS_init_r()` at `Util.cpp:2763` — the leak occurs with or without these changes; parallel decode only raises the number of leaked handles per task from one to the inflight-batch count. *(raised by xaxys)*

- **`internal/datacoord/util.go:377` — `MaxConnections` stays 0 on the `local` storage branch, so `minio.maxConnections` remains ignored in that deployment shape.** The `local` branch (`util.go:377-381`) builds only `RootPath`/`StorageType`, leaving `MaxConnections` at 0, so every `> 0` guard added by this PR (the two `loon_ffi/util.cpp` producers, `internal/storagev2/packed/ffi_common.go:2123`, and the extfs mirror) correctly omits the key and milvus-storage's registered default of `uint32(100)` applies (`cpp/src/properties.cpp:392-396` at `f9f5bd1`, pinned by `internal/core/thirdparty/milvus-storage/CMakeLists.txt:17`). Failure: a `common.storageType=local` cluster whose external collection data actually lives in S3, where an operator raises `minio.maxConnections` to 500 → `createStorageConfig` takes the `local` branch, neither `fs.max_connections` nor `extfs.<cid>.max_connections` is emitted, the external S3 client uses the built-in default of 100, and the configured value is dropped with no log indication. Unrelated to this PR: `master`'s `createStorageConfig` sets `MaxConnections` on neither branch, so the gap predates this change on both paths; this PR only fills in the `minio` branch, leaves the `local` line untouched, and its `> 0` guard actually raises the effective value there from `max(io_capacity, 25)` to the default 100. Suggestion for the follow-up: populate `MaxConnections` on the `local` branch as well (external data may still be on S3), or document that the omission is intentional. *(raised by czs007)*

internal/core/src/storage/RecordBatchSize.h
```diff
@@ -0,0 +1,38 @@
...
+
+// Returns the bytes referenced by this record-batch slice. Unlike
+// TotalBufferSize, this accounts for array offsets and does not charge every
+// slice for its entire shared backing buffer. Falls back to the row count if
+// Arrow cannot compute the referenced ranges.
+inline int64_t
+EstimateRecordBatchBytes(const arrow::RecordBatch& batch) {
+    auto size_result = arrow::util::ReferencedBufferSize(batch);
+    if (!size_result.ok()) {
+        return batch.num_rows();
```
marcelo-cjl: [P1] Arrow view types bypass the byte budget

In the pinned Arrow 17 implementation, ReferencedBufferSize returns NotImplemented for STRING_VIEW, BINARY_VIEW, and LIST_VIEW. This fallback then returns batch.num_rows(), while the caller treats the result as bytes. These view types can be produced by external/Vortex readers and are normalized only later, inside the submitted decode task.

For example, a 64 MiB batch containing 8,192 rows would be charged as only 8,192 bytes. The retained Arrow memory therefore diverges sharply from pending_bytes, making the byte-based backpressure ineffective. The separate batch-count limit can still allow several GiB to remain in flight across concurrent index builds, potentially causing OOM.

internal/datacoord/util.go
```diff
@@ -395,6 +395,7 @@ func createStorageConfig() *indexpb.StorageConfig {
 			UseVirtualHost:    Params.MinioCfg.UseVirtualHost.GetAsBool(),
 			CloudProvider:     Params.MinioCfg.CloudProvider.GetValue(),
 			RequestTimeoutMs:  Params.MinioCfg.RequestTimeoutMs.GetAsInt64(),
+			MaxConnections:    uint32(Params.MinioCfg.MaxConnections.GetAsInt()),
```
marcelo-cjl, line 403: [P2] Local primary storage drops max-connections for external S3 reads

MaxConnections is populated only in the non-local branch. When common.storageType is local, createStorageConfig leaves this field at zero even though an external collection may still reference an s3:// URI. The C++ property construction omits the zero value, so the external filesystem cannot inherit the configured minio.maxConnections and instead uses its default connection limit.

As a result, the operator setting is ignored for the local-primary-storage plus external-S3 topology, leaving the max-connections propagation incomplete for this supported path.

czs007:

Verified:
- Both independent reviewers completed a final re-review round and confirmed no additional issues beyond what was already resolved — no findings survived to the reporting stage.

PR #51709, opened 2026-07-22
fix: [3.0] keep BM25/MinHash searchable after add_function_field with stale segment index meta

MrPresent-Han:
issue: #51366
pr: #51692

Cherry-pick of #51692 to the 3.0 branch.

## Problem

A field added by `add_function_field` (a BM25 or MinHash function output field) can stay unsearchable: the search keeps failing with `field index meta not found for field N` (an `AssertInfo` in `IndexMeta.cpp`) even though `describe_index` reports the index `Finished`, until the collection is released and loaded again.

Root cause: a segment captures its index meta at construction time — the per-segment `ChunkedSegmentSealedImpl::col_index_meta_`, and the growing `IndexingRecord::index_meta_`. A schema-only reopen refreshes the schema but not this snapshot, so a field added by `add_function_field` is missing from it. On segments without a built sparse index (sealed segments below the index-build threshold, and growing segments), BM25/MinHash brute force fetched the field's index params via `GetFieldIndexMeta`, which asserts on the missing field.

## Fix

- Guard the brute-force index-param fetch at both `vector_search` entry points (`ChunkedSegmentSealedImpl` and `SegmentGrowingImpl` via `SearchOnGrowing`) with `HasField`, so a field missing from the stale snapshot no longer trips the assert.
- Source the brute-force index params (BM25 `k1`/`b`, MinHash `mh_lsh_band`/`mh_element_bit_width`) from the collection-level index meta at plan creation (`plan_c.cpp`) into `SearchInfo::brute_force_index_params_` — the same source already used for `metric_type`. Brute force falls back to it when the per-segment meta lacks the field. These params never route through `search_params`, so the query hook cannot drop them, and MinHash defaults to knowhere's construction values when the index omits them.

### DIFF
--- a/internal/core/src/common/QueryInfo.h
+++ b/internal/core/src/common/QueryInfo.h
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <memory>
+#include <optional>
 
 #include "ArrayOffsets.h"
 #include "common/Tracer.h"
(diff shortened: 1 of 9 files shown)


liliu-z (commented):
/lgtm
/approve

PR #51717, opened 2026-07-22
fix: [cp3.0] keep bitmap cursors aligned on skipped offset paths

liliu-z:
Backport of #51540 to the 3.0 branch.

pr: #51540
issue: #51537
follow-up issue: #51736
proposed master fix: #51737

## What changed

- Preserve the evaluator callback contract for SkipIndex-pruned row and element candidates by applying validity first, then invoking the evaluator with a null batch so callback-local bitmap cursors still advance.
- Keep nullable scalar-index reverse-lookup misses aligned and stop applying raw chunk-0 SkipIndex statistics to the segment-wide scalar index.
- Use `IArrayOffsets` as the logical source of truth for flattened ARRAY elements, including nullable rows with retained physical payloads.
- Fix growing-segment ARRAY length recovery to read the dense `Array` layout and assign zero logical elements to invalid rows.
- Make the remaining TIMESTAMPTZ, unary ARRAY, and unary JSON offset evaluators safe for null batches produced by reverse-lookup misses.
- Backport 15 focused regressions covering growing and sealed row offsets, element offsets, element full scans, nullable ARRAY accounting, and scalar-index-only evaluation.

## Scope

This backport keeps the scope of #51540. It fixes logical-candidate accounting when an evaluated batch contains elements.

It does not change the existing framework behavior for a whole row batch whose logical element count is zero. That distinct follow-up is tracked in #51736, with the proposed master fix in #51737.

## Backport notes

- Cherry-picked source commits `dc33021eb740d7e6c2f56756f72850540ab919c1` and `193e29e9a313d76df7f911ebec7f3080b46ca331` in source order with provenance and DCO trailers.
- Both commits applied cleanly on the latest `upstream/3.0` with no 3.0-specific code changes.
- Each stable patch-id matches its source commit exactly; range-diff differs only by cherry-pick provenance and Li Liu sign-off trailers.
- The only source-parent versus 3.0-base differences in touched pre-existing files are unrelated null-rejecting expression support and a growing column-load unwind guard; neither overlaps the backported hunks.
- Audited all 3.0 production callers of the affected offset and element helpers; their evaluators are null-batch safe after this patch.

## Verification

- `git diff --check upstream/3.0...HEAD` — passed.
- `clang-format-15 --dry-run --Werror` on all changed C++ files — passed.
- Source PR CI Loop on the exact current head `193e29e9` passed build, code-check, integration and Go tests, plus 8,249/8,249 C++ tests.
- Local full C++ rebuild was not completed because the macOS arm64 Conan profile has no compatible binaries and began rebuilding the full third-party dependency graph; target-branch CI for this backport must provide branch-specific compile and regression coverage.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -603,6 +603,16 @@ class SegmentExpr : public Expr {
                                               std::move(valid_res));
     }
 
+    // Candidate evaluator contract:
+    // - every logical candidate position advances the evaluator exactly once;
+    // - data == nullptr means the reader already decided the candidate result
+    //   (for example SkipIndex pruning or a NULL reverse lookup), so the
+    //   evaluator must only advance position-indexed state by size and return;
+    // - result and validity for a null batch are owned by the reader.
+    //
+    // This contract keeps callback-local bitmap_input cursors aligned across
+    // raw-data, element-level, and scalar-index-only candidate paths.
+
     // when we have scalar index and index contains raw data, could go with index chunk by offsets
     template <typename T, typename BatchEvaluator, typename... ValTypes>
     int64_t
@@ -1052,6 +1062,11 @@ class SegmentExpr : public Expr {
                       "Json element type is not supported for "
                       "element-level filtering");
 
+        auto array_offsets = segment_->GetArrayOffsets(field_id_);
+        AssertInfo(array_offsets != nullptr,
+                   "ArrayOffsets not found for field {}",
+                   field_id_.get());
+
         int64_t processed_rows = 0;
         int64_t processed_elems = 0;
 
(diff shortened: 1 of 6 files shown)


internal/core/src/exec/expression/Expr.h
```diff
@@ -1156,16 +1107,16 @@ class SegmentExpr : public Expr {
...
-                                    valid_res[processed_elems + k] = false;
-                            }
-                        } else {
+                        const bool is_row_valid =
+                            !valid_data.data() || valid_data[j];
+                        // ArrayOffsets defines a NULL array row as having zero
+                        // logical elements even when its physical payload is
+                        // non-empty. Keep the flattened result aligned with
+                        // that logical address space.
+                        const auto elem_count =
```
buqian-zilliz, line 1116: **[P1] Advance past zero-element row batches**

Because null rows now contribute zero logical elements, an entire row batch can have `real_batch_size == 0` even when later rows still contain elements. The element-level visitors return `nullptr` before `ProcessDataChunksForElementLevel()` advances `current_data_chunk_`/`current_data_chunk_pos_`, and full-mode `ElementFilterBitsNode` then fails its non-null result assertion while the segment-wide element count is still positive. With the default 8,192-row batch, 8,192 null or empty arrays followed by a non-empty row deterministically fails the query. Distinguish a zero-element nonempty row batch from true exhaustion by advancing across such row batches and continuing until elements are produced or all rows are consumed.

liliu-z: Confirmed. This is the whole-zero-element-row-batch case explicitly excluded from #51540 scope. It is tracked in #51736, with the proposed master fix in #51737.

#51717 remains a patch-equivalent backport of #51540 and does not include this follow-up.

Issue #51249, opened 2026-07-10
[Bug]: LoadScope_Reopen fails to load DISKANN index because num_load_thread is missing


zhuwenxing:
# Issue Title

[Bug]: LoadScope_Reopen fails to load DISKANN index because num_load_thread is missing

## Environment

- Milvus version: `master-20260709-b542464f30` (`harbor.milvus.io/milvusdb/milvus:master-20260709-b542464f`)
- Deployment mode: cluster, streaming mode enabled
- MQ type: Woodpecker
- SDK version: PyMilvus `3.1.0rc63`
- OS: Rocky Linux 8.7, kernel `4.18.0-553.129.1.el8_10.x86_64`, amd64
- Kubernetes: v1.24.2
- CPU/Memory: 4 QueryNode replicas; each requests 2 CPU / 25 GiB and limits 8 CPU / 32 GiB
- GPU: N/A
- Others: `common.storage.useLoonFFI=true`; distributed MinIO; Milvus CR remains Healthy

### K8s Pod List

```text
stable-16cu-workload-etcd-0                                  1/1  Running  0  4h32m
stable-16cu-workload-etcd-1                                  1/1  Running  0  4h32m
stable-16cu-workload-etcd-2                                  1/1  Running  0  4h32m
stable-16cu-workload-milvus-datanode-76ff5464f6-hxvxl        1/1  Running  0  4h31m
stable-16cu-workload-milvus-datanode-76ff5464f6-kdv6h        1/1  Running  0  4h31m
stable-16cu-workload-milvus-datanode-76ff5464f6-kqf77        1/1  Running  0  4h31m
stable-16cu-workload-milvus-datanode-76ff5464f6-zpf8m        1/1  Running  0  4h31m
stable-16cu-workload-milvus-mixcoord-57cc9c65df-jxzv6        1/1  Running  0  4h31m
stable-16cu-workload-milvus-proxy-d85dd57f9-mx7hn            1/1  Running  0  4h31m
stable-16cu-workload-milvus-querynode-0-77899cd587-5xd4v     1/1  Running  0  4h31m
stable-16cu-workload-milvus-querynode-0-77899cd587-82rpj     1/1  Running  0  4h31m
stable-16cu-workload-milvus-querynode-0-77899cd587-r9pbk     1/1  Running  0  4h31m
stable-16cu-workload-milvus-querynode-0-77899cd587-zxlkw     1/1  Running  0  4h31m
[...]

PR #51718, opened 2026-07-22
fix: [3.0] inject QueryNode-local index load params on segment Reopen

foxspy:
## Summary

The segment Reopen path handed the `SegmentLoadInfo` to segcore without injecting the QueryNode-local index load parameters that the normal load path injects. For a DISKANN index the required `num_load_thread` (derived from local QueryNode resources, not persisted in index metadata) was therefore missing, and segcore aborted index loading with `param num_load_thread is empty` (VectorDiskIndex). Reopen tasks failed and were recreated indefinitely, adding sustained scheduler / index-loading / logging pressure while the new DISKANN index never loaded on the affected segments.

This extracts the per-index load-parameter preparation from `segmentLoader.Load` into a shared `prepareIndexLoadParams` helper and calls it from `LocalSegment.Reopen` as well, so the full-load and Reopen paths inject the same parameters. A unit test reproduces the missing parameter (fails before the fix, passes after).

issue: #51249
pr: #51716

### DIFF
--- a/internal/querynodev2/segments/segment.go
+++ b/internal/querynodev2/segments/segment.go
@@ -1303,6 +1303,13 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
 	}
 	defer s.ptrLock.Unpin()
 
+	// Reopen forwards the SegmentLoadInfo straight to segcore, so it must inject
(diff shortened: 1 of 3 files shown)


liliu-z (commented):
/lgtm
/approve

Issue #51723, opened 2026-07-22
datacoord: two manifest-commit sites miss the base==current CAS, and L0 commits object storage under the meta lock


xiaofan-luan:
## Background

#51376 fixed the schema-bump in-place manifest commit, which could silently drop a concurrently-built index: DataCoord adopted a compaction result by a *version-newer* check only, without verifying it was built on the segment's **current** manifest. The fix added an adoption-layer `base == current` CAS (mirroring the stats path's `errStatsResultStale`).

An audit of the remaining StorageV3 manifest-commit sites found the same invariant missing at two more places, plus an unrelated locking issue at a third.

## 1. Sort / shared stats manifest update — missing base CAS

`internal/datacoord/task_stats.go` `SetJobInfo`: the shared "update manifest version" branch (the sort sub-job and other non-text/JSON sub-jobs) adopts the manifest via a blind `UpdateManifest(segID, manifest)`, unlike the text/JSON branches which go through `updateStatsResultIfManifestMatches` (which carries the base CAS). This is safe today only because the sort target is a birth-exclusive, freshly-allocated segment — but the asymmetry is a latent lost-update if that target ever became a live segment with concurrent manifest writers.

## 2. External-collection refresh column patch — missing base CAS

`internal/datacoord/task_refresh_external_collection.go` `applyExternalRefreshPatch` adopts an in-place column patch (`AppendSegmentManifestColumns`, OVERWRITE, built on the segment's dispatch-time manifest) with a blind `cloned.ManifestPath = incoming.GetManifestPath()` — no check that the patch was built on the current manifest. A concurrent external-JSON index build that advanced the manifest in between is dropped. It self-heals (the patch clears the text/JSON stats placeholders, so the stats inspector rebuilds), but it opens a query-degradation window and wastes a rebuild — and, more importantly, it is the same lost-update class as #51376 at an uncovered site.

## 3. L0 delete compaction — object-storage commit under the meta lock

[...]

PR #51724, opened 2026-07-22
fix: base==current CAS for the sort-stats and external-refresh manifest adoptions

xiaofan-luan:
## What / why

The same StorageV3 segment manifest is advanced concurrently by several producers — an external-collection refresh column patch, a sort-stats result, and a text/JSON index build. They adopted a result by a *version-newer* check only, without verifying it was built on the segment's **current** manifest, so a later write could silently overwrite a concurrent commit (lost update). See #51723 for the audit.

This PR adds the `base == current` CAS at those adoption sites, and — because a CAS that only *detects* a conflict is not usable on its own (the previous behaviour either silently completed with missing data, or failed the whole job) — the recovery machinery to rebuild safely on the current manifest, plus the fencing needed to keep re-dispatch correct.

## Changes

**1. `base == current` CAS at the two adoption sites** (`task_stats.go`, `task_refresh_external_collection.go`, `task_update.go`, new `SegmentInfo.base_manifest`)
The worker records the manifest each result was built on (`base_manifest`); the coordinator adopts only when it still equals the segment's current manifest. The refresh CAS runs **inside** the `UpdateSegmentsInfo` / `segMu` critical section (in the upsert operator, via the synchronized `modPack.Get`) so the decision is atomic with the patch.

**2. Adopt only a legal *successor*, not just a matching base** (shared `validateManifestSuccessor`, `meta.go`)
`base == current` alone is not enough: a buggy / mixed-version / corrupt worker could carry the right base yet a result that points at another segment's manifest or an older version, silently corrupting the segment pointer. The result must be an idempotent replay (`result == current`) or a strictly-forward, same-base-path, parseable successor (`packed.CompareManifestPath`). This is the check the schema-bump adoption already did; it is extracted into one primitive and used by both so the paths cannot drift.

**3. Refresh: rebuild on conflict instead of silently completing / failing**
On a stale-manifest conflict the job-level apply aborts atomically and the checker resets the job's finished tasks to Init, so the worker rebuilds the patch on the current manifest (rather than keeping the segment as-is and reporting the refresh finished with columns still missing). A concurrent aggregator that observes a mid-retry task no-ops (`errExternalRefreshNotReady`) instead of failing the job.

**4. Classify refresh task failures — retry the transient ones**
Previously any task failure failed the whole refresh job. Now request/data errors (collection gone, invariant violations) fail; transient failures (RPC, allocation, worker object-store / manifest I/O, cancellation) drop the worker-side task and reset it for re-dispatch, mirroring the stats path. `ResetTaskForRetry` clears state/progress/result atomically. The DataNode manager reports `Retry` (not `Failed`) for those so DataCoord re-dispatches. Permanence is decoupled from the merr Input/System blame classification via an explicit `errExternalRefreshPermanent` marker.

**5. Fence worker attempts by version (ABA)**
Re-dispatch reuses the same taskID, so a stale/late Drop or result-write from a superseded attempt could clobber the re-dispatched one. `task_version` is carried through Create/Query/Drop; the DataNode registers each attempt under it, supersedes older attempts, and drops writes/`DeleteIfVersion` from a stale version; DataCoord fences its meta writes by the attempt version too. The version lives on the persisted task record (etcd), so it is monotonic across a DataCoord restart.

**6. A task the worker no longer tracks re-dispatches, not fails**
When DataCoord queries a task it believes is in flight but the DataNode has lost it (typically a DataNode restart drops the in-memory task map), the worker reports `Retry` so DataCoord re-runs it on a live node instead of failing the refresh job over a transient loss.

## Compatibility

- **Sort / shared index stats** adoption **fails open** on an empty base — a birth commit (freshly allocated sort target with no manifest yet) or an older DataNode that cannot report a base. This is not a regression: before this PR the stats path adopted blindly for everyone; new DataNodes are now protected (they set a base), and a fully-upgraded cluster is fully protected. base-fencing is enforced only where the worker does set a base.
- **External-collection refresh** adoption **fails closed** on an empty base (rejects). It is a manual, low-frequency operation that is not run during a rolling upgrade, so it has no old-worker compatibility need and takes the stronger guarantee on an existing segment.

## Not in this PR (deferred)

- **L0 "move the object-store commit off the meta lock"** — the in-lock commit is correct; moving it off-lock re-introduces a lost-update TOCTOU unless the in-lock apply re-validates `base == current` and retries. A performance optimization, not a correctness fix; lands separately. Tracked in #51723.
- **milvus-table deltalog refresh function-output rebuild** — a separate correctness concern in the deltalog path (the rebuilt manifest drops target-local function-output column groups the fake binlogs still claim), unrelated to the manifest CAS; handled on its own.

## Tests

- `task_stats_test.go`: `TestSetJobInfoSortResultManifestHandling` (stale→reject / fresh→adopt / baseless→adopt / birth→adopt / replay→no-op).
- `task_refresh_external_collection_test.go`: `TestApplyExternalCollectionSegmentUpdate_StalePatchAborts` (stale & empty base → abort+rebuild, matching → patched); CreateTaskOnWorker / QueryTaskOnWorker classification (transient → re-dispatch, permanent → fail); version-fenced re-dispatch.
- `meta_test.go`: `TestValidateManifestSuccessor` (replay / forward / empty / stale / rollback / cross-segment / unparsable).
- `external_collection_refresh_meta_test.go`: version-fenced writes (stale attempt dropped, current lands, v0 unconditional).
- `manager_test.go`: version fence reproduces the ABA (a superseded attempt's late result is dropped), `DeleteIfVersion` stale-drop fence, transient→Retry / ParameterInvalid→Failed classification.
- `services_test.go`: a task the worker no longer tracks reports `Retry`.

`data_coord.pb.go`'s large diff is the deterministic `[]byte` rawDesc re-wrap from inserting fields (regenerated with the repo's `cmake_build/bin/protoc`; regenerating the unchanged proto yields a 0-line diff).

Relates to #51376. Audit: #51723.

https://claude.ai/code/session_01SFhVdnFbWiAuEco1q5txtV


xiaofan-luan:
Thanks — all three were real and are addressed in `e0c0c2e`.

**P1 #1 (external CAS was dead code).** Correct: the old version wired the CAS into `refreshExternalCollectionTask.SetJobInfo`, which has no production caller — the real path is `applyFinishedJobSegments` → `applyExternalCollectionSegmentUpdate`. Reworked so the base is **persisted on the result segment** (new `SegmentInfo.base_manifest`, set by the worker in `patchSegmentForMissingColumns`); it rides the existing `task.updated_segments` persistence to the real job-level adoption, so the CAS runs on the actual path and survives restarts.

**P1 #2 (stale would hard-fail the job).** Correct: `UpdateJobStateWithPreApply` persists `JobStateFailed` on any pre-apply error. So a stale patch is no longer returned as an error — it is **reclassified as "kept"**: the segment keeps its current manifest (preserving the concurrent commit), is neither dropped nor patched, and the next refresh's missing-column detection re-materializes the columns. No `JobStateFailed`, no re-dispatch machinery.

**P2 #3 (L0 commits serialized).** Correct: restored the bounded pool. `saveSegmentMeta` resolves base+entries sequentially (cheap, in-memory) then runs the object-store commits through `conc.NewPool(L0ManifestUpdatePoolSize)` — the same bound the in-lock path used — off the meta lock.

Note: the large `data_coord.pb.go` diff is the deterministic `[]byte` rawDesc re-wrap from inserting a field (regenerated with the repo's `cmake_build/bin/protoc`; regenerating the *unchanged* proto yields a 0-line diff, so it matches CI).

xiaofan-luan:
Thanks again — all three are correct. Second revision (`e4434ef`):

**P1-B (external CAS was TOCTOU).** Fixed. The CAS now runs *inside* the `UpdateSegmentsInfo` / `segMu` critical section — in the upsert operator via the synchronized `modPack.Get`, not the pre-lock `mt.segments.GetSegment`. The base==current decision is now atomic with the patch: a stale patch is kept-as-is (the segment is in neither `segmentsToDrop` nor the patch set), preserving the concurrent commit.

**P1-A (L0 off-lock overwrite) + P2 (partial success not cached).** Both correct, and both only exist because I moved the L0 commit off-lock. I've **dropped that change** from this PR. The original in-lock `commitL0ManifestUpdates` is correct — the base read, the `OverwriteResolver` commit and the pointer move are atomic under `segMu`, so `current == base` at apply time and there is no lost update. Committing off-lock re-introduces exactly the TOCTOU you describe (an index adopts M8 between the off-lock M7 read and the in-lock apply, then the version-newer check adopts M9-built-from-M7) unless the in-lock apply re-validates `current == base` and retries, and the pool caches partial successes. That is a performance change (lock-across-S3), not a correctness fix, so it is deferred to a separate PR with the proper in-lock `{base, committed}` CAS + retry.

So this PR is now the two adoption-layer CAS fixes only (sort/shared stats + external refresh), both on the real path and inside the lock. #51723 still tracks the L0 lock-across-S3 item for that follow-up.

internal/datanode/external/task_update.go
```diff
@@ -638,6 +638,10 @@ func (t *RefreshExternalCollectionTask) patchSegmentForMissingColumns(
 
 	patched := proto.Clone(seg).(*datapb.SegmentInfo)
 	patched.ManifestPath = newManifestPath
+	// Record the manifest this column patch was built on so DataCoord can reject
+	// (keep-as-is) the patch if the segment's manifest has advanced since — a
+	// concurrent index build committed in between — instead of dropping it.
+	patched.BaseManifest = seg.GetManifestPath()
```
weiliu1031: [P1] `seg` is not always the live segment from the refresh request here. When the deltalogs also need refreshing, `organizeSegments` first creates an uncommitted intermediate manifest M1 and passes that segment into this method. This records M1 as `BaseManifest`, although DataCoord still holds the original manifest M0. The adoption CAS therefore rejects the final result even without a concurrent writer, and every retry repeats the same sequence. Please capture the original live manifest before the first rewrite and preserve it through all subsequent manifest transformations.

weiliu1031 (commented):
Three blocking findings and one non-blocking follow-up are attached inline.

internal/datacoord/task_refresh_external_collection.go
```diff
@@ -415,6 +426,28 @@ func applyExternalCollectionSegmentUpdate(
...
+				// advanced since — a concurrent text/JSON index build committed and
+				// was adopted in between — adopting the patch would drop that
+				// commit. Abort the whole apply atomically (modPack.fail sets
+				// updatePack.err, so UpdateSegmentsInfo returns before persisting
+				// anything) and signal errExternalRefreshStaleManifest: the refresh
+				// checker resets the job's tasks to Init and the worker rebuilds the
+				// patch on the current manifest, instead of silently completing with
+				// a segment still missing the refreshed columns. An empty base
+				// (older worker) skips the CAS, preserving prior behavior.
+				if base := incoming.GetBaseManifest(); base != "" && existing.GetManifestPath() != base {
```
weiliu1031: [P1] An empty base is not necessarily a birth result on this path. `refreshMilvusTableSegmentManifest` can rewrite the manifest of an existing segment for a deltalog-only refresh without setting `BaseManifest`. If another stats/index task advances the live manifest while that result is running, this condition skips the CAS and adopts the stale deltalog result over the concurrent commit. Please carry the original manifest as the base for every same-ID manifest rewrite; only genuinely new segments should have no base.

internal/datacoord/external_collection_refresh_checker.go
```diff
@@ -254,6 +255,14 @@ func (c *externalCollectionRefreshChecker) aggregateJobState(job *datapb.Externa
 				func(latestJob *datapb.ExternalCollectionRefreshJob) error {
 					return c.applyJobInfo(c.ctx, latestJob)
 				})
+			if errors.Is(err, errExternalRefreshStaleManifest) {
+				// The apply raced a concurrent manifest commit. Do not finish or
+				// fail the job; rebuild its results on the current manifest by
+				// resetting its tasks. The inspector re-dispatches them and the
+				// per-job timeout bounds the loop if the conflict never clears.
+				c.retryTasksOnStaleManifest(job.GetJobId(), err)
```
weiliu1031: [P1] The task reset happens after `UpdateJobStateWithPreApply` has released the job lock. Another completion callback may already have aggregated the job as `Finished`, then enter pre-apply after this path has reset one task to `Init`. `applyFinishedJobSegments` returns a normal non-finished-task error, which `UpdateJobStateWithPreApply` persists as a terminal `Failed` job. Please establish retry ownership/generation under the same job critical section, or revalidate the aggregate generation and treat callbacks from the previous generation as no-ops.

internal/datacoord/external_collection_refresh_checker.go
```diff
@@ -330,6 +339,35 @@ func (c *externalCollectionRefreshChecker) aggregateJobState(job *datapb.Externa
...
+func (c *externalCollectionRefreshChecker) retryTasksOnStaleManifest(jobID int64, cause error) {
+	mlog.Warn(c.ctx, "external refresh apply raced a concurrent manifest commit; resetting tasks to rebuild on the current manifest",
+		mlog.FieldJobID(jobID),
+		mlog.Err(cause))
+	tasks := c.refreshMeta.GetTasksByJobID(jobID)
+	for _, task := range tasks {
+		if task.GetState() != indexpb.JobState_JobStateFinished {
+			continue
+		}
+		if err := c.refreshMeta.UpdateTaskState(task.GetTaskId(),
```
weiliu1031: [P3] Resetting through `UpdateTaskState` leaves `Progress=100`, `ResultReady=true`, and the previous result payload intact. The rebuild still executes correctly, but progress APIs report `Pending` or `InProgress` with 100% until the new result arrives. A dedicated retry-reset operation should clear progress and the previous result fields atomically.

internal/datacoord/task_refresh_external_collection.go
liliu-z: The milvus-table deltalog-only path rewrites an existing segment manifest without setting `BaseManifest`, so this condition skips CAS and can overwrite a newer stats/index manifest. Please carry the original dispatch-time manifest for every same-ID rewrite.

internal/datacoord/task_refresh_external_collection.go
```diff
@@ -415,6 +426,28 @@ func applyExternalCollectionSegmentUpdate(
...
+				// decision is atomic with the patch. The worker built this column
+				// patch on incoming.BaseManifest; if the segment's manifest has
+				// advanced since — a concurrent text/JSON index build committed and
+				// was adopted in between — adopting the patch would drop that
+				// commit. Abort the whole apply atomically (modPack.fail sets
+				// updatePack.err, so UpdateSegmentsInfo returns before persisting
+				// anything) and signal errExternalRefreshStaleManifest: the refresh
+				// checker resets the job's tasks to Init and the worker rebuilds the
+				// patch on the current manifest, instead of silently completing with
+				// a segment still missing the refreshed columns. An empty base
```
liliu-z: Empty base is fail-open for an existing segment. That also leaves rolling upgrades or any future producer that forgets the field unprotected. Existing-segment adoption should require a non-empty base; only new segment birth should allow empty.

internal/datanode/external/task_update.go
liliu-z: This can record an unadopted intermediate manifest as the base. With deltalog refresh plus missing columns, the pipeline is M1 -> M2 -> M3, but DataCoord still has M1 and this returns base=M2, so CAS fails even with no concurrent writer. Keep the original M1 as the adoption base through the whole pipeline.

internal/datacoord/external_collection_refresh_checker.go
liliu-z: The stale retry transition happens after the job lock is released. Two concurrent job checks can both aggregate Finished; one resets a task to Init, then the other enters pre-apply, sees a non-finished task, and persists the job as Failed. The per-job critical section needs to cover aggregate -> apply -> stale reset.

internal/datacoord/external_collection_refresh_checker.go
liliu-z: Resetting the same task ID does not create a fenced new attempt. DataNode deduplicates by `(clusterID, taskID)`, while the task Version is not sent in Create/Query/Drop. Retry can replay the cached finished result and depends on Drop succeeding. Please key/fence worker attempts by version, or allocate a new task ID.

internal/datanode/external/task_update.go
```diff
@@ -638,6 +638,10 @@ func (t *RefreshExternalCollectionTask) patchSegmentForMissingColumns(
 
 	patched := proto.Clone(seg).(*datapb.SegmentInfo)
 	patched.ManifestPath = newManifestPath
+	// Record the manifest this column patch was built on so DataCoord can reject
```
liliu-z: CAS alone does not make the milvus-table deltalog path safe. It rebuilds the target manifest from the source snapshot, which does not include target-local function-output column groups, while the cloned fake binlogs can still claim those fields exist. Please update deltalogs on the current target manifest, or fully rebuild the target-local outputs too.

liliu-z (commented):
A few blocking correctness issues below.

liliu-z:
One broader suggestion: this should be closed at one shared DataCoord manifest-adoption primitive instead of adding job-specific CAS checks in stats, refresh, compaction, etc.

All existing-segment writers eventually publish `SegmentInfo.ManifestPath`, so their different result protos can be normalized to something like:

`{segmentID, expectedBase, resultManifest, mutateOtherFields}`

Then enforce under `segMu`:

- `current == result` => idempotent success
- existing-segment update requires a non-empty base
- `current != expectedBase` => stale
- result must stay on the same base path and move forward
- apply job-specific fields and persist atomically

New-segment birth should use a separate path. It would also be good to use `FailResolver` at the storage writer for existing-manifest commits where possible; writer CAS and DataCoord adoption CAS protect different gaps.

internal/datacoord/external_collection_refresh_checker.go
```diff
@@ -330,6 +347,37 @@ func (c *externalCollectionRefreshChecker) aggregateJobState(job *datapb.Externa
...
+		mlog.FieldJobID(jobID),
+		mlog.Err(cause))
+	tasks := c.refreshMeta.GetTasksByJobID(jobID)
+	for _, task := range tasks {
+		if task.GetState() != indexpb.JobState_JobStateFinished {
+			continue
+		}
+		// Fence to the task's current attempt version (freshly read here), so a
+		// concurrent supersede does not turn this into a stale reset.
+		if _, err := c.refreshMeta.ResetTaskForRetry(task.GetTaskId(), task.GetVersion(),
```
weiliu1031, line 384: nit: Could this reset also verify that the owning job is still non-terminal under the same critical section? `UpdateJobStateWithPreApply` releases the job lock before `resetJobTasksForRetry` runs, so a concurrent timeout can mark the job Failed before this version-only reset moves a Finished task back to Init. The inspector may then enqueue work for an already-terminal job. An atomic job-state/task-state/version check, with an additional terminal-job guard before dispatch, would keep the states consistent.

internal/datanode/external/manager.go
```diff
@@ -168,19 +191,52 @@ func (m *ExternalCollectionManager) Delete(clusterID string, taskID int64) *Task
...
 }
 
-// UpdateState updates only the state/failReason fields.
-func (m *ExternalCollectionManager) UpdateState(clusterID string, taskID int64, state indexpb.JobState, failReason string) {
+// DeleteIfVersion cancels and removes the task entry only when its version
+// matches (version 0 forces removal, for legacy drops that carry no version).
+// A drop issued for a superseded attempt (older version) is ignored so it cannot
+// delete the entry of the re-dispatched attempt now occupying the same taskID.
+// Returns the removed entry, or nil if nothing was removed.
+func (m *ExternalCollectionManager) DeleteIfVersion(clusterID string, taskID int64, version int64) *TaskInfo {
```
weiliu1031, line 187: nit: Now that the production Drop path uses `DeleteIfVersion`, `Delete` and `CancelTask` appear to be test-only. Could we remove those legacy entry points and update the tests to exercise the version-aware lifecycle API, so future callers cannot bypass the attempt fence? `UpdateState` is also test-only, and `mutateTaskApplied` is a single-call forwarding helper, so they could be included in the same cleanup if desired.

weiliu1031 (commented):
Two non-blocking nits.

PR #51725, opened 2026-07-23
fix: validate MATCH predicates are element-level

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/46517

### DIFF
--- a/internal/core/src/exec/expression/MatchExpr.cpp
+++ b/internal/core/src/exec/expression/MatchExpr.cpp
@@ -301,6 +301,10 @@ PhyMatchFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                "Match result should be ColumnVector");
     AssertInfo(match_result_col_vec->IsBitmap(),
                "Match result should be bitmap");
+    AssertInfo(match_result_col_vec->size() == elem_count,
+               "Match predicate result size {} should equal element count {}",
+               match_result_col_vec->size(),
+               elem_count);
     TargetBitmapView match_result_bitset_view(
         match_result_col_vec->GetRawData(), match_result_col_vec->size());
     TargetBitmapView match_result_valid_view(
--- a/internal/parser/planparserv2/parser_visitor.go
+++ b/internal/parser/planparserv2/parser_visitor.go
@@ -1245,7 +1245,7 @@ func validateMatchPredicateElementLevel(expr *planpb.Expr) (bool, error) {
 
 	switch realExpr := expr.GetExpr().(type) {
 	case *planpb.Expr_ColumnExpr:
-		return validateMatchColumnInfo(realExpr.ColumnExpr.GetInfo())
+		return false, merr.WrapErrParameterInvalidMsg("raw element columns are not supported inside MATCH predicate")
 	case *planpb.Expr_TermExpr:
 		return validateMatchColumnInfo(realExpr.TermExpr.GetColumnInfo())
 	case *planpb.Expr_UnaryRangeExpr:
@@ -1257,7 +1257,7 @@ func validateMatchPredicateElementLevel(expr *planpb.Expr) (bool, error) {
 	case *planpb.Expr_CompareExpr:
 		return false, merr.WrapErrParameterInvalidMsg("column-to-column comparison is not supported inside MATCH predicate")
 	case *planpb.Expr_JsonContainsExpr:
-		return validateMatchColumnInfo(realExpr.JsonContainsExpr.GetColumnInfo())
+		return false, merr.WrapErrParameterInvalidMsg("array/JSON contains expressions are not supported inside MATCH predicate")
 	case *planpb.Expr_NullExpr:
 		return validateMatchColumnInfo(realExpr.NullExpr.GetColumnInfo())
 	case *planpb.Expr_ExistsExpr:
@@ -1289,15 +1289,7 @@ func validateMatchPredicateElementLevel(expr *planpb.Expr) (bool, error) {
 		}
 		return leftElementLevel && rightElementLevel, nil
 	case *planpb.Expr_CallExpr:
-		hasElementLevel := false
-		for _, param := range realExpr.CallExpr.GetFunctionParameters() {
-			paramElementLevel, err := validateMatchPredicateElementLevel(param)
-			if err != nil {
-				return false, err
-			}
-			hasElementLevel = hasElementLevel || paramElementLevel
-		}
-		return hasElementLevel, nil
+		return false, merr.WrapErrParameterInvalidMsg("function calls are not supported inside MATCH predicate")
 	case *planpb.Expr_ValueExpr, *planpb.Expr_AlwaysTrueExpr:
 		return false, nil
 	case *planpb.Expr_ElementFilterExpr:
--- a/internal/parser/planparserv2/parser_visitor.go
+++ b/internal/parser/planparserv2/parser_visitor.go
@@ -1228,6 +1228,89 @@ func isValidStructSubField(tokenText string) bool {
 	return len(tokenText) >= 4 && tokenText[:2] == "$[" && tokenText[len(tokenText)-1] == ']'
 }
 
+func validateMatchColumnInfo(columnInfo *planpb.ColumnInfo) (bool, error) {
+	if columnInfo == nil {
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate column info is missing")
+	}
+	if !columnInfo.GetIsElementLevel() {
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate can only use element-level fields")
+	}
+	return true, nil
+}
+
+func validateMatchPredicateElementLevel(expr *planpb.Expr) (bool, error) {
+	if expr == nil {
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate is missing")
+	}
+
+	switch realExpr := expr.GetExpr().(type) {
+	case *planpb.Expr_ColumnExpr:
+		return validateMatchColumnInfo(realExpr.ColumnExpr.GetInfo())
+	case *planpb.Expr_TermExpr:
+		return validateMatchColumnInfo(realExpr.TermExpr.GetColumnInfo())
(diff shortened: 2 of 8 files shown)


internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1228,6 +1228,97 @@ func isValidStructSubField(tokenText string) bool {
...
+	case *planpb.Expr_BinaryExpr:
+		leftElementLevel, err := validateMatchPredicateElementLevel(realExpr.BinaryExpr.GetLeft())
+		if err != nil {
+			return false, err
+		}
+		rightElementLevel, err := validateMatchPredicateElementLevel(realExpr.BinaryExpr.GetRight())
+		if err != nil {
+			return false, err
+		}
+		return leftElementLevel || rightElementLevel, nil
```
buqian-zilliz: **[P1] Require both logical branches to preserve element cardinality**

`leftElementLevel || rightElementLevel` admits a logical predicate when one branch is only row-sized. For example, `MATCH_ANY(s, ($[b] == 2) && (true || $[a] == 1))` leaves the right branch as `AlwaysTrueExpr`, so validation succeeds through the left branch. In sequential MATCH evaluation the comparison returns `elem_count` bits but `AlwaysTrueExpr` returns `batch_rows`; `ThreeValuedLogicOp::And` sizes both views from the first vector, which can read past the shorter bitmap before the final size assertion runs. Require both branches to be element-cardinality-aware, or normalize constant branches to `elem_count` before combining them.

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1228,6 +1228,97 @@ func isValidStructSubField(tokenText string) bool {
...
+		return validateMatchColumnInfo(realExpr.ColumnExpr.GetInfo())
+	case *planpb.Expr_TermExpr:
+		return validateMatchColumnInfo(realExpr.TermExpr.GetColumnInfo())
+	case *planpb.Expr_UnaryRangeExpr:
+		return validateMatchColumnInfo(realExpr.UnaryRangeExpr.GetColumnInfo())
+	case *planpb.Expr_BinaryRangeExpr:
+		return validateMatchColumnInfo(realExpr.BinaryRangeExpr.GetColumnInfo())
+	case *planpb.Expr_BinaryArithOpEvalRangeExpr:
+		return validateMatchColumnInfo(realExpr.BinaryArithOpEvalRangeExpr.GetColumnInfo())
+	case *planpb.Expr_CompareExpr:
```
buqian-zilliz, line 1257: **[P1] Reject CompareExpr until element-level execution exists**

This case accepts `MATCH_ANY(sa, $[a] == $[b])` because both columns are marked element-level. C++ then discards that mode in `ParseCompareExprs`, constructs the physical comparison with the subfields' storage type `ARRAY`, and `PhyCompareFilterExpr` has no `ARRAY` dispatch, so the query passes this validation and later fails with `DataTypeInvalid`. Reject `CompareExpr` here until its physical node preserves element offsets and scalar element types, or implement that execution path.

internal/core/src/exec/expression/MatchExpr.cpp
```diff
@@ -301,6 +301,10 @@ PhyMatchFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                "Match result should be ColumnVector");
     AssertInfo(match_result_col_vec->IsBitmap(),
                "Match result should be bitmap");
+    AssertInfo(match_result_col_vec->size() == elem_count,
```
buqian-zilliz, line 304: **[P2] Fix overflow shortcuts before enforcing element count**

The new size invariant is violated by valid element-level range predicates that take an overflow shortcut. For example, `MATCH_ANY(struct_array, $[int8_field] == 128)` is accepted, but in sequential mode `PhyUnaryRangeFilterExpr::PreCheckOverflow` allocates a row-sized result from `batch_size_` and `active_count_`, while this code expects the flattened `elem_count`. Any batch that does not contain exactly one element per row now throws here instead of returning no matches; the binary-range overflow shortcut has the same problem. Make those shortcuts use element-level batch cardinality and validity before enforcing this assertion.

internal/parser/planparserv2/parser_visitor.go
buqian-zilliz, line 1: **[P1] Preserve int64 precision in mixed JSON IN lists**

When a JSON `IN` list mixes integer and floating literals, this conversion narrows every integer to `float64` without checking the exact-integer limit. For example, `json["n"] in [9007199254740993, 0.5]` rewrites the integer to `9007199254740992`; because C++ now sees only float values, it bypasses the existing unsafe-int precise path and can return false positives for adjacent int64 values. Preserve the integer tag and use the precise numeric path, or explicitly handle or reject unsafe mixed lists.

buqian-zilliz, line 1: **[P1] Normalize mixed JSON template lists too**

The normalization only runs in the literal-list branch. For a template RHS, `FillTermExpressionValue` later copies mixed JSON values unchanged, while C++ chooses the term executor from the first value. A template list `[1, 1.5]` selects the int64 path and asserts when the float is inserted, whereas `[1.5, 1]` takes the double path, making valid template queries order-dependent. Move this normalization into a shared helper and apply it after template filling as well.

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1228,6 +1228,89 @@ func isValidStructSubField(tokenText string) bool {
...
+	return true, nil
+}
+
+func validateMatchPredicateElementLevel(expr *planpb.Expr) (bool, error) {
+	if expr == nil {
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate is missing")
+	}
+
+	switch realExpr := expr.GetExpr().(type) {
+	case *planpb.Expr_ColumnExpr:
```
buqian-zilliz, line 1247: **[P1] Reject raw element columns and calls until C++ supports them**

`IsElementLevel` does not make a raw column physically element-typed. `$[flag]` is serialized with storage type `ARRAY`, so `MATCH_ANY(sa, $[flag])` passes this validator and then fails C++'s boolean type check; through the `CallExpr` case, `empty($[name])` similarly resolves as `empty(ARRAY)` instead of the registered `empty(VARCHAR)`. Reject raw columns and calls inside MATCH for now, or make the C++ column and call paths expose the element type and flattened cardinality before accepting them.

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1228,6 +1228,89 @@ func isValidStructSubField(tokenText string) bool {
...
+		return validateMatchColumnInfo(realExpr.TermExpr.GetColumnInfo())
+	case *planpb.Expr_UnaryRangeExpr:
+		return validateMatchColumnInfo(realExpr.UnaryRangeExpr.GetColumnInfo())
+	case *planpb.Expr_BinaryRangeExpr:
+		return validateMatchColumnInfo(realExpr.BinaryRangeExpr.GetColumnInfo())
+	case *planpb.Expr_BinaryArithOpEvalRangeExpr:
+		return validateMatchColumnInfo(realExpr.BinaryArithOpEvalRangeExpr.GetColumnInfo())
+	case *planpb.Expr_CompareExpr:
+		return false, merr.WrapErrParameterInvalidMsg("column-to-column comparison is not supported inside MATCH predicate")
+	case *planpb.Expr_JsonContainsExpr:
```
buqian-zilliz, line 1259: **[P1] Reject JsonContains until it returns element-sized bitmaps**

`JsonContainsExpr` is accepted solely because its column has `IsElementLevel`, but `PhyJsonContainsFilterExpr` still uses row batch sizing and row offsets. `MATCH_ANY(sa, array_contains($[tag], "x"))` therefore returns one child bit per row in sequential mode and hits the new `elem_count` assertion for variable-length structs; offset mode can interpret flattened element IDs as row IDs. Reject this node inside MATCH until its executor is element-aware, or lower it to an element comparison.

internal/parser/planparserv2/parser_visitor.go
SpadeA-Tang: It's not relevant with this PR.

SpadeA-Tang: It's not relevant with this PR.

internal/parser/planparserv2/parser_visitor.go
SpadeA-Tang: I have banned this.

internal/parser/planparserv2/parser_visitor.go
SpadeA-Tang: ditto

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1228,6 +1228,81 @@ func isValidStructSubField(tokenText string) bool {
...
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate is missing")
+	}
+
+	switch realExpr := expr.GetExpr().(type) {
+	case *planpb.Expr_ColumnExpr:
+		return false, merr.WrapErrParameterInvalidMsg("raw element columns are not supported inside MATCH predicate")
+	case *planpb.Expr_TermExpr:
+		return validateMatchColumnInfo(realExpr.TermExpr.GetColumnInfo())
+	case *planpb.Expr_UnaryRangeExpr:
+		return validateMatchColumnInfo(realExpr.UnaryRangeExpr.GetColumnInfo())
```
buqian-zilliz, line 1252: **[P2] Reject malformed template ranges in MATCH validation**

Checking only the column flag lets template arithmetic bypass the new unsupported-comparison guard. For `MATCH_ANY(sa, {k} + $[a] < $[b])`, the template-marked left `BinaryArithExpr` is mis-lowered into a `UnaryRangeExpr` on `$[b]` with neither a concrete value nor a template name, and with `IsTemplate` unset. This branch accepts that malformed node, so `FillExpressionValue` skips it and C++ later asserts while decoding the unset `GenericValue`. Reject range nodes lacking a value/template name here, or fix the comparison lowering to reject this unsupported multi-column shape.

zhengbuqian (approved):

Issue #51736, opened 2026-07-23
[Bug]: ARRAY element full scan stops on zero-element row batches before later elements


liliu-z:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: current master (`b7cbb91a161c1830bd0a54e024c48cb8dc8f468d`)
- Deployment mode: reproducible in both sealed and growing segment execution paths
- Frequency: deterministic

### Current Behavior

During an element-level ARRAY full scan, a row batch may contain rows but contribute zero logical elements when every ARRAY in that batch is NULL or empty.

`GetNextBatchSizeForElementLevel()` can therefore return a positive row count with an element count of zero. The expression visitors currently treat the zero element count as exhaustion and return `nullptr` before their data or index processing path advances the row cursor.

If later rows still contain elements, `ElementFilterBitsNode` has a positive segment-wide element count but receives a null expression result and fails its non-null assertion.

With the default 8,192-row expression batch, 8,192 NULL or empty ARRAY rows followed by a non-empty ARRAY row deterministically reproduce the problem.

### Expected Behavior

A valid row batch containing zero logical elements must be distinguished from true exhaustion. Evaluation should return a valid zero-length result, advance exactly one corresponding row batch, and continue until it either produces elements or consumes all rows.

Compound expressions must keep every child on the same row batch while doing this.

### Steps To Reproduce

1. Create a nullable `ARRAY<INT32>` field used by an element filter.
2. Set `EXEC_EVAL_EXPR_BATCH_SIZE` to 4.
3. Insert 9 rows: rows 0 through 7 are NULL or empty arrays, and row 8 is `[7, 9]`.
4. Run a full-mode compound element predicate such as `$[price_array] >= 0 AND $[price_array] == 9` with all documents active.
5. Observe that evaluation stops on the first zero-element row batch instead of returning row 8, element index 1.

The same failure occurs on sealed and growing segments.

### Anything else?

- A leaf must not internally collapse multiple zero-element row batches: AND short-circuiting advances skipped children once per evaluation, so collapsing batches would desynchronize child cursors.
- Unary and binary-range integer-overflow fast paths also need element-sized results; otherwise they bypass the corrected batch contract.
- This case was explicitly excluded from the scope of #51540.
- It was identified while reviewing the 3.0 backport #51717: https://github.com/milvus-io/milvus/pull/51717#discussion_r3632208484
- Proposed master fix: #51737

PR #51737, opened 2026-07-23
fix: advance element filters across zero-element batches

liliu-z:
issue: #51736

Follow-up to #51540.

Found during review of #51717: https://github.com/milvus-io/milvus/pull/51717#discussion_r3632208484

## What changed

- Make `GetNextRealBatchSize()` distinguish true exhaustion from a non-empty row batch containing zero logical ARRAY elements.
- Let unary range, binary range, term, and binary arithmetic element visitors return a valid zero-length bitmap and advance exactly one row batch.
- Make unary and binary-range integer-overflow fast paths use element counts and the same cursor contract.
- Add sealed and growing regressions with one full NULL-array batch, one full valid-empty-array batch, and later matching elements. Compound AND/OR predicates verify skipped children stay aligned.

## Why

A complete row batch can contain rows but contribute zero logical elements. Treating its element count of zero as EOF returns `nullptr` before the row cursor advances. Full-mode `ElementFilterBitsNode` still has later segment elements to evaluate and fails its non-null expression-result assertion.

The fix preserves one evaluation per row batch: zero-element batches produce zero-length results instead of being collapsed inside a leaf, so compound-expression cursor movement remains synchronized.

## Scope

This fixes the batch contract for the scalar element visitors that use `GetNextRealBatchSize()` (unary range, binary range, term, and binary arithmetic). Other element-expression families are unchanged.

## Verification

- `git diff --check upstream/master...HEAD` — passed.
- `clang-format-15 --dry-run --Werror` on all changed C++ files — passed.
- Audited all 9 production callers of `GetNextRealBatchSize()` and both integer-overflow fast paths.
- Added deterministic sealed/growing regression coverage for two consecutive zero-element row batches, compound cursor alignment, and unary/binary-range overflow paths.
- Local full C++ build was not completed because the macOS arm64 Conan profile has no compatible binaries and would rebuild the full third-party dependency graph; target-branch CI is pending for compile and test execution.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -532,15 +532,28 @@ class SegmentExpr : public Expr {
                    : batch_size_;
     }
 
-    int64_t
+    std::optional<int64_t>
     GetNextRealBatchSize(const OffsetVector* input, bool element_level) {
         if (input != nullptr) {
+            if (input->empty()) {
+                return std::nullopt;
+            }
             return input->size();
         } else if (element_level) {
-            auto [_, elem_count] = GetNextBatchSizeForElementLevel();
+            auto [batch_rows, elem_count] = GetNextBatchSizeForElementLevel();
+            // A non-empty row batch can legitimately contain no logical
+            // elements. Keep 0 distinct from exhaustion so callers still
+            // advance their row cursor for this batch.
+            if (batch_rows == 0) {
+                return std::nullopt;
+            }
             return elem_count;
         }
-        return GetNextBatchSize();
+        auto batch_size = GetNextBatchSize();
+        if (batch_size == 0) {
+            return std::nullopt;
+        }
+        return batch_size;
     }
 
     // Get the next batch size for element-level processing
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -414,27 +414,43 @@ class SegmentExpr : public Expr {
         return true;
     }
 
+    int64_t
+    DataCursorRow(size_t chunk, int64_t chunk_pos) const {
+        return segment_->is_chunked()
+                   ? segment_->num_rows_until_chunk(field_id_, chunk) +
+                         chunk_pos
+                   : static_cast<int64_t>(chunk) * size_per_chunk_ + chunk_pos;
+    }
+
     void
     MoveCursorForDataMultipleChunk() {
         int64_t processed_size = 0;
         for (size_t i = current_data_chunk_; i < num_data_chunk_; i++) {
             auto data_pos =
                 (i == current_data_chunk_) ? current_data_chunk_pos_ : 0;
+            const auto active_remaining =
+                active_count_ - DataCursorRow(i, data_pos);
+            if (active_remaining <= 0) {
+                break;
+            }
             // if segment is chunked, type won't be growing
             int64_t size = segment_->chunk_size(field_id_, i) - data_pos;
 
-            size = std::min(size, batch_size_ - processed_size);
+            size = std::min(
+                {size, batch_size_ - processed_size, active_remaining});
+            if (size <= 0) {
+                continue;
+            }
 
             processed_size += size;
+            current_data_chunk_ = i;
+            current_data_chunk_pos_ = data_pos + size;
             if (processed_size >= batch_size_) {
-                current_data_chunk_ = i;
-                current_data_chunk_pos_ = data_pos + size;
-                current_data_global_pos_ =
-                    current_data_global_pos_ + processed_size;
                 break;
             }
-            // }
         }
+        current_data_global_pos_ =
+            DataCursorRow(current_data_chunk_, current_data_chunk_pos_);
     }
     // Non-chunked segments are always Growing (Sealed is always chunked).
     void
@@ -443,22 +459,31 @@ class SegmentExpr : public Expr {
         for (size_t i = current_data_chunk_; i < num_data_chunk_; i++) {
             auto data_pos =
                 (i == current_data_chunk_) ? current_data_chunk_pos_ : 0;
+            const auto active_remaining =
+                active_count_ - DataCursorRow(i, data_pos);
+            if (active_remaining <= 0) {
+                break;
+            }
             auto size = (i == (num_data_chunk_ - 1) &&
                          active_count_ % size_per_chunk_ != 0)
                             ? active_count_ % size_per_chunk_ - data_pos
                             : size_per_chunk_ - data_pos;
 
-            size = std::min(size, batch_size_ - processed_size);
+            size = std::min(
+                {size, batch_size_ - processed_size, active_remaining});
+            if (size <= 0) {
+                continue;
+            }
 
             processed_size += size;
+            current_data_chunk_ = i;
+            current_data_chunk_pos_ = data_pos + size;
             if (processed_size >= batch_size_) {
-                current_data_chunk_ = i;
-                current_data_chunk_pos_ = data_pos + size;
-                current_data_global_pos_ =
-                    current_data_global_pos_ + processed_size;
                 break;
             }
         }
+        current_data_global_pos_ =
+            DataCursorRow(current_data_chunk_, current_data_chunk_pos_);
     }
 
     void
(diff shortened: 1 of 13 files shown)


internal/core/src/exec/expression/Expr.h
buqian-zilliz, line 1: **[P2] Zero-element batches still scan raw array rows**

Returning an engaged `0` correctly distinguishes this batch from exhaustion, but each raw-data caller still enters `ProcessDataChunksForElementLevel()`. That helper prefetches or opens array chunks and walks every row even though the offsets lookup has already proved that the batch contains no logical elements. Long sparse gaps therefore add O(number of empty rows) raw-data work before the next element, and chunk-level skip indexes cannot eliminate this when a chunk also contains later values. Add a shared zero-element fast path that advances exactly one row batch and returns an empty `ColumnVector` without touching array payloads.

SpadeA-Tang:
except for the comment
rest /lgtm

internal/core/src/exec/expression/Expr.h
```diff
@@ -414,27 +414,43 @@ class SegmentExpr : public Expr {
         return true;
     }
 
+    int64_t
+    DataCursorRow(size_t chunk, int64_t chunk_pos) const {
```
zhengbuqian: bad name, consider rename to "RowOffsetInSegment", or "OffsetInSegment"?

internal/core/src/exec/expression/Expr.h
```diff
@@ -414,27 +414,43 @@ class SegmentExpr : public Expr {
...
+        return segment_->is_chunked()
+                   ? segment_->num_rows_until_chunk(field_id_, chunk) +
+                         chunk_pos
+                   : static_cast<int64_t>(chunk) * size_per_chunk_ + chunk_pos;
+    }
+
     void
     MoveCursorForDataMultipleChunk() {
```
zhengbuqian: for code cleaness: method almost identical to MoveCursorForDataSingleChunk, the only difference is how size is computed. we can use just move the logic into MoveCursorForData and use segment_->is_chunked() to decide how to compute size. so we can get rid of these 2 helpers.

internal/core/src/exec/expression/Expr.h
```diff
@@ -2916,20 +2979,104 @@ class SegmentExpr : public Expr {
...
+        // state. A leading zero-element batch must not touch ARRAY payloads.
+        auto array_offsets = segment_->GetArrayOffsets(field_id_);
+        AssertInfo(array_offsets != nullptr,
+                   "ArrayOffsets not found for field {}",
+                   field_id_.get());
+        const auto initial_rows = std::min(batch_size_, active_count_);
+        const auto elem_start = array_offsets->ElementIDRangeOfRow(0).first;
+        const auto elem_end =
+            array_offsets->ElementIDRangeOfRow(initial_rows).first;
+        return elem_end > elem_start;
```
buqian-zilliz, line 2981: **[P3] Deferred raw-data prefetch only inspects the leading batch, so later empty ARRAY chunks are still prefetched**

`ShouldPrefetchRawDataEagerly()` decides eager vs deferred prefetch using only rows `[0, min(batch_size_, active_count_))`. When that first batch contains at least one element, the expression-specific `PrefetchRawData()` runs eagerly and iterates every chunk from `0`; the per-chunk SkipIndex decision is keyed on scalar value ranges and does not reliably prune chunks whose rows are all empty/NULL. `AdvanceEmptyElementBatch()` then avoids *reading* later zero-element row batches during evaluation, but the prefetch I/O for those empty chunks has already been issued. For sparse ARRAY fields where the empty rows are not in the first batch, this retains the original eager-prefetch cost. This is a follow-up optimization rather than a correctness regression: consider deferring prefetch until the first non-empty batch for the current cursor, or extending the heuristic beyond the leading batch.

zhengbuqian (approved):

internal/core/src/exec/expression/Expr.h
liliu-z: Renamed to `RowOffsetInSegment` in e2b6631. It reads no cursor state — it just converts `(chunk, chunk_pos)` to a segment-global row offset — so the old name was misleading.

internal/core/src/exec/expression/Expr.h
liliu-z: Merged in e2b6631, and the duplication collapsed a bit further than expected.

The `active_remaining` clamp this PR added makes the single-chunk trailing-chunk special case redundant: for non-chunked segments `num_data_chunk_ == upper_div(active_count_, size_per_chunk_)`, so `min(size_per_chunk_, active_count_ - chunk * size_per_chunk_)` reproduces `active_count_ % size_per_chunk_ - data_pos` exactly on the last chunk. So the per-chunk row count is now a single helper:

```cpp
int64_t
ChunkRowCount(size_t chunk) const {
    if (segment_->is_chunked()) {
        return segment_->chunk_size(field_id_, chunk);
    }
    return std::min(size_per_chunk_,
                    active_count_ - static_cast<int64_t>(chunk) * size_per_chunk_);
}
```

and the two loops became one `MoveCursorForData()`. `ComputeRawDataPrefetchStartChunk()` had a third copy of the same ternary and now calls the helper too. Net -34 lines.

Equivalence was checked exhaustively rather than by eye: brute-forced old vs new for both changed formulas over `size_per_chunk_` 1..64 x `active_count_` 0..512 x every valid `(chunk, data_pos)` — 9,566,706 cases, 0 mismatches. The chunked branch is byte-identical (`segment_->chunk_size()` both before and after).

internal/core/src/exec/expression/Expr.h
liliu-z: Done — this is implemented as `AdvanceEmptyElementBatch()` in `Expr.h`. It returns a zero-length `ColumnVector`, advances exactly one row batch via `MoveCursor()`, and never touches ARRAY payloads:

```cpp
VectorPtr
AdvanceEmptyElementBatch(const OffsetVector* input, bool element_level, int64_t result_count) {
    if (!element_level || input != nullptr || result_count != 0) {
        return nullptr;
    }
    MoveCursor();
    return std::make_shared<ColumnVector>(TargetBitmap(0, false), TargetBitmap(0, true));
}
```

It is called immediately after `GetNextRealBatchSize()` and before any payload work — including before `PreCheckOverflow()` — in all element-level leaves: `UnaryExpr.cpp:1608,1772`, `TermExpr.cpp:938,997,1038`, `BinaryRangeExpr.cpp:427,503`, `BinaryArithOpEvalRangeExpr.cpp:1375,2292`. So `ProcessDataChunksForElementLevel()` is never entered for a zero-element batch.

The async raw-data prefetch needed a separate guard since it runs ahead of evaluation; that is `ShouldPrefetchRawDataEagerly()` / `EnsureRawDataPrefetched()`.

internal/core/src/exec/expression/Expr.h
liliu-z: The description is accurate — `ShouldPrefetchRawDataEagerly()` only inspects `[0, min(batch_size_, active_count_))`, and if that batch has any element the eager `PrefetchRawData()` covers every chunk from the current start chunk. Not taking it in this PR, for two reasons.

1. Not a regression, as noted. master prefetches eagerly and unconditionally, so this path is no worse than before; the deferral only ever removes I/O.

2. The suggested direction works against the design. Prefetch exists to issue I/O *ahead of* evaluation. Deciding it from the live cursor is exactly what the current code deliberately avoids — the prefetch worker races with compound-expression cursor moves, which is why the heuristic reads immutable initial-batch geometry instead. Extending it past the leading batch means either scanning the whole `ArrayOffsets` at prefetch time or re-deciding per batch; the first pays a scan to maybe save a fetch, the second serializes prefetch behind evaluation.

Separately, this PR is a correctness fix for the zero-element batch cursor contract; the eager/deferred prefetch machinery is already the widest part of its scope. Further prefetch tuning for sparse ARRAY fields belongs in its own perf PR with a benchmark, so tracking it there rather than here.

One thing checked while reading this: `ElementIDRangeOfRow(initial_rows)` with `initial_rows == active_count_` is in bounds — both `ArrayOffsetsSealed` (`ArrayOffsets.cpp:197`) and `ArrayOffsetsGrowing` (`:565`) have an explicit `row_id == row_count` branch.

Issue #51384, opened 2026-07-15
[Bug]: Driver rethrows operator exceptions as ExecOperatorException, destroying segcore error codes


xiaofan-luan:
## Is there an existing issue for this?
- [x] I have searched the existing issues

## Environment
master

## Current Behavior
`CALL_OPERATOR` in exec/Driver.cpp catches every operator exception and rethrows it as `ExecOperatorException`, discarding the original `SegcoreError` code. Any carefully classified error raised inside expression evaluation (e.g. `ExprInvalid` for a user-fixable plan/deployment problem) reaches the CGO boundary as a generic UnexpectedError -- terminal client errors become retriable-looking internal errors for ALL expression failures, on every query path.

## Expected Behavior
The driver wrapper preserves and forwards the SegcoreError code (ExecOperatorException carries it, or SegcoreError is rethrown as-is), and the CGO boundary translation keeps it, so error classification done at throw sites survives to the client.

## Anything else?
Found while classifying MATCH_* errors: even after fixing throw sites to ExprInvalid, the code cannot survive the driver today. Cross-cutting, deserves its own PR.

PR #51739, opened 2026-07-23
fix: [cp3.0] preserve segcore error codes through the exec driver

xiaofan-luan:
issue: #51384
pr: #51389

## What
Backport the complete patch series from #51389 to the 3.0 branch. This preserves classified SegcoreError codes through the exec driver, handles cancellation consistently, corrects storage error classification, and keeps internal expression/operator invariants classified as system errors.

## Validation
- `git range-diff` confirms all 11 backported commits are patch-equivalent to the source PR.
- Cherry-pick completed without conflicts.
- Local C++ build was started but intentionally stopped before completion at the requester’s direction; CI is the remaining validation.


liliu-z (commented):
/lgtm
/approve

Issue #51741, opened 2026-07-23
[Bug]: ARRAY-backed field-to-field comparisons are accepted by the parser but fail in QueryNode


liliu-z:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Environment

- Milvus version: master
- Components: `internal/parser/planparserv2`, QueryNode `CompareExpr`

### Current Behavior

The parser accepts field-to-field comparisons when either operand is backed by an ARRAY field, for example:

```text
array_a == array_b
array_a != array_b
array_a == scalar_a
array_a[0] == scalar_a
```

It emits a `CompareExpr`, but that expression only carries field IDs and storage data types. It does not carry an ARRAY element path, and the QueryNode `CompareExpr` executor does not support `DataType::ARRAY` operands. The request therefore passes query-plan construction and fails later during execution with an internal segcore error.

This behavior predates #51389. That PR preserves the error code selected downstream; it does not introduce the unsupported expression.

### Expected Behavior

Until ARRAY-backed field-to-field execution is implemented, reject these expressions during query-plan construction with a clear QueryPlan input error.

Keep the supported ARRAY forms unchanged, including:

```text
array_a == [1, 2]
[1, 2] == array_a
array_a[0] == 1
array_a[0] < 10
```

### Steps To Reproduce

1. Create a collection with an `ARRAY<INT64>` field and another comparable field.
2. Insert and load data.
3. Query or search with `array_a == array_a` or `array_a[0] == scalar_a`.
4. Observe that parsing succeeds but QueryNode execution fails on the ARRAY storage type.

### Anything else?

Related error-propagation PR: #51389.

liliu-z:
Fix PR: #51742.

liliu-z:
3.0 backport PR: #51746.

PR #51742, opened 2026-07-23
fix: reject unsupported ARRAY-backed field comparisons

liliu-z:
issue: #51741
related: #51389

## What

- Reject field-to-field comparisons when either operand is backed by an ARRAY field before constructing `CompareExpr`.
- Return an `ErrQueryPlan` input error instead of allowing the unsupported plan to fail later in QueryNode.
- Keep ARRAY-to-literal and ARRAY-element-to-literal comparisons unchanged.
- Add regression coverage for full ARRAY operands, ARRAY element paths, both operand positions, equality/inequality/relational operators, and the preserved literal forms.

## Why

`planparserv2` currently accepts these expressions because their scalar or element types are comparable. However, `CompareExpr` only carries field IDs and storage data types: it does not carry an ARRAY element path, and the C++ executor has no `DataType::ARRAY` operand path. As a result, the request passes planning and fails later with an internal segcore error.

This is a pre-existing parser/executor contract gap. #51389 only preserves the downstream error code; it does not introduce the unsupported expression. Rejecting the request at plan construction keeps that error-propagation change from exposing this planner bug as a newly visible system error.

## Preserved behavior

These forms remain valid:

```text
ArrayField == [1, 2, 3, 4]
[1, 2, 3, 4] == ArrayField
ArrayField != [1, 2, 3, 4]
ArrayField[0] == 1
```

## Verification

```text
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2/...
```

The full `planparserv2` package suite passes locally. No live Milvus e2e run was performed for this parser-only change.

### DIFF
--- a/internal/parser/planparserv2/plan_parser_v2_test.go
+++ b/internal/parser/planparserv2/plan_parser_v2_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/milvus-io/milvus/internal/util/function/rerank"
 	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/proto/planpb"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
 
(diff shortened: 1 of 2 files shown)


zhengbuqian (approved):
/lgtm

liliu-z:
3.0 backport: #51746. It is marked as a preemptive backport and should not merge before this master PR.

Issue #42937, opened 2025-06-25
[Enhancement]: Update Knowhere Version Track


foxspy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

This issue is mainly used to record the upgrade record of the knowhere version upgrade in milvus

PR #51748, opened 2026-07-23
fix: [3.0] revert targetVecIndexVersion bump and set to 8

foxspy:
## Summary

Revert #50238, which set `dataCoord.targetVecIndexVersion` to `10`, and set it to `8` instead.

issue: #42937


liliu-z (commented):
/lgtm
/approve

PR #51270, opened 2026-07-11
fix: keep delegator deleteMut free of bulk delete replay and batch-pin ts column in delete apply

tinswzy:
issue: #49435

On streaming standalone, concurrent Strong `count(*)` + upsert intermittently fails with `code=505 channel tsafe stalled`.

### Root cause (measured with per-step instrumentation on the reproducing benchmark)

1. `loadStreamDelete` held the delegator's `deleteMut.RLock` across the whole post-load delete replay (forward → `LoadDeltaData` → segcore apply). Under replay amplification — a freshly compacted segment replays the entire buffered delete history, and batches of ~100K delete pks matched 50–100M rows (multiple delete versions per pk × multiple physical row versions) — a single apply took 20–50s, starving `ProcessDelete`, the single writer that also advances tsafe. The channel froze and every Strong query on it failed; on standalone (single replica) the retries land on the same delegator, so the failure is user-visible.
2. Inside the apply, `search_batch_pks` resolved every matched row's insert timestamp through `ReadTimestamp()` — a per-row cachinglayer pin — measured at ~60% of the call.

### Fixes

1. **Delegator (chase-then-lock)**: the bulk snapshot flush and iterative catch-up rounds now run lock-free; only the final trickle catch-up + `addDistribution` remain under `deleteMut.RLock`. The original atomicity invariant (no delete can fall between "deletes applied" and "segment visible") is unchanged — the locked section is structurally identical to before — while its payload drops from the whole replay to typically zero-to-a-few-hundred rows. Includes a regression test that blocks `worker.Delete` during load and asserts `ProcessDelete` is not starved (fails on the old code with a 50s+ stall).
2. **Segcore**: batch-pin the timestamp column once per `search_batch_pks` call from the same published snapshot and read raw pointers in the loop. Pins are refcounts on the snapshot's cells, so segment-reopen's atomic publication is not blocked; per-call snapshot consistency is identical to the per-row variant.

### Verification

2 rounds × 4 concurrent instances of the reproducing benchmark (sift-2M, partition-key collection, Strong `count(*)` + upsert, 8C/16Gi standalone) — all 8 runs passed:

| metric | before (4 runs) | after (8 runs) |
|---|---|---|
| `channel tsafe stalled` | ≥1000 lines per pod | **0** |
| client-visible errors | present | **0** (incl. `count(*)==2,000,000` checks) |
| `LoadDeletedRecord` avg | ~10.9s (peaks 28–54s) | 1.5–2.1s |
[...]

PR #51750, opened 2026-07-23
fix: [3.0] keep delegator deleteMut free of bulk delete replay and batch-pin ts column in delete apply

tinswzy:
pr: #51270


czs007 (approved):
/lgtm

Issue #51062, opened 2026-07-06
[Bug]: BM25 search sees VARCHAR placeholder for sparse function field after schema evolution


lyyyuna:
### Describe the bug

During schema evolution stability testing, a BM25 search failed after repeatedly adding/dropping fields and function fields.

The collection schema had a BM25 sparse function output field `sparse_bm25` whose field type was `VECTOR_SPARSE_U32_F32`, but the search request was rejected because the server parsed the search placeholder type as `VARCHAR`.

This looks like a transient schema/function metadata inconsistency in the search path after function field schema evolution. There was no QueryNode crash in this run.

### Error

```text
2026-07-04 12:05:43,347 [ERROR][_log_rpc_error]: RPC error: [search],
<MilvusException: (code=2000, message=failed to search:
failed to search/query delegator 4 for channel by-dev-rootcoord-dml_0_467446420583745078v0:
fail to search on QueryNode 4:
worker(4) query failed:
parser searchRequest failed:
Assert "check_data_type(field_meta, ph, element.element_level_)"
=> vector type must be the same,
field sparse_bm25 - type VECTOR_SPARSE_U32_F32,
search ph type VARCHAR
at ../internal/core/src/query/Plan.cpp:163:
segcore error[segcoreCode=2001])>, <elapsed:9149.8ms>
```

### Environment

- Milvus image: `harbor.tencent-stage.zilliz.cn/milvusdb/milvus:master-20260704-33f93c2a-amd64`
- Test repo commit: `727a2ebcb3285442bccc8062f54d32fffc97dd8c`
- Workflow: `schema-evolution-l3-7f9b5`
- vcluster: `vc-yiyang-li`
- Namespace: `argo-schema-evolution-l3-7f9b5`
- Argo test pod: `schema-evolution-l3-7f9b5-test-4117748329`
- QueryNode pod: `schema-evolution-milvus-querynode-844577d876-t8hqh`
- QueryNode worker id in error: `QueryNode 4`
- Error time: `2026-07-04 12:05:43 UTC`

### QueryNode logs around the failure

QueryNode did not print the `Plan.cpp:163` assertion directly. The assertion was returned to the client as the PyMilvus RPC error above.

Around the same time, QueryNode was loading/releasing sealed segments for the same collection:

```text
2026-07-04 12:05:41.807 [INFO] [querynodev2/services.go:467]
"received load segments request" collectionID=467446420583745078 partitionID=467446420583745079
[...]

PR #51753, opened 2026-07-23
fix: reconcile BM25 IDF oracle on the load path when a load races(#51062)

MrPresent-Han:
issue: #51062

### DIFF
--- a/internal/querynodev2/delegator/delegator.go
+++ b/internal/querynodev2/delegator/delegator.go
@@ -228,6 +228,41 @@ func (sd *shardDelegator) publishIDFOracle(idfOracle IDFOracle) {
 	sd.idfOracle.Store(&idfOracleHolder{oracle: idfOracle})
 }
 
+// reconcileFunctionRuntimeLocked reconciles the delegator's function runtime —
+// the per-vchannel function runner and the BM25 IDF oracle — to the collection's
+// current schema. The caller MUST hold schemaChangeMutex.
+//
+// Both artifacts are driven only by UpdateSchema (the stream/DDL path); the load
+// path never touches them. A segment load can advance the collection schema ahead
+// of the DDL and gate that UpdateSchema out as a no-op (#51062), so the no-op
+// branch calls this to still refresh the runner and create/sync the oracle. It
+// reconciles against sd.collection.Schema() (monotonic:
+// prepareCollectionSchemaUpdate rejects older versions), so a stale schema can never
+// roll either back; the oracle is never torn down and is published once, so a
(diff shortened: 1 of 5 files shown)


internal/querynodev2/delegator/delegator.go
```diff
@@ -1277,11 +1306,15 @@ func (sd *shardDelegator) UpdateSchema(ctx context.Context, schema *schemapb.Col
...
+		// Collection schema is already current (e.g. a segment load advanced it ahead
+		// of this DDL), so the schema/worker apply below is a no-op. But the delegator
+		// IDF oracle is driven only by the stream, so still reconcile it here to the
+		// collection's current schema (#51062).
+		mlog.Info(ctx, "delegator collection schema current, reconcile IDF oracle only",
 			mlog.Uint64("schemaVersion", schemaVersion),
 			mlog.Uint64("schemaBarrierTs", schemaBarrierTs),
 		)
-		return nil
+		return sd.reconcileIDFOracleLocked()
```
liliu-z: The no-op branch reconciles only the IDF oracle, not the per-vchannel BM25 function runner that the full-apply branch refreshes via function.GetManager().Update (delegator.go:1382); on the load path that runner is refreshed only by syncCollectionIndexMeta when the load carries index info (delegator_data.go:594-615) and never by ProcessInsert. Because sd.collection is the shared node collection (delegator.go:1511), a co-located worker load can advance the schema to the add-field version via PutOrRef (services.go:539) without going through this delegator's syncCollectionIndexMeta, so when the DDL then lands here the runner is left at the old version and a search for the new BM25 output field fails in prepareSearchFunction -> RunWithRunner with a retriable 'field is not available in function runner schema' error (not a silent raw-VARCHAR-to-segcore path), persisting on a growing-only channel that never receives a later indexed load/reopen. Also call function.GetManager().Update(sd.collectionID, delegatorFunctionRunnerKey(sd.vchannelName), sd.collection.Schema()) in this branch so the runner advances together with the oracle.

MrPresent-Han: Good catch — adopted. Instead of a separate call in the gated branch, the runner refresh is folded into the shared reconcile path: `function.GetManager().Update` moved from the full-apply branch into `reconcileFunctionRuntimeLocked` (renamed from `reconcileIDFOracleLocked`), so both branches reconcile the runner and the oracle together from the monotonic collection schema, keeping the same panic-on-invalid-metadata semantics. Added `TestUpdateSchemaGatedBranchReconcilesFunctionRuntime` covering this scenario end-to-end: collection pre-advanced by a load, gated event arrives, then the runner serves the new output field and the oracle exists.

Issue #51544, opened 2026-07-17
[Bug]: [Nightly] Text LOB bulk import setup fails because pytest container has no kubeconfig


NicoYuan1986:
## Environment

- **Milvus Version**:
  - master Nightly, commit `f662cec` / `master-20260716-f662cec16d`, DAILY-CI-TRIGGER build 299
  - 3.0 Nightly, commit `3f52db0`, DAILY-CI-TRIGGER build 299
- **Deployment Mode**: cluster
- **MQ**: pulsar / woodpecker / woodpecker-service
- **SDK**: Python client Nightly test framework, pymilvus `2.7.0rc46` observed in master build 299 logs
- **OS**: Nightly CI Kubernetes environment

### Option B: Steps

1. Run DAILY-CI-TRIGGER build 299:
   - https://jenkins-milvus-ci.milvus.io/blue/organizations/jenkins/MILVUS-CI-V2-PR-PIPELINES%2FDAILY-CI-TRIGGER/detail/DAILY-CI-TRIGGER/299/pipeline
2. Check the master 4AM dispatcher:
   - https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-dispatcher/14735/
3. Check the 3.0 4AM dispatcher:
   - https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-dispatcher/14739/
4. Observe the bulk import Text LOB case failing during pytest fixture setup:
   - `bulk_insert/test_bulk_insert_api.py::TestBulkInsert::test_text_lob_bulk_import_json_csv_parquet[json]`
   - `bulk_insert/test_bulk_insert_api.py::TestBulkInsert::test_text_lob_bulk_import_json_csv_parquet[csv]`
   - `bulk_insert/test_bulk_insert_api.py::TestBulkInsert::test_text_lob_bulk_import_json_csv_parquet[parquet]`

### Trigger Conditions

- **Frequency**: observed in all checked deployment modes for this case in DAILY-CI-TRIGGER build 299.
- **Scope**: master and 3.0 Nightly.
- **Deployment modes observed**:
  - `distributed-pulsar-mmap`
  - `distributed-woodpecker`
  - `distributed-woodpecker-service`
- **Failed downstream builds**:
  - master dispatcher #14735:
    - `distributed-pulsar-mmap`: build 34378
[...]

PR #51755, opened 2026-07-23
test: update client cases for storage and sdk changes

NicoYuan1986:
issue: #51544

## What changed

- Derive the default Python client test namespace from `MILVUS_NS` or `MILVUS_HELM_NAMESPACE` before falling back to `chaos-testing`.
- Add an optional external-source MinIO endpoint override for external table tests so the upload endpoint and server-side `external_source` endpoint can be configured independently.
- Classify newly added proto-only `DataType` values `Date`, `Time`, and `Decimal` in the external table add-field matrix.
- Update the multi-primary-key invalid schema test to expect validation when the second primary key is added to the schema.
- Extend TEXT LOB layout validation to accept current Vortex segment data files while keeping the existing parquet layout path.

## Why

- Nightly test scripts already provide the Helm namespace through environment variables, but some fixtures still fell back to a hard-coded namespace.
- External table tests need to distinguish the endpoint used by the Python test client from the endpoint Milvus uses to read `external_source` data.
- The add-field matrix should stay explicit as new proto data types are introduced.
- Newer PyMilvus validates duplicate primary keys earlier, before collection creation.
- Current Storage V3 deployments may persist TEXT LOB segment data as Vortex files instead of parquet files.

## Verification

- `test_text_lob_bulk_import_json_csv_parquet`
- `test_match_query_parent_output_with_diskann_index`
- `test_text_lob_inline_vs_lob_object_layout`
- `test_milvus_client_collection_invalid_schema_multi_pk`
- `test_milvus_client_external_table_add_all_supported_field_types`
- `test_milvus_client_xgboost_supports_all_numeric_feature_types`


tests/python_client/conftest.py
```diff
@@ -34,7 +34,12 @@ def pytest_addoption(parser):
     parser.addoption("--password", action="store", default="Milvus", help="password for connection")
     parser.addoption("--db_name", action="store", default="default", help="database name for connection")
     parser.addoption("--secure", action="store", default=False, help="secure for connection")
-    parser.addoption("--milvus_ns", action="store", default="chaos-testing", help="milvus_ns")
+    parser.addoption(
+        "--milvus_ns",
+        action="store",
+        default=os.getenv("MILVUS_NS", os.getenv("MILVUS_HELM_NAMESPACE", "chaos-testing")),
```
yanliang567, line 40: [P1] This does not address the failure reported in #51544. `TestcaseBaseBulkInsert.init_minio_client` still calls `get_milvus_instance_name` / `get_pod_ip_name_pairs`, both of which initialize the Kubernetes client; with `IN_CLUSTER` unset they still call `load_kube_config()` and fail before the namespace value can help. Nightly already passes `--minio_host`, so please consume `minio_host`/`minio_bucket` in that fixture (or set `IN_CLUSTER=true`) and verify the original failing Nightly case.

tests/python_client/milvus_client/test_milvus_client_text_lob.py
```diff
@@ -691,48 +691,86 @@ def text_reference_values(values):
...
+    lob_file_ids = []
+    for key in lob_keys:
+        name = os.path.basename(key)
+        if name.endswith(".vx"):
+            lob_file_ids.append(uuid.UUID(name.removesuffix(".vx")))
+    assert lob_file_ids, f"expected Vortex LOB objects under {lob_prefix}; data={vortex_keys}, lobs={lob_keys}"
+
+    log.info(f"TEXT LOB vortex layout: data={vortex_keys}, lobs={lob_keys}")
+    refs = {}
+    lob_file_id = lob_file_ids[0]
```
yanliang567: [P2] The Vortex branch fabricates refs from the expected inputs instead of reading persisted segment data. It classifies rows using `inline_threshold`, assigns every large row the first `.vx` UUID and offset 0, and the test then asserts those synthesized values. A regression that stores row 3100 as a LOB, or uses the wrong file/offset, can still pass. Please decode the reference column from the Vortex segment/manifest (or use a storage reader that exposes the actual refs) before asserting the layout.

tests/python_client/common/external_table_common.py
```diff
@@ -42,8 +42,10 @@ def _minio_address(minio_host):
 
 
 def get_minio_config(minio_host=None, minio_bucket=None):
+    address = _minio_address(minio_host)
     return {
-        "address": _minio_address(minio_host),
+        "address": address,
+        "external_source_address": _minio_address(os.getenv("MILVUS_EXTERNAL_SOURCE_MINIO_HOST") or address),
```
yanliang567: [P2] The external-source address is now independent, but its TLS setting is not: `build_external_spec()` still derives `use_ssl` from `cfg["secure"]`, which is based on the upload-side `minio_host`. If uploads use HTTP and `MILVUS_EXTERNAL_SOURCE_MINIO_HOST=https://...` (or vice versa), Milvus connects with the wrong protocol. Please track an external-source-specific secure flag from the override and use it in the extfs spec.

yanliang567 (commented):
Found three issues during review; details are in the inline comments.

czs007:
**Verification: this PR fixes #51544.**

We ran the failing test from #51544 on a bare-metal standalone instance, once on `master` and once on this PR's head:

```
test:                    bulk_insert/test_bulk_insert_api.py::TestBulkInsert::test_text_lob_bulk_import_json_csv_parquet[json]
master                   errored during setup
this PR (head):          passed
```

Because this PR only touches `tests/`, both runs used the same server binary on the same running instance — only the test files were swapped between the two runs, so the difference is attributable to this PR alone.
Wall time 43s for the test itself.

Caveat: this is a single test case, not a full CI run.

This PR looks stalled — last commit 40 days ago, and it currently has merge conflicts; CI is red (Ruff (changed files only)); no review yet. Since the change itself is verified to fix the issue, it seems worth rebasing so that it can land.

czs007:
duplicated with https://github.com/milvus-io/milvus/pull/53305

Issue #51756, opened 2026-07-23
[Enhancement]: JSON stats queries unnecessarily access shared BSON and may apply raw JSON null validity incorrectly


zhagnlu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

JSON stats build records a full layout for each JSON path/type, including whether data is stored in a typed shredding column or in shared BSON. However, during load, Milvus only restores the shredding column mapping and drops shared layout metadata.

  As a result, JSON stats query execution cannot tell whether an exact path/type is fully covered by shredding columns. The query path conservatively executes shared BSON fallback even when the target path/type already has a typed shredding column.

  This causes pure typed JSON stats queries to unnecessarily pin/load the shared BSON inverted index and potentially read/parse shared BSON payloads.

  There is also a related correctness risk in the unary JSON stats path: applying raw top-level JSON field null/validity can mask rows that should be determined by JSON stats shredding/shared validity instead.

  Examples

  For data where /a_STRING is stored as a shredding column:

  json["a"] == "foo"
  json["a"] in ["foo", "bar"]
  json["a"] > "foo"

  The result can be answered by the shredding column alone, but Milvus still executes shared BSON fallback.

  For numeric paths, both /a_INT64 and /a_DOUBLE need to be covered before shared fallback can be safely skipped. If only one numeric type is covered, shared fallback is still required because the other numeric type may be in shared BSON.

Impact

  - Extra query latency from unnecessary BSON index pin/load.
  - Extra memory or mmap pressure from shared BSON index access.
  - Extra CPU from BSON payload parsing.
  - Pure shredding-column queries can be affected by shared BSON index load failures, even though they should not need shared BSON.
  - Raw JSON field null validity may incorrectly affect JSON stats query results.

  Expected Behavior

  - Exact path/type JSON stats queries should skip shared BSON fallback when all target types are already covered by shredding columns.
  - Numeric JSON stats queries should skip shared fallback only when both INT64 and DOUBLE shredding columns are available.
  - EXISTS and prefix-style queries should remain conservative unless full shared layout metadata is available.
  - JSON stats query validity should come from stats data, not raw top-level JSON field validity.

  Fix Direction

  Use existing loaded shredding metadata to decide whether shared fallback is necessary, without loading the full shared layout into memory.

  Add a helper like:

  HasAllShreddingFields(pointer, target_types)

  Then skip ExecuteForSharedData() for exact path/type cases where shredding columns fully cover the query.

[...]

PR #51757, opened 2026-07-23
enhance: optimize JSON stats fallback and null handling

zhagnlu:
issue: #51756

### DIFF
--- a/internal/core/src/exec/expression/JsonContainsExpr.cpp
+++ b/internal/core/src/exec/expression/JsonContainsExpr.cpp
@@ -694,7 +694,8 @@ PhyJsonContainsFilterExpr::ExecJsonContainsByStats() {
                 }
             }
         };
-        {
+        if (!index->HasAllShreddingFields(pointer,
+                                          {milvus::index::JSONType::ARRAY})) {
             milvus::ScopedTimer timer(
                 "json_contains_stats_shared_data",
                 [this](double us) { json_stats_shared_latency_us_ += us; });
@@ -934,7 +935,8 @@ PhyJsonContainsFilterExpr::ExecJsonContainsArrayByStats() {
             }
             res_view[row_offset] = false;
         };
-        {
+        if (!index->HasAllShreddingFields(pointer,
+                                          {milvus::index::JSONType::ARRAY})) {
             milvus::ScopedTimer timer(
                 "json_contains_array_stats_shared_data",
                 [this](double us) { json_stats_shared_latency_us_ += us; });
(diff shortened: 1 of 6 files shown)


internal/core/src/exec/expression/JsonContainsExpr.cpp
```diff
@@ -1387,7 +1389,8 @@ PhyJsonContainsFilterExpr::ExecJsonContainsAllByStats() {
                 res_view[row_offset] = (remaining == 0);
             }
         };
-        {
+        if (!index->HasAllShreddingFields(pointer,
```
buqian-zilliz: **[P3] Avoid building the shared matcher on the shredded fast path**

The fully shredded path still constructs `shared_matcher` and `shared_found_large` before this guard. `ContainsAllMatcher` hashes every query term into a new map, and lists over 64 terms also allocate the auxiliary bitmap, even though `ExecuteForSharedData` is skipped. For large `json_contains_all` lists this leaves O(term-count) fallback-only setup on the intended fast path; compute the coverage flag first and construct the shared matcher/executor only inside the fallback branch.

internal/core/src/exec/expression/UnaryExpr.cpp
```diff
@@ -1448,18 +1423,28 @@ PhyUnaryRangeFilterExpr::ExecRangeVisitorImplJsonByStats() {
...
+                skip_shared_data = index->HasAllShreddingFields(
+                    pointer, {milvus::index::JSONType::BOOL});
+            } else if constexpr (std::is_same_v<GetType, proto::plan::Array>) {
+                skip_shared_data = index->HasAllShreddingFields(
+                    pointer, {milvus::index::JSONType::ARRAY});
+            }
         }
 
-        {
+        if (!skip_shared_data) {
```
buqian-zilliz, line 1331: **[P3] Defer fallback regex compilation until shared data is needed**

This guard is evaluated only after the shared-path `LikePatternMatcher`/`PartialRegexMatcher` has been constructed. On a fully shredded STRING path, a regex query still allocates and compiles a new RE2, and LIKE still parses its pattern, even though the shared executor never runs. Decide `skip_shared_data` before building the matcher/context and create those fallback-only objects inside this branch.

internal/core/src/exec/expression/JsonContainsExpr.cpp
zhagnlu: fixed

internal/core/src/exec/expression/UnaryExpr.cpp
zhagnlu: fixed

zhengbuqian (approved):

Issue #51727, opened 2026-07-23
[Bug]: DataCoord skips index creation for function-output fields on recovered StorageV3 segments


tedxu:
Follow-up from PR #50410 review by @congqixia.

After PR #50410, AlterSegments skips persisting per-field binlog KVs for
StorageV3 segments, so a recovered V3 SegmentInfo reloads with empty
Binlogs arrays (ManifestPath is authoritative).

internal/datacoord/index_inspector.go getSegmentBinlogFields (~:182)
derives field presence only from segment.GetBinlogs(). For a recovered
V3 segment, a function-output field that exists in the manifest but not
in the empty in-memory array is treated as missing, so
canCreateIndexForSegment (~:215) skips creating its index and does not
self-heal.

Fix direction: resolve function-output field presence from the manifest
(or Stats) for V3 rather than treating an empty Binlogs array as
evidence the field was not backfilled.

PR #50410, opened 2026-06-09
enhance: persist segment summary metadata on SegmentInfo

tedxu:
See #50406

DataCoord today derives aggregate segment metrics by iterating
SegmentInfo's FieldBinlog arrays on every scheduling decision.
For V3 segments the per-field binlog KV entries in etcd are pure
write amplification because the manifest is already the
authoritative path source.

This PR adds a Statistics message embedded on SegmentInfo so
DataCoord reads aggregates (sizes, counts, timestamp range,
null counts, quantiles) directly from a persisted field with no
branching on storage version at the read sites.

Wire formats

SaveBinlogPathsRequest.stats is populated by every flush. The
receiver consumes only stats_binlog_size, and only for V3
segments where stats live in the manifest. V2 derives the
cumulative footprint from the statslog FieldBinlog array.

CompactionSegment.stats is populated by every compactor at
write completion and copied verbatim onto the new SegmentInfo.
Insert and delta aggregates come from the FieldBinlog arrays
the compactor ships; stats blob footprint comes from a counter
the BinlogRecordWriter accumulates inside writeStats /
appendV3Stats and exposes via GetStatsBlobSize.

V3 etcd KV write skip

AlterSegments skips per-FieldBinlog KV writes for V3 segments;
the manifest is authoritative for paths and Stats carries the
aggregates DataCoord needs. Pre-existing V3 segments with binlog
KVs still load correctly via applyBinlogInfo.

Migration

Lazy. NewSegmentInfo back-fills Stats from arrays on first load
when the persisted field is nil. The back-fill lives in memory
until the next mutating operator persists the SegmentInfo proto.

Other notable changes

ShouldDoSingleCompaction reports expired_fraction one bucket
below the precise lower bound to prevent byte-size over-trigger
when binlog sizes vary.

EnsureStats is non-mutating to avoid a data race with
concurrent RLock readers; eager init happens at NewSegmentInfo
and under the meta.segMu write lock in mutating operators.

V3 L0 compaction outputs now carry TimestampFrom/To on the
resulting deltalog, matching the V1 path.

Test plan

- [x] Unit tests for storage.StatisticsCollector and
  BuildStatsFromFieldBinlogs
- [x] Unit tests for UpdateSegmentStats V2/V3 differentiation
- [x] Unit tests for ShouldDoSingleCompaction under-estimate
- [x] Unit tests for CompactionSegment.Stats V3 blob size flow
- [x] Unit tests for import L0 / non-L0 timestamp plumbing
[...]

PR #51768, opened 2026-07-23
fix: gate V3 function-output index build on segment schema version

tedxu:
issue: #51727

Recovered V3 segments load with empty Binlogs after PR #50410 (the
manifest is authoritative), so the index-eligibility gate in datacoord
rejected materialized function-output fields permanently and their
index never built or self-healed.

This derives index readiness from the segment schema version instead of
the binlog projection. A segment defers only when the collection has
functions and the segment schema version is behind the collection
version, meaning its function outputs are not yet materialized.

Adding a function to an existing collection requires StorageV3 with
bump-schema-version reconciliation enabled, and that reconciliation
advances the segment schema version together with materialization, so a
segment caught up to the collection schema version is guaranteed to have
its outputs materialized. The gate therefore never builds an empty
index, needs no manifest read, and survives recovery because the schema
version is a scalar on SegmentInfo that PR #50410 still persists.

getSegmentBinlogFields and the binlog fast path are removed.

### DIFF
--- a/internal/datacoord/backfill_result.go
+++ b/internal/datacoord/backfill_result.go
@@ -190,10 +190,9 @@ func buildV2Groups(bucket string, entry *BackfillSegment) (map[int64]*datapb.Fie
 		}
 		out[fid] = &datapb.FieldBinlog{
 			FieldID: fid,
-			// ChildFields carries the real field IDs that index creation
-			// (getSegmentBinlogFields) and backfill-compaction detection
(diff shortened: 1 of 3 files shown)


MrPresent-Han:
checking

MrPresent-Han:
/lgtm
conditionally lgtm, two preconditons must be fulfilled:
1. external collection support intact schema-evolution
2. external backfill can drop existing index on the target field when finished

Issue #51769, opened 2026-07-23
bug: MixCoord may release its shared session before child coordinators stop


xiaocai2333:
## Summary

On the 2.6 branch, RootCoord, DataCoord, and QueryCoord share the same etcd Session when running in MixCoord mode. During shutdown, QueryCoord stops first and calls Session.Stop, which revokes the shared MixCoord lease before DataCoord and RootCoord have finished stopping.

## Root cause

The shared Session is created and injected by mixCoordImpl, but every child coordinator treats it as an internally owned Session and stops it from its own Stop method.

The current shutdown sequence is:

1. mixCoordImpl.Stop calls QueryCoord.Stop.
2. QueryCoord.Stop calls the shared Session.Stop.
3. The MixCoord lease is revoked and another MixCoord can become active.
4. The old DataCoord may still be draining compaction and index work and persisting metadata.

This breaks the expected fencing invariant that the old MixCoord finishes all coordinator work before the new MixCoord becomes active.

## Impact

A new MixCoord can overlap with an old DataCoord that is still completing in-flight work. This can cause stale task recovery and leave segment index records in a non-terminal state without being scheduled.

This has the same shared-session lifecycle root cause as #48242. The corresponding master fix is #48252, but the 2.6 branch is still affected.

## Proposed fix

Use explicit Session ownership:

- A coordinator stops a Session only when it created that Session itself.
- Sessions injected by MixCoord are not stopped by child coordinators.
- mixCoordImpl stops the shared Session only after QueryCoord, DataCoord, and RootCoord have all stopped.

This issue is scoped to the 2.6 branch.

PR #48252, opened 2026-03-15
fix: remove IsTriggerKill SIGINT from datacoord and querycoord session watchers

xiaofan-luan:
## Summary

Removes the `IsTriggerKill` / SIGINT block from `datacoord.stopServiceWatch()` and `querycoordv2.watchNodes()`.

### Root Cause

In MixCoord mode all three coordinators share the same etcd `Session` object. During shutdown, when any coordinator calls `session.Stop()` it cancels the shared context, closing the other coordinators' etcd watches — triggering `stopServiceWatch()` / `watchNodes()` which sent SIGINT to the process during a normal, coordinated teardown.

### Why remove SIGINT?

- `go s.Stop()` already handles unexpected session loss — SIGINT was just a backstop in case `Stop()` hangs
- No other component (rootcoord, proxy, datanode, querynode) has this logic
- The false-positive risk (killing the process during normal shutdown) outweighs the marginal benefit

issue: #48242

PR #51771, opened 2026-07-23
fix: defer MixCoord session shutdown

xiaocai2333:
## What this PR does

Fixes ownership of the shared etcd Session used by MixCoord on the `2.6` branch.

- Tracks whether QueryCoord and DataCoord own Sessions created internally.
- Prevents child coordinators from stopping a Session injected by MixCoord.
- Makes MixCoord revoke its shared Session only after QueryCoord, DataCoord, and RootCoord have stopped.
- Adds ownership and shutdown-order tests.

## Why

QueryCoord could revoke the shared MixCoord Session while DataCoord was still draining compaction and index work. That could activate a new MixCoord before the old coordinator stack had fully stopped.

The Session creator now owns its lifecycle, preserving the lease as the fencing boundary until all child coordinators finish shutdown.

## Verification

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/coordinator ./internal/querycoordv2 ./internal/datacoord ./internal/rootcoord -run "TestMixCoordStopOwnsSharedSession|TestSessionOwnership"`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/coordinator ./internal/querycoordv2 ./internal/datacoord`
- `MILVUS_CONF_ROCKSMQ_PATH=<isolated-temp-path>/rocksmq go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/rootcoord`

issue: #51769
pr: #48252

### DIFF
--- a/internal/coordinator/mix_coord.go
+++ b/internal/coordinator/mix_coord.go
@@ -350,6 +350,12 @@ func (s *mixCoordImpl) Stop() error {
 	if err := s.rootcoordServer.Stop(); err != nil {
 		log.Error("Failed to stop rootCoord", zap.Error(err))
 	}
+
+	// MixCoord owns the shared session. Revoke it only after all child
+	// coordinators have stopped using it.
+	if s.session != nil {
+		s.session.Stop()
+	}
 	s.cancel()
 	return nil
 }
--- a/internal/coordinator/mix_coord_test.go
+++ b/internal/coordinator/mix_coord_test.go
@@ -32,9 +32,11 @@ import (
 	"github.com/milvus-io/milvus-proto/go-api/v2/milvuspb"
 	"github.com/milvus-io/milvus/internal/datacoord"
 	"github.com/milvus-io/milvus/internal/querycoordv2"
+	"github.com/milvus-io/milvus/internal/rootcoord"
 	"github.com/milvus-io/milvus/internal/util/dependency"
 	kvfactory "github.com/milvus-io/milvus/internal/util/dependency/kv"
 	"github.com/milvus-io/milvus/internal/util/pathutil"
+	"github.com/milvus-io/milvus/internal/util/sessionutil"
 	"github.com/milvus-io/milvus/internal/util/testutil"
 	"github.com/milvus-io/milvus/pkg/v2/proto/datapb"
 	"github.com/milvus-io/milvus/pkg/v2/util/etcd"
@@ -42,6 +44,36 @@ import (
 	"github.com/milvus-io/milvus/pkg/v2/util/paramtable"
 	"github.com/milvus-io/milvus/pkg/v2/util/tikv"
 )
+
+func TestMixCoordStopOwnsSharedSession(t *testing.T) {
+	mockey.PatchConvey("shared session stops after child coordinators", t, func() {
+		ctx := context.Background()
+		coord, err := NewMixCoordServer(ctx, dependency.NewDefaultFactory(true))
+		assert.NoError(t, err)
+		coord.session = sessionutil.NewSession(ctx)
+
+		stopOrder := make([]string, 0, 4)
+		mockey.Mock((*querycoordv2.Server).Stop).To(func(*querycoordv2.Server) error {
+			stopOrder = append(stopOrder, "querycoord")
+			return nil
+		}).Build()
+		mockey.Mock((*datacoord.Server).Stop).To(func(*datacoord.Server) error {
+			stopOrder = append(stopOrder, "datacoord")
+			return nil
+		}).Build()
+		mockey.Mock((*rootcoord.Core).Stop).To(func(*rootcoord.Core) error {
+			stopOrder = append(stopOrder, "rootcoord")
+			return nil
+		}).Build()
+		mockey.Mock((*sessionutil.Session).Stop).To(func(*sessionutil.Session) {
+			stopOrder = append(stopOrder, "session")
+		}).Build()
+
+		err = coord.Stop()
+		assert.NoError(t, err)
+		assert.Equal(t, []string{"querycoord", "datacoord", "rootcoord", "session"}, stopOrder)
+	})
+}
 
 func TestMixcoord_EnableActiveStandby(t *testing.T) {
 	randVal := rand.Int()
--- a/internal/datacoord/server.go
+++ b/internal/datacoord/server.go
@@ -133,7 +133,9 @@ type Server struct {
 	factory         dependency.Factory
 
 	session          sessionutil.SessionInterface
+	sessionOwned     bool
 	icSession        sessionutil.SessionInterface
+	icSessionOwned   bool
 	dnSessionWatcher sessionutil.SessionWatcher
 	qnSessionWatcher sessionutil.SessionWatcher
 
@@ -397,7 +399,9 @@ func (s *Server) SetDataNodeCreator(f func(context.Context, string, int64) (type
 
 func (s *Server) SetSession(session sessionutil.SessionInterface) error {
 	s.session = session
+	s.sessionOwned = false
 	s.icSession = session
+	s.icSessionOwned = false
 	if s.session == nil {
 		return merr.WrapErrServiceNotReadyMsg("session is nil, the etcd client connection may have failed")
 	}
@@ -520,11 +524,13 @@ func (s *Server) initSegmentManager() error {
 func (s *Server) initSession() error {
 	if s.icSession == nil {
 		s.icSession = sessionutil.NewSession(s.ctx)
+		s.icSessionOwned = true
 		s.icSession.Init(typeutil.IndexCoordRole, s.address, true, true)
 		s.icSession.SetEnableActiveStandBy(s.enableActiveStandBy)
 	}
 	if s.session == nil {
 		s.session = sessionutil.NewSession(s.ctx)
+		s.sessionOwned = true
 
 		s.session.Init(typeutil.DataCoordRole, s.address, true, true)
 		s.session.SetEnableActiveStandBy(s.enableActiveStandBy)
@@ -986,18 +992,22 @@ func (s *Server) Stop() error {
 		s.qnSessionWatcher.Stop()
 	}
 
-	if s.session != nil {
-		s.session.Stop()
-	}
-
-	if s.icSession != nil {
-		s.icSession.Stop()
-	}
+	s.stopSessions()
 
 	s.stopServerLoop()
 	log.Info("datacoord serverloop stopped")
 	log.Warn("datacoord stop successful")
 	return nil
+}
+
+func (s *Server) stopSessions() {
+	if s.sessionOwned && s.session != nil {
+		s.session.Stop()
+	}
+
+	if s.icSessionOwned && s.icSession != nil {
+		s.icSession.Stop()
+	}
 }
 
 // CleanMeta only for test
--- a/internal/datacoord/session_ownership_test.go
+++ b/internal/datacoord/session_ownership_test.go
@@ -1,0 +1,45 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may not use this file
+// except in compliance with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package datacoord
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/milvus-io/milvus/internal/util/sessionutil"
+)
+
+func TestSessionOwnership(t *testing.T) {
+	injectedSession := sessionutil.NewMockSession(t)
+	server := &Server{}
+
+	err := server.SetSession(injectedSession)
+	assert.NoError(t, err)
+	assert.False(t, server.sessionOwned)
+	assert.False(t, server.icSessionOwned)
+	server.stopSessions()
+
+	dataSession := sessionutil.NewMockSession(t)
+	indexSession := sessionutil.NewMockSession(t)
+	dataSession.EXPECT().Stop().Once()
+	indexSession.EXPECT().Stop().Once()
+	server.session = dataSession
+	server.sessionOwned = true
+	server.icSession = indexSession
+	server.icSessionOwned = true
+	server.stopSessions()
+}
--- a/internal/querycoordv2/server.go
+++ b/internal/querycoordv2/server.go
@@ -79,6 +79,7 @@ type Server struct {
 	tikvCli             *txnkv.Client
 	address             string
 	session             sessionutil.SessionInterface
+	sessionOwned        bool
 	sessionWatcher      sessionutil.SessionWatcher
 	sessionWatcherMu    sync.Mutex
 	kv                  kv.MetaKv
@@ -155,6 +156,7 @@ func (s *Server) Register() error {
 
 func (s *Server) SetSession(session sessionutil.SessionInterface) error {
 	s.session = session
+	s.sessionOwned = false
 	if s.session == nil {
 		return merr.WrapErrServiceNotReadyMsg("session is nil, the etcd client connection may have failed")
 	}
@@ -247,6 +249,7 @@ func (s *Server) initSession() error {
 	// Init QueryCoord session
 	if s.session == nil {
 		s.session = sessionutil.NewSession(s.ctx)
+		s.sessionOwned = true
 		s.session.Init(typeutil.QueryCoordRole, s.address, true, true)
 		s.enableActiveStandBy = Params.QueryCoordCfg.EnableActiveStandby.GetAsBool()
 		s.session.SetEnableActiveStandBy(s.enableActiveStandBy)
@@ -587,12 +590,16 @@ func (s *Server) Stop() error {
 	s.cancel()
 	s.wg.Wait()
 
-	if s.session != nil {
-		s.session.Stop()
-	}
+	s.stopSession()
 
 	log.Info("QueryCoord stop successfully")
 	return nil
+}
+
+func (s *Server) stopSession() {
+	if s.sessionOwned && s.session != nil {
+		s.session.Stop()
+	}
 }
 
 // UpdateStateCode updates the status of the coord, including healthy, unhealthy
--- a/internal/querycoordv2/session_ownership_test.go
+++ b/internal/querycoordv2/session_ownership_test.go
@@ -1,0 +1,38 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may not use this file
+// except in compliance with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package querycoordv2
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/milvus-io/milvus/internal/util/sessionutil"
+)
+
+func TestSessionOwnership(t *testing.T) {
+	injectedSession := sessionutil.NewMockSession(t)
+	server := &Server{}
+
+	err := server.SetSession(injectedSession)
+	assert.NoError(t, err)
+	assert.False(t, server.sessionOwned)
+	server.stopSession()
+
+	server.sessionOwned = true
+	injectedSession.EXPECT().Stop().Once()
+	server.stopSession()
+}
--- a/internal/rootcoord/root_coord.go
+++ b/internal/rootcoord/root_coord.go
@@ -119,10 +119,11 @@ type Core struct {
 	quotaCenter    *QuotaCenter
 	keyManager     *KeyManager
 
-	stateCode atomic.Int32
-	initOnce  sync.Once
-	startOnce sync.Once
-	session   sessionutil.SessionInterface
+	stateCode    atomic.Int32
+	initOnce     sync.Once
+	startOnce    sync.Once
+	session      sessionutil.SessionInterface
+	sessionOwned bool
 
 	factory dependency.Factory
 
@@ -309,6 +310,7 @@ func (c *Core) SetTiKVClient(client *txnkv.Client) {
 
 func (c *Core) SetSession(session sessionutil.SessionInterface) error {
 	c.session = session
+	c.sessionOwned = false
 	if c.session == nil {
 		return merr.WrapErrServiceNotReadyMsg("session is nil, the etcd client connection may have failed")
 	}
@@ -706,7 +708,7 @@ func (c *Core) cancelIfNotNil() {
 }
 
 func (c *Core) revokeSession() {
-	if c.session != nil {
+	if c.sessionOwned && c.session != nil {
 		// wait at most one second to revoke
 		c.session.Stop()
 		log.Ctx(c.ctx).Info("rootcoord session stop")
--- a/internal/rootcoord/session_ownership_test.go
+++ b/internal/rootcoord/session_ownership_test.go
@@ -1,0 +1,39 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may not use this file
+// except in compliance with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rootcoord
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/milvus-io/milvus/internal/util/sessionutil"
+)
+
+func TestSessionOwnership(t *testing.T) {
+	injectedSession := sessionutil.NewMockSession(t)
+	core := &Core{}
+
+	err := core.SetSession(injectedSession)
+	assert.NoError(t, err)
+	assert.False(t, core.sessionOwned)
+
+	core.revokeSession()
+
+	core.sessionOwned = true
+	injectedSession.EXPECT().Stop().Once()
+	core.revokeSession()
+}


czs007:
This PR correctly centralizes shared-session ownership in `mixCoordImpl`, but moving the lease revoke to the very end of `Stop()` widens several windows in which the process dies before the revoke ever runs, leaving standby takeover to wait out the session TTL.

## Medium

**`internal/coordinator/mix_coord.go:114`**
The standby→active activation runs in a bare goroutine spawned from `Register()` (no `wg.Add`), and `Stop()` neither checks activation state nor waits for it. If `Stop()` lands inside the activation window, the child `Stop()` calls have real side effects on the still-initializing coordinators: `querycoordv2.Server.Stop()` unconditionally runs `s.cancel()` + `s.wg.Wait()`, and `rootcoord`'s `cancelIfNotNil()` cancels the rootcoord ctx (derived from the mixcoord ctx), aborting the in-flight `initInternal` errgroup. The error propagates back through `ProcessActiveStandBy` → the activation goroutine, where `s.ctx.Err()` is still `nil` (because `s.cancel()` is sequenced after `session.Stop()`), so it hits `panic(err)` and exits the process. Before this PR the revoke happened inside the first child's `Stop()`, immediately after the trigger point; now two full child `Stop()` calls (including rootcoord's `wg.Wait()`) sit between the trigger and the revoke, substantially widening the race. If the panic wins, the ACTIVE key and the three legacy role keys survive until `common.session.ttl` (default 15s) expires. Note the exposure is broader than standby deployments: `initSession` hardcodes `SetEnableActiveStandBy(true)`, so even single-replica standalone has this goroutine and any SIGTERM during startup lands in the window. Also, datacoord and querycoord are constructed with the `NewMixCoordServer` argument ctx rather than the cancellable mixcoord ctx, so `s.cancel()` does not reach them — a datacoord still in `Init`/`Start` will keep bringing up GC/compaction loops after the lease is gone.
*Suggestion:* track the activation goroutine with the server `WaitGroup` and `Wait()` before revoking, or set a `stopping` flag in `Stop()` so `activateFunc` returns `nil` promptly. At minimum, document the uncovered window in the PR description or a code comment. *(surfaced during verification)*

**`internal/coordinator/mix_coord.go:356`**
`s.cancel()` is sequenced *after* `s.session.Stop()`, so any goroutine blocked in `ProcessActiveStandBy` observes `s.ctx.Err() == nil` and takes the `panic(err)` branch instead of returning cleanly. The standby path is affected the same way: `session.Stop()` cancels only the Session's own (child) ctx, the `registerActiveFn` etcd Txn fails with `context canceled`, and that is treated as non-retryable and returned immediately — while the mixcoord ctx is still live. Since `session.Stop()` also blocks in `wg.Wait()` (up to ~1s waiting on keepalive revoke), the standby goroutine reliably wins the race, making `exit(2)` the normal shutdown path for standby nodes. This is pre-existing rather than introduced here, but the PR touches exactly these lines.
*Suggestion:* move `s.cancel()` above `s.session.Stop()`; the session ctx is a child of the mixcoord ctx, so revoke semantics are unchanged. Note this only fixes half the problem — see the activation-window item regarding child ctxs that `s.cancel()` does not govern. *(surfaced during verification)*

**`internal/coordinator/mix_coord.go:356`**
The whole MixCoord `Stop()` is wrapped by `exitWhenStopTimeout` with `rootCoord.gracefulStopTimeout` (default 5s); on timeout it dumps pprof and calls `os.Exit(1)`, so any remaining statements in `Stop()` never execute. This PR moves the shared-session revoke from the first child's `Stop()` to the *last* statement of `mixCoordImpl.Stop()` — after all three child `Stop()` calls. The PR's own stated motivation ("DataCoord may still be draining compaction and index work") describes precisely the case where that chain exceeds 5s, in which case the process is force-killed and the lease is never explicitly revoked; the ACTIVE key and legacy role keys then wait out `common.session.ttl` (default 15s). In other words, the deferred revoke only takes effect when the graceful chain finishes within 5s — exactly the scenario where it wasn't needed.
*Suggestion:* either revoke at a bounded point with its own timeout, align `rootCoord.gracefulStopTimeout` with the expected drain duration, or state explicitly in the PR description that the force-exit path degrades to TTL-based fencing. *(surfaced during verification)*

## Low

**`internal/rootcoord/root_coord.go:710`**
The only non-test construction site for all three coordinators is `mix_coord.go:92-94`, and mixcoord's `Init()` calls `initSession()` → `SetSession()` on all three before any child `Init()`. All three `SetSession` implementations hardcode `sessionOwned = false` (and datacoord additionally `icSessionOwned = false`). The `s.session == nil` branches that would set ownership to `true` are therefore unreachable in production, which makes the newly added ownership checks in `revokeSession()` / `stopSession()` / `stopSessions()` permanent no-ops and the four `*Owned` fields dead weight. That is the intended behavior today, but it turns "no owning entry point exists" into an implicit invariant: someone later creating a session inside a child, or deleting these apparently-dead fields, would silently change revoke semantics.
*Suggestion:* document the invariant at `SetSession`/`stopSessions`, or drop the child-side `session.Stop()` paths entirely and make the owner solely responsible, rather than keeping ownership branches that can never fire. *(surfaced during verification)*

**`internal/coordinator/mix_coord_test.go:48`**
`TestMixCoordStopOwnsSharedSession` mocks out `Stop` for all three coordinators *and* `(*sessionutil.Session).Stop`, asserting only the ordering of four strings. Consequently, if someone re-adds `s.session.Stop()` to any child's real `Stop()` path, this test still passes — the child `Stop` has been replaced by a string append. The only real regression guards are the three `session_ownership_test.go` files, which inject a `NewMockSession` without an `EXPECT().Stop()` so an erroneous revoke fails on an unexpected call — but those cover only the `revokeSession()`/`stopSession()`/`stopSessions()` helpers, never a full child `Server.Stop()`.
*Suggestion:* add a test that injects a `NewMockSession` (with no `EXPECT().Stop()`) into querycoordv2/datacoord and invokes the real `Stop()`, so an accidental revoke fails loudly. *(surfaced during verification)*

**`internal/coordinator/mix_coord.go:423`**
The gRPC shutdown path calls `GracefulStop()` before `mixCoord.Stop()`. `GracefulStop()` sets `s.streamingCoord = nil` without synchronization, while the activation goroutine's `initInternal` reads the same field as its first step (`s.streamingCoord.Start(s.ctx)`) — an unsynchronized read/write that is both a race-detector data race and a potential nil-deref panic. This predates the PR, but the PR's correctness now depends on `Stop()` reaching its final `s.session.Stop()`: a panic during the `GracefulStop` phase occurs before `mixCoord.Stop()` even begins, so the lease is never revoked and takeover waits for the session TTL. Same failure surface as the activation-window item.
*Suggestion:* guard `streamingCoord` with a mutex or atomic, or simply `Stop()` it without nilling the field. *(surfaced during verification)*

xiaocai2333:
@czs007 Thanks — the review is right that the value of this PR depends entirely on `Stop()` actually reaching the revoke, and every one of the six findings reproduces against the code. Addressed in 324487c999; the PR description now carries the limitations too.

**`mix_coord.go:114` — activation window (Medium) — fixed.**
Took the `stopping` flag option. It is set as the first statement of `Stop()`, and the goroutine's check is widened from `s.ctx.Err() == context.Canceled` to `s.ctx.Err() != nil` (the old form also missed `DeadlineExceeded`). The goroutine body is now the named method `processActiveStandBy`, so the three branches — shutdown in progress, cancelled ctx, genuine failure still panics — are unit-tested directly instead of through a racing goroutine.

I did not take the `wg.Add`/`Wait()` option: waiting on that goroutine before the revoke risks a new deadlock, because a standby node blocks in `ProcessActiveStandBy` until the session context dies, which only happens at `session.Stop()` — the very thing we would be waiting to do.

**`mix_coord.go:356` — `s.cancel()` ordering (Medium) — fixed.**
Moved above `s.session.Stop()`. Confirmed the revoke semantics claim before doing it: `processKeepAliveResponse`'s deferred revoke uses `context.Background()` with its own 1s timeout, and `Session.Stop()` still blocks on `wg.Wait()` for it, so cancelling the parent first neither skips nor truncates the revoke. `TestMixCoordStopOwnsSharedSession` now asserts `coord.ctx.Err()` is already `context.Canceled` at the moment the session is revoked.

**`mix_coord.go:356` — `exitWhenStopTimeout` (Medium) — documented, not fixed.**
Confirmed: `rootCoord.gracefulStopTimeout` defaults to 5s (`DefaultCoordGracefulStopTimeout`) and the handler calls `os.Exit(1)`. I took the third option you offered. The other two do not hold up: bounding the revoke separately does not help because `os.Exit` skips defers, so nothing inside `Stop()` can make it unconditional; and raising the default would change shutdown behaviour for every coordinator, which is out of scope for this fix. There is now a comment at the revoke site and a "Known limitation" section in the PR body stating that on force-exit the lease is not revoked explicitly and fencing degrades to `common.session.ttl` (15s) rather than breaking.

**`root_coord.go:710` — unreachable ownership branches (Low) — documented.**
Verified independently: `mix_coord.go:92-94` is the only non-test construction site, `cmd/components/` has no standalone datacoord/querycoord/rootcoord binary, and `initSession` on all three is only reachable with an already-injected session. Kept the flags rather than deleting the child-side paths, and documented the invariant on `SetSession` and on `stopSession`/`stopSessions`/`revokeSession`, so the branches read as intentional no-ops. One detail worth recording: RootCoord's `Core` has no session-creating path at all — `c.session` is only ever assigned in `SetSession` — so `sessionOwned` there is permanently false, unlike datacoord/querycoord whose `initSession` does set it. Its comment says so.

**`mix_coord_test.go:48` — test proves nothing (Low) — fixed.**
Added `TestStopDoesNotRevokeInjectedSession` to querycoordv2, rootcoord and datacoord. Each drives the **real** `Server.Stop()` / `Core.Stop()` with an injected `MockSession` that has no `EXPECT().Stop()`, so any revoke on the shutdown path fails as an unexpected call. DataCoord's runs against a fully started server from `newTestServer` rather than a bare struct, so it covers the real shutdown sequence.

Negative control, since a guard test that cannot fail is the exact problem you raised: removing the `sessionOwned` check from `datacoord.stopSessions` and `querycoordv2.stopSession` makes both new tests fail; restoring it makes them pass again.

**`mix_coord.go:423` — `streamingCoord` race (Low) — fixed.**
Added `streamingCoordMu` with `get/setStreamingCoord`. `GracefulStop` swaps the pointer out under the lock and stops it outside, so it is also idempotent; `initInternal` reads through the getter and returns `ServiceNotReady` if it is already gone instead of nil-dereferencing; `RegisterStreamingCoordGRPCService` is nil-guarded.

**One correction on the first item.** The claim that `s.cancel()` does not reach datacoord and querycoord because they get the argument ctx does not quite hold: `NewQueryCoord` and `datacoord`'s own `Stop()` each derive and cancel their own context, so their shutdown is not gated on the mixcoord ctx. The real hole is narrower and worse — `datacoord.Server.Stop()` CAS-guards on `StateCode_Healthy`, so a DataCoord still in `Init` is not stopped *at all*, regardless of which ctx it was constructed with. That is a pre-existing shutdown-contract issue rather than a session-ownership one, and changing it on a 2.6 backport is not something I want to fold in here, so it is called out in the PR description and left for a follow-up.

Verification: `./internal/coordinator`, `./internal/querycoordv2`, `./internal/rootcoord` and `./internal/datacoord` all pass in full with `-tags dynamic,test -gcflags="all=-N -l" -count=1`.

czs007:
Adversarial review of the MixCoord shutdown hardening confirms the core design (per-creator session ownership, deferred revoke after children stop) but found one blocking gap and two lifecycle races on the shutdown path.

## High

**`internal/coordinator/mix_coord.go:475` — `stopping` is armed in `Stop()` but not in `GracefulStop()`, leaving the activation window open**

`stopping.Store(true)` appears only at the top of `Stop()` (mix_coord.go:361). `GracefulStop()` (mix_coord.go:475-484) clears `streamingCoord` under the lock but neither sets `stopping` nor cancels `s.ctx`. The gRPC wrapper calls `GracefulStop()` first (service.go:283), then blocks in `utils.GracefulStopGRPCServer` (up to `gracefulStopTimeout`, default 3s in configs/milvus.yaml:917) and `s.grpcWG.Wait()`, and only then calls `Stop()` (service.go:294). Throughout that window both guard conditions at mix_coord.go:136 are false — `s.ctx` is derived from `context.WithCancel(context.Background())` (cmd/roles/roles.go:375) and SIGTERM does not cancel it.

*Failure:* a standby MixCoord wins the ACTIVE key while `Stop()` is blocked in `GracefulStopGRPCServer` after `streamingCoord` was nil'd → `initInternal` returns `ServiceNotReady`, neither guard at mix_coord.go:136 holds, and the process `panic`s at mix_coord.go:140 before `s.cancel()` and `s.session.Stop()` run, so the shared etcd lease is never revoked and the incoming MixCoord is blocked until `common.session.ttl` (default 15s) expires — the exact outcome this PR set out to prevent, with a non-zero exit on top.

Note the trigger is broader than a nil `streamingCoord`: *any* non-ctx error surfaced by `ProcessActiveStandBy` in that window takes the same panic path.

*Suggestion:* set `s.stopping.Store(true)` at the top of `GracefulStop()` as well (or have the nil-`streamingCoord` branch itself signal shutdown). `TestMixCoordProcessActiveStandBy` sets `stopping`/`cancel()` by hand and never drives the `GracefulStop → Stop` sequence, so please add coverage for that ordering. (raised by tinswzy, chyezh)

## Medium

**`internal/coordinator/mix_coord.go:190` — `Start()` and `Stop()` run concurrently on the same `streamingCoord` outside the lock**

The mutex protects the *field*, not the *server*: `initInternal` takes the pointer under the lock (mix_coord.go:185) and then calls `streamingCoord.Start(s.ctx)` outside it, while `GracefulStop` calls `Stop()` on the same object outside the lock (mix_coord.go:481-483). Both mutate streamingcoord's global singleton registries — `Start` → `balance.Register` / `broadcast.Register` (streamingcoord/server/server.go:62,73); `Stop` → `balance.Release()` / `broadcast.Release()` / `resource.Release()` (server.go:88-92) — and `Release()` is a silent no-op when the future is not yet ready (balance/singleton.go:31-33, broadcast/singleton.go:65-67).

*Failure:* activation grabs the pointer at mix_coord.go:185 and releases the lock just as `GracefulStop` nils the field and calls `Stop()` → the `Release()` calls no-op because the futures are not ready, `resource.Release()` closes `streamingNodeManagerClient` anyway, and the subsequently registered balancer/broadcaster are never closed by anyone; their goroutines hang off `context.WithCancelCause(context.Background())` (balancer_impl.go:55), so `s.cancel()` at mix_coord.go:384 cannot reach them and the shutting-down MixCoord keeps running WAL balancing / pchannel assignment against an already-closed manager client, plus a permanently registered `paramtable` config handler, until the process exits.

*Suggestion:* arming `stopping` earlier does not close this race. Serialize `Start`/`Stop` — e.g. hold the lock in `GracefulStop` until `Stop()` returns, or give the server lifetime/once semantics that make `Start` and `Stop` mutually exclusive. (raised by tinswzy)

**`internal/coordinator/mix_coord.go:160` — `activateFunc` has no shutdown barrier, and `Stop()` does not wait for an in-flight activation**

`stopping` is only consumed *after* `ProcessActiveStandBy` returns (mix_coord.go:136), to decide panic-vs-silent-return. Neither `activateFunc` (mix_coord.go:160-174) nor `initInternal` checks it, and the activation goroutine (`go s.processActiveStandBy(...)`, mix_coord.go:122) is not tracked in `s.wg`, so `Stop()` never waits for it.

*Failure:* SIGTERM arrives while `activateFunc → initInternal` is mid-flight (e.g. parked in `rootcoordServer.Init()`) → `Stop()` stops the three child coordinators (mix_coord.go:367-379) while `initInternal` concurrently `Init()`s and `Start()`s the same three; the restarted datacoord/querycoord hang off the outer ctx `c` (mix_coord.go:101-102) rather than the cancellable one, so `s.cancel()` cannot stop their server loops, etcd watches, and compaction goroutines, and `startAndUpdateHealthy()` (mix_coord.go:172, 260-264) flips the state code back to `StateCode_Healthy` — the shutting-down pod reports ready again and k8s keeps routing traffic to it.

*Suggestion:* this predates the PR, but hardening the MixCoord shutdown path is precisely its goal: check the shutdown flag at the entry of `activateFunc` (and again before `startAndUpdateHealthy`), and have `Stop()` wait for any in-flight activation. Deriving querycoord/datacoord from the cancellable `ctx` would also make `s.cancel()` effective. (surfaced during verification)

xiaocai2333:
Second adversarial review round — all three findings reproduce, all three fixed in 5e289eafa7. The High is a real hole in my own previous commit, not a pre-existing one.

**High — `stopping` armed in `Stop()` but not `GracefulStop()` — fixed, and this was my bug.**

Confirmed exactly as described. `service.go:283` calls `GracefulStop()`, then blocks in `utils.GracefulStopGRPCServer` and `s.grpcWG.Wait()`, and only reaches `Stop()` at `service.go:294`. `s.ctx` descends from `context.WithCancel(context.Background())` (`roles.go:375`) and SIGTERM does not cancel it, so throughout that window both guards in `processActiveStandBy` were false — my flag placement assumed `Stop()` was the entry point of shutdown, and it is not.

Worse than "the guard did not cover it": the `ServiceNotReady` I added to `initInternal` for the nil-`streamingCoord` case was itself a *new* trigger for the panic path inside precisely that window. The previous commit narrowed one race and opened another.

Fixed by arming `stopping` at the top of `GracefulStop()`; `Stop()` still sets it for direct callers. `TestMixCoordGracefulStopArmsStopping` now drives the real `GracefulStop → Stop` ordering, asserts `s.ctx` is still live at that point (so the test would not pass for the wrong reason), and asserts the activation does not panic. Reverting the one-line fix makes it fail on both the flag assertion and `should not panic`.

**Medium — `Start`/`Stop` concurrent on the same `streamingCoord` — fixed.**

Correct, and the distinction between guarding the *field* and guarding the *server* is the part I got wrong. Verified the consequence: `balance.Release()` returns early when `!s.Ready()` (`balance/singleton.go:31-33`), so a `Stop()` that overlaps a `Start()` releases nothing while the subsequently registered balancer and broadcaster are left with no owner.

`startStreamingCoord` and `GracefulStop` now both hold `streamingCoordMu` across the entire `Start()`/`Stop()` call, so the two are mutually exclusive: whichever wins, the loser either starts a live server or sees nil and returns `ServiceNotReady`. Renamed the field comment to say it is a lifetime lock so the next reader does not repeat my mistake.

**Medium — `activateFunc` has no shutdown barrier — fixed (the barrier, not the wait).**

Took the first half of the suggestion: `activateFunc` now returns `nil` early both on entry and again after `initInternal`, so winning the ACTIVE key mid-shutdown neither re-`Init`/`Start`s the child coordinators nor reaches `startAndUpdateHealthy`. The state-code flip was the concrete harm — a draining pod reporting `Healthy` again keeps k8s routing to it — and there is a second one worth recording: `startPosixCleanupTask` would start a fresh cleanup goroutine after `stopPosixCleanupTask` has already consumed its stop-once, so nothing would ever stop it. Both are covered by `TestMixCoordActivateFuncSkipsWhenStopping`, including the case where shutdown begins *during* `initInternal`.

I did not add `Stop()` waiting on the activation goroutine. A standby node blocks in `ProcessActiveStandBy` until the session context dies, which is exactly what `session.Stop()` does at the end of `Stop()` — waiting for it first would deadlock. The barrier makes the in-flight activation harmless without needing the wait; if you think the wait is still required on top, say so and I will look at restructuring the ordering rather than bolting on a `wg.Wait()`.

**Deriving querycoord/datacoord from the cancellable ctx** — still out of scope for this PR, as agreed on the previous round. To be precise about why: both derive and cancel their own contexts in their own `Stop()`, so `s.cancel()` not reaching them is not the actual defect. The actual defect is that `datacoord.Server.Stop()` CAS-guards on `StateCode_Healthy` and silently no-ops for a server still in `Init` — changing that is a shutdown-contract change, not a session-ownership one, and I would rather not land it on a 2.6 backport. Noted in the PR description as a follow-up.

**Verification.** `./internal/coordinator`, `./internal/querycoordv2`, `./internal/rootcoord`, `./internal/datacoord` all pass with `-tags dynamic,test -gcflags="all=-N -l" -count=1`. Note these four must run with `-p 1` locally: run in parallel they are four separate test binaries contending for the hard-coded global rocksmq lock at `/tmp/milvus/rocksmq/_meta_kv/LOCK`, which fails inside `dependency.NewDefaultFactory` before any code under test runs. Not a regression from this PR — CI runs each package in its own environment — but it cost me a round trip, so recording it here.

czs007:
Adversarial review surfaced two medium-severity concurrency issues in the MixCoord shutdown path; the session-ownership refactor and the deferred-revoke ordering themselves verified as correct.

## Medium

**`internal/coordinator/mix_coord.go:128` — activation goroutine is untracked, so the shared session lease can be revoked while child coordinators are still starting**

`Register()` spawns `processActiveStandBy` with no WaitGroup or lifecycle handle, and `Stop()` never joins it. The two `stopping.Load()` guards (lines 169 and 187) straddle `initOnce.Do(initInternal)` — there is no barrier *inside* it, so the seconds-long span in which `initInternal` starts RootCoord and then runs `Init()`/`Start()` for DataCoord and QueryCoord in an errgroup (lines 222-247) is entirely unsynchronized with `Stop()`. Note that this path is always live: `SetEnableActiveStandBy(true)` is unconditional at line 457.

Failure: MixCoord receives SIGTERM (pod delete / rolling upgrade) while the activation goroutine is inside `initInternal` → `datacoordServer.Stop()` returns without doing anything because the CAS at `internal/datacoord/server.go:964` only fires from `Healthy`, so DataCoord's `Start()` runs to completion and flips itself Healthy — launching the garbage collector, compaction trigger and index/stats inspectors — *after* `Stop()` has already cancelled `s.ctx` and revoked the shared etcd lease at lines 399/411. Those loops write DataCoord meta with no lease fencing them, while the ACTIVE key is already free for a standby MixCoord to claim.

This breaks the invariant the PR states for itself at lines 401-403 ("revoke it only after all child coordinators have stopped"). Suggestion: track the activation goroutine and make the healthy-transition and the revoke one serialized lifecycle operation — an activation-scoped context plus a join before the revoke — or at minimum re-check `stopping` and abort inside `initInternal` between child startups. *(raised by tinswzy, chyezh)*

**`internal/coordinator/mix_coord.go:512` — `GracefulStop` blocks on `streamingCoordMu` with no cancellation path, burning the coord stop budget**

`startStreamingCoord` holds `streamingCoordMu` across the whole `streamingCoord.Start(s.ctx)` call (lines 445-452), while `GracefulStop` sets `stopping` and then blocks acquiring that same mutex (lines 508-513). The only cancellation of `s.ctx` is at line 399, inside `Stop()`, which the gRPC wrapper reaches only after `GracefulStop()` returns (`internal/distributed/mixcoord/service.go:283` → `:294`) — so while `GracefulStop` is parked on the mutex, nothing can unblock `Start`. This lock scope is new in this PR; previously `GracefulStop` did not take the mutex and could not be blocked by an in-flight `Start`.

Failure: SIGTERM arrives while an activation is inside `streamingCoord.Start` and etcd is slow but reachable (large pchannel catalog / degraded etcd), so the synchronous `RecoverChannelManager` and `checkIfAllNodeGreaterThan260` work exceeds the remaining budget → `rc.svr.Stop` overruns the 5s `RootCoordCfg.GracefulStopTimeout` and `exitWhenStopTimeout` calls `os.Exit(1)` (`cmd/components/util.go:22-33`) before line 411 is ever reached, so the shared lease is never explicitly revoked and failover waits out the full `common.session.ttl`.

Suggestion: cancel a shutdown-scoped activation context (or pass a cancellable context into `Start`) before waiting for the lock. *(raised by tinswzy, chyezh)*

xiaocai2333:
Third round — both findings reproduce, both fixed in de85dd1dd5. One of them is again a defect my own previous commit introduced.

**Medium — `GracefulStop` blocks on `streamingCoordMu` with no cancellation path — fixed, and this one is mine.**

Correct, including the observation that the lock scope is new in this PR. Round 2 made `GracefulStop` hold `streamingCoordMu` across `Stop()` to serialize it against `startStreamingCoord`, but `startStreamingCoord` holds the same lock across the entire `Start(s.ctx)` call, and the only cancellation of `s.ctx` is in `Stop()` — which the gRPC wrapper reaches only after `GracefulStop()` returns (`service.go:283` → `:294`). So the fix for one race created a mutual wait with no exit: a slow-but-reachable etcd inside `RecoverChannelManager` parks `GracefulStop` until the 5s coord budget expires and `os.Exit(1)` fires, skipping the revoke. That is the second time in this PR that a fix opened a new hole on the same path.

Fixed with an activation-scoped context: `activationCtx`/`activationCancel` created in `NewMixCoordServer` as a child of `s.ctx`, passed to `startStreamingCoord`, and cancelled at the top of `GracefulStop` *before* it contends for the lock. `TestMixCoordGracefulStopUnblocksInFlightStart` parks a mocked `Start` on `<-ctx.Done()` and fails with a 10s timeout if the cancel is removed; it also asserts `s.ctx` is still live at that point, so it cannot pass for the wrong reason.

**Medium — activation goroutine untracked, no barrier inside `initInternal` — fixed, including the part I previously deferred.**

Reproduced. The two `stopping` checks straddled `initOnce.Do(initInternal)` with nothing inside it, and the DataCoord consequence you traced is exactly right: `datacoord.Server.Stop()` CAS-fires only from `Healthy`, so a DataCoord still in `Init` is not stopped, then finishes `Start()`, flips itself `Healthy` and launches GC / compaction / inspectors — after the lease was revoked. You are also right that this breaks the invariant the PR states for itself, which is why I am no longer treating it as out of scope: I declined this in round 1 on the grounds that it was a pre-existing DataCoord issue, and that framing was wrong once it is the stated invariant that fails.

Three parts:

1. `initInternal` now calls `abortIfStopping` before each stage — streamingCoord start, RootCoord `Init`, RootCoord `Start`, and inside both errgroup branches before DataCoord/QueryCoord `Init` and `Start`.
2. The activation goroutine is tracked in `s.wg`, and `Stop()` joins it after `s.cancel()` and before the revoke.
3. To make that join deadlock-free, `QueryCoord` and `DataCoord` are now constructed from the cancellable mixcoord ctx instead of the argument ctx. This is the change your round-1 review suggested and I declined; it is a prerequisite here, because every blocking point the activation can park on must be released by `s.cancel()` for the join to terminate — the ACTIVE-key wait (session ctx), `streamingCoord.Start` (activation ctx), RootCoord `Init`/`Start` (already derived), and now the other two children. Ordering matters and is asserted: `TestMixCoordStopJoinsActivation` fails with "session revoked while activation was still running" if the join is removed, and with "Stop() deadlocked" if the join is moved above `s.cancel()`.

I did not change DataCoord's `Healthy`-only CAS in `Stop()`. With the barriers plus the join, MixCoord no longer leaves a DataCoord starting behind its back, so the CAS is no longer reachable in the failure mode you described. The CAS is still questionable on its own terms — a server killed during `Init` cleans up nothing — but that is a DataCoord shutdown-contract change and I would rather not land it in a 2.6 backport on top of this. Recorded as a follow-up in the PR description rather than silently dropped.

**Verification.** All four packages pass with `-tags dynamic,test -gcflags="all=-N -l" -count=1 -p 1`, and `gofmt` is clean. Every fix in this round has a negative control that reproduces the specific failure, including the two-way one on the join placement.

Worth stating plainly: rounds 2 and 3 each found that my previous round's fix introduced a new defect on this same path. The shutdown sequence has three entry points (`GracefulStop`, `Stop`, the activation goroutine) and three contexts, and I have been fixing it one narrow window at a time. If a fourth round finds another, the right response is probably to restructure the lifecycle rather than patch the next window.

czs007:
## Adversarial review — 3 issues

The core fix (session ownership flag) is correct and clean, but the ordering guarantee added around `Stop()` does not hold on the real code path, and one in-flight-activation window remains open.

---

### High

**`internal/coordinator/mix_coord.go:459` — the lease is already revoked by `s.cancel()`, so `s.wg.Wait()` provides no ordering guarantee**

`Stop()` runs `s.cancel()` (:450) → `s.wg.Wait()` (:459) → `s.session.Stop()` (:471), and the comment at :452-463 claims the lease only disappears after the activation goroutine is joined. It does not: the session is built from the mixcoord child ctx (`mix_coord.go:515` → `session_util.go:286`), the actual `etcdCli.Revoke` lives in the keepalive goroutine's `defer` (`session_util.go:603`), and that goroutine's only exit condition is the session ctx (`session_util.go:621`). `Session.Stop()` (`session_util.go:979-985`) is just `cancel()` + `wg.Wait()` and issues no revoke — so the revoke fires at :450, concurrently with the join, and :471 degenerates into a second `wg.Wait()`.

*Failure:* a MixCoord whose activation goroutine has passed the `abortIfStopping("dataCoord start")` barrier and is inside `datacoordServer.Start()` receives SIGTERM → the lease is revoked at :450 while DataCoord is still starting, the lease-bound ACTIVE key vanishes, the standby MixCoord activates immediately, and two DataCoords drive compaction/GC against the same meta.

*Suggestion:* anchor the session on a ctx not reached by `s.cancel()` so the revoke is driven solely by `s.session.Stop()` at :471. That requires giving `ProcessActiveStandBy` a separate cancellation signal (it watches the ACTIVE key on the session ctx), otherwise :459 deadlocks. If the restructure is deferred, at minimum rewrite the :452-463 comment to describe actual behaviour, and change the two assertions that currently encode the defect as expected — both new tests mock `(*sessionutil.Session).Stop`, so `joinedBeforeRevoke` in `TestMixCoordStopJoinsActivation` is falsely green, and `assert.ErrorIs(t, ctxErrAtRevoke, context.Canceled)` in `TestMixCoordStopOwnsSharedSession` asserts that the ctx is already cancelled at revoke time. *(raised by czs007)*

---

### Medium

**`internal/coordinator/mix_coord.go:438` — child coordinators are stopped before the activation join, so an in-flight `DataCoord.Start()` escapes `Stop()`**

`Stop()` orders `queryCoordServer.Stop()` (:434) → `datacoordServer.Stop()` (:438) → `rootcoordServer.Stop()` (:442) → `s.cancel()` (:450) → `s.wg.Wait()` (:459). DataCoord's `Stop()` is gated on `CompareAndSwap(Healthy, Abnormal)` and returns `nil` for any other state (`datacoord/server.go:964`), while `startDataCoord` sets `Healthy` last (`datacoord/server.go:351-357`). Once activation has passed the `abortIfStopping` barrier the barrier no longer applies — `Start()` never re-checks `stopping`, and the errgroup ctx is discarded at `mix_coord.go:258` (`g, _ := errgroup.WithContext(...)`).

*Failure:* SIGTERM arrives after activation enters `datacoordServer.Start()` but before `UpdateStateCode(Healthy)` → the CAS at `datacoord/server.go:964` fails against `Initializing`, `Stop()` returns `nil` having stopped nothing, and `Start()` then completes and brings up garbage collector, compaction inspector, stats/index inspectors and the server loop, which keep writing meta after `mixCoordImpl.Stop()` has returned and the lease is gone.

So `s.wg.Wait()` waits for "startup finished", not "stopped"; the PR description's claim that the barrier removes this state holds only for phases not yet entered.

*Suggestion:* re-run the idempotent child-coordinator stop sequence after :459 (DataCoord's CAS makes repeat calls safe), or move all three `Stop()` calls after the join — the latter depends on fixing the revoke timing above. *(raised by czs007)*

---

### Low

**`internal/coordinator/mix_coord.go:209` — TOCTOU on the second `stopping` check can flip a draining pod back to `Healthy`**

Only a `log.Info` separates the `s.stopping.Load()` check at :209 from `s.startAndUpdateHealthy()` at :214, which sets `stateCode` to `Healthy` at :327. The shutdown path never resets that field — the only two `UpdateStateCode` calls in the file are `StandBy` (:183) and `Healthy` (:327).

*Failure:* `GracefulStop()` executes `s.stopping.Store(true)` (:568) between :209 and :214 → :327 sets `Healthy`, and the draining MixCoord keeps reporting `Healthy` to `GetComponentStates` / readiness probes for the remaining gRPC drain window (up to `common.gracefulStopTimeout`), so traffic continues to be routed to it until the process exits.

The window is two instructions wide, hence low severity.

*Suggestion:* make the `Healthy` transition a CAS that only applies when shutdown has not begun, or hold a single lock across both the `stopping` store and the state transition. Separately, the code comment at :207-208 ("restart the POSIX cleanup task after `stopPosixCleanupTask` has already consumed its stop-once") describes a consequence that cannot occur — `startPosixCleanupTask` is guarded by `posixCleanupStartOnce` (:385) — so it is worth correcting while here. *(raised by czs007)*

xiaocai2333:
Fourth round. The High is correct and it is the most serious finding on this PR: the ordering guarantee has been broken since my *first* review commit, and the test I added for it was falsely green. Fixed in 03b2c16caa by reducing scope rather than adding another layer.

**High — the lease is already revoked by `s.cancel()` — confirmed, and it is my regression.**

Verified every link: `NewSessionWithEtcd` derives the session ctx from the passed ctx (`session_util.go:295`), the real `etcdCli.Revoke` lives in the keepalive goroutine's `defer` (`:603`), that goroutine's only exit condition is the session ctx (`:621`), and `Session.Stop()` is just `cancel()` + `wg.Wait()` (`:979-985`). So cancelling `s.ctx` revokes the lease exactly as surely as calling `Session.Stop()`.

Worth being precise about the provenance, because it is not a pre-existing issue: the original commit `0ca26ddc12` ordered `session.Stop()` → `s.cancel()` and was **correct**. I hoisted `s.cancel()` above it in round 1 to make the `ProcessActiveStandBy` panic guard fire, and justified it in the PR with "the session ctx is a child of the mixcoord ctx, so revoke semantics are unchanged". That justification came from checking that the revoke uses `context.Background()` and therefore still *succeeds* — I verified whether the revoke could happen and never verified *when*. Round 3 then inserted `s.wg.Wait()` between a revoke that had already fired and a `Session.Stop()` that had become a no-op, so the join guaranteed nothing, and `TestMixCoordStopJoinsActivation` passed only because it mocked `(*sessionutil.Session).Stop` and measured the mock instead of the lease.

Fix:

- `s.cancel()` moved back below `s.session.Stop()`, restoring the original ordering. The panic guard no longer needs the early cancel — the `stopping` flag armed in `GracefulStop` is both earlier and more reliable than the ctx ever was.
- The activation join and its WaitGroup tracking are removed. I am not going to re-derive it a third way: the only signal that unblocks `ProcessActiveStandBy` is the session ctx, and cancelling that is what revokes the lease, so "join, then revoke" is not expressible in the current sessionutil design. Anchoring the session on an independent ctx would make it expressible, but that changes session lifetime semantics for every component and is not something I want to land in a 2.6 backport. `Register()` now documents the constraint instead of implying a guarantee that does not exist.
- `TestMixCoordStopJoinsActivation` deleted rather than repaired — its premise was wrong, not its mechanics.
- The assertion in `TestMixCoordStopOwnsSharedSession` is inverted: `s.ctx` must still be **live** when the session is stopped. It previously asserted the defect. Negative control: re-hoisting `s.cancel()` now fails it with "the lease is revoked too early". Had that assertion existed in round 1, the regression would not have shipped.

**Medium — in-flight `DataCoord.Start()` escapes `Stop()` — acknowledged, not fixed, now stated as a limitation.**

Your trace is right, including the part that matters most: `abortIfStopping` only guards phases not yet entered, so `s.wg.Wait()` was waiting for "startup finished", not "stopped". The PR description claimed the barriers removed this state; that claim was too strong and I have corrected it rather than leaving it standing.

Closing it needs either the join (not expressible, above) or a change to DataCoord's `Healthy`-only CAS so a server killed mid-`Init` still cleans up. The second is a DataCoord shutdown-contract change; I would rather do it in its own PR against master than bundle it here. Recorded under known limitations with the failure mode spelled out.

**Low — TOCTOU on the second `stopping` check — narrowed and documented; the bogus comment is fixed.**

The two-instruction window is real and I have not closed it — the comment now says so instead of implying the check is sufficient. You are also right that the rest of that comment was simply false: I wrote that skipping the check would "restart the POSIX cleanup task after `stopPosixCleanupTask` has consumed its stop-once", but `startPosixCleanupTask` is guarded by `posixCleanupStartOnce`, so that consequence cannot occur. I invented a justification that does not hold. Removed.

**Verification.** All four packages pass with `-tags dynamic,test -gcflags="all=-N -l" -count=1 -p 1`; `gofmt` clean. The net diff of this round is −42 lines.

**On the process.** Rounds 2 and 3 each found that my previous fix had introduced a new defect, and round 4 found that the guarantee underpinning all three was never real. I said last round that a fourth finding of this kind should mean restructuring rather than patching the next window; the honest version of that here is to stop widening the PR. What remains is the piece that was correct from the start — the ownership flag, which this review confirms — plus the barriers and locks that were each independently verified, with the two windows I cannot close stated as limitations instead of dressed up as fixes.

czs007:
Adversarial review surfaced one concurrency issue in the coordinator shutdown path that can defeat the very lease-revoke guarantee this PR adds.

## Medium

**`internal/coordinator/mix_coord.go:253`** — RootCoord's `Start()` is not serialized against `Stop()`. The activation goroutine performs a single `abortIfStopping("rootCoord start")` check before entering `rootcoordServer.Start()`, and nothing guards the two against each other afterwards. The first step of `startInternal()` is `c.proxyWatcher.WatchProxy(c.ctx)` (`internal/rootcoord/root_coord.go:645-648`), which issues an etcd `Get` on `c.ctx`; if `mixCoordImpl.Stop()` reaches `s.rootcoordServer.Stop()` (`mix_coord.go:444`) it cancels that same context via `c.cancelIfNotNil()` (`root_coord.go:746`), the `Get` returns `context.Canceled`, and `startInternal` calls `log.Fatal`, which exits the process through `os.Exit(1)` (the following `panic(err)` is only a fallback). The `stopping` flag does not help here — this path never reaches the `processActiveStandBy` stopping branch.

*Failure:* a standby MixCoord wins the ACTIVE key, passes the `abortIfStopping` check, and is inside `WatchProxy`'s etcd `Get` when SIGTERM arrives and `Stop()` cancels RootCoord's context → the process exits at `root_coord.go:646` before `mix_coord.go:465` runs `s.session.Stop()`, so the shared etcd lease is never explicitly revoked and the successor MixCoord must wait out `common.session.ttl` (15s by default) before it can become active.

*Suggestion:* serialize RootCoord's `Start()` and `Stop()` (or have `Stop()` wait for an in-flight `Start()` to unwind cleanly), and treat a shutdown-induced context cancellation as a normal early return rather than `log.Fatal`/`panic`, so the shutdown path always reaches `session.Stop()`. (raised by xaxys)

xiaocai2333:
Fifth round. The finding reproduces; recorded as a known limitation rather than fixed, reasoning below. Also fixed the `ci-v2/code-check` failure (misspell) that this round's head carried.

**Medium — RootCoord's `Start()` is not serialized against `Stop()` — confirmed, not fixed.**

Traced the whole chain and it holds:

- `startInternal`'s first step is `c.proxyWatcher.WatchProxy(c.ctx)` (`root_coord.go:645`), which derives a timeout child of `c.ctx` and issues `getSessionsOnEtcd` on it (`proxy_watcher.go:89-92`).
- `mixCoordImpl.Stop()` reaches `s.rootcoordServer.Stop()` at `mix_coord.go:444`, which calls `cancelIfNotNil()` (`root_coord.go:746`) cancelling that same `c.ctx`.
- The `Get` returns `context.Canceled`, `startInternal` calls `log.Fatal` → `os.Exit(1)`.
- `s.session.Stop()` is at `mix_coord.go:466`, twenty-odd lines further down, so the process is gone before the revoke and the successor waits out `common.session.ttl`.

You are also right that the `stopping` flag cannot help: this path never returns to `processActiveStandBy`. And it is not limited to `WatchProxy` — the `c.restore(c.ctx)` call immediately after has a bare `panic(err)` with the same trigger.

I am not fixing it here. Unlike the defects in rounds 2-4, this one is not something this PR introduced: it is RootCoord's existing "startup failure kills the process" semantics, and the fix — treating a shutdown-induced cancellation as a normal early return at both exit points — is a decision about RootCoord's startup contract rather than about session ownership. After four rounds in which widening this PR repeatedly introduced new defects, the useful thing I can do is record it precisely instead of reaching into another component again.

Recorded in three places:

- The PR description's known-limitations section, which now frames both this and the DataCoord case as one limitation — *a shutdown landing inside a child's `Start()` is not controlled* — with the RootCoord variant flagged as the more severe of the two, since it kills the process rather than merely outliving `Stop()`.
- `abortIfStopping`'s doc comment (be5b6f6508), which previously read as blanket protection. It now states that it guards the gaps *between* stages only, and names both consequences. This matters because the helper is mine and a future reader would otherwise reasonably assume the window was closed.
- This comment, for the follow-up.

**Unrelated: `ci-v2/code-check` was failing on this round's head.** `static-check` reported 9 `misspell` findings — the comments and test descriptions I added across the review rounds used British `cancelled`/`cancelling` while `.golangci.yml` sets `misspell: locale: US`. Fixed in cd5201ad2a (comment and test-string only). Worth noting how it went unnoticed: for four rounds I verified with `go test` and `gofmt` but never ran the linter, so the spelling accumulated across every round and only surfaced on the fifth push. I now run `golangci-lint` over the changed packages before pushing — full linter set, all four packages, 0 issues.

**State of the PR.** Findings went 6 → 3 → 2 → 3 → 1 across the rounds, and this is the first round that did not find a defect introduced by the previous round's fix — which I take as the scope reduction in round 4 having been the right call. What remains is the ownership flag plus the barriers and locks that each have a negative control, with three windows recorded as limitations rather than claimed as fixed.

czs007:
This adversarial review confirmed two issues (one high, one medium) in the session watcher shutdown and retry paths; the remaining candidates were refuted or already documented as known limitations.

### High

- **`session_util.go:969` — Watcher rebuild loop cannot be interrupted by `Stop()`, which can push shutdown past the graceful budget and skip the explicit session revoke.** The watcher's `cancel` cancels a child of `s.ctx` (`session_util.go:859/880`), but the rebuild loop calls `retry.Do(w.s.ctx, w.handleReWatch, ...)` on the parent context, so `w.cancel()` never reaches it and `Stop()` (`session_util.go:831-833`) blocks in `wg.Wait()`. Because this PR deliberately orders `s.session.Stop()` and `s.cancel()` after the child coordinators' `Stop()` (`mix_coord.go:476/483`), `s.ctx` is guaranteed live while querycoord/datacoord stop their watchers. If a rebuild is in flight and `handleReWatch` keeps returning retriable errors, the default retry config (10 attempts, 200ms doubling capped at 3s) can sleep ~18s+ — well over the 5s `gracefulStopTimeout` noted at `mix_coord.go:470-471`, so `os.Exit(1)` fires and skips the explicit session revoke this PR exists to guarantee, falling back to TTL-based fencing. This is not a deterministic hang (a successful rebuild lets the goroutine exit normally), but the failure window is exactly the stale-token/compaction scenario the PR targets. `ProxyWatcher` has the same flaw: `proxy_watcher.go:155` retries on rootcoord's `c.ctx`, but `Core.Stop()` calls `proxyWatcher.Stop()` (`root_coord.go:739`) before `c.cancelIfNotNil()` (`root_coord.go:746`), and the retry loop never observes `closeCh`. **Suggestion:** have `sessionWatcher` store the child context created in `start(ctx)` and pass it to `retry.Do` so `w.cancel()` interrupts in-flight rebuilds; for `ProxyWatcher`, tie the retry context's cancellation to `closeCh`. *(raised by xaxys, tinswzy)*

### Medium

- **`session_util.go:982` — Retry predicate for watch rebuild is too narrow: a transient `GetSessions` error aborts on the first attempt and panics outside shutdown.** `retry.Do` first checks `IsRecoverable` (which transient `Unavailable`/`DeadlineExceeded` errors pass), then the caller's predicate, returning immediately on a mismatch (`retry.go:62`, `retry.go:79-87`). `handleReWatch`'s first call is `GetSessions` (`session_util.go:982`); a transient etcd-unavailable error matches none of the four sentinels in `IsRetriableWatchErr` (`etcd_util.go:70-78`: `ErrCompacted`/`ErrInvalidAuthToken`/`ErrUserEmpty`/`ErrAuthOldRevision`), so the retry exits on attempt one, `handleWatchErr` closes the event channel, and `handleWatchResponse` panics when `w.s.ctx` is not cancelled (`session_util.go:903-908`). An etcd member restart that both invalidates the token and causes brief unavailability would still crash a standby — precisely the scenario Fix 2 targets — and the "only transient errors are retried" comment does not match the actual predicate. `ProxyWatcher` shares the issue via the same predicate at `proxy_watcher.go:155`. **Suggestion:** widen the retry predicate to `IsRetriableWatchErr(err)` OR transient error classes (e.g. `Unavailable`/`DeadlineExceeded`), or add a dedicated transient-error retry around `GetSessions`. *(raised by tinswzy, xaxys)*

xiaocai2333:
Both findings are valid and are fixed in 544f2c22e5.

### High — watcher Stop could not interrupt rebuild retry

- `sessionWatcher` now stores and uses its child context for the initial watch, re-list retry/backoff, replacement watch, and event delivery. `Stop()` cancels that context before `wg.Wait()`.
- `ProxyWatcher` now owns a watch context that `Stop()` cancels before waiting. This does not depend on RootCoord canceling its parent context; the test asserts the parent context is still live when Stop completes.
- Proxy re-watch now replaces `eventCh` in the existing watcher goroutine instead of calling `WatchProxy` and spawning another watcher goroutine.
- Cancellation racing with a watch error or channel closure exits quietly rather than panicking or closing the Session event channel.

### Medium — rebuild retry predicate was too narrow

- Kept `IsRetriableWatchErr` narrow: it only decides which terminal watch responses trigger rebuilding.
- Added `IsRetriableWatchRebuildErr` for the unary re-list request. It includes the recoverable watch sentinels plus transient etcd/gRPC `Unavailable`, `Internal`, and `DeadlineExceeded` failures, including wrapped etcd errors such as no-leader/unhealthy.
- Moved the bounded retry around only the re-list/Get. The rewatch/application callback runs once and remains fail-fast even if it returns a transient-looking gRPC error.
- Permission, bad-credential, malformed-request, and generic callback errors still fail on the first attempt.

Verification added for both SessionWatcher and ProxyWatcher:

- recoverable watch error → transient Get failure → successful rebuild;
- Stop interrupts active retry in under one second;
- RootCoord/Session parent context remains live while watcher Stop succeeds;
- non-retriable request errors fail after one attempt;
- callback errors are not retried;
- idempotent Stop and cancellation/channel-close races do not panic or close the Session event channel.

The focused tests were repeated 10 times, the relevant existing compaction/watch tests pass, `pkg/v2/util/etcd` passes, and golangci-lint reports 0 issues for the changed packages. The PR description has also been updated with the shutdown/retry behavior and the exact verification commands.

xiaocai2333:
This PR was intentionally reset to the same Session-ownership fix as #51820. The eight follow-up commits covering auth-watch recovery and broader activation/shutdown hardening have been removed; earlier review discussion referring to those removed commits is now historical.

liliu-z (commented):
/approve

czs007:
Adversarial re-verification of this PR (head `0ca26ddc120a`) upheld all six previously reported findings — three medium-severity shutdown-ordering/concurrency issues and three low-severity coverage/reachability issues — with no new critical or high findings in an independent pass.

## Re-review of `0ca26ddc120a`

All earlier findings were re-examined against the current diff and surrounding source this round:

- **[medium] mix_coord.go:114 — activation goroutine can preempt the deferred lease revoke: Confirmed.** Activation still runs in a bare goroutine with no `wg.Add`, and `Stop()` neither checks activation state nor waits for it, so an activation error can `panic` before the revoke at the end of `Stop()`.
- **[medium] mix_coord.go:356 — `s.session.Stop()` ordered before `s.cancel()`: Confirmed.** During a concurrent activation, `Session.Stop()` cancels only the session's derived context; `s.ctx.Err()` is still nil, so the Register loop panics instead of returning cleanly.
- **[low] mix_coord.go:379 — ownership guards never fire with owned=true in production: Confirmed**, with a corrected framing — calling the guard a "no-op" was imprecise, since the `sessionOwned=false` branch is the fix itself; what stands is that the `owned=true` branch is unreachable in production.
- **[low] mix_coord_test.go:48 — mixcoord Stop test cannot catch a shared-lease-revoke regression: Confirmed.** The test replaces the children's real `Stop()` bodies entirely, so a re-introduced `session.Stop()` inside a child would go undetected.
- **[medium] mix_coord.go:342 — revoke sits inside the 5s graceful-stop force-exit window: Confirmed.** `exitWhenStopTimeout` calls `os.Exit(1)` after 5 seconds (default `DefaultCoordGracefulStopTimeout`), so the PR's own motivating scenario (DataCoord still draining compaction) is exactly the one where the explicit revoke is skipped.
- **[low] mix_coord.go:423 — `GracefulStop()` races the activation goroutine on `streamingCoord`: Confirmed.** The unsynchronized Stop-then-nil write can race the activation goroutine's read, causing a nil dereference before `Stop()` reaches the final revoke.

## This round's verified findings

### Medium

- **`internal/coordinator/mix_coord.go:114`** — The activation logic in `Register()` runs in a bare goroutine that is never added to the WaitGroup, and `Stop()` neither checks activation state nor waits for it. A concurrent `Stop()` can cancel an in-flight QueryCoord Init/Start (querycoordv2's `Stop()` unconditionally runs `s.cancel(); s.wg.Wait()`), and since the MixCoord context is not yet canceled at that point, the activation error takes the `panic(err)` branch — potentially terminating the process before the lease revoke at mix_coord.go:356-358 runs. Suggestion: track the activation goroutine in the WaitGroup and wait for it in `Stop()`, or gate the panic branch on a shutdown signal. (raised by xaxys, tinswzy)

- **`internal/coordinator/mix_coord.go:356`** — The newly added `s.session.Stop()` is ordered before the pre-existing `s.cancel()`. `Session.Stop()` only cancels the session's own derived context; the standby/activation loop then receives `context canceled` as a non-retryable error while `s.ctx.Err()` is still nil, so the Register loop panics instead of returning cleanly. Suggestion: call `s.cancel()` first, or guard the panic branch with a shutdown flag. (raised by xaxys, tinswzy)

- **`internal/coordinator/mix_coord.go:342`** — The deferred revoke is placed after the three full child `Stop()` calls, all inside `exitWhenStopTimeout(rc.svr.Stop, ...)` in cmd/components/mix_coord.go:67-69, which defaults to only 5 seconds and does `dumpPprof()` + `os.Exit(1)` on timeout. The PR's own stated scenario — DataCoord draining compaction for more than 5 seconds — is therefore precisely the case where the process is force-killed before the explicit revoke, leaving the lease to expire by TTL and weakening the fencing guarantee the PR claims to strengthen. Suggestion: move the revoke out of the 5-second force-exit path (or reorder cancel/revoke); otherwise document this limitation explicitly in the PR description. (raised by xaxys, tinswzy)

### Low

- **`internal/coordinator/mix_coord.go:379`** — The `sessionOwned=true` branch is unreachable in production. The only non-test call site of the three child constructors is mix_coord.go:92-94, and `mixCoordImpl.Init()` injects the shared session via `SetSession` (which sets `sessionOwned=false`) before any child `Init()` runs, so the children never see a nil session. The `false` branch is the actual fix and works as intended, but the `owned=true` machinery — and the three `session_ownership_test` assertions on it — only exercise a test-only path, so the PR description's claim about internally-created sessions overstates the production effect. Suggestion: adjust the PR description accordingly. (raised by xaxys, tinswzy)

- **`internal/coordinator/mix_coord_test.go:48`** — `TestMixCoordStopOwnsSharedSession` mocks the three children's real `Stop()` methods and `Session.Stop()` into string appends and asserts only the call order, so if a child's real `Stop()` ever re-introduces a direct `session.Stop()` on the shared session, the test still passes. The three `session_ownership_test.go` files are the real regression guards, but they only test each child in isolation with `sessionOwned=true` — the path that never executes in production. Suggestion: add a test that exercises the children's real `Stop()` paths against the shared session. (raised by xaxys, tinswzy)

- **`internal/coordinator/mix_coord.go:423`** — `GracefulStop()` performs `s.streamingCoord.Stop(); s.streamingCoord = nil` without synchronization, while the activation goroutine's `initInternal` reads `s.streamingCoord` and calls `Start` (mix_coord.go:167) with no lock. `GracefulStop` is reachable from both grpc `Server.Stop` and `mixCoordImpl.Stop`, either of which can run concurrently with an in-flight activation — a race-detector-visible data race and possible nil dereference that would kill the process before `Stop()` reaches the final lease revoke. Pre-existing, but this PR's correctness now depends on `Stop()` running to completion, putting it on the fix's critical path. Suggestion: synchronize access to `s.streamingCoord` (e.g. a mutex or atomic swap). (raised by xaxys, tinswzy)

PR #51774, opened 2026-07-23
fix: keep raw fallback cursors consistent

zhengbuqian:
issue: #51489
issue: #51567
issue: #51568

This PR fixes the following issues:

- Corrects predicate rewriting for dynamically typed JSON values: `IN`/`NOT IN`, `== OR`, and `!= AND` are merged only within the same literal type.
- Prevents precision loss in JSON `int64`/floating-point comparisons near `2^53`, including range rewriting and template filling.
- Enables JSON BinaryRange predicates to use JSON path indexes and JSON shredding, with consistent semantics across raw-data, stats, and index paths.
- Rejects BinaryRange predicates with incomparable bounds, including incompatible types, NaN, null, boolean, and array values.
- Determines index-to-raw-data fallback before execution, preventing cursor/offset corruption across multiple batches.
- Fixes missing same-type metadata for templated `json_contains_any/all` expressions and prevents unsupported index execution.
- Improves ARRAY semantics, including empty-array comparisons, non-empty `==`/`!=`, NULL handling, nested offsets, whole-array `IN`/`NOT IN`, and element-type validation.
- Fixes TIMESTAMPTZ arithmetic comparisons when an index cannot provide raw-value lookup.
- Adds proper TermExpr target-type validation so unsupported fields such as vectors and geometry are rejected even with empty or templated RHS values.
- Updates the related Go SDK, E2E, and unit-test expectations.

### DIFF
--- a/internal/core/src/exec/expression/BinaryRangeExpr.cpp
+++ b/internal/core/src/exec/expression/BinaryRangeExpr.cpp
@@ -127,8 +127,15 @@ PhyBinaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                                JsonNumericBoundRequiresPreciseInt64Comparison(
                                    expr_->upper_val_));
 
+            // Keep sparse JsonFlat offset batches candidate-local when raw
+            // JSON is resident.  An index-only segment cannot use the generic
+            // JSON reverse-lookup path, so it must query the typed JsonFlat
+            // executor and gather the requested rows instead.
+            const bool use_json_flat_raw_offsets =
+                exec_path_ == ExprExecPath::ScalarIndex && has_offset_input_ &&
+                PinnedJsonIndexIsFlat() && num_data_chunk_ > 0;
             if (exec_path_ == ExprExecPath::ScalarIndex &&
-                (!has_offset_input_ || !PinnedJsonIndexIsFlat())) {
+                !use_json_flat_raw_offsets) {
                 if (is_numeric) {
                     if (!use_double && PinnedJsonIndexIsFlat()) {
                         result = ExecRangeVisitorImplForIndex<int64_t>(input);
@@ -437,6 +444,10 @@ PhyBinaryRangeFilterExpr::ExecRangeVisitorImplForIndex(OffsetVector* input) {
         return func(index_ptr, val1, val2, lower_inclusive, upper_inclusive);
     };
     if (input != nullptr) {
+        if (PinnedJsonIndexIsFlat()) {
+            return ProcessIndexChunksAndGatherByOffsets<T>(
+                execute_sub_batch, *input, val1, val2);
+        }
         if (cached_result_ == nullptr) {
             auto scalar_index =
                 dynamic_cast<const Index*>(pinned_index_[0].get());
--- a/internal/core/src/exec/expression/BinaryRangeExpr.cpp
+++ b/internal/core/src/exec/expression/BinaryRangeExpr.cpp
@@ -132,10 +132,7 @@ PhyBinaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
 
             if (exec_path_ == ExprExecPath::ScalarIndex && !has_offset_input_) {
                 if (is_numeric) {
-                    if (has_unsafe_int_bound) {
-                        result =
-                            ExecRangeVisitorImplForJsonPreciseNumeric(context);
-                    } else if (!use_double && PinnedJsonIndexIsFlat()) {
+                    if (!use_double && PinnedJsonIndexIsFlat()) {
                         result = ExecRangeVisitorImplForIndex<int64_t>();
                     } else {
                         proto::plan::GenericValue double_lower_val;
@@ -1167,19 +1164,41 @@ PhyBinaryRangeFilterExpr::DetermineExecPath() {
         return;
     }
 
-    SegmentExpr::DetermineExecPath();
-    if (exec_path_ != ExprExecPath::ScalarIndex) {
-        return;
-    }
-
     auto data_type = expr_->column_.data_type_;
     if (expr_->column_.element_level_) {
         data_type = expr_->column_.element_type_;
     }
 
-    // ARRAY type cannot use scalar index
+    if (data_type == DataType::JSON) {
+        const auto lower_type = expr_->lower_val_.val_case();
+        const auto upper_type = expr_->upper_val_.val_case();
+        const auto is_numeric =
+            (lower_type == proto::plan::GenericValue::ValCase::kInt64Val ||
+             lower_type == proto::plan::GenericValue::ValCase::kFloatVal) &&
+            (upper_type == proto::plan::GenericValue::ValCase::kInt64Val ||
+             upper_type == proto::plan::GenericValue::ValCase::kFloatVal);
+        const auto has_unsafe_int_bound =
+            is_numeric &&
+            ((lower_type == proto::plan::GenericValue::ValCase::kInt64Val &&
+              !IsInt64SafeForJsonDoubleIndex(expr_->lower_val_.int64_val())) ||
+             (upper_type == proto::plan::GenericValue::ValCase::kInt64Val &&
+              !IsInt64SafeForJsonDoubleIndex(expr_->upper_val_.int64_val())));
+        if (lower_type == proto::plan::GenericValue::ValCase::kStringVal ||
+            has_unsafe_int_bound) {
+            exec_path_ = ExprExecPath::RawData;
+            return;
+        }
+    }
+
+    // ARRAY type cannot use scalar index.
     if (data_type == DataType::ARRAY) {
         exec_path_ = ExprExecPath::RawData;
+        return;
+    }
+
+    SegmentExpr::DetermineExecPath();
+    if (exec_path_ != ExprExecPath::ScalarIndex) {
+        return;
     }
 }
 
--- a/internal/core/src/exec/expression/BinaryRangeExpr.cpp
+++ b/internal/core/src/exec/expression/BinaryRangeExpr.cpp
@@ -121,19 +121,17 @@ PhyBinaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
                       proto::plan::GenericValue::ValCase::kFloatVal) &&
                  (upper_type == proto::plan::GenericValue::ValCase::kInt64Val ||
                   upper_type == proto::plan::GenericValue::ValCase::kFloatVal));
-            const auto has_unsafe_int_bound =
-                is_numeric &&
-                ((lower_type == proto::plan::GenericValue::ValCase::kInt64Val &&
-                  !IsInt64SafeForJsonDoubleIndex(
-                      expr_->lower_val_.int64_val())) ||
-                 (upper_type == proto::plan::GenericValue::ValCase::kInt64Val &&
-                  !IsInt64SafeForJsonDoubleIndex(
-                      expr_->upper_val_.int64_val())));
+            const auto requires_precise_int64_comparison =
+                is_numeric && (JsonNumericBoundRequiresPreciseInt64Comparison(
+                                   expr_->lower_val_) ||
+                               JsonNumericBoundRequiresPreciseInt64Comparison(
+                                   expr_->upper_val_));
 
-            if (exec_path_ == ExprExecPath::ScalarIndex && !has_offset_input_) {
+            if (exec_path_ == ExprExecPath::ScalarIndex &&
+                (!has_offset_input_ || !PinnedJsonIndexIsFlat())) {
                 if (is_numeric) {
                     if (!use_double && PinnedJsonIndexIsFlat()) {
-                        result = ExecRangeVisitorImplForIndex<int64_t>();
+                        result = ExecRangeVisitorImplForIndex<int64_t>(input);
                     } else {
                         proto::plan::GenericValue double_lower_val;
                         if (lower_type ==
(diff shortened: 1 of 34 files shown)


internal/core/src/exec/expression/TimestamptzArithCompareExpr.h
```diff
@@ -50,6 +50,11 @@ class PhyTimestamptzArithCompareExpr : public SegmentExpr {
     void
     Eval(EvalCtx& context, VectorPtr& result) override;
 
+    void
+    DetermineExecPath() override {
```
liliu-z: `PhyTimestamptzArithCompareExpr::DetermineExecPath()` now forces `exec_path_ = RawData` unconditionally, so `UseIndexCursor()` is always false and the offset-input reverse-lookup branch in `ProcessDataByOffsets` (Expr.h:696, gated on `UseIndexCursor() && num_data_chunk_ == 0`) is never taken. On a sealed TIMESTAMPTZ field with a scalar index whose raw column has been dropped — the default residency when `queryNode.preferFieldDataWhenIndexHasRawData` is false — an offset-input `ts + interval <op> X` filter instead falls through to `segment_->chunk_data<T>` (Expr.h:796), whose `chunk_data_impl` asserts `field_data_ready_bitset` and throws a SegcoreError (not a crash) because the data is absent. This regresses the pre-existing `OffsetsEvalIndexOnlyCorrectnessTest.NullableTimestamptzCallbackHandlesNullData` (test_offsets_eval_correctness.cpp:1034, which calls DropFieldData then evaluates with an OffsetVector); the new SealedScalarIndexStillUsesRawDataPath test leaves field data loaded so it cannot catch this. Retain ScalarIndex when `num_data_chunk_ == 0` so the index reverse-lookup path stays reachable for the dropped-data case.

internal/core/src/exec/expression/UnaryExpr.cpp
```diff
@@ -2023,17 +2019,41 @@ PhyUnaryRangeFilterExpr::DetermineExecPath() {
...
+        return;
+    }
+
+    if (data_type == DataType::ARRAY) {
+        const auto val_case = expr_->val_.val_case();
+        const auto can_use_array_index =
+            val_case == proto::plan::GenericValue::ValCase::kArrayVal &&
+            expr_->val_.array_val().array_size() > 0 &&
+            (expr_->op_type_ == proto::plan::OpType::Equal ||
+             expr_->op_type_ == proto::plan::OpType::NotEqual);
```
buqian-zilliz, line 2102: **[P1] Keep ARRAY NotEqual off the candidate-only index path**

Keeping `NotEqual` eligible here routes it to `ExecArrayEqualForIndex`, which seeds candidates only from rows containing the first distinct literal element and post-filters only that subset. A stored `[3]` evaluated against `[1,2]` never becomes a candidate, so its result remains false even though the arrays are unequal. Restrict this path to `Equal`, or compute equality for all valid rows and complement it while preserving the validity mask.

zhengbuqian: fixed

internal/core/src/exec/expression/BinaryRangeExpr.cpp
```diff
@@ -1148,19 +1171,56 @@ PhyBinaryRangeFilterExpr::DetermineExecPath() {
...
+            (lower_type == proto::plan::GenericValue::ValCase::kInt64Val ||
+             lower_type == proto::plan::GenericValue::ValCase::kFloatVal) &&
+            (upper_type == proto::plan::GenericValue::ValCase::kInt64Val ||
+             upper_type == proto::plan::GenericValue::ValCase::kFloatVal);
+        const auto requires_precise_int64_comparison =
+            is_numeric &&
+            (JsonNumericBoundRequiresPreciseInt64Comparison(
+                 expr_->lower_val_) ||
+             JsonNumericBoundRequiresPreciseInt64Comparison(expr_->upper_val_));
+        if (requires_precise_int64_comparison) {
```
buqian-zilliz, line 1222: **[P2] Preserve STRING_TO_DOUBLE semantics in precise ranges**

This precision gate sends floating-point bounds from 2^53 through 2^63 to RawData before the JSON path index is pinned. A DOUBLE path index configured with STRING_TO_DOUBLE intentionally indexes numeric strings as doubles, but `ExecRangeVisitorImplForJsonPreciseNumeric` calls `at_numeric()` and marks those same strings UNKNOWN. For example, a numeric string at 9007199254740992 stops matching a range around 9007199254740992.0 at this threshold. Keep the cast index for casted rows or make the fallback apply the configured cast; otherwise matching rows are silently lost.

internal/parser/planparserv2/convert_field_data_to_generic_value.go
```diff
@@ -12,6 +12,19 @@ import (
 func convertArrayValue(templateName string, templateValue *schemapb.TemplateArrayValue) (*planpb.GenericValue, error) {
 	var arrayValues []*planpb.GenericValue
 	var elementType schemapb.DataType
+	// An empty SDK list has no element from which to infer a concrete oneof
+	// type. Preserve it as an untyped empty array; consumers with field context
+	// can still interpret its semantics (for example, ARRAY == []).
+	if templateValue != nil && templateValue.GetData() == nil {
```
buqian-zilliz, line 18: **[P2] Validate empty IN templates against the target field**

Accepting every zero-oneof `TemplateArrayValue` as an untyped empty array lets empty placeholders bypass all compatibility checks in `FillTermExpressionValue`, because its casting loop executes zero times. For example, `float_vector IN {values}` with `values=[]` now reaches segcore as an empty `TermExpr`, whose executor has no vector dispatch and fails during execution, while any nonempty list is rejected by the parser. Validate the target field type even when the term list is empty, or restrict this representation to consumers that explicitly support it.

internal/core/src/exec/expression/BinaryRangeExpr.cpp
```diff
@@ -434,6 +430,25 @@ PhyBinaryRangeFilterExpr::ExecRangeVisitorImplForIndex() {
         BinaryRangeIndexFunc<T> func;
         return func(index_ptr, val1, val2, lower_inclusive, upper_inclusive);
     };
+    if (input != nullptr) {
```
buqian-zilliz, line 446: **[P2] Avoid full-index materialization for sparse offsets**

For an offset batch this branch still runs the range query against the entire JSON path index, retains an `active_count`-sized result bitmap plus two full validity bitmaps, and only then gathers the requested rows. Sparse iterative filtering therefore pays segment-wide CPU and memory even when the JSON field or a raw-serving index can evaluate those candidates directly. Keep semantically equivalent raw or reverse-lookup evaluation for sparse offsets, or add an offset-aware index path instead of materializing the full segment.

internal/core/src/exec/expression/BinaryRangeExpr.cpp
```diff
@@ -616,11 +631,11 @@ PhyBinaryRangeFilterExpr::ExecRangeVisitorImplForJson(EvalCtx& context) {
     const auto& bitmap_input = context.get_bitmap_input();
     auto* input = context.get_offset_input();
     FieldId field_id = expr_->column_.field_id_;
-    if (!has_offset_input_ && exec_path_ == ExprExecPath::JsonStats) {
+    if (exec_path_ == ExprExecPath::JsonStats) {
```
buqian-zilliz, line 651: **[P2] Keep sparse offsets out of the full JsonStats scan**

Removing the offset-input guard sends sparse candidate batches through `ExecRangeVisitorImplForJsonStats`. On a cache miss that path allocates two `active_count`-sized bitmaps, evaluates the relevant shredding data and all shared-data hits, and only then gathers the requested offsets. Since JSON stats are enabled by default, a small iterative follow-up can become a full-segment pass for every unique range. Retain candidate-local evaluation when data is available, or implement an offset-aware stats lookup.

zhengbuqian: not planning for this PR, tracked in #51832

internal/core/src/exec/expression/BinaryRangeExpr.cpp
zhengbuqian: not planning for this PR, tracked in #51832

internal/core/src/exec/expression/BinaryRangeExpr.cpp
zhengbuqian: not planning for this PR, tracked in #51833

internal/core/src/exec/expression/BinaryRangeExpr.cpp
```diff
@@ -121,22 +121,17 @@ PhyBinaryRangeFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
...
-
-            if (exec_path_ == ExprExecPath::ScalarIndex && !has_offset_input_) {
+            const auto requires_precise_int64_comparison =
+                is_numeric && (JsonNumericBoundRequiresPreciseInt64Comparison(
+                                   expr_->lower_val_) ||
+                               JsonNumericBoundRequiresPreciseInt64Comparison(
+                                   expr_->upper_val_));
+
+            if (exec_path_ == ExprExecPath::ScalarIndex &&
+                (!has_offset_input_ || !PinnedJsonIndexIsFlat())) {
```
buqian-zilliz: **[P1] Keep offset-input JsonFlat ranges off the raw reverse-lookup path**

The new condition routes `ScalarIndex + offset input + JsonFlatIndex` into the raw JSON branch. On a sealed segment whose raw JSON data has been dropped, `ProcessDataByOffsets<Json>` enters `ProcessIndexLookupByOffsets<Json>`, casts the pinned `JsonFlatIndex` to `ScalarIndex<Json>`, and dereferences the resulting null pointer while loading validity. A follow-up binary-range filter can therefore abort on an index-only segment. Keep this case on a JsonFlat-aware index/gather path, or only choose raw evaluation when field data is present.

internal/parser/planparserv2/convert_field_data_to_generic_value.go
buqian-zilliz, line 18: **[P2] Do not admit nested empty ARRAY values into TermExpr**

This zero-oneof conversion is recursive, so an ARRAY term template can now contain multiple untyped empty arrays. For an ARRAY field, `arr IN {values}` with `values = [[], []]` passes `FillTermExpressionValue` as two `kArrayVal` entries and remains a multi-value `TermExpr`; the C++ ARRAY term dispatcher handles only bool/int/float/string values and throws on `kArrayVal`. Reject this template shape or lower whole-array term values to equality predicates before segcore.

internal/parser/planparserv2/utils.go
```diff
@@ -744,6 +745,112 @@ func castRangeValue(dataType schemapb.DataType, value *planpb.GenericValue) (*pl
...
+		return 0, true
+	}
+
+	return 0, false
+}
+
+func validateBinaryRangeBounds(lower, upper *planpb.GenericValue, lowerInclusive, upperInclusive bool) error {
+	cmp, comparable := compareBinaryRangeBounds(lower, upper)
+	if !comparable {
+		return nil
```
zhengbuqian: why return error == nil if not comparable?

zhengbuqian: seems to me comparable can be false only when:

1. number comparison and any is float NaN 2. string comparison and any is not string 3. any of upper/lower is not number/string

then can't we simply reject such expressions? 

if so we can simplify a lot?

internal/parser/planparserv2/utils.go
```diff
@@ -744,6 +745,112 @@ func castRangeValue(dataType schemapb.DataType, value *planpb.GenericValue) (*pl
 	return value, nil
 }
 
+// compareInt64ToFloat64 compares an int64 and a float64 without converting the
+// integer to float64. It returns whether the values are comparable separately
+// so NaN can be left to execution-time predicate semantics.
+func compareInt64ToFloat64(lhs int64, rhs float64) (int, bool) {
```
zhengbuqian: we have a very similar method in rewriter/range.go

internal/parser/planparserv2/utils.go
aoiasd: Confirmed that Segcore only supports mixed numeric bounds (int64/float), which are unified under one numeric executor. It does not support cross-family bounds such as number/string: the scalar-index path rejects them, while RawData/JsonStats select the executor from the lower bound and fail when converting the upper bound. Therefore 1 <= A < "z" should be rejected here instead of being deferred to execution, and the corresponding valid-expression tests should be removed or changed.

internal/core/src/exec/expression/BinaryRangeExpr.cpp
```diff
@@ -761,11 +787,15 @@ PhyBinaryRangeFilterExpr::ExecRangeVisitorImplForJson(EvalCtx& context) {
...
+PhyBinaryRangeFilterExpr::ExecRangeVisitorImplForJsonStats(
+    OffsetVector* input) {
     using GetType = std::conditional_t<std::is_same_v<ValueType, std::string>,
                                        std::string_view,
                                        ValueType>;
-    auto real_batch_size = GetNextBatchSize();
+    auto real_batch_size =
+        input != nullptr
+            ? input->size()
+            : std::min(batch_size_, active_count_ - current_data_global_pos_);
```
zhagnlu, line 798: wrap as a interface, used for offset/indexed/stats type.

zhagnlu: actually, GetNextBatchSize() add the jsonStats switch case

internal/core/src/exec/expression/Expr.h
zhagnlu, line 444: Can we use exec_path_ instead of UseIndexCursor ?

Issue #51776, opened 2026-07-23
[Bug]: QueryNode can skip segment cleanup after release wait timeout


bigsheeper:
### Is there an existing issue for this?

- [x] I have searched the existing issues

Search terms checked: `MaxWLockConditionalWaitTime`, `load state lock wait timeout`, `StartReleaseAll`, `ReleaseSegments`, and `C.DeleteSegment`.

### Environment

```markdown
- Milvus version: master f113bdbc8e39efa8821b7b3f7151287ce1bd0b9c
- Deployment mode: any QueryNode deployment
- MQ type: N/A
- SDK version: N/A
- OS: Linux
```

### Current Behavior

`LoadStateLock.waitOrPanic` returns after `common.locks.maxWLockConditionalWaitTime` even when the release predicate is still false. `LocalSegment.Release` treats the resulting nil guard as "release is already done" and returns without deleting the native segcore segment, refunding the PK candidate, or clearing loaded binlog accounting.

The manager path then continues as if release succeeded and removes the segment from the on-releasing set, hiding the leak.

Production path:

```text
QueryCoord ReleaseSegments
  -> QueryNode.ReleaseSegments
  -> segmentManager.Remove / Clear
  -> segmentManager.release
  -> LocalSegment.Release
  -> LoadStateLock.StartReleaseAll
  -> waitOrPanic timeout returns before refCnt reaches 0
```

A normal conflicting reader can hold the ref count through search/retrieve:

```text
QueryNode.SearchSegments / QuerySegments
  -> segments.SearchHistorical / Retrieve
  -> validateOnHistorical
  -> Segment.PinIfNotReleased
  -> segcore search/retrieve returns later
```

### Expected Behavior

A release wait timeout should log the timeout, but it must not make `StartReleaseAll` return before the segment becomes releasable and the release state transition is installed. Otherwise callers cannot distinguish "release already done" from "release abandoned after timeout".

### Steps To Reproduce

Add this regression test:

```go
func TestStartReleaseAllWaitsAfterTimeoutUntilRefCntZero(t *testing.T) {
    paramtable.Init()
    paramtable.Get().Save(paramtable.Get().CommonCfg.MaxWLockConditionalWaitTime.Key, "0.05")
    defer paramtable.Get().Reset(paramtable.Get().CommonCfg.MaxWLockConditionalWaitTime.Key)

    l := NewLoadStateLock(LoadStateDataLoaded)
    assert.True(t, l.PinIfNotReleased())

[...]

PR #51777, opened 2026-07-23
fix: wait for segment release after lock timeout

bigsheeper:
issue: #51776

## What this PR changes

`LoadStateLock.waitOrPanic` no longer returns when `common.locks.maxWLockConditionalWaitTime` elapses while the predicate is still false. It now keeps the existing timeout log as a watchdog, but continues waiting until the guarded state transition can run.

This prevents `LocalSegment.Release` from receiving a nil guard after a timeout and treating a still-pinned segment as already released, which skipped native segment deletion and resource cleanup while manager bookkeeping continued.

## Tests

- Repro before fix:
  - `GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/segments/state -run TestStartReleaseAllWaitsAfterTimeoutUntilRefCntZero`
  - failed with `StartReleaseAll returned before refcnt dropped to zero`
- After fix:
  - `GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/segments/state`
  - passed
- Static check:
  - `GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage make static-check`
  - passed with `0 issues`

## Notes

The broader `./internal/querynodev2/segments` package test is still blocked in this local worktree by C++ link dependencies (`libgrpc++.so.1.67` / related grpc symbols from `libmilvus-common.so`), even when pkg-config is pointed at the main checkout artifacts.

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -1118,7 +1118,7 @@ common:
     threshold:
       info: 500 # minimum milliseconds for printing durations in info level
       warn: 1000 # minimum milliseconds for printing durations in warn level
-    maxWLockConditionalWaitTime: 600 # maximum seconds for waiting wlock conditional
+    maxWLockConditionalWaitTime: 600 # seconds before logging a wlock conditional wait that is taking long; the wait itself is not bounded by this value
   traceLogMode: 0 # trace request info
   visibilityFilterEnabled: true # whether to apply row visibility filtering (timestamp, delete, and TTL) on querynode. When disabled, all rows are returned regardless of insert/delete timestamps.
   bloomFilterEnabled: true # whether to load and apply bloom filter/pk stats on querynode
--- a/internal/querynodev2/segments/state/load_state_lock.go
+++ b/internal/querynodev2/segments/state/load_state_lock.go
@@ -189,23 +189,18 @@ func (ls *LoadStateLock) canReleaseAll(state loadStateEnum) bool {
 }
 
 func (ls *LoadStateLock) waitOrPanic(ready func(state loadStateEnum) bool, then func()) {
-	ch := make(chan struct{})
 	maxWaitTime := paramtable.Get().CommonCfg.MaxWLockConditionalWaitTime.GetAsDuration(time.Second)
-	go func() {
-		ls.cv.L.Lock()
-		defer ls.cv.L.Unlock()
-		defer close(ch)
-		for !ready(ls.state) {
-			ls.cv.Wait()
-		}
-		then()
-	}()
-
-	select {
-	case <-time.After(maxWaitTime):
+	timer := time.AfterFunc(maxWaitTime, func() {
 		mlog.Error(context.TODO(), "load state lock wait timeout", mlog.Duration("maxWaitTime", maxWaitTime))
-	case <-ch:
+	})
+	defer timer.Stop()
+
+	ls.cv.L.Lock()
+	defer ls.cv.L.Unlock()
+	for !ready(ls.state) {
+		ls.cv.Wait()
 	}
+	then()
 }
 
 type StatePredicate func(state loadStateEnum) bool
--- a/internal/querynodev2/segments/state/load_state_lock.go
+++ b/internal/querynodev2/segments/state/load_state_lock.go
@@ -169,13 +169,13 @@ func (ls *LoadStateLock) StartReleaseAll() (g LoadStateLockGuard) {
 	return g
 }
 
-// blockUntilDataLoadedOrReleased blocks until the segment is loaded or released.
-func (ls *LoadStateLock) BlockUntilDataLoadedOrReleased() bool {
-	var ok bool
+// BlockUntilDataLoadedOrReleased blocks until the segment is loaded or released.
+// It has no return value on purpose: waitOrPanic no longer gives up, so any
+// status it could report would be a constant.
+func (ls *LoadStateLock) BlockUntilDataLoadedOrReleased() {
 	ls.waitOrPanic(func(state loadStateEnum) bool {
 		return state == LoadStateDataLoaded || state == LoadStateReleased
-	}, func() { ok = true })
-	return ok
+	}, func() {})
 }
 
 // waitUntilCanReleaseData waits until segment is release data able.
@@ -190,8 +190,12 @@ func (ls *LoadStateLock) canReleaseAll(state loadStateEnum) bool {
 
 func (ls *LoadStateLock) waitOrPanic(ready func(state loadStateEnum) bool, then func()) {
 	maxWaitTime := paramtable.Get().CommonCfg.MaxWLockConditionalWaitTime.GetAsDuration(time.Second)
+	// Watchdog only: the wait below is deliberately unbounded so that then() always
+	// runs and native cleanup is never skipped. This fires once, purely to surface
+	// a release that is taking longer than expected.
 	timer := time.AfterFunc(maxWaitTime, func() {
-		mlog.Error(context.TODO(), "load state lock wait timeout", mlog.Duration("maxWaitTime", maxWaitTime))
+		mlog.Warn(context.TODO(), "load state lock still waiting, the wait is not bounded and continues until the state is ready",
+			mlog.Duration("maxWaitTime", maxWaitTime))
 	})
 	defer timer.Stop()
 
--- a/internal/querynodev2/segments/state/load_state_lock_test.go
+++ b/internal/querynodev2/segments/state/load_state_lock_test.go
@@ -189,6 +189,42 @@ func TestStartReleaseAll(t *testing.T) {
 	assert.Equal(t, LoadStateReleased, l.state)
 }
 
+func TestStartReleaseAllWaitsAfterTimeoutUntilRefCntZero(t *testing.T) {
+	paramtable.Init()
+	paramtable.Get().Save(paramtable.Get().CommonCfg.MaxWLockConditionalWaitTime.Key, "0.05")
+	defer paramtable.Get().Reset(paramtable.Get().CommonCfg.MaxWLockConditionalWaitTime.Key)
+
+	l := NewLoadStateLock(LoadStateDataLoaded)
+	assert.True(t, l.PinIfNotReleased())
+
+	ch := make(chan LoadStateLockGuard, 1)
+	go func() {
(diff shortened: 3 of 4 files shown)


czs007:

**Verified:**

- `internal/querynodev2/segments/state/load_state_lock.go:198-203` — no lost wakeup: the waiter holds the write lock continuously across the predicate check and into `cv.Wait()` (which releases it atomically), while `Unpin` takes the read lock to `Dec()`+`Broadcast()`, so a decrement cannot interleave between predicate evaluation and the wait; the re-check after wake is correct.
- `internal/querynodev2/segments/state/load_state_lock.go:200-203` — the original defect is genuinely fixed: `then()` now runs synchronously in the caller's goroutine under the lock, so the named return guard is assigned before `waitOrPanic` returns, eliminating the old race where a post-timeout `then()` left `Release` with a `nil` guard and skipped `C.DeleteSegment` while `refCnt > 0`; the previously leaked goroutine parked in `cv.Wait()` is also gone.
- `internal/querynodev2/segments/state/load_state_lock.go:182-189` — the new blocking behavior is correctly scoped: `canReleaseData` does not consult `refCnt`, so only `canReleaseAll` gates on `refCnt == 0` and the `ReleaseScopeData` path is unaffected.
- `internal/querynodev2/segments/state/load_state_lock.go:161-163` and `internal/querynodev2/segments/segment.go:1377-1379` — `nil`-guard semantics are preserved: a legitimate double release still returns `nil` for an already-`LoadStateReleased` segment and `Release` early-returns, with concurrent `StartReleaseAll` callers serialized on the write lock.
- `internal/querynodev2/segments/manager.go:677-683,737-757` — blast radius of a slow release is bounded: `Remove`/`RemoveBy` drop the segment from the fine-grained `ConcurrentMap` before calling `mgr.release`, so only the releasing goroutine waits rather than the whole segment map.
- `internal/querynodev2/segments/state/load_state_lock_test.go:192-227,247-279` — tests genuinely cover the new semantics: each asserts non-return for 200ms *after* the 50ms watchdog elapses, then asserts return once `refCnt`/state changes.

czs007:
## Re-review of `7e2bce2e2521`

The commit is unchanged since our last review; all four previously posted findings still stand as non-blocking `low` items, and this round's independent pass found no new critical or high-severity issues.

- **`internal/querynodev2/segments/state/load_state_lock.go:198` — `waitOrPanic` lost its bounded exit: Confirmed.** The code is unchanged — the wait is still an unconditional `for !ready(ls.state) { ls.cv.Wait() }` with no cancellation path, and `canReleaseAll` still gates on `refCnt.Load() == 0`, so a leaked pin blocks the release goroutine indefinitely. Details and updated blast radius below.
- **`internal/querynodev2/segments/state/load_state_lock.go:193` — watchdog log is Error-level, one-shot, and still worded as a timeout: Confirmed.** `time.AfterFunc` fires exactly once at `Error` level saying "wait timeout" even though nothing times out and the wait continues, so a slow-but-successful release emits a false error while a permanently stuck one emits one line and then goes silent — the two are indistinguishable.
- **`pkg/util/paramtable/component_param.go:1086` — `common.locks.maxWLockConditionalWaitTime` doc still advertises a wait cap: Confirmed.** Its only non-test consumer is `load_state_lock.go:192`, where it now controls solely the watchdog log delay; the exported doc and the `configs/milvus.yaml:1113` comment still describe it as a maximum wait, so an operator lowering it to bound release latency gets no behavioral change.
- **`internal/querynodev2/segments/state/load_state_lock.go:173` — `BlockUntilDataLoadedOrReleased` bool return is always true: Confirmed.** `ok` is set only inside `then()`, and `then()` now always runs once the function returns normally, making the value constant; there is no production caller (only `load_state_lock_test.go:115`, which ignores it).

## Findings this round

### Low

**`internal/querynodev2/segments/state/load_state_lock.go:198` — `waitOrPanic` has no bounded exit, so an unreturned pin blocks the releaser forever and stalls the `ReleaseSegments` RPC and the delegator's `growingSegmentLock`.**
`waitOrPanic` (191-204) is now just `for !ready(ls.state) { ls.cv.Wait() }`; the `time.AfterFunc` only logs and provides no cancellation. Since `canReleaseAll` requires `refCnt.Load() == 0` (`load_state_lock.go:188`) and pin ownership is handed across functions — `GetAndPin`/`GetAndPinBy` return an already-pinned segment and rely on the caller to `mgr.Unpin` (`manager.go:514-539`, `541-600`, `602-606`), with the pin held for the whole cgo search/query — any pin that is not returned blocks the release permanently. Failure: a growing segment's pin is not returned in time (a `GetAndPin`ned search stuck inside cgo, or an error path that misses `mgr.Unpin`) while querycoord issues `ReleaseSegments` for that segment → `waitOrPanic` blocks forever in `cv.Wait`, so the `ReleaseSegments` handler — which calls `Remove` synchronously in a loop (`services.go:703-709`) — never returns, the remaining segment IDs in that request are never released and `Collection.Unref` never runs; on the growing path the delegator holds `sd.growingSegmentLock` across `worker.ReleaseSegments` (`delegator_data.go:1254-1282`), so `ProcessInsert` blocks on the same lock when creating a new growing segment (`delegator_data.go:208`) and the shard's write pipeline stalls; meanwhile `manager.go:796-810` never runs, leaving the `growingOnReleasingSegments`/`sealedOnReleasingSegments` entry in place so `Exist()` stays true and `QueryNodeNumSegments` never decrements.
Scope note: `Remove`/`RemoveBy` do the map removal before `mgr.release` (`manager.go:677-683`, `754-756`), so a stuck release does not additionally hold the segment map.
Suggestion: this is an acceptable, deliberate trade-off against the previous "timeout → nil guard → skip native cleanup" leak, so it need not block this PR, but please follow up with a bounded cancellation path (e.g. context- or deadline-aware wait). Any such mechanism must leave the segment in a pending-release state on abort — it must not go back to treating a nil guard as a successful release. Pairing it with periodic (rather than one-shot) stuck-release reporting or a metric would also make the new failure mode observable. *(raised by tinswzy, chyezh)*

bigsheeper:
Thanks — all four confirmed. Three of them are this PR's own debt and are fixed in `477f1d2eb8` (which also rebases onto latest master, 25 commits behind). The fourth I'm leaving as the deliberate trade-off, per your own note.

**Fixed:**

- **Watchdog wording and level (`:193`).** Now `Warn`, and the message says the wait continues rather than claiming a timeout. Your framing of why this matters is the part I'd underline: a slow-but-successful release and a permanently stuck one were emitting the *same single Error line*, so the log couldn't distinguish them.
- **`common.locks.maxWLockConditionalWaitTime` doc.** Reworded in both `component_param.go` and `configs/milvus.yaml`. This was the one I considered most worth fixing here — it's an operator-facing promise this PR silently invalidated: lowering the value to bound release latency does nothing now. The yaml text is exactly what `make generate-yaml` emits; I regenerated and diffed to confirm, since a hand-written comment that drifts from the param doc breaks the config check.
- **`BlockUntilDataLoadedOrReleased` bool (`:173`).** Dropped the return value. You're right that this PR is what made it constant — before the change the timeout path returned false. No production caller exists (only `load_state_lock_test.go:115`, which ignores it), so removing it was cleaner than documenting a constant that reads like a status.

**Not fixed — the unbounded wait (`:198`).** I agree with your classification. The new behavior is strictly better than what it replaced: the old code could return with a nil guard and skip `C.DeleteSegment` while `refCnt > 0`, leaking native memory; the new one can only block, loudly. Blocking is the better failure.

I verified the blast radius you described and it holds: `services.go:704-710` removes synchronously in a loop, so a stuck release strands the remaining segment IDs in that request and `Collection.Unref` never runs; the delegator holds `sd.growingSegmentLock` across `worker.ReleaseSegments`, and `ProcessInsert` takes the same lock when creating a growing segment, so the shard's write path stalls; and `manager.go:796-810` never runs, leaving the metric undecremented and `Exist()` true.

Because that cost is real, I put it in the commit message rather than leaving it implicit — a future reader should see why the bounded-cancellation follow-up matters. I'm not attempting that path here, and I've noted your constraint on it: an aborted wait must leave the segment in a pending-release state, never revert to treating a nil guard as success.

**One process note.** These two rounds disagree. The first (03:20) posted "Adversarial review found no issues requiring changes" with six specific verifications; the second (04:40) reports "four previously posted findings still stand" — but those four had never appeared on this PR. Whatever produced the first verdict missed everything the second round found. Flagging it because a clean verdict is being treated as a merge signal, and here it wasn't a reliable one.

czs007:

**Verified:**
- `pkg/util/typeutil/load_state_lock.go:191-208` — the new inline unbounded wait guarantees `then()` always runs, so `StartReleaseAll` can no longer return a nil guard and `LocalSegment.Release` can no longer skip native `DeleteSegment`/`pkCandidate.Refund()`/`SubLoadedBinlogSize`.
- `pkg/util/typeutil/load_state_lock.go:175` — dropping the `bool` return from `BlockUntilDataLoadedOrReleased` is safe: a repo-wide grep shows the only references are `load_state_lock_test.go:115` and the internal `StartReleaseData`/`StartReleaseAll`, none of which read the removed value.
- `pkg/util/typeutil/load_state_lock.go:130,151,176` — all three `waitOrPanic` callers are internal and re-verified; no external caller depends on the old timeout-returns behavior.
- `pkg/util/typeutil/load_state_lock.go:202-207` — lock/wakeup semantics are unchanged: the waiter holds `cv.L` across the `ready` check and `cv.Wait()`, so `Unpin`'s `Broadcast` cannot interleave and no wakeup is lost.
- `pkg/util/typeutil/load_state_lock.go:196-200` — the watchdog is side-effect free: `time.AfterFunc` only logs, `defer timer.Stop()` cancels it on the fast path, and a misconfigured `maxWaitTime <= 0` yields at most one benign warning.
- `pkg/util/typeutil/load_state_lock_test.go:192-223,247-282` — both new tests (blocked past the 50ms watchdog until unpin, and `then()` firing only after the predicate holds) would fail against the old timeout-returns path.

czs007 (approved):
/lgtm

PR #51779, opened 2026-07-23
fix: key proxy shard leader cache by collection id to close alias repoint race

xiaofan-luan:
## Summary

Re-key the proxy shard leader cache by the cluster-unique collection id instead of by `(database, collection name)`, closing the concurrent-`AlterAlias` stale-alias routing race and making shard-cache invalidation O(1).

## Root cause

`shardClientMgrImpl.collLeader` was keyed `map[database]map[collectionName]*shardLeaders`, and `GetShard` / `GetShardLeaderList` never checked the cached entry's `collectionID` against the id resolved for the current request. After `AlterAlias` repoints an alias to a new collection:

1. an in-flight fill for the old target re-inserts the old collection's shard leaders under the alias name (after the invalidation that raced it), and
2. a later request that resolves the alias to the **new** id hits that stale entry and is served the **old** collection's channels and leaders (both come from the same alias-keyed entry).

The result is queries on the alias silently routed to the old collection until the entry ages out.

## Changes

- `collLeader` is now `map[int64]*shardLeaders` (collection id is cluster-unique, so database and name drop out of the key). Name/alias/database resolution stays upstream in the meta cache; the shard cache only ever sees a resolved id.
- `InvalidateShardLeaderCache([]int64)` becomes an O(len(ids)) direct delete instead of a full-cache scan under the write lock.
- The cache deliberately does **not** track database membership. `collection -> database` is mutable (cross-db `RenameCollection` keeps the id), so keying/indexing on it would let a stale attribution drop a live collection or leak one that moved. `RemoveDatabase` is a no-op: `DropDatabase` requires an empty database, so its collections were already dropped and evicted by id first.
- `DeprecateShardCache` is removed and folded into `InvalidateShardLeaderCache`; the search/query/statistic retry paths and the `InvalidateCollectionMetaCache` handler now evict by collection id.

## Test plan

- [x] `TestShardCacheAliasRepointResolvesByCollectionID` fails on the old name-keyed cache (served the old collection's channel) and passes after the re-key.
- [x] `TestRemoveDatabase` fails on a db-indexed version (evicts a live id by stale db attribution) and passes with the id-keyed no-op.
- [x] Migrated `TestInvalidateShardLeaderCache`; `internal/proxy/shardclient` full package suite green; `go build ./internal/proxy/...` clean.
- [ ] Full `internal/proxy` suite and static-check via CI.

## Follow-ups

- Same-id stale-fill after a shard-leader balance still self-heals via the not-shard-leader retry; a fill generation gate to drop late fills is a follow-up (also closes the rare dropped-collection late-fill straggler, memory-only).

issue: #51533

https://claude.ai/code/session_017QZmbdsWhTJLB4Nn5yAxdG

### DIFF
--- a/internal/proxy/shardclient/manager.go
+++ b/internal/proxy/shardclient/manager.go
@@ -41,7 +41,6 @@ import (
 type ShardClientMgr interface {
 	GetShard(ctx context.Context, withCache bool, database, collectionName string, collectionID int64, channel string) ([]NodeInfo, error)
 	GetShardLeaderList(ctx context.Context, database, collectionName string, collectionID int64, withCache bool) ([]string, error)
-	DeprecateShardCache(database, collectionName string)
 	InvalidateShardLeaderCache(collections []int64)
 	ListShardLocation() map[int64]NodeInfo
 	RemoveDatabase(database string)
@@ -63,8 +62,13 @@ type shardClientMgrImpl struct {
 
 	mixCoord types.MixCoordClient
 
-	leaderMut  sync.RWMutex
-	collLeader map[string]map[string]*shardLeaders // database -> collectionName -> collection_leaders
+	leaderMut sync.RWMutex
+	// collLeader keys shard leaders by the cluster-unique collection id, so name/alias/database
+	// resolution (done upstream against the meta cache) can never serve one collection's shard
+	// leaders under another's name after an alias repoint or a cross-db rename. Eviction is by
+	// collection id only -- the cache deliberately does not depend on the mutable
+	// collection->database mapping. See issue #51533.
+	collLeader map[int64]*shardLeaders // collectionID -> collection_leaders
 }
 
 const (
@@ -92,7 +96,7 @@ func NewShardClientMgr(mixCoord types.MixCoordClient, options ...shardClientMgrO
 		purgeInterval:   defaultPurgeInterval,
 		expiredDuration: defaultExpiredDuration,
 
-		collLeader: make(map[string]map[string]*shardLeaders),
+		collLeader: make(map[int64]*shardLeaders),
 		mixCoord:   mixCoord,
 	}
 	for _, opt := range options {
@@ -109,7 +113,7 @@ func (c *shardClientMgrImpl) SetClientCreatorFunc(creator queryNodeCreatorFunc) 
 func (m *shardClientMgrImpl) GetShard(ctx context.Context, withCache bool, database, collectionName string, collectionID int64, channel string) ([]NodeInfo, error) {
 	method := "GetShard"
 	// check cache first
-	cacheShardLeaders := m.getCachedShardLeaders(database, collectionName, method)
+	cacheShardLeaders := m.getCachedShardLeaders(collectionID, method)
 	if cacheShardLeaders == nil || !withCache {
 		// refresh shard leader cache
 		newShardLeaders, err := m.updateShardLocationCache(ctx, database, collectionName, collectionID)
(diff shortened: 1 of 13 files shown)


xiaofan-luan:
Addressed the review (P2 blocking + P3). Force-pushed: master `368f10ec0d`, cp3.0 #51780 `b943608f54`. Shard-client package suite is green on both branches.

**P2 — the `dbCollections` / `entry.database` secondary index could go stale under cross-db rename.** This is correct, and it's the same mutable-key trap this PR removes for the collection name, reintroduced for the database. Rather than patch the index's consistency, I removed it, because it turned out to be redundant for correctness:

- Cross-db `RenameCollection` keeps the collection id, and its `InvalidateCollectionMetaCache` broadcast carries the real collection id, so the proxy already evicts the entry **by id** on rename (`impl.go` → `InvalidateShardLeaderCache([id])`).
- `DropDatabase` requires an **empty** database — rootcoord rejects a non-empty drop (`meta_table.go`: "not empty, must drop all collections before drop database"). So every collection is dropped individually and evicted by id first; by the time `RemoveDatabase` runs there is nothing left to clean.

So the shard cache never needs to track database membership. Changes:
- Removed `dbCollections` and `shardLeaders.database`.
- `RemoveDatabase` is now a documented no-op.
- This also removes the `RemoveDatabase` O(N-in-db)-under-write-lock cost that the index had introduced.
- Added `TestRemoveDatabase` that fails on the index version (it evicted a live id by stale db attribution) and passes now.

**P3 — README** structure descriptions updated to the id-keyed layout.

**On the fill generation gate:** the remaining case is the pre-existing Class-2 late-fill (a stale fill landing after an invalidation) — the original name-keyed cache had it too. With the index gone, the cross-db-rename scenario now yields a **correct live entry**, not a leak. The only residual is a dropped-collection late-fill straggler (memory-only, never served since its id no longer resolves). I'm keeping the fill gate as a follow-up, but happy to fold it into this PR if you'd prefer.

internal/proxy/shardclient/manager.go
```diff
@@ -207,10 +205,7 @@ func (m *shardClientMgrImpl) updateShardLocationCache(ctx context.Context, datab
 	}
 
 	m.leaderMut.Lock()
-	if _, ok := m.collLeader[database]; !ok {
-		m.collLeader[database] = make(map[string]*shardLeaders)
-	}
-	m.collLeader[database][collectionName] = newShardLeaders
+	m.collLeader[collectionID] = newShardLeaders
```
liliu-z, line 208: One small follow-up: a `GetShardLeaders` RPC that started before invalidation can still land here afterward and repopulate this ID. This is pre-existing, but with `RemoveDatabase` now a no-op it may leave a dropped collection entry around and keep its QN client from being purged. Could we fence this write with a small per-ID fill token/generation?

liliu-z (commented):
/lgtm
/approve

internal/proxy/shardclient/manager.go
xiaofan-luan: Good catch, and agreed on the analysis. This is the pre-existing late-fill-vs-invalidate race, and you're right that keying by collection id changes the reclaim profile: the old `(db, name)` key would get overwritten by a same-name recreate, whereas a dropped id is never reused, so a late fill that lands after invalidation leaves the entry behind — and with `RemoveDatabase` now a no-op, nothing sweeps it, so its QN client isn't purged either.

The impact is memory-only though, not a correctness issue: a dropped collection's id is never resolved by the meta cache, so the entry is never served — it just holds a little memory and a QN client ref until the process restarts. Given the low impact I'd prefer to leave this as a follow-up rather than add the fence in this PR. The per-ID fill token/generation you suggested is exactly the right fix for that follow-up (a TTL on the shard-leader cache would also reclaim it).

PR #51780, opened 2026-07-23
fix: [cp3.0] key proxy shard leader cache by collection id to close alias repoint race

xiaofan-luan:
Cherry-pick from master.
pr: #51779

## Summary

Backport of the proxy shard leader cache re-key to the cluster-unique collection id, closing the concurrent-`AlterAlias` stale-alias routing race and making shard-cache invalidation O(1). 3.0's `shardclient` package matches the master baseline, so this is a clean cherry-pick.

## Root cause

`shardClientMgrImpl.collLeader` was keyed `map[database]map[collectionName]*shardLeaders`, and `GetShard` / `GetShardLeaderList` never validated the cached entry's `collectionID` against the id resolved for the request. After `AlterAlias` repoints an alias to a new collection, an in-flight fill re-inserts the old collection's shard leaders under the alias name, and a later request resolving the alias to the new id is served the old collection's channels and leaders (both from the same alias-keyed entry).

## Changes

- `collLeader` is now `map[int64]*shardLeaders` (collection id is cluster-unique; database and name drop out of the key).
- `InvalidateShardLeaderCache([]int64)` becomes an O(len(ids)) delete instead of a full-cache scan under the write lock.
- The cache does not track database membership (`collection -> database` is mutable via cross-db `RenameCollection`). `RemoveDatabase` is a no-op: `DropDatabase` requires an empty database, so its collections were already dropped and evicted by id.
- `DeprecateShardCache` is folded into `InvalidateShardLeaderCache`; the search/query/statistic retry paths and the `InvalidateCollectionMetaCache` handler evict by collection id.

## Test plan

- [x] `internal/proxy/shardclient` full package suite green on 3.0, including `TestShardCacheAliasRepointResolvesByCollectionID` (regression), `TestRemoveDatabase`, and `TestInvalidateShardLeaderCache`.
- [x] `go build ./internal/proxy/shardclient/...` clean; caller edits applied cleanly with no remaining `DeprecateShardCache` references.
- [ ] Full `internal/proxy` suite and static-check via 3.0 CI.

## Follow-ups

- Fill generation gate to drop late fills (same-id stale-fill after balance self-heals via not-shard-leader retry today).

issue: #51533

https://claude.ai/code/session_017QZmbdsWhTJLB4Nn5yAxdG

### DIFF
--- a/internal/proxy/impl.go
+++ b/internal/proxy/impl.go
@@ -139,17 +139,15 @@ func (node *Proxy) InvalidateCollectionMetaCache(ctx context.Context, request *p
 	collectionID := request.CollectionID
 	msgType := request.GetBase().GetMsgType()
 	var aliasName []string
+	// The shard leader cache is keyed by the cluster-unique collection id (issue #51533), so an
+	// alias/collection name is no longer a cache key. Evicting the affected collection id covers
+	// every name/alias that resolves to it; alias repoints and renames don't move a collection's
+	// shard leaders, so for those this is a harmless best-effort refresh. The names argument is
+	// kept only so the call sites (which still pass alias names for the meta-cache path) are
+	// untouched.
 	deprecateShardCaches := func(names ...string) {
-		seen := make(map[string]struct{}, len(names))
-		for _, name := range names {
-			if name == "" {
-				continue
-			}
-			if _, ok := seen[name]; ok {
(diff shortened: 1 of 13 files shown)


liliu-z (commented):
/lgtm
/approve

PR #51785, opened 2026-07-23
enhance: add global schema cache and per-segment ownership on QueryNode

xiaofan-luan:
Tracking issue: #51796

## Summary

This PR makes QueryNode schema ownership explicit and shared:

- Segcore has one process-wide `SchemaCache`. Collections and segments keep the returned `SchemaPtr`, and the cache entry is removed when the last reference is released.
- QueryCoord sends the immutable logical schema separately from its effective mmap/warmup load policy. New QueryNodes cache only the logical schema; the request-level effective schema remains for older QueryNodes during rolling upgrade.
- QueryNode publishes logical schema, compact load policy, partial-load fields, barrier timestamp, and the internal segcore revision as one immutable collection snapshot. Segment creation and reopen consume one snapshot.
- Sealed segments retain parsed mmap/warmup policy after the wire proto is compacted. Growing segments keep their cached logical schema reference but do not retain another load-policy schema.
- Partial-load membership uses a field-ID set instead of scanning `load_fields` for every field.

`schema.Version` remains the persisted logical/materialization version. A newer barrier may refresh same-version runtime properties such as TTL without changing that logical version. QueryNode uses a collection-local segcore revision only to publish those runtime snapshots; it is not a DataCoord/index version and is not reported as one.

## Correctness covered

- First sealed load preserves the mmap/warmup policy parsed before `load_schema` is cleared.
- Reloaded collection metadata updates `load_fields` and load policy together.
- Create and reopen cannot combine schema and policy from different snapshots.
- QueryCoord policy decoration cannot create two different global-cache payloads for the same logical schema version.
- Property-only TTL/external updates preserve master behavior and do not advance the version used by DataCoord index/backfill fences.
- Cache entries are reclaimed after the final collection/segment schema reference is released.

## Verification

Local darwin/arm64 verification on the pushed head:

- `make build-cpp-with-unittest`: passed.
- C++ `SegmentLoadInfoTest.*`, `SchemaCacheTest.*`, and `CApiTest.UpdateSchemaTest`: 120/120 passed.
- QueryNode `TestCollectionManager` and focused compact/reopen tests: passed.
- `./internal/querycoordv2/task`: passed.
- `./internal/util/segcore`: passed.
- Focused RootCoord TTL/external property tests: passed.
- `git diff --check`: passed.

The complete `./internal/querynodev2/segments` package exceeded the local Go test default 10-minute timeout; the PR-specific suites above completed successfully. CI for the new head is pending.


xiaofan-luan:
### Review follow-ups addressed + root-cause tracked

The P1 review findings are fixed in the latest push:

- **Per-delegator schema convergence** — `delegator.UpdateSchema` now gates side effects on this delegator's own `servedSchemaVersion`, not the shared `Collection`. This fixes the case where a peer shard-delegator (co-located) or a Full/Delta/Watch `PutOrRef` pre-advances the shared `Collection` and suppresses this delegator's idfOracle / function-runner rebuild. The cursor is committed only after all side effects succeed (so a failure leaves it un-advanced for WAL replay).
- **Index meta after reopen** — a new `CollectionManager.UpdateIndexMeta` (index meta only, no served-schema advance) is applied after reopen on both worker and delegator, so newly-indexed fields pass segcore's search-plan `HasField` check (reopen decodes the index but the `SyncType_UpdateVersion` path carries no `IndexInfoList`).
- **Sole served-schema update failure** — the delete pipeline now panics for WAL replay on a genuine `UpdateSchema` failure instead of swallowing it and advancing TSafe; a transient not-ready (`ChannelNotAvailable`) error during the `pipeline.Start` → `delegator.Start` window is skipped rather than panicked to avoid killing the flowgraph at startup.

The per-delegator fix is a **scoped** fix. The deeper root cause is that a single global `Collection` schema is shared per `collectionID` per node (by delegators and workers), so it can be advanced by more than one path. The proper root-fix — giving each vchannel its own schema and removing the shared global schema — is tracked separately in #51814 (it dovetails with the online schema-evolution design's per-vchannel readiness aggregation) and is intentionally kept out of this PR to bound its scope.

Missing regression-test coverage for the reopen applied-vs-served split is tracked in #51805.

bigsheeper:

Adversarial review of this PR surfaced one critical DataCoord recovery defect plus five smaller correctness, error-handling, and crash-safety issues across the index-snapshot and QueryNode delegator paths.

## Critical

**`internal/metastore/kv/datacoord/kv_catalog.go:634` — `applySnapshot` parses a root-prefixed etcd key with `TrimPrefix`, so DataCoord cannot recover.**
`etcdKV.WalkWithPrefix` prepends `rootPath` to the prefix (`internal/kv/etcd/etcd_kv.go:89`) but hands the callback the **raw** etcd key (`etcd_kv.go:108`). DataCoord's catalog KV is rooted at a non-empty `metaRootPath` (`internal/datacoord/server.go:616-617`), so the callback actually receives `by-dev/meta/field-index-snapshot/10/200/2`. `strings.TrimPrefix(key, util.FieldIndexSnapshotPrefix+"/")` never matches, `relative` stays the full key, `Split` yields 6 parts, and `len(parts) != 3` falls through to `WrapErrDataIntegrityMsg` at line 636.
Failure: any collection that has gone through one revisioned index DDL leaves `<rootPath>/field-index-snapshot/{coll}/{rev}/{indexID}` in etcd → on the next DataCoord/MixCoord restart, `ListIndexes` fails, `indexMeta.reloadFromKV` propagates the error (`internal/datacoord/index_meta.go:216-220`), and the coordinator crash-loops with no self-healing path.
Suggestion: parse from the tail (last 3 path segments), matching the two prefix-robust siblings in the same function — `applyLegacy` uses `path.Base(path.Dir(key))` (`kv_catalog.go:668`) and `ListIndexSnapshotRevisions` uses `path.Base(key)` (`kv_catalog.go:692`). Note the unit test at `kv_catalog_test.go:1030-1032` feeds bare keys, which is why CI stays green; please extend it to cover rooted keys.
*(raised by xiaocai2333, sijie-ni-0214)*

## Medium

**`internal/querynodev2/delegator/delegator_data.go:246` — `ProcessInsert` panics on a benign `ErrSegmentNotLoaded`.**
After the new growing segment is `Put` into the segment manager (line 241), the index-version check at line 242 calls `local.UpdateGrowingIndexConfig`, whose first step is `if !s.ptrLock.PinIfNotReleased() { return merr.WrapErrSegmentNotLoaded(...) }` (`internal/querynodev2/segments/segment.go:1342-1344`). Neither `ReleaseSegments` (`internal/querynodev2/services.go:745`) nor `UnsubDmChannel` (`services.go:450`) holds `growingSegmentLock`, so a concurrent release is expected, not a corruption signal.
Failure: QueryCoord issues `ReleaseSegments` (Growing scope) or `UnsubDmChannel` for a segment just created by `ProcessInsert` while `currentIndexVersion > creationIndexVersion` → `PinIfNotReleased` returns false, the error surfaces, and the QueryNode process panics, cutting all queries on that shard until the segments are reloaded.
Suggestion: mirror the sibling `UpdateIndexInfoListWithError`, which treats this exact error as `continue` (`delegator.go:482-486`) — log and proceed instead of `panic(err)`.
*(raised by xiaocai2333, sijie-ni-0214)*

**`internal/core/src/segcore/SegmentIndexMeta.h:98` — metric-omitted search fails with `FieldNotLoaded` when a growing segment's index config is empty or lagging.**
This PR makes the metric resolvable only from the segment itself: `ResolveMetricTypeFromIndexMeta` returns empty when `index_meta` lacks the field (lines 72-79), and `ResolveSearchMetricType` throws `FieldNotLoaded` when both the segment metric and the request metric are empty (lines 96-100). A growing segment's `index_meta` comes from the delegator's self-built load info, which carries only `IndexInfos: fieldIndexes` (`delegator_data.go:193`) — the channel-level snapshot from `WatchDmChannels` (`services.go:284`), which lags after a new index is created until `UpdateIndexInfoList` refreshes it. Proxy leaves `metricType` empty whenever the user does not specify one (`internal/proxy/search_util.go:524-527`), which is the common case. Sealed segments are unaffected because `AppendCollectionIndexConfig` backfills config-only entries into the load info (`services.go:569`).
Failure: `create index` on a loaded collection, then a streaming write creates a growing segment before QueryCoord's target refresh reaches the delegator, then a search without `metric_type` in `search_params` → segcore throws `FieldNotLoaded` ("field index of the field: X is not loaded, please reload the collection") and the whole search request errors out until the snapshot refreshes.
Suggestion: let the growing path fall back to the collection-level index config, or return an empty metric so the existing Go-side tolerance ("segment reports no metric → skip") at `tasks/search_task.go:339-344` can apply, instead of throwing from segcore.
*(raised by xiaocai2333, sijie-ni-0214)*

## Low

**`internal/metastore/kv/datacoord/update.go:70` — `CleanupIndexSnapshot` removes the marker before the legacy prefix; a crash in that window resurrects legacy indexes.**
The three non-atomic removals run marker (line 70) → snapshot prefix (line 73) → legacy prefix (line 76). Marker-first is correct for snapshot keys, but it also exposes legacy keys: `applyLegacy` skips entries only while `revisions[collectionID]` exists (`kv_catalog.go:670-672`), so once the marker is gone, leftover legacy protos with `IsDeleted=false` load as active index definitions.
Failure: a collection with an orphan legacy `field-index/{coll}/{idx}` key (left by an earlier partial `RemoveIndex`) is dropped, GC gating passes, and the process crashes after line 70 but before line 76 → on restart the stale index defs load as active for a collection that no longer exists, and it never self-heals: `GetIndexSnapshotCollectionIDs` reads the in-memory revision map (`internal/datacoord/index_meta.go:1400-1404`), which is empty after restart, so cleanup never re-fires, and `recycleUnusedIndexes` only reclaims `IsDeleted` entries.
Suggestion: reverse the order — remove the legacy prefix first, then the snapshot prefix, then the marker, so both key families stay consistent at every crash point.
*(raised by xiaocai2333)*

**`internal/querycoordv2/meta/target_manager.go:422` — the revision-0 fallback places `indexInfoVersion` outside the revision domain.**
When `indexInfoPresent && indexInfoVersion == 0`, the fallback assigns `time.Now().UnixNano()` (~1.75e18), while real revisions live in the `1<<62` domain (~4.6e18, `internal/datacoord/ddl_callbacks.go:48-59`). The forward direction is safe because the wall-clock value is always below any real revision, so a real revision is always accepted. The reverse is not: once a delegator has seen a real revision, any later snapshot that falls back again is permanently rejected by the monotonicity check in `setFieldIndexes` (`delegator.go:174-182`). Given the PR body already declares this master-breaking with no rolling active/standby compatibility, this is a hardening/documentation request rather than a live bug.
Suggestion: add a `max()` guard, or document the invariant explicitly — the fallback value must always be below the revision domain, which makes this direction irreversible.
*(raised by xiaocai2333, sijie-ni-0214)*

**`internal/querynodev2/delegator/delegator.go:1499` — `UpdateSchema` converts a returnable `function.GetManager().Update` error into a `panic`, bypassing the newly added retry.**
The pre-existing `updateDelegatorSchemaLocked` returns the error for the same call (`delegator.go:280-282`); the new `UpdateSchema` path panics instead, while the caller in `internal/querynodev2/pipeline/delete_node.go:120-132` just gained a bounded retry (5 attempts with backoff) that this error class can never reach. The inline rationale checks out — `Update`'s synchronous errors come only from pure metadata validation (`internal/util/function/manager.go:474-490`, `1048-1092`, `546-556`) and runner construction is asynchronous — so this is not a vacuous retry, but unilaterally promoting a returnable error to a process crash is a behavior change worth an explicit decision: any internal inconsistency kills the QueryNode, and WAL replay redelivers the same `UpdateSchema` event.
Suggestion: return the error and let the existing retry path handle it, panicking only after the attempts are exhausted (`delete_node.go:145-150`).
*(raised by xiaocai2333, sijie-ni-0214)*

bigsheeper:
BTW, lgtm for the scheam version part.

czs007:

Adversarial review of the current head found four items: two high-severity (a PR description that no longer matches the code, and an unbounded in-memory map in the new schema registry) and two low-severity documentation/API-hygiene gaps in newly added files.

## High

**`PR description` — the PR body describes code that is not in this head.** Several load-bearing claims are contradicted by the diff at `0d519ef67f`:

- The entire **"Versioned index snapshots"** section has no corresponding code. `ListIndexesResponse` carries only `status` + `index_infos` with no revision field (`pkg/proto/index_coord.proto:289-292`), the diff touches no file under `internal/metastore/kv/datacoord/`, and the only DataCoord changes are `index_inspector.go`, `task_index.go`, and their tests — so there is no revision-scoped staging prefix, no revision key as a visibility marker, no chunked snapshot above the KV txn limit, and no obsolete-snapshot reclamation. On the QueryCoord side only `task/scheduler.go` and `task/utils.go` change, so there is no target-version logical clock either.
- "QueryNode's collection-wide `CollectionIndexMeta` and `SetIndexMeta` path are removed" did not happen. `ComposeIndexMeta` still builds a collection-wide `segcorepb.CollectionIndexMeta` including `MaxIndexRowCount` (`internal/querynodev2/segments/index_meta.go:31,54-56`), and it is still installed collection-wide via `PutOrRef` / `UpdateIndexMeta` (`internal/querynodev2/segments/collection.go:45,54`; `internal/querynodev2/services.go:644-645`; `internal/querynodev2/handlers.go:167-168`; `internal/querynodev2/delegator/delegator_data.go:611-612,865-866`). That file is not in the diff at all, and the comment at `internal/querynodev2/services.go:637-638` states outright that reopen still refreshes the collection index meta.
- The Verification section cites results for packages this PR never touches — "Full `internal/metastore/kv/datacoord` test binary: passed", "Full `internal/querycoordv2/meta` test binary: passed", and the 64-index DropAll staging / crash-recovery / slow-clock promotion tests. Those tests do not exist in this diff.
- The Compatibility note justifies the master-breaking label with "new revision-scoped DataCoord metadata", which likewise is not present.

This violates CLAUDE.md G3 ("Commit messages and PR body may assert ONLY benefits verified end-to-end"), which is an enforced gate in this repo. **Suggestion:** rewrite the body to describe only what this head does — per-vchannel/per-segment schema authority, `schema.Version` as the sole ordering key, reopen no longer advancing the served cursor, the property-driven version bump, the function-output materialization gate in `index_inspector.go`, and the empty-search validation — and drop or defer the index-snapshot section along with its Verification entries. *(raised by czs007)*

**`internal/querynodev2/schemaregistry/registry.go:59` — `segmentGenerationMap` grows without bound for the lifetime of a vchannel.** `AdvanceSegmentGenerationFloor` inserts one entry per `(segmentID, segmentType)` on every successful publication (`internal/querynodev2/segments/manager.go:689-690`, `:947-948`), and no path ever deletes an individual key: `removeSegmentWithType` (`manager.go:1236-1258`) only touches `globalSegments` and `secondaryIndex`, and the map is discarded wholesale only when the last channel owner is released (`registry.go:398`, `:609`). A delegator holds a channel owner for its entire life, so reclamation happens only on channel unwatch or process restart. Failure: a QueryNode hosting several high-throughput shard delegators with no release/balance churns new growing and compaction-produced sealed segment IDs continuously → the map climbs monotonically to O(10^5–10^6) entries per vchannel (tens of MB of resident tombstone state, multiplied per channel), with no metric exposing it. The comment at `registry.go:55-58` acknowledges the O(segment IDs seen per channel lifetime) cost, but "intentional" is not the same as "bounded". **Suggestion:** bound it — drop the tombstone once the segment slot is released *and* its recorded generation is below the current channel schema floor (the floor already fences the stale-publication case that motivated the tombstone), or use a bounded LRU and treat a miss as "no floor". At minimum, export the per-channel map size as a metric so the growth is observable. *(raised by czs007)*

## Low

**`internal/querynodev2/schemaregistry/registry.go:562` — `RemoveChannel` is an unannotated force-release path with no production callers.** It calls `delete(r.channels, key)` and then `releaseChannel` without checking `channel.owners`, bypassing the owner ref-count protocol that the rest of the file depends on (`RetainChannel` / `RetainExistingChannel` / `ReleaseChannel`, plus the `channel.removed || channel.owners == 0` guards in `Put`, `Acquire`, `AdvanceSchemaFloor`, and `AdvanceSegmentGenerationFloor`); `releaseChannel` unconditionally sets `removed = true; owners = 0` and clears both `versions` and `segmentGenerationMap` (`registry.go:597-611`). Its sibling `RemoveCollection` already documents exactly this constraint ("It is reserved for whole-registry/test teardown: collection ref-count teardown must not call it…", `registry.go:574-578`), but `RemoveChannel` carries no equivalent note, and its only caller in the tree is `registry_test.go:66`. **Suggestion:** give `RemoveChannel` the same test-only comment as `RemoveCollection`, or move it to a `_testonly.go` file, so both force-release paths state the same constraint. Defensive hardening only — nothing triggers it today. *(raised by czs007)*

**`internal/rootcoord/schema_version.go:55` — the deny-list's implicit constraint is undocumented.** `schemaRefreshExcludedPropertyKeys` narrows only the *decision* of whether to bump the version: `schemaPropertiesChanged` compares property sets after `projectSchemaProperties` filtering (`schema_version.go:78-88`), while the broadcast snapshot still carries the *full* property set (`ddl_callbacks_alter_collection_properties.go:195-196`; `internal/metastore/model/collection.go:164`). The result is an unwritten rule: a change to a deny-listed property alters the persisted schema payload without bumping `schema.Version`, and segcore's process-global `SchemaCache::AcquireOrCreate` keys purely on `(collection_id, schema_proto.version())` and reuses the parsed object without comparing payloads (`internal/core/src/segcore/SchemaCache.cpp:469-487`) — so two QueryNodes can hold different payloads under the same key with no invalidation. This is safe today: `Schema::ParseFrom` reads only `mmap.enabled` (`Schema.cpp:268`), `ttl_field` (`:270-287`), and the `warmup.*` keys (`:290-297`), none of which are deny-listed, and `collection.ttl.seconds` (`pkg/common/common.go:267`) does not collide with the C++-side `ttl_field` (`Consts.h:95`). **Suggestion:** state the constraint explicitly in the comment block above the deny-list — any key added here must be read by neither `Schema::ParseFrom` nor any QueryNode consumer of `Collection.properties`. *(raised by czs007)*

czs007:

The PR body describes a versioned index-snapshot mechanism and a QueryNode index-meta removal that do not exist in this diff, so the description needs to be brought in line with the code before merge.

## Medium

**`PR description:1` — The `## Versioned index snapshots` section has no corresponding implementation, and contradicts this PR's own design doc.**
The body claims DataCoord persists a collection-level index revision and returns it from `ListIndexes`, that snapshots are staged under a revision prefix and committed with a trailing marker, that oversized snapshots are auto-chunked past the KV transaction limit, that recovery reads only the marker-selected snapshot, that stale snapshots and the final revision/snapshot domain are reclaimed, and that QueryCoord target versions moved to a logical clock. None of this is in the diff:

- `pkg/proto/index_coord.proto` is untouched by this PR; at head `a89e2e7`, `ListIndexesResponse` (index_coord.proto:289-292) still carries only `status` + `index_infos` — there is no revision field for `ListIndexes` to return.
- Across the full 169-file patch, `staging`, `DropAll`, `IndexRevision`, `logical clock`, and `ListIndexes` all have zero hits; `revision` hits twice, both inside the design doc and both saying the opposite.
- DataCoord changes are limited to four files (`index_inspector.go`, `index_inspector_test.go`, `task_index.go`, `task_index_test.go`), covering function-output materialization gating and VectorArray estimation order. `internal/metastore/kv/datacoord` is not in the changed-file list, so no persistence, GC, or recovery path exists.
- QueryCoord changes are limited to `task/scheduler.go` (error-classification allowlist, `shouldRecordSegmentTaskError`) and `task/utils.go` (a comment rewrite plus the existing `SchemaBarrierTs` assignment). `querycoordv2/meta` and target-version allocation are untouched.
- `WatchDmChannelsRequest.index_info_list` (query_coord.proto:372) is a pre-existing master field; this PR's only `query_coord.proto` changes are `LoadMetaInfo.schema_barrier_ts` and `UpdateSchemaRequest`.

The design doc added by this PR states the opposite directly (`docs/design-docs/design_docs/segcore/20260204-loaddiff_based_segment_load.md:204-212`): "`IndexMeta` snapshot still has no monotonic revision … this change must not be described as providing online index-DDL convergence." The top-of-body blockquote ("after the new index-snapshot revision domain is in use") and the Compatibility note ("the new revision-scoped DataCoord metadata") are likewise unsupported.

**Suggestion:** Delete the `## Versioned index snapshots` section outright, along with the revision wording in the top blockquote and the Compatibility note. Keep only what the diff actually contains — SchemaCache, `schema.Version`, `FieldLoadPolicy`, and the reopen path. *(raised by xiaocai2333, bigsheeper)*

**`PR description:1` — The claim that QueryNode's collection-wide `CollectionIndexMeta` / `SetIndexMeta` path was removed is false.**
The body states: "QueryNode's collection-wide `CollectionIndexMeta` and `SetIndexMeta` path are removed; search, refine, sealed loading, and growing interim indexes resolve configuration from segment/load metadata." At head `a89e2e7` that path is fully intact, and part of it is added by this PR:

- `UpdateIndexMeta` is a new method on the `collectionManager` interface (`internal/querynodev2/segments/collection.go:51-54`, implementation at `:142`), along with a newly generated `MockCollectionManager_UpdateIndexMeta_Call`.
- The cgo chain is unchanged: `Collection.updateIndexMeta` (collection.go:288) → `CCollection.UpdateIndexMeta` → `C.SetIndexMeta` (`internal/util/segcore/collection.go:111`); `internal/core/src/segcore/collection_c.cpp:74` still defines `SetIndexMeta`.
- `internal/util/segcore/collection.go` deletes `CCollection.Schema()` and `CCollection.UpdateSchema()` but explicitly retains the `indexMeta` field, `IndexMeta()`, and `UpdateIndexMeta()`.
- `internal/querynodev2/segments/index_meta.go` (`ComposeIndexMeta`, :31) is not in this PR's changed-file list at all.
- This PR adds new call sites into that path, in `internal/querynodev2/services.go` and `delegator/delegator_data.go`.

What this PR actually removes is collection-wide *schema* ownership, not index meta. As written, the sentence invites reviewers to skip the newly added `UpdateIndexMeta` call sites.

**Suggestion:** Narrow the sentence to the schema-side removal only, or state that collection-wide `CollectionIndexMeta` is retained and that versioned index-meta convergence is follow-up work — which matches the "Index metadata boundary" section of the design doc in this PR. *(raised by xiaocai2333)*

MrPresent-Han:
I will take over this

internal/core/src/segcore/SegmentLoadInfo.h
```diff
@@ -1111,6 +1174,66 @@ class SegmentLoadInfo {
...
+            collect_field_policy(field);
+        }
+        for (const auto& struct_field : load_schema.struct_array_fields()) {
+            for (const auto& field : struct_field.fields()) {
+                collect_field_policy(field);
+            }
+        }
+        // SegmentLoadInfo retains only the effective policy, not another full
+        // schema copy per segment.
+        info_.clear_load_schema();
```
buqian-zilliz, line 1237: **[P1] Preserve the extracted load policy across reconstruction**

`InitLoadPolicy` clears the only serialized copy of the policy, but `ChunkedSegmentSealedImpl::Load` later constructs a fresh `SegmentLoadInfo` from `snapshot->load_info->GetProto()`. The new object therefore sees no `load_schema`, marks the policy unsupplied, and falls back to the cached logical schema. Initial sealed loading consequently ignores the per-segment mmap/warmup snapshot whenever it differs from the logical schema, potentially turning an mmap load into resident memory. Copy the existing `SegmentLoadInfo`, including `load_policy_`, or otherwise preserve the extracted policy across this reconstruction.

internal/querynodev2/segments/collection.go
```diff
@@ -398,6 +361,8 @@ func (c *Collection) CreateCSegment(req *segcore.CreateCSegmentRequest) (segcore
 		return nil, merr.WrapErrServiceInternal("create segment on released collection")
 	}
 	req.Collection = c.ccollection
+	req.LoadSchema = c.LoadSchema()
+	req.LoadFields = c.loadFields.Collect()
```
buqian-zilliz: **[P1] Refresh load fields when applying an existing load update**

`c.loadFields` is initialized only by `NewCollection`, while subsequent `PutOrRef` calls discard the incoming `LoadMetaInfo` after extracting its barrier. QueryCoord supports changing the load-field set on an already loaded collection and forwards the new set in later load metadata, but these new forwarding lines keep sending the original set to every created or reopened segment. Shrinking the set continues eagerly loading removed fields, while expanding it leaves newly requested fields lazy, so the accepted load configuration is not applied until the `Collection` is destroyed. Update the field set atomically from accepted load updates using the same barrier ordering as `loadSchema`.

internal/querynodev2/segments/segment.go
```diff
@@ -1295,11 +1295,12 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
...
 
-	schema, schemaVersion := s.collection.SchemaAndSegcoreVersion()
+	schema := s.collection.Schema()
 	err := s.csegment.Reopen(ctx, &segcore.ReopenRequest{
-		LoadInfo:      newLoadInfo,
-		Schema:        schema,
-		SchemaVersion: schemaVersion,
+		LoadInfo:   newLoadInfo,
+		Schema:     schema,
+		LoadSchema: s.collection.LoadSchema(),
```
buqian-zilliz: **[P2] Capture logical and load schemas from one snapshot**

`Schema()` and `LoadSchema()` independently load `c.schema`, and `Reopen` does not hold `schemaTransitionMu`. A concurrent `applyLoadUpdate` can publish between the two calls, producing a reopen request with the old logical schema and the new load-policy schema. Segcore then computes its diff against mismatched schema epochs; for example, a field introduced by the new version can be filtered out until another reopen occurs. Return both pointers from one `collectionSchemaSnapshot` load or hold the schema-transition read lock while capturing them. `CreateCSegment` has the analogous native-schema/Go-policy transition window.

internal/core/src/segcore/SegmentLoadInfo.h
```diff
@@ -609,6 +618,59 @@ class SegmentLoadInfo {
...
 
+    [[nodiscard]] bool
+    ShouldLoadField(FieldId field_id) const {
+        if (!schema_->ShouldLoadField(field_id)) {
+            return false;
+        }
+        if (info_.load_fields().empty()) {
+            return true;
+        }
+        return std::find(info_.load_fields().begin(),
```
buqian-zilliz: **[P2] Cache load-field membership instead of scanning per field**

`ShouldLoadField` linearly scans `info_.load_fields()` for every field. The normal non-partial path materializes every schema field into that list, so the empty-list fast path is not used, and this helper is called inside per-field and column-group loops during load and reopen. Each pass therefore becomes O(F²) per segment, regressing the previous unordered-set lookup and multiplying the cost across the segment fan-out. Build a membership set once when initializing the policy, or preserve empty-as-all on the wire.

internal/util/segcore/segment.go
```diff
@@ -453,6 +459,8 @@ func ConvertToSegcoreSegmentLoadInfo(src *querypb.SegmentLoadInfo) (*segcorepb.S
 		UseTakeForOutput:     src.GetUseTakeForOutput(),
 		EstimatedBytesPerRow: src.GetEstimatedBytesPerRow(),
 		CommitTimestamp:      src.GetCommitTimestamp(),
+		LoadFields:           loadFields,
+		LoadSchema:           loadSchema,
```
buqian-zilliz, line 464: **[P2] Avoid sending a full schema with every segment policy**

This embeds a complete `CollectionSchema` in every segment creation and reopen message even though C++ only extracts mmap and warmup settings. Go serializes the full schema per segment, C++ scans all top-level and nested fields, copies the effective policy, and immediately clears the schema. A collection-wide operation over S segments therefore adds O(S × schema size) serialization, parsing, and transient allocation. Carry a compact load-policy message or share a policy snapshot keyed by collection and barrier instead.

internal/rootcoord/ddl_callbacks_alter_collection_properties.go
```diff
@@ -178,17 +181,13 @@ func (c *Core) broadcastAlterCollectionForAlterCollection(ctx context.Context, r
 			header.UpdateMask.Paths = append(header.UpdateMask.Paths, message.FieldMaskCollectionSchema)
 		}
 
-		// Build schema snapshot with updated properties (schema version should NOT be changed for properties-only alter).
+		// These values are consumed from the schema snapshot, so changing them
+		// creates a new logical schema version.
 		schema := coll.ToCollectionSchemaPB()
+		schema.Version = coll.SchemaVersion + 1
```
liliu-z: Setting `collection.ttl.field` or updating external source/spec now bumps `schema.Version` (line 187), which DataCoord treats as "segment is physically behind the collection schema". For any collection with functions, index_inspector.go:188-193 defers every index build on segments whose `SchemaVersion` is behind, and the only mechanism that advances a flushed segment's version without a data change — BumpSchemaVersionCompaction — is disabled by default (configs/milvus.yaml:837) and explicitly skipped for external collections (compaction_policy_bump_schema_version.go:94-97). Since DataCoord itself issues this alter after every external-collection refresh (server_external_schema_updater.go:40-47) while the segments that refresh just stamped carry the pre-alter version (task_refresh_external_collection.go:608), a function-bearing external collection ends every refresh cycle with unindexable segments and no self-healing path. The same bump also fails in-flight backfill commits, which fence on exact version equality (services_commit_backfill.go:419).

liliu-z:
**internal/core/src/segcore/SegmentInterface.h:0** SegmentLoadInfo.h:1232-1234 claims "SegmentLoadInfo retains only the effective policy, not another full schema copy per segment", but that clear only runs for sealed segments, which route through segcore::SegmentLoadInfo in ChunkedSegmentSealedImpl::SetLoadInfo. Growing segments use the base override at SegmentInterface.h:667, which stores the whole proto — including the full CollectionSchema that collection.go:364 now attaches to every segment — into `load_info_` and never parses or clears it, so the copy is resident for the segment's lifetime. Reopen additionally marshals the same schema twice per call (internal/util/segcore/segment.go:365 and :377).

internal/querynodev2/segments/collection.go
```diff
@@ -64,13 +64,8 @@ type collectionSchemaUpdatePlan struct {
 	// logicalSchemaVersion is schema.Version from the accepted schema payload.
 	// It is the Go-side structural schema freshness key.
 	logicalSchemaVersion uint64
-	// schemaBarrierTs fences stale load results and orders same-version schema
-	// payload refreshes such as collection property snapshots.
+	// schemaBarrierTs fences stale load results and orders load-policy snapshots.
```
liliu-z: The PR body claims versioned index snapshots, a QueryCoord target-version logical clock, empty-search metric validation, a per-vchannel schema registry, BM25 stats retry, and removal of the collection-wide `CollectionIndexMeta`/`SetIndexMeta` path. None of these are in the 28-file diff, and `updateIndexMeta` is still defined at collection.go:369 and still invoked unconditionally at collection.go:417-420. The Verification section likewise lists results for packages this diff does not touch, so a reviewer cannot map any listed claim onto reviewable code.

internal/core/src/segcore/SegmentLoadInfo.h
xiaofan-luan: Resolved in the current head. `ChunkedSegmentSealedImpl::Load` now copies the existing `SegmentLoadInfo` wrapper instead of reconstructing from its compacted proto, and the wrapper copy constructor preserves `load_policy_` and `load_fields_`. `SegmentLoadInfoTest.KeepsLoadPolicySeparateFromLogicalSchema` covers the compacted-proto plus wrapper-copy case.

internal/querynodev2/segments/collection.go
xiaofan-luan: Resolved in the current head. `loadFields` is part of the immutable `collectionSchemaSnapshot`; accepted `PutOrRef` load updates copy the new field set and publish it together with the logical schema, load policy, barrier, and segcore revision.

internal/querynodev2/segments/segment.go
xiaofan-luan: Resolved in the current head. `Reopen` captures logical schema, compact load policy, load fields, and segcore revision through one `segmentLoadState()` snapshot. Segment creation likewise captures the native collection and the matching load snapshot under the collection's segment-creation fence.

internal/core/src/segcore/SegmentLoadInfo.h
xiaofan-luan: Resolved in the current head. `SegmentLoadInfo::BuildCache` builds an `unordered_set<FieldId>` once, and `ShouldLoadField` now performs constant-time membership checks rather than scanning the repeated proto field for every schema field.

internal/util/segcore/segment.go
xiaofan-luan: Resolved in the current head. QueryNode stores one compact collection-level policy snapshot via `loadPolicyFromSchema`; sealed create/reopen messages carry only mmap/warmup keys and field IDs, while growing segment creation sets `LoadSchema` to nil. The full logical schema is sent only on the separate schema path required by reopen.

internal/rootcoord/ddl_callbacks_alter_collection_properties.go
xiaofan-luan: Resolved by removing the RootCoord version-bump change from this PR. `schema.Version` remains the persisted logical/materialization version; same-version TTL/external runtime-property refreshes are ordered by the barrier and a QueryNode-local segcore revision, so they do not advance DataCoord's physical schema fence.

internal/querynodev2/segments/collection.go
xiaofan-luan: Resolved. The PR body has been rewritten to match the current focused diff: process-wide SchemaCache ownership, collection/segment schema handles, logical versus effective load schema, compact load policy, atomic collection snapshots, and cache reclamation. The removed index-snapshot, per-vchannel registry, BM25 retry, and CollectionIndexMeta-removal claims are no longer present.

Issue #51793, opened 2026-07-24
[Feature]: support nested array


SpadeA-Tang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

We need to support nested array.

PR #51794, opened 2026-07-24
feat: support recursively nested Array fields

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/51793
milvus-proto: https://github.com/milvus-io/milvus-proto/pull/640

design doc: docs/design-docs/design_docs/20260724-recursive-array-struct-column-layout.md‎

TODO:
1. support filter expression
2. support import


go.mod
```diff
@@ -321,3 +321,5 @@ replace (
 	github.com/milvus-io/milvus/pkg/v3 => ./pkg
 	github.com/tecbot/gorocksdb => github.com/milvus-io/gorocksdb v0.0.0-20220624081344-8c5f4212846b // indirect
 )
+
+replace github.com/milvus-io/milvus-proto/go-api/v3 => /home/spadea/working2/milvus-proto/go-api
```
buqian-zilliz: **[P1] Replace the private proto checkout with a published version**

This replacement points every clean checkout at `/home/spadea/working2/milvus-proto/go-api`; the same absolute path is also committed in `client/go.mod` and `pkg/go.mod`. The declared proto pseudo-version predates the new `TypeSchema`/`ElementSchema` API, so CI and release builds without the author's directory cannot load the module or compile these references. Publish the proto change, bump the dependency in all three modules, and remove the host-local replacements.

internal/proxy/util.go
```diff
@@ -571,6 +579,95 @@ func validateElementType(dataType schemapb.DataType) error {
 	return merr.WrapErrParameterInvalidMsg("element type %s is not supported", dataType.String())
 }
 
+func validateTypeSchema(collectionName string, fieldName string, typeSchema *schemapb.TypeSchema) error {
+	if typeSchema == nil {
+		return merr.WrapErrParameterMissingMsg("element_schema should be specified for nested array field %s", fieldName)
+	}
+	if typeSchema.GetNullable() {
```
buqian-zilliz, line 565: **[P1] Honor nullable nested Array levels**

This rejects every `TypeSchema` level with `nullable=true`, while the new layout and builders carry an independent validity bitmap for each nested Array node and encode null children as `DATA_NOT_SET`. Consequently schemas such as `Array<Array<Int32>>` with nullable inner arrays can never be created or inserted, and null cannot be distinguished from an empty inner array through the public schema. Accept nullable nested nodes and make `checkNestedArrayValue` validate an absent child against that level's nullable flag.

client/go.mod
buqian-zilliz, line 1: **[P1] Add recursive Array support to client/v3**

The client module is switched to the new proto API, but client/v3 still cannot represent this schema or its values: `entity.Field` has no `ElementSchema`, `ProtoMessage`/`ReadProto` drop it, and the Array column parser rejects results whose element type is `Array`. Consequently the official Go client cannot create, insert, describe, or decode the feature added here. Add recursive schema and column/value round-tripping alongside the dependency bump.

internal/proxy/util.go
```diff
@@ -571,6 +579,95 @@ func validateElementType(dataType schemapb.DataType) error {
...
+			return merr.WrapErrParameterInvalidMsg("element_schema data type should be Array for nested array field %s", field.GetName())
+		}
+		if err := validateMaxCapacityPerRow(collectionName, field); err != nil {
+			return err
+		}
+		return validateTypeSchema(collectionName, field.GetName(), field.GetElementSchema())
+	}
+
+	if field.GetElementType() == schemapb.DataType_Array {
+		return merr.WrapErrParameterMissingMsg("element_schema should be specified for nested array field %s", field.GetName())
```
buqian-zilliz: **[P1] Expose ElementSchema through REST v2**

REST v2 has no way to satisfy this new `ElementSchema` requirement. Its create-collection request model only accepts `elementDataType` and populates `ElementType`, so requesting an Array element reaches this branch and is rejected for missing `ElementSchema`; the REST insert and describe converters also handle only flat `ElementType` values. Add recursive REST schema and value encoding/decoding so nested arrays can be created and round-tripped through this public API.

internal/storage/insert_data.go
```diff
@@ -2028,27 +2028,49 @@ func (data *StringFieldData) GetMemorySize() int {
...
+	// Array rows are stored as recursively nested ScalarField_ArrayData values.
+	// Inspect the payload before elementType so nested arrays still work when the
+	// top-level FieldSchema uses element_schema and therefore leaves element_type
+	// unset.
+	if arrayData := data.GetArrayData(); arrayData != nil {
+		var size int
+		for _, element := range arrayData.GetData() {
+			size += getArrayScalarFieldMemorySize(element, arrayData.GetElementType())
+		}
+		return size
```
buqian-zilliz, line 2044: **[P2] Count nested container overhead in Array memory usage**

The recursive branch sums only leaf payload bytes. A row containing thousands of empty child arrays therefore reports essentially zero bytes even though it allocates thousands of `ScalarField`/`ArrayArray` objects and pointer slices. `ArrayFieldData.GetMemorySize` feeds the DataNode write-buffer statistics and force-sync watermark, so these values can keep reported usage below the limit while real heap usage grows. Include the per-node/container overhead or use a conservative recursive serialized-size estimate.

internal/core/src/mmap/ChunkedColumn.h
```diff
@@ -667,14 +668,19 @@ class ChunkedArrayColumn : public ChunkedColumnBase {
...
+        : ChunkedColumnBase(std::move(slot), field_meta),
+          field_meta_has_element_schema_(field_meta.has_element_schema()) {
     }
 
     void
     BulkArrayAt(milvus::OpContext* op_ctx,
                 std::function<void(const ArrayView&, size_t)> fn,
                 const int64_t* offsets,
                 int64_t count) const override {
+        if (field_meta_has_element_schema_) {
```
buqian-zilliz: **[P2] Reject or implement array_length for recursive Arrays**

`array_length` is still accepted for any Array field by the plan parser, but execution reads Array fields through the legacy `ArrayView`/Span path. This new guard therefore makes an accepted `array_length(nested_array)` predicate throw `Unsupported` at runtime for sealed segments, and the corresponding Span guard does the same for growing segments. Add an `ArrayValueView` length path, or reject recursive-Array expressions during plan validation so accepted queries do not fail on QueryNode.

go.mod
SpadeA-Tang: This will be modifed once milvus-proto is ready.

internal/proxy/util.go
SpadeA-Tang: element-null is not support in this stage.

client/go.mod
SpadeA-Tang: Go SDK will not be covered in this PR.

internal/proxy/util.go
SpadeA-Tang: ditto.

internal/storage/insert_data.go
SpadeA-Tang: This is general issue for other types too where the overhead may not be included in the memory size.

internal/core/src/storage/Util.cpp
buqian-zilliz, line 1: **[P1] Support recursive ARRAYs in external Arrow normalization**

`NormalizeExternalArrow` still treats every ARRAY as a flat Arrow `List<T>` and passes `field_meta.get_element_type()` into `ConvertListToProtobufBinary`. For recursive fields that value is `NONE`, while a valid source column is `List<List<T>>`; `ArrowListElementTypeToMilvus` rejects the inner `LIST` before the column can be loaded. External collection validation allows ARRAY fields, so creating a recursive ARRAY mapping succeeds but sealed load, retrieve, and search fail as soon as this normalizer sees it. Drive conversion from `ElementSchema` and recursively serialize nested lists, or reject recursive ARRAYs for external collections.

internal/proxy/util.go
```diff
@@ -579,9 +676,13 @@ func validateFieldType(schema *schemapb.CollectionSchema) error {
 		case schemapb.DataType_None:
 			return merr.WrapErrParameterInvalidMsg("data type None is not valid")
 		case schemapb.DataType_Array:
-			if err := validateElementType(field.GetElementType()); err != nil {
-				return err
+			if field.GetElementSchema() == nil {
```
buqian-zilliz: **[P2] Reject recursive ARRAY indexes before scheduling**

Recursive ARRAY schemas accepted here can still be submitted for scalar indexing: explicit INVERTED validation accepts any ARRAY, and scalar auto-index derives a primitive index from `FieldSchema.ElementType`. That value must be `NONE` when `ElementSchema` is present, so the index worker passes `NONE` to `CreateCompositeScalarIndex` and fails with `DataTypeInvalid` before the build. The raw-data readers also do not receive the recursive `TypeSchema`. Reject recursive-ARRAY indexes at the request boundary, or carry `ElementSchema` through index selection and data loading.

internal/rootcoord/util.go
```diff
@@ -468,6 +468,27 @@ func checkStructArrayFieldSchema(schemas []*schemapb.StructArrayFieldSchema) err
 				msg := fmt.Sprintf("fields in StructArrayField can only be array or array of vector, but field %s is %s", field.Name, field.DataType.String())
 				return merr.WrapErrParameterInvalidMsg(msg)
 			}
+			if field.GetDataType() == schemapb.DataType_ArrayOfVector &&
+				(field.GetElementSchema() != nil || field.GetElementType() == schemapb.DataType_ArrayOfVector) {
+				return merr.WrapErrParameterInvalidMsg("nested ArrayOfVector is not supported for field %s", field.Name)
+			}
+			if field.GetElementSchema() != nil {
```
buqian-zilliz: **[P2] Recursively validate ElementSchema in RootCoord**

This RootCoord check validates only the first `ElementSchema` node and never walks the rest of the chain; `checkFieldSchema` does not validate `ElementSchema` for ordinary fields either. A caller that reaches RootCoord directly can therefore persist a schema with a non-ARRAY child, both `element_type` and `element_schema`, or no leaf element type. Recursive C++ readers later call `ColumnarArrayChunk::ValidateArrayType` and assert on that metadata. Apply the same recursive `TypeSchema` validation at this DDL boundary before storing ordinary or StructArray fields.

internal/proxy/util.go
```diff
@@ -571,6 +579,95 @@ func validateElementType(dataType schemapb.DataType) error {
...
+		if field.GetElementType() != schemapb.DataType_None {
+			return merr.WrapErrParameterInvalidMsg("element_type should not be specified for nested array field %s", field.GetName())
+		}
+		if field.GetElementSchema().GetDataType() != schemapb.DataType_Array {
+			return merr.WrapErrParameterInvalidMsg("element_schema data type should be Array for nested array field %s", field.GetName())
+		}
+		if err := validateMaxCapacityPerRow(collectionName, field); err != nil {
+			return err
+		}
+		return validateTypeSchema(collectionName, field.GetName(), field.GetElementSchema())
```
buqian-zilliz: **[P2] Gate partial Array operations on recursive fields**

`ARRAY_APPEND` and `ARRAY_REMOVE` are still admitted for every `DataType_Array` field, but recursive schemas require `FieldSchema.ElementType=NONE`. A correctly tagged payload with `ArrayArray.element_type=Array` is rejected against `NONE`; omitting the tag only postpones failure until merge, where `ApplyArrayRowOp` dispatches on the base element type `Array` and returns unsupported. Reject these operations up front when `ElementSchema` is present, or implement recursive append and removal semantics.

internal/core/src/common/ChunkWriter.cpp
```diff
@@ -34,10 +40,691 @@
...
+GetColumnarArrayElementType(const proto::schema::TypeSchema& type) {
+    return type.has_element_schema() ? DataType::ARRAY
+                                     : DataType(type.element_type());
+}
+
+template <typename T>
+void
+AppendFixedValue(std::vector<char>& data, T value) {
+    const auto old_size = data.size();
+    data.resize(old_size + sizeof(T));
```
buqian-zilliz: **[P2] Bulk-copy fixed-width recursive Array leaves**

`AppendFixedValue` calls `vector<char>::resize` for every numeric leaf value and then immediately overwrites the value-initialized bytes with `memcpy`. Because the recursive builder never reserves `fixed_data`, large growing inserts and sealed loads pay per-element initialization plus repeated geometric reallocations and copies across the entire leaf column. Sum the leaf element count, resize once, and copy each protobuf repeated-field range contiguously.

go.mod
buqian-zilliz: This issue still exists: the current head still depends on the author's absolute local milvus-proto checkout in all three modules, so clean checkouts and CI cannot resolve the new proto API.

internal/rootcoord/util.go
SpadeA-Tang: RootCoord is an internal service that cannot bypass Proxy validation. Full validation for other field types also happens only in Proxy.

internal/core/src/storage/Util.cpp
SpadeA-Tang: Banned temporarily for external collection.

internal/proxy/util.go
SpadeA-Tang: Banned now.

internal/proxy/util.go
SpadeA-Tang: Now, reject ARRAY_APPEND/ARRAY_REMOVE for nested array.

internal/core/src/common/ChunkWriter.cpp
SpadeA-Tang: optimized.

pkg/util/typeutil/field_schema.go
```diff
@@ -89,3 +89,76 @@ func CreateFieldSchemaHelper(schema *schemapb.FieldSchema) *FieldSchemaHelper {
...
+	error,
+) {
+	if typeSchema == nil {
+		return schemapb.DataType_None, schemapb.DataType_None,
+			merr.WrapErrParameterInvalidMsg(
+				"type_schema kind should be specified for field %s", fieldName)
+	}
+
+	switch kind := typeSchema.GetKind().(type) {
+	case *schemapb.TypeSchema_ArrayElement:
```
zhengbuqian, line 144: nit: should we also rename `array_element` to `array_element_type`?

pkg/util/typeutil/field_schema.go
```diff
@@ -89,3 +89,76 @@ func CreateFieldSchemaHelper(schema *schemapb.FieldSchema) *FieldSchemaHelper {
...
+		if kind.ArrayElement == nil {
+			return schemapb.DataType_None, schemapb.DataType_None,
+				merr.WrapErrParameterInvalidMsg(
+					"type_schema array element should be specified for field %s", fieldName)
```
zhengbuqian: `array element` -> `array_element`? same for other error messages, to be more careful.

pkg/util/typeutil/field_schema.go
```diff
@@ -89,3 +89,76 @@ func CreateFieldSchemaHelper(schema *schemapb.FieldSchema) *FieldSchemaHelper {
...
+	typeSchema := field.GetTypeSchema()
+	if typeSchema == nil {
+		return nil
+	}
+
+	dataType, elementType, err := validateTypeSchemaNode(field.GetName(), typeSchema)
+	if err != nil {
+		return err
+	}
+	if dataType != field.GetDataType() {
```
zhengbuqian: we allow not setting data_type and element_type, right? so them being none should be ok. error only if it was set but not match

pkg/util/typeutil/field_schema.go
```diff
@@ -89,3 +89,76 @@ func CreateFieldSchemaHelper(schema *schemapb.FieldSchema) *FieldSchemaHelper {
...
+		return merr.WrapErrParameterInvalidMsg(
+			"type_schema data type %s does not match data_type %s for field %s",
+			dataType.String(), field.GetDataType().String(), field.GetName())
+	}
+	if elementType != field.GetElementType() {
```
zhengbuqian: should we disallow `elementType` to be nested data type? to enforece all usage of newly supported nested data type to use the new `TypeSchema` instead of `data_type` + `element_type`. same for `data_type`: we accept setting both only for existing supported data types for compatibility reason. any new type must use only `TypeSchema`.

internal/util/indexparamcheck/index_params_validation.go
```diff
@@ -220,6 +220,10 @@ func FillDimension(field *schemapb.FieldSchema, indexParams map[string]string) e
 // add-function-field bound-index path enforces identical rules at every DDL
 // boundary, including callers that reach rootcoord directly.
 func ValidateFieldIndexParams(field *schemapb.FieldSchema, indexParamsMap map[string]string) error {
+	if field.GetDataType() == schemapb.DataType_Array && field.GetElementType() == schemapb.DataType_Array {
```
zhengbuqian: should also check type_schema. 

Let's define the 2 type systems:

* **legacy type system**: `data_type + element_type`. this can only model types before this PR. we also define such data types to be **legacy types**.
* **schematic type system**: data type defined by `type_schema`. we also define such data types to be **schematic type**

as discussed in `field_schema.go`, how about enforce this:

* schematic type can only specify `type_schema`, and leave `data_type + element_type` to be unset/NONE.
* legacy type must continue set `data_type + element_type`, so we don't have to change all existing code, and existing checks on legacy type can continue checking only `data_type + element_type`. but they may optionally set `type_schema`. if both set, at any API entrance, we check them to be consistent.

then for checks related to schematic types, we should only check against `type_schema`.

I'll reference this as Comment A, in other comments.

zhengbuqian: this is basically a migration process suggestion, I'll get this reviewed and agreed upon by the maintainers asap.

internal/storage/utils.go
```diff
@@ -107,6 +114,11 @@ func ValidateStorageV1InsertWritableSchema(schema *schemapb.CollectionSchema) er
 	return nil
 }
 
+func isNestedArrayField(field *schemapb.FieldSchema) bool {
+	return field.GetDataType() == schemapb.DataType_Array &&
```
zhengbuqian: comment A.

also rename this method to `isLegacyTypeField`? as we'll want to reject any schematic type in storage v1, not just array of array.

internal/proxy/util.go
```diff
@@ -452,8 +452,12 @@ func validateDimension(field *schemapb.FieldSchema) error {
 }
 
 func validateMaxLengthPerRow(collectionName string, field *schemapb.FieldSchema) error {
```
zhengbuqian: this wrapper seems unnecessary

internal/proxy/util.go
```diff
@@ -464,38 +468,42 @@ func validateMaxLengthPerRow(collectionName string, field *schemapb.FieldSchema)
...
 	// if not exist type params max_length, return error
 	if !exist {
-		return merr.WrapErrParameterMissingMsg("type param(max_length) should be specified for the field(%s) of collection %s", field.GetName(), collectionName)
+		return merr.WrapErrParameterMissingMsg("type param(max_length) should be specified for the field(%s) of collection %s", fieldName, collectionName)
 	}
 
 	return nil
 }
 
 func getMaxCapacityPerRow(collectionName string, field *schemapb.FieldSchema) (int64, error) {
```
zhengbuqian: ditto

internal/proxy/task_upsert_partial_op.go
```diff
@@ -45,6 +45,7 @@ func hasNonReplacePartialOp(req *milvuspb.UpsertRequest) bool {
 //   - Each op targets a field present in the collection schema.
 //   - Array-specific ops (ARRAY_APPEND, ARRAY_REMOVE) require the target
 //     field to have DataType_Array.
+//   - Recursive Array fields do not support ARRAY_APPEND or ARRAY_REMOVE.
```
zhengbuqian: add `// TODO(NestedSchema)` .

for all features that are currently not supported, add this TODO.

internal/core/src/storage/Util.cpp
```diff
@@ -1022,25 +1022,32 @@ std::vector<std::future<std::unique_ptr<DataCodec>>>
...
+            return res;
+        };
 
     for (auto& file : remote_files) {
-        futures.emplace_back(pool.Submit(
-            DownloadAndDeserialize, remote_chunk_manager, is_field_data, file));
+        futures.emplace_back(pool.Submit(DownloadAndDeserialize,
+                                         remote_chunk_manager,
+                                         is_field_data,
+                                         array_type,
```
zhengbuqian, line 1055: do we also need std::move here?

pkg/util/typeutil/field_schema.go
SpadeA-Tang: What you mean? Change proto field name?

pkg/util/typeutil/field_schema.go
SpadeA-Tang: For array of array of int, the elementType here should be array. element_type in FieldSchema is not allowed to be array. So, it cannot be nested array for old data_type + element_type.

pkg/util/typeutil/field_schema.go
SpadeA-Tang: Yes, I will fix it.

internal/util/indexparamcheck/index_params_validation.go
SpadeA-Tang: You mean, if we receive a type such as  Array<Int> but set the types in TypeSchema without data_type and element_type, we should reject it?

internal/storage/utils.go
SpadeA-Tang: element_type is set be array is illegal. I originally intended to check whether type_schema was set. I will fix it.

pkg/util/typeutil/field_schema.go
zhengbuqian: yes, will that be clearer? changing name now is of zero effort.

SpadeA-Tang:
@zhengbuqian PTAL

internal/util/indexparamcheck/index_params_validation.go
zhengbuqian: * legacy type: continue using `data_type` + `element_type`, can optionally set `type_schema`
* schematic type: may only set `type_schema`

wdyt?

internal/core/src/common/FieldMeta.h
```diff
@@ -255,6 +264,34 @@ class FieldMeta {
...
+            return *type_schema_;
+        }
+        milvus::proto::schema::TypeSchema type_schema;
+        auto* element = type_schema.mutable_array_element();
+        element->set_leaf_type(ToProtoDataType(element_type_));
+        return type_schema;
+    }
+
+    bool
+    has_element_schema() const {
```
zhengbuqian: this method is unclear to me? are we using this to decide if this field is exactly 2 level nested `array<array<T>>`?

now after `type_schema_.has_value()` it assumes the `kind` to be `array_element`? and then it checks the `array_element` of `type_schema_.array_element`. maybe it is not what you wanted.

internal/core/src/common/FieldMeta.h
```diff
@@ -255,6 +264,34 @@ class FieldMeta {
...
+        return type_schema_.has_value() &&
+               type_schema_->array_element().has_array_element();
+    }
+
+    const milvus::proto::schema::TypeSchema&
+    get_element_schema() const {
+        Assert(has_element_schema());
+        return type_schema_->array_element();
```
zhengbuqian: `get_element_schema()` returns the 1st level `kind` of `type_schema_` but `has_element_schema()` checked the 2nd level `kind`.

internal/core/src/common/FieldMeta.h
```diff
@@ -255,6 +264,34 @@ class FieldMeta {
         return element_type_;
     }
 
+    DataType
+    get_array_element_type() const {
```
zhengbuqian: how does this differ from `get_element_type()`?

internal/core/src/common/FieldMeta.h
SpadeA-Tang: I changed it to `is_nested_array`.

internal/core/src/common/FieldMeta.h
SpadeA-Tang: This method is deleted.

internal/core/src/common/FieldMeta.h
SpadeA-Tang: This method is deleted.

internal/core/src/storage/Util.cpp
SpadeA-Tang: No, it's in for loop.

pkg/util/typeutil/field_schema.go
zhengbuqian: ping? any thoughts?

pkg/util/typeutil/field_schema.go
```diff
@@ -89,3 +89,82 @@ func CreateFieldSchemaHelper(schema *schemapb.FieldSchema) *FieldSchemaHelper {
...
+				field.GetName())
+		}
+		return nil
+	}
+
+	if err := validateTypeSchemaNode(field.GetName(), typeSchema); err != nil {
+		return err
+	}
+	if !IsNestedArrayTypeSchema(typeSchema) {
+		return merr.WrapErrParameterInvalidMsg(
```
zhengbuqian, line 200: I thought our previous discussion was:

* legacy type: continue using data_type + element_type, can optionally set type_schema
* schematic type: may only set type_schema

wdyt?

internal/proxy/util.go
```diff
@@ -571,21 +571,128 @@ func validateElementType(dataType schemapb.DataType) error {
 	return merr.WrapErrParameterInvalidMsg("element type %s is not supported", dataType.String())
 }
 
+func validateNestedArrayTypeSchema(collectionName string, fieldName string, typeSchema *schemapb.TypeSchema) (int64, error) {
```
zhengbuqian: validation performed in proxy and root coord is somewhat duplicated but not identical? consider consolidate the validation logic in 1 place

pkg/util/typeutil/field_schema.go
```diff
@@ -89,3 +89,82 @@ func CreateFieldSchemaHelper(schema *schemapb.FieldSchema) *FieldSchemaHelper {
...
+		return false
+	}
+	_, ok := elementSchema.GetKind().(*schemapb.TypeSchema_ArrayElement)
+	return ok
+}
+
+// ValidateFieldTypeSchema validates the wire representation of nested Arrays.
+// Non-nested fields use only data_type/element_type. Nested Arrays use
+// data_type=Array, element_type=Array, and a recursive type_schema.
+func ValidateFieldTypeSchema(field *schemapb.FieldSchema) error {
```
zhengbuqian, line 184: missing leaf type check?

internal/core/src/common/ColumnarArrayChunk.h
```diff
@@ -0,0 +1,619 @@
...
+}  // namespace array_detail
+
+// A read-only view over one recursively serialized Array column chunk.
+//
+// Root layout:
+//   [validity bitmap, when nullable][alignment padding]
+//   [offsets: row_count + 1][child node][MMAP_ARRAY_PADDING]
+//
+// Nested Array node layout:
+//   [validity bitmap][alignment padding][offsets: row_count + 1][child node]
```
zhengbuqian: validity bitmap of child node should also depend on inner node's nullable?

pkg/util/typeutil/field_schema.go
SpadeA-Tang: Not this PR. We should have a separate one to make all codes use type_schema only.

internal/proxy/util.go
SpadeA-Tang: All field  validation follows this pattern where Proxy do complete DDL check while RootCoord do some light defensive check.

pkg/util/typeutil/field_schema.go
SpadeA-Tang: validateTypeSchemaNode checks leaf. Do you mean whether it's valid/supported in array?

zhengbuqian: yes, I believe we allow only certain types as T in Array<Array<T>>

pkg/util/typeutil/field_schema.go
zhengbuqian: yes but we should not return error if the sdk set both for legacy type?

internal/core/src/common/ColumnarArrayChunk.h
zhengbuqian: ping

zhengbuqian: i think this will be handled by the element null PRs right?

zhengbuqian (approved):

internal/core/src/common/ColumnarArrayChunk.h
SpadeA-Tang: The validity now is already dependended on the `nullable` of the related type schema.

SpadeA-Tang: Though it's not allowed to set it true now.

pkg/util/typeutil/field_schema.go
SpadeA-Tang: I will handle this.

congqixia (approved):
/approve
lgtm for quick temp solution
we need to address current storage format before make solid solution for in-memory one

Issue #51799, opened 2026-07-24
[Bug]: C++ unit tests mutate process-global SegcoreConfig (inline static); chunk_rows=2 leak causes deterministic UT shard timeouts


xiaocai2333:
## Summary

`SegcoreConfig`'s members are all `inline static` (class-level globals, since #27673), so the common test pattern

```cpp
SegcoreConfig config = SegcoreConfig::default_config();  // looks like a copy
config.set_chunk_rows(kChunkRows);                        // actually mutates the process-global
```

does **not** isolate anything — it silently reconfigures every later test in the same process.

#51540 added three such sites in `test_offsets_eval_correctness.cpp` with `kChunkRows = 2` and no restore. After any of them runs, every subsequent growing segment in the process is built with 2-row chunks; a 100k-row segment becomes ~50,000 chunks, and the nullable access path (`get_span_base` → `ThreadSafeChunkVector::get_element_offset`, an O(n) walk under a shared lock per call — `mmap/ChunkVector.h`) turns quadratic.

## Impact

Whether CI dies depends purely on gtest shard composition: if a `chunk_rows=2` test and the heavy `ExprTestSuite/ExprTest.TestTermJsonNullable` instances land in the same shard with no test in between that sets `chunk_rows` back to a large value, each Nullable instance goes from ~3s to ~120s under coverage instrumentation and the shard blows the 1800s budget ("shard completed with 0 tests captured"). Because sharding is deterministic per test set, any PR that adds/removes C++ tests re-rolls the layout and can hit this deterministically on every run (observed on #51486: two identical failures at 19208/19278, while test-set-preserving PRs inherit master's currently-safe layout).

## Minimal reproduction (two master-only tests)

```bash
./internal/core/output/unittest/all_tests \
  "--gtest_filter=OffsetsEvalNullableElementTest.GrowingLoadFieldDataUsesLogicalCountForNullPayloadRows:ExprTestSuite/ExprTest.TestTermJsonNullable/0"
```

`TestTermJsonNullable/0` takes ~1.5s alone, ~20s (locally) / ~120s (CI, with coverage) when run after the `chunk_rows=2` test. Verified on a binary with zero non-master changes.

## Fixes

1. Immediate (PR attached): wrap the three `set_chunk_rows` sites in the existing `ScopedSegcoreConfigRestore` RAII guard (`test_utils/SegcoreConfigUtils.h`, already used by 3 other test files). Verified red→green with the reproduction above.
[...]

PR #51800, opened 2026-07-24
fix: restore process-global SegcoreConfig after chunk_rows overrides in tests

xiaocai2333:
issue: #51799

## What

`SegcoreConfig`'s members are all `inline static`, so the test pattern `SegcoreConfig config = SegcoreConfig::default_config(); config.set_chunk_rows(2);` mutates the **process-global** config rather than a copy. The three such sites in `test_offsets_eval_correctness.cpp` (added in #51540) leave `chunk_rows = 2` behind for every later test in the process: a 100k-row growing segment becomes ~50k chunks, and the nullable access path (`get_span_base` → `ThreadSafeChunkVector::get_element_offset`, O(n) under a shared lock per call) turns quadratic — `ExprTestSuite/ExprTest.TestTermJsonNullable` goes from ~1.5s to ~120s under coverage instrumentation.

Whether CI survives depends only on gtest shard layout: if a `chunk_rows=2` site and the heavy Nullable instances share a shard with no restoring test in between, the shard exceeds the 1800s budget and dies with "shard completed with 0 tests captured". Sharding is deterministic per test set, so an affected PR fails on **every** run (observed: #51486 runs 19208/19278, identical failures; test-set-preserving PRs inherit master's currently-safe layout and pass — until any test-count change re-rolls it, including master's own).

This PR wraps the three sites in the existing `ScopedSegcoreConfigRestore` RAII guard (`test_utils/SegcoreConfigUtils.h`, already used by `SegmentGrowingIndexTest.cpp`, `ChunkedSegmentSealedBinlogIndexTest.cpp`, `flush_growing_segment_test.cpp`). No behavior change for the guarded tests themselves.

## Verification

Minimal reproduction (both tests are on master, unmodified):

```bash
./internal/core/output/unittest/all_tests \
  "--gtest_filter=OffsetsEvalNullableElementTest.GrowingLoadFieldDataUsesLogicalCountForNullPayloadRows:ExprTestSuite/ExprTest.TestTermJsonNullable/0"
```

- Before this fix: probe instances take 19.7–20.9s each (43s under load), after the 2ms poison test.
- After this fix: 1.5–1.8s, identical to running the probe alone. Red/green verified locally under ASAN.

Deeper follow-ups (root `inline static` design, ~11 milder unrestored override sites, O(n) `get_element_offset`) are catalogued in #51799.

### DIFF
--- a/internal/core/unittest/test_offsets_eval_correctness.cpp
+++ b/internal/core/unittest/test_offsets_eval_correctness.cpp
@@ -33,6 +33,7 @@
 #include "segcore/SegcoreConfig.h"
 #include "test_utils/DataGen.h"
 #include "test_utils/GenExprProto.h"
+#include "test_utils/SegcoreConfigUtils.h"
 #include "test_utils/cachinglayer_test_utils.h"
 #include "test_utils/storage_test_utils.h"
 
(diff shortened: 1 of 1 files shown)


xiaocai2333:
fixed by #51486

Issue #51788, opened 2026-07-24
[Bug]: Rollback from 3.0 to latest 2.6 leaves collections unserviceable due to index store path version incompatibility


yanliang567:
# [Bug]: Rollback from 3.0 to latest 2.6 leaves collections unserviceable due to index store path version incompatibility

## Environment

- **Milvus Version**:
  - Base: `harbor.milvus.io/milvusdb/milvus:v2.6.18`
  - Target: `harbor.milvus.io/manta/milvus:3.0-20260723-bd384f0`
  - Rollback: `harbor.milvus.io/milvusdb/milvus:2.6-20260723-d73c8e53`
- **Deployment Mode**:
  - standalone + Rocksmq
  - cluster + Pulsar
- **SDK**: PyMilvus from the QA workflow runner
- **Environment**: 4am QA Kubernetes
- **Workflows**:
  - standalone: `milvus-standalone-2-6-upgrade-rollback-frw2l`
  - cluster+Pulsar: `c26rb-pr10-837343`
- **Scenario**: `2.6.18 -> 3.0 latest -> latest 2.6 rollback`

### Option B: Steps

1. Deploy Milvus `v2.6.18`.
2. Create rollback-safe 2.6 collections and insert baseline data.
   - 3 collections
   - 5000 rows per collection
   - scalar/vector index coverage includes:
     - `STL_SORT`
     - `HNSW`
     - `BITMAP`
     - `INVERTED`
     - `AUTOINDEX`
     - `NGRAM`
     - `TRIE`
     - `DISKANN`
     - `BIN_IVF_FLAT`
     - `IVF_RABITQ`
     - `SPARSE_INVERTED_INDEX`
3. Validate baseline query/search/count/DML on 2.6.
4. Upgrade the same deployment to `harbor.milvus.io/manta/milvus:3.0-20260723-bd384f0`.
5. Keep rollback-unsafe storage features disabled:
   - `common.storage.useLoonFFI=false`
   - `dataNode.storage.format` is not set to `vortex`
   - `common.storage.jsonShreddingEnabled=false`
6. In the 3.0 phase, run post-upgrade validations:
   - index compatibility validation
   - DML/DQL against existing collections
   - create new collections and query/search them
   - rebuild/load selected vector and scalar indexes
7. Roll back Milvus to `harbor.milvus.io/milvusdb/milvus:2.6-20260723-d73c8e53`.
8. Wait for rollback serviceability and query baseline checkpoint collections.

[...]

PR #51802, opened 2026-07-24
fix: gate collection-rooted index layout behind an explicit opt-in

bigsheeper:
## Summary

The collection-rooted index layout (`index_v1`) turns on as soon as every QueryNode reports the current version. That gate only models a rolling upgrade — it has no notion of a rollback — so a `2.6 -> 3.0` upgrade starts writing index files that a 2.6 binary cannot address. After rolling back to 2.6, DataCoord rebuilds every index path under the legacy `index_files/` prefix, sealed segment load fails on the missing objects (`Path does not exist` / `GetObjectSize 404`), and collections never become query-serviceable again (`channel distribution is not serviceable`).

This PR makes the layout an explicit opt-in so an upgrade never silently produces artifacts an older binary cannot read.

## Changes

- New config `dataCoord.index.storePathVersion` (default `0` = legacy build-rooted; `1` = collection-rooted), `refreshable`, `Export: true` with docs. Only `0` and `1` are accepted; any other value — including a malformed string, which the paramtable getters silently coerce to `0` — falls back to layout 0 and logs a rated warning.
- `GetClusterMinIndexStorePathVersion` returns `BUILD_ROOTED` unless the config opts in. The existing QueryNode session-version gate stays as an **additional** condition (AND), so enabling it during a `2.6 -> 3.0` upgrade keeps building layout 0 until every QueryNode is upgraded. That gate compares the version each QueryNode publishes in its session, which is the compile-time `common.Version` constant, so it separates release lines (2.6.x vs 3.0.x) but does not differentiate binaries within the same release line — the config doc and code comment say so explicitly.
- Regenerated `configs/milvus.yaml`.

Falling back to layout 0 is safe in both directions because the layout is recorded per `SegmentIndex`: existing records keep being read and GC'd under the layout they were built with, so the switch can be flipped at runtime.

## Behavior change

A brand-new 3.0 cluster previously enabled the collection-rooted layout automatically: every QueryNode reports the current version, so the old gate returned `COLLECTION_ROOTED`. With `dataCoord.index.storePathVersion` defaulting to `0`, a fresh 3.0 deployment now builds indexes under the legacy `index_files/` layout until the config is explicitly set to `1`.

## Scope / follow-up

This governs newly built segment indexes only. Copy-segment / import-copy and snapshot restore intentionally preserve the layout recorded on the source, so a cluster with the switch off can still receive layout 1 files by copying from a cluster that had it on. Not addressed here.

Read-side rollback compatibility on the 2.6 line (letting 2.6 resolve `index_v1` paths) is a separate change and is tracked in the issue.

Making the QueryNode half of the gate meaningful *within* a release line would need a per-binary path-capability field in the session payload, rather than the release-line version constant it compares today. Also a follow-up, not this PR.

## Test Plan

- [x] Unit: config gate returns `BUILD_ROOTED` when the config is `0` (even with an all-current cluster) and when the config is `1` but any QueryNode session is older; refreshable flip; malformed value falls back to legacy.
- [x] Unit: unsupported values (`2`, `-1`, non-numeric) fall back to layout 0 instead of being read as an opt-in.
- [x] `go test -tags dynamic,test -gcflags="all=-N -l" ./internal/datacoord/` full package green.
- [ ] QA: `2.6.x -> 3.0 -> 2.6.x` rollback case asserting baseline collections become query/search serviceable (follow-up in QA workflows).

issue: #51788

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -801,6 +801,10 @@ dataCoord:
   syncSegmentsInterval: 300 # The time interval for regularly syncing segments
   index:
     memSizeEstimateMultiplier: 2 # When the memory size is not setup by index procedure, multiplier to estimate the memory size of index data
+    # Object storage layout for newly built segment index files. 0: index_files/{buildID}/{indexVersion}/{partitionID}/{segmentID}, 1: index_v1/{collectionID}/{partitionID}/{segmentID}/{buildID}/{indexVersion}.
+    # Layout 1 cannot be read by 2.6, so enabling it commits the upgrade and gives up the ability to roll back to 2.6. Existing index files keep the layout they were built with, so this switch only affects index files built from now on.
+    # Layout 1 is additionally gated on every QueryNode having been upgraded, so setting this to 1 in a mixed-version cluster still builds layout 0 until the upgrade finishes.
+    storePathVersion: 0
   enableGarbageCollection: true # Switch value to control if to enable garbage collection to clear the discarded data in MinIO or S3 service.
   gc:
     interval: 3600 # The interval at which data coord performs garbage collection, unit: second.
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -801,9 +801,9 @@ dataCoord:
   syncSegmentsInterval: 300 # The time interval for regularly syncing segments
   index:
     memSizeEstimateMultiplier: 2 # When the memory size is not setup by index procedure, multiplier to estimate the memory size of index data
-    # Object storage layout for newly built segment index files. 0: index_files/{buildID}/{indexVersion}/{partitionID}/{segmentID}, 1: index_v1/{collectionID}/{partitionID}/{segmentID}/{buildID}/{indexVersion}.
+    # Object storage layout for newly built segment index files. 0: index_files/{buildID}/{indexVersion}/{partitionID}/{segmentID}, 1: index_v1/{collectionID}/{partitionID}/{segmentID}/{buildID}/{indexVersion}. Any other value falls back to 0.
     # Layout 1 cannot be read by 2.6, so enabling it commits the upgrade and gives up the ability to roll back to 2.6. Existing index files keep the layout they were built with, so this switch only affects index files built from now on.
-    # Layout 1 is additionally gated on every QueryNode having been upgraded, so setting this to 1 in a mixed-version cluster still builds layout 0 until the upgrade finishes.
+    # Layout 1 is additionally gated on no QueryNode still reporting an older release line, so setting this to 1 during a 2.6 -> 3.0 upgrade still builds layout 0 until every QueryNode is upgraded. That check reads the version each QueryNode publishes in its session, which is a per-release-line constant, so it separates 2.6.x from 3.0.x but does not differentiate binaries within the same release line.
     storePathVersion: 0
   enableGarbageCollection: true # Switch value to control if to enable garbage collection to clear the discarded data in MinIO or S3 service.
   gc:
--- a/internal/datacoord/index_engine_version_manager.go
+++ b/internal/datacoord/index_engine_version_manager.go
@@ -13,6 +13,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/indexpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/lock"
+	"github.com/milvus-io/milvus/pkg/v3/util/metautil"
 )
 
 // IndexEngineVersionManager manages the index engine versions reported by all QueryNodes in the cluster.
@@ -125,7 +126,22 @@ func (m *versionManagerImpl) Update(session *sessionutil.Session) {
 	m.addOrUpdate(session)
 }
 
+// GetClusterMinIndexStorePathVersion returns the index file layout to use for new index builds.
+//
+// COLLECTION_ROOTED requires BOTH:
+//   - the operator to opt in via dataCoord.index.storePathVersion, because a binary older than
+//     this one cannot read that layout and the opt-in is what gives up rollback compatibility;
+//   - every QueryNode to already run this version, because QueryNodes rebuild the remote index
+//     prefix themselves (storage/FileManager.h GetRemoteIndexObjectPrefix), so an older one
+//     would look for the files under the legacy layout.
+//
+// Falling back to BUILD_ROOTED is always safe: the layout is recorded per SegmentIndex, so
+// records built earlier keep being read and GC'd under the layout they were built with.
 func (m *versionManagerImpl) GetClusterMinIndexStorePathVersion() indexpb.IndexStorePathVersion {
+	if !metautil.IsCollectionRooted(indexpb.IndexStorePathVersion(Params.DataCoordCfg.IndexStorePathVersion.GetAsInt32())) {
+		return indexpb.IndexStorePathVersion_INDEX_STORE_PATH_VERSION_BUILD_ROOTED
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
--- a/internal/datacoord/index_engine_version_manager.go
+++ b/internal/datacoord/index_engine_version_manager.go
@@ -3,6 +3,8 @@ package datacoord
 import (
 	"context"
 	"math"
+	"strconv"
+	"strings"
 
 	"github.com/blang/semver/v4"
 	"github.com/samber/lo"
@@ -13,7 +15,6 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/indexpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/lock"
-	"github.com/milvus-io/milvus/pkg/v3/util/metautil"
 )
 
 // IndexEngineVersionManager manages the index engine versions reported by all QueryNodes in the cluster.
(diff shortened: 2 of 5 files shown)


bigsheeper:
Reviewer: yihao (Codex)

I reviewed this PR and did not find a blocking correctness issue.

Checked:
- New `SegmentIndex.IndexStorePathVersion` assignment in DataCoord is gated by `GetClusterMinIndexStorePathVersion()`, so default config `0` keeps new index builds on the legacy `index_files` layout.
- DataNode index build, load, and GC continue to use persisted `IndexStorePathVersion` metadata instead of re-reading current config, so existing indexes should keep their original layout.
- copy/import/restore-style paths preserve source index metadata; I did not find a bypass that creates new collection-rooted paths while the gate is disabled.

Non-blocking suggestion:
- Consider adding a test or comment for `storePathVersion=2`. `metautil.IsCollectionRooted()` treats `>= 1` as collection-rooted, while the operator-facing config/docs describe `0/1`; this is probably intentional forward compatibility, but the config semantics are slightly implicit.

Verification note:
- GitHub CI had `ci-v2/build-ut-cov` and `ci-v2/go-sdk` passing when reviewed; `e2e-default` was still pending.
- I could not run the local targeted DataCoord test because this machine is missing `milvus_core.pc` / `PKG_CONFIG_PATH` setup.

czs007:
Adversarial review surfaced two low-severity, non-blocking items — one defensive-hardening suggestion around the config gate and one release-note gap for a default behavior change — neither of which should hold up the merge.

## Low

**`internal/datacoord/index_engine_version_manager.go:141` — config gate uses a `>=` comparison instead of normalizing to `{0, 1}`**

The gate calls `metautil.IsCollectionRooted(...)`, which is implemented as `pathVersion >= COLLECTION_ROOTED` (`pkg/util/metautil/segment_index.go:38-40`), so a misconfigured `storePathVersion: 2` also passes. Practical impact is limited: `GetClusterMinIndexStorePathVersion` only ever returns the two enum constants (lines 142-158), the raw config integer is never persisted (the sole write site is `internal/datacoord/index_inspector.go:266`, which stores the function's return value), and the C++ side (`internal/core/src/storage/FileManager.h:307-310`) likewise only ever sees `0` or `1` from `SegmentIndex`. So `2` behaves identically to `1` rather than selecting an undefined layout, and opting out of rollback capability with a non-zero value is the intended semantic — the real exposure is an operator typo (meaning `0`, typing `2`). Note there was a rating disagreement here: one reviewer scored this medium, another considered it low / optional hardening that should not block the merge.

*Suggestion:* switch the gate to an exact `config == COLLECTION_ROOTED` check for defensive normalization, leaving room for a future layout `2`, and log a warning on unsupported values. This also helps with observability of bad input: `GetAsInt32()` silently coerces non-numeric values to `0` (`pkg/util/paramtable/param_item.go:224-232`), and the current tests only assert the fallback result without any visible signal.

*(raised by chyezh, tinswzy)*

**`internal/datacoord/index_engine_version_manager.go:141` — behavior change needs a release note: brand-new 3.0 clusters now default to the legacy layout**

Before this change, `GetClusterMinIndexStorePathVersion` considered only QueryNode session versions, so a fresh 3.0 deployment (where every QN version equals `common.Version`) returned `COLLECTION_ROOTED` and automatically enabled the `index_v1/` layout. With the new early return, the condition becomes "config opt-in AND all QNs upgraded", and `dataCoord.index.storePathVersion` defaults to `"0"` (`pkg/util/paramtable/component_param.go:5985-5993`, `configs/milvus.yaml:804`). A brand-new 3.0 cluster therefore uses the legacy `index_files/` layout until the config is explicitly enabled — the PR's own test (`Test_..._IndexStorePathVersionConfigGate`) asserts exactly this default.

*Suggestion:* add a release note documenting the change so users do not assume new clusters are already on the collection-rooted layout by default.

*(raised by tinswzy)*

czs007:
## Adversarial Review Summary

One low-severity documentation issue was confirmed: the new config documentation overstates the granularity of the QueryNode gating for index storage path layout 1.

### Low

**`pkg/util/paramtable/component_param.go:5991`** (also mirrored in `configs/milvus.yaml:803`)

**Problem:** The doc string states that *"Layout 1 is additionally gated on every QueryNode having been upgraded, so setting this to 1 in a mixed-version cluster still builds layout 0 until the upgrade finishes."* The actual gate is `version.LT(common.Version)` in `internal/datacoord/index_engine_version_manager.go:152`, which compares the version each QueryNode reports in its session (`internal/util/sessionutil/session_util.go:289` sends `common.Version` verbatim). That value is a compile-time constant in `pkg/common/version.go:9` — `3.0.0-beta` on master, `2.6.21` on the 2.6 line. So the gate can only distinguish *release lines*: a 2.6 QueryNode does compare as older than `3.0.0-beta`, which makes the 2.6 → 3.0 mixed-upgrade scenario this PR targets work correctly, but every 3.0-line binary reports the identical literal `3.0.0-beta`, making `LT` always false. Within the 3.0 line the gate is effectively blind and cannot distinguish two 3.0 binaries with different index-layout read capabilities. An operator can easily read "every QueryNode having been upgraded" as per-binary capability detection. The same wording appears in the new comment at `index_engine_version_manager.go:131-136`.

**Suggestion:** Reword to something like *"...still builds layout 0 until no QueryNode reports an older release-line version (2.6.x)"*, and state explicitly that the check does not differentiate between binaries within the same release line. Update the mirrored text in `configs/milvus.yaml:803` and the code comment at `index_engine_version_manager.go:131-136` to match. Longer term, a per-binary path-capability field in the session payload would make this gate meaningful within a release line — a follow-up, not a blocker for this PR.

*(raised by tinswzy, chyezh)*

bigsheeper:
Thanks — both items verified against the current branch and picked up in `d994ace`.

**1. `>=` comparison.** Confirmed: `IsCollectionRooted` is `pathVersion >= COLLECTION_ROOTED` (`pkg/util/metautil/segment_index.go:38-40`), so `storePathVersion: 2` did pass the gate. Your impact analysis also holds — `GetClusterMinIndexStorePathVersion` only ever returns the two enum constants, the sole write site of its result is `internal/datacoord/index_inspector.go:266`, and `FileManager.h` never sees anything but 0/1 from `SegmentIndex` — so `2` behaved exactly like `1` rather than selecting an undefined layout.

Fixed anyway as defensive normalization: the gate no longer goes through `IsCollectionRooted`; it parses the raw config value and accepts only `{0, 1}`. Everything else — including a malformed string, which `GetAsInt32` silently coerces to `0` with no signal today — falls back to the legacy layout and logs a rated warning, so an operator typo is visible instead of being read as an opt-in. Unit coverage added for `2` and `-1`. Behavior for the two supported values is unchanged.

**2. Release note.** Correct — with the config defaulting to `0`, a brand-new 3.0 cluster now uses the legacy `index_files/` layout until the operator opts in, where before it auto-enabled `index_v1/`. This repo has no release-note block in the PR flow, so I've called it out explicitly under a new **Behavior change** section in the PR description, and it is captured on issue #51788.

bigsheeper:
Confirmed and fixed in `d994ace`.

Verified the mechanism end to end: the gate is `version.LT(common.Version)` (`internal/datacoord/index_engine_version_manager.go`) against the version each QueryNode publishes in its session (`internal/util/sessionutil/session_util.go:289`), and that value is the compile-time constant in `pkg/common/version.go:9` — `3.0.0-beta` on master. So the check separates release lines (a 2.6 QueryNode does compare older, which is what makes the `2.6 -> 3.0` case this PR targets work) but is blind within the 3.0 line, where every binary reports the identical literal.

Reworded all three places to say exactly that — the config doc in `pkg/util/paramtable/component_param.go`, the mirrored text in `configs/milvus.yaml`, and the comment above `GetClusterMinIndexStorePathVersion`: layout 1 waits until no QueryNode still reports an older release line, and the check does not differentiate binaries within the same release line.

Agreed that a per-binary path-capability field in the session payload is the real fix for intra-release-line gating; recorded it as a follow-up in the PR description rather than doing it here.

czs007:

Verified:

- `internal/datacoord/index_engine_version_manager.go:37` — the config is read as a raw string and parsed with an explicit `strconv.ParseInt` rather than a coercing getter, so malformed, negative, and unsupported-but-parseable values (`"2"`, `"-1"`) warn and fall back to `BUILD_ROOTED` instead of being silently indistinguishable from `"0"`.
- `internal/datacoord/index_engine_version_manager.go:66` — the config check returns `BUILD_ROOTED` before `m.mu` is acquired and then falls through to the existing session-version check, so the two independent conditions are ANDed without introducing a lock-ordering or atomicity concern.
- `internal/datacoord/index_engine_version_manager.go:168` — every read of `sessionVersion` stays under the same mutex used by the startup/add/remove/update paths, so the new gate adds no unsynchronized state.
- `internal/datacoord/index_inspector.go:266`, `internal/datacoord/task_index.go:500` — the selected layout is stamped onto the new `SegmentIndex` at `AddSegmentIndex` and the persisted per-segment value is propagated into `CreateJobRequest`, so the gate governs the actual build path and a default-config `2.6 → 3.0` upgrade writes build-rooted files.
- `internal/datacoord/handler.go:878`, `internal/datacoord/garbage_collector.go:1510` — index loading and GC both derive paths from each persisted `SegmentIndex.IndexStorePathVersion`, so flipping the refreshable config from `1` back to `0` affects only new builds and never reinterprets already-built indexes.
- `internal/datacoord/index_engine_version_manager_test.go:65`, `internal/datacoord/index_engine_version_manager_test.go:103` — coverage exercises both halves of the gate (default, opt-in, flip-back, malformed/out-of-range/negative fallbacks, plus empty and mixed-version clusters), and the tests run sequentially with `defer Reset`, so the default-value assertion is not order-fragile.

czs007 (approved):
/lgtm

Issue #51765, opened 2026-07-23
[Bug]: StreamingNode WAL recovery deadlocks with HWM sealing and leaks unbounded goroutines


yanliang567:
## Environment

- **Milvus Version**: `harbor.milvus.io/milvusdb/milvus:master-20260720-b9621d7d`, commit `b9621d7d325dea78fc8fd33c9d2ba0dcbda24c04`
- **Latest master checked**: `955f761fcd07ef4406d481d4f53100ee06318362`; the critical code path is unchanged
- **Deployment Mode**: cluster, deployed by Milvus Operator `1.3.6`
- **Topology**: 2 StreamingNodes, 6 QueryNodes, 4 DataNodes, 1 MixCoord, 1 Proxy
- **MQ**: Pulsar
- **SDK**: test SDK `2.7.0rc175`, client version parameter `2.6`
- **Kubernetes**: 4AM cluster, namespace `qa-milvus`

### Option B: Steps

1. Deploy a Milvus cluster with StreamingNode enabled and two StreamingNode replicas.
2. Use 16 DML channels and large segments (`dataCoord.segment.maxSize=4096`).
3. Create a channel recovery containing growing segments while the StreamingNode global flush pressure can trigger the HWM seal policy.
4. During WAL recovery, allow `RecoverShardManager()` to register recovered segment stats before the WAL Future is set.
5. Run a mixed search/query/upsert workload to continue producing segment-capacity seal notifications.
6. Observe one recovered WAL staying in `in_recovery`, all automatic segment sealing on that StreamingNode stopping, and goroutine count growing without bound.

Observed workflow:

- Argo: https://argo-workflows.zilliz.cc/workflows/qa/laion1b-test-cron-sn-nullable-1784728800?tab=workflow
- Workflow window: `2026-07-22 14:00:00` to `21:13:54 UTC`
- Workload operations: search, query, load, and upsert

### Trigger Conditions

- **Frequency**: timing-sensitive; one fully diagnosed occurrence, persistent until process restart
- **First observed after**: WAL recovery of `dml_1@term100`; deadlock begins in the same millisecond as HWM sealing
- **Amplifier**: continued writes generate capacity seal notifications after the global seal worker is blocked
- **Does NOT recover when**: the workload stops, channel ownership moves to a newer term, or DataCoord/index/compaction queues are empty

## Expected Behavior

- WAL recovery should finish or abort cleanly when its context is canceled.
[...]

PR #51803, opened 2026-07-24
fix: prevent WAL recovery deadlock

chyezh:
issue: #51765

- WAL recovery: wait for WAL readiness asynchronously and stop pending flush work on cancellation
- seal scheduling: bound seal notifications without spawning one goroutine per event
- WAL open lifecycle: clean partially initialized resources exactly once on failure

### DIFF
--- a/internal/streamingnode/server/wal/adaptor/opener.go
+++ b/internal/streamingnode/server/wal/adaptor/opener.go
@@ -80,6 +80,38 @@ type openerAdaptorImpl struct {
 	interceptorBuilders []interceptors.InterceptorBuilder
 }
 
+type walOpenResources struct {
+	once            sync.Once
+	released        bool
+	roWAL           *roWALAdaptorImpl
+	param           *interceptors.InterceptorBuildParam
+	recoveryStorage recovery.RecoveryStorage
+	flusher         *flusherimpl.WALFlusherImpl
+}
+
+func (r *walOpenResources) Close() {
+	if r.released {
+		return
+	}
+	r.once.Do(func() {
+		if r.flusher != nil {
+			r.flusher.Close()
+		} else if r.recoveryStorage != nil {
+			r.recoveryStorage.Close()
+		}
+		if r.param != nil {
+			r.param.Clear()
+		}
+		if r.roWAL != nil {
+			r.roWAL.Close()
+		}
+	})
+}
+
+func (r *walOpenResources) Release() {
+	r.released = true
+}
+
 // Open opens a wal instance for the channel.
 func (o *openerAdaptorImpl) Open(ctx context.Context, opt *wal.OpenOption) (wal.WAL, error) {
 	if !o.lifetime.Add(typeutil.LifetimeStateWorking) {
@@ -190,6 +222,9 @@ func (o *openerAdaptorImpl) openRWWAL(ctx context.Context, l walimpls.WALImpls, 
 	roWAL := adaptImplsToROWAL(l, func() {
 		o.walInstances.Remove(id)
 	})
+	resources := &walOpenResources{roWAL: roWAL}
+	defer resources.Close()
+
 	cpProto, err := resource.Resource().StreamingNodeCatalog().GetConsumeCheckpoint(ctx, opt.Channel.Name)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get checkpoint from catalog")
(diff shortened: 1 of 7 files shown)


tinswzy:

Reviewed at head `c415ef1d` against master. This is a clean, well-tested fix for the three related WAL-recovery issues. No blocking problems found — the core changes are correct. A few low/nit observations below.

## Verification of the core fixes

- **Deadlock fix (`segment_flush_worker.go:23-43`)** — correct. `shardManagerImpl.AsyncFlushSegment` (`shard_manager_segment.go:242`) holds `m.mu.Lock()` across `pm.AsyncFlushSegment → asyncFlushSegment`, which previously called the blocking `m.wal.Get()` while the recovery path that resolves the WAL future needs that same lock. Moving `GetWithContext(ctx)` into the goroutine releases the lock as intended. Goroutine-leak safety is also fine: the `ctx` is the shard manager ctx (`shard_manager.go:67`), cancelled in `Close()` (`:366`), so a never-ready WAL unblocks on shutdown rather than leaking. Good test coverage in `TestAsyncFlushSegmentDoesNotHoldShardManagerLockWhileWaitingForWAL`.

- **Resource cleanup (`wal_open_resources.go`)** — the `else if` between `flusher` and `recoveryStorage` in `Close()` is correct, not a leak: whenever `resources.flusher` is set it was constructed with the same `rs` (`opener.go:253-260`, `:326-331`), and `WALFlusherImpl.Close()` closes it (`wal_flusher.go:149`). This actually removes a latent double-`Close()` of the recovery storage from the old flushing-stage path (which did `rs.Close()` *and* `flusher.Close()`). The replicate-recovery failure path is now cleaned up too — deferred `Close() → param.Clear() → ShardManager.Close()` unregisters the seal operator (`interceptor.go:42-43`), matching `TestOpenRWWALCleansRecoveredShardManagerOnReplicateRecoveryFailure`.

## Findings

**Low — `stats_seal_worker.go:40-45`: seal signal can be silently dropped.**
`NotifySealSegment` now drops the notification when the 100-entry `sealNotifier` is full (`select { … default: }`). This is a good bound versus the old unbounded-goroutine send, but note the failure mode: the consumer `loop()` calls `asyncMustSealSegment → operator.AsyncFlushSegment`, which takes the shard-manager lock (`shard_manager_segment.go:242`). During the very lock-contention window this PR addresses, the consumer can stall, the buffer fills, and `PolicyCapacity` (`stats_manager.go:214`) / `PolicyBinlogNumber` (`:379`) signals get dropped. It is not a permanent leak — both call sites re-trigger on the next `AllocRows`/`UpdateOnSync`, and the 1-minute timer is a backstop — but for a segment that crosses capacity and then goes idle, the seal can be deferred until the time policy (segment maxLife) fires rather than promptly. Consider logging drops at debug level for observability, or a short comment documenting the "drop-and-rely-on-retrigger" contract.

**Nit — `wal_open_resources.go:25-29`:** the flusher/recoveryStorage ownership coupling described above is non-obvious. A one-line comment ("flusher owns and closes recoveryStorage; do not close it here when flusher is set") would prevent a future refactor from reintroducing either a leak (flusher set without rs) or a double-close.

**Nit — `wal_open_resources.go:13,39-41`:** `released` is read in `Close()` and written in `Release()` without synchronization. Correct today because all access is on the single `openRWWAL` goroutine (the `defer` runs on the same goroutine after `Release()`), but the `sync.Once` field implies concurrency that doesn't actually exist here. Worth a brief comment noting the single-goroutine assumption, or folding `released` into the `once`/an atomic if that assumption might change.

Nothing here blocks merge.

chyezh:
Follow-up on the Team II review:

- **Dropped seal notification:** assessed as an intentional bounded-backpressure trade-off. After this PR, `AsyncFlushSegment` no longer waits for WAL readiness synchronously. HWM, memory, and time policies bypass `sealNotifier`; only capacity and binlog notifications use it. With a 100-entry buffer and a short non-blocking consumer path, a full queue is unlikely in normal operation. If it does occur, active segments retrigger through later allocation or sync events, while time policies provide a backstop for inactive segments. No additional functional change is needed.
- **Resource ownership:** confirmed that `WALFlusherImpl.Close()` owns and closes `RecoveryStorage`. The existing `if / else if` in `walOpenResources.Close()` is required to avoid a double close.
- **`released` synchronization:** `Release()` and the deferred `Close()` are called by the same `openRWWAL` goroutine. `sync.Once` only makes repeated `Close()` calls idempotent; it does not imply concurrent access with `Release()`.

These invariants are now documented in code by commit `7b88c1648f`. No further functional fix is planned for these findings.

liliu-z (commented):
/lgtm
/approve

PR #51804, opened 2026-07-24
fix: [cp 3.0] prevent WAL recovery deadlock

chyezh:
issue: #51765
pr: #51803

- WAL recovery: wait for WAL readiness asynchronously and stop pending flush work on cancellation
- seal scheduling: bound seal notifications without spawning one goroutine per event
- WAL open lifecycle: clean partially initialized resources exactly once on failure

### DIFF
--- a/internal/streamingnode/server/wal/adaptor/opener.go
+++ b/internal/streamingnode/server/wal/adaptor/opener.go
@@ -80,6 +80,38 @@ type openerAdaptorImpl struct {
 	interceptorBuilders []interceptors.InterceptorBuilder
 }
 
+type walOpenResources struct {
+	once            sync.Once
+	released        bool
+	roWAL           *roWALAdaptorImpl
+	param           *interceptors.InterceptorBuildParam
+	recoveryStorage recovery.RecoveryStorage
+	flusher         *flusherimpl.WALFlusherImpl
+}
+
+func (r *walOpenResources) Close() {
+	if r.released {
+		return
+	}
+	r.once.Do(func() {
+		if r.flusher != nil {
+			r.flusher.Close()
+		} else if r.recoveryStorage != nil {
+			r.recoveryStorage.Close()
+		}
+		if r.param != nil {
+			r.param.Clear()
+		}
+		if r.roWAL != nil {
+			r.roWAL.Close()
+		}
+	})
+}
+
+func (r *walOpenResources) Release() {
+	r.released = true
+}
+
 // Open opens a wal instance for the channel.
 func (o *openerAdaptorImpl) Open(ctx context.Context, opt *wal.OpenOption) (wal.WAL, error) {
 	if !o.lifetime.Add(typeutil.LifetimeStateWorking) {
@@ -190,6 +222,9 @@ func (o *openerAdaptorImpl) openRWWAL(ctx context.Context, l walimpls.WALImpls, 
 	roWAL := adaptImplsToROWAL(l, func() {
 		o.walInstances.Remove(id)
 	})
+	resources := &walOpenResources{roWAL: roWAL}
+	defer resources.Close()
+
 	cpProto, err := resource.Resource().StreamingNodeCatalog().GetConsumeCheckpoint(ctx, opt.Channel.Name)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get checkpoint from catalog")
@@ -199,20 +234,19 @@ func (o *openerAdaptorImpl) openRWWAL(ctx context.Context, l walimpls.WALImpls, 
 	// recover the wal state.
 	param, err := buildInterceptorParams(ctx, l, cp)
 	if err != nil {
-		roWAL.Close()
 		return nil, errors.Wrap(err, "when building interceptor params")
 	}
+	resources.param = param
 	rs, snapshot, err := recovery.RecoverRecoveryStorage(ctx, newRecoveryStreamBuilder(roWAL), cp, param.LastTimeTickMessage)
 	if err != nil {
-		param.Clear()
-		roWAL.Close()
 		return nil, errors.Wrap(err, "when recovering recovery storage")
 	}
+	resources.recoveryStorage = rs
 
 	// Handle alter WAL if found in snapshot
 	// This flushes all remaining data and triggers WAL switch to the target implementation
 	if snapshot.AlterWALInfo != nil && snapshot.AlterWALInfo.FoundAlterWALMsg {
-		return o.handleAlterWAL(ctx, l, opt, roWAL, param, rs, snapshot)
+		return o.handleAlterWAL(ctx, opt, roWAL, rs, resources, snapshot)
 	}
 
 	param.LastConfirmedMessageID = determineLastConfirmedMessageID(param.LastTimeTickMessage.MessageID(), snapshot.TxnBuffer)
@@ -255,9 +289,11 @@ func (o *openerAdaptorImpl) openRWWAL(ctx context.Context, l walimpls.WALImpls, 
 			RecoverySnapshot:   snapshot,
 			RateLimitComponent: roWAL.WALRateLimitComponent,
 		})
+		resources.flusher = flusher
 	}
 	wal := adaptImplsToRWWAL(roWAL, o.interceptorBuilders, param, flusher)
 	o.walInstances.Insert(id, wal)
+	resources.Release()
 	return wal, nil
 }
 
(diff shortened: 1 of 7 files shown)


tinswzy:

Reviewed the full diff against the `3.0` base. This is a faithful, self-consistent backport of the master fix (#51803). The three changes are coherent and I traced the resource-lifecycle and deadlock logic end-to-end. No blocking issues — I recommend merge after the CI cross-check below. Notes are non-blocking.

**Overall:** mergeable; findings are low/nit only.

### Verified correct (not issues, for reviewer confidence)
- `wal_open_resources.go:24-29` — `Close()` deliberately calls `flusher.Close()` **or** `recoveryStorage.Close()`, never both. This is correct because `WALFlusherImpl.Close()` closes the recovery storage internally (`internal/streamingnode/server/flusher/flusherimpl/wal_flusher.go:149`). This replaces the old `handleAlterWALFlushingStage` code that called `rs.Close()` *and* `flusher.Close()` — the new single-owner path avoids a double-close of the recovery storage. Good.
- `opener.go:194` + `264` — `defer resources.Close()` with `resources.Release()` on the success path is idempotent: `Release()` flips `released`, and the alter-WAL path calls `resources.Close()` internally then the deferred `Close()` no-ops via `sync.Once`. On every early-return error, the deferred `Close()` reclaims `roWAL`/`param`/`rs`/`flusher`. No leak, no double-free.
- `segment_flush_worker.go:23-30` — moving `m.wal.GetWithContext(ctx)` inside the goroutine (was `m.wal.Get()` executed synchronously by the caller holding the shard-manager lock) is the actual deadlock fix. `GetWithContext` exists on `syncutil.Future` in 3.0 (`pkg/util/syncutil/future.go:27`), and `ctx` is `m.ctx` (partition-manager scoped, cancelled on close), so the goroutine unblocks on shutdown rather than leaking.

### low — dropped capacity-seal notification for an idle full segment
`stats/stats_seal_worker.go:40-45`: switching from a per-event goroutine to a non-blocking `select { case ch <- …: default: }` means that when `sealNotifier` (buffer 100) is full, a `PolicyCapacity`/`PolicyBinlogNumber` notification is silently dropped.
- Self-healing for active segments: `StatsManager.AllocRows` (`stats_manager.go:212-214`) re-notifies on the next insert while `shouldBeSealed` stays true, so a still-written segment gets re-queued.
- Gap: a segment that reaches capacity and then receives **no further inserts** while the queue is saturated relies entirely on the 1-minute time-policy sweep (`loop()` → `notifyToSealSegmentWithTimePolicy`) as a backstop, delaying its seal by up to `defaultSealWorkerTimerInterval`. This is an acceptable trade-off under overload (a full 100-deep queue already implies the worker is sealing aggressively), but worth calling out since the old code guaranteed eventual enqueue. No change requested; flagging for awareness.

### nit — CI/backport hygiene
- The new tests pull in symbols that must exist on `3.0`: `replicates.RecoverReplicateManager` and `mock_utils.MockSealOperator` (`opener_test.go`), and `syncutil`/`policy`/`utils` helpers plus `MiniCluster`-free `newShardManagerWithGrowingSegment` (`shard_manager_test.go`). They appear present in this branch, but please confirm the streamingnode unit suites are green in CI on `3.0` — mock/import drift between master and `3.0` is the usual backport failure mode here.
- `walOpenResources.released` is read/written without synchronization. This is safe today because `Close()`/`Release()` are only ever called from the single `openRWWAL` goroutine, but a short comment stating that invariant would prevent a future concurrent caller from being introduced unsafely.

Nothing here blocks the backport.

chyezh:
Follow-up on the Team II review:

- **Dropped seal notification:** assessed as an intentional bounded-backpressure trade-off. The deadlock fix removes the synchronous WAL readiness wait from the seal consumer. HWM, memory, and time policies bypass `sealNotifier`; only capacity and binlog notifications use it. With a 100-entry buffer and a short non-blocking consumer path, a full queue is unlikely in normal operation. Active segments retrigger through later allocation or sync events, while time policies provide a backstop for inactive segments. No additional functional change is needed.
- **Resource ownership:** confirmed that `WALFlusherImpl.Close()` owns and closes `RecoveryStorage`, so the `if / else if` cleanup is correct and avoids a double close.
- **`released` synchronization:** `Release()` and deferred `Close()` remain on the same `openRWWAL` goroutine; no current data race exists.
- **Backport hygiene:** the full 3.0 CI loop passed, including build, code check, Go and integration unit tests, and C++ coverage tests.

No blocking issue remains from these findings.

tinswzy:

**team2 automated re-review** at head `a1edd4b0` (single agent, for reference). Re-reviewed the full 3.0 backport with focus on the delta since the previous review (head `88d6ac3`).

**Overall: mergeable.** No blocking issues. The delta since the last review is a single new commit that is a *pure code-move*, and I re-verified the resource-lifecycle, deadlock, and bounded-seal behavior end-to-end against the code at this head. CI loop is green at `a1edd4b` (build/code-check/UT-GO/UT-integration/CPP-cov).

### Delta since previous review — verified safe
- `a1edd4b (isolate WAL open resource cleanup)` is a **no-logic-change move**: `walOpenResources` struct + `Close()`/`Release()` are lifted verbatim from `opener.go` into the new `wal_open_resources.go`, same package `adaptor`, identical bodies (`sync.Once` + `released` guard, `if flusher != nil … else if recoveryStorage != nil …`). No dangling import left behind — `sync` is still used in `opener.go:76` (`sync.Mutex`), and `recovery`/`flusherimpl`/`interceptors` remain referenced by `openRWWAL`. Nothing else in the diff changed from the previously reviewed revision.

### Parity / lifecycle re-verification (not issues — for reviewer confidence)
- **Double-close avoidance is correct.** `WALFlusherImpl.Close()` owns and closes the recovery storage (`flusher/flusherimpl/wal_flusher.go:149` → `impl.RecoveryStorage.Close()`), so `walOpenResources.Close()`'s `flusher` XOR `recoveryStorage` branch is right: on the flusher path only `flusher.Close()` runs (which closes `rs`), on the `DisableFlusher`/pre-flusher path `recoveryStorage.Close()` runs directly.
- **No use-after-close in the alter-WAL path.** `handleAlterWALFlushingStage` calls `resources.Close()` (`opener.go:394`) and afterward touches only `snapshot`/`opt`/catalog — never `rs`/`param`/`roWAL`. The subsequent `handleAlterWALAdvanceCheckpointsStage` uses only catalog + MixCoord client, so closing `roWAL`/`rs` before it is safe (matches the original pre-refactor ordering). The deferred `resources.Close()` in `openRWWAL` then no-ops via `sync.Once`. No double-free, no leak on any early-return.
- **Shard-manager cleanup on the replicate-recovery-failure path actually fires.** The new `TestOpenRWWALCleansRecoveredShardManagerOnReplicateRecoveryFailure` exercises the real leak this refactor closes: on `RecoverReplicateManager` failure the deferred `resources.Close()` → `param.Clear()` → `ShardManager.Close()` (`interceptors/interceptor.go:43-45`), which unregisters the seal operators so a later `RegisterSealOperator` does not conflict. Confirmed `resources.param` is set before that failure point, so the recovered shard manager is genuinely reclaimed.
- **Deadlock fix is intact.** `segment_flush_worker.go` moves the blocking `m.wal.Get()` into the spawned goroutine as `m.wal.GetWithContext(ctx)`; `GetWithContext` exists on `syncutil.Future` in 3.0 (`pkg/util/syncutil/future.go:27`), and `ctx` cancellation makes the goroutine exit cleanly instead of blocking a shard-manager-locked caller.

### low — dropped capacity/binlog seal notification under a saturated queue (unchanged, acceptable)
`stats/stats_seal_worker.go:40-44`: the non-blocking `select { case ch <- …: default: }` drops a `PolicyCapacity`/`PolicyBinlogNumber` notification when `sealNotifier` (buffer 100) is full. Re-confirmed the backstops at this head:
- Only capacity (`stats_manager.go:214`) and binlog-number (`stats_manager.go:379`) route through the bounded channel; HWM/memory/time/blocking-L0 policies use separate notifiers or the timer and are unaffected.
- Active segments re-trigger via `AllocRows` re-notify on the next insert while `shouldBeSealed` holds; a full-but-idle segment falls back to the time-policy sweep (`loop()` → `notifyToSealSegmentWithTimePolicy`, ticker `defaultSealWorkerTimerInterval = 1m`, then subject to the time-policy threshold).

This is the intended bounded-backpressure trade-off already acknowledged by the author; not requesting a change. Minor note for accuracy: the fallback delay for a truly idle full segment is bounded by the time-policy selection criteria, not strictly one 1-minute tick — worth keeping in mind if a future report shows delayed seals under sustained overload.

### nit
- `walOpenResources.released` is still read/written without synchronization. Safe today because `Release()`/`Close()` execute only on the single `openRWWAL` goroutine; a one-line comment stating that single-goroutine invariant would guard against a future concurrent caller being added unsafely.

Nothing here blocks the backport.

liliu-z (commented):
/lgtm
/approve

Issue #41767, opened 2025-05-12
[Bug]: [IvfRabitq]The default nprobe value for searching is confusing


yanliang567:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master-20250508-e2ddbe49-amd64
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus 2.6.0rc119
```

### Current Behavior

1. create IVFRABITQ index with 
index_params.add_index(field_name=self.float_vector_field_name,
                               metric_type="COSINE",
                               index_type="IVF_RABITQ",
                               params={"nlist": 128, "refine": 'true', "refine_type": "SQ8"})
2. load 
3. search
3.1 if searching with "params": {"nprobe": "32"},  it returns success. I guess it uses  8 as default value after converting failure
3.2 if searching with "params": {"nprobe": None}, it returns error.  the error: 
`pymilvus.exceptions.MilvusException: <MilvusException: (code=65535, message=fail to search on QueryNode 1: worker(1) query failed: N6milvus21ExecOperatorExceptionE :Operator::GetOutput failed for [Operator:PhyVectorSearchNode, plan node id: 153933] :  => failed to search: config={{"k":10,"metric_type":"COSINE","nprobe":null,"span_id":"3d25d910c3c88201","trace_flags":0,"trace_id":"fe0c094ce5c6f362c743e26a48846750"}} type conflict in json: Type conflict in json: param 'nprobe' (null) should be integer at /workspace/source/internal/core/src/index/VectorMemIndex.cpp:404`

### Expected Behavior

the two checking failures lead to different behaviors, which makes user feel confusing.
shall we use default value for any datatype checking failure? or shall we return errors for any checking failures? 

### Steps To Reproduce

```markdown

```

foxspy:
This is by design. The upper layer milvus does json to map[string]string, and converts it back to json when calling knowhere. To ensure compatibility, knowhere will convert the string back to the corresponding type according to the parameter type definition in knowhere （see https://github.com/zilliztech/knowhere/blob/main/src/common/config.cc#L66）.

PR #51809, opened 2026-07-24
fix: validate vector search params via knowhere Config::Load at plan creation (#47729)

### LINKED ISSUE #47729: [Bug]: Index parameter nprobe validation missing - accepts nprobe=0

### Environment

```markdown
- Milvus version:v2.6.10
- Deployment mode(standalone or cluster):Standalone
- MQ type(rocksmq, pulsar or kafka):rocksmq
- SDK version(e.g. pymilvus v2.0.0rc2):PyMilvus v2.6.10
- OS(Ubuntu or CentOS): Windows (Client) / Linux (Server)
- CPU/Memory:N/A 
- GPU:N/A 
- Others: Docker Compose deployment
```

### Current Behavior

Milvus v2.6.10 accepts `nprobe=0` when searching with IVF indexes (IVF_FLAT, IVF_PQ, IVF_SQ8, SCANN, BIN_IVF_FLAT), which violates the documented requirement that `nprobe` must be a positive integer (> 0).

### Expected Behavior

Search should fail with a clear error message indicating that `nprobe` must be a positive integer greater than 0.

Example expected error:
```
<ParamError: (code=1100, message=Invalid search parameter 'nprobe': value must be greater than 0, but got 0)>
```

### Steps To Reproduce

```markdown
1. Create a collection with vector field
2. Insert test data into collection
3. Create an IVF index with valid `nlist` parameter
4. Attempt to search with `nprobe=0`
5. Observe that search succeeds (should fail)
**Complete Reproduction Script:**

from pymilvus import connections, Collection, FieldSchema, CollectionSchema, DataType, utility

# Connect to Milvus
connections.connect(alias="default", host="localhost", port="19530")

def cleanup_collection(collection_name):
    """Clean up existing collection"""
    if utility.has_collection(collection_name):
        collection = Collection(collection_name)
        collection.drop()
        print(f"Cleaned up: {collection_name}")

# Test IVF_FLAT with nprobe=0
collection_name = "test_ivf_flat_nprobe_0"
cleanup_collection(collection_name)

# Create collection
fields = [
    FieldSchema(name="id", dtype=DataType.INT64, is_primary=True, auto_id=False),
    FieldSchema(name="vector", dtype=DataType.FLOAT_VECTOR, dim=128)
]
schema = CollectionSchema(fields=fields)
collection = Collection(name=collection_name, schema=schema)

# Insert test data
data = [[i for i in range(10)], [[0.1]*128 for _ in range(10)]]
collection.insert(data)
collection.flush()

# Create IVF_FLAT index
index_params = {
    "metric_type": "L2",
    "index_type": "IVF_FLAT",
    "params": {"nlist": 100}
}
collection.create_index(field_name="vector", index_params=index_params)

# Load collection
collection.load()

# Try to search with nprobe=0 (invalid)
search_params = {"metric_type": "L2", "params": {"nprobe": 0}}
res = collection.search(
    data=[[0.1]*128],
    anns_field="vector",
    param=search_params,
    limit=10
)
# Result: Search succeeds (should fail) - BUG CONFIRMED

# Cleanup
cleanup_collection(collection_name)
connections.disconnect("default")


**IVF_PQ with nprobe=0:**

# Use same setup as above, but with IVF_PQ index
index_params = {
    "metric_type": "L2",
    "index_type": "IVF_PQ",
    "params": {"nlist": 100, "m": 16, "nbits": 8}
}
collection.create_index(field_name="vector", index_params=index_params)

# Search with nprobe=0
search_params = {"metric_type": "L2", "params": {"nprobe": 0}}
res = collection.search(
    data=[[0.1]*128],
    anns_field="vector",
    param=search_params,
    limit=10
)
# Result: Search succeeds (should fail) - BUG CONFIRMED


**SCANN with nprobe=0:**

# Use same setup as above, but with SCANN index
index_params = {
    "metric_type": "L2",
    "index_type": "SCANN",
    "params": {"nlist": 100}
}
collection.create_index(field_name="vector", index_params=index_params)

# Search with nprobe=0
search_params = {"metric_type": "L2", "params": {"nprobe": 0}}
res = collection.search(
    data=[[0.1]*128],
    anns_field="vector",
    param=search_params,
    limit=10
)
# Result: Search succeeds (should fail) - BUG CONFIRMED


**Note:** The `Collection` API is used instead of `MilvusClient` API to avoid automatic index creation issues. The `MilvusClient` API automatically creates a default index when calling `create_collection()`, which prevents creating a custom IVF index afterward.
```

### Milvus Log

N/A - Issue is reproducible via SDK behavior observation.

### Contract Violation

1. **Parameter Validation Contract**: According to [Milvus Index Documentation](https://milvus.io/docs/index.md), `nprobe` parameter must be a positive integer for IVF indexes.

2. **Documentation References**:
   - [Milvus Index Documentation](https://milvus.io/docs/index.md) - Official documentation for index parameters
   - `nprobe` is a search parameter for IVF indexes (IVF_FLAT, IVF_PQ, IVF_SQ8, SCANN)
   - Parameter definition: "nprobe: Number of cluster units to search... Must be a positive integer."
   - Multiple technical articles and community resources confirm this parameter requirement

### Impact

- **Incorrect Search Behavior**: Searching with `nprobe=0` may lead to undefined search results or performance issues
- **User Confusion**: Users may perform searches with invalid parameters without realizing the parameter is invalid
- **Documentation Violation**: Actual behavior contradicts documented requirements

### Suggested Fix

1. **Add Parameter Validation**:
   - Validate `nprobe` parameter before executing search
   - Ensure `nprobe >= 1`
   - Return clear error message if validation fails

2. **Improve Error Messages**:
   - Identify specific parameter name ("nprobe")
   - Specify expected valid range (e.g., "must be between 1 and nlist")
   - Provide actual invalid value that was provided

yihui504:
Fixes #47729

Rewritten per reviewer feedback — the prior proxy-layer approach duplicated the `nprobe` range and changed existing error contracts. This version routes search params through knowhere's own `Config::Load`, so the range/type contracts stay owned by knowhere (`src/index/ivf/ivf_config.h`) and are never copied into milvus code.

## Approach

- **Where**: `internal/core/src/segcore/plan_c.cpp`, right after `PopulateBruteForceIndexParams`. Plan creation runs before any segment dispatch, so this covers indexed / brute-force / growing / binlog-index paths uniformly.
- **How**: `ValidateVectorSearchParams` seeds `metric_type`, builds the knowhere config via `IndexStaticFaced<T>::CreateConfig`, then runs `Config::FormatAndCheck` + `Config::Load(SEARCH)`. Knowhere produces any type/range error. Dispatches by data type (`fp32`/`fp16`/`bf16`/`bin1`/`int8`/`sparse_u32_f32`), lifted from `vector_index_c.cpp:63-92`.
- **Error code**: `ErrorCode::InvalidParameter` (2042, `inputError`).

## Tests

`SearchParamsValidator_test.cpp`:
- empty params load across IVF_FLAT / IVF_PQ / IVF_SQ8 / SCANN / HNSW / DISKANN
- `nprobe` ∈ {0, -1, 65537, `"0"`, `null`, `true`, `32.0`} on IVF_FLAT → rejected with knowhere's own message
- same values on HNSW → accepted (no collateral damage)
- valid IVF nprobe passes

Python E2E regression (`nprobe=0` real search on growing / sealed-unindexed / sealed-with-binlog-index) to follow.


internal/proxy/search_util.go
```diff
@@ -553,6 +553,15 @@ func parseSearchInfo(searchParamsPair []*commonpb.KeyValuePair, schema *schemapb
 		searchParamStr = ""
 	}
 
+	// 4.1 validate nprobe: must be a positive integer if present.
+	// nprobe is an IVF-family search parameter; 0 (or negative) is always invalid and
+	// would otherwise silently yield empty/undefined results without any error.
+	if nprobeRes := gjson.Get(searchParamStr, nprobeKey); nprobeRes.Exists() {
+		if nprobeRes.Int() <= 0 {
```
xiaofan-luan: `gjson.Result.Int()` is coercive rather than a type check: invalid strings, `null`, `false`, objects/arrays become `0`, and fractional numbers are truncated. As a result, values such as `"invalid"`, `null`, and `0.5` are reported as `nprobe <= 0` with code 1100 instead of retaining their existing type/index-specific errors; this is what causes the current E2E regressions. Also, `parseSearchInfo` does not know whether the selected index is IVF, so this check affects unrelated index families too.

internal/proxy/task_search_test.go
```diff
@@ -3509,6 +3509,34 @@ func TestSearchTask_parseSearchInfo(t *testing.T) {
...
+			{"missing nprobe", `{"metric_type": "L2"}`, false},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				params := getValidSearchParams()
+				resetSearchParamsValue(params, ParamsKey, tt.paramsValue)
+				_, err := parseSearchInfo(params, nil, nil, false)
+				if tt.expectError {
+					assert.Error(t, err)
```
xiaofan-luan: These assertions only verify that some error contains `nprobe`, so they do not protect the error code/message contract and cannot catch the E2E regressions. Please cover invalid string, `null`, fractional values, and a non-IVF index, and assert the expected code/reason. Since the original bug is path-dependent, add an actual search regression for `nprobe=0` on both indexed and brute-force/growing paths.

xiaofan-luan (commented):
Thanks for rebasing this. I don't think the validation should live in proxy as written. Indexed IVF search already validates `nprobe` through Knowhere's `IvfConfig` range `[1, 65536]`; the uncovered behavior is path-dependent because growing/raw sealed brute-force search bypasses Knowhere `Config::Load`. Adding a coercive check in `parseSearchInfo` runs before the index type is known and changes existing error contracts, as shown by the current E2E failures.

Please move or factor the validation into a common segcore path before the indexed-vs-brute-force dispatch (or another layer that knows the index family), while preserving the existing Knowhere type/range errors. The regression coverage should include `nprobe=0` on both indexed IVF and brute-force/growing paths.

xiaofan-luan:
please fix issues as comment, happy to review again once done.

yihui504:
> please fix issues as comment, happy to review again once done.

@xiaofan-luan a new fix is now ready for review,if there's something more for me to do,please just let me know!

xiaofan-luan:
Thanks for sticking with this through the rewrite, and sorry for the extra round-trip — this asks you to move the check one more time, so let me state the criterion up front so it's the last move.

## The criterion

`nprobe`'s valid range is **owned by knowhere**. It is declared exactly once, in `src/index/ivf/ivf_config.h:38-44`:

```cpp
KNOWHERE_CONFIG_DECLARE_FIELD(nprobe)
    .set_default(8)
    .for_search().for_range_search().for_iterator()
    .set_range(1, 65536);
```

#47729 is not "Milvus is missing a check". It is "**one execution path bypasses the owner**". So the fix has to route that path back through the owner, at a layer that knows which index family the request targets. Any patch that re-states the range in Milvus code — Go or C++ — is a second copy of the contract, and it will drift.

That is what my earlier review meant by *"preserving the existing Knowhere type/range errors"* and *"another layer that knows the index family."* I wasn't concrete enough about it, which is part of why this landed where it did.

## Where the current version lands

The message text and `[1, 65536]` bound are exact — I checked them against the pinned `v3.0.6` (`config.h:367-386`, `Range::to_string`). But the range was *copied*, not *reused*, and that's why two holes remain.

**1. String form bypasses the check, so #47729 still reproduces.**

Commit `be26264` made strings fall through. The comment justifying it doesn't hold: #41767 is closed, and knowhere does not ignore string `nprobe`. The indexed path runs `LoadConfig` (`src/index/index.cc:31-39`), which calls `FormatAndCheck` **before** `Config::Load`; `src/common/config.cc:87-112` takes any `Entry<CFG_INT>` whose value `is_string()`, runs `std::stoll`, and writes the integer back — then the normal range check applies.

| `search(params={"nprobe": "0"})` | Result |
|---|---|
| Sealed + IVF index | `"0"` → `0` → rejected, `Out of range in json: param 'nprobe' (0) should be in range [1, 65536]` |
| Growing / unindexed sealed | accepted, returns results |

Same path-dependence as before, different spelling. String params aren't exotic — the repro in #41767 is literally `"nprobe": "32"`, and Milvus itself emits string form at `IndexConfigGenerator.cpp:79-80` (`std::to_string`).

**2. The sealed binlog-index branch is never reached by the check.**

`ChunkedSegmentSealedImpl::vector_search` (`ChunkedSegmentSealedImpl.cpp:3620-3695`) has three branches:

| Branch | Condition | Path | Validated? |
|---|---|---|---|
| 1 | `binlog_index_bitset` (`:3621`) | `GetSearchConf` (`:3625`) → `SearchOnSealedIndex` | **no** |
| 2 | `index_ready_bitset` (`:3642`) | `SearchOnSealedIndex` | yes, knowhere |
| 3 | else (`:3661`) | `SearchOnSealedColumn` | yes, this PR |

Branch 1 calls `VecIndexConfig::GetSearchConf` (`IndexConfigGenerator.cpp:137-153`), which replaces `search_params_` wholesale and copies back only `maintain_params = {"radius", "range_filter", "drop_ratio_search", "dim_max_score_ratio"}` (`IndexConfigGenerator.h:38-39`). The user's `nprobe` is dropped, and the interim index's own value is used — normalized in `component_param.go:4339-4356`, so it can never be out of range. With `interimIndex.enableIndex=true`, `nprobe=0` on a sealed-not-yet-indexed segment still passes silently.

Note the growing path is fine, because `SearchOnGrowing.cpp:117` runs the check **before** the interim-index branch at `:129`. That asymmetry is the tell: on growing it's a pre-dispatch check, on sealed it's one post-dispatch branch out of three.

**Both of these disappear if the check moves — that's the point, not two separate asks.**

## The fix

Do it once at plan construction, where the index type is known and no segment dispatch has happened yet: `internal/core/src/segcore/plan_c.cpp:63-69`. That block already has `field_index_meta` and a mutable `search_info`.

No knowhere change is needed — `Config::FormatAndCheck` and `Config::Load` are public statics (`config.h:341-346`), and `IndexStaticFaced<T>::CreateConfig` is public.

```cpp
// plan_c.cpp, right after PopulateBruteForceIndexParams(...)
ValidateVectorSearchParams(search_info,
                           field_index_meta.GetIndexType(),
                           (*schema)[milvus::FieldId(field_id)].get_data_type());
```

```cpp
template <typename T>
static void
LoadAndCheck(const std::string& index_type, const knowhere::Json& params) {
    auto cfg = knowhere::IndexStaticFaced<T>::CreateConfig(
        index_type, knowhere::Version::GetCurrentVersion().VersionNumber());
    knowhere::Json json(params);
    std::string msg;
    auto status = knowhere::Config::FormatAndCheck(*cfg, json, &msg);
    if (status == knowhere::Status::success) {
        status = knowhere::Config::Load(
            *cfg, json, knowhere::PARAM_TYPE::SEARCH, &msg);
    }
    if (status != knowhere::Status::success) {
        ThrowInfo(milvus::ErrorCode::InvalidParameter, msg);
    }
}
```

Dispatch by data type to `fp32 / fp16 / bf16 / bin1 / int8 / sparse_u32_f32` — you can lift that block verbatim from `vector_index_c.cpp:63-92`, including its `VECTOR_ARRAY` element-type handling.

**One detail that matters:** seed the json with `metric_type` and `topk` before validating, mirroring `PrepareSearchParams` (`VectorIndex.h:169-173`):

```cpp
json[knowhere::meta::METRIC_TYPE] = search_info.metric_type_;
json[knowhere::meta::TOPK] = search_info.topk_;
```

Then the validated input is byte-identical to what the indexed path validates, so the two cannot disagree — they're running the same code.

What this buys, without copying a single constant:

- `nprobe=0` and `nprobe="0"` rejected on **every** segment path, with knowhere's own message.
- All three sealed branches covered, because it runs before the dispatch.
- **Non-IVF is untouched**: `Config::Load` iterates only `cfg.__DICT__` (`config.h:346-347`), and `HnswConfig`/`DiskANNConfig`/`SparseInvertedIndexConfig` don't declare `nprobe`, so a stray `nprobe` on HNSW stays ignored exactly as today. This is the concern from my earlier review, and it resolves itself here because the arbiter is knowhere's field declaration rather than a hand-written list in Milvus.
- Same-class bugs (`ef=0`, out-of-range `search_list_size`, …) get fixed for free.

## Error code

Please use `ErrorCode::InvalidParameter` (2042), not `UnexpectedError`. 2042 is already flagged `inputError: true` in `pkg/util/merr/segcore.go:108`, so a bad user parameter surfaces as a parameter error rather than a non-retriable internal server error. `UnexpectedError` (2001) maps to the generic `ErrSegcore` catch-all with no input-error flag.

## One risk to close, and the tests

Running `Config::Load(SEARCH)` against raw user params fails if some index family declares a `for_search()` field with no default and without `allow_empty_without_default`. I checked `metric_type` (default `"L2"`), `k` (default `10`) and `nprobe` (default `8`), so the common path is safe — but I only checked `BaseConfig` and `IvfConfig`.

So please add, alongside the functional tests:

1. **A sweep over every supported index type**, asserting a minimal/empty search-param set still loads. This is what makes the change safe, and it's cheap.
2. `nprobe` ∈ {`0`, `-1`, `65537`, `"0"`, `null`, `true`, `32.0`} on IVF_FLAT → all rejected, message identical to the indexed path.
3. The same values on HNSW → all accepted (proving no collateral damage).
4. A real search regression for `nprobe=0` covering growing, sealed-unindexed, **and** sealed-with-binlog-index (`interimIndex.enableIndex=true`) — branch 1 above is the one the current patch misses, so please make sure a test would have caught it.

## Also

The PR description still describes the v1 proxy approach and claims *"Added 5 cases to `TestSearchTask_parseSearchInfo`"*, which isn't in this diff. Please update it before merge — otherwise the squash commit message won't match the change.

yihui504:
@xiaofan-luan one deliberate deviation from your sketch, flagging it so it's not a surprise:

You suggested seeding both `metric_type` and `topk`. I seed only `metric_type`. Seeding the raw `topk` made `Config::Load` reject legitimate HNSW searches — `test_bitmap_index_search_group_by` does `ef=32, k=1000`, and milvus adjusts `ef = max(ef, k)` only at search time (not plan creation), so seeding `k=1000` here trips knowhere's `ef should be larger than k`. nprobe and other range/type checks don't depend on `topk`, so dropping the seed preserves the validation goal.

UT-CPP (8445 passed, 0 failed) and e2e-default are green on `d5459dc`. Python E2E regression (nprobe=0 on growing / sealed-unindexed / sealed-binlog-index) is next.

yihui504:
@xiaofan-luan the plan-level routing is implemented per your sketch (latest 8a30e1f, CI green: UT-CPP 8445 passed + e2e-default). Could you check whether the tests match your expectation?

C++ unit tests (SearchParamsValidator_test.cpp) — your items 1–3:

empty params load across IVF_FLAT / IVF_PQ / IVF_SQ8 / SCANN / HNSW / DISKANN
nprobe ∈ {0, -1, 65537, "0", null, true, 32.0} on IVF_FLAT → rejected, message identical to the indexed path
same values on HNSW → accepted (nprobe not declared, ignored as today)
Python E2E (test_nprobe_validation.py) — your item 4, actual search regression:

sealed segment with IVF index loaded: nprobe=0 → rejected
growing segment (insert without flush): nprobe=0 → rejected
On the binlog-index path (your item 4, branch 1): I could not exercise it as a standalone search test. milvus's e2e framework has no API to toggle queryNode.segcore.interimIndex.enableIndex at runtime (it's env/startup only), and interimIndex has never had an e2e test in milvus. Since ValidateVectorSearchParams runs at plan creation — before the ChunkedSegmentSealedImpl branch 1 dispatch — the binlog-index path is covered by construction. If you want it exercised as an actual search, I'd appreciate guidance on how you'd expect interimIndex to be enabled in e2e.

(One deviation from your sketch, flagged earlier: I seed only metric_type, not topk — seeding the raw topk tripped knowhere's ef > k check on legitimate HNSW searches like ef=32, k=1000. nprobe/range checks don't depend on topk.)

yihui504:
@xiaofan-luan following up after the latest label-manager pass — no changes to the fix itself, but two review items are now closed and one semantic decision is flagged for your call:

**1. Sweep extended to every CPU family (your item 1).** `EmptyParamsLoadAllIndexTypes` now covers IVF_RABITQ / IVF_RABITQ_FASTSCAN / BIN_FLAT / BIN_IVF_FLAT / SPARSE_INVERTED_INDEX / SPARSE_WAND on top of the original 6 dense-float families (binaries use HAMMING, sparse uses IP). GPU families are deliberately excluded: `CreateConfig` throws for types not registered in this CPU build, and the catch in `LoadAndCheck` would make the sweep silently pass without exercising knowhere — a note in the test explains this. (Verified against the new knowhere pin: every `for_search()` field in the sparse config has a default, so an empty param set loads.)

**2. HNSW invalid set completed (your item 3).** `HnswIgnoresStrayNprobe` now ignores the same 7 values {0, -1, 65537, "0", null, true, 32.0} that `IvfRejectsInvalidNprobe` rejects, value-for-value.

**3. Unindexed-sealed is intentionally not validated — flagging for your sign-off.** When `index_type` is empty (collection loaded without any field index → brute-force path), `ValidateVectorSearchParams` skips: there is no knowhere owner to route to, since nprobe's range lives in the IVF index config that doesn't exist here. This is the same "no declared owner → param ignored" semantics as HNSW ignoring a stray nprobe. If you'd rather vector search params be rejected outright on unindexed collections, that would be a BaseConfig-level default in knowhere rather than a milvus decision — happy to move it there if you prefer.

Also noting: the knowhere pin moved v3.0.6 → `6a3b7df1`; I checked `ivf_config.h` — nprobe's `[1, 65536]` declaration is unchanged, so the asserted messages still match. Branch has been updated with master.

yihui504:
@xiaofan-luan gentle nudge — the branch is now fully green and waiting on review:

| check | result |
|---|---|
| UT-CPP-Cov | 8653 passed, 0 failed |
| UT-GO / UT-Integration / Code-Check | ✅ |
| go-sdk (rerun) / e2e-default | ✅ |
| DCO | ✅ |

One blocking item needs a maintainer click: the `Python Lint (tests/)` workflow on this fork PR is pending approval — [run 31915636379](https://github.com/milvus-io/milvus/actions/runs/31915636379) (`action_required`). The earlier failure on `df85c69` was a `ruff format` issue in `test_nprobe_validation.py`, fixed in `f58f475` (verified locally with ruff 0.15.11 against `tests/ruff.toml`), so the run should pass once approved.

The review questions from my last comment (unindexed-sealed skip semantics, GPU sweep exclusion) still stand whenever you have a moment — happy to adjust either way.

PR #51127, opened 2026-07-07
enhance: batch woodpecker client appends & expose sync knobs for small-batch high-concurrency QPS

tinswzy:
## What & why

Optimizes QPS in **small-batch high-concurrency** write scenarios (#50829).

The performance win lives in woodpecker `v0.1.34`: a client-side batched send/ack
pipeline (zilliztech/woodpecker#208) that coalesces consecutive small appends into
a single `AddEntries` request. This PR bumps the dependency and adds the tuning
knobs on the milvus side.

## Changes

- Bump `github.com/zilliztech/woodpecker` `v0.1.33` → `v0.1.35` (root + `pkg/`).
  `v0.1.34` brings the batching pipeline; [`v0.1.35`](https://github.com/zilliztech/woodpecker/releases/tag/v0.1.35) is a
  small reliability/bugfix release on top (batch-ack/send timeout bounds, retry
  data-race fixes, disk-watermark backpressure).
- Add three woodpecker parameters (`service_param.go`) and regenerate
  `configs/milvus.yaml`; assert defaults in `service_param_test.go`:
  - `woodpecker.client.segmentAppend.maxBatchEntries` — default **1000**
  - `woodpecker.client.segmentAppend.maxBatchBytes` — default **2000000** (~2MB)
  - `woodpecker.logstore.segmentSyncPolicy.maxIntervalForService` — default **10ms**

## Where each knob takes effect (important)

- **`maxBatchEntries` / `maxBatchBytes`** are **client-side** (consumed in
  `woodpecker/segment/segment_handle.go`), wired through `setCustomWpConfig`, and
  run inside the milvus process in **both** embed and service modes. This is the
  effective QPS lever.
  **Blast radius:** the bump *by itself* (batching landed in `v0.1.34`) puts **every** deployment —
  embed/standalone included — on the new `BatchAppendOp` write path by default
  (`getDefaultWoodpeckerConfig()` ships `MaxBatchEntries: 1000`; `NewSegmentHandle`
  is shared by both modes). The `maxBatchEntries` knob added here is the
  **off-switch** for that change (`1` disables batching), not its cause.
- **`maxIntervalForService`** is a **woodpecker service-storage-mode server** knob
  (only the staged-storage writer reads it, and that writer is instantiated only
  when `woodpecker.storage.type == "service"`). It is **not consumed inside the
  milvus process** — embed mode never uses staged storage, and service mode runs a
  pure client. It is therefore exposed only as a config surface for standalone
[...]

PR #51810, opened 2026-07-24
enhance: [3.0] batch woodpecker client appends & expose sync knobs for small-batch high-concurrency QPS

tinswzy:
pr: #51127


czs007:

## Overview

This PR's config plumbing is clean, but the default-on batching flip plus the woodpecker `v0.1.35` bump introduce two behavior changes that need an explicit answer, along with three smaller config/UX inconsistencies.

## Medium

**`configs/milvus.yaml:197` — batching on by default invokes the new `AddEntries` RPC, which a stale (v0.1.33) log-store server does not implement, and the resulting error is not classified retryable.**
`AddEntries` does not exist anywhere in v0.1.33 (repo-wide grep returns zero files); v0.1.35 adds it, and the batched send has no `codes.Unimplemented` fallback to per-entry `AddEntry`. The raw gRPC error is stored as the per-replica channel error and fed straight into the retry gate, where `werr.IsRetryableErr` returns `false` for any non-`woodpeckerError`, so the op skips retry entirely and goes to final failure. **Failure:** in a `woodpecker.storage.type=service` deployment where Milvus is upgraded to this build while some log-store pods still run v0.1.33 → every batched append to a stale replica returns `Unimplemented`, `HandleAppendRequestFailure` skips retry and marks final failure, so one stale replica forces continuous segment rolling and `(Wq-Aq+1)` stale replicas fast-fail WAL writes for the affected logs.
Scope note: this is reachable only via `logStoreClientRemote`. The default `woodpecker.storage.type` is `minio` (`pkg/util/paramtable/service_param.go:1082`), and both `minio` and `local` go through `NewEmbedClient` → `NewLogStoreClientPoolLocal`, whose `AppendEntries` just loops the in-process `AddEntry` — no new RPC, no quorum fan-out. So embedded deployments are unaffected.
*Suggestion:* answer this explicitly in the PR body (or add a mixed-version test) for service-mode rolling upgrades. The only Milvus-side lever is defaulting `maxBatchEntries` to `1`, since woodpecker's own built-in default is already `MaxBatchEntries: 1000` (`common/config/configuration.go:814`) — the bump turns batching on regardless of the paramtable default. Fallback-on-`Unimplemented` or capability negotiation belongs in the library. *(raised by xiaocai2333, chyezh)*

**`pkg/streaming/walimpls/impls/wp/builder.go:78` — woodpecker v0.1.35 disk-watermark write rejection is enabled by default with no Milvus-side override.**
v0.1.33 has no `DiskWatermarkPolicy` at all; v0.1.35 adds it with `Enabled: true`, `MinFreeBytes: 1Gi`, `HardThresholdRatio: 0.90`. `NewLogStore` registers the task whenever the policy is enabled, `RootPath` is non-empty (Milvus always sets it, `builder.go:156-162`) and the storage type is `service` or `local` — so an embed deployment with `woodpecker.storage.type=local` does register it. The maintenance manager runs each task once immediately at startup, and `diskRejectBpsFor` returns full rejection whenever `free <= MinFreeBytes` regardless of `ThrottleEnabled`. **Failure:** `storage.type=local` with the WAL volume at or below 1 GiB free (small PVC, dev box, or a disk that fills) → the first sample at startup sets `diskRejectBps` to max, every `AddEntry`/`AddEntryBatch` returns `ErrLogStoreDiskPressure`; the error is retryable, so appends burn `MaxRetries`, final-fail and roll segments, blocking WAL writes for the whole node.
There is no escape hatch: `setCustomWpConfig` never touches `DiskWatermarkPolicy`, and `getWpConfig` calls `config.NewConfiguration()` with no file arguments, so there is not even a YAML surface for it.
*Suggestion:* expose and wire the watermark knobs through paramtable, or preserve the previously-disabled behavior for embedded local storage. *(raised by chyezh)*

## Low

**`pkg/util/paramtable/service_param.go:951` — `maxIntervalForService` is exported into `milvus.yaml` but read by nothing in this repo, and its doc contradicts the effective default.**
The param is `Export: true` with `DefaultValue: "10ms"`, and `builder.go` deliberately does not wire it, on the stated assumption that open-source installs hand the same `milvus.yaml` to the woodpecker server. Nothing in this repo does that: the only consumer is the staged-storage writer inside a standalone log-store server, that server reads its own config file (`cmd/external/user_config.go:317-318`), and no artifact under `configs/`, `deployments/` or `scripts/` forwards `milvus.yaml` to it. **Failure:** an operator tunes the key on a service-mode deployment to trade durability latency for QPS → no Milvus code path reads it (only the definition and a unit test reference it), the knob has zero effect and no warning is emitted, while the exported doc asserts "default is 10 milliseconds" when the effective server default is still woodpecker's 1ms (`common/config/configuration.go:853`).
*Suggestion:* either drop the key from `milvus.yaml` (keep the tuning note in the woodpecker repo or the PR body), or state in the doc that the value is inert unless `milvus.yaml` is also fed to the log-store server. *(surfaced during verification)*

## Nitpick

**`pkg/streaming/walimpls/impls/wp/builder.go:104` — `maxBatchEntries: 0` is treated as invalid and yields full 1000-entry batching, the opposite of the sibling branch's literal-`0` handling.**
The two guards added next to each other treat an explicit `0` in opposite ways: for `maxBatchBytes` a literal `"0"` is honored as "no limit" (`builder.go:112-113`), while for `maxBatchEntries` any non-positive value is declared invalid and woodpecker's built-in `1000` is kept. woodpecker itself would have clamped `<1` to `1`, i.e. batching disabled, so the wrapper actively inverts the outcome. **Failure:** operator sets `maxBatchEntries: 0` (or a negative value) intending to disable client-side batching → the value is discarded, batching runs at full width at 1000, and the warn log says only "invalid … keeping woodpecker built-in default" — which also means the documented mitigation for the `AddEntries` mixed-version risk above silently does not apply.
*Suggestion:* honor a literal `"0"` as the disable spelling, symmetric with the bytes branch, or make the warn message state explicitly that batching remains enabled at N and that `1` is the disable value. *(surfaced during verification)*

**`pkg/streaming/walimpls/impls/wp/builder.go:122` — only the exact literal `"0"` disables the byte cap; equivalent zero spellings silently keep the default.**
`GetAsSize()` returns `0` both for a real parse failure and for zero-valued sized spellings: `"0mb"` hits the m/mb branch and returns `0`, `"0b"` has no recognized suffix so `ParseInt("0b")` fails and returns `0`. Neither equals the trimmed literal `"0"`. **Failure:** `maxBatchBytes: 0b` (or `0mb`/`0kb`), intending "no byte limit" → both land in the else branch, log "invalid woodpecker maxBatchBytes" and leave the 2MB cap in effect — the opposite of the operator's intent.
*Suggestion:* the behavior is defensible; the doc at `service_param.go:818` should say that only the bare literal `0` removes the limit. *(raised by xiaocai2333)*

czs007:

## Re-review of `5b33042cf9fe`

- **`configs/milvus.yaml:197` — default-on batching invokes `AddEntries`, unimplemented on a stale v0.1.33 log-store server, and not classified retryable (service mode only): Confirmed.** Both reviewers re-walked the full chain on this branch's dependency set (`sequential_executor.go:106-140` → `batch_append_op.go:129` → `logstore_client_remote.go:206/219` raw gRPC error → `segment_handle.go:417-419` `werr.IsRetryableErr` → false); it still stands, rescoped to `woodpecker.storage.type=service` deployments — see the medium section below.
- **`pkg/streaming/walimpls/impls/wp/builder.go:78` — v0.1.35 disk-watermark write rejection is enabled by default with no Milvus-side override: Confirmed.** `configuration.go:899-906` still ships the policy `Enabled: true` and `setCustomWpConfig` wires no `DiskWatermarkPolicy` field and exposes no paramtable key — see the medium section below.
- **`pkg/util/paramtable/service_param.go:951` — `maxIntervalForService` is exported but read by nothing in this repo, and its documented "default is 10ms" contradicts the effective 1ms default: Confirmed in part, retracted in part.** The dead-config half stands (the only non-test consumers are `writer_impl.go:142` and `cmd/external/user_config.go:317`, both outside this process). The "documented default contradicts the effective 1ms default" half was wrong on our side: `configs/milvus.yaml:245` spells out in the same comment that this raises the interval from woodpecker's built-in 1ms default.
- **`pkg/streaming/walimpls/impls/wp/builder.go:122` — only the exact literal `"0"` disables the byte cap: Confirmed.** `param_item.go:354-359` maps both `"0mb"` and `"0b"` to `0`, neither of which equals the trimmed literal, so both take the warn branch.
- **`pkg/streaming/walimpls/impls/wp/builder.go:104` — `maxBatchEntries: 0` is treated as invalid and yields full 1000-entry batching, the opposite of the sibling branch's literal-0 handling: Confirmed.** The asymmetry between `builder.go:104-109` and `:110-117` is unchanged.

---

This round's review of the backport found no newly introduced must-fix issues; the items below are the two default-on behaviors that come with the v0.1.33→v0.1.35 dependency bump, plus minor configuration and test-coverage notes.

## Medium

- **`configs/milvus.yaml:197` — default-on batching calls an RPC a stale log-store server does not implement, and the error is not retryable.** `maxBatchEntries` defaults to 1000, so any time ≥2 `AppendOp`s are queued the batch path calls `AddEntries` (`sequential_executor.go:115` → `batch_append_op.go:134` → `logstore_client_remote.go:199`), an RPC that does not exist in v0.1.33 at all. Failure: in a `woodpecker.storage.type=service` deployment where Milvus is upgraded first and the standalone log-store server is still v0.1.33, concurrent writes queue ≥2 ops → `AddEntries` returns `codes.Unimplemented`, which is a raw gRPC status rather than a `woodpeckerError`, so `werr.IsRetryableErr` returns false (`errors.go:313-319`) and the batch skips `MaxRetries` straight to final failure and the quorum-loss path, failing that pchannel's WAL writes for the whole mixed-version window — and because a batch of size 1 falls back to the unary `AddEntry`, this looks healthy at low QPS and breaks under load. Suggestion: document the upgrade ordering (log-store server before Milvus) in the PR body and release notes, and call out `maxBatchEntries: 1` as the escape hatch; a `codes.Unimplemented` fallback to unary `AddEntry` in woodpecker would remove the window entirely. Default `minio`/embed deployments are unaffected. (Also tracked above under the re-review of `5b33042cf9fe`.) (raised by czs007, xiaocai2333)
- **`pkg/streaming/walimpls/impls/wp/builder.go:78` — v0.1.35's disk-watermark write rejection is on by default with no Milvus-side override.** The bump brings in `DiskWatermarkPolicy`, published with `Enabled: true, HardThresholdRatio: 0.90, MinFreeBytes: 1Gi, ThrottleEnabled: true` (`configuration.go:899-906`), and `server/logstore.go:138-142` registers `newDiskWatermarkTask` whenever the policy is enabled, `Storage.RootPath != ""` (Milvus always sets it, `builder.go:155-161`) and storage is service or local. Failure: on a `woodpecker.storage.type=local` deployment whose WAL disk crosses the 0.90 hard watermark or drops below 1GiB free → `admitAppend` rejects writes with `ErrLogStoreDiskPressure`, which is retryable but exhausts the 3 client retries and then fails the append, with no key in `milvus.yaml` to relax the threshold or disable the backpressure. Suggestion: wire at least `Enabled` and the threshold ratios through `setCustomWpConfig` with matching `ParamItem`s. Non-blocking for the default `minio` deployment, where the task is never registered. (raised by czs007, xiaocai2333)

## Low

- **`pkg/util/paramtable/service_param.go:951` — `maxIntervalForService` has no consumer inside this repo.** The new exported param is never assigned in `setCustomWpConfig`; `builder.go:127-136` carries an explicit NOTE that it is intentionally not wired. Its only non-test consumers in v0.1.35 are `server/storage/stagedstorage/writer_impl.go:142` and `cmd/external/user_config.go:317`, so it only takes effect when the deployment layer renders `milvus.yaml` to a standalone woodpecker server — a path outside this repo. Suggestion: keep it, but make the "only effective when this file is mounted into the woodpecker log-store server" caveat as prominent in the yaml comment as it is in the Go NOTE, so operators don't expect it to change in-process behavior. (raised by czs007, xiaocai2333)
- **`configs/milvus.yaml:245` — existing service-mode deployments move from a 1ms to a 10ms sync interval.** `cmd/external/user_config.go:317-318` overrides woodpecker's built-in 1ms default (`configuration.go:853`) whenever the mounted value is >0. Failure: an existing `woodpecker.storage.type=service` deployment that hands the same `milvus.yaml` to its standalone log-store server upgrades to this file → the staged writer's sync ticker goes from 1ms to 10ms, adding up to ~9ms of persistence latency per flush, with no explicit operator action required to trigger it. The yaml comment does state this relationship, so it is no longer silent. Suggestion: surface it in the release notes as an operator-visible upgrade note. (raised by czs007, xiaocai2333)
- **`pkg/streaming/walimpls/impls/wp/builder_test.go:61` — no test anchors woodpecker's built-in `maxBatchBytes` default.** `ZeroEntriesKeepsWoodpeckerDefault` (`:61-65`) genuinely anchors the entries default — it builds a fresh `config.NewConfiguration()`, takes the else branch, and asserts `1000`. There is no bytes-side equivalent: `MilvusDefaults` (`:30-36`) asserts `2000000` only after the `v > 0` branch has written Milvus's own `DefaultValue`, and `InvalidValuesKeepWoodpeckerDefaults` (`:51-59`) uses 777/888 sentinels that exercise the branch, not the default. The coincidence that woodpecker's built-in is also `2000000` (`configuration.go:815`) hides the gap. Suggestion: add a one-line assert on a fresh `NewConfiguration()` so a change to woodpecker's built-in `MaxBatchBytes` fails the test. (raised by czs007, xiaocai2333)

## Nitpick

- **`pkg/streaming/walimpls/impls/wp/builder.go:112` — only the literal `"0"` lifts the byte cap; `"0mb"` / `"0b"` fall into the warn branch.** `GetAsSize` maps `"0mb"` to `0*1024*1024` and `"0b"` to a failed parse returning `0` (`param_item.go:354-372`), neither of which equals the trimmed literal `"0"`, so both keep woodpecker's 2MB default while logging "invalid ... keeping woodpecker built-in default". The behavior is safe; only the log message misleads. Suggestion: treat any parsed size of exactly 0 as "no limit", or reword the warning to say which spellings are accepted. (raised by czs007, xiaocai2333)
- **`pkg/streaming/walimpls/impls/wp/builder.go:104` — `maxBatchEntries: 0` keeps full 1000-entry batching, the opposite of the sibling byte branch's literal-0 semantics.** `builder.go:104-109` sends 0 and -1 into the warn branch and keeps woodpecker's built-in 1000, while `:110-117` treats a literal `"0"` for bytes as "no limit". The asymmetry is documented in the comment at `:97-103` and `configs/milvus.yaml:197` tells operators to use `1` rather than `0` to disable batching. Suggestion: none required; consider mirroring the comment into the yaml for `maxBatchEntries` so the `0` vs `1` distinction is visible where operators set it. (raised by czs007, xiaocai2333)

## Carried over from #51127

- **`go.mod:267` — woodpecker v0.1.33→v0.1.34 bump introduces the new `AddEntries` streaming RPC; concurrent writes fail unretryably during a mixed-version rolling upgrade: Confirmed on this branch.** Same mechanism as the item above, re-verified at `sequential_executor.go:115` / `batch_append_op.go:134` / `logstore_client_remote.go:199` / `segment_handle.go:418-419`; note the range moved — this PR bumps straight to v0.1.35, so the affected window is any server older than v0.1.34.
- **`pkg/streaming/walimpls/impls/wp/builder.go:108` — `MaxIntervalForService`'s Go wiring is dead code inside the Milvus process and the PR description over-claims: Resolved upstream.** The wiring is gone and `builder.go:128-137` is now an explicit NOTE that the field is intentionally not set; the commit message states it is not consumed inside the Milvus process, and this backport's body only references `pr: #51127` rather than repeating the earlier claim. The dead-config half is reported separately above as a documentation-clarity note.
- **`pkg/streaming/walimpls/impls/wp/builder.go:96` — a typo in `maxBatchBytes`/`maxBatchEntries` silently becomes 0, removing the byte cap or disabling batching, with no validation on the Milvus path: Resolved upstream.** `builder.go:104-117` now guards both sides — a non-positive `maxBatchEntries` warns and keeps the built-in, and an unparseable `maxBatchBytes` (e.g. `"1,000"`) is distinguished from an explicit `"0"` — and `builder_test.go:51-59` covers the typo path.
- **`pkg/streaming/walimpls/impls/wp/builder.go:96` — batching params are tagged `refreshable:"true"` but wired as a one-shot copy in `Build()`, so runtime refresh does not take effect: Retracted.** Our earlier comment was wrong here: on this branch `service_param.go:725-726` declares both `AppendMaxBatchEntries` and `AppendMaxBatchBytes` as `refreshable:"false"`, which matches the one-shot copy exactly. `SyncMaxIntervalForService` is still `refreshable:"true"` at `:747`, but it has no in-process consumer at all, so nothing can go stale.
- **`configs/milvus.yaml:245` — after upgrading, existing service-mode deployments silently move from a 1ms to a 10ms woodpecker sync interval: Resolved upstream for the "silently" part.** The yaml comment at `:245` now states explicitly that this raises the interval from woodpecker's built-in 1ms default and changes the cadence for existing service-mode deployments. The behavior change itself still applies and is listed above as an operator-visible upgrade note.
- **`pkg/util/paramtable/service_param.go:722` — the PR body's claim that the three params were "previously declared in the struct but never Init'd" contradicts the diff: Resolved upstream.** That sentence appears in neither the merged commit message nor this PR's body; the body now describes the params as added by this change.
- **`pkg/streaming/walimpls/impls/wp/builder_test.go:51` — no test independently anchors woodpecker's built-in defaults (1000/2000000), leaving the invalid-value fallback contract without an anchor: Confirmed on this branch, narrowed to `maxBatchBytes`.** `builder_test.go:61-65` now anchors the entries default on a fresh `config.NewConfiguration()`; the bytes side has no equivalent, which is the low-severity test-coverage item listed above.

- **`configs/milvus.yaml:192` — the legacy-metadata-prefix compatibility guarantee this comment now documents has a silent failure path.** `detectEffectiveMetadataPrefix` (`meta/metadata_etcd.go:135-146`) runs once per client construction and does not fail closed: on an etcd error it logs a warning and returns the configured prefix instead of the legacy `"woodpecker"` root, so on a cluster whose metadata lives under the legacy root, one timed-out `hasLegacyMetadata` Txn during startup points the provider at an empty namespace, and `opener.go:22-32` then reacts to `LogExists == false` by calling `CreateLog` and creating a fresh empty log for that pchannel. Unrelated to this PR: the detection code is byte-for-byte identical in v0.1.33 and v0.1.35 (the file's only diff is four logger-level/field changes), and this PR's only `MetaPrefix` change is swapping `paramtable.Get().WoodpeckerCfg.MetaPrefix` for the same-valued `cfg.MetaPrefix` at `builder.go:91` — the fallback exists with or without this change. No tracking issue yet; the fix belongs upstream (fail closed on probe error), or the yaml comment should state that the fallback is best-effort. (raised by czs007)

czs007 (approved):
/lgtm

PR #51812, opened 2026-07-24
fix: [cp3.0]gate collection-rooted index layout behind an explicit opt-in

bigsheeper:

pr: #51802
issue: #51788

## Summary

Cherry-picked from master PR #51802 (open - preemptive backport).

Gates the collection-rooted index layout (`index_v1`) behind `dataCoord.index.storePathVersion`, defaulting to `0` (legacy build-rooted) on 3.0.x. This stops a `2.6 -> 3.0` upgrade from silently writing index files a 2.6 binary cannot address, which otherwise leaves collections query-unserviceable after rolling back to 2.6.

Only `0` and `1` are accepted; any other value — including a malformed string, which the paramtable getters silently coerce to `0` — falls back to layout 0 and logs a rated warning. The QueryNode half of the gate compares the version each QueryNode publishes in its session, which is the compile-time `common.Version` constant, so it separates release lines (2.6.x vs 3.0.x) but does not differentiate binaries within the same release line; the config doc and code comment say so explicitly.

A brand-new 3.0 cluster previously enabled the collection-rooted layout automatically, so with this change a fresh 3.0 deployment builds indexes under the legacy `index_files/` layout until the config is explicitly set to `1`.

**Note**: Original PR is still open (under review). This CP PR should be updated if the original PR changes.

## Verification

- [x] Re-synced with master PR #51802 after its review follow-up commit (unsupported-value rejection + gate doc rewording).
- [x] File count matches original PR (5 files)
- [x] Changed lines identical to master PR
- [x] No conflict markers
- [x] `go test -tags dynamic,test -gcflags="all=-N -l" ./internal/datacoord/ -run 'Test_IndexEngineVersionManager|TestServer_initServiceDiscovery_BindIndexNodeDoesNotAffectQueryNodePathVersion'` green on this branch.

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -797,6 +797,10 @@ dataCoord:
   syncSegmentsInterval: 300 # The time interval for regularly syncing segments
   index:
     memSizeEstimateMultiplier: 2 # When the memory size is not setup by index procedure, multiplier to estimate the memory size of index data
+    # Object storage layout for newly built segment index files. 0: index_files/{buildID}/{indexVersion}/{partitionID}/{segmentID}, 1: index_v1/{collectionID}/{partitionID}/{segmentID}/{buildID}/{indexVersion}.
+    # Layout 1 cannot be read by 2.6, so enabling it commits the upgrade and gives up the ability to roll back to 2.6. Existing index files keep the layout they were built with, so this switch only affects index files built from now on.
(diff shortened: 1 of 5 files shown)


bigsheeper:
rerun go-sdk
rerun ut

czs007 (approved):
/lgtm

Issue #51815, opened 2026-07-24
[Enhancement]: Limit DataCoord stats task submission by global scheduler pending queue


xiaocai2333:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Limit DataCoord stats-task submission using the total number of pending tasks in the global scheduler.

When the global scheduler has more than 100 pending tasks, DataCoord should skip submitting new stats tasks. The threshold should use `dataCoord.sortCompactionTriggerCount` (default: 100), preserving the existing 2.5 sort/stats-task behavior where submission is still allowed at exactly 100 pending tasks and skipped starting at 101.

JSON key-index (JSON shredding) stats tasks should use the same scheduler-level limit instead of the separate 10-tasks-per-10-minutes limiter.

### Why is this needed?

The current per-trigger throttling does not account for backlog shared by all task types in the global scheduler. Under sustained stats-task generation, DataCoord can continue adding work while the scheduler is already overloaded.

Using the scheduler pending queue as the admission signal provides one consistent backpressure mechanism for text, BM25, JSON key-index, and other stats sub-jobs.

### Anything else?

JSON key-index task discovery continues to run on `dataCoord.taskCheckInterval` (60 seconds by default). The admission check should happen before allocating a task ID, persisting task metadata, or enqueueing the task.

PR #51816, opened 2026-07-24
enhance: limit stats task submission by scheduler pending queue

xiaocai2333:
issue: #51815

## What this PR does

- Skips new stats-task submission when the global scheduler already holds more pending **stats** tasks than `dataCoord.statsTaskPendingLimit` (100 by default).
- Counts only stats tasks (`GetPendingTaskCount(taskcommon.Stats)`), so index-build / compaction / import / analyze backlog no longer consumes the stats admission budget. Tasks waiting for a retry-backoff deadline **are counted**: they still sit in the pending queue and will be dispatched as soon as the deadline passes, and counting them keeps the gate closed during a stats failure storm instead of letting it fail open. Tasks already dispatched to workers remain excluded as before.
- Uses queue depth as admission control, not as a submission-rate limit. The scheduler can drain and refill the queue as worker capacity becomes available, which is intentional for improving stats-task throughput.
- Removes the separate JSON key-index limit of 10 tasks per 10 minutes. JSON discovery continues on `dataCoord.taskCheckInterval` and shares the runnable scheduler-backlog limit with text stats tasks.
- Alternates whether text indexing or JSON shredding gets first claim on the shared backlog budget each discovery round, preventing either sub-job from being permanently starved by fixed ordering.
- Filters segments whose stats task is already recorded in meta inside each trigger's selector, and skips collections with no eligible field before scanning segments.
- Checks admission before task-ID allocation, metadata persistence, and enqueue. `SubmitStatsTask` retains the check for direct callers and `AddStatsTask` retains the final duplicate guard.
- Exposes a per-type pending-task count through the global scheduler (`GetPendingTaskCount(taskType)`) and priority queue (`TaskCountBy`), while the scheduler's own queue-depth log continues to use the allocation-free total count.
- Marks `dataCoord.jsonShreddingTriggerCount` and `dataCoord.jsonShreddingTriggerInterval` deprecated. The keys remain accepted, `jsonShreddingTriggerCount: 0` remains a compatibility kill switch, and DataCoord warns when deprecated non-default values are configured.

## Throughput and upgrade behavior

The global gate bounds stats pending-queue depth rather than submission rate. When workers drain pending tasks, later discovery rounds can submit more work up to the configured backlog limit. This allows JSON shredding and text-index stats tasks to use available worker capacity instead of remaining capped at 10 JSON tasks per 10 minutes.

On upgrade, segments with an old or missing JSON stats format can become candidates together. Operators who need to pause that migration can set `common.enabledJSONShredding=false` or keep `dataCoord.jsonShreddingTriggerCount=0`; `dataCoord.statsTaskPendingLimit` controls the shared stats backlog.

## Review fixes

- Fixed the admission accuracy issue: the count is scoped to `taskcommon.Stats` so unrelated index/compaction/import backlog cannot starve stats submission, and backoff tasks are counted (both `CreateTaskOnWorker` and `QueryTaskOnWorker` failures converge on `recordTaskFailure` and re-push into the pending queue), so a worker-side failure storm cannot make the gate fail open.
- Did not restore the old JSON rate limiter because removing that throughput ceiling is the purpose of this PR. The upgrade behavior and operator controls are documented explicitly instead.

## Verification

- Passed: `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord/...`
- Passed from `pkg/`: `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/paramtable -run 'TestComponentParam$'`
- Passed: `go test -count=1 ./cmd/tools/config -run TestYamlFile`
- Passed with golangci-lint v2.11.3 over `./internal/datacoord/...` and `pkg/util/paramtable/...`
- Added focused coverage for stats-only pending counts (`TestGlobalScheduler_GetPendingTaskCountIsScopedByTaskType`: other task types excluded; `TestGlobalScheduler_GetPendingTaskCountIncludesBackoff`: backoff tasks included). Existing tests cover global admission boundaries, alternating sub-job order, duplicate filtering, and the deprecated JSON zero-value kill switch.

### DIFF
--- a/pkg/util/paramtable/component_param.go
+++ b/pkg/util/paramtable/component_param.go
@@ -5361,10 +5361,12 @@ type dataCoordConfig struct {
 	StatsTaskSlotUsage                   ParamItem `refreshable:"true"`
 	AnalyzeTaskSlotUsage                 ParamItem `refreshable:"true"`
 
-	EnableSortCompaction             ParamItem `refreshable:"true"`
-	TaskCheckInterval                ParamItem `refreshable:"true"`
-	SortCompactionTriggerCount       ParamItem `refreshable:"true"`
-	JSONStatsTriggerCount            ParamItem `refreshable:"true"`
+	EnableSortCompaction       ParamItem `refreshable:"true"`
+	TaskCheckInterval          ParamItem `refreshable:"true"`
+	SortCompactionTriggerCount ParamItem `refreshable:"true"`
+	// Deprecated: JSON stats tasks now use SortCompactionTriggerCount for throttling.
+	JSONStatsTriggerCount ParamItem `refreshable:"true"`
+	// Deprecated: JSON stats tasks now run on TaskCheckInterval.
 	JSONStatsTriggerInterval         ParamItem `refreshable:"true"`
 	JSONStatsMaxShreddingColumns     ParamItem `refreshable:"true"`
 	JSONStatsShreddingRatioThreshold ParamItem `refreshable:"true"`
@@ -6857,7 +6859,7 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	p.JSONStatsTriggerCount = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerCount",
 		Version:      "2.6.5",
-		Doc:          "jsonkey stats task count per trigger",
+		Doc:          "Deprecated: JSON stats tasks now use dataCoord.sortCompactionTriggerCount for throttling.",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerCount"},
 		PanicIfEmpty: false,
@@ -6868,7 +6870,7 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	p.JSONStatsTriggerInterval = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerInterval",
 		Version:      "2.6.5",
-		Doc:          "jsonkey task interval per trigger",
+		Doc:          "Deprecated: JSON stats tasks now run on dataCoord.taskCheckInterval.",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerInterval"},
 		PanicIfEmpty: false,
--- a/pkg/util/paramtable/component_param.go
+++ b/pkg/util/paramtable/component_param.go
@@ -5612,13 +5612,11 @@ type dataCoordConfig struct {
 	StatsTaskSlotUsage                   ParamItem `refreshable:"true"`
 	AnalyzeTaskSlotUsage                 ParamItem `refreshable:"true"`
 
-	EnableSortCompaction       ParamItem `refreshable:"true"`
-	TaskCheckInterval          ParamItem `refreshable:"true"`
-	SortCompactionTriggerCount ParamItem `refreshable:"true"`
-	StatsTaskPendingLimit      ParamItem `refreshable:"true"`
-	// Deprecated: JSON stats tasks are throttled by StatsTaskPendingLimit.
-	JSONStatsTriggerCount ParamItem `refreshable:"true"`
-	// Deprecated: JSON stats tasks now run on TaskCheckInterval.
+	EnableSortCompaction             ParamItem `refreshable:"true"`
+	TaskCheckInterval                ParamItem `refreshable:"true"`
+	SortCompactionTriggerCount       ParamItem `refreshable:"true"`
+	StatsTaskPendingLimit            ParamItem `refreshable:"true"`
+	JSONStatsTriggerCount            ParamItem `refreshable:"true"`
 	JSONStatsTriggerInterval         ParamItem `refreshable:"true"`
 	JSONStatsMaxShreddingColumns     ParamItem `refreshable:"true"`
 	JSONStatsShreddingRatioThreshold ParamItem `refreshable:"true"`
@@ -7237,22 +7235,22 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	p.JSONStatsTriggerCount = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerCount",
 		Version:      "2.6.5",
-		Doc:          "Deprecated: JSON stats tasks are throttled by dataCoord.statsTaskPendingLimit.",
+		Doc:          "jsonkey stats task count per trigger",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerCount"},
 		PanicIfEmpty: false,
-		Export:       false,
+		Export:       true,
 	}
 	p.JSONStatsTriggerCount.Init(base.mgr)
 
 	p.JSONStatsTriggerInterval = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerInterval",
 		Version:      "2.6.5",
-		Doc:          "Deprecated: JSON stats tasks now run on dataCoord.taskCheckInterval.",
+		Doc:          "jsonkey task interval per trigger",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerInterval"},
 		PanicIfEmpty: false,
-		Export:       false,
+		Export:       true,
 	}
 	p.JSONStatsTriggerInterval.Init(base.mgr)
 
--- a/pkg/util/paramtable/component_param.go
+++ b/pkg/util/paramtable/component_param.go
@@ -5361,11 +5361,13 @@ type dataCoordConfig struct {
 	StatsTaskSlotUsage                   ParamItem `refreshable:"true"`
 	AnalyzeTaskSlotUsage                 ParamItem `refreshable:"true"`
 
-	EnableSortCompaction             ParamItem `refreshable:"true"`
-	TaskCheckInterval                ParamItem `refreshable:"true"`
-	SortCompactionTriggerCount       ParamItem `refreshable:"true"`
-	StatsTaskPendingLimit            ParamItem `refreshable:"true"`
-	JSONStatsTriggerCount            ParamItem `refreshable:"true"`
+	EnableSortCompaction       ParamItem `refreshable:"true"`
+	TaskCheckInterval          ParamItem `refreshable:"true"`
+	SortCompactionTriggerCount ParamItem `refreshable:"true"`
+	StatsTaskPendingLimit      ParamItem `refreshable:"true"`
+	// Deprecated: JSON stats tasks are throttled by StatsTaskPendingLimit.
+	JSONStatsTriggerCount ParamItem `refreshable:"true"`
+	// Deprecated: JSON stats tasks now run on TaskCheckInterval.
 	JSONStatsTriggerInterval         ParamItem `refreshable:"true"`
 	JSONStatsMaxShreddingColumns     ParamItem `refreshable:"true"`
 	JSONStatsShreddingRatioThreshold ParamItem `refreshable:"true"`
@@ -6868,22 +6870,22 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	p.JSONStatsTriggerCount = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerCount",
 		Version:      "2.6.5",
-		Doc:          "jsonkey stats task count per trigger",
+		Doc:          "Deprecated: JSON stats tasks are throttled by dataCoord.statsTaskPendingLimit.",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerCount"},
 		PanicIfEmpty: false,
-		Export:       true,
+		Export:       false,
 	}
 	p.JSONStatsTriggerCount.Init(base.mgr)
 
 	p.JSONStatsTriggerInterval = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerInterval",
 		Version:      "2.6.5",
-		Doc:          "jsonkey task interval per trigger",
+		Doc:          "Deprecated: JSON stats tasks now run on dataCoord.taskCheckInterval.",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerInterval"},
 		PanicIfEmpty: false,
-		Export:       true,
+		Export:       false,
 	}
 	p.JSONStatsTriggerInterval.Init(base.mgr)
 
--- a/pkg/util/paramtable/component_param.go
+++ b/pkg/util/paramtable/component_param.go
@@ -5361,13 +5361,11 @@ type dataCoordConfig struct {
 	StatsTaskSlotUsage                   ParamItem `refreshable:"true"`
 	AnalyzeTaskSlotUsage                 ParamItem `refreshable:"true"`
 
-	EnableSortCompaction       ParamItem `refreshable:"true"`
-	TaskCheckInterval          ParamItem `refreshable:"true"`
-	SortCompactionTriggerCount ParamItem `refreshable:"true"`
-	StatsTaskPendingLimit      ParamItem `refreshable:"true"`
-	// Deprecated: JSON stats tasks are throttled by StatsTaskPendingLimit.
-	JSONStatsTriggerCount ParamItem `refreshable:"true"`
-	// Deprecated: JSON stats tasks now run on TaskCheckInterval.
+	EnableSortCompaction             ParamItem `refreshable:"true"`
+	TaskCheckInterval                ParamItem `refreshable:"true"`
+	SortCompactionTriggerCount       ParamItem `refreshable:"true"`
+	StatsTaskPendingLimit            ParamItem `refreshable:"true"`
+	JSONStatsTriggerCount            ParamItem `refreshable:"true"`
 	JSONStatsTriggerInterval         ParamItem `refreshable:"true"`
 	JSONStatsMaxShreddingColumns     ParamItem `refreshable:"true"`
 	JSONStatsShreddingRatioThreshold ParamItem `refreshable:"true"`
@@ -6870,22 +6868,22 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	p.JSONStatsTriggerCount = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerCount",
 		Version:      "2.6.5",
-		Doc:          "Deprecated: JSON stats tasks are throttled by dataCoord.statsTaskPendingLimit.",
+		Doc:          "jsonkey stats task count per trigger",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerCount"},
 		PanicIfEmpty: false,
-		Export:       false,
+		Export:       true,
 	}
 	p.JSONStatsTriggerCount.Init(base.mgr)
 
 	p.JSONStatsTriggerInterval = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerInterval",
 		Version:      "2.6.5",
-		Doc:          "Deprecated: JSON stats tasks now run on dataCoord.taskCheckInterval.",
+		Doc:          "jsonkey task interval per trigger",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerInterval"},
 		PanicIfEmpty: false,
-		Export:       false,
+		Export:       true,
 	}
 	p.JSONStatsTriggerInterval.Init(base.mgr)
 
--- a/pkg/util/paramtable/component_param.go
+++ b/pkg/util/paramtable/component_param.go
@@ -5615,7 +5615,8 @@ type dataCoordConfig struct {
 	EnableSortCompaction       ParamItem `refreshable:"true"`
 	TaskCheckInterval          ParamItem `refreshable:"true"`
 	SortCompactionTriggerCount ParamItem `refreshable:"true"`
-	// Deprecated: JSON stats tasks now use SortCompactionTriggerCount for throttling.
+	StatsTaskPendingLimit      ParamItem `refreshable:"true"`
+	// Deprecated: JSON stats tasks are throttled by StatsTaskPendingLimit.
 	JSONStatsTriggerCount ParamItem `refreshable:"true"`
 	// Deprecated: JSON stats tasks now run on TaskCheckInterval.
 	JSONStatsTriggerInterval         ParamItem `refreshable:"true"`
@@ -7223,14 +7224,24 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	}
 	p.SortCompactionTriggerCount.Init(base.mgr)
 
+	p.StatsTaskPendingLimit = ParamItem{
+		Key:          "dataCoord.statsTaskPendingLimit",
+		Version:      "2.7.0",
+		Doc:          "skip submitting new stats tasks when the global scheduler holds more pending tasks than this limit",
+		DefaultValue: "100",
+		PanicIfEmpty: false,
+		Export:       false,
+	}
+	p.StatsTaskPendingLimit.Init(base.mgr)
+
 	p.JSONStatsTriggerCount = ParamItem{
 		Key:          "dataCoord.jsonShreddingTriggerCount",
 		Version:      "2.6.5",
-		Doc:          "Deprecated: JSON stats tasks now use dataCoord.sortCompactionTriggerCount for throttling.",
+		Doc:          "Deprecated: JSON stats tasks are throttled by dataCoord.statsTaskPendingLimit.",
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerCount"},
 		PanicIfEmpty: false,
-		Export:       true,
+		Export:       false,
 	}
 	p.JSONStatsTriggerCount.Init(base.mgr)
 
@@ -7241,7 +7252,7 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 		DefaultValue: "10",
 		FallbackKeys: []string{"dataCoord.jsonStatsTriggerInterval"},
 		PanicIfEmpty: false,
-		Export:       true,
+		Export:       false,
 	}
 	p.JSONStatsTriggerInterval.Init(base.mgr)
 
--- a/pkg/util/paramtable/component_param.go
+++ b/pkg/util/paramtable/component_param.go
@@ -5364,7 +5364,8 @@ type dataCoordConfig struct {
 	EnableSortCompaction       ParamItem `refreshable:"true"`
 	TaskCheckInterval          ParamItem `refreshable:"true"`
 	SortCompactionTriggerCount ParamItem `refreshable:"true"`
-	// Deprecated: JSON stats tasks now use SortCompactionTriggerCount for throttling.
+	StatsTaskPendingLimit      ParamItem `refreshable:"true"`
+	// Deprecated: JSON stats tasks are throttled by StatsTaskPendingLimit.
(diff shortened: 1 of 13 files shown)


czs007:
## Adversarial Review Summary

This review surfaced one high-severity upgrade-path regression around JSON stats throttling, plus three medium-severity issues covering scheduling admission accuracy, trigger fairness, and silent config deprecation.

---

### 🔴 High

**`internal/datacoord/stats_inspector.go:308` — Removing the JSON stats submission throttle lets upgrade-triggered re-shredding consume full worker capacity**

This PR removes the per-round / per-interval cap on JSON key-index task submission (`maxJSONStatsTaskCount >= JSONStatsTriggerCount → break`, combined with the 10-minute `JSONStatsTriggerInterval` reset — effectively 10 tasks per 10 minutes globally). The only remaining brake is `canSubmitStatsTask`'s `pendingTaskCount > statsTaskPendingLimit` (100), which bounds *queue depth*, not *submission rate*: `schedule()` drains the pending queue onto any node with a free slot every `TaskScheduleInterval` (100ms), so the queue is emptied and refilled every round — equivalent to re-shredding at full worker capacity. Meanwhile `needDoJSONKeyIndex` returns true whenever a segment's `JsonKeyStatsDataFormat < common.JSONStatsDataFormatVersion` (currently 3, already bumped from 2 historically), and `enabledJSONShredding` defaults to true, so every sealed segment with a JSON field becomes a candidate on the first `taskCheckInterval` tick after an upgrade. Because `pendingTasks` is a min-heap ordered by task ID (FCFS) and is shared with index build, compaction, and import, these early re-shred tasks are dispatched ahead of newer work on the same slot pool.

*Failure scenario:* upgrading an existing cluster with JSON fields to this build (segments at `JsonKeyStatsDataFormat < 3`, or with no `JsonKeyStats` at all) with `enabledJSONShredding` left at its default → from the first 60s tick onward, re-shred tasks for all pre-existing sealed segments hold the pending queue at 100 and saturate DataNode/IndexNode slots (~16 slots per 512MB segment); FCFS ordering puts newly created index-build and compaction tasks behind them, slowing both for the entire migration window (potentially hours). Before this change the same migration was rate-limited to 10 tasks per 10 minutes.

*Suggestion:* either retain a per-round / per-interval submission cap for the JSON sub-job (reusing `JSONStatsTriggerCount` is the cheapest option and preserves the `0 = disabled` semantics), or, if dropping the throttle is intentional, export `statsTaskPendingLimit` in `configs/milvus.yaml` and document this upgrade-time resource behavior change in the release notes. *(raised by czs007, chyezh)*

---

### 🟡 Medium

**`internal/datacoord/task/global_scheduler.go:276` — `GetPendingTaskCount` counts backoff tasks, reporting saturation while workers sit idle**

`schedule()` collects tasks in failure backoff into `delayed` (`:276-279`) and pushes them back into `pendingTasks` at the end of the round (`:321-323`); `check()` and the `CreateTaskOnWorker` failure path re-push as well (`:295`, `:347`). None of these hold a worker slot, yet all are counted by `GetPendingTaskCount()` → `pendingTasks.TaskCount()`, which returns `len(pqp.tasks)` without distinguishing backoff state. Since `pendingTasks` is shared across index, compaction, import, analyze, and stats, a batch of persistently failing unrelated tasks can pin the admission signal at "saturated." The backoff ceiling (60s) is the same order as the stats trigger interval (60s), so this false saturation is a stable state rather than transient jitter.

*Failure scenario:* object-storage credential/permission errors cause >100 import or index-build tasks to fail repeatedly (each entering up to 60s of backoff while remaining in `pendingTasks`) while DataNode/IndexNode slots are largely idle → `canSubmitStatsTask` continuously reads `pendingTaskCount > 100` and returns false, all three trigger loops return early, and no text index or JSON shredding task is submitted for the duration of the failure storm even though workers are idle.

*Suggestion:* exclude backoff tasks from the count, or use the number of tasks actually occupying slots as the admission signal. *(raised by czs007)*

**`internal/datacoord/stats_inspector.go:138` — Fixed trigger order plus whole-function `return` on limit starves JSON stats behind text stats**

`triggerStatsTaskLoop` invokes triggers in a fixed order — text → BM25 → JSON (`:138-140`) — and BM25 is a no-op since `enableBM25()` is constantly false (`:324-326`). Both the text and JSON loops `return` from the whole function when the pending limit is hit (text at `:225-227` and `:265-267`, JSON at `:283-285` and `:309-311`), and text always runs first. As a result, JSON receives only whatever capacity text leaves behind rather than a fair share.

*Failure scenario:* a cluster with both text-match-enabled fields and JSON fields, with a large backlog of text index candidates (e.g. text match newly enabled on a large collection, or `TextStatsLogs` missing after an upgrade) such that text fills pending past `statsTaskPendingLimit` on every 60s tick → `triggerJSONKeyIndexStatsTask` returns at its first `canSubmitStatsTask` check every round, submitting zero JSON shredding tasks for the entire text migration and leaving JSON queries on the unaccelerated path until the text backlog drains.

*Suggestion:* rotate the trigger order across ticks, or give each sub-job type an independent quota. *(raised by chyezh, czs007)*

**`internal/datacoord/stats_inspector.go:277` — `jsonShreddingTriggerCount` is silently deprecated, breaking the existing "set 0 to disable JSON shredding" operator switch**

Previously, `if maxJSONStatsTaskCount >= Params.DataCoordCfg.JSONStatsTriggerCount.GetAsInt() { break }` meant a configured value of 0 broke immediately on the first segment — effectively disabling JSON key-index submission entirely. That was an exported, documented knob (`jsonShreddingTriggerCount: 10` in `configs/milvus.yaml`) operators could use to stop shredding. After this PR the key is still accepted by paramtable (the `ParamItem` and `FallbackKeys` are retained, only `Doc` and `Export` changed), but there is no remaining read site anywhere in the code — `git grep` for `JSONStatsTriggerCount`/`JSONStatsTriggerInterval` outside `component_param.go` returns nothing. The configuration neither errors nor warns; its semantics silently disappear. This is distinct from the throttle removal above: here an *already-configured user setting* is silently ignored after upgrade.

*Failure scenario:* an operator who previously set `dataCoord.jsonShreddingTriggerCount` to 0 (or a small value) to suppress JSON shredding upgrades to this build → the setting is silently ignored with no log message, JSON key-index tasks resume for all candidate segments and saturate worker slots up to `statsTaskPendingLimit` — precisely the behavior the operator meant to prevent — with no visible cause, since the config file still shows 0.

*Suggestion:* either preserve the `0 = disabled` semantics (at least for the JSON sub-job), or emit an explicit startup warning when a deprecated key is present with a non-default value, and note the change in the release notes. *(raised by czs007)*

xiaocai2333:
Thanks @czs007 @chyezh — the review was against `4b7a085`; head is now `ccbc1c9978`. Point-by-point below, including the two I am **not** changing and why.

### 🔴 High — JSON re-shredding can run at full worker capacity after an upgrade

Accepted as an accurate description; the "the gate bounds queue depth, not submission rate" framing corrected my own earlier reasoning (I had treated the ~limit+1 instantaneous ceiling as a rate bound, which it is not — `schedule()` drains and the loop refills within the same tick). I verified the amplification chain: `common.JSONStatsDataFormatVersion = 3` (`pkg/common/common.go:194`) plus `needDoJSONKeyIndex`'s `< version` branch plus `common.enabledJSONShredding` defaulting to true does make every pre-existing sealed segment with a JSON field a candidate on the first tick after upgrade.

**Not adding a per-sub-job submission cap back** — sharing one budget with FCFS ordering is the intended model here. Taking the documented-fallback option you offered instead:

- The PR body now has an **"Upgrade behavior to be aware of"** section stating the depth-not-rate semantics, the `JsonKeyStatsDataFormat < 3` amplification, the FCFS interaction with index/compaction on the same slot pool, and the three levers (`common.enabledJSONShredding=false`, `dataCoord.jsonShreddingTriggerCount=0`, or lowering `dataCoord.statsTaskPendingLimit`).
- `jsonShreddingTriggerCount: 0` keeps working as a kill switch — see the third item below.

### 🟡 `jsonShreddingTriggerCount`'s `0 = disabled` semantics silently disappeared

Fixed in `ccbc1c9978`. This one was a genuine miss on my side. Verified against the base: `if maxJSONStatsTaskCount >= JSONStatsTriggerCount.GetAsInt() { break }` with the key at 0 broke on the very first segment, so 0 did disable JSON key-index submission outright.

Following the existing `queryNode.cache.warmup` precedent (`internal/util/initcore/init_core.go:330-338`) — keep one value's effect, warn about the rest:

- `triggerJSONKeyIndexStatsTask` returns early while the key reads 0, with a rated warning pointing at `common.enabledJSONShredding`.
- DataCoord warns at startup for each deprecated throttle key left at a non-default value, naming the key and its value, so a setting like `3` that no longer does anything is visible in the log rather than silently ignored.

### 🟡 Fixed trigger order starves JSON behind text

Fixed in `ccbc1c9978`. Extracted `triggerStatsTasks(round)` and alternate text/JSON per round (BM25 stays last; it returns immediately while `enableBM25()` is false). Same conclusion was reached independently in a local review pass, so three sources agree — thanks for pinning down the "zero JSON tasks for the whole text migration" phrasing.

### 🟡 `GetPendingTaskCount` counts backoff tasks

**Not changing, deliberately.** The specifics are right — `taskRetryBackoffMaxInterval` defaults to 60s and `taskCheckInterval` to 60s, so the false-saturation state is stable rather than transient, and backoff tasks hold no worker slot.

The reason for leaving it: when >100 tasks are cycling through retry, the actionable problem is whatever makes them fail (the S3 credential/permission case in your scenario), not the stats tasks that are being held back — a stats task admitted into that same storm would very likely fail too. Throttling on total queue depth also keeps the signal simple and type-agnostic. Recording it as a known limitation rather than a refutation: if this shows up in practice, excluding tasks whose `notBefore` has not elapsed is the narrow fix.

### Not-for-PR observations

Both agree with where this landed:

- **`statsTaskPendingLimit` not exported** — intentional, matching its neighbours `taskCheckInterval` and `sortCompactionTriggerCount`. It stays overridable via env/etcd, and requiring it in the yaml template is not a goal. Thanks for downgrading the original `high`.
- **`SubmitStatsTask` returning `nil` when refused** — same conclusion here: the only callers are the three trigger loops in this file, each of which rediscovers the segment on the next `taskCheckInterval` tick, so nothing is dropped today. Left as-is rather than widening the interface for a hypothetical caller.

### Also fixed since the reviewed commit

`4b7a085` had Code-Check red (`QF1001: could apply De Morgan's law` on the BM25 selector, introduced in `c275193`), which also skipped UT-GO and UT-Integration. Fixed in `ad18bd64fd`; `golangci-lint` v2.11.3 (the CI version) is now clean over `./internal/datacoord/...`, `pkg/util/paramtable/...` and `tests/go_client`, and that check is part of the pre-push loop now.

### Verification of the new behavior

Each new test was mutation-checked — the behavior was removed and the test confirmed to fail, then restored:

- drop the per-round alternation → the sub-job-order test fails;
- drop the `0 = disabled` path → the deprecated-zero test fails (strict allocator mock catches the ID allocation);
- drop the selector's `HasStatsTask` filter → the skip test fails;
- turn the collection-level `return` into `continue` → the backlog test fails on the admission-call count.

`./internal/datacoord/...` and `cmd/tools/config -run TestYamlFile` pass locally.

czs007:
**Overview:** Adversarial review found two medium-severity scheduling issues, both stemming from the new stats-task admission gate relying on the depth of a shared pending queue instead of a stats-specific or rate-based signal.

## Medium

**`internal/datacoord/stats_inspector.go:254` — `canSubmitStatsTask` treats backing-off tasks as worker saturation**

`canSubmitStatsTask` (L254-265) reads only `scheduler.GetPendingTaskCount()`, which is just the length of the shared `pendingTasks` queue (`task/global_scheduler.go:127-129` → `task/priority_queue.go:105-110`). That queue mixes two very different populations: tasks genuinely waiting for a slot (pushed back after `pickNode` returns `NullNodeID`, `global_scheduler.go:281-285`) and tasks in retry backoff (`global_scheduler.go:276-279`, `321-323`), which occupy zero worker slots. Backoff is easy to enter — for stats tasks, any failure in `UpdateTaskVersion`/`CreateStats` resets the task to `Init` via the deferred `resetTask` (`task_stats.go:155-160`), and a failed `QueryStats` RPC does the same through `dropAndResetTaskOnWorker` (`task_stats.go:253-254`) — and it persists, since `dataCoord.taskRetryBackoffMaxInterval` defaults to 60s. The queue is also shared across index/compaction/import/analyze/stats, so it is not a stats-specific signal either.

*Failure:* etcd write jitter or a half-dead datanode pushes >100 tasks (of any type) into `recordTaskFailure` backoff → `GetPendingTaskCount()` stays above `statsTaskPendingLimit`, `canSubmitStatsTask` is permanently false, and `triggerTextStatsTask`/`triggerJSONKeyIndexStatsTask` return at the first collection, so no new text-index or JSON-shredding stats tasks are created at all — even while every healthy datanode slot is idle, leaving those indexes missing for as long as the failing tasks keep cycling.

*Suggestion:* exclude backing-off tasks from `GetPendingTaskCount()`, or expose the count per task type. Otherwise the premise in the comment at L250-253 — "a deep queue means the workers are already saturated" — does not hold. *(raised by bigsheeper)*

---

**`internal/datacoord/stats_inspector.go:325` — replacing the JSON-shredding rate limit with a queue-depth cap can starve other task types during upgrade migration**

The removed `jsonShreddingTriggerCount/Interval` were a *rate* limit (at most N per interval); `statsTaskPendingLimit` is a *depth* limit. Since `schedule()` drains pending onto workers every `TaskScheduleInterval` (100ms) and `triggerJSONKeyIndexStatsTask` refills every check tick, JSON shredding now effectively runs at full worker throughput. That matters on upgrade: `needDoJSONKeyIndex` returns true for any sealed segment with `GetJsonKeyStats() == nil` or `JsonKeyStatsDataFormat < 3` (L216-238), `EnableJSONKeyStatsIndex` is unconditionally true for JSON fields (`pkg/util/typeutil/field_schema.go:56-58`), and `common.enabledJSONShredding` defaults to true — so every sealed segment becomes a candidate at once. The pending queue is a min-heap keyed by `taskID` and shared with index/compaction/import (`task/priority_queue.go:37-48`).

*Failure:* a cluster that never had JSON shredding enabled upgrades to this version with tens of thousands of JSON-bearing sealed segments → from the first `taskCheckInterval` tick the shared pending queue sits near the 100-task limit filled with JSON shredding tasks whose smaller taskIDs dequeue ahead of newly created index-build/compaction/import tasks, stalling new index builds and compaction for the duration of the migration window; the previous version rate-limited the same migration to 10 tasks per 10 minutes.

*Suggestion:* keep a per-round submission cap (e.g. a submit-per-tick quota keyed by `subJobType`), or raise/retune `statsTaskPendingLimit` to a value comparable to the old rate. The current operator options are all-or-nothing (`enabledJSONShredding=false`, `jsonShreddingTriggerCount=0`) or lowering `statsTaskPendingLimit`, which also throttles text indexing. *(surfaced during verification)*

xiaocai2333:
Updated in `a864d335a0` after re-checking the PR's throughput objective:

- Kept the admission fix from `88f179e205`: `GetPendingTaskCount` counts only tasks eligible for dispatch and excludes tasks whose retry `notBefore` deadline is still in the future. The focused test enters backoff through the real `recordTaskFailure` path and verifies that an expired backoff becomes countable again.
- Reverted the accidental restoration of `jsonShreddingTriggerCount` + `jsonShreddingTriggerInterval`. Removing that 10-tasks-per-10-minutes ceiling is the purpose of this PR; the shared runnable-backlog gate remains depth-based admission control, not a replacement rate limiter.
- Restored the intended deprecation behavior and retained `jsonShreddingTriggerCount: 0` only as the compatibility kill switch.

Full DataCoord tests, paramtable/config tests, and scoped golangci-lint v2.11.3 pass locally. The PR body now explicitly documents the throughput choice, upgrade behavior, and operator controls.

czs007:

Both reviewers independently re-examined the final round and confirmed there were no additional findings, and no issues survived verification.

**Verified:**
- Final-round re-review by `gzh` — independently re-examined the change set and confirmed no additional issues.
- Final-round re-review by `yezh` — independently re-examined the change set and confirmed no additional issues.

*Note: the reviewer notes supplied for this round record only the reviewers' concurring verdicts (each linking to their full untruncated report) and do not enumerate specific file:line checks, so a more granular verified list cannot be produced without inventing entries.*

pkg/util/paramtable/component_param.go
```diff
@@ -6854,25 +6857,35 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	}
 	p.SortCompactionTriggerCount.Init(base.mgr)
 
+	p.StatsTaskPendingLimit = ParamItem{
+		Key:          "dataCoord.statsTaskPendingLimit",
+		Version:      "2.7.0",
```
czs007: there is no 2.7.0

xiaocai2333: Fixed in 27cac735bd — changed to 3.0.0 (also applied to the 3.0 backport #52010).

czs007:

## Overview

The new stats admission gate in `canSubmitStatsTask` counts the *global* scheduler queue with a backoff carve-out, which makes the gate both fail open during worker-side failure storms and fail closed under unrelated index/compaction backlog.

## High

**`internal/datacoord/task/global_scheduler.go:132` — `GetPendingTaskCount()` fails open during a worker-side failure storm, disabling the admission gate and degrading submission to O(N²)**

A stats task that fails on the dispatch path has neither a terminal state nor a retry ceiling: `CreateTaskOnWorker` (`internal/datacoord/task_stats.go:155-159`) defers `resetTask`, which writes `JobState_JobStateInit` (`task_stats.go:132-135`), and both `req, err := st.prepareJobRequest(...)` (`task_stats.go:198`) and `if err = cluster.CreateStats(...)` (`task_stats.go:211`) assign the outer `err`, so both reach it. Back in `schedule()` (`global_scheduler.go:360-365`) the state is still `Init`, so the task goes through `recordTaskFailure` + `pendingTasks.Push` with no failure-count ceiling — `recordTaskFailure` (`:80-106`) only pushes `notBefore` out, capped at `TaskRetryBackoffMaxInterval` (60s default). The task then cycles `pending → popped → fail → backoff → pending` indefinitely, and the new filter `!ok || !now.Before(backoff.notBefore)` (`:132-138`) only sees it in the sliver between `notBefore` elapsing and the next 100 ms `schedule()` tick — roughly 0.17% visibility under a 60s backoff, so 10k failing tasks are counted as ~17 against a default limit of 100.

Failure scenario: DataNodes registered with free slots but `CreateStats`/`prepareJobRequest` persistently failing (half-dead node, throttled object storage, binlog read errors) on a cluster that just upgraded `JSONStatsDataFormatVersion` → the count trends to ~0, `dataCoord.statsTaskPendingLimit` stops applying, and a single discovery tick writes one `StatsTask` per candidate segment cluster-wide into etcd and `pendingTasks`; meanwhile each per-segment `GetPendingTaskCount` call scans the whole pending map under the priority-queue `RLock` without the usual early stop at ~101, so the tick degrades to O(N²) and contends with `schedule()`.

Suggestion: scope the count to stats work (`TaskCountBy` filtering `GetTaskType() == taskcommon.Stats`) and drop the backoff carve-out — tasks in backoff genuinely occupy queue depth. *(raised by czs007, chyezh)*

**`internal/datacoord/stats_inspector.go:254` — admission is gated on all-type queue depth, so index/compaction/import backlog starves text/JSON stats submission indefinitely**

`canSubmitStatsTask` calls `si.scheduler.GetPendingTaskCount()` directly, and that queue holds every task type: compaction (`compaction_inspector.go:294`), import (`import_inspector.go:123`), index build (`index_inspector.go:245`), analyze (`analyze_inspector.go:59`), copy segment (`copy_segment_inspector.go:255`), external refresh. Index build is one task per `(segment, index)`, and `indexInspector.reloadFromMeta` (`index_inspector.go:265-289`) enqueues every `Unissued`/`Retry`/`InProgress` segment index at once on DataCoord restart, so on a write-heavy cluster the backlog routinely exceeds the default limit of 100. Before this PR `triggerTextStatsTask` had no admission check at all and the JSON path used its own `maxJSONStatsTaskCount` limiter, neither of which was affected by other task types — this is a behavioural regression.

Failure scenario: build indexes on a 500-segment collection, or restart DataCoord with an index backlog, pushing pending index + compaction tasks above 100 → `triggerTextStatsTask` returns at the first collection (`:273-275`) and `triggerJSONKeyIndexStatsTask` likewise (`:337-339`), so no text-index or JSON-shredding task is submitted for the entire backlog window and never recovers until the backlog drops below 101; newly flushed segments hit by `TEXT_MATCH` queries then throw `TextIndexNotFound` (`internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:1909-1919`) rather than degrading.

Suggestion: same fix as above — restrict `TaskCountBy` to `taskcommon.Stats`, or give stats a dedicated counting entry point. *(surfaced during verification)*

## Medium

**`internal/datacoord/task/global_scheduler.go:348` — version-constrained, currently undispatchable tasks are counted as runnable, contradicting the interface comment**

When `pickNodeWithMinimumVersion` finds no compatible worker, the `versionConstrained` branch appends to `delayed` and re-pushes at `:391-393` without calling `recordTaskFailure`, so these tasks never get a `backoffs` entry. The filter treats `!ok` as runnable (`:135-136`), so they are counted as "eligible for dispatch now" for as long as no compatible worker exists — the opposite of the doc comment at `:42-44`, and the mirror image of the backoff carve-out above. The only version-constrained task today is `copySegmentTask` (`copy_segment_task.go:324-331`, DataNode ≥ 3.0.1 at `:80`), and `copy_segment_checker.go:344-345` chunks by `maxSegmentsPerCopyTask` (default 100), so crossing the limit needs >10k segments in external restore jobs.

Failure scenario: restore an external snapshot with >10k segments while every DataNode is still below 3.0.1 (rolling upgrade in progress or aborted), producing >101 undispatchable tasks → the count stays above `dataCoord.statsTaskPendingLimit`, both stats triggers return at the first collection, and no text-index or JSON-shredding task is submitted for the whole upgrade window; if the rollout is aborted this does not self-heal.

Suggestion: covered by the same stats-type-scoped count; if the general-purpose counter is kept for other callers, exclude version-constrained tasks so the value matches its documented meaning. *(raised by czs007, chyezh)*

czs007:

**Verified:**
- `internal/datacoord/task/global_scheduler.go:136-139` — `GetPendingTaskCount` now counts every queued task of the requested type without excluding entries whose retry `notBefore` is in the future, so tasks in backoff remain visible to admission control.
- `internal/datacoord/task/global_scheduler.go:364-367` — failed `CreateTaskOnWorker` dispatches are returned to `pendingTasks` via `recordTaskFailure` + `Push`, and are therefore counted; the gate no longer fails open during a worker-side failure storm (the `check()` path at `global_scheduler.go:416-419` behaves the same for `Retry` tasks).
- `internal/datacoord/stats_inspector.go:257-260` — `canSubmitStatsTask` calls `GetPendingTaskCount(taskcommon.Stats)`, and `statsTask.GetTaskType()` returns `taskcommon.Stats` (`internal/datacoord/task_stats.go:83-85`), so index/compaction/import/analyze backlogs no longer consume the stats admission budget.
- `internal/datacoord/task/priority_queue.go:113-124` — the `TaskCountBy` filter invokes `GetTaskType()` under `RLock`; all 11 `GetTaskType()` implementations return constants and take no locks, so there is no lock-ordering or deadlock risk.
- `internal/datacoord/task/global_scheduler_test.go:110-158` — `TestGlobalScheduler_GetPendingTaskCountIsScopedByTaskType` (mixed Stats/Compaction/Index → 1/2/1) and `TestGlobalScheduler_GetPendingTaskCountIncludesBackoff` (count remains 2 after `recordTaskFailure`) distinguish the intended behavior.
- `pkg/taskcommon/type.go:19` — `taskcommon.Type` is an alias of `string`, so the generated mock signature `GetPendingTaskCount(taskType string)` matches the interface; `go vet -tags dynamic,test ./internal/datacoord/task/...` passes.

No surviving issues, disputed items, or pre-existing hazards were reported for this PR.

czs007:

**Verified:**
- `internal/datacoord/stats_inspector.go:455-489` — `SubmitStatsTask` ordering (segment existence → external branch → `HasStatsTask` dedup → admission → `AllocID`) preserves the external-collection short-circuit and lets duplicates bypass the admission budget.
- `internal/datacoord/task/task_stats.go:83-85` — `taskcommon.Stats` is produced only by `statsTask`, so `GetPendingTaskCount(taskcommon.Stats)` covers exactly the stats population and nothing else consumes its budget.
- `internal/datacoord/stats_inspector.go:99-117` — `warnDeprecatedThrottleConfigs` compares against `DefaultValue` via `ParamItem.getWithRaw`, so clean installs and untouched upgraded configmaps stay silent; the old `jsonStatsTriggerCount: 0` kill-switch is preserved through `FallbackKeys`.
- `internal/datacoord/task/priority_queue.go:73-132` — `Push`/`Pop`/`Remove` mutate `tasks` and `heap` under the same write lock, so `TaskCount()` is exactly equivalent to the `len(TaskIDs())` it replaces.
- `internal/datacoord/task/global_scheduler.go:322-396` — every popped task lands in `delayed`, `runningTasks`, or is re-pushed on failure; no task-loss path, and `delayed` is refilled before `AwaitAll`.
- `internal/datacoord/stats_inspector_test.go:335-349` — `TestSubmitStatsTask` still exercises the dedup guard: the first call genuinely populates `"10-Sort"` (collection 1 has no `ExternalSource`), so the strict zero-expectation mocks assert the guard, not a hidden fixture.

## Low (non-blocking)

- **PR description** — The PR body states the opposite of the code: `internal/datacoord/task/global_scheduler.go:45-47` documents that backoff tasks *are* counted, `GetPendingTaskCount` (`:136-140`) applies no `taskInBackoff` filter, and `TestGlobalScheduler_GetPendingTaskCountIncludesBackoff` asserts `count == 2` after `recordTaskFailure`. The body's "Tasks waiting for a retry-backoff deadline are excluded" and the "runnable-vs-backoff" verification bullet describe the semantics before 74abd95 reverted them. Suggestion: update the PR description (and the Verification section) to say backoff tasks are included and why, so operators do not derive an inverted capacity conclusion from it. *(raised by czs007, tinswzy)*

- **`internal/datacoord/task/global_scheduler.go:322`** — During the `schedule()` pop loop, popped backoff tasks sit in the local `delayed` slice until `:393-395` (before `AwaitAll`), and tasks already popped but not yet returned from `CreateTaskOnWorker` are in neither `pendingTasks` nor `runningTasks`; both are invisible to `GetPendingTaskCount`. Since `execPool.Submit` blocks when all 128 workers are busy (`pkg/util/conc/pool.go:63-65`), the window is extended when a round dispatches >128 tasks and the worker RPC is slow. Failure: pending stats tasks near `statsTaskPendingLimit` + >128 dispatchable tasks in one round with slow `CreateTaskOnWorker` + the 60s `statsInspector` tick landing in that window → `canSubmitStatsTask` reads a deflated count and admits up to roughly "hidden task count" extra stats tasks that tick (extra `AllocID` + etcd writes + `Enqueue`); the overshoot is bounded and self-corrects on the next tick. Trigger conditions are narrow (slots available but RPC slow) and the PR is still a net improvement over having no admission gate. Suggestion: follow-up — maintain a per-type counter updated in `Enqueue`/terminal transitions, or peek-and-skip instead of pop-and-repush; not required in this PR. *(raised by tinswzy, czs007)*

## Nitpick (non-blocking)

- **`internal/datacoord/stats_inspector.go:276`** (also `:340`, `:389`) — `canSubmitStatsTask` runs before `needTriggerFieldIDs` is computed, so each call performs a `TaskCountBy` scan (`priority_queue.go:112-123`, RLock over the whole shared pending queue across index/compaction/stats types) even for collections that would be skipped by the `len(needTriggerFieldIDs) == 0` `continue` at `:290`/`:352`. Each hold is a single short scan and cannot stall dispatch; it is wasted CPU, measurable only when collection count × queue backlog are both large. Suggestion: move the `canSubmitStatsTask` check after the `continue`. *(raised by czs007, tinswzy)*

czs007 (approved):
/lgtm

Issue #51749, opened 2026-07-23
[Enhancement]: Add mmap-backed storage for sealed nullable vector offset mapping


marcelo-cjl:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Add mmap-backed and resource-accounted storage for `SealedOffsetMapping`, the logical-offset <-> physical-offset mapping used by nullable vector sealed segments.

Current master (`0ec31c4b0bee1e796fb9b4d1547ce98ae388c343`) still keeps this mapping entirely in memory:

- `internal/core/src/common/OffsetMapping.h`
  - `SealedOffsetMapping` owns `std::vector<int32_t> l2p_vec_` and `p2l_vec_` for dense mappings.
  - It owns `std::unordered_map<int32_t, int32_t> l2p_map_` and `p2l_map_` for sparse valid rows.
- `internal/core/src/common/OffsetMapping.cpp`
  - `Build()` chooses sparse map mode with `valid_count * 10 < total_count`; otherwise it resizes and fills the two dense vectors.
- `internal/core/src/mmap/ChunkedColumnInterface.h`
  - nullable sealed field loading builds `valid_data_` from chunks and then calls `offset_mapping_.Build(valid_data_.data(), valid_data_.size())`.

There is no mmap-backed array/storage abstraction in `SealedOffsetMapping`, and I do not see matching resource accounting for this additional mapping memory.

A reasonable target design would be:

1. Add a mmap-capable read-only storage path for the dense `l2p` / `p2l` arrays, similar in spirit to Cardinal's mmapable tier id mappings.
2. Decide the sparse mode policy explicitly:
   - keep `unordered_map` in memory but include it in resource accounting, or
   - replace it with mmapable compact arrays / sorted pairs when lookup latency is acceptable.
3. Wire the mapping allocation to Milvus' existing mmap path, lifecycle, and cache/resource accounting instead of creating ad hoc temporary files.
4. Gate the behavior by config and keep the current in-memory implementation as the default or fallback until validated.
5. Add tests/benchmarks for large nullable vector sealed segments covering dense, sparse, all-valid, and all-null distributions.

### Why is this needed?

Nullable vector compact storage decouples logical row offsets from physical vector offsets. For sealed segments, Milvus can mmap field data and indexes, but the offset mapping itself remains a separate resident memory structure.

For large loaded collections with nullable vector fields, this can become a non-trivial memory cost outside the main mmap/resource plan:

[...]

PR #51817, opened 2026-07-24
enhance: support mmap for sealed offset mapping

marcelo-cjl:
## Summary
- add mmap-backed storage options for sealed nullable vector offset mappings
- wire disk vector index load to dedicated `enable_mmap_i2o_map` and `enable_mmap_o2i_map` config keys instead of generic `enable_mmap`
- keep i2o/o2i mmap controls independent and preserve sparse-map behavior per direction
- extend existing offset mapping, chunked column, and nullable disk index tests with mmap on/off coverage

issue: #51749

## Test plan
- [x] `export jobs="$(nproc)" CONAN_CPU_COUNT="$(nproc)" GOMAXPROCS="$(nproc)" && source ./scripts/setenv.sh && make SKIP_3RDPARTY=1 build-cpp-with-unittest jobs="$(nproc)"`
- [x] `source ./scripts/setenv.sh && ./internal/core/output/unittest/all_tests --gtest_filter='OffsetMapping.*:*ChunkedColumnInterfaceTest*:*VectorArrayColumnInterfaceTest*:DiskAnnFileManagerTest.FilterValidDataDiskFileSlices:DiskAnnFileManagerTest.CacheValidDataDiskFileSlices:DiskAnnFileManagerTest.CacheValidDataMultipleDiskFileSlices:DiskAnnFileManagerTest.LoadStreamIndexCachesOnlyValidDataSidecar:DiskAnnFileManagerTest.CacheRawDataToDiskValidDataFile:DiskAnnFileManagerTest.BuildAllNullNullableDiskVectorIndexFromDataset:DiskAnnFileManagerTest.LoadAllNullNullableDiskVectorIndexFromDataset'`
- [x] `git diff --check upstream/master..HEAD`

### DIFF
--- a/internal/core/src/common/OffsetMapping.cpp
+++ b/internal/core/src/common/OffsetMapping.cpp
@@ -6,6 +6,7 @@
 #include <utility>
 
 #include "common/EasyAssert.h"
+#include "storage/MmapManager.h"
 
 namespace milvus {
 namespace {
@@ -33,6 +34,48 @@ ShouldSkipBitsetTransform(const BitsetView& bitset,
 
 }  // namespace
 
+void
+OffsetMappingArray::Resize(
+    size_t size,
+    int32_t value,
+    bool enable_mmap,
+    const storage::MmapChunkManagerPtr& mmap_chunk_manager,
+    const storage::MmapChunkDescriptorPtr& mmap_descriptor) {
+    Clear();
+    if (size == 0) {
+        return;
+    }
+
+    if (enable_mmap) {
+        AssertInfo(mmap_chunk_manager != nullptr,
+                   "offset mapping mmap chunk manager is null");
+        AssertInfo(mmap_descriptor != nullptr,
+                   "offset mapping mmap descriptor is null");
+        mmap_chunk_manager_ = mmap_chunk_manager;
+        mmap_descriptor_ = mmap_descriptor;
+        mmap_data_ = static_cast<int32_t*>(mmap_chunk_manager_->Allocate(
+            mmap_descriptor_, sizeof(int32_t) * size));
+        AssertInfo(mmap_data_ != nullptr,
+                   "failed to allocate offset mapping mmap buffer, size: {}",
+                   sizeof(int32_t) * size);
+        size_ = size;
+        std::fill_n(mmap_data_, size_, value);
+        return;
+    }
+
+    vec_.assign(size, value);
+    size_ = size;
+}
+
+void
+OffsetMappingArray::Clear() {
+    size_ = 0;
+    vec_.clear();
+    mmap_data_ = nullptr;
+    mmap_descriptor_ = nullptr;
+    mmap_chunk_manager_ = nullptr;
+}
+
 int64_t
 OffsetMapping::GetPhysicalOffset(int64_t logical_offset) const {
     return logical_offset;
@@ -55,6 +98,11 @@ OffsetMapping::GetValidCount() const {
 
 bool
 OffsetMapping::IsEnabled() const {
+    return false;
+}
+
+bool
+OffsetMapping::IsMmap() const {
     return false;
 }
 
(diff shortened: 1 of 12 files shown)


internal/core/src/index/VectorDiskIndex.cpp
```diff
@@ -374,8 +376,22 @@ VectorDiskAnnIndex<T>::Load(milvus::tracer::TraceContext ctx,
...
-        BuildValidDataFromBitmap(
-            this, disk_valid_data.total_count, disk_valid_data.bitmap.data());
+        auto offset_mapping_options = GetOffsetMappingMmapOptions(load_config);
+        if (NeedOffsetMappingMmap(offset_mapping_options,
+                                  disk_valid_data.total_count,
+                                  disk_valid_data.valid_count)) {
+            const auto& mmap_config =
+                storage::MmapManager::GetInstance().GetMmapConfig();
+            offset_mapping_options.mmap_chunk_manager =
+                std::make_shared<storage::MmapChunkManager>(
```
liliu-z: `make_shared<MmapChunkManager>` here constructs a fresh per-index mmap manager whose constructor zeroes the process-global `allocated_size_`, corrupting the node-wide mmap memory accounting that is shared by every other index/segment on the node. Although off by default, this path is operator/tenant-reachable at runtime with no Milvus source change: `knowhere.DISKANN.load.enable_mmap_i2o_map=true` flows through the generic `knowhere.<INDEX>.load.*` passthrough merged at `pkg/util/indexparams/index_params.go:383` (or via cloud `autoIndex.params.load`), reaching `GetOffsetMappingMmapOptions`→`NeedOffsetMappingMmap`→this constructor. The corrupted (undercounted) accounting then defeats mmap capacity limits/eviction and can drive the node to OOM; reuse the existing global MmapChunkManager instead of constructing a per-index one.

marcelo-cjl: Done. Removed the per-index MmapChunkManager creation from VectorDiskIndex::Load. Offset mapping now uses direct file-backed mmap files under the index-local id_mapping_mmap directory, so it stays colocated with the index files and no longer resets the process-global MmapChunkManager accounting.

internal/core/src/index/VectorIndexValidDataUtils.h
```diff
@@ -124,6 +127,33 @@ CountValidDataBitmap(size_t count, const uint8_t* bitmap) {
...
+                      size_t total_count,
+                      size_t valid_count) {
+    return (options.enable_mmap_o2i_map && total_count > 0) ||
+           (options.enable_mmap_i2o_map && valid_count > 0);
+}
+
+inline OffsetMappingBuildOptions
+GetOffsetMappingMmapOptions(const Config& config) {
+    OffsetMappingBuildOptions options;
+    options.enable_mmap_i2o_map =
```
foxspy, line 150: The load path already passes the final mmap decision through the existing `ENABLE_MMAP` config, but `GetOffsetMappingMmapOptions()` ignores it and only reads the newly introduced `enable_mmap_i2o_map` / `enable_mmap_o2i_map` keys. As a result, even when the caller loads the index with `ENABLE_MMAP=true`, the offset mapping still stays on heap. Please reuse the existing load-time `ENABLE_MMAP` value to enable both i2o and o2i mappings, remove the duplicate config keys, and update `MmapOptionsUseOnlyIdMappingConfigKeys` to verify this contract.

internal/core/src/common/OffsetMapping.cpp
```diff
@@ -33,6 +69,94 @@ ShouldSkipBitsetTransform(const BitsetView& bitset,
...
+    if (enable_mmap) {
+        AssertInfo(!filepath.empty(), "offset mapping mmap filepath is empty");
+        AssertInfo(size <= std::numeric_limits<size_t>::max() / sizeof(int32_t),
+                   "offset mapping mmap buffer is too large");
+
+        const auto mmap_size = sizeof(int32_t) * size;
+        std::filesystem::create_directories(
+            std::filesystem::path(filepath).parent_path());
+
+        const int fd = open(filepath.c_str(), O_RDWR | O_CREAT | O_TRUNC, 0600);
```
foxspy, line 95: Suggestion: could these arrays reuse the singleton `MmapManager::GetInstance().GetMmapChunkManager()` instead of managing files directly with `open` / `ftruncate` / `mmap`? Registering one descriptor for the offset mapping and allocating both i2o/o2i arrays from it would keep the mmap root, disk-limit accounting, and release lifecycle consistent with the rest of segcore. The important part is to reuse the singleton manager rather than construct a per-index `MmapChunkManager`, since the latter was the source of the accounting-reset issue in the earlier revision. This is non-blocking if the direct file lifecycle is intentional.

internal/core/src/index/VectorIndexValidDataUtils.h
marcelo-cjl: I kept these as dedicated id-mapping load knobs intentionally. `ENABLE_MMAP` controls the index payload mmap path, while offset mapping has direction-specific behavior and can be enabled independently through `enable_mmap_i2o_map` / `enable_mmap_o2i_map`. Wiring `ENABLE_MMAP=true` to force both mappings would change that contract and remove the per-direction control, so I did not reuse `ENABLE_MMAP` for idmap mmap.

internal/core/src/common/OffsetMapping.cpp
marcelo-cjl: Done. I kept the direct file-backed lifecycle intentional so the offset mapping files stay under the index-local `id_mapping_mmap` directory alongside the DiskANN index files. For accounting, this now follows the sealed index load-resource path: when either idmap mmap knob is enabled, `IndexLoadResource` adds one conservative `num_rows * sizeof(int32_t)` estimate to `final_disk_cost` and `max_disk_cost`; enabling both directions still counts as one estimate and does not add memory cost. This avoids moving the files to the global mmap root while still charging the disk load resource.

foxspy (approved):
/lgtm

PR #51818, opened 2026-07-24
fix: synchronize delegator function runtime with collection schema

aoiasd:
issue: #49716

## Summary
Track the collection schema version applied to each shard delegator's FunctionRunner and IDF runtime. Catch up channel-local runtime when the shared Collection was advanced by another channel or load path, while serializing concurrent updates and preserving retry after failed updates.

### DIFF
--- a/internal/querycoordv2/observers/target_observer.go
+++ b/internal/querycoordv2/observers/target_observer.go
@@ -25,6 +25,7 @@ import (
 	"golang.org/x/time/rate"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/querycoordv2/meta"
 	"github.com/milvus-io/milvus/internal/querycoordv2/params"
 	"github.com/milvus-io/milvus/internal/querycoordv2/session"
@@ -526,7 +527,10 @@ func (ob *TargetObserver) shouldUpdateCurrentTarget(ctx context.Context, collect
 func (ob *TargetObserver) syncNextTargetToDelegator(ctx context.Context, collectionID int64, collReadyDelegatorList []*meta.DmChannel, newVersion int64) bool {
 	var partitions []int64
 	var indexInfo []*indexpb.IndexInfo
+	var schema *schemapb.CollectionSchema
+	var schemaBarrierTs uint64
 	var err error
+	metadataInitialized := false
 	for _, d := range collReadyDelegatorList {
 		updateVersionAction := ob.genSyncAction(ctx, d.View, newVersion)
 		replica := ob.meta.GetByCollectionAndNode(ctx, collectionID, d.Node)
@@ -536,7 +540,15 @@ func (ob *TargetObserver) syncNextTargetToDelegator(ctx context.Context, collect
 			return false
 		}
 		// init all the meta information
-		if partitions == nil {
+		if !metadataInitialized {
+			collectionInfo, err := ob.broker.DescribeCollection(ctx, collectionID)
+			if err != nil {
+				mlog.Warn(ctx, "failed to describe collection", mlog.Err(err))
+				return false
+			}
+			schema = collectionInfo.GetSchema()
+			schemaBarrierTs = collectionInfo.GetUpdateTimestamp()
+
 			partitions, err = utils.GetPartitions(ctx, ob.targetMgr, collectionID)
 			if err != nil {
 				mlog.Warn(ctx, "failed to get partitions", mlog.Err(err))
@@ -549,9 +561,10 @@ func (ob *TargetObserver) syncNextTargetToDelegator(ctx context.Context, collect
 				mlog.Warn(ctx, "fail to get index info of collection", mlog.Err(err))
 				return false
 			}
+			metadataInitialized = true
 		}
 
-		if !ob.syncToDelegator(ctx, replica, d.View, updateVersionAction, partitions, indexInfo) {
+		if !ob.syncToDelegator(ctx, replica, d.View, updateVersionAction, schema, schemaBarrierTs, partitions, indexInfo) {
 			return false
 		}
 	}
@@ -559,7 +572,7 @@ func (ob *TargetObserver) syncNextTargetToDelegator(ctx context.Context, collect
 }
 
 func (ob *TargetObserver) syncToDelegator(ctx context.Context, replica *meta.Replica, LeaderView *meta.LeaderView, action *querypb.SyncAction,
-	partitions []int64, indexInfo []*indexpb.IndexInfo,
+	schema *schemapb.CollectionSchema, schemaBarrierTs uint64, partitions []int64, indexInfo []*indexpb.IndexInfo,
 ) bool {
 	replicaID := replica.GetID()
 
@@ -577,11 +590,13 @@ func (ob *TargetObserver) syncToDelegator(ctx context.Context, replica *meta.Rep
 		ReplicaID:    replicaID,
 		Channel:      LeaderView.Channel,
 		Actions:      []*querypb.SyncAction{action},
+		Schema:       schema,
 		LoadMeta: &querypb.LoadMetaInfo{
-			LoadType:      ob.meta.GetLoadType(ctx, LeaderView.CollectionID),
-			CollectionID:  LeaderView.CollectionID,
-			PartitionIDs:  partitions,
-			ResourceGroup: replica.GetResourceGroup(),
+			LoadType:        ob.meta.GetLoadType(ctx, LeaderView.CollectionID),
+			CollectionID:    LeaderView.CollectionID,
+			PartitionIDs:    partitions,
+			ResourceGroup:   replica.GetResourceGroup(),
+			SchemaBarrierTs: schemaBarrierTs,
 		},
 		Version:       time.Now().UnixNano(),
 		IndexInfoList: indexInfo,
--- a/internal/querycoordv2/observers/target_observer.go
+++ b/internal/querycoordv2/observers/target_observer.go
@@ -525,12 +525,30 @@ func (ob *TargetObserver) shouldUpdateCurrentTarget(ctx context.Context, collect
 // 1. if next target is changed before delegator becomes serviceable, we need to sync the new next target to delegator to support partial search
 // 2. if next target is ready to read, we need to sync the next target to delegator to support full search
 func (ob *TargetObserver) syncNextTargetToDelegator(ctx context.Context, collectionID int64, collReadyDelegatorList []*meta.DmChannel, newVersion int64) bool {
-	var partitions []int64
-	var indexInfo []*indexpb.IndexInfo
-	var schema *schemapb.CollectionSchema
-	var schemaBarrierTs uint64
-	var err error
-	metadataInitialized := false
+	if len(collReadyDelegatorList) == 0 {
+		return true
+	}
+
+	collectionInfo, err := ob.broker.DescribeCollection(ctx, collectionID)
+	if err != nil {
+		mlog.Warn(ctx, "failed to describe collection", mlog.Err(err))
+		return false
+	}
+	schema := collectionInfo.GetSchema()
+	schemaBarrierTs := collectionInfo.GetUpdateTimestamp()
+
+	partitions, err := utils.GetPartitions(ctx, ob.targetMgr, collectionID)
+	if err != nil {
+		mlog.Warn(ctx, "failed to get partitions", mlog.Err(err))
+		return false
+	}
+
+	indexInfo, err := ob.broker.ListIndexes(ctx, collectionID)
+	if err != nil {
+		mlog.Warn(ctx, "fail to get index info of collection", mlog.Err(err))
+		return false
+	}
+
 	for _, d := range collReadyDelegatorList {
 		updateVersionAction := ob.genSyncAction(ctx, d.View, newVersion)
 		replica := ob.meta.GetByCollectionAndNode(ctx, collectionID, d.Node)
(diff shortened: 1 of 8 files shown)


internal/querynodev2/delegator/delegator.go
```diff
@@ -1543,17 +1580,15 @@ func NewShardDelegator(ctx context.Context, collectionID UniqueID, replicaID Uni
...
-	})
-
-	if hasBM25Field {
+	if len(sd.bm25Functions) > 0 {
 		idfOracle := NewIDFOracle(sd.vchannelName, collection.Schema().GetFunctions())
 		idfOracle.Start()
 		sd.distribution.SetIDFOracle(idfOracle)
 		sd.publishIDFOracle(idfOracle)
 	}
+	sd.collectionVersion.Store(collection.SchemaVersion())
```
MrPresent-Han: `collection` here is the shared per-collection object (`manager.Collection.Get(collectionID)` at L1523 is keyed by collectionID only, so every channel's delegator on this node points at the same instance). This constructor reads it four separate times:

- L1527 `Alloc(..., collection.Schema())` — builds the FunctionRunner
- L1583 `newBM25FunctionSet(collection.Schema())`
- L1586 `NewIDFOracle(..., collection.Schema().GetFunctions())`
- L1591 `collection.SchemaVersion()` — this new line

Nothing serializes this window against another channel: `PutOrRef` in `services.go` returns before the constructor runs, the collection's schema-transition lock is only held inside `applySchemaUpdate`, and `schemaChangeMutex` is per-delegator. So channel B's `delete_node → UpdateSchema → collectionManager.UpdateSchema` can advance the shared `Collection` in the middle of this.

If the advance lands between the `Schema()` reads and this `Store()`, the delegator ends up with `collectionVersion` = new version but runner / `bm25Functions` / IDF oracle = old schema. All three catch-up sites added by this PR compare with `<`:

- L1339 (no-op branch of `UpdateSchema`)
- L1399 (normal branch)
- `delegator_data.go:619` (`syncCollectionIndexMeta`)

so they all evaluate false and this channel never catches up — no self-heal until the delegator is rebuilt.

Not a regression (the multi-read predates this PR, and on master the no-op branch returned `nil` anyway), but it is a hole in this PR's own fix. Suggest taking one snapshot right after L1523 and reusing it:

```go
collection := manager.Collection.Get(collectionID)
if collection == nil {
    return nil, merr.WrapErrCollectionNotFound(collectionID, "not in delegator manager")
}
schema, schemaVersion, _ := collection.SchemaSnapshot()
```

then use `schema` at L1527/L1583/L1586 and `schemaVersion` here.

(FWIW calling `UpdateDelegatorSchema` from the constructor instead does not work: `collectionVersion` starts at 0 and the guard at L248 is `>=`, so a collection that was never altered — `schema.Version == 0` — would return early and never build the IDF oracle.)

internal/querynodev2/delegator/delegator.go
```diff
@@ -226,6 +233,53 @@ func (sd *shardDelegator) publishIDFOracle(idfOracle IDFOracle) {
...
+		return nil
+	}
+
+	newSet := newBM25FunctionSet(schema)
+	if newSet.HasIncompatibleCommonFunction(sd.bm25Functions) {
+		return merr.WrapErrServiceInternal("unsupported incompatible BM25 function schema change on loaded collection")
+	}
+	if err := function.GetManager().Update(sd.collectionID, delegatorFunctionRunnerKey(sd.vchannelName), schema); err != nil {
+		return err
+	}
```
MrPresent-Han, line 282: Minor / observability, not a correctness blocker.

Every error exit of `functionRunnerManager.Update` is an invariant violation rather than an environmental failure:

- `key == ""` / `schema == nil` — impossible from this call site
- unregistered key, or `version >= schema.GetVersion()` — returns `nil`, no-op
- `buildFunctionRunnerVersion` → `embeddingFunctionMetadata` — function has no input/output fields, or references a field ID absent from the schema (`manager.go:1040/1045/1060`)
- `ensureVersion` — the same `schema.Version` registered with two different function metadata sets (`manager.go:539/545`)

Note it only computes signatures and field-ID maps; actual runner construction is async, so no provider/credential/tokenizer failure can surface here. That matches the comment this PR removes ("A synchronous error here means the internal function metadata is invalid") and explains why the old code chose `panic`.

Returning an error instead is the better shape — it is testable, and the PR's own `TestUpdateSchemaReturnsInvalidFunctionMetadata` covers it. The concern is what happens downstream: the only caller of `UpdateSchema` is `pipeline/delete_node.go:98`, which just logs at Warn and moves on. The message is acked and never replayed.

By that point L1383 (`worker.UpdateSchema`) and L1394 (`collectionManager.UpdateSchema`) have already landed, so the node is left with workers + local `Collection` on the new schema while this delegator's FunctionRunner and IDF oracle stay on the old one. `collectionVersion` is correctly left un-stamped, so the state is *retryable* — but nothing drives a retry: all three catch-up sites (L1339, L1399, `delegator_data.go:619`) are passive and need either another schema DDL or another `LoadSegments` carrying `IndexInfoList`. If neither arrives the delegator stays stale indefinitely, with a single Warn line as the only evidence.

`panic` should be kept here. Every exit listed above is a corrupted-invariant case with no local recovery: by the time it fires the schema has already landed on the workers and on the local `Collection`, so this delegator cannot be brought back to a consistent state without being rebuilt. Crashing gives that for free — restart, re-watch, `NewShardDelegator` rebuilds runner + IDF oracle from the current schema. Returning an error hands it to a caller that cannot act on it, so "loud crash + automatic recovery" becomes "silent wrong BM25 results with one Warn line".

Testability is not an argument for the change either: the base already covered it with `require.PanicsWithError` in `TestUpdateSchemaPanicsOnInvalidFunctionMetadata`.

Suggest: keep `return err` for the BM25 incompatibility guard above (that is a rejection, not corruption), and restore `panic(err)` for `function.GetManager().Update`. If a softer failure mode is wanted instead, it needs to be an actual recovery path — mark the delegator for rebuild — not a Warn.


Worth noting the WAL side too: `delete_node.go` logs the Warn and then unconditionally calls `UpdateTSafe` at L109, so the schema event is consumed and TSafe advances regardless. Queries move forward on the new TSafe while this delegator's function runtime is still on the old schema.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1279,15 +1333,14 @@ func (sd *shardDelegator) UpdateSchema(ctx context.Context, schema *schemapb.Col
 			mlog.Uint64("schemaVersion", schemaVersion),
 			mlog.Uint64("schemaBarrierTs", schemaBarrierTs),
 		)
+		// Another channel may have advanced the shared Collection before this
+		// delegator applied its channel-local FunctionRunner and IDF state.
+		currentSchema, currentVersion, _ := sd.collection.SchemaSnapshot()
+		if sd.collectionVersion.Load() < currentVersion {
+			return sd.UpdateDelegatorSchema(ctx, currentSchema, currentVersion)
+		}
```
MrPresent-Han: This branch now advances the channel-local runtime but leaves `sd.schemaBarrierTs` alone, and the snapshot's barrier is discarded with `_`.

`addDistributionIfSchemaBarrierOK` (`delegator_data.go:788`) fences in-flight loads by comparing the load's own `req.GetLoadMeta().GetSchemaBarrierTs()` (passed at `delegator_data.go:754`) against the delegator-local `sd.schemaBarrierTs`. Reaching this branch means another channel already applied the DDL to the shared `Collection`, so this delegator never raised its own barrier for it — L1347-1348 is only on the path below.

Result: a load that started before the DDL still passes the fence and publishes its segments, while `UpdateDelegatorSchema` has already switched the FunctionRunner and IDF oracle to the new schema. For a newly added BM25 function that means segments with no stats for the new output field landing in the distribution after the oracle expects them.

On master this branch was a plain `return nil`, so the barrier and the runtime were stale together and stayed consistent. Advancing only the runtime is what makes them diverge.

`schemaChangeMutex` is already held here (L1328), so the fix is local — keep the third return value and raise the barrier before returning:

```go
currentSchema, currentVersion, currentBarrierTs := sd.collection.SchemaSnapshot()
if sd.schemaBarrierTs < currentBarrierTs {
    sd.schemaBarrierTs = currentBarrierTs
}
if sd.collectionVersion.Load() < currentVersion {
    return sd.UpdateDelegatorSchema(ctx, currentSchema, currentVersion)
}
return nil
```


Keep the `<` guard so the barrier only ever moves forward, consistent with L1347.

Two related spots worth a look while you are here:

- L1399 (the applied path) is fine — L1347 already raised the barrier before it.
- `delegator_data.go:619` also advances only the version. That one looks intentional, since a load is the thing being fenced rather than the thing that should move the fence — but it is not obvious from the code. A one-line comment saying why the load path deliberately does not touch `schemaBarrierTs` would help.

Test gap: the suite covers `collectionVersion` on every catch-up path but never `schemaBarrierTs`. Suggest a case that takes the no-op branch, asserts the barrier was raised to the snapshot value, and asserts `addDistributionIfSchemaBarrierOK` then rejects an entry carrying the pre-DDL barrier.

Longer term (not for this PR): this and the constructor snapshot issue are the same root — the delegator keeps two independent local shadows of one atomic `Collection` snapshot, advanced at different sites under different guards. Folding `{version, barrierTs}` into a single applied-snapshot struct that every catch-up path updates as a unit would make "only half advanced" structurally impossible.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -587,9 +587,9 @@ func (sd *shardDelegator) loadBM25StatsForReopen(ctx context.Context, infos []*q
...
 }
 
-// syncCollectionIndexMeta refreshes the delegator node's CCollection IndexMeta after a
-// forwarded worker load. Worker LoadSegments already updates IndexMeta on the target
-// worker, but the delegator (which executes growing search locally) must stay in sync.
+// syncCollectionIndexMeta refreshes the delegator node's CCollection IndexMeta and
+// channel-local function runtime after a forwarded worker load. Worker LoadSegments
+// already updates IndexMeta on the target worker, but the delegator must stay in sync.
 func (sd *shardDelegator) syncCollectionIndexMeta(ctx context.Context, req *querypb.LoadSegmentsRequest) error {
 	if len(req.GetIndexInfoList()) == 0 {
```
MrPresent-Han: Can we rename this method, not just indexMeta
This method actually serve as a complement for update delegator schema

MrPresent-Han:
related: https://github.com/milvus-io/milvus/issues/51062

liliu-z (commented):
/lgtm
/approve

internal/querycoordv2/observers/target_observer.go
```diff
@@ -577,11 +584,13 @@ func (ob *TargetObserver) syncToDelegator(ctx context.Context, replica *meta.Rep
 		ReplicaID:    replicaID,
 		Channel:      LeaderView.Channel,
 		Actions:      []*querypb.SyncAction{action},
+		Schema:       schema,
```
buqian-zilliz, line 587: **[P1] Apply schema metadata during target-version sync**

`genSyncAction` always creates a `SyncType_UpdateVersion` action, but QueryNode's `SyncDistribution` only reads `req.Schema` and `req.LoadMeta` in the `SyncType_Set` branch; the UpdateVersion branch only calls `SyncTargetVersion`. Consequently this RPC can succeed and let QueryCoord promote the target without updating the delegator's schema barrier, FunctionRunner, or IDF runtime, so an already-loaded channel that missed the schema event remains stale indefinitely. Apply or catch up the schema in the UpdateVersion path, or introduce a dedicated schema-sync action.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1341,36 +1421,16 @@ func (sd *shardDelegator) UpdateSchema(ctx context.Context, schema *schemapb.Col
...
-		sd.distribution.SetIDFOracle(idfOracle)
-		if current := sd.distribution.current.Load(); current != nil {
-			idfOracle.SetNext(current)
-		}
-		sd.publishIDFOracle(idfOracle)
-	} else if idfOracle != nil && !newSet.Equal(oldSet) {
-		if err := idfOracle.SyncFunctions(schema.GetFunctions()); err != nil {
-			return err
-		}
+	if err := sd.updateDelegatorSchemaLocked(ctx); err != nil {
```
buqian-zilliz, line 1424: **[P1] Validate BM25 compatibility before distributed side effects**

`updateDelegatorSchemaLocked` performs `HasIncompatibleCommonFunction`, but this call now happens only after every worker has accepted `UpdateSchema` and `collectionManager.UpdateSchema` has published the new shared schema. If the deterministic compatibility check rejects the change, the method returns into the new pipeline panic with workers and the shared Collection on the new schema while this delegator's FunctionRunner, IDF runtime, and collection version remain old. Run this validation before worker fan-out and collection publication, or make the transition transactional.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1328,8 +1399,17 @@ func (sd *shardDelegator) UpdateSchema(ctx context.Context, schema *schemapb.Col
 	}
 
 	_, err = executeSubTasks(ctx, tasks, nil, func(ctx context.Context, req *querypb.UpdateSchemaRequest, worker cluster.Worker) (*StatusWrapper, error) {
-		ctx = retry.WithMaxAttemptsContext(ctx, 1)
-		status, err := worker.UpdateSchema(ctx, req)
+		var status *commonpb.Status
+		err := retry.Do(ctx, func() error {
```
buqian-zilliz, line 1403: **[P1] Include worker acquisition in the retry boundary**

`organizeSubTask` resolves each worker before this closure and keeps `worker == nil` when `GetWorker` fails. `executeSubTasks` then takes its nil-worker branch without invoking `retry.Do`, so a transient session lookup, missing address, or client-construction failure reaches `deleteNode`'s new panic with zero retries. Re-resolve the worker inside each retry attempt, or retry task construction before failing the schema event.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -567,8 +573,17 @@ func (sd *shardDelegator) loadBM25StatsForReopen(ctx context.Context, infos []*q
...
-			if err := idfOracle.LoadSealedForReopen(ctx, info.GetSegmentID(), info, cm, activateIfReadable); err != nil {
+			var activateIfReadable bool
+			err := retry.Do(ctx, func() error {
+				activateIfReadable = sd.distribution.IsReadableSealedSegment(info.GetSegmentID())
+				return idfOracle.LoadSealedForReopen(ctx, info.GetSegmentID(), info, cm, activateIfReadable)
+			},
+				retry.Attempts(reopenBM25LoadRetryCount+1),
+				retry.Sleep(reopenBM25LoadRetryInitialBackoff),
+				retry.MaxSleepTime(reopenBM25LoadRetryMaxBackoff),
+				retry.RetryErr(merr.IsRetryableErr),
```
buqian-zilliz, line 584: **[P2] Retry transient streaming I/O failures**

`merr.IsRetryableErr` only accepts errors carrying a retriable Milvus error, while `LoadSealedForReopen` can propagate raw lazy object-reader, `io.Copy`, `os.Create`, and `f.Sync` errors; `merr.Wrapf` preserves those raw types. Transient mid-stream MinIO, GCS, Azure, or filesystem failures therefore stop after the first attempt even though this block is meant to retry them. Map storage errors to typed retryable errors before this predicate, or reject only known permanent errors.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1328,8 +1399,17 @@ func (sd *shardDelegator) UpdateSchema(ctx context.Context, schema *schemapb.Col
...
+			rpcCtx := retry.WithMaxAttemptsContext(ctx, 1)
+			var err error
+			status, err = worker.UpdateSchema(rpcCtx, req)
+			return merr.CheckRPCCall(status, err)
+		},
+			retry.Attempts(updateSchemaWorkerRetryCount+1),
```
buqian-zilliz, line 1409: **[P2] Do not sleep after the terminal retry**

`retry.Do` sleeps after every failed attempt, including the last configured one. With six schema attempts, the final RPC runs at about 15.5 seconds and this code then holds `schemaChangeMutex` and blocks tSafe for another 16 seconds before returning and panicking, without issuing another RPC; the BM25 path similarly wastes 6.4 seconds. Use retry logic that omits the terminal sleep or schedules an actual attempt after each backoff; the one-minute cap is unreachable with these attempt counts.

Issue #46358, opened 2025-12-16
[Feature]: Support reopen segment on querynodes


congqixia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Support reopen segments on querynodes applying data or schema change

### Describe the solution you'd like.

Impl segment reopening by calculating load diff in segcore based on segment managed loading procedure

PR #51822, opened 2026-07-24
enhance: [3.0] remove legacy query node drop index path (#51740)

congqixia:
Cherry-pick from master
pr: #51740
Related to #46358

Remove the unused QueryCoord drop-index task and QueryNode RPC, delegator, worker, segment, and segcore entry points.

Keep index removal routed through Reopen and LoadDiff, and regenerate the affected proto and mocks.

---------


liliu-z (commented):
/lgtm

PR #51823, opened 2026-07-24
enhance: configurable backoff for import write retry

bigsheeper:
## Summary

Make the import write-retry backoff (initial interval + max cap) configurable and larger by default, while keeping the existing unlimited-retry behavior unchanged. Under sustained object-storage request-rate throttling — e.g. AWS S3 `503 SlowDown` during a large bulk import writing many small objects to a cold bucket prefix — the tight fixed `10s` cap makes the whole-file write retry re-fire too aggressively; larger, tunable backoff reduces that pressure without a rebuild.

issue: #51824

### Changes

- Add two refreshable paramtable items:
  - `dataNode.import.writeRetryInitialInterval` (default `1s`)
  - `dataNode.import.writeRetryMaxInterval` (default `60s`)
- Wire them into the import write retry options in `internal/datanode/importv2/util.go` via `retry.Sleep` / `retry.MaxSleepTime`. `retry.Attempts` is unchanged (`dataNode.import.maxWriteRetryAttempts`, default `0` = unlimited).

### Test Plan

- [x] `pkg/util/paramtable` unit tests pass (new param defaults asserted)
- [x] `internal/datanode/importv2` unit tests pass
- [ ] CI: static-check + full test suite

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -907,6 +907,8 @@ dataNode:
     readBufferSizeInMB: 16 # The base insert buffer size (in MB) during import. The actual buffer size will be dynamically calculated based on the number of shards.
     readDeleteBufferSizeInMB: 16 # The delete buffer size (in MB) during import.
     memoryLimitPercentage: 10 # The percentage of memory limit for import/pre-import tasks.
+    writeRetryInitialInterval: 1 # Initial backoff interval in seconds for import write retry.
+    writeRetryMaxInterval: 60 # Maximum backoff interval in seconds for import write retry.
   compaction:
     levelZeroBatchMemoryRatio: 0.5 # The minimal memory ratio of free memory for level zero compaction executing in batch mode
     levelZeroMaxBatchSize: -1 # Max batch size refers to the max number of L1/L2 segments in a batch when executing L0 compaction. Default to -1, any value that is less than 1 means no limit. Valid range: >= 1.
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -907,8 +907,13 @@ dataNode:
     readBufferSizeInMB: 16 # The base insert buffer size (in MB) during import. The actual buffer size will be dynamically calculated based on the number of shards.
     readDeleteBufferSizeInMB: 16 # The delete buffer size (in MB) during import.
     memoryLimitPercentage: 10 # The percentage of memory limit for import/pre-import tasks.
-    writeRetryInitialInterval: 1 # Initial backoff interval in seconds for import write retry.
-    writeRetryMaxInterval: 60 # Maximum backoff interval in seconds for import write retry.
+    # Initial backoff interval in seconds for import write retry. Must be a positive integer;
+    # a non-positive or unparseable value falls back to the default.
+    writeRetryInitialInterval: 1
+    # Maximum backoff interval in seconds for import write retry. Must be a positive integer;
+    # a non-positive or unparseable value falls back to the default. Set it to at least twice
+    # writeRetryInitialInterval, otherwise the effective cap is raised to twice the initial interval.
+    writeRetryMaxInterval: 60
   compaction:
     levelZeroBatchMemoryRatio: 0.5 # The minimal memory ratio of free memory for level zero compaction executing in batch mode
     levelZeroMaxBatchSize: -1 # Max batch size refers to the max number of L1/L2 segments in a batch when executing L0 compaction. Default to -1, any value that is less than 1 means no limit. Valid range: >= 1.
--- a/internal/datanode/importv2/util.go
+++ b/internal/datanode/importv2/util.go
@@ -107,9 +107,12 @@ func NewSyncTask(ctx context.Context,
 	}
 
 	writeRetryAttempts := paramtable.Get().DataNodeCfg.ImportMaxWriteRetryAttempts.GetAsUint()
+	initialInterval := time.Duration(paramtable.Get().DataNodeCfg.ImportWriteRetryInitialInterval.GetAsInt()) * time.Second
+	maxInterval := time.Duration(paramtable.Get().DataNodeCfg.ImportWriteRetryMaxInterval.GetAsInt()) * time.Second
 	retryOpts := []retry.Option{
-		retry.Attempts(writeRetryAttempts), // default retry always
-		retry.MaxSleepTime(10 * time.Second),
+		retry.Attempts(writeRetryAttempts), // 0 = unlimited, preserved on purpose
+		retry.Sleep(initialInterval),
+		retry.MaxSleepTime(maxInterval),
 	}
 	task := syncmgr.NewSyncTask().
 		WithAllocator(allocator).
--- a/internal/datanode/importv2/util.go
+++ b/internal/datanode/importv2/util.go
@@ -106,22 +106,29 @@ func NewSyncTask(ctx context.Context,
 		syncPack.WithBM25Stats(bm25Stats)
 	}
 
-	writeRetryAttempts := paramtable.Get().DataNodeCfg.ImportMaxWriteRetryAttempts.GetAsUint()
-	initialInterval := time.Duration(paramtable.Get().DataNodeCfg.ImportWriteRetryInitialInterval.GetAsInt()) * time.Second
-	maxInterval := time.Duration(paramtable.Get().DataNodeCfg.ImportWriteRetryMaxInterval.GetAsInt()) * time.Second
-	retryOpts := []retry.Option{
-		retry.Attempts(writeRetryAttempts), // 0 = unlimited, preserved on purpose
-		retry.Sleep(initialInterval),
-		retry.MaxSleepTime(maxInterval),
-	}
 	task := syncmgr.NewSyncTask().
 		WithAllocator(allocator).
 		WithMetaCache(metaCache).
 		WithSchema(metaCache.GetSchema(0)). // TODO specify import schema if needed
 		WithSyncPack(syncPack).
 		WithStorageConfig(storageConfig).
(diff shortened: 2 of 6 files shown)


bigsheeper:
rerun ut

czs007:
Adversarial review of the import write-retry configuration found one blocking correctness issue and one low-severity documentation/validation gap; the default 1s→60s path itself is wired correctly.

## High

**`internal/datanode/importv2/util.go:22` — no lower bound on `writeRetryInitialInterval` allows a zero-delay unbounded retry loop**

`writeRetryInitialInterval` is read with `GetAsInt()` and passed straight to `retry.Sleep()` with no floor. An explicit `0`, or any unparseable value such as `1s` / `500ms` / `0.5` (the key has no `InSeconds` suffix, so a duration-style value is an easy operator mistake, and `getAndConvert` silently coerces parse failures to `0` rather than the declared default — `param_item.go:436-438,469-475`), yields `initialInterval == 0`. Options apply in order `Attempts(0) → Sleep(0) → MaxSleepTime(60s)`: in `options.go:78-86`, `c.sleep*2 == 0` is not `> 3s`, so the clamp does not fire, and the final config is `sleep=0, maxSleepTime=60s, attempts=0`. In `retry.Do`, `time.After(0)` fires immediately and `c.sleep *= 2` stays `0` forever (`retry.go:113,123-126`), while the loop condition `c.attempts == 0 || i < c.attempts` (`retry.go:51`) never terminates. The import task ctx is `context.WithCancel(context.Background())` (`task_import.go:63`), so the deadline escape at `retry.go:95-108` never triggers either. Negative values behave identically. This surface is new: pre-PR, `c.sleep` came from the 200ms default and was not operator-settable.

Failure: an operator sets `dataNode.import.writeRetryInitialInterval: 0` (or a duration-looking value) while `maxWriteRetryAttempts` keeps its default of `0` = unlimited, then runs a bulk import against object storage returning persistent `503 SlowDown` → the write retry loop spins with zero backoff and no attempt cap (measured in-tree: 2000 iterations in ~11ms with these exact options), each iteration issuing a real object-storage write, so the datanode hammers S3/MinIO indefinitely and amplifies the very throttling storm this PR is meant to relieve, with no path to failing out.

Suggestion: clamp to a positive minimum at the call site (e.g. `max(initialInterval, 100*time.Millisecond)`), or reject non-positive values at `Init` with a `Formatter` — the sibling `ImportMemoryLimitPercentage` (`component_param.go:7332-7348`) is in-repo precedent for exactly this validation. A behavioral test asserting the effective sleep schedule for both the defaults and a `0` config would lock the fix in; the diff currently adds no test that `NewSyncTask` threads these values through in the order-sensitive way it does. *(raised by tinswzy, chyezh)*

## Low

**`internal/datanode/importv2/util.go:23` — configured max interval is not a hard cap when `max ≤ 2×initial`**

`MaxSleepTime` forces `maxSleepTime = 2*c.sleep` whenever `c.sleep*2 > maxSleepTime` (`options.go:88-97`), and since `retry.Sleep(initialInterval)` is applied first, its own clamp has already raised `maxSleepTime` to `2×initial` before `MaxSleepTime` is consulted.

Failure: an operator sets `writeRetryInitialInterval: 60` with `writeRetryMaxInterval: 60` (or any pair where `max ≤ 2×initial`) to bound retry pressure → the schedule becomes 60s, 120s, 120s, … so the effective cap is silently 120s instead of the configured 60s, with no warning in logs or config output.

This is harmless at the shipped 1s/60s defaults, but the yaml comment ("Maximum backoff interval") is misleading for custom values. Suggestion: note in the config comment that `max` should be `≥ 2×initial`, or enforce that relationship at Init. *(raised by tinswzy, chyezh)*

bigsheeper:
Thanks — both findings verified against the current head (`666df5d1e`), and the High one is fixed in this PR (`b5c4853c3c`).

**High (`util.go:22`) — confirmed, fixed.** Traced every step:

- `Sleep(0)` is not clamped: `c.sleep*2 == 0 > maxSleepTime` is false (`options.go:76-83`), and `MaxSleepTime(60s)` then sets `maxSleepTime=60s` without touching `sleep`. Effective config is `sleep=0, maxSleepTime=60s, attempts=0`.
- `c.sleep *= 2` is a no-op at 0 (`retry.go:123-126`), and the loop condition `c.attempts == 0 || i < c.attempts` (`retry.go:51`; same in `Handle`, `:107`) never terminates.
- The deadline escape at `retry.go:95-108` cannot fire: the import task ctx is `context.WithCancel(context.Background())` (`task_import.go:63`) and is threaded unchanged to the sync call (`:306,312`) — no `WithTimeout`/`WithDeadline` anywhere on that path.
- Your point about `getAndConvert` is the sharper half: `getAsInt` coerces a parse failure to the literal `0`, **not** the declared `DefaultValue` (`param_item.go:437`, `:469-475`), so `writeRetryInitialInterval: 1s` — a very natural thing to write given the key has no `InSeconds` suffix — lands in exactly the same place as an explicit `0`. Negative values behave identically.
- Measured it: with these exact options a failing write ran 6710 iterations inside one 100ms window.

Fixed with a `Formatter` on both intervals rather than a call-site clamp, following the `ImportMemoryLimitPercentage` precedent you pointed at. A non-positive or unparseable value falls back to its declared default (1s / 60s) with a warning, so the configured value and the effective value stay identical — a call-site clamp would have left them diverging. I applied it to `writeRetryMaxInterval` too, since a `0` there disables the backoff cap the same way.

Tests added, as suggested: `component_param_test.go` pins the fallback for `0` / `-1` / `1s` and the pass-through for valid values; `internal/datanode/importv2/util_test.go` asserts the end-to-end schedule against an import-shaped ctx (cancellable, no deadline). Both fail without the formatter — the importv2 one at 6710 calls.

**Low (`util.go:23`) — confirmed, documented rather than enforced.** Reproduced: `Sleep` runs first and raises `maxSleepTime` to `2*initial`, so `initial=60, max=60` yields an effective cap of 120s. The shipped 1s/60s defaults are unaffected. I documented the `max >= 2*initial` expectation in the config doc and `milvus.yaml`, and pinned the semantics in `pkg/util/retry/options_test.go`, rather than enforcing the relationship — a `Formatter` only sees its own key, so cross-parameter validation would need a separate mechanism, which seems disproportionate for a case that cannot bite at the defaults. Happy to add it if you'd rather fail closed.

czs007:

Both reviewers independently re-examined the changes in the final round and confirmed they had no additional findings to raise.

bigsheeper:
/hold

let me test the import throughput before merge

bigsheeper:
**Validation results**

Ran a 3TB bulk import on a UAT cluster carrying this change, compared against the production incident that motivated it.

| | Incident (200ms / 10s) | This change (1s / 60s) | Incident ÷ this change |
|---|---|---|---|
| Data-landing phase | 9h18m | 51m | **10.9×** |
| S3 throttling window | 160 min | 3 min | **53×** |
| Throttling persisting after submission tailed off | 136 min | not observed | — |

The third row is the one that matters. In the incident, job submission thinned out ~2.5h before the storm ended, yet `SlowDown` kept firing for another 136 minutes — the retries were sustaining the congestion on their own. With the larger backoff, throttling appeared 12 minutes *after* submission had already stopped, and cleared in 3.

Same worker cap (50) and same collection shape (12 shards, 14 fields, 11 indexes) on both runs; different clusters (prod vs UAT).

czs007:

Verified:
- pkg/util/paramtable/component_param.go:7380 — both new formatters reset zero, negative, and unparseable values to the defaults (1/60), so the previously reported zero-delay retry loop cannot be re-entered through normal misconfiguration.
- pkg/util/retry/options.go:78 — the Attempts → Sleep → MaxSleepTime option order produces the intended backoff schedule (1s start, doubling, capped at 60s), matching the documented behavior in configs/milvus.yaml:910-916.
- internal/datanode/importv2/util.go:131 — `ImportMaxWriteRetryAttempts` is passed through unchanged, so the default infinite-retry semantics (0 = unlimited) are preserved; only the backoff intervals change, as the PR states.
- pkg/util/retry/retry.go:112 — the backoff wait selects on `ctx.Done()`, so the 60s cap does not delay an already-delivered cancellation.
- pkg/config/manager.go:141 — config changes evict the cached values, so formatters apply to every source including runtime overrides; stale unformatted values cannot bypass validation.
- internal/flushcommon/syncmgr/write_buffer.go:1499 — the flush-path `WithWriteRetryOptions` call sites keep their own hard-coded values, confirming this PR's scope is import-only.

## Re-review of `b5c4853c3caf`

No earlier findings from this mechanism were left open on this commit — there is nothing to adjudicate. (The fix for the previously reported zero-delay retry loop was independently re-verified this round; see the Verified list above.)

## Low — non-blocking

These do not require changes before merge.

- **internal/datanode/importv2/util.go:125** — `time.Duration(params.ImportWriteRetryInitialInterval.GetAsInt()) * time.Second` overflows int64 when the configured value exceeds 9,223,372,036 seconds, producing a negative sleep. If `dataNode.import.writeRetryInitialInterval` is set to such a value (e.g. a nanosecond count like 60000000000 entered by mistake) while an import write keeps failing and `maxWriteRetryAttempts` stays at its default of 0 → the negative sleep makes `time.After` fire immediately and `retry.Do` degrades into a zero-delay infinite hot loop (one core saturated, warn-log flooding) until the context is cancelled. The formatter only rejects `interval <= 0` and sets no upper bound, and the doubling logic keeps the sleep negative so it never recovers. Suggestion: clamp the value to a sane upper bound (e.g. 3600) in the formatter or in `newWriteRetryOptions`. (raised by czs007)

czs007 (approved):
/lgtm

Issue #51710, opened 2026-07-22
[Bug]: Concurrent SegmentGrowingImpl::Reopen and bulk_subscript cause schema use-after-free


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

No issue matching this exact ASan use-after-free signature was found. Related but different issues include #51662, #51594, and #50799.

### Environment

- Milvus version: `master-20260721-c532b95dba`
- Git commit: `c532b95dbac1d6ca1093bf8377c7965e63b23e2c`
  - The full commit is from the Jenkins `COMMIT_SHA` parameter.
  - The affected deployment reported `git_commit: c532b95dba` through `get_server_version(detail=True)`.
- Deployment mode: Standalone
- Message queue: Woodpecker
- SDK: PyMilvus `3.1.0rc69`
- OS: Linux/amd64 container
- CPU/Memory:
  - Requests: 3 CPU / 8 GiB
  - Limits: 8 CPU / 16 GiB
- GPU: None
- Kubernetes namespace: `jenkins-milvus-ci`
- Kubernetes node: `4am-node34`
- Image: `harbor.milvus.io/milvusdb/milvus:e2e-pr51674-c532b95d-260721-amd`
- Helm chart: `zilliztech/milvus` version `5.0.6`

### Kubernetes Pod List

At deployment startup:

```text
NAME                                                                  READY   STATUS    RESTARTS
e2e-disp-stdalone-g1-35556-etcd-0                                     1/1     Running   0
e2e-disp-stdalone-g1-35556-milvus-standalone-7cdc944756-99tzq         1/1     Running   0
e2e-disp-stdalone-g1-35556-minio-549c55b657-5l5qh                     1/1     Running   0
```

After the failure:

```text
NAME                                                                  READY   STATUS    RESTARTS
e2e-disp-stdalone-g1-35556-etcd-0                                     1/1     Running   0
e2e-disp-stdalone-g1-35556-milvus-standalone-7cdc944756-99tzq         1/1     Running   1
e2e-disp-stdalone-g1-35556-minio-549c55b657-5l5qh                     1/1     Running   0
```

The previous state of the Milvus container was `Error`.

### Current Behavior

The standalone Milvus process is terminated by AddressSanitizer while a growing-segment flush reads field metadata concurrently with a schema reopen.

The read path is:

```text
FlushGrowingSegmentData
  -> BuildArrayForChunk
  -> BuildVectorArrayFromSegment
[...]

PR #51826, opened 2026-07-24
fix:fix growing segment schema snapshot lifetime

zhagnlu:
issue: #51710

### DIFF
--- a/internal/core/src/exec/expression/MatchExpr.cpp
+++ b/internal/core/src/exec/expression/MatchExpr.cpp
@@ -247,9 +247,9 @@ PhyMatchFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
     auto input = context.get_offset_input();
     SetHasOffsetInput(input != nullptr);
 
-    auto schema = segment_->get_schema();
+    auto schema = segment_->get_schema_snapshot();
     auto field_meta =
-        schema.GetFirstArrayFieldInStruct(expr_->get_struct_name());
+        schema->GetFirstArrayFieldInStruct(expr_->get_struct_name());
 
     auto array_offsets = segment_->GetArrayOffsets(field_meta.get_id());
     AssertInfo(array_offsets != nullptr, "Array offsets not available");
--- a/internal/core/src/exec/operator/FilterBitsNode.cpp
+++ b/internal/core/src/exec/operator/FilterBitsNode.cpp
@@ -46,7 +46,7 @@ BuildExprCacheKey(const plan::FilterBitsNode& filter,
     auto* segment =
         query_context != nullptr ? query_context->get_segment() : nullptr;
     if (segment != nullptr &&
-        segment->get_schema().get_ttl_field_id().has_value()) {
+        segment->get_schema_snapshot()->get_ttl_field_id().has_value()) {
         key += fmt::format("|entity_ttl_physical_time_us:{}",
                            query_context->get_entity_ttl_physical_time_us());
     }
--- a/internal/core/src/query/SearchOnGrowing.cpp
+++ b/internal/core/src/query/SearchOnGrowing.cpp
@@ -62,11 +62,11 @@ FloatSegmentIndexSearch(const segcore::SegmentGrowingImpl& segment,
                         const BitsetView& bitset,
                         milvus::OpContext* op_context,
                         SearchResult& search_result) {
-    auto& schema = segment.get_schema();
+    auto schema = segment.get_schema_snapshot();
     auto& indexing_record = segment.get_indexing_record();
 
     auto vecfield_id = info.field_id_;
-    auto& field = schema[vecfield_id];
+    auto& field = (*schema)[vecfield_id];
     auto is_sparse = field.get_data_type() == DataType::VECTOR_SPARSE_U32_F32;
     // TODO(SPARSE): see todo in PlanImpl.h::PlaceHolder.
     auto dim = is_sparse ? 0 : field.get_dim();
@@ -107,13 +107,13 @@ SearchOnGrowing(const segcore::SegmentGrowingImpl& segment,
                 const BitsetView& bitset,
                 milvus::OpContext* op_context,
                 SearchResult& search_result) {
-    auto& schema = segment.get_schema();
+    auto schema = segment.get_schema_snapshot();
     auto& record = segment.get_insert_record();
 
     // step 1.1: get meta
     // step 1.2: get which vector field to search
     auto vecfield_id = info.field_id_;
-    auto& field = schema[vecfield_id];
+    auto& field = (*schema)[vecfield_id];
     CheckBruteForceSearchParam(field, info);
 
     auto data_type = field.get_data_type();
--- a/internal/core/src/segcore/SegmentGrowingImpl.cpp
+++ b/internal/core/src/segcore/SegmentGrowingImpl.cpp
@@ -350,11 +350,12 @@ SchemaHasTextField(const Schema& schema) {
 void
 SegmentGrowingImpl::InitializeArrayOffsets() {
     std::unique_lock lock(array_offsets_map_mutex_);
+    auto schema = get_schema_snapshot();
 
     // Group fields by struct_name
     std::unordered_map<std::string, std::vector<FieldId>> struct_fields;
 
-    for (const auto& [field_id, field_meta] : schema_->get_fields()) {
+    for (const auto& [field_id, field_meta] : schema->get_fields()) {
         auto struct_name = GetStructNameForArrayField(field_meta);
         if (struct_name.has_value()) {
             struct_fields[*struct_name].push_back(field_id);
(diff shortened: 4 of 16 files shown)


internal/core/src/segcore/SegmentGrowingImpl.h
```diff
@@ -218,6 +218,12 @@ class SegmentGrowingImpl : public SegmentGrowing {
         return *schema_;
     }
 
+    SchemaPtr
+    get_schema_snapshot() const override {
+        std::shared_lock lck(sch_mutex_);
```
buqian-zilliz: **[P1] Avoid reversing the schema and segment lock order**

`get_schema_snapshot()` acquires `sch_mutex_`, but existing query and retrieve paths acquire `mutex_` first and then reach schema-snapshotting helpers. `Reopen()` takes the opposite order: `sch_mutex_` first, then `mutex_` while publishing text indexes and in `UpdateResourceTracking()`. A concurrent query and schema reopen can therefore wait on each other indefinitely. Publish and load `schema_` atomically, or enforce one lock order and pass one operation-level snapshot through the call chain.

internal/core/src/segcore/SegmentInterface.cpp
```diff
@@ -748,8 +748,8 @@ SegmentInternalInterface::set_field_avg_size(FieldId field_id,
                                              int64_t field_size) {
     AssertInfo(field_id.get() >= 0,
                "invalid field id, should be greater than or equal to 0");
-    auto& schema = get_schema();
-    auto& field_meta = schema[field_id];
+    auto schema = get_schema_snapshot();
```
buqian-zilliz, line 751: **[P1] Do not reacquire the schema lock during Insert**

`Insert()` holds `sch_mutex_` in shared mode for the entire insert and calls this method for every variable-width field. This new snapshot call reacquires the same non-recursive `std::shared_mutex`, which is undefined and can self-deadlock when a writer is pending. Pass the already-held `Schema` or `FieldMeta` into the update, as the new overloads for `try_remove_chunks()` and `UpdateResourceTracking()` do.

internal/core/src/query/SearchOnGrowing.cpp
```diff
@@ -62,11 +62,11 @@ FloatSegmentIndexSearch(const segcore::SegmentGrowingImpl& segment,
                         const BitsetView& bitset,
                         milvus::OpContext* op_context,
                         SearchResult& search_result) {
-    auto& schema = segment.get_schema();
+    auto schema = segment.get_schema_snapshot();
```
buqian-zilliz, line 65: **[P2] Keep growing index search on one schema snapshot**

`SearchOnGrowing()` already holds a `SchemaPtr` and `FieldMeta` from its snapshot, but this helper takes a second snapshot. A concurrent `Reopen()` can drop `vecfield_id` between the two reads, making the lookup below assert even though the in-flight search is still using metadata from the first snapshot; only the indexed path has this failure. Pass the caller's snapshot or `FieldMeta` into `FloatSegmentIndexSearch()` so one search observes one schema version.

zhagnlu: The existing behavior remains the same; no other changes will be made to this PR for the time being.

internal/core/src/segcore/SegmentInterface.cpp
zhagnlu: fixed

internal/core/src/segcore/SegmentGrowingImpl.h
zhagnlu: using atomic replaced

internal/core/src/segcore/SegmentGrowingImpl.cpp
```diff
@@ -1640,7 +1671,8 @@ SegmentGrowingImpl::bulk_subscript(milvus::OpContext* op_ctx,
                                    FieldId field_id,
                                    const int64_t* seg_offsets,
                                    int64_t count) const {
-    auto& field_meta = schema_->operator[](field_id);
+    auto schema = get_schema_snapshot();
```
buqian-zilliz, line 1674: **[P1] Keep vector flushes on the flush-task schema**

`FlushGrowingSegmentData` intentionally uses the flush-task schema because it can be older than the runtime schema, and its materialized-field projection can still include a vector field later dropped from runtime. Fixed-width and sparse vector extraction then calls this overload, which reloads the current runtime schema; after `Reopen()` drops the field, the lookup on the next line throws even though the flush-task `FieldMeta` and raw data are valid. Pass the flush-task `FieldMeta` or schema through vector extraction so the entire flush uses one schema version.

internal/core/src/segcore/SegmentGrowingImpl.cpp
```diff
@@ -1267,7 +1292,7 @@ SegmentGrowingImpl::Delete(int64_t size,
     deleted_record_.StreamPush(sort_pks, sort_timestamps.data());
 
     // step 3: update resource tracking
-    UpdateResourceTracking();
+    UpdateResourceTracking(*schema);
```
buqian-zilliz: **[P1] Prevent Delete from replacing a newer resource charge**

`schema` is captured before parsing and sorting the delete, but `Delete()` does not hold `sch_mutex_`. A concurrent `Reopen()` can publish a newer schema, backfill a new field, and charge that larger resource estimate; this delete can then refund that charge and replace `tracked_resource_` with an estimate from the old schema, omitting the new field. Because the estimate is computed before `resource_tracking_mutex_` is acquired, that mutex only makes refund/charge atomic, not fresh. Serialize this operation with `Reopen()`, or compute and version-check the latest schema while serializing the estimate and update.

internal/core/src/segcore/SegmentGrowingImpl.cpp
zhagnlu: This is a valid issue, but I think it is out of scope for this PR.
This PR is focused on schema snapshot lifetime/stability: making growing readers hold a stable SchemaPtr instead of referencing schema_ that can be replaced by Reopen().
I’d prefer to handle that in a follow-up PR to avoid mixing the snapshot-lifetime fix with a growing-source flush behavior change.

internal/core/src/segcore/SegmentGrowingImpl.cpp
zhagnlu: fixed

internal/core/src/exec/operator/FilterBitsNode.cpp
```diff
@@ -46,7 +46,7 @@ BuildExprCacheKey(const plan::FilterBitsNode& filter,
     auto* segment =
         query_context != nullptr ? query_context->get_segment() : nullptr;
     if (segment != nullptr &&
-        segment->get_schema().get_ttl_field_id().has_value()) {
+        segment->get_schema_snapshot()->get_ttl_field_id().has_value()) {
```
buqian-zilliz, line 49: **[P2] Include the TTL field identity in the expression cache key**

`CompileExpressions()` injects the TTL predicate from the segment schema, but this cache key records only the query's physical time when a TTL field exists; the injected field ID and schema generation are absent from `filter.ToString()`. A properties-only schema update can switch `ttl_field` from field A to field B without changing the segment ID or request timestamp, and segment reopen does not erase `ExprResCacheManager`. Stage 2 can therefore reuse Stage 1's bitmap computed for the old TTL field. Include the schema generation or TTL field ID in the key, or invalidate the segment cache when the schema is published.

internal/core/src/exec/expression/MatchExpr.cpp
```diff
@@ -247,9 +247,9 @@ PhyMatchFilterExpr::Eval(EvalCtx& context, VectorPtr& result) {
     auto input = context.get_offset_input();
     SetHasOffsetInput(input != nullptr);
 
-    auto schema = segment_->get_schema();
+    auto schema = segment_->get_schema_snapshot();
```
buqian-zilliz, line 250: **[P2] Keep match evaluation on one schema snapshot**

`PhyMatchFilterExpr` is reused across filter batches, but every `Eval()` resolves the struct from the latest segment schema. Growing `Reopen()` can publish a schema that drops the struct while this query is still running; the old field data and array offsets are retained, yet the next batch throws in `GetFirstArrayFieldInStruct()`, and the validity helpers independently look up the old field in the new schema. Retain a query-scoped schema or representative `FieldMeta` and use it for every batch and validity lookup.

internal/core/src/segcore/SegmentGrowingImpl.cpp
buqian-zilliz: This issue still exists: recapturing the schema in `UpdateResourceTracking()` does not serialize the estimate with `Reopen()`. The estimate is still computed before `resource_tracking_mutex_` is acquired, so `Reopen()` can publish and charge a newer schema before `Delete()` replaces that charge with its stale estimate.

internal/core/src/exec/operator/FilterBitsNode.cpp
zhagnlu: This is another problem,  need to fix other prs, current pr not fix it

internal/core/src/exec/expression/MatchExpr.cpp
zhagnlu: same as upper comment, need to fixed in other prs, not releated with this pr point

internal/core/src/segcore/SegmentGrowingImpl.h
```diff
@@ -215,7 +215,16 @@ class SegmentGrowingImpl : public SegmentGrowing {
 
     const Schema&
     get_schema() const override {
```
zhengbuqian, line 217: can we just migrate all get_schema to get_schema_snapshot?

zhengbuqian (approved):

PR #51828, opened 2026-07-24
fix: [3.0] size storage v2 cells by projected columns

sparknack:
pr: #51653
issue: #51652

## What changed
- Backport projected-column size estimation for Storage V2 cells to 3.0.
- Preserve a positive loading reservation for fully deleted row groups.
- Bump milvus-storage to f9f5bd1 for per-column decoded-memory estimates.

## Test
- `make build-cpp-with-unittest`
- `./internal/core/output/unittest/all_tests --gtest_filter='*ManifestGroupTranslatorTest*'` (30/30 passed)
- `make verifiers`


internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +132,207 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+            ? static_cast<uint64_t>(fallback_bytes_per_row)
+            : kLastResortBytesPerRow;
+    size_t live_row_groups = 0;
+    size_t positive_fallback_row_groups = 0;
+    for (size_t i = 0; i < row_group_sizes.size(); ++i) {
+        if (row_group_rows[i] == 0) {
+            // Empty logical row groups still participate in cell ranges. Keep
+            // their loading reservation positive so the batch planner can
+            // reach the reader and receive the valid empty batch.
+            row_group_sizes[i] =
```
buqian-zilliz, line 292: **[P1] Zero-row Iceberg chunks can crash merged reads**

The positive sentinel makes fully deleted Iceberg row groups eligible for the pinned f9f5bd1 batch reader. When a contiguous live row group is followed by a zero-row group, the reader merges both into one range, consumes the sole live RecordBatch, increments its batch index, and then dereferences past the vector while reconstructing the zero-row chunk. Loading this valid layout can therefore crash QueryNode. Synthesize an explicit empty batch for each zero-row chunk, or fix the dependency before batching these chunks.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +132,207 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+    }
+
+    SizeEstimateSource size_estimate_source;
+    std::string total_estimate_error;
+    if (projected_estimate_available) {
+        // TODO: Lance single-page variable-width estimates can be lower than
+        // decoded Arrow memory. Add a per-projected-field sampled safety floor
+        // once that metadata is propagated; fallback_bytes_per_row currently
+        // covers the whole column group and would erase projection savings.
+        row_group_sizes = std::move(projected_row_group_sizes);
```
buqian-zilliz, line 218: **[P1] Do not use Lance lower bounds for memory admission**

This accepts any successful projected estimate even when fallback_bytes_per_row contains the DataNode sample of actual Arrow memory. In f9f5bd1, a single-page variable-width Lance column deliberately returns encoded page bytes, which are only a lower bound on decoded memory. These sizes drive cache accounting and the global transient-load budget, so compressed external data can allocate far more than reserved and OOM QueryNode. Keep a conservative sampled floor until a per-projection sample or guaranteed upper bound is available.

internal/core/thirdparty/milvus-storage/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( milvus-storage_VERSION e658197)
+set( milvus-storage_VERSION f9f5bd1)
```
buqian-zilliz, line 17: **[P2] Pinned reader retains quadratic projection metadata**

The f9f5bd1 reader builds a per-row-group map containing every column's estimate inside every projected ChunkReader. The lazy JSON path retains one reader per shredded column, so an N-column group with R row groups now retains O(N²R) metadata before any cell is loaded; the default 1,024-column limit can consume hundreds of megabytes or gigabytes per segment. Share immutable indexed size metadata across readers or retain only requested-column estimates before taking this dependency bump.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -130,21 +132,207 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
+        }
+        if (row_group_rows[i] > std::numeric_limits<uint64_t>::max() /
+                                    positive_fallback_bytes_per_row) {
+            throw std::runtime_error(fmt::format(
+                "positive fallback row group size exceeds the uint64_t "
+                "range, rows {}, bytes per row {}",
+                row_group_rows[i],
+                positive_fallback_bytes_per_row));
+        }
+        row_group_sizes[i] =
```
buqian-zilliz, line 308: **[P2] Preserve known-zero estimate semantics**

The new API reports unavailable estimates as errors and uses zero for a known-zero or physically absent field, but this assignment treats every live-row zero aggregate as unknown and expands it to 4KB per row or the full external sample. A large schema-evolved all-null chunk can therefore be accounted as gigabytes, while the same absent field in a mixed projection is masked by positive siblings and receives no floor. Preserve the distinction and apply a small type-aware estimate per projected field; reserve coarse fallbacks for error paths.

zhengbuqian (approved):

Issue #48060, opened 2026-03-06
[Bug]: [benchmark][streaming-cluster] loaded collection failed and queryNode CrashLoop after ManifestGroupTranslator streaming refactor


wangting0128:
## Summary

After PR #47954 (`ed05357f`, merged 2026-02-28) refactored `ManifestGroupTranslator::get_cells()` from bulk loading to streaming (per-batch) loading, QueryNode enters a CrashLoopBackOff when loading segments with large VARCHAR INVERTED indexes. The streaming approach holds CGO threads for extended periods, starving the Go runtime's etcd keepalive goroutine, causing the session lease to expire and the process to exit.

## Environment

- **Milvus version**: 2.6-20260305-3ec52c5b-amd64 (branch 2.6)
- **Deployment**: Kubernetes, cluster mode with streaming architecture
- **QueryNode resources**: 8 CPU / 32 GB Memory
- **Last passing version**: 2.6-20260226-cf2c455d-amd64
- **First failing version**: 2.6-20260305-3ec52c5b-amd64

## Root Cause

PR #47954 changed `ManifestGroupTranslator::get_cells()`:

```
Before: chunk_reader_->get_chunks(indices, ALL)     // one bulk CGO call, loads all row groups at once
After:  chunk_reader_->get_chunks(rg_indices, 1)    // per-batch CGO call, streamed through CellReaderChannel
```

While this reduces peak memory from `O(all_cells)` to `O(single_cell)`, the streaming approach significantly increases the total wall-clock time that CGO threads are occupied during segment loading. When many sealed segments (20+) are loaded simultaneously—each with 4 indexes using **sync warmup**—all OS threads become blocked in CGO calls (`get_chunks` + `create_group_chunk`), preventing the Go scheduler from running the etcd keepalive goroutine.

The etcd session lease expires after ~45-56 seconds, causing QueryNode to self-terminate. On restart, the coordinator re-dispatches the same segment load requests, creating a crash loop.

### Comparison of three test runs (same test case `test_inverted_locust_varchar_dml_dql_streaming_cluster`)

| | Feb 13 (edef4897) | Feb 26 (cf2c455d) | Mar 5 (3ec52c5b) |
|---|---|---|---|
| **Result** | PASSED | PASSED | **FAILED** |
| **QN Restarts** | 4 (initial only) | 3 (initial only) | 2→8 (CrashLoop) |
[...]

PR #51829, opened 2026-07-24
enhance: bump Arrow recipe revision for Parquet fixes

jiaqizho:
issue: #48060 

Update the Arrow recipe revision to improve buffer reservation and data length estimation, reducing repeated resizes when decoding large string columns from Parquet files.


liliu-z (commented):
/lgtm
/approve

PR #51831, opened 2026-07-24
fix: [3.0] base==current CAS for the sort-stats and external-refresh manifest adoptions

weiliu1031:
issue: #51376
pr: #51724

## Summary

Backport the complete final merged manifest-adoption CAS change from
master to `3.0`.

- Carry the manifest base used by external-refresh and sort-stats workers.
- Adopt results only when the base still matches the current manifest and
  the result is a legal successor.
- Reset stale work for redispatch instead of completing with lost updates.
- Fence superseded external-refresh attempts with `task_version`.
- Retry only known-transient worker failures; fail unknown and permanent
  build or data errors instead of looping in `RefreshPending`.

## Backport notes

This branch was rebuilt directly from the latest `milvus/3.0` and the
complete final squash commit of the master PR was cherry-picked.

The cherry-pick conflicted only in `internal/datacoord/meta_test.go`,
`pkg/proto/data_coord.proto`, and the generated
`pkg/proto/datapb/data_coord.pb.go`.

The `3.0` branch does not contain the master-only `Statistics` schema.
`SegmentInfo.base_manifest` therefore keeps field number 39 while field
38 remains unused. The Go protobuf was regenerated from the resolved
`3.0` schema. Existing `3.0` meta tests were preserved and the master
manifest-successor cases were appended.

## Validation

- `make generated-proto-without-cpp`
- `git diff --cached --check`
- `gofmt -d` on all changed Go files
- Local Go tests were not run for this rebuilt commit; PR CI will validate
  the complete `3.0` backport.


xiaofan-luan:
Closing this backport — the master change it ports (#51724) is being reverted in #51872 so we can rework it rather than patch it in place.

Two reasons, both in the revert description:

1. **Design rework.** The worker attempt fencing (`task_version` on the wire, newer attempt evicts the resident one, version guard drops the evicted attempt's late writes) is being replaced by reject-on-occupied registration, matching what the index / stats / analyze workers already do — that removes the class of stale write instead of guarding against it, and takes the version off the wire. And `base_manifest` should not live on `SegmentInfo`: it is an adoption token belonging to the result message, as it already does for `StatsResult` and `CompactionSegment`.

2. **Incomplete coverage.** The same failure class is unaddressed for index metadata — an in-place refresh patch invalidates the segment's text/JSON stats but leaves `indexMeta` untouched, and the index inspector skips any segment that already has a `SegmentIndex` entry, so the index is never rebuilt. That is reachable on the purely sequential path, no concurrency needed.

Master and 3.0 will be re-submitted together once the rework lands. Sorry for the churn, and thanks for putting the backport together — nothing was wrong with the port itself.

Issue #51837, opened 2026-07-24
[Bug]: File resource sync advances version before analyzer update succeeds


bigsheeper:
## What happened

`SyncManager.Sync` updates its local `resourceMap` and stores the target file resource `version` before `analyzer.UpdateGlobalResourceInfo` succeeds.

If `analyzer.UpdateGlobalResourceInfo` returns an error, `Sync` returns that error to the coordinator. The coordinator retries the same target version, but the next `Sync(version, resources)` call is skipped by this guard:

```go
if version <= m.version.Load() {
    return nil
}
```

That makes the retry report success even though the analyzer global resource map was never updated. A node can be marked synced for a file resource version while analyzer-backed resources remain stale or missing.

## Impact

High correctness issue for file resources in `sync` mode. QueryNode/DataNode can download the files but fail to update analyzer runtime options, then acknowledge the version on retry. Text analyzer resources that depend on the global resource map may remain unavailable until a later version bump or process restart.

## Reproduction

A minimal failing unit test patches `analyzer.UpdateGlobalResourceInfo` to fail after the file is downloaded:

```go
func (suite *SyncManagerSuite) TestSync_AnalyzerUpdateFailureDoesNotAdvanceVersion() {
    mockey.PatchConvey("failed analyzer update keeps version retryable", suite.T(), func() {
        expectedErr := errors.New("mock analyzer update failed")
        mockey.Mock(analyzer.UpdateGlobalResourceInfo).Return(expectedErr).Build()

        resources := []*internalpb.FileResourceInfo{
            {Id: 1, Name: "test.file", Path: "/storage/test.file"},
        }
        suite.mockStorage.EXPECT().Reader(context.Background(), "/storage/test.file").Return(newMockReader("test content"), nil).Once()

        err := suite.manager.Sync(1, resources)
        suite.ErrorIs(err, expectedErr)
        suite.Equal(uint64(0), suite.manager.GetVersion())
    })
}
```

Current output on master (`7fbb6f4bf8`):

```text
Error: Not equal:
    expected: 0x0
    actual  : 0x1
Test: TestSyncManagerSuite/TestSync_AnalyzerUpdateFailureDoesNotAdvanceVersion
```

Command used:

```bash
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/util/fileresource -run TestSyncManagerSuite/TestSync_AnalyzerUpdateFailureDoesNotAdvanceVersion -v
```

[...]

PR #51838, opened 2026-07-24
fix: keep file resource sync retryable on analyzer failure

bigsheeper:
issue: #51837

## What changed

Move `SyncManager.Sync`'s `analyzer.UpdateGlobalResourceInfo` call before advancing the local file resource `version` and `resourceMap`.

If analyzer runtime option update fails, the node now keeps the old synced version so the coordinator's next `SyncFileResource` retry will execute the sync path again instead of being skipped as already complete.

## Verification

- `GOWORK=/home/sheep/workspace/milvus/.worktrees/fix-fileresource-sync-version/go.work PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64 go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/util/fileresource -v`
- `GOWORK=/home/sheep/workspace/milvus/.worktrees/fix-fileresource-sync-version/go.work make static-check`

## Notes

The local worktree needed untracked helper symlinks for existing C++ output and generated migration legacy proto files so Milvus tooling could run from the required `.worktrees/` location. Those files are not included in the commit.

### DIFF
--- a/internal/util/fileresource/manager.go
+++ b/internal/util/fileresource/manager.go
@@ -232,6 +232,9 @@ func (m *SyncManager) Sync(version uint64, resourceList []*internalpb.FileResour
 		}
 	}
 
+	if err := analyzer.UpdateGlobalResourceInfo(newResourceMap); err != nil {
+		return err
+	}
 	for _, resourceID := range removes {
 		err := os.RemoveAll(path.Join(m.localPath, fmt.Sprint(resourceID)))
 		if err != nil {
@@ -240,9 +243,6 @@ func (m *SyncManager) Sync(version uint64, resourceList []*internalpb.FileResour
 	}
 	m.resourceMap = newResourceMap
 	m.version.Store(version)
-	if err := analyzer.UpdateGlobalResourceInfo(newResourceMap); err != nil {
-		return err
-	}
 	notifyListeners(SyncEvent{Version: version, Resources: resolvedResources})
 	return nil
 }
--- a/internal/util/fileresource/manager_test.go
+++ b/internal/util/fileresource/manager_test.go
@@ -28,10 +28,13 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/bytedance/mockey"
+	"github.com/cockroachdb/errors"
 	"github.com/stretchr/testify/suite"
 	"go.uber.org/atomic"
 
 	"github.com/milvus-io/milvus/internal/mocks"
+	"github.com/milvus-io/milvus/internal/util/analyzer"
 	"github.com/milvus-io/milvus/pkg/v3/proto/internalpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/conc"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
@@ -225,6 +228,22 @@ func (suite *SyncManagerSuite) TestSync_UpdateAndRemoveNotifyListener() {
 	suite.Require().Len(listener.events, 3)
 	suite.Empty(listener.events[2].Resources)
 	suite.NoDirExists(path.Join(suite.tempDir, "2"))
+}
+
+func (suite *SyncManagerSuite) TestSync_AnalyzerUpdateFailureDoesNotAdvanceVersion() {
+	mockey.PatchConvey("failed analyzer update keeps version retryable", suite.T(), func() {
+		expectedErr := errors.New("mock analyzer update failed")
+		mockey.Mock(analyzer.UpdateGlobalResourceInfo).Return(expectedErr).Build()
+
+		resources := []*internalpb.FileResourceInfo{
+			{Id: 1, Name: "test.file", Path: "/storage/test.file"},
+		}
+		suite.mockStorage.EXPECT().Reader(context.Background(), "/storage/test.file").Return(newMockReader("test content"), nil).Once()
+
+		err := suite.manager.Sync(1, resources)
+		suite.ErrorIs(err, expectedErr)
+		suite.Equal(uint64(0), suite.manager.GetVersion())
+	})
 }
 
 func (suite *SyncManagerSuite) TestMode() {
--- a/internal/util/fileresource/manager_test.go
+++ b/internal/util/fileresource/manager_test.go
@@ -233,7 +233,7 @@ func (suite *SyncManagerSuite) TestSync_UpdateAndRemoveNotifyListener() {
 func (suite *SyncManagerSuite) TestSync_AnalyzerUpdateFailureDoesNotAdvanceVersion() {
 	mockey.PatchConvey("failed analyzer update keeps version retryable", suite.T(), func() {
 		expectedErr := errors.New("mock analyzer update failed")
-		mockey.Mock(analyzer.UpdateGlobalResourceInfo).Return(expectedErr).Build()
+		mocker := mockey.Mock(analyzer.UpdateGlobalResourceInfo).Return(expectedErr).Build()
 
 		resources := []*internalpb.FileResourceInfo{
 			{Id: 1, Name: "test.file", Path: "/storage/test.file"},
@@ -243,6 +243,16 @@ func (suite *SyncManagerSuite) TestSync_AnalyzerUpdateFailureDoesNotAdvanceVersi
 		err := suite.manager.Sync(1, resources)
 		suite.ErrorIs(err, expectedErr)
 		suite.Equal(uint64(0), suite.manager.GetVersion())
+
+		// the same version is retried and reaches the analyzer again, re-downloading the resource
+		suite.mockStorage.EXPECT().Reader(context.Background(), "/storage/test.file").Return(newMockReader("test content"), nil).Once()
+		mocker.Return(nil)
+
+		suite.Require().NoError(suite.manager.Sync(1, resources))
+		suite.Equal(uint64(1), suite.manager.GetVersion())
+		content, err := os.ReadFile(path.Join(suite.tempDir, "1", "test.file"))
+		suite.NoError(err)
+		suite.Equal("test content", string(content))
 	})
 }
 


czs007:
Adversarial review found one low-severity, pre-existing issue; the change itself is correct and mergeable.

## Low

**`internal/util/fileresource/manager.go:235`** — `Sync` downloads each changed resource into `dir(newId)` before any early-return point, but the `removes` list is derived solely from the stale `m.resourceMap` (`:223-233`) and only runs at the end (`:238-243`). Any early return therefore leaves a downloaded-but-untracked directory behind, and `SyncManager` has no orphan GC (`GcLoop`/`CleanResource` at `:352-376` belong to `RefManager` and reclaim by ref-count, not by scanning untracked dirs). Failure scenario: version N contains resources A (id1→id2) and B (id4→id5); A downloads into `dir(id2)` successfully, then B's object-storage reader fails and returns early at `:212-214`, leaving `m.resourceMap` at `{A:id1, B:id4}` — a later version N+1 = `{A:id7, B:id5}` computes `removes` as only `id1`/`id4`, so `dir(id2)` is never referenced and never reclaimed, accumulating one leaked copy per failure until the process restarts.

This is not introduced by this PR: two such early-return paths already exist on master (`MkdirAll` failure at `:187-190`, object-storage download failure at `:212-214`, the latter reachable under ordinary network/storage errors); the new `UpdateGlobalResourceInfo` call adds a third. The leak is also bounded by process lifetime — `cleanLocalDir(RootCachePath)` wipes the cache directory at querynode startup (`cmd/roles/roles.go:192-200`, non-posix mode). The reorder in this PR is strictly an improvement, since it switches the analyzer's `resource_map` to the new ids before the old directories are deleted, closing the window where the analyzer could resolve a just-removed path.

Suggestion: handle separately from this PR — either add a `defer` in `Sync` that rolls back directories downloaded during a failed attempt, or add a directory-scan cleanup reconciled against `m.resourceMap`. (raised by chyezh, tinswzy)

bigsheeper:
Thanks — verified the finding against the current head (`17c4fdf4b`). It holds, and it is genuinely pre-existing. Not addressing it in this PR; two additions below.

**Verified:**

- `Sync` downloads into `localResourcePath` before any early return, and `removes` is computed from the stale `m.resourceMap` only after the download loop, with the `RemoveAll` loop at the very end. So an early return does leave a downloaded-but-untracked directory.
- The two pre-existing early returns are the `os.MkdirAll` failure and the download closure's `return err` (reader / `os.Create` / `io.Copy`). Neither is in this PR's diff — it only moves the `UpdateGlobalResourceInfo` call, which adds a third.
- `CleanResource` and `GcLoop` are `RefManager` methods and reclaim by ref-count (`m.ref[key] <= 0`), never by scanning for untracked directories. `SyncManager` has no GC of its own. Confirmed.

**Addition 1 — the startup bound holds, and here is why it applies at all.** You asserted the leak is bounded by `cleanLocalDir(RootCachePath)`; I checked the two links that make that true, since neither is obvious:

- `RootCachePath` resolves to `<localStorage>/cache` while `FileResourcePath` resolves to `<localStorage>/cache/<nodeID>/file_resource` (`internal/util/pathutil/path_util.go:32-46`), so the leaked directories really are inside the wiped subtree.
- `fileresource.InitManager` is called from exactly one place, `internal/querynodev2/server.go:415`, so `SyncMode` only ever runs on querynode — which is the one role whose startup performs that wipe.

**Addition 2 — the bound is narrower than stated.** `runQueryNode` (`cmd/roles/roles.go:192-200`) only calls `cleanLocalDir` when `CommonCfg.EnablePosixMode` is false; under posix mode the in-code comment says the clean task is done by mixcoord instead. I did not trace the mixcoord path, so "bounded by process lifetime" is verified for non-posix mode only. Under posix mode I can't currently claim the leak is reclaimed on restart.

**On your note that the reorder is an improvement:** agreed, and it's a benefit I had not claimed in the PR description. Before the change the old directories were deleted while the analyzer still held the *old* `resource_map`, so there was a window where it could resolve a path that had just been removed; moving `UpdateGlobalResourceInfo` ahead of the `RemoveAll` loop closes it. To be precise about how far I verified that: it follows from the statement ordering by inspection — I did not fault-inject the window, so I'd rather not state it as a tested guarantee.

**Why not here:** the leak is orthogonal to the retryability fix in this PR and predates it. Fixing it properly means either a rollback `defer` over the directories downloaded during a failed attempt, or a directory scan reconciled against `m.resourceMap` — both of which want their own tests and their own review.

czs007:
## Re-review of `17c4fdf4bd34`

- **Confirmed** — *low* — `internal/util/fileresource/manager.go:235` (Sync early-return leaves orphaned downloaded resource directories, pre-existing): both reviewers re-verified that downloads land in `dir(newId)` before any removal candidates are computed, and that `removes` is derived solely from the stale `m.resourceMap`, so the orphan path still stands.

---

Adversarial review found one low-severity, pre-existing issue; the ordering fix itself is correct and mergeable.

### Low

**`internal/util/fileresource/manager.go:235` — Sync early-return leaves an orphaned resource directory on disk**

`Sync` creates and downloads into `<localPath>/<newId>/` (`os.MkdirAll` at line 187, `io.Copy` at line 206) before any fallible step completes, while removal candidates are derived exclusively from the stale `m.resourceMap` (lines 223-233) and executed only afterwards (lines 238-243) — so an ID downloaded in the current round can never enter `removes`. Every early-return path (MkdirAll failure at 187-190, object-storage download failure at 212-214, and the analyzer-update return added by this PR at 235-237) leaves the created directory behind while `m.resourceMap` keeps its old value, and the package has no directory-scanning orphan GC (only three `os.RemoveAll` calls against known paths, at 182, 236, 359).

Failure scenario: on a DataNode with `dnMode=sync`, version 1 downloads resource 10 and then fails on resource 11 (object-store error) → the user deletes resource 10 and version 2 syncs successfully without it → `<localStorage>/cache/<nodeID>/file_resource/10/` is never removed — it is neither in the new list (so never overwritten) nor in `m.resourceMap` (so never in `removes`) — and `runDataNode` (cmd/roles/roles.go:207-209) performs no startup `cleanLocalDir`, so it survives restarts and permanently occupies local disk.

Suggestion: no change is required in this PR — the defect predates it, and it self-heals under normal retries, since the version is not advanced on early return and a retry of the same name/ID does `os.RemoveAll(localResourcePath)` at line 182 before re-downloading; only a changed ID (or a permanently abandoned sync) turns it into a permanent leak. Worth tracking separately as either a `defer`-based cleanup of partially downloaded directories on the error paths, or a startup/periodic orphan sweep on the DataNode path. (raised by chyezh, xiaocai2333)

### Verified

- `internal/util/fileresource/manager.go:235-246` — the analyzer update now succeeds before old directories are removed and before `resourceMap`/`version` advance; on failure the RPC returns an error while retaining the previous local version and map, and no later error path can leave the version uncommitted.
- `internal/util/fileresource/manager.go:159-220` — every new resource file is fully downloaded before the analyzer receives the new resource map, so analyzer consumers cannot resolve a newly published ID before its file exists, and the analyzer never points at a directory that `removes` will delete.
- `internal/util/fileresource/manager.go:150-157` — synchronization remains serialized under the manager lock, and the relocated analyzer call stays inside the same critical section, so the version guard cannot race concurrent `Sync` calls.
- `internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/options/runtime_option.rs:40-43,129-131` — the analyzer update replaces the global `ResourceInfo` while holding its own write lock, so readers cannot observe a partially replaced resource map.
- `internal/coordinator/file_resource_observer.go:198-219,234-261` — failed node RPCs are not recorded in the coordinator distribution, so the existing retry loop resends the same target version.
- `internal/util/fileresource/manager_test.go:233-246` — the added failure-path test asserts both error propagation and that a failed analyzer update does not advance the version, while the existing success/update/removal coverage still exercises the reordered path.

bigsheeper:
Correcting myself: my previous comment contained a factual error, and it changed the conclusion I drew. Your `dnMode=sync` framing is right and mine was wrong.

**What I got wrong.** I wrote that `fileresource.InitManager` is called from exactly one place and that `SyncMode` therefore only ever runs on querynode. There are five call sites:

- `internal/querynodev2/server.go:415` — `common.fileResource.mode.queryNode`, default `sync`
- `internal/streamingnode/server/server.go:49` — same param, default `sync`
- `internal/datanode/data_node.go:204`, `:212`, `:215` — `common.fileResource.mode.dataNode`, **also default `sync`**

I reached the wrong conclusion by reading only the head of a grep whose output was dominated by unrelated `function.GetManager()` hits. The specific claim "called from exactly one place" was never true.

**What that changes.** I used it to argue the orphan is bounded by process lifetime, via the querynode startup wipe. That bound does not hold. `cleanLocalDir` has exactly one production call site — `cmd/roles/roles.go:198`, inside `runQueryNode`, and only when `EnablePosixMode` is false. `runDataNode` and `runStreamingNode` do not call it at all.

So for a datanode or streamingnode running the default `sync` mode, an orphaned `<localStorage>/cache/<nodeID>/file_resource/<id>/` directory **survives restarts indefinitely**. That is precisely the scenario your re-review describes, and it makes the finding meaningfully worse than my reply implied: not "leaks until the process restarts" but "leaks permanently" on two of the three roles that run this code.

My earlier "the bound is narrower than stated" note was directionally right but far too small — I framed it as a posix-mode caveat on querynode, when the real gap is two entire roles with no startup sweep at all.

**What stands.** The rest of my verification holds: `Sync` downloads into `dir(newId)` before any early return, `removes` derives only from the stale `m.resourceMap`, and `CleanResource`/`GcLoop` are `RefManager` methods that reclaim by ref-count. And I agree with your conclusion that no change is required in this PR — the defect predates it and the reorder here is still an improvement. I'm only correcting the reasoning I offered about its blast radius, since a reader would otherwise take "bounded by process lifetime" from me and under-rate it.

czs007:
## Re-review of `17c4fdf4bd34`

- **Orphan resource directories left behind by `Sync`'s early-return paths (`internal/util/fileresource/manager.go:235`, no orphan GC in `SyncManager`) — Confirmed:** cleanup candidates are still derived only from the old `resourceMap`, and `SyncManager` still has no GC loop, so directories downloaded in a failed round can remain permanently.
- **Orphan resource directories left behind by `Sync`'s early-return path (`internal/util/fileresource/manager.go:235`) — Confirmed:** same defect as above (the two earlier bullets describe one issue); it is re-reported once below with the fuller reproduction chain.
- **Failure-path unit test does not assert that a retry advances the version (`internal/util/fileresource/manager_test.go:233`) — Confirmed:** the test asserts the analyzer error is returned and the version stays at 0, but never clears the mock and re-runs `Sync(1, resources)` to prove the version can then reach 1.
- **`notifyListeners` swallows listener errors after the version is committed (`internal/util/fileresource/manager.go:246`) — Confirmed as latent only:** `notifyListeners` still logs and discards `OnFileResourceSync` errors, but the sole registered production listener (`internal/util/function/chain/expr/xgboost_cache.go:108`) unconditionally returns nil, so there is no reachable production failure today.

---

This round's independent audit found no new critical or high-severity problems with the reordering of `analyzer.UpdateGlobalResourceInfo`; one pre-existing low-severity issue is re-reported below.

### Low

**`internal/util/fileresource/manager.go:187` — orphan resource directories from `Sync`'s early-return paths (pre-existing, not introduced by this PR)**

`Sync` creates each new resource directory with `os.MkdirAll` (line 187) and downloads into it (lines 192–214), but the deletion candidate set `removes` is derived exclusively from the *old* `m.resourceMap` (lines 223–233) and applied at lines 238–243 — directories created in the current round can never enter it. Failure scenario: with fileResource sync enabled, the coordinator distributes version N containing resources A (id=10) and B (id=11); on a QueryNode, A downloads successfully into `<localPath>/10/`, then B's object-storage `Reader` fails and `Sync` returns early at lines 212–214; if A is subsequently removed from the coordinator before a retry succeeds, `<localPath>/10/` is referenced by neither the old map nor any future resource list → the directory stays on the node's local disk permanently, and since `SyncManager` has no GC loop (unlike `RefManager.CleanResource`/`GcLoop` at lines 352–376), repeated occurrences steadily consume local disk with model/dictionary files that can reach hundreds of MB, recoverable only by manual cleanup.

The new early return added by this PR at lines 235–237 shares the same state effect, but the defect is reachable through the pre-existing download path alone and is not a merge blocker. Suggested fix: download into a staging directory and clean it up on failure, or add an explicit reconciliation scan over `<localPath>/<resourceID>` that removes directories absent from the current resource list. *(raised by chyezh, xiaocai2333)*

bigsheeper:
Thanks — third round confirms all three. One of them is this PR's own debt and is fixed in `e44fbb2f66`; per finding below. (The branch is also rebased onto latest master, so the fix commit you reviewed as `17c4fdf4bd34` is now `0ce4885898`; its content is unchanged.)

**`manager_test.go:233` — test gap: confirmed, fixed.**

You're right that the case only proved half the claim: it asserted the analyzer error propagates and the version stays at 0, but never re-ran `Sync(1, resources)` to show the retry converges — and "retryable" is the whole point of the change.

Being precise about what the gap actually was, since I first mis-stated it to myself: the existing assertion is not toothless. I reverted `manager.go` to the pre-fix ordering and `suite.Equal(uint64(0), GetVersion())` does fail (`expected 0x0, actual 0x1`). What it could not show is that the retry *reaches* anything — with the fix reverted, the second `Sync(1, ...)` is short-circuited by the `version <= m.version.Load()` guard at `manager.go:155-157` and never touches the downloader, which under the new assertions surfaces as `1 out of 2 expectation(s) were met`. So the added half covers a failure mode the old half structurally could not.

What the retry path does, verified by trace and now pinned by the test:

- after the failed attempt `m.resourceMap` is untouched (`newResourceMap` is a separate map, assigned only at `manager.go:244`) and `m.version` is still 0, so the guard lets the same version back in;
- the resource is absent from `resourceMap`, so `Sync` re-runs `RemoveAll` + `MkdirAll` and **re-downloads** — hence a second `Reader` expectation rather than `.Twice()` on the first, since a reused `newMockReader` would be exhausted and would silently write an empty file. The test asserts the file content after the retry so that regression cannot pass.

On who drives the retry: `FileResourceObserver.syncLoop` (`internal/coordinator/file_resource_observer.go:103-121`) re-issues on error via `RetryNotify` → 3s sleep → `Notify`, re-reading `targetVersion` from meta — so the node does see the same version again, unbounded. That is the loop the pre-fix ordering permanently defeated.

**`manager.go:187` orphan resource directories — confirmed again, still not fixed here.**

Nothing new from my side; my position is unchanged from last round, including the correction I posted about its blast radius (default `sync` mode on datanode and streamingnode has no startup `cleanLocalDir`, so these directories survive restarts, not merely the process). Fixing it means a staging directory with rollback on failure, or a reconciliation scan of `<localPath>/<resourceID>` against the current list — both want their own tests and review, and neither is coupled to the six-line reorder here.

**`manager.go:246` `notifyListeners` swallowing listener errors — agreed, latent, and same family.**

Re-checked: `notifyListeners` (`manager.go:85-92`) logs and discards, and the sole production listener `xgboostModelCache.OnFileResourceSync` (`xgboost_cache.go:108-120`) has no error path, so it is unreachable today. It is the same shape as the orphan finding — a post-commit failure that nothing reclaims — so I'd rather fix that class in one follow-up than bolt a partial fix onto this PR.

czs007:

Verified:
- `internal/util/fileresource/manager.go:235` — the analyzer state is published before `resourceMap` and `version` are committed, and an analyzer failure returns immediately, leaving both local synchronization markers unchanged.
- `internal/util/fileresource/manager.go:155` — because a failed attempt does not advance `version`, the coordinator's retry with the same target version is not skipped by the early return, so retries are genuinely effective.
- `internal/util/fileresource/manager.go:151` — the manager mutex serializes downloads, analyzer publication, cleanup, and version advancement, so retries or concurrent syncs cannot interleave state commits.
- `internal/coordinator/file_resource_observer.go:209` — only a success status updates the coordinator's distribution; an analyzer failure is never recorded as synced, and the error path triggers a delayed re-notification with the same target version.
- `internal/util/analyzer/canalyzer/c_analyzer_factory.go:109` — the underlying typed segcore error is preserved as-is, so node RPCs propagate a proper failure status.
- `internal/util/fileresource/manager_test.go:233` — the regression test injects an analyzer failure, asserts the version stays at 0, then retries the identical version and requires a second download (also catching an incorrectly advanced `resourceMap`) before verifying version 1 and the file content; this test fails on the old implementation.

czs007 (approved):
/lgtm

PR #51839, opened 2026-07-24
enhance: [3.0] amortize per-row chunk resolution in storage-v2 bulk access and expr offset paths

xiaofan-luan:
## Summary

Backport of the storage-v2 per-row chunk resolution amortization to 3.0.

issue: #50991
pr: #51678

Cherry-pick of the single commit on the master PR (clean cherry-pick, no
conflicts; `append_valid_data` / `ThreadSafeValidData` / `ChunkDataAccessor`
call sites verified consistent on the 3.0 tree).

See #51678 for the full description: borrow + per-distinct-chunk dedup in
`ProxyChunkColumn` bulk accessors, chunk-cached expression offset paths
(sealed and growing), batched `BulkIsValid` in `bulk_subscript`, and the
enabling `ThreadSafeValidData` chunked-storage rewrite that makes growing
nullable validity pointers stable under concurrent insert.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Md98mTWZSM9xDmLjeVYdvh

PR #51840, opened 2026-07-24
enhance: batch growing validity lookups instead of copying the whole bitmap

xiaofan-luan:
## Summary

Follow-up to #51678 (**stacked on it — the first commit is that PR's; will
rebase to a single commit once it merges**).

`FilterVectorValidOffsets` (growing `bulk_subscript`, nullable vector fields)
called `vec_base->get_valid_data()` **on every query**: a flat O(segment rows)
copy of the entire validity bitmap, used only to answer `count` offsets on the
non-mapping branch — and on the mapping-storage branch (the common case for
nullable vectors) only for an emptiness check.

## Changes

- New `ThreadSafeValidData::bulk_is_valid(offsets, count, out)`: one lock
  acquisition, `count` chunk-indexed reads; out-of-range offsets yield `false`
  (same tolerant semantics as the loop it replaces).
- `FilterVectorValidOffsets` gates on the existing `empty()` accessor and uses
  `bulk_is_valid` on the non-mapping branch. Behavior unchanged, including the
  mapping branch and the `valid_offsets` / `valid_count` accounting.
- `get_data()` now carries a WARNING doc: it materializes a flat copy of the
  whole bitmap — avoid on hot paths; prefer `empty()` / `is_valid()` /
  `bulk_is_valid()` / `get_chunk_data()`.

Per-query cost on this path drops from O(segment rows) copy + O(count) reads
to a single lock + O(count) reads.

Remaining `get_data()` callers (insert-path interim-index append in
`FieldIndexing.cpp`) need a flat buffer for index build and are left as-is.

https://claude.ai/code/session_01Md98mTWZSM9xDmLjeVYdvh

### DIFF
--- a/internal/core/src/common/GroupChunk.h
+++ b/internal/core/src/common/GroupChunk.h
@@ -44,6 +44,22 @@ class GroupChunk {
         return it->second;
     }
 
+    // Borrow the chunk for a field WITHOUT sharing ownership. The returned
+    // pointer is valid only while this GroupChunk stays pinned/resident (e.g.
+    // for the lifetime of the CellAccessor that pinned it). Bulk accessors use
+    // this on the hot path to avoid the per-row shared_ptr refcount bump that
+    // GetChunk() incurs — an atomic RMW on a shared control block that contends
+    // across concurrent search threads. For anything that must outlive the pin,
+    // use GetChunk() instead.
+    Chunk*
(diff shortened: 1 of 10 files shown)


xiaofan-luan:
Folded into #51678 (and its 3.0 backport #51839) per maintainer preference — no separate follow-up PR.

PR #51841, opened 2026-07-24
enhance: tighten sealed segment reopen COW lifecycle

congqixia:
Related to #51068

- reject online reopen when segment row count changes
- publish commit timestamps with segment snapshot state
- retire replaced index warmups only after successful publication
- enforce text index source and schema identity across reopen
- remove redundant mutex locks from snapshot-owned sealed leaf readers
- propagate captured snapshot and runtime state through nested helpers
- defer the Search/Retrieve synchronization split to a follow-up

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -13,6 +13,8 @@
 #include "segcore/default_fs.h"
 
 #include <cxxabi.h>
+#include <boost/uuid/random_generator.hpp>
+#include <boost/uuid/uuid_io.hpp>
 #include <fmt/core.h>
 #include <folly/ScopeGuard.h>
 #include <folly/Try.h>
@@ -1358,6 +1360,82 @@ ChunkedSegmentSealedImpl::NormalizePublishedState(
     SetSystemFieldReadyInState(state, state.load_info.get());
 }
 
+void
+ChunkedSegmentSealedImpl::ValidateTextIndexState(
+    const PublishedSegmentState& state) const {
+    if (state.schema == nullptr || state.load_info == nullptr ||
+        state.runtime == nullptr) {
+        bool has_runtime_text_index =
+            state.runtime != nullptr && !state.runtime->text_indexes.empty();
+        bool has_text_index_metadata =
+            state.load_info != nullptr &&
+            (!state.load_info->GetCreatedTextIndexes().empty() ||
+             !state.load_info->GetTextStatsLogs().empty());
+        AssertInfo(!has_runtime_text_index && !has_text_index_metadata,
+                   "published text-index state requires schema, load info, "
+                   "and runtime resources");
+        return;
+    }
+
+    const auto& markers = state.load_info->GetCreatedTextIndexes();
+    const auto& stats = state.load_info->GetTextStatsLogs();
+    const auto& indexes = state.runtime->text_indexes;
+
+    for (const auto& [field_id, _] : markers) {
+        AssertInfo(field_exists_in_schema(state.schema, field_id),
+                   "raw-built text index marker references dropped field {}",
+                   field_id.get());
+        AssertInfo(stats.find(field_id.get()) == stats.end(),
+                   "text index field {} is both raw-built and pre-built",
+                   field_id.get());
+        auto runtime_it = indexes.find(field_id);
+        AssertInfo(runtime_it != indexes.end() &&
+                       std::holds_alternative<
+                           std::shared_ptr<index::TextMatchIndexHolder>>(
+                           runtime_it->second) &&
+                       std::get<std::shared_ptr<index::TextMatchIndexHolder>>(
+                           runtime_it->second) != nullptr,
+                   "raw-built text index marker for field {} has no raw "
+                   "runtime holder",
+                   field_id.get());
+    }
+
+    for (const auto& [field_id_value, _] : stats) {
+        auto field_id = FieldId(field_id_value);
+        AssertInfo(field_exists_in_schema(state.schema, field_id),
+                   "pre-built text index metadata references dropped field {}",
+                   field_id_value);
+        AssertInfo(markers.find(field_id) == markers.end(),
+                   "text index field {} is both pre-built and raw-built",
+                   field_id_value);
+        auto runtime_it = indexes.find(field_id);
+        using PrebuiltTextIndexSlot =
+            std::shared_ptr<cachinglayer::CacheSlot<index::TextMatchIndex>>;
+        AssertInfo(
+            runtime_it != indexes.end() &&
+                std::holds_alternative<PrebuiltTextIndexSlot>(
+                    runtime_it->second) &&
+                std::get<PrebuiltTextIndexSlot>(runtime_it->second) != nullptr,
+            "pre-built text index metadata for field {} has no cache "
+            "slot",
+            field_id_value);
+    }
+
+    for (const auto& [field_id, runtime_index] : indexes) {
+        AssertInfo(field_exists_in_schema(state.schema, field_id),
+                   "runtime text index references dropped field {}",
+                   field_id.get());
+        bool raw = std::holds_alternative<
+            std::shared_ptr<index::TextMatchIndexHolder>>(runtime_index);
+        AssertInfo(raw ? markers.find(field_id) != markers.end()
+                       : stats.find(field_id.get()) != stats.end(),
+                   "runtime {} text index for field {} has no matching "
+                   "published metadata",
+                   raw ? "raw-built" : "pre-built",
+                   field_id.get());
+    }
+}
+
 std::shared_ptr<ChunkedSegmentSealedImpl::PublishedSegmentState>
 ChunkedSegmentSealedImpl::BuildNextPublishedState(
     const std::shared_ptr<const PublishedSegmentState>& current,
@@ -1444,9 +1522,11 @@ ChunkedSegmentSealedImpl::CloneLoadInfoWithDefaultFilled(
 
 std::shared_ptr<const SegmentLoadInfo>
 ChunkedSegmentSealedImpl::CloneLoadInfoWithTextIndexCreated(
-    const std::shared_ptr<const SegmentLoadInfo>& current, FieldId field_id) {
+    const std::shared_ptr<const SegmentLoadInfo>& current,
+    FieldId field_id,
+    RawTextIndexSource source) {
     auto load_info_copy = std::make_shared<SegmentLoadInfo>(*current);
-    load_info_copy->SetTextIndexCreated(field_id);
+    load_info_copy->SetTextIndexCreated(field_id, source);
     return std::const_pointer_cast<const SegmentLoadInfo>(load_info_copy);
 }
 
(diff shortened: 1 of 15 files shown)


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7534,6 +7511,13 @@ ChunkedSegmentSealedImpl::Reopen(
 
     SegmentLoadInfo current_mutable(*current->load_info);
     SegmentLoadInfo new_local(new_load_info, target_schema);
```
liliu-z, line 7890: `new_local` is not pruned against `target_schema` on the full Reopen path. If a field is dropped but the incoming load info still has its `textstatslogs`, this Reopen publishes stale stats and the next Reopen fails the `HasFieldInSchema` assertion. Can we run the same schema pruning used by `ReplaceSchemaForReopen()` here?

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7534,6 +7511,13 @@ ChunkedSegmentSealedImpl::Reopen(
...
+    AssertInfo(
+        current->load_info->GetNumOfRows() == new_local.GetNumOfRows(),
```
liliu-z: This only checks the top-level proto row count. Storage V2 derives the actual staged row count from column-group `entries_num`, and the V2 loader has no check like the V1 path. So top-level N + replacement entries M can still publish `runtime.row_count=M` while the delete bitmap stays at N. Please validate the field/group row counts against the published N too.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7534,6 +7511,13 @@ ChunkedSegmentSealedImpl::Reopen(
...
+        "online reopen cannot change row count for segment {}: current {}, "
+        "incoming {}; use full segment replacement instead",
+        id_,
+        current->load_info->GetNumOfRows(),
+        new_local.GetNumOfRows());
     new_local.InheritCachedColumnGroupsFrom(*current->load_info);
     for (auto fid : current->load_info->GetCreatedTextIndexes()) {
```
liliu-z: Keeping this marker by FieldId is not enough to prove the RawBuilt text index is unchanged. A same-ID/same-row-count binlog or manifest replacement can change the column data while this marker makes `ComputeDiffTextIndexes()` skip rebuilding, so we can publish new column data with the old text index. Please track a real source generation/fingerprint, rebuild on the staged runtime, or at least reject data-replacing Reopen for RawBuilt fields.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -846,6 +884,89 @@ JsonStatsLoadIdentityEqual(const proto::segcore::JsonKeyStats& lhs,
...
+            "online reopen cannot change text index source for segment {}, "
+            "field {} from pre-built to {}; use full segment replacement "
+            "instead",
+            new_info.GetSegmentID(),
+            field_id,
+            incoming_meta.enable_match() ? "raw-built" : "none");
+        AssertInfo(
+            TextStatsLoadIdentityEqual(current_stats, *incoming_stats),
+            "online reopen cannot replace pre-built text index identity for "
+            "segment {}, field {}; use full segment replacement instead",
```
liliu-z: `use full segment replacement` is not actionable today. QueryCoord turns manifest/DataVersion lag into `ActionTypeReopen`, and a Reopen failure has no Reduce+Grow/full-load fallback, so this can retry forever (external refresh is a real path). We need either staged replacement here or a structured `full reload required` error that QueryCoord handles.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7555,10 +7701,15 @@ ChunkedSegmentSealedImpl::Reopen(
 
     SegmentLoadInfo current_mutable(*current->load_info);
     SegmentLoadInfo new_local(new_load_info, target_schema);
+    // Construction filters derived caches, while reopen normalization also
+    // removes proto-backed state for fields dropped from the target schema.
+    new_local.ReplaceSchemaForReopen(target_schema);
+    new_local.ValidateReopenRowCount(current->load_info->GetNumOfRows());
```
liliu-z, line 7892: The reopen path's new rejections — row-count mismatch (ValidateReopenRowCount at 7707) and pre-built/raw-built text-source changes (ComputeDiff at 7716) — throw via AssertInfo, which carries the default UnexpectedError code. The adjacent stale-schema rejection at 7692 instead throws a distinguishable ThrowInfo(kCollectionSchemaVersionNotReady), showing the codebase does emit specific codes for reopen rejections that need distinct Go handling. On the Go side, querynode ReopenSegments returns the error verbatim with no full-reload fallback (segment_loader.go:2475), and the querycoord IndexChecker re-issues the identical reopen every ~10s (default IndexCheckInterval) with no error-code awareness. So any reachable 'requires full replacement' rejection (see the legacy row-count and raw-built->prebuilt cases) wedges the segment: it never picks up its updated index/stats. Emit a distinguishable code for these rejections so a follow-up Go path can fall back to a full segment reload instead of retrying the in-place reopen.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -75,6 +75,50 @@ SegmentLoadInfo::HasManifestColumn(const std::string& column_name) const {
...
+                }
+            }
+        }
+        if (!retained) {
+            continue;
+        }
+
+        int64_t actual = 0;
+        for (const auto& binlog : field_binlog.binlogs()) {
+            actual += binlog.entries_num();
```
liliu-z, line 271: For non-manifest (legacy V1) segments, ValidateReopenRowCount sums each retained field's insert-binlog entries_num and asserts the sum equals num_of_rows. Legacy binlogs do not store entries_num in their header: patchEntryNumber recovers the count by reading payloads ("legacy segment binlog found... header does not store entry numb") and writes it back only to index binlogs, and only during full load — it is never invoked in the ReopenSegments path. So the insert binlogs in the reopen SegmentLoadInfo still carry entries_num==0, the per-field sum is 0, and the assert false-rejects a valid reopen even though the authoritative segment num_of_rows already matched (checked by the first assert at line 80). Treat entries_num==0 as unknown and skip the per-field consistency loop, relying on the segment-level num_of_rows guard.

internal/querynodev2/segments/segment.go
```diff
@@ -1288,28 +1356,38 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
...
-	schema, schemaVersion := s.collection.SchemaAndSegcoreVersion()
-	err := s.csegment.Reopen(ctx, &segcore.ReopenRequest{
-		LoadInfo:      newLoadInfo,
+	err = s.csegment.Reopen(ctx, &segcore.ReopenRequest{
+		LoadInfo:      projectedLoadInfo,
 		Schema:        schema,
 		SchemaVersion: schemaVersion,
 	})
 	if err != nil {
+		if errors.Is(err, merr.ErrNeedFullSegmentReplacement) {
```
buqian-zilliz, line 1378: **[P1] Handle the full-replacement control signal**

The new control error is only wrapped and returned. `segmentLoader.ReopenSegments` propagates it, and no production QueryCoord or QueryNode path matches `ErrNeedFullSegmentReplacement` to issue a full load or atomic segment swap. Therefore valid transitions that segcore intentionally rejects in place, including row-count or storage-kind changes and prebuilt text-index replacement, leave the old segment installed and the reopen action failing. Handle this sentinel by scheduling or performing full segment replacement instead of treating it as a terminal reopen error.

internal/core/src/segcore/SegmentLoadInfo.cpp
```diff
@@ -75,6 +234,55 @@ SegmentLoadInfo::HasManifestColumn(const std::string& column_name) const {
...
+                  GetSegmentID(),
+                  expected,
+                  GetNumOfRows());
+    }
+
+    if (HasManifestPath()) {
+        return;
+    }
+
+    for (const auto& field_binlog : GetBinlogPaths()) {
```
buqian-zilliz, line 253: **[P1] Validate index row counts before reopen**

`ValidateReopenRowCount` validates the top-level count and binlog totals but never examines `FieldIndexInfo.num_rows`. A reopen can keep the top-level row count at N, replace an index whose populated metadata and files describe M rows, and still publish because runtime validation compares only `RuntimeResourceState::row_count` with the top-level load info. Such an index can produce offsets outside delete or timestamp bitmaps, or omit rows. Validate every retained index with a positive `num_rows` against `expected` before diff computation or index I/O.

internal/querynodev2/segments/segment.go
```diff
@@ -1288,28 +1356,38 @@ func (s *LocalSegment) Reopen(ctx context.Context, newLoadInfo *querypb.SegmentL
 	}
 	defer s.ptrLock.Unpin()
 
+	schema, schemaVersion := s.collection.SchemaAndSegcoreVersion()
+	projectedLoadInfo, err := projectSegmentLoadInfoToSchema(newLoadInfo, schema)
```
buqian-zilliz, line 1360: **[P2] Project reopen metadata before resource admission**

This projection runs inside `LocalSegment.Reopen`, but `segmentLoader.ReopenSegments` has already called `requestResource` on the original infos. The estimator still charges dropped-field stats and text data and, with tiered eviction disabled, a binlog group containing only dropped children retains its default vector flags and is charged in full. A valid field-drop reopen can therefore fail resource admission before canonical metadata reaches segcore. Project the reopen infos before `prepare` and `requestResource`, as the full `Load` path now does.

internal/querynodev2/segments/segment_loader.go
```diff
@@ -239,9 +239,17 @@ func (loader *segmentLoader) Load(ctx context.Context,
...
 	for _, segment := range segments {
-		configureUseTakeForOutput(segment, collection.Schema())
+		projected, err := projectSegmentLoadInfoToSchema(segment, schema)
+		if err != nil {
+			return nil, merr.Wrapf(err, "failed to project load metadata for segment %d", segment.GetSegmentID())
+		}
+		configureUseTakeForOutput(projected, schema)
+		projectedSegments = append(projectedSegments, projected)
 	}
+	segments = projectedSegments
```
buqian-zilliz, line 252: **[P2] Keep delegator post-load metadata canonical**

`segments = projectedSegments` only changes the worker-side local slice. In the remote full-load path, the shard delegator still owns the original `req.Infos`; after the worker succeeds it synchronizes the new schema and passes those unprojected infos to `loadBM25Stats`. Dropped BM25 fields can therefore be downloaded and registered in the IDF oracle even though the worker segment pruned them. Propagate the projected infos to delegator post-load processing, or filter BM25 resolution against the current schema there.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -7344,12 +7597,67 @@ ChunkedSegmentSealedImpl::PrepareLoadDiffForReopen(
...
+            // Make the cloned runtime describe the target source set before
+            // choosing whether to rebuild from a field column or scalar-index
+            // raw data. Pure drops normally happen in FinalizeLoadDiff, which
+            // is too late for source selection here.
+            if (diff.field_data_to_drop.erase(field_id) > 0) {
+                committer.Commit([&](RuntimeResourceState& runtime,
+                                     PublishedSegmentState& staged_state) {
+                    bool drop_binlog_index = get_bit_if_present(
+                        staged_state.binlog_index_bitset, field_id);
+                    DropFieldData(field_id, schema_snapshot, &runtime);
```
buqian-zilliz, line 7617: **[P2] Defer column warmup cancellation until publication**

Calling `DropFieldData` during preparation immediately invokes `CancelWarmup()` on the old column even though that column still belongs to the currently published snapshot. `BuildRawTextIndexWithSchema` and the later cancellation, validation, or publish-gate steps can still fail; in that case the old snapshot remains live, but its one-shot asynchronous warmup has been canceled and no failure path restarts it, leaving subsequent reads to load remaining cells on demand. Retain removed columns as staged retirement state and cancel their warmup only after publication succeeds.

Issue #51842, opened 2026-07-24
Growing nullable validity ignores reserved_offset for out-of-order concurrent inserts


liliu-z:
## Problem

`SegmentGrowingImpl::Insert` writes timestamps, row IDs, and field values into the range identified by `reserved_offset`, but nullable validity data is still appended at `ThreadSafeValidData::length_`:

```cpp
insert_record_.get_valid_data(field_id)->set_data_raw(
    num_rows, &insert_record_proto->fields_data(data_offset), field_meta);
```

`ThreadSafeValidData::set_data_raw` then writes at `length_` and increments it, without receiving `reserved_offset`.

This violates the documented growing-segment contract: `Insert(reservedOffset, ...)` may execute concurrently and must write into `[reservedOffset, reservedOffset + size)`.

This is pre-existing behavior, discovered while reviewing #51678; it is not introduced or worsened by that PR. The normal QueryNode ingestion path currently serializes writes to the same segment, so production exposure is limited, but direct concurrent callers or future same-segment ingestion parallelism can trigger it.

## Deterministic reproduction

Use a schema containing a nullable field and two distinguishable validity patterns:

1. Reserve A: `PreInsert(2) -> 0`, expected range `[0, 2)`.
2. Reserve B: `PreInsert(2) -> 2`, expected range `[2, 4)`.
3. Let A have validity `[true, false]`.
4. Let B have validity `[false, true]`.
5. Execute `Insert(B)` before `Insert(A)`.

Field values are written by reserved offset:

```text
values   = [A, B]
```

Validity is appended by completion order:

```text
actual   = [B.validity, A.validity]
expected = [A.validity, B.validity]
```

Nullable scalar paths may silently associate NULL markers with the wrong rows. Some nullable vector mapping-storage paths may assert when a later reserved range tries to read validity offsets beyond the current `length_`.

`AckResponder` prevents partially completed ranges from becoming visible, but it cannot reorder a validity bitmap that was already written into the wrong logical range.

## Suggested direction

- Pass `reserved_offset` into the growing validity write path.
- Allocate through `reserved_offset + num_rows`.
- Write validity into `[reserved_offset, reserved_offset + num_rows)`.
- Make the high-watermark/publication invariant explicit; segment visibility can continue to rely on `AckResponder`.

## Acceptance criteria

- Add a deterministic nullable test that reserves A then B, executes B before A, and verifies value/validity alignment after both are acknowledged.
- Cover at least one nullable scalar field.
[...]

PR #51843, opened 2026-07-24
fix: honor reserved offset for growing nullable validity

liliu-z:
## What this fixes

`PreInsert` assigns each growing insert a logical range. Field values, row IDs, and timestamps honor `reserved_offset`, but nullable validity previously appended by insert completion order. If a later reservation completes first, NULL markers can attach to the wrong scalar rows, and nullable vector mapping storage can read beyond the initialized validity range.

## Changes

- Pass `reserved_offset` into growing validity writes.
- Reserve through `reserved_offset + num_rows` and write into that logical range.
- Keep the initialized high-watermark monotonic while `AckResponder` remains responsible for publishing the contiguous visible prefix.
- Preserve the fixed-chunk, pointer-stable validity storage introduced by #51678.
- Add deterministic scalar and nullable-vector regressions for reverse insert completion order.

## Scope

This PR fixes logical validity-range alignment for out-of-order insert completion. It does not claim to solve the separate pre-existing physical-offset allocation race between truly simultaneous nullable-vector mapping writers.

## Testing

- `Growing.OutOfOrderInsertKeepsNullable*`: 2/2 passed
- `Growing.*`: 19/19 passed
- clang-format 15 dry-run and `git diff --check` passed

issue: #51842
related: #51678

### DIFF
--- a/internal/core/src/segcore/ConcurrentVector.h
+++ b/internal/core/src/segcore/ConcurrentVector.h
@@ -61,27 +61,39 @@ class ThreadSafeValidData {
     }
 
     void
-    set_data_raw(const std::vector<FieldDataPtr>& datas) {
+    set_data_raw(size_t offset, const std::vector<FieldDataPtr>& datas) {
         std::unique_lock<std::shared_mutex> lck(mutex_);
+        size_t total = 0;
+        for (auto& field_data : datas) {
+            total += field_data->get_num_rows();
+        }
+        if (total == 0) {
+            return;
+        }
+        reserve_to(offset + total);
+
+        auto write_offset = offset;
         for (auto& field_data : datas) {
             auto num_row = field_data->get_num_rows();
-            reserve_to(length_ + num_row);
-            write_bits(length_, num_row, [&field_data](size_t i) {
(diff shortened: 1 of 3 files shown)


liliu-z:
Superseded by #51960, which landed the same fix for the `Insert` path — growing nullable validity now writes into the range `PreInsert` reserved, using a no-gap assert at the write site rather than the hole-tolerant high-watermark this PR proposed. #51960's approach is simpler and it also handles the `Reopen` backfill retry explicitly.

The linked issue #51842 is about the `Insert` path specifically, so it is covered by #51960 as well.

One inconsistency was left behind: inside `load_field_data_common`, column data, the interim index, timestamps and row IDs are all addressed by `reserved_offset`, while validity still appends by arrival order (the `FieldData` overload of `set_data_raw` never got the parameter). That is a behavioral no-op today — a growing segment is loaded before it is published and always at offset 0 — so it is a consistency cleanup, not a fix. Split out into #52637 rather than force-pushing this branch, since the discussion here is about a different change.

Closing rather than rebasing: this branch is 168 commits behind and its commit is fully contained in #51960.

Issue #51649, opened 2026-07-21
Enhance[non-blocking-3.0]: bump_schema_version enhancement points


MrPresent-Han:
Tracker for **non-blocking** enhancement / optimization points found while reviewing the `bump_schema_version_compaction` feature (DataCoord + DataNode).

> [!IMPORTANT]
> **NOT a Milvus 3.0 release blocker.** No item here has been confirmed to cause silent data corruption or source-data loss. A few narrow schema-evolution cases can fail the task or panic, while the remaining items are stability, robustness, efficiency, or cleanup enhancements. Items may be promoted to release blockers only if the affected schema-evolution path is enabled as a supported GA workflow.

**Severity legend** — every item is tagged so priority is unambiguous:
- 🟡 **feature hardening** — a narrow supported schema-evolution sequence can deterministically fail or panic; no silent data loss.
- 🟠 **stability / robustness** — no data risk, but can overload a DataNode or turn an internal error into a task failure/panic; worth doing, still non-blocking.
- 🟢 **efficiency / harmless** — wasted CPU/IO, memory retention, or churn only; data is always correct; safe to defer.
- ⚪ **cleanup** — dead / redundant code; no runtime effect at all.

---

## E1 — Size-aware slot for bump compaction  🟠 stability

**Problem.** `bumpSchemaVersionTask.GetTaskSlot()` (`internal/datacoord/compaction_task_bump_schema_version.go:77`) returns a **flat** `BumpSchemaVersionCompactionSlotUsage` (default **1**) for **every** segment, regardless of size or path. The DataNode picks one of three paths at execution:

| path | trigger | cost |
|------|---------|------|
| pure-bump (metadata only) | no missing function, no dropped field | cheap |
| missing-function materialize | function output field missing (BM25/MinHash) | medium (full-segment) |
| full schema rewrite | dropped field(s) | heavy (full-segment read + write) |

The heavy paths cost like a `SortCompaction` (full read+write scaling with segment size) but are charged **1 slot**. Slot is the **only** DataNode admission gate — `internal/datanode/services.go:618-637` (`availableSlots = totalSlots - compactionUsed - ...`, no separate memory watermark). So a mass schema change (drop-field → full-rewrite, or add-function-field → materialize) over a large collection with many big segments can over-admit heavy tasks → **DataNode memory pressure / OOM / saturation**.

**Nature.** Availability / stability risk, **not** a correctness bug (worst case = overloaded DataNode; data stays correct).

[...]

Issue #51729, opened 2026-07-23
[Bug]: Schema-bump materialization undercounts Statistics for recovered StorageV3 input segments


tedxu:
Follow-up from PR #50410 review by @congqixia.

runMissingFunctionMaterialization builds the output Statistics from
mergedInsertLogs and segment.GetDeltalogs()
(internal/datanode/compactor/bump_schema_version_compactor.go ~:1150,
stats.go ~:42 -> BuildStatsFromFieldBinlogs).

After PR #50410 the compaction plan for a recovered StorageV3 input
segment ships empty FieldBinlogs/Deltalogs (per-field KVs are skipped),
so the computed Stats undercounts InsertBinlogSize/DeltaBinlogSize/
DeleteNumRows. The receiver (meta.go in-place branch) adopts this
non-nil computed Stats, making the undercount durable. The manifest FFI
does not expose byte MemorySize, and CompactionSegmentBinlogs has no
Stats field.

Fix direction: ship the input segment's SegmentInfo.Stats in the
compaction plan and preserve it, applying only the partial-writer deltas
(also affects the V2 partial-writer path and runFullSchemaRewrite).

PR #51848, opened 2026-07-25
enhance: harden compaction logic against potential risks

MrPresent-Han:
### What

Hardens the `bump_schema_version` / `add_function_field` compaction path against a set of potential risks identified during a deep review, and rounds out ref file-resource (RefMode) handling. All items strengthen existing logic; wire protocols and default behaviors are unchanged.

> **Note:** the RecordReader ownership / borrow-contract portion originally in this PR has been split out into #51884 (master) / #51891 (3.0) and is no longer part of this diff.

### Enhanced items

**RefMode handling**
- **datacoord** — carry the schema's analyzer `FileResources` into the bump compaction plan so the inline text-match index build can resolve them under ref mode, mirroring sort/mix compaction.
- **rootcoord** — reject `add_function_field` when the added function's input-field analyzer references a file resource: the output backfill runs that analyzer inside the compaction record-materializer, which cannot resolve a referenced resource under ref mode, so the combination is intentionally left unsupported for now. The check is scoped to the function's input field, so collections that merely have other resource-referencing analyzer fields are unaffected.

**Materializer robustness**
- decide selection-column and source-TTL presence from `existingFields` instead of probing records — V2/V3 records are strict about `Column` on absent fields.
- build selection columns eagerly so a build failure surfaces with its root cause instead of deep inside the writer.
- fill missing schema fields before running functions so newly added inputs are readable by function runners.
- skip the analyzer file-resource download when no field enables match (that download exists solely to build text-match indexes).
- drop an unused parameter in the full-rewrite path.

**Function runner robustness**
- **MinHash** — keep the derived `num_hashes` runner-local instead of writing it back into the shared plan schema, which clustering reads from parallel runner constructions.
- **multi-analyzer BM25** — DDL validates alias syntax while preserving legacy dangling aliases for compatibility; at runtime unresolved analyzer names fall back to the default analyzer, an unresolvable default returns an error instead of recursing forever, and a constructor warning surfaces undefined alias targets.

### Tests

- Full `internal/datanode/compactor` suite green under `-tags dynamic,test`; targeted `internal/storage` and `internal/datanode/external` suites green.
- New regression tests proven to catch the previous behavior: selection over a record lacking a schema field, and a full rewrite whose source segment lacks the designated TTL field.
- storage / function / rootcoord unit tests green.

issue: #51649

### DIFF
--- a/internal/datacoord/compaction_task_bump_schema_version.go
+++ b/internal/datacoord/compaction_task_bump_schema_version.go
@@ -26,15 +26,18 @@ import (
 	"go.uber.org/atomic"
 	"google.golang.org/protobuf/proto"
 
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/compaction"
 	"github.com/milvus-io/milvus/internal/datacoord/allocator"
 	"github.com/milvus-io/milvus/internal/datacoord/session"
 	"github.com/milvus-io/milvus/internal/metastore/kv/binlog"
+	"github.com/milvus-io/milvus/internal/util/fileresource"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
 	"github.com/milvus-io/milvus/pkg/v3/taskcommon"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
+	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
 
 var _ CompactionTask = (*bumpSchemaVersionTask)(nil)
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
 		JsonParams:                compactionParams,
 		CurrentScalarIndexVersion: t.ievm.ResolveScalarIndexVersion(),
 	}
+
+	// set analyzer resource for text match index if use ref mode.
+	// Only the full-rewrite path consumes these (inline createTextIndex, mirroring
+	// sort/mix); which path a segment takes is decided on the datanode, so attach
+	// best-effort: on failure keep building the plan without resources instead of
+	// failing tasks (bump-only / partial backfill) that never need them. A
+	// full-rewrite that does need them fails at createTextIndex resource
+	// resolution, next to the actual consumer.
+	if fileresource.IsRefMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue()) &&
+		schemaNeedsTextIndex(taskProto.GetSchema()) &&
+		len(taskProto.GetSchema().GetFileResourceIds()) > 0 {
+		resources, err := t.meta.GetFileResources(context.Background(), taskProto.GetSchema().GetFileResourceIds()...)
+		if err != nil {
+			mlog.Warn(context.TODO(), "get file resources for schema bump compaction failed, building plan without them",
+				mlog.Int64("collectionID", taskProto.GetCollectionID()), mlog.Err(err))
+		} else {
+			plan.FileResources = resources
+		}
+	}
+
 	segments := make([]*SegmentInfo, 0, len(taskProto.GetInputSegments()))
 	for _, segID := range taskProto.GetInputSegments() {
 		segInfo := t.meta.GetHealthySegment(context.TODO(), segID)
@@ -145,6 +168,15 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
 	plan.BeginLogID = logIDRange.Begin
 	WrapPluginContext(taskProto.GetCollectionID(), taskProto.GetSchema().GetProperties(), plan)
 	return plan, nil
+}
+
+func schemaNeedsTextIndex(schema *schemapb.CollectionSchema) bool {
+	for _, field := range schema.GetFields() {
+		if typeutil.CreateFieldSchemaHelper(field).EnableMatch() {
+			return true
+		}
+	}
+	return false
 }
 
 func (t *bumpSchemaVersionTask) GetSlotUsage() int64 {
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -131,19 +131,19 @@ func (t *bumpSchemaVersionCompactionTask) preCompact() error {
 	return nil
 }
 
-func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunctions []*schemapb.FunctionSchema) (*schemapb.CollectionSchema, []int64, error) {
+func (t *bumpSchemaVersionCompactionTask) buildFunctionBackfillReadSchema(missingFunctions []*schemapb.FunctionSchema, sourceFields map[int64]struct{}) (*schemapb.CollectionSchema, []int64, error) {
 	schema := t.plan.GetSchema()
 	seen := make(map[int64]struct{})
-	var fields []*schemapb.FieldSchema
-	var fieldIDs []int64
-	addInputField := func(field *schemapb.FieldSchema) {
+	var readFields []*schemapb.FieldSchema
+	var readFieldIDs []int64
+	addReadField := func(field *schemapb.FieldSchema) {
 		fieldID := field.GetFieldID()
 		if _, ok := seen[fieldID]; ok {
 			return
 		}
 		seen[fieldID] = struct{}{}
-		fields = append(fields, field)
-		fieldIDs = append(fieldIDs, fieldID)
+		readFields = append(readFields, field)
+		readFieldIDs = append(readFieldIDs, fieldID)
 	}
 	for _, functionSchema := range missingFunctions {
 		if err := validateSupportedMissingFunctionMaterialization(functionSchema); err != nil {
@@ -157,39 +157,59 @@ func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunc
 			if err := validateMaterializationInputField(functionSchema, inputField); err != nil {
 				return nil, nil, err
 			}
-			addInputField(inputField)
+			addReadField(inputField)
 
-			additionalFields, err := additionalFunctionInputFields(schema, functionSchema, inputField)
+			implicitFields, err := implicitFunctionInputFields(schema, functionSchema, inputField)
 			if err != nil {
 				return nil, nil, err
 			}
-			for _, additionalField := range additionalFields {
-				if err := validateMaterializationInputField(functionSchema, additionalField); err != nil {
+			for _, implicitField := range implicitFields {
+				if err := validateMaterializationInputField(functionSchema, implicitField); err != nil {
 					return nil, nil, err
 				}
-				addInputField(additionalField)
+				addReadField(implicitField)
+			}
+		}
+	}
+	// If no explicit or implicit input exists physically, the filtered read
+	// schema would be empty. Add one system column as a row-count anchor.
+	hasReadableSourceField := false
+	for _, fieldID := range readFieldIDs {
+		if _, ok := sourceFields[fieldID]; ok {
+			hasReadableSourceField = true
+			break
+		}
+	}
+	if !hasReadableSourceField {
+		for _, fieldID := range []int64{common.RowIDField, common.TimeStampField} {
+			if _, ok := sourceFields[fieldID]; !ok {
+				continue
+			}
+			if field := typeutil.GetField(schema, fieldID); field != nil {
+				addReadField(field)
+				break
 			}
 		}
 	}
 	return &schemapb.CollectionSchema{
 		Name:               schema.GetName(),
 		Description:        schema.GetDescription(),
-		Fields:             fields,
+		Fields:             readFields,
 		EnableDynamicField: schema.GetEnableDynamicField(),
 		Properties:         schema.GetProperties(),
-	}, fieldIDs, nil
+	}, readFieldIDs, nil
 }
 
-func additionalFunctionInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
+func implicitFunctionInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
 	switch functionSchema.GetType() {
 	case schemapb.FunctionType_BM25:
-		return bm25AdditionalInputFields(schema, inputField)
+		return bm25ImplicitInputFields(schema, functionSchema, inputField)
 	default:
 		return nil, nil
 	}
 }
 
-func bm25AdditionalInputFields(schema *schemapb.CollectionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
+func bm25ImplicitInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
 	params, ok := typeutil.CreateFieldSchemaHelper(inputField).GetMultiAnalyzerParams()
 	if !ok {
 		return nil, nil
@@ -198,14 +218,14 @@ func bm25AdditionalInputFields(schema *schemapb.CollectionSchema, inputField *sc
 		ByField string `json:"by_field"`
 	}
 	if err := json.Unmarshal([]byte(params), &multiAnalyzerParams); err != nil {
-		return nil, err
+		return nil, merr.WrapErrDataIntegrity(err, "failed to parse multi_analyzer_params for function %s", functionSchema.GetName())
 	}
 	if multiAnalyzerParams.ByField == "" {
-		return nil, merr.WrapErrParameterInvalidMsg("multi_analyzer_params missing required 'by_field' key")
+		return nil, merr.WrapErrDataIntegrityMsg("multi_analyzer_params for function %s is missing required 'by_field' key", functionSchema.GetName())
 	}
 	byField := typeutil.GetFieldByName(schema, multiAnalyzerParams.ByField)
 	if byField == nil {
-		return nil, merr.WrapErrParameterInvalidMsg("input field not found in schema")
+		return nil, merr.WrapErrDataIntegrityMsg("multi_analyzer_params for function %s references missing input field %s", functionSchema.GetName(), multiAnalyzerParams.ByField)
 	}
 	return []*schemapb.FieldSchema{byField}, nil
 }
@@ -288,16 +308,6 @@ func validateMaterializationOutputField(functionSchema *schemapb.FunctionSchema,
 	return nil
 }
 
-func partialMaterializerExistingFields(schema *schemapb.CollectionSchema, missingFunctions []*schemapb.FunctionSchema, existingFields map[int64]struct{}) map[int64]struct{} {
-	fields := collectionSchemaFields(schema)
-	for _, functionSchema := range missingFunctions {
-		for _, outputIndex := range functionOutputIndexesToMaterialize(functionSchema, existingFields) {
-			delete(fields, functionSchema.GetOutputFieldIds()[outputIndex])
-		}
-	}
-	return fields
-}
-
 func (t *bumpSchemaVersionCompactionTask) openRecordReader(segment *datapb.CompactionSegmentBinlogs, schema *schemapb.CollectionSchema) (storage.RecordReader, map[int64]struct{}, error) {
 	reader, existingFields, err := newCompactionSegmentRecordReader(t.ctx, segment, schema, t.compactionParams.StorageConfig,
 		storage.WithCollectionID(segment.GetCollectionID()),
@@ -328,35 +338,35 @@ func (t *bumpSchemaVersionCompactionTask) fullRewriteSegmentID() (int64, error) 
 	return idRange.GetBegin(), nil
 }
 
-func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchema, entityFilter compaction.EntityFilter, ttlFieldID int64, useTTLField bool, ttlValues []int64) (*recordSelection, []int64, error) {
+func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchema, entityFilter compaction.EntityFilter, ttlFieldID int64, useTTLField bool, ttlDefault int64, useTTLDefault bool) (*recordSelection, error) {
 	pkArray := record.Column(pkField.GetFieldID())
 	var pkAt func(int) any
 	switch pkField.GetDataType() {
 	case schemapb.DataType_Int64:
 		int64Array, ok := pkArray.(*array.Int64)
 		if !ok {
-			return nil, nil, merr.WrapErrServiceInternal("int64 primary key field not found in full schema rewrite record")
+			return nil, merr.WrapErrServiceInternal("int64 primary key field not found in full schema rewrite record")
 		}
 		pkAt = func(i int) any { return int64Array.Value(i) }
 	case schemapb.DataType_VarChar:
 		stringArray, ok := pkArray.(*array.String)
 		if !ok {
-			return nil, nil, merr.WrapErrServiceInternal("varchar primary key field not found in full schema rewrite record")
+			return nil, merr.WrapErrServiceInternal("varchar primary key field not found in full schema rewrite record")
 		}
 		pkAt = func(i int) any { return stringArray.Value(i) }
 	default:
-		return nil, nil, merr.WrapErrServiceInternal("invalid primary key data type for full schema rewrite")
+		return nil, merr.WrapErrServiceInternal("invalid primary key data type for full schema rewrite")
 	}
 
 	timestampArray, ok := record.Column(common.TimeStampField).(*array.Int64)
 	if !ok {
-		return nil, nil, merr.WrapErrServiceInternal("timestamp field not found in full schema rewrite record")
+		return nil, merr.WrapErrServiceInternal("timestamp field not found in full schema rewrite record")
 	}
 	var ttlArray *array.Int64
 	if useTTLField {
 		ttlArray, ok = record.Column(ttlFieldID).(*array.Int64)
 		if !ok {
-			return nil, nil, merr.WrapErrServiceInternal("TTL field not found in full schema rewrite record")
+			return nil, merr.WrapErrServiceInternal("TTL field not found in full schema rewrite record")
 		}
 	}
 
@@ -367,6 +377,8 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
 		expireTs := int64(-1)
 		if useTTLField && ttlArray.IsValid(i) {
 			expireTs = ttlArray.Value(i)
+		} else if !useTTLField && useTTLDefault {
+			expireTs = ttlDefault
 		}
 		if entityFilter.Filtered(pkAt(i), typeutil.Timestamp(timestampArray.Value(i)), expireTs) {
 			if sliceStart != -1 {
@@ -378,9 +390,6 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
 			continue
 		}
 
-		if useTTLField && expireTs > 0 {
-			ttlValues = append(ttlValues, expireTs)
-		}
 		if sliceStart == -1 {
 			sliceStart = i
 		}
@@ -390,9 +399,24 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
 		selection.length += record.Len() - sliceStart
 	}
 	if filteredRows == 0 {
-		return nil, ttlValues, nil
+		return nil, nil
 	}
-	return selection, ttlValues, nil
+	return selection, nil
+}
+
+func fullRewriteTTLDefault(schema *schemapb.CollectionSchema, ttlFieldID int64) (int64, bool, error) {
+	if ttlFieldID < common.StartOfUserFieldID {
+		return 0, false, nil
+	}
+	ttlField := typeutil.GetField(schema, ttlFieldID)
+	if ttlField == nil || ttlField.GetDefaultValue() == nil {
+		return 0, false, nil
+	}
+	value, ok := ttlField.GetDefaultValue().GetData().(*schemapb.ValueField_TimestamptzData)
+	if !ok {
+		return 0, false, merr.WrapErrDataIntegrityMsg("TTL field %d has a non-timestamptz default value", ttlFieldID)
+	}
+	return value.TimestamptzData, true, nil
 }
 
 func (t *bumpSchemaVersionCompactionTask) runSchemaVersionBumpOnly() *datapb.CompactionPlanResult {
@@ -425,6 +449,7 @@ func (t *bumpSchemaVersionCompactionTask) runSchemaVersionBumpOnly() *datapb.Com
 }
 
 func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields map[int64]struct{}) (*datapb.CompactionPlanResult, error) {
+	// 1. Set up source filtering, reading, and materialization.
 	segment := t.plan.GetSegmentBinlogs()[0]
 	collectionID := segment.GetCollectionID()
 	newSegmentID, err := t.fullRewriteSegmentID()
@@ -461,10 +486,8 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 	}
 	defer materializer.Close()
 
-	// Prepare LOB (TEXT) handling: a schema bump keeps existing TEXT LOB data
-	// unchanged, so REUSE_ALL — the writer preserves the encoded LOB reference
-	// bytes (WithTextRefsAsBinary) and the source LOB files are merged into the
-	// output manifest afterwards (applyLOBCompaction). Mirrors sort compaction.
+	// 2. Set up LOB reuse and the output writer.
+	// Existing TEXT LOB references stay unchanged in REUSE_ALL mode.
 	if err := t.initLOBCompactionContext(t.ctx); err != nil {
 		return nil, err
 	}
@@ -504,6 +527,16 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		}
 	}()
 
+	// 3. Filter, materialize, and rewrite source records.
+	// Source TTL presence is decided from existingFields, never by probing the
+	// record: V2/V3 records panic on Column for a field they do not carry.
+	_, sourceHasTTL := existingFields[ttlFieldID]
+	sourceHasTTLField := ttlFieldID >= common.StartOfUserFieldID && sourceHasTTL
+	ttlDefault, hasTTLDefault, err := fullRewriteTTLDefault(t.plan.GetSchema(), ttlFieldID)
+	if err != nil {
+		return nil, err
+	}
+	preMaterializeFilter := len(delta) > 0 || t.plan.GetCollectionTtl() > 0 || sourceHasTTLField || hasTTLDefault
 	var totalRows int64
 	for {
 		record, err := reader.Next()
@@ -514,11 +547,9 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 			return nil, err
 		}
 
-		sourceHasTTLField := ttlFieldID >= common.StartOfUserFieldID && record.Column(ttlFieldID) != nil
-		preMaterializeFilter := len(delta) > 0 || t.plan.GetCollectionTtl() > 0 || sourceHasTTLField
 		var selection *recordSelection
 		if preMaterializeFilter {
-			selection, _, err = selectFullRewriteRecord(record, pkField, entityFilter, ttlFieldID, sourceHasTTLField, nil)
+			selection, err = selectFullRewriteRecord(record, pkField, entityFilter, ttlFieldID, sourceHasTTLField, ttlDefault, hasTTLDefault)
 			if err != nil {
 				return nil, err
 			}
@@ -549,6 +580,7 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		cleanupMaterializedRecord(wrapped)
 	}
 
+	// 4. Finalize writer output and assemble the rewritten segment.
 	if err := writer.Close(); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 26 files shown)


internal/datacoord/compaction_task_bump_schema_version.go
```diff
@@ -113,6 +114,21 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
...
 		CurrentScalarIndexVersion: t.ievm.GetCurrentScalarIndexEngineVersion(),
 	}
+
+	// set analyzer resource for text match index if use ref mode.
+	// bump full-rewrite builds text indexes inline (createTextIndex) and needs the analyzer
+	// resources, mirroring sort/mix compaction; without them ref mode cannot resolve the
+	// referenced analyzer resource and the compaction fails.
+	if fileresource.IsRefMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue()) &&
+		len(taskProto.GetSchema().GetFileResourceIds()) > 0 {
+		resources, err := t.meta.GetFileResources(context.Background(), taskProto.GetSchema().GetFileResourceIds()...)
```
liliu-z, line 130: Under ref mode this new block calls `t.meta.GetFileResources`, which reads datacoord's `resourceIDMap` — a map written only by `UpdateFileResources`, which has no production caller (`reloadFromKV` never loads it; the datacoord catalog's `SaveFileResource`/`ListFileResource` are never invoked; the file-resource observer syncs rootcoord's meta to leaf nodes, not datacoord's). In production the lookup therefore always returns `file resource N not found` and ref-mode bump compaction fails at plan build; the added test passes only because it seeds the map by hand via `UpdateFileResources`. This is a pre-existing systemic gap that mix (`compaction_task_mix.go:378`) and clustering (`compaction_task_clustering.go:378`) already share and this PR extends to the bump path (note: sort compaction does not issue this call, contrary to the original wording) — it needs a real production populator for `resourceIDMap`.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -187,6 +190,27 @@ func (c *Core) broadcastAlterCollectionSchemaAdd(ctx context.Context, broadcaste
...
+func (c *Core) rejectAddFunctionInputAnalyzerFileResource(ctx context.Context, schema *schemapb.CollectionSchema, function *schemapb.FunctionSchema) error {
+	analyzerInfos, err := collectFunctionInputAnalyzerInfos(schema, function)
+	if err != nil {
+		return err
+	}
+	resourceIds, err := c.validateAnalyzerInfos(ctx, analyzerInfos)
+	if err != nil {
+		return err
+	}
+	if len(resourceIds) > 0 {
```
liliu-z: `rejectAddFunctionInputAnalyzerFileResource` rejects whenever the input analyzer references a file resource with no file-resource-mode gate, even though its own comment scopes the problem to ref mode. In the default sync mode the resources are pre-synced to datanodes and registered in the global analyzer map (`SyncManager.Sync` -> `analyzer.UpdateGlobalResourceInfo`), so the backfill materializer's `NewAnalyzer(params, "")` resolves them fine; only ref mode (`RefManager`, which never populates the global map and only downloads per-call for the text-index build) genuinely cannot. As written this blocks valid sync-mode function-field additions outright; gate the rejection on ref mode, mirroring the datacoord side.

internal/rootcoord/create_collection_task.go
```diff
@@ -928,6 +928,15 @@ func validateMultiAnalyzerParams(params string, coll *schemapb.CollectionSchema,
 	if !hasDefault {
 		return merr.WrapErrParameterInvalidMsg("multi analyzer must set default analyzer for all unknown value")
 	}
+
+	// Every alias must target a defined analyzer. A dangling target would either
+	// silently tokenize with the default analyzer (wrong BM25 output) or, when
+	// "default" itself is aliased to a missing name, recurse forever at runtime.
+	for aliasName, target := range aliasMapping {
```
liliu-z: The new dangling-alias check in `validateMultiAnalyzerParams` runs over the whole schema: `collectAnalyzerInfos` iterates every field, and both the add path (`ddl_callbacks_alter_collection_schema.go:160`) and the drop path (`:424`) reach it via `prepareAlterCollectionAnalyzerFileResources` -> `validateSchemaAnalyzerFileResources`. A legacy collection carrying a runtime-tolerated dangling alias (the runner now only warns and falls back to default) can therefore no longer add, alter, or drop any field, because unrelated DDL re-trips strict validation on pre-existing config. This makes the DDL-time check inconsistent with the runtime tolerance the same PR adds. Restrict the strict alias check to newly added/modified analyzer configs rather than the full existing schema. (The original issue cites internal/proxy/create_collection_task.go, which does not exist; the code is in internal/rootcoord/.)

internal/rootcoord/create_collection_task.go
```diff
@@ -941,6 +950,23 @@ func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.Ana
...
+// input fields only — a subset of collectAnalyzerInfos used to decide whether an
+// added function's input analyzer references a file resource.
+func collectFunctionInputAnalyzerInfos(collSchema *schemapb.CollectionSchema, function *schemapb.FunctionSchema) ([]*querypb.AnalyzerInfo, error) {
+	inputIDs := typeutil.NewSet(function.GetInputFieldIds()...)
+	analyzerInfos := make([]*querypb.AnalyzerInfo, 0)
+	for _, field := range collSchema.GetFields() {
+		if !inputIDs.Contain(field.GetFieldID()) {
+			continue
+		}
+		if err := validateAnalyzer(collSchema, field, &analyzerInfos); err != nil {
```
liliu-z, line 980: The new reject guard's `collectFunctionInputAnalyzerInfos` calls `validateAnalyzer`, which returns early for fields with `enable_analyzer=false` (create_collection_task.go:978). But MinHash — which `add_function_field` supports (alter_schema_add.go:151) — consumes the input field's `analyzer_params` unconditionally (minhash_function.go:98), so a MinHash input with `enable_analyzer=false` and a resource-referencing `analyzer_params` slips past the guard and then fails ref-mode backfill. Validate `analyzer_params` for MinHash inputs regardless of the `enable_analyzer` flag. Niche config, but it defeats the guard's purpose. (Original cites internal/proxy/create_collection_task.go:978, which is neither the right module nor inside a diff hunk; the in-diff guard call is at rootcoord line 963.)

internal/datacoord/compaction_task_bump_schema_version.go
MrPresent-Han: Confirmed the core issue: DataCoord receives the schema FileResourceIds, but its resourceIDMap has no production population or startup recovery path. UpdateFileResources is currently test-only, so the hand-seeded unit test proves only the Bump-side contract.

This is a pre-existing shared gap affecting Mix, Sort, namespace clustering, and text-index stats. I opened #51861 to track the production RootCoord -> DataCoord metadata propagation/recovery and asked @aoiasd to take ownership. I do not have repository permission to set the assignee metadata directly.

The boundary for this PR is intentionally narrower: assuming DataCoord FileResource meta is populated, BumpSchemaVersion copies the referenced FileResourceInfo entries into plan.FileResources so DataNode createTextIndex can Download/BuildExtraResourceInfo/Release them. This PR does not claim to fix the shared metadata population mechanism; RefMode paths requiring these resources remain blocked until #51861 is fixed.

One small correction: SortCompaction also reaches this lookup because DataCoord routes Sort and Mix through newMixCompactionTask, and the condition explicitly includes both SortCompaction and MixCompaction.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han: Confirmed. The current guard rejects every file-resource mode, while SyncMode pre-syncs the files and registers them in the global analyzer resource map, so NewAnalyzer(params, "") can resolve the references during backfill. RefMode does not populate that map, and CloseMode has no resource-resolution path either. I will narrow the guard to allow only SyncMode (!fileresource.IsSyncMode(mode) && len(resourceIds) > 0) and cover Sync/Ref/Close in tests.

internal/rootcoord/create_collection_task.go
MrPresent-Han: Confirmed. Revalidating every analyzer in the full schema turns a runtime-tolerated legacy dangling alias into a blocker for unrelated add/alter/drop-field DDL. We will preserve the legacy behavior rather than add a new-vs-existing validation branch: remove the new DDL-time dangling-target rejection, keep a warning plus default fallback for an ordinary dangling alias, and let a dangling default alias fail through the runtime terminal guard instead of recursing. The local tests now explicitly preserve DDL acceptance for both legacy cases.

internal/rootcoord/create_collection_task.go
MrPresent-Han: Confirmed. collectFunctionInputAnalyzerInfos currently reuses validateAnalyzer, whose enable_analyzer=false early return is correct for match/BM25 validation but wrong for this resource collector. MinHash is valid without enable_analyzer and still calls getAnalyzerParams plus NewAnalyzer unconditionally, so resource-backed analyzer_params can bypass the guard. I will make the collector function-aware: MinHash inputs will have analyzer_params collected and validated regardless of enable_analyzer, without changing the existing MinHash enable-analyzer semantics. Together with the mode fix, SyncMode remains allowed while RefMode and CloseMode reject this unsupported backfill configuration; a regression test will cover the disabled-analyzer MinHash case.

internal/util/function/multi_analyzer_bm25_function.go
```diff
@@ -140,6 +164,12 @@ func (v *MultiAnalyzerBM25FunctionRunner) getAnalyzer(name string, analyzers map
 		return analyzers[name], nil
 	}
 
+	// "default" is the terminal fallback. If even it cannot resolve (missing, or
+	// aliased to an undefined analyzer), fail instead of recursing forever. Guard
+	// on the pre-alias name so a dangling default alias cannot loop.
+	if origName == "default" {
```
liliu-z, line 189: The PR body claims "DDL now validates that every alias targets a defined analyzer," but validateMultiAnalyzerParams only checks the alias map is well-formed JSON and never verifies targets resolve; the PR's own new tests assert dangling aliases pass DDL. The real change is the getAnalyzer recursion guard here — it returns an error instead of recursing forever when even "default" cannot resolve — plus a constructor warning; dangling aliases still fall back to default for compatibility. Reword the PR body so a reviewer does not go verify a validation guarantee the code intentionally omits.

MrPresent-Han: Good catch. You are right: DDL intentionally accepts dangling alias targets for compatibility; the actual changes are the constructor warning and the terminal guard when the default analyzer cannot resolve. I have reworded the PR body to match the implemented behavior.

internal/datanode/compactor/record_materializer.go
```diff
@@ -96,56 +101,52 @@ func (m *RecordMaterializer) Wrap(rec storage.Record) (storage.Record, error) {
 
 func (m *RecordMaterializer) WrapWithSelection(rec storage.Record, selection *recordSelection) (storage.Record, error) {
 	base := rec
-	var selected *selectedRecord
 	if selection != nil {
-		selected = newSelectedRecord(rec, m.schema, selection)
+		selected, err := newSelectedRecord(rec, m.schema, m.existingFields, selection)
+		if err != nil {
+			return nil, err
```
liliu-z, line 172: `WrapWithSelection` now returns early on a `newSelectedRecord` error (`return nil, err`) without releasing `rec`, whereas its later error paths release `base` (which owns `rec`). The full-rewrite caller only calls `record.Release()` when `selection == nil` (bump_schema_version_compactor.go:581), trusting WrapWithSelection to own the record when a selection is present — so if `buildSelectedColumn` fails (e.g. a column type mismatch) with a non-nil selection, the source record (potentially cgo-backed) leaks on that error return. This is an error-path-only leak: the happy path releases via `releaseWrappedRecord` and the reader is `defer`-closed, so a blanket `defer record.Release()` would double-free; fix by making WrapWithSelection's release contract consistent across its error paths.

MrPresent-Han: Resolved by the rebase onto master after #51884 merged: `WrapWithSelection` now operates under the borrow contract — `rec` is reader-owned and released by the reader's next `Next()`/`Close()` (the full-rewrite loop defer-closes the reader), so the early error return correctly leaves `rec` untouched and the transfer-era leak this thread describes no longer exists. The eager-build error path itself is exercised by `TestRecordMaterializerWrapWithSelectionReturnsColumnBuildError`. (Note: this thread never auto-outdated because the rebase force-push orphaned the commit it was anchored to — GitHub keeps such comments current forever.)

internal/util/function/minhash_function.go
```diff
@@ -156,26 +160,22 @@ func NewMinHashFunctionRunner(
...
-		}
+		// Keep the derived value runner-local. funSchema is shared with the
+		// compaction plan and may be read by concurrent runner constructions.
 		numHashes = int(outputDim / 32)
-		funSchema.Params = append(funSchema.Params, &commonpb.KeyValuePair{
-			Key:   NumHashesKey,
-			Value: strconv.Itoa(numHashes),
-		})
+	}
+	if int64(numHashes)*32 != outputDim {
```
buqian-zilliz: **[P1] Validate num_hashes without overflowing the dimension check**

The new equality guard can overflow before it compares the values. On a 64-bit target, `num_hashes=576460752303423489` with output dim `32` makes `int64(numHashes)*32` wrap to `32`; the DDL validator uses the same overflowing multiplication, so the schema is accepted. `initializePermutations` then attempts impossibly large slice allocations, which can panic or exhaust the process. Compare the parsed `int64` directly with `outputDim/32` and enforce the allocation/native integer bounds before converting it to `int`.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -1130,15 +1180,12 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
...
-				// comes from the freshly committed V3 stats (tracked on
-				// writerResult); insert/delta aggregates come from the
-				// merged arrays.
+				// Missing function outputs add new stats keys, so the final
+				// footprint is the base manifest plus this task's new blobs.
 				Stats: buildCompactionOutputStats(
 					mergedInsertLogs,
 					segment.GetDeltalogs(),
-					writerResult.statsBlobSize,
+					baseStatsBlobSize+writerResult.statsBlobSize,
```
buqian-zilliz, line 1191: **[P2] Preserve all cumulative V3 statistics during partial backfill**

This change preserves only the base `StatsBinlogSize`. After a DataCoord restart, V3 segments intentionally reload with empty `FieldBinlogs` and `Deltalogs`, while the manifest and `SegmentInfo.Stats` remain authoritative. In that state `mergedInsertLogs` contains only the newly backfilled columns, so `buildCompactionOutputStats` drops the original insert/delta sizes and counts, timestamp range and quantiles, and null counts. DataCoord then replaces the durable statistics wholesale, making segment sizing, expiry/delete compaction, and nullable-vector index scheduling incorrect. Carry or merge the base `SegmentInfo.Stats` and add the new-column/stat-blob deltas instead of rebuilding cumulative statistics from the restart-empty summaries.

internal/rootcoord/create_collection_task.go
```diff
@@ -941,15 +941,46 @@ func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.Ana
...
 	if active && !h.EnableAnalyzer() {
 		return merr.WrapErrParameterInvalidMsg("field %s which has enable_match or is input of BM25 function must also enable_analyzer", fieldSchema.Name)
 	}
 	if !h.EnableAnalyzer() {
+		// A MinHash input legally keeps enable_analyzer=false yet its runner
+		// consumes analyzer_params. Collect them so any referenced file
+		// resources reach schema.FileResourceIds and ref-counting — otherwise
+		// the DDL gate admits a dependency the ledger never records and
+		// RemoveFileResource can delete a still-required resource.
+		if typeutil.IsMinHashFunctionInputField(collSchema, fieldSchema) {
```
buqian-zilliz: **[P2] Resolve MinHash inputs by name before assigning canonical IDs**

`validateSchemaAnalyzerFileResources` runs before `assignFieldAndFunctionID`, but this new branch calls `IsMinHashFunctionInputField`, which trusts any supplied nonzero IDs and does not fall back to names when they mismatch. A direct CreateCollection request can therefore provide matching input names plus stale nonzero `FieldID`/`InputFieldIds`, skip collection of the disabled-analyzer field's `analyzer_params`, and then have those IDs rewritten from names before persistence. The persisted function uses the resource without listing it in `FileResourceIds` or reserving its refcount, so the resource can be removed and later runner construction fails. Match by input names in this pre-normalization path, or canonicalize IDs before resource validation.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Confirmed. This is the known recovered-StorageV3 statistics gap from #50410, not a new regression in this PR. #51848 preserves `StatsBinlogSize` as base + new only; it does not provide complete cumulative-statistics preservation.

This is tracked by #51729. @tedxu has an open follow-up, #52006, which changes schema-bump materialization to report an increment and merges it into the existing `SegmentInfo.Stats` on the forward manifest-CAS adoption path, preserving the pre-existing delta/timestamp aggregates and old-field null counts. I will defer this cross-component change to that PR rather than duplicate it here.

internal/datacoord/compaction_task_bump_schema_version.go
MrPresent-Han: Confirmed. This is the external RootCoord -> DataCoord FileResource metadata/recovery contract gap tracked by #51861. It predates this PR and is deliberately outside #51848 scope.

This PR only guarantees that, once DataCoord has the resource metadata, BumpSchemaVersion forwards it to the DataNode text-index consumer. The lookup is best-effort, so a miss logs a warning and builds a resource-free plan rather than failing plan construction; a full rewrite that actually needs the resource then fails at resource resolution. I will not add a local workaround or claim end-to-end RefMode support here.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han: Confirmed and fixed in the current head. The guard now rejects only `len(resourceIDs) > 0 && !fileresource.IsSyncMode(mode)`, so SyncMode remains allowed while RefMode and CloseMode reject resource-backed backfill. `TestRejectAddFunctionInputAnalyzerFileResource` covers all three modes.

internal/rootcoord/create_collection_task.go
MrPresent-Han: Fixed in the latest push. Since `collectAnalyzerInfos` runs before `assignFieldAndFunctionID`, the MinHash-input check now matches by name (`isMinHashInputFieldByName`) instead of by field id, so a request carrying a nonzero stale/mismatched FieldID/InputFieldIds no longer skips collection of the disabled-analyzer field's analyzer_params — its file resources reach `FileResourceIds` and ref-counting. Regression test `TestCollectAnalyzerInfosMinHashInputMatchedByNameBeforeIDAssignment` uses a field with FieldID=999 and InputFieldIds=[888] (names agree, ids disagree) and is proven non-vacuous: it fails with the previous id-based predicate.

I opted for the by-name fix over reordering canonicalization ahead of validation: `assignFieldAndFunctionID` sits after `appendDynamicField`/`handleNamespaceField` and is skipped entirely on the `preserveFieldID` path, so moving it earlier is a broad, risky reorder. Note the same pre-normalization id-trust also affects `IsBm25FunctionInputField` in the active-field check on this path (milder: a stale-id BM25 input escapes the 'must enable_analyzer' rejection rather than corrupting the ledger) — I'll track that plus the broader function↔file-resource binding/ledger gaps as a follow-up.

internal/storage/arrow_util.go
```diff
@@ -327,6 +328,16 @@ func GenerateEmptyArrayFromSchema(schema *schemapb.FieldSchema, numRows int) (ar
 			bd.AppendValues(
 				lo.RepeatBy(numRows, func(_ int) []byte { return schema.GetDefaultValue().GetBytesData() }),
 				nil)
+		case schemapb.DataType_Geometry:
```
buqian-zilliz, line 331: **[P1] Apply WKT conversion when copying null geometry values**

This conversion only covers a Geometry field that is wholly absent. Once the column exists, `RecordBuilder.Append` and `storage.Sort` handle a null through `appendValueAt`'s generic `BinaryBuilder` branch, which reads `defaultValue.GetBytesData()`. Geometry defaults are persisted as WKT in `StringData`, so that branch appends a valid zero-length byte value instead of the WKB default, and a later sort, mix, or full-rewrite compaction can silently change those rows. Pass the field type into that path and perform the same WKT-to-WKB conversion, or store one canonical WKB default for every writer.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -187,6 +194,27 @@ func (c *Core) broadcastAlterCollectionSchemaAdd(ctx context.Context, broadcaste
...
+func (c *Core) rejectAddFunctionInputAnalyzerFileResource(ctx context.Context, schema *schemapb.CollectionSchema, function *schemapb.FunctionSchema) error {
+	analyzerInfos, err := collectFunctionInputAnalyzerInfos(schema, function)
+	if err != nil {
+		return err
+	}
+	resourceIds, err := c.validateAnalyzerInfos(ctx, analyzerInfos)
+	if err != nil {
+		return err
+	}
+	mode := Params.CommonCfg.DNFileResourceMode.GetValue()
```
buqian-zilliz, line 210: **[P1] Gate on the effective DataNode file-resource mode**

`common.fileResource.mode.dataNode` is refreshable, but this check reads its live value while both `FileResourceObserver.dnMode` and the DataNode's global file manager snapshot the mode only at initialization. After changing `ref` or `close` to `sync`, RootCoord admits this DDL and DataCoord stops attaching per-task resources, while already-running DataNodes still use the old mode and cannot resolve the analyzer during backfill. The schema and bound index can therefore commit with no viable backfill path. Gate on the effective node mode and readiness, or make this setting restart-only and require all DataNodes to restart and sync before accepting the DDL.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -129,6 +130,12 @@ func (c *Core) broadcastAlterCollectionSchemaAdd(ctx context.Context, broadcaste
 		if err := validator.ValidateFunction(schema, plan.Function.GetName(), true); err != nil {
 			return merr.Wrap(err, "invalid function schema")
 		}
+		if err := schemautil.ValidateAddFunctionInputNotText(schema, plan.Function); err != nil {
```
buqian-zilliz, line 133: **[P1] Run MinHash parameter validation on the direct RootCoord path**

The direct RootCoord path still never runs `ValidateMinHashFunction`: `validator.ValidateFunction(..., true)` skips `embedding.ValidateFunctions`, while the bound-index checks validate the vector and index dimension without relating it to `num_hashes` or checking the other MinHash parameters. A direct `AlterCollectionSchema` call can therefore commit, for example, output dim 64 with `num_hashes=3`; the new runner rejects this when schema-bump compaction constructs the materializer, so existing rows never receive the output. Run the pure MinHash schema and parameter validation here before broadcasting while keeping external runtime-provider checks disabled.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -748,17 +773,37 @@ type bumpSchemaVersionBatchWriter interface {
...
+	// it again from deferred cleanup; only the returned output still needs release.
+	l.exhausted = true
+	return output, err
+}
+
+func (l *bumpSchemaVersionWriterLease) Cleanup() {
+	if l == nil || l.exhausted {
+		return
+	}
+	output, _ := l.Close()
```
buqian-zilliz: **[P2] Abort the partial writer instead of closing it on errors**

`Cleanup` calls `Close`, which finalizes the FFI writer's pending column groups and files; destroying the returned `WriterOutput` only releases its C payload. Any earlier read, materialization, write, stats, or cancellation error after batches were written therefore flushes an uncommitted partial backfill to object storage, leaving unreferenced files and turning error cleanup into segment-scale I/O. Expose the underlying writer's `Destroy` or abort operation for this path and reserve `Close` for successful finalization.

internal/datacoord/compaction_task_bump_schema_version.go
```diff
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
...
+	// best-effort: on failure keep building the plan without resources instead of
+	// failing tasks (bump-only / partial backfill) that never need them. A
+	// full-rewrite that does need them fails at createTextIndex resource
+	// resolution, next to the actual consumer.
+	if fileresource.IsRefMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue()) &&
+		schemaNeedsTextIndex(taskProto.GetSchema()) &&
+		len(taskProto.GetSchema().GetFileResourceIds()) > 0 {
+		resources, err := t.meta.GetFileResources(context.Background(), taskProto.GetSchema().GetFileResourceIds()...)
+		if err != nil {
+			mlog.Warn(context.TODO(), "get file resources for schema bump compaction failed, building plan without them",
```
buqian-zilliz, line 132: **[P2] Resolve ref-mode analyzer resources before starting the rewrite**

In ref mode, this error branch dispatches a plan without `FileResources`. If the segment takes the full-rewrite path and its match analyzer uses those resources, DataNode writes and closes the replacement segment, adds stats and LOB manifest updates, and only then reaches `createTextIndex`, where resource resolution fails. That leaves the task failed after full-segment I/O with unadopted output and repeats the work on retry. Make the lookup failure fatal for full rewrites, or validate and download the required resources before producing rewrite output.

internal/util/function/minhash_function.go
buqian-zilliz: This issue still exists: both the runner constructor and DDL validator still multiply `num_hashes` by 32 before a checked bounds comparison, allowing a wrapped value to pass and reach the permutation allocations.

internal/storage/arrow_util.go
MrPresent-Han: Tracked in companion issue #52105 — out of scope for this PR. This copy path (`appendValueAt`'s BinaryBuilder branch) predates the PR (#45572) and is not touched by it; this PR only added geometry default support to `GenerateEmptyArrayFromSchema` (the wholly-absent-column prefill path). Also worth noting the trigger window is narrower than it first appears: a normal insert materializes null inputs into valid WKB whenever the default is already in place (proxy `FillWithDefaultValue` geometry case + ValidData set true), so regular insert + compaction cannot hit the branch — it requires storing nulls first and adding the default afterwards via schema evolution. The issue captures the exact sequence, the two affected call sites, and the suggested fix (thread the field type into `appendValueAt`, convert StringData WKT via `ConvertWKTToWKB`).

MrPresent-Han: Correction to my previous reply after deeper, executable verification: there is a more direct channel than the evolution-added-default sequence — **Parquet import**. `ReadNullableGeometryData` fills the default WKB into null slots but never marks them valid (unlike the generic `fillWithDefaultValueImpl`), the rows persist as genuine nulls, and import auto-schedules sort compaction under default configuration, so the rewrite happens within the same import lifecycle. So the trigger does not require any unusual operation sequence. #52105 has been updated accordingly, including the two-half fix (consumer `appendValueAt` + producer reader normalization) and the note that full-rewrite participates only via its selection path. Scope-wise this remains pre-existing (`appendValueAt` branch from #45572, reader gap older still) and is tracked there.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han: The config-vs-capability divergence described here is real, but it is a pre-existing platform property of `common.fileResource.mode.dataNode`, not something this gate introduces — and it cannot be resolved from this PR:

- The DataNode global file manager and `FileResourceObserver` both snapshot the mode at initialization, so the `refreshable` marking already has no effect on any live component; hot-updating the value changes nothing on running nodes in either direction (both sync→ref and ref→sync transitions effectively require a full rolling restart today).
- Independently, a DataNode configured `sync` without a resource address silently degrades to close mode at init (#48631), so even a restart-only policy cannot make the config value a reliable capability signal.
- There is currently no channel reporting a node's *effective* mode, so no coordinator-side check — this one or any existing consumer of the param — can gate on capability rather than configuration.

This gate uses the config value as deployment intent, which is the same contract every existing consumer of this param relies on. Gating on effective node mode requires capability reporting (heartbeat/RPC surface) plus removing the misleading `refreshable` marking and hardening the silent degrade — cross-component platform work that deserves a dedicated follow-up rather than a partial patch here. Keeping this PR scoped to the DDL-side guard.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Valid finding — after a DataCoord restart a V3 segment's FieldBinlog/Deltalog arrays are intentionally empty (per-field KVs are not persisted since #50410), so rebuilding an absolute `Statistics` from the plan-carried arrays drops the cumulative insert/delta sizes, timestamp range and null counts, and the completion path adopts it wholesale.

This is being addressed in a dedicated fix: #52006 (@tedxu) changes schema-bump materialization to report a stats *increment* instead of an absolute rebuild, so the durable cumulative `SegmentInfo.Stats` is merged rather than replaced. Deferring to that PR rather than duplicating the fix here.

internal/util/function/minhash_function.go
MrPresent-Han: Fixed in 173d7ba2fa — both sites now compare via division instead of multiplication: the runner checks `int64(numHashes) != outputDim/32` (outputDim is already validated positive and `%32==0` above), and `ValidateMinHashFunction` checks `outputDim%32 != 0 || outputDim/32 != int64(numHashes)`. Added overflow regression tests with `num_hashes = 2^59+1` for both paths; both fail against the multiply-based checks.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han: Fixed in 173d7ba2fa — MinHash parameter validation moved out of the `disableRuntimeCheck` bucket: `validator.ValidateFunction` now runs `ValidateMinHashFunction` unconditionally (same name filter), and the duplicate call inside `embedding.ValidateFunctions` was removed so the runtime bucket carries only model runtime checks. This also covers the proxy create path when the runtime check is disabled by config. Regression test: the direct path (`disableRuntimeCheck=true`) now rejects `num_hashes=3` with `dim=64`.

internal/datacoord/compaction_task_bump_schema_version.go
MrPresent-Han: Addressed in 173d7ba2fa on the DataNode side: `runFullSchemaRewrite` now fails fast — before any segment read or write — when the node's effective file-resource mode is ref, the schema has enable_match fields with `FileResourceIds`, and the plan carries no resources (mirroring DataCoord's attach predicate). DataCoord's best-effort attach stays as-is: it cannot know whether a segment will take the bump-only/partial path, so a fatal lookup there would fail tasks that never need the resources. Note the gate checks the node's effective mode (`GlobalFileManager.Mode()`), not the config value. Regression test asserts the gate fires before any segment access.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Fixed in 173d7ba2fa: the FFI layer already exposes a no-flush abort (`FFIPackedWriter.Destroy`, "releases writer resources without committing pending output"); wired it through as `packedRecordBatchWriter.Abort` and switched the lease's `Cleanup` to abort instead of close, so error paths no longer flush pending column groups to storage. The explicit `Close` on the success path is unchanged. One boundary worth stating: column-group files already flushed during `Write` cannot be undone by an abort — they stay manifest-unreferenced and remain GC's responsibility; this change removes the additional farewell flush at error time.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -425,6 +450,21 @@ func (t *bumpSchemaVersionCompactionTask) runSchemaVersionBumpOnly() *datapb.Com
 }
 
 func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields map[int64]struct{}) (*datapb.CompactionPlanResult, error) {
+	// Ref-mode analyzer file resources are attached to the plan by DataCoord
+	// best-effort (bump-only and partial backfill never need them). This path
+	// does — at createTextIndex, after the whole segment has been read,
+	// materialized and rewritten. When they are known missing, fail before any
+	// of that work instead.
+	if fileresource.GlobalFileManager != nil &&
```
MrPresent-Han: [P1 / fix in this PR] This collection-wide fail-fast mis-kills a legal schema: field A may use enable_match with a built-in analyzer and require no file resource, while unrelated field B contributes FileResourceIds as a non-match analyzer or MinHash input. schemaNeedsTextIndex is then true and the collection resource union is non-empty, but an empty plan does not prove that createTextIndex is missing a resource it needs. Remove this guard and add the A-with-built-in-match + B-with-unrelated-resource regression case. Complete match-field-scoped resource preparation and fail-early behavior is tracked by #52141.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han, line 210: [Final conclusion / skip] This gate reads deployment configuration, not the effective DataNode manager mode. Hot switching and sync-to-close silent downgrade remain real platform gaps, but fixing them requires node-reported effective mode/resource version/epoch plus scheduler readiness integration. They predate #51848 and are outside this PR; keep the current scoped guard and do not claim hot-switch or mixed-node coverage.

internal/storage/arrow_util.go
MrPresent-Han, line 331: [Final conclusion / skip] This new branch correctly converts the default for a wholly absent Geometry column. A separate pre-existing path can retain NULL validity in Parquet import and later rewrite the value as an empty invalid WKB through appendValueAt. That end-to-end issue is tracked by #52105 and assigned to congqixia. Do not expand #51848 or claim complete Geometry default coverage here.

MrPresent-Han (commented):
Final review synchronization after the end-to-end audit. In-scope changes for this PR are: (1) remove the collection-wide coarse FileResource fail-fast and add the mis-kill regression test; (2) aggregate, rate-limit, and enrich the dangling-alias warning. The DataNode effective-mode gap, Geometry NULL/default issue (#52105), DataCoord FileResource producer/recovery gap (#51861), cumulative Statistics preservation (#51729/#52006), and runtime FunctionRunner classification cleanup are pre-existing/out of scope. The BM25 enable_analyzer residual finding is withdrawn. Rebase onto current master and post-rebase CI remain mechanical merge prerequisites.

internal/datacoord/compaction_task_bump_schema_version.go
```diff
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
...
+	// Only the full-rewrite path consumes these (inline createTextIndex, mirroring
+	// sort/mix); which path a segment takes is decided on the datanode, so attach
+	// best-effort: on failure keep building the plan without resources instead of
+	// failing tasks (bump-only / partial backfill) that never need them. A
+	// full-rewrite that does need them fails at createTextIndex resource
+	// resolution, next to the actual consumer.
+	if fileresource.IsRefMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue()) &&
+		schemaNeedsTextIndex(taskProto.GetSchema()) &&
+		len(taskProto.GetSchema().GetFileResourceIds()) > 0 {
+		resources, err := t.meta.GetFileResources(context.Background(), taskProto.GetSchema().GetFileResourceIds()...)
```
MrPresent-Han, line 130: [Final conclusion / skip] DataCoord's FileResource map still has no production producer or startup recovery path; UpdateFileResources is test-only. This shared pre-existing gap is tracked by #51861 and also affects other RefMode compactions. Keep this best-effort forwarding, but do not claim end-to-end RefMode support in #51848; no linked fix PR was found during this review.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -1130,15 +1198,12 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
...
-				// comes from the freshly committed V3 stats (tracked on
-				// writerResult); insert/delta aggregates come from the
-				// merged arrays.
+				// Missing function outputs add new stats keys, so the final
+				// footprint is the base manifest plus this task's new blobs.
 				Stats: buildCompactionOutputStats(
 					mergedInsertLogs,
 					segment.GetDeltalogs(),
-					writerResult.statsBlobSize,
+					baseStatsBlobSize+writerResult.statsBlobSize,
```
MrPresent-Han, line 1191: [Final conclusion / skip] Adding the base manifest size here is a valid local improvement, but recovered StorageV3 can still rebuild an absolute Statistics object from incomplete plan arrays and overwrite cumulative durable data. This predates #51848 and is tracked by #51729 / #52006. If #52006 lands first, the rebase must preserve its increment-and-merge semantics.

internal/util/function/multi_analyzer_bm25_function.go
```diff
@@ -140,6 +164,14 @@ func (v *MultiAnalyzerBM25FunctionRunner) getAnalyzer(name string, analyzers map
 		return analyzers[name], nil
 	}
 
+	// "default" is the terminal fallback. If even it cannot resolve (missing, or
+	// aliased to an undefined analyzer), fail instead of recursing forever. Guard
+	// on the pre-alias name so a dangling default alias cannot loop. This is a
+	// system-side failure: ordinary unknown names fall back by design, so only a
+	// broken persisted configuration reaches this point — never request content.
+	if origName == "default" {
```
MrPresent-Han, line 189: [Final conclusion / skip] The new terminal guard correctly classifies broken persisted configuration as ErrFunctionFailed. Remaining constructor inconsistencies, such as MinHash metadata errors returning ParameterInvalid and top-level multi-analyzer JSON returning a bare error, predate this PR. Keep them out of #51848 and handle any broader runtime-classification cleanup separately.

internal/rootcoord/create_collection_task.go
```diff
@@ -941,15 +958,48 @@ func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.Ana
...
+	return analyzerInfos, nil
+}
+
 // validateAnalyzer validates active match/BM25 fields and fields with
-// enable_analyzer=true. Analyzer params are ignored while analyzer is disabled.
+// enable_analyzer=true. Analyzer params are ignored while analyzer is disabled,
+// except for MinHash inputs, which consume them regardless.
 func validateAnalyzer(collSchema *schemapb.CollectionSchema, fieldSchema *schemapb.FieldSchema, analyzerInfos *[]*querypb.AnalyzerInfo) error {
 	h := typeutil.CreateFieldSchemaHelper(fieldSchema)
 	active := h.EnableMatch() || typeutil.IsBm25FunctionInputField(collSchema, fieldSchema)
```
MrPresent-Han, line 992: [Final conclusion / withdrawn] The public schema contract does require a BM25 VARCHAR/TEXT input to set enable_analyzer=true. However, the BM25 runner does not use this flag as a runtime capability gate and falls back to default analyzer params when they are absent. Therefore a stale-ID defense-in-depth miss does not establish the previously claimed runtime/createTextIndex failure. No #51848 action is required for this point.

internal/util/function/multi_analyzer_bm25_function.go
```diff
@@ -119,10 +130,23 @@ func NewMultiAnalyzerBM25FunctionRunner(coll *schemapb.CollectionSchema, schema
 		runner.analyzers[name] = analyzer
 	}
 
+	// A dangling alias is accepted for compatibility. At runtime an ordinary
+	// dangling alias falls back to the default analyzer, while a dangling
+	// "default" alias makes the default unresolvable and fails the request.
+	// Warn once per construction so operators can spot the misconfiguration.
+	for aliasName, target := range runner.alias {
+		if _, ok := runner.analyzers[target]; !ok {
+			mlog.Warn(context.TODO(), "multi analyzer alias targets undefined analyzer; affected values fall back to the default analyzer, or fail at runtime if the default alias itself dangles",
```
MrPresent-Han: [P2 / fix in this PR] This emits one warning per dangling alias per runner construction. DataNode full-rewrite materialization constructs a runner per stale segment and repeats on retry, producing roughly segments x aliases x retries warnings. Aggregate all dangling aliases into one deterministic warning, use mlog.RatedWarn as a safety valve, and include database/collection/schema version, input field, and function identifiers. Preserve DDL acceptance and runtime fallback semantics.

Issue #51854, opened 2026-07-25
[Bug]: growing-source flush errors panic before writebuffer retry callback


bigsheeper:
## Bug Report

Growing-source flush errors can panic the DataNode before the writebuffer retry callback runs.

### What happened

`writeBufferBase.getGrowingSourceSyncTask` wires `GrowingSourceSyncTask.WithFailureCallback(wb.errHandler)`. The default writebuffer error handler panics. `GrowingSourceSyncTask.Run` also calls `t.HandleError(err)` in its own defer. For ordinary growing-source errors, such as `FlushGrowingData` returning an error, that defer invokes the default panic handler before `keyLockDispatcher` can run the callback list provided by `syncManager.SyncData`.

The writebuffer callback in `writeBufferBase.submitSyncTasks` is where growing-source progress is rolled back, `AbortSyncing` is applied, failure counters are recorded, non-retryable layout mismatches are classified, and retry scheduling happens. A panic in `GrowingSourceSyncTask.Run` prevents that recovery path from executing.

### Expected behavior

A normal growing-source flush error should be returned to the sync manager callback path so the writebuffer can roll back `SyncingRows`, record `failureCount` / `lastFailure`, and schedule retry where appropriate. It should not panic before that callback runs.

### Repro evidence

I added this failing repro before filing the issue:

```bash
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:$LD_LIBRARY_PATH \
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 \
  ./internal/flushcommon/writebuffer \
  -run 'TestL0WriteBuffer/TestBufferDataGrowingSourceMode/source_task_error_with_default_handler_reaches_retry_callback'
```

Current result on `milvus-io/master` (`785ab5117c06`):

```text
panic: flush growing source data: mock growing source flush error [recovered, repanicked]
...
github.com/milvus-io/milvus/internal/flushcommon/syncmgr.(*GrowingSourceSyncTask).HandleError
    internal/flushcommon/syncmgr/growing_source.go:597
github.com/milvus-io/milvus/internal/flushcommon/syncmgr.(*GrowingSourceSyncTask).Run.func1
    internal/flushcommon/syncmgr/growing_source.go:625
```

### Why this is production-impacting

[...]

PR #51855, opened 2026-07-25
enhance: make growing source HandleError idempotent

bigsheeper:
## What

`GrowingSourceSyncTask.HandleError` is invoked twice for a single failure:

- `GrowingSourceSyncTask.Run` calls it from its own defer
- `syncManager.submit` prepends a handler that calls it again (`sync_manager.go:164`), ahead of the writebuffer callback

So `failureCallback` runs twice and `DataNodeFlushBufferCount{status="fail"}` is incremented twice per failure. This PR makes both single-shot.

## Scope

**No behavior change today.** The failure callback wired for growing-source tasks is `wb.errHandler` (`write_buffer.go:1568`), whose default implementation panics (`options.go:48-51`), and no production code overrides it. That panic escapes `Run` before `keyLockDispatcher` reaches the callback list (`key_lock_dispatcher.go:178-181`), so the second `HandleError` call is never reached and the guard cannot fire. It takes effect once that callback stops panicking.

Deliberately **not** addressed here: that the panicking callback preempts the writebuffer recovery path (rollback of `SyncingRows`, failure counters, retry scheduling) — see #51854. Changing that is a behavior change and belongs in its own PR.

## Verification

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/syncmgr/` — ok, including the new `TestGrowingSourceSyncTaskHandleErrorIsIdempotent` (fails without the guard: callback count 2)
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/writebuffer/` — ok
- `golangci-lint run --build-tags dynamic,test ./internal/flushcommon/...` — 0 issues

issue: #51854

### DIFF
--- a/internal/flushcommon/syncmgr/growing_source.go
+++ b/internal/flushcommon/syncmgr/growing_source.go
@@ -400,6 +400,7 @@ type GrowingSourceSyncTask struct {
 
 	writeRetryOpts  []retry.Option
 	failureCallback func(error)
+	errorHandled    bool
 	tr              *timerecord.TimeRecorder
 }
 
@@ -593,6 +594,12 @@ func (t *GrowingSourceSyncTask) HandleError(err error) {
 	if errors.IsAny(err, merr.ErrSegmentNotFound, merr.ErrChannelNotFound) {
 		return
 	}
+	// Run's defer and syncManager.submit's handler both call HandleError for the
+	// same failure; keep the callback and the failure metric single-shot.
+	if t.errorHandled {
+		return
+	}
+	t.errorHandled = true
 	if t.failureCallback != nil {
 		t.failureCallback(err)
 	}
--- a/internal/flushcommon/syncmgr/growing_source_test.go
+++ b/internal/flushcommon/syncmgr/growing_source_test.go
@@ -145,6 +145,31 @@ func TestGrowingSourceSyncTaskHandleErrorSkipsFailureCallbackForStaleMetaErrors(
 	}
 }
 
+// GrowingSourceSyncTask.Run calls HandleError from its own defer and
+// syncManager.submit's handler calls it again for the same failure, so the
+// failure callback would otherwise run twice.
+func TestGrowingSourceSyncTaskHandleErrorIsIdempotent(t *testing.T) {
+	paramtable.Get().Init(paramtable.NewBaseTable())
+	metrics.DataNodeFlushBufferCount.Reset()
+
+	callbackCount := 0
+	task := NewGrowingSourceSyncTask().
+		WithFailureCallback(func(error) {
+			callbackCount++
+		})
+
+	err := merr.WrapErrServiceInternalMsg("flush growing source data failed")
+	task.HandleError(err)
+	task.HandleError(err)
+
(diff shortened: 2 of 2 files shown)


czs007:

Verified:
- `internal/flushcommon/syncmgr/growing_source.go:597` — `errorHandled` is set before the failure callback is invoked, so the task defer and the sync-manager wrapper cannot run it twice.
- `internal/flushcommon/syncmgr/key_lock_dispatcher.go:178` — `Task.Run` and all callbacks execute sequentially inside the same worker closure, so the new unsynchronized boolean is never accessed concurrently on the production path.
- `internal/flushcommon/writebuffer/write_buffer.go:965` — recovering the task-level callback panic lets the subsequent write-buffer callback roll back `SyncingRows`, record the failure, and schedule the retry.
- `internal/flushcommon/syncmgr/growing_source.go:602` — the recovery is scoped strictly around `failureCallback`; the original sync error is unchanged and still propagates through the callback chain and the returned future.
- `internal/flushcommon/syncmgr/growing_source.go:594` — segment-not-found and channel-not-found errors still short-circuit ahead of the guard, bypassing callbacks and failure metrics exactly as before.
- `internal/flushcommon/writebuffer/l0_write_buffer_test.go:979` — the new test drives a real sync manager with the production-equivalent panicking handler and asserts the original error reaches the future, failure state is recorded, and buffered rows are restored for retry.

czs007:

Verified:

- `internal/flushcommon/syncmgr/growing_source.go:593-606` — the stale-meta check (`ErrSegmentNotFound` / `ErrChannelNotFound`) runs before the single-shot flag is set, so stale errors still return early without consuming the guard, invoking the callback, or touching the failure metric.
- `internal/flushcommon/syncmgr/growing_source.go:625-633`, `internal/flushcommon/syncmgr/sync_manager.go:155-170`, `internal/flushcommon/syncmgr/key_lock_dispatcher.go:173-184` — both `HandleError` invocations for the same returned error happen sequentially on the same worker goroutine, so the duplicate call is suppressed and the new boolean is never accessed concurrently.
- `internal/flushcommon/writebuffer/write_buffer.go:1552-1586` — every retry constructs a fresh `GrowingSourceSyncTask`, so the task-local guard cannot leak across flush attempts and suppress a later attempt's failure handling.
- `internal/flushcommon/writebuffer/write_buffer.go:962-1041` — callback ordering and error propagation are unchanged; the rollback/retry callback still receives the original task error after the manager handler runs.
- `internal/flushcommon/writebuffer/options.go:48-51` — the default failure callback still panics as before; setting the flag ahead of the callback does not alter today's panic path or add reachable behavior.
- `internal/flushcommon/syncmgr/growing_source_test.go:107-170` — existing stale-error coverage is intact, and the new test asserts exactly one callback and one `DataNodeFlushBufferCount` increment across two non-stale invocations, so it fails deterministically without the guard.
- Repository-wide call-site search confirms `write_buffer.go:1553` is the only production site constructing this task; no task reuse or other concurrent `HandleError` entry point exists.

czs007 (approved):
/lgtm

Issue #51858, opened 2026-07-26
[Bug]: [3.0] e2e create_index for STL_SORT/BITMAP JSON path index fails — CI pins scalar index version to 3 but feature requires >= 4


xiaocai2333:
## Description

On the **3.0** branch, e2e CI deterministically fails to create `STL_SORT` / `BITMAP` JSON path indexes:

```
RPC error: [create_index], MilvusException: (code=1100,
message=JSON path index with STL_SORT requires scalar index engine version >= 4,
current resolved version: 3: invalid parameter)
```

This fails `test_milvus_client_index.py`, `test_milvus_client_query.py`, `test_milvus_client_json_path_index.py` (and cascades to compact/delete/hybrid_search files sharing the setup) on every 3.0 PR whose e2e-amd/arm test distribution includes those files — e.g. #51839, #51780, #51748, #51488. PRs that pass simply drew a test distribution that skipped those files (the pool dispatcher splits ~40-60 files per run), so the failure looks flaky but is systemic.

## Root cause

Two 3.0 changes conflict inside the same CI config:

1. #48953 (backported to 3.0) requires scalar index engine version **>= 4** for `STL_SORT`/`BITMAP` JSON path multi-type indexes (`common.MinScalarIndexVersionForJsonPathMultiType = 4`).
2. #51792 set the CI Helm values to `forceRebuildScalarSegmentIndex: true` + `targetScalarIndexVersion: 3`.

`ResolveScalarIndexVersion()` (`internal/datacoord/index_engine_version_manager.go`) with `forceRebuild == true` returns `target` verbatim → **3**, which is `< 4`, so `create_index` in `internal/datacoord/index_service.go` rejects the request.

The node actually supports up to `MaximumScalarIndexEngineVersion = 4` (`pkg/common/common.go`), so v4 is a valid, supported target — the CI config simply pins one version too low for the JSON-path-index tests that also run in the same suite. master does not pin this value at all, so it is 3.0-specific.

## Fix

Bump `targetScalarIndexVersion: 3` → `4` in the 16 `tests/_helm/values/**` files #51792 touched. This keeps the force-rebuild testing intent (now rebuilding to the max supported version, 4) while satisfying the JSON-path-index requirement. No product-code change; test/CI config only.

PR: (attached)

xiaocai2333:
[...]

PR #51792, opened 2026-07-24
test: Set CI Helm values index versions

yanliang567:
pr: #51748

## Summary

- Pin dataCoord.targetVecIndexVersion to 10 in CI Helm values that inject Milvus config
- Enable scalar segment index rebuild and pin dataCoord.targetScalarIndexVersion to 3 for the same CI configs
- Add the ARM distributed woodpecker service boundary values file with ARM scheduling settings

issue: #51791

## Test Plan

- ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f) }' tests/_helm/values/*/*
- ruby -e 'require "yaml"; missing=[]; checked=0; ARGV.each do |f| outer=YAML.load_file(f); user=outer.dig("extraConfigFiles", "user.yaml"); next unless user; checked += 1; inner=YAML.safe_load(user); dc=inner["dataCoord"] || {}; ok = dc["targetVecIndexVersion"] == 10 && dc["forceRebuildScalarSegmentIndex"] == true && dc["targetScalarIndexVersion"] == 3; missing << f unless ok; end; puts checked; exit(missing.empty? ? 0 : 1); end' tests/_helm/values/*/*
- ruby -e 'require "yaml"; f="tests/_helm/values/e2e-arm/distributed-woodpecker-service-boundary"; outer=YAML.load_file(f); raise "top" unless outer.dig("nodeSelector", "jenkins-e2e-arm") == "true"; raise "etcd" unless outer.dig("etcd", "nodeSelector", "jenkins-e2e-arm") == "true"; raise "minio" unless outer.dig("minio", "nodeSelector", "jenkins-e2e-arm") == "true"; raise "top toleration" unless outer["tolerations"].any? { |t| t["key"] == "jenkins-e2e-arm" }; puts "arm_boundary_scheduling_ok"'

PR #51859, opened 2026-07-26
test: [3.0] pin CI scalar index version to 4 so JSON path index tests can run

xiaocai2333:
issue: #51858

## Problem

On 3.0, e2e `create_index` for `STL_SORT`/`BITMAP` JSON path indexes fails deterministically:

```
code=1100, JSON path index with STL_SORT requires scalar index engine version >= 4, current resolved version: 3
```

#51792 pinned the CI Helm values to `forceRebuildScalarSegmentIndex: true` + `targetScalarIndexVersion: 3`. With force-rebuild set, `ResolveScalarIndexVersion()` returns the target verbatim (3). But `STL_SORT`/`BITMAP` JSON path multi-type indexes require version **>= 4** (`common.MinScalarIndexVersionForJsonPathMultiType`, on 3.0 via #48953). So `create_index` (`internal/datacoord/index_service.go`) rejects those requests, failing `test_milvus_client_index.py` / `test_milvus_client_query.py` / `test_milvus_client_json_path_index.py` (and cascading files) on every 3.0 PR whose e2e test distribution runs them — e.g. #51839 (the current 3.0 tip), #51780, #51748, #51488.

## Fix

Bump `targetScalarIndexVersion: 3` → `4` in the 16 `tests/_helm/values/**` files #51792 touched. The node supports up to `MaximumScalarIndexEngineVersion = 4`, so 4 is a valid target; force-rebuild is preserved (now rebuilding to the max supported version) and the JSON-path-index requirement is met. master does not pin this value at all — this is a 3.0-only CI config fix, no product code changed.

## Verification

- `ResolveScalarIndexVersion()` with target=4 + forceRebuild=true returns 4 ≥ `MinScalarIndexVersionForJsonPathMultiType` (4) → `create_index` accepts STL_SORT/BITMAP JSON path indexes.
- All 16 values files parse as valid YAML after the change.

### DIFF
--- a/tests/_helm/values/e2e-amd/distributed-pulsar
+++ b/tests/_helm/values/e2e-amd/distributed-pulsar
@@ -87,7 +87,7 @@ extraConfigFiles:
     dataCoord:
       targetVecIndexVersion: 10
       forceRebuildScalarSegmentIndex: true
-      targetScalarIndexVersion: 3
+      targetScalarIndexVersion: 4
       compaction:
         bumpSchemaVersion:
           enabled: true
--- a/tests/_helm/values/e2e-amd/distributed-woodpecker-service-boundary
+++ b/tests/_helm/values/e2e-amd/distributed-woodpecker-service-boundary
@@ -93,7 +93,7 @@ extraConfigFiles:
     dataCoord:
       targetVecIndexVersion: 10
       forceRebuildScalarSegmentIndex: true
-      targetScalarIndexVersion: 3
+      targetScalarIndexVersion: 4
       enableCompaction: true
       compaction:
         enableAutoCompaction: false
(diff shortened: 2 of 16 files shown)


xiaocai2333:
For the record: this PR was the wrong fix. The actual fix is #51798 ("fix: [3.0] honor resolved index versions", merged 2026-07-27 00:16 +0800), which raised `CurrentScalarIndexEngineVersion` from 3 to 4 on 3.0 — so `ResolveScalarIndexVersion()` now returns 4 and satisfies `MinScalarIndexVersionForJsonPathMultiType`. The CI Helm pin this PR touched was not the effective input. Verified after the fix: #51868's e2e passed 103/103 with `test_milvus_client_index.py` and `test_milvus_client_json_path_index.py` exercised and zero version errors. Tracking issue #51858 closed accordingly.

Issue #51852, opened 2026-07-25
[Bug]: Object-storage error mapping drops cancellation and leaves LIST errors untyped


microbluey:
### Is there an existing issue for this?

- [x] I have searched the existing issues and pull requests.

### Environment

- Milvus version: master `785ab5117c061b914ea58401944186e5ce25dede`
- Deployment mode: any deployment using remote object storage
- Storage: S3/MinIO-compatible, Azure Blob Storage, or GCS

### Current behavior

Object-storage operations do not preserve or classify errors consistently.

#### 1. Cancellation and deadline identity is lost

`mapObjectStorageError` falls through to:

```go
merr.WrapErrIoFailed(fileName, err)
```

for `context.Canceled` and `context.DeadlineExceeded`.

`WrapErrIoFailed` copies `err.Error()` into an `ErrIoFailed` description but does not retain the original error as a cause. Consequently:

```go
mapped := mapObjectStorageError("object", context.Canceled)

errors.Is(mapped, context.Canceled) // false
merr.Code(mapped)                   // generic I/O code, not CanceledCode
```

The same applies to `context.DeadlineExceeded`.

This also affects `RemoteChunkManager` metrics. GET, PUT, STAT, REMOVE, and COPY classify cancellation using `errors.Is(err, context.Canceled)`. Because the adapter has already replaced the error, canceled operations are counted as `fail`, not `cancel`.

#### 2. LIST/WALK bypasses the common provider mapper

Most adapter operations route provider errors through `mapObjectStorageError`, but listing is inconsistent:

- MinIO/S3-compatible `WalkWithObjects` returns `object.Err` directly.
- Azure `WalkWithObjects` returns both flat- and hierarchy-pager errors directly.
- GCP native listing already uses `mapObjectStorageError`.

As a result, the same provider error—for example S3 `SlowDown`—is a typed, retryable Milvus error for STAT/PUT but remains a raw SDK error for LIST. Callers such as GC and cleanup paths therefore cannot rely on one error taxonomy across object-storage operations.

### Expected behavior

- `context.Canceled` and `context.DeadlineExceeded` remain detectable through `errors.Is`.
- `merr.Code` reports the existing canceled/timeout codes for those errors.
- Existing provider mappings remain unchanged.
- MinIO and Azure LIST/WALK errors use the same provider mapping as other object-storage operations.
- Canceled object-storage operations are recorded under the cancel metric label rather than failure.

### Proposed fix

[...]

PR #51860, opened 2026-07-26
fix: preserve object storage cancellation errors

microbluey:
issue: #51852

## What this PR does

- preserves `context.Canceled` and `context.DeadlineExceeded` at the object-storage mapping boundary so `errors.Is` and the existing canceled/timeout status codes remain available to callers
- maps MinIO list-channel errors and both Azure list-pager error paths through `mapObjectStorageError`, matching the existing GCP listing behavior
- adds focused coverage for direct and wrapped context termination, cancellation metric classification, and representative MinIO LIST provider errors

## Why

`mapObjectStorageError` previously converted context termination to `ErrIoFailed` by copying only the message. That removed the cause chain, changed the status code to generic I/O failure, and caused canceled GET/PUT operations to be counted as failures. LIST/WALK was also inconsistent because MinIO and Azure returned raw SDK errors while other operations used Milvus I/O errors.

## Verification

Passed locally:

- `gofmt`
- `git diff --check`
- `go test -count=1 ./util/merr -run 'Test.*Code|Test.*Canceled|Test.*Timeout'` from `pkg/`

The focused `internal/storage` package tests require the Milvus native toolchain. I attempted the official builder path locally, but the repository-pinned builder tag was not yet published; the available amd64 fallback builder under Apple Silicon/Rosetta failed while linking oneTBB with `GNU make jobserver: Bad file descriptor`, including with a fully serial build. The focused storage tests are therefore left for the upstream native CI runner.

### DIFF
--- a/internal/storage/azure_object_storage.go
+++ b/internal/storage/azure_object_storage.go
@@ -157,7 +157,7 @@ func (AzureObjectStorage *AzureObjectStorage) WalkWithObjects(ctx context.Contex
 		for pager.More() {
 			pageResp, err := pager.NextPage(ctx)
 			if err != nil {
-				return err
+				return mapObjectStorageError(prefix, err)
 			}
 			for _, blob := range pageResp.Segment.BlobItems {
 				if !walkFunc(&ChunkObjectInfo{FilePath: *blob.Name, ModifyTime: *blob.Properties.LastModified}) {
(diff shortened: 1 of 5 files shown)


microbluey:
/assign @liliu-z

Friendly ping for review — CI is green (Build / Code-Check / UT-Integration / UT-GO all passed on `d081830`, `ci-passed` + `dco-passed`), and the change is scoped to the object-storage error boundary:

- `mapObjectStorageError` now preserves `context.Canceled` / `context.DeadlineExceeded` in the cause chain, so `errors.Is` and the existing canceled/timeout status codes keep working instead of collapsing to a generic `ErrIoFailed` (which also made canceled GET/PUT count as failures in the metrics).
- MinIO list-channel errors and both Azure list-pager paths now go through the same mapper, matching the existing GCP listing behavior.

Happy to split this into separate cancellation and LIST-consistency commits, or adjust the approach, if that makes review easier. Tracking issue: #51852.

Issue #51863, opened 2026-07-26
[Bug]: QueryNode ReleaseSegments can hang after delegator close


bigsheeper:
## Summary

QueryNode can hang a ReleaseSegments or SyncDistribution control path when it races with delegator close. The delegator removal path waits on the channel returned by distribution.RemoveDistributions, but after distribution.Close the snapshot loop is stopped and notifySnapshotUpdate becomes a no-op. If RemoveDistributions captures the current snapshot's cleared channel in that state, nothing closes the channel, so the caller waits forever.

## Impact

This can leak the RPC goroutine and leave QueryCoord-side segment release or distribution synchronization work stuck during channel unsubscribe/release races. It is a normal control-plane race, not a panic-only path.

## Call chain

- QueryNode.ReleaseSegments -> shardDelegator.ReleaseSegments -> distribution.RemoveDistributions -> waits on returned channel
- QueryNode.UnsubDmChannel removes the delegator and calls delegator.Close concurrently
- distribution.Close stops the snapshot loop; later RemoveDistributions cannot generate the next snapshot that would close the captured signal

## Repro

A focused regression test added before the fix fails on master:

[2026/07/27 02:13:19.148 +09:00] [INFO] [distance/calc_distance_amd64.go:15] ["Hook avx for go simd distance computation"]
[2026/07/27 02:13:19.189 +09:00] [DEBUG] [config/refresher.go:69] ["start refreshing configurations"] [source=FileSource]
[2026/07/27 02:13:19.190 +09:00] [DEBUG] [config/etcd_source.go:52] ["init etcd source"] [etcdInfo="{\"UseEmbed\":false,\"EnableAuth\":true,\"UserName\":\"etcdadmin\",\"PassWord\":\"etcdadmin\",\"UseSSL\":false,\"Endpoints\":[\"localhost:2379\"],\"KeyPrefix\":\"by-dev\",\"CertFile\":\"/path/to/etcd-client.pem\",\"KeyFile\":\"/path/to/etcd-client-key.pem\",\"CaCertFile\":\"/path/to/ca.pem\",\"MinVersion\":\"1.3\",\"DialTimeout\":5000000000,\"RefreshInterval\":5000000000}"]
[2026/07/27 02:13:19.191 +09:00] [INFO] [etcd/etcd_util.go:227] ["create etcd client(enable auth)"] [useSSL=false] [endpoints="[localhost:2379]"] [minVersion=1.3]
[...]

PR #51864, opened 2026-07-26
fix: unblock delegator distribution removal after close

bigsheeper:
issue: #51863

## What this PR changes

- Expires the current delegator distribution snapshot during Close so any RemoveDistributions caller that already captured the old cleared signal is unblocked.
- Returns an already-closed signal when RemoveDistributions observes a closed distribution, since the snapshot loop can no longer generate a new snapshot.
- Adds regressions for both orderings: removal after close, and close after removal captured the signal.

## Verification

- GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator -run 'TestDistributionSuite/TestRemoveDistributionsAfterCloseReturnsClosedSignal'
- GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator -run 'TestDistributionSuite/Test(RemoveDistributionsAfterCloseReturnsClosedSignal|CloseClearsPendingRemoveDistributionSignal)'
- GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator
- GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib make static-check

### DIFF
--- a/internal/querynodev2/delegator/distribution.go
+++ b/internal/querynodev2/delegator/distribution.go
@@ -213,6 +213,17 @@ func (d *distribution) PinReadableSegments(requiredLoadRatio float64, partitions
 	d.mut.RLock()
 	defer d.mut.RUnlock()
 
+	// Close() expires the current snapshot, so with no reader pinned its cleared
+	// channel is already closed. Pinning it afterwards would hand out a snapshot
+	// whose reader barrier is spent, and the next RemoveDistributions would let
+	// ReleaseSegments unload those segments while this reader is still fanning out
+	// to them. This check is atomic against Close(): the pin holds d.mut.RLock and
+	// Close holds d.mut.Lock, so either the pin wins and its inUse keeps the
+	// barrier up, or it observes closed here and gives up.
+	if d.closed.Load() {
+		return nil, nil, nil, -1, merr.WrapErrChannelNotAvailable(d.channelName, "channel distribution is closed")
+	}
+
 	requireFullResult := requiredLoadRatio >= 1.0
 	loadRatioSatisfy := d.queryView.GetLoadedRatio() >= requiredLoadRatio
 	var isServiceable bool
--- a/internal/querynodev2/delegator/distribution.go
+++ b/internal/querynodev2/delegator/distribution.go
@@ -600,11 +600,17 @@ func (d *distribution) RemoveDistributions(sealedSegments []SegmentEntry, growin
 	}
 
 	var signal chan struct{}
-	if d.closed.Load() {
-		signal = getClosedCh()
-	} else if current := d.current.Load(); current != nil {
+	if current := d.current.Load(); current != nil {
 		// Capture current snapshot's cleared channel. The next genSnapshot will
 		// create a new snapshot and expire this one, closing the channel.
+		//
+		// This holds after Close() too: Close() expires the current snapshot, so
+		// its cleared channel closes as soon as the last in-flight reader unpins
+		// it -- keeping the reader barrier this function documents, instead of
+		// short-circuiting to an already-closed channel and letting the caller
+		// release segments out from under a running QueryStream. genSnapshot()
+		// refuses to install a new snapshot once closed, so this channel is
+		// guaranteed to be reached by an Expire.
 		signal = current.cleared
 	} else {
 		signal = make(chan struct{})
@@ -629,6 +635,21 @@ func (d *distribution) RemoveDistributions(sealedSegments []SegmentEntry, growin
 // in which, user could use found nodeID=>segmentID list.
 // mutex RLock is required before calling this method.
 func (d *distribution) genSnapshot() chan struct{} {
+	// Once closed, d.current must stay the expired snapshot installed by Close():
+	// RemoveDistributions hands that snapshot's cleared channel to callers as the
+	// reader barrier, and a fresh unexpired snapshot here would leave that channel
+	// forever open (the loop is stopping, so nothing would expire it) and
+	// reintroduce the post-close hang. Reachable from the snapshot loop -- an
+	// iteration that already took a notifier token can be parked on d.mut while
+	// Close() holds it -- and from AddGrowing/SyncTargetVersion/Flush, none of
+	// which are ordered against Close().
+	if d.closed.Load() {
+		if current := d.current.Load(); current != nil {
+			return current.cleared
+		}
+		return getClosedCh()
+	}
+
 	// stores last snapshot
 	// ok to be nil
 	last := d.current.Load()
--- a/internal/querynodev2/delegator/distribution.go
+++ b/internal/querynodev2/delegator/distribution.go
@@ -599,10 +599,12 @@ func (d *distribution) RemoveDistributions(sealedSegments []SegmentEntry, growin
 		delete(d.growingSegments, growing.SegmentID)
 	}
 
-	// Capture current snapshot's cleared channel. The next genSnapshot will
-	// create a new snapshot and expire this one, closing the channel.
 	var signal chan struct{}
-	if current := d.current.Load(); current != nil {
+	if d.closed.Load() {
+		signal = getClosedCh()
+	} else if current := d.current.Load(); current != nil {
+		// Capture current snapshot's cleared channel. The next genSnapshot will
+		// create a new snapshot and expire this one, closing the channel.
 		signal = current.cleared
 	} else {
 		signal = make(chan struct{})
@@ -816,7 +818,12 @@ func (d *distribution) Flush() {
 // Close stops the background snapshot loop and waits for it to exit.
 func (d *distribution) Close() {
 	d.closeOnce.Do(func() {
+		d.mut.Lock()
 		d.closed.Store(true)
+		if current := d.current.Load(); current != nil {
+			current.Expire(d.getCleanup(current.version))
+		}
+		d.mut.Unlock()
 		close(d.snapshotClose)
 	})
 	<-d.snapshotDone
--- a/internal/querynodev2/delegator/distribution_test.go
+++ b/internal/querynodev2/delegator/distribution_test.go
@@ -521,7 +521,7 @@ func (s *DistributionSuite) TestRemoveDistributionsAfterCloseReturnsClosedSignal
 
 	ch := s.dist.RemoveDistributions([]SegmentEntry{{NodeID: 1, SegmentID: 1}}, nil)
 
-	timeout := time.NewTimer(100 * time.Millisecond)
+	timeout := time.NewTimer(time.Second)
 	defer timeout.Stop()
 	select {
 	case <-ch:
@@ -540,7 +540,7 @@ func (s *DistributionSuite) TestCloseClearsPendingRemoveDistributionSignal() {
 	ch := s.dist.RemoveDistributions([]SegmentEntry{{NodeID: 1, SegmentID: 1}}, nil)
 	s.dist.Close()
 
-	timeout := time.NewTimer(100 * time.Millisecond)
+	timeout := time.NewTimer(time.Second)
 	defer timeout.Stop()
 	select {
 	case <-ch:
--- a/internal/querynodev2/delegator/distribution_test.go
+++ b/internal/querynodev2/delegator/distribution_test.go
@@ -591,6 +591,33 @@ func (s *DistributionSuite) TestRemoveDistributionsAfterCloseKeepsReaderBarrier(
 	}
 }
 
+// TestPinReadableSegmentsAfterCloseFails covers the opposite ordering of the test
+// above: Close() first, pin afterwards. Close() expires the current snapshot, so
+// with no reader pinned its cleared channel is already closed -- a late pin would
+// get a snapshot whose reader barrier is spent, and the next RemoveDistributions
+// would return that already-closed channel. QueryStream takes no delegator
+// lifetime, so it can still reach PinReadableSegments after Close().
+func (s *DistributionSuite) TestPinReadableSegmentsAfterCloseFails() {
+	s.dist.AddDistributions(SegmentEntry{
+		NodeID:      1,
+		SegmentID:   1,
+		PartitionID: 1,
+	})
+	// update target version, make distribution serviceable
+	s.dist.SyncTargetVersion(&querypb.SyncAction{
+		TargetVersion:         1000,
+		SealedSegmentRowCount: map[int64]int64{1: 100},
+	}, []int64{1})
+
+	// Serviceable on its own is not enough once the distribution is closed.
+	s.Require().True(s.dist.Serviceable())
+	s.dist.Close()
+
+	_, _, _, version, err := s.dist.PinReadableSegments(1.0, 1)
+	s.Error(err)
+	s.Equal(int64(-1), version)
+}
+
 func (s *DistributionSuite) TestPeek() {
 	type testCase struct {
 		tag      string
@@ -1568,6 +1595,17 @@ func (s *DistributionSuite) TestSyncTargetVersion_RedundantGrowingLogic() {
 	}
 }
 
+// stopSnapshotLoop stops the background snapshot goroutine without marking the
+// distribution closed. Tests that patch channelQueryView methods need the loop
+// quiescent, but PinReadableSegments refuses a closed distribution, so Close()
+// cannot serve that purpose. It shares closeOnce so a later Close() is a no-op.
+func stopSnapshotLoop(d *distribution) {
+	d.closeOnce.Do(func() {
+		close(d.snapshotClose)
+	})
+	<-d.snapshotDone
+}
+
 func TestPinReadableSegments(t *testing.T) {
 	// Helper function to create test distribution
 	setupDistribution := func() *distribution {
@@ -1585,7 +1623,7 @@ func TestPinReadableSegments(t *testing.T) {
 		// Flush pending snapshot and stop background goroutine to avoid
 		// data race with mockey patches in sub-tests.
 		dist.Flush()
-		dist.Close()
+		stopSnapshotLoop(dist)
 		return dist
 	}
 
@@ -1671,7 +1709,7 @@ func TestPinReadableSegments_ServiceableLogic(t *testing.T) {
 
 	// Stop background goroutine to avoid data race with mockey patches.
 	dist.Flush()
-	dist.Close()
+	stopSnapshotLoop(dist)
 
 	// Test case: requireFullResult=true, Serviceable=false, GetLoadedRatio=1.0
 	// This tests the case where load ratio is satisfied but serviceable is false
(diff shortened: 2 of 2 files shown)


czs007:
Adversarial review surfaced one concurrency issue in the new `close` fast path of `RemoveDistributions` that should be addressed before merge.

## Medium

**`internal/querynodev2/delegator/distribution.go:603`** — The newly added `if d.closed.Load() { signal = getClosedCh() }` returns an already-closed channel after `Close()`, discarding the reader barrier the function's own doc comment promises (distribution.go:574-576: the signal is closed only once the snapshot containing the removed segments has expired). That barrier is the sole gate before `worker.ReleaseSegments` (`delegator_data.go:1249-1251` waits on the signal, 1275 releases). The assumption that readers are drained before `Close()` does not hold for `QueryStream`: unlike `Search` (delegator.go:481-484) and `Query` (delegator.go:716-719), `QueryStream` (delegator.go:626-630) never calls `sd.lifetime.Add`, so `sd.lifetime.Wait()` at delegator.go:1391 does not wait for it while it holds a pinned snapshot (delegator.go:664-669) and runs `worker.QueryStreamSegments` (delegator.go:697). `ReleaseSegments` is likewise unprotected by the delegator lifetime, and services.go:683 can obtain the delegator concurrently with the `GetAndRemove` + `Close()` at services.go:413-421.

Failure: an in-flight `QueryStream` has pinned a snapshot containing segment S and is executing on the worker while `UnsubDmChannel` drives `distribution.Close()` (setting `d.closed`) and a concurrent `ReleaseSegments(shard=X, Scope=Historical, S)` reaches `RemoveDistributions` → the fast path returns the closed channel, `<-signal` no longer blocks, S is released on the target node mid-read, and the stream's subtask fails with `merr.WrapErrSegmentNotLoaded(S, "segment not found")` (segments/manager.go:570-573), returning an error to the proxy instead of the result the documented semantics would have produced.

Suggestion: preserve the barrier in the close path rather than short-circuiting it. Note that simply returning `current.cleared` in the close branch is not safe as-is: `Close()` expires `current` before `close(d.snapshotClose)` (distribution.go:821-827), so a `snapshotLoop` iteration already holding a notifier token and blocked on `d.mut` (distribution.go:196-200) can install a fresh, unexpired snapshot into `d.current` after the critical section — whose `cleared` would never close, reintroducing the #51863 deadlock. Either reorder `Close()` to stop the loop before expiring, or check `current.expired` in the close branch and only fall back to `getClosedCh()` when it holds. A regression test covering "pin snapshot → Close → post-close removal → assert the signal stays blocked until `Unpin`" would lock this in. *(raised by chyezh, tinswzy)*

bigsheeper:
Confirmed, and fixed in `26b19156b9`. Your reading of the failure is right, including the caveat about the naive fix — which I hit, and which pushed the fix somewhere slightly different from either option you offered.

**Verified, point by point:**

- The doc comment does promise a reader barrier (`distribution.go:573-575`, "all in-flight reads using that snapshot have finished"), and `snapshot.go:124` backs it — `cleared` closes only on `expired && inUse == 0`.
- That signal is the sole gate before `worker.ReleaseSegments`: `RemoveDistributions` → `<-signal` → `worker.ReleaseSegments`.
- `QueryStream` really is the hole. `Search` (`delegator.go:481-484`) and `Query` (`:716-719`) both take `sd.lifetime.Add` + `defer Done`; `QueryStream` (`:626+`) only checks `Serviceable()`. So `sd.lifetime.Wait()` at `:1391` does not wait for it, while it holds `PinReadableSegments` + `defer Unpin(version)` and runs `worker.QueryStreamSegments`.
- `ReleaseSegments` takes no delegator lifetime, and `services.go:683` can `Get` the delegator concurrently with the `GetAndRemove` + `Close()` at `:413`/`:421`.

**On your caveat — it's right, and it's worse than stated.** You noted the snapshot loop can be parked on `d.mut` holding a notifier token and install a fresh unexpired snapshot after `Close()`'s critical section. True. But `genSnapshot` has four callers, not one: `AddGrowing:460`, `SyncTargetVersion:547` and `Flush:813` also call it under `d.mut`, and none of them are ordered against `Close()` either. So option (a) (reorder `Close()` to stop the loop first) would still leave three paths open, and option (b) (`current.expired` check) narrows the window rather than closing it — when `expired` is false, falling back to `getClosedCh()` drops the barrier again.

So I guarded `genSnapshot` itself: once `d.closed` is set it installs nothing and returns the current snapshot's `cleared`. That makes `d.current` permanently the expired snapshot `Close()` installed, which lets `RemoveDistributions` go back to plain `current.cleared` — barrier preserved, and #51863 can't return because `Close()` already expired it, so `cleared` closes as soon as `inUse` hits zero. One guard covers all four callers. All `genSnapshot` callers discard its return value (checked), so the early return is safe.

**Regression test** is the shape you asked for: pin → `Close()` → post-close removal → assert the signal stays blocked → `Unpin` → assert it closes. I verified it isn't vacuous by temporarily restoring the short-circuit; it fails with `remove distribution signal closed while a reader still pinned the snapshot`, and passes with the fix.

Full `./internal/querynodev2/delegator` is green (29.6s), lint clean on the package.

**Not fixed here:** `QueryStream`'s missing `lifetime.Add` is a pre-existing defect and is the precondition that makes this finding reachable at all. Adding it changes QueryStream's lifetime semantics and deserves its own change rather than riding along in a distribution-barrier fix. The fix here is correct independently — it restores the documented barrier regardless of which readers the lifetime happens to drain — but I want to be explicit that the underlying asymmetry is still there.

czs007:
Adversarial review surfaced one non-blocking concurrency gap; the two-commit fix itself is correct and the reader barrier holds in the closed path.

## Medium

**`internal/querynodev2/delegator/distribution.go:212` — `PinReadableSegments` does not check `d.closed`, so a post-close pin can bypass the `RemoveDistributions` reader barrier.**

`Close()` (distribution.go:840-851) sets `d.closed` and calls `current.Expire(...)` under `d.mut`, freezing `d.current` at the closing snapshot `S0`. If no reader is pinned at that moment (`inUse == 0`), `checkCleared` (snapshot.go:123-141) fires its `once` and closes `S0.cleared` permanently. `PinReadableSegments` gates only on `d.queryView.Serviceable()` — which `Close()` never flips — and `snapshot.Get()` (snapshot.go:82-85) does not check `expired`, so `S0` can still be pinned afterwards. Every subsequent `RemoveDistributions` then returns that already-closed channel.

Failure: a `QueryStream` RPC passes the `sd.Serviceable()` check at delegator.go:628, takes the `waitTSafe` fast path (delegator.go:1116-1119, no serviceability recheck) and reaches `PinReadableSegments` after `distribution.Close()` has completed, pinning the cleared `S0` → the concurrent `ReleaseSegments` path sees `<-signal` return immediately (delegator_data.go:1251) and physically unloads those segments at delegator_data.go:1275 while the stream is still fanning out to them, failing the query with `SegmentNotLoaded`; the barrier stays defeated for every later removal, since the channel never reopens.

Reachability comes from `QueryStream` not registering with `sd.lifetime` (unlike `Search` at delegator.go:481 and `Query` at delegator.go:716), so `sd.lifetime.Wait()` at delegator.go:1391 does not drain it. This predates the PR and was previously masked: master's `Close()` (distribution.go:817-823) did not expire the snapshot, so post-close removals hung forever and the physical release never happened. Replacing that hang with a real release is a strict improvement — hence non-blocking — but it makes the window observable.

Suggestion: fast-fail on `d.closed` at the top of `PinReadableSegments`. Pin holds `d.mut.RLock()` and `Close` holds `d.mut.Lock()`, so the check is atomic with respect to close: either the pin wins (`inUse > 0`, barrier holds) or it observes `closed` and returns an error. Note that `TestRemoveDistributionsAfterCloseKeepsReaderBarrier` (distribution_test.go:558) pins *before* `Close()` and therefore does not cover the close-then-pin ordering; a companion test for that order would be worth adding. (raised by xiaocai2333, chyezh)

bigsheeper:
Confirmed and fixed in `815989839f` (the branch is also rebased onto latest master, so the two earlier commits have new SHAs; their content is unchanged). Your reading of the window is right, and I'd rather close it here than ship this PR with it open — without the fix, the net effect of this PR is trading a hang for a possible use-after-release, which is not the trade the PR claims to make.

**Verified, point by point, against the head you reviewed (`26b19156b9`):**

- `Close()` (`distribution.go:840-851`) sets `closed` and expires the current snapshot under `d.mut`, and `genSnapshot` refuses to install a new one once closed (`:646-651`), so `d.current` is frozen at S0.
- `checkCleared` (`snapshot.go:123-141`) fires its `once` and closes `S0.cleared` permanently when `expired && inUse == 0`.
- `snapshot.Get()` (`snapshot.go:82-86`) only does `inUse.Inc()` plus the filter — it never consults `expired`, so S0 stays pinnable after that.
- `PinReadableSegments` had no `closed` check: the flag appeared only at `:180`, `:400`, `:646` and `:843`.
- `QueryStream` spans `delegator.go:626-711` and contains **zero** `lifetime.` calls, unlike Search (`:481`), Query (`:716`), GetStatistics (`:833`) and UpdateSchema (`:1263`) — so `sd.lifetime.Wait()` at `:1391` does not drain it.
- `waitTSafe`'s fast path (`:1116-1119`) returns without rechecking serviceability.
- `ReleaseSegments` takes the signal at `delegator_data.go:1249-1251` and physically unloads at `:1274-1276`.
- Pre-existing and previously masked: master's `Close()` (`:817-823`) never expired the snapshot, so post-close removals blocked forever and the release never happened.

**One correction to the write-up, which narrows the window without closing it.** You wrote that `PinReadableSegments` gates only on `d.queryView.Serviceable()`, "which `Close()` never flips". That is accurate for `distribution.Close()`. But the reachability chain runs through `QueryStream`'s `sd.Serviceable()` at `delegator.go:628`, which is `sd.IsWorking(sd.lifetime.GetState())` (`:250-252`) — and that *is* flipped, by `sd.lifetime.SetState(lifetime.Stopped)` at the very top of `shardDelegator.Close()` (`:1386`), well before `sd.distribution.Close()` at `:1398`.

So a QueryStream arriving after `SetState` is rejected at `:628` and never gets near the pin. The real window is the in-flight batch: requests that passed `:628` *before* `SetState` and have not yet reached `PinReadableSegments` at `:664`, with `speedupGuranteeTS` and `waitTSafe` in between. It is bounded, but it is real, and QueryStream can be long-lived.

**Evidence stronger than the trace: five existing tests were relying on the hole.** After adding the guard, `TestPinReadableSegments` (all four sub-cases), `TestPinReadableSegments_ServiceableLogic`, `_LoadRatioLogic` and `_EdgeCases` all failed with `"channel distribution is closed" does not contain "channel distribution is not serviceable"`. Their setup does `dist.Flush(); dist.Close()` under the comment *"Stop background goroutine to avoid data race with mockey patches"* — they use `Close()` purely to quiesce the snapshot loop and then go on to pin successfully. That is the close-then-pin ordering succeeding in-tree today; the comment makes clear it was a convenience, not an intended contract. Those five sites now stop the goroutine directly through a `stopSnapshotLoop` test helper, which keeps what they actually wanted (`-race` on those tests is clean).

**Fix.** `PinReadableSegments` fast-fails on `d.closed` at the top, returning the same `merr.WrapErrChannelNotAvailable` shape the not-serviceable branch already returns (`:232`), so none of the four callers — Search `:533`, QueryStream `:664`, Query `:755`, GetStatistics `:853` — needed changing. As you noted, the check is atomic against close: the pin holds `d.mut.RLock` for its whole body and `Close()` holds `d.mut.Lock`, so either the pin wins and its `inUse` keeps the barrier up, or it observes `closed` and gives up. There is no third interleaving.

**Test.** Added `TestPinReadableSegmentsAfterCloseFails` — the ordering you pointed out is missing. It asserts `Serviceable()` is still true before `Close()`, so the case pins down that serviceability alone does not stop it, then requires an error and `version == -1` after. Reverting the guard makes it fail at `distribution_test.go:617-618` (`An error is expected but got nil`). Full `./internal/querynodev2/delegator/...` passes on the rebased head and `make static-check` reports 0 issues across all four scopes.

**Not fixed here, and flagging it rather than leaving it implicit:** `PinOnlineSegments` (`:277-289`) has the same structural hole — no `closed` check, same `current.Get()`. Its callers `forwardStreamingByBF` / `forwardStreamingDirect` (`delta_forward.go:197`, `:301`) come from `ProcessDelete` (`delegator_data.go:261`), which takes no `lifetime` either. I left it alone because it has no error return, so guarding it means either changing its signature or letting callers silently proceed on empty sealed/growing — a behavior change on the delete-forwarding path that wants its own test and review. Worth a follow-up; I have not audited what those callers would do with an empty result.

czs007:
Adversarial review of this PR is complete; one low-severity test-stability issue was identified, with no production logic defects found.

### Low

- **`internal/querynodev2/delegator/distribution_test.go:524`** (also line 543) — These tests use `time.NewTimer(100 * time.Millisecond)` as a positive wait for the snapshot-cleared signal, but that signal is closed by an asynchronous goroutine cascade (`snapshot.go:134-139`), and `Flush()` builds a snapshot chain before it fires. Under heavy CI load, scheduling delays beyond 100ms can fail the test even though the implementation is correct. Suggestion: use the 1-second bound already used by comparable tests in this file (e.g. line 498 and the `time.After(time.Second)` at line 589), or an `Eventually` assertion. The 100ms window remains appropriate only for the negative check at line 582. *(raised by xaxys)*

bigsheeper:
Confirmed and fixed in `53abfd7676` (also rebased onto latest master, so the earlier commits have new SHAs; content unchanged).

**`distribution_test.go:524` / `:543` — 100ms positive waits. Confirmed, widened to one second.**

Verified each part of the claim against the head you reviewed:

- Both are *positive* waits — `case <-timeout.C: s.Fail(...)` in `TestRemoveDistributionsAfterCloseReturnsClosedSignal` and `TestCloseClearsPendingRemoveDistributionSignal`.
- The signal really does arrive through an asynchronous cascade: `snapshot.go:133-139` closes `cleared` inside `go func() { <-s.last.cleared; ...; close(s.cleared) }()`, and `Flush()` builds the chain ahead of it.
- The in-file precedent you pointed at holds: `:498` already uses `time.NewTimer(time.Second)` and `:589` uses `time.After(time.Second)` for the same kind of wait.
- `:582` is the negative check (`case <-ch: s.Fail(...)`), where a short observation window is the point, so it keeps its 100ms.

**One correction to the framing, in my favour of taking it more seriously rather than less.** This is not adjacent flakiness I inherited — `git diff` against the merge-base shows both tests are **added by this PR**. And the inconsistency was internal to the PR: `:589`, which I wrote two commits later, already allowed a second. So the same change shipped two conventions for the same wait.

Widening does not weaken either assertion: they fail when the signal *never* arrives, and a one-second bound catches that exactly as well as 100ms. I checked that rather than asserting it — setting the bound to `time.Nanosecond` makes both fail with their own messages (`remove distribution signal not closed after distribution close`, `pending remove distribution signal not closed by distribution close`), so they are not vacuous at the wider bound.

I used the existing one-second timer rather than `Eventually`, to avoid introducing a third waiting idiom into a file that already has two.

`TestDistributionSuite` passes under `-count=3 -race`; full `./internal/querynodev2/delegator/...` passes on the rebased head; `make static-check` reports 0 issues across all four scopes.

czs007:

Verified:
- `internal/querynodev2/delegator/distribution.go:613` — `RemoveDistributions` retains the current snapshot's `cleared` channel after closure, so callers neither hang on an unexpired replacement snapshot nor bypass an active reader barrier.
- `internal/querynodev2/delegator/distribution.go:851` — `Close` takes the exclusive lock to set `closed` and expire, while `PinReadableSegments` checks `closed` under `RLock`, so every ordering either pins before expiration or rejects the late pin.
- `internal/querynodev2/delegator/distribution.go:648` — the closed-state guard prevents a queued notifier or concurrent snapshot caller from installing a fresh, unexpired snapshot after `Close`, which would otherwise leave the removal signal permanently open.
- `internal/querynodev2/delegator/snapshot.go:76` — `Expire` and `Done` both recheck `expired` and `inUse`, so either ordering closes the signal after the last reader drains, and `sync.Once` prevents duplicate cleanup or double channel closure.
- `internal/querynodev2/delegator/delegator_data.go:1249` — `ReleaseSegments` still waits on the returned signal, which now represents the exact snapshot whose readers must drain.
- `internal/querynodev2/delegator/distribution_test.go:514` — tests cover the key orderings: removal after close, close after signal capture, close with an actively pinned reader, release after unpin, and rejection of a pin after close.

bigsheeper:
Closing: superseded by #52167.

This PR fixed a hang where `ReleaseSegments` waited on a snapshot signal that the stopped background snapshot loop could no longer clear. #52167 made snapshot publication synchronous and **removed that background loop entirely**, so the state this PR guarded against no longer exists: the signal is now cleared by `genSnapshot` on the calling goroutine, and the reader barrier still holds via the snapshot's in-use count. The hang described in #51863 no longer reproduces on master.

The later commits here only covered holes opened by this PR's own approach — refusing readable pins after `Close()` was needed because this PR made `Close()` expire the current snapshot, and master's `Close()` no longer does.

Nothing left to rebase. Closing in favour of #52167.

Issue #51869, opened 2026-07-27
[Bug]: partition stats with vector centroids never deserialize (sonic vs encoding/json on null-into-interface)


xiaofan-luan:
## Describe the bug

Partition stats containing vector centroids can never be deserialized. `shardDelegator.loadPartitionStats` fails for every such snapshot with:

```
failed to parse partition stats from bytes  bytes_length=346
error="json: cannot unmarshal  into Go value of type storage.VectorFieldValue"
```

The error is only logged (`mlog.Error` + `continue`), never returned, so the delegator's `partitionStats` silently stays empty and segment pruning degrades to a full scan with no other signal.

## Root cause

`FieldStats.Centroids` is `[]VectorFieldValue` (an interface), which JSON cannot allocate directly. `FieldStats.UnmarshalJSON` worked around this by having `initCentroids` pre-allocate concrete `*FloatVectorFieldValue` values into the slice and letting the decoder merge into them.

That workaround depends on the decoder's merge-into-existing-value behaviour, which is not portable. `internal/storage` decodes via `internal/json`, which is **bytedance/sonic**, not `encoding/json`. Sonic rejects `null` into a non-empty interface where `encoding/json` silently leaves it nil:

```
STDLIB  null -> bloomfilter.BloomFilterInterface   err=<nil>
SONIC   null -> bloomfilter.BloomFilterInterface   err=json: cannot unmarshal  into Go value of type ...
```

A vector `FieldStats` always serializes `"bf":null`, and `initCentroids`' auxiliary struct declared a `bloomfilter.BloomFilterInterface` field it never even read. So on every vector field:

1. the auxiliary unmarshal fails on `bf`
2. `initCentroids` swallows the error with a bare `return`
3. nothing is pre-allocated
4. the real centroid decode then fails with the misleading `cannot unmarshal into VectorFieldValue`

## Why existing tests did not catch it

`TestVectorFieldStatsMarshal` covers this path but stayed green because it discards the `UnmarshalJSON` error return, and sonic fills the slice while still reporting the failure. The only caller that checks the error is `loadPartitionStats`, which just logs it.

## Steps to reproduce

```
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 \
  -run 'TestDelegatorDataSuite/TestLoadPartitionStats' ./internal/querynodev2/delegator/
```

Reproduces on current master (785ab5117c) on darwin/arm64, go1.26.5.

[...]

PR #51870, opened 2026-07-27
fix: decode field stats centroids explicitly instead of relying on decoder merge

xiaofan-luan:
issue: #51869

## What

`FieldStats.Centroids` is `[]VectorFieldValue`, an interface slice JSON cannot allocate into. `UnmarshalJSON` worked around that by having `initCentroids` pre-allocate concrete `*FloatVectorFieldValue` values and letting the decoder merge into them.

That depends on merge-into-existing-value behavior, which is not portable. This package decodes with `internal/json` (**bytedance/sonic**), and sonic rejects `null` into an interface **whose method set carries `UnmarshalJSON`** — which `bloomfilter.BloomFilterInterface` does — where `encoding/json` silently leaves it nil:

```
STDLIB  null -> bloomfilter.BloomFilterInterface   err=<nil>
SONIC   null -> bloomfilter.BloomFilterInterface   err=json: cannot unmarshal  into Go value of type ...
```

A vector `FieldStats` always serializes `"bf":null`, and `initCentroids`' auxiliary struct declared a `bloomfilter.BloomFilterInterface` field it never read. So on every vector field: the auxiliary unmarshal failed on `bf` → `initCentroids` swallowed it with a bare `return` → nothing was pre-allocated → the real centroid decode failed with a misleading `cannot unmarshal into VectorFieldValue`.

Partition stats with centroids therefore never loaded. `loadPartitionStats` only logs that error and continues, so the delegator's `partitionStats` stayed empty and **segment pruning silently degraded to a full scan**.

## Platform: the optdec decoder path

sonic's `sonic.go` build tag admits both amd64 and arm64, but the two dispatch to **different decoder implementations**:

- `internal/decoder/api/decoder_arm64.go` → `optdec` — the only choice on arm64
- `internal/decoder/api/decoder_amd64.go` → `jitdec` by default, but `optdec` when `SONIC_USE_OPTDEC=1` (`internal/envs/decode.go`)

Only `optdec` rejects `null` into an `UnmarshalJSON`-bearing interface. So this hits **arm64 by default, and amd64 as well whenever optdec is enabled**. Milvus does not set `SONIC_USE_OPTDEC`, so amd64 deployments are unaffected in practice today — but the trigger is the decoder, not the architecture.

That matches the evidence: `TestLoadPartitionStats` asserts `len(Centroids) == 2` and has been green on master in ci-v2 (linux/amd64, default `jitdec`) all along, while failing on darwin/arm64.

So on **ARM deployments** partition stats with centroids never loaded and segment pruning silently ran as a full scan — invisible to CI.

## How

Decode each centroid explicitly into its concrete type — no reliance on decoder merge semantics. Alongside that, in the same block:

- stop swallowing the decode failure (it now propagates)
- reject unsupported vector types loudly instead of producing an empty snapshot
- classify every failure on this path as `ErrDataIntegrity` (1009). Malformed array, malformed centroid, unsupported type and missing key all mean "the stats buffer on disk does not match the expected shape", which is what [`error_handling_casebook.md`](docs/dev/error_handling_casebook.md) assigns to `ErrDataIntegrity`; `WrapErrDataIntegrity` is documented for exactly this case ("json unmarshal of stats buffer"). Inner errors go in the error argument, not the format string (Pattern 4), so the sonic cause stays in the chain
- replace the previously unconditional deref of `"centroids"` (which could nil-panic) with a check that distinguishes the two cases. The field carries no `omitempty`, so a snapshot Milvus wrote always has the key — `null` when there is nothing to store. An explicit `null` decodes as empty; a **missing** key on a float vector is reported as a data-integrity error, because accepting it silently hands `FilterSegmentsByVector` an empty centroid set, which it treats as "search this segment" without logging — the same silent full-scan degradation this PR is fixing. Types without centroid support are still free to omit the key.

## Why existing tests missed it

`TestVectorFieldStatsMarshal` covers this path but stayed green for two reasons: it discarded the `UnmarshalJSON` error return (sonic fills the slice while still reporting failure), and it decoded into a `FieldStats` the test constructed itself rather than the zero-value elements production decodes into.

Both are fixed: the error is asserted, and `TestVectorFieldStatsDecodeIntoZeroValue` covers the shapes production actually uses — zero-value `FieldStats` inside a slice and a full `PartitionStatsSnapshot` round trip — plus every error path this change introduces: explicit `null`, missing key, unsupported vector type, malformed centroid, malformed centroid array — the last three asserting `errors.Is(err, merr.ErrDataIntegrity)` rather than just `Error()`.

## Verification

Based on master (785ab5117c), darwin/arm64, go1.26.5.

| | before | after |
|---|---|---|
| `TestDelegatorDataSuite` (incl. `TestLoadPartitionStats`) | FAIL | ok |
| `internal/querynodev2/delegator` (package) | FAIL | ok |
| `internal/storage` centroid tests | passed for the wrong reason | ok |
| `golangci-lint` misspell (`locale: US`) | 1 issue | 0 issues |

Confirmed failing on unmodified master before the change, so this is not a regression from any in-flight PR.

https://claude.ai/code/session_01FzrY94C2vx4GmVGCfTp8D6

### DIFF
--- a/internal/storage/field_stats.go
+++ b/internal/storage/field_stats.go
@@ -171,10 +171,10 @@ func (stats *FieldStats) UnmarshalJSON(data []byte) error {
 			}
 			stats.BF = bf
 		}
-	} else {
-		stats.initCentroids(data, stats.Type)
-		err = json.Unmarshal(*messageMap["centroids"], &stats.Centroids)
-		if err != nil {
+	} else if value, ok := messageMap["centroids"]; ok && value != nil {
+		// Guarded: an unsupported/vector-less type reaches this branch too, and
+		// dereferencing a missing "centroids" key used to panic.
+		if err := stats.unmarshalCentroids(*value, stats.Type); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


liliu-z (commented):
/lgtm
/approve

PR #51872, opened 2026-07-27
fix: revert "base==current CAS for the sort-stats and external-refresh manifest adoptions"

xiaofan-luan:
Reverts #51724 (`785ab5117c`).

Reverting to rework the change rather than patch it in place.

## Why

**1. Design rework.**

- *Attempt fencing.* #51724 carried `task_version` on the wire so a re-dispatched attempt could evict the resident one, with a version guard dropping the evicted attempt's late writes. That is the only worker in the tree with evict-on-conflict semantics — index / stats / analyze all reject a dispatch onto an occupied taskID (`LoadOrStoreIndexTask` / `LoadOrStoreStatsTask`). Rejecting removes the whole class of stale write instead of guarding against it, and takes the version off the wire entirely. Evicting is what forces the version to exist: it cancels the resident attempt, and a cancelled attempt's partial result then has to be fenced out of the entry the new attempt owns.

- *`base_manifest` placement.* #51724 added `base_manifest` to `SegmentInfo` — the segment entity — with a comment stating it must never be set on a persisted live segment. It is an adoption token belonging to the result, which is where the two comparable paths already put it (`StatsResult.base_manifest`, `CompactionSegment.base_manifest`). The refresh response reused `SegmentInfo` as its payload element, which is what left nowhere else to put it; the result should have its own message.

**2. Incomplete coverage.**

The same failure class is unaddressed for index metadata. An in-place external-refresh patch clears the segment's `TextStatsLogs` and `JsonKeyStats`, but nothing touches `indexMeta.segmentIndexes`, and `createIndexesForSegment` skips any segment that already has a `SegmentIndex` entry — so the index is never rebuilt. A segment can therefore carry refreshed data with a vector/scalar index built from the superseded data, and this does not need a concurrent commit: the purely sequential refresh path reaches it.

`SegmentIndex` has no record of the manifest it was built from, so `FinishTask` has no equivalent of the CAS this PR added for the manifest pointer. Landing the manifest CAS while that hole is open reads as "the adoption races are fixed" when one of the producers named in the original description is still unguarded.

## Consequence of the revert

The lost-update this fixed is reopened: manifest adoption returns to a version-newer check without verifying the result was built on the segment's current manifest, so a later write can silently overwrite a concurrent commit. Tracked in the audit issue.

issue: #51723

## Follow-up

Master and the 3.0 backport will be re-submitted together once the rework lands. #51831 (the 3.0 backport of this change, not yet merged) is being closed.

Verified: the revert applies cleanly on current master, restores the pre-#51724 state exactly for all 26 touched files (`git diff` against `785ab5117c^` for those paths is empty), and `./internal/...` builds.

https://claude.ai/code/session_01CMUfpxrLwZHnS5xZc2zEyW


xiaofan-luan:
Correction to this revert's stated reasons.

Reason 2 ("incomplete coverage" — that an in-place external-refresh patch leaves a stale index) does not hold. I filed it as #51873 and have since closed it as invalid:

- `validateExternalRefreshPatch` rejects any row-count change, so an in-place patch cannot alter the row set.
- The only two producers of an in-place patch append missing external columns or refresh deltalogs. Appending a column does not affect an index on another field, and deletions are applied as a query-time bitset rather than baked into the index.
- Schema-bump compaction performs the same class of in-place manifest change and also leaves `indexMeta` alone, deliberately — it rewrites no data.

Reason 1 (the design rework: reject-on-occupied task registration instead of evict-plus-version-fence, and moving `base_manifest` off the `SegmentInfo` entity onto a dedicated result message) stands, and is what the rework in #51874 implements. The revert itself was the right call; one of its two stated justifications was not.

Issue #49233, opened 2026-04-22
[Bug]: refresh against nonexistent S3 bucket loops forever (NoSuchBucket misclassified as transient)


zhuwenxing:
## Environment
- Milvus version: master-20260421-cd88c796fa
- Deployment mode: standalone
- MQ type: woodpecker
- SDK version: pymilvus 2.7.0rc198
- OS: Ubuntu 20.04.6 LTS (amd64)
- CPU/Memory: 64 cores / 528GB (host)
- GPU: N/A

### K8s Pod List
```
NAME                                                  READY   STATUS    RESTARTS      AGE
ext-table-master-etcd-0                               1/1     Running   0             4h35m
ext-table-master-milvus-standalone-59447ddfc7-hldbl   1/1     Running   5 (33m ago)   18h
ext-table-master-minio-7b74464967-n8f5l               1/1     Running   0             4h35m
```

## Current Behavior
Creating an external collection with `external_source=s3://bucket-does-not-exist-xxx/path/`
is accepted (the URL passes scheme validation). `refresh_external_collection`
on it returns a positive `job_id`. The job then loops forever in
`RefreshPending` state, with `reason=''` returned to the client (i.e. no
diagnostic surfaces through the progress API).

Observed:
```
create accepted: 's3://bucket-does-not-exist-adc81e/path/'
refresh job_id=465787408941891028
  t=5s  state=RefreshPending reason=''
  t=10s state=RefreshPending reason=''
  t=15s state=RefreshPending reason=''
  t=20s state=RefreshPending reason=''
  t=25s state=RefreshPending reason=''
```

DataCoord log every ~10 s:
```
[WARN] [datacoord/external_collection_refresh_manager.go]
  "async task creation failed, will retry on next checker tick"
  jobID=...
  error="... AWS Error NO_SUCH_BUCKET during ListObjectsV2 operation:
         The specified bucket does not exist: loon FFI transient error"
```

## Expected Behavior
`NoSuchBucket` is a **terminal** S3 error — no amount of retrying will turn a
nonexistent bucket into an existing one. The refresh job should:

1. Transition to `RefreshFailed` after the first attempt, with a clear
   `reason="external bucket does not exist"` (ideally without leaking internal
   endpoint DNS — see the related "DNS leak" bug report).
2. Stop consuming scheduler cycles for the lifetime of the collection.

The same applies to other terminal S3 errors: `NoSuchKey`, `AccessDenied`,
[...]

PR #51874, opened 2026-07-27
fix: base==current CAS for the sort-stats and external-refresh manifest adoptions

xiaofan-luan:
Rework of #51724, which was reverted in #51872. Rebased on post-revert master.

## What / why

The same StorageV3 segment manifest is advanced concurrently by several producers — an external-collection refresh column patch, a sort-stats result, and a text/JSON index build. They adopted a result by a *version-newer* check only, without verifying it was built on the segment's **current** manifest, so a later write could silently overwrite a concurrent commit (lost update). Audit: #51723.

## Unchanged from #51724

- **`base == current` CAS at the adoption sites.** A result is adopted only when the manifest it was built on still equals the segment's current manifest. The refresh CAS runs **inside** the `UpdateSegmentsInfo` / `segMu` critical section (in the upsert operator, via the synchronized `modPack.Get`) so the decision is atomic with the patch.
- **Adopt only a legal *successor*.** A matching base is not sufficient — a buggy / mixed-version / corrupt worker could carry the right base yet a result pointing at another segment or an older version. Extracted into the shared `validateManifestSuccessor` so the refresh and schema-bump paths cannot drift.
- **Rebuild on conflict**, refresh task failure classification, and "a task the worker no longer tracks re-dispatches rather than fails".

### 1. Task registration rejects instead of evicting

#51724 carried `task_version` on the wire so a re-dispatch could evict the resident attempt, with a version guard dropping the evicted attempt's late writes. That was the only worker in the tree with evict-on-conflict semantics.

Registration now **rejects** a dispatch onto an occupied taskID with `ErrTaskDuplicate`, matching `LoadOrStoreIndexTask` / `LoadOrStoreStatsTask`. DataCoord drops the resident attempt before re-dispatching — which the create path already did, and which is now load-bearing rather than best-effort. This removes the class of stale write instead of guarding against it, and takes the version off the wire entirely.

Why eviction forces a version to exist: eviction cancels the resident attempt, and a cancelled attempt's *partial* result then has to be fenced out of the entry the new attempt owns. Rejection leaves the resident attempt running to completion, so its result is never partial.

The attempt version stays on the persisted task record, where it still fences DataCoord's own attempt-scoped meta writes (`UpdateTaskState` / `ResetTaskForRetry`).

### 2. Result writes go through the attempt's own handle

Rejecting alone still leaves a window: a drop detaches the entry, the re-dispatch registers a fresh one under the same taskID, and the cancelled attempt's late write resolves the key to the *new* attempt. Writes now go through the `*TaskInfo` the attempt registered rather than a `(clusterID, taskID)` lookup, so a detached attempt writes into an object nobody reads.

### 3. `base_manifest` moves off the segment entity

#51724 added `base_manifest` to `SegmentInfo` — the segment entity — with a comment stating it must never be set on a persisted live segment. It is an adoption token belonging to the result, as it already is for `StatsResult.base_manifest` and `CompactionSegment.base_manifest`.

The result now has its own message carrying only what the worker produced:

```proto
message ExternalRefreshSegmentResult {
  int64 segment_id = 1;
  string manifest = 2;
  string base_manifest = 3;   // adoption token; empty for a new segment
  int64 num_rows = 4;
  int64 storage_version = 5;
  int32 schema_version = 6;
  repeated FieldBinlog binlogs = 7;
}
```

DataCoord materializes the `SegmentInfo` itself, so collection / partition / channel / state / level are no longer worker-supplied. A collection mismatch became **unrepresentable** rather than rejected at runtime.

`SegmentInfo.base_manifest` and `RefreshExternalCollectionTaskRequest.task_version` are dropped from the proto. Both were introduced by #51724 and are reverted out of master, so no released wire format is affected.

### 4. The stats CAS expected value comes from DataCoord, not the worker

The base used to be echoed back by the worker, which made the fence depend on the DataNode's version: a worker too old to report one sent an empty base, and an empty base had to be adopted (fail open) forever. That also meant an empty base was overloaded — legitimate sort birth, pre-CAS worker, or a bug that dropped it — so the third case was silently tolerated.

DataCoord already knows the value: it is the manifest it dispatched on. It is now stamped on the task record when the attempt is assigned:

```proto
message StatsTask {
  ...
  string dispatched_manifest = 14;
}
```

Stamped inside the existing `UpdateVersion` write, so no extra etcd round trip. Consequences:

- An old DataNode is fenced exactly like a current one — the compatibility exemption disappears rather than being carried indefinitely.
- A buggy or mixed-version worker cannot misreport its base; the value never passes through it.
- DataCoord and DataNode no longer have to ship together for this fence to behave.
- Empty now means exactly one thing: an attempt dispatched by a DataCoord predating the field and still in flight across the upgrade. That window drains in one upgrade cycle, after which this can become fail-closed.

The stamp and `CreateStatsRequest.manifest_path` must be the same string — a stamp staler than the request rejects every result, a stamp fresher than it silently adopts a stale one. Both are threaded from a single local in `CreateTaskOnWorker` and `prepareJobRequest` takes it as a parameter, so the two cannot be reordered apart or drift by re-reading the segment. `TestCreateTaskOnWorkerStampMatchesDispatchedRequest` pins it.

### 5. A Sort's cross-segment adoption is decided coordinator-side

A Sort reads the origin and writes a freshly allocated target, so its base describes a different segment than the one being written; comparing them rejects every sort result. #51724 handled this by having the worker blank the base for Sort — which is what overloaded "empty base" in the first place.

The coordinator now makes the call, since it is the only side that knows both segments:

```go
crossSegment := segmentID != originSegmentID
if !crossSegment && base != "" && currentManifest != base {
    return reject(nil)
}
```

The comparison uses the segment actually being written versus the origin, not `TargetSegmentID` directly, so a task record with an unset or stale `TargetSegmentID` still gets the in-place fence rather than skipping it.

The successor fence stays **unconditional**, so a sort retry onto a target an earlier attempt already materialized still cannot roll it backwards or point it at a third segment. Note the pre-existing sort tests all use `TargetSegmentID == SegmentID`; a real cross-segment sort had no coverage, and `TestSetJobInfoSortCrossSegmentSkipsBaseFence` adds it.

### 6. Rejected-result cleanup cannot be steered into another segment

`collectRejectedStatsResultFiles` feeds `MultiRemove`, so the manifest base path it resolves decides what gets **deleted**. It took that from `result.BaseManifest`. Once a Sort reports the origin's manifest (item 5), that would aim deletion at the origin's live files.

The base path now comes from a manifest DataCoord vouches for, belonging to the segment the attempt was writing. `result.Manifest` is not trusted either — pointing at a foreign or unparseable segment is one of the reasons a result gets rejected in the first place.

This is defusing, not fixing a live fault: it is unreachable today, because cleanup runs only for external collections and Sort is not submitted for external collections. The two gates are in unrelated files, which is exactly why it is worth removing.

### 7. Refresh meta: one lock, one critical section, one commit

The stale-manifest rollback used to run after the job lock was released. That window produced an `errExternalRefreshNotReady` signal, two terminal-state re-checks, and a version fence, all guarding against states the window created.

- The job/task lock pair collapses to a single job-aggregate lock (keyed by collectionID; a collection holds at most one active job, so it is job-granular in practice).
- `FinishJobWithApply` re-derives the all-tasks-finished aggregate **inside** the lock instead of trusting one the caller computed outside it, and performs the segment apply and the stale-manifest rollback in the same critical section.
- Finished commits together with the clearing of every consumed task result as **one composite write**, job last as the commit marker — previously 1 + N writes with an observable partial state.

Net: `errExternalRefreshNotReady`, `ClearTaskResult`, `ClearTaskResultsByJobID`, `AddTaskIDToJob`, `checker.resetJobTasksForRetry` and one lock are deleted.

### 8. Naming

`updatedSegments` → `segment_results` on the response and the persisted task, matching the `ExternalRefreshSegmentResult` element type and the existing `QueryCopySegmentResponse.segment_results`. Field numbers are unchanged, and neither message is protojson-serialized or exposed above DataCoord, so this is not a wire contract change. Reviewers may want to read this as a separate pass — it is mechanical and touches many of the same files.

### 9. Index metadata is fenced on the segment manifest

The revert's third reason — an in-place patch clears the segment's `TextStatsLogs` and `JsonKeyStats` but leaves `indexMeta.segmentIndexes` built on the superseded manifest — is closed here rather than deferred. Unlike (1) and (2) it needs no concurrent commit: the purely sequential refresh path reaches it.

`SegmentIndex` now records what the build read (`data_manifest`, field 23), stamped from the segment's `ManifestPath` at build creation. One field, three fences:

- **Adoption** — `applyExternalCollectionSegmentUpdate` removes affected indexes *before* publishing the patch, then repeats it best-effort after the commit to catch a build that started against the old manifest mid-apply.
- **Repair** — `getUnIndexTaskSegments` / `createIndexesForSegment` treat `DataManifest != segment.ManifestPath` as stale, drop the entry and rebuild. This is the durable signal; neither invalidation pass is atomic with the segment write, so it has to survive a crash or a failed cleanup.
- **Serving** — `GetIndexInfos` snapshots manifests and index rows together under `segMu` and filters stale entries, so a row that outlives a failed cleanup is never served.

The fence keys off the segment's current manifest rather than off the producer, so it covers any in-place manifest advance, not only refresh.

On #51723's open question — which patch shapes actually invalidate an existing index — this takes the conservative side: the index is dropped when the patch changes segment metadata at all (`!externalRefreshPatchIsNoop`) or when the recorded manifest already differs, rather than reasoning about which column groups a given patch left alone. Over-invalidating costs a rebuild; under-invalidating serves wrong data.

**Upgrade note:** a StorageV3 index built before this change has an empty `data_manifest` and is treated as stale once, so upgrading rebuilds it. That is deliberate — such a build has unknown provenance and may be exactly the case #51723 describes — but it is a one-time reindex to plan for on clusters already running StorageV3. StorageV2 segments have no manifest pointer, so the fence is inert there, as is the race.

### 10. An over-limit commit is split, not refused

A refresh commit used to be refused when it needed more than `metastore.maxEtcdTxnNum` (64) operations. With the Finished marker always taking one slot that capped a refresh at **63 changed segments**, which is not workable for real external collections — and it also mis-sized a crash replay, which re-sends every already-committed segment as a patch that writes nothing.

`UpdateSegmentsInfoWithActions` now writes the commit as several transactions when it does not fit in one, ordered so the **Finished marker lands strictly last**. That is the invariant worth keeping: readers are protected by "the marker is never durable before the segment generation it describes", not by "all segments move together". A crash between chunks therefore lands in the state the apply path already recovers from — segments committed, job not Finished — so the job stays non-terminal, the checker re-dispatches, and every already-committed segment returns as a full no-op the upsert operator skips. The marker is still folded into the final chunk when it fits, so a commit that did fit in one transaction is exactly as atomic as before.

The trade is a window in which the collection is partially refreshed, closed by that retry. Two further consequences, stated rather than hidden: a multi-chunk commit performs several etcd round trips while holding `segMu` (smaller than, but the same shape as, the L0 issue below), and in-memory segment state is advanced per successful chunk so it never lags what the catalog already holds.

### Rejected-attempt files are not reclaimed

By the time DataCoord refuses an attempt, the worker has already written its manifest — and any function-output column data or translated deltalogs — to final StorageV3 paths. `task_update.go`'s header states this as the design: *"failed manifests remain in final locations but are not registered in meta"*. The CAS adds one more way to land in that class rather than creating it; the pre-existing rejections (segment validation, version-newer) leak identically. `garbage_collector.go` skips V3 files under a registered segment ("orphan files managed by loon"), and a CAS rejection only ever applies to an already-registered segment, so general orphan GC does not reach these either.

It is deliberately not fixed here, because the obvious cleanup is unsafe in two independent ways. `AppendSegmentManifestColumns` inherits the old manifest's column groups, so a rejected manifest references files the **live** manifest still points at. And external source columns are referenced rather than copied — *"External input files are referenced, never copied"* — so deleting what a rejected manifest points at can reach into the customer's own bucket. The only unambiguously reclaimable object is the rejected manifest's own `_metadata/manifest-N.avro` record, which is kilobytes; the function-output data is the actual volume. Reclaiming that safely needs a loon-side "files exclusively owned by version N" primitive, which is a follow-up rather than something to improvise in Go here.

The stats path's `cleanupRejectedStatsResultFiles` is not a counter-example: it deletes text/JSON index files, which are Milvus-written and exclusively owned by the attempt, and it still validates every path against a DataCoord-vouched prefix before removing it.

### A failed explore cannot tell a bad source from a storage outage

The loon FFI boundary maps every storage failure onto the single `ErrLoonTransient` sentinel, so an unreachable, denied, absent or malformed source — `NoSuchBucket` included — is indistinguishable from a real S3 blip at `createTasksForJob`'s explore. The classification therefore stays on the terminal side: `ErrLoonTransient` fails the job, matching #49233. A genuine transient outage is surfaced as `RefreshFailed` and has to be re-issued.

Weighing it the other way — leaving it retriable — is worse: the job sits in `Init`/`RefreshPending` while the checker re-runs the same doomed explore for the full `ExternalCollectionJobTimeout` (1h default), and the user gets no signal at all. Closing this properly means giving the FFI boundary a precise error category, which belongs with the storage work rather than here.

### Index invalidation is whole-segment, and runs before the CAS

Two consequences of fencing indexes on the segment manifest, both accepted here rather than fixed:

- The manifest pointer is segment-granular, so it cannot express "the manifest moved but this field's column group did not". Any in-place patch therefore rebuilds **every** index on the segment — including, for a missing-column append, indexes over fields whose data was untouched. Narrowing this needs per-column-group staleness, not a different predicate.
- The pre-commit invalidation is a catalog write, while the CAS is evaluated later, inside `UpdateSegmentsInfo`'s operator. A rejected patch rolls the segment back but not the index removals, so a rejected attempt still costs a rebuild. Correctness is unaffected — the rebuild reads the unchanged current manifest — but the work is wasted, and the rejection path is the one this PR adds.

### L0 commits to object storage under the meta lock

`commitL0ManifestUpdates` still runs inside `m.segMu` in `UpdateSegmentsInfo`: the L0 delete-compaction object-storage commit, and its optimistic-retry loop, hold the global segment-meta lock across S3 I/O. Under object-store throttling that stalls every other segment-meta operation behind it. L0 is the only manifest-commit site that commits to object storage inside the DataCoord meta lock — bump and stats commit on the DataNode, off-lock.

This PR does not touch it; it is item (3) of #51723, and the audit's status list should be updated to reflect that (1), (2) and (4) land here while (3) stays open.

issue: #51723

## Tests

Carried over from #51724: `TestSetJobInfoSortResultManifestHandling`, `TestApplyExternalCollectionSegmentUpdate_StalePatchAborts`, `TestValidateManifestSuccessor`, refresh task failure classification, and "worker no longer tracks the task → Retry".

New for the redesign:
- `TestExternalCollectionManager_RejectsDispatchOntoResidentAttempt` / `_DetachedHandleCannotClobberSuccessor` / `_DeleteIsUnconditional`
- `TestApplyExternalCollectionSegmentUpdate_MaterializesIntoRequestedCollection`
- `TestSetJobInfoFencesOldDataNodeViaDispatchedManifest` — a worker reporting no base is still fenced off its dispatched manifest; only a pre-field DataCoord attempt fails open
- `TestCreateTaskOnWorkerStampMatchesDispatchedRequest` — the stamp and the request's manifest cannot diverge
- `TestSetJobInfoSortCrossSegmentSkipsBaseFence` — origin base on a fresh target is adopted; a retry rollback or foreign manifest on a materialized target is still refused
- `TestCollectRejectedStatsResultFiles/worker_supplied_manifests_cannot_steer_deletion`
- `TestExternalCollectionRefreshMeta_FinishJobWithApply` — terminal-state skip, tasks-not-all-finished skip, commit, stale-manifest rollback, and write-failure atomicity
- `TestSegmentIndexManifestStale` — the fence's truth table: nil and StorageV2 (empty/empty) are not stale, a pre-field StorageV3 record is rebuilt once, matching manifests are left alone
- `TestIndexInspector_RebuildsStaleManifestIndex` — an index whose `data_manifest` trails the segment is dropped and rebuilt by the inspector
- `TestApplyExternalCollectionSegmentUpdate_InvalidatesFakeFinishedIndex` — a Finished index over a refreshed segment does not survive the patch
- `TestCreateTaskOnWorkerFailureReleasesEnqueuedNodeLabel` — moving the compaction reserve write ahead of the RPC (item 1) also moved the read of `originNodeID`; this pins the `Executing` gauge to the label the enqueue filed it under. Negative control: with the read left after the reserve write it fails on both gauges (`-1` bucket not released, dispatched node driven negative)

## Verification boundary

Stating this explicitly so the claims above are not read as stronger than the evidence.

- All evidence is unit tests with mockey stubs. There is **no integration or e2e run**.
- Four behaviours were checked with negative controls — reverting the mechanism and confirming the new test fails: the cross-segment skip, the dispatched-manifest fence, the stamp/request invariant, and the deletion-path trust fix (which produced a delete path under a foreign segment, `files/insert_log/9/9/9999/...`).
- `-race` is clean on `internal/datacoord`, but these tests are largely single-goroutine, so that is **weak evidence for the lock unification** in item 7. That conclusion comes from auditing every `jobLock` acquisition site and every callee reachable inside the critical section, not from a concurrent stress test.
- The concurrent scenario the CAS exists for — an index build committing a manifest while a refresh is mid-apply — has **never been triggered end to end**; it is simulated by making `preApply` return the sentinel.
- The stats rebuild loop has no bound: a rejected result drops the task and the inspector re-submits on its next scan, which converges but is not capped by a retry count or timeout the way the refresh job is. JSON re-submission is additionally rate-limited by `JSONStatsTriggerCount` / `JSONStatsTriggerInterval`, so its retry can be delayed by a trigger interval.

Local run on macOS arm64: `internal/datacoord/...` and `internal/datanode/...` pass, plus `golangci-lint` with the repo config and `--build-tags dynamic,test`. The one failure, `TestCompactionTriggerManagerSuite` / `TestGetExpectedSegmentSize/all_DISKANN`, was confirmed to reproduce identically in a clean worktree at the base commit — local knowhere is built with `BUILD_DISK_ANN=OFF`, so `IsDiskVecIndex("DISKANN")` is false; Linux CI builds it on.

`data_coord.pb.go`'s large diff is the deterministic `[]byte` rawDesc re-wrap from changing fields; regenerated with the repo's `make generated-proto-without-cpp`.

https://claude.ai/code/session_01W6dkeNTbtN6cqoUn7cZqHY


xiaofan-luan:
### Code review

Found 1 issue:

1. Metrics drift on the failed-dispatch path of `mixCompactionTask.CreateTaskOnWorker` (bug due to reordering `updateAndSaveTaskMeta(setState(executing), setNodeID(nodeID))` to before the `CreateCompaction` RPC). `originNodeID := t.GetTaskProto().GetNodeID()` now reads the already-persisted new `nodeID` instead of the pre-attempt value, so on every failed dispatch the `Executing` gauge is decremented under `node=<nodeID>` — a bucket that was never incremented — while the `Executing{node="-1"}` bucket incremented at enqueue time (when NodeID is still `NullNodeID`) is never decremented. Repeated failures permanently inflate `DataCoordCompactionTaskNum{node="-1",state="executing"}` and drive the real-node bucket negative. Functional drop/retry behavior is unaffected (`cancelAndReset` uses the `nodeID` parameter). Fix: capture `originNodeID` before the reserve write, or decrement under `NullNodeID`.

https://github.com/milvus-io/milvus/blob/9e640741d99ccd768366771aca8c4df5e7fcfdca/internal/datacoord/compaction_task_mix.go#L113-L138

Enqueue-side increment for comparison:

https://github.com/milvus-io/milvus/blob/9e640741d99ccd768366771aca8c4df5e7fcfdca/internal/datacoord/compaction_inspector.go#L300-L302

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #51875, opened 2026-07-27
enhance: export filesystem metrics on scrape instead of on task completion


xiaofan-luan:
## Summary

The `milvus_storage_filesystem_*` metrics are collected correctly but exported on the wrong trigger: they are pushed only when a load / sync / compaction / index-build task finishes. Anything that reads object storage *outside* those events — most importantly vector raw data on the query path — never shows up.

## Background

The data source is milvus-storage's `FilesystemMetrics`: relaxed atomic counters incremented at the real object-storage calls (`S3Client::GetObject` / `PutObject`, and their local-filesystem equivalents in `local_fs_producer`). Milvus reads them over cgo (`loon_filesystem_get_metrics`) and publishes 8 gauges, registered for every role in `cmd/roles/roles.go`.

The push happens in exactly four places:

| Call site | Trigger |
|---|---|
| `internal/querynodev2/services.go` | after `LoadSegments` completes |
| `internal/flushcommon/syncmgr/task.go` | after each `SyncTask` |
| `internal/datanode/compactor/executor.go` | after a compaction task |
| `internal/datanode/index/scheduler.go` | after an index-build task |

## Problem

**1. Vector raw data read on the query path is invisible — on the default configuration.**

`queryNode.segcore.tieredStorage.warmup.vectorField` defaults to `disable`, and `CacheWarmupPolicy_Disable` is implemented as a plain `return;` — nothing is preloaded. So vector raw data is fetched from object storage the first time a query touches it, via `if (need_load) RunLoad(...)` on the pin path. That happens long after `LoadSegments` returned, so none of it is ever reflected in the gauges.

Not every query reads it — a search returning only ids and distances does not — but returning vector output fields, late-materialization rescoring, range search and index-less scans all do, and vector raw data is typically the largest part of a collection.

(The other three warmup policies default to `sync`, so scalar fields and indexes really are resident after load and genuinely do not read object storage at query time. For those, a frozen value is an accurate reflection.)

**2. Nothing is visible once eviction is enabled.**

`tieredStorage.evictionEnabled` defaults to `false`, so this does not bite by default — but eviction is the entire point of tiered storage. With it on, every evicted-then-reaccessed cell is a cold load on the query path, and none of it is reported.

**3. The exported value does not mean what its name says.**

[...]

PR #51876, opened 2026-07-27
enhance: export filesystem metrics on scrape instead of on task completion

xiaofan-luan:
## What & why

The `milvus_storage_filesystem_*` gauges are collected correctly — milvus-storage increments them at the real `S3Client::GetObject`/`PutObject` calls — but they are **pushed only when a load / sync / compaction / index-build task finishes**. Object storage read outside those events is never reflected.

That is not a corner case on the default configuration. `queryNode.segcore.tieredStorage.warmup.vectorField` defaults to `disable`, and `CacheWarmupPolicy_Disable` is implemented as a plain `return;` — nothing is preloaded. So **vector raw data is fetched from object storage the first time a query touches it**, via `if (need_load) RunLoad(...)` on the pin path, long after `LoadSegments` returned. Enabling eviction puts every cold reload on the same invisible path.

To be precise about the scope: the other three warmup policies default to `sync`, so scalar fields and indexes really are resident after load and genuinely do not read object storage at query time — for those, a frozen value is accurate. Vector raw data is the part that goes unaccounted, and it is usually the largest part of a collection.

The value is also mislabelled. `filesystem_read_bytes` reads as "bytes read so far", but actually means *"the process total as of the last task completion"*. On a query-serving node it is frozen and no dashboard reveals that — a number that looks current but is stale is worse than a missing one.

## Change

Export on the **pull model**: read the counters when Prometheus scrapes.

- The 8 `GaugeVec`s become `Desc`s plus a collector emitting `CounterValue` at scrape time. The values are monotonic process counters, so counter is the honest type, and `MustNewConstMetric` removes any need to track deltas.
- `pkg/` is its own module and cannot import the cgo storage package, so the provider is installed through a callback — the same pattern as the existing `SetPoolCollectFn`.
- The provider crosses cgo, so a failure or panic in it degrades to "no series this scrape" rather than taking down `/metrics` for everything else on the endpoint.

**Keeping multi-backend coverage.** Pulling only the default filesystem would have been a regression: sync, compaction and index tasks each carry their own `StorageConfig`, and the storage layer cannot enumerate live filesystems (`loon_filesystem_get_metrics` takes one handle; `FilesystemCache` exposes only `get`/`size`/`remove`). Those four call sites now **register** their config instead of publishing it — idempotent, no cgo, and cheaper than the per-task cgo call plus eight `Set`s they did before — and the collector reads every registered backend.

## Compatibility

Metric names and the `fs` label are unchanged. Only the type (`gauge` → `counter`) and the freshness of the value change, so existing queries keep working and `rate()` becomes meaningful.

The names deliberately keep **no `_total` suffix** despite becoming counters, so existing dashboards are not broken. This deviates from the Prometheus naming convention; several Milvus counters (e.g. `milvus_proxy_scanned_remote_mb`) are already in that position.

## Tests

`pkg/metrics` — counter type and values, multiple backends, silence when no provider is installed (emitting zeros would read as "no traffic" rather than "not measured"), and `/metrics` surviving a provider panic.

`internal/storagev2` — registration is idempotent (it is on the sync path), nil/keyless configs are ignored, the default backend is registered lazily on first scrape (paramtable is not populated at `init` time), and a registered backend round-trips through collection.

## Out of scope

- Splitting traffic by purpose (load vs query vs compaction) — the storage layer's counters are process-global; separating them needs per-context accounting inside milvus-storage.
- Per-collection / per-segment attribution — no such label exists.
- Read amplification (bytes fetched vs bytes actually needed) — additionally requires a "useful bytes" measure that does not exist yet.

issue: #51875

### DIFF
--- a/cmd/roles/roles.go
+++ b/cmd/roles/roles.go
@@ -38,6 +38,7 @@ import (
 	"github.com/milvus-io/milvus/internal/distributed/streaming"
 	"github.com/milvus-io/milvus/internal/http"
 	"github.com/milvus-io/milvus/internal/http/healthz"
+	"github.com/milvus-io/milvus/internal/storagev2"
 	"github.com/milvus-io/milvus/internal/util/dependency"
 	kvfactory "github.com/milvus-io/milvus/internal/util/dependency/kv"
 	"github.com/milvus-io/milvus/internal/util/initcore"
@@ -70,6 +71,9 @@ func init() {
 	metrics.RegisterMetaMetrics(Registry.GoRegistry)
 	metrics.RegisterMsgStreamMetrics(Registry.GoRegistry)
 	metrics.RegisterStorageMetrics(Registry.GoRegistry)
+	// The storage layer's counters are read at scrape time; pkg/metrics is its
+	// own module and cannot import this cgo package, hence the callback.
+	metrics.SetFilesystemStatsFn(storagev2.CollectFilesystemStats)
 }
 
 // stopRocksmqIfUsed closes the RocksMQ if it is used.
--- a/internal/datanode/compactor/executor.go
+++ b/internal/datanode/compactor/executor.go
@@ -158,11 +158,8 @@ func (e *executor) completeTask(planID int64, result *datapb.CompactionPlanResul
 
 		task.compactor.Complete()
 
-		// Publish filesystem metrics after compaction task completion
-		storageConfig := task.compactor.GetStorageConfig()
-		if _, err := storagev2.PublishFilesystemMetricsWithConfig(storageConfig); err != nil {
-			mlog.Warn(context.TODO(), "failed to publish filesystem metrics", mlog.Err(err))
-		}
+		// Make this backend visible to the filesystem metrics scrape.
+		storagev2.RegisterFilesystemConfig(task.compactor.GetStorageConfig())
 		return
 	}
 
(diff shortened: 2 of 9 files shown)


xiaofan-luan:
Withdrawing this: the scrape-time model it implements cannot be done safely with today's storage API.

`loon_filesystem_get` → `FilesystemCache::get` is **get-or-create**, backed by an LRU of capacity 16. Calling it from a Prometheus scrape means a scrape can construct an S3 client (including on roles that never touch storage), reorder the LRU, and evict a filesystem the workload is actively using — and past 16 backends the repeated rebuilds would reset the counters this PR exports as monotonic. A metrics scrape must not mutate what it observes.

Two further problems found in review, both recorded in #51875:
- the Go-side registry keyed by address+bucket is coarser than the real cache key, which also covers cloud_provider, region, SSL, timeouts, max_connections and credentials;
- non-default backends are only registered after a task succeeds, so they stay invisible during long runs and on the first failure.

Prerequisite is a read-only lookup/snapshot API in milvus-storage. #51875 has been rewritten to describe the problems and that dependency, with the proposed approach removed so nobody re-implements it from the issue.

Separately, #51875 now records a bug that exists on master independently of any of this: `makePropertiesFromConfig` omits `fs.max_connections` while the real storage-v2 path sets it, and `GetCacheKey()` hashes it — so with a non-default `minio.maxConnections` these metrics report 0 rather than stale values. That one is fixable on its own.

PR #51877, opened 2026-07-27
fix: stop handing Go dangling pointers from GetAnalyzeResultMeta

### LINKED ISSUE #51871: [Bug]: GetAnalyzeResultMeta hands Go dangling pointers into a destroyed local copy

### Environment

```markdown
- Milvus version: master (f875b692b3)
- Deployment mode: N/A (unit test / segcore)
- OS: observed on macOS arm64; the defect itself is platform-independent UB
```

### Current Behavior

`GetAnalyzeResultMeta` hands Go pointers into a C++ object that has already been destroyed, so the analyze (KMeans clustering) result paths that cross the cgo boundary are read from freed memory.

Observed in `TestIndexServiceSuite/Test_CreateAnalyzeTask/normal_case`: the analyze job itself succeeds, but the centroids path comes back as garbage bytes instead of a path:

```
[INFO] [index/task_analyze.go:190] ["analyze result"] [centroidsFile="���5�E� y"]
```

and the assertion fails:

```
Error: "\x8b\x7f\x98\xcfb\xf7\x96K\x10i\xc7" does not contain "/centroids"
```

### Expected Behavior

`centroidsFile` is the real centroids object path (containing `/centroids`), and `id_mapping_paths` are the real per-segment offset-mapping paths.

### Root Cause

`internal/core/src/clustering/analyze_c.cpp:179-189`:

```cpp
auto res = real_analyze->GetClusteringResultMeta();   // (1) local copy, by value
*centroid_path = res.centroid_path.data();            // (2) pointer INTO the local copy
*centroid_file_size = res.centroid_file_size;

auto& map_ = res.id_mappings;                         // (3) reference into the local copy
const char** id_mapping_paths_ = (const char**)id_mapping_paths;
size_t i = 0;
for (auto it = map_.begin(); it != map_.end(); ++it, i++) {
    id_mapping_paths_[i] = it->first.data();          // (4) pointers INTO the local copy
    id_mapping_sizes[i] = it->second;
}
```

`GetClusteringResultMeta()` returns **by value** (`internal/core/src/clustering/KmeansClustering.h:57-65`):

```cpp
ClusteringResultMeta
GetClusteringResultMeta() {
    ...
    return cluster_result_;
}
```

and `ClusteringResultMeta` owns heap-allocating members (`KmeansClustering.h:40-45`):

```cpp
struct ClusteringResultMeta {
    std::string centroid_path;
    int64_t centroid_file_size;
    std::unordered_map<std::string, int64_t> id_mappings;
};
```

So `res` is a local copy whose `std::string` buffers are freed when `GetAnalyzeResultMeta` returns. Every pointer written through `centroid_path` / `id_mapping_paths` dangles by the time the caller reads it.

The Go side then reads that freed memory (`internal/util/analyzecgowrapper/analyze.go:107`):

```go
centroidsFilePath := C.GoString(cCentroidsFilePath)
```

This is undefined behaviour rather than a deterministic failure. On glibc/x86 the freed block usually still holds the original bytes at the moment of the read, which is why CI does not catch it; a different allocator (macOS arm64) recycles the block immediately and the corruption becomes visible. Note the window is not empty even on Linux — the loop at (4) and the `status` writes run after the pointer at (2) has been handed out.

### Suggested Fix

`KmeansClustering` outlives the call (it is only destroyed by `DeleteAnalyze` / `at.analyze.Delete()`), so `cluster_result_` itself is a safe thing to point into. The minimal fix is to stop copying:

```cpp
// KmeansClustering.h
const ClusteringResultMeta&
GetClusteringResultMetaRef() {
    if (!is_runned_) {
        throw SegcoreError(ErrorCode::UnexpectedError,
                           "clustering result is not ready before kmeans run");
    }
    return cluster_result_;
}
```

```cpp
// analyze_c.cpp
const auto& res = real_analyze->GetClusteringResultMetaRef();
```

Alternative: `strdup` each string on the C++ side and free them on the Go side — heavier, because it also requires changing the Go ownership rules.

### Adjacent defect in the same function (worth fixing together)

`internal/util/analyzecgowrapper/analyze.go:92-94`:

```go
cCentroidsFilePath := C.CString("")
cCentroidsFileSize := C.int64_t(0)
defer C.free(unsafe.Pointer(cCentroidsFilePath))
```

Deferred call arguments are evaluated at the `defer` statement, so this frees the **initial empty-string** allocation, not whatever pointer the C++ side later writes into `cCentroidsFilePath`. Once the C++ side is changed to hand out a pointer the caller must not free (the reference fix above), the `defer C.free` should free the initial allocation explicitly and unambiguously — or the initial `C.CString("")` should be dropped entirely, since the value is overwritten before it is ever read.

### Steps To Reproduce

```bash
# on macOS arm64 (any allocator that recycles the freed block promptly)
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 \
  ./internal/datanode/ -run "TestIndexServiceSuite/Test_CreateAnalyzeTask"
```

Requires a local etcd and MinIO. The test reaches `JobStateFinished` and then fails the `s.Contains(centroidsFile, "/"+common.Centroids)` assertion with a garbage string.

### Anything else?

Found while running the datanode test suite locally; unrelated to the change being worked on. Filed separately so it can be fixed on its own.

xiaofan-luan:
issue: #51871

Fixes #51871.

## The bug

`GetAnalyzeResultMeta` handed Go pointers into a C++ object that was already destroyed:

```cpp
auto res = real_analyze->GetClusteringResultMeta();   // (1) local COPY, by value
*centroid_path = res.centroid_path.data();            // (2) pointer INTO the copy
*centroid_file_size = res.centroid_file_size;

const auto& map_ = res.id_mappings;
for (auto it = map_.begin(); it != map_.end(); ++it, i++) {
    id_mapping_paths_[i] = it->first.data();          // (3) and into the copy's map keys
    id_mapping_sizes[i] = it->second;
}
```

`GetClusteringResultMeta()` returned `ClusteringResultMeta` **by value**, and that struct owns a `std::string` plus an `unordered_map<std::string, int64_t>`. `res` is therefore a local copy whose buffers are freed when the function returns — every pointer written through `centroid_path` / `id_mapping_paths` dangles before the Go side reads it with `C.GoString`.

## The fix

`KmeansClustering` outlives the call — the Go side owns it until `DeleteAnalyze` — so `cluster_result_` is a safe thing to point into. The accessor now returns a const reference to the member instead of a copy, and the out parameter becomes `const char**` to match. cgo strips `const` from pointer types, so the Go call site needs no change.

Also drops the Go-side `C.CString("")` seed. The value is overwritten by the callee before it is ever read, and its

```go
defer C.free(unsafe.Pointer(cCentroidsFilePath))
```

was correct only by accident: deferred call arguments are evaluated at the `defer` statement, so it freed the initial empty-string allocation rather than the pointer C++ wrote back. Had it evaluated late — the intuitive reading — it would have freed C++-owned memory. Starting from a nil `*C.char` removes both the stray allocation and the trap.

## Why CI did not catch it

This is undefined behaviour, not a deterministic failure. On glibc/x86 the freed block usually still holds the original bytes at the moment of the read. A different allocator recycles it immediately, which is where it surfaced.

Note the window is not empty on Linux either: the id-mapping loop and the `status` writes both run after the pointer at (2) has been handed out.

## Verification

`TestIndexServiceSuite/Test_CreateAnalyzeTask` fails before this change and passes after it, on macOS arm64 with a locally rebuilt core.

Before — the analyze job completes, but the centroids path is 11 bytes of binary garbage, and the test fails on `s.Contains(centroidsFile, "/"+common.Centroids)`:

```
[INFO] [index/task_analyze.go:190] ["analyze result"] [centroidsFile="<garbage>"]
Error: "\x8b\x7f\x98\xcfb\xf7\x96K\x10i\xc7" does not contain "/centroids"
```

After:

```
[INFO] [index/task_analyze.go:190] ["analyze result"] [centroidsFile="index-service-ut/analyze_stats/200/1/1/2/111/centroids"]
ok  github.com/milvus-io/milvus/internal/datanode
```

https://claude.ai/code/session_01CMUfpxrLwZHnS5xZc2zEyW

### DIFF
--- a/internal/core/src/clustering/KmeansClustering.h
+++ b/internal/core/src/clustering/KmeansClustering.h
@@ -56,7 +56,14 @@ class KmeansClustering {
     Run(const milvus::proto::clustering::AnalyzeInfo& config);
 
     // should never be called before run
-    ClusteringResultMeta
+    //
+    // Returns a reference, not a copy, and that is load-bearing: the cgo
+    // boundary (GetAnalyzeResultMeta) hands the caller raw char* pointers into
+    // this struct's std::string members. A by-value return would make those
+    // pointers reference a local copy that is destroyed on return, so Go would
+    // read freed memory. The pointers stay valid for as long as this
+    // KmeansClustering object lives, which the Go side owns until DeleteAnalyze.
+    const ClusteringResultMeta&
     GetClusteringResultMeta() {
         if (!is_runned_) {
             throw SegcoreError(
--- a/internal/core/src/clustering/analyze_c.cpp
+++ b/internal/core/src/clustering/analyze_c.cpp
@@ -163,7 +163,7 @@ DeleteAnalyze(CAnalyze analyze) {
 
 CStatus
 GetAnalyzeResultMeta(CAnalyze analyze,
-                     char** centroid_path,
+                     const char** centroid_path,
                      int64_t* centroid_file_size,
                      void* id_mapping_paths,
                      int64_t* id_mapping_sizes) {
@@ -176,11 +176,14 @@ GetAnalyzeResultMeta(CAnalyze analyze,
                    "was null");
         auto real_analyze =
             reinterpret_cast<milvus::clustering::KmeansClustering*>(analyze);
-        auto res = real_analyze->GetClusteringResultMeta();
+        // Bind by reference: the pointers handed out below point into this
+        // struct's std::string members, so they must reference the object owned
+        // by real_analyze rather than a local copy that dies on return.
+        const auto& res = real_analyze->GetClusteringResultMeta();
         *centroid_path = res.centroid_path.data();
         *centroid_file_size = res.centroid_file_size;
 
-        auto& map_ = res.id_mappings;
+        const auto& map_ = res.id_mappings;
         const char** id_mapping_paths_ = (const char**)id_mapping_paths;
         size_t i = 0;
         for (auto it = map_.begin(); it != map_.end(); ++it, i++) {
--- a/internal/core/src/clustering/analyze_c.h
+++ b/internal/core/src/clustering/analyze_c.h
@@ -30,7 +30,7 @@ DeleteAnalyze(CAnalyze analyze);
 
 CStatus
 GetAnalyzeResultMeta(CAnalyze analyze,
-                     char** centroid_path,
+                     const char** centroid_path,
                      int64_t* centroid_file_size,
                      void* id_mapping_paths,
                      int64_t* id_mapping_sizes);
--- a/internal/util/analyzecgowrapper/analyze.go
+++ b/internal/util/analyzecgowrapper/analyze.go
@@ -89,9 +89,17 @@ func (ca *CgoAnalyze) Delete() error {
 func (ca *CgoAnalyze) GetResult(size int) (string, int64, []string, []int64, error) {
 	cOffsetMappingFilesPath := make([]unsafe.Pointer, size)
 	cOffsetMappingFilesSize := make([]C.int64_t, size)
-	cCentroidsFilePath := C.CString("")
+	// Out-parameters only: GetAnalyzeResultMeta overwrites this with a pointer
+	// into C++-owned memory (the KmeansClustering object's result meta, valid
+	// until DeleteAnalyze), which Go must not free. Starting from nil rather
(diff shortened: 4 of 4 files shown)


liliu-z (commented):
/lgtm
/approve

Issue #51643, opened 2026-07-21
[Bug]: StreamingNode panics after growing segment search invalid offset with Vortex flush error


yanliang567:
## Environment

- **Milvus Version**: `harbor.milvus.io/milvusdb/milvus:master-20260720-b9621d7d`
- **Deployment Mode**: cluster, deployed by Milvus Operator `1.3.6`
- **Milvus CR Version**: `2.7.0`
- **MQ**: Pulsar
- **SDK**: test SDK `2.7.0rc175`, client version parameter `2.6`
- **OS/K8s**: 4AM Kubernetes cluster, namespace `qa-milvus`

### Option B: Steps

1. Deploy a Milvus cluster with StreamingNode enabled:
   - instance: `qa-milvus/laion1b-test-2`
   - image: `master-20260720-b9621d7d`
   - `streamingMode=true`
   - 2 StreamingNodes, 6 QueryNodes, 4 DataNodes
2. Enable Vortex/Loon FFI storage format:
   - `common.storage.useLoonFFI=true`
   - `dataNode.storage.format=vortex`
3. Run mixed read/write workload:
   - Argo workflow: `qa/laion1b-test-cron-sn-nullable-1784556000`
   - case: `test_concurrent_locust_custom_parameters`
   - operations include `search`, `query`, `load`, and `upsert`
4. During the workload, one StreamingNode panics and restarts:
   - pod: `laion1b-test-2-milvus-streamingnode-75ff47cdc9-4jr6m`
   - exit code: `134`
   - K8s reason: `Error`, not `OOMKilled`

### Trigger Conditions

- **Frequency**: observed once in the latest mixed read/write run.
- **First observed after**: about 5h47m into the workflow.
- **Does NOT look like**:
  - client-only error: server-side StreamingNode panicked first
  - OOM: K8s last state is `Reason=Error`, `ExitCode=134`
  - insert/upsert rate limit: final upsert failures were `0`

## Expected Behavior

Milvus should not return out-of-range offsets during search result reduce/export on a growing segment. If growing source flush to Vortex/object storage fails, the error should be handled/retried without panicking and killing the StreamingNode process.

## Actual Behavior

The StreamingNode first logged an internal segcore offset consistency violation on a growing segment:

- segment: `467804729870388070`
- segment type: `Growing`
- row count: `43456`
- invalid offset: `43467`
- source: `../internal/core/src/segcore/reduce/Reduce.cpp:147`

[...]

PR #51878, opened 2026-07-27
fix: bound growing reads and make flush retries deterministic

xiaofan-luan:
issue: #51643

## What this fixes

Growing rows can be published into column storage and nullable offset mappings before the acknowledged row count advances. Search and flush must use the same frozen acknowledged prefix. At the same time, transient flush failures must preserve exactly one logical owner for the data: they must not enter the default panic path, duplicate an attempt, lose a parked pack during Drop, hide retained memory from backpressure, or commit channel metadata before segment data is durable.

This PR fixes both sides of that correctness boundary.

## Growing search correctness

- `SearchInfo` carries `active_count_`, frozen once by the plan layer.
- Nullable vector search converts that logical bound with `OffsetMapping::ValidCountBelow`, so it never scans rows beyond the same MVCC snapshot used by the non-nullable path.
- `GrowingOffsetMapping::Append` checks both physical and logical continuity. The binary search requires monotonic physical-to-logical order, so an out-of-order future insert fails loudly instead of silently returning unacknowledged rows.
- The growing flush admission bound uses `RowNum()`, the acknowledged prefix validated by segcore.

## Flush retry, memory, and cleanup correctness

- `SyncManager` is the sole owner of `HandleError`; `SyncTask.Run` no longer invokes the default panic callback before recovery can run.
- Ordinary sync has one per-segment logical state and at most one in-flight attempt. Retry generations reject stale timers, the same immutable task is reused, and a later batch cannot overtake a failed earlier batch.
- Follow-up ownership is installed before the previous logical completion is published, so backpressure never observes a full buffered tail without an owner.
- Ordinary insert, delete, and BM25 stats payload remains in `MemorySize()` and `DataNodeFlowGraphBufferDataSize` after yield. It is removed exactly once only when `ReleaseData()` actually runs.
- BM25 stats enter the existing insert threshold from the buffered stage onward. The already-accounted byte estimate is frozen and transferred to the sync task, so a runtime `bm25StatsBytesPerEntry` refresh cannot make the gauge negative or create an accounting gap.
- Per-segment backpressure reuses the existing `FlushInsertBufferSize` and `FlushDeleteBufferBytes` thresholds. The combined retained payload plus current write-buffer tail blocks `BufferData` after the triggering task has been submitted, so oversized batches can still make progress without allowing storage outages to accumulate invisible packs.
- The memory manager counts all resident ordinary payload but chooses eviction candidates only from immediately evictable buffers, avoiding a busy loop when the watermark is dominated by in-flight flushes.
- After object storage succeeds, the task freezes binlog metadata and releases the row payload. A metadata-only retry does not rewrite object data or continue consuming payload capacity.
- Success, cancellation, fatal errors, callback panics, dispatcher shutdown, and retry submission failure converge on exactly-once accounting and cleanup.
- `Drop` and `SaveAll` atomically take ownership of pending ordinary syncs, drain ordinary and growing data under their context, and call `DropChannel` only after persistence succeeds. The streaming flusher lifecycle context now reaches writeNode, BufferManager, WriteBuffer, and the final broker `DropVirtualChannel` RPC; BufferManager also applies `GracefulStopTimeout` as an upper bound.
- Dispatcher close fences semaphore waiters, async handoff, queued cancellation, callbacks, futures, and task accounting.

## Scope

This PR introduces no separate process-wide packed-memory budget, lease, or weighted byte semaphore. Ordinary backpressure reuses the existing per-segment flush thresholds and DataNode memory watermark. Growing-source data remains owned by segcore and is not counted as a yielded pack.

Storage-layer retryable/permanent classification remains out of scope; retry behavior uses the classifications currently exposed by the sync path.

## Validation

- CI-aligned golangci-lint 2.11.3 for flushcommon, storage, and streaming flusher packages - 0 issues
- explicit govet and test depguard checks - 0 issues
- remote focused Go tests passed for payload exact-once accounting, storage failure retention, metadata-only release, BufferData backpressure, memory-manager no-spin, bounded Drop context propagation, BM25 buffered/in-flight accounting, and BM25 config-refresh transfer
- remote full pipeline and streaming flusher package tests passed
- race tests passed for payload concurrent release, BufferData backpressure, and the complete write-buffer base suite
- local compile-only checks passed for writebuffer, pipeline, storage, and streaming flusher packages
- `git diff --check`, gofmt, gofumpt, and gci checks
- remote C++ fast build - 737 targets completed
- remote `OffsetMapping.*` - 17/17 passed

The local macOS test binaries cannot execute because the environment does not provide `libmilvus-storage.dylib`; the Go tests above were executed on the configured Linux development host with the CI build tags and Mockey gcflags.

### DIFF
--- a/internal/core/src/common/OffsetMapping.cpp
+++ b/internal/core/src/common/OffsetMapping.cpp
@@ -370,6 +370,12 @@ GrowingOffsetMapping::Append(const bool* valid_data,
                "into one growing segment must stay serial.",
                physical_idx,
                valid_count_);
+    AssertInfo(start_logical == total_count_,
+               "growing offset mapping requires logical offsets to be appended "
+               "in order: got start_logical={}, expected {}. Inserts into one "
+               "growing segment must stay serial.",
+               start_logical,
+               total_count_);
     for (int64_t i = 0; i < count; ++i) {
         if (valid_data[i]) {
             const auto logical_offset = start_logical + i;
--- a/internal/core/src/common/OffsetMapping.cpp
+++ b/internal/core/src/common/OffsetMapping.cpp
@@ -1,6 +1,7 @@
 #include "common/OffsetMapping.h"
 
 #include <algorithm>
+#include <limits>
 #include <mutex>
 #include <shared_mutex>
 #include <utility>
@@ -51,6 +52,12 @@ OffsetMapping::IsValid(int64_t logical_offset) const {
 int64_t
 OffsetMapping::GetValidCount() const {
     return 0;
+}
+
+int64_t
+OffsetMapping::ValidCountBelow(int64_t logical_bound) const {
+    // No mapping: physical space == logical space.
+    return std::max<int64_t>(0, logical_bound);
 }
 
 bool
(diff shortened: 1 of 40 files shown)


xiaofan-luan:
This PR has been split into three focused replacements:

- #51960 — Segcore growing search and field-map snapshots
- #51958 — deterministic, memory-bounded flush retry and Drop lifecycle
- #51959 — vector test-data generator OpenMP overhead

The three PRs have disjoint scopes whose union exactly matches this PR's 40 changed files and `3403 additions / 378 deletions`. They are rebased onto current `master`; #51960 also preserves the newer upstream `SearchOnGrowing` schema-snapshot change.

Closing this PR as superseded. The branch remains available as the fully integrated reference and previously validated CI head.

PR #51040, opened 2026-07-03
enhance: Reduce QueryNode distribution report payload

sijie-ni-0214:
## Summary

issue: #51039

Reduce the repeated `GetDataDistribution` RPC payload between QueryNode and QueryCoord by adding delta report support.

This PR lets QueryCoord advertise delta support with its last applied distribution timestamp. QueryNode tracks dirty and removed sealed segments plus dirty and removed channel leader views, then returns a delta when it is safe to do so. QueryCoord applies delta responses through `Patch` methods on the segment and channel distribution managers, while full reports remain the fallback for unsupported or stale requests.

## Changes

- Add delta fields to `GetDataDistributionRequest` and `GetDataDistributionResponse`.
- Track QueryNode distribution deltas for segment upsert/remove and channel leader view upsert/remove.
- Build full reports or delta reports based on QueryCoord support and timestamp compatibility.
- Patch QueryCoord segment/channel distribution managers for delta responses.
- Optimize QueryNode metrics collection by replacing allocation-heavy segment grouping with `Segment.RangeBy`, and make `RangeBy` stop as soon as the visitor returns false.
- Preserve node stats correctness in delta responses with total segment/channel counts.
- Add focused tests for delta tracking, report construction, patch application, and failure-path leader view refreshes.

## Known trade-off

Delta reporting intentionally does not mark a channel for every pure streaming row-count or delta-watermark update, since doing so would make hot DML continuously advance the distribution timestamp and reduce the payload-saving benefit.

Channel membership and correctness-sensitive leader-view changes are still reported through explicit marks, including new growing segments, flush/seal/handoff, target synchronization, load/release, `SyncDistribution`, and readiness/serviceability changes. The remaining accepted staleness is limited to `NumOfGrowingRows` and segment `LastDeltaTimestamp` until the next marked channel update or full report.

## Test Plan

- [x] `git diff --check`
- [x] `source scripts/setenv.sh && go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" -ldflags="-extldflags \"-Wl,-rpath,${MILVUS_WORK_DIR}/cmake_build/lib -Wl,-rpath,${MILVUS_WORK_DIR}/internal/core/output/lib -Wl,-rpath,${MILVUS_WORK_DIR}/cmake_build/thirdparty/boost_ext\"" ./internal/querynodev2/segments -run TestManager -timeout 600s`
[...]

Issue #51039, opened 2026-07-03
[Enhancement]: Reduce GetDataDistribution RPC payload size


sijie-ni-0214:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Optimize the QueryNode `GetDataDistribution` report path so QueryCoord can request and apply incremental distribution changes instead of receiving the full segment distribution and leader view payload every time.

The optimization should preserve the existing full-report behavior as a fallback, while allowing compatible QueryCoord and QueryNode versions to exchange deltas for changed sealed segments and channel leader views.

### Why is this needed?

`GetDataDistribution` is called repeatedly by QueryCoord to keep QueryNode distribution state up to date. On deployments with many loaded sealed segments and channels, the current full report can become large because every response includes all segment version info, channel version info, and leader views even when only a small subset changed.

This increases RPC payload size and repeated processing cost on both QueryNode and QueryCoord, especially for workloads with frequent small distribution updates such as load/release, target sync, index changes, or leader view refreshes.

### Anything else?

A proposed implementation can add delta support to `GetDataDistribution` by:

- letting QueryCoord advertise delta support and send its last applied distribution timestamp;
- tracking dirty/removed segments and dirty/removed channel leader views on QueryNode;
- returning full reports when delta cannot be built safely;
- patching QueryCoord's segment and channel distribution managers when a delta response is received;
- keeping total segment/channel counts in delta responses for node stats.

PR #51881, opened 2026-07-27
enhance: [cp2.6] Reduce QueryNode distribution report payload

sijie-ni-0214:
Cherry-pick from master
pr: #51040

issue: #51039

## Summary

Backport delta support for `GetDataDistribution` so QueryCoord can patch QueryNode segment and channel distribution state instead of receiving a full report for every update.

- Track dirty and removed sealed segments and channel leader views on QueryNode.
- Build delta reports when the QueryCoord timestamp is compatible, with full reports as the safe fallback.
- Patch QueryCoord segment and channel distribution managers from delta responses.
- Preserve node statistics using total segment and channel counts in delta responses.
- Reduce QueryNode metrics allocations with in-place `RangeBy` aggregation and early termination.

## 2.6 adaptation

The 2.6 IndexChecker still updates sealed-segment indexes through `LoadScope_Index`, while master uses the newer reopen path. Mark index-only loads as segment upserts so the following delta report carries the refreshed index metadata. The focused regression test verifies the affected segment IDs and `IndexInfo` contents.

## Test Plan

- [x] `git diff --check upstream/2.6...HEAD`
- [x] `make static-check`
- [x] Native build probe: `build_required=false`
- [x] `go test -count=1 -tags dynamic,test -gcflags='all=-N -l' ./internal/querynodev2 -run 'TestQueryNodeService/TestLoadIndex_Success$' -timeout 300s` (with the required macOS rpaths and SDK sysroot)


czs007:

**Verified:**

- `dist_handler.go:201-203` — the coord unconditionally skips L0 segments and the QN delta path filters L0 out of segment marks, so `SegmentDistManager` cannot diverge on L0 between the two report modes.
- `channel_dist_manager.go:314` — `Update` fully replaces a node's channel set, so delta reporting a transiently non-serviceable channel via `removedChannelNames` reproduces exactly the coord state the base already produced; not a regression.
- `services.go` (`GetDataDistribution`, no-change early return) — the `!hasDataDistributionChange` branch returns before `clearDelta()` / `lastReportedTs = 0`, so a "no change" poll never drops accumulated dirty or removed marks.
- `services.go` (full-report path) — even when `canBuildDelta` is false, the marker maps are detached and cleared and `lastReportedTs` is set to `lastModifyTs` under the generation guard, so full→delta and delta→full transitions re-seat the delta base without lost or double-counted state.
- `delegator.go` (`NewShardDelegator`) — `distribution` is assigned in the struct literal before the `opts` loop runs, so `WithLeaderViewUpdatedCallback` reliably sets both `sd.leaderViewUpdatedCallback` and `sd.distribution.leaderViewUpdatedCallback`; serviceability transitions do reach the tracker.
- `distribution.go` (`updateServiceable`) — reordering `loadedRatio.Store` ahead of the `queryView.Serviceable()` read preserves value semantics, and the callback stays edge-triggered on a real `old != new` transition (covered by `TestDistribution_UpdateServiceableNotifiesOnChange`).
- Protocol compatibility — the new request/response fields are additive, so an old QN ignores `support_delta` and returns `IsDelta=false`, and a new QN sees `canBuildDelta == false` against an old coord; mixed-version clusters degrade safely to full reports.

czs007 (approved):
/lgtm

PR #51882, opened 2026-07-27
fix: [2.6] size the scalar-index valid bitmap by active count on growing segments (#51523)

xiaocai2333:
issue: #51237
pr: #51523

Backport of #51523 to 2.6. Not a clean cherry-pick — the 2.6 expression layer differs, see "Differences from master" below.

## Root cause

A growing segment carries a segment-level interim R-Tree index for a geometry field as soon as the collection has an `RTREE` index on it, making geometry the only field type whose expressions take the **ScalarIndex** path on a growing segment. The index bitmap is segment-global (a scalar index always has exactly one chunk), but `ProcessIndexOneChunk()` / `ProcessIndexOneChunkForValid()` sliced it with `size_per_chunk_` as the upper bound — on a growing segment that is the raw-data chunk granularity (`segcore.chunkRows`, 128 by default), unrelated to the bitmap.

On 2.6 the visible symptom is the sealed-only assert rather than the truncated bitmap, because 2.6 already carried a narrow `GEOMETRY && Growing` special case in `ProcessIndexOneChunkForValid` that masked the truncation. A `geometry_field is not null` query over more than one chunk of growing rows crashes with:

```
Assert "segment_->type() == SegmentType::Sealed" => index mode only for sealed segment
```

## Fix

- `ProcessIndexOneChunk()` / `ProcessIndexOneChunkForValid()`: slice by `active_count_` (the rows visible to this query) instead of `size_per_chunk_`, and assert the cached bitmap covers the requested range instead of silently clamping to its length. No-op on sealed segments, where `size_per_chunk() == get_row_count()`. The old third `min` term was also wrong on its own — it clamped against the bitmap's *full* length rather than the length remaining after `data_pos`, so a short bitmap plus `data_pos > 0` made `append()` read past the end. This subsumes the `GEOMETRY && Growing` special case, which bounded by the bitmap size — that may run ahead of `active_count_` under concurrent inserts.
- `MoveCursorForIndex()`: drop the sealed-only assert — the index cursor is a global row position for a growing segment with a segment-level scalar index too.

## Differences from master

- 2.6 has no `execute_all_at_once_` fast path in the index-chunk loop, so the corresponding commit of #51523 ("guard the all-at-once index fast path") has no counterpart here. The slicing lives in the per-chunk helpers rather than inline in `ProcessIndexChunks()`.
- The GIS `MoveCursor()` growing fix that #51523's 3.0 backport (#51583) carries is **already on 2.6** via #51487, so it is not repeated here.

## Tests

The three regression tests from #51523, ported as-is:

- `TestGeometryNullExprOnGrowingWithRTreeIndex` — `is not null` / `is null` on a growing segment with the R-Tree index, plus a batched conjunction and the short-circuit path.
- `TestGeometrySpatialExprOnGrowingWithRTreeIndex` — real spatial predicates, indexed vs raw growing segment over the same rows, must select identical rows.
- `TestGeometryConjunctionShortCircuitOnGrowingWithRTreeIndex` — conjunction whose numeric side short-circuits the early batches, anchored on a standalone spatial run rather than an indexed-vs-raw comparison alone.

## Verification (local, 2.6 worktree, rebased on 759be43b1e)

Built with `make build-cpp-with-unittest` and ran the C++ unit tests:

- **Positive**: the three new tests pass — 162 parameterized instances; together with the neighbouring `GISFilterGrowing*` / `GISFilterDoesNotSupportOffsetInput` tests from #51487, 165 pass.
- **Negative control**: reverting only `Expr.h` to stock 2.6 (keeping the tests) makes `TestGeometryNullExprOnGrowingWithRTreeIndex` fail — 18 failures, all `Assert "segment_->type() == SegmentType::Sealed" => index mode only for sealed segment` at `Expr.h:281`. So the fix is load-bearing on today's 2.6.
- **Scope note, stated plainly**: on this base the other two tests pass with or without the change — `TestGeometryConjunctionShortCircuitOnGrowingWithRTreeIndex` is already covered by #51487's GIS `MoveCursor()` fix, and `TestGeometrySpatialExprOnGrowingWithRTreeIndex` is coverage for the growing R-Tree spatial path (as on master). They are kept because they pin behavior this change touches, not because they reproduce a live 2.6 failure.
- **No regressions**: full `ExprTestSuite/*` — 4428 tests, all pass. This matters because the `active_count_` slicing applies to every scalar-index expression, sealed included.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -278,8 +278,9 @@ class SegmentExpr : public Expr {
 
     void
     MoveCursorForIndex() {
-        AssertInfo(segment_->type() == SegmentType::Sealed,
-                   "index mode only for sealed segment");
+        // The index cursor is a global row position. This holds for sealed
+        // segments and for growing segments with a segment-level scalar
+        // index (the geometry interim R-Tree, see issue #51237).
         auto size =
             std::min(active_count_ - current_index_chunk_pos_, batch_size_);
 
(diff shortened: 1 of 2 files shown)


liliu-z (commented):
/lgtm
/approve

PR #51884, opened 2026-07-27
enhance: unify RecordReader ownership on a borrow contract

MrPresent-Han:
### What

Unifies `RecordReader.Next()` on a single borrow contract and aligns every consumer, so record reference counts balance on all compaction paths. First of a phased series split out of #51848.

- **storage** — records returned by `Next()` are reader-owned and valid until the next `Next`/`Close`. The V1 `CompositeBinlogRecordReader` now holds and releases its current record, mirroring the packed/manifest readers, which already work this way.
- **compactor** — the record materializer releases only its derived arrays (function outputs, missing-field fills, selection columns) through an idempotent cleanup channel (`cleanupMaterializedRecord`) and never the borrowed base record. `retainBase` and `releaseWrappedRecord` are removed.
- **consumers** — the bump/mix direct paths stop releasing reader-owned records; `DeleteCodec` and the external deltalog / function-executor paths no longer release reader-owned records; the sort path closes the outermost reader wrapper so the timestamp-overwrite wrapper releases its last record.
- **cross-batch holders** — `storage.Sort` keeps records across reader advances through its existing explicit `Retain`/`Release`, which balances under the unified contract.

### Why

Consumers treated reader records as caller-owned on some paths and reader-owned on others. Under the Arrow `assert` build tag, the direct-consumer paths (bump / mix on storage v2/v3) report `too many releases`, and the wrapper-reader paths balanced only through a compensating `retainBase`. One declared contract removes the asymmetry.

### Tests

- Full `internal/datanode/compactor` suite green under `-tags dynamic,test,assert` (Arrow refcount assertions enabled) — the direct-consumer paths fail there without this change.
- New refcount unit tests: the V1 reader owns its current record (verified with `memory.CheckedAllocator`), cleanup leaves the borrowed base untouched, and a retained record survives a reader advance and balances to zero.
- Targeted `internal/storage` and full `internal/datanode/external` suites green under the same tags.

issue: #51649

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -520,20 +520,17 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		if preMaterializeFilter {
 			selection, _, err = selectFullRewriteRecord(record, pkField, entityFilter, ttlFieldID, sourceHasTTLField, nil)
 			if err != nil {
-				record.Release()
 				return nil, err
 			}
 			if selection != nil && selection.Len() == 0 {
-				record.Release()
 				continue
 			}
 		}
 
+		// record stays owned by the reader (released on its next Next/Close);
+		// only the derived arrays of the wrapped record are cleaned up here.
 		wrapped, err := materializer.WrapWithSelection(record, selection)
 		if err != nil {
-			if selection == nil {
-				record.Release()
-			}
 			return nil, err
 		}
 		out := overwriteRecordTimestamps(wrapped, segment.GetCommitTimestamp())
@@ -541,7 +538,7 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 			if out != wrapped {
 				out.Release()
 			}
-			releaseWrappedRecord(wrapped, record)
+			cleanupMaterializedRecord(wrapped)
 			return nil, err
 		}
 		if out != wrapped {
@@ -549,7 +546,7 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		}
 
 		totalRows += int64(wrapped.Len())
-		releaseWrappedRecord(wrapped, record)
+		cleanupMaterializedRecord(wrapped)
 	}
 
 	if err := writer.Close(); err != nil {
@@ -1021,10 +1018,11 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx 
 		readDuration += time.Since(readStart)
 
 		computeStart := time.Now()
+		// record stays owned by the reader (released on its next Next/Close);
+		// only the derived arrays of the wrapped record are cleaned up here.
 		wrapped, err := materializer.Wrap(record)
 		computeDuration += time.Since(computeStart)
 		if err != nil {
-			record.Release()
 			span.End()
 			return nil, err
 		}
@@ -1033,7 +1031,7 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx 
 			outputFieldID := outputField.GetFieldID()
 			outputCol := wrapped.Column(outputFieldID)
 			if outputCol == nil {
-				releaseWrappedRecord(wrapped, record)
+				cleanupMaterializedRecord(wrapped)
 				span.End()
 				return nil, merr.WrapErrServiceInternalMsg("output field %d not found in materialized record", outputFieldID)
 			}
@@ -1041,7 +1039,7 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx 
 			if stats, ok := statsByField[outputFieldID]; ok {
 				bm25MemSize, err := appendBM25StatsFromArrowArray(stats, outputCol)
 				if err != nil {
-					releaseWrappedRecord(wrapped, record)
+					cleanupMaterializedRecord(wrapped)
 					span.End()
 					return nil, err
 				}
@@ -1052,7 +1050,7 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx 
 
 		writeStart := time.Now()
 		if err := writerResult.writer.Write(wrapped); err != nil {
-			releaseWrappedRecord(wrapped, record)
+			cleanupMaterializedRecord(wrapped)
 			span.End()
 			return nil, err
 		}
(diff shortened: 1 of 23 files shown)


internal/storage/serde_delta.go
```diff
@@ -232,35 +231,37 @@ func (crr *simpleArrowRecordReader) Next() (Record, error) {
...
+			field2Col[FieldID(i)] = i
 		}
-		crr.r.r = record
-		return true
+		return NewSimpleArrowRecord(record, field2Col), true
 	}
 
-	if ok := composeRecord(); !ok {
-		if err := crr.iterateNextBatch(); err != nil {
+	for {
```
MrPresent-Han, line 252: Why a loop instead of the previous single-retry: the old flow (`if !ok { advance once; if still !ok → io.EOF }`) conflated three distinct conditions behind `Next()==false`, causing two silent data-loss modes:

1. **Read error swallowed as exhaustion.** arrow's `Next() bool` folds "exhausted" and "read failed" into `false`; the distinction only exists in `Err()`, which was never consulted. A corrupt blob mid-stream was treated as "done", silently dropping the remaining delete records (deleted rows would resurrect after compaction). Note the `err != io.EOF` guard: pqarrow stores `io.EOF` in `Err()` on *normal* exhaustion, so only non-EOF errors are real failures.

2. **Premature EOF on an empty blob.** With `[non-empty, empty, non-empty]`, the single retry was spent crossing into the empty blob; the second `composeRecord` miss then declared EOF, discarding every blob after the hole. Current writers likely never produce zero-row deltalogs, but the reader should not depend on that unstated invariant.

The loop makes "true EOF" have exactly one exit: `iterateNextBatch` returning `io.EOF` when it advances past the last blob. Each iteration handles one blob transition, so any number of consecutive empty blobs is skipped uniformly. Covered by `TestSimpleArrowRecordReaderEmptyMiddleBlob` and `TestSimpleArrowRecordReaderSurfacesReadError` (both fail against the previous flow).

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -520,36 +520,33 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
...
+			cleanupMaterializedRecord(wrapped)
 			return nil, err
 		}
 		if out != wrapped {
 			out.Release()
 		}
 
 		totalRows += int64(wrapped.Len())
-		releaseWrappedRecord(wrapped, record)
+		cleanupMaterializedRecord(wrapped)
```
liliu-z, line 549: In runFullSchemaRewrite, when the segment has a non-zero commit timestamp (import/CDC) and some rows are dropped by deletes/TTL so selection != nil, `wrapped` is or wraps a `selectedRecord` whose projected columns are built lazily on first Column() access. overwriteRecordTimestamps snapshots refcounts via `rec.Retain()` before writer.Write, but the real writer materializes those columns afterward, so they escape the snapshot; the following `out.Release()` (a full Release over every column) and this `cleanupMaterializedRecord(wrapped)` (cleanupDerived over selectedRecord.columns) then each release the late-created columns once, driving them to refcount -1 — a double-free that panics under the Arrow assert tag and corrupts memory otherwise. The same double-release is on the error path at line 541. Records with a fixed column set (compositeRecord base, no selection) are unaffected because Retain/Release are exact inverses; the PR fixes the borrowed-base accounting but not these derived columns, so the 'balance on all compaction paths' claim does not hold on the selection + commit-ts path. It is latent today because the compactor tests mock the writer and never combine a non-zero commit timestamp with deletes/TTL. Fix by materializing selectedRecord.columns before the timestamp wrap; do not drop cleanupMaterializedRecord, which would instead leak the computed/pre-snapshot columns at refcount 1.

MrPresent-Han: Thanks — the structural point is verified and fixed in 695915c3b4, with one correction on the concrete failure mode.

**Fixed as suggested:** selection columns are now built eagerly (`newSelectedRecord` slices every `existingFields`-gated schema field up front), so the column set is fixed before `overwriteRecordTimestamps` takes its Retain snapshot, and `Retain`/`Release` stay exact inverses. `cleanupMaterializedRecord` is kept, per your note.

**Correction on the failure mode:** while writing the refcount proof for this fix I found `RecordBuilder.Build()` never released its builder-side creator refs — `array.NewRecord` retains every column, and the `NewArray` ref was simply orphaned, so every `Build()`-produced column carried a permanent +1. That changes the arithmetic of the path you traced: the "extra" release on the commit-ts path consumed that orphaned ref, landing exactly at 0 — no −1 / double-free actually occurred (which is why the `assert` builds stayed green). The real defect was the mirror image: the non-commit-ts path left every selection column at refcount 1, reclaimed only by GC. Your structural conclusion still stands — a lazily created column escapes the snapshot — and if `Build()` had been fixed first, the lazy path would have become exactly the double-free you described.

Both ends are addressed now: eager column build + `Build()` releases its creator refs (all four production `RecordBuilder` call sites audited; none relied on the orphaned ref). Coverage added: CheckedAllocator lifecycle tests for selection columns (exact zero on success, mid-build failure, and partial build with an already-built column), plus a `RecordBuilder` ownership contract test — each proven non-vacuous (they fail with leak reports when either half of the fix is reverted).

internal/storage/serde_delta.go
```diff
@@ -232,35 +231,37 @@ func (crr *simpleArrowRecordReader) Next() (Record, error) {
...
+		if rec, ok := composeRecord(); ok {
+			return rec, nil
+		}
+		// Next()==false means either batch exhaustion or a read error;
+		// pqarrow stores io.EOF in Err() on normal exhaustion
+		if err := crr.rr.Err(); err != nil && err != io.EOF {
```
liliu-z, line 258: The new crr.rr.Err() branch returns the pqarrow decode error untyped, so merr.Code falls through to errUnexpected (65535) — mislabeling a corrupt in-memory deltalog payload as an opaque internal bug. Elsewhere the storage layer classifies malformed binlog/parquet bytes as ErrDataIntegrity (see payload_reader.go); wrap it the same way with merr.WrapErrDataIntegrity(err, "read deltalog record batch"). Impact is observability-only — both codes are non-retriable and nothing branches on the class — so this is low severity.

MrPresent-Han: Fixed in 07b5788efc: the branch now returns `merr.WrapErrDataIntegrity(err, "read deltalog record batch")`, matching the payload_reader.go classification for corrupt stored bytes (1009, non-retriable). The inner pqarrow error carries no merr code, so the classification factory does not mask anything, and the error chain is preserved — the test asserts both `ErrorIs(io.ErrUnexpectedEOF)` and `ErrorIs(merr.ErrDataIntegrity)`. Also audited consumers: the only type-sensitive checks on this path are `err == io.EOF` loop terminators, which the branch explicitly excludes, so no control flow changes.

liliu-z (commented):
/lgtm
/approve

Issue #50406, opened 2026-06-09
[Enhancement]: Persist segment summary metadata on SegmentInfo


tedxu:
Today DataCoord derives aggregate segment metrics (total size,
deletion count, file counts, etc.) by iterating SegmentInfo's
FieldBinlog arrays on every scheduling decision. For V3 segments
the per-field binlog KV entries in etcd are pure write
amplification because the manifest is already the authoritative
path source.

Add a Statistics summary field to SegmentInfo so DataCoord reads
aggregates directly from the persisted proto.

Requirements:

1. DataCoord scheduling decisions (compaction trigger, GC, load
   balancing) read aggregate metrics from the persisted proto,
   without iterating FieldBinlog arrays or reading manifests
   from object storage.

2. Aggregate metadata survives DataCoord restart.

3. V3 segments do not write per-field binlog KV entries to
   etcd. The manifest is the source of truth for paths; only
   aggregate summaries are persisted in etcd.

4. Backward compatible: existing V2 segments and existing V3
   segments that currently have binlog KVs continue to work.

5. Aggregate values stay consistent with the underlying data.
   For V3 segments they are extracted at manifest generation;
   for V2 segments they are computed from binlog arrays.

PR #51885, opened 2026-07-27
enhance: read StorageV3 segments via manifest in clustering analyze

tedxu:
See #50406

Clustering compaction analyze (both scalar and mapping paths)
reads segment data via newCompactionSegmentRecordReader, which
already handles V3 manifests. However the analyze callers still
gate on FieldBinlogs to derive binlogNum and field sets. After
a DataCoord restart V3 segments have empty FieldBinlogs (data
is in the manifest), so the analyze paths need to obtain field
info from the manifest when FieldBinlogs is absent.

issue: #50406


internal/core/src/clustering/KmeansClustering.cpp
```diff
@@ -82,71 +102,97 @@ KmeansClustering::FetchDataFiles(uint8_t* buf,
...
+    const int64_t expected_remote_file_size,
+    const std::string& manifest_path,
+    const Config& base_config,
+    int64_t& offset) {
+    // base_config already carries STORAGE_VERSION/DATA_TYPE/ELEMENT_TYPE/DIM
+    // (built once in Run, mirroring index_c.cpp get_config). Only the manifest
+    // varies per segment.
+    Config config = base_config;
+    config[SEGMENT_MANIFEST_KEY] = manifest_path;
+    auto field_datas = file_manager_->CacheRawDataToMemory(config);
```
congqixia, line 128: I found one issue around the claimed memory bound for V3 sampling.

  `KmeansClustering::FetchSegmentViaManifest` calls `CacheRawDataToMemory` first, and only then `CopyFieldDatasToBuf` truncates the copied bytes to `expected_train_size`. But the
  manifest path eventually goes through `GetFieldDatasFromManifest`, which reads all record batches into `field_datas` before returning. So for a large StorageV3 segment, random
  sampling can still load the whole segment into temporary memory even when only a small train-size budget remains.

  This means the train-size memory bound is not actually preserved for V3, despite the PR description. The new C++ test sets `train_size == full segment size`, so it does not cover the
  partial-sampling case.

  Could we either make the manifest reader support bounded/streaming reads, or change the sampling path so it stops reading once the train budget is reached? A regression test with
  `train_size < single V3 segment size` would also be helpful.

congqixia (commented):
overall lgtm
please check NIT comment

internal/core/src/clustering/KmeansClustering.cpp
tedxu: Good catch. You’re right this is separate from the sampling-granularity change.

This PR already changed training sampling to segment-level/inter-segment by shuffling segment IDs and reading whole segments until the train budget is reached, so V1/V2/V3 are aligned on that point.

But for the V3 manifest path, `FetchSegmentViaManifest()` still goes through `CacheRawDataToMemory()`, and the manifest reader builds `field_datas` before `CopyFieldDatasToBuf()` truncates to `train_size`. So while the sampling granularity changed, the temporary memory bound for partial V3 sampling is not actually guaranteed yet.

Also agreed the current test only covers the full-read case, since `train_size == full segment size`.

I’ll update the PR description/comment to avoid over-claiming here, and we can either handle bounded/streaming manifest reads in this PR or follow up with a dedicated fix and regression test for `train_size < single V3 segment size`.

congqixia (approved):
/lgtm

Issue #51796, opened 2026-07-24
[Enhancement] Collapse QueryNode schema versioning to a single global schema.Version (remove barrier-ts / segcoreSchemaVersion)


xiaofan-luan:
## Summary

QueryNode currently orders schema updates across **two** version domains:
- the logical, globally-assigned `schema.Version`, and
- a node-local barrier timestamp `schemaBarrierTs` plus a derived node-local monotonic `segcoreSchemaVersion` counter (introduced in #50577 to fix the out-of-order-schema panic #50364, and to let `ttl_field` / collection-property refreshes reach segcore without bumping `schema.Version`).

This proposes **collapsing to a single, globally-assigned `schema.Version`**: every schema DDL — including collection property / `ttl_field` alters that carry a schema snapshot — bumps `schema.Version`; segcore gates on `schema.Version`; `schemaBarrierTs` + `segcoreSchemaVersion` are removed.

## Motivation

The dual-domain model is a recurring source of complexity and bugs. The `add_function_field` reopen load-wins race (#50989 / #51062) required reasoning across both domains, and any per-node version reservation carries collision hazards. Because `schema.Version` is assigned once by RootCoord, a reopen load and the stream `UpdateSchema` carry the **same** version for the same schema — no local allocation, no reservation, no collision. Net ~ -270 lines.

## Approach

- **RootCoord**: property / `ttl_field` schema-snapshot alters set `schema.Version = current + 1` (previously kept the same version and were ordered by barrier-ts).
- **QueryNode**: `prepareCollectionSchemaUpdate` orders strictly by `schema.Version` (`<= current -> skip`; anti-rollback / out-of-order-delivery protection from #50364 preserved as the `<` case; same-version is a no-op; `>` applies). `segcoreSchemaVersion` collapses to `schema.Version`; barrier-ts plumbing removed across collection/delegator/pipeline/services/querycoord. The in-flight load fence migrates from wall-clock `schemaBarrierTs` to the load's `schema.Version` vs the delegator's served version.
- **segcore (C++)**: `Schema::ParseFrom` sources `schema_version_` from the proto version, so the CCollection/segment version gates run in the `schema.Version` domain.
- Proto `schema_barrier_ts` left defined-but-unused (wire-stable, no field-number reuse).

## Relationship

[...]

PR #51886, opened 2026-07-27
fix: gate segment load on delegator schema version

congqixia:
Related to #51796

Reject load segment and reopen requests when the delegator has not consumed the matching schema version from WAL yet. The gate is implemented in the delegator, because all load requests must pass through it.

Maintain an independent delegator schema view that is advanced only by WAL UpdateSchema events. Load requests may observe that view, but must not advance it, must not update delegator function runners, and must not share schema ownership with other QueryNode components.

Stop using SchemaBarrierTs as the delegator load gate. Keep it only as a compatibility/update timestamp field for workers and collection manager same-version schema payloads.

Treat ErrCollectionSchemaVersionNotReady as a temporary QueryCoord load result: keep the segment task started, leave rpcReturned false, and let scheduler retry after the delegator catches up.

Add focused coverage for the delegator schema gate, load-path schema isolation, QueryCoord retry behavior, and the versioned merr wrapper.

### DIFF
--- a/internal/querycoordv2/task/executor.go
+++ b/internal/querycoordv2/task/executor.go
@@ -222,13 +222,23 @@ func (ex *Executor) executeSegmentAction(task *SegmentTask, step int) {
 // not really executes the request
 func (ex *Executor) loadSegment(task *SegmentTask, step int) error {
 	action := task.Actions()[step].(*SegmentAction)
-	defer action.rpcReturned.Store(true)
+	markRPCReturned := true
+	defer func() {
+		if markRPCReturned {
+			action.rpcReturned.Store(true)
+		}
+	}()
 	ctx := task.Context()
 
 	var err error
 	defer func() {
 		if err != nil {
-			task.Fail(err)
+			if errors.Is(err, merr.ErrCollectionSchemaVersionNotReady) {
+				markRPCReturned = false
+				task.SetReason(err.Error())
+			} else {
+				task.Fail(err)
+			}
 		}
 		ex.removeTask(task, step)
 	}()
@@ -278,7 +288,11 @@ func (ex *Executor) loadSegment(task *SegmentTask, step int) error {
 	status, err := ex.cluster.LoadSegments(task.Context(), view.Node, req)
 	err = merr.CheckRPCCall(status, err)
 	if err != nil {
-		mlog.Warn(context.TODO(), "failed to load segment", mlog.Err(err))
+		if errors.Is(err, merr.ErrCollectionSchemaVersionNotReady) {
+			mlog.Info(ctx, "load segment waits for delegator schema version", mlog.Err(err))
+		} else {
+			mlog.Warn(ctx, "failed to load segment", mlog.Err(err))
+		}
 		return err
 	}
 
--- a/internal/querycoordv2/task/executor.go
+++ b/internal/querycoordv2/task/executor.go
@@ -27,6 +27,8 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/samber/lo"
 	"go.uber.org/atomic"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
@@ -233,7 +235,7 @@ func (ex *Executor) loadSegment(task *SegmentTask, step int) error {
 	var err error
 	defer func() {
 		if err != nil {
-			if errors.Is(err, merr.ErrCollectionSchemaVersionNotReady) {
+			if isLoadSegmentPendingError(err) {
 				markRPCReturned = false
 				task.SetReason(err.Error())
 			} else {
@@ -288,8 +290,8 @@ func (ex *Executor) loadSegment(task *SegmentTask, step int) error {
 	status, err := ex.cluster.LoadSegments(task.Context(), view.Node, req)
 	err = merr.CheckRPCCall(status, err)
 	if err != nil {
-		if errors.Is(err, merr.ErrCollectionSchemaVersionNotReady) {
-			mlog.Info(ctx, "load segment waits for delegator schema version", mlog.Err(err))
+		if isLoadSegmentPendingError(err) {
+			mlog.Info(ctx, "load segment waits for transient runtime readiness", mlog.Err(err))
 		} else {
 			mlog.Warn(ctx, "failed to load segment", mlog.Err(err))
 		}
@@ -300,6 +302,25 @@ func (ex *Executor) loadSegment(task *SegmentTask, step int) error {
 	mlog.Info(context.TODO(), "load segments done", mlog.Duration("elapsed", elapsed))
 
 	return nil
+}
+
+func isLoadSegmentPendingError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if merr.IsRetryableErr(err) {
+		return true
+	}
+	if errors.IsAny(err,
+		context.DeadlineExceeded,
+		grpc.ErrClientConnClosing,
+		merr.ErrChannelNotAvailable,
+		merr.ErrNodeNotAvailable,
+		merr.ErrNodeNotFound,
+	) {
+		return true
+	}
+	return funcutil.IsGrpcErr(err, codes.Unavailable, codes.DeadlineExceeded, codes.Aborted)
 }
 
 // If we enable following checking when loading segments,
--- a/internal/querycoordv2/task/task_test.go
+++ b/internal/querycoordv2/task/task_test.go
@@ -181,6 +181,7 @@ func (suite *TaskSuite) BeforeTest(suiteName, testName string) {
 		"TestLoadSegmentTask",
 		"TestLoadSegmentTaskNotIndex",
 		"TestSegmentTaskWaitsDistAfterLoadRPC",
+		"TestLoadSegmentTaskWaitsDelegatorSchema",
 		"TestLoadSegmentTaskFailed",
 		"TestTaskCanceled",
 		"TestMoveSegmentTask",
@@ -637,6 +638,86 @@ func (suite *TaskSuite) TestLoadSegmentTaskNotIndex() {
 		suite.Equal(TaskStatusSucceeded, task.Status())
 		suite.NoError(task.Err())
 	}
+}
+
+func (suite *TaskSuite) TestLoadSegmentTaskWaitsDelegatorSchema() {
+	ctx := context.Background()
+	timeout := 10 * time.Second
+	targetNode := int64(3)
+	partition := int64(100)
+	segment := suite.loadSegments[0]
+	channel := &datapb.VchannelInfo{
+		CollectionID: suite.collection,
+		ChannelName:  Params.CommonCfg.RootCoordDml.GetValue() + "-test",
+	}
+
+	suite.broker.EXPECT().DescribeCollection(mock.Anything, suite.collection).RunAndReturn(func(ctx context.Context, i int64) (*milvuspb.DescribeCollectionResponse, error) {
+		return &milvuspb.DescribeCollectionResponse{
+			Schema: &schemapb.CollectionSchema{
+				Name:    "TestLoadSegmentTaskWaitsDelegatorSchema",
+				Version: 1,
+				Fields: []*schemapb.FieldSchema{
+					{FieldID: 100, Name: "vec", DataType: schemapb.DataType_FloatVector},
+				},
+			},
+		}, nil
+	})
+	suite.broker.EXPECT().ListIndexes(mock.Anything, suite.collection).Return([]*indexpb.IndexInfo{
+		{
+			CollectionID: suite.collection,
+		},
+	}, nil)
+	suite.broker.EXPECT().GetSegmentInfo(mock.Anything, segment).Return([]*datapb.SegmentInfo{
+		{
+			ID:            segment,
+			CollectionID:  suite.collection,
+			PartitionID:   partition,
+			InsertChannel: channel.ChannelName,
+		},
+	}, nil)
+	suite.broker.EXPECT().GetIndexInfo(mock.Anything, suite.collection, segment).Return(nil, nil)
+	schemaErr := merr.WrapErrCollectionSchemaVersionNotReadyWithVersion("TestLoadSegmentTaskWaitsDelegatorSchema", 0, 1)
+	suite.cluster.EXPECT().LoadSegments(mock.Anything, targetNode, mock.Anything).Return(merr.Status(schemaErr), nil).Once()
+
+	suite.dist.ChannelDistManager.Update(targetNode, &meta.DmChannel{
+		VchannelInfo: channel,
+		Node:         targetNode,
+		Version:      1,
+		View: &meta.LeaderView{
+			ID:           targetNode,
+			CollectionID: suite.collection,
+			Channel:      channel.ChannelName,
+			Status:       &querypb.LeaderViewStatus{Serviceable: true},
+		},
+	})
+	task, err := NewSegmentTask(
+		ctx,
+		timeout,
+		WrapIDSource(0),
+		suite.collection,
+		suite.replica,
+		commonpb.LoadPriority_LOW,
+		NewSegmentAction(targetNode, ActionTypeGrow, channel.GetChannelName(), segment),
+	)
+	suite.NoError(err)
+	suite.NoError(suite.scheduler.Add(task))
+
+	segments := []*datapb.SegmentInfo{{
+		ID:            segment,
+		InsertChannel: channel.ChannelName,
+		PartitionID:   1,
+	}}
+	suite.broker.EXPECT().GetRecoveryInfoV2(mock.Anything, suite.collection).Return([]*datapb.VchannelInfo{channel}, segments, nil)
+	suite.target.UpdateCollectionNextTarget(ctx, suite.collection)
+	suite.AssertTaskNum(0, 1, 0, 1)
+
+	suite.dispatchAndWait(targetNode)
+	suite.AssertTaskNum(1, 0, 0, 1)
+	suite.Equal(TaskStatusStarted, task.Status())
+	suite.NoError(task.Err())
+	suite.Contains(task.GetReason(), "collection schema version not ready")
+	action := task.Actions()[0].(*SegmentAction)
+	suite.False(action.rpcReturned.Load())
 }
 
 func (suite *TaskSuite) TestLoadSegmentTaskFailed() {
--- a/internal/querycoordv2/task/task_test.go
+++ b/internal/querycoordv2/task/task_test.go
@@ -661,12 +661,12 @@ func (suite *TaskSuite) TestLoadSegmentTaskWaitsDelegatorSchema() {
 				},
 			},
 		}, nil
-	}).Twice()
+	}).Times(3)
 	suite.broker.EXPECT().ListIndexes(mock.Anything, suite.collection).Return([]*indexpb.IndexInfo{
 		{
 			CollectionID: suite.collection,
 		},
-	}, nil).Twice()
+	}, nil).Times(3)
 	suite.broker.EXPECT().GetSegmentInfo(mock.Anything, segment).Return([]*datapb.SegmentInfo{
 		{
 			ID:            segment,
@@ -674,10 +674,11 @@ func (suite *TaskSuite) TestLoadSegmentTaskWaitsDelegatorSchema() {
 			PartitionID:   partition,
 			InsertChannel: channel.ChannelName,
 		},
-	}, nil).Twice()
-	suite.broker.EXPECT().GetIndexInfo(mock.Anything, suite.collection, segment).Return(nil, nil).Twice()
+	}, nil).Times(3)
+	suite.broker.EXPECT().GetIndexInfo(mock.Anything, suite.collection, segment).Return(nil, nil).Times(3)
 	schemaErr := merr.WrapErrCollectionSchemaVersionNotReadyWithVersion("TestLoadSegmentTaskWaitsDelegatorSchema", 0, 1)
 	suite.cluster.EXPECT().LoadSegments(mock.Anything, targetNode, mock.Anything).Return(merr.Status(schemaErr), nil).Once()
+	suite.cluster.EXPECT().LoadSegments(mock.Anything, targetNode, mock.Anything).Return(merr.Status(merr.WrapErrNodeNotAvailable(targetNode, "worker churn")), nil).Once()
 	suite.cluster.EXPECT().LoadSegments(mock.Anything, targetNode, mock.Anything).Return(merr.Success(), nil).Once()
 
 	suite.dist.ChannelDistManager.Update(targetNode, &meta.DmChannel{
@@ -718,6 +719,13 @@ func (suite *TaskSuite) TestLoadSegmentTaskWaitsDelegatorSchema() {
 	suite.NoError(task.Err())
 	suite.Contains(task.GetReason(), "collection schema version not ready")
 	action := task.Actions()[0].(*SegmentAction)
+	suite.False(action.rpcReturned.Load())
+
+	suite.dispatchAndWait(targetNode)
+	suite.AssertTaskNum(1, 0, 0, 1)
+	suite.Equal(TaskStatusStarted, task.Status())
+	suite.NoError(task.Err())
+	suite.Contains(task.GetReason(), "node not available")
 	suite.False(action.rpcReturned.Load())
 
 	suite.dispatchAndWait(targetNode)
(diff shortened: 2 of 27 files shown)


internal/querynodev2/delegator/delegator.go
```diff
@@ -205,6 +211,63 @@ type shardDelegator struct {
...
+	}
+	return typeutil.Clone(sd.schemaView.schema), sd.schemaView.version
+}
+
+func (sd *shardDelegator) shouldUpdateDelegatorSchemaLocked(schema *schemapb.CollectionSchema) bool {
+	if schema == nil {
+		return false
+	}
+	sd.ensureDelegatorSchemaViewLocked()
+	return uint64(schema.GetVersion()) > sd.schemaView.version
```
liliu-z, line 277: Blocking: current master supports a same `schema.Version` with a newer `SchemaBarrierTs` (for example a `ttl_field` refresh). This version-only gate drops that WAL event before the worker, collection, runner, and IDF updates run. Please keep the barrier in the per-delegator view and freshness check too.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -775,13 +811,10 @@ func (sd *shardDelegator) withPostLoadLimit(ctx context.Context, fn func() error
 	return fn()
 }
 
-func (sd *shardDelegator) addDistributionIfSchemaBarrierOK(schemaBarrierTs uint64, entries ...SegmentEntry) error {
-	sd.schemaChangeMutex.RLock()
-	defer sd.schemaChangeMutex.RUnlock()
-	if schemaBarrierTs < sd.schemaBarrierTs {
-		return merr.WrapErrServiceInternal("schema barrier changed")
+func (sd *shardDelegator) addDistributionIfLoadSchemaOK(ctx context.Context, req *querypb.LoadSegmentsRequest, entries ...SegmentEntry) error {
+	if err := sd.checkLoadSchemaVersionReady(ctx, req); err != nil {
```
liliu-z: Blocking: the final schema check and distribution publish are no longer atomic. This helper releases `schemaChangeMutex.RLock` before `AddDistributions`, so `UpdateSchema` can run in between, miss the not-yet-published segment in `PinOnlineSegments`, publish the new schema, and then let the stale segment become visible. Please hold the same RLock across both the check and `AddDistributions`.

internal/querycoordv2/task/executor.go
```diff
@@ -222,13 +222,23 @@ func (ex *Executor) executeSegmentAction(task *SegmentTask, step int) {
...
+			action.rpcReturned.Store(true)
+		}
+	}()
 	ctx := task.Context()
 
 	var err error
 	defer func() {
 		if err != nil {
-			task.Fail(err)
+			if errors.Is(err, merr.ErrCollectionSchemaVersionNotReady) {
```
liliu-z: Blocking dependency: this retries `LoadSegments`, but today a failed WAL `UpdateSchema` is only logged and TSafe still advances. The delegator view can then stay behind permanently, and retrying the load cannot make progress. Please land/stack the schema apply retry/replay change (and do not advance TSafe on failure) before enabling this wait path.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -110,6 +111,35 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
+			return true
+		}
+		mlog.RatedWarn(ctx, rate.Limit(1), "failed to update schema in delete node, retrying before advancing tsafe",
+			mlog.Int64("collectionID", dNode.collectionID),
+			mlog.String("channel", dNode.channel),
+			mlog.Int32("schemaVersion", nodeMsg.schema.GetVersion()),
+			mlog.Uint64("schemaBarrierTs", nodeMsg.schemaBarrierTs),
+			mlog.Err(err))
+
+		if dNode.closed.Load() {
```
liliu-z, line 182: Blocking: this shutdown check is not reachable through the real pipeline close path. `streamPipeline.Close()` waits for the work goroutine before `pipeline.Close()` calls `deleteNode.Close()`. If `UpdateSchema` keeps failing, `Operate()` waits for `closed` while `Close()` waits for `Operate()`, so shutdown deadlocks. Please wire this retry to a cancellation signal that is closed before `closeWg.Wait()` and add a real pipeline-close test.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -110,6 +111,35 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
 	return nil
 }
 
+func (dNode *deleteNode) updateSchemaUntilApplied(ctx context.Context, nodeMsg *deleteNodeMsg) bool {
+	for {
+		err := dNode.delegator.UpdateSchema(ctx, nodeMsg.schema, nodeMsg.schemaBarrierTs)
```
liliu-z: Blocking: this retries every `UpdateSchema` error forever. Some errors are permanent, for example an incompatible BM25 schema or a released/missing collection, so one event can wedge the vchannel and TSafe forever. Please retry only transient errors and use an explicit fail-stop/restart/replay path for permanent ones.

internal/querynodev2/delegator/delegator.go
liliu-z, line 277: Blocking for rolling upgrade: an old RootCoord can still emit the same `schema.Version` with a newer barrier for a TTL-field update. A new QueryNode drops that WAL DDL here and still advances TSafe, leaving the local/segcore schema stale. The RootCoord version bump only fixes homogeneous new clusters. Please keep the barrier in the delegator view/gate, or explicitly enforce an upgrade order that makes old RC -> new QN impossible.

congqixia: old rc with new qn shall never be possible in prod

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -95,21 +105,70 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
+	dNode.delegator.UpdateTSafe(nodeMsg.timeRange.timestampMax)
+	return nil
+}
+
+func (dNode *deleteNode) updateSchemaUntilApplied(ctx context.Context, nodeMsg *deleteNodeMsg) bool {
+	for {
+		if dNode.closed.Load() {
+			return false
+		}
+		err := dNode.delegator.UpdateSchema(ctx, nodeMsg.schema, nodeMsg.schemaBarrierTs)
```
liliu-z: Blocking: the pipeline is started before delegator.Start(), so this can run while the delegator is still Initializing. UpdateSchema uses IsWorking, returns non-retryable ErrChannelNotAvailable, and the code below panics. Please allow NotStopped here (as in #51785) or change the startup order, and add a regression test for this window.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -95,21 +105,70 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
+		if err == nil {
+			return true
+		}
+
+		if dNode.closed.Load() {
+			return false
+		}
+		if !merr.IsRetryableErr(err) {
```
liliu-z: Blocking: this classification does not match the real UpdateSchema failures. A missing/unhealthy worker becomes ErrServiceInternal or ErrNodeNotAvailable, and raw gRPC Unavailable/deadline errors are not merr errors, so normal transient churn is treated as permanent and panics the QueryNode. Please classify transport/lifecycle/distribution failures explicitly and retry them with a bound before fail-stop.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -95,21 +105,70 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
 
 	if nodeMsg.schema != nil {
 		ctx := context.TODO()
```
liliu-z: Blocking: PreClose can stop the retry timer, but this context cannot cancel an UpdateSchema RPC already in flight. A stuck worker call can still keep streamPipeline.Close waiting. Please use a cancellable node/pipeline context tied to PreClose.

internal/querycoordv2/task/task_test.go
```diff
@@ -639,6 +640,86 @@ func (suite *TaskSuite) TestLoadSegmentTaskNotIndex() {
...
+	suite.target.UpdateCollectionNextTarget(ctx, suite.collection)
+	suite.AssertTaskNum(0, 1, 0, 1)
+
+	suite.dispatchAndWait(targetNode)
+	suite.AssertTaskNum(1, 0, 0, 1)
+	suite.Equal(TaskStatusStarted, task.Status())
+	suite.NoError(task.Err())
+	suite.Contains(task.GetReason(), "collection schema version not ready")
+	action := task.Actions()[0].(*SegmentAction)
+	suite.False(action.rpcReturned.Load())
```
liliu-z, line 722: Blocking test gap: this only proves the task stays Started after the first NotReady response. Please make a second dispatch return success with a refreshed schema and assert the task finishes, so the actual re-pack/reschedule convergence path is covered.

internal/querynodev2/delegator/delegator.go
liliu-z: I still consider this blocking. The standard rollingUpdate.sh is RootCoord-first, but rollback and custom deployments are not protected by any runtime/version gate. Please either keep the same-version/newer-barrier compatibility for one release, or add an explicit capability gate plus a mixed-version test.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -95,21 +105,70 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
-		if err := dNode.delegator.UpdateSchema(ctx, nodeMsg.schema, nodeMsg.schemaBarrierTs); err != nil {
-			mlog.Warn(ctx, "failed to update schema in delete node",
+		if !dNode.updateSchemaUntilApplied(ctx, nodeMsg) {
+			return nil
+		}
+	}
+
+	// update tSafe
+	dNode.delegator.UpdateTSafe(nodeMsg.timeRange.timestampMax)
```
liliu-z, line 122: Blocking: this can still advance TSafe after a schema message was lost before reaching this node. filterNode ignores the error from out.append(msg), so an AddCollectionField Body() decode failure leaves nodeMsg.schema nil and is silently skipped. Please propagate/fail-stop that error so the no-TSafe-advance and WAL-replay invariant is end-to-end.

internal/rootcoord/ddl_callbacks_alter_collection_properties.go
```diff
@@ -178,16 +185,18 @@ func (c *Core) broadcastAlterCollectionForAlterCollection(ctx context.Context, r
 			header.UpdateMask.Paths = append(header.UpdateMask.Paths, message.FieldMaskCollectionSchema)
 		}
 
-		// Build schema snapshot with updated properties (schema version should NOT be changed for properties-only alter).
+		// Build schema snapshot with updated properties. Any schema payload
+		// change must advance schema.Version so WAL consumers can gate on it.
 		schema := coll.ToCollectionSchemaPB()
+		schema.Version = coll.SchemaVersion + 1
```
liliu-z, line 191: Blocking for the version-only gate: this bump is now a correctness requirement, but the PR only tests the producer-side metadata value. Please add a TTL/external consumer test through WAL to the delegator, and centralize or assert the invariant that every non-nil schema payload advances Version; otherwise a missed property path is silently skipped.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1520,6 +1578,7 @@ func NewShardDelegator(ctx context.Context, collectionID UniqueID, replicaID Uni
 		vchannelName:      channel,
 		version:           version,
 		collection:        collection,
+		schemaView:        newDelegatorSchemaView(collection.Schema()),
```
liliu-z: The constructor reads collection.Schema() separately for the function-runner Alloc (:1548) and for the new schemaView snapshot (:1581). collection.Schema() is an atomic snapshot that a concurrent collectionManager.UpdateSchema (from another vchannel of the same collection) can bump between the two reads, leaving V1 runners pinned while schemaView records V2; UpdateSchema's no-op gate then drops the WAL event that would refresh the runners (skips unless schema.Version > schemaView.version). Because this PR also removes the load-path function.GetManager().Update resync, searches on the newly added function field return a persistent ErrServiceUnavailable until the next DDL or a reload. The window widens for function-bearing collections whose Alloc is slow.

internal/rootcoord/ddl_callbacks_alter_collection_properties.go
congqixia: Thanks for calling this out. For this PR we are assuming old RootCoord + new QueryNode is not a supported production deployment state. The
  supported rollout order keeps RootCoord upgraded before QueryNode, so QueryNode should only consume schema payloads produced under the new
  invariant: every non-nil schema payload change must bump schema.Version and be written to WAL.

  Because of that constraint, I do not want to keep same-version/newer-SchemaBarrierTs as a second schema freshness path in delegator. That
  would reintroduce barrier as schema-version semantics and weaken the version-only load gate.

liliu-z:
The current Code-Check failure is just this gocritic issue:

```text
internal/querynodev2/pipeline/delete_node_test.go:185:19: unlambda: replace func() bool { return updateSchemaCalled.Load() } with updateSchemaCalled.Load
```

So this can be:

```go
suite.Eventually(updateSchemaCalled.Load, time.Second, time.Millisecond)
```

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -110,16 +121,126 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
 	return nil
 }
 
+func (dNode *deleteNode) updateSchemaUntilApplied(ctx context.Context, nodeMsg *deleteNodeMsg) bool {
+	start := time.Now()
+	for {
+		if dNode.closed.Load() || ctx.Err() != nil {
+			return false
+		}
+		err := dNode.delegator.UpdateSchema(ctx, nodeMsg.schema, nodeMsg.schemaBarrierTs)
```
liliu-z: Blocking: the retry limit still does not bound a single UpdateSchema call. The max duration is checked only after this call returns, while dNode.ctx has no deadline. A stuck local or remote worker call can therefore block WAL and TSafe forever. Please use a context.WithTimeout for the whole apply loop so the 5-minute bound also cancels an in-flight attempt.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -692,15 +779,19 @@ func (sd *shardDelegator) LoadSegments(ctx context.Context, req *querypb.LoadSeg
 	}
 	log.Debug(ctx, "work loads segments done")
 
-	if err := sd.syncCollectionIndexMeta(ctx, req); err != nil {
-		log.Warn(ctx, "failed to sync collection index meta on delegator", mlog.Err(err))
+	if err := sd.checkLoadSchemaVersionReady(ctx, req); err != nil {
```
liliu-z, line 741: Blocking: if a V1 full load finishes on the worker and this check observes V2, we return and retry with V2, but the worker keeps the V1 segment. The retry only updates CCollection and segmentLoader.prepare skips the already-loaded segment, so it is published without reopening or reloading the V2 load info. Please repair or replace an existing segment from an older load epoch, and cover this with a real worker-segment test.

internal/querynodev2/delegator/delegator.go
```diff
@@ -1487,11 +1553,12 @@ func NewShardDelegator(ctx context.Context, collectionID UniqueID, replicaID Uni
 	if collection == nil {
 		return nil, merr.WrapErrCollectionNotFound(collectionID, "not in delegator manager")
 	}
-	if err := function.GetManager().Alloc(collectionID, delegatorFunctionRunnerKey(channel), collection.Schema()); err != nil {
+	initialSchema := collection.Schema()
```
liliu-z: Blocking: one shared read fixes the split snapshot, but this is still not the schema of this Watch. A V1 Watch can start before its V2 WAL barrier while another vchannel has already advanced the shared Collection to V2; this delegator and the pipeline then start with V2, process pre-barrier inserts with future functions, and skip their own V2 event. Pass one clone of the Watch request schema into all per-vchannel runtime state.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -94,32 +112,136 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
+			mlog.Duration("retryDuration", time.Since(start)),
+			mlog.Err(wrapped))
+		panic(wrapped)
+	}
+
+	for {
+		if dNode.closed.Load() {
+			return false
+		}
+		err := dNode.delegator.UpdateSchema(applyCtx, nodeMsg.schema, nodeMsg.schemaBarrierTs)
```
liliu-z, line 147: Blocking: `applyCtx` still does not bound this call. `UpdateSchema` reaches `GetWorker`, but the QueryNode worker builder calls `GetSessions(node.ctx)` instead of using the passed context, so a stuck etcd lookup ignores the five-minute deadline and blocks WAL/TSafe forever. Propagate this context through worker discovery and client creation.

internal/querynodev2/pipeline/delete_node.go
```diff
@@ -94,32 +112,136 @@ func (dNode *deleteNode) Operate(in Msg) Msg {
...
+	}
+	if merr.IsRetryableErr(err) {
+		return true
+	}
+	if errors.Is(err, context.DeadlineExceeded) ||
+		errors.Is(err, merr.ErrChannelNotAvailable) ||
+		errors.Is(err, merr.ErrNodeNotAvailable) {
+		return true
+	}
+	switch status.Code(err) {
```
liliu-z, line 215: Blocking: normal worker churn still falls through as permanent here. Discovery can return a raw etcd error; a disappeared session becomes `ErrParameterInvalid` from an empty address; closed clients can return `grpc.ErrClientConnClosing`. None matches this list, so a transient topology or connection failure panics the QueryNode. Classify the real lookup/RPC errors as retryable.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -595,24 +625,27 @@ func (sd *shardDelegator) syncCollectionIndexMeta(ctx context.Context, req *quer
...
 		loadMeta = &querypb.LoadMetaInfo{
 			CollectionID: req.GetCollectionID(),
 		}
+	} else {
+		loadMeta = typeutil.Clone(loadMeta)
 	}
+	loadMeta.SchemaBarrierTs = 0
 
 	meta := segments.ComposeIndexMeta(ctx, req.GetIndexInfoList(), schema)
 	if err := sd.collectionManager.PutOrRef(req.GetCollectionID(), schema, meta, loadMeta); err != nil {
```
liliu-z, line 653: Blocking: this can roll the shared IndexMeta back. Shard B can process a V2 load after shard A moved the shared Collection to V3; `PutOrRef` rejects the schema rollback, but `applyLoadUpdate` still replaces the whole IndexMeta with B's V2 list. That leaves schema V3 plus IndexMeta V2. Fence or version the IndexMeta update too, or make it segment-local.

internal/querycoordv2/task/executor.go
liliu-z: Blocking: only schema mismatches stay pending. A Reopen may already succeed on the worker and report the new DataVersion, then fail in delegator BM25 post-load repair on a transient object-store read. Failing the task here loses the retry because SegmentChecker now sees the target version, leaving the IDF oracle incomplete. Keep runtime-repair failures pending and retryable too.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -1124,7 +1202,7 @@ func (sd *shardDelegator) loadStreamDelete(ctx context.Context,
 			// Atomically add to distribution while still holding RLock, so no
 			// ProcessDelete can run between "deletes applied" and "segment
 			// visible".
-			if err := sd.addDistributionIfSchemaBarrierOK(schemaBarrierTs, entries...); err != nil {
+			if err := sd.addDistributionIfLoadSchemaOK(ctx, req, entries...); err != nil {
```
liliu-z, line 1176: Blocking: this late gate can fail after `loadBM25Stats` already installed V1 stats. That entry is not rolled back, and the V2 full-load retry calls `LoadSealed`, which returns early for an existing segment, so a BM25 field added in V2 never gets stats. Roll back or epoch-key the side effect, or use the field-aware reopen repair on retry.

internal/querynodev2/delegator/delegator_data.go
```diff
@@ -692,6 +733,11 @@ func (sd *shardDelegator) LoadSegments(ctx context.Context, req *querypb.LoadSeg
 	}
 	log.Debug(ctx, "work loads segments done")
 
+	if err := sd.checkLoadSchemaVersionReady(ctx, req); err != nil {
+		sd.releaseOrphanLoadedSegments(ctx, worker, targetNodeID, req, orphanLoadInfos)
```
liliu-z: Blocking under overlapping deliveries such as QueryCoord or leader failover: the singleflight key and this cleanup have no load-epoch ownership. An older request can share or outlive a newer one, then release the segment by ID after the newer load succeeds. Normal single-QueryCoord retries are serialized, but failover/duplicate RPCs still need an epoch-owned result and cleanup.

internal/querynodev2/segments/segment_loader.go
```diff
@@ -455,6 +458,56 @@ func (loader *segmentLoader) prepare(ctx context.Context, segmentType SegmentTyp
...
+type loadSchemaVersionedSegment interface {
+	LoadSchemaVersion() uint64
+}
+
+func (loader *segmentLoader) reopenLoadedSegmentsIfStale(ctx context.Context, collection *Collection, segmentType SegmentType, infos ...*querypb.SegmentLoadInfo) error {
+	if segmentType != SegmentTypeSealed {
+		return nil
+	}
+
+	requiredSchemaVersion := collection.SchemaVersion()
```
liliu-z: Blocking: the repair epoch comes from the mutable shared Collection, not the load request. A V2 request can run after another shard moved Collection to V3, so Reopen applies V2 `SegmentLoadInfo` under V3; `LocalSegment.Reopen` then reads the version again after C++ returns and can mark a V2 reopen as V3. Thread one exact request schema/version through the whole reopen and store that captured version.

internal/querynodev2/segments/segment_loader.go
```diff
@@ -455,6 +458,56 @@ func (loader *segmentLoader) prepare(ctx context.Context, segmentType SegmentTyp
...
+
+func loadedSegmentNeedsReopen(segment Segment, incoming *querypb.SegmentLoadInfo, requiredSchemaVersion uint64) bool {
+	current := segment.LoadInfo()
+	if incoming.GetDataVersion() > current.GetDataVersion() {
+		return true
+	}
+	if incoming.GetStorageVersion() > current.GetStorageVersion() {
+		return true
+	}
+	if incoming.GetManifestPath() != "" && incoming.GetManifestPath() != current.GetManifestPath() {
```
liliu-z, line 515: Blocking: path inequality is not a freshness check. A delayed request carrying an older manifest path will reopen and roll a newer loaded segment back. Compare manifest revisions, for example with `packed.CompareManifestPath`, and reject a lower composite load epoch instead of reopening it.

internal/querynodev2/delegator/delegator.go
```diff
@@ -938,6 +1002,7 @@ func organizeSubTask[T any](ctx context.Context,
 			req:      req,
 			targetID: workerID,
 			worker:   worker,
+			err:      err,
```
liliu-z, line 1018: Blocking: UpdateSchema uses `skipEmpty=false`, so an offline-filtered empty sealed group still reaches this with worker ID `-1`; `GetWorker(-1)` returns an error that is now propagated into schema fail-stop even though there are zero segments to update. Skip empty sealed groups while still issuing the local worker update explicitly.

internal/querynodev2/pipeline/filter_node.go
```diff
@@ -95,7 +95,15 @@ func (fNode *filterNode) Operate(in Msg) Msg {
 				mlog.Err(err),
 			)
 		} else {
-			out.append(msg)
+			if err := out.append(msg); err != nil {
```
liliu-z, line 98: Follow-up after tracing all current producers: valid AlterCollection rejects a schema mask with a nil schema before WAL append, so I no longer consider this a merge blocker. This validation is still useful hardening for malformed/replayed WAL.

internal/querynodev2/segments/segment_loader.go
```diff
@@ -241,6 +241,9 @@ func (loader *segmentLoader) Load(ctx context.Context,
 	for _, segment := range segments {
 		configureUseTakeForOutput(segment, collection.Schema())
 	}
+	if err := loader.reopenLoadedSegmentsIfStale(ctx, collection, segmentType, segments...); err != nil {
```
liliu-z: Blocking: this repair runs before `prepare`, so a V2 request arriving while V1 is still loading sees no segment here, gets skipped as `isLoading`, and `waitSegmentLoadDone` accepts the V1 result without rechecking the load epoch. Recheck after the wait, or make the loading state epoch-aware.

Issue #40451, opened 2025-03-07
[Enhancement]: New Search Architecture Based On Streaming Service


chyezh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

New Search Architecture Based On Streaming Service 

### Why is this needed?

Current Streaming Service supports a embedded querynode to implement search/query, also see #38399. 
But old delegator logic is too heavy for streaming node, and we cannot split the batch and streaming process based on current delegator arch completely:

- Cannot merge the flush process and search search built process, so there're always double consuming from wal when recovery, double memory usage if collection is loaded.
- Cannot put all meta management of growing data on streaming node and make a light weight history meta coordinator.
- Cannot remove the forwarding RPC of the delegator to reduce streamingnode's work. 

Here's the new distributed architecture for search/query process based on streaming service:

![Image](https://github.com/user-attachments/assets/e7c816e5-4559-4d87-8d0b-be4206ac145d)

The query process is implemented as shown in the diagram above, following a two-phase query approach:  

- Coordinator will generate global versioned query view and sync the view to all streaming node and query node, and keep consistency by  a cross-node state machine.
- QueryNode will subscribe the pure delete stream from the streaming node and apply the delete request to all segment on it.
- **Phase 1:** The Proxy generates a shard-level query plan using the highest version of the QueryView from the StreamingNode:  
  - Includes MVCC.  
  - Query optimization (BM25, segment filtering, etc.).  
  - Query view versioning.  

- **Phase 2:** The Proxy sends the query plan to both the StreamingNode and QueryNode:  
  - The StreamingNode and QueryNode execute all query operations on their respective segments based on the specified view version (similar to the current SearchSegments process, but using version numbers instead of a segment list).  

- **Final Step:** The Proxy reduces all results and returns them to the user.  

During this process, if a node crashes or the view becomes invalid, the process is canceled and the query operation is retried.

Here's the TODO list:

- [ ] Versioned Data View to keep a view of historical data on a shard.
- [ ] Versioned Query View to keep a distributed loaded info of a loaded shard.
    - #40467
    - #40518
    - #40521
- [ ] New balancer to generate query view automatically by current cluster info.
- [ ] Cross node state machine to keep consistency query view between streaming node, query node and coord.
- [ ] Pure delete stream subscription start from any checkpoint.
- [ ] Segment loader scheduler on query node to act with query view.
- [ ] Client of new search architecture implementation for proxy.
[...]

PR #51887, opened 2026-07-27
feat: add query view state machines

chyezh:
issue: #40451

- QueryView protocol: add lifecycle states, per-node view payloads, bidirectional sync messages, and generated Go bindings
- state machines: add shared QueryView value objects and focused Coord, StreamingNode, and QueryNode lifecycle implementations with unit tests
- documentation: describe state transitions, recovery of versioned Up views, StreamingNode query-lease ownership, and TODO integration boundaries


czs007:

Reviewed at head 4f2c846 (all 3 commits, full diff vs master). Overall: mergeable as scaffolding — the three state machines are internally consistent, the pending-persist/pending-sync consume model is clean, and the unit tests are thorough (lifecycle, crash-recovery regression paths, idempotency, re-push fast-forward). I traced the cross-node chains (Coord↔SN↔QN, including the Dropping-not-persisted recovery regression and the UpRecovering silent-unrecoverable path) and found no blocking correctness issue. Findings below are non-blocking: one contract-hardening item and several doc/robustness nits.

**Medium — `OnQueryNodeLost` is silently dropped in Ready/Up/Down/Unrecoverable, which can deadlock Dropping unless the future Manager re-delivers the loss.**
`internal/views/coord/coordview/state_machine.go:159-174` only handles Preparing (→ Unrecoverable) and Dropping (→ count as Dropped). Failure scenario: a QN dies while the view is Up; service discovery delivers `OnQueryNodeLost` once → no-op; later the view goes Down → Dropping. `allNodesDropped()` (line 384) requires `qnStates[deadQN] == Dropped`, which can now only be set by a Dropped report (never comes) or another `OnQueryNodeLost` call while in Dropping. If the Manager's discovery integration is edge-triggered (fires once at removal), the view is stuck in Dropping forever — never deleted from ETCD, SM instance leaked, and `handleDropping` keeps re-pushing Dropped to a dead node. The doc (§1.5/§1.6) implies level-triggered, sync-scoped delivery ("loss is delivered only for active QN-targeted syncs"), but nothing enforces it. Suggested fix: make the SM self-sufficient — record the loss in every state (e.g., a `lostQNs` set, or set `qnStates[id] = Dropped` immediately in non-Preparing states) and consult it in `allNodesDropped()`; alternatively, state the level-triggered re-delivery requirement explicitly in the `OnQueryNodeLost` godoc so the Manager PR can't miss it.

**Low — doc/code mismatches in `query_view_state_machine.md` (this PR's third commit is specifically a doc-alignment commit).**
1. §1.3 (line 82) lists an Up-state transition trigger "a QueryNode is lost while a Preparing sync is pending" — the code has no such transition (`OnQueryNodeLost` is a no-op in Up, and a view in Up cannot have a pending Preparing sync since Up requires all QNs Ready). This row should be removed.
2. §1.4 (peer states, lines 119-124) and §1.5 entry conditions (line 129, "while Coord is in Preparing, Ready, or Up") omit the Down → Unrecoverable transition that the code implements at `coordview/state_machine.go:280-283` and explicitly tests (`TestDown_UnrecoverableFromSN/QN`). This path is reachable: SN goes Unrecoverable silently during UpRecovering, Coord pushes Down, SN re-reports Unrecoverable. The doc should include it.
3. §1.5 entry conditions also omit the Manager-forced `EnterUnrecoverable` (preemption / RequestRelease), which the code exposes at `coordview/state_machine.go:147`.

**Low — QN can report never-assigned segments as ready.**
`internal/querynodev2/qnview/state_machine.go:125-133`: for a known partition, delta segment IDs not present in `assignedSegments` are still inserted into `readySegments`, and `buildReport` (lines 253-256) publishes them in `ready_segment_ids`. Failure scenario: a stale load-completion callback for a segment that belongs to a different/older view inflates the reported ready set, and Coord's `QNReadySegments()` — documented as Balancer/Manager decision input — then over-reports progress. Suggested fix: skip IDs not in `assignedSet` entirely (they already don't count toward `readyCount`, so recording them only pollutes the report).

**Low — panic-on-malformed-proto at a wire trust boundary, plus a nil-SN-view classification hazard.**
`internal/views/qviews/view.go:57-68` panics on a `QueryViewOfShard` that has neither `StreamingNode` nor exactly one `QueryNode`. This factory is the designated converter for follower responses arriving over `SyncQueryView`, so a malformed or truncated response from one node would panic the coordinator process. Consider a `(QueryViewAtWorkNode, error)` signature for the wire-facing path. Relatedly, classification keys on `StreamingNode != nil`, while `snview`'s `cloneStreamingNodeView` (`snview/state_machine.go:341-346`) deliberately tolerates a nil SN view — an SN state machine constructed with a nil `snView` would emit reports that get classified as QN reports (or panic) on the Coord side. Normalizing a nil SN view to an empty message at `newSNQueryViewStateMachine`/`buildReport` would close this off cheaply.

**Nit — dead code.** `StateTransition`/`NewStateTransition`/`IsStateTransition` (`internal/views/qviews/utils.go:68-93`) are referenced only by their own test. If they are staging for the Manager's transition-notification API, a TODO comment would help; otherwise drop them.

**Nit — inconsistent clone discipline across the three SMs.** SN SM defensively clones on construction and on every accessor; QN SM stores `meta`/`qnView` without cloning and `Meta()`/`QNView()` (`qnview/state_machine.go:77-84`) return internal pointers; Coord `View()` and `QNReadySegments()` (`coordview/state_machine.go:110-123`) also expose internals (the tests even pin `View()` returning the same reference). Any future caller mutating a returned proto corrupts SM state silently. Suggest aligning on the SN convention, or at minimum documenting "returned protos must not be mutated" on the accessors.

No blocking issues found; the medium item is a contract to pin down before/within the Manager integration PR rather than a bug in this diff.

chyezh:
Follow-up to the automated review at https://github.com/milvus-io/milvus/pull/51887#issuecomment-5102134978.

Addressed the three documentation mismatches in `70583c5dd9`:

- removed the impossible QueryNode-loss-during-Preparing trigger from the Coord Up transition
- documented the implemented Down -> Unrecoverable transition and peer-state handling
- documented Manager-driven `EnterUnrecoverable` for preemption/RequestRelease and updated the Dropping entry sources

The remaining findings were checked against the complete qv integration and do not require changes in this PR:

- `OnQueryNodeLost`: ReliableSyncer rechecks `IsNodeAlive` when the Dropped sync is enqueued. If the QN is already absent, it immediately invokes `OnQueryNodeLost` for that Dropping target; if a pending per-node syncer is removed, it drains the pending entries and invokes the same callback. The state machine therefore receives the loss when the Dropping cleanup quorum needs it, so recording QN loss while Up is unnecessary.
- unassigned ready segments: `QueryViewSegmentReadinessManager` creates readiness waiters only from the segments assigned in that QueryView, keyed by QueryViewKey/version. Its callback cannot report a segment from another view through the production path.
- malformed proto / nil SN view: `SyncQueryView` is an internal trusted protocol and its reports are constructed by the node state machines from Coord-built views. The one-node payload and non-nil SN view are construction preconditions; expanding the wire-facing API for impossible internal inputs would add defensive complexity without a production behavior requirement.
- clone discipline: QueryView is immutable by lifecycle and ownership semantics. No production caller mutates the state-machine snapshot through these accessors.
- `StateTransition`: retained as part of the extracted qv shared state-machine utility surface. It has no runtime effect; any cleanup can be handled separately from this behavior-preserving extraction.

No state-machine behavior was changed by this follow-up.

czs007:

🤖 **team2 automated re-review** (single agent · not cross-verified, for reference) · by `zack`

Re-reviewed at head 70583c5 (delta vs previously reviewed 4f2c846 is exactly the one-file doc commit; no state-machine behavior changed, matching the follow-up's claim). **Verdict: mergeable.** All three documentation fixes from the follow-up are verified correct against the code, and the follow-up's rationale for the remaining findings is not contradicted by anything in this branch.

Verification of the doc fixes in `docs/design-docs/design_docs/qviews/query_view_state_machine.md`:

1. **Up trigger** — the impossible "QueryNode lost while a Preparing sync is pending" row is removed; §1.3 now matches `handleUp` (`internal/views/coord/coordview/state_machine.go:268-272`) and the `OnQueryNodeLost` no-op in Up (`state_machine.go:159-174`). ✓
2. **Down → Unrecoverable** — §1.4 now documents the transition and both SN/QN Unrecoverable peer reactions, matching `handleDown` (`state_machine.go:281-283`) and the existing `TestDown_UnrecoverableFromSN` / `TestDown_UnrecoverableFromQN` tests; §1.5 entry conditions and the §1.6 Dropping provenance ("Preparing, Ready, Up, or Down") are updated consistently. ✓
3. **Manager-driven `EnterUnrecoverable`** — §1.5 now documents it with exactly the state guard the code enforces (Preparing/Ready/Up, `state_machine.go:147-152`). ✓

On the rejected findings: the `OnQueryNodeLost` re-delivery explanation (ReliableSyncer rechecking `IsNodeAlive` at Dropped-sync enqueue and on pending-syncer removal) and the `QueryViewSegmentReadinessManager` assigned-segments guarantee reference integration code that is not part of this PR, so they cannot be audited here — but both are internally consistent with the doc's stated delivery contract (§1.5 note: loss delivered only for active QN-targeted syncs), and the Dropping-cleanup scenario they describe covers the deadlock case raised previously. I am not re-raising these; they should be re-checked when the Manager/ReliableSyncer PR lands, since that is where the contract becomes enforceable.

One residual finding, same class as the mismatches this commit fixed:

**Nit — §1.4 Down peer-state list omits SN reporting Dropped.**
`query_view_state_machine.md:115-124` lists SN peer states in Down as "Up / Down / Unrecoverable" and the transition trigger as "SN confirms Down" only. The code explicitly accepts SN = Dropped as equivalent to Down (`state_machine.go:288`, with the rationale comment at lines 276-278: Coord crash recovery regresses Dropping → Down while the SN already advanced to Dropped), and this path is pinned by `TestDown_SNDropped_FastForwardToDropping`. Suggest adding "Dropped → Coord transitions to Dropping (SN already advanced past Down; occurs when Coord crash recovery regresses from Dropping to Down)" to the peer-state list and "SN confirms Down (or reports Dropped)" to the trigger. Doc-only; fine to fold into any later commit.

No blocking issues. The scaffolding remains internally consistent and well-tested; head 70583c5 is mergeable.

czs007:

🤖 **team2 automated re-review** (single agent · not cross-verified, for reference) · by `zack`

Re-reviewed at head 3031646. The delta vs the previously reviewed 70583c5 is exactly one doc commit (`3031646e56`, 5 insertions in `query_view_state_machine.md` §1.4) — no state-machine code changed. **Verdict: mergeable.**

The delta resolves the single residual nit from the previous re-review, and I verified it end-to-end against the code:

1. **Coord side** — `handleDown` (`internal/views/coord/coordview/state_machine.go:280-296`) treats an SN report of Dropped identically to Down (`report.State() == QueryViewStateDown || report.State() == QueryViewStateDropped` → Dropping), with the rationale comment at lines 276-279 matching the doc's wording. The persistence claim in the new text is also accurate: `EnterDown` sets `pendingPersist` (Down is persisted, line 184) while the Down → Dropping transition sets only `pendingSync` (Dropping is not persisted, line 289), so "regresses from the unpersisted Dropping state to persisted Down" is exactly what the code does. ✓
2. **Test** — `TestDown_SNDropped_FastForwardToDropping` (`state_machine_test.go:950-969`) pins the path: drive to Down, SN reports Dropped, assert Dropping + no pending persist + pending Dropped sync. ✓
3. **SN side of the chain** — the scenario is actually producible by the SN state machine in this PR: `handleCoordDown` (`snview/state_machine.go:256-273`) re-reports the current state (including Dropped) when the regressed Coord re-pushes Down, and if the SN is still mid-Dropping when the push arrives, the later `OnDropped` callback (`snview/state_machine.go:186-192`) emits the Dropped report that unblocks Coord. So the documented recovery fast-forward is closed at both ends. ✓

Full-PR rescan for remaining code-vs-doc mismatches: I re-walked all seven Coord states in §1 against `coordview/state_machine.go`, §2 against `snview/state_machine.go`, and §3 against `qnview/state_machine.go`, and found no remaining discrepancy. One case I deliberately checked and am *not* raising: a QN reporting Dropped while Coord is in Down (the QN-side analogue of the scenario this commit documents) is handled benignly by the code (`updateNodeState` at line 347 records it; the `isSNReport` gate at lines 285-287 makes it a no-op for transitions), but it has no production delivery path — `syncViewsForState` (line 325) excludes QNs from Down pushes, so no QN-targeted sync is active in Down. The doc omitting it is consistent with its convention of listing reachable peer states; nothing for the author to change.

Previously rejected findings stand as resolved per the author's follow-up; no new evidence contradicts that rationale. No substantive issues — head 3031646 is mergeable.

internal/querynodev2/qnview/state_machine.go
```diff
@@ -0,0 +1,262 @@
...
+		if changed {
+			sm.pendingReport = sm.buildReport()
+		}
+		return
+	}
+	if sm.readyCount >= sm.totalSegments {
+		sm.state = qviews.QueryViewStateReady
+		sm.pendingReport = sm.buildReport()
+	} else {
+		sm.pendingReport = sm.buildReport()
```
liliu-z: buildReport clones the entire qnView proto (state_machine.go:252) and is invoked eagerly on every OnSegmentsReady event (lines 138/144/146). Because pendingReport is overwritten and only the latest survives to ConsumeReport, any events between consumes do wasted full-view clones, and with the progressive per-segment ready reporting described in the design doc this clone cost scales with the segment count. Consider constructing the report lazily in ConsumeReport (mark dirty on change) rather than on every event.

czs007:
Adversarial review of the query-view state machines surfaced 10 verified issues — 6 high (state-machine correctness and crash-consistency ordering gaps), 1 medium, and 3 low.

### High

- **`internal/views/coord/coordview/state_machine.go:314`** — `transitionToUnrecoverable` sets the state and `pendingPersist` but never clears `pendingSync`, unlike every other intent-changing transition (`EnterDown` L185, `EnterDropping` L197, `handleDown` L294, `handleDropping` L309). With 6 call sites (L150, L166, L227, L250, L270, L282), any path can leak a stale sync: e.g. `handlePreparing` queues `pendingSync = Up` on all-Ready, then a QN `Unrecoverable` report drives `handleReady` → `transitionToUnrecoverable`, which persists Unrecoverable but leaves the Up sync alive — the Manager would push Up for a view the Coord has already declared dead, causing the SN to activate and persist a view no QN backs. Suggestion: set `pendingSync = nil` inside `transitionToUnrecoverable`. *(raised by tinswzy, xaxys)*

- **`internal/streamingnode/server/wal/snview/state_machine.go:240`** — `handleCoordUp`'s `default` branch calls `buildReport()` while `coordVisibleState()` maps `UpRecovering` to `Up`, so an SN that crashed after persisting Up (Coord still in Ready, re-pushing Up) reports Up immediately on restart while still catching up on the WAL. The Coord then confirms the new view and Downs the old one — the new view can't serve query plans yet and the old one is deactivated. `state_machine_test.go:785` currently locks in this behavior. Suggestion: in `UpRecovering`, defer the Up report until `OnRecoveringDone`. *(raised by xaxys)*

- **`internal/views/coord/coordview/state_machine.go:49`** — There is no ordering barrier between write-ahead persistence and node sync: the constructor `NewCoordQueryViewStateMachine` (L60-61) and `EnterDown` (L184-185) set both `pendingPersist` and `pendingSync`, which `ConsumePersist`/`ConsumeSync` drain independently. If the Manager pushes Preparing first, nodes reach Ready/Up and the SN persists the view, then the Coord crashes before consuming the ETCD persist, restart leaves ETCD with no record of a view the SN is actively serving — an uncleanable ghost Up view. `state_machine_test.go:831` shows consuming Sync alone is API-legal, contradicting the design doc's "persist Preparing, then push". Since the Manager is not in this PR, the invariant should be enforced at the SM API level (open Sync only after persist is confirmed) or at minimum documented as a hard caller contract. *(raised by xaxys)*

- **`internal/streamingnode/server/wal/snview/state_machine.go:244`** — On Ready→Up, `pendingReport` and `pendingPersist` are produced as two independent `Consume*` outputs with no ordering guarantee. If the report is drained first, the Coord persists Up, and the SN crashes before `ConsumePersist`, the restarted SN has no persisted view — and Coord-side recovery explicitly does no re-push for Up views (`coordview/state_machine.go:90-92`), leaving the Coord with a permanently-Up view the SN can never reconstruct. Suggestion: document and enforce persist-before-report in the (TODO) caller, since the SM cannot enforce it itself. *(raised by tinswzy)*

- **`internal/streamingnode/server/wal/snview/state_machine.go:256`** — The symmetric window on teardown: Down/Dropped paths (L256/L275, and `OnDropped`) generate the report and the local recovery-record deletion as independent pendings, so the report can go out while the delete is still pending. If the Coord completes global Dropped and removes the ETCD record, then the SN crashes before deleting its local record, restart resurrects the view as UpRecovering/Up with no Coord counterpart to ever Drop it — leaking resources, and the stale view can be selected as the highest available Up version for query planning. Suggestion: guarantee local deletion completes before the Down/Dropped report is emitted. *(raised by xaxys)*

- **`pkg/proto/view.proto:53`** — `QueryViewOfStreamingNode` is an empty message and `StreamingNode.Key()` is `sn@<pchannel>`, so `updateNodeState` (`coordview/state_machine.go:349-351`) unconditionally overwrites `snState` with any SN report for that pchannel — a delayed report from a pre-reassignment owner is indistinguishable from the current owner's (e.g. a stale Ready/Up prematurely satisfying `allNodesReady`). The transport layer is TODO and may fence old owners via per-node stream sessions, but that defense doesn't exist yet nor is it stated as a contract. Suggestion: add a term/epoch field now (adding a field to an existing proto3 message is wire-compatible), or explicitly document that report freshness is guaranteed by the transport session layer. *(raised by tinswzy)*

### Medium

- **`internal/querynodev2/qnview/state_machine.go:126`** — `OnSegmentsReady` filters by partition but not by segment: every reported segment ID for a known partition is recorded in `pSet`, with only `readyCount` gated by `assignedSet`. `buildReport` (L255) then echoes the full set into `ReadySegmentIds`, and the Coord's `updateQNReadySegments` copies it verbatim. Cosmetic until the Balancer (TODO) consumes the field, but `ReadySegmentIds` should be constrained to a subset of `SegmentIds` at the source. *(raised by tinswzy, xaxys)*

### Low

- **`internal/views/coord/coordview/state_machine.go:344`** — `updateNodeState` writes into `qnStates` without a membership check, unlike `OnQueryNodeLost` (L160). A misrouted report inserts a phantom QN that blocks `allNodesReady`/`allNodesDropped`, or — if it reports Unrecoverable — sends the whole view to Unrecoverable. Suggestion: add the same membership guard as `OnQueryNodeLost`. *(raised by tinswzy)*

- **`internal/querynodev2/qnview/state_machine.go:55`** — A zero-segment view starts in Preparing with `readyCount == totalSegments == 0`, and the only transition to Ready lives inside `OnSegmentsReady`; if the loader never calls it for an empty assignment, the QN hangs in Preparing forever and the Coord's `allNodesReady` never trips. Suggestion: transition to Ready in the constructor when `total == 0`, or document that callers must invoke `OnSegmentsReady(nil)` for empty assignments (0 >= 0 already triggers the transition). *(raised by tinswzy)*

- **`internal/querynodev2/qnview/state_machine.go:61`** — The QN SM holds caller-provided `meta`/`qnView` pointers directly and `Meta()`/`QNView()` return internal pointers, while the SN SM consistently `proto.Clone`s on both construction and read. External mutation of the proto after construction bypasses the `assignedSegments`/`totalSegments` snapshot, desynchronizing the Ready check from the actual assignment. Suggestion: align with snview — clone on construction and on read accessors. *(surfaced during verification)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3031646e5687a6a9c7ae6faa87af3b17f06b699b...czs007:team2-fix/Td268aa29f5?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3031646e5687a6a9c7ae6faa87af3b17f06b699b...czs007:team2-fix/Td1c56af79a?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3031646e5687a6a9c7ae6faa87af3b17f06b699b...czs007:team2-fix/T7f60c61d1d?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3031646e5687a6a9c7ae6faa87af3b17f06b699b...czs007:team2-fix/T4a6a223de1?expand=1

internal/querynodev2/qnview/state_machine.go
chyezh: Fixed in `952dcd6bea`.

The QN state machine now keeps only a `reportPending` intent during state/progress events and constructs the protobuf lazily in `ConsumeReport()`. Multiple readiness events before a consume therefore coalesce into one report built from the latest state instead of cloning the full QN view for every event. Entering Dropping still clears any stale report intent, preserving the previous cancellation semantics.

This does not introduce a new concurrency model: the state machine remains explicitly non-thread-safe and all calls are serialized by the owner, so report construction still observes a consistent state snapshot. Added `TestSegments_ReportBuiltLazilyAndCoalescesProgress`; the focused test was verified RED before the implementation, and the QN race test plus all QV/proto tests pass.

chyezh:
Follow-up to the adversarial review at https://github.com/milvus-io/milvus/pull/51887#issuecomment-5104052191.

Two responsibility boundaries explain the reported high-level concerns:

1. **UpRecovering receiving a re-pushed Up is not a critical consistency window.** This state machine coordinates the distributed QueryView state across nodes. Once SN has persisted Up, SN owns enough recovery information to reconstruct the view after restart; that is the state-machine durability objective. Local WAL catch-up and query serviceability remain SN-local recovery concerns. Reporting the Coord-visible Up state does not mean the local recovery implementation must already be serving queries.
2. **StreamingNode owner freshness/fencing is a Syncer responsibility, not a per-view state-machine responsibility.** The Syncer owns the node/pchannel-scoped stream session and must reject or retire reports from stale owners. Adding assignment epochs to the QueryView state machine would duplicate transport-session semantics at the wrong layer.

The remaining adversarial findings are false positives when evaluated through the complete production call chains:

- stale `pendingSync` and Coord persist-before-sync ordering: ShardViewManager processes each state-machine event synchronously, drains pending persist before pending sync, and flushes persistence before dispatching sync. Direct tests that intentionally leave pending outputs unconsumed bypass the Manager contract. The generated candidate patch is therefore not applied.
- SN persist/report and teardown ordering: `snShardView.consumeReportPersistAndCleanup` calls `consumeAndPersist` before `consumeReport`, covering both activation and recovery-record deletion paths.
- unassigned ready segments and zero-segment readiness: QueryViewSegmentReadinessManager creates callbacks only from the current QueryView assignment and explicitly invokes `OnReady` for an empty assignment.
- phantom QN reports: ReliableSyncer routes through a target-node stream and matches the complete QueryViewKey before invoking the state machine.
- proto clone discipline: QueryView is immutable by lifecycle and ownership semantics; no production path mutates the state-machine snapshot through these accessors.

Separately, liliu-zs report-construction performance comment was actionable and is fixed in `952dcd6bea` by constructing QN reports lazily in `ConsumeReport`; this changes no state-machine consistency or concurrency semantics.

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/3031646e5687a6a9c7ae6faa87af3b17f06b699b...czs007:team2-fix/T5361a67e0a?expand=1

czs007:

Adversarial review of the query-view scaffolding found one robustness gap in the shared value layer; the three state machines and version-ordering helpers were checked independently and held up.

## Low

- **`internal/views/qviews/view.go:57`** — `NewQueryViewAtWorkNodeFromProto` disambiguates a StreamingNode report from a QueryNode report *solely* by `pb.StreamingNode != nil`, but `QueryViewOfStreamingNode` is an empty message (`pkg/proto/view.proto:55-56`), so nothing forces a construction path to set it, and `proto.Clone` on a typed-nil preserves the nil all the way through (`view.go:79` → `snview/state_machine.go:341-345` `cloneStreamingNodeView(nil)` returns nil → `buildReport` emits a report with `StreamingNode == nil` and a full `QueryNode` topology). Failure: the Balancer-produced `QueryViewOfShard` leaves `StreamingNode` unset, Coord pushes it to the SN, and the SN's report comes back with `StreamingNode == nil` → with one QN replica the SN report is silently reclassified as a QueryNode report, so `coordview/state_machine.go:347` writes the SN's state into `qnStates`, `snState` stays `Nil`, `allNodesReady` (`:372-382`) never holds and the view is stuck in `Preparing` forever; with more than one QN replica it hits `panic("invalid node view proto")` at `view.go:67` and takes down Coord's sync path. This was reproduced against the branch: `NewFullQueryViewAtStreamingNode(meta, nil, [{NodeId:7}]).IntoProto()` round-trips back as `*qviews.QueryViewAtQueryNode / qn@7`, and the two-QN case recovers `invalid node view proto`. Note that `view_test.go:40-48` currently pins this ambiguity as if it were a contract. Suggestion: classify on an explicit `NodeType` field or a `oneof` in the proto rather than on field presence; at minimum make the SN-side construct/clone path guarantee a non-nil `StreamingNode` (have `cloneStreamingNodeView` return `&viewpb.QueryViewOfStreamingNode{}` for a nil input) and assert the invariant at the Coord construction site in `coordview/state_machine.go:335`. Since the transport/Manager layer is still TODO, this is not reachable in-repo today — but it is a trap laid for the integration layer that will consume these protos. (raised by xiaocai2333)

chyezh:
Follow-up to https://github.com/milvus-io/milvus/pull/51887#issuecomment-5125504117.

The remaining Low finding depends on bypassing the production QueryView construction contract and passing a nil StreamingNode view directly into an internal constructor. The complete qv integration does not produce that input:

- the Balancer creates views through `NewQueryViewAtCoordBuilder`
- `QueryViewAtCoordBuilder.Build()` unconditionally initializes `StreamingNode` with `&viewpb.QueryViewOfStreamingNode{}`
- `ShardViewManager.AddPreparing()` accepts the builder and calls `Build()` to create the Coord state machine
- subsequent Coord syncs, SN reports, persistence, and recovery preserve that non-nil discriminator
- the builder tests explicitly assert that the generated StreamingNode view is non-nil

Therefore, the reproduced `NewFullQueryViewAtStreamingNode(..., nil, ...)` case is a semantically invalid direct-constructor input rather than a reachable production path. Adding a node-type field, normalizing nil, or changing the state-machine/proto behavior would add defensive handling for an input excluded by the ownership and construction contract, so no change is needed for this PR.

I am triggering another adversarial review against the current head `21d2b91a62`. Please evaluate findings against the complete production construction and owner call chains, and distinguish invalid direct state-machine/constructor inputs from reachable behavior.

czs007:

## Re-review of `21d2b91a6285`

- **[low] `internal/views/qviews/view.go:57` — `NewQueryViewAtWorkNodeFromProto` distinguishes node types solely by `pb.StreamingNode != nil`, while the SN-side report clears that field to nil when `snView == nil`: Confirmed.** The dispatch chain is unchanged at this head, and reviewers additionally traced that `proto.Clone` preserves a typed-nil message rather than materializing one, so a nil `streaming_node` survives every hop from the SN report back into the Coord parse path. Restated with the full mechanism below.

---

Adversarial review confirms one low-severity latent correctness gap in the node-type dispatch of the new view package; no new critical or high issues were found this round.

## Low

**`internal/views/qviews/view.go:57` — SN views with an unset `streaming_node` are misclassified as QN views, or panic.**
The dispatch in `NewQueryViewAtWorkNodeFromProto` is `if pb.StreamingNode != nil { SN } else if len(pb.QueryNode) == 1 { QN } else { panic }` (view.go:57-68), and nothing else disambiguates the two node types. `QueryViewOfStreamingNode` is an empty message (`pkg/proto/view.proto:59-60`), so a producer has no data-carrying reason to set it, and nil is never intercepted downstream: `NewCoordQueryViewStateMachine` (coordview/state_machine.go:49-63) does not validate the field; `syncViewsForState` forwards it verbatim (coordview/state_machine.go:335) and `proto.Clone` on a typed-nil message returns a typed nil, so the assertion at view.go:81 succeeds with the field still nil; on the SN side `cloneStreamingNodeView(nil)` returns nil (snview/state_machine.go:341-346), so `buildReport` emits `StreamingNode == nil` alongside a non-empty `QueryNode` list (snview/state_machine.go:317-326).
*Failure:* a `QueryViewOfShard` arrives with `streaming_node` unset and exactly one `query_node` → the SN report is silently parsed as `*QueryViewAtQueryNode`, `updateNodeState` writes the SN's state into `qnStates[nodeID]` while `snState` stays `QueryViewStateNil`, `allNodesReady()` can never be satisfied and the view is stuck in `Preparing` forever; with 0 or ≥2 query nodes it panics at view.go:67 instead.
*Suggestion:* make the "an SN view must set `streaming_node`" contract explicit at the view entry point — either reject a `QueryViewOfShard` with an unset `streaming_node` in `NewCoordQueryViewStateMachine`, or normalize it by filling in the empty message — rather than relying on every future producer remembering to set a field that carries no data.
*Severity note:* rated low because no in-tree producer exists yet — at this head, `qviews` / `coordview` / `snview` / `qnview` have no non-test callers outside their own packages, and the Balancer/RPC transport layer is not yet wired up, so production reachability is unproven. The guard is cheap to add now, before producers land.
(raised by xaxys, czs007)

chyezh:
Follow-up to https://github.com/milvus-io/milvus/pull/51887#issuecomment-5138865607.

The re-review confirms only the conditional behavior after the QueryView construction invariant has already been violated; it does not establish a reachable correctness gap. This PR intentionally extracts the state machines while leaving the Balancer/Manager integration outside the picked scope, so the absence of an in-tree producer at this head is expected rather than evidence that future producers are unconstrained.

In the complete qv construction chain, the Balancer creates views through `NewQueryViewAtCoordBuilder`; `Build()` unconditionally sets `StreamingNode: &viewpb.QueryViewOfStreamingNode{}`; `ShardViewManager.AddPreparing()` accepts that builder and calls `Build()`; and subsequent sync, report, persistence, and recovery paths preserve the discriminator. A direct `NewCoordQueryViewStateMachine` / `NewFullQueryViewAtStreamingNode` call with an unset SN view bypasses that ownership and construction contract and is therefore a semantically invalid internal input.

Per the existing code convention, we should not add validation, normalization, or a new proto discriminator solely to handle an input excluded by the production preconditions. No code change is planned for this finding. This is consistent with the adversarial review result of 0 must-fix issues and mergeable as-is.

czs007:

A single Low-severity construction-contract issue remains on `internal/views/qviews/view.go`; no new critical or high issues were found in this round.

## Re-review of `21d2b91a6285`

- **`internal/views/qviews/view.go:57` — "NewQueryViewAtWorkNodeFromProto relies solely on `pb.StreamingNode != nil` to distinguish node types, while the SN-side report clears that field to nil when `snView == nil`": **Confirmed** — re-verified at this head: the dispatch has no other discriminator, and `cloneStreamingNodeView(nil)` returning nil (`snview/state_machine.go:341-345`) is put straight back into the report by `buildReport` (`:324`).
- **`internal/views/qviews/view.go:57` — "NewQueryViewAtWorkNodeFromProto decides node type purely by `pb.StreamingNode != nil`, so an SN view without `streaming_node` set is misclassified as a QN view or panics": **Confirmed** — this is a duplicate of the finding above and stands for the same reason; the misclassification and the panic branch are both real behaviours of `view.go:57-68`.

## Low

- **`internal/views/qviews/view.go:57`** — `NewQueryViewAtWorkNodeFromProto` disambiguates node types using only `if pb.StreamingNode != nil` (SN), else `len(pb.QueryNode) == 1` (QN), else `panic`; there is no node-type enum and no vchannel/node-id cross-check. Because `QueryViewOfStreamingNode` is an empty message (`pkg/proto/view.proto:53-56`), a producer has no data-carrying reason to set the field — presence is pure convention — and the in-repo chain is nil-preserving with no validation: `coordview/state_machine.go:335` passes `sm.view.StreamingNode` through unchecked, `proto.Clone` on a typed-nil returns a typed-nil so `view.go:81` neither panics nor fills it in, and `cloneStreamingNodeView(nil)` propagates the nil back into the SN report. An SN report with `StreamingNode == nil` and exactly one QN entry would then be classified as `QueryViewAtQueryNode` and written into `qnStates` instead of `snState` (`coordview/state_machine.go:344-352`), so the view could never satisfy `allNodesReady`/`allNodesDropped` (`:372-394`); with any other QN count it panics on the Coord. This is a latent construction-contract fragility rather than a reachable defect today — the only in-tree callers of the converter are tests, and the Balancer/transport that would build the wire proto is not part of this PR. Suggestion: add a one-line guard in `NewQueryViewAtWorkNodeFromProto` that rejects a proto carrying neither a `streaming_node` nor a well-formed single QN entry, or set `StreamingNode` unconditionally at the single construction site once the Balancer lands. *(raised by xaxys, czs007)*

docs/design-docs/design_docs/qviews/query_view_state_machine.md
```diff
@@ -0,0 +1,459 @@
...
+  - Up (recovery scenario: SN restored an old Up view from persistence) → Coord fast-forwards to Up.
+- **QN**: Preparing / Ready / Unrecoverable
+  - Preparing → Coord waits.
+  - Ready → Coord marks this QN as ready; checks if all nodes are ready.
+  - Unrecoverable → Coord transitions to Unrecoverable.
+
+### 1.2 Ready
+
+**Entry Conditions:**
+- All SN and QNs have reported Ready (automatic transition from Preparing).
```
weiliu1031, line 50: what happens if new qn startup before all available qn and sn report ready status? should the data view bind to current node view ?

docs/design-docs/design_docs/qviews/query_view_state_machine.md
```diff
@@ -0,0 +1,459 @@
...
+
+**Possible Peer States (and Coord's reaction):**
+- **SN**: Ready / Up
+  - Ready (Up push not yet delivered) → Coord re-pushes Up.
+  - Up → Coord transitions to Up.
+- **QN**: Ready / Unrecoverable
+  - Ready → Coord does nothing; normal.
+  - Unrecoverable → Coord transitions to Unrecoverable.
+
+> Note: Ready is NOT persisted. On Coord crash recovery, ETCD still shows Preparing; Coord re-pushes and catches up from node responses.
```
weiliu1031, line 70: which state will a newly started qn reports?

docs/design-docs/design_docs/qviews/query_view_state_machine.md
```diff
@@ -0,0 +1,459 @@
...
+    report directly; the query path detects the unavailable view and triggers
+    replacement.
+- **QN**: Ready / Unrecoverable
+  - Ready → Coord does nothing; normal.
+  - Unrecoverable → Coord transitions to Unrecoverable.
+
+### 1.4 Down
+
+**Entry Conditions:**
+- Higher-version view confirms Up; Coord immediately transitions the old Up
```
weiliu1031, line 103: will view be binded to collection level or replica level for single shard collection? if a collection decrease from 5 replica to 3 replica, what will happens?

docs/design-docs/design_docs/qviews/query_view_state_machine.md
```diff
@@ -0,0 +1,459 @@
...
+**Possible Coord States (and this node's reaction):**
+- Coord in Up / Down → SN does nothing; normal.
+- Coord pushes Down → SN transitions to Down.
+- Other signals → SN ignores.
+
+### 2.4 UpRecovering (StreamingNode-Only Proto State)
+
+UpRecovering is defined in the proto enum (`QueryViewStateUpRecovering = 8`) but is only used by StreamingNode.
+Coord and QueryNode never enter this state. For Coord-visible reporting, UpRecovering maps to Up
+(Coord is unaware of UpRecovering and considers the view to be in Up state).
```
weiliu1031, line 294: if uprecovering state is maps to up, which means this sn will be used for query, and the query request will be stuck at sn for waiting sn to finish upRecovering state ?

docs/design-docs/design_docs/qviews/query_view_state_machine.md
```diff
@@ -0,0 +1,459 @@
...
+  view to Down.
+- ReleaseCollection.
+- Recovery: loaded from ETCD in Down state.
+
+**Automatic Behavior:**
+1. Persist Down to ETCD (if transitioning from Up).
+2. Push Down to SN.
+
+> Query lease ownership: Coord does not wait for a lease period before entering
+> Down and always keeps at most one Up view. After receiving Down, StreamingNode
```
weiliu1031, line 113: will query stuck the state mechine, if query is executing, the qn may become dropped state?  or even if qn becomes down state, the references will keep the target segments alive?

docs/design-docs/design_docs/qviews/README.md
```diff
@@ -0,0 +1,336 @@
...
+- In this example, `SegmentID @DataVersion` labels the DataVersion when the
+  segment joined the view. DataView itself stores loadable membership and the
+  collection-level DataVersion, not a per-segment view-version field.
+- After Compaction, old Segments are permanently removed from the view.
+
+### 5.4 Constraints
+
+- Each DataVersion can correspond to one or more segment membership changes.
+- Once a Segment is compacted, it is removed from the storage view and will never return.
+- The view is not affected by offline tasks such as indexing.
```
weiliu1031, line 128: so new data view may cause performance downgrade if lack of index?

docs/design-docs/design_docs/qviews/query_view_state_machine.md
chyezh: A QueryView's node assignment is fixed when it is generated. A QN that starts while the view is Preparing is not added to that in-flight view; the DataView remains bound to the current assignment, and the new QN may be used by a later QueryView.

docs/design-docs/design_docs/qviews/query_view_state_machine.md
chyezh: A newly started QN does not report any QueryView state by itself. After Coord assigns it a Preparing view, it reports Preparing as segment readiness progresses, and Ready after all assigned segments are ready.

docs/design-docs/design_docs/qviews/query_view_state_machine.md
chyezh: No. During task acquisition, a QueryView reference delays resource release. After segment handles are acquired, their query references keep the target segments alive, so the view can advance to Dropped while the in-flight query finishes safely. QN has no Down state, and new acquisitions for the dropped view are rejected.

docs/design-docs/design_docs/qviews/query_view_state_machine.md
chyezh: QueryView is scoped per ShardOnReplica, so each `(replica, vchannel)` has an independent view. When scaling from 5 replicas to 3, the two removed replica views are released through Down -> Dropping -> Dropped, while the retained replicas continue independently.

docs/design-docs/design_docs/qviews/README.md
chyezh: For flushed segments, loading before index completion is allowed, so a temporary performance downgrade is possible. Compaction output segments are not loaded through DataView until indexing is complete. A future optimization can also prevent flushed segments from joining DataView before their indexes are ready.

docs/design-docs/design_docs/qviews/query_view_state_machine.md
chyezh: UpRecovering maps to Up only in the Coord-facing control plane. The SN query path requires the local state to be exactly Up, so an UpRecovering view is not selected; the request receives a retriable view error instead of waiting for recovery.

czs007:
> Also reported: 1 disputed item (our reviewers disagree — for the author to adjudicate, not a required fix).

A stale-sync leak in the coordinator state machine's Unrecoverable transition is the one confirmed finding this round, plus one open disagreement between our reviewers on the recovery-path Up report.

---

## Re-review of `21d2b91a6285`

- `internal/views/qviews/view.go:57` — `NewQueryViewAtWorkNodeFromProto` distinguishes node type solely by `pb.StreamingNode != nil`, while the SN-side report clears that field when `snView == nil`: **Not re-examined** — nobody looked at it this round; it stands.
- `internal/views/qviews/view.go:57` — an SN view with `streaming_node` unset can be misclassified as a QN view or panic under the same nil-based type check: **Not re-examined** — nobody looked at it this round; it stands.
- `internal/views/qviews/view.go:57` — node-type determination relies only on the `StreamingNode` field being non-nil, misclassifying or panicking when it is nil: **Not re-examined** — nobody looked at it this round; it stands.

## High

- `internal/views/coord/coordview/state_machine.go:314` — `transitionToUnrecoverable` sets `pendingPersist` but does not clear `pendingSync`, so an Up sync queued while in Ready (lines 240/262) survives the transition to Unrecoverable; the constructor-enqueued Preparing sync likewise survives the `OnQueryNodeLost` path. With the state machine in Ready and a pending Up sync accumulated before consumption (a usage pattern `TestPendingOverwrite_LastWins` explicitly endorses), a node reporting Unrecoverable followed by a normal `ConsumePersist` + `ConsumeSync` → Unrecoverable is persisted to etcd while the stale Up is still dispatched to the streaming node, which activates a view the coordinator has already declared dead (in the QN-lost scenario, one pointing at a downed query node), and since the coordinator ignores node reports in Unrecoverable, the divergence persists until `EnterDropping` pushes Dropped. Note the trigger requires the Manager consumer specified in the design doc but not yet merged in this PR. No documented contract requires draining after every event — the tests codify the opposite — so a fully contract-compliant consumer hits this; by contrast, `EnterDown` (line 183) and `EnterDropping` (line 196) both overwrite `pendingSync` on transition. Suggestion: set `pendingSync = nil` in `transitionToUnrecoverable`, and add a regression test that enqueues a sync, skips the intermediate consume, and then enters Unrecoverable. (raised by bigsheeper, sijie-ni-0214)

## Disputed — our reviewers disagree

- `internal/streamingnode/server/wal/snview/state_machine.go:68` — on the recovery path the local state is `UpRecovering` but is reported to the coordinator as Up, flagged as prematurely confirming serviceability. Refuting evidence: `state_machine.go:240-252,305-320` report Up for an unchanged `UpRecovering` state when Up is re-pushed, and `coordview/state_machine.go:248-259` accepts it as Ready → Up while `:176-185` treats a higher-version Up as the trigger to Down the old view — if v2 was persisted but its original Up report was lost, then after a restart v1 may finish recovery while v2 is still recovering; the retried Up falsely confirms v2, the coordinator downs v1, and no view can generate new query plans until v2 catches up. Verification wanted to drop it because design doc §2.4 defines this mapping as intended: coordinator-side Up means ownership/metadata is established, with serviceability guarded by SN-local query planning, which the code matches. (raised by bigsheeper, sijie-ni-0214; drop refuted by xaxys)

These disputed items are flagged for the author to adjudicate; they are not filed as required fixes.

chyezh:
Follow-up to the adversarial review at https://github.com/milvus-io/milvus/pull/51887#issuecomment-5247418041.

Fixed the confirmed stale pending Sync issue in `240f9e9fb4`, after first aligning the Coord state machine with qv's atomic `ConsumeFlush` flow in the independent commit `fc2b6a0f0c`.

`transitionToUnrecoverable` now clears `pending.Sync`, so an unconsumed Preparing or Up sync cannot survive alongside the Unrecoverable persist. Added regression tests for both reachable cases: QueryNode loss before the initial Preparing flush is consumed, and an Unrecoverable report while a Ready-generated Up sync is still pending.

Verified all four QueryView state-machine packages with coverage above 90%, plus the Coord race test and `git diff --check`. The disputed `UpRecovering` to Coord-visible Up behavior remains unchanged and follows the documented ownership/recovery semantics.

czs007:

Verified:
- `internal/views/coord/coordview/state_machine.go:314-317` — `transitionToUnrecoverable()` now explicitly sets `sm.pending.Sync = nil` after queuing the `Unrecoverable` persistence state, cancelling both constructor-queued `Preparing` syncs and Ready-queued `Up` syncs so no stale sync survives the transition.
- `internal/views/coord/coordview/state_machine_test.go:383-395` — regression test covers the QN-lost-after-construction scenario and asserts the atomic flush contains the `Unrecoverable` persist with an empty `Sync`.
- `internal/views/coord/coordview/state_machine_test.go:397-412` — regression test covers the case where an `Up` sync was already queued after Ready, and likewise asserts no sync remains after the transition.

Reviewers also re-checked the incremental range `21d2b91a6285..HEAD` for newly introduced issues and re-ran the affected package tests; both came back clean.

czs007 (approved):
/lgtm

czs007:

## Re-review of `240f9e9fb4eb`

No findings from the earlier comment remained on record for this commit, so there is nothing to confirm or retract. This round's independent re-examination of the same head surfaced no new issues.

**Verified:**

- `pkg/proto/view.proto:78-105` — the version, state, node-payload, and bidirectional-stream fields have no stale tag conflicts, and view batches plus typed close messages are represented atomically; generated bindings match the source proto.
- `internal/views/qviews/view.go:70-125` — StreamingNode payloads retain the complete topology while QueryNode payloads contain exactly one node view, and outgoing protos are deep-copied.
- `internal/views/qviews/utils.go:112-122` — DataVersion and QueryViewVersion comparisons are lexicographic, as documented.
- `internal/views/coord/coordview/state_machine.go:139-162,225-318` — normal, recovery fast-forward, unrecoverable, and cleanup transitions preserve the documented lifecycle, and entering Unrecoverable cancels stale syncs so no outdated Preparing/Up/Down messages are sent.
- `internal/streamingnode/server/wal/snview/state_machine.go:159-180,275-300` — a failed Up recovery preserves its retryable recovery record, and Down deletes the record while deferring resource release until Dropped, consistent with the lease boundary.
- `internal/querynodev2/qnview/state_machine.go:173-180` — only assigned, deduplicated segments count toward Ready, and reports are lazily built and coalesced to the latest progress, with stale reports cleared on Dropped.

All three state machines document and rely on serialized access from their Manager, and no hidden concurrent access was found in the current implementation. The reviewed diff was confirmed byte-identical to the PR changes, and the four target packages pass their test suites under the prescribed build tags.

chyezh:
query view doesn't affect the e2e, so the e2e can be skipped.

Issue #51888, opened 2026-07-27
[Bug]: REST v2 insert rejects collections whose only vector fields are inside a Struct Array


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus image: `harbor.tencent-stage.zilliz.cn/manta/milvus:master-20260727-996402d`
- Version: `master-20260727-996402d4e8`
- Git commit: `996402d4e8`
- Build time: `Mon Jul 27 09:01:21 UTC 2026`
- Deployment mode: `STANDALONE`
- Message queue: `woodpecker`

Pods:

- `rest-struct-only-vector-jz-etcd-0` — 1/1 Running, 0 restarts
- `rest-struct-only-vector-jz-milvus-standalone-5d4dd4df4-6bws2` — 1/1 Running, 0 restarts
- `rest-struct-only-vector-jz-minio-656bc9645d-qp4dw` — 1/1 Running, 0 restarts

### Current Behavior

REST v2 rejects insert requests when the collection has no top-level vector field and its only vector field is an `ArrayOfVector` sub-field inside a Struct Array.

The insert fails consistently with:

```json
{
  "code": 1804,
  "message": "fail to deal the insert data, error: collection: demo has no vector field or functions: invalid parameter"
}
```

The collection schema is accepted and persisted correctly. Adding an unrelated top-level vector field makes the same Struct Array payload succeed.

### Expected Behavior

REST v2 should recognize vector fields inside Struct Array fields. The insert should succeed when the collection's only vector fields are Struct Array sub-fields.

### Steps To Reproduce

```python
import os

import requests

base = os.environ["MILVUS_URI"].rstrip("/")
headers = {
    "Authorization": f"Bearer {os.environ['MILVUS_TOKEN']}",
    "Content-Type": "application/json",
}
collection = "rest_struct_only_vector_repro"


def post(path, payload):
    response = requests.post(
        base + path,
        headers=headers,
        json=payload,
        timeout=30,
    )
    response.raise_for_status()
    return response.json()


try:
    post(
        "/v2/vectordb/collections/create",
        {
            "collectionName": collection,
            "schema": {
                "fields": [
                    {
                        "fieldName": "primary_key",
                        "dataType": "Int64",
                        "isPrimary": True,
                    }
                ],
                "structFields": [
                    {
                        "fieldName": "doc",
                        "fields": [
                            {
                                "fieldName": "v1",
                                "dataType": "ArrayOfVector",
[...]

PR #51890, opened 2026-07-27
fix: [3.0] recognize struct array vectors in REST inserts

zhuwenxing:
pr: #51889
issue: #51888

## What changed

- Backport the REST Struct Array vector detection fix from #51889 to the `3.0` branch.
- Use `typeutil.GetVectorFieldSchemas` so nested `ArrayOfVector` fields count as vectors.
- Add regression coverage for a Struct Array as the only vector-bearing field.
- Preserve rejection for schemas with no vector fields or functions.

## Root cause

`nameDims` is populated only from top-level fields. The validation ran before Struct Array conversion, so it incorrectly rejected schemas whose vectors exist only in Struct Array sub-fields.

## Verification

- `TestAnyToColumnsStructArrayAsOnlyVectorField`
- `TestAnyToColumnsRejectsSchemaWithoutVectorOrFunction`
- The same change passed the full `internal/distributed/proxy/httpserver` Go test package on master.
- REST v2 E2E on the master fix with Float, Float16, Int8, and Binary vector sub-fields: `code=0`, `insertCount=1`
- 3.0 CI pending.

### DIFF
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -1810,7 +1810,7 @@ func anyToColumns(rows []map[string]interface{}, validDataMap map[string][]bool,
 			IsDynamic: field.IsDynamic,
 		}
 	}
-	if len(nameDims) == 0 && len(sch.Functions) == 0 && !partialUpdate {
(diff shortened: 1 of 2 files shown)


zhengbuqian (approved):
/lgtm

PR #51891, opened 2026-07-27
enhance: [3.0] unify RecordReader ownership on a borrow contract (#51884)

MrPresent-Han:
Cherry-pick of #51884 to 3.0. Clean pick — identical patch-id with the master commit.

Unifies all `RecordReader` implementations on a single borrow contract: records returned by `Next()` are reader-owned and valid until the next `Next()`/`Close()`; callers that need a record longer must `Retain()` it.

- storage: V1 `CompositeBinlogRecordReader` holds and releases its current record, mirroring the packed/manifest readers; `simpleArrowRecordReader` returns a fresh wrapper per batch (a retained record survives advance), makes the per-blob closer idempotent, surfaces mid-stream read errors via `rr.Err()`, and advances across empty blobs instead of returning premature EOF
- compactor: the record materializer releases only its derived arrays through an idempotent cleanup channel, never the borrowed base record
- consumers: bump/mix/DeleteCodec/external paths stop releasing reader-owned records; cross-batch holders keep explicit `Retain`/`Release`
- tests: refcount and lifecycle unit tests for both readers; suites verified under `-tags dynamic,test,assert` on master

issue: #51649
pr: #51884

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -514,20 +514,17 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		if preMaterializeFilter {
 			selection, _, err = selectFullRewriteRecord(record, pkField, entityFilter, ttlFieldID, sourceHasTTLField, nil)
 			if err != nil {
-				record.Release()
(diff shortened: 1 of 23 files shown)


liliu-z (commented):
/lgtm
/approve

Issue #45881, opened 2025-11-26
[Feature]: Support External Collection with Lakehouse Integration


yanbinyang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Keywords

External Collection, Lakehouse, Paimon, Iceberg, Virtual Segment, Zero-ETL

### Overview

Milvus currently requires users to import external data (from lakehouse systems like Apache Paimon/Iceberg) through bulk insert operations, which creates data duplication and increases storage costs. This MEP proposes an External Collection mechanism that enables Milvus to query lakehouse tables directly without data copying, implementing a zero-ETL architecture for vector similarity search.

### Background & Motivation

With the rapid adoption of lakehouse architectures (Iceberg, Paimon), organizations store massive amounts of vector embeddings in lakehouse tables. Current pain points include:

-  Data Duplication: Users must copy data from lakehouse to Milvus, doubling storage costs
-  Complex ETL Pipelines: Maintaining sync between lakehouse and Milvus requires additional infrastructure
-  Delayed Updates: Batch import causes data freshness issues
-  Operational Overhead: Managing two separate storage systems increases complexity

A native lakehouse integration allows:

- Zero-ETL: Direct query on lakehouse tables without data movement
- Cost Efficiency: Single copy of data in lakehouse
- Simplified Architecture: Eliminate sync pipelines
- Real-time Freshness: Query latest lakehouse snapshots directly

### High-Level Design

1. External Collection: New collection type with `collection.external.enabled=true` property
2. Lakehouse Catalog: Abstract interface supporting multiple lakehouse formats
3. Virtual Segment: Segment metadata pointing to lakehouse data splits (no binlog storage)
4. Data Refresh：Strategy for syncing lakehouse updates to external collection, controlled by `collection.external.data_refresh` property:
    - LATEST(default): Virtual Segments are automatically refreshed to reference the latest lakehouse snapshot when collection.load() is called. Each load operation triggers a new Fill Job.
    - STATIC: Virtual Segments reference a fixed lakehouse snapshot captured at collection creation time, the snapshot can be manul refresh. Each refresh also trigger a new Fill Job.
7. Bulk Fill Pipeline: Batch job to create Virtual Segments from lakehouse snapshots
8. Table Reader: Read lakehouse data on-demand during query/index build

### User Workflow

```python
# Define schema (must match lakehouse table schema)
schema = CollectionSchema(fields=[
[...]

PR #51892, opened 2026-07-27
fix: preserve segment ownership across parallel refresh tasks

weiliu1031:
issue: #45881

## Summary

Make external collection refresh parallel-safe while preserving file-range
task splitting.

- Explore the source once per planning attempt and build balanced continuous
  base ranges from `dataCoord.externalCollectionFilesPerTask`.
- Compute transitive continuous-range closure so every surviving file used by
  a baseline segment is visible to exactly one owner task.
- Persist the complete ownership plan before scheduling and send each
  DataNode only its owned baseline segments.
- Persist task-local kept and updated results, validate exclusive ownership,
  and apply one aggregate segment update after all sibling tasks finish.
- Make job-level segment apply replay-safe after restart and re-enqueue
  in-progress siblings of failed jobs only for worker cancellation.
- Use attempt-scoped Explore manifests with terminal job cleanup.
- Document the task partitioning and ownership mechanism.
- Add structured diagnostics for source file diffs, task closure, segment
  outcomes, physical row totals, row deltas, and update reasons.

## Why

A parallel task inspected only its assigned file range but previously
received every active baseline segment. DataNode interprets an old fragment
missing from that range as removed, so a task could invalidate a segment
whose files were assigned to a sibling task.

The resulting job could report completion while exposing only part of the
source rows.

Unlike the single-task fallback proposed in
[#51772](https://github.com/milvus-io/milvus/pull/51772), this change retains
configured parallelism while giving every baseline segment one exclusive
owner.

## Behavior boundaries

- `externalCollectionFilesPerTask` remains the target size of the initial
  balanced split, not a hard maximum.
- Ownership closure may merge task ranges, including falling back to one task
  when segment file references transitively cross every boundary.
- DataCoord plans ownership at file-path level. DataNode still validates exact
  fragments using file path and row range.
- Tasks without the persisted ownership-plan version are rejected rather than
  inferred during recovery.
- The change is limited to external collection refresh processing.

## Validation

```bash
source ~/.profile && source scripts/setenv.sh

go test -tags dynamic,test -gcflags="all=-N -l" \
  -ldflags="-r ${RPATH}" -count=1 -v \
  github.com/milvus-io/milvus/internal/datacoord \
  github.com/milvus-io/milvus/internal/datanode/external \
  -run 'TestPlanExternalRefreshOwnership|TestCreateTasksForJob_PersistedOwnershipDrivesWorkerRequest|TestCreateTasksForJob_CompositePersistenceFailureIsUnpublished|TestExternalCollectionRefreshManager_ApplyFinishedJobSegments|TestExternalCollectionRefreshMeta_PublishTaskPlan|TestExternalCollectionRefreshInspector_ReloadFromMeta|TestRefreshExternalCollectionTaskSuite'

git diff --check
```

The full Go suite and external-table E2E suite were not run.


internal/datacoord/task_refresh_external_collection.go
```diff
@@ -507,6 +548,31 @@ func normalizeExternalRefreshUpdatedSegment(
...
+		existing.GetPartitionID() == incoming.GetPartitionID() &&
+		existing.GetInsertChannel() == incoming.GetInsertChannel() &&
+		existing.GetState() == incoming.GetState() &&
+		existing.GetNumOfRows() == incoming.GetNumOfRows() &&
+		existing.GetManifestPath() == incoming.GetManifestPath() &&
+		existing.GetSchemaVersion() == incoming.GetSchemaVersion() &&
+		existing.GetStorageVersion() == incoming.GetStorageVersion() &&
+		existing.GetLevel() == incoming.GetLevel() &&
+		existing.GetIsSorted() == incoming.GetIsSorted() &&
+		proto.Equal(
```
liliu-z: The replay-dedup check compares `proto.Equal(&datapb.SegmentInfo{Binlogs: existing.GetBinlogs()}, ...{Binlogs: incoming.GetBinlogs()})`, but V3 catalog persistence strips binlogs on write (`buildAlterSegmentsKvs` calls `resetBinlogFields` and skips the per-FieldBinlog KVs for V3 segments) and every external-refresh segment is V3, so a reloaded `existing` segment comes back with empty `Binlogs` while the persisted task result still carries the non-empty fake binlogs that validation requires. If DataCoord restarts after apply persists the segments but before the Finished-job write, replay sees the two binlog sets differ, reports an ID collision, and marks the already-applied job Failed via `modPack.fail`. Compare only durable fields (`NumOfRows`/`Stats`) and drop the `Binlogs` term.

pkg/proto/data_coord.proto
```diff
@@ -1600,6 +1600,12 @@ message ExternalCollectionRefreshTask {
   repeated int64 kept_segments = 15;  // Segment IDs kept by this task result
   repeated SegmentInfo updated_segments = 16;  // Segment upsert payload produced by this task result
   bool result_ready = 17;  // True once this task has persisted job-level refresh result payload
+  // Version of the persisted file-range and segment-ownership contract.
+  // Zero denotes a legacy task without ownership metadata and is rejected.
+  int32 ownership_plan_version = 19;
+  // Immutable baseline segment IDs assigned exclusively to this task.
+  // These IDs are ownership input, not task results.
+  repeated int64 owned_segment_ids = 20;
```
liliu-z, line 1609: A task persists baseline segment IDs twice: owned_segment_ids (set at task creation) and kept_segments (set at task result). For an unchanged refresh the ownership closure can collapse every baseline segment into one task (the PR documents this fall-back), so both repeated-int64 arrays hold all N IDs. SaveExternalCollectionRefreshTask marshals the whole task into a single uncompressed etcd value, so at roughly 18·N bytes the record crosses etcd's ~1.5 MiB value limit near N≈87k; UpdateTaskResult then fails permanently and the refresh times out. Store the baseline IDs once (e.g. persist removed instead of kept, deriving kept from owned) or shard the record.

internal/datacoord/external_collection_refresh_meta.go
```diff
@@ -452,10 +448,24 @@ func (m *externalCollectionRefreshMeta) UpdateJobProgress(jobID int64, progress
...
 
-// AddTaskIDToJob adds a taskID to job's task_ids list
-func (m *externalCollectionRefreshMeta) AddTaskIDToJob(jobID int64, taskID int64) error {
-	_, err := m.mutateJob(jobID, "add taskID to job", func(job *datapb.ExternalCollectionRefreshJob) (bool, error) {
-		job.TaskIds = append(job.TaskIds, taskID)
+// PublishTaskPlan commits the complete ordered task list after every task
+// record has been persisted. An empty job task list is the unpublished state.
+func (m *externalCollectionRefreshMeta) PublishTaskPlan(jobID int64, taskIDs []int64) error {
+	_, err := m.mutateJob(jobID, "publish external refresh task plan", func(job *datapb.ExternalCollectionRefreshJob) (bool, error) {
+		if job.GetState() != indexpb.JobState_JobStateInit {
```
liliu-z: The Init-state and no-double-publish guards live only in PublishTaskPlan, which is referenced exclusively from tests. The production publish path createTasksForJob → AddTasksToJob (line 929 → 489) re-fetches the job under lock but never checks its state. Because processJob runs tryTimeoutJob on Init jobs, a job can be marked Failed and have its explore manifest removed (handleJobFailed → cleanupExploreTempForJob) during the slow Explore inside createTasksForJob; AddTasksToJob then still appends the tasks and the manager enqueues them against the now-Failed, manifest-less job (wasted dispatch plus DataNode-created manifests that leak because the Failed job is never applied). Enforce the Init-state / no-double-publish guard on the AddTasksToJob path, or route publication through PublishTaskPlan.

internal/datacoord/task_refresh_external_collection.go
weiliu1031: Fixed. Replay no longer compares the fake binlogs that V3 catalog persistence strips. externalRefreshNewSegmentAlreadyApplied now requires the durable segment, collection, and partition identity to match, then uses packed.CompareManifestPath(existing, incoming) >= 0. An equal or newer version of the same manifest base path is treated as an idempotent replay, while reuse of the segment ID for a different manifest still remains a collision. Restart/replay coverage was added for this path.

pkg/proto/data_coord.proto
weiliu1031: Fixed by moving ownership-plan task results out of the etcd task value. Versioned tasks serialize kept_segments and updated_segments as a deterministic protobuf in object storage; etcd retains only result_storage_version, result_path, and the SHA-256 checksum alongside the immutable ownership input. Loading verifies both the checksum and embedded task identity. Legacy version-0 inline results remain readable, while new ownership tasks no longer duplicate the kept IDs or carry SegmentInfo payloads in one etcd value.

internal/datacoord/external_collection_refresh_meta.go
weiliu1031: Fixed at the production publication boundary. AddTasksToJob now re-fetches the job while holding the job and task locks, then rejects non-Init jobs, empty plans, and jobs that already have task IDs before the catalog update. createTasksForJob publishes through this method, so timeout/failure and late Explore publication are serialized: whichever acquires the job lock first wins, and a stale Explore result cannot publish tasks to a terminal job. The late-plan cleanup case is covered by a production-path test.

internal/datacoord/external_collection_refresh_manager.go
```diff
@@ -731,68 +849,101 @@ func (m *externalCollectionRefreshManager) createTasksForJob(
...
+	baselineSegments := make([]*datapb.SegmentInfo, 0, len(currentSegments))
+	for _, segment := range currentSegments {
+		baselineSegments = append(baselineSegments, segment.SegmentInfo)
+	}
+
+	segmentFragments, err := packed.BuildCurrentSegmentFragments(
+		baselineSegments,
+		createStorageConfig(),
+		nil,
+	)
```
tedxu: This call reads related manifests sequentially, and duplicates the work each DataNodes does. Consider make it parallel, and dispatch necessary data to DataNodes to avoid the duplicated read.

tedxu (commented):
`externalRefreshOwnershipPlanVersionV1` and `refreshExternalCollectionTask.SetJobInfo` is now unreachable.

internal/datacoord/external_collection_refresh_planner.go
```diff
@@ -0,0 +1,197 @@
...
+// planExternalRefreshOwnership assigns each explored file index and baseline
+// segment to exactly one continuous task range. filesPerTask determines the
+// balanced base ranges; segment references then merge boundaries until every
+// surviving file of a segment is visible to its owner task. Manifest I/O and
+// task/segment ID allocation deliberately stay outside this pure planner.
+func planExternalRefreshOwnership(
+	files []*datapb.ExternalFileInfo,
+	segmentFragments packed.SegmentFragments,
+	filesPerTask int64,
+) ([]externalRefreshOwnershipTaskPlan, externalRefreshOwnershipSummary, error) {
```
tedxu, line 68: This function and its chunking logic could both be optimized further. Based on our offline discussion, we'll refactor them in future PRs.

tedxu (commented):
/lgtm
/approve

Issue #43638, opened 2025-07-30
[Enhancement]: Update Woodpecker Version Track


tinswzy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

This issue is mainly used to record the upgrade record of the Woodpecker version upgrade in milvus

PR #51896, opened 2026-07-27
enhance: upgrade Woodpecker to v0.1.36 and expose direct read config

tinswzy:
issue: #43638

## What

Upgrade Woodpecker from `v0.1.35` to `v0.1.36-0.20260727122421-dea6d6414ced`.

The pinned commit is the current Woodpecker master commit intended for the upcoming `v0.1.36` release. The final tag has not been cut yet.

This PR also exposes the Woodpecker service-mode direct-read settings through Milvus configuration:

- `woodpecker.client.directRead.enabled`
- `woodpecker.client.directRead.maxBatchSize`
- `woodpecker.client.directRead.maxFetchThreads`

Client-side group commit is already exposed through `woodpecker.client.segmentAppend.maxBatchEntries` and `maxBatchBytes`. Woodpecker has no separate group-commit enable flag; setting `maxBatchEntries` to `1` disables batching.

## Why

Woodpecker already provides disk-watermark monitoring and retriable write backpressure to warn and slow/block new appends before the local WAL disk is exhausted. This upgrade adds the complementary reclaim path: after a staged segment is durably compacted to object storage, Woodpecker can promptly release each replica's local `data.log` instead of waiting for truncate retention GC.

Together, these mechanisms reduce the risk of Woodpecker local disks reaching `no space left on device` under sustained DML workloads:

- keep the existing soft/hard disk-watermark warning and append backpressure;
- asynchronously notify replicas after compaction and reclaim compacted local staged data only after confirming the object-storage footer;
- recover cleanup progress across restarts, retry transient failures, and fall back to object storage for reads after local data is reclaimed;
- expose compaction, cleanup, reclaimed-byte, client-frontier, and decommission observability;
- add end-to-end coverage for compacted local data cleanup.

The upgrade also includes several reliability fixes:

- refuse compaction on replicas that are behind the coordinator-confirmed `LastEntryId`;
- synchronize writer shutdown with background auditor/notification tasks;
- require decommission completion before operator scale-down and PVC reclaim.

Relevant Woodpecker changes:

- https://github.com/zilliztech/woodpecker/pull/237
- https://github.com/zilliztech/woodpecker/pull/243
- https://github.com/zilliztech/woodpecker/pull/245
- https://github.com/zilliztech/woodpecker/pull/246
- https://github.com/zilliztech/woodpecker/pull/247
- https://github.com/zilliztech/woodpecker/pull/248

## Verification

- `go mod verify`
- `(cd pkg && go mod verify)`
- `make generate-yaml`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./cmd/tools/config -run '^TestYamlFile$'`
- `(cd pkg && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/paramtable -run '^TestServiceParam$')`
- `(cd pkg && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./streaming/walimpls/impls/wp -run '^TestSetCustomWpConfig(BatchParams|DirectReadParams)$')`

The complete Woodpecker adapter package test was also attempted locally, but its WAL integration setup requires etcd at `localhost:2379`; the local environment timed out connecting to etcd before exercising the WAL cases. Upstream Woodpecker component, integration, and e2e coverage is linked above.


czs007:
Adversarial review of the Woodpecker v0.1.36 upgrade and `directRead` config passthrough found no blocking issues; two minor items are noted below.

### Low

- **`go.mod:267`** — The Woodpecker dependency is pinned to the pseudo-version `v0.1.36-0.20260727122421-dea6d6414ced` (also in `pkg/go.mod:40`, with matching `go.sum` updates), while the PR title says "upgrade Woodpecker to v0.1.36". Once the `v0.1.36` tag is actually cut, replace both pseudo-version pins with the released tag before this reaches a release branch. *(raised by xiaocai2333)*

### Nitpick

- **`pkg/streaming/walimpls/impls/wp/builder.go:121`** — The three `directRead` knobs (`Enabled`/`MaxBatchSize`/`MaxFetchThreads`) are passed through via `GetAsBool()`/`GetAsSize()`/`GetAsInt()` with no bounds checking on the Milvus side (e.g. `maxBatchSize <= 0`, `maxFetchThreads <= 0`). This matches the established pattern of every sibling Woodpecker passthrough (`AppendMaxBatch*`, `SegmentRolling*`) and the defaults (`16M`/`4`) are valid, so it is not a blocker. Suggestion: add uniform clamping/validation in Woodpecker's own config layer, and consider a malformed-value test case in `builder_test.go` (the new `TestSetCustomWpConfigDirectReadParams` only covers defaults and valid custom values). *(raised by xiaocai2333)*

czs007:
Adversarial review of this change surfaced one high-severity issue that should be addressed before merge.

### High

- **`pkg/streaming/walimpls/impls/wp/builder.go:121`** — The new `DirectReadMaxBatchSize` / `DirectReadMaxFetchThreads` settings overwrite Woodpecker's validated defaults directly with `GetAsSize()` / `GetAsInt()` results, without validation and without a subsequent `Configuration.Validate()` call on the build path. These helpers return `0` for malformed strings, and explicit zero/negative values pass through as well. Since `directRead.enabled` defaults to `true`, a `maxBatchSize` of `0` causes the object-storage block-read loop for a sealed segment to schedule no blocks and return EOF; the upstream log reader interprets that EOF as segment completion and advances to the next segment, potentially skipping an entire segment of unread WAL data during consumption or recovery. A non-positive `maxFetchThreads` also degrades the intended bounded worker pool into one with unlimited concurrency. Suggestion: preserve Woodpecker's defaults or reject the configuration when either numeric value is non-positive or malformed, following the guarded pattern already used for the batch parameters at lines 104–117 of the same file, and add invalid-value test cases alongside the valid-only cases at `builder_test.go:97-112`. *(raised by xaxys, chyezh)*

tinswzy:
@czs007 Thanks for catching this. I traced both failure paths and agree that the unchecked passthrough was unsafe.

Fixed in `dbbd4186c0`:

- `directRead.maxBatchSize` and `directRead.maxFetchThreads` are now applied only when the parsed values are positive.
- Malformed, zero, or negative values keep the already-validated Woodpecker configuration instead of overwriting it with `0` / a non-positive pool size, and emit a warning with the original value.
- Added coverage for malformed, zero, and negative values, proving that the validated values remain unchanged.

I used targeted guards rather than calling `Configuration.Validate()` after every Milvus override, since a global validation pass could change startup behavior for unrelated existing Woodpecker passthrough settings.

Verified with:

```text
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./streaming/walimpls/impls/wp -run '^TestSetCustomWpConfigDirectReadParams$'
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/paramtable -run '^TestServiceParam$'
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./cmd/tools/config -run '^TestYamlFile$'
```

The earlier pseudo-version item is also resolved; the PR now pins the released `v0.1.36` tag.

czs007:
Adversarial review confirmed the dependency bump and direct-read config passthrough are sound; two low-severity items remain, both non-blocking.

## Low

**`pkg/streaming/walimpls/impls/wp/builder.go:121` — malformed `DirectReadEnabled` value silently disables direct read**

`wpConfig.Woodpecker.Client.DirectRead.Enabled = cfg.DirectReadEnabled.GetAsBool()` is applied unguarded, while the two numeric direct-read params six lines below get explicit warn-and-keep-default handling. `GetAsBool` funnels into `getAsBool -> getAndConvert(v, strconv.ParseBool, false)` (`pkg/util/paramtable/param_item.go:432-434`, `469-475`), which swallows the parse error and returns the zero value `false` — it does *not* fall back to the ParamItem's declared `DefaultValue: "true"` (`pkg/util/paramtable/service_param.go:863-869`). An ops typo such as `enabled: ture` in `milvus.yaml` therefore turns direct read off fleet-wide with no warning, and the effect is real: woodpecker gates both the object-storage client (`woodpecker_client.go:194-196`) and the per-segment direct-read path (`segment_handle.go:542-543`) on this flag, so every sealed-segment read silently reverts to the gRPC quorum path.

*Suggestion:* warn on unparseable input, mirroring the `maxBatchSize`/`maxFetchThreads` guards at `builder.go:127-138`; at minimum, document that malformed values land on `false` rather than on the declared `true` default. *(raised by xiaocai2333, xaxys)*

**`go.mod:267` — PR description contradicts the committed woodpecker pin**

The PR body states the upgrade is to `v0.1.36-0.20260727122421-dea6d6414ced` and that "the final tag has not been cut yet," but every committed pin is the clean release tag `v0.1.36` (`go.mod:267`, `pkg/go.mod:40`, both `go.sum` files, and both Helm references). The tag does exist upstream (`e99c5e63e0dbb1eb8d4b64d5b2b22f6fe9ce57bc`) and the committed `go.sum` hash matches the published zip byte-for-byte, so `go mod download` is not at risk — this is purely a stale description.

*Suggestion:* update the PR description before merge so reviewers and backporters are not chasing a pseudo-version that does not appear in the diff. *(raised by xiaocai2333)*

czs007 (approved):
/lgtm

PR #51898, opened 2026-07-27
build(deps): update Knowhere sparse codec fix

sparknack:
## What changed
- Pin Knowhere to sparknack/knowhere commit 62486cd1 from https://github.com/zilliztech/knowhere/pull/1738.
- Preserve the v8/v9 sparse flat-codec fix until it lands upstream.

## Test
- ninja -C cmake_build -j1 knowhere
- make verifiers (did not complete: local parallel C++ compilation hit host resource/compiler failures outside the Knowhere target)


sparknack:
just a test pr, close.

PR #51901, opened 2026-07-27
fix: keep sync dispatcher cleanup after task panic

bigsheeper:
issue: #51854

## What

Recover task and callback panics inside the key-locked sync dispatcher, convert them to typed system errors, and still run dispatcher completion bookkeeping.

This keeps the dispatcher future complete, releases pending capacity, clears the in-flight segment key, and lets later callbacks run even when an earlier task/callback panics. The growing-source writebuffer recovery callback can then roll back sync state and schedule retry instead of being skipped by the default panicking error handler.

## Verification

Local worktree note: this branch lives under `.worktrees/`, so the Go commands used `GOWORK=off`; the worktree also used ignored symlinks to the main checkout's `internal/core/output` and generated migration `legacypb` artifact.

- Repro first: `PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:$LD_LIBRARY_PATH GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/syncmgr -run 'TestKeyLockDispatcher/TestPanicStillCompletesAndDrains'` failed before the fix with `panic: mock task panic [recovered, repanicked]`.
- `PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:$LD_LIBRARY_PATH GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/syncmgr -run 'TestKeyLockDispatcher/TestPanicStillCompletesAndDrains'`
- `PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:$LD_LIBRARY_PATH GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/syncmgr`
- `PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:$LD_LIBRARY_PATH GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/writebuffer`
- `PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:$LD_LIBRARY_PATH GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/...`
- `GOWORK=off make static-check` -> 0 issues across core, pkg, client, and go_client e2e phases.

### DIFF
--- a/internal/flushcommon/syncmgr/key_lock_dispatcher.go
+++ b/internal/flushcommon/syncmgr/key_lock_dispatcher.go
@@ -9,6 +9,7 @@ import (
 	"github.com/milvus-io/milvus-proto/go-api/v3/msgpb"
 	"github.com/milvus-io/milvus/pkg/v3/metrics"
 	"github.com/milvus-io/milvus/pkg/v3/util/conc"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 	"github.com/milvus-io/milvus/pkg/v3/util/syncutil"
 )
@@ -170,18 +171,21 @@ func (d *keyLockDispatcher[K]) dispatchLocked(key K, pt *pendingTask) {
 	// By spawning a goroutine, the current worker function can return and
 	// release its slot, allowing the goroutine's Submit to proceed.
 	go func() {
-		f := d.workerPool.Submit(func() (struct{}, error) {
+		f := d.workerPool.Submit(func() (ret struct{}, err error) {
 			nodeID := paramtable.GetStringNodeID()
 			metrics.WALFlusherSyncDispatcherQueueDuration.WithLabelValues(nodeID).Observe(time.Since(pt.enqueueAt).Seconds())
 
 			startTime := time.Now()
-			err := pt.task.Run(pt.ctx)
+			defer func() {
+				metrics.WALFlusherSyncDispatcherExecuteDuration.WithLabelValues(nodeID).Observe(time.Since(startTime).Seconds())
+				onComplete(err)
+			}()
+
+			err = runTaskSafely(pt.ctx, pt.task)
 			for _, cb := range pt.callbacks {
-				err = cb(err)
+				err = runTaskCallbackSafely(cb, err)
 			}
-			metrics.WALFlusherSyncDispatcherExecuteDuration.WithLabelValues(nodeID).Observe(time.Since(startTime).Seconds())
-			onComplete(err)
-			return struct{}{}, err
+			return ret, err
 		})
 
 		// Detect pool rejection (e.g., pool closed during shutdown).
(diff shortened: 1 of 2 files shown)


czs007:
Both full reports contain nothing beyond the one-line concurrences already in the notes — no `file:line` checks anywhere. Since the rules forbid inventing Verified entries, the comment states that explicitly instead of fabricating a list.

---

**Verified:**
All findings raised in earlier rounds were resolved or dismissed during adversarial verification, and both final-round reviewers independently re-examined the change set and confirmed no additional issues remain. The final-round reviewer notes recorded concurrence only, without citing specific file/line checks, so no per-file verification entries are listed here.

---

Two things you should know before posting:

1. **No `Verified:` file list was possible.** The prompt requires 3–6 `file:line` entries taken only from the reviewer notes, but both notes (and their full reports, which I fetched) contain only "no additional issues" concurrences with zero file references. I wrote the section to say so rather than invent entries.
2. **Possible inconsistency in the inputs:** reviewer `yezh` wrote "Blocking: I agree with the existing high and medium findings," yet the post-verification issue JSON is empty (all drops excluded). If those high/medium findings were dropped in error rather than resolved, the clean-result comment above would misstate the review outcome — worth a quick confirmation before publishing.

bigsheeper:
Closing this — the premise is wrong.

**A panic in a sync task is meant to stop the process.** The chain is deliberate: `conc.Pool.Submit` recovers and re-panics (`pool.go:71-76`, `panic(x) // throw panic out`); the default `PanicHandler` re-panics unless containment is requested (`options.go:53-58`); containment is opt-in and defaults off (`concealPanic: false`, `options.go:78`) with no production caller of `WithConcealPanic` anywhere in the repo. On this path specifically, `wb.errHandler` defaults to panicking (`write_buffer.go:1568`, `options.go:48-51`) and nothing overrides it.

This PR neutralizes that at one call site. `runTaskSafely` recovers before the panic reaches any of it, so the stop-the-node mechanism still exists but no longer does anything here — and nothing in `key_lock_dispatcher.go` says so. The description gives the motivation away: the recovery callback should run "instead of being skipped by the default panicking error handler". That handler is the design, not an obstacle.

If the underlying want is legitimate — #51854 exists because the panicking callback preempts writebuffer recovery — it needs to be a design change argued in the open: either a non-panicking `errHandler` for growing-source with a justification, or `conc.WithConcealPanic(true)` on this pool with the reason recorded in code. A hand-rolled `recover()` is neither.

Back to #51854.

Issue #51900, opened 2026-07-27
[Enhancement]: Include field name and actual type in range / IN / comparison type-mismatch errors


jsshin8128:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Type-mismatch errors for range (`a < f < b`), `in`, and comparison expressions in `internal/parser/planparserv2` should name the field that failed the check and its actual type. Today they don't. Range is the worst case: no field name, no field type, no offending bound value, and three separate mismatch cases that all come back as the same string, `invalid range operations`.

Current behavior on master (`61f7a6ca0`). I dropped a throwaway test into the package and ran these against its own `newTestSchema`:

| expression | current error |
| --- | --- |
| `1.5 < Int64Field < 3.5` | `invalid range operations` |
| `"a" < Int64Field < "z"` | `invalid range operations` |
| `1 < VarCharField < 5` | `invalid range operations` |
| `0 < BoolField < 1` | `invalid range operations on boolean expr` |
| `Int64Field in [1, "two", 3]` | `value 'string_val:"two"' in list cannot be casted to Int64` |
| `VarCharField > 5` | `comparisons between VarChar and Int64 are not supported` |

Repro:

```go
helper := newTestSchemaHelper(t)
_, err := ParseExpr(helper, `1.5 < Int64Field < 3.5`, nil)
// cannot parse expression: 1.5 < Int64Field < 3.5: invalid range operations: query plan failed
```

Something like this would be a lot easier to act on:

```
invalid range operations: bound value 1.5 (Double) does not match field Int64Field (Int64)
invalid range operations on boolean expr: field BoolField (Bool)
value "two" (VarChar) in list cannot be casted to field Int64Field (Int64)
comparisons between VarCharField (VarChar) and 5 (Int64) are not supported
```

The field name is already available as the parse-tree text of the operand. That's how PR #51022 picks it up for the arithmetic operators.

Sites:

- `internal/parser/planparserv2/utils.go`, `castRangeValue()`, lines 728 / 731 / 734 / 738. Callers are `parser_visitor.go`, `VisitRange` at 1408 and 1413, `VisitReverseRange` at 1495 and 1500.
- `internal/parser/planparserv2/parser_visitor.go`, `VisitTerm()`, line 1172.
- `internal/parser/planparserv2/utils.go`, `HandleCompare()`, line 532. This one already prints both data types, it's only missing the field names.

Done when:

[...]

PR #51022, opened 2026-07-02
enhance: add field names and types to arithmetic type mismatch errors

sharrmeen:
## Description

Previously, when a user tried to do arithmetic or bitwise operations on a non-integer field, the error message was completely generic (e.g., `"modulo requires an integer field"`). This made it hard to understand which field caused the issue.

This PR threads the original field name and its actual data type through the parser into the error generation. 

**Before:**
`modulo requires an integer field`

**After:**
`modulo requires an integer field, but field 'price' is Double`

## Scope of Changes

1. **Full Operator Coverage:** While the original issue only specifically mentioned bitwise and modulo, this PR goes further and applies the exact same field-name error formatting to **all** standard arithmetic operators (`+`, `-`, `*`, `/`) via `canArithmetic`.
2. **Introduced `arithField` Struct:** To safely pass the field metadata (`dataType`, `elementType`, and `name`) down to `checkValidModArith` and `canArithmetic`, I introduced a small `arithField` struct. This replaces the previous approach of passing 5+ loose positional arguments, which might be prone to compiler-silent bugs (like accidentally swapping `leftType` and `rightType`).
3. **E2E Testing:** Added `TestExpr_TypeMismatchErrorMessages` to ensure the entire flow (from raw string parsing down to error generation) works as intended and outputs the correct strings.

## Related Issue

issue: #50965

PR #51903, opened 2026-07-27
enhance: include field names and types in range, IN, and comparison errors

jsshin8128:
issue: #51900

## Summary

Type-mismatch errors for range, `in`, and comparison expressions in `planparserv2` did not indicate which field was rejected. Range errors had the least context: they omitted the field name, field type, and offending bound value, and all three mismatch cases returned the same generic message, `invalid range operations`. The `in` error also printed literals in protobuf text format (`string_val:"two"`).

* `castRangeValue()` now takes the source text of the column operand and reports both sides of the mismatch. The boolean case also includes the field name.
* `HandleCompare()` now takes the source text of both operands, so comparison errors can identify the operands instead of only their types. Both call sites, `VisitEquality` and `VisitRelational`, pass `ctx.Expr(0).GetText()` and `ctx.Expr(1).GetText()`.
* `VisitTerm()` now reports the field a list value failed to cast to.
* New helpers:

  * `formatGenericValue()` formats a literal in the same way the user wrote it.
  * `getValueDataType()` returns the parsed type of a literal.
  * `columnIdentifierText()` extracts the field name from the parse tree, since `ColumnInfo` only carries a field ID.

Field names are derived from the operand's parse-tree text, consistent with the approach used for arithmetic operators in PR #51022.

```diff
1.5 < Int64Field < 3.5
- invalid range operations
+ invalid range operations: bound value 1.5 (Double) does not match field Int64Field (Int64)
```

```diff
0 < BoolField < 1
- invalid range operations on boolean expr
+ invalid range operations on boolean expr: field BoolField (Bool)
```

```diff
1 < ArrayField[0] < "z"
- invalid range operations
+ invalid range operations: bound value "z" (VarChar) does not match field ArrayField[0] (Int64)
```

```diff
Int64Field in [1, "two", 3]
- value 'string_val:"two"' in list cannot be casted to Int64
+ value "two" (VarChar) in list cannot be casted to field Int64Field (Int64)
```

```diff
VarCharField > Int64Field
- comparisons between VarChar and Int64 are not supported
+ comparisons between VarCharField (VarChar) and Int64Field (Int64) are not supported
```

Error factories and error codes are unchanged. These remain input errors created via `merr.WrapErrQueryPlanMsg` or `merr.WrapErrParameterInvalidMsg`.

`castRangeValue()` keeps the `invalid range operations` prefix so existing assertions and any tooling that depends on that string continue to work.

## Not covered

Template-filled bounds such as `1 < f < {v}` are validated in `FillBinaryRangeExpressionValue()` through `castValue()`, which still reports:

```text
cannot cast value to Int64, value: string_val:"a"
```

`castValue()` and `checkContainsElement()` (used by the `json_contains*` path) have the same limitation and are tracked as follow-ups in #51900. This PR does not modify them.

There is no conflict with PR #51022. That PR updates `canArithmetic()`, `checkValidModArith()`, and their call sites, none of which are touched here.

## Verification

* Added `TestExpr_TypeMismatchNamesFieldAndType` in `plan_parser_v2_test.go`.

  * Covers 11 expressions across both range directions, array element access, `in`, field-vs-literal comparisons, and field-vs-field comparisons.
  * Verifies that errors include field name, field type, offending value, and value type.
  * Ensures protobuf-style representations like `string_val:` and `int64_val:` never appear in error output.
* Updated `Test_castRangeValue` for the new signature and added message-level assertions.
* Added `Test_formatGenericValue` and `Test_getValueDataType` in `utils_test.go`.

```text
LOCAL_STORAGE_SIZE=10 go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2/...
ok  github.com/milvus-io/milvus/internal/parser/planparserv2            5.232s
ok  github.com/milvus-io/milvus/internal/parser/planparserv2/rewriter   5.106s
```

The before/after messages above are the actual parser outputs captured by running each expression through `ParseExpr` on this branch.

### DIFF
--- a/internal/parser/planparserv2/parser_visitor.go
+++ b/internal/parser/planparserv2/parser_visitor.go
@@ -1205,8 +1205,15 @@ func (v *ParserVisitor) VisitTerm(ctx *parser.TermContext) interface{} {
 		for i, e := range array {
 			castedValue, err := castValue(dataType, e)
 			if err != nil {
+				// dataType is printed as-is after the target is narrowed above. If the array
+				// target is not narrowed, dataType remains Array, so the element type is appended
+				// to reflect the expected shape of each value in the list.
+				fieldDataType := dataType.String()
+				if typeutil.IsArrayType(dataType) {
+					fieldDataType = getDataType(childExpr)
+				}
 				return merr.WrapErrParameterInvalidMsg("value %s (%s) in list cannot be casted to field %s (%s)",
-					formatGenericValue(e), getValueDataType(e), ctx.Expr(0).GetText(), dataType.String())
+					literalText(termElementText(ctx.Expr(1), i), e), getValueDataType(e), ctx.Expr(0).GetText(), fieldDataType)
 			}
 			values[i] = castedValue
 		}
@@ -1408,6 +1415,21 @@ func columnIdentifierText(identifier, child, structSubField, structIndexField an
 	}
 }
 
+// termElementText returns the source text of the i-th element in a term list. It
+// returns an empty string when the list is not a literal, such as when it is
+// produced by constant folding, since there is no source token available to quote.
+func termElementText(list parser.IExprContext, i int) string {
+	arrayCtx, ok := list.(*parser.ArrayContext)
+	if !ok {
+		return ""
+	}
+	elements := arrayCtx.AllExpr()
+	if i >= len(elements) {
+		return ""
+	}
+	return elements[i].GetText()
+}
+
 func (v *ParserVisitor) getChildColumnInfo(identifier, child, structSubField, structIndexField antlr.TerminalNode) (*planpb.ColumnInfo, error) {
 	if identifier != nil {
 		childExpr, err := v.translateIdentifier(identifier.GetText())
@@ -1546,12 +1568,12 @@ func (v *ParserVisitor) VisitRange(ctx *parser.RangeContext) interface{} {
 	lowerValue := lowerValueExpr.GetValue()
 	upperValue := upperValueExpr.GetValue()
 	if !isTemplateExpr(lowerValueExpr) {
-		if lowerValue, err = castRangeValue(fieldText, fieldDataType, lowerValue); err != nil {
+		if lowerValue, err = castRangeValue(fieldText, ctx.Expr(0).GetText(), fieldDataType, lowerValue); err != nil {
 			return err
 		}
 	}
 	if !isTemplateExpr(upperValueExpr) {
-		if upperValue, err = castRangeValue(fieldText, fieldDataType, upperValue); err != nil {
+		if upperValue, err = castRangeValue(fieldText, ctx.Expr(1).GetText(), fieldDataType, upperValue); err != nil {
 			return err
 		}
 	}
@@ -1633,12 +1655,12 @@ func (v *ParserVisitor) VisitReverseRange(ctx *parser.ReverseRangeContext) inter
 	lowerValue := lowerValueExpr.GetValue()
 	upperValue := upperValueExpr.GetValue()
 	if !isTemplateExpr(lowerValueExpr) {
-		if lowerValue, err = castRangeValue(fieldText, fieldDataType, lowerValue); err != nil {
+		if lowerValue, err = castRangeValue(fieldText, ctx.Expr(1).GetText(), fieldDataType, lowerValue); err != nil {
 			return err
 		}
 	}
 	if !isTemplateExpr(upperValueExpr) {
-		if upperValue, err = castRangeValue(fieldText, fieldDataType, upperValue); err != nil {
+		if upperValue, err = castRangeValue(fieldText, ctx.Expr(0).GetText(), fieldDataType, upperValue); err != nil {
 			return err
 		}
 	}
--- a/internal/parser/planparserv2/parser_visitor.go
+++ b/internal/parser/planparserv2/parser_visitor.go
@@ -546,7 +546,7 @@ func (v *ParserVisitor) VisitEquality(ctx *parser.EqualityContext) interface{} {
 		return ret
 	}
 
-	expr, err := HandleCompare(ctx.GetOp().GetTokenType(), leftExpr, rightExpr)
+	expr, err := HandleCompare(ctx.GetOp().GetTokenType(), leftExpr, rightExpr, ctx.Expr(0).GetText(), ctx.Expr(1).GetText())
 	if err != nil {
 		return err
 	}
@@ -601,7 +601,7 @@ func (v *ParserVisitor) VisitRelational(ctx *parser.RelationalContext) interface
 		return err
 	}
 
-	expr, err := HandleCompare(ctx.GetOp().GetTokenType(), leftExpr, rightExpr)
+	expr, err := HandleCompare(ctx.GetOp().GetTokenType(), leftExpr, rightExpr, ctx.Expr(0).GetText(), ctx.Expr(1).GetText())
 	if err != nil {
 		return err
 	}
@@ -1205,7 +1205,8 @@ func (v *ParserVisitor) VisitTerm(ctx *parser.TermContext) interface{} {
 		for i, e := range array {
 			castedValue, err := castValue(dataType, e)
 			if err != nil {
-				return merr.WrapErrParameterInvalidMsg("value '%s' in list cannot be casted to %s", e.String(), dataType.String())
+				return merr.WrapErrParameterInvalidMsg("value %s (%s) in list cannot be casted to field %s (%s)",
+					formatGenericValue(e), getValueDataType(e), ctx.Expr(0).GetText(), dataType.String())
 			}
 			values[i] = castedValue
 		}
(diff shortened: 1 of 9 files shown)


xiaofan-luan:
Thanks for the detailed issue and for following up with a PR! The scope is clear and focused, and the before/after examples plus targeted tests make this easy to review. Please keep it moving through review—we appreciate the contribution!

jsshin8128:
The first CI run failed in `ci-v2/e2e-default` and `ci-v2/go-sdk` due to this change. `ci-v2/build-ut-cov` passed, which confirms the issue is not in the Go unit tests.

The Python and Go SDK test suites assert these parser error messages verbatim. My initial impact analysis only covered Go sources, so I missed the assertions under `tests/python_client`. In total, eight assertions were still expecting the old messages:

* `tests/python_client/milvus_client/test_milvus_client_query.py` (4 cases)
* `tests/python_client/milvus_client/test_milvus_client_search_v2.py`
* `tests/python_client/milvus_client/test_milvus_client_search_by_pk.py`
* `tests/python_client/milvus_client/test_milvus_client_search_none_default.py`
* `tests/python_client/testcases/test_delete.py`
* `tests/go_client/common/utils.go`

Fixed in `79ae96cd4`.

The updated expected strings were not written manually. I ran the same expressions through `ParseExpr` against schemas that mirror the field names and types used in each test, then copied the actual parser output.

`tests/go_client/testcases/query_test.go` still asserts `cannot cast value to Int64` for template values. That path goes through `castValue()`, which is not modified in this PR.

@xiaofan-luan Thanks for the kind words! I'll keep an eye on the CI run and address any review feedback as it comes in!

jsshin8128:
Rebased onto `master` and resolved conflicts in `parser_visitor.go`, `plan_parser_v2_test.go`, and `tests/go_client/common/utils.go`.

Conflict resolution details:

* On `master`, `VisitEquality` now hoists `leftExpr` and `rightExpr` above the constant-folding branch and calls `contextualizeEmptyArrayLiteral`. I kept that structure and only added the two operand source texts to the `HandleCompare` call.
* Kept both new tests in `plan_parser_v2_test.go`: `TestExpr_JSONBinaryRangePreciseBoundValidation` from `master` and `TestExpr_TypeMismatchNamesFieldAndType` from this PR.
* The Go SDK assertion for `floatVec in [0]` is no longer needed from this PR. On `master`, the target-type guard added in `VisitTerm` rejects the expression with `term expression is not supported for data type FloatVector` before it reaches `castValue()`. I kept the `master` implementation and removed my change for that case. The Python assertions are unaffected.

I reran the seven expressions covered by the Python suite on the rebased branch, and all of them still produce exactly the messages expected by the updated assertions.

`go test ./internal/parser/planparserv2/...` also passes locally.

xiaofan-luan (commented):
Thanks for the focused improvement. I left two inline suggestions for message-fidelity cases that are not covered by the current tests.

internal/parser/planparserv2/utils.go
```diff
@@ -555,11 +555,47 @@ func getDataType(expr *ExprWithType) string {
...
+// returns an internal protobuf representation (e.g. `string_val:"two"`), which must
+// never be exposed in a user-facing error message.
+func formatGenericValue(value *planpb.GenericValue) string {
+	switch v := value.GetVal().(type) {
+	case *planpb.GenericValue_BoolVal:
+		return strconv.FormatBool(v.BoolVal)
+	case *planpb.GenericValue_Int64Val:
+		return strconv.FormatInt(v.Int64Val, 10)
+	case *planpb.GenericValue_FloatVal:
+		return strconv.FormatFloat(v.FloatVal, 'g', -1, 64)
```
xiaofan-luan, line 574: `formatGenericValue` cannot preserve the literal as written because it only sees the value after parsing. On this head, `Int64Field in [1.0]` reports `value 1 (Double)`, and `0x1p+1 < Int64Field < 3` reports `bound value 2 (Double)`. Single-quoted and raw strings are normalized as well. This conflicts with the issue/test contract (“exactly as the user wrote it”) and makes output such as `1 (Double)` confusing. Could we thread the original parse-tree/token text into the range and `IN` error paths instead of reconstructing it from `GenericValue`?

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1205,7 +1205,8 @@ func (v *ParserVisitor) VisitTerm(ctx *parser.TermContext) interface{} {
 		for i, e := range array {
 			castedValue, err := castValue(dataType, e)
 			if err != nil {
-				return merr.WrapErrParameterInvalidMsg("value '%s' in list cannot be casted to %s", e.String(), dataType.String())
+				return merr.WrapErrParameterInvalidMsg("value %s (%s) in list cannot be casted to field %s (%s)",
+					formatGenericValue(e), getValueDataType(e), ctx.Expr(0).GetText(), dataType.String())
```
xiaofan-luan: `dataType.String()` drops the array element type for a whole-array `IN` target. For example, `ArrayField in [1]` reports `field ArrayField (Array)`, while `ArrayField in [[1, 2]]` is valid; reporting `Array[Int64]` would make the expected value shape clear. Could we use the existing full type formatting (for example `getDataType(childExpr)`) while retaining the element type for indexed/element-level paths, and add an unindexed-array regression case?

internal/parser/planparserv2/utils.go
jsshin8128: Thanks for the comment!

Confirmed and fixed in 3b8e51954. Range and in expressions now consistently pass through the operand source text, matching the existing behavior in the comparison path. This ensures values like 1.0, 1e3, 0x1p+1, and 'two' are reported exactly as they appear in the query.

formatGenericValue is now only used as a fallback for operands that do not have a source token, such as template-injected values. The doc comment has also been updated to remove the incorrect claim that it preserves the original spelling.

Regression tests cover all four cases, including the reverse-range direction.

internal/parser/planparserv2/parser_visitor.go
jsshin8128: Thanks for the comment!

Confirmed and fixed in 3b8e51954. Unindexed array targets now use getDataType(childExpr), so ArrayField in [1] correctly reports the target type as Array[Int64]. Indexed and element-level targets are already narrowed to their element type earlier in the pipeline, so their output remains unchanged. Regression tests have been added for the unindexed ArrayField and StringArrayField cases.

jsshin8128:
Ruff (changed files only) is failing on this PR. The findings are in the Python test files where assertions were updated, but none of them are on lines modified in this PR. Running Ruff 0.15.11 with tests/ruff.toml against the corresponding files from master produces the same 69 findings. ruff format --check would also reformat four of those files, resulting in a diff of roughly 3,500 lines.

How would you like to handle this? I can open a separate cleanup PR for those files, or include the fixes here if you prefer to keep everything in a single change.

internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1389,6 +1397,39 @@ func (v *ParserVisitor) getColumnInfoFromStructIndexField(identifier string) (*p
...
+	default:
+		return ""
+	}
+}
+
+// termElementText returns the source text of the i-th element in a term list. It
+// returns an empty string when the list is not a literal, such as when it is
+// produced by constant folding, since there is no source token available to quote.
+func termElementText(list parser.IExprContext, i int) string {
+	arrayCtx, ok := list.(*parser.ArrayContext)
```
xiaofan-luan, line 1422: `termElementText` treats any non-direct `ArrayContext` as if no source token were available, but a parenthesized list is a `ParensContext` that still contains the original literal text. On this head, `Int64Field in ([1.0])` reports `value 1 (Double)`, `Int64Field in (([1e3]))` reports `value 1000 (Double)`, and `Int64Field in (['two'])` reports `value "two" (VarChar)`. This leaves the issue's “print values the way the user wrote them” requirement incomplete. Could we unwrap `ParensContext` layers before resolving the `ArrayContext`, and add regression cases for these forms?

internal/parser/planparserv2/utils.go
```diff
@@ -555,11 +555,65 @@ func getDataType(expr *ExprWithType) string {
...
+		return text
+	}
+	return formatGenericValue(value)
+}
+
+// getValueDataType returns the parsed type of a literal so type-mismatch errors
+// can include the types of both operands.
+func getValueDataType(value *planpb.GenericValue) string {
+	if valueExpr := toValueExpr(value); valueExpr != nil {
+		return valueExpr.dataType.String()
```
xiaofan-luan, line 605: `getValueDataType` returns `valueExpr.dataType.String()`, so an array-valued mismatch loses its element type even though `GenericValue.ArrayVal` carries it. For example, `Int64Field in [[1, 2]]` reports `value [1,2] (Array)` instead of `Array[Int64]`, making it impossible to distinguish that value from `Array[VarChar]`. This is the same ambiguity the PR now fixes for the target field type. Could we use the existing full type formatting (`getDataType(valueExpr)`) here and add a nested-array mismatch regression case?

xiaofan-luan (commented):
I found two remaining edge cases in the new error-message formatting: parenthesized IN lists still lose literal spelling, and array-valued mismatches still omit the element type.

xiaofan-luan:
I fix the ruff problem.
please rebase with master and the ruff problem should be fixed @jsshin8128

xiaofan-luan:
thanks for the patience

Issue #51905, opened 2026-07-27
[Bug]: master does not compile: BinaryArithRangeJSONCompareNotEqual is undefined


czs007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- Milvus version: master @ 997e8eefef
- Deployment mode(standalone or cluster): N/A (build failure)
- MQ type(rocksmq, pulsar or kafka): N/A
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): Ubuntu 22.04, gcc 11
- CPU/Memory: N/A
- GPU: N/A
- Others: -
```

### Current Behavior

`master` does not compile. `make build-cpp` fails in segcore:

```
internal/core/src/exec/expression/BinaryArithOpEvalRangeExpr.cpp:401:38:
  error: 'json_v' was not declared in this scope; did you mean 'bson_t'?
internal/core/src/exec/expression/BinaryArithOpEvalRangeExpr.cpp:406:38:
  error: 'json_v' was not declared in this scope; did you mean 'bson_t'?
internal/core/src/exec/expression/BinaryArithOpEvalRangeExpr.cpp:400:60:
  error: 'BinaryArithRangeJSONCompareNotEqual' was not declared in this scope;
         did you mean 'BinaryArithRangeJSONCompare'?
```

The Shl/Shr cases that 997e8eefef (#51051) added to the JSON **NotEqual** branch call
`BinaryArithRangeJSONCompareNotEqual`:

```cpp
case proto::plan::ArithOpType::Shl: {
    BinaryArithRangeJSONCompareNotEqual(
        (int64_t(json_v) << int64_t(right_operand)) != val);
    break;
}
```

That macro does not exist. `BinaryArithOpEvalRangeExpr.cpp` defines only
`BinaryArithRangeJSONCompare` (L201), `BinaryArithRangeJONCompareArrayLength` (L243),
`BinaryArithRangeArrayCompare` (L763) and `BinaryArithRangeArrayLengthCompate` (L785), and
nothing in the tree produces the name by token pasting. `json_v` is declared inside the
expansion of `BinaryArithRangeJSONCompare`, so referencing it outside that macro is what
produces the first two errors.

Every other case in the same switch — Div, Mod, ArrayLength, BitAnd, BitOr, BitXor —
already passes its `!= val` comparison to `BinaryArithRangeJSONCompare`, so the two shift
cases are simply calling the wrong name.

[...]

PR #51051, opened 2026-07-04
feat: implement shift (<<, >>) and bitwise NOT (~) in filter expressi…

AyushKashyapII:
## Summary

Closes #50964
Follows #50666 (which added `&`, `|`, `^`)

Completes the bitwise operator family in Milvus filter expressions by adding
shift left (`<<`), shift right (`>>`), and bitwise NOT (`~`).

**`~` is not a new execution path** — it is rewritten at parse time to
`x ^ -1`, an exact identity in two's-complement arithmetic. This reuses the
existing `BitXor` executor path on every field type with no new proto value
and no new C++ executor code.

## Changes

| Layer | What changed |
|---|---|
| `pkg/proto/plan.proto` | Added `Shl = 10`, `Shr = 11` to `ArithOpType` enum |
| `pkg/proto/planpb/plan.pb.go` | Regenerated with protoc v27.0 (not hand-edited) |
| `internal/parser/planparserv2/` | `VisitShift` wired; `VisitUnary` BNOT arm added; `ShiftLeft`, `ShiftRight`, `BitNot` constant-folding; shift-amount guard `[0, 64)` enforced at parse time |
| `internal/core/src/exec/expression/` | `ArithOpHelper<Shl/Shr>` specializations; dispatch arms across all 4 execution paths (scalar, index, JSON, array) |
| `internal/core/src/bitset/` | `Shl`/`Shr` added to enum, `ArithCompareOperator`, runtime dispatch in `bitset.h`, all 5 platform instantiation files, and early-exit SIMD fallback guards in all 4 SIMD impl headers |
| `internal/core/src/expr/ITypeExpr.h` | `Shl`/`Shr` cases in `fmt::formatter` |

## Behavior

- `(flags << 1) == 4` ✅
- `(flags >> 2) == 1` ✅  
- `~flags == 0` ✅ (rewritten to `flags ^ -1 == 0`)
- `(FloatField << 1) == 4` ❌ parse error — integer types only
- `(flags << 64) == 0` ❌ parse error — shift amount must be in `[0, 64)`
- `(flags >> 2) & 1 == 1` ❌ nested arithmetic — pre-existing executor limitation shared by all arithmetic operators, documented as non-goal

## Non-goals

- SIMD-accelerated shift evaluation (scalar fallback used, same as `&`/`|`/`^`)
- Nested arithmetic in a single predicate — requires a new executor node type, separate follow-up
- A first-class `BitNot` plan operator — the `x ^ -1` rewrite is exact and needs no new executor code

[...]

PR #51906, opened 2026-07-27
fix: use the defined JSON compare macro for shift ops in the NotEqual branch

czs007:
issue: #51905

`master` currently does not compile. The Shl/Shr cases that #51051 added to the JSON
**NotEqual** branch call `BinaryArithRangeJSONCompareNotEqual`, which is not defined
anywhere — `BinaryArithOpEvalRangeExpr.cpp` only defines `BinaryArithRangeJSONCompare`
(L201), `BinaryArithRangeJONCompareArrayLength` (L243), `BinaryArithRangeArrayCompare`
(L763) and `BinaryArithRangeArrayLengthCompate` (L785), and nothing produces the name by
token pasting. `json_v` is declared inside the expansion of the real macro, hence:

```
BinaryArithOpEvalRangeExpr.cpp:401:38: error: 'json_v' was not declared in this scope
BinaryArithOpEvalRangeExpr.cpp:400:60: error: 'BinaryArithRangeJSONCompareNotEqual' was not
                                             declared in this scope
```

Every other case in the same switch — Div, Mod, ArrayLength, BitAnd, BitOr, BitXor —
already passes its `!= val` comparison to `BinaryArithRangeJSONCompare`, so the two shift
cases just use the same macro. No semantics change.

Also adds the two test cases the branch was missing. The JSON shift cases added by #51051
cover `<`, `>=` and `~json["int"] < 0`; `~` is rewritten to `(x ^ -1)` and dispatches
through the BitXor case, so nothing exercised the JSON NotEqual branch with an explicit
shift — which is why the broken lines got through.

### Verification

- Before: `make build-cpp` fails in segcore. After: builds clean.
- `*TestBinaryArithOpEvalRange*` — 540/540 passed.
- To confirm the new cases actually reach the changed lines, one reference lambda was
  temporarily inverted; the suite then failed with
  `(json["int"] << 1) != 2@0!!{"int":1622650073,...}` and passed again once restored.

### DIFF
--- a/internal/core/src/exec/expression/BinaryArithOpEvalRangeExpr.cpp
+++ b/internal/core/src/exec/expression/BinaryArithOpEvalRangeExpr.cpp
@@ -397,12 +397,12 @@ PhyBinaryArithOpEvalRangeExpr::ExecRangeVisitorImplForJson(
                         break;
                     }
                     case proto::plan::ArithOpType::Shl: {
-                        BinaryArithRangeJSONCompareNotEqual(
+                        BinaryArithRangeJSONCompare(
                             (int64_t(json_v) << int64_t(right_operand)) != val);
                         break;
                     }
                     case proto::plan::ArithOpType::Shr: {
-                        BinaryArithRangeJSONCompareNotEqual(
+                        BinaryArithRangeJSONCompare(
                             (int64_t(json_v) >> int64_t(right_operand)) != val);
                         break;
                     }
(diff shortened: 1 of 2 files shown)


czs007:
Superseded by #51904, which landed as 4ffe2f3caf about an hour before I opened this — same two-line macro replacement and equivalent regression cases. My mistake for not re-checking upstream before pushing. Closing.

Issue #51902, opened 2026-07-27
[Bug]: disk write worker pool swallows write failures and drops queued writes on reconfig


czs007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (997e8eefef), also present on 2.6.x / 3.0.x
- Deployment mode(standalone or cluster): both (QueryNode only)
- MQ type(rocksmq, pulsar or kafka): N/A
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): Linux
- CPU/Memory: N/A
- GPU: N/A
- Others: requires `common.diskWriteNumThreads` > 0 (non-default)
```

### Current Behavior

Found by code inspection while reading the C++ storage layer; not observed in a running
cluster yet.

**Scope first, so this is not mis-triaged: it only affects deployments that explicitly set
`common.diskWriteNumThreads` to a non-zero value.** The default is `0`
(`configs/milvus.yaml`, `pkg/util/paramtable/component_param.go`), which makes
`FileWriteWorkerPool` keep `executor_ == nullptr`, so `FileWriter::use_writer_pool_` is
`false` and every write goes down the direct `WriteInternal()` path. Default clusters are
not affected. The parameter exists to cap disk-write concurrency so index writes do not
starve reads, so disk-sensitive deployments do enable it.

With the pool enabled there are two defects. The second one is silent only because of the
first.

#### 1. `Future::wait()` does not rethrow, so write failures are discarded

`internal/core/src/storage/FileWriter.cpp` — `FileWriter::Write()` (L300) and
`FileWriter::Finish()` (L365):

```cpp
auto task = [this, data, nbyte, promise]() {
    try {
        WriteInternal(data, nbyte);
        promise->setValue(folly::Unit{});
    } catch (...) {
        promise->setException(folly::exception_wrapper(std::current_exception()));
    }
};

if (FileWriteWorkerPool::GetInstance().AddTask(task)) {
    try {
        future.wait();                                  // <-- only blocks
    } catch (const std::exception& e) {                 // <-- unreachable
        Cleanup();
        ThrowInfo(ErrorCode::FileWriteFailed, ...);
    }
}
```

Per folly's contract (`folly/futures/Future.h`, `Future<T>& wait() &`), the only
postcondition is `isReady() == true`. It does not return the value and does not rethrow the
[...]

PR #51908, opened 2026-07-27
fix: surface disk write failures and stop dropping queued writes on pool reconfig

czs007:
issue: #51902

Defects in the optional disk-write worker pool, plus the guards that turn the resulting
corruption into an error. All of this is behind `common.diskWriteNumThreads > 0`; the
default `0` bypasses the pool entirely and is unaffected.

### 1. `Future::wait()` never rethrows, so failed writes were reported as success

`FileWriter::Write()` / `Finish()` handed the write to the pool and then called
`future.wait()` inside a `try`/`catch`. folly's `Future<T>& wait() &` only blocks — its
only postcondition is `isReady() == true` — so the `catch` was dead code and a failed write
returned to the caller as a success. The result is now taken out with `getTry()` and
`ErrorCode::FileWriteFailed` is raised on the calling thread.

A broken promise is handled separately: it means the task was destroyed before it ran, so
nothing was written and the writer is untouched, and the write is redone inline. For that
to be reachable the queued task must be the promise's only owner, so the closure captures
the promise by move and is itself moved into `AddTask()` — otherwise the caller-side owners
keep the Promise alive, `BrokenPromise` is never published, and the wait hangs.

### 2. Reconfiguring the pool discarded queued writes

`SetWorker()` replaced the executor on every call. `stop()` makes each worker `return` as
soon as its current task ends (`threadsToStop_ > 0 && !isJoin_` in
`CPUThreadPoolExecutor::threadRun()`), so everything still queued was destroyed with the
executor, breaking those tasks' promises — which, via defect 1, the caller saw as success.

Merely draining instead of discarding is not enough: while the outgoing executor works
through its backlog the replacement is already accepting writes, so a same-size refresh
runs up to `2 * nr_worker` writes at once, breaking the cap this setting exists to enforce.
`SetWorker()` therefore no longer replaces the executor at all:

- unchanged count → return early
- `n → m`, both > 0 → `setNumThreads(m)`, resized in place
- `0 → n` → create; `n → 0` → `join()` (drains) then release

The resize happens under `executor_mutex_`, which also gates `AddTask()` for its duration,
so two live pools never coexist. Shrinking via `setNumThreads()` does not repeat the
original bug: it retires only the surplus threads while the executor and its queue survive.

### 3. A failed writer was left booby-trapped

`Cleanup()` closes `fd_` and frees `aligned_buf_` but leaves `capacity_`/`offset_` intact,
and `PositionedWriteWithCheck()` calls it *before* rethrowing. A caller that swallowed the
error and kept going therefore hit a null dereference in `Write()`'s fast path or in the
`memset()` inside `FlushWithDirectIO()`, and with an empty buffer `Finish()` skipped the
flush and returned the stale `file_size_` as a successful finalization. Both entry points
now check the file is still open, like `PositionedFileWriter::WriteAt()` already does.

Also takes `executor_mutex_` in `HasPool()`, which previously read the `shared_ptr`
concurrently with `SetWorker()`'s write.

## Verification

Five regression tests; each was confirmed to fail against the code without the
corresponding fix:

| test | without the fix | with it |
|---|---|---|
| `PooledWriteFailureIsReportedToCaller` | `Actual: it throws nothing` | throws `FileWriteFailed` |
| `ExecutorReconfigDoesNotDropQueuedTasks` | timeout at 1 of 9 tasks | 9/9 run |
| `ExecutorShutdownDrainsQueuedTasks` | 1 of 9 tasks | 9/9 run |
| `WriteAfterFailedWriteIsRejected` | killed by `SIGSEGV` (shown with `EXPECT_EXIT(..., KilledBySignal(SIGSEGV))`) | throws |
| `FinishAfterFailedWriteIsRejected` | `Actual: it throws nothing` | throws |

Failures are fault-injected against `/dev/full`, which is seekable and fails every write
with `ENOSPC`, so the error travels the real path: `PWriteAll` → promise → `getTry()` →
rethrow on the calling thread.

Full C++ suite: 8030 tests, **all passing except 7 `AzureChunkManagerTest` cases that fail
because the local azurite container had exited mid-run** (`Request was cancelled by
context`); after restarting it that suite passes 9/9. `internal/core/src/storage/azure/`
does not reference `FileWriter`.

Also exercised on a running standalone with `common.diskWriteNumThreads: 1`:

- QueryNode does create the pool — `[CGO] [FileWriter.h] Set the number of write worker to 1`.
- Before this PR, changing **one unrelated** key (`common.diskWriteRateLimiter.avgKBps`)
  fired `SetWorker()` **twice**, since QueryNode registers nine `common.diskWrite*` keys
  against the same handler. With the early return those reconfigurations are now no-ops.
- 20k×128 HNSW, five release/load cycles while flipping `avgKBps` every 50 ms: searches
  returned correct results, no write errors, no panics.

## Not verified

- The promise-ownership change is reasoned, not covered by a test: after the `SetWorker()`
  rework no path discards a queued task except the pool destructor.
- I did not instrument whether individual index-file writes exceed the 64 KB buffer, so I
  cannot claim the standalone run above actually raced a reconfiguration against a queued
  write. That mechanism is pinned by the unit tests instead.
- I did not reproduce end-to-end data loss on a pre-fix binary.
- `~FileWriteWorkerPool()` still uses `stop(); join();`, left alone deliberately — at static
  destruction a fast shutdown is reasonable and no caller can consume the result.

### DIFF
--- a/internal/core/src/storage/FileWriter.cpp
+++ b/internal/core/src/storage/FileWriter.cpp
@@ -267,6 +267,10 @@ FileWriter::WriteInternal(const void* data, size_t nbyte) {
 
 void
 FileWriter::Write(const void* data, size_t nbyte) {
+    // a failed write closes the file and frees the aligned buffer, so refuse to
+    // keep going instead of memcpy'ing into the freed buffer below
+    AssertInfo(fd_ != -1, "File is not open: {}", filename_);
+
     // if the data can fit in the aligned buffer, we can just copy it to the aligned buffer
     if (nbyte <= capacity_ - offset_) {
         const char* src = static_cast<const char*>(data);
@@ -283,7 +287,10 @@ FileWriter::Write(const void* data, size_t nbyte) {
 
     auto promise = std::make_shared<folly::Promise<folly::Unit>>();
     auto future = promise->getFuture();
-    auto task = [this, data, nbyte, promise]() {
+    // the queued task has to be the only owner of the promise: if it is
+    // discarded before it runs, destroying it is what publishes BrokenPromise
+    // and wakes the wait below, so no owner may be left on this side
+    auto task = [this, data, nbyte, promise = std::move(promise)]() {
         try {
             WriteInternal(data, nbyte);
             promise->setValue(folly::Unit{});
@@ -295,15 +302,24 @@ FileWriter::Write(const void* data, size_t nbyte) {
 
     // try to add the task to the writer pool
     // fallback to write the data directly if the task cannot be added to the pool
-    if (FileWriteWorkerPool::GetInstance().AddTask(task)) {
-        try {
-            future.wait();
-        } catch (const std::exception& e) {
-            Cleanup();
-            ThrowInfo(ErrorCode::FileWriteFailed,
-                      "Failed to write to file: {}, error: {}",
-                      filename_,
-                      e.what());
+    if (FileWriteWorkerPool::GetInstance().AddTask(std::move(task))) {
+        // folly's Future::wait() only blocks until the future is ready, it does
+        // not rethrow the stored exception, so the result has to be taken out
+        // explicitly. Otherwise a failed write is reported as a success.
+        auto result = std::move(future).getTry();
+        if (result.hasException()) {
+            if (result.exception().is_compatible_with<folly::BrokenPromise>()) {
+                // the task was destroyed before it ran, so nothing was written
+                // and the writer is untouched: do the write on this thread
+                WriteInternal(data, nbyte);
+            } else {
+                auto reason = result.exception().what().toStdString();
+                Cleanup();
+                ThrowInfo(ErrorCode::FileWriteFailed,
+                          "Failed to write to file: {}, error: {}",
+                          filename_,
+                          reason);
+            }
         }
     } else {
         WriteInternal(data, nbyte);
@@ -342,11 +358,16 @@ FileWriter::FlushWithBufferedIO() {
 
 size_t
 FileWriter::Finish() {
+    // an earlier failure closed the file and freed the aligned buffer while
+    // leaving offset_ alone, so finalizing here would either memset the freed
+    // buffer or report a stale file_size_ as a successful finalization
+    AssertInfo(fd_ != -1, "File is not open: {}", filename_);
+
     // if the aligned buffer is not empty, we should flush it to the file
     if (offset_ != 0) {
         auto promise = std::make_shared<folly::Promise<folly::Unit>>();
         auto future = promise->getFuture();
-        auto task = [this, promise]() {
+        auto task = [this, promise = std::move(promise)]() {
             try {
                 if (use_direct_io_) {
                     FlushWithDirectIO();
@@ -360,15 +381,26 @@ FileWriter::Finish() {
             }
         };
 
-        if (FileWriteWorkerPool::GetInstance().AddTask(task)) {
-            try {
-                future.wait();
-            } catch (const std::exception& e) {
-                Cleanup();
-                ThrowInfo(ErrorCode::FileWriteFailed,
-                          "Failed to flush file: {}, error: {}",
-                          filename_,
-                          e.what());
+        if (FileWriteWorkerPool::GetInstance().AddTask(std::move(task))) {
+            // see the notes in Write(): wait() never rethrows, and a broken
+            // promise means the task never ran
+            auto result = std::move(future).getTry();
+            if (result.hasException()) {
+                if (result.exception()
+                        .is_compatible_with<folly::BrokenPromise>()) {
+                    if (use_direct_io_) {
+                        FlushWithDirectIO();
+                    } else {
+                        FlushWithBufferedIO();
+                    }
+                } else {
+                    auto reason = result.exception().what().toStdString();
+                    Cleanup();
+                    ThrowInfo(ErrorCode::FileWriteFailed,
+                              "Failed to flush file: {}, error: {}",
+                              filename_,
+                              reason);
+                }
             }
         } else {
             if (use_direct_io_) {
--- a/internal/core/src/storage/FileWriter.h
+++ b/internal/core/src/storage/FileWriter.h
@@ -366,22 +366,40 @@ class FileWriteWorkerPool {
                 std::thread::hardware_concurrency());
             nr_worker = std::thread::hardware_concurrency();
         }
-        std::shared_ptr<folly::CPUThreadPoolExecutor> old_executor = nullptr;
+        std::shared_ptr<folly::CPUThreadPoolExecutor> retired_executor = nullptr;
         {
+            // held across the resize so no task is submitted to a pool that is
+            // being taken down, and so two reconfigurations cannot interleave
             std::lock_guard<std::mutex> lock(executor_mutex_);
-            old_executor = executor_;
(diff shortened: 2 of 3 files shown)


internal/core/src/storage/FileWriter.cpp
```diff
@@ -296,14 +300,23 @@ FileWriter::Write(const void* data, size_t nbyte) {
...
-        } catch (const std::exception& e) {
-            Cleanup();
-            ThrowInfo(ErrorCode::FileWriteFailed,
-                      "Failed to write to file: {}, error: {}",
-                      filename_,
-                      e.what());
+        // folly's Future::wait() only blocks until the future is ready, it does
+        // not rethrow the stored exception, so the result has to be taken out
+        // explicitly. Otherwise a failed write is reported as a success.
+        auto result = std::move(future).getTry();
```
buqian-zilliz, line 309: **[P2] Make the queued task the sole Promise owner**

`promise` and the local `task` closure both retain the producer while this `getTry()` blocks, and `AddTask(task)` copies the closure. If the executor destroys its queued copy without running it (the pool destructor still uses `stop()`), the Promise cannot be destroyed to publish `BrokenPromise`; this wait therefore never returns and the fallback below is unreachable. `Finish()` has the same ownership pattern. Move the submitted task into `AddTask()` and release every caller-side Promise owner before blocking.

internal/core/src/storage/FileWriter.cpp
```diff
@@ -361,14 +374,25 @@ FileWriter::Finish() {
...
-                future.wait();
-            } catch (const std::exception& e) {
-                Cleanup();
-                ThrowInfo(ErrorCode::FileWriteFailed,
-                          "Failed to flush file: {}, error: {}",
-                          filename_,
-                          e.what());
+            // see the notes in Write(): wait() never rethrows, and a broken
+            // promise means the task never ran
+            auto result = std::move(future).getTry();
```
buqian-zilliz, line 387: **[P2] Reject Finish on a writer invalidated by an earlier failure**

`PositionedWriteWithCheck()` closes `fd_` and frees `aligned_buf_` on a failed write but leaves `offset_` unchanged. If the caller catches the now-propagated error and calls `Finish()`, there is no invalid-state check before this flush: direct I/O can enter `FlushWithDirectIO()` and `memset()` the freed buffer when `offset_ != 0`, while `offset_ == 0` skips flushing and returns the stale `file_size_` as if finalization succeeded. Check the failed or closed state at the start of `Finish()`; at minimum, require `fd_ != -1`.

internal/core/src/storage/FileWriter.h
```diff
@@ -380,7 +380,10 @@ class FileWriteWorkerPool {
             }
         }
         if (old_executor != nullptr) {
-            old_executor->stop();
+            // join() instead of stop(): stop() makes every worker return as
+            // soon as its current task ends and leaves the queued tasks
+            // unexecuted, which breaks the promises they carry. join() lets the
+            // workers drain the queue before exiting.
```
buqian-zilliz: **[P2] Keep the concurrency cap across executor replacement**

`SetWorker()` publishes the replacement executor and releases `executor_mutex_` before draining `old_executor`. While this drain processes the entire old backlog, `AddTask()` can run new writes on the replacement, so a same-size refresh can execute up to `2 * nr_worker` write tasks for the duration of the backlog. Because all disk-writer configuration updates call this reconfiguration path, routine updates can violate the configured maximum write concurrency and increase disk-read contention. Avoid rebuilding when the count is unchanged, and resize, drain, or gate submissions before exposing a replacement.

internal/core/src/storage/FileWriter.cpp
czs007: Fixed in 94afa2d. You are right that the caller kept the producer alive: `AddTask(std::function<void()>)` takes its argument by value and I passed the lvalue `task`, so while `getTry()` blocked the `shared_ptr<Promise>` was still held by the local `promise`, by the local closure, and by the executor's copy. Dropping the executor's copy took the count 3 -> 2, the Promise was never destroyed, `BrokenPromise` was never published, and the wait would have hung rather than reaching the fallback.

The closure now captures the promise by `std::move` and is itself moved into `AddTask()`, in both `Write()` and `Finish()`, so the queued task is the sole owner.

I could not build a deterministic test for this one: after the `SetWorker()` rework there is no path left that discards a queued task except the pool destructor. So this change is reasoned, not covered by a test.

internal/core/src/storage/FileWriter.cpp
czs007: Fixed in 94afa2d, and this corrected a wrong assumption of mine. I had deliberately guarded only `Write()`, reasoning that `Finish()` on a closed writer would end up in `pwrite(-1, ...)` and throw `EBADF` rather than crash. That is wrong: `FlushWithDirectIO()` does the `memset(aligned_buf_ + offset_, ...)` *before* the write, so with the buffer already freed it is the same null dereference as the fast path.

Your second failure mode turned out to be the one my test actually exercises. In the `/dev/full` scenario the write fails inside the `left_size >= capacity_` loop, before `offset_ = 0` is reached and with `offset_` already 0, so `Finish()` skipped the flush entirely and returned the stale `file_size_` as a successful finalization — silently.

`Finish()` now starts with `AssertInfo(fd_ != -1, ...)`, matching `Write()` and `PositionedFileWriter::WriteAt()`. Added `FinishAfterFailedWriteIsRejected`; without the guard it fails with `Expected: writer.Finish() throws ... Actual: it throws nothing`.

On the double-`Finish()` concern that made me skip this originally: the only non-test caller is `MmapChunkTarget::flush()`, and every caller reaches it through `release()` exactly once, so no existing path calls `Finish()` twice.

internal/core/src/storage/FileWriter.h
czs007: Fixed in 94afa2d, and you are right that this was a regression my own change introduced: replacing `stop()` with `join()` traded dropped tasks for a window in which the old executor drains its backlog while the replacement is already accepting work, so a same-size refresh could run up to `2 * nr_worker` writes — breaking the exact cap this setting exists to enforce. It compounds with something I saw on a live standalone: changing one unrelated key (`common.diskWriteRateLimiter.avgKBps`) fires `SetWorker()` *twice*, because QueryNode registers nine `common.diskWrite*` keys against the same handler.

Took your suggestion in full — `SetWorker()` no longer replaces the executor:

- unchanged count -> return early (so the routine reconfigurations above become no-ops)
- `n -> m`, both > 0 -> `executor_->setNumThreads(m)`, resized in place
- `0 -> n` -> create; `n -> 0` -> `join()` then release

I checked that shrinking via `setNumThreads()` does not repeat the original bug: it calls `removeThreads(n, /*isJoin*/ false)`, but that only retires the surplus threads while the executor and its queue survive, so the remaining workers still drain the backlog. The whole resize is done under `executor_mutex_`, which also gates `AddTask()` for its duration, so there is never a moment with two live pools.

This makes the dropped-task defect structurally impossible rather than fixed by discipline. `ExecutorShutdownDrainsQueuedTasks` covers the one remaining teardown path (`n -> 0`); it fails against the old `stop(); join();`. `ExecutorReconfigDoesNotDropQueuedTasks` now waits for completion instead of assuming `SetWorker()` blocks — it still fails against the old implementation, by timeout at 1 of 9 tasks.

PR #51909, opened 2026-07-27
fix: [2.6] keep proxy metric status label stable, split cause into its own label (#51495)

czs007:
Cherry-pick of #51496
pr: #51495
issue: #51493

### Why this branch needs it

#51496 landed this fix on `hotfix-2.6.20` (shipped as v2.6.20) but **never on `2.6`**. `2.6` still carries the v2.6.19 in-place status split — `metrics.go` has `FailInputLabel` / `FailSystemLabel` / `RejectedUserLabel` / `RejectedSystemLabel`, and `req_count` has no `cause` label at all.

So the next release cut from `2.6` would emit `fail_input` / `fail_system` again: a user upgrading **v2.6.20 -> 2.6.21** would see the dashboards and alert rules that v2.6.20 had just repaired break a second time — and this time as a regression from a working state, which is harder to notice than the original break.

### What

```
```

|---|---|

Outcomes that have no responsible party emit `cause=""`. Prometheus treats an empty label value as equivalent to the label being absent, so those series stay byte-identical to what pre-2.6.19 emitted — the new label materializes only where it carries information.

Old queries work unchanged and still count each request once: `status="fail"` is again every hard failure, because Prometheus aggregates over `cause` for free.

### Cherry-pick notes

`git cherry-pick` of #51496 applied with **no conflicts** (19 files). Cross-checked file by file against `hotfix-2.6.20`: 16 files are byte-identical, and the 3 that differ do so only in `2.6`-only content unrelated to this change (`internal/proxy/util.go` — `validateElementType`; `handler_v2.go` and `util_test.go` — surrounding `2.6` code). The cause-related lines in all three were extracted and compared separately: identical.

The emit surface was **re-verified on this branch rather than assumed**: all 12 `WithLabelValues` sites for the two metrics were checked for label arity by a paren-balanced scan, because adding a label makes arity a runtime panic, not a compile error. (`2.6` has 12 sites where master/3.0 have 62 — it lacks the `InterceptorImpl` / `snapshot_impl.go` surface.) No stale `FailInputLabel` / `FailSystemLabel` / `RejectedUserLabel` / `RejectedSystemLabel` references remain, and all `ParseMetricLabel` / `failMetricLabel` call sites use the two-value signature.

- Green: `internal/proxy`, `internal/distributed/proxy`, `internal/distributed/proxy/httpserver`, `internal/distributed/proxy/client`, `pkg/metrics`, `pkg/util/requestutil`, `pkg/util/merr`.
- `make verifiers`: 0 issues.

### Related

- #51496 — `hotfix-2.6.20` (merged, the version this is taken from)
- #51526 — `3.0`
- #51907 — `master` (aligns `CauseNA` there from `"na"` to `""`)

### DIFF
--- a/deployments/monitor/grafana/milvus-dashboard.json
+++ b/deployments/monitor/grafana/milvus-dashboard.json
@@ -4794,7 +4794,7 @@
                 "uid": "${datasource}"
               },
               "exemplar": true,
-              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=~\"fail_.*\"}[2m])/120) by(function_name, pod, node_id)",
+              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=\"fail\"}[2m])/120) by(function_name, pod, node_id)",
               "interval": "",
               "legendFormat": "{{function_name}}-{{pod}}-{{node_id}}",
               "queryType": "randomWalk",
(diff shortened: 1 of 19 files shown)


liliu-z (commented):
/lgtm
/approce

liliu-z (commented):
/lgtm
/approve

aa0901292300-bit:
- [ ] - *

PR #51375, opened 2026-07-14
feat: add FMINDEX scalar index for exact LIKE prefix/infix/suffix on VARCHAR

xiaofan-luan:
issue: #51577

## What this PR does

Adds a new **FMINDEX** scalar index for **VARCHAR**, backed by a vendored
byte-exact FM-index (BWT + quad wavelet matrix + sampled suffix array; libsais
for suffix-array construction). It answers the three anchored `LIKE` forms
**exactly, with no candidate recheck**:

- `col LIKE 'p%'`  → PrefixMatch
- `col LIKE '%p'`  → PostfixMatch
- `col LIKE '%p%'` → InnerMatch

Backward search is O(|pattern|); located hits are genuine in-document matches (an
out-of-alphabet separator per row means no match spans two rows).

## Scope (v1)

**Delivered:** VARCHAR columns + the three anchored `LIKE` forms, guarded by a
count-first cost model.

**Declined / deferred (fall back to the existing path, always exact):**
- The equality family `==` / `!=` / `IN` / `NOT IN` (incl. wildcard-free
  `LIKE 'abc'`, which the planner lowers to `==`) is **declined** by
  `ShouldUseOp` and routed to the scan / an equality index — FMINDEX is a
  substring index, not an equality index. `In()`/`NotIn()` remain required
  `ScalarIndex` overrides but `ThrowInfo(Unsupported)` since they are never
  routed.
- General `LIKE` (interior `%`/`_`), `RegexMatch`, and lexicographic range —
  brute-force scan (two-phase acceleration is a follow-up).
- `JSON` / `TEXT` / `ARRAY` data types — rejected at index creation.

## Design doc

`docs/design-docs/design_docs/20260708-fm_index_scalar_index.md` (updated to this
scope; includes an honest capacity model).

## Highlights

- Vendored `internal/core/src/index/fmindex/*` + `internal/core/thirdparty/libsais`
  (Apache-2.0), CMake wiring. Core library: `xiaofan-luan/fm-index-lite`.
- `FMIndex : ScalarIndex<std::string>`: Build/Upload/Load via the V3 packed-file
  path; zero-copy **mmap** load (LoadView) with a move-based Deserialize
  fallback; `ShouldUseOp` routing gate.
- **Count-first cost guard** `queryNode.fmindexCostRatio` (default 0.001), wired
  paramtable → CGO (`SegcoreSetFMIndexCostRatio`) → `SegcoreConfig` →
[...]

PR #51910, opened 2026-07-27
fix: classify NGRAM as a scalar index

xiaofan-luan:
## What changed

- Classify `NGRAM` as a scalar index in `IsScalarIndexType`.
- Preserve the stored NGRAM AutoIndex parameters when the resolved scalar AutoIndex configuration changes.
- Report `CurrentScalarIndexVersion`, rather than the vector `CurrentIndexVersion`, for NGRAM segment-index state.

## Why

NGRAM is implemented by the scalar-index engine, but its index type was omitted from `IsScalarIndexType` when NGRAM support was introduced. The helper's two DataCoord call sites therefore treated NGRAM as a vector index for version reporting and skipped scalar AutoIndex compatibility handling.

This was discovered while reviewing #51375 and is split out so the behavior change can be reviewed and tested independently from FMINDEX.

## Tests

- Added direct coverage for `IsScalarIndexType(IndexNGRAM)`.
- Added DataCoord coverage for NGRAM AutoIndex parameter preservation.
- Added DataCoord coverage verifying NGRAM reports `CurrentScalarIndexVersion`.
- `go test -c -tags dynamic,test` succeeds for `./internal/util/indexparamcheck` and `./internal/datacoord`.

Targeted binaries could not be executed locally because this checkout does not have the required `libmilvus_core.dylib` / `libmilvus-storage.dylib`; CI will run the tests in the complete build environment.

### DIFF
--- a/internal/datacoord/index_meta_test.go
+++ b/internal/datacoord/index_meta_test.go
@@ -250,6 +250,38 @@ func TestMeta_ScalarAutoIndex(t *testing.T) {
 		assert.Equal(t, newIndexParams[0].Key, common.IndexTypeKey)
 		assert.Equal(t, newIndexParams[0].Value, "INVERTED")
 	})
+
+	t.Run("ngram index param rewrite", func(t *testing.T) {
+		m.indexes[collID] = map[UniqueID]*model.Index{
+			indexID: {
+				CollectionID: collID,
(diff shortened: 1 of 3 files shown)


czs007 (approved):
/lgtm

PR #51913, opened 2026-07-28
fix: restore growing index raw data ownership

chyezh:
issue: #51916

- growing segment writes: restore vector interim-index raw-data ownership and stop repopulating reclaimed chunks after synchronization
- field loading: apply the same vector-only ownership rule while preserving nullable validity metadata
- index append failures: document the invalid recreate path after the only complete raw-data source has been reclaimed

### DIFF
--- a/internal/core/src/segcore/FieldIndexing.cpp
+++ b/internal/core/src/segcore/FieldIndexing.cpp
@@ -361,6 +361,13 @@ VectorFieldIndexing::AppendSegmentIndexSparse(int64_t reserved_offset,
             sync_with_index_.store(true);
         } catch (SegcoreError& error) {
             LOG_ERROR("growing sparse index add error: {}", error.what());
+            // Known design defect: after a raw-data-owning interim index has
+            // synchronized, its source chunks may have been reclaimed and
+            // field_raw_data is no longer a complete recovery source. This
+            // catch is reachable when AddWithDataset fails, but recreating the
+            // index here is not a valid recovery path and must not exist under
+            // single-owner semantics. The failure should instead be propagated
+            // and the segment marked unrecoverable.
             recreate_index(get_data_type(), field_raw_data);
         }
     } else {
@@ -387,6 +394,14 @@ VectorFieldIndexing::AppendSegmentIndexSparse(int64_t reserved_offset,
                 index_->AddWithDataset(dataset, conf);
             } catch (SegcoreError& error) {
                 LOG_ERROR("growing sparse index add error: {}", error.what());
+                // Known design defect: after a raw-data-owning interim index
+                // has synchronized, its source chunks may have been reclaimed
+                // and field_raw_data is no longer a complete recovery source.
+                // This catch is reachable when AddWithDataset fails, but
+                // recreating the index here is not a valid recovery path and
+                // must not exist under single-owner semantics. The failure
+                // should instead be propagated and the segment marked
+                // unrecoverable.
                 recreate_index(get_data_type(), field_raw_data);
             }
         }
@@ -497,6 +512,13 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
             sync_with_index_.store(true);
         } catch (SegcoreError& error) {
             LOG_ERROR("growing index add error: {}", error.what());
+            // Known design defect: after a raw-data-owning interim index has
+            // synchronized, its source chunks may have been reclaimed and
+            // field_raw_data is no longer a complete recovery source. This
+            // catch is reachable when AddWithDataset fails, but recreating the
+            // index here is not a valid recovery path and must not exist under
+            // single-owner semantics. The failure should instead be propagated
+            // and the segment marked unrecoverable.
             recreate_index(get_data_type(), field_raw_data);
         }
     } else {
@@ -525,6 +547,14 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
                 index_->AddWithDataset(dataset, conf);
             } catch (SegcoreError& error) {
                 LOG_ERROR("growing index add error: {}", error.what());
+                // Known design defect: after a raw-data-owning interim index
+                // has synchronized, its source chunks may have been reclaimed
+                // and field_raw_data is no longer a complete recovery source.
+                // This catch is reachable when AddWithDataset fails, but
+                // recreating the index here is not a valid recovery path and
+                // must not exist under single-owner semantics. The failure
+                // should instead be propagated and the segment marked
+                // unrecoverable.
                 recreate_index(get_data_type(), field_raw_data);
             }
         }
--- a/internal/core/src/segcore/SegmentGrowingImpl.cpp
+++ b/internal/core/src/segcore/SegmentGrowingImpl.cpp
@@ -715,10 +715,14 @@ SegmentGrowingImpl::Insert(int64_t reserved_offset,
                 &insert_record_proto->fields_data(data_offset),
                 field_meta);
         }
-        // A synchronized interim index that owns raw data consumes later
-        // batches directly. Do not repopulate chunks reclaimed by
-        // try_remove_chunks.
-        if (!indexing_record_.HasRawData(field_id)) {
+        // A synchronized vector interim index that owns raw data consumes
+        // later batches directly. Do not repopulate chunks reclaimed by
+        // try_remove_chunks. Scalar indexes do not take over raw-data
+        // ownership from ConcurrentVector.
+        const bool index_owns_raw_data =
+            IsVectorDataType(field_meta.get_data_type()) &&
+            indexing_record_.HasRawData(field_id);
+        if (!index_owns_raw_data) {
             if (field_meta.get_data_type() == DataType::TEXT) {
                 auto spillover = GetTextLobSpillover(field_id);
                 AssertInfo(spillover != nullptr,
@@ -981,9 +985,13 @@ SegmentGrowingImpl::load_field_data_common(
     if (insert_record_.is_valid_data_exist(field_id)) {
         insert_record_.get_valid_data(field_id)->set_data_raw(field_data);
     }
-    // Keep the load path aligned with Insert: once the interim index owns raw
-    // data, append the loaded batch to the index without rebuilding raw chunks.
-    if (!indexing_record_.HasRawData(field_id)) {
+    // Keep the load path aligned with Insert: once a vector interim index owns
+    // raw data, append the loaded batch to the index without rebuilding raw
+    // chunks. Scalar indexes do not take over raw-data ownership.
+    const bool index_owns_raw_data =
+        IsVectorDataType(field_meta.get_data_type()) &&
+        indexing_record_.HasRawData(field_id);
+    if (!index_owns_raw_data) {
         insert_record_.get_data_base(field_id)->set_data_raw(reserved_offset,
                                                              field_data);
     }
--- a/internal/core/src/segcore/SegmentGrowingImpl.cpp
+++ b/internal/core/src/segcore/SegmentGrowingImpl.cpp
@@ -715,36 +715,38 @@ SegmentGrowingImpl::Insert(int64_t reserved_offset,
                 &insert_record_proto->fields_data(data_offset),
                 field_meta);
         }
-        // Growing-source flush reads raw data from insert_record_. Keep it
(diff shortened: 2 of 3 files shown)


czs007:

Reviewed at head 99dc161c28 (full diff vs merge base, plus the surrounding ownership machinery: `IndexingRecord`/`FieldIndexing`, `try_remove_chunks`, `bulk_subscript`, and the growing-source flush in `segment_c.cpp`). Overall the change is correct on the main paths and mergeable: the flush path already prefers the interim index for vector fields (`can_read_from_index` → `bulk_subscript` → `GetDataFromIndex`, with an explicit payload-size cross-check in `BuildFixedWidthVectorArrayFromBulkSubscript`), and nullable validity lives in the shared `ThreadSafeValidData`, which `Insert`/`load_field_data_common` still populate unconditionally. Findings below are non-blocking hardening and test-coverage items.

**Medium — the raw-write skip is not gated on vector type; correctness for GEOMETRY/TEXT rests on a distant constant.**
`internal/core/src/segcore/SegmentGrowingImpl.cpp:721` (Insert) and `:986` (load) skip the chunk write whenever `indexing_record_.HasRawData(field_id)` is true, for any field type. Today this is only reachable for float/f16/bf16/sparse vectors, but `field_indexings_` also holds GEOMETRY entries (`IndexingRecord::Initialize`, FieldIndexing.h:417-437), whose `sync_data_with_index()` does return true once the RTree is built (FieldIndexing.h:202-215, set in `process_geometry_data`). The only thing preventing the skip from firing for GEOMETRY is `RTreeIndex::HasRawData()` returning a hard-coded `false` (RTreeIndex.h:145-147). If that ever changes (or a scalar interim index gains raw data), Insert silently stops writing geometry/TEXT chunks — but `try_remove_chunks` only reclaims `IsVectorDataType` fields, geometry/TEXT readers and the flush builders still read `view_element` from chunks, and `IndexingRecord::GetDataFromIndex` silently no-ops for non-dense/sparse types — i.e., silent data loss with no assertion tripping. The TEXT LOB spillover write now also sits behind this condition. Suggested fix: make the invariant local, e.g. `if (!(IsVectorDataType(field_meta.get_data_type()) && indexing_record_.HasRawData(field_id)))`, or `AssertInfo(IsVectorDataType(...))` when the skip triggers, in both Insert and `load_field_data_common`.

**Medium — after ownership transfer, the index-add error path has no recovery source and bricks reads of the field.**
`AppendSegmentIndexDense`/`Sparse` catch `SegcoreError` from `AddWithDataset` and call `recreate_index` (FieldIndexing.cpp:477, 500, 528), which resets `built_`/`sync_with_index_` to false and discards the index. Once this PR's skip is active, raw chunks are deliberately and permanently empty, so after such a failure the data exists nowhere: the next build attempt reads `field_raw_data->get_chunk_data(0)` from an empty chunk vector, and `SegmentGrowingImpl::HasRawData` (SegmentGrowingImpl.h:583-600) now returns false with `num_chunk()==0`, so every `bulk_subscript` hits `AssertInfo(HasRawData(...), "Growing segment loss raw data")`. To be fair, this path was already broken before the PR (chunks were cleared each insert, so `recreate_index` rebuilt from partial/misaligned chunks — silent corruption); the PR converts silent corruption into deterministic failure, which is arguably better. Still, worth an explicit decision: either treat add-failure-after-ownership as fatal for the segment with a clear error, or defer the skip until the append for the batch has succeeded. At minimum a comment noting that `recreate_index` cannot recover once ownership has transferred.

**Medium — regression test verifies sizes but never content, and the load-path claim is untested.**
`internal/core/src/segcore/SegmentGrowingIndexTest.cpp:441-452`: the final assertions (`valid_data_size == row_count`, `data_size == valid_count * dim`) are self-consistent even if `GetDataFromIndex` returns mis-mapped or garbage vectors — validity comes from the always-maintained `ThreadSafeValidData`, so the test cannot detect wrong vector content for the second batch, which is exactly the data that now only lives in the index. Suggest comparing the returned floats against `dataset.raw_` for offsets [100, 200) (respecting the validity bitmap), and asserting `valid_count > 0` so the check cannot pass vacuously. Separately, the PR body's second bullet ("field loading: preserve nullable validity metadata while appending later batches through the interim index") has no test — `load_field_data_common` with a synced index (skip active, validity preserved) is never exercised.

**Low — unsynchronized window between the `HasRawData` check and the chunk write.**
The check at SegmentGrowingImpl.cpp:721 and the subsequent `set_data_raw` take no lock, while `try_remove_chunks` (line 402) clears under a `chunk_mutex_.try_lock()`. With concurrent inserts (the `PreInsert`/`ConcurrentVector` design permits them), a batch can pass the check just before another thread's append flips sync, then write chunks concurrently with — or after — a `clear()`, leaving a stale partial chunk and a raw offset mapping that no longer matches. Readers are unaffected (they prefer the index) and the next insert re-clears, so this is benign in steady state and the same interleaving existed pre-PR with the populate-then-clear pattern; flagging mainly so the ownership comment can state that stale chunks after a lost race are tolerated by design.

**Nit —** after ownership, the raw vector's `get_offset_mapping().GetTotalCount()`/`get_valid_count()` freeze (the new test relies on this). The only remaining consumer is the `valid_count >= build_threshold` gate in `IndexingRecord::AppendingIndex` (FieldIndexing.cpp:74-76, 136-138), which stays safe only because ownership implies the threshold was already crossed. A short comment there would prevent future consumers from treating the frozen mapping as a row count.

No critical/high issues found; the ownership model is sound given that the flush path reads index-owned vectors and validity is tracked independently of the reclaimed chunks.

czs007:

Reviewed at updated head cb0f579 (full diff vs master, both commits, plus the ownership machinery: `IndexingRecord::HasRawData`, `try_remove_chunks`, `AppendSegmentIndexDense/Sparse`, `recreate_index`, `bulk_subscript`, `ThreadSafeValidData`, `FieldDataImpl::FillFieldData`, and `DataGen`). The update addresses the earlier round well: the guard is now vector-only (correctly excluding GEOMETRY entries that also register in `field_indexings_`), the load path preserves nullable validity, and the invalid-recreate comments are placed on exactly the right catches (the build-failure catches at `FieldIndexing.cpp:339`/`:492`, where raw chunks are still complete, are correctly left unannotated). Direction is mergeable; findings below are one test-correctness issue worth fixing before merge and documented/pre-existing items.

**High (pre-existing; this PR documents it — recommend a minimal real fix rather than comment-only):** `FieldIndexing.cpp:522`, `:558` (dense) and `:371`, `:405` (sparse) — when `AddWithDataset` fails after synchronization, `recreate_index` swaps in an **empty** index while `built_`, `sync_with_index_`, and `index_cur_` all keep their old values and the raw chunks have already been reclaimed. `HasRawData()` therefore remains true, so every subsequent read (search, `bulk_subscript`, growing-source flush) is served from an index missing all rows, and the new Insert/load guards keep skipping raw writes based on that same stale `HasRawData()` — one failed append silently loses the entire segment with no error surfaced. Since single-owner semantics now make this the *only* copy, comment-only is thin protection. Suggested minimal hardening in this PR: in the post-sync catches, do not call `recreate_index`; instead set an atomic `unrecoverable_` flag (checked in `HasRawData()`/query paths, or simply rethrow so `Insert` fails loudly). That is a few-line change and converts silent wrong results into a diagnosable error.

**Medium (new test, should fix before merge):** `SegmentGrowingIndexTest.cpp:437` — `CreateNullableFloatFieldData` calls `field_data->FillFieldData(data.vectors().float_vector().data().data(), valid_bitmap.data(), row_count, 0)`, but the source `DataArray` from `DataGen` stores nullable float vectors **compactly** (`generate_float_vector(field_meta, valid_count)` with the default `null_percent = 50`, so ~`row_count/2` rows), while `FillFieldData` unconditionally copies `element_count * dim` values (`FieldData.cpp:92-95`). With `row_count = 100`, `dim = 4` this reads ~800 bytes past the end of the protobuf buffer (heap-buffer-overflow under ASAN) and leaves rows `[valid_count, row_count)` of the `FieldData` as garbage. The test still passes only because the post-sync append consumes the buffer prefix (`total_count == add_count` → `data_ptr = p + 0`). Fix: build a full-layout buffer by scattering the compact valid vectors into their row slots (zero-filling null slots), or drive the test through real storage deserialization.

**Medium (pre-existing, but worth verifying now that the index is the sole owner):** `SegmentGrowingImpl.cpp:998-1005` feeds `FieldData::Data()` into `AppendSegmentIndexDense`, whose nullable branch assumes "data_source contains valid vectors compactly" (`FieldIndexing.cpp:525`). `FieldDataImpl`'s nullable fill semantics are full-layout (nulls occupy slots), and arrow-based binlog deserialization also materializes null slots. If real loaded nullable-vector `FieldData` is full-layout, a post-sync load appends the *physical prefix* of the buffer — i.e. null-placeholder rows — instead of the scattered valid vectors, and under single-owner semantics that wrong copy is now the only copy. The new test does not catch this because its hand-crafted buffer happens to put the valid vectors at the prefix. Please confirm which layout the real load path delivers for nullable dense vectors and add a regression through the real deserialization path if it is full-layout.

**Low:** the ownership predicate `IsVectorDataType(...) && indexing_record_.HasRawData(...)` (`SegmentGrowingImpl.cpp:722`, `:991`) is broader than the set of types the read side can actually serve: `IndexingRecord::GetDataFromIndex` (`FieldIndexing.h:471-477`) silently no-ops for any type outside {FLOAT, FLOAT16, BFLOAT16, SPARSE_U32_F32}, while `CreateIndex` already accepts `VECTOR_INT8` (`FieldIndexing.cpp:830`). Today INT8 can never reach sync (the `AppendingIndex` dispatch falls through for it), so this is unreachable — but the invariant holds by coincidence across three files. Suggest making `GetDataFromIndex` `ThrowInfo` on an unhandled type instead of silently returning, and/or extracting a single shared "index can own raw data for this type" predicate used by both the write-skip guard and the read dispatch.

**Nit:** `SegmentGrowingImpl.cpp:818` still adds the skipped batch's bytes to `stats_.mem_size` even when the data now lives only in the interim index (and `try_remove_chunks` never decrements it) — post-sync memory accounting is systematically inflated. Pre-existing drift, fine as a follow-up.

Verified-good points for the record: the guard is evaluated *after* the nullable validity write, and `ConcurrentVector::get_valid_data()` routes to the same shared `ThreadSafeValidData`, so `AppendSegmentIndexDense`'s validity view stays complete for post-sync batches on both paths; the double-checked-lock read paths in `bulk_subscript_impl` tolerate the ownership handoff; and the Insert regression genuinely pins the no-repopulate behavior (frozen chunk count and offset-mapping total, with data still served correctly from the index).

chyezh:
@czs007 Follow-up on https://github.com/milvus-io/milvus/pull/51913#issuecomment-5099748438:

- **Post-sync `AddWithDataset()` failure:** confirmed as a real pre-existing correctness defect. `recreate_index()` replaces `index_` without resetting `built_`, `sync_with_index_`, or `index_cur_`, so the empty replacement index can still be treated as the synchronized raw-data owner. A proper fix needs an explicit unrecoverable state across append and read paths; a blind `throw;` is not sufficient because Insert has already mutated segment state and the index-add exception may not provide transactional rollback. Tracked separately in #51922. This PR keeps the ownership/performance scope and documents that the current recreate path is invalid after ownership transfer.

- **`CreateNullableFloatFieldData()` overflow:** this does not apply to the test. `storage::CreateFieldData(VECTOR_FLOAT, ..., nullable=true)` returns `FieldData<FloatVector>`, and the virtual call dispatches to `FieldDataVectorImpl::FillFieldData()` (`FieldData.cpp:669-750`). That overload counts the validity bitmap and copies only `valid_count * dim` values, so the compact payload produced by `DataGen` is the expected input and is not read as `row_count * dim`.

- **Real nullable-vector load layout:** the internal binlog path is also compact. `add_vector_payload()` uses `arrow::BinaryBuilder`, appends vector bytes only for valid rows, and uses `AppendNull()` for null rows (`storage/Util.cpp:143-169`). `PayloadReader` then feeds the `BinaryArray` contiguous payload through the same vector-specific `FillFieldData()` path (`FieldData.cpp:358-379`). The load regression therefore models the production `FieldData::Data()` layout used by `load_field_data_common()`.

The INT8/shared ownership predicate and `stats_.mem_size` observations are valid future hardening/accounting follow-ups, but neither is reachable as a correctness regression in the current supported raw-owning interim-index types.

liliu-z (commented):
/lgtm
/approve

Issue #51916, opened 2026-07-28
[Bug]: [Benchmark][Cluster] Bounded Search repeatedly times out waiting for tSafe during concurrent upsert on StreamingNode


wangting0128:
- [x] I have searched the existing issues

### Environment

- Milvus version: `harbor.milvus.io/milvusdb/milvus:3.0-20260727-b598b28977-v2-amd64`
- Deployment mode: cluster with StreamingNode enabled
- MQ: Pulsar
- SDK: pymilvus `2.7.0rc175` (`client_version=2.6`)
- OS/Kubernetes: 4AM cluster, namespace `qa-milvus`
- Deployment config: `zong-server-laion-sn`
- Workload: `test_concurrent_locust_custom_parameters`
- Test case parameters: `client-laion-sn-nullable`
- Collection: `laion_stable_10_nullable`
- Operations: concurrent `search`, `query`, `load`, and `upsert`

### Server Pods

All relevant pods were Running and had no restart during the workflow.

| Role | Milvus node ID | Pod | Kubernetes node | Image |
|---|---:|---|---|---|
| Proxy | - | `laion1b-test-2-milvus-proxy-64f9664566-4fqxq` | `4am-node25` | `3.0-20260727-b598b28977-v2-amd64` |
| StreamingNode / QueryNode leader for `dml_0` | 174 | `laion1b-test-2-milvus-streamingnode-5989cd97bf-lmfr5` | `4am-node14` | `3.0-20260727-b598b28977-v2-amd64` |
| StreamingNode / QueryNode leader for `dml_1` | 166 | `laion1b-test-2-milvus-streamingnode-5989cd97bf-82lhb` | `4am-node26` | `3.0-20260727-b598b28977-v2-amd64` |

### Current Behavior

During sustained concurrent upsert and read traffic, Bounded-consistency Search repeatedly waits for the DML-channel `tSafe` and exhausts the client-side 10-second deadline.

The failure lasts throughout the 7h10m workflow, rather than being limited to deployment or recovery.

```
Search requests: 249851
Search failures: 9557 (3.83%)
Query failures: 0
Upsert failures: 0
```

**Error breakdown**

| Client-visible error | Count | Note |
|---|---:|---|
| `StatusCode.DEADLINE_EXCEEDED` | 9551 | Main failure mode |
| Direct segcore `invalid offset` error | 2 | Separate growing-segment correctness error |
[...]

PR #51917, opened 2026-07-28
fix: [cp 3.0] restore growing index raw data ownership

chyezh:
issue: #51916
pr: #51913

- growing segment writes: restore vector interim-index raw-data ownership and stop repopulating reclaimed chunks after synchronization
- field loading: apply the same vector-only ownership rule while preserving nullable validity metadata
- index append failures: document the invalid recreate path after the only complete raw-data source has been reclaimed

### DIFF
--- a/internal/core/src/segcore/FieldIndexing.cpp
+++ b/internal/core/src/segcore/FieldIndexing.cpp
@@ -361,6 +361,13 @@ VectorFieldIndexing::AppendSegmentIndexSparse(int64_t reserved_offset,
             sync_with_index_.store(true);
         } catch (SegcoreError& error) {
             LOG_ERROR("growing sparse index add error: {}", error.what());
+            // Known design defect: after a raw-data-owning interim index has
+            // synchronized, its source chunks may have been reclaimed and
+            // field_raw_data is no longer a complete recovery source. This
+            // catch is reachable when AddWithDataset fails, but recreating the
+            // index here is not a valid recovery path and must not exist under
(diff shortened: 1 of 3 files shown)


xaxys:

Reviewed at head `2a271a3227` against the 3.0 base.

## Key issues

- **High — Index append failures silently leave the segment without a valid vector-data owner.** `internal/core/src/segcore/FieldIndexing.cpp:362-371,395-405,513-523,548-559` catches `AddWithDataset` failures and calls `recreate_index(...)`, despite the new comments correctly stating that recovery is invalid. The ownership paths at `internal/core/src/segcore/SegmentGrowingImpl.cpp:721-763` and `:990-1003` may already have stopped writing raw chunks. Recreation produces an empty index while `built_`, `sync_with_index_`, and `index_cur_` retain their previous values, and the insertion/loading operation continues to acknowledgement. If Knowhere rejects an append—for example because of memory pressure—the segment can report the new rows as committed while searches, retrievals, and growing-source flushes read from an empty or incomplete index.

## Fix suggestions

- Propagate post-synchronization `AddWithDataset` failures, or transition the segment into an explicit unrecoverable state; do not recreate the index and continue.
- Permit recreation only when complete raw data is demonstrably available, and reset all index state (`built_`, `sync_with_index_`, and `index_cur_`) consistently.
- Add fault-injection coverage for dense nullable/non-nullable and sparse indexes after ownership transfer, verifying that a failed append is not acknowledged or flushed as successful.

liliu-z (commented):
/lgtm
/approve

PR #51699, opened 2026-07-22
fix: [cp2.6]Return early after parameter validation fails in RESTful v2 function DDL handlers

bigsheeper:
Cherry-pick from master (PR still open)

pr: #51511
issue: #51503

## Summary

Cherry-picked from master PR #51511 (open — preemptive backport). Diff is byte-identical to the master PR (handler_v2.go + handler_v2_test.go, +30 lines).

**Note**: Original PR #51511 is still open; update this CP PR if it changes.

## Verification

- [x] Diff identical to master PR (2 files, +30)
- [x] internal/distributed/proxy/httpserver compiles on 2.6
- [x] No conflict markers
- [ ] make static-check (skipped by cherry-pick workflow)

PR #51918, opened 2026-07-28
fix: [hotfix-2.6.21] keep proxy metric status label stable, split cause into its own label (#51495)

czs007:
Cherry-pick of #51496
pr: #51495
issue: #51493

### What

`hotfix-2.6.21` still carries the v2.6.19 in-place status split — `metrics.go` has `FailInputLabel` / `FailSystemLabel` / `RejectedUserLabel` / `RejectedSystemLabel`, and `milvus_proxy_req_count` has no `cause` label. This brings it in line with `hotfix-2.6.20` (#51496) and `2.6` (#51909).

```
```

|---|---|

Outcomes that have no responsible party emit `cause=""`. Prometheus treats an empty label value as equivalent to the label being absent, so those series stay byte-identical to what pre-2.6.19 emitted.

Old queries work unchanged and still count each request once: `status="fail"` is again every hard failure, because Prometheus aggregates over `cause` for free.

### Cherry-pick notes

Applied with **no conflicts** (19 files). Cross-checked file by file against the `2.6` version (#51909): 18 of 19 are byte-identical; `handler_v2.go` differs by exactly the two `return nil, err` lines added by #51699, which is on `2.6` but not on this branch — unrelated to this change.

The emit surface was **re-verified on this branch rather than assumed**: all 12 `WithLabelValues` sites for the two metrics were checked for label arity by a paren-balanced scan, because adding a label makes arity a runtime panic, not a compile error. No stale `FailInputLabel` / `FailSystemLabel` / `RejectedUserLabel` / `RejectedSystemLabel` references remain, and all `ParseMetricLabel` / `failMetricLabel` call sites use the two-value signature.

- Green: `pkg/metrics`, `pkg/util/requestutil`, `pkg/util/merr`.
- The proxy packages (`internal/proxy`, `internal/distributed/proxy/...`) are green on `2.6` with byte-identical content in #51909.

### Related

- #51496 — `hotfix-2.6.20`
- #51909 — `2.6`
- #51526 — `3.0`
- #51907 — `master`

### DIFF
--- a/deployments/monitor/grafana/milvus-dashboard.json
+++ b/deployments/monitor/grafana/milvus-dashboard.json
@@ -4794,7 +4794,7 @@
                 "uid": "${datasource}"
               },
               "exemplar": true,
-              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=~\"fail_.*\"}[2m])/120) by(function_name, pod, node_id)",
+              "expr": "sum(increase(milvus_proxy_req_count{app_kubernetes_io_instance=~\"$instance\", app_kubernetes_io_name=\"$app_name\", namespace=\"$namespace\", status=\"fail\"}[2m])/120) by(function_name, pod, node_id)",
               "interval": "",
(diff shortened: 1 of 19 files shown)


liliu-z (commented):
/approve

Issue #51919, opened 2026-07-28
[Bug]: ThreadPool::Submit can orphan accepted tasks when worker expansion fails


xaxys:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Environment

- Milvus version: current `master` (`14564e2615` during the audit)
- Component: C++ shared `milvus::ThreadPool`
- Deployment mode: standalone or cluster
- Trigger class: process/OS thread-resource exhaustion during dynamic pool expansion

## Current Behavior

`ThreadPool::Submit()` does not have a single observable acceptance boundary.

The current order is:

1. construct a packaged task;
2. enqueue its wrapper into `work_queue_`;
3. update metrics and acquire the pool mutex;
4. if the pool is below its configured maximum and no worker is idle, construct another `std::thread`;
5. only then call `task_ptr->get_future()` and return the future.

If dynamic `std::thread` construction throws `std::system_error` after step 2, the caller observes a failed submission even though the task is already owned by the queue. No future is returned, so the caller cannot wait for, cancel, or otherwise track that accepted task. An existing worker can execute it later.

The realistic trigger is rare: the pool must be expanding while the OS rejects a new thread, typically with `EAGAIN` because of a process/system thread limit or temporarily unavailable resources. The rarity does not remove the lifetime hazard. Callers commonly rely on vectors of returned futures, `WaitAllFutures`, `DrainFutures`, or scope guards to keep captured state alive. Those mechanisms cannot drain a task whose future was never returned.

One concrete example is the parallel packed-index read path: an accepted task may retain raw pointers or references to an entry buffer/download state and may write to an fd. If `Submit()` throws after enqueue, the caller can unwind, destroy the buffer/state, or close the fd while the queued task still runs. Depending on fd reuse and timing, this can become a use-after-free or a write to an unrelated file.

## Affected Surface

The audit found 31 direct production `Submit()` call sites in 11 implementation files on `master`:

- storage: `IndexEntryReader`, `DiskFileManagerImpl`, `IndexEntryEncryptedLocalWriter`, and `Util`;
- index/json stats: `JsonKeyStats`;
- segcore: `ChunkedSegmentSealedImpl`, `SegmentGrowingImpl`, `Utils`, `memory_planner`, `reduce/Reduce`, and `search_result_export_c`.

[...]

PR #51920, opened 2026-07-28
fix: return futures for accepted thread pool tasks

xaxys:
issue: #51919

## What does this PR do?

This PR closes a rare but unsafe acceptance gap in the shared C++ `ThreadPool::Submit()` implementation.

On the original base of this PR, `Submit()` enqueued a task and only returned its future after post-enqueue metrics, locking, and optional dynamic worker creation. If `std::thread` construction failed while expanding the pool, the task remained queued but `Submit()` threw. The caller therefore had no future for an accepted task and could unwind while that task still referenced caller-owned state.

The concrete trigger is uncommon—an expanding pool must hit an OS/process thread-resource limit—but the consequence is not local to one reader. Current `master` has 31 direct production submissions across 11 implementation files. Several callers drain returned futures specifically to protect stack references, buffers, output state, readers, or file descriptors; no caller can drain a future that was never returned.

## Divergence from current master

Current master now contains the alternative implementation merged through [#51403](https://github.com/milvus-io/milvus/pull/51403). It obtains the future, publishes the task first, and treats dynamic worker expansion as best effort: worker-spawn failure is caught, the accepted task remains queued, and the future is returned. This resolves the original thread-construction failure for ordinary submissions without rejecting the task.

This PR keeps a different contract. When projected demand requires another worker, provisioning completes before queue publication. If provisioning fails, `Submit()` throws and the current task has not been accepted. The difference is therefore no longer whether the original failure is noticed; it is whether failed expansion should leave the task queued or reject it before publication.

Two theoretical boundaries remain in the current-master policy:

1. the post-enqueue `LOG_WARN` calls in failure handlers may themselves allocate and throw under extreme memory pressure;
2. if every existing worker is blocked and a task submitted from a worker synchronously waits on another task in the same pool, failed expansion can leave that child future waiting indefinitely.

A static audit of the current production call graph found no same-pool nested synchronous submission. Existing nested load paths deliberately cross pools, such as MIDDLE to HIGH/LOW, and one LOW-pool decode path explicitly documents that it must not be called from a LOW-pool worker. Same-pool submit-and-wait should remain disallowed; callers should use a different pool, continuations, or return the future to a non-worker coordinator instead.

The PR is being held open for now while the desired acceptance policy is considered. No additional implementation change is proposed solely for a nested-submission scenario that has no current production instance.

## How does it work?

The submission contract is now based on queue publication:

- the packaged-task future is obtained before publication;
- while holding the pool mutex, `Submit()` evaluates projected queue demand and creates/registers any required expansion worker before the task is enqueued;
- if `std::thread` construction fails, the submission throws while the task is still unaccepted and absent from `work_queue_`;
- after enqueue succeeds, optional metric failures are contained without suppressing the future of the accepted task;
- normal task results and task-thrown exceptions continue to propagate through the future.

Provisioning before publication also preserves nested-submission behavior. If the only worker submits a child task and expansion fails, the child submission remains a visible exception instead of returning an unserviceable child future.

The implementation remains directly in the existing `Submit()` template. It does not add a test peer, injectable worker factory, new public/private helper, `ThreadPool` data member, or other test-only production seam. The current master already provides a deterministic worker-spawn hook; this PR reuses that existing seam to verify the rejection boundary.

## Affected surface

The shared fix covers all current and future users. The audited direct callers are:

- storage: `IndexEntryReader`, `DiskFileManagerImpl`, `IndexEntryEncryptedLocalWriter`, and `Util`;
- index/json stats: `JsonKeyStats`;
- segcore: `ChunkedSegmentSealedImpl`, `SegmentGrowingImpl`, `Utils`, `memory_planner`, `reduce/Reduce`, and `search_result_export_c`.

This is intentionally a pool-level fix. Adding ownership gates independently to consumers would duplicate the same contract, impose per-consumer overhead, and leave other callers exposed.

## Behavior and performance

The normal submission path adds no allocation, atomic state, worker wait, or `ThreadPool` data member. It reorders existing locking, worker provisioning, queue publication, and future retrieval so failure and acceptance are unambiguous. Paired empty-task microbenchmarks on the same provisioning-before-publication ordering showed no normal-path regression; details are posted separately for review.

The intentional exceptional-path behavior is:

- worker provisioning failure rejects the not-yet-published task;
- after publication, the caller receives the future that tracks the accepted task.

The PR does not change the existing pool-shutdown lifecycle. Production `ThreadPools` are shut down only from teardown/destruction after their users stop submitting work.

## Validation

Validation covers ordinary return values and task-thrown exceptions through the reordered submission path. It also reuses the deterministic worker-spawn hook already present on current master to verify that a provisioning failure propagates to the caller and the rejected task never enters the queue or executes. This PR does not introduce a new production test seam or change the public API or object layout.

### DIFF
--- a/internal/core/src/storage/ThreadPool.h
+++ b/internal/core/src/storage/ThreadPool.h
@@ -139,64 +139,67 @@ class ThreadPool {
         std::function<decltype(f(args...))()> func =
             std::bind(std::forward<F>(f), std::forward<Args>(args)...);
         auto task_ptr =
-            std::make_shared<std::packaged_task<decltype(f(args...))()>>(func);
-
-        auto enqueue_time = std::chrono::steady_clock::now();
-        auto* queue_metric = metric_queue_duration_;
-        auto* execute_metric = metric_execute_duration_;
-        std::function<void()> wrap_func = [task_ptr,
-                                           enqueue_time,
-                                           queue_metric,
-                                           execute_metric]() {
-            auto execute_start = std::chrono::steady_clock::now();
-            if (queue_metric) {
-                queue_metric->Observe(
-                    std::chrono::duration<double>(execute_start - enqueue_time)
-                        .count());
-            }
-            auto observe_execute = [&]() {
-                if (execute_metric) {
-                    execute_metric->Observe(
-                        std::chrono::duration<double>(
-                            std::chrono::steady_clock::now() - execute_start)
-                            .count());
-                }
-            };
-            try {
-                (*task_ptr)();
-            } catch (...) {
-                observe_execute();
-                throw;
-            }
-            observe_execute();
-        };
-
+            std::make_shared<std::packaged_task<decltype(f(args...))()>>(
+                std::move(func));
         auto future = task_ptr->get_future();
 
-        std::unique_lock<std::mutex> lock(mutex_);
-        work_queue_.enqueue(std::move(wrap_func));
+        auto* queue_metric = metric_queue_duration_;
+        auto* execute_metric = metric_execute_duration_;
 
-        if (idle_threads_size_ > 0) {
-            condition_lock_.notify_one();
-        }
-        if (work_queue_.size() > static_cast<size_t>(idle_threads_size_) &&
-            current_threads_size_ < max_threads_size_.load()) {
-            // Dynamic increase thread number
-            try {
+        // Lifetime-safety contract: do not move work_queue_.enqueue() above
+        // this block. Every fallible worker-provisioning step must finish
+        // before task publication. A caller may release task-captured state
+        // when Submit() throws; enqueueing first could leave an accepted but
+        // untracked task using that state after the caller unwinds.
+        {
+            std::lock_guard<std::mutex> lock(mutex_);
+            if (work_queue_.size() >= static_cast<size_t>(idle_threads_size_) &&
+                current_threads_size_ < max_threads_size_.load()) {
+                // Dynamic increase thread number
                 if (worker_spawn_hook_for_test_) {
                     worker_spawn_hook_for_test_();
                 }
                 threads_.emplace_back(&ThreadPool::Worker, this);
                 current_threads_size_++;
-            } catch (const std::exception& e) {
-                LOG_WARN(
-                    "Failed to expand thread pool {}: {}", name_, e.what());
-            } catch (...) {
-                LOG_WARN("Failed to expand thread pool {}", name_);
+            }
+
+            // Queue duration starts after lock wait and worker provisioning, so
+            // it measures time from queue publication to task execution.
+            auto enqueue_time = std::chrono::steady_clock::now();
+            std::function<void()> wrap_func =
+                [task_ptr, enqueue_time, queue_metric, execute_metric]() {
+                    auto execute_start = std::chrono::steady_clock::now();
+                    if (queue_metric) {
+                        queue_metric->Observe(std::chrono::duration<double>(
+                                                  execute_start - enqueue_time)
+                                                  .count());
+                    }
+                    auto observe_execute = [&]() {
+                        if (execute_metric) {
+                            execute_metric->Observe(
+                                std::chrono::duration<double>(
+                                    std::chrono::steady_clock::now() -
+                                    execute_start)
+                                    .count());
+                        }
+                    };
+                    try {
+                        (*task_ptr)();
+                    } catch (...) {
+                        observe_execute();
+                        throw;
+                    }
+                    observe_execute();
+                };
+            work_queue_.enqueue(wrap_func);
+            if (idle_threads_size_ > 0) {
+                condition_lock_.notify_one();
             }
         }
-        lock.unlock();
 
+        // Queue publication is the acceptance boundary. Do not let optional
+        // post-enqueue metrics prevent the accepted task's future from being
+        // returned.
         try {
             if (metric_submitted_) {
                 metric_submitted_->Increment();
@@ -204,14 +207,10 @@ class ThreadPool {
             if (metric_queue_depth_) {
                 metric_queue_depth_->Set(work_queue_.size());
             }
-        } catch (const std::exception& e) {
-            LOG_WARN("Failed to update thread pool {} submit metrics: {}",
-                     name_,
-                     e.what());
         } catch (...) {
-            LOG_WARN("Failed to update thread pool {} submit metrics", name_);
+            // Metrics are best-effort. Nothing after queue publication may
+            // throw before the accepted task's future is returned.
         }
-
         return future;
     }
 
@@ -289,7 +288,7 @@ class ThreadPool {
     prometheus::Histogram* metric_execute_duration_{nullptr};
 
  private:
-    friend class ThreadPoolTest_WorkerSpawnFailureDoesNotFailQueuedTask_Test;
+    friend class ThreadPoolTest_WorkerSpawnFailureRejectsUnqueuedTask_Test;
 
     // Deterministic test seam for worker-spawn failure coverage.
     std::function<void()> worker_spawn_hook_for_test_;
--- a/internal/core/src/storage/ThreadPoolTest.cpp
+++ b/internal/core/src/storage/ThreadPoolTest.cpp
@@ -19,6 +19,7 @@
 #include <concepts>
 #include <future>
 #include <limits>
+#include <stdexcept>
 #include <string>
 #include <system_error>
 
@@ -189,7 +190,7 @@ TEST_F(ThreadPoolTest, LoadFileOverheadControllerIsLazyAndStable) {
     EXPECT_EQ(file_group, file_owner.GetOrCreate(/*executor_workers=*/8));
 }
 
-TEST_F(ThreadPoolTest, WorkerSpawnFailureDoesNotFailQueuedTask) {
+TEST_F(ThreadPoolTest, WorkerSpawnFailureRejectsUnqueuedTask) {
     ThreadPool pool(1.0, "test_pool");
 
     std::promise<void> blocker_started;
@@ -208,18 +209,24 @@ TEST_F(ThreadPoolTest, WorkerSpawnFailureDoesNotFailQueuedTask) {
     };
 
     std::atomic<int> task_runs{0};
-    std::future<void> task_future;
-    EXPECT_NO_THROW(
-        task_future = pool.Submit([&task_runs]() { task_runs.fetch_add(1); }));
+    EXPECT_THROW(pool.Submit([&task_runs]() { task_runs.fetch_add(1); }),
+                 std::system_error);
 
     unblock_worker.set_value();
     blocker.get();
 
-    ASSERT_TRUE(task_future.valid());
-    ASSERT_EQ(task_future.wait_for(std::chrono::seconds(2)),
-              std::future_status::ready);
-    task_future.get();
-    EXPECT_EQ(task_runs.load(), 1);
+    EXPECT_EQ(task_runs.load(), 0);
+}
+
+TEST_F(ThreadPoolTest, SubmissionPreservesResultAndTaskException) {
+    ThreadPool pool(1.0, "test_pool");
+
+    auto result = pool.Submit([](int value) { return value + 1; }, 41);
+    EXPECT_EQ(result.get(), 42);
+
+    auto failure =
+        pool.Submit([]() { throw std::runtime_error("task failed"); });
+    EXPECT_THROW(failure.get(), std::runtime_error);
 }
 
 }  // namespace milvus


xaxys:
Validation update for `86568ae067`:

### Test-boundary decision

The final revision keeps the fix directly in the existing `ThreadPool::Submit()` template. The earlier `ThreadPoolTestPeer`, private `SubmitImpl`, injectable worker factory, and deterministic worker-creation failure case have been removed.

Portable C++ does not provide a process-local hook for forcing `std::thread` construction to fail. Keeping the injection seam would make production structure depend on one test. Alternatives such as `pthread_create` symbol interposition, seccomp, process-wide resource limits, or forced OOM are platform-specific and can affect unrelated tests in the shared `all_tests` binary.

The failure boundary is therefore explicit in the production ordering:

1. create the packaged task and obtain its future;
2. under `mutex_`, calculate projected demand and construct/register any required worker;
3. publish the task with `work_queue_.enqueue()` only after provisioning succeeds;
4. after publication, isolate optional metric failures and return the tracking future.

If `std::thread` construction throws, step 3 has not happened and no orphaned task exists.

`ThreadPoolTest.*` was run for 20 repetitions: **160/160 passed**. The added test verifies ordinary return values and task-thrown exceptions through the reordered path. All 11 affected production translation units compile against the final inline template.

### Normal-path submission microbenchmark

The previously reported benchmark used the same provisioning-before-publication production ordering. The final change only removes the test-only wrapper/factory and inlines that logic back into `Submit()`; it adds no runtime work.

Configuration: four-worker pool, 10,000 warm-up tasks, then 200,000 empty tasks per sample; seven samples per process; six interleaved baseline/fix process pairs.

| Pair | upstream/master | provisioning-before-publication |
|---:|---:|---:|
| 1 | 3,426 ns | 2,345 ns |
| 2 | 3,211 ns | 2,262 ns |
| 3 | 3,187 ns | 2,130 ns |
| 4 | 3,601 ns | 2,238 ns |
| 5 | 3,338 ns | 2,133 ns |
| 6 | 3,365 ns | 2,265 ns |
| Aggregate median | ~3,352 ns | ~2,250 ns |

This synthetic result is not claimed as a production throughput improvement. It establishes that the new lock/publication ordering does not add normal-path submission overhead.

### Audited impact surface

On current `master`, the shared pool has **31 direct production `Submit()` call sites in 11 implementation files**: seven in `IndexEntryReader` and 24 in storage, json-stats, and segcore consumers. Moving the acceptance boundary into `ThreadPool` removes the post-enqueue/no-future window for all of them without per-consumer ownership gates or test-only production APIs.

xaxys:
Latest-head clarification for `86568ae067`:

- the final implementation is contained directly in the existing `ThreadPool::Submit()` template;
- it does not contain `ThreadPoolTestPeer`, `SubmitImpl`, a friend declaration, an injectable worker/thread factory, or any other test-only production seam;
- the remaining unit test exercises public submission behavior (result and task-exception propagation). It does not claim to deterministically inject `std::thread` construction failure.

A portable process-local hook for forcing the `std::thread` constructor to fail does not exist. The available alternatives are either production injection structure or process/platform-wide fault injection, so the failure invariant is instead made explicit in the source ordering: required worker provisioning completes before queue publication.

czs007:
Adversarial review confirmed one issue requiring follow-up; all correctness, concurrency, and exception-path checks on the `Submit()` reorder passed.

### Medium

- **ThreadPool.h:165-183 / ThreadPoolTest.cpp:142** — Test coverage: the core invariant of this PR — that when `std::thread` construction throws, `Submit()` fails before `work_queue_.enqueue`, leaving no orphaned task in the queue — has no regression test. The new test `SubmissionPreservesResultAndTaskException` (ThreadPoolTest.cpp:142-151) only asserts return-value and task-body exception propagation, both of which are guaranteed by `packaged_task` and passed on master before this PR; it exercises the success path through the reordered code but never verifies the failure ordering, so a revert of the reorder would go undetected. The deterministic worker-creation-failure harness (`ThreadPoolTestPeer` / injectable worker factory) was removed because portable C++ cannot inject `std::thread` construction failure in-process, so the invariant is currently guaranteed only by source order. Suggestion: track this as a known gap — e.g. restore an injectable worker-factory seam behind a test-only hook, or at minimum document the required ordering at the provisioning site so future edits don't silently break it. *(raised by chyezh, tinswzy)*

xaxys:
Thanks — I agree with the coverage characterization: `SubmissionPreservesResultAndTaskException` verifies that the reorder preserves normal future semantics, but it does not exercise the worker-creation failure boundary. It should not be presented as a regression test for that exceptional path.

I am intentionally not adding a deterministic failure hook to the production `ThreadPool` class. Even a friend declaration or private injectable worker factory makes the formal implementation structurally depend on test code. The production API and class design should be driven by runtime requirements, not by the need to mock one standard-library constructor.

There is no portable standard-C++ alternative:

- `std::thread` has no failure-injection API, its constructor is not virtual, and it cannot be legally replaced or mocked from a test;
- exhausting the process/thread limit by creating real threads is nondeterministic and unsafe for CI;
- interposing `pthread_create` or a private libstdc++ symbol is platform/ABI-specific and changes thread creation for the shared `all_tests` process, so it is not an acceptable cross-platform unit-test contract;
- placing such a test in a separate executable would not satisfy the existing CI path unless that target became another maintained test surface.

For validation, I retained the earlier friend-based implementation on the separate local branch `local/threadpool-test-peer-proof` at `241b1a8e10`. It contains `ThreadPoolTestPeer`, the injectable worker factory, and the deterministic `WorkerCreationFailureRejectsNestedSubmitBeforeEnqueue` test. That version passed the local `ThreadPoolTest.*` run for 20 repetitions (180/180 cases), confirming that worker-creation failure rejects the submission before queue publication and leaves no orphaned task.

That branch is evidence for the behavior, not code proposed for merge. The production PR keeps the invariant explicit at the provisioning site:

1. obtain the task future;
2. provision and register any required worker;
3. publish the task to `work_queue_` only after provisioning succeeds.

If a production-motivated thread-creation abstraction is introduced later, the deterministic test can be added against that abstraction. Until then, I prefer the documented source-order invariant over adding permanent test-only structure to `ThreadPool`.

xaxys:
Follow-up: the latest head is `6f75579ede`.

No test hook or production API was added. The only change from the previous head is a stronger source-level maintenance contract immediately above worker provisioning:

> Do not move `work_queue_.enqueue()` above this block. Every fallible worker-provisioning step must finish before task publication.

The comment also records the lifetime reason: after `Submit()` throws, the caller may release state captured by the task; publishing before a fallible step could therefore leave an accepted but untracked task using released state. This documents the invariant without making `ThreadPool` depend on test-only structure.

xaxys:
Head update: `3af369c66a` is the same single PR commit rebased onto current `upstream/master` (`56c3b79dcd`). There is no additional implementation change.

This also replaces the stale Codecov comparison that analyzed `ade6477` against an outdated base. The previous report already showed 100% coverage for modified coverable lines; the newly triggered CI/Codecov run should report against the current PR head and base.

czs007:
## Re-review of `86568ae06757`

- **Confirmed** — `ThreadPoolTest.cpp:142` (medium, missing regression test for clean rejection when worker creation fails): both reviewers re-audited the unchanged head and confirmed the test still does not exercise this invariant; details below.

---

Adversarial re-review of this commit found one medium test-coverage gap (carried over from the prior round and re-confirmed); no new critical or high issues.

### Medium

- **`internal/core/src/storage/ThreadPoolTest.cpp:142`** — The new test `SubmissionPreservesResultAndTaskException` only asserts normal result propagation (`result.get() == 42`) and exceptions thrown by the task body. Both behaviors are guaranteed by `std::packaged_task` itself and pass identically on pre-PR code, so the test does not protect the invariant this PR introduces: when dynamic `std::thread` construction fails, `Submit()` must throw *before* `work_queue_.enqueue(wrap_func)` (see the lifetime-safety contract at `ThreadPool.h:165-169`), leaving no accepted-but-futureless orphan task. A regression that restores the old enqueue-before-provisioning order would still pass this test, and the injectable worker-factory seam was removed from the final head. **Suggestion:** add a deterministic worker-construction failure injection point, force pool expansion to fail, and assert that `Submit()` throws while the task is neither enqueued nor executed. *(raised by chyezh, tinswzy)*

czs007:
Adversarial review confirmed one medium and two low issues; the core reorder is correct and mergeable, with the items below recommended as follow-ups.

### Medium

**`ThreadPoolTest.cpp:142` — Rejection invariant on worker-provisioning failure has no regression test.** The PR's central invariant — that a `std::thread` construction failure during dynamic expansion must occur *before* `work_queue_.enqueue`, so `Submit` throws and the task is never enqueued (the contract stated in the `ThreadPool.h:165-169` lifetime-safety comment) — is currently guarded only by that source comment. The new `SubmissionPreservesResultAndTaskException` test only asserts `result.get() == 42` and future-propagated task exceptions, both of which are guaranteed by `std::packaged_task` and pass on the pre-change code; if the enqueue were moved back ahead of provisioning, the test would still be green. Note the PR body already states a deterministic thread-constructor failure is intentionally not injected, so there is no overclaim. **Suggestion:** as a follow-up, add an injectable worker factory (or land the deterministic case from the author's `local/threadpool-test-peer-proof` branch) so the failure path fails loudly if the ordering regresses. *(raised by chyezh, tinswzy)*

### Low

**`ThreadPool.h:135` — Queue-duration metric is now inflated by lock wait and thread creation.** `enqueue_time` is captured before the lock, but the enqueue was moved inside the `mutex_` critical section (`ThreadPool.h:181`) and may be preceded by a millisecond-scale `std::thread` construction (`ThreadPool.h:175`). The queue-duration histogram (`ThreadPool.h:143-147`) therefore now includes mutex contention and expansion time in "queued" time — a measurement-accuracy regression introduced by this reorder, with no correctness impact. **Suggestion:** capture `enqueue_time` immediately before (or at) the actual `enqueue` call so the metric measures only time spent in the queue. *(raised by chyezh, tinswzy)*

**`ThreadPool.h:136` — Metric pointers are read outside `mutex_`, racing with `SetMetrics` (pre-existing).** `Submit` reads `metric_queue_duration_`/`metric_execute_duration_` without the lock (`ThreadPool.h:136-137`) and reads `metric_submitted_`/`metric_queue_depth_` after the lock scope ends (`ThreadPool.h:191-195`), while `SetMetrics` (`ThreadPool.h:242-250`) writes these pointers under `mutex_` — a formal data race. This predates the PR (the old code also read these pointers before locking) and is low risk in practice since `SetMetrics` is called once during initialization. **Suggestion:** address separately, e.g. make the pointers atomic or require `SetMetrics` before any `Submit`. *(raised by tinswzy)*

xaxys:
Thanks — I agree with the queue-duration finding and fixed it in `f2368dd329`.

`enqueue_time` is now captured after `mutex_` has been acquired and after any required worker has been created and registered, immediately before the wrapper is published to `work_queue_`. The histogram therefore measures queue-publication-to-execution time and no longer includes submit-side mutex contention or dynamic thread-creation latency.

The acceptance/lifetime invariant is unchanged: every fallible worker-provisioning step still completes before `work_queue_.enqueue()`. If worker creation fails, the task remains unpublished and `Submit()` throws without leaving an orphaned task.

I am not changing the metric-pointer synchronization in this PR. The formal race observation is valid for the public class interface, but it is pre-existing and the production lifecycle prevents it: `ThreadPools::GetThreadPool()` holds its global unique lock, constructs the pool, calls `SetMetrics()`, and only then returns the first externally usable reference. `ThreadPools.cpp` is the only production `SetMetrics()` caller, so production does not rebind these pointers concurrently with `Submit()`.

If that API is hardened separately, I would prefer constructor-time metric binding or making `SetMetrics()` an initialization-only internal operation over adding atomic loads to every metric pointer access.

czs007:
Adversarial review confirmed two non-blocking issues: a test-coverage gap around the relocated worker-provisioning branch (medium) and a minor, intentional lock-contention trade-off in `Submit()` (low).

### Medium

- **`internal/core/src/storage/ThreadPoolTest.cpp:142` — new test gives only racy coverage of the relocated provisioning branch; the failure-path invariant is untested.** Whether `SubmissionPreservesResultAndTaskException` enters the dynamic-provisioning block depends on OS scheduling: the single min-worker only increments `idle_threads_size_` (`ThreadPool.cpp:118-119`) after it wins the mutex, so the guard at `ThreadPool.h:145` evaluates `0 >= 0` (branch taken) or `0 >= 1` (branch skipped) depending on which thread runs first — this is flaky-by-schedule, not deterministic regression coverage. More importantly, the PR's core invariant — that a `std::thread` construction failure at `ThreadPool.h:148` rejects the submission without enqueuing the task (enqueue now at `ThreadPool.h:182`, after provisioning) — has no test at all; the deterministic test hook was intentionally removed. What the test does stably verify is `packaged_task` result/exception propagation, which the old code already provided. **Suggestion:** add a deterministic regression test for the worker-construction-failure → reject-and-not-enqueue path as a follow-up, and avoid describing the moved branch as covered. Both reviewers agree this is a follow-up, not a merge blocker. *(raised by chyezh, tinswzy)*

### Low

- **`internal/core/src/storage/ThreadPool.h:182` — `enqueue()` moved inside the pool-level `mutex_` critical section, serializing producers with worker bookkeeping (accepted, intentional trade-off).** Previously `work_queue_.enqueue()` ran outside `mutex_` under only SafeQueue's own `shared_mutex`; it now runs inside the `mutex_` critical section (`ThreadPool.h:144` → `h:182`). Lock ordering is consistently `mutex_` → SafeQueue lock across `Submit`, `Worker` dequeue, and `SetMetrics`, so there is no deadlock risk, but all producers now contend with each worker's idle/dequeue bookkeeping, and the held region also includes the `std::function` wrapper construction (heap-allocating, `h:157-181`) and `steady_clock::now()` (`h:156`). This is a deliberate cost of the lifetime-safety contract documented at `h:138-142` (provisioning must precede task publication). **Suggestion:** no change needed; recorded for awareness. *(raised by tinswzy)*

czs007:
Adversarial review of this PR confirmed the core reorder is correct and mergeable; the verified issues below are one out-of-scope caller-side lifetime hazard, one acknowledged test gap, and two rare error-path holes in the acceptance contract.

### High

- **`ChunkedSegmentSealedImpl.cpp:2407` — Partial batch-submission failure can unwind before futures are guarded (caller-side, outside this diff).** Batched callers install the `folly::makeGuard` futures drain *after* the submission loop (`ChunkedSegmentSealedImpl.cpp:2462`, `Reduce.cpp:220/482`). If a mid-loop `Submit()` throws (e.g. `std::thread` construction failure), previously accepted tasks are not yet guarded; `std::future` from a `packaged_task` does not block on destruction, so the caller unwinds while accepted tasks still reference stack-captured state (`&fs`, `&field_ids_for_stats`, `&debug_key`, or the `ReduceHelper` `this`). The comment at `ChunkedSegmentSealedImpl.cpp:2405` ("Futures are always joined below before this function returns") does not hold on that path. This lives in caller files this PR does not touch and should not block the diff — **suggestion:** in a follow-up PR, install the futures guard before the submission loop (on the empty vector). *(raised by chyezh, tinswzy)*

### Medium

- **`ThreadPoolTest.cpp:142` — No regression test for the worker-creation failure path.** The PR's core behavior — a failed worker expansion leaves no task in `work_queue_` and makes `Submit()` throw — is untested; `SubmissionPreservesResultAndTaskException` only covers happy-path results and task-exception propagation. The injectable worker-factory seam was intentionally removed and the author has confirmed this as an accepted trade-off; recorded here to keep the gap visible. **Suggestion:** none required now; revisit if a non-invasive injection seam becomes available. *(raised by tinswzy)*

### Low

- **`ThreadPool.h:198` — `LOG_WARN` in the post-enqueue catch block can itself throw, breaking the "no throw after acceptance" contract.** The catch handlers call `LOG_WARN`, which expands to `fmt::format` calls (including `SERVER_MODULE_FUNCTION`) that allocate strings and can throw `std::bad_alloc` under memory pressure. That exception is uncaught and escapes `Submit()` after the task is already accepted — reintroducing the class of bug this PR eliminates. **Suggestion:** wrap the logging in its own `catch (...) {}` or use a non-allocating log path so the boundary from enqueue to `return future` is strictly noexcept. *(raised by tinswzy)*

- **`ThreadPool.h:150` — `threads_` map insertion can throw before the move, destroying a joinable thread and calling `std::terminate()`.** In `threads_[t.get_id()] = std::move(t)`, `unordered_map::operator[]` may throw `std::bad_alloc` during node allocation/rehash before the move-assignment runs; `t` still owns a joinable thread, and its destructor during unwind terminates the process — contradicting the clean-throw contract endorsed by the new comment at lines 138–142. This is a pre-existing pattern, but the PR relocates and re-endorses it. **Suggestion:** `join()`/`detach()` on insertion failure (e.g. `reserve` up front or wrap the insert in a try block), or narrow the comment's claim. *(raised by tinswzy)*

xaxys:
Thanks. I audited these findings against the current source and updated the PR at `9f0671a5e9`.

The partial batch-submission lifetime hazard is valid, but it should not be fixed in #51920. This PR defines the acceptance boundary for one `Submit()` call; it cannot make a sequence of independent submissions transactional or drain futures returned by earlier calls. The affected loops are caller-side code outside this diff and the same exposure already existed for any later submission/allocation failure. A separate PR should address the full pattern consistently, with an exact pre-submit `reserve()` plus a drain guard installed before the first submission, or a shared batch-submission helper. The exact reserve is important because a guard alone cannot track a future if vector insertion throws after `Submit()` has already returned it.

I fixed the newly introduced post-acceptance logging issue. Metric updates remain best-effort and are caught after queue publication, but the catch path no longer calls `LOG_WARN` or performs any other potentially throwing work. From publication to `return future`, metric failures are now contained without reopening the accepted-task-without-future window.

I am not changing `threads_[t.get_id()] = std::move(t)` here. The possible `bad_alloc` followed by destruction of a joinable local thread is a real extreme-OOM concern, but the insertion pattern predates this PR. The reordered implementation performs it before queue publication, so it cannot leave the submitted task accepted without a future. Hardening worker registration should be handled separately rather than expanding this focused acceptance-contract fix.

The deterministic worker-construction failure test gap remains the previously documented trade-off: portable `std::thread` has no non-invasive failure-injection mechanism, and this PR does not add a test-only seam to production code.

czs007:
Adversarial review of this PR surfaced one non-blocking issue; the core `Submit()` reorder was independently verified as correct and mergeable.

### Medium

- **`ThreadPoolTest.cpp:147` — New test does not cover the rejection boundary on worker-provisioning failure.** `SubmissionPreservesResultAndTaskException` verifies normal-result and task-thrown-exception propagation through the returned future, but both paths exercise tasks that were already accepted. The PR's core invariant — that a `std::thread` construction failure (e.g. `std::system_error` on OS thread limits) cleanly rejects the task before enqueue, leaving no accepted-but-unqueued work — is never triggered by any test. This is acknowledged in the PR description as intentionally omitted (no portable in-process failure hook without adding a test seam to a production header). **Suggestion:** track this as a follow-up and add a deterministic provisioning-failure injection (e.g. a minimal test seam or a fault-injection build) so the rejection boundary at `ThreadPool.h:138-152` gets direct coverage. *(raised by tinswzy)*

xaxys:
Rebased lastest master branch that fixed ut

czs007:
> Also reported: 2 pre-existing hazards (real, but not introduced here — follow-up only, does **not** block this PR).

## Overview

This round surfaced one observability regression introduced by the `Submit()` reorder in `ThreadPool.h`, plus three real but pre-existing exception-safety / resource hazards in callers that belong in separate follow-up work.

## Low

**`internal/core/src/storage/ThreadPool.h:156` — the new lock wait falls outside the `queue_duration` histogram**

This PR moves `work_queue_.enqueue()` from outside the lock into the `mutex_` critical section (`ThreadPool.h:144-186`) and moves the `enqueue_time` sample from the top of `Submit()` to after the lock is taken and worker provisioning has completed (`:156`). Before the change, a task was enqueued before the lock was acquired, so lock wait could not delay when the task started; after the change a task cannot start until `mutex_` is held — and `mutex_` is the central lock held by every `Worker()` on each loop iteration, on each dequeue, and by `FinishThreads()` while joining (`ThreadPool.cpp:98-110`, `:118-165`). That new submit-side wait is not inside `internal_storage_pool_queue_duration_seconds` (its start point is now after the lock) and no other metric covers it, so a submit-side slowdown is invisible in monitoring.

Suggestion: move the `enqueue_time` sample back above the `std::lock_guard` (a one-line change), or state in the PR description that `queue_duration` semantics changed from "Submit entry → execution start" to "enqueue → execution start", and say what observes the lock wait instead. *(raised by czs007)*

- **`internal/core/src/segcore/reduce/Reduce.cpp:220`** (originally reported under `internal/core/src/query/Reduce.cpp`, which does not exist — the code is in `segcore/reduce/`) — the submission loop in `FillPrimaryKey()` (`:212-219`) emplaces all N futures before the reclaiming `folly::makeGuard` is constructed at `:220`. If the k+1-th `pool.Submit()` throws, the guard does not exist yet, and `std::packaged_task`-backed `std::future` destructors do not block, so the first k already-accepted tasks keep running and keep dereferencing the captured `this` and `op_ctx_` (`:213`, `:216`). `RefineDistances()` is structurally identical (submission at `:460-481`, guard only at `:482`) and additionally captures the raw `dense_blob` pointer. Failure: a querynode near `RLIMIT_NPROC` / its container memory cap runs a `num_segments_ > 1` search and `pthread_create` fails while the MIDDLE pool expands, making the k+1-th `Submit()` throw → the k accepted tasks still call `FillPrimaryKeys(plan_, *search_result, op_ctx_)` after `ReduceHelper` and `OpContext` were destroyed with the `PrepareSearchResultsForExportImpl` frame (the exception is swallowed at `search_result_export_c.cpp:1178`), a heap-use-after-free that corrupts an already-freed `SearchResult` at best and segfaults the querynode at worst. Unrelated to this PR: the diff touches only `storage/ThreadPool.h` and `ThreadPoolTest.cpp`, and `Reduce.cpp` is untouched — the old `Submit()` could throw before returning a future too, so this exposure predates the PR (which in fact narrows it slightly, since the failing task is no longer enqueued). Fix in follow-up by constructing the guard before the submission loop, or putting submission and drain in one RAII scope. *(raised by czs007, chyezh)*

- **`internal/core/src/segcore/memory_planner.cpp:538`** — the submission loop in `LoadCellBatchAsync()` (`:538-621`) has no guard at all; an exception escapes the function and the `futures` return value is lost with it. Both callers' channels are bounded (`GroupChunkTranslator.cpp:353-355`, `ManifestGroupTranslator.cpp:468-470`, capacity `GetMaxThreadNum() * kChannelCapacityMultiplier` over `tbb::concurrent_bounded_queue`, which blocks when full), and both the `while (channel->pop(...))` drain and the catch-block drain sit *after* the `LoadCellBatchAsync()` call, so a throw during submission skips draining entirely. Failure: a storage-v2 segment load splits into multiple batches where one batch has at least channel-capacity cells, and `pool.Submit()` throws mid-loop on a thread/memory-constrained node → accepted tasks fill the bounded channel and then block forever in `channel->push()` (`:615`/`:618`) after `budget.AcquireUntil()` (`:562`) already succeeded, so the scope-exit `release_guard` (`:571-577`) never runs and `batch_budget_bytes` stays permanently checked out of the process-wide singleton budget (`EntryStreamUtils.h:79-83`); `remaining` never reaches 0 so `channel->close()` is never called, the worker thread is lost, and every subsequent segment load on that node waits on credit that is never returned — recoverable only by restarting the process. Unrelated to this PR: `memory_planner.cpp` and both translators are untouched by the diff, and the old `Submit()` could likewise throw before returning a future, so the "throw mid-submission → drain skipped → producer wedged in push" path already held; this PR neither widens nor narrows it. Follow-up fix: bring submission and drain into one RAII scope inside `LoadCellBatchAsync()`. *(raised by czs007, chyezh)*

- **`internal/core/src/storage/ThreadPool.h:148`** — in the relocated worker-provisioning block, `threads_[t.get_id()] = std::move(t)` (`:150`) allocates a map node and can throw `std::bad_alloc`. Failure: in a memory-capped container where `pthread_create` still succeeds (stack already allocated) but the heap is near exhaustion, that node allocation throws → the already-started, joinable `std::thread` is destroyed during unwinding, calling `std::terminate()` and aborting the querynode/datanode process instead of producing the clean rejection this PR is about. Unrelated to this PR: those three lines were moved character-for-character from after `enqueue` to before it, and the identical pattern already exists untouched in `ThreadPool::Init()` (`ThreadPool.cpp:72-77`), so the behavior is unchanged by this diff — it is raised only because these are the lines the PR relocates and the new contract comment above them (`:138-142`, "every fallible worker-provisioning step must finish before task publication") reads as covering it. Note the follow-up fix cannot be a bare `try/catch`: detaching the worker would leave `FinishThreads()` (`ThreadPool.cpp:102-107`) dereferencing `end()` in a release build once the `assert` is compiled out, so it needs a registry that tolerates an unregistered worker. *(raised by czs007)*

xaxys:
Thanks. I reviewed both queue-duration findings together. They describe two different metric semantics rather than one implementation error.

I recommend keeping the current sampling point after acquiring `mutex_` and completing worker provisioning. The metric is named `internal_storage_pool_queue_duration_seconds`, and its help text describes task queue duration. Its clearest meaning is therefore approximately queue publication to execution start. Moving the sample above the lock would mix submit-side mutex contention and dynamic worker creation into queue residence. For example, a slow thread creation with no queue backlog would appear as high queue duration even though the task executes immediately after publication, which would make the metric less useful for diagnosing queue saturation or worker consumption.

The newer observation is still valid: because publication now occurs after the pool lock and provisioning, that pre-publication delay is not represented by any pool-specific metric. I view this as an observability gap, not a reason to overload the queue-duration histogram. If production monitoring needs that distinction, the clean follow-up is a separate `submit_duration` metric covering Submit entry through publication/return, while retaining queue duration for publication through execution. The old implementation was not strictly Submit-entry timing either, since it sampled after task binding and packaged-task allocation.

I also reviewed the metric-pointer race finding. It is formally possible through the public class API if `SetMetrics()` runs concurrently with `Submit()`, but the production lifecycle prevents that: `ThreadPools::GetThreadPool()` constructs the pool and calls `SetMetrics()` while holding its global unique lock, then publishes the first usable pool reference. `ThreadPools.cpp` is the only production caller and does not rebind metrics at runtime. I therefore do not recommend adding atomic pointer loads in this PR. A cleaner independent hardening would bind an immutable metrics bundle during construction, or make `SetMetrics()` explicitly initialization-only.

For this PR, my recommendation is no code change: preserve queue-publication-to-execution semantics, document the metric boundary in the PR discussion, and handle submit-side latency or immutable metric binding separately if operational evidence shows they are needed.

czs007:
> Also reported: 3 pre-existing hazards (real, but not introduced here — follow-up only, does **not** block this PR).

One sentence up front: this round re-confirms the earlier findings on `016575bf6027`, adds one same-class caller-side hazard, and leaves this PR's own findings to a single test-coverage gap.

## Re-review of `016575bf6027`

- **`internal/core/src/segcore/reduce/Reduce.cpp:220` — Submit() throwing leaves accepted tasks holding destroyed objects after unwinding: Confirmed.** Both reviewers re-walked the code and agree the draining guard is still installed only after the submit loop, so packaged-task futures for already-accepted work are destroyed without waiting.
- **`internal/core/src/segcore/memory_planner.cpp:538` — partial submission failure blocks workers forever on the bounded channel and leaks the process-level transient budget: Confirmed**, with one correction to the causal chain: the deadlock arises because the throw escapes *before* the callers' own drain/`WaitAllFutures` `catch` blocks, leaving no consumer, rather than because the channel is merely left unclosed.
- **`internal/core/src/storage/ThreadPool.h:148` — `unordered_map::operator[]` during worker registration can throw `bad_alloc` and trigger `std::terminate`: Confirmed.** The identical pattern remains in the unchanged `ThreadPool.cpp:73-76`, so this is pre-existing rather than a regression, and it stays low.
- **`internal/core/src/storage/ThreadPool.h:156` — submit-side lock wait now falls outside the `queue_duration` histogram: Confirmed on the literal claim, and explicitly non-blocking.** The sampling point did move after `mutex_` acquisition, but this is the deliberate design (the histogram now measures publication→execution); it is an accepted observability trade-off, not a correctness defect, and needs no change in this PR.

## Low

- **`internal/core/src/storage/ThreadPoolTest.cpp:142` — the worker-provisioning rejection path has no test coverage.** `SubmissionPreservesResultAndTaskException` asserts only the normal return value and task-thrown exception propagation; this PR's core semantic — that a failed worker provisioning cleanly rejects the submission (task never enters `work_queue_`, no future returned, exception visible) — has no assertion anywhere in the 153-line file. The PR description's argument for not injecting `std::thread` construction failure in this unit binary (no portable hook, unwillingness to add a test-only seam to a production header) is a reasonable trade-off. If you do want coverage, no production seam is needed: a gtest fork-style case (`EXPECT_EXIT`/child process) with `setrlimit(RLIMIT_NPROC)` can force `pthread_create` `EAGAIN` in the child, then assert that `Submit` throws and that `GetThreadNum()` and queue depth are unchanged. (raised by xiaocai2333, chyezh)

- **`internal/core/src/segcore/reduce/Reduce.cpp:212` — `FillPrimaryKey()` installs its draining `folly::makeGuard` only after the submit loop (`:220`), so a mid-loop `Submit()` failure unwinds while the guard does not yet exist.** Failure: a `num_segments_ > 1` search where MIDDLE-pool expansion hits the process thread limit (`RLIMIT_NPROC`/`EAGAIN`) and submission `k+1` throws `std::system_error` → the first `k` packaged-task futures are destroyed without blocking, the exception propagates to the `catch` at `search_result_export_c.cpp:1179` and destroys the stack-owned `OpContext` (`:1165`) and `ReduceHelper` (`:1166`), while background tasks keep calling `segment->FillPrimaryKeys(plan_, *search_result, op_ctx_)` on freed state and writing into a `SearchResult` still held by Go — segfault or silent memory corruption fully decoupled from its root cause. Fix is to move the guard ahead of the submit loop. Unrelated to this PR: `Reduce.cpp` is untouched by this diff, and on master `Submit()` throws on the same expansion failure with the same missing guard and the same discarded futures. (raised by xiaocai2333, chyezh)
- **`internal/core/src/segcore/memory_planner.cpp:530` — `LoadCellBatchAsync()` has no cleanup for partial submission; the throw escapes before the callers' `try`, so their drain/`WaitAllFutures` handlers never run.** `remaining` starts at `batches.size()` and only accepted tasks decrement it in `task_guard` (`:554-558`), and the throw site precedes the callers' `try` (`GroupChunkTranslator.cpp:365` vs `:385`; `ManifestGroupTranslator.cpp:475` vs `:501`). Failure: a storage-v2 column-group load split into `N > 1` batches with total cells exceeding `GetMaxThreadNum() * 1.5`, where load-pool expansion hits the thread limit and submission `j` throws → the callers' drain + `WaitAllFutures` cleanup is bypassed, leaving no consumer; the `j-1` accepted producers fill the bounded channel and block forever on `channel->push` (`:620/623`), never reaching `task_guard` (`Channel::close()` is itself a blocking push, `Channel.h:57-63`), starving that priority pool's subsequent segment loads with no error or timeout while the already-transferred `TransientMemoryBudget` is never returned. Fix inside `LoadCellBatchAsync()`: account for unsubmitted batches, drain the channel, wait for accepted tasks, release drained results, then rethrow. Unrelated to this PR: `memory_planner.cpp` and both translators are untouched by this diff, and master's `Submit()` breaks the loop the same way — this PR only shifts by one whether the throwing submission was enqueued, which does not change the deadlock conditions. (raised by xiaocai2333, chyezh)
- **`internal/core/src/segcore/reduce/Reduce.cpp:464` — `RefineDistances()` has the same pre-guard partial-submission hole**, creating its draining guard only at `:482`, after all submissions. Failure: a `num_segments_ > 1` search with global refine enabled (non-empty `placeholder_group_`) where MIDDLE-pool `std::thread` construction fails and submission `k+1` throws → the first `k` `RefineOneSegment` futures are destroyed non-blockingly, the `catch` at `search_result_export_c.cpp:1179` destroys the stack-owned `OpContext`/`ReduceHelper`, and running tasks keep using the dangling `op_ctx_` in `segment->CalcDistByIDs` (`:395`) and the captured `dense_blob` (`:389-390`) while writing `search_result->distances_`/`seg_offsets_` already returned to Go — use-after-free or silently corrupted distances/ordering. Move the guard before the loop. Unrelated to this PR: `Reduce.cpp` is untouched by this diff and the same conditions hold on master. (raised by chyezh, xiaocai2333)
- **`internal/core/src/storage/ThreadPool.h:150` — `threads_[t.get_id()] = std::move(t)` runs after the worker thread has started, so an allocation failure aborts the process.** Failure: under memory exhaustion the `std::thread` is created successfully but the `unordered_map` node allocation throws `std::bad_alloc` → the local `t` is still joinable, unwinding runs `~thread()` → `std::terminate()` → whole-process abort, i.e. exactly the opposite of the clean, catchable rejection this PR is built for. Worth noting because this is the fallible step whose "finish before publication" ordering the PR depends on, and its failure mode is not a catchable exception. Suggested follow-up: reserve the map node before constructing the `std::thread`, or provide a failure path that can safely stop and join the unregistered worker. Unrelated to this PR: these lines are a verbatim relocation of master's `ThreadPool.h:176-183` block to before `enqueue`, with identical statements and semantics, and the same pattern persists in the untouched `ThreadPool.cpp:73-76`. (raised by xiaocai2333, chyezh)

xaxys:
**Since another `ThreadPool` change merged into master, this PR is not nessesary now**

Current decision context after rebasing onto the ThreadPool changes from [#51403](https://github.com/milvus-io/milvus/pull/51403):

### Current master policy

- obtain the task future;
- enqueue the task first;
- attempt dynamic worker expansion afterward;
- catch worker-spawn failure and keep the accepted task queued;
- return the future so an existing worker can execute the task later.

This fixes the original ordinary failure case: `std::thread` construction failure itself no longer leaves an accepted task without its future.

### #51920 policy

- obtain the future;
- determine whether projected demand requires another worker;
- complete required worker provisioning before enqueue;
- if provisioning fails, throw while the task is still unpublished;
- otherwise enqueue and return the future.

**The divergence is now an acceptance-policy choice: current master accepts the task and treats expansion as best effort, while #51920 rejects a task whose required provisioning failed.**

### Remaining boundaries in the master implementation

1. The `LOG_WARN` calls executed after enqueue are theoretically fallible because formatting/allocation can throw under extreme memory pressure. That is a narrow remaining accepted-task-without-future window, independent of the normal `std::thread` exception that master now catches.
2. If all workers are blocked and a worker submits a child task to the same pool and synchronously waits for it, failed expansion could leave the child future waiting indefinitely.

The second point is structural rather than a current production bug. I audited the current production Submit call graph and found no same-pool nested synchronous submission. Existing nested load paths deliberately cross MIDDLE to HIGH/LOW pools, and the LOW decode path explicitly states that it must not be entered from a LOW-pool worker. Same-pool submit-and-wait should remain disallowed; future callers should use a different pool, a continuation, or return the future to a non-worker coordinator.

The upstream failure-injection test is also not a nested-submission example: an external test thread releases the blocked worker, which then consumes the queued task normally.

Related conflict resolution in [#51696](https://github.com/milvus-io/milvus/pull/51696): current master added `InspectStreamLoadInfo()` for V3 scalar-index resource estimation. The rebased PR preserves that lightweight path at 64 KiB and skips a separate magic GET, while normal `IndexEntryReader::Open()` validates magic and retains up to 256 KiB for actual entry reuse. The inspection still causes a separate estimation-time Range GET and destroys its reader afterward, so #51696 cannot eliminate or reuse that request.\n\nFor now I am holding #51920 open without further code changes. The remaining decision is whether Milvus wants the current best-effort expansion policy, with a small separate hardening for post-enqueue logging, or the stronger reject-before-publication policy in this PR.

czs007:

Adversarial review confirmed one high-severity finding introduced by this PR's change to the `Submit()` failure contract.

## High

- **`internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:2451`** (also `internal/core/src/segcore/reduce/Reduce.cpp:213` and `Reduce.cpp:464`) — This PR changes worker-spawn failure in `ThreadPool::Submit()` from being swallowed to throwing `std::system_error` (a contract now pinned by `ThreadPoolTest.cpp`), but all three call sites submit tasks in a loop with by-reference captures (`&fs`, `&field_ids_for_stats`, `&debug_key`, or `ReduceHelper::this`) and only construct their `futures_guard` drain guard *after* the loop completes. If thread creation fails mid-loop (pid/cgroup thread limits, `EAGAIN`, thread-stack allocation failure) after at least one task has been accepted → `Submit()` throws, stack unwinding destroys the futures vector and the referenced stack locals while earlier-accepted tasks are still running, producing a use-after-free (the `std::future` from a `packaged_task` does not block in its destructor). The fix does not require making batch submission atomic: hoist each drain guard above its submit loop, or use an RAII future collector that drains every previously accepted task on unwind. Since this hazard is armed by this PR's throw-on-spawn-failure contract, the caller-side guard fix should land before or together with this PR — recommend merging that separate fix first if it is kept as its own PR. *(raised by czs007, chyezh)*

PR #51925, opened 2026-07-28
fix: [3.0] synchronize delegator function runtime with collection schema (#51818)

aoiasd:
issue: #49716
pr: https://github.com/milvus-io/milvus/pull/51818

## Summary
Track the collection schema version applied to each shard delegator's FunctionRunner and IDF runtime. Catch up channel-local runtime when the shared Collection was advanced by another channel or load path, while serializing concurrent updates and preserving retry after failed updates.


liliu-z (commented):
/lgtm
/approve

Issue #50102, opened 2026-05-27
[Bug]: MixCoord active-standby failover panics with etcd auth enabled


hejiahang929:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version:v2.6.16
- Deployment mode :cluster
- MQ type:kafka
- OS: Ubuntu
- CPU/Memory: 8C/16G
```

### Current Behavior

```markdown
## Bug: standby MixCoord panics after promotion when etcd auth token expires

### Milvus version

v2.6.16

### Deployment mode

Kubernetes, non-Helm deployment.

MixCoord is started as `mixture` with:

```yaml
args:
  - run
  - mixture
  - -rootcoord
  - -querycoord
  - -datacoord
```

MixCoord active-standby is enabled:

```yaml
mixCoord:
  enableActiveStandby: true
```

Two MixCoord pods are running.

etcd auth is enabled. All Milvus components share the same config file and use the same etcd auth credentials.

### Problem

When the standby MixCoord pod has just started, deleting the active MixCoord pod works correctly. The standby pod is promoted to active successfully.

However, after the standby MixCoord pod has been running for a while, deleting the active MixCoord pod causes the standby pod to switch from standby to active, then panic and exit with code 134.

The failed pod can restart successfully afterwards, which suggests the etcd username/password config is correct and the issue is related to the expired etcd auth token held by the standby process.

### Relevant logs

```text
[INFO] ["try to register as ACTIVE mixcoord service..."]
[INFO] ["register ACTIVE mixcoord"] [revision=274616]
[INFO] ["serverName: mixcoord quit STANDBY mode, this node will become ACTIVE, ID: 145"]
[INFO] ["mixCoord switch from standby to active, activating"]
[INFO] ["RootCoord init successfully"]
[INFO] ["succeed to init sessions on etcd"]

panic: rpc error: code = Unauthenticated desc = etcdserver: invalid auth token

goroutine 2893 [running]:
github.com/milvus-io/milvus/internal/util/proxyutil.(*ProxyWatcher).startWatchEtcd(...)
    /go/src/github.com/milvus-io/milvus/internal/util/proxyutil/proxy_watcher.go:146
github.com/milvus-io/milvus/internal/util/proxyutil.(*ProxyWatcher).WatchProxy.func1()
    /go/src/github.com/milvus-io/milvus/internal/util/proxyutil/proxy_watcher.go:112
```

There are also etcd client warnings before the panic:

[...]

PR #51926, opened 2026-07-28
fix: [2.6] re-watch etcd on recoverable auth-token errors

xiaocai2333:
## What this PR does

Backports #50614 to the `2.6` branch.

With etcd auth enabled and MixCoord active-standby enabled, a long-running standby can hold an auth token that etcd has invalidated. When the standby is promoted, an established watch stream returns `Unauthenticated: invalid auth token`. The 2.6 watcher code only recovers from `ErrCompacted` and panics on other watch errors, causing the promoted MixCoord to exit.

This PR:

- Adds `etcd.IsRetriableWatchErr` for compacted revisions and recoverable etcd auth-token errors.
- Makes `ProxyWatcher` re-list sessions and establish a new watch with bounded retry instead of panicking on those errors.
- Applies the same recovery behavior to `sessionutil` watchers.
- Keeps permission errors, invalid credentials, and other non-recoverable errors fail-fast.
- Avoids panicking when re-watch fails because the owning context is already canceled.
- Includes the classification, re-watch, and compaction regression tests from #50614.

Re-listing uses the unary etcd client path, which refreshes an invalid auth token before the replacement watch stream is opened.

## Verification

Passed:

- `(cd pkg && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/etcd)`
- `git diff --check`
- `gofmt` check on all changed Go files

The following targeted watcher test was started locally but its first-time C++/Go dependency compilation was stopped due to local build time; CI will run it:

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 -p 1 ./internal/util/sessionutil ./internal/util/proxyutil -run 'TestWatcherHandleWatchResp|TestProxyManager_ErrCompacted'`

issue: #50102
pr: #50614

### DIFF
--- a/internal/util/sessionutil/session_util.go
+++ b/internal/util/sessionutil/session_util.go
@@ -791,7 +791,12 @@ type SessionEvent struct {
 }
 
 type sessionWatcher struct {
-	s         *Session
+	s *Session
+	// ctx is the watcher-owned context (child of the session context), canceled
+	// by Stop(). Every watcher-side etcd call — the watch itself, the re-watch
+	// retry and its GetSessions — must use it, so Stop() interrupts in-flight
+	// recovery instead of waiting on the session context.
+	ctx       context.Context
 	cancel    context.CancelFunc
 	rch       clientv3.WatchChan
 	eventCh   chan *SessionEvent
@@ -859,9 +864,10 @@ func (s *Session) WatchServices(prefix string, revision int64, rewatch Rewatch) 
 	ctx, cancel := context.WithCancel(s.ctx)
 	w := &sessionWatcher{
 		s:        s,
+		ctx:      ctx,
 		cancel:   cancel,
 		eventCh:  make(chan *SessionEvent, 100),
-		rch:      s.etcdCli.Watch(s.ctx, path.Join(s.metaRoot, DefaultServiceRoot, prefix), clientv3.WithPrefix(), clientv3.WithPrevKV(), clientv3.WithRev(revision)),
+		rch:      s.etcdCli.Watch(ctx, path.Join(s.metaRoot, DefaultServiceRoot, prefix), clientv3.WithPrefix(), clientv3.WithPrevKV(), clientv3.WithRev(revision)),
 		prefix:   prefix,
 		rewatch:  rewatch,
 		validate: func(s *Session) bool { return true },
@@ -880,9 +886,10 @@ func (s *Session) WatchServicesWithVersionRange(prefix string, r semver.Range, r
 	ctx, cancel := context.WithCancel(s.ctx)
 	w := &sessionWatcher{
 		s:        s,
+		ctx:      ctx,
 		cancel:   cancel,
 		eventCh:  make(chan *SessionEvent, 100),
-		rch:      s.etcdCli.Watch(s.ctx, path.Join(s.metaRoot, DefaultServiceRoot, prefix), clientv3.WithPrefix(), clientv3.WithPrevKV(), clientv3.WithRev(revision)),
+		rch:      s.etcdCli.Watch(ctx, path.Join(s.metaRoot, DefaultServiceRoot, prefix), clientv3.WithPrefix(), clientv3.WithPrevKV(), clientv3.WithRev(revision)),
 		prefix:   prefix,
 		rewatch:  rewatch,
 		validate: func(s *Session) bool { return r(s.Version) },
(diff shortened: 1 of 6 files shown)


internal/util/sessionutil/session_util.go
```diff
@@ -942,32 +951,43 @@ func (w *sessionWatcher) handleWatchResponse(wresp clientv3.WatchResponse) {
...
+		log.Ctx(w.s.ctx).Warn("Watch service found error", zap.Error(err))
 		w.closeEventCh()
 		return err
 	}
 
+	// Re-establish the watch with a bounded retry. handleReWatch issues a unary
+	// request first, which refreshes the etcd auth token via clientv3's unary
+	// retry interceptor. Only keep retrying transient errors (e.g. a still-stale
+	// auth token); a non-transient failure aborts immediately.
+	if reErr := retry.Do(w.s.ctx, w.handleReWatch, retry.RetryErr(etcd.IsRetriableWatchErr)); reErr != nil {
```
xiaofan-luan: **[P2] Bind the re-watch retry to the watcher lifetime**

`sessionWatcher.Stop()` cancels the watcher child context, but this retry uses `w.s.ctx`; `handleReWatch` also uses `w.s.ctx` for both `GetSessions` and the replacement watch. QueryCoord and DataCoord call watcher `Stop()` before canceling the session context. If shutdown races with auth recovery, `Stop()` can therefore wait for these retries or the unbounded `GetSessions`, and recovery can establish another watch after the watcher has already been stopped.

Please retain the watcher-owned context and use it for retry, Get, Watch, and the context-done check. The ProxyWatcher retry at `proxy_watcher.go:155` has the same lifetime mismatch: `Stop()` closes `closeCh`, while the retry only observes the component context.

pkg/util/etcd/etcd_util.go
```diff
@@ -39,6 +39,43 @@ import (
...
+// source of Unauthenticated, so matching the sentinel is both sufficient and
+// narrower than trusting the code.
+//
+// Callers should wrap the re-watch in a bounded retry (retry.Attempts) so a
+// genuinely persistent failure does not loop forever.
+func IsRetriableWatchErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	return errors.Is(err, rpctypes.ErrCompacted) ||
```
xiaofan-luan: **[P1] Match the error shape returned by an etcd watch cancellation**

With etcd v3.5.23, `serverWatchStream` places `rpctypes.ErrGRPCInvalidAuthToken.Error()` — the full `rpc error: code = Unauthenticated ...` string — in `WatchResponse.CancelReason`. clientv3 then runs `v3rpc.Error(errors.New(resp.CancelReason))`; this remains a plain error because the v3rpc lookup keys contain only the status descriptions. Consequently `errors.Is(err, rpctypes.ErrInvalidAuthToken)` is false for this production path, so both watchers still take the non-retriable panic path.

The current unit test injects the sentinel directly and does not cover this transport shape. Please add a cancel-reason-shaped test and normalize or match the actual clientv3 error before classifying it.

xiaocai2333: Fixed in 982eaa6221. Verified against etcd v3.5.23 source: the server puts `ErrGRPCInvalidAuthToken.Error()` (the full `rpc error: code = Unauthenticated desc = ...` string) into `CancelReason` (server `v3rpc/watch.go:275`), and clientv3 rebuilds it with `v3rpc.Error(errors.New(resp.CancelReason))` (`watch.go:461`) whose lookup table is keyed by the bare description only — so `errors.Is` indeed never matched on this path.

`IsRetriableWatchErr` now falls back to matching the sentinel descriptions inside the error message, covering both the direct CancelReason shape and the substream shape that nests it inside a `FailedPrecondition` status (`watch.go:131`). Permission-denied cancel reasons keep the same transport shape and remain non-retriable. Added cancel-reason-shaped test cases for all three auth sentinels, the nested substream shape, and a permission-denied negative case.

internal/util/sessionutil/session_util.go
xiaocai2333: Fixed in 982eaa6221. `sessionWatcher` now keeps the watcher-owned context (child of the session ctx, canceled by `Stop()`) and uses it for the initial watch, the re-watch retry, `GetSessions`, the replacement watch, and the teardown check — so `Stop()` interrupts in-flight recovery instead of waiting on the session ctx, and the shutdown race no longer panics (the canceled watcher ctx makes teardown quiet).

For `ProxyWatcher`, `closeCh` is bridged into a child context that gates the retry and its backoff sleeps, each attempt checks `IsClosed` first, and a closed watcher is treated as quiet teardown rather than the panic path. `WatchProxy` itself still runs on the component ctx so a successfully re-established watch isn't torn down when the bridge is released.

Added regression tests on both sides: re-watch attempts keep failing with a retriable auth error while `Stop()` is called concurrently — the retry now aborts within ~200ms without panicking (previously it would sleep through the full backoff budget and then panic).

czs007:

Both independent reviewers converged on the same verdict in the final round, with no additional findings raised on either side.

Issue #51930, opened 2026-07-28
[Bug]: Milvus may crash when Sonic JIT module registration races with Go plugin loading


congqixia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 3.0 beta & 2.6.21
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

Milvus can terminate with an unrecoverable Go runtime error when Sonic JIT
  module registration runs concurrently with `plugin.Open`:

  ```text
  runtime: plugin has empty pluginpath
  fatal error: runtime: plugin has empty pluginpath

  Serializing Milvus call sites with the mutex in hookutil.LoadPlugin does not
  fully prevent the crash. That mutex only serializes calls made through Milvus;
  it does not synchronize plugin.Open with Sonic's concurrent runtime
  moduledata registration.

  Sonic's loader appends synthetic JIT runtime.moduledata entries whose
  pluginpath is empty. In its default build path, this operation does not hold
  the Go standard library's plugin.pluginsMu.

  When the operation overlaps with the dlopen/lastmoduleinit section of
  plugin.Open, the Go runtime may treat Sonic's synthetic module as the newly
  loaded plugin and fail because its pluginpath is empty.

  The failure is raised through runtime.throw, so it cannot be recovered by
  application code.

### Expected Behavior

Concurrent Sonic JIT compilation and Go plugin loading should not corrupt the runtime module list or terminate the Milvus process.

### Steps To Reproduce

```markdown
1. Build the Milvus executable and Go plugins with Go 1.26.4 and the following
     build tags:

     dynamic,sonic,with_jemalloc

  2. Start multiple goroutines that trigger Sonic JIT compilation.
  3. Concurrently call plugin.Open to load a Go plugin.
  4. Repeat the operation under load.
  5. Observe an intermittent fatal runtime error:

     runtime: plugin has empty pluginpath

  A focused reproducer using concurrent Sonic JIT goroutines and plugin.Open
  can reproduce the same failure.
```

### Milvus Log

  runtime: plugin has empty pluginpath
  fatal error: runtime: plugin has empty pluginpath

  The affected executable and plugins all report go1.26.4, so this is not
[...]

PR #51931, opened 2026-07-28
fix: synchronize Sonic JIT registration with Go plugin loading

congqixia:
Enable Sonic's bytedance_tango build path for the Milvus CPU and GPU executables. This makes Sonic runtime module registration use the same plugin.pluginsMu critical section as plugin.Open.

Add -checklinkname=0 because Go 1.26 otherwise rejects Sonic loader's linkname reference to the private plugin.pluginsMu symbol.

issue: #51930

Tested:
- built a focused host and plugin with Tango synchronization enabled
- passed 50 concurrent Sonic JIT and plugin.Open stress runs
- passed hookutil tests with the required dynamic,test flags
- verified CPU and GPU Makefile command expansion


scripts/verify_sonic_plugin_build.sh
```diff
@@ -0,0 +1,152 @@
...
+    fi
+    if [[ $(build_setting "${plugin}" "GOOS") != "${host_goos}" ]]; then
+        echo "${plugin}: GOOS differs from host ${host_goos}" >&2
+        exit 1
+    fi
+    if [[ $(build_setting "${plugin}" "GOARCH") != "${host_goarch}" ]]; then
+        echo "${plugin}: GOARCH differs from host ${host_goarch}" >&2
+        exit 1
+    fi
+    if [[ $(build_setting "${plugin}" "GOAMD64") != "${host_goamd64}" ]]; then
```
liliu-z: The plugin/host comparison checks GOAMD64 but never GOARM64. On the arm64 images Milvus ships, GOAMD64 is absent from buildinfo so that comparison is a no-op, while GOARM64 (confirmed present as `build GOARM64=v8.0`) goes entirely unchecked — a microarch-skewed plugin (e.g. v8.0 host, v9.0 plugin) passes verification cleanly. Since the verifier deliberately guards GOAMD64 as a microarch check, add the symmetric GOARM64 comparison.

scripts/verify_sonic_plugin_build.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly REQUIRED_LDFLAG="-checklinkname=0"
+readonly -a REQUIRED_TAGS=(dynamic sonic with_jemalloc bytedance_tango)
```
liliu-z: REQUIRED_TAGS is a fixed list of four tags and the verifier only checks that each is present — it never compares the plugin's tag set against the host's. The GPU host is built with an extra `cuda` tag (Makefile:129), and `cuda` selects different source in packages the plugin ABI transitively imports (paramtable/autoindex_param_cuda.go, hardware/gpu_mem_info_cuda.go, gated by `//go:build cuda`). A plugin built without `cuda` therefore passes verification yet diverges in the paramtable package hash, so plugin.Open rejects it at runtime — a false PASS. Compare the host's actual compatibility-affecting tags instead of a hardcoded subset.

scripts/verify_sonic_plugin_build.sh
```diff
@@ -0,0 +1,152 @@
...
+
+    tags=$(build_setting "${artifact}" "-tags")
+    for required_tag in "${REQUIRED_TAGS[@]}"; do
+        if ! has_csv_value "${tags}" "${required_tag}"; then
+            echo "${artifact}: missing build tag ${required_tag}; tags=${tags:-<empty>}" >&2
+            return 1
+        fi
+    done
+
+    ldflags=$(build_setting "${artifact}" "-ldflags")
```
liliu-z: The ldflags check requires the recorded `-ldflags` build setting to contain `-checklinkname=0`, but Go omits `-ldflags` from buildinfo entirely under `-trimpath`. Any trimpath'd artifact reads back empty and is wrongly rejected. Milvus's own Makefile doesn't use -trimpath, so this is a false-reject on external/reproducible plugin builds (where -trimpath is common) rather than a runtime hazard; verify via a signal that survives -trimpath, or document that trimpath'd plugins can't be ldflags-verified.

scripts/verify_sonic_plugin_build.sh
```diff
@@ -0,0 +1,152 @@
...
+if [[ $(build_setting "${host}" "-buildmode") != "exe" ]]; then
+    echo "${host}: expected -buildmode=exe" >&2
+    exit 1
+fi
+validate_common_settings "${host}"
+
+host_go_version=$(go_version "${host}")
+host_goos=$(build_setting "${host}" "GOOS")
+host_goarch=$(build_setting "${host}" "GOARCH")
+host_goamd64=$(build_setting "${host}" "GOAMD64")
```
liliu-z: The verifier captures and compares GOOS/GOARCH/GOAMD64 and the Go version, but never GOEXPERIMENT. A FIPS build sets `GOEXPERIMENT=boringcrypto` on the host (Makefile:70, MILVUS_FIPS_ENABLED=ON), which is recorded in buildinfo and changes the crypto package ABI. A plugin built without boringcrypto passes every check here yet diverges in crypto package hashes, so plugin.Open rejects it at runtime — the same fixed-subset gap as the GOAMD64/tags checks, on a security-sensitive axis. Compare GOEXPERIMENT alongside the other target settings.

Makefile
```diff
@@ -14,7 +14,12 @@ PWD 	  := $(shell pwd)
 GOPATH	:= $(shell $(GO) env GOPATH)
 SHELL 	:= /bin/bash
 OBJPREFIX := "github.com/milvus-io/milvus/cmd/milvus"
-MILVUS_GO_BUILD_TAGS := "dynamic,sonic,with_jemalloc"
+# Temporary synchronization workaround for the race between Go plugin loading
+# and github.com/bytedance/sonic/loader registering JIT moduledata. All Go
+# plugins loaded by Milvus must use the same tag and linker flag.
+SONIC_PLUGIN_SYNC_TAG := bytedance_tango
+SONIC_PLUGIN_SYNC_LDFLAG := -checklinkname=0
+MILVUS_GO_BUILD_TAGS := dynamic,sonic,with_jemalloc,$(SONIC_PLUGIN_SYNC_TAG)
```
liliu-z, line 22: Adding `bytedance_tango` to the host build tags swaps `sonic/loader`'s lock-free `register.go` for `register_tango.go`, changing the compiled source (and pkghash) of every package that transitively imports sonic. Go's `plugin.Open` rejects a plugin whose shared-package fingerprints differ from the host, so pre-built Sonic-using plugins (cipher, hook) fail to load until rebuilt with the identical `bytedance_tango` tag and `-checklinkname=0` flag. That load failure crashes startup: `InitOnceCipher` panics unconditionally (internal/util/hookutil/cipher.go:448) and `InitOnceHook` panics whenever `common.panicWhenPluginFail` is true, which is the default (internal/util/hookutil/hook.go:126). The build-parity verifier added in the first commit was removed in the second, so nothing enforces or documents the required host↔plugin flag parity.

congqixia: Thanks, this is a valid compatibility concern.

  The Go plugins, including `hook.so`, are built and released through a separate plugin pipeline. After this Milvus change lands, all plugin artifacts shipped with the updated Milvus image will be rebuilt with the same Go toolchain
  and build contract:

  ```text
  -tags dynamic,sonic,with_jemalloc,bytedance_tango
  -ldflags=-checklinkname=0

  The updated Milvus binary will not be released together with existing pre-Tango plugin artifacts. Plugin loading and startup will be verified using the rebuilt artifacts before release.

  The verifier was removed from this PR because it was a standalone script that was not integrated into the Milvus build or CI and cannot access artifacts produced by the external plugin pipeline. Compatibility validation will instead
  be performed where those final plugin artifacts are built and assembled.

liliu-z (commented):
/lgtm
/approve

Issue #41675, opened 2025-05-07
[Feature]: Support Orderby


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Many users have requested support for result ordering using ORDER BY. In Milvus 3.0, we introduce incremental enhancements to address this need across different query types.

Search: By default, results are ordered by similarity score (i.e., relevance). To support ORDER BY on a scalar field, Milvus retrieves the target field and performs result reordering at the Proxy level. This approach is lightweight in complexity and does not require changes to the search pipeline.

Query: Results are currently ordered by primary key. Supporting ORDER BY on a scalar field follows the same pattern—retrieve the relevant field and sort at the Proxy. Since queries return all results in a single RPC call, memory pressure is not a concern.

Range Search Iterator: Results are ordered by similarity by default. Arbitrary scalar ORDER BY is not supported for now. However, we plan to support ORDER BY RANDOM(). This mode enables fast iteration by returning data in segment order, without worrying about maintaining a strict global order.

Query Iterator: Like regular queries, results are ordered by primary key. Arbitrary ORDER BY is not planned in this mode either, but ORDER BY RANDOM() will be supported to allow efficient, segment-wise iteration.

SunMeng657:
This ORDER BY feature for search/query is exactly what I need—it’ll let me filter and analyze data with precision, streamlining my workflow big-time.
Closely watching, eager for release. Thanks for pushing Milvus forward!

Issue #36380, opened 2024-09-20
[Feature]: 希望在 query 查询是增加根据标量字段排序的能力


yuanjianchen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

在使用 milvus 时会有一些场景只有标量查询，希望在处理标量查询时能提供排序策略，例如通过时间排序

### Describe the solution you'd like.

表达式中增加 order by  [fieldName] [direction] 这种语法

xiaofan-luan:
/assign @MrPresent-Han 

query with order by should not be hard to implement. Iterator might be harder

MrPresent-Han:
I will handle this

PR #51937, opened 2026-07-28
enhance: add first-class orderByFields/groupByFields params to RESTful v2

MrPresent-Han:
Search order-by previously worked over REST only through the undeclared `searchParams.order_by_fields` passthrough, and query GROUP BY aggregation was unreachable because the query handler forwards a fixed whitelist of query params.

This PR adds first-class RESTful v2 request fields:

- **`SearchReqV2.orderByFields`** — forwarded as the `order_by_fields` search param the proxy already consumes. Rejected when combined with `searchAggregation` (the aggregation pipeline ignores order-by), and when specified together with the legacy `searchParams.order_by_fields` passthrough (ambiguous: the first key silently wins on the proxy side otherwise). The legacy passthrough alone keeps working.
- **`QueryReqV2.groupByFields`** — forwarded as the `group_by_fields` query param, enabling GROUP BY aggregation over REST together with aggregation expressions (`count(*)`, `sum(x)`, ...) in `outputFields`.

Verified by httpserver unit tests covering param forwarding and all rejection paths; proxy-side order-by / group-by aggregation execution is covered by existing proxy tests. REST e2e is left to the CI e2e suites.

related: #41675, #36380

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -1377,9 +1377,18 @@ func (h *HandlersV2) addCollectionStructField(ctx context.Context, c *gin.Contex
 	return resp, err
 }
 
-// copy from internal/proxy/task_query.go
-func matchCountRule(outputs []string) bool {
-	return len(outputs) == 1 && strings.ToLower(strings.TrimSpace(outputs[0])) == "count(*)"
+// outputsContainCountStar reports whether any output field is the count(*)
+// aggregate, mirroring the Proxy's count(*) detection: a global (no GROUP BY)
+// query carrying count(*) plus a limit is rejected as meaningless pagination,
+// so REST must not attach its synthetic default limit to such a request —
+// including mixed aggregates like ["count(*)", "sum(x)"].
+func outputsContainCountStar(outputs []string) bool {
+	for _, o := range outputs {
+		if strings.EqualFold(strings.TrimSpace(o), "count(*)") {
+			return true
+		}
+	}
+	return false
 }
 
 func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbName string) (interface{}, error) {
@@ -1408,14 +1417,26 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 	}
 	req.ExprTemplateValues = generateExpressionTemplate(httpReq.ExprParams)
 	c.Set(ContextRequest, req)
+	if len(httpReq.GroupByFields) > 0 && len(httpReq.OrderByFields) > 0 {
+		err := merr.WrapErrParameterInvalidMsg("orderByFields and groupByFields cannot be used simultaneously")
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
+	}
 	if httpReq.Offset > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
 	}
-	if httpReq.Limit > 0 && !matchCountRule(httpReq.OutputFields) {
+	// REST applies a default limit, but Proxy rejects pagination for a global
+	// query containing count(*). A grouped count(*) uses limit/offset to
+	// paginate groups, so its limit must still be forwarded.
+	isGlobalCount := len(httpReq.GroupByFields) == 0 && outputsContainCountStar(httpReq.OutputFields)
+	if httpReq.Limit > 0 && !isGlobalCount {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.LimitKey, Value: strconv.FormatInt(int64(httpReq.Limit), 10)})
 	}
 	if len(httpReq.OrderByFields) > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OrderByFieldsKey, Value: strings.Join(httpReq.OrderByFields, ",")})
+	}
+	if len(httpReq.GroupByFields) > 0 {
+		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.GroupByFieldsKey, Value: strings.Join(httpReq.GroupByFields, ",")})
 	}
 	resp, err := wrapperProxyWithLimit(ctx, c, req, h.checkAuth, false, "/milvus.proto.milvus.MilvusService/Query", true, h.proxy, func(reqCtx context.Context, req any) (interface{}, error) {
 		return h.proxy.Query(reqCtx, req.(*milvuspb.QueryRequest))
@@ -1904,12 +1925,30 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
+		if len(httpReq.OrderByFields) > 0 {
+			err := merr.WrapErrParameterInvalidMsg("orderByFields and searchAggregation cannot be used simultaneously")
+			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+			return nil, err
+		}
+		if searchParamsContainAny(httpReq.SearchParams, proxy.OrderByFieldsKey) {
+			err := merr.WrapErrParameterInvalidMsg("searchParams.order_by_fields and searchAggregation cannot be used simultaneously")
+			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+			return nil, err
+		}
 		req.SearchAggregation, err = convertSearchAggregationReq(httpReq.SearchAggregation)
 		if err != nil {
 			mlog.Warn(ctx, "high level restful api, convert SearchAggregation failed", mlog.Err(err))
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
+	}
+
+	// The legacy searchParams.order_by_fields passthrough would silently shadow the
+	// dedicated field (first key wins on the proxy side), so reject the ambiguity.
+	if len(httpReq.OrderByFields) > 0 && searchParamsContainAny(httpReq.SearchParams, proxy.OrderByFieldsKey) {
+		err := merr.WrapErrParameterInvalidMsg("ambiguous order by: use either orderByFields or searchParams.order_by_fields, not both")
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
 	}
 
 	searchParams, err := generateSearchParams(httpReq.SearchParams)
@@ -1929,6 +1968,9 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 	if httpReq.GroupByField != "" && httpReq.GroupSize > 0 {
 		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamGroupSize, Value: strconv.FormatInt(int64(httpReq.GroupSize), 10)})
 		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamStrictGroupSize, Value: strconv.FormatBool(httpReq.StrictGroupSize)})
+	}
+	if len(httpReq.OrderByFields) > 0 {
+		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: proxy.OrderByFieldsKey, Value: strings.Join(httpReq.OrderByFields, ",")})
 	}
 	if len(httpReq.FunctionScore.Functions) != 0 {
 		if req.FunctionScore, err = genFunctionScore(ctx, &httpReq.FunctionScore); err != nil {
--- a/internal/proxy/util.go
+++ b/internal/proxy/util.go
@@ -3250,7 +3250,7 @@ func reconstructStructFieldData(
 	outputFields []string,
 	schema *schemapb.CollectionSchema,
 ) ([]*schemapb.FieldData, []string) {
-	if len(outputFields) == 1 && outputFields[0] == "count(*)" {
+	if len(fieldsData) == 0 || (len(outputFields) == 1 && outputFields[0] == "count(*)") {
 		return fieldsData, outputFields
 	}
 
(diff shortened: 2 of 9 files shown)


internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1411,12 +1411,19 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 	if httpReq.Offset > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
 	}
-	if httpReq.Limit > 0 && !matchCountRule(httpReq.OutputFields) {
+	// REST applies a default limit, but Proxy rejects pagination for a global
+	// count(*). A grouped count(*) uses limit/offset to paginate groups, so its
+	// limit must still be forwarded.
+	isGlobalCountOnly := len(httpReq.GroupByFields) == 0 && matchCountRule(httpReq.OutputFields)
```
liliu-z: REST's matchCountRule only recognizes a lone count(*) output, so a global (no groupByFields) query that contains count(*) alongside another aggregate — e.g. outputFields: ["count(*)", "sum(x)"] — fails isGlobalCountOnly, inherits the default Limit: 100, and forwards it. The proxy then rejects the request at task_query.go:836, whose guard keys on hasCountStar() (true for any count(*) aggregate), returning "count entities with pagination is not allowed" (code 1100) even though the user never requested pagination. Suppress the default limit whenever a global query contains a count(*) aggregate, not only when count(*) is the sole output field, so these multi-aggregate global queries match the proxy's hasCountStar rejection and are admitted. (Pre-existing edge, but on the predicate this PR rewrote; workaround is an explicit limit:0.)

MrPresent-Han: Fixed in d6f2bfcef6. Replaced `matchCountRule` (which only matched a lone `count(*)`) with `outputsContainCountStar`, so REST now suppresses its synthetic default limit for **any** global (no groupByFields) query whose outputFields contain `count(*)` — including mixed aggregates like `["count(*)", "sum(x)"]` — matching the proxy's `hasCountStar()` guard at task_query.go. Grouped `count(*)` still forwards limit/offset to paginate groups.

Added regression coverage in `TestQueryGroupByFields`: a global `["count(*)", "word_count"]` asserts no limit is forwarded, and a grouped `["count(*)", "word_count"]` + limit asserts the limit is forwarded. The TestDML mock that models the proxy rejection was switched to the same helper so it now mirrors `hasCountStar` semantics. Thanks for the precise catch.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1411,12 +1406,15 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
...
 	}
-	if httpReq.Limit > 0 && !matchCountRule(httpReq.OutputFields) {
+	if httpReq.Limit > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.LimitKey, Value: strconv.FormatInt(int64(httpReq.Limit), 10)})
 	}
 	if len(httpReq.OrderByFields) > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OrderByFieldsKey, Value: strings.Join(httpReq.OrderByFields, ",")})
 	}
+	if len(httpReq.GroupByFields) > 0 {
+		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.GroupByFieldsKey, Value: strings.Join(httpReq.GroupByFields, ",")})
```
buqian-zilliz, line 1439: **[P1] Do not truncate groups before the cross-shard merge**

`groupByFields` now makes grouped aggregation reachable through REST, but the REST default `limit=100` is forwarded as the QueryNode `topK`. For the no-ORDER-BY group pipeline, each QueryNode passes that value to `GroupAggReducer` and stops admitting new groups before Proxy performs the global merge. If the same group spans shards and different shards retain different local groups, the returned `count` or `sum` is silently partial (with `limit=1`, one shard can keep A while another keeps B). Avoid shard-local truncation for unordered grouped aggregates; apply limit/offset only after all partial aggregates are merged, or use a merge-safe distributed top-k protocol.

internal/proxy/util.go
```diff
@@ -3250,10 +3250,6 @@ func reconstructStructFieldData(
 	outputFields []string,
 	schema *schemapb.CollectionSchema,
 ) ([]*schemapb.FieldData, []string) {
-	if len(outputFields) == 1 && outputFields[0] == "count(*)" {
```
buqian-zilliz, line 3253: **[P2] Preserve aggregate output fields after an empty page**

With global count pagination now allowed, `limit=1, offset=1` legitimately slices away the sole aggregate row. Aggregation requests carry no `OutputFieldsId`, so the pipeline returns empty `FieldsData`; for a collection that has any `StructArrayFields`, removing this fast path causes `reconstructStructFieldData` to rebuild `OutputFields` from the empty data slice and return `[]` instead of the requested `["count(*)"]`. Preserve `outputFields` whenever `fieldsData` is empty, or retain an aggregation-specific fast path.

internal/proxy/search_util.go
```diff
@@ -177,7 +177,7 @@ func parseOrderByFields(searchParamsPair []*commonpb.KeyValuePair, schema *schem
 	}
 
 	var orderByFields []OrderByField
-	pairs := strings.Split(orderByStr, ",")
+	pairs := splitTopLevel(orderByStr, ',', -1)
```
buqian-zilliz: **[P2] Escape GJSON metacharacters in newly accepted JSON keys**

`splitTopLevel` now accepts quoted separator keys such as `metadata["[a,b]"]:asc`, but the downstream lookup does not preserve that property name literally: the JSON pointer `/[a,b]` becomes the GJSON path `[a,b]`, which GJSON interprets as a multipath subselector. The sorter therefore reads `a` and `b` (or missing values) instead of the `"[a,b]"` property and silently returns the wrong order. Either reject keys that cannot be represented safely or traverse the JSON pointer literally/escape every GJSON metacharacter before accepting these paths.

enhance_restful_0730.md
```diff
@@ -0,0 +1,125 @@
+# RESTful orderByFields / groupByFields review notes — 2026-07-30
```
buqian-zilliz: **[P3] Remove the committed review scratchpad**

`enhance_restful_0730.md` is a dated internal review/worktree scratchpad rather than durable product documentation. It contains obsolete proposals, "current worktree" state, and test instructions that will become stale immediately. Remove it from the PR; if any behavior needs permanent documentation, move a cleaned, stable description into the established documentation area.

internal/proxy/task_query.go
```diff
@@ -831,11 +818,6 @@ func (t *queryTask) PreExecute(ctx context.Context) error {
 		}
 	}
 
-	// count(*) without GROUP BY is a single-value result, pagination is meaningless.
-	// But count(*) with GROUP BY + limit is valid (limits the number of groups returned).
-	if t.hasCountStar() && t.queryParams.limit != typeutil.Unlimited && len(t.GetGroupByFieldIds()) == 0 {
-		return merr.WrapErrParameterInvalidMsg("count entities with pagination is not allowed")
-	}
 	t.plan.Namespace = namespaceForPlan(t.schema.CollectionSchema, t.request.Namespace)
```
liliu-z, line 839: Removing the count-pagination guard lets count(*) run with pagination; when offset skips the single aggregate row the proxy returns empty FieldsData (the added Python test asserts res == []). The Go client v3 then rejects that response: parseSearchResult takes the len(fieldDataList)==0 branch and, for a collection without a dynamic field, returns "extra output fields [count(*)] found and result does not contain dynamic field" (read.go:305), because the count(*) output field matches no schema field and there is no dynamic column to absorb it. So count(*) with a row-skipping offset+limit — previously rejected cleanly by the proxy — now surfaces an opaque client-side decode error instead of a count or a clean validation error.

liliu-z, line 839: This PR removes the guard that rejected count(*) combined with pagination (the 'count entities with pagination is not allowed' check) — a user-visible change to count/pagination semantics — but the PR body describes only orderByFields/groupByFields and never mentions it. Per the repo's disclose-behavior-changes convention, call the guard removal out in the PR body so reviewers and users are not surprised (it also introduces the Go-client decode regression noted separately). The 'missing linked issue' concern is weaker since the PR does link related issues.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: Rebuilt on current master in 2e91ec3259. The fix remains: REST uses `outputsContainCountStar`, so its synthetic default limit is omitted for any global output containing `count(*)`, including `["count(*)", "sum(x)"]`; grouped counts still forward limit/offset to paginate groups. The regression now uses the valid mixed aggregate `["count(*)", "sum(word_count)"]` and checks both cases.

internal/proxy/task_query.go
MrPresent-Han: Agreed. The rebuild restores the existing Proxy count-pagination guard unchanged: `internal/proxy/task_query.go` is no longer modified, so global `count(*)` with limit/offset remains rejected as before and cannot reach the Go-client empty-result decode path. The empty-FieldsData handling retained in `reconstructStructFieldData` is an independent defensive fix for legitimate empty aggregation pages.

MrPresent-Han: Rebuilt on current master in 2e91ec3259. This branch no longer removes or changes the Proxy count-pagination guard, so there is no count/pagination semantic change to disclose. The commit is limited to REST `groupByFields` / Search `orderByFields` support and the correctness guards required by those newly exposed paths.

internal/proxy/util.go
MrPresent-Han: Fixed in 2e91ec3259. `reconstructStructFieldData` now returns the requested `outputFields` unchanged whenever `FieldsData` is empty, so StructArray schema reconstruction cannot erase aggregate metadata. The rebuild also restores the pre-existing global `count(*)` pagination guard, but this empty-page preservation remains needed independently. Added a regression asserting an empty aggregate page keeps `sum(price)` in OutputFields.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: Fixed in 2e91ec3259. The no-ORDER-BY group reduce pipeline now always constructs `DeduplicateByGroups` with `groupLimit=-1` on both QueryNode and Delegator. The Proxy `MilvusAggReducer` remains the first layer that applies the user limit after all partial groups have merged. The regression passes topK=2 through two partial results and verifies that all three groups remain, with the split `blue` group fully accumulated. REST explicitly rejects the new `groupByFields + orderByFields` combination because it would need a separate merge-safe distributed ordered top-k design.

internal/util/queryutil/pipeline_builders.go
```diff
@@ -73,10 +73,12 @@ func buildOrderByReducePipeline(name string, schema *schemapb.CollectionSchema,
...
 
-// buildGroupByReducePipeline: [DeduplicateByGroups(topK)] → output
-func buildGroupByReducePipeline(name string, schema *schemapb.CollectionSchema, topK int64, groupByFieldIDs []int64, aggregates []*planpb.Aggregate) *Pipeline {
+// buildGroupByReducePipeline: [DeduplicateByGroups(unlimited)] → output.
+// QueryNode and Delegator must retain every partial group so Proxy can merge
+// all contributions before it applies the user-visible limit and offset.
+func buildGroupByReducePipeline(name string, schema *schemapb.CollectionSchema, groupByFieldIDs []int64, aggregates []*planpb.Aggregate) *Pipeline {
 	b := NewPipelineBuilder(name)
-	b.Add(OpReduceByGroups, pin(), pout(), NewDeduplicateByGroupsOperator(schema, groupByFieldIDs, aggregates, topK))
+	b.Add(OpReduceByGroups, pin(), pout(), NewDeduplicateByGroupsOperator(schema, groupByFieldIDs, aggregates, -1))
```
buqian-zilliz, line 81: **[P1] Enforce the output-size budget on all-groups reduction**

Passing `-1` here makes every QueryNode and Delegator retain and serialize every group before Proxy applies the requested page. `BuildQueryReducePipeline` receives `maxOutputSize` (100 MB by default), and the sibling reducers enforce it, but this grouped branch discards it; `GroupAggReducer` only caps the group count at 100000 and VarChar grouping keys are allowed. A `limit=1` query over many wide groups can therefore allocate and transfer far beyond the configured output quota. Keep the all-groups correctness fix, but enforce an intermediate byte/memory budget or use a merge-safe bounded strategy.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1408,15 +1417,27 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 	}
 	req.ExprTemplateValues = generateExpressionTemplate(httpReq.ExprParams)
 	c.Set(ContextRequest, req)
+	if len(httpReq.GroupByFields) > 0 && len(httpReq.OrderByFields) > 0 {
```
buqian-zilliz, line 1420: **[P2] Allow ordered grouped queries through REST**

Allow GROUP BY to use the query path's supported ORDER BY validation instead of rejecting the combination here. `queryTask` already restricts ORDER BY targets to GROUP BY columns, and the query reducers have dedicated GROUP BY+ORDER pipelines. This blanket rejection prevents the new REST `groupByFields` API from requesting deterministic ordering for its limit/offset pagination. Forward both parameters and let Proxy reject unsupported targets.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1408,15 +1417,27 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
...
+		return nil, err
+	}
 	if httpReq.Offset > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
 	}
-	if httpReq.Limit > 0 && !matchCountRule(httpReq.OutputFields) {
+	// REST applies a default limit, but Proxy rejects pagination for a global
+	// query containing count(*). A grouped count(*) uses limit/offset to
+	// paginate groups, so its limit must still be forwarded.
+	isGlobalCount := len(httpReq.GroupByFields) == 0 && outputsContainCountStar(httpReq.OutputFields)
```
buqian-zilliz, line 1431: **[P2] Reject offsets that are silently ignored for global counts**

Reject `offset` for a global `count(*)` aggregation instead of forwarding it without a limit. When `outputFields` mixes `count(*)` with another aggregate, this new branch suppresses the REST default limit while the earlier code still appends a positive offset. `parseQueryParams` only reads offset when a limit is present, so Proxy silently ignores the caller's offset and returns the unpaginated aggregate row. Return a parameter error when `isGlobalCount && offset > 0`, or otherwise give both pagination fields explicit semantics.

internal/proxy/util.go
```diff
@@ -3250,7 +3250,7 @@ func reconstructStructFieldData(
...
+	if len(fieldsData) == 0 || (len(outputFields) == 1 && outputFields[0] == "count(*)") {
```
buqian-zilliz, line 3253: **[P2] Reconstruct StructArray names for empty searches**

This early return bypasses the only step that collapses transformed StructArray subfields back to their parent field. `translateOutputFields` expands a requested StructArray into internal names such as `items[text]` and `items[embedding]`, while the zero-hit requery path returns `FieldsData=[]`; the empty response therefore exposes those internal names, whereas the same request with hits returns one `items` ArrayOfStruct field. Preserve aggregation output names on empty pages without skipping schema-based StructArray output reconstruction.

PR #51938, opened 2026-07-28
enhance: [cp 3.0] add first-class orderByFields/groupByFields params to RESTful v2

MrPresent-Han:
Cherry-pick of #51937 to 3.0.

Search order-by previously worked over REST only through the undeclared `searchParams.order_by_fields` passthrough, and query GROUP BY aggregation was unreachable because the query handler forwards a fixed whitelist of query params. This adds first-class `SearchReqV2.orderByFields` and `QueryReqV2.groupByFields` request fields with conflict/ambiguity rejection; the legacy passthrough alone keeps working. See #51937 for details.

issue: #41675, #36380
pr: #51937


internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1377,9 +1377,18 @@ func (h *HandlersV2) addCollectionStructField(ctx context.Context, c *gin.Contex
...
-func matchCountRule(outputs []string) bool {
-	return len(outputs) == 1 && strings.ToLower(strings.TrimSpace(outputs[0])) == "count(*)"
+// outputsContainCountStar reports whether any output field is the count(*)
+// aggregate, mirroring the Proxy's count(*) detection: a global (no GROUP BY)
+// query carrying count(*) plus a limit is rejected as meaningless pagination,
+// so REST must not attach its synthetic default limit to such a request —
+// including mixed aggregates like ["count(*)", "sum(x)"].
+func outputsContainCountStar(outputs []string) bool {
+	for _, o := range outputs {
+		if strings.EqualFold(strings.TrimSpace(o), "count(*)") {
```
liliu-z: Could we use `agg.MatchAggregationExpression` here? Proxy accepts valid forms such as `COUNT ( * )`, but this check does not. REST then adds the default limit and Proxy rejects the request as paginated, so the mixed-aggregate fix is still incomplete.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1411,12 +1420,19 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 	if httpReq.Offset > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
 	}
-	if httpReq.Limit > 0 && !matchCountRule(httpReq.OutputFields) {
+	// REST applies a default limit, but Proxy rejects pagination for a global
+	// query containing count(*). A grouped count(*) uses limit/offset to
+	// paginate groups, so its limit must still be forwarded.
+	isGlobalCount := len(httpReq.GroupByFields) == 0 && outputsContainCountStar(httpReq.OutputFields)
+	if httpReq.Limit > 0 && !isGlobalCount {
```
liliu-z: `Limit` is prefilled with 100, so this cannot tell the synthetic default from an explicit `"limit": 10`. For a global mixed aggregate, the explicit limit is now silently dropped; an explicit offset is also ignored by Proxy once limit is absent. Please track field presence and suppress only the default. Explicit pagination should still reach Proxy and be rejected.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1904,6 +1920,16 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
+		if len(httpReq.OrderByFields) > 0 {
```
liliu-z, line 1905: Could we enforce this in Proxy too? A gRPC/SDK caller can still send `SearchAggregation` with `order_by_fields`; both are parsed, but `newBuiltInPipeline` picks `searchWithAggPipe`, so order-by is silently ignored. Keeping the REST check is fine, but the invariant needs to be protocol-independent.

internal/distributed/proxy/httpserver/handler_v2_test.go
```diff
@@ -3761,6 +3761,136 @@ func TestQueryOrderByFields(t *testing.T) {
...
+		{
+			// Preserve legacy REST behavior: pagination is omitted for global count(*).
+			path:        QueryAction,
+			requestBody: []byte(`{"collectionName": "book", "filter": "book_id > 0", "outputFields": ["count(*)"], "limit": 10}`),
+		},
+		{
+			// Global mixed aggregate: default limit must still be suppressed, else
+			// Proxy rejects it as "count entities with pagination is not allowed".
+			path:        QueryAction,
+			requestBody: []byte(`{"collectionName": "book", "filter": "book_id > 0", "outputFields": ["count(*)", "word_count"]}`),
```
liliu-z: This is not a valid global aggregate. Once `count(*)` is present, raw `word_count` is rejected by `NewAggregationFieldMap`, but the mock makes this case look successful. Please use `sum(word_count)` and keep the no-limit assertion.

internal/distributed/proxy/httpserver/handler_v2_test.go
```diff
@@ -3761,6 +3761,136 @@ func TestQueryOrderByFields(t *testing.T) {
 	assert.Equal(t, []string{"word_count:desc,book_id:asc"}, orderByValues)
 }
 
+func TestQueryGroupByFields(t *testing.T) {
```
liliu-z, line 3759: These tests stop at a mocked Proxy and only verify forwarded KVs. The REST E2E suite does not send the new top-level `orderByFields` or `groupByFields`, so CI will not execute these API paths. Please add real REST cases that assert the returned search order and grouped query values.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: Thanks. We removed the REST count(*) special-case together with Proxy’s global count pagination prohibition. REST now forwards every positive limit, so there is no suppression path whose behavior can diverge for forms such as COUNT ( * ).

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: Agreed. Rather than trying to infer JSON field presence around a suppression branch, we removed both the suppression and Proxy’s global count pagination rejection. REST now forwards default and explicit positive limits; aggregation consumes all matching rows and pagination only slices final output rows. Regression coverage now checks omitted and explicit limits.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: Good catch. Added the protocol-independent validation in searchTask.initSearchAggregation: a non-empty order_by_fields now conflicts with search_aggregation. This covers direct gRPC/SDK callers; the existing REST validation remains. Added a focused Proxy test for the conflict.

internal/distributed/proxy/httpserver/handler_v2_test.go
MrPresent-Han: Fixed. The global mixed-aggregate test now uses count(*) with sum(word_count), which is row-aligned and valid. The test keeps the default-limit forwarding assertion; raw word_count remains covered only as a GROUP BY key.

internal/distributed/proxy/httpserver/handler_v2_test.go
MrPresent-Han: Added real REST E2E coverage. The query suite now verifies deterministic groupByFields output together with group ordering and per-group count(*); it also verifies that omitted and explicit REST limits do not truncate global count(*). Existing REST orderByFields cases assert returned row order.

PR #51940, opened 2026-07-28
fix: validate MATCH predicates are element-level [3.0]

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/46517
pr: #51725

### DIFF
--- a/internal/parser/planparserv2/parser_visitor.go
+++ b/internal/parser/planparserv2/parser_visitor.go
@@ -1206,6 +1206,81 @@ func isValidStructSubField(tokenText string) bool {
 	return len(tokenText) >= 4 && tokenText[:2] == "$[" && tokenText[len(tokenText)-1] == ']'
 }
 
+func validateMatchColumnInfo(columnInfo *planpb.ColumnInfo) (bool, error) {
+	if columnInfo == nil {
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate column info is missing")
+	}
+	if !columnInfo.GetIsElementLevel() {
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate can only use element-level fields")
+	}
+	return true, nil
+}
+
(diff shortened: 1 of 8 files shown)


internal/parser/planparserv2/parser_visitor.go
```diff
@@ -1206,6 +1206,81 @@ func isValidStructSubField(tokenText string) bool {
...
+		return false, merr.WrapErrParameterInvalidMsg("MATCH predicate is missing")
+	}
+
+	switch realExpr := expr.GetExpr().(type) {
+	case *planpb.Expr_ColumnExpr:
+		return false, merr.WrapErrParameterInvalidMsg("raw element columns are not supported inside MATCH predicate")
+	case *planpb.Expr_TermExpr:
+		return validateMatchColumnInfo(realExpr.TermExpr.GetColumnInfo())
+	case *planpb.Expr_UnaryRangeExpr:
+		return validateMatchColumnInfo(realExpr.UnaryRangeExpr.GetColumnInfo())
```
buqian-zilliz, line 1230: **[P2] Reject malformed templated range predicates**

Checking only the column flag here lets a malformed templated comparison pass the new MATCH validation. For `MATCH_ANY(sa, {k} + $[a] < $[b])`, the left arithmetic node is marked templated, so `handleCompare` takes its generic `left.expr.GetIsTemplate()` branch even though it is not a `ValueExpr`; it builds a `UnaryRangeExpr` on `$[b]` with neither `Value` nor `TemplateVariableName` and leaves `IsTemplate` false. This branch accepts that node, `FillExpressionValue` skips it, and segcore later decodes the unset `GenericValue` and fails. Require a concrete value or template name before accepting a unary range, or reject or fix this unsupported comparison shape.

zhengbuqian (approved):
regarding the above comments, I assume master currently have the same problem?

SpadeA-Tang:
> regarding the above comments, I assume master currently have the same problem?

yes

PR #51939, opened 2026-07-28
test: Increase PyMilvus version to 3.1.0rc78 from master for master branch

sijie-ni-0214:
## What changed

- Update `pymilvus` and `pymilvus[bulk_writer]` from 3.1.0rc69 to 3.1.0rc78.
- Align deprecated `drop_collection_function` tests with the unsupported legacy RPC response.
- Keep `drop_function_field` coverage for the supported cascade deletion behavior.
- Preserve the underlying TIMESTAMPTZ validation error returned by `AlterCollectionSchema`.
- Assert the complete deprecated DropCollectionFunction response in the TextEmbedding E2E test.

This supersedes https://github.com/milvus-io/milvus/pull/51849.

Related PyMilvus changes:
- https://github.com/milvus-io/pymilvus/pull/3704
- https://github.com/milvus-io/pymilvus/pull/3727
- https://github.com/milvus-io/pymilvus/pull/3729

## Verification

- `ruff check tests/python_client/testcases/test_text_embedding_function_e2e.py`
- `ruff format --check tests/python_client/testcases/test_text_embedding_function_e2e.py`
- `gofumpt -l internal/rootcoord/ddl_callbacks_alter_collection_schema.go internal/rootcoord/ddl_callbacks_alter_collection_schema_test.go`
- `go test -tags dynamic,test -gcflags="all=-N -l" ./internal/rootcoord -run TestDDLCallbacksBroadcastAlterCollectionSchema`
- PyMilvus 3.1.0rc78 TestPyPI dependency dry-run
- `git diff --check`

PR #51942, opened 2026-07-28
test: [cp3.0]Update PyMilvus to 3.0.1rc74

sijie-ni-0214:
Cherry-pick from master
pr: #51939

## Summary

- Align deprecated DropCollectionFunction expectations with the legacy RPC rejection.
- Update the 3.0 Python client test dependency to pymilvus-3.0.1rc74.
- Format test_milvus_client_timestamptz.py and replace its wildcard import with explicit imports so it passes Ruff. The file has accumulated historical formatting debt, so this mechanical formatting accounts for the large diff (832 added / 443 removed lines); it does not alter the test cases beyond the already-described expectation update.

## Test Plan

- ruff check and formatting check for test_milvus_client_timestamptz.py
- Python bytecode compilation for test_milvus_client_timestamptz.py
- Collected the three updated Drop Function cases with pytest
- Verified the RC74 client route with the L0 Search Iterator test
- git diff --check


congqixia (approved):
/approve

PR #51946, opened 2026-07-28
enhance: [2.6] Enable storage version compaction by default

congqixia:
Change dataCoord.compaction.storageVersion.enabled from false to true.


LoveEachDay (approved):
/lgtm

PR #51952, opened 2026-07-28
doc: add fuzzy BM25 search design

aoiasd:
## Summary

Define a three-phase architecture for segment term dictionaries, exact fuzzy BM25 expansion, and Global FST acceleration. The design also establishes the migration path toward one Text Term Index shared by fuzzy BM25 and text match features.

### DIFF
--- a/docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
+++ b/docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
@@ -1,0 +1,1036 @@
+# Text Term Index Foundation and Fuzzy BM25 Search
+
+**Author:** zhicheng
+**Date:** 2026-07-28
+**Status:** Draft
+
+---
+
+## 1. Summary
+
+This document proposes a three-phase design for fuzzy BM25 search in Milvus.
+The first phase establishes a segment-level Text Term Index foundation: terms
+produced by the analyzer are preserved through WAL, flush, compaction, load,
+and recovery, and QueryNode Workers expose a common term-expansion interface.
+Fuzzy BM25 then expands query tokens on the Workers before the Delegator builds
+the BM25 IDF vector.
+
+The second phase adds a Global FST on the Streaming Query Node (SN) as an exact
+acceleration layer. The Global FST supplies the vocabulary for covered
+segments, while the Delegator requests only the uncovered segment dictionaries
+from Workers. The third phase allows users to explicitly skip the uncovered
+segment RPC and trade freshness for latency.
+
+The initial field-level switch, `enable_fuzzy_bm25`, is intentionally
+transitional. The long-term direction is for `enable_match`, or a dedicated
+Match Index definition, to build one complete Milvus Text Term Index containing
+the term enumeration, FST, and postings. Exact text match, fuzzy text match,
+fuzzy BM25, prefix, wildcard, and other term-based text features should share
+that index instead of maintaining feature-specific files and execution paths.
+
+---
+
+## 2. Background
+
+### 2.1 Current BM25 search path
+
+For a BM25 function, the WAL write path materializes the sparse BM25 output
+before appending an Insert message. The analyzer emits tokens, each token is
+hashed to a `uint32`, and the per-row token frequencies are stored in a sparse
+vector.
+
+On the query path, the shard Delegator currently performs the following work
+before sending the search request to QueryNode Workers:
+
+```text
+raw query text
+  -> BM25 function runner
+  -> analyzer
+  -> hash tokens and build query TF sparse vector
+  -> IDFOracle.BuildIDF
+  -> BM25 query sparse vector
+  -> SearchSegments on Workers
+```
+
+`IDFOracle` is owned by the Delegator because BM25 IDF must be computed from a
+consistent shard-level view rather than independently by every segment.
+
+Fuzzy BM25 cannot use this path unchanged. Fuzzy expansion requires the
+original token strings and the searchable term vocabulary. Expansion must
+therefore happen before hashing and before `BuildIDF`.
+
+### 2.2 Current fuzzy text match path
+
+Milvus already implements the filter operator `text_match_fuzzy`. It is a
+segment-local operation over the existing Tantivy `TextMatchIndex`:
+
+```text
+query text
+  -> segment TextMatchIndex
+  -> Tantivy analyzer
+  -> Levenshtein automaton x Tantivy term dictionary FST
+  -> segment postings
+  -> result bitset
+```
+
+This path is appropriate for a filter because term expansion and posting lookup
+are both local to the segment. It does not need to expose the expanded terms to
+the Delegator.
+
+Fuzzy BM25 is different. The expanded terms must be combined across all target
+segments before the Delegator constructs one common BM25 query vector. If each
+segment independently built and scored a different vector, scores from
+different segments would not be comparable.
+
+The existing `TextMatchIndex` proves that an FST is useful for more than fuzzy
+BM25. Segment-level FST traversal is also a natural building block for fuzzy
+match, prefix, wildcard, regexp/automaton intersection, term enumeration, and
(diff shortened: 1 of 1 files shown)


docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,1173 @@
...
+the existing fuzzy text-match behavior. A dictionary term is a candidate when
+its distance from the query term is no greater than `max_edit_distance`.
+
+The feature is useful for queries containing typographical errors, such as
+matching `milvuz` with the indexed term `milvus`, while retaining BM25
+relevance ranking instead of returning only a Boolean match.
+
+This document proposes a three-phase design for fuzzy BM25 search in Milvus.
+The first phase establishes a segment-level FST foundation: terms produced by
+the analyzer are preserved through WAL and immutable FST fragments
```
zhengbuqian: why is WAL mentioned here?

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,1173 @@
...
+segment resources.
+
+The second phase adds a Global BM25 FST on the Streaming Query Node (SN) as an
+exact acceleration layer. The Global BM25 FST supplies the vocabulary for
+covered sealed segments, while the Delegator requests only uncovered sealed
+dictionaries from Workers and continues to expand growing terms locally. The
+third phase allows users to explicitly skip the uncovered sealed-segment RPC
+and trade freshness for latency.
+
+The initial field-level switch, `enable_fuzzy_bm25`, is intentionally
```
zhengbuqian: we need to allow altering this switch on the fly. in a new PR of course.

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,1173 @@
+# Segment FST Foundation and Fuzzy BM25 Search
+
+**Author:** zhicheng
```
zhengbuqian: this document is ok but too lengthy, make it more concise.

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,1173 @@
...
+---
+
+## 6. Three-phase delivery
+
+### Phase 1: Segment FST foundation and exact fuzzy BM25
+
+- Add the temporary field switch `enable_fuzzy_bm25`.
+- Preserve message-level deduplicated terms in Insert WAL messages.
+- Maintain the growing dictionary and mutable Trie on the Delegator.
+- Write one small immutable FST fragment per sync.
```
zhengbuqian: why is this necessary? do we ever need to read this fst fragment? why can't we just build it from scratch in sort compaction?

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,1173 @@
...
+
+---
+
+## 7. Field-level feature gate
+
+Phase 1 adds `enable_fuzzy_bm25` to the BM25 input string field.
+
+```text
+enable_analyzer = true
+enable_fuzzy_bm25 = true
```
zhengbuqian: another such flag or make it a field or even Function params? which is better?

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,1173 @@
+# Segment FST Foundation and Fuzzy BM25 Search
+
+**Author:** zhicheng
+**Date:** 2026-07-28
```
zhengbuqian: also missing the feature of requiring min prefix length

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
aoiasd: flag enable_fuzzy for bm25 function seems better.

docs/design-docs/design_docs/20260728-fuzzy-bm25-search.md
```diff
@@ -0,0 +1,392 @@
...
+FST for that field. This also preserves the path toward a shared field-scoped
+Segment Text Index.
+
+The Function must have the normal BM25 shape. Its input field must be `VARCHAR`
+or `TEXT` and set `enable_analyzer = true`. Omission or `false` keeps fuzzy BM25
+disabled and does not require Segment FST maintenance.
+
+The first implementation may reject online changes to this Function parameter
+because existing segments do not yet have complete Segment FSTs. A follow-up PR
+should support `false -> true` by building and publishing complete FSTs for
```
zhengbuqian: in that PR, we'll disallow fuzzy search until the FSTs are finished building, right?

aoiasd: Yes. The design now requires `fuzziness > 0` searches to be rejected during a future `false -> true` backfill, while exact BM25 remains available. Fuzzy readiness is published only after every visible segment has complete FST coverage.

congqixia:
**Comment: Global BM25 FST placement is ambiguous**

The design says the Global BM25 FST is "loaded beside the Delegator" (§3.1) and that the Delegator maintains growing Tries, batches Worker expansion, and runs search (§3.2). But "beside the Delegator" is ambiguous given the current topology: in the new architecture the Delegator is a *role* that physically runs inside the streamingnode's embedded querynode (SQNode), while the Worker role runs in the standalone querynode process.

This matters for three things the doc does not pin down:

1. **Physical owner**: does the Global FST live in the SQNode (streamingnode embedded querynode, sharing its memory budget with the WAL buffer + embedded segcore) or in a standalone querynode? The doc's own Phase 2 motivation (reduce sealed-segment RPCs) implies it should be co-located with the Delegator, but "loaded beside" never states the process.
2. **Builder ownership**: §3.2 lists only "QueryNode Delegator / Worker" responsibilities; the Global FST *builder* is not assigned to any component. With the Delegator now embedded in the streamingnode, an unassigned builder is more consequential than in the old topology.
3. **Memory/rollout implications**: the streamingnode already carries heavy memory; an additional shard-and-field-level vocabulary (or an admission estimate for it) should be explicitly accounted for, like the growing-Trie expansion factor.

Suggest being explicit about the physical placement (process), the builder, and the memory accounting for Phase 2.

congqixia:
**Comment: inconsistent external-collection support between fuzzy BM25 and exact BM25**

The design lists External Collections as a non-goal for the first release (§2), which is reasonable for the WAL-sidecar + segment-FST path. However, on current master exact (non-fuzzy) BM25 **already works on external collections**: refresh materializes the BM25 output field and serializes per-field BM25 stats into the packed manifest (`datanode/external/function_executor.go`), and QueryNode's IDF oracle loads those sealed stats and builds IDF in the normal search path (verified end-to-end by `TestExternalTableBM25Function`). So the exact/fuzzy split on external collections is asymmetric by design.

This asymmetry has two implications the doc should address:

1. **Documentation gap**: fuzzy BM25 being unavailable on external collections while exact BM25 works is user-visible. The doc should state explicitly that fuzzy (`fuzziness > 0`) returns an error on external-collection fields (or is simply not enabled) — right now a reader would have to infer it from the non-goal. The non-goal bullet itself is clear enough, but the user contract (§4) doesn't say what happens when a user enables fuzzy on an external collection.

2. **Future convergence**: since exact BM25 stats already flow through the packed manifest on external collections (no WAL sidecar needed), a later "fuzzy on external" path could reuse that same manifest-driven vocabulary build instead of the WAL term sidecar. The doc's Phase 3 evolution should acknowledge this alternative data source; otherwise the stated "no external collections" reason (no WAL terms) reads as a permanent limitation rather than a scoped first-release boundary.

Issue #51941, opened 2026-07-28
[Bug]: [Benchmark][Cluster] Nullable growing-vector offset-mapping snapshot race during concurrent upsert/search


wangting0128:
- [x] I have searched the existing issues

### Environment

- Milvus version: `harbor.milvus.io/milvusdb/milvus:51916-growing_index_ownership-3.0-20260728-2a271a3227-v2-amd64`
- Relevant code change: [#51913](https://github.com/milvus-io/milvus/pull/51913), `fix: harden growing index raw data ownership`
- Deployment mode: StreamingNode cluster; MQ: Pulsar
- SDK: pymilvus `2.7.0rc175` (`client_version=2.6`)
- Kubernetes: 4AM cluster, namespace `qa-milvus`
- Deployment config: `zong-server-laion-sn`; collection: `laion_stable_10_nullable`
- Workload: `test_concurrent_locust_custom_parameters` with concurrent Search, Query, Load, and Upsert

### Server Pod Topology

Snapshot collected with:

```bash
kubectl get pod -o wide -n qa-milvus | grep -E "NAME|laion1b-test-2"
```

All Milvus server pods were `Running`, `1/1 Ready`, and had `0` restarts.

| Component | Pod | Pod IP | Node |
|---|---|---|---|
| MixCoord | `laion1b-test-2-milvus-mixcoord-76bb5c97c4-92qlb` | `10.104.32.93` | `4am-node39` |
| Proxy | `laion1b-test-2-milvus-proxy-7db4b6b4d8-79w6w` | `10.104.34.189` | `4am-node37` |
| DataNode | `laion1b-test-2-milvus-datanode-d78996ff6-l6zwt` | `10.104.14.5` | `4am-node18` |
| DataNode | `laion1b-test-2-milvus-datanode-d78996ff6-m7869` | `10.104.19.169` | `4am-node28` |
| DataNode | `laion1b-test-2-milvus-datanode-d78996ff6-qlvqb` | `10.104.32.90` | `4am-node39` |
| DataNode | `laion1b-test-2-milvus-datanode-d78996ff6-rq92k` | `10.104.24.79` | `4am-node29` |
| QueryNode | `laion1b-test-2-milvus-querynode-0-f5f76fd74-7mlf2` | `10.104.19.170` | `4am-node28` |
| QueryNode | `laion1b-test-2-milvus-querynode-0-f5f76fd74-c8h4b` | `10.104.34.190` | `4am-node37` |
[...]

PR #51953, opened 2026-07-28
fix: stabilize growing nullable bitset snapshot

marcelo-cjl:
## Summary
- Materialize growing nullable-vector MVCC no-filter bitsets in physical offset space.
- Use the transformed physical bitset size as the growing search boundary.
- Avoid scanning nullable vector rows appended after the query snapshot.

issue: #51941

## Test plan
- [x] `make build-cpp-with-unittest`
- [x] `make run-test-cpp filter='Util_Segcore.*' jobs="$(nproc)"`
- [x] `clang-format-15 --dry-run -Werror internal/core/src/common/OffsetMapping.cpp internal/core/src/query/SearchOnGrowing.cpp internal/core/src/segcore/UtilsTest.cpp`
- [x] `git diff --check`

### DIFF
--- a/internal/core/src/query/SearchOnGrowing.cpp
+++ b/internal/core/src/query/SearchOnGrowing.cpp
@@ -196,22 +196,19 @@ SearchOnGrowing(const segcore::SegmentGrowingImpl& segment,
 
         TargetBitmap transformed_bitset;
         BitsetView search_bitset = bitset;
-        if (has_offset_mapping && !bitset.empty()) {
+        int64_t active_count = 0;
+        if (has_offset_mapping) {
             auto status =
                 offset_mapping.TransformBitset(bitset, transformed_bitset);
             if (status == OffsetMapping::BitsetTransformStatus::AllFiltered) {
                 FillEmptySearchResult(search_result, num_queries, info.topk_);
                 return;
             }
-            if (status == OffsetMapping::BitsetTransformStatus::NoFilter) {
-                search_bitset = BitsetView{};
-            }
+            active_count = static_cast<int64_t>(transformed_bitset.size());
+        } else {
+            active_count = std::min(int64_t(bitset.size()),
+                                    segment.get_active_count(timestamp));
         }
-
-        auto active_count = has_offset_mapping
-                                ? offset_mapping.GetValidCount()
-                                : std::min(int64_t(bitset.size()),
-                                           segment.get_active_count(timestamp));
 
         // Check for nullable vector field with all null values
         if (active_count == 0) {
@@ -219,8 +216,7 @@ SearchOnGrowing(const segcore::SegmentGrowingImpl& segment,
             FillEmptySearchResult(search_result, num_queries, info.topk_);
             return;
         }
-        if (has_offset_mapping && !bitset.empty() &&
-            !transformed_bitset.empty()) {
+        if (has_offset_mapping && !transformed_bitset.empty()) {
             search_bitset =
                 search_result.PinBitset(std::move(transformed_bitset));
         }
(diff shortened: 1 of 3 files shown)


internal/core/src/query/SearchOnGrowing.cpp
```diff
@@ -196,31 +196,27 @@ SearchOnGrowing(const segcore::SegmentGrowingImpl& segment,
 
         TargetBitmap transformed_bitset;
         BitsetView search_bitset = bitset;
-        if (has_offset_mapping && !bitset.empty()) {
+        int64_t active_count = 0;
+        if (has_offset_mapping) {
```
liliu-z, line 200: Blocking: `bitset` can be empty here, and empty means “no filter / all rows visible”. `BitsetView::all()` is true for an empty view, so the growing transform returns `AllFiltered` and this path silently returns zero results (`visibilityFilterEnabled=false` is one example). Please preserve the empty-view semantics and add a growing nullable regression test. Since an empty view has no size, the logical snapshot bound should come from the frozen plan `active_count`, not the current mapping.

liliu-z (commented):
/lgtm
/approve

internal/core/src/query/SearchOnGrowing.cpp
foxspy: A concrete triggering scenario is a nullable-vector growing segment searched with no scalar or visibility filter, for example when visibility_filter_enabled=false and all rows in the snapshot are visible. VectorSearchNode then passes an empty BitsetView to mean no filtering. Because empty.all() is true, TransformBitset returns AllFiltered and SearchOnGrowing returns immediately, so a nonempty segment can silently produce zero hits. Please handle the empty view before all() and preserve the query-frozen logical active_count when deriving the physical snapshot boundary.

internal/core/src/query/SearchOnGrowing.cpp
```diff
@@ -196,31 +196,27 @@ SearchOnGrowing(const segcore::SegmentGrowingImpl& segment,
...
             auto status =
                 offset_mapping.TransformBitset(bitset, transformed_bitset);
             if (status == OffsetMapping::BitsetTransformStatus::AllFiltered) {
                 FillEmptySearchResult(search_result, num_queries, info.topk_);
                 return;
             }
-            if (status == OffsetMapping::BitsetTransformStatus::NoFilter) {
-                search_bitset = BitsetView{};
-            }
+            active_count = static_cast<int64_t>(transformed_bitset.size());
```
foxspy, line 207: There is still a storage-publication race when nullable inserts and growing searches run concurrently. ConcurrentVector::set_data_raw() publishes offset_mapping_.Append() before allocating the required chunk and copying the vector data. A search can therefore derive this larger active_count after the mapping update but before the backing physical positions are ready. If the append crosses a chunk boundary, get_chunk_data() can assert because the chunk does not exist yet; within an existing chunk, search can read positions whose vectors have not been copied, producing incorrect distances or top-K results. The mapping count and physical storage need a single publication boundary, with a concurrent chunk-boundary regression test.

internal/core/src/query/SearchOnGrowing.cpp
marcelo-cjl: This issue does exist.

There are two related problems here. First, when visibilityFilterEnabled=false, or when there is no scalar/visibility filter and all snapshot rows are visible, VectorSearchNode can pass an empty BitsetView to mean no-filter/all-visible. The growing nullable transform could then treat that empty view as all(), return AllFiltered, and produce empty results.

Second, the visible logical bound should not come from the mutable offset mapping valid_count. The offset mapping can grow after the query snapshot, so using its current valid_count can let newly appended physical vector entries affect the search range.

The follow-up fix is in #51988. It moves the snapshot bound to the upper search node: for growing all-visible search, VectorSearchNode materializes an all-visible logical bitset using the query-frozen active_count_. Then TransformBitset only maps that bounded logical view to physical vector ids, so newly appended mapping entries cannot extend the search range. The 3.0 pick is #51989.

Issue #50377, opened 2026-06-08
[Bug]: Standalone SIGSEGV in GIS ST_WITHIN query on nullable geometry growing segment


zhuwenxing:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## Environment

- Milvus version: `harbor.milvus.io/milvusdb/milvus:master-20260606-43010a4f`
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): pulsar
- SDK version(e.g. pymilvus v2.0.0rc2): Python client test in Jenkins deploy_test_cron, exact pymilvus version unknown from archived logs
- OS(Ubuntu or CentOS): Linux container on Kubernetes
- CPU/Memory: unknown from archived logs
- GPU: N/A
- Others:
  - Jenkins job: `deploy_test_cron`, build `4066`
  - Jenkins URL: https://qa-jenkins.milvus.io/job/deploy_test_cron/4066/
  - Namespace: `chaos-testing`
  - Instance: `pulsar-standalone-reinstall-4066`
  - The crash happened during the "Send Concurrent Requests" stage.

### K8s Pod List

The instance was later cleaned up, so the live pod is no longer available. The following pod list was captured during the failing Jenkins run:

```text
pulsar-standalone-reinstall-4066-etcd-0                           1/1     Running       0              24m     10.104.27.246   4am-node31   <none>           <none>
pulsar-standalone-reinstall-4066-etcd-1                           1/1     Running       0              24m     10.104.15.110   4am-node20   <none>           <none>
pulsar-standalone-reinstall-4066-etcd-2                           1/1     Running       0              24m     10.104.20.180   4am-node22   <none>           <none>
pulsar-standalone-reinstall-4066-milvus-standalone-5bbd5c8ds4q4   1/1     Running       3 (15m ago)    24m     10.104.32.211   4am-node39   <none>           <none>
pulsar-standalone-reinstall-4066-minio-759f696ddd-lq5f4           1/1     Running       0              24m     10.104.18.183   4am-node25   <none>           <none>
pulsar-standalone-reinstall-4066-pulsarv3-bookie-0                1/1     Running       0              24m     10.104.15.107   4am-node20   <none>           <none>
pulsar-standalone-reinstall-4066-pulsarv3-bookie-1                1/1     Running       0              24m     10.104.23.238   4am-node27   <none>           <none>
[...]

PR #48701, opened 2026-04-02
enhance: add visibilityFilterEnabled and bloomFilterEnabled toggles for querynode optimization

sparknack:
## Summary
- Add `common.visibilityFilterEnabled` toggle to bypass row visibility filtering (timestamp, delete, and TTL) on querynode. When disabled, all rows are returned regardless of insert/delete timestamps, improving performance for workloads that don't need timestamp consistency.
- Add `common.bloomFilterEnabled` toggle to skip bloom filter/PK stats loading on querynode, reducing memory for workloads without delete or search-by-id. Delete forwarding falls back to broadcast mode when disabled.

issue: #48703

## Test plan
- [ ] Unit tests for bloom filter disabled path (`TestLoadSegmentsWithoutBloomFilter`, `TestLoadWithoutBloomFilter`)
- [ ] Unit tests for nil candidate broadcast in delete forwarding
- [ ] Unit tests for visibility filter bypass with `visibilityFilterEnabled=false`
- [ ] Verify no nil pointer panics when bloom filter is disabled
- [ ] Integration test with `bloomFilterEnabled=false` and `visibilityFilterEnabled=false`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #51960, opened 2026-07-28
fix: stabilize growing segment read snapshots

xiaofan-luan:
issue: #51643
Related: #50377, #51953

Split from #51878.

Rebased on top of #51953 ("stabilize growing nullable bitset snapshot"), which fixed the brute-force half of the same bug by materializing the no-filter bitset in physical space under one lock. That `TransformBitset` contract (and its test) is kept; this PR layers the plan-frozen bound on top so the interim-index, iterator, element-level and empty-bitset paths are covered as well, and replaces the chunk-lock reader protocol with generation pinning so reclamation no longer starves under query load.

## What this fixes

A growing-segment search must operate on one acknowledged snapshot. Today the visible-row bound is re-derived inside the search kernels, and for a nullable vector field it is derived from the offset mapping's current size. Both race with concurrent inserts: a growing segment publishes rows into column storage — and into a nullable field's offset mapping — before `ack_responder_` advances, so a later read admits rows that are neither acknowledged nor visible at the query timestamp. Those rows then fail Reduce's offset validation, or slip into the result when the ack happens to catch up first.

Separately, the old reader protocol held a segment-wide chunk lock for whole scans, which made `try_remove_chunks`' try-lock fail under any query load: a segment that stopped taking writes never reclaimed its raw-chunk memory once the interim index owned the data.

### Acknowledged-row snapshot

- `SearchInfo::active_count_` is frozen once in the plan layer (`PhyVectorSearchNode`) for growing segments and carried into the search kernels, which no longer re-derive it. `-1` means "not supplied": sealed searches and direct callers keep their existing behaviour.
- The raw growing path uses that logical bound directly, and converts it into the mapping's physical space for nullable vectors through the new `OffsetMapping::ValidCountBelow`, rather than `GetValidCount()`.
- The interim-index path applies the same bound explicitly. Knowhere treats an empty bitset as `IDSelectorAll`, so a no-filter fast path must still carry a prefix length; otherwise rows appended after the bound can occupy ANN top-k slots before Reduce validates their offsets. This covers nullable and non-nullable indexes, including the iterator paths.
- An empty `BitsetView` means "no filter", not "zero rows". Clamping the bound to `bitset.size()` zeroed it, so a nullable vector field came back with an empty result instead of every visible row.
- Element-level (embedding-list) search no longer clamps its row-space scan bound with the element-space bitset size: a nullable embedding-list column stores null rows as empty arrays, so the flattened element count can run below the row count, and the `min()` silently dropped the tail rows from brute-force scans. The scan bound also converts through `ValidCountBelow` whenever the mapping is enabled, so a compacted column is never walked with a logical bound past its physical rows.

### Chunk lifetime: snapshot pinning instead of a chunk lock

Knowhere's brute-force iterators retain raw pointers into chunk storage past the end of the call that produced them, while `try_remove_chunks` reclaims that storage once the interim index owns the raw data.

- `VectorBase::acquire_chunks()` returns a `ChunkSnapshot`: shared ownership of the chunk collection plus the chunk count published with it, latched together under the container's own mutex. Readers address every chunk through the snapshot with no lock. `SearchOnGrowing` pins one generation up front and stores the reference in `SearchResult::resource_pins_` for iterators that outlive the call.
- `ThreadSafeChunkVector::clear()` swaps the collection out instead of clearing in place; the old generation dies with its last holder, on whichever thread that is. The segment-wide chunk mutex is gone entirely — no try-lock, so reclamation proceeds immediately and unconditionally, and retention is bounded by query lifetime, not segment lifetime.
- The chunk store is an owned append-only slab directory (`ChunkSlabDirectory`: a fixed spine of geometrically growing slabs) rather than `std::deque` or `tbb::concurrent_vector`. Unsynchronised indexed reads need an index structure that never moves; deque reallocates the node map `operator[]` walks, and oneTBB specifies `concurrent_vector::reserve()` — the call that keeps an allocation failure from ever counting unconstructed slots — as concurrently **unsafe** with every other member, element access included, so the previous revision depended on a private implementation detail of the pinned 2021.9. The directory allocates slabs before publishing their pointers and move-constructs elements into pre-reserved storage, so nothing a reader can see ever moves and a throw changes nothing a reader can see.
- A chunk's element offset is a multiplication instead of a sum over every preceding chunk's size; the expression layer asks once per chunk, so the sum made one growing-segment scan O(num_chunks^2) — ~300M iterations on a multi-million-row segment at the default chunkRows of 128.

### Offset mapping: single array, lock-free, batch-friendly

`OffsetMapping` becomes a pure interface with `NoOp` / `Sealed` / `Growing` `final` implementations in their own files; adding a conversion to the interface now fails to compile in an implementation that misses it, instead of silently inheriting an identity mapping — which on a growing mapping that has nulls means handing the scan offsets past the end of physical storage.

The growing implementation replaces two shared_mutex-guarded hash maps (~64 B per valid row, a lock RMW per lookup) with one dense int32 physical→logical array over a fixed spine of geometrically growing chunks, published by a single release-store to a packed (valid, total) counter:

- 4 bytes per valid row and nothing for null rows — a sparse nullable column pays for the vectors it stores, not for every logical row.
- logical→physical is a binary search over the strictly-increasing array. An all-valid mapping short-circuits to the identity, and batch conversions gallop from a cursor on ascending inputs, so full-range passes cost O(rows) rather than a binary search per row.
- The three production full-scan loops — the VECTOR_ARRAY chunk views the expression layer walks, `bulk_subscript_vector_array_impl`, and flush's `BuildVectorArrayForChunk` — convert one contiguous batch at a time, which also drops a per-row validity lock from each.
- `GrowingOffsetMapping::Append` asserts the ascending-order precondition. Nothing re-establishes that order after the fact, so the assertions are the enforcement; the contract holds because the querynode pipeline drives each vchannel from a single goroutine. `SegmentGrowingImpl::Insert` documents this, including the fact that plain schemas *do* tolerate concurrent inserts — only compact (nullable vector) storage does not.
- `Append` reserves only what the batch's valid rows need: an all-null schema-evolution backfill (an entire segment's rows in one batch) allocates nothing, instead of parking a full worst-case reservation (~64 MiB per 10M rows) that nothing ever reads.
- Validity is written at the logical offset the caller reserved, instead of appended at the current length.

### Reopen retry convergence

The ordering assertions above surfaced a retry hazard: `Reopen` backfills new columns before building text indexes and publishes the schema last, so a text-index failure leaves a new nullable-vector column (and its order-asserted mapping) already backfilled while the schema stays old — and the retry re-entered `fill_empty_field` from logical offset 0. The backfill now fills only the missing suffix (the content is uniform nulls/defaults, so this is equivalent), making retries convergent; it also switches the validity write to the offset-addressed overload, fixing a pre-existing bug where a retry doubled the validity length. Covered by `ReopenRetryAfterTextIndexFailureBackfillsOnce`.

### Offset-mapping contract cleanups

- `SealedOffsetMapping::Build` resets state before the degenerate-input early return, so rebuilding with null/empty input leaves a disabled empty mapping (as the header states) instead of silently retaining the previous build.
- `GrowingOffsetMapping::FilterValidLogicalOffsets` on a disabled mapping now treats logical and physical spaces as coinciding, matching every other conversion and the NoOp/Sealed implementations, instead of rejecting all offsets against `total_count_ == 0`.

### `common.visibilityFilterEnabled`: bypass removed, surface kept as a deprecated shim

The knob (added by #48701) gated a global bypass in `PhyMvccNode` that returned every row regardless of insert, delete and TTL timestamps — and returned deleted rows, since deletes were forwarded to the segment but never applied while it was on (the warning in `query_node.go` documented exactly this). The bypass branch and its tests are removed; filtering is always enforced.

Because the parameter shipped in v3.0.0 with `Export: true`, its compatibility surface is retained as a deprecated shim rather than removed outright:

- `common.visibilityFilterEnabled` / `CommonCfg.VisibilityFilterEnabled` still parse (marked deprecated, no longer exported into new configs). A configuration that sets the key to `false` fails querynode startup with an explicit error, so an operator who relied on the bypass gets a clear migration signal instead of a silent behavior change during a rolling upgrade.
- The `SegcoreSetVisibilityFilterEnabled` C symbol survives as a documented no-op, so callers built against the v3.0.0 interface keep linking.

The sealed-segment fast path from #48272 — which skips the redundant timestamp mask only when it is provably a no-op — is preserved and covers the legitimate performance case. This still deserves a release-note entry.

### Drive-by

Two uncallable member functions in touched files are removed: `InsertRecordGrowing::drop_field_data()` and the one-argument `SegmentGrowingImpl::try_remove_chunks(FieldId)` overload.

## Deliberately not included

An earlier revision of this branch also carried the following. They were dropped to keep the change to what the fix actually needs:

- **Serializing nullable-vector inserts behind a condition variable.** The querynode pipeline already delivers a growing segment's inserts one at a time (`streamPipeline::work` → `insertNode::Operate` → `ProcessInsert`), so the gate never fired. `GrowingOffsetMapping::Append`'s assertions state the precondition instead of silently correcting a caller bug.
- **Publishing the offset mapping only after compact chunk data is copied.** With the bound above in place, no reader resolves offsets inside that window.
- **Copy-on-write field maps in `InsertRecordGrowing`.** `field_map_mutex_` already prevents the schema-evolution rehash race; replacing it with atomic snapshots was a lock-removal optimisation, not part of this fix, and it weakened `get_span_base`'s single-snapshot guarantee.

## Validation

- clang-format 15.0.7 clean on every changed file; full C++ build with unit tests.
- Local suites: `GrowingOffsetMapping` / `OffsetMapping` / `SealedOffsetMapping` / `ElementFilter*`, plus the wide `*Growing*` / `*CachedSearchIterator*` / `*ConcurrentVector*` / `*ChunkVector*` / `*VectorArray*` sweep — 700+ tests green. (Two text-column flush cases skip in this environment; DiskANN emb-list cases fail locally with "index not supported", an environment limitation independent of this change.)
- New regression tests in this revision:
  - `ElementFilterGrowingNullable.SearchAndSubscriptAcrossPhysicalChunks`: nullable embedding list `[null, valid-empty, one-vector]` with `chunk_rows=1`, so every stored row sits in its own physical chunk. Brute-force and iterator-v2 searches must both return (doc 2, elem 0); `bulk_subscript` must hand back null / valid-empty / valid rows unchanged.
  - `FlushGrowingSegmentTest.FlushNullableEmbListMixedRows`: flush round-trip of the same shape through `BuildVectorArrayForChunk`, with `chunk_rows=1` (the stored rows cross a physical chunk boundary inside compact storage) and content-level assertions — per-row validity `[null, valid, valid]`, valid-row count, the valid-but-empty row stays a zero-length list rather than becoming null (and vice versa), and the payload row round-trips element for element.
  - `GrowingOffsetMapping`: gallop-vs-point-lookup equivalence across chunk boundaries, unsorted/duplicate inputs, all-valid identity fast path, and all-null backfill behaviour.


internal/core/src/segcore/SegmentGrowingTest.cpp
```diff
@@ -1956,6 +1958,307 @@ TEST(Growing, ConcurrentInsertResourceTracking) {
...
+        if (with_new_fields) {
+            sch->AddDebugField("nvec",
+                               DataType::VECTOR_FLOAT,
+                               dim,
+                               knowhere::metric::L2,
+                               /*nullable=*/true);
+            sch->AddDebugVarcharField(FieldName("match_text"),
+                                      DataType::VARCHAR,
+                                      256,
+                                      /*nullable=*/false,
```
foxspy: This test does not currently reach the intended text-index/analyzer failure. `match_text` is added as a non-nullable field without a default value, so `Reopen()` fails earlier while backfilling the 32 existing rows. The first `EXPECT_ANY_THROW` masks that mismatch by accepting the wrong exception, and the second `Reopen()` fails for the same missing-field reason. As a result, the test does not exercise retry convergence after a text-index build failure.

Could we make `match_text` nullable or provide a valid default value, and verify that the first failure is specifically caused by the invalid analyzer JSON? After correcting the analyzer configuration, the second `Reopen()` should succeed, and the nullable-vector mapping should still contain `N` backfilled rows rather than `2N`. This would ensure the test covers the intended partial-backfill and retry path.

xiaofan-luan: Addressed in the current revision (0795fd251a): `match_text` is now nullable, with a comment explaining why a non-nullable field without a default would fail earlier in the backfill and never reach the text-index build. The broad `EXPECT_ANY_THROW` is gone — attempt 1 catches the exception and asserts the "failed to create text writer" message, pinning the failure to the text-index build specifically. Attempt 2 with corrected analyzer params succeeds (schema version advances), and the nullable-vector mapping is asserted to hold N backfilled rows (not 2N) both after the failed attempt and after the successful retry; a follow-up insert then verifies offsets continue from the backfilled prefix (total reaches 2N only through real rows).

chyezh:
I reviewed this PR against the additional MVCC counterexamples from #52005. The vector read-snapshot work is valuable: freezing the growing `active_count`, pinning chunk generations, and carrying the physical prefix into brute-force/interim-index/iterator paths address the insert-suffix and chunk-reclamation races. However, I found one high-severity residual MVCC gap and one compatibility concern.

### P1 / High: the acknowledged row bound is not paired with a timestamp-safe delete snapshot

PR #51960 stabilizes the insert/chunk side, but delete publication remains outside that snapshot. `LoadPush()` can first add newer deletes to the cumulative `deleted_mask_`; a later `StreamPush()` then clones the entire mask into `latest_snapshot_` while assigning only that streaming batch's `max_ts`.

Concrete counterexample:

```text
insert A@100, B@100
LoadPush(delete B@400)
StreamPush(delete A@200)
Query@250
```

The published snapshot can become:

```text
max_ts = 200
bitset = {A, B}
```

Because `250 >= 200`, the fast path applies the whole bitmap and incorrectly hides B, even though B's delete is in the future.

There is a second recovery-order problem at `DeletedRecord.h:153`: deduplication uses only `deleted_mask_[row_id]`. For:

```text
LoadPush(delete A@400)
StreamPush(delete A@200)
```

the earlier effective delete at 200 can be discarded because the physical row is already marked. The timestamp-aware fallback can then expose A at Query@250. Consequently, disabling the latest-snapshot optimization does not fully solve the out-of-order replay case.

This matches #52005's failure signature: a future upsert delete crosses an older MVCC boundary while the future insert remains hidden.

Suggested fix:

- The snapshot watermark must be at least the maximum timestamp of every delete represented by its bitmap, including loaded/replayed deletes.
- Preserve the earliest effective delete timestamp per physical row when deletes arrive out of order, or retain the older skip-list entry instead of deduplicating solely by a boolean mask.
- If recovery cannot establish these invariants, keep queries on the timestamp-aware path or return a retryable recovery error rather than serving the channel.

Tests should cover optimization-on/off parity for:

```text
LoadPush(B@400) -> StreamPush(A@200) -> Query@250
LoadPush(A@400) -> StreamPush(A@200) -> Query@250
StreamPush(newer) -> StreamPush(older) -> Query between timestamps
```

and the full resurrection sequence:

```text
insert v1@100
delete@200
insert v2@300
delete@400
```

with queries at 150/200/250/300/350/400/450 through the delete fast path, fallback, brute-force search, and interim-index search.

### MVCC scenario assessment

1. **Future delete vs. older snapshot:** not handled; the delete bitmap can leak future deletes.
2. **Growing upsert/resurrection:** correct only when mutations are delivered monotonically.
3. **Multiple physical versions:** ordered behavior is correct. Production uses `include_same_ts=false`, so delete@D affects versions with `insert_ts < D`; a version inserted exactly at D survives. Delete@D is visible at Query@D.
4. **One coherent generation across ack/timestamps/PK/chunks/index/deletes:** partial. The PR stabilizes `active_count` and vector chunks, but not PK/delete publication.
5. **Unsorted recovery/replay:** not handled. Out-of-order delete replay can mislabel the bitmap and lose an earlier delete. If insert timestamps themselves can be unsorted, `get_active_count()` also still relies on `std::upper_bound` over an ordered timestamp prefix.
6. **Chunk/index replacement lifetime:** handled for the vector read paths changed by this PR; `ChunkSnapshot` and `resource_pins_` cover the relevant UAF/stale-pointer risk.
7. **Optimized/fallback equivalence:** vector paths are aligned, but delete bitmap and skip-list paths are not equivalent under out-of-order recovery.

### P2 / Medium: preserve the shipped visibility-filter compatibility surface

The PR removes `common.visibilityFilterEnabled`, `CommonCfg.VisibilityFilterEnabled`, and `SegcoreSetVisibilityFilterEnabled`. Since the parameter shipped in v3.0.0 with `Export: true`, removing these outright breaks source/link compatibility and silently ignores existing configurations.

The unsafe bypass should remain impossible, but the old parameter/member/C symbol can be retained as deprecated compatibility shims that always enforce filtering, or reject `false` with a clear startup/configuration error. This also gives operators an explicit rolling-upgrade and migration signal.

xiaofan-luan:
@chyezh Thanks for the deep review. I verified both counterexamples against `DeletedRecord.h` and they hold:

- `StreamPush` publishes the cumulative `deleted_mask_` paired with only the streaming batch's `max_ts` (`UpdateLatestSnapshot`), so a loaded/replayed future delete already present in the mask leaks through the fast path exactly as you describe.
- The dedup at `DeletedRecord.h:153` is timestamp-blind, so `LoadPush(A@400)` -> `StreamPush(A@200)` drops the earlier effective delete from the skip list, and the timestamp-aware fallback misses it as well — disabling the snapshot optimization is indeed not sufficient.

Both sit in delete publication, which this PR deliberately does not touch (`DeletedRecord.h` is not in the diff), and the failure reproduces on master with the same sequences — it is pre-existing rather than introduced here. To keep this PR's already-verified scope (insert-suffix bound, chunk generation pinning, offset mapping) reviewable, we will fix the delete-side MVCC in a dedicated follow-up PR linked to #52005, following the invariants you outlined:

- snapshot watermark >= the max timestamp of every delete represented in the bitmap, loaded/replayed deletes included;
- timestamp-aware dedup that preserves the earliest effective delete per physical row (or retains the older skip-list entry) instead of a boolean mask check;
- recovery either establishes these invariants or keeps queries on the timestamp-aware path / returns a retryable error;
- the optimization-on/off parity matrix and the insert/delete resurrection sequence you listed as regression tests.

I will open the follow-up and link it back here. Will respond on the `visibilityFilterEnabled` compatibility point separately.

xiaofan-luan:
On the `visibilityFilterEnabled` compatibility point (P2): implemented in f959c5894f as a deprecated shim per your suggestion — the bypass stays removed and filtering is always enforced, but the parameter / `CommonCfg` member still parse (marked deprecated, no longer exported), setting the key to `false` now fails querynode startup with an explicit error instead of a silent behavior change, and `SegcoreSetVisibilityFilterEnabled` survives as a documented no-op for link compatibility. PR body updated accordingly.

czs007:

Adversarial review of the growing-segment snapshot changes surfaced no correctness or concurrency defects; the only actionable item on this PR is a documentation/dead-code nitpick.

## Nitpicks

- **`internal/core/src/query/SearchOnGrowing.cpp:302`** — The newly added `NoFilter → search_bitset = BitsetView{}` branch is unreachable at this call site. `GrowingOffsetMapping::TransformBitset` returns `NoFilter` only when `counts.total == 0` (`GrowingOffsetMapping.cpp:198-200`) or when `bitset.empty()` (`:209-211`), and both are already excluded by the guards above: `has_offset_mapping = offset_mapping.IsEnabled() && !is_element_level_search` (`:290-291`, where `IsEnabled()` is `LoadCounts().total > 0` and `total` is monotonically increasing — single store at `GrowingOffsetMapping.cpp:121`) and `if (has_offset_mapping && !bitset.empty())` (`:295`). Suggestion: either drop these three lines, or keep them as defensive code with a one-line comment noting that the guards above make this branch unreachable here. Note the same branch *is* live at `SearchOnIndex.cpp:93` (sealed path via `SealedOffsetMapping`), so the `NoFilter` state should stay on the interface either way. (raised by czs007, tinswzy)

- **`internal/core/src/segcore/ConcurrentVector.h:483`** — `offset_mapping_.Append(...)` (`:483-486`, which publishes via the release-store of `counts_` at `GrowingOffsetMapping.cpp:121`) happens *before* `emplace_to_at_least` and `set_data` write the actual vector data (`:503-507`), leaving a window in which the mapping already counts a valid row whose chunk is not yet allocated or written. This is currently safe because every production read path bounds its scan at or below the ack boundary (`SearchOnGrowing.cpp:241-243` freezes the plan/ack bound, `:332-335` converts it via `ValidCountBelow(logical_bound)`), and no production code uses `GetValidCount()` as a scan bound. The invariant is already documented on the interface (`OffsetMapping.h:78-91`, `GrowingOffsetMapping.h:89-99`); the remaining gap is that the *writer* side carries no such note. Suggested follow-up: add a comment at `ConcurrentVector.h:483` stating that the mapping is published ahead of the chunk data, so readers must use the ack bound (`ValidCountBelow`) and never the mapping's current size. This is unrelated to this PR because the `Append`-before-`set_data` ordering is unchanged context in the diff and already exists on master — this PR only replaces the per-row `is_valid` loop above it with `bulk_is_valid_range`, and in fact moves in the safer direction by removing the one reader that relied on the mapping's current size (`active_count = transformed_bitset.size()` → `ValidCountBelow(plan_bound)`). (raised by tinswzy)

PR #51961, opened 2026-07-28
enhance: [GoSDK] fix client telemetry correctness gaps and add request ID propagation

xiaofan-luan:
issue: #47281

## What

The Go SDK is currently the only SDK that implements the client side of `ClientTelemetryService`. Auditing it against the server implementation in `internal/rootcoord/telemetry` turned up five issues. The protocol-critical parts (config_hash algorithm, `last_command_timestamp` watermark, reply retry-on-failed-heartbeat, reservoir P99) were all correct — these are the gaps around them.

## Changes

**1. No way to correlate a client call with server-side logs.**

Adds opt-in `WithClientRequestID(ctx, id)` and `NewClientRequestID()`, sending the `client_request_id` header that `pkg/tracer/client_request_id_propagator.go` already consumes. pymilvus exposes the same thing via `CallContext(client_request_id=...)`; the Go SDK had no equivalent.

This is **deliberately opt-in rather than automatic**. The server propagator builds a remote span context with no sampled flag, and `SetTracerProvider` uses `ParentBased(TraceIDRatioBased(fraction))`, whose default `remoteParentNotSampled` is `NeverSample()`. Injecting the header unconditionally would therefore *exclude* every Go SDK request from trace sampling whenever `trace.sampleFraction > 0` — the opposite of the intent. Use it for log correlation; it is not a substitute for W3C `traceparent` propagation.

**2. Doomed heartbeat retried forever against an older server.**

```go
resp, err := m.client.telemetryService.ClientHeartbeat(ctx, req)
if err != nil {
    // Log error but continue - telemetry is best-effort
    return   // <- nothing was actually logged
}
```

Against a server predating `ClientTelemetryService` this fires every 30s for the lifetime of the client, with no backoff, no self-disable, and no diagnostics. Now latches off on `codes.Unimplemented`, exits the heartbeat loop, and exposes `IsSupported()` / `LastHeartbeatError()` — the `client/` module has no logger, so an accessor is the only way to surface a best-effort failure. `connectInternal` already handled `Unimplemented` this way; the heartbeat path just missed it.

Also passes `grpc_retry.Disable()` on the heartbeat, so a heartbeat against an `Unavailable` server costs one RPC instead of six with exponential backoff (the client-wide interceptor retries `Unavailable`/`ResourceExhausted`).

**3. `max_latency_ms` collected but never sent.**

`common.Metrics` has the field and the collector computed it, but every proto conversion site omitted it, so the server could never display it.

**4. `client_id` not stable across restarts.**

It was a fresh `uuid.New()` per process, so server-side telemetry history fragmented on every restart and `client:<id>` scoped commands could not target a long-lived client. Adds `TelemetryConfig.ClientID` to pin it. Unchanged behaviour when unset.

**5. Test-only fork that had already diverged.**

`processCommands` / `calculateConfigHash` / `collectProtoMetrics` / `getPendingReplies` were a parallel implementation reachable only from tests. They had already drifted: the test-only `calculateConfigHash` hashed `ID:Type`, while production hashes `ID + Type + payload`. Since `config_hash` must match the server byte-for-byte or configs get re-pushed on every heartbeat, the tests were validating an algorithm that never runs.

Deletes the fork, collapses the duplicated snapshot→proto conversion into a single `toProtoOperationMetrics()`, and repoints the tests at the production functions.

## Tests

- `TestIsValidTraceIDHex`, `TestNewClientRequestID`, `TestClientRequestIDFromContext`, `TestExtraInfoInjectsClientRequestID` — header is absent unless requested, malformed IDs are dropped, valid ones round-trip.
- `TestHeartbeatAgainstUnimplementedServer` — bufconn server with no services registered; asserts the latch and that the error is observable.
- `TestHeartbeatLoopExitsWhenUnsupported` — one-hour interval, so a loop ignoring the latch hangs the test.
- `TestMaxLatencyIsReported`, `TestTelemetryClientID`.

## Verification

`go test ./...` green across the `client/` module, `golangci-lint run` 0 issues, `gofmt` clean.

Verification is unit-level. **This has not been run against a live Milvus server** — in particular the end-to-end behaviour of `client_request_id` reaching server logs is reasoned from the propagator and sampler source, not observed.

### DIFF
--- a/client/milvusclient/interceptors.go
+++ b/client/milvusclient/interceptors.go
@@ -18,6 +18,8 @@ package milvusclient
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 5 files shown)


xiaofan-luan:

Issue #51963, opened 2026-07-28
[Bug]: client telemetry and tracing — sampling suppression, SDK correctness gaps, unusable async API, stale MEP


xiaofan-luan:
Umbrella issue for an audit of the client telemetry (`ClientTelemetryService`) and client-side tracing paths. Five problems, one PR.

---

## 1. `client_request_id` silently disables trace sampling (the actual bug)

A request carrying the `client_request_id` (or legacy `client-request-id`) header is **never sampled**, regardless of `trace.sampleFraction`, and the suppression propagates to every downstream component — the whole call tree is missing from Jaeger/OTLP.

It is silent because the header still does its main job: the trace ID reaches the logs. You keep log correlation and lose tracing with no signal.

**Cause.** `clientRequestIDPropagator.Extract` carries the client-supplied ID in a synthesized **remote parent span context** with no sampled flag (`pkg/tracer/client_request_id_propagator.go:52-56`):

```go
return trace.ContextWithRemoteSpanContext(ctx, trace.NewSpanContext(trace.SpanContextConfig{
    TraceID: traceID,
    SpanID:  spanID,
    Remote:  true,
    // TraceFlags defaults to 0x00 => not sampled
}))
```

`SetTracerProvider` uses `ParentBased(TraceIDRatioBased(fraction))` (`pkg/tracer/tracer.go:68-76`), and OTel's `ParentBased` defaults `remoteParentNotSampled` to `NeverSample()` (`sdk@v1.43.0 trace/sampling.go:200`). The `root` sampler — the only one that consults `sampleFraction` — is reached **only when there is no parent at all**. Synthesizing a parent takes the decision away from the operator and hardcodes it to drop.

The underlying design error: a client-supplied *correlation ID* was placed in the *parent span context*, a slot that in OTel also encodes an upstream sampling decision.

**Repro:**
```
1. Start Milvus with trace.exporter=jaeger and trace.sampleFraction=1.0
2. Search with no trace headers                    -> trace appears
3. Same search with client_request_id set to 32 hex
   (pymilvus: CallContext(client_request_id=...))  -> no trace, at any component
   The trace ID is still in the logs, which is why this goes unnoticed.
```

**Blast radius:** users who explicitly set a request ID — i.e. people actively debugging, exactly when traces matter most. The common support flow ("send us a request id") currently disables tracing for the request being investigated.

### 1b. Propagator ordering makes the guard dead code

```go
propagation.NewCompositeTextMapPropagator(
[...]

PR #51964, opened 2026-07-28
fix: client_request_id no longer disables trace sampling

xiaofan-luan:
issue: #51963

## Problem

A request carrying `client_request_id` was **never sampled**, regardless of `trace.sampleFraction`, and the suppression propagated to every downstream component — the whole call tree vanished from Jaeger/OTLP.

It was silent because the header still did its main job: the trace ID reached the logs. You kept log correlation and lost tracing without any signal.

`clientRequestIDPropagator` carries the client ID in a synthesized **remote parent span context** with no sampled flag (`client_request_id_propagator.go:52-56`). `ParentBased` maps *remote parent, not sampled* → `NeverSample` (`sdk@v1.43.0 trace/sampling.go:200`), and its `root` sampler — the only one that consults `sampleFraction` — is reached **only when there is no parent at all**. Synthesizing a parent therefore took the decision away from the operator and hardcoded it to drop.

Root design error: a client-supplied *correlation ID* was placed in the *parent span context*, a slot that in OTel also encodes an upstream sampling decision.

## Fix

Separate the two meanings.

1. The propagator marks the span context it synthesized, via an **in-process context value** — never a span-context field, so nothing leaks onto the wire or downstream.
2. `clientRequestIDSampler` samples a synthetic parent **as if it were a root span**, so `trace.sampleFraction` governs it. Everything else delegates to `ParentBased` unchanged, so a genuine upstream `traceparent` — including `sampled=0` — keeps standard W3C semantics.

The marker is validated against the current parent span ID rather than trusted alone. It stays in the context for the whole request, so without that check an in-process **child** span would be re-sampled independently of its parent and split the trace. `TestHasSyntheticParent` and `TestChildSpanFollowsParentDecision` cover this.

## Also: propagator ordering

```go
propagation.NewCompositeTextMapPropagator(
    clientRequestIDPropagator{},   // ran first
    otel.GetTextMapPropagator(),   // TraceContext + Baggage, ran second
)
```

`clientRequestIDPropagator.Extract` opens with `if trace.SpanContextFromContext(ctx).IsValid() { return ctx }` — a guard meant to yield to a real traceparent. Running **first**, nothing has been extracted yet, so it could never fire. The two only composed correctly because `TraceContext` ran later and overwrote the synthesized context.

Swapped the order so the guard does what it says, and precedence is explicit instead of incidental — otherwise a future reordering silently flips behavior.

## Behaviour matrix

| Incoming | Before | After |
|---|---|---|
| `client_request_id`, `sampleFraction=1.0` | **dropped** | sampled, trace ID = client's ID |
| `client_request_id`, `sampleFraction=0.0` | dropped | dropped (ratio governs) |
| `traceparent` sampled=1, fraction=0.0 | sampled | sampled (unchanged) |
| `traceparent` sampled=0, fraction=1.0 | dropped | dropped (unchanged, W3C respected) |
| both headers | traceparent wins | traceparent wins (now by design, not by accident) |
| neither | ratio | ratio (unchanged) |

## Tests

`pkg/tracer/sampler_test.go` drives the real composite propagator and a real `TracerProvider`, covering every row above plus the child-span and marker-scoping cases.

Reverting the sampler branch makes `TestClientRequestIDIsSampledByRatio/sampled_when_fraction_is_1.0` fail with `sampled=false` at `sampleFraction=1.0` — i.e. the tests reproduce the bug rather than merely describing it.

## Verification

`go test ./tracer/...` green (including the pre-existing propagator tests), `golangci-lint` 0 issues, `gofmt` clean, `go build ./...` green for the `pkg` module. No exported API changed — every new symbol is unexported and `SetTracerProvider`'s signature is untouched.

Verification is unit-level. **Not yet exercised against a live Jaeger/OTLP backend**; the end-to-end claim in the matrix is derived from the sampler and propagator semantics plus these tests.

## Note

One pre-existing cosmetic consequence is now visible: a trace started from `client_request_id` has a root span whose `parentSpanID` refers to the synthesized span, which is never reported. Jaeger and Zipkin both render this as a normal partial trace (the span becomes the root). Removing the fake parent entirely would mean seeding the trace ID through a custom `IDGenerator` instead — a larger change on the hot path for every span, which seemed disproportionate for this fix. Happy to follow up if reviewers prefer it.

### DIFF
--- a/pkg/tracer/client_request_id_propagator.go
+++ b/pkg/tracer/client_request_id_propagator.go
@@ -29,12 +29,48 @@ const (
 	clientRequestIDKey       = "client_request_id"
 )
 
+// syntheticParentKey marks a context whose remote span context was synthesized from a
+// client_request_id header rather than extracted from a real upstream traceparent. The
+// value is the synthesized SpanID, so the sampler can tell that span context apart from
+// any other parent it might later be asked about.
+//
+// This is an in-process marker only: it is a context value, never a span-context field, so
+// it is not propagated to downstream components. Those receive a normal traceparent whose
+// sampled flag already reflects the decision made here.
+type syntheticParentKey struct{}
+
+// withSyntheticParent records that spanID was synthesized locally rather than propagated.
+func withSyntheticParent(ctx context.Context, spanID trace.SpanID) context.Context {
+	return context.WithValue(ctx, syntheticParentKey{}, spanID)
+}
+
+// hasSyntheticParent reports whether the parent span context in ctx is one this package
+// synthesized from a client_request_id.
+//
+// It re-checks the span context rather than trusting the marker alone: the marker stays in
+// the context for the whole request, so a child span started later in the same process
+// would otherwise be misidentified as having a synthetic parent and be re-sampled
+// independently of its real parent.
+func hasSyntheticParent(ctx context.Context) bool {
+	if ctx == nil {
+		return false
+	}
+	spanID, ok := ctx.Value(syntheticParentKey{}).(trace.SpanID)
+	if !ok {
+		return false
+	}
+	psc := trace.SpanContextFromContext(ctx)
+	return psc.IsRemote() && psc.SpanID() == spanID
+}
+
 type clientRequestIDPropagator struct{}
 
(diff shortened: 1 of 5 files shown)


xiaofan-luan:

PR #51965, opened 2026-07-28
doc: refresh client telemetry MEP to match the implementation

xiaofan-luan:
The client telemetry MEP (`20260131-client_side_telemetry.md`) was written before the feature landed and never updated. Several sections describe interfaces that were never built. Anyone implementing the client half for another SDK from this document would produce something that does not interoperate with the server.

Found while auditing the Go SDK client implementation against `internal/rootcoord/telemetry`.

## Corrected

| Doc said | Reality |
|---|---|
| Commands `set_sampling_rate` / `enable_collections` / `update_config` | None exist. Real set: `push_config`, `collection_metrics`, `show_errors`, `show_latency_history`, `get_config` — and only `push_config` may be persistent |
| `TargetScope` = `"*"` / `"client_id:xxx"` / `"db:name"` | `global` / `client:<id>` / `database:<db>` |
| REST at `/api/v1/telemetry/*` | `/_telemetry/*` on the internal HTTP port; two endpoints were missing from the doc |
| RPCs on `service RootCoord` | Own `service ClientTelemetryService` |
| `ClientHeartbeatResponse.NewConfigHash` | No such field — the client recomputes the hash locally |
| "Persistent commands re-sent until explicitly deleted" | Suppressed whenever `config_hash` matches |
| "Telemetry disabled by default" | `DefaultTelemetryConfig()` has `Enabled: true` |
| `CommandHandler func(*ClientCommand) string` | Returns `*CommandReply` |
| clientID is "a stable UUID" | Stable per `Client`, **not** across process restarts |

Added the JSON payload schema for each of the five real command types, since `payload` is raw JSON and there was nothing to implement against.

## Added

Load-bearing details that were absent and that a port to another SDK cannot be written without:

- The exact `config_hash` algorithm. Client and server must agree **byte-for-byte** (sha256 over persistent configs sorted by ID, writing `ID + Type + Payload`, truncated to 16 hex chars, `""` for the empty set) or configs get re-pushed on every single heartbeat.
- The `last_command_timestamp` watermark, its strict `>` comparison, and the resulting exactly-once delivery — including that a failed command is never redelivered.
- Replies as the garbage-collection mechanism: any reply deletes the command *regardless of success*, and `ttl_seconds: 0` commands (which is what `/config` and `/history` push) leak forever if the client never replies.
- Persistent configs are JSON-**merged** on re-push to the same scope, not replaced.
- The hardcoded server limits, and explicitly that **there are no `milvus.yaml` / paramtable keys** for this feature.
- Metrics ingest truncation rules (drop `RequestCount == 0` collection entries, cap at 100 operations / 1 MB).
- The `legacy:<host>:<hash>` client-ID fallback and the fact that it collides same-host same-SDK processes into one entry, which breaks `client:` scoping.
- That `/config` and `/history` are fire-and-forget: they return a command ID, and the result must be read back from `GET /_telemetry/clients` by matching `command_id` in `command_replies`.

## Implementation status

Recorded two things the doc implied were universal:

- The client half is implemented in the **Go SDK only**. pymilvus, Java, Node.js and Rust do not implement it.
- `CommandRouter` validates payload shapes but is never invoked from `HandleHeartbeat`, and `PushCommand` does not validate `command_type` against it.

Status moved `Draft` → `Implemented`, and the header notes the refresh so the diff is not mistaken for a design change. No code changes.

### DIFF
--- a/docs/design-docs/design_docs/20260131-client_side_telemetry.md
+++ b/docs/design-docs/design_docs/20260131-client_side_telemetry.md
@@ -2,10 +2,17 @@
 
 - **Created:** 2026-01-31
 - **Author(s):** @xiaofanluan
-- **Status:** Draft
+- **Status:** Implemented
 - **Component:** SDK | Proxy | Coordinator
-- **Related Issues:** #46934
-- **Released:** [TBD]
+- **Related Issues:** #46934, #47281
+- **Implemented by:** #47523, #47542
+
+> This document was refreshed to match the implementation. Where the original draft
+> described interfaces that were never built (`set_sampling_rate`, `enable_collections`,
+> `update_config`, `/api/v1/telemetry/*`, `NewConfigHash`), those sections have been
+> replaced with what the code actually does. Source of truth is
+> `internal/rootcoord/telemetry/`, `internal/proxy/telemetry_*.go` and
+> `client/milvusclient/telemetry.go`.
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


xiaofan-luan:

PR #51966, opened 2026-07-28
enhance: add a command-reply endpoint and synchronous mode to the telemetry API

xiaofan-luan:
issue: #47281

## Problem

Telemetry commands are answered **asynchronously**, on the client's next heartbeat. `GET /_telemetry/clients/{id}/config` and `.../history` push a command and return:

```json
{"command_id": "...", "status": "pending", "message": "Command sent to client..."}
```

The reply then lands in a `StoredCommandReply` that **no endpoint exposes**. The only way to read it was to list every client and scan the `command_replies` array by hand, matching on command ID — and that array is smuggled through `ClientInfo.Reserved["command_replies"]` as a JSON string. `TelemetryManager.GetClientCommandReplies()` exists but has no production caller.

Net effect: the pull-from-client half of the API is impractical to consume from anything except the WebUI, which carries its own bespoke polling. Every other consumer — a script, a CLI, an agent — has to reimplement push → poll-all-clients → match → time out.

## Changes

**New:** `GET /_telemetry/commands/{commandId}/reply`

| Query param | Effect |
|---|---|
| `client_id` | Turns a scan of all clients into a targeted lookup |
| `wait` | Block until the reply arrives (e.g. `30s`, or a bare number of seconds) |

**Also:** the same `?wait=` on `/config` and `/history`, so a caller can push and collect in one request.

```bash
# one-shot
curl -u root:pw "$M/_telemetry/clients/$CID/config?wait=35s"
{"command_id":"...","client_id":"...","status":"done","waited_ms":2140,"reply":{...}}

# or push now, collect later
curl -u root:pw "$M/_telemetry/clients/$CID/config"
{"command_id":"abc","status":"pending",...}
curl -u root:pw "$M/_telemetry/commands/abc/reply?client_id=$CID&wait=30s"
```

## Design notes

**One response shape everywhere.** Sync and async return the same body, so callers do not special-case the two modes: check `status` (`"done"` / `"pending"`), read `reply` when done.

**Pending is 200, not 404 or an error.** A missing reply means "not answered *yet*" — and a client that will never answer is indistinguishable from one that is merely slow (replies are also evicted past 50 per client). Making that an error would force every caller to treat a normal state as an exception. Timing out likewise reports `"pending"`, with `waited_ms`.

**Bounded.** `wait` is clamped to 90s (a couple of heartbeat intervals; beyond that the client is not answering) and is additionally bounded by the request context, so a caller hanging up releases the handler promptly.

**Cheap to poll.** Reply lookups pass `IncludeMetrics=false`. Replies are populated independently of that flag, so polling does not drag every client's metrics along with it. Poll interval is 500ms — replies only ever land on a heartbeat, so faster buys nothing.

## Compatibility

Behaviour without `?wait=` is unchanged: `command_id` and `status: "pending"` are still returned, which is exactly what the WebUI branches on (`telemetry.html`: `if (data.status === 'pending')` → `pollForConfigReply(data.command_id, ...)`). The `message` string is reworded and `client_id` is added; nothing consumes either. The WebUI's own polling could later be replaced by the new endpoint, but this PR leaves it alone to keep the diff focused.

No proto changes — the endpoint is built on the existing `GetClientTelemetry` RPC.

## Tests

`internal/proxy/telemetry_command_reply_test.go`:

- `parseWaitParam`: duration strings, bare/fractional seconds, empty, negative, over-cap clamping, garbage rejection
- Reply found / not found / reply for a *different* command not matched
- `client_id` produces a targeted request **with `IncludeMetrics=false`** (asserted via `MatchedBy`)
- `wait` polls until the reply lands — mock returns pending then done; asserts it waited at least one poll interval and returned well before the budget
- `wait` expiring reports `pending`, not an error
- Context cancellation unblocks promptly and is not reported as a lookup failure

## Verification

`go test -tags dynamic,test` green for the new tests and the existing `Telemetry` suite in `internal/proxy`; `golangci-lint` 0 issues; `gofmt` clean.

Verification is unit-level — **not exercised against a live cluster**. The end-to-end timing (reply actually arriving within one heartbeat) is reasoned from the heartbeat protocol, not observed.

### DIFF
--- a/internal/http/router.go
+++ b/internal/http/router.go
@@ -165,6 +165,8 @@ const (
 	TelemetryClientHistoryPath = "/_telemetry/clients/:clientId/history"
 	// TelemetryCommandsPath is the path to manage client commands.
 	TelemetryCommandsPath = "/_telemetry/commands"
+	// TelemetryCommandReplyPath is the path to fetch a client's reply to a pushed command.
+	TelemetryCommandReplyPath = "/_telemetry/commands/:commandId/reply"
 	// TelemetryUIPath is the path for telemetry management web UI.
 	TelemetryUIPath = "/telemetry"
 )
--- a/internal/proxy/impl.go
+++ b/internal/proxy/impl.go
@@ -6509,6 +6509,7 @@ func (node *Proxy) RegisterRestRouter(router gin.IRouter) {
 	router.GET(http.TelemetryClientsPath+"/:clientId/config", telemetryAuth, getTelemetryClientConfig(node))
 	router.GET(http.TelemetryClientHistoryPath, telemetryAuth, getTelemetryClientHistory(node))
 	router.POST(http.TelemetryCommandsPath, telemetryAuth, postTelemetryCommand(node))
+	router.GET(http.TelemetryCommandReplyPath, telemetryAuth, getTelemetryCommandReply(node))
 	router.DELETE(http.TelemetryCommandsPath+"/:commandId", telemetryAuth, deleteTelemetryCommand(node))
 }
 
--- a/internal/proxy/telemetry_command_reply.go
+++ b/internal/proxy/telemetry_command_reply.go
@@ -1,0 +1,164 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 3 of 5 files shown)


xiaofan-luan:

Issue #648, opened 2019-12-02
remove redundant table flag


cydrain:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
one of following 2 flags is redundant,
constexpr int64_t FLAG_MASK_NO_USERID = 0x1;
constexpr int64_t FLAG_MASK_HAS_USERID = 0x1 << 1;

suggest to only keep FLAG_MASK_HAS_USERID

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

cydrain:
When table is created, the flag = 0;
when add vectors into table without IDs, flag = 1;
when add vectors into table with IDs, flag = 2;
There are 3 states, we cannot indicate 3 states using 1 bit.

Close this issue.

PR #51967, opened 2026-07-28
enhance: fix client telemetry and tracing gaps found auditing the Go SDK against the server

xiaofan-luan:
issue: #51963

Audit of the client telemetry (`ClientTelemetryService`) and client-side tracing paths, comparing the Go SDK against `internal/rootcoord/telemetry` and `pkg/tracer`. Five commits, reviewable independently.

Most of the protocol core holds up — `config_hash`, reply-retry-on-failed-heartbeat, reservoir P99. These are the gaps around it.

**Correction:** earlier revisions of this description also listed the `last_command_timestamp` watermark as already correct. It is not. The cursor is a millisecond timestamp compared strictly, so a command created in the *same millisecond* as one the client has already processed is filtered out of every later heartbeat and never delivered — it survives only until its TTL, with no error to whoever pushed it. Reproduced in isolation: the second command comes back in zero heartbeats. Relaxing to `>=` does not fix it, because the client prunes its executed-ID set at the watermark and would re-run commands sitting exactly on it; a correct fix needs a per-command cursor (a monotonic sequence, or `(timestamp, command_id)` compared lexicographically) on both sides. That is a protocol change and out of scope here — it is pre-existing, documented in the design doc, and tracked in #51963.

---

## 1. `fix:` `client_request_id` no longer disables trace sampling

**The one behavioural bug.** A request carrying `client_request_id` was **never sampled**, whatever `trace.sampleFraction` said, and the suppression propagated to every downstream component. Silent, because the trace ID still reached the logs — you kept log correlation and lost tracing with no signal.

`clientRequestIDPropagator` carries the client ID in a synthesized **remote parent span context** with no sampled flag. `ParentBased` maps *remote parent, not sampled* → `NeverSample` (`sdk@v1.43.0 trace/sampling.go:200`), and its `root` sampler — the only one consulting `sampleFraction` — is reached **only when there is no parent at all**.

Root design error: a client-supplied *correlation ID* was placed in the *parent span context*, a slot that in OTel also encodes an upstream sampling decision.

**Fix.** Separate the two meanings. The propagator marks the context it synthesized (an **in-process context value**, never a span-context field, so nothing reaches the wire), and `clientRequestIDSampler` samples a synthetic parent **as if it were a root span**. Everything else delegates to `ParentBased`, so a genuine `traceparent` — including `sampled=0` — keeps standard W3C semantics.

The marker is validated against the current parent span ID rather than trusted alone: it stays in the context for the whole request, so without that check an in-process **child** span would be re-sampled independently of its parent and split the trace.

Also swapped the composite propagator order. `clientRequestIDPropagator.Extract` opens with an "already have a span context" guard meant to yield to a real `traceparent`, but running **first** it could never fire — the two only composed correctly because `TraceContext` ran later and overwrote it. Now precedence is explicit rather than incidental.

| Incoming | Before | After |
|---|---|---|
| `client_request_id`, fraction=1.0 | **dropped** | sampled, trace ID = client's ID |
| `client_request_id`, fraction=0.0 | dropped | dropped (ratio governs) |
| `traceparent` sampled=1, fraction=0.0 | sampled | unchanged |
| `traceparent` sampled=0, fraction=1.0 | dropped | unchanged (W3C respected) |
| both | traceparent wins | traceparent wins (by design now) |

Reverting the sampler branch makes `TestClientRequestIDIsSampledByRatio/sampled_when_fraction_is_1.0` fail with `sampled=false` at `sampleFraction=1.0` — the tests reproduce the bug, not just describe it.

## 2. `enhance:` Go SDK telemetry correctness

- **`Unimplemented` was retried forever at full rate.** Against an older server the heartbeat fired every 30s indefinitely, and the error was dropped despite a comment claiming it was logged. Now the interval **backs off exponentially** while the server keeps answering `Unimplemented` — doubling per consecutive rejection, capped at 30 minutes, never shortening below the configured interval — and the streak resets on the first reply. Deliberately *not* latched off: a client load-balances across proxies, so mid rolling-upgrade one heartbeat can land on an old proxy while the rest of the cluster already supports the service, and a client that gave up would stay dark until restarted. `IsSupported()` / `LastHeartbeatError()` expose the state — the `client/` module has no logger, so an accessor is the only way to surface a best-effort failure. `IsSupported()` is optimistic: it returns true before the first heartbeat, so it means "no evidence of an old server", not "confirmed working". Also opts the heartbeat out of the client-wide retry interceptor (six RPCs with backoff → one against an `Unavailable` server).
- **`max_latency_ms`** was collected but omitted at every proto conversion site, so the server could never show it.
- **`TelemetryConfig.ClientID`** to pin a client identity across process restarts; unchanged when unset.
- **Opt-in `WithClientRequestID` / `NewClientRequestID`**, matching pymilvus's `CallContext(client_request_id=...)`. Kept opt-in even though commit 1 fixes the server: an SDK cannot assume the server has the fix, and sending it unconditionally would silently disable tracing against any older cluster.
- **Deleted a test-only fork that had already diverged.** `processCommands` / `calculateConfigHash` / `collectProtoMetrics` / `getPendingReplies` were reachable only from tests, and the test-only `calculateConfigHash` hashed `ID:Type` while production hashes `ID + Type + payload`. Since `config_hash` must match the server byte-for-byte, the tests were validating an algorithm that never runs. Collapsed the duplicated proto conversion into one `toProtoOperationMetrics()` and pointed the tests at production code.

## 3. `enhance:` command-reply endpoint

Commands are answered on the client's next heartbeat, but the reply landed in a `StoredCommandReply` **no endpoint exposed** — smuggled through `ClientInfo.Reserved["command_replies"]` as a JSON string, with `GetClientCommandReplies()` having no production caller. The only way to read one was to list every client and scan by hand.

Adds `GET /_telemetry/commands/{commandId}/reply`, with `?client_id=` to look up one client instead of scanning every cached one.

```bash
curl -u root:pw "$M/_telemetry/commands/$CMD/reply?client_id=$CID"
{"command_id":"...","client_id":"...","status":"done","responded":1,"observed_clients":1,"replies":[...]}
```

Deliberately **no server-side blocking mode**. An earlier revision offered `?wait=`; it was removed after review. Each lookup carries every matching client's *entire* stored reply history (50 entries, encoded into `ClientInfo.Reserved` regardless of `IncludeMetrics`), the 1MiB payload cap applies only to built-in handlers, and `authorizationEnabled` defaults to false — so a blocking mode turned one unauthenticated request into dozens of full-history transfers. A caller that wants to wait polls on its own schedule, which keeps one request to one internal query. `GetClientTelemetryRequest.command_id` (milvus-io/milvus-proto#647) would remove the amplification at its source; using it is gated on a proto bump.

`pending` is 200, not an error — a reply that has not arrived is indistinguishable from one that never will.

No proto changes — built on the existing `GetClientTelemetry` RPC.

## 4. `doc:` MEP refreshed to match the implementation

The MEP was written before the feature landed and never updated; anyone porting the client half from it would build something that does not interoperate. Corrected the command types (three of the documented ones were never implemented), `TargetScope` values, REST paths, service name, the non-existent `NewConfigHash`, persistent-command resend behaviour, the default-enabled flag, and the `CommandHandler` signature.

Added what was missing and is load-bearing for a port: the exact `config_hash` algorithm, the watermark's exactly-once consequence, reply-as-garbage-collection, the `ttl_seconds: 0` leak, persistent-config JSON merge, the hardcoded server limits (and that there are **no** `milvus.yaml` keys), and the `legacy:` client-ID collision.

## 5. `enhance:` alignment

The four above were written independently and contradict each other once bundled — the SDK comment described sampling behaviour that commit 1 fixes, and the MEP documented the absence of the endpoint commit 3 adds. Reconciled.

---

## Compatibility

- **No proto changes and no dependency changes.** `go.mod` and `go.sum` are byte-identical to master across all four modules. (Earlier revisions bumped milvus-proto; that is fully reverted — see milvus-io/milvus-proto#647 and #648 for the upstream fields, which milvus does not yet consume.)
- No exported Go API removed, `SetTracerProvider`'s signature unchanged; every new symbol in `pkg/tracer` is unexported.
- Telemetry REST behaviour without `?wait=` is unchanged: `command_id` and `status: "pending"` are still returned, which is what the WebUI branches on.
- Go SDK: `client_request_id` is off unless the caller opts in, so no request gains a header by default.

## Verification

`go test` green for the `client/` module, `pkg/tracer`, and the `Telemetry` suite in `internal/proxy`; `golangci-lint` 0 issues across all touched packages; `gofmt` clean.

**All verification is unit-level — nothing here has been run against a live cluster or a live Jaeger/OTLP backend.** The sampling matrix is derived from the sampler/propagator semantics plus tests that fail without the fix; the end-to-end claim that a reply arrives within one heartbeat is reasoned from the protocol, not observed.

## Not addressed

Found during the audit, left out deliberately: client telemetry exists only in the Go SDK (pymilvus/Java/Node/Rust do not implement it); there are no `milvus.yaml`/paramtable keys for the feature; the REST surface is Basic Auth with no RBAC; non-persistent commands live only on the RootCoord that received the push; `ttl_seconds: 0` commands leak if a client never replies; and `PushCommand` does not validate `command_type`. All recorded in #51963.

---

## Review follow-ups

Post-review fixes are in the commits after the five above; the substantive ones:

- **Broadcast commands survived only until the first reply.** `cleanupRepliedCommands` deleted any replied one-time command regardless of scope, so on a `global` command the fastest client's reply cancelled delivery to everyone else — a fleet-wide `collection_metrics` silently reached part of the fleet. Deletion on reply is now limited to `client:` scope; `global`/`database:` expire on their TTL.
- **`?wait=` was removed entirely** (an earlier revision added it, then fixed its deadline handling and its stopping condition; all of that is gone). Each lookup carries every matching client's *entire* stored reply history, the 1MiB payload cap covers only the built-in handlers, and `authorizationEnabled` defaults to false — so a blocking mode turned one unauthenticated request into dozens of full-history transfers. Collecting a reply is always deferred now.
- **The milvus-proto bump was reverted.** Earlier revisions bumped it for `GetClientTelemetryRequest.command_id`; it dragged in three rounds of unrelated collateral (regenerated mocks, a field-ID collision in `planparserv2`'s test schema, and a broken `fastpb` hand-decoder contract for a new `schemapb.IDs` oneof variant). `go.mod` and `go.sum` are byte-identical to master again.
- **`observed_clients`** (was `known_clients`) — the count reflects what the lookup scanned, not what the command targeted, so it never established completeness and no longer claims to.
- **`ClientMetricsCache` data race** — the heartbeat wrote five fields while admin readers polled them unsynchronized. Now behind an `RWMutex`, with readers cloning/encoding off the lock. Pre-existing; fixed here because this PR adds another poller.


czs007:

Adversarial review of the new telemetry command/reply path surfaced two medium-severity defects in the reply endpoint plus four low-severity issues in error classification, command TTL semantics, and documentation.

## Medium

**`internal/proxy/telemetry_command_reply.go:93` — the reply endpoint silently returns one arbitrary reply for commands that many clients answer**

`findCommandReply`'s nested loop returns on the first `command_id` match, and the outer iteration order comes from `sync.Map.Range` (`manager.go:663`), which is unspecified. A command with neither `target_client_id` nor `target_database` is stored with scope `global` (`command_store.go:277-282`) and `matchesScope` returns true for every client (`manager.go:601-603`), so every connected client receives and answers it with the same `command_id` — delivery is not raced against reply collection, since a client receives the command on one heartbeat and answers on the next, well before the first reply triggers `cleanupRepliedCommands` (`manager.go:401-403`). Failure: `POST /_telemetry/commands` with `command_type=show_errors` and empty `target_client_id`/`target_database` against N>1 connected clients, then `GET /_telemetry/commands/{id}/reply` without `client_id` → `status:"done"` with a single arbitrary reply and `client_id` (replaying the request may return a different client's data), the other N-1 answers dropped with no "more exist" signal, so an operator reads one sample as the cluster-wide answer.
*Suggestion:* return `replies` as an array with per-entry `client_id`, and/or a `responded`/`total_known_clients` count so callers can distinguish complete from partial. If the single-valued shape is intentional, require `client_id` for non-client-scoped commands instead of picking a winner. *(raised by czs007)*

**`internal/rootcoord/telemetry/manager.go:701` — every poll ships the target client's entire stored reply set (up to ~50 MB), repeated up to ~181 times per request**

`manager.go:701-713` unconditionally JSON-encodes all of `cache.CommandReplies` into `ClientInfo.Reserved["command_replies"]` on every `GetClientTelemetry` — outside the `IncludeMetrics` guard at `:697` — and the proxy unmarshals it back each time (`telemetry_models.go:100-105`). The cache keeps 50 replies (`manager.go:425,460-462`) and each payload is capped at 1 MB client-side (`client/milvusclient/telemetry.go:1538,1768-1774`), so the blob's upper bound is ~50 MB; `waitForCommandReply` repeats the lookup every 500 ms for up to 90 s (`telemetry_command_reply.go:36,109,132-133`) even for a fully targeted lookup. Failure: a client that has answered many large `show_errors`/`show_latency_history` commands stops heartbeating (hung process or partition) and an operator issues `GET /_telemetry/clients/{id}/config?wait=90s` → ~181 rounds of marshal + gRPC transfer + unmarshal of ~50 MB for one HTTP request, occupying rootcoord CPU and the proxy↔rootcoord link, still ending in `status:"pending"`. Note the 500 ms constant contradicts its own comment at `telemetry_command_reply.go:34-36` ("Replies only ever land on a heartbeat, so polling faster buys nothing") against a 30 s default heartbeat (`telemetry.go:68`) — 60× redundant. To be fair on scale: shipping the whole blob unconditionally is pre-existing (the WebUI already polls it at 1 s in `telemetry.html:2158`); what this PR adds is the server-side 500 ms × 181 amplifier, roughly 3× the existing path.
*Suggestion:* add a server-side filter (a `command_id` field on `GetClientTelemetryRequest`, or a dedicated `GetCommandReply` RPC) so a poll returns at most the requested reply; failing that, cap the encoded `command_replies` size and widen `telemetryReplyPollInterval` to seconds. *(raised by czs007)*

## Low

**`internal/proxy/telemetry_http_handler.go:532` — two commands that previously never expired are now silently subject to a 1-hour TTL**

`getTelemetryClientHistory` (`:532`) and `getTelemetryClientConfig` (`:592`) both pass `TtlSeconds: 0`. Previously the store persisted 0 verbatim and every expiry check is `cmd.TTLSeconds > 0 && ...` (`command_store.go:411,437,501`), i.e. 0 meant never expire — exactly what the old comment ("Keep until client replies") described. The new `resolveCommandTTL` (`command_store.go:141-145`) maps 0 to `defaultCommandTTLSeconds = 3600` (`:133`) and stores it at `:343`; these two call sites only changed their comments. Failure: a client whose `heartbeat_interval_ms` was pushed to 7200000 (2 h — neither side bounds it, `client/milvusclient/telemetry.go:1347-1356`), or simply one partitioned for over an hour, then `GET /_telemetry/clients/{id}/config` → the command is reaped as expired before the client ever fetches it, the endpoint returns `status:"pending"` forever, with no log explaining why; pre-change the same request would have been delivered on the next heartbeat.
*Suggestion:* make the semantics explicit at these two call sites — pass `-1` to keep the old "wait for the client" behaviour, or document on the endpoints that a request is silently discarded once the client has been offline for an hour and must be re-issued. The TTL backstop itself is reasonable; only the two endpoints that relied on immortality need to choose. *(surfaced during verification)*

**`internal/rootcoord/telemetry/manager.go:806` — "client not currently connected" is classified as a non-retriable `InputError` though it is a transient cold-cache condition**

`manager.go:805-813` returns `WrapErrParameterInvalidMsg` on a `clientMetrics.Load` miss, and `ErrParameterInvalid` is `1100, retriable=false, InputError` (`pkg/util/merr/errors.go:193`). `m.clientMetrics` is rootcoord-local memory rebuilt only from heartbeats (`manager.go:370-389`), so after a restart or failover it is empty for up to a full heartbeat interval (30 s by default, unbounded by config). Failure: right after a mixcoord/rootcoord restart, a provisioning script pushes a persistent client-scoped `push_config` for a legitimately pinned-ID client whose first heartbeat has not landed → rejection as code 1100 / non-retriable / `InputError`, with text blaming the client's `ClientID` rather than the cold server cache; the script gives up instead of retrying a request that would succeed seconds later, and the failure lands in input-error rather than system-error metrics.
*Suggestion:* split the branches — keep `ParameterInvalid` for the known-but-unstable client (`:815`), and use a retriable system error (`ErrServiceNotReady`/`ErrServiceUnavailable`) for the unknown client, since the server cannot distinguish "never existed" from "hasn't heartbeated yet". This mirrors the pattern already documented for `ErrCollectionNotFound`/`ErrDatabaseNotFound` at `errors.go:73,143`. *(raised by czs007)*

**`internal/proxy/telemetry_http_handler.go:201` — the `wait` doc points operators at a config that does not govern this route**

(Line correction: the text is at `:200-202`, not `:427`.) The comment states that `proxy.http.writeTimeout` still applies, but `RegisterRestRouter` is mounted on `metricsGinHandler` and published via `mhttp.Register` (`internal/distributed/proxy/service.go:168-185`), i.e. the management server at `internal/http/server.go:297`, which sets only a hardcoded 10 s `ReadTimeout` and no configurable write timeout. `Params.HTTPCfg.WriteTimeout` applies only to `s.httpServer` (`service.go:230-234`), serving `/v1` and `/v2/vectordb`. There is no behavioural impact — the 10 s `ReadTimeout` does not cut a 90 s handler, because `net/http` clears the read deadline when starting the background read for a bodyless request, so the 90 s cap is genuinely reachable — but the PR's own design doc says this correctly (`20260131-client_side_telemetry.md:60`, "Served on the internal HTTP port (9091 by default)"), so the two contradict each other and operators would be sent to tune a knob that cannot affect this route.
*Suggestion:* reword to state that the route is served on the 9091 management port with no configurable write timeout. *(raised by czs007)*

**`docs/design-docs/design_docs/20260131-client_side_telemetry.md:55` — the design doc calls telemetry opt-in, but it is on by default**

The new paragraph says telemetry is opt-in at the level of whether a caller constructs a `TelemetryConfig` at all, yet `client/milvusclient/client.go:91-92` unconditionally creates and starts the manager, `NewClientTelemetryManager` substitutes `DefaultTelemetryConfig()` when the config is nil (`telemetry.go:528-530`), that default has `Enabled: true` (`telemetry.go:65-70`), and `Start()` only returns early when `Enabled == false` (`telemetry.go:558-565`). A caller who follows the doc and omits `TelemetryConfig` still sends heartbeats and metrics every 30 s.
*Suggestion:* state that telemetry is enabled by default and must be disabled explicitly via `TelemetryConfig{Enabled: false}`. *(raised by chyezh)*

- **`internal/rootcoord/telemetry/manager.go:702` — `ClientMetricsCache` is read and written without any lock.** Writes at `manager.go:384-389` (`ClientInfo`, `LatestMetrics`, `ConfigHash`, `LastCommandTS`) and in `processCommandReplies` (`:444` append, `:461` reslice) race reads at `:680`, `:698`, `:702/:710` (`json.Marshal(cache.CommandReplies)`) and `:718`; only `LastHeartbeat`, `AccessedDatabases` and `ClientIDStable` are concurrency-safe (`:87-102`). `IncludeMetrics:false` does not narrow the exposure — `:701-713` and `:718-724` both sit outside the `:697` guard, so even `findCommandReply`'s deliberately cheap call races `LatestMetrics` and `CommandReplies`; any locking fix must cover `CommandReplies`, `LatestMetrics` and `ClientInfo` together. Failure: a heartbeat rewrites the `CommandReplies` slice header (`:444`/`:461`) while another goroutine marshals it at `:710` → a torn ptr/len pair yields out-of-bounds reads of the old backing array and stale `*StoredCommandReply` pointers, producing garbled or duplicated reply data at best and a nil-pointer panic in the RootCoord RPC handler at worst; `go test -race` reports a DATA RACE. Unrelated to this PR: not one of the involved read or write lines is touched here (the only additions to `manager.go` are the atomic `ClientIDStable` field and `validatePersistentTarget`, which deliberately reads only atomics — see the comment at `:797-799`), and the path is already polled on master, where `telemetry.html:2158` hits `GET /_telemetry/clients` once per second and `manager.go:704` already marshals `cache.CommandReplies`. This PR raises the rate to one poll per 500 ms, but the race exists without it. *(raised by czs007)*

PR #51140, opened 2026-07-07
feat: support bloom_match approximate membership filter expression

xiaofan-luan:
issue: #51139
design doc: [docs/design-docs/design_docs/20260707-bloom-filter-expression.md](https://github.com/xiaofan-luan/milvus/blob/feat-bloom-filter-expr/docs/design-docs/design_docs/20260707-bloom-filter-expression.md) (included in this PR)
depends on: milvus-io/milvus-proto#632 (merged) — adds `bytes_val` to `TemplateValue`, pinned here via a Go pseudo-version

> History squashed to a single commit on 2026-07-16 (reviewer request); same content as the previous 8-commit series plus the JSON-path extension below.
> Rebased over the GoSDK decoupling (#51420) on 2026-07-17: the `sbbf` package moved from `pkg/util/sbbf` into the standalone client module as `client/v3/sbbf` (the client must not depend on `milvus/pkg` anymore), dropping its `merr` usage (plain errors inside the client module; the proxy boundary classifies via `merr.WrapErrParameterInvalidErr`). Golden-vector fixtures moved along; C++ conformance tests read the new path. Also reconciled with #51521 (template fill now descends into MATCH_*) — element-level `bloom_match` inside `MATCH_*`/`element_filter` is **rejected at the proxy** rather than evaluated (see Probe targets).
> Rebased onto master (#51540) on 2026-07-22: the index-only reverse-lookup path now follows #51540's callback-cursor convention. The previously-bundled #51286 pruning optimization was **dropped** (see below); `bloom_match` results are unchanged.

## What

Adds `bloom_match`, an approximate set-membership filter expression — a compact alternative to exact `IN [...]` for large membership sets:

```
bloom_match(field, {blob})                              -- probably-in-set
not bloom_match(field, {blob})                          -- exclude semantics
bloom_match(meta["user_id"], {blob})                    -- JSON path (incl. dynamic fields)
```

The filter blob is **built on the client** (`milvusclient.NewBloomFilterBlob(members, fpr)`), shipped as a native protobuf `bytes` template value, and the plan carries only the blob. segcore evaluates it as a dedicated `BloomFilterExpr` physical expression on the data path — one XXH64 + one cache-line block probe per row; sealed and growing segments behave identically. The `CallExpr` surface syntax never reaches querynodes (rewritten to a dedicated `BloomFilterExpr` plan node at the proxy).

[...]

PR #51968, opened 2026-07-29
feat: support roaring_match exact integer membership filter expression

xiaofan-luan:
issue: #52094
design doc: [docs/design-docs/design_docs/20260714-roaring-exact-membership-expression.md](docs/design-docs/design_docs/20260714-roaring-exact-membership-expression.md)

## Summary

Adds `roaring_match(field, {bitmap})`, an exact membership expression for top-level `INT8`, `INT16`, `INT32`, and `INT64` fields. The Go client builds an MRB1 envelope around a portable Roaring64 bitmap, Proxy validates and embeds it in the expression plan, and QueryNode independently validates, decodes, and evaluates it.

Unlike `bloom_match`, Roaring membership has no false positives and is allowed in delete expressions.

## Scope and behavior

- Adds Go SDK bitmap construction and Search, Query, and Delete template-parameter support.
- Preserves signed integer values through their int64 two-complement bit pattern.
- Supports raw-field probing and bounded scalar-index reverse lookup.
- Rejects unsupported field types, malformed blobs, and element-filter placement.
- Allows membership expressions in `ScoreFunction.filter`, matching the behavior `bloom_match` shipped with in #51140. `PhyRoaringFilterExpr` consumes offset input, so the boost path evaluates it natively against the candidate offsets rather than materializing a whole-segment bitset; NULL rows fold to FALSE and receive no boost. Scorer-filter blobs are charged to the same per-request budget as main-predicate blobs.
- Shares one configurable 64 MiB per-filter body limit between Bloom and Roaring.
- Shares one 128 MiB aggregate serialized membership-plan limit across Search, HybridSearch, Query, and complex Delete.
- Keeps the fixed Roaring limits of 262,144 high containers and a 64 MiB decoded-memory estimate per filter.
- Marshals complex-delete plans once before shard fan-out.
- Replaces Bloom and Roaring bytes with `<blob>` in Proxy plan logs; QueryNode plan-proto debug rendering is bounded and removes membership and unknown-field bytes.
- Prints the template placeholder, never the template's value. The expression text carries only `{ids}` -- the values travel in a separate map -- so it is logged unchanged and ordinary diagnostics are untouched. Template values are elided wherever they could surface: request logs (`elideRequestForLog`, covering Search + sub-requests, Query, Delete, HybridSearch), plan logs once the blob is materialized into the plan (`RedactPlanForLog` at both Proxy sites, `PlanProtoDebugString` at both QueryNode sites), and errors, which report the template name and never its value. Slow-query snapshots store expression text only.

## Compatibility

The public plan protobuf gains `RoaringFilterExpr`.

### Breaking change: `DeleteOption.Request()` (Go SDK, `client/v3`)

```go
// before
Request() *milvuspb.DeleteRequest
// after
Request() (*milvuspb.DeleteRequest, error)
```

This **is** a source-breaking change to a published (`client/v3.0.0-beta`) interface, not just an internal detail. It breaks two groups:

- callers that use the returned request directly — `req := opt.Request()` must become `req, err := opt.Request()`;
- **any third-party `DeleteOption` implementation**, which no longer satisfies the interface until its `Request()` is updated.

`Client.Delete`'s own behavior for valid input is unchanged, but that does not cover the compile-time break above — calling it out explicitly.

Kept rather than reverted, for two reasons:

- `SearchOption.Request()` and `QueryOption.Request()` in this same module already return `(request, error)`. Delete was the outlier; this converges the SDK rather than adding a new shape.
- The discarded error mattered on the delete path specifically. With the old signature a template value that fails conversion is silently omitted from `ExprTemplateValues`, so the request reaches the server with a predicate referencing an unbound `{key}`. Surfacing it client-side is the point of the change.

### Configuration

The two Bloom-specific proxy params are renamed to their shared membership names:

| old | new |
| --- | --- |
| `proxy.maxBloomFilterSize` | `proxy.maxMembershipFilterSize` |
| `proxy.maxBloomFilterPlanSize` | `proxy.maxMembershipFilterPlanSize` |

The old YAML keys remain accepted as `FallbackKeys`, so deployments that already set them need no change. The corresponding Go fields (`ProxyCfg.MaxBloomFilterSize`, `ProxyCfg.MaxBloomFilterPlanSize`, `DefaultMaxBloomFilterPlanSize`) are removed rather than kept as aliases: they were added in #51140 on 2026-07-28 and are in no release tag.

## Verification

Passed in this checkout:

- `internal/parser/planparserv2/...`, `internal/proxy/accesslog/...`
- `client/v3/roaringfilter`
- targeted Roaring/template tests in `client/v3/milvusclient`
- `pkg/v3/util/roaringfilter`
- proxy membership configuration tests in `pkg/v3/util/paramtable`
- `git diff --check`
- C++ `all_tests` built from scratch in this checkout; `RoaringFilterExpr*`, `RoaringMembership*`, `PlanProto*`, `BloomFilterExpr*`, `BoostScoreRunner*`, `WeightScorer*`, `RandomScorer*` -- 82 tests, all passing, including the new scorer-execution test.

That build also surfaced a compile break in `PlanProtoTest.cpp`: it used `roaring::Roaring64Map` while including only `common/RoaringMembership.h`, which pulls in the 32-bit `roaring.hh`. Fixed by including `<roaring/roaring64map.hh>`.


czs007:

One-sentence overview: this PR's new `RoaringFilterExpr` blob reaches two debug-logging paths that were never extended to redact it, and two smaller issues (an unpruned index reverse-lookup and member-key values in validation errors) round out the findings.

## Medium

**`internal/core/src/query/PlanProto.cpp:970` — `LogPlanProtoDebug` leaks the exact roaring member set via `ShortDebugString`.**
The 4096-byte guard (`kMaxPlanDebugBytes`) was sized for `bloom_match` on the assumption that "a blob with many values is necessarily > 4096 bytes". Roaring breaks that assumption: the PR body's own table shows 10M contiguous ids encode to a 2.2 KiB MRB1 body, so the whole plan lands under the threshold and takes the dump branch. Unlike SBBF, roaring is an exact, reversible encoding — the octal-escaped `bitmap_blob` fully reconstructs the caller's member set. Failure: with QueryNode debug logging on, a `search`/`query` carrying a dense or contiguous `roaring_match` set (10M contiguous ids → 2.2 KiB body, plan ≤ 4096 bytes) → the complete `bitmap_blob` is written to the QueryNode log and can be decoded offline into all 10M member values, then retained by log collection. This violates this PR's own design doc MUSTs (`docs/design-docs/design_docs/20260714-roaring-exact-membership-expression.md:471`, `:495`). Suggestion: strip the blob fields of `RoaringFilterExpr`/`BloomFilterExpr` before dumping, or reuse the "size + cardinality only" strategy this PR already applies at `internal/core/src/expr/ITypeExpr.h:966-976`, rather than gating on a byte threshold. Both call sites are affected (`:982` search, `:1006` retrieve/delete). *(raised by xaxys, xiaocai2333)*

**`internal/parser/planparserv2/bloom_match.go:441` — `RedactPlanForLog` does not cover roaring plans, so proxy debug logs print `bitmap_blob`.**
`collectBloomFilterExprs` only matches `*planpb.Expr_BloomFilterExpr`, so there are two holes: (1) a pure-roaring plan hits `if len(blooms) == 0 { return p.plan.String() }` and returns the unredacted string; (2) a mixed bloom + roaring plan also leaks — the replacement branch elides only the bloom blobs while `RoaringFilterExpr.BitmapBlob` is still printed verbatim by `plan.String()`. There is no byte threshold here as there is on the C++ path. Failure: proxy at debug level receives a Search/HybridSearch/Query containing `roaring_match` (alone or mixed with `bloom_match`) with a blob near the `proxy.maxRoaringFilterSize` default of 32 MiB → the "proxy init search request" log emits the full escaped blob, leaking the exact member set and building a ~128 MiB string per log call (once per sub-request under HybridSearch), spiking memory and CPU. Call sites: `internal/proxy/task_search.go:708` and `:1127`. Suggestion: generalize `collectBloomFilterExprs` to collect both blob-bearing node types (keeping the node set in sync with `hasRoaringFilterExpr`), or have `RedactPlanForLog` reuse the traversal in `PlanContainsRoaringFilter`. *(raised by xaxys)*

## Low

**`internal/core/src/exec/expression/RoaringFilterExpr.cpp:294` — index-only path does `Reverse_Lookup` for the whole batch without applying the upstream predicate.**
The no-offset branch of `ExecVisitorImplForIndex` builds `batch_offsets` covering the entire `[current_index_chunk_pos_, +real_batch_size)` range, and `ProcessIndexLookupByOffsets` performs an unconditional `Reverse_Lookup` per offset (`internal/core/src/exec/expression/Expr.h:673-675`) before the callback — but the `bitmap_input` check happens inside the callback at `:272`, so rows already excluded upstream still pay for an index lookup. Failure: on a sealed segment where the raw data is not loaded but a fast-reverse-lookup scalar index is pinned (`queryNode.indexOffsetCacheEnabled=true`), a query shaped like `high_selectivity_pred and roaring_match(id, {bm})` → every batch performs `real_batch_size` reverse lookups instead of one per surviving row (≈100× wasted lookups at 1% selectivity). This is filed as low and non-blocking because the code is line-for-line structurally identical to the existing `PhyBloomFilterExpr` index path (`internal/core/src/exec/expression/BloomFilterExpr.cpp:330-336`), and the real fix is not local: `ProcessIndexLookupByOffsets` writes results via `res[i]` (candidate-list index), so skipping rows requires changing the shared helper's interface to support sparse candidates. Suggestion: leave a TODO here pointing at the shared helper so this gets fixed together with the bloom path. *(raised by xaxys, xiaocai2333)*

**`internal/core/src/common/RoaringMembership.cpp:267` — validation errors of the form `key %d follows %d` carry member key values, on both the C++ and Go sides.**
When roaring deserialization validation fails, the high key / Roaring32 child key — i.e. the high bits of member values — is formatted straight into the error string and thence into logs. The Go proxy validator has the same two messages (`pkg/util/roaringfilter/portable_roaring64.go:62-65`, `:128-131`), and that is the path that actually reaches the client: `internal/parser/planparserv2/roaring_match.go:95-97` wraps it with `%v` into a `ParameterInvalid` message. Failure: a client submits an MRB1 blob whose high keys or Roaring32 keys are not strictly increasing (hand-assembled, or reordered in transit while retaining real data) → the `ParameterInvalid` message returned to the client and written to the proxy log contains a real high key (the upper 32 bits of a member value), and the same string reaches segcore errors and logs on QueryNode. The exposure is far smaller than a whole-blob leak, but it is the same MUST at design doc `:471`. Suggestion: drop the key values and report the container ordinal instead ("container *i* has a non-increasing key"); fix all four sites together. *(raised by xaxys, xiaocai2333)*

xiaofan-luan:
Thanks for the detailed review. All four findings are addressed on the current head (`bf29f4f0`). Details below so each one can be checked directly.

### Medium 1 — `LogPlanProtoDebug` leaking the roaring member set through the byte threshold

Agreed on the root cause: a byte threshold is the wrong gate for an encoding as compact and reversible as roaring.

The threshold is no longer what protects the blob. `PlanProtoDebugString` now copies the plan and runs `RedactMembershipFilterBlobs()` (`internal/core/src/query/PlanProto.cpp:987`), a recursive walk that replaces `BloomFilterExpr.filter_blob` and `RoaringFilterExpr.bitmap_blob` with `<N bytes elided>` **regardless of plan size**. The 2.2 KiB / 10M-contiguous-ids case in the review lands well under `kMaxPlanDebugBytes` and is still elided. `kMaxPlanDebugBytes` is kept only as a short-circuit so an oversized plan is not copied and stringified at all. The walk also clears unknown fields, so an old-QueryNode descriptor pool that preserves field 23 as unknown bytes does not print it either. Both call sites (search and retrieve/delete) go through this function.

Covered by `PlanProto.RedactsMembershipBlobsInDebugString` and the unknown-field case in `PlanProtoTest.cpp`.

### Medium 2 — `RedactPlanForLog` not covering roaring plans

Fixed. `membershipRedactedPlan.String()` (`internal/parser/planparserv2/bloom_match.go:480`) now collects **both** node types, and does so for the main predicate and for every `ScoreFunction.filter`:

```go
collectBloomFilterExprs(planPredicates(p.plan), &blooms)
collectRoaringFilterExprs(planPredicates(p.plan), &roarings)
for _, scorer := range p.plan.GetScorers() { /* both, again */ }
```

Both holes you identified are closed: a pure-roaring plan no longer takes the early `len(blooms) == 0` return, and a mixed plan elides both blob kinds. The swap is still pointer-assignment on the original slices with a `defer` restore, so there is no clone of a multi-MiB plan.

### Low 1 — index-only path doing `Reverse_Lookup` for the whole batch

Fixed rather than left as a TODO. As you noted, the real fix is in the shared helper, so that is where it went: `Expr.h` gained `ProcessIndexLookupByOffsetsWithMask` / `ProcessDataByOffsetsWithMask` (`internal/core/src/exec/expression/Expr.h:723`), which take a candidate mask indexed by candidate position. Masked-out candidates never reach `Reverse_Lookup`; the evaluator is still invoked with `data == nullptr` so its candidate-position cursor stays aligned, and `res[i]` / `valid_res[i]` keep their upstream values. `PhyRoaringFilterExpr` uses the mask-aware entry points (`RoaringFilterExpr.cpp:310`).

The bloom path is untouched here and can adopt the same helpers in a follow-up.

### Low 2 — validation errors carrying member key values

Fixed at all four sites. The messages now report the **container ordinal** instead of the key, exactly as suggested:

- `internal/core/src/common/RoaringMembership.cpp` (high container, Roaring32 container)
- `pkg/util/roaringfilter/portable_roaring64.go`
- `client/roaringfilter/portable_roaring64.go`

e.g. `invalid portable roaring64 body: high container %d has a non-increasing key`. No member-derived value reaches the client error, the proxy log, or segcore.

---

More generally on the diagnostics MUSTs in the design doc: expression text is logged unchanged because it only ever carries the `{ids}` placeholder, while the **template values** are elided wherever a request or plan is rendered — gRPC trace logs, the RESTful trace log's `exprParams`, proxy plan logs, and QueryNode plan debug. Errors report a template's name, never its value.

internal/parser/planparserv2/roaring_match.go
```diff
@@ -0,0 +1,259 @@
...
+		if e, ok := node.GetExpr().(*planpb.Expr_RoaringFilterExpr); ok {
+			*out = append(*out, e.RoaringFilterExpr)
+		}
+		return false
+	})
+}
+
+// PlanContainsRoaringFilter reports whether the plan's main predicate or a
+// scorer filter contains a roaring_match expression.
+func PlanContainsRoaringFilter(plan *planpb.PlanNode) bool {
```
liliu-z, line 241: nit: `PlanContainsRoaringFilter` is a line-for-line copy of `PlanContainsBloomFilter`, and both re-implement the switch that `planPredicates` already does — three copies of the same switch in this package. Worth collapsing into one helper that takes the `has*` predicate, in the same spirit as the `walkExpr` extraction in this PR. Same applies to `collectRoaringFilterExprs` / `collectBloomFilterExprs`.

internal/core/src/common/RoaringMembership.h
```diff
@@ -0,0 +1,79 @@
...
+        uint64_t low_container_count;
+        uint64_t estimated_decoded_bytes;
+    };
+
+    static constexpr std::string_view kMagic = "MRB1";
+    static constexpr uint16_t kVersion = 1;
+    static constexpr uint16_t kFormatPortableRoaring64 = 1;
+    static constexpr size_t kHeaderSize = 32;
+    static constexpr size_t kMaxBodySize = 128 * 1024 * 1024;
+    static constexpr uint64_t kMaxHighContainerCount = uint64_t{1} << 18;
```
liliu-z, line 45: These 9 limits are hand-copied in three places: here, `pkg/util/roaringfilter`, `client/roaringfilter`. They have to agree — the client pre-rejects what the proxy would reject, and the proxy pre-rejects what segcore would reject — but nothing checks that. Changing `kMaxHighContainerCount` here alone compiles and passes CI; you'd find out when the proxy admits a blob every querynode refuses.

The two configurable limits already have a single owner in paramtable. Same could work here — push at init like the other segcore limits in `initcore` — or at minimum a test that reads all three sets and asserts they match.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -477,6 +477,40 @@ func restfulSizeMiddleware(handler gin.HandlerFunc, observeOutbound bool) gin.Ha
...
+// redactExprParams returns a copy of params with every value replaced by
+// proxy.RedactedValue, leaving the placeholder names visible. A template value
+// is user payload -- a membership filter ships a multi-MiB blob through
+// exprParams -- and a trace log only needs to show which placeholders were
+// bound, not what they were bound to.
+//
+// The gRPC path redacts expr_template_values the same way, but its helper
+// switches on protobuf request types and so cannot recognize these RESTful
+// shapes; they are handled here instead.
+// hasExprParams reports whether any sub-request binds expression templates, so
```
liliu-z, line 489: Missing blank line — the `redactExprParams` doc comment above now attaches to `hasExprParams`, and `redactExprParams` ends up undocumented.

internal/core/src/exec/expression/RoaringFilterExpr.cpp
```diff
@@ -0,0 +1,325 @@
...
+        processed_cursor += size;
+    };
+
+    int64_t processed_size;
+    if (has_offset_input_) {
+        processed_size = ProcessDataByOffsets<T>(
+            execute_sub_batch, std::nullptr_t{}, input, res, valid_res);
+    } else {
+        processed_size = ProcessDataChunks<T>(
+            execute_sub_batch, std::nullptr_t{}, res, valid_res);
```
liliu-z, line 224: `skip_func` is nullptr, so no chunk is ever skipped. Bloom has to pass nullptr — an SBBF has no range — but a roaring bitmap does: `Minimum()`/`Maximum()` are O(1), so a chunk whose SkipIndex range doesn't overlap could be dropped without reading a row. `TermExpr`, `UnaryExpr`, `BinaryRangeExpr` and `BinaryArithOpEvalRangeExpr` all pass a real one, so roaring is the only expression here that could use SkipIndex and doesn't.

Fine as follow-up, but worth a TODO. Caveat for whoever picks it up: keys are two's-complement, so a chunk range spanning zero is two key ranges.

pkg/util/roaringfilter/roaring_filter.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+// Filter is an immutable exact-membership bitmap after construction.
+type Filter struct {
+	bitmap *roaring64.Bitmap
+}
+
+// Build deduplicates members into a portable Roaring64 bitmap and wraps it in
+// an MRB1 envelope.
+func Build(members []int64) ([]byte, error) {
```
liliu-z, line 82: `Build` here has no non-test caller — the proxy only reaches `Validate`, and the client builds through `client/roaringfilter`. Mirror image on the client side: `Validate`, `Parse`, `Filter` and all 284 lines of `portable_roaring64.go` are test-only there, since `NewRoaringBitmapBlob` stops at `Build`. So each package carries a full copy of the other's job, ~590 lines, which is also why the two `portable_roaring64.go` files ended up 2 lines apart.

Cutting the client down to `Build` and this one down to `Validate` leaves zero overlap and nothing to keep in sync. Tests that need to decode can call `roaring64.Bitmap.ReadFrom` directly. Unless some of these are meant as public API for SDK users — worth saying so if they are.

liliu-z (commented):
/lgtm
/approve

PR #51969, opened 2026-07-29
feat: target telemetry config at a workload label

xiaofan-luan:
issue: #51963
design doc: `docs/design-docs/design_docs/20260131-client_side_telemetry.md` (updated in this PR)

> **Draft — blocked on milvus-io/milvus-proto#646.**
> `go.mod` carries a temporary `replace` onto that PR's commit so this compiles and is fully reviewable. It must be dropped and re-pinned to a merged milvus-proto version before merge.
>
> Stacked on #51967, which is where client-scoped configs became one-time.

## Why

#51967 rejects `persistent=true` with a `client:` target, because a client ID names a *process*: the SDK mints a fresh one per start, so a config targeted at one silently stops applying the moment that client restarts.

That closed a leak but left a gap. The remaining durable scopes are `global` and `database:<db>` — there is no way to express *"turn the sampling rate down on the ingest workers, and keep it that way"* short of configuring an entire database or the whole cluster.

## What

A label names a **workload** rather than a process, so it keeps matching across restarts — which is exactly what makes it a legitimate target for persistent config.

**Clients declare labels** through the existing `ClientInfo.Reserved` map, so the client half needs no schema change:

```go
milvusclient.New(ctx, &milvusclient.ClientConfig{
    TelemetryConfig: &milvusclient.TelemetryConfig{
        Labels: map[string]string{"app": "ingest-worker"},
    },
})
```

→ `Reserved["label.app"] = "ingest-worker"`

**Operators target them:**

```bash
curl -u root:pw -XPOST "$M/_telemetry/commands" -d '{
  "command_type": "push_config",
  "target_label": "app=ingest-worker",
  "payload": {"sampling_rate": 0.1},
  "persistent": true
}'
```

Also available in the WebUI as a new **Workload (label)** target.

Keys that are empty or contain `=` are dropped client-side rather than sent — neither could ever be selected by a `label:key=value` scope, so emitting them would just produce labels that silently never match.

## Scope model after this PR

| Scope | Selects | Lifetime | Persistent allowed |
|---|---|---|---|
| `global` | every client | durable | yes |
| `database:<db>` | clients that accessed the db | durable | yes |
| `label:<k>=<v>` | clients declaring the label | durable | **yes** ← new |
| `client:<id>` | one process | ephemeral | no |

Lifetime is what decides persistence eligibility, and that is now a coherent rule rather than a special case.

## One deliberate non-change

An earlier draft made the three target fields mutually exclusive, which felt like the better API — and it broke `TestCommandWithTargetClientTakesPrecedenceOverDatabase`, an existing test pinning the behaviour that client wins when both are set.

Reverted to precedence (client → label → database). Adding a scope should not change how the existing ones resolve; tightening that is a separate decision, not a side effect of this feature. The proto comment documents the precedence.

## Tests

- `TestResolveTargetScope` — all four scopes, label trimming, malformed selectors (`app`, `=ingest`), and both precedence pairs
- `TestMatchesLabel` — hit/miss, empty-value semantics (`tier=` matches a declared empty label but `app=` does not match a set one), whitespace, and that `client_id` is not addressable as a label
- `TestLabelScopedConfigMayBePersistent` / `TestLabelScopedConfigSurvivesReload` — the actual point: a label-scoped config is accepted as persistent and is still there, still addressable, after a simulated RootCoord restart
- `TestClientInfoLabels` — SDK publishes under the prefix, coexists with `client_id`, drops unaddressable keys

## Verification

`go test` green for the `client/` module and `internal/rootcoord/telemetry`; `golangci-lint` 0 issues on both; `gofmt` clean.

Unit-level only — **not run against a live cluster**. In particular the end-to-end path (client declares a label → server matches it → config survives a real RootCoord restart) is covered by a simulated reload, not an actual restart.

### DIFF
--- a/client/milvusclient/interceptors.go
+++ b/client/milvusclient/interceptors.go
@@ -18,6 +18,8 @@ package milvusclient
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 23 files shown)


xiaofan-luan:
Closing: the label scope is not worth a permanent proto field for what it buys today.

`push_config` exposes exactly three knobs (`enabled`, `heartbeat_interval_ms`, `sampling_rate`), all of them about telemetry itself. Per-workload targeting of those three does not justify permanent API surface plus a client-side opt-in that applications have to adopt before anything matches.

The gap it filled is real but narrow, and `database:` scope covers most of it. If `push_config` ever grows knobs that matter more, this can come back.

#51967 is unaffected: the TTL default and the client-scope persistence rule stand on their own.

PR #51973, opened 2026-07-29
test: isolate external storage cleanup

### LINKED ISSUE #51971: [Bug]: Unit tests can delete shared etcd and object storage data
## Environment

- **Milvus Version**: master at `ab4935563cb6a8e4f8a917608ea2d0a4b2333aca`
- **Deployment Mode**: N/A (Go unit tests connecting to configured external services)
- **MQ**: N/A
- **SDK**: N/A
- **OS**: macOS 26.5.2 arm64 (the issue is platform-independent)

### Option B: Steps

1. Configure `configs/milvus.yaml` so etcd and MinIO/S3 point to shared, non-disposable services.
2. Put sentinel data outside the unit tests' intended scope, for example:
   - etcd: `by-dev/meta/production-sentinel`
   - object storage: `files/production-sentinel`
3. Run the affected tests:

```bash
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/util/sessionutil -run '^TestUpdateSessions$'
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/segments -run '^TestSearch$'
```

4. Inspect the shared services after teardown:
   - `TestUpdateSessions` constructs an etcd KV with an empty root and defers `RemoveWithPrefix("")`, which deletes every key visible to the test credentials.
   - `SearchSuite` calls `RemoveWithPrefix(minio.rootPath)`; with the default root `files`, this deletes every object under `files/`, not only objects written by the test.

### Trigger Conditions

- **Frequency**: Always when the affected teardown runs and the configured credentials permit deletion.
- **First observed after**: Running the affected unit tests against shared external storage.
- **Does NOT happen when**: The tests use disposable local etcd/MinIO instances, where the destructive cleanup is not externally visible.

## Expected Behavior

Each unit test must write into a test-specific storage namespace and remove only that namespace during teardown. Existing data outside the test namespace must remain unchanged.

## Actual Behavior

The tests read the real storage configuration through `paramtable.Init()` but perform teardown with an empty etcd prefix or the global object-storage root. This can remove unrelated Milvus metadata and objects from a shared environment.

## Error Logs

No error is emitted. The destructive cleanup succeeds silently.

## Non-default Configuration

```yaml
etcd:
  endpoints: shared-etcd.example:2379
  rootPath: by-dev
  metaSubPath: meta

minio:
  address: shared-s3.example
  bucketName: shared-milvus
  rootPath: files

common:
  storageType: remote
```

## Analysis Hints

- `internal/util/sessionutil/session_util_test.go::TestUpdateSessions`
- `internal/querynodev2/segments/search_test.go::SearchSuite.TearDownTest`
- Scope the etcd KV to the test's existing random `metaRoot`.
- Construct the SearchSuite chunk manager with a test-specific root and remove only that root.

sijie-ni-0214:
issue: #51971

Fixes #51971

## Background

Several Go unit tests initialize `paramtable` from `configs/milvus.yaml`, so they may connect to the etcd and object storage configured for a real or shared environment.

Two teardown paths were not scoped to test-owned data:

- `TestUpdateSessions` created an etcd KV with an empty root and deferred `RemoveWithPrefix("")`. With an unrestricted client, this removes every key visible to the test credentials.
- `SearchSuite` removed `minio.rootPath` directly. With the default value `files`, this removes every object under `files/`, including data unrelated to the test.

This stayed mostly hidden when tests used disposable local etcd and MinIO instances, but it is destructive when a developer's configuration points to shared services.

## Changes

- Bind the `TestUpdateSessions` etcd KV to the test's existing random `metaRoot`, so empty-prefix cleanup is limited to that namespace.
- Give each `SearchSuite` test its own chunk-manager root and remove only that root during teardown.
- Check chunk-manager initialization errors instead of ignoring them.
- Keep the change limited to test setup and teardown; no production storage interfaces are changed.

## Verification

- [x] `TestUpdateSessions` with `-tags dynamic,test -gcflags="all=-N -l" -count=1`
- [x] Full `TestSearch` suite with `-tags dynamic,test -gcflags="all=-N -l" -count=1`
- [x] An etcd sentinel outside the test root remained after `TestUpdateSessions`
- [x] A MinIO sentinel under the configured global `files/` root remained after `TestSearch`
- [x] `make static-check` — 0 issues

### DIFF
--- a/internal/querynodev2/segments/search_test.go
+++ b/internal/querynodev2/segments/search_test.go
@@ -37,6 +37,7 @@ import (
 
 type SearchSuite struct {
 	suite.Suite
+	rootPath     string
 	chunkManager storage.ChunkManager
 
 	manager      *Manager
@@ -58,8 +59,11 @@ func (suite *SearchSuite) SetupTest() {
 	ctx := context.Background()
 	msgLength := 100
 
-	chunkManagerFactory := storage.NewChunkManagerFactoryWithParam(paramtable.Get())
-	suite.chunkManager, _ = chunkManagerFactory.NewPersistentStorageChunkManager(ctx)
+	suite.rootPath = suite.T().Name()
+	chunkManagerFactory := storage.NewTestChunkManagerFactory(paramtable.Get(), suite.rootPath)
+	chunkManager, err := chunkManagerFactory.NewPersistentStorageChunkManager(ctx)
+	suite.Require().NoError(err)
+	suite.chunkManager = chunkManager
 	initcore.InitRemoteChunkManager(paramtable.Get())
 	initcore.InitLocalChunkManager(suite.T().Name())
 	initcore.InitMmapManager(paramtable.Get(), 1)
@@ -139,10 +143,19 @@ func (suite *SearchSuite) SetupTest() {
 }
 
 func (suite *SearchSuite) TearDownTest() {
-	suite.sealed.Release(context.Background())
-	DeleteCollection(suite.collection)
 	ctx := context.Background()
-	suite.chunkManager.RemoveWithPrefix(ctx, paramtable.Get().MinioCfg.RootPath.GetValue())
+	if suite.sealed != nil {
+		suite.sealed.Release(ctx)
+		suite.sealed = nil
+	}
+	if suite.collection != nil {
+		DeleteCollection(suite.collection)
+		suite.collection = nil
+	}
+	if suite.chunkManager != nil {
+		suite.chunkManager.RemoveWithPrefix(ctx, suite.rootPath)
+		suite.chunkManager = nil
+	}
 }
 
 func (suite *SearchSuite) TestSearchSealed() {
--- a/internal/util/sessionutil/session_util_test.go
+++ b/internal/util/sessionutil/session_util_test.go
@@ -122,7 +122,7 @@ func TestUpdateSessions(t *testing.T) {
 
 	metaRoot := fmt.Sprintf("%d/%s", rand.Int(), DefaultServiceRoot)
 	etcdCli, _ := kvfactory.GetEtcdAndPath()
-	etcdKV := etcdkv.NewEtcdKV(etcdCli, "")
+	etcdKV := etcdkv.NewEtcdKV(etcdCli, metaRoot)
 
 	defer etcdKV.RemoveWithPrefix(ctx, "")
 
@@ -161,7 +161,7 @@ func TestUpdateSessions(t *testing.T) {
 	notExistSessions, _, _ := s.GetSessions(ctx, "testt")
 	assert.Equal(t, len(notExistSessions), 0)
 
-	etcdKV.RemoveWithPrefix(ctx, metaRoot)
+	etcdKV.RemoveWithPrefix(ctx, "")
 	assert.Eventually(t, func() bool {
 		sessions, _, _ := s.GetSessions(ctx, "test")
 		return len(sessions) == 0


czs007:
Test-only change; the etcd empty-prefix cleanup fix and the object-storage teardown re-scoping are both correct, and the remaining findings are low-severity nits inherited from existing suite conventions rather than regressions introduced here.

## Low

**`internal/querynodev2/segments/search_test.go:145` — teardown can nil-panic and mask the real failure**
The newly added `suite.Require().NoError(err)` at line 65 introduces an earlier `FailNow` point in `SetupTest`. testify runs `TearDownTest` from a `defer`, so it still executes after `runtime.Goexit`; at that point `suite.sealed` is unassigned for the first subtest, and `TearDownTest`'s first statement `suite.sealed.Release(ctx)` panics on the nil `Segment`, hiding the actual chunk-manager creation error. (Note: `suite.chunkManager` is assigned on the preceding line, so it is not the nil dereference — `suite.sealed` is.) Suggestion: guard the teardown, e.g. `if suite.sealed != nil { suite.sealed.Release(ctx) }` (and similarly for `suite.chunkManager`). The same pattern exists in `retrieve_test.go:185` and `segment_test.go:154`, so a shared follow-up is also reasonable. *(raised by xaxys, xiaocai2333)*

**`internal/querynodev2/segments/search_test.go:62` — deterministic root path plus fixed IDs leaves a cross-process collision window**
`suite.rootPath = suite.T().Name()` is deterministic and the suite uses fixed `collectionID=100` / `partitionID=10` / `segmentID=1`, so two concurrent processes running this test against a shared bucket can overwrite or delete each other's objects. This is still a strict improvement over the previous teardown, which removed the global `MinioCfg.RootPath` (`files/`), and it matches the existing convention across the sibling suites, so it is not a blocker. Suggestion: append a random run suffix to `rootPath` in a follow-up, ideally suite-family-wide. *(raised by xaxys, xiaocai2333)*

**`internal/querynodev2/segments/search_test.go:148` — `RemoveWithPrefix` return value ignored**
Cleanup failures pass silently and can leave residual objects behind. This matches both the pre-change line and the sibling suites, so it is pre-existing style rather than something this PR introduces. Suggestion: assert or log the error, e.g. `suite.NoError(suite.chunkManager.RemoveWithPrefix(ctx, suite.rootPath))`. *(raised by xaxys, xiaocai2333)*

**`internal/storage/utils.go:1751` — shared test factory hardcodes `"minio"` and drops several storage options**
`NewTestChunkManagerFactory` pins `persistentStorage` to `"minio"` and, compared with the production `NewChunkManagerFactoryWithParam`, omits `SslCACert`, `SslTLSMinVersion`, `UseVirtualHost`, `Region`, `RequestTimeoutMs`, and `GcpCredentialJSON`, so tests built on it cannot exercise `storageType=local` or those configuration paths. This PR does not introduce or modify the helper — it only aligns `search_test.go` with the convention already used by 11 call sites under `querynodev2` (plus a near-duplicate at `internal/mocks/util/mock_segcore/mock_data.go:457`). Suggestion: track a separate repo-level change to have the helper honor `CommonCfg.StorageType` and forward the missing options; no change required in this PR. *(raised by xaxys, xiaocai2333)*

sijie-ni-0214:
Addressed the teardown masking issue in commit 61a40c2dd3.

The chunk manager is now assigned to the suite only after creation succeeds. This also avoids retaining a typed-nil storage.ChunkManager interface when the factory returns an error. TearDownTest now guards and clears the sealed segment, collection, and chunk manager before cleanup.

Verification:
- With MinIO unavailable, TestSearchGrowing reports the original storage initialization error without a teardown panic.
- With local MinIO running, the full TestSearch suite passes.
- make static-check passes.

The remaining low-severity findings follow existing sibling-suite or shared-helper conventions and are left out of this focused PR.

czs007 (approved):
/lgtm

Issue #51948, opened 2026-07-28
Proxy retries QueryNode queue-full errors


MrPresent-Han:
## Problem

When a QueryNode returns `ErrServiceTooManyRequests` because its DQL queue is full, the proxy shard load balancer treats the response as a generic execution failure. It excludes the node and immediately retries the same workload across replicas. With hybrid search and many shards, these proxy-side retries amplify an existing overload.

## Expected behavior

Queue saturation is a transient overload signal from an otherwise healthy QueryNode. The proxy should stop its request-level retry without excluding the node and return the original code 4 status with `Retriable=true`, allowing the client to apply backoff.

## Affected branches

The retry path is present in `2.6`, `3.0`, and `master`.

PR #51974, opened 2026-07-29
fix: bound QueryNode queue-full retries

MrPresent-Han:
issue: #51948

3.0 PR: #51975
2.6 PR: #51949

## What changed

- Treat `ErrServiceTooManyRequests` as request-scoped QueryNode saturation.
- Exclude the saturated QueryNode for the current request and preserve one failover pass across the remaining replicas in the exact cached leader snapshot used for selection.
- Once all selectable replicas in that snapshot have been attempted, stop immediately before another retry backoff or any shard-leader refresh, and return the original code 4 status with `Retriable=true`.
- Keep shard-leader refresh and exclusion-set recycling only for non-overload failures.
- Do not add a healthy but saturated QueryNode to the cross-request blacklist.
- Cover single-replica saturation, successful multi-replica failover, all-replicas-saturated behavior, cache snapshot shrink/expansion, and the absence of cached or uncached shard lookup during the overload sweep.

## Root cause

The load-balancing policy treated queue-full as a generic retriable execution failure. Once every replica was request-level excluded, it cleared the exclusion set and cycled through the same saturated replicas again according to `retryOnReplica`. Hybrid-search fanout amplified those repeated dispatches.

An intermediate version stopped that QueryNode retry cycle only after `GetShard(false)`. Because uncached shard lookup calls `MixCoord.GetShardLeaders`, the single-replica overload path added one coordinator RPC per failed channel workload. The overload check now runs before that refresh.

## Behavior

- Single replica: execute on the QueryNode once, then return the overload status without another retry backoff or shard-leader refresh.
- Multiple replicas: a queue-full response can fail over to each cached, untried replica in the captured selection snapshot within the same request.
- All selectable replicas saturated or excluded: each saturated replica is attempted at most once; no cached/uncached shard lookup or exclusion-set clearing occurs after overload begins.
- The original retriable status reaches the client; automatic backoff still depends on the SDK or application retry policy.

## Scope

This PR bounds request-level replica retries when `ErrServiceTooManyRequests` reaches Proxy as code 4. It does not change Delegator partial-result mixed-error aggregation. In the opt-in partial-result path, aggregation can mask code 4 when queue-full is combined with other worker errors; that pre-existing case remains a separate follow-up and is not covered by this PR's retry-bounding claim.

## Validation

```text
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy/shardclient -run 'TestLBPolicySuite/TestExecuteWithRetryTooManyRequests'
ok github.com/milvus-io/milvus/internal/proxy/shardclient 6.260s

go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy/shardclient
ok github.com/milvus-io/milvus/internal/proxy/shardclient 153.604s
```

### DIFF
--- a/internal/proxy/shardclient/lb_policy.go
+++ b/internal/proxy/shardclient/lb_policy.go
@@ -151,110 +151,127 @@ func preferredNodeID(workload CollectionWorkLoad, channel string) int64 {
 	return nodeID
 }
 
+func allShardLeadersExcluded(shardLeaders []NodeInfo, excludeNodes typeutil.UniqueSet) bool {
+	if len(shardLeaders) == 0 {
+		return false
+	}
+	for _, node := range shardLeaders {
+		if !excludeNodes.Contain(node.NodeID) {
+			return false
+		}
+	}
+	return true
+}
+
 // try to select the best node from the available nodes
 func (lb *LBPolicyImpl) selectNode(ctx context.Context, balancer LBBalancer, workload ChannelWorkload, excludeNodes *typeutil.UniqueSet) (NodeInfo, bool, error) {
+	targetNode, selectedByBalancer, _, err := lb.selectNodeWithUncachedRetry(ctx, balancer, workload, excludeNodes)
+	return targetNode, selectedByBalancer, err
+}
+
+func (lb *LBPolicyImpl) selectNodeFromShardLeaders(ctx context.Context, balancer LBBalancer, workload ChannelWorkload, shardLeaders []NodeInfo, excludeNodes *typeutil.UniqueSet) (targetNode NodeInfo, selectedByBalancer bool, err error) {
 	log := mlog.With(
 		mlog.Int64("collectionID", workload.CollectionID),
 		mlog.String("channelName", workload.Channel),
 	)
-	// Select node using specified nodes
-	trySelectNode := func(withCache bool) (NodeInfo, bool, error) {
+
+	candidateNodes := make(map[int64]NodeInfo)
+	serviceableNodes := make(map[int64]NodeInfo)
+	defer func() {
+		if err != nil {
+			candidatesInStr := lo.Map(shardLeaders, func(node NodeInfo, _ int) string {
+				return node.String()
+			})
+			serviceableNodesInStr := lo.Map(lo.Values(serviceableNodes), func(node NodeInfo, _ int) string {
+				return node.String()
+			})
+			log.Warn(ctx, "failed to select shard",
+				mlog.Int64s("excluded", excludeNodes.Collect()),
+				mlog.String("candidates", strings.Join(candidatesInStr, ", ")),
+				mlog.String("serviceableNodes", strings.Join(serviceableNodesInStr, ", ")),
+				mlog.Err(err))
+		}
+	}()
+
+	// Filter nodes based on excludeNodes
+	for _, node := range shardLeaders {
+		if !excludeNodes.Contain(node.NodeID) {
+			if node.Serviceable {
+				serviceableNodes[node.NodeID] = node
+			}
+			candidateNodes[node.NodeID] = node
+		}
+	}
+	if len(candidateNodes) == 0 {
+		err = merr.WrapErrChannelNotAvailable(workload.Channel, "no available shard leaders")
+		return NodeInfo{}, false, err
+	}
+
+	if preferredNode, ok := serviceableNodes[workload.PreferredNodeID]; ok {
+		recordPreferredNodeSelection(metrics.PreferredNodeHitLabel)
+		return preferredNode, false, nil
+	} else if workload.PreferredNodeID != 0 {
+		recordPreferredNodeSelection(metrics.PreferredNodeUnavailableLabel)
+	}
+
+	balancer.RegisterNodeInfo(lo.Values(candidateNodes))
+
+	// prefer serviceable nodes
+	var targetNodeID int64
+	if len(serviceableNodes) > 0 {
+		targetNodeID, err = balancer.SelectNode(ctx, lo.Keys(serviceableNodes), workload.Nq)
+	} else {
+		targetNodeID, err = balancer.SelectNode(ctx, lo.Keys(candidateNodes), workload.Nq)
+	}
+	if err != nil {
+		return NodeInfo{}, false, err
+	}
+
+	if _, ok := candidateNodes[targetNodeID]; !ok {
+		err = merr.WrapErrNodeNotAvailable(targetNodeID)
+		return NodeInfo{}, false, err
+	}
+
+	return candidateNodes[targetNodeID], true, nil
+}
+
+func (lb *LBPolicyImpl) selectNodeWithUncachedRetry(ctx context.Context, balancer LBBalancer, workload ChannelWorkload, excludeNodes *typeutil.UniqueSet) (NodeInfo, bool, []NodeInfo, error) {
+	log := mlog.With(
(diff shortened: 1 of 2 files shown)


internal/proxy/shardclient/lb_policy.go
```diff
@@ -264,13 +264,17 @@ func (lb *LBPolicyImpl) ExecuteWithRetry(ctx context.Context, workload ChannelWo
...
 	var lastErr error
+	var overloadErr error
 	var err error
 	var shardLeaders []NodeInfo
 	requestExcludedNodes := typeutil.NewUniqueSet()
 	tryExecute := func() (bool, error) {
 		// Get fresh blacklist on each retry to include newly blacklisted nodes
 		blacklist := lb.blacklist.GetBlacklistedNodes(workload.Channel)
 		if len(shardLeaders) > 0 && requestExcludedNodes.Len() >= len(shardLeaders) {
+			if overloadErr != nil {
```
weiliu1031: **[P1] Determine exhaustion from the actual selectable nodes**

This check only compares `requestExcludedNodes` with the initial `shardLeaders`, while the candidate set passed to `selectNode` excludes `blacklist ∪ requestExcludedNodes`.

For example, with leaders `[1, 2]`, if node 1 is already blacklisted and node 2 returns queue-full, `requestExcludedNodes.Len()` is only 1, so this guard is skipped. `selectNode` then finds no cached candidate, calls `GetShard(false)`, and may clear the combined exclusion set before selecting node 1 or node 2 again.

This reintroduces both the coordinator refresh and repeated dispatch that this change promises to prevent. Please determine exhaustion from the actual candidate set and ensure the overload path cannot refresh shard leaders or clear exclusions.

internal/proxy/shardclient/lb_policy.go
```diff
@@ -335,6 +346,14 @@ func (lb *LBPolicyImpl) ExecuteWithRetry(ctx context.Context, workload ChannelWo
 			if merr.GetErrorType(err) == merr.InputError {
 				return false, err
 			}
+			// Queue saturation is a transient overload signal from a healthy
+			// QueryNode. Try each replica at most once, then return the overload
+			// to the caller instead of cycling through saturated replicas.
+			if errors.Is(err, merr.ErrServiceTooManyRequests) {
```
liliu-z, line 373: In opt-in partial-result mode (proxy.partialResultRequiredDataRatio < 1; default is 1, which disables it and makes executeSubTasks fail-fast with a single error) a remote worker's queue-full raised inside the delegator's fan-out is aggregated via merr.Combine when partial results are not satisfied. multiErrors projects its wire Code from the tail error (Unwrap returns the last error, and Code returns the first milvusError found), so when code-4 is not the tail the QueryNode status carries a different code and the proxy's errors.Is(err, ErrServiceTooManyRequests) branch does not fire, falling back to the generic retry path. This gap is narrow: it does not affect top-level shard-leader admission overload (clean code-4), all-queue-full aggregation (tail is code-4), or the default config, and it is not a regression (matches pre-PR behavior there). Per G3, either preserve code-4 through the aggregation or scope the PR's bounding claim to exclude this partial-result sub-case.

internal/proxy/shardclient/lb_policy.go
MrPresent-Han: changed

internal/proxy/shardclient/lb_policy.go
MrPresent-Han: Agreed that this partial-result aggregation gap exists. It is pre-existing and belongs to Delegator/mixed-error projection rather than the Proxy LB retry state machine, so it remains outside this PR.

I narrowed both the amended commit message and the PR description accordingly: this PR bounds request-level replica retries only when `ErrServiceTooManyRequests` reaches Proxy as code 4. The opt-in partial-result case where `merr.Combine` masks code 4 remains unchanged and is explicitly excluded from this PR's bounding claim as a separate follow-up.

The default `partialResultRequiredDataRatio=1` path, top-level admission queue-full, and clean/all-queue-full code-4 paths remain covered by this change.

internal/proxy/shardclient/lb_policy.go
MrPresent-Han: Addressed in `d6d66dcaf5`.

The overload path now builds `excludeNodes` from `blacklist ∪ requestExcludedNodes`, captures the exact `selectedShardLeaders` snapshot that selected the first saturated node, and uses `selectNodeFromShardLeaders` for the rest of that request-level sweep. Selection and exhaustion therefore use the same immutable node set, and the overload path cannot call `GetShard(true/false)` or clear the combined exclusions.

Exhaustion is checked both before selection and immediately after inserting the current queue-full node, so a blacklisted replica plus a newly saturated replica terminates without a cache refresh or repeated dispatch.

Regression coverage includes the existing-blacklist case, cache shrink, cache expansion between the initial lookup and selection, cross-replica failover, and each replica being tried at most once. The focused queue-full suite and the full `internal/proxy/shardclient` package both pass.

internal/proxy/shardclient/lb_policy.go
```diff
@@ -335,6 +367,21 @@ func (lb *LBPolicyImpl) ExecuteWithRetry(ctx context.Context, workload ChannelWo
...
+				if overloadErr == nil {
+					overloadShardLeaders = append([]NodeInfo(nil), selectedShardLeaders...)
```
buqian-zilliz, line 375: **[P2] Derive the overload sweep budget from the captured snapshot**

The overload snapshot is captured from `selectedShardLeaders`, but `retry.Handle`'s total attempt count was fixed earlier from a different `shardLeaders` read. `selectNodeWithUncachedRetry` re-reads the cache and can fall back to an uncached refresh, and earlier generic failures can consume attempts before this snapshot is captured. If the snapshot is larger than the remaining budget, the loop exits after its last backoff before every untried replica in `overloadShardLeaders` is selected, returning overload even though a healthy replica may remain. Derive or extend the overload budget from the captured snapshot and current exclusions, or run this sweep outside the original fixed counter.

internal/proxy/shardclient/lb_policy.go
```diff
@@ -264,44 +281,59 @@ func (lb *LBPolicyImpl) ExecuteWithRetry(ctx context.Context, workload ChannelWo
...
-					return true, lastErr
-				}
-				return true, err
+		excludeNodes := typeutil.NewUniqueSet(blacklist...)
+		excludeNodes.Insert(requestExcludedNodes.Collect()...)
+		balancer := lb.getBalancer()
+		var targetNode NodeInfo
+		var selectedByBalancer bool
+		var selectedShardLeaders []NodeInfo
+		if overloadErr != nil {
```
buqian-zilliz, line 300: **[P2] Account every failed target during the overload sweep**

Once `overloadErr` is set, later `GetClient` failures and non-retryable execution errors only call `ChannelBlacklist.Add` instead of request-excluding the selected node. `Add` is a no-op when `proxy.replicaBlacklistDuration <= 0`, so in that supported configuration the frozen-snapshot path can select the same failed replica repeatedly and lose the one-pass bound. Even when the blacklist is enabled, these branches—and the generic retryable-error branch—return `shouldRetry=true` without checking whether the current failure exhausted the snapshot, so `retry.Handle` sleeps before invoking the callback only to return `overloadErr`. During overload mode, record every unsuccessful target in a request-local attempted set and check exhaustion before returning any failure.

Issue #51782, opened 2026-07-23
[Enhancement]: segcore HashTable::groupProbe issues prefetches with ~zero distance — probe pipelining lost in Velox port


czs007:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## What would you like to be added?

Restore the 4-way interleaved probe pipeline in `HashTable::groupProbe()` (`internal/core/src/exec/HashTable.cpp`) so that the existing `__builtin_prefetch` calls in `ProbeState` actually get enough distance to hide memory latency.

### Current behavior

`ProbeState` contains two software prefetches, clearly intended to hide hash-table cache misses:

- `preProbe()` prefetches the bucket right after computing its offset ([HashTable.cpp:76-77](https://github.com/milvus-io/milvus/blob/master/internal/core/src/exec/HashTable.cpp#L76-L77)):

  ```cpp
  __builtin_prefetch(reinterpret_cast<uint8_t*>(table.table_) + bucketOffset_);
  ```

- `loadNextHit()` prefetches the group row ([HashTable.cpp:135](https://github.com/milvus-io/milvus/blob/master/internal/core/src/exec/HashTable.cpp#L135)):

  ```cpp
  group_ = table.row(bucketOffset_, hit);
  __builtin_prefetch(group_);
  ```

However, the driver loop processes rows one at a time with a **single** `ProbeState` ([HashTable.cpp:264-272](https://github.com/milvus-io/milvus/blob/master/internal/core/src/exec/HashTable.cpp#L264-L272)):

```cpp
ProbeState state;
for (int32_t idx = 0; idx < lookup.hashes_.size(); idx++) {
    ...
    state.preProbe(*this, lookup.hashes_[idx], idx);  // prefetch(table_ + bucketOffset_)
    state.firstProbe<ProbeState::Operation::kInsert>(*this); // loadTags(table_, bucketOffset_)
    fullProbe(lookup, state);
}
```

[...]

PR #51976, opened 2026-07-29
enhance: pipeline group hash probes

MrPresent-Han:
## What changed

- Restore a four-way interleaved `preProbe` / `firstProbe` / `fullProbe` pipeline in Segcore's grouping hash table.
- Refresh bucket tags for later lanes so inserts from earlier lanes in the same batch are visible, while avoiding a duplicate comparison of the prefetched candidate.
- Reserve capacity for the worst-case four insertions before starting a batch and retain the scalar tail.
- Add coverage for duplicate keys, same-bucket collisions, same-tag stale candidates, scalar tails, and the rehash boundary.

## Why

The scalar driver consumed each bucket and group-row prefetch almost immediately, leaving insufficient distance to hide random-memory latency. Interleaving four independent probe states restores the software pipeline used by Velox while preserving Milvus's incremental rehash behavior.

## Validation

- `clang-format --dry-run --Werror internal/core/src/exec/HashTable.cpp internal/core/src/exec/HashTable.h internal/core/unittest/test_query_group_by.cpp`
- `git diff --check`
- A local Debug C++ unit-test build was attempted, but the fresh C++20 Conan graph required source-building 48 dependencies and was stopped during OpenBLAS. C++ unit tests were not completed locally; CI is pending.

issue: #51782

### DIFF
--- a/internal/core/src/exec/HashTable.cpp
+++ b/internal/core/src/exec/HashTable.cpp
@@ -91,11 +91,17 @@ class ProbeState {
 
     template <Operation op, typename Compare, typename Insert, typename Table>
     inline char*
-    fullProbe(Table& table, Compare compare, Insert insert) {
+    fullProbe(Table& table, Compare compare, Insert insert, bool extraCheck) {
         AssertInfo(op == Operation::kInsert,
                    "Only support insert operation for group cases");
         if (group_ && compare(group_, row_)) {
             return group_;
+        }
+        auto* alreadyChecked = group_;
+        if (extraCheck) {
+            tagsInTable_ = table.loadTags(bucketOffset_);
+            hits_ = static_cast<BaseHashTable::MaskType>(
+                milvus::toBitMask(tagsInTable_ == wantedTags_));
         }
         const auto kEmptyGroup = BaseHashTable::TagVector::broadcast(0);
         for (int64_t numProbedBuckets = 0;
@@ -103,7 +109,8 @@ class ProbeState {
              ++numProbedBuckets) {
             while (hits_ > 0) {
                 loadNextHit<op>(table);
-                if (compare(group_, row_)) {
+                if (!(extraCheck && group_ == alreadyChecked) &&
+                    compare(group_, row_)) {
                     return group_;
                 }
             }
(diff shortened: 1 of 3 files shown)


internal/core/src/exec/HashTable.cpp
```diff
@@ -254,21 +261,53 @@ HashTable::fullProbe(HashLookup& lookup, ProbeState& state) {
...
+    ProbeState state1;
+    ProbeState state2;
+    ProbeState state3;
+    ProbeState state4;
+    int32_t probeIndex = 0;
+    const auto numProbes = static_cast<int32_t>(lookup.hashes_.size());
+
+    for (; probeIndex + kProbeBatchSize <= numProbes;
+         probeIndex += kProbeBatchSize) {
+        while (static_cast<uint64_t>(numDistinct_) + kProbeBatchSize >
```
buqian-zilliz: **[P3] Avoid speculative rehashes at batch boundaries**

This reserves capacity for four possible inserts before any lane is probed. When `numDistinct_` is one to three entries below `rehashSize()` and the next four rows all match existing groups, the scalar loop would keep the current table, but this path performs an O(`numDistinct_`) rehash and permanently doubles `table_` even though cardinality does not grow. With a large group limit, a duplicate-heavy batch that lands on this boundary can therefore cause an avoidable latency and memory spike. Please process the near-threshold rows scalarly, or otherwise defer resizing until an insertion actually reaches the limit, and enter the four-lane pipeline only when four inserts fit.

MrPresent-Han: Fixed in `8d1ced5725`. Near-threshold rows are now probed one at a time and capacity is re-evaluated after each row; once rehashing creates room, execution re-enters the four-lane pipeline. Added `TestExistingBatchDoesNotRehashNearThreshold`, which fills to `rehashSize() - 1`, probes four existing keys, and verifies that `rehashSize()` remains unchanged.

zhengbuqian (approved):

liliu-z (commented):
/lgtm
/approve

PR #51977, opened 2026-07-29
enhance: backport pipelined group hash probes to 3.0

MrPresent-Han:
Backport of #51976 to the 3.0 branch.

## What changed

- Restore the four-way interleaved grouping hash-probe pipeline.
- Refresh later-lane bucket tags after earlier lanes may insert.
- Reserve capacity before each four-row batch and retain the scalar tail.
- Backport the collision, duplicate-key, tail, and rehash-boundary tests.

## Validation

- Cherry-picked cleanly from master commit `c8049dfb72`.
- `clang-format --dry-run --Werror internal/core/src/exec/HashTable.cpp internal/core/src/exec/HashTable.h internal/core/unittest/test_query_group_by.cpp`
- `git diff --check trunk/3.0..HEAD`
- A local Debug C++ unit-test build was attempted on the master branch, but the fresh C++20 Conan graph required source-building 48 dependencies and was stopped during OpenBLAS. C++ unit tests were not completed locally; CI is pending.

pr: #51976
issue: #51782

### DIFF
--- a/internal/core/src/exec/HashTable.cpp
+++ b/internal/core/src/exec/HashTable.cpp
@@ -91,11 +91,17 @@ class ProbeState {
 
     template <Operation op, typename Compare, typename Insert, typename Table>
     inline char*
-    fullProbe(Table& table, Compare compare, Insert insert) {
+    fullProbe(Table& table, Compare compare, Insert insert, bool extraCheck) {
         AssertInfo(op == Operation::kInsert,
                    "Only support insert operation for group cases");
         if (group_ && compare(group_, row_)) {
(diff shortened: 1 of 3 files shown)


MrPresent-Han:
Backport updated in `3ed8354f08` with the speculative-rehash fix from master (`8d1ced5725`). Near-threshold rows now use scalar probing until a four-row batch safely fits, with regression coverage ensuring an existing-key batch does not grow the table.

liliu-z (commented):
/lgtm
/approve

Issue #51781, opened 2026-07-23
[Bug]: Growing segment in-flight queries can observe post-Reopen erased derived resources (per-call capture instead of per-operation)


xiaofan-luan:
## Current behavior

#51694 makes the sealed segment's `PublishedSegmentState` the strong owner of segment-derived resources (GEOMETRY caches, TEXT LOB readers), so an in-flight query that captured a published snapshot keeps a consistent generation of schema + resources.

The growing segment has no equivalent single publication point. `SegmentGrowingImpl` keeps `schema_`, `text_field_resources_`, and `geometry_caches_` as three separately-locked members, and read paths capture the text resource **per `bulk_subscript` call** (`CaptureTextFieldResource`) rather than per operation.

`Reopen` erases `text_field_resources_` / `geometry_caches_` entries for removed fields under their own locks. A query admitted against the old schema that is still executing can therefore observe the post-erase state on its next accessor call and fail on the `AssertInfo` (e.g. "TEXT field {} has no LOB reader" / missing resource), even though the rows it reads are still the old generation's.

## Expected behavior

A growing-segment query should capture schema + derived resources once per operation (analogous to the sealed `CapturePublishedState`), so a field drop racing an in-flight query cannot make that query fail mid-flight.

Note this is fail-loud, not a data-correctness issue: resources are `shared_ptr`-held, so already-captured references stay alive (no use-after-free, no stale reads). The gap is availability — a spurious query error during the drop-field window.

## Possible direction

Unify growing's `schema_ + text_field_resources_ + geometry_caches_` into one atomically-swapped state object that queries capture once at admission, mirroring the sealed design landed in #51694.

Related: #51693, #51694.

PR #51985, opened 2026-07-29
fix: bind segment-derived resources to published state

MrPresent-Han:
issue: #51693

Supersedes #51694. This is the takeover branch with the original change plus the follow-up review fixes.

## What changed

- Bind sealed-segment GEOMETRY caches and TEXT LOB reader handles to `PublishedSegmentState::RuntimeResourceState`, so old published snapshots retain their matching derived resources.
- Replace the capacity-based `TextColumnCache` with reference-counted `SharedTextLobReader` handles behind a process-global weak registry.
- Acquire partition TEXT readers only for loaded internal V3 TEXT columns; schema-only and external TEXT fields do not create readers.
- Keep growing-segment TEXT representation state and GEOMETRY caches in lifetime-safe shared resource state, including field-drop cleanup.
- Capture schema snapshots with `get_schema_snapshot()` so in-flight expressions do not retain dangling schema references across reopen.
- Invalidate expression-result bitsets while the publish gate is still closed when `LoadDiff::ReplacesExistingFieldData()` is true.
- Preserve structured storage error categories across the TEXT reader boundary.

## Takeover review fixes

- Pin the selected GEOMETRY cache with a `shared_ptr` for the complete GIS refinement pass.
- Pass the operation's GEOS context into non-prepared `equals` and `dwithin` fallbacks, avoiding a context from another generation or thread.
- Use the expression's captured schema snapshot in the unary JSON/ngram path.
- Route the sealed test-only index drop through the reopen lock and captured published schema.
- Fail clearly if Loon FFI properties were not initialized before growing TEXT reader setup.
- Tighten the design docs so they describe the read-lease guarantee without claiming a universal per-operation snapshot pin.

## Scope

The expression-result cache remains generation-free (`segment_id + predicate`, validated by row count). A data-replacing reopen invalidates it before admitting readers; an add-only reopen retains valid entries.

Complete sealed-segment accessor routing under one per-operation snapshot remains tracked by #51594. Complete growing-segment schema/resource publication remains tracked by #51781.

## Validation

- `make build-cpp-with-unittest` passed, including the Rust/Vortex bridge, `libmilvus_core.so`, and final `unittest/all_tests` link.
- Focused C++ tests passed: 253/253 across GEOMETRY/GIS, TEXT LOB, `SegmentLoadInfo`/`LoadDiff`, sealed reopen/COW state, expression-cache invalidation, growing reopen, and the segment read gate.
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/merr/...` passed from `pkg/`.
- `clang-format-15 --dry-run --Werror` passed for the touched C++ files.
- `git diff --check` passed.
- A full `make test-go` result is not claimed. A prior local attempt was not clean because of environment/baseline failures including occupied port 19530, `/var/lib/milvus` permissions, a noexec Go cache, stale missing package paths, Woodpecker timeout, and live etcd/mixcoord interaction.


internal/core/src/segcore/TextLobReader.h
```diff
@@ -0,0 +1,148 @@
...
+        google::protobuf::RepeatedPtrField<std::string>* dst);
+
+ private:
+    std::shared_ptr<milvus_storage::lob_column::LobColumnReader>
+    GetOrCreateReaderLocked(std::shared_ptr<arrow::fs::FileSystem> fs,
+                            const milvus_storage::api::Properties& properties);
+};
+
+// TextLobReaderRegistry only deduplicates SharedTextLobReader objects by LOB
+// path. It owns no readers and has no capacity or eviction policy: the map
```
buqian-zilliz, line 98: **[P1] Bound the retained TEXT LOB file-reader cache**

Loaded segment runtime states strongly own each SharedTextLobReader, while this registry explicitly provides no capacity or eviction. The underlying LobColumnReader retains every opened VortexFormatReader in an unbounded file_id map and releases it only on ClearCache or destruction. Consequently, as a partition/field accumulates LOB files through flush and compaction, readers touched by retired segments remain open as long as any segment on that path stays loaded; the removed outer cache previously evicted inactive path readers at 64. Bound or close the inner per-file cache independently of the lightweight snapshot handle, otherwise long-lived TEXT workloads can exhaust file descriptors and native memory.

internal/core/src/segcore/SegmentGrowingImpl.cpp
```diff
@@ -2700,6 +2813,24 @@ SegmentGrowingImpl::Reopen(SchemaPtr sch) {
...
         }
 
+        // Derived resources have the same logical lifetime as their fields.
+        // Queries that already captured a shared_ptr keep the old generation
+        // alive; future queries cannot reacquire it after schema publication.
+        if (!removed_fields.empty()) {
+            {
+                std::unique_lock lock(text_field_resources_mutex_);
+                for (auto field_id : removed_fields) {
+                    text_field_resources_.erase(field_id);
```
buqian-zilliz, line 2823: **[P1] Retire growing resources with the schema generation**

These resources are erased while schema_ still exposes the dropped field. Growing query paths do not hold sch_mutex_, and TEXT retrieval captures TextFieldResourceState lazily, so a request can observe the old schema after this erase and then fail because its spillover/reader state is gone. EnsureArrayOffsetsForStructField also runs before the schema store and can throw, leaving the old schema published without the resources it requires. Publish schema and resources as one generation, or retain retired resources until old readers drain.

internal/core/src/segcore/SegmentGrowingImpl.cpp
```diff
@@ -3155,7 +3302,34 @@ SegmentGrowingImpl::fill_empty_field(const FieldMeta& field_meta) {
...
+                   "TEXT default data for field {} has {} rows, expected {}",
+                   field_id.get(),
+                   values.data_size(),
+                   missing);
+        std::vector<std::string> refs(missing);
+        for (int64_t i = 0; i < missing; ++i) {
+            if (field_meta.is_nullable() && !data->valid_data(i)) {
+                refs[i] = spillover->WriteAndEncode(nullptr, 0);
+            } else {
+                refs[i] = spillover->WriteAndEncode(values.data(i));
```
buqian-zilliz, line 3318: **[P1] Make TEXT default backfill idempotent**

For a non-empty TEXT default, this loop appends the identical payload once per existing row, turning a schema add into O(row_count * default_size) spillover I/O and disk usage even though all rows can share one immutable encoded reference. More importantly, ordinary string columns use the disabled NoOpOffsetMapping, so a failed Reopen retry sees filled == 0, appends the full backfill again, and only overwrites the in-memory references, leaving the previous bytes unreachable. Repeated lazy reopen retries can fill the mmap disk; encode the default once and reuse its reference, and stage or truncate spillover writes so retries are idempotent.

internal/core/src/segcore/SegmentLoadInfo.h
```diff
@@ -155,6 +155,22 @@ struct LoadDiff {
...
+    // must be invalidated. New-field-only changes (fields_to_fill_default,
+    // *_to_load, index (re)builds over unchanged values) do not alter an
+    // existing predicate's result and are intentionally excluded.
+    [[nodiscard]] bool
+    ReplacesExistingFieldData() const {
+        return !binlogs_to_replace.empty() ||
+               !column_groups_to_replace.empty() ||
+               !column_groups_to_lazyreplace.empty() ||
+               !field_data_to_drop.empty() || !fields_to_reload.empty() ||
+               manifest_updated || load_external_manifest;
```
buqian-zilliz, line 171: **[P2] Preserve filter caches for add-only manifest updates**

ComputeDiff sets manifest_updated whenever the manifest-path JSON changes; every new manifest version changes that string, including add-only internal updates for which ComputeDiffColumnGroups emits only *_to_load entries. This predicate therefore classifies those reopens as data replacement and purges every cached filter bitset even though no previously queryable value changed, defeating the intended add-only cache retention. For internal manifests, invalidate from the actual replace/drop/reload buckets rather than the version bump alone, while keeping conservative handling for external full reloads.

Issue #51992, opened 2026-07-29
[Feature]: Support staged online schema evolution


congqixia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Milvus currently handles most schema mutations as a single broadcasted schema replacement. In a distributed deployment, Proxy, StreamingNode, QueryNode, DataCoord, and QueryCoord may observe schema changes at different times. This creates correctness risks for add field, drop field, function output fields, query-side load/balance, and stale read/write requests.

Known risks include mixed schema/data visibility, stale queries referencing dropped fields, function output fields becoming readable before backfill/index/IDF readiness, and load/balance racing with schema WAL consumption.

### Describe the solution you'd like.

Introduce a staged online schema evolution protocol for collection schema DDL.

  The design should cover:

  - field lifecycle states such as adding/created/dropping/dropped
  - full, write, read, and describe schema views
  - WAL barrier semantics for schema-changing AlterCollection messages
  - one schema version per segment
  - schema DDL admission gates
  - load/balance serialization during schema change for the Milvus 3.0 rollout
  - readiness gates for backfill, index, QueryCoord/QueryNode, BM25 IDF, and external collection refresh
  - safe drop-field behavior where stale destructive reads fail as input errors
  - primary/secondary cluster boundaries where historical processing is handled independently

  Related design document:

  - `docs/design-docs/design_docs/20260715-online-schema-evolution.md`

### Describe an alternate solution.

One alternative is to keep the current single-step schema replacement model and patch individual race cases. This is risky
  because each component may still observe schema and data readiness independently.

  Another alternative is to implement full multi-version segment serving immediately. This would provide stronger semantics
  but is too large for the first rollout and is not required to fix the current schema/load/balance race.

### Anything else? (Additional Context)

The first production iteration should prioritize the Milvus 3.0 safety gate:

  1. Block or drain load/reload/manual balance/automatic balance while schema change is being applied for the collection.
  2. Release the gate only after schema-change broadcast and required QueryCoord/QueryNode update callbacks complete.
  3. Ensure each segment remains bound to exactly one schema version.

  Full atomic schema/data-view publish, transparent multi-version segment serving, and Batch Update semantics can be handled
  in later iterations.

PR #51993, opened 2026-07-29
doc: add online schema evolution design

congqixia:
Related to #51992

Also capture review conclusions for the Milvus 3.0 safety gate around schema changes, load/balance serialization, IDF readiness, external collection refresh, and primary/secondary cluster boundaries.

### DIFF
--- a/docs/design-docs/design_docs/20260715-online-schema-evolution.md
+++ b/docs/design-docs/design_docs/20260715-online-schema-evolution.md
@@ -1,0 +1,1253 @@
+# Design Document: Online Schema Evolution for Milvus
+
+**Date**: July 2026
+**Status**: Draft
+**Scope**: Add field, drop field, add function field, drop function field, schema DDL admission gates
+**Feature DRI**: @Congqi Xia
+**Primary Approver**: @tedxu
+**Independent Approver**: @chyezh
+**Design Review**: 2026-07-16 & 2026-07-27
+**Review Record**:
+
+- 2026-07-16: https://zilliverse.feishu.cn/docx/J2TRdjO2xoxDHzxy7BAcbsj8nnh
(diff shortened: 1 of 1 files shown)


congqixia:
/hold
wait for approver stamps

docs/design-docs/design_docs/20260715-online-schema-evolution.md
```diff
@@ -0,0 +1,1253 @@
...
+For add field/function field:
+
+```
+create invisible field -> build snapshot data/index/query view -> publish visible field + data view
+```
+
+For drop field/function field:
+
+```
+mark invisible/dropping -> drain stale read/write windows -> physically remove field
```
tedxu: Note that the `dropping` to `dropped` status transition doesn't depend on physical data removal. The compaction task that removes the data runs asynchronously and won't affect the state change.

tedxu (commented):
/lgtm

chyezh:
Two constraints should be made explicit in the design:

1. The Phase 2 broadcast is cluster-local and must not be replicated to the backup/secondary cluster. Each cluster must run its own readiness gates and emit its own local Phase 2 broadcast only after those gates pass.

2. The schema-version sequence should be N -> N+1 -> N+1 for both Add and Drop. Phase 1 installs the target write-schema epoch and increments the schema version. Phase 2 only publishes visibility or performs final cleanup, so it must keep N+1 and must not create another write/segment schema epoch. In particular, the Drop Phase 1 increment is required, while the FieldState/final-cleanup change in Phase 2 does not increment the schema version.

docs/design-docs/design_docs/20260715-online-schema-evolution.md
congqixia: done

Issue #45691, opened 2025-11-19
[Feature]: support search from external tables


sunby:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

support create vector index and search from external tables like iceberg, lance, and other table format.

PR #51997, opened 2026-07-29
fix: use serving cluster ID for external table embeddings

weiliu1031:
issue: #45691

## What changed

- Add the serving cluster ID to
  `RefreshExternalCollectionTaskRequest`.
- Populate the value when DataCoord builds the refresh payload.
- Reuse the payload value for unified `CreateTask` properties.
- Normalize legacy payloads from `CreateTask` properties when the
  payload cluster ID is empty.
- Read the normalized payload directly for manager registration and
  function execution.
- Remove the separate constructor parameter and worker config fallback.

## Why

A shared DataNode worker can execute external-table refresh tasks from
multiple serving clusters. Its process-level cluster prefix identifies
the worker rather than the cluster that submitted the task, causing
model authentication and instance metadata to use the wrong identity.

## Behavior

External refresh tasks carry their serving cluster ID in the task
payload. DataNode uses the request value when registering the manager
task and running embeddings.

During rolling upgrades, an older DataCoord may omit the new payload
field. DataNode fills an empty value from `CreateTask` routing
properties before manager registration and embedding execution.
Non-empty payload values remain authoritative. DataNode never uses its
process configuration and does not reconcile mismatched non-empty
values with properties.

## Validation

Not run for this exact candidate during preview.


congqixia (commented):
I found one issue:

  In `internal/datanode/services.go`, `CreateTask` now registers the external refresh task with `req.GetClusterID()`, but it does not validate that the payload cluster ID is present.
  `QueryTask` and `DropTask` still use the cluster ID from task properties, so with a mixed-version producer or any request that omits the new payload field, DataNode will accept the
  task under key `("", taskID)`. Later DataCoord queries with the serving cluster prefix and gets `task result not found`; DropTask also cannot cancel the registered task.

  Could we either fail fast when `RefreshExternalCollectionTaskRequest.ClusterID` is empty, or keep a compatibility fallback to the `CreateTask` properties cluster ID when the payload
  field is missing? A regression test for the missing-payload-cluster case would make this contract explicit.

weiliu1031:
Thanks for catching this. I added a rolling-upgrade fallback at the
CreateTask boundary: when
RefreshExternalCollectionTaskRequest.clusterID is empty, DataNode
copies the serving cluster ID from task properties before manager
registration and embedding execution. Non-empty payload values remain
authoritative, and downstream code remains payload-based. I also added
coverage for the legacy payload and the missing-property error path.

congqixia (approved):
/lgtm

PR #51156, opened 2026-07-08
fix: treat unset IdField in zero-hit search results as empty in chain converter

tinswzy:
issue: #50969

### Problem

`hybrid_search` + `group_by` returning an empty result fails with `service internal error: unsupported ID type` (nightly `test_hybrid_search_group_by_empty_results`, master only). Despite the issue's original scope note, CI Loki shows the identical failure daily on the pulsar-based distributed deployment as well as both woodpecker deployments — the path is MQ-agnostic. Full RCA: https://github.com/milvus-io/milvus/issues/50969#issuecomment-4911084753

Zero-hit results reach the proxy carrying `Ids: &schemapb.IDs{}` — a non-nil shell whose `IdField` oneof is unset (`emptySearchResultData` in `querynodev2/tasks/search_task.go`, and the zero-hit fill in `querynodev2/segments/search_reduce.go`). With a single channel, `reduceAdvanceGroupBy` forwards that sub-result verbatim (skipping `setupIdListForSearchResult`), and the rerank operator's `chain.FromSearchResultData` only routed `Ids == nil` to the empty-`$id`-column path — the shell fell into `importIDs`' default branch and errored. With ≥2 channels the normal reduce path installs a typed empty `IdField`, which is why the bug only surfaces on single-shard collections. The same passthrough exists in `reduceSearchResultDataNoGroupBy`'s single-shard fast path, so plain search + function rerank with empty results could hit the same error.

### Fix (consumer-side)

In `FromSearchResultData`, an unset `IdField` carries no IDs and is now treated exactly like `Ids == nil`:

- shell / nil + zero rows → empty Int64 `$id` column (existing `importEmptyIDs` path)
- shell + rows present → still rejected as `unsupported ID type` (malformed input, unchanged semantics)
- nil + rows present → unchanged tolerated passthrough without `$id` column (aggregation-style results)

Fixing at the consumer covers all producers of the zero-hit shell (querynode empty result, delegator reduce, both proxy single-channel fast paths).

### Testing

- New regression test `TestFromSearchResultData_EmptyIDsZeroRows`: table-driven over `Ids == nil` and the unset-oneof shell (nq=1 and nq=2), asserting both produce the identical empty DataFrame; written first and observed failing with `unsupported ID type` before the fix.
- New `TestFromSearchResultData_NilIDsWithRows` pins the pre-existing nil-with-rows passthrough; existing `TestFromSearchResultData_UnsupportedIDType` (shell with rows → error) still passes.
[...]

Issue #50969, opened 2026-07-01
[Bug]: [Nightly] hybrid_search group_by empty result returns unsupported ID type in distributed woodpecker service mode


NicoYuan1986:
## Environment

- **Milvus Version**: master nightly
- **Deployment Mode**: distributed-woodpecker-service
- **MQ**: woodpecker service
- **SDK**: Python client test framework
- **OS**: Nightly CI environment

### Option B: Steps

1. Run Nightly CI trigger build 270 on master:
   https://jenkins-milvus-ci.milvus.io/blue/organizations/jenkins/MILVUS-CI-V2-PR-PIPELINES%2FDAILY-CI-TRIGGER/detail/DAILY-CI-TRIGGER/270/pipeline
2. Check the distributed-woodpecker-service downstream run.
3. Run the Python client case:
   `test_hybrid_search_group_by_empty_results`
4. The test executes `hybrid_search` with `group_by` where the search condition is expected to return empty results.

### Trigger Conditions

- **Frequency**: observed in Nightly CI
- **Scope**: only master
- **Deployment**: only distributed-woodpecker-service
- **Case**: `test_hybrid_search_group_by_empty_results`

## Expected Behavior

`hybrid_search` with `group_by` should return an empty result successfully when the search condition matches no entities.

The test should receive an empty search result and pass the expected empty-result check.

## Actual Behavior

The server returns a service internal error:

```text
unsupported ID type
```

The Python test side then continues to call `len()` on the returned Error object and raises a secondary `TypeError`.

Failure summary:

```text
hybrid_search group_by empty result returns service internal error: unsupported ID type
Test side then calls len() on the Error object and raises TypeError
case: test_hybrid_search_group_by_empty_results
scope: only master; only distributed-woodpecker-service
```

## Error Logs

Nightly CI link:

```text
https://jenkins-milvus-ci.milvus.io/blue/organizations/jenkins/MILVUS-CI-V2-PR-PIPELINES%2FDAILY-CI-TRIGGER/detail/DAILY-CI-TRIGGER/270/pipeline
```

Key error:

```text
service internal error: unsupported ID type
```

Secondary test-side error:

```text
TypeError from calling len() on the returned Error object
```

## Non-default Configuration

```yaml
deployment: distributed-woodpecker-service
```

## Analysis Hints (Optional)

This is a Nightly-only failure observed on master in distributed-woodpecker-service mode.

A related but separate closed issue exists:

[...]

PR #51999, opened 2026-07-29
fix: [3.0] treat unset IdField in zero-hit search results as empty in chain converter

tinswzy:
Cherry-pick from master

issue: #50969
pr: https://github.com/milvus-io/milvus/pull/51156

## Summary

Backport #51156 to the 3.0 branch.

Zero-hit search results may reach the proxy as a non-nil `Ids: &schemapb.IDs{}` shell whose `IdField` oneof is unset. The single-channel reduce fast path forwards that shell to `chain.FromSearchResultData`, where it was previously treated as an unsupported ID type.

Treat an unset `IdField` like `Ids == nil` for zero-row results, while preserving the existing behavior for malformed shells with rows and nil IDs with rows.

## Testing

- `gofmt -l internal/util/function/chain/converter.go internal/util/function/chain/converter_test.go` is clean.
- `git diff --check upstream/3.0...HEAD` is clean.
- Focused chain tests compile and link locally. Execution in the temporary worktree is blocked by the local native runtime configuration (`libmilvus_core.dylib` has no runtime rpath); the 3.0 CI run is required for full execution verification.

### DIFF
--- a/internal/util/function/chain/converter.go
+++ b/internal/util/function/chain/converter.go
@@ -278,7 +278,11 @@ func FromSearchResultData(resultData *schemapb.SearchResultData, alloc memory.Al
 	}
 
 	// Import ID column ($id)
-	if ids := resultData.GetIds(); ids != nil {
+	// Zero-hit results may carry Ids as a non-nil shell whose IdField oneof is
+	// unset (e.g. querynode emptySearchResultData forwarded verbatim by the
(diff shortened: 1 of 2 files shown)


czs007 (approved):
/lgtm

czs007 (approved):
/lgtm

Issue #52002, opened 2026-07-29
[Enhancement]: Index and stats task slots are charged per column group instead of per column


xiaocai2333:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

DataCoord should size the index build and stats task slots by the data the task actually reads, which is a single column, instead of by the whole column group that column lives in.

Today:

- `indexInspector.createIndexForSegment` uses `segment.getFieldBinlogSize(fieldID)`, which returns the `MemorySize` of the **whole column group** whenever the field is listed in `ChildFields`.
- `statsInspector.SubmitStatsTask` uses `originSegment.getSegmentSize()` (doubled for the json key index job), i.e. the **whole segment**, even though the json key index and the text index tasks only read the columns they index.

Both then feed `calculateIndexTaskSlot` / `calculateStatsTaskSlot`, which scale linearly above 512MB.

### Why is this needed?

Storage v3 readers project a single column: `GetFieldDatasFromStorageV2` builds an arrow schema holding only the target column before reading (`internal/core/src/storage/Util.cpp`), so parquet only touches that column inside the column group file. The slot usage therefore over-states the work by the ratio of the whole group to the indexed column.

This is worst for external collections. Their segments carry one synthetic column group listing every field in `ChildFields`, whose `MemorySize` covers all columns at once (`buildFakeBinlogs` in `internal/datanode/external/task_update.go`). Every field of an external segment is charged the whole segment:

- a 128 dim float vector over 3000 rows really needs ~1.5MB, i.e. the smallest slot bucket (1 slot), but a 2GB synthetic group asks for `(2GB / 512MB) * dataCoord.slot.indexTaskSlotUsage` = 256 slots;
- a json key index task over a wide external table is sized as if it read every column of the segment, twice.

The effect is much reduced concurrency on the datanode pool for external collections, and to a lesser extent for regular storage v3 segments whose scalar fields share the short column group (dense vector fields usually get a column group of their own via `avgSizePolicy`, so they are already accurate).

### Anything else?

A schema-only estimation is not enough on its own: `typeutil.EstimateSizePerRecord` gives an exact size for fixed width types (scalars, dense vectors) but only a configured guess for variable length types (`common.dynamicFieldLengthAvg`, 512B per row by default for JSON), which can understate a large json column by an order of magnitude and lead to over-admission.

[...]

PR #52003, opened 2026-07-29
enhance: attribute column group size per field for task slot estimation

xiaocai2333:
issue: #52002

### What this PR does

DataCoord sized the index build and the json/text index stats tasks by the whole column group (or the whole segment) holding the indexed column, while those tasks read a single column: a manifest-backed StorageV3 read goes through `GetFieldDatasFromManifest`, which asks the reader for just the target column (`internal/core/src/storage/Util.cpp`). (`GetFieldDatasFromStorageV2`, in contrast, materializes the whole column group — which is why per-field attribution is applied only to manifest-backed segments.)

External collections were the worst case, since their segments carry one synthetic column group listing every field (`buildFakeBinlogs` in `internal/datanode/external/task_update.go`): a 128 dim float vector over 3000 rows really needs ~1.5MB (1 slot) but a 2GB synthetic group asked for `(2GB / 512MB) * indexTaskSlotUsage` = 256 slots.

This PR attributes the measured column group size to the fields it holds:

- fields whose schema gives an exact size (fixed width scalars except Bool, non-nullable dense vectors) take exactly that size, and **only these** are deducted from the group budget;
- **nullable fixed width fields** are charged their schema width plus a small nullable-encoding overhead (binary offsets + validity bitmap, 8 B/row) — an upper bound, since null rows only shrink the stored data. This is what keeps the attribution effective on non-milvus-table external collections, where create time forces every user field nullable (Pass 2 of `NormalizeAndValidateExternalCollectionSchema`), so nothing there is exact;
- everything else — variable length fields (VarChar / Text / JSON / Array / Geometry / sparse and array vectors) and Bool (bit-packed on disk, so the 1 B/row schema guess is not exact) — is charged the **whole measured residual** of its group (group size minus the exact fields), computed in whole bytes;
- the result is bounded by the column group size.

The residual is deliberately **not** split proportionally between several variable length fields: the schema weights are configured guesses (`common.dynamicFieldLengthAvg`), and splitting by them could understate a fat column and over-admit tasks. Charging the full residual once per group keeps the estimate an upper bound — it can over-estimate an individual column, never under-estimate it. A schema-only estimation would not be enough on its own for the same reason: the configured guess can understate a large json column by an order of magnitude.

### Where it applies

| task | before | after |
| --- | --- | --- |
| index build | whole column group | the indexed column's share |
| json key index stats | whole segment, doubled | the json columns' share, doubled ¹ |
| text index stats | whole segment | the match enabled columns' share ¹ |
| sort and other stats jobs | whole segment | unchanged |

¹ The share of a variable length column is the group size minus the exact (fixed width, non-nullable) fields. On a non-milvus-table external collection every user field is nullable, so nothing is deductible and these tasks keep the whole (synthetic) group — the attribution helps them on internal StorageV3, milvus-table external, and groups holding function-output fields.

### Scope and fallbacks

Manifest-backed V3 segments do not persist their `FieldBinlog` arrays (the datacoord catalog skips the per-FieldBinlog KVs, see `isV3Segment`), so the per-column attribution is effective for segments flushed or refreshed **during the current DataCoord process lifetime**. A segment recovered after a restart carries no binlogs and keeps the previous whole-segment estimation until it is rewritten; the recovery paths read only the local collection cache and never load through rootcoord.

Any segment without a recorded group size, a field with no binlog, or a schema that cannot be resolved also keeps the previous whole segment estimation, so every fallback is exactly as conservative as today.

### Tests

`internal/datacoord` passes in full. New unit tests cover the estimation (`estimateFieldsReadSize`, `estimateGroupFieldsSize`, `fieldSizePerRecord`, `segmentColumnGroups`, `isFixedWidthType`), both call sites (`estimateIndexFieldSize`, `estimateStatsTaskSize`, `statsTaskFieldIDs`), the recovery paths with catalog-realistic segments (manifest set, binlog arrays empty), and the Bool / nullable / sub-row-remainder attribution cases.

### DIFF
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -581,9 +581,15 @@ func calculateIndexTaskSlot(fieldSize, numRows int64, indexParams []*commonpb.Ke
 // a field, i.e. the estimation cannot be off. Variable length types only get a
 // configured guess out of the schema (see typeutil.EstimateSizePerRecord), so
 // they are attributed from the measured data instead whenever possible.
+//
+// Bool is deliberately absent: the schema charges it 1 byte per row while the
+// measured group size accounts it bit-packed (~1/8 byte per row, see
+// ActualSizeInBytes handling of arrow.BOOL), and external sampling can round
+// it down to 0. Treating it as exact would over-deduct from the residual and
+// understate the variable length columns sharing its group.
 func isFixedWidthType(dataType schemapb.DataType) bool {
 	switch dataType {
-	case schemapb.DataType_Bool, schemapb.DataType_Int8, schemapb.DataType_Int16,
+	case schemapb.DataType_Int8, schemapb.DataType_Int16,
 		schemapb.DataType_Int32, schemapb.DataType_Int64, schemapb.DataType_Float,
 		schemapb.DataType_Double, schemapb.DataType_Timestamptz,
 		schemapb.DataType_BinaryVector, schemapb.DataType_FloatVector,
@@ -655,6 +661,13 @@ func segmentColumnGroups(segment *SegmentInfo) map[int64]*columnGroup {
 
 // estimateFieldsReadSize estimates how much data a task reads for fieldIDs of a
 // segment.
+//
+// It works off segment.GetBinlogs(), which for manifest-backed StorageV3
+// segments only exists in memory: the catalog does not persist their
+// FieldBinlog arrays (see isV3Segment), so after a DataCoord restart such
+// segments carry no binlogs and every field errors out here, letting the
+// callers keep their conservative whole-segment fallback until the segment is
+// rewritten.
 //
 // Both the index build and the json/text stats tasks request a single column at
 // a time. Per-field attribution is safe only for a StorageV3 segment with a
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -30,6 +30,7 @@ import (
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
+	"github.com/milvus-io/milvus/internal/storage"
 	"github.com/milvus-io/milvus/internal/util/indexparamcheck"
 	"github.com/milvus-io/milvus/internal/util/vecindexmgr"
 	"github.com/milvus-io/milvus/pkg/v3/common"
@@ -644,14 +645,15 @@ func segmentColumnGroups(segment *SegmentInfo) map[int64]*columnGroup {
 // estimateFieldsReadSize estimates how much data a task reads for fieldIDs of a
 // segment.
 //
-// Both the index build and the json/text stats tasks read a single column at a
-// time: for storage v3 the reader is built with a schema holding only the target
-// column, so parquet only touches that column inside the column group file (see
-// GetFieldDatasFromStorageV2 in internal/core/src/storage/Util.cpp). The binlog
-// size however is recorded per column group, so a field sharing a group with
-// others gets charged for all of them. This is worst for external collections,
-// whose segments carry one synthetic group holding every column at once (see
-// buildFakeBinlogs in internal/datanode/external).
+// Both the index build and the json/text stats tasks request a single column at
+// a time. Per-field attribution is safe only for a StorageV3 segment with a
+// manifest: GetFieldDatasFromManifest asks the reader for the target column,
+// while GetFieldDatasFromStorageV2 materializes every column of the selected
+// column group. The binlog size however is recorded per column group, so a
+// projected field sharing a group with others otherwise gets charged for all of
+// them. This is worst for external collections, whose segments carry one
+// synthetic group holding every column at once (see buildFakeBinlogs in
+// internal/datanode/external).
 //
 // The group size is the measured truth for the group as a whole, so it is used
 // as the budget: fields whose schema gives an exact size take exactly that, and
@@ -670,6 +672,12 @@ func estimateFieldsReadSize(schema *schemapb.CollectionSchema, segment *SegmentI
 	numRows := segment.GetNumOfRows()
 	if numRows <= 0 {
 		return 0, nil
+	}
+	if !supportsFieldProjection(segment) {
+		// Callers gate this optimization on supportsFieldProjection. Reaching
+		// this branch means an internal contract was violated, not bad input.
+		return 0, merr.WrapErrServiceInternalMsg(
+			"segment %d does not have a StorageV3 manifest for projected field reads", segment.GetID())
 	}
 
 	groups := segmentColumnGroups(segment)
@@ -692,6 +700,14 @@ func estimateFieldsReadSize(schema *schemapb.CollectionSchema, segment *SegmentI
 		total += size
 	}
 	return total, nil
+}
+
+// supportsFieldProjection matches the actual reader selection in the index and
+// stats workers. A non-empty manifest selects GetFieldDatasFromManifest, which
+// projects the requested column. StorageV2 files without a manifest are read by
+// GetFieldDatasFromStorageV2, which materializes the whole column group.
+func supportsFieldProjection(segment *SegmentInfo) bool {
+	return segment.GetStorageVersion() == storage.StorageV3 && segment.GetManifestPath() != ""
 }
 
 // estimateGroupFieldsSize attributes the measured size of one column group to
@@ -719,6 +735,10 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 	estimates := make(map[int64]fieldEstimate, len(group.fields))
 	var fixedPerRecord int64
 	var hasVariableField bool
+	var externalColumnResolver *typeutil.StorageColumnResolver
+	if typeutil.IsExternalCollection(schema) {
+		externalColumnResolver = typeutil.NewStorageColumnResolver(schema)
+	}
 	for _, fieldID := range group.fields {
 		size, exact, err := fieldSizePerRecord(schema, fieldID)
 		if err != nil {
@@ -733,7 +753,17 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 		}
 		estimates[fieldID] = fieldEstimate{sizePerRecord: size, exact: exact}
 		if exact {
-			fixedPerRecord += size
+			measured := true
+			if externalColumnResolver != nil {
+				field := typeutil.GetFieldByID(schema, fieldID)
+				_, measured = externalColumnResolver.SourceDataColumnName(field)
+				// Refresh adds generated function outputs to MemorySize after
+				// sampling the source columns, so they are measured too.
+				measured = measured || typeutil.IsFunctionOutputField(schema, field)
+			}
+			if measured {
+				fixedPerRecord += size
+			}
 			continue
 		}
 		hasVariableField = true
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -803,44 +803,6 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 	return total, nil
 }
 
-// collectionCache memoizes collection lookups over a loop that resolves the
-// same collections repeatedly. It goes through the handler, which loads from
-// rootcoord on a cache miss: the datacoord collection cache is filled by an
-// async goroutine (see Server.initMeta), so it is usually still empty while the
-// inspectors recover their tasks at startup. That load costs up to 5 retries
-// with a 10s timeout, hence the memoization - including of misses.
-type collectionCache struct {
-	handler Handler
-	cached  map[int64]*collectionInfo
-}
-
-func newCollectionCache(handler Handler) *collectionCache {
-	return &collectionCache{handler: handler, cached: make(map[int64]*collectionInfo)}
-}
-
-func (c *collectionCache) get(ctx context.Context, collectionID int64) *collectionInfo {
-	if coll, ok := c.cached[collectionID]; ok {
-		return coll
-	}
-	coll := resolveCollection(ctx, c.handler, collectionID)
-	c.cached[collectionID] = coll
-	return coll
-}
-
-// resolveCollection returns the collection info, loading it from rootcoord when
-// the datacoord cache does not hold it yet.
-func resolveCollection(ctx context.Context, handler Handler, collectionID int64) *collectionInfo {
-	if handler == nil {
-		return nil
-	}
-	coll, err := handler.GetCollection(ctx, collectionID)
-	if err != nil {
-		mlog.Warn(ctx, "failed to resolve collection", mlog.FieldCollectionID(collectionID), mlog.Err(err))
-		return nil
-	}
-	return coll
-}
-
 func calculateStatsTaskSlot(segmentSize int64) int64 {
 	defaultSlots := Params.DataCoordCfg.StatsTaskSlotUsage.GetAsInt64()
 	if segmentSize > 512*1024*1024 {
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -655,10 +655,14 @@ func segmentColumnGroups(segment *SegmentInfo) map[int64]*columnGroup {
 //
 // The group size is the measured truth for the group as a whole, so it is used
 // as the budget: fields whose schema gives an exact size take exactly that, and
-// the remaining bytes are shared between the variable length fields
-// proportionally to their schema estimate. That keeps a 128 dim vector at its
-// real size while a large json column still gets the bytes it actually
-// occupies, instead of the configured per row guess.
+// whatever remains is charged to the variable length fields. That keeps a 128
+// dim vector at its real size, while a json column is charged what the data
+// really contains instead of the configured per row guess
+// (common.dynamicFieldLengthAvg), which can be an order of magnitude off. The
+// residual is not split proportionally between several variable length fields:
+// the schema weights are guesses, and splitting by them could understate a fat
+// column, i.e. over-admit tasks. Charging it once keeps the result an upper
+// bound, which is what a slot must be.
 func estimateFieldsReadSize(schema *schemapb.CollectionSchema, segment *SegmentInfo, fieldIDs []int64) (int64, error) {
 	if len(fieldIDs) == 0 {
 		return 0, merr.WrapErrParameterInvalidMsg("no field to estimate size for")
@@ -703,28 +707,41 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 		return group.size, nil
 	}
 
+	requestedFields := make(map[int64]struct{}, len(fieldIDs))
+	for _, fieldID := range fieldIDs {
+		requestedFields[fieldID] = struct{}{}
+	}
+
 	type fieldEstimate struct {
 		sizePerRecord int64
 		exact         bool
 	}
 	estimates := make(map[int64]fieldEstimate, len(group.fields))
-	var fixedPerRecord, variableWeight int64
+	var fixedPerRecord int64
+	var hasVariableField bool
 	for _, fieldID := range group.fields {
 		size, exact, err := fieldSizePerRecord(schema, fieldID)
 		if err != nil {
-			return 0, err
+			if _, requested := requestedFields[fieldID]; requested {
+				return 0, err
+			}
+			// A group member the schema cannot resolve, e.g. a field dropped
+			// after this segment was flushed and before compaction rewrote it.
+			// Skipping it leaves its bytes in the residual instead of giving up
+			// on the whole group.
+			continue
 		}
 		estimates[fieldID] = fieldEstimate{sizePerRecord: size, exact: exact}
 		if exact {
 			fixedPerRecord += size
 			continue
 		}
-		variableWeight += size
+		hasVariableField = true
 	}
 
 	// bytes the variable length fields of this group take per row, measured
 	residualPerRecord := group.size/numRows - fixedPerRecord
-	if residualPerRecord < 0 || variableWeight == 0 {
+	if residualPerRecord < 0 || !hasVariableField {
 		// the schema does not add up to the measured data, e.g. because some
 		// field of the group is not materialized in it. Fall back to the schema
 		// estimation alone, still bounded by the group size.
@@ -732,18 +749,62 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 	}
 
 	var total int64
+	var residualCharged bool
 	for _, fieldID := range fieldIDs {
 		estimate := estimates[fieldID]
-		sizePerRecord := estimate.sizePerRecord
 		if !estimate.exact && residualPerRecord > 0 {
-			sizePerRecord = residualPerRecord * sizePerRecord / variableWeight
+			// every variable length field of the group shares the same residual,
+			// so it is charged once no matter how many of them are requested
+			if !residualCharged {
+				total += residualPerRecord * numRows
+				residualCharged = true
+			}
+			continue
 		}
-		total += sizePerRecord * numRows
+		total += estimate.sizePerRecord * numRows
 	}
 	if total > group.size {
 		return group.size, nil
 	}
 	return total, nil
+}
+
+// collectionCache memoizes collection lookups over a loop that resolves the
+// same collections repeatedly. It goes through the handler, which loads from
+// rootcoord on a cache miss: the datacoord collection cache is filled by an
+// async goroutine (see Server.initMeta), so it is usually still empty while the
+// inspectors recover their tasks at startup. That load costs up to 5 retries
+// with a 10s timeout, hence the memoization - including of misses.
+type collectionCache struct {
+	handler Handler
+	cached  map[int64]*collectionInfo
+}
+
+func newCollectionCache(handler Handler) *collectionCache {
+	return &collectionCache{handler: handler, cached: make(map[int64]*collectionInfo)}
+}
+
+func (c *collectionCache) get(ctx context.Context, collectionID int64) *collectionInfo {
+	if coll, ok := c.cached[collectionID]; ok {
+		return coll
+	}
+	coll := resolveCollection(ctx, c.handler, collectionID)
+	c.cached[collectionID] = coll
+	return coll
+}
+
+// resolveCollection returns the collection info, loading it from rootcoord when
+// the datacoord cache does not hold it yet.
+func resolveCollection(ctx context.Context, handler Handler, collectionID int64) *collectionInfo {
+	if handler == nil {
+		return nil
+	}
+	coll, err := handler.GetCollection(ctx, collectionID)
+	if err != nil {
+		mlog.Warn(ctx, "failed to resolve collection", mlog.FieldCollectionID(collectionID), mlog.Err(err))
+		return nil
+	}
+	return coll
 }
 
 func calculateStatsTaskSlot(segmentSize int64) int64 {
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -659,10 +659,14 @@ func segmentColumnGroups(segment *SegmentInfo) map[int64]*columnGroup {
 //
 // The group size is the measured truth for the group as a whole, so it is used
 // as the budget: fields whose schema gives an exact size take exactly that, and
-// the remaining bytes are shared between the variable length fields
-// proportionally to their schema estimate. That keeps a 128 dim vector at its
-// real size while a large json column still gets the bytes it actually
-// occupies, instead of the configured per row guess.
+// whatever remains is charged to the variable length fields. That keeps a 128
+// dim vector at its real size, while a json column is charged what the data
+// really contains instead of the configured per row guess
+// (common.dynamicFieldLengthAvg), which can be an order of magnitude off. The
+// residual is not split proportionally between several variable length fields:
+// the schema weights are guesses, and splitting by them could understate a fat
+// column, i.e. over-admit tasks. Charging it once keeps the result an upper
+// bound, which is what a slot must be.
 func estimateFieldsReadSize(schema *schemapb.CollectionSchema, segment *SegmentInfo, fieldIDs []int64) (int64, error) {
 	if len(fieldIDs) == 0 {
 		return 0, merr.WrapErrParameterInvalidMsg("no field to estimate size for")
@@ -707,28 +711,41 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 		return group.size, nil
 	}
 
+	requestedFields := make(map[int64]struct{}, len(fieldIDs))
+	for _, fieldID := range fieldIDs {
+		requestedFields[fieldID] = struct{}{}
+	}
+
 	type fieldEstimate struct {
 		sizePerRecord int64
 		exact         bool
 	}
 	estimates := make(map[int64]fieldEstimate, len(group.fields))
-	var fixedPerRecord, variableWeight int64
+	var fixedPerRecord int64
+	var hasVariableField bool
 	for _, fieldID := range group.fields {
 		size, exact, err := fieldSizePerRecord(schema, fieldID)
 		if err != nil {
-			return 0, err
+			if _, requested := requestedFields[fieldID]; requested {
+				return 0, err
+			}
+			// A group member the schema cannot resolve, e.g. a field dropped
+			// after this segment was flushed and before compaction rewrote it.
+			// Skipping it leaves its bytes in the residual instead of giving up
+			// on the whole group.
+			continue
 		}
 		estimates[fieldID] = fieldEstimate{sizePerRecord: size, exact: exact}
 		if exact {
 			fixedPerRecord += size
 			continue
 		}
-		variableWeight += size
+		hasVariableField = true
 	}
 
 	// bytes the variable length fields of this group take per row, measured
 	residualPerRecord := group.size/numRows - fixedPerRecord
-	if residualPerRecord < 0 || variableWeight == 0 {
+	if residualPerRecord < 0 || !hasVariableField {
 		// the schema does not add up to the measured data, e.g. because some
 		// field of the group is not materialized in it. Fall back to the schema
 		// estimation alone, still bounded by the group size.
@@ -736,18 +753,62 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 	}
 
 	var total int64
+	var residualCharged bool
 	for _, fieldID := range fieldIDs {
 		estimate := estimates[fieldID]
-		sizePerRecord := estimate.sizePerRecord
 		if !estimate.exact && residualPerRecord > 0 {
-			sizePerRecord = residualPerRecord * sizePerRecord / variableWeight
+			// every variable length field of the group shares the same residual,
+			// so it is charged once no matter how many of them are requested
+			if !residualCharged {
+				total += residualPerRecord * numRows
+				residualCharged = true
+			}
+			continue
 		}
-		total += sizePerRecord * numRows
+		total += estimate.sizePerRecord * numRows
 	}
 	if total > group.size {
 		return group.size, nil
 	}
 	return total, nil
+}
+
+// collectionCache memoizes collection lookups over a loop that resolves the
+// same collections repeatedly. It goes through the handler, which loads from
+// rootcoord on a cache miss: the datacoord collection cache is filled by an
+// async goroutine (see Server.initMeta), so it is usually still empty while the
+// inspectors recover their tasks at startup. That load costs up to 5 retries
+// with a 10s timeout, hence the memoization - including of misses.
+type collectionCache struct {
+	handler Handler
+	cached  map[int64]*collectionInfo
+}
+
+func newCollectionCache(handler Handler) *collectionCache {
+	return &collectionCache{handler: handler, cached: make(map[int64]*collectionInfo)}
+}
+
+func (c *collectionCache) get(ctx context.Context, collectionID int64) *collectionInfo {
+	if coll, ok := c.cached[collectionID]; ok {
+		return coll
+	}
+	coll := resolveCollection(ctx, c.handler, collectionID)
+	c.cached[collectionID] = coll
+	return coll
+}
+
+// resolveCollection returns the collection info, loading it from rootcoord when
+// the datacoord cache does not hold it yet.
+func resolveCollection(ctx context.Context, handler Handler, collectionID int64) *collectionInfo {
+	if handler == nil {
+		return nil
+	}
+	coll, err := handler.GetCollection(ctx, collectionID)
+	if err != nil {
+		mlog.Warn(ctx, "failed to resolve collection", mlog.FieldCollectionID(collectionID), mlog.Err(err))
+		return nil
+	}
+	return coll
 }
 
 func calculateStatsTaskSlot(segmentSize int64) int64 {
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -597,6 +597,12 @@ func isFixedWidthType(dataType schemapb.DataType) bool {
 
 // fieldSizePerRecord returns the per row size of a field derived from the
 // schema, together with whether that size is exact.
+//
+// A nullable field is never exact, whatever its data type: rows holding null
+// store less than the full width (nullable vectors are even stored with a
+// variable length encoding, see CreateArrowSchema in
+// internal/core/src/storage/Util.cpp), so its real size is data dependent and
+// must come from the measured residual instead.
 func fieldSizePerRecord(schema *schemapb.CollectionSchema, fieldID int64) (int64, bool, error) {
 	field := typeutil.GetFieldByID(schema, fieldID)
 	if field == nil {
@@ -614,7 +620,8 @@ func fieldSizePerRecord(schema *schemapb.CollectionSchema, fieldID int64) (int64
 		return 0, false, merr.WrapErrParameterInvalidMsg("cannot estimate size of field %d with data type %s",
 			fieldID, field.GetDataType().String())
 	}
-	return int64(sizePerRecord), isFixedWidthType(field.GetDataType()), nil
+	exact := isFixedWidthType(field.GetDataType()) && !field.GetNullable()
+	return int64(sizePerRecord), exact, nil
 }
 
 // columnGroup is a set of fields stored together, which is what a binlog entry
@@ -773,24 +780,27 @@ func estimateGroupFieldsSize(schema *schemapb.CollectionSchema, group *columnGro
 		hasVariableField = true
 	}
 
-	// bytes the variable length fields of this group take per row, measured
-	residualPerRecord := group.size/numRows - fixedPerRecord
-	if residualPerRecord < 0 || !hasVariableField {
+	// bytes the variable length fields of this group take in total, measured.
+	// Kept in whole bytes rather than per record: a per record truncation
+	// could lose up to numRows bytes and drop the estimate into a lower slot
+	// bucket.
+	residualBytes := group.size - fixedPerRecord*numRows
+	if residualBytes < 0 || !hasVariableField {
 		// the schema does not add up to the measured data, e.g. because some
 		// field of the group is not materialized in it. Fall back to the schema
 		// estimation alone, still bounded by the group size.
-		residualPerRecord = 0
+		residualBytes = 0
 	}
 
 	var total int64
 	var residualCharged bool
 	for _, fieldID := range fieldIDs {
 		estimate := estimates[fieldID]
-		if !estimate.exact && residualPerRecord > 0 {
+		if !estimate.exact && residualBytes > 0 {
 			// every variable length field of the group shares the same residual,
 			// so it is charged once no matter how many of them are requested
 			if !residualCharged {
-				total += residualPerRecord * numRows
+				total += residualBytes
 				residualCharged = true
 			}
 			continue
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/util/indexparamcheck"
 	"github.com/milvus-io/milvus/internal/util/vecindexmgr"
 	"github.com/milvus-io/milvus/pkg/v3/common"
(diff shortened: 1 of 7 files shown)


czs007:

## Overview

The measured-residual sizing this PR introduces is correct in aggregate but under-charges fat variable-length columns and silently degrades to whole-group sizing on the index-recovery path; two smaller robustness/test gaps are also noted below.

## High

**`internal/datacoord/util.go:595` — residual is split by the same schema guess the PR rejects, under-estimating variable-length reads**
The residual budget is derived from the measured `group.size` (util.go:582), but the weights used to split it across variable-length columns are still the schema estimates this PR's own description rejects (`VarChar = min(max_length, 256)`, `JSON/Array/Geometry = 512`, `pkg/util/typeutil/field_length_limit.go:20-23`). When a column group holds two or more variable-length columns whose real/estimated ratios differ, the requested column receives less budget than it actually reads — an under-estimate, i.e. worker memory over-commitment, whereas the pre-PR whole-group estimate could only over-estimate. Failure: an external collection with `pk int64` + `content VarChar(max_length=65535, enable_match=true)` at a real 4KB/row + `meta JSON` at 100B/row over 1M rows (synthetic group measures ≈4.1KB/row) gives `TextIndexJob` `4076*256/768 ≈ 1358B/row` for `content` → ~4GB of real reads estimated as ~1.36GB, `calculateStatsTaskSlot` drops from 8× to 2× `StatsTaskSlotUsage`, roughly 4× more concurrent text-index builds are admitted on one worker, and the slot mechanism stops bounding memory (OOM risk on the index node).
This path is not limited to legacy configs: `common.storage.stv2.splitSystemColumn.enabled` defaults to true (`component_param.go:1114-1121`), so fields 0/1/PK are split into a system group and the short group is left holding exactly the JSON + match-enabled VarChar shape (`split_policy.go:151-157,267-276` only split vectors/`DataType_Text` and columns with avg ≥ 1024B). The worst case is this PR's headline scenario — an external collection's synthetic group holds every column at once (`internal/datanode/external/task_update.go:1099-1121`) and bypasses avg-size splitting entirely, so the skew is unbounded.
Suggestion: charge the full `residualPerRecord` to a requested variable-length field. It is already capped at `group.size` by util.go:599-601, so this removes the under-estimate while keeping the headline win on fixed-width columns. *(raised by czs007, chyezh)*

**`internal/datacoord/index_inspector.go:331` — collection cache is usually empty during index-task recovery, so the new estimate silently falls back to whole-group slots**
`reloadFromMeta` now computes `fieldSize` via `i.estimateIndexFieldSize(...)`, which reads `i.meta.GetCollection` (index_inspector.go:300-303). That cache is populated by an async goroutine (`server.go:638-645`), while `indexInspector.Start()` → `reloadFromMeta()` is the first synchronous step of `startDataCoord` (`server.go:413-415`, `756-759`). The async fill also depends on rootcoord's `DescribeCollection`, which is behind a health gate (`root_coord.go:1228-1233`) and retries at 1s intervals, with 3 RPCs per collection (`meta.go:442-472`) — so an empty cache here is the norm, not an edge case. Failure: restart datacoord/mixcoord while an external collection has unfinished index tasks → `coll == nil` → `estimateIndexFieldSize` returns `getFieldBinlogSize()` for the whole synthetic group, so a 2GB group with a 128-dim vector column (~1.5MB real) charges `(2GB/512MB) * IndexTaskSlotUsage = 256` slots — exactly the number this PR exists to fix — and post-restart index builds serialize behind their own slot usage, with only a single Warn (index_inspector.go:305-311).
Suggestion: use the loading variant `handler.GetCollection` (`handler.go:569-591`); the inspector already holds the handler. Note it performs up to 5 retries with a 10s timeout per cache miss, and `reloadFromMeta` calls this per segment×index — memoize by `collectionID` (or batch-prefetch the collection list before the loop), otherwise startup stretches into minutes. *(raised by czs007, chyezh)*

## Low

**`internal/datacoord/util.go:568` — a field dropped after the segment was written aborts the whole group estimate, silently disabling the optimization**
`estimateGroupFieldsSize` iterates all of `group.fields` (util.go:568-572) and returns an error for the entire group if any single field fails to resolve; `fieldSizePerRecord` returns `merr.WrapErrFieldNotFound` for fields absent from the schema (util.go:451-455). Dropping a field physically removes it from `schema.Fields` (`ddl_callbacks_alter_collection_schema.go:456-476`), while already-flushed segments still list it in `ChildFields` until compaction rewrites them. Failure: flush segments on an external collection, then `AlterCollectionSchema`-drop any field in the group (not necessarily the indexed one), then trigger an index build or json/text stats before compaction → `ErrFieldNotFound` aborts the group, both call sites fall back to whole-group/whole-segment size (256 slots for a 2GB synthetic group), and the fix is fully bypassed for those segments with only a Warn to show it.
Suggestion: skip unresolvable group members individually — leaving their bytes in the residual — rather than abandoning the group. The current direction is safe (over-estimate, equal to pre-PR behavior), but the optimization silently stops working on any collection that has dropped a field. *(raised by czs007)*

**`internal/datacoord/index_inspector_test.go:705` — fixtures' `ChildFields` omit system fields 0/1, so they do not match real external segments**
The new cases populate `ChildFields` with user fields only (`[]int64{100, vecField}`; `[]int64{pkFieldID, textFieldID, jsonFieldID}` in `stats_inspector_test.go`), but real external segments derive `ChildFields` from the full `schema.GetFields()` (`task_update.go:1099-1105`), and the persisted schema includes `RowIDField(0)` / `TimeStampField(1)` (`create_collection_task.go:464-479`), which `DescribeCollection` does not filter (`root_coord.go:1200` → `model/collection.go:154-171`). In production those two `Int64` fields are charged 8B each into `fixedPerRecord`, compressing the residual. Current behavior is correct; the gap is representativeness. Suggestion: add a case with `ChildFields: []int64{0, 1, 100, 102}` to pin the 8B fixed-width treatment of system columns so it is not later "fixed" into a skip. *(raised by czs007)*

- `internal/datacoord/stats_inspector.go:113` — the stats-task recovery path still uses `taskSlot = calculateStatsTaskSlot(segment.getSegmentSize())` while only `SubmitStatsTask` (stats_inspector.go:369) was migrated to the new estimate, so recovered tasks charge whole-segment slots. Failure: restart datacoord with Init/Retry/InProgress json-key-index or text-index stats tasks on a 2GB synthetic column group → each recovered task charges `(2GB/512MB) * StatsTaskSlotUsage` (doubled for `JsonKeyIndexJob`), frozen at enqueue time, so post-restart stats/index work runs near-serially with no warning. Unrelated to this PR because those lines are untouched by the diff (`reloadFromMeta` is outside the changed range) and produce identical values before and after it. The persisted `StatsTask` already carries `SubJobType` (`stats_task_meta.go:78,89`), so `si.estimateStatsTaskSize(segment, st.GetSubJobType())` is a one-line change — but it depends on the same collection-cache-availability problem as the index-recovery finding above, and should be done together with it, otherwise it silently falls back to `segmentSize` anyway. Recommend merging that separate fix before this PR if it lands independently. *(raised by czs007, chyezh)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/53f18e86181314f768b59eb7114e9eba3df44707...czs007:team2-fix/T372dea6a5b?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/53f18e86181314f768b59eb7114e9eba3df44707...czs007:team2-fix/T9c40a0236f?expand=1

xiaocai2333:
Addressed the adversarial review in 99fd0ef58a. Summary of what changed and why:

**High 1 — residual split by the same schema guess the PR rejects (`util.go:595`)**
Fixed. A variable-length column is now charged the **whole** measured residual of its group rather than a weighted share, exactly as suggested. The residual is charged once per group, so requesting several variable-length columns of the same group (e.g. two json columns in one `JsonKeyIndexJob`) does not count it twice, and the result stays bounded by `group.size`. Your `pk + content VarChar(65535) @4KB/row + meta JSON @100B/row` case now charges `content` the full 4076B/row residual instead of ~1358B/row, so the estimate can no longer land below the real read.

**High 2 — collection cache empty during recovery (`index_inspector.go:331`)**
Fixed. Both inspectors now resolve the collection through `handler.GetCollection`, with a `collectionCache` that memoizes per collection **including misses**, so the 5-retry/10s-timeout load is paid at most once per collection instead of once per segment×index. Verified the ordering you described: `reloadCollectionsFromRootcoord` runs in a goroutine inside `Server.initMeta` while `indexInspector.Start()` is synchronous in `startDataCoord`. New test `TestIndexInspector_ReloadFromMetaResolvesCollectionThroughHandler` pins both the per-column slot after a restart with an empty cache and the single rootcoord call for two recovered tasks (`.Once()`).

**Low 1 — a dropped field aborts the whole group (`util.go:568`)**
Fixed. An unresolvable group member is skipped and its bytes stay in the residual; the estimation only fails when the **requested** field itself cannot be resolved. Covered by `group member missing from schema is skipped` and `group member of unestimable size is skipped`.

**Low 2 — fixtures omit system fields 0/1 (`index_inspector_test.go:705`)**
Fixed. Added `system fields of a group are charged as fixed width` with `ChildFields: {0, 1, pk, json}`, pinning the 8B Int64 treatment of `RowIDField`/`TimeStampField` and the residual they leave for the json column. Confirmed `appendSysFields` puts both into the persisted schema, so this matches production.

**Pre-existing hazard — stats recovery path (`stats_inspector.go:113`)**
Included in this PR after all. You noted it should be done together with the collection-cache fix, and since that fix landed here, leaving the stats recovery path on whole-segment sizing while the index recovery path is fixed would be inconsistent. `reloadFromMeta` now calls `estimateStatsTaskSize` with the persisted `SubJobType`, sharing the same memoized cache. New test `TestReloadFromMetaEstimatesPerSubJobType` asserts a recovered `JsonKeyIndexJob` drops the pk column while a recovered `Sort` job still reads the whole segment.

`internal/datacoord` passes in full; all new functions keep 100% statement coverage including every fallback branch.

czs007:

Both findings concern the new per-field/per-group resource estimation: the slot request it produces can be far smaller than the bytes the worker actually reads, which over-admits concurrent index/stats tasks.

## High

**`internal/datacoord/util.go:578` — system fields in `ChildFields` are deducted from a budget that never included them, under-charging variable-length columns.**
External-segment sampling excludes RowID, Timestamp and VirtualPK via `SourceDataColumnName` (`pkg/util/typeutil/schema.go:2986-2988`, `2832-2836`), so the measured group `MemorySize` contains no bytes for those fields — but `buildFakeBinlogs` still lists every schema field in `ChildFields` (`internal/datanode/external/task_update.go:1099-1105`), and the estimator subtracts their fixed widths from that budget again (`fixedPerRecord += size`; `residualPerRecord := group.size/numRows - fixedPerRecord`). Failure: a 1M-row external segment with an Int64 PK plus a ~20B/row JSON column records `MemorySize` ≈ 28MB with `ChildFields = [0, 1, 100, 101]` → `fixedPerRecord` = 24B (RowID + Timestamp + PK), so JSON is charged 4MB (8MB after the stats x2) and the task requests `StatsTaskSlotUsage/8` while really handling ~40MB (`StatsTaskSlotUsage/4`), admitting 2x more such tasks per worker than the slot budget was sized for. The error grows without bound as the external table narrows — a 24B/row table with a virtual PK gets residual 0 and lands in the smallest bucket regardless of real size. Suggestion: exclude fields that the sampling path skipped (`IsExternalSystemOrVirtualField`) from `fixedPerRecord`, or have `buildFakeBinlogs` size `MemorySize` over the same field set it writes into `ChildFields` so budget and deduction agree; the current test `util_test.go:1313-1334` pins the mismatched behaviour as intended and should be updated with it. *(raised by xaxys, czs007)*

**`internal/datacoord/util.go:918` — per-field attribution is applied to segments whose reader does not project columns, which under default config is every segment.**
The doc comment at `util.go:912-933` and the PR description justify per-field estimation by citing `GetFieldDatasFromStorageV2` as proof that parquet touches only the target column. That function is the counter-example: it builds `arrow_schema` and never uses it, passing `nullptr` in the schema slot of `LoadWithStrategy` (`internal/core/src/storage/Util.cpp:1489-1508`), which `internal/core/src/segcore/memory_planner.h:91` documents as "read all columns of the files"; `Util.cpp:1517-1521` then indexes the result as `table->column(col_offset)` with the group-relative offset, which is only valid because every group column was materialized. The projecting path (`GetFieldDatasFromManifest`, `needed_columns = {column_name}`, `Util.cpp:1600`) is selected only when the manifest path is non-empty (`MemFileManagerImpl.cpp:239-278`, `DiskFileManagerImpl.cpp:976-991`), and manifests are written only on the StorageV3/loon path (`syncmgr/pack_writer_v2.go:269,316` returns an always-empty manifest path; `importv2/util.go:77-83` sets one only `if useLoonFFI`) — while new segments are StorageV2 unless `common.storage.useLoonFFI`, default false (`segment_alloc_worker.go:148-152`, `component_param.go:1098-1100`). Failure: with default config, a 10M-row StorageV2 segment whose short group (`column_group_splitter.go:92-94`) holds pk + RowID + Timestamp + ~40 scalars + a ~1KB/row JSON column measures ~2GB; an inverted index on one Int64 estimates 8B*10M = 80MB, so `calculateIndexTaskSlot` (`util.go:415-427`) drops the request from `(2GB/512MB)*ScalarIndexTaskSlotUsage` to `ScalarIndexTaskSlotUsage/16` (~64x less) while the build still streams the full 2GB group (`InvertedIndexTantivy.cpp:212` → `Util.cpp:1856-1859` → `CacheRawDataToMemory` → `GetFieldDatasFromStorageV2`) → ~64x more concurrent scalar-index/stats builds admitted per indexnode, each saturating network and holding up to `DEFAULT_FIELD_MAX_MEMORY_LIMIT` (128MB, `Consts.h:86`) of arrow decode buffers, until the worker OOMs. The text-index stats task compounds it: it builds all match-enabled fields concurrently in one errgroup (`internal/datanode/index/task_stats.go:523-530`) while the estimator charges the group residual exactly once (`util.go:607-621`). Vector fields and `Text` fields get their own single-field groups (`split_policy.go:151-158`) and StorageV1 binlogs are single-field, so those paths are exact either way. Suggestion: gate per-field attribution on the actually-projecting path — `estimateFieldsReadSize` already receives the `*SegmentInfo` (`util.go:534-550`), so `segment.GetStorageVersion() == storage.StorageV3 && segment.GetManifestPath() != ""` (the predicate already exists as `canBuildExternalJSONKeyIndex`, `stats_inspector.go:199-201`) and fall back to `getFieldBinlogSize`/`getSegmentSize` otherwise; this keeps the headline win for the external/manifest-backed segments the PR targets. Please also correct the `util.go:912-933` comment and the PR description, which both cite `GetFieldDatasFromStorageV2` as evidence of column pruning, and add a no-manifest StorageV2 segment test asserting the fallback. *(raised by czs007)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/99fd0ef58aa8a95cdfde883871d3bc5661eafa66...czs007:team2-fix/Ta43c6eef1c?expand=1

czs007:

## Re-review of `99fd0ef58aa8`

- **`internal/datacoord/util.go:578` — system fields in `ChildFields` are deducted from a budget that never included them: Confirmed.** Both reviewers re-verified every link independently against the current tree and could not refute it; the persisted schema does carry `RowID`/`Timestamp` as resolvable `Int64` fields, so the "skip unresolved members" branch never fires for them.
- **`internal/datacoord/util.go:918` — per-field attribution applied to segments whose reader does not project columns: Confirmed.** `GetFieldDatasFromStorageV2` still passes `nullptr` in the schema slot of `LoadWithStrategy` and materializes the full row-group table before selecting one column, so the premise behind the attribution does not hold on the default StorageV2 path.

Both findings from the previous round stand after independent re-verification, and this round surfaced no additional issues beyond them.

## High

**`internal/datacoord/util.go:592` — system/virtual fields are deducted as fixed width from a sampling budget that never measured them, under-sizing variable-length columns on external segments.**
`estimateGroupFieldsSize` accumulates every resolvable fixed-width member of `group.fields` into `fixedPerRecord` (`util.go:592`) and then treats `group.size/numRows - fixedPerRecord` as the measured residual for variable-length columns (`util.go:599`). For external segments, half of that subtraction removes bytes that were never counted: `buildFakeBinlogs` copies *every* schema field into `ChildFields` (`internal/datanode/external/task_update.go:1101-1104`), while `MemorySize` is built purely from sampled source columns (`task_update.go:982`, filtered through `SourceDataColumnName` at `task_update.go:1156-1159`), and `SourceDataColumnName` explicitly returns `false` for `RowID`, `Timestamp`, and `VirtualPK` (`pkg/util/typeutil/schema.go:2986`, `2832-2836`). Those fields do exist in the schema DataCoord sees — `appendSysFields` adds `RowID`(0)/`Timestamp`(1) as `Int64` at collection creation (`internal/rootcoord/create_collection_task.go:464-478`), `ToCollectionSchemaPB` does not strip them (`internal/metastore/model/collection.go:159`), and PK-less external tables additionally get an `Int64` `VirtualPK` injected by the proxy (`internal/proxy/util.go:888-893`). Net effect: at least 16 B/row (24 B/row with `VirtualPK`) is subtracted from a budget that never contained it. Failure: an external parquet collection with an injected `VirtualPK` and one `enable_match` varchar column measured at 40 B/row over 50M rows, running a `TextIndexJob` stats task → the true 2 GB read (24 slots) is charged as `(40-24)*50M = 800 MB` (8 slots), a 3× under-request, so 3× the designed number of such tasks are admitted concurrently and IndexNode memory is oversold. Note the residual stays in `(0, true value)`, so the negative-value fallback at `util.go:600` never triggers.
*Suggestion:* derive the deductible set from the same membership rule that built `MemorySize` (drop members for which `SourceDataColumnName` fails), or stop listing unmeasured fields in `buildFakeBinlogs`. Please also revisit the new test at `internal/datacoord/util_test.go:335-356`: its `groupSize := (8 + 8 + pkSize + residualPerRow) * numRows` includes the two system columns in the group size, which is the opposite of what real external segments produce — as written it locks in the incorrect invariant rather than covering this case. *(raised by czs007, chyezh)*

**`internal/datacoord/util.go:503` — per-field attribution is applied to the StorageV2 read path, which does not project columns, producing a large under-request.**
The premise documented at `util.go:503-506` ("for storage v3 the reader is built with a schema holding only the target column") holds only for the manifest (loon FFI) path. `GetFieldDatasFromStorageV2` builds `arrow_schema` and then never uses it, passing `nullptr` to `LoadWithStrategy` (`internal/core/src/storage/Util.cpp:1488-1490`, `1501-1508`); by contract a null schema means read all columns of the files (`internal/core/src/segcore/memory_planner.h:91`), and the code materializes the entire row-group table before selecting `column(col_offset)` (`Util.cpp:1520`). Real projection only happens on the manifest path via `GetFieldDatasFromManifest` (`Util.cpp:1600`, `1642`), which is chosen based on whether the request carries a manifest (`MemFileManagerImpl.cpp:234/271`, `DiskFileManagerImpl.cpp:970/985`; Go side `internal/datacoord/task_index.go:559`). StorageV2 is the *default* write version and has no `ManifestPath` — the streaming path defaults to `storage.StorageV2` and only upgrades to V3 for TEXT fields or `common.storage.useLoonFFI` (`internal/streamingnode/server/wal/interceptors/shard/shards/segment_alloc_worker.go:149-152`; non-streaming inference is equivalent at `internal/flushcommon/writebuffer/write_buffer.go:1384-1391`), and `useLoonFFI` defaults to `false` (`pkg/util/paramtable/component_param.go:1097-1103`). StorageV2 binlogs are multi-field column groups as well (`internal/flushcommon/syncmgr/pack_writer_v2.go:301`), so the `len(group.fields) > 1` branch is hit. Failure: a default-config StorageV2 segment where 40 `Int64` scalars share one remanent-short column group (~320 B/row) over 2M rows (~640 MB group), indexing one of those `Int64` fields → the estimate is `8B*2M = 16 MB`, landing in the `>10MB` bucket of `calculateIndexTaskSlot` for `max(16/16,1) = 1` slot (pre-PR: `(640MB/512MB)*16 = 16` slots); the scheduler admits on 1 slot while each task materializes the full 640 MB group, so 16 concurrent builds reach ~10 GB and OOM IndexNode. Before this change, `segment.getFieldBinlogSize` returned the column-group size, which matched what is actually read — the PR turns a correct estimate into an under-estimate here.
*Suggestion:* gate the per-field attribution on `ManifestPath != ""` (or `StorageVersion == StorageV3`) and keep the previous conservative sizing for StorageV2; alternatively, pass the projection schema into `LoadWithStrategy` and fix up the resulting column offset. *(raised by czs007, chyezh)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/99fd0ef58aa8a95cdfde883871d3bc5661eafa66...czs007:team2-fix/T77ee6919d8?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/99fd0ef58aa8a95cdfde883871d3bc5661eafa66...czs007:team2-fix/T1c0ff0ab29?expand=1

czs007:

This PR's per-field slot attribution is sound and bounded by the column group size on every path; the remaining items are a set of small, one-directional under-estimates plus one avoidable lookup on the recovery path.

## Low

**`internal/datacoord/index_inspector.go:317` — recovery path resolves the collection even when the estimator discards it**
`collections.get(i.ctx, segment.GetCollectionID())` is an argument expression, so it is evaluated unconditionally, before `estimateIndexFieldSize` applies its own gate — the first thing that function does is `if !supportsFieldProjection(segment) { return binlogSize }` (index_inspector.go:279-281), discarding `coll` for every StorageV1/StorageV2 / manifest-less segment. The submit path already has the gate on the outside (index_inspector.go:215-217); the recovery path does not. Failure: a datacoord restart on a StorageV1/V2 cluster with index tasks still pending for a collection rootcoord no longer has → `reloadFromMeta` blocks on the full `retry.Attempts(5)` (~3s of backoff, ctx-capped at 10s) per such collection, synchronously before `StateCode_Healthy` (server.go:414-417), and the resolved collection is thrown away one line later. Suggestion: mirror the submit path — resolve only when `supportsFieldProjection(segment)`. (raised by tinswzy)

**`internal/datacoord/stats_inspector.go:115` — stats recovery path resolves the collection for sub-jobs that never read it**
`coll := collections.get(si.ctx, segment.GetCollectionID())` runs before `estimateStatsTaskSize` decides whether the value is relevant. Inside, it is used only under `if coll != nil && supportsFieldProjection(segment)` and only when `statsTaskFieldIDs` returns a non-empty list (stats_inspector.go:471-472), and that function returns `nil` for every sub-job other than `JsonKeyIndexJob`/`TextIndexJob` (stats_inspector.go:575-577). Since `Sort` is by far the most common recovered stats sub-job, the majority of iterations resolve a collection whose value is discarded. Suggestion: gate the lookup on `supportsFieldProjection(segment)` and on the sub-job being `JsonKeyIndexJob`/`TextIndexJob`. (raised by tinswzy)

**`internal/datacoord/util.go:757` — external `measured` gate mirrors `SourceDataColumnName` but omits the sampler's `avgBytes > 0` guard**
The estimator decides a fixed-width external field contributed to the measured group size by asking `externalColumnResolver.SourceDataColumnName(field)` (util.go:759), but the producer of that measured size requires strictly more: `if avgBytes, ok := fieldSizes[columnName]; ok && avgBytes > 0` (internal/datanode/external/task_update.go:1160). That `> 0` matters for Bool: the sampler uses integer division, `sampled_sizes.emplace_back(column_name, col_bytes / num_rows)` (external_utils_c.cpp:306), and Arrow BOOL passes through `NormalizeExternalArrowByType` unconverted, so its bit-packed buffer gives `(num_rows/8)/num_rows == 0` — mapped external Bool columns are deterministically excluded from the group's `MemorySize`, yet the estimator still deducts their 1 B/row into `fixedPerRecord`. Failure: an external collection whose synthetic column group holds a mapped Bool alongside a JSON/VarChar column, then a json-key-index/text-index stats task or index build on that column → `residualPerRecord` is 1 B/row below the bytes the group actually contains, roughly a 10% under-estimate for a compact JSON column. Suggestion: only deduct fields actually present in the measured size, i.e. replicate the `> 0` condition. (raised by tinswzy)

**`internal/datacoord/util.go:596` — nullable fixed-width fields treated as exact without the validity bitmap**
`fieldSizePerRecord` marks a field `exact` purely from its data type (util.go:580-592) and takes its size from `typeutil.EstimateSizePerRecord`, which returns payload bytes only and never adds the `(rows+7)/8` validity bitmap a nullable column materializes (pkg/util/typeutil/schema.go:146-214 — `fs.GetNullable()` is never consulted). For a non-requested group member this is safe (the bitmap stays in the residual, which over-charges — conservative); for a requested nullable fixed-width field it under-charges by `numRows/8` bytes. Failure: a scalar index build on a nullable Int8/Int16 column sharing a StorageV3 column group, where the true materialized size sits within ~11% above one of `calculateIndexTaskSlot`'s 10MB/100MB/512MB boundaries → the estimate lands one bucket low and the task reserves fewer slots than the data it materializes. Suggestion: add the bitmap for nullable fields, or stop marking nullable fields exact. Magnitudes: ~12.5% for nullable Int8, ~1.5% for Int64, negligible for dense vectors; nullable Bool is over-charged instead, per the item above. (raised by tinswzy)

## Nitpick

**`internal/datacoord/util.go:773` — integer floor in the per-record residual discards up to `numRows-1` bytes**
`residualPerRecord := group.size/numRows - fixedPerRecord` floors the division and the result is multiplied back by `numRows` at util.go:789, so up to `numRows-1` bytes are silently dropped across the segment, always in the under-estimating direction. Suggestion: compute the residual in total bytes — `group.size - fixedPerRecord*numRows` — which removes the loss entirely. (raised by tinswzy)

**`internal/datacoord/util.go:666` — doc comment claims an "upper bound", and the PR description still describes the superseded proportional split**
(1) The `estimateFieldsReadSize` doc comment asserts "Charging it once keeps the result an upper bound, which is what a slot must be" (util.go:666-667). Given the validity-bitmap omission (util.go:596), the external-Bool over-deduction (util.go:757) and the integer floor (util.go:773), the value is not strictly an upper bound on bytes read; the invariant the code actually enforces is that every path stays bounded by the column group size (clamp at util.go:796-798, schema-guess floor at util.go:774-778). Suggestion: reword to "bounded by the column group size" so the stated guarantee matches the enforced one. (2) The PR body still says the remaining measured bytes are shared between variable-length fields "proportionally to their schema estimate", but commit 74bf9f5547 reversed that and the merged code charges the residual once via `residualCharged` (util.go:782-791). Suggestion: update the PR description to match. (raised by tinswzy)

czs007:

Verified:
- `internal/datacoord/util.go:738-765` — fixed-width external fields are subtracted from the variable-width residual only when `StorageColumnResolver.SourceDataColumnName` confirms they were actually sampled, or when they are explicit function outputs, using the same predicate as the write-side measurement in `internal/datanode/external/task_update.go:982-994,1144-1163`.
- `pkg/util/typeutil/schema.go:2973-2995` — `SourceDataColumnName` excludes RowID, Timestamp, VirtualPK, and function outputs, so unmeasured system/virtual fields no longer erode the residual; the milvus-table format is covered because `ValidateMilvusTableSchemaIdentity` requires `external_field` on every target field.
- `internal/datacoord/util.go:705-710` — `supportsFieldProjection` limits per-field attribution to StorageV3 segments with a manifest, and this gate matches the reader selection in `DiskFileManagerImpl.cpp:970-985` and `MemFileManagerImpl.cpp:250-272`, which only take the manifest path when one is present.
- `internal/core/src/storage/Util.cpp:1601-1644` — the manifest read path projects only the target column, while the StorageV2 path (`Util.cpp:1489-1521`) materializes the whole column group, so non-projecting segments correctly retain the full group/segment size via `internal/datacoord/index_inspector.go:214-217` and `internal/datacoord/stats_inspector.go:471`.
- `internal/datacoord/util.go:555` — all remaining two-argument `calculateIndexTaskSlot` call sites in `index_inspector_test.go` were migrated to the three-argument signature, with no leftover two-argument callers repo-wide.

## Re-review of `f5f604baa3b1`

This round's verification concentrated on confirming the closure of the two previously reported high-severity issues and auditing the incremental commit; the earlier low/nitpick findings below were not individually re-examined:

- **Not re-examined** — `internal/datacoord/index_inspector.go:317`: recovery path resolves the collection even when the estimator immediately discards it. Nobody looked at it this round — it stands.
- **Not re-examined** — `internal/datacoord/stats_inspector.go:115`: stats recovery path resolves the collection unconditionally, including for Sort/BM25 jobs that never read it. Nobody looked at it this round — it stands.
- **Not re-examined** — `internal/datacoord/util.go:757`: external `measured` gate mirrors `SourceDataColumnName` but omits the sampler's `avgBytes > 0` guard. Nobody looked at it this round — it stands.
- **Not re-examined** — `internal/datacoord/util.go:596`: nullable fixed-width fields treated as exact without accounting for the validity bitmap. Nobody looked at it this round — it stands.
- **Not re-examined** — `internal/datacoord/util.go:773`: integer floor in the per-record residual discards up to numRows-1 bytes. Nobody looked at it this round — it stands.
- **Not re-examined** — `internal/datacoord/util.go:666`: doc comment claims an "upper bound", and the PR description still describes the superseded proportional split. Nobody looked at it this round — it stands.

No new findings were raised in this round.

internal/datacoord/util.go
```diff
@@ -571,6 +573,270 @@ func calculateIndexTaskSlot(fieldSize, numRows int64, indexParams []*commonpb.Ke
...
+}
+
+func (c *collectionCache) get(ctx context.Context, collectionID int64) *collectionInfo {
+	if coll, ok := c.cached[collectionID]; ok {
+		return coll
+	}
+	coll := resolveCollection(ctx, c.handler, collectionID)
+	c.cached[collectionID] = coll
+	return coll
+}
```
czs007: `collectionCache`: I don't think this type needs to exist, and its main rationale no longer holds now that the coordinators are merged.

**1. The hit path is already memoized one layer down.** `ServerHandler.GetCollection` (handler.go:569-573) returns from `meta.GetCollection` when present, and on a miss `loadCollectionFromRootCoord` ends with `s.meta.AddCollection(collInfo)` (server.go:1207). So after the first successful resolution of a collection, every later call is a `ConcurrentMap` lookup — even without this cache the cost is already O(distinct collections), not O(recovered tasks).

**2. There is no RPC involved anymore.** `s.broker` wraps `s.mixCoord`, which is the `*mixCoordImpl` itself (mix_coord.go:159), and `mixCoordImpl.DescribeCollectionInternal` is a plain passthrough to `rootcoordServer` (mix_coord.go:541-543) — an in-process call into rootcoord's in-memory meta. rootcoord is also already `Healthy` by then: `rootcoordServer.Init()/Start()` complete before the errgroup starts datacoord (mix_coord.go:172-197, root_coord.go:746).

That leaves only the trailing "including of misses" half. That one is real: a collection that genuinely cannot be resolved (dropped, GC hasn't caught up) burns the `retry.Do(Attempts(5))` backoff (~3s wall clock) on every call, and `ErrCollectionNotFound` is deliberately kept retriable for exactly this path (merr/errors.go:73), so it can't be shortened at the retry layer. But it's a cost this PR introduces: before it, `reloadFromMeta` resolved no collection at all, and it runs synchronously inside `Start()`, before `UpdateStateCode(Healthy)`.

Suggestion — drop the type entirely:

- **`reloadFromMeta`** (index_inspector.go:306/317, stats_inspector.go:105/115): use `i.meta.GetCollection` / `si.mt.GetCollection` instead. Pure local map read — no retry, no backoff, nothing added to the startup critical path. On a cold cache it falls back to the segment/binlog size, which is exactly today's behavior, so no regression; and every task created after startup still gets the accurate estimate.
- **`createIndexForSegment`** (index_inspector.go:216): the collection was resolved two frames up — `canCreateIndexForSegment` (:179) calls `handler.GetCollection` and fails closed on error. Returning it from there and passing it down removes the second lookup, and the inconsistent nil handling between the two (defer-the-build vs fall-back-to-binlog-size).
- **`SubmitStatsTask`** (stats_inspector.go:372): all three callers already hold the `*collectionInfo` from `mt.GetCollections()` (:208, :248, :287), and have already computed the same predicate that `statsTaskFieldIDs` recomputes (:214-221 vs :436-441). Passing the collection (or the field IDs) down removes the lookup and the duplicated predicate, and closes a small window where a `DropCollection` between `GetCollections()` and `SubmitStatsTask` makes every segment pay the 3s backoff on every tick.

The estimation logic itself looks right to me — this is only about how it gets its schema.

xiaocai2333: Addressed in 8a470e7299 — `collectionCache` is gone, along with `resolveCollection`. Verified your premises before changing anything: `loadCollectionFromRootCoord` does end with `meta.AddCollection`, and the scheduler's `pickNode` assigns an over-sized request to the most-available node on a best-effort basis, so the cold-cache fallback on recovery degrades to the pre-PR conservative slots without starving the task.

What changed, per your three suggestions:

- **`reloadFromMeta`** (both inspectors): reads `meta.GetCollection` / `mt.GetCollection` only — a local map read, nothing on the startup critical path. On a cold cache the estimation falls back to the binlog/segment size. Tests pin both directions: a cached collection yields the per-column estimate, and the handler mocks carry **no** `GetCollection` expectation, so any resolution attempt during recovery fails the test.
- **`createIndexForSegment`**: `canCreateIndexForSegment` now returns the collection it resolved and it is passed down, removing the second lookup and the inconsistent nil handling (the submit path can no longer hit the fall-back-to-binlog-size branch with a nil collection).
- **`SubmitStatsTask`**: reads the local meta cache via `getCollection`, closing the DropCollection-race backoff window. I stopped short of changing the interface signature to pass the collection/field IDs down: `statsTaskFieldIDs` has to stay for the recovery path anyway (a recovered task only has its `SubJobType`), so the signature change would buy one local map read at the cost of touching the interface and its mocks.

`internal/datacoord` passes in full; net -57 lines.

czs007:

Adversarial review surfaced two low-severity correctness issues, both in the new field-size estimation logic this PR adds to `internal/datacoord/util.go`.

## Low

- **`internal/datacoord/util.go:773` — Integer division truncation can drop the group residual estimate into a lower size bucket.** `residualPerRecord := group.size/numRows - fixedPerRecord` truncates before being multiplied back at line 789, losing `group.size mod numRows` (up to ~`numRows` bytes). Failure scenario: a shared column group in a StorageV3 manifest segment with a very large row count (e.g. 10M rows, truncation loss up to ~10MB) whose true read size sits just above a `calculateIndexTaskSlot`/`calculateStatsTaskSlot` bucket boundary (e.g. 512MiB) → the truncated estimate falls into the lower bucket, the task requests fewer slots than its real read volume warrants (a 4× gap at the 512MiB boundary), and the scheduler over-admits concurrent tasks, causing excess reads/memory pressure on workers. Suggestion: compute the total residual bytes first and then amortize, instead of truncating per-record. *(raised by chyezh, bigsheeper)*

- **`internal/datacoord/util.go:580` — Exact estimation ignores the variable-length encoding overhead of nullable vector columns.** `isFixedWidthType` (util.go:580–592) classifies vectors as fixed-width exact by `DataType` alone, without checking `field.GetNullable()`. On the storage side, nullable `FloatVector`/`BinaryVector`/`Float16Vector`/`BFloat16Vector`/`Int8Vector` are encoded as `arrow::binary()` (32-bit offsets + validity; see `internal/core/src/storage/Util.cpp:756–827`), while `fieldSizePerRecord` relies on `typeutil.EstimateSizePerRecord` (`pkg/util/typeutil/schema.go:176–186`), which only counts the raw payload (e.g. `dim*4`). Failure scenario: building an index on a nullable vector field in a StorageV3 manifest segment, especially at small dims (e.g. dim=8 nullable BinaryVector, 4B offset ≥ 1B payload per row) → the field size is underestimated (~0.8% at dim=128, several-fold at very small dims), so tasks near a bucket boundary drop into a lower slot tier and under-reserve resources; conversely, when the column is a non-requested group member, the skewed `fixedPerRecord` pollutes residual attribution for the variable-length columns. Nullable vectors with many nulls also store far less payload than the full-size estimate, which can push `fixedPerRecord` above the measured average and trigger the residual-zeroing fallback. Suggestion: have `fieldSizePerRecord` return `exact=false` for `Nullable` fields. *(raised by chyezh, bigsheeper)*

xiaocai2333:
Both Low findings are addressed in ff63071b68.

**`util.go:773` — integer truncation of the residual**
Fixed as suggested: the residual is now computed in whole bytes (`residualBytes := group.size - fixedPerRecord*numRows`) and charged once, instead of truncating per record and multiplying back. New test `residual keeps sub-row remainder bytes` pins a group size that is not a multiple of the row count and asserts no byte is lost.

**`util.go:580` — nullable vectors treated as exact fixed width**
Confirmed the encoding claim against `CreateArrowSchema` (`internal/core/src/storage/Util.cpp`): nullable vectors are `arrow::binary()`, non-nullable are `fixed_size_binary`. Fixed slightly more broadly than suggested: `fieldSizePerRecord` reports `exact=false` for **every** nullable field, not only vectors — nullable scalars also store less than the full width for null rows, so their schema estimate is just as data dependent. Nullable fields now take their bytes from the measured residual, which covers the encoding overhead, the many-nulls case, and stops them from polluting `fixedPerRecord`. New test `nullable fixed width field is charged from the residual` pins both the residual attribution and that the non-nullable members of the group keep their exact budget.

`internal/datacoord` passes in full.

czs007:

Adversarial review found two issues that survived verification: a Boolean size-accounting mismatch introduced by this PR that biases the new residual-based read-size estimate low, and a minor inconsistency in how the rewritten stats-task estimator treats text-match vs. JSON-key index jobs.

## Medium

- **`internal/datacoord/util.go` (new `isFixedWidthType` / `estimateGroupFieldsSize`) — Bool columns are measured bit-packed but deducted at 1 byte/row, so `residualBytes` is under-estimated for variable-length columns.** The measured `group.size` comes from `ActualSizeInBytes` (`internal/storage/serde.go:1470-1476`), which sizes `arrow.BOOL` value buffers as `BytesForBits(length)` ≈ numRows/8; the deduction side lists `DataType_Bool` as exact and charges it via `typeutil.EstimateSizePerRecord` (`pkg/util/typeutil/schema.go:151-152`) at 1 byte/row. Each non-nullable Bool column therefore over-deducts ~0.875 byte/row, and any requested variable-length column in the same group (JSON key index, text-match VarChar, INVERTED on JSON) is under-counted, breaking the "upper bound" invariant the function's own comment declares. The external path is worse: `internal/core/src/segcore/external_utils_c.cpp:306` samples `col_bytes / num_rows` with integer division, so Bool ≈ 0, and `internal/datanode/external/task_update.go:1211` only sums `avgBytes > 0`, so Bool contributes 0 bytes to the synthesized `memorySize` while datacoord still deducts a full 1 byte/row for it. Failure: a StorageV3 segment whose column group holds ≥1 non-nullable Bool plus a requested variable-length column (or any external collection with an `external_field`-mapped Bool) → `estimateFieldsReadSize` comes in below the true column size, and near the 10MB/100MB/512MB buckets or on >512MB segments (`util.go:567-568,580-581` scale slots by size/512MB) the task slot is too small, so the scheduler admits more index/stats tasks than the datanode memory budget allows. The error is bounded (≤1 byte/row × Bool field count, ~64 B/row at the default `maxFieldNum=64`), hence medium. Suggestion: move `DataType_Bool` out of the exact set (treat it like nullable fields via the residual), or deduct it as `BytesForBits(numRows)`; if Bool is moved out, a Bool-targeted BITMAP/INVERTED build would be charged the whole residual — conservative but still capped by group size, so directionally safe. (raised by chyezh, bigsheeper)

## Low

- **`internal/datacoord/stats_inspector.go:488` — `estimateStatsTaskSize` applies `× 2` only to `StatsSubJob_JsonKeyIndexJob` and returns bare `readSize` for `TextIndexJob`, though both build and hold INVERTED indexes.** The text-index task also constructs an INVERTED index per `EnableMatch` field (`internal/datanode/index/task_stats.go:573-611`, `eg.Go` with no concurrency cap). The JSON-only `× 2` predates this PR, but the PR narrows the text-task base from the whole segment (`getSegmentSize()`) to the match columns' share, removing the implicit headroom the old whole-segment base provided. Suggestion: either apply the same multiplier to `TextIndexJob` or document why the asymmetry is intended; no measurements currently show index resident memory exceeding the source column, hence low. (provenance unclear — please confirm whether this PR introduces it) (raised by bigsheeper)

czs007:

This PR narrows per-field size estimates for index and stats task slot accounting; the estimator is conservative in the right direction, but it depends on in-memory binlog arrays that manifest-backed StorageV3 segments never persist, and the new recovery tests encode a state that cannot occur after a restart.

## Medium

**`internal/datacoord/util.go:637` — `segmentColumnGroups` reads `segment.GetBinlogs()`, which is empty for every manifest-backed StorageV3 segment after a DataCoord restart, so the optimization is inert for exactly the segments it targets.**
V3 segments do not persist their binlog arrays: `buildAlterSegmentsKvs` strips them (`kv_catalog.go:327`, `resetBinlogFields(cloned)`) and then skips the per-`FieldBinlog` KVs entirely for `isV3Segment` (`kv_catalog.go:357-358`; `isV3Segment` is exactly `ManifestPath != ""`, `kv_catalog/util.go:116-118`, a superset of `supportsFieldProjection` at `util.go:720-722`). On reload, `applyBinlogInfo` can only back-fill from those never-written KVs (`kv_catalog.go:233-235`), and nothing rebuilds them from the manifest. Result: `segmentColumnGroups` returns an empty map and `estimateFieldsReadSize` returns `WrapErrFieldNotFound` for every field (`util.go:698-700`), so both call sites fall back to the whole-segment size. Failure: an external collection is refreshed into StorageV3 segments carrying a `ManifestPath` and synthetic ~2GB `Binlogs`, and DataCoord restarts (rolling upgrade / OOM / pod reschedule) with index or stats tasks still pending → the reloaded `SegmentInfo.Binlogs` is empty, `reloadFromMeta` and every newly triggered `createIndexForSegment` / `SubmitStatsTask` hit the fallback, and `calculateIndexTaskSlot` again takes the `fieldSize > 512MB` branch (`util.go:571-573`) to request `(2GB/512MB) * IndexTaskSlotUsage` = 256 slots — the exact number this PR sets out to eliminate — permanently, until compaction rewrites the segment. The direction is safe (the fallback equals the pre-PR value), so this is not a correctness regression. Suggestion: either derive group sizes for manifest-backed segments from the persisted `SegmentInfo.Stats`, or short-circuit explicitly with `if len(segment.GetBinlogs()) == 0 { <conservative fallback> }` before calling the estimator and log it at Debug/RatedWarn. Either way, please scope the benefit claimed in the PR body to "effective within the flushing process's lifetime, lost after a restart". *(surfaced during verification)*

**`internal/datacoord/index_inspector_test.go:266` — the two new recovery tests construct a segment/cache state that cannot occur during recovery, turning a guaranteed post-restart fallback into a green test.**
`TestIndexInspector_ReloadFromMetaUsesCachedCollection` (`index_inspector_test.go:266`) and `TestReloadFromMetaEstimatesPerSubJobType` (`stats_inspector_test.go:583`) both hand-build a segment that carries `ManifestPath` *and* non-empty `Binlogs` (`index_inspector_test.go:325-334`, `stats_inspector_test.go:611-621`) — impossible for a recovered V3 segment per the persistence chain above — and additionally pre-seed the collection cache (`index_inspector_test.go:295`), whereas real recovery runs synchronously in `Start()` while the collection cache loads asynchronously (`server.go:646-651`), making a cold cache the common case. Failure: rebuild the segments the way `ListSegments` actually returns them (`ManifestPath` kept, `Binlogs` empty) or leave `m.collections` cold, and the assertions at `index_inspector_test.go:356-359` and `stats_inspector_test.go:656-657` flip from the narrowed value to the whole-segment fallback → the tests go red, showing the current green does not protect the `reloadFromMeta` path and will lead future changers to believe narrowing survives a restart. Suggestion: rewrite these to assert the real recovered behaviour (no `Binlogs`), and keep the narrowing assertions on the in-memory paths, `createIndexForSegment` / `SubmitStatsTask`. *(surfaced during verification)*

## Low

**`internal/datacoord/util.go:586` — `isFixedWidthType` charges `Bool` at 1 B/row, which does not match the arrow bit-packed accounting behind `group.size`.**
`group.size` traces back to `ActualSizeInBytes` (`serde.go:1470-1476`), where `arrow.BOOL` is accounted as `BytesForBits(length)` (~0.125 B/row per buffer), while `fieldSizePerRecord` uses `typeutil.EstimateSizePerRecord`, which charges Bool 1 B/row (`pkg/util/typeutil/schema.go:151-152`). Every other type in the fixed-width list matches on both sides; Bool is the only entry that over-deducts (~0.75–0.875 B/row), which shrinks `residualBytes = group.size - fixedPerRecord*numRows` (`util.go:786`) and therefore understates the budget of the variable-length columns in the same group. Failure: a StorageV3 short column group with 8 Bool columns plus one `enable_match` VarChar column over 10M rows → `residualBytes` comes out ~70MB low, and since FMINDEX slots are a continuous function of `fieldSize` (`util.go:556`, `estimateFMIndexBuildPeakBytes`) rather than bucketed, the under-estimate propagates directly into fewer requested slots (~0–1 under default parameters), i.e. slight over-admission. Bounded in magnitude, but it is the one place where the estimator leans non-conservative. Suggestion: drop `schemapb.DataType_Bool` from `isFixedWidthType` so it is charged out of the residual, or account for it at 1/8 B/row. *(surfaced during verification)*

**`internal/datacoord/stats_inspector.go:477` — the estimation-failure path uses an unrated `mlog.Warn` inside a periodic trigger loop.**
Combined with the empty-binlog behaviour above, this fires whenever the collection cache is warm (`coll != nil`), the segment is manifest-backed, and the sub-job is Text/JsonKeyIndex — not just the `AddField` / missing-`MemorySize` edge cases. `estimateStatsTaskSize` runs at `stats_inspector.go:375`, *before* the `AddStatsTask` dedup at `:393-395`, and `triggerTextStatsTask` has no per-round cap, so dedup does not suppress the log. Failure: a collection with N manifest-backed V3 segments pending text / json key index after a DataCoord restart → N warning lines every `dataCoord.taskCheckInterval` tick (default 60s, `component_param.go:7204-7207`) until those stats tasks finish; at thousands of segments the rate can bury concurrent real alerts and inflate log-ingestion cost. Suggestion: downgrade to Debug or switch to `RatedWarn`, matching the existing `mlog.RatedInfo` used for "stats task already exists" a few lines below. *(surfaced during verification)*

**`internal/datacoord/util.go:674` — the PR description still documents a proportional residual split that the final implementation does not do.**
The PR body's "What this PR does" states that remaining measured bytes are shared between variable-length fields "proportionally to their schema estimate", and "Points worth a careful review" #2 elaborates on the risk of that split being off for skewed columns. The shipped code does the opposite: the doc comment at `util.go:674-678` states the residual is *not* split proportionally, and `util.go:799-806` charges the full residual once per group to any requested variable-length field. The risk profiles are inverted — reviewers will validate a proportional algorithm that does not exist and assume individual columns may be under-estimated, when the implementation is an upper bound that can only over-estimate. Suggestion: sync the PR body to the implementation, and fix one further inaccuracy while there — the body says `GetFieldDatasFromStorageV2` builds a single-column arrow schema, but that function reads the whole row group and then picks by `col_offset` (`internal/core/src/storage/Util.cpp:1474-1489`); the projecting reader is `GetFieldDatasFromManifest`, as the comment at `util.go:660-663` correctly states. *(surfaced during verification)*

xiaocai2333:
All five findings are addressed in 7ce647d3b2 (branch also rebased onto latest master).

**Medium 1 — attribution inert for recovered manifest-backed segments**
Confirmed your persistence chain (`resetBinlogFields` + the `isV3Segment` skip in `buildAlterSegmentsKvs`, nothing rebuilds the arrays on reload). Took the explicit short-circuit option: `estimateIndexFieldSize` returns the conservative size at Debug level when a manifest-backed segment carries no binlogs, and `estimateStatsTaskSize` gates the narrowing on `len(segment.GetBinlogs()) > 0`. The fallback value was already the Stats-derived whole-segment size (`getFieldBinlogSize` falls through to `getSegmentSize()` when nothing matches), so the direction is unchanged. I stopped short of deriving group sizes from `SegmentInfo.Stats`: it only records the aggregate insert/stats/delta sizes, not per-group splits, so there is nothing finer to attribute from after a restart. The PR body's new "Scope and fallbacks" section states the limitation exactly as you phrased it — effective within the process lifetime, lost after a restart until the segment is rewritten.

**Medium 2 — recovery tests encode an impossible state**
`TestIndexInspector_ReloadFromMetaUsesCachedCollection` is deleted (its premise cannot occur), and the two remaining recovery tests now build segments the way `ListSegments` returns them: `ManifestPath` set, `Binlogs` empty, size on `Stats` only — and assert the whole-segment fallback. The narrowing assertions live on the in-memory paths as you suggested: `TestIndexInspector_CreateIndexForSegment_FMIndexUsesMemoryBasedSlots` for the index side and `TestEstimateStatsTaskSize` for the stats side.

**Low — Bool bit-packing**
Dropped `Bool` from `isFixedWidthType`, so it is charged out of the measured residual instead of over-deducting 1 B/row from it. New case `bool group member does not reduce the measured residual` pins a bit-packed bool sharing a group with a json column.

**Low — unrated Warn in the trigger loop**
Switched to `mlog.RatedWarn(1/s)`. The empty-binlog case no longer reaches that log at all (gated before estimation), so what remains rated is genuine estimation failures.

**Low — PR body describes a proportional split**
Body rewritten to match the implementation: full residual charged once per group as an upper bound, and the projection claim corrected (`GetFieldDatasFromManifest` projects; `GetFieldDatasFromStorageV2` materializes the whole group).

All touched suites (`TestIndexInspector*`, `TestStatsInspectorSuite`, `TestUtilSuite`) pass on the rebased branch; the full-package run locally only fails on the pre-existing minio/etcd-dependent tests (no local minio), left to the CI loop.

czs007:

Field-level size attribution added by this PR never fires for user fields on non-milvus-table external collections, so the headline slot savings do not materialize on the main external path, and the new tests use a schema shape production cannot produce, which hides it.

## High

**`internal/datacoord/util.go:629` — field-level attribution is permanently inert for user fields on non-milvus-table external collections**

`exact := isFixedWidthType(field.GetDataType()) && !field.GetNullable()` evaluates to `false` for *every* user field mapped via `external_field` on a non-milvus-table external collection. At create time, Pass 2 of `NormalizeAndValidateExternalCollectionSchema` (`pkg/util/typeutil/schema.go:2792-2797`) rewrites each non-system, non-function-output field to `Nullable=true` in place, before the schema is persisted (`internal/rootcoord/create_collection_task.go:191-194`, `internal/proxy/task.go:522`), so DataCoord caches the rewritten schema. milvus-table sources return early at `schema.go:2779-2781` and are unaffected.

Consequence: RowID / Timestamp / `__virtual_pk__` keep `Nullable=false` but report `measured=false` (`schema.go:3050-3052`), so they are excluded from `fixedPerRecord` (`util.go:783-788`); the pk/vector/JSON fields are all nullable, so `exact=false` and the code takes the `hasVariableField` branch → `fixedPerRecord == 0` → `residualBytes = group.size` (`util.go:800`). Since external segments carry a single synthetic column group from `buildFakeBinlogs` (`internal/datanode/external/task_update.go:1150-1172`, `FieldID=DefaultShortColumnGroupID`, all fields in `ChildFields`), the new estimate equals what the old `segment.getFieldBinlogSize` path returned via `ChildFields` matching (`internal/datacoord/segment_info.go:524-545`) — byte for byte.

Failure: a parquet external collection (128-dim FloatVector + JSON, both mapped through `external_field`) whose refresh produces a 2 GB synthetic column group triggers vector index build / `JsonKeyIndexJob` → `estimateIndexFieldSize` returns 2 GB, the task still requests `calculateIndexTaskSlot(2GB) = (2GB/512MB) * indexTaskSlotUsage` slots, external index-build concurrency stays suppressed, and the "1 slot" outcome promised in the PR body never appears.

This is still a sound upper bound, not a correctness regression, but the PR body's lead example ("128-dim vector + 2 GB synthetic group → 256 slots down to 1") does not hold on the main external path. Function-output fields are exempt from Pass 2, keep `Nullable=false`, and are explicitly counted as measured at `util.go:786`, so they do benefit; milvus-table external and internal StorageV3 also work as designed.

Suggestion: either relax the `exact` predicate for the external case (e.g. treat "nullable but the measured group demonstrably contains this column" separately), or narrow the claimed benefit in the PR body and table to function-output fields, milvus-table external, and internal V3. *(surfaced during verification)*

## Medium

**`internal/datacoord/index_inspector_test.go:1008` — external test fixtures use a schema shape non-milvus-table production cannot produce, masking the issue above**

The three external fixtures added by this PR set `ExternalField` but leave `Nullable` unset: `index_inspector_test.go:894` and `:898-903` (`TestIndexInspector_estimateIndexFieldSize`), `index_inspector_test.go:1056-1064` (`TestIndexInspector_CreateIndexForSegment_ExternalTaskSlot`), and `stats_inspector_test.go:997-1010` (`TestEstimateStatsTaskSize`). `NormalizeAndValidateExternalCollectionSchema` cannot emit that shape for a non-milvus-table source (Pass 2 forces `Nullable=true`); the stats fixture additionally sets `ExternalSource` without `ExternalSpec`, a combination rejected outright at `schema.go:2697-2701`.

As a result the tests exercise the `exact=true` branch (pk counted into `fixedPerRecord`, residual charged to the JSON/varchar columns), while non-milvus-table production takes the `fixedPerRecord=0`, residual-equals-whole-group branch.

Failure: running those fixtures through `typeutil.NormalizeAndValidateExternalCollectionSchema` first (or simply adding `Nullable: true` to pk/vec/json) and re-running `go test -tags dynamic,test -gcflags="all=-N -l" ./internal/datacoord/ -run 'TestIndexInspector_estimateIndexFieldSize|TestIndexInspector_CreateIndexForSegment_ExternalTaskSlot|TestEstimateStatsTaskSize'` → the assertions at `index_inspector_test.go:1110`, `stats_inspector_test.go:1064` and `:1071` fail, because the estimate falls back to the full 2 GB group / whole-segment size.

Suggestion: normalize the fixtures through `NormalizeAndValidateExternalCollectionSchema` (and supply the missing `ExternalSpec`) before asserting, so the tests reflect the real production schema shape — which will immediately surface the High above rather than hiding it behind a green run. *(surfaced during verification)*

xiaocai2333:
Both findings are addressed in c8113ae166 (rebased onto latest master). Verified every premise before changing anything: Pass 2 of `NormalizeAndValidateExternalCollectionSchema` does force `Nullable=true` on every mapped user field of a non-milvus-table source, `SourceDataColumnName` does exclude system/virtual fields, and with all fields nullable the old code did return the synthetic group size byte for byte.

**High — attribution inert for user fields on non-milvus-table external collections**
Fixed by splitting the two uses of a schema size that the old `exact` flag conflated:

- *Deduction* from the group budget still requires exact (fixed width, non-nullable) — deducting a nullable field's full width could overstate its real bytes and understate everyone else's, so that side is unchanged.
- *Charging* a requested field now has a middle tier: a **nullable fixed width** field is charged its schema width plus the nullable encoding overhead (nullable vectors are binary-encoded with 4-byte offsets, plus the validity bitmap — bounded at 8 B/row), capped by the measured residual and the group size. Null rows only shrink the stored data, so this is an upper bound.

With that, the headline case works on the real production shape: a nullable 128-dim vector in a 2 GB all-nullable synthetic group is charged `(512+8) * numRows` ≈ 1.5 MB → smallest slot bucket. What this deliberately does **not** rescue: a variable length target column (json/text stats) in an all-nullable group has no schema bound and nothing deductible, so it conservatively keeps the whole group. The PR body's bullet list and "Where it applies" table now state exactly that split (footnote ¹), instead of implying the narrowing applies uniformly.

**Medium — fixtures use a schema shape production cannot produce**
All estimation fixtures are now production-shaped: every `ExternalField` user field carries `Nullable: true`, and schemas with `ExternalSource` also carry an `ExternalSpec` (the two are validated as an atomic pair). The stats fixture that needed deductible fixed width columns became an internal StorageV3 schema — a legitimate shape for that scenario — and a new sub-test pins the non-milvus-table external fallback (`non milvus-table external collection keeps the whole segment`). New unit case `all nullable external group still attributes fixed width columns` pins the High directly: on the all-nullable shape, the vector gets its bounded width and the json column gets the whole group, so a regression on either side goes red.

Affected suites (`TestIndexInspector*`, `TestStatsInspectorSuite`, `TestUtilSuite`, `TestSubmitStats*`) pass on the rebased branch; full-package validation left to the CI loop as before.

czs007:

**Verified:**

- `internal/datacoord/util.go:629` — the new `case estimate.fixedWidth` branch charges `(sizePerRecord + nullableFixedWidthPadPerRow) * numRows` and is capped both by `residualBytes` and by the trailing `total > group.size → group.size` guard, so the estimate cannot underestimate a group.
- `pkg/util/typeutil/schema.go:2783-2796` — Pass 2 marks every non-system, non-function-output user field `Nullable=true` (milvus-table returns early), so indexed vector fields on external collections now reach the `fixedWidth` charging branch instead of falling through to whole-group residual billing.
- `internal/core/src/storage/Util.cpp:153-176` — nullable vectors are written via `arrow::BinaryBuilder` (4 B/row int32 offset) plus a validity bitmap (1/8 B/row) ≈ 4.125 B/row, confirming the 8 B/row pad is a genuine upper bound and errs high, never low.
- `internal/datanode/external/task_update.go:681,1204-1214` — the `MemorySize` producer uses the same `StorageColumnResolver.SourceDataColumnName` predicate as the deduction side in `estimateGroupFieldsSize`, so the "deduct only measured exact fields" rule is consistent across both ends.
- `internal/datacoord/meta.go:500-506` — after the `collectionCache` removal, both inspectors' `reloadFromMeta` read only the local concurrent-map `meta.GetCollection`, adding no synchronous rootcoord calls in `Start()`; new tests pin the "no `GetCollection` expectation" invariant.
- `pkg/util/typeutil/schema.go:3368` — `GetFieldByID(nil, …)` and `IsExternalCollection(nil)` are nil-safe, so an unvalidated `collectionInfo.Schema` degrades to `merr.WrapErrFieldNotFound` and a full-segment-size fallback rather than a panic.

PR #52006, opened 2026-07-29
fix: report a stats increment from schema bump materialization

tedxu:
issue: #51729

Schema-bump materialization appends function-output columns to an
existing StorageV3 segment. It reported an absolute Statistics computed
from the compaction plan's FieldBinlog arrays. Since #50410 those arrays
are empty for a V3 segment recovered after a DataCoord restart, so the
reported footprint counted only the newly written columns, and the
receiver adopted it verbatim and made the undercount durable. That value
feeds compaction triggering, residual size, load balancing and GC.

CompactionSegment.stats is now interpreted by adoption path rather than
extended with a new field. Where the receiver builds a new SegmentInfo
it remains the output's absolute footprint. Where it updates an existing
SegmentInfo in place it is an increment added onto that segment's
current Stats. Only schema-bump materialization produces an increment,
computed from data the compactor already holds, so it no longer reads
the plan's arrays at all.

The accumulation is gated on the forward-commit branch of the existing
manifest CAS, so a replayed result cannot double count. The increment
carries only insert_binlog_size, insert_binlog_count, stats_binlog_size
and null_counts. Delta and timestamp aggregates are absent from it and
left untouched by the receiver.

The result's insert_logs carries only the column groups this run wrote,
and the receiver merges them into SegmentInfo.Binlogs rather than
replacing the array, sharing the helper the V2 backfill path uses. The
merge is idempotent and deliberately ungated, since a post-restart
replay is exactly when the array is empty and most needs repopulating.
That array is the sole input to the index-eligibility gate, so the
result must keep carrying it.

Materialization now asserts it was not handed dropped fields, because
the increment is only valid for an add-only operation. Any drop routes
to runFullSchemaRewrite, which still ships an absolute Stats.

Verification, stated precisely

Verified locally: the DataCoord half, internal/datacoord and
internal/metastore/kv/datacoord. The post-restart accumulation test was
checked against a deliberately broken receiver and fails there, so it
covers the reported defect rather than merely passing.

NOT verified locally: internal/datanode/compactor could not be built on
the development machine, because its prebuilt C++ core is missing a
symbol that internal/util/initcore requires. Those changes were reviewed
statically against real symbol definitions but never compiled or
executed before CI. CI is their first execution.

No end-to-end reproduction of the original defect exists. The fix is
supported by a receiver-level simulation of the recovered-V3 shape and
by static tracing, not by reproducing the bug in a running cluster.

Deliberate metric semantics changes

DataCoordCompactedSegmentSize now observes only the bytes a
materialization wrote, instead of the whole segment's footprint.
DataNodeWriteDataCount{delete} no longer counts a segment's pre-existing
deletes as newly written by a materialization.

Not addressed

Segments whose Stats was already corrupted keep the wrong value until a
sort or compaction rewrites them. Repair would need byte sizes that are
precisely what is unavailable.

### DIFF
--- a/internal/datacoord/meta_test.go
+++ b/internal/datacoord/meta_test.go
@@ -2152,16 +2152,23 @@ func (suite *MetaBasicSuite) TestCompleteBumpSchemaVersionCompactionMutation() {
 		suite.EqualValues(3, infos[0].GetStats().GetDeleteNumRows())
 	})
 
-	suite.Run("materialization non-nil result stats overwrites old stats", func() {
-		// runMissingFunctionMaterialization grows Binlogs and ships a freshly
-		// computed Stats; the receiver must adopt it, not keep the pre-bump value.
+	suite.Run("in-place result stats is added onto existing stats", func() {
 		currentManifest := packed.MarshalManifestPath("/data/segments/1", 10)
 		resultManifest := packed.MarshalManifestPath("/data/segments/1", 12)
 		segs := makeSegments(1, commonpb.SegmentState_Flushed)
 		old := segs.GetSegment(1)
 		old.StorageVersion = storage.StorageV3
 		old.ManifestPath = currentManifest
-		old.Stats = &datapb.Statistics{InsertBinlogSize: 1234, InsertBinlogCount: 7}
+		old.Stats = &datapb.Statistics{
+			InsertBinlogSize:  1000,
+			InsertBinlogCount: 4,
+			StatsBinlogSize:   20,
+			DeleteNumRows:     3,
+			DeltaBinlogSize:   77,
+			TimestampFrom:     10,
+			TimestampTo:       50,
+			NullCounts:        map[int64]int64{100: 0},
+		}
 		m := &meta{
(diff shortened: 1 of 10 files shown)


internal/datacoord/meta_test.go
```diff
@@ -2180,16 +2187,181 @@ func (suite *MetaBasicSuite) TestCompleteBumpSchemaVersionCompactionMutation() {
...
+					BaseManifest:   currentManifest,
+					StorageVersion: storage.StorageV3,
+					Stats:          &datapb.Statistics{InsertBinlogSize: 1},
+				},
+			},
+		}
+		_, _, err := m.completeBumpSchemaVersionCompactionMutation(task, result)
+		suite.NoError(err)
+
+		fields := getSegmentBinlogFields(m.segments.GetSegment(1))
```
MrPresent-Han, line 2362: Please rebase this branch onto the latest master. PR #51768 (4a246f3c) removed getSegmentBinlogFields and changed index eligibility to the segment/collection schema-version gate. The current merge tree therefore leaves this call undefined, so the DataCoord tests will not compile. Please update or remove this test and the related obsolete binlog-gate comments after rebasing, then rerun CI.

Issue #52014, opened 2026-07-29
[Bug]: ArrayOfVector FP16/BF16 element search skips placeholder conversion


bigsheeper:
## Bug Report

FP32 search placeholders are not converted when the target field is an `ArrayOfVector` whose element type is `Float16Vector` or `BFloat16Vector`.

### What happens

Proxy resolves the searched struct sub-field and calls `ConvertPlaceholderGroup`, but `ConvertPlaceholderGroup` uses `fieldSchema.GetDataType()` only. For `ArrayOfVector` sub-fields this value is `ArrayOfVector`, so the existing fp32 -> fp16/bf16 conversion path is skipped and the placeholder is forwarded as `FloatVector`.

For element-level `ArrayOfVector` search with non-EmbList metrics, QueryNode/core expects the placeholder type to match the vector element type. The unchanged `FloatVector` placeholder is rejected with a vector type mismatch, so supported schemas fail unless clients pre-encode FP16/BF16 placeholders.

### Proof

On current master (`55b6f54617`), a focused proof test against `ConvertPlaceholderGroup` fails:

```bash
GOWORK=off go test -tags proof,dynamic,test -gcflags="all=-N -l" -count=1 \
  ./internal/proxy/vector_type_convert.go \
  ./internal/proxy/vector_type_convert_proof_test.go \
  -run TestProofConvertPlaceholderGroupArrayOfVectorElementType
```

Failure excerpt:

```text
expected: 102
actual  : 101
Test: TestProofConvertPlaceholderGroupArrayOfVectorElementType/float16_element

expected: 103
actual  : 101
Test: TestProofConvertPlaceholderGroupArrayOfVectorElementType/bfloat16_element
```

`101` is `FloatVector`, `102` is `Float16Vector`, and `103` is `BFloat16Vector`.

### Expected behavior

For `ArrayOfVector` fields, `ConvertPlaceholderGroup` should use the field `ElementType` for vector type compatibility and fp32 -> fp16/bf16 conversion, while preserving existing pass-through behavior for non-converted vector types.

PR #52015, opened 2026-07-29
fix: convert ArrayOfVector element placeholders

bigsheeper:
issue: #52014

## What changed

- Use `ArrayOfVector.ElementType` when proxy decides whether a search placeholder needs fp32 -> fp16/bf16 conversion.
- Add coverage for `ArrayOfVector` fields whose element type is `Float16Vector` or `BFloat16Vector`.

## Verification

- Red proof before fix:
  `GOWORK=off go test -tags proof,dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy/vector_type_convert.go ./internal/proxy/vector_type_convert_proof_test.go -run TestProofConvertPlaceholderGroupArrayOfVectorElementType`
  - Failed as expected: placeholder type stayed `FloatVector` instead of `Float16Vector`/`BFloat16Vector`.
- Green proof after fix:
  `GOWORK=off go test -tags proof,dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy/vector_type_convert.go ./internal/proxy/vector_type_convert_proof_test.go -run TestProofConvertPlaceholderGroupArrayOfVectorElementType`
- Focused package tests:
  `LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy -run 'TestConvertPlaceholderGroup|TestNormalizeFp32ToFp16Bf16VectorFieldData'`
- Static check:
  `LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage GOWORK=off make static-check`

### DIFF
--- a/internal/proxy/vector_type_convert.go
+++ b/internal/proxy/vector_type_convert.go
@@ -79,6 +79,9 @@ func ConvertPlaceholderGroup(phgBytes []byte, fieldSchema *schemapb.FieldSchema)
 	placeholder := phg.Placeholders[0]
 	phType := placeholder.Type
 	fieldType := fieldSchema.GetDataType()
+	if fieldType == schemapb.DataType_ArrayOfVector {
+		fieldType = fieldSchema.GetElementType()
+	}
 
 	// Check if types already match
 	if isVectorTypeMatch(placeholder.Type, fieldType) {
--- a/internal/proxy/vector_type_convert_test.go
+++ b/internal/proxy/vector_type_convert_test.go
@@ -226,6 +226,42 @@ func TestConvertPlaceholderGroupToBFloat16(t *testing.T) {
 	assert.Equal(t, 4*2, len(resultPhg.Placeholders[0].Values[0]))
 }
 
+func TestConvertPlaceholderGroupArrayOfVectorElementType(t *testing.T) {
+	for _, tt := range []struct {
+		name            string
+		elementType     schemapb.DataType
+		placeholderType commonpb.PlaceholderType
+	}{
+		{
+			name:            "float16 element",
+			elementType:     schemapb.DataType_Float16Vector,
+			placeholderType: commonpb.PlaceholderType_Float16Vector,
+		},
+		{
+			name:            "bfloat16 element",
+			elementType:     schemapb.DataType_BFloat16Vector,
+			placeholderType: commonpb.PlaceholderType_BFloat16Vector,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			vectors := [][]float32{{0.1, 0.2, 0.3, 0.4}}
+			phgBytes := createFloat32PlaceholderGroup(vectors)
+			fieldSchema := &schemapb.FieldSchema{
+				DataType:    schemapb.DataType_ArrayOfVector,
+				ElementType: tt.elementType,
+			}
+
+			convertedBytes, _, err := ConvertPlaceholderGroup(phgBytes, fieldSchema)
+			assert.NoError(t, err)
+
+			var resultPhg commonpb.PlaceholderGroup
+			err = proto.Unmarshal(convertedBytes, &resultPhg)
+			assert.NoError(t, err)
+			assertPlaceholderVectorsMatchFloat32(t, tt.placeholderType, tt.elementType, resultPhg.Placeholders, vectors)
+		})
+	}
+}
+
 func TestConvertPlaceholderGroupTypeMismatch(t *testing.T) {
 	// Test: Float16Vector searching BFloat16Vector field -> should error
 	// (Only fp32 -> fp16/bf16 conversion is supported)


czs007:

This PR's change to `ConvertPlaceholderGroup` is correct and minimal for the element-level `ArrayOfVector` path, and verification surfaced no issues introduced by this PR that require changes here.

Verified:
- `internal/proxy/vector_type_convert.go:82-84` — the `ElementType` override is gated to `DataType_ArrayOfVector`, so scalar `Array` fields (which also carry `ElementType`) are untouched.
- `internal/proxy/vector_type_convert.go:80,88` — the returned `phType` stays the pre-conversion `placeholder.Type`, so downstream element-level vs. emb-list routing at `task_search.go:1069` is unaffected.
- `internal/proxy/task_search.go:570,1054` — both the hybrid sub-request and ordinary search paths reach `ConvertPlaceholderGroup` with the full `FieldSchema`, so `GetElementType()` is populated.
- `internal/proxy/vector_type_convert.go:100-102` — a stale schema with unset `ElementType` yields `DataType_None` and falls through to pass-through, identical to pre-fix behavior and with no panic.
- `internal/core/src/query/Plan.cpp:69-116,137-168` — element-level `ArrayOfVector` searches require plain `Float16Vector`/`BFloat16Vector` placeholders, exactly the types this change emits.
- `internal/proxy/vector_type_convert_test.go:229-260` — both FP16 and BF16 `ArrayOfVector` element types are covered, including resulting placeholder type and converted bytes.

- `internal/proxy/vector_type_convert.go:93` — **`EmbList*` placeholder types are not converted for `ArrayOfVector` fields.** The five `EmbList*` types are absent from `placeholderTypeToDataType` (lines 46-53), so `isVectorTypeMatch` returns false and the `!isVectorType` guard at lines 93-95 returns the bytes untouched before the `fieldType = ElementType` value is ever consulted at line 97. Concretely: `ArrayOfVector(ElementType=Float16Vector)` + MaxSim metric + an fp32 `EmbListFloatVector` placeholder reaches segcore, where `check_data_type` demands `EmbListFloat16Vector` for a `VECTOR_FLOAT16` element (`Plan.cpp:79-86`) and the `AssertInfo` at `Plan.cpp:160-165` throws "vector type must be the same". This is the shape `ArrayOfVector` normally takes, since `task_index.go:367-372` maps the autoindex metric to an EmbList metric when the user specifies none. A fix would add the `EmbList*` entries to the map and target `EmbListFloat16Vector`/`EmbListBFloat16Vector` in `convertPlaceholder`; the byte-level fp32→fp16 conversion of the concatenated values already works unchanged. Unrelated to this PR: for an `EmbList*` placeholder the control flow is byte-identical before and after the change — `isVectorTypeMatch` misses the map either way and the same early return fires, so the fp32 emb-list query fails at the identical segcore assert without this PR. Recommend merging that separate fix before this PR if the emb-list path matters for the release. (raised by czs007, xiaocai2333)

- `internal/proxy/vector_type_convert.go:163` — **The insert/upsert path still rejects `ArrayOfVector`, leaving a write/search asymmetry.** `normalizeFP32ToFP16BF16VectorField` (called from `task_insert.go:251` and `task_upsert.go:1438`) bails out at lines 164-167 because an `ArrayOfVector` field's `DataType` is neither `Float16Vector` nor `BFloat16Vector`; insert then fails in `checkArrayOfVectorFieldData` with "need float16 vector array" (`validate_util.go:1256-1262`). The guard is not the only blocker: lines 168-170 read `vectors.GetFloatVector()`, which is nil for an `ArrayOfVector` field whose payload lives in `vectors.GetVectorArray()` (`validate_util.go:1177`), so a real fix must iterate the array elements and rewrite each element's oneof — a separate change, not a one-line guard tweak. Ordering is favorable at least: normalize (`task_insert.go:251`) runs before `Validate` (`task_insert.go:289`), so a fix there would take effect. Unrelated to this PR: `normalizeFP32ToFP16BF16VectorField`, `checkArrayOfVectorFieldData`, and both call sites are unmodified, and an fp32 write to an fp16 `ArrayOfVector` field was rejected at exactly the same line before this PR — only the asymmetry framing is new. (raised by xiaocai2333)

czs007 (approved):
/lgtm

PR #52019, opened 2026-07-30
enhance: [3.0] upgrade Woodpecker to v0.1.37 and expose direct read config

tinswzy:
pr: #51896 
update wp to v0.1.37


czs007:
> **Must-fix issues introduced by this PR: 3**
> **Merge recommendation:** Mergeable after the 3 must-fix items below are fixed in this PR.

## Overview
This PR wires up the Woodpecker DirectRead config correctly, but the cherry-pick dropped the implementation half of the `maxBatchEntries` / `maxBatchBytes` batch parameters while keeping the struct fields and their test, leaving two dead config knobs, a deterministically failing test, and no escape hatch for the v0.1.36 batching change.

## Critical

**`pkg/util/paramtable/service_param.go:725` — `AppendMaxBatchEntries` / `AppendMaxBatchBytes` are declared but never `Init`'d and never applied.**
The two new `ParamItem` fields (`service_param.go:725-726`) have no `ParamItem` definition + `.Init(base.mgr)` in `WoodpeckerConfig.Init` (that block only adds the three `DirectRead*` items, `service_param.go:843-868`), so their `Key` and `DefaultValue` are the empty string. `setCustomWpConfig` (`pkg/streaming/walimpls/impls/wp/builder.go:88-116`) jumps straight from `SegmentAppend.QueueSize` (line 95) to `DirectRead.Enabled` (line 99) and never writes `SegmentAppend.MaxBatchEntries` / `.MaxBatchBytes`. `configs/milvus.yaml:194-196` likewise lists only `queueSize` / `maxRetries`. A repo-wide `git grep AppendMaxBatch` hits only these two struct lines and the test file.
Failure: an operator sets `woodpecker.client.segmentAppend.maxBatchEntries=1` in `milvus.yaml`/etcd to disable client-side group commit → the setting is silently ignored (the key is never registered and never read), the client keeps Woodpecker's built-in 1000 / 2MB batching, and no warning is logged.
Comparing against upstream master (this PR's body references #51896), master has all three pieces: the `ParamItem` definitions + `Init` at `service_param.go:808-824`, the `builder.go:104-116` assignments (including an explicit `strings.TrimSpace(...) == "0"` branch, hence the extra `strings` import), and the `maxBatchEntries` / `maxBatchBytes` entries at `configs/milvus.yaml:197-198`. `builder_test.go` is byte-identical to master's — only the implementation was lost.
Suggestion — pick one:
- Restore all three pieces from master: keys `woodpecker.client.segmentAppend.maxBatchEntries` / `maxBatchBytes`, defaults `1000` / `2000000`, plus the `strings` import. Note that `maxBatchBytes` treats an explicit `"0"` as "no byte limit", so it cannot reuse the single `v > 0` guard used for the DirectRead params.
- Or drop the two struct fields and `TestSetCustomWpConfigBatchParams`, keeping this PR to the DirectRead scope its title describes.

*(raised by xiaocai2333, xaxys)*

**`pkg/streaming/walimpls/impls/wp/builder_test.go:15` — `TestSetCustomWpConfigBatchParams` fails deterministically (3 subtests red, 2 more pass vacuously).**
This is a direct consequence of the issue above, and it is not flaky. Because the `ParamItem`s are never `Init`'d, `params.WoodpeckerCfg.AppendMaxBatchEntries.Key` (line 15) and `.AppendMaxBatchBytes.Key` (line 16) are `""`, so the `params.Save("", …)` calls at lines 25-26 write to an empty key — `BaseTable.Save` (`pkg/util/paramtable/base_table.go:301-313`) does not validate the key and always returns `nil`, so `require.NoError` passes and hides the problem. Even with correct keys the assertions would still fail, since `setCustomWpConfig` never applies the fields.
Failure: CI or a local run of the `pkg/streaming/walimpls/impls/wp` package unit tests → the package is deterministically red, blocking the `go-ut` job on the 3.0 branch.
Measured at PR head `68903a4dc7` (`go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./streaming/walimpls/impls/wp/... -run TestSetCustomWpConfigBatchParams -v`, from the `pkg` module):
- FAIL `ValidValues` — line 41 expected 500, actual 1000; line 42 expected 1048576, actual 2000000
- FAIL `ExplicitZeroBytesMeansNoByteLimit` — line 48 expected 0, actual 2000000
- FAIL `EntriesOneDisablesBatching` — line 73 expected 1, actual 1000
- PASS `MilvusDefaults`, `InvalidValuesKeepWoodpeckerDefaults`, `ZeroEntriesKeepsWoodpeckerDefault`

The `actual` values 1000 / 2000000 are Woodpecker's built-in defaults (`woodpecker@v0.1.36/common/config/configuration.go:890-891`). Note that `InvalidValuesKeepWoodpeckerDefaults` and `ZeroEntriesKeepsWoodpeckerDefault` pass vacuously — the fields are never written, so "keeps the default" holds trivially; they are not real coverage and should be re-checked once the implementation is restored. `TestSetCustomWpConfigDirectReadParams` (all 5 subtests) passes, and `go vet` / `go build ./...` are clean, so this is purely missing implementation, not a build problem.
Suggestion: fix the implementation above, then re-run this test and tighten the two vacuous subtests so they fail if the override path regresses.

*(raised by xiaocai2333, xaxys)*

## High

**`pkg/streaming/walimpls/impls/wp/builder.go:95` — the bump to Woodpecker v0.1.36 enables client-side batched writes by default, and 3.0 has no config switch to turn it off.**
This PR moves Woodpecker from v0.1.33 to v0.1.36 (`go.mod:267`, `pkg/go.mod:40`). In v0.1.36, `NewSegmentHandle` builds a `BatchingSequentialExecutor` from `cfg.Woodpecker.Client.SegmentAppend.MaxBatchEntries` (default 1000, `segment_handle.go:104,117`), and any value `> 1` coalesces consecutive appends into a single `AddEntries` RPC (`sequential_executor.go:117`). v0.1.33's `logstore.proto` has no `AddEntries` method at all (only `AddEntry`, line 52; `AddEntries` is new at v0.1.36 `logstore.proto:57`), and `BatchAppendOp` has no `codes.Unimplemented` fallback — `batch_append_op.go:134-139` fails the whole node on any `AppendEntries` error, and grepping `Unimplemented` across `woodpecker@v0.1.36/woodpecker/` only hits `segment_compacted_notify_manager.go:333` (`CompactSegmentWithExpected` / `NotifySegmentCompacted`), not the append path. Combined with the issue above, there is no way to set `maxBatchEntries=1` on 3.0 — master has that escape hatch, this cherry-pick lost it.
Failure: a rolling upgrade in Woodpecker service storage mode where Milvus (embedding client v0.1.36) is upgraded before the Woodpecker logstore server (still v0.1.33), and any write queues more than one append → `AddEntries` returns `codes.Unimplemented`, `BatchAppendOp.failNode` fails the entire batch into `HandleAppendRequestFailure`, and once retries are exhausted WAL appends error out, so writes keep failing for the whole window with no config available to fall back to single-entry `AddEntry` — only a binary rollback.
Suggestion: restore the `maxBatchEntries` / `maxBatchBytes` paramtable + `builder.go` mapping together with the fix above (at minimum to bring back the escape hatch), and document the upgrade-order constraint in the PR description / release notes: upgrade the Woodpecker server to >= v0.1.34 before upgrading Milvus.

*(raised by xaxys)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/68903a4dc7405e31336ecc9ee8ab8a30c6832560...czs007:team2-fix/T78dd44cf62?expand=1

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/68903a4dc7405e31336ecc9ee8ab8a30c6832560...czs007:team2-fix/T4c21a9080a?expand=1

czs007 (approved):

Issue #52005, opened 2026-07-29
[Bug]: [Benchmark][Cluster] Bounded count(*) returns incomplete result after StreamingNode recovery


wangting0128:
- [x] I have searched the existing issues

### Environment

- Milvus version: `harbor.milvus.io/milvusdb/milvus:3.0-20260728-36456931`
- Deployment mode: StreamingNode cluster, Milvus Operator
- MQ: Pulsar WAL
- Storage: StorageV3/Vortex with loon FFI and S3-compatible object storage
- SDK: pymilvus `2.7.0rc175`
- OS/Kubernetes: 4AM cluster, namespace `qa-milvus`
- Deploy config: `zong-server-laion-sn`
- Collection: `laion_stable_10_nullable`
- Workload: `test_concurrent_locust_custom_parameters`
- Concurrent operations: Load, Bounded `count(*)` Query, Search, and Upsert

### Current Behavior

After a StreamingNode panics and restarts because a Vortex object-storage flush fails, Bounded `count(*)` queries first fail with `channel not available` and then may succeed while silently returning fewer rows than the expected collection count.

The observed storage trigger was:

```text
AwsErrorNoSuchUpload(code 101)
AWS Error NO_SUCH_UPLOAD during CompleteMultipartUpload operation:
The specified multipart upload does not exist.
```

The affected StreamingNode then panicked at `2026-07-29 09:38:30 UTC` and restarted at `09:40:13 UTC`.

During the recovery window, the client first received explicit channel-unavailable errors. After the channel was again serviceable, successful Bounded `count(*)` results were still incomplete.

Representative request:

- requestId / traceID: `dc275eafa52a4b4da6d05e900e29fdad`
- time: `2026-07-29 09:42:56 UTC`
- expected count: `100000000`
- returned count: `99999763`
- missing rows: `237`

The Proxy dispatched the request to both collection channels, `dml_0` and `dml_1`, and returned a successful response:

```text
[Proxy] [Query received]
[OutputFields="[count(*)]"]
[guarantee_ts=468010446628585472]

[Proxy] [get query result]
[channel=laion1b-test-2-rootcoord-dml_1_467422281854355718v1]

[Proxy] [get query result]
[channel=laion1b-test-2-rootcoord-dml_0_467422281854355718v0]

[Proxy] [Count result]
[count(*)=99999763]
```

The test checker reported several incorrect successful counts in the same recovery window:

```text
99999847 == 100000000
99999841 == 100000000
[...]

PR #52022, opened 2026-07-30
fix: keep latest delete snapshot timestamp monotonic

chyezh:
issue: #52005

- segcore delete MVCC: retain a monotonic maximum timestamp across load and stream delete batches before publishing the cumulative delete bitmap
- out-of-order delete history: append decreasing timestamp corrections per affected physical row in a rollback overlay for historical queries
- concurrent historical queries: keep rollback corrections append-only so readers cannot miss both the old and new visibility boundary during node replacement

### DIFF
--- a/internal/core/src/segcore/DeletedRecord.h
+++ b/internal/core/src/segcore/DeletedRecord.h
@@ -17,6 +17,7 @@
 #include <mutex>
 #include <shared_mutex>
 #include <tuple>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 #include <folly/ConcurrentSkipList.h>
@@ -79,7 +80,8 @@ class DeletedRecord {
         : insert_record_(insert_record),
           search_pk_func_(std::move(search_pk_func)),
           segment_id_(segment_id),
-          deleted_lists_(SortedDeleteList::createInstance()) {
+          deleted_lists_(SortedDeleteList::createInstance()),
+          rollback_delete_list_(SortedDeleteList::createInstance()) {
     }
 
     ~DeletedRecord() {
@@ -141,6 +143,7 @@ class DeletedRecord {
         Timestamp max_timestamp = 0;
 
         SortedDeleteList::Accessor accessor(deleted_lists_);
+        SortedDeleteList::Accessor rollback_accessor(rollback_delete_list_);
         for (size_t i = 0; i < pks.size(); ++i) {
             auto deleted_ts = timestamps[i];
             if (deleted_ts > max_timestamp) {
@@ -152,10 +155,6 @@ class DeletedRecord {
             timestamps,
             [&](const SegOffset offset, const Timestamp delete_ts) {
                 auto row_id = offset.get();
-                // if already deleted, no need to add new record
-                if (deleted_mask_.size() > row_id && deleted_mask_[row_id]) {
-                    return;
-                }
                 // Skip delete when delete_ts <= insert_ts.
                 // Normal segment callers search PKs with include_same_ts=false,
                 // so only rows with insert_ts < delete_ts reach this callback.
@@ -171,6 +170,31 @@ class DeletedRecord {
                 if (insert_ts != 0 && delete_ts <= insert_ts) {
                     return;
                 }
+
+                // The main delete list keeps the first-arriving delete for a
+                // physical row. If an out-of-order delete for the same row
+                // arrives later, preserve the minimum duplicate timestamp in
+                // a sparse overlay so historical queries can move the row's
+                // visibility boundary backward without expanding per-row
+                // storage for the whole segment.
+                if (deleted_mask_.size() > row_id && deleted_mask_[row_id]) {
+                    auto [it, inserted] =
+                        rollback_min_timestamps_.try_emplace(row_id, delete_ts);
+                    if (inserted) {
+                        rollback_accessor.insert(
+                            std::make_pair(delete_ts, row_id));
+                        mem_add += DELETE_PAIR_SIZE;
+                    } else if (delete_ts < it->second) {
+                        auto previous_ts = it->second;
+                        rollback_accessor.insert(
+                            std::make_pair(delete_ts, row_id));
+                        rollback_accessor.erase(
+                            std::make_pair(previous_ts, row_id));
+                        it->second = delete_ts;
+                    }
+                    return;
+                }
+
                 accessor.insert(std::make_pair(delete_ts, row_id));
                 if constexpr (is_sealed) {
                     Assert(deleted_mask_.size() > 0);
@@ -277,6 +301,20 @@ class DeletedRecord {
                 bitset.set(it->second);
             }
             it++;
+        }
+
+        // Historical snapshots are built from the first-arriving delete list.
+        // Always overlay late duplicate deletes on the slow path so a smaller
+        // timestamp for an already-deleted row is still visible without
+        // invalidating or rebuilding all dumped snapshots.
+        SortedDeleteList::Accessor rollback_accessor(rollback_delete_list_);
+        auto rollback_it = rollback_accessor.begin();
+        while (rollback_it != rollback_accessor.end() &&
+               rollback_it->first <= query_timestamp) {
+            if (rollback_it->second < insert_barrier) {
+                bitset.set(rollback_it->second);
+            }
+            rollback_it++;
         }
     }
 
@@ -430,6 +468,11 @@ class DeletedRecord {
         search_pk_func_;
     int64_t segment_id_{0};
     std::shared_ptr<SortedDeleteList> deleted_lists_;
+    // Sparse timestamp correction for physical rows whose duplicate deletes
+    // arrive out of order. The main list keeps the first-arriving timestamp;
+    // this overlay keeps the minimum timestamp among later arrivals.
+    std::shared_ptr<SortedDeleteList> rollback_delete_list_;
+    std::unordered_map<Offset, Timestamp> rollback_min_timestamps_;
     // max timestamp of deleted records which replayed in load process
     Timestamp max_load_timestamp_{0};
     // max timestamp of all delete batches processed by this record;
--- a/internal/core/src/segcore/DeletedRecord.h
+++ b/internal/core/src/segcore/DeletedRecord.h
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <limits>
 #include <memory>
 #include <mutex>
@@ -103,6 +104,7 @@ class DeletedRecord {
         }
 
         auto max_deleted_ts = InternalPush(pks, timestamps);
+        max_delete_timestamp_ = std::max(max_delete_timestamp_, max_deleted_ts);
 
         if (max_deleted_ts > max_load_timestamp_) {
             max_load_timestamp_ = max_deleted_ts;
@@ -120,9 +122,10 @@ class DeletedRecord {
         }
 
         auto max_ts = InternalPush(pks, timestamps);
+        max_delete_timestamp_ = std::max(max_delete_timestamp_, max_ts);
 
         if (ENABLE_LATEST_DELETE_SNAPSHOT_OPTIMIZATION.load()) {
-            UpdateLatestSnapshot(max_ts);
+            UpdateLatestSnapshot(max_delete_timestamp_);
         }
 
         bool can_dump = timestamps[0] >= max_load_timestamp_;
@@ -429,6 +432,9 @@ class DeletedRecord {
     std::shared_ptr<SortedDeleteList> deleted_lists_;
     // max timestamp of deleted records which replayed in load process
     Timestamp max_load_timestamp_{0};
+    // max timestamp of all delete batches processed by this record;
+    // a conservative upper bound for every delete in deleted_mask_
+    Timestamp max_delete_timestamp_{0};
     int32_t sealed_row_count_;
     // used to remove duplicated deleted records for fast access
     BitsetType deleted_mask_;
--- a/internal/core/src/segcore/DeletedRecord.h
+++ b/internal/core/src/segcore/DeletedRecord.h
@@ -173,10 +173,10 @@ class DeletedRecord {
 
                 // The main delete list keeps the first-arriving delete for a
                 // physical row. If an out-of-order delete for the same row
-                // arrives later, preserve the minimum duplicate timestamp in
-                // a sparse overlay so historical queries can move the row's
-                // visibility boundary backward without expanding per-row
-                // storage for the whole segment.
+                // arrives later, append each decreasing timestamp to a sparse
+                // overlay so historical queries can move the row's visibility
+                // boundary backward without expanding per-row storage for the
+                // whole segment.
                 if (deleted_mask_.size() > row_id && deleted_mask_[row_id]) {
                     auto [it, inserted] =
                         rollback_min_timestamps_.try_emplace(row_id, delete_ts);
@@ -185,12 +185,10 @@ class DeletedRecord {
                             std::make_pair(delete_ts, row_id));
                         mem_add += DELETE_PAIR_SIZE;
                     } else if (delete_ts < it->second) {
-                        auto previous_ts = it->second;
                         rollback_accessor.insert(
                             std::make_pair(delete_ts, row_id));
-                        rollback_accessor.erase(
-                            std::make_pair(previous_ts, row_id));
                         it->second = delete_ts;
+                        mem_add += DELETE_PAIR_SIZE;
                     }
                     return;
                 }
@@ -468,9 +466,12 @@ class DeletedRecord {
         search_pk_func_;
     int64_t segment_id_{0};
     std::shared_ptr<SortedDeleteList> deleted_lists_;
-    // Sparse timestamp correction for physical rows whose duplicate deletes
-    // arrive out of order. The main list keeps the first-arriving timestamp;
-    // this overlay keeps the minimum timestamp among later arrivals.
+    // Append-only timestamp corrections for physical rows whose duplicate
+    // deletes arrive out of order. The main list keeps the first-arriving
+    // timestamp, while rollback_min_timestamps_ tracks the current minimum so
+    // only corrections that move the visibility boundary backward are added.
+    // Older corrections stay in the list so a concurrent reader can observe
+    // either boundary without racing with node removal.
     std::shared_ptr<SortedDeleteList> rollback_delete_list_;
     std::unordered_map<Offset, Timestamp> rollback_min_timestamps_;
     // max timestamp of deleted records which replayed in load process
--- a/internal/core/src/segcore/DeletedRecordTest.cpp
+++ b/internal/core/src/segcore/DeletedRecordTest.cpp
@@ -46,6 +46,35 @@ using namespace milvus;
 using namespace milvus::segcore;
 using namespace milvus::exec;
 
+namespace {
+
+class LatestDeleteSnapshotOptimizationGuard {
+ public:
+    explicit LatestDeleteSnapshotOptimizationGuard(bool enabled)
+        : original_(
+              ENABLE_LATEST_DELETE_SNAPSHOT_OPTIMIZATION.exchange(enabled)) {
+    }
+
+    ~LatestDeleteSnapshotOptimizationGuard() {
+        ENABLE_LATEST_DELETE_SNAPSHOT_OPTIMIZATION.store(original_);
+    }
+
+ private:
+    bool original_;
+};
+
+void
+SearchDirectDeleteOffsets(
+    const std::vector<PkType>& pks,
+    const Timestamp* timestamps,
+    std::function<void(SegOffset offset, Timestamp ts)> cb) {
+    for (size_t i = 0; i < pks.size(); ++i) {
+        cb(SegOffset(std::get<int64_t>(pks[i])), timestamps[i]);
+    }
+}
+
+}  // namespace
+
 TEST(DeleteMVCC, common_case) {
     milvus::exec::expression::FunctionFactory& factory =
         milvus::exec::expression::FunctionFactory::Instance();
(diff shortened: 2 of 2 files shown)


xaxys:

Reviewed at head `f2a1bd1684`.

## Key issues

- **High — `LoadPush` can leave an already-published snapshot stale.** `internal/core/src/segcore/DeletedRecord.h:106`: after `InternalPush`, the new code only executes `max_delete_timestamp_ = std::max(max_delete_timestamp_, max_deleted_ts);`. It does not refresh or invalidate `latest_snapshot_`. Failure scenario: `StreamPush` deletes row A at timestamp 100 and publishes snapshot `{max_ts=100, A}`; a later `LoadPush` deletes row B at timestamp 200. A query at timestamp 200 still satisfies `query_timestamp >= snapshot->max_ts` and returns the stale `{A}` bitmap without scanning the delete list, making row B incorrectly visible. This ordering is possible because online deletes and later delta/L0 loading both update an existing segment.

## Fix suggestions

- After every successful `LoadPush`, either rebuild `latest_snapshot_` with `max_delete_timestamp_` when the optimization is enabled, or atomically invalidate it so queries use the slow path until a complete snapshot is rebuilt.
- Add the missing reverse-order regression test: `StreamPush` first, then `LoadPush`, followed by a query at or after the loaded delete timestamp. A segment-level test using `Delete` followed by `LoadDeltaData` would also cover the production call path.

chyezh:
Regarding the reported stale-snapshot sequence, I do not think `StreamPush(A@100) -> LoadPush(B@200) -> Query@200` is reachable for the native QueryNode path covered by #52005. The class-level sequence is constructible, but the production path adds tSafe and segment-publication constraints:

1. For an already-readable segment, `deleteNode.Operate` synchronously runs `ProcessDeleteBatches`, which forwards `worker.Delete -> LocalSegment.Delete -> DeletedRecord::StreamPush`, before calling `UpdateTSafe`. Query waits for tSafe and normally uses the returned tSafe as its MVCC timestamp. Therefore a query at MVCC 200 cannot run while a relevant online delete at 200 has only reached `LoadPush`; the corresponding `StreamPush@200` completes before tSafe reaches 200 and republishes the cumulative snapshot.

2. For a segment being loaded/recovered, L0/deltalog `LoadPush` happens before the segment is added to the readable distribution. `loadStreamDelete` catches up the delete buffer and publishes the distribution under the delete-buffer barrier. This is a fresh `DeletedRecord`, so there is no pre-existing `{max_ts=100, A}` snapshot to become stale before publication.

3. The `LoadScope_Delta` LeaderChecker path repairs a worker segment that is not yet present in the leader/readable view; the delta load completes before that distribution is published.

The production failure addressed by this PR is the opposite ordering: a future delete is already present in the cumulative bitmap, then an older `StreamPush` publishes that bitmap with a regressed batch timestamp. A query between the two timestamps is serviceable and incorrectly consumes the future delete.

If there is a production call path where `LoadPush` adds a previously-unapplied delete to an already-readable segment after tSafe has passed that delete timestamp, without either the normal `StreamPush` or load-time delete-buffer catch-up, please point to that path. That would make the reported reverse-order case actionable.

czs007:

## Overview

This PR's monotonic `max_delete_timestamp_` change is correct in its core invariant, but it turns one pre-existing out-of-order-delete scenario from accidentally-correct into a missed deletion, and it surfaces an adjacent hazard in the runtime toggle path.

## High

- **`internal/core/src/segcore/DeletedRecord.h:156` — Out-of-order deletes on the same PK are swallowed by the per-row dedup, and this PR turns that case from accidentally-correct into a missed deletion.**
  The `InternalPush` callback dedups by row (`if (deleted_mask_.size() > row_id && deleted_mask_[row_id]) return;`, L157-158), keeping only the *first-arriving* delete ts, while `max_timestamp` at L144-149 is computed over the whole batch and is unaffected by that dedup. So when the same PK is replayed as delete@200 first and delete@100 second, `(100, row)` never enters the SkipList, and this PR now pins `max_delete_timestamp_` monotonically at 200.
  Failure: on a sealed segment with `queryNode.segcore.enableLatestDeleteSnapshotOptimization` at its default `true`, PK X (row r) replayed as delete@200 via LoadPush/deltalog then delete@100 via StreamPush (out-of-order delivery across L0 segments / WAL replay), queried at mvcc ts=150 → row r is no longer marked deleted and is returned in the result set; because `(100, r)` never enters the SkipList, every query with ts ∈ [100, 200) keeps missing it, whereas before this PR the same sequence returned the correct result.
  Walking A/B: before, `UpdateLatestSnapshot(max_ts=100)` gave snapshot.max_ts=100, so `Query(ts=150)` satisfied `150 >= 100`, took the fast path, OR'd in `deleted_mask_` (which contains r), and r was correctly deleted. After, `UpdateLatestSnapshot(max_delete_timestamp_=200)` makes `Query(ts=150)` fall to the slow path with an empty `snapshots_`, where the SkipList holds only `(200, r)` and `200 > 150`, so r is not deleted.
  The dedup line itself is pre-existing and untouched by this diff — but the timing this PR's own new tests construct (LoadPush with a newer ts, then StreamPush with an older one; `DeletedRecordTest.cpp:208`) flips from correct to incorrect under a *same-PK* variant, and all three new tests use two distinct PKs (`DeletedRecordTest.cpp:215/220`, `243/248`, `271/277`), which happens to avoid this branch entirely.
  Suggestion: when deduping, still insert if the incoming ts is smaller (or lower that row's ts in the SkipList to the min), and add a same-PK regression case. *(raised by xaxys, xiaocai2333)*

- **`internal/core/src/segcore/DeletedRecord.h:127` — Disabling the optimization at runtime only skips snapshot publication; `Query` still consumes the stale snapshot unconditionally, so every new delete becomes invisible to the fast path.**
  Snapshot publication in `StreamPush` is flag-guarded (L127-129), but the `Query` fast path (L243-249) reads `latest_snapshot_` without checking the flag and nothing clears `latest_snapshot_` on disable; since the snapshot bitset is a clone of `deleted_mask_` taken at publish time (L224-227 → L64-66), deletes added after the flag flips can never reach it.
  Failure: a sealed segment has already published `snapshot(max_ts=100, bitset={A})`; an operator flips `queryNode.segcore.enableLatestDeleteSnapshotOptimization` from true to false via the dynamic callback (`internal/util/initcore/init_core.go:666` → `internal/core/src/common/Common.cpp:123`); the segment then receives delete B@200; any query at mvcc ts >= 100 → takes the fast path with the stale A-only bitset and returns B's row as live, and every subsequent delete on that segment stays invisible to the fast path until the flag is re-enabled and another StreamPush occurs.
  Why it is unrelated to this PR: the flag guard, the unconditional fast-path read, and the publish-time bitset clone are untouched by this diff — the only change to `StreamPush` is passing `max_delete_timestamp_` instead of `max_ts`, and in this sequence the single snapshot published before the flag flip has max_ts=100 either way, so the behavior is identical before and after.
  Worth noting for prioritization: disabling this flag is the most natural mitigation an operator would reach for against the bug this PR fixes, which lands them directly on this path. The new third test covers only disabled→enabled (`DeletedRecordTest.cpp:264-291`), not enabled→disabled. Suggested fix for the follow-up: check the flag on the `Query` fast path, or `std::atomic_store(&latest_snapshot_, nullptr)` when the flag transitions to false. No tracking issue yet; recommend merging that separate fix before this PR is relied on operationally. *(raised by xaxys, xiaocai2333)*

chyezh:
Addressed the same-row out-of-order delete case in commit `a621c6ea04`.

The fix keeps the first-arriving delete in the main delete list and preserves the minimum timestamp among later arrivals for an already-deleted physical row in a sparse rollback overlay. Historical slow-path queries merge this overlay after the normal snapshot/main-list scan, so `LoadPush(X@200) -> StreamPush(X@100) -> Query@150` now marks X deleted. The cumulative latest bitmap and its fast path are unchanged, and the unique deleted-row count is still one.

The insert-timestamp check now runs before duplicate handling, so the overlay cannot delete a physical row version before that version was inserted; growing-segment same-PK resurrection semantics remain intact.

Added `LatestSnapshotPreservesEarlierDeleteForSameRow`, covering query timestamps 50, 150, and 200. The new PR CI for `a621c6ea04` is currently running.

czs007:

## Overview

Commit 1 (`max_delete_timestamp_` monotonicity) traces clean, but commit 2's rollback overlay has a concurrency window on its `erase` path, materializes entries for duplicates that can never change a query result, and has no test coverage on the branch carrying both.

---

## High

**`internal/core/src/segcore/DeletedRecord.h:189` — insert-then-erase leaves a window where a concurrent query sees neither overlay entry**

On the duplicate-delete path, the code inserts `(delete_ts, row_id)` — which sorts *before* the existing entry, since `delete_ts < previous_ts` — and only then erases `(previous_ts, row_id)`. folly's `Accessor::erase` forwards to `remove()`, which unlinks the node via `preds[k]->setSkip(k, nodeToDelete->skip(k))` (`ConcurrentSkipList.h:392-393`), so a reader whose cursor already advanced past the new smaller key hops straight over the removed larger key and observes neither. `mask_with_delete` takes no lock on either segment implementation (`ChunkedSegmentSealedImpl.cpp:3596-3600`, `SegmentGrowingImpl.cpp:396-400`), so `Query()` is genuinely concurrent with `InternalPush`.

*Failure:* a row receives three out-of-order deletes (ts=500 lands in the main list, ts=300 creates the overlay entry, ts=100 triggers the insert+erase pair) while a search runs the slow path with `query_timestamp` in `[300, 500)` and its rollback cursor parked between the two keys → that search misses the delete bit entirely and returns the deleted row as a live hit, since the main-list entry's ts=500 also exceeds `query_timestamp`.

This is not a use-after-free — `Query()` holds its own `Accessor` at `DeletedRecord.h:310` and folly's `NodeRecycler` defers destruction while any accessor holds a ref — but it is a transient missed delete: a row correctly reported deleted at ts T is momentarily reported live at the same ts T. That is visibility going backwards, precisely what this PR exists to prevent.

*Suggestion:* drop the `erase` and keep only `it->second = delete_ts`. This closes the window at zero correctness cost — a stale larger-timestamp overlay entry can only re-set a bit that the smaller-timestamp entry already set earlier in the same scan. *(raised by czs007)*

---

## Medium

**`internal/core/src/segcore/DeletedRecord.h:180` — overlay is materialized for ordinary duplicates that do not lower the visibility boundary**

The `try_emplace` + unconditional `rollback_accessor.insert` at `:180-186` fires on the *first* duplicate delete of any already-deleted physical row, regardless of whether `delete_ts` is below that row's main-list timestamp. Before this PR the duplicate returned immediately at zero cost. The overlay is also invisible to accounting: `mem_add` adds `DELETE_PAIR_SIZE` but the `unordered_map` node is uncounted, and `size()` (`:433-437`) reports only `deleted_lists_`, which is what feeds the cachinglayer charge at `:214-229`.

*Failure:* an upsert-heavy segment where rows are re-deleted in normal increasing-timestamp order (as in the existing `DeleteMVCC.delete_exist_duplicate_pks`, where `pk(3)`/`pk(2)` at ts 9 re-hit rows already deleted at ts 5/6) → one skiplist node plus one map node retained per duplicate-deleted row, none of it reflected in `size()`, and every slow-path `Query()` pays an extra full overlay scan from `begin()` (`:310-318`) that dumped snapshots can never skip, unlike the main list which starts at the snapshot cursor (`:288-297`).

*Suggestion:* skip materialization when `delete_ts` does not lower the row's existing boundary — that removes the entire cost for the common in-order upsert case. If the overlay is kept, consider folding its size into `size()` so the memory charge is not under-reported. *(raised by czs007)*

**`internal/core/src/segcore/DeletedRecordTest.cpp:187` — no test reaches the overlay erase branch, and `size()` hides overlay growth**

Every `DeletedRecord` test in the repo stops at two deletes per physical row, so nothing exercises `DeletedRecord.h:187-194` — the only lines in this PR that call `erase`, and the lines carrying the concurrency window above. `LatestSnapshotPreservesEarlierDeleteForSameRow` (`:236-266`) is the only same-row case and does a single `LoadPush` + single `StreamPush`, so `try_emplace` returns `inserted=true` and the branch is never entered; `common_case` deletes `pk(1)` twice; `delete_exist_duplicate_pks` deletes each physical row at most twice; `perform`, `LatestSnapshotOptimizationBenchmark` and `SnapshotDumpProgress` delete each row exactly once. Separately, `size()` counts `deleted_lists_` only, so `ASSERT_EQ(2, delete_record.size())` at `:170` and `ASSERT_EQ(4/5, ...)` at `:455`/`:474` stay green while the overlay silently grows — the suite cannot observe overlay size or content in either direction.

*Suggestion:* add a test with three out-of-order deletes on one row (ts 500, then 300, then 100, asserting the query boundary lands at 100) to cover the erase branch, and expose an accessor or size view over `rollback_delete_list_` so overlay growth becomes observable. *(raised by czs007)*

chyezh:
Updated the PR at head `98866366ca` to address the High issue in https://github.com/milvus-io/milvus/pull/52022#issuecomment-5131775672.

The rollback skip list is now append-only: when a duplicate delete moves a physical row's boundary farther backward, the new `(timestamp, row)` pair is inserted and the previous correction is no longer erased. `rollback_min_timestamps_` still tracks the current minimum, so equal or larger later arrivals do not add more nodes. A concurrent reader can therefore observe the previous correction or the new correction without racing with node removal.

Each retained correction now increments `mem_size_`. Added `RollbackOverlayKeepsDecreasingDeleteHistory` with `500 -> 300 -> 100`; it verifies the unique deleted-row count remains one, the main delete plus both retained corrections account for three delete pairs, and the historical/latest query boundaries remain correct.

`size()` intentionally continues to count unique deleted physical rows because it backs `get_deleted_count()`; rollback history must not inflate that value. The separate Medium concern about first ordinary in-order duplicates being materialized is unchanged by this commit and should be reassessed as a performance/space trade-off in the next review.

No local C++ build or unit test was run under the feature-worktree policy; new PR CI has been triggered for this head.

czs007:

Adversarial review of the delete-snapshot / rollback-overlay change: the MVCC boundary logic is sound, but the rollback overlay allocates on every duplicate delete — a documented-as-common path — and is rescanned in full on every slow-path query.

## High

**`internal/core/src/segcore/DeletedRecord.h:180` — rollback overlay allocates for every duplicate delete, and the growth is invisible to cache accounting.**
In the `deleted_mask_[row_id]` branch, `rollback_min_timestamps_.try_emplace` has nothing to compare against, because the main list's timestamp for the row is not tracked. The *first* duplicate delete of a row is therefore always appended to `rollback_delete_list_`, even when its timestamp is >= the main-list entry and carries no information. This is the common case, not a corner case: the Go layer deliberately forwards already-applied deletes (`internal/querynodev2/segments/segment.go:974-978`), L0-forward batches are not timestamp-sorted (`segment.go:1093-1095`), and upsert workloads produce one delete record per upsert for the same pk, all forwarded to older sealed segments that still hold it. Failure: any sealed segment receiving a second delete for a pk it already deleted — an upsert re-deleting the same pk, or the delegator re-forwarding L0 deletes after a segment/delegator reload → `rollback_delete_list_` + `rollback_min_timestamps_` grow to ~1 skiplist node + 1 hash-map node per deleted row (≈2× the `DeletedRecord` footprint) with zero query benefit, and none of it reaches the caching layer, since the charge at `DeletedRecord.h:215-216` uses `size()`, which counts only `deleted_lists_` (`DeletedRecord.h:431-435`) — so querynode resident memory exceeds what the caching layer believes is loaded and eviction/OOM protection under-reacts. This also defeats the stated goal of the comment at `DeletedRecord.h:174-179`.
*Suggestion:* record the main list's timestamp for a row when it is first inserted — either seed `rollback_min_timestamps_[row_id]` at the `accessor.insert` site (line 196), or use a dense `std::vector<Timestamp>` sized to `sealed_row_count_` — and append to the overlay only when `delete_ts` is strictly smaller. *(raised by czs007, xaxys)*

## Medium

**`internal/core/src/segcore/DeletedRecord.h:308` — slow-path query rescans the whole rollback overlay from `begin()` with no snapshot/cursor skipping.**
The overlay walk starts at `begin()` on every slow-path query, unlike the main-list walk, which resumes from the last dumped snapshot (`DeletedRecord.h:286-295`). The scan is thus O(rows-with-duplicate-deletes) per query, while the main walk is bounded by the post-snapshot tail. Failure: any search/query on a segment with a non-empty overlay where `query_timestamp < latest_snapshot_->max_ts` — the normal case under Bounded consistency, where the guarantee timestamp is `now - common.gracefulTime` while deletes keep arriving → each query traverses the entire overlay skiplist from `begin()`, adding per-query latency proportional to the total number of rows with duplicate deletes rather than to the un-snapshotted tail; on a heavily-upserted segment this is a multi-hundred-thousand-node walk on the search hot path.
*Suggestion:* at minimum add a cheap "overlay is empty → skip" guard around lines 308-316, backed by an atomic counter so the `Accessor` is not even constructed; the High fix above additionally keeps the overlay genuinely sparse. *(raised by czs007)*

## Low

**`internal/querynodev2/segments/segment.go:974` — load-bearing comment asserts an idempotency invariant this PR removes.**
The comment states that "segcore `DeletedRecord::InternalPush` is idempotent on (PK, ts): duplicate deletes against already-deleted rows are discarded internally". After commit 2 (`285d52b86e`) that is false in both halves: the early `if (deleted_mask_[row_id]) return;` guard is gone, and a duplicate now allocates a rollback-overlay node — unconditionally on the first duplicate for a row, and again on every strictly-smaller timestamp. Pushing the identical (PK, ts) batch twice therefore grows memory, so it is not idempotent even for equal timestamps. The conclusion ("Do NOT add a ts-watermark skip here") is still correct and more important than before, but its stated rationale is now wrong, and it is the justification for a silent-data-loss-avoidance decision.
*Suggestion:* update this comment and the back-reference at `segment.go:1093-1095` ("segcore dedups at (PK, ts) level") to state the real reason duplicates must keep being forwarded. *(raised by czs007)*

**`internal/core/src/segcore/DeletedRecordTest.cpp:303` — missing test for a duplicate delete whose timestamp is greater than the main entry.**
The five new cases (lines 208, 236, 267, 303, 331) cover monotonic snapshot, same-row earlier delete, decreasing chain, non-regressing snapshot ts, and the OFF→ON toggle, but none pins the no-information case. The existing `mem_size()` assertion (`:284-285`) covers only the strictly-decreasing chain (500 → 300 → 100).
*Suggestion:* add `LoadPush(pk@100)` then `StreamPush(pk@500)` asserting `mem_size() == 1 * DELETE_PAIR_SIZE`. It fails today (the current code yields `2 * DELETE_PAIR_SIZE`, since `try_emplace` reports `inserted=true` and appends), and it is exactly the assertion that pins the High fix — worth landing together with it. *(raised by czs007)*

- **`internal/core/src/segcore/DeletedRecord.h:108`** — `LoadPush` never republishes `latest_snapshot_`. `UpdateLatestSnapshot` has exactly one call site, in `StreamPush` (`DeletedRecord.h:129-131`); `LoadPush` mutates `deleted_mask_` via `InternalPush` but never republishes. QueryCoord's LeaderChecker issues `LoadSegments` with `LoadScope_Delta` (`internal/querycoordv2/task/utils.go:145`), which reaches an already-loaded sealed segment (`internal/querynodev2/handlers.go:118-125`) and lands in `LoadDeletedRecord -> LoadPush` (`ChunkedSegmentSealedImpl.cpp:3213`) — on a segment that has normally already published a snapshot via streaming deletes. Failure: a serving sealed segment that published `latest_snapshot_` at `max_ts=T1` then receives a LeaderChecker delta load carrying genuinely new deletes → `deleted_mask_` gains the rows but the published bitset is stale, so every query with `query_timestamp >= T1` takes the fast path at lines 266-271 and ORs the stale bitset, returning deleted rows until the next `StreamPush` on that segment. This is the under-deletion mirror of the over-deletion the PR fixes. Unrelated to this PR: the PR only adds the `max_delete_timestamp_` bookkeeping line to `LoadPush` — `LoadPush` had no `UpdateLatestSnapshot` call before it either, and the stale window is identical on master.
- **`internal/core/src/segcore/DeletedRecord.h:265`** — the fast path does not check `ENABLE_LATEST_DELETE_SNAPSHOT_OPTIMIZATION`, so disabling the flag at runtime freezes a stale snapshot. The fast path fires purely on `latest_snapshot_` being set, while the only publisher (line 129-131) is gated on that flag, which is refreshable (`pkg/util/paramtable/component_param.go:3764`, `:5016-5023`) and wired to a live callback (`internal/util/initcore/init_core.go:666-673`). Failure: an operator flips `queryNode.segcore.enableLatestDeleteSnapshotOptimization` to false at runtime while deletes keep arriving → `UpdateLatestSnapshot` is skipped but the fast path still consumes the frozen snapshot, so queries with ts >= the frozen `max_ts` omit every delete applied after the flip; the operator's escape hatch makes correctness worse instead of falling back to the safe slow path. Possible follow-up: check the flag in the fast path, or clear `latest_snapshot_` in `SetEnableLatestDeleteSnapshotOptimization` (`internal/core/src/common/Common.cpp:121-126`). Unrelated to this PR: neither the fast-path guard nor the publisher's flag gate is touched by the diff — the only change on line 130 is the argument `max_ts` → `max_delete_timestamp_`.
- **`internal/core/src/segcore/DeletedRecord.h:215`** — the caching-layer charge passes a delete-entry *count* where `memory_bytes` is expected. `new_estimated_size = size()` returns `accessor.size()` (entry count), is compared against a 64KB threshold, and is handed to `ChargeLoadedResource({delta_size, 0})`, whose first field is `memory_bytes` (`cachinglayer/Manager.h:96-101`, `Utils.h:76-78`); the correct byte figure is already tracked in `mem_size_` (`DeletedRecord.h:437-440`). Failure: any sealed segment accumulating deletes with the caching layer enabled → memory is reported at ~1 byte per delete entry instead of >=16 bytes plus node overhead, and nothing is charged at all below 65536 entries, so `dlist_` accounting and the `cache_loaded_bytes` metric understate real resident memory. Unrelated to this PR: the whole `if constexpr (is_sealed)` block at lines 212-227 is untouched context in the diff, and the units mismatch exists identically on master. That said, any fix for the High finding has to touch these lines anyway — recommend merging that separate fix before this PR if you want them addressed together.

czs007:

Re-review of `98866366caca`: the seven findings we previously posted on this commit are adjudicated below, and this round reports two medium findings against the new append-only rollback overlay in `DeletedRecord.h` — its unreclaimed resident growth and its full-scan cost on slow-path queries.

## Re-review of `98866366caca`

- `internal/core/src/segcore/DeletedRecord.h:180` — *Rollback overlay allocates for every duplicate delete, and the growth is invisible to cache accounting* — **Retracted**: our earlier comment was wrong as stated — overlay inserts are gated by the per-row minimum map (a node is added only when a delete timestamp drops below that row's running minimum, not on every duplicate delete), and each node is accounted (`mem_add += DELETE_PAIR_SIZE`, accumulated into `mem_size_` and charged to the caching layer via `ChargeLoadedResource`, `DeletedRecord.h:186–227`). The surviving residual concern — superseded nodes are never reclaimed — is re-filed at medium below.
- `internal/core/src/segcore/DeletedRecord.h:108` — *LoadPush never republishes `latest_snapshot_`, leaving the fast path stale after a delta-load into a serving segment* — **Not re-examined**: nobody looked at it this round — it stands.
- `internal/core/src/segcore/DeletedRecord.h:308` — *Slow-path query rescans the whole rollback overlay from `begin()` with no snapshot/cursor skipping* — **Confirmed**: still present at head; restated below with its concrete failure impact.
- `internal/core/src/segcore/DeletedRecord.h:265` — *Fast path does not check `ENABLE_LATEST_DELETE_SNAPSHOT_OPTIMIZATION`, so disabling the flag at runtime freezes a stale snapshot* — **Not re-examined**: nobody looked at it this round — it stands.
- `internal/core/src/segcore/DeletedRecord.h:215` — *Caching-layer charge passes a delete-entry count where `memory_bytes` is expected* — **Not re-examined**: the unit passed at the `ChargeLoadedResource` call site was not re-checked this round — it stands.
- `internal/querynodev2/segments/segment.go:974` — *Load-bearing comment asserts an idempotency invariant this PR removes* — **Not re-examined**: nobody looked at it this round — it stands.
- `internal/core/src/segcore/DeletedRecordTest.cpp:303` — *Missing test for a duplicate delete whose timestamp is greater than the main entry* — **Not re-examined**: nobody looked at it this round — it stands.

## Medium

- **`internal/core/src/segcore/DeletedRecord.h:180` — Superseded rollback-overlay nodes are never reclaimed, so overlay memory grows monotonically under out-of-order duplicate deletes.** After the insert-then-erase race fix, the overlay is strictly append-only: each time a delete timestamp below a row's running minimum arrives, a new node is inserted and the node it supersedes stays in the list forever. When a sealed segment holds R physical versions of a primary key and duplicate deletes for it arrive out of order (L0 replay concatenates timestamps unsorted across L0 segments, `segment.go:1094`) → each new per-row minimum appends R nodes to `rollback_delete_list_` with no reclamation path, so resident overlay memory grows until the segment is released and the growing overlay also lengthens every slow-path scan. The growth is bounded: because the min-map admits only new per-row minima and arrivals are concatenations of ascending runs (datanode delete buffers are written in write-ahead-log order; L0 compaction only concatenates), the ceiling is roughly R × (ascending-run count + 1) rather than R × D, and every node is charged to the caching layer — hence medium. Suggestion: add a reader-safe reclamation path for superseded nodes as a follow-up, and consider seeding the min-map when the main-table entry is inserted so a first duplicate with a *larger* timestamp no longer inserts a redundant node via the unconditional `try_emplace` branch (weighing the map's own resident cost). *(raised by xaxys, tinswzy)*

- **`internal/core/src/segcore/DeletedRecord.h:308` — The rollback overlay is outside the snapshot/dump mechanism, so every slow-path query scans it in full.** Overlay merging starts at `begin()` on each slow-path `Query`; the `snapshots_`/`snap_next_pos_` positioning applies only to `deleted_lists_`, and `DumpSnapshot` never materializes overlay entries. Once the overlay has accumulated N nodes and queries take the slow path (historical/MVCC reads with a query timestamp below the latest snapshot's `max_ts`, or a continuous delete stream advancing `max_delete_timestamp_` so most queries miss the fast path) → each query linearly scans up to all N overlay entries with no snapshot acceleration, so per-query latency grows linearly with overlay size. Suggestion: fold overlay entries into `DumpSnapshot`, or maintain a per-snapshot cursor into the overlay so slow-path queries skip the already-materialized prefix. *(raised by tinswzy, xaxys)*

PR #52023, opened 2026-07-30
feat: impl array_length/array_contains in Match Faimly for recursively nested Array fields

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/51793

design doc: docs/design-docs/design_docs/20260724-recursive-array-struct-column-layout.md‎


internal/core/src/common/ArrayValue.h
buqian-zilliz, line 1: **[P3] Removed null/initialization assertions from ArrayValueView allow silent null dereference**

The hot-path accessors of `ArrayValueView` (`size()`, `is_null()`, `element_type()`, `type()`, `begin()`, `end()`, `child()`, `output_data()`) no longer call `AssertInitialized()`, and `array_at()` no longer calls `AssertNotNull()`. This removes per-element branches but turns previously-diagnosed misuse into undefined behavior: `type()`, `child()`, and `element_type()` dereference `type_`/`child_`, which are `nullptr` for a default-constructed view, and `array_at()` on a null view now reads `offsets()[row]`/`offsets()[row + 1]` without the former guard. The new comment only documents that a default-constructed view must be assigned before use and does not cover the null-view case. Since this non-owning view is now used broadly across the new ARRAY expression paths, consider keeping a cheap `is_null_` guard in `array_at()` (or asserting in callers) and documenting the null-view contract explicitly.

SpadeA-Tang: It's intended.

internal/core/src/exec/expression/BinaryArithOpEvalRangeExpr.cpp
```diff
@@ -49,7 +49,12 @@ PhyBinaryArithOpEvalRangeExpr::Eval(EvalCtx& context, VectorPtr& result) {
     auto input = context.get_offset_input();
     SetHasOffsetInput((input != nullptr));
     auto data_type = expr_->column_.data_type_;
-    if (expr_->column_.element_level_) {
+    bool is_nested_array = false;
```
buqian-zilliz: **[P3] Recompute the invariant `is_nested_array` via a schema snapshot on every `Eval` call**

`PhyBinaryArithOpEvalRangeExpr::Eval` now calls `segment_->get_schema_snapshot()` and then indexes it with `operator[]` on every evaluation to recompute `is_nested_array`. This flag is invariant for the life of the expression, yet the lookup is performed on the per-batch filter hot path. For sealed segments `get_schema_snapshot()` is an atomic load of the published state followed by a `Schema` `shared_ptr` copy. When `data_type == DataType::ARRAY`, `is_nested_array` is equivalent to `expr_->column_.element_type_ == DataType::ARRAY`, so this can be derived directly from the already-available column info (or cached once at expression construction) instead of fetching the schema snapshot each batch.

internal/core/src/common/ArrayValue.h
```diff
@@ -238,25 +233,21 @@ class ArrayValueView {
...
 
     ArrayOffset
     end() const {
-        AssertInitialized();
         return end_;
     }
 
     const Chunk&
     child() const {
-        AssertInitialized();
```
zhengbuqian, line 259: consider using assert? which does not impact release mode performance.

zhengbuqian (approved):

PR #49081, opened 2026-04-16
fix: add standalone migrate data timeout to prevent shutdown hang

chyezh:
issue: #48777

- standalone shutdown: wait 10 seconds before checking for another active QueryNode, then skip migration when none is available
- session discovery: retry etcd session lookups until a snapshot is available so transient failures do not bypass rolling-upgrade migration
- streaming shutdown: close the TimeTickSyncInspector during resource release so its background and async sync goroutines cannot outlive the StreamingNode
- configuration: export queryNode.standaloneMigrateDataTimeout with a 10s default

Issue #48777, opened 2026-04-07
[Bug]: Standalone shutdown hangs indefinitely — querynode "migrate data" retry loop does not respect shutdown context


jeffdgr8:
## Environment

- **Milvus Version**: `milvusdb/milvus:v2.6.13` (also reproduced on `v2.6.14`)
- **Deployment Mode**: standalone
- **MQ**: woodpecker
- **SDK**: pymilvus v2.6.11 (bug is server-side, SDK not involved)
- **OS**: Ubuntu (Docker on Linux, host kernel 6.x)

### Option B: Steps (for systemic / stability / timing issue)

1. **Deploy**: Single standalone container with etcd and MinIO via docker-compose:
   ```yaml
   services:
     etcd:
       image: quay.io/coreos/etcd:v3.5.18
       volumes:
         - milvus_etcd_data:/etcd
       environment:
         - ETCD_AUTO_COMPACTION_MODE=revision
         - ETCD_AUTO_COMPACTION_RETENTION=1000
         - ETCD_QUOTA_BACKEND_BYTES=4294967296
         - ETCD_SNAPSHOT_COUNT=50000
       command: etcd -advertise-client-urls=http://etcd:2379 -listen-client-urls http://0.0.0.0:2379 --data-dir /etcd
       healthcheck:
         test: ["CMD", "etcdctl", "endpoint", "health"]
         interval: 30s
         timeout: 20s
         retries: 3
       deploy:
         resources:
           limits:
             memory: 512M

     minio:
       image: minio/minio:latest
       ports:
         - "9000:9000"
         - "9001:9001"
       volumes:
         - minio_data:/data
       environment:
         MINIO_ROOT_USER: minioadmin
         MINIO_ROOT_PASSWORD: minioadmin
       command: server /data --console-address ":9001"
       healthcheck:
         test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
         interval: 30s
         timeout: 20s
         retries: 3
       deploy:
         resources:
           limits:
             memory: 512M

     milvus:
       image: milvusdb/milvus:v2.6.13
       command: ["milvus", "run", "standalone"]
       security_opt:
         - seccomp:unconfined
       ports:
         - "19530:19530"
         - "9091:9091"
       volumes:
         - milvus_data:/var/lib/milvus
       environment:
         ETCD_ENDPOINTS: etcd:2379
         MINIO_ADDRESS: minio:9000
[...]

PR #52026, opened 2026-07-30
fix: [cp 2.6] prevent standalone shutdown hang

chyezh:
issue: #48777
pr: #49081

- standalone shutdown: bound migrate waiting and stop when no active QueryNode can accept data
- session discovery: retry until a valid snapshot is available so rolling-upgrade migration is not bypassed
- streaming shutdown: close TimeTickSyncInspector during resource release
- configuration: expose queryNode.standaloneMigrateDataTimeout with a 10s default

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -581,6 +581,7 @@ queryNode:
     nqMergeRatio: 16 # Maximum ratio between merged total NQ and the smaller task NQ when grouping query node read tasks.
     maxDeadlineMergeGap: 50ms # Maximum allowed gap between task context deadlines when grouping query node read tasks. Tasks with only one deadline cannot be grouped. Set a negative duration to disable this check.
     topKMergeRatio: 20
+  standaloneMigrateDataTimeout: 10s # Duration string (e.g. 10s, 3m). In standalone mode, after this duration, the node stops waiting for data migration if no other active query node is available.
   search:
     enableResultZeroCopy: false # When true, delegator passes reduced SearchResultData directly instead of re-marshaling to SlicedBlob. Toggle at runtime for instant fallback.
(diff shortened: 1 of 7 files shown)


xaxys:

Reviewed at head `6ba74fbcdc`.

### Key issues

- **High — `internal/querynodev2/server.go:459-466,518-522`:** `hasOtherActiveQueryNode` uses `retry.Do(node.ctx, ...)` with `retry.AttemptAlways()` and an always-true retry predicate. If `GoingStop()` succeeds but etcd subsequently becomes unavailable, or a persistent malformed QueryNode session prevents `GetSessions` from succeeding, this call never returns. The existing graceful-stop timer cannot be selected while execution is blocked inside the helper, and `node.ctx` is only canceled after the migration loop. Shutdown therefore hangs until the outer component watchdog force-exits, which can take the default 1800 seconds and defeats this backport's purpose.

### Fix suggestions

- Pass a bounded context into `hasOtherActiveQueryNode`, limited by the existing graceful-stop deadline, and use that context for both `retry.Do` and `GetSessions`.
- Continue retrying transient lookup failures within that budget, but when the budget expires, follow the normal migration-timeout cleanup path instead of remaining in the helper indefinitely.
- Extend `internal/querynodev2/server_test.go:306-322` with a persistent-error case that verifies the helper or `Stop()` returns when its shutdown context expires; the current test only covers one transient failure followed by success.

I found no other substantive backport-specific issues. The `pkg/v3` to `pkg/v2` adaptation, 2.6 zap logger usage, TimeTick inspector close ordering, paramtable registration, and test API usage otherwise appear consistent with the 2.6 branch.

czs007 (approved):
/lgtm

PR #52027, opened 2026-07-30
fix: [cp 3.0] prevent standalone shutdown hang

chyezh:
issue: #48777
pr: #49081

- standalone shutdown: bound migrate waiting and stop when no active QueryNode can accept data
- session discovery: retry until a valid snapshot is available so rolling-upgrade migration is not bypassed
- streaming shutdown: close TimeTickSyncInspector during resource release
- configuration: expose queryNode.standaloneMigrateDataTimeout with a 10s default

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -629,6 +629,7 @@ queryNode:
     nqMergeRatio: 16 # Maximum ratio between merged total NQ and the smaller task NQ when grouping query node read tasks.
     maxDeadlineMergeGap: 50ms # Maximum allowed gap between task context deadlines when grouping query node read tasks. Tasks with only one deadline cannot be grouped. Set a negative duration to disable this check.
     topKMergeRatio: 20
+  standaloneMigrateDataTimeout: 10s # Duration string (e.g. 10s, 3m). In standalone mode, after this duration, the node stops waiting for data migration if no other active query node is available.
   search:
     enableResultZeroCopy: false # When true, delegator passes reduced SearchResultData directly instead of re-marshaling to SlicedBlob. Toggle at runtime for instant fallback.
   levelZeroForwardPolicy: FilterByBF # delegator level zero deletion forward policy, possible option["FilterByBF", "RemoteLoad"]
--- a/internal/querynodev2/server.go
+++ b/internal/querynodev2/server.go
@@ -78,6 +78,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/metricsinfo"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
+	"github.com/milvus-io/milvus/pkg/v3/util/retry"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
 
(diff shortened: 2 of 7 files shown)


xiaocai2333:
🤖 **team2 automated review** (single agent · not cross-verified, for reference) · by `xiaocai`

Reviewed at head 417d498ae3 against base 3.0, comparing the net diff to master PR #49081. **This is a clean backport — mergeable, no blocking issues.**

The added/removed lines across all 7 files are byte-for-byte identical to master #49081; the only differences are baseline context line numbers (e.g. `configs/milvus.yaml` hunk at 629 vs 635, `component_param.go` at 3622/4723 vs 3679/4780), which is expected since the surrounding 3.0 code is offset. I verified each of the four logical pieces:

1. **Standalone migrate timeout / session retry** (`internal/querynodev2/server.go:471-531`) — `hasOtherActiveQueryNode()` and the standalone deadline check in `Stop()` are identical, including the `retry.AttemptAlways()` + `retry.Sleep(time.Second)` + `retry.RetryErr(always)` semantics that keep transient session-lookup failures from being misread as "no active node" and bypassing rolling-upgrade migration. The 3.0 branch already uses `mlog`, and the `Warn` call, `lo.Map` over `sealedSegments`/`growingSegments`, and `channelNum` all match — the patch applied without conflict, confirming those symbols exist on 3.0.
2. **TimeTick inspector close** (`internal/streamingnode/server/resource/resource.go:89`) — `r.timeTickInspector.Close()` is added first in `Release()`, before `wbMgr.Stop()` / `syncMgr.Close()`. Ordering is sound: the inspector is the producer that triggers syncs, so stopping it before the sync manager it drives is correct (no new syncs enqueued into an already-stopped consumer). `Close()` = `taskNotifier.Cancel()` + `BlockUntilFinish()` + `wg.Wait()`, so background + asyncSync goroutines are fully drained.
3. **Config / paramtable compatibility** — `StandaloneMigrateDataTimeout` field, `Init` block, `configs/milvus.yaml` entry, and 10s default all match master. The `Version: "2.6.16"` value is **correct for this branch**: the 3.0 line tracks 2.6.x param versioning (params up to 2.6.19 exist in this file), so it is consistent, not a stale copy from master.
4. **Regression tests** — `TestHasOtherActiveQueryNode`, `TestStopStandaloneMigrateTimeout`, and `TestReleaseClosesTimeTickInspector` are carried over verbatim, exercising: no-other-node → false, active-other → true, stopping-other → false, ctx-cancel → error (not "no node"), transient-error-retry via mockey, standalone fast-exit under a 0s timeout, and goroutine-count drop after `Release()`.

Minor observations (non-blocking, all inherited from master — no need to diverge from the upstream patch):

- **nit** — `internal/streamingnode/server/resource/resource_test.go:288` (`TestReleaseClosesTimeTickInspector`): the inspector is closed twice — once via `t.Cleanup(inspector.Close)` and again inside `Release()`. This is harmless because `Close()` is idempotent (`taskNotifier.Cancel()` on an already-cancelled context and `wg.Wait()` on a drained group are safe), but the `t.Cleanup` is effectively redundant.
- **nit** — `server_test.go` builds the "other" session via a hand-written JSON literal (`{"ServerID":99999,...,"Stopping":false}`) written directly to etcd rather than through `sessionutil`. It works for the current `Session` shape but is coupled to the on-disk JSON field names; if `Session` fields are ever renamed this test won't fail-fast at compile time. Acceptable for a backport.

Note: per the submitter, local tests were skipped. The change is logic-identical to an already-merged, reviewed master PR, so risk is low; CI on this PR should still be green before merge. No code changes requested.

czs007 (approved):
/lgtm

PR #52036, opened 2026-07-30
enhance: compute column size estimates on demand

weiliu1031:
issue: #51652

Follow-up to [#51653](https://github.com/milvus-io/milvus/pull/51653). Depends on [milvus-storage#605](https://github.com/milvus-io/milvus-storage/pull/605).

### What changed

- Bump milvus-storage to compute per-column chunk estimates on demand instead of retaining per-column metadata for every chunk.
- Consume bulk estimates once and map projected columns by logical column-group order.
- Prefer logical-column totals for full projections while keeping aggregate estimates as a compatibility fallback.
- Update Storage V2 and JSON stats call sites and tests for the new interface.

### Why

The initial projected-size implementation retained a column-size vector per row group and a field-name map per chunk. In wide column groups and lazy per-field readers, this metadata scaled with both chunk and column counts.

### Behavior boundary

Formats without per-column estimates continue to use the existing sampled, aggregate, or last-resort fallback paths. This PR does not change chunk decoding behavior.

### Validation

- `git diff --check`
- Unit tests not run.


internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -152,80 +152,143 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
         projected_estimate_error = "projection contains no columns";
     }
 
-    if (projected_estimate_available && full_projection) {
-        // Full projections cover the whole column group. Use the aggregate
-        // estimate directly to avoid a linear name scan for every projected
-        // column in wide groups such as JSON shredding output.
+    if (projected_estimate_available) {
+        auto column_size_result =
+            chunk_reader_->get_chunk_column_estimated_size();
```
buqian-zilliz: **[P2] Share bulk estimates across lazy translators**

`get_chunk_column_estimated_size()` returns the complete `[column][chunk]` matrix, but lazy storage-v2 and JSON-stats loading creates one `ManifestGroupTranslator` per projected field. With `L` lazy fields in an `N`-column group containing `R` row groups, this recomputes and transiently allocates the `N × R` matrix `L` times, becoming `O(N² × R)` when the whole group is lazy; some storage-v2 paths also construct these translators concurrently. Compute the bulk estimates once per column group and share/index them across projected translators, or expose a projection-specific estimate API.

internal/core/thirdparty/milvus-storage/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( milvus-storage_VERSION f9f5bd1)
-set( GIT_REPOSITORY  "https://github.com/milvus-io/milvus-storage.git")
+set( milvus-storage_VERSION 2cd7257)
+set( GIT_REPOSITORY  "https://github.com/jiaqizho/milvus-storage.git")
```
buqian-zilliz: **[P2] Do not fetch a build dependency from a personal fork**

The build now fetches milvus-storage from a contributor-owned fork. The pinned commit is the head of the still-open upstream dependency PR, so a fork deletion, visibility change, or force-push that makes the object unreachable would break clean Milvus builds and leaves a production build input outside the organization-owned repository. Merge the dependency change first, then restore `milvus-io/milvus-storage` and pin the merged official commit.

internal/core/src/segcore/storagev2translator/ManifestGroupTranslator.cpp
```diff
@@ -152,80 +164,150 @@ ManifestGroupTranslator::ManifestGroupTranslator(
...
         projected_estimate_error = "projection contains no columns";
     }
 
-    if (projected_estimate_available && full_projection) {
-        // Full projections cover the whole column group. Use the aggregate
-        // estimate directly to avoid a linear name scan for every projected
-        // column in wide groups such as JSON shredding output.
+    if (projected_estimate_available) {
+        if (!column_size_estimate.has_value()) {
+            column_size_estimate = FetchColumnSizeEstimates(*chunk_reader_);
```
sparknack, line 177: Since `column_size_estimate` also originates from `FetchColumnSizeEstimates()`, could this check and subsequent refetch be redundant?

weiliu1031: Good question. This is not a subsequent refetch: `nullopt` means the caller has not attempted to fetch the estimate.

Batch/lazy paths precompute the estimate once and share the same result across multiple translators. Standalone/eager paths create only one translator and let it perform the initial fetch. An engaged optional also represents a failed prefetch, preventing the translator from repeating the same failed operation.

The optional fallback mainly keeps standalone construction self-contained; we could alternatively require every caller to provide a precomputed result and remove this fallback.

Issue #52038, opened 2026-07-30
[Bug]: Credential material can be written to Milvus logs


XuanYang-cn:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Environment

- Milvus version: current `master`
- Deployment mode: standalone or distributed, depending on the triggered path
- Relevant log settings: debug logging or detailed request trace logging; some affected paths log at info or warning level

## Current Behavior

Several Milvus code paths serialize credential-bearing objects or request values directly into logs. Depending on the operation, the logged data can include plaintext or reversibly encoded passwords, API keys, password verifiers, and external-storage credentials.

| Trigger | Credential material that can reach logs |
|---|---|
| Authentication cache miss or API-key verification failure | API keys, bcrypt password hashes, and SHA-256 password verifiers |
| REST user creation or password update with detailed request tracing enabled | Base64-encoded plaintext passwords, including old and new passwords |
| A request hook returns an error for a credential operation | The complete credential request, including password fields |
| External-collection creation or refresh | Credentials in `external_source` URI components and `external_spec`, including object-storage access keys and secret keys |
| Stats-task pre-execution | The complete storage configuration, including secret access keys and GCP credential JSON |

These values can be retained in local logs, centralized log systems, support bundles, or log backups and become accessible to identities that are permitted to read operational logs but should not receive credentials.

## Expected Behavior

Milvus logs must not contain plaintext, reversibly encoded, hashed, or otherwise reusable credential material. Operational logs may identify the operation, account, task, non-sensitive storage type, outcome, and error without serializing credential values or complete credential-bearing objects.

## Steps To Reproduce

Use unique sentinel values rather than real credentials.

1. Configure Milvus with debug logging or detailed request trace logging where required by the selected path.
2. Perform one of the affected operations listed above using sentinel credential values.
3. Trigger the corresponding logging condition, such as an authentication cache miss, hook error, external-collection request, or stats task.
4. Search the Proxy or DataNode logs for the sentinel values.

## Actual Behavior

One or more sentinel credential values appear verbatim or in a reversible representation in the generated logs.

## Milvus Log

Sanitized examples of the affected output shapes:

```text
credential={username:<user> encrypted_password:<bcrypt-verifier> sha256_password:<sha256-verifier>}
request={password:<base64-encoded-plaintext> oldPassword:<base64-encoded-plaintext> newPassword:<base64-encoded-plaintext>}
[...]

PR #52039, opened 2026-07-30
fix: prevent sensitive credentials from reaching logs

XuanYang-cn:
Fail closed when redacting external source URIs and project external specs to approved metadata with the complete extfs object redacted. Preserve detailed validation errors.

Log API-key hook failures without the token, redact RBAC password hashes, remove TiKV/etcd values and DataNode storage access-key fields from logs, and stop plaintext Go-client test password logs. Add regression coverage for the affected paths.

See also: #52038


internal/datanode/index/task_stats.go
```diff
@@ -182,7 +182,7 @@ func (st *statsTask) PreExecute(ctx context.Context) error {
 		mlog.FieldSegmentID(st.req.GetSegmentID()),
 		mlog.Int64("storageVersion", st.req.GetStorageVersion()),
 		mlog.Int64("preExecuteRecordSpan(ms)", preExecuteRecordSpan.Milliseconds()),
-		mlog.Any("storageConfig", st.req.StorageConfig),
+		mlog.String("storageType", st.req.GetStorageConfig().GetStorageType()),
```
XuanYang-cn: What about other configs? Not all the configs ar credentials right? Maybe redact the credentials part?

internal/datanode/index/task_stats_test.go
```diff
@@ -187,6 +218,38 @@ func (s *TaskStatsSuite) TestSortSegmentWithBM25() {
...
+	defer cancel()
+
+	manager := NewTaskManager(ctx)
+	task := NewStatsTask(ctx, cancel, &workerpb.CreateStatsRequest{
+		ClusterID:    s.clusterID,
+		TaskID:       100,
+		CollectionID: s.collectionID,
+		PartitionID:  s.partitionID,
+		SegmentID:    102,
+		// #nosec G101 -- non-secret sentinel credentials verify that logs elide these fields.
```
XuanYang-cn: ?

internal/proxy/util.go
```diff
@@ -1497,11 +1497,9 @@ func PasswordVerify(ctx context.Context, username, rawPwd string) bool {
 }
 
 func VerifyAPIKey(rawToken string) (string, error) {
-	hoo := hookutil.GetHook()
-	user, err := hoo.VerifyAPIKey(rawToken)
+	user, err := hookutil.GetHook().VerifyAPIKey(rawToken)
 	if err != nil {
-		mlog.Warn(context.TODO(), "fail to verify apikey", mlog.String("api_key", rawToken), mlog.Err(err))
```
XuanYang-cn, line 1503: No need to remove this log right?

pkg/util/externalspec/external_spec.go
```diff
@@ -173,7 +173,7 @@ func ValidateExternalSource(source string) error {
 	}
 	u, err := url.Parse(source)
 	if err != nil {
-		return merr.Wrap(err, "invalid external_source URL")
+		return merr.WrapErrParameterInvalidMsg("external_source is not a valid URL")
```
XuanYang-cn, line 182: missing the err

pkg/util/externalspec/external_spec.go
```diff
@@ -197,18 +197,43 @@ func ValidateExternalSource(source string) error {
 // Called from Proxy and RootCoord (defense in depth) on create-collection.
 func ValidateSourceAndSpec(externalSource, externalSpec string) error {
 	if err := ValidateExternalSource(externalSource); err != nil {
-		return merr.WrapErrParameterInvalid("valid external_source", externalSource, err.Error())
+		return merr.WrapErrParameterInvalidMsg("external_source is invalid")
```
XuanYang-cn: ditto

pkg/util/externalspec/external_spec.go
```diff
@@ -197,18 +197,43 @@ func ValidateExternalSource(source string) error {
...
 	}
 	spec, err := ParseExternalSpec(externalSpec)
 	if err != nil {
-		return merr.WrapErrParameterInvalid("valid external_spec", "<redacted>", err.Error())
+		return merr.WrapErrParameterInvalidMsg("external_spec is invalid")
 	}
 	if err := ValidateExtfsComplete(externalSource, spec.Extfs); err != nil {
-		return merr.WrapErrParameterInvalid("valid external_spec", "<redacted>", err.Error())
+		return merr.WrapErrParameterInvalidMsg("external_spec is invalid")
 	}
```
XuanYang-cn, line 214: dittlo

liliu-z:
**internal/datanode/index/task_index.go:342** task_index.go:342 logs the full buildIndexParams at INFO on every index build, and buildIndexParams.StorageConfig carries the raw SecretAccessKey (:278) and GcpCredentialJSON (:290) — live S3/GCP storage credentials. This is the identical credential-in-log pattern this PR fixes in the sibling file task_stats.go:185 (switched to logging only storageType), but the BuildIndex path was missed, so AK/SK- and GCP-backed collections still write cloud secrets to INFO logs. Log only non-secret fields (e.g. storageType) or redact the credential keys before logging.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -477,6 +477,34 @@ func restfulSizeMiddleware(handler gin.HandlerFunc, observeOutbound bool) gin.Ha
 
 func getTraceLogRequestFieldWithoutSensitiveInfo(req any) mlog.Field {
 	switch request := req.(type) {
+	case *CollectionReq:
```
liliu-z, line 480: getTraceLogRequestFieldWithoutSensitiveInfo's type switch handles CollectionReq/RefreshExternalCollectionReq/PasswordReq/NewPasswordReq but not *ImportReq, so an import request falls through to the default branch (line 522, `return proxy.GetRequestFieldWithoutSensitiveInfo(req)`) and is logged raw via mlog.Any. ImportReq.Options can carry the `ezk` option (importutilv2/option.go:73, the base64 encryption-zone key for encrypted backup data), so under TraceLogMode 2/3 that key is written verbatim. Add an *ImportReq case that drops/redacts the ezk option. (The original note's 'when the redaction hook returns an error' path does not exist here.)

internal/proxy/trace_log_interceptor.go
```diff
@@ -111,6 +111,20 @@ func GetRequestFieldWithoutSensitiveInfo(req interface{}) mlog.Field {
 			ModifiedUtcTimestamps: updateCredentialReq.ModifiedUtcTimestamps,
 		})
 	}
+	createCollectionReq, ok := req.(*milvuspb.CreateCollectionRequest)
```
liliu-z, line 118: The grpc TraceLogInterceptor logs GetRequestFieldWithoutSensitiveInfo(req) for every unary method under TraceLogMode 2/3, and this PR extends that redaction chain for CreateCollection/RefreshExternalCollection — but it omits *internalpb.ImportRequest (grpc ImportV2). ImportRequest.Options carries the `ezk` encryption-zone key, so an import falls through to the default `return mlog.Any("request", req)` and the key is logged raw. This is the grpc-side twin of the REST import gap (Issue 2); both surfaces need an ImportRequest case that redacts the ezk option.

internal/datanode/index/task_stats.go
XuanYang-cn: Fixed in a6f4b84280: clone the complete storage config, redact access key ID, secret access key, SSL CA certificate, and GCP credential JSON, while preserving operational fields such as address, bucket, root path, and storage type.

internal/datanode/index/task_stats_test.go
XuanYang-cn: Fixed in a6f4b84280: removed the suppression. The credential sentinels are assembled at runtime, so gosec does not mistake test data for hard-coded credentials. Static-check passes.

internal/proxy/util.go
XuanYang-cn: Kept one sanitized warning at AuthenticationInterceptor, where the generic VerifyAPIKey error is logged. Re-adding the lower-level hook error log would duplicate the warning and could expose hook details. The test now verifies the warning exists without logging the raw or encoded token.

pkg/util/externalspec/external_spec.go
XuanYang-cn: Fixed in a6f4b84280: the URL parse cause is retained in the returned error after redacting the submitted source value.

pkg/util/externalspec/external_spec.go
XuanYang-cn: Fixed in a6f4b84280: source validation now includes the sanitized underlying cause while keeping submitted credentials out of the error.

pkg/util/externalspec/external_spec.go
XuanYang-cn: Fixed in a6f4b84280: spec parsing and completeness validation now include the sanitized cause, with submitted spec values redacted.

internal/distributed/proxy/httpserver/handler_v2.go
XuanYang-cn: No change needed: ezk is encrypted by contract, so it is safe to include in REST trace logs.

internal/proxy/trace_log_interceptor.go
XuanYang-cn: No change needed: ezk is encrypted by contract, so it is safe to include in gRPC trace logs.

pkg/util/externalspec/external_spec.go
```diff
@@ -173,7 +247,7 @@ func ValidateExternalSource(source string) error {
...
 	u, err := url.Parse(source)
 	if err != nil {
-		return merr.Wrap(err, "invalid external_source URL")
+		return merr.WrapErrParameterInvalidMsg("external_source is not a valid URL: %s", redactErrorMessage(err, source))
```
XuanYang-cn: I dont think we need to redact the error message, because it's invalid, invalidation in the error can be print. This is over complicated for a error from Parse, Validate and urlParse error message.

XuanYang-cn: Fixed: removed the error-message redaction helpers and now preserve the original parse and validation details while keeping the typed error code.

XuanYang-cn:
@liliu-z Fixed in 31369f02e4: BuildIndex now logs a cloned BuildIndexInfo with storage credentials and external spec/source secrets redacted. The original unredacted parameters still go to CreateIndex.

czs007:

## Overview

This review covers credential-redaction coverage and test consistency in the external-collection / storage-config logging paths; the findings below are grouped by severity, with pre-existing hazards separated out at the end.

---

## High

**`internal/proxy/impl_test.go:3348` — stale test cases assert redaction behavior this PR removed**
The final commit (`31369f02e4`) deleted `redactErrorMessage` / `externalSpecErrorValues` and changed `ValidateSourceAndSpec` to `merr.Wrap(err, ...)`, which preserves the inner error verbatim. But the two cases added in the previous commit — `invalid format value redacted` (:3348) and `invalid extfs value redacted` (:3349) — still carry `secretSubstrs` and are checked by `assert.NotContains(t, resp.GetStatus().GetReason(), secret)` at :3363-3365. Both cases reach `ValidateSourceAndSpec` via `impl.go:7344`, and `merr.Status` sets `Reason: err.Error()` (`pkg/util/merr/utils.go:121`), so the sentinels are present in the reason string. `external_spec_test.go:225/233` asserts `assert.Contains` on the same function and the same input class, so the two test files now contradict each other.
Failure: running `go test ./internal/proxy/ -run TestProxy_RefreshExternalCollection_AtomicSourceSpec` → both sub-cases fail on the observed reasons `external_spec is invalid: parse external spec: unsupported format "FORMAT_REFRESH_SECRET_SENTINEL", ...` and `external_spec is invalid: extfs.cloud_provider="cloud_provider_refresh_secret_sentinel" is not supported: ...`, turning CI red for the whole PR.
Suggestion: drop `secretSubstrs` from these two cases (or flip them to `assert.Contains`) and rename them so `"…redacted"` no longer describes removed behavior. The `secretSubstrs` on the first two cases (:3341/:3342) hit the `srcSet != specSet` branch, whose message carries only the `source_set`/`spec_set` booleans, and should stay. *(raised by czs007)*

---

## Medium

**`pkg/util/externalspec/external_spec.go:218` — `RedactExternalSource` returns opaque / host-less URIs verbatim, leaking embedded credentials**
The function returns the input unchanged whenever `url.Parse` succeeds and `User`, `RawQuery`, and `Fragment` are all empty — it requires neither an allowlisted scheme nor a non-empty host. `s3:ACCESS_KEY:SECRET_KEY@bucket/path` parses as an opaque URI (`Opaque="ACCESS_KEY:SECRET_KEY@bucket/path"`, `User=nil`) and is therefore treated as a normal URI and returned as-is (confirmed by running it against the PR head). `ValidateExternalSource` then rejects it for an empty host — and that rejection branch is exactly what logs the supposedly-redacted field at WARN (`handler_v2.go:2330`, `:2341`).
Failure: `POST /v2/vectordb/collections/create` with `schema.externalSource = s3:AKIAXXXXXXXX:wJalrSecretKey@bucket/path` → the AK/SK pair is written in cleartext to the WARN log by the rejection path, and again to the INFO trace log under `TraceLogMode=2/3`; disabling trace logging does not stop the WARN leak.
Suggestion: before returning a source unchanged, require the same structural validity as `ValidateExternalSource` (scheme in the allowlist, non-empty host, empty `Opaque`); otherwise return `<redacted>`. Add secret-bearing opaque and scheme-less cases — `TestRedactExternalSource` (`external_spec_test.go:162-177`) currently covers only userinfo/query/fragment/malformed. *(raised by chyezh)*

**`internal/proxy/util.go:1502` — `VerifyAPIKey` discards the hook error, so the authentication-failure reason is no longer recorded anywhere**
The change from "Warn (with `rawToken` and `err`) + token-bearing error" to a bare `return "", merr.WrapErrParameterInvalidMsg("invalid API key")` correctly removes the token, but it also drops the hook's `err` entirely, with nothing else recording it. Both callers (`authentication_interceptor.go:86`, `distributed/proxy/service.go:147`) only log `mlog.Warn(ctx, "fail to verify apikey", mlog.Err(err))`, and by then `err` is the information-free `"invalid API key"`.
Failure: the plugin hook's `VerifyAPIKey` fails because the auth backend is unreachable (timeout / 5xx) rather than because the key is invalid → the log reads only `fail to verify apikey error=invalid API key: invalid parameter`, and operators cannot distinguish a revoked key from a down auth backend; the reason no longer exists in any log.
Suggestion: log a Warn carrying `err` only (no token) before returning, or keep the cause as an inner error via `merr.Wrap(err, "invalid API key")` if the wrapped text is safe to surface. *(raised by czs007)*

---

## Low

**`pkg/util/typeutil/schema.go:2656` — the two external-spec validation paths in this PR adopt opposite error-redaction policies**
This PR changes `return err` to `return merr.WrapErrParameterInvalidMsg("external_spec is invalid")`, collapsing every reason from `ParseExternalSpec` (unsupported format, disallowed extfs key, malformed boolean) into one uninformative message. The same PR changes `ValidateSourceAndSpec` (`external_spec.go:200-206`) to `merr.Wrap(err, ...)`, preserving those same underlying errors in full — and its tests explicitly assert the original values appear (`external_spec_test.go:225/233`), while `schema_test.go:5509-5514` asserts the opposite for the same input class. Both paths parse the same `external_spec` string.
Suggestion: pick one policy. Either keep detail on both sides (restore `return err` in `schema.go`), or converge on both (reintroduce redaction in `external_spec.go` and fix the contradictory assertions noted in the High item above). *(raised by czs007)*

---

- **`internal/proxy/trace_log_interceptor.go:128`** — the Restore/Export snapshot branches (`trace_log_interceptor.go:128-138`, `handler_v2.go:508-521`) still use `RedactExternalSpec`, which only masks the 4 keys in the `secretExtfsKeys` allowlist and preserves unknown extfs keys and unknown top-level fields verbatim (`external_spec.go:540-541`); `RestoreExternalSnapshotReq.SnapshotMetadataURI` and `ExportSnapshotReq.TargetS3Path` are not redacted at all. Under `TraceLogMode=2/3` the request log happens before business validation, so `{"extfs":{"future_password":"SECRET"}}` and `s3://b/meta?X-Amz-Signature=SECRET` land in INFO logs. Unrelated to this PR: both snapshot cases are pure context lines in the diff, and `git show 7280a8601e:internal/proxy/trace_log_interceptor.go` already has the identical `RedactExternalSpec` call with the URI fields unredacted. Follow-up: route these through `RedactExternalSpecForLog` and apply `RedactExternalSource` to the two URI fields, with regression tests for unknown spec keys and query-bearing URIs. *(raised by chyezh)*
- **`pkg/util/externalspec/external_spec.go:179`** — validation errors echo the original `external_spec` / `external_source` values back to the client and into logs: `unsupported format %q`, `extfs.cloud_provider=%q is not supported`, and `WrapErrParameterInvalidErr(err, ...)` on the `url.Parse` failure (a `url.Error` inherently contains the full URL), all reaching `Status.Reason` via `merr.Status`. Unrelated to this PR: replaying the same inputs against merge-base `7280a8601e` produces `...unsupported format "FORMAT_REFRESH_SECRET_SENTINEL", ...[actual=<redacted>]` — the value was already carried out by `err.Error()`; for `external_source` this PR is a net improvement, since the old `WrapErrParameterInvalid("valid external_source", externalSource, ...)` printed the full userinfo-bearing URL as the `actual` parameter. Follow-up: if the "do not redact error messages" decision stands, document the trade-off in a comment on `ValidateSourceAndSpec`. *(surfaced during verification)*
- **`internal/rootcoord/create_collection_task.go:563`** — internal components still log the unredacted `externalSource` (`create_collection_task.go:563`, `datacoord/ddl_callbacks_external_collection.go:42`, `datacoord/external_collection_refresh_manager.go:580`, `datacoord/server_external_schema_updater.go:54`, `datanode/services.go:994`). `ValidateExternalSource` does not reject query/fragment, so `s3://bucket/path?X-Amz-Security-Token=SECRET` is a valid request whose STS token is written to disk by RootCoord/DataCoord/DataNode even though the Proxy now treats query/fragment as credential carriers. Unrelated to this PR: none of those five `mlog.String("externalSource", …)` call sites appear in the diff. Follow-up: either push `RedactExternalSource` down to these log points, or reject query/fragment-bearing sources in `ValidateExternalSource` so both functions agree. *(surfaced during verification)*
- **`internal/datanode/index_services.go:89`** — the chunk-manager creation failure branches log `mlog.String("accessKey", req.GetStorageConfig().GetAccessKeyID())` at ERROR and return immediately, at three sites (`:89` CreateJob, `:291`, `:394` CreateStats/CreateTask). These run before this PR's new `redactBuildIndexParamsForLog` / `redactStorageConfigForLog` points, so a misconfigured endpoint or CA makes the task leak the AccessKeyID and never reach `PreExecute`/`Execute`. Unrelated to this PR: `index_services.go` is not in the diff's file list (only `task_index.go`, `task_stats.go`, and two test files). Follow-up: drop the `accessKey` field or reuse `redactStorageConfigForLog`. *(surfaced during verification)*
- **`internal/proxy/hook_interceptor.go:48`** — the Before/After failure branches still pass `mlog.Err(err)` (:48, :58) and build `status.Error(codes.InvalidArgument, "detail: "+err.Error())` (:52, :62); the Mock branch (:36-43) does the same. A third-party hook that interpolates a `CreateCredential`/`UpdateCredential` password into its own error would bypass this PR's request redaction via the error field. Unrelated to this PR: the diff only swapped the request field from `RedactReqForLog` to `GetRequestFieldWithoutSensitiveInfo` on those two lines; the error handling is unchanged context. This is defensive hardening against plugin implementations rather than a demonstrated leak in this repo — the new test uses a fixed-text error. Follow-up: constrain `err.Error()` on these branches (length/allowlist), or document the constraint on the hook interface. *(surfaced during verification)*

XuanYang-cn:
Addressed in 61fec39b9c:

- Updated the stale Proxy cases and restored detailed external-spec validation errors consistently.
- Made `RedactExternalSource` fail closed for opaque, invalid, unsupported, credential-bearing, query, and fragment URIs.
- Preserved trusted API-key hook failure detail without logging or returning the token.
- Applied the agreed audit fixes for snapshot/DataCoord external specs, RBAC hashes, KV values, DataNode access keys, and test password logs.

Raw snapshot URL/path handling and hook-wide error sanitization remain unchanged per the audit consensus.

internal/kv/etcd/etcd_kv.go
```diff
@@ -396,7 +403,7 @@ func (kv *etcdKV) MultiSave(ctx context.Context, kvs map[string]string) error {
 	CheckTnxStringValueSizeAndWarn(ctx, kvs)
 	_, err := kv.executeTxn(kv.getTxnWithCmp(ctx1), ops...)
 	if err != nil {
-		mlog.Warn(ctx, "Etcd MultiSave error", mlog.Any("kvs", kvs), mlog.Int("len", len(kvs)), mlog.Err(err))
+		mlog.Warn(ctx, "Etcd MultiSave error", mlog.Strings("keys", sortedMapKeys(kvs)), mlog.Int("len", len(kvs)), mlog.Err(err))
```
XuanYang-cn: I think we only needs to lo.keys, dont need to sort it

internal/kv/etcd/etcd_kv.go
```diff
@@ -418,7 +425,7 @@ func (kv *etcdKV) MultiSaveBytes(ctx context.Context, kvs map[string][]byte) err
 	CheckTnxBytesValueSizeAndWarn(ctx, kvs)
 	_, err := kv.executeTxn(kv.getTxnWithCmp(ctx1), ops...)
 	if err != nil {
-		mlog.Warn(ctx, "Etcd MultiSaveBytes err", mlog.Any("kvs", kvs), mlog.Int("len", len(kvs)), mlog.Err(err))
+		mlog.Warn(ctx, "Etcd MultiSaveBytes err", mlog.Strings("keys", sortedMapKeys(kvs)), mlog.Int("len", len(kvs)), mlog.Err(err))
```
XuanYang-cn: ditto

internal/kv/etcd/etcd_kv.go
```diff
@@ -496,7 +503,7 @@ func (kv *etcdKV) MultiSaveAndRemove(ctx context.Context, saves map[string]strin
 	resp, err := kv.executeTxn(kv.getTxnWithCmp(ctx1, cmps...), ops...)
 	if err != nil {
 		mlog.Warn(ctx, "Etcd MultiSaveAndRemove error",
-			mlog.Any("saves", saves),
+			mlog.Strings("saveKeys", sortedMapKeys(saves)),
```
XuanYang-cn: ditto

internal/kv/etcd/etcd_kv.go
```diff
@@ -531,7 +538,7 @@ func (kv *etcdKV) MultiSaveBytesAndRemove(ctx context.Context, saves map[string]
 	_, err := kv.executeTxn(kv.getTxnWithCmp(ctx1), ops...)
 	if err != nil {
 		mlog.Warn(ctx, "Etcd MultiSaveBytesAndRemove error",
-			mlog.Any("saves", saves),
+			mlog.Strings("saveKeys", sortedMapKeys(saves)),
```
XuanYang-cn: ditto

internal/kv/etcd/etcd_kv.go
```diff
@@ -593,7 +600,7 @@ func (kv *etcdKV) MultiSaveAndRemoveWithPrefix(ctx context.Context, saves map[st
 	resp, err := kv.executeTxn(kv.getTxnWithCmp(ctx1, cmps...), ops...)
 	if err != nil {
 		mlog.Warn(ctx, "Etcd MultiSaveAndRemoveWithPrefix error",
-			mlog.Any("saves", saves),
+			mlog.Strings("saveKeys", sortedMapKeys(saves)),
```
XuanYang-cn: ditto

internal/kv/etcd/etcd_kv.go
```diff
@@ -627,7 +634,7 @@ func (kv *etcdKV) MultiSaveBytesAndRemoveWithPrefix(ctx context.Context, saves m
 	_, err := kv.executeTxn(kv.getTxnWithCmp(ctx1), ops...)
 	if err != nil {
 		mlog.Warn(ctx, "Etcd MultiSaveBytesAndRemoveWithPrefix error",
-			mlog.Any("saves", saves),
+			mlog.Strings("saveKeys", sortedMapKeys(saves)),
```
XuanYang-cn: ditto

internal/datanode/index/task_index.go
```diff
@@ -236,6 +239,29 @@ func (it *indexBuildTask) PreExecute(ctx context.Context) error {
...
+	redacted := proto.Clone(params).(*indexcgopb.BuildIndexInfo)
+	if config := redacted.GetStorageConfig(); config != nil {
+		for _, secret := range []*string{
+			&config.AccessKeyID,
+			&config.SecretAccessKey,
+			&config.SslCACert,
+			&config.GcpCredentialJSON,
+		} {
+			if *secret != "" {
+				*secret = "<redacted>"
```
XuanYang-cn: Is Milvus previous faver "<redacted>" or "*****", I dont mind use which, just keep it align with original design. If they both exists already in the milvus, keep this pr unchange.

internal/datanode/index/task_stats.go
```diff
@@ -147,6 +147,25 @@ func (st *statsTask) IsVectorIndex() bool {
 	return false
 }
 
+func redactStorageConfigForLog(config *indexpb.StorageConfig) *indexpb.StorageConfig {
```
XuanYang-cn, line 158: The code dup itself with the redactBuildIndexParamsForLog

internal/datanode/sensitive_log_test.go
XuanYang-cn, line 1: prefer to not add a new test file.

internal/kv/tikv/txn_tikv.go
```diff
@@ -159,6 +160,12 @@ type txnTiKV struct {
 	requestTimeout time.Duration
 }
 
+func sortedMapKeys[V any](values map[string]V) []string {
```
XuanYang-cn: ditto this whole file. no need to sort.

internal/proxy/trace_log_interceptor.go
```diff
@@ -135,6 +153,41 @@ func GetRequestFieldWithoutSensitiveInfo(req interface{}) mlog.Field {
...
 }
 
+func redactRestoreRBACRequestForLog(req *milvuspb.RestoreRBACMetaRequest) *milvuspb.RestoreRBACMetaRequest {
+	if req == nil {
+		return nil
+	}
+	redacted := proto.Clone(req).(*milvuspb.RestoreRBACMetaRequest)
+	for _, user := range redacted.GetRBACMeta().GetUsers() {
+		if user != nil {
+			user.Password = ""
```
XuanYang-cn: Over this implimentation, the redact results are not consists with each other, keep one form. either redacted, or ******.

pkg/util/externalspec/external_spec.go
```diff
@@ -176,7 +176,7 @@ func ValidateExternalSource(source string) error {
 	}
 	u, err := url.Parse(source)
 	if err != nil {
-		return merr.Wrap(err, "invalid external_source URL")
+		return merr.WrapErrParameterInvalidErr(err, "external_source is not a valid URL")
```
XuanYang-cn: Lets restore it, since it adds no changes but only error message

internal/proxy/trace_log_interceptor.go
```diff
@@ -111,16 +111,34 @@ func GetRequestFieldWithoutSensitiveInfo(req interface{}) mlog.Field {
 			ModifiedUtcTimestamps: updateCredentialReq.ModifiedUtcTimestamps,
 		})
 	}
+	restoreRBACReq, ok := req.(*milvuspb.RestoreRBACMetaRequest)
+	if ok {
+		return mlog.Any("request", redactRestoreRBACRequestForLog(restoreRBACReq))
+	}
+	createCollectionReq, ok := req.(*milvuspb.CreateCollectionRequest)
+	if ok {
+		return mlog.Any("request", redactCreateCollectionRequestForLog(createCollectionReq))
```
liliu-z, line 120: This PR redacts a CollectionSchema's ExternalSource/ExternalSpec for the CreateCollection gRPC log here, but the same schema is Info-logged unredacted on the load path: querycoordv2/services.go:200 and :274 log `req.Schema`, and querynodev2/segments/collection.go:138/140/166 log the collection `schema`. segment_loader.go:1243-1244 feeds `schema.GetExternalSpec()` straight into external-storage authentication, which proves the spec still carries live extfs credentials (AK/SK) at load time, so every load of an AK/SK-mode external collection writes the access key/secret into Info logs. Apply the same schema redaction at those load-path log sites.

internal/kv/etcd/etcd_kv.go
XuanYang-cn: Fixed: use lo.Keys directly; save values remain omitted from the log.

internal/datanode/index/task_index.go
XuanYang-cn: Kept unchanged: upstream/master already uses both <redacted> and masked-star forms, as noted in the comment.

internal/kv/etcd/etcd_kv.go
XuanYang-cn: Fixed: use lo.Keys directly; save values remain omitted from the log.

internal/proxy/trace_log_interceptor.go
XuanYang-cn: Fixed: RBAC password hashes now use the existing sensitiveMark (*****), instead of an empty string.

internal/kv/etcd/etcd_kv.go
XuanYang-cn: Fixed: use lo.Keys directly; save values remain omitted from the log.

internal/datanode/index/task_stats.go
XuanYang-cn: Fixed: both typed storage configs now reuse one shared credential-field redaction helper.

internal/kv/etcd/etcd_kv.go
XuanYang-cn: Fixed: use lo.Keys directly; save values remain omitted from the log.

internal/kv/etcd/etcd_kv.go
XuanYang-cn: Fixed: use lo.Keys directly; save values remain omitted from the log.

internal/proxy/trace_log_interceptor.go
XuanYang-cn: Fixed: added mlog.FieldSchema backed by the shared schema sanitizer and migrated every direct CollectionSchema log across Proxy, QueryCoord, QueryNode, and DataNode, including all cited load paths.

internal/datanode/sensitive_log_test.go
XuanYang-cn: Fixed: moved the regression into internal/datanode/services_test.go and removed the new test file.

pkg/util/externalspec/external_spec.go
XuanYang-cn: Fixed: restored the original merr.Wrap(err, "invalid external_source URL") behavior.

internal/kv/etcd/etcd_kv.go
XuanYang-cn: Fixed: use lo.Keys directly; save values remain omitted from the log.

internal/kv/tikv/txn_tikv.go
XuanYang-cn: Fixed: replaced the sorted helper with lo.Keys throughout the TiKV file.

pkg/util/externalspec/external_spec.go
```diff
@@ -196,22 +199,80 @@ func ValidateExternalSource(source string) error {
 }
 
 // ValidateSourceAndSpec validates URL + JSON shape + ValidateExtfsComplete.
-// Errors are wrapped via merr.WrapErrParameterInvalid for direct return.
+// Errors retain their typed validation code while adding field context.
 // Called from Proxy and RootCoord (defense in depth) on create-collection.
 func ValidateSourceAndSpec(externalSource, externalSpec string) error {
 	if err := ValidateExternalSource(externalSource); err != nil {
-		return merr.WrapErrParameterInvalid("valid external_source", externalSource, err.Error())
+		return merr.Wrap(err, "external_source is invalid")
```
tedxu, line 206: Malformed `external_source` still leaks embedded credentials here, and the new test pins that behavior.

`ValidateExternalSource` only checks `u.User != nil` *after* a successful parse. Its parse-failure branch returns `merr.Wrap(err, "invalid external_source URL")`, and Go's `url.Error.Error()` embeds the raw URL verbatim:

```
"s3://ACCESS_SENTINEL:SECRET_SENTINEL@bad host/path"
  -> parse "s3://ACCESS_SENTINEL:SECRET_SENTINEL@bad host/path": invalid character " " in host name
```

That string is now propagated unchanged by this line, and it reaches both the client-visible `Status.Reason` and the logs (`mlog.Err(err)`) at every call site: `internal/proxy/impl.go:7346`, `internal/rootcoord/create_collection_task.go:214` and `:490`, `internal/datacoord/external_collection_refresh_manager.go:886`. `RedactExternalSource` correctly returns `<redacted>` for exactly this input on the log-field path — the error path bypasses it entirely.

The new `malformed_source_includes_parse_error` case in `external_spec_test.go` asserts the sentinel **is** contained, so this is locked in by a test.

Secondary effect on the wire code: `merr.Wrap` is the right call per the error-handling guide, and every other branch keeps `ErrParameterInvalid` because `ValidateExternalSource` / `ParseExternalSpec` / `ValidateExtfsComplete` return real `milvusError`s. But the parse branch wraps a bare `url.Error`, so `merr.Code()` finds no `milvusError` in the chain and falls through to `errUnexpected` (65535). Before this change the outer `WrapErrParameterInvalid` re-labeled it to 1100, so a malformed URI now surfaces as "unexpected error" while every sibling validation failure stays ParameterInvalid.

Suggested fix, in `ValidateExternalSource` — don't embed the `url.Parse` error:

```go
u, err := url.Parse(source)
if err != nil {
    return merr.WrapErrParameterInvalidMsg("external_source is not a valid URL")
}
```

That fixes the leak and restores the code in one move. Then flip the test to assert the sentinel is absent, and add `assert.ErrorIs(t, err, merr.ErrParameterInvalid)` to the `ValidateSourceAndSpec` cases — the current tests only check message substrings, so the code regression is invisible to them.

XuanYang-cn: Fixed in 2ddb3b7c01: malformed external_source URLs now return a typed ParameterInvalid error without embedding the raw URL or url.Parse detail. Tests assert the sentinel is absent and validation codes remain typed.

PR #52042, opened 2026-07-30
fix: [3.0] use serving cluster ID for external table embeddings

weiliu1031:
issue: #45691
pr: #51997

## What changed

- Backport payload-based serving cluster ID propagation for external
  refresh tasks.
- Populate `RefreshExternalCollectionTaskRequest.clusterID` in
  DataCoord.
- Reuse the payload value for unified `CreateTask` properties.
- Normalize legacy payloads from `CreateTask` properties when the
  payload cluster ID is empty.
- Read the normalized payload directly for manager registration and
  embedding execution.
- Remove the constructor-level cluster ID and worker config fallback.

## Why

A shared DataNode worker can execute external-table refresh tasks from
multiple serving clusters. Its process-level cluster prefix identifies
the worker rather than the cluster that submitted the task, causing
embedding authentication to use the wrong identity.

## Branch difference

The payload propagation and rolling-upgrade fallback are identical to
master. The generated `data_coord.pb.go` reflects the 3.0 proto
baseline. No branch-specific behavior changes are introduced.

## Validation

Not run for this exact candidate during preview.


congqixia (approved):
/lgtm

Issue #52043, opened 2026-07-30
[Bug]: Snapshot V2 includes uncommitted Storage V3 growing segments


weiliu1031:
### Environment

- Deployment: VDC UAT, AWS us-west-2
- Milvus build: `unknown-20260729-6afd297f7`
- Git commit: `6afd297f7`
- Python: `3.12.12`
- pymilvus: `3.0.1rc71`
- pymilvus commit: `cb16a04cb0f0f7c07260c56c4bfd5e8442692ed7`
- Test: `tests/sdk/test_milvus3_vdc_sdk.py::test_snapshot_point_in_time_restore`
- RPC timeout: 15 seconds
- Test timeout: 45 seconds

### Current Behavior

A new Storage V3 Growing segment receives a non-empty manifest path
using `ManifestEarliest` version 0 before it has committed any files.

Snapshot V2 treats every non-empty manifest path as evidence that the
segment contains data. The empty Growing segment is therefore included
in the snapshot.

Restore schedules a `CopySegment` task for that segment and eventually
fails with:

```text
no insert/delete binlogs for segment
```

The same VDC instance reproduced the failure in three consecutive runs.
Snapshot creation and restore submission succeeded; the asynchronous
restore task entered a server-side failed state.

### Expected Behavior

A Storage V3 segment whose only storage metadata is a version 0
`ManifestEarliest` placeholder should not be included in Snapshot V2.

A committed manifest-only Storage V3 segment must remain eligible for
snapshot and restore.

### Steps To Reproduce

1. Create a Storage V3 collection.
2. Insert two rows and flush.
3. Insert one additional row without flushing.
4. Confirm the Growing row is visible with a Strong query.
5. Create a Snapshot V2 snapshot.
6. Restore the snapshot into a new collection.
7. Poll the restore task until it fails.

### Milvus Log

The representative server timeline shows:

- the flushed segment committed a manifest and entered `Flushed`;
- a second segment was created in `Growing` state with one row and no
  binlogs or deltalogs;
- the snapshot recorded both segments;
- restore created two `CopySegment` inputs;
- the copy task failed with
  `no insert/delete binlogs for segment`.

### Anything else?

The snapshot filter currently considers a segment to contain data when
any of these conditions is true:

```text
len(Binlogs) > 0 || len(Deltalogs) > 0 || ManifestPath != ""
```

For Storage V3, a non-empty manifest path is insufficient because
`ManifestEarliest` is assigned when a Growing segment is allocated.

[...]

Issue #51836, opened 2026-07-24
[Bug]: Storage V3 snapshot restore rejects manifest-only segments as missing binlogs


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: `master-20260723-f113bdbc8e`
- Git commit: `f113bdbc8e39efa8821b7b3f7151287ce1bd0b9c`
  - This is the exact `GitCommit` reported by both DataNode and MixCoord startup banners.
- Deployment mode: cluster
- MQ type: Pulsar
- SDK version: pymilvus `3.1.0rc69`
- OS: Linux container; the exact distribution was not retained
- CPU/Memory:
  - DataNode: request 4 CPU / 16 GiB, limit 16 CPU / 32 GiB
  - MixCoord: request 1 CPU / 4 GiB, limit 16 CPU / 32 GiB
- GPU: N/A
- Storage configuration:
  - `common.storage.enablev2=true`
  - `common.storage.useLoonFFI=true`
  - MinIO object storage
- Streaming service: enabled
- Chaos type: Proxy pod failure

Pod state after chaos recovery:

```text
NAME                                                               READY   STATUS      RESTARTS     AGE
proxy-pod-failure-592-etcd-0                                      1/1     Running     0            61m
proxy-pod-failure-592-etcd-1                                      1/1     Running     0            61m
proxy-pod-failure-592-etcd-2                                      1/1     Running     0            61m
proxy-pod-failure-592-milvus-datanode-67f9f6c849-f777h            1/1     Running     2 (59m ago)  61m
proxy-pod-failure-592-milvus-datanode-67f9f6c849-fb2mc            1/1     Running     0            61m
proxy-pod-failure-592-milvus-mixcoord-5bb9f8b666-2hd92            1/1     Running     1 (59m ago)  61m
proxy-pod-failure-592-milvus-proxy-7db54d8d56-wsdfv               1/1     Running     5 (31m ago)  61m
proxy-pod-failure-592-milvus-querynode-8946b5585-5ll6f            1/1     Running     2 (59m ago)  61m
proxy-pod-failure-592-milvus-querynode-8946b5585-bxwkm            1/1     Running     0            61m
proxy-pod-failure-592-milvus-querynode-8946b5585-pnht4            1/1     Running     0            61m
proxy-pod-failure-592-milvus-streamingnode-867989d888-m7xkh       1/1     Running     1 (59m ago)  61m
[...]

PR #52044, opened 2026-07-30
fix: exclude uncommitted V3 manifests from snapshots

weiliu1031:
issue: #52043
pr: #50410

## What this PR does

Snapshot V2 used any non-empty `ManifestPath` as evidence that a segment
contains data. Storage V3 assigns a `ManifestEarliest` version 0 path as
soon as a Growing segment is allocated, before it has committed files.

That placeholder therefore passed snapshot selection and created a
`CopySegment` restore task for a segment with no insert or delete
binlogs.

This PR separates segment-state filtering from data-presence filtering:

- existing insert and delete binlog arrays continue to qualify a
  segment;
- manifest-only Storage V3 segments qualify only when their manifest
  version is greater than `ManifestEarliest`;
- version 0 placeholders are excluded before channel timestamp
  filtering;
- malformed persisted manifest paths return a data-integrity error.

## Behavior boundary

The change keeps committed manifest-only Storage V3 segments eligible,
including segments reloaded without legacy `FieldBinlog` arrays. It
only removes uncommitted placeholders from Snapshot V2.

It intentionally does not loosen `CopySegment` validation. The legal
Flushed manifest-only restore problem tracked by
[#51836](https://github.com/milvus-io/milvus/issues/51836) remains a
separate downstream case.

## Test coverage

The DataCoord regression coverage includes:

- Growing V3 + version 0 + no legacy binlogs: excluded;
- Flushed V3 + committed manifest + no legacy arrays: included;
- mixed committed, Growing placeholder, and empty segments: only the
  committed segment is snapshotted;
- malformed manifest metadata: reported as `ErrDataIntegrity`;
- legacy and non-V3 behavior: unchanged.

## Validation

Not run for this exact candidate because matching `check-commit`
evidence is unavailable.

### DIFF
--- a/internal/datacoord/handler.go
+++ b/internal/datacoord/handler.go
@@ -33,6 +33,7 @@ import (
 	"github.com/milvus-io/milvus/internal/metastore/kv/binlog"
 	"github.com/milvus-io/milvus/internal/metastore/model"
 	"github.com/milvus-io/milvus/internal/storage"
+	"github.com/milvus-io/milvus/internal/storagev2/packed"
 	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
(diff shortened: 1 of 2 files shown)


congqixia (approved):
/lgtm

Issue #52045, opened 2026-07-30
[Bug]: Bulk import decodes the struct array column once per sub-field


bigsheeper:
## Environment

- **Milvus Version**: reproduced on `master` (verified at commit `de2d7e5cae`). Originally observed on a production cloud instance whose exact image tag is not confirmed; the relevant code is unchanged on master.
- **Deployment Mode**: standalone (import task executed on a separate worker node)
- **MQ**: not relevant — the defect is in the import file reader, independent of MQ
- **SDK**: n/a (triggered by bulk import of a parquet file)
- **OS**: Linux

### Option B: Steps

1. Create a collection with a **struct array field** that has **N sub-fields** (N > 1). The amplification factor scales with N, so make one sub-field dominate the payload — e.g. an `ArrayOfVector` sub-field with `dim=3584`, alongside 8 scalar `Array` sub-fields.
2. Write a parquet file for that schema.
3. Bulk import the file.
4. Measure the bytes actually pulled from object storage during `preimport` / `import` and compare against the file size.

### Verified local reproduction

A Go test in `internal/util/importutilv2/parquet` that wraps `storage.ChunkManager` with a byte counter and drives `parquet.NewReader(...).Read()` to EOF. No production-code instrumentation is needed, because `NewReader` already takes the `ChunkManager` interface.

Fixture: 9 struct sub-fields (8 × `Array`/VarChar + 1 × `ArrayOfVector`, `dim=3584`), 500 rows, single row group.

```
RESULT rows=500 fileSize=74475374 readBytes=670224762 ratio=9.00x elapsed=6.412125145s
```

### Trigger Conditions

- **Frequency**: always — fully deterministic
- **Scales with**: the number of struct sub-fields (N sub-fields → ~N× the struct column is read)
- **Does NOT happen when**: the collection has no struct array field, or the struct has exactly one sub-field. Top-level (non-struct) fields are unaffected: each top-level column index is consumed by exactly one reader.

## Expected Behavior

All sub-field readers of one struct array column together read that column **once**. For the fixture above, `readBytes` should be ≈ `fileSize` (ratio ≈ 1.0).

## Actual Behavior

The column is downloaded and decoded **once per sub-field**, and all but one leaf is discarded each time. Measured ratio 9.00x for 9 sub-fields.

Observed on a production instance with a 580 MB (608,065,141 bytes) parquet file containing only 10,000 rows:

| Phase | Duration |
|---|---|
| preimport | 2m10.0s |
[...]

PR #52046, opened 2026-07-30
fix: read only the needed leaf columns for struct array sub-fields

bigsheeper:
issue: #52045

## Summary

`NewStructFieldReader` created one arrow `ColumnReader` per struct sub-field but passed the **same top-level column index** to `GetColumn` for all of them. `GetColumn` hardcodes `filterLeaves: false`, so every sub-field reader built leaf readers for the entire `list<struct>` column, decoded all of them on each `NextBatch`, and kept only its own. The column was downloaded and decoded once per sub-field.

This PR resolves the leaf column indices belonging to each sub-field from the pqarrow manifest and reads through `GetFieldReader`, which prunes the rest.

## Measured

Local fixture with **9 struct sub-fields** including a dim-3584 `ArrayOfVector`, 500 rows, driven through `parquet.NewReader(...).Read()` to EOF with a byte-counting `ChunkManager`:

| | file size | bytes read | ratio | read loop |
|---|---|---|---|---|
| before | 74,475,374 | 670,224,762 | **9.00x** | 6.41s |
| after | 74,479,069 | 74,479,065 | **1.00x** | 1.27s |

A unit test with 3 sub-fields (`TestStructArrayReadAmplification`, added here) measures 2.96x before and 1.00x after, and now guards the regression in CI.

Causation was isolated: reverting only `struct_field_reader.go` while keeping the new tests reproduces 2.96x with the read-amplification assertion message, so the improvement is attributable to the leaf pruning rather than to any test-fixture change in this PR.

The elapsed-time gain is smaller than the byte gain because decode CPU does not shrink proportionally.

## Not measured

The production instance that surfaced this (see the issue) spent 2m10s in `preimport` and 2m14s in `import` on a 580 MB file, with node metrics showing ~4.7 GB pulled during `preimport` alone. The corresponding end-to-end improvement there is **expected but not verified** — the measurements above use a local `ChunkManager`, not object storage over the network. Since the production path is more IO-bound, the real gain is plausibly larger, but that remains to be confirmed.

## Implementation notes

Two details that are easy to get wrong:

1. **The leaf walk terminates on `len(Children) == 0`, not `SchemaField.IsLeaf()`.** arrow assigns `ColIndex` only in `populateLeaf` and zero-initialises every other `SchemaField`, so `IsLeaf()` (`ColIndex != -1`) reports true for group nodes and hands back a bogus index of 0. arrow's own `getReader` guards the same way. Using `IsLeaf()` silently resolves a nested sub-field such as `ArrayOfVector` (which is `list<list<float>>`) to the file's first leaf column. This is covered by an assertion that every resolved leaf's parquet path contains the sub-field name; mutation testing confirms the assertion fails when the condition is swapped.
2. **With pruning, arrow drops the pruned children and rebuilds the struct type from the survivors**, leaving exactly one child, so the sub-field always sits at `structArray.Field(0)`.

Failures in the new leaf resolution use `ErrImportSysFailed` rather than `ErrImportFailed`: by the time it runs, `columnIndex` and `fieldIndex` have already been validated against the same file's arrow types, and the arrow schema is derived from the manifest, so these branches can only fire on an internal inconsistency, not on malformed user data.

## Test plan

- [x] `TestStructArrayReadAmplification` — byte-count assertion, 2.96x → 1.00x
- [x] `TestCollectSubFieldLeaves` — every resolved leaf's parquet path belongs to its sub-field, sets are disjoint
- [x] `TestStructArrayLeafPruningWithNulls` — null rows survive pruning: readback `ValidData` is exactly `[true false true false true]` and non-null rows' vector contents match what was written
- [x] `./internal/util/importutilv2/...` and `./internal/datanode/importv2/...` green
- [x] `TestParquetReaderWithStructArray` (3 sub-fields × 5 vector types), `TestParquetReaderMissingNullableStructArray`, `TestStructFieldReader_NullableArrayOfVector` green
- [x] `make static-check` — 0 issues
- [x] `TestBulkInsert/TestImportWithVectorArray` — passes for the Parquet path (all 5 vector types). The unmodified test also covers CSV and JSON and currently aborts on an unrelated pre-existing failure in `internal/util/importutilv2/common/vector_array.go` (`expected slice as vector, but got string`, CSV + Float16Vector), which this branch does not touch; the Parquet iterations were verified by temporarily restricting `fileTypeArr` locally.

## Out of scope

Deliberately not included, each worth its own change:

- `EstimateReadCountPerBatch`'s `assumedArrayLen = 10` makes batches only ~112 rows for wide `ArrayOfVector` schemas. It does not affect bytes read.
- `RemoteChunkManager.Reader()` records no byte-size metric, unlike `Read()`/`ReadAt()`, so the streaming path this fix lives on is invisible to `PersistentDataKvSize`. Adding it would change an existing metric's distribution.
- `ArrowReadProperties.Parallel` is still off, so a struct's leaves load serially.

### DIFF
--- a/internal/util/importutilv2/parquet/leaf_pruning_test.go
+++ b/internal/util/importutilv2/parquet/leaf_pruning_test.go
@@ -1,0 +1,494 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parquet
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"math/rand"
+	"os"
+	"testing"
+
+	"github.com/apache/arrow/go/v17/arrow"
+	"github.com/apache/arrow/go/v17/arrow/array"
+	"github.com/apache/arrow/go/v17/arrow/memory"
+	"github.com/apache/arrow/go/v17/parquet"
+	"github.com/apache/arrow/go/v17/parquet/file"
+	"github.com/apache/arrow/go/v17/parquet/pqarrow"
+	"github.com/cockroachdb/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+
+	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
+	"github.com/milvus-io/milvus/internal/storage"
+	"github.com/milvus-io/milvus/pkg/v3/common"
+	"github.com/milvus-io/milvus/pkg/v3/objectstorage"
+)
+
+// leafPruningTestSchema builds a collection schema whose struct array field has
+// three sub-fields, one of them an ArrayOfVector. Callers should pass a dim large
+// enough that the vector sub-field dominates the on-disk size, otherwise read
+// amplification is not measurable against parquet metadata overhead.
+//
+// Every field is non-nullable. Callers that need null rows must set Nullable on
+// the fields they care about before writing.
+func leafPruningTestSchema(dim string, maxCapacity string) *schemapb.CollectionSchema {
+	return &schemapb.CollectionSchema{
+		Name: "test_leaf_pruning",
+		Fields: []*schemapb.FieldSchema{
+			{
+				FieldID:      100,
+				Name:         "id",
+				IsPrimaryKey: true,
+				DataType:     schemapb.DataType_Int64,
+			},
+			{
+				FieldID:  101,
+				Name:     "varchar_field",
+				DataType: schemapb.DataType_VarChar,
+				TypeParams: []*commonpb.KeyValuePair{
+					// testutil.CreateInsertData fills VarChar via
+					// testutils.GenerateStringArray, which builds 5-10 word
+					// sentences up to roughly 100 characters. A smaller
+					// max_length makes reader.Read fail CheckVarcharLength
+					// before any assertion in these tests is reached.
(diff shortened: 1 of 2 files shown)


czs007:

**Verified:**

- `struct_field_reader.go:132` — `fieldIndex: 0` is guaranteed rather than assumed, because `getReader`'s STRUCT branch compacts `childReaders`/`childFields` in place and rebuilds the type via `arrow.StructOf(childFields...)`, placing the single surviving child at position 0 by construction; the all-children-filtered `nil` case is caught by the guard at `struct_field_reader.go:106`.
- `struct_field_reader.go:92` vs `util.go:191-197` — the two index spaces agree, since `fieldIndex` is resolved by name against `fileReader.Schema()`, which builds a fresh manifest from the same parquet schema and props and emits fields in manifest order, so `structType.Fields()[i]` and `structField.Children[i]` are the same node.
- `struct_field_reader.go:102` — `columnIndex` semantics are unchanged between `GetColumn` and `GetFieldReader` (both terminate in `fr.getReader`), and although `GetFieldReader` does not bounds-check the index itself, `collectSubFieldLeaves:45` validates it first, so the direct index cannot go out of range.
- Struct reconstruction cannot flip mode: `newStructReader` sets `hasRepeatedChild = true` unless some child reports no repeated child, and `internal/proxy/util.go:588-591` rejects any struct sub-field that is not `Array`/`ArrayOfVector`, so every child is a `listReader` and the flag stays `true` both before and after pruning — the `validityIO.Read` vs `DefRepLevelsToBitmap` choice is unaffected.
- Same-mode, different level source is equivalent: `DefRepLevelsToBitmap` delegates to `DefRepLevelsToListInfo`, whose loop skips `repLevels[idx] > info.RepLevel` and treats equality as a no-write continuation, so both the scalar and the one-level-deeper vector leaf emit exactly one bitmap slot per struct element and skip empty/null outer-list rows identically; `TestStructArrayLeafPruningWithNulls:486` pins this with an exact expected bitmap rather than a row count.
- `struct_field_reader.go:75-83` — terminating on `len(field.Children) == 0` is required, not stylistic, because `SchemaField.IsLeaf()` tests `ColIndex != -1` and zero-initialised group nodes report leaf with a bogus `ColIndex` of 0; `getReader` guards identically, and `TestCollectSubFieldLeaves:195-197` asserts each resolved leaf's parquet path contains the sub-field name.
- Blast radius is contained: `NewStructFieldReader` has a single caller (`util.go:204`), no path reads a struct `FieldReader`'s `columnReader.Field()`, and the two type-sniffing uses (`field_reader.go:74`, `:199-204`) sit on the non-struct constructor path; memory accounting also improves, since `reader.go:63-66` divides the 64 MB budget by field count assuming one leaf set per reader — an assumption the old `GetColumn` path violated.

liliu-z (commented):
/lgtm
/approve

Issue #52052, opened 2026-07-30
[Bug]: [2.6] Configuration views can expose sensitive values


XuanYang-cn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 2.6.x
- Deployment mode(standalone or cluster): standalone and cluster
- MQ type(rocksmq, pulsar or kafka): any
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): any
- CPU/Memory: N/A
- GPU: N/A
- Others: management/WebUI HTTP endpoints on port 9091
```

### Current Behavior

Milvus 2.6 uses separate key-name denylists for configuration HTTP responses.

`Manager.GetConfigsView()` returns plaintext values with source annotations. The Proxy cluster configuration route applies a narrow second masking pass, while `/management/config/get` uses a different substring list. The rules have drifted, so credential-related built-ins and dynamic secret names can be returned in plaintext.

Examples include access-key identifiers, usernames, Pulsar auth parameters, OTLP headers, and dynamic keys containing auth token, client secret, private-key passphrase, or credential.

### Expected Behavior

Configuration HTTP views should use one shared sensitive-key classifier. The cluster and hook views should return `*****` for sensitive values, and `/management/config/get` should deny sensitive-key reads. Internal raw configuration APIs should keep returning plaintext for trusted callers.

### Steps To Reproduce

```markdown
1. On Milvus 2.6, add a runtime configuration such as dynamic.authToken=plain-text-sensitive-value.
2. Request GET /api/v1/_cluster/configs on port 9091.
3. Observe that the dynamic auth-token value is returned in plaintext.
4. Request /management/config/get for an unmatched credential-related key such as minio.accessKeyID or kafka.saslUsername.
5. Observe the inconsistent protection caused by the separate denylist.
```

### Anything else?

- Broader management-plane hardening: #49846
- Master implementation being tracked: #51956
- Original WebUI masking report: #44065
- Closed draft 2.6 PR: #52054
- Keep this issue open until a PR carrying the fix to the `2.6` branch exists; merging #51956 into master alone does not resolve it.

PR #52054, opened 2026-07-30
fix: [2.6] redact sensitive configuration values

XuanYang-cn:
Configuration views in 2.6 rely on separate key-name denylists, allowing
credential-related and dynamic secret keys to be returned in plaintext.

Centralize sensitive-key classification in the config manager and apply the
same ***** mask to cluster and hook configuration views. Reuse the classifier
to deny sensitive management config reads, while keeping raw configuration
APIs unchanged and leaving the Proxy response helper serialization-only.

Add regression coverage for 34 built-in sensitive keys, four dynamic secret
patterns, raw API behavior, configuration sources and runtime overlays,
management reads, Proxy serialization, and hook-view compatibility.

See also: #52052

### DIFF
--- a/internal/coordinator/restful_mgr_routes.go
+++ b/internal/coordinator/restful_mgr_routes.go
@@ -1443,10 +1443,7 @@ func (s *mixCoordImpl) HandleGetConfig(writer http.ResponseWriter, request *http
 		if key == "" {
 			continue
 		}
-		// Redact sensitive config keys (passwords, secrets, tokens).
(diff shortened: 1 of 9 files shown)


XuanYang-cn:
Closing this draft because master PR #51956 is the broader implementation and this 2.6 change would duplicate that work. Keeping #52052 open to track the release branch. This issue should close only after a PR carrying the fix to 2.6 exists; merging #51956 into master alone is not sufficient.

Issue #52056, opened 2026-07-30
[Enhancement] Accelerate batch WAL reopen during MQ type switching


tinswzy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Improve WAL reopen scheduling when multiple PChannels become unavailable at the same time.

During an MQ type switch, all PChannels may report assignment errors and require WAL reopen almost simultaneously. Currently, `AssignmentDiscover` handles each report synchronously and calls `MarkAsUnavailable` with only one PChannel.

Although the balancer supports marking multiple PChannels unavailable in one operation and opens different PChannels concurrently, processing reports one by one may split them into multiple balance rounds. WAL recovery and reopen operations therefore start at different times, introducing noticeable latency when a deployment has many PChannels.

The flush checkpoint wait also uses a fixed one-second polling interval, which adds unnecessary delay for PChannels that finish recovery quickly.

Proposed improvements:

1. Buffer assignment error reports for a short interval.
2. Deduplicate them by PChannel and retain the latest term.
3. Mark all collected PChannels unavailable in one batch.
4. Start WAL recovery and reopen for these PChannels concurrently.
5. Check the flush checkpoint using a short initial interval with exponential backoff, capped at one second.

### Why is this needed?

MQ type switching commonly causes many PChannels to reopen together. Sequentially triggering balance rounds makes the total switch time grow with the number of PChannels, even though the PChannels are independent and can recover concurrently.

In a local test with 16 PChannels, batching unavailable reports and accelerating flush checkpoint polling reduced bidirectional MQ switching time from around 20 seconds to approximately 7–9 seconds.

This optimization is independent of transparent cross-WAL scanner switching. It improves general scheduling and recovery efficiency whenever multiple WAL instances need to reopen simultaneously.

### Anything else?

The per-PChannel open/reopen logic is already concurrent once the PChannels are included in the same balance result. The main optimization is batching upstream assignment error reports so they enter the same balance round.

PR #52057, opened 2026-07-30
enhance: accelerate batch WAL reopen during MQ switching

tinswzy:
issue: #52056

## What was changed

- Batch assignment error reports from the same burst before calling `MarkAsUnavailable`.
- Deduplicate reported PChannels by name and retain the latest term.
- Let PChannels from the same burst enter one balance round so their WAL recovery and reopen operations can run concurrently.
- Replace the fixed one-second flush checkpoint ticker with exponential backoff from 20 ms to 1 second while retaining a one-minute overall timeout.

## Why

An MQ type switch can make many PChannels report assignment errors at nearly the same time. Handling every report synchronously may split independent PChannels into multiple balance rounds, making their WAL reopen operations start sequentially and adding noticeable switching latency.

## Test coverage

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/streamingcoord/server/service/discover ./internal/streamingnode/server/wal/adaptor`
- `go test -race -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/streamingcoord/server/service/discover`
- `make build-go`
- Local standalone MQ switch with 16 PChannels and concurrent writes:
  - RocksMQ to Woodpecker: 8.568 seconds, all 603 inserted PKs matched.
  - Woodpecker to RocksMQ: 7.424 seconds, all 1,143 inserted PKs matched.


czs007:

## Overview

This PR's batching of assignment-error reports is sound in its concurrency structure, but the batched failure path, a newly-live flush timeout, and test initialization need attention.

---

## High

**`internal/streamingcoord/server/service/discover/discover_server.go:170` — batched `MarkAsUnavailable` is all-or-nothing, and a failed batch is never re-driven**

The PR replaces the per-report `MarkAsUnavailable` call with a 10ms-batched call, but the downstream path is all-or-nothing: `channel/manager.go:513-521` returns `ErrChannelNotExist` from inside the per-channel loop before anything is persisted, and `manager.go:523` propagates a `SavePChannels` failure for the whole slice. Here at `:170` the error is only logged at Warn. Before this change a failure lost one channel's mark; now it loses every channel in the burst.

*Failure:* 16 pchannels of one streamingnode report within the same 10ms window and etcd flaps during `SavePChannels` (or one channel name is absent from `cm.channels`) → all 16 UNAVAILABLE marks are lost, no rebalance is triggered, and the client's retries are silently dropped by `assignment/discoverer.go:126` (`Term <= lastErrorReportedTerm[name]`) because the term only advances via the rebalance that the lost mark was supposed to cause — so those pchannels stay ASSIGNED-but-unwritable until the assignment stream is rebuilt.

*Suggestion:* Don't swallow the error at `:170`. Retry the whole batch, or fall back to per-channel marking so that one bad channel doesn't take healthy ones down with it.

*(raised by czs007)*

---

## Medium

**`internal/streamingnode/server/wal/adaptor/opener.go:351` — the one-minute flush timeout goes from dead code to genuinely live**

Previously `case <-time.After(defaultWALSwitchFlushTimeout)` sat inside the for-select, so it was re-armed on every iteration while the 1s ticker always fired first — the branch was unreachable, and the effective semantics were "wait until flush completes or ctx is cancelled." Hoisting it into `timeoutTimer` outside the loop is the right shape, but it activates a timeout that never fired before.

*Failure:* a pchannel with a large growing-segment backlog (or slow object storage) needs more than 60s for the flusher checkpoint to reach `AlterWALTs` → `handleAlterWALFlushingStage` returns `timeout waiting for flush completion` at `:383`, `openRWWAL`'s `defer resources.Close()` (`:201`) tears down the flusher and recovery storage, and `wal_lifetime.go:158-167` marks the channel unavailable; convergence then requires several 60s-abort + rebalance rounds, with the pchannel unwritable throughout.

*Suggestion:* Describe this in the PR body as a behavior change rather than "retaining a one-minute overall timeout," and re-check whether 60s is sufficient for large backlogs — consider making it configurable or scaling it with backlog size.

*(raised by czs007)*

---

## Low

**`internal/metastore/kv/streamingcoord/kv_catalog.go:171` — this PR is the first caller that can push `MarkAsUnavailable` past `maxEtcdTxnNum`**

`SavePChannels` splits into non-atomic `MultiSave` calls at `metastore.maxEtcdTxnNum` (default 64, `paramtable/service_param.go:570`). That is a pre-existing property of the catalog layer shared with other multi-channel callers, but before this PR `MarkAsUnavailable`'s only caller passed exactly one channel; it can now pass up to `assignmentErrorQueueCapacity = 128`, so the split becomes reachable for the first time.

*Failure:* 128 pchannels configured and 65–128 marks land in one batch, with `MultiSave` failing on the second chunk → the first 64 are already UNAVAILABLE in etcd while `cm.channels` and `cm.version` are not updated at all (`manager.go:544-548` is skipped), so the coordinator keeps serving those channels as ASSIGNED, and the etcd-only marks only take effect after a streamingcoord restart, triggering a reassignment with no live trigger behind it.

*Suggestion:* No separate atomicity work is needed — handling the error at `discover_server.go:170` as described above covers this case too.

*(raised by czs007)*

**`internal/streamingcoord/server/service/discover/discover_server_test.go:107` — the two new tests implicitly depend on `TestAssignmentDiscover` running first**

The package has no `TestMain`, and `resource.Resource()` is just `r.Load()` (`resource.go:79`), which is nil until initialized. Both new tests call `NewAssignmentDiscoverServer`, which dereferences `resource.Resource().Logger()` at `discover_server.go:38`. Only `TestAssignmentDiscover` calls `resource.InitForTest`, so whole-package runs happen to pass on source ordering alone.

*Failure:* running `go test -run TestMarkUnavailableLoopBatchesAndDeduplicates ./internal/streamingcoord/server/service/discover/` (or any future reordering / skip of `TestAssignmentDiscover`) → SIGSEGV nil-pointer panic at `resource.go:125` → `discover_server.go:38` → `discover_server_test.go:107`, taking down the package's whole test binary. This was reproduced against PR head `70f8d269e1`.

*Suggestion:* Add one `resource.InitForTest()` line to each of the two new tests.

*(raised by czs007)*

czs007:

- Candidate fix (diff on top of this PR head): https://github.com/milvus-io/milvus/compare/70f8d269e17c73cc4e6b65cd1bcd7ec394c6970a...czs007:team2-fix/Ta8734aaeaf?expand=1

czs007:

Verified:
- Retry/backoff path — the previously reported backoff-bypass concern was re-examined and independently confirmed as the sole medium-severity finding, with no further defects found around it.
- Scope of the change — a second independent reviewer swept the same diff and reported no additional issues beyond that one finding.

czs007:

Verified:
- `internal/streamingcoord/server/service/discover/discover_server.go:79-82` — each `ReportError` now issues `MarkAsUnavailable` with exactly one `PChannelInfo`, so a failure can only affect that single report.
- `internal/streamingcoord/server/balancer/channel/manager.go:513-525` — the internal all-or-nothing handling can no longer discard unrelated channels from a burst, since every call it receives is single-channel.
- `internal/streamingcoord/server/balancer/balancer_impl.go:683-700` — concurrency is handled as independent per-channel balance operations, replacing the removed 10 ms batch-and-deduplication path.
- `internal/streamingcoord/server/balancer/balancer_test.go:372-395` — verifies that a second report and assignment proceed while the first assignment is still in flight.

## Re-review of `bb0daef8f68c`

No earlier findings from this mechanism were on record for this commit, so there is nothing to adjudicate. Reviewers additionally confirmed that the previously flagged high-severity behavior in the error-report path is resolved in the current state of the branch (see the Verified list above).

No new findings — including non-blocking, pre-existing, or disputed items — survived this round of verification.

Issue #52067, opened 2026-07-31
bug: generic query reduction and empty-response correctness gaps


MrPresent-Han:
### Scope

This issue consolidates generic QueryNode / Delegator / Proxy correctness defects found while reviewing REST work. They affect SDK and gRPC clients too, so they are intentionally outside REST API request mapping.

### 1. Grouped reduction ignores maxOutputSize

`BuildQueryReducePipeline` receives `quotaAndLimits.limits.maxOutputSize`, and plain/non-grouped order-by reducers enforce it. Both grouped branches create `NewDeduplicateByGroupsOperator` without passing that budget, while `GroupAggReducer` only caps `common.groupBy.maxGroups` (default 100000).

To retain correct partial aggregates, QueryNode and Delegator must retain all groups before Proxy merges them. Wide VarChar group keys can then allocate and serialize far beyond the configured 100 MB budget even for `LIMIT 1`.

Expected: preserve complete partial groups, but enforce a byte/memory budget at each grouped reduce stage and return an explicit error on overflow. Never use the user page limit to truncate partial groups.

### 2. GROUP BY with prefix ORDER BY can return partial aggregates

```sql
GROUP BY (a, b)
ORDER BY a
LIMIT 1
```

This is valid SQL: the order is unspecified among groups with equal `a`, but every returned aggregate must be complete. The generic QueryNode/Delegator pipeline currently applies `OrderByLimit(topK)` before Proxy merges shard contributions. Nodes can retain different tied groups; Proxy may select a group without receiving another node partial contribution, producing partial `count`, `sum`, or `avg`.

Expected: fix the distributed reduction without narrowing the valid API. Possible mechanisms include a consistent invisible group-key tie-breaker, retaining all boundary ties, or deferring the cut until the global merge. REST should forward the valid combination with the same semantics as existing SDK/gRPC requests. Related: #50925.

### 3. Empty Query/Search responses leak StructArray child field names

`translateOutputFields` expands requested StructArray `items` into internal child names such as `items[text]` and `items[embedding]`. `reconstructStructFieldData` collapses them back only if `FieldsData` is non-empty. A zero-hit response therefore exposes internal child names in `OutputFields`, while an equivalent response with hits returns `items`.

Expected: for empty results, map transformed child output names back to parent StructArray names using the schema, preserving order and deduplicating parents. Aggregate names such as `sum(price)` must remain intact.

MrPresent-Han:
[...]

PR #52071, opened 2026-07-31
enhance: add REST query group by and search order by

MrPresent-Han:
## Summary

- Add `groupByFields` to the REST v2 query request and forward it as `group_by_fields`.
- Add `orderByFields` to the REST v2 search request and forward it as `order_by_fields`.
- Forward `groupByFields` and query `orderByFields` together, matching the existing query API semantics.
- Reject only newly introduced ambiguous or unsupported REST combinations.

## Compatibility

- Preserve the master behavior for global lone `count(*)`: REST omits limit and an offset remains ignored as before.
- Preserve master behavior for mixed global aggregates containing `count(*)`: REST forwards the default limit and Proxy rejects count pagination.
- Preserve the legacy `searchParams.order_by_fields` passthrough, including its existing interaction with `searchAggregation`.
- Generic reduce, proxy pipeline, and response reconstruction behavior are unchanged; follow-up items are tracked in #52067.

## Tests

- `source scripts/setenv.sh && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/distributed/proxy/httpserver`

related: #41675, #36380

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -1411,11 +1411,18 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 	if httpReq.Offset > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
 	}
-	if httpReq.Limit > 0 && !matchCountRule(httpReq.OutputFields) {
+	// Keep the legacy global count(*) behavior: REST never forwards its limit
+	// for a lone count(*). GroupByFields is new, so grouped counts still forward
+	// limit and offset to paginate groups.
+	isLegacyGlobalCount := len(httpReq.GroupByFields) == 0 && matchCountRule(httpReq.OutputFields)
+	if httpReq.Limit > 0 && !isLegacyGlobalCount {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.LimitKey, Value: strconv.FormatInt(int64(httpReq.Limit), 10)})
 	}
 	if len(httpReq.OrderByFields) > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OrderByFieldsKey, Value: strings.Join(httpReq.OrderByFields, ",")})
+	}
+	if len(httpReq.GroupByFields) > 0 {
+		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.GroupByFieldsKey, Value: strings.Join(httpReq.GroupByFields, ",")})
 	}
 	resp, err := wrapperProxyWithLimit(ctx, c, req, h.checkAuth, false, "/milvus.proto.milvus.MilvusService/Query", true, h.proxy, func(reqCtx context.Context, req any) (interface{}, error) {
 		return h.proxy.Query(reqCtx, req.(*milvuspb.QueryRequest))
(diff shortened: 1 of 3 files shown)


internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1411,12 +1411,19 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
...
+	// limit and offset to paginate groups.
+	isLegacyGlobalCount := len(httpReq.GroupByFields) == 0 && matchCountRule(httpReq.OutputFields)
+	if httpReq.Limit > 0 && !isLegacyGlobalCount {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.LimitKey, Value: strconv.FormatInt(int64(httpReq.Limit), 10)})
 	}
 	if len(httpReq.OrderByFields) > 0 {
 		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.OrderByFieldsKey, Value: strings.Join(httpReq.OrderByFields, ",")})
 	}
+	if len(httpReq.GroupByFields) > 0 {
+		req.QueryParams = append(req.QueryParams, &commonpb.KeyValuePair{Key: proxy.GroupByFieldsKey, Value: strings.Join(httpReq.GroupByFields, ",")})
```
buqian-zilliz, line 1425: **[P1] Do not truncate grouped partials before the global merge**

Forwarding `groupByFields` makes limited grouped aggregation reachable from REST, but the distributed reduction path applies the request's `topK` before Proxy performs the global group merge. The no-order branch drops new groups after `topK` at each QueryNode/Delegator, while the group-and-order branch applies a local `OrderByLimit(topK)`; for `GROUP BY (a,b) ORDER BY a LIMIT 1`, tied `(a,b)` groups can therefore be retained differently on different workers. Proxy may then return a count or sum missing contributions from workers that pruned that group. Retain complete partial groups until the global merge with a separate memory bound, or reject the unsafe REST combinations until the reducer protocol is fixed.

buqian-zilliz, line 1425: **[P1] Enforce the output byte budget for grouped reductions**

This new REST entry point reaches grouped reducers that bypass `quotaAndLimits.limits.maxOutputSize`. QueryNode and Delegator pass the byte budget into `BuildQueryReducePipeline`, but both GROUP BY branches ignore it; GROUP BY with ORDER BY can retain up to `common.groupBy.maxGroups` (100,000 by default) before sorting. Wide VarChar keys can therefore allocate and serialize far beyond the configured 100 MiB even when the final page is tiny. Plumb byte accounting through grouped reduction and fail explicitly on overflow; using the page limit to truncate partial groups would preserve the correctness defect above.

MrPresent-Han: Thanks for the detailed analysis. This is a pre-existing generic QueryNode/Delegator/Proxy grouped-reduction problem, not specific to this REST mapping. It affects existing SDK/gRPC requests that use `group_by_fields`, including the local group-limit and partial-group case described here. We have recorded the follow-up in #52067. The fix will be made in the generic reduction pipeline so partial groups are retained through the global merge with an independent memory bound. We will not introduce a REST-only rejection here, because that would narrow valid semantics and diverge from the existing APIs.

MrPresent-Han: Thanks. The missing `maxOutputSize` enforcement is likewise a generic grouped-reduction issue. #52067 tracks it. We will fix it at the common reducer level by adding byte accounting and failing explicitly on overflow while retaining complete partial groups; using the user page limit as the cutoff is not correct. This REST PR remains a request-mapping change.

congqixia (approved):
/lgtm

Issue #50263, opened 2026-06-03
[Feature]: Support Row Level Security


zhengbuqian:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Allow defining of policies to enforce row level access control

PR #52073, opened 2026-07-31
feat: [RLS2] synchronize row-level security metadata

aoiasd:
relate: #50263
design doc PR: #53173

## Summary
Adds bulk RLS metadata loading, versioned Proxy snapshots, notifications, TTL reconciliation, and fail-closed refresh.


internal/rootcoord/root_coord.go
```diff
@@ -3535,14 +3535,19 @@ func (c *Core) CreateRowPolicy(ctx context.Context, req *milvuspb.CreateRowPolic
...
-	_, err = c.mutateRLSPolicyWithCollectionLock(ctx, req.GetDbName(), req.GetCollectionName(), func() (UniqueID, error) {
+	collectionID, err := c.mutateRLSPolicyWithCollectionLock(ctx, req.GetDbName(), req.GetCollectionName(), func() (UniqueID, error) {
 		return c.meta.CreateRLSPolicy(ctx, req, policyID)
 	})
 	if err != nil {
 		logger.Warn(ctx, "failed to create RLS policy", mlog.Err(err))
 		metrics.RootCoordDDLReqCounter.WithLabelValues(method, metrics.FailLabel).Inc()
 		return merr.Status(err), nil
 	}
+	if err := c.notifyRLSMetaChanged(ctx, commonpb.MsgType_CreateRowPolicy, req.GetDbName(), req.GetCollectionName(), collectionID); err != nil {
```
buqian-zilliz: **[P1] Make post-commit RLS propagation fail closed**

The RLS catalog mutation has already committed before this call, but the fan-out still runs on the caller's context and any TSO or proxy failure is returned as if the mutation failed. If the deadline expires or one proxy cannot refresh, that proxy keeps its previous `lastSuccessfulRefresh`, so the TTL check continues treating the stale snapshot as valid for up to the default 3600 seconds. A policy tightening or tag removal can therefore be committed in RootCoord while requests are still authorized by the old rules. Make post-commit propagation durable and record the incoming version as a required freshness floor so failed refreshes fail closed.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,124 @@
...
+		dbName = resp.GetDbName()
+	}
+	if resp.GetCollectionName() != "" {
+		collectionName = resp.GetCollectionName()
+	}
+
+	refreshedAt := time.Now()
+	if refreshPolicies {
+		updated := m.setRLSPolicySnapshot(dbName, collectionID, policySnapshot{
+			Version:     int64(version),
```
buqian-zilliz, line 103: **[P2] Bind snapshot ordering to the metadata revision**

`version` is allocated before `GetRLSMetadata` and is not a revision of the response. A lower-version refresh can read newer metadata, then a delayed higher-version notification that read older metadata can overwrite it; the inverse ordering can also discard the newer response as stale. If the later change's notification was missed, revoked access can remain cached and freshly timestamped for nearly a full TTL. Return an authoritative metadata revision captured atomically with the RootCoord snapshot and compare that revision here.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,124 @@
...
+		return merr.WrapErrServiceInternalMsg("failed to refresh RLS snapshots without manager or coord client")
+	}
+	if collectionID == 0 {
+		return merr.WrapErrServiceInternalMsg("failed to refresh RLS snapshots with empty collection id")
+	}
+	if !refreshPolicies && !refreshPrincipalTags {
+		return merr.WrapErrServiceInternalMsg("failed to refresh RLS snapshots without a target")
+	}
+
+	resp, err := coord.GetRLSMetadata(ctx, &rootcoordpb.GetRLSMetadataRequest{
```
buqian-zilliz, line 80: **[P2] Avoid returning principal tags for policy-only refreshes**

Both policy-only and principal-only refreshes call this combined RPC. A single policy mutation therefore makes every Proxy serialize and download the entire principal-tag snapshot and then discard it; the default limits allow about 55 MiB of raw tag keys and values per collection before protobuf and map overhead. RootCoord also deep-clones these maps before marshaling, so fan-out cost scales with proxy count. Add a requested snapshot kind or field mask, separate RPCs, or send the changed delta.

internal/proxy/rls/reconciler.go
```diff
@@ -0,0 +1,160 @@
...
+			commonpbutil.WithSourceID(paramtable.GetNodeID()),
+		),
+		AllowUnavailable: false,
+	})
+	if err := merr.CheckRPCCall(resp, err); err != nil {
+		return nil, merr.Wrap(err, "failed to list collections for RLS snapshot reconciliation")
+	}
+
+	targets := make([]reconciliationTarget, 0)
+	for _, dbCollections := range resp.GetDbCollections() {
```
buqian-zilliz: **[P2] Batch reconciliation across collections**

Every collection without state is treated as due, so each Proxy issues one `GetRLSMetadata` RPC per collection at startup and again when the common TTL expires, including collections with no RLS metadata. With the default 65,536 general-capacity ceiling, ten proxies and one-partition/one-shard collections can generate roughly 655,000 full metadata RPCs in one synchronized cycle, with no batching, jitter, or rate limit. Reconcile only RLS-bearing or known collections, or add a paginated bulk/version endpoint.

internal/proxy/meta_cache.go
```diff
@@ -1534,6 +1535,9 @@ func (m *MetaCache) RemoveDatabase(ctx context.Context, database string) {
 	evicted := 0
 	for start := 0; start < len(ids); start += removeBatch {
 		end := min(start+removeBatch, len(ids))
+		for _, id := range ids[start:end] {
+			rls.RemoveCollection(ctx, id)
```
buqian-zilliz: **[P2] Make RLS cleanup follow the authoritative collection lifecycle**

These IDs come from MetaCache's lazy `nameIdx`, while the RLS reconciler creates manager state for every collection returned by `ShowCollectionIDs`. Dropping a database therefore removes RLS state only for collections that happened to be described through MetaCache; all other policy and tag snapshots remain indefinitely because the manager discards `dbName` and reconciliation never prunes missing IDs. An in-flight refresh can likewise reinsert an ID after `RemoveCollection`. Track database or lifecycle generations in the RLS manager and prune or tombstone states absent from the authoritative collection set.

internal/proxy/proxy.go
```diff
@@ -277,6 +284,14 @@ func (node *Proxy) Start() error {
 	}
 	mlog.Debug(node.ctx, "start id allocator done", mlog.String("role", typeutil.ProxyRole))
 
+	node.wg.Add(1)
+	go func() {
+		defer node.wg.Done()
+		if err := rls.DefaultManager().RunReconciler(node.ctx, node.mixCoord, node.tsoAllocator.AllocOne); err != nil {
```
buqian-zilliz: **[P1] Close the registration gap before bootstrap reconciliation**

`RunReconciler` performs its bootstrap sweep before this Proxy is registered: `Server.start` calls `Proxy.Start` and only then `Proxy.Register`. If a policy tightening or tag revocation commits after this goroutine has fetched a collection but before RootCoord's watcher adds the new Proxy, the notification omits it. The old snapshot has just been stamped with a fresh `RefreshedAt`, so the request-path TTL check will trust it for up to the default hour. Establish registration/notification visibility before taking the authoritative bootstrap snapshot, and do not mark the Proxy healthy until that barrier completes.

internal/rootcoord/ddl_callbacks_rls.go
```diff
@@ -238,31 +240,74 @@ func unmarshalRLSPrincipalMessage(header *message.AlterRLSMetadataMessageHeader,
...
+	if c.tsoAllocator == nil || c.proxyClientManager == nil {
+		return
+	}
+	coll, err := c.meta.GetCollectionByIDWithMaxTs(ctx, collectionID)
+	if err != nil {
+		mlog.Warn(ctx, "skip best-effort RLS metadata notification because collection metadata is unavailable",
+			mlog.FieldCollectionID(collectionID), mlog.Err(err))
+		return
+	}
+	ctx = context.WithoutCancel(ctx)
```
buqian-zilliz: **[P2] Bound detached notifications to the coordinator lifecycle**

`context.WithoutCancel` removes both cancellation and the deadline before one untracked goroutine is launched for every committed RLS mutation. `InvalidateCollectionMetaCache` waits for every Proxy, and each client can spend roughly 53 seconds retrying by default (or retry forever when `maxAttempts` is zero), so an outage plus sustained mutations can accumulate detached fan-out work. `Core.Stop` cancels `c.ctx` and waits `c.wg`, neither of which covers these goroutines, so they can continue using the allocator and clients after shutdown. Route notifications through a bounded/coalescing worker owned by the Core lifecycle and give each fan-out an explicit timeout.

internal/rootcoord/root_coord.go
buqian-zilliz: This issue still exists. The callback now detaches the fan-out and logs failures, but a failed or skipped notification still records no required minimum version or invalidation state, so the previous snapshot remains trusted until proxy.rls.metaRefreshInterval expires.

internal/proxy/rls/snapshot.go
buqian-zilliz: This issue still exists. GetRLSMetadata still returns no authoritative metadata revision, while refresh ordering is based on a TSO allocated before the read, so a larger version can still carry older metadata.

internal/proxy/rls/snapshot.go
buqian-zilliz: This issue still exists. GetRLSMetadata has no requested snapshot kind or field mask, so every policy-only or principal-only refresh still transfers and clones both complete snapshots.

internal/proxy/rls/reconciler.go
buqian-zilliz: This issue still exists. Each Proxy still enumerates all collections and performs one full GetRLSMetadata RPC per due collection without batching, RLS-only filtering, jitter, or rate limiting.

internal/proxy/meta_cache.go
buqian-zilliz: This issue still exists. DropDatabase still derives IDs only from MetaCache, while the RLS manager neither tracks database membership nor prunes or tombstones states, so uncached or in-flight refreshed collections can remain or be reinserted.

internal/proxy/proxy.go
aoiasd: Thanks. The consistency target here is best-effort notification with TTL-bounded eventual convergence rather than strict startup linearizability. We will move the bootstrap reconciler startup until after the Proxy session is registered, so the normal path registers the Proxy before its first sweep.

We intentionally will not block Proxy Healthy on a watcher-visibility/bootstrap barrier. A residual race caused by watcher propagation delay or notification failure may require the snapshot TTL to expire before convergence, which is acceptable for the current design.

internal/proxy/impl.go
```diff
@@ -151,6 +156,21 @@ func (node *Proxy) InvalidateCollectionMetaCache(ctx context.Context, request *p
 		}
 	}
 
+	switch msgType {
+	case rlsutil.MsgTypeCreateRowPolicy,
+		rlsutil.MsgTypeUpdateRowPolicy,
+		rlsutil.MsgTypeDropRowPolicy,
+		rlsutil.MsgTypeSetRLSPrincipalTags,
+		rlsutil.MsgTypeDeleteRLSPrincipalTags:
+		rls.RemoveCollection(ctx, collectionID)
```
zhengbuqian, line 187: we invalidates the entire RLS cache when any policy/principal got update, right? this may be too frequent, as principal is set per end user, principal creation/update will be frequent operations, and entire cache re-initialization will be too frequent.

aoiasd:
@Mergifyio refresh

internal/rootcoord/ddl_callbacks_rls.go
```diff
@@ -261,21 +281,30 @@ func (c *DDLCallback) alterRLSMetadataV2AckCallback(ctx context.Context, result
 		if err != nil {
 			return err
 		}
-		return c.meta.ApplyAlterRLSPrincipal(ctx, principal)
+		err = c.meta.ApplyAlterRLSPrincipal(ctx, principal)
```
buqian-zilliz, line 287: **[P1] Propagate principal persistence failures from the ACK callback**

`principal, err := unmarshalRLSPrincipalMessage(...)` declares a case-clause-local `err`. The subsequent assignment from `ApplyAlterRLSPrincipal` updates that shadow, so the outer `err` checked after the switch remains nil. If `SaveRLSPrincipal` fails, the callback still expires Proxy caches and acknowledges the tag mutation even though it was not persisted. Assign the unmarshalling result to the outer error variable or return the apply error directly.

internal/rootcoord/ddl_callbacks_rls.go
```diff
@@ -261,21 +281,30 @@ func (c *DDLCallback) alterRLSMetadataV2AckCallback(ctx context.Context, result
...
 	default:
 		return merr.WrapErrServiceInternalMsg("alter RLS metadata message has no metadata")
 	}
+	if err != nil {
+		return err
+	}
+	return c.ExpireCaches(ctx, header)
```
buqian-zilliz, line 298: **[P1] Handle RLS headers created before cache expirations existed**

`cache_expirations` was added to existing WAL message headers, so pending or CDC-replayed `AlterRLSMetadata` and `DropRLSMetadata` messages created before this change legitimately return nil from `GetCacheExpirations()`. Both callbacks pass that nil value to `ExpireCaches`, which dereferences `cacheExpirations.CacheExpirations` and panics after applying the metadata mutation. Detect the missing field and synthesize the collection-scoped RLS invalidation from the existing header and body before calling the shared helper.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (m *manager) ensureFreshMetadata(ctx context.Context, collectionID UniqueID) error {
+	if m == nil || collectionID == 0 {
+		return merr.WrapErrServiceInternalMsg("failed to validate RLS metadata freshness with invalid manager or collection id")
+	}
+	refreshTTL := paramtable.Get().ProxyCfg.RLSMetaRefreshInterval.GetAsDuration(time.Second)
+	if refreshTTL <= 0 {
+		return merr.WrapErrServiceInternalMsg("failed to validate RLS metadata freshness with invalid TTL %s", refreshTTL)
+	}
+	m.refreshLocks.RLock(collectionID)
```
buqian-zilliz: **[P2] Do not hold invalidation locks across metadata RPCs**

The lazy refresh path acquires the per-collection read lock before entering singleflight and retains it through TSO allocation and the `GetRLSMetadata` RPC; the targeted refresh path does the same. `RemoveCollection` needs the corresponding write lock, while RLS ACK callbacks now synchronously wait for every Proxy invalidation. A slow or retrying metadata RPC can therefore block a restrictive policy mutation or collection cleanup. The code already captures a state incarnation that can be detached safely, so use that fencing design instead of holding the lock across network I/O.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,276 @@
...
+		snapshot.RefreshedAt = time.Now()
+	}
+	state.principalTagVersion = snapshot.Version
+	state.principalTagLastSuccessfulRefresh = snapshot.RefreshedAt
+	state.principalTags = map[string]map[string]rlsutil.TagValue{}
+	for principalName, tags := range snapshot.PrincipalTags {
+		if principalName == "" {
+			continue
+		}
+		state.principalTags[principalName] = clonePrincipalTags(tags)
```
buqian-zilliz: **[P2] Avoid cloning the newly decoded principal maps**

`refreshSnapshotsUnlocked` creates private principal-tag maps while decoding the RPC response, but `setRLSPrincipalTagsSnapshot` immediately allocates and copies every nested map again before storing them. At the configured limits this repeats hashing and allocation for up to 50,000 tag entries and 1,000 maps per collection while the original decoded structure is still live. No other owner retains that structure, so transfer it into the collection state or make the setter explicitly consume the snapshot.

internal/distributed/mixcoord/client/client.go
```diff
@@ -2232,6 +2232,17 @@ func (c *Client) CommitBackfillResult(ctx context.Context, req *datapb.CommitBac
...
 }
 
+func (c *Client) GetRLSMetadata(ctx context.Context, req *rootcoordpb.GetRLSMetadataRequest, opts ...grpc.CallOption) (*rootcoordpb.GetRLSMetadataResponse, error) {
+	req = typeutil.Clone(req)
+	commonpbutil.UpdateMsgBase(
+		req.GetBase(),
+		commonpbutil.FillMsgBaseFromClient(paramtable.GetNodeID(), commonpbutil.WithTargetID(c.grpcClient.GetNodeID())),
+	)
+	return wrapGrpcCall(ctx, c, func(client MixCoordClient) (*rootcoordpb.GetRLSMetadataResponse, error) {
+		return client.GetRLSMetadata(ctx, req)
```
buqian-zilliz: **[P3] Forward the GetRLSMetadata call options**

`GetRLSMetadata` accepts `opts ...grpc.CallOption`, but its closure invokes `client.GetRLSMetadata(ctx, req)` without forwarding them. Caller-supplied readiness, compression, credentials, or message-size behavior is therefore silently ignored. Pass `opts...` to preserve the generated gRPC client contract.

internal/rootcoord/root_coord.go
buqian-zilliz: This issue still exists. The synchronous callback retries ordinary fan-out failures, but a missing Proxy registration still makes the fan-out return success without recording any freshness requirement on that Proxy. A missed restrictive update can therefore remain trusted until the TTL expires.

internal/proxy/rls/snapshot.go
buqian-zilliz: This issue still exists. Although the RPC now supports `RLSMetadataKind`, every RLS mutation removes the entire collection state and the lazy refresh path always requests `RLS_METADATA_KIND_ALL`. A policy-only mutation therefore still causes every Proxy to reload all principal tags.

internal/rootcoord/ddl_callbacks_rls.go
aoiasd: These RLS WAL messages and `cache_expirations` are introduced together in this unreleased feature stack, so there is no released producer or persisted RLS message without this field to replay, including through CDC. All current policy/drop producers populate it; the only intentional empty case is creating a previously unseen principal, which is handled before `ExpireCaches`, so we intentionally do not add compatibility for this unreachable legacy format.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (m *manager) ensurePrincipalTags(ctx context.Context, collectionID UniqueID, principalName string) (map[string]rlsutil.TagValue, error) {
+	if m == nil || collectionID == 0 || principalName == "" {
+		return nil, merr.WrapErrPrivilegeNotPermitted("RLS principal is required")
+	}
+
+	key := principalKey{collectionID: collectionID, principalName: principalName}
+	coord, _, validateFreshness, _ := m.refreshDependencies()
+	if entry := m.getPrincipalTagsEntry(key); entry != nil {
```
buqian-zilliz, line 168: **[P2] Enforce the principal-cache TTL on lookup**

`metaRefreshInterval` is documented as the maximum principal-cache age and is refreshable, but this path returns any cached entry without comparing `refreshedAt` with the current TTL; expiry depends solely on a fixed ten-minute scanner. After a missed invalidation, a revoked tag can therefore remain effective for the configured TTL plus almost ten minutes, and lowering the TTL below ten minutes still permits nearly ten minutes of stale authorization data. Evict or reject expired entries during lookup, or schedule expiry against each entry's deadline, so the configured interval is an actual upper bound.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,236 @@
...
+			return nil, merr.WrapErrServiceUnavailableMsg("RLS collection %d was removed during principal refresh", collectionID)
+		}
+		finished := false
+		defer func() {
+			if !finished {
+				m.finishPrincipalRefresh(key, state, token, nil)
+			}
+		}()
+
+		resp, err := coord.GetRLSMetadata(refreshCtx, &rootcoordpb.GetRLSMetadataRequest{
```
buqian-zilliz, line 195: **[P2] Allow request cancellation during shared metadata refreshes**

The caller's `ctx` is checked only before and after synchronous `singleflight.Do`, while the actual `GetRLSMetadata` call uses the manager-lifetime `refreshCtx`. If the coordinator stalls, an expired request remains blocked until that shared call returns, and every request for the same key waits with it; the policy refresh path uses the same pattern for both TSO allocation and RPC. Keep the shared refresh independent if needed, but use a bounded refresh context and let each caller select between the shared result and its own `ctx.Done()`.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,236 @@
...
+				refreshedAt: time.Now(),
+				tags:        clonePrincipalTags(tags),
+			}
+		}
+		current := m.finishPrincipalRefresh(key, state, token, entry)
+		finished = true
+		if !current {
+			return nil, merr.WrapErrServiceUnavailableMsg("RLS principal %q metadata changed during refresh", principalName)
+		}
+		if tags == nil {
```
buqian-zilliz, line 231: **[P2] Cache empty tag results for tagless principals**

A principal with no stored tag binding is a normal state, but this branch returns an empty map without installing a cache entry. `singleflight` only coalesces concurrent misses, so after it completes every subsequent RLS request for the same tagless principal repeats a coordinator RPC and RootCoord scan; varying principal names bypasses coalescing entirely. Cache an explicit empty-tag result with the same TTL, and invalidate that key when its first tag binding is created instead of omitting the create invalidation.

internal/rootcoord/meta_table.go
```diff
@@ -3410,6 +3411,61 @@ func (mt *MetaTable) ListRLSPolicies(ctx context.Context, req *rlsutil.ListRowPo
...
 
+func (mt *MetaTable) GetRLSMetadata(ctx context.Context, collectionID int64, kind rootcoordpb.RLSMetadataKind, principalName string) (*model.RLSMetadata, error) {
+	if collectionID == 0 {
+		return nil, merr.WrapErrServiceInternalMsg("failed to get RLS metadata with empty collection id")
+	}
+
+	mt.ddLock.RLock()
+	defer mt.ddLock.RUnlock()
+
+	coll, err := mt.getLatestCollectionByIDInternal(ctx, collectionID, false)
```
buqian-zilliz: **[P2] Avoid cloning unrelated collection state for RLS reads**

Expanding `getLatestCollectionByIDInternal(..., false)` one level shows that it calls `filterUnavailablePartition`, which shallow-clones the RLS policy map and principal slice and allocates/scans the full partition list. This method then clones the requested RLS objects again. Every principal-cache miss therefore performs `O(partitions + policies + principals)` allocation and scanning while holding the global `ddLock` read lock, even when only one principal is requested. Use a lightweight locked lookup that checks collection availability without cloning unrelated state, then clone only the requested policy or principal data.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,495 @@
...
+
+type policySnapshot struct {
+	Version     int64
+	RefreshedAt time.Time
+	Policies    []*rlsutil.RowPolicy
+}
+
+type SnapshotVersionAllocator func(ctx context.Context) (uint64, error)
+
+type Manager interface {
```
buqian-zilliz: **[P3] Land the RLS cache with its production consumer**

No non-test production code calls `ensureFreshMetadata`, `ensurePrincipalTags`, or `RefreshPolicySnapshot`, and the exported `Manager` interface exposes no snapshot read operation. The live paths only initialize the singleton and invalidate entries in a map that production code never populates, so this PR adds the RPC, proto, configuration, cache, fan-out, and mock surfaces without changing request behavior. Either connect the cache to the RLS request path in this PR or introduce this infrastructure together with its first consumer.

internal/rootcoord/root_coord.go
buqian-zilliz: This issue still exists: invalidation still succeeds when no Proxy clients are registered, and no required metadata revision is recorded for a Proxy that missed the fan-out, so its stale RLS state can remain trusted until TTL reconciliation.

internal/proxy/rls/snapshot.go
aoiasd: This is intentional for the v1 consistency contract. Proxy invalidation is the primary path; principal TTL is a fallback implemented by the periodic scanner. We explicitly accept a stale window of the configured TTL plus at most one 10-minute scan interval, since RLS metadata mutations are expected to be infrequent. The design document describes periodic eviction rather than a strict per-lookup deadline, so no code change is planned here.

internal/proxy/rls/snapshot.go
aoiasd: Fixed in 7633fda8e6. Policy and principal lookups now wait through native `singleflight.Group.DoChan` and select against the individual request `ctx.Done()`. A canceled caller returns immediately, while the shared refresh continues on the manager lifecycle context and can populate the cache for other callers. We intentionally do not derive the synchronization lifetime from the gRPC request; manager shutdown cancels it.

internal/proxy/rls/snapshot.go
aoiasd: The missing binding is intentionally not negative-cached. A principal is a request-supplied identifier, not a Milvus entity, and no stored binding therefore evaluates with empty tags. First-time Set intentionally sends no invalidation because there is no positive cache entry to evict. Negative caching would either hide the first binding until TTL or require create fan-out, and arbitrary request principal names could grow the Proxy cache without the persisted-principal limit. We keep the current no-negative-cache contract.

internal/rootcoord/meta_table.go
aoiasd: Fixed in 7633fda8e6. `GetRLSMetadata` now checks the collection directly under `ddLock` and clones only the requested RLS policy/principal data. It no longer calls `getLatestCollectionByIDInternal`, so it avoids the unrelated partition scan and shallow collection copies. The remaining named-principal scan is the known deferred indexing optimization.

internal/proxy/rls/manager.go
aoiasd: This PR is intentionally the synchronization/cache foundation of the dependent RLS stack. It initializes and invalidates the manager; #52074 adds evaluation, and #52075 wires the first production consumers into Query, Search, Delete, Upsert, and related paths. Adding a token consumer here would cross the agreed cache -> evaluator -> enforcement split without changing final behavior. The dormant layer is safe to merge because requests do not rely on it until the descendants land, and the stack will merge in order.

internal/rootcoord/root_coord.go
aoiasd: RLS intentionally follows the existing collection/schema metadata invalidation contract. The shared ProxyClientManager treats an empty client set and disappeared or unimplemented clients as successful best-effort fan-out; returning an error for zero clients would block callbacks when no Proxy is legitimately running and still would not detect a partially missing Proxy. Missed RLS notification is repaired by request-time policy TTL refresh, with refresh failures failing closed, and principal TTL eviction plus lazy reload. We explicitly accept this rare eventual-convergence window in v1; strict freshness would require a generic Proxy membership and revision-floor protocol outside this PR.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	if coord == nil {
+		return nil, merr.WrapErrServiceInternalMsg("failed to refresh RLS principal tags without coord client")
+	}
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	cacheKey := principalRefreshKey(key)
+	resultCh := m.principalRefreshes.DoChan(cacheKey, func() (any, error) {
```
buqian-zilliz, line 176: **[P2] Bound shared refresh work after callers cancel**

`DoChan` lets an individual request return on `ctx.Done()`, but the shared `GetRLSMetadata` call runs on the manager-lifetime `refreshCtx`, which has no deadline. `singleflight` retains every waiter channel until that function finishes, so a stalled RootCoord call lets repeated timed-out requests for the same key grow the retained waiter list without bound; distinct principals or collections can additionally accumulate stuck goroutines and RPCs. Derive a bounded context from `refreshCtx` for each shared refresh and cap concurrent distinct refreshes, so the work remains shared without outliving all callers indefinitely.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,248 @@
...
+		})
+		if err := merr.CheckRPCCall(resp, err); err != nil {
+			return nil, merr.Wrapf(err, "failed to get RLS principal %q tags", principalName)
+		}
+		if resp.GetCollectionId() != collectionID {
+			return nil, merr.WrapErrServiceInternalMsg("RLS metadata collection id mismatch: requested %d, received %d", collectionID, resp.GetCollectionId())
+		}
+		var tags map[string]rlsutil.TagValue
+		for _, principal := range resp.GetPrincipals() {
+			if principal.GetPrincipalName() == principalName {
```
buqian-zilliz, line 211: **[P2] Reject principal records from the wrong collection**

After validating only the response-level collection ID, this loop accepts a matching principal without checking the record's own `CollectionId`. RootCoord reload groups records by their collection-key prefix but preserves the payload's collection ID, so an inconsistent record stored under collection A with collection B in its payload is returned under outer collection A and its tags are cached for A. Reject a matching record whose `CollectionId` differs from the requested collection, as `rowPoliciesFromInfo` already does for policies, so corrupt authorization metadata fails closed.

pkg/proto/root_coord.proto
```diff
@@ -369,3 +372,27 @@ message CreatePartitionResponse {
   common.Status status = 1;
   int64 partitionID = 2;
 }
+
+enum RLSMetadataKind {
+  RLS_METADATA_KIND_ALL = 0;
```
buqian-zilliz: **[P2] Do not make the bulk snapshot the default request**

Because this is a proto3 enum, an omitted `kind` becomes zero and therefore selects `RLS_METADATA_KIND_ALL`. The introduced Proxy callers request policies or one named principal, and there is no current production caller for `ALL`; nevertheless, a missing field triggers the broadest operation, cloning up to 1,000 principals with 50 tags each plus every policy while holding the global `ddLock` read lock and potentially producing a response of tens of megabytes. Reserve zero for an unspecified value and reject it, and keep `ALL` as a nonzero option only if a bounded consumer actually needs it.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+	resultCh := m.metadataRefreshes.DoChan(policyRefreshKey(collectionID), func() (any, error) {
+		if !m.snapshotRefreshDue(collectionID, refreshTTL, time.Now()) {
+			return struct{}{}, nil
+		}
+		coord, allocVersion, validateFreshness, refreshCtx := m.refreshDependencies()
+		if !validateFreshness || coord == nil || allocVersion == nil || refreshCtx == nil {
+			return struct{}{}, merr.WrapErrServiceInternalMsg("failed to refresh RLS metadata without required dependencies")
+		}
+		version, err := allocVersion(refreshCtx)
```
buqian-zilliz: **[P3] Remove the redundant timestamp allocation from policy refreshes**

`allocVersion` calls Proxy's `timestampAllocator.AllocOne`, which performs a separate RootCoord `AllocTimestamp` RPC before the subsequent `GetRLSMetadata` RPC. The resulting value is only compared locally, while per-collection `singleflight` and `policyRefreshGeneration` already reject refreshes that cross invalidation or removal; no production caller submits independently versioned `RefreshPolicySnapshot` calls. Remove the `SnapshotVersionAllocator` and local version plumbing in favor of the existing generation fence, or return an authoritative metadata revision from `GetRLSMetadata` if revision ordering is required. This removes one coordinator round trip and one failure dependency from every expired policy refresh.

pkg/proto/root_coord.proto
aoiasd: Fixed in c6563e242d. RLS_METADATA_KIND_UNSPECIFIED is now zero and rejected; the existing production wire values remain POLICIES=1 and PRINCIPALS=2, while explicit bulk ALL is 3. Bulk callers must now opt in, and the omitted-kind case has a focused test.

internal/proxy/rls/snapshot.go
aoiasd: The shared refresh intentionally runs on the Proxy manager lifecycle context, while each caller waits through DoChan and can return immediately on its own context cancellation. An arbitrary cache-local deadline would fail closed on a legitimate slow coordinator call and would contradict the agreed synchronization semantics. Proxy shutdown cancels the manager context. A distinct concurrency or liveness policy can be added later if coordinator behavior requires it, but this PR does not invent one.

internal/proxy/rls/manager.go
aoiasd: Fixed in c6563e242d. The SnapshotVersionAllocator and local version plumbing were removed, so an expired policy refresh now performs only GetRLSMetadata. Per-collection singleflight coalesces production refreshes, and the collection state plus policyRefreshGeneration fence rejects results crossing invalidation or collection removal. If independent refresh ordering is needed later, it must use an authoritative metadata revision rather than a pre-read TSO.

internal/proxy/rls/snapshot.go
aoiasd: The authoritative principal ownership is the etcd key rls/principal/{collectionID}/{principalName}. RootCoord groups and returns the record under that collection ID; the CollectionId inside the payload is redundant. A mismatch requires manually corrupted metadata or another upstream internal bug. Per the current design decision, Proxy trusts the global collection-ID key and does not duplicate this integrity check.

internal/distributed/mixcoord/client/client.go
aoiasd: Fixed in c6563e242d. The MixCoord wrapper now forwards opts to client.GetRLSMetadata.

internal/proxy/rls/manager.go
aoiasd: Fixed in c6563e242d. The full principal snapshot path and its second clone were removed. A miss now fetches and caches one principal only; the remaining single bounded tag-map copy isolates cache ownership.

internal/rootcoord/ddl_callbacks_rls.go
aoiasd: Fixed in c6563e242d. Detached context.WithoutCancel notification goroutines were removed. Cache expiration now runs synchronously in the broadcaster ACK callback and inherits its lifecycle and retry behavior.

internal/proxy/rls/snapshot.go
aoiasd: Fixed in c6563e242d by removing the non-authoritative TSO/version ordering entirely. Production policy refreshes are coalesced per collection, and generation plus state identity prevents a result from crossing invalidation or collection removal.

internal/rootcoord/ddl_callbacks_rls.go
aoiasd: Fixed in c6563e242d. ApplyAlterRLSPrincipal errors are checked and returned immediately, before cache expiration and ACK completion.

internal/proxy/rls/manager.go
aoiasd: Fixed in c6563e242d. Collection and state locks are held only while capturing or publishing refresh state; GetRLSMetadata runs outside them. Generation, token, and state-identity fences reject stale publication after invalidation or removal.

internal/proxy/rls/snapshot.go
aoiasd: Fixed in c6563e242d. Policy refreshes request only POLICIES; principal misses request PRINCIPALS for one principal_name. Proxy invalidation is likewise split between the collection policy cache and one collectionID/principal entry.

internal/proxy/rls/reconciler.go
aoiasd: Fixed in c6563e242d. The background reconciler and startup collection sweep were removed. RLS policy metadata is loaded lazily only for an RLS-enforced request, and principal tags are loaded per principal.

internal/proxy/impl.go
aoiasd: Fixed in c6563e242d. Policy mutations invalidate only the collection policy snapshot; principal updates and deletes invalidate only the collectionID/principal cache entry. First-time principal tags need no fan-out because missing principals are not negative-cached.

internal/proxy/meta_cache.go
aoiasd: Fixed in c6563e242d. DropCollection removes RLS state by global collection ID, DropDatabase walks its detached cache bucket, and state identity plus refresh tokens prevent in-flight refreshes from publishing into removed state.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,232 @@
...
+				tags, err = rlsutil.TagsFromJSON(principal.GetTags())
```
buqian-zilliz, line 212: **[P2] Preserve numeric tag kinds across the snapshot RPC**

`RLSPrincipalInfo.tags` is JSON, so this new refresh path cannot preserve the declared numeric kind. `TagsToJSON` writes a `TagValueKindDouble` value such as `1.0` as JSON `1`, and this call to `TagsFromJSON` successfully parses that token as `int64` before the fallback for out-of-range values is reached. The Proxy therefore caches the value as `TagValueKindInt64`, which changes the typed literal supplied to RLS expressions. Encode tags with typed protobuf values or explicit per-value type markers so every legal `TagValue` round-trips without changing kind.

aoiasd: Confirmed. We intentionally keep the JSON representation in this synchronization foundation instead of introducing a one-off typed encoding here. The dependent evaluator PR #52074 treats int64 and double as one numeric family and normalizes values to the target field type only when the conversion is exact, so the 1.0 to 1 normalization does not change the authorization result. Preserving the originally declared numeric kind requires one end-to-end client, API, WAL, catalog, and cache representation and will be handled as a follow-up rather than partially in #52073.

internal/rootcoord/ddl_callbacks_rls.go
```diff
@@ -124,18 +132,24 @@ func (c *Core) broadcastSetRLSPrincipalTags(ctx context.Context, req *rlsutil.Se
 	if err != nil {
 		return err
 	}
-	return broadcastAlterRLSPrincipal(ctx, broadcaster, principal)
+	var cacheExpirations *message.CacheExpirations
+	if !principal.IsNew {
```
buqian-zilliz: **[P2] Invalidate first-time principals to fence an in-flight miss**

Skipping the expiration for a new principal is unsafe even though missing principals are not stored in the cache. A pre-create `GetRLSMetadata` miss still owns a `principalRefreshToken` and a `singleflight` entry; if it snapshots no record before this broadcast is applied but completes afterward, requests that begin after the successful mutation can join that old flight and receive empty tags because no invalidation bumps the token or forgets the flight. Always send the principal-scoped invalidation for creates as well; `RemovePrincipal` is harmless when no positive entry exists and also fences any in-flight miss.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,475 @@
...
+	return state
+}
+
+func (m *manager) beginPolicyRefresh(collectionID UniqueID) (*collectionState, uint64) {
+	key := newCollectionKey(collectionID)
+	m.mu.Lock()
+	state := m.collections[key]
+	if state == nil {
+		state = newCollectionState()
+		m.collections[key] = state
```
buqian-zilliz: **[P2] Bound the aggregate RLS metadata cache**

`beginPolicyRefresh` permanently adds each accessed collection to `collections`, successful policy snapshots have no idle or capacity eviction, and the scanner only ages out principal entries. At the supported ceiling of 65,536 collections and 100 policies per collection with multi-KiB expressions, a long-lived Proxy can retain many GiB after a large working set has been touched; refreshing data does not bound that footprint. Add an aggregate entry or byte budget with idle/LRU eviction and reload evicted snapshots lazily.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,475 @@
...
+	if refreshTTL <= 0 {
+		return merr.WrapErrServiceInternalMsg("failed to validate RLS metadata freshness with invalid TTL %s", refreshTTL)
+	}
+	if !m.snapshotRefreshDue(collectionID, refreshTTL, time.Now()) {
+		return nil
+	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	resultCh := m.metadataRefreshes.DoChan(policyRefreshKey(collectionID), func() (any, error) {
```
buqian-zilliz: **[P2] Back off failed policy refreshes instead of retrying at request rate**

When an expired policy refresh fails, `policyLastSuccessfulRefresh` remains expired and the completed `singleflight` call is immediately removed. Every subsequent request can therefore start another `GetRLSMetadata` RPC; a fast coordinator error or corrupt response turns query traffic across collections and Proxies into a retry storm while requests are already failing closed. Record a per-collection retry deadline with bounded exponential backoff and jitter so requests during the backoff fail closed without issuing another RPC.

internal/proxy/rls/manager.go
aoiasd: This is outside the current v1 scale target. The design assumes RLS policy and tag mutations are low-frequency and does not target tens of thousands of simultaneously active RLS collections in one Proxy cache. DropCollection and DropDatabase remove their collection state, while active policy snapshots remain collection-scoped and reload lazily after invalidation or TTL. We will add an aggregate LRU or byte budget if production collection scale makes this measurable, rather than adding an unvalidated eviction policy in this PR.

internal/rootcoord/ddl_callbacks_rls.go
aoiasd: Fixed in 382d97019b. Every SetRLSPrincipalTags mutation, including the first stored binding, now carries a principal-scoped cache expiration. The ACK callback applies the post-image before invalidation; Proxy RemovePrincipal removes the refresh token and forgets the singleflight key, so a pre-create miss cannot publish or be joined after the mutation. The concurrency test now exercises an in-flight empty result followed by invalidation and a fresh lookup.

internal/proxy/rls/manager.go
aoiasd: Fixed in 382d97019b. Failed policy refreshes now record a per-collection bounded exponential backoff with jitter using the existing typeutil BackoffWithInstant helper. An expired snapshot remains fail-closed during the deadline and requests do not issue another metadata RPC; successful refresh and policy invalidation reset the backoff. The focused test verifies two immediate requests cause one RPC and invalidation permits immediate recovery.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (m *manager) ensurePrincipalTags(ctx context.Context, collectionID UniqueID, principalName string) (map[string]rlsutil.TagValue, error) {
+	if m == nil || collectionID == 0 || principalName == "" {
+		return nil, merr.WrapErrPrivilegeNotPermitted("RLS principal is required")
+	}
+
+	key := principalKey{collectionID: collectionID, principalName: principalName}
+	coord, validateFreshness, _ := m.refreshDependencies()
+	if entry := m.getPrincipalTagsEntry(key); entry != nil {
+		return clonePrincipalTags(entry.tags), nil
```
buqian-zilliz: **[P2] Avoid cloning the principal tag map on every cache hit**

Every positive cache hit calls `clonePrincipalTags`, which allocates a new map and rehashes all tags before returning. This is the intended per-request lookup for RLS enforcement; with up to 50 tags per principal, steady traffic pays `O(tags × QPS)` allocation and CPU even though cached entries are immutable and invalidation replaces or removes them. Expose an immutable lookup/view or another evaluator-owned read-only representation, and copy only at mutation or publication boundaries.

aoiasd: Fixed in 473d5a1717. Principal tags are still cloned once when the cache entry is published, then the package-internal cache-hit and singleflight paths share that immutable map. The only production consumer reads tag values without mutating or retaining the map; invalidation replaces or removes the complete entry. The focused cache test now confirms the first refreshed result and the subsequent hit use the same published map.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,228 @@
...
+
+		resp, err := coord.GetRLSMetadata(refreshCtx, &rootcoordpb.GetRLSMetadataRequest{
+			Base: commonpbutil.NewMsgBase(
+				commonpbutil.WithSourceID(paramtable.GetNodeID()),
+			),
+			CollectionId:  collectionID,
+			Kind:          rootcoordpb.RLSMetadataKind_RLS_METADATA_KIND_PRINCIPALS,
+			PrincipalName: principalName,
+		})
+		if err := merr.CheckRPCCall(resp, err); err != nil {
```
buqian-zilliz, line 203: **[P2] Back off failed principal refreshes**

`singleflight` only coalesces callers while this RPC is running. After a transport, status, or decode failure, deferred cleanup removes the principal token and the completed singleflight entry disappears, so the next cache-miss request immediately issues another `GetRLSMetadata` call. Once entries expire during a RootCoord outage, RLS-enforced query QPS can therefore become coordinator retry QPS even though every request fails closed. Add a bounded per-collection or global principal-refresh backoff/circuit breaker, and reset it on success or invalidation.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+func (m *manager) finishPolicyRefresh(collectionID UniqueID, state *collectionState, generation uint64, snapshot *policySnapshot) bool {
+	m.mu.Lock()
+	state.mu.Lock()
+	state.policyRefreshes--
+	current := m.collections[collectionID] == state && state.policyRefreshGeneration == generation
+	if current {
+		if snapshot != nil {
+			current = state.setRLSPolicySnapshotLocked(*snapshot)
```
buqian-zilliz: **[P3] Prepare policy snapshots outside the global manager lock**

`finishPolicyRefresh` holds the process-wide `m.mu` write lock while `setRLSPolicySnapshotLocked` validates and clones the collection's policies and their action slices into a new map. Every lookup for every other collection first needs the same mutex in `getCollectionState`, so concurrent TTL refreshes serialize this allocation work and stall unrelated RLS requests. Build the validated immutable map before taking `m.mu`, then hold the generation and state locks only long enough to verify the incarnation and swap the prepared snapshot.

internal/proxy/rls/snapshot.go
```diff
@@ -0,0 +1,298 @@
...
+
+type CoordClient interface {
+	GetRLSMetadata(ctx context.Context, in *rootcoordpb.GetRLSMetadataRequest, opts ...grpc.CallOption) (*rootcoordpb.GetRLSMetadataResponse, error)
+}
+
+func (m *manager) ensurePoliciesFresh(ctx context.Context, collectionID UniqueID) error {
+	if m == nil || collectionID == 0 {
+		return merr.WrapErrServiceInternalMsg("failed to validate RLS policy freshness with invalid manager or collection id")
+	}
+	coord, refreshCtx := m.dependencies()
```
buqian-zilliz: **[P3] Avoid the lifecycle mutex on every RLS cache hit**

Both `ensurePoliciesFresh` and `ensurePrincipalTags` call `dependencies()` before checking whether the requested metadata is already cached. That helper acquires the process-wide `lifecycleMu`, while the refresh paths fetch the dependencies again in `refreshPolicies` or the principal singleflight closure. The preflight check therefore adds a shared mutex acquisition to every intended request-path cache hit and duplicates it on misses. Move the dependency check after the cache-hit path, or rely on the existing checks when a refresh actually starts.

internal/proxy/rls/snapshot.go
buqian-zilliz: This issue still exists: failed principal refreshes still clear their token without recording any backoff, so the next cache miss can immediately start another coordinator RPC.

internal/proxy/rls/manager.go
buqian-zilliz: This issue still exists: validation is now outside the critical section, but replacement-map allocation and policy/action cloning still occur while the process-wide manager lock is held.

pkg/proto/root_coord.proto
```diff
@@ -369,3 +372,28 @@ message CreatePartitionResponse {
...
+  int64 collection_id = 2;
+  RLSMetadataKind kind = 3;
+  // Optional principal filter. It is valid only with PRINCIPALS and allows
+  // Proxy to load one principal by the immutable collection ID.
+  string principal_name = 4;
+}
+
+message GetRLSMetadataResponse {
+  common.Status status = 1;
+  string db_name = 2;
```
buqian-zilliz: **[P3] Remove unused names from the snapshot response**

The production callers of `GetRLSMetadata` validate `collection_id` and consume policies or principals; neither reads `db_name` nor `collection_name`. Returning these fields makes `MetaTable.GetRLSMetadata` resolve a database name, including an O(number of databases) fallback while holding `ddLock`, and threads two unused values through the model and RPC. Remove these fields and construct the response from `collection_id` plus the requested RLS payload.

internal/proxy/metacache/meta_cache.go
```diff
@@ -1507,6 +1508,9 @@ func (m *MetaCache) RemoveDatabase(ctx context.Context, database string) {
 	evicted := 0
 	for start := 0; start < len(ids); start += removeBatch {
 		end := min(start+removeBatch, len(ids))
+		for _, id := range ids[start:end] {
+			rls.MarkCollectionDropped(id)
```
buqian-zilliz, line 1512: **[P2] Tombstone RLS state independently of MetaCache residency**

`RemoveDatabase` obtains `ids` only from the current `nameIdx` bucket, but ordinary collection-cache invalidations remove that entry while the separate RLS manager retains its policy and principal state. If the metadata entry is not refilled before `DropDatabase`, this loop never calls `MarkCollectionDropped` for that collection; its RLS state survives the database-drop acknowledgment, and an in-flight request that resolved the ID earlier can still read that state after the drop. Track database membership in the RLS manager or carry authoritative collection IDs in the drop invalidation so every RLS state for the database is tombstoned.

pkg/proto/root_coord.proto
```diff
@@ -369,3 +372,28 @@ message CreatePartitionResponse {
   common.Status status = 1;
   int64 partitionID = 2;
 }
+
+enum RLSMetadataKind {
+  RLS_METADATA_KIND_UNSPECIFIED = 0;
+  RLS_METADATA_KIND_POLICIES = 1;
+  RLS_METADATA_KIND_PRINCIPALS = 2;
+  RLS_METADATA_KIND_ALL = 3;
```
buqian-zilliz, line 380: **[P3] Remove the unused full-snapshot selector**

The production Proxy callers request either `POLICIES` or one `PRINCIPALS` record; no caller uses `RLS_METADATA_KIND_ALL`. Supporting it adds a branch that clones every policy and principal while holding `MetaTable.ddLock` and permits an unnecessarily broad response shape. Remove `ALL` and its server branch until a concrete caller needs a full snapshot; the two targeted selectors already cover the current cache.

aoiasd: `ALL` is intentionally retained as the explicitly requested collection-scoped bulk RLS metadata API contract. The current Proxy cache now uses the narrower `POLICIES` and single-`PRINCIPALS` selectors on its hot path, but that implementation choice does not remove the bulk API requirement. Since no production caller currently selects `ALL`, its cloning branch also has no runtime lock cost on existing request paths. Removing the enum value and server behavior would narrow the agreed internal contract, so this PR keeps it unchanged.

internal/proxy/metacache/meta_cache.go
aoiasd: This sequence is not reachable through the production DDL path. RootCoord rejects `DropDatabase` while the database contains any collection (`CheckIfDatabaseDroppable`). `DropCollection` holds a shared DB resource lock, while `DropDatabase` takes the exclusive DB lock, and the collection lock is retained through its ACK callback. That callback sends an expiration carrying the authoritative collection ID; Proxy calls `rls.MarkCollectionDropped(collectionID)` independently of whether MetaCache/nameIdx still contains the collection, and callback errors are retried. Therefore an ordinary MetaCache invalidation does not create a gap: before the database can become droppable, the collection is either dropped and already tombstoned by ID, or moved by cross-DB rename and its RLS state must remain live. Tracking mutable DB membership in the RLS manager would instead risk tombstoning a collection that moved to another database. No production-code change is needed here.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+func (m *manager) finishPolicyRefresh(collectionID UniqueID, state *collectionState, generation uint64, snapshot *policySnapshot) bool {
+	if state == nil {
+		return false
+	}
+	var refreshedAt time.Time
+	var policies map[string]*rlsutil.RowPolicy
+	if snapshot != nil {
+		refreshedAt, policies = preparePolicySnapshot(*snapshot)
```
buqian-zilliz: **[P3] Publish decoded policies without recloning them**

`rowPoliciesFromInfo` already allocates a private `RowPolicy` and actions slice for every protobuf entry. `preparePolicySnapshot` then allocates a second map and `cloneRowPolicy` copies each policy and its actions again, even though no production owner retains the decoded slice. Have the conversion build the name-keyed map, or otherwise transfer ownership, and publish it directly; this removes the one-caller snapshot pipeline and duplicate allocations on every policy refresh.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,422 @@
...
+		droppedCollections: map[UniqueID]struct{}{},
+	}
+}
+
+func (m *manager) init(ctx context.Context, coord CoordClient) error {
+	if m == nil || coord == nil {
+		return merr.WrapErrServiceInternalMsg("failed to initialize RLS metadata manager without required dependencies")
+	}
+
+	m.lifecycleMu.Lock()
```
buqian-zilliz: **[P3] Remove reinitialization-only lifecycle state**

`rls.Init` has one production call from `Proxy.Init`, before the manager can serve refreshes, yet `init` implements cancel-and-replace reinitialization and every refresh calls `dependencies()` under `lifecycleMu`. Treat the coordinator and Proxy-lifetime context as immutable after initialization, use that context directly for the scanner and refreshes, and remove `lifecycleCancel` and `lifecycleMu`; this deletes reinitialization-only state and avoids a lock on every metadata refresh.

internal/proxy/rls/manager.go
aoiasd: Fixed in 9f54662abf. rowPoliciesFromInfo now builds the final name-keyed map, and the refresh publishes that owned map directly. The one-caller snapshot/clone pipeline was removed; the same pass also removed the redundant principal-tag clone.

internal/proxy/rls/manager.go
aoiasd: Fixed in 9f54662abf. The coordinator client and Proxy-lifetime context are now immutable after Init; refreshes and the scanner use them directly, and lifecycleMu, lifecycleCancel, dependencies(), and the reinitialization-only test were removed.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,376 @@
...
+
+func (m *manager) invalidatePolicies(collectionID UniqueID) {
+	if m == nil || collectionID == 0 {
+		return
+	}
+	m.mu.RLock()
+	state := m.collections[collectionID]
+	if state != nil {
+		state.mu.Lock()
+		state.policyGeneration++
```
buqian-zilliz: **[P2] Make retry-driven invalidations idempotent**

The ACK callback returns the aggregate fan-out error after it has already invalidated healthy Proxies, so one failed Proxy makes the broadcaster resend the same expiration to every Proxy on each retry. This method treats each duplicate as a new mutation: it increments `policyGeneration`, clears any newly refreshed snapshot, and calls `Forget`; principal invalidation similarly removes its active token. Once the stacked execution consumer uses this cache, retries can repeatedly fail healthy in-flight requests with `metadata changed during refresh` and multiply RootCoord refresh RPCs until the failing Proxy recovers. Carry a stable mutation revision and ignore duplicate invalidations, or retain per-Proxy acknowledgements and retry only failed targets.

aoiasd: Fixed in 2673b89d. Each RLS invalidation now carries the stable control-channel timetick. Proxy keeps a bounded FIFO set of exact processed revisions per collection, so retries are idempotent without assuming callback order; a first-seen out-of-order mutation still invalidates the relevant cache. Tests cover policy/principal retries, out-of-order revisions, bounded-history eviction, and scanner retention.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,429 @@
...
+		state.mu.Unlock()
+		m.mu.Unlock()
+		return
+	}
+	delete(state.principalTags, principalName)
+	delete(state.principalRefreshTokens, principalName)
+	delete(state.principalBackoffs, principalName)
+	state.mu.Unlock()
+	m.mu.Unlock()
+	m.principalRefreshes.Forget(principalRefreshKey(key))
```
buqian-zilliz: **[P2] Do not detach a refresh started after invalidation**

`invalidatePolicies` and `invalidatePrincipalTags` publish the invalidated state by releasing `state.mu` and `m.mu`, but call `singleflight.Forget` only afterwards. A request entering in that gap can register a new post-invalidation flight; `Forget` then detaches that fresh flight and allows another request to start a second one. Policy refreshes issue duplicate RootCoord RPCs, while principal refreshes replace each other's token, causing one legitimate caller to fail with `metadata changed during refresh` despite fetching current metadata. Call `Forget` after mutating the state but before releasing the invalidation locks, or include the generation or token in the singleflight key.

internal/proxy/rls/manager.go
```diff
@@ -0,0 +1,429 @@
...
+
+type collectionState struct {
+	mu sync.RWMutex
+	// Exact revisions deduplicate retries without assuming callback order.
+	invalidationRevisions     map[typeutil.Timestamp]struct{}
+	invalidationRevisionOrder []typeutil.Timestamp
+
+	// Policy snapshot and refresh state.
+	policyGeneration      uint64
+	activePolicyRefreshes int
```
buqian-zilliz: **[P3] Remove unreachable policy-refresh accounting**

`activePolicyRefreshes` is only consulted by the empty-state deletion branch, but none of the supported completion paths can reach that branch: success sets `policyRefreshedAt`, failure sets `policyBackoff`, producer-supplied invalidations retain their nonzero revision, and collection drops detach the state from `m.collections`. Remove the field, its increment and decrement, and the unreachable cleanup branch; the existing generation and state-incarnation checks already fence late refreshes.

internal/proxy/rls/manager.go
aoiasd: Fixed in 80542ca7c6. Invalidation no longer uses singleflight.Forget. Policy flights are keyed by the captured policy generation. Principal token allocation/reuse and DoChan registration are now one manager/state-lock critical section, and the collection-monotonic token is part of the flight key. State identity plus generation/token checks fence old flights before the RPC and again at publication, so a post-invalidation refresh cannot be detached or overwritten.

internal/proxy/rls/manager.go
aoiasd: Fixed in 80542ca7c6. Removed activePolicyRefreshes, its increment/decrement, and the opportunistic empty-state cleanup branch. Policy completion now uses the existing collection-state identity and policyGeneration fences; finishPolicyRefresh only needs the manager read lock.

internal/rootcoord/ddl_callbacks.go
```diff
@@ -116,29 +120,32 @@ func (c *DDLCallback) ExpireCaches(ctx context.Context, expirations any) error {
...
-		if err := c.expireCache(ctx, cacheExpiration); err != nil {
+		if err := c.expireCache(ctx, cacheExpiration, opts...); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func (c *DDLCallback) expireCache(ctx context.Context, cacheExpiration *message.CacheExpiration) error {
+func (c *DDLCallback) expireCache(ctx context.Context, cacheExpiration *message.CacheExpiration, opts ...proxyutil.ExpireCacheOpt) error {
```
buqian-zilliz: **[P2] Skip the discarded TSO allocation in RLS ACK callbacks**

The new timestamp option is applied only after `expireCache` has already called `GenerateTSO(1)`. `ExpireCacheConfig.Apply` then overwrites that generated timestamp with the broadcast result's `MaxTimeTick`, so every RLS alter/drop consumes a needless TSO allocation. More importantly, an allocator failure makes the already-persisted callback fail and retry even though the broadcast result already contains the stable revision required for invalidation. Pass the supplied timestamp into `expireCache` before allocation and call `GenerateTSO` only when no override is present.

aoiasd: Fixed in 94e087e401. RLS ACK cache expiration now passes the broadcast MaxTimeTick directly to expireCache; GenerateTSO runs only when no timestamp is supplied. Removed the redundant SetMsgTimestamp override, and the regression test uses a failing allocator while verifying timestamps 101/102 are forwarded.

pkg/util/paramtable/component_param.go
```diff
@@ -3359,6 +3360,17 @@ Disabled if the value is less or equal to 0.`,
 	}
 	p.RLSMaxArrayLiteralElements.Init(base.mgr)
 
+	p.RLSMetaRefreshInterval = ParamItem{
+		Key:          "proxy.rls.metaRefreshInterval",
+		Version:      "3.0.0",
```
zhengbuqian, line 3365: let's target at 3.0.4 for now.

zhengbuqian (approved):

internal/rootcoord/meta_table.go
```diff
@@ -3410,6 +3411,51 @@ func (mt *MetaTable) ListRLSPolicies(ctx context.Context, req *rlsutil.ListRowPo
...
+		if principalName != "" {
+			return nil, merr.WrapErrServiceInternalMsg("RLS principal filter is only supported for principal metadata")
+		}
+		metadata.Policies = model.RLSPolicyMapToSlice(coll.RLSPolicies)
+	case rootcoordpb.RLSMetadataKind_RLS_METADATA_KIND_PRINCIPALS:
+		if principalName == "" {
+			metadata.Principals = model.CloneRLSPrincipals(coll.RLSPrincipals)
+			break
+		}
+		for _, principal := range coll.RLSPrincipals {
```
zhengbuqian, line 3447: this is linear scan, and could be too slow in case of a million principals? can we use some other data structures to store this in mem?

zhengbuqian (commented):
also is it possible to only load RLS meta in coord for loaded collections?

PR #52074, opened 2026-07-31
feat: [RLS3] evaluate row-level security policies

aoiasd:
relate: #50263
design doc PR: #53173

## Summary
Adds policy combination, template compilation, and a three-valued local evaluator for write checks.

### DIFF
--- a/internal/proxy/rls/manager.go
+++ b/internal/proxy/rls/manager.go
@@ -18,13 +18,15 @@ package rls
 
 import (
 	"context"
+	"slices"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
 
-	"google.golang.org/protobuf/proto"
-
-	"github.com/milvus-io/milvus/pkg/v3/proto/rootcoordpb"
+	"github.com/milvus-io/milvus/internal/parser/planparserv2"
+	"github.com/milvus-io/milvus/internal/util/rlsutil"
+	"github.com/milvus-io/milvus/pkg/v3/proto/planpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/conc"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
@@ -36,7 +38,7 @@ type UniqueID = typeutil.UniqueID
 type policySnapshot struct {
 	Version     int64
 	RefreshedAt time.Time
-	Policies    []*rootcoordpb.RLSPolicyInfo
+	Policies    []*rlsutil.RowPolicy
 }
 
 type principalTagsSnapshot struct {
@@ -47,11 +49,28 @@ type principalTagsSnapshot struct {
 
 type SnapshotVersionAllocator func(ctx context.Context) (uint64, error)
 
-type SnapshotManager interface {
+type Manager interface {
 	Init(ctx context.Context, coord CoordClient, allocVersion SnapshotVersionAllocator) error
 	RefreshPolicySnapshot(ctx context.Context, coord CoordClient, dbName string, collectionName string, collectionID UniqueID, version uint64) error
 	RefreshPrincipalTagsSnapshot(ctx context.Context, coord CoordClient, dbName string, collectionName string, collectionID UniqueID, version uint64) error
 	RunReconciler(ctx context.Context, coord CoordClient, allocVersion SnapshotVersionAllocator) error
+
+	GetRLSUsingPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs) (*planpb.Expr, error)
+	ApplyRLSUsingPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs, plan *planpb.PlanNode) error
+	GetRLSCheckPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs) (*planpb.Expr, error)
+}
+
+type exprKind int
+
+const (
+	usingExprKind exprKind = iota
+	checkExprKind
+)
+
+type compiledKey struct {
+	action   rlsutil.PolicyAction
+	kind     exprKind
+	timezone string
 }
 
 type collectionState struct {
@@ -60,8 +79,9 @@ type collectionState struct {
 	principalTagVersion               int64
 	policyLastSuccessfulRefresh       time.Time
 	principalTagLastSuccessfulRefresh time.Time
-	policies                          map[string]*rootcoordpb.RLSPolicyInfo
+	policies                          map[string]*rlsutil.RowPolicy
 	principalTags                     map[string]map[string]string
+	compiled                          map[compiledKey]*compiledExpression
 }
 
 type collectionKey struct {
@@ -74,12 +94,13 @@ type manager struct {
 	dependencyMu      sync.RWMutex
 	coord             CoordClient
 	allocVersion      SnapshotVersionAllocator
+	validateFreshness bool
 	metadataRefreshes conc.Singleflight[struct{}]
 }
 
 var defaultManager = newManager()
 
-func DefaultManager() SnapshotManager {
+func DefaultManager() Manager {
 	return defaultManager
 }
 
@@ -98,17 +119,25 @@ func (m *manager) configure(coord CoordClient, allocVersion SnapshotVersionAlloc
 	defer m.dependencyMu.Unlock()
 	m.coord = coord
 	m.allocVersion = allocVersion
+	m.validateFreshness = true
 }
 
-func (m *manager) refreshDependencies() (CoordClient, SnapshotVersionAllocator) {
+func (m *manager) refreshDependencies() (CoordClient, SnapshotVersionAllocator, bool) {
 	m.dependencyMu.RLock()
 	defer m.dependencyMu.RUnlock()
-	return m.coord, m.allocVersion
+	return m.coord, m.allocVersion, m.validateFreshness
 }
 
 func (m *manager) ensureFreshMetadata(ctx context.Context, collectionID UniqueID) error {
 	if m == nil || collectionID == 0 {
 		return merr.WrapErrServiceInternalMsg("failed to validate RLS metadata freshness with invalid manager or collection id")
+	}
+	coord, allocVersion, validateFreshness := m.refreshDependencies()
+	if !validateFreshness {
+		return nil
+	}
+	if coord == nil || allocVersion == nil {
+		return merr.WrapErrServiceInternalMsg("failed to refresh RLS metadata without required dependencies")
 	}
 	refreshTTL := paramtable.Get().ProxyCfg.RLSMetaRefreshInterval.GetAsDuration(time.Second)
 	if refreshTTL <= 0 {
@@ -122,8 +151,8 @@ func (m *manager) ensureFreshMetadata(ctx context.Context, collectionID UniqueID
 		if policyDue, principalTagsDue := m.snapshotRefreshDue(collectionID, refreshTTL, time.Now()); !policyDue && !principalTagsDue {
 			return struct{}{}, nil
 		}
-		coord, allocVersion := m.refreshDependencies()
-		if coord == nil || allocVersion == nil {
+		coord, allocVersion, validateFreshness := m.refreshDependencies()
+		if !validateFreshness || coord == nil || allocVersion == nil {
 			return struct{}{}, merr.WrapErrServiceInternalMsg("failed to refresh RLS metadata without required dependencies")
 		}
 		version, err := allocVersion(ctx)
@@ -154,12 +183,13 @@ func (m *manager) setRLSPolicySnapshot(_ string, collectionID UniqueID, snapshot
 	}
 	state.policyVersion = snapshot.Version
 	state.policyLastSuccessfulRefresh = snapshot.RefreshedAt
-	state.policies = map[string]*rootcoordpb.RLSPolicyInfo{}
+	state.policies = map[string]*rlsutil.RowPolicy{}
+	state.compiled = map[compiledKey]*compiledExpression{}
 	for _, policy := range snapshot.Policies {
 		if policy == nil || policy.GetPolicyName() == "" {
 			continue
 		}
-		state.policies[policy.GetPolicyName()] = proto.Clone(policy).(*rootcoordpb.RLSPolicyInfo)
+		state.policies[policy.GetPolicyName()] = cloneRowPolicy(policy)
 	}
 	return true
 }
@@ -190,7 +220,7 @@ func (m *manager) setRLSPrincipalTagsSnapshot(_ string, collectionID UniqueID, s
 	return true
 }
 
-func (m *manager) removeCollection(_ context.Context, collectionID UniqueID) {
+func (m *manager) removeCollection(ctx context.Context, collectionID UniqueID) {
 	if m == nil || collectionID == 0 {
 		return
 	}
@@ -200,10 +230,179 @@ func (m *manager) removeCollection(_ context.Context, collectionID UniqueID) {
 	delete(m.collections, newCollectionKey(collectionID))
 }
 
+func (m *manager) GetRLSUsingPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs) (*planpb.Expr, error) {
+	return m.getRLSPredicate(ctx, collectionID, principalName, action, enforceRLS, usingExprKind, schemaHelper, visitorArgs, func(policy *rlsutil.RowPolicy) string {
+		return policy.GetUsingExpr()
+	})
+}
+
+func (m *manager) ApplyRLSUsingPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs, plan *planpb.PlanNode) error {
+	predicate, err := m.GetRLSUsingPredicate(ctx, collectionID, principalName, action, enforceRLS, schemaHelper, visitorArgs)
+	if err != nil {
+		return err
+	}
+	return MergePredicateToPlan(plan, predicate)
+}
+
+func (m *manager) GetRLSCheckPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs) (*planpb.Expr, error) {
+	return m.getRLSPredicate(ctx, collectionID, principalName, action, enforceRLS, checkExprKind, schemaHelper, visitorArgs, func(policy *rlsutil.RowPolicy) string {
+		return policy.GetCheckExpr()
+	})
+}
+
+func (m *manager) getRLSPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, enforceRLS bool, kind exprKind, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs, exprSelector func(*rlsutil.RowPolicy) string) (*planpb.Expr, error) {
+	if !enforceRLS {
+		return nil, nil
+	}
+	if m == nil || collectionID == 0 {
+		return nil, denyNoApplicableRLSPolicy(action, kind)
+	}
+	if err := m.ensureFreshMetadata(ctx, collectionID); err != nil {
+		return nil, merr.Wrapf(err, "failed to validate RLS metadata for collection %d", collectionID)
+	}
+
+	state := m.getCollectionState(newCollectionKey(collectionID))
+	if state == nil {
+		return nil, denyNoApplicableRLSPolicy(action, kind)
+	}
+	compiledExpr, tags, err := state.getCompiledExprAndTags(principalName, action, kind, schemaHelper, visitorArgs, exprSelector)
+	if err != nil || compiledExpr == nil {
+		if err != nil {
+			return nil, err
+		}
+		return nil, denyNoApplicableRLSPolicy(action, kind)
+	}
+
+	expr, err := compiledExpr.Instantiate(principalName, tags)
+	if err != nil {
+		return nil, err
+	}
+	if expr == nil {
+		return nil, denyNoApplicableRLSPolicy(action, kind)
+	}
+	if isAlwaysTrueExpr(expr) {
+		return nil, nil
+	}
+	return expr, nil
+}
+
+func denyNoApplicableRLSPolicy(action rlsutil.PolicyAction, kind exprKind) error {
+	return merr.WrapErrPrivilegeNotPermitted("%s operation denied by RLS: no applicable %s policies", rlsActionOperation(action), kind.policyLabel())
+}
+
+func (kind exprKind) policyLabel() string {
+	switch kind {
+	case checkExprKind:
+		return "check"
+	default:
+		return "using"
+	}
+}
+
+func rlsActionOperation(action rlsutil.PolicyAction) string {
+	switch action {
+	case rlsutil.PolicyActionQuery:
+		return "query"
+	case rlsutil.PolicyActionQueryIterator:
+		return "query iterator"
+	case rlsutil.PolicyActionSearch:
+		return "search"
+	case rlsutil.PolicyActionSearchIterator:
+		return "search iterator"
+	case rlsutil.PolicyActionHybridSearch:
+		return "hybrid search"
+	case rlsutil.PolicyActionDelete:
+		return "delete"
+	case rlsutil.PolicyActionInsert:
+		return "insert"
+	case rlsutil.PolicyActionUpsert:
+		return "upsert"
+	default:
+		return "unknown"
+	}
+}
+
+func orderedPolicies(policiesByName map[string]*rlsutil.RowPolicy) []*rlsutil.RowPolicy {
+	names := make([]string, 0, len(policiesByName))
+	for name := range policiesByName {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	policies := make([]*rlsutil.RowPolicy, 0, len(names))
+	for _, name := range names {
+		policies = append(policies, policiesByName[name])
+	}
+	return policies
+}
+
+func (state *collectionState) getCompiledExprAndTags(principalName string, action rlsutil.PolicyAction, kind exprKind, schemaHelper *typeutil.SchemaHelper, visitorArgs *planparserv2.ParserVisitorArgs, exprSelector func(*rlsutil.RowPolicy) string) (*compiledExpression, map[string]string, error) {
+	key := compiledKey{
+		action: action,
+		kind:   kind,
+	}
+	if visitorArgs != nil {
+		key.timezone = visitorArgs.Timezone
+	}
+
+	state.mu.RLock()
+	if len(state.policies) == 0 {
+		state.mu.RUnlock()
+		return nil, nil, nil
+	}
+	if compiledExpr, ok := state.compiled[key]; ok {
+		tags := state.principalTags[principalName]
+		state.mu.RUnlock()
+		return compiledExpr, tags, nil
+	}
+	state.mu.RUnlock()
+
+	state.mu.Lock()
+	defer state.mu.Unlock()
+	if len(state.policies) == 0 {
+		return nil, nil, nil
+	}
+
+	if state.compiled == nil {
+		state.compiled = map[compiledKey]*compiledExpression{}
+	}
+	if compiledExpr, ok := state.compiled[key]; ok {
+		return compiledExpr, state.principalTags[principalName], nil
+	}
+
+	policies := orderedPolicies(state.policies)
+	templates, combinedExpr := preparePolicyExprTemplates(policies, action, exprSelector)
+	if maxExpressionLength := paramtable.Get().ProxyCfg.RLSMaxCombinedExpressionLength.GetAsInt(); len(combinedExpr) > maxExpressionLength {
+		return nil, nil, merr.WrapErrServiceQuotaExceededMsg("RLS combined expression exceeds max length %d", maxExpressionLength)
+	}
+	compiledExpr, err := compileExprTemplates(schemaHelper, templates, visitorArgs)
+	if err != nil {
+		return nil, nil, err
+	}
+	state.compiled[key] = compiledExpr
+	return compiledExpr, state.principalTags[principalName], nil
+}
+
 func newCollectionState() *collectionState {
 	return &collectionState{
-		policies:      map[string]*rootcoordpb.RLSPolicyInfo{},
+		policies:      map[string]*rlsutil.RowPolicy{},
 		principalTags: map[string]map[string]string{},
+		compiled:      map[compiledKey]*compiledExpression{},
+	}
+}
+
+func cloneRowPolicy(policy *rlsutil.RowPolicy) *rlsutil.RowPolicy {
+	if policy == nil {
+		return nil
+	}
+	return &rlsutil.RowPolicy{
+		PolicyName:  policy.PolicyName,
+		PolicyType:  policy.PolicyType,
+		Actions:     slices.Clone(policy.Actions),
+		UsingExpr:   policy.UsingExpr,
+		CheckExpr:   policy.CheckExpr,
+		Description: policy.Description,
+		PolicyId:    policy.PolicyId,
 	}
 }
 
--- a/internal/proxy/rls/snapshot.go
+++ b/internal/proxy/rls/snapshot.go
@@ -22,6 +22,7 @@ import (
 
 	"google.golang.org/grpc"
 
+	"github.com/milvus-io/milvus/internal/util/rlsutil"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/rootcoordpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/commonpbutil"
@@ -89,7 +90,7 @@ func (m *manager) refreshSnapshots(ctx context.Context, coord CoordClient, dbNam
 		updated := m.setRLSPolicySnapshot(dbName, collectionID, policySnapshot{
 			Version:     int64(version),
 			RefreshedAt: refreshedAt,
-			Policies:    resp.GetPolicies(),
+			Policies:    rowPoliciesFromInfo(resp.GetPolicies()),
 		})
 		if !updated {
 			mlog.Debug(ctx, "skip stale RLS policy snapshot",
@@ -121,4 +122,27 @@ func (m *manager) refreshSnapshots(ctx context.Context, coord CoordClient, dbNam
 		}
 	}
 	return nil
+}
+
+func rowPoliciesFromInfo(policies []*rootcoordpb.RLSPolicyInfo) []*rlsutil.RowPolicy {
+	converted := make([]*rlsutil.RowPolicy, 0, len(policies))
+	for _, policy := range policies {
+		if policy == nil {
+			continue
+		}
+		actions := make([]rlsutil.PolicyAction, len(policy.GetActions()))
+		for i, action := range policy.GetActions() {
+			actions[i] = rlsutil.PolicyAction(action)
+		}
+		converted = append(converted, &rlsutil.RowPolicy{
+			PolicyName:  policy.GetPolicyName(),
+			PolicyType:  rlsutil.PolicyType(policy.GetPolicyType()),
+			Actions:     actions,
+			UsingExpr:   policy.GetUsingExpr(),
+			CheckExpr:   policy.GetCheckExpr(),
+			Description: policy.GetDescription(),
+			PolicyId:    policy.GetPolicyId(),
+		})
+	}
+	return converted
 }
--- a/internal/proxy/rls/util.go
+++ b/internal/proxy/rls/util.go
@@ -1,0 +1,967 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rls
+
+import (
+	"context"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
+	"github.com/milvus-io/milvus/internal/parser/planparserv2"
+	"github.com/milvus-io/milvus/internal/parser/planparserv2/rewriter"
+	"github.com/milvus-io/milvus/internal/util/rlsutil"
+	"github.com/milvus-io/milvus/pkg/v3/mlog"
+	"github.com/milvus-io/milvus/pkg/v3/proto/planpb"
+	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
+	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
+	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
+)
+
+type compiledExpression struct {
+	permissive  []*compiledPolicyExpression
+	restrictive []*compiledPolicyExpression
+}
+
+type compiledPolicyExpression struct {
+	expr           *planpb.Expr
+	needsPrincipal bool
+	tagVariables   map[string]string
+}
+
+type policyExprTemplate struct {
+	policyType     rlsutil.PolicyType
+	expr           string
+	needsPrincipal bool
+	tagVariables   map[string]string
+}
+
+func preparePolicyExprTemplates(policies []*rlsutil.RowPolicy, action rlsutil.PolicyAction, exprSelector func(*rlsutil.RowPolicy) string) ([]policyExprTemplate, string) {
+	templates := make([]policyExprTemplate, 0)
+	permissiveExprs := make([]string, 0)
+	restrictiveExprs := make([]string, 0)
+
+	for _, policy := range policies {
+		if !PolicyMatchesAction(policy, action) {
+			continue
+		}
+		policyExpr := strings.TrimSpace(exprSelector(policy))
+		if policyExpr == "" {
+			continue
+		}
+		var policyNeedsPrincipal bool
+		var tagVariables map[string]string
+		policyExpr, policyNeedsPrincipal, tagVariables = toTemplateExpr(policyExpr)
+		templates = append(templates, policyExprTemplate{
+			policyType:     policy.GetPolicyType(),
+			expr:           policyExpr,
+			needsPrincipal: policyNeedsPrincipal,
+			tagVariables:   tagVariables,
+		})
+		switch policy.GetPolicyType() {
+		case rlsutil.PolicyTypePermissive:
+			permissiveExprs = append(permissiveExprs, policyExpr)
+		case rlsutil.PolicyTypeRestrictive:
+			restrictiveExprs = append(restrictiveExprs, policyExpr)
+		}
+	}
+
+	if len(permissiveExprs) == 0 {
+		if len(restrictiveExprs) > 0 {
+			return templates, "false"
+		}
+		return templates, ""
+	}
+
+	groups := make([]string, 0, 2)
+	groups = append(groups, joinExprs(permissiveExprs, "or"))
+	if len(restrictiveExprs) > 0 {
+		groups = append(groups, joinExprs(restrictiveExprs, "and"))
+	}
+	return templates, joinExprs(groups, "and")
+}
+
+func toTemplateExpr(expr string) (string, bool, map[string]string) {
+	return funcutil.ConvertRLSTemplateVariables(strings.TrimSpace(expr))
+}
+
+func compileExprTemplates(schemaHelper *typeutil.SchemaHelper, templates []policyExprTemplate, visitorArgs *planparserv2.ParserVisitorArgs) (*compiledExpression, error) {
+	if len(templates) == 0 {
+		return nil, nil
+	}
+	compiled := &compiledExpression{}
+	for _, template := range templates {
+		policyExpr, err := compilePolicyExprTemplate(schemaHelper, template.expr, template.needsPrincipal, template.tagVariables, visitorArgs)
+		if err != nil {
+			return nil, err
+		}
+		if policyExpr == nil {
+			continue
+		}
+		switch template.policyType {
+		case rlsutil.PolicyTypePermissive:
+			compiled.permissive = append(compiled.permissive, policyExpr)
+		case rlsutil.PolicyTypeRestrictive:
+			compiled.restrictive = append(compiled.restrictive, policyExpr)
+		}
+	}
+	if len(compiled.permissive) == 0 && len(compiled.restrictive) == 0 {
+		return nil, nil
+	}
+	return compiled, nil
+}
+
+func compilePolicyExprTemplate(schemaHelper *typeutil.SchemaHelper, expr string, needsPrincipal bool, tagVariables map[string]string, visitorArgs *planparserv2.ParserVisitorArgs) (*compiledPolicyExpression, error) {
+	expr = strings.TrimSpace(expr)
+	if expr == "" {
+		return nil, nil
+	}
+	if schemaHelper == nil {
+		return nil, merr.WrapErrServiceInternalMsg("failed to compile RLS expression template with nil schema helper")
+	}
+	parsedExpr, err := planparserv2.ParseExprTemplate(schemaHelper, expr, visitorArgs)
+	if err != nil {
+		return nil, merr.Wrapf(err, "failed to parse RLS expression template")
+	}
+	return &compiledPolicyExpression{
+		expr:           parsedExpr,
+		needsPrincipal: needsPrincipal,
+		tagVariables:   tagVariables,
+	}, nil
+}
+
+func (e *compiledExpression) Instantiate(principalName string, principalTags map[string]string) (*planpb.Expr, error) {
+	if e == nil {
+		return nil, nil
+	}
+	if len(e.permissive) == 0 {
+		if len(e.restrictive) > 0 {
+			return alwaysFalsePredicate(), nil
+		}
+		return nil, nil
+	}
+
+	permissiveExprs, err := instantiatePolicyExprs(e.permissive, principalName, principalTags)
+	if err != nil {
+		return nil, err
+	}
+	restrictiveExprs, err := instantiatePolicyExprs(e.restrictive, principalName, principalTags)
+	if err != nil {
+		return nil, err
+	}
+
+	finalExpr := combinePredicates(permissiveExprs, planpb.BinaryExpr_LogicalOr)
+	if len(restrictiveExprs) > 0 {
+		finalExpr = combinePredicate(finalExpr, combinePredicates(restrictiveExprs, planpb.BinaryExpr_LogicalAnd), planpb.BinaryExpr_LogicalAnd)
+	}
+	return finalExpr, nil
+}
+
+func instantiatePolicyExprs(policies []*compiledPolicyExpression, principalName string, principalTags map[string]string) ([]*planpb.Expr, error) {
+	exprs := make([]*planpb.Expr, 0, len(policies))
+	for _, policy := range policies {
+		expr, err := policy.Instantiate(principalName, principalTags)
+		if err != nil {
+			return nil, err
+		}
+		if expr != nil {
+			exprs = append(exprs, expr)
+		}
+	}
+	return exprs, nil
+}
+
+func (e *compiledPolicyExpression) Instantiate(principalName string, principalTags map[string]string) (*planpb.Expr, error) {
+	if e == nil || e.expr == nil {
+		return nil, nil
+	}
+	if e.needsPrincipal && principalName == "" {
+		return alwaysFalsePredicate(), nil
+	}
+
+	values := make(map[string]*planpb.GenericValue, len(e.tagVariables)+1)
+	if e.needsPrincipal {
+		values[funcutil.RLSPrincipalTemplateName] = planparserv2.NewString(principalName)
+	}
+	for tagKey, variable := range e.tagVariables {
+		tagValue, ok := principalTags[tagKey]
+		if !ok {
+			return alwaysFalsePredicate(), nil
+		}
+		values[variable] = planparserv2.NewString(tagValue)
+	}
+
+	expr := proto.Clone(e.expr).(*planpb.Expr)
+	if err := planparserv2.FillExpressionValue(expr, values); err != nil {
+		return nil, err
+	}
+	return rewriter.RewriteExpr(expr), nil
+}
+
+func combinePredicates(exprs []*planpb.Expr, op planpb.BinaryExpr_BinaryOp) *planpb.Expr {
+	if len(exprs) == 0 {
+		return nil
+	}
+	combined := exprs[0]
+	for _, expr := range exprs[1:] {
+		combined = combinePredicate(combined, expr, op)
+	}
+	return combined
+}
+
+func combinePredicate(left *planpb.Expr, right *planpb.Expr, op planpb.BinaryExpr_BinaryOp) *planpb.Expr {
+	if left == nil {
+		return right
+	}
+	if right == nil {
+		return left
+	}
+	switch op {
+	case planpb.BinaryExpr_LogicalAnd:
+		if isAlwaysTrueExpr(left) {
+			return right
+		}
+		if isAlwaysTrueExpr(right) {
+			return left
+		}
+	case planpb.BinaryExpr_LogicalOr:
+		if isAlwaysTrueExpr(left) || isAlwaysTrueExpr(right) {
+			return alwaysTruePredicate()
+		}
+	}
+	return rewriter.RewriteExpr(&planpb.Expr{
+		Expr: &planpb.Expr_BinaryExpr{
+			BinaryExpr: &planpb.BinaryExpr{
+				Op:    op,
+				Left:  left,
+				Right: right,
+			},
+		},
+	})
+}
+
+func PolicyMatchesAction(policy *rlsutil.RowPolicy, action rlsutil.PolicyAction) bool {
+	if policy == nil {
+		return false
+	}
+	for _, policyAction := range policy.GetActions() {
+		if policyAction == action {
+			return true
+		}
+	}
+	return false
+}
+
+func joinExprs(exprs []string, op string) string {
+	nonEmpty := make([]string, 0, len(exprs))
+	for _, expr := range exprs {
+		expr = strings.TrimSpace(expr)
+		if expr != "" {
+			nonEmpty = append(nonEmpty, parenthesizeExpr(expr))
+		}
+	}
+	if len(nonEmpty) == 0 {
+		return ""
+	}
+	return strings.Join(nonEmpty, " "+op+" ")
+}
+
+func parenthesizeExpr(expr string) string {
+	return "(" + strings.TrimSpace(expr) + ")"
+}
+
+func QueryAction(isIterator bool) rlsutil.PolicyAction {
+	if isIterator {
+		return rlsutil.PolicyActionQueryIterator
+	}
+	return rlsutil.PolicyActionQuery
+}
+
+func SearchAction(isAdvanced bool, isIterator bool) rlsutil.PolicyAction {
+	if isAdvanced {
+		return rlsutil.PolicyActionHybridSearch
+	}
+	if isIterator {
+		return rlsutil.PolicyActionSearchIterator
+	}
+	return rlsutil.PolicyActionSearch
+}
+
+func ResolveRuntimePrincipal(rlsEnabled bool, principalName string, operation string) (string, bool, error) {
+	if !rlsEnabled {
+		return "", false, nil
+	}
+	if strings.TrimSpace(principalName) == "" {
+		return "", false, merr.WrapErrPrivilegeNotPermitted("%s operation denied by RLS: rls_principal is required", operation)
+	}
+	maxPrincipalNameLength := paramtable.Get().ProxyCfg.RLSMaxPrincipalNameLength.GetAsInt()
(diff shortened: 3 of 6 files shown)


internal/proxy/rls/manager.go
```diff
@@ -200,10 +231,164 @@ func (m *manager) removeCollection(_ context.Context, collectionID UniqueID) {
...
+	if compiledExpr, ok := state.compiled[key]; ok {
+		return compiledExpr, state.principalTags[principalName], nil
+	}
+
+	policies := orderedPolicies(state.policies)
+	templates, combinedExpr := preparePolicyExprTemplates(policies, action, exprSelector)
+	if maxExpressionLength := paramtable.Get().ProxyCfg.RLSMaxCombinedExpressionLength.GetAsInt(); len(combinedExpr) > maxExpressionLength {
+		return nil, nil, merr.WrapErrServiceQuotaExceededMsg("RLS combined expression exceeds max length %d", maxExpressionLength)
+	}
+	compiledExpr, err := compileExprTemplates(schemaHelper, templates, visitorArgs)
```
buqian-zilliz: **[P1] Compile policy timestamps in the collection timezone**

Compile RLS policy literals with the collection timezone, not `visitorArgs.Timezone`. RootCoord validates policies and the write-check path parses them with `schemaHelper.GetTimezone()`, while query and search parser timezones can be overridden by the requester. For an allowed Timestamptz equality policy such as `ts == '2026-01-01 00:00:00'`, changing the request timezone changes the resolved instant and can admit a different row set. Construct the policy parser arguments from the schema timezone here and remove the request timezone from the cache key.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,967 @@
...
+			return rowFloat == float64(targetInt.Int64Val), nil
+		}
+		return false, nil
+	}
+	if targetFloat, ok := target.GetVal().(*planpb.GenericValue_FloatVal); ok {
+		rowNumber, ok := numericValue(rowValue)
+		if !ok {
+			return false, nil
+		}
+		return rowNumber == targetFloat.FloatVal, nil
```
buqian-zilliz, line 820: **[P1] Evaluate FLOAT predicates at FLOAT precision**

Coerce FLOAT literals to float32 before local comparisons. FLOAT `FieldData` yields float32 values, but `GenericValue_FloatVal` retains the parsed float64 literal, so widening the row and comparing it directly makes ordinary values such as `0.1` unequal. Segcore casts the literal to float32, so read-side and write-side policy results diverge; under `not`, `not in`, or a negated contains expression, this mismatch can flip to true and admit a write that the policy should reject. Apply the same coercion in `scalarArrayContains` for FLOAT arrays.

internal/proxy/rls/manager.go
```diff
@@ -200,10 +231,164 @@ func (m *manager) removeCollection(_ context.Context, collectionID UniqueID) {
...
+	if visitorArgs != nil {
+		key.timezone = visitorArgs.Timezone
+	}
+
+	state.mu.RLock()
+	if len(state.policies) == 0 {
+		state.mu.RUnlock()
+		return nil, nil, nil
+	}
+	if compiledExpr, ok := state.compiled[key]; ok {
```
buqian-zilliz: **[P2] Honor lowered expression limits on cache hits**

Revalidate the refreshable combined-expression limit on cache hits. The only `RLSMaxCombinedExpressionLength` check is below this early return, so after an expression is cached, lowering `proxy.rls.maxCombinedExpressionLength` has no effect until the policy snapshot changes; warm and cold proxies can then enforce different limits. Store the combined length with each cache entry and compare it on every hit, or invalidate/version the cache when this configuration changes.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,967 @@
...
+	return rewriter.RewriteExpr(expr), nil
+}
+
+func combinePredicates(exprs []*planpb.Expr, op planpb.BinaryExpr_BinaryOp) *planpb.Expr {
+	if len(exprs) == 0 {
+		return nil
+	}
+	combined := exprs[0]
+	for _, expr := range exprs[1:] {
+		combined = combinePredicate(combined, expr, op)
```
buqian-zilliz: **[P2] Rewrite each policy group only once**

`combinePredicates` runs `RewriteExpr` after every appended policy, so the ith append traverses and rebuilds the entire i-policy prefix. With up to 100 policies, every cache-hit instantiation can perform O(P²) AST work and allocate repeated intermediate trees on the query/search hot path. Build the complete OR or AND tree, preferably balanced, and invoke the rewriter once after the group is complete.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,967 @@
...
+		return truthUnknown, merr.WrapErrServiceInternalMsg("RLS term expression does not support field-to-field IN")
+	}
+	rowValue, err := rowData.value(expr.GetColumnInfo(), rowIdx)
+	if err != nil {
+		return truthUnknown, err
+	}
+	if rowValue == nil {
+		return truthUnknown, nil
+	}
+	for _, value := range expr.GetValues() {
```
buqian-zilliz, line 690: **[P2] Precompute IN membership for batch validation**

This loop linearly scans every literal for every row and every applicable `IN` term. The configured limit permits up to 1,024 literals, so large insert or upsert batches can spend O(rows × literals × policies) comparisons even though the parsed predicate is reused for the entire batch. Precompute a typed membership set, or use the already sorted term values with logarithmic lookup, once per instantiated predicate.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,967 @@
...
+			return truthUnknown, merr.WrapErrServiceInternalMsg("RLS array_contains expression requires exactly one element")
+		}
+		contains, err := scalarArrayContains(arrayValue, expr.GetElements()[0])
+		if err != nil {
+			return truthUnknown, err
+		}
+		return truthValueFromBool(contains), nil
+	case planpb.JSONContainsExpr_ContainsAll:
+		for _, element := range expr.GetElements() {
+			contains, err := scalarArrayContains(arrayValue, element)
```
buqian-zilliz, line 727: **[P2] Scan row arrays once for contains-all and contains-any**

`array_contains_all` and `array_contains_any` call `scalarArrayContains` once per requested element, and each call rescans the complete row array. This makes each row O(literal_elements × array_length) for a valid policy containing a large literal list. Build a typed lookup structure for the row array once, or precompile the target set and scan the row array once, so the work is O(literal_elements + array_length).

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,967 @@
...
+		return 0, false, merr.WrapErrServiceInternalMsg("RLS row index %d exceeds valid data length %d", rowIdx, len(validData))
+	}
+	if !validData[rowIdx] {
+		return 0, false, nil
+	}
+	if dataLen == len(validData) {
+		return rowIdx, true, nil
+	}
+	dataIdx := 0
+	for i := 0; i < rowIdx; i++ {
```
buqian-zilliz, line 964: **[P2] Precompute compact nullable-array offsets**

For compact nullable array data, every access recounts valid entries from row zero to `rowIdx`. Because row validation already iterates rows sequentially, one referenced array field becomes O(N²), multiplied by repeated array predicates. Precompute logical-to-physical offsets once in `newRowData` or `fieldReader`, or maintain a forward cursor, as the scalar nullable iterator already does.

internal/proxy/rls/snapshot.go
```diff
@@ -135,3 +136,26 @@ func (m *manager) refreshSnapshotsUnlocked(ctx context.Context, coord CoordClien
 	}
 	return nil
 }
+
+func rowPoliciesFromInfo(policies []*rootcoordpb.RLSPolicyInfo) []*rlsutil.RowPolicy {
```
zhengbuqian, line 140: why do we need this `rlsutil.RowPolicy` struct? can we just use `rootcoordpb.RLSPolicyInfo`?

zhengbuqian: this is more a comment for the previous PR.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1238 @@
...
+		return containsValidArrayElement(data.StringData.GetData(), validData, func(value string) bool {
+			return value == targetValue.StringVal
+		})
+	default:
+		return false, merr.WrapErrServiceInternalMsg("unsupported RLS array element type %T", data)
+	}
+}
+
+func containsValidArrayElement[T any](values []T, validData []bool, matches func(T) bool) (bool, error) {
+	if len(validData) > 0 && len(validData) != len(values) {
```
buqian-zilliz: **[P2] Guard the element-array layout before evaluation**

`containsValidArrayElement` only accepts the post-`fillWithValue` dense layout where `len(validData) == len(values)`, while the Proxy input boundary accepts compact element-nullable payloads where `CountValidRows(validData) == len(values)`. Neither `ValidateCheckForWrite` nor `ValidateRowsByPredicate` documents or enforces that callers must run the fill step first, so wiring the check before normalization turns valid null-containing arrays into internal errors. Accept both layouts, or make the dense-layout precondition explicit and enforce the call order at this API boundary.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1238 @@
...
+
+	return ValidateRowsByPredicate(ctx, fieldsData, rowNum, parsedExpr, operation, exprKind)
+}
+
+func ValidateRowsByPredicate(ctx context.Context, fieldsData []*schemapb.FieldData, rowNum int, parsedExpr *planpb.Expr, operation string, exprKind string) error {
+	if parsedExpr == nil || rowNum == 0 {
+		return nil
+	}
+
+	rowData := newRowData(fieldsData)
```
buqian-zilliz: **[P2] Build readers only for fields referenced by the predicate**

`newRowData` eagerly constructs a reader for every submitted field even though `parsedExpr` identifies the field IDs actually used by the check. Compact nullable scalars make `GetDataIterator` allocate a row-sized index slice, and every nullable array allocates `arrayDataIndices` even for a full-size payload where it is only an identity map. On wide insert or upsert batches this adds `O(fields × rows)` preprocessing and peak memory before evaluating the first row. Build readers only for `ReferencedFieldIDs(parsedExpr)` or initialize them lazily, and use direct indexing for full-size arrays.

internal/proxy/rls/predicate.go
```diff
@@ -0,0 +1,213 @@
...
+	compiled, err := state.getCompiledExpression(action, kind, schema)
+	if err != nil {
+		return nil, err
+	}
+	if compiled == nil {
+		return nil, denyNoApplicableRLSPolicy(action, kind)
+	}
+
+	var tags map[string]rlsutil.TagValue
+	if compiled.needsTags() {
```
buqian-zilliz, line 96: **[P2] Skip tag RPCs for restrictive-only policy sets**

`compiledExpression.Instantiate` returns an unconditional false predicate when there is no permissive policy, so restrictive-policy tags cannot affect the result. However, `needsTags` scans both groups and this branch refreshes principal metadata first. With a tag-based restrictive-only set, an unstored principal is not negative-cached, so every denied request can issue a RootCoord RPC and may surface a service-unavailable error instead of the deterministic RLS denial. Short-circuit before tag resolution, or make `needsTags` return false when the permissive group is empty.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1238 @@
...
+	if err != nil {
+		return err
+	}
+	if checkExpr == nil {
+		return nil
+	}
+	return ValidateRowsByPredicate(ctx, fieldsData, rowNum, checkExpr, operation, "check")
+}
+
+func validateRows(ctx context.Context, fieldsData []*schemapb.FieldData, schemaHelper *typeutil.SchemaHelper, rowNum int, expr string, operation string, exprKind string) error {
```
buqian-zilliz: **[P3] Remove the unused raw-expression evaluator path**

`validateRows` has no non-test caller; the new write entry point resolves a compiled check through `resolveCheckPredicate` and calls `ValidateRowsByPredicate` directly. Keeping this second raw-string parsing path in production also expands the evaluator beyond the validated policy contract, including ordered and `NotEqual` comparisons that `ValidateParsedExpression` rejects. Move any test-only convenience into test code and remove this path and its unsupported comparison branches.

internal/proxy/rls/util.go
buqian-zilliz: This issue still exists: `evalTermExpr` still scans every literal for every row, so batch validation remains linear in the literal count for each row and policy.

internal/proxy/rls/util.go
buqian-zilliz: This issue still exists: `ContainsAll` and `ContainsAny` still call `scalarArrayContains` once per requested element, rescanning the row array for each literal.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1322 @@
...
+	tagVariables   map[string]string
+}
+
+func (e *compiledExpression) needsTags() bool {
+	if e == nil || len(e.permissive) == 0 {
+		return false
+	}
+	for _, policies := range [...][]*compiledPolicyExpression{e.permissive, e.restrictive} {
+		for _, policy := range policies {
+			if len(policy.tagVariables) > 0 {
```
buqian-zilliz, line 55: **[P2] Avoid tag RPCs when constant policies decide the result**

`needsTags` treats every tag-bearing policy as required before simplifying the policy groups. For example, a permissive `true` policy OR a tag-based permissive policy, with no restrictive policies, is unconditionally true, but this code still refreshes principal metadata first. After the cache TTL expires, a RootCoord outage or refresh backoff can therefore reject an operation whose authorization result does not depend on tags. Simplify the permissive and restrictive groups before calculating tag dependencies, or derive the required tags from the simplified combined expression.

internal/proxy/rls/predicate.go
buqian-zilliz, line 96: **[P2] Cache successful empty principal-tag lookups**

`ensurePrincipalTags` represents a successful lookup with no principal tag binding by returning an empty map without publishing a `principalTagsEntry`. Consequently, every subsequent request that genuinely needs tags for that principal issues another `GetRLSMetadata` RPC after the previous in-flight call completes. High-cardinality traffic from tagless principals can therefore create a per-request RootCoord path. Cache an entry containing an immutable empty map and `refreshedAt`; the existing TTL and principal invalidation paths can refresh it safely.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1322 @@
...
+			return alwaysFalsePredicate(), nil
+		}
+		normalizedTagValue, ok := normalizeRLSTagValue(e.expr, variable, tagValue)
+		if !ok {
+			return alwaysFalsePredicate(), nil
+		}
+		values[variable] = rlsTagValueToGenericValue(normalizedTagValue)
+	}
+
+	expr := proto.Clone(e.expr).(*planpb.Expr)
```
buqian-zilliz, line 328: **[P2] Reuse principal-independent policy ASTs**

`Instantiate` always allocates a values map, deep-clones the protobuf expression, fills it, and rewrites it, even when the policy references neither the current principal nor any tags. Literal-only policies are valid, so a request can repeat this work for up to 100 policies on the query or write hot path despite the compiled cache. Store a pre-instantiated immutable AST for variable-free policies, or cache the final variable-free group, and clone only subtrees whose template values must be filled.

internal/proxy/rls/predicate.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, merr.WrapErrServiceQuotaExceededMsg("RLS combined expression exceeds max length %d", maxLength)
+	}
+	compiled, err := compileExprTemplates(schema, templates, timezone, kind)
+	if err != nil {
+		return nil, err
+	}
+	if state.compiled == nil {
+		state.compiled = make(map[compiledKey]*compiledCacheEntry)
+	}
+	state.compiled[key] = &compiledCacheEntry{
```
buqian-zilliz: **[P2] Compile each policy once per schema snapshot**

`compiled` is keyed by action and retains a separately parsed copy of every applicable policy expression for each key. A policy covering every operation can therefore retain seven parsed copies of its using expression and two copies of its check expression. With up to 100 policies, 16 KiB of combined source per entry, and no idle collection eviction, compilation CPU and retained memory scale with actions times collections instead of policies times collections. Compile each policy expression once per kind and schema snapshot, then let action-specific entries contain lightweight references to those immutable compiled policies.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1322 @@
...
+	}
+	if matcher := rowData.termMatcher(expr); matcher != nil {
+		if _, matched := matcher.index(rowValue); matched {
+			return truthTrue, nil
+		}
+		if _, usable := scalarLiteralKey(matcher.dataType, rowValue); usable {
+			return truthFalse, nil
+		}
+	}
+	for _, value := range expr.GetValues() {
```
buqian-zilliz: **[P3] Remove the duplicate fallback evaluator paths**

`ParseExprTemplate` casts term literals, the parser constrains array literals, and `ValidateParsedExpression` restricts the supported field kinds before these evaluators run. Therefore `termMatcher` or `arrayMatcher` can be nil only for a malformed internal AST, but the fallback branches implement a second set of term and array-contains semantics for that state, including the old repeated-scan path. Make matcher construction return a data-integrity error and keep only the cached matcher evaluation path; `valueEqual` remains available for unary equality.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. Variable-free policy groups now eliminate canonical true/false before tag dependency calculation, independent of enabledOptimizeExpr. Decisive permissive-true and restrictive-false paths therefore skip principal metadata refresh; optimizer-disabled mixed-policy regressions cover both cases.

internal/proxy/rls/predicate.go
aoiasd: Fixed in 428f87d. A successful no-binding lookup now publishes a timestamped empty principalTagsEntry. Existing TTL, notification invalidation, and refresh fencing handle refresh; the regression covers miss, cache hit, invalidation, and reload.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. Variable-free policies are combined into one cached static group and skip values allocation, fill, and per-policy rewrite. The combined group is cloned once per request because rewrite and plan merge mutate ASTs, avoiding shared-AST races.

internal/proxy/rls/predicate.go
aoiasd: Acknowledged; deferring this bounded optimization. Compilation is lazy per action and kind, policies are capped at 100, combined source is capped at 16 KiB, and DropCollection removes the full state. A per-policy cache adds another schema/config invalidation identity; we will revisit it if profiling shows action-by-collection retention is material.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. Matcher construction now returns ErrDataIntegrity for malformed internal ASTs, term and ARRAY evaluation uses only the cached matcher path, and the duplicate linear fallback helpers were removed.

internal/proxy/rls/predicate.go
```diff
@@ -0,0 +1,213 @@
...
+		state.mu.RUnlock()
+		return nil, nil
+	}
+	if entry := state.compiled[key]; entry.matches(schemaVersion, timezone, maxLength) {
+		state.mu.RUnlock()
+		return entry.expression, nil
+	}
+	state.mu.RUnlock()
+
+	state.mu.Lock()
```
buqian-zilliz: **[P3] Compile policy expressions outside the collection lock**

`getCompiledExpression` holds `state.mu` exclusively while sorting, converting, parsing, and schema-validating every applicable policy. If a policy or principal invalidation arrives during a cold compilation, it acquires the manager-wide `m.mu` and then waits for this lock, causing requests and metadata operations for unrelated collections to block behind it as well. Snapshot the immutable policy inputs under `state.mu`, compile outside the lock, and reacquire it only to verify that the snapshot is still current and publish the cache entry. The publication fence must cover normal TTL snapshot replacement as well as `policyGeneration` changes.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. IN predicates now build one typed, batch-local membership matcher and use O(1) lookup per row; FLOAT normalization remains float32. Added matcher-cache and precision regressions.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. ARRAY contains predicates cache the typed target matcher and scan each row array once; contains-all uses generation-stamped state without per-row allocation. Duplicate, empty, nullable, and cross-row cases are covered.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. Compact nullable ARRAY readers precompute the logical-to-physical row map once and use O(1) indexing; dense rows use direct indexing.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. The shared element traversal accepts both exact dense and exact compact-valid layouts and rejects every other shape. Dense, compact, and malformed regressions cover the boundary.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. Row preparation now builds readers only for ReferencedFieldIDs; dense ARRAY input uses direct indexing and only compact input allocates a row map.

internal/proxy/rls/predicate.go
aoiasd: Fixed in 428f87d. A restrictive-only set is recognized as deterministically false before tag resolution, so it denies without a principal metadata RPC. Added an RPC-failure regression asserting zero calls.

internal/proxy/rls/util.go
aoiasd: Fixed in 428f87d. The production raw-string evaluator path and unsupported ordered/NotEqual branches were removed; test-only parsing remains in test code, and production evaluation accepts only validated compiled predicates.

internal/proxy/rls/manager.go
aoiasd: Fixed on the current branch. Policy parsing now uses only the collection schema timezone, and the compiled-cache identity includes both schema version and timezone. The regressions cover request-timezone overrides and timezone changes without a schema-version bump.

internal/proxy/rls/util.go
aoiasd: Fixed on the current branch. FLOAT literals are normalized to float32 for scalar equality, IN matchers, and ARRAY contains matchers, matching FieldData and Segcore precision. The 0.1 regressions cover all paths.

internal/proxy/rls/manager.go
aoiasd: Fixed on the current branch. The refreshable maxCombinedExpressionLength value is part of the compiled-cache identity, so a warm entry misses immediately after the limit changes. The regression covers lowering the limit after a successful cache hit.

internal/proxy/rls/util.go
aoiasd: Fixed on the current branch. Policies are assembled into complete balanced groups and RewriteExpr is called once on the final instantiated expression, rather than after every append.

internal/proxy/rls/predicate.go
aoiasd: Fixed in c3fe6b6. The slow path now snapshots immutable policy pointers and their generation under the collection lock, then sorts, converts, parses, and validates outside that lock. Publication rechecks the generation, and successful TTL snapshot replacement now advances it as well as explicit invalidation. A collection-local compile mutex coalesces concurrent cold misses without being acquired by metadata invalidation; the regression verifies invalidation does not wait for compilation.

internal/proxy/rls/snapshot.go
```diff
@@ -209,21 +214,18 @@ func (m *manager) ensurePrincipalTags(ctx context.Context, collectionID UniqueID
 		if err != nil {
 			return nil, merr.WrapErrDataIntegrity(err, "decode RLS principal %q tags", principalName)
 		}
-		var entry *principalTagsEntry
-		if tags != nil {
-			entry = &principalTagsEntry{
-				refreshedAt: time.Now(),
-				tags:        tags,
-			}
+		if tags == nil {
```
buqian-zilliz, line 244: **[P2] Bound negative principal-cache growth**

`ensurePrincipalTags` now publishes a `principalTagsEntry` even when RootCoord has no matching principal, but `principalTags` has no count or byte bound. `ResolveRuntimePrincipal` accepts runtime names up to the 64 KiB transport limit, so high-cardinality miss traffic can retain one large key and empty map per name for the full default one-hour TTL, issue one RootCoord lookup per name, and make the scanner traverse the entire map while holding `state.mu`. Bound negative entries per collection, for example with a count- and byte-limited LRU, or cap their lifetime and cardinality separately from positive entries.

internal/proxy/rls/predicate.go
```diff
@@ -0,0 +1,228 @@
...
+		state.mu.Lock()
+		if state.policyGeneration != generation {
+			state.mu.Unlock()
+			continue
+		}
+		if entry := state.compiled[key]; entry.matches(schemaVersion, timezone, maxLength) {
+			state.mu.Unlock()
+			return entry.expression, nil
+		}
+		if compileErr != nil {
```
buqian-zilliz: **[P2] Cache deterministic compilation failures**

When compilation fails deterministically, this branch returns without publishing a cache entry. Every subsequent request therefore copies and sorts the policy snapshot and reparses the same expressions while queued behind the collection-wide `policyCompileMu`. Lowering the refreshable combined-expression limit below an existing snapshot, or encountering a schema incompatibility, leaves the collection in this serialized compile loop until its policy or schema changes. Cache the error alongside the generation, schema version, timezone, and maximum-length identity, and invalidate it through the existing generation path.

internal/rootcoord/meta_table.go
```diff
@@ -3027,9 +3026,14 @@ func validateRLSPolicyExpression(schemaHelper *typeutil.SchemaHelper, exprKind s
 	if err != nil {
 		return merr.WrapErrParameterInvalidErr(err, "invalid RLS %s expression", exprKind)
 	}
-	if err := validateRLSParsedExpr(parsedExpr, allowedTemplateVariables); err != nil {
+	if err := rlsutil.ValidateParsedExpression(parsedExpr, allowedTemplateVariables); err != nil {
```
buqian-zilliz, line 2994: **[P2] Migrate policies accepted by the previous validator**

`validateRLSPoliciesWithSchema` revalidates persisted policies during field and function DDL through the stricter shared validator. Before this change, policies using tag templates against boolean or timestamptz fields and `using_expr` on element-nullable arrays were accepted; `ValidateParsedExpression` and `ValidateUsingExpressionSchema` now reject them. Such metadata has no migration or upgrade preflight, so runtime compilation will fail closed after upgrade and an otherwise compatible schema or function alteration will also be rejected. Add an explicit migration or preflight for existing policies, or version the stored expression contract and handle the legacy forms at the serving boundary.

aoiasd: This does not apply to a supported upgrade path. RLS1/RLS2 contained internal metadata scaffolding, but no RLS mutation method was exposed by Proxy or RootCoord RPC in any release; RootCoord exposed only the read-side GetRLSMetadata snapshot. Therefore no supported deployed writer could create a legacy persisted population under the older validator. RLS3 establishes the first serveable expression contract, while manually injected or corrupt metadata continues to fail closed, so no migration is required.

internal/proxy/rls/snapshot.go
aoiasd: Fixed in a18079ba. The per-collection principal-tag cache now has refreshable limits of 65,536 entries and 64 MiB of logical payload by default, accounting for the principal name plus tag keys and values. This covers both positive entries and negative misses; an oversized lookup still serves the current request but is not retained. Publication, FIFO eviction, TTL cleanup, and explicit invalidation share the same locked accounting helpers, with regressions for count/byte bounds and dynamic limit reduction.

internal/proxy/rls/predicate.go
aoiasd: Fixed in a18079ba. Deterministic ServiceQuotaExceeded and DataIntegrity compilation failures are cached under the same policy generation, schema version, timezone, and maximum-length identity as successful results. Transient/internal failures remain uncached. Tests cover recovery after a limit change or policy replacement and verify that internal failures are retried.

internal/proxy/rls/manager.go
```diff
@@ -414,6 +437,63 @@ func principalRefreshKey(key principalKey, token principalRefreshToken) string {
...
+// principal cache within its count and logical-payload byte limits.
+func (state *collectionState) putPrincipalTagsLocked(principalName string, entry *principalTagsEntry) {
+	state.removePrincipalTagsLocked(principalName)
+	if entry == nil {
+		return
+	}
+	entry.cacheBytes = principalTagsCacheBytes(principalName, entry.tags)
+	maxBytes := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheBytes.GetAsInt64()
+	if entry.cacheBytes <= maxBytes {
+		entry.orderElement = state.principalCacheOrder.PushBack(principalName)
```
buqian-zilliz: **[P2] Protect hot principals from negative-cache churn**

`principalCacheOrder` is insertion-ordered: cache hits returned by `getPrincipalTagsEntry` never move `orderElement`, so every new successful miss evicts the oldest entry once either limit is full, even if that entry is a heavily used real principal. Because runtime principal names may be 64 KiB, roughly 1,024 distinct misses can consume the default 64 MiB budget and flush older positive entries; legitimate requests then fall back to RootCoord and can fail on refresh backoff or an outage. Promote entries on hits, or keep negative entries in a separately bounded cache, so miss churn cannot evict active principals.

aoiasd: Fixed in 33e5d866. Positive and negative principal entries still share the same per-collection count/byte budget, but use separate insertion queues and trimming always evicts negative entries first. If the cache is full of positive entries, a new miss evicts itself instead of displacing a real principal, while cache hits remain read-only and avoid a collection write lock. Regressions cover both non-empty and existing-empty positive principals under negative pressure.

internal/proxy/rls/snapshot.go
```diff
@@ -200,30 +205,30 @@ func (m *manager) ensurePrincipalTags(ctx context.Context, collectionID UniqueID
...
+			tags = map[string]rlsutil.TagValue{}
+		}
+		entry := &principalTagsEntry{
+			refreshedAt: time.Now(),
+			tags:        tags,
+			missing:     !found,
```
buqian-zilliz: **[P2] Bound negative-cache memory across collections**

`missing: !found` makes successful lookups of nonexistent principals cacheable for up to the metadata TTL, while both new limits are enforced independently per collection and `manager.collections` has no aggregate budget. Runtime principal names are accepted up to 64 KiB, so roughly 1,024 distinct misses can consume the default 64 MiB allowance in each collection; 100 active collections can therefore retain about 6.4 GiB per Proxy before map/list overhead. Add a process-wide bound for negative entries, or a much smaller dedicated negative quota, so collection count cannot multiply this allowance.

aoiasd: Thanks, this is a valid aggregate-scale concern. In this iteration we intentionally scope the cache budget per collection and are not targeting a very large number of concurrently RLS-enabled collections. The required invariants here are that each collection is bounded and that dropping it releases all cache state. The implementation enforces both entry and logical-byte limits per collection, where logical bytes include the principal name plus all tag keys and values. `markCollectionDropped` removes the entire collection state, and refresh publication is fenced so an in-flight refresh cannot recreate it after drop. We will consider a process-wide or dedicated negative-cache budget when we expand the supported collection-scale target, rather than add cross-collection eviction policy in this PR.

internal/proxy/rls/predicate.go
```diff
@@ -0,0 +1,238 @@
...
+	checkExprKind
+)
+
+type compiledKey struct {
+	action rlsutil.PolicyAction
+	kind   exprKind
+}
+
+type compiledCacheEntry struct {
+	generation    uint64
```
buqian-zilliz: **[P3] Remove the redundant generation from compiled cache entries**

`state.compiled` is cleared at both sites that increment `policyGeneration`, and the slow path already rechecks `state.policyGeneration` before publishing a result. Consequently, every entry remaining in this map necessarily belongs to the current generation; storing `generation` again in each `compiledCacheEntry` and threading it through `matches` adds a second invalidation invariant without distinguishing cache identities. Remove this field and retain the generation only as the out-of-lock compilation fence; cache hits need to compare only the schema version, timezone, and active length limit.

aoiasd: Fixed in 8527d873. `compiledCacheEntry` now contains only schema version, timezone, expression-length limit, result, and deterministic error. `policyGeneration` remains solely as the out-of-lock compilation/publication fence; both generation-advance sites clear the whole compiled map. The same change also switches the `errors.Is` call to the repository-approved CockroachDB errors package, resolving the depguard Code Check failure.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1312 @@
...
+		return err
+	}
+	if checkExpr == nil {
+		return nil
+	}
+	return ValidateRowsByPredicate(ctx, fieldsData, rowNum, checkExpr, operation, "check")
+}
+
+func ValidateRowsByPredicate(ctx context.Context, fieldsData []*schemapb.FieldData, rowNum int, parsedExpr *planpb.Expr, operation string, exprKind string) error {
+	if parsedExpr == nil || rowNum == 0 {
```
buqian-zilliz: **[P2] Validate the full FieldData row count**

`ValidateRowsByPredicate` returns success when `rowNum == 0`, and negative values also skip the loop; more generally, it never checks that `rowNum` matches the logical row count recorded by `newRowData`. A caller that supplies a smaller count therefore leaves trailing rows completely unchecked, which can admit writes that do not satisfy the RLS check. Derive the count from the normalized `FieldData`, or reject negative values and every count/data mismatch, including zero with non-empty fields, before evaluating rows.

aoiasd: Fixed in a959e61d. ValidateRowsByPredicate now rejects negative counts and zero/non-empty or positive/empty contradictions, builds the normalized readers once, and requires every predicate-referenced scalar/ARRAY reader to have an exact logical row count and valid nullable layout before evaluating any row. The regressions cover smaller counts (including a denied trailing row), zero, negative, missing/oversized data, compact nullable arrays, and unrelated fields that remain owned by the existing payload-alignment layer.

internal/proxy/rls/predicate.go
```diff
@@ -0,0 +1,237 @@
...
+	expression    *compiledExpression
+	err           error
+}
+
+func (entry *compiledCacheEntry) matches(schemaVersion int32, timezone string, maxLength int) bool {
+	return entry != nil && entry.schemaVersion == schemaVersion && entry.timezone == timezone && entry.maxLength == maxLength
+}
+
+// ResolveUsingPredicate resolves the row filter for an RLS-protected operation.
+func ResolveUsingPredicate(ctx context.Context, collectionID UniqueID, principalName string, action rlsutil.PolicyAction, schema *typeutil.SchemaHelper) (*planpb.Expr, error) {
```
buqian-zilliz, line 58: **[P3] Avoid landing the evaluator without a production consumer**

`ResolveUsingPredicate`, `MergePredicateToPlan`, and `ValidateCheckForWrite` have no non-test production callers at this head, so the 1,595 lines added across `predicate.go` and `util.go` do not participate in any request path. This leaves a large dormant API whose real consumer requirements cannot be validated and may force churn when the later wiring arrives. Move the compiler and evaluator into the request-path PR that calls them, or add those production call sites here.

aoiasd: This split is intentional for the current two-PR stack. PR #52074 establishes the tested policy compiler/evaluator boundary, and its direct child #52075 wires these APIs into all production request paths. Moving only a subset of consumers here would introduce partial enforcement, while folding the whole child would remove the review boundary. This layer is safe to merge inertly; user-visible RLS enforcement is only claimed once #52075 lands.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1358 @@
...
+		}
+		return value, nil
+	default:
+		return nil, merr.WrapErrServiceInternalMsg("RLS expression references unsupported field %s with type %s", reader.field.GetFieldName(), reader.field.GetType().String())
+	}
+}
+
+func isRLSScalarType(dataType schemapb.DataType) bool {
+	switch dataType {
+	case schemapb.DataType_Bool,
```
buqian-zilliz, line 1059: **[P2] Reject String policies that the write evaluator cannot execute**

`ValidateParsedExpression` admits `schemapb.DataType_String` because `typeutil.IsPrimitiveType` and `typeutil.IsStringType` classify it as a supported scalar, so RootCoord can persist a `check_expr` for a legacy String field. The write evaluator omits `String` here and from `rowData.value` and `rlsScalarDataLen`; the underlying `typeutil.GetDataIterator` also has no `String` case. Evaluating such a policy therefore fails every checked write with an internal unsupported-field error instead of applying the predicate. Either reject `String` during shared policy admission and schema compatibility validation, or support and normalize it consistently throughout the evaluator.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1358 @@
...
+		return nil, nil
+	}
+	if e.isStatic() {
+		return proto.Clone(e.expr).(*planpb.Expr), nil
+	}
+	if e.needsPrincipal && principalName == "" {
+		return alwaysFalsePredicate(), nil
+	}
+
+	values := make(map[string]*planpb.GenericValue, len(e.tagVariables)+1)
```
buqian-zilliz: **[P2] Short-circuit missing-tag policies before constructing ASTs**

`compiledPolicyExpression.Instantiate` constructs the `values` map before confirming that every required tag exists, and each missing tag returns a freshly allocated false-predicate tree. `instantiatePolicyExprs` then retains one tree per policy, combines up to the default 100-policy limit, and runs the rewriter even though the group result is already known; negative-cache hits make this RPC-free path cheap to repeat. Validate and normalize required tags before allocating, represent a false policy without an AST, omit it from permissive OR groups, and short-circuit restrictive AND groups so at most one final false expression is materialized.

internal/proxy/rls/util.go
aoiasd: Fixed in e7808fd8c7. Shared RLS expression validation now rejects deprecated top-level DataType_String fields, so policy admission, schema compatibility checks, and Proxy compilation of persisted policies fail consistently before the write evaluator. Added check_expr/Insert, IN, and existing-policy schema tests; Array<String> remains unchanged because its evaluator path is supported.

internal/proxy/rls/util.go
aoiasd: Fixed in e7808fd8c7. Missing or incompatible tags are now represented without an AST: restrictive AND policies short-circuit, permissive OR policies omit false branches, and an all-false result materializes exactly one final false predicate. Tag normalization happens once before allocating the values map or cloning the policy AST. Added optimizer-disabled shape and fail-closed tests for partial/all permissive misses and restrictive misses.

fadedthedemon104-boop (approved):

internal/proxy/rls/manager.go
```diff
@@ -252,9 +278,10 @@ func (m *manager) expirePrincipalTags(now time.Time) {
 		state.mu.Lock()
 		for principalName, entry := range state.principalTags {
 			if entry == nil || !entry.refreshedAt.Add(refreshTTL).After(now) {
-				delete(state.principalTags, principalName)
+				state.removePrincipalTagsLocked(principalName)
```
buqian-zilliz: **[P2] Expire principal entries without scanning the full cache**

With the new negative cache, `principalTags` can reach 65,536 entries per collection. This scanner traverses the entire map every ten minutes while holding `state.mu` exclusively; with the default one-hour TTL, the first five scans examine every entry without expiring anything, blocking cache hits and adding `O(total cached principals)` periodic work across collections. The positive and negative queues are already in insertion/refresh order, so expire entries from each queue's front until the first fresh entry, while retaining the separate trimming step for refreshed capacity limits.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1382 @@
...
+	case *schemapb.ScalarField_StringData:
+		return containsValidArrayElement(data.StringData.GetData(), validData, func(value string) bool { return visit(value) })
+	default:
+		return false, merr.WrapErrServiceInternalMsg("unsupported RLS array element type %T", data)
+	}
+}
+
+func containsValidArrayElement[T any](values []T, validData []bool, matches func(T) bool) (bool, error) {
+	if len(validData) > 0 && len(values) != len(validData) {
+		validRows := int(funcutil.CountValidRows(validData))
```
buqian-zilliz, line 1367: **[P2] Avoid recounting array validity for every policy**

`containsValidArrayElement` runs for every row and every applicable `array_contains*` policy. For compact element data, it calls `CountValidRows` over the full validity bitmap before scanning the values, so up to 100 policies on the same array can rescan the same 4,096-element bitmap 100 times per row. Validate and cache each `ScalarField` row's compact/full layout once in `rowData`, or memoize it on first access, and reuse that result across policy evaluations.

aoiasd: Fixed in f6bc3b583e. `rowData` now memoizes each `ScalarField` element layout across all policies evaluating the same row, including cached malformed-layout errors. The map is cleared once at the start of each outer row, so it cannot grow with the write batch. Tests cover dense, compact, malformed, and cross-matcher reuse.

internal/proxy/rls/manager.go
aoiasd: Fixed in f6bc3b583e. Expiration now removes expired prefixes from the positive and negative refresh-ordered queues and stops at the first fresh entry. Refresh timestamps are assigned while holding the collection-state lock so queue order matches freshness; capacity trimming remains a separate step.

internal/rootcoord/meta_table.go
```diff
@@ -3567,9 +3443,24 @@ func (mt *MetaTable) ListRLSPrincipals(ctx context.Context, req *rlsutil.ListRLS
 		return nil, err
 	}
 
-	names := lo.Map(coll.RLSPrincipals, func(principal *model.RLSPrincipal, _ int) string {
-		return principal.PrincipalName
-	})
+	principals, err := mt.catalog.ListRLSPrincipals(ctx, coll.CollectionID)
```
buqian-zilliz, line 3445: **[P2] Paginate principal listing before removing the storage bound**

`ListRLSPrincipals` now calls a catalog method that performs `LoadWithPrefix`, unmarshals every principal protobuf and tag map, and sorts the complete result before this method allocates and sorts another names slice. This PR simultaneously removes `maxPrincipalsPerCollection`, while the list request has no page size or continuation token. A single request, or `GetRLSMetadata` with `RLS_METADATA_KIND_ALL`, can therefore read and materialize an unbounded tag corpus and exceed RootCoord memory or RPC limits. Retain a creation bound until listing is paginated, or add a paged, key-only catalog path for these operations.

internal/proxy/rls/manager.go
```diff
@@ -250,27 +284,23 @@ func (m *manager) expirePrincipalTags(now time.Time) {
...
-			}
-		}
-		for principalName, backoff := range state.principalBackoffs {
-			if backoff == nil || !now.Before(backoff.NextInstant()) {
-				delete(state.principalBackoffs, principalName)
-			}
-		}
+		state.expirePrincipalTagsLocked(state.positivePrincipalCacheOrder, refreshTTL, now)
+		state.expirePrincipalTagsLocked(state.negativePrincipalCacheOrder, refreshTTL, now)
+		state.trimPrincipalStateLocked()
```
buqian-zilliz, line 291: **[P2] Expire refresh backoffs after their retry window**

The periodic scanner now expires tag entries but never removes a `principalBackoffEntry` whose `NextInstant` has passed; `trimPrincipalStateLocked` only removes such entries when a configured capacity is exceeded. A transient failure for each of many principals can therefore retain the full per-collection backoff map, list nodes, timers, and names indefinitely when those principals are not requested again. Restore time-based cleanup from `principalBackoffOrder`, or remove expired backoff state before starting a later refresh.

internal/metastore/model/collection.go
```diff
@@ -42,9 +42,9 @@ type Collection struct {
 	Fields            []*Field
 	StructArrayFields []*StructArrayField
 	Functions         []*Function
-	// RLS metadata is an internal RootCoord cache. It is persisted in its own
-	// KV namespace and must not be marshaled into collection info.
-	RLSPolicies          map[string]*RLSPolicy
+	// RLS policies are cached by RootCoord and persisted in their own KV namespace.
+	RLSPolicies map[string]*RLSPolicy
+	// RootCoord leaves RLSPrincipals empty and reads principal tags from the catalog on demand.
```
buqian-zilliz: **[P3] Remove the permanently empty principal field**

The new invariant says RootCoord never populates `Collection.RLSPrincipals`, and this PR removes its remaining production population and read sites, but the field and its `ShallowClone`, `Clone`, and `ApplyUpdates` plumbing remain. This preserves a second, impossible representation beside the catalog and invites future code to read an always-empty cache. Remove `RLSPrincipals` and those branches; catalog point lookups and listing now provide the sole principal representation.

internal/rootcoord/meta_table.go
aoiasd: Implemented a bounded non-paginated scan: List/ALL now stops at 65,536 principals or 64 MiB of logical principal-name + tag payload and returns ServiceQuotaExceeded. Exact (collectionID, principal) lookup remains point-based. Cursor pagination and a bounded Coord-side point-read cache are follow-up PRs; startup full loading will not be restored.

internal/metastore/model/collection.go
aoiasd: Removed Collection.RLSPrincipals and its clone, update, recovery, and test plumbing. Principal metadata now has a single representation in the catalog namespace and is accessed by point lookup or the explicit bounded list path.

internal/proxy/rls/manager.go
aoiasd: Fixed by tracking the last failure time on every backoff entry and expiring idle entries after the refresh interval. Exponential retry state is retained within that window, while success, invalidation, collection drop, and capacity eviction still clear it.

internal/rootcoord/meta_table.go
```diff
@@ -3415,44 +3254,84 @@ func (mt *MetaTable) GetRLSMetadata(ctx context.Context, collectionID int64, kin
...
+	for _, policy := range metadata.Policies {
+		policy.DBID = dbID
+		policy.CollectionID = collectionID
+	}
+	metadata.Principals = make([]*model.RLSPrincipal, len(principals))
+	for i, principal := range principals {
+		if principal == nil {
+			return nil, merr.WrapErrDataIntegrityMsg("RLS principal metadata is nil for collection %d", collectionID)
+		}
+		metadata.Principals[i] = model.CloneRLSPrincipal(principal)
```
buqian-zilliz: **[P2] Remove the redundant deep copy of catalog principals**

`GetRLSPrincipal` and `ListRLSPrincipals` return request-local models decoded from KV, yet this loop deep-copies every tag map before `Core.GetRLSMetadata` immediately serializes it. A point lookup duplicates up to 50 tags, while `RLS_METADATA_KIND_ALL` can duplicate the entire 65,536-entry/64 MiB logical result, with Go map overhead present in both copies at peak. Normalize `DBID` and `CollectionID` on the returned models and assign them directly, or document a shared-ownership contract if a catalog implementation actually returns shared state.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1406 @@
...
+
+	if len(permissiveExprs) == 0 {
+		if len(restrictiveExprs) > 0 {
+			return templates, "false"
+		}
+		return templates, ""
+	}
+
+	groups := make([]string, 0, 2)
+	groups = append(groups, joinExprs(permissiveExprs, "or"))
```
buqian-zilliz: **[P3] Track the combined length without constructing the expression**

`preparePolicyExprTemplates` allocates permissive and restrictive string slices and builds both group strings plus the final combined string, but its only caller immediately discards the contents and reads `len(combinedExpr)`. With up to 100 policies and 4 KiB expressions, a rejected compilation can allocate and copy hundreds of KiB solely to compute a quota value. Return the exact formatted length while iterating, accounting for parentheses and separators, and remove `permissiveExprs`, `restrictiveExprs`, `groups`, and `joinExprs`.

internal/rootcoord/meta_table.go
buqian-zilliz: This issue still exists. The bounded walker prevents one request from materializing unbounded memory, but principal creation remains unbounded and neither complete-list API has a cursor. Once stored data exceeds either reused cache limit, every full listing fails with `ServiceQuotaExceeded` and there is no way to enumerate the collection.

internal/rootcoord/meta_table.go
aoiasd: Fixed in 476a2df939. RootCoordCatalog now explicitly guarantees caller-owned Get/List principal results, MetaTable normalizes and returns them directly, and a KV regression mutates and rereads returned results to pin that ownership contract.

internal/proxy/rls/util.go
aoiasd: Fixed in 476a2df939. Proxy now accumulates the exact formatted byte length while preparing templates instead of constructing the combined expression. The permissive/restrictive grouping, restrictive-only false case, trimming, UTF-8 byte counting, and greater-than limit boundary are preserved.

internal/rootcoord/meta_table.go
aoiasd: Agreed. The current 65,536-principal / 64 MiB quota is a deliberate interim fail-fast boundary for the non-paginated API. Public/internal cursor pagination will be added in a follow-up PR; normal enforcement remains a point read, and a bounded Coord-side point-read cache is also planned separately.

internal/util/rlsutil/types.go
```diff
@@ -54,6 +54,22 @@ func NewDoubleTagValue(value float64) TagValue {
...
+// PrincipalTagsSize returns the logical bytes occupied by one principal and
+// its tags. Numeric values use their fixed-width in-memory representation.
+func PrincipalTagsSize(principalName string, tags map[string]TagValue) int64 {
+	size := int64(len(principalName))
+	for key, value := range tags {
+		size += int64(len(key))
+		switch value.Kind {
+		case TagValueKindString:
+			size += int64(len(value.StringValue))
+		case TagValueKindInt64, TagValueKindDouble:
```
zhengbuqian: add default branch, since we'll add more tag value kinds in the future, so we don't miss any code change. 

if you are a coding agent reading this, check other places for such issues and fix them as well

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1399 @@
...
+}
+
+type compiledPolicyExpression struct {
+	expr           *planpb.Expr
+	needsPrincipal bool
+	tagVariables   map[string]string
+}
+
+func (e *compiledExpression) needsTags() bool {
+	if e == nil || len(e.permissive) == 0 {
```
zhengbuqian: why don't we also check e.restrictive?

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1399 @@
...
+	restrictive []*compiledPolicyExpression
+}
+
+type compiledPolicyExpression struct {
+	expr           *planpb.Expr
+	needsPrincipal bool
+	tagVariables   map[string]string
+}
+
+func (e *compiledExpression) needsTags() bool {
```
zhengbuqian: precompute and store needsTags

congqixia (commented):
P0/P1/P2-graded review of the RLS evaluator (RLS3). The core combination logic (permissive OR / restrictive AND with always-true/false simplification), template compilation/caching, and the three-valued row evaluator are sound and very well covered by tests; I found **no P0 issues**. See inline comments for the P1/P2 items, plus two non-inline notes:

- **[P1] CI gate is red**: `ci-v2/build-ut-cov` is **FAILED** on this head (`GO failed: script returned exit code 1`, ~107 min) while `ci-v2/e2e-default`, `ci-v2/go-sdk`, and the main trigger pass. I could not access the Jenkins log (403 without auth), so I cannot tell whether this is a UT timeout or a real compile/UT failure — please confirm before merge.
- **[P2] `proxy.rls.maxPrincipalsPerCollection` was removed with no replacement creation-time cap**: principal growth in the catalog is now unbounded (only the proxy cache and the bulk-list limits apply, and bulk list hard-fails with `ServiceQuotaExceeded` beyond them). The in-code TODO covers pagination; a follow-up cap would also restore the ability to bound principal growth at the control plane.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1399 @@
...
+		if err != nil {
+			return truthUnknown, err
+		}
+		return truthValueFromBool(matched), nil
+	default:
+		return truthUnknown, merr.WrapErrServiceInternalMsg("unsupported RLS contains operator %s", expr.GetOp().String())
+	}
+}
+
+func valueEqual(rowValue any, target *planpb.GenericValue) (bool, error) {
```
congqixia: [P1] `valueEqual` dereferences `target` unconditionally (`target.GetVal()` at lines 1231/1238/1258/1264). The internal RLS path always fills values via `FillExpressionValue` (which errors when a template variable is missing), so this is only reachable through the exported `ValidateRowsByPredicate`/`ResolveUsingPredicate` APIs with a caller-supplied or half-instantiated `UnaryRangeExpr` whose `Value` is nil. A `nil` target would panic instead of returning an error. Suggest a `if target == nil { return false, nil }` guard.

internal/proxy/rls/util.go
```diff
@@ -0,0 +1,1399 @@
...
+				restrictiveLength += len(" and ")
+			}
+			restrictiveCount++
+			restrictiveLength += len(strings.TrimSpace(policyExpr)) + 2
+		}
+	}
+
+	if permissiveCount == 0 {
+		if restrictiveCount > 0 {
+			return templates, len("false")
```
congqixia: [P2] When `permissiveCount == 0` and `restrictiveCount > 0`, the combined length is hardcoded to `len("false")`, so restrictive-only policy sets never hit the `maxCombinedExpressionLength` quota. The restrictive templates are still parsed in `compileExprTemplates` and a parse failure there becomes a cached `DataIntegrity` error (predicate.go:196 caches DataIntegrity) -> permanent deny, even though `Instantiate` returns `alwaysFalse` without ever evaluating them. Consider skipping compilation of the dead restrictive group, or enforce the length limit on it.

internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -2439,21 +2437,38 @@ func (kc *Catalog) DropRLSPrincipal(ctx context.Context, collectionID int64, pri
...
+	}
+	maxEntries := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheEntries.GetAsInt()
+	maxBytes := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheBytes.GetAsInt64()
+	principals := make([]*model.RLSPrincipal, 0, min(maxEntries, rlsPrincipalScanBatchSize))
+	var totalBytes int64
+	// TODO: Add cursor pagination so callers can enumerate collections that
+	// exceed the non-paginated list limits.
+	err := walker.WalkWithPrefix(ctx, BuildRLSPrincipalPrefix(collectionID), rlsPrincipalScanBatchSize, func(_, value []byte) error {
+		if len(principals) >= maxEntries {
+			return merr.WrapErrServiceQuotaExceededMsg("RLS principal list exceeds the %d-entry limit", maxEntries)
```
congqixia: [P2] Bulk `ListRLSPrincipals` hard-fails with `ServiceQuotaExceeded` for any collection exceeding the entry/byte limits instead of returning a partial list or paginating. This also means `GetRLSMetadata(kind=ALL/PRINCIPALS)` fails wholesale for large collections. The `// TODO: Add cursor pagination` below acknowledges this; worth tracking as an explicit follow-up since the removed `maxPrincipalsPerCollection` quota used to bound this.

internal/proxy/rls/snapshot.go
```diff
@@ -33,6 +36,18 @@ type CoordClient interface {
 	GetRLSMetadata(ctx context.Context, in *rootcoordpb.GetRLSMetadataRequest, opts ...grpc.CallOption) (*rootcoordpb.GetRLSMetadataResponse, error)
 }
 
+func wrapMetadataRefreshError(err error, format string, args ...any) error {
+	if merr.IsMilvusError(err) {
+		return merr.Wrapf(err, format, args...)
+	}
+	code := status.Code(err)
+	if errors.IsAny(err, context.Canceled, context.DeadlineExceeded) ||
```
congqixia: [P2] Caller/cancellation contexts are classified as retriable service unavailability (`context.Canceled`/`codes.Canceled` -> `WrapErrServiceUnavailableErr`), which then records an exponential backoff for the principal via `recordPrincipalRefreshFailureLocked`. This is harmless (fail-closed, backoffs expire by TTL) but a client cancel arguably should not poison backoff state. Consider returning the cancel as-is instead of mapping it to ServiceUnavailable.

pkg/util/paramtable/component_param.go
```diff
@@ -3361,6 +3351,28 @@ Disabled if the value is less or equal to 0.`,
 	}
 	p.RLSMaxArrayLiteralElements.Init(base.mgr)
 
+	p.RLSMaxPrincipalCacheEntries = ParamItem{
```
congqixia: [P2] `proxy.rls.maxPrincipalsPerCollection` is removed here and in configs/milvus.yaml with no replacement creation-time cap; principal count in the catalog is now unbounded. If the intent is to let the new cache/list limits bound growth, a brief design note or follow-up issue (e.g., a per-collection creation quota or cursor pagination) would help operators. See the review-level note.

internal/metastore/kv/rootcoord/kv_catalog.go
aoiasd: Agreed. The current non-paginated API intentionally fails once the bounded materialization limits are exceeded. Cursor pagination is recorded as an explicit follow-up and is outside this PR.

internal/proxy/rls/util.go
aoiasd: `GenericValue.GetVal()` is a generated nil-safe getter: it checks the receiver and returns nil when `target` is nil (`plan.pb.go:841-845`). These type assertions therefore fail normally; this path does not panic, so no guard is needed.

internal/proxy/rls/util.go
aoiasd: The loop below checks both permissive and restrictive groups. The early return only covers a restrictive-only set: without any permissive policy the effective predicate is always false, so restrictive tags cannot change the authorization result. The precomputed `needsTags` value keeps this behavior and adds an explicit comment.

internal/proxy/rls/util.go
aoiasd: `maxCombinedExpressionLength` applies to the effective combined predicate. With no permissive policy that predicate is exactly `false`, so the length of `false` is intentional. We still compile the persisted restrictive policies to detect schema/data-integrity corruption and fail closed.

pkg/util/paramtable/component_param.go
aoiasd: The creation-time cap was intentionally removed for the million-principal target. Proxy caching and current non-paginated list materialization remain bounded; cursor pagination and a bounded RootCoord point-lookup cache are tracked as follow-ups rather than restoring a principal-count cap.

internal/proxy/rls/snapshot.go
aoiasd: The caller context is not passed to the shared refresh RPC; the RPC uses the manager-owned `refreshCtx`. Caller cancellation only exits the waiting select and does not finish the refresh as failed or record backoff. A cancellation reaching this wrapper comes from the shared RPC/manager lifecycle and remains a transient service failure.

internal/util/rlsutil/types.go
aoiasd: Fixed. Added a default branch that returns an internal error, and audited the sibling TagValueKind switches so unknown future kinds also fail closed.

internal/proxy/rls/util.go
aoiasd: Fixed. `needsTags` is now precomputed and stored in the compiled policy expression, so request-time resolution no longer rescans the compiled tree.

internal/proxy/rls/manager.go
```diff
@@ -414,6 +465,122 @@ func principalRefreshKey(key principalKey, token principalRefreshToken) string {
...
+		order.Remove(entry.orderElement)
+		entry.orderElement = nil
+	}
+}
+
+func (state *collectionState) trimPrincipalStateLocked() {
+	maxEntries := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheEntries.GetAsInt()
+	maxBytes := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheBytes.GetAsInt64()
+	for len(state.principalTags)+len(state.principalBackoffs) > maxEntries || state.principalCacheBytes+state.principalBackoffBytes > maxBytes {
+		if oldest := state.principalBackoffOrder.Front(); oldest != nil {
```
buqian-zilliz: **[P2] Preserve refresh backoff when the principal cache is full**

`recordPrincipalRefreshFailureLocked` inserts a backoff and immediately calls this trimmer. When `principalTags` already reaches either per-collection limit, the loop removes entries from `principalBackoffOrder` before evicting any tag entry, so the newly created backoff is discarded. During RootCoord or metadata-store failures, repeated requests for the same uncached principal can therefore continuously start refresh RPCs as the 64 slots turn over instead of observing the intended 1-30 second retry suppression. Reserve a separately bounded backoff budget or evict an appropriate cache entry before dropping an active backoff.

aoiasd: Fixed in cb17ef9535. The trimmer now protects the backoff being recorded, evicts older backoffs or cached tag entries first, and drops the protected entry only if it alone exceeds the hard bound. TestManagerPrincipalRefreshFailureKeepsNewestBackoffWhenCacheIsFull verifies the full-cache case and confirms an immediate retry does not start another metadata RPC.

internal/proxy/rls/manager.go
```diff
@@ -414,6 +465,129 @@ func principalRefreshKey(key principalKey, token principalRefreshToken) string {
...
+
+func (state *collectionState) trimPrincipalStateLocked(protectedBackoff string) {
+	maxEntries := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheEntries.GetAsInt()
+	maxBytes := paramtable.Get().ProxyCfg.RLSMaxPrincipalCacheBytes.GetAsInt64()
+	for len(state.principalTags)+len(state.principalBackoffs) > maxEntries || state.principalCacheBytes+state.principalBackoffBytes > maxBytes {
+		if oldest := state.principalBackoffOrder.Front(); oldest != nil && oldest.Value.(string) != protectedBackoff {
+			state.removePrincipalBackoffLocked(oldest.Value.(string))
+			continue
+		}
+		oldest := state.negativePrincipalCacheOrder.Front()
```
buqian-zilliz: **[P2] Avoid flushing the cache for an oversized protected backoff**

`recordPrincipalRefreshFailureLocked` can create a backoff whose `cacheBytes` exceeds a dynamically lowered `maxPrincipalCacheBytes`, because existing principal names remain addressable up to the 64 KiB transport bound. This branch evicts negative and positive cache entries before the later fallback drops the oversized protected backoff, so one failed metadata lookup can flush every fresh principal entry while retaining no retry suppression; repeated requests then repeat the RPC and eviction cycle. Check whether the protected backoff can fit before trimming and discard it directly without evicting unrelated cached principals.

internal/proxy/rls/manager.go
buqian-zilliz: This issue still exists: `protectedBackoff` only protects the entry currently being recorded. At a full cache, a failure for a second principal removes the older, still-active backoff before evicting any cached tag entry. Alternating requests for distinct uncached principals during a metadata outage can therefore discard each retry window and continuously restart refresh RPCs as the 64 slots turn over. Evict tag-cache entries before any unexpired backoff, or give backoffs a separate bounded budget.

internal/proxy/rls/manager.go
aoiasd: Confirmed. This affects cache retention and can increase metadata RPCs when the configured byte limit is lower than one principal backoff, but it does not affect authorization correctness or violate the configured hard bound. We will track the trim-order optimization as a non-blocking follow-up and keep RLS3 unchanged for merge.

PR #52075, opened 2026-07-31
feat: [RLS4] expose and enforce row-level security

aoiasd:
relate: #50263
design doc PR: #53173

## Summary
Enforces RLS across query, search, delete, insert, and upsert, including privileged `skip_rls`.


internal/proxy/task_delete.go
```diff
@@ -334,17 +335,40 @@ func (dr *deleteRunner) Init(ctx context.Context) error {
...
+		if err := checkSkipRLSPrivilege(ctx, dr.req.GetDbName(), dr.req.GetCollectionName(), "delete"); err != nil {
+			return err
+		}
+		rlsEnabled = false
+	}
+	principalName, enforceRLS, err := rls.ResolveRuntimePrincipal(rlsEnabled, dr.req.GetRlsPrincipal(), "delete")
+	if err != nil {
+		return err
+	}
+	if err := rls.DefaultManager().ApplyRLSUsingPredicate(ctx, dr.collectionID, principalName, milvuspb.RowPolicyAction_RowPolicyActionDelete, enforceRLS, dr.schema.schemaHelper, visitorArgs, dr.plan); err != nil {
```
buqian-zilliz: **[P1] Keep RLS delete authorization and tombstones on one snapshot**

The RLS predicate only constrains the preliminary QueryStream at `dr.ts`, but every matched PK batch is then enqueued as a new `deleteTask`, whose queue allocation assigns a later timestamp. Because the emitted tombstone contains only the PK and that later timestamp, a concurrent upsert can move an authorized row into a tenant the caller may not delete after the RLS read, and the later tombstone will still remove that new version. Bind deletion to the authorized row version/snapshot or re-evaluate the predicate at the mutation snapshot.

internal/proxy/task_insert.go
```diff
@@ -291,6 +296,22 @@ func (it *insertTask) PreExecute(ctx context.Context) error {
...
+		if err := checkSkipRLSPrivilege(ctx, it.insertMsg.GetDbName(), it.insertMsg.GetCollectionName(), "insert"); err != nil {
+			return err
+		}
+		it.rlsEnabled = false
+	}
+	principalName, enforceRLS, err := rls.ResolveRuntimePrincipal(it.rlsEnabled, it.rlsPrincipal, "insert")
+	if err != nil {
+		return err
+	}
+	if err := rls.ValidateCheckForWrite(ctx, it.collectionID, principalName,
```
buqian-zilliz, line 333: **[P1] Handle empty ValidData before evaluating write policies**

`ValidateCheckForWrite` panics when a `check_expr` references a non-nullable scalar field with a default value. `validateUtil.FillWithDefaultValue` leaves such fields with a non-nil empty `ValidData`; `typeutil.GetDataIterator` interprets any non-nil slice as nullable, builds a zero-length index table, and indexes it for row 0. The same evaluator is used by insert and upsert, so valid writes can fail deterministically. Handle `len(ValidData) == 0` as dense data, or preserve nil, before local RLS evaluation.

pkg/common/common.go
```diff
@@ -683,16 +683,9 @@ func ValidateRLSProperties(kvs ...*commonpb.KeyValuePair) error {
 	for _, kv := range kvs {
 		switch kv.GetKey() {
 		case RLSEnabledKey:
-			enabled, err := IsRLSEnabled(kv)
-			if err != nil {
+			if _, err := IsRLSEnabled(kv); err != nil {
```
buqian-zilliz, line 755: **[P2] Reject duplicate rls.enabled properties**

Reject duplicate `rls.enabled` entries. Collection properties are repeated KVs; creation preserves their order, `IsRLSEnabled` and the proxy cache take the first match, while alter processing normalizes through `ToMap` and takes the last. A collection created with `true` followed by `false` therefore starts with RLS enabled, but a later unrelated property update can canonicalize the value to false and silently disable enforcement. Reject a second occurrence or normalize once with explicit precedence.

internal/proxy/task_upsert.go
```diff
@@ -1669,16 +1707,25 @@ func (it *upsertTask) PreExecute(ctx context.Context) error {
...
-	if it.req.GetPartialUpdate() {
-		err = it.queryPreExecute(ctx)
+	if it.rlsEnabled && it.req.GetSkipRls() {
+		if err := checkSkipRLSPrivilege(ctx, it.req.GetDbName(), collectionName, "upsert"); err != nil {
+			return err
+		}
+		it.rlsEnabled = false
+	}
+
+	if it.req.GetPartialUpdate() || it.rlsEnabled {
```
buqian-zilliz, line 1828: **[P2] Avoid fetching full rows for RLS-only upsert checks**

This branch sends every RLS-enabled full upsert through `queryPreExecute`, where `retrieveByPKs` performs a strong query requesting `*` for all output fields. The non-partial path uses the result only for the `using_expr` check and then discards it, so unrelated vectors and text are transferred and materialized and can push otherwise valid batches over `maxOutputSize`. Request only the PK and fields referenced by the compiled predicate for non-partial upserts, retaining full rows only for partial-merge reconstruction.

internal/proxy/task_search.go
```diff
@@ -1174,15 +1177,35 @@ func (t *searchTask) tryGeneratePlan(params []*commonpb.KeyValuePair, dsl string
...
+		if err := checkSkipRLSPrivilege(t.ctx, t.request.GetDbName(), t.request.GetCollectionName(), operation); err != nil {
+			return nil, nil, 0, false, nil, internalpb.SearchType_DEFAULT, err
+		}
+		t.rlsEnabled = false
+	}
+	principalName, enforceRLS, err := rls.ResolveRuntimePrincipal(t.rlsEnabled, t.request.GetRlsPrincipal(), operation)
+	if err != nil {
+		return nil, nil, 0, false, nil, internalpb.SearchType_DEFAULT, err
+	}
+	rlsPredicate, err := rls.DefaultManager().GetRLSUsingPredicate(t.ctx, t.GetCollectionID(), principalName, rls.SearchAction(t.IsAdvanced, searchInfo.isIterator), enforceRLS, t.schema.schemaHelper, visitorArgs)
```
buqian-zilliz: **[P2] Resolve hybrid-search RLS once per request**

`initAdvancedSearchRequest` calls this path once per subrequest, so the same collection, principal, and HybridSearch predicate can be resolved up to 1024 times. Even on a compiled-cache hit, each call repeats freshness and locking work and clones, fills, rewrites, and recombines every applicable policy expression. Resolve one predicate for the top-level hybrid request and merge a clone into each subplan; this also keeps all subplans on one authorization snapshot.

internal/proxy/task.go
```diff
@@ -2184,6 +2190,16 @@ func (t *alterCollectionTask) PreExecute(ctx context.Context) error {
 	if err := common.ValidateNamespaceShardingEnabledNotAltered(t.GetProperties(), t.GetDeleteKeys()); err != nil {
 		return err
 	}
+	if err := common.ValidateRLSProperties(t.GetProperties()...); err != nil {
```
zhengbuqian, line 2137: Require ManageRLS before mutating RLS properties. A caller permitted to alter the collection but without ManageRLS can set rls.enabled=false or rls.force=false (or delete these keys), thereby disabling or weakening RLS. Please perform a collection-scoped ManageRLS check whenever either RLS key appears in Properties or DeleteKeys.

aoiasd:
@Mergifyio refresh

pkg/common/common.go
```diff
@@ -884,30 +884,17 @@ func ValidateRLSProperties(kvs ...*commonpb.KeyValuePair) error {
 				return merr.WrapErrParameterInvalidMsg("duplicated collection property %q", RLSEnabledKey)
 			}
 			seen[RLSEnabledKey] = struct{}{}
-			enabled, err := IsRLSEnabled(kv)
-			if err != nil {
+			if _, err := IsRLSEnabled(kv); err != nil {
```
buqian-zilliz: **[P2] Complete `rls.enabled` property transitions**

This line now permits `rls.enabled=true`, but `alterCollectionTask.PreExecute` still calls `ValidateRLSEnabledNotAltered`, so setting or deleting the switch through `AlterCollection` is rejected. Existing collections therefore cannot opt in, and enabled collections cannot execute the disable transition that must remove their policies and tag bindings and invalidate serving state. Implement the enable/disable transition atomically, including metadata readiness or cleanup and Proxy invalidation, or keep the runtime switch gated until that path exists.

internal/proxy/impl.go
```diff
@@ -6835,6 +6862,298 @@ func (node *Proxy) OperatePrivilegeGroup(ctx context.Context, req *milvuspb.Oper
...
+		return &milvuspb.ListRowPoliciesResponse{
+			Status:         merr.Status(err),
+			DbName:         req.GetDbName(),
+			CollectionName: req.GetCollectionName(),
+		}, nil
+	}
+	return resp, nil
+}
+
+func (node *Proxy) SetRLSPrincipalTags(ctx context.Context, req *milvuspb.SetRLSPrincipalTagsRequest) (*commonpb.Status, error) {
```
buqian-zilliz: **[P2] Make principal-tag patch semantics explicit**

`SetRLSPrincipalTags` forwards the JSON object to a metadata path that merges supplied keys into the existing map; omitted keys remain active, although the public method name and request comment do not state patch semantics. A caller rotating an authorization tag set can therefore believe an omitted grant was revoked while it remains usable. Expose distinct patch and replace operations, or rename and document this as an incremental upsert together with the required `DeleteRLSPrincipalTags` step for removals.

internal/proxy/task_upsert.go
```diff
@@ -302,13 +310,16 @@ func (it *upsertTask) prepareUpsert(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		if !primaryField.GetAutoID() {
+		fullAutoID = primaryField.GetAutoID()
+		if !fullAutoID && !it.rlsEnabled {
```
buqian-zilliz: **[P2] Skip the strong read for permissive full upserts**

By making only non-RLS full upserts return here, every RLS-enabled non-partial, non-AutoID upsert falls through to a synchronous Strong retrieve. When `ResolveUsingPredicate` simplifies the effective `using_expr` to true, it returns nil, so the retrieved rows are not used for authorization or reconstruction; the request still incurs a distributed query, PK materialization, and shard I/O. Preserve the direct delete-and-insert path when the resolved predicate is nil and there is no resolution error; keep the read for partial updates, AutoID classification, or a nontrivial predicate.

pkg/proto/root_coord.proto
```diff
@@ -180,6 +180,16 @@ service RootCoord {
...
     rpc OperatePrivilegeGroup(milvus.OperatePrivilegeGroupRequest) returns (common.Status) {}
 
+    // Row-Level Security (RLS) management APIs.
+    rpc CreateRowPolicy(milvus.CreateRowPolicyRequest) returns (common.Status) {}
+    rpc DropRowPolicy(milvus.DropRowPolicyRequest) returns (common.Status) {}
+    rpc ListRowPolicies(milvus.ListRowPoliciesRequest) returns (milvus.ListRowPoliciesResponse) {}
+    rpc UpdateRowPolicy(milvus.UpdateRowPolicyRequest) returns (common.Status) {}
+    rpc SetRLSPrincipalTags(milvus.SetRLSPrincipalTagsRequest) returns (common.Status) {}
+    rpc GetRLSPrincipalTags(milvus.GetRLSPrincipalTagsRequest) returns (milvus.GetRLSPrincipalTagsResponse) {}
+    rpc ListRLSPrincipals(milvus.ListRLSPrincipalsRequest) returns (milvus.ListRLSPrincipalsResponse) {}
```
buqian-zilliz: **[P2] Paginate principal enumeration before exposing it**

`ListRLSPrincipalsRequest` has no page size or continuation token, so this newly exposed RPC must enumerate the collection in one operation. The current backing scan can materialize and decode up to 65,536 principals or 64 MiB of tag metadata, sort full slices, and then copy all names into one response; lowering the refreshable limits can also make an existing collection impossible to enumerate. Add cursor-based pagination backed by a name-only, page-bounded catalog scan.

internal/proxy/task_delete.go
buqian-zilliz: This issue still exists: complex delete still authorizes at `dr.ts`, then enqueues each PK batch as a new `deleteTask` whose scheduler-assigned timestamp is later and whose tombstone carries no checked row version.

internal/proxy/task_insert.go
aoiasd: Fixed in the current head: Proxy normalizes empty ValidData before local write-policy evaluation, so dense/default-backed fields are no longer interpreted as a zero-length nullable index.

internal/proxy/task_upsert.go
aoiasd: Fixed. Non-partial upserts now request only the primary key plus fields referenced by the instantiated USING predicate; full-row retrieval remains limited to partial-update reconstruction.

internal/proxy/task_search.go
aoiasd: Fixed. The top-level search task resolves and stores the HybridSearch RLS predicate once, then clones and merges it into each sub-plan.

internal/proxy/impl.go
aoiasd: The selected API semantics are now explicit in the design document: Set incrementally upserts supplied keys and preserves unspecified tags; Delete removes selected keys, and an empty key list removes all bindings. A combined upsert+delete atomic patch remains a follow-up.

pkg/proto/root_coord.proto
aoiasd: Accepted follow-up. The current unary API uses bounded backend scans and fails with ServiceQuotaExceeded before materializing more than 65,536 principals or 64 MiB. Cursor pagination and a bounded Coord-side point cache will be added separately.

internal/proxy/task_upsert.go
aoiasd: Fixed. For non-partial, non-AutoID full upserts, a successfully resolved nil/always-true USING predicate preserves the direct delete-and-insert path and skips the strong read.

internal/proxy/task_delete.go
aoiasd: This concurrency is accepted by the current mutation semantics. The delete is authorized against its read snapshot; once authorized, a concurrent upsert that changes the row's later RLS visibility does not revoke the already-decided delete. Providing row-version-bound tombstones would require a broader storage/WAL contract and is outside this RLS iteration.

pkg/common/common.go
aoiasd: Fixed. ValidateRLSProperties now rejects duplicate rls.enabled and rls.force keys, and the property reader uses one consistent last-value rule.

internal/proxy/task.go
aoiasd: Fixed. AlterCollection now requires collection-scoped ManageRLS whenever rls.enabled or rls.force appears in Properties or DeleteKeys.

pkg/common/common.go
aoiasd: The runtime transition remains intentionally gated in this iteration: ValidateRLSEnabledNotAltered rejects setting or deleting rls.enabled after creation. A future transition implementation will perform enable initialization or disable cleanup and invalidation atomically.

Issue #51245, opened 2026-07-10
[Enhancement]: Add async storage v2 load path for field data and scalar indexes


sparknack:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Add an async load path for storage v2/v3 data used by QueryNode sealed segment load. The work can land in several PRs:

- Async field-data loading for storage v3 manifest/column-group load, guarded by a temporary rollout switch.
- Async scalar index loading for ScalarIndexV3, using the same scheduling/admission model where possible.
- Follow-up rollout work to enable the async path by default after validation.
- Cleanup work to remove the temporary switch and obsolete synchronous concepts once the async implementation becomes the only path.

### Why is this needed?

The current sealed segment load path materializes storage v3 data through mostly synchronous translator work. For wide schemas or many row groups, this limits IO parallelism and makes load latency sensitive to object-storage round trips.

An async load pipeline lets field-data reads, materialization, mmap writes, and later scalar index reads be scheduled with explicit admission control and priority. This should improve load throughput while keeping memory usage bounded and preserving the existing synchronous path during rollout.

### Anything else?

The first PR should focus on field data only and keep index loading unchanged. It should add a temporary disabled-by-default switch so deployments can opt in to the async path safely before the old path is removed.

PR #52077, opened 2026-07-31
enhance: add mmap field data writeback

sparknack:
issue: #51245

## Summary

- Add a disabled-by-default `queryNode.mmap.writeback` option to sync mmap field-data files written through buffered I/O before mapping them.
- Skip the `fdatasync` barrier when `FileWriter` resolves to direct I/O, while keeping actual buffered writes eligible for writeback.
- Propagate a typed `MmapChunkWritebackMode` through sealed-segment field-data translators and mmap chunk creation.

This PR only adds the mmap writeback primitive used by the async field-data load work tracked in #51245; it does not enable the async load path.

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -207,6 +207,13 @@ FormatRuntimeFieldIds(
     return fmt::format("{}", ids);
 }
 
+static MmapChunkWritebackMode
+CreateMmapChunkWritebackMode(const storage::MmapConfig& mmap_config) {
+    return mmap_config.GetMmapWriteback()
+               ? MmapChunkWritebackMode::FdatasyncOnFinish
+               : MmapChunkWritebackMode::Disabled;
+}
+
 static void
 CheckVectorOutputCellsLoaded(int64_t segment_id,
                              FieldId field_id,
@@ -2443,6 +2450,7 @@ ChunkedSegmentSealedImpl::load_column_group_data_internal(
     size_t num_rows = storage::GetNumRowsForLoadInfo(load_info);
     ArrowSchemaPtr arrow_schema = schema_snapshot->ConvertToArrowSchema();
     auto& mmap_config = storage::MmapManager::GetInstance().GetMmapConfig();
+    auto writeback_mode = CreateMmapChunkWritebackMode(mmap_config);
 
     for (auto& [id, info] : load_info.field_infos) {
         AssertInfo(info.row_count > 0,
@@ -2544,7 +2552,8 @@ ChunkedSegmentSealedImpl::load_column_group_data_internal(
                 mmap_config.GetMmapPopulate(),
                 milvus_field_ids.size(),
                 load_info.load_priority,
-                warmup_policy);
+                warmup_policy,
+                writeback_mode);
         auto chunked_column_group =
             std::make_shared<ChunkedColumnGroup>(std::move(translator));
 
@@ -2604,6 +2613,7 @@ ChunkedSegmentSealedImpl::load_column_group_data_internal(
     size_t num_rows = storage::GetNumRowsForLoadInfo(load_info);
     ArrowSchemaPtr arrow_schema = schema_snapshot->ConvertToArrowSchema();
     auto& mmap_config = storage::MmapManager::GetInstance().GetMmapConfig();
+    auto writeback_mode = CreateMmapChunkWritebackMode(mmap_config);
 
     for (auto& [id, info] : load_info.field_infos) {
         AssertInfo(info.row_count > 0,
@@ -2697,7 +2707,8 @@ ChunkedSegmentSealedImpl::load_column_group_data_internal(
                 mmap_config.GetMmapPopulate(),
                 milvus_field_ids.size(),
                 load_info.load_priority,
-                warmup_policy);
+                warmup_policy,
+                writeback_mode);
         auto chunked_column_group =
             std::make_shared<ChunkedColumnGroup>(std::move(translator));
 
@@ -2798,6 +2809,7 @@ ChunkedSegmentSealedImpl::load_field_data_internal(
     SCOPE_CGO_CALL_METRIC();
 
     auto& mmap_config = storage::MmapManager::GetInstance().GetMmapConfig();
+    auto writeback_mode = CreateMmapChunkWritebackMode(mmap_config);
 
     size_t num_rows = storage::GetNumRowsForLoadInfo(load_info);
     auto visible_runtime_owner =
@@ -2888,7 +2900,8 @@ ChunkedSegmentSealedImpl::load_field_data_internal(
                         info.enable_mmap,
                         mmap_config.GetMmapPopulate(),
                         load_info.load_priority,
-                        warmup_policy);
+                        warmup_policy,
+                        writeback_mode);
 
             auto data_type = field_meta.get_data_type();
             auto slot = cachinglayer::Manager::GetInstance().CreateCacheSlot(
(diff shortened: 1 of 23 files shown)


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -2659,7 +2667,8 @@ ChunkedSegmentSealedImpl::load_column_group_data_internal(
                 mmap_config.GetMmapPopulate(),
                 milvus_field_ids.size(),
                 load_info.load_priority,
-                warmup_policy);
+                warmup_policy,
+                writeback_mode);
```
buqian-zilliz, line 2556: **[P1] Propagate writeback through staged sealed-load paths**

`writeback_mode` is only passed by the non-staged overload shown here. The matching `StagedStateCommitter` overloads still construct `GroupChunkTranslator`, `ChunkTranslator`, and `DefaultValueChunkTranslator` without the new argument, so each silently uses `Disabled`. `Load()` and `Reopen()` route storage-v1/storage-v2 binlog loads and default-field creation through those staged overloads; therefore `queryNode.mmap.writeback=true` does not request `fdatasync` on these normal paths. Derive and pass the mode in all staged overloads as well, and consider removing the defaults so future omissions fail to compile.

internal/core/src/storage/FileWriter.cpp
```diff
@@ -211,6 +216,36 @@ FileWriter::PositionedWriteWithCheck(const void* data,
...
+void
+FileWriter::SyncFileData() {
+#ifdef __APPLE__
+    auto ret = fsync(fd_);
+#else
+    auto ret = fdatasync(fd_);
+#endif
+    if (ret != 0) {
+        auto saved_errno = errno;
+        Cleanup();
```
buqian-zilliz, line 232: **[P2] Clean up the mmap file after a sync failure**

`fdatasync` can report delayed `ENOSPC` or `EIO` after buffered writes have already created the file. This branch closes the descriptor and throws, but `MmapChunkTarget::release()` has not yet constructed the `ChunkMmapGuard` that owns and unlinks the path, so the file is orphaned. The storage-v2 translators generate a new suffix for each attempt, so retries can accumulate full-size files and make disk-pressure recovery progressively worse. Keep the path under an RAII cleanup guard until ownership is transferred to `ChunkMmapGuard`, or explicitly unlink it when release fails.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
sparknack: Fixed. The staged storage-v2 binlog, storage-v1 binlog, and default-field paths now derive and pass the configured writeback mode explicitly. The staged manifest path was already passing it. I also audited all eight production translator construction sites; each now forwards writeback_mode.

internal/core/src/storage/FileWriter.cpp
sparknack: Fixed. MmapChunkTarget now retains RAII ownership of both the mapping and file path until ChunkMmapGuard is constructed. Any flush/fdatasync/mmap/guard-construction failure unwinds through the target destructor and removes the temporary file. Added MmapChunkTargetCleansUpFileWhenReleaseFails to cover the failure path.

internal/core/src/storage/FileWriter.cpp
```diff
@@ -379,6 +414,8 @@ FileWriter::Finish() {
         }
     }
 
+    SyncWrittenData();
```
buqian-zilliz: **[P2] Keep fdatasync under the disk-write concurrency limit**

`SyncWrittenData()` runs on the caller only after the final flush task has released its `FileWriteWorkerPool` slot. `common.diskWriteNumThreads` is documented to cap concurrent disk-write operations to protect read latency, but with writeback enabled each load caller can now block in `fdatasync` while the pool starts another write or flush task, so even a limit of 1 permits many concurrent syncs. Since `fdatasync` is the operation that forces dirty pages to storage, parallel field loading can create a sync storm. Run the final sync under the same pool or admission mechanism, including the no-tail-flush case, or add equivalent concurrency control.

sparknack: Fixed. `FileWriter::Finish()` now runs the final tail flush and `fdatasync` in one `FileWriteWorkerPool` task. The sync-only path with no tail flush is also submitted to the pool, so `common.diskWriteNumThreads` covers both cases. Added `FinishKeepsFdatasyncInWriterPool` to block the sync point and verify a second writer cannot enter a single-thread pool concurrently. Verified with all 49 `FileWriterTest` cases and `make verifiers`.

internal/core/src/storage/FileWriter.cpp
```diff
@@ -342,17 +381,26 @@ FileWriter::FlushWithBufferedIO() {
...
+            // Keep the final flush and sync in the same worker-pool task so
+            // diskWriteNumThreads limits both operations.
+            if (offset_ != 0) {
                 if (use_direct_io_) {
                     FlushWithDirectIO();
                 } else {
                     FlushWithBufferedIO();
                 }
+            }
+            SyncWrittenData();
```
buqian-zilliz: **[P2] Consume the pooled finish future's exception**

`SyncWrittenData()` now runs inside a worker task that catches failures and stores them with `promise->setException()`, but the caller only uses `future.wait()`. In the Folly version used here, `wait()` merely blocks until the future is ready and does not rethrow the stored exception; that requires consuming the result with `get()` or `value()`. With `common.diskWriteNumThreads > 0`, an `fdatasync`/`fsync` failure is therefore ignored, `Finish()` returns success, and `MmapChunkTarget` maps the file even though the requested writeback barrier failed. Consume the future result before cleanup so the exception reaches the existing RAII unlink path.

sparknack: Fixed. The pooled finish path now consumes the result with `std::move(future).get()`, so exceptions stored by `promise->setException()` are rethrown into the existing cleanup/error path. Added `FinishPropagatesPooledSyncFailure`, which injects a sync failure under a single-thread `FileWriteWorkerPool`; it failed with `wait()` and passes with `get()`. Verified all 50 `FileWriterTest` cases and `make verifiers`.

internal/core/src/storage/FileWriter.cpp
```diff
@@ -362,20 +400,16 @@ FileWriter::Finish() {
 
         if (FileWriteWorkerPool::GetInstance().AddTask(task)) {
             try {
-                future.wait();
+                std::move(future).get();
```
buqian-zilliz, line 403: **[P1] Propagate failures from pooled writes too**

`get()` now propagates errors from the final flush/sync task, but earlier pooled writes still use `future.wait()` in `FileWriter::Write()`. Folly's `wait()` only blocks and does not rethrow the exception stored by `promise->setException()`. If `WriteInternal()` hits `ENOSPC`/`EIO`, it calls `Cleanup()`—closing the fd and freeing `aligned_buf_`—yet `Write()` returns success; `MmapChunkTarget` then advances `size_`, so the next padding write can dereference the cleared buffer or `release()` can mmap a short file and later trigger `SIGBUS`. Consume the pooled write future with `get()` as well so every data-write failure reaches the new RAII cleanup path.

sparknack: Fixed. The pooled `Write()` path now consumes its future with `std::move(future).get()`, so `WriteInternal()` failures propagate through the existing cleanup/error path. Added `WritePropagatesPooledFailure`, which invalidates the writer fd and forces a pooled write; it failed with `wait()` because `Write()` threw nothing, and passes with `get()`. I also audited both `setException()` paths in `FileWriter`; `Write()` and `Finish()` now both consume their futures with `get()`. Verified all 51 `FileWriterTest` cases and `make verifiers`.

zhengbuqian (approved):

Issue #52083, opened 2026-07-31
[Enhancement]: Support AutoIndex for the bound index of add_function_field


MrPresent-Han:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Allow omitting index params (or passing `AUTOINDEX`) when adding a function output field via `add_function_field` / `AlterCollectionSchema` ADD. The bound index of the new vector output field should resolve to concrete build params from the AutoIndex config on the server side, the same way `create_index` supports AUTOINDEX, instead of requiring an explicit `index_type`.

Scope:
- Resolve via AutoIndex only when `index_type` is absent or `AUTOINDEX` (with `AUTOINDEX`, only `metric_type` may additionally be passed, mirroring the create_index rule); explicit `index_type` behavior stays unchanged.
- Persist the auto-resolved bound index in the create_index autoindex convention (`UserIndexParams = AUTOINDEX + final metric`) so a later `create_index(AUTOINDEX)` on the same field is deduplicated as idempotent.
- For BM25 functions, a non-user-specified metric is forced to BM25 (the function type authoritatively determines it) instead of failing on the config default metric.

### Why is this needed?

#51105 bound an index to the function output field at DDL time but made an explicit `index_type` mandatory. A user adding e.g. a BM25 sparse field often does not care about index details and expects the AutoIndex experience that `create_index` already provides.

### Anything else?

Server-side only for now; making `index_params` optional in pymilvus `add_function_field` is a follow-up.

PR #52084, opened 2026-07-31
enhance: support AutoIndex for add_function_field bound index (#52083)

MrPresent-Han:
issue: #52083

### What

Allow omitting index params (or passing `AUTOINDEX`) when adding a function output field via `add_function_field` (`AlterCollectionSchema` ADD). The bound index of the new vector output field now resolves to concrete build params from the AutoIndex config, mirroring the `create_index` AUTOINDEX rules. Previously an explicit `index_type` was mandatory (#51105).

### How

- `indexparamcheck.PrepareFunctionOutputIndexParams` gains an AutoIndex resolution branch: when `index_type` is absent or `AUTOINDEX`, per-data-type AutoIndex config supplies the build params (dense incl. large-topk/refine handling, sparse, binary incl. deduplicate-metric selection, int8). With `AUTOINDEX`, only `metric_type` may additionally be passed ("only metric type can be passed when use AutoIndex", same as create_index); a user-specified metric wins over the config one. Explicit `index_type` behavior is byte-for-byte unchanged.
- Resolution happens pre-broadcast in rootcoord `prepareBoundFieldIndex` (single source of truth; the fully resolved FieldIndex is carried in the AlterCollectionMessage, so ack-callback replay stays a pure idempotent apply). Proxy `preExecuteAdd` validates with the same shared helper.
- The auto-resolved bound index persists in the create_index autoindex convention: `UserIndexParams = [index_type: AUTOINDEX, metric_type: <final>]` (`WrapUserIndexParams`). Datacoord `checkParams` dedupes by TypeParams+UserIndexParams, so a later `create_index(AUTOINDEX)` on the field returns idempotent success instead of "at most one distinct index per field", and the autoindex config-upgrade metric compatibility logic applies. No datacoord changes needed.
- One deliberate divergence from create_index: for BM25 functions a **non-user-specified** metric is forced to BM25 (the DDL knows the function type authoritatively; the sparse config default IP is not a user choice). A user-specified wrong metric is still rejected.
- `wrapUserIndexParams` / `getDenseFloatAutoIndexParams` / `adjustAutoIndexParamsByDataType` are moved (not copied) from proxy into `indexparamcheck` so proxy and rootcoord share one implementation.
- The empty-params gate in `schemautil.ValidateAlterSchemaAddFunctionPlan` is lifted (a vector output field still ALWAYS gets a bound index — now materialized via AutoIndex when params are omitted, so the bump-schema-version indexed-visibility invariant is preserved). The RESTful `add_function_field` `indexParams` field becomes optional accordingly.

### Verification

- UT: resolution table in `indexparamcheck` (explicit passthrough, empty/AUTOINDEX/metric-only/legacy-params-JSON resolution, only-metric restriction, BM25 metric forcing and wrong-metric rejection, dense/binary/dedup config selection, unsupported type, unknown function type); rootcoord `TestPrepareBoundFieldIndex` autoindex subcases incl. persisted-shape asserts; rootcoord DDL broadcast integration case (no-params add succeeds, bound index materialized with resolved params + canonical UserIndexParams); datacoord equivalence test (persisted bound autoindex + later `create_index(AUTOINDEX)` → `errIndexOperationIgnored`; different explicit index still conflicts); proxy `preExecuteAdd` accept/reject cases; RESTful omitted-`indexParams` case.
- e2e: the former negative case (raw `alter_collection_schema` without index params rejected) is flipped to a positive: the DDL succeeds and `describe_index` shows the resolved `SPARSE_INVERTED_INDEX` with metric `BM25`.

### Out of scope / follow-ups

- pymilvus still requires `index_params` in `add_function_field`; making it optional is a follow-up SDK change (raw gRPC/RESTful callers can omit it already).
- Unifying create_index's own two AUTOINDEX resolution branches with the shared helper.
- No change to the cloud (`autoIndex.enable=true`) behavior of overriding explicit index types — the bound-index explicit path keeps honoring user params as before.

### DIFF
--- a/internal/proxy/task.go
+++ b/internal/proxy/task.go
@@ -1085,9 +1085,10 @@ type alterCollectionSchemaTask struct {
 	Condition
 	*milvuspb.AlterCollectionSchemaRequest
 	*milvuspb.AlterCollectionSchemaResponse
-	ctx       context.Context
-	mixCoord  types.MixCoordClient
-	oldSchema *schemapb.CollectionSchema
+	ctx                  context.Context
+	mixCoord             types.MixCoordClient
+	oldSchema            *schemapb.CollectionSchema
+	collectionProperties []*commonpb.KeyValuePair
 }
 
 func (t *alterCollectionSchemaTask) TraceCtx() context.Context {
@@ -1187,10 +1188,11 @@ func (t *alterCollectionSchemaTask) preExecuteAdd(ctx context.Context) error {
 
 	// Validate the bound index params against the new field, mirroring the
 	// create_index path (index name rules, checker existence, data-type
-	// compatibility, train params, dimension filling). Validation-only: the
-	// request keeps the user's original params so the persisted UserIndexParams
-	// match the create_index convention; rootcoord independently derives the
-	// normalized IndexParams at prepare.
+	// compatibility, train params, dimension filling, AUTOINDEX resolution when
+	// index_type is omitted). Validation-only: the request keeps the user's
+	// original params so the persisted UserIndexParams match the create_index
+	// convention; rootcoord independently derives the normalized IndexParams at
+	// prepare.
 	if plan.Kind == schemautil.AlterSchemaAddFunctionField {
 		if err := validateIndexName(plan.IndexName); err != nil {
 			return err
@@ -1198,8 +1200,8 @@ func (t *alterCollectionSchemaTask) preExecuteAdd(ctx context.Context) error {
 		if err := indexparamcheck.ValidateIndexParamsSize(plan.IndexExtraParams...); err != nil {
 			return err
 		}
-		indexParamsMap, err := indexparamcheck.PrepareFunctionOutputIndexParams(
-			plan.Function.GetType(), plan.Field.GetName(), plan.IndexExtraParams)
+		indexParamsMap, _, err := indexparamcheck.PrepareFunctionOutputIndexParams(
+			plan.Function.GetType(), plan.Field, t.collectionProperties, plan.IndexExtraParams)
 		if err != nil {
 			return err
 		}
--- a/internal/proxy/task_index.go
+++ b/internal/proxy/task_index.go
@@ -52,8 +52,6 @@ const (
 	AutoIndexName = common.AutoIndexName
 	DimKey        = common.DimKey
 	IsSparseKey   = common.IsSparseKey
-
-	RefineTypeKey = "refine_type"
 )
 
 // mapVectorMetricToEmbListMetric maps element-level vector metrics to EmbList-level metrics for ArrayOfVector.
@@ -74,58 +72,6 @@ func mapVectorMetricToEmbListMetric(metricType string) string {
 	default:
 		return metricType
 	}
-}
-
-// adjustAutoIndexParamsByDataType adjusts autoindex params based on vector data type
-// If data_type is bf16 and refine_type is fp16/fp32, adjust to BF16
-// If data_type is fp16 and refine_type is bf16/fp32, adjust to FP16
-// Other refine_type values (e.g., sq8) are not modified
-func adjustAutoIndexParamsByDataType(config map[string]string, dataType schemapb.DataType) map[string]string {
-	if config == nil {
-		return config
-	}
-	refineType, hasRefine := config[RefineTypeKey]
-	if !hasRefine {
-		return config
-	}
-
-	refineTypeLower := strings.ToLower(refineType)
-	var requiredRefineType string
-
-	switch dataType {
-	case schemapb.DataType_Float16Vector:
-		// fp16 data requires fp16 refine, adjust if refine_type is bf16/fp32
-		if refineTypeLower == "bf16" || refineTypeLower == "fp32" {
-			requiredRefineType = "FP16"
-		}
-	case schemapb.DataType_BFloat16Vector:
-		// bf16 data requires bf16 refine, adjust if refine_type is fp16/fp32
-		if refineTypeLower == "fp16" || refineTypeLower == "fp32" {
-			requiredRefineType = "BF16"
-		}
-	}
-
-	if requiredRefineType == "" {
-		return config
-	}
-
-	adjusted := make(map[string]string, len(config))
-	for k, v := range config {
-		adjusted[k] = v
-	}
-	adjusted[RefineTypeKey] = requiredRefineType
-	return adjusted
-}
-
-func getDenseFloatAutoIndexParams(collectionProperties []*commonpb.KeyValuePair) map[string]string {
-	// autoindex is enabled only for cloud instance.
-	// and large_topk query mode is set in collection properties.
-	// we will use large_topk index params for dense float vector index when these two conditions are met.
-	if Params.AutoIndexConfig.Enable.GetAsBool() && common.IsQueryModeLargeTopK(collectionProperties...) {
-		return Params.AutoIndexConfig.LargeTopKIndexParams.GetAsJSONMap()
-	}
-
-	return Params.AutoIndexConfig.IndexParams.GetAsJSONMap()
 }
 
 type createIndexTask struct {
@@ -189,19 +135,6 @@ func (cit *createIndexTask) OnEnqueue() error {
 	return nil
 }
 
-func wrapUserIndexParams(metricType string) []*commonpb.KeyValuePair {
-	return []*commonpb.KeyValuePair{
-		{
-			Key:   common.IndexTypeKey,
-			Value: AutoIndexName,
-		},
-		{
-			Key:   common.MetricTypeKey,
-			Value: metricType,
-		},
-	}
-}
-
 func (cit *createIndexTask) parseFunctionParamsToIndex(indexParamsMap map[string]string) error {
 	if !cit.fieldSchema.GetIsFunctionOutput() {
 		return nil
@@ -220,6 +153,9 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	// Whether the metric came from the USER (before any autoindex config merge)
+	// — a config-injected metric is not a user choice.
+	_, userMetricSpecified := indexParamsMap[common.MetricTypeKey]
 
 	if jsonCastType, exist := indexParamsMap[common.JSONCastTypeKey]; exist {
 		indexParamsMap[common.JSONCastTypeKey] = strings.ToUpper(strings.TrimSpace(jsonCastType))
@@ -318,14 +254,14 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 
 			if typeutil.IsDenseFloatVectorType(cit.fieldSchema.DataType) ||
 				(typeutil.IsArrayOfVectorType(cit.fieldSchema.DataType) && typeutil.IsDenseFloatVectorType(cit.fieldSchema.ElementType)) {
-				autoIndexParams := getDenseFloatAutoIndexParams(cit.collectionProperties)
+				autoIndexParams := indexparamcheck.GetDenseFloatAutoIndexParams(cit.collectionProperties)
 				// override float vector index params by autoindex
 				// filter incompatible refine_type for fp16/bf16 vectors
 				dataType := cit.fieldSchema.DataType
 				if typeutil.IsArrayOfVectorType(cit.fieldSchema.DataType) {
 					dataType = cit.fieldSchema.ElementType
 				}
-				autoIndexParams = adjustAutoIndexParamsByDataType(autoIndexParams, dataType)
+				autoIndexParams = indexparamcheck.AdjustAutoIndexParamsByDataType(autoIndexParams, dataType)
 				for k, v := range autoIndexParams {
 					indexParamsMap[k] = v
 				}
@@ -387,7 +323,7 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 				if len(indexParamsMap) == numberParams {
 					// though we already know there must be metric type, how to make this safer to avoid crash?
 					metricType := autoIndexConfig[common.MetricTypeKey]
-					cit.newExtraParams = wrapUserIndexParams(metricType)
+					cit.newExtraParams = indexparamcheck.WrapUserIndexParams(metricType)
 					useAutoIndex(autoIndexConfig)
 					return nil
 				}
@@ -402,7 +338,7 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 					}
 
 					// only metric type is passed.
-					cit.newExtraParams = wrapUserIndexParams(metricType)
+					cit.newExtraParams = indexparamcheck.WrapUserIndexParams(metricType)
 					useAutoIndex(autoIndexConfig)
 					// make the users' metric type first class citizen.
 					indexParamsMap[common.MetricTypeKey] = metricType
@@ -415,13 +351,13 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 			var config map[string]string
 			if typeutil.IsDenseFloatVectorType(cit.fieldSchema.DataType) ||
 				(typeutil.IsArrayOfVectorType(cit.fieldSchema.DataType) && typeutil.IsDenseFloatVectorType(cit.fieldSchema.ElementType)) {
-				config = getDenseFloatAutoIndexParams(cit.collectionProperties)
+				config = indexparamcheck.GetDenseFloatAutoIndexParams(cit.collectionProperties)
 				// filter incompatible refine_type for fp16/bf16 vectors
 				dataType := cit.fieldSchema.DataType
 				if typeutil.IsArrayOfVectorType(cit.fieldSchema.DataType) {
 					dataType = cit.fieldSchema.ElementType
 				}
-				config = adjustAutoIndexParamsByDataType(config, dataType)
+				config = indexparamcheck.AdjustAutoIndexParamsByDataType(config, dataType)
 			} else if typeutil.IsSparseFloatVectorType(cit.fieldSchema.DataType) ||
 				(typeutil.IsArrayOfVectorType(cit.fieldSchema.DataType) && typeutil.IsSparseFloatVectorType(cit.fieldSchema.ElementType)) {
 				// override sparse float vector index params by autoindex
@@ -452,6 +388,24 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 			if !metricTypeExist && typeutil.IsArrayOfVectorType(cit.fieldSchema.DataType) {
 				if m, ok := indexParamsMap[common.MetricTypeKey]; ok {
 					indexParamsMap[common.MetricTypeKey] = mapVectorMetricToEmbListMetric(m)
+				}
+			}
+		}
+
+		// For a BM25 function output field the required metric is known
+		// authoritatively; when the metric in the map was injected by the
+		// autoindex config (not user-specified), force it to BM25 instead of
+		// failing the BM25 metric check below — same rule as the
+		// add_function_field bound-index resolution. The user-facing wrapped
+		// extra params must carry the same final metric.
+		if !userMetricSpecified && cit.fieldSchema.GetIsFunctionOutput() &&
+			cit.functionSchema.GetType() == schemapb.FunctionType_BM25 {
+			if m, ok := indexParamsMap[common.MetricTypeKey]; ok && m != metric.BM25 {
+				indexParamsMap[common.MetricTypeKey] = metric.BM25
+				for _, kv := range cit.newExtraParams {
+					if kv.GetKey() == common.MetricTypeKey {
+						kv.Value = metric.BM25
+					}
 				}
 			}
 		}
@@ -918,7 +872,7 @@ func (dit *describeIndexTask) Execute(ctx context.Context) error {
 		if params == nil {
 			metricType, err := funcutil.GetAttrByKeyFromRepeatedKV(MetricTypeKey, indexInfo.GetIndexParams())
 			if err == nil {
-				params = wrapUserIndexParams(metricType)
+				params = indexparamcheck.WrapUserIndexParams(metricType)
 			}
 		}
 		fieldName := field.Name
--- a/internal/rootcoord/ddl_callbacks_alter_collection_schema.go
+++ b/internal/rootcoord/ddl_callbacks_alter_collection_schema.go
@@ -37,6 +37,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/streaming/util/message"
 	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
+	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 	"github.com/milvus-io/milvus/pkg/v3/util/timestamptz"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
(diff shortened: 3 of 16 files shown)


internal/util/indexparamcheck/index_params_validation.go
```diff
@@ -112,25 +112,151 @@ func FillFunctionOutputIndexParams(functionType schemapb.FunctionType, indexPara
...
+
+	autoIndexCfg := &paramtable.Get().AutoIndexConfig
+	var config map[string]string
+	switch {
+	case typeutil.IsDenseFloatVectorType(field.GetDataType()):
+		config = AdjustAutoIndexParamsByDataType(GetDenseFloatAutoIndexParams(collectionProperties), field.GetDataType())
+	case typeutil.IsSparseFloatVectorType(field.GetDataType()):
+		config = autoIndexCfg.SparseIndexParams.GetAsJSONMap()
+	case typeutil.IsBinaryVectorType(field.GetDataType()):
+		if userMetricSpecified && funcutil.SliceContain(DeduplicateMetrics, userMetric) {
```
buqian-zilliz: **[P1] Resolve omitted MinHash params as a MinHash index**

A MinHash output is a BinaryVector, but when the newly accepted omitted/AUTOINDEX request has no metric, `userMetricSpecified` is false and this branch selects the generic binary config (default `BIN_IVF_FLAT`/`HAMMING`). QueryNode requires `MHJACCARD` for MinHash searches and the dedicated config is `MINHASH_LSH`/`MHJACCARD`, so the DDL succeeds while persisting an index that cannot serve the function correctly. Select the MinHash config from `functionType` when the metric is omitted and reject conflicting user metrics.

internal/proxy/task_index.go
```diff
@@ -387,7 +320,7 @@ func (cit *createIndexTask) parseIndexParams(ctx context.Context) error {
 				if len(indexParamsMap) == numberParams {
 					// though we already know there must be metric type, how to make this safer to avoid crash?
 					metricType := autoIndexConfig[common.MetricTypeKey]
-					cit.newExtraParams = wrapUserIndexParams(metricType)
+					cit.newExtraParams = indexparamcheck.WrapUserIndexParams(metricType)
```
buqian-zilliz, line 326: **[P2] Make BM25 AutoIndex requests reach deduplication**

The new bound-index path stores an omitted-params BM25 index as `AUTOINDEX` with metric `BM25`, but the follow-up create path canonicalizes the sparse AutoIndex config's default metric (`IP`) here and retains the same value in the cloud branch. `FillFunctionOutputIndexParams` then rejects an omitted or AUTOINDEX-only `create_index` before DataCoord can dedupe it against the bound index. Apply the same function-aware BM25 override before building `newExtraParams` and validating the create request.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -255,6 +255,13 @@ func (c *Core) prepareBoundFieldIndex(ctx context.Context, coll *model.Collectio
 	typeParams := lo.Filter(plan.Field.GetTypeParams(), func(kv *commonpb.KeyValuePair, _ int) bool {
 		return kv.GetKey() != common.MmapEnabledKey && kv.GetKey() != common.WarmupKey
 	})
+	userIndexParams := plan.IndexExtraParams
+	if autoResolved {
+		// Canonical AUTOINDEX user params (create_index convention): datacoord's
+		// checkParams dedupes a later create_index(AUTOINDEX) on this field by
+		// comparing exactly these pairs.
+		userIndexParams = indexparamcheck.WrapUserIndexParams(indexParamsMap[common.MetricTypeKey])
```
buqian-zilliz, line 282: **[P2] Canonicalize cloud UserIndexParams consistently**

`autoResolved` also covers requests where `index_type` was omitted, but this line always adds an explicit `AUTOINDEX` pair. With `autoIndex.enable=true`, `createIndexTask` forwards the caller's original user params instead of canonicalizing them, so repeating a valid metric-only BM25 request sends one pair while this bound index stores two. DataCoord rejects that length mismatch before metric compatibility and returns a distinct-index conflict; canonicalize the cloud create path or preserve one representation for both producers.

internal/util/indexparamcheck/index_params_validation.go
MrPresent-Han: Fixed in e7c3f5b44d. The binary branch now selects the deduplicate config (`MINHASH_LSH`/`MHJACCARD`) whenever the function is MinHash and the user did not specify a metric — same function-knows-best rule as the BM25 metric forcing, so an omitted-params MinHash add no longer persists a `BIN_IVF_FLAT`/`HAMMING` index that cannot serve MinHash searches. Covered by new UTs at both layers (`indexparamcheck`: "MinHash without metric picks deduplicate config"; rootcoord: `TestPrepareBoundFieldIndex/MinHash empty params resolve to deduplicate config`).

On the second suggestion (rejecting conflicting user metrics): I kept the user-specified metric first-class (e.g. explicit `HAMMING` still resolves via the generic binary config) rather than adding a rejection, because the explicit-index_type path accepts MinHash+HAMMING today (`FillFunctionOutputIndexParams` has no MinHash metric rule) — rejecting only on the AutoIndex path would make the two paths asymmetric. A Fill-level MinHash metric rule that covers both paths (and create_index) feels like the right follow-up; happy to file it separately if you agree.

internal/proxy/task_index.go
MrPresent-Han: Fixed in e7c3f5b44d. `parseIndexParams` now records whether the metric was user-specified before any autoindex config merge, and for a BM25 function output field a config-injected metric is forced to BM25 before `FillFunctionOutputIndexParams` runs — in the OSS branch the wrapped `newExtraParams` metric is rewritten to the same final value, and in the cloud branch only the build params are forced while the caller's raw extra params are preserved (matching that branch's existing convention). So an omitted/AUTOINDEX-only `create_index` on the BM25 output field now resolves to `[AUTOINDEX, BM25]` and reaches DataCoord's dedupe against the bound index instead of failing the metric check at the proxy.

New UT: `Test_parseIndexParams_AutoIndexBM25FunctionOutput` (empty params, AUTOINDEX-only, user-specified metric unchanged, cloud-mode variant).

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han: Fixed in e7c3f5b44d, via the second option (one representation per producer mode) rather than canonicalizing the cloud create path: `prepareBoundFieldIndex` now persists exactly what `create_index` would persist for the equivalent request in the same deployment mode — `wrapUserIndexParams(finalMetric)` when `autoIndex.enable=false` (OSS convention, unchanged), and the caller's raw extra params (empty for an omitted request) when `autoIndex.enable=true`, matching the cloud branch that forwards original user params. That removes the UserIndexParams length mismatch: in cloud, a bound index from an omitted-params add now compares like a prior `create_index` with empty extra params.

I stayed away from canonicalizing the cloud create path itself because it would regress existing deployments: cloud indexes already persisted with raw params would mismatch newly-wrapped repeat requests in `checkParams`, turning today's idempotent re-creates into distinct-index conflicts. New UT: `TestPrepareBoundFieldIndex/cloud mode persists raw user params for auto-resolved index`.

internal/util/indexparamcheck/index_params_validation.go
```diff
@@ -112,25 +112,156 @@ func FillFunctionOutputIndexParams(functionType schemapb.FunctionType, indexPara
...
+	case typeutil.IsDenseFloatVectorType(field.GetDataType()):
+		config = AdjustAutoIndexParamsByDataType(GetDenseFloatAutoIndexParams(collectionProperties), field.GetDataType())
+	case typeutil.IsSparseFloatVectorType(field.GetDataType()):
+		config = autoIndexCfg.SparseIndexParams.GetAsJSONMap()
+	case typeutil.IsBinaryVectorType(field.GetDataType()):
+		// A MinHash output field with no user metric must resolve to the
+		// deduplicate config (MINHASH_LSH/MHJACCARD): the generic binary config
+		// would build an index that cannot serve MinHash searches. Same
+		// function-knows-best rule as the BM25 metric forcing below.
+		if (userMetricSpecified && funcutil.SliceContain(DeduplicateMetrics, userMetric)) ||
```
buqian-zilliz, line 250: **[P2] Honor the deduplicate AutoIndex feature gate**

When `autoIndex.enable=true` but `autoIndex.params.deduplicate.enable=false`, a metric-only `MHJACCARD` request or an omitted-metric MinHash request reaches this branch and selects `DeduplicateIndexParams` anyway. The existing create-index AutoIndex path rejects the same deduplicate selection under that configuration, so add-function-field bypasses the operator's rollout gate and can materialize `MINHASH_LSH` while it is disabled. Check `EnableDeduplicateIndex` before selecting this config in cloud AutoIndex mode.

scripts/core_build.sh
```diff
@@ -295,6 +295,7 @@ if [[ ${RUN_CPPLINT} == "ON" ]]; then
   fi
   echo "clang-format check passed!"
 else
+  jobs=6
```
buqian-zilliz: **[P2] Preserve the configured build parallelism**

Do not overwrite the configured parallelism here. The script deliberately accepts a caller-supplied `jobs` value and otherwise derives it from `nproc`/`sysctl` near the top, but this assignment runs before every actual build, so `jobs=1` for memory-constrained runners and larger values for capable builders are both ignored. This can cause avoidable OOM build failures or arbitrary throttling. Remove this assignment and retain the existing initialization logic.

internal/util/indexparamcheck/index_params_validation.go
```diff
@@ -112,25 +112,156 @@ func FillFunctionOutputIndexParams(functionType schemapb.FunctionType, indexPara
...
 	indexParamsMap, err := ExpandIndexParams(extraParams)
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
-	indexType := indexParamsMap[common.IndexTypeKey]
-	if indexType == "" || indexType == common.AutoIndexName {
-		return nil, merr.WrapErrParameterInvalidMsg(
-			"an explicit index_type is required for the bound index of function output field %q", fieldName)
+	if indexType := indexParamsMap[common.IndexTypeKey]; indexType != "" && indexType != common.AutoIndexName {
```
buqian-zilliz: **[P2] Do not treat an explicit empty index type as AutoIndex**

This condition folds an explicitly supplied `index_type: ""` into the new AutoIndex path. In the default OSS path, `create_index` distinguishes an absent key from an existing empty value: the latter skips both AutoIndex handlers and is rejected as an invalid index type. The bound-index path previously rejected it too, so this now silently turns a malformed request into a concrete server-chosen index and breaks the claimed create-index parity. Resolve only when the key is absent or its value is exactly `AUTOINDEX`.

MrPresent-Han: Fixed in 78f9843f56. `PrepareFunctionOutputIndexParams` now uses the same presence-based rule as `create_index`: AutoIndex resolution triggers only when the `index_type` KEY is absent or its value is exactly `AUTOINDEX`. A present-but-empty value stays on the explicit path and is rejected by the checker-existence validation at both boundaries ("invalid index type"), matching the create_index behavior.

New UTs: `indexparamcheck` "explicit empty index type stays on explicit path" (asserts resolved=false and the downstream rejection) and rootcoord `TestPrepareBoundFieldIndex/explicit empty index type rejected, not auto-resolved`.

scripts/core_build.sh
MrPresent-Han: Fixed in 78f9843f56 — and thanks for catching it: this line should never have been in the PR at all. It was a machine-local build-concurrency cap from my dev checkout that got swept into the commit by a blanket `git add -u` during an amend. `scripts/core_build.sh` is now reverted to the upstream state and is no longer part of this PR; the configured/derived parallelism logic is untouched.

internal/util/indexparamcheck/index_params_validation.go
MrPresent-Han: Fixed in 78f9843f56. The deduplicate-config selection (both the user-`MHJACCARD` and the omitted-metric MinHash entry) now honors the feature gate the same way `create_index` does: with `autoIndex.enable=true` and `autoIndex.params.deduplicate.enable=false` the request is rejected ("Deduplicate index is not enabled, cannot resolve the bound index ... via AutoIndex") instead of materializing `MINHASH_LSH` behind the operator's rollout gate. I deliberately left the OSS mode ungated because the OSS `create_index` branch selects `DeduplicateIndexParams` without consulting `EnableDeduplicateIndex` — keeping the two paths behaviorally identical per deployment mode; if you'd rather gate OSS as well, that should change both paths together as a separate discussion.

New UTs: "cloud dedup gate rejects MinHash AutoIndex when disabled" (omitted metric + explicit MHJACCARD) and "cloud dedup gate enabled resolves MinHash AutoIndex".

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -193,8 +193,8 @@ func (c *Core) broadcastAlterCollectionSchemaAdd(ctx context.Context, broadcaste
 // apply is a pure idempotent write (a replayed callback rebuilds the identical
 // index), and every input-dependent rejection happens before anything commits.
 func (c *Core) prepareBoundFieldIndex(ctx context.Context, coll *model.Collection, plan *schemautil.AlterSchemaAddPlan) (*indexpb.FieldIndex, error) {
-	indexParamsMap, err := indexparamcheck.PrepareFunctionOutputIndexParams(
-		plan.Function.GetType(), plan.Field.GetName(), plan.IndexExtraParams)
+	indexParamsMap, autoResolved, err := indexparamcheck.PrepareFunctionOutputIndexParams(
```
buqian-zilliz, line 197: **[P2] Persist Knowhere defaults for MinHash fallback consistency**

Apply the build-stage Knowhere defaults before materializing this bound index. The normal `create_index` path merges `KnowhereConfig` at `BuildStage` before persisting `IndexParams`, but this path serializes the helper result directly. DataCoord adds those defaults only to per-segment build jobs, while QueryNode uses the unmerged collection metadata for MinHash brute-force fallback. With configured `MINHASH_LSH` values such as `mh_lsh_band` or `mh_element_bit_width`, indexed segments use the configured values while growing or stale segments fall back to 1/8, so one query can evaluate segments with incompatible MinHash settings. Merge the same defaults before validation and serialization.

internal/proxy/task.go
```diff
@@ -1187,19 +1187,28 @@ func (t *alterCollectionSchemaTask) preExecuteAdd(ctx context.Context) error {
...
 	if plan.Kind == schemautil.AlterSchemaAddFunctionField {
 		if err := validateIndexName(plan.IndexName); err != nil {
 			return err
 		}
 		if err := indexparamcheck.ValidateIndexParamsSize(plan.IndexExtraParams...); err != nil {
 			return err
 		}
-		indexParamsMap, err := indexparamcheck.PrepareFunctionOutputIndexParams(
-			plan.Function.GetType(), plan.Field.GetName(), plan.IndexExtraParams)
+		collID, err := globalMetaCache.GetCollectionID(ctx, t.GetDbName(), t.GetCollectionName())
```
buqian-zilliz: **[P3] Reuse the collection metadata already fetched**

`AlterCollectionSchema` has already fetched a `DescribeCollectionResponse` containing the collection ID and properties, but this new precheck performs `GetCollectionID` and `GetCollectionInfo` again. In remote service-provider mode, the first cache miss issues a second coordinator describe and can fail an otherwise valid DDL; in cached mode it still repeats cache and metric work. The currently allowed BM25/MinHash outputs are sparse or binary, so `PrepareFunctionOutputIndexParams` does not consult `collectionProperties` for them. Reuse the response metadata or defer the lookup until a supported data type actually needs it.

internal/util/indexparamcheck/index_params_validation.go
buqian-zilliz: This issue still exists for the REST endpoint: `IndexParam.IndexType` is a plain string and `convertToExtraParams` drops empty values, so `indexType: ""` becomes an absent key before the new presence check and is still AutoIndex-resolved. Preserve JSON field presence or reject the explicit empty value in the HTTP layer.

MrPresent-Han: You're right about the mechanics — at the REST layer an explicit `"indexType": ""` is normalized away by `convertToExtraParams` before the presence check can see it. We'd prefer to keep that behavior for compatibility, though: the same helper backs `/indexes/create` and `/collections/create` (with-index quick setup), where `indexType: ""` has always normalized to absent-and-AutoIndex-resolved, and `metricType` follows the identical empty-means-unset convention. Distinguishing `""` from an omitted field would also require changing the JSON shape to `*string`, and tightening only this endpoint would make it stricter than its sibling index-creating endpoints.

So the layered contract stays: the gRPC boundary is strict (present-but-empty `index_type` is rejected, as fixed in 78f9843f56), while REST keeps its established empty-means-unset normalization consistently across endpoints. Happy to revisit REST-wide presence semantics separately if you think it's worth it.

internal/proxy/task.go
MrPresent-Han: Good catch, fixed in 592ed830b8. `AlterCollectionSchema` in `impl.go` already has the `DescribeCollectionResponse`, so the task now carries its `Properties` in a new `collectionProperties` field and `preExecuteAdd` uses that directly — the extra `GetCollectionID` + `GetCollectionInfo` calls (and the second describe they could trigger on a cold cache in remote mode) are gone. As you noted the currently supported BM25/MinHash outputs are sparse/binary and don't consult properties, but reusing the already-fetched metadata is strictly better: fewer calls, one fewer independent failure path, and `oldSchema` + properties now come from the same snapshot. Net -12 lines; the metacache-init the tests previously needed is dropped too.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
MrPresent-Han: Fixed in 7768d58ae8. `prepareBoundFieldIndex` now applies `KnowhereConfig.MergeIndexParams(indexType, BuildStage, ...)` after validation and before serialization — the same non-destructive build-stage merge `create_index` does — gated on `KnowhereConfig.Enable` (default true). So an operator-configured build param like `knowhere.MINHASH_LSH.build.mh_lsh_band` is now persisted in the bound index's `IndexParams`, keeping the built index and QueryNode's brute-force fallback on the same MinHash settings.

This covers both the explicit and the AutoIndex-resolved bound paths (the gap predates this PR — it's been there for the explicit path since the bound-index feature landed in #51105 — so fixing both here). It's a no-op under the default empty knowhere config. New UT: `TestPrepareBoundFieldIndex/knowhere build defaults merged into persisted index params`.

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -214,6 +215,19 @@ func (c *Core) prepareBoundFieldIndex(ctx context.Context, coll *model.Collectio
...
+	// Merge the YAML build-stage knowhere defaults into the persisted params, as
+	// create_index does before persisting. Without this, an operator-configured
+	// build param (e.g. MINHASH_LSH mh_lsh_band/mh_element_bit_width) reaches the
+	// index build via datacoord's own merge but not the collection metadata that
+	// QueryNode uses for brute-force fallback, so indexed and fallback segments
+	// could evaluate one query with incompatible settings. Pre-existing gap for
+	// the explicit bound path too; fixing both here. No-op with the default
+	// (empty) knowhere config.
+	if Params.KnowhereConfig.Enable.GetAsBool() {
+		if indexParamsMap, err = Params.KnowhereConfig.MergeIndexParams(indexType, paramtable.BuildStage, indexParamsMap); err != nil {
```
buqian-zilliz, line 223: **[P2] Validate Knowhere defaults after merging them**

Move this merge before `ValidateFieldIndexParams`, or re-run field validation on the merged map. The map checked above does not contain the Knowhere build-stage values added here, while the later `ValidateIndexParams` only checks duplicate, mmap, and offset-cache structure. An invalid or oversized operator-supplied build default can therefore be serialized into the bound `FieldIndex` and alter-schema WAL instead of rejecting the DDL as `create_index` does, leaving the mandatory BM25/MinHash backfill index unable to build.

MrPresent-Han: Good catch, fixed in ce52a937b8. Moved the knowhere merge to before `ValidateFieldIndexParams` (right after the checker-existence check), so the merged build-stage defaults now go through the full field validation — size, dimension, and `CheckTrain` — exactly as `create_index` validates its merged map. An invalid or oversized operator-supplied build default now rejects the DDL instead of being serialized into the bound `FieldIndex`/WAL. `indexType` is captured before the merge and the merge never touches it, so the checker-existence check above is unaffected.

New UT: `TestPrepareBoundFieldIndex/oversized knowhere build default rejected after merge` (sets an over-limit `knowhere.MINHASH_LSH.build.mh_lsh_band` and asserts the "exceeds limit" rejection).

internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -255,6 +271,16 @@ func (c *Core) prepareBoundFieldIndex(ctx context.Context, coll *model.Collectio
...
 		return kv.GetKey() != common.MmapEnabledKey && kv.GetKey() != common.WarmupKey
 	})
+	// Persist exactly what create_index would persist for the equivalent request,
+	// per deployment mode, so datacoord's checkParams (which compares these pairs)
+	// dedupes a later create_index instead of reporting a distinct-index conflict:
+	// OSS wraps autoindex requests into the canonical AUTOINDEX+metric pair
+	// (wrapUserIndexParams), while the cloud branch (autoIndex.enable=true) keeps
+	// the caller's raw extra params.
+	userIndexParams := plan.IndexExtraParams
+	if autoResolved && !Params.AutoIndexConfig.Enable.GetAsBool() {
```
buqian-zilliz, line 281: **[P2] Canonicalize equivalent cloud AutoIndex forms**

When `autoIndex.enable=true`, this keeps the bound index's caller spelling, so the two accepted AutoIndex forms persist different `UserIndexParams`: omission stores zero pairs, while `index_type=AUTOINDEX` stores one. A later `create_index` using the other form reaches DataCoord's `checkParams`, which rejects on the length mismatch before AutoIndex compatibility handling and returns a distinct-index conflict even though both requests resolve to the same index. Normalize the omitted and explicit `AUTOINDEX` forms consistently, or canonicalize them during deduplication.

MrPresent-Han: This is a real inconsistency, but it's a pre-existing property of cloud-mode `create_index` rather than something new here, and it can't be fixed in the bound path alone without reintroducing the earlier cross-producer mismatch — so I'd prefer to defer it to a shared follow-up.

Details:
- The two forms persist the **same** built index (`IndexParams` is identical — same concrete type/metric/build params). Only `UserIndexParams` — the raw-request record `checkParams` uses for idempotency dedup — differs: omitted → `[]`, `AUTOINDEX` → `[{index_type:AUTOINDEX}]` (`index_type` isn't a configable key, so it isn't filtered out before the length comparison).
- In cloud mode, `create_index` itself already has this: its cloud branch stores the caller's raw extra params verbatim (doesn't rebuild `newExtraParams`), so `create_index(field)` and `create_index(field, AUTOINDEX)` on the same field already fail to cross-dedupe today, independent of this PR. The bound path mirrors that raw storage on purpose — that's exactly what the earlier review asked for (match what `create_index` persists per deployment mode).
- In OSS mode there's no issue: both `create_index` (via `handle()`) and this bound path normalize both spellings to `WrapUserIndexParams(metric)`, so they dedupe consistently.
- Normalizing only the bound path in cloud mode would diverge from cloud `create_index` (which stays raw) and reintroduce the cross-producer length mismatch from the earlier comment. The correct, uniform fix is to canonicalize the AUTOINDEX forms during dedup in DataCoord's `checkParams` so `[]`/`[AUTOINDEX]`/`[AUTOINDEX,metric]` compare equal when they resolve to the same index — that fixes both `create_index` and `add_function_field` at once.

I'll file that `checkParams` canonicalization as a separate follow-up (it touches shared dedup + existing cloud `create_index` behavior, with backward-compat implications for already-persisted indexes), and keep this PR mirroring `create_index` per mode. Let me know if you'd rather pull it in here.

PR #52090, opened 2026-07-31
fix: [3.0] support TLS-only MixCoord client connections (#52089)

santiago-wjq:
Cherry-pick of #52089 to the 3.0 branch.

## What this PR does

- adds the default-off `mixCoord.grpc.clientTlsEnabled` setting for outbound MixCoord connections
- applies it to both the classic MixCoord gRPC client and the StreamingCoord client
- preserves cluster-wide internal TLS precedence and leaves component gRPC servers unchanged

The setting uses `tls.caPemPath` and `internaltls.sni` when enabled. It does not depend on `common.security.tlsMode`, so deployments with external TLS and plaintext internal coordinator endpoints retain their current behavior.

## Verification

- `cd pkg && go test ./util/paramtable -run 'TestResolveMixCoordClientTLS|TestInternalTLSParams' -count=1`
- added regression coverage for `tlsMode=1` with the new setting disabled, directional certificate loading, internal TLS precedence, and fail-closed certificate errors
- `git diff --check`

The two client-package tests require the repository's native RocksDB/Milvus Core/Kafka build dependencies, which are not installed in the local environment; CI will execute them.

issue: #52091
pr: #52089

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -328,6 +328,9 @@ rocksmq:
 # Related configuration of mixCoord
 mixCoord:
   enableActiveStandby: false
+  grpc:
+    # Enable TLS only for outbound connections to MixCoord. This does not enable TLS on this component's gRPC server.
+    clientTlsEnabled: false
 
 # Related configuration of rootCoord, used to handle data definition language (DDL) and data control language (DCL) requests
 rootCoord:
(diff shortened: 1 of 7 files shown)


santiago-wjq:
Closing as superseded by the Cloud/Kite-side topology fix. The original Milvus GCP model keeps component-to-component traffic on the internal TLS setting; external TLS mode must not be reused for the QueryNode-to-coordinator bus.

Issue #52092, opened 2026-07-31
[Bug]: ScoreBasedBalancer's scoreUnbalanceTolerationFactor check never blocks moves once target priority goes negative


next0262:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: v2.6.17 (also confirmed still present on master)
- Deployment mode(standalone or cluster): cluster
- MQ type(rocksmq, pulsar or kafka): pulsar
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): Ubuntu
- CPU/Memory: 
- GPU: 
- Others: queryCoord.balancer = ScoreBasedBalancer (default), scoreUnbalanceTolerationFactor = 0.05 (default)
```

### Current Behavior

Was looking into why our balancer kept moving a handful of small segments back and forth between the same two querynodes over and over (same segment going A->B->A within a minute, repeatedly), even though nothing looked wrong with either node. Traced it to `commonScoreBasedBenefitEvaluator.HasEnoughBenefit` in `internal/querycoordv2/assign/assign_policy_components.go`:

```go
oldPriorityDiff := math.Abs(sourceScore - targetScore)
if oldPriorityDiff < targetScore*params.Params.QueryCoordCfg.ScoreUnbalanceTolerationFactor.GetAsFloat() {
    return false
}
```

`sourceScore`/`targetScore` here are `NodeItem.getPriority()`, i.e. `currentScore - assignedScore`. A balance target is by definition under its quota, so its priority is almost always negative. That makes `targetScore * factor` negative, while `oldPriorityDiff` is always >= 0 - so this never triggers, meaning the toleration check basically never blocks a move no matter what `scoreUnbalanceTolerationFactor` is set to.

For comparison, v2.3's `score_based_balancer.go` had pretty much the same formula but compared against the node's total score, which was always positive back then:

```go
unbalance := float64(fromPriority - toPriority)
if unbalance < float64(toPriority)*params.Params.QueryCoordCfg.ScoreUnbalanceTolerationFactor.GetAsFloat() {
    break
}
```

The v2.4 refactor changed priority to a delta-from-average (can go negative) but nobody updated this check, so it's basically been dead code ever since.

[...]

PR #52093, opened 2026-07-31
fix: use assigned score instead of priority as toleration baseline in ScoreBasedBalancer

next0262:
issue: #52092

`HasEnoughBenefit`'s toleration check compares the priority diff against `targetScore * factor`, but target priority (`currentScore - assignedScore`) is negative for basically every real balance target, so the check ends up a no-op no matter what the factor is set to. Been broken since the v2.4 priority refactor - v2.3 compared against the node's total score, which was always positive.

Switched to `targetNode.GetAssignedScore()` (always non-negative) as the baseline instead - restores the original "ignore diffs under X% of node score" behavior. Added a test for the negative-priority case.

### DIFF
--- a/internal/querycoordv2/assign/assign_policy_components.go
+++ b/internal/querycoordv2/assign/assign_policy_components.go
@@ -63,8 +63,9 @@ type benefitEvaluator interface {
 	// HasEnoughBenefit checks if the assignment from source to target is beneficial enough
 	// sourceScore: current score of source node
 	// targetScore: current score of target node
+	// targetAssignedScore: assigned (quota) score of target node, always non-negative
 	// resourceScore: score of the resource to be assigned
-	HasEnoughBenefit(sourceScore float64, targetScore float64, resourceScore int64) bool
+	HasEnoughBenefit(sourceScore float64, targetScore float64, targetAssignedScore float64, resourceScore int64) bool
 }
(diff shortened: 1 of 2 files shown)


next0262:
Gentle ping @weiliu1031 @congqixia 🙏

We hit this in production last week — with the toleration check effectively dead, the
balancer kept shuttling the same small segments between nodes (A→B→A within 46s, 180+
moves/day), and the cold-cache landings amplified a search outage. We're on a
`globalRowCountFactor` workaround until this lands.

One-line fix + regression test, CI green. Details: #52092

weiliu1031:
@next0262 Nice catch! The negative target priority makes the toleration check effectively a no-op. Using the target node's assigned score as a stable, non-negative baseline looks like the right minimal fix and restores the intended behavior.

Thanks for tracking this down, explaining the root cause clearly, and adding a focused regression test!

next0262:
Thanks for the quick review! @weiliu1031 🙏 Could this be cherry-picked to release-2.6?
We're running 2.6.x in production where we hit this.

weiliu1031:
> Thanks for the quick review! @weiliu1031 🙏 Could this be cherry-picked to release-2.6? We're running 2.6.x in production where we hit this.

done

Issue #52048, opened 2026-07-30
[Bug]: REST API v2 returns Array fields in raw protobuf ScalarField format instead of JSON arrays


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
Milvus version: v2.6.21 (also reproduced on v2.6.20; v2.6.21 re-verified on 2026-07-24)
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq (standalone default)
- SDK version(e.g. pymilvus v2.0.0rc2): REST API v2 (RESTful HTTP via curl; no SDK)
- OS(Ubuntu or CentOS): macOS host / Linux container (Docker Desktop)
- CPU/Memory: 16 GB
- GPU: N/A
- Others: FLAT index, COSINE metric
```

### Current Behavior

When retrieving Array-type fields via the REST API v2 (query, search, or get endpoints), the response contains the raw protobuf `ScalarField` message structure instead of a native JSON array.

Observed output for a VarChar Array field:
```json
{"tags": {"Data": {"StringData": {"data": ["a", "b", "c"]}}}}
```
instead of the expected `{"tags": ["a", "b", "c"]}`.

This affects ALL Array element types and all three retrieval endpoints (`query`, `search`, `get`):
- VarChar Array → `{"Data": {"StringData": {"data": [...]}}}`
- Int64 Array → `{"Data": {"LongData": {"data": [...]}}}`
- Float Array → `{"Data": {"FloatData": {"data": [...]}}}`

Empty arrays are even worse — they return `{"tags": {"Data": {"StringData": {}}}}` (the `data` key is missing entirely) instead of `{"tags": []}`.

Actual output from the reproduction below:
```json
// Query:
{"code":0,"data":[
  {"pk":1,"str_arr":{"Data":{"StringData":{"data":["hello","world"]}}},"int_arr":{"Data":{"LongData":{"data":[10,20,30]}}}},
  {"pk":2,"str_arr":{"Data":{"StringData":{}}},"int_arr":{"Data":{"LongData":{}}}}
]}

// Search:
{"code":0,"data":[
  {"distance":1,"pk":1,"str_arr":{"Data":{"StringData":{"data":["hello","world"]}}},"int_arr":{"Data":{"LongData":{"data":[10,20,30]}}}}
]}

// Get:
{"code":0,"data":[
[...]

PR #52096, opened 2026-07-31
fix: return native JSON arrays from REST APIs

xiaofan-luan:
issue: #52048

## What changed

- Unwrap protobuf `ScalarField` rows before serializing REST Array fields.
- Return native JSON arrays for all supported scalar element types.
- Normalize empty arrays to `[]` and preserve nullable rows as `null`.
- Apply `Accept-Type-Allow-Int64` semantics to Int64 array elements.
- Add exact JSON regression coverage for mixed Bool, Int, Int64, Float, Double, and String arrays, field-specific nullable bitmaps, empty arrays, escaped/non-ASCII strings, search scores, custom primary keys, and both Int64 response modes. Raw JSON assertions verify values beyond the JavaScript safe-integer range are emitted with the exact expected digits and quoting.

## Root cause

`buildQueryResp` placed the per-row `*schemapb.ScalarField` directly into the response map. The JSON encoder therefore exposed the protobuf oneof wrapper (`Data.StringData`, `Data.LongData`, and so on) instead of the contained values.

## Validation

```text
go test -tags dynamic,test -gcflags='all=-N -l' -count=1 ./internal/distributed/proxy/httpserver -run 'TestBuildQueryResp|TestScalarFieldToRESTAny|TestArray$'
```

The complex JSON regression test was also run in a temporary worktree with only the old direct-`ScalarField` response assignment restored. Both Int64 response-mode cases failed and showed the leaked protobuf wrappers, confirming the test catches the original defect.

The broader HTTP server package reaches an unrelated pre-existing `TestSearchV2/search#09` assertion failure that also reproduces on untouched `origin/master` under the same local test environment.

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -448,11 +448,21 @@ proxy:
     # itself is the risk the setting exists to remove; values below 1 are read as 1.
     maxExprParamsDepth: 100
     # high-level restful api, return a JSON field as the document it holds rather than as a string.
-    # A JSON field reads back as "{\"a\":1}" by default, while the same value in the dynamic field reads back as {"a":1}.
-    # Turning this on removes that difference. Rows written before the insert path stopped storing non-JSON bytes may not
-    # hold a document; if any row in a response is such a row, every JSON field in that response falls back to the string
-    # form and a warning is logged, so a caller always sees one shape or the other and never a mixture.
-    nativeJSONResponse: false
+    # Turn this off only to keep clients written against the older shape working while they migrate: there a JSON field read
+    # back as "{\"a\":1}" while the same value in the dynamic field read back as {"a":1}. The insert path follows the same
+    # switch, so either shape can be sent back unchanged: while the field reads back as text, a JSON string is read as the
+    # document it spells; once it reads back as the document itself, a string is stored as the string it is. Rows written
+    # before the insert path stopped storing non-JSON bytes may not hold a document; if any row in a response is such a row,
+    # every JSON field in that response falls back to the string form and a warning is logged, so a caller always sees one
+    # shape or the other and never a mixture.
+    nativeJSONResponse: true
+    # high-level restful api, whether to return Array fields wrapped in the raw protobuf ScalarField shape
+    # ({"tags":{"Data":{"StringData":{"data":["a","b"]}}}}) instead of a native JSON array ({"tags":["a","b"]}).
+    # Only enable this to keep clients written against the old, incorrect shape working while they migrate;
+    # it will be removed in a future release. It covers the shape of a top-level Array field and nothing else:
+    # a struct array's sub-fields are unaffected, and so is Accept-Type-Allow-Int64, which renders an Int64 as a
+    # string wherever one appears, including inside either kind of array.
+    legacyArrayResponse: false
     enablePprof: true # Whether to enable pprof middleware on the metrics port
     readHeaderTimeout: 5s # HTTP server timeout for reading request headers
     readTimeout: 0s # HTTP server timeout for reading the entire request, including the body. 0 disables this timeout
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -432,6 +432,11 @@ proxy:
     debug_mode: false # Whether to enable http server debug mode
     port:  # high-level restful api
     acceptTypeAllowInt64: true # high-level restful api, whether http client can deal with int64
+    # high-level restful api, whether to return Array fields wrapped in the raw protobuf ScalarField shape
+    # ({"tags":{"Data":{"StringData":{"data":["a","b"]}}}}) instead of a native JSON array ({"tags":["a","b"]}).
+    # Only enable this to keep clients written against the old, incorrect shape working while they migrate;
+    # it will be removed in a future release.
+    legacyArrayResponse: false
     enablePprof: true # Whether to enable pprof middleware on the metrics port
     readHeaderTimeout: 5s # HTTP server timeout for reading request headers
     readTimeout: 0s # HTTP server timeout for reading the entire request, including the body. 0 disables this timeout
--- a/internal/distributed/proxy/httpserver/handler_v1.go
+++ b/internal/distributed/proxy/httpserver/handler_v1.go
@@ -557,7 +557,7 @@ func (h *HandlersV1) query(c *gin.Context) {
 			mlog.Warn(ctx, "high level restful api, fail to deal with query result", mlog.Any("response", response), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+				HTTPReturnMessage: resultErrMessage(err),
 			})
 		} else {
 			HTTPReturnStream(c, http.StatusOK, gin.H{HTTPReturnCode: http.StatusOK, HTTPReturnData: outputData})
@@ -637,7 +637,7 @@ func (h *HandlersV1) get(c *gin.Context) {
 			mlog.Warn(ctx, "high level restful api, fail to deal with get result", mlog.Any("response", response), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+				HTTPReturnMessage: resultErrMessage(err),
 			})
 		} else {
 			HTTPReturnStream(c, http.StatusOK, gin.H{HTTPReturnCode: http.StatusOK, HTTPReturnData: outputData})
@@ -1036,7 +1036,7 @@ func (h *HandlersV1) search(c *gin.Context) {
 				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: resultErrMessage(err),
 				})
 			} else {
 				HTTPReturnStream(c, http.StatusOK, gin.H{HTTPReturnCode: http.StatusOK, HTTPReturnData: outputData})
--- a/internal/distributed/proxy/httpserver/handler_v1.go
+++ b/internal/distributed/proxy/httpserver/handler_v1.go
@@ -557,7 +557,7 @@ func (h *HandlersV1) query(c *gin.Context) {
 			mlog.Warn(ctx, "high level restful api, fail to deal with query result", mlog.Any("response", response), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+				HTTPReturnMessage: resultErrMessage(err),
 			})
 		} else {
 			HTTPReturnStream(c, http.StatusOK, gin.H{HTTPReturnCode: http.StatusOK, HTTPReturnData: outputData})
@@ -636,7 +636,7 @@ func (h *HandlersV1) get(c *gin.Context) {
 			mlog.Warn(ctx, "high level restful api, fail to deal with get result", mlog.Any("response", response), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+				HTTPReturnMessage: resultErrMessage(err),
 			})
 		} else {
 			HTTPReturnStream(c, http.StatusOK, gin.H{HTTPReturnCode: http.StatusOK, HTTPReturnData: outputData})
@@ -1034,7 +1034,7 @@ func (h *HandlersV1) search(c *gin.Context) {
 				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: resultErrMessage(err),
 				})
 			} else {
 				HTTPReturnStream(c, http.StatusOK, gin.H{HTTPReturnCode: http.StatusOK, HTTPReturnData: outputData})
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -1428,7 +1428,7 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 			mlog.Warn(ctx, "high level restful api, fail to deal with query result", mlog.Any("response", resp), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+				HTTPReturnMessage: resultErrMessage(err),
 			})
 		} else {
 			scannedRemoteBytes, scannedTotalBytes, cacheHitRatio, isValid := proxy.GetStorageCost(queryResp.GetStatus())
@@ -1496,7 +1496,7 @@ func (h *HandlersV2) get(ctx context.Context, c *gin.Context, anyReq any, dbName
 			mlog.Warn(ctx, "high level restful api, fail to deal with get result", mlog.Any("response", resp), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+				HTTPReturnMessage: resultErrMessage(err),
 			})
 		} else {
 			scannedRemoteBytes, scannedTotalBytes, cacheHitRatio, isValid := proxy.GetStorageCost(queryResp.GetStatus())
@@ -2007,7 +2007,7 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 				mlog.Warn(ctx, "high level restful api, fail to deal with search aggregation result", mlog.Any("result", searchResp.Results), mlog.Err(err))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: resultErrMessage(err),
 				})
 			} else {
 				respBody := gin.H{
@@ -2035,7 +2035,7 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: resultErrMessage(err),
 				})
 			} else {
 				if len(searchResp.Results.Recalls) > 0 {
@@ -2207,7 +2207,7 @@ func (h *HandlersV2) advancedSearch(ctx context.Context, c *gin.Context, anyReq 
 				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: resultErrMessage(err),
 				})
 			} else {
 				if proxy.Params.QueryNodeCfg.StorageUsageTrackingEnabled.GetAsBool() && isValid {
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -1480,7 +1480,7 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 			mlog.Warn(ctx, "high level restful api, fail to deal with query result", mlog.Any("response", resp), mlog.Err(err))
 			HTTPReturn(c, http.StatusOK, gin.H{
 				HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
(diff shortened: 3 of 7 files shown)


czs007:

This PR's core array-unwrapping fix is sound; the findings below are all in the surrounding error-message redaction and compatibility-switch work.

## Medium

**`internal/distributed/proxy/httpserver/utils.go:3975` — dynamic-field JSON failures now report `$meta` and lose the only actionable detail.**
The dynamic-field branch of `case schemapb.DataType_JSON` now returns `wrapOutputFieldErr(fieldData.GetFieldName(), err)`, but that branch is entered only when `GetIsDynamic() == true`, so the name is always the internal `$meta` (`pkg/common/common.go:70`, created at `internal/rootcoord/create_collection_task.go:346`) — never a name from the caller's `outputFields`. This contradicts the contract this PR states at `utils.go:3695-3697` ("The name comes from the caller's own outputFields, so echoing it back tells them which column to drop or report"). Failure: a dynamic-field collection holds a legacy row whose `$meta` bytes are not a single JSON document (the case this PR itself documents at `utils.go:4030-4035`) and the caller queries `/v2/vectordb/entities/query` → HTTP 200 with code 1805 and message `fail to parse search result, error: failed to serialize output field $meta`, pointing at a field the caller never requested while the actionable cause (`dynamic field does not hold a single JSON document`, or the decoder error) survives only in the server log at `utils.go:3973`, which a hosted-deployment caller cannot read.
The same reduction hits the aggregation path: every error in `buildSearchAggregationResp` (`utils.go:4060-4104`) and `buildAggBucketResp` is a bare `merr.WrapErrServiceInternalMsg` with neither `asSafeMessage` nor `wrapOutputFieldErr`, so `resultErrMessage` falls through to `return base` — a server-side reduce contract violation such as `agg_topks sum %d does not match bucket count %d` (`utils.go:4086`) reaches the user as just `fail to parse search result`.
Suggestion: mark the `utils.go:4030`-area errors with `asSafeMessage` (their text describes stored data, not internal layout) or attribute them to the dynamic field names the caller actually requested; for the aggregation path, preserve at least a distinguishing category in the client-visible message. *(raised by tinswzy)*

**`internal/distributed/proxy/httpserver/utils.go:2545` — the Int64-to-string change for `ArrayOfStruct` sub-fields is not covered by `legacyArrayResponse`.**
`scalarArrayToInterfaces` gained an `enableInt64` parameter, and its `LongData` branch now emits `strconv.FormatInt(v, 10)` instead of the raw `int64` it always emitted before. That function is the struct-array sub-field value path (`utils.go:2433`), which already returned native values before this PR, so this is a pure JSON type flip: `{"my_struct":[{"score":1}]}` becomes `{"my_struct":[{"score":"1"}]}`. Failure: an existing REST client that sends no `Accept-Type-Allow-Int64` header (absent header parses to `false` at `handler_v2.go:1477`) queries a struct-array field with an Int64 array sub-field → elements arrive as JSON strings, breaking strongly-typed deserialization, numeric comparison, and summation on the client, and setting `proxy.http.legacyArrayResponse=true` does not restore the old shape. The flag's only read is `utils.go:3839` and its only use is `utils.go:3907` inside `case schemapb.DataType_Array`; it is never threaded into `structArrayRowAccessor.row` / `scalarArrayToInterfaces` (`utils.go:3997` passes only `enableInt64`).
Suggestion: either thread `legacyArrayResponse` down through `row()` into `scalarArrayToInterfaces`, or state explicitly in the param `Doc` (`pkg/util/paramtable/http_param.go:119-124`) and the `configs/milvus.yaml` comment that the switch covers top-level `Array` fields only and that the struct-array Int64 change cannot be rolled back — as written, the Doc describes only the `ScalarField` wrapper shape and reads as if it covered all of this PR's Array response changes. *(raised by tinswzy)*

## Nitpick

**`internal/distributed/proxy/httpserver/utils_test.go:2204` — the legacy-shape test marshals with a different encoder than the one serving responses.**
`TestBuildQueryRespLegacyArrayResponse` builds its expected payload with `gojson.Marshal` (`encoding/json`), while responses are rendered by `json_render.go:33` via `internal/json` → `sonic.ConfigStd` (`internal/json/sonic.go:26`). The flag exists purely for compatibility with the pre-fix output, and that output came from sonic marshalling a `*schemapb.ScalarField` (an interface-typed exported oneof), so asserting through `encoding/json` does not verify what the server emits with the flag on. Every sibling test added in this file (`utils_test.go:2392/2417/2424/2445/2598/2694`) uses the internal `json.Marshal`; only the compat-focused one opts out. Separately, the comment at `utils_test.go:2216` says "byte for byte" but the assertion is `assert.JSONEq`, which parses both sides and does not compare bytes. Note the two encoders were checked against this value and currently agree, so there is no behavioral divergence today — this is test fidelity only.
Suggestion: use the internal `json.Marshal`, drop the `gojson` import, and remove "byte for byte" from the comment. *(raised by tinswzy)*

czs007:

This PR's array-unwrapping fix is correct in its core path; the findings below concern client-visible shape changes when `Accept-Type-Allow-Int64` is `false`, plus one pre-existing gap on the aggregation response path.

## Medium

**`internal/distributed/proxy/httpserver/utils.go:1268` — Int64 array elements are serialized as strings, but the insert path cannot parse quoted integers.**
When `Accept-Type-Allow-Int64` is `false` (header explicitly set, or `proxy.http.acceptTypeAllowInt64=false`), the newly added `scalarFieldToRESTAny` (`utils.go:3752-3756`) renders `Array(Int64)` elements via `formatInt64`, i.e. as strings. The insert path at `utils.go:1268-1270` unmarshals the raw request text straight into `[]int64` for `ElementType=Int64`, so `["9007199254740993"]` is rejected — unlike the top-level scalar `Int64` case at `utils.go:1177-1179`, which goes through `json.Number(dataString).Int64()` and accepts the quoted form. Read-back → write-back does not round-trip. Failure: a client sends `Accept-Type-Allow-Int64: false` (or the deployment sets `proxy.http.acceptTypeAllowInt64=false`), queries a collection with an `Array(Int64)` field via `/v2/vectordb/entities/query`, gets `{"ids":["9007199254740993"]}`, and POSTs the same row back to `/v2/vectordb/entities/insert` → the insert is rejected with `ParameterInvalid`, breaking read-modify-write flows entirely under that header.
Suggestion: accept quoted integers in the `ElementType=Int64` branch of the insert path, mirroring the top-level scalar handling (`json.Number(...).Int64()` fallback), so the response shape this PR produces is accepted back on write. Note this is invisible in CI because `tests/restful_client_v2/api/milvus.py:208` pins the header to `true` on every request; the default path is unaffected since `pkg/util/paramtable/http_param.go:69-77` defaults to `true` and `utils.go:4556-4561` fills in the header when absent. *(raised by czs007, xaxys)*

## Low

**`internal/distributed/proxy/httpserver/utils.go:3907` — the `legacyArrayResponse` compatibility switch does not cover the `ArrayOfStruct` Int64 shape change.**
The legacy branch at `utils.go:3905-3911` short-circuits only `case schemapb.DataType_Array`; the `ArrayOfStruct` branch at `utils.go:3996-3997` calls `structArrayAccessors[j].row(int(dataIdx), enableInt64)` unconditionally, and this PR's `scalarArrayToInterfaces` (`utils.go:2540-2546`) now emits `strconv.FormatInt` when `enableInt64` is false. `ArrayOfStruct` never had the protobuf-wrapper bug, so this is a pure output-shape change with no escape hatch. Failure: a deployment sets `proxy.http.acceptTypeAllowInt64=false` (or a client sends `Accept-Type-Allow-Int64: false`) and queries a collection whose struct-array sub-field is `Array(Int64)` → elements change from `9007199254740993` to `"9007199254740993"`, numeric-parsing clients fail with a type error, and setting `proxy.http.legacyArrayResponse=true` does not restore the old shape because that switch only applies to `DataType_Array`.
The new semantics (pushing `Accept-Type-Allow-Int64` down into nested Int64) are reasonable; the issue is only that the compatibility switch's coverage does not match the shapes the PR actually changes. Suggestion: either make `legacyArrayResponse` also short-circuit `ArrayOfStruct`, or state explicitly in its `Doc` (`pkg/util/paramtable/http_param.go:117-125`) and in `configs/milvus.yaml` that it does not cover that shape. *(raised by czs007, xaxys)*

- `pkg/util/typeutil/schema.go:3619` — Array/JSON output fields are silently rendered as `null` in aggregation `top_hits`. `getData` has no `DataType_Array`/`JSON`/`Geometry`/`ArrayOfStruct` case and falls through to `return nil`; `GetDataIterator` (`schema.go:3552-3555`) forwards that, `readValueByFieldID` (`internal/proxy/search_agg/computer.go:392-395`) treats `nil` as `isNull`, `buildHitResult` (`computer.go:292-296`) drops the `isNull` flag and writes `hit.Fields[fieldID]=nil`, `serializeAggHitFields` (`internal/proxy/search_pipeline.go:1062-1065`) leaves the oneof unset, and `aggHitFieldValueToRESTAny` (`utils.go:4268-4269`) returns `nil` → JSON `null`. Failure: a search request with `search_aggregation` + `top_hits` and `outputFields` containing an `ARRAY(VARCHAR)` field `tags` → `buckets[i].hits[k].tags` is always `null` even though the row has values, with no error code or log, and inconsistent with the same field on the non-aggregation query path. Nothing rejects such an output field — `internal/proxy/task_search.go:432-434` copies every `OutputFieldsId` into `UserOutputFieldIDs` unfiltered. This is unrelated to this PR because the diff does not touch `pkg/util/typeutil/schema.go`, `internal/proxy/search_agg/computer.go`, `internal/proxy/search_pipeline.go` or `internal/proxy/task_search.go`, and the `utils.go` hunks stop at `buildSearchAggregationResp` without reaching `buildAggHitsResp`/`aggHitFieldValueToRESTAny`. Suggested follow-up: add the missing cases to `getData` plus an `AggHitField` array variant, or explicitly reject non-scalar output fields in `initSearchAggregation`. No tracking issue yet. *(raised by czs007)*

czs007:

## Overview

Adversarial review found four issues in the new REST array/JSON conversion paths: two correctness defects that let non-finite floats reach storage through the newly added lenient array fallback, and two behavior/compatibility problems where a response-shape flag changes response or write semantics contrary to its documented contract.

## High

**`internal/distributed/proxy/httpserver/utils.go:1908` (and `:1928`) — quoted Float/Double array elements accept NaN/Inf**
The new quoted-string branches parse with `cast.ToFloat32E` / `cast.ToFloat64E` and never check `math.IsNaN` / `math.IsInf`; `cast.ToFloat32E("NaN")`, `("Inf")`, `("infinity")` all return non-finite values with `err == nil`. Failure: a REST insert of `{"arr": ["NaN"]}` (or `["Inf"]`, `["-Infinity"]`) into an `Array(Float)`/`Array(Double)` field → the value is persisted, and any later query/search hitting that row fails to encode (`json: unsupported value: NaN or ±Infinite`) after gin has already written `200`, so the client gets HTTP 200 with an empty body and the row is permanently unreadable over REST, silently. Before this PR the payload was rejected at `json.Unmarshal`; it now reaches the new `checkAndSetData` fallback (`utils.go:1212-1243`), and downstream `checkArrayElement` (`internal/proxy/validate_util.go:1108-1129`) only does oneof shape checks — no `typeutil.VerifyFloats32/64`, unlike `checkFloatFieldData` (`validate_util.go:1040`) for plain columns. The adjacent Number branch at `utils.go:1904` does reject `Inf`, so the function is not self-consistent internally. Suggestion: after parsing the string, apply `math.IsNaN`/`math.IsInf` plus the float32 range check used by the Number branch, and add insert→query regression cases for `"NaN"`, `"Inf"`, `"-Inf"`. The same code is reused by `buildStructSubArrayScalar` (`utils.go:1957`), so struct sub-arrays are covered by the same fix. *(raised by chyezh)*

**`internal/distributed/proxy/httpserver/utils.go:1926` — out-of-range Double literals silently become ±Inf**
The Double branch appends `v.Float()` for `gjson.Number` with no range or finiteness check, and `gjson`'s `Float()` swallows `strconv.ParseFloat`'s `ErrRange`. Failure: a REST insert of `{"arr": [1e999]}` into an `Array(Double)` field → strict `json.Unmarshal` rejects it, the new fallback re-parses it, `+Inf` is written and persisted, and subsequent query/search on that row fails sonic encoding and returns HTTP 200 with an empty body — the row becomes permanently unreadable over REST. The PR note that "only a payload the decode already rejected reaches here" holds, but the converse is the problem: payloads that were previously rejected are now accepted. The Float branch at `utils.go:1903-1906` already uses `strconv.ParseFloat(v.Raw, 64)` with an `IsInf` guard; the Double branch lacks the equivalent. Suggestion: parse with `strconv.ParseFloat(v.Raw, 64)` and error on `err != nil || math.IsInf(parsed, 0)`. (The same missing check in `buildStructSubArrayScalar`'s Double path predates this PR; what is new is the top-level `Array` column entry into it.) *(surfaced during verification)*

## Medium

**`internal/distributed/proxy/httpserver/utils.go:3990` — `legacyArrayResponse` bypasses `scalarFieldToRESTAny` and ignores `enableInt64`**
When `legacyArrayResponse` is true, `buildQueryResp` assigns the raw `*schemapb.ScalarField` into the row and `continue`s, skipping `scalarFieldToRESTAny` entirely, so `enableInt64` is never applied. Failure: a deployment with `proxy.http.legacyArrayResponse=true` querying an `Array(Int64)` field containing `9007199254740993` without `Accept-Type-Allow-Int64: true` → the response is `{"tags":{"Data":{"LongData":{"data":[9007199254740993]}}}}`, and `JSON.parse` in JavaScript yields `9007199254740992` — silent precision loss, in exactly the client population the legacy flag exists for. This contradicts the doc added in the same PR at `pkg/util/paramtable/http_param.go:122-127`, which states the flag "covers the shape of a top-level Array field and nothing else: … and so is Accept-Type-Allow-Int64, which renders an Int64 as a string wherever one appears, including inside either kind of array." Suggestion: build the legacy wrapper explicitly and format `LongData` per `enableInt64` (reusing `formatInt64`, `utils.go:4379`), or amend the doc to state that legacy mode also reverts Int64 rendering. Cover `legacyArrayResponse=true` × both `Accept-Type-Allow-Int64` values with a value beyond 2^53. *(raised by chyezh)*

**`pkg/util/paramtable/http_param.go:106` — a response-format flag now also gates write-path JSON unwrapping, and its default was flipped**
`nativeJSONResponse` was introduced on master only (d9bef57b8f, 2026-08-10) and is not an ancestor of `v3.0.0` or `milvus/3.0`, so the relevant baseline is `v3.0.0`, whose `checkAndSetData` unconditionally unwraps JSON strings (`reallyData[fieldName] = []byte(dataString)`, with `dataString = fieldValue.String()` stripping the quotes). This PR does two things: gates that unwrap on `!nativeJSONResponse` (`internal/distributed/proxy/httpserver/utils.go:1272`) and flips the default from `false` to `true`. Failure: upgrading from `v3.0.0` with default configuration and sending the unchanged REST insert `{"data":[{"id":1,"meta":"{\"a\":1}"}]}` → `document = fieldValue.Raw`, so the literal string `"{\"a\":1}"` is stored instead of the object `{"a":1}`, the insert reports success with no warning, and `meta["a"] == 1` no longer matches that row. Because the flag is refreshable, toggling it at runtime changes write semantics mid-flight and leaves a single collection holding both shapes; setting it back to `false` only affects subsequent writes. Suggestion: decouple the write-side unwrap from the response-side format (a separate flag, or fix the write semantics outright), and call this out in the release notes / migration guide as a stored-data change rather than a response-format change. Also worth cross-checking the `Version` field, which still reads `"2.7.0"` while the newly added `legacyArrayResponse` uses `"3.0.0"` and `v3.0.0` has shipped. *(raised by xiaocai2333)*

czs007:

This PR's lenient array-element fallback accepts quoted `NaN`/`Inf` values for `Array(Float)` fields, which is a regression from the previous strict decode path.

## High

**`internal/distributed/proxy/httpserver/utils.go:1908`** — The Float `case v.Type == gjson.String` branch in `parseScalarArrayElements` calls `cast.ToFloat32E(v.String())` and only checks `err != nil`, with no `math.IsNaN` / `math.IsInf` guard. With cast v1.3.1 (`go.mod:32`), `ToFloat32E` on a string delegates to `strconv.ParseFloat(s, 32)` (`caste.go:168-173`), which returns `(NaN, nil)` for `"NaN"` and `(±Inf, nil)` for `"Inf"`/`"+Inf"`/`"Infinity"` — case-insensitively — so those values pass straight through. Failure: a default-config `POST /v2/vectordb/entities/insert` whose `DataType_Array`/`ElementType_Float` field is `["NaN"]` (or `["Inf"]`, `["Infinity"]`, `["nan"]`, and likewise for a Float sub-field of a struct array) → the insert is accepted and NaN/±Inf is persisted, and any later query/search that selects that field on that row fails mid-encode in `jsonRender` (`json_render.go:32-36`) with `json: unsupported value: NaN` after a 200 status and partial body have already been written, leaving the client with a truncated, syntactically invalid response.

Reachability and the absence of a downstream backstop: the top-level `Array(Float)` path first tries `unmarshalScalarArray` (`utils.go:1212`), whose `json.Unmarshal` into `[]float32` rejects the string element, then falls back to `parseScalarArrayElements` (`utils.go:1235`) and hits this branch. Array fields are validated by `checkArrayFieldData` → `checkArrayElement`, whose Float case (`validate_util.go:1108-1118`) only compares the reflect type and never calls `VerifyFloats32` — unlike a plain Float column, which does go through `VerifyFloats32` (`validate_util.go:1039-1041`, `pkg/util/typeutil/float_util.go:51-58`). The in-code comment "cast.ToFloat32E is what a plain Float column falls back to" therefore holds at the parse layer but not at the validation layer. `parseScalarArrayElements` is also reused by `buildStructSubArrayScalar` (`utils.go:1965-1966`), so struct-array Float sub-fields share the gap.

Suggestion: mirror the Double branch, which is already correct at `utils.go:1936-1941`:

```go
parsed, err := cast.ToFloat32E(v.String())
if err != nil || math.IsInf(float64(parsed), 0) || math.IsNaN(float64(parsed)) {
    return nil, elementErr(idx, v, "expect float in the float32 range")
}
```

*(raised by xiaocai2333, bigsheeper)*

xiaofan-luan:
Thanks — all three points addressed, plus one follow-up filed. Head is now `8c5304ecf1`.

**[P1] `legacyArrayResponse` bypassing Int64 safety — fixed as suggested (option A).**
The legacy branch now goes through `legacyArrayValue(row, enableInt64)`. When the caller allows native Int64, or the row holds no Int64 at all, the `*schemapb.ScalarField` is handed to the encoder exactly as before, byte for byte. Otherwise the wrapper is spelled out with the value formatted per the header: `{"Data":{"LongData":{"data":["9007199254740993","1"]}}}`. Nested arrays recurse, and `element_type` follows the same `omitempty` rule the message does, so nothing but the digits changes. The reasoning is that the switch restores a *shape* — what those clients parse — while `Accept-Type-Allow-Int64` is about what their JSON parser can hold, which the shape does not change. `TestLegacyArrayResponseKeepsInt64Safe` covers `legacyArrayResponse=true` against both header values with `9007199254740993`, and asserts the untouched path for a row with no Int64 in it.

**[P2] Lenient Double accepting infinities — fixed, and generalized.**
Confirmed: `gjson.Result.Float()` reads `1e400` as `+Inf` while `json.Unmarshal` into `[]float64` rejects it, so the fallback was accepting a value the strict path refused. All four float element branches (Float/Double × number/string) now go through `typeutil.VerifyFloat`, the check a plain column of the same type reaches — rather than a hand-written `IsNaN || IsInf`, so the rule keeps one definition. For Float it is applied *after* the cast to float32, where a magnitude past that range has already become an infinity, which also retires a separate `math.MaxFloat32` comparison.

A related hole surfaced while doing this and is fixed in the same change: `cast.ToFloat32E` returns `(NaN, nil)` for `"NaN"` and `(±Inf, nil)` for `"Inf"`/`"Infinity"` — case-insensitively — and an array element has no later backstop, since `checkArrayElement` compares the element's type and never calls `VerifyFloats32` the way a plain column does. One of those stored is a row that cannot be served: the encoder fails on it after the status and part of the body are already written. The rejection cases now cover `NaN`, `nan`, `Inf`, `+Inf`, `-inf` and `Infinity` for both Float and Double, on the top-level column and the struct sub-field path.

**Version fields** — corrected. `compatibilityMode`, `maxExprParamsDepth`, `nativeJSONResponse` and `legacyArrayResponse` all claimed releases that shipped without them; they now read `3.0.1`, the milestone this PR targets.

**[P1] `nativeJSONResponse` default — keeping `true` in this PR, deliberately.**

The part of the objection I agree with: an upgrading client does not read the release notes for us, and a *write*-side change is the dangerous kind, because it surfaces weeks later rather than on the first request. That is real and it is why this needs a prominent release-note entry, which I will make sure it gets.

Where I land differently:

- The setting has never shipped. It exists only on master (`d9bef57b8f`, 2026-08-11) and is on neither `3.0` nor `2.6`, so this is not a change to a released default — it is the choice of which default this feature is *born* with. That choice is only available once.
- A default that is wrong is not made safe by being opt-in; it is made permanent. The shape being fixed is a real inconsistency — the same value reads back as text in a declared JSON field and as a document in the dynamic field — and behind an opt-in almost nobody will turn it on, so both shapes live forever and every client keeps paying for the difference.
- The rollback is one switch and it covers both directions. That is the point of coupling the write-side unwrap to it, rather than a second flag: while the field reads back as the text of its document, that text is the field's wire form, and the insert path reading a document out of it is decoding, not guessing. Once the field reads back as the document itself, a string is a string — unwrapping it would store a value the caller never sent and leave no way to store `"123"` as the text it is. Splitting the two into separate flags would make the four combinations expressible, and two of them are incoherent (a deployment that returns documents but re-reads strings as documents cannot round-trip its own responses).
- On refreshability: a mixed collection is possible, and that is inherent to any refreshable write-path setting, `compatibilityMode` included. Worth documenting; I do not think it argues for a separate flag, which would make the mixture more likely rather than less.

That said, this is a protocol decision rather than a code one, and if the maintainers would rather ship the flip on its own I will pull it out of this PR into a standalone one and leave the dialect coupling here — the coupling is what makes `nativeJSONResponse=false` a complete rollback, and it is a no-op while the default stays `false`. Say the word and I will split it.

**Follow-up filed:** #52551 — `topHits` silently returns `null` for Array and JSON output fields. Pre-existing (this diff does not touch `typeutil/schema.go`, `search_agg/computer.go`, `search_pipeline.go` or `task_search.go`); `AggHitField` has no array variant, so a proper fix needs a proto change.

czs007:

**Verified:**
- `internal/distributed/proxy/httpserver/utils.go:1913-1926` — the Float quoted-string branch now runs `typeutil.VerifyFloat` after `cast.ToFloat32E`, so `"NaN"`, `"Inf"`, `"+Inf"`, `"Infinity"` (which `strconv.ParseFloat` accepts) are rejected.
- `pkg/util/typeutil/float_util.go:51-58` — `VerifyFloat` explicitly returns an error on `math.IsNaN` and `math.IsInf(±1)`, which is exactly the check that was missing before.
- `internal/distributed/proxy/httpserver/utils.go:1902-1912` — the Float number branch casts to `float32` first and then calls `VerifyFloat`, so an overflow to ±Inf is rejected rather than stored.
- `internal/distributed/proxy/httpserver/utils.go:1940-1953` — the Double number and string branches were moved from hand-written `IsInf`/`IsNaN` to the same `VerifyFloat` call, so float and double now share one rule.
- `internal/distributed/proxy/httpserver/utils.go:1212, 1235, 1978-1979` — the reachable path was re-traced end to end: a top-level `Array(Float)` insert of `["NaN"]` fails strict `unmarshalScalarArray`, falls back to the lenient `parseScalarArrayElements`, and is rejected there; the struct sub-field path calls the same function, so it is covered too.
- `internal/distributed/proxy/httpserver/utils_test.go:1798-1805` — regression cases cover `["NaN"] ["nan"] ["Inf"] ["+Inf"] ["-inf"] ["Infinity"]` for Float and `["NaN"] ["Inf"] ["-Infinity"]` for Double, including the case variants.

One caveat on the depth of that verification: the local tree had no compiled C++ core (`pkg-config` could not find `milvus_core`), so `go test ./internal/distributed/proxy/httpserver/` could not be run — the conclusions above come from static review and path tracing, not a local green test run.

## Low — non-blocking, no change required in this PR

- `internal/distributed/proxy/httpserver/utils.go:1902` — Range checking for Float array elements moved from "check the float64 magnitude before conversion" to "check for Inf after conversion", which is a relaxation relative to this PR's base: the base `buildStructSubArrayScalar` carried `math.Abs(parsed) > math.MaxFloat32` at this position (base `utils.go:1877`), and this PR removes it in favor of `typeutil.VerifyFloat`, which only rejects NaN/±Inf. Failure: writing the literal `3.4028235e38` into a Float element of a struct array sub-field (or a top-level Array column whose strict decode fails and falls back to the lenient branch, e.g. `[3.4028235e38, "1"]`) → the request was rejected before this PR and is now silently accepted and stored as `3.4028234663852886e38`, while the same literal in a plain Float column is still rejected at `utils.go:1306` ("value is outside the float32 range"), so a client migrating a value between a scalar column and an array element sees inconsistent behavior. The affected interval is narrow: `3.4028236e38` already casts to `+Inf` and is rejected, so no Inf leaks through and the stored value is always a finite float32; the actual boundary is the midpoint between `MaxFloat32` and `2^128`, `3.4028235677973366e38` (the midpoint itself rounds up to Inf and is rejected). Suggestion: if you want the two paths aligned, move the plain-column pre-conversion magnitude check into the `gjson.Number` branch as well. This is a consistency nit and does not block merging.
  (raised by bigsheeper)

czs007 (approved):
/lgtm

Issue #52050, opened 2026-07-30
[Bug]:  Int8/Int16/Int32 typed fields silently accept out-of-range values with two's complement wrapping


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
Milvus version: v2.6.21 (also reproduced on v2.6.20; v2.6.21 re-verified on 2026-07-24)
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq (standalone default)
- SDK version(e.g. pymilvus v2.0.0rc2): REST API v2 (RESTful HTTP via curl; no SDK)
- OS(Ubuntu or CentOS): macOS host / Linux container (Docker Desktop)
- CPU/Memory: 16 GB
- GPU: N/A
- Others: FLAT index, COSINE metric
```

### Current Behavior

When inserting a value that exceeds the valid range of an Int8, Int16, or Int32 typed field, Milvus **silently accepts the insert** (returns `code=0`) and **stores a corrupted value** using two's complement truncation. No error or warning is returned.

```
Int8 field (valid range: -128 to 127):
  Insert 128  → stored as -128  (no error!)
  Insert -129 → stored as 127   (no error!)
  Insert 256  → stored as 0     (no error!)
  Insert 1000 → stored as -24   (no error!)

Int16 field (valid range: -32768 to 32767):
  Insert 32768  → stored as -32768  (no error!)
  Insert 65535  → stored as -1      (no error!)
  Insert 100000 → stored as -31072  (no error!)

Int32 field (valid range: -2147483648 to 2147483647):
  Insert 2147483648  → stored as -2147483648  (no error!)
  Insert 4294967295  → stored as -1           (no error!)
```

Additionally, the filter literal is NOT wrapped, so `insert(i8=128)` followed by `query(filter="i8 == 128")` returns **zero results** — the data appears to have vanished, and only `i8 == -128` finds it.

### Expected Behavior

Reject out-of-range values with a clear error, matching PostgreSQL, MySQL (strict mode), and MongoDB, which reject values that don't fit the declared type rather than silently truncating:
```json
{"code": 1100, "message": "value 128 out of range for Int8 field 'i8' (valid range: -128 to 127)"}
```

### Steps To Reproduce

```markdown
# Create collection with Int8/Int16/Int32 fields
curl -s -X POST http://localhost:19530/v2/vectordb/collections/create \
[...]

PR #52097, opened 2026-07-31
fix: reject out-of-range narrow integers in REST

xiaofan-luan:
## What changed

- Parse REST scalar `Int8`, `Int16`, and `Int32` values with the target bit width before narrowing.
- Parse scalar `Int64` values from the raw JSON literal instead of gjson's
  `String()` rendering, so decimal and exponent forms keep full precision.
- Validate StructArray `Array<Int8/Int16/Int32/Int64>` sub-fields before converting them to protobuf integer arrays.
- Preserve exact integer-valued decimal and exponent forms in StructArray input while rejecting fractions and out-of-range values.
- Return errors that include the field/sub-field, offending value, and valid range.
- Add unit and endpoint coverage for REST v1/v2 insert and upsert.

## Root cause

The scalar REST adapter used `cast.ToInt8E`, `cast.ToInt16E`, and `cast.ToInt32E`. Those helpers parsed into a machine-sized integer and then narrowed without checking the destination range, so values such as `128` became `int8(-128)` before Proxy validation could see the original value.

The StructArray adapter had the same ordering problem: it called `int32(v.Int())` or `v.Int()` before validation. That allowed wraparound inputs such as `4294967296`, Int32 overflow, Int64 overflow, and fractional values to be silently changed.

The scalar `Int64` adapter had the same origin. It read `fieldValue.String()`,
which gjson renders through the parsed `float64` as soon as the raw text is not
all digits, so the value was already wrong before `json.Number` saw it:

|---|---|---|
| `{"count": 9007199254740993.0}` | `9007199254740992` (no error) | `9007199254740993` |
| `{"count": 1e-400}` | `0` (no error) | rejected |

Only the JSON-number form is rerouted through `parseJSONInteger`. Quoted
integers keep `json.Number`'s base-10 parsing: this path also carries Int64
primary keys, and strconv's base detection would silently reinterpret a
zero-padded id such as `"010"` as octal.

## Compatibility

- Valid min/max boundaries remain accepted.
- Scalar integer values remain accepted as either JSON numbers or quoted integer strings.
- StructArray and scalar integer-valued decimal/exponent forms remain accepted (`1.0`, `1e3`, `100e-2`), and Int64 values are parsed exactly rather than through float64.
- Two scalar Int64 inputs change: `9007199254740993.0` now keeps its exact value instead of being silently rounded, and `1e-400` is rejected instead of being silently stored as `0`. Quoted integers are unchanged.
- REST v1/v2 insert and upsert intentionally preserve the existing public error code `1804` (`ErrInvalidInsertData`). The underlying validation error remains `ErrParameterInvalid` (`1100`); this PR does not change the wire error-code contract.
- Rejected input does not reach Proxy insert/upsert calls.

## Validation

- `git diff --check` passed.
- The following tests passed with `-tags dynamic,test -gcflags="all=-N -l" -count=1` in a local source snapshot compatible with the available native artifacts:
  - `TestCheckAndSetData`
  - `TestCheckAndSetDataStructArrayRows`
  - `TestNarrowIntegerOverflowV1`
  - `TestNarrowIntegerOverflowV2`
  - `TestStructArrayScalarSubFieldTypes`
  - `TestStructArrayNarrowIntegerSubFieldValidation`
- The exact-head local package run is blocked before tests start because the existing native artifacts do not expose `SegcoreSetFMIndexCostRatio`; CI will build against matching native sources.

issue: #52050

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -432,6 +432,12 @@ proxy:
     debug_mode: false # Whether to enable http server debug mode
     port:  # high-level restful api
     acceptTypeAllowInt64: true # high-level restful api, whether http client can deal with int64
+    # high-level restful api, restore the value handling of releases that predate the REST insert validation work.
+    # When true the server keeps the previous lenient behaviour: a missing or null non-nullable field is stored as an
+    # empty value, out-of-range integers wrap instead of being rejected, numbers reach VarChar and JSON fields through
+    # their float64 rendering, and integers too large for the JSON engine become 0. This is a temporary escape hatch
+    # for clients that have not been corrected yet: every one of those behaviours silently changes what is stored.
+    compatibilityMode: false
     enablePprof: true # Whether to enable pprof middleware on the metrics port
     readHeaderTimeout: 5s # HTTP server timeout for reading request headers
     readTimeout: 0s # HTTP server timeout for reading the entire request, including the body. 0 disables this timeout
--- a/internal/distributed/proxy/httpserver/handler_v1_test.go
+++ b/internal/distributed/proxy/httpserver/handler_v1_test.go
@@ -869,6 +869,35 @@ func TestInsertForDataType(t *testing.T) {
 	}
 }
 
+func TestNarrowIntegerOverflowV1(t *testing.T) {
+	paramtable.Init()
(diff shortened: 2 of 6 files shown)


xiaofan-luan:

Issue #52095, opened 2026-07-31
[Enhancement]: Go client is missing six index types the server supports


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

`client/index` declares 24 index types. Auditing it against the server's own
list (`internal/util/indexparamcheck/conf_adapter_mgr.go` for scalar,
knowhere's `comp/index_param.h` for vector) and against the Java / C++ / Rust
SDKs turns up six types Milvus accepts that the Go client never declared:

| index type | kind | java | cpp | rust | go |
|---|---|---|---|---|---|
| `NGRAM` | scalar | ✅ | ✅ | ✅ | ❌ |
| `FMINDEX` | scalar | ❌ | ❌ | ❌ | ❌ |
| `HNSW_SQ` | vector | ✅ | ✅ | ✅ | ❌ |
| `HNSW_PQ` | vector | ✅ | ✅ | ✅ | ❌ |
| `HNSW_PRQ` | vector | ✅ | ✅ | ✅ | ❌ |
| `AISAQ` | vector | ✅ | ✅ | ✅ | ❌ |

(`FMINDEX` is new — scalar index engine version 5, merged in #51375 — so no SDK
has it yet. The other five have been on the server for a while and simply never
reached this client.)

### Why is this needed?

Nothing is hard-blocked: `IndexType` is a string alias and `NewGenericIndex`
takes a raw param map, so a caller can always pass the literal. But:

- there is no typed constructor, so users hand-build the param map and a
  misspelled key (`efconstruction` for `efConstruction`) fails at the server
  instead of at compile time;
- `DescribeIndex` returns a type string the package has no constant for.

This is a recurring gap rather than a one-off. `RTREE` and `BITMAP` reached all
five clients; `NGRAM` reached three and was dropped here and in the Node SDK.
There is no process keeping SDK index-type lists in step with the server, so
each new type loses a different subset of clients.

### Anything else?

Suggested shape, following the conventions already in the package:

- `NewNgramIndex(minGram, maxGram int)` — positional, because
  `ngram_index_checker` rejects an index that omits either bound, so there is no
  default to fall back to.
- `NewFMIndex()` with fluent `WithSaSampleRate` / `WithBlockBytes` (the
  `JSONPathIndex` shape). Both are optional, so an unset one must be left out of
  the map entirely — emitting a zero fails the server's range check rather than
  selecting its default.
[...]

PR #52098, opened 2026-08-01
feat: add the index types the Go client was missing, and fix the broken ones (#52095)

xiaofan-luan:
issue: #52095
design doc: docs/design-docs/design_docs/20260708-fm_index_scalar_index.md

The design doc covers `FMINDEX`, the one genuinely new index type here; it landed
on master with the server implementation in #51375. The other five types this PR
adds — `NGRAM`, `HNSW_SQ`, `HNSW_PQ`, `HNSW_PRQ`, `AISAQ` — are existing engine
capabilities that the Go client had simply never declared, and the rest of the
change is bug fixes, so neither needs a design of its own.

`client/index` had drifted behind the server. Audited it against the index checker
registry (`internal/util/indexparamcheck/conf_adapter_mgr.go`), the knowhere v3.0.6
config classes — the version Milvus pins — and the Java / C++ / Rust / Node SDKs.

### Missing index types

| | |
|---|---|
| scalar | `NGRAM`, `FMINDEX` |
| vector | `HNSW_SQ`, `HNSW_PQ`, `HNSW_PRQ`, `AISAQ` |

All but `FMINDEX` already exist in the Java, C++ and Rust SDKs; `NGRAM` has been on
master for a while and simply never reached this client, so a Go user had to fall
back to `NewGenericIndex` with a hand-built param map, and `DescribeIndex` returned
a type the package had no constant for.

New constructors follow the shape each family already uses — `NewNgramIndex`,
`NewFMIndex`, `NewHNSWSQ/PQ/PRQIndex` with matching AnnParams, and `NewAISAQIndex`.

`refine` is deliberately confined to the three quantized HNSW variants. Plain HNSW
maps to `FaissHnswFlatConfig`, whose `CheckAndAdjust` rejects the build with
`refine is not supported for this index` if `refine` or `refine_type` is set, so
exposing it there would only produce failures.

### Bugs fixed, all surfaced by the same audit

- `WithRabitQueryBits` emitted `rbq_query_bits`. knowhere's `IvfRaBitQConfig`
  declares the field as `rbq_bits_query`, and `KNOWHERE_CONFIG_DECLARE_FIELD`
  stringifies the field name into the JSON key, so the transposed spelling reached
  no consumer. It appears nowhere else in the repo — the server's own e2e tests pass
  `rbq_bits_query` and the out-of-range error they assert on quotes it. Unknown
  search params are not rejected, so callers setting query bits silently got 0.
- `NewIvfRabitQIndex` set `baseIndex.indexType` to `BinIvfFlat`, so `IndexType()`
  reported `BIN_IVF_FLAT` while `Params()` emitted `IVF_RABITQ`.
- `gpuIVFPQIndex` and `gpuCagra` both emitted `IndexTypeKey: GPUIvfFlat`, so
  `NewGPUIVPPQIndex` and `NewGPUCagraIndex` built `GPU_IVF_FLAT` indexes — the
  latter with CAGRA-only graph params attached to it.
- `cagraGraphDegreeKey` was `` `"graph_degree"` ``; the backquoted literal contains
  the quote characters, so the emitted key matched nothing.
- None of the four GPU builders set `baseIndex.indexType`, so `IndexType()` returned
  `""` for all of them.
- `gpuIVFFlatIndex` / `gpuIVFPQIndex` emitted `nlist` / `m` / `nbits` as `"0"`,
  because the constructors never accepted them. Zero is below every accepted range,
  so the build was rejected outright; they are now omitted so the server applies its
  default, and remain reachable through `WithExtraIndexParams`.

### Compatibility

Existing constructors are untouched — same parameters, same `Index` return type — so
callers that take them as function values still compile. Verified by diffing every
`New*` signature against the base commit, and `go vet` passes on `tests/go_client`,
which exercises these constructors throughout.

Params the typed constructors do not model are reachable through
`WithExtraIndexParams`, which merges raw params into any `Index`. `index_type` and
`metric_type` are reserved there: overriding them would put `Params()` and
`IndexType()` into disagreement. This matches how pymilvus behaves — it forwards
whatever param dict it is given (its `valid_index_params_keys` frozenset is dead
code, referenced nowhere and still listing `n_trees` and `PQM` from the Annoy era).

Also adds the search params that only needed a method on an existing type:
`hsnwAnnParam.WithSeedEf`, `diskANNParam.WithBeamwidth`, and `sparseAnnParam`'s
`WithSearchAlgo` / `WithRefineFactor` / `WithDimMaxScoreRatio`.

Opt-in params distinguish "never set" from "set to something invalid": a value the
caller supplied is forwarded verbatim so the server's range check reports the
mistake, rather than being mistaken for an omission and silently replaced by the
default.

### Known follow-up, not addressed here

`ivfRabitQAnnParam.WithRefineK` takes an `int`, but knowhere types `refine_k` as
`CFG_FLOAT` and passes it through as `faiss::IndexRefineSearchParameters::k_factor`
— a multiple of top-k, so `1.5` is meaningful and currently inexpressible. The new
`hnswQuantAnnParam.WithRefineK` takes a `float64`; changing the existing one would
break source compatibility, so it is left for a maintainer decision.

### Not added

`SCANN_DVR`, `IVF_FLAT_CC`, `IVF_SQ_CC`, `IVF_RABITQ_FASTSCAN`, `SVS_*`,
`SPARSE_*_CC` and the `GPU_FAISS_*` / `GPU_CUVS_*` families. They exist in knowhere
but no SDK exposes them; they are internal or experimental.

Also left alone: the stale `IVF_HNSW` constant. The server no longer supports it,
but removing it would break anyone still referencing it.

### Testing

`go build`, `go vet` and `go test ./index/...` pass on the client module; `go vet`
passes on `tests/go_client`.

### DIFF
--- a/client/index/common.go
+++ b/client/index/common.go
@@ -47,9 +47,13 @@ const (
 	IvfSQ8     IndexType = "IVF_SQ8"
 	IvfRabitQ  IndexType = "IVF_RABITQ"
 	HNSW       IndexType = "HNSW"
+	HNSWSQ     IndexType = "HNSW_SQ"
+	HNSWPQ     IndexType = "HNSW_PQ"
+	HNSWPRQ    IndexType = "HNSW_PRQ"
 	IvfHNSW    IndexType = "IVF_HNSW"
 	AUTOINDEX  IndexType = "AUTOINDEX"
 	DISKANN    IndexType = "DISKANN"
+	AISAQ      IndexType = "AISAQ"
 	SCANN      IndexType = "SCANN"
 	MinHashLSH IndexType = "MINHASH_LSH"
 
@@ -67,5 +71,7 @@ const (
 	Sorted   IndexType = "STL_SORT"
 	Inverted IndexType = "INVERTED"
 	BITMAP   IndexType = "BITMAP"
+	NGRAM    IndexType = "NGRAM"
+	FMINDEX  IndexType = "FMINDEX"
 	RTREE    IndexType = "RTREE"
 )
--- a/client/index/disk_ann.go
+++ b/client/index/disk_ann.go
@@ -16,6 +16,8 @@
 
 package index
 
+import "strconv"
+
 const (
 	diskANNSearchListKey = `search_list`
 )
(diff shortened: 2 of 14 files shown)


xiaofan-luan:
Pushed a fix for the `ci-v2/build-ut-cov` failure. It was Code-Check, not the unit
tests: `misspell` flagged two comments using the British spelling `honoured`
(`client/index/disk_ann.go:148`, `client/index/index_test.go:55`). Fixed to
`honored`, and `golangci-lint v2.11.3` with the repo's own `client/.golangci.yml`
now reports 0 issues on the client module.

The `ci-v2/e2e-default` failure is unrelated to this PR:

```
FAILED milvus_client/test_milvus_client_struct_array_nullable.py::
  TestMilvusClientStructArraySchemaEvolution::
  test_struct_array_compaction_preserves_null_empty_offsets_and_indexes
  - assert -1 > 0
= 1 failed, 6137 passed, 340 skipped, 1 xfailed in 4872.93s
```

`compact()` returning -1 in that struct-array test is the failure described in
\#51686, which was closed on 2026-07-24 but appears to be recurring: #52103 hit the
same e2e failure in the same window, while #52099 / #52097 / #52093 passed. This PR
changes only `client/` (14 files, all under `client/index`) and adds no behaviour
to the paths that test exercises.

congqixia (approved):
/lgtm

Issue #52053, opened 2026-07-30
[Bug]: REST API v2 does not support autoId=true for Int64 PK (VarChar generates empty string IDs)


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
Milvus version: v2.6.21 (also reproduced on v2.6.20; v2.6.21 re-verified on 2026-07-24)
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq (standalone default)
- SDK version(e.g. pymilvus v2.0.0rc2): REST API v2 (RESTful HTTP via curl; no SDK)
- OS(Ubuntu or CentOS): macOS host / Linux container (Docker Desktop)
- CPU/Memory: 16 GB
- GPU: N/A
- Others: FLAT index, L2 metric
```

### Current Behavior

When a collection is created with `autoId: true` on the primary key field:

**Int64 PK + autoId** — inserting data *without* the PK field (the expected autoId usage) fails with a parse error:
```
code=1804, "strconv.ParseInt: parsing \"\": invalid syntax"
```
The REST layer converts the missing field to an empty string and tries to parse it as Int64 **before** checking the autoId flag. The only workaround is providing an explicit PK value — which defeats the purpose of autoId.

**VarChar PK + autoId** — inserting data without the PK field "succeeds" but generates **empty string** `""` as the ID for every row, silently creating duplicate PKs:
```
Insert 1: insertIds = [""]
Insert 2: insertIds = ["", ""]
```

Actual output from the reproduction below:
```
Int64 autoId (omit pk):   code=1804, "strconv.ParseInt: parsing \"\": invalid syntax"
Int64 autoId (pk=null):   code=1804, "strconv.ParseInt: parsing \"\": invalid syntax"
Int64 autoId (pk=12345):  code=0, insertIds=[12345] — explicit value used, NOT auto
VarChar autoId (omit pk): code=0, insertIds=["", ""]  — empty string for ALL rows
VarChar autoId (3 rows):  code=0, insertIds=["", "", ""] — duplicate empty PKs!
```

### Expected Behavior

**Int64 PK + autoId=true**: Omitting the PK field should trigger auto-generation of a unique ID (as it does via gRPC/pymilvus)
- **VarChar PK + autoId=true**: Should generate a unique string ID (e.g., UUID) — not empty string

### Steps To Reproduce

```markdown
[...]

PR #52099, opened 2026-08-01
fix: validate required REST insert fields

xiaofan-luan:
issue: #52053

## What changed

- Detect missing JSON fields before REST type conversion instead of collapsing them to empty strings.
- Return a parameter-missing error for absent required fields.
- Return a parameter-invalid error when a non-nullable field is explicitly set to `null`.
- Preserve the existing skip behavior for partial updates, nullable/default fields, AutoID primary keys, and function output fields.
- Add regression coverage for missing Int64 and VarChar primary keys, null VarChar primary keys, and the valid VarChar AutoID path.

## Root cause

`gjson.Result.String()` returns an empty string for both a missing field and several explicit empty/null values. The REST conversion path performed this conversion before checking field presence, so a missing required VarChar primary key was accepted as `""`, while numeric fields failed later with a parsing error.

The AutoID reproduction in #52053 places `autoId` on the primary field, while REST v2 defines it at `schema.autoId`. This PR does not change that schema contract; it fixes the required-field validation issue exposed by the report.

## User impact

Malformed inserts now fail immediately with a clear required-field or nullability error instead of silently inserting an empty VarChar primary key.

## Validation

- `git diff --check`
- Verified GJSON presence semantics independently: missing, explicit `null`, and explicit empty string remain distinguishable through `Exists()`.
- Focused HTTP server tests were attempted with the required `dynamic,test` tags and disabled optimizations. The local native build is currently blocked by an initcore/cgo header mismatch (`C.SegcoreSetFMIndexCostRatio`). A `CGO_ENABLED=0` run is not supported because several Milvus packages require native implementations. CI should run the added regression tests in the complete build environment.

## Compatibility switch

Rejecting a missing field turns requests that used to return `code=0` into
400s, and omitting a field a client has no value for is a common shape, so
`proxy.http.allowMissingFields` is added as an escape hatch.

- Default `false`: the request is rejected. This is the fix.
- Set to `true`: the old behaviour is restored, the value is stored empty --
  `""` for VarChar, String, Geometry and Timestamptz, empty bytes for JSON.

It is refreshable, so an operator hitting this on upgrade can flip it without a
restart while clients are corrected.

The switch deliberately does not cover struct array fields: a missing struct
array already failed before this change, inside `parseStructArrayRow`, so there
is no lenient behaviour to fall back to and skipping the field would be new
behaviour rather than old.

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -432,6 +432,12 @@ proxy:
     debug_mode: false # Whether to enable http server debug mode
     port:  # high-level restful api
     acceptTypeAllowInt64: true # high-level restful api, whether http client can deal with int64
+    # high-level restful api, restore the value handling of releases that predate the REST insert validation work.
+    # When true the server keeps the previous lenient behaviour: a missing or null non-nullable field is stored as an
+    # empty value, out-of-range integers wrap instead of being rejected, numbers reach VarChar and JSON fields through
+    # their float64 rendering, and integers too large for the JSON engine become 0. This is a temporary escape hatch
+    # for clients that have not been corrected yet: every one of those behaviours silently changes what is stored.
+    compatibilityMode: false
     enablePprof: true # Whether to enable pprof middleware on the metrics port
     readHeaderTimeout: 5s # HTTP server timeout for reading request headers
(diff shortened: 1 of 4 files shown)


xiaofan-luan:

Issue #52100, opened 2026-08-01
[Enhancement]: Move the first build of the growing interim vector index off the insert path


foxspy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Build the first interim vector index of a growing segment (IVF_FLAT_CC / SCANN_DVR / sparse CC variants) in a background pool instead of on the insert CGO thread. Insert only publishes a raw-data watermark; a background task builds, catches up to the watermark, and atomically flips the sync flag. Until then, search keeps brute-force scanning raw chunks — the existing fallback, so the read path needs zero changes.

An escape-hatch config `queryNode.segcore.interimIndex.asyncGrowingBuild` (default true) falls back to the legacy synchronous build; a runtime toggle affects growing segments created afterwards.

### Why is this needed?

Today the first build runs synchronously on the insert CGO thread with `NUM_BUILD_THREAD=1`. When a growing segment crosses the build threshold (`max(maxIndexRowCount * indexBuildRatio, nlist * 39)`), that one-shot single-threaded IVF/SCANN training blocks the insert pipeline for the whole build. Since `delegator.UpdateTSafe` sits downstream of insertNode in the pipeline, tSafe stops advancing while the build runs, and readers waiting on their guarantee timestamp hit `WaitTsafeStallTimeout` / context deadlines.

The resulting read timeouts are a visibility-watermark stall caused by the write path, not slow search itself.

### Anything else?

Design notes, verification matrix and disclosed caveats are in the linked PR.

PR #52101, opened 2026-08-01
enhance: build the first growing interim vector index asynchronously

foxspy:
## Summary

The feature remains opt-in while exact-head CI and review run.

The first build of a growing segment's interim vector index previously ran synchronously on the insert CGO thread with `NUM_BUILD_THREAD=1`. That build can block `insertNode.Operate`, delay `delegator.UpdateTSafe`, and make readers waiting on their guarantee timestamp hit `WaitTsafeStallTimeout` even when vector search itself is healthy.

This PR moves only the **first** interim-index build to a dedicated segcore pool. Insert continues publishing raw data while the background task builds and catches up. The index is exposed to readers only after the remaining watermark is finalized under `append_mutex_`; until then, search keeps using the existing raw-data fallback.

The feature is controlled by `queryNode.segcore.interimIndex.asyncGrowingBuild` and now defaults to **`false`**. A hot update affects growing segments created afterwards.

### Design

`VectorFieldIndexing` uses a one-way write-side state machine:

`kNotBuilt -> kBuilding -> (kSynced | kDisabled)`

- **Handoff:** insert advances a monotonic `pending_upto_` raw-data watermark after raw storage is published.
- **Catch-up:** the background task builds the threshold prefix and drains newer rows in bounded staging slices.
- **Finalize budget:** before taking `append_mutex_`, the task estimates the remaining finalize cost from observed catch-up throughput. It finalizes under the lock only when the estimate is within `asyncGrowingFinalizeBudgetMs` (default `20ms`).
- **Deadline fallback:** if catch-up cannot reach the finalize budget within `asyncGrowingCatchupDeadlineMs` (default `30000ms`), the unpublished index is discarded and the field remains on raw search. There is no unbounded forced locked drain.
- **Failure:** build/catch-up failures transition to `kDisabled`; the half-built index is never published and raw search remains authoritative.
- **Teardown:** queued work can be abandoned; running work is joined through the shared task control block.
- **Pool:** capacity follows `queryNode.segcore.interimIndex.buildParallelRate`; pool capacity/activity/queue metrics are exported.

The steady-state `Add` path after publication remains synchronous; this PR only changes the first build.

### Configuration

```yaml
queryNode:
  segcore:
    interimIndex:
      asyncGrowingBuild: false
      asyncGrowingFinalizeBudgetMs: 20
      asyncGrowingCatchupDeadlineMs: 30000
```

### Metrics

- `internal_core_growing_index_latency{type="first_build"|"catchup"}`
- `internal_core_growing_index_rows{type="catchup_rows"}`
- `internal_core_growing_index_inflight{type="first_build"}`
- `internal_core_growing_index_failures{type="build"}`
- `internal_core_growing_index_pool{type="capacity"|"active_threads"|"idle_threads"|"queue_depth"}`

### Validation

Local exact-head checks completed:

- `make milvus -j2` (Release)
- `go test ./util/paramtable -run '^TestComponentParam$' -count=1`
- one DCO-signed commit rebased onto current `master`
- `git diff --check`

CAP/PERF mixed read/write validation: https://github.com/milvus-io/milvus/pull/52101#issuecomment-5239441009

The validation shows improved query isolation and materially lower tail stalls near insert saturation, while sustained insert capacity remains unchanged. It should not be interpreted as uniformly lower latency, cheaper vector computation, or a measured maximum query-throughput limit.

### Boundaries

- Default behavior remains synchronous unless the feature is explicitly enabled.
- The final publish window can still block insert briefly, bounded by the configured estimated budget rather than a zero-gap requirement.
- A running background build can delay segment destruction until that build exits; queued work is abandoned without waiting.
- Background build/catch-up extends the duration of the existing transient memory undercount; peak behavior is not addressed here.
- Steady-state post-publication `Add` remains synchronous.

issue: #52100


foxspy:
## CAP/PERF validation results

> **Build provenance:** these results were collected on follow-up commit [`ccc7c8c16`](https://github.com/milvus-io/milvus/commit/ccc7c8c16af297dcd5e4ed369017822942c44ddf), which is one commit ahead of this PR's current head [`e17e9abb4`](https://github.com/milvus-io/milvus/commit/e17e9abb4e449538ac5174a662345b915f485bdf). The tested image was `v3.0.0-ccc7c8c16-5288(master-20260806-ccc7c8c16-1565e90)`. This is therefore follow-up validation based on this PR, not exact-current-head acceptance evidence.

### Method

- Dataset: OpenAI 5M, 1536 dimensions; Strong consistency; COSINE; topK=100.
- Insert: 100 rows/request, target 1500 -> 3000 -> 4500 -> 6000 rows/s, 30 minutes per sync/async mode.
- Search: one query vector/request; concurrency 5 for ~15 minutes, then 10 for ~15 minutes.
- Supplemental PERF run: concurrency 80 only, 900 seconds per mode, with the same dataset/search semantics and offered insert rate of 3000 rows/s.
- A/B order: sync (`asyncGrowingBuild=false`) then async (`true`), using separate collections on the same instance.
- Environments: CAP `class-2-disk-enterprise` and PERF `class-2-enterprise`.
- QPS below is the arithmetic mean of the c=5 and c=10 windows. Insert capacity is based on actual acknowledged rows/s, not offered load.
- These are closed-loop fixed-concurrency measurements. QPS is therefore primarily `concurrency / average latency`; it is useful for comparing sync/async at the same concurrency, but it is not a direct measurement of maximum query capacity.

### Results

| Env | Offered rows/s | Sync ack | Async ack | Sync QPS | Async QPS | QPS change | Boundary |
|---|---:|---:|---:|---:|---:|---:|---|
| CAP | 1500 | 1500.00 | 1500.00 | 97.34 | 113.54 | +16.6% | count/sample passed |
| CAP | 3000 | ~3000 | ~3000 | 36.19 | 40.06 | +10.7% | source did not cycle; final ~2m14s had no write pressure |
| CAP | 4500 | 3067.32 | 3070.96 | 28.71 | 39.11 | +36.2% | saturated at ~3.07k; count/sample/cycling passed |
| CAP | 6000 | 3070.77 | 3070.81 | 29.83 | 36.79 | +23.3% | saturated at ~3.07k; count/sample/cycling passed |
| PERF | 1500 | 1500.00 | 1500.00 | 259.53 | 260.33 | +0.3% | count/sample passed |
| PERF | 3000 | 2395.77 | 2408.44 | 106.46 | 135.77 | +27.5% | saturated at ~2.4k; RateLimiter + memory quota |
| PERF | 4500 | 2409.83 | 2385.94 | 100.35 | 124.45 | +24.0% | one scheduled sampled ID missing in each mode |
| PERF | 6000 | 2400.06 | 2378.26 | 88.78 | 107.41 | +21.0% | saturated at ~2.4k; RateLimiter + memory quota |

The main result is that async did **not** increase sustained insert capacity: CAP remained at about 3.07k rows/s and PERF at about 2.4k rows/s. Its benefit appeared in query isolation under saturated write pressure. PERF was nearly unchanged at 1500 rows/s (+0.3% QPS), but async improved mean QPS by 21.0%-27.5% across the saturated PERF tiers and by 23.3%-36.2% at CAP 4500/6000.

### Detailed PERF 3000 profiling rerun

| Concurrency | Metric | Sync | Async | Change |
|---:|---|---:|---:|---:|
| 5 | QPS | 77.59 | 110.14 | +41.9% |
| 5 | P99 | 1437.40 ms | 266.92 ms | -81.4% |
| 5 | P99.9 | 4805.03 ms | 2436.75 ms | -49.3% |
| 10 | QPS | 173.49 | 202.90 | +17.0% |
| 10 | P99 | 473.39 ms | 238.42 ms | -49.6% |
| 10 | P99.9 | 3157.72 ms | 1572.93 ms | -50.2% |

Both modes had zero search failures. Insert throughput was again unchanged (`2368.44 -> 2351.11 rows/s`) and quota-limited.

The c=80 supplement showed that the async benefit persists beyond c=10, but also exposed the concurrency knee:

| Metric (c=80) | Sync | Async | Change |
|---|---:|---:|---:|
| QPS | 171.76 | 232.03 | **+35.1%** |
| Average latency | 465.63 ms | 344.71 ms | **-26.0%** |
| P50 | 208.58 ms | 278.69 ms | **+33.6%** |
| P95 | 3015.86 ms | 858.88 ms | **-71.5%** |
| P99 | 4101.18 ms | 2279.76 ms | **-44.4%** |
| P99.9 | 7487.13 ms | 4511.66 ms | **-39.7%** |
| Max | 12069.92 ms | 8960.00 ms | **-25.8%** |

The lower average latency together with a higher P50 means that async changed the distribution rather than uniformly accelerating every request: typical requests paid about 70 ms of additional steady-state overhead (`208.58 -> 278.69 ms`), while the multi-second P95/P99 stalls became much less frequent/severe. In other words, async trades a median-latency regression for substantially better tail isolation and closed-loop throughput.

Both c=80 modes ran for 900 seconds with zero search failures and acknowledged exactly 2.7M rows at approximately 3000 rows/s, with zero insert retries/errors; `count(*)` and a fixed 200-ID sample both passed. This shorter run did not reach the quota-limited steady state seen in the 30-minute run, so it does not revise the sustained insert-capacity conclusion.

The c=80 client was fully occupied, but the server was not CPU-saturated. Async c=10 to c=80 increased concurrency 8x while QPS moved only `202.90 -> 232.03` (+14.4%) and average latency rose `49.26 -> 344.71 ms` (~7x). QueryNode CPU averaged only about 3.2 cores over the async window, while goroutine snapshots showed many parked/search-waiting goroutines. This suggests that c=80 is beyond the efficient scaling knee and that the remaining limitation is more likely execution parallelism/queueing/contention than total CPU capacity. A c=10/20/40/80/120/160 sweep with a no-write control is still required to establish the absolute query-capacity plateau.

The 120-second CPU profiles did not show lower Knowhere/index-engine compute share in async mode; async used similar or more total CPU while completing more queries. This supports the interpretation that async reduces synchronous build/finalize stalls or contention, rather than making vector distance computation itself faster. Exact function-level attribution was not possible because the matching ARM binaries/debug symbols were unavailable locally.

### Correctness and interpretation boundaries

- Final `count(*)` matched acknowledged rows in all runs.
- In the consolidated PERF 4500 run, sync and async each missed one scheduled sampled ID (`4070463` and `4023496`). In the later PERF 3000 profiling rerun, sync missed one sampled ID (`4188059`) while async missed none. These are correctness caveats, not proven data loss, and should be reproduced separately.
- CAP 3000 was not a strict sustained 30-minute result because the 5M source did not cycle.
- CAP and PERF use different instance classes; sync/async comparisons are valid within each environment, but their absolute QPS should not be treated as a like-for-like instance comparison.
- The A/B order was fixed sync -> async rather than randomized/interleaved, so order effects have not been fully excluded.
- c=80 improves throughput and severe tail latency but regresses P50; the feature should therefore be described as reducing periodic stalls/long-tail interference, not uniformly accelerating every query.
- The c=80 result does not prove a 232-QPS hard server limit. It shows that this specific Strong/COSINE/topK100/1536D + 3000 rows/s closed-loop workload was already near its concurrency-scaling knee by c=80.

### Verdict

The follow-up build is **positive for mixed read/write query stability near insert saturation**: across c=5, c=10, and c=80, async consistently improves throughput and materially reduces P99/P99.9 stalls. It is **neutral for sustained insert capacity**, costs more CPU, and does not improve every percentile (c=80 P50 regressed). The evidence supports the intended isolation benefit, not a claim of higher insert capacity, cheaper vector computation, or a proven maximum query throughput.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -567,6 +677,551 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
...
+                throw_deadline(frozen_gap);
+            }
+            continue;
+        }
+
+        auto lock_hold_start = std::chrono::steady_clock::now();
+        AddRange(field_raw_data, final_target, /*interruptible=*/false);
+        // get_valid_data() materializes an O(rows) copy, so it is fetched once
+        // here and never inside the catch-up loop.
+        auto valid_data = field_raw_data->get_valid_data();
```
liliu-z: The finalize budget is computed purely in physical (non-null) rows, but the work it gates is O(logical rows) for nullable vector fields. `frozen_gap` at line 1081 is `PhysicalTarget(...) - index_cur_`, and `consume_rate()` is physical rows per ms — yet after taking `append_mutex_` the code calls `get_valid_data()` (documented in ConcurrentVector.h:169 as a flat copy of the WHOLE bitmap, O(segment rows)) and then `UpdateValidData(..., pending - index_logical)`, which runs two O(logical count) passes inside `GrowingOffsetMapping::Append`. A field that is mostly null has a near-zero physical gap, so the 20ms estimate always passes while the locked section still walks every logical row in the segment, blocking the insert path on `append_mutex_`. Note the same repo already avoids this call for exactly this reason at SegmentGrowingImpl.cpp:3341. Either fold logical rows into the estimate or hoist the validity work out of the lock.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -567,6 +677,551 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
...
+        int64_t target = PhysicalTarget(field_raw_data, pending_upto_.load());
+        int64_t gap = target - static_cast<int64_t>(index_cur_.load());
+        if (gap > 0 && ElapsedMs(catchup_start) >= async_catchup_deadline_ms_ &&
+            estimated_finalize_ms(gap) > async_finalize_budget_ms_) {
+            throw_deadline(gap);
+        }
+        if (gap > 0) {
+            auto round_start = std::chrono::steady_clock::now();
+            int64_t round_from = static_cast<int64_t>(index_cur_.load());
+            AddRange(field_raw_data, target, /*interruptible=*/true);
```
liliu-z: `asyncGrowingCatchupDeadlineMs` is only evaluated at round boundaries (lines 1017 and 1060), never inside `AddRange`. The per-slice loop in `AddRange` checks only `IsCancelled()`, so a single round drains everything up to the watermark it captured at line 1015 without ever re-reading the clock. With a large buffered backlog on the first round, the effective time before the fallback fires can be several multiples of the configured deadline. This does not hold `append_mutex_` (interruptible rounds run unlocked) and teardown stays responsive, so the impact is extra background CPU past the operator's configured limit rather than an insert stall — but the config doc presents the value as a total bound, which it is not.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -264,103 +374,169 @@ VectorFieldIndexing::GetDataFromIndex(const int64_t* seg_offsets,
...
+        int64_t copy_start = std::max(from, chunk_id * size_per_chunk);
+        int64_t copy_end = std::min(to, (chunk_id + 1) * size_per_chunk);
+        int64_t copy_count = copy_end - copy_start;
+        // For mapping storage, chunk data is already compactly stored,
+        // so we can copy directly from chunk
+        auto chunk_data =
+            static_cast<const value_type*>(vec->get_chunk_data(chunk_id));
+        int64_t chunk_offset = copy_start - chunk_id * size_per_chunk;
+        for (int64_t i = 0; i < copy_count; ++i) {
+            staging[copied + i] = chunk_data[chunk_offset + i];
```
liliu-z: `CopySparseRows` copy-assigns each `SparseRow`, and knowhere's copy assignment is a deep copy (heap allocation + memcpy per row), so the whole catch-up range is re-allocated row by row. The header advertises a zero-copy single-chunk fast path, but it is unreachable in production: `kCatchupSparseRows` is 4096 while `queryNode.segcore.chunkRows` defaults to 128, so every slice spans ~32 chunks and always takes the copying branch. `knowhere::sparse::SparseRow` already has a non-owning constructor `SparseRow(size_t count, uint8_t* data, bool own_data)`, so views into the live chunk memory are available — the chunks cannot be reclaimed here anyway since `sync_with_index_` is still false. Peak memory is not doubled (staging holds at most 4096 rows), but the per-row allocations add up over the catch-up range and inflate the time the deadline is measured against.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -567,6 +677,551 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
...
+    for (;;) {
+        if (IsCancelled()) {
+            // Segment tearing down: stay kBuilding, never publish.
+            return;
+        }
+        int64_t target = PhysicalTarget(field_raw_data, pending_upto_.load());
+        int64_t gap = target - static_cast<int64_t>(index_cur_.load());
+        if (gap > 0 && ElapsedMs(catchup_start) >= async_catchup_deadline_ms_ &&
+            estimated_finalize_ms(gap) > async_finalize_budget_ms_) {
+            throw_deadline(gap);
```
liliu-z, line 1133: A catch-up deadline miss is terminal for the whole life of the growing segment: `throw_deadline` raises, `handle_failure` stores `kDisabled`, and both `AppendSegmentIndexDense` and `AppendSegmentIndexSparse` treat `kDisabled` as a no-op break with no path back. The behaviour itself is documented (configs/milvus.yaml, the paramtable Doc, and the WARN log all say the segment permanently falls back to raw search), so this is a design question rather than an undocumented defect — but worth confirming deliberately: the sustained-write burst that trips the 30s deadline is typically transient, while the penalty (brute-force scan for that field until the segment seals) is not, and the feature is most likely to disable itself precisely under the load it was built for. Consider re-arming once the watermark stops outrunning the index.

docs/superpowers/plans/2026-08-07-growing-index-catchup-budget.md
```diff
@@ -0,0 +1,31 @@
+# Async Growing Index Catch-Up Budget Plan
```
liliu-z: This file (the reviewer cited a filename that does not exist in the diff — the actual path is docs/superpowers/plans/2026-08-07-growing-index-catchup-budget.md) creates a new docs/superpowers/ tree that exists nowhere else in the repo. Its contents are a scratch agent work plan, not a design doc: it pins an exact branch head SHA, and its implementation sequence includes steps like committing and pushing a dedicated branch and deploying an AWS UAT instance. Repo convention puts feature design docs under docs/design-docs. Either rewrite it as a design doc there or drop it from the PR.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -567,6 +677,551 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
...
+        // Terminal state FIRST, and with nothing throwable between the lock
+        // and here: everything below (recreate_index's allocations, the log
+        // sinks) can throw, and a throw that escaped before this point would
+        // strand the field in kBuilding forever -- no kDisabled, no counter,
+        // no terminal log, and every later insert silently no-ops into the
+        // kBuilding branch.
+        built_ = false;
+        index_cur_.store(0);
+        state_.store(GrowingIndexState::kDisabled);
+        milvus::monitor::internal_core_growing_index_build_failures.Increment();
```
liliu-z: The catch-up deadline fallback and a genuine knowhere build error land in the same handler, so they emit the identical WARN text and increment the same `internal_core_growing_index_failures{type="build"}` counter. The PR ships that counter as an operational signal, but a deadline miss is a documented, tunable capacity outcome while a build throw is an index defect — different remediations, and an operator alerting on the counter cannot tell them apart. A distinct label (for example type="deadline") would keep the signal actionable.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -175,6 +248,48 @@ VectorFieldIndexing::VectorFieldIndexing(const FieldMeta& field_meta,
...
+    // (which returns immediately above, never reaching here) and the
+    // Submit-catch block in AppendBuild (which resets build_ctrl_ to nullptr
+    // on a win, so the destructor takes the early `build_ctrl_ == nullptr`
+    // return instead of getting here). TryAbandon() above already lost, so
+    // the running task owns the state machine and phase can only be
+    // kRunning.
+    AssertInfo(build_ctrl_->phase.load() != BuildTaskCtrl::Phase::kAbandoned,
+               "abandoned build ctrl must have been cleared before the "
+               "destructor waits");
+    build_ctrl_->finished.wait();
```
liliu-z, line 360: The join added here runs too late to protect mmap-backed raw chunks. `~SegmentGrowingImpl()` releases the segment's mmap blocks in its destructor *body* (`mcm->UnRegister(mmap_descriptor_)`, SegmentGrowingImpl.h:486-489), and a destructor body always completes before any member destructor runs — so `indexing_record_` → `~VectorFieldIndexing()` → `build_ctrl_->finished.wait()` only joins the background build after `MmapBlocksHandler::Deallocate` has already reset each Fixed block and pushed it onto `fix_size_blocks_cache_` for immediate re-issue to another segment's `Allocate` (large blocks are `Close()`d/munmapped). Meanwhile the build is reading `FixedLengthChunk::data_` through `vec->get_chunk_data(chunk_id)` in `CopyDenseRows`/`CopySparseRows`, and `BuildFirstIndexDense` can hand that raw chunk pointer straight to knowhere for the whole build. The destructor comment's justification — "IndexingRecord is declared after InsertRecord, so we are destroyed first and field_raw_data outlives this join" — covers only the `ConcurrentVector` objects, not their mmap backing store, which the enclosing destructor body frees eagerly. Correcting the original claim: mmap is not unconditional for growing dense vectors — `InsertRecord::append_field_meta` gates it, so the trigger is `asyncGrowingBuild=true` plus either `queryNode.mmap.growingMmapEnabled=true` or (interim index + default `queryNode.mmap.vectorField=true` + `denseVectorIndexType: SCANN_DVR`); the existing teardown tests run under neither, which is why they pass. Cancel/join the build before the descriptor is unregistered, or move the `UnRegister` out of the destructor body and let the member `mmap_descriptor_` shared_ptr (declared first, destroyed last) release it.

foxspy: Thanks, confirmed. The explicit destructor-body unregister has been removed, so the mmap descriptor is released by member destruction only after the indexing record has canceled/joined any in-flight async build. I also added a regression test that blocks the build during segment destruction and verifies the mmap blocks remain registered until the join completes.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -497,135 +841,566 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
...
+        bool is_mapping_storage = field_raw_data->is_mapping_storage();
+        // Under IdMap the logical registration travels with each physical Add
+        // (per-slice IdMapData in AddRange); only a null-only logical tail can
+        // remain after the drain, and it is registered inside the lock below.
+        int64_t index_logical =
+            is_mapping_storage
+                ? static_cast<int64_t>(index_->GetIdMap().OutCount())
+                : pending;
+        int64_t logical_tail = pending - index_logical;
+        double frozen_estimate_ms = estimated_finalize_ms(frozen_gap);
```
liliu-z, line 1219: With an all-null tail, `frozen_gap` is zero, so the estimate is 0ms even though `RegisterNullLogicalTail` still scans the tail under the lock. Could the finalize budget account for logical rows too?

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -497,135 +841,566 @@ VectorFieldIndexing::AppendSegmentIndexDense(int64_t reserved_offset,
...
+            AssertInfo(
+                static_cast<int64_t>(index_->GetIdMap().OutCount()) == pending,
+                "id map logical count ({}) did not reach frozen watermark ({})",
+                index_->GetIdMap().OutCount(),
+                pending);
+        }
+        // Atomic with the final AddRange under the same lock: readers that
+        // observe sync_with_index_==true are guaranteed the index covers
+        // every row whose AppendingIndex call has returned.
+        sync_with_index_.store(true);
```
liliu-z, line 1256: If writes stop before the build finishes, IVF_FLAT_CC keeps both the raw chunks and the index until another insert. Could publication trigger cleanup, coordinated with any in-flight inserts?

Issue #52049, opened 2026-07-30
[Bug]:  REST API returns wrong values for large integers (>2^53) in dynamic fields


Nqn7m11:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
Milvus version: v2.6.21 (also reproduced on v2.6.20; v2.6.21 re-verified on 2026-07-24)
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): rocksmq (standalone default)
- SDK version(e.g. pymilvus v2.0.0rc2): REST API v2 (RESTful HTTP via curl; no SDK)
- OS(Ubuntu or CentOS): macOS host / Linux container (Docker Desktop)
- CPU/Memory: 16 GB
- GPU: N/A
- Others: dynamic schema (auto-created collection)
```

### Current Behavior

When a dynamic field stores an integer larger than 2^53 (9007199254740992), the REST API response returns a **different, incorrect value**. The data is stored correctly internally (filter expressions can match the exact value), but the JSON response serializes the integer through float64, losing precision.

```
Stored: {"x": 9007199254740993}  (2^53 + 1)
Filter: x == 9007199254740993    → matches correctly ✓
Output: {"x": 9007199254740992}  → WRONG! Lost last bit

Stored: {"x": 9223372036854775807}  (INT64_MAX)
Filter: x == 9223372036854775807    → matches correctly ✓
Output: {"x": 9223372036854776000}  → WRONG! Off by 193
```

Typed Int64 schema fields are NOT affected — only dynamic fields have this issue.

Actual output from the reproduction below:
```json
{"code":0,"data":[
  {"id":1,"big_val":9007199254740992},
  {"id":2,"big_val":9223372036854776000},
  {"id":3,"big_val":1234567890123456800}
]}
```

### Expected Behavior

The REST API should return the exact integer value that was stored. Int64 values up to 9223372036854775807 should be serialized as exact integers in the JSON response, not rounded through float64:
```json
{"code":0,"data":[
  {"id":1,"big_val":9007199254740993},
  {"id":2,"big_val":9223372036854775807},
  {"id":3,"big_val":1234567890123456789}
]}
```

### Steps To Reproduce

```markdown
# Create collection with dynamic schema
curl -s -X POST http://localhost:19530/v2/vectordb/collections/create \
  -H "Content-Type: application/json" \
  -d '{"collectionName": "repro_h12", "dimension": 4}'

sleep 1

[...]

PR #52103, opened 2026-08-01
fix: preserve int64 precision for REST dynamic-field values > 2^53

lxcxjxhx:
## Motivation

When a dynamic field stores an integer larger than 2^53 (e.g. `9223372036854775807`), the REST v2 query response returns a **different, incorrect value** because `buildQueryResp` decodes the stored dynamic-field JSON with `json.Unmarshal` into `map[string]interface{}`, which represents every JSON number as `float64`. float64 has only a 53-bit mantissa, so large ints silently lose precision (e.g. `9223372036854775807` is returned as `9223372036854776000`).

This corrupts 64-bit IDs, Snowflake/Twitter IDs and nanosecond timestamps read back through the REST API. Typed Int64 fields are unaffected because they go through the native int64 path. Full precision-loss matrix is in issue #52049.

## Changes

- `internal/distributed/proxy/httpserver/utils.go`: in `buildQueryResp`, decode the dynamic-field JSON with `json.Decoder` + `UseNumber()` instead of `json.Unmarshal`, so numeric values stay as `json.Number` and serialize back with their exact digits.
- `internal/distributed/proxy/httpserver/utils_test.go`: add `TestBuildQueryRespDynamicFieldLargeIntPrecision` covering `9223372036854775807`, `9007199254740993` and a small integer, asserting the `json.Number` type and exact re-serialization.

`buildQueryResp` is the single shared response builder for query, get and search (v1 + v2), so all REST response paths are covered by this one fix.

## Testing

- Verified the exact decode path (`json.NewDecoder(...).UseNumber().Decode`) with `internal/json` locally: `9223372036854775807` round-trips exactly and re-serializes as `{"big_val":9223372036854775807,...}` (json.Number is encoded verbatim by the sonic encoder used by the REST renderer).
- `gofmt` clean on both changed files.
- Full `go build ./internal/distributed/proxy/httpserver/` is blocked on this Windows host by an unrelated pre-existing CGO dependency (confluent-kafka-go/librdkafka is not available on Windows); the same package builds on Linux CI.

## Checklist

- [x] Code follows the project's coding standards
- [x] Tests have been added/updated
- [ ] Documentation has been updated (if applicable)
- [x] No new dependencies introduced (or justified)
- [x] Changes are backward compatible

issue: #52049

### DIFF
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -17,6 +17,7 @@
 package httpserver
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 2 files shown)


lxcxjxhx:
Closing: the UseNumber fix for dynamic-field int64 precision has already been merged into master (with additional trailing-content validation). The conflict exists because the same area was rewritten upstream.

Issue #52106, opened 2026-08-01
[Enhancement]: Reduce plan parser rewriter column key allocations


bigsheeper:
## What

The plan parser rewriter builds column grouping keys in `columnKey` with repeated `fmt.Fprintf` calls. This runs while range, JSON term, and text-match rewrite passes group predicates by `ColumnInfo`, including complex expressions that contain several scalar, JSON, and array predicates.

Replacing the formatter calls with append-based integer/bool/string formatting preserves the exact key format while reducing allocation pressure in the full parser benchmark.

## Evidence

Duplicate search before filing:
- Open PR search for `columnKey rewriter optimization` and `plan parser rewriter performance`: no matches.
- Open issue search for the same terms: no matches.

Behavior pinning tests already exist and pass:

```bash
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2/rewriter
```

Benchmark command, run before and after the local change:

```bash
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -run '^$' -bench 'BenchmarkOptimizationComparison/(complex_(and_chain|mixed|nested|with_json|full))$' -benchmem -count=5 ./internal/parser/planparserv2
```

Median samples from the benchmark output:

- `complex_nested`: 175797 ns/op, 39957 B/op, 685 allocs/op -> 162741 ns/op, 39593 B/op, 674 allocs/op.
- `complex_full`: 154299 ns/op, 37447 B/op, 640 allocs/op -> 150081 ns/op, 37096 B/op, 628 allocs/op.
- `complex_with_json`: 105629 ns/op, 24221 B/op, 424 allocs/op -> 103996 ns/op, 23886 B/op, 412 allocs/op.

No benchmark case increased allocs/op.

## Scope

Blast radius is one non-exported helper in `internal/parser/planparserv2/rewriter/util.go`. No exported signatures change.

PR #52107, opened 2026-08-01
enhance: reduce parser rewriter column key allocations

bigsheeper:
issue: #52106

### What changed

Replaced the repeated `fmt.Fprintf` calls in the plan parser rewriter's `columnKey` helper with append-based formatting. The helper keeps the exact same key layout while avoiding formatter overhead in predicate grouping paths.

### Why

`columnKey` is used while grouping range, JSON term, and text-match predicates during expression rewrite. Complex parser expressions can call this helper several times in one parse.

### Verification

Behavior tests:

```bash
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2 ./internal/parser/planparserv2/rewriter
```

Benchmark command:

```bash
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -run '^$' -bench 'BenchmarkOptimizationComparison/(complex_(and_chain|mixed|nested|with_json|full))$' -benchmem -count=5 ./internal/parser/planparserv2
```

Median samples (before -> after):

| Benchmark | ns/op | B/op | allocs/op |
|---|---|---|---|
| `complex_nested` | 175797 -> 162741 (-7.4%) | 39957 -> 39593 (-0.9%) | 685 -> 674 (-11) |
| `complex_full` | 154299 -> 150081 (-2.7%) | 37447 -> 37096 (-0.9%) | 640 -> 628 (-12) |
| `complex_with_json` | 105629 -> 103996 (-1.5%) | 24221 -> 23886 (-1.4%) | 424 -> 412 (-12) |

Static check:

```bash
GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib make static-check
```

Passed with 0 issues. The environment variables point at existing local C++ build artifacts because this nested worktree does not have its own `internal/core/output`.

### DIFF
--- a/internal/parser/planparserv2/rewriter/util.go
+++ b/internal/parser/planparserv2/rewriter/util.go
@@ -1,10 +1,9 @@
 package rewriter
 
 import (
-	"fmt"
 	"math"
 	"sort"
-	"strings"
+	"strconv"
 
 	"github.com/samber/lo"
 
@@ -14,15 +13,21 @@ import (
 )
 
 func columnKey(c *planpb.ColumnInfo) string {
-	var b strings.Builder
-	fmt.Fprintf(&b, "%d|%t|%d|",
-		c.GetFieldId(),
-		c.GetIsElementLevel(),
-		len(c.GetNestedPath()))
-	for _, p := range c.GetNestedPath() {
-		fmt.Fprintf(&b, "%d:%s|", len(p), p)
+	nestedPath := c.GetNestedPath()
+	buf := make([]byte, 0, 32)
+	buf = strconv.AppendInt(buf, c.GetFieldId(), 10)
+	buf = append(buf, '|')
+	buf = strconv.AppendBool(buf, c.GetIsElementLevel())
+	buf = append(buf, '|')
+	buf = strconv.AppendInt(buf, int64(len(nestedPath)), 10)
+	buf = append(buf, '|')
+	for _, p := range nestedPath {
+		buf = strconv.AppendInt(buf, int64(len(p)), 10)
+		buf = append(buf, ':')
+		buf = append(buf, p...)
+		buf = append(buf, '|')
 	}
-	return b.String()
+	return string(buf)
 }
 
 // effectiveDataType returns the real scalar type to be used for comparisons.


czs007:

`Verified:`
- `internal/parser/planparserv2/rewriter/util.go:32` — the header segment is byte-for-byte equivalent: `strconv.AppendInt` / `AppendBool` reproduce the old `"%d|%t|%d|"` exactly, including negative and int64-boundary field ids.
- `internal/parser/planparserv2/rewriter/util.go:38` — the loop segment keeps the `len(p):` length prefix and copies the path bytes verbatim, so the anti-collision property is preserved even for embedded `:`, `|`, NUL, empty segments, or invalid UTF-8.
- `internal/parser/planparserv2/rewriter/util.go:8` — import change is clean: `fmt` and `strings` were used only inside `columnKey`, leaving no dangling references after their removal and the addition of `strconv`.
- `pkg/proto/planpb/plan.pb.go:1275` — the getters line up with the new call sites (`GetFieldId()` → `int64`, `GetIsElementLevel()` → `bool`, `GetNestedPath()` → `[]string`) and are nil-safe, so a nil column still yields `0|false|0|` as before.
- `internal/parser/planparserv2/rewriter/range.go:155` — all consumers (plus `json_term.go:123` and `text_match.go:33`) use the return value purely as a local grouping key with unchanged suffixes; since the key bytes are unchanged, grouping and dedup semantics are unaffected.
- `internal/parser/planparserv2/rewriter/util.go:44` — no aliasing or lifetime concern: `make([]byte, 0, 32)` is only a capacity hint and `string(buf)` copies, with `buf` never escaping.
- `internal/parser/planparserv2/rewriter/range_test.go:360` — existing tests (with `range_json_test.go:417` and `text_match_test.go:35`) already cover same-path merging, non-merging across JSON paths / array indices / fields, and TEXT_MATCH grouping.

Note: the dynamic Go test run could not be executed in the read-only sandbox (`/tmp/go-build*` was not creatable); the conclusions above rest on source-level equivalence analysis.

bigsheeper:
Closing this — the win is too small to be worth the review cost.

The expression parse runs once per request in the proxy task (`task_query.go:528`, `task_delete.go:338`), not per row or per segment. So on the most complex expression benchmarked here, the change saves ~13 µs per query (175797 -> 162741 ns/op) and ~11 allocations; on the other two cases it is 1.5-2.7%. That is not measurable end to end, and it costs readability: hand-rolled `strconv.Append*` byte-building in place of straightforward `fmt` code.

No behavior change either way — the key layout was byte-identical. Worth revisiting if profiling ever puts expression rewrite on a real hot path.

PR #51321, opened 2026-07-13
enhance: batch QueryCoord metadata updates

sijie-ni-0214:
issue: #51320

## What this PR does

- Batch streaming QueryNode replica recovery across collections using `metastore.maxEtcdTxnNum` as the accumulation threshold.
- Batch streaming QueryNode cleanup by removal count and skip stale cleanup entries without blocking valid entries in the same batch.
- Persist each replica batch before updating in-memory state; a failed batch leaves that batch unchanged in memory while later observer batches continue.
- Fall back to the default transaction size when the configured value is invalid.

## Design notes

- `metastore.maxEtcdTxnNum` must not exceed the etcd server `--max-txn-ops` limit.
- A single collection is never split across recovery transactions because its SQ node assignment is computed as one collection-wide unit. A collection with more replicas than the threshold therefore uses one transaction, matching the previous behavior.
- Recovery and cleanup hold the affected collection locks through the corresponding `SaveReplica` call. Locks are acquired in sorted order; this is the deliberate consistency and transaction-reduction trade-off of the batch path.

## Test plan

- [x] QueryCoord meta package tests
- [x] QueryCoord utils package tests
- [x] QueryCoord replica observer package tests
- [x] Paramtable `TestServiceParam`
- [x] `gofumpt -l` on changed Go files
- [x] `git diff --check`

Issue #51320, opened 2026-07-13
[Enhancement]: Batch QueryCoord replica metadata updates


sijie-ni-0214:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Use the configured metastore transaction limit to batch QueryCoord partition and replica metadata operations, including streaming QueryNode replica recovery and cleanup. Recovery calculations should remain collection-scoped while persistence is grouped into bounded transactions.

### Why is this needed?

Per-item metadata transactions increase etcd round trips and make large replica cleanup or recovery operations slow. Sharing the metastore transaction limit also avoids a separate hard-coded batch size and keeps QueryCoord writes within the configured transaction boundary.

### Anything else?

The change includes coverage for stale metadata, persistence failures, invalid batch-size configuration, and continued processing after a failed batch.

PR #52110, opened 2026-08-02
enhance: [cp2.6] batch QueryCoord metadata updates

sijie-ni-0214:
Cherry-pick from master
pr: #51321

issue: #51320

## What this PR does

- Batch streaming QueryNode replica recovery across collections using `metastore.maxEtcdTxnNum` as the accumulation threshold.
- Batch streaming QueryNode cleanup by removal count and skip stale cleanup entries without blocking valid entries in the same batch.
- Persist each replica batch before updating in-memory state; a failed batch leaves that batch unchanged while later observer batches continue.
- Fall back to the default transaction size when the configured value is invalid.

## Backport notes

- Adapt the logging calls to the 2.6 `pkg/v2/log` and `zap` APIs.
- Use the v2 milvus-proto imports required by the 2.6 branch.

## Test Plan

- [x] `CONAN_CMD=conan-1 make SKIP_3RDPARTY=1`
- [x] Package tests for `internal/querycoordv2/meta`, `internal/querycoordv2/observers`, and `internal/querycoordv2/utils`
- [x] `TestServiceParam` in the `pkg` module
- [x] Targeted golangci-lint for all changed packages
- [x] `gofumpt v0.5.0` and `gci v0.11.2` checks on all changed Go files
- [x] `git diff --check`

### DIFF
--- a/internal/querycoordv2/meta/replica_manager.go
+++ b/internal/querycoordv2/meta/replica_manager.go
@@ -63,7 +63,7 @@ type ReplicaManagerInterface interface {
 	// Node management
 	RecoverNodesInCollection(ctx context.Context, collectionID typeutil.UniqueID, rgs map[string]*ResourceGroup) error
 	RemoveNode(ctx context.Context, collectionID typeutil.UniqueID, replicaID typeutil.UniqueID, nodes ...typeutil.UniqueID) error
-	RemoveSQNode(ctx context.Context, collectionID typeutil.UniqueID, replicaID typeutil.UniqueID, nodes ...typeutil.UniqueID) error
+	RemoveSQNodesInCollections(ctx context.Context, removals []SQNodeRemoval) error
 
 	// Metadata access
 	GetResourceGroupByCollection(ctx context.Context, collection typeutil.UniqueID) typeutil.Set[string]
@@ -90,6 +90,12 @@ type ReplicaManager struct {
 
 	idAllocator func() (int64, error)
 	catalog     metastore.QueryCoordCatalog
+}
+
+type SQNodeRemoval struct {
+	CollectionID int64
+	ReplicaID    int64
+	Nodes        []int64
 }
 
 func NewReplicaManager(idAllocator func() (int64, error), catalog metastore.QueryCoordCatalog) *ReplicaManager {
@@ -523,7 +529,8 @@ func (m *ReplicaManager) getSrcReplicasAndCheckIfTransferable(collectionID typeu
 				len(srcReplicas),
 				collectionID,
 				srcRGName,
-				replicaNum))
+				replicaNum),
+		)
 		return nil, err
 	}
 	return srcReplicas, nil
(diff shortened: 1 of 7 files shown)


tinswzy:

Reviewed at head `8c58eb0` against master PR #51321 (source `0f1624a18a`). This is a faithful, compilation-clean cherry-pick — no blocking issues, mergeable. Only non-blocking notes below.

## Parity verification (all confirmed)
- **`replica_manager.go`** — `RemoveSQNodesInCollections` / `RecoverSQNodesInCollections` / `recoverSQNodesInCollectionLocked` are byte-for-byte identical to master. Persist-before-memory ordering (`SaveReplica` → then `putReplicasInMemory`), multi-collection lock acquisition in ascending-sorted order (deadlock-safe), stale-entry skipping (`!ok || CollectionID mismatch` → `continue`; `ContainROSQNode` re-check so a node recovered back to RW is not wrongly removed), and empty-batch short-circuits all match. Proto import correctly adapted to `pkg/v2/proto/querypb`; `replicaPB`, `ContainROSQNode`, `RemoveSQNode`, `putReplicasInMemory` all exist on the 2.6 types.
- **`replica_observer.go` (the resolved conflict)** — logic is identical to master; only the logging is adapted (`mlog.Warn(ctx, …, mlog.Err)`→`log.Warn(…, zap.Error)`, `logger.Info(context.TODO(), …)`→`logger.Info(…)`), which matches the stated backport intent. Recovery batches by replica count, removal batches at `len(removals) >= batchSize`, flush-before-next-batch and failed-batch continuation preserved. `zap`/`paramtable`/`meta` imports present.
- **`utils/meta.go`** — both `RecoverSQNodesInCollection` call sites correctly rewritten to the `[]int64{id}` slice form.
- **`service_param.go`** — `MaxEtcdTxnNum` Formatter identical; `strconv` imported, `getAsInt` returns 0 on parse error so `"invalid"/"0"/"-1"` all fall back to 64 (covered by the new `TestServiceParam` assertions). Note the version tag is `2.6.3`, appropriate for the branch.
- **Callers/mocks** — no leftover references to the removed interface methods `RemoveSQNode` / `RecoverSQNodesInCollection`; the only remaining `RemoveSQNode` hits are the unrelated `*mutableReplica` method. Ported tests compile against 2.6 signatures (`NewReplicaObserver` 3-arg, `PutCollectionWithoutSave`, `NewMeta`, `ChannelDistManagerInterface`/`SegmentDistManagerInterface` fields are interface-typed so the counting wrappers assign cleanly).
- **Tests** — meta, observer, and service_param additions match master; coverage is solid (single-batch persist, stale-collection skip, save-error-keeps-memory on both recover and remove paths, cleanup batch save, stale-removal-for-recovered-RW-node, skip-stale-replicas, and observer-level batching-by-replica-count + continue-after-batch-error + flush-before-next-batch).

## Non-blocking findings

**low — recovery batch can exceed `MaxEtcdTxnNum` for a single oversized collection** (`internal/querycoordv2/observers/replica_observer.go:145`)
```go
if recoveryReplicaCount > 0 && recoveryReplicaCount+replicaCount > batchSize {
    flushRecoveries()
}
recoveryCollections = append(recoveryCollections, collectionID)
recoveryReplicaCount += replicaCount
```
Because the batch is flushed *before* appending an overflowing collection, normal committed batches stay ≤ `batchSize`. But if one collection's modified-replica count alone exceeds `batchSize`, it is committed as a single `SaveReplica`, and `Catalog.SaveReplica` (`kv_catalog.go:82`) forwards everything to `etcdKV.MultiSave`, which issues **one** etcd txn with no internal chunking (`etcd_kv.go:390`). If that count exceeds etcd's `--max-txn-ops` (default 128) the write fails and that collection stops recovering. Failure scenario: a collection with >128 replicas whose SQ assignments all change in one tick. Practically unreachable (replicas-per-collection is small, only *modified* replicas are saved, 128 headroom over the 64 default) and identical to master — flagging only for completeness. The removal path is unaffected: it flushes at per-replica granularity so each `SaveReplica` is ≤ `batchSize`.

**nit — Formatter hardcodes the default** (`pkg/util/paramtable/service_param.go:554`)
The invalid-value fallback returns the literal `"64"`, duplicating `DefaultValue: "64"` on line 548; if the default is ever changed, the fallback silently diverges. Consider `return p.MaxEtcdTxnNum.DefaultValue` (or a shared const). Cosmetic, matches master.

**nit — longer lock hold under batching** (`internal/querycoordv2/meta/replica_manager.go`, `RecoverSQNodesInCollections` / `RemoveSQNodesInCollections`)
Every collection lock in a batch is now held for the full duration of the single `SaveReplica` etcd round-trip, versus one lock per save previously. Lock acquisition is globally ordered (ascending sort) so there is no deadlock, but concurrent per-collection operations on any collection in the batch now serialize behind that one etcd write. This is the intended fewer-txns tradeoff — noting it as an operational consideration, not a defect.

Nothing here blocks the backport; the cherry-pick is accurate and the conflict resolution is correct.

czs007 (approved):
/lgtm

PR #52111, opened 2026-08-02
enhance: [cp3.0] batch streaming query node replica updates

sijie-ni-0214:
Cherry-pick from master
pr: #51321

issue: #51320

## Test Plan

- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" ./internal/querycoordv2/meta ./internal/querycoordv2/observers ./internal/querycoordv2/utils -timeout 300s`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" ./util/paramtable -timeout 300s` (from `pkg/`)

### DIFF
--- a/internal/querycoordv2/meta/replica_manager.go
+++ b/internal/querycoordv2/meta/replica_manager.go
@@ -63,7 +63,7 @@ type ReplicaManagerInterface interface {
 	// Node management
 	RecoverNodesInCollection(ctx context.Context, collectionID typeutil.UniqueID, rgs map[string]*ResourceGroup) error
 	RemoveNode(ctx context.Context, collectionID typeutil.UniqueID, replicaID typeutil.UniqueID, nodes ...typeutil.UniqueID) error
-	RemoveSQNode(ctx context.Context, collectionID typeutil.UniqueID, replicaID typeutil.UniqueID, nodes ...typeutil.UniqueID) error
+	RemoveSQNodesInCollections(ctx context.Context, removals []SQNodeRemoval) error
 
 	// Metadata access
 	GetResourceGroupByCollection(ctx context.Context, collection typeutil.UniqueID) typeutil.Set[string]
@@ -90,6 +90,12 @@ type ReplicaManager struct {
 
 	idAllocator func() (int64, error)
 	catalog     metastore.QueryCoordCatalog
+}
+
+type SQNodeRemoval struct {
+	CollectionID int64
+	ReplicaID    int64
+	Nodes        []int64
 }
 
 func NewReplicaManager(idAllocator func() (int64, error), catalog metastore.QueryCoordCatalog) *ReplicaManager {
@@ -523,7 +529,8 @@ func (m *ReplicaManager) getSrcReplicasAndCheckIfTransferable(collectionID typeu
 				len(srcReplicas),
 				collectionID,
 				srcRGName,
-				replicaNum))
+				replicaNum),
+		)
 		return nil, err
 	}
 	return srcReplicas, nil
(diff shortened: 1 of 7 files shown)


czs007:

This PR is a clean cherry-pick of #51321 onto 3.0; the findings below are low-severity batching/locking/logging observations plus one pre-existing configuration hazard, and every carried-over point from #51321 is adjudicated at the end.

## Low

- **`internal/querycoordv2/observers/replica_observer.go:141`** — The batch is sized from an unlocked snapshot of the replica count (`replicaCount := len(ob.meta.GetByCollection(ctx, collectionID))`, split at `:145`), but `RecoverSQNodesInCollections` re-reads the latest replicas *after* taking `collLock` (`replica_manager.go:938-950`) and persists those, so a concurrent `Spawn` landing in that window makes the real op count exceed the estimate. Failure: with `metastore.maxEtcdTxnNum` set to 128 (= etcd's default `--max-txn-ops`), a tick that snapshots exactly 128 replicas into one batch and then races a `LoadCollection`/`Spawn` that adds one replica → `SaveReplica` issues a 129-op transaction, etcd rejects it with "too many operations in txn request", and the whole batch's streaming-query-node RO/RW migration is dropped for that round with only a warn (recovered on the next tick). Suggestion: derive the batch size from the replica list read under `collLock`, or re-check the op count inside `RecoverSQNodesInCollections` and split before calling `SaveReplica`. (raised by xaxys, tinswzy)

- **`internal/querycoordv2/observers/replica_observer.go:190`** — Batching lost the per-replica failure context in the logs: the batch-level warn at `:158-160` carries only `mlog.Err(err)` with no collection/replica identity, while the field-carrying logger built at `:178-190` (collectionID, replicaID, removedNodes, roNodes, rwNodes) is now used only for an unconditional "will remove it from replica" Info emitted *before* the save; the recovery path at `:134-136` has the same gap. When a batch save fails, operators see N "will remove" Info lines plus one context-free warn and cannot tell which collections/replicas were left uncleaned. Suggestion: attach the batch's collection/replica IDs to the batch-level warns, and either move the per-replica Info after a successful `SaveReplica` or downgrade it to debug. (surfaced during verification)

- **`internal/querycoordv2/observers/replica_observer.go:127`** — `MaxEtcdTxnNum` is used directly as the batch size while `Catalog.SaveReplica` (`kv_catalog.go:89-99`) does a single `MultiSave` and `etcdKV.MultiSave` (`etcd_kv.go:375-395`) puts every op into one transaction; the new formatter (`service_param.go:575-581`) clamps only the lower bound (`<1 → 64`), not the upper. Failure: `metastore.maxEtcdTxnNum` set to 512 against a server still at `--max-txn-ops=128`, with more than 128 replicas → every tick packs up to 512 puts into one txn that etcd rejects, so streaming-query-node recovery and RO cleanup never persist and never self-heal. Why it is unrelated to this PR: the "config value larger than the server's `--max-txn-ops` breaks the write" contract already held repo-wide — rootcoord/datacoord/streamingcoord/streamingnode catalogs have long used the same value as the `SaveByBatchWithLimit` chunk limit that still lands in an unchunked `MultiSave`; this PR only makes querycoord replica writes a new consumer of that existing contract, and the added formatter neither introduced nor widened the missing upper bound. No tracking issue yet. (raised by tinswzy)

## Carried over from #51321

- **`internal/metastore/kv/querycoord/kv_catalog.go:92` — SaveReplica not atomic, memory/etcd divergence (high): Resolved upstream.** The reviewers re-checked this branch: `SaveReplica` (`kv_catalog.go:89-99`) still performs a single `MultiSave` with no `SaveByBatchWithLimit` chunking, and `etcd_kv.go:375-395` puts all ops into one `executeTxn`, so the partial-commit precondition the finding depended on does not exist here.
- **`internal/querycoordv2/observers/replica_observer.go:145` — oversized-collection batch path has no test coverage (medium): Confirmed on this branch**, at `replica_observer_test.go:432-465`. The correction is that the path is not untested-because-undriven: `TestCheckStreamingQueryNodesFlushesRemovalBeforeScanningNextBatch` sets `maxTxnNum=1` against a collection with 2 replicas, so it already takes the `:145` pass-through branch and emits a 2-op recovery batch — but every assertion in that case targets removal ordering, and none checks the recovery batch size; `TestCheckStreamingQueryNodesBatchesRecoveryByReplicaCountAndContinuesAfterError` uses `maxTxn=3` with ≤2 replicas per collection so it never exceeds the bound, and `replica_observer_test.go:515-553` is a cleanup case. With the single-transaction `SaveReplica` on this branch the consequence is a rejected oversized txn retried next tick, not memory/etcd divergence, so the severity is low; adding one assertion on the emitted recovery batch size would close it.
- **`internal/querycoordv2/meta/replica_manager.go:839` — one stale replica cancels the whole SQ-node removal batch (low): Resolved upstream.** Reviewers verified `RemoveSQNodesInCollections` at `replica_manager.go:837-851` now `continue`s on missing replicas, mismatched collection IDs, and empty node sets instead of returning `WrapErrReplicaNotFound`, restoring the per-item skip semantics.
- **`internal/metastore/kv/querycoord/kv_catalog.go:89` — SaveReplica unchunked, maxEtcdTxnNum upper bound unclamped (low): Confirmed on this branch**, at `kv_catalog.go:89-99` plus `etcd_kv.go:375-395` and `service_param.go:575-581`. Verification classified it as pre-existing rather than introduced here — it is filed above under pre-existing hazards, with the note that the server-side `--max-txn-ops` is itself configurable so a hardcoded 128 clamp would be wrong.
- **`internal/querycoordv2/observers/replica_observer.go:145` — batches still oversized when one collection exceeds batchSize (low): Confirmed on this branch**, at `replica_observer.go:145-149`: the flush is guarded by `recoveryReplicaCount > 0` and checked only before appending the next collection, so a single collection larger than `batchSize` is appended unconditionally. Failure: a collection loaded with more replicas than the server's `--max-txn-ops` and needing recovery in one round → a deterministic oversized transaction that etcd rejects every tick, freezing that collection's streaming-query-node assignment. The author has stated this is deliberate, to keep assignment collection-wide, and the removal path at `:185-193` is correctly bounded.
- **`internal/metastore/kv/querycoord/kv_catalog.go:89` — PR description and test plan claim catalog batching changes that do not exist (low): Resolved upstream.** Reviewers read this backport's body and test plan: it contains only the cherry-pick/pr/issue lines plus meta/observers/utils/paramtable test items, with the nonexistent catalog-batching bullet and `TestMetaOperationsUseConfiguredBatchSize` claim removed, so the misleading text is gone here.
- **`internal/querycoordv2/meta/replica_manager.go:938` — collLock window widened to a whole batch, spanning an etcd round trip (low): Confirmed on this branch**, at `replica_manager.go:825-869` and `938-974` (the earlier `internal/querycoordv2/pkg/meta/...` path in our comment was wrong; the correct path is `internal/querycoordv2/meta/replica_manager.go`). Both paths lock the whole batch in ascending order, hold those locks across one `SaveReplica` etcd round trip, then update memory. Failure: etcd slow or briefly unavailable so `SaveReplica` blocks up to `etcd.requestTimeout` (default 10s) on a 64-collection batch → all 64 `collLock`s are held for ~10s, blocking `Spawn`/`RemoveNode`/`RecoverNodesInCollection`/`TransferReplica` (and thus `LoadCollection` and balance) on those collections, and one failure now discards the round's updates for all 64 rather than one. Reviewers confirmed the ascending lock order matches the existing `Put` path (`:376-387`) so there is no deadlock, and `KeyLock` allocates an independent `RefLock` per key; this is an accepted upstream trade-off, noted here rather than raised as a required fix.
- **`pkg/util/paramtable/service_param.go:575` — maxEtcdTxnNum clamps only the lower bound (low): Confirmed on this branch**, same root as the item above at `service_param.go:575-581`; it is covered by the pre-existing-hazard entry rather than repeated as a fix request for this PR.

sijie-ni-0214:
Thanks for the thorough review.

I agree with the two low-severity observations.

For the recovery batch-size estimate, a concurrent `Spawn` can make the final replica count exceed the pre-lock estimate. `SaveReplica` is still a single etcd transaction, and memory is updated only after it succeeds, so this results in a failed observer round rather than partial persistence or memory/etcd divergence; it will be retried on the next observer cycle. Enforcing the limit after acquiring collection locks is not a mechanical split because SQN assignment is computed collection-wide, so I prefer to keep this cherry-pick aligned with #51321 and handle stronger transaction-cap enforcement separately.

I also agree that batch failure logs should include the affected collection/replica IDs, and that the cleanup success log would be clearer after a successful save. This is an observability improvement suitable for a follow-up.

Thanks again for confirming that the remaining items are either pre-existing or accepted trade-offs and that this PR is mergeable as-is.

czs007 (approved):
/lgtm

Issue #52112, opened 2026-08-02
[Bug]: Search parses naive TIMESTAMPTZ filters before resolving timezone


bigsheeper:
### Is there an existing issue for this?

- [x] I have searched the existing issues and open PRs

Related closed issues found during duplicate search:
- #44594 covered rejection of the `timezone` search parameter itself and was closed as fixed.
- #46426 covered hybrid-search timezone API support and is closed/stale.

I did not find an open issue or open PR for the current proxy plan-generation ordering bug.

### Environment

```markdown
- Milvus version: master e022d123b0dda7567bc7ea8693fc71bc7c9bde5e
- Deployment mode: unit-test repro in proxy
```

### Current Behavior

`searchTask.PreExecute` builds the search plan before it resolves `t.resolvedTimezoneStr` from request `SearchParams` or collection properties.

For a normal `Search` request with a naive `TIMESTAMPTZ` literal and `timezone=Asia/Shanghai`, the parser receives an empty timezone and interprets the literal with the process local timezone instead of the request timezone. On a machine with `JST+0900`, the generated predicate bound for `ts > ISO '2025-01-01T00:00:00'` is:

```text
actual:   1735689600000000
expected: 1735660800000000
```

That is an 8-hour-vs-9-hour interpretation error for the filter boundary, so valid searches can return the wrong rows.

### Expected Behavior

Search and hybrid-search plan generation should use the resolved request-level timezone, or the collection timezone fallback, when parsing naive `TIMESTAMPTZ` filter literals. This should match `queryTask`, which resolves timezone before creating plan args.

### Steps To Reproduce

A focused unit test reproduces the issue by asserting the serialized search plan predicate:

```bash
GOWORK=off \
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
CGO_LDFLAGS='-Wl,-rpath-link,/home/sheep/workspace/milvus/internal/core/output/lib' \
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy \
  -run 'TestSearchTaskPreExecuteUsesTimezoneForTimestamptzFilter'
```

Before the fix, this fails:

```text
expected: 1735660800000000
actual  : 1735689600000000
```

[...]

PR #52113, opened 2026-08-02
fix: resolve search timezone before plan generation

bigsheeper:
issue: #52112

### What changed

Search `PreExecute` now resolves and validates the request or collection timezone before `initSearchRequest` / `initAdvancedSearchRequest` generates the search plan. This keeps naive `TIMESTAMPTZ` filter literals aligned with the same timezone later used for result formatting.

### Why

Before this change, plan generation could parse a naive `TIMESTAMPTZ` literal with the process local timezone because `t.resolvedTimezoneStr` was still empty. On a `JST+0900` host, a search request with `timezone=Asia/Shanghai` generated the wrong predicate bound:

```text
expected: 1735660800000000
actual:   1735689600000000
```

### Verification

```bash
GOWORK=off \
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
CGO_LDFLAGS='-Wl,-rpath-link,/home/sheep/workspace/milvus/internal/core/output/lib' \
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy \
  -run 'TestSearchTaskPreExecuteUsesTimezoneForTimestamptzFilter'
```

```bash
GOWORK=off \
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
CGO_LDFLAGS='-Wl,-rpath-link,/home/sheep/workspace/milvus/internal/core/output/lib' \
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy
```

```bash
GOWORK=off \
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib \
CGO_LDFLAGS='-Wl,-rpath-link,/home/sheep/workspace/milvus/internal/core/output/lib' \
make static-check
```

### DIFF
--- a/internal/proxy/task_search.go
+++ b/internal/proxy/task_search.go
@@ -263,6 +263,18 @@ func (t *searchTask) PreExecute(ctx context.Context) error {
 	// searches with small result size could no longer need requery.
 	traceVal, _ := funcutil.GetAttrByKeyFromRepeatedKV(PipelineTraceKey, t.request.GetSearchParams())
 	t.traceEnabled = strings.EqualFold(traceVal, "true")
+	timezone, exist := funcutil.TryGetAttrByKeyFromRepeatedKV(common.TimezoneKey, t.request.SearchParams)
+	if exist {
+		if !timestamptz.IsTimezoneValid(timezone) {
+			log.Info(ctx, "get invalid timezone from request", mlog.String("timezone", timezone))
+			return merr.WrapErrParameterInvalidMsg("unknown or invalid IANA Time Zone ID: %s", timezone)
+		}
+		log.Debug(ctx, "determine timezone from request", mlog.String("user defined timezone", timezone))
+	} else {
+		timezone = getColTimezone(collectionInfo)
+		log.Debug(ctx, "determine timezone from collection", mlog.Any("collection timezone", timezone))
+	}
+	t.resolvedTimezoneStr = timezone
 	// initSearchAggregation must run before init{,Advanced}SearchRequest so
 	// that t.SearchRequest.GroupByFieldIds and t.aggCtx are populated before
 	// queryInfo is built and captured — otherwise queryInfo.GroupByFieldIds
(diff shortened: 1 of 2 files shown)


czs007:

Verified:
- `internal/proxy/task_search.go:266` — the request timezone is resolved and validated before either search-plan initialization path, so the dependency it relies on (`collectionInfo`, assigned and nil-checked at line 187) is already available at the new location.
- `internal/proxy/task_search.go:277`, `internal/proxy/task_search.go:1184` — `resolvedTimezoneStr` is assigned before `tryGeneratePlan` passes it into `ParserVisitorArgs`, so naive `TIMESTAMPTZ` literals are bound with the intended timezone rather than the zero value.
- `internal/proxy/task_search.go:289`, `internal/proxy/task_search.go:560` — every advanced/hybrid sub-request generates its plan through the same timezone-aware helper, so both plan-generation paths are covered by the move.
- `internal/proxy/task_search.go:274`, `internal/proxy/util.go:3343` — requests without an explicit timezone fall back to the collection timezone and then to UTC consistently.
- `internal/proxy/task_search.go:1475` — the same resolved timezone remains available for PostExecute result formatting, so the output path is unchanged, including for requests with no `timestamptz` filter.
- `internal/proxy/task_search.go:268` — only an invalid user-supplied IANA timezone is classified as `ParameterInvalid`; the earlier failure point is pure fail-fast and preserves existing error semantics, and timezone state stays request-local on `searchTask` with no shared mutable state introduced.
- `internal/proxy/task_search_test.go:114`, `internal/proxy/task_search_test.go:184` — the test runs the full `PreExecute` path, unmarshals the serialized plan, and asserts the exact predicate bound; since the pre-fix state resolves an empty timezone through `time.LoadLocation("")` to UTC (`pkg/util/timestamptz/timestamptz.go:73`), the assertion is deterministic and catches the regression on any CI host regardless of local time zone.

czs007:
This fix needs a 2.6 cherry-pick: the ordering bug was introduced on master by #46546 and reached 2.6 the same day via #46556 (d6863c5994), so all releases since v2.6.8 parse naive `TIMESTAMPTZ` literals in search filters as UTC regardless of the request/collection timezone. Verified on current 2.6 HEAD: `task_search.go` assigns `resolvedTimezoneStr` at L309, after plan generation at L235/L237, with the same `ParserVisitorArgs` wiring at L931.

On test coverage: the e2e suite already covers exactly this scenario — naive literal + non-UTC timezone with precise row assertions — in `test_milvus_client_timestamptz_different_timezone_query`, but only through `query()`, whose plan generation resolves the timezone first and was never affected. The search-path e2e cases pass `timezone` without any `TIMESTAMPTZ` literal filter (`test_milvus_client_timestamptz_search` has no filter at all; the only filtered search cases use `is null`/`is not null`). Duplicating those timezone-boundary assertions onto `search()` would keep this path covered end-to-end and would have caught this seven months earlier.

czs007 (approved):
/lgtm

Issue #52116, opened 2026-08-02
[Enhancement]: tests/ — resolve 3 shadowed duplicate definitions and the ruff rules deferred by the baseline cleanup


xiaofan-luan:
Follow-up to #52114 / #52115, which cleared the ruff baseline in `tests/` but deliberately deferred everything that would change test behaviour. Two separable pieces of work.

## 1. Three duplicate definitions that silently shadow each other — resolved by #52115

In each of these, the later definition wins and the earlier one **never runs**, despite looking like a live test:

| file | name | detail |
|---|---|---|
| `tests/restful_client/testcases/test_vector_operations.py` | `TestDeleteVector` | class defined twice. The shadowed one is `@pytest.mark.L0` and holds `test_delete_vector_default` (2x2 parametrized); the surviving one holds only negative cases. The file already uses the `TestInsertVectorNegative` / `TestSearchVectorNegative` convention, so the second class was a copy whose name was never updated. |
| `tests/python_client/testcases/test_compaction.py` | `test_compact_after_delete` | defined twice in `TestCompactionParams`, covering different things. The shadowed one is parametrized (`delete_pos`) and covers *partial* deletes plus a compaction-plan check; the surviving one deletes everything. Partial-delete coverage was therefore not running at all. |
| `tests/python_client/utils/util_pymilvus.py` | `add_vector_field` | defined twice with different signatures — `(entities, is_normal=False)` appends a vector field and returns entities; `(nb, dimension=default_dim)` builds a field dict, discards it, and returns only the generated name. The second (broken) one wins. |

**Correction to an earlier version of this issue:** it claimed the `add_vector_field` signature mismatch meant some caller was silently getting the wrong function. That was a guess and it was wrong — grepping the repo shows **neither definition has any caller**. Every `add_vector_field` match is something else: `test_milvus_client_add_vector_field` test methods, the `Op.add_vector_field` enum member in `chaos/checker.py`, and a `Checker` method of the same name. Both definitions are dead code.

#52115 resolves all three (fourth commit): renames the negative class to `TestDeleteVectorNegative`, renames the shadowed compaction case to `test_compact_after_delete_partial`, and deletes the broken `add_vector_field`. Net effect is 4 L0 and 2 L3 cases start running and nothing stops running.

## 2. Rules disabled to keep the gate green — F403/F405 resolved, rest remain

These are in the global `ignore` list in `tests/ruff.toml`. Each could be re-enabled after a dedicated cleanup:

[...]

PR #52117, opened 2026-08-02
test: resolve three shadowed duplicate definitions so the dead cases run again

xiaofan-luan:
issue: #52116

> **Stacked on #52115.** Until that merges, the diff here shows its 171 files too — **review only the last commit, `d92224eee1`** (4 files, +2/−16). The diff collapses to just that commit once #52115 lands.

Three places in `tests/` define the same name twice in the same scope. Python keeps the later definition, so the earlier one is dead while still reading as live code — a test that looks like it exists and never runs. This resolves all three and removes the F811 per-file exemptions that #52115 parked them under; F811 is now clean across `tests/` with no suppressions.

### 1. `restful_client/testcases/test_vector_operations.py` — `TestDeleteVector`

Declared twice. The second one contains only negative cases:

```python
@pytest.mark.L0
class TestDeleteVector(TestBase):          # line 778 — shadowed, never ran
    def test_delete_vector_default(...)    # 2x2 parametrized

@pytest.mark.L1
class TestDeleteVector(TestBase):          # line 838 — wins
    def test_delete_vector_with_invalid_api_key(...)
    def test_delete_vector_with_invalid_collection_name(...)
    def test_delete_vector_with_non_primary_key(...)
```

The file already uses `TestInsertVectorNegative` and `TestSearchVectorNegative`, so this is a copy that never got its class name updated. Renamed the second to `TestDeleteVectorNegative`.

**This is the one to look at:** it brings back an `L0` case, so 4 parametrizations of `test_delete_vector_default` start running in PR CI. They have never executed, so if the restful delete path has drifted since they were written, this is where it shows up. That is the point of the fix, but it is worth knowing before merging.

### 2. `python_client/testcases/test_compaction.py` — `test_compact_after_delete`

Defined twice in `TestCompactionParams`, covering different things:

- shadowed (L3, `@parametrize("delete_pos", [1, tmp_nb // 2])`): deletes one entity, then half, and checks the compaction plan
- surviving (L2): deletes everything

Renamed the shadowed one to `test_compact_after_delete_partial`, which is what it actually tests. The surviving definition is untouched — same name, same body, same mark. Note there is already a separate `test_compact_after_delete_all`, so the partial case was the only coverage of partial deletes, and it was not running.

### 3. `python_client/utils/util_pymilvus.py` — `add_vector_field`

Defined twice with **different signatures**:

```python
def add_vector_field(entities, is_normal=False):     # line 455 — shadowed
    ...
    entities.append(field)
    return entities

def add_vector_field(nb, dimension=default_dim):     # line 527 — wins
    field_name = gen_unique_str()
    field = {...}          # built, then discarded
    return field_name      # returns only the name
```

Removed the second one. It is both the broken one (constructs `field` and throws it away) and the one that was winning. Grepped the whole repo: **neither definition has a single caller** — every `add_vector_field` hit is something else (`test_milvus_client_add_vector_field` test methods, the `Op.add_vector_field` enum member in `chaos/checker.py`, and a `Checker` method of the same name). So this is dead-code removal, not an API change.

*Correcting something I wrote in #52116: I had guessed callers were silently getting the wrong function. There are no callers at all.*

### Net effect on what executes

| | |
|---|---|
| start running | `test_delete_vector_default` ×4 (**L0**), `test_compact_after_delete_partial` ×2 (L3) |
| stop running | nothing |
| unchanged | the 3 negative cases, now under `TestDeleteVectorNegative` (still L1) |

Computed by parsing each file before and after and resolving shadowing the way Python does, not by reading the diff.

### Checks

- `ruff check tests` → clean, with the three `per-file-ignores` entries deleted
- `ruff format --check tests` → 307 files already formatted
- `compileall` over all 307 files under Python 3.12 → passes
- no external reference to `TestDeleteVector`, `test_compact_after_delete`, or `add_vector_field` exists anywhere in the repo (`*.py`, `*.sh`, `*.yaml`, `*.json`)


xiaofan-luan:
Folded into #52115 as its fourth commit — keeping this as a single PR. Same change, same verification; see the 'Three shadowed duplicate definitions' section there.

PR #52119, opened 2026-08-03
enhance: [cp3.0] Reduce QueryNode distribution report payload

sijie-ni-0214:
Cherry-pick from master
pr: #51040

issue: #51039

## Summary

- Add QueryNode-to-QueryCoord delta distribution reports and manager patching, falling back to full reports when the timestamp cannot be safely used.
- Reduce QueryNode metrics allocation by iterating segments in place.
- Preserve the 3.0 removal of the legacy QueryNode DropIndex path (#51822); the master commit only changed that already-removed method as context.

## Test Plan

- [x] `git diff --check upstream/3.0...HEAD`
- [x] gofumpt and gci checks for all non-generated changed Go files
- [x] `go test -tags dynamic,test -gcflags="all=-N -l" ./internal/querycoordv2/dist ./internal/querycoordv2/meta` (with macOS rpaths)
- [x] QueryNode, delegator, and segments tests are blocked by the pre-existing stale Cgo output: `SegcoreSetEnableGISSplitFusion` is missing from `internal/core/output/include/segcore/segcore_init_c.h`.
- [x] `make SKIP_3RDPARTY=1` cannot refresh that output because the unmodified 3.0 baseline fails in `ChunkedSegmentSealedImpl.cpp:8168-8169` on macOS OpenMP structured-binding capture support.
- [x] `make static-check` is blocked by the same Cgo typecheck failure.


xaxys:

## Key issues

- **High — stale removal markers can delete live segment/channel entries.** `internal/querynodev2/services.go:1342-1344` and `:1371-1373` emit detached removal markers without checking current state, while `internal/querynodev2/data_distribution_report.go:96-121` makes the last marker win.
  - Segment: `SegmentManager.Remove` removes the active entry before waiting for release (`internal/querynodev2/segments/manager.go:649-683`), and reload during that interval is explicitly allowed (`internal/querynodev2/segments/segment_loader.go:439-445`). A reload can mark an upsert before the old release returns; the late remove at `internal/querynodev2/services.go:704-708` then causes QueryCoord to remove the newly loaded segment.
  - Channel: Watch can pass the unsubscribe check at `internal/querynodev2/services.go:218-229`; Unsub can then observe no delegator at `:382-385`; Watch can insert and mark the channel at `:270` and `:361`; finally Unsub can overwrite that marker at `:433`. QueryCoord consequently removes a live shard leader.

- **Medium — unserviceable delta responses lose subsequent offline-routing changes.** `internal/querycoordv2/dist/dist_handler.go:304-309` resets the cursor only for full responses. `internal/querynodev2/delegator/distribution.go:465-477` makes an offline segment wire-visible by changing its node/version, but `:380-392` dirties the channel only when the serviceability boolean changes. After one segment makes the channel unserviceable and is reported, another segment going offline while it remains unserviceable produces no timestamp or delta. QueryCoord retains the old route, so `internal/querycoordv2/checkers/leader_checker.go:179-180` can incorrectly consider it current and skip repair.

- **Medium — the leader-view marker can be consumed before the new snapshot is published.** `internal/querynodev2/delegator/distribution.go:410-440` updates the authoritative map but only queues asynchronous snapshot generation. `GetDataDistribution` later reads the published snapshot at `internal/querynodev2/services.go:1429-1431`, while leader loads/SyncDistribution publish their marker and timestamp when the RPC returns (`:463`, `:510`, and `:1460-1462`). A pull in this window clears the sole dirty marker while serializing the old leader view. If serviceability remains unchanged, the callback at `internal/querynodev2/delegator/distribution.go:380-392` sends no corrective delta.

- **Medium — the delta revision is not strictly monotonic.** `internal/querynodev2/services.go:1830-1835` assigns `time.Now().UnixNano()` directly. Equal clock values or a backward wall-clock adjustment leave a dirty marker behind a revision that is equal to or below QueryCoord’s cursor; `internal/querynodev2/data_distribution_report.go:68-74` then reports “no change,” and `internal/querycoordv2/dist/dist_handler.go:175-179` rejects the response. The distribution can remain stale until a later revision exceeds the previous value.

## Fix suggestions

1. Resolve the union of dirty and removed keys against current state when constructing a delta: present means upsert; absent means removal. Add blocked-release/reload and Watch/Unsub interleaving tests.
2. Keep the full-pull fallback while a leader is unserviceable, or dirty the channel after every wire-visible offline-routing mutation, not only serviceability transitions.
3. Publish the snapshot before advancing the marker/timestamp, report from the locked authoritative maps, or emit a follow-up marker after every completed snapshot mutation.
4. Generate revisions with `max(time.Now().UnixNano(), lastModifyTs+1)` or use a logical monotonic sequence.

sijie-ni-0214:
Thanks for the detailed review. I rechecked the relevant interleavings and have the following conclusions:

1. The stale removal-marker interleaving does not occur in the normal control path. QueryCoord serializes SegmentTasks by (replicaID, segmentID, isGrowing), and serializes Watch/Unsub tasks by (replicaID, channel). Therefore a reload/watch cannot overtake the corresponding release/unsub task and leave a late detached removal marker behind it.

2. An unserviceable delegator is intentionally not retained as a routable leader. Its dirty channel is not silently discarded: since QueryNode omits an unserviceable leader view, the remaining dirty channel is reported through RemovedChannelNames. ChannelDistManager.Patch then removes the previous leader view, rather than retaining its stale route; the checker recreates the leader. Subsequent offline mutations on that already-removed leader do not require per-segment deltas.

3. Confirmed. The asynchronous delegator snapshot can advance the distribution marker before GetSegmentInfo observes the new snapshot. This is a pre-existing snapshot-publication issue, not specific to this delta backport, and it also affects delete forwarding. It needs the synchronous snapshot-publication fix from https://github.com/milvus-io/milvus/commit/3f2888a2afb9ec3d9e1801f76ed1588e63fd58b7, which I will backport separately before merging this backport.

4. Agreed that UnixNano is not a strict logical clock. This is a low-probability, recoverable liveness edge case: a later greater revision restores reporting. I prefer not to add a logical revision mechanism to this focused backport.

czs007:

This backport of #51040 was re-reviewed end to end against the delta data-distribution path; one new low-severity staleness gap was found in this round, and every finding from the upstream review is adjudicated at the bottom.

## Low

- **`internal/querynodev2/services.go:1577`** — `Delete` (and `DeleteBatch` at `:1635`) apply deletes to sealed segments without calling any `distDeltaTracker` mark or `updateDistributionModifyTS()`, yet `segment.Delete` advances that segment's `LastDeltaTimestamp`, which is reported through `buildSegmentVersionInfo` (`data_distribution_report.go:130`). Before this PR any `modifyTs` bump rebuilt every `SegmentVersionInfo`, so the delete-driven timestamp was refreshed incidentally; in delta mode only segments in `dirtySegments` are rebuilt (`services.go:1334-1345`, `:1372-1379`). Failure: a sealed segment that only keeps receiving forwarded `Delete`/`DeleteBatch`, with no further `LoadSegments`/`ReleaseSegments`/`SyncDistribution` → QueryCoord's view of that segment (`LoadedTimestamp` in `/metrics`, derived from `LastDeltaTimestamp` at `meta/segment_dist_manager.go:157`) stays frozen at the last full report, so an operator reading "this segment has seen no recent deletes" is reading a stale value. Suggestion: mark the successfully deleted sealed segment IDs via `markSegmentUpsert` in both handlers (with the usual `defer updateDistributionModifyTS()`), or explicitly document that `LastDeltaTimestamp` is best-effort under delta reporting. Scope is observability only — `segment_dist_manager.go:157` is the field's sole consumer; no checker or balance policy reads it, and query correctness is unaffected. *(raised by chyezh, tinswzy)*

## Carried over from #51040

1. `distribution.go:460` — leader-view serviceable degradation never reported under delta. **Resolved upstream**: `delegator/distribution.go:353-392` detects the serviceability transition and invokes the callback, and `services.go:1856-1861` dirties the channel and advances the modify timestamp; reviewers additionally verified the wiring is live — `NewShardDelegator` assigns `distribution` in the struct literal (`delegator.go:1588`) before options run (`:1605`), so the nil-guard in `WithLeaderViewUpdatedCallback` passes.
2. `services.go:1349` — concurrent `GetDataDistribution` can silently lose a delta by unconditionally writing `lastReportedTs`. **Resolved upstream**: the delta base is invalidated while building and stale commits are rejected via `deltaGeneration` (`services.go:1301-1322`, `:1434-1438`), with the interleavings covered at `data_distribution_report_test.go:202-347`.
3. `internal/querycoordv2/dist/dist_handler.go:305` — **Not re-examined**: the original entry carried only a file:line with no proposition or failure scenario, so it could not be adjudicated this round; it stands as recorded.
4. `dist_handler.go:379` — no kill switch for the delta protocol. **Resolved upstream**: `SupportDelta` is gated by `queryCoord.enableDataDistributionDelta` (`dist_handler.go:375-381`, `pkg/util/paramtable/component_param.go:3338-3345`, exposed at `configs/milvus.yaml:493`).
5. `services.go:1279` — full scan moved inside the global mutex, serialized against the data-plane RPC completion path. **Resolved upstream**: the tracker mutex is released at `services.go:1322`, before the segment and delegator scans at `:1326-1432`, so the reported contention path does not exist on this branch.
6. `data_distribution_report_test.go:958` — tautological unit test. **Resolved upstream**: real concurrent report/reset coverage, including stale-commit rejection, now exists at `data_distribution_report_test.go:202-347`.
7. `services.go:1362` — growing segments created by the insert pipeline do not dirty the channel. **Resolved upstream**: `delegator/delegator_data.go:165,243,257-258` flags newly created growing segments and notifies the leader-view callback.
8. `internal/querynodev2/services.go:503` — `LoadSegments` marks the tracker dirty even on load failure. **Confirmed on this branch**: the unconditional defers at `services.go:510` and `:549-558` still mark entries on the failure paths at `:512-533` and `:579-580`. Delta construction re-reads real state (`:1336-1340`: a segment that failed to load has `GetSealed == nil` and lands in `removedSegmentIDs`), so the demonstrated impact is one redundant reconcile rather than incorrect distribution data.
9. `internal/querynodev2/services.go:1298` — `TotalSegmentCount` can disagree with the reported segment set. **Confirmed on this branch** at `services.go:1326-1332`: the count and the reported set come from separate traversals of a concurrent map (`segments/manager.go:608-623` takes no lock), so under delta the `segmentCnt` written from `resp.TotalSegmentCount` (`dist/dist_handler.go:181-184`) can transiently disagree with the node's actual entries and mislead `assign/assign_policy_roundrobin.go:174` for one round; it self-corrects on the next report. Two details in our earlier write-up were wrong and are corrected here: the traversals are not "each independently locked" (they are lock-free), and the `services.go:2316` line reference did not exist.
10. `internal/querynodev2/segments/manager.go:283` — shared iteration helper semantics changed for all callers. **Retracted**: our earlier comment framed this as a hazard, but the only existing callback that can return `false` mid-iteration, `GetAndPinBy` (`manager.go:514-538`), already unpins everything and returns no segment set on error, so no current caller has an observable behavior change. The change at `:272-285` in fact aligns the candidate-map branch with the `segmentIDs` branch (`:250-253`), which already returned from the whole helper.
11. `internal/querynodev2/services.go:1349` — growing-row staleness effectively unbounded. **Confirmed on this branch** at `delegator/delegator_data.go:243,257-259`: `growingSegmentAdded` is set only when a *new* growing segment is created, so continued inserts into an existing growing segment neither dirty the channel nor advance `lastModifyTs`, while `NumOfGrowingRows` is recomputed live per report from `segment.InsertCount()` (`services.go:1396`). Under delta, only channels present in `dirtyChannels` have their leader view rebuilt (`services.go:1360-1364`), so a busy channel's row count can stay frozen at its last marked value while unrelated load/release events keep `lastModifyTs` moving. This is a scheduling input, not just an observation: `assign/assign_policy_score.go:93,97` and `assign/assign_policy_rowcount.go:76` accumulate node load from `ch.View.NumOfGrowingRows`, so the node is under-weighted and keeps attracting new segments/channels. Query correctness is unaffected.

sijie-ni-0214:
Thanks for the adversarial review. I agree that this backport introduces no must-fix issue. The `LastDeltaTimestamp` and `NumOfGrowingRows` refresh gaps will be handled separately so that high-frequency DML does not negate the payload reduction. The synchronous delegator snapshot fix in https://github.com/milvus-io/milvus/pull/52167 will also be backported before this PR is merged.

czs007 (approved):
/lgtm

PR #52120, opened 2026-08-03
enhance: parallelize json stats build

zhagnlu:
JSON stats build benchmark — 500K rows per segment
Concurrent segments  ||   Non-parallel   ||      8 workers   ||       Speedup
1              ||                             6.810 s       ||         1.641 s        ||      4.15× faster
4              ||                            7.140 s        ||         4.806 s       ||      1.49× faster
8               ||                           10.878 s       ||        9.658 s  ||           1.13× faster
Setup: Release build, 12 CPUs, local filesystem, one warm-up followed by five measured runs; median values are reported. Data generation is excluded from the timing.
Note: “Non-parallel” means each segment is built serially internally. Multiple segment builds are still launched concurrently in 8-segment cases.

### DIFF
--- a/internal/core/src/futures/Executor.h
+++ b/internal/core/src/futures/Executor.h
@@ -33,4 +33,7 @@ getSearchCPUExecutor();
 folly::CPUThreadPoolExecutor*
 getLoadCPUExecutor();
 
+folly::CPUThreadPoolExecutor*
+getJsonStatsBuildExecutor();
+
 };  // namespace milvus::futures
--- a/internal/core/src/futures/future_c.cpp
+++ b/internal/core/src/futures/future_c.cpp
@@ -9,6 +9,8 @@
 // is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
 // or implied. See the License for the specific language governing permissions and limitations under the License
 
+#include <algorithm>
+
 #include "Executor.h"
 #include "Future.h"
 #include "folly/executors/CPUThreadPoolExecutor.h"
@@ -18,6 +20,7 @@
 #include "log/Log.h"
 #include "monitor/Monitor.h"
 #include "prometheus/gauge.h"
+#include "storage/EntryStreamUtils.h"
 
 extern "C" void
 future_cancel(CFuture* future) {
@@ -73,4 +76,27 @@ executor_set_load_thread_num(int thread_num) {
     milvus::monitor::internal_cgo_pool_size_load.Set(thread_num);
     LOG_INFO("future executor setup load cpu executor with thread num: {}",
              thread_num);
+}
+
+extern "C" void
+executor_set_json_stats_build_thread_num(int thread_num) {
+    auto normalized_thread_num = std::max(1, thread_num);
+    milvus::futures::getJsonStatsBuildExecutor()->setNumThreads(
+        normalized_thread_num);
+    LOG_INFO(
+        "future executor setup json stats build cpu executor with thread num: "
+        "{}",
+        normalized_thread_num);
+}
+
+extern "C" void
+executor_set_json_stats_build_max_inflight_bytes(int64_t max_inflight_bytes) {
+    auto normalized_max_inflight_bytes =
+        static_cast<size_t>(std::max<int64_t>(0, max_inflight_bytes));
+    milvus::storage::TransientMemoryBudget::SetJsonStatsBuildBudgetBytes(
+        normalized_max_inflight_bytes);
+    LOG_INFO(
+        "json stats build transient memory budget setup with max inflight "
+        "bytes: {}",
+        normalized_max_inflight_bytes);
 }
--- a/internal/core/src/index/json_stats/JsonKeyStats.cpp
+++ b/internal/core/src/index/json_stats/JsonKeyStats.cpp
@@ -14,17 +14,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <nlohmann/json.hpp>
-#include <string.h>
+#include <algorithm>
+#include <atomic>
 #include <chrono>
 #include <cstdint>
 #include <exception>
 #include <filesystem>
 #include <initializer_list>
 #include <iosfwd>
+#include <limits>
+#include <map>
+#include <memory>
+#include <string.h>
 #include <unordered_set>
+#include <utility>
 #include <variant>
-#include "segcore/default_fs.h"
+#include <vector>
 
 #include "NamedType/named_type_impl.hpp"
 #include "NamedType/underlying_functionalities.hpp"
@@ -40,6 +45,9 @@
 #include "common/Tracer.h"
 #include "common/jsmn.h"
 #include "fmt/core.h"
+#include "folly/ScopeGuard.h"
+#include "folly/futures/Future.h"
+#include "futures/Executor.h"
 #include "index/Utils.h"
 #include "index/json_stats/JsonKeyStats.h"
 #include "index/json_stats/bson_builder.h"
@@ -54,21 +62,22 @@
 #include "milvus-storage/reader.h"
 #include "mmap/ChunkedColumnGroup.h"
 #include "mmap/Types.h"
+#include "nlohmann/json.hpp"
 #include "nlohmann/detail/iterators/iteration_proxy.hpp"
 #include "nlohmann/json_fwd.hpp"
 #include "parquet/metadata.h"
-#include "segcore/storagev1translator/BsonInvertedIndexTranslator.h"
 #include "segcore/ChunkedSegmentSealedImpl.h"
 #include "segcore/storagev2translator/ManifestGroupTranslator.h"
 #include "segcore/Utils.h"
+#include "segcore/default_fs.h"
+#include "segcore/storagev1translator/BsonInvertedIndexTranslator.h"
 #include "storage/DiskFileManagerImpl.h"
+#include "storage/EntryStreamUtils.h"
 #include "storage/FileManager.h"
 #include "storage/LocalChunkManager.h"
 #include "storage/LocalChunkManagerSingleton.h"
 #include "storage/MemFileManagerImpl.h"
 #include "storage/MmapManager.h"
-#include "storage/Util.h"
-#include "folly/ScopeGuard.h"
 #include "storage/ThreadPools.h"
 #include "storage/Types.h"
 #include "storage/Util.h"
@@ -77,6 +86,105 @@
 namespace milvus::index {
 
 namespace {
+
+constexpr int64_t kJsonStatsRowsPerRange = 16 * 1024;
+constexpr size_t kJsonStatsMaterializeMinReservationBytes = 64 * 1024;
+constexpr size_t kJsonStatsMaterializeInputExpansionFactor = 2;
+constexpr size_t kJsonStatsMaterializeRowOverheadBytes = 32;
+constexpr size_t kJsonStatsMaterializeColumnOverheadBytes = 1024;
+
+struct CollectKeyInfoResult {
+    uint64_t sequence;
+    int64_t row_count;
+    std::map<JsonKey, KeyStatsInfo> infos;
+};
+
+template <typename Func>
+void
+ForEachJsonStatsRow(const JsonStatsRowRange& range, Func&& func) {
+    int64_t visited_rows = 0;
+    for (const auto& slice : range.slices) {
+        AssertInfo(slice.data != nullptr,
+                   "json stats row range contains null field data");
+        AssertInfo(slice.row_count > 0,
+                   "json stats row range contains an empty slice");
+        AssertInfo(slice.local_begin >= 0,
+                   "json stats field data slice has negative local begin: {}",
+                   slice.local_begin);
+        AssertInfo(
+            slice.local_begin <= slice.data->get_num_rows() - slice.row_count,
+            "json stats field data slice [{}, {}) exceeds field data rows {}",
+            slice.local_begin,
+            slice.local_begin + slice.row_count,
+            slice.data->get_num_rows());
+        AssertInfo(
+            slice.global_begin == range.global_begin + visited_rows,
+            "json stats field data slice starts at global row {}, expected {}",
+            slice.global_begin,
+            range.global_begin + visited_rows);
+
+        for (int64_t offset = 0; offset < slice.row_count; ++offset) {
+            func(slice.data,
+                 slice.local_begin + offset,
+                 slice.global_begin + offset);
+        }
+        visited_rows += slice.row_count;
+    }
+    AssertInfo(visited_rows == range.row_count,
+               "json stats row range contains {} rows, expected {}",
+               visited_rows,
+               range.row_count);
+}
+
+size_t
+SaturatingAdd(size_t lhs, size_t rhs) {
+    if (rhs > std::numeric_limits<size_t>::max() - lhs) {
+        return std::numeric_limits<size_t>::max();
+    }
+    return lhs + rhs;
+}
+
+size_t
+SaturatingMultiply(size_t lhs, size_t rhs) {
+    if (lhs != 0 && rhs > std::numeric_limits<size_t>::max() / lhs) {
+        return std::numeric_limits<size_t>::max();
+    }
+    return lhs * rhs;
+}
+
+size_t
+EstimateJsonStatsMaterializeReservationBytes(const JsonStatsRowRange& range,
+                                             size_t column_count) {
+    size_t input_bytes = 0;
+    ForEachJsonStatsRow(
+        range,
+        [&](const FieldDataPtr& data,
+            int64_t local_row,
+            int64_t /* global_row */) {
+            if (data->IsNullable() && !data->is_valid(local_row)) {
+                return;
+            }
+            auto raw_value = data->RawValue(local_row);
+            if (raw_value == nullptr) {
+                return;
+            }
+            const auto* json = static_cast<const milvus::Json*>(raw_value);
+            input_bytes = SaturatingAdd(input_bytes, json->data().size());
+        });
+
+    auto reservation_bytes = SaturatingMultiply(
+        input_bytes, kJsonStatsMaterializeInputExpansionFactor);
+    reservation_bytes = SaturatingAdd(
+        reservation_bytes,
+        SaturatingMultiply(static_cast<size_t>(range.row_count),
+                           kJsonStatsMaterializeRowOverheadBytes));
+    reservation_bytes = SaturatingAdd(
+        reservation_bytes,
+        SaturatingMultiply(column_count,
+                           kJsonStatsMaterializeColumnOverheadBytes));
+    return std::max(kJsonStatsMaterializeMinReservationBytes,
+                    reservation_bytes);
+}
 
 struct JsonStatsParquetMetadata {
     std::shared_ptr<arrow::Schema> schema;
@@ -162,6 +270,60 @@ GetJsonStatsFieldsFromSchema(const std::shared_ptr<arrow::Schema>& schema) {
 }
 
 }  // namespace
+
+std::vector<JsonStatsRowRange>
+CreateJsonStatsRowRanges(const std::vector<FieldDataPtr>& field_datas,
+                         int64_t max_rows_per_range) {
+    AssertInfo(max_rows_per_range > 0,
+               "json stats max rows per range must be positive, got {}",
+               max_rows_per_range);
+
+    std::vector<JsonStatsRowRange> row_ranges;
+    JsonStatsRowRange current_range{};
+    int64_t global_begin = 0;
+
+    for (const auto& data : field_datas) {
+        AssertInfo(data != nullptr,
+                   "cannot create json stats row range from null field data");
+        const auto data_rows = data->get_num_rows();
+        AssertInfo(data_rows >= 0,
+                   "json stats field data has negative row count: {}",
+                   data_rows);
+
+        int64_t local_begin = 0;
+        while (local_begin < data_rows) {
+            if (current_range.row_count == 0) {
+                current_range.sequence =
+                    static_cast<uint64_t>(row_ranges.size());
+                current_range.global_begin = global_begin + local_begin;
+            }
+
+            const auto remaining_rows =
+                max_rows_per_range - current_range.row_count;
+            const auto slice_rows =
+                std::min(data_rows - local_begin, remaining_rows);
+            current_range.slices.push_back(JsonStatsFieldDataSlice{
+                data,
+                local_begin,
+                slice_rows,
+                global_begin + local_begin,
+            });
+            current_range.row_count += slice_rows;
+            local_begin += slice_rows;
+
+            if (current_range.row_count == max_rows_per_range) {
+                row_ranges.push_back(std::move(current_range));
+                current_range = JsonStatsRowRange{};
+            }
+        }
+        global_begin += data_rows;
+    }
+
+    if (current_range.row_count > 0) {
+        row_ranges.push_back(std::move(current_range));
+    }
+    return row_ranges;
+}
 
 JsonKeyStats::JsonKeyStats(const storage::FileManagerContext& ctx,
                            bool is_load,
@@ -389,6 +551,125 @@ JsonKeyStats::CollectKeyInfo(const std::vector<FieldDataPtr>& field_datas,
     return infos;
 }
 
+std::map<JsonKey, KeyStatsInfo>
+JsonKeyStats::CollectKeyInfo(const std::vector<JsonStatsRowRange>& row_ranges,
+                             bool nullable) {
+    std::map<JsonKey, KeyStatsInfo> infos;
+    int64_t expected_global_begin = 0;
+    uint64_t expected_sequence = 0;
+    for (const auto& range : row_ranges) {
+        AssertInfo(range.sequence == expected_sequence,
+                   "json stats row range sequence is {}, expected {}",
+                   range.sequence,
+                   expected_sequence);
+        AssertInfo(range.global_begin == expected_global_begin,
+                   "json stats row range starts at global row {}, expected {}",
+                   range.global_begin,
+                   expected_global_begin);
+        expected_global_begin += range.row_count;
+        ++expected_sequence;
+    }
+
+    if (row_ranges.empty()) {
+        num_rows_ = 0;
+        return infos;
+    }
+
+    auto* executor = milvus::futures::getJsonStatsBuildExecutor();
+    const auto max_inflight = std::min(
+        row_ranges.size(), std::max<size_t>(1, executor->numThreads()));
+
+    // Futures are kept in sequence order. Only max_inflight tasks are active
+    // at a time, so completed local maps cannot accumulate for every range.
+    std::vector<folly::Future<CollectKeyInfoResult>> futures;
+    futures.reserve(row_ranges.size());
+    size_t next_submit = 0;
+    size_t next_consume = 0;
+    std::exception_ptr first_exception;
+
+    auto submit_next = [&]() {
+        try {
+            auto range = row_ranges[next_submit];
+            auto future = folly::via(
+                executor, [this, range = std::move(range), nullable]() {
+                    std::map<JsonKey, KeyStatsInfo> local_infos;
+                    ForEachJsonStatsRow(
+                        range,
+                        [&](const FieldDataPtr& data,
+                            int64_t local_row,
+                            int64_t /* global_row */) {
+                            if ((nullable || data->IsNullable()) &&
+                                !data->is_valid(local_row)) {
+                                return;
+                            }
+                            auto json_str = static_cast<const milvus::Json*>(
+                                                data->RawValue(local_row))
+                                                ->data()
+                                                .data();
+                            CollectSingleJsonStatsInfo(json_str, local_infos);
+                        });
+                    return CollectKeyInfoResult{
+                        range.sequence,
+                        range.row_count,
+                        std::move(local_infos),
+                    };
+                });
+            futures.emplace_back(std::move(future));
+            ++next_submit;
+        } catch (...) {
+            if (first_exception == nullptr) {
+                first_exception = std::current_exception();
+            }
+        }
+    };
+
+    while (next_submit < max_inflight && first_exception == nullptr) {
+        submit_next();
+    }
+
+    int64_t num_rows = 0;
+    uint64_t next_result_sequence = 0;
+    while (next_consume < futures.size()) {
+        try {
+            auto result = std::move(futures[next_consume]).get();
+            if (first_exception == nullptr) {
+                AssertInfo(result.sequence == next_result_sequence,
+                           "json stats collect result sequence is {}, expected "
+                           "{}",
+                           result.sequence,
+                           next_result_sequence);
+                for (const auto& [key, local_info] : result.infos) {
+                    infos[key].hit_row_num_ += local_info.hit_row_num_;
+                }
+                num_rows += result.row_count;
+                ++next_result_sequence;
+            }
+        } catch (...) {
+            if (first_exception == nullptr) {
+                first_exception = std::current_exception();
+            }
+        }
+        ++next_consume;
+
+        if (first_exception == nullptr && next_submit < row_ranges.size()) {
+            submit_next();
+        }
+    }
+
+    // Every submitted future has been consumed before an error escapes. This
+    // keeps `this` and the FieldData referenced by worker lambdas alive.
+    if (first_exception != nullptr) {
+        std::rethrow_exception(first_exception);
+    }
+
+    AssertInfo(num_rows == expected_global_begin,
+               "collected {} json stats rows, expected {}",
+               num_rows,
+               expected_global_begin);
+    num_rows_ = num_rows;
+    return infos;
+}
+
 std::map<JsonKey, JsonKeyLayoutType>
 JsonKeyStats::ClassifyJsonKeyLayoutType(
     const std::map<JsonKey, KeyStatsInfo>& infos) {
@@ -495,7 +776,7 @@ void
 JsonKeyStats::AddKeyStats(const std::vector<std::string>& path,
                           JSONType type,
                           const std::string& value,
-                          std::map<JsonKey, std::string>& values) {
+                          std::map<JsonKey, std::string>& values) const {
     auto path_str = JsonPointer(path);
     auto key = JsonKey(path_str, type);
     values[key] = value;
@@ -507,7 +788,7 @@ JsonKeyStats::TraverseJsonForBuildStats(
     jsmntok* tokens,
     int& index,
     std::vector<std::string>& path,
-    std::map<JsonKey, std::string>& values) {
+    std::map<JsonKey, std::string>& values) const {
     jsmntok current = tokens[0];
     AssertInfo(current.type != JSMN_UNDEFINED,
                "current token type is undefined for json: {}",
@@ -597,26 +878,9 @@ JsonKeyStats::TraverseJsonForBuildStats(
     }
 }
 
-void
-JsonKeyStats::BuildKeyStatsForNullRow() {
-    // add empty value for column keys that not hit
-    for (const auto& key : column_keys_) {
-        parquet_writer_->AppendValue(key.ToColumnName(), "");
-    }
-
-    // add null bson to shared column
-    BsonDocument null_doc;
-    parquet_writer_->AppendSharedRow(null_doc.data(), null_doc.length());
-
-    parquet_writer_->AddCurrentRow();
-}
-
-void
-JsonKeyStats::BuildKeyStatsForRow(const char* json_str, uint32_t row_id) {
-    LOG_TRACE("build key stats for row {} with json {} for segment {}",
-              row_id,
-              json_str,
-              segment_id_);
+bool
+JsonKeyStats::ParseJsonForBuildStats(
+    const char* json_str, std::map<JsonKey, std::string>& values) const {
     jsmn_parser parser;
     jsmn_init(&parser);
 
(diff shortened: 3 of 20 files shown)


internal/core/src/storage/EntryStreamUtils.h
```diff
@@ -88,11 +88,22 @@ class TransientMemoryBudget {
...
+    }
+
     static void
     SetLoadTransientBudgetBytes(size_t bytes) {
         GetLoadTransientBudget().SetCapacityBytes(bytes);
     }
 
+    static void
+    SetJsonStatsBuildBudgetBytes(size_t bytes) {
+        GetJsonStatsBuildBudget().SetCapacityBytes(bytes);
```
buqian-zilliz, line 107: **[P1] Keep the JSON budget from overwriting load admission**

The new JSON-stats budget is routed through `SetCapacityBytes`, whose implementation also updates the singleton `LoadMemoryOverheadController`. `InitSegcore` applies `common.loadTransientBudgetBytes` first and then the default 512 MiB JSON budget, so startup silently replaces the admission policy for unrelated segment loads; later hot reloads of either setting remain last-writer-wins. Keep the load overhead-controller update specific to the load budget and update only the JSON budget's own capacity here.

internal/core/src/index/json_stats/bson_inverted.cpp
```diff
@@ -76,6 +76,23 @@ BsonInvertedIndex::AddRecord(const std::string& key,
...
+        return;
+    }
+
+    auto& destination = inverted_index_map_[key];
+    if (destination.empty()) {
+        destination = std::move(records);
+        return;
+    }
+
+    destination.reserve(destination.size() + records.size());
```
buqian-zilliz: **[P1] Avoid quadratic growth when merging BSON postings**

Reserving exactly `destination.size() + records.size()` on every 16K-row chunk merge defeats the vector's geometric growth and repeatedly copies the entire accumulated postings prefix. For a key present in every range, 10 million rows produce about 611 chunks and more than 3 billion prior `int64_t` copies for that one key. Remove the exact reserve, grow geometrically, or reserve the final count once.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -706,31 +996,437 @@ JsonKeyStats::BuildKeyStatsForRow(const char* json_str, uint32_t row_id) {
...
+        try {
+            range = row_ranges[next_submit];
+            reservation_bytes = EstimateJsonStatsMaterializeReservationBytes(
+                range, static_cast<size_t>(schema->num_fields()));
+        } catch (...) {
+            remember_exception(std::current_exception());
+            return false;
+        }
+
+        auto accounted_bytes =
```
buqian-zilliz: **[P2] Drain active tasks if bookkeeping allocation fails**

This allocation is outside `submit_next`'s exception guard. If it throws while an earlier refill task is active, `BuildKeyStatsParallel` unwinds without draining that future or releasing its memory reservation. The global budget can remain permanently charged, and the worker still captures `this`, so destruction of the failed build object can race worker access. Put all potentially throwing submission setup under guarded cleanup that always drains and releases already-submitted tasks before propagating an error.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -840,8 +1536,10 @@ JsonKeyStats::AddBucketName(const std::string& remote_prefix) {
 void
 JsonKeyStats::BuildWithFieldData(const std::vector<FieldDataPtr>& field_datas,
                                  bool nullable) {
+    const auto row_ranges =
+        CreateJsonStatsRowRanges(field_datas, kJsonStatsRowsPerRange);
     // collect key stats info and classify key type
-    auto infos = CollectKeyInfo(field_datas, nullable);
+    auto infos = CollectKeyInfo(row_ranges, nullable);
```
buqian-zilliz: **[P2] Use the serial path when parallelism is unavailable**

The production path always selects the future-based chunk pipeline, including when there is only one row range or the effective executor limit is one. Those supported cases gain no concurrency but still pay for local-map merging, a full reservation scan, future handoff, temporary Arrow arrays, and `AppendArraySlice` copies back into the writer builders. Select the retained serial collect/materialize path when effective concurrency is at most one.

internal/core/src/storage/EntryStreamUtils.h
zhagnlu: fixed

internal/core/src/index/json_stats/bson_inverted.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -78,6 +87,105 @@ namespace milvus::index {
...
+
+    auto reservation_bytes = SaturatingMultiply(
+        input_bytes, kJsonStatsMaterializeInputExpansionFactor);
+    reservation_bytes = SaturatingAdd(
+        reservation_bytes,
+        SaturatingMultiply(static_cast<size_t>(range.row_count),
+                           kJsonStatsMaterializeRowOverheadBytes));
+    reservation_bytes = SaturatingAdd(
+        reservation_bytes,
+        SaturatingMultiply(column_count,
```
buqian-zilliz: **[P1] Account for Arrow column buffers before admitting workers**

`maxInflightBytes` is enforced against this estimate before dispatch, but the estimate charges only 1 KiB per output column instead of the buffers that `MaterializeKeyStatsRange` must allocate for every row/column pair. Arrow numeric builders append a zero value even for nulls, so a valid sparse 16,384-row range with 1,024 shredded `int64` columns reserves about 1.5 MiB while allocating at least 130 MiB; four such ranges can be admitted concurrently and exceed 520 MiB before reconciliation. With a lowered budget, the overshoot is even larger relative to the configured cap, and none of these chunks runs exclusively because each looked small at admission time. Reserve a conservative row-by-column bound or split ranges to fit before scheduling.

internal/core/src/index/json_stats/parquet_writer.cpp
```diff
@@ -199,15 +200,75 @@ JsonStatsParquetWriter::AppendRow(
...
+        auto result = WriteCurrentBatch();
+        if (!result.ok()) {
+            return result.status();
+        }
+    }
+    int64_t offset = 0;
+    while (offset < batch->num_rows()) {
+        const auto available_rows = target_batch_size - unflushed_row_count_;
+        const auto rows_to_append = std::min<int64_t>(
+            static_cast<int64_t>(available_rows), batch->num_rows() - offset);
```
buqian-zilliz: **[P2] Reject negative batch sizes before the signed conversion**

`dataCoord.jsonShreddingWriteBatchSize` accepts a signed `int64` value without lower-bound validation and is passed into the writer's `size_t batch_size_`. A configured value such as `-1` therefore makes `target_batch_size` equal to `SIZE_MAX`; on the supported 64-bit builds, converting `available_rows` back to `int64_t` produces a negative `rows_to_append`, so `AppendArraySlice` fails and multi-range JSON stats builds abort. Validate the configuration before converting it to `size_t`, or keep this arithmetic in a checked signed type.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -712,29 +1001,441 @@ JsonKeyStats::BuildKeyStatsForRow(std::string_view json_str, uint32_t row_id) {
...
+
+JsonKeyStats::MaterializedChunk
+JsonKeyStats::MaterializeKeyStatsRange(
+    const JsonStatsRowRange& range,
+    bool nullable,
+    const std::shared_ptr<arrow::Schema>& schema) const {
+    AssertInfo(schema != nullptr,
+               "json stats materialize schema must not be null");
+
+    auto [builders, builders_map] = CreateArrowBuilders(key_types_);
```
buqian-zilliz, line 1098: **[P2] Reserve the per-range Arrow builders before filling them**

`range.row_count` is fixed when these fresh builders are created, but none of them is reserved before the per-row append loop. Every `Append` or `AppendNull` therefore grows each builder geometrically from empty; with 1,024 columns, 16,384-row ranges, and four workers, this causes tens of thousands of allocation-and-copy growth steps per wave and repeats for every range. Call `Reserve(range.row_count)` on each builder once, and reserve variable-width data where a byte estimate is available.

internal/core/src/index/json_stats/parquet_writer.cpp
```diff
@@ -199,15 +200,75 @@ JsonStatsParquetWriter::AppendRow(
...
+
+        for (int column = 0; column < batch->num_columns(); ++column) {
+            arrow::ArraySpan array_span(*batch->column(column)->data());
+            auto status = builders_[column]->AppendArraySlice(
```
buqian-zilliz: **[P2] Avoid rematerializing every worker batch in the writer**

`MaterializeKeyStatsRange` has already produced complete Arrow arrays, but this loop copies every column into writer-owned builders, which `WriteCurrentBatch` then finishes into a second set of arrays. An N-byte JSON-stats output therefore incurs another O(N) pass of single-threaded memory traffic, while the writer can retain up to `batch_size_` copied rows in addition to the worker batches covered by the transient budget; with the defaults that is five 16,384-row ranges. Feed record-batch slices directly to the packed writer, or aggregate them without rebuilding every array.

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: > Thanks for pointing this out. We now call Reserve(range.row_count) on every fresh Arrow builder before the append loop, which avoids repeated geometric growth for the validity, offset, and fixed-width buffers. Variable-width payload reservation would require a
  > reliable per-column byte estimate; using the total JSON input size for every variable-width builder would significantly over-allocate memory, so that part is not applied blindly.

internal/core/src/index/json_stats/parquet_writer.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/parquet_writer.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -712,29 +1042,450 @@ JsonKeyStats::BuildKeyStatsForRow(std::string_view json_str, uint32_t row_id) {
...
+                    bson_inverted_index_->AddRecords(key, std::move(postings));
+                }
+
+                processed_rows += result.row_count;
+                ++next_result_sequence;
+            }
+        } catch (...) {
+            remember_exception(std::current_exception());
+        }
+        release_task_budget(task);
```
buqian-zilliz, line 1464: **[P1] Keep writer-retained Arrow buffers inside the memory budget**

The reservation is released immediately after `AppendRecordBatch`, but the pinned packed/parquet writer caches zero-copy `RecordBatch` slices. Even after a flush, a small remainder slice can retain the entire Arrow backing buffer. A 16,384-row range with 1,024 fixed-width columns is about 130 MiB, so every concurrent build can retain another full chunk outside the process-wide budget while refill admits the next budgeted wave. This defeats `maxInflightBytes` and can still OOM the DataNode. Keep the reservation until downstream releases the buffers, or use a bounded handoff that does not retain the worker-owned batch.

internal/core/src/futures/future_c.cpp
```diff
@@ -69,3 +72,26 @@ executor_set_load_thread_num(int thread_num) {
     LOG_INFO("future executor setup load cpu executor with thread num: {}",
              thread_num);
 }
+
+extern "C" void
+executor_set_json_stats_build_thread_num(int thread_num) {
+    auto normalized_thread_num = std::max(1, thread_num);
+    milvus::futures::getJsonStatsBuildExecutor()->setNumThreads(
```
buqian-zilliz, line 79: **[P1] Prevent executor resize exceptions from crossing the cgo boundary**

Folly's `setNumThreads` can throw while increasing the pool if worker-thread creation fails. This new hot-reload entry point is an `extern "C"` void function without an exception guard, so such an exception crosses the cgo boundary and terminates the DataNode instead of retaining the previous pool size. Catch the exception here and expose an error to Go, or return a status that the watcher can log safely.

internal/core/src/index/json_stats/parquet_writer.cpp
```diff
@@ -199,15 +200,81 @@ JsonStatsParquetWriter::AppendRow(
...
+    // them straight through so the writer does not copy every value into a
+    // second set of builders merely to finish those builders back into arrays.
+    // Keep batch_size_ as an upper bound for a single write; Slice shares the
+    // original buffers and therefore remains zero-copy.
+    const auto max_batch_rows = static_cast<int64_t>(batch_size_);
+    int64_t offset = 0;
+    while (offset < batch->num_rows()) {
+        const auto rows_to_write =
+            std::min(max_batch_rows, batch->num_rows() - offset);
+        auto slice = offset == 0 && rows_to_write == batch->num_rows()
```
buqian-zilliz, line 259: **[P2] Avoid slicing batches while accounting their full backing buffers**

When `jsonShreddingWriteBatchSize` is below the fixed 16,384-row materialization range, `Slice` retains the original full-range buffers. The pinned packed writer measures each slice from those complete buffers, and its parquet writer derives average row size from that value. For example, a 1,024-row slice is charged like the full 16,384-row batch, causing premature flushes and row groups roughly one sixteenth of the intended physical size. Cap materialization ranges by `batch_size_`, compact the slices before handoff, or make downstream accounting offset-aware.

internal/core/src/index/json_stats/JsonKeyStats.h
```diff
@@ -68,11 +68,32 @@
...
+
+struct JsonStatsFieldDataSlice {
+    FieldDataPtr data;
+    int64_t local_begin;
+    int64_t row_count;
+    int64_t global_begin;
+};
+
+struct JsonStatsRowRange {
+    uint64_t sequence;
```
buqian-zilliz, line 87: **[P3] Remove duplicated range ordering state**

`row_ranges` are emitted into a vector and their futures are submitted and consumed in that same order, so `sequence` is always the vector index. The global start is also duplicated across each range, every slice, and each materialized result, creating representable inconsistent states and requiring several validation loops and counters. Keep `JsonStatsRowRange::global_begin`, derive slice row IDs from the visited offset, and identify results by their task index.

internal/core/src/index/json_stats/JsonKeyStats.cpp
buqian-zilliz: This issue still exists: fixed-width Arrow buffers are now charged, but variable-width output is still assumed to fit within twice the raw JSON size. Compact numeric arrays can expand to BSON by substantially more than that, so several ranges can pass admission and exceed the budget before reconciliation. Please include a conservative BSON expansion bound or adapt range sizes before dispatch.

internal/core/src/index/json_stats/parquet_writer.cpp
buqian-zilliz: This issue still exists at the native request boundary: the formatter normalizes requests created by an upgraded DataCoord, but a base-version DataCoord can still send zero or a negative value to a new DataNode. A zero stride prevents `offset` from advancing, while a negative value wraps through `size_t`. Normalize or reject the value in `JsonKeyStats` or `JsonStatsParquetWriter`.

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.h
zhagnlu: fixed

internal/core/src/futures/future_c.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -446,6 +640,100 @@ JsonKeyStats::CollectKeyInfo(const std::vector<FieldDataPtr>& field_datas,
...
+    size_t next_submit = 0;
+    size_t next_consume = 0;
+    std::exception_ptr first_exception;
+
+    auto submit_next = [&]() {
+        try {
+            auto range = row_ranges[next_submit];
+            auto future = folly::via(
+                executor, [this, range = std::move(range), nullable]() {
+                    std::map<JsonKey, KeyStatsInfo> local_infos;
```
buqian-zilliz, line 657: **[P1] Bound the parallel key-collection maps**

`maxInflightBytes` only gates materialization, while this phase creates one unbudgeted `std::map<JsonKey, KeyStatsInfo>` per active range. The shredding-column cap is applied only after collection, so the default eight workers can retain eight large maps in completed futures in addition to the merged map; concurrent builds multiply that process-wide footprint and can OOM even when operators lower the configured budget. Charge these local maps to a shared or phase-specific budget, or merge into bounded shards instead of retaining one full map per range.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -933,7 +1650,20 @@ JsonKeyStats::BuildWithFieldData(const std::vector<FieldDataPtr>& field_datas,
...
     parquet_writer_->Init(std::move(writer_context));
-    BuildKeyStats(field_datas, nullable);
+    if (use_parallel_pipeline) {
+        // Arrow slices retain the complete backing buffers. Keep every
+        // materialized range within the writer batch limit so
+        // AppendRecordBatch can hand it off without slicing; otherwise
+        // downstream memory accounting would charge a small slice for the
+        // full range and undersize row groups.
+        const auto materialize_rows_per_range =
+            std::min(kJsonStatsRowsPerRange, write_batch_size_);
```
buqian-zilliz, line 1643: **[P1] Avoid one range allocation per row for small batches**

`jsonShreddingWriteBatchSize` accepts every positive value, and this now uses it directly as the eager materialization-range size. With a valid value such as `1`, `CreateJsonStatsRowRanges` allocates one heap-backed `JsonStatsRowRange` per row, then `BuildKeyStatsParallel` reserves `tasks` for the full range count before submitting work. On a production segment this metadata alone can consume hundreds of MB outside `maxInflightBytes` and fail the build before the budget can throttle it. Generate ranges and task slots through a bounded active window, or enforce a practical minimum without materializing row-count-sized vectors upfront.

internal/core/src/index/json_stats/parquet_writer.cpp
```diff
@@ -224,7 +225,73 @@ JsonStatsParquetWriter::AppendRow(
...
+    }
+
+    if (batch->num_rows() == 0) {
+        return arrow::Status::OK();
+    }
+
+    // Preserve row order if a caller switches from the row-oriented API to
+    // pre-materialized record batches. The parallel JSON-stats path does not
+    // normally have pending builder rows.
+    if (unflushed_row_count_ > 0) {
```
buqian-zilliz, line 260: **[P3] Remove unreachable mixed-mode batch handling**

`AppendRecordBatch` has only one production caller, immediately after `Init`, and `BuildWithFieldData` now caps every materialized range at `batch_size_`. The pending row-builder flush and slicing loop therefore support states that caller cannot produce, adding a second write mode and retaining unnecessary zero-copy slice handling. Assert those existing preconditions and call `packed_writer_->Write(batch)` once.

internal/core/src/index/json_stats/JsonKeyStats.cpp
buqian-zilliz: This issue still exists: capping the range removes slicing, but the packed writer still retains the newest full record batch until a later flush or close while `BuildKeyStatsParallel` releases its reservation immediately.

internal/core/src/index/json_stats/JsonKeyStats.cpp
buqian-zilliz: This issue still exists: fixed-width Arrow buffers are now counted, but variable-width BSON output is still admitted with only a two-times raw-input estimate, which compact arrays can exceed before reconciliation.

internal/core/src/index/json_stats/parquet_writer.cpp
buqian-zilliz: This issue still exists: the DataCoord formatter normalizes same-version configuration, but the native constructor still accepts the signed request value unchanged, so rolling-upgrade requests can still pass `0` or a negative batch size.

internal/core/src/index/json_stats/parquet_writer.cpp
zhagnlu: ignore, not bug

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: In practice, jsonShreddingWriteBatchSize is typically above 16K (the default is 81920), so materialization still uses 16K-row ranges. Although smaller values are currently accepted, we don’t expect such tiny batch sizes in production.

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: For reference, with 32-byte paths (~144 B per distinct path/type entry, measured with x86-64 libstdc++ and jemalloc), 100k distinct keys in each of the eight local maps plus the merged map consume ~123.6 MiB total per build (~110 MiB additional over the merged map alone). At this cardinality, the map footprint looks manageable.

This estimate covers map allocations for one build; longer paths, higher key cardinality, or concurrent builds would increase the memory requirement.

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: > This appears to be a generic ownership and memory-accounting issue in the milvus-storage packed/parquet writer, rather than something specific to JSON Stats.
  >
  > PackedRecordBatchWriter::Write() splits columns zero-copy, ParquetFileWriter::Write() caches the resulting record batches, and Flush() may retain zero-copy remainder slices. Therefore, Write() can return while the writer still references the caller-owned Arrow
  > backing buffers. A small remainder slice may pin the entire input batch.
  >
  > Meanwhile, the packed writer removes the flushed group from current_memory_usage_, so its accounting can diverge from the physically retained memory.
  >
  > Could this be fixed at the milvus-storage ownership boundary, either by providing a detached/consumed write that compact-copies any retained remainder, or by exposing the retained-buffer lifetime and actual retained bytes to callers? JSON Stats could then release
  > its reservation only after the ownership handoff is complete. 
  > a common problem not only json stats, not important problem. next pr optimize it

internal/core/src/index/json_stats/parquet_writer.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: > This appears to be a generic ownership and memory-accounting issue in the milvus-storage packed/parquet writer, rather than something specific to JSON Stats.
  >
  > PackedRecordBatchWriter::Write() splits columns zero-copy, ParquetFileWriter::Write() caches the resulting record batches, and Flush() may retain zero-copy remainder slices. Therefore, Write() can return while the writer still references the caller-owned Arrow
  > backing buffers. A small remainder slice may pin the entire input batch.
  >
  > Meanwhile, the packed writer removes the flushed group from current_memory_usage_, so its accounting can diverge from the physically retained memory.
  >
  > Could this be fixed at the milvus-storage ownership boundary, either by providing a detached/consumed write that compact-copies any retained remainder, or by exposing the retained-buffer lifetime and actual retained bytes to callers? JSON Stats could then release
  > its reservation only after the ownership handoff is complete. 
  > a common problem not only json stats, not important problem. next pr optimize it

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: For this PR, I'd like to defer refining the BSON expansion estimate. `maxInflightBytes` is a soft admission budget for concurrent materialization. Predicting the full peak precisely is difficult because parser/DOM temporaries, intermediate copies, Arrow buffer growth, and allocator overhead also contribute; improving the BSON estimate alone would not make it an exact memory cap.

The current reconciliation updates the accounting after each chunk completes to constrain subsequent admission. Already-running chunks can temporarily exceed their reservations, so transient peaks remain possible. Let's keep the current approach for now and revisit tighter estimation and range sizing separately.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -771,29 +1072,431 @@ JsonKeyStats::BuildKeyStatsForRow(std::string_view json_str, uint32_t row_id) {
...
+
+JsonKeyStats::MaterializedChunk
+JsonKeyStats::MaterializeKeyStatsRange(
+    const JsonStatsRowRange& range,
+    bool nullable,
+    const std::shared_ptr<arrow::Schema>& schema) const {
+    AssertInfo(schema != nullptr,
+               "json stats materialize schema must not be null");
+
+    auto [builders, builders_map] = CreateArrowBuilders(key_types_);
```
buqian-zilliz: **[P1] Avoid rebuilding all shared-key entries for every range**

`CreateArrowBuilders(key_types_)` creates a fresh `builders_map` entry for every classified key, including every `SHARED` path, even though `MaterializeKeyStatsRange` never consults that map for shared keys. `max_shredding_columns_` caps only non-shared columns, so high-cardinality JSON can make each of the default eight workers duplicate an unbounded node-based map; these temporary maps are omitted from both the admission estimate and `MaterializedChunk::memory_bytes`, and the work repeats for every range. This can exhaust memory despite `maxInflightBytes` and turns materialization setup into `O(ranges * all_keys)`. Build task-local mappings only for `column_keys_` plus the single shared builder, or share immutable lookup metadata across ranges.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -933,7 +1658,20 @@ JsonKeyStats::BuildWithFieldData(const std::vector<FieldDataPtr>& field_datas,
...
     parquet_writer_->Init(std::move(writer_context));
-    BuildKeyStats(field_datas, nullable);
+    if (use_parallel_pipeline) {
+        // Arrow slices retain the complete backing buffers. Keep every
+        // materialized range within the writer batch limit so
+        // AppendRecordBatch can hand it off without slicing; otherwise
+        // downstream memory accounting would charge a small slice for the
+        // full range and undersize row groups.
+        const auto materialize_rows_per_range =
+            std::min(kJsonStatsRowsPerRange, write_batch_size_);
```
buqian-zilliz, line 1643: **[P2] Honor write batch sizes above 16K in the parallel path**

The parallel branch sets each materialization range to `min(16384, write_batch_size_)` and sends every range directly to `AppendRecordBatch`, which immediately invokes `PackedRecordBatchWriter::Write`. Consequently the default `jsonShreddingWriteBatchSize` of 81,920, and every configured value above 16,384, no longer controls the write batch once parallelism is selected: the default path performs five splitter and writer submissions for the rows the serial path submits once. This changes the existing configuration contract and multiplies downstream split, heap, and Parquet flush bookkeeping. Accumulate ordered materialized chunks up to `write_batch_size_`, or add a separate materialization-chunk setting instead of silently clamping the write batch.

internal/core/src/futures/future_c.cpp
```diff
@@ -82,3 +87,39 @@ executor_set_load_thread_num(int thread_num) {
...
+    auto normalized_thread_num = std::max(1, thread_num);
+    try {
+        milvus::futures::getJsonStatsBuildExecutor()->setNumThreads(
+            normalized_thread_num);
+        LOG_INFO(
+            "future executor setup json stats build cpu executor with thread "
+            "num: {}",
+            normalized_thread_num);
+        return milvus::SuccessCStatus();
+    } catch (const std::exception& e) {
```
buqian-zilliz: **[P2] Preserve OOM classification when resizing the executor**

`std::bad_alloc` derives from `std::exception`, so an allocation failure from `setNumThreads` is passed to `FailureCStatus(&e)` and becomes `UnexpectedError`; the shared `CGO_CATCH_AND_RETURN_CSTATUS` catch tail explicitly maps it to `MemAllocateFailed`. This setter is used during startup and hot reload, so resource exhaustion loses the established memory-error classification at the C/Go boundary. Use the shared catch tail, or catch `std::bad_alloc` before `std::exception`, as the memory-budget setter already does.

zhagnlu: fixed

zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: fixed

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: jsonShreddingWriteBatchSize controls the upstream RecordBatch size. Downstream flushing is byte-based: the packed writer uses a 16 MiB buffer, and the Parquet writer builds ~1 MiB row groups across batches.

  The extra submissions do add splitter/heap bookkeeping, but coalescing already-materialized Arrow chunks can add copying and increase peak memory. We would need benchmarks to establish whether that tradeoff is beneficial.

  The configuration behavior change above 16K is real. I’d prefer to clarify that the setting is an upper bound in the parallel path rather than add a concatenation step solely to preserve the previous submission size.

ignore it

internal/core/src/exec/expression/ExprCache.cpp
```diff
@@ -388,37 +693,46 @@ ExprResCacheManager::Put(const Key& key, const Value& value) {
...
-            file = std::make_unique<DiskSlotFile>(
+        if (file_it == disk_files_.end() || !file_it->second) {
+            auto new_file = std::make_unique<DiskSlotFile>(
                 key.segment_id,
                 path,
                 static_cast<int64_t>(value.result->size()),
                 config_.disk_max_file_size);
+            if (file_it == disk_files_.end()) {
+                file_it =
+                    disk_files_.emplace(key.segment_id, std::move(new_file))
```
buqian-zilliz: **[P2] Clean up the disk file if map insertion fails**

`DiskSlotFile` is constructed before `disk_files_.emplace()`, and its constructor has already opened and truncated `seg_<id>.cache`. If the map insertion throws, destroying `new_file` only closes the descriptor; the path remains. The best-effort boundary then swallows the exception, no map, Clock, or metric state records the file, and `EraseSegment()` also returns on a map miss without unlinking it. Allocation failures across multiple segments can therefore accumulate cache files outside `disk.maxBytes` accounting. Insert a placeholder before constructing the file, or catch insertion failure and unlink the newly created path.

internal/core/src/exec/expression/RawExprCacheAdapter.cpp
```diff
@@ -0,0 +1,511 @@
...
+            return;
+        }
+
+        auto capture_lease = manager.TryAcquireMaterialization(active_count);
+        if (!capture_lease.has_value()) {
+            AbortCapture();
+            return;
+        }
+        state_.capture_lease.emplace(std::move(*capture_lease));
+        state_.captured_result.reserve(active_count);
```
buqian-zilliz: **[P2] Avoid full-bitmap capture for latency-rejected keys**

`EnsureAdmission()` starts a full result/validity capture as soon as frequency admission succeeds, while `minEvalDurationUs` is checked only later in `PutAdmitted()`. A frequent predicate that remains below the latency threshold therefore never creates an entry, yet every later miss reserves and copies a full-segment bitmap pair before being rejected. For example, an all-skipped `SkipIndex` predicate over 100 million rows can repeatedly reserve and copy about 25 MB while denying that materialization budget to actual hits. Retain a per-key latency-rejection backoff, or perform a timing-only admitted pass before allocating capture buffers.

internal/core/src/exec/expression/ExprCache.h
```diff
@@ -80,7 +101,7 @@ class FrequencyTracker {
         }
     }
 
-    static constexpr size_t kNumSlots = 16384;  // 16K slots = 16KB memory
+    static constexpr size_t kNumSlots = 256 * 1024;  // 256K slots = 256 KiB
```
buqian-zilliz: **[P2] Do not synchronously decay the entire 256K admission table**

Increasing `kNumSlots` from 16K to 256K without changing the 10,000-record decay interval makes every 10,000th admission synchronously execute 262,144 atomic loads and 262,144 atomic stores on the query path. Besides averaging roughly 52 table operations per admission, this concentrates more than half a million atomic operations into one request and creates periodic tail-latency spikes under miss-heavy workloads. Decay a bounded rotating slice per record, use epoch-tagged counters, or scale the decay interval with the table size.

internal/core/src/exec/expression/RawExprCacheAdapter.cpp
```diff
@@ -0,0 +1,511 @@
...
+void
+RawExprCacheAdapter::Eval(EvalCtx& context, VectorPtr& result) {
+    const bool has_offset_input = context.get_offset_input() != nullptr;
+    SetHasOffsetInput(has_offset_input);
+    // A cache hit bypasses input_->Eval(), so mirror the offset mode that the
+    // wrapped evaluator would otherwise record for its MoveCursor() guard.
+    input_->SetHasOffsetInput(has_offset_input);
+
+    WaitPrefetch();
+    EnsureLookup();
```
buqian-zilliz: **[P3] Remove repeated lookup synchronization from every raw batch**

`Eval()` first calls `WaitPrefetch()`, whose no-future path invokes `EnsureLookup()`, and then immediately invokes `EnsureLookup()` again. On a miss, the adapter also calls `input_->WaitPrefetch()`, although each eligible wrapped evaluator begins its own `Eval()` with that same wait. After the first batch this leaves two completed `std::call_once` checks and duplicate wrapped-prefetch checks on every batch. Let the adapter's `WaitPrefetch()` establish lookup state once and let the wrapped evaluator own its prefetch wait.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -446,6 +644,100 @@ JsonKeyStats::CollectKeyInfo(const std::vector<FieldDataPtr>& field_datas,
...
+    };
+
+    while (next_submit < max_inflight && first_exception == nullptr) {
+        submit_next();
+    }
+
+    int64_t num_rows = 0;
+    while (next_consume < futures.size()) {
+        try {
+            auto result = std::move(futures[next_consume]).get();
```
buqian-zilliz, line 690: **[P2] Avoid ordered-future stalls in both parallel stages**

`CollectKeyInfo()` holds only one executor-sized window and consumes it strictly by range index. Because ranges are balanced by row count rather than JSON size or complexity, one slow first range can block this `get()` while the later futures finish; on an otherwise idle executor, those completed futures still occupy the window, so no more ranges are submitted and every other worker goes idle. `BuildKeyStatsParallel()` repeats the same ordered-window pattern. Refill from completion notifications under a bounded pending-result budget, while retaining results by range index for ordered merging and writing.

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -771,29 +1064,431 @@ JsonKeyStats::BuildKeyStatsForRow(std::string_view json_str, uint32_t row_id) {
...
+        }
+    };
+
+    auto submit_next = [&](bool block_for_budget) -> bool {
+        size_t reservation_bytes = 0;
+        JsonStatsRowRange range;
+        try {
+            range = row_ranges[next_submit];
+            reservation_bytes =
+                EstimateJsonStatsMaterializeReservationBytes(range, *schema);
```
buqian-zilliz, line 1344: **[P2] Reuse input sizes instead of rescanning every row serially**

Every materialization submission synchronously calls `EstimateJsonStatsMaterializeReservationBytes()` on the caller thread. That helper walks every row and calls `is_valid()` and `RawValue()`; the underlying `FieldData` accessors acquire multiple shared locks per row. A million-row build therefore adds millions of serialized lock acquisitions in a third full pass after key collection, which can become the throughput limit for short JSON values and delay task submission. Return each range's input-byte total from the key-collection worker and reuse it for admission instead of scanning the rows again.

internal/core/src/futures/Executor.h
```diff
@@ -33,4 +33,7 @@ getSearchCPUExecutor();
 folly::CPUThreadPoolExecutor*
 getLoadCPUExecutor();
 
+folly::CPUThreadPoolExecutor*
+getJsonStatsBuildExecutor();
```
zhengbuqian, line 37: seems we are adding another executor, but since this will be used only by the index node, I think it's fine. please confirm @sparknack

internal/core/src/index/json_stats/JsonKeyStats.cpp
```diff
@@ -84,6 +94,103 @@ namespace {
...
+    return lhs + rhs;
+}
+
+size_t
+SaturatingMultiply(size_t lhs, size_t rhs) {
+    if (lhs != 0 && rhs > std::numeric_limits<size_t>::max() / lhs) {
+        return std::numeric_limits<size_t>::max();
+    }
+    return lhs * rhs;
+}
```
zhengbuqian: we already have such helpers in common/Utils.h

zhengbuqian (approved):
good job! 

2 nit comments

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: > Thanks for pointing this out. The ordered window can indeed cause head-of-line blocking when JSON complexity is heavily skewed; splitting by row count does not eliminate that possibility. We don’t currently have evidence that this is a significant bottleneck for
  > this build path. Completion-driven refill would add out-of-order result buffering, memory accounting, and failure-handling complexity to both stages. I’d prefer to keep the bounded ordered implementation in this PR and revisit it if profiling shows meaningful worker
  > idle time.

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: > Agreed that this adds a serial pass with per-row accessor overhead. However, it only checks validity and reads the JSON value’s size—there is no parsing or key-map construction—so we expect it to be relatively cheap. Reusing collection totals also needs extra
  > bookkeeping when the writer batch size splits collection ranges into smaller materialization ranges. I’d prefer to keep the simpler approach for now and revisit it if profiling shows this pass is a bottleneck.

internal/core/src/index/json_stats/JsonKeyStats.cpp
zhagnlu: fixed

internal/core/src/futures/Executor.h
zhagnlu: yes, index node use it.

Issue #52029, opened 2026-07-30
[Enhancement]: Parallelize multi-segment insert processing in QueryNode delegator for partition-key collections


chyezh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Parallelize the independent per-segment work within QueryNode delegator's `ProcessInsert` when one insert batch contains records for multiple growing segments.

The current call path is:

1. `internal/querynodev2/pipeline/insert_node.go` groups insert records by segment ID.
2. `internal/querynodev2/delegator/delegator_data.go::ProcessInsert` iterates over the resulting `map[segmentID]*InsertData` serially.
3. For every segment, it creates or obtains the growing segment, calls `Segment.Insert`, updates PK candidates and BM25 IDF statistics, and registers newly created growing segments.

The per-segment operations should run concurrently with bounded parallelism because different segment IDs are independent. `ProcessInsert` must still wait for all segment tasks before returning so that insert ordering across pipeline batches remains unchanged.

The implementation should preserve the existing behavior and concurrency boundaries:

- Do not concurrently process two records for the same segment within one invocation.
- Keep new-growing-segment verification and registration safe with `growingSegmentLock` or an equivalent critical section.
- Preserve released/excluded segment skip behavior.
- Preserve PK candidate and BM25 IDF oracle updates.
- Notify the leader-view update callback once after the batch if one or more growing segments were added.
- Propagate unexpected segment creation or insert failures to the caller with behavior equivalent to the current panic path; a panic must not be lost inside a worker goroutine.
- Bound concurrency to avoid creating an unbounded number of goroutines or oversubscribing the online-write CGO workload.
- Do not change public APIs or insert visibility semantics.

### Why is this needed?

Partition-key collections distribute rows across multiple physical partitions and commonly produce insert batches that touch multiple growing segments. The default partition-key collection has 16 physical partitions, and a sufficiently distributed batch can therefore contain many segment entries.

`ProcessInsert` currently executes `Segment.Insert` serially for every segment. Its wall-clock latency becomes approximately the sum of all per-segment insert costs, even though the segment operations are independent. Since this work runs in the QueryNode insert pipeline, the serial loop limits ingestion throughput and can increase growing-data visibility and channel time-tick lag for partition-key workloads.

Parallelizing across segments should reduce batch processing latency while keeping each individual segment's mutation order unchanged.

Suggested acceptance criteria:

- A unit test demonstrates that inserts for different segments can overlap in time.
[...]

PR #52121, opened 2026-08-03
enhance: parallelize QueryNode segment inserts

chyezh:
issue: #52029

- QueryNode insert dispatch: add a dedicated bounded pool that follows the segcore mutate pool sizing factor
- delegator insert processing: process distinct growing segments concurrently while preserving registration, PK and BM25 updates, and deterministic wait-all failure semantics

### DIFF
--- a/internal/querynodev2/delegator/delegator_data.go
+++ b/internal/querynodev2/delegator/delegator_data.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"slices"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -150,6 +151,12 @@ type DeleteBatch struct {
 	Data []*DeleteData
 }
 
+type segmentInsertResult struct {
+	growingSegmentAdded bool
+	err                 error
+	panicValue          any
+}
+
 // Append appends another delete data into this one.
 func (d *DeleteData) Append(ad DeleteData) {
 	d.PrimaryKeys = append(d.PrimaryKeys, ad.PrimaryKeys...)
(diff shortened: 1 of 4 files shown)


tinswzy:

Reviewed at head `9cc72d9` (6 commits) against master. Overall: this is a well-constructed change and I'd consider it mergeable. The concurrency model is sound — each task operates on a distinct `segmentID` (map keys), `ProcessInsert` is never called concurrently for the same delegator (the insert pipeline node is single-goroutine per channel), and every shared structure the workers touch is internally synchronized: `idfOracle` (`o.Lock()` in both `RegisterGrowing`/`UpdateGrowing`), `distribution` (`d.mut`), and `segmentManager` (`typeutil.ConcurrentMap`). The `growingSegmentLock` still makes the `GrowingSegmentExists`→register check-then-act atomic per segment. `results[index]` has a clean happens-before via `future.Await()`. The pool separation is correct: `LocalSegment.Insert` (segment.go:926) submits to `mutatePool` and awaits, so fanning out on a separate `insertPool` avoids the nested-submit deadlock, and `conc.Pool` is blocking (`nonBlocking:false`), so a saturated pool queues rather than returning an overload error that would panic the node. The wait-all is also required for correctness because `ProcessInsert` runs under `WithInsertSchemaTransition`'s `schemaTransitionMu.RLock()` (insert_node.go:152) — draining all workers before returning keeps the schema pinned for the whole batch and avoids leaving goroutines touching C++ segments while the process unwinds on panic. Panic/error/nil-panic/`growingSegmentAdded`-across-panic semantics are all implemented carefully and backed by targeted tests. Findings below are non-blocking.

**Low — secondary worker panics/errors are silently dropped in the aggregation (`delegator_data.go:216-224`)**
```go
for _, result := range results {
    if result.panicked { panic(result.panicValue) }
    if result.err != nil { panic(result.err) }
}
```
The loop iterates in sorted `segmentID` order and propagates only the *first* failure. If two segments in one batch both panic (e.g. two malformed `InsertRecord`s, or two `growing.Insert` faults), only the lowest-ID panic value is re-raised; the other captured `panicValue` is neither logged nor surfaced anywhere — `runProcessInsertTask` (delegator_data.go:228) recovers it without logging, and `processInsert` only logs `err` paths, not panics. On a crash you lose the diagnostic for every non-propagated segment. Suggest logging each captured panic/err (segmentID + value) inside `runProcessInsertTask`, or in the aggregation loop before `panic(...)`, so all concurrent failures appear in the log even though only one is re-raised.

**Low — `insertPool` shares `mutatePool`'s size factor, so a large batch can monopolize the mutate pool (`pool.go:186-188`)**
```go
func insertPoolSize() int { return mutatePoolSize() }  // == CPU * mutatePoolSizeFactor (default 2)
```
`insertPool.Cap() == mutatePool.Cap()`, and each `insertPool` worker consumes one `mutatePool` worker for its nested `growing.Insert().Await()`. A single delegator whose batch has ≥ `CPU*factor` distinct growing segments can therefore occupy the entire mutate pool with its nested inserts, delaying concurrent `segment.Delete`/`Insert` cgo work from *other* delegators on the same QueryNode until the batch drains. This is head-of-line latency, not a deadlock (blocking pool resolves it), and the added parallelism is real (previously a batch serialized its cgo inserts one-at-a-time; now up to `mutatePool.Cap()` overlap). Still, since the fan-out cannot exceed mutate-pool throughput anyway, consider either a dedicated `insertPoolSizeFactor` or a comment documenting that the two pools are intentionally coupled and that insert fan-out can transiently saturate the mutate pool.

**Nit — `observeProcessCost()` is skipped on the panic path (`delegator_data.go:216-225`)**
The metric observation sits after the panic loop, so any batch that ends in panic never records `QueryNodeProcessCost`. This matches the pre-PR behavior (the old code also panicked mid-loop), so it's not a regression, but since the code now deliberately reaches a single tail, a `defer`-based observation would make the timing complete for the (rare) failure batches too.

**Nit — pre-existing: a freshly created growing segment is not released on `ErrSegmentNotLoaded`/`ErrSegmentNotFound` (`delegator_data.go:280-284`)**
```go
if errors.IsAny(err, merr.ErrSegmentNotLoaded, merr.ErrSegmentNotFound) {
    log.Warn(...); return nil   // new segment from NewSegment() above is not Release()d
}
```
When `newGrowingSegment` is true and `growing.Insert` returns one of these skip errors, the just-created segment is dropped without `Release()`. This is unchanged from the original loop (which `continue`d without releasing) and is an unusual state for a brand-new segment, so it's not introduced here — flagging only because this code was restructured and it's easy to fix now: release when `newGrowingSegment && err is skip-class`.

**Test coverage — strong.** `TestProcessInsertRunsDistinctSegmentsConcurrently`, `...MixedSkipAndSuccess`, `...WaitsForAllTasksBeforePanickingOnError`, `...PreservesWorkerPanicValueAfterAllTasks`, `...PreservesNilWorkerPanicAfterAllTasks`, `...NotifiesAfterPostRegistrationPanic`, and `...RegistersParallelBM25Stats` collectively exercise overlap, skip-vs-success mixing, wait-all-before-propagation, `panic(nil)` preservation, notify-before-panic ordering, and concurrent BM25 registration. Two small gaps worth a follow-up: (1) no assertion pinning `notifyCount==0` before all tasks complete for the plain error/worker-panic cases the way the nil-panic and post-registration tests do (they check `Zero(notifyCount)` pre-release; the error/worker-panic tests only check the final `==1`); (2) no test with more distinct segments than `insertPool.Cap()` to cover the blocking-submit/queueing path. The concurrency tests rely on `insertPool.Cap() >= 2`; with the default factor of 2 that holds even at 1 CPU (`1*2=2`), so this is not a flakiness risk, just an implicit assumption worth a comment.

Issue #52086, opened 2026-07-31
[Bug]: Streaming service cutover can deadlock while closing legacy timetick producers


chyezh:
## Environment

- **Milvus Version**: 2.6.18, Git commit `7ea6e3526`
- **Deployment Mode**: standalone
- **MQ**: Pulsar
- **Upgrade Path**: rolling upgrade from Milvus 2.5 to 2.6.18
- **pulsar-client-go**: v0.19.0
- **OS**: Kubernetes

### Option B: Steps

1. Deploy a Milvus 2.5 standalone instance backed by Pulsar.
2. Keep the legacy RootCoord DML/timetick msgstream active.
3. Roll the instance to Milvus 2.6.18 and let StreamingCoord complete the 2.6 version gate.
4. Trigger `ChannelManager.MarkStreamingHasEnabled` while `timetickSync.startWatch` is broadcasting timeticks.
5. The same streaming-ready notification concurrently:
   - asks `timetickSync.startWatch` to stop; and
   - closes the legacy DML msgstreams and their Pulsar producers.
6. Observe that StreamingCoord stops balancing and DDL/DCL operations no longer complete.

### Trigger Conditions

- **Frequency**: intermittent; requires a Send/Close timing race during the one-time 2.5 to 2.6 streaming-service cutover
- **First observed after**: approximately one minute after the new 2.6 standalone pod started
- **Does NOT happen when**: the streaming version was already persisted and the one-time cutover path is skipped

## Expected Behavior

The legacy RootCoord timetick senders should stop before their msgstreams are closed, or all in-flight sends should finish or return an error. `MarkStreamingHasEnabled` must not hold the ChannelManager lock while waiting for external listeners.

The streaming version transition should finish, WAL assignments should switch from read-only to read-write, and DDL/DCL requests should continue.

## Actual Behavior

`MarkStreamingHasEnabled` cancels four listeners and then calls `BlockUntilFinish` while holding `cm.cond.L`.

Three listeners exit normally, but `timetickSync.startWatch` never returns and therefore never executes its deferred notifier `Release`. The ChannelManager lock remains held indefinitely. This blocks the balancer and all callers of `WaitUntilStreamingEnabled`.

[...]

PR #52123, opened 2026-08-03
fix: prevent streaming cutover deadlock

chyezh:
issue: #52086

- legacy RootCoord DML streams: serialize admitted broadcasts with close and reject sends after shutdown
- streaming cutover: preserve the listener-completion barrier and wake parked waiters after cleanup

### DIFF
--- a/internal/rootcoord/dml_channels.go
+++ b/internal/rootcoord/dml_channels.go
@@ -36,8 +36,9 @@ import (
 )
 
 type dmlMsgStream struct {
-	ms    msgstream.MsgStream
-	mutex sync.RWMutex
+	ms     msgstream.MsgStream
+	mutex  sync.RWMutex
+	closed bool
 
 	refcnt int64 // current in use count
 	used   int64 // total used counter in current run, not stored in meta so meant to be inaccurate
@@ -82,6 +83,43 @@ func (dms *dmlMsgStream) DecRefCnt() {
 	} else {
 		mlog.Warn(context.TODO(), "Try to remove channel with no ref count", mlog.Int64("idx", dms.idx))
 	}
+}
+
+func (dms *dmlMsgStream) broadcast(ctx context.Context, pack *msgstream.MsgPack) (map[string][]msgstream.MessageID, error) {
+	dms.mutex.RLock()
+	defer dms.mutex.RUnlock()
+
+	if dms.closed {
+		return nil, merr.WrapErrServiceNotReadyMsg("legacy dml msgstream is closed")
+	}
+	if dms.refcnt <= 0 {
+		return nil, nil
+	}
+	return dms.ms.Broadcast(ctx, pack)
+}
+
+func (dms *dmlMsgStream) broadcastMark(ctx context.Context, pack *msgstream.MsgPack) (map[string][]msgstream.MessageID, error) {
+	dms.mutex.RLock()
+	defer dms.mutex.RUnlock()
+
+	if dms.closed {
+		return nil, merr.WrapErrServiceNotReadyMsg("legacy dml msgstream is closed")
+	}
+	if dms.refcnt <= 0 {
+		return nil, merr.WrapErrServiceInternalMsg("channel not in use")
+	}
+	return dms.ms.Broadcast(ctx, pack)
+}
+
+func (dms *dmlMsgStream) close() {
+	dms.mutex.Lock()
+	defer dms.mutex.Unlock()
+
+	if dms.closed {
+		return
+	}
+	dms.closed = true
+	dms.ms.Close()
 }
 
 // channelsHeap implements heap.Interface to performs like an priority queue.
@@ -180,6 +218,7 @@ func newDmlChannels(initCtx context.Context, factory msgstream.Factory, chanName
 
 	for i, name := range names {
 		var ms msgstream.MsgStream
+		var closeNotifier *snmanager.StreamingReadyNotifier
 		if !streamingutil.IsStreamingServiceEnabled() {
 			ms = d.newMsgstream(initCtx, factory, name)
 		} else {
@@ -191,14 +230,7 @@ func newDmlChannels(initCtx context.Context, factory msgstream.Factory, chanName
 			if !notifier.IsReady() {
 				logger.Info(initCtx, "streaming service is not enabled, create a msgstream to use")
 				ms = d.newMsgstream(initCtx, factory, name)
-				go func() {
-					defer notifier.Release()
-					<-notifier.Ready()
-					// release the msgstream.
-					logger.Info(initCtx, "streaming service is enabled, release the msgstream...")
-					ms.Close()
-					logger.Info(initCtx, "streaming service is enabled, release the msgstream done")
-				}()
+				closeNotifier = notifier
 			} else {
 				logger.Info(initCtx, "streaming service has been enabled, msgstream should not be created")
 				notifier.Release()
@@ -213,6 +245,18 @@ func newDmlChannels(initCtx context.Context, factory msgstream.Factory, chanName
 		}
 		d.pool.Insert(name, dms)
 		d.channelsHeap = append(d.channelsHeap, dms)
+
+		if closeNotifier != nil {
+			go func(notifier *snmanager.StreamingReadyNotifier, stream *dmlMsgStream, pchannel string) {
+				defer notifier.Release()
+				<-notifier.Ready()
+				// release the msgstream.
+				logger := mlog.With(mlog.String("pchannel", pchannel))
+				logger.Info(initCtx, "streaming service is enabled, release the msgstream...")
+				stream.close()
+				logger.Info(initCtx, "streaming service is enabled, release the msgstream done")
+			}(closeNotifier, dms, name)
+		}
 	}
 
 	heap.Init(&d.channelsHeap)
@@ -322,15 +366,10 @@ func (d *dmlChannels) broadcast(chanNames []string, pack *msgstream.MsgPack) err
 			return err
 		}
 
-		dms.mutex.RLock()
-		if dms.refcnt > 0 {
-			if _, err := dms.ms.Broadcast(d.ctx, pack); err != nil {
-				mlog.Error(d.ctx, "Broadcast failed", mlog.Err(err), mlog.String("chanName", chanName))
-				dms.mutex.RUnlock()
-				return err
-			}
+		if _, err := dms.broadcast(d.ctx, pack); err != nil {
+			mlog.Error(d.ctx, "Broadcast failed", mlog.Err(err), mlog.String("chanName", chanName))
+			return err
 		}
-		dms.mutex.RUnlock()
 	}
 	return nil
 }
@@ -343,25 +382,17 @@ func (d *dmlChannels) broadcastMark(chanNames []string, pack *msgstream.MsgPack)
 			return result, err
 		}
 
-		dms.mutex.RLock()
-		if dms.refcnt > 0 {
-			ids, err := dms.ms.Broadcast(d.ctx, pack)
-			if err != nil {
-				mlog.Error(d.ctx, "BroadcastMark failed", mlog.Err(err), mlog.String("chanName", chanName))
-				dms.mutex.RUnlock()
-				return result, err
-			}
-			for cn, idList := range ids {
-				// idList should have length 1, just flat by iteration
-				for _, id := range idList {
-					result[cn] = id.Serialize()
-				}
-			}
-		} else {
-			dms.mutex.RUnlock()
-			return nil, merr.WrapErrServiceInternalMsg("channel not in use: %s", chanName)
+		ids, err := dms.broadcastMark(d.ctx, pack)
+		if err != nil {
+			mlog.Error(d.ctx, "BroadcastMark failed", mlog.Err(err), mlog.String("chanName", chanName))
+			return result, merr.Wrapf(err, "channel %s", chanName)
 		}
-		dms.mutex.RUnlock()
+		for cn, idList := range ids {
+			// idList should have length 1, just flat by iteration
+			for _, id := range idList {
+				result[cn] = id.Serialize()
+			}
+		}
 	}
 	return result, nil
 }
--- a/internal/rootcoord/dml_channels.go
+++ b/internal/rootcoord/dml_channels.go
@@ -113,12 +113,13 @@ func (dms *dmlMsgStream) broadcastMark(ctx context.Context, pack *msgstream.MsgP
 
 func (dms *dmlMsgStream) close() {
 	dms.mutex.Lock()
-	defer dms.mutex.Unlock()
-
 	if dms.closed {
+		dms.mutex.Unlock()
 		return
 	}
 	dms.closed = true
+	dms.mutex.Unlock()
+
 	dms.ms.Close()
 }
 
--- a/internal/rootcoord/dml_channels_test.go
+++ b/internal/rootcoord/dml_channels_test.go
@@ -21,8 +21,11 @@ import (
 	"context"
 	"math/rand"
 	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
+	"github.com/bytedance/mockey"
 	"github.com/cockroachdb/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -31,6 +34,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/mq/common"
 	"github.com/milvus-io/milvus/pkg/v3/mq/msgstream"
 	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 )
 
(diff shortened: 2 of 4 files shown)


tinswzy:

Reviewed at head `df21829` against `master` (merge base `bcbc1c6`). Read-only review of both commits individually and the aggregate diff.

**Overall take:** Mergeable. The production change is a genuine fix and is well-encapsulated; findings below are non-blocking (low/nit + a couple of observations that help future readers). No critical/high issues found.

**Scope note (informational, not a defect).** The aggregate PR diff touches only 3 files: `internal/rootcoord/dml_channels.go` and two test files. `internal/streamingcoord/server/balancer/channel/manager.go` does **not** appear in the net diff — commit 2 (`df21829`, "preserve streaming cutover barrier") reverts commit 1's (`2a53233`, "prevent streaming cutover deadlock") `manager.go` hunk exactly (blob returns to `67df9fafe8`). So `MarkStreamingHasEnabled`/`WaitUntilStreamingEnabled` are byte-identical to master; the "barrier" half of this PR is effectively a no-op revert plus a new regression test. That is a fine outcome, but reviewers should understand the only behavioral change lands in `dml_channels.go`. Worth confirming the two commits were intended to net-cancel in `manager.go` rather than land a change there.

**Findings**

low — `dml_channels.go:114-123` (`close`) now serializes behind in-flight `Broadcast`, but `Broadcast` runs with `d.ctx == context.TODO()` (`dml_channels.go:211`, passed at `:369`/`:385`).
```go
func (dms *dmlMsgStream) close() {
	dms.mutex.Lock()          // waits for any in-flight broadcast to release RLock
	defer dms.mutex.Unlock()
	...
	dms.ms.Close()
}
```
This correctly removes the master data race (old code called `ms.Close()` off-lock concurrently with `ms.Broadcast()` → `producer.Send()`, which is the actual bug being fixed — good). The residual risk: `ms.Broadcast()` → `producer.Send()` has no deadline (context.TODO), so if the legacy broker stalls during cutover, `close()` blocks holding the write lock. Because the closed `dmlMsgStream` stays in `channelsHeap`, `channelsHeap.Less` → `RefCnt()`/`Used()` take that stream's `RLock`, so a stuck `close()` also stalls `addChannels`/`removeChannels`/`getChannelNames` (all under `d.mut`), and stalls `MarkStreamingHasEnabled` (blocked in `BlockUntilFinish`) which holds `cm.cond.L`. This is strictly better than the racy master, but the "prevent deadlock" goal isn't fully achieved as long as the legacy `Send` is unbounded. Suggestion: pass a bounded context to `dms.ms.Broadcast` (or document that legacy `Send` is guaranteed to return) so cutover cannot wedge on an unreachable broker.

low — `manager.go:330-358` (`MarkStreamingHasEnabled`) does not broadcast the cond on enable; parked `WaitUntilStreamingEnabled` waiters are not woken by the enable event. In the final (reverted) code the enable path sets `streamingVersion` and never calls `cond.Broadcast()`/`UnsafeBroadcast()`. A goroutine already parked inside `cond.Wait(ctx)` (`context_condition_variable.go:59`, already unlocked, selecting on `ch`) before enable will only wake on an incidental later `LockAndBroadcast` (e.g. `TriggerWatchUpdate`), not on enable itself. This is pre-existing master behavior (the PR reverts to it), and the new test does not exercise it — its waiter arrives after `Cancel()` and parks on `Lock()`, not on `Wait()`, so it returns as soon as the version is already set. Given the PR's stated goal of preserving the barrier, consider adding `cm.cond.UnsafeBroadcast()` *after* `BlockUntilFinish()` (still under the lock, before returning). That wakes parked waiters deterministically without reintroducing commit 1's ordering bug (which broadcast *before* `BlockUntilFinish`, publishing enablement before listeners completed cutover).

nit — `dml_channels.go:369` vs `:388`: inconsistent error context. `d.broadcast` returns the inner error unwrapped, while `d.broadcastMark` wraps with `merr.Wrapf(err, "channel %s", chanName)`. The non-mark path loses the channel name on failure. Minor; align the two if the channel name is useful for triage. (Use of `merr.Wrapf` rather than `WrapErrXxxErr` here is correct — it preserves the inner code.)

nit — `dml_channels.go:249-259`: the goroutine rebuilds `logger := mlog.With(mlog.String("pchannel", pchannel))`, duplicating the `logger` already constructed at `:229` for the same pchannel. Harmless; could pass it in.

**Test coverage — good.** `TestDmlMsgStreamCloseWaitsForBroadcast` directly and correctly validates the core invariant (close blocks until in-flight broadcast completes; post-close `broadcast`/`broadcastMark` return `ErrServiceNotReady`; the underlying stream is not touched after close — `broadcastCalls == 1`). `TestMarkStreamingHasEnabledWaitsForListenersBeforePublishing` is a valid regression guard: under commit 1's pattern `WaitUntilStreamingEnabled` would observe the version and return early, tripping the `t.Fatalf`; under the reverted code it blocks on `Lock()` until listeners finish. Both require `-gcflags="all=-N -l"` for the mockey patch (already the repo default) — fine.

**Classification check — reasonable.** `closed → merr.WrapErrServiceNotReadyMsg` (retriable/System) is the right choice for the transient cutover window: a caller retrying re-resolves routing to the streaming WAL rather than the permanently-closed legacy stream. `refcnt<=0 → merr.WrapErrServiceInternalMsg("channel not in use")` matches master. One thing to confirm out-of-band: that every caller of `dmlChannels.broadcast`/`broadcastMark` treats `ErrServiceNotReady` as retry-and-re-route (not retry-same-stream), otherwise a caller pinned to the closed legacy stream would retry `ServiceNotReady` indefinitely post-cutover.

Nice cleanup overall: all `dms.ms.*` access is now funneled through the locked `broadcast`/`broadcastMark`/`close` helpers (verified no remaining direct `ms.Broadcast`/`ms.Close` call sites in `internal/rootcoord`), and the goroutine passes `notifier`/`stream`/`pchannel` explicitly to avoid loop-capture.

czs007:

Adversarial review surfaced one medium-severity concurrency issue in the legacy DML close path, where the new close-under-lock coupling can stall the entire `ChannelManager` during streaming cutover.

## Medium

- **`internal/rootcoord/dml_channels.go:114`** — `dmlMsgStream.close()` calls `ms.Close()` while holding `dms.mutex.Lock()`. That write lock must wait for every in-flight `broadcast`/`broadcastMark` (`dml_channels.go:88-112`) to release its `RLock`, and those hold the read lock across `dms.ms.Broadcast()` → `producer.Send()` (`pkg/mq/msgstream/mq_msgstream.go:346-352`), which can block for an unbounded time under MQ backpressure or unreachability. The release goroutine's `defer notifier.Release()` (`dml_channels.go:251`) therefore cannot run until `close()` returns, and `MarkStreamingHasEnabled` waits on that `BlockUntilFinish` *while holding* `cm.cond.L` (`internal/streamingcoord/server/balancer/channel/manager.go:332, 353-355`). Failure scenario: streaming cutover fires `MarkStreamingHasEnabled` while RootCoord still has one in-flight legacy DDL/timetick `dms.broadcast()` whose Pulsar/Kafka `producer.Send()` is blocked on an unreachable broker or a full pending queue → `close()` parks on `dms.mutex.Lock()`, `Release()` is never reached, and `cm.cond.L` stays held for the full duration of the MQ stall, freezing `AddPChannels` (`manager.go:278`), `AllocVirtualChannels` (`:394`), `TriggerWatchUpdate` (`:323`), `waitChanges`/watch RPCs (`:748`) and `WaitUntilStreamingEnabled` (`:245`), so collection creation and channel assignment stop and the cutover cannot complete. The lock graph is acyclic (no `dms.mutex → cm.cond.L` edge), so this is a stall rather than a deadlock — but the stall is as long as the MQ produce hangs, and Go's writer-preferring `RWMutex` means the pending `Lock()` also blocks all subsequent `RLock()` users of that stream (`RefCnt`/`IncRefcnt`/`broadcast`), widening the blast radius. This coupling is introduced here: on base, the release goroutine called `ms.Close()` directly without touching `dms.mutex`, and `mqMsgStream.Broadcast` does not take `closeRWMutex`, so `Release()` never waited on an in-flight broadcast — the PR's own `TestDmlMsgStreamCloseWaitsForBroadcast` asserts the new behaviour. **Suggestion:** flip only the `closed` flag under `dms.mutex` and move `ms.Close()` outside the lock; alternatively bound `close()` with a timeout, or decouple `notifier.Release()` from `Close()` completion so cutover no longer waits on a legacy MQ produce. (surfaced during verification)

chyezh:
Followed up on the medium-severity finding.

The residual liveness risk is valid if an MQ `Producer.Send()` itself never returns. However, I do not think this requires another change in this PR:

- The concrete failure in #52086 is the Pulsar v0.19 concurrent Send/Close race: Close can terminate the producer event loop while an admitted request is still buffered, leaving synchronous Send without a callback. Holding the read lock through Broadcast and acquiring the write lock before Close guarantees that all admitted sends finish before the producer is closed, and that no new send can start afterward.
- A send that hangs independently of Close already prevents `timetickSync.startWatch` from passing its `wg.Wait()`, so the streaming-enable listener cannot finish. The PR removes the Send/Close-induced hang; it does not attempt to provide bounded progress for every MQ outage.
- Setting `closed` under the write lock and invoking `ms.Close()` after unlocking would preserve Send/Close serialization, but it would still wait to acquire the write lock behind the stuck Send. It therefore does not address the failure scenario described in the finding.
- Adding a reliable timeout/cancellation path is broader than this change: `mqMsgStream.Broadcast(ctx, ...)` currently sends with the message `TraceCtx`, rather than the supplied `ctx`. Decoupling notifier completion from Close would also change the streaming cutover barrier/lifecycle semantics.

Given the adversarial review result of zero must-fix issues, I am keeping this PR focused on the reported Send/Close race. Bounded MQ-send cancellation and cutover lifecycle handling can be addressed separately if needed.

czs007:

## Re-review of `1077eccd5ede`

- **[medium] `internal/rootcoord/dml_channels.go:114` — `close()` calls `ms.Close()` under the `dms.mutex` write lock, propagating legacy-MQ blocking to `cm.cond.L`** — **Confirmed**: both reviewers re-walked the full chain this round and every link is still present (`close()` holds the write lock across `ms.Close()`; `broadcast`/`broadcastMark` hold `RLock` across the whole `Broadcast()`; `MarkStreamingHasEnabled` calls `BlockUntilFinish()` with `cm.cond.L` held).

## Findings — this round

Re-verification of commit `1077eccd5ede` confirms the previously reported liveness risk and surfaces no new critical/high issues.

### Medium

- **`internal/rootcoord/dml_channels.go:114` — `close()` holds the `dms.mutex` write lock while calling `ms.Close()`, adding a path that propagates legacy-MQ blocking into `cm.cond.L` and the rootcoord DDL path.** `close()` (`dml_channels.go:114-123`) takes `dms.mutex.Lock()` and calls `ms.Close()` under it, while `broadcast`/`broadcastMark` (`:88-112`) hold `RLock` for the entire duration of `dms.ms.Broadcast()`. In `pkg/mq/msgstream/mq_msgstream.go:319-357`, `Broadcast` calls `producer.Send()` synchronously on a `spanCtx` derived from `MsgSpanFromCtx(v.TraceCtx(), v)`, so the caller-supplied `ctx` does not bound it; Go's `RWMutex` also blocks all subsequent readers once a writer is queued. Failure: during a 2.5→2.6 streaming cutover with Pulsar degraded or under heavy backpressure, the rootcoord timetick path blocks in `producer.Send()` while holding `RLock`, so each pchannel's cutover goroutine (`:250-259`) stalls on `dms.mutex.Lock()` and never reaches `defer notifier.Release()` → `MarkStreamingHasEnabled` (`streamingcoord/server/balancer/channel/manager.go:331-360`) blocks in `BlockUntilFinish()` while holding `cm.cond.L`, indefinitely stalling every `cm.cond.L` user in the same (mixcoord) process — `WaitUntilStreamingEnabled:244`, `IsStreamingEnabledOnce:236`, `CurrentPChannelsView:381`, `AllocVirtualChannels:394`, `AddPChannels:277` — halting DDL such as collection creation as well as channel assignment; concurrently `addChannels` (`:369-381`) holds `d.mut` while waiting on the `dms.mutex` write lock held by `ms.Close()`, which also stalls `getChannelNames` (`:269-271`). Nothing in that chain has a timeout. This is not a lock cycle — no `dms.mutex` holder acquires `cm.cond.L` in reverse — and it requires MQ `Send`/`Close` to hang, which is why it is rated medium rather than blocking. Before this PR, `ms.Close()` was called entirely without holding `dms.mutex`, so this edge is new. Suggestion: keep `SaveVersion` under the lock but move the point at which the enabled state becomes externally visible to after the listeners finish — run `Cancel`/`BlockUntilFinish` outside `cm.cond.L`, then re-acquire the lock to set the reader-visible fields and `UnsafeBroadcast()`. The barrier is then guaranteed by publication order rather than by lock hold time. Note that this PR's first commit `2a53233e70` already had `Cancel`/`BlockUntilFinish` outside `cm.cond.L`, and the second commit `df21829013` moved them back inside to restore the barrier — so a lock that does not span `BlockUntilFinish` is demonstrably workable. An alternative is to track and cancel/bound in-flight sends before draining. (raised by tinswzy, sijie-ni-0214)

internal/rootcoord/dml_channels.go
```diff
@@ -84,6 +85,43 @@ func (dms *dmlMsgStream) DecRefCnt() {
...
+
+func (dms *dmlMsgStream) close() {
+	dms.mutex.Lock()
+	defer dms.mutex.Unlock()
+
+	if dms.closed {
+		return
+	}
+	dms.closed = true
+	dms.ms.Close()
```
czs007, line 123: (non-blocking) `ms.Close()` runs while the write lock is held, and `RefCnt()` / `Used()` (`dml_channels.go:50-61`) both take the read lock. Go's `sync.RWMutex` blocks new readers as soon as a writer is queued, so a slow `ms.Close()` (a Pulsar producer `Close()` flushes pending sends) blocks `channelsHeap.Less` (`:135-151`), and therefore `getChannelNames` / `addChannels` / `removeChannels`, which call it while holding `d.mut`. That couples legacy-stream teardown to rootcoord's channel-allocation path — a coupling master did not have, since `Close()` used to run off-lock.

The write lock is only needed to flip the flag and drain in-flight readers; once `closed` is true nobody can reach `ms` again, so the close itself is safe outside the lock:

```go
func (dms *dmlMsgStream) close() {
	dms.mutex.Lock()
	if dms.closed {
		dms.mutex.Unlock()
		return
	}
	dms.closed = true
	dms.mutex.Unlock() // no one can enter ms from here on
	dms.ms.Close()
}
```

internal/rootcoord/dml_channels.go
```diff
@@ -213,6 +245,18 @@ func newDmlChannels(initCtx context.Context, factory msgstream.Factory, chanName
...
+		if closeNotifier != nil {
+			go func(notifier *snmanager.StreamingReadyNotifier, stream *dmlMsgStream, pchannel string) {
+				defer notifier.Release()
+				<-notifier.Ready()
+				// release the msgstream.
+				logger := mlog.With(mlog.String("pchannel", pchannel))
+				logger.Info(initCtx, "streaming service is enabled, release the msgstream...")
+				stream.close()
+				logger.Info(initCtx, "streaming service is enabled, release the msgstream done")
+			}(closeNotifier, dms, name)
```
czs007, line 259: (non-blocking, but worth recording on #52086) The cutover path now contains an unbounded wait: `MarkStreamingHasEnabled` holds `cm.cond.L` for its whole body and blocks in `BlockUntilFinish()` (`manager.go:352-354`) → this listener goroutine → `stream.close()` → write lock → in-flight `ms.Broadcast()`. `d.ctx` is `context.TODO()` (`:211`), so if the legacy broker stalls, `cm.cond.L` stays held indefinitely and the whole `ChannelManager` wedges (assignment watch, `WaitUntilStreamingEnabled`, `RegisterStreamingEnabledNotifier`). On master `close()` did not wait for the in-flight broadcast, so this specific wedge is new here. The trade is still net positive — it removes a real use-after-close race — but the "prevent deadlock" goal isn't fully reached, and it would be easy for a future reader to assume it is.

One thing to note before someone reaches for the obvious fix: passing a deadline-bearing context to `dms.ms.Broadcast` would **not** bound it. `mqMsgStream.Broadcast` ignores its `ctx` argument — it derives `spanCtx` from `MsgSpanFromCtx(v.TraceCtx(), v)` and sends with `producer.Send(spanCtx, msg)` (`pkg/mq/msgstream/mq_msgstream.go:326`, `:347`), and the `TimeTickMsg` built in `sendTimeTickToChannel` never sets `BaseMsg.ctx`, so `TraceCtx()` returns nil. Boundedness has to come from a producer-level send timeout, or from moving the drain off the cutover path (flip `closed`, `notifier.Release()`, then drain and close asynchronously) — the latter only if it's acceptable for an in-flight legacy timetick to land after enablement.

internal/rootcoord/dml_channels.go
chyezh: Fixed in 43f7c30963. `close()` now acquires the write lock to drain existing broadcasts and set `closed`, then releases the lock before calling `ms.Close()`. New broadcasts still observe `closed` and cannot overlap the underlying close, while `RefCnt()` / `Used()` are no longer blocked by a slow `ms.Close()`. I also extended the regression test to hold the underlying Close blocked and verify state reads plus post-close broadcast rejection.

internal/rootcoord/dml_channels.go
chyezh: Agreed that this residual unbounded-wait risk remains. I am keeping the current synchronous drain for this fix because it preserves the cutover barrier: releasing the notifier or draining asynchronously could allow an in-flight legacy timetick to land after streaming enablement. Bounding it safely needs producer-level send timeout/cancellation or a broader lifecycle design, so I am treating this as a non-blocking follow-up rather than adding that complexity to #52086.

czs007:

Verified:
- `internal/rootcoord/dml_channels.go:114-124` — `close()` takes the write lock to set `closed` before releasing it, so the lock acquisition drains all in-flight `broadcast` readers and `ms.Close()` can never run concurrently with `ms.Broadcast()`.
- `internal/rootcoord/dml_channels.go:88-111` — `broadcast`/`broadcastMark` check `closed` before `refcnt` and call `ms.Broadcast` under the read lock, so once the close flag is set no new send can reach a closed producer.
- `internal/rootcoord/dml_channels.go:49-61` — `ms.Close()` is invoked outside the lock, so a slow underlying close no longer blocks `RefCnt()`/`Used()` state reads; `TestDmlMsgStreamCloseWaitsForBroadcast` exercises this with a blocking close hook.
- `internal/rootcoord/dml_channels.go:222-259` — the listener goroutine explicitly captures `dms` and the channel name (no loop-variable capture), and `notifier.Release()` still runs after `stream.close()`, preserving the cutover completion barrier.
- `internal/streamingcoord/server/balancer/channel/manager.go:349-357` — notifiers are cancelled and awaited via `BlockUntilFinish()` before `UnsafeBroadcast()` is called while holding `cm.cond.L`, satisfying the condition-variable contract and closing the liveness gap where parked `WaitUntilStreamingEnabled` waiters were never woken after `streamingVersion` was set.
- `internal/streamingcoord/server/balancer/channel/manager_test.go:510-630` — tests cover both sides of that change: streaming-enabled state is not published before listeners finish, and parked condition waiters are actually woken afterwards.

czs007 (approved):
/lgtm

PR #52124, opened 2026-08-03
enhance: optimize bulk import metadata and batching

### LINKED ISSUE #52122: Optimize bulk import preimport and batching performance
## Background

Bulk import currently spends avoidable time in two places for large vector datasets:

1. PreImport scans the input files to derive row count and memory estimates even when the reader can get row count from file metadata.
2. Import reads small batches because `EstimateReadCountPerBatch` is capped at 1000 rows when the buffer can hold more rows. This increases the number of batches, InsertData assembly calls, and sync tasks.
3. The import storage config does not pass through the configured object storage `MaxConnections`, so import-side remote IO cannot use the configured connection pool consistently.

## Proposed optimization

- Add a metadata-based PreImport fast path for readers that can expose row count directly, with conservative fallback conditions for cases that still require row-level hashing or validation.
- Increase import reader batch size based on the task buffer and estimated record size, while reserving half of the buffer as working-set headroom for system fields, hashing, and InsertData copies.
- Pass `MinioCfg.MaxConnections` into import `StorageConfig` so import readers use the configured object storage connection limit.

## Measurement

Dataset: `cohere_1M`  
Metric: import task completion time from REST progress `details[*].state=Completed`; post-import Sorting/global job completion is excluded.

| Variant | Import task completion | Relative to baseline | PreImport read stat | Sync task count | Sync task sum |
| --- | ---: | ---: | ---: | ---: | ---: |
| baseline | 339.370s | 1.00x | 35.147s | 1000 | 294.575s |
| PreImport metadata fast path only | 291.181s | 1.17x | 0.051s | 1000 | 282.339s |
| Batch half-buffer only | 150.622s | 2.25x | 30.649s | 368 | 112.205s |
| Storage MaxConnections only | 330.334s | 1.03x | 30.614s | 1000 | 291.669s |
| All optimizations | 117.489s | 2.89x | 0.041s | 368 | 111.563s |

marcelo-cjl:
Fixes #52122

## What this PR does

This PR optimizes bulk import performance in three places:

- **PreImport metadata fast path**
  - Adds a `NumRows()` capability for NumPy and Parquet import readers.
  - Lets PreImport derive row count and memory stats from file metadata instead of scanning batches when the schema is safe for this shortcut.
  - Keeps the original scan path for partition-key collections, variable-length or nullable fields, and multi-vchannel non-autoID cases where row-level hashing or exact data sizing is required.

- **Larger import reader batches**
  - Removes the fixed 1000 rows-per-batch cap from `EstimateReadCountPerBatch`.
  - Computes batch rows from `bufferSize / estimatedRecordSize / 2`.
  - Reserves half of the task buffer as headroom for system fields, row hashing, and InsertData copies created after reader output.

- **Import storage connection config**
  - Passes `MinioCfg.MaxConnections` into import `StorageConfig` so import readers use the configured object storage connection limit consistently.

## Measurement

Dataset: `cohere_1M`  
Metric: import task completion time from REST progress `details[*].state=Completed`; post-import Sorting/global job completion is excluded.

| Variant | Import task completion | Relative to baseline | PreImport read stat | Sync task count | Sync task sum |
| --- | ---: | ---: | ---: | ---: | ---: |
| baseline | 339.370s | 1.00x | 35.147s | 1000 | 294.575s |
| PreImport metadata fast path only | 291.181s | 1.17x | 0.051s | 1000 | 282.339s |
| Batch half-buffer only | 150.622s | 2.25x | 30.649s | 368 | 112.205s |
| Storage MaxConnections only | 330.334s | 1.03x | 30.614s | 1000 | 291.669s |
| All optimizations | 117.489s | 2.89x | 0.041s | 368 | 111.563s |

## Tests

- `go test ./internal/util/importutilv2/common ./internal/util/importutilv2/numpy ./internal/util/importutilv2/parquet`
- `go test -tags dynamic,test ./internal/datanode/importv2 ./internal/datacoord -run '^$'`

### DIFF
--- a/internal/datacoord/util.go
+++ b/internal/datacoord/util.go
@@ -400,6 +400,7 @@ func createStorageConfig() *indexpb.StorageConfig {
 			GcpCredentialJSON: Params.MinioCfg.GcpCredentialJSON.GetValue(),
 			SslTlsMinVersion:  Params.MinioCfg.SslTLSMinVersion.GetValue(),
 			UseCrc32CChecksum: Params.MinioCfg.UseCRC32C.GetAsBool(),
+			MaxConnections:    uint32(Params.MinioCfg.MaxConnections.GetAsInt()),
 		}
 	}
 
--- a/internal/datanode/importv2/task_preimport.go
+++ b/internal/datanode/importv2/task_preimport.go
@@ -185,6 +185,54 @@ func (t *PreImportTask) Execute() []*conc.Future[any] {
 	return futures
 }
 
+type numRowsReader interface {
+	NumRows() (int64, bool)
+}
+
+func (t *PreImportTask) canUseMetadataFastPath() bool {
+	if len(t.partitionIDs) != 1 || len(t.vchannels) == 0 {
+		return false
+	}
+	if !isFixedSizeImportSchema(t.GetSchema()) {
+		return false
+	}
+	if _, err := typeutil.GetPartitionKeyFieldSchema(t.GetSchema()); err == nil {
+		return false
+	}
+	pkField, err := typeutil.GetPrimaryFieldSchema(t.GetSchema())
+	if err != nil {
+		return false
+	}
+	return pkField.GetAutoID() || len(t.vchannels) == 1
+}
+
+func isFixedSizeImportSchema(schema *schemapb.CollectionSchema) bool {
+	for _, field := range typeutil.GetAllFieldSchemas(schema) {
+		if field.GetNullable() {
+			return false
+		}
+		switch field.GetDataType() {
+		case schemapb.DataType_Bool,
+			schemapb.DataType_Int8,
+			schemapb.DataType_Int16,
+			schemapb.DataType_Int32,
+			schemapb.DataType_Int64,
+			schemapb.DataType_Float,
+			schemapb.DataType_Double,
+			schemapb.DataType_Timestamptz,
+			schemapb.DataType_BinaryVector,
+			schemapb.DataType_FloatVector,
+			schemapb.DataType_Float16Vector,
+			schemapb.DataType_BFloat16Vector,
+			schemapb.DataType_Int8Vector:
+			continue
+		default:
+			return false
+		}
+	}
+	return true
+}
+
 func (t *PreImportTask) readFileStat(reader importutilv2.Reader, fileIdx int) error {
 	fileSize, err := reader.Size()
 	if err != nil {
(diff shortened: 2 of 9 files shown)


internal/datanode/importv2/task_preimport.go
```diff
@@ -197,6 +245,56 @@ func (t *PreImportTask) readFileStat(reader importutilv2.Reader, fileIdx int) er
 				"fileSize=%d, maxSize=%d", fileSize, int64(maxSize))
 	}
 
+	if nrReader, ok := reader.(numRowsReader); ok && t.canUseMetadataFastPath() {
+		numRows, ok := nrReader.NumRows()
```
liliu-z, line 249: The metadata fast path takes the NumPy row count from the header shape (FieldReader.NumRows reads Header.Descr.Shape) without reading the payload; validateHeader checks only the shape rank and the vector dim (shape[1]), never shape[0] against the actual byte length, whereas the scan path it replaces read the bytes and failed on a short/corrupt file. A truncated or crafted .npy whose shape[0] overstates the real rows now produces an inflated TotalRows/TotalMemorySize that feeds AssignSegments (`for size > 0 { AllocImportSegment; size -= segmentMaxSize }`), ID pre-allocation, and disk-quota checks, so datacoord can over-allocate import segments before the mismatch finally surfaces when the import task reads the short payload. Validate shape[0]*per-row-bytes against the payload/file size before trusting the count (Parquet is unaffected — its NumRows comes from the validated file footer).

internal/datacoord/util.go
```diff
@@ -400,6 +400,7 @@ func createStorageConfig() *indexpb.StorageConfig {
 			GcpCredentialJSON: Params.MinioCfg.GcpCredentialJSON.GetValue(),
 			SslTlsMinVersion:  Params.MinioCfg.SslTLSMinVersion.GetValue(),
 			UseCrc32CChecksum: Params.MinioCfg.UseCRC32C.GetAsBool(),
+			MaxConnections:    uint32(Params.MinioCfg.MaxConnections.GetAsInt()),
```
liliu-z, line 403: This populates MaxConnections in the shared createStorageConfig(), but three consumers of that proto drop it, so the setting silently never reaches them: the index worker rebuilds indexcgopb.StorageConfig field-by-field and stops at UseCrc32CChecksum (internal/datanode/index/task_index.go:275-293), the analyze worker does the same for clusteringpb.StorageConfig (internal/datanode/index/task_analyze.go:100-118), and StorageV3's MakePropertiesFromStorageConfig never emits the max-connections property (internal/storagev2/packed/ffi_common.go, even though PropertyFSMaxConnections exists at :58). On those paths C++ gets max_connections=0 and the `if (storage_config.max_connections > 0)` guard in ChunkManager.cpp leaves the client at its default (25 for the AWS/MinIO ChunkManager), so index build and analyze keep using the default regardless of config while the packed import read/write and stats paths honor it. Copy MaxConnections in those field-by-field builds and add the property in MakePropertiesFromStorageConfig for consistent behavior.

Issue #52033, opened 2026-07-30
[Bug]: Azure Blob Storage ManagedIdentity authentication fails when checking Hierarchical Namespace support


ducminhle:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 3.0.0
- Deployment mode(standalone or cluster): standalone
- MQ type(rocksmq, pulsar or kafka): woodpecker   
- SDK version(e.g. pymilvus v2.0.0rc2): 
- OS(Ubuntu or CentOS): Ubuntu (Docker 29.4.3)
- CPU/Memory: 8 core, 32GB
- GPU: None
- Others:
```

### Current Behavior
I have deployed Milvus [v3.0.0](https://github.com/milvus-io/milvus/tree/v3.0.0), with Azure managed identity (set 2 env: AZURE_CLIENT_ID and AZURE_TENANT_ID)
When using Azure Blob Storage as the object storage backend, inserting data from Attu appears to succeed. The inserted entities are visible in the Attu UI, but Milvus logs the following error during the insert operation:

```
Assert "result.ok()"  => [StorageV2] Failed to create packed writer: IOError: Failed to create packed column group writer. [path=files/insert_log/468031747677751091/468031747677751092/468031747677951240/0/468031747677951418]: IOError: Failed to open output stream: Unknown error: Check for Hierarchical Namespace support on 'https://milvusdemosa.blob.core.windows.net:443/milvus' failed: N5Azure4Core11Credentials23AuthenticationExceptionE: GetToken(): error response: 404 Not Found

<?xml version="1.0" encoding="utf-8"?>
<Error xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <Code>ResourceNotFound</Code>
    <Message>The specified resource does not exist.</Message>
    <Details>'' isn't a valid resource name.</Details>
</Error>. Detail: PackedStorageIO: IOError: Failed to open output stream: Unknown error: Check for Hierarchical Namespace support on 'https://milvusdemosa.blob.core.windows.net:443/milvus' failed: N5Azure4Core11Credentials23AuthenticationExceptionE: GetToken(): error response: 404 Not Found

<?xml version="1.0" encoding="utf-8"?>
<Error xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
[...]

PR #52127, opened 2026-08-03
fix: upgrade azure-sdk-for-cpp to 1.16.4

jiaqizho:
issue: #52033
Pin the published 1.16.4 Conan recipe revision. Azure Core 1.16.4 contains the CURL transport lifetime fix that destroys the easy handle before the shared SSL handle, resolving the ASAN failure seen with 1.16.1.


jiaqizho:
Need CI update image(upgrade azurite)

liliu-z (commented):
/lgtm
/approve

Issue #50892, opened 2026-06-29
[Bug]: HNSW_PQ/IVF_PQ collection load SIGSEGVs on NVIDIA Grace (Neoverse-V2) — bundled OpenBLAS 0.3.20 crashes under concurrent segment loads


FantasqueX:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 2.6.19 (image milvusdb/milvus:v2.6.19)
  - Deployment mode(standalone or cluster): standalone (docker-compose)
  - MQ type(rocksmq, pulsar or kafka): standalone embedded default (not relevant to this issue)
  - SDK version(e.g. pymilvus v2.0.0rc2): pymilvus 2.6.x
  - OS(Ubuntu or CentOS): Ubuntu 22.04.5 LTS (host + image)
  - CPU/Memory: ARM Neoverse-V2 (NVIDIA Grace), 144 cores / ample free RAM (not OOM)
  - GPU: N/A (CPU index)
  - Others: bundled BLAS is libopenblas0-pthread:arm64 0.3.20+ds-1, linked by libknowhere.so alongside libgomp
```

### Current Behavior

Loading an HNSW_PQ collection built with a high PQ sub-quantizer count (pq_m=128, dim=768, ~10M rows, ~45 sealed segments) **segfaults milvus-standalone (exit code 139)** during `load_collection`. It is deterministic — 4 of 4 load attempts crashed in my testing.

The crash happens inside faiss's per-segment PQ SDC-table computation, which calls OpenBLAS `sgemm`. With `OPENBLAS_VERBOSE=2` set on the container, OpenBLAS logs `Core: armv8` — it does NOT recognize Neoverse-V2 and falls back to the *generic* ARMv8 kernels, and the crash is in that generic `sgemm_incopy_ARMV8` copy kernel.

This is NOT OOM and NOT data-size dependent. The trigger is concurrency: many segments load in parallel, each running `compute_sdc_table` (an OpenMP region) -> many concurrent multithreaded `sgemm` calls contending on OpenBLAS's internal copy buffers. A collection with few segments does not reliably crash; one with many segments crashes ~100%. Because `compute_sdc_table` runs for any PQ-based index, IVF_PQ / SCANN / HNSW_PQ on aarch64 are all potentially affected.

### Expected Behavior

Loading a PQ-based collection on aarch64 / Neoverse-V2 (Grace) should not crash the server.

### Steps To Reproduce

```markdown
[...]

PR #49912, opened 2026-05-19
build(deps): Upgrade knowhere and manage milvus-common via conan

yhmo:
issue: https://github.com/milvus-io/milvus/issues/47425

# Plan: upgrade Milvus to Knowhere with Conan `milvus-common` and Conan OpenBLAS

## Context

Milvus currently fetches both `knowhere` and `milvus-common` from source with CMake `FetchContent`. The requested Knowhere commit `0e53e57cd1f67f74decbdf9af42e74d0e3850155` expects `milvus-common` and Linux OpenBLAS from Conan 2.x instead:

- `milvus-common/1.0.0-9ca5ea6@milvus/dev#274d428d85f1d3d996e1092f0c9c7144`
- `find_package(milvus-common REQUIRED)` and target `milvus-common::milvus-common`
- Linux `openblas/0.3.30` with CMake target `OpenBLAS::OpenBLAS`
- macOS Apple BLAS/LAPACK via `BLA_VENDOR Apple`, not Conan OpenBLAS

Milvus also currently installs system OpenBLAS in builder Dockerfiles, runtime Dockerfiles, dependency scripts, and packaging scripts. The intended outcome is one build-time OpenBLAS source on Linux: Conan. Runtime packaging should use the Conan-built shared libraries instead of relying on distro `libopenblas-dev`/`openblas-devel` where possible.

## Implementation

1. **Update Conan dependency pins in `internal/core/conanfile.py`.**
   - Add direct requirement `milvus-common/1.0.0-9ca5ea6@milvus/dev#274d428d85f1d3d996e1092f0c9c7144`.
   - Align dependencies required by Knowhere `0e53e57c`:
     - `folly/2026.04.20.00@milvus/dev#06852bea5b6449f0c4eb0df002b5779c`
     - `fmt/11.2.0#eb98daa559c7c59d591f4720dde4cd5c`
     - `lz4/1.10.0#982d9b673900f665a1da109e09c17cab`
     - `fast_float/8.0.0@milvus/dev#c7802833c74c5a86ffed70e4af1a795e`
     - Linux-only `openblas/0.3.30`
   - Add `openblas/*:dynamic_arch = True` to default options to match Knowhere.

2. **Move Milvus core build settings to C++20.**
   - Change `internal/core/CMakeLists.txt` from `CMAKE_CXX_STANDARD 17` to `20`.
   - Keep `OPENTELEMETRY_STL_VERSION=2017` unchanged; it controls OpenTelemetry ABI mode, not the language standard.
[...]

PR #52129, opened 2026-08-03
fix: pin dynamic pthread OpenBLAS recipe on Linux

CLiqing:
## Summary

Backport the OpenBLAS 0.3.30 fix to Milvus 2.6 in Conan 1 syntax:

- require the known-good recipe revision `openblas/0.3.30#b818732ec6d4be12891e042d331a7e48` on Linux
- enable `dynamic_arch` for runtime CPU dispatch
- link OpenBLAS statically
- remove the stale `use_openmp` setting; this recipe uses the native pthread backend

Pinning the recipe revision prevents `conan install --update` from selecting the newer OpenMP recipe revision. `dynamic_arch` avoids the generic ARMv8 kernel path observed on AWS Graviton4 / Neoverse-V2.

issue: #50892
pr: #49912

### DIFF
--- a/internal/core/conanfile.py
+++ b/internal/core/conanfile.py
@@ -84,11 +84,12 @@ class MilvusConan(ConanFile):
         "glog:shared": True,
         "prometheus-cpp:with_pull": False,
         "fmt:header_only": True,
+        "openblas:dynamic_arch": True,
+        "openblas:shared": False,
         "onetbb:tbbmalloc": False,
         "onetbb:tbbproxy": False,
         "gdal:shared": True,
         "gdal:fPIC": True,
-        "openblas:use_openmp": True,
     }
 
     def configure(self):
@@ -100,6 +101,8 @@ class MilvusConan(ConanFile):
             self.options["arrow"].with_jemalloc = False
 
     def requirements(self):
+        if self.settings.os == "Linux":
+            self.requires("openblas/0.3.30#b818732ec6d4be12891e042d331a7e48")
         if self.settings.os != "Macos":
             self.requires("libunwind/1.7.2")
         # Override s2n 1.4.1 (from aws-c-io) to 1.6.0 for OpenSSL 3.x FIPS detection
--- a/internal/core/CMakeLists.txt
+++ b/internal/core/CMakeLists.txt
@@ -70,6 +70,12 @@ if (CMAKE_COMPILER_IS_GNUCC)
 endif ()
 
 set( CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake" )
+if (LINUX)
+    # Knowhere v2.6.18 still uses CMake's FindBLAS/FindLAPACK modules.  Add
+    # the Conan OpenBLAS compatibility modules before the system modules.
+    list(PREPEND CMAKE_MODULE_PATH
+         "${CMAKE_CURRENT_SOURCE_DIR}/cmake/openblas_compat")
+endif()
 include( Utils )
 
 # **************************** Build time, type and code version ****************************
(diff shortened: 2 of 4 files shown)


CLiqing:
## Investigation update

The current build failure is confirmed to be dependency resolution, not a source build failure:

```text
Unable to find openblas/0.3.30 in remotes
```

Milvus 2.6 still uses the Conan 1 `default-conan-local` remote, where the highest available OpenBLAS recipe is `0.3.27`. The master-aligned `0.3.30` package is available from the Conan 2 `default-conan-local2` remote.

`0.3.27` is not an equivalent fallback: it predates OpenBLAS PR #4702 (correct Neoverse-V2 detection, included from 0.3.28), and its standard recipe does not provide the current `use_openmp` option.

Follow-up: the compatible OpenBLAS 0.3.30 package/recipe for the 2.6 Conan 1 remote will be handled separately. This PR should be revisited after that artifact is available.

CLiqing:
The Conan 1 `openblas/0.3.30` recipe has been updated and verified locally with the 2.6 dependency graph. It now accepts `dynamic_arch=True`, `shared=False`, `use_openmp=True`, and `use_thread=True`, maps OpenMP to the OpenBLAS CMake build, and exports the GCC OpenMP runtime.

Rebased this PR onto the latest 2.6 head and rerunning build / ARM e2e.

internal/core/conanfile.py
```diff
@@ -84,11 +84,12 @@ class MilvusConan(ConanFile):
         "glog:shared": True,
         "prometheus-cpp:with_pull": False,
         "fmt:header_only": True,
+        "openblas:dynamic_arch": True,
```
liliu-z, line 87: Enabling `openblas:dynamic_arch` is unsafe on arm64 with GCC < 10, because the pinned recipe revision builds OpenBLAS through its CMake path: `cmake/arch.cmake:37-40` only adds NEOVERSEV1/NEOVERSEN2/ARMV8SVE/A64FX to `DYNAMIC_CORE` when the compiler is GCC >= 10, yet nothing in the CMake path ever defines `NO_SVE` (only `NO_SME` is auto-probed, in `cmake/system_check.cmake:142-151`), so `driver/others/dynamic_arm64.c` still declares those four `gotoblas_*` tables extern and takes their addresses at lines 234-240 and 272-278 while `kernel/CMakeLists.txt:1376-1381` never emits them. The ARM CI and release pipelines build inside `milvus-env:ubuntu20.04` (`ci/jenkins/PR-Arm.groovy:64`, `ci/jenkins/PublishArmBasedImages.groovy:60`), whose base image `ubuntu:focal-20220426` ships gcc 9.4.0, and `scripts/3rdparty_build.sh:161` passes `--build=missing`, so an arm64 build that has to compile this brand-new recipe-revision/option-set from source dies with undefined references to `gotoblas_NEOVERSEV1` et al. rather than degrading gracefully. Note the original claim that only two arm64 builder images are published is inaccurate — `publish-builder.yaml` builds `linux/arm64` for all four (ubuntu22.04/GCC 12 and amazonlinux2023/GCC 11 are unaffected) — but the arm64 path Milvus actually exercises is the affected ubuntu20.04 one, so either move that builder to GCC 10+, thread `NO_SVE` through the recipe, or confirm a prebuilt arm64 binary for this exact revision and option set exists in `default-conan-local` first.

CLiqing: Thanks for raising this. We verified that both the current Milvus 2.6 ARM builder and the ARM artifact produced by this PR use GCC 12.3, so the OpenBLAS DYNAMIC_ARCH issue with GCC 9 does not affect the supported release pipeline. GCC 9 source builds are outside the scope of this backport.
We prefer to keep OpenBLAS 0.3.30 to stay aligned with master, which also uses the pthread backend. The reported pthread race requires a specific concurrent large-GEMM timing pattern, and no corresponding Milvus/Knowhere failure has been reported. A broader OpenBLAS upgrade can be handled separately.

liliu-z (commented):
/lgtm
/approve

PR #52131, opened 2026-08-03
test: strengthen add function field search readiness coverage

lyyyuna:
## Summary

- Align add-function-field search polling with the schema-evolution readiness policy.
- Record transient search errors and empty-result timing, then require a 2-second post-hit wait followed by five successful fixed verification rounds.
- Require exact ordered topK IDs for whole-result comparisons and fail immediately on non-empty unexpected results.
- Add regression coverage for the #51366 path with a post-created base index and 512 total rows.

Related: #51366

## Test Plan

- `git diff --check`
- `ruff check tests/python_client/milvus_client/test_add_function_field_feature.py`
- `ruff format --check tests/python_client/milvus_client/test_add_function_field_feature.py`
- `python -m py_compile tests/python_client/milvus_client/test_add_function_field_feature.py`
- Targeted E2E on `schema-evolution-v3-latest` using `3.0-20260810-e2820e27`: 3 passed in 320.79s
- Full `test_add_function_field_feature.py` run after the 2-second wait change on `schema-evolution-v3-latest`: 23 passed in 900.08s

### DIFF
--- a/tests/python_client/milvus_client/test_add_function_field_feature.py
+++ b/tests/python_client/milvus_client/test_add_function_field_feature.py
@@ -1509,7 +1509,7 @@ class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
 
         self.drop_collection(client, collection_name)
 
-    @pytest.mark.tags(CaseLabel.L1)
+    @pytest.mark.tags(CaseLabel.L2)
     def test_add_bm25_function_field_drop_function_field_combo(self):
         """
         TC-10: Add Function Field with Drop Function / Drop Field combination.
@@ -1828,7 +1828,7 @@ class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
         client.drop_alias(reverse_alias_name)
         self.drop_collection(client, reverse_collection_name)
 
-    @pytest.mark.tags(CaseLabel.L1)
+    @pytest.mark.tags(CaseLabel.L2)
     def test_add_bm25_function_field_after_delete_upsert_history(self):
         """
         TC-13: Delete / Upsert before add_function_field.
@@ -2034,7 +2034,7 @@ class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
 
         self.drop_collection(client, collection_name)
 
-    @pytest.mark.tags(CaseLabel.L1)
+    @pytest.mark.tags(CaseLabel.L2)
     def test_add_function_field_matches_existing_bm25_and_minhash_output(self):
         """
         TC-15: BM25 / MinHash old-vs-added function output equivalence.
@@ -2899,7 +2899,7 @@ class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
 
         self.drop_collection(client, collection_name)
 
-    @pytest.mark.tags(CaseLabel.L1)
+    @pytest.mark.tags(CaseLabel.L2)
     def test_add_bm25_function_field_with_multi_analyzer_by_field(self):
         """
         TC-24: Multi-analyzer / by_field BM25 input.
@@ -3041,7 +3041,7 @@ class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
 
         self.drop_collection(client, collection_name)
 
-    @pytest.mark.tags(CaseLabel.L1)
+    @pytest.mark.tags(CaseLabel.L2)
     def test_add_function_field_function_type_support_matrix(self):
         """
         TC-26: FunctionType support matrix.
--- a/tests/python_client/milvus_client/test_add_function_field_feature.py
+++ b/tests/python_client/milvus_client/test_add_function_field_feature.py
@@ -20,6 +20,76 @@ from utils.util_log import test_log as log
 
 
 class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
+    @staticmethod
+    def _search_hit_ids_match(hit_ids, expected_id, exact_ids=False, expected_ids=None):
+        if expected_ids is not None:
+            return bool(hit_ids) and all(hit_id in expected_ids for hit_id in hit_ids)
+        if exact_ids:
+            return hit_ids == [expected_id]
+        return bool(hit_ids) if expected_id is None else expected_id in hit_ids
+
+    @staticmethod
+    def _is_known_backfill_type_mismatch(exc):
+        message = str(exc).lower()
+        return "vector_sparse_u32_f32 vs varchar" in message or (
+            "vector type must be the same" in message and "vector_sparse_u32_f32" in message and "varchar" in message
+        )
+
+    @classmethod
+    def _search_error_category(cls, exc):
+        message = str(exc).lower()
+        if cls._is_known_backfill_type_mismatch(exc):
+            return "known_backfill_type_mismatch"
+        if "field index" in message and "not loaded" in message:
+            return "field_index_not_loaded"
+        if "field index meta not found" in message:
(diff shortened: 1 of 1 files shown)


tests/python_client/milvus_client/test_add_function_field_feature.py
```diff
@@ -3255,6 +3353,115 @@ def assert_minhash_hit(collection_name, query_text, expected_id, label, timeout=
...
+            FieldSchema("sparse_bm25", DataType.SPARSE_FLOAT_VECTOR),
+            Function(
+                name="bm25_repro_fn",
+                function_type=FunctionType.BM25,
+                input_field_names=["text"],
+                output_field_names=["sparse_bm25"],
+            ),
+            index_params=sparse_index,
+        )
+        client.load_collection(collection_name)
```
yanliang567: Could we remove this second `load_collection` call? The collection was already loaded before `add_function_field`, while the test docstring and #51366 expect the new field to become searchable without a user-triggered reload. Keeping this call means a regression where schema evolution only becomes visible after another load request would still pass, so the `post-create base index + no reload` combination is not actually covered. TC-14 covers no-reload only when the base index is supplied during collection creation.

tests/python_client/milvus_client/test_add_function_field_feature.py
```diff
@@ -3255,6 +3353,115 @@ def assert_minhash_hit(collection_name, query_text, expected_id, label, timeout=
...
+        assert stats["row_count"] == total_row_count
+
+        # Function-output backfill and QueryNode visibility are not represented by
+        # describe_index readiness. Poll the actual BM25 search directly.
+        search_result = self.wait_for_search_hit(
+            client,
+            collection_name,
+            data=[query_text],
+            anns_field="sparse_bm25",
+            expected_id=None,
```
yanliang567, line 3537: Using `expected_id=None` makes `wait_for_search_hit` treat any non-empty result as ready, and `_assert_post_hit_search_rounds` then checks only `bool(hit_ids)` in all five rounds. The range assertion below validates only the final round, so unexpected IDs during the first hit or first four verification rounds are accepted, contrary to the PR's fail-immediately policy. Please pass an expected-ID set or validator (for example, non-empty and every ID in `[30000, 30500)`) and apply it to every polling and post-hit round.

yanliang567 (commented):
I found two verification gaps in the new #51366 regression coverage and post-hit readiness checks. Please see the inline comments.

tests/python_client/milvus_client/test_add_function_field_feature.py
lyyyuna: Removed the second `load_collection` after `add_function_field`. TC-35 now keeps the collection loaded from before the schema change and verifies that the added BM25 field becomes searchable without a user-triggered reload. The updated test passed 3 consecutive runs on `schema-evolution-v3-latest`.

tests/python_client/milvus_client/test_add_function_field_feature.py
lyyyuna: Added `expected_ids` validation to the shared search-wait helper. The first successful poll and every fixed post-hit round now require a non-empty result whose IDs are all in the expected range; empty results continue polling, while any non-empty out-of-range or mixed result fails immediately. TC-35 passes `range(30000, 30500)`, and the matching branches were checked directly in addition to 3 consecutive TC-35 runs.

tests/python_client/milvus_client/test_add_function_field_feature.py
```diff
@@ -20,6 +20,76 @@
 
 
 class TestMilvusClientAddFunctionFieldFeature(TestMilvusClientV2Base):
+    @staticmethod
+    def _search_hit_ids_match(hit_ids, expected_id, exact_ids=False, expected_ids=None):
+        if expected_ids is not None:
+            return bool(hit_ids) and all(hit_id in expected_ids for hit_id in hit_ids)
```
yanliang567, line 35: Could this path also validate the expected result count? TC-35 inserts 500 documents containing the query token and searches with `limit=5`, but this predicate accepts a single hit such as `[30000]` as ready. If only a subset of function outputs is visible, the initial poll and all five post-hit rounds can pass. Please require `len(hit_ids) == limit` (or an explicit expected count) for this path in addition to validating membership, so partial visibility is not reported as ready.

tests/python_client/milvus_client/test_add_function_field_feature.py
```diff
@@ -3255,6 +3369,115 @@ def assert_minhash_hit(collection_name, query_text, expected_id, label, timeout=
...
+            ],
+        )
+        client.flush(collection_name)
+
+        stats = client.get_collection_stats(collection_name)
+        assert stats["row_count"] == total_row_count
+
+        # Function-output backfill and QueryNode visibility are not represented by
+        # describe_index readiness. Poll the actual BM25 search directly.
+        search_result = self.wait_for_search_hit(
```
yanliang567, line 3532: Could we wait for and assert the sparse index state before starting this search poll? The #51366 reproduction reaches `state=Finished` with `pending_index_rows=0` before search remains broken. Here search starts immediately after the flush; with 512 rows below the physical-index threshold, searchable raw data could make the test pass without exercising that post-index-ready state. Please wait for `sparse_bm25` readiness and assert its state/row counters before polling search.

yanliang567 (commented):
I found two remaining verification gaps in the new #51366 regression coverage: partial result visibility can still be accepted as ready, and the test does not establish the sparse-index-ready precondition from the original reproduction. Please see the inline comments.

tests/python_client/milvus_client/test_add_function_field_feature.py
lyyyuna: Fixed in afaf9b422c. The shared readiness helper now accepts an explicit expected_count and applies it to both the initial successful poll and every fixed post-hit verification round. TC-35 passes search_limit=5, so a partial non-empty result fails immediately unless it contains exactly 5 IDs and every ID is in the expected range.

tests/python_client/milvus_client/test_add_function_field_feature.py
lyyyuna: Fixed in afaf9b422c. After the 500-row flush, TC-35 now waits for sparse_bm25 readiness and explicitly asserts state=Finished, total_rows=512, indexed_rows=512, and pending_index_rows=0 before starting the search poll. The collection is still not released or reloaded. The updated TC-35 passed on schema-evolution-v3-latest via port-forward: 1 passed in 43.04s.

lyyyuna:
Follow-up on the `time.sleep(10)` -> `time.sleep(2)` change:

I reran the full `test_add_function_field_feature.py` file serially against the same `schema-evolution-v3-latest` environment through port-forward.

- Before: **23 passed in 1324.57s**
- After: **23 passed in 900.08s**
- Observed reduction: **424.49s (~7m 04s, 32.0%)**

The readiness polling and the 5 strict post-hit search rounds remain unchanged; only the fixed settling wait was shortened.

lyyyuna:
Updated the test case levels to keep the pre-merge suite focused while retaining the extended combinations in nightly coverage.

Level distribution:

| Level | Before | After |
|---|---:|---:|
| L0 | 3 | 3 |
| L1 | 13 | 8 |
| L2 | 7 | 12 |
| Total | 23 | 23 |

Five extended combination/matrix cases were moved from L1 to L2:

- Add Function Field + Drop Function/Field lifecycle
- Delete/Upsert history before backfill
- Existing-vs-added BM25/MinHash output equivalence
- Multi-analyzer by-field behavior
- FunctionType support matrix

The nullable-input backfill case remains L1 so nullable historical-data handling is still covered by the pre-merge CI.

Issue #48957, opened 2026-04-13
[Enhancement]: V3 streaming load for scalar indexes to reduce peak memory


sparknack:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

V3 format scalar indexes (ScalarIndexSort, StringIndexSort, StringIndexMarisa, BitmapIndex) currently load the entire index data into memory at once before parsing/mmapping, causing high peak memory usage during load.

This issue tracks a set of related enhancements to support **streaming load** for V3 scalar indexes, plus persisting auxiliary data structures (like \`idx_to_offsets\` and CSR tables) into the V3 format so mmap mode can map them directly instead of reconstructing them in heap memory.

### Why is this needed?

- **Reduce peak memory during load**: streaming entries with a sliding window avoids the 2x transient copy
- **Reduce steady-state memory in mmap mode**: persisting auxiliary structures (e.g. \`idx_to_offsets\`, CSR \`str_ids_to_offsets_\`) lets the OS page cache manage them instead of keeping 4GB+ vectors on the heap at 1B rows
- **Consistency**: unify the V3 load pattern across all scalar index types

### Scope

1. Add \`ReadEntryStream\` API with sliding window
2. Streaming V3 \`LoadEntries\` for ScalarIndexSort / StringIndexSort / StringIndexMarisa / BitmapIndex
3. Persist and mmap \`idx_to_offsets\` + \`valid_bitset\` in ScalarIndexSort V3
4. Flatten \`str_ids_to_offsets_\` to CSR and persist/mmap in StringIndexMarisa V3
5. Zero-copy non-mmap path for StringIndexSort via \`LoadFromBuffer\`
6. Persist and mmap \`idx_to_offsets\` in StringIndexSort V3
7. Update resource estimates accordingly

PR #52132, opened 2026-08-03
enhance: separate and cache index load resource estimates

sparknack:
issue: #48957

- Separate metadata-only index load estimates from scalar file inspection and resource planning.
- Reuse cached estimates only for vector and legacy scalar paths that do not require file context.
- Preserve packed scalar V3 and hybrid file inspection, including shared stream runtime-unit planning, and update callers and tests.

### DIFF
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -25,6 +25,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "common/Consts.h"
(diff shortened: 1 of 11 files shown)


internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp
```diff
@@ -92,66 +80,74 @@ SealedIndexTranslator::SealedIndexTranslator(
...
-            index_load_info_.num_rows,
-            index_load_info_.dim,
-            index_load_info_.index_files,
-            file_manager_context_,
-            stream_load_info,
-            use_shared_memory_overhead_group);
     if (index_load_info_.load_resource_request.has_value()) {
-        return *index_load_info_.load_resource_request;
+        load_resource_request_ = *index_load_info_.load_resource_request;
     }
```
yhmo, line 150: internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp line:150
High ---- This unconditional assignment restores the cached metadata-only request after the V3/hybrid path has inspected the files and stored `plan.request` above. As a result, any caller that supplies a cache still gets stale resident/peak costs in `estimated_byte_size_of_cell`, so encrypted or otherwise file-specific transient memory can be under-reserved despite the inspection. Please keep the cached request only in the early-return branch and add a successful V3 file-context test that verifies the planned resource values win.

internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp
```diff
@@ -92,66 +80,74 @@ SealedIndexTranslator::SealedIndexTranslator(
...
-            index_load_info_.enable_mmap,
-            index_load_info_.num_rows,
-            index_load_info_.dim,
-            index_load_info_.index_files,
-            file_manager_context_,
-            stream_load_info,
-            use_shared_memory_overhead_group);
     if (index_load_info_.load_resource_request.has_value()) {
-        return *index_load_info_.load_resource_request;
+        load_resource_request_ = *index_load_info_.load_resource_request;
```
buqian-zilliz, line 149: **[P2] Do not overwrite the inspected load plan with the cached estimate**

The early guard only reuses a cached request when file inspection is unnecessary, but this trailing assignment unconditionally restores the cache after `PlanScalarIndexLoad()` has produced the file-aware V3/HYBRID request. Any producer that supplies a cached admission estimate for such a load will make `estimated_byte_size_of_cell()` use stale resident and peak costs, potentially under-reserving encrypted-stream memory or losing the resolved HYBRID subtype estimate. Remove this trailing override; the early-return branch is the only safe place to consume the cache.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -316,6 +362,194 @@ InspectScalarIndexStreamLoadInfo(
...
+IndexType
+ScalarIndexTypeFromSpec(const IndexLoadSpec& spec) {
+    auto index_type_it = spec.index_params.find("index_type");
+    AssertInfo(index_type_it != spec.index_params.end(), "index type is empty");
+    return index_type_it->second;
+}
+
+int32_t
+ScalarIndexVersionFromSpec(const IndexLoadSpec& spec) {
+    auto config = ParseConfigFromIndexParams(spec.index_params);
```
buqian-zilliz, line 374: **[P3] Avoid rebuilding the JSON config in every planning phase**

`ScalarIndexVersionFromSpec()` materializes a complete nlohmann `Config` from the string map on every call. `BuildCache()` invokes it once through `RequiresFileContextForLoadResource()` and again through estimation, while the file-aware translator invokes it independently from the requires, inspect, and plan phases. This adds repeated map copies and JSON allocations for every scalar index during segment load or reopen. Read the version directly from `index_params`, or parse it once and carry it through the planning phases.

PR #52134, opened 2026-08-03
enhance: bump schema version for property schema WAL updates

congqixia:
Ensure alter collection property changes that carry schema payloads advance schema.Version before writing the AlterCollection message to WAL. Cover TTL field and external source/spec updates, and guard against schema update masks without a newer schema payload.

Related to #51796

### DIFF
--- a/internal/rootcoord/ddl_callbacks_alter_collection_properties.go
+++ b/internal/rootcoord/ddl_callbacks_alter_collection_properties.go
@@ -123,20 +123,24 @@ func (c *Core) broadcastAlterCollectionForAlterCollection(ctx context.Context, r
 				header.UpdateMask.Paths = append(header.UpdateMask.Paths, message.FieldMaskCollectionConsistencyLevel)
 			}
 		case common.CollectionExternalSource:
-			if udpates.Schema == nil {
-				udpates.Schema = &schemapb.CollectionSchema{}
-			}
-			udpates.Schema.ExternalSource = prop.GetValue()
-			if !funcutil.SliceContain(header.UpdateMask.Paths, message.FieldMaskCollectionExternalSpec) {
-				header.UpdateMask.Paths = append(header.UpdateMask.Paths, message.FieldMaskCollectionExternalSpec)
+			if prop.GetValue() != coll.ExternalSource {
+				if udpates.Schema == nil {
+					udpates.Schema = &schemapb.CollectionSchema{}
+				}
+				udpates.Schema.ExternalSource = prop.GetValue()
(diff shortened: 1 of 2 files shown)


internal/rootcoord/ddl_callbacks_alter_collection_properties.go
```diff
@@ -178,16 +185,18 @@ func (c *Core) broadcastAlterCollectionForAlterCollection(ctx context.Context, r
 			header.UpdateMask.Paths = append(header.UpdateMask.Paths, message.FieldMaskCollectionSchema)
 		}
 
-		// Build schema snapshot with updated properties (schema version should NOT be changed for properties-only alter).
+		// Build schema snapshot with updated properties. Any schema payload
+		// change must advance schema.Version so WAL consumers can gate on it.
 		schema := coll.ToCollectionSchemaPB()
+		schema.Version = coll.SchemaVersion + 1
```
liliu-z, line 191: Bumping schema.Version here makes TTL/external property alters advance the collection schema version even though no field changes. On a function collection, datacoord's index gate defers index builds for any segment whose SchemaVersion trails the collection's, and a schema change force-flushes all in-flight growing segments at the old version, so every segment resident at alter time falls behind. For a non-external function collection the only catch-up path is bump-schema-version compaction, which defaults to false and is not forced on for functions declared at create_collection time, so those index builds are stranded permanently and that data stays unsearchable. Segments allocated after the alter carry the new version and index normally (the collection is not bricked), and external collections self-heal on the next external refresh; the durable failure is the TTL/external alter on a non-external function collection with bump compaction off. Fix by making the gate compare function-output field sets instead of raw versions, or by rejecting/skipping the version bump on function collections when bump compaction is disabled (as add-function already does).

liliu-z, line 191: Advancing schema.Version on a TTL/external alter now causes concurrent inserts carrying the prior version to be rejected with ErrCollectionSchemaMismatch; before this PR these alters kept the same version and did not reject writes. On a multi-shard collection the WAL append across vchannels is non-atomic, so a version bump landing mid-commit can commit some vchannels while others reject; the proxy returns the first error without rolling back the committed subset, and the bundled Go SDK's retryIfSchemaError re-runs the entire Insert/Upsert, duplicating the already-committed rows with fresh auto-IDs. The duplication machinery is pre-existing (also reachable via dynamic-field and add-field bumps) and needs a narrow race between the broadcast and an in-flight multi-vchannel insert, so this is worth flagging rather than gating the PR.

Issue #52139, opened 2026-08-03
[Bug]: Streaming QueryNodes can block sealed segment balancing in channel-exclusive mode


weiliu1031:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: reproduced against the current master code path
- Deployment mode: cluster
- MQ type: Streaming Service enabled
- SDK version: N/A
- OS: N/A
- QueryCoord balancer: ChannelLevelScoreBalancer
- Channel exclusive mode: enabled
```

### Current Behavior

When Streaming Service and `ChannelLevelScoreBalancer` are enabled,
valid Streaming QueryNodes can prevent sealed segments from being
balanced between normal QueryNodes.

For example:

```text
Normal QueryNodes:
  rwNodes = [344, 345, 346, 347]

Streaming QueryNodes:
  rwSQNodes = [339, 333]
```

Nodes `339` and `333` are valid RWSQ nodes that hold channels and act as
delegators. Sealed segments are loaded on nodes `344` through `347`.

For each channel, `ChannelLevelScoreBalancer` obtains its normal
QueryNodes from `GetChannelRWNodes()` and compares them with the channel
distribution:

```text
channelRwNodes = [344, 345, 346, 347]
channelDist    = [339, 333]
channelRONodes = [339, 333]
```

The shared outbound-node list therefore classifies valid RWSQ channel
holders as outbound nodes.

With Streaming Service enabled, channel moves are skipped in this
branch. The balancer then attempts to move sealed segments from nodes
`339` and `333`. Because those nodes hold no sealed segments, it
generates an empty plan.

Normal `genSegmentPlan()` processing is in the alternative branch, so
it is never reached. Sealed segment imbalance between normal
QueryNodes can therefore persist indefinitely.

### Expected Behavior

Channel holders and sealed-segment workers should use independent node
sets.

- Valid RWSQ channel holders should not be classified as sealed-segment
  outbound nodes solely because they appear in channel distribution.
- Sealed segments outside their channel-exclusive normal QueryNode set
  should retain outbound migration priority.
- If no executable outbound segment plan is generated, normal segment
  balancing should continue between the channel's normal QueryNodes.

### Steps To Reproduce

```markdown
1. Enable Streaming Service.
2. Enable ChannelLevelScoreBalancer and channel-exclusive mode.
3. Create a replica with normal QueryNodes 344-347 and RWSQ nodes
   339 and 333.
[...]

PR #52140, opened 2026-08-03
fix: prevent streaming query nodes from blocking segment balance

weiliu1031:
issue: #52139

### What changed

- Separate channel outbound nodes from sealed-segment outbound nodes in
  `ChannelLevelScoreBalancer`.
- Leave Streaming QueryNode channel placement to score-based channel
  balance and stopping balance.
- Preserve migration priority for sealed segments outside their
  channel-exclusive normal QueryNode set.
- Continue normal segment balancing when outbound segments cannot
  produce an executable move plan.
- Add Streaming Service regression coverage and update the
  channel-exclusive design document.

### Why

With Streaming Service enabled, valid RWSQ nodes hold channels while
sealed segments reside on normal QueryNodes. The previous shared
outbound-node list compared both distributions against the normal
QueryNode set.

This classified RWSQ channel holders as outbound nodes. Streaming mode
then skipped channel movement and attempted to move sealed segments
from those RWSQ nodes. When they held no sealed segments, the result was
an empty plan, and normal segment balancing was never reached.

### Behavior boundary

- Streaming channel placement remains handled by score-based channel
  balance and stopping balance.
- A sealed segment outside its channel-exclusive normal QueryNode set
  is still treated as outbound workload.
- Executable outbound segment plans retain priority over normal segment
  balancing.
- When no executable outbound segment plan exists, normal segment
  balancing continues.

### Validation

Not run for this exact candidate (`check-commit` unavailable).

### DIFF
--- a/docs/design-docs/design_docs/20260204-channel_exclusive_mode.md
+++ b/docs/design-docs/design_docs/20260204-channel_exclusive_mode.md
@@ -340,14 +340,16 @@ BalanceReplica(ctx, replica)
 3. Channel-Aware Balancing (Exclusive Mode Enabled)
     For each channel:
         ├─ Get channel's exclusive RW nodes
-        ├─ Identify outbound nodes (nodes no longer in RW node list)
-        ├─ If outbound nodes exist:
-        │  ├─ genChannelPlanForOutboundNodes()
+        ├─ Identify channel and segment outbound nodes independently
+        ├─ In streaming mode, leave channel placement to streaming and stopping balance
+        ├─ If channel outbound nodes exist:
+        │  └─ genChannelPlanForOutboundNodes()
+        ├─ If segment outbound nodes exist:
         │  └─ genSegmentPlanForOutboundNodes()
-        └─ If no outbound nodes:
+        └─ If no channel outbound exists and no executable segment outbound plan exists:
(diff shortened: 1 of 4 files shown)


chyezh:
One test-stability concern in `internal/datanode/index/task_test.go` around the updated Mockey callback:

The callback stores the `*clusteringpb.AnalyzeInfo` argument in `captured` and inspects it after `task.Execute` and the mocked call have returned. With runtime patching, retaining a pointer argument outside the callback can preserve a pointer to a non-escaping temporary from the original call site, making the assertion compiler-dependent or flaky.

Since this test only needs `MaxConnections`, could we snapshot the scalar inside the callback instead?

```go
var capturedMaxConnections uint32

patch := mockey.Mock(analyzecgowrapper.Analyze).To(
    func(
        _ context.Context,
        info *clusteringpb.AnalyzeInfo,
        _ *indexcgopb.StoragePluginContext,
    ) (analyzecgowrapper.CodecAnalyze, error) {
        capturedMaxConnections = info.GetStorageConfig().GetMaxConnections()
        return nil, nil
    },
).Build()
```

Then assert `capturedMaxConnections == 237`. If the full message is needed, cloning it inside the callback would also avoid retaining the original pointer.

Issue #52135, opened 2026-08-03
[Bug]: REST v2 omits Struct Array element offsets from element-level hybrid search results


zhuwenxing:
## Environment

- Milvus version: `master-20260803-f929d68ba6`
- Git commit: `f929d68ba6` (from the exact affected standalone Pod startup banner)
- Build time: `Mon Aug 3 07:03:30 UTC 2026`
- Deployment mode: standalone
- MQ type: Woodpecker
- SDK version: N/A for reproduction (RESTful v2 with `requests 2.34.2`); PyMilvus `3.0.1rc71` was used for parity verification
- OS: Ubuntu 22.04.4 LTS
- CPU/Memory: 2 CPU / 8 GiB
- GPU: N/A
- Kubernetes: v1.34.1-tke.5

### K8s Pod List

```text
NAME                                                            READY   STATUS    RESTARTS   AGE
frame-struct-hybrid-master-etcd-0                               1/1     Running   0          49m
frame-struct-hybrid-master-milvus-standalone-64f98df459-94txn   1/1     Running   0          49m
frame-struct-hybrid-master-minio-777c7c6db4-77wdz               1/1     Running   0          49m
```

## Current Behavior

RESTful v2 successfully executes an element-level hybrid search on a Struct Array vector subfield, but omits `SearchResultData.element_indices` from the JSON response.

The request returns HTTP 200 and `code: 0`, but callers cannot determine which Struct Array element produced the selected result.

Actual response:

```json
{
  "code": 0,
  "cost": 0,
  "data": [
    {
      "distance": 1,
      "id": 1
    }
  ],
  "topks": [1]
}
```

The selected element is `objects[1]`, but no element offset is returned.

For the same type of result, PyMilvus reads `SearchResultData.element_indices` and exposes it as:

```json
{
  "frame_id": 102,
  "distance": 0.5,
  "offset": 1
}
```

## Expected Behavior

RESTful v2 should preserve the element index returned by the search pipeline and expose it in each element-level search hit, for example:

```json
{
  "distance": 1,
  "id": 1,
  "offset": 1
}
```

This is required to identify which object inside `Array<Struct>` produced the result.

## Steps To Reproduce

1. Start a master Milvus build supporting Struct Array element-level hybrid search.
2. Install the Python dependency:

   ```bash
   pip install requests
   ```

3. Set `MILVUS_URI` and `MILVUS_TOKEN`.
4. Run the script below.
[...]

PR #52143, opened 2026-08-03
fix: return element offsets in REST v2 search results

zhuwenxing:
issue: #52135

## What this PR does

- Exposes `SearchResultData.element_indices` as `offset` on each REST v2 search hit.
- Covers both `/v2/vectordb/entities/search` and `/v2/vectordb/entities/hybrid_search`.
- Leaves non-element-level search responses unchanged.
- Returns an internal error if element offsets are not aligned with serialized hits.
- Adds handler, conversion, and RESTful v2 E2E regression coverage.

## Why

The search pipeline preserves Struct Array element indices through search, hybrid reranking, and group-by. The REST v2 response conversion previously serialized IDs, scores, and output fields without element indices, so callers could identify the row but not the matched Struct Array element.

## Tests

- `TestSearchV2ReturnsElementOffsets`
- `TestBuildSearchResp`
- Full `internal/distributed/proxy/httpserver` Go unit-test package
- `tests/restful_client_v2/testcases/test_vector_operations.py::TestHybridSearchVector::test_hybrid_search_struct_array_vector_returns_element_offset[128]`

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -2059,12 +2059,13 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(nil), HTTPReturnData: []interface{}{}, HTTPReturnCost: cost})
 		} else {
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
-			outputData, err := buildSearchResp(searchResp.Results, allowJS, collSchema)
-			if err != nil {
-				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
+			outputData, buildErr := buildSearchResp(searchResp.Results, allowJS, collSchema)
+			if buildErr != nil {
+				err = buildErr
+				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(buildErr))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + buildErr.Error(),
 				})
 			} else {
 				if len(searchResp.Results.Recalls) > 0 {
@@ -2231,12 +2232,13 @@ func (h *HandlersV2) advancedSearch(ctx context.Context, c *gin.Context, anyReq 
 			HTTPReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(nil), HTTPReturnData: []interface{}{}, HTTPReturnCost: cost})
 		} else {
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
-			outputData, err := buildSearchResp(searchResp.Results, allowJS, collSchema)
-			if err != nil {
-				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
+			outputData, buildErr := buildSearchResp(searchResp.Results, allowJS, collSchema)
+			if buildErr != nil {
+				err = buildErr
+				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(buildErr))
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
-					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
+					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + buildErr.Error(),
 				})
 			} else {
 				if proxy.Params.QueryNodeCfg.StorageUsageTrackingEnabled.GetAsBool() && isValid {
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -1848,6 +1848,24 @@ func invalidSearchResultStatus(buildErr error) *commonpb.Status {
 	return status
 }
 
+func searchResultLogFields(results *schemapb.SearchResultData) []mlog.Field {
+	if results == nil {
+		return []mlog.Field{mlog.Bool("resultNil", true)}
+	}
+	return []mlog.Field{
+		mlog.Int64("nq", results.GetNumQueries()),
+		mlog.Int64("topK", results.GetTopK()),
+		mlog.Int("rowCount", len(results.GetScores())),
+		mlog.Int("outputFieldCount", len(results.GetFieldsData())),
+		mlog.Int("aggregationBucketCount", len(results.GetAggBuckets())),
+	}
+}
+
+func logSearchResultBuildFailure(ctx context.Context, message string, results *schemapb.SearchResultData, buildErr error) {
+	fields := append(searchResultLogFields(results), mlog.Err(buildErr))
+	mlog.Warn(ctx, message, fields...)
+}
+
 func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbName string) (interface{}, error) {
 	httpReq := anyReq.(*SearchReqV2)
 	req := &milvuspb.SearchRequest{
@@ -2040,7 +2058,7 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
 			outputData, err := buildSearchAggregationResp(searchResp.Results, allowJS, collSchema)
 			if err != nil {
-				mlog.Warn(ctx, "high level restful api, fail to deal with search aggregation result", mlog.Any("result", searchResp.Results), mlog.Err(err))
+				logSearchResultBuildFailure(ctx, "high level restful api, fail to deal with search aggregation result", searchResp.Results, err)
 				HTTPReturn(c, http.StatusOK, gin.H{
 					HTTPReturnCode:    merr.Code(merr.ErrInvalidSearchResult),
 					HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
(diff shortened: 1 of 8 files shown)


internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2847,6 +2847,41 @@ func buildQueryResp(rowsNum int64, needFields []string, fieldDataList []*schemap
...
+	}
+	if len(elementOffsets) != len(rows) {
+		return nil, merr.WrapErrServiceInternalMsg(
+			"search result element_indices length (%d) does not match row count (%d)",
+			len(elementOffsets), len(rows),
+		)
+	}
+
+	for i, offset := range elementOffsets {
+		rows[i][HTTPReturnElementOffset] = offset
```
buqian-zilliz, line 2909: **[P2] Avoid overwriting valid fields named offset**

`buildQueryResp` has already flattened schema fields and written the primary key under its configured name before this assignment. Because `offset` is a valid, non-reserved field name, an element-level search on a collection whose primary key or requested field is named `offset` silently replaces that user value with the element index; the response can therefore lose the row identity of the hit. Please use a collision-free response shape or explicitly reserve or reject this name instead of overwriting user data.

zhuwenxing: Fixed in e663cad15a. REST v2 now rejects element-level search results when the reserved `offset` response field conflicts with a primary key or output field, preventing user data from being overwritten. Added helper and handler regression coverage for both collision cases.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2847,6 +2847,49 @@ func buildQueryResp(rowsNum int64, needFields []string, fieldDataList []*schemap
...
+		results.GetScores(),
+		enableInt64,
+		collectionSchema,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	elementOffsets := results.GetElementIndices().GetData()
+	if len(elementOffsets) == 0 {
```
buqian-zilliz: **[P2] Validate present-but-empty element indices**

`ElementIndices` presence, rather than a nonzero data length, is the production marker for an element-level result. With a non-nil but empty `LongArray` and non-empty serialized rows, this early return silently downgrades the response to a normal search and bypasses the length-mismatch error, so clients again receive hits without offsets. Return early only when `results.GetElementIndices() == nil`; otherwise compare its data length with `len(rows)`, including the valid zero-row case.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -2053,7 +2053,7 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(nil), HTTPReturnData: []interface{}{}, HTTPReturnCost: cost})
 		} else {
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
-			outputData, err := buildQueryResp(0, searchResp.Results.OutputFields, searchResp.Results.FieldsData, searchResp.Results.Ids, searchResp.Results.Scores, allowJS, collSchema)
+			outputData, err := buildSearchResp(searchResp.Results, allowJS, collSchema)
```
buqian-zilliz: **[P2] Propagate response-conversion failures**

Because this short declaration is inside a nested block, its `err` shadows the proxy call's outer error. When the new `buildSearchResp` validation rejects an offset collision or misaligned element indices, the handler writes an error JSON but returns the original successful `SearchResults` and `nil`; `wrapperPost` consequently records the request as successful and skips its wrapper-level failure log. Assign to the outer error or return the conversion error explicitly here and at the hybrid-search call site.

internal/distributed/proxy/httpserver/utils.go
buqian-zilliz: This issue still exists for dynamic output fields. When a client requests `offset` and that key is absent from a returned entity's dynamic JSON, `buildQueryResp` omits it from the row, so the row-based collision check passes and the element index is inserted under `offset`. The response then presents metadata as the requested field value. Please reserve or reject the requested output name, including dynamic projections, instead of checking only materialized row keys.

zhuwenxing: Fixed in e4d8ace704. buildSearchResp now rejects the reserved offset output field for element-level results even when the dynamic JSON does not materialize that key. A regression test covers the absent dynamic-field case.

internal/distributed/proxy/httpserver/utils.go
zhuwenxing: Fixed in e4d8ace704. ElementIndices presence is now determined by the non-nil protobuf field, and its data length is validated against the row count, including empty arrays and zero-row results.

internal/distributed/proxy/httpserver/handler_v2.go
zhuwenxing: Fixed in e4d8ace704. Both search and hybrid search now propagate buildSearchResp failures through the outer handler error. Regression tests verify that the wrapper records the conversion failure instead of success.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -2053,12 +2053,13 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(nil), HTTPReturnData: []interface{}{}, HTTPReturnCost: cost})
 		} else {
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
-			outputData, err := buildQueryResp(0, searchResp.Results.OutputFields, searchResp.Results.FieldsData, searchResp.Results.Ids, searchResp.Results.Scores, allowJS, collSchema)
-			if err != nil {
-				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
+			outputData, buildErr := buildSearchResp(searchResp.Results, allowJS, collSchema)
+			if buildErr != nil {
+				err = buildErr
```
buqian-zilliz: **[P2] Record conversion errors as processed failures**

Assigning `buildErr` avoids a success metric, but the handler still returns the original successful `SearchResults` in `resp` while the HTTP body reports `ErrInvalidSearchResult`. `wrapperPost` therefore passes an OK response plus a non-nil error to `ParseMetricLabel`, which classifies a post-search conversion failure as `rejected`, not a processed `fail`; a server-side element-index mismatch is placed in the rejection metric and the fail-only structured log is skipped. Return or record a non-OK response status matching the emitted conversion error here and at the hybrid-search call site so the client response, metrics, and logging agree.

internal/distributed/proxy/httpserver/utils.go
buqian-zilliz: This issue still exists. For Struct Array search results, `reconstructStructFieldDataForSearch` rebuilds `SearchResultData.OutputFields` from materialized field names, so an explicitly requested dynamic `offset` becomes `$meta` before `buildSearchResp` runs. If the returned dynamic JSON does not materialize that key, the row scan also misses it and the element index is still written under `offset`. Please validate against the original requested output fields, or reserve the name before reconstruction, rather than relying on the reconstructed response field list.

zhuwenxing: Fixed in fc7c188dd5. buildSearchResp now receives the original requested output fields from both Search and HybridSearch and checks the reserved `offset` name before injecting element offsets, even when Struct Array reconstruction rewrites Results.OutputFields to `$meta`. Added unit and handler regressions covering this case.

internal/distributed/proxy/httpserver/handler_v2.go
zhuwenxing: Fixed in fc7c188dd5. Both Search and HybridSearch now replace the returned SearchResults.Status with ErrInvalidSearchResult when REST result conversion fails and derive the error and HTTP body from that same status. This makes ParseMetricLabel report fail/system and enables the structured failure log. Added handler tests for field collisions and mismatched element_indices.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -2059,12 +2066,14 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
...
 		} else {
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
-			outputData, err := buildQueryResp(0, searchResp.Results.OutputFields, searchResp.Results.FieldsData, searchResp.Results.Ids, searchResp.Results.Scores, allowJS, collSchema)
-			if err != nil {
-				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
+			outputData, buildErr := buildSearchResp(searchResp.Results, httpReq.OutputFields, allowJS, collSchema)
+			if buildErr != nil {
+				searchResp.Status = invalidSearchResultStatus(buildErr)
+				err = merr.Error(searchResp.Status)
+				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(buildErr))
```
buqian-zilliz: **[P1] Avoid logging client-triggered search payloads**

On a collection where `offset` is a valid field or dynamic key, a client can request output fields such as `["offset", "*"]` and deterministically reach this branch after the proxy returns all requested data. `mlog.Any("result", searchResp.Results)` is plain `zap.Any`: the asynchronous logger truncates only after encoding, while the synchronous core has no per-entry cap. Each request can therefore format and allocate the entire payload and write result IDs, fields, or vectors into logs, creating a data-exposure and CPU/memory/log-volume amplification path here and in hybrid search. Log only shape/count metadata and the conversion error, and reject the reserved selector before result conversion.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -2059,12 +2066,14 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(nil), HTTPReturnData: []interface{}{}, HTTPReturnCost: cost})
 		} else {
 			allowJS, _ := strconv.ParseBool(c.Request.Header.Get(HTTPHeaderAllowInt64))
-			outputData, err := buildQueryResp(0, searchResp.Results.OutputFields, searchResp.Results.FieldsData, searchResp.Results.Ids, searchResp.Results.Scores, allowJS, collSchema)
-			if err != nil {
-				mlog.Warn(ctx, "high level restful api, fail to deal with search result", mlog.Any("result", searchResp.Results), mlog.Err(err))
+			outputData, buildErr := buildSearchResp(searchResp.Results, httpReq.OutputFields, allowJS, collSchema)
```
buqian-zilliz, line 2087: **[P2] Preserve offsets in aggregation top hits**

Element-level ArrayOfVector searches remain valid with `searchAggregation`, and the aggregation reduce preserves distinct rows and their `ElementIndices`. However, this call is only reached in the non-aggregation branch; `HitResult`, `AggHit`, and `buildAggHitsResp` carry only the primary key, score, and fields. `topHits` can therefore return multiple entries for the same primary key from different array elements without exposing which element produced each score. Propagate the selected row's element index through aggregation serialization as `offset`, or reject this combination until it can be represented.

internal/distributed/proxy/httpserver/utils.go
buqian-zilliz: This issue still exists. The new check compares raw REST `outputFields` entries exactly, but the proxy trims them and also accepts the explicit `$meta["offset"]` form before normalizing the dynamic key to `offset`. After Struct Array reconstruction rewrites `Results.OutputFields` to `$meta`, an absent dynamic key escapes the row scan and the element index is emitted under `offset`. Please validate the normalized requested-field namespace rather than the raw strings.

zhuwenxing:
Closing this PR for now. Before exposing element-level indices as `offset` in REST v2 search results, we need to resolve the name collision between the system-generated element offset and a user-defined field, including a primary key, named `offset`.

The current collision can overwrite the user's value and cause SDKs to interpret the element index as the hit ID. This is tracked in #52461.

Once the response contract clearly separates system metadata from user-defined fields across REST and SDKs, we can revisit this implementation.

Issue #52149, opened 2026-08-03
[Bug]: Clustering compaction drops commit_timestamp and silently deletes committed import rows


bigsheeper:
## Environment

- **Milvus Version**: master (reproduced at 80435d159d); 3.0 carries the same code. 2.5/2.6 do not have the `commit_timestamp` feature and are unaffected.
- **Deployment Mode**: any (cluster or standalone)
- **MQ**: any
- **SDK**: n/a (server-side data path)
- **OS**: any

### Option B: Steps

1. Deploy: any topology with clustering compaction enabled on a collection that has a clustering key.
2. Setup: create the collection, then start a 2PC import (or CDC-replicated import) into it.
3. Workload: while the import is still running (before it commits), issue a `delete` for a primary key that the import is going to write.
4. Trigger: let the import commit, then let clustering compaction pick up the resulting segment. After commit the segment is Flushed, sorted, visible, `is_importing=false` and `commit_timestamp != 0`, which satisfies every predicate in `clusteringCompactionPolicy.triggerOneCollection` (`internal/datacoord/compaction_policy_clustering.go#L147-L156`) — that policy has no `commit_timestamp` and no `IsSorted` guard.
5. Observed via: query results — the imported row is gone. Compaction itself reports success.

### Trigger Conditions

- **Frequency**: always, when a delete falls in the window between the imported rows' timestamps and the import's commit timestamp.
- **Window size**: for import, roughly the duration of the import job. For CDC/replication the source cluster's TSO can be far older than the target's, so the window can be arbitrarily long.
- **Does NOT happen when**: the segment has `commit_timestamp == 0` (ordinary write path), or when mix/sort compaction handles the segment instead — those paths read the field correctly.

## Expected Behavior

A delete whose timestamp precedes the segment's `commit_timestamp` must not apply, because the row did not exist at that time. Output rows must be normalized to `commit_ts` so that clearing `commit_timestamp` on the compaction output is sound.

## Actual Behavior

Two distinct defects, both rooted in the same partial copy at `internal/datanode/compactor/clustering_compactor.go#L499-L509`, which builds a `datapb.CompactionSegmentBinlogs` without `CommitTimestamp` and passes it to `mappingSegment()`.

[...]

PR #52150, opened 2026-08-03
fix: preserve commit_timestamp through clustering compaction

bigsheeper:
issue: #52149

## Summary

Clustering compaction handed `mappingSegment()` a hand-built partial copy of `CompactionSegmentBinlogs` that omitted `CommitTimestamp`, so every consumer inside saw 0. Two things went wrong as a result.

**Deletes that predate the commit were applied.** `NewEntityFilter` received `commit_ts = 0`, so `isEntityDeleted` compared delete timestamps against stale row timestamps rather than `max(row_ts, commit_ts)`. Rows an import had legitimately written were dropped while compaction reported success and DataCoord marked the inputs dropped.

**Row-timestamp normalization never reached storage.** It assigned `v.Timestamp`, but `storage.ValueSerializer` builds every column from `vv.Value` and never reads `vv.Timestamp` — the two are separate copies produced at deserialization. Output rows kept stale timestamps even though `completeClusterCompactionMutation` clears `commit_timestamp` on the output on the assumption that they were rewritten, so the output segment lost commit_ts protection permanently.

## Changes

- Drop the partial copy and pass the plan's segment straight to `mappingSegment()`. The copy was shallow — it shared the same `FieldBinlogs`/`Deltalogs` slices and `*datapb.Binlog` pointers, and binlog paths are already materialized in place by `DecompressCompactionBinlogsWithRootPath` at the top of `Compact()` — so it isolated nothing while making it easy to forget a field.
- Replace the ineffective per-value assignment with `wrapReaderWithTimestampOverwrite`, nested exactly as `sort_compaction.go` does it, so the timestamp column itself is rewritten and the entity filter, the stats and the writer all observe one value.

## Test Plan

- [x] New unit test asserts both halves: the pre-commit delete does not remove its row (10240 rows, not 10239), and every output binlog carries `timestamp_from == timestamp_to == commit_ts`. Both assertions fail on master. Removing only the missing field turns the first green and leaves the second red, which is why both are needed.
- [x] `internal/datanode/compactor` and `internal/compaction` package tests pass. Segments with `commit_timestamp == 0` take an unchanged path: `wrapReaderWithTimestampOverwrite` returns the reader untouched when the value is 0.
- [x] `make static-check` clean across core, pkg, client and go_client.
- [x] Audited every construction site of `CompactionSegmentBinlogs` (7 in total) and the `SegmentLoadInfo` chain; this copy was the only place losing the field.
- [ ] Not verified end to end on a live cluster. The delete and TTL failure modes were traced by hand and reproduced in unit tests only.

## Note for reviewers

`DataNodeCompactionMissingDeleteCount` will now count pre-commit deletes on import segments as missing, because they are correctly not applied. Mix compaction already behaves this way; this is expected, not a regression.

### DIFF
--- a/internal/datanode/compactor/clustering_compactor.go
+++ b/internal/datanode/compactor/clustering_compactor.go
@@ -497,16 +497,8 @@ func (t *clusteringCompactionTask) mapping(ctx context.Context,
 	mapStart := time.Now()
 	futures := make([]*conc.Future[any], 0, len(inputSegments))
 	for _, segment := range inputSegments {
-		segmentClone := &datapb.CompactionSegmentBinlogs{
-			SegmentID: segment.SegmentID,
-			// only FieldBinlogs and deltalogs needed
-			Deltalogs:      segment.Deltalogs,
-			FieldBinlogs:   segment.FieldBinlogs,
-			StorageVersion: segment.StorageVersion,
-			Manifest:       segment.GetManifest(),
-		}
 		future := t.mappingPool.Submit(func() (any, error) {
-			err := t.mappingSegment(ctx, segmentClone)
+			err := t.mappingSegment(ctx, segment)
 			return struct{}{}, err
 		})
 		futures = append(futures, future)
--- a/internal/datanode/compactor/clustering_compactor.go
+++ b/internal/datanode/compactor/clustering_compactor.go
@@ -618,6 +618,7 @@ func (t *clusteringCompactionTask) mappingSegment(
 		return err
 	}
 	rr = newMaterializedRecordReader(rr, materializer)
+	rr = wrapReaderWithTimestampOverwrite(rr, segment.GetCommitTimestamp())
 	defer rr.Close()
 
 	hasTTLField := t.ttlFieldID >= common.StartOfUserFieldID
@@ -657,12 +658,6 @@ func (t *clusteringCompactionTask) mappingSegment(
 			}
 			if entityFilter.Filtered(v.PK.GetValue(), uint64(v.Timestamp), expireTs) {
 				continue
-			}
-
-			// Normalize import segment timestamps: overwrite to commit_ts
-			// so the output segment becomes a normal segment (commit_ts = 0).
-			if commitTs := segment.GetCommitTimestamp(); commitTs != 0 {
-				v.Timestamp = int64(commitTs)
 			}
 
 			clusteringKey := row[t.clusteringKeyField.FieldID]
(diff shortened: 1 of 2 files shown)


czs007:

Verified:
- `internal/datanode/compactor/clustering_compactor.go:499-511` — passing the full `segment` instead of the hand-rolled partial clone restores `CommitTimestamp`, and `mappingSegment` only reads the segment (no field writes back), so sharing the object across goroutines is safe.
- `internal/datanode/compactor/clustering_compactor.go:283` — binlog `LogPath`s are materialized before concurrent mapping, and `DecompressBinLogWithRootPath` is a no-op once `LogPath != ""`, so the newly-visible `CollectionID`/`PartitionID` do not change the read path; `CommitTimestamp` is the only field the change actually activates.
- `internal/compaction/entity_filter.go:89-120` — the filter uses `max(row_ts, commit_ts)`, so pre-commit deletes are correctly ignored while post-commit deletes still apply, matching the existing sort-compaction behaviour and leaving TTL/upsert boundaries unchanged.
- `internal/datanode/compactor/timestamp_overwrite.go:66-107` — the wrapper's `Close` releases the previous batch and closes the inner materialized reader (retain/release balanced, no double-close), and a `commit_ts == 0` segment bypasses the wrapper entirely with no behaviour change.
- `internal/datacoord/meta.go:2425-2449` — output positions are recomputed from the rewritten binlog timestamps before the commit ts is cleared, so the precondition the rewrite relies on holds.
- `internal/datanode/compactor/clustering_compactor_test.go:543-564` — the row-count assertion covers the pre-commit-delete case and the per-binlog `TimestampFrom == TimestampTo == commit_ts` assertions cover the older "only the bypass field was changed" defect, so the two halves of the fix are independently exercised.

czs007:

Verified:
- `internal/datanode/compactor/clustering_compactor.go:499-503,568-572` — the complete plan segment (including `CommitTimestamp`) is passed through to mapping and the EntityFilter, so pre-commit deletes are evaluated against the correct effective timestamp.
- `internal/datanode/compactor/clustering_compactor.go:620-670` — the timestamp overwrite happens before deserialization on the shared path preceding scalar/vector dispatch, updating both `Value.Timestamp` and the timestamp inside `Value.Value` consumed by storage writers.
- `internal/datanode/compactor/timestamp_overwrite.go:74-107` — a zero commit timestamp bypasses the wrapper entirely, underlying reader errors propagate unchanged, and retained Arrow records are released on advance and on close.
- `internal/datacoord/meta.go:2425-2449` — `CommitTimestamp` is cleared only after rewritten output has been successfully generated, keeping output normalization safe.
- `internal/datanode/compactor/clustering_compactor.go:499-506` — each input segment is read by a single future after path decompression completes, and shared cluster writers remain protected by `ClusterBuffer` locks, so no new race is introduced.
- `internal/datanode/compactor/clustering_compactor_test.go:543-563` — the regression test constructs `row_ts < delete_ts < commit_ts` and independently verifies both that all 10,240 rows survive the pre-commit delete and that every output binlog is normalized to the commit timestamp, so it is not a false positive.

czs007 (approved):
/lgtm

PR #52151, opened 2026-08-03
enhance: support TEXT fields in go client

zhagnlu:
issue: #48783

### DIFF
--- a/client/entity/field.go
+++ b/client/entity/field.go
@@ -52,6 +52,8 @@ func (t FieldType) Name() string {
 		return "String"
 	case FieldTypeVarChar:
 		return "VarChar"
+	case FieldTypeText:
+		return "Text"
 	case FieldTypeArray:
 		return "Array"
 	case FieldTypeJSON:
@@ -95,6 +97,8 @@ func (t FieldType) String() string {
 	case FieldTypeString:
 		return "string"
 	case FieldTypeVarChar:
+		return "string"
+	case FieldTypeText:
 		return "string"
 	case FieldTypeArray:
 		return "Array"
@@ -140,6 +144,8 @@ func (t FieldType) PbFieldType() (string, string) {
 		return "String", "string"
 	case FieldTypeVarChar:
 		return "VarChar", "string"
+	case FieldTypeText:
+		return "Text", "string"
 	case FieldTypeJSON:
 		return "JSON", "JSON"
 	case FieldTypeGeometry:
@@ -187,6 +193,8 @@ const (
 	FieldTypeJSON FieldType = 23
 	// FieldTypeGeometry field type Geometry
 	FieldTypeGeometry FieldType = 24
+	// FieldTypeText field type text
+	FieldTypeText FieldType = 25 // variable-length strings without a required max_length
 	// FieldTypeTimestamptz field type timestamptz
 	FieldTypeTimestamptz FieldType = 26
 	// FieldTypeBinaryVector field type binary vector
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -967,6 +967,16 @@ func checkAndSetData(body []byte, collSchema *schemapb.CollectionSchema, partial
 					}
 				}
 
+				if fieldType == schemapb.DataType_Text {
+					if !fieldValue.Exists() {
+						continue
+					}
+					if fieldValue.Type != gjson.String {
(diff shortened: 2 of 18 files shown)


client/entity/field.go
```diff
@@ -187,6 +193,8 @@ const (
 	FieldTypeJSON FieldType = 23
 	// FieldTypeGeometry field type Geometry
 	FieldTypeGeometry FieldType = 24
+	// FieldTypeText field type text
+	FieldTypeText FieldType = 25 // variable-length strings without a required max_length
```
buqian-zilliz, line 197: **[P2] Reject TEXT sub-fields in StructSchema validation**

`FieldTypeText` is now a public field type, but `StructSchema.Validate` still accepts it. `Schema.ProtoMessage` rewrites every non-vector struct sub-field to `DataType_Array` with the original type as `ElementType`, so a TEXT sub-field becomes `Array<Text>`; the server rejects TEXT as an array element, and the struct column builder has no TEXT case either. Consequently, `CreateCollection` passes the SDK's automatic validation and fails only on the server. Please reject TEXT sub-fields locally, or add end-to-end struct-array TEXT support before accepting them.

zhagnlu: This PR only guarantees top-level TEXT support; the current server still rejects Array<Text>. Server support and Go struct-column encoding will be added together in a follow-up. PyMilvus also currently preserves and serializes this schema shape without rejecting it locally.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -775,9 +775,7 @@ func checkAndSetData(body []byte, collSchema *schemapb.CollectionSchema, partial
 					reallyData[fieldName] = result
 				case schemapb.DataType_Timestamptz:
 					reallyData[fieldName] = dataString
-				case schemapb.DataType_VarChar:
-					reallyData[fieldName] = dataString
-				case schemapb.DataType_String:
+				case schemapb.DataType_String, schemapb.DataType_VarChar, schemapb.DataType_Text:
```
buqian-zilliz: **[P2] Validate TEXT values before stringifying REST rows**

The new TEXT arm stores `fieldValue.String()` without first establishing that the field exists and is a JSON string. For a required non-nullable TEXT field, both an omitted key and an explicit `null` become `""`; `anyToColumns` then sees the synthesized map entry and bypasses its missing-field error, so the insert silently writes an empty string. Numbers and booleans are similarly stringified instead of rejected. Validate `Exists()` and `gjson.String` (after the nullable/default handling) before assigning the TEXT value.

internal/distributed/proxy/httpserver/wrap_request.go
```diff
@@ -136,7 +136,7 @@ func (f *FieldData) AsSchemapb() (*schemapb.FieldData, error) {
 				},
 			},
 		}
-	case schemapb.DataType_VarChar:
+	case schemapb.DataType_VarChar, schemapb.DataType_Text:
```
buqian-zilliz: **[P2] Do not erase nullable TEXT values in low-level REST**

Unmarshalling the low-level TEXT payload directly into `[]string` silently turns a JSON `null` element into the zero value `""`. Because this wrapper cannot carry `ValidData`, Proxy later synthesizes an all-true bitmap for nullable fields, so `field: ["a", null]` is stored as `["a", ""]` instead of preserving null (or rejecting the request). Decode nullable elements and populate `ValidData`, or explicitly reject null elements.

internal/distributed/proxy/httpserver/utils.go
zhagnlu: fixed

client/entity/field.go
```diff
@@ -96,6 +98,8 @@ func (t FieldType) String() string {
 		return "string"
 	case FieldTypeVarChar:
 		return "string"
+	case FieldTypeText:
+		return "string"
```
buqian-zilliz, line 102: **[P3] Keep TEXT distinct in column type mismatch diagnostics**

`FieldTypeText.String()` returns the same `"string"` label as `FieldTypeVarChar`, while `processInsertColumns` compares the distinct enum values and formats both with `%v`. Passing `WithVarcharColumn` for a TEXT field (or `WithTextColumn` for a VarChar field) therefore fails with `has type string but collection field definition is string`, which hides the actual mismatch. Please use `Name()` or another schema-type formatter in this diagnostic so it identifies `Text` versus `VarChar`.

zhengbuqian (approved):

Issue #51318, opened 2026-07-13
[Backfill] Harden CommitBackfillResult: V3 function-output index never builds, missing result fencing, unauthenticated management endpoint


xiaofan-luan:
## Summary

The Spark-driven backfill commit path (`DataCoord.CommitBackfillResult`), used to materialize
function-output columns (e.g. TextEmbedding) for existing rows after an online schema change, has
three correctness/hardening gaps that should be closed before online schema change is enabled by
default. These were surfaced while reviewing #50990. Two related items were already handled there
(partial-batch failure reporting, and a milvus-side `schemaVersion` fence that stays dormant until
Spark stamps it); this issue tracks the remaining work.

Scope note: the commit has two storage paths that behave differently — **V2** (`column_groups` →
`buildV2Groups`) and **V3** (`version` + `manifestPaths`). The findings are per-path where it matters.

## 1. V3 backfill never updates ChildFields → function-output index never builds

The V3 commit (`UpdateManifestVersion` in `internal/datacoord/meta.go`) only advances the segment's
`ManifestPath`; it never touches `segment.Binlogs` / `ChildFields`. But index eligibility
(`getSegmentBinlogFields` / `canCreateIndexForSegment` in `internal/datacoord/index_inspector.go`)
reads the field set from `segment.GetBinlogs()` (meta), not from the V3 manifest. So a freshly
backfilled function-output field is invisible to index creation — `canCreateIndexForSegment` takes
the `"function output field has no binlog, skip create index"` branch and returns false.

Result: for existing (sealed, pre-change) **V3** segments, the function-output vector index is
**never built** — brute-force scan forever, even though the embeddings become readable (the manifest
advance does trigger a QueryNode Reopen).

The **V2** path does not have this bug: `buildV2Groups` (`internal/datacoord/backfill_result.go`)
sets `ChildFields`, and `UpdateSegmentColumnGroupsOperator` bumps `DataVersion`, so the field is
indexable.

There is also an **unfenced race** with `BumpSchemaVersionCompaction` (default off): it produces new
segment IDs and drops the old ones, while a concurrent/later Spark backfill still targets the old
segment IDs → the backfill commit is rejected (`segment state is not Flushed`). Same root cause:
index eligibility reads meta `Binlogs` instead of the V3 manifest, and there is no coordination
between bump-compaction and backfill.

Suggested fix: for V3, either make index eligibility read the manifest's field set, or have the V3
[...]

PR #52155, opened 2026-08-04
test: add Spark-Milvus backfill E2E coverage

### LINKED ISSUE #52130: [Enhancement]: Create E2E test for sprak backfill

Create E2E test for sprak backfill

### Why is this needed?

_No response_

### Anything else?

_No response_

FeilongHou:
## What this PR does

Adds dedicated Spark-Milvus Backfill E2E coverage and the runtime tooling needed to execute it against real Storage V3 and Storage V2 segments.

- Adds opt-in, serial-only pytest coverage for Backfill modes, null semantics, snapshot boundaries, schema fencing, manifest versions, compaction, and invalid inputs.
- Adds Kubernetes Job and reusable Spark Toolbox runners with credential redaction, RBAC validation, evidence collection, and cleanup.
- Builds `zilliztech/spark-milvus` in the toolbox pod and runs Spark locally through the validated wrapper and native-library runtime.
- Separates Storage V3 and Storage V2 execution through pytest markers and independent Milvus deployments.
- Uses the dedicated `SparkBackfill` tag, so the normal `L0 L1 L2 ClusterOnly` nightly selection does not include this suite.
- Documents manual toolbox deployment and dedicated-suite execution.

Fixes #52130

## Pipeline isolation

The suite is ignored during ordinary Python client collection unless `--run-spark-backfill` is explicitly provided. The dedicated pipeline additionally selects `--tags SparkBackfill`; existing PR, nightly, and weekly tag sets therefore select zero Spark Backfill cases.

V3 and V2 remain separate marker selections because they require different Milvus storage configurations and must run serially without pytest-xdist.

## Test plan

- [x] Dedicated Spark Backfill pipeline completed successfully.
- [x] Storage V3: 30 cases collected; 26 passed and 4 strict xfails track the known schema-fencing gap in #51318.
- [x] Dedicated `SparkBackfill` tag collects 31 cases: 30 V3 and 1 V2.
- [x] Normal nightly tags `L0 L1 L2 ClusterOnly` select 0 Spark Backfill cases, even when the opt-in collection gate is opened.
- [x] `ruff check tests/python_client/common/common_type.py tests/python_client/conftest.py tests/python_client/spark_backfill`
- [x] `git diff upstream/master...HEAD --check`

## Operational notes

- Spark execution uses `spark-submit --master local[2]` in one toolbox pod.
- The toolbox is discovered by `app=spark-milvus-toolbox` and is managed outside pytest.
- V3 and V2 must use independent Milvus deployments.


yanliang567 (changes requested):
Requesting changes because the suite currently misses several production-risk paths and can report green while important backfill behavior is broken.

1. **[P1] The positive V3 coverage never exercises the real online-schema-change/function-output path.** The case factory creates `bf_score`, `bf_label`, `bf_vector`, and the `bf_vector` index before inserting source rows. The positive cases then only validate query visibility after commit. This cannot detect the known V3 `ChildFields`/index-eligibility failure in #51318. Please add a case that inserts and flushes the original rows first, adds the function-output field/function/index afterward, snapshots and backfills, then verifies index completion and deterministic ANN search results.

2. **[P1] The V2 commit verification checks only the target fields.** `_commit_and_wait()` queries only `bf_score`, `bf_label`, and `bf_vector`. Because the V2 commit path rewrites segment column groups, a regression that drops or corrupts `base_int`, `base_float`, `text`, or `vector` could still pass. Please verify the complete source rows after both the initial commit and replacement commit.

3. **[P1] Active compaction protection is not covered.** Both compaction tests wait until `compaction_expire_time` has elapsed before triggering/validating compaction, and one disables auto-compaction. A regression that ignores `compaction_protection_seconds` entirely would still pass. Please trigger compaction while the protection TTL is active and assert that the snapshot segment IDs remain persistent and the generated result remains committable.

4. **[P1] The documented existing-Secret RBAC mode is incomplete.** When `--spark-storage-secret-name` is provided without local AK/SK environment variables, `spark_storage_credentials` calls `read_namespaced_secret`. However, `required_rbac_permissions(create_secret=False)` and the README do not require `get` on `secrets`. A user following the documented least-privilege Role will fail before the suite starts. Please add the permission to the preflight check and documentation, or avoid reading the Secret from pytest.

5. **[P2] Additional snapshots can leak when setup validation fails.** `BackfillCase.create_snapshot()` registers the snapshot in `snapshot_names` only after describe/object reads/parsing/storage validation. Any failure between server-side creation and registration leaves an untracked snapshot that teardown cannot drop. Register it immediately after successful creation or delete it on subsequent failure.

Local verification on head `8a32b42a7c0c826591f46b81b92b042ebdad0af8`:
- Ruff check and format check passed.
- Default collection excludes this suite.
- Explicit opt-in collects 31 cases.
- `L0 L1 L2 ClusterOnly` selects 0 cases.
- `-n 2` fails fast as intended.

The current `ci-v2/e2e-default` status is still failing; Jenkins logs require authentication, so I could not correlate that failure further.

yanliang567 (commented):
## Test review

Thanks for this — the infra quality is high (credential redaction, RBAC preflight, safe bundle extraction with SHA256 + manifest runtime checks, strict-xfail tracking of known gaps, and clean V2/V3 marker isolation are all well done). Below are review notes focused on coverage gaps and a few fragile assertions.

### Coverage gaps

1. **The Connector "Read" path has zero test coverage.** `BackfillCase.run_read_probe()` (`case.py`), `read_probe.py`, the `operation="read"` branch in both `remote_entrypoint.py` and `toolbox_runner.py`, and `contracts.extract_read_probe_result()` are all dead code — nothing calls `run_read_probe`. The PR title and README advertise "Read/Backfill" coverage, but only Backfill is actually exercised. Either add a Read E2E case (the probe already computes `count`/`primaryKeys`/`projection`/`sqlRows`/`topK` and would be cheap to assert) or drop the read plumbing from this PR.

2. **Missing scenario: schema change + backfill + search *without* reload.** The closest existing case is `test_v3_online_minhash_function_output_builds_index_and_searches`, but it explicitly does `release_collection` + `load_collection` before searching, so it never verifies online search on a newly-backfilled field. The V2 case does assert online visibility without reload, but via `query` (`wait_for_visible_rows`), not `search`. The most failure-prone path — add a field to a sealed collection, backfill it, then search the new vector field without reloading — is currently unexercised. Recommend adding one positive case (add vector field → backfill → build index → search directly), or recording "reload required" as an explicit known limitation rather than implicitly working around it.

3. **Test-count claim doesn't reconcile with the diff.** The PR body says "30 V3 + 1 V2, 4 strict xfails". Counting parametrized instances in the diff: `test_v3_backfill_e2e.py` yields 21 instances, `test_v3_negative_e2e.py` yields 11, total 32 V3. The xfail instances are 5, not 4, because `test_schema_change_after_snapshot_before_spark_rejects_stale_result` is parametrized over `["add", "drop"]` (2 instances). Please state the exact `-m`/`--tags` selection the dedicated pipeline used, since the numbers don't obviously match the diff (the 3 compaction-marked cases may have been excluded).

4. **V2 coverage is thin** — a single positive case, no V2 negative, compaction, or schema-fence coverage. The README candidly documents the V2 Read/planner limitations, so this is understandable, but worth calling out as an explicit known gap for V2 parity.

### Assertion / robustness concerns

5. **`build_ground_truth` is a hand-authored oracle.** The tests validate the connector against a locally written model of `coalesce`/`overwrite`/`replace` semantics. In particular, `replace` sets unmatched rows' target fields to `None` ("replace clears unmatched rows"); if that differs from the connector's actual contract, the tests pass/fail against the wrong oracle. Please cross-check the `replace` semantics against the connector spec.

6. **`infer_root_path` hardcodes the `/snapshots/` path layout** (`case.py`). This couples the suite to Milvus's internal snapshot path structure, which the milvus-table design explicitly avoided exposing. A path-layout change will silently break every case's `s3_root_path`. Prefer deriving the root from the snapshot location more defensively or exposing it explicitly.

7. **Negative-case assertions couple to connector log strings** — `"types to match"`, `"snapshot float"`, `"parquet double"`/`"parquet string"`, `"dimension mismatch"`, `"expected a json array"`, `"mode must be one of"`, `"numberformatexception"`, `"batchsize must be positive"`, `"duplicate primary key"`. These are exact-substring matches against connector messages; any wording change spuriously fails the suite.

8. **`log_contains_message` normalization maps `pk` → `primary key`**, which can produce false-positive substring matches (e.g. a log containing the words "primary key" for an unrelated reason). Consider matching on a more stable signal than normalized log text where possible.

9. **`test_expired_protection_with_compaction_rejects_old_result_and_new_snapshot_succeeds` asserts `old_status in {200, 500}`.** The granular `segment_statuses` assertions carry the real signal, but accepting both HTTP statuses masks a status-code regression; tighten to the expected code.

10. **`pytest_ignore_collect` couples the entire python_client suite's collection to `spark_backfill.config` being importable.** Today the `config → k8s_runner` chain is stdlib-only so it's safe, but a future import added to either module would break all normal CI at collection time with an `ImportError`. Consider moving `is_spark_backfill_path` to a dependency-free module (or a plain path-string check) so the gate cannot accidentally drag in heavy/optional deps.

11. **`add_minhash_function_field` hardcodes the REST endpoint and payload shape** (`/v2/vectordb/collections/add_function_field`), bypassing pymilvus. Any API/payload change silently breaks the MinHash case.

12. **Compaction cases depend on wall-clock time** (`compaction_protection_seconds=1800` / `int(time.time()) < expire_time`). The 1800s window is generous so flake risk is low, but a slow cluster could still race; the `time.sleep(expire_time - now + 2)` in the expired cases is likewise a fixed heuristic.

### Positives (for the record)

- `xfail(strict=True, raises=<specific error>)` correctly forces XPASS → fail when the tracked gap (#51318) is fixed.
- Multi-layer credential redaction (`_reject_sensitive_keys`, `_redact`, `redact_text`, `_redacted_pipeline`) plus Secret-only injection is thorough.
- `safe_extract_tar` (rejects symlink/hardlink and `..`) + archive/file SHA256 + runtime Spark/Scala/Java/OS/arch validation is solid supply-chain hygiene.
- Snapshot-boundary semantics (extra PK, post-snapshot insert, post-snapshot delete) are well covered.

yanliang567 (changes requested):
Following up on the current head (`119d187d47`), I still see three test-correctness issues that should be addressed before merge:

1. **[P1] V2 still does not verify source-field integrity after Commit.**

   In `test_v2_backfill_e2e.py`, `_commit_and_wait()` builds the expected result with `TARGET_FIELDS` and calls `wait_for_visible_rows(..., TARGET_FIELDS)`, so it queries only `bf_score`, `bf_label`, and `bf_vector`. A V2 regression that drops or corrupts `base_int`, `base_float`, `text`, or `vector` can therefore still pass both the initial and replacement commits. Please merge the original source fields into the expected rows and query/assert the complete row, as the V3 helper already does.

2. **[P1] The visibility checks are preceded by a forced load refresh, so they do not validate automatic online propagation.**

   `BackfillCase.drop_snapshots_and_refresh()` always calls `client.refresh_load()`, and the common V3 validation helper invokes it before `wait_for_visible_rows()`. The V2 replacement path does the same through `drop_snapshot=True`; the MinHash path additionally performs `release_collection()` followed by `load_collection()` before search. As a result, these cases can pass even if Commit does not automatically propagate the new manifest/DataVersion to the loaded collection. This also conflicts with the helper/README claim that visibility is verified "without reload." Please separate snapshot cleanup from load refresh and assert query/search visibility before any `refresh_load`, release, or reload. At least one function-output search case should run directly against the already-loaded collection.

3. **[P2] The Read known-gap xfail can hide any Spark Read failure, not only the documented `minio://` incompatibility.**

   The test uses `xfail(..., raises=BackfillContractError)`, while `BackfillCase.run_read_probe()` converts every failed Read job into `BackfillContractError`. Authentication failures, bad S3 credentials, wrapper/build regressions, class-loading failures, or an unrelated Connector crash would all be reported as the expected xfail. Please detect the specific unsupported-scheme failure and raise a dedicated exception for that condition; all other Read failures should remain real failures.

FeilongHou:
> Following up on the current head (`119d187d47`), I still see three test-correctness issues that should be addressed before merge:
> 
> 1. **[P1] V2 still does not verify source-field integrity after Commit.**
>    In `test_v2_backfill_e2e.py`, `_commit_and_wait()` builds the expected result with `TARGET_FIELDS` and calls `wait_for_visible_rows(..., TARGET_FIELDS)`, so it queries only `bf_score`, `bf_label`, and `bf_vector`. A V2 regression that drops or corrupts `base_int`, `base_float`, `text`, or `vector` can therefore still pass both the initial and replacement commits. Please merge the original source fields into the expected rows and query/assert the complete row, as the V3 helper already does.
> 2. **[P1] The visibility checks are preceded by a forced load refresh, so they do not validate automatic online propagation.**
>    `BackfillCase.drop_snapshots_and_refresh()` always calls `client.refresh_load()`, and the common V3 validation helper invokes it before `wait_for_visible_rows()`. The V2 replacement path does the same through `drop_snapshot=True`; the MinHash path additionally performs `release_collection()` followed by `load_collection()` before search. As a result, these cases can pass even if Commit does not automatically propagate the new manifest/DataVersion to the loaded collection. This also conflicts with the helper/README claim that visibility is verified "without reload." Please separate snapshot cleanup from load refresh and assert query/search visibility before any `refresh_load`, release, or reload. At least one function-output search case should run directly against the already-loaded collection.
> 3. **[P2] The Read known-gap xfail can hide any Spark Read failure, not only the documented `minio://` incompatibility.**
>    The test uses `xfail(..., raises=BackfillContractError)`, while `BackfillCase.run_read_probe()` converts every failed Read job into `BackfillContractError`. Authentication failures, bad S3 credentials, wrapper/build regressions, class-loading failures, or an unrelated Connector crash would all be reported as the expected xfail. Please detect the specific unsupported-scheme failure and raise a dedicated exception for that condition; all other Read failures should remain real failures.

2. **[P1] The visibility checks are preceded by a forced load refresh, so they do not validate automatic online propagation.**
It is by design. Users are expected to manually refresh the collection after dropping the snapshot.


3. **[P2] The Read known-gap xfail can hide any Spark Read failure, not only the documented `minio://` incompatibility.**
currently it is still failing. Because Spark reads all segments info including deleted and duplicate. But the logical count(*) and data are correct from Milvus side.

PR #21909, opened 2023-02-01
Avoid counting all bits to improve query performance

yah01:
Signed-off-by: yah01 <yang.cen@zilliz.com>
/kind improvement
more improvement detail: #20462

PR #52157, opened 2026-08-04
fix: route Tantivy logs through Milvus logging

### LINKED ISSUE #51912: [Enhancement]: Route background Tantivy logs through the Milvus logging system

Route logs emitted by Tantivy background work through the standard Milvus logging system instead of writing them directly to `stderr`.

The Tantivy Rust binding currently initializes `env_logger` in `internal/core/thirdparty/tantivy/tantivy-binding/src/log.rs`. `env_logger` writes to `stderr` by default. Milvus calls `tantivy_set_log_level` from `internal/core/src/common/Common.cpp`, but this only updates the Rust log filter and does not redirect log records into the Milvus logging pipeline.

Add a thread-safe logging bridge for the Rust binding, for example a Rust `log::Log` implementation that forwards records over FFI to the existing Milvus C++/Go logging sink. The bridge should:

- Preserve the Rust log severity mapping (`trace`, `debug`, `info`, `warn`, and `error`).
- Include useful Tantivy target/module and source information when available.
- Respect the configured Milvus log level, output, format, and rotation behavior.
- Work safely for Tantivy background threads without recursion, deadlock, or lifetime issues across FFI.
- Avoid direct `stderr` output for normal Tantivy runtime logs.

Acceptance criteria:

- Normal Tantivy background logs no longer write directly to `stderr`.
- Tantivy records appear in the configured Milvus log output with the correct severity and recognizable Tantivy source context.
- Changing the Milvus log level controls Tantivy log visibility consistently.
- Tests cover severity mapping and logging from a background thread.

### Why is this needed?

Direct `stderr` output bypasses Milvus logging configuration and observability conventions. These records can have a different format, miss component/source context, ignore the configured destination and rotation policy, and be handled inconsistently by log collection in different deployments. Routing them through the standard Milvus logging path makes Tantivy diagnostics consistent, configurable, and easier to correlate with the rest of the process logs.

### Anything else?

Relevant code paths:

- `internal/core/thirdparty/tantivy/tantivy-binding/src/log.rs`: initializes `env_logger`.
- `internal/core/thirdparty/tantivy/tantivy-binding/src/log_c.rs`: currently exposes only log-level control.
- `internal/core/src/common/Common.cpp`: propagates the Milvus log level to Tantivy.
- `internal/core/src/common/logging_c.cpp` and `internal/util/cgo/logging/logging.go`: existing C++ to Milvus logging bridge that can serve as the integration pattern.

fancyboi999:
## What this PR does

- replaces the Tantivy binding's standalone `env_logger` with a process-wide logger that forwards structured records through the Milvus logging stack
- preserves Tantivy severity, target, source file, line, and message across the Rust/C++/Go boundary
- keeps runtime log-level changes effective, including when Milvus configures the level before the first Tantivy operation
- falls back to `env_logger` when a Go logging sink is unavailable, so C++-only binaries do not silently drop Tantivy logs

The callback is registered once and the logging hot path does not take a shared lock. Strings received by Go are copied before they are passed to Zap because a Core may retain an entry after the FFI callback returns.

## Why

The binding currently installs its own `env_logger`, so Tantivy records bypass Milvus log formatting, level control, and output routing. Its fixed `Info` filter also suppresses later `debug` and `trace` settings even though `tantivy_set_log_level` updates the facade's max level.

## Tests

- `cargo +1.89 test --test log_bridge -- --nocapture`
- `cargo +1.89 fmt --all -- --check`
- `cargo +1.89 check --lib`
- `go test -tags dynamic,test -gcflags='all=-N -l' ./internal/util/cgo/logging -run 'Test(TantivyLoggingBridge|Logging|MapTantivySeverity|LogTantivyRecord|LogTantivyRecordRespectsCoreLevel)$' -count=1 -v`
- the same focused Go tests with `-race`
- `go vet -tags dynamic,test ./internal/util/cgo/logging`
- C++ syntax checks for normal and `WITHOUT_GO_LOGGING` builds on Linux; production and fallback bridge harnesses on macOS

`cargo +1.89 test --lib` is currently blocked on `master` by the existing `create_text_writer` test call in `src/index_reader_text.rs:266`, which is missing the new `enable_background_merge` argument.

Fixes #51912

### DIFF
--- a/internal/core/src/common/logging_c.cpp
+++ b/internal/core/src/common/logging_c.cpp
@@ -169,7 +169,7 @@ InitGoogleLoggingWithoutZapSink() {
         return;
     }
     google::InitGoogleLogging("milvus");
-    // log is caught by zap, so we don't need to log to stderr/stdout/files anymore.
+    // No Zap sink is installed, so route glog output to stdout.
     FLAGS_logtostdout = true;
     FLAGS_logtostderr = false;
     FLAGS_alsologtostderr = false;
--- a/internal/core/src/common/logging_c.cpp
+++ b/internal/core/src/common/logging_c.cpp
@@ -18,9 +18,11 @@
 
 #include <glog/logging.h>
 #include <string.h>
+#include <cstdint>
 #include <string>
 
 #include "glog/log_severity.h"
+#include "tantivy-binding.h"
 
 #ifdef WITHOUT_GO_LOGGING
 
@@ -32,6 +34,19 @@ goZapLogExt(int severity,
             int line,
             const char* msg,
             int msg_len) {
+}
+
+// Go logging is optional in C++ unit-test binaries. This weak fallback tells
+// Rust to keep using env_logger when no Go implementation is linked.
+extern "C" bool __attribute__((weak)) goZapLogTantivy(int severity,
+                                                      const char* target,
+                                                      uintptr_t target_len,
+                                                      const char* file,
+                                                      uintptr_t file_len,
+                                                      uint32_t line,
+                                                      const char* msg,
+                                                      uintptr_t msg_len) {
+    return false;
 }
 
 #elif defined(__APPLE__)
@@ -47,6 +62,17 @@ extern "C" void __attribute__((weak)) goZapLogExt(int severity,
                                                   int msg_len) {
 }
 
+extern "C" bool __attribute__((weak)) goZapLogTantivy(int severity,
+                                                      const char* target,
+                                                      uintptr_t target_len,
+                                                      const char* file,
+                                                      uintptr_t file_len,
+                                                      uint32_t line,
+                                                      const char* msg,
+                                                      uintptr_t msg_len) {
+    return false;
+}
+
 #else
 
 // Go export function.
@@ -59,7 +85,36 @@ goZapLogExt(int severity,
             const char* msg,
             int msg_len);
 
+extern "C" bool
+goZapLogTantivy(int severity,
+                const char* target,
+                uintptr_t target_len,
+                const char* file,
+                uintptr_t file_len,
+                uint32_t line,
+                const char* msg,
+                uintptr_t msg_len);
+
 #endif
+
+static bool
+TantivyLog(TantivyLogLevel severity,
+           const char* target,
+           uintptr_t target_len,
+           const char* file,
+           uintptr_t file_len,
+           uint32_t line,
+           const char* msg,
+           uintptr_t msg_len) {
+    return goZapLogTantivy(static_cast<int>(severity),
+                           target,
+                           target_len,
+                           file,
+                           file_len,
+                           line,
+                           msg,
+                           msg_len);
+}
 
 class GoZapSink : public google::LogSink {
     void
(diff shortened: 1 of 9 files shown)


fancyboi999:
Hi @czs007 and @xige-16, could you please take a look when convenient?

This PR routes Tantivy records through Milvus `mlog` while preserving severity, target, and caller metadata. It also keeps an `env_logger` fallback for binaries without the Go logging sink.

The focused integration coverage exercises the real `tantivy_index_exist` FFI path through Rust -> C++ -> Go -> `mlog`, plus background-thread logging, runtime level changes, and the Go race detector. After the review, I will follow the approval notifier and `/assign @liliu-z` for final approval.

fancyboi999:
CI update: the original Code Check failure is fixed in `c15f77b0c`. [Execution #21909](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-build-ut-ciloop-pipeline/21909/pipeline-overview/) passed Code Check (0 issues for core/pkg/client), Build, UT-Integration, and UT-CPP-Cov.

The remaining UT-Go failure is an unrelated current-master test regression in `internal/datanode/index.TestAnalyzeTaskSuite/TestMaxConnectionsReachesAnalyze`. The [archived log](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-build-ut-ciloop-pipeline/21909/artifact/ut-go-test.log) shows the mock at `task_test.go:286` still has the old two-argument `Analyze` signature. #51789 added `*indexcgopb.StoragePluginContext` as a third argument before #51708 merged the stale mock, so the test panics during mock setup before any logging code executes.

I requested the rerun before the detailed artifact was available. Since this is a deterministic signature mismatch, execution #21930 is expected to reproduce it. I am keeping this PR scoped to the Tantivy logging change.

internal/core/src/common/logging_c.cpp
```diff
@@ -113,7 +169,7 @@ InitGoogleLoggingWithoutZapSink() {
         return;
     }
     google::InitGoogleLogging("milvus");
-    // log is catched by zap, so we don't need to log to stderr/stdout/files anymore.
+    // log is caught by zap, so we don't need to log to stderr/stdout/files anymore.
```
yhmo: internal/core/src/common/logging_c.cpp line:172
Low ---- This comment is still inverted for InitGoogleLoggingWithoutZapSink: this branch never installs g_sink and explicitly enables FLAGS_logtostdout, so logs are not caught by Zap here. Please describe the stdout/no-sink behavior instead; otherwise future changes to these flags can be based on the wrong output-routing assumption.

internal/util/cgo/logging/tantivy_logging_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestTantivyLoggingBridge(t *testing.T) {
+	readEntries := captureMlogEntries(t, "info")
+
+	InitGoogleLoggingWithZapSink()
+	require.False(t, tantivyIndexExist("/nonexistent/path/to/tantivy-index"))
```
yhmo, line 33: internal/util/cgo/logging/tantivy_logging_test.go line:33
Low ---- This exercises the real Rust->C++->Go->mlog bridge only synchronously. The background-thread test in tests/log_bridge.rs replaces the callback with the Rust capture_log, so no committed test covers a Rust-created worker entering the actual Go callback/mlog path (the concurrency boundary this change is meant to make safe). Could we add a deterministic background emitter through the production callback and run that under -race?

internal/core/src/common/logging_c.cpp
fancyboi999: Fixed in 8a33f011d. The comment now states that no Zap sink is installed and glog output is routed to stdout, matching the flags in this branch.

internal/util/cgo/logging/tantivy_logging_test.go
fancyboi999: Fixed in 8a33f011d. I added a deterministic Rust C-ABI emitter that creates and joins a Rust worker, and both the Rust integration test and the Go bridge test now invoke it. The Go test exercises the production Rust -> C++ -> Go -> mlog callback path and asserts the emitted messages, severities, target, and caller. The full logging package passes locally with `go test -race -tags dynamic,test ./internal/util/cgo/logging -count=1 -v`.

PR #52161, opened 2026-08-04
enhance: [2.6] skip re-applying baked deletes on segment load (#49435)

tinswzy:
### What & why

Addresses the residual `channel tsafe stalled` on #49435. After the previous delegator lock fix (#51754), the failure still reproduces occasionally, and we traced it to **CPU throttling**: a compaction burst triggers many concurrent segment loads whose delete-replay saturates an all-in-one standalone's cores and starves the single-threaded delete flowgraph that advances `tsafe`.

The root of that CPU-heavy replay is a **data-amplification**: each newly-loaded compacted segment re-applies the whole accumulated delete history, because rows that are written once and never updated keep the replay window pinned to the start of the run, so the in-memory delete buffer never trims (measured `matched ≈ 55M rows per apply`).

### How

Record, per segment, a **`delete_covered_position`** — the WAL position up to which deletes are already baked into the segment's data by compaction. On load, the delegator replays only the delete tail after that position (and pins the delete buffer there) instead of from `start_position` (minTs).

- **proto**: add optional `SegmentInfo.delete_covered_position`, mirrored into `SegmentLoadInfo`.
- **datacoord**: compaction sets it on its output segment as the **MIN** delete coverage across inputs (each input's coverage = max of its own covered position and its deltalog `TimestampTo`); `nil` when any input's coverage is unknown.
- **querycoord**: propagate it into `SegmentLoadInfo`.
- **delegator**: replay/pin from `delete_covered_position` when present, else fall back to `start_position`.

The field is **optional** and the position is a **conservative lower bound** on the true coverage, so replaying from it can only re-apply already-applied deletes (idempotent) and never loses one; segments without it keep the previous behavior.

### Notes

- Targets `2.6` first (the delegator-only lock fix #51754 already landed here); will forward-port to master/3.0 after validation.
- Unit tests added: `computeDeleteCoveredPosition` / `inputDeleteCoverageTs` (datacoord) and `segmentDeleteReplayStartTs` (delegator).

issue: #49435


czs007:

## Overview

The `DeleteCoveredPosition` mechanism is sound in its safe-degradation behavior, but its coverage is narrower than the PR description implies, and two of the three compaction paths derive coverage from live metadata rather than from the snapshot the datanode actually baked.

## Medium

**`internal/datacoord/meta.go:1899` — Clustering compaction output never sets `DeleteCoveredPosition`, so the #49435 optimization is lost for the entire clustering lineage.**
`completeClusterCompactionMutation` builds its output `datapb.SegmentInfo` (lines 1898–1919) without a `DeleteCoveredPosition` and without `Deltalogs`; this PR only touched the mix (`:2034`) and sort (`:2529`) paths. Input deletes *are* baked on the datanode side (`clustering_compactor.go:573`, `:661`), but the output segment records no coverage. Because clustering output is `L2` + `IsInvisible` (`:1912`, `:1916`), it is always re-processed by sort compaction, and sort at `:2529` reads exactly this segment — coverage `nil` and no deltalogs → `inputDeleteCoverageTs` returns `known=false` → `computeDeleteCoveredPosition` returns `nil` (`:1816-1830`). The `nil` therefore propagates all the way to the final visible segment, not just an intermediate one. Behavior remains a safe degradation (the delegator falls back to `start_position`, `delegator_data.go:820-824` — idempotent, no deletes lost).
*Suggestion:* scope the "fixes #49435" claim in the PR body so it does not cover clustering-enabled clusters, and add coverage propagation for the clustering path as follow-up work. *(surfaced during verification)*

**`internal/datacoord/meta.go:1792` — "any unknown input → nil" makes coverage permanently `nil` for the most common mix-compaction shape, so the optimization rarely engages.**
`inputDeleteCoverageTs` only returns `known=true` when a segment carries `delete_covered_position > 0` or carries deltalogs (`:1792-1805`); a segment that was never hit by a delete has neither, and `computeDeleteCoveredPosition` then discards coverage for the whole output (`:1819-1822`). *Failure: an ordinary mix compaction whose inputs include at least one never-deleted segment — typically a freshly flushed segment that has been through sort compaction, whose `Deltalogs` is always nil (`sort_compaction.go:325-337` → `meta.go:2545`) and whose coverage is inherited as nil (`meta.go:2529`) → the output segment's `DeleteCoveredPosition` is empty, the delegator falls back to `start_position` (`delegator_data.go:820-824`), and replay behaves exactly as before this PR, so #49435's replay amplification, CPU saturation and stalled tsafe remain while metrics show nothing.* In 2.6, deletes for freshly flushed L1 segments all go through L0 segments (`flushcommon/syncmgr/meta_writer.go:111` leaves `deltaFieldBinlogs` empty for L1 insert packs), and mix compaction's typical shape is "a few new small segments + some older ones" — one such input is enough for the `min` semantics to zero out the whole result. Note this is not a coding error: for a segment with no deltalog, coverage genuinely cannot be proven from current metadata (deletes may still sit in an uncompacted L0 segment), so returning `nil` is the necessary conservative choice.
*Suggestion:* (a) make the "fixes #49435" wording in the PR body conditional; (b) add a coverage hit-rate metric or log line — otherwise, after rollout there is no way to distinguish "the optimization engaged but wasn't enough" from "it never engaged at all". Closing the gap properly needs an independent per-segment "L0-applied watermark" maintained by L0 compaction (advanced even when no delete hit the segment), rather than inferring coverage from the presence of deltalogs. *(surfaced during verification)*

## Low

**`internal/datacoord/meta.go:2034` — Mix compaction derives `DeleteCoveredPosition` from live meta rather than from the snapshot that was actually baked, adding an undocumented dependency on the inspector's per-channel mutual exclusion.**
Mix snapshots deltalogs at plan-build time (`compaction_task_mix.go:394`) and the datanode bakes only that snapshot, while `completeMixCompactionMutation` recomputes coverage from `compactFromSegInfos`, which is a fresh read of live meta (`meta.go:2004`). The race this would imply is *not* reachable: `compaction_inspector.go:265-267` excludes Mix/Sort on a channel that already has an executing L0 task (and `:257-259` the reverse); tasks enter `executingTasks` at `:286` *before* being enqueued at `:287`, so both `BuildCompactionRequest` and `completeMixCompactionMutation` run inside that exclusion window, as does L0's `saveSegmentMeta` (`compaction_task_l0.go:168`); removal happens only at `:626` once `Process()` returns true, which for L0 requires `meta_saved`/`completed`/`failed`/`timeout` (`compaction_task_l0.go:227-238`); and restart leaves no scheduling window (`server.go:609` `loadMeta()` precedes `:631` `compactionInspector.start()`). Even if it were reachable, deletes still resident in L0 segments are forwarded in full by the delegator's Phase 0 `forwardL0Deletion` (`delegator_data.go:848`), unfiltered by `coveredTs`. What remains is that correctness now silently depends on "live meta's deltalog set == the snapshot the datanode baked", an equality guaranteed only by the scheduler and recorded nowhere in the code.
*Suggestion:* either document the dependency with a comment/assertion at `computeDeleteCoveredPosition` / `completeMixCompactionMutation`, or derive coverage from the plan snapshot instead of live meta so the computation is self-contained. *(surfaced during verification)*

**`internal/datacoord/meta.go:2529` — The sort branch has the same live-meta derivation, and its output genuinely carries no deltalogs.**
Factual correction worth recording: the sort path's datanode output does **not** carry `Deltalogs` — the `datapb.CompactionSegment` built at `sort_compaction.go:325-337` has no such field (only `InsertLogs`/`Field2StatslogPaths`/`Bm25Logs`/`Manifest` etc.), so `meta.go:2545` `Deltalogs: resultSegment.GetDeltalogs()` is always nil; deletes are baked directly into the data via `entityFilter` (`sort_compaction.go:196`, `:206`). The corresponding race is unreachable for the same reason as above: `compaction_inspector.go:265` groups `SortCompaction` with `MixCompaction` in the same exclusion class, and import-triggered sort tasks go through the same queue (`import_checker.go:381`) — there is no second creation path bypassing the scheduler. The residual issue is identical: `meta.go:2529` derives coverage from a live read of `oldSegment` (`meta.go:2494`) rather than the baked plan snapshot.
*Suggestion:* handle this together with `meta.go:2034` — whichever fix or comment/assertion is applied there must cover this branch too, not mix alone. *(surfaced during verification)*

czs007:

Adversarial review of the delete-coverage change surfaced one correctness risk in how `DeleteCoveredPosition` is computed at compaction completion, one proto field-number conflict with master, and one test-coverage gap.

## High

**`internal/datacoord/meta.go:1793` — `computeDeleteCoveredPosition` reads live meta at completion, not the plan snapshot the compactor actually baked**

The doc comment on `computeDeleteCoveredPosition` rests correctness on an invariant that does not exist in the code: that the compaction inspector keeps L0 compaction mutually exclusive with Mix/Sort/Clustering compaction on the same channel. `selectFlushedSegment()` (`compaction_task_l0.go:306-313`) picks L0 compaction *targets* filtering only on state / `!IsImporting` / `Level != L0` / `StartPosition < pos` — there is no `!isCompacting` check. `isCompacting` is only consulted for *input* selection in the policies (`compaction_policy_l0.go:94`, `compaction_policy_single.go:231`, `compaction_util.go:93`), which does not constrain L0 target selection at all, and `L0CompactionCompactionTaskFilter()` (`compaction_inspector.go:734`) has no production caller — only tests reference it.

Failure: a Sort/Mix compaction is executing on datanode with segment A's deltalogs snapshotted into the plan at `TimestampTo=500` (`compaction_task_mix.go:400`), a concurrent L0 compaction completes and merges a deltalog carrying deletes in `(500, 900]` into A via `AddBinlogsOperator` (`compaction_task_l0.go:448` → `meta.go:1075`) while dropping the L0 segment that held them → completion reads live meta (`meta.go:1989`, `meta.go:2043`; sort path `meta.go:2538`) and stamps output segment B with `DeleteCoveredPosition=900`, even though B baked only deletes ≤500 and carries no deltalogs of its own (`sort_compaction.go:318-330`); A is then dropped, and on load `segmentDeleteReplayStartTs` (`delegator_data.go:820-825`) returns 900 so `ListAfter(900)` skips `(500, 900]` and those deleted rows are permanently resurrected in search/query — silently, with no error and no self-healing. Phase 0 `forwardL0Deletion` (`delegator_data.go:845`) cannot backstop this, because those deletes were precisely the ones moved out of the L0 segment and into A. Sort compaction has the widest exposure, since every flushed segment is sorted.

The underlying race (L0 writing deltalogs onto a segment about to be dropped) predates this PR, but the previous "replay everything from `start_position`" behaviour re-applied those deletes on every load; this PR removes that backstop while grounding coverage in an invariant that does not hold, turning a masked race into a visible data error.

Suggestion: compute `inputDeleteCoverageTs` per input at `BuildCompactionRequest` time and carry it in the `CompactionTask` proto (or derive it from `plan.SegmentBinlogs[i].GetDeltalogs()`), then read it back at completion, so coverage is by construction consistent with the deltalog set the compactor actually applied. If changing the proto is undesirable, a smaller fix is to compare each input's current deltalog set against the plan snapshot at completion and return `nil` (safe degradation to `start_position`) on any mismatch — note that `ValidateSegmentStateBeforeCompleteCompactionMutation` (`meta.go:2100-2121`) only checks input health today and cannot detect this drift. *(raised by czs007)*

## Medium

**`pkg/proto/data_coord.proto:435` — new field numbers are already taken on master, conflicting with the stated forward-port plan**

This PR assigns `msg.MsgPosition delete_covered_position = 33` in `SegmentInfo` and `= 25` in `SegmentLoadInfo` (`pkg/proto/query_coord.proto:391`). Both numbers are already occupied on master with incompatible types: `SegmentInfo` 33 is `repeated int64 expirQuantiles` (34–38 are also taken), and `SegmentLoadInfo` 25 is `int32 data_version`. The two branches agree field-for-field up through 32 (`manifest_path = 32`), so this is the first divergence — and `SegmentInfo` is persisted to etcd, while the PR body states the change "will forward-port to master/3.0 after validation."

Failure: a 2.6 cluster running this PR persists `SegmentInfo` with field 33 into etcd and is then upgraded in place to a master/3.0 binary (or a mixed-version deployment reads the same meta after a forward-port that had to pick a different number) → the `MsgPosition` bytes are parsed as packed varints into `expirQuantiles`, feeding fabricated percentile values to `ShouldCompactExpiryWithTTLField` (`compaction_trigger.go:723`) so TTL-expiry compaction fires spuriously or is wrongly suppressed, and the segment's delete coverage is silently lost on the new version.

Suggestion: agree the field numbers with the master side before merging — e.g. use the same numbers on both branches (39 / 30, or whatever is free on master) — or at minimum add a `reserved` entry on master documenting 33 / 25 so the numbers cannot be reused. *(raised by czs007)*

## Low

**`internal/datacoord/meta_delete_coverage_test.go:117` — the premise this PR's correctness depends on is untested**

The file has four cases: `TestInputDeleteCoverageTs` (:49) and `TestComputeDeleteCoveredPosition` (:79) are pure-function tests, and `TestClusteringDeleteCoveragePropagation` (:117) / `TestClusteringDeleteCoverageUnknownInput` (:183) exercise only the ClusteringCompaction and SortCompaction branches of `CompleteCompactionMutation`. The new `completeMixCompactionMutation` branch (`meta.go:2043`) has no meta-level test at all. On the delegator side, `delegator_delete_coverage_test.go:28` only covers `segmentDeleteReplayStartTs` as a pure function — nothing verifies that `loadStreamDelete` actually replays from the covered start point or that `deleteBuffer.Pin` moves with it. Most importantly, no test covers the case where the plan's snapshotted deltalogs differ from live meta at completion, which is both the entry point for the high-severity issue above and the assumption the entire correctness argument rests on.

Suggestion: add a meta-level test for the mix-compaction branch, a delegator test asserting the replay/pin start point, and a test that mutates an input segment's deltalogs between plan construction and `CompleteCompactionMutation` to pin down the intended behaviour under drift. *(raised by czs007)*

tinswzy:
Thanks @czs007 — the second round caught a real must-fix. Pushed `10f4994a35`.

**HIGH (live-meta coverage → resurrection).** Correct, and my round-1 "scheduler mutual exclusion" comment was wrong: `selectFlushedSegment` has no `!isCompacting` check and `L0CompactionCompactionTaskFilter` has no production caller, so L0 can attach deltalogs to an input between plan-build and completion. Coverage is now computed at **plan-build** from the snapshotted inputs (exactly what the datanode bakes), carried on `CompactionTask.delete_covered_ts`, and read back at completion — completion no longer reads live meta. Added a completion test asserting the output takes the task's ts, not a drifted live deltalog.

**MEDIUM (field-number clash with master).** Confirmed (`SegmentInfo` 33 = `expirQuantiles`, `SegmentLoadInfo` 25 = `data_version` on master). Renumbered to `SegmentInfo` 39 / `SegmentLoadInfo` 30 (+ the new `CompactionTask` carrier 30); the master forward-port will use the same numbers.

**LOW (tests).** Added pure-function tests plus a completion drift test. A full delegator `loadStreamDelete` replay/pin integration test remains a follow-up.

Also: clustering coverage was added in `6e7b0396f0`; the "any-unknown-input → nil" behavior stays as the necessary conservative choice (in 2.6 many inputs carry deletes only in L0, so coverage is often absent → safe fallback to `start_position`) — the proper long-term fix is an L0-maintained per-segment applied watermark rather than inferring from deltalog presence, which I'll open as a follow-up. And I switched the field from a `MsgPosition` to a plain `uint64` `delete_covered_ts`, since it is only ever used to filter the delete buffer by time (never to seek).

czs007:

Adversarial review confirmed three issues introduced by this PR — two high-severity problems around `DeleteCoveredTs` consistency and un-reverted experiment defaults, and one medium-severity scheduling gap that weakens the coverage-timestamp invariant.

## High

- **`internal/datacoord/compaction_task_clustering.go:155` — Clustering redispatch can pair an old plan's result with a newer `DeleteCoveredTs`, silently resurrecting deleted data.**
  When a status query fails, the task falls back to `pipelining` (lines 155–165); redispatch rebuilds the plan from live metadata and unconditionally overwrites `DeleteCoveredTs` (lines 371–398). Because the plan ID is unchanged, the DataNode rejects the resubmission as a duplicate (`internal/datanode/compactor/executor.go:109-115`), but the clustering `doCompact` error branch is missing a `return` (lines 767–792; the mix counterpart has one), so the task is still marked `executing`. Failure scenario: a datacoord restart re-enqueues the task via `NeedReAssignNodeID` (defeating channel-level mutual exclusion) while an L0 compaction appends newer deltalogs to the input segments → the rebuilt request raises `DeleteCoveredTs` from 100 to 200, the duplicate rejection is swallowed, and completion stamps the old plan's result (baked only through 100) with coverage 200 (`meta.go:1950`), so the delegator skips delete replay in (100, 200] on load and deleted rows become visible again. The window requires restart + reassignment to the original node + duplicate rejection, so it is narrow, but the consequence is silent data resurrection. Suggested fix: add the missing `return` in the `doCompact` error branch (line 791), and make `setDeleteCoveredTs` write-once (only write when `GetDeleteCoveredTs() == 0`; deltalogs only grow, so keeping the first value can only underestimate coverage, which is safe and idempotent). Please also add a redispatch/duplicate regression test — the existing `meta_delete_coverage_test.go:95-131` only covers direct completion. Note: the missing `return` itself is pre-existing (from 142bd2fc05), but the unconditional overwrite that makes it harmful is added by this PR. (raised by chyezh, xaxys, czs007)

- **`pkg/util/paramtable/component_param.go:5283` — Temporary experiment defaults were not reverted (`blockingL0EntryNum` 5,000,000 → 500,000; `blockingL0SizeInMB` 64 → 8).**
  The comments at lines 5276–5296 explicitly say "#49435 EXPERIMENT (temporary — revert before merge)", yet the lowered defaults remain in place. Failure scenario: a deployment built from this branch without explicit overrides for these two settings reaches 500k entries or 8 MB of L0 backlog per shard → `sealByBlockingL0` (`internal/datacoord/segment_allocation_policy.go:225-232`) seals the oldest growing segment roughly 8–10× earlier than intended, multiplying small sealed segments and the resulting flush/sort-compaction/indexing/load bursts — the same class of CPU-starvation regression this PR aims to eliminate. The stock `configs/milvus.yaml` (lines 670/673) still carries the upstream values, so minimal or custom-config deployments are the ones exposed. Suggested fix: revert both defaults to 5,000,000 and 64 before merge. (raised by chyezh, czs007, xaxys)

## Medium

- **`internal/datacoord/compaction_inspector.go:257` — L0 tasks on the same channel are not mutually exclusive, so `delete_covered_ts` can exceed the contiguous success frontier.**
  The scheduler's `Level0DeleteCompaction` branch (lines 257–263) checks only `mixChannelExcludes`/`clusterChannelExcludes`, not the `l0ChannelExcludes` set seeded by executing L0 tasks (lines 222–223), so two L0 tasks can run concurrently on one channel. Failure scenario: an older L0 task covering (t1, t2] is still executing when a newer task covering (t2, t3] completes and attaches its deltalog, then the older task fails; a subsequent mix/clustering compaction over those segments stamps `delete_covered_ts = t3` even though (t1, t2] was never baked (`meta.go:1794-1809` infers coverage from the maximum deltalog `TimestampTo`) → the delegator skips buffer replay up to t3 on load. In the direct case this is masked by Phase 0 L0 forwarding (`delegator_data.go:842-849` → `delta_forward.go:56-71` is not filtered by covered_ts, and the failed task's L0 segments are not dropped), so turning it into an actual delete loss additionally requires a restart/stale-target race — hence medium severity as a non-blocking hardening item. Suggested fix: have `schedule()` check `l0ChannelExcludes` for L0 tasks as well, restoring per-channel L0 serialization so `delete_covered_ts` becomes a strict frontier, and add a test for the later-L0-success-then-earlier-L0-failure sequence. (raised by chyezh, czs007, xaxys)

tinswzy:
Closing this.

The core mechanism here — a per-segment delete-coverage watermark (`delete_covered_ts`), used both to skip re-applying baked deletes and to start the on-load delete-buffer replay from `covered_ts` instead of `start_position` — turns out to be effectively redundant with the existing deleteCP-based buffer bound, and same-build A/B testing shows no measurable benefit for the #49435 tsafe stall.

**Why it's redundant**
- The delegator delete buffer is already trimmed to `[deleteCP, now]`, so any on-load replay is floored at deleteCP regardless of `covered_ts`.
- Both `covered_ts` and deleteCP are bounded by the L0-compaction frontier; deleteCP is the channel-wide baked boundary tracking that frontier, so `covered_ts <= deleteCP` in the normal (uniform) case. Replaying from `covered_ts` then lands on the same deleteCP floor, and there are no baked deletes left in `[deleteCP, now]` to skip — so neither the "skip re-apply" nor the "covered start" path trims anything.
- The oldest-first L0 pick added here makes baking advance uniformly across segments, which removes the only case (skewed baking, where one lagging L0 pins deleteCP while other segments are baked ahead) in which `covered_ts` could exceed deleteCP. The safety fix also eliminates the mechanism's only upside.

**Empirical (same-build A/B on #49435)**
Coverage engaged on ~52–56% of compactions but produced no measurable improvement. The arm with coverage on but the blocking-L0 seal left loose kept an ~8.9M-row delete buffer and still stalled — identical to baseline — because buffer size (and thus the stall) is set by deleteCP/frontier, which `covered_ts` structurally cannot move. This matches @czs007's earlier review concern that `delete_covered_position` rarely engages meaningfully in the common mix path.

**What actually addresses #49435**
The stall is the frontier (→ deleteCP → buffer) not advancing under heavy upsert on a small collection, so each segment load replays a large buffer on the single-threaded delete path and freezes tsafe. The effective lever is advancing the frontier by tuning the blocking-L0 seal threshold so growing segments seal promptly under delete pressure (validated: delete buffer 8.9M→1.5M rows, stall→0, at ~7× more compaction). The root-cause fix is decoupling the buffer replay from tsafe advancement via a per-segment delete-applied watermark, so a large replay can't freeze tsafe. Both will be follow-up PRs.

Extracting the oldest-first L0 sort fix separately if it proves worth keeping on its own.

PR #52162, opened 2026-08-04
enhance: make StorageV2 multipart upload size configurable

Slontia:
## What this PR does

- Introduces `dataNode.storage.multiPartUploadSize` (default 10 MiB, valid range 5 MiB – 5 GiB, hot-update supported) and wires it through the StorageV2 and StorageV3 write paths: flush sync task (V2 packed writer, V3 packed writer and TEXT/LOB FFI segment writer), mix compaction (`mergeSplit` and `mergeSort` via `buildWriterOptions`), sort compaction, clustering compaction, bump-schema-version compaction, and the stats task.
- Fixes a pre-existing bug in `MultiSegmentWriter.rotateWriter()` where `WithUploader` and `WithVersion` were appended to `w.rwOption` in-place on each rotation, causing the option slice to grow linearly with the number of rotations; replaced with a local copy-before-append, plus a regression test asserting the option slice stays constant across rotations.

## Why

StorageV2/V3 packed writers default to 10 MiB per part. S3 enforces a hard limit of 10,000 parts per multipart upload, so any single output file exceeding ~97 GiB triggers an HTTP 400 error. The C++ AWS SDK then retries indefinitely, leaving compaction tasks permanently stuck in the `executing` state with no way to cancel them.

Setting `dataNode.storage.multiPartUploadSize: 209715200` (200 MiB) raises the per-file ceiling to ~1.95 TiB, which covers the large compaction output segments that trigger this failure. Out-of-range values fall back to the default with a warning.

issue: #53601

Note for maintainers: we run 2.6.x in production and would appreciate it if this fix could be considered for backport to 2.6 after it lands on master.

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -933,6 +933,10 @@ dataNode:
   bloomFilterApplyParallelFactor: 2 # parallel factor when to apply pk to bloom filter, default to 2*CPU_CORE_NUM
   storage:
     format: parquet # storage format for insert data, options: [parquet, vortex]
+    # The size of each part in multipart upload to object storage. Unit: Byte. Valid range: 5MiB ~ 5GiB.
+    # Only affects StorageV2 writers. Increase this value when large compaction output files exceed the
+    # S3 multipart upload limit of 10000 parts (e.g. set to 209715200 for 200MiB to support ~1.95TiB objects).
+    multiPartUploadSize: 10485760 # 10MiB default
     deltalog: json # deltalog format, options: [json, parquet]
   ip:  # TCP/IP address of dataNode. If not specified, use the first unicastable address
   port: 21124 # TCP port of dataNode
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -934,8 +934,9 @@ dataNode:
   storage:
     format: parquet # storage format for insert data, options: [parquet, vortex]
     # The size of each part in multipart upload to object storage. Unit: Byte. Valid range: 5MiB ~ 5GiB.
-    # Only affects StorageV2 writers. Increase this value when large compaction output files exceed the
-    # S3 multipart upload limit of 10000 parts (e.g. set to 209715200 for 200MiB to support ~1.95TiB objects).
+    # Affects StorageV2 (packed Parquet) and StorageV3 writers. Increase this value when large compaction
+    # output files exceed the S3 multipart upload limit of 10000 parts
+    # (e.g. set to 209715200 for 200MiB to support ~1.95TiB objects).
     multiPartUploadSize: 10485760 # 10MiB default
     deltalog: json # deltalog format, options: [json, parquet]
   ip:  # TCP/IP address of dataNode. If not specified, use the first unicastable address
--- a/internal/datanode/compactor/clustering_compactor.go
+++ b/internal/datanode/compactor/clustering_compactor.go
@@ -1071,6 +1071,7 @@ func (t *clusteringCompactionTask) getWriterOpts() []storage.RwOption {
 		storage.WithStorageConfig(t.compactionParams.StorageConfig),
 		storage.WithUseLoonFFI(t.compactionParams.UseLoonFFI),
 		storage.WithWriterFormat(t.compactionParams.GetStorageFormat()),
+		storage.WithMultiPartUploadSize(paramtable.Get().DataNodeCfg.MultiPartUploadSize.GetAsInt64()),
 	}
 
 	if t.lobContext != nil && t.lobContext.ShouldRewriteAnyField() {
--- a/internal/datanode/compactor/segment_writer.go
+++ b/internal/datanode/compactor/segment_writer.go
@@ -188,14 +188,16 @@ func (w *MultiSegmentWriter) rotateWriter() error {
 
 	chunkSize := w.binLogMaxSize
 
-	w.rwOption = append(w.rwOption,
+	opts := make([]storage.RwOption, len(w.rwOption), len(w.rwOption)+2)
+	copy(opts, w.rwOption)
+	opts = append(opts,
 		storage.WithUploader(func(ctx context.Context, kvs map[string][]byte) error {
 			return w.binlogIO.Upload(ctx, kvs)
 		}),
 		storage.WithVersion(w.storageVersion),
 	)
 	rw, err := storage.NewBinlogRecordWriter(w.ctx, w.collectionID, w.partitionID, newSegmentID,
-		w.schema, w.allocator.logIDAlloc, chunkSize, w.maxRows, w.rwOption...,
+		w.schema, w.allocator.logIDAlloc, chunkSize, w.maxRows, opts...,
 	)
 	if err != nil {
 		return err
--- a/internal/flushcommon/syncmgr/task.go
+++ b/internal/flushcommon/syncmgr/task.go
@@ -32,7 +32,6 @@ import (
 	"github.com/milvus-io/milvus/internal/storage"
 	"github.com/milvus-io/milvus/internal/storagecommon"
 	"github.com/milvus-io/milvus/internal/storagev2"
-	"github.com/milvus-io/milvus/internal/storagev2/packed"
 	"github.com/milvus-io/milvus/pkg/v3/metrics"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
--- a/internal/flushcommon/syncmgr/task.go
+++ b/internal/flushcommon/syncmgr/task.go
@@ -152,12 +152,12 @@ func (t *SyncTask) Run(ctx context.Context) (err error) {
 	case storage.StorageV2:
 		// New sync task means needs to flush data immediately, so do not need to buffer data in writer again.
 		writer := NewBulkPackWriterV2(t.metacache, t.schema, t.chunkManager, t.allocator, 0,
-			packed.DefaultMultiPartUploadSize, t.storageConfig, columnGroups, t.writeRetryOpts...)
+			paramtable.Get().DataNodeCfg.MultiPartUploadSize.GetAsInt64(), t.storageConfig, columnGroups, t.writeRetryOpts...)
 		t.insertBinlogs, t.deltaBinlog, t.statsBinlogs, t.bm25Binlogs, t.manifestPath, t.flushedSize, t.stats, err = writer.Write(ctx, t.pack)
 		statsWriter = writer
 	case storage.StorageV3:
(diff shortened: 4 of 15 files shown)


internal/flushcommon/syncmgr/task.go
```diff
@@ -152,12 +152,12 @@ func (t *SyncTask) Run(ctx context.Context) (err error) {
...
 		// New sync task means needs to flush data immediately, so do not need to buffer data in writer again.
 		writer := NewBulkPackWriterV2(t.metacache, t.schema, t.chunkManager, t.allocator, 0,
-			packed.DefaultMultiPartUploadSize, t.storageConfig, columnGroups, t.writeRetryOpts...)
+			paramtable.Get().DataNodeCfg.MultiPartUploadSize.GetAsInt64(), t.storageConfig, columnGroups, t.writeRetryOpts...)
 		t.insertBinlogs, t.deltaBinlog, t.statsBinlogs, t.bm25Binlogs, t.manifestPath, t.flushedSize, t.stats, err = writer.Write(ctx, t.pack)
 		statsWriter = writer
 	case storage.StorageV3:
 		writer := NewBulkPackWriterV3(t.metacache, t.schema, t.chunkManager, t.allocator, 0,
-			packed.DefaultMultiPartUploadSize, t.storageConfig, columnGroups, segmentInfo.ManifestPath(), t.writeRetryOpts...)
+			paramtable.Get().DataNodeCfg.MultiPartUploadSize.GetAsInt64(), t.storageConfig, columnGroups, segmentInfo.ManifestPath(), t.writeRetryOpts...)
```
yhmo, line 159: internal/flushcommon/syncmgr/task.go line:160
High ---- This passes the configured value into the StorageV3 writer, but the downstream newPackedRecordBatchWriter ignores its multiPartUploadSize argument and calls packed.NewFFIPackedWriter without adding fs.multi_part_upload_size to the FFI properties. As a result V3 uploads still use the default part size and can hit the same 10,000-part/stuck-compaction failure this PR is meant to fix. Please propagate the value into the FFI properties and add a property-level V3 regression test.

internal/datanode/compactor/clustering_compactor.go
```diff
@@ -1071,6 +1071,7 @@ func (t *clusteringCompactionTask) getWriterOpts() []storage.RwOption {
 		storage.WithStorageConfig(t.compactionParams.StorageConfig),
 		storage.WithUseLoonFFI(t.compactionParams.UseLoonFFI),
 		storage.WithWriterFormat(t.compactionParams.GetStorageFormat()),
+		storage.WithMultiPartUploadSize(paramtable.Get().DataNodeCfg.MultiPartUploadSize.GetAsInt64()),
```
yhmo, line 1074: internal/datanode/compactor/clustering_compactor.go line:1074
Medium ---- The propagation audit is still missing live packed-writer paths: bumpSchemaVersionCompactionTask omits this option for its full rewrite and hard-codes packed.DefaultMultiPartUploadSize for its partial V3 writer; BulkPackWriterV3.writeDelta also owns the configured size but does not pass it to NewDeltalogWriter. Those outputs retain the ~97 GiB ceiling. Please thread the new setting through these sibling writers and cover the inventory in tests.

configs/milvus.yaml
```diff
@@ -933,6 +933,10 @@ dataNode:
   bloomFilterApplyParallelFactor: 2 # parallel factor when to apply pk to bloom filter, default to 2*CPU_CORE_NUM
   storage:
     format: parquet # storage format for insert data, options: [parquet, vortex]
+    # The size of each part in multipart upload to object storage. Unit: Byte. Valid range: 5MiB ~ 5GiB.
+    # Only affects StorageV2 writers. Increase this value when large compaction output files exceed the
```
yhmo: configs/milvus.yaml line:937
Low ---- This says V2 only, but the PR intentionally passes the setting to StorageV3 flush and manifest writers as well. Please describe the actual packed-writer scope so operators understand that changing the value also affects V3 memory use and retry cost.

internal/datanode/compactor/segment_writer.go
```diff
@@ -188,14 +188,16 @@ func (w *MultiSegmentWriter) rotateWriter() error {
 
 	chunkSize := w.binLogMaxSize
 
-	w.rwOption = append(w.rwOption,
+	opts := make([]storage.RwOption, len(w.rwOption), len(w.rwOption)+2)
```
yhmo, line 191: internal/datanode/compactor/segment_writer.go line:191
Low ---- Could we extend the existing multi-rotation test to assert that len(writer.rwOption) remains unchanged after several rotations? The previous implementation still produced valid segments, so current behavioral tests would not catch the linear option-slice growth that this change is fixing.

internal/flushcommon/syncmgr/task.go
```diff
@@ -152,12 +151,12 @@ func (t *SyncTask) Run(ctx context.Context) (err error) {
...
```
yhmo, line 159: internal/flushcommon/syncmgr/task.go line:159
Medium ---- This StorageV3 path selects NewPackedTextBatchWriter when TEXT column configs are present, but that function never uses its multiPartUploadSize parameter and calls packed.NewFFISegmentWriter with only the SegmentWriterConfig and storage config. Large TEXT/LOB flushes and compactions therefore remain on the default part size even after this setting is raised. Please pass fs.multi_part_upload_size into the SegmentWriter properties and add a TEXT-path regression test.

internal/flushcommon/syncmgr/task.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

configs/milvus.yaml
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/datanode/compactor/segment_writer.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/flushcommon/syncmgr/task.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/datanode/compactor/clustering_compactor.go
Slontia: Addressed in the latest revision: `bumpSchemaVersionCompactionTask` now passes the configured `dataNode.storage.multiPartUploadSize` to both partial-writer variants (`NewPartialPackedRecordBatchWriter` and `NewPartialPackedRecordBatchWriterWithTextRefsAsBinary`), the full-rewrite path propagates it via the shared writer options, and `BulkPackWriterV3.writeDelta` passes it to `NewDeltalogWriter`. Also rebased onto latest master. Note: the streamingnode WAL `pack_writer.go` path still uses `packed.DefaultMultiPartUploadSize` — it belongs to a different component/config namespace, so I left it as a follow-up rather than expanding this PR's scope.

internal/storagev2/packed/ffi_common.go
```diff
@@ -66,10 +66,11 @@ var (
...
 
-	PropertyWriterPolicy             = C.GoString(C.loon_properties_writer_policy)
-	PropertyWriterFormat             = C.GoString(C.loon_properties_writer_format)
-	PropertyWriterSchemaBasedPattern = C.GoString(C.loon_properties_writer_schema_base_patterns)
-	PropertyWriterSchemaBasedFormats = "writer.split.schema_based.formats"
+	PropertyWriterPolicy              = C.GoString(C.loon_properties_writer_policy)
+	PropertyWriterFormat              = C.GoString(C.loon_properties_writer_format)
+	PropertyWriterSchemaBasedPattern  = C.GoString(C.loon_properties_writer_schema_base_patterns)
+	PropertyWriterSchemaBasedFormats  = "writer.split.schema_based.formats"
+	PropertyWriterMultiPartUploadSize = "writer.multi_part_upload_size"
```
xiaofan-luan: [P1] Use the multipart property recognized by the pinned storage library

At this head, milvus-storage is pinned to `15ab3d7`, where the property is [`fs.multi_part_upload_size`](https://github.com/milvus-io/milvus-storage/blob/15ab3d7d3f2c222a9ae70bbdae7520b8eb2ca810/cpp/include/milvus-storage/properties.h#L90), rather than `writer.multi_part_upload_size`.

The FFI property conversion accepts undefined keys, but the [Parquet writer](https://github.com/milvus-io/milvus-storage/blob/15ab3d7d3f2c222a9ae70bbdae7520b8eb2ca810/cpp/src/format/parquet/parquet_writer.cpp#L174-L182) and [filesystem configuration](https://github.com/milvus-io/milvus-storage/blob/15ab3d7d3f2c222a9ae70bbdae7520b8eb2ca810/cpp/src/filesystem/fs.cpp#L279-L280) read the `fs.*` key. Consequently, setting this configuration to 200 MiB still leaves V3 ordinary and TEXT writers using the 10 MiB default, without reporting a configuration error. The original multipart-limit failure remains possible.

Please use the exported `C.loon_properties_fs_multi_part_upload_size` and add regression coverage that verifies the effective part size consumed by the storage library for both ordinary and TEXT writers.

xiaofan-luan (commented):
Rechecked head `63839349920f5c877740dc993aaf7617bab60dd5`. The V3 property propagation still has the issue below. Validation was source tracing against the pinned storage dependency; no large-object upload integration test was run.

internal/datanode/compactor/clustering_compactor.go
xiaofan-luan: Thanks for updating the partial writers and `BulkPackWriterV3.writeDelta`; those now receive the configured value. One propagation gap remains at head `63839349920f5c877740dc993aaf7617bab60dd5`:

[P2] The full schema rewrite does not use the shared writer options. Dropping a physical field routes to `runFullSchemaRewrite()` at lines 120-121. That method builds its own [`writerOpts` at lines 676-684](https://github.com/milvus-io/milvus/blob/63839349920f5c877740dc993aaf7617bab60dd5/internal/datanode/compactor/bump_schema_version_compactor.go#L676-L684), without `WithMultiPartUploadSize`, and passes them directly to `NewBinlogRecordWriter` at lines 703-711. It does not call `newV3WriterResult()` or the mix-compaction option builder.

This leaves the full rewrite on `DefaultWriterOptions()`'s 10 MiB value. After fixing the FFI property name, a large V3 segment written with 200 MiB parts can therefore still encounter the original multipart limit when a field is dropped and the segment is rewritten.

Please also add `WithMultiPartUploadSize` to the full-rewrite options and cover both additive and full-rewrite schema-bump branches.

Issue #52164, opened 2026-08-04
[Bug]: planparserv2 test fixture allocates field IDs that collide with any new DataType enum value


bigsheeper:
## Environment

- **Milvus Version**: master @ 80435d159d
- **Deployment Mode**: n/a — unit test
- **MQ**: n/a
- **SDK**: n/a
- **OS**: n/a

### Option B: Steps

1. Bump `github.com/milvus-io/milvus-proto/go-api/v3` from the version master currently pins (`v3.0.0-20260712212354-7efdf25099dd`) to any version that contains `DataType_Decimal = 30` / `DataType_UUID = 31`, e.g. `v3.0.0-20260803063812-659295cc95a3`.
2. Run `go test -tags dynamic,test -gcflags="all=-N -l" ./internal/parser/planparserv2/`

### Trigger Conditions

- **Frequency**: always, once the proto contains a `DataType` enum value >= 30
- **Does NOT happen when**: proto is pinned at a version whose highest non-vector `DataType` value is <= 29

## Expected Behavior

Adding a `DataType` enum value in milvus-proto does not break unrelated parser tests.

## Actual Behavior

Every test that builds a schema through `newTestSchema` fails:

```
duplicated fieldID: 130: invalid parameter
```

6+ tests fail, including `TestExpr_BloomMatch`, `TestPlanContainsBloomFilter`, `TestCheckIdentical`, and all `TestFillExpressionValue` subtests.

## Error Logs

```
--- FAIL: TestExpr_BloomMatch (0.00s)
    plan_parser_v2_test.go:90:
        Error: Received unexpected error:
               duplicated fieldID: 130: invalid parameter
               Wraps: (2) duplicated fieldID: 130
               Wraps: (3) invalid parameter
FAIL internal/parser/planparserv2.TestExpr_BloomMatch
```

## Non-default Configuration

None.

## Analysis Hints

`newTestSchema` (`internal/parser/planparserv2/plan_parser_v2_test.go:24`) allocates a field per `DataType` enum value as `100 + value`:

```go
for name, value := range schemapb.DataType_value {
    newField := &schemapb.FieldSchema{
        FieldID: int64(100 + value), ...
    }
}
```

and then appends fields with hardcoded IDs **130, 131, 132, 133, 134** — the dynamic field, `StringArrayField`, and the struct-array fields.

[...]

PR #52165, opened 2026-08-04
fix: keep planparserv2 test field IDs clear of the DataType-derived range

bigsheeper:
issue: #52164

## Summary

`newTestSchema` in `planparserv2` derives one field ID per `DataType` enum value as `100+value`, then appends hand-written fields at **130-134**. Both live in one ID space, so the fixture only held while the highest non-vector `DataType` was 29. The proto now has `Decimal = 30` and `UUID = 31`, which generate 130 and 131 and collide with the dynamic field and `StringArrayField` — `CreateSchemaHelper` then rejects every schema the fixture builds, failing 8 tests.

This is latent rather than new: any future `DataType` addition would have hit it.

## Changes

**`internal/parser/planparserv2/plan_parser_v2_test.go`**
Move the hand-written fields to named constants past `100+max(DataType)` (301 today, via `DataType_Struct(201)`), and have the assertions reference those constants instead of repeating the literals.

**`pkg/util/fastpb/`**
Bumping milvus-proto far enough to reach that `DataType` range also trips `TestProtoContract_FieldSetsPinned`, because `schemapb.IDs` gained a third oneof variant, `uuid_id`.

The guard's own guidance requires a new oneof variant to be decoded in-pass: anything left to the deferred `protoMerge` is applied *after* the decode loop, so it wins regardless of wire position and breaks oneof last-wins. Added `case 3`, plus a test that puts `uuid_id` ahead of `int_id`/`str_id` on one wire — a single-variant payload cannot distinguish the two, since `proto.Marshal` only emits the variant that is set.

`InsertRequest` and `UpsertRequest` also gained `rls_principal` / `skip_rls`. Those are plain scalars, correct via the existing `default -> rest -> protoMerge` fallback, so only their golden sets are updated.

**`go.mod` ×4** — bump `milvus-proto/go-api/v3` to `v3.0.0-20260803063812-659295cc95a3`.

## Test Plan

- [x] `./internal/parser/planparserv2/` — ok (was 8 failures)
- [x] `pkg/util/fastpb/` — ok
- [x] `go build -tags dynamic ./internal/...`, `pkg`, `client` — ok
- [x] `gofumpt` / `gci` on all changed files — clean

Both new tests were verified to discriminate, each reverted afterwards:

- Removing `case 3` fails `TestEquiv_IDs_OneofLastWins` (both subtests) and nothing else. Note the plain round-trip cases added alongside it do **not** catch this — they pass with `case 3` removed, which is why the hand-built wire test exists.
- The fixture change is pinned by the existing 8 tests, which fail on the collision without it.

`make static-check` was **not** run: the local golangci-lint binary is incompatible with the current Go toolchain (`unsupported version: 2`). The gofumpt/gci checks above were run directly. Relying on CI for the full lint.

## Note for reviewers

This bump is what several pending PRs need in order to consume `PrivilegeImportBinlog` (milvus-proto#643) — master still pins the 2026-07-12 proto, three weeks behind. The `uuid_id` and RLS fields come from other people's proto changes that landed in between; there is no milvus-side implementation for either on master yet (`DataType_UUID`, `UUIDArray`, `IDs_UuidId` have no non-test usage). Landing this separately keeps the decoder change reviewable on its own rather than buried in an unrelated feature PR.

### DIFF
--- a/client/go.mod
+++ b/client/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/cockroachdb/errors v1.9.1
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
-	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260712212354-7efdf25099dd
+	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260803063812-659295cc95a3
 	github.com/quasilyte/go-ruleguard/dsl v0.3.23
 	github.com/samber/lo v1.52.0
 	github.com/stretchr/testify v1.11.1
--- a/client/go.sum
+++ b/client/go.sum
@@ -158,6 +158,8 @@ github.com/mediocregopher/radix/v3 v3.4.2/go.mod h1:8FL3F6UQRXHXIBSPUs5h0RybMF8i
 github.com/microcosm-cc/bluemonday v1.0.2/go.mod h1:iVP4YcDBq+n/5fb23BhYFvIMq/leAFZyRl6bYmGDlGc=
 github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260712212354-7efdf25099dd h1:q24BFW1uhH7VxueDw3t1V6a9RZuMorjyOya5+oxIUAE=
 github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260712212354-7efdf25099dd/go.mod h1:rbKpv5JToISTKTTLl0duL5r6wbYnjJ9SsD0QgXMzKy0=
+github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260803063812-659295cc95a3 h1:EnDFYLoneAhdn5EyJ1OOTcBXT90d4FgXW9BDRBVngTQ=
+github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260803063812-659295cc95a3/go.mod h1:rbKpv5JToISTKTTLl0duL5r6wbYnjJ9SsD0QgXMzKy0=
 github.com/mitchellh/go-homedir v1.1.0/go.mod h1:SfyaCUpYCn1Vlf4IUYiD9fPX4A5wJrkLzIz1N1q0pr0=
 github.com/mitchellh/mapstructure v1.1.2/go.mod h1:FVVH3fgwuzCH5S8UJGiWEs2h04kUh9fWfEaFds41c1Y=
 github.com/modern-go/concurrent v0.0.0-20180228061459-e0a39a4cb421/go.mod h1:6dJC0mAP4ikYIbvyc7fijjWJddQyLn8Ig3JB5CqoB9Q=
--- a/client/milvusclient/mock_milvus_server_test.go
+++ b/client/milvusclient/mock_milvus_server_test.go
@@ -2135,6 +2135,65 @@ func (_c *MilvusServiceServer_DeleteCredential_Call) RunAndReturn(run func(conte
 	return _c
 }
 
+// DeleteRLSPrincipalTags provides a mock function with given fields: _a0, _a1
+func (_m *MilvusServiceServer) DeleteRLSPrincipalTags(_a0 context.Context, _a1 *milvuspb.DeleteRLSPrincipalTagsRequest) (*commonpb.Status, error) {
+	ret := _m.Called(_a0, _a1)
+
+	if len(ret) == 0 {
+		panic("no return value specified for DeleteRLSPrincipalTags")
+	}
+
+	var r0 *commonpb.Status
+	var r1 error
+	if rf, ok := ret.Get(0).(func(context.Context, *milvuspb.DeleteRLSPrincipalTagsRequest) (*commonpb.Status, error)); ok {
+		return rf(_a0, _a1)
+	}
+	if rf, ok := ret.Get(0).(func(context.Context, *milvuspb.DeleteRLSPrincipalTagsRequest) *commonpb.Status); ok {
+		r0 = rf(_a0, _a1)
+	} else {
+		if ret.Get(0) != nil {
+			r0 = ret.Get(0).(*commonpb.Status)
+		}
+	}
+
+	if rf, ok := ret.Get(1).(func(context.Context, *milvuspb.DeleteRLSPrincipalTagsRequest) error); ok {
+		r1 = rf(_a0, _a1)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
+}
+
+// MilvusServiceServer_DeleteRLSPrincipalTags_Call is a *mock.Call that shadows Run/Return methods with type explicit version for method 'DeleteRLSPrincipalTags'
+type MilvusServiceServer_DeleteRLSPrincipalTags_Call struct {
+	*mock.Call
+}
+
+// DeleteRLSPrincipalTags is a helper method to define mock.On call
+//   - _a0 context.Context
+//   - _a1 *milvuspb.DeleteRLSPrincipalTagsRequest
+func (_e *MilvusServiceServer_Expecter) DeleteRLSPrincipalTags(_a0 interface{}, _a1 interface{}) *MilvusServiceServer_DeleteRLSPrincipalTags_Call {
+	return &MilvusServiceServer_DeleteRLSPrincipalTags_Call{Call: _e.mock.On("DeleteRLSPrincipalTags", _a0, _a1)}
+}
+
+func (_c *MilvusServiceServer_DeleteRLSPrincipalTags_Call) Run(run func(_a0 context.Context, _a1 *milvuspb.DeleteRLSPrincipalTagsRequest)) *MilvusServiceServer_DeleteRLSPrincipalTags_Call {
+	_c.Call.Run(func(args mock.Arguments) {
+		run(args[0].(context.Context), args[1].(*milvuspb.DeleteRLSPrincipalTagsRequest))
+	})
+	return _c
+}
+
+func (_c *MilvusServiceServer_DeleteRLSPrincipalTags_Call) Return(_a0 *commonpb.Status, _a1 error) *MilvusServiceServer_DeleteRLSPrincipalTags_Call {
+	_c.Call.Return(_a0, _a1)
+	return _c
+}
+
+func (_c *MilvusServiceServer_DeleteRLSPrincipalTags_Call) RunAndReturn(run func(context.Context, *milvuspb.DeleteRLSPrincipalTagsRequest) (*commonpb.Status, error)) *MilvusServiceServer_DeleteRLSPrincipalTags_Call {
+	_c.Call.Return(run)
+	return _c
+}
+
 // DeleteUserTags provides a mock function with given fields: _a0, _a1
 func (_m *MilvusServiceServer) DeleteUserTags(_a0 context.Context, _a1 *milvuspb.DeleteUserTagsRequest) (*commonpb.Status, error) {
 	ret := _m.Called(_a0, _a1)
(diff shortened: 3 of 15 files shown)


czs007:
This PR is a mechanical milvus-proto bump plus a new fastpb `uuid_id` decoder, regenerated mocks, and a planparserv2 fixture refactor; review surfaced one test-coverage gap in the new decoder's contract test and one pre-existing validation hazard.

## Low

- **`pkg/util/fastpb/proto_contract_test.go:70` — `schemapb.UUIDArray` is missing from the leaf-array golden set.** This PR adds a hand-written decode branch for `UUIDArray` (`pkg/util/fastpb/searchresult.go:266-271`, via `decodeRepeatedBytes`) and updates several other golden sets (`schemapb.IDs` → `{1,2,3}`, `InsertRequest`/`UpsertRequest` extended), but the leaf-array block that pins `BytesArray`/`JSONArray` still has no `UUIDArray` entry, so the new hand-decoded type has no contract tripwire. This is a coverage gap rather than a decode bug: `decodeRepeatedBytes` falls back to `fallbackUnmarshal` on any non-`(1, bytes)` field (`pkg/util/fastpb/helpers.go:386-388`), and `fallbackUnmarshal` calls `proto.Reset` before `proto.Unmarshal` (`helpers.go:35-38`), so a future added field would silently degrade the fast path instead of mis-parsing. Suggestion: add `{"schemapb.UUIDArray", &schemapb.UUIDArray{}, []int{1}}` alongside the existing `BytesArray`/`JSONArray` entries. *(raised by zack, xiaocai)*

- **`internal/rootcoord/util.go:584` — `validateFieldDataType` gates only on enum membership, so unimplemented scalar types can be created but never written.** The check is `schemapb.DataType_name` lookup plus a `DataType_None` exclusion, with no positive supported-type whitelist; insert then hits the `default` branch at `pkg/util/funcutil/func.go:705`. Failure: a raw-gRPC client sends `CreateCollection` with a scalar field of `DataType_UUID(31)` (or `Decimal(30)`, or the already-present `Mol`/`Date`/`Time`) → the collection is created and persisted, but every subsequent `Insert` returns `ParameterInvalid "31 is not supported now"` from `funcutil.GetNumRowOfFieldDataWithSchema`, leaving a permanently unwritable collection recoverable only by `DropCollection`. Unrelated to this PR: the diff touches neither file (it contains only go.mod/go.sum bumps, mockery output, a planparserv2 test fixture, and `pkg/util/fastpb`), and the old pinned proto already carried `Mol=27`/`Date=28`/`Time=29` with zero non-test references under `internal/`+`pkg/`, so the identical create-succeeds/insert-fails behavior reproduces on master today — the bump only widens an already-unguarded set from three values to five. The durable fix is a positive supported-type whitelist, best owned by a separate hardening change or by the PR that implements these types. *(raised by gzh, xiaocai, zack)*

bigsheeper:
Thanks — the Low finding holds and is fixed in `e892af941b`. I verified each premise against the head you reviewed (`237a3769`) rather than judging from the description.

**Low (`proto_contract_test.go:70`) — confirmed, fixed.**

Your reading is right, and the gap is mine: the leaf-array block's own stated scope is "leaf arrays with their own hand-written field switches", and `UUIDArray` qualifies on exactly the same grounds as the two types already pinned there. `searchresult.go:266-271` decodes it through `decodeRepeatedBytes` (`helpers.go:371`) — the identical helper that backs `BytesArray` and `JSONArray`, both listed as `[]int{1}`. This PR updated the `schemapb.IDs` entry to `{1,2,3}` for the new `uuid_id` variant but did not extend the leaf-array block, so the new hand-decoded type shipped without a tripwire. Added `{"schemapb.UUIDArray", &schemapb.UUIDArray{}, []int{1}}` as you suggested.

Your severity call is also right, and I checked it rather than taking it: this is coverage, not a decode bug. `decodeRepeatedBytes` fast-paths only `num == 1 && wtype == 2` and sends anything else to `fallbackUnmarshal(full, m)` (`helpers.go:386-388`), which does `proto.Reset` then `proto.Unmarshal` (`helpers.go:35-38`) — so a future field on `UUIDArray` degrades the whole submessage to the official codec instead of mis-parsing. What was missing was only the signal that it happened.

I confirmed the new row actually is a tripwire rather than a passing assertion: with the golden set deliberately set to `{1, 2}` the test fails with `schemapb.UUIDArray proto field set changed (want [1 2], got [1])`; restored to `{1}` it passes. `./pkg/util/fastpb/...` is green and `golangci-lint` reports 0 issues on the package.

**Pre-existing hazards — agreed, not taken here.**

I'm leaving `internal/rootcoord/util.go:584` (`validateFieldDataType` gating on enum membership only) alone, per your own classification. I checked the premise you gave for it: this PR's diff contains only go.mod/go.sum bumps, mockery output, a planparserv2 test fixture, and `pkg/util/fastpb`, so it touches neither `rootcoord/util.go` nor `funcutil.GetNumRowOfFieldDataWithSchema` — the create-succeeds/insert-fails behavior reproduces on master today, and the bump only widens the unguarded enum set. A positive supported-type whitelist is the right fix but belongs to a separate hardening change, not to a proto bump. No tracking issue filed yet.

czs007:

## Overview

This PR's substantive change is the in-pass `uuid_id` oneof decode in the fastpb `IDs` decoder, plus a proto bump, regenerated mocks, and a test-fixture field-ID remap; the notes below are one coverage gap in the new decoder branch and one pre-existing hazard found nearby.

## Low

- **`pkg/util/fastpb/message_merge_test.go:135` — the new `uuid_id(3)` duplicate branch has no direct test.**
  This PR widened the oneof duplicate guard at `pkg/util/fastpb/searchresult.go:247` from `num == 1 || num == 2` to include `3`, so a wire payload carrying `uuid_id` twice now takes `fallbackUnmarshal(full, ids)` (`searchresult.go:248-250`). Before this PR, field 3 fell through `default → rest → protoMerge` — an entirely different code path. `TestOneofRepeatedMerge` currently has only an `IDs.IntId` duplicate-same-variant case (`message_merge_test.go:135-144`), with no `uuid_id` counterpart; every other `UuidId` occurrence in the package's tests (`equiv_test.go:85/90/91/101`, `proto_contract_test.go:78`) is a single occurrence, and the new `TestEquiv_IDs_OneofLastWins` covers `uuid_id` followed by a *different* variant, not `uuid_id` twice.
  *Suggestion:* add an `IDs.UuidId` subtest alongside `message_merge_test.go:135`, mirroring the existing `IntId` shape. The current behavior was verified correct against `proto.Unmarshal` on the PR head — this is about locking it in, not fixing a defect.
  (raised by tinswzy)

- **`internal/parser/planparserv2/plan_parser_v2_test.go:267` — a hardcoded `FieldID: 199` still sits inside the `100 + DataType` derived window.** `withNullField` in `TestExpr_NullLiteral_LegacyNullField` appends a field with ID `199`, which falls in the `[100, 301]` range this PR's own new comment identifies as growing with the `DataType` enum (`DataType_Struct = 201`); it happens not to collide only because the enum has no value `99` today (values are 0-5/10/11/20-31/100-106/200/201). Folding it into the same family of constants introduced here (e.g. `legacyNullFieldID = 1005`) would fully decouple the fixture from `DataType` growth. *Unrelated to this PR:* line 267 is untouched by the diff — which only adds the constant block, replaces five `FieldID`s in `newTestSchema`, and remaps eight assertion literals — and the line behaves identically with or without this PR.
  (surfaced during verification)

bigsheeper:
Thanks — both items are confirmed and taken in `6fe14eb9d4`, including the one you scoped as follow-up only. Verified against the head you reviewed (`e892af941b`) rather than judging from the description.

**Low (`message_merge_test.go:135`) — confirmed, fixed. Your severity was right but the reasoning is stronger than "coverage".**

Everything you traced holds: this PR widened the duplicate guard at `searchresult.go:247` from `num == 1 || num == 2` to include `3`, so a repeated `uuid_id` now takes `fallbackUnmarshal` where it previously fell through `default → rest → protoMerge`; `TestRepeatedMessageOneofMerge` had an `IntId` duplicate case and no `UuidId` counterpart; and every other `UuidId` occurrence in the package is a single occurrence.

What I did not expect until I ran the RED: the new subtest is not pinning a path that happens to agree. Reverting the guard to `num == 1 || num == 2` makes it fail with `got=uuid_id:{data:"\xbb"}` against `want=uuid_id:{data:"\xaa" data:"\xbb"}`. proto3 merges a repeated submessage appearing twice field-by-field, so two `uuid_id` occurrences must concatenate their `data`; the hand-written `case 3` this PR added would replace instead. The guard widening is what keeps that correct, and the two changes are only correct together. Your "verified correct on the PR head, this is about locking it in" reading is right — but what needed locking in is a live correctness pairing, not an inert branch. Good catch.

**Pre-existing (`plan_parser_v2_test.go:267`) — confirmed, and I took it in this PR rather than deferring it.**

Your analysis is exact, including why it does not collide today. I pulled the full `DataType_value` table from the generated proto to check: values are `0-5, 10-11, 20-31, 100-106, 200-201`, so the derived set is `{100-105, 110, 111, 120-131, 200-206, 300, 301}` and `199` is genuinely absent — it is clear only because the enum has no `99`, exactly as you said.

I disagree with the follow-up-only classification, though, and I want to say why rather than just acting. This PR's whole purpose is to keep `newTestSchema`'s hand-written field IDs clear of the `DataType`-derived window, and it introduced a constant block to do it. Leaving one hardcoded ID inside `[100, 301]` in the same function family is this PR's own goal left unfinished, not an inherited hazard — the same shape as the `130` collision it exists to fix. It is `legacyNullFieldID = 1005` now, two lines.

**One thing your review did not cover, which I checked because of it.** You flagged the site at `:267`; I audited every site that appends to a `newTestSchema` result rather than only that one. There are two. The other (`:1051`, `max_edit_distance`) uses field ID `9999`, already far outside the window, so I left it alone — recorded in the commit message so the audit is visible rather than assumed.

Verification: RED/GREEN on the new subtest as described above; `TestRepeatedMessageOneofMerge` all six subtests pass; `./internal/parser/planparserv2/` and `pkg/util/fastpb` are green; `golangci-lint` reports 0 issues on both packages.

czs007:

One medium-severity validation gap surfaced: schema creation accepts the newly-nameable `Decimal`/`UUID` field types that no write path supports.

## Medium

**`internal/proxy/util.go:574` — `validateFieldType` does not reject `Decimal`/`UUID`, so collection creation succeeds but inserts fail later.**

`validateFieldType` is an implicit allow-list-by-omission: it handles only `String`, `None`, and `Array` (delegating to `validateElementType`), and the `switch` has no `default`, so every other `DataType` enum value passes through (util.go:574-586). `ValidateField` (util.go:694-710) likewise branches only on vector / `VarChar` / `Array` / `ArrayOfVector` and never matches `Decimal(30)` / `UUID(31)`. Both are the only field-type gates in `createCollectionTask.PreExecute` (task.go:589, task.go:643).

Failure: creating a collection with `{"fieldName":"u","dataType":"UUID"}` via REST v2 `POST /v2/vectordb/collections/create` (or `field.data_type=31` over gRPC) → `CreateCollection` returns success and the collection is persisted, but any subsequent `Insert` carrying that field fails at validate_util.go:434 → funcutil/func.go:705 with `UUID is not supported now`; the collection is permanently unusable for writes and can only be dropped and recreated, with an error message that does not point at the real cause.

The two entry paths differ in provenance and are worth separating:
- **gRPC:** `data_type` is an open `int32` enum on the wire, so a client could already submit `30`/`31` before this PR and `validateFieldType` would already let it through — a pre-existing, type-generic gap.
- **REST v2:** `dataType` is a *string* resolved through `schemapb.DataType_value` (request_v2.go:790, handler_v2.go:2532), which returns `data type %s is invalid(case sensitive)` on a miss. The previously pinned proto (`v3.0.0-20260712212354`) contains no `DataType_Decimal` / `DataType_UUID`, while the bumped proto (`v3.0.0-20260803063812`, schema.pb.go:48-49) defines `Decimal = 30` / `UUID = 31`. The REST layer was therefore blocked only incidentally by the enum name table, and this PR's proto bump removes that accidental barrier — `{"dataType": "UUID"}` goes from rejected to accepted.

Suggestion: convert `validateFieldType` to an explicit allow-list with a rejecting `default`, or add an explicit rejection for `Decimal`/`UUID` until their data path lands. Note that this PR changes none of the production lines above (the diff is limited to go.mod/go.sum, regenerated mocks, tests, and `pkg/util/fastpb/searchresult.go`), so this is best handled alongside the UUID/Decimal feature work rather than as a gate on a proto-plumbing bump. (provenance unclear — please confirm whether this PR introduces it) (raised by chyezh, tinswzy)

bigsheeper:
Answering the provenance question you flagged: **yes for REST v2, no for gRPC** — and I think that makes this stronger than "pre-existing, handle with the feature work".

I verified each step rather than reasoning from the description:

- `validateFieldType` (`internal/proxy/util.go:574`) is an allow-list by omission exactly as you describe: it handles `String`, `None` and `Array`, and the `switch` has no `default`, so every other `DataType` falls through. `ValidateField` never matches `Decimal`/`UUID` either, and those two are the only field-type gates in `createCollectionTask.PreExecute` (`task.go:589`, `task.go:645`).
- Insert then dies at `funcutil/func.go:705` — the `default` branch returning `%s is not supported now` — so the collection is created and permanently unwritable, with an error that does not name the cause.
- The proto delta is real. I grepped both module versions: the pinned `v3.0.0-20260712212354` contains **no** `Decimal`/`UUID` entries at all, and `v3.0.0-20260803063812` defines `"Decimal": 30` and `"UUID": 31`.
- REST v2 resolves the string through `schemapb.DataType_value` at both entry points (`request_v2.go:790`, `handler_v2.go:2532`), and misses are rejected with `data type %s is invalid(case sensitive)`.

So your reading holds exactly: the enum name table was the only thing rejecting `{"dataType": "UUID"}` over REST, and this PR's bump removes it. On gRPC the field is an open `int32` and `30`/`31` already fell through before the bump — that half is pre-existing and type-generic, as you said.

**Where I'd put it differently from your closing note.** You suggest handling this with the UUID/Decimal feature work rather than gating a proto-plumbing bump, on the grounds that this PR changes none of the production lines involved. The second half is true — the diff is go.mod/go.sum, regenerated mocks, a planparserv2 fixture, `pkg/util/fastpb`, and a merge test. But "touches no production line" and "introduces no behavior change" come apart here: for REST, `{"dataType":"UUID"}` goes from rejected to accepted *because of this PR*, and unlike the `validateFieldDataType` hazard you raised in the earlier round, this half does **not** reproduce on master. A bump that silently widens an externally reachable API surface is a behavior change even when its diff is all go.mod.

That said, I am not treating it as a merge gate on my own judgment, because there is a real question I cannot answer from the code: whether `Decimal`/`UUID` are meant to be reachable from the moment the proto lands (someone's feature branch depending on create-time acceptance) or only when their data path merges. If it is the latter, the fix belongs here and is small — an explicit rejection in `validateFieldType` for the types with no write path, which restores the pre-bump REST behavior rather than adding new policy, and incidentally closes the pre-existing gRPC half for those two types as well.

I would rather do that than ship a bump that leaves a create-succeeds/insert-always-fails shape reachable over REST. Say which way you want it and I will either add the rejection here or open a follow-up issue naming both entry paths and the `validateFieldType` allow-list conversion as the durable fix.

Verification note for the rest of the PR: the two items from your previous round are in `6fe14eb9d4`. The `IDs.UuidId` duplicate subtest turned out to pin a live correctness pairing rather than an inert branch — reverting the guard widening makes it fail with `got={0xbb}` against `want={0xaa,0xbb}`, since proto3 merges a repeated submessage appearing twice and the hand-written `case 3` would replace instead.

czs007:

## Re-review of `6fe14eb9d44a`

- **Confirmed** — `internal/proxy/util.go:574` (validateFieldType does not reject Decimal/UUID): re-verified against code this round; the blacklist-style switch still lets `Decimal(30)`/`UUID(31)` through with no `default` branch. Both reviewers additionally established that the gap is pre-existing rather than introduced by this PR, so it is re-scoped to the pre-existing-hazards section below instead of standing as a finding against this change.

## This round's findings

Adversarial review found no new issues introduced by this PR; the one confirmed hazard is pre-existing and listed separately below.

Verified:
- pkg/util/fastpb/searchresult.go:247 — the duplicate-oneof guard now covers `num == 3`, so a repeated `uuid_id` routes to the official codec and preserves proto3 merge semantics instead of an incorrect in-pass replace.
- pkg/util/fastpb/searchresult.go:266 — the in-pass `uuid_id` decode copies each bytes element and falls back to a full official re-decode of `UUIDArray` on any unexpected tag or wire type, so malformed input degrades rather than mis-parses.
- pkg/util/fastpb/equiv_test.go:1064 — the hand-built dual-variant wire test is the only construction that can distinguish in-pass decode from the deferred merge path, so a missing `case 3` cannot silently regress.
- pkg/util/fastpb/proto_contract_test.go:70 — the golden field sets were widened for the new Insert/Upsert scalars and `UUIDArray`, and the new scalars are correctly handled by the existing default-to-fallback path with no hand-written cases needed.
- internal/parser/planparserv2/plan_parser_v2_test.go:861 — the relocated fixture field IDs (1000–1005) all sit outside the generated window [100, 301], including the previously stray ID 199; the change is test-only.
- internal/mocks/mock_proxy.go — both checked-in hand-written mocks covering the MilvusService surface were regenerated for the 6 new RPCs, and real servers embed `UnimplementedMilvusServiceServer`, so no implementer is left compile-broken.

- `internal/proxy/util.go:574` — `validateFieldType` is a blacklist-style switch (only `String`/`None`/`Array` branches, no `default`), so fields declared as `Decimal(30)` or `UUID(31)` pass proxy and rootcoord validation at CreateCollection; a client creating a collection with such a field succeeds and leaves unusable metadata → the failure only surfaces deep in the serialization/storage path at insert time as an internal error rather than a clear `ParameterInvalid`, since `validateUtil.Validate`'s empty `default` branch silently skips unknown data types. This is unrelated to this PR because the PR only bumps the milvus-proto dependency and regenerates mocks/test fixtures — it touches no validation code, and the open-enum gap predates the bump (raw values 30/31 took the same accepting path before these names existed). Recommend fixing it as part of the Decimal/UUID feature implementation; no tracking issue yet. (raised by xiaocai2333, czs007)

czs007 (approved):
/lgtm

Issue #52166, opened 2026-08-04
[Bug]: Delegator may miss forwarding deletes to a newly loaded sealed segment


sijie-ni-0214:
## Environment

- **Milvus Version**: master at `4d7d00e7f5`
- **Deployment Mode**: streaming mode, standalone or cluster
- **MQ**: streaming WAL backend independent
- **SDK**: N/A
- **OS**: N/A

### Option B: Steps (timing issue)

1. Start loading a sealed segment through the shard delegator.
2. Let `loadStreamDelete` finish replaying buffered deletes and enter its final `deleteMut.RLock` barrier.
3. In that barrier, call `AddDistributions`. The segment is added to `sealedSegments`, but the current implementation only notifies the asynchronous snapshot loop.
4. Release the barrier before the snapshot loop publishes a new `current` snapshot.
5. Deliver another streaming delete during this window.
6. `ProcessDelete` calls `PinOnlineSegments`, reads the old snapshot, and does not forward the delete to the newly loaded sealed segment.

### Trigger Conditions

- **Frequency**: intermittent and timing-dependent
- **First observed after**: a sealed segment finishes delete replay while snapshot publication is delayed
- **Does NOT happen when**: `AddDistributions` publishes the updated snapshot synchronously before releasing the final delete barrier

## Expected Behavior

Once a sealed segment has completed delete replay and the final load barrier is released, every subsequent streaming delete should observe that segment in the delegator distribution and be forwarded to it.

## Actual Behavior

There is a visibility gap between updating the live distribution map and asynchronously publishing the next snapshot. A delete arriving in this gap pins the old snapshot and can skip the newly loaded sealed segment. The missed delete is not replayed again before the segment becomes readable.

## Error Logs

No error is emitted. This is a silent consistency issue that can leave a loaded sealed segment without a delete that arrived after its load-time replay.

## Non-default Configuration

None.

## Analysis Hints (Optional)

- Final load/delete barrier: `internal/querynodev2/delegator/delegator_data.go`, `loadStreamDelete`
- Snapshot publication: `internal/querynodev2/delegator/distribution.go`, `AddDistributions`
- Streaming delete routing: `internal/querynodev2/delegator/delta_forward.go`, `PinOnlineSegments`
- Related recovery optimization issue: #51107
- The fix should preserve the atomic contract between "all prior deletes applied" and "segment visible to subsequent delete forwarding".

PR #51108, opened 2026-07-07
enhance: Defer delegator snapshot generation during recovery

sijie-ni-0214:
issue: #51107

## What this PR does

This PR optimizes QueryNode delegator recovery by deferring the first public snapshot generation while restored segment distribution is still being rebuilt before QueryCoord syncs the target view.

During pre-sync recovery, restored distributions stay in the live distribution maps and online segment routing reads those maps under lock. This avoids publishing a new snapshot for every restored segment while still allowing delete forward to see restored segments. Once QueryCoord syncs the target view, snapshot generation remains synchronous where delete forwarding and the IDF oracle depend on the newly generated snapshot.

## Commit breakdown

- `786cc5a3f0 fix: sync delegator snapshot for delete forwarding`
  - Restores synchronous snapshot generation on target-version sync so delete forward and IDF oracle state observe the current distribution immediately.

- `14e9a75547 enhance: defer pre-sync recovery snapshots`
  - Defers public snapshot generation during pre-sync recovery before QueryCoord syncs the query view.
  - Lets online segment reads use live distribution maps under lock during that window, preserving delete forward correctness without repeatedly publishing snapshots.

- `b27d19ff8e enhance: reduce production snapshot memory usage`
  - Uses `SealedSegmentRowCount` as the authoritative sealed target source inside QueryNode.
  - Preallocates per-node snapshot grouping to reduce allocations and memory overhead while building production snapshots.

## Test Plan

- [x] `git diff --check`
- [x] `gofumpt -l internal/querynodev2/delegator/distribution.go internal/querynodev2/delegator/distribution_test.go internal/querynodev2/delegator/delegator_data_test.go internal/querynodev2/delegator/delta_forward_test.go internal/querynodev2/services_test.go`
- [x] `go test -v -count=1 -tags dynamic,test -gcflags=\"all=-N -l\" ./internal/querynodev2/delegator -run \"Test(AddDistributionsPublishesSnapshotSynchronously|DistributionPreSyncDefersRecoverySnapshot)$\"`
- [x] `go test -v -count=1 -tags dynamic,test -gcflags=\"all=-N -l\" ./internal/querynodev2/delegator -run \"TestStreamingForward/(TestBFStreamingForward|TestDirectStreamingForward)$|TestDelegatorDataSuite/TestSyncTargetVersion$|TestDistributionSuite/TestSyncTargetVersion_RedundantGrowingLogic$\"`
[...]

PR #48824, opened 2026-04-08
enhance: optimize load performance for high concurrency

sunby:
issue: https://github.com/milvus-io/milvus/issues/48823
1. Async snapshot generation in distribution: move genSnapshot to a background goroutine, reducing lock hold time in AddDistributions, AddGrowing, MarkOfflineSegments, and RemoveDistributions.
2. Batch partition recovery in CollectionManager.Recover to reduce meta store write calls.
3. Remove embedded StreamingNode check in resource_manager.
4. Skip partition lookup in executor.getMetaInfo for loadMeta.
5. Fix proto.Clone scope in cluster.LoadSegments.
6. Downgrade schemaChangeMutex to RLock in addDistributionIfVersionOK.
7. Fix delegator Close order: stop snapshot loop before refunding.

PR #52167, opened 2026-08-04
fix: sync delegator snapshot for delete forwarding

sijie-ni-0214:
issue: #52166

## What this PR does

This PR restores synchronous publication of delegator distribution snapshots after sealed segment add, remove, and offline updates.

It closes the visibility gap where a sealed segment had completed load-time delete replay and was present in the live distribution map, but a subsequent streaming delete could still pin the previous snapshot and skip that segment before the asynchronous snapshot loop published the update.

The change also removes the background snapshot loop and test-only flush path, and serializes distribution shutdown with late sealed candidate additions so Bloom filter resource accounting is refunded correctly.

## Why

`loadStreamDelete` publishes a newly loaded sealed segment while holding the final `deleteMut.RLock` barrier. Snapshot publication must complete before that barrier is released so the next `ProcessDelete` call observes the segment through `PinOnlineSegments`.

Updating only the live distribution map and notifying an asynchronous snapshot worker breaks that atomic contract.

## Test Plan

- [x] `git diff --check upstream/master...HEAD`
- [x] `make static-check`
- [x] `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" ./internal/querynodev2/delegator -run 'Test(AddDistributionsPublishesSnapshotSynchronously|Distribution_AddDistributionsAfterClose|Distribution_CloseRefundsExistingCandidate)$|TestStreamingForward/(TestBFStreamingForward|TestDirectStreamingForward)$|TestDistributionSuite/(TestRemoveDistribution|TestMarkOfflineSegments)$'`

### DIFF
--- a/internal/querynodev2/delegator/delegator.go
+++ b/internal/querynodev2/delegator/delegator.go
@@ -1454,11 +1454,7 @@ func (sd *shardDelegator) Close() {
 		sd.growingSourceProvider.Deactivate()
 	}
 
-	// Stop background snapshot loop before refunding candidates
 	sd.distribution.Close()
-
-	// Refund all sealed segment candidates in distribution
-	sd.distribution.RefundAllCandidates()
 
 	// clean idf oracle
 	if idfOracle := sd.getIDFOracle(); idfOracle != nil {
--- a/internal/querynodev2/delegator/delegator_data_test.go
+++ b/internal/querynodev2/delegator/delegator_data_test.go
@@ -575,7 +575,6 @@ func (s *DelegatorDataSuite) TestProcessDelete() {
 		},
 	}, 10)
 
-	s.delegator.distribution.Flush()
 	s.False(s.delegator.distribution.Serviceable())
 
 	worker1.EXPECT().LoadSegments(mock.Anything, mock.AnythingOfType("*querypb.LoadSegmentsRequest")).
@@ -598,7 +597,6 @@ func (s *DelegatorDataSuite) TestProcessDelete() {
 		Version: time.Now().UnixNano(),
 	})
 	s.Require().NoError(err)
-	s.delegator.distribution.Flush()
 	s.True(s.delegator.distribution.Serviceable())
 	// Test normal errors with retry and fail
 	worker1.ExpectedCalls = nil
@@ -611,7 +609,6 @@ func (s *DelegatorDataSuite) TestProcessDelete() {
 			RowCount:    1,
 		},
 	}, 10)
-	s.delegator.distribution.Flush()
 	s.False(s.delegator.distribution.Serviceable(), "should retry and failed")
 
 	// refresh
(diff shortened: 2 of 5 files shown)


czs007:

This PR correctly moves snapshot publication back under `d.mut` so a newly added segment is visible before the delete-forwarding barrier releases; the two notes below concern the performance and lock-hygiene cost of that change.

## Low

- **`internal/querynodev2/delegator/distribution.go:403`** — `AddDistributions` now calls `d.genSnapshot()` + `d.updateServiceable("AddDistributions")` synchronously inside `d.mut.Lock()`, and this path is reached from `delegator_data.go:1143` / `:1179` while `sd.deleteMut.RLock()` is held. This has two costs: (a) the drained-fast path, which previously did only map writes plus a non-blocking `notifySnapshotUpdate()`, now performs an O(N) walk over all sealed + growing segments with fresh `nodeSegments`/`dist`/`growing` allocations (`distribution.go:597-622`) inside the `deleteMut` barrier; (b) the removed capacity-1 `snapshotNotifier` used to coalesce concurrent rebuilds, so total snapshot work goes from ~O(N) to O(requests × N), all under the `d.mut` write lock that search-side `PinReadableSegments`/`PinOnlineSegments` contend for at `distribution.go:241`. Failure scenario: a channel with thousands of sealed segments under a balance/recovery load storm with concurrent delete traffic → each load request adds an O(N) snapshot build inside the `deleteMut.RLock` barrier, repeatedly stalling `ProcessDelete` (`delegator_data.go:284-285`, the sole writer advancing tsafe) and delaying search visibility, while the extra rebuilds serialize against search `Pin*` RLocks and raise latency jitter and GC pressure during load. This is bounded (pure in-memory; `leaderViewUpdatedCallback` → `markLeaderViewUpdated` at `services.go:1856-1862` only marks a map, so no re-entrancy) and is inherent to publishing before unlock, so it is a trade-off rather than a defect — but note that it partially reverts the load-path gains from #48824. Suggestion: state the trade-off explicitly in the PR description and include measured numbers for a large channel (thousands of sealed segments) under a balance/recovery load storm; if the regression is material, consider restoring coalescing for the non-barrier callers while keeping the synchronous publish on the barrier path. *(raised by tinswzy)*

## Nitpick

- **`internal/querynodev2/delegator/distribution.go:571`** — The `d.mut.Unlock()` that previously preceded the remove log was moved after it, so `mlog.Info("remove segments from distribution", ...)` (`:571-576`) now runs inside the write lock held from `:549` to `:582`. The same new pattern appears in `MarkOfflineSegments`, where `:433-434` takes `d.mut.Lock()` with a deferred unlock and the offline log at `:450-452` executes under it. Since `d.mut` is the same lock search-side `PinReadableSegments`/`PinOnlineSegments` acquire for reading (`:241`), this needlessly widens the mutual-exclusion window. Suggestion: capture the log fields while holding the lock and move both `mlog.Info` calls after `Unlock()`. Impact is small (zap writes are typically buffered) — code hygiene only. *(raised by tinswzy)*

sijie-ni-0214:
Thanks for the review. The synchronous rebuild is an intentional correctness trade-off, and #51108 will address the recovery-path overhead. The logging point is valid, but we will keep this PR focused and leave the code unchanged.

czs007 (approved):
/lgtm

PR #51360, opened 2026-07-14
enhance: enforce function-field binding principle for function DDL

MrPresent-Han:
## What

Enforce a function–field binding principle for function DDL (add / drop / alter function on an existing collection), so a function and its output field stay coupled and already-stored vectors can never be silently invalidated by DDL.

BM25 and MinHash follow the strict binding: add/drop only via the unified `add_function_field` / `drop_function_field`, output field always coupled. **TextEmbedding is a temporary carve-out** — its legacy `AddCollectionFunction` / `DropCollectionFunction` paths are retained, so attach-over-existing-field and detach are still possible for it, because `add_function_field` embedding backfill is not yet implemented. It will be folded into the unified path in a follow-up. So the "always coupled" invariant currently holds for BM25/MinHash, not TextEmbedding.

## Changes

- **AlterCollectionSchema invariant**: reject standalone add-function (a function must be added together with its new output field), reject detaching a function without dropping its output field, and reject function cascade (a new function's input being another function's output).
- **Legacy RPCs**: `AddCollectionFunction` / `DropCollectionFunction` are rejected for BM25/MinHash (must use `add_function_field` / `drop_function_field`) and retained only for TextEmbedding (see carve-out). `AlterCollectionFunction` is kept and gains the whitelist below. In the alter and legacy-add paths, function field IDs are always re-derived from field names (never trusted from the request), so a request cannot inject an unrelated field ID that a later `drop_function_field` would delete.
- **alter_function whitelist**: only connection/runtime params may change (TextEmbedding: `url`, `credential`, `timeout_ms`, `max_client_batch_size`, `region`, `location`, `projectid`, `user`); BM25/MinHash have no alterable params. Function identity (type, name, input/output fields) and output-shaping params (`dim`, `model_name`, `endpoint` — the TEI model identity, `normalize`, `truncate*`, prompts, ...) are immutable. Params are normalized (keys lowercased, duplicate keys rejected) so a crafted duplicate key cannot bypass the diff.
- **drop_function_field**: allow any output-producing function (dropping needs no backfill), removing the previous BM25/MinHash-only restriction.

## Not in scope / follow-up

- Extending `add_function_field` to TextEmbedding (post-creation add would require backfilling every existing row through the external embedding model) — deferred; folding the TextEmbedding legacy paths into the unified binding follows this.
- MinHash input-field analyzer immutability lives on a different DDL path (`AlterCollectionField`); tracked as a separate follow-up.

## Breaking changes

[...]

PR #52170, opened 2026-08-04
test: add REST ordering, aggregation, and schema evolution coverage

lyyyuna:
## Summary

- Add REST v2 query and search coverage for `orderByFields`, query aggregation, and `searchAggregation`.
- Add REST v2 schema-evolution coverage for drop field and atomic add/drop function-field operations.
- Extend the REST test client with `add_function_field` and `drop_function_field` helpers.
- Keep schema-evolution cases focused on REST request parameters and resulting metadata, without duplicating segment lifecycle or multi-replica coverage.

## Coverage

- Query/search ordering with filters, pagination, nullable fields, deterministic member assertions, and deliberately distinct ANN/scalar/PK ordering.
- Query/search aggregation with exact metrics and hit-score oracles, nullable group keys, composite/nested groups, `searchSize`, `topHits`, and multi-query ownership.
- Drop field invalid identifier combinations, protected or unknown fields, repeated drops, and complete before/after schema-index snapshots.
- Add function field required objects, field-name binding, default AutoIndex behavior, output-field/function/index metadata, and rejected-request atomicity.
- Drop function field selective cascade across two bindings, invalid or repeated drops, and field dependency validation.

Related changes: #52071, #51360, #51411

## Test Plan

- `python3 -m pytest testcases/test_query_order_by.py testcases/test_search_order_by.py testcases/test_query_aggregation.py testcases/test_search_aggregation.py testcases/test_drop_field_operations.py testcases/test_function_field_operations.py --endpoint http://127.0.0.1:19530 --token root:Milvus --minio_host 127.0.0.1 -q -s`
  - 66 passed
- `python3 -m pytest milvus_client/test_milvus_client_search_aggregation.py --host 127.0.0.1 --port 19530 --user root --password Milvus --token root:Milvus -q -s`
  - 51 passed, 1 existing xfailed
- `ruff check` on all seven changed test files
- `ruff format --check` on all seven changed test files
- `git diff --check`

### DIFF
--- a/tests/python_client/milvus_client/test_milvus_client_search_aggregation.py
+++ b/tests/python_client/milvus_client/test_milvus_client_search_aggregation.py
@@ -161,6 +161,61 @@ class TestSearchAggregation(TestMilvusClientV2Base):
         prices = [hit.fields[self.price_field] for hit in hits]
         assert prices == sorted(prices)
 
+    @staticmethod
+    def _l2_score(query_vector, vector):
+        return math.fsum((left - right) ** 2 for left, right in zip(query_vector, vector))
+
+    def _expected_brand_aggregation(self, query_vector, bucket_size, top_hit_size):
+        ranked_rows = []
+        for row in self._rows():
+            vector = row[self.vector_field]
+            brand = row[self.brand_field]
+            if vector is None or brand is None:
+                continue
+            ranked_rows.append(
+                (
+                    self._l2_score(query_vector, vector),
+                    row[self.primary_field],
+                    brand,
+                )
+            )
+        ranked_rows.sort()
+
+        expected_groups = {}
+        for score, primary_key, brand in ranked_rows:
+            if brand not in expected_groups:
+                if len(expected_groups) >= bucket_size:
+                    continue
+                expected_groups[brand] = []
+            if len(expected_groups[brand]) < top_hit_size:
+                expected_groups[brand].append((primary_key, score))
+            if len(expected_groups) == bucket_size and all(
+                len(hits) == top_hit_size for hits in expected_groups.values()
+            ):
+                break
+
+        assert len(expected_groups) == bucket_size
+        assert all(len(hits) == top_hit_size for hits in expected_groups.values())
+        return expected_groups
+
+    def _normalize_brand_aggregation(self, buckets):
+        return tuple(
+            (
+                self._key_value(bucket, self.brand_field),
+                bucket.count,
+                bucket.metrics["doc_count"],
+                tuple(
+                    (
+                        hit.pk,
+                        round(float(hit.score), 6),
+                        tuple(sorted(hit.fields.items())),
+                    )
+                    for hit in bucket.hits
+                ),
+            )
+            for bucket in buckets
+        )
+
     @pytest.mark.tags(CaseLabel.L1)
     def test_search_aggregation_minimal_parameters(self):
         """
@@ -189,6 +244,70 @@ class TestSearchAggregation(TestMilvusClientV2Base):
             assert len(buckets) == 1
             assert [entry["field_name"] for entry in buckets[0].key] == [self.brand_field]
             assert buckets[0].count >= 1
+
+    @pytest.mark.tags(CaseLabel.L1)
+    @pytest.mark.parametrize(
+        "bucket_size",
+        [2, 4],
+        ids=["two-buckets", "all-four-buckets"],
+    )
+    def test_search_aggregation_limit_and_size_combinations(self, bucket_size):
+        """
+        target: verify ordinary search limit does not change exact search_aggregation buckets or topHits
+        method: compare limit smaller than, equal to, and greater than one bucket size against an independent FLAT/L2 oracle
+        expected: every limit returns identical exact bucket keys, counts, hit IDs, fields, and L2 scores
+        """
+        top_hit_size = 2
+        query_vector = self._query_vectors()[0]
+        limits = [1, bucket_size, bucket_size + 1]
+        expected_groups = self._expected_brand_aggregation(query_vector, bucket_size, top_hit_size)
+        client = self._client(alias=self.shared_alias)
+        normalized_by_limit = {}
+
+        for limit in limits:
+            res, _ = self.search(
+                client,
+                self.collection_name,
+                data=[query_vector],
+                anns_field=self.vector_field,
+                search_params=self._search_params(),
+                limit=limit,
+                filter=self._non_null_expr(self.brand_field),
+                output_fields=[self.brand_field],
+                search_aggregation=SearchAggregation(
+                    fields=[self.brand_field],
+                    size=bucket_size,
+                    metrics={"doc_count": {"count": "*"}},
+                    order=[{"_key": "asc"}],
+                    top_hits=TopHits(size=top_hit_size, sort=[{"_score": "asc"}]),
+                ),
+            )
+
+            assert len(res.agg_buckets) == 1
+            buckets = res.agg_buckets[0]
+            assert len(buckets) == bucket_size
+            keys = [self._key_value(bucket, self.brand_field) for bucket in buckets]
+            assert keys == sorted(expected_groups)
+
+            for bucket in buckets:
+                brand = self._key_value(bucket, self.brand_field)
+                expected_hits = expected_groups[brand]
+                assert bucket.count == top_hit_size
+                assert bucket.metrics["doc_count"] == top_hit_size
+                assert len(bucket.hits) == top_hit_size
+                assert [hit.pk for hit in bucket.hits] == [primary_key for primary_key, _ in expected_hits]
+                assert [hit.score for hit in bucket.hits] == pytest.approx(
+                    [score for _, score in expected_hits],
+                    rel=1e-5,
+                    abs=1e-6,
+                )
+                assert [hit.fields for hit in bucket.hits] == [{self.brand_field: brand}] * top_hit_size
+                self._assert_scores_ascending(bucket.hits)
+
+            normalized_by_limit[limit] = self._normalize_brand_aggregation(buckets)
+
+        baseline = normalized_by_limit[limits[0]]
+        assert all(normalized_by_limit[limit] == baseline for limit in limits[1:])
 
     @pytest.mark.tags(CaseLabel.L1)
     def test_search_aggregation_range_search(self):
--- a/tests/python_client/milvus_client/test_milvus_client_search_aggregation.py
+++ b/tests/python_client/milvus_client/test_milvus_client_search_aggregation.py
@@ -277,7 +277,13 @@ class TestSearchAggregation(TestMilvusClientV2Base):
                 search_aggregation=SearchAggregation(
                     fields=[self.brand_field],
                     size=bucket_size,
-                    metrics={"doc_count": {"count": "*"}},
+                    metrics={
+                        "doc_count": {"count": "*"},
+                        "score_sum": {"sum": "_score"},
+                        "score_avg": {"avg": "_score"},
+                        "score_min": {"min": "_score"},
+                        "score_max": {"max": "_score"},
+                    },
                     order=[{"_key": "asc"}],
                     top_hits=TopHits(size=top_hit_size, sort=[{"_score": "asc"}]),
                 ),
@@ -303,6 +309,11 @@ class TestSearchAggregation(TestMilvusClientV2Base):
                 )
                 assert [hit.fields for hit in bucket.hits] == [{self.brand_field: brand}] * top_hit_size
                 self._assert_scores_ascending(bucket.hits)
+                expected_scores = [score for _, score in expected_hits]
+                assert bucket.metrics["score_sum"] == pytest.approx(sum(expected_scores))
+                assert bucket.metrics["score_avg"] == pytest.approx(sum(expected_scores) / len(expected_scores))
+                assert bucket.metrics["score_min"] == pytest.approx(min(expected_scores))
+                assert bucket.metrics["score_max"] == pytest.approx(max(expected_scores))
 
             normalized_by_limit[limit] = self._normalize_brand_aggregation(buckets)
 
--- a/tests/restful_client_v2/testcases/test_drop_field_operations.py
+++ b/tests/restful_client_v2/testcases/test_drop_field_operations.py
@@ -35,6 +35,53 @@ class TestCollectionDropField(TestBase):
         assert rsp["code"] == 0, rsp
         return collection_name
 
+    def _describe_index_state(self, collection_name, index_name):
+        rsp = self.index_client.index_describe(collection_name=collection_name, index_name=index_name)
+        assert rsp["code"] == 0, rsp
+        assert len(rsp["data"]) == 1, rsp
+        index = rsp["data"][0]
+        assert "fail" not in index["indexState"].lower(), rsp
+        assert index["failReason"] == "", rsp
+        index_params = [(param["key"], param["value"]) for param in index["indexParams"]]
+        index_param_keys = [key for key, _ in index_params]
+        assert len(index_param_keys) == len(set(index_param_keys)), index
+        return {
+            "fieldName": index["fieldName"],
+            "indexName": index["indexName"],
+            "metricType": index["metricType"],
+            "indexType": index["indexType"],
+            "indexParams": sorted(index_params),
+        }
+
+    def _assert_collection_state(self, collection_name, *, expected_fields, expected_indexes):
+        desc = self.collection_client.collection_describe(collection_name)
+        assert desc["code"] == 0, desc
+        raw_fields = desc["data"]["fields"]
+        field_names = [field["name"] for field in raw_fields]
+        assert field_names == list(expected_fields), desc
+        assert len(field_names) == len(set(field_names)), desc
+        fields = {field["name"]: field for field in raw_fields}
+
+        raw_functions = desc["data"].get("functions", [])
+        function_names = [function["name"] for function in raw_functions]
+        assert len(function_names) == len(set(function_names)), desc
+        functions = {function["name"]: function for function in raw_functions}
+        assert set(fields) == set(expected_fields), desc
+        assert functions == {}, desc
+
+        indexes = self.index_client.index_list(collection_name=collection_name)
+        assert indexes["code"] == 0, indexes
+        assert len(indexes["data"]) == len(set(indexes["data"])), indexes
+        assert set(indexes["data"]) == set(expected_indexes), indexes
+        index_metadata = {
+            index_name: self._describe_index_state(collection_name, index_name) for index_name in indexes["data"]
+        }
+        return {
+            "fields": fields,
+            "functions": functions,
+            "indexes": index_metadata,
+        }
+
     @pytest.mark.parametrize(
         "field_name,field_id,expected_message",
         [
@@ -51,8 +98,9 @@ class TestCollectionDropField(TestBase):
         method: omit both identifiers, provide both, or provide a non-positive fieldId
         expected: REST rejects each request before attempting a schema mutation
         """
+        missing_collection_name = gen_collection_name(prefix=f"{self.__class__.__name__}Missing")
         rsp = self.collection_client.drop_field(
-            "drop_field_parameter_validation",
+            missing_collection_name,
             field_name=field_name,
             field_id=field_id,
         )
@@ -76,13 +124,22 @@ class TestCollectionDropField(TestBase):
         """
         collection_name = self._create_collection()
 
+        before_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "tag", "dense"],
+            expected_indexes=["dense_idx"],
+        )
+
         rsp = self.collection_client.drop_field(collection_name, field_name=field_name)
         assert rsp["code"] == 1100, rsp
         assert expected_message in rsp["message"], rsp
 
-        desc = self.collection_client.collection_describe(collection_name)
-        assert desc["code"] == 0, desc
-        assert {field["name"] for field in desc["data"]["fields"]} == {"id", "tag", "dense"}
+        after_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "tag", "dense"],
+            expected_indexes=["dense_idx"],
+        )
+        assert after_state == before_state
 
     def test_drop_field_second_request_is_rejected(self):
         """
@@ -92,8 +149,32 @@ class TestCollectionDropField(TestBase):
         """
         collection_name = self._create_collection()
 
+        before_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "tag", "dense"],
+            expected_indexes=["dense_idx"],
+        )
+
         first = self.collection_client.drop_field(collection_name, field_name="tag")
         assert first["code"] == 0, first
+        after_first_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "dense"],
+            expected_indexes=["dense_idx"],
+        )
+        expected_after_first_state = {
+            "fields": {name: field for name, field in before_state["fields"].items() if name != "tag"},
+            "functions": before_state["functions"],
+            "indexes": before_state["indexes"],
+        }
+        assert after_first_state == expected_after_first_state
+
         second = self.collection_client.drop_field(collection_name, field_name="tag")
         assert second["code"] == 1100, second
         assert "field not found" in second["message"], second
+        after_second_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "dense"],
+            expected_indexes=["dense_idx"],
+        )
+        assert after_second_state == after_first_state
--- a/tests/restful_client_v2/testcases/test_drop_field_operations.py
+++ b/tests/restful_client_v2/testcases/test_drop_field_operations.py
@@ -1,0 +1,99 @@
+import pytest
+from base.testbase import TestBase
+from utils.constant import CaseLabel
+from utils.utils import gen_collection_name
+
+
+@pytest.mark.tags(CaseLabel.L1)
+class TestCollectionDropField(TestBase):
+    """Supplemental REST v2 parameter coverage for collection field drops."""
+
+    def _create_collection(self):
+        collection_name = gen_collection_name(prefix=self.__class__.__name__)
+        rsp = self.collection_client.collection_create(
+            {
+                "collectionName": collection_name,
+                "schema": {
+                    "autoId": False,
+                    "enableDynamicField": False,
+                    "fields": [
+                        {"fieldName": "id", "dataType": "Int64", "isPrimary": True},
+                        {"fieldName": "tag", "dataType": "VarChar", "elementTypeParams": {"max_length": "64"}},
+                        {"fieldName": "dense", "dataType": "FloatVector", "elementTypeParams": {"dim": "4"}},
+                    ],
+                },
+                "indexParams": [
+                    {
+                        "fieldName": "dense",
+                        "indexName": "dense_idx",
+                        "indexType": "AUTOINDEX",
+                        "metricType": "L2",
+                    }
+                ],
+            }
+        )
+        assert rsp["code"] == 0, rsp
+        return collection_name
+
+    @pytest.mark.parametrize(
+        "field_name,field_id,expected_message",
+        [
+            (None, None, "exactly one of fieldName or fieldId is required"),
+            ("tag", 101, "exactly one of fieldName or fieldId is required"),
+            (None, 0, "fieldId must be greater than 0"),
+            (None, -1, "fieldId must be greater than 0"),
+        ],
+        ids=["missing-identifier", "both-identifiers", "zero-field-id", "negative-field-id"],
+    )
+    def test_drop_field_identifier_parameter_validation(self, field_name, field_id, expected_message):
+        """
+        target: verify REST drop field requires one valid identifier form
+        method: omit both identifiers, provide both, or provide a non-positive fieldId
+        expected: REST rejects each request before attempting a schema mutation
+        """
+        rsp = self.collection_client.drop_field(
+            "drop_field_parameter_validation",
+            field_name=field_name,
+            field_id=field_id,
+        )
+        assert rsp["code"] == 1100, rsp
+        assert expected_message in rsp["message"], rsp
+
+    @pytest.mark.parametrize(
+        "field_name,expected_message",
+        [
+            ("id", "cannot drop primary key field"),
+            ("dense", "cannot drop the last vector field"),
+            ("missing_field", "field not found"),
+        ],
+        ids=["primary-key", "last-vector", "unknown-field"],
+    )
+    def test_drop_field_rejects_protected_or_unknown_field(self, field_name, expected_message):
+        """
+        target: verify REST surfaces server-side field drop validation
+        method: request a primary-key, last-vector, or unknown field drop by name
+        expected: each request fails without removing any schema field
+        """
+        collection_name = self._create_collection()
+
+        rsp = self.collection_client.drop_field(collection_name, field_name=field_name)
+        assert rsp["code"] == 1100, rsp
+        assert expected_message in rsp["message"], rsp
+
+        desc = self.collection_client.collection_describe(collection_name)
+        assert desc["code"] == 0, desc
+        assert {field["name"] for field in desc["data"]["fields"]} == {"id", "tag", "dense"}
+
+    def test_drop_field_second_request_is_rejected(self):
+        """
+        target: verify REST drop field reports a field that was already removed
+        method: drop the same scalar field twice by name
+        expected: the first request succeeds and the second reports field not found
+        """
+        collection_name = self._create_collection()
+
+        first = self.collection_client.drop_field(collection_name, field_name="tag")
+        assert first["code"] == 0, first
+        second = self.collection_client.drop_field(collection_name, field_name="tag")
+        assert second["code"] == 1100, second
+        assert "field not found" in second["message"], second
--- a/tests/restful_client_v2/testcases/test_drop_field_operations.py
+++ b/tests/restful_client_v2/testcases/test_drop_field_operations.py
@@ -1,3 +1,5 @@
+import time
+
 import pytest
 from base.testbase import TestBase
 from utils.constant import CaseLabel
@@ -18,8 +20,8 @@ class TestCollectionDropField(TestBase):
                     "enableDynamicField": False,
                     "fields": [
                         {"fieldName": "id", "dataType": "Int64", "isPrimary": True},
-                        {"fieldName": "tag", "dataType": "VarChar", "elementTypeParams": {"max_length": "64"}},
                         {"fieldName": "dense", "dataType": "FloatVector", "elementTypeParams": {"dim": "4"}},
+                        {"fieldName": "tag", "dataType": "VarChar", "elementTypeParams": {"max_length": "64"}},
                     ],
                 },
                 "indexParams": [
@@ -35,13 +37,21 @@ class TestCollectionDropField(TestBase):
         assert rsp["code"] == 0, rsp
         return collection_name
 
+    def _wait_index_ready(self, collection_name, index_name, timeout=30):
+        t0 = time.time()
+        while time.time() - t0 < timeout:
+            rsp = self.index_client.index_describe(collection_name=collection_name, index_name=index_name)
+            assert rsp["code"] == 0, rsp
+            assert len(rsp["data"]) == 1, rsp
+            index = rsp["data"][0]
+            if index["indexState"] == "Finished":
+                return index
+            time.sleep(1)
+        raise AssertionError(f"index {index_name} of collection {collection_name} not ready after {timeout}s")
+
     def _describe_index_state(self, collection_name, index_name):
-        rsp = self.index_client.index_describe(collection_name=collection_name, index_name=index_name)
-        assert rsp["code"] == 0, rsp
-        assert len(rsp["data"]) == 1, rsp
-        index = rsp["data"][0]
-        assert "fail" not in index["indexState"].lower(), rsp
-        assert index["failReason"] == "", rsp
+        index = self._wait_index_ready(collection_name, index_name)
+        assert index["failReason"] == "", index
         index_params = [(param["key"], param["value"]) for param in index["indexParams"]]
         index_param_keys = [key for key, _ in index_params]
         assert len(index_param_keys) == len(set(index_param_keys)), index
@@ -50,6 +60,7 @@ class TestCollectionDropField(TestBase):
             "indexName": index["indexName"],
             "metricType": index["metricType"],
             "indexType": index["indexType"],
+            "indexState": index["indexState"],
             "indexParams": sorted(index_params),
         }
 
@@ -58,9 +69,19 @@ class TestCollectionDropField(TestBase):
         assert desc["code"] == 0, desc
         raw_fields = desc["data"]["fields"]
         field_names = [field["name"] for field in raw_fields]
+        field_ids = [int(field["id"]) for field in raw_fields]
         assert field_names == list(expected_fields), desc
         assert len(field_names) == len(set(field_names)), desc
+        assert all(field_id > 0 for field_id in field_ids), desc
+        assert len(field_ids) == len(set(field_ids)), desc
         fields = {field["name"]: field for field in raw_fields}
+
+        raw_properties = desc["data"].get("properties", [])
+        property_keys = [prop["key"] for prop in raw_properties]
+        assert len(property_keys) == len(set(property_keys)), desc
+        properties = {prop["key"]: prop["value"] for prop in raw_properties}
+        assert "max_field_id" in properties, desc
+        assert int(properties["max_field_id"]) >= max(field_ids), desc
 
         raw_functions = desc["data"].get("functions", [])
         function_names = [function["name"] for function in raw_functions]
@@ -80,6 +101,7 @@ class TestCollectionDropField(TestBase):
             "fields": fields,
             "functions": functions,
             "indexes": index_metadata,
+            "properties": properties,
         }
 
     @pytest.mark.parametrize(
@@ -98,14 +120,25 @@ class TestCollectionDropField(TestBase):
         method: omit both identifiers, provide both, or provide a non-positive fieldId
         expected: REST rejects each request before attempting a schema mutation
         """
-        missing_collection_name = gen_collection_name(prefix=f"{self.__class__.__name__}Missing")
+        collection_name = self._create_collection()
+        before_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "dense", "tag"],
+            expected_indexes=["dense_idx"],
+        )
         rsp = self.collection_client.drop_field(
-            missing_collection_name,
+            collection_name,
             field_name=field_name,
             field_id=field_id,
         )
         assert rsp["code"] == 1100, rsp
         assert expected_message in rsp["message"], rsp
+        after_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "dense", "tag"],
+            expected_indexes=["dense_idx"],
+        )
+        assert after_state == before_state
 
     @pytest.mark.parametrize(
         "field_name,expected_message",
@@ -126,7 +159,7 @@ class TestCollectionDropField(TestBase):
 
         before_state = self._assert_collection_state(
             collection_name,
-            expected_fields=["id", "tag", "dense"],
+            expected_fields=["id", "dense", "tag"],
             expected_indexes=["dense_idx"],
         )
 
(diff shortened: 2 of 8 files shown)


tests/restful_client_v2/testcases/test_search_aggregation.py
```diff
@@ -0,0 +1,268 @@
...
+        target: verify all supported metric operations with one grouping field
+        method: retain all bucket rows with topHits, then group by category and calculate count/sum/avg/min/max
+        expected: every bucket key, count, and metric matches the inserted rows
+        """
+        rsp = self.vector_client.vector_search(
+            {
+                "collectionName": self.collection_name,
+                "data": [[1.0, 0.0]],
+                "annsField": "vector",
+                "limit": len(ROWS),
```
yanliang567: All positive search-aggregation cases use `limit=len(ROWS)`, no selective filter, and derive the expected buckets from the entire fixture. An implementation that ignores the ANN candidates/query vector and aggregates the whole collection would therefore still pass. Please use a topK smaller than the collection size (or a selective filter), establish the exact candidate IDs, and derive the aggregates and `topHits` only from those candidates.

tests/restful_client_v2/testcases/test_query_aggregation.py
```diff
@@ -0,0 +1,303 @@
...
+    def test_query_group_by_single_field_count_sum(self):
+        """
+        target: verify REST query supports groupByFields with aggregation expressions
+        method: group by category and calculate count(*) and sum(score)
+        expected: every category returns the exact aggregate values
+        """
+        rows = self._query(
+            {
+                "collectionName": self.collection_name,
+                "filter": "id >= 0",
```
yanliang567: Every positive grouped-query case uses the tautological `id >= 0` filter and computes its ground truth from all rows. If the REST path drops the filter whenever `groupByFields` is present, this suite remains green. Please add a selective filter that changes group membership and aggregate values, and assert the resulting count/sum.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,276 @@
...
+        }
+        payload["indexParams"] = {
+            "fieldName": "embedding",
+            "indexName": "embedding_idx",
+            "metricType": "L2",
+            "indexType": "HNSW",
+            "params": {"M": 8, "efConstruction": 64},
+        }
+
+        rsp = self.collection_client.add_function_field(payload)
```
yanliang567, line 472: The expected behavior here is that the failed request creates no partial schema objects, but the test only checks the error response. The missing-request-part cases above have the same gap, and the name-mismatch case does not check for an orphan index. Please describe the collection and list indexes after each failure and assert that the complete field/function/index state is unchanged.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,276 @@
...
+        self._add_bm25_function_field(collection_name)
+
+        rsp = self.collection_client.drop_function_field(
+            {"collectionName": collection_name, "functionName": FUNCTION_NAME}
+        )
+        assert rsp["code"] == 0, rsp
+
+        desc = self.collection_client.collection_describe(collection_name)
+        assert desc["code"] == 0, desc
+        assert OUTPUT_FIELD_NAME not in {field["name"] for field in desc["data"]["fields"]}
```
yanliang567: These assertions prove that the requested BM25 objects disappeared, but not that the cascade preserved unrelated objects. A faulty implementation that also deletes `text`, `dense`, or `dense_idx` would pass. Please assert the exact surviving field/function/index sets after the drop.

tests/restful_client_v2/testcases/test_search_order_by.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+    def test_search_order_by_single_field(self, order_by, field_name, expected_values, expected_ids):
+        """
+        target: verify REST search supports top-level orderByFields in ascending and descending order
+        method: search all deterministic ANN candidates and sort them by one scalar field
+        expected: result values follow the requested scalar order
+        """
+        rows = self._search(orderByFields=order_by)
+        assert [row[field_name] for row in rows] == expected_values
+        assert [row["id"] for row in rows] == expected_ids
```
yanliang567, line 138: The fixture has deterministic IP distances, but the test only verifies reordered IDs/scalars. If entity columns are reordered while `distance` remains in the original ANN order, all current positive cases pass. Please assert the ID-to-distance association as well (`1 -> 1.0`, `2 -> 0.9`, `3 -> 0.8`, `4 -> 0.7`).

tests/restful_client_v2/testcases/test_search_order_by.py
```diff
@@ -0,0 +1,190 @@
...
+        ]
+
+    @pytest.mark.tags(CaseLabel.L1)
+    def test_search_order_by_with_offset(self):
+        """
+        target: verify REST search applies offset after top-level orderByFields
+        method: take the top three ANN candidates, sort by price ascending, skip one, and return two
+        expected: pagination returns ids 3 and 1 after scalar ordering
+        """
+        rows = self._search(limit=2, offset=1, orderByFields=["price:asc"])
```
yanliang567, line 189: This data does not distinguish candidate-scoped ordering from incorrectly sorting the whole collection: the ANN-excluded row (ID 4) also has the worst price, so both implementations return `[3, 1]`. Please give an ANN-excluded row the best scalar sort value and assert that it still cannot enter this page.

yanliang567 (changes requested):
Thanks for adding this REST coverage. I am requesting changes because several assertions still allow regressions in aggregation scoping and schema-operation atomicity to pass. Details are inline.

tests/restful_client_v2/testcases/test_search_order_by.py
lyyyuna: Fixed in fc4fdfb1da. Every successful search performed through the helper now validates the ID-to-distance association: 1->1.0, 2->0.9, 3->0.8, and 4->0.7, including filtered and paginated subsets.

tests/restful_client_v2/testcases/test_query_aggregation.py
lyyyuna: Fixed in fc4fdfb1da. The single-field grouped query now uses a selective category-and-ID filter that reduces the result to category_1/category_3 and IDs [20, 90). The expected grouped count and sum are computed from the same filtered fixture rows and asserted exactly.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Fixed in fc4fdfb1da. Added a shared exact base-schema assertion and call it after the missing-part, field-name mismatch, and unsupported-function-type failures. Each case now verifies fields are exactly id/text/dense, functions are empty, and the only index is dense_idx.

tests/restful_client_v2/testcases/test_search_order_by.py
lyyyuna: Fixed in fc4fdfb1da. ID 4 now has the best ascending price (5) while remaining outside the ANN candidate window in the offset case. Full-result expectations were updated, and the offset result remains IDs [3, 1], proving candidate-scoped sorting.

tests/restful_client_v2/testcases/test_search_aggregation.py
lyyyuna: Fixed in fc4fdfb1da. The positive search-aggregation cases now use a selective filter that establishes exact candidate IDs 1/2/3, while IDs 4/5 remain in the collection. Bucket keys, metrics, counts, and topHits are all derived from the filtered candidate rows. The legacy order-by compatibility case also uses an exact two-ID candidate set.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Fixed in fc4fdfb1da. The cascade-drop test now asserts the exact surviving state through the shared helper: fields id/text/dense, no functions, and only dense_idx.

tests/restful_client_v2/testcases/test_search_aggregation.py
```diff
@@ -0,0 +1,285 @@
...
+
+ROWS = [
+    {"id": 1, "category": "A", "brand": "X", "price": 30, "vector": [1.0, 0.0]},
+    {"id": 2, "category": "A", "brand": "Y", "price": 10, "vector": [0.9, 0.0]},
+    {"id": 3, "category": "B", "brand": "X", "price": 20, "vector": [0.8, 0.0]},
+    {"id": 4, "category": "C", "brand": "Y", "price": 40, "vector": [0.7, 0.0]},
+    {"id": 5, "category": "A", "brand": "X", "price": 5, "vector": [0.6, 0.0]},
+]
+SEARCH_LIMIT = 3
+SEARCH_FILTER = "id <= 3"
```
yanliang567: This selective filter still admits exactly the same number of rows as the ANN `limit`: IDs 1/2/3 and `limit=3`. It proves that the filter is respected, but an implementation that aggregates every filtered row while ignoring the query vector and ANN topK would still pass every positive case. Please let the filter admit more rows than the ANN limit and make at least one ranked-out row change a bucket, metric, or topHit if it is incorrectly included.

tests/restful_client_v2/testcases/test_search_aggregation.py
```diff
@@ -0,0 +1,285 @@
...
+                    "size": 3,
+                    "metrics": {
+                        "item_count": {"op": "count", "fieldName": "*"},
+                        "total_price": {"op": "sum", "fieldName": "price"},
+                        "average_price": {"op": "avg", "fieldName": "price"},
+                        "minimum_price": {"op": "min", "fieldName": "price"},
+                        "maximum_price": {"op": "max", "fieldName": "price"},
+                    },
+                    "order": [{"key": "_key", "direction": "asc"}],
+                    "topHits": {"size": SEARCH_LIMIT, "sort": [{"fieldName": "price", "direction": "asc"}]},
```
yanliang567: Could this use a `topHits.size` smaller than the cardinality of at least one bucket? Category A currently has two candidates and this requests size 3; the other configured sizes are also greater than or equal to their bucket sizes. A server that ignores `topHits.size` and returns every bucket row therefore passes. For example, request size 1 for category A, keep its aggregate count at 2, and assert the exact price-sorted winner.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        collection_name = self._create_collection()
+
+        rsp = self.collection_client.drop_field(collection_name, field_name=field_name)
+        assert rsp["code"] == 1100, rsp
+        assert expected_message in rsp["message"], rsp
+
+        desc = self.collection_client.collection_describe(collection_name)
+        assert desc["code"] == 0, desc
+        assert {field["name"] for field in desc["data"]["fields"]} == {"id", "tag", "dense"}
```
yanliang567: This verifies that the field names survived the rejected drop, but not the existing `dense_idx`. A non-atomic implementation could remove the vector index before rejecting the last-vector-field mutation and this test would remain green. Please snapshot/list the indexes and assert the exact index state is unchanged after each rejected request.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_drop_field_second_request_is_rejected(self):
+        """
+        target: verify REST drop field reports a field that was already removed
+        method: drop the same scalar field twice by name
+        expected: the first request succeeds and the second reports field not found
+        """
+        collection_name = self._create_collection()
+
+        first = self.collection_client.drop_field(collection_name, field_name="tag")
```
yanliang567, line 193: The first successful drop is only checked by response code, and the state after the rejected second request is not inspected. A code-0 no-op followed by a `field not found` response could pass without ever removing `tag`; unrelated partial changes during the second call would also be missed. Please assert the exact fields/indexes immediately after the first call and verify the same state after the second.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,278 @@
...
+        expected_message,
+    ):
+        """
+        target: verify drop_function_field validates the functionName parameter
+        method: send an empty name and a name absent from the collection schema
+        expected: REST returns binding or invalid-parameter errors respectively
+        """
+        collection_name = self._create_base_collection()
+
+        rsp = self.collection_client.drop_function_field(
```
yanliang567, line 550: These failed drop cases only assert the response and start from a collection with no function. That does not prove a bad function name cannot partially mutate unrelated fields/indexes or remove a valid function. Please seed a valid BM25 function for the unknown-name case and assert the exact field/function/index state is unchanged after every failed drop (the empty-name binding case can at least assert the base state).

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,278 @@
...
+
+        blocked = self.collection_client.drop_field(collection_name, field_name="text")
+        assert blocked["code"] == 1100, blocked
+        assert "referenced by function" in blocked["message"], blocked
+
+        dropped = self.collection_client.drop_function_field(
+            {"collectionName": collection_name, "functionName": FUNCTION_NAME}
+        )
+        assert dropped["code"] == 0, dropped
+        unblocked = self.collection_client.drop_field(collection_name, field_name="text")
```
yanliang567, line 613: A success code alone does not prove the now-unblocked `text` field was actually removed. A no-op implementation would pass this test. Please describe the collection and list indexes afterward, asserting fields are exactly `id` and `dense`, functions are empty, and only `dense_idx` remains.

tests/restful_client_v2/testcases/test_query_order_by.py
```diff
@@ -44,8 +44,8 @@ def _parse_order_by_field(order_by_field):
 def _expected_rows(order_by_fields, limit, row_filter=None, offset=0):
     """Build a deterministic reference order for local assertions.
 
-    The final ID key stabilizes this test helper only. Query ORDER BY does not
-    guarantee an implicit PK tie-breaker when all explicit sort fields match.
+    Query ORDER BY uses the primary key as a stable tie-breaker when all
```
yanliang567: This comment contradicts the current Query ORDER BY design: equal-key order is explicitly non-deterministic, and appending the primary key implicitly is listed as a future, not-implemented option. Please restore the previous test-helper wording and only assert exact tie ordering when the request explicitly includes a final key such as `id:asc`.

yanliang567 (commented):
The latest update fixes the earlier query-filter, search distance-association, offset-candidate, failed-add atomicity, and cascade-preservation findings. I found several remaining verification gaps around ANN aggregation scope, topHits truncation, and drop-operation state validation. Details are inline.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Updated in 69c5157396. The unknown-function case now seeds a valid BM25 function first and verifies the exact fields, function metadata, dense/sparse indexes, and sparse index parameters before and after the rejected drop. The empty-name binding case verifies the unchanged base schema and index state.

tests/restful_client_v2/testcases/test_search_aggregation.py
lyyyuna: Updated in 69c5157396. The nested aggregation case now retains two rows for the parent category through the child topHits window, while the parent requests topHits size 1. It asserts category A count=2 and the exact price-sorted parent winner, ID 2 with price 10, so an implementation returning every bucket row would fail.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Updated in 69c5157396. After the final successful text-field drop, the test now describes the collection and lists indexes, asserting fields are exactly id/dense, functions are empty, and dense_idx is the only remaining index.

tests/restful_client_v2/testcases/test_query_order_by.py
lyyyuna: Updated in 69c5157396. The helper comment now states that ID only stabilizes local expected data and is not an implicit API tie-breaker. The tied multi-field test no longer asserts exact IDs, while the offset test explicitly requests price:asc,id:asc before making exact page and repeatability assertions.

tests/restful_client_v2/testcases/test_search_aggregation.py
lyyyuna: Updated in 69c5157396. The REST scenario now lets the filter admit all five rows while using limit=1 and aggregation size=3/topHits=3. It asserts the exact A/B/C bucket keys, counts, all metric values, and price-sorted hits, making the observed limit-vs-size behavior explicit instead of relying on filter cardinality matching limit. I also added PyMilvus parameterized coverage for limit<size, limit=size, and limit>size; each case verifies that aggregation size controls the bucket count and that each bucket retains the requested two hits.

tests/restful_client_v2/testcases/test_drop_field_operations.py
lyyyuna: Updated in 69c5157396. The repeated Drop Field test now checks the initial state, verifies that the first successful call actually removes tag while preserving dense_idx, and verifies the identical state after the rejected second call.

tests/restful_client_v2/testcases/test_drop_field_operations.py
lyyyuna: Updated in 69c5157396. Each protected/unknown Drop Field case now verifies the exact field set and dense_idx before the request and again after the rejection, so index removal or any other partial mutation is detected.

tests/restful_client_v2/testcases/test_search_order_by.py
```diff
@@ -0,0 +1,199 @@
...
+    def _assert_id_distance_association(rows):
+        for row in rows:
+            assert row["id"] in EXPECTED_DISTANCE_BY_ID
+            assert row["distance"] == pytest.approx(EXPECTED_DISTANCE_BY_ID[row["id"]])
+
+    @pytest.mark.tags(CaseLabel.L0)
+    @pytest.mark.parametrize(
+        "order_by,field_name,expected_values,expected_ids",
+        [
+            (["price"], "price", [5, 10, 20, 30], [4, 2, 3, 1]),
```
yanliang567: Both parameter rows produce the same ID order `[4, 2, 3, 1]`: `price:asc` and `rating:desc`. An implementation that always sorts by price and ignores the requested field would therefore pass both the ID and scalar-value assertions. Please change the rating values so its descending order is distinct from price ascending and from the ANN order.

tests/restful_client_v2/testcases/test_search_order_by.py
```diff
@@ -0,0 +1,199 @@
...
+        assert all(set(row) == {"id", "category", "distance"} for row in rows)
+
+    @pytest.mark.tags(CaseLabel.L0)
+    def test_search_order_by_multi_fields(self):
+        """
+        target: verify REST search supports multiple top-level orderByFields
+        method: sort by category ascending and price descending
+        expected: secondary sorting is applied within the duplicate category
+        """
+        rows = self._search(orderByFields=["category:asc", "price:desc"])
```
yanliang567: The expected multi-field order `[1, 2, 3, 4]` is identical to the raw IP ANN order, and the ANN order inside category A already matches `price:desc`. Ignoring all `orderByFields`, or ignoring only the secondary key, still passes. Please make the secondary key reverse the ANN order within A, for example `price:asc` with expected IDs `[2, 1, 3, 4]`, or otherwise decouple vector rank from the scalar oracle.

yanliang567 (commented):
The latest update fixes the earlier findings. The inline comments below identify remaining false-green gaps around ORDER BY field selection, filter/candidate ordering, search-aggregation candidate retention, limit independence, and function-field cascade/atomicity.

Additional P2 follow-ups worth covering: a positive REST MinHash add/drop path; `isFunctionOutput` metadata; multi-NQ aggregation bucket ownership and hit distance binding; GROUP BY fields omitted from `outputFields`; protected/repeated Drop Field operations through positive `fieldId`; and PK-set/uniqueness checks in tied query-order results.

tests/restful_client_v2/testcases/test_search_order_by.py
```diff
@@ -0,0 +1,199 @@
...
+        assert [row["id"] for row in rows] == expected_ids
+
+    @pytest.mark.tags(CaseLabel.L0)
+    def test_search_order_by_with_filter(self):
+        """
+        target: verify REST search combines filter with top-level orderByFields
+        method: restrict results to category A and sort the matching rows by price ascending
+        expected: filtering happens before ordering and returns ids 2 and 1
+        """
+        rows = self._search(filter='category == "A"', orderByFields=["price:asc"])
```
yanliang567: The helper default `limit=4` covers the whole fixture, so this cannot distinguish filter-before-candidate-selection from an incorrect implementation that selects all four ANN candidates and filters afterward. Please use a smaller limit and put higher-scoring non-A rows ahead of the A rows; the filtered search should still return the requested number of A candidates before scalar ordering.

tests/restful_client_v2/testcases/test_search_aggregation.py
```diff
@@ -0,0 +1,282 @@
...
+                    "size": 3,
+                    "metrics": {
+                        "item_count": {"op": "count", "fieldName": "*"},
+                        "total_price": {"op": "sum", "fieldName": "price"},
+                        "average_price": {"op": "avg", "fieldName": "price"},
+                        "minimum_price": {"op": "min", "fieldName": "price"},
+                        "maximum_price": {"op": "max", "fieldName": "price"},
+                    },
+                    "order": [{"key": "_key", "direction": "asc"}],
+                    "topHits": {"size": 3, "sort": [{"fieldName": "price", "direction": "asc"}]},
```
yanliang567: Category A has exactly three rows and `topHits.size=3`; B and C have only one each. No row is ever ranked out of its group, so aggregating every filtered row or ignoring vector-score retention produces the same counts, metrics, and hits. Please add more rows to one group than the retained group size, give the extra row a deliberately worse IP score, and assert that it cannot affect the bucket count/metrics/topHits.

tests/python_client/milvus_client/test_milvus_client_search_aggregation.py
```diff
@@ -190,6 +190,60 @@ def test_search_aggregation_minimal_parameters(self):
...
+                size=bucket_size,
+                metrics={"doc_count": {"count": "*"}},
+                order=[{"_key": "asc"}],
+                top_hits=TopHits(size=top_hit_size, sort=[{"_score": "asc"}]),
+            ),
+        )
+
+        assert len(res.agg_buckets) == 1
+        buckets = res.agg_buckets[0]
+        assert len(buckets) == bucket_size
```
yanliang567: These assertions verify only result shape. Any sorted set of `bucket_size` brands with any two same-brand hits can pass, and separate parameterized invocations are never compared, so results may incorrectly change with `limit`. Please build an independent FLAT/L2 oracle for exact bucket keys, counts, hit IDs, and scores, then normalize and compare equal-`bucket_size` cases across `limit < size`, `limit == size`, and `limit > size`.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        collection_name = self._create_base_collection()
+        self._add_bm25_function_field(collection_name)
+
+        rsp = self.collection_client.drop_function_field(
+            {"collectionName": collection_name, "functionName": FUNCTION_NAME}
+        )
+        assert rsp["code"] == 0, rsp
+
+        self._assert_base_schema_state(collection_name)
```
yanliang567: This collection contains only one function. A faulty cascade that removes every function output field and index would produce the same base state and pass. Please seed two independently named function fields, drop only one by name, and assert that the other function, output field, and bound index remain exactly unchanged.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,326 @@
...
+        method: drop the same function field twice
+        expected: the second request reports that the function no longer exists
+        """
+        collection_name = self._create_base_collection()
+        self._add_bm25_function_field(collection_name)
+        payload = {"collectionName": collection_name, "functionName": FUNCTION_NAME}
+
+        first = self.collection_client.drop_function_field(payload)
+        assert first["code"] == 0, first
+        second = self.collection_client.drop_function_field(payload)
```
yanliang567, line 585: Neither call is followed by a schema/index-state assertion. A first call that reports success but leaves stale objects, or a rejected second call that performs a partial mutation, remains undetected. Please assert the exact base schema/index state immediately after the first successful drop and again after the rejected second request.

lyyyuna: Fixed in 07c10807e7. The test captures the original base state, asserts the exact base field/function/index state immediately after the successful first drop, then asserts the rejected second drop leaves that complete state unchanged.

tests/restful_client_v2/testcases/test_search_order_by.py
lyyyuna: Fixed in 07c10807e7. The filtered case requests limit=3 while higher-scoring B/C rows precede some A rows, and it asserts exactly three A hits ordered as [2,5,1]. This detects candidate selection before filtering.

tests/restful_client_v2/testcases/test_search_order_by.py
lyyyuna: Fixed in 07c10807e7. The multi-field case now uses category asc plus price asc and expects [2,5,1,3,4], so the secondary order inside category A differs from ANN, rating, and PK order.

tests/python_client/milvus_client/test_milvus_client_search_aggregation.py
lyyyuna: Fixed in 07c10807e7. The test now builds an independent FLAT/L2 oracle for exact bucket keys, counts, hit IDs, fields, and scores, and compares normalized results across limit < size, limit == size, and limit > size. The full SDK file passed: 51 passed, 1 existing xfail.

tests/restful_client_v2/testcases/test_search_aggregation.py
lyyyuna: Fixed in 07c10807e7. Category A now has four filtered rows while the retained size is three. The low-IP ID 6 must not contribute to count, metrics, or topHits, and a high-IP filtered-out row also verifies filter scoping.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Fixed in 07c10807e7. The case now seeds two independently named BM25 bindings, captures full field/function/index metadata, drops only one binding, and compares the complete surviving binding and dense index state unchanged.

tests/restful_client_v2/testcases/test_search_order_by.py
lyyyuna: Fixed in 07c10807e7. The fixture now produces distinct orders for ANN [3,4,1,2,5], price asc [4,2,3,5,1], and rating desc [1,4,5,2,3], with exact ID and scalar-value assertions.

lyyyuna:
Addressed all latest inline review findings in 07c10807e7:

- Decoupled search ORDER BY fixtures from ANN, PK, and alternative scalar orders, including filter-before-candidate selection.
- Added exact aggregation retention, filter, `size`/`searchSize`, nested child `searchSize`, `topHits`, multi-NQ, ID, field, and score assertions.
- Added an independent FLAT/L2 SDK oracle and cross-limit result equality checks.
- Strengthened add/drop field and function-field checks with complete survivor snapshots, duplicate metadata detection, index health checks, selective two-function cascade, and repeated-drop atomicity.

Validation:
- REST v2 targeted suite: 66 passed.
- PyMilvus search aggregation file: 51 passed, 1 existing xfailed.
- Ruff check/format and `git diff --check`: passed.

@yanliang567 PTAL.

yanliang567 (commented):
Review of head `85e95bd2`.

I found one P1 verification gap and several P2 coverage gaps:

### P1

- `tests/restful_client_v2/testcases/test_drop_field_operations.py:56-83` and `test_function_field_operations.py:120-237`: the before/after snapshots compare fields, functions, and indexes, but omit collection `properties`. Add/drop schema DDL updates `properties.max_field_id`; a rejected request that partially mutates this high-water mark, or a successful drop that lowers it incorrectly, can therefore pass. Please include normalized properties in the snapshots and assert the expected `max_field_id` behavior.

### P2

- `test_function_field_operations.py:195-217`: the new field/function `id` values are stripped before comparison. Please verify positive, unique field/function IDs and the input/output ID mapping, together with the updated `max_field_id`.

- `test_search_order_by.py:6-12,81-150`: all complete scalar sort keys are unique, so the suite does not verify equal-key tie behavior or pagination across a tied block. Add candidates with identical sort keys but deliberately different ANN/PK order, then assert the tie order.

- Search aggregation has no REST/SDK E2E case using `_score` as a metric source (`sum/avg/min/max`). The implementation has a dedicated score-field path, but current E2E coverage only uses `_score` for `topHits.sort`.

- `test_query_aggregation.py` always includes group keys in `outputFields`. Please add a REST case with `groupByFields=["category"]` and output fields containing only aggregates, and verify the response does not inject the group key. The SDK path has coverage, but the REST response remapping is not covered here.

The existing REST `test_collection_operations.py::test_drop_field_by_name_and_id` already covers the positive `fieldId` drop path, so I did not duplicate that finding.

Static validation is clean: Ruff check/format and `git diff --check` pass.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,192 @@
...
+        after_first_state = self._assert_collection_state(
+            collection_name,
+            expected_fields=["id", "dense"],
+            expected_indexes=["dense_idx"],
+        )
+        expected_after_first_state = {
+            "fields": {name: field for name, field in before_state["fields"].items() if name != "tag"},
+            "functions": before_state["functions"],
+            "indexes": before_state["indexes"],
+            "properties": before_state["properties"],
```
yanliang567, line 207: This still does not verify that `max_field_id` is a monotonic high-water mark. `tag` is created before `dense`, so deleting `tag` leaves a higher field ID; an implementation that incorrectly recomputes `max_field_id` from the surviving fields produces the same property value and passes. Please make the removable field the highest-ID field, assert the high-water mark is unchanged after the drop, then add another field and assert its ID is strictly greater rather than reused.

yanliang567 (commented):
Review of head `992d89cd`. I found three remaining false-green gaps and several P2 follow-ups. Details are inline. Non-blocking aggregation follow-ups: assert that child metrics/hits do not leak into a parent with no such configuration, and cover metric/topHits source fields omitted from `outputFields`.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+        assert rsp["code"] == 0, rsp
+        return collection_name
+
+    def _describe_index_state(self, collection_name, index_name):
+        rsp = self.index_client.index_describe(collection_name=collection_name, index_name=index_name)
+        assert rsp["code"] == 0, rsp
+        assert len(rsp["data"]) == 1, rsp
+        index = rsp["data"][0]
+        assert "fail" not in index["indexState"].lower(), rsp
```
yanliang567: This accepts non-ready states such as `Unissued` or `InProgress`, and the normalized snapshot below drops `indexState` entirely. A rejected or successful drop that regresses `dense_idx` from ready to rebuilding can therefore pass the before/after comparison. Please wait for the exact ready/Finished state and retain the health state in the snapshot.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,578 @@
...
+            )
+        )
+        assert rsp["code"] == 0, rsp
+
+    def _describe_index_state(self, collection_name, index_name):
+        rsp = self.index_client.index_describe(collection_name=collection_name, index_name=index_name)
+        assert rsp["code"] == 0, rsp
+        assert len(rsp["data"]) == 1, rsp
+        index = rsp["data"][0]
+        assert "fail" not in index["indexState"].lower(), rsp
```
yanliang567: The same index-health gap exists here: any state whose name lacks `fail` is accepted, and `indexState` is removed from the returned snapshot. The AutoIndex and cascade/atomicity cases can stay green when a new or surviving index never reaches a usable state. Please poll for exact ready/Finished state and include it in the snapshot.

tests/restful_client_v2/testcases/test_search_aggregation.py
```diff
@@ -0,0 +1,506 @@
...
+        assert rsp["aggTopks"] == [3]
+
+        buckets = rsp["data"][0]["buckets"]
+        expected = [
+            ("A", "X", 35, [5, 1], [5, 30], [0.6, 1.0]),
+            ("B", "X", 20, [3], [20], [0.8]),
+            ("A", "Y", 10, [2], [10], [0.9]),
+        ]
+        assert all([key["fieldName"] for key in bucket["key"]] == ["category", "brand"] for bucket in buckets)
+        for bucket, (category, brand, total_price, hit_ids, hit_prices, hit_scores) in zip(buckets, expected):
```
yanliang567, line 231: `all(...)` is true for an empty or short bucket list, and `zip(buckets, expected)` validates only the common prefix. `aggTopks == [3]` checks separate metadata, not the actual bucket-list cardinality, so a response missing the last composite bucket can pass. Please assert `len(buckets) == len(expected)` or the complete ordered composite-key list before the loop.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,192 @@
...
+        ],
+        ids=["missing-identifier", "both-identifiers", "zero-field-id", "negative-field-id"],
+    )
+    def test_drop_field_identifier_parameter_validation(self, field_name, field_id, expected_message):
+        """
+        target: verify REST drop field requires one valid identifier form
+        method: omit both identifiers, provide both, or provide a non-positive fieldId
+        expected: REST rejects each request before attempting a schema mutation
+        """
+        missing_collection_name = gen_collection_name(prefix=f"{self.__class__.__name__}Missing")
```
yanliang567: All identifier-validation requests target a nonexistent collection, so this cannot verify the docstring's no-mutation guarantee. In particular, a both-identifiers request could partially mutate a real collection before returning the parameter error and this test would remain green. Please use a real collection for the mutation-capable cases and compare the complete field/function/index/property state before and after.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,578 @@
...
+                (KEEP_FUNCTION_NAME, KEEP_OUTPUT_FIELD_NAME, KEEP_INDEX_NAME),
+            ],
+        )
+
+        rsp = self.collection_client.drop_function_field(
+            {"collectionName": collection_name, "functionName": FUNCTION_NAME}
+        )
+        assert rsp["code"] == 0, rsp
+
+        after_state = self._assert_bm25_function_schema_state(
```
yanliang567, line 507: The selective-cascade comparison verifies the surviving binding by REST names and metadata, but not its `input_field_ids`/`output_field_ids`. The SDK ID assertion covers only the single-function add path. Please capture the `bm25_keep_fn` SDK binding before and after the drop and assert that it still points to `text` and `sparse_keep`.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,578 @@
...
+            collection_name,
+            expected_bindings=[(FUNCTION_NAME, OUTPUT_FIELD_NAME, OUTPUT_FIELD_NAME)],
+            expected_index_types={OUTPUT_FIELD_NAME: "AUTOINDEX"},
+        )
+        assert {name: after_state["fields"][name] for name in before_state["fields"]} == before_state["fields"]
+        assert after_state["indexes"]["dense_idx"] == before_state["indexes"]["dense_idx"]
+
+    def test_add_function_field_rejects_unsupported_function_type(self):
```
yanliang567, line 444: The positive add/drop paths are all BM25, although the contract and the assertion below say MinHash is also supported. Please add a positive REST MinHash binding path and verify exact `BinaryVector`, `MINHASH_LSH`, `MHJACCARD`, field/function ID binding, and cascade-removal metadata.

tests/restful_client_v2/testcases/test_drop_field_operations.py
lyyyuna: Fixed in 4bf6d62e00. The fixture now declares `dense` before `tag`, making `tag` the highest-ID field. The test asserts max_field_id equals tag's field ID before the drop, stays unchanged after both the first and second drop requests, and that a newly added field receives a strictly higher ID with max_field_id advancing to it.

tests/restful_client_v2/testcases/test_drop_field_operations.py
lyyyuna: Fixed in 4bf6d62e00. `_describe_index_state` now goes through `_wait_index_ready`, which polls index_describe until indexState is `Finished` (30s timeout, hard failure otherwise), and the normalized snapshot includes `indexState`. A rejected or successful drop that regresses `dense_idx` out of the ready state now fails the test.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Fixed in 4bf6d62e00. Same hardening as the drop-field helper: `_describe_index_state` now waits for indexState `Finished` via `_wait_index_ready` and keeps `indexState` in the returned snapshot, so the AutoIndex and cascade/atomicity cases can no longer stay green on a non-ready index.

tests/restful_client_v2/testcases/test_search_aggregation.py
lyyyuna: Fixed in 4bf6d62e00. Added `assert len(buckets) == len(expected)` before the zip, so an empty or short bucket list now fails instead of passing vacuously through `all(...)` and prefix-only zip comparison.

tests/restful_client_v2/testcases/test_drop_field_operations.py
lyyyuna: Fixed in 4bf6d62e00. The parametrized identifier-validation case now creates a real collection, captures the full schema + index state before the request, and asserts the identical state after each rejection — including the both-identifiers request — so any partial mutation is detected.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Fixed in 4bf6d62e00. Added `_capture_sdk_function_binding`, which reads `function_id`, `input_field_ids`, and `output_field_ids` through the SDK describe view. The selective-cascade test now captures the surviving binding before the drop and asserts its IDs are unchanged afterward.

tests/restful_client_v2/testcases/test_function_field_operations.py
lyyyuna: Fixed in 4bf6d62e00. Added L0 case `test_add_and_drop_minhash_function_field`: a positive REST MinHash binding path that asserts exact `BinaryVector` output (dim = num_hashes * 32), `MINHASH_LSH` index with `MHJACCARD` metric, function type 4, field/function ID binding cross-checked via SDK, and cascade drop restoring the exact base schema state.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,238 @@
...
+                        "indexType": "AUTOINDEX",
+                        "metricType": "L2",
+                    }
+                ],
+            }
+        )
+        assert rsp["code"] == 0, rsp
+        return collection_name
+
+    def _wait_index_ready(self, collection_name, index_name, timeout=30):
```
yanliang567, line 40: Non-blocking (follow-up in a later PR): `_wait_index_ready` and `_describe_index_state` are duplicated almost verbatim between this file and `test_function_field_operations.py`. Consider extracting them into a shared helper under `common/`.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,698 @@
...
+        assert dense_index["indexType"] == "AUTOINDEX"
+        index_metadata = {"dense_idx": dense_index}
+        for function_name, output_field_name, index_name in expected_bindings:
+            assert int(fields[output_field_name]["id"]) > 0
+            output_field = {key: value for key, value in fields[output_field_name].items() if key != "id"}
+            assert output_field == {
+                "autoId": False,
+                "clusteringKey": False,
+                "description": "",
+                "isFunctionOutput": True,
```
yanliang567, line 291: Non-blocking (follow-up): asserting the output field with full-dict equality (including `isFunctionOutput`, `clusteringKey`, `description`, `nullable`, `partitionKey`, `primaryKey`) couples the test to the complete REST describe payload shape — any new field added to the describe response will break it. Consider asserting only the fields of interest instead.

tests/restful_client_v2/testcases/test_drop_field_operations.py
```diff
@@ -0,0 +1,238 @@
...
+            "functions": functions,
+            "indexes": index_metadata,
+            "properties": properties,
+        }
+
+    @pytest.mark.parametrize(
+        "field_name,field_id,expected_message",
+        [
+            (None, None, "exactly one of fieldName or fieldId is required"),
+            ("tag", 101, "exactly one of fieldName or fieldId is required"),
```
yanliang567, line 111: Non-blocking (follow-up): `field_id=101` is now stale — after declaring `dense` before `tag`, `101` is `dense`'s field ID rather than `tag`'s. It is harmless for the 'both identifiers' assertion, but the magic number is misleading; consider dropping the concrete id.

tests/restful_client_v2/testcases/test_query_aggregation.py
```diff
@@ -0,0 +1,349 @@
...
+                "limit": 2,
+                "offset": 1,
+                "outputFields": ["category", "count(*)", "sum(score)"],
+                "groupByFields": ["category"],
+                "orderByFields": ["category:desc"],
+            }
+        )
+
+        assert rows == [
+            {"category": "category_3", "count(*)": 40, "sum(score)": 3980},
```
yanliang567, line 293: Non-blocking (follow-up): hardcoded `count(*)`/`sum(score)` values here differ from the rest of this file, which derives expectations via `_expected_grouped_rows`. Using the ground-truth helper would keep this resilient to `_rows()` changes.

tests/restful_client_v2/testcases/test_query_aggregation.py
```diff
@@ -0,0 +1,349 @@
...
+            if aggregate_function == "count":
+                expected[aggregate_field] = len(values)
+            elif aggregate_function == "sum":
+                expected[aggregate_field] = sum(values)
+            elif aggregate_function == "min":
+                expected[aggregate_field] = min(values)
+            elif aggregate_function == "max":
+                expected[aggregate_field] = max(values)
+            elif aggregate_function == "avg":
+                expected[aggregate_field] = sum(values) / len(values)
```
yanliang567, line 52: Non-blocking (follow-up): `sum(values)/len(values)` (and `min(values)`/`max(values)` above) will raise on a group whose aggregated field is all-NULL. Not triggered by the current data, but the oracle is fragile if the null distribution changes.

tests/restful_client_v2/testcases/test_function_field_operations.py
```diff
@@ -0,0 +1,698 @@
...
+        return {
+            "fieldName": index["fieldName"],
+            "indexName": index["indexName"],
+            "metricType": index["metricType"],
+            "indexType": index["indexType"],
+            "indexState": index["indexState"],
+            "indexParams": sorted(index_params),
+        }
+
+    def _capture_sdk_function_binding(self, collection_name, function_name, output_field_name):
```
yanliang567, line 170: Non-blocking (follow-up): `_capture_sdk_function_binding` opens and closes a new `MilvusClient` on every call (called twice in the cascade test). Consider reusing a single client instance.

yanliang567 (commented):
Non-blocking follow-up for the next PR (left as a general note because the lines sit outside the diff): in `test_query_order_by.py`, `test_query_order_by_single_field_asc` and `test_query_order_by_single_field_desc` still assert the exact ID order via `_assert_expected_ids`, which relies on `score = (i*17+11) % NB` being a permutation with no ties. This is inconsistent with the `_assert_expected_members` style used for the multi-field case and becomes fragile if `_rows()` introduces ties.

Issue #52163, opened 2026-08-04
[Enhancement]: Complete RESTful v2 coverage and semantic parity for public PyMilvus 3.0 APIs


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Create a tracking task to close the confirmed coverage and semantic gaps between RESTful v2 and the public PyMilvus 3.0 `MilvusClient` API.

This is not a request to expose every Milvus gRPC method. Internal, diagnostic, and component-management RPCs that are not available through the public high-level PyMilvus client are explicitly out of scope.

#### Audit baseline

- Milvus `upstream/3.0`: `07df976ff67561906ffb8bcbbc4964fbb861d7ca`
- PyMilvus `upstream/3.0`: `cb16a04cb0f0f7c07260c56c4bfd5e8442692ed7`
- Method scope: public high-level `pymilvus.MilvusClient`
- Verification method: read-only source audit
- Runtime and E2E confirmation: still required as implementation acceptance work

The audit found 17 usable or intended public high-level PyMilvus capabilities without RESTful v2 equivalents.

PyMilvus also declares `get_flush_all_state()`, but its current high-level implementation does not pass the required `flush_all_ts` to the underlying handler. It is therefore not counted as an independent working capability in this audit. RESTful v2 should nevertheless provide usable completion semantics together with `flush_all()`.

#### Current RESTful v2 coverage: missing capabilities

| Capability | Public PyMilvus API | Backend gRPC API | Current RESTful v2 state |
| --- | --- | --- | --- |
| Server version | `get_server_version()` | `GetVersion` | No equivalent RESTful v2 endpoint |
| Detailed server/build information | `get_server_version(detail=True)` | `Connect` | No equivalent RESTful v2 endpoint |
| Replica inspection | `describe_replica()` | `GetReplicas` | No equivalent RESTful v2 endpoint |
| Global flush | `flush_all()` | `FlushAll`, `GetFlushAllState` | No global flush endpoint or usable completion workflow |
| Loaded segment inspection | `list_loaded_segments()` | `GetQuerySegmentInfo` | No equivalent RESTful v2 endpoint |
| Persistent segment inspection | `list_persistent_segments()` | `GetPersistentSegmentInfo` | `/segments/describe` invokes `GetSegmentsInfo`; its request and response semantics are different |
[...]

PR #52171, opened 2026-08-04
enhance: add native snapshot RESTful v2 APIs

zhuwenxing:
issue: #52163

## What this PR does

This PR adds RESTful v2 coverage for the native snapshot lifecycle exposed by the public PyMilvus 3.0 `MilvusClient`.

Snapshot identity is collection-scoped. Resource operations resolve `dbName` and `collectionName` to a collection ID, and snapshot names are unique only within that collection. The REST contract therefore requires `collectionName` for create, drop, list, describe, and pin operations, and requires `sourceCollectionName` for restore.

The following endpoints are added under `/v2/vectordb/snapshots/`:

- `create` → `CreateSnapshot`
- `drop` → `DropSnapshot`
- `list` → `ListSnapshots`
- `describe` → `DescribeSnapshot`
- `restore` → `RestoreSnapshot`
- `pin` → `PinSnapshotData`
- `unpin` → `UnpinSnapshotData`

Native restore returns a `jobId`. The existing snapshot job endpoints remain responsible for asynchronous job operations and inspection:

- `/v2/vectordb/jobs/snapshot/restore_external` → `RestoreExternalSnapshot`
- `/v2/vectordb/jobs/snapshot/export` → `ExportSnapshot`
- `/v2/vectordb/jobs/snapshot/describe` → `GetRestoreSnapshotState`
- `/v2/vectordb/jobs/snapshot/list` → `ListRestoreSnapshotJobs`

The `export` and `restore_external` backends currently return `ServiceUnimplemented`; this PR does not implement those gRPC capabilities.

Request and response fields follow RESTful v2 camelCase conventions while preserving the public user-facing snapshot semantics, including collection-scoped snapshot identity, snapshot metadata, restore job IDs, and pin IDs.

All handlers use the existing authentication, authorization, rate-limit, timeout, and gRPC error-projection paths. Route metadata and `fullMethod` values match the gRPC methods actually invoked.

## Tests

- Added handler tests covering request mappings and response projections.
- Added validation coverage for collection-scoped identity, required fields, and numeric constraints.
- Added a RESTful v2 end-to-end lifecycle test covering create, list, describe, pin, unpin, restore, restore-state polling, and drop.
- Covered all `/v2/vectordb/jobs/snapshot` routes in the end-to-end case: `describe` and `list` validate completed restore jobs, while `export` and `restore_external` validate their current unimplemented responses.
- Moved snapshot cleanup into teardown so snapshots are removed before their collections.
- Full `internal/distributed/proxy/httpserver` Go unit test suite passed.
- RESTful v2 end-to-end test passed.
- Ruff lint and formatting checks passed.

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -182,6 +182,8 @@ var routeToMethod = map[string]string{ //nolint:gosec // not credentials, just a
 	"/v2/vectordb/snapshots/list":                    "ListSnapshots",
 	"/v2/vectordb/snapshots/describe":                "DescribeSnapshot",
 	"/v2/vectordb/snapshots/restore":                 "RestoreSnapshot",
+	"/v2/vectordb/snapshots/pin":                     "PinSnapshotData",
+	"/v2/vectordb/snapshots/unpin":                   "UnpinSnapshotData",
 	"/v2/vectordb/jobs/external_collection/refresh":  "RefreshExternalCollection",
 	"/v2/vectordb/jobs/external_collection/describe": "GetRefreshExternalCollectionProgress",
 	"/v2/vectordb/jobs/external_collection/list":     "ListRefreshExternalCollectionJobs",
@@ -354,6 +356,8 @@ func (h *HandlersV2) RegisterRoutesToV2(router gin.IRouter) {
 	router.POST(SnapshotCategory+ListAction, timeoutMiddleware(wrapperPost(func() any { return &OptionalCollectionNameReq{} }, wrapperTraceLog(h.listSnapshots))))
 	router.POST(SnapshotCategory+DescribeAction, timeoutMiddleware(wrapperPost(func() any { return &SnapshotReq{} }, wrapperTraceLog(h.describeSnapshot))))
 	router.POST(SnapshotCategory+RestoreAction, timeoutMiddleware(wrapperPost(func() any { return &RestoreSnapshotReq{} }, wrapperTraceLog(h.restoreSnapshot))))
+	router.POST(SnapshotCategory+PinAction, timeoutMiddleware(wrapperPost(func() any { return &PinSnapshotDataReq{} }, wrapperTraceLog(h.pinSnapshotData))))
+	router.POST(SnapshotCategory+UnpinAction, timeoutMiddleware(wrapperPost(func() any { return &UnpinSnapshotDataReq{} }, wrapperTraceLog(h.unpinSnapshotData))))
 	router.POST(ExternalCollectionJobCategory+RefreshAction, timeoutMiddleware(wrapperPost(func() any { return &RefreshExternalCollectionReq{} }, wrapperTraceLog(h.refreshExternalCollection))))
 	router.POST(ExternalCollectionJobCategory+DescribeAction, timeoutMiddleware(wrapperPost(func() any { return &RefreshExternalCollectionProgressReq{} }, wrapperTraceLog(h.getRefreshExternalCollectionProgress))))
 	router.POST(ExternalCollectionJobCategory+ListAction, timeoutMiddleware(wrapperPost(func() any { return &OptionalCollectionNameReq{} }, wrapperTraceLog(h.listRefreshExternalCollectionJobs))))
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -177,6 +177,10 @@ var routeToMethod = map[string]string{ //nolint:gosec // not credentials, just a
 	"/v2/vectordb/jobs/snapshot/export/describe":     "GetExportSnapshotState",
 	"/v2/vectordb/jobs/snapshot/describe":            "GetRestoreSnapshotState",
 	"/v2/vectordb/jobs/snapshot/list":                "ListRestoreSnapshotJobs",
+	"/v2/vectordb/snapshots/create":                  "CreateSnapshot",
+	"/v2/vectordb/snapshots/drop":                    "DropSnapshot",
+	"/v2/vectordb/snapshots/list":                    "ListSnapshots",
+	"/v2/vectordb/snapshots/describe":                "DescribeSnapshot",
 	"/v2/vectordb/jobs/external_collection/refresh":  "RefreshExternalCollection",
 	"/v2/vectordb/jobs/external_collection/describe": "GetRefreshExternalCollectionProgress",
 	"/v2/vectordb/jobs/external_collection/list":     "ListRefreshExternalCollectionJobs",
@@ -344,6 +348,10 @@ func (h *HandlersV2) RegisterRoutesToV2(router gin.IRouter) {
 	router.POST(SnapshotJobCategory+DescribeExportAction, timeoutMiddleware(wrapperPost(func() any { return &JobIDReq{} }, wrapperTraceLog(h.getExportSnapshotState))))
 	router.POST(SnapshotJobCategory+DescribeAction, timeoutMiddleware(wrapperPost(func() any { return &JobIDReq{} }, wrapperTraceLog(h.getRestoreSnapshotState))))
 	router.POST(SnapshotJobCategory+ListAction, timeoutMiddleware(wrapperPost(func() any { return &OptionalCollectionNameReq{} }, wrapperTraceLog(h.listRestoreSnapshotJobs))))
+	router.POST(SnapshotCategory+CreateAction, timeoutMiddleware(wrapperPost(func() any { return &CreateSnapshotReq{} }, wrapperTraceLog(h.createSnapshot))))
+	router.POST(SnapshotCategory+DropAction, timeoutMiddleware(wrapperPost(func() any { return &SnapshotReq{} }, wrapperTraceLog(h.dropSnapshot))))
+	router.POST(SnapshotCategory+ListAction, timeoutMiddleware(wrapperPost(func() any { return &OptionalCollectionNameReq{} }, wrapperTraceLog(h.listSnapshots))))
+	router.POST(SnapshotCategory+DescribeAction, timeoutMiddleware(wrapperPost(func() any { return &SnapshotReq{} }, wrapperTraceLog(h.describeSnapshot))))
 	router.POST(ExternalCollectionJobCategory+RefreshAction, timeoutMiddleware(wrapperPost(func() any { return &RefreshExternalCollectionReq{} }, wrapperTraceLog(h.refreshExternalCollection))))
 	router.POST(ExternalCollectionJobCategory+DescribeAction, timeoutMiddleware(wrapperPost(func() any { return &RefreshExternalCollectionProgressReq{} }, wrapperTraceLog(h.getRefreshExternalCollectionProgress))))
 	router.POST(ExternalCollectionJobCategory+ListAction, timeoutMiddleware(wrapperPost(func() any { return &OptionalCollectionNameReq{} }, wrapperTraceLog(h.listRefreshExternalCollectionJobs))))
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -178,6 +178,7 @@ var routeToMethod = map[string]string{ //nolint:gosec // not credentials, just a
 	"/v2/vectordb/snapshots/drop":                    "DropSnapshot",
 	"/v2/vectordb/snapshots/list":                    "ListSnapshots",
 	"/v2/vectordb/snapshots/describe":                "DescribeSnapshot",
+	"/v2/vectordb/snapshots/restore":                 "RestoreSnapshot",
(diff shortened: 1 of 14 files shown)


internal/distributed/proxy/httpserver/snapshot_handler_v2.go
```diff
@@ -0,0 +1,176 @@
...
+	c.Set(ContextRequest, req)
+
+	resp, err := wrapperProxyWithLimit(ctx, c, req, h.checkAuth, false, "/milvus.proto.milvus.MilvusService/RestoreSnapshot", true, h.proxy, func(reqCtx context.Context, req any) (interface{}, error) {
+		return h.proxy.RestoreSnapshot(reqCtx, req.(*milvuspb.RestoreSnapshotRequest))
+	})
+	if err == nil {
+		HTTPReturn(c, http.StatusOK, gin.H{
+			HTTPReturnCode: merr.Code(nil),
+			HTTPReturnData: gin.H{
+				"jobId": resp.(*milvuspb.RestoreSnapshotResponse).GetJobId(),
```
liliu-z: restoreSnapshot serializes jobId as a raw JSON number (line 131), but jobId comes from the TSO-based global ID allocator so it exceeds 2^53; a default JS/Node client rounds it on JSON.parse and then polls /v2/vectordb/jobs/snapshot/describe — whose JobIDReq.JobID input is a decimal string parsed with ParseInt — with the wrong id, getting a permanent not-found. Emit it via formatRESTInt64 (honoring Accept-Type-Allow-Int64) as file_resource_handler_v2.go:72 does; the pre-existing sibling emits at handler_v2.go:3872 and 3906 share the bug and should be aligned for consistency.

internal/distributed/proxy/httpserver/request_v2.go
```diff
@@ -377,6 +377,55 @@ type JobIDReq struct {
...
+	SnapshotName   string `json:"snapshotName" binding:"required"`
+	TTLSeconds     int64  `json:"ttlSeconds" binding:"gte=0"`
+}
+
+func (req *PinSnapshotDataReq) GetDbName() string { return req.DbName }
+
+func (req *PinSnapshotDataReq) GetCollectionName() string { return req.CollectionName }
+
+type UnpinSnapshotDataReq struct {
+	PinID int64 `json:"pinId" binding:"required,gt=0"`
```
liliu-z: pinId is TSO-derived (>2^53) yet the pin response emits it as a raw JSON number (snapshot_handler_v2.go:155) and the unpin DTO takes it as a JSON int64 (request_v2.go:426), so it needs a two-sided fix — a default JS/Node client both rounds it on output and cannot resubmit the exact value through an int64 JSON field. Because UnpinSnapshot returns success for an unknown pinId (snapshot_meta.go:1607), unpin silently no-ops against the rounded id, and a permanent ttl=0 pin then keeps DropSnapshot failing with ErrSnapshotPinned (snapshot_meta.go:663) indefinitely. Emit pinId via formatRESTInt64 and change PinID to a decimal-string input.

internal/distributed/proxy/httpserver/snapshot_handler_v2.go
```diff
@@ -0,0 +1,176 @@
...
+	if err == nil {
+		snapshot := resp.(*milvuspb.DescribeSnapshotResponse)
+		HTTPReturn(c, http.StatusOK, gin.H{
+			HTTPReturnCode: merr.Code(nil),
+			HTTPReturnData: gin.H{
+				"snapshotName":   snapshot.GetName(),
+				"description":    snapshot.GetDescription(),
+				"collectionName": snapshot.GetCollectionName(),
+				"partitionNames": snapshot.GetPartitionNames(),
+				"createTs":       snapshot.GetCreateTs(),
```
liliu-z: describeSnapshot emits createTs as a raw JSON number (line 105), but createTs is a TSO and exceeds 2^53, so default JS/Node clients lose its low bits on JSON.parse — the same raw-int64 pattern the reviewers flagged for jobId/pinId but missed here. Impact is milder (the physical-time high bits survive, so wall-clock decoding still works; only exact-value equality/ordering and the v2 string-id convention break), but it should also go through formatRESTInt64.

internal/distributed/proxy/httpserver/snapshot_handler_v2.go
zhuwenxing: Fixed in 197610f9b0. Restore job IDs and the existing snapshot job responses now use formatRESTInt64 and honor Accept-Type-Allow-Int64. Added boundary coverage for values above 2^53.

internal/distributed/proxy/httpserver/request_v2.go
zhuwenxing: Fixed in 197610f9b0. pinId responses now use formatRESTInt64, while unpin accepts and validates a positive decimal int64 string. Added round-trip coverage for values above 2^53.

internal/distributed/proxy/httpserver/snapshot_handler_v2.go
zhuwenxing: Fixed in 197610f9b0. createTs now uses formatRESTInt64 and honors Accept-Type-Allow-Int64, with boundary coverage above 2^53.

internal/distributed/proxy/httpserver/snapshot_handler_v2.go
```diff
@@ -0,0 +1,186 @@
...
+	return resp, err
+}
+
+func (h *HandlersV2) restoreSnapshot(ctx context.Context, c *gin.Context, anyReq any, dbName string) (interface{}, error) {
+	httpReq := anyReq.(*RestoreSnapshotReq)
+	req := &milvuspb.RestoreSnapshotRequest{
+		Name:                 httpReq.SnapshotName,
+		DbName:               dbName,
+		CollectionName:       httpReq.SourceCollectionName,
+		TargetDbName:         httpReq.TargetDbName,
```
liliu-z: The restore handler forwards `targetDbName` straight to the Proxy, but nothing in the authorization path ever looks at it: `PrivilegeInterceptor` resolves the policy database from `RestoreSnapshotRequest.DbName` (the source db) and the policy object from field 4, `collection_name` (the source collection). Restore then creates a brand-new collection in the target db via `broker.CreateCollection`, an internal coord call that never re-enters proxy RBAC — so a caller holding `PrivilegeRestoreSnapshot` on `tenant_a/orders` can set `targetDbName: tenant_b` and materialize a collection in a database they hold no grant on. The same function already handles the analogous cross-db case for RenameCollection by escalating to a cluster-scoped check, so the precedent for the fix exists. The gap is in the shared Proxy path and is equally reachable over gRPC today, but this PR makes it a second, documented transport, so it is worth resolving before shipping rather than inheriting silently.

liliu-z: `targetDbName` is the only database field on this route that does not go through the REST db-resolution chain. `wrapperPost` resolves the source db from `sourceDbName`, then the `dbName` header, then `default`, and the handler passes that resolved value as `DbName` — but `TargetDbName` is forwarded verbatim, and neither the database interceptor nor rootcoord maps it to the caller's current database. A user working in `tenant_a` who omits `targetDbName` gets the restored collection created in `default`, with no error. The PR's own rate-limit code already encodes this assumption, so quota is charged to `default` too and the two stay consistent — but consistently wrong. Defaulting an empty `targetDbName` to the same resolved db the handler already computed would remove the asymmetry.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -3890,12 +3905,12 @@ func (h *HandlersV2) exportSnapshot(ctx context.Context, c *gin.Context, anyReq
...
 }
 
-func restoreSnapshotJobToREST(info *milvuspb.RestoreSnapshotInfo) gin.H {
+func restoreSnapshotJobToREST(info *milvuspb.RestoreSnapshotInfo, allowInt64 bool) gin.H {
 	if info == nil {
 		return gin.H{}
 	}
 	return gin.H{
-		"jobId":          info.GetJobId(),
+		"jobId":          formatRESTInt64(info.GetJobId(), allowInt64),
```
liliu-z: This flips `jobId` from a JSON number to a JSON string (absent the `Accept-Type-Allow-Int64` header) on three routes that already shipped in v3.0.0: `/v2/vectordb/jobs/snapshot/restore_external`, `/describe`, and `/list`. The PR's own test diff records the break — `"jobId":2001` became `"jobId":"2001"`. Stringifying int64 is the right REST v2 convention, but it is a wire-format change on a released surface, so it should be a deliberate call rather than a side effect of adding new routes. The sweep also stops at the snapshot family: the external-collection job routes still emit `jobId` as a raw number, so the two job families now disagree.

internal/distributed/proxy/httpserver/snapshot_handler_v2.go
zhuwenxing: Fixed in 93d40946c5. An omitted `targetDbName` now defaults to the resolved source database in both the REST handler and `DatabaseInterceptor`. Rate-limit accounting uses the same fallback. Added REST, interceptor, and quota-routing coverage.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: [P1 / blocker] This compatibility issue remains on current head `0c000ce`.

The three routes already released in v3.0.0 returned a JSON number, while the current default returns a string unless an existing client knows to send `Accept-Type-Allow-Int64: true`. Marking the thread resolved does not provide a migration path for typed clients that unmarshal `jobId` as an integer.

Please either preserve the released default schema in this PR or make the type migration an explicit, documented/versioned compatibility decision.

internal/proxy/database_interceptor.go
```diff
@@ -358,6 +358,9 @@ func fillDatabase(ctx context.Context, req interface{}) (context.Context, interf
 		if r.DbName == "" {
 			r.DbName = GetCurDBNameFromContextOrDefault(ctx)
 		}
+		if r.TargetDbName == "" {
+			r.TargetDbName = r.DbName
```
MrPresent-Han: [P1 / blocker] Default an omitted target database from the active database, not from the explicit source database.

PyMilvus 3.0 documents `source_db_name` and `target_db_name` as independently defaulting to the active database. With active database `tenant_a`, explicit source `archive`, and no target, this assignment restores into `archive`; `restoreSnapshotTask` forwards it unchanged, and RBAC/rate limiting incorrectly classify the operation as same-database.

Please capture the active database from context before filling either field, then default each omitted field independently from it. The REST handler should mirror the active `DB-Name` header/default instead of the resolved source. Add the missing explicit-source plus different-active-database test.

internal/proxy/privilege_interceptor.go
```diff
@@ -102,6 +102,18 @@ func PrivilegeInterceptor(ctx context.Context, req interface{}) (context.Context
...
+	// RestoreSnapshot is collection-scoped within one database. Restoring into
+	// another database creates a collection there, so require the same privilege
+	// at cluster scope (db="*") instead of authorizing only against the source.
+	if r, ok := req.(*milvuspb.RestoreSnapshotRequest); ok {
+		targetDBName := r.GetTargetDbName()
+		if targetDBName == "" {
+			targetDBName = dbName
+		}
+		if dbName != targetDBName {
+			dbName = util.AnyWord
```
MrPresent-Han: Blocker: cross-DB restore changes `dbName` to `*` before DB-scoped alias resolution. Authorization can therefore check the literal alias while execution resolves the canonical source collection, so RBAC may validate the wrong resource. Please preserve `sourceDBName` for alias resolution and use a separate `policyDBName = "*"` only for Casbin; add alias coverage such as `tenant_a/orders_alias -> payroll`.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: [P1 / blocker] Re-opening this compatibility issue as a fresh thread because the previous thread is resolved. These three v3.0.0 routes changed `jobId` from the released JSON number default to a JSON string unless clients send `Accept-Type-Allow-Int64: true`, which breaks typed clients without a migration path. Please preserve the released default schema here, or make this an explicit documented/versioned compatibility decision.

zhuwenxing: Fixed in d5a2a0b163. Restored the released JSON number format for `jobId` on the v3.0.0 `restore_external`, `describe`, and `list` routes. The new native snapshot endpoints continue to follow the string-int64 convention by default. Boundary tests cover values above 2^53 and both `Accept-Type-Allow-Int64` modes.

internal/proxy/privilege_interceptor.go
zhuwenxing: Fixed in 97ff162a3b. Alias resolution now continues to use the source database, while a separate `policyDBName` is promoted to `"*"` only for cross-database Casbin enforcement. Added coverage for `tenant_a/orders_alias -> payroll` with a cluster-scoped grant on `payroll`.

PR #52173, opened 2026-08-04
enhance: batch Arrow field serialization

sunby:
## What

- add an optional `serdeEntry.serializeBatch` path and keep row-wise `serialize` as the fallback
- use Arrow batch append APIs for bool/integer/floating/timestamptz, varchar/string, JSON/geometry, dense vectors, and sparse float vectors
- preserve compact nullable-vector layout while validating dimensions, physical row counts, validity lengths, and schema/data nullability
- preserve scalar metadata when merging `FieldData` so batch validation receives the original nullable/type information

`Array`, `ArrayOfVector`, and `Text` remain on the existing row-wise path because their serialization requires per-row conversion or supports multiple physical representations.

## Why

- keep batch serialization owned by each `serdeEntry`, consistent with the existing type dispatch
- avoid per-row builder dispatch and temporary FloatVector byte slices
- retain amortized Arrow builder growth when one `RecordBuilder` is reused across chunks

## Coverage

Added coverage for:

- every type that enables `serializeBatch`, including nullable rows
- both directions of schema/data nullability mismatch for primitive, dense-vector, and sparse-vector paths
- merged `FieldData` nullable/type metadata
- FloatVector source-byte reuse and amortized builder growth

Local tests were not run per request. Static verification completed with `golangci-lint` for `./internal/storage/...` and the repository clang-format script. Full `make lint-fix` was attempted but is blocked by the checkout's missing generated package `cmd/tools/migration/legacy/legacypb`.

issue: #53181


internal/storage/serde.go
```diff
@@ -1366,6 +1366,99 @@ func NewSimpleArrowRecord(r arrow.Record, field2Col map[FieldID]int) *simpleArro
...
+		if remaining := typedBuilder.Cap() - typedBuilder.Len(); remaining < logicalRows {
+			typedBuilder.Resize(typedBuilder.Len() + logicalRows)
+		}
+		appendValue = typedBuilder.Append
+		appendNull = typedBuilder.AppendNull
+	case *array.BinaryBuilder:
+		if remaining := typedBuilder.Cap() - typedBuilder.Len(); remaining < logicalRows {
+			typedBuilder.Resize(typedBuilder.Len() + logicalRows)
+		}
+		typedBuilder.ReserveData(len(bytesData))
```
buqian-zilliz: **[P2] Per-chunk exact buffer resizing causes O(n²) reallocations in reused builders**

`appendFloatVectorFieldData` pre-sizes with `typedBuilder.Resize(typedBuilder.Len()+logicalRows)` and `typedBuilder.ReserveData(len(bytesData))`. In the vendored Arrow Go (v17.0.1), `BinaryBuilder.ReserveData`/`Resize` grow the underlying buffers via `roundUpToMultipleOf64` (linear), whereas the previous per-row `Append` path grew via `bitutil.NextPowerOf2` (geometric). When one `array.RecordBuilder` is reused across many `BuildRecord` calls — `BulkPackWriterV2.serializeBinlog` iterates every chunk in `pack.insertData` — each chunk triggers a full-buffer `Reallocate` that copies all previously accumulated data, making total copying quadratic in the number of chunks. This hits the `BinaryBuilder` data and offsets buffers (nullable path) and the validity bitmap; for `FixedSizeBinaryBuilder`, `Resize` also only resizes the validity bitmap after the first use, so its data buffer is pre-sized only on the first chunk. Prefer geometric growth (e.g. `Reserve`/`NextPowerOf2`) or size once from the total row count when it is known.

internal/storage/serde.go
```diff
@@ -110,10 +110,270 @@ type serdeEntry struct {
...
+			"field validity length mismatch, rows=%d, validity=%d, fieldData=%T",
+			rowCount,
+			len(validData),
+			data,
+		)
+	}
+	return validData, nil
+}
+
+func serializePrimitiveFieldData[T any](
```
buqian-zilliz, line 164: **[P3] Scalar/sparse batch serializers do not validate nullability against schema**

`serializePrimitiveFieldData` and `serializeSparseFloatVectorFieldData` take `nullable` from the schema but never compare it with `data.GetNullable()`, unlike `serializeDenseVectorFieldData` which returns an explicit nullable-mismatch error. If the data nullability ever disagrees with the schema (a non-nullable schema field carrying `Nullable=true` data with `ValidData`), `batchValidity` returns a nil validity slice, so `AppendValues` treats every row as valid: null rows are silently written as zero/empty values for scalars, and for sparse vectors the compacted null rows are dropped from the output instead of being rejected. Add the same `data.GetNullable() != nullable` validation to both batch serializers so a mismatch fails loudly rather than corrupting data.

internal/storage/serde.go
buqian-zilliz, line 1: **[P3] PR description/tests do not reflect the expanded batch-serialization scope**

The PR title and body still describe only the `FloatVector` fast path and list a single `FloatVector` test, but the final commit also enables batch Arrow serialization for every scalar type plus `BinaryVector`, `Float16Vector`, `BFloat16Vector`, `Int8Vector`, and `SparseFloatVector`. Reviewers reading the description will underestimate the blast radius, and the single described test does not exercise the newly batched scalar/vector paths. Please update the title/body (and ideally add coverage) to match the actual scope.

internal/storage/serde.go
```diff
@@ -110,10 +110,280 @@ type serdeEntry struct {
...
 
+type batchValuesBuilder[T any] interface {
+	AppendValues([]T, []bool)
+}
+
+type validFieldData interface {
+	GetValidData() []bool
+}
+
+func validateBatchNullability(data FieldData, nullable bool) error {
```
buqian-zilliz, line 128: **[P2] Strict nullability validation assumes every FieldData producer sets `Nullable`**

The new `validateBatchNullability` is called from `batchValidity` for every batch-serialized type and directly from `serializeDenseVectorFieldData`, and it now hard-fails whenever `data.GetNullable()` differs from the schema's `nullable`. That invariant is only safe if every `FieldData` producer that reaches `BuildRecord` populates `Nullable` from the schema. Some construction sites still leave `Nullable` as the zero value `false`: the scalar cases of `AddInsertData` in `internal/storage/data_codec.go` append `ValidData` for nullable fields but never set `Nullable`, and `buildSparseFloatVectorArrowArray` in `internal/datanode/compactor/record_materializer.go` constructs `SparseFloatVectorFieldData` without setting `Nullable`. If either flow reaches `BuildRecord` for a nullable schema field, serialization now fails with a `field nullable mismatch` error where the previous row-wise path performed no such check. Please verify these producers (or set `Nullable` from the schema) so the new validation does not reject otherwise-valid data.

sunby: Thanks for pointing this out. I traced both construction sites:

- The current non-test caller of `AddInsertData` is the V1 binlog deserialization path used by `indexBuildTask.parseFieldMetaFromBinlog`. It only uses the deserialized `InsertData` to recover field ID/type metadata and does not pass it to `BuildRecord`.
- `buildSparseFloatVectorArrowArray` does call `BuildRecord`, but nullable function-output fields are rejected before this function is reached, so both the schema and the constructed `SparseFloatVectorFieldData` are non-nullable on the current path.

Therefore neither of the two named paths currently triggers the new mismatch check. I agree that `AddInsertData` still produces inconsistent nullability metadata. Fixing that correctly requires propagating descriptor nullability explicitly and auditing the legacy row-based producer path, which is separate from this Arrow batch-serialization change. I plan to address that normalization in a follow-up PR rather than expand this PR further. The strict validation here remains useful to prevent nullable data from being silently serialized under a non-nullable schema.

zhengbuqian (approved):

internal/storage/serde.go
```diff
@@ -110,10 +110,280 @@ type serdeEntry struct {
...
+	serializeBatch func(b array.Builder, data FieldData, nullable bool, elementType schemapb.DataType) error
 }
 
 type TextLobRef []byte
 
+type batchValuesBuilder[T any] interface {
+	AppendValues([]T, []bool)
+}
+
+type validFieldData interface {
```
buqian-zilliz: **[P3] Remove the duplicate validity interface**

`FieldData` already declares `GetValidData() []bool`, so every value accepted by `batchValidity` necessarily satisfies `validFieldData`. The additional interface assertion and its unreachable error branch add a second contract without providing any protection. Remove `validFieldData` and call `data.GetValidData()` directly.

internal/storage/serde.go
```diff
@@ -110,10 +110,280 @@ type serdeEntry struct {
 	// 	nil is serialized to null without checking the type nullability.
 	//	elementType is only used for ArrayOfVector
 	serialize func(b array.Builder, v any, elementType schemapb.DataType) error
+	// serializeBatch serializes complete field data through Arrow batch APIs.
+	// If nil, BuildRecord falls back to serialize one row at a time.
+	serializeBatch func(b array.Builder, data FieldData, nullable bool, elementType schemapb.DataType) error
```
buqian-zilliz: **[P3] Drop the unused element type from the batch callback**

Every registered `serializeBatch` implementation discards `elementType`, while `ArrayOfVector`, the only type that needs it, remains on the row-wise path. Remove this fourth callback parameter and the corresponding unused parameters from the batch helpers until a batch serializer actually needs it; `BuildRecord` only needs to pass `elementType` to the existing row-wise `serialize` callback.

Issue #51205, opened 2026-07-09
Support FileResource sync on Proxy nodes


junjiejiangjjj:
## Motivation

FileResource is currently synchronized to execution nodes such as QueryNode through the existing FileResourceObserver flow. Proxy also needs access to the same FileResource files locally, but it is not currently part of the sync target set.

This causes Proxy to miss FileResource downloads even when the cluster-level FileResource metadata has been updated.

## Proposal

Add Proxy as a FileResource sync target by reusing the existing `fileresource.Manager` mechanism.

The change should:

- Add a Proxy-side FileResource manager initialization path.
- Add an internal `SyncFileResource` RPC on Proxy.
- Wire Proxy clients into `FileResourceObserver`.
- Let RootCoord inject the existing `proxyClientManager` into the observer.
- Add a `common.fileResource.mode.proxyNode` config item.
- Keep the behavior aligned with existing QueryNode FileResource sync semantics.

## Scope

This issue only covers enabling Proxy to reuse the existing FileResource sync/download mechanism.

It does not include PyUDF-specific validation, runtime behavior, resolver logic, or resource filtering.

## Acceptance Criteria

- Proxy can initialize a FileResource manager in sync mode.
- RootCoord/FileResourceObserver can push FileResource updates to Proxy nodes.
- Proxy downloads FileResource files through the existing `fileresource.Sync` path.
- Existing QueryNode/DataNode FileResource behavior is unchanged.
- Tests cover Proxy manager initialization and observer sync to Proxy.

PR #51206, opened 2026-07-09
enhance: support file resource sync on proxy

junjiejiangjjj:
issue: https://github.com/milvus-io/milvus/issues/51205
  
Add Proxy as a FileResource sync target by wiring it into the existing
  FileResourceObserver flow. Proxy now initializes a fileresource manager,
  exposes SyncFileResource over internal RPC, and RootCoord injects the
  proxy client manager so resource updates can be pushed to Proxy nodes.

PR #52174, opened 2026-08-04
enhance: support file resource sync on proxy 

junjiejiangjjj:
issue: #51205

Add Proxy as a FileResource sync target by wiring it into the existing FileResourceObserver flow. Proxy now initializes a fileresource manager, exposes SyncFileResource over internal RPC, and RootCoord injects the proxy client manager so resource updates can be pushed to Proxy nodes.


internal/proxy/proxy.go
```diff
@@ -187,6 +190,23 @@ func (node *Proxy) initRateCollector() error {
 
 // Init initialize proxy.
 func (node *Proxy) Init() error {
+	fileMode := fileresource.GetLocalMode()
+	if fileMode == fileresource.SyncMode {
+		if node.factory == nil {
+			return merr.WrapErrServiceInternalMsg("proxy dependency factory is nil")
+		}
+		node.factory.Init(paramtable.Get())
+		chunkManager, err := node.factory.NewPersistentStorageChunkManager(node.ctx)
```
liliu-z, line 206: This PR defaults proxy file-resource sync to `sync` (common.fileResource.mode.proxy, component_param.go:1587) and, in that mode, proxy `Init()` calls `NewPersistentStorageChunkManager`, which for remote storage eagerly probes the bucket (`NewMinioClient` -> `BucketExists` with retry). If object storage is unreachable or misconfigured at startup, `Init()` returns the wrapped error and the proxy fails to start — a dependency it did not have before (the old `NewProxy` discarded the factory). Unlike the data node path, which degrades to CloseMode when no storage address is configured (data_node.go:202-204), the proxy has no graceful fallback. Default proxy sync off, or degrade gracefully when storage is unavailable.

internal/util/fileresource/manager.go
```diff
@@ -220,41 +340,85 @@ func (m *SyncManager) Sync(version uint64, resourceList []*internalpb.FileResour
...
+		m.refManager.Close()
+	}
+}
+
 func (m *SyncManager) Mode() Mode { return SyncMode }
 
 func NewSyncManager(downloader storage.ChunkManager) *SyncManager {
+	baseManager := newBaseManager()
+	refManager := &RefManager{
+		BaseManager: baseManager,
```
liliu-z: `NewSyncManager` now embeds a `RefManager` that shares the same `localPath` as the `SyncManager` (`BaseManager: baseManager`). `SyncManager.Sync` installs files at `localPath/<id>`, while the `RefManager` installs and GC-deletes at `localPath/<RootPath>/<id>` (CleanResource -> `os.RemoveAll(path.Join(m.localPath, key))`). When the downloader's `RootPath()` cleans to `""` or `"."` (e.g. `minio.rootPath: "/"` or empty, which the formatter permits), the two paths collapse onto `localPath/<id>`, so the 15-minute RefManager GC removes files that `Sync` owns. This is reachable on a data node in sync mode, where compaction calls `GlobalFileManager.Download`/`Release`. Give the RefManager and sync trees distinct subdirectories so the paths can't collapse.

internal/rootcoord/root_coord.go
```diff
@@ -228,6 +228,10 @@ func (c *Core) GetQuotaCenter() *QuotaCenter {
 	return c.quotaCenter
 }
 
+func (c *Core) GetProxyClientManager() proxyutil.ProxyClientManagerInterface {
```
liliu-z, line 231: QueryNode (querycoordv2/server.go:683) and DataNode (datacoord/server.go:872) call `fileResourceObserver.Notify()` on `SessionAddEvent`, but the proxy registration path in rootcoord (`proxyWatcher.AddSessionFunc(..., c.proxyClientManager.AddProxyClient)`, root_coord.go:506) has no equivalent hook, and the observer only `Notify()`s on startup and on file-resource DDL. The proxy `SyncManager` starts at version 0 with no startup pull, so a proxy that joins or restarts after coordinator startup is never pushed the current resources until the next DDL, and proxy-local consumers (e.g. the XGBoost function cache) fail with `file resource not found` indefinitely. Add a `Notify()` on the proxy add-session path (and consider including proxies in a readiness gate).

internal/datanode/data_node.go
```diff
@@ -196,7 +196,7 @@ func (node *DataNode) Init() error {
 		syncMgr := syncmgr.NewSyncManager(nil)
 		node.syncMgr = syncMgr
 
-		fileMode := fileresource.ParseMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue())
+		fileMode := fileresource.GetLocalMode()
```
liliu-z, line 199: The PR made `InitManager` return an error (including a conflict when re-initialized with a different mode) and updated QueryNode (initError), StreamingNode (panic), and Proxy (return err) to handle it — but the data node's three `InitManager` calls (data_node.go:204/212/215) discard the returned error. In a mixed/standalone process, when the data node hits the no-storage-address branch it calls `InitManager(nil, CloseMode)` while a co-located QueryNode/StreamingNode calls `InitManager(cm, SyncMode)`; whichever runs first wins, and if the data node wins, the other component's `InitManager` returns the mode-conflict error and fails fatally (QueryNode initError -> waitForAllComponentsReady fails, or StreamingNode panic), crashing the process. Handle the data node InitManager errors and make the process-resolved mode authoritative so co-located components cannot install conflicting modes.

internal/util/fileresource/manager.go
```diff
@@ -220,41 +340,85 @@ func (m *SyncManager) Sync(version uint64, resourceList []*internalpb.FileResour
 		})
 	}
 
-	for name, id := range m.resourceMap {
-		if newId, ok := newResourceMap[name]; ok {
-			if newId != id {
-				// remove old resource with same name
-				removes = append(removes, id)
+	if err := analyzer.UpdateGlobalResourceInfo(newResourceMap); err != nil {
```
liliu-z, line 293: In `SyncManager.Sync`, `analyzer.UpdateGlobalResourceInfo(newResourceMap)` (manager.go:343) publishes the new name->id map to the global tokenizer state before the staging->final `os.Rename` (manager.go:360) that puts files in place. Between publish and rename a concurrent analyzer/consumer that resolves a newly-added id to `localPath/<id>` finds no file (the rename is atomic, so this is a missing-file window, not a partial read). If a later rename in the loop fails, the function returns without advancing the version and without rolling back the analyzer state, and the deferred `cleanupStaged()` deletes the staged files — so the analyzer keeps pointing at ids whose files do not exist until a subsequent successful sync. Publish to the analyzer only after all renames commit, mirroring how `notifyListeners` is already called last.

internal/proxy/impl.go
```diff
@@ -123,6 +124,24 @@ func (node *Proxy) GetStatisticsChannel(ctx context.Context, req *internalpb.Get
 	}, nil
 }
 
+func (node *Proxy) SyncFileResource(ctx context.Context, req *internalpb.SyncFileResourceRequest) (*commonpb.Status, error) {
+	if err := merr.CheckHealthy(node.GetStateCode()); err != nil {
+		return merr.Status(err), nil
+	}
+
+	if err := fileresource.Sync(context.TODO(), req.GetVersion(), req.GetResources()); err != nil {
```
liliu-z, line 132: The three file-resource RPC handlers pass `context.TODO()` to `fileresource.Sync` instead of the incoming request `ctx`: proxy/impl.go:132, datanode/services.go:980, and querynodev2/services.go:1921. `Sync` now uses the ctx both to wait for manager initialization (`case <-ctx.Done()`) and to bound each per-file download (`context.WithTimeout(ctx, downloadTimeout)`), so dropping it discards the caller's cancellation/deadline and the trace/log context, and violates the repo convention of threading the real function-parameter ctx. Pass the request `ctx` through all three handlers.

internal/proxy/proxy.go
junjiejiangjjj: Proxy FileResource intentionally defaults to `sync`, as Proxy is a FileResource consumer for XGBoost reranking and future PyUDF support. Defaulting to `close` would make these features unavailable under the default configuration.

internal/util/fileresource/manager.go
junjiejiangjjj: Fixed. `SyncManager` no longer embeds `RefManager`. The resolved mode now creates only one manager implementation, so RefManager GC cannot remove SyncManager-owned files.

internal/rootcoord/root_coord.go
junjiejiangjjj: Fixed. RootCoord now calls `fileResourceObserver.Notify()` after adding or reconciling Proxy clients, so newly joined, restarted, or rediscovered Proxies receive the current snapshot.

internal/datanode/data_node.go
junjiejiangjjj: Restored the original `sync.Once` initialization behavior. This PR now only changes components to read the process-resolved mode and no longer introduces mode-conflict errors.

internal/util/fileresource/manager.go
junjiejiangjjj: Fixed. The Analyzer resource map is now updated only after all staged files are activated. Old resources and manager state remain unchanged if activation or Analyzer update fails.

internal/proxy/impl.go
junjiejiangjjj: This preserves the existing behavior. Before this PR, FileResource downloads used `context.Background()` and did not inherit the RPC context. The new context is used to apply the configured per-file download timeout consistently across all three components.

internal/proxy/proxy.go
```diff
@@ -187,6 +190,21 @@ func (node *Proxy) initRateCollector() error {
...
+		if err != nil {
+			return merr.Wrap(err, "initialize Proxy file resource storage")
+		}
+		fileresource.InitManager(chunkManager, fileMode)
```
liliu-z, line 210: Proxy calls fileresource.InitManager before node.initSession, so the manager's localPath is built from paramtable.GetNodeID() while it is still 0 (initSession assigns the real ID). QueryNode and DataNode initialize the manager after their session, so only proxy bakes in node 0 and downloads to <root>/cache/0/file_resource instead of <root>/cache/<proxyID>/file_resource. Compounding this, mixCoord.IsServerActive — used by the POSIX-mode cache cleanup checkExpiredPOSIXDIR — only checks queryCoord/datacoord sessions, so a proxy's <root>/cache/<id> directory is never seen as active and can be removed. Move InitManager after initSession and include proxy sessions in IsServerActive.

internal/util/fileresource/manager.go
```diff
@@ -116,20 +131,116 @@ const (
...
+	defer cancel()
+
+	size, err := downloader.Size(downloadCtx, resource.GetPath())
+	if err != nil {
+		return merr.Wrapf(err, "get file resource size for %s", resource.GetPath())
+	}
+	if size < 0 {
+		return merr.WrapErrIoFailedMsg("file resource %s reports negative size %d", resource.GetPath(), size)
+	}
+	if size > m.maxFileSize {
```
liliu-z: downloadFile now rejects any resource larger than maxFileSize (default 1 GiB) on every sync node (QueryNode/DataNode/proxy), whereas the pre-PR SyncManager used an unbounded io.Copy. A >1 GiB resource that synced before this change will fail with ErrIoEntityTooLarge after upgrade. Note the limit is already configurable via common.fileResource.maxFileSize, so the reviewer's 'make it configurable' is unnecessary — the real risk is a silent default-driven regression, which should be called out in the upgrade notes.

internal/proxy/proxy.go
junjiejiangjjj: Fixed. Proxy now initializes the file-resource manager after `initSession`, so the manager uses the actual proxy node ID. `mixCoord.IsServerActive` also checks proxy sessions through `rootcoordServer.ServerExist`.

internal/util/fileresource/manager.go
junjiejiangjjj: The default has been changed to `0`, which disables the file-size limit. The limit is applied only when `common.fileResource.maxFileSize > 0`, so existing resources larger than 1 GiB remain supported by default and no upgrade behavior change is introduced.

internal/datanode/data_node.go
liliu-z, line 199: The DataNode builds its file-resource manager from the process-wide resolved mode (`fileresource.GetLocalMode()`), but DataCoord decides whether to ship file resources for ref-style download using the DataNode's own `DNFileResourceMode` (compaction_task_mix.go:375, stats_inspector.go:225). In a co-located process where `DNFileResourceMode=ref` while another role forces the resolved mode to sync (e.g. a default-sync QueryNode in standalone), the node installs a SyncManager whose `Download` is a no-op, yet DataCoord still populates `plan.FileResources` and the compaction/stats task resolves the dictionary at the ref-style `<base>/<storage_name>/<id>/<file>` path that the SyncManager never creates. Custom-dictionary text indexing on that DataNode then fails to locate the files. Drive the node manager mode and the DataCoord task-planning gate from a single mode source.

internal/util/fileresource/manager.go
```diff
@@ -116,20 +131,121 @@ const (
...
+	defer cancel()
+
+	size, err := downloader.Size(downloadCtx, resource.GetPath())
+	if err != nil {
+		return merr.Wrapf(err, "get file resource size for %s", resource.GetPath())
+	}
+	if size < 0 {
+		return merr.WrapErrIoFailedMsg("file resource %s reports negative size %d", resource.GetPath(), size)
+	}
+	if m.maxFileSize > 0 && size > m.maxFileSize {
```
liliu-z: An oversized resource is not written into any committed snapshot — `downloadFile` rejects it at manager.go:178 with ErrIoEntityTooLarge and `SyncManager.Sync` returns before renaming files into place, advancing the version, updating `resourceMap`, or notifying listeners (verified by TestSync_SizeLimit). The real problem is that this rejection fails the whole sync version atomically while the oversized resource remains in committed RootCoord meta (AddFileResource only checks path existence, not size, before persisting): every sync node then re-rejects that version on each retry, so the file-resource version can never advance and no update — including unrelated resources bundled in the same version — propagates until an admin manually removes the offending resource. Only triggered when maxFileSize>0 (non-default). Validate size before persisting to meta, or skip-and-continue past an over-limit resource instead of failing the whole version.

internal/datanode/data_node.go
junjiejiangjjj: Fixed. DataCoord now always passes the required `FileResources` without checking the DataNode mode. The DataNode uses its actual manager mode to decide how to handle them: `RefMode` downloads per task, while `SyncMode` uses the resources already synchronized locally. Tests were added for both compaction and stats paths.

internal/util/fileresource/manager.go
junjiejiangjjj: Fixed. RootCoord now validates the file size before persisting the resource, so an oversized file cannot block subsequent resource synchronization. The node-side size check has also been removed. Thanks!

internal/util/proxyutil/proxy_client_manager.go
```diff
@@ -302,6 +303,21 @@ func (p *ProxyClientManager) RefreshPolicyInfoCache(ctx context.Context, req *pr
 	return group.Wait()
 }
 
+func (p *ProxyClientManager) SyncFileResource(ctx context.Context, request *internalpb.SyncFileResourceRequest) map[int64]error {
+	results := make(map[int64]error)
+	p.proxyClient.Range(func(proxyID int64, client types.ProxyClient) bool {
```
liliu-z: SyncFileResource iterates proxy clients and issues each RPC inline, so total latency is the sum over the fleet and one hung proxy blocks delivery to every proxy behind it. Every sibling broadcast in this same file already uses errgroup — UpdateCredentialCache at :263 and RefreshPolicyInfoCache at :288 — so this is a departure from the established pattern in the file it lives in. It matters because the caller holds FileResourceObserver.syncMu and RootCoord's AddFileResource/RemoveFileResource invoke Sync() inline on the DDL RPC path, so the serial chain stalls the user-facing DDL. The path in the original report was wrong — this is internal/util/proxyutil/proxy_client_manager.go; internal/distributed/proxy/client_manager/ does not exist.

internal/util/fileresource/manager.go
```diff
@@ -147,71 +249,53 @@ func (m *SyncManager) GetVersion() uint64 {
...
+				Path:      resource.GetPath(),
+				LocalPath: localFilePath,
+			})
+			continue
 		}
 
-		err = os.MkdirAll(localResourcePath, os.ModePerm)
-		if err != nil {
-			return err
+		stagingResourcePath := fmt.Sprintf("%s.downloading-%d", localResourcePath, version)
```
liliu-z: `stagingResourcePath := fmt.Sprintf("%s.downloading-%d", localResourcePath, version)` is cleaned by `os.RemoveAll` keyed on the current version, and a restarted node normally resyncs at the same version so the exact `<id>.downloading-<V>` directory is removed. But if metadata advances while the node is down, the resync arrives at V+1 and the old V staging directory is never matched again, leaving a partial download on local disk that accumulates across such restarts. Sweep the parent directory for any `*.downloading-*` prefix match rather than only the current version's exact path.

internal/coordinator/file_resource_observer.go
```diff
@@ -263,12 +266,36 @@ func (m *FileResourceObserver) Sync() error {
...
 		}
 
-		nodeIDs = append(nodeIDs, dnnodes...)
+		for _, nodeID := range dnnodes {
+			activeNodes[nodeID] = struct{}{}
+		}
+	}
+
+	// sync file resource to proxy if file resource mode was Sync
+	if m.proxyMode == fileresource.SyncMode && m.proxyManager != nil {
```
liliu-z, line 278: The new proxy branch omits the per-node version guard that both sibling branches have, so it re-broadcasts the full resource list to every proxy on every Sync rather than only to proxies behind targetVersion. Combined with the retry loop this compounds: any failing proxy leaves syncErr non-nil, syncLoop calls RetryNotify, and 3 seconds later the whole fleet is re-broadcast — serially, since ProxyClientManager.SyncFileResource has no fan-out — for as long as that one proxy stays unreachable. The QueryNode and DataNode branches skip already-synced nodes and are unaffected. Add the same `if info, ok := m.distribution.Get(nodeID); !ok || info.Version < targetVersion` guard on the proxy path.

internal/rootcoord/root_coord.go
```diff
@@ -228,6 +228,37 @@ func (c *Core) GetQuotaCenter() *QuotaCenter {
...
+	return c.proxyClientManager
+}
+
+func (c *Core) ServerExist(serverID int64) bool {
+	sessions, _, err := c.session.GetSessions(c.ctx, typeutil.ProxyRole)
+	if err != nil {
+		mlog.Warn(c.ctx, "failed to get sessions", mlog.Err(err))
+		return false
```
liliu-z, line 239: Core.ServerExist returns false when the etcd session lookup fails, and mixCoordImpl.IsServerActive now consults it, so a transient etcd error during the POSIX cleanup sweep causes checkExpiredPOSIXDIR to os.RemoveAll a live proxy's entire node cache directory — which under this PR is where its downloaded file resources live. Recovery does not happen on its own: SyncManager.Sync early-returns on `version <= m.version.Load()`, so the in-memory version still matches and the deleted files are never re-fetched until a new resource version lands or the proxy restarts. The sweep runs every minute and does an uncached GetSessions(ProxyRole) per candidate directory. Treat a lookup error as 'unknown, keep' rather than 'dead, delete'.

internal/util/proxyutil/proxy_client_manager.go
junjiejiangjjj: QueryNode and DataNode file-resource synchronization in `FileResourceObserver.Sync` also issues RPCs serially. This PR follows the existing file-resource synchronization model for Proxy so that the three node types keep consistent behavior. Parallelizing the fan-out should be handled as a separate cross-component optimization covering QueryNode, DataNode, and Proxy together. We will add the missing per-node version guard in this PR so retries do not resend the resource list to already-synced proxies.

internal/util/fileresource/manager.go
junjiejiangjjj: Good catch. The versioned staging directory is introduced by this PR, and cleaning only the current version can leave an older partial directory after a restart. We will update the cleanup to sweep stale `*.downloading-*` directories, including staging directories for resources no longer present in the current metadata, and add test coverage.

internal/coordinator/file_resource_observer.go
junjiejiangjjj: Agreed. Proxy should use the same per-node version guard as the QueryNode and DataNode branches. We will filter the target proxies using `distribution` before sending RPCs, while still keeping every live proxy in `activeNodes`, so an already-synced proxy is neither called again nor removed from the distribution during cleanup.

internal/rootcoord/root_coord.go
junjiejiangjjj: The failure mode is valid. However, the `ServerExist bool` behavior predates this PR: QueryCoord and DataCoord also return false when `GetSessions` fails, and the POSIX cleanup already treats that as inactive. This PR extends the same existing behavior to Proxy. To keep the node types consistent, we do not plan to change only the Proxy lookup semantics here. Distinguishing “not found” from “lookup failed” should be handled as a separate cross-component cleanup fix covering QueryNode, DataNode, and Proxy together.

cmd/roles/roles.go
```diff
@@ -482,6 +498,10 @@ func (mr *MilvusRoles) Run() {
 		}()
 	}
 
+	effectiveFileResourceMode := mr.resolveFileResourceMode()
```
liliu-z, line 501: With this PR a Proxy-only process starts a file-resource manager that writes into `<localStorage>/cache/<serverID>/file_resource/`, but nothing ever reclaims those directories under the default `common.enablePosixMode: false`. `runQueryNode` has a `!EnablePosixMode`-guarded `cleanLocalDir(RootCachePath)`; `runProxy` has none, and the POSIX-mode reclaimer this PR extended (`rootcoordServer.ServerExist` in `IsServerActive`) returns early when POSIX mode is off. Each Proxy restart allocates a fresh ServerID and a fresh subdirectory, so on any persistent local-storage mount (bare metal, docker-compose named volume, PVC) the cache grows without bound; k8s emptyDir is unaffected. Note that simply copying `cleanLocalDir` into `runProxy` is wrong — `runComponent` starts its future immediately and standalone shares one nodeID across roles, so the Proxy wipe would delete the directory the concurrently-starting QueryNode is downloading into. Hoist one guarded wipe next to `fileresource.SetLocalMode` here, before any component starts, and drop the one in `runQueryNode`.

junjiejiangjjj: The cleanup gap already exists for DataNode-only deployments. We will address it in a follow-up PR with unified process-level cleanup for all roles, rather than handling Proxy separately in this PR.

pkg/util/paramtable/component_param.go
```diff
@@ -1578,6 +1581,69 @@ If enabled, IPv6 ULA/global addresses will be prioritized ahead of IPv4.`,
 	}
 	p.DNFileResourceMode.Init(base.mgr)
 
+	p.ProxyFileResourceMode = ParamItem{
+		Key:          "common.fileResource.mode.proxy",
```
liliu-z, line 1585: #51205 and #51206 both use `common.fileResource.mode.proxyNode`, but this registers `common.fileResource.mode.proxy` with no fallback. A deployment using `proxyNode: close` silently keeps the default `sync` mode and still takes the object-storage startup path. Please keep the specified key, or accept both keys during migration.

internal/proxy/proxy.go
liliu-z: The merged XGBoost path is still L0-only and its design explicitly defers Proxy/L2; PyUDF is still draft. So this default adds a new object-storage startup dependency before any merged Proxy consumer uses it. Could we default it off or enable it only with a Proxy-side consumer?

internal/rootcoord/root_coord.go
liliu-z: The fact that QN/DN already have this behavior does not make the new Proxy path safe. A later metadata version does not heal it either: the name/ID fast path in `SyncManager.Sync` reuses `LocalPath` without checking the file, so an unchanged resource stays missing until Proxy restart or resource replacement. Can we make liveness tri-state here, or re-download missing files before adding Proxy to this sweeper?

cmd/roles/roles.go
liliu-z: `emptyDir` survives container restarts inside the same Pod, so crash loops leak there too until the Pod is replaced. With Proxy sync enabled by default, every restart can leave another full snapshot under a new ServerID. I do not think process-level cleanup is safe to defer.

pkg/util/paramtable/component_param.go
junjiejiangjjj: Milvus names this component `Proxy`, not `ProxyNode`, so `common.fileResource.mode.proxy` is the intended key. It is also consistent with the key added to `configs/milvus.yaml`.

The `proxyNode` key only appeared in the closed, unmerged PRs and has never been part of a released configuration, so there is no existing configuration to migrate or preserve.

internal/rootcoord/root_coord.go
junjiejiangjjj: Thanks for the clarification. I agree that the failure and recovery gap is valid, but it is a common FileResource issue rather than a Proxy-specific one. QueryNode and DataNode already use the same boolean liveness semantics, where a session lookup error is treated as inactive, and all sync-mode consumers share the same `SyncManager` behavior that reuses an unchanged name/ID without verifying that the local file still exists.

This PR extends those existing behaviors to Proxy, but fixing only the Proxy branch would leave inconsistent cleanup and recovery semantics across FileResource consumers. I prefer to address this in a separate PR that fixes the full path consistently for QueryNode, DataNode, and Proxy: preserve an unknown liveness state instead of treating it as dead, delete cache directories only after absence is confirmed, and add a common repair path for missing local files.

liliu-z (commented):
/lgtm
/approve

liliu-z (commented):
/lgtm

Issue #52178, opened 2026-08-05
[Bug]: Compaction workers swallow writer errors and crash/hang on abnormal inputs, allowing silent row loss


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```
- Milvus version: master
- Deployment mode: any
```

### Current Behavior

Compaction workers ignore several writer-side errors, so a task can report `completed` (or wedge) after silently corrupting its output. Four defects, one shared contract violation — a worker-side writer error/panic must make the task terminally fail, never continue silently:

**1. Mix compaction discards `RecordBuilder.Append` errors** (`internal/datanode/compactor/mix_compactor.go:372` and `:385`)

After delete/TTL filtering, retained rows are rebuilt through `storage.RecordBuilder`, but both `rb.Append(...)` call sites drop the returned error. `Append` is not atomic: on failure `nRows` is not advanced and values already appended are not rolled back. One swallowed failure means every retained row of that record batch is silently dropped while batches without filtered rows are still written — the task completes with a NON-EMPTY result, so the empty-result guard on the DataCoord side never fires, the result is adopted and the input segments are dropped.

The trigger is deterministic, not transient: `appendValueAt` (`internal/storage/arrow_util.go`) can only fail on type mismatches ("invalid value type", "unsupported builder type"). Any Arrow representation divergence — e.g. a TEXT column read back as `String` while the rebuilder allocates a `Binary` builder — reproduces it on every compaction of the affected segment. Reproduced in a regression test with real V1 binlogs.

`NamespaceCompactor` reuses the same path and inherits the defect.

**2. Clustering compaction discards record-boundary `Flush` errors** (`internal/datanode/compactor/clustering_compactor.go:703-705`)

Cluster buffers are force-flushed after each input record, error ignored. `SerializeWriterImpl.Flush` keeps its buffer on failure and the V1 composite writer has no per-field rollback, so continuing replays the retained batch into fields that already persisted it — per-field files fork (verified 2:1 write divergence under fault injection) while the row-count metadata stays consistent, masking the corruption from count checks.

**3. MergeSort infers comparators from `recs[0]`** (`internal/storage/sort.go`) — a first reader that yields zero physical batches (metadata/physical divergence produced by other bugs) causes a nil dereference panic; position-sensitive, looks flaky.

[...]

PR #52179, opened 2026-08-05
fix: propagate swallowed RecordBuilder.Append errors in mix compaction

MrPresent-Han:
## What this PR does

Mix compaction discarded both `RecordBuilder.Append` errors on the retained-row rebuild path (`mix_compactor.go`).

`Append` is not atomic (no rollback of already-appended values; `nRows` only advances on full success) and its only failure modes are deterministic type mismatches, so any Arrow representation divergence made every compaction of the affected segment silently drop the retained rows of each delete-containing batch — while batches without deletes were still written, producing a **non-empty completed result** that the DataCoord empty-result guard cannot catch: the result was adopted and the input segments dropped. Both call sites now propagate the error and release the poisoned builder (never reaching `Build`); the builder is also released on the normal and writer-error paths, closing an adjacent lifetime leak. `NamespaceCompactor` inherits the fix.

## Not in this PR (tracked in the issue)

Scope is deliberately frozen to the confirmed Mix silent-loss defect. Moved to follow-ups per the issue's out-of-scope list:

- clustering compaction's record-boundary `Flush` swallow and the failure-containment stack it requires (buffer poisoning, sibling latch/drain, in-flight I/O cancellation, causal-error preservation) — an implementation is archived and can seed a dedicated PR;
- MergeSort `recs[0]` nil-guard and executor panic containment (PR-B);
- generic writer Abort/Commit lifecycle; row-conservation result contract;
- legacy V1 `SegmentWriter` swallow sites (only production exposure is the legacy V1 stats-task path).

## Tests

- `TestMixCompactionPropagatesRecordBuilderAppendError` — real V1 binlogs with a TEXT field (deserialized as arrow `String`) against the production `Binary` rebuilder, so the mismatch comes from the production reader itself; two subcases position the filtered row to exercise both the mid-record and the tail `Append` call sites; each asserts the task fails and returns no adoptable result. Red on the previous code (which completed silently).
- Full `internal/datanode/compactor` suite passes with `-tags dynamic,test -gcflags="all=-N -l" -count=1`.

issue: #52178

### DIFF
--- a/internal/datanode/compactor/mix_compactor.go
+++ b/internal/datanode/compactor/mix_compactor.go
@@ -369,7 +369,13 @@ func (t *mixCompactionTask) writeSegment(ctx context.Context,
 					rb = storage.NewRecordBuilder(writerSchema)
 				}
 				if sliceStart != -1 {
-					rb.Append(r, sliceStart, i)
+					// Append is not atomic: a failed builder may hold a partial
+					// row and must never reach Build.
+					if err = rb.Append(r, sliceStart, i); err != nil {
+						rb.Release()
+						cleanupMaterializedRecord(r)
+						return 0, 0, err
+					}
 				}
 				sliceStart = -1
 				continue
@@ -382,7 +388,11 @@ func (t *mixCompactionTask) writeSegment(ctx context.Context,
 
 		if rb != nil {
 			if sliceStart != -1 {
-				rb.Append(r, sliceStart, r.Len())
+				if err = rb.Append(r, sliceStart, r.Len()); err != nil {
+					rb.Release()
+					cleanupMaterializedRecord(r)
+					return 0, 0, err
+				}
 			}
 			if rb.GetRowNum() > 0 {
 				err := func() error {
@@ -395,10 +405,12 @@ func (t *mixCompactionTask) writeSegment(ctx context.Context,
 					return mWriter.Write(out)
 				}()
 				if err != nil {
+					rb.Release()
 					cleanupMaterializedRecord(r)
 					return 0, 0, err
 				}
 			}
+			rb.Release()
 		} else {
 			out := overwriteRecordTimestamps(r, seg.GetCommitTimestamp())
 			err := mWriter.Write(out)
--- a/internal/datanode/compactor/mix_compactor_test.go
+++ b/internal/datanode/compactor/mix_compactor_test.go
@@ -192,6 +192,104 @@ func TestMixCompactionMaterializesMissingFieldWithDeleteFilter(t *testing.T) {
 	s.EqualValues(1, fieldBinlogEntriesForTest(segment.GetInsertLogs(), StringField))
 }
 
+const textFieldForAppendErrorTest = int64(130)
+
+func (s *MixCompactionTaskStorageV1Suite) setupTestWithTextField() {
+	s.mockBinlogIO = mock_util.NewMockBinlogIO(s.T())
+	s.meta = genTestCollectionMeta()
+	s.meta.Schema.Fields = append(s.meta.Schema.Fields, &schemapb.FieldSchema{
+		FieldID:  textFieldForAppendErrorTest,
+		Name:     "text_col",
+		DataType: schemapb.DataType_Text,
+		Nullable: true,
+	})
+	params, err := compaction.GenerateJSONParams(s.meta.GetSchema())
+	s.Require().NoError(err)
+	plan := &datapb.CompactionPlan{
+		PlanID:                 999,
+		Type:                   datapb.CompactionType_MixCompaction,
+		Schema:                 s.meta.GetSchema(),
+		PreAllocatedSegmentIDs: &datapb.IDRange{Begin: 19531, End: math.MaxInt64},
+		PreAllocatedLogIDs:     &datapb.IDRange{Begin: 9530, End: 19530},
+		MaxSize:                64 * 1024 * 1024,
+		JsonParams:             params,
+	}
+	pk, err := typeutil.GetPrimaryFieldSchema(s.meta.GetSchema())
+	s.Require().NoError(err)
+	s.task = NewMixCompactionTask(context.Background(), s.mockBinlogIO, nil, plan, compaction.GenParams(), []int64{pk.FieldID})
+}
+
+// V1 binlogs deserialize TEXT as arrow String while the retained-row rebuilder
+// allocates a Binary builder for TEXT, so with any filtered row rb.Append fails
+// deterministically: the task must fail instead of silently dropping the
+// retained rows of that batch.
+func TestMixCompactionPropagatesRecordBuilderAppendError(t *testing.T) {
+	cases := []struct {
+		name           string
+		deletePKOffset int64
+	}{
+		// deleting the first row leaves the retained range open at record end,
+		// exercising the tail rb.Append; deleting the second row closes the
+		// retained range mid-record, exercising the in-loop rb.Append.
+		{name: "filtered_first_row_hits_tail_append", deletePKOffset: 0},
+		{name: "filtered_second_row_hits_mid_append", deletePKOffset: 1},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			s := newMixCompactionStorageV1SuiteForDirectTest(t)
+			s.setupTestWithTextField()
+
+			segmentID := int64(700)
+			alloc := allocator.NewLocalAllocator(888888, math.MaxInt64)
+			segWriter, err := NewSegmentWriter(s.meta.GetSchema(), 65535, compactionBatchSize, segmentID, PartitionID, CollectionID, []int64{})
+			s.Require().NoError(err)
+			for i := int64(0); i < 2; i++ {
+				row := getRow(segmentID+i, 0)
+				row[textFieldForAppendErrorTest] = "text-payload"
+				err = segWriter.Write(&storage.Value{
+					PK:        storage.NewInt64PrimaryKey(segmentID + i),
+					Timestamp: int64(tsoutil.ComposeTSByTime(getMilvusBirthday())),
+					Value:     row,
+				})
+				s.Require().NoError(err)
+			}
+			segWriter.FlushAndIsFull()
+			s.segWriter = segWriter
+
+			kvs, fBinlogs, err := serializeWrite(context.TODO(), alloc, s.segWriter)
+			s.Require().NoError(err)
+
+			deleteTs := tsoutil.ComposeTSByTime(getMilvusBirthday().Add(10 * time.Second))
+			blob, err := getInt64DeltaBlobs(segmentID, []int64{segmentID + tc.deletePKOffset}, []uint64{deleteTs})
+			s.Require().NoError(err)
+			deltaPath := "deltalog/append-error-" + tc.name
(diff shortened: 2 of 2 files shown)


internal/datanode/compactor/clustering_compactor.go
```diff
@@ -700,8 +712,8 @@ func (t *clusteringCompactionTask) mappingSegment(
 		}
 
 		// all cluster buffers are flushed for a certain record, since the values read from the same record are references instead of copies
-		for _, buffer := range t.clusterBuffers {
-			buffer.Flush()
+		if err := t.flushClusterBuffersForRecord(); err != nil {
```
buqian-zilliz: **[P2] Stop sibling mapping workers after a flush failure**

`mappingSegment` only returns from the worker that observes this error. `mapping()` has already submitted all input segments, and `conc.AwaitAll` returns at the first errored future without cancelling or draining the rest. Those workers share `t.clusterBuffers`, while `SerializeWriterImpl.Flush` leaves `pos` and the buffered values unchanged on error. A sibling can therefore append to or retry the poisoned batch; a transient retry can replay partially written fields, and if the error persists one sibling can leave `pos == batchSize`, causing the next sibling's `WriteValue` to index `buffer[batchSize]` and panic. Latch this as a task-wide or buffer-wide failure, stop sibling writes, and wait for all mapping futures before cleanup.

MrPresent-Han: Good catch — all three legs verified and fixed in the amended commit: the first mapping failure is now latched task-wide (`mappingFailed`) so sibling workers stop at their record boundary; `mapping()` drains every worker via `conc.BlockOnAll` before returning to cleanup; and `SerializeWriterImpl.WriteValue` now rejects writes into a buffer retained by a failed flush instead of indexing past it — which also kills the pre-existing self-panic route (a persistently failing flush would eventually index `buffer[batchSize]` even single-threaded, before this PR). Deterministic concurrency UT for the latch is hard to build stably, so the guard + drain serve as the double insurance here.

internal/datanode/compactor/clustering_compactor.go
```diff
@@ -507,11 +511,17 @@ func (t *clusteringCompactionTask) mapping(ctx context.Context,
...
+			}
 			return struct{}{}, err
 		})
 		futures = append(futures, future)
 	}
-	if err := conc.AwaitAll(futures...); err != nil {
+	// Drain every worker before returning: workers share the cluster buffers,
+	// so an early error return would let siblings keep writing into a possibly
+	// poisoned buffer while cleanup runs.
+	if err := conc.BlockOnAll(futures...); err != nil {
```
buqian-zilliz: **[P2] Drain nested flush futures before cleanup**

`BlockOnAll` only joins the mapping futures. `flushLargestBuffers` and `flushAll` still use `conc.AwaitAll`, which returns after the first errored future in iteration order, leaving later `FlushChunk` or `Close` jobs active. `mapping` can therefore return and deferred cleanup can release the flush pool while those jobs are still mutating or uploading from the shared writers, because the underlying pool's `Release` does not wait for running workers. Join every submitted flush future before propagating the error.

internal/datanode/compactor/clustering_compactor.go
```diff
@@ -632,6 +654,11 @@ func (t *clusteringCompactionTask) mappingSegment(
 
 	offset := int64(-1)
 	for {
+		if t.mappingFailed.Load() {
+			// A sibling worker failed; the shared buffers may hold a poisoned
+			// batch, so stop contributing instead of writing into them.
+			return merr.WrapErrServiceInternal("clustering mapping aborted: a sibling mapping worker failed")
```
buqian-zilliz: **[P2] Preserve the causal mapping error**

`mappingFailed` records only a boolean. If a later-submitted worker is the real failure, an earlier future can subsequently hit this branch and return the synthetic service-internal error. `BlockOnAll` keeps the first error by future slice order, so that sibling-abort error can replace the actual download, deserialization, or flush error returned from `mapping`, leaving the task without the causal failure needed for diagnosis. Preserve the first real error and return it after draining all workers.

internal/datanode/compactor/clustering_compactor.go
buqian-zilliz: This issue still exists: `mappingFailed` is set only after the failing `mappingSegment` returns and is checked only before the next record, so a sibling already processing a record or waiting on the buffer lock can still append to the poisoned buffer and invoke `Flush` again. A failed record-boundary flush can retain fewer than `batchSize` values, so the new bounds check does not reject that retry. Poison the buffer under its lock, or otherwise reject every subsequent write and flush immediately after the first failure.

MrPresent-Han: Fixed in the amended commit by moving the poison down to the resource: `SerializeWriterImpl` now latches `failedErr` on the first failed flush and rejects every subsequent `WriteValue`/`Flush` — this covers the partial-buffer (`pos < batchSize`) retry the bounds check missed, and any mid-record sibling that was already past the record-boundary check (all access goes through `ClusterBuffer`'s lock, so the flag flips race-free). The task-wide latch is now just best-effort early-stop. Added `TestSerializeWriterStickyAfterFlushError` proving a partial-buffer flush failure never replays into the underlying writer.

internal/datanode/compactor/clustering_compactor.go
MrPresent-Han: Fixed: `flushLargestBuffers` and `flushAll` now join every submitted future via `conc.BlockOnAll` before propagating the error, so by the time `cleanUp` releases the flush pool there are no running jobs left to race against.

internal/datanode/compactor/clustering_compactor.go
MrPresent-Han: Fixed: the first causal error is captured once (`mappingErrOnce`/`mappingErr`) by the failing worker before the latch flips, and `mapping()` returns it after draining all workers — a sibling-abort error from an earlier-indexed future can no longer mask the real download/deserialize/flush failure.

internal/datanode/compactor/clustering_compactor.go
```diff
@@ -782,7 +819,10 @@ func (t *clusteringCompactionTask) flushLargestBuffers(ctx context.Context) erro
 			break
 		}
 	}
-	if err := conc.AwaitAll(futures...); err != nil {
+	// Join every flush job before returning: pool Release does not wait for
+	// running workers, so an early error return would race cleanup against
+	// jobs still mutating the shared writers.
+	if err := conc.BlockOnAll(futures...); err != nil {
```
buqian-zilliz: **[P2] Poison buffers before draining nested FlushChunk failures**

`BlockOnAll` drains the nested jobs, but `mappingErr` and `mappingFailed` are not published until the enclosing `mappingSegment` returns. `MultiSegmentWriter.FlushChunk` can succeed in `SerializeWriter.Flush` and then fail in the direct `BinlogRecordWriter.FlushChunk` call, so `SerializeWriterImpl.failedErr` remains unset; for the V1 writer this can occur after the per-field writers have already been finalized. The `ClusterBuffer` lock is then released while this wait is still in progress, allowing a sibling to write to or retry the finalized writer and surface a secondary error before the original worker unwinds. Latch and poison the original `FlushChunk` error under the buffer lock, or publish it from the flush job before exposing the writer, while still draining all futures.

internal/storage/serde.go
```diff
@@ -1279,29 +1279,45 @@ type SerializeWriterImpl[T any] struct {
...
 	buffer []T
 	pos    int
+	// failedErr poisons the writer after the first failed flush: the buffer is
+	// retained and the underlying writer may hold a partial batch, so any
+	// further write or flush would replay it.
+	failedErr error
 }
 
 func (sw *SerializeWriterImpl[T]) Flush() error {
+	if sw.failedErr != nil {
```
buqian-zilliz: **[P2] Add an abort path for poisoned writers**

Once `failedErr` is set, this guard also makes `Close` return before `rw.Close`, because `Close` starts by calling `Flush`. Packed V2/V3 record writers own native writer handles that are released only through their close or destroy paths, while clustering cleanup only releases the worker pools. A failed write can therefore leak the native writer indefinitely when the failed task is discarded. Add an abort/discard path that releases the underlying writer without flushing or committing the retained batch, and invoke it on compaction failure.

internal/datanode/compactor/clustering_compactor.go
MrPresent-Han: Valid — the chunk-flush leg (`MultiSegmentWriter.FlushChunk`'s direct `BinlogRecordWriter.FlushChunk` call) is outside the serialize-layer poison, and the V1 per-field `Finalize` loop can leave a partially-finalized writer for a sibling to hit. That said, we are freezing this PR's scope at its original promise: swallowed errors are propagated and the task fails terminally. Buffer-level failure lifecycle (poisoning at the `ClusterBuffer` lock boundary) belongs together with the abort/discard path in the failure-lifecycle follow-up now listed in #52178's out-of-scope section — fixing it piecemeal here would keep growing this diff into a writer-lifecycle redesign.

internal/storage/serde.go
MrPresent-Han: Valid, but pre-existing rather than introduced here: `Close` has always started with `Flush`, so any flush error already skipped `rw.Close` before this PR — the sticky error only makes that deterministic for later attempts. A correct fix needs a non-committing Abort/Destroy on the writer contract (V1's `Close` is committing — finalizing per-field blobs on a failed task is exactly what must not happen), which is the generic writer-lifecycle work listed as out-of-scope in #52178; tracked there as a follow-up together with the buffer-level poison above.

internal/datanode/compactor/clustering_compactor.go
```diff
@@ -507,11 +515,23 @@ func (t *clusteringCompactionTask) mapping(ctx context.Context,
 		}
 		future := t.mappingPool.Submit(func() (any, error) {
 			err := t.mappingSegment(ctx, segmentClone)
+			if err != nil {
+				t.mappingErrOnce.Do(func() { t.mappingErr = err })
+				t.mappingFailed.Store(true)
```
buqian-zilliz: **[P2] Cancel sibling I/O before draining mapping futures**

`mappingFailed` is not polled until a worker has composed delete logs, downloaded vector offsets, constructed its reader, and reached a record boundary. This branch flips the flag but leaves the task context active; that context is rooted at the DataNode lifetime rather than the compaction RPC, and `BlockOnAll` then waits for every future. A sibling already blocked in download or reader I/O therefore keeps the compaction slot occupied until that operation finishes or times out. Cancel the task context when the first causal error is latched, stop submitting additional segments, and then drain the futures.

MrPresent-Han:
**Scope re-cut**: this PR is now narrowed to the confirmed Mix silent-row-loss defect only (2 files). The clustering record-boundary `Flush` swallow and the failure-containment stack it pulled in (buffer poisoning, sibling latch/drain, nested-future draining, causal-error preservation, in-flight I/O cancellation — threads above) have been moved out wholesale: fixing the flush swallow in a concurrent shared-buffer path turns out to require the full failure-lifecycle contract, which deserves its own design-first PR rather than growing this one. The removed implementation is archived and referenced from #52178's out-of-scope list; the review findings in the threads above remain valid against that future work. Apologies for the churn — earlier "fixed in the amended commit" replies on the clustering threads are superseded by this re-scope.

Issue #52189, opened 2026-08-05
[Enhancement]: QueryNode read scheduler rejects requery tasks when unsolved queue is full, wasting completed ANN computation


MrPresent-Han:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Admit requery tasks (the output-field fetch phase of Search/HybridSearch) through a dedicated bounded lane in the QueryNode read scheduler, dispatched ahead of the regular policy queue, instead of subjecting them to the shared `queryNode.scheduler.unsolvedQueueSize` admission check.

Proposed mechanism (kept deliberately small):
- Identify requery by the existing `QueryLabel="requery"` already carried end-to-end in `RetrieveRequest` (no proto change).
- Admission: a separate bounded quota `queryNode.scheduler.requeryUnsolvedQueueSize` (default 1024, `<=0` disables the lane and falls back to today's behavior).
- Dispatch: pop the requery lane first, then the policy queue. No changes inside the fifo / user-task-polling policies.
- The lane cannot starve regular tasks: requery is derived traffic — at most one requery per admitted search, so its arrival rate is capped by upstream search admission and auto-decreases under overload.

### Why is this needed?

When the unsolved queue is full, `handleAddTaskRequest` rejects all read tasks indiscriminately with `ErrTooManyRequests`. For a requery task this is the worst possible place to shed load: the parent request has already paid the expensive ANN phase, and rejecting the cheap PK-based fetch discards all of it. The client retry then re-runs the full ANN phase against an already overloaded cluster, amplifying the overload.

Observed in a production incident: a HybridSearch (12 shards, dense+BM25, topk=200) spent 9.03s completing both ANN passes successfully across all QueryNodes, then its requery phase was rejected 834 times across all retry layers (`limit by queryNode.scheduler.unsolvedQueueSize: too many concurrent requests, queue is full[limit=1024]`) and the whole 12.87s request failed — 100% of the completed ANN work was wasted.

Related: #49707 / PR #50454 add proxy-side backpressure and faster load shedding — orthogonal and complementary: they reduce how much waste is generated; this change protects the completion of work already paid for.

### Anything else?

Salvage-value asymmetry: completing a requery costs a milliseconds-level PK fetch and rescues seconds of already-spent ANN compute; admitting one more new search instead buys none of that.

PR #50454, opened 2026-06-10
enhance: add querynode read backpressure

chyezh:
issue: #49707

- querynode scheduler: clean deadline-reached read tasks under queue pressure and split queue outcome metrics for scheduled, expired, deadline advance, and cleared paths
- proxy DQL scheduling: add adaptive slowdown/recover backpressure from downstream resource errors with dynamic concurrency controls, metrics, and transition logs
- proxy shard routing: stop retrying resource-insufficient DQL errors and cancel sibling shard requests on failure
- streaming delegator: cancel and wait for sibling hybrid search subrequests after subrequest failure

PR #41965, opened 2025-05-20
fix: cost metrics collection logic for replica selection

weiliu1031:
issue: #41621
- Deprecate EnableWorkerSQCostMetrics parameter
- Always collect cost metrics from all search and retrieve results
- Update code with comments explaining the changes rationale

PR #24413, opened 2023-05-25
add request cost metrics on qn

weiliu1031:
Signed-off-by: Wei Liu <wei.liu@zilliz.com>
issue: #24366

PR #52190, opened 2026-08-05
enhance: add dedicated priority lane for requery tasks in read scheduler (#52189)

MrPresent-Han:
issue: #52189

### What this PR does

When the QueryNode unsolved queue is full, rejecting the requery phase of
Search/HybridSearch discards ANN work that has already completed and causes a
client retry to repeat that work against an overloaded cluster.

This PR adds a dedicated bounded requery lane as a policy-layer decorator over
the FIFO read scheduler:

- **Identification**: uses the existing `QueryLabel="requery"` carried by
  `RetrieveRequest`; no proto or Proxy plumbing change is required.
- **Admission**: adds refreshable
  `queryNode.scheduler.requeryUnsolvedQueueSize` with an independent default
  of 1024. A positive integer sets the lane capacity, `<= 0` disables it, and
  an invalid value emits a rate-limited warning and falls back to 1024.
- **Full-lane handling**: admission performs one expired-task cleanup and one
  retry. If the retry is still rejected, only the current request fails and
  the current receive batch ends, giving the scheduler a chance to drain work;
  later requests remain pending for the next batch.
- **Dispatch**: replaces strict priority with request-weighted requery credit.
  Credit starts at 3 and each lane pop consumes one. Every queued task tracks
  how many pre-merge scheduler tasks it represents; credit is refreshed to
  `max(3, originalRequestCount)` only when a regular task is actually handed
  to execution, so a staged regular task later dropped by expiration or clear
  opens no lane window, and a task already canceled when popped never counts
  as regular service.
- **Accounting**: regular admission continues to use the regular-only counter,
  while public task/NQ totals include both regular and requery backlog for
  `CostAggregation`, load balancing, and aggregate ReadyLen metrics. The
  delegated query cost path now keeps all workers' costs, finishing the
  `EnableWorkerSQCostMetrics` deprecation declared in #41965 (the search path
  already does), and the merged cost's `totalNQ` takes the per-field maximum
  across workers so a backlogged worker's signal survives response-time
  snapshot selection; remote requery backlog therefore reaches the leader's
  `CostAggregation` and the look-aside balancer. The consumer-less
  `queryNode.enableWorkerSQCostMetrics` parameter is removed (it was never
  exported to milvus.yaml, and unknown keys in existing configs are ignored).
- **Observability**: adds
  `milvus_querynode_read_task_requery_ready_len` for the lane depth, including
  a task staged for execution handoff; adds
  `milvus_querynode_read_task_reject_total{lane=regular|requery}` counting
  admission rejections per class; and adds
  `milvus_querynode_read_task_requery_queue_duration` as the requery-only
  waiting-time subset per outcome, while the existing aggregate
  `read_task_queue_duration` histogram keeps its label schema untouched. `originalRequestCount` is internal
  scheduling metadata and is not exported as a Prometheus metric.
- **Lifecycle**: expiration cleanup covers both policy queues and the
  scheduler-local staged task before capacity enforcement. Admin clear,
  shutdown drain, counter settlement, and queue-duration metrics preserve the
  existing scheduler ownership semantics.
- **Policy compatibility**: the lane wraps FIFO only. It remains disabled under
  `user-task-polling`, preserving per-user round-robin scheduling and
  `maxPendingTaskPerUser`.

### Why weighted priority

Requery has high salvage value because it completes work whose ANN phase has
already run, but strict priority can starve regular reads. A fixed task ratio is
also insufficient because one regular SearchTask may represent several merged
requests while their requeries remain separate tasks. The finite credit window
preserves regular progress, and `originalRequestCount` compensates for the queue
unit mismatch introduced by merging.

### Compatibility

- Requests without the requery label follow the regular path unchanged.
- `user-task-polling` remains unchanged because it is not wrapped.
- Regular and requery capacities are independent. With both defaults, the
  intentional waiting bound is 1024 regular + 1024 requery tasks.
- The scheduler is shared by QueryNode workers and the streaming-node
  delegator, so both FIFO admission points benefit.
- CPU/GPU cross-pool head-of-line blocking needs pool-aware dispatch and is
  tracked separately in #52394.
- The change is complementary to #50454: that work reduces overload traffic;
  this change protects completion of work already performed.

### Tests

Focused policy, scheduler, metric, config, and paramtable tests cover:

- weighted credit and merged-task request counts;
- credit granted at execution handoff rather than Pop, and a staged regular
  task dropped before handoff granting no lane window;
- already-canceled regular tasks not replenishing credit;
- per-lane rejection counting and the requery-only queue-duration subset
  metric;
- merged request cost keeping the per-field maximum `totalNQ` across worker
  snapshots without mutating worker responses;
- lane-full cleanup/retry and receive-batch yielding;
- independent regular/requery admission budgets and aggregate load accounting;
- live staged tasks counting against capacity and expired staged tasks releasing
  capacity for both regular and requery admission;
- fixed default, explicit, disabled, invalid fallback, and independent hot
  update behavior;
- expiration, clear, shutdown, and dynamic disable;
- unchanged `user-task-polling` behavior.

No performance claim is made beyond the scheduling mechanism itself; overload
benchmarking remains follow-up validation.

### DIFF
--- a/internal/querynodev2/handlers.go
+++ b/internal/querynodev2/handlers.go
@@ -231,15 +231,10 @@ func (node *QueryNode) queryChannel(ctx context.Context, req *querypb.QueryReque
 	}
 	// aggregate cost
 	requestCosts := lo.FilterMap(results, func(result *internalpb.RetrieveResults, _ int) (*internalpb.CostAggregation, bool) {
-		if paramtable.Get().QueryNodeCfg.EnableWorkerSQCostMetrics.GetAsBool() {
-			return result.GetCostAggregation(), true
-		}
-
-		if result.GetBase().GetSourceID() == paramtable.GetNodeID() {
-			return result.GetCostAggregation(), true
-		}
-
-		return nil, false
+		// delegator node won't be used to load sealed segment if stream node is enabled
+		// and if growing segment doesn't exists, delegator won't produce any cost metrics
+		// so all workers' costs are collected (the enableWorkerSQCostMetrics toggle was removed)
+		return result.GetCostAggregation(), true
 	})
 	resp.CostAggregation = segmentutil.MergeRequestCost(requestCosts)
 	if resp.CostAggregation == nil {
--- a/internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
+++ b/internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
@@ -172,7 +172,8 @@ func (s *scheduler) schedule() {
 				// drain policy maintained task
 				for task.valid() {
 					execChan <- task.Task
-					s.updateWaitingTaskCounter(-1, -nq)
+					s.onTaskServed(task)
+					s.onTaskDequeued(task, task.NQ())
 					task = s.produceExecChan(now)
 				}
 				mlog.Info(context.TODO(), "all task put into exeChan, schedule worker exit")
@@ -181,7 +182,7 @@ func (s *scheduler) schedule() {
 			}
 			// Receive add operation request and return the process result.
 			// And consume recv chan as much as possible.
-			s.consumeRecvChan(req, maxReceiveChanBatchConsumeNum, now)
+			s.consumeRecvChan(req, maxReceiveChanBatchConsumeNum, now, &task)
 		case req := <-s.clearChan:
 			var result ClearResult
 			result, task = s.clearQueuedTasks(req.filter, req.reason, task, now)
@@ -189,7 +190,8 @@ func (s *scheduler) schedule() {
 		case execChan <- execTask:
 			// Task sent, drop the ownership of sent task.
 			// Update waiting task counter.
-			s.updateWaitingTaskCounter(-1, -nq)
+			s.onTaskServed(task)
+			s.onTaskDequeued(task, nq)
 			// And produce new task into execChan as much as possible.
 			task = s.produceExecChan(now)
 		}
@@ -197,10 +199,10 @@ func (s *scheduler) schedule() {
 }
 
 // consumeRecvChan consume the recv chan as much as possible.
-func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
+func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time, staged **queuedTask) {
 	// Check the dynamic wait task limit.
 	maxWaitTaskNum := paramtable.Get().QueryNodeCfg.MaxUnsolvedQueueSize.GetAsInt64()
-	if !s.handleAddTaskRequest(req, maxWaitTaskNum, now) {
+	if !s.handleAddTaskRequest(req, maxWaitTaskNum, now, staged) {
 		return
 	}
 
@@ -211,7 +213,7 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
 			if !ok {
 				return
 			}
-			if !s.handleAddTaskRequest(req, maxWaitTaskNum, now) {
+			if !s.handleAddTaskRequest(req, maxWaitTaskNum, now, staged) {
 				return
 			}
 		default:
@@ -222,33 +224,97 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
 
 // HandleAddTaskRequest handle a add task request.
 // Return true if the process can be continued.
-func (s *scheduler) handleAddTaskRequest(req addTaskReq, maxWaitTaskNum int64, now time.Time) bool {
-	if maxWaitTaskNum > 0 && s.GetWaitingTaskTotal() >= maxWaitTaskNum {
-		s.cleanupExpiredTasks(now)
+func (s *scheduler) handleAddTaskRequest(req addTaskReq, maxWaitTaskNum int64, now time.Time, staged **queuedTask) bool {
+	requery := false
+	if classifier, ok := s.policy.(laneClassifier); ok {
+		requery = classifier.UseLane(req.task)
 	}
 
 	if err := req.task.Context().Err(); err != nil {
 		mlog.Warn(context.TODO(), "task canceled before enqueue", mlog.Err(err))
 		req.err <- err
-	} else if maxWaitTaskNum > 0 && s.GetWaitingTaskTotal() >= maxWaitTaskNum {
-		err := merr.WrapErrTooManyRequests(
-			int32(maxWaitTaskNum),
-			fmt.Sprintf("limit by %s", paramtable.Get().QueryNodeCfg.MaxUnsolvedQueueSize.Key),
-		)
-		req.err <- err
-	} else {
-		// Push the task into the policy to schedule and update the counter of the ready queue.
-		queued := newQueuedTask(req.task, now)
-		nq := queued.NQ()
-		newTaskAdded, err := s.policy.Push(queued)
-		if err == nil {
-			s.updateWaitingTaskCounter(int64(newTaskAdded), nq)
-		}
-		req.err <- err
+		return maxWaitTaskNum <= 0 || s.getRegularWaitingTaskTotal() < maxWaitTaskNum
 	}
 
-	// Continue processing if the queue isn't reach the max limit.
-	return maxWaitTaskNum <= 0 || s.GetWaitingTaskTotal() < maxWaitTaskNum
+	capacity, available := s.waitingTaskCapacityAvailable(requery, maxWaitTaskNum)
+	if !available {
+		s.cleanupExpiredTasks(now, staged)
+		if err := req.task.Context().Err(); err != nil {
+			mlog.Warn(context.TODO(), "task canceled before enqueue", mlog.Err(err))
+			req.err <- err
+			return maxWaitTaskNum <= 0 || s.getRegularWaitingTaskTotal() < maxWaitTaskNum
+		}
+		capacity, available = s.waitingTaskCapacityAvailable(requery, maxWaitTaskNum)
+	}
+	if !available {
+		recordReadTaskReject(requery)
+		if requery {
+			req.err <- requeryLaneCapacityError(capacity)
+		} else {
+			req.err <- merr.WrapErrTooManyRequests(
+				int32(capacity),
+				fmt.Sprintf("limit by %s", paramtable.Get().QueryNodeCfg.MaxUnsolvedQueueSize.Key),
+			)
+		}
+		return false
+	}
+
+	// Allocate only after the common canceled/full rejection paths have passed.
+	queued := newQueuedTask(req.task, now)
+	queued.requery = requery
+	nq := queued.NQ()
+	newTaskAdded, err := s.pushTask(queued, now, staged)
+	if err == nil {
+		s.updateWaitingTaskCounter(queued, int64(newTaskAdded), nq)
+	}
+	req.err <- err
+	if errors.Is(err, merr.ErrServiceTooManyRequests) {
+		recordReadTaskReject(requery)
+		if requery {
+			// The lane remained full after cleanup and retry. Yield the scheduler
+			// loop so execChan handoff can drain work; unread requests remain pending.
+			return false
+		}
+	}
+	// Continue processing only while the regular queue still has room.
+	return maxWaitTaskNum <= 0 || s.getRegularWaitingTaskTotal() < maxWaitTaskNum
+}
+
+func (s *scheduler) waitingTaskCapacityAvailable(requery bool, maxRegular int64) (int64, bool) {
+	if requery {
+		capacity := requeryLaneCapacity()
+		return capacity, capacity > 0 && s.getRequeryWaitingTaskTotal() < capacity
+	}
+	return maxRegular, maxRegular <= 0 || s.getRegularWaitingTaskTotal() < maxRegular
+}
+
+// pushTask retries a requery-lane capacity rejection once after the scheduler
+// has completed expired-task lifecycle handling. pushTaskOnce performs
+// admission before mutating the policy.
+func (s *scheduler) pushTask(task *queuedTask, now time.Time, staged **queuedTask) (int, error) {
+	newTaskAdded, err := s.pushTaskOnce(task)
+	if !task.requery || !errors.Is(err, merr.ErrServiceTooManyRequests) {
+		return newTaskAdded, err
+	}
+
+	if ctxErr := task.Context().Err(); ctxErr != nil {
+		return 0, ctxErr
+	}
+	s.cleanupExpiredTasks(now, staged)
+	if ctxErr := task.Context().Err(); ctxErr != nil {
+		return 0, ctxErr
+	}
+	return s.pushTaskOnce(task)
+}
+
+func (s *scheduler) pushTaskOnce(task *queuedTask) (int, error) {
+	if task.requery {
+		capacity := requeryLaneCapacity()
+		if capacity <= 0 || s.getRequeryWaitingTaskTotal() >= capacity {
+			return 0, requeryLaneCapacityError(capacity)
+		}
+	}
+	return s.policy.Push(task)
 }
 
 // produceExecChan produce task from scheduler into exec chan as much as possible
@@ -266,7 +332,8 @@ func (s *scheduler) produceExecChan(now time.Time) *queuedTask {
 		select {
 		case execChan <- execTask:
 			// Update waiting task counter.
-			s.updateWaitingTaskCounter(-1, -nq)
+			s.onTaskServed(task)
+			s.onTaskDequeued(task, nq)
 			// Task sent, drop the ownership of sent task.
 			task = nil
 		default:
@@ -338,6 +405,26 @@ func readTaskExecuteOutcome(err error) string {
 	return metrics.FailLabel
 }
 
+// onTaskDequeued updates the matching queue-class counters when a task leaves
+// scheduler ownership at execChan handoff, expiration, clear, or shutdown.
+func (s *scheduler) onTaskDequeued(task *queuedTask, nq int64) {
+	if !task.valid() {
+		return
+	}
+	s.updateWaitingTaskCounter(task, -1, -nq)
+}
+
+// onTaskServed reports an execChan handoff to the policy. It runs on the
+// schedule goroutine only, so the policy needs no locking.
+func (s *scheduler) onTaskServed(task *queuedTask) {
+	if !task.valid() {
+		return
+	}
+	if observer, ok := s.policy.(taskServedObserver); ok {
+		observer.onTaskServed(task)
+	}
+}
+
 // setupExecListener setup the execChan and next task to run.
 func (s *scheduler) setupExecListener(lastWaitingTask *queuedTask, now time.Time) (*queuedTask, int64, chan Task) {
 	var execChan chan Task
@@ -350,7 +437,7 @@ func (s *scheduler) setupExecListener(lastWaitingTask *queuedTask, now time.Time
 				break
 			}
 			if err := lastWaitingTask.Context().Err(); err != nil {
-				s.updateWaitingTaskCounter(-1, -lastWaitingTask.NQ())
+				s.onTaskDequeued(lastWaitingTask, lastWaitingTask.NQ())
 				s.recordReadTaskQueueDuration(lastWaitingTask, now, readTaskQueueOutcomeExpired)
 				lastWaitingTask.Done(err)
 				lastWaitingTask = nil
@@ -369,12 +456,19 @@ func (s *scheduler) setupExecListener(lastWaitingTask *queuedTask, now time.Time
 	return lastWaitingTask, nq, execChan
 }
 
-func (s *scheduler) cleanupExpiredTasks(now time.Time) {
+func (s *scheduler) cleanupExpiredTasks(now time.Time, staged **queuedTask) {
 	deadlineAdvance := paramtable.Get().QueryNodeCfg.SchedulePolicyTaskDeadlineAdvance.GetAsDurationByParse()
 	cleanupTime := now.Add(deadlineAdvance)
+	if staged != nil && (*staged).cleanupReady(cleanupTime) {
+		task := *staged
+		*staged = nil
+		s.onTaskDequeued(task, task.NQ())
+		// Queue duration was already recorded when this task was selected by Pop.
+		task.Done(cleanupTaskError(task))
+	}
 	tasks := s.policy.Cleanup(cleanupTime)
 	for _, task := range tasks {
-		s.updateWaitingTaskCounter(-1, -task.NQ())
+		s.onTaskDequeued(task, task.NQ())
 		s.recordReadTaskQueueDuration(task, now, readTaskQueueOutcomeExpired)
 		task.Done(cleanupTaskError(task))
 	}
@@ -396,7 +490,7 @@ func (s *scheduler) clearQueuedTasks(filter TaskFilter, reason string, task *que
 		nq := removedTask.NQ()
 		result.QueuedCleared++
 		result.QueuedNQCleared += nq
-		s.updateWaitingTaskCounter(-1, -nq)
+		s.onTaskDequeued(removedTask, nq)
 		s.recordReadTaskQueueDuration(removedTask, now, readTaskQueueOutcomeCleared)
 		removedTask.Done(clearErr)
 	}
@@ -413,42 +507,86 @@ func clearTaskQueueError(reason string) error {
 // setupReadyLenMetric update the read task ready len metric.
 func (s *scheduler) setupReadyLenMetric() {
 	waitingTaskCount := s.GetWaitingTaskTotal()
+	nodeID := paramtable.GetStringNodeID()
 
 	// Update the ReadyQueue counter for quota.
 	collector.Counter.Set(metricsinfo.ReadyQueueType, waitingTaskCount)
-	// Record the waiting task length of policy as ready task metric.
-	metrics.QueryNodeReadTaskReadyLen.WithLabelValues(paramtable.GetStringNodeID()).Set(float64(waitingTaskCount))
-	metrics.QueryNodeReadTaskReadyNQ.WithLabelValues(paramtable.GetStringNodeID()).Set(float64(s.GetWaitingTaskTotalNQ()))
+	// Aggregate gauges include both regular and requery waiting tasks. The
+	// lane-specific gauge uses the same counter as admission, so a task staged
+	// for execChan handoff remains visible until ownership is transferred.
+	metrics.QueryNodeReadTaskReadyLen.WithLabelValues(nodeID).Set(float64(waitingTaskCount))
+	metrics.QueryNodeReadTaskReadyNQ.WithLabelValues(nodeID).Set(float64(s.GetWaitingTaskTotalNQ()))
+	metrics.QueryNodeReadTaskRequeryReadyLen.WithLabelValues(nodeID).Set(float64(s.getRequeryWaitingTaskTotal()))
 }
 
 func (s *scheduler) recordReadTaskQueueDuration(task *queuedTask, now time.Time, outcome string) {
 	if !task.valid() {
 		return
 	}
-	metrics.QueryNodeReadTaskQueueDuration.WithLabelValues(
-		paramtable.GetStringNodeID(),
-		outcome,
-	).Observe(float64(task.queueDuration(now).Microseconds()) / 1000.0)
+	nodeID := paramtable.GetStringNodeID()
+	durationMs := float64(task.queueDuration(now).Microseconds()) / 1000.0
+	// The aggregate histogram keeps its pre-existing label schema and covers
+	// both classes; the requery-only subset goes to a separate metric.
+	metrics.QueryNodeReadTaskQueueDuration.WithLabelValues(nodeID, outcome).Observe(durationMs)
+	if task.requery {
+		metrics.QueryNodeReadTaskRequeryQueueDuration.WithLabelValues(nodeID, outcome).Observe(durationMs)
+	}
+}
+
+func laneLabel(requery bool) string {
+	if requery {
+		return metrics.ReQueryLabel
+	}
+	return metrics.RegularLabel
+}
+
+func recordReadTaskReject(requery bool) {
+	metrics.QueryNodeReadTaskRejectCnt.WithLabelValues(paramtable.GetStringNodeID(), laneLabel(requery)).Inc()
 }
 
 // scheduler counter implement, concurrent safe.
 type schedulerCounter struct {
-	waitingTaskTotal   atomic.Int64
-	waitingTaskTotalNQ atomic.Int64
+	regularWaitingTaskTotal   atomic.Int64
+	regularWaitingTaskTotalNQ atomic.Int64
+	requeryWaitingTaskTotal   atomic.Int64
+	requeryWaitingTaskTotalNQ atomic.Int64
 }
 
 // GetWaitingTaskTotal get ready task counts.
 func (s *schedulerCounter) GetWaitingTaskTotal() int64 {
-	return s.waitingTaskTotal.Load()
+	return s.regularWaitingTaskTotal.Load() + s.requeryWaitingTaskTotal.Load()
 }
 
 // GetWaitingTaskTotalNQ get ready task NQ.
 func (s *schedulerCounter) GetWaitingTaskTotalNQ() int64 {
-	return s.waitingTaskTotalNQ.Load()
+	return s.regularWaitingTaskTotalNQ.Load() + s.requeryWaitingTaskTotalNQ.Load()
 }
 
-// updateWaitingTaskCounter update the waiting task counter for observing.
-func (s *schedulerCounter) updateWaitingTaskCounter(num int64, nq int64) {
-	s.waitingTaskTotal.Add(num)
-	s.waitingTaskTotalNQ.Add(nq)
+func (s *schedulerCounter) getRegularWaitingTaskTotal() int64 {
+	return s.regularWaitingTaskTotal.Load()
+}
+
+func (s *schedulerCounter) getRegularWaitingTaskTotalNQ() int64 {
+	return s.regularWaitingTaskTotalNQ.Load()
+}
+
+func (s *schedulerCounter) getRequeryWaitingTaskTotal() int64 {
+	return s.requeryWaitingTaskTotal.Load()
+}
+
+func (s *schedulerCounter) getRequeryWaitingTaskTotalNQ() int64 {
+	return s.requeryWaitingTaskTotalNQ.Load()
+}
+
+// updateWaitingTaskCounter dispatches updates by the immutable queue-class
+// stamp so regular admission and exported load accounting can use different
+// views without losing lifecycle symmetry.
+func (s *schedulerCounter) updateWaitingTaskCounter(task *queuedTask, num int64, nq int64) {
+	if task.requery {
+		s.requeryWaitingTaskTotal.Add(num)
+		s.requeryWaitingTaskTotalNQ.Add(nq)
+		return
+	}
+	s.regularWaitingTaskTotal.Add(num)
+	s.regularWaitingTaskTotalNQ.Add(nq)
 }
--- a/internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
+++ b/internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
@@ -13,6 +13,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/metrics"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/util/conc"
+	"github.com/milvus-io/milvus/pkg/v3/util/contextutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/lifetime"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/metricsinfo"
@@ -33,6 +34,7 @@ func newScheduler(policy schedulePolicy) Scheduler {
 	mlog.Info(context.TODO(), "query node use concurrent safe scheduler", mlog.Int("max_concurrency", maxReadConcurrency))
 	return &scheduler{
 		policy:           policy,
+		requeryQueue:     newMergeTaskQueue("requery"),
 		receiveChan:      make(chan addTaskReq),
 		clearChan:        make(chan clearQueuedReq),
 		execChan:         make(chan Task),
@@ -61,12 +63,16 @@ type clearQueuedResp struct {
 
 // scheduler is a general concurrent safe scheduler implementation by wrapping a schedule policy.
 type scheduler struct {
-	policy      schedulePolicy
-	receiveChan chan addTaskReq
-	clearChan   chan clearQueuedReq
-	execChan    chan Task
-	pool        *conc.Pool[any]
-	gpuPool     *conc.Pool[any]
+	policy schedulePolicy
+	// requeryQueue is a bounded lane dispatched ahead of the policy queue; it is
+	// owned by the schedule goroutine and its tasks stay out of the policy
+	// waiting counters. See tryAddRequeryTask for the rationale.
+	requeryQueue *mergeTaskQueue
+	receiveChan  chan addTaskReq
+	clearChan    chan clearQueuedReq
+	execChan     chan Task
+	pool         *conc.Pool[any]
+	gpuPool      *conc.Pool[any]
 
 	// wg is the waitgroup for internal worker goroutine
 	wg sync.WaitGroup
@@ -172,7 +178,7 @@ func (s *scheduler) schedule() {
 				// drain policy maintained task
 				for task.valid() {
 					execChan <- task.Task
-					s.updateWaitingTaskCounter(-1, -nq)
+					s.onTaskDequeued(task, nq)
 					task = s.produceExecChan(now)
 				}
 				mlog.Info(context.TODO(), "all task put into exeChan, schedule worker exit")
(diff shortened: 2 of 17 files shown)


internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -338,19 +380,37 @@ func readTaskExecuteOutcome(err error) string {
 	return metrics.FailLabel
 }
 
+// popTask pops the next task to run: the requery lane first (strict priority —
+// completing a requery salvages already-spent ANN computation), then the policy.
+func (s *scheduler) popTask(now time.Time) *queuedTask {
+	if task := s.requeryQueue.pop(); task.valid() {
```
buqian-zilliz: **[P1] Guarantee service for the regular policy queue**

`popTask` never reaches `policy.Pop` while the requery lane remains nonempty, so ordinary search and query tasks have no guaranteed service. Deriving requery traffic from admitted searches does not bound the work arriving at this scheduler: retrieval fans out across channels and workers, requests can be retried, and output-field fetch cost varies with result size. Under sustained overload, a replenished lane can therefore hold regular tasks until their contexts expire. Use bounded-burst or weighted priority, or reserve service for the policy queue.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -338,19 +380,37 @@ func readTaskExecuteOutcome(err error) string {
...
+		return task
+	}
+	return s.policy.Pop(now)
+}
+
+// onTaskDequeued updates the policy waiting counters when a task leaves the
+// scheduler; requery-lane tasks are tracked by the lane itself only.
+func (s *scheduler) onTaskDequeued(task *queuedTask, nq int64) {
+	if task.valid() && task.requery {
```
buqian-zilliz: **[P1] Include requery backlog in node load accounting**

The lane is never added to the waiting counters, and this branch explicitly keeps it out of their dequeue accounting. Those counters are not only used for regular-queue admission: QueryNode copies `GetWaitingTaskTotalNQ()` into `CostAggregation.TotalNQ`, which the proxy look-aside balancer cubes when scoring nodes. A node can therefore have up to `requeryUnsolvedQueueSize` priority tasks queued while advertising little or no backlog, attracting more reads and retries during overload. Keep a policy-only counter for `unsolvedQueueSize`, but maintain separate total or lane counters for exported load accounting.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -251,6 +261,38 @@ func (s *scheduler) handleAddTaskRequest(req addTaskReq, maxWaitTaskNum int64, n
...
+		req.err <- err
+	} else if int64(s.requeryQueue.len()) >= laneSize {
+		req.err <- merr.WrapErrTooManyRequests(
+			int32(laneSize),
+			fmt.Sprintf("limit by %s", paramtable.Get().QueryNodeCfg.RequeryUnsolvedQueueSize.Key),
+		)
+	} else {
+		queued := newQueuedTask(req.task, now)
+		queued.requery = true
+		s.requeryQueue.push(queued)
```
buqian-zilliz: **[P2] Preserve per-user isolation in the requery lane**

Pushing directly into `requeryQueue` bypasses `schedulePolicy.Push`. When `scheduleReadPolicy` is `user-task-polling`, requery tasks consequently skip both per-user round-robin scheduling and `maxPendingTaskPerUser`. One tenant can fill the global lane, cause other tenants' requeries to be rejected, and run its backlog ahead of every normal read despite the configured fair policy. Enforce equivalent per-user admission and scheduling inside the priority lane, or use a priority-aware user policy.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Resolved in 3c5b8e7748. Requery priority now lives in requeryPriorityPolicy and uses a K=2 bounded burst: while both queues have work, two lane tasks are followed by one regular-policy task. The inner-empty fallback is also counted as the first task of the next burst, covering a regular task that arrives while that task awaits handoff. Covered by TestRequeryPriorityPolicyBoundedBurst and TestRequeryPriorityPolicyCountsFallbackInNextBurst; the scheduler package passes with -count=3.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Resolved in 3c5b8e7748. The scheduler now maintains separate regular and requery count/NQ pairs. Public waiting getters return their sum for Ready metrics and CostAggregation.TotalNQ, while regular admission reads only the regular counter; both classes decrement at actual execChan handoff, cleanup, clear, and shutdown. Covered by the separate-admission/exported-counter, cleanup, clear, and shutdown tests; the scheduler package passes with -count=3.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Resolved in 3c5b8e7748. The priority decorator is wired only around FIFO. user-task-polling remains unwrapped, so requery tasks continue through its existing per-user round-robin and maxPendingTaskPerUser admission without a global lane. Covered by TestRequeryLaneWiringDisablesLaneForUserTaskPolling and TestUserTaskPollingQuotaRejectDoesNotTriggerGlobalCleanup; the scheduler package passes with -count=3.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,121 @@
...
+
+func newRequeryPriorityPolicy(inner schedulePolicy) *requeryPriorityPolicy {
+	return &requeryPriorityPolicy{
+		inner: inner,
+		lane:  newMergeTaskQueue("requery"),
+	}
+}
+
+func (p *requeryPriorityPolicy) UseLane(task Task) bool {
+	return task != nil && requeryLaneCapacity() > 0 &&
```
buqian-zilliz: **[P2] Avoid resolving lane capacity for every regular task**

`UseLane` resolves `requeryLaneCapacity()` before checking the context label, so every ordinary search or query admitted by this single scheduler goroutine now executes `ParamItem.GetValue()`. That path walks `Manager.GetConfig` through multiple concurrent maps and, for normal file/etcd-backed values, a source read lock; the default `auto` path also reads `MaxUnsolvedQueueSize`, and actual requeries repeat the capacity lookup in `Push`. This adds avoidable shared synchronization to the admission hot path. Check `ReQueryLabel` first and resolve the capacity only for actual requery tasks, preferably once per admission.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,121 @@
...
+func (p *requeryPriorityPolicy) Push(task *queuedTask) (int, error) {
+	if !task.requery {
+		return p.inner.Push(task)
+	}
+
+	// Classification freezes the queue class, not the refreshable capacity.
+	// If the lane is disabled after classification, reject rather than falling
+	// through to a regular queue whose admission check was intentionally skipped.
+	capacity := requeryLaneCapacity()
+	if capacity <= 0 || int64(p.lane.len()) >= capacity {
```
buqian-zilliz: **[P2] Count the staged requery against lane capacity**

`p.lane.len()` does not include the requery task that `setupExecListener` has already popped into the scheduler-local handoff slot. When the worker pool is saturated, `exec()` blocks in `Pool.Submit`, so that staged task can remain waiting while the scheduler accepts more requests. With a configured capacity of 1, one requery can therefore be staged and another stored in `lane`, leaving two waiting requeries despite the advertised maximum. Regular admission avoids this by counting its staged task until `execChan` handoff. Enforce the limit against the requery waiting count, or otherwise include the staged task.

MrPresent-Han: Fixed in 517028464e. Requery admission now enforces capacity against `getRequeryWaitingTaskTotal()`, so a popped/staged task remains counted until execChan handoff. Added a cap=1 regression test covering the staged slot; scheduler tests passed for 3 runs.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Fixed in 517028464e. `UseLane` now checks `ReQueryLabel` before reading capacity, and `requeryUnsolvedQueueSize` uses ParamItem fallback plus the typed `GetAsInt64` cache, removing the per-admission trim/equal/parse path. Verified with the scheduler package tests (3 runs), paramtable tests, and the config YAML test.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,115 @@
...
+func (p *requeryPriorityPolicy) Pop(now time.Time) *queuedTask {
+	if p.burst < requeryPriorityBurstLimit {
+		if task := p.lane.pop(); task.valid() {
+			p.burst++
+			return task
+		}
+	}
+
+	if task := p.inner.Pop(now); task.valid() {
+		p.burst = 0
```
buqian-zilliz: **[P2] Do not count canceled regular tasks as burst service**

`Pop` resets the burst as soon as `inner.Pop` returns, before the scheduler checks whether that task is already canceled. `setupExecListener` drops a canceled task and calls `Pop` again; because the burst is now zero, up to two live lane tasks are selected before the next FIFO entry is inspected. A backlog of N canceled regular tasks can therefore put 2N requeries ahead of a live regular task behind them, so the bounded-burst policy does not bound service for executable regular work and the live task can time out under overload. Skip canceled inner entries before resetting the burst, or reset the burst only after a regular task survives validation or is handed off.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -223,32 +223,65 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
...
+func (s *scheduler) pushTask(task *queuedTask, now time.Time) (int, error) {
+	newTaskAdded, err := s.pushTaskOnce(task)
+	if !task.requery || !errors.Is(err, merr.ErrServiceTooManyRequests) {
+		return newTaskAdded, err
+	}
+
+	if ctxErr := task.Context().Err(); ctxErr != nil {
+		return 0, ctxErr
+	}
+	s.cleanupExpiredTasks(now)
```
buqian-zilliz: **[P2] Avoid rescanning both queues for every full-lane rejection**

When the requery lane is full, every failed admission reaches this call. `requeryPriorityPolicy.Cleanup` scans both the lane and the entire inner queue, while `handleAddTaskRequest` returns `true` whenever only the regular queue has room; `consumeRecvChan` can therefore process up to 100 full-lane requests and perform 100 O(lane+regular) scans before the scheduler gets another chance to hand a task to `execChan`. All requests in that batch reuse the same `now`, so the repeated deadline scans are largely redundant. Under the overload this feature targets, rejection traffic can monopolize the sole scheduler goroutine and delay draining the priority lane. Stop batching after a lane-cap rejection, or clean only the relevant lane at most once per batch before retrying.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"time"
+
+	"github.com/milvus-io/milvus/pkg/v3/metrics"
+	"github.com/milvus-io/milvus/pkg/v3/util/contextutil"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
+	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
+)
+
+const requeryPriorityBurstLimit = 2
```
liliu-z: When both queues are backed up, Pop grants at most two lane dispatches per regular dispatch, but one regular dispatch can be a merged SearchTask covering many original requests, and each of those requests later issues its own requery that the lane cannot merge (SearchTask is the only MergeTask implementation, so QueryTask/QueryStreamTask always occupy one lane slot each). Lane arrivals are therefore denominated in parent requests while lane drain is denominated in tasks: once the average merge group exceeds 2, lane depth grows until it reaches requeryUnsolvedQueueSize and requeries are rejected again, which is the outcome the lane exists to prevent. Consider deriving the lane credit from the popped task's original-request count rather than a fixed 2.

pkg/util/paramtable/component_param.go
```diff
@@ -4723,6 +4724,19 @@ Max read concurrency must greater than or equal to 1, and less than or equal to
 	}
 	p.MaxUnsolvedQueueSize.Init(base.mgr)
 
+	p.RequeryUnsolvedQueueSize = ParamItem{
+		Key:          "queryNode.scheduler.requeryUnsolvedQueueSize",
+		Version:      "3.0.0",
+		DefaultValue: "auto",
```
liliu-z: With the default `auto` the lane inherits unsolvedQueueSize (1024) and enforces it against its own counter, while regular admission checks a separate counter, so the node's queued-task ceiling becomes 1024 regular + 1024 requery. The split is intentional, but neither the param Doc nor configs/milvus.yaml says the effective bound is the sum, so an operator who tuned unsolvedQueueSize as the queue-depth bound silently gets twice as much. Document the summed ceiling or carve the lane slots out of the existing budget.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,115 @@
...
+func requeryLaneCapacityError(capacity int64) error {
+	return merr.WrapErrTooManyRequests(
+		int32(capacity),
+		fmt.Sprintf("limit by %s", paramtable.Get().QueryNodeCfg.RequeryUnsolvedQueueSize.Key),
+	)
+}
+
+// requeryLaneCapacity uses ParamItem's typed cache. The "auto" default is
+// resolved through the regular queue capacity fallback configured on the item.
+func requeryLaneCapacity() int64 {
```
liliu-z, line 132: requeryLaneCapacity goes through ParamItem.GetAsInt64, whose converter returns 0 on a parse error without logging, so a typo in queryNode.scheduler.requeryUnsolvedQueueSize disables the lane and reverts the node to pre-PR behaviour with no signal — the PR's own test codifies "invalid" resolving to 0. The aggregate ReadyLen gauge already counted these tasks before and after, so nothing exposes lane depth or lane rejections separately either, leaving no way to correlate a requery latency regression with the misconfiguration. Validate the raw value and warn once when it fails to parse, and export lane depth.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,115 @@
...
+		return 0, requeryLaneCapacityError(capacity)
+	}
+
+	p.lane.push(task)
+	return 1, nil
+}
+
+func (p *requeryPriorityPolicy) Pop(now time.Time) *queuedTask {
+	if p.burst < requeryPriorityBurstLimit {
+		if task := p.lane.pop(); task.valid() {
```
buqian-zilliz, line 82: **[P2] Keep CPU requery priority from blocking idle GPU workers**

`p.lane.pop()` chooses a requery before the scheduler considers which execution pool can accept it. Requery `QueryTask`s always use the CPU pool, GPU-index searches use the independent `gpuPool`, and the single `exec()` receiver blocks in `Pool.Submit` when the selected pool is full. With CPU workers saturated and GPU workers idle, two queued requeries can therefore block submission of a ready GPU search at the FIFO head, wasting the separate GPU capacity. Make dispatch pool-aware or use independent handoff loops so CPU priority cannot head-of-line block GPU work.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -223,32 +223,65 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
...
+	if ctxErr := task.Context().Err(); ctxErr != nil {
+		return 0, ctxErr
+	}
+	return s.pushTaskOnce(task)
+}
+
+func (s *scheduler) pushTaskOnce(task *queuedTask) (int, error) {
+	if task.requery {
+		capacity := requeryLaneCapacity()
+		if capacity <= 0 || s.getRequeryWaitingTaskTotal() >= capacity {
```
buqian-zilliz, line 313: **[P2] Evict an expired staged requery before enforcing the lane cap**

The requery counter includes the task already popped from the lane and staged for `execChan`, but that task is no longer visible to `policy.Cleanup`. If `exec()` is blocked in `Pool.Submit` and the staged task expires, `setupExecListener` does not recheck it because `lastWaitingTask` is still valid. At a capacity of 1, every new requery is then rejected and the cleanup-and-retry path cannot reclaim the slot until a CPU worker becomes available just to hand off and drop the expired task. Revalidate and remove the staged task before applying the capacity check, or include it in cleanup.

pkg/util/paramtable/component_param.go
```diff
@@ -4723,6 +4724,19 @@ Max read concurrency must greater than or equal to 1, and less than or equal to
...
+		FallbackKeys: []string{p.MaxUnsolvedQueueSize.Key},
```
buqian-zilliz: **[P2] Keep auto capacity refreshes from caching a stale fallback**

`FallbackKeys` makes the effective value come from `unsolvedQueueSize`, but `GetAsInt64` caches it using the primary `requeryUnsolvedQueueSize` raw value. On a cache miss, an admission can read the old fallback, race with a fallback refresh that clears the cache, and then reinstall the old capacity because `CASCachedValue` validates only the unchanged primary value `auto`. Subsequent admissions keep using that stale limit indefinitely, so lowering or disabling the inherited lane capacity may silently fail. Include the effective fallback value or version in cache validation, or resolve this dependency without the primary-only typed cache.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -223,32 +223,65 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
 // HandleAddTaskRequest handle a add task request.
 // Return true if the process can be continued.
 func (s *scheduler) handleAddTaskRequest(req addTaskReq, maxWaitTaskNum int64, now time.Time) bool {
-	if maxWaitTaskNum > 0 && s.GetWaitingTaskTotal() >= maxWaitTaskNum {
+	queued := newQueuedTask(req.task, now)
```
buqian-zilliz: **[P3] Keep rejected admission paths allocation-free**

`newQueuedTask` now runs before cancellation and capacity checks. Since accepted instances can be retained by the policy queue, this allocation site escapes, so canceled requests and requests rejected by the regular or requery capacity checks allocate even though they are never enqueued. During queue-full storms, that adds avoidable allocation and GC pressure to the single admission goroutine. Compute the lane classification first and allocate and stamp `queuedTask` only on the push path.

internal/util/searchutil/scheduler/requery_policy.go
buqian-zilliz: This issue still exists: `Pop` resets the burst before the scheduler validates the regular task's context, so canceled regular entries still count as burst service and can repeatedly delay the next live regular task.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
buqian-zilliz: This issue still exists: each full-lane rejection still scans both queues, and the batch continues whenever the regular queue has room, so one receive batch can repeat the same O(lane+regular) cleanup up to 100 times before scheduling resumes.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Fixed in b42efd8d12. Requery credit is now refreshed only when the popped regular task is still live; a task already canceled at Pop leaves the current credit unchanged, so scheduler cleanup cannot open another lane window. The base credit is now 3. Covered by TestRequeryPriorityPolicyCanceledRegularDoesNotRefreshCredit; the scheduler package passed 3 consecutive runs.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Fixed in b42efd8d12. Admission still performs one cleanup and one retry for the current requery, but a final lane-cap rejection now ends only the current receive batch; unread requests remain on receiveChan for the next scheduler iteration. Covered by TestRequeryLaneFullStopsReceiveBatch; the scheduler package passed 3 consecutive runs.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Fixed in b42efd8d12. Each queuedTask now carries an internal originalRequestCount initialized to 1; successful merges accumulate the incoming count, and a live regular Pop refreshes lane credit to max(3, originalRequestCount). This metadata is scheduler-only and is not exported as a Prometheus metric. Covered by queue and weighted-credit policy tests; the scheduler package passed 3 consecutive runs.

pkg/util/paramtable/component_param.go
MrPresent-Han: Addressed in b42efd8d12. The independent budgets are intentional and are now explicit in the regular/requery Param docs, generated configs/milvus.yaml, code comments, and PR body: with auto=N, the effective waiting bound is N regular + N requery = 2N. The YAML consistency test passes.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Fixed in b42efd8d12. The ParamItem formatter now strictly validates integer values, emits a rate-limited warning, and falls back to the current regular queue capacity; steady-state admission still uses the typed int64 cache. Added milvus_querynode_read_task_requery_ready_len using the same requery waiting counter as admission, including a task staged for execChan handoff. Paramtable, metrics, scheduler, and YAML consistency tests pass.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Agreed. The single blocking dispatcher predates this PR, but the requery lane widens the trigger by allowing a later CPU requery to bypass an earlier GPU regular task. A correct fix needs pool-aware or independent CPU/GPU handoff, including separate staged ownership and lifecycle/accounting handling; partial `Pool.Free()` or asynchronous `Submit` workarounds would be racy or hide backlog. Tracking the complete scheduler change in #52394 and deferring it from this bounded requery-lane PR.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Fixed in b3d08134ca. Capacity cleanup now receives the scheduler-local staged slot and uses the same cleanup cutoff as the policy queues. An expired staged regular or requery task is detached first, then its class counter/NQ is settled and Done is called; queue-duration remains the existing Pop-selection metric. Added regular and requery cap=1 regressions plus a live-staged guard. Scheduler package passed 3 full runs and the key cases passed 20 runs.

pkg/util/paramtable/component_param.go
MrPresent-Han: Fixed in b3d08134ca by removing the cross-param dependency: requeryUnsolvedQueueSize no longer supports auto or FallbackKeys. Its independent default is 1024; invalid values emit a rate-limited warning and fall back to 1024, while hot updates depend only on this key. Param docs, generated YAML text, tests, and the PR body were updated. Paramtable and YAML consistency tests pass.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Fixed in b3d08134ca. Admission now classifies the raw Task first, checks cancellation and regular/requery capacity (including cleanup/recheck), and only then allocates and stamps queuedTask on the Push path. The policy keeps its final capacity validation for dynamic-update races. Scheduler tests passed 3 full runs; the rejection/staged cases passed 20 runs.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -222,33 +222,88 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
...
-		nq := queued.NQ()
-		newTaskAdded, err := s.policy.Push(queued)
-		if err == nil {
-			s.updateWaitingTaskCounter(int64(newTaskAdded), nq)
+		return maxWaitTaskNum <= 0 || s.getRegularWaitingTaskTotal() < maxWaitTaskNum
+	}
+
+	capacity, available := s.waitingTaskCapacityAvailable(requery, maxWaitTaskNum)
+	if !available {
+		s.cleanupExpiredTasks(now, staged)
```
buqian-zilliz, line 241: **[P2] Recheck cancellation after full-queue cleanup**

The request context is checked only before this cleanup. If its deadline fires while the O(queue) scan runs, the following capacity recheck can either admit the now-canceled task when a slot is reclaimed or return `TooManyRequests` instead of `ctx.Err()` when the queue remains full. In FIFO, an admitted canceled `SearchTask` can also merge into a live entry, so its NQ is executed because later cancellation checks inspect only the primary queue entry. Recheck `req.task.Context().Err()` immediately after cleanup before either rejecting or pushing.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,125 @@
...
+func (p *requeryPriorityPolicy) Push(task *queuedTask) (int, error) {
+	if !task.requery {
+		return p.inner.Push(task)
+	}
+
+	// Classification freezes the queue class, not the refreshable capacity.
+	// If the lane is disabled after classification, reject rather than falling
+	// through to a regular queue whose admission check was intentionally skipped.
+	capacity := requeryLaneCapacity()
+	if capacity <= 0 || int64(p.lane.len()) >= capacity {
```
buqian-zilliz: **[P3] Include staged work in the refreshed capacity recheck**

`pushTaskOnce` first validates the scheduler-owned requery count, but this final check re-reads the refreshable limit and compares it only with `p.lane.len()`. If the limit drops from 2 to 1 between those reads while one requery is staged and the physical lane is empty, both checks pass and this push leaves two scheduler-owned requeries under a limit of one; the excess can persist while `execChan` handoff is blocked. Validate the refreshed limit against the full requery waiting count, or make the refresh/check/enqueue sequence use one authoritative capacity snapshot.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Fixed in 4f4b812740. After full-queue cleanup, admission now rechecks req.task.Context().Err() before capacity reevaluation, rejection, allocation, or merge. Added regressions proving that cancellation during cleanup cannot merge extra NQ/originalRequestCount into a live SearchTask, and that ctx.Err() takes precedence over TooManyRequests when the queue remains full. The focused cases passed 20 runs and the scheduler package passed 3 full runs.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Fixed in 4f4b812740. scheduler.pushTaskOnce is now the sole requery capacity authority: it takes one capacity snapshot and compares it with getRequeryWaitingTaskTotal(), which includes both the physical lane and a staged requery, before policy mutation. Removed the policy Push second capacity read and lane.len-only check. Added the exact cap 2 to 1 hot-shrink regression with one staged task; the focused cases passed 20 runs and the scheduler package passed 3 full runs.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -222,33 +222,93 @@ func (s *scheduler) consumeRecvChan(req addTaskReq, limit int, now time.Time) {
 
 // HandleAddTaskRequest handle a add task request.
 // Return true if the process can be continued.
-func (s *scheduler) handleAddTaskRequest(req addTaskReq, maxWaitTaskNum int64, now time.Time) bool {
-	if maxWaitTaskNum > 0 && s.GetWaitingTaskTotal() >= maxWaitTaskNum {
-		s.cleanupExpiredTasks(now)
+func (s *scheduler) handleAddTaskRequest(req addTaskReq, maxWaitTaskNum int64, now time.Time, staged **queuedTask) bool {
+	requery := false
```
buqian-zilliz, line 228: **[P2] Prioritize requeries before the shared admission channel**

Requery classification happens only after `Add` has won the single unbuffered `receiveChan`. Under regular-queue saturation, a requery competes with an arbitrary backlog of regular senders; each regular request can be received, scan and clean the full queue, and be rejected before the requery reaches this code. The requery can therefore hit its deadline despite an empty dedicated lane, discarding the completed ANN work this lane is meant to salvage. Classify before the shared wait and use a separate or priority-aware admission path so pending regular admissions cannot head-of-line block requeries.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
```diff
@@ -433,22 +513,47 @@ func (s *scheduler) recordReadTaskQueueDuration(task *queuedTask, now time.Time,
...
 // GetWaitingTaskTotal get ready task counts.
 func (s *schedulerCounter) GetWaitingTaskTotal() int64 {
-	return s.waitingTaskTotal.Load()
+	return s.regularWaitingTaskTotal.Load() + s.requeryWaitingTaskTotal.Load()
 }
 
 // GetWaitingTaskTotalNQ get ready task NQ.
 func (s *schedulerCounter) GetWaitingTaskTotalNQ() int64 {
-	return s.waitingTaskTotalNQ.Load()
+	return s.regularWaitingTaskTotalNQ.Load() + s.requeryWaitingTaskTotalNQ.Load()
```
buqian-zilliz, line 562: **[P2] Preserve remote requery backlog in delegated-query cost**

Although this getter now aggregates the requery lane, the default delegated Query/requery path discards that value. `QuerySegments` writes each worker's total, but `queryChannel` retains remote-worker costs only when `queryNode.enableWorkerSQCostMetrics` is true, whose default is false. If the delegator has no local subtask, such as with sealed-only data, every cost is filtered out and the leader returns an empty `CostAggregation`; `queryTask` then overwrites the balancer entry with `ResponseTime == 0`, so `calculateScore` ignores `TotalNQ`. Include remote-worker costs for delegated Query as Search already does, or aggregate their backlog independently, otherwise a full requery lane remains invisible to load balancing.

internal/util/searchutil/scheduler/requery_policy.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	if task := p.inner.Pop(now); task.valid() {
+		// Credit is a backlog-aware weighted burst, not a reservation for a
+		// specific parent request. A live regular task refreshes the current
+		// window, while a task already canceled at Pop does not count as regular
+		// service.
+		if task.Context().Err() == nil {
+			p.requeryCredit = max(requeryPriorityBaseCredit, task.originalRequestCount)
```
buqian-zilliz: **[P2] Do not grant credit to a staged regular task that is dropped**

`Pop` replenishes credit while the selected regular task is live but before it leaves scheduler ownership. While `exec()` is blocked, that task can remain staged until its deadline enters `taskDeadlineAdvance`; full-queue cleanup or `ClearQueued` then removes it without notifying this policy. The credit survives, so up to `max(3, originalRequestCount)` requeries are selected before the next regular task even though the credited task was never handed off. A prefix of staged expirations compounds these bursts and can make live regular work time out. Replenish credit on successful `execChan` handoff, or undo it when a staged regular task is removed.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Agreed. The dedicated lane currently provides independent capacity and priority only after a task has crossed the shared unbuffered admission channel; it cannot reorder callers already blocked in Add. A complete fix needs a priority-aware or separate admission path and must preserve regular progress, dynamic disable, user-task-polling, cancellation, and stop/drain lifecycle behavior. Tracking that scheduler-level change in #52447 and keeping it out of this bounded policy-lane PR.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Fixed in 44234fc6be. The delegated query cost path now keeps all workers' costs unconditionally, matching the search reduce path and finishing the EnableWorkerSQCostMetrics deprecation declared in #41965 (the param keeps its declaration but has no remaining consumer). A remote worker's requery backlog reported through GetWaitingTaskTotalNQ now survives queryChannel aggregation and reaches the look-aside balancer even when the delegator has no local subtask.

internal/util/searchutil/scheduler/requery_policy.go
MrPresent-Han: Fixed in 44234fc6be. Credit is now granted only at execChan handoff: the scheduler reports served tasks to the policy through a taskServedObserver probe at its handoff sites, and Pop no longer refreshes credit. A staged regular task dropped by expiration cleanup or clear is never reported as served, so it cannot open a lane window; this also subsumes the earlier canceled-at-Pop rule. Existing policy sequence tests keep their expected orders under a served-pop helper, proving the normal path is unchanged. Covered by TestRequeryPriorityPolicyCreditGrantedAtServeNotPop and TestRequeryLaneStagedRegularDroppedGrantsNoCredit; the scheduler package passed 3 full runs.

pkg/metrics/querynode_metrics.go
```diff
@@ -343,6 +364,7 @@ var (
 		}, []string{
 			nodeIDLabelName,
 			outcomeLabelName,
+			laneLabelName,
```
buqian-zilliz: **[P2] Preserve the existing queue-duration metric schema**

Adding `lane` to this existing histogram changes the label-set contract of every `milvus_querynode_read_task_queue_duration` series. During a rolling upgrade, old pods emit one `(node_id, outcome)` series while new pods emit separate `(node_id, outcome, lane)` series; rules or joins that expect one series per node/outcome can split, duplicate, or fail. Keep the original aggregate histogram and add a separate lane-specific metric, as done for ready length, or otherwise preserve an aggregate projection under the existing name.

internal/querynodev2/handlers.go
```diff
@@ -231,15 +231,10 @@ func (node *QueryNode) queryChannel(ctx context.Context, req *querypb.QueryReque
...
-
-		if result.GetBase().GetSourceID() == paramtable.GetNodeID() {
-			return result.GetCostAggregation(), true
-		}
-
-		return nil, false
+		// delegator node won't be used to load sealed segment if stream node is enabled
+		// and if growing segment doesn't exists, delegator won't produce any cost metrics
+		// so we deprecate the EnableWorkerSQCostMetrics param
+		return result.GetCostAggregation(), true
```
buqian-zilliz, line 237: **[P3] Retire the cost-metrics toggle at the configuration boundary**

This unconditional path leaves `queryNode.enableWorkerSQCostMetrics` with no consumers, but the refreshable parameter still defaults to `false` and is documented as controlling whether worker cost is used. Operators can continue setting and observing a live-looking value that no longer changes behavior. Mark the parameter as deprecated/no-op, or remove it through the intended configuration migration, and update its documentation in the same change.

buqian-zilliz, line 237: **[P3] Fold cost aggregation into a single pass**

Now that this predicate is unconditional, `lo.FilterMap` allocates an `O(len(results))` slice only for `MergeRequestCost` to scan it, and the following `lo.SumBy` traverses the same worker results again. This runs for every delegated Query/requery and grows with worker fan-out. Select the response-time cost and accumulate the other required fields in one pass over `results` to avoid the temporary allocation and redundant traversal.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
buqian-zilliz: This issue still exists: the configuration gate is gone, but `segmentutil.MergeRequestCost` returns only the `CostAggregation` with the largest `ResponseTime`. When a different worker carries the high `TotalNQ`, that backlog is still discarded before `LookAsideBalancer.calculateScore`. Aggregate the backlog signal independently of response-time selection while avoiding duplicate worker snapshots.

internal/querynodev2/handlers.go
MrPresent-Han: Keeping this as-is. The temporary slice and extra traversals are bounded by the per-channel worker fan-out (single digits), which is negligible next to the RPCs this path just completed. Folding everything into one pass would inline the cost-merge semantics into this handler and orphan segmentutil.MergeRequestCost, which we are deliberately keeping as the single authority for merge semantics (including the totalNQ backlog aggregation added for the thread above). The separate TotalRelatedDataSize accumulation over results is the pre-existing pattern and is unchanged.

internal/util/searchutil/scheduler/concurrent_safe_scheduler.go
MrPresent-Han: Fixed in e999d75917. segmentutil.MergeRequestCost (whose only production caller is this query path) now keeps latency fields from the highest-response-time snapshot while taking the per-field maximum of totalNQ across all worker costs, on a copy rather than mutating a worker's response. A per-field maximum keeps the backlogged worker visible to LookAsideBalancer.calculateScore and stays safe against duplicate worker snapshots because nothing is accumulated. Covered by the new TestMergeRequestCost. The search path's private mergeRequestCost keeps the pre-existing snapshot-only projection (dating to #24413) and is left for a follow-up to keep this PR's blast radius on the query path.

pkg/metrics/querynode_metrics.go
MrPresent-Han: Fixed in e999d75917. The lane label is reverted: read_task_queue_duration keeps its original (node_id, outcome) schema and continues to cover both classes, and the requery-only subset is exported as the new milvus_querynode_read_task_requery_queue_duration with the same outcome label — the same separate-metric pattern this PR uses for ready length. Regular-only waiting time is derivable as the bucket-wise difference. Covered by the updated scheduler metric tests.

internal/querynodev2/handlers.go
MrPresent-Han: Fixed in e999d75917 by removing the parameter entirely: the ParamItem definition, struct field, and test assertion are gone, and the three code comments that referenced it are reworded. The key was never exported to milvus.yaml, and paramtable ignores unknown keys, so existing configs that still set it keep starting up unchanged; 2.x branches keep their own definition.

liliu-z:
**internal/querynodev2/handlers_test.go:0** component_param.go drops the EnableWorkerSQCostMetrics field from queryNodeConfig, but handlers_test.go:192 still reads paramtable.Get().QueryNodeCfg.EnableWorkerSQCostMetrics.Key, and that file is not part of this diff. No other definition of the symbol exists in the tree, so go test -tags dynamic,test ./internal/querynodev2/... fails to compile, and scripts/run_go_codecov.sh runs internal/... with -failfast, so the whole internal test run aborts. The worker_cost and remote_cost subtest cases at handlers_test.go:184 also no longer test anything now that queryChannel collects every worker's cost unconditionally.

internal/util/searchutil/scheduler/requery_lane_test.go
```diff
@@ -0,0 +1,508 @@
...
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for blocker to start")
+	}
+	require.NoError(t, s.Add(lane1))
+	require.NoError(t, s.Add(lane2))
+	require.NoError(t, s.Add(lane3))
+	require.NoError(t, s.Add(regular))
+	require.Eventually(t, func() bool {
+		policy := s.policy.(*requeryPriorityPolicy)
+		return policy.lane.len() > 0 && s.getRequeryWaitingTaskTotal() > 0
```
liliu-z, line 486: requery_lane_test.go:486 reads policy.lane.len(), i.e. the plain int q.count at queues.go:25, from the require.Eventually goroutine while the running scheduler goroutine mutates the same field in markRemoved (queues.go:112) via requeryPriorityPolicy.Pop and produceExecChan. requery_policy.go:35 documents that the lane is owned by the scheduler goroutine and needs no locks, and a returning Add() only orders that task's push, not the later pops. go test -race -count=1000 on this test reports "WARNING: DATA RACE" at queues.go:25 vs queues.go:112 and fails with "race detected during execution of test"; scripts/run_go_codecov.sh runs internal/... with -race -failfast, so this intermittently aborts the whole run.

liliu-z:
**internal/querynodev2/segments/utils.go:0** segmentutil.MergeRequestCost now takes the per-field maximum of totalNQ, but Search and HybridSearch merge through the duplicate mergeRequestCost at internal/querynodev2/segments/utils.go:155, which still returns the whole snapshot of the highest-ResponseTime worker. Both paths write the same per-node entry in the look-aside balancer (task_search.go:1578 and task_query.go:1135 both reach look_aside_balancer.go:187 metrics.cost.Store(cost)), so with worker snapshots (ResponseTime=800, TotalNQ=10) and (ResponseTime=200, TotalNQ=2000) a query response reports 2000 and the next search response overwrites it with 10. Every Search with output fields issues a requery Query, so both writers are always live and the requery backlog this PR surfaces is erased from the node score on each search response.

internal/util/searchutil/scheduler/requery_lane_test.go
MrPresent-Han: Fixed in e926b64be7. TestRequeryLaneShutdownDrainsCounters no longer reads the scheduler-goroutine-owned policy.lane.len() from the require.Eventually goroutine; it now waits for the atomic requery waiting count to reach 2, which still proves that one task may be staged while another remains in the physical lane. Verified the focused test with -race -count=1000 and the full scheduler package with -race.

PR #52194, opened 2026-08-05
fix:[3.0] upgrade aws-sdk-cpp to 1.11.842 for checksum retries

jiaqizho:
pr: #52193
issue: #50915
aws-sdk-cpp 1.11.692 can reuse checksum state from a partially consumed aws-chunked stream when a PutObject request is retried after a transport or low-speed timeout. The retry sends the complete body with a stale checksum trailer, causing MinIO to reject the write with HTTP 400 XAmzContentChecksumMismatch.

Upgrade the direct Conan dependency to aws-sdk-cpp 1.11.842 and its corresponding recipe revision. This version includes the upstream partially consumed stream fix(aws-sdk-cpp#3706), which only preserves a calculated checksum after the stream reaches EOF and correctly handles a rewound request body during retries.


liliu-z (commented):
/lgtm
/approve

Issue #52195, opened 2026-08-05
Track vector index size in segment statistics


yuruiz:
## Summary

Persist the serialized size of completed vector indexes in each segment’s `Statistics`, including indexes built for compaction output segments.

## Scope

- Add `VectorIndexSize` and reserved `ScalarIndexSize` protobuf fields.
- Populate only `VectorIndexSize` after a vector index build completes.
- Aggregate completed vector indexes for the segment and preserve existing statistics.

`ScalarIndexSize` is intentionally not populated in this change.

PR #52196, opened 2026-08-05
feat: add vector index size statistics

yuruiz:
## Summary

- add vector and reserved scalar index-size fields to segment statistics
- persist the aggregate serialized size of finished vector indexes
- cover compaction-output segments through the async index completion path

issue: #52195

## Verification

```text
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord -run Test_indexTaskSuite
cd pkg && go test -count=1 ./proto/datapb
git diff --check
```


internal/datacoord/task_index.go
```diff
@@ -130,13 +130,44 @@ func (it *indexBuildTask) UpdateTaskVersion(nodeID int64) error {
...
+		// Hold this from index-meta completion through segment-stat persistence.
+		// That makes simultaneous vector field completions observe a stable sum.
+		it.meta.vectorIndexSizeMu.Lock()
+		defer it.meta.vectorIndexSizeMu.Unlock()
+	}
+
 	if err := it.meta.indexMeta.FinishTask(result); err != nil {
 		return err
 	}
+	if isFinishedVectorIndex {
```
liliu-z: FinishTask persists the index as Finished via catalog.AlterSegmentIndexes and updateVectorIndexSize persists the segment's VectorIndexSize via catalog.AlterSegments as two separate, non-atomic etcd writes. A DataCoord crash between them leaves the index Finished but VectorIndexSize at 0, and reloadFromMeta only re-enqueues Unissued/Retry/InProgress indexes (index_inspector.go:271-274), so a Finished index is never re-run to backfill it. The same gap means every vector index that finished before this feature/upgrade stays at 0, since it never entered the completion hook and nothing reconciles finished indexes at startup.

internal/datacoord/meta.go
```diff
@@ -1491,6 +1493,25 @@ func UpdateSegmentStats(segmentID int64, requestStats *datapb.Statistics) Update
...
+	return func(modPack *updateSegmentPack) bool {
+		segment := modPack.Get(segmentID)
+		if segment == nil {
+			mlog.Warn(context.TODO(), "meta update: update vector index size failed - segment not found",
+				mlog.Int64("segmentID", segmentID))
+			return false
+		}
+
+		stats := segment.EnsureStats()
+		stats.VectorIndexSize = vectorIndexSize
```
liliu-z, line 1528: VectorIndexSize is a DataCoord-owned Stats field written only here, but the in-place schema-bump materialization path adopts the compactor's Stats wholesale — `if s := resultSegment.GetStats(); s != nil { cloned.Stats = s }` at meta.go:3522 — and the compactor-computed Stats carries no VectorIndexSize, so it is zeroed on a segment whose vector index is still Finished. Because that path keeps the segment ID, createIndexesForSegment skips the already-present SegmentIndex (index_inspector.go:159-160), so updateVectorIndexSize never re-runs and the size stays 0 permanently. Preserve DataCoord-owned Stats fields when adopting compactor Stats there, mirroring the schema-bump-only sub-case that ships Stats=nil.

internal/datacoord/task_index.go
```diff
@@ -130,13 +130,44 @@ func (it *indexBuildTask) UpdateTaskVersion(nodeID int64) error {
...
 
+// updateVectorIndexSize persists the serialized footprint of every finished
+// vector index on this segment. Compaction produces a new segment before its
+// indexes are built, so index completion is the first point where the actual
+// vector-index file size is available. Its caller serializes vector build
+// completion so every update observes the latest aggregate.
+func (it *indexBuildTask) updateVectorIndexSize() error {
+	var vectorIndexSize int64
+	for _, segmentIndex := range it.meta.indexMeta.GetSegmentIndexes(it.CollectionID, it.SegmentID) {
+		if vecindexmgr.GetVecIndexMgrInstance().IsVecIndex(segmentIndex.IndexType) &&
```
liliu-z: updateVectorIndexSize sums only indexes where IsVecIndex(segmentIndex.IndexType) is true, but the copy-segment path inserts Finished vector SegmentIndex records with IndexSerializedSize populated and IndexType left empty (copy_segment_task.go:908-925). IsVecIndex("") is false, so those indexes are excluded from the aggregate; and because they bypass setJobInfo entirely, their segment's VectorIndexSize is never computed at all. Records whose persisted IndexType is empty (legacy indexes finished before the index_type proto field, number 21, was populated) hit the same gap after upgrade. Populate/fallback IndexType on the copy path and reconcile already-Finished records.

internal/datacoord/task_index.go
```diff
@@ -130,13 +130,44 @@ func (it *indexBuildTask) UpdateTaskVersion(nodeID int64) error {
...
 	it.SetState(indexpb.JobState(result.GetState()), result.GetFailReason())
 	return nil
 }
 
+// updateVectorIndexSize persists the serialized footprint of every finished
+// vector index on this segment. Compaction produces a new segment before its
+// indexes are built, so index completion is the first point where the actual
+// vector-index file size is available. Its caller serializes vector build
+// completion so every update observes the latest aggregate.
+func (it *indexBuildTask) updateVectorIndexSize() error {
```
liliu-z: updateVectorIndexSize is the only place the aggregate VectorIndexSize is maintained, and it runs solely on vector-index build completion. DropIndex followed by index-file GC never triggers it, so the segment keeps counting the dropped index's bytes and over-reports until some later vector build on that same segment happens to recompute. getSegmentIndexes already excludes IsDeleted field indexes, so a recompute would be correct — it simply is never triggered on the drop/GC path. Trigger a recompute (or decrement) when an index is dropped.

internal/datacoord/task_index.go
```diff
@@ -130,13 +130,44 @@ func (it *indexBuildTask) UpdateTaskVersion(nodeID int64) error {
 }
 
 func (it *indexBuildTask) setJobInfo(result *workerpb.IndexTaskInfo) error {
+	isFinishedVectorIndex := result.GetState() == commonpb.IndexState_Finished &&
+		vecindexmgr.GetVecIndexMgrInstance().IsVecIndex(it.IndexType)
+	if isFinishedVectorIndex {
+		// Hold this from index-meta completion through segment-stat persistence.
+		// That makes simultaneous vector field completions observe a stable sum.
+		it.meta.vectorIndexSizeMu.Lock()
```
liliu-z: The serialization is a dedicated process-wide meta.vectorIndexSizeMu (meta.go:107), not the segment meta lock as claimed, and it is held (via defer) across two sequential etcd writes — FinishTask's AlterSegmentIndexes and updateVectorIndexSize's AlterSegments. A single mutex serializes every vector-index completion across all segments and collections, so mass index builds (e.g. bulk load) queue their completions behind it, and a persistently failing stat write on one segment can stall completions for all others. The stable-sum guarantee is per-segment, so a per-segment or per-collection lock would remove the cross-segment serialization.

internal/datacoord/task_index.go
```diff
@@ -130,13 +130,44 @@ func (it *indexBuildTask) UpdateTaskVersion(nodeID int64) error {
...
+		defer it.meta.vectorIndexSizeMu.Unlock()
+	}
+
 	if err := it.meta.indexMeta.FinishTask(result); err != nil {
 		return err
 	}
+	if isFinishedVectorIndex {
+		if err := it.updateVectorIndexSize(); err != nil {
+			return err
```
liliu-z: setJobInfo returns on an updateVectorIndexSize error before calling SetState, after FinishTask has already persisted the index as Finished. The sole caller ignores the returned error (task_index.go:665), so the task is left with in-memory state != Finished while meta says Finished; the scheduler keeps it in runningTasks and re-issues QueryTaskOnWorker, re-running FinishTask under vectorIndexSizeMu on every poll until the stat write succeeds. If the worker has meanwhile lost the completed result (e.g. IndexNode restart during the stuck window), QueryTaskOnWorker resets the index to Init (task_index.go:676), which the scheduler treats as a failure and re-queues for a full rebuild. Mark the task Finished regardless of the stat write — move SetState before updateVectorIndexSize, or make the stat update best-effort/logged — since the index build genuinely succeeded.

internal/datacoord/meta.go
yuruiz: Fixed. The in-place schema-bump compaction path now preserves DataCoord-owned vector/scalar index-size fields when it adopts compactor Statistics. Data-log Statistics remain sourced from the compactor.

internal/datacoord/task_index.go
yuruiz: Fixed. The copy path now persists IndexType from the target index definition and reconciles the copied segment afterward. Reconciliation also falls back to the index definition for legacy SegmentIndex records with an empty IndexType.

internal/datacoord/task_index.go
yuruiz: Fixed. Replaced the process-wide mutex with a KeyLock keyed by segment ID. Completion and aggregation remain serialized per segment while independent segments can finish concurrently.

internal/datacoord/task_index.go
yuruiz: Fixed. setJobInfo now marks the in-memory task Finished immediately after FinishTask succeeds. The independent statistics update is best-effort and logged on failure; startup reconciliation repairs it.

internal/datacoord/task_index.go
yuruiz: Fixed. Once index definitions are marked deleted, the drop-index callback reconciles every segment in the collection. Deleted definitions are excluded from the aggregate, so the reported vector-index size is reduced immediately.

internal/datacoord/task_index.go
yuruiz: Fixed. DataCoord now reconciles finished vector-index sizes at index-inspector startup, covering both upgrade backfill and a crash between FinishTask and the separate segment-stats write.

Issue #52197, opened 2026-08-05
[Bug]: RBAC backup repeats tenant-wide grantee scans


shaoting-huang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master @ c552dfda0f
- Deployment mode(standalone or cluster): N/A -- rootcoord KV catalog backup path
- MQ type(rocksmq, pulsar or kafka): N/A
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): N/A
- CPU/Memory: N/A
- GPU: N/A
- Others: RBAC metadata with legacy 16-hex grantee IDs
```

### Current Behavior

`BackupRBAC` filters builtin roles and then calls `ListGrant` once per custom role. In metadata that still contains legacy 16-hex grantee IDs, each `ListGrant` call may perform a tenant-wide `GranteePrefix` scan through the legacy shared-ID ownership check, so backups with multiple custom roles repeat the same tenant-wide scan.

### Expected Behavior

`BackupRBAC` should load tenant-wide grantee keys at most once when custom roles exist, reuse that data for legacy shared-ID ownership checks, and still fail closed if two roles share a legacy grantee ID. Backups with only builtin roles should skip the tenant-wide grantee preload.

### Steps To Reproduce

```markdown
1. Create RBAC metadata with multiple custom roles whose grants use legacy 16-hex grantee IDs.
2. Run `BackupRBAC`.
3. Observe that each per-role `ListGrant` call can repeat a tenant-wide `GranteePrefix` scan through the legacy ownership check.
4. Add a second role with the same legacy grantee ID and observe that the backup path must still reject the shared ID instead of returning grants.
```

### Milvus Log

N/A -- root cause analysis and regression tests target rootcoord KV catalog behavior.

### Anything else?

Relevant code paths: `internal/metastore/kv/rootcoord/kv_catalog.go` (`BackupRBAC`, `ListGrant`, `ListPolicy`, `loadGranteeIDPrefixWithLoadedGrantees`).

PR #52198, opened 2026-08-05
fix: [RBAC] avoid repeated backup grantee scans

shaoting-huang:
- Add a lazy, memoized BackupRBAC grantee provider so tenant-wide `GranteePrefix` data is loaded only when a legacy 16-hex grantee ID needs a shared-ID collision check.
- Keep public `ListGrant` and existing `ListPolicy` behavior unchanged while preserving `Grantor.User` population during RBAC backup.
- Add regression coverage for no-legacy backups avoiding the extra full-tenant scan, lazy snapshot timing after role-local reads, shared legacy ID fail-closed behavior, and invalid loaded grantee data.

related: #52197

### DIFF
--- a/internal/metastore/kv/rootcoord/kv_catalog.go
+++ b/internal/metastore/kv/rootcoord/kv_catalog.go
@@ -1514,7 +1514,21 @@ func (kc *Catalog) loadGranteeIDPrefix(ctx context.Context, tenant string, grant
 	return kc.loadGranteeIDPrefixWithLoadedGrantees(ctx, tenant, granteeKey, idStr, nil, nil)
 }
 
+func validateLoadedGrantees(granteeKeys []string, granteeValues []string) error {
+	if (granteeKeys == nil) != (granteeValues == nil) {
+		return merr.WrapErrServiceInternalMsg("loaded grantee keys and values must both be nil or both be non-nil, got keys=%d values=%d", len(granteeKeys), len(granteeValues))
+	}
+	if granteeKeys != nil && len(granteeKeys) != len(granteeValues) {
+		return merr.WrapErrServiceInternalMsg("loaded grantee keys and values must have equal length, got keys=%d values=%d", len(granteeKeys), len(granteeValues))
+	}
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -2110,12 +2132,32 @@ func (kc *Catalog) BackupRBAC(ctx context.Context, tenant string) (*milvuspb.RBA
 		return entity.GetRole(), true
 	})
 
+	var granteeKeys []string
+	var granteeValues []string
+	if len(roleEntity) > 0 {
```
liliu-z, line 2201: The preload runs unconditionally whenever the tenant has at least one custom role, but the snapshot is only consumed inside loadGranteeIDPrefixWithLoadedGrantees when a grant carries a legacy (16-char) grantee ID, so on a cluster with no legacy IDs — the normal state after migration — every backup now pays one extra full-tenant grantee scan (all keys and values) that is never read. The per-role scans at kv_catalog.go:1785 also remain even though the snapshot is a strict superset of what they return, so the new scan is additive rather than replacing anything; driving role enumeration off the snapshot, or loading it lazily on the first legacy ID as DeleteGrant already does at line 1999, would make it free instead. Reusing the snapshot for enumeration would also close the new staleness window, where a revoke landing between the preload and a per-role read makes the backup fail with 'shared legacy grantee id' for a sharing that no longer exists (the pre-PR fresh scan at check time would not have seen it).

shaoting-huang: Implemented the lazy path on aed5b947b5: BackupRBAC now creates a memoized provider but the tenant-wide GranteePrefix load only runs inside the legacy 16-hex collision branch, after the role-local grant read found that legacy ID. Full-length grantee ID backups no longer run the extra tenant-wide scan. Added regressions for the no-legacy path, stale eager-preload timing, shared legacy ID fail-closed behavior, and invalid loaded tenant keys.

zhengbuqian (approved):

Issue #51162, opened 2026-07-08
[Feature]: Support element-level nullability for Array/ArrayOfVector fields


SpadeA-Tang:
Add element-level nullability support for ARRAY and ArrayOfVector fields.

implementation:
- https://github.com/milvus-io/milvus/pull/52203
- https://github.com/milvus-io/milvus/pull/51164
- https://github.com/milvus-io/milvus/pull/52398
- https://github.com/milvus-io/milvus/pull/52399
- https://github.com/milvus-io/milvus/pull/52400
- https://github.com/milvus-io/milvus/pull/52401
- https://github.com/milvus-io/milvus/pull/52402
- https://github.com/milvus-io/milvus/pull/52403
- https://github.com/milvus-io/milvus/pull/52404

PR #52203, opened 2026-08-05
fix: adapt to the valid_data change in milvus-proto

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/51162


go.mod
```diff
@@ -69,7 +69,7 @@ require (
 	github.com/hamba/avro/v2 v2.29.0
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/magiconair/properties v1.8.7
-	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260712212354-7efdf25099dd
+	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260805030012-e994281e8de4
```
buqian-zilliz: **[P1] Gate element-null until nested masks are supported**

The upgraded proto revision is the `support element-null` change, so `FieldSchema.element_nullable` and nested `ScalarField/VectorField.valid_data` become part of the accepted API. This branch has no production `ElementNullable` handling or rejection, and the client, proxy, and core paths discard those inner masks (`Array`/`VectorArray` keep only payload values). A row such as `[value, NULL, value]` is therefore persisted or returned as shifted compact values or as concrete zero/vector values, corrupting element indices and query results. Please include the end-to-end validation, storage, and readback support in this change, or reject or pin the feature until that support lands.

pkg/util/typeutil/field_data.go
```diff
@@ -5,6 +5,50 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 )
 
+// GetFieldDataValidData returns the validity of the immediate values carried by
+// FieldData. New payloads store it on ScalarField or VectorField; FieldData is
+// retained as a legacy fallback for older payloads.
+func GetFieldDataValidData(fieldData *schemapb.FieldData) []bool {
+	if validData := fieldData.GetValidData(); len(validData) > 0 {
```
buqian-zilliz: **[P2] Honor field-specific validity before the legacy bitmap**

`GetFieldDataValidData` gives the deprecated top-level `FieldData.valid_data` precedence over `ScalarField.valid_data` / `VectorField.valid_data`, while the upgraded proto contract defines the field-specific bitmap as authoritative and the top-level field only as fallback. Several internal result, merge, and export paths call this helper, or the identically ordered client and C++ helpers, without a conflict check, so a payload containing both sources can silently apply stale null bits and select, drop, or reorder the wrong rows. Check the field-specific bitmap first in all three helpers, then fall back to the legacy field.

zhengbuqian: I think it makes sense to favor the deprecated valid_data more in the schema.proto evolution phase. 

but should we apply more lightweight checks? such as "length of valid data must equal to number of values".

pkg/util/typeutil/field_data.go
```diff
@@ -5,6 +5,50 @@ import (
...
+	}
+	if scalars := fieldData.GetScalars(); scalars != nil {
+		return scalars.GetValidData()
+	}
+	return fieldData.GetVectors().GetValidData()
+}
+
+// SetFieldDataValidData writes validity to the current field-specific location
+// and clears the legacy FieldData.valid_data source.
+func SetFieldDataValidData(fieldData *schemapb.FieldData, validData []bool) {
```
zhengbuqian, line 32: do we call this function to process the response before returning to the user? will there be compatibility issues if the user was still using the old SDK/proto?

internal/proxy/validate_util.go
```diff
@@ -110,6 +110,9 @@ func (v *validateUtil) Validate(data []*schemapb.FieldData, helper *typeutil.Sch
 		return merr.WrapErrServiceInternal("nil schema helper provided for Validation")
 	}
 	for _, field := range data {
+		if typeutil.HasFieldDataValidDataConflict(field) {
```
zhengbuqian: why? we should allow the user/sdk to set both? for example the requests sent by a new sdk should set both so the request can be understood by both the old/new milvus?

pkg/util/typeutil/field_data.go
SpadeA-Tang: > I think it makes sense to favor the deprecated valid_data more in the schema.proto evolution phase.
> 
> but should we apply more lightweight checks? such as "length of valid data must equal to number of values".

Vertification should be performed in the earlier phase such as pre execute in insert. It just a warpper of GetValidData() which is used directly before and has no checks.

internal/proxy/validate_util.go
SpadeA-Tang: Will we have the case that user uses new sdk to access a old milvus backend?

zhengbuqian: you can't stop the user from doing that.

pkg/util/typeutil/field_data.go
```diff
@@ -1,10 +1,127 @@
...
+	if fieldData == nil {
+		return
+	}
+	fieldData.ValidData = nil
+	switch field := fieldData.Field.(type) {
+	case *schemapb.FieldData_Scalars:
+		if field.Scalars == nil {
```
zhengbuqian: what does `field.Scalars == nil` mean in such case? is it possible?

pkg/util/typeutil/field_data.go
```diff
@@ -1,10 +1,127 @@
...
+			field.Vectors = &schemapb.VectorField{}
+		}
+		field.Vectors.ValidData = validData
+	}
+}
+
+// HasFieldDataValidDataConflict reports whether the legacy and current
+// validity locations are both populated with different values. Struct fields
+// are checked recursively.
+func HasFieldDataValidDataConflict(fieldData *schemapb.FieldData) bool {
```
zhengbuqian: nit: how about rename this to validate valid data? returning a false representing good and true representing bad is somewhat confusing..

pkg/util/typeutil/field_data.go
```diff
@@ -1,10 +1,127 @@
...
+	if legacy := fieldData.GetValidData(); len(legacy) > 0 {
+		return legacy
+	}
+	var current []bool
+	if scalars := fieldData.GetScalars(); scalars != nil {
+		current = scalars.GetValidData()
+	} else if vectors := fieldData.GetVectors(); vectors != nil {
+		current = vectors.GetValidData()
+	}
+	return current
```
zhengbuqian: return getFieldSpecificValidData to avoid duplicate code

pkg/util/typeutil/field_data.go
```diff
@@ -1,10 +1,127 @@
...
+}
+
+// SetFieldDataValidData writes validity to the current field-specific location
+// and clears the legacy FieldData.valid_data source.
```
zhengbuqian, line 31: to be symmetric, rename to SetFieldSpecificValidData?

pkg/util/typeutil/field_data.go
```diff
@@ -1,10 +1,127 @@
...
+}
+
+func getFieldSpecificValidData(fieldData *schemapb.FieldData) []bool {
+	if scalars := fieldData.GetScalars(); scalars != nil {
+		return scalars.GetValidData()
+	}
+	return fieldData.GetVectors().GetValidData()
+}
+
+func normalizeFieldDataValidData(fieldData *schemapb.FieldData) {
```
zhengbuqian, line 79: normalizeFieldDataValidData is called only in ValidateAndNormalizeFieldDataValidData, consider making it a lambda.

pkg/util/typeutil/field_data.go
```diff
@@ -1,10 +1,127 @@
 package typeutil
 
 import (
+	"slices"
+
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 )
 
+// GetFieldDataValidData returns the validity of the immediate values carried by
```
zhengbuqian, line 20: I'd prefer more consistent method naming:

* first we categorize the 2 types of valid data: legacy valid data and field specific valid data(or new valid data for brevity)
* then we can have a clear set of helper methods: 
  * `GetFieldDataValidData`: prefer legacy if present, or else new valid data
  * `ValidateFieldDataValidDataKeepOnlyNew`
  * `Get/Set FieldDataNewValidData` now means the field specific valid data, maybe a bool argument to indicate whether or not to unser the legacy valid data.
  * `Get/Set FieldDataLegacyValidData` now means the legacy valid data

the current code works but is somewhat messy

pkg/util/typeutil/field_data.go
zhengbuqian: this can also just be a lambda in ValidateAndNormalizeFieldDataValidData

pkg/util/typeutil/field_data.go
zhengbuqian: that's the idea, we may do not need all those helpers. and be careful with the method scope: some do not need to be exported.

internal/proxy/impl.go
```diff
@@ -2877,9 +2877,24 @@ func (node *Proxy) Search(ctx context.Context, request *milvuspb.SearchRequest)
 	if err != nil {
 		rsp.Status = merr.Status(err)
 	}
+	projectSearchResultValidDataForLegacy(rsp)
 	return rsp, nil
 }
 
+func projectSearchResultValidDataForLegacy(result *milvuspb.SearchResults) {
```
zhengbuqian, line 2903: is this the same as the helper in typeutil?

internal/proxy/util.go
```diff
@@ -1216,6 +1216,21 @@ func validateFieldDataColumns(columns []*schemapb.FieldData, schema *schemaInfo)
 	return nil
 }
 
+// validateAndNormalizeFieldDataValidData validates compatibility fields once
+// when a user payload enters the proxy, then keeps only the current
+// field-specific representation for internal processing.
+func validateAndNormalizeFieldDataValidData(fields []*schemapb.FieldData) error {
```
zhengbuqian, line 1222: ditto? if different, why not place it there?

internal/core/src/common/Utils.h
```diff
@@ -46,6 +46,35 @@ namespace milvus {
 #define VEC_FIELD_DATA(data_array, type) \
     (data_array->vectors().type##_vector().data())
 
+inline const google::protobuf::RepeatedField<bool>&
+GetFieldDataRowValidData(const DataArray& field_data) {
+    if (field_data.valid_data_size() > 0) {
```
zhengbuqian: is it still possible for C++ code to see legacy valid data from the input? is it still necessary for C++ code to set that in the output?

if we did all the transition in golang

internal/core/src/exec/expression/GISConjunctExpr.cpp
```diff
@@ -445,7 +446,7 @@ PhyGISRefineConjunctExpr::Eval(EvalCtx& context, VectorPtr& result) {
             auto geometry_array =
                 static_cast<const milvus::proto::schema::GeometryArray*>(
                     &data_array->scalars().geometry_data());
-            const auto& vd = data_array->valid_data();
+            const auto& vd = GetFieldDataRowValidData(*data_array);
```
zhengbuqian, line 449: seeing this makes me wonder: is it possible to disallow accessing the valid_data member in the future code changes through specific lint rules? unless NOLINT is used. so we can prevent unwanted regression. this holds for both c++ and golang code.

for golang we have mature Go ruleguard ready to use, and for C++ we can use clang-query wrapper, it should be ready but seems not enabled yet. 

we can do this in a separate PR. I suggest we do this asap. in the fuure type_schema we will also rely a lot on this to enture code correctness.

internal/proxy/impl.go
SpadeA-Tang: This handles SearchReults which the helper in typeutil handles FieldData.

internal/proxy/util.go
SpadeA-Tang: It should stay in proxy: typeutil handles a single FieldData, while this helper traverses multiple fields and provides proxy-level  error handling.

internal/core/src/common/Utils.h
SpadeA-Tang: Theretically, it should be specific valid data all the time in C++. We use consider legacy for defensive purpose.

internal/core/src/exec/expression/GISConjunctExpr.cpp
SpadeA-Tang: Added a type-aware Ruleguard lint rule to prevent direct access to valid_data on schemapb.FieldData, ScalarField, and VectorField.

  It rejects:

  - Direct field reads and writes
  - GetValidData() calls and method values
  - ValidData composite-literal initialization
  - Both pointer and value receivers

zhengbuqian (approved):

Issue #52105, opened 2026-08-01
[Bug]: Row-copying compaction fills a null geometry cell with an empty value instead of the field default


MrPresent-Han:
### Problem

During row-copying record rebuilds, `appendValueAt` (`internal/storage/arrow_util.go:185-198`) materializes a field's default value into null cells. The function dispatches on the arrow builder type, and its `*array.BinaryBuilder` branch reads the default via `GetBytesData()` — correct for JSON, whose defaults are carried in `BytesData`. Geometry columns share the same `BinaryBuilder` (`serdeMap[Geometry]` → arrow Binary), but geometry defaults are carried as WKT text in `StringData` (enforced at DDL time by `checkGeometryDefaultValue`, `internal/rootcoord/util.go:368`). The protobuf getter returns `nil` for the mismatched oneof, and `BinaryBuilder.Append(nil)` writes a `valid=true, length=0` value. A null geometry cell is therefore rewritten as a zero-length value instead of the WKB form of the default. VarChar and JSON cells on the same path are handled as intended.

Downstream, `ConvertWKBToWKT` fails on a zero-length input (`internal/proxy/validate_util.go:68-78`), so the compaction itself succeeds and publishes the segment, while a later Query/Search touching the field reports an error.

### Trigger channels

**Channel 1 — Parquet import (verified end-to-end, default configuration).** `ReadNullableGeometryData` (`internal/util/importutilv2/parquet/field_reader.go:814-840`) fills the default WKB into null slots but returns the original `validData` unchanged — unlike the generic `fillWithDefaultValueImpl` (`:287-308`), it never marks those rows valid. The backing default is then dropped by hashing (`GeometryFieldData.GetRow` returns nil for invalid rows) and the rows persist as genuine nulls through V1/V2/V3. Since import schedules sort compaction automatically under default configuration (`enableSortCompaction()` = `dataCoord.enableCompaction` ∧ `dataCoord.sortCompaction.enable`, both default true; `internal/datacoord/import_checker.go:361-441`), the rewrite happens within the same import lifecycle:

1. Collection has a nullable geometry field with a WKT default.
2. A Parquet file contains that column with at least one NULL cell.
3. Import completes; sort compaction is scheduled and runs.
4. The null cells are written out as zero-length values.

[...]

PR #52204, opened 2026-08-05
fix: Preserve geometry defaults in rebuilds

congqixia:
issue: #52105

Geometry fields store defaults as WKT strings, while JSON fields use byte defaults. Row-copy rebuilds only read binary defaults from BytesData, so null geometry cells could be rewritten as valid empty WKB values during compaction.

Pass the field schema into the row-copy helper so geometry defaults are converted from WKT to WKB before appending. Also make the Parquet geometry reader mark default-filled null rows valid, matching the generic nullable default handling used by other scalar readers.

Add regression coverage for geometry defaults in empty-array generation, RecordBuilder rebuilds, and Parquet nullable geometry imports.

### DIFF
--- a/internal/storage/arrow_util.go
+++ b/internal/storage/arrow_util.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/storagev2/packed"
+	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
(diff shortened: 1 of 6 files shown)


internal/storage/arrow_util.go
```diff
@@ -189,6 +191,14 @@ func appendValueAt(builder array.Builder, a arrow.Array, idx int, defaultValue *
 		if ba.IsNull(idx) {
 			// could be internal $meta json
 			if defaultValue != nil {
+				if field.GetDataType() == schemapb.DataType_Geometry {
+					val, err := common.ConvertWKTToWKB(defaultValue.GetStringData())
```
MrPresent-Han: Could we precompute the geometry default WKB once per field before entering the row-copy loop? `appendValueAt` is called for every emitted row from both `RecordBuilder.Append` and `Sort`, and this branch runs for every null geometry value. `ConvertWKTToWKB` performs a full WKT parse and WKB serialization with temporary allocations, so historical segments containing many null geometry rows may repeat the same conversion a large number of times during compaction. Caching the converted `[]byte` in the builder/Sort setup and reusing it here would preserve the behavior while avoiding the per-row parsing cost. A many-null-row benchmark or regression test would also help cover this path.

MrPresent-Han (commented):
One performance consideration on the row-copy path.

Issue #52202, opened 2026-08-05
[Bug]: Snapshot restore resets consistency level and drops collection properties


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: master-20260805-c552dfda0f7
- Git commit: c552dfda0f7, reported by get_server_version(detail=True) from the affected standalone instance
- Build time: Wed Aug 5 09:01:18 UTC 2026
- Deployment mode: Standalone on Kubernetes
- MQ type: Woodpecker
- SDK version: PyMilvus 3.1.0rc62
- Python version: 3.12.12
- OS: Linux container
- CPU/Memory: Not separately recorded
- GPU: None
- Pod status:
  - snapshot-master-longrun-etcd-0: Running, Ready 1/1, 0 restarts
  - snapshot-master-longrun-milvus-standalone-5dfd5477b9-s2ct4: Running, Ready 1/1, 0 restarts
  - snapshot-master-longrun-minio-57c66c9797-g4lqr: Running, Ready 1/1, 0 restarts

### Current Behavior

Restoring a collection snapshot completes successfully, but the restored collection does not preserve the source collection metadata:

- Consistency level changes from Bounded (2) to Strong (0).
- User-configured collection properties are missing.
- The restore job still reports RestoreSnapshotCompleted.

Observed property mismatches:

~~~text
collection.ttl.seconds:
  source: "360"
  restored: missing

collection.autocompaction.enabled:
  source: "False"
  restored: missing

mmap.enabled:
  source: "False"
  restored: missing

allow_insert_auto_id:
  source: "False"
  restored: missing
~~~

### Expected Behavior

Snapshot restore should preserve:

1. The source collection consistency level.
2. Every property returned by describe_collection(source)["properties"], with the same value.

Restore-specific internal properties may be added to the target collection, but source properties must not be removed or changed.

Aliases are excluded from this report because aliases are mutable database-scoped bindings rather than intrinsic collection properties. Their snapshot/restore semantics require a separate product decision.

### Steps To Reproduce

Set MILVUS_URI and, when authentication is enabled, MILVUS_TOKEN, then run:

~~~python
import json
import os
import time
from uuid import uuid4

from pymilvus import MilvusClient

client_args = {"uri": os.environ["MILVUS_URI"]}
if os.environ.get("MILVUS_TOKEN"):
    client_args["token"] = os.environ["MILVUS_TOKEN"]

client = MilvusClient(**client_args)

[...]

PR #52205, opened 2026-08-05
fix: preserve collection metadata during snapshot restore

weiliu1031:
issue: #52202

## What changed

- Store the collection consistency level and properties in snapshot metadata.
- Restore those values while isolating restore-only property mutations.
- Recover metadata from schema properties for snapshots created before this
  fix.
- Cover current snapshots, legacy snapshots, and the Go client restore flow.

## Why

CollectionDescription already defined canonical fields for consistency and
properties, but snapshot generation left them unset. Restore consumed those
empty fields, so the target silently used Strong consistency and lost source
collection settings despite reporting success.

## Behavior boundary

The restored collection now preserves source consistency and collection
properties. Restore may still add internal properties such as
preserve_field_ids. Alias behavior is unchanged.

## Validation

Before rebasing onto the latest master:

- `GOTOOLCHAIN=go1.26.5 make milvus`
- DataCoord targeted tests with `-tags dynamic,test` and
  `-gcflags="all=-N -l"`:
  - `TestGenSnapshot_PreservesCollectionMetadata`
  - `TestSnapshotManager_RestoreCollection_PreservesMetadata`
- Go client E2E with `-tags dynamic,test` and
  `-gcflags="all=-N -l"`:
  - `TestSnapshotRestoreEmptyCollection`
- `gofmt`
- `git diff --check`

The rebased commit retained the same feature patch-id and touched paths. The
post-rebase build and tests were not rerun.

### DIFF
--- a/internal/datacoord/handler.go
+++ b/internal/datacoord/handler.go
@@ -874,6 +874,8 @@ func (h *ServerHandler) GenSnapshot(ctx context.Context, collectionID UniqueID) 
 			Schema:              schema,
 			NumShards:           int64(resp.GetShardsNum()),
 			NumPartitions:       resp.GetNumPartitions(),
+			ConsistencyLevel:    resp.GetConsistencyLevel(),
+			Properties:          common.CloneKeyValuePairs(resp.GetProperties()),
 			Partitions:          partitionMapping,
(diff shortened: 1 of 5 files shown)


congqixia (approved):
/lgtm

PR #52206, opened 2026-08-05
fix: [3.0] preserve collection metadata during snapshot restore

weiliu1031:
issue: #52202
pr: #52205

## What changed

Backport the snapshot metadata fix to the 3.0 branch:

- Store the collection consistency level and properties in snapshot metadata.
- Restore those values while isolating restore-only property mutations.
- Recover metadata from schema properties for snapshots created before the
  fix.
- Add regression coverage for current snapshots, legacy snapshots, and the
  Go client restore flow.

## Why

Snapshot generation left the canonical consistency and property fields unset.
Restore consumed those empty fields, causing the target collection to use
Strong consistency and lose source collection settings despite reporting
success.

## Branch notes

The affected snapshot generation and restore paths match master. The change
applies cleanly without 3.0-specific behavior changes.

## Validation

- `git diff --check`
- Build and tests were not run for this 3.0 backport.

### DIFF
--- a/internal/datacoord/handler.go
+++ b/internal/datacoord/handler.go
@@ -847,6 +847,8 @@ func (h *ServerHandler) GenSnapshot(ctx context.Context, collectionID UniqueID) 
 			Schema:              schema,
 			NumShards:           int64(resp.GetShardsNum()),
 			NumPartitions:       resp.GetNumPartitions(),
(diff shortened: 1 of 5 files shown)


congqixia (approved):
/lgtm

PR #51232, opened 2026-07-10
fix: Avoid structured binding capture in OpenMP build

sijie-ni-0214:
## Summary

This PR avoids directly capturing a structured binding in `ChunkedSegmentSealedImpl::LoadBatchJsonKeyIndexes`.

Clang rejects direct structured-binding captures when compiling this code path with OpenMP enabled. The lambda now uses a generalized init-capture for `field_id`, preserving the existing value-capture behavior while avoiding the unsupported capture form.

issue: #51231

## Test Plan

- [x] `git diff --check`
- [x] `/tmp/milvus-clang-format-15-bin/clang-format-15 --dry-run --Werror --style=file internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp`
- [x] `make SKIP_3RDPARTY=1`
- [x] `PATH="/tmp/milvus-clang-format-15-bin:$PATH" make verifiers`

PR #52207, opened 2026-08-05
fix: avoid structured binding capture in OpenMP build

zhikunyao:
issue: #51231

### What changed

- Use explicit init-captures for `cg_index` and `field_ids` in `ChunkedSegmentSealedImpl::LoadColumnGroups`.
- Avoid directly capturing structured-binding variables when OpenMP is enabled.

### Why

PR #52036 introduced a structured-binding loop and captured both bindings in the submitted thread-pool lambda. Clang rejects those captures when compiling with OpenMP enabled, causing the MacOS Code Checker to fail with:

```text
error: capturing a structured binding is not yet supported in OpenMP
```

This follows the same explicit init-capture pattern used by #51232 and #51302.

### Verification

- Reproduced the failure with Homebrew Clang 17 and `-fopenmp=libomp`; the explicit init-capture variant passes syntax compilation.
- `CLANG_FORMAT=/opt/homebrew/opt/llvm@15/bin/clang-format make cppcheck`
- `git diff --check`
- MacOS Code Checker will be triggered with `/ci-rerun` after the PR is created.


zhikunyao:
MacOS Code Checker validation passed for commit `b07e0b13cb19b4eac4a97a3dae736cdde0d31d29`.

- Run: https://github.com/milvus-io/milvus/actions/runs/31024626335
- Checkout: exact PR commit `b07e0b13cb19b4eac4a97a3dae736cdde0d31d29`
- `Code Check`: success
- Overall conclusion: success

The `/ci-rerun-code-check-macos` comment was accepted by Jenkins, but its rerun helper skipped because it searches for an existing `pull_request` run while `.github/workflows/mac.yaml` is currently `workflow_dispatch`-only. I therefore dispatched the same `mac.yaml` workflow manually from `master`, with the workflow input pinned to the exact PR commit above.

Issue #52208, opened 2026-08-05
[Enhancement]: avoid allocations in typeutil murmur3 hash helpers


bigsheeper:
## What would you like to be improved?

`pkg/util/typeutil.Hash32Bytes` currently creates a new `murmur3.New32()` hasher for each call, and `Hash32Uint64` allocates an 8-byte slice before hashing. These helpers are used by int64 routing hash paths such as proxy insert/delete routing and import partition hashing, so the per-key allocations show up in hot routing loops.

## Why is this useful?

The same observable hash output can be produced without per-call heap allocation:

- `murmur3.Sum32(b)` is equivalent to creating a 32-bit Murmur3 hasher, writing `b`, and reading `Sum32()` for a one-shot byte slice.
- `Hash32Uint64` can encode into a stack `[8]byte` before calling `Hash32Bytes`.
- Existing exact-value and routing-modulo tests already pin the output behavior.

## Proof

Existing behavior-pinning tests pass:

```bash
cd pkg
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -run "TestHash32_Uint64|TestHash32_String|TestHash32_Uint64_Value|TestHashPK2Channels|TestHashPK2ChannelsMatchesLegacyModulo|TestLocateRoutingIndexesMatchesLegacyModulo|TestHashKey2PartitionsMatchesLegacyModulo" ./util/typeutil
```

Before benchmark on current `milvus-io/master` with added local benchmark coverage:

```text
BenchmarkHash32Bytes-2     128.9-131.5 ns/op   80 B/op  1 allocs/op
BenchmarkHash32Uint64-2    155.9-158.7 ns/op   88 B/op  2 allocs/op
BenchmarkHash32Int64-2     156.5-162.5 ns/op   88 B/op  2 allocs/op
```

After using the one-shot Murmur3 hash and a stack uint64 buffer:

```text
BenchmarkHash32Bytes-2     36.39-37.14 ns/op   0 B/op  0 allocs/op
BenchmarkHash32Uint64-2    31.65-32.72 ns/op   0 B/op  0 allocs/op
BenchmarkHash32Int64-2     34.95-35.35 ns/op   0 B/op  0 allocs/op
```

## Duplicate search

No matching existing issue or PR was found for:

- `Hash32Bytes Hash32Uint64 murmur3 typeutil allocation performance`
- `HashPK2Channels locateRoutingIndexes HashRoutingTable typeutil hash routing performance`

PR #52209, opened 2026-08-05
enhance: avoid hash helper allocations

bigsheeper:
issue: #52208

## What this PR does

This PR removes per-call allocations from `pkg/util/typeutil` Murmur3 hash helpers used by int64 routing hash paths:

- `Hash32Bytes` now uses `murmur3.Sum32` directly instead of allocating a new hasher.
- `Hash32Uint64` now encodes into a stack `[8]byte` instead of allocating an 8-byte slice.
- The hash helpers move from `spaolacci/murmur3` to `twmb/murmur3` (see below).
- Adds benchmarks for `Hash32Bytes`, `Hash32Uint64`, and `Hash32Int64`.

The exported signatures and hash outputs stay unchanged.

## Why `twmb/murmur3`

`spaolacci/murmur3`'s one-shot `Sum32` walks the input through a bare `uintptr` and converts it back to an `unsafe.Pointer` on every 4-byte block (`murmur32.go:125-129`). The runtime's `checkptr` instrumentation — on for every `-race` build, which includes CI's `ut-integration` — rejects that outright, so any hash of four or more bytes becomes a **fatal error**, not a test failure. The first revision of this PR hit exactly that: `ut-integration` died in `TestHashPK2ChannelsIntegration` with `checkptr: pointer arithmetic result points to invalid allocation`.

Reverting `Hash32Bytes` to the streaming `New32()` + `Write()` form avoids `checkptr` but restores 80 B/op and 1 alloc/op, since `New32()` returns a `hash.Hash32` interface whose `digest32` necessarily escapes — that gives up the point of this PR.

`twmb/murmur3` is the maintained fork of the same algorithm and loads each block by index rather than by pointer arithmetic, so it is safe under `checkptr`. It was already present in the module graph as an indirect dependency; this PR promotes it to a direct dependency of `pkg` and adds the `go.sum` entry `tests/go_client` needs.

**Digests are unchanged.** `TestHash32_Uint64_Value` already pins 20 known outputs and passes. A differential run against `spaolacci`'s streaming hasher additionally matched on every length 0-64, the `int64` boundary values, 200k random inputs, and 100k sequential `uint64`s in the exact 8-byte routing shape. This matters because these values decide PK-to-channel routing, so a digest change would misroute data in existing collections.

## Numbers

`BenchmarkHash32Bytes` / `Hash32Uint64` / `Hash32Int64` (added here). Median of 5, `-gcflags="all=-N -l" -benchmem -count=5`; before and after measured back to back in the same session on the same machine.

time:

| | before | after |
|---|---|---|
| `Hash32Bytes` | 140.7 ns | **48.68 ns (2.89x)** |
| `Hash32Uint64` | 179.6 ns | **59.73 ns (3.01x)** |
| `Hash32Int64` | 162.0 ns | **62.70 ns (2.58x)** |

B/op:

| | before | after |
|---|---|---|
| `Hash32Bytes` | 80 | **0** |
| `Hash32Uint64` | 88 | **8 (-91%)** |
| `Hash32Int64` | 88 | **8 (-91%)** |

`Hash32Bytes` reaches zero allocations. `Hash32Uint64` and `Hash32Int64` retain a single 8-byte allocation rather than reaching zero: `twmb`'s `Sum32` passes the slice through `strslice`, which takes the address of its parameter, so escape analysis moves the caller's stack array to the heap (`moved to heap: b` under `-gcflags=-m`). Reaching zero would mean building the string header with `unsafe` in `typeutil` itself, which is the same class of trick that caused the `checkptr` failure above, so this PR does not do it.

## Verification

```bash
cd pkg
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/typeutil
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -run "^$" -bench "BenchmarkHash32(Bytes|Uint64|Int64)$" -benchmem -count=5 ./util/typeutil
```

Reproducing the original failure and confirming the fix — this is the instrumentation CI's `-race` build enables:

```bash
cd pkg
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l -d=checkptr" -count=1 ./util/typeutil
```

On master this passes; on the first revision of this PR it aborted with `checkptr: pointer arithmetic result points to invalid allocation`; with `twmb/murmur3` it passes again.

```bash
make static-check
```

`make static-check` result: 0 issues in core, pkg, client, and go_client e2e checks.


bigsheeper:
Pushed `55937fdc5d` — fixes the `ut-integration` failure on `d44cad8`, and rebases onto latest master (`a9d0453b`). No review comments outstanding on this PR; this note is just the root cause and what changed, since the failure was not obvious from the CI summary.

**What the CI failure actually was.** Not a failing assertion — a fatal runtime error that killed the whole `tests/integration` binary, which is why the log shows `FAIL github.com/milvus-io/milvus/tests/integration` repeatedly with no `--- FAIL:` line:

```
fatal error: checkptr: pointer arithmetic result points to invalid allocation
  murmur3.Sum32WithSeed (murmur32.go:129)
  typeutil.Hash32Bytes (hash.go:55)
  typeutil.Hash32Uint64 → Hash32Int64 → HashPK2Channels
  tests/integration.TestHashPK2ChannelsIntegration (util_insert_test.go:539)
```

`spaolacci/murmur3`'s one-shot `Sum32` converts the data pointer to a bare `uintptr`, does arithmetic on it, then converts back to `unsafe.Pointer` for each 4-byte block (`murmur32.go:125-129`). The runtime's `checkptr` instrumentation rejects that round-trip unconditionally, and `checkptr` is on for every `-race` build — which `ut-integration` is. So this fires for **any** input of four or more bytes; it is not flaky and not stack-vs-heap dependent. The previous streaming code (`New32()` + `Write()`) took a different path in the same library that reads block bytes by index, which is why master was fine.

I reproduced it in isolation before changing anything:

```bash
cd pkg && GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l -d=checkptr" -count=1 ./util/typeutil
# on d44cad8: fatal error, same stack
# on master:  ok
```

**Why not just revert `Hash32Bytes`.** Reverting to `New32()` + `Write()` does avoid `checkptr`, but measured 80 B/op, 1 alloc/op for `Hash32Bytes` and 88 B/op, 2 allocs/op for `Hash32Uint64` — i.e. back to master, giving up the PR.

**The fix.** Switch the helpers to `twmb/murmur3`, the maintained fork of the same algorithm, which loads blocks by index instead of by pointer arithmetic. It was already in the module graph as an indirect dependency; this promotes it to direct in `pkg` and adds the `go.sum` entry `tests/go_client` needs. Five files, +12/-5.

**Digest stability — the part worth checking.** These values decide PK-to-channel routing, so a changed digest would misroute data in existing collections. Two independent checks:

- `TestHash32_Uint64_Value` already pins 20 known outputs; unchanged and passing.
- A differential run against `spaolacci`'s streaming hasher matched bit-for-bit on every length 0-64 (three fill patterns), the `int64` boundary values in both endiannesses, 200k random inputs, and 100k sequential `uint64`s in the exact 8-byte routing shape.

Worth noting `tests/integration/util_insert.go` still computes its reference hash with `spaolacci` streaming, so `TestHashPK2ChannelsIntegration` now cross-validates the two libraries against each other rather than just checking our code against itself.

**One thing I did not achieve, stated plainly.** `Hash32Uint64` / `Hash32Int64` keep **one 8-byte allocation** instead of reaching zero. `twmb`'s `Sum32` routes through `strslice`, which takes the address of its parameter, so escape analysis moves the caller's stack array to the heap (`moved to heap: b` under `-gcflags=-m`). Getting to zero would mean constructing the string header with `unsafe` inside `typeutil` — the same class of trick that caused this failure — so I left it. The PR description's benchmark table has been corrected accordingly; it previously claimed 0 allocs for all three. Net against master:

| | master | now |
|---|---|---|
| `Hash32Bytes` | 128.9-131.5 ns · 80 B · 1 alloc | 48.9-49.9 ns · 0 B · 0 alloc |
| `Hash32Uint64` | 155.9-158.7 ns · 88 B · 2 allocs | 60.3-63.1 ns · 8 B · 1 alloc |
| `Hash32Int64` | 156.5-162.5 ns · 88 B · 2 allocs | 64.3-66.3 ns · 8 B · 1 alloc |

`make static-check`: 0 issues across core, pkg, client, go_client.

czs007:

Verified:
- `pkg/util/typeutil/hash.go:74` — the new `murmur3.Sum32(b) & 0x7fffffff` is digest-equivalent to the old implementation, and the production routing path is pinned by 20 hardcoded expected values in `hash_test.go:74-133`, which would fail on any divergence.
- `pkg/util/typeutil/hash.go:84-86` — int64 PK/partition-key routing always feeds exactly 8 bytes (two full blocks, no tail path) via `common.Endian` (LittleEndian), so byte order, buffer size, and the 31-bit mask are all preserved.
- `pkg/util/typeutil/hash.go:58-60` — independent differential testing across 257,000 random inputs of lengths 0–256 produced no digest difference between the old and new libraries; the seed, 4-byte block, tail, and fmix stages all match.
- `pkg/util/typeutil/hash.go:50-52` — string PK routing uses crc32, not murmur3, so the unpinned tail path is never exercised in production.
- `pkg/go.mod:72` / `go.sum:731-732` — twmb v1.1.6 was already the root module's selected version (previously indirect), so promoting it to a direct dependency introduces no version skew; the spaolacci downgrade to indirect in `pkg` is correct since no direct import remains there.
- `tests/integration/util_insert.go:362` — the integration suite still uses spaolacci as an independent oracle, providing continuous cross-validation that the two libraries produce identical digests.

pkg/util/typeutil/hash.go
```diff
@@ -52,21 +52,19 @@ func (stringRoutingHasher) Hash(key string) (uint64, error) {
...
+// Keep this on twmb/murmur3: spaolacci/murmur3's Sum32 walks the input through a
+// bare uintptr, which the runtime's checkptr instrumentation (on for every -race
+// build) rejects. Both produce identical digests.
 func Hash32Bytes(b []byte) (uint32, error) {
-	h := murmur3.New32()
-	if _, err := h.Write(b); err != nil {
-		return 0, err
-	}
-	return h.Sum32() & 0x7fffffff, nil
+	return murmur3.Sum32(b) & 0x7fffffff, nil
```
czs007, line 60: nit (non-blocking): after this change, first-party code imports two different murmur3 libraries.

This file now uses `twmb`, while `tests/integration/util_insert.go:26` still computes the same routing hash with `spaolacci`'s streaming hasher — `hashInt64ForChannel` (util_insert.go:355-364) is a hand-copied mirror of `Hash32Int64`, and its comment still says "same as typeutil.Hash32Bytes", which now names a different library than the one `typeutil` calls.

To be clear about what is and isn't changing here: neither library can be dropped, and this PR does not add one. Both were already in the graph on master — `spaolacci` via `pkg/v3/mq/msgstream → pulsar-client-go/pulsar/internal`, `twmb` via `pkg/v3/util/tikv → tikv/client-go/v2/internal/latch` (`latch.go:46`). What does change is the invariant: the routing digest now has to agree across two independent third-party implementations. I checked that it does — a differential run of `spaolacci` streaming vs `twmb.Sum32` (both masked with `& 0x7fffffff`) over all lengths 0-256, the `int64` boundary values in the 8-byte little-endian routing shape, 500k random `uint64` in that same shape, and 200k random variable-length blobs gave 0 mismatches. So this is not a correctness problem today.

The suggestion is to remove the coupling rather than depend on it: make `hashInt64ForChannel` call `typeutil.Hash32Int64` directly. `util_insert.go` does not import `typeutil` today, but `util_insert_test.go:26` already does, so there is no cycle — it's a one-line change. After it, `twmb` is the only murmur3 that first-party code imports, and `spaolacci` drops to indirect in the root `go.mod` as well.

Worth noting because the safety net is thinner than it looks: the only thing currently guarding the cross-library invariant is `TestHashPK2ChannelsIntegration` (util_insert_test.go:513), and it asserts *distribution* ("each channel should have exactly numRows/numChannels PKs"), not per-value equality. It would probably go red if the two libraries ever diverged, but the failure would read as a balance problem rather than a hash mismatch.

bigsheeper: Taken in `9d773410ed`. I verified each premise against the head you reviewed (`59a983d5`) before changing anything, and there is one place where your description is more conservative than the facts.

**Confirmed.** `tests/integration/util_insert.go:26` was the **only** remaining first-party import of `spaolacci/murmur3` in the repo — I grepped rather than trusting the claim. `hashInt64ForChannel` (`:355-364`) was indeed a hand-copied mirror of `Hash32Int64` down to the `& 0x7fffffff` mask, and its comment named `typeutil.Hash32Bytes`, which by then called a different library. `TestHashPK2ChannelsIntegration` (`util_insert_test.go:513-553`) does assert distribution (`channelCounts[ch] == numRows/numChannels`), not per-value equality, so it was the only guard and it was the wrong shape for the invariant.

**Where you understated it.** You wrote that `util_insert.go` does not import `typeutil` today but `util_insert_test.go:26` already does, "so there is no cycle". True, but it was never even a test-only question: `util_query.go:38` — a non-test file in the same `integration` package — already imports `typeutil`. There was no import-graph consideration at all.

**What changed.** `hashInt64ForChannel` now calls `typeutil.Hash32Int64` directly; `encoding/binary` and the `spaolacci` import are gone. `go mod tidy` moved `spaolacci/murmur3` from the direct block to the indirect one in the root `go.mod` (one line; `go.sum` unchanged), so `twmb` is now the only murmur3 first-party code imports.

**On the safety net — I am deliberately not adding the per-value assertion you implied was missing.** Before the change, a differential test would have been the right guard, and I ran one myself while evaluating this (`pk = 1 / 42 / -1 / 1<<62` in the 8-byte little-endian routing shape: `pk=42` → `1871679806` from both libraries → shard 2 under 4 shards). After the change there is nothing left to assert: the helper *is* the production function, so divergence is impossible by construction rather than merely unobserved. Adding a test that compares `Hash32Int64` against itself would be noise.

Verified locally: `make static-check` clean (0 issues across core/pkg/client/go_client), `go vet -tags dynamic,test ./tests/integration/` clean, and `TestGenerateBalancedInt64PKs` / `TestGenerateBalancedVarCharPKs` / `TestGenerateChannelBalancedPrimaryKeys` / `TestHashPK2ChannelsIntegration` all pass.

czs007:

Verified:
- `pkg/util/typeutil/hash.go:59-60` — `Hash32Bytes` preserves the zero-seed MurmurHash3_x86_32 digest and the existing `0x7fffffff` mask, so routing output is byte-identical after the library swap.
- `pkg/util/typeutil/hash_test.go:76-135` — twenty pinned hardcoded digests (e.g. `Hash32Int64(0)==1669671676`) act as a regression net; any digest drift from the replacement would fail the test immediately.
- `pkg/util/typeutil/hash.go:64-72` — the fixed `[8]byte` buffer is function-local, little-endian encoding is unchanged, and the `int64`→`uint64` conversion preserves the exact two's-complement bytes for negative values.
- `pkg/util/typeutil/hash.go:59-72` — the helpers are stateless pure functions with unchanged exported signatures; returning `nil` directly matches the old streaming hasher, whose `Write` never returned an error.
- `tests/integration/util_insert.go:319-356` — balanced-PK generation now calls the same `Hash32Int64` helper as production routing, eliminating drift between two hand-maintained implementations without changing channel selection.
- `pkg/go.mod` / `tests/go_client/go.mod` — no source file still imports `spaolacci/murmur3`, so demoting it to indirect and promoting `twmb/murmur3` to direct is self-consistent and survives `go mod tidy`.

czs007 (approved):
/lgtm

Issue #52212, opened 2026-08-06
[Bug]: REST insert stores invalid JSON for a JSON field holding a string or an out-of-range number


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (also present on 2.6 / 3.0)
- Deployment mode(standalone or cluster): any
- SDK version: REST API v1 / v2 (no SDK)
```

### Current Behavior

`checkAndSetData` stores a declared `JSON` field as `[]byte(fieldValue.String())`
(`internal/distributed/proxy/httpserver/utils.go`). `gjson`'s `String()` is a
rendering, not the original token:

- for a JSON string it strips the quotes
- for a number it falls back to `strconv.FormatFloat` of the parsed `float64`
  whenever the raw text is not all digits

So the bytes written to storage are not always a JSON document:

| request | stored bytes | valid JSON? |
|---|---|---|
| `{"j": "hello"}` | `hello` | **no** |
| `{"j": ""}` | (empty) | **no** |
| `{"j": "{not json"}` | `{not json` | **no** |
| `{"j": 9007199254740993.0}` | `9007199254740992` | yes, but silently rounded |
| `{"j": 10.0}` | `10` | yes, but retyped |

Nothing downstream catches this: the proxy's `checkJSONFieldData`
(`internal/proxy/validate_util.go`) only checks for a nil array and the max
length, it never validates that the bytes parse as JSON.

Those bytes are not a JSON document. With simdjson v3.12.2 (the version pinned
in `internal/core/thirdparty/simdjson/CMakeLists.txt`):

```
document hello   -> TAPE_ERROR: The JSON document has an improper structure
document "hello" -> parses OK
```

Verified end to end against a running standalone (`master-20260803`). The
failure is entirely silent, which makes it worse than an error:

- the insert returns `code=0`
- a query echoes `meta` back verbatim, so the row looks stored
- but the row matches **no** JSON predicate, including an exact match on the
  value that was written

```
insert {"id":2,"meta":"pending"}      -> stored bytes: pending      (invalid)
insert {"id":4,"meta":"\"pending\""}  -> stored bytes: "pending"    (valid)

filter meta == "pending"              -> matches id 4 only
filter meta["status"] == "ok"         -> does not match id 2
filter meta["status"] != "ok"         -> does not match id 2
filter not exists meta["status"]      -> matches id 2
```

[...]

PR #52213, opened 2026-08-06
fix: keep the original JSON token for REST JSON and dynamic fields

xiaofan-luan:
Two places in `checkAndSetData` destroyed data by re-rendering a JSON value
instead of keeping the token the client sent. Both store JSON text, so the
decode/encode round trip could only lose information.

## What changed

- Store a declared `JSON` field from the original `gjson` token (`Raw`) instead
  of `fieldValue.String()`.
- Keep the original number literal for a **dynamic field** via `json.Number`,
  and reject the values the JSON engine cannot represent.
- A JSON document supplied as a JSON string is still unwrapped, so that input
  form keeps working.
- Export `Valid` from `internal/json` (used for the unwrap check).
- Add table-driven coverage for every reachable JSON token kind, asserting both
  the exact stored bytes and that they remain a valid JSON document.

## Root cause

`gjson`'s `String()` renders a value rather than returning its token. For a JSON
string it strips the quotes; for a number it falls back to
`strconv.FormatFloat` of the parsed `float64` whenever the raw text is not all
digits. The bytes written to storage were therefore not always a JSON document:

| request | stored before | valid JSON? | stored after |
|---|---|---|---|
| `{"j": "hello"}` | `hello` | **no** | `"hello"` |
| `{"j": ""}` | (empty) | **no** | `""` |
| `{"j": "{not json"}` | `{not json` | **no** | `"{not json"` |
| `{"j": 9007199254740993.0}` | `9007199254740992` | rounded | `9007199254740993.0` |
| `{"j": 10.0}` | `10` | retyped | `10.0` |

Nothing downstream caught it: the proxy's `checkJSONFieldData` only checks the
array is non-nil and within `JSONMaxLength`. Segcore then cannot parse the
field — with simdjson v3.12.2 (the pinned version), a bare `hello` produces
`TAPE_ERROR`, so the whole JSON field of that row is unreadable.

## Dynamic field

The dynamic-field branch decoded numbers with a "does the raw text contain a
dot" heuristic and then re-encoded them. The integer side used `cast.ToInt64`,
which discards its error and returns `0`, so everything strconv could not parse
was silently stored as `0`:

|---|---|---|
| `{"x": 1e300}` | `0` | `1e300` |
| `{"x": 1e19}` | `0` | `1e19` |
| `{"x": 12345678901234567890}` | `0` | unchanged, exact |
| `{"x": 10.0}` | `10` | `10.0` |

Integers beyond 64 bits are now rejected rather than stored: simdjson reports
`BIGINT_ERROR` for those literals, so storing them would turn every query that
touches the path into an error. The message points at the two ways to express
such a value (a floating-point literal, or a string).

The same check is shared with a JSON field whose document is itself a number,
so the two field kinds no longer disagree about the same literal. Numbers
nested inside an object or array are deliberately not walked — that would cost
a full document scan on the insert path — and a test pins that boundary.

This also removes a decode/encode round trip: on a 12-field dynamic payload it
measures ~40% faster with ~40% fewer allocations.

## Why storing `Raw` is safe

The request body is validated as JSON by the gin binder
(`ShouldBindBodyWith(req, binding.JSON)`) before `checkAndSetData` runs, and the
same cached bytes are what `checkAndSetData` parses. Every `gjson.Result.Raw`
inside a validated body is a well-formed JSON value by construction, so no extra
validation pass is needed on the insert path.

The binder also gates the number range: `data` is decoded into
`[]map[string]interface{}`, so values outside the float64 range (`1e400`,
`-1e309`) are rejected at bind time and never reach this code. An earlier
revision of this PR claimed those were stored as `+Inf`; that was wrong, and the
corresponding test case has been removed.

## Compatibility

- Objects, arrays, booleans and plain-integer numbers are unchanged.
- A JSON document passed as a JSON string is still unwrapped.
- Values that were previously mangled now round-trip: a JSON string keeps its
  quotes, and a number keeps its original text instead of being rendered through
  `float64`. Note that segcore still reads any literal containing `.` or `e` as
  a double, so `9007199254740993.0` remains a double at query time — this PR
  stops the text from being rewritten, it does not change segcore's number model.

## Validation

- `gofmt` clean.
- `ci-v2/build-ut-cov` green on the first revision (the `httpserver` package
  cannot be built locally against the available native artifacts, which predate
  `SegcoreSetFMIndexCostRatio` / `InitLoonReaderThreadPool`).
- The expected values were cross-checked against the same `gjson` v1.17.3 +
  `sonic` v1.15.2 versions this repo pins; the pre-change behavior reproduces
  the invalid-JSON outputs in the table above.

issue: #52212

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -432,6 +432,12 @@ proxy:
     debug_mode: false # Whether to enable http server debug mode
     port:  # high-level restful api
     acceptTypeAllowInt64: true # high-level restful api, whether http client can deal with int64
+    # high-level restful api, restore the value handling of releases that predate the REST insert validation work.
+    # When true the server keeps the previous lenient behaviour: a missing or null non-nullable field is stored as an
+    # empty value, out-of-range integers wrap instead of being rejected, numbers reach VarChar and JSON fields through
+    # their float64 rendering, and integers too large for the JSON engine become 0. This is a temporary escape hatch
+    # for clients that have not been corrected yet: every one of those behaviours silently changes what is stored.
+    compatibilityMode: false
     enablePprof: true # Whether to enable pprof middleware on the metrics port
     readHeaderTimeout: 5s # HTTP server timeout for reading request headers
     readTimeout: 0s # HTTP server timeout for reading the entire request, including the body. 0 disables this timeout
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -499,6 +499,9 @@ func jsonNumberLiteral(field string, raw string) (json.Number, error) {
 func checkAndSetData(body []byte, collSchema *schemapb.CollectionSchema, partialUpdate bool) ([]map[string]interface{}, map[string][]bool, error) {
 	var reallyDataArray []map[string]interface{}
 	validDataMap := make(map[string][]bool)
+	// Escape hatch for clients that relied on the previous value handling.
+	// Read once per request rather than per field.
+	compatibilityMode := paramtable.Get().HTTPCfg.CompatibilityMode.GetAsBool()
 	dataResult := gjson.GetBytes(body, HTTPRequestData)
 	dataResultArray := dataResult.Array()
 	if len(dataResultArray) == 0 {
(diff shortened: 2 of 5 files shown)


xiaofan-luan:

PR #52214, opened 2026-08-06
enhance: support mixed exact and prefix removals in composite txn

shaoting-huang:
issue: #51351

## What

The composite-update txn engine (#51624) rejects mixing exact and prefix removals in one atomic transaction, which blocked two call sites that #51624 explicitly deferred:

- **rootcoord `dropRole`**: DropRole (exact removals) + DeleteGrant (prefix removals) were two independent catalog calls — a crash in between leaves orphaned grants.
- **querycoord `ReleaseCollection`**: exact collection-key removal + partition prefix removal were two steps — a crash in between leaves a collection released with partitions still persisted.

This PR adds mixed-removal support to the engine and folds both flows into single composite transactions. Three commits, one concern each:

1. `enhance: support mixed exact and prefix removals in composite txn` — new `TxnKV.MultiSaveAndRemoveMixed`: etcd implements it natively (one txn mixing exact `OpDelete` and `WithPrefix` deletes), TiKV materializes prefixes into exact deletes inside the same transaction snapshot; `txn.Commit` lifts the mix ban on the atomic path. Chunked-fallback semantics (commit marker lands last) unchanged.
2. `enhance: drop role and its grants in one composite txn` — role record + user-role mappings + grantee-privileges/grantee-id subtrees in one composite `Update`; role record remains the visibility marker (lands last on fallback). Persisted keys/values are byte-for-byte identical to the legacy two-call end state.
3. `enhance: release collection load info atomically` — collection load-info exact removal + partition prefix removal in one transaction.

## Crash-safety hardening found in review

Adversarial review with fault injection caught an ordering bug in the drop-role fallback path: the grantee-privileges subtree (the only index from which a retry can rediscover grantee-id subtrees) was removed in the first prefix chunk. A crash between chunks made the retry converge "successfully" while permanently orphaning the remaining grantee-id subtrees — and since `crypto.GranteeID` is deterministic, re-creating a same-name role and granting on the same object silently resurrects the stale privileges. Fixed by ordering the index subtree last; regression test `TestCatalog_Update_DropRoleWithGrants_SecondPrefixChunkCrashRetry` injects a second-chunk crash and asserts retry converges to the legacy end state.

## Testing

TDD throughout (every test was written first and observed failing). Coverage follows the #51624 test bar: byte-for-byte encoding equivalence with the legacy paths, op ordering, atomic/fallback crash injection with retry convergence, idempotent retry. Full `internal/metastore/...`, `internal/kv/...`, `internal/querycoordv2/meta/...` and `internal/rootcoord/...` suites pass locally.

## Notes

- `MetaTable.DropGrant` / `IMetaTable.DropGrant` now have no production callers (the drop-role flow no longer uses them); removal is deferred to a follow-up cleanup to keep this PR focused.
- Concurrency precondition documented on `metastore.DropRoleGrants`: apply-time reads are committed without predicates, callers must serialize against concurrent grant writes (rootcoord does via MetaTable's permission lock).

### DIFF
--- a/internal/metastore/kv/querycoord/kv_catalog.go
+++ b/internal/metastore/kv/querycoord/kv_catalog.go
@@ -242,14 +242,11 @@ func (s Catalog) GetResourceGroups(ctx context.Context) ([]*querypb.ResourceGrou
 }
 
 func (s Catalog) ReleaseCollection(ctx context.Context, collection int64) error {
-	// remove collection and obtained partitions
+	// remove collection and obtained partitions in one txn, so a crash cannot
+	// leave partitions behind without their collection load info
 	collectionKey := EncodeCollectionLoadInfoKey(collection)
-	err := s.cli.Remove(ctx, collectionKey)
-	if err != nil {
-		return err
-	}
 	partitionsPrefix := fmt.Sprintf("%s/%d/", PartitionLoadInfoPrefix, collection)
-	return s.cli.RemoveWithPrefix(ctx, partitionsPrefix)
+	return s.cli.MultiSaveAndRemoveMixed(ctx, nil, []string{collectionKey}, []string{partitionsPrefix})
 }
 
 func (s Catalog) ReleasePartition(ctx context.Context, collection int64, partitions ...int64) error {
--- a/internal/metastore/kv/querycoord/kv_catalog_test.go
+++ b/internal/metastore/kv/querycoord/kv_catalog_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/milvus-io/milvus/internal/kv/mocks"
 	. "github.com/milvus-io/milvus/internal/querycoordv2/params"
 	"github.com/milvus-io/milvus/pkg/v3/kv"
+	"github.com/milvus-io/milvus/pkg/v3/kv/predicates"
 	"github.com/milvus-io/milvus/pkg/v3/proto/querypb"
 	"github.com/milvus-io/milvus/pkg/v3/util/etcd"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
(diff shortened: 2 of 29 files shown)


internal/metastore/update_action.go
```diff
@@ -284,3 +299,22 @@ func SaveReplica(r *querypb.Replica) UpdateAction {
...
+// last on the chunked fallback path, so a crash mid-drop leaves the role
+// visible (and the drop retryable) instead of orphaning its grants.
+func DropRole(tenant string, roleName string) UpdateAction {
+	return UpdateAction{Type: ActionDelete, Entry: RoleEntry{Tenant: tenant, Name: roleName}}
+}
+
+// DropRoleGrants returns an UpdateAction that removes every grant of a role.
+// Both DropRole and DropRoleGrants read the current key set at apply time and
+// commit it without predicate checks, so the caller must serialize them
+// against concurrent grant/user-role writes (rootcoord does this via
```
liliu-z, line 315: The parenthetical claim that rootcoord serializes these actions via MetaTable's permission lock is not accurate: `MetaTable.AlterCollection` rewrites the same grantee-privileges and grantee-id keys through `catalog.MigrateGrantCollectionName` while holding only `ddLock`, and both collection-drop paths call `catalog.DeleteGrantByCollectionName` the same way, so neither takes `permissionLock`. Drop-role also broadcasts under `NewExclusivePrivilegeResourceKey` while rename broadcasts under `NewExclusiveDBNameResourceKey`, so the broadcast scheduler does not order them either; the underlying race is pre-existing and this PR narrows it (one lock hold instead of two), but the new doc comment should not promise a guarantee that does not hold.

internal/metastore/kv/querycoord/kv_catalog.go
```diff
@@ -242,14 +242,11 @@ func (s Catalog) GetResourceGroups(ctx context.Context) ([]*querypb.ResourceGrou
...
+	// remove collection and obtained partitions in one txn, so a crash cannot
+	// leave partitions behind without their collection load info
 	collectionKey := EncodeCollectionLoadInfoKey(collection)
-	err := s.cli.Remove(ctx, collectionKey)
-	if err != nil {
-		return err
-	}
 	partitionsPrefix := fmt.Sprintf("%s/%d/", PartitionLoadInfoPrefix, collection)
-	return s.cli.RemoveWithPrefix(ctx, partitionsPrefix)
+	return s.cli.MultiSaveAndRemoveMixed(ctx, nil, []string{collectionKey}, []string{partitionsPrefix})
```
liliu-z, line 249: On TiKV this swaps a server-side range delete for a client-side scan that buffers one `txn.Delete` per partition key inside the transaction: the old path called `RemoveWithPrefix` → `txn.DeleteRange`, while `MultiSaveAndRemoveMixed` iterates the prefix range and deletes key by key, and `runWriteTxnWithRetry` re-runs that whole scan on every attempt. Nothing chunks this path — `maxTiKVTxnOps` only bounds `txn.Commit`, which `ReleaseCollection` bypasses by calling the KV directly — so the buffered mutation set scales with the loaded-partition count (1024 at the default `rootCoord.maxPartitionNum`, more if it is raised).

internal/metastore/kv/txn/commit.go
```diff
@@ -83,12 +83,12 @@ func commitAtomic(ctx context.Context, txn kv.TxnKV, b *Builder) error {
...
 	}
-	// A single etcd txn cannot express exact and prefix deletes together:
 	// MultiSaveAndRemoveWithPrefix deletes EVERY listed key by prefix, so an
 	// exact Remove("coll-1") routed through it would also nuke "coll-10" and
-	// "coll-1/x". Reject the mix rather than silently widen the delete.
+	// "coll-1/x". A mixed set therefore routes through MultiSaveAndRemoveMixed,
+	// the one TxnKV call that keeps both delete shapes in a single txn.
 	if len(removals) > 0 && len(prefixRemovals) > 0 {
-		return merr.WrapErrParameterInvalidMsg("composite update cannot mix exact and prefix removals in one atomic transaction")
+		return txn.MultiSaveAndRemoveMixed(ctx, saves, removals, prefixRemovals)
```
liliu-z, line 91: `ValidateNoSaveUnderRemovedPrefix` only runs on the mixed branch, because only `MultiSaveAndRemoveMixed` calls it. The prefix-only atomic branch routes to `MultiSaveAndRemoveWithPrefix` and the chunked fallback flushes puts and prefix deletes as separate unvalidated calls, so the same builder contents error when they fit `MaxTxnOps` and apply silently when they do not. No current `Update` caller stages a save alongside a prefix removal so this is latent today, but the check belongs in `Commit` rather than in one of three routes.

internal/metastore/kv/querycoord/kv_catalog_test.go
```diff
@@ -339,6 +340,118 @@ func (suite *CatalogTestSuite) TestLoadRelease() {
...
+		suite.NoError(err)
+		keys, _, err := suite.kv.LoadWithPrefix(ctx, EncodePartitionLoadInfoPrefix(collection))
+		suite.NoError(err)
+		suite.Len(keys, len(partitions))
+	}
+
+	// prefix removal fails while exact removal would succeed: the release must
+	// not leave partitions behind without the collection load info
+	suite.saveCollectionWithPartitions(ctx, 1, 100, 101)
+	suite.catalog.cli = &releaseFailKV{MetaKv: suite.kv, removeWithPrefixErr: injected}
```
liliu-z, line 419: `ReleaseCollection` no longer calls `RemoveWithPrefix`, so `releaseFailKV.RemoveWithPrefix` never fires and the error at line 420 is always nil — the `if err != nil` branch is dead and this block only re-asserts what `TestReleaseCollectionRemovesAllKeys` already covers. The second block (`mixedErr: injected`) is the one that genuinely exercises the failure path; either delete the first block or point its injection at `MultiSaveAndRemoveMixed` and assert it fired. Line 432's `suite.catalog.cli = suite.kv` is also dead, being overwritten on the next line.

Issue #52188, opened 2026-08-05
[Bug]: REST v2 large search responses are fully buffered, causing high latency and memory amplification


yanliang567:
## Environment

- **Milvus Version**: Observed on a managed Zilliz Cloud deployment on 2026-08-05; the exact server image/commit is not exposed. The buffering behavior described below is also present on Milvus `master` at `c552dfda0f7b`.
- **Deployment Mode**: Managed deployment
- **MQ**: Unknown / managed
- **SDK**: REST Data Plane V2; comparison clients were PyMilvus `2.6.15rc6` and Go client v2 `2.6.5`
- **OS**: macOS arm64 client

### Option B: Steps

1. Deploy Milvus with the REST Data Plane V2 API enabled.
2. Create and load a collection containing approximately 1 million rows. The test collection used a 3584-dimensional float vector field plus several scalar/VARCHAR output fields.
3. Build a COSINE vector index and load the collection.
4. Send three serial REST requests to `/v2/vectordb/entities/search` with:
   - `nq=1`
   - `limit=300000`
   - COSINE metric
   - six output fields
5. Send three serial REST requests to `/v2/vectordb/entities/hybrid_search` with:
   - two ANN sub-requests, each with `limit=300000`
   - final `limit=300000`
   - RRF reranker with `k=60`
   - the same six output fields
6. Record client wall latency, response size, and the matching Proxy ACCESS log `time_cost`.
7. Run the equivalent requests through gRPC clients using the same vectors, parameters, output fields, and result count.

Example REST search payload construction:

```python
import random

payload = {
    "collectionName": "rest_large_result_repro",
    "data": [[random.random() for _ in range(3584)]],
    "annsField": "vector",
    "limit": 300000,
    "outputFields": [
        "primary_key",
        "field_1",
        "field_2",
        "field_3",
        "field_4",
        "field_5",
    ],
    "searchParams": {"metricType": "COSINE"},
}
```

### Trigger Conditions

- **Frequency**: 6/6 large-result REST requests were slow.
- **First observed after**: Requesting 300,000 result rows with six output fields.
- **Does NOT happen when**: The underlying gRPC `Search`/`HybridSearch` operation does not show comparable server compute latency. The exact REST result-count boundary has not yet been measured.

## Expected Behavior

Large JSON responses are expected to take longer than protobuf responses, but the REST response path should not fully materialize and buffer the entire response before sending it.

For a large result set, the desired behavior is:

- bounded response-generation memory usage;
[...]

PR #52215, opened 2026-08-06
fix: stream large REST search responses

MrPresent-Han:
issue: #52188

## What changed

- replace full V2 Query/Get/Search/Hybrid Search row-map materialization with a validated lazy row source
- stream every V2 row source directly after timeout arbitration through a 64 KiB downstream buffer; the previous row-count fallback is removed
- keep generic non-row JSON encoding in the handler goroutine, where timeout arbitration can still win before the first real response write
- validate and pre-encode Search aggregation results per bucket under the request context, then stream the prepared buckets after arbitration
- propagate renderer/registration failures back to V2 handlers so the Proxy response status, request metrics, and traces record result-conversion failures
- make response preflight context-aware, including nullable/vector/struct/dynamic JSON/value checks
- preserve all-null `TEXT` rows and guard mixed zero-hit recall calculation against `0/0 -> NaN`
- preserve configured HTTP `writeTimeout`; when it is disabled, install the request write deadline and clear it after the request so keep-alive connections are not poisoned
- stop pooling timeout response buffers larger than 1 MiB
- preserve the legacy V1 materialized response behavior

## Root cause

The previous REST V2 path built a complete `[]map[string]interface{}`, encoded the full JSON document, copied it into the timeout middleware buffer, and only then wrote to the client. Large TopK responses therefore incurred multiple response-sized live allocations and could not produce an early first byte.

The timeout recorder now stores only a deferred cooperative stream renderer. If timeout wins before commit, the renderer is discarded and the existing structured 408 is published. If handler completion wins, recorder state is detached under the mutex and the row/bucket stream writes through the real response writer without retaining a full-document timeout buffer.

## Timeout and error semantics

- preflight and handler-side encoding observe the request context at bounded loop intervals
- if the deadline expires before the first real response write, the middleware can still return the structured 408 response
- once success bytes are committed, a timeout stops further production/writes; it cannot replace the partial success response with a 408
- with `proxy.http.writeTimeout=0`, the stream writer sets the connection write deadline to the request deadline and clears it on completion
- with a configured server write timeout, the request path does not overwrite that connection deadline
- timeout/renderer registration races prefer the request context error rather than reporting a false `ErrInvalidSearchResult`
- malformed internal Query/Search results are preflighted before stream commit and reported as `ErrInvalidSearchResult`; the underlying Proxy response status and handler error are also updated for metrics/tracing

## Measured results

For the 300,000-row benchmark fixture:

| Path | Baseline | New | Baseline bytes/op | New bytes/op |
|---|---:|---:|---:|---:|
| Complete validated response | 1.078 s | 0.967 s | 1,140,356,304 | 271,695,912 |
| Prebuilt rows through timeout middleware | 1.958 s | 0.599 s | 1,473,877,744 | 141,357,640 |

- complete-response max RSS: 970,008 -> 341,800 KiB
- prebuilt-row timeout max RSS: 1,064,592 -> 542,764 KiB
- complete-response first write: approximately 1.1107 s -> 27.7 ms
- prebuilt-row first write: approximately 1.9550 s -> 129.7 us

Standalone verification returned 300,000 unique rows in all three runs per route:

- Search: 36,815,636 bytes, 0.423-0.517 s TTFB, 0.848-0.944 s wall
- Hybrid Search: 40,670,066 bytes, 1.747-1.861 s TTFB, 2.156-2.268 s wall

The 10-row benchmark is approximately 6% slower (51.7 us -> 55.0 us), with 15.5 -> 19.4 KiB/op and 137 -> 142 allocs/op.

A few-wide-row benchmark (3 rows, 4,303,499-byte response) currently reports:

- direct: 20,345,184 B/op, 7,343 allocs/op
- timeout: 18,839,064 B/op, 113 allocs/op

`B/op` is cumulative allocation, not peak live memory.

## Validation

Current head `acc9789908`:

- focused renderer-registration, cancellation, nullable-vector, timeout, malformed-result, endpoint, all-null TEXT, and mixed zero-hit recall tests: passed
- full `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/distributed/proxy/httpserver/...`: passed
- timeout/render-related `-race` run with the required tags/gcflags: passed
- `gofmt` and `git diff --check`: passed
- DCO: passed

- `make test-proxy`: passed (`internal/proxy` 80.8% coverage; HTTP server 88.6% coverage)

## Limitations

- the 64 KiB buffer bounds downstream buffering, but Sonic still materializes one complete encoded row in scratch: temporary encoding memory is `O(max encoded row)`, not `O(64 KiB)`; follow-up: #52244
- Search aggregation is prepared per bucket; cancellation is checked between buckets and nested loops, while one bucket marshal remains the largest non-preemptible encoding unit
- generic non-row values remain handler-buffered, and an individual monolithic custom JSON marshal may not observe cancellation until it writes
- actual slow-client and mid-stream TCP-disconnect runs were not performed; injected writer failures verify that production stops on the first observed write error
- HTTP/2 and external reverse-proxy/compression behavior were not measured separately
- this change does not claim OOM prevention or zero small-response regression


internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -297,14 +369,15 @@ func timeoutMiddleware(handler gin.HandlerFunc) gin.HandlerFunc {
...
 			}
 			gCtx.Next()
-			if err := recorder.CommitTo(realWriter); err != nil {
-				mlog.Warn(context.TODO(), "failed to write response body", mlog.Err(err))
-				recorder.Close()
-				bufPool.Put(buffer)
+			commitErr := func() error {
+				defer bufPool.Put(buffer)
+				defer recorder.Close()
+				return recorder.CommitTo(realWriter)
```
buqian-zilliz: **[P1] Keep deferred encoding within the request timeout**

Once the finish arm is selected, CommitTo performs deferred Prepare, Row, and JSON encoding synchronously with no remaining timeout arbitration or context check. In the base path, row construction and encoding happened before the handler signaled completion, so RequestTimeoutMs covered that server work. A backend that finishes just before the deadline can now spend substantially longer serializing a large response and still return 200. Keep deferred rendering under the request deadline, or stop rendering when the request context expires.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,261 @@
...
 
 // Render writes data with custom ContentType.
 func (r jsonRender) Render(w http.ResponseWriter) error {
 	r.WriteContentType(w)
+	if rows, ok := getJSONRowSource(r.Data[HTTPReturnData]); ok {
+		rowCount := rows.Len()
+		if rowCount < 0 {
+			return merr.WrapErrServiceInternalMsg("JSON row source returned negative row count %d", rowCount)
+		}
+		if rowCount <= streamingJSONRowThreshold {
```
buqian-zilliz: **[P2] Use a byte budget for the materialized fallback**

The <=128 branch is bounded only by row count, not encoded size. Default Query/Search limits are 100, while a valid row can contain a high-dimensional vector or large JSON, VARCHAR, or array values, so a tens-of-megabytes response still materializes every row map and the complete encoded document before the first write. Gate this fallback on an estimated byte budget, or stream wide rows regardless of Len().

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,261 @@
...
+			if rowIndex > 0 {
+				if err := bufferedWriter.WriteByte(','); err != nil {
+					return err
+				}
+			}
+			row, err := r.rows.Row(rowIndex)
+			if err != nil {
+				return err
+			}
+			encodedRow, err := json.Marshal(row)
```
buqian-zilliz: **[P2] Avoid an owned JSON allocation for every row**

json.Marshal(row) returns an owned slice on every iteration; Sonic ConfigStd.Marshal copies its pooled output before returning. A 300,000-row response therefore creates roughly 300,000 short-lived byte slices and response-sized allocation and copy traffic in the hot loop before those bytes are written. Reuse one streaming encoder or an encode-into scratch buffer outside the loop.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2664,186 +2695,601 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+			dataIndex, valid, err := rows.fieldDataAccessors[fieldIndex].rowIndex(rowIndex)
+			if err != nil {
+				return err
+			}
+			if !valid {
+				continue
+			}
+			raw := jsonRows[dataIndex]
+			var value map[string]interface{}
+			if err := json.Unmarshal(raw, &value); err != nil {
```
buqian-zilliz: **[P2] Avoid constructing complex rows twice**

This preflight unmarshals every dynamic JSON row into a full map and discards it, while Row unmarshals the same bytes again. validateStructArrays likewise calls accessor.row(), allocating per-element maps and encoded vector values that Row rebuilds later. Large dynamic or array-of-struct responses therefore perform the dominant parsing and allocation work twice before and during streaming, delaying the first byte and increasing GC pressure. Validate syntax and shape directly from raw payloads, or otherwise avoid constructing final row values twice.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,138 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+	}
+	if w.deadlineSet {
+		return nil
+	}
+	w.deadlineSet = true
+	deadline, ok := w.ctx.Deadline()
+	if !ok {
+		return nil
+	}
+	if err := w.controller.SetWriteDeadline(deadline); err != nil && !errors.Is(err, http.ErrNotSupported) {
```
buqian-zilliz, line 450: **[P1] Do not let Request-Timeout extend WriteTimeout**

`prepare` unconditionally replaces the deadline already installed by `http.Server.WriteTimeout` on the underlying connection or stream. Because this context deadline can come from the uncapped client-controlled `Request-Timeout` header, a fast request followed by a slow reader can extend a deployment's configured write timeout from seconds to minutes or hours. This defeats the server-wide slow-client and resource bound for every V2 response committed through this middleware. Make the effective deadline the earlier of the server write timeout and the request deadline instead of always overwriting it.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2624,15 +2739,15 @@ func newFieldDataRowAccessor(fieldData *schemapb.FieldData) (*fieldDataRowAccess
...
+			// The generic validator classifies caller-provided FieldData as an
+			// input error. Query response FieldData is produced inside Milvus,
+			// so a compact-layout violation is deliberately system-blame.
+			return nil, merr.WrapErrServiceInternalErr(err,
+				"query response nullable vector field %s has invalid compact data", fieldData.GetFieldName())
 		}
 		accessor.compactIndices = compactIndices
 		return accessor, nil
 	}
-	if validCount == 0 {
```
buqian-zilliz, line 3630: **[P2] Preserve all-null TEXT result columns**

Removing this zero-valid-row fast path forces every nullable scalar through `fieldDataValueCount`. That switch does not handle `DataType_Text`, even though TEXT query and requery results retain `Type=Text` and store values in `StringData`. For a nullable TEXT column whose `ValidData` contains N false entries and whose physical string payload is empty, the base code emitted N rows with this field set to null; this head now fails construction with `ErrInvalidSearchResult`. Keep the zero-valid-row path while separately rejecting inconsistent payloads, or handle TEXT as a string result type before making the count unconditional.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,161 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+// ResetWriteDeadline clears only a deadline installed by this writer. This is
+// required for keep-alive connections when the server WriteTimeout is disabled;
+// otherwise a completed request's deadline would poison later requests using
+// the same connection.
+func (w *requestDeadlineWriter) ResetWriteDeadline() {
+	if !w.deadlineOwn {
+		return
+	}
+	w.deadlineOwn = false
+	_ = w.controller.SetWriteDeadline(time.Time{})
```
buqian-zilliz: **[P1] Keep the request write deadline through server finalization**

Clearing the deadline here occurs before `net/http` has completed the response. A `ResponseWriter.Write` may only buffer data; after Gin returns, HTTP/1 still runs `finishRequest` to flush the response and chunk terminator, while HTTP/2 flushes the final DATA or trailers with END_STREAM. With the default `proxy.http.writeTimeout=0`, this removes the only write deadline before those operations, so a client that stops reading or withholds HTTP/2 flow-control credit can block finalization indefinitely and retain a server goroutine or stream past the request timeout. Leave the deadline in place until `net/http` performs its own cleanup; HTTP/1 clears it after `finishRequest`, and HTTP/2 stops the timer when the stream closes.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,161 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+	w.deadlineSet = true
+	deadline, ok := w.ctx.Deadline()
+	if !ok {
+		return nil
+	}
+	// net/http already owns the connection deadline when WriteTimeout is
+	// configured. Replacing it with a later, client-controlled request deadline
+	// would weaken the server-wide slow-client bound, and the existing deadline
+	// is not observable through ResponseController. Preserve it unchanged.
+	if paramtable.Get().HTTPCfg.WriteTimeout.GetAsDurationByParse() > 0 {
```
buqian-zilliz: **[P2] Enforce the earlier request deadline with WriteTimeout enabled**

`WriteTimeout > 0` does not imply that the server-owned deadline is earlier than this request's deadline. For example, with `writeTimeout=2m` and the default 30-second request timeout, this branch leaves a blocked HTTP/1 socket write or HTTP/2 flow-control wait governed only by the two-minute server deadline. Context cancellation does not interrupt an in-progress `ResponseWriter.Write`, and after the middleware selects the `finish` case it is no longer watching the request timer, so the per-request timeout can be exceeded. Preserve the earlier of the server and request deadlines instead of skipping request-level write interruption unconditionally.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2664,194 +2845,781 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+			return nil, err
+		}
+		if fieldData == nil {
+			return nil, merr.WrapErrServiceInternalMsg("query response field at index %d is nil", fieldIndex)
+		}
+	}
+	inferredRowCount := rowsNum == 0
+	strictRowCount := inferredRowCount && strictInferredRowCount
+	if inferredRowCount {
+		if strictInferredRowCount && ids != nil {
```
buqian-zilliz, line 3943: **[P1] Treat empty rerank IDs as zero rows**

`strictInferredRowCount` treats every non-nil `Ids` wrapper as authoritative before looking at `FieldsData`. The production all-empty rerank branch returns `Ids: &schemapb.IDs{}` with no ID oneof while preserving the requested nonzero `TopK`; when `outputFields` explicitly includes the primary key or `*`, `searchTask.PostExecute` also appends a zero-row primary-key `FieldData`. The handler therefore does not take its `TopK == 0` shortcut, this branch calls `queryResponseIDCount`, and a valid no-hit Search or Hybrid Search response becomes `ErrInvalidSearchResult`. The base implementation inferred zero rows from the present column and returned an empty data array. Treat an empty IDs wrapper as zero IDs, or fall back to the available zero-row column before rejecting it.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Thanks for flagging this. The path is now bounded at both layers: full-result preflight and row streaming periodically check the request context, while network writes go through `requestDeadlineWriter`. A deadline reached before the first byte returns a structured 408; after the response starts, encoding stops on the context/write error.

internal/distributed/proxy/httpserver/json_render.go
MrPresent-Han: Addressed: the `<=128` materialized fallback has been removed entirely. Every `jsonRowSource` now uses the streaming path, so row count no longer decides whether the full response is buffered.

internal/distributed/proxy/httpserver/json_render.go
MrPresent-Han: Addressed: the row path now reuses one Sonic encoder writing directly to the reusable buffered writer, instead of calling `Marshal` and allocating an owned slice for every row. Scratch space for a single extremely wide row remains O(max encoded row size) and is tracked separately in #52244.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Addressed: dynamic JSON preflight now validates syntax, object shape, and finite numbers with `gjson` without constructing maps. ArrayOfStruct preflight validates row shape and values directly without materializing the final row representation.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Addressed: when `proxy.http.writeTimeout` is configured, `requestDeadlineWriter` no longer overwrites the deadline installed by net/http. A regression test verifies that this path does not call `SetWriteDeadline`.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Addressed: `DataType_Text` is now counted through its `StringData` payload, preserving logical all-null rows while still detecting inconsistent physical data. A V2 Query endpoint test covers the all-null TEXT case.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Thanks for catching this. Addressed in `a4534e4908`: `ResetWriteDeadline` and all renderer/CommitTo early resets were removed, so the request write deadline remains active through net/http finalization. HTTP/1 clears it after `finishRequest`; HTTP/2 stops the timer when the stream closes.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: This limitation is valid, but I am not expanding the implementation further in this PR. `http.ResponseController` exposes a setter but no way to read the deadline already installed by net/http, so unconditionally applying the request deadline can extend an earlier server deadline. The current behavior preserves configured `WriteTimeout` as the hard bound for blocked writes and uses the request context to bound cooperative preflight and encoding. Computing the exact minimum requires explicit server-deadline propagation or net/http support and should be handled as a separate follow-up.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Thanks for catching this. Addressed in `a4534e4908`: a non-nil `schemapb.IDs{}` with an unset oneof is treated as zero IDs. The regression coverage includes both the row utility and a V2 Search endpoint response with `TopK > 0`, `Topks = [0]`, an empty IDs wrapper, and an empty explicitly requested primary-key column.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -1625,33 +1753,42 @@ func scalarArrayToInterfaces(sf *schemapb.ScalarField) []interface{} {
...
+			"struct vector payload length %d is not divisible by row width %d for type %s",
+			dataLength, width, elemType)
 	}
+	return int(int64(dataLength) / width), nil
 }
 
 func vectorFieldToInterfaces(vf *schemapb.VectorField, elemType schemapb.DataType, dim int64) ([]interface{}, error) {
-	if dim <= 0 {
-		return nil, merr.WrapErrParameterInvalidMsg("invalid dim %d", dim)
+	count, err := vectorFieldElemCount(vf, elemType, dim)
```
buqian-zilliz, line 2722: **[P2] Check cancellation during ArrayOfVector conversion**

`vectorFieldToInterfaces` converts a complete ArrayOfVector cell without observing the request context. A valid cell can contain up to 4096 vectors, and the current dimension limits allow the Float16/BFloat16/Binary base64 paths or the Int8 copy path to process hundreds of megabytes after the request deadline. In V2 this helper runs from the deferred `Row` path only after `streamingJSONRender` has flushed the success prefix; cancellation is checked again only after the conversion finishes, so the request goroutine can exceed its timeout and then leave a partial 200 response. Pass the context into this helper and check it inside the per-vector conversion loops.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2664,194 +2845,783 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+	if rows.scores != nil && int64(len(rows.scores)) > i {
+		row[HTTPReturnDistance] = rows.scores[i] // only 8 decimal places
+	}
+	return row, nil
+}
 
+func buildQueryResp(rowsNum int64, needFields []string, fieldDataList []*schemapb.FieldData, ids *schemapb.IDs,
+	scores []float32, enableInt64 bool, collectionSchema *schemapb.CollectionSchema,
+) ([]map[string]interface{}, error) {
+	rows, err := newQueryResponseRowsWithMode(context.Background(), rowsNum, needFields, fieldDataList, ids, scores, enableInt64, collectionSchema, false)
```
buqian-zilliz, line 4716: **[P2] Avoid a second full pass over V1 responses**

`buildQueryResp` is still used by the V1 Query/Get/Search handlers, but this call now runs `validateDynamicJSON`, `validateStructArrays`, and `validateJSONValues` over the complete result before the existing `Row` loop materializes it. Large V1 responses therefore parse dynamic JSON twice and traverse float, vector, and struct payloads in an extra response-sized pass, even though V1 has not committed a deferred stream. This adds CPU and latency to the legacy path that this PR otherwise leaves materialized. Keep the bounds checks needed for safe indexing, but skip or fuse the streaming-only full-value preflight in non-streaming mode.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Updated in 5521239c46. With server WriteTimeout configured, requestDeadlineWriter leaves the existing net/http deadline untouched during normal output, so a later request timeout cannot extend it. A one-shot request-context callback only forces SetWriteDeadline(time.Now()) when the request deadline has expired and Write or Flush is actively blocked. This preserves an earlier server deadline while allowing an earlier request deadline to interrupt the operation. Regression tests cover normal writes plus blocked Write and Flush, and the focused race run passes.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Superseding my earlier response: addressed in 5521239c46. When WriteTimeout is configured, the server-owned deadline remains unchanged, while a one-shot request-context callback interrupts only an active Write or Flush when the shorter request deadline expires. This implements the effective earlier bound without replacing or extending a potentially earlier server deadline. Both blocked Write and blocked Flush are covered by unit tests, and the focused race run passes.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Addressed in 5521239c46. vectorFieldToInterfaces now receives the deferred row context, checks cancellation before conversion, and checks it before every vector for Float, Float16, BFloat16, Binary, and Int8 paths. The struct-array accessor passes its request context through. A table-driven regression test cancels after conversion has started and verifies context.Canceled for every supported element type. The full HTTP server unit suite and the focused race run pass.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Thanks. This is a valid V1-only performance observation, but HTTP V1 is deprecated and this issue and PR are scoped to the V2 large-response path. I am intentionally not adding a V1-specific validation mode or benchmark to this already large patch. buildQueryResp remains a materialized compatibility path with unchanged response values; the additional validation pass on the deprecated interface is accepted rather than expanding the scope further. No code change is planned for this thread.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,172 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+func (w *requestDeadlineWriter) interruptActiveWrite() {
+	w.writeMu.Lock()
+	defer w.writeMu.Unlock()
+	if errors.Is(w.ctx.Err(), context.DeadlineExceeded) && w.writeActive {
+		_ = w.controller.SetWriteDeadline(time.Now())
+	}
+}
+
+func (w *requestDeadlineWriter) Write(data []byte) (int, error) {
+	w.setWriteActive(true)
```
buqian-zilliz: **[P2] Do not mark writes active before preparation succeeds**

`Write` and `FlushError` set `writeActive` before `prepare`. With `WriteTimeout` enabled, the request can expire after `prepare`'s initial context check but before or during `context.AfterFunc` registration; the callback can then set the real write deadline to `time.Now()` even though no underlying I/O started. `prepare` returns `DeadlineExceeded`, so `CommitTo` attempts its structured 408 through the same expired writer and that fallback can fail. Prepare while idle, then mark the operation active and recheck the context immediately before the actual write or flush.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,172 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+func (w *requestDeadlineWriter) setWriteActive(active bool) {
+	w.writeMu.Lock()
+	w.writeActive = active
+	w.writeMu.Unlock()
+}
+
+func (w *requestDeadlineWriter) interruptActiveWrite() {
+	w.writeMu.Lock()
+	defer w.writeMu.Unlock()
+	if errors.Is(w.ctx.Err(), context.DeadlineExceeded) && w.writeActive {
```
buqian-zilliz: **[P2] Keep the request deadline effective through response finalization**

`interruptActiveWrite` only clamps the deadline while `writeActive` is true. When `proxy.http.writeTimeout` is nonzero, a write can return after buffering data; once `CommitTo` returns, the middleware cancels `topCtx`, so the callback sees `context.Canceled` and an inactive writer while net/http performs its final HTTP/1 flush or HTTP/2 completion. A longer server `WriteTimeout` can therefore retain the connection goroutine beyond `RequestTimeoutMs` or the request-header deadline. Preserve the effective earlier deadline through response finalization, for example by propagating the server deadline so the true minimum can be installed safely.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,172 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+	return n, err
+}
+
+func (w *requestDeadlineWriter) FlushError() error {
+	w.setWriteActive(true)
+	defer w.setWriteActive(false)
+	if err := w.prepare(); err != nil {
+		return err
+	}
+	err := w.controller.Flush()
```
buqian-zilliz: **[P2] Preserve underlying flush errors through Gin**

`w.controller` wraps Gin's response writer. In the versions used here, `http.ResponseController.Flush` stops at Gin's void `http.Flusher.Flush` method before following `Unwrap`, and Gin discards the underlying `FlushError`. A server write-timeout or transport flush failure can therefore return nil here while `ctx.Err()` is still nil, causing the renderer to build another 64 KiB batch—or a large single row—before the next `Write` observes the broken connection. Build the controller on the unwrapped writer or add a `FlushError` adapter that preserves the underlying error.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2664,194 +2864,783 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+			case schemapb.DataType_BinaryVector:
+				row[fieldData.GetFieldName()] = fieldData.GetVectors().GetBinaryVector()[dataIdx*(fieldData.GetVectors().GetDim()/8) : (dataIdx+1)*(fieldData.GetVectors().GetDim()/8)]
+			case schemapb.DataType_FloatVector:
+				row[fieldData.GetFieldName()] = fieldData.GetVectors().GetFloatVector().GetData()[dataIdx*fieldData.GetVectors().GetDim() : (dataIdx+1)*fieldData.GetVectors().GetDim()]
+			case schemapb.DataType_Float16Vector:
+				row[fieldData.GetFieldName()] = fieldData.GetVectors().GetFloat16Vector()[dataIdx*(fieldData.GetVectors().GetDim()*2) : (dataIdx+1)*(fieldData.GetVectors().GetDim()*2)]
+			case schemapb.DataType_BFloat16Vector:
+				row[fieldData.GetFieldName()] = fieldData.GetVectors().GetBfloat16Vector()[dataIdx*(fieldData.GetVectors().GetDim()*2) : (dataIdx+1)*(fieldData.GetVectors().GetDim()*2)]
+			case schemapb.DataType_SparseFloatVector:
+				row[fieldData.GetFieldName()] = typeutil.SparseFloatBytesToMap(fieldData.GetVectors().GetSparseFloatVector().Contents[dataIdx])
```
buqian-zilliz: **[P2] Make sparse-vector conversion cancellation-aware**

`validateJSONValues` checks the context while scanning sparse payloads, but `Row` then repeats the full payload walk in `SparseFloatBytesToMap` without context checks and grows an unreserved map. A valid sparse row can contain a very large number of nonzero entries, so a deadline during this second pass still leaves O(nnz) CPU and allocation after the streamed success response has started. Use a context-aware conversion with periodic checks and preallocate `len(contents)/8`, or encode sparse entries directly.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2870,47 +3659,75 @@ func buildSearchAggregationResp(results *schemapb.SearchResultData, enableInt64
...
+		bucket, err := buildAggBucketResp(ctx, pbBucket, enableInt64, collectionSchema)
+		if err != nil {
+			return nil, err
+		}
+		encoded, err := json.Marshal(bucket)
+		if err != nil {
+			return nil, merr.WrapErrServiceInternalErr(err, "encode search aggregation bucket %d", bucketIndex)
 		}
-		output = append(output, gin.H{"buckets": buckets})
+		encodedBuckets = append(encodedBuckets, encoded)
```
buqian-zilliz, line 4838: **[P2] Do not retain every encoded aggregation bucket**

`json.Marshal` returns an owned slice for each bucket, and appending all of them here keeps the complete encoded aggregation payload alive until the deferred stream finishes. The aggregation guard limits entry count rather than encoded bytes and can be disabled, while top hits can carry large string or byte fields, so this path still holds the protobuf result plus O(total response size) encoded JSON and cannot emit a first byte until the last bucket is marshaled. Preflight marshal-sensitive values without retaining every encoding and encode buckets in `WriteJSON`, or enforce an explicit byte budget.

xiaofan-luan:
Reviewed current head `5521239c46`. The main V2 row-streaming direction looks sound, but I don't think this is ready to merge yet:

- **Current CI blocker:** Code-Check fails `depguard` because `timeout_middleware.go` and `json_render_test.go` import the standard-library `errors`; this repository requires `github.com/cockroachdb/errors`.
- **Timeout/write correctness:** the pre-I/O deadline race, request-deadline gap during net/http finalization, and swallowed underlying flush errors are still present:
  - https://github.com/milvus-io/milvus/pull/52215#discussion_r3729047466
  - https://github.com/milvus-io/milvus/pull/52215#discussion_r3729047800
  - https://github.com/milvus-io/milvus/pull/52215#discussion_r3729048143
- **Remaining response-bound issues:** sparse-vector materialization does not observe cancellation, and SearchAggregation still retains every encoded bucket:
  - https://github.com/milvus-io/milvus/pull/52215#discussion_r3729048574
  - https://github.com/milvus-io/milvus/pull/52215#discussion_r3729048928

The single extremely-wide-row Sonic scratch boundary tracked by #52244 is clearly separated and does not need to block this PR. I recommend addressing or explicitly resolving/de-scoping the items above before merge.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Addressed in `d5d64a5efa`. `Write` and `FlushError` now run `prepare()` while the writer is inactive, then lock `writeMu`, recheck the request context, and mark the operation active immediately before the underlying output call. The active flag is cleared as soon as that call returns. If the deadline fires during preparation, the callback observes an inactive writer and does not clamp the real write deadline, preserving the uncommitted structured-408 path. Regression coverage includes this prepare-time race plus blocked `Write` and `Flush`; the focused race run and full HTTP-server unit suite pass.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: This is a valid remaining limitation. I tried propagating an application-owned server deadline and installing `min(serverDeadline, requestDeadline)`, but reverted that approach in `d5d64a5efa`: `ResponseController` cannot read the absolute deadline already installed by net/http, so setting a reconstructed deadline can extend an earlier native server bound. The current implementation deliberately preserves native `WriteTimeout` semantics and uses the request deadline only to interrupt an active `Write`/`Flush`. Therefore, with `WriteTimeout > 0`, a blocked HTTP/1 final flush or HTTP/2 completion after the handler returns is not fully bounded by the request deadline. A complete fix requires connection/stream-level deadline ownership and separate HTTP/1 and HTTP/2 slow-reader verification, which is outside this PR. I will keep this as an explicit follow-up limitation and will not claim finalization is fully covered here.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: Addressed in `d5d64a5efa`. `requestDeadlineWriter` now builds a separate flush controller on Gin's unwrapped underlying writer, while explicitly calling Gin's `WriteHeaderNow()` before flushing so status/header semantics are preserved. This bypasses Gin's void `Flush()` and returns the underlying `FlushError` unchanged. The regression test injects a sentinel flush error and verifies both that it is propagated and that no subsequent row is produced.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Addressed in `d5d64a5efa`. The HTTP response projection now uses a context-aware sparse converter that checks cancellation before allocating the map, preallocates `len(contents)/8`, checks the context every 256 entries, and checks once more after conversion. `Row()` propagates the cancellation error. The regression test cancels during sparse conversion and verifies `context.Canceled` is returned.

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Agreed; this remains intentionally outside this PR. The current aggregation path still retains all encoded bucket slices, so its additional live encoding memory is `O(total encoded aggregation response size)` and first byte waits for the last bucket. The incremental design and acceptance criteria are tracked in #52283: retain protobuf buckets plus lightweight boundaries, preflight without keeping every encoding, then construct/encode/release one bucket at a time. This PR does not claim complete aggregation-response streaming.

MrPresent-Han:
@xiaofan-luan Thanks for the summary. Current head is `d5d64a5efa`.

- The depguard blocker is fixed by replacing the two standard-library `errors` imports with `github.com/cockroachdb/errors`. Targeted `golangci-lint` reports 0 issues; the new remote CI run is currently pending.
- The pre-I/O deadline race is fixed: preparation happens while inactive, followed by a locked context recheck immediately before marking the underlying Write/Flush active.
- The finalization observation is valid but is intentionally not claimed as fixed when `WriteTimeout > 0`. I reverted the application-owned server-deadline/minimum approach because it could extend an earlier opaque net/http deadline. The remaining HTTP/1 final-flush / HTTP/2 completion gap requires connection/stream-level deadline ownership and stays follow-up work.
- Gin's swallowed flush error is fixed by flushing through the unwrapped writer while preserving `WriteHeaderNow()`; the underlying error is returned and row production stops.
- Sparse conversion now checks cancellation before allocation and periodically during conversion, with NNZ-sized map preallocation.
- Aggregation bucket retention remains out of scope and is tracked by #52283; this PR does not claim complete aggregation-response streaming.

Verification on this head: full HTTP-server unit suite, focused race tests, Proxy HTTP timeout configuration tests, `go vet`, `gofmt`, `git diff --check`, and targeted `golangci-lint` all passed.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,225 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+	if w.serverWriteTimeout > 0 {
+		context.AfterFunc(w.ctx, func() {
+			w.interruptActiveWrite()
+		})
+		return w.ctx.Err()
+	}
+	// Leave the request deadline in place until net/http finishes the response.
+	// The server performs its final buffered flush after the handler returns and
+	// then clears the HTTP/1 connection deadline or closes the HTTP/2 stream.
+	if err := w.controller.SetWriteDeadline(deadline); err != nil && !errors.Is(err, http.ErrNotSupported) {
```
buqian-zilliz: **[P2] Do not send the timeout fallback through an expired deadline**

When `proxy.http.writeTimeout` is zero, `prepare` installs `topCtx`'s deadline on the real response writer before `beginWrite` performs its second context check. If the deadline expires in that gap, the first write is correctly skipped and `CommitTo` selects the unwritten 408 fallback, but `writeRequestTimeout` sends that fallback through an HTTP/1 writer whose deadline is already expired or an HTTP/2 stream that has already timed out, and it ignores the write result. The client can therefore receive no JSON 408 even though no success byte was written. Arm the request deadline together with the first underlying output operation, or replace the expired deadline with a short bounded fallback deadline before writing the timeout response.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2664,194 +2885,799 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+
+func (rows *queryResponseRows) validateJSONValues() error {
+	for fieldIndex, fieldData := range rows.fieldDataList {
+		if err := checkResponseContext(rows.ctx, int64(fieldIndex)); err != nil {
+			return err
+		}
+		if fieldData.GetType() == schemapb.DataType_ArrayOfStruct {
+			continue
+		}
+		for rowIndex := int64(0); rowIndex < rows.rowsNum; rowIndex++ {
```
buqian-zilliz, line 4383: **[P2] Skip row scans for field types with nothing to validate**

`validateJSONValues` enters this row loop for every field except `ArrayOfStruct`, although `validateQueryResponseFieldValue` immediately returns nil unless the type is Float, Double, FloatVector, SparseFloatVector, or Array. Bool, integer, string/Text, JSON, Geometry, and byte-vector columns therefore each add `rowsNum` context/modulo checks, nullable-index lookups, and validity branches before the stream is registered. At the intended large-result scale, wide responses can perform millions of no-op iterations before their first byte. Filter for the handful of types that can contain non-finite JSON numbers before entering the row loop.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -170,25 +197,225 @@ func (w *timeoutResponseRecorder) Pusher() http.Pusher {
...
+	if !ok {
+		return nil
+	}
+	// net/http already owns the connection or stream deadline when the startup
+	// WriteTimeout snapshot is positive. Do not replace that deadline because it
+	// may be earlier than the request deadline. Instead, if the request deadline
+	// wins while a write is blocked, force the active write to expire immediately.
+	// Runtime config changes must not alter this ownership for an existing server.
+	if w.serverWriteTimeout > 0 {
+		context.AfterFunc(w.ctx, func() {
```
buqian-zilliz: **[P3] Stop the write callback after successful rendering**

`context.AfterFunc` starts its callback in a new goroutine when the context is canceled, but the returned stop function is discarded. For every successful response with a positive `proxy.http.writeTimeout` that reaches `prepare`, the middleware's deferred `cancel()` runs after `CommitTo`; `writeActive` is already false, so the callback only acquires `writeMu` and exits. This adds an avoidable goroutine and scheduler work per response on that configured path. Retain the stop function and invoke it when commit or rendering finishes, while leaving the callback armed during active I/O.

internal/distributed/proxy/httpserver/timeout_middleware.go
buqian-zilliz: This issue still exists. The moved deadline setup removes the old prepare/beginWrite gap, but the first prefix write can still be accepted only into Gin/net/http buffers: Gin marks the response as written before forwarding it, while net/http may not flush those bytes to the socket. If the request deadline fires in that window, runOutput returns DeadlineExceeded and the expired transport prevents the later flush, yet the Written() check still suppresses the structured 408. The handoff must distinguish buffered acceptance from a successfully flushed first byte.

MrPresent-Han: The buffering window is real, but it is after the response has crossed the logical commit boundary of Gin/net/http. Gin `WriteHeaderNow` marks the writer as written and calls the underlying `WriteHeader` before forwarding the body; after that, net/http will not accept a different status even if the header or body is still buffered. net/http exposes neither rollback nor a reliable signal that the first byte reached the peer, so treating a failed or expired first flush as unwritten and sending a 408 could append a timeout payload to a partially committed 200.

The original issue was narrower: the request deadline could expire the transport before any underlying output operation began, making an otherwise unwritten 408 impossible. Current head keeps the transport untouched until the first output passes the final context check, and the renderer immediately follows the prefix write with an explicit `FlushError`. A deadline racing with that write/flush is therefore a post-commit transport failure, not an unwritten response that can safely be replaced.

Achieving a wire-visible atomic handoff would require protocol-specific connection or stream ownership for HTTP/1 and HTTP/2, outside this PR. I propose keeping `Written()` as the safe irreversibility boundary and documenting this as a transport-level limitation rather than reopening the original P2.

internal/distributed/proxy/httpserver/timeout_middleware.go
MrPresent-Han: fixed

internal/distributed/proxy/httpserver/utils.go
MrPresent-Han: Addressed in current head `c3fd259dfb`. `validateJSONValues` now filters at the field level, so only Float, Double, FloatVector, SparseFloatVector, and Array enter the per-row validation loop; all other field types `continue` before any row scan. Added `TestQueryResponseRowsSkipsJSONValueValidationForNoopFields`, which uses a VarChar field and verifies that row-level context checks are not consumed.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,248 @@
...
+			continue
+		}
+
+		// Commit the response prefix before producing the streamed value. The
+		// 64 KiB buffer bounds downstream write buffering; an individual Sonic
+		// row encode may still require scratch space proportional to that row.
+		if err := bufferedWriter.Flush(); err != nil {
+			return err
+		}
+		if err := deadlineWriter.FlushError(); err != nil {
```
liliu-z: `streamingJSONRender.Render` unconditionally forces a transport-level flush before emitting the `data` array, so every V2 Query/Get/Search/HybridSearch response now commits its headers mid-handler regardless of size. Two consequences that the PR description does not call out: small responses that previously fit inside net/http's 2 KiB pre-chunking buffer and shipped with a `Content-Length` are now always `Transfer-Encoding: chunked`, and because `realWriter.Written()` becomes true at that point the middleware permanently loses its ability to substitute the structured 408 or an `ErrInvalidSearchResult` body for the rest of the request — even for a two-row result. Gating the early flush on a size or row-count threshold would keep the early-first-byte win for large results without changing the wire format of every small one.

MrPresent-Han: Addressed in 69bd94bfa5. I removed the unconditional pre-data flush, so small responses remain uncommitted while encoding, preserving HTTP/1 Content-Length and leaving the existing pre-commit timeout/error path in control. Large responses still begin streaming naturally once the bounded 64 KiB buffer fills. Added coverage for small-response Content-Length and a structured 408 before commit.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,241 @@
...
+	return bufferedWriter.Flush()
+}
+
+func (source *jsonRowStreamSource) WriteJSON(ctx context.Context, w io.Writer) error {
+	if _, err := w.Write([]byte{'['}); err != nil {
+		return err
+	}
+	rowEncoder := json.NewEncoder(w)
+	for rowIndex := int64(0); rowIndex < source.rowCount; rowIndex++ {
+		if err := ctx.Err(); err != nil {
```
liliu-z: The response body is now produced under the request-scoped timeout context, so `proxy.http.requestTimeoutMs` (default 30000ms) bounds client download time, not just server processing. Before this PR `CommitTo` wrote the fully-buffered body with a single `realWriter.Write(w.body.Bytes())` that never consulted the request context; now `jsonRowStreamSource.WriteJSON` aborts between rows on `ctx.Err()`, and `endWrite` additionally pins the connection write deadline to the request deadline when `proxy.http.writeTimeout` is 0 (the shipped default). A client that is slow enough to take longer than 30s to drain a multi-MB search response therefore receives a truncated, unparseable JSON body after a 200 status line. The failure is not silent — `timeout_middleware.go:589` logs a Warn — but `ProxyFunctionCall` was already incremented with the success label inside the handler goroutine (`handler_v2.go:435-444`) before the renderer ran, so success-rate metrics stay clean. The same ordering also lets the middleware emit a structured 408 for a request the metrics already counted as a success when the deadline fires before the first byte reaches the wire. Consider detaching/extending the deadline once the stream commits, and reflecting post-commit stream failures in the request metrics.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,241 @@
...
+		}
+		encodedKey, err := json.Marshal(key)
+		if err != nil {
+			return nil, err
+		}
+		field := encodedJSONField{key: encodedKey}
+		if key == HTTPReturnData {
+			field.stream = streamSource
+		} else {
+			field.value, err = json.Marshal(data[key])
```
liliu-z, line 195: `newStreamingJSONRender` pre-encodes every non-data field with `json.Marshal`, so any non-finite scalar in the metadata map aborts the entire response as `ErrInvalidSearchResult`. `recallCal` returns NaN when a query in a multi-query batch has zero hits (0/0), and this PR contains no guard for it — internal/proxy/util.go is not in the diff — so the PR description's "guard mixed zero-hit recall calculation against 0/0 -> NaN" is not backed by the change; the new test at handler_v2_stream_test.go:454 in fact locks in NaN recalls producing ErrInvalidSearchResult. This is not a regression (sonic's ConfigStd rejected NaN in the old whole-document encode too, yielding an empty 200 body), but it is an inconsistency in this PR's own design: row values get an explicit non-finite preflight via `validateJSONValues`, while the metadata scalars the PR newly pre-encodes get none. Sanitize NaN/Inf recall before encoding, or fix the zero-hit division in recallCal.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1436,30 +1450,35 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
...
-				HTTPReturnMessage: merr.ErrInvalidSearchResult.Error() + ", error: " + err.Error(),
-			})
+		outputData, buildErr := newQueryResponseRowsWithContext(ctx, int64(0), queryResp.OutputFields, queryResp.FieldsData, nil, nil, allowJS, collSchema)
+		if buildErr != nil {
+			err = buildErr
+			if isRequestContextError(buildErr) {
+				return resp, err
+			}
+			queryResp.Status = merr.Status(merr.ErrInvalidSearchResult)
+			mlog.Warn(ctx, "high level restful api, fail to deal with query result", mlog.Any("response", resp), mlog.Err(buildErr))
```
liliu-z: `mlog.Any("response", resp)` serializes the entire QueryResults/SearchResultData proto — every vector, PK and dynamic field of a full topK result — into a Warn log. The line itself predates this PR, but the PR materially widens what reaches it: the new preflight in `newQueryResponseRowsWithContext` rejects non-finite floats, per-field row-count and ID-count mismatches, invalid or over-nested dynamic JSON, and vector payloads whose length is not a multiple of the row width, none of which previously produced this log. The PR is also internally inconsistent here — the render-failure warnings it adds alongside (e.g. handler_v2.go:1480) deliberately log only the error. Log status code, error and result counts instead of the whole proto; the same applies at handler_v2.go:1533, 2065, 2103 and 2259.

internal/distributed/proxy/httpserver/json_render.go
MrPresent-Han: Addressed in 7266ca86af. `recallCal` now returns 0 when `results` is empty, and the mixed-query test covers `Topks=[2,0]` with recalls `[0.5,0]`. Backported to #52323 in 61ac6c6979.

internal/distributed/proxy/httpserver/json_render.go
MrPresent-Han: Addressed in 7266ca86af. `requestTimeout` now applies only until the first real response output; after commit, rendering uses the original request context, so `WriteTimeout` and client cancellation govern the remaining stream. I kept request success metrics unchanged in this scoped fix; post-commit stream-write observability can be handled separately. Backported to #52323 in 61ac6c6979.

internal/distributed/proxy/httpserver/handler_v2.go
MrPresent-Han: Addressed in 7266ca86af. All five warnings now log status/error and result counts only; the full Query/Search proto is no longer serialized or written to logs. Backported to #52323 in 61ac6c6979.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,268 @@
...
+	WriteJSON(context.Context, io.Writer) error
+}
+
+type streamIdleProgressWriter struct {
+	writer     io.Writer
+	controller *streamIdleController
+}
+
+func (w *streamIdleProgressWriter) Write(data []byte) (int, error) {
+	n, err := w.writer.Write(data)
```
buqian-zilliz: **[P1] Arm the idle timeout before the first stream write**

`recordProgress` is the only place that starts the idle timer, but it runs only after the wrapped `Write` has completed. The first 64 KiB flush can therefore block on HTTP/1 backpressure or HTTP/2 flow control before `streamIdleTimeout` exists (the default HTTP/2 stream window is only 65,535 bytes). A response that fits in one buffer gets no useful enforcement window: the timer starts after its sole write and is immediately stopped when `Render` returns. With the default `writeTimeout=0`, a client can hold the write until the much longer, client-overridable request timeout. Arm the idle timer when streaming begins, then reset it after successful progress.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,268 @@
...
+		row, err := source.rows.Row(rowIndex)
+		if err != nil {
+			if ctxErr := responseContextError(ctx); ctxErr != nil {
+				return ctxErr
+			}
+			return err
+		}
+		// Reuse one streaming encoder. Sonic still materializes one encoded row
+		// in scratch, but avoids an additional owned []byte and copy per row.
+		if err := rowEncoder.Encode(row); err != nil {
```
buqian-zilliz, line 288: **[P2] Make wide-row encoding cancellation-aware**

`rowEncoder.Encode` is still an all-or-nothing CPU step: Sonic first encodes the complete row into a scratch buffer and only then calls the writer, so request cancellation and the stream-idle controller cannot be observed while the row is serialized. A valid row can be very wide (the defaults allow up to 256 fields and 2 MiB per TEXT value), and Sonic deliberately does not pool encoded buffers larger than 1 MiB. If a deadline expires while a later wide row is being encoded, the goroutine continues doing row-sized CPU and allocation work after the 200 prefix may already have been flushed, then returns a partial response. Encode row fields incrementally with periodic context checks, or enforce a bounded per-row encoded size before streaming.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,269 @@
...
+	WriteJSON(context.Context, io.Writer) error
+}
+
+type streamIdleProgressWriter struct {
+	writer     io.Writer
+	controller *streamIdleController
+}
+
+func (w *streamIdleProgressWriter) Write(data []byte) (int, error) {
+	n, err := w.writer.Write(data)
```
buqian-zilliz: **[P2] Bound writes before treating them as stream progress**

Bound the writes used to measure stream progress. Go's `bufio.Writer` forwards a write larger than its buffer directly when the buffer is empty, and Sonic's stream encoder submits each fully encoded row in one `Write`. After the first 64 KiB flush, a multi-megabyte row or aggregation bucket can therefore reach this wrapper as one oversized call. The idle controller is rearmed only after that call returns, so a client that is continuously draining HTTP/2 frames or HTTP/1 bytes can still be canceled once the single `Write` lasts longer than `streamIdleTimeout`. Split outgoing data into bounded chunks, or encode rows incrementally, and rearm after each completed chunk.

internal/distributed/proxy/httpserver/json_render.go
buqian-zilliz: This issue still exists: each complete row is still passed to one `rowEncoder.Encode(row)` call, so Sonic can finish a potentially multi-megabyte scratch encoding before the context-aware writer is touched. No incremental field encoding or per-row encoded-size bound was added.

internal/distributed/proxy/httpserver/json_render.go
liliu-z: The idle timer only re-arms after an underlying write call returns, so a single row larger than the 64 KiB buffer counts as zero progress for the whole duration of its write. sonic's StreamEncoder encodes a row into one buffer and hands it to `Write` in one call, and `bufio.Writer` bypasses its buffer for writes larger than the remaining space, so a multi-MB row becomes one underlying write. A client draining that row steadily but slowly still trips `expire()` -> `cancel(errStreamIdleTimeout)` -> `interruptActiveWrite()` -> `SetWriteDeadline(time.Now())`, and the client gets a truncated 200. This only bites deployments that opt in (`proxy.http.streamIdleTimeout` defaults to `0s`, which disables the controller), but for those it turns healthy-slow into a silent truncation. Split the row output into bounded chunks and arm between chunks so progress is observed incrementally.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2664,194 +2885,805 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+			}
+		}
+	}
+	return nil
+}
+
+// Sonic's encoder uses a 4096-entry stack. Dynamic JSON's top-level object is
+// flattened into the response row, so 4093 nested child containers is the last
+// depth that fits once the row map and encoder root state are included.
+const maxDynamicJSONValueNestingDepth = 4093
```
liliu-z: `maxDynamicJSONValueNestingDepth = 4093` is roughly 2x too permissive, so the preflight that is supposed to guarantee the streaming renderer can never fail mid-body has a hole. sonic's 4096-entry encoder stack is consumed at about two slots per level for `map[string]interface{}` values, so encoding fails at 2048 nested containers, not 4093. A dynamic `$meta` value between those two bounds passes `validDynamicJSONValue`, then fails inside `rowEncoder.Encode(row)` after the renderer has already been committed — which is exactly the mid-stream failure the preflight exists to prevent. Reachable via an SDK insert of a deeply nested dynamic field followed by a REST query. Lower the constant to the depth sonic can actually encode.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -297,30 +666,25 @@ func timeoutMiddleware(handler gin.HandlerFunc) gin.HandlerFunc {
...
+				defer bufPool.Put(buffer)
+				defer recorder.Close()
+				return recorder.CommitTo(requestCtx, realWriter)
+			}()
+			if commitErr != nil {
+				if errors.Is(commitErr, context.DeadlineExceeded) && !realWriter.Written() {
+					writeRequestTimeout(gCtx, realWriter)
+					return
+				}
+				mlog.Warn(gCtx.Request.Context(), "failed to write response body", mlog.Err(commitErr))
```
liliu-z, line 797: The PR adds handler-side propagation of renderer failures to `ErrInvalidSearchResult`, but with the deferred renderer the row materialization and encoding now run after the handler has returned, inside `CommitTo`. A failure there (e.g. a `Row()` error or an encode failure that slipped past preflight) is only logged; `realWriter.WriteHeader(200)` has already been called, so the client gets a bodiless 200 and cannot distinguish it from an empty result set. Worth emitting an error body when `!realWriter.Written()`, and gating the `ErrInvalidSearchResult` label on renderer/encoding errors specifically so a plain connection failure isn't relabelled as an invalid search result.

internal/distributed/proxy/httpserver/json_render.go
```diff
@@ -17,26 +17,286 @@
...
+	if err := responseContextError(r.ctx); err != nil {
+		return err
+	}
+
+	deadlineWriter := newRequestDeadlineWriter(r.ctx, w)
+	defer deadlineWriter.stopDeadlineInterrupt()
+	outputWriter := io.Writer(deadlineWriter)
+	if controller := streamIdleControllerFromContext(r.ctx); controller != nil {
+		controller.arm()
+		defer controller.stop()
```
buqian-zilliz, line 216: **[P2] Keep the idle bound through response finalization**

Stopping the controller here drops the new idle bound before net/http has finalized the response. With the default `writeTimeout=0`, a small final write can succeed after only entering net/http's internal buffers; the actual HTTP/1 socket flush happens after the Gin handler returns, while `requestDeadlineWriter` has left only the potentially much longer request deadline on the transport. A slow client can therefore hold the connection well past `streamIdleTimeout` after the last application write. Keep a rolling idle deadline effective through transport finalization, at least when no server `WriteTimeout` owns the deadline.

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -297,30 +681,32 @@ func timeoutMiddleware(handler gin.HandlerFunc) gin.HandlerFunc {
...
+						writeRequestTimeout(gCtx, realWriter)
+						return
+					}
+					var renderErr *deferredRenderError
+					if errors.As(commitErr, &renderErr) && !errors.Is(commitErr, context.Canceled) {
+						returnInvalidSearchResult(gCtx, renderErr.cause)
+						return
+					}
+				}
+				mlog.Warn(gCtx.Request.Context(), "failed to write response body", mlog.Err(commitErr))
```
buqian-zilliz, line 797: **[P2] Record post-commit stream failures in the request outcome**

After a deferred stream has written its first bytes, any idle timeout, request deadline, or transport error reaches this branch, but it only emits a warning. The successful handler code was already propagated as `HTTPReturnCode=0`, so the structured REST access log reports `Successful` and there is no delivery-failure counter even though the client receives truncated JSON. Mark the outer Gin context and error type, and record a stream-delivery failure before returning; the already-committed HTTP status does not need to be rewritten.

internal/distributed/proxy/httpserver/json_render.go
buqian-zilliz: This issue still exists: the new 64 KiB chunking only applies after Sonic has fully encoded a row and invoked Write; `rowEncoder.Encode(row)` remains an all-or-nothing CPU and allocation step that cannot observe request or stream-idle cancellation.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -3666,272 +3886,857 @@ func (accessor *fieldDataRowAccessor) rowIndex(rowIdx int64) (int64, bool, error
...
+		return !math.IsNaN(value.Num) && !math.IsInf(value.Num, 0)
+	}
+	if value.Type != gjson.JSON {
+		return true
+	}
+	if nestingDepth > maxDynamicJSONValueNestingDepth {
+		return false
+	}
+	valid := true
+	value.ForEach(func(_, child gjson.Result) bool {
```
buqian-zilliz: **[P2] Avoid recursive GJSON rescans during preflight**

`validDynamicJSONValue` recursively calls `ForEach` on each nested `gjson.Result`. `ForEach` reparses the child's complete raw subtree to locate its extent, so a deeply nested dynamic document with a wide leaf is scanned again for every ancestor—the same O(depth * bytes) pattern that `checkEngineCompatible` in this file explicitly avoids. This runs for every dynamic row before the deferred renderer is registered and has no context check within one document, so a valid nested value can consume substantial CPU past the request deadline before any byte is streamed. Use a single-pass token/depth scanner and check the request context during that walk.

internal/proxy/accesslog/info/info.go
```diff
@@ -29,34 +29,36 @@ type getMetricFunc func(i AccessInfo) string
...
+	"$method_name":              getMethodName,
+	"$method_status":            getMethodStatus,
+	"$trace_id":                 getTraceID,
+	"$user_addr":                getAddr,
+	"$user_name":                getUserName,
+	"$response_size":            getResponseSize,
+	"$error_code":               getErrorCode,
+	"$error_msg":                getErrorMsg,
+	"$error_type":               getErrorType,
+	"$stream_termination":       getStreamTermination,
```
buqian-zilliz, line 41: **[P2] Avoid prefix-overlapping access-log placeholders**

`$stream_termination` is a prefix of `$stream_termination_cause`, while `Formatter.build` ranges over `MetricFuncMap` in unspecified map order and splits the remaining format text on each key. If the shorter key is processed first, it also consumes the prefix of the cause placeholder; the longer key is never registered, and the default query/search log emits the termination value followed by the literal `_cause` instead of the actual cause. The result therefore varies between formatter builds or processes. Tokenize placeholders atomically, process keys longest-first, or use non-overlapping names.

internal/distributed/proxy/httpserver/json_render.go
buqian-zilliz: This issue still exists: `rowEncoder.Encode(row)` still serializes one complete row before any bytes reach the cancellation-aware progress writer, so a wide row can continue consuming CPU and allocation after the request or stream-idle deadline.

buqian-zilliz: This issue still exists: `rowEncoder.Encode(row)` still serializes one complete row before any bytes reach the cancellation-aware progress writer, so a wide row can continue consuming CPU and allocation after the request or stream-idle deadline.

PR #52217, opened 2026-08-06
fix: materialize geometry WKT default as WKB for absent columns (#51649)

MrPresent-Han:
### What

\`GenerateEmptyArrayFromSchema\` had no branch for Geometry defaults: DDL persists a geometry default as WKT (\`ValueField.StringData\`) while storage carries WKB, so materializing a wholly-absent geometry column during compaction (e.g. a column added by \`add_field\` that old segments never wrote) fell into the default-value switch's error branch and failed the compaction. Add the Geometry case: convert the WKT default to WKB and back the generated array with it.

The sibling path where the geometry column physically exists but a cell is NULL (\`appendValueAt\`) is a separate pre-existing issue tracked by #52105.

Split from #51848.

issue: #51649
related: #52105

### Tests

- New \`TestGenerateEmptyArrayGeometryDefault\`: asserts valid bits and exact WKB bytes for every generated row; green locally under \`-tags dynamic,test\`.

### DIFF
--- a/internal/storage/arrow_util.go
+++ b/internal/storage/arrow_util.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/storagev2/packed"
+	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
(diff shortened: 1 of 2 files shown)


MrPresent-Han:
Superseded by #52204, which covers this GenerateEmptyArrayFromSchema branch and additionally fixes the sibling appendValueAt path and the Parquet reader normalization (the full #52105 chain), from the original author of the appendValueAt default handling.

PR #52218, opened 2026-08-06
fix: [cp3.0] sync delegator snapshot for delete forwarding

sijie-ni-0214:
Cherry-pick from master
pr: #52167

issue: #52166

## What this PR does

This PR backports synchronous publication of delegator distribution snapshots after sealed segment add, remove, and offline updates.

It closes the visibility gap where a newly loaded sealed segment had completed delete replay and entered the live distribution map, but a subsequent streaming delete could still pin the previous snapshot and skip that segment before the asynchronous snapshot worker published the update.

The change also removes the background snapshot loop and test-only flush path, and serializes distribution shutdown with late sealed candidate additions so Bloom filter resource accounting is refunded correctly.

## Test Plan

- [x] `git diff --check upstream/3.0...HEAD`
- [x] `make` on macOS with LLVM 19
- [x] `make static-check`
- [x] `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l" ./internal/querynodev2/delegator -run 'Test(AddDistributionsPublishesSnapshotSynchronously|Distribution_AddDistributionsAfterClose|Distribution_CloseRefundsExistingCandidate)$|TestStreamingForward/(TestBFStreamingForward|TestDirectStreamingForward)$|TestDistributionSuite/(TestRemoveDistribution|TestMarkOfflineSegments)$' -timeout 300s`

### DIFF
--- a/internal/querynodev2/delegator/delegator.go
+++ b/internal/querynodev2/delegator/delegator.go
@@ -1432,11 +1432,7 @@ func (sd *shardDelegator) Close() {
 		sd.growingSourceProvider.Deactivate()
 	}
 
-	// Stop background snapshot loop before refunding candidates
 	sd.distribution.Close()
-
-	// Refund all sealed segment candidates in distribution
-	sd.distribution.RefundAllCandidates()
 
 	// clean idf oracle
 	if idfOracle := sd.getIDFOracle(); idfOracle != nil {
--- a/internal/querynodev2/delegator/delegator_data_test.go
+++ b/internal/querynodev2/delegator/delegator_data_test.go
@@ -550,7 +550,6 @@ func (s *DelegatorDataSuite) TestProcessDelete() {
 		},
 	}, 10)
 
-	s.delegator.distribution.Flush()
 	s.False(s.delegator.distribution.Serviceable())
 
 	worker1.EXPECT().LoadSegments(mock.Anything, mock.AnythingOfType("*querypb.LoadSegmentsRequest")).
@@ -573,7 +572,6 @@ func (s *DelegatorDataSuite) TestProcessDelete() {
 		Version: time.Now().UnixNano(),
 	})
 	s.Require().NoError(err)
-	s.delegator.distribution.Flush()
 	s.True(s.delegator.distribution.Serviceable())
 	// Test normal errors with retry and fail
 	worker1.ExpectedCalls = nil
@@ -586,7 +584,6 @@ func (s *DelegatorDataSuite) TestProcessDelete() {
 			RowCount:    1,
 		},
 	}, 10)
-	s.delegator.distribution.Flush()
 	s.False(s.delegator.distribution.Serviceable(), "should retry and failed")
 
 	// refresh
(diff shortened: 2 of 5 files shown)


czs007:

This backport of #52167 introduces no new issues beyond the two findings already raised on the upstream PR, both of which remain present here by design.

**Verified:**
- `internal/querynodev2/delegator/distribution.go:359` — `AddDistributions` and `Close` serialize through `d.mut`, and whichever loses the race refunds its own candidate exactly once (asserted by `TestDistribution_AddDistributionsAfterClose` and `TestDistribution_CloseRefundsExistingCandidate`).
- `internal/querynodev2/delegator/distribution.go:576` — `RemoveDistributions` still returns the cleared signal for the snapshot it expires, and the caller waits on it only after `d.mut` is released (`delegator_data.go:1260-1262`), so there is no hold-and-wait.
- `internal/querynodev2/delegator/distribution.go:781` — after `Close` drains candidates, entries left with a nil `Candidate` are explicitly skipped by the only downstream consumer, `BatchGetFromSegments` (`distribution.go:708`, `732`).
- `internal/querynodev2/delegator/snapshot.go:134-139` — the new synchronous `genSnapshot`/`updateServiceable` acquire no additional locks, and expired-snapshot cleanup runs on a background goroutine without holding `d.mut`, so no new lock ordering is introduced.
- `internal/querynodev2/delegator/distribution_test.go:398` — `TestAddDistributionsPublishesSnapshotSynchronously` pins a snapshot immediately after `AddDistributions` and asserts segment visibility, which would fail under the old asynchronous loop, so the test genuinely covers the fixed semantics.
- Repo-wide search confirms `RefundAllCandidates`, `notifySnapshotUpdate`, and `snapshotLoop` have no remaining references; `distribution.Close()` is called only from `delegator.go:1435`.

## Carried over from #52167

- **Confirmed on this branch** (low) — `internal/querynodev2/delegator/distribution.go:398`: every accepted `AddDistributions` call runs `genSnapshot()` (a full scan of all sealed/growing entries) and `updateServiceable()` synchronously while holding `d.mut`, and the load path enters this while holding `deleteMut.RLock()` (`delegator_data.go:1124-1179`); removing the capacity-one notifier also eliminates snapshot-update coalescing. When a shard has a large sealed-segment population and a burst of segment loads completes while streaming deletes are pending → `ProcessDelete`'s acquisition of `deleteMut.Lock()` and tSafe advancement are delayed, causing latency spikes in delete visibility on that channel. This is the intentional trade-off that closes the snapshot-visibility gap and was settled upstream; if it warrants follow-up, consider incremental/copy-on-write snapshot construction, or preparing the snapshot outside the barrier and doing only final validation and pointer publication inside it. *(raised by tinswzy, chyezh)*
- **Confirmed on this branch** (nitpick) — `internal/querynodev2/delegator/distribution.go:566`: `RemoveDistributions` executes `mlog.Info` while holding the `d.mut` write lock (held until line 577), and `MarkOfflineSegments` follows the same new pattern (`distribution.go:445-447`, lock held until 451). If the logging sink stalls during a balance/release (slow disk, full stderr pipe) → all `PinReadableSegments`/`PinOnlineSegments` callers (search, query, delete forwarding) on the shard block for the duration of the log I/O. Settled upstream as a design trade-off; if revisited, capture the log fields under the lock, publish the snapshot, unlock, and emit the log afterward. *(raised by tinswzy, chyezh)*

czs007 (approved):
/lgtm

PR #52219, opened 2026-08-06
fix: run MinHash schema validation on all DDL paths and reject TEXT input (#51649)

MrPresent-Han:
### What

Admission gaps around MinHash functions, each letting DDL persist collections that can never work:

- **`disableRuntimeCheck=true` paths skipped MinHash parameter validation.** `ValidateMinHashFunction` only ran inside `embedding.ValidateFunctions`, which `validator.ValidateFunction(..., disableRuntimeCheck=true)` skips entirely. The check is pure schema validation with no external calls, so hoist it into `validator.ValidateFunction` where it runs regardless of the runtime-check switch, and drop the now-redundant special case in the embedding executor.
- **The direct RootCoord CreateCollection path ran no function validation at all.** `prepareSchema` only validated fields and assigned IDs — a client hitting RootCoord directly could persist a MinHash (or any function) schema the proxy path rejects. Run `validator.ValidateFunction(schema, "", true)` after `assignFieldAndFunctionID` (IDs canonical). Scoped to newly created schemas: the `preserveFieldID` branch restores an existing collection (snapshot/replication), whose historical schema must not be re-judged by current-version rules, and **external collections are exempt** — `prepareSchema` sees the resolved schema (nullability inferred from the external source), not the user-submitted form the proxy validated, so judging it by user-schema rules rejects legal external tables (caught by the go-sdk e2e `TestExternalTableTextEmbeddingFunction`).
- **The validation itself was overflow-bypassable.** `ValidateMinHashFunction` compared `outputDim != int64(numHashes * 32)`: with dim=32 and `num_hashes=2^59+1` the multiplication wraps to 32 and the check passes, letting the runner later attempt an impossibly large permutation allocation. Compare via division (`outputDim%32 == 0 && outputDim/32 == numHashes`) so the hoisted check is actually authoritative. Defense-in-depth at the runner's allocation site is a follow-up (kept with the runner-hardening split of #51848).
- **`CheckFunctionInputField` admitted TEXT inputs for MinHash, contradicting the runner-side contract.** `ValidateMinHashFunction` has always been VarChar-only ("is not string type"), so the two validators disagreed on the same schema; with the gaps above closed, align the admission check to VARCHAR-only. Note: inserts into an existing TEXT-input collection do work today (TEXT arrives as `StringData` on the wire); the mismatch bites in the validator disagreement and in backfill-style paths reading persisted TEXT back as LOB references — this PR fixes admission, it does not claim runtime insert breakage.

Split from #51848.

issue: #51649

### Tests

- `Test_createCollectionTask_prepareSchema_validatesFunctions`: the real RootCoord create path rejects an overflowing `num_hashes` and accepts a valid MinHash schema.
- `Test_createCollectionTask_prepareSchema_skipsFunctionValidationForExternal`: an external table with a nullable TextEmbedding input (the resolved form) passes.
- `TestValidateMinHashFunctionRejectsNumHashesOverflowingDimCheck`: overflow boundary (`2^59+1`) rejected by the division-based check.
- New `validator_test.go`: MinHash params checked with runtime checks disabled.
- `TestValidateFunctionInputField` gains MinHash VarChar-accepted / TEXT-rejected cases.
- rootcoord `prepareSchema` suites, `internal/util/function`, `validator`, and `embedding` suites green locally under `-tags dynamic,test`.

### DIFF
--- a/internal/rootcoord/create_collection_task.go
+++ b/internal/rootcoord/create_collection_task.go
@@ -29,6 +29,7 @@ import (
 	"github.com/milvus-io/milvus/internal/json"
 	"github.com/milvus-io/milvus/internal/storagev2/packed"
 	"github.com/milvus-io/milvus/internal/streamingcoord/server/balancer"
+	"github.com/milvus-io/milvus/internal/util/function/validator"
 	"github.com/milvus-io/milvus/internal/util/hookutil"
 	"github.com/milvus-io/milvus/pkg/v3/common"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
@@ -703,6 +704,20 @@ func (t *createCollectionTask) prepareSchema(ctx context.Context) error {
 	} else {
 		if err := t.assignFieldAndFunctionID(t.body.CollectionSchema); err != nil {
 			return err
+		}
+		// Function schema validation for requests entering RootCoord directly —
+		// the proxy validates before forwarding, but nothing else on this path
+		// does. Runs only for newly created schemas: preserveFieldID restores an
+		// existing collection (snapshot/replication), whose historical schema
+		// must not be re-judged by current-version rules. External collections
+		// are exempt too: this schema is the RESOLVED one, not what the user
+		// submitted — e.g. nullability is inferred from the external source, so
+		// judging it by user-schema rules rejects legal external tables (the
+		// proxy already validated the user-submitted form).
+		if !typeutil.IsExternalCollection(t.body.CollectionSchema) {
+			if err := validator.ValidateFunction(t.body.CollectionSchema, "", true); err != nil {
+				return err
+			}
 		}
 	}
 	if err := typeutil.ValidateExternalCollectionResolvedSchema(t.body.CollectionSchema); err != nil {
--- a/internal/proxy/util_test.go
+++ b/internal/proxy/util_test.go
@@ -3356,6 +3356,31 @@ func TestValidateFunctionInputField(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("Valid MinHash function input - varchar", func(t *testing.T) {
+		function := &schemapb.FunctionSchema{
+			Type: schemapb.FunctionType_MinHash,
+		}
+		fields := []*schemapb.FieldSchema{
+			{DataType: schemapb.DataType_VarChar},
+		}
+		err := validator.CheckFunctionInputField(function, fields)
+		assert.NoError(t, err)
+	})
+
+	t.Run("Invalid MinHash function input - TEXT rejected", func(t *testing.T) {
+		// align with ValidateMinHashFunction's VarChar-only input contract:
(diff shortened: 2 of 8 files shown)


internal/rootcoord/create_collection_task.go
```diff
@@ -704,6 +705,14 @@ func (t *createCollectionTask) prepareSchema(ctx context.Context) error {
 		if err := t.assignFieldAndFunctionID(t.body.CollectionSchema); err != nil {
 			return err
 		}
+		// Function schema validation for requests entering RootCoord directly —
+		// the proxy validates before forwarding, but nothing else on this path
+		// does. Runs only for newly created schemas: preserveFieldID restores an
+		// existing collection (snapshot/replication), whose historical schema
+		// must not be re-judged by current-version rules.
+		if err := validator.ValidateFunction(t.body.CollectionSchema, "", true); err != nil {
```
buqian-zilliz: **[P2] Do not skip fresh milvus-table validation**

`preserveFieldID` is not limited to replay or restore: `prepareMilvusTableSnapshotSchema` also sets it for a brand-new milvus-table external collection after aligning field IDs. Because this validation is confined to the `else` branch, a request reaching RootCoord directly can still persist an invalid MinHash schema, such as a `num_hashes` value that does not match the binary-vector dimension. Please distinguish fresh snapshot alignment from historical replay and run the pure function validation for the fresh path as well.

buqian-zilliz: **[P2] Reject functions over the generated namespace field**

This validation runs after `handleNamespaceField` injects `$namespace_id`, so a direct RootCoord request can name that server-generated VARCHAR field as a TextEmbedding input and pass validation, even though the Proxy path rejects the same schema because the field is absent from the client schema. Normal inserts then call `genFunctionFields` before `addNamespaceData`, and the executor fails with `input field not found` on every insert. Validate against the client schema before generated fields are appended, or explicitly reject system fields as function inputs.

MrPresent-Han: Thanks for catching this. We agree that the remaining case requires a caller to bypass Proxy and submit a fresh milvus-table schema directly to RootCoord. The supported create path validates the user-submitted schema in Proxy before RootCoord performs snapshot field-ID alignment, while `preserveFieldID` is also used by restore/replay paths that intentionally avoid re-judging persisted schemas with current-version rules. We won't split fresh alignment from replay/restore validation in this PR; we'll keep the current scope and leave the direct-RootCoord milvus-table case as follow-up.

MrPresent-Han: Thanks for flagging this. This is a valid pre-existing namespace/function issue, but it is not introduced by this PR: before this change RootCoord already injected `$namespace_id` before `assignFieldAndFunctionID` and performed no function validation, so the same schema could already be persisted. This PR is scoped to the MinHash admission gaps and does not change namespace-field generation or function execution ordering. We won't address generated system fields as function inputs here and will leave that broader behavior for a separate follow-up.

PR #52220, opened 2026-08-06
fix: enforce the add-function backfill gates on the direct RootCoord path (#51649)

MrPresent-Han:
### What

The add-function backfill prerequisites — the TEXT-input rejection **and** the StorageV3 / schema-bump-compaction / storage-version-compaction requirements — lived as proxy-private functions, so they only guarded requests entering through the proxy. A request hitting RootCoord`"'s `add_function_field` DDL directly bypassed them and could register a BM25/MinHash function whose backfill compaction is impossible, leaving pre-existing segments unable to materialize the output (old and new rows with inconsistent function data).

- Move both gates to schemautil (`ValidateAddFunctionInputNotText`, `ValidateAddFunctionBackfillConfig`), each with its test, so they are callable outside the proxy.
- Enforce both in `broadcastAlterCollectionSchemaAdd`, right after function-schema validation — the direct path now rejects the same configurations as the proxy path.
- Switch the proxy call sites to the schemautil symbols and drop the private copies.

Split from #51848.

issue: #51649

### Tests

- `TestDDLCallbacksAlterCollectionSchemaRejectsTextFunctionInput`: direct path rejects a BM25 add over a TEXT input.
- `TestDDLCallbacksAlterCollectionSchemaRequiresBackfillConfig`: direct path rejects an add when any of StorageV3 / bump-schema / storage-version compaction is disabled.
- `TestValidateAddFunctionInputNotText`, `TestValidateAddFunctionBackfillConfig` moved/added in schemautil.
- rootcoord + schemautil + proxy tests green locally under `-tags dynamic,test`; residual references to the old proxy-private symbols grep to zero.

### DIFF
--- a/internal/rootcoord/ddl_callbacks_alter_collection_schema.go
+++ b/internal/rootcoord/ddl_callbacks_alter_collection_schema.go
@@ -130,6 +130,26 @@ func (c *Core) broadcastAlterCollectionSchemaAdd(ctx context.Context, broadcaste
 		if err := validator.ValidateFunction(schema, plan.Function.GetName(), true); err != nil {
 			return merr.Wrap(err, "invalid function schema")
 		}
+		isExternalCollection := typeutil.IsExternalCollection(schema)
+		// Non-milvus-table external sources expose TEXT as ordinary UTF8 input.
+		// Internal and milvus-table manifests expose TEXT as binary LOB references.
+		if !isExternalCollection || typeutil.NewStorageColumnResolver(schema).IsMilvusTable() {
+			if err := schemautil.ValidateAddFunctionInputNotText(schema, plan.Function); err != nil {
+				return err
+			}
+		}
+		// External collections materialize function outputs during refresh, not
+		// through internal schema-bump/storage-version compaction.
+		if !isExternalCollection {
+			if err := schemautil.ValidateAddFunctionBackfillConfig(
+				Params.DataCoordCfg.EnableCompaction.GetAsBool(),
+				Params.CommonCfg.UseLoonFFI.GetAsBool(),
+				Params.DataCoordCfg.BumpSchemaVersionCompactionEnabled.GetAsBool(),
+				Params.DataCoordCfg.StorageVersionCompactionEnabled.GetAsBool(),
+			); err != nil {
+				return err
+			}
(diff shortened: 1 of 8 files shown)


internal/rootcoord/ddl_callbacks_alter_collection_schema.go
```diff
@@ -130,6 +130,9 @@ func (c *Core) broadcastAlterCollectionSchemaAdd(ctx context.Context, broadcaste
 		if err := validator.ValidateFunction(schema, plan.Function.GetName(), true); err != nil {
 			return merr.Wrap(err, "invalid function schema")
 		}
+		if err := schemautil.ValidateAddFunctionInputNotText(schema, plan.Function); err != nil {
```
buqian-zilliz: **[P1] Revalidate the backfill configuration in RootCoord**

The new direct-RootCoord guard mirrors only the TEXT-input check, while `validateAddFunctionRequiresStorageV3` remains proxy-private. A caller that reaches `Core.AlterCollectionSchema` directly can therefore add a BM25/MinHash function over a VARCHAR input when StorageV3, schema-bump compaction, or storage-version compaction is disabled. RootCoord will broadcast the new schema and bound index, but pre-existing segments cannot materialize the output, leaving old and new rows with inconsistent function data. Enforce the same three backfill prerequisites at this coordinator boundary before allocating or broadcasting the schema change.

MrPresent-Han: Good catch — addressed in 4f2989a828. I extended the same move-and-enforce treatment to the backfill-config prerequisites: `validateAddFunctionRequiresStorageV3` is now `schemautil.ValidateAddFunctionBackfillConfig` (StorageV3/useLoonFFI, bumpSchemaVersion.enabled, storageVersion.enabled), and `broadcastAlterCollectionSchemaAdd` calls it right after the TEXT-input gate, before allocating the bound index or broadcasting the schema change. The proxy call site now uses the shared symbol and the private copy is dropped. Added `TestDDLCallbacksAlterCollectionSchemaRequiresBackfillConfig` asserting the direct RootCoord path rejects the add when any of the three is disabled.

PR #52221, opened 2026-08-06
fix: account MinHash input analyzer resources without enable_analyzer (#51649)

MrPresent-Han:
### What

A MinHash function input legally keeps `enable_analyzer=false`, yet its runner consumes the field's `analyzer_params` unconditionally (`NewMinHashFunctionRunner` -> `getAnalyzerParams`). The DDL-side collector `validateAnalyzer` however returned early for `enable_analyzer=false` fields, so a resource-backed `analyzer_params` (e.g. `resource://dict`) on a MinHash input never reached `ValidateAnalyzer`, never landed in `schema.FileResourceIds`, and never entered file-resource ref-counting — `RemoveFileResource` could then delete a resource the collection still depends on at runtime.

Fix: in the `enable_analyzer=false` early-return branch, still collect `analyzer_params` when the field is a MinHash function input. Matching is **by name** (`isMinHashInputFieldByName`): this runs before `assignFieldAndFunctionID`, so a direct CreateCollection request may carry stale non-canonical IDs and an ID-based check would silently miss the input.

Split from #51848.

issue: #51649

### Tests

- `TestCollectAnalyzerInfosIncludesDisabledMinHashInput`: disabled-analyzer MinHash input still emits its analyzer info.
- `TestCollectAnalyzerInfosMinHashInputMatchedByNameBeforeIDAssignment`: stale mismatched IDs, matching names — still collected.
- Both plus the existing `Test_validateMultiAnalyzerParams` suite green locally under `-tags dynamic,test`.

### DIFF
--- a/internal/rootcoord/create_collection_task.go
+++ b/internal/rootcoord/create_collection_task.go
@@ -935,6 +935,23 @@ func validateMultiAnalyzerParams(params string, coll *schemapb.CollectionSchema,
 	return nil
 }
 
+// isMinHashInputFieldByName reports whether fieldSchema is a MinHash function
+// input, keyed on the field name. Used in the pre-normalization schema
+// validation path where field IDs are not yet canonical.
+func isMinHashInputFieldByName(collSchema *schemapb.CollectionSchema, fieldSchema *schemapb.FieldSchema) bool {
+	for _, fn := range collSchema.GetFunctions() {
+		if fn.GetType() != schemapb.FunctionType_MinHash {
+			continue
+		}
+		for _, name := range fn.GetInputFieldNames() {
+			if name == fieldSchema.GetName() {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.AnalyzerInfo, error) {
 	analyzerInfos := make([]*querypb.AnalyzerInfo, 0)
 	for _, field := range collSchema.GetFields() {
@@ -946,7 +963,8 @@ func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.Ana
 }
 
 // validateAnalyzer validates active match/BM25 fields and fields with
-// enable_analyzer=true. Analyzer params are ignored while analyzer is disabled.
+// enable_analyzer=true. Analyzer params are ignored while analyzer is disabled,
+// except for MinHash inputs, which consume them regardless.
 func validateAnalyzer(collSchema *schemapb.CollectionSchema, fieldSchema *schemapb.FieldSchema, analyzerInfos *[]*querypb.AnalyzerInfo) error {
 	h := typeutil.CreateFieldSchemaHelper(fieldSchema)
 	active := h.EnableMatch() || typeutil.IsBm25FunctionInputField(collSchema, fieldSchema)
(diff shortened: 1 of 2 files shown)


internal/rootcoord/create_collection_task.go
```diff
@@ -946,14 +963,25 @@ func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.Ana
...
 	}
 	if !h.EnableAnalyzer() {
+		// A MinHash input legally keeps enable_analyzer=false yet its runner
+		// consumes analyzer_params. Collect them so any referenced file
+		// resources reach schema.FileResourceIds and ref-counting — otherwise
+		// the DDL gate admits a dependency the ledger never records and
+		// RemoveFileResource can delete a still-required resource. Match by name:
+		// this runs before assignFieldAndFunctionID, so field/function IDs are not
+		// yet canonical and an ID-based check would miss a stale-ID input.
+		if isMinHashInputFieldByName(collSchema, fieldSchema) {
```
buqian-zilliz, line 982: **[P2] Reconcile preserved MinHash input IDs before collecting resources**

`preserve_field_ids=true` skips `assignFieldAndFunctionID`, so the assumption here that IDs will be canonicalized does not hold on that accepted path. A schema can name field A as the MinHash input while preserving field B's ID; this code validates and records A's analyzer params, but `NewMinHashFunctionRunner` later selects B by `InputFieldIds`. B's file resource can therefore remain outside `FileResourceIds` and still be removed. Re-resolve or validate function field IDs on the preserve path, or use the same identity source as runtime.

internal/rootcoord/create_collection_task.go
```diff
@@ -946,14 +963,25 @@ func collectAnalyzerInfos(collSchema *schemapb.CollectionSchema) ([]*querypb.Ana
...
+			appendAnalyzerParams(fieldSchema, analyzerInfos)
```
buqian-zilliz, line 983: **[P2] Avoid loading MinHash-only resources in text-index compaction**

Collecting MinHash-only analyzer resources here makes `FileResourceIds` non-empty even when no field has `enable_match`. In `DNFileResourceMode=ref`, DataCoord puts every schema resource into mix/sort compaction plans, and DataNode's shared `createTextIndex` downloads them before its first `EnableMatch()` check. A MinHash-only collection therefore performs a cold remote download and can fail compaction on that I/O even though no text index is produced; the zero-ref cache entry is removed by the 15-minute GC, so the cost recurs. Gate text-index resource attachment or download on an actual match-enabled field.

internal/rootcoord/create_collection_task.go
buqian-zilliz: This issue still exists. The preserve-field-ID path still skips name-to-ID reconciliation: resource collection selects the MinHash input by `InputFieldNames`, while `NewMinHashFunctionRunner` selects the runtime field by `InputFieldIds`. A preserved schema with inconsistent names and IDs can therefore record field A's resource while consuming field B's analyzer.

internal/rootcoord/create_collection_task.go
buqian-zilliz: This issue still exists. Ref-mode mix/sort compaction still attaches every schema file resource, and `createTextIndex` still downloads those resources before checking whether any field has `EnableMatch()`. A MinHash-only collection can therefore still incur and fail on an analyzer-resource download even though no text index will be built.

MrPresent-Han:
Closing this PR because MinHash-related changes are no longer in scope.

Issue #51861, opened 2026-07-26
[Bug]: Populate DataCoord file-resource metadata in production


MrPresent-Han:
### Problem

DataCoord has an in-memory file-resource cache (`resourceIDMap`) and `UpdateFileResources` / `GetFileResources`, but there is no production path that populates or restores it:

- `UpdateFileResources` is called only by unit tests.
- DataCoord `reloadFromKV` does not call `ListFileResource`.
- The DataCoord catalog exposes `SaveFileResource` / `ListFileResource`, but they have no production callers.
- The coordinator file-resource observer synchronizes resources to QueryNode/DataNode in sync mode, but does not update DataCoord metadata.

As a result, RefMode jobs that resolve schema `FileResourceIds` through DataCoord fail with `file resource N not found` even though RootCoord owns the resource metadata.

### Affected paths

- Mix and Sort compaction plan construction.
- Namespace clustering compaction plan construction.
- Text-index stats jobs.
- Bump-schema-version compaction after PR #51848 wires it to the same existing contract.

### Expected behavior

RootCoord authoritative `FileResourceInfo` entries and their version should be available in DataCoord:

1. Restore the current resource set when DataCoord starts or restarts.
2. Propagate resource add/remove/version updates while the cluster is running.
3. Keep `GetFileResources(ids...)` usable in RefMode without relying on tests to seed the map manually.

### Scope note

PR #51848 only adds the BumpSchemaVersion-side wiring and contract tests. The production population/recovery mechanism is intentionally tracked here as the shared cross-compaction fix.

MrPresent-Han:
@aoiasd This is the pre-existing DataCoord file-resource metadata population gap discussed in PR #51848. Could you please take ownership of the shared fix? I do not have repository permission to set the assignee metadata directly.

MrPresent-Han:
non-blocking-3.0

MrPresent-Han:
Ai-Scanned-Found Problem, need to be confirmed

Issue #52141, opened 2026-08-03
[Enhancement]: Preflight text-index file resources before schema-bump full rewrite


MrPresent-Han:
### Problem

In the schema-bump full-rewrite path, text-index file resources are prepared only after the expensive rewrite has already completed:

```text
read source segment
→ materialize and write replacement segment
→ commit/merge manifest and LOB references
→ Download FileResources
→ BuildExtraResourceInfo
→ create text index
```

The final resource preparation happens in `internal/datanode/compactor/compactor_common.go::createTextIndex`, called near the end of `runFullSchemaRewrite`.

A missing resource, download failure, malformed resource mapping, dictionary-file failure, or analyzer initialization failure therefore discards the completed rewrite work. Policy retry repeats the same full-segment I/O.

PR #51848 temporarily added a coarse fail-fast based on:

```go
schemaNeedsTextIndex(schema) &&
len(schema.FileResourceIds) > 0 &&
len(plan.FileResources) == 0
```

That check is being removed because `schema.FileResourceIds` is a collection-wide analyzer-resource union. It can reject an inline/built-in match analyzer merely because an unrelated non-match field owns a file resource.

### Goal

Introduce a task-scoped `prepareTextIndexResources` stage for schema-bump full rewrite so resource preparation failures are reported before any segment reader/writer I/O.

Target flow:

```text
determine the resources required by enable_match analyzers
→ check required resources against the plan
→ download required resources
→ build AnalyzerExtraInfo
→ validate match analyzers
→ hold a resource lease
→ full schema rewrite
→ merge LOB references
→ create text index using the prepared context
→ release the lease
```

The actual `CreateIndex` call must remain after output-manifest and LOB-reference assembly; only resource preparation moves earlier.

### Required behavior

- Do not use collection-level `len(FileResourceIds) > 0` as proof that a match field needs resources.
- Determine the resource set required by `enable_match` fields.
- Compare the required and provided sets and report missing resources before rewrite.
- Download only text-index-required resources.
- Build and reuse the same `AnalyzerExtraInfo` during final index construction.
- Keep the acquired resources alive through rewrite and index upload.
- Release exactly once on success, cancellation, and every failure path.
- Preserve typed storage, timeout, and cancellation errors with `merr.Wrap/Wrapf`.
- Initially scope the implementation to schema-bump full rewrite. Mix/sort/stats-task adoption can be evaluated separately.
[...]

PR #52222, opened 2026-08-06
enhance: carry analyzer file resources into bump compaction plan (#51649)

MrPresent-Han:
### What

Two mirrored gaps around analyzer file resources in the schema-bump compaction path, under DataNode file-resource **ref mode**:

- **DataCoord**: bump `BuildCompactionRequest` never attached the schema's `FileResources` to the plan, while sort/mix compactions do. The bump full-rewrite path builds the text-match index inline (`createTextIndex`), which under ref mode resolves analyzer resources from the plan — without them the rebuild cannot resolve a referenced dictionary. Attach them **best-effort**: only when a field enables match and the schema lists resource ids; on lookup failure keep building the plan without them, since bump-only / partial-backfill tasks never consume resources and must not fail over an unused lookup (a full rewrite that does need them still fails next to the actual consumer, `createTextIndex`).
- **DataNode**: `createTextIndex` downloaded the plan's file resources before checking whether any field enables match — the download exists solely to build text-match indexes. Skip all setup when no field enables match (`schemaNeedsTextIndex`), so unrelated resource-owning analyzer fields (the `FileResourceIds` union is collection-wide, not per-field) never block or slow a compaction that builds no text index.

Known limitation, not claimed fixed here: DataCoord's file-resource view has no production population/recovery path yet (`UpdateFileResources` is test-only) — tracked by #51861. This PR wires the plan-side carry so bump matches sort/mix once #51861 lands.

Split from #51848.

issue: #51649

### Tests

- `TestBuildCompactionRequest_BumpFileResourcesInRefMode`: ref mode carries resources; sync mode skips; missing resource degrades to empty carry without failing the plan; no-match schema skips.
- `TestFullRewriteDoesNotRejectUnrelatedRefModeResources`: a match field on the built-in analyzer is not rejected because an unrelated field owns a resource.
- Full datacoord bump suite plus all compactor `*TextIndex*` regressions green locally under `-tags dynamic,test`.

### DIFF
--- a/internal/datacoord/compaction_task_bump_schema_version.go
+++ b/internal/datacoord/compaction_task_bump_schema_version.go
@@ -26,15 +26,18 @@ import (
 	"go.uber.org/atomic"
 	"google.golang.org/protobuf/proto"
 
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/compaction"
 	"github.com/milvus-io/milvus/internal/datacoord/allocator"
 	"github.com/milvus-io/milvus/internal/datacoord/session"
 	"github.com/milvus-io/milvus/internal/metastore/kv/binlog"
+	"github.com/milvus-io/milvus/internal/util/fileresource"
 	"github.com/milvus-io/milvus/pkg/v3/mlog"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
 	"github.com/milvus-io/milvus/pkg/v3/taskcommon"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
+	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
 
 var _ CompactionTask = (*bumpSchemaVersionTask)(nil)
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
 		JsonParams:                compactionParams,
 		CurrentScalarIndexVersion: t.ievm.ResolveScalarIndexVersion(),
 	}
+
+	// set analyzer resource for text match index if use ref mode.
+	// Only the full-rewrite path consumes these (inline createTextIndex, mirroring
+	// sort/mix); which path a segment takes is decided on the datanode, so attach
+	// best-effort: on failure keep building the plan without resources instead of
+	// failing tasks (bump-only / partial backfill) that never need them. A
+	// full-rewrite that does need them fails at createTextIndex resource
+	// resolution, next to the actual consumer.
+	if fileresource.IsRefMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue()) &&
+		schemaNeedsTextIndex(taskProto.GetSchema()) &&
+		len(taskProto.GetSchema().GetFileResourceIds()) > 0 {
+		resources, err := t.meta.GetFileResources(context.Background(), taskProto.GetSchema().GetFileResourceIds()...)
+		if err != nil {
+			mlog.Warn(context.TODO(), "get file resources for schema bump compaction failed, building plan without them",
+				mlog.Int64("collectionID", taskProto.GetCollectionID()), mlog.Err(err))
+		} else {
+			plan.FileResources = resources
+		}
+	}
+
 	segments := make([]*SegmentInfo, 0, len(taskProto.GetInputSegments()))
 	for _, segID := range taskProto.GetInputSegments() {
 		segInfo := t.meta.GetHealthySegment(context.TODO(), segID)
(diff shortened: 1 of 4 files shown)


internal/datacoord/compaction_task_bump_schema_version.go
```diff
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
...
+
+	// set analyzer resource for text match index if use ref mode.
+	// Only the full-rewrite path consumes these (inline createTextIndex, mirroring
+	// sort/mix); which path a segment takes is decided on the datanode, so attach
+	// best-effort: on failure keep building the plan without resources instead of
+	// failing tasks (bump-only / partial backfill) that never need them. A
+	// full-rewrite that does need them fails at createTextIndex resource
+	// resolution, next to the actual consumer.
+	if fileresource.IsRefMode(paramtable.Get().CommonCfg.DNFileResourceMode.GetValue()) &&
+		schemaNeedsTextIndex(taskProto.GetSchema()) &&
```
buqian-zilliz, line 128: **[P2] Do not exclude analyzer-backed function materialization**

The new gate assumes that analyzer resources are needed only when a text-match index is built, but `add_function_field` backfills BM25 and MinHash through `NewRecordMaterializer`. Those runners construct analyzers before `createTextIndex`; in DataNode ref mode, resource-backed analyzers need downloaded resources plus analyzer extra info at that point. Without `enable_match` this condition omits resources, while with `enable_match` the partial path never calls `createTextIndex` and the full-rewrite path reaches it only after the materializer was created, so the supported schema bump deterministically fails. Plumb resource extra info into the materializer/function runners, or reject resource-backed BM25/MinHash backfill in ref mode before scheduling it.

internal/datacoord/compaction_task_bump_schema_version.go
```diff
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
...
+		len(taskProto.GetSchema().GetFileResourceIds()) > 0 {
+		resources, err := t.meta.GetFileResources(context.Background(), taskProto.GetSchema().GetFileResourceIds()...)
```
buqian-zilliz, line 130: **[P2] Filter resources to the match-enabled analyzers**

`CollectionSchema.FileResourceIds` is a collection-wide union over analyzer-enabled fields, but this passes the entire union whenever any field has `enable_match`. `createTextIndex` then downloads every plan resource before it filters to match fields, so a built-in match analyzer becomes dependent on dictionaries owned only by unrelated non-match or BM25 fields; an unavailable unrelated file aborts the full rewrite, and healthy large dictionaries add needless ref-mode I/O. Resolve and download only the resources referenced by match-enabled analyzers, or defer that field-level selection to the DataNode.

internal/datacoord/compaction_task_bump_schema_version.go
```diff
@@ -113,6 +116,26 @@ func (t *bumpSchemaVersionTask) BuildCompactionRequest() (*datapb.CompactionPlan
...
+		if err != nil {
+			mlog.Warn(context.TODO(), "get file resources for schema bump compaction failed, building plan without them",
```
buqian-zilliz, line 132: **[P2] Preflight unresolved resources before rewriting the segment**

Ignoring this lookup error sends an incomplete plan to the DataNode even when the task later selects the full-rewrite path. `runFullSchemaRewrite` closes the writer and commits the replacement files and manifest before `createTextIndex` initializes the match analyzer; if that analyzer needs a missing resource, the task fails only after the full rewrite. Cleanup clears the original segment's compacting flag without advancing its schema version, so the periodic bump policy can select it again and repeat the rewrite, leaving each unadopted output for garbage collection. Preserve enough resource-resolution state to preflight the required match analyzers after the DataNode chooses the path but before opening the rewrite writer.

MrPresent-Han: Thanks. The late resource-preparation failure described here is real, but it is not introduced by this PR. Before this change, Bump plans never carried FileResources, so a resource-backed match analyzer already reached createTextIndex only after the full rewrite and failed there.

This PR improves the successful metadata-lookup path. It intentionally keeps plan construction best-effort because DataCoord cannot know whether the DataNode will choose bump-only, partial materialization, or full rewrite.

Moving resource download and validation ahead of expensive work is a broader FileResource lifecycle/robustness change shared with other consumers, and we do not want this focused fix PR to redesign that mechanism. Bump-specific path-aware preflight is tracked by #52141, while #51861 tracks DataCoord resource population. No code change is planned here.

MrPresent-Han:
Closing this PR. A correct solution needs explicit per-analyzer resource dependency selection and a robust resource lifecycle rather than making the Bump path consume the collection-wide FileResourceIds union.

Keeping #51649 open. #51861 tracks DataCoord resource population/recovery, and #52141 tracks Bump text-index resource selection and preflight.

PR #52223, opened 2026-08-06
fix: remove FieldBinlog emptiness guards that reject V3 segments

tedxu:
See #50406

V3 segments store data in LOON manifests. After a DataCoord
restart, resetBinlogFields nils the arrays before serialization
and applyBinlogInfo has no per-field KVs to restore for V3,
leaving FieldBinlogs empty by design.

The emptiness guards in mix_compactor (isEmpty / lo.EveryBy)
and clustering_compactor (binlogNum in mappingSegment and
scalarAnalyzeSegment) treated empty arrays as illegal plans,
causing deterministic rejection and infinite fail-retry loops
for restarted V3 segments.

Remove the guards. The downstream record reader already routes
V3 segments to ManifestRecordReader via a manifest check,
independent of FieldBinlogs. Use plan.GetTotalRows() for
bloom-filter sizing instead of iterating binlog arrays.

issue: #50406

### DIFF
--- a/internal/datanode/compactor/clustering_compactor.go
+++ b/internal/datanode/compactor/clustering_compactor.go
@@ -592,20 +592,6 @@ func (t *clusteringCompactionTask) mappingSegment(
 		}
 	}
 
-	// Get the number of field binlog files from non-empty segment
-	var binlogNum int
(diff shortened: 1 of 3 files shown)


XuanYang-cn (approved):
/lgtm

PR #52224, opened 2026-08-06
fix: [cp3.0]read only the needed leaf columns for struct array sub-fields

bigsheeper:

pr: #52046
issue: #52045

## Summary

Cherry-picked from master PR #52046 (open - preemptive backport).

`NewStructFieldReader` passed the same top-level column index to `GetColumn` for every struct sub-field, so each sub-field reader decoded all leaves of the whole `list<struct>` column. This resolves each sub-field's leaf column indices from the pqarrow manifest and reads through `GetFieldReader`, pruning the rest.

**Note**: Original PR is still open. This CP PR should be updated if the original PR changes.

## Verification

- [x] File count matches original PR (2 files, +559/-3)
- [x] Per-file diff verified identical to master PR (no conflicts during cherry-pick)
- [x] No conflict markers

### DIFF
--- a/internal/util/importutilv2/parquet/leaf_pruning_test.go
+++ b/internal/util/importutilv2/parquet/leaf_pruning_test.go
@@ -1,0 +1,494 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


liliu-z (commented):
/lgtm
/approve

PR #52228, opened 2026-08-06
fix: set correct content type for 9091 APIs

### LINKED ISSUE #52230: [Bug]: Incorrect Content-Type for JSON responses on port 9091

### Environment

```markdown
- Milvus version: 3.0
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

Some APIs on port 9091 return a JSON body without setting `Content-Type: application/json`. Go may therefore send the response as `text/plain`.

Some error responses are also built by hand, which may produce invalid JSON when the error message contains quotes or control characters.

### Expected Behavior

- JSON responses should use `Content-Type: application/json`.
- Text responses should use `Content-Type: text/plain`.
- All JSON responses should be valid JSON.

### Steps To Reproduce

```markdown
1. Start a Milvus instance.

2. Call a management API on port 9091:

  curl -i http://localhost:9091/management/querycoord/node/list

3. Check the response headers and body.

  The body is JSON, but the response header is: Content-Type: text/plain; charset=utf-8

4. Call an error path:

  curl -i "http://localhost:9091/management/querycoord/transfer/segment?source_node_id=invalid"

  Some error paths build the response by hand and may return invalid JSON.

```

### Milvus Log

_No response_

### Anything else?

_No response_

fudongyingluck:
## What changed

This PR fixes  Fixes #52230 response headers and JSON output for several APIs on port 9091.

It covers:

- `/expr`
- `/healthz`
- `/management/stop`
- Proxy management APIs

It also uses JSON marshal instead of building JSON strings by hand.

## Notes
- Zap had the same content type issue. A fix has already been sent: https://github.com/uber-go/zap/pull/1567

- I first wanted to put `writeJSONWithMsg` in a shared package, but I could not find a good package for it. For now, the helper stays close to the related handlers.

- Port 9091 currently uses both `http.ServeMux` and Gin. In the long term, moving these routes to Gin may be a good option. It could make routing, middleware, logging, and recovery easier to share, and reduce code complexity. However, this would be a large change, so I would like to hear your thoughts first.  

## Test

- Proxy management unit tests passed.
- The updated 9091 APIs were tested with a running standalone Milvus.
- Response status, content type, and JSON body were checked.

milvus-emoji-05-thx-unofficial-small


internal/http/server.go
```diff
@@ -160,13 +160,12 @@ func RegisterStopComponent(triggerComponentStop func(role string) error) {
...
-				w.WriteHeader(http.StatusInternalServerError)
-				fmt.Fprintf(w, `{"msg": "failed to trigger component stop, %s"}`, err.Error())
+				writeJSONWithMsg(w, http.StatusInternalServerError,
+					fmt.Sprintf("failed to trigger component stop, %s", err.Error()))
 				return
 			}
 			mlog.Info(ctx, "finish to trigger component stop", mlog.String("role", role))
-			w.WriteHeader(http.StatusOK)
-			w.Write([]byte(`{"msg": "OK"}`))
+			writeJSONWithMsg(w, http.StatusOK, "OK")
```
yhmo, line 168: internal/http/server.go line:168
Medium ---- Could we also route `RegisterCheckComponentReady` through `writeJSONWithMsg`? That active `/management/check/ready` route is registered next to `/management/stop` in `cmd/roles/roles.go`, but it still writes hand-built JSON after `WriteHeader` and never sets `Content-Type`. Because its callback error includes the caller-supplied `role`, a value containing a quote still produces invalid JSON, leaving the same #52230 behavior on a 9091 management API.

fudongyingluck: Thanks for the suggestion! I’ve updated the RegisterCheckComponentReady to use writeJSONWithMsg as well.

yhmo: The issue is confirmed fixed. Resolving this conversation.

fudongyingluck:
I checked the logs. The ci-rerun-e2e-default failure seems unrelated to the changes in this PR.

fudongyingluck:
@yhmo Do I need to do anything else for this PR?

PR #47943, opened 2026-02-27
enhance: upgrade Go dependencies (casbin, gin, lo, cockroachdb/errors)

liliu-z:
## Summary
- Upgrade Go dependencies across root, pkg, and client modules:
  - `casbin/casbin` v2.44.2 → v2.135.0
  - `gin-gonic/gin` v1.9.1 → v1.11.0
  - `samber/lo` v1.27.0 → v1.52.0
  - `cockroachdb/errors` v1.9.1 → v1.12.0
  - `google.golang.org/protobuf` v1.36.5 → v1.36.9
- Adapt source code to breaking API changes:
  - `lo.Last()` now returns `(T, bool)` instead of `(T, error)` (lo v1.52)
  - `gin.LogFormatterParams.Keys` is now `map[any]any` instead of `map[string]any` (gin v1.11)

issue: https://github.com/milvus-io/milvus/issues/33482

## Test plan
- [x] `go mod tidy` clean on all three modules (root, pkg, client)
- [x] Local lint passes with no new errors
- [ ] CI code-check passes
- [ ] CI ut-go passes
- [ ] CI e2e passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #52231, opened 2026-08-06
fix: [2.6] restore REST v2 proxy request metrics

### LINKED ISSUE #52177: [Bug]: REST v2 requests no longer update milvus_proxy_req_count on v2.6.18

- [x] I have searched the existing issues and pull requests.

### Environment

- Milvus version: v2.6.18 and current 2.6 branch
- Deployment mode: cluster (the affected code path is shared by REST v2 deployments)
- SDK version: REST v2 API
- Others:
  - Gin v1.9.1
  - Regression introduced by the 2.6 backport in #50006 / commit `e261ec51f4b6bd6c6dd1ac5f5cd14c1e25223e16`

### Current Behavior

Successful REST v2 operations increment `milvus_proxy_restful_api_req_count{path=...}`, but do not increment the canonical operation counter:

```promql
milvus_proxy_req_count{
  function_name="Search",
  status=~"total|success"
}
```

The same applies to the other REST v2 operations mapped by `routeToMethod`, including Query, HybridSearch, Insert, Upsert, Delete, and Import.

This silently makes request-rate dashboards and alerts based on `milvus_proxy_req_count` under-report REST traffic. Other execution-level metrics such as search vector count still update, so a deployment can show non-zero VPS with near-zero QPS.

### Expected Behavior

Each completed REST v2 request should update `milvus_proxy_req_count` exactly as implemented by #47213:

1. increment `status="total"`;
2. increment exactly one outcome series such as `status="success"` or `status="fail"`;
3. preserve the mapped `function_name`, database, and collection labels.

### Steps To Reproduce

1. Start Milvus v2.6.18 with metrics enabled.
2. Record the values of:

   ```promql
   milvus_proxy_restful_api_req_count{path="/v2/vectordb/entities/search"}
   ```

   and:

   ```promql
   milvus_proxy_req_count{function_name="Search",status=~"total|success"}
   ```

3. Send a valid successful request to `POST /v2/vectordb/entities/search`.
4. Observe that `milvus_proxy_restful_api_req_count` increments, while the Search total/success operation counters do not.

The failure can also be reproduced directly from the context behavior:

1. The 2.6 timeout middleware invokes the route handler on `handlerCtx := gCtx.Copy()`.
2. Gin v1.9.1 `Context.Copy()` does not copy the private `fullPath` field.
3. `wrapperPost` evaluates `routeToMethod[gCtx.FullPath()]` on the copied context.
4. `FullPath()` returns an empty string, so the lookup fails and the function returns before either `ProxyFunctionCall` increment.

### Milvus Log

There is no error log. The failure is a silent metric-coverage regression.

### Anything else?

Relevant source:

- v2.6.18 copies the handler context in [timeout_middleware.go](https://github.com/milvus-io/milvus/blob/v2.6.18/internal/distributed/proxy/httpserver/timeout_middleware.go#L259-L270).
- The REST metric update depends on `gCtx.FullPath()` in [handler_v2.go](https://github.com/milvus-io/milvus/blob/v2.6.18/internal/distributed/proxy/httpserver/handler_v2.go#L386-L408).
- v2.6.18 uses [Gin v1.9.1](https://github.com/milvus-io/milvus/blob/v2.6.18/go.mod#L16).
- Gin v1.9.1 [Context.Copy()](https://github.com/gin-gonic/gin/blob/v1.9.1/context.go#L112-L130) omits `fullPath`; [FullPath()](https://github.com/gin-gonic/gin/blob/v1.9.1/context.go#L160-L162) only returns that field.
- Gin fixed this in [gin-gonic/gin#3784](https://github.com/gin-gonic/gin/pull/3784), available from v1.10.
- REST-to-`milvus_proxy_req_count` instrumentation was originally added in [#47213](https://github.com/milvus-io/milvus/pull/47213).
- The regression was introduced in the 2.6 backport [#50006](https://github.com/milvus-io/milvus/pull/50006). Master was not affected because it used a newer Gin version whose `Copy()` preserves `fullPath`.

Potential fixes include:

- use `gCtx.Request.URL.Path` for the static REST v2 route mapping;
- capture/preserve the route path before copying the context; or
- backport Gin's `fullPath` copy fix / upgrade Gin after dependency review.

Please add a regression test that executes a registered REST route through `timeoutMiddleware(wrapperPost(...))` and asserts the `milvus_proxy_req_count` total/outcome deltas.

fancyboi999:
issue: #52177
pr: #47943

## What changed

- preserve the registered Gin route pattern when the timeout middleware copies the request context
- use the preserved route for REST v2 proxy metrics and the affected authorization/rate-limit audit hooks
- add regression coverage through the real `timeoutMiddleware(wrapperPost(...))` chain and a dynamic-route case

## Root cause

Milvus 2.6 uses Gin v1.9.1. In that version, `Context.Copy()` does not retain Gin's private `fullPath` field. REST v2 handlers run on a copied context, so `gCtx.FullPath()` becomes empty inside `wrapperPost`. The route-to-method lookup then misses and returns before incrementing `milvus_proxy_req_count`, even though the request itself succeeds.

The middleware now captures the registered route pattern before copying the context and makes it available to downstream handlers. It intentionally does not fall back to `Request.URL.Path`, because dynamic routes need the stable registered template rather than high-cardinality concrete paths.

Master is already unaffected through #47943, which upgraded Gin to v1.11.0. Gin v1.11.0 copies `fullPath` in `Context.Copy()`; the 2.6 branch remains on v1.9.1 and needs this scoped compatibility fix.

## Validation

- negative regression check: restoring the old `gCtx.FullPath()` lookup returned HTTP 200 but produced metric delta `0` instead of `1`
- focused regression tests: PASS with `-race` (`2/2` tests)
- full `internal/distributed/proxy/httpserver` package: PASS with `-race`, `89.9%` statement coverage
- repository-pinned `golangci-lint v2.11.3` on the changed package: `0 issues`
- repository format target and `git diff --check`: PASS
- C++ `milvus_core` production target linked successfully for the cgo-backed Go tests; unrelated C++ unit-test executables were not run

Closes #52177

### DIFF
--- a/internal/distributed/proxy/httpserver/timeout_middleware.go
+++ b/internal/distributed/proxy/httpserver/timeout_middleware.go
@@ -225,6 +225,15 @@ var timeoutContextKeysToPropagate = []string{
 	"traceID",
 }
 
+const copiedContextRoutePathKey = "milvus_copied_context_route_path"
+
+func matchedRoutePath(c *gin.Context) string {
+	if path := c.FullPath(); path != "" {
+		return path
+	}
+	return c.GetString(copiedContextRoutePathKey)
+}
+
 func propagateTimeoutContextKeys(dst *gin.Context, src *gin.Context) {
 	for _, key := range timeoutContextKeysToPropagate {
 		if value, ok := src.Get(key); ok {
@@ -256,7 +265,12 @@ func timeoutMiddleware(handler gin.HandlerFunc) gin.HandlerFunc {
 		buffer := bufPool.Get()
 		buffer.Reset()
 		recorder := newTimeoutResponseRecorder(buffer)
+		// gin.Context.Copy did not retain the matched route path before Gin 1.11.
+		// Preserve the registered route pattern for metrics and audit hooks running
+		// on the copied context; Request.URL.Path is unsuitable for dynamic routes.
+		routePath := matchedRoutePath(gCtx)
 		handlerCtx := gCtx.Copy()
+		handlerCtx.Set(copiedContextRoutePathKey, routePath)
 		handlerCtx.Request = req
 		handlerCtx.Writer = recorder
 
--- a/internal/distributed/proxy/httpserver/timeout_middleware.go
+++ b/internal/distributed/proxy/httpserver/timeout_middleware.go
@@ -265,7 +265,7 @@ func timeoutMiddleware(handler gin.HandlerFunc) gin.HandlerFunc {
 		buffer := bufPool.Get()
 		buffer.Reset()
 		recorder := newTimeoutResponseRecorder(buffer)
-		// gin.Context.Copy did not retain the matched route path before Gin 1.11.
+		// gin.Context.Copy did not retain the matched route path before Gin 1.10.
 		// Preserve the registered route pattern for metrics and audit hooks running
 		// on the copied context; Request.URL.Path is unsuitable for dynamic routes.
 		routePath := matchedRoutePath(gCtx)
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -384,7 +384,7 @@ func wrapperPost(newReq newReqFunc, v2 handlerFuncV2) gin.HandlerFunc {
 			zap.Any("url", gCtx.Request.URL.Path))
 
 		resp, err := v2(ctx, gCtx, req, dbName)
-		methodTag, ok := routeToMethod[gCtx.FullPath()]
+		methodTag, ok := routeToMethod[matchedRoutePath(gCtx)]
 		if !ok {
 			return
 		}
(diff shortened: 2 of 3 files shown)


fancyboi999:
All required CI checks are green, including `codecov/patch` at 92.85% (target: 80%). This fix is intentionally scoped to the 2.6 branch: master upgraded to Gin v1.11.0 in #47943, where `Context.Copy()` preserves the matched route path.

@cydrain @czs007, when you have a chance, could you please review this 2.6 fix?

internal/distributed/proxy/httpserver/timeout_middleware.go
```diff
@@ -256,7 +265,12 @@ func timeoutMiddleware(handler gin.HandlerFunc) gin.HandlerFunc {
 		buffer := bufPool.Get()
 		buffer.Reset()
 		recorder := newTimeoutResponseRecorder(buffer)
+		// gin.Context.Copy did not retain the matched route path before Gin 1.11.
```
yhmo: internal/distributed/proxy/httpserver/timeout_middleware.go line:268
Low ---- Could we change this compatibility boundary to "before Gin 1.10" (or just "Gin 1.9.1")? Gin v1.10.0 already copies `fullPath` in `Context.Copy()` via gin-gonic/gin#3784, so saying "before Gin 1.11" may cause a future 1.10-based branch to keep this shim unnecessarily.

fancyboi999: Updated the compatibility boundary to "before Gin 1.10" in 0e28f33a7. I verified this against Gin v1.9.1, which does not copy `fullPath`, and Gin v1.10.0 / gin-gonic/gin#3784, which does. Thanks for catching the inaccurate boundary.

Issue #49980, opened 2026-05-21
[Bug]: Concurrent Partial Updates on the same entity cause "Lost Update" (Data Override)


wu-xingying:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 2.6.8
- Deployment mode(standalone or cluster):cluster
- MQ type(rocksmq, pulsar or kafka):    kafka
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

When multiple clients/threads perform Partial Update ( client.upsert(partial_update=True) ) concurrently on the same entity (same Primary Key) but updating different fields , a severe "Lost Update" issue occurs.

Because Milvus handles Partial Update via a Read-Modify-Write pattern in the Proxy layer, multiple concurrent requests will fetch the same snapshot of the old data. They merge their specific fields independently in memory and then append the Delete + Insert messages to the message queue.
The last appended message will overwrite the previous ones, causing the fields updated by the earlier concurrent requests to be silently lost, even though all upsert requests return Success to the client.

### Expected Behavior

When two concurrent partial updates target different fields of the same entity (e.g., Thread A updates age , Thread B updates name ), both updates should be preserved (e.g., the final entity should have the new age AND the new name ).

### Steps To Reproduce

We can easily reproduce this by using threading.Barrier in Python to force two partial updates to hit the Proxy at the exact same time.

```markdown
import json 
import random 
import threading 
import time 

from pymilvus import DataType, MilvusClient 


URI = "http://localhost:19530" 
TOKEN = "root:xxxxxxx" 

COLLECTION_NAME = "test_partial_update_lost_update" 
DIM = 4 
TEST_ID = 1 
ROUNDS = 50 


def make_client(): 
    return MilvusClient(uri=URI, token=TOKEN) 


def random_vec(): 
    return [round(random.random(), 4) for _ in range(DIM)] 


def print_section(title): 
    print("\n" + "=" * 24 + f" {title} " + "=" * 24) 


def pretty_row(row): 
    ordered = { 
        "id": row.get("id"), 
        "age": row.get("age"), 
        "name": row.get("name"), 
        "optional_desc": row.get("optional_desc"), 
        "info": row.get("info"), 
        "dynamic_key_A": row.get("dynamic_key_A"), 
        "vector": row.get("vector"), 
    } 
[...]

PR #52235, opened 2026-08-06
enhance: Define partial update CAS attempt proof

weiliu1031:
issue: #49980

## Design review

- Primary approver: @chyezh
- Independent approver: @liliu-z
- Review date: 2026-08-07

## What changed

- Add the design proposal for optimistic partial-update commit admission.
- Limit `PartialUpdateCAS` attempt proof to the read timestamp and observed
  PChannel term.
- Reserve the removed primary-key field ID and collection ID protobuf field
  numbers and names.
- Derive collection and schema identity from the Insert header and resolve the
  authoritative PK descriptor from ShardManager.
- Reserve a retryable streaming status for rebuilding a complete
  partial-update attempt.
- Regenerate the corresponding Go protobuf bindings.

## Why

Concurrent partial updates require a stable internal contract between the
Proxy snapshot read and StreamingNode commit admission. The proof carries only
state that StreamingNode cannot derive, while the Insert header and
ShardManager remain the authoritative source of write identity.

## Behavior boundaries

- This PR does not enable CAS or change runtime write behavior.
- No public RPC or client API is added.
- Removed protobuf field numbers and names are reserved and cannot be reused.
- Runtime implementation remains in
  [#51845](https://github.com/milvus-io/milvus/pull/51845).

## Validation

- `GOTOOLCHAIN=auto make generated-proto-without-cpp`
- `git diff --check`


chyezh:
Could we simplify `PartialUpdateCAS` so it contains only the attempt-scoped CAS proof?

```proto
message PartialUpdateCAS {
    uint64 read_ts = 1;
    int64 observed_pchannel_term = 2;

    reserved 3, 4;
    reserved "primary_key_field_id", "collection_id";
}
```

`read_ts` and `observed_pchannel_term` cannot be derived by StreamingNode and therefore need to come from Proxy. However, `collection_id` and `primary_key_field_id` can already be derived from the Insert message and the local schema:

- `InsertMessageHeader` carries `collection_id` and `schema_version`, and the upstream shard interceptor validates the schema version.
- The partial-update interceptor already has `GetPrimaryKeyDescriptor(collectionID, schemaVersion)`, so it can obtain the actual PK field ID from ShardManager and extract the PKs from the Insert body.

Keeping the latter two fields creates duplicate sources of truth, and the current implementation does not fully cross-check them:

- Commit admission calls `fences.Verify` with `txnState.meta.GetCollectionId()` without verifying that it matches the Insert header collection ID. Incorrect metadata could therefore check the fence for the wrong collection.
- PK extraction uses `meta.GetPrimaryKeyFieldId()` and only verifies that the field exists and is Int64 or VarChar; it does not verify that the field is the PK defined by the schema.

I suggest deriving these values on StreamingNode:

1. Read `collection_id` and `schema_version` from the Insert header.
2. Require CAS Inserts to carry an explicit `schema_version`.
3. Resolve the PK descriptor through ShardManager and extract the real PK field.
4. Store the derived collection ID in `pendingTxn`.
5. Verify that all CAS chunks in the transaction use the same proof, collection, and schema.
6. Use `pendingTxn.collectionID` for collection-fence validation at commit.

This keeps the wire metadata limited to information that is truly not derivable, removes Proxy and StreamingNode consistency risks, and gives the protocol a clearer fail-closed boundary.

Issue #52236, opened 2026-08-06
[Bug]: MilvusService RPCs can miss RBAC privilege annotations without CI failure


shaoting-huang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master @ 82ecdb68f5
- Deployment mode(standalone or cluster): standalone or cluster
- MQ type(rocksmq, pulsar or kafka): N/A
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): N/A
- CPU/Memory: N/A
- GPU: N/A
- Others:
```

### Current Behavior

`PrivilegeInterceptor` treats a request type without the `privilege_ext_obj` proto annotation as having no privilege requirement. A new `MilvusService` RPC can therefore be added without the annotation and CI does not fail.

### Expected Behavior

CI should enumerate every `MilvusService` RPC from the compiled descriptor and fail when a non-allowlisted request type lacks `privilege_ext_obj`.

### Steps To Reproduce

1. Add a new `MilvusService` RPC request type without `privilege_ext_obj`, or remove the annotation from a currently protected request.
2. Run the RBAC annotation coverage test.
3. The test should fail and list the uncovered RPC.

### Milvus Log

N/A -- descriptor coverage gap.

### Anything else?

The fail-open behavior is in `internal/proxy/privilege_interceptor.go`; the privilege mapping is stored on request message descriptors via `commonpb.E_PrivilegeExtObj`.

PR #52241, opened 2026-08-06
fix: [RBAC] guard service privilege annotations

shaoting-huang:
## What

Guard the service RBAC surface so that a future RPC added without a `privilege_ext_obj` annotation fails CI instead of silently becoming public (the proxy interceptor treats unannotated requests as public).

## Scope

This PR now focuses on **non-stream hardening** after the stream authentication/RBAC work moved to #52608:

1. **RBAC annotation coverage guard** (`pkg/util/funcutil/rbac_annotation_coverage_test.go`): descriptor-driven test that enumerates `MilvusService` + `ClientTelemetryService` and requires every unannotated RPC to carry a documented allowlist entry. Checks stale entries (annotation landed), stream allowance, comment quality, and reports all buckets in one run. The `CreateReplicateStream` / `DumpMessages` allowlist entries now reference #52608's `streamMethodAuthorizers` as the authorization authority.
2. **Close reviewed bypasses**:
   - `Dummy` query dispatch disabled (was a debug shim that bypassed query authz/rate limit).
   - `RunAnalyzer` now requires collection-scoped `Search` privilege.
   - `GetClientTelemetry` / `PushClientCommand` / `DeleteClientCommand` root-gated; `ClientHeartbeat` stays authenticated telemetry polling.
3. **Harden CDC replication input** (`pkg/streaming/util/message`): `NewReplicateMessage` performs cheap structural validation — nil message, non-panicking message-id parse, reserved properties, header/cipher-header proto decode (cipher-enabled types only) — so malformed frames are rejected before reaching panic-only decoders downstream. The decrypt-and-full-body-unmarshal validation is intentionally dropped: it retried forever against `context.Background()` on revoked KMS keys, wedging the replicate `recvLoop`; the remaining structural checks keep external input away from `Payload()`'s panic contract.

> The stream authentication + RBAC interceptor chain (CreateReplicateStream, DumpMessages) originally in this PR is now in #52608; the client metadata stream interceptor is likewise covered there.

issue: #52236

Signed-off-by: shaoting-huang <shaoting.huang@zilliz.com>


pkg/util/funcutil/rbac_annotation_coverage_test.go
```diff
@@ -0,0 +1,260 @@
...
+	// CalcDistance is a deprecated endpoint that returns service-unavailable before object access.
+	"CalcDistance": "deprecated endpoint rejected before object access",
+	// CheckHealth is the unauthenticated readiness probe used by clients and orchestration.
+	"CheckHealth": "health probe",
+	// ComputePhraseMatchSlop runs text analysis from request literals without object lookup.
+	"ComputePhraseMatchSlop": "stateless text-analysis helper",
+	// Connect is the handshake that establishes client connection metadata before RBAC context exists.
+	"Connect": "client handshake",
+	// CreateReplicateStream is the CDC data-replication stream entrypoint.
+	"CreateReplicateStream": "CDC replication stream",
```
weiliu1031: **[P0] Do not allowlist unauthenticated WAL streams**

The external Proxy installs authentication, privilege, and rate-limit interceptors only in the unary chain, so `CreateReplicateStream` and `DumpMessages` never pass through them. `CreateReplicateStream` accepts caller-supplied frames and appends them to the secondary WAL; `NewReplicateMessage` also reaches `MustUnmarshalMessageID`, so malformed input can panic before the cluster-ID check. `DumpMessages` opens a WAL scanner and returns raw messages. Adding these methods to this allowlist makes the coverage test accept the gap without providing an alternative authorization path. Please add stream authentication and authorization using real cluster credentials rather than the caller-controlled `cluster-id`, and use non-panicking input validation before allowlisting them.

pkg/util/funcutil/rbac_annotation_coverage_test.go
```diff
@@ -0,0 +1,260 @@
...
+	// DescribePrewarmTask only polls the state of an asynchronous prewarm job.
+	"DescribePrewarmTask": "async prewarm status polling",
+	// DescribeSegmentIndexData is an unimplemented feder endpoint that returns service-unavailable.
+	"DescribeSegmentIndexData": "unimplemented feder endpoint",
+	// Dummy is a legacy test/debug shim that dispatches to already-typed request handlers.
+	"Dummy": "legacy debug shim",
```
weiliu1031: **[P1] Dummy bypasses Query authorization**

`DummyRequest` has no privilege annotation, so `PrivilegeInterceptor` returns success before checking any grant. `Proxy.Dummy` then builds a `QueryRequest` from caller-controlled database and collection fields and calls `node.Query` directly; an ordinary Go call does not re-enter the gRPC privilege or rate-limit interceptors. For example, `output_fields: ["count(*)"]` is valid with the empty expression constructed here, allowing an authenticated low-privilege user to execute query workloads against collections they cannot query directly and bypass DQL query quotas, even though the result is discarded. Please remove or restrict this debug endpoint, or explicitly authorize and rate-limit the constructed `QueryRequest` before dispatch.

weiliu1031 (commented):
Two authorization gaps need to be addressed before these allowlist entries are treated as intentional exceptions.

pkg/util/funcutil/rbac_annotation_coverage_test.go
```diff
@@ -0,0 +1,260 @@
...
+	"testing"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+
+	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
+)
+
+const milvusServiceName protoreflect.Name = "MilvusService"
```
liliu-z: The guard hard-codes a single service name, so `ClientTelemetryService` — registered on the same external gRPC server, one line below `MilvusService` — is never checked. All four of its unary RPCs (`ClientHeartbeat`, `GetClientTelemetry`, `PushClientCommand`, `DeleteClientCommand`) lack `privilege_ext_obj`, and `PrivilegeInterceptor` fails open on an unannotated request, so any authenticated caller reaches them with zero grants; `PushClientCommand`/`DeleteClientCommand` are mutating and forward straight to rootcoord. Iterate `File_milvus_proto.Services()` rather than pinning one name, and annotate or explicitly allowlist these four.

pkg/util/funcutil/rbac_annotation_coverage_test.go
```diff
@@ -0,0 +1,260 @@
...
+			stale = append(stale, fmt.Sprintf("%s is not a MilvusService RPC; remove it from rbacAnnotationAllowlist", methodName))
+		}
+	}
+
+	sort.Strings(undocumented)
+	sort.Strings(unallowlistedStreams)
+	sort.Strings(stale)
+	sort.Strings(missing)
+
+	requireNoRBACAnnotationCoverageErrors(t, "allowlist entries without one-line justification", undocumented)
```
liliu-z: `requireNoRBACAnnotationCoverageErrors` uses `t.Fatalf`, which aborts the test goroutine, so only the first non-empty bucket is ever reported — and `missing`, the security-relevant one this guard exists for, is checked last. A developer who adds an unannotated RPC while an unrelated allowlist entry has gone stale sees only the stale-entry message, fixes it, then re-runs to discover the missing annotation. Using `t.Errorf` for the first three checks (or reporting `missing` first) surfaces everything in one run.

pkg/streaming/util/message/builder.go
```diff
@@ -84,13 +106,134 @@ func NewReplicateMessage(clustrID string, im *commonpb.ImmutableMessage) (Replic
...
+		return nil
+	}
+	if !msg.MessageType().CanEnableCipher() {
+		return merr.WrapErrParameterInvalidMsg("message type %s cannot enable cipher", msg.MessageType().String())
+	}
+	return validateReplicateProtoProperty(msg, messageCipherHeader, "cipher header", &messagespb.CipherHeader{})
+}
+
+func validateReplicatePayload(msg *immutableMessageImpl) error {
+	if msg.properties.Exist(messageCipherHeader) {
```
liliu-z: validateReplicatePayload skips body validation for any replicated message that carries a `_ch` property, and validateReplicateCipherHeader above it only checks that the CipherHeader proto decodes and that the type may be encrypted — nothing verifies the payload is really encrypted or decryptable. A replicate client can therefore attach a well-formed CipherHeader to a garbage Insert payload and reach exactly the panic-only decoders this validation was added to guard: the shard interceptor calls materializeFunctionFields, which for a function-enabled collection calls MustBody() -> Payload(), and Payload() panics on both getDecryptorWithRetry failure and Decrypt failure (on builds with no cipher plugin, mustGetCipher panics outright). That crashes the streamingnode and takes down the write path for every collection on it. Worth rejecting a caller-supplied cipher header on the replicate path when no cipher is registered, or validating decryptability before accepting the message.

shaoting-huang: Fixed in 54de95f5bc. `NewReplicateMessage` now treats `_ch` as a signal to validate decryptability instead of skipping payload validation: it rejects nil cipher registration, decryptor lookup failure, nil decryptor, decrypt failure, and decrypted payload unmarshal failure as `ErrParameterInvalid` before returning a message. The decryptor lookup uses the existing retry helper after the explicit nil-cipher guard.

Added regression coverage in `pkg/streaming/util/message/builder_test.go` and `builder_internal_test.go` for the requested garbage encrypted Insert payload plus lookup/decrypt/unmarshal failure cases.

Local validation:
- `(cd pkg && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./streaming/util/message -run "TestNewReplicateMessage")`
- `(cd pkg && go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./streaming/util/message)`
- `make fmt`
- `PKG_CONFIG_PATH=/tmp/eng34-pkgconfig:/home/shaoting/workspace/milvus/internal/core/output/lib/pkgconfig:/home/shaoting/workspace/milvus/cmake_build/thirdparty/milvus-storage make static-check`
- `git diff --check`

pkg/util/funcutil/rbac_annotation_coverage_test.go
shaoting-huang: Fixed. I added server-side stream authentication and privilege checks for `CreateReplicateStream` and `DumpMessages`, mapped both to `UpdateReplicateConfiguration`, and added client stream metadata propagation. I also replaced the panic-prone replicate message path with explicit input validation and regression tests.

pkg/util/funcutil/rbac_annotation_coverage_test.go
shaoting-huang: Fixed. I disabled `Dummy` query dispatch, so it no longer builds and calls `QueryRequest` directly or bypasses query authorization/rate limiting.

pkg/util/funcutil/rbac_annotation_coverage_test.go
shaoting-huang: Fixed. The coverage test now reports all non-empty error buckets with `t.Error` instead of stopping at the first failure, and I added a regression test to verify multiple buckets are surfaced in one run.

pkg/util/funcutil/rbac_annotation_coverage_test.go
shaoting-huang: Fixed. The RBAC coverage test now includes `ClientTelemetryService`. `GetClientTelemetry`, `PushClientCommand`, and `DeleteClientCommand` are guarded as root-only in proxy handler code; `ClientHeartbeat` stays as authenticated heartbeat/command polling and is documented in the allowlist.

shaoting-huang:
Update: the stream authentication + RBAC part of this PR (GrpcStreamAuthInterceptor / StreamPrivilegeInterceptor / client metadata stream interceptor) has been split out into #52608, which now owns the external stream chain (auth + RBAC + fail-closed + rate limit + access log) for CreateReplicateStream and DumpMessages, addressing #52387.

This PR now scopes to the non-stream hardening: the RBAC annotation coverage guard, Dummy/RunAnalyzer/telemetry bypasses, and the CDC replication input hardening. The guard allowlist entries for the two stream methods reference #52608's `streamMethodAuthorizers` as the authority.

pkg/streaming/util/message/builder.go
```diff
@@ -84,13 +106,167 @@ func NewReplicateMessage(clustrID string, im *commonpb.ImmutableMessage) (Replic
...
+	}
+	header := &messagespb.CipherHeader{}
+	if err := DecodeProto(value, header); err != nil {
+		return nil, merr.WrapErrParameterInvalidMsg("invalid cipher header: %s", err.Error())
+	}
+	if cipher == nil {
+		return nil, merr.WrapErrParameterInvalidMsg("invalid cipher header: cipher not registered")
+	}
+
+	decryptor, err := getDecryptorWithRetry(header.EzId, header.CollectionId, header.SafeKey)
```
yhmo: pkg/streaming/util/message/builder.go line:218
Medium ---- decryptReplicatePayloadForValidation runs synchronously inside the proxy replicate recvLoop (handleReplicateMessage -> NewReplicateMessage -> validateReplicatePayload) and calls getDecryptorWithRetry, which on a KMS-invalid error retries with exponential backoff against context.Background() with no deadline or attempt cap. A single replicated message whose cipher key is revoked or mid-rotation therefore blocks the entire recvLoop forever, stalling every subsequent message on the stream. Consider passing a bounded/cancellable context into the validation decrypt (e.g., derived from stream.Context()) so a bad key fails the stream cleanly instead of wedging it.

pkg/util/funcutil/rbac_annotation_coverage_test.go
```diff
@@ -0,0 +1,350 @@
...
+			if allowlisted && (strings.TrimSpace(reason) == "" || strings.Contains(reason, "\n")) {
+				undocumented = append(undocumented, string(methodName))
+			}
+
+			hasAnnotation := hasPrivilegeExtObj(method.Input())
+			methodLabel := fmt.Sprintf("%s.%s (%s)", service.Name(), methodName, method.Input().FullName())
+			switch {
+			case isStreamMethod(method) && !allowlisted:
+				unallowlistedStreams = append(unallowlistedStreams, methodLabel)
+			case hasAnnotation && allowlisted && !isStreamMethod(method):
```
yhmo, line 144: pkg/util/funcutil/rbac_annotation_coverage_test.go line:144
Low ---- The stale-entry check is gated on `!isStreamMethod`, so an allowlisted stream RPC that later gains a privilege_ext_obj annotation on its request type is never flagged as stale (a stream method with annotation and an allowlist entry silently falls through all three cases). This is the opposite of the intended drift signal: after #52608 lands, an annotated stream method would stay in rbacAnnotationAllowlist forever with no test feedback. Consider also flagging allowlisted stream methods whose input now has an annotation, or documenting why stream entries should be treated as permanent.

Issue #52232, opened 2026-08-06
[Bug]: avoid index build task fails for 0 element in nested array index


SpadeA-Tang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version:
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

Avoid index build task fails for 0 element in nested array index

### Steps To Reproduce

```markdown

```

PR #52245, opened 2026-08-06
fix: build empty index when data is empty in nested index

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/52232


czs007 (approved):
/lgtm

Issue #52240, opened 2026-08-06
[Enhancement]: Remove the /expr diagnostic endpoint and expression runtime


XuanYang-cn:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## What would you like to be added?

Remove the Milvus diagnostic expression executor completely.

The removal includes the `/expr` HTTP endpoint, its `/static/` executor page,
the global expression runtime and component-object registry, endpoint-specific
authentication and RBAC behavior, endpoint-specific configuration, and the Go
expression dependency and personal-fork replacement.

After removal, requests to `/expr` and the expression executor under `/static/`
must return HTTP 404. The normal Milvus WebUI and Milvus query/filter expression
features must remain unchanged.

Land the removal on `master` first, then backport the accepted change to the
`3.0` and `2.6` release branches. A release-branch PR must link the master PR.

### Problem Statement

Milvus includes a debug-only HTTP endpoint that compiles and runs expressions
against a global registry containing configuration and live component objects.
It is disabled by default and has authentication controls, but keeping it still
creates an unnecessary runtime-inspection, maintenance, and supply-chain
surface.

The runtime depends on `github.com/expr-lang/expr` through a replacement to
`github.com/SimFG/expr`, a personally governed fork. The same replacement is
present in both Milvus Go modules and in the current `master`, `3.0`, and `2.6`
branches. The fork contains behavior required only by the diagnostic executor,
so removing the executor allows Milvus to remove the dependency instead of
transferring or maintaining the fork.

The endpoint also requires component constructors to register internal runtime
objects globally, adds endpoint-specific configuration and RBAC privilege
surface, and ships a browser executor page. These pieces have no role in normal
vector search, scalar filtering, or query expression evaluation.

### Solution

Remove the diagnostic executor rather than replacing its dependency or adding
more security controls. Milvus should no longer initialize an expression
runtime, expose component objects for reflective inspection, advertise an
`Expr` privilege, or ship endpoint-specific configuration and UI assets.

Existing deployments that still contain the removed configuration keys or a
stored `Expr` privilege grant must continue starting safely. The obsolete
settings and grants should have no effect and must not recreate access to the
removed endpoint.

### User Stories

1. As a Milvus operator, I want the diagnostic executor removed, so that production nodes do not expose an unnecessary code-execution-style inspection surface.
2. As a security reviewer, I want `/expr` to be absent rather than disabled, so that configuration mistakes cannot re-enable it.
[...]

PR #52246, opened 2026-08-06
enhance: remove diagnostic expression executor

XuanYang-cn:
Remove the diagnostic HTTP endpoint, executor UI, runtime registry,
endpoint-specific configuration and RBAC behavior, and forked expression
dependency. Ignore obsolete stored Expr privilege grants during upgrades.

See also: #52240


pkg/util/constant.go
```diff
@@ -515,12 +510,8 @@ func PrivilegeNameForMetastore(name string) string {
 }
 
 func isPrivilegeNameForMetastoreDefined(name string) bool {
-	if _, ok := commonpb.ObjectPrivilege_value[name]; ok {
-		return true
-	}
-	// TODO: drop this special case once PrivilegeExpr is promoted to a proto enum value
-	// in milvus-io/milvus-proto (commonpb.ObjectPrivilege_PrivilegeExpr).
-	return name == PrivilegeExpr
+	_, ok := commonpb.ObjectPrivilege_value[name]
```
liliu-z, line 519: Dropping the `PrivilegeExpr` special case makes `IsPrivilegeNameDefined("Expr")` false, which has two composing effects. First, `CheckIfPrivilegeGroupCreatable` (internal/rootcoord/meta_table.go:2340) no longer rejects `Expr`, so a custom privilege group can now be named `Expr`. Second, `Catalog.RestoreRBAC` (internal/metastore/kv/rootcoord/kv_catalog.go:2162-2168) now falls into its `default` branch for a legacy backup grant named `Expr` and writes it as `PrivilegeGroupNameForMetastore("Expr")` = `PrivilegeGroupExpr` — byte-identical to a grant of a custom group named `Expr`. The `obsoleteExprPrivilege` filters this PR adds to `ListGrant`/`ListPolicy` match `PrivilegeExpr`, not `PrivilegeGroupExpr`, so they do not catch it. Restoring a pre-upgrade backup onto a cluster that has a custom `Expr` group therefore silently grants the role everything that group holds, with no error and no audit trail. Reserving `Expr` as a group name closes this by construction. Note the reviewer's original anchor (rbac_task.go:120) is not touched by this PR; the root cause is here.

internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -1710,6 +1715,8 @@ func (kc *Catalog) ListGrant(ctx context.Context, tenant string, entity *milvusp
 			privilegeName := util.PrivilegeNameForAPI(granteeIDInfos[0])
 			if granteeIDInfos[0] == util.AnyWord {
 				privilegeName = util.AnyWord
+			} else if granteeIDInfos[0] == obsoleteExprPrivilege {
+				continue
```
liliu-z, line 1806: The obsolete-Expr handling is applied only to the read/backup side, not to the restore input side, so the PR's stated goal of ignoring obsolete Expr grants during upgrades is not met for backup/restore. `CheckIfRBACRestorable` (internal/rootcoord/meta_table.go:2279-2281) rejects any grant whose privilege is neither a restorable group nor `IsPrivilegeNameDefined`; since `Expr` is no longer defined, restoring a backup taken from a pre-upgrade cluster where any role held `Expr` now fails outright with `privilege [Expr] does not exist`, aborting the entire RestoreRBAC request rather than skipping the one obsolete grant. Skipping `Expr` in `CheckIfRBACRestorable` and in `Catalog.RestoreRBAC`'s grant loop would make the upgrade path complete.

internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -2244,6 +2253,9 @@ func (kc *Catalog) ListPrivilegeGroups(ctx context.Context) ([]*milvuspb.Privile
 			mlog.Error(ctx, "failed to unmarshal privilege group info", mlog.Err(err))
 			return nil, err
 		}
+		privGroupInfo.Privileges = lo.Reject(privGroupInfo.Privileges, func(privilege *milvuspb.PrivilegeEntity, _ int) bool {
+			return privilege.GetName() == obsoleteExprPrivilegeForAPI
```
liliu-z, line 2352: The new `lo.Reject` strips every privilege entry named `Expr` from every group on read, but nothing reserves `Expr` as a group name any more, and nested groups are explicitly permitted. An admin can create a custom group `Expr` and add it to another group `G` — `CheckIfPrivilegeGroupAlterable` accepts it because `Expr` is a key in `currenctGroups` — after which `ListPrivilegeGroups` silently drops `Expr` from `G`'s privilege list, so roles granted `G` never receive the nested group's privileges. The write succeeds, the read lies, and there is no error or log. Reserving `Expr` at group creation (the fix for the aliasing issue above) also closes this.

internal/metastore/kv/rootcoord/kv_catalog.go
XuanYang-cn: Fixed in 25e04380f5: restore validation accepts legacy `Expr`, and catalog replay skips that obsolete grant instead of persisting it as `PrivilegeGroupExpr`.

pkg/util/constant.go
XuanYang-cn: Fixed in 25e04380f5: `Expr` is reserved for privilege-group creation while remaining non-grantable, preventing restore aliasing with a custom group.

internal/metastore/kv/rootcoord/kv_catalog.go
XuanYang-cn: Fixed in 25e04380f5: `Expr` cannot be created as a custom privilege group, so nested-group reads no longer collide with the obsolete privilege filter.

XuanYang-cn:
Superseded by #52876, which is the current implementation for #52240 and carries the active review and CI state. Closing this PR; please continue review and delivery in #52876.

PR #52247, opened 2026-08-06
fix: build empty index when data is empty in nested index [3.0]

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/52232
pr: https://github.com/milvus-io/milvus/pull/52245


czs007 (approved):
/lgtm

PR #52248, opened 2026-08-06
fix: [RBAC] enforce partition statistics privilege

### LINKED ISSUE #52251: [Bug]: Statistics APIs miss RBAC permission checks

### Environment

- Milvus version: master before milvus-proto `v3.0.0-20260806081414-16b288837fbd`
- Deployment mode(standalone or cluster): standalone or cluster
- MQ type(rocksmq, pulsar or kafka): N/A
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): N/A
- CPU/Memory: N/A
- GPU: N/A
- Others: Authorization/RBAC enabled

### Current Behavior

`GetCollectionStatistics` and `GetPartitionStatistics` requests are not protected by the `PrivilegeGetStatistics` RBAC privilege because the generated proto/API metadata does not include the privilege annotation for these statistics APIs.

As a result, an authenticated principal without `PrivilegeGetStatistics` can pass the privilege interceptor for collection and partition statistics requests.

### Expected Behavior

When authorization is enabled, `GetCollectionStatistics` and `GetPartitionStatistics` should require `PrivilegeGetStatistics` on the target collection. Users without this privilege should receive a permission denied error, and users with the privilege should be allowed.

### Steps To Reproduce

1. Enable authorization.
2. Create or use a non-root user with a role on a collection.
3. Grant the role a collection privilege such as `Load`, but do not grant `GetStatistics`.
4. Call `GetCollectionStatistics` or `GetPartitionStatistics` for that collection.
5. The request is allowed instead of being denied by RBAC.

### Milvus Log

N/A

### Anything else?

Fixed by https://github.com/milvus-io/milvus/pull/52248.

shaoting-huang:
- Bump `milvus-proto/go-api/v3` to `v3.0.0-20260806081414-16b288837fbd` in the root, `pkg`, Go client, and go-client test modules so interceptor-backed `GetPartitionStatistics` uses the `PrivilegeGetStatistics` metadata added by the proto update.
- Add proxy privilege-interceptor regression coverage with an authenticated principal: requests without `PrivilegeGetStatistics` return `codes.PermissionDenied`, while granting `PrivilegeGetStatistics` allows the same collection and partition statistics requests. `GetCollectionStatistics` is covered as baseline parity; `GetPartitionStatistics` is the newly protected request in this bump.
- Refresh generated proxy and Go client mocks for the expanded proto service interface, keep their imports in repository GCI grouping, and merge current master fixes for plan-parser fixture IDs plus fastpb UUID ID decoding/proto-contract goldens.
- Scope note: this PR covers gRPC and REST v2 paths that run `PrivilegeInterceptor`; legacy metrics-port `/api/v1` handlers, unannotated segment-info RPCs, and the Decimal/UUID validation class are pre-existing or separate proto-bump gaps left out of scope.
- Scope note: this proto bump also remaps row-policy RPC privileges: `CreateRowPolicy` and `DropRowPolicy` now require `PrivilegeManageRLS`, and `ListRowPolicies` now requires `PrivilegeViewRLS`. Those RPCs are not implemented in proxy yet, and the new privilege names are not wired into built-in privilege groups in this PR.

related: #52251


go.mod
```diff
@@ -69,7 +69,7 @@ require (
 	github.com/hamba/avro/v2 v2.29.0
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/magiconair/properties v1.8.7
-	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260712212354-7efdf25099dd
+	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260806081414-16b288837fbd
```
liliu-z, line 72: GetPersistentSegmentInfo and GetQuerySegmentInfo carry no privilege_ext_obj annotation in either the old or the new pinned proto, and PrivilegeInterceptor treats a missing annotation as allow, so both remain callable by any authenticated principal with zero grants. Both return PartitionID and NumRows per segment, which is a superset of what GetPartitionStatistics exposes, so the row-count and partition-layout data this PR just gated on PrivilegeGetStatistics is still reachable through an unguarded path. This is pre-existing rather than a regression, but it means the fix is incomplete for the data it is meant to protect — annotating these two request messages in the same proto bump would close it.

pkg/go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/jolestar/go-commons-pool/v2 v2.1.2
 	github.com/json-iterator/go v1.1.13-0.20220915233716-71ac16282d12
 	github.com/klauspost/compress v1.18.0
-	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260712212354-7efdf25099dd
+	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260806081414-16b288837fbd
```
liliu-z, line 24: The proto version was bumped in go.mod and pkg/go.mod but not in tests/go_client/go.mod, which still pins v3.0.0-20260712212354-7efdf25099dd while replacing pkg with ../../pkg — MVS therefore wants to upgrade and module resolution now fails under -mod=readonly. I measured `go list -mod=readonly -m all` in tests/go_client: rc=0 at the parent commit, rc=1 at HEAD with "updates to go.mod needed, disabled by -mod=readonly". Makefile static-check runs golangci-lint inside tests/go_client, so this breaks CI. client/go.mod is also still on the old pin but resolves cleanly (rc=0) since it does not replace pkg; bump it anyway for consistency, since scripts/update-api-version.sh is the documented procedure and updates all four go.mod files.

go.mod
liliu-z, line 72: The legacy /api/v1 REST surface calls proxy methods directly and never invokes PrivilegeInterceptor, so the grants this PR relies on are not consulted there. Note the scope is wider and less specific than "statistics": every handler on that router bypasses the privilege check (search, query, insert, delete, credential management included), the router is mounted on the metrics port rather than the main client port, and the bypass predates this PR — GetCollectionStatistics was already annotated and already bypassed on this route. Worth recording as a known gap in the PR's coverage claim rather than treating it as something this change introduced or can fix.

liliu-z, line 72: The PR body says it protects GetCollectionStatistics and GetPartitionStatistics, but GetCollectionStatisticsRequest already carried PrivilegeGetStatistics in the previously pinned proto — only GetPartitionStatistics actually gains enforcement here. The collection half of the new test therefore asserts behavior that already held before the change. Worth restating the body as "GetPartitionStatistics now requires PrivilegeGetStatistics" and noting the upgrade impact: roles assembled from individual grants that omit GetStatistics will start getting PermissionDenied on that call, though roles granted via the built-in ReadOnly/ReadWrite/Admin groups are unaffected since those already include it.

liliu-z, line 72: The bumped proto adds Decimal=30 and UUID=31 to schemapb.DataType, and since field-type validation gates on generated-enum membership rather than a supported-type list, those two types now pass CreateCollection and AddField and fail later at insert with "unsupported data type". Real but minor and not new in kind: Date=28 and Time=29 were already in the enum before this bump and have zero references anywhere in the repo, so the same accept-then-fail-at-insert hole already shipped. The failure is a clean ErrParameterInvalid, not corruption. If you want to close the class, replace the enum-membership predicate with an explicit whitelist at all four sites; otherwise this is fine to leave.

liliu-z, line 72: This bump turns `internal/parser/planparserv2` from green to a panicking FAIL, but the root cause is the test fixture, not the parser. The new proto adds `DataType` enum values `Decimal = 30` and `UUID = 31`; `newTestSchema` assigns `FieldID: 100 + int64(value)` while ranging over `schemapb.DataType_value`, so `DecimalField` gets FieldID 130 and collides with the hard-coded dynamic field also at FieldID 130. `CreateSchemaHelper` then returns `duplicated fieldID: 130`, `ParseExpr` returns a nil expr, and `ShowExpr(nil)` dereferences it and SIGSEGVs, aborting the whole package before any other test runs. The fix is to renumber the fixture's hard-coded IDs (130-134 in plan_parser_v2_test.go) out of the `100+enum` range, not to change the parser. Note the '43 failing tests' figure is not reproducible: the panic kills the test binary, so `go test` emits zero `--- FAIL` lines and only a package-level FAIL.

pkg/go.mod
liliu-z, line 24: This bump turns `pkg/util/fastpb` red: `TestProtoContract_FieldSetsPinned` trips on three types. `ScalarField` gains field 17 `valid_data` and `VectorField` gains field 9 `valid_data` -- both plain repeated fields, which that test's own guidance calls correct-by-construction because the decoders route unknown numbers through `default -> rest -> protoMerge`; those two only need the golden set updated. `schemapb.IDs` gains field 3 `uuid_id`, and that one is a new oneof variant, which the same guidance says MUST be handled in-pass. `dec.ids` handles only numbers 1 and 2 in-pass and drops field 3 into the deferred `protoMerge`, so the value is still decoded correctly but oneof last-wins ordering diverges from the official codec whenever more than one arm appears on the wire. Implement in-pass `uuid_id` decoding, then update the golden set. Two corrections to the original claim: the cited path `pkg/util/fastpb/fastpb_test.go:100` does not exist (the test is `proto_contract_test.go`), and UUID ids are not 'silently skipped or misparsed' -- the fallback decodes them; the PR also does not touch fastpb at all, so nothing has been re-pinned.

go.mod
liliu-z, line 72: The scope note claims `GetPartitionStatistics` is the only newly protected request in this bump, but the bump also silently re-maps the privilege required by three existing row-policy RPCs: `CreateRowPolicy` and `DropRowPolicy` move from `PrivilegeCreateOwnership`/`PrivilegeDropOwnership` to `PrivilegeManageRLS`, and `ListRowPolicies` moves from `PrivilegeSelectOwnership` to `PrivilegeViewRLS`. Both new privilege names exist only in the proto enum; they appear nowhere in `pkg/util/constant.go`, so they are in no built-in privilege group (`ClusterAdmin` carries CreateOwnership/DropOwnership, `ClusterReadOnly` carries SelectOwnership) and are absent from `util.ObjectPrivileges`, which `Core.isValidPrivilege` requires for v1 `OperatePrivilege` grants. Nothing breaks today because those three RPCs are unimplemented at the proxy and fall through to the `UnimplementedMilvusServiceServer` embed, but the gap will surface the moment RLS lands. Worth calling out in the PR body rather than leaving a security PR's scope statement inaccurate.

shaoting-huang: Thanks, I documented this as a known out-of-scope gap in the PR body. This PR keeps the fix limited to the statistics annotation bump; segment-info annotations should be handled in a separate proto change.

pkg/go.mod
shaoting-huang: Fixed. I bumped the proto version in root, pkg, client, and tests/go_client, then verified readonly module resolution for tests/go_client.

go.mod
shaoting-huang: Thanks, I added a scope note for the legacy metrics-port REST surface. This PR covers gRPC and REST v2 paths that go through PrivilegeInterceptor; the legacy direct-call routes are a pre-existing gap.

shaoting-huang: Thanks, I documented this as an existing validation-class gap and left it out of scope for this RBAC fix. Closing it should use an explicit supported-type whitelist separately.

shaoting-huang: Updated the PR body. It now states that GetPartitionStatistics is the newly protected request, while GetCollectionStatistics is covered only as baseline parity.

pkg/go.mod
shaoting-huang: Fixed by merging the upstream fastpb UUID ID decoding change and updating the proto-contract goldens for IDs, ScalarField.valid_data, and VectorField.valid_data.

go.mod
shaoting-huang: Fixed by merging the upstream fixture hardening. The hand-written test FieldIDs now start at 1000, away from the generated 100 + DataType enum range.

shaoting-huang: Updated the PR body with a scope note for the row-policy privilege remap. The RPCs are still unimplemented in proxy, and wiring the new RLS privileges into built-in groups is left out of this PR.

PR #52250, opened 2026-08-06
enhance: add PLE ruff rule

even-even:
Hello
In PR:
1) add https://docs.astral.sh/ruff/rules/#error-ple ruff rule
2) fixes in code

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -15,7 +15,7 @@
 #   BASE_REF      diff base ref. Auto-detected from the current branch's open PR
 #                 via `gh pr view`. If no PR exists, you MUST set it explicitly,
 #                 e.g.: make ci BASE_REF=origin/master
-#   RUFF_VERSION  ruff version (default: 0.15.11, pinned to .github/workflows/python-lint.yaml)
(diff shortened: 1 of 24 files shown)


tests/Makefile
```diff
@@ -20,7 +20,7 @@
 # Requires `uvx` (ships with `uv`). Ruff is fetched & cached on first run.
 # Auto-detection requires `gh` authenticated (`gh auth status`).
 
-RUFF_VERSION ?= 0.15.11
+RUFF_VERSION ?= 0.16.1
```
yhmo, line 23: tests/Makefile line:23
Medium ---- This updates the local Ruff runner to 0.16.1, but `.github/workflows/python-lint.yaml` still pins all three CI invocations to 0.15.11. The Makefile comment and `tests/README.md` promise that these versions match, so `make ci` and pre-commit can now accept or rewrite code differently from CI. Please update the workflow pins (and the stale line-18 comment) to 0.16.1, or keep all entry points on 0.15.11.

even-even: thanks for review
I unify ruff version to 0.16.1

yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #52138, opened 2026-08-03
enhance: merge ARRAY contains predicates

sunby:
issue: #51996

## What

- Merge compatible OR chains on the same physical ARRAY column into one ContainsAny predicate.
- Merge compatible AND chains on the same physical ARRAY column into one ContainsAll predicate.
- Preserve first-encounter order, remove duplicate values, and support nested or balanced logical trees and existing Any or All predicates.
- Keep JSON columns and incompatible values unchanged.
- Add rewriter unit and integration coverage, query-level equivalence coverage, and documentation.

## Testing

- LOCALSTORAGEPATH=/tmp/milvus-codex-array-contains go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2/...
- LOCALSTORAGEPATH=/tmp/milvus-codex-array-contains bin/golangci-lint run --build-tags dynamic,test --timeout=10m --config .golangci.yml ./internal/parser/planparserv2/rewriter
- Python AST syntax check for tests/python_client/milvus_client/test_milvus_client_query.py
- cd internal/core && CLANG_FORMAT=/opt/homebrew/opt/llvm@15/bin/clang-format ./run_clang_format.sh .

## Notes

- JSON columns are intentionally excluded from this optimization.
- The query-level growing and sealed/indexed equivalence case was added but not executed locally because no usable Milvus endpoint was available.
- make lint-fix was invoked; its repository-wide golangci phase was blocked by stale local internal/core/output headers, while the scoped rewriter lint completed with zero issues.

PR #52252, opened 2026-08-06
enhance: refactor expression rewriter rules

sunby:
## What

- Move concrete expression rewrite logic into the `rewriter/rules` package, with one rule per file and explicit `Match` and `Apply` stages.
- Replace the legacy visitor orchestration with a child-first explicit post-order traversal.
- Traverse a replacement subtree again after a rule applies, so chained rules complete before the waiting parent is processed.
- Build the active rule registry from configuration while keeping Term sorting and deduplication always enabled.
- Preserve existing public helper APIs through compatibility wrappers.
- Add regression coverage for rule matching and ordering, chained rewrites, child-before-parent behavior, nil replacements, and non-terminating-rule guards.
- Move the ARRAY contains implementation already merged by #52138 into the rules package and preserve its original OR/AND pipeline order.

The ARRAY_CONTAINS optimization behavior remains owned by #52138; this PR only adapts that now-master code to the rule abstraction.

## Testing

- `env LOCAL_STORAGE_SIZE=1 go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/parser/planparserv2/...`
- CI-equivalent changed-file Go format check
- targeted `golangci-lint --build-tags dynamic,test` for `./internal/parser/planparserv2/rewriter/...`: 0 issues
- `internal/core/run_clang_format.sh internal/core` with clang-format 15
- `git diff --check`

## Notes

- `make lint-fix` completed gofumpt and gci, but its repository-wide golangci phase is blocked in the untouched `internal/util/analyzecgowrapper` package by a local CGo header mismatch for the `Analyze` function signature.
- Preserving first-occurrence output order for independent groups in Term/Range/TextMatch rewrites remains a follow-up TODO.


internal/parser/planparserv2/rewriter/rewriter.go
```diff
@@ -0,0 +1,121 @@
...
+
+			ruleApplications++
+			if ruleApplications >= maxRuleApplications {
+				r.hitApplicationLimit = true
+				return result
+			}
+
+			// The replacement may have a different shape and may expose new rule
+			// inputs in its children, so process the replacement post-order again.
+			stack = append(stack, rewriteTask{slot: task.slot})
```
buqian-zilliz, line 92: **[P3] Always-on term sorting is executed twice for every multi-value IN/NOT IN list**

`sortTermValuesRule` is always registered and mutates the `TermExpr` in place, returning `changed=true` the first time it sorts an unsorted value list. `rewriter.rewrite` then treats that as a replacement and re-traverses the same node (`stack = append(stack, rewriteTask{slot: task.slot})`), so `Apply` runs again and `sortTermValues` re-sorts the already-sorted list. Each `sortTermValues` call also makes a full slice copy (`before := append([]*planpb.GenericValue(nil), term.GetValues()...)`) and `sortGenericValues` allocates a new deduplicated slice via `lo.UniqBy`, so the always-on sort work is roughly doubled for large `IN`/`NOT IN` value lists in the per-query planning path. Consider a mechanism that continues the rule scan for same-node in-place mutations, or a `sortTermValuesRule` that only reports a change when the node shape (and therefore the next applicable rule) can actually differ.

sunby: not essential for now.

internal/parser/planparserv2/rewriter/rules/rule_canonicalize_term.go
```diff
@@ -0,0 +1,71 @@
...
+				return newAlwaysTrueExpr(), true
+			}
+			if len(values) == 1 {
+				return newUnaryRangeExpr(term.GetColumnInfo(), planpb.OpType_Equal, values[0]), true
+			}
+		}
+	}
+
+	if len(term.GetValues()) == 1 {
+		return newUnaryRangeExpr(term.GetColumnInfo(), planpb.OpType_Equal, term.GetValues()[0]), true
```
buqian-zilliz, line 43: **[P2] Single-value NOT IN now emits NOT(col == value) for canRewriteNotEqual-guarded columns**

The refactor moves single-value `IN -> ==` normalization into `canonicalizeTermRule`, which runs before `simplifyUnaryRule`. Previously `visitUnaryExpr` handled `NOT(TermExpr)` up front and, when `canRewriteNotEqual(col, v)` was false, returned `NOT(TermExpr)` without converting the inner term. Now `canonicalizeTermRule.Apply` unconditionally rewrites a one-element `TermExpr` to `col == v`, and `simplifyUnaryRule.Match` then rejects that `Equal` child because `canRewriteNotEqual` is false, so the emitted plan becomes `NOT(col == v)`. Under three-valued logic no NULL is folded, so this is likely semantically equivalent, but it changes the plan shape specifically for the non-JSON nested-path columns that `canRewriteNotEqual` is deliberately conservative about. Please confirm this is intended, or apply the same `canRewriteNotEqual`/nested-path guard to the single-value `IN -> ==` conversion in the `NOT` context.

internal/parser/planparserv2/rewriter/rewriter.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (r *rewriter) rewrite(expr *planpb.Expr) *planpb.Expr {
+	if expr == nil {
+		return nil
+	}
+
+	r.hitApplicationLimit = false
+	maxRuleApplications := r.maxRuleApplications
+	if maxRuleApplications <= 0 {
+		maxRuleApplications = defaultMaxRuleApplicationsPerInputNode * countRewriteNodes(expr)
```
buqian-zilliz, line 42: **[P3] Defensive rewrite budget adds a full extra traversal and can fail silently**

`rewrite` calls `countRewriteNodes(expr)` on every invocation to size the defensive budget, which walks the whole tree once before the main traversal walks it again, adding an O(n) pass to the per-query planning path. When the budget is exhausted the function returns `result` after setting `hitApplicationLimit`, but that field is only read in tests, so production callers receive a partially rewritten (though still semantically equivalent) tree with no signal that optimization was truncated. Consider a fixed/static budget, counting nodes during the main traversal, or surfacing the limit flag.

internal/parser/planparserv2/rewriter/api.go
```diff
@@ -0,0 +1,21 @@
+package rewriter
+
+import (
```
zhengbuqian, line 3: what is this file for? I did not see any usage?

internal/parser/planparserv2/rewriter/README.md
```diff
@@ -19,15 +19,31 @@ The rewriter can be configured via the following parameter (refreshable at runti
...
+- A rewrite-application budget is only a defensive guard against mutually inverse or continually expanding custom rules. Normal rules should terminate without reaching it.
+
 ### Implemented Rules
 
-1) IN / NOT IN normalization and merges (`term_in.go`)
+1) IN / NOT IN normalization and merges (`rules/term_in.go`)
 - OR-equals to IN (same column):
   - `a == v1 OR a == v2 ...` → `a IN (v1, v2, ...)`
-  - Numeric columns only merge when count > threshold (default 150); others when count > 1.
+  - At least two compatible equality predicates are required.
```
zhengbuqian, line 43: do not remove the raw explanation message?

internal/parser/planparserv2/rewriter/rules/util.go
```diff
@@ -88,11 +89,21 @@ func isNumericType(dt schemapb.DataType) bool {
...
-		return
+		return false
 	}
+	before := append([]*planpb.GenericValue(nil), term.GetValues()...)
 	term.Values = sortGenericValues(term.Values)
+	if len(before) != len(term.GetValues()) {
+		return true
+	}
+	for i := range before {
+		if !proto.Equal(before[i], term.GetValues()[i]) {
```
zhengbuqian, line 102: no need to use proto equal. ptr comparison should be sufficient?

internal/parser/planparserv2/rewriter/rules/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package rules defines expression rewrite rules, their default ordering, and
+// the shared transformations used by those rules.
+package rules
```
zhengbuqian, line 3: why is this file necessary?

internal/parser/planparserv2/rewriter/rules/array.go
```diff
@@ -1,10 +1,16 @@
...
+package rules
 
 import (
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/pkg/v3/proto/planpb"
 )
 
+// NormalizeEmptyArrayComparisons lowers whole-array empty comparisons to
+// executable array-length predicates.
+func NormalizeEmptyArrayComparisons(expr *planpb.Expr) *planpb.Expr {
```
zhengbuqian, line 10: we can simply make normalizeEmptyArrayComparisons public

internal/parser/planparserv2/rewriter/rules/json_term.go
```diff
@@ -7,6 +7,11 @@ import (
 
 var jsonTermKindOrder = []string{"bool", "int64", "float", "string", "array"}
 
+// NormalizeTermExprs makes every TermExpr executable by a scalar executor.
+func NormalizeTermExprs(expr *planpb.Expr) *planpb.Expr {
```
zhengbuqian, line 11: simply make normalizeTermExprs public

PR #52253, opened 2026-08-06
enhance: extend composite Update coverage across coordinator catalogs

shaoting-huang:
issue: #51351
**Stacked on #52214** — the first 3 commits (`support mixed exact and prefix removals` / `drop role and its grants` / `release collection load info`) belong to that PR; review this one from `enhance: persist collection and partition load info atomically` onward (8 commits).

## What

#51624 introduced the composite `Update(actions...)` primitive but migrated only 7 call sites. This PR extends coverage to every remaining "logically one change, persisted as several independent writes" site across the coordinators, so each lands in one composite transaction (atomic when it fits the txn limit; commit-marker-last on the chunked fallback):

1. **querycoord collection load** — load-info + partition load-infos in one txn (was: two steps; a crash left a collection loaded with zero partitions).
2. **rootcoord rename** — collection record + db pointer + grant-key rewrite in one txn (was: three sequential writes).
3. **rootcoord drop collection** — collection drop + its grant cleanup in one txn (was: two orchestrated calls, orphaned grants on crash).
4. **rootcoord grant mutations** — AlterGrant's three modes (grant / revoke / legacy-id migration) each single-txn; RestoreRBAC batched composite (per-batch atomic, oversized single entities split).
5. **datacoord import** — job + first task batch in one txn, job record as commit marker (the intentional per-vchannel 2PC in `HandleCommitVchannel` is untouched).
6. **datacoord compaction completion** — clustering's two-part write unified; sort / bump-schema-version paths moved onto the same composite encoding (byte-for-byte identical to legacy `AlterSegments`).
7. **streamingcoord** — the whole catalog moved onto the shared txn engine (replicate-configuration composite with config as marker); streaming mocks generated.
8. **cleanup** — unreachable legacy methods removed (`MetaTable.DropGrant`/`IMetaTable.DropGrant` and friends); `ActionUpdate` doc corrected to its actual replace/upsert semantics (as discussed in #51624 review).

`DropExternalCollectionRefreshJob/Task` are intentionally left in place: #51892 is rewriting that subsystem; this PR does not touch external-collection-refresh files.

## Crash-safety hardening found in adversarial review

- **Import ambiguous-commit rollback (data-loss class)**: the initial import composite rolled back task records whenever `catalog.Update` returned an error — but on an applied-yet-timed-out etcd txn the job marker was already durable, and the rollback deleted task records a committed job referenced; after a restart the checker would drive a task-less job to "completed" without importing anything. Fixed with a read-back disambiguation (`readBackJobMarker`): marker committed → adopt as success; marker provably absent → roll back; read-back failed → leave for restart adoption. Regression tests inject the ambiguous failure both ways.
- **RestoreRBAC oversized entity**: a single entity exceeding `MaxTxnOps` (e.g. a user with 64+ role mappings) is now split instead of producing a rejected oversized txn.
- **Transient-error handling in import checker**: marker-only write failures retry on the next tick instead of permanently failing the job.

## Testing

TDD throughout (each commit carries tests written first and observed failing). Test bar follows #51624: byte-for-byte encoding equivalence against the legacy paths, ordering, atomic + chunked-fallback crash injection (including second-chunk crashes), idempotent retry, ambiguous-failure injection. Full `internal/metastore/...`, `internal/kv/...`, `internal/querycoordv2/meta/...`, `internal/rootcoord/...`, `internal/datacoord/...` suites pass locally.

### DIFF
--- a/internal/datacoord/import_checker.go
+++ b/internal/datacoord/import_checker.go
@@ -256,30 +256,32 @@ func (c *importChecker) getLackFilesForImports(job ImportJob) []*datapb.ImportFi
 
 func (c *importChecker) checkPendingJob(job ImportJob) {
 	log := mlog.With(mlog.FieldJobID(job.GetJobID()))
+	// len(lacks) == 0 is NOT an early exit: after a crash between the task
+	// flush and the job commit marker (the chunked-fallback window of
+	// AddTasksToJob), every task already exists while the job is still
+	// Pending, and the retry must still land the PreImporting marker below.
 	lacks := c.getLackFilesForPreImports(job)
-	if len(lacks) == 0 {
-		return
-	}
 	fileGroups := lo.Chunk(lacks, Params.DataCoordCfg.FilesPerPreImportTask.GetAsInt())
 
-	newTasks, err := NewPreImportTasks(fileGroups, job, c.alloc, c.importMeta)
+	var newTasks []ImportTask
+	if len(fileGroups) > 0 {
+		var err error
+		newTasks, err = NewPreImportTasks(fileGroups, job, c.alloc, c.importMeta)
+		if err != nil {
+			log.Warn(c.ctx, "new preimport tasks failed", mlog.Err(err))
+			return
+		}
+	}
+	// The task batch and the job transition persist as one composite write,
+	// job record last as the commit marker, so a crash cannot leave task
+	// records under a job that never observed them.
+	err := c.importMeta.AddTasksToJob(c.ctx, job.GetJobID(), newTasks, UpdateJobState(internalpb.ImportJobState_PreImporting))
 	if err != nil {
-		log.Warn(c.ctx, "new preimport tasks failed", mlog.Err(err))
+		log.Warn(c.ctx, "add preimport tasks failed", mlog.Err(err))
 		return
 	}
 	for _, t := range newTasks {
-		err = c.importMeta.AddTask(c.ctx, t)
-		if err != nil {
-			log.Warn(c.ctx, "add preimport task failed", WrapTaskLog(t, mlog.Err(err))...)
-			return
-		}
 		log.Info(c.ctx, "add new preimport task", WrapTaskLog(t, mlog.Any("fileStats", t.GetFileStats()))...)
-	}
-
-	err = c.importMeta.UpdateJob(c.ctx, job.GetJobID(), UpdateJobState(internalpb.ImportJobState_PreImporting))
-	if err != nil {
-		log.Warn(c.ctx, "failed to update job state to PreImporting", mlog.Err(err))
-		return
 	}
 	pendingDuration := job.GetTR().RecordSpan()
 	metrics.ImportJobLatency.WithLabelValues(metrics.ImportStagePending).Observe(float64(pendingDuration.Milliseconds()))
@@ -326,10 +328,11 @@ func (c *importChecker) checkPreImportingJob(job ImportJob) {
 		return
 	}
 
+	// len(lacks) == 0 is NOT an early exit: after a crash between the task
+	// flush and the job commit marker (the chunked-fallback window of
+	// AddTasksToJob), every task already exists while the job is still
+	// PreImporting, and the retry must still land the Importing marker below.
 	lacks := c.getLackFilesForImports(job)
-	if len(lacks) == 0 {
-		return
-	}
 
 	requestSize, err := CheckDiskQuota(c.ctx, job, c.meta, c.importMeta)
 	if err != nil {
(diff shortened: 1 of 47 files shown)


internal/datacoord/import_checker.go
```diff
@@ -340,22 +343,44 @@ func (c *importChecker) checkPreImportingJob(job ImportJob) {
...
+			// failed transition-only write as retriable (warn, next tick
+			// replays it), and a transient store error must not kill the job
+			// - keep that semantic.
+			log.Warn(c.ctx, "update import job state to Importing failed, wait for retry", mlog.Err(err))
+			return
+		}
+		// Task creation failed (the batch was rolled back); mirror the legacy
+		// AddTask-failure semantic and fail the job.
+		log.Warn(c.ctx, "add new import tasks failed", mlog.Err(err))
+		updateJobState(internalpb.ImportJobState_Failed, UpdateJobReason(err.Error()))
```
liliu-z, line 372: AddTasksToJob returns an error for two very different outcomes and the caller cannot tell them apart. `jobMarkerAbsent` means the write provably missed and the task records were rolled back; `jobMarkerUnknown` means the read-back itself failed, so the composite may well have committed — which is why import_meta.go:348-357 deliberately skips the rollback and leaves the records "for restart reload". But whenever `len(newTasks) > 0` this branch terminalizes the job as Failed, and UpdateJob refuses any transition out of Failed (import_meta.go:180-184), so the restart-adoption path the unknown branch exists to preserve can never run. Note this only preserves the legacy AddTask-failure semantic (it is not a regression) and checkGC does eventually reclaim a Failed job, so the harm is a spuriously and irreversibly failed import rather than unreclaimed data. Surfacing the marker state (or a typed error) would let the unknown case retry on the next tick like the marker-only branch just above.

internal/datacoord/import_meta.go
```diff
@@ -256,6 +260,137 @@ func (m *importMeta) AddTask(ctx context.Context, task ImportTask) error {
...
+// concurrent task or job write can interleave. Terminal jobs (Completed/Failed)
+// reject the batch without a write, mirroring UpdateJob's guard.
+func (m *importMeta) AddTasksToJob(ctx context.Context, jobID int64, tasks []ImportTask, actions ...UpdateJobAction) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	job, ok := m.jobs[jobID]
+	if !ok {
+		return nil
+	}
+	if job.GetState() == internalpb.ImportJobState_Completed ||
```
liliu-z, line 302: The terminal-state guard here (and the `!ok` guard just above it) returns nil, so the caller cannot distinguish "batch persisted" from "batch silently discarded". checkPreImportingJob goes on to log `add new import task` for tasks that were never written and to emit the ImportStagePreImport latency metric plus the "preimport done" log; checkPendingJob does the same with the pending-stage metric. More concretely, NewImportTasks has already persisted every segment of the batch in `Importing` state before this call, and importInspector.processFailed — the only path that ever marks import segments Dropped — iterates task records, so a discarded batch leaves those segments with nothing to reclaim them. Returning an error would at least surface the discard to the caller.

internal/datacoord/import_meta.go
```diff
@@ -256,6 +260,137 @@ func (m *importMeta) AddTask(ctx context.Context, task ImportTask) error {
...
+			// write actually landed. Adopt it exactly as a success - rolling
+			// back now would delete task records a committed job references.
+			mlog.Warn(ctx, "composite import write reported an error but its job marker is committed, adopting",
+				mlog.FieldJobID(jobID), mlog.Err(err))
+		case jobMarkerAbsent:
+			// The marker is definitively not on disk, so no job observed the
+			// batch. Roll back any task record the chunked fallback already
+			// flushed: the retry re-allocates task IDs, so these keys would
+			// never be written - or dropped - again by this process.
+			for _, task := range tasks {
```
liliu-z, line 334: The rollback deletes the flushed task records but leaves the segments NewImportTasks already persisted in the `Importing` state. importInspector.processFailed is the only path that marks import segments Dropped, and it iterates task records, so deleting the record removes the one handle a restart reload would have used to retire those segments. The common all-or-nothing failure leaks the same segments today (legacy AddTask failed the job with the suffix of segments already allocated), so this is a hardening gap rather than a regression — but since this PR is introducing an explicit rollback, dropping the preallocated segments alongside the task records would close it.

internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -2135,15 +2270,105 @@ func (kc *Catalog) BackupRBAC(ctx context.Context, tenant string) (*milvuspb.RBA
...
+	// delete twice into one txn.
+	//
+	// An entity whose kvs alone exceed the limit cannot be written atomically
+	// at all - one oversized txn is simply rejected by the store, failing the
+	// restore permanently where the legacy per-key path succeeded. Its kvs are
+	// therefore split across batches, saves (in deterministic key order)
+	// before removals, giving up per-entity atomicity for the same
+	// convergence-on-retry guarantee the restore as a whole already relies on.
+	stage := func(addSaves map[string]string, addRemovals []string, keepExisting bool) error {
+		if n := len(saves) + len(removals); n > 0 && n+len(addSaves)+len(addRemovals) > limit {
```
liliu-z, line 2321: `keepExisting` is evaluated against `saves`, which `flush()` resets, so the first-wins rule only holds inside one batch: when `stage` flushes before adding, a later duplicate grant lands in the next txn and overwrites the earlier one, making the outcome depend on MaxTxnOps. Two corrections to the framing though. The pre-PR path had no first-grant-wins semantic to break — `AlterGrant` returned the `IgnorableError` for an already-granted privilege and `RestoreRBAC` propagated it, aborting the whole restore; skipping duplicates is a deliberate new behavior in this PR. And the value that actually diverges is the grantee-id leaf's grantor username, not the privilege set, so the effective grants are identical either way. Still worth hoisting the seen-key set outside the batch loop so the result is batch-size independent.

Issue #52255, opened 2026-08-06
[Bug]: REST VARCHAR field stores a rewritten number instead of the literal, and silently accepts objects


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (also present on 2.6 / 3.0)
- Deployment mode(standalone or cluster): any
- SDK version: REST API v1 / v2 (no SDK)
```

### Current Behavior

The VARCHAR / STRING branch of `checkAndSetData`
(`internal/distributed/proxy/httpserver/utils.go`) stores
`fieldValue.String()`. That is a rendering, not the token the caller sent:
gjson formats a number by parsing it into a `float64` and printing it with the
`'f'` verb, which never uses exponent notation.

| request | stored | note |
|---|---|---|
| `{"name": 1.50}` | `1.5` | trailing zero dropped |
| `{"name": 1.0}` | `1` | decimal point dropped |
| `{"name": 9007199254740993.0}` | `9007199254740992` | last digit lost |
| `{"name": 1e19}` | `10000000000000000000` | rewritten |
| `{"name": 1e300}` | 301 bytes of digits | 5 bytes in, 301 bytes stored |
| `{"name": {"a": 1}}` | `{"a":1}` | object silently stringified |

Measured expansion for the exponent case:

```
1e19   ( 4 bytes) ->  20 bytes stored
1e100  ( 5 bytes) -> 101 bytes stored
1e300  ( 5 bytes) -> 301 bytes stored
1e308  ( 5 bytes) -> 309 bytes stored
```

It is bounded by the float64 range (~309 bytes), so it is not a
denial-of-service, but a field whose `max_length` is smaller than the expansion
fails with a size the caller never sent, which is very hard to diagnose from
the client side.

The object case is worse in a different way: addressing the wrong field is a
common client mistake, and storing the object's text hides it instead of
reporting it.

### Expected Behavior

Take the literal the caller wrote:

- string -> the decoded string
- number, boolean -> `value.Raw`
- object, array -> reject; there is no meaningful string form

### Steps To Reproduce

```markdown
1. Create a collection with a VARCHAR field `name` (max_length 256).
2. POST /v2/vectordb/entities/insert with `{"name": 1e300}`.
3. The request fails with a max_length error naming 301 characters, although
   the request carried 5.
4. Insert `{"name": {"a": 1}}`; it succeeds and stores the text `{"a":1}`.
```

### Anything else?

[...]

PR #52256, opened 2026-08-06
fix: store the literal a REST caller wrote in a VARCHAR field

xiaofan-luan:
## What changed

- Build a VARCHAR / STRING value from the original token: `value.Raw` for
  numbers and booleans, the decoded string for strings.
- Reject a JSON object or array for a string field.
- Add coverage for every JSON value kind, for the exponent expansion, and for
  the rejected structures.

## Root cause

The branch stored `fieldValue.String()`, which is a rendering rather than the
token. gjson formats a number by parsing it into `float64` and printing it with
the `'f'` verb, which never uses exponent notation:

|---|---|---|
| `{"name": 1.50}` | `1.5` | `1.50` |
| `{"name": 1.0}` | `1` | `1.0` |
| `{"name": 9007199254740993.0}` | `9007199254740992` | `9007199254740993.0` |
| `{"name": 1e19}` | `10000000000000000000` | `1e19` |
| `{"name": 1e300}` | 301 bytes of digits | `1e300` |
| `{"name": {"a": 1}}` | `{"a":1}` | rejected |

Measured expansion: `1e19` -> 20 bytes, `1e100` -> 101, `1e300` -> 301,
`1e308` -> 309. Bounded by the float64 range, so not a denial-of-service, but a
field whose `max_length` is below the expansion fails with a size the caller
never sent.

The object case hides a different problem: addressing the wrong field is a
common client mistake, and storing the object's text swallows it.

## Compatibility

This changes what is stored for inputs that already worked, so it is worth
calling out explicitly:

- **A number in a string field keeps its literal form.** `{"sku": 12345}` is
  unchanged, since a plain integer renders to itself. Inputs written with a
  decimal point or an exponent change: `1.0` was stored as `1` and is now
  stored as `1.0`. A client that round-trips through this field and compares
  strings will see the difference.
- **An object or an array is now rejected** instead of stored as its text.
- Strings, plain integers, negatives and booleans are untouched.

Geometry and Timestamptz share the code path and are deliberately left alone:
`ConvertWKTToWKB` and `ParseTimeTz` already reject anything that does not
parse, so those two never store a silently mangled value.

## Validation

- `gofmt` clean.
- The expected values were cross-checked against the `gjson` v1.17.3 this repo
  pins; all 14 cases match and the pre-change behavior reproduces every row of
  the table above.
- The `httpserver` package cannot be compiled locally against the available
  native artifacts (they predate `SegcoreSetFMIndexCostRatio` /
  `InitLoonReaderThreadPool`), so the package tests are left to CI.

issue: #52255

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -432,6 +432,12 @@ proxy:
     debug_mode: false # Whether to enable http server debug mode
     port:  # high-level restful api
     acceptTypeAllowInt64: true # high-level restful api, whether http client can deal with int64
+    # high-level restful api, restore the value handling of releases that predate the REST insert validation work.
+    # When true the server keeps the previous lenient behaviour: a missing or null non-nullable field is stored as an
+    # empty value, out-of-range integers wrap instead of being rejected, numbers reach VarChar and JSON fields through
+    # their float64 rendering, and integers too large for the JSON engine become 0. This is a temporary escape hatch
+    # for clients that have not been corrected yet: every one of those behaviours silently changes what is stored.
+    compatibilityMode: false
     enablePprof: true # Whether to enable pprof middleware on the metrics port
     readHeaderTimeout: 5s # HTTP server timeout for reading request headers
     readTimeout: 0s # HTTP server timeout for reading the entire request, including the body. 0 disables this timeout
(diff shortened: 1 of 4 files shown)


xiaofan-luan:

PR #52257, opened 2026-08-06
feat: flush growing segments from querynode source through a unified write buffer

xiaofan-luan:
issue: #51643

Second half of #51878 (growing-search landed as #51953 + #51960). Stacked on #52478
(channel-op liveness), whose two fixes are duplicated here until it merges.

design doc: `docs/design-docs/design_docs/20260806-datanode-flush-coordination.md`

## Why

Growing segment rows used to exist twice in memory: once in the QueryNode growing
segment (serving queries), once in the DataNode write buffer (awaiting flush). This PR
flushes directly from the QueryNode copy — the write buffer keeps a ledger, not the
bytes — coordinated by a `(startTs, endTs]` position fence resolved in segcore to an
acked row prefix (fences never split a timestamp group; adjacent fences are disjoint
and complete; one executable definition, C++ side).

Gated by `common.storage.enableGrowingSourceFlush` (default **false**) and StorageV3.

## One write buffer, two payloads

The write buffer no longer has two flush paths. A segment's unflushed insert data is a
`segmentPayload` with two implementations:

- **ownedPayload** — rows buffered locally (non-v3, fallback, deletes/L0).
- **refPayload** — rows pinned in the segcore growing segment; the payload is a ledger
  of batch positions. `Snapshot` resolves the fence and pins the source, and **may
  fail** (source Unavailable / not yet materialized) — that failure is the first-class
  retryable shape of a reference, handled by the same debt machinery as any retry.

Everything above the payload is ONE machine: one coordinator, one per-segment queue,
one admission lease (retained across retries — the node-wide bound on materialized
payload), one debt model (metacache Sealed/Dropped state + non-empty payload/queue +
a two-verb rate-limit intent), one selection path (trigger parity with the previous
dedicated growing loop is pinned by test), one completion function. `Sealed → Flushing`
is claimed one-way at task construction. `CommitFlush` is the only settlement;
`AbortFlush` keeps the floor — a terminal failure keeps the checkpoint pinned because
WAL replay is then the only copy.

## Checkpoint: two derived pin classes, no registry

The channel checkpoint never advances past the replay origin of anything not yet
durable. That is enforced by exactly two DERIVED candidate classes:

1. `payload.EarliestPosition()` — buffered AND in-flight data, both modes (owned keeps
   a floor per snapshot until its commit; ref keeps ledger batches until the fence acks).
2. The metacache seal fence, recorded set-if-nil at seal time — covers the
   metadata-only final flush from the seal instant, before any task exists.

There is no pin registry: nothing to unregister, no path that can forget to release.
Metadata-only DROP tasks deliberately pin nothing — coordinator meta converges with
zero DataNode ack (RootCoord FastAck + infinitely-retried NotifyDropPartition →
`DropSegmentsOfPartition`; channel drop via two independent retried drivers →
`UpdateDropChannelSegmentInfo`; trace cited in code).

## Retry: dual-drive

Flowgraph timeticks drive retries as the fast path; one bufferManager ticker is the
backstop, so no wait (admission, capacity, drain) has to pump retries itself. The one
exception is documented: a dropping buffer leaves the manager map before its final
drain, so that wait drives its own retries.

## Task layer

Two-phase reorder dispatcher (parallel Prepare, per-key ordered Commit) replaces the
key-lock dispatcher. The growing task commits rows out of the source only after their
metadata is durable (point of no return, ordering pinned by test). One sync-error
classifier decides retry/terminal/cancel for both task kinds.

## Release protocol (growing segment = the only copy)

- Channel release (`UnsubDmChannel`): fence → ManualFlush → drain; the fence strictly
  precedes the WAL append; a drain error fails the release. Transient prepare failure
  refuses the unsub for retry, gated on a registered provider — NOT the local segment
  snapshot (the provider can owe debt for a segment the QueryNode has not materialized).
- Partial release (`ReleasePartitions` / target updates): enforced at the delegator
  layer — if the local write buffer still owes a growing-source flush, append one
  rate-limited ManualFlush (nudge, best-effort, at most one per 10×flushRetryInterval
  per collection/vchannel) and refuse retryably; the checker re-issues after the debt
  settles. Deliberately not the channel-wide admission fence, which only reopens on a
  new provider registration.
- `QueryNode.Stop`: bounded drain (clamped [5s, 60s]) before destroying providers,
  placed before `pipelineManager.Close()` (TSafe freezes after). Matters most for the
  embedded-querynode stop during a streaming-node rolling upgrade, when the local WAL
  is still alive to serve the drain. Destroying on timeout stays safe: rows remain in
  the WAL behind the pinned checkpoint.

## Also

- QueryCoord channel-task timeout now bounds the task context (segment/leader tasks
  deliberately unbounded); `KeyLock.LockCtx` makes queued channel-op waits cancellable
  (both stacked from #52478).
- importv2: terminal state published only by the scheduler (BlockOnAll fence); refused
  submissions are Abandoned; clones share ctx/cancel; import write-retry budget merged
  with the configurable backoff.
- Behavior deltas called out in the design doc: stale-flush trigger gains the policy's
  0–10% jitter; channel drop of a mid-flush growing segment lets Dropped supersede the
  flush; stale drop debt for a segment missing from the metacache is refused instead of
  building a task that dies on SegmentNotFound.

## Test

`internal/flushcommon/...`, `internal/datanode/...` (incl. importv2, compactor),
`internal/querynodev2/...`, `internal/querycoordv2/task`, `internal/streamingnode/...`,
`internal/distributed/streaming/...` pass, `-race` on the concurrency-sensitive suites;
gofmt/golangci-lint clean. C++ fence semantics covered by row-content tests (grouped
timestamps, start-exclusive/end-inclusive, adjacent windows disjoint-and-complete by
pk values, flush/pk-reader agreement). A 12-item invariant audit maps each design
invariant to a named discriminating test; the load-bearing ones (checkpoint pins, floor
retention on terminal failure, fence-before-append, point-of-no-return ordering,
retry-survives-admission-wait, exactly-one-metadata-flush, double-submission gate) are
mutation-verified — each test was confirmed RED against a deliberately broken
implementation.

## Known limitations

- DataCoord's `Flushing` state is dead — nothing writes it; cleanup independent.
- `HandleLoonFFIResult` flattens FFI failures into `ErrLoonTransient`; until fixed,
  layout mismatch is classified by message text.
- The segcore backstop for "materialized field with no data" returns `UnexpectedError`
  (retryable); routing it terminal is deferred to the error-code rework.
- The integration test cannot yet positively assert the growing branch engaged
  (needs a per-source metric label; logged TODO).
- DataNode collection metrics are ref-counted per (node, collection); per-vchannel
  rekey is a follow-up.


xiaofan-luan:
Closing this in favour of #53619, which takes the opposite direction.

This PR built further on growing-source flush: it made the QueryNode copy the
flush source through a unified write buffer, adding about 20k lines across 127
files. #53619 instead removes growing-source flush entirely and converges the
flush-path state it was tangled with, deleting about 15.8k lines.

The reasoning for the reversal:

- The feature has been gated off by default since it landed
  (`common.storage.enableGrowingSourceFlush`), so nothing runs it, yet it
  duplicates most of `SyncTask`, adds a seven-dimension progress state machine to
  `writeBufferBase`, and spans a release protocol across QueryNode and both
  halves of StreamingNode.
- It has four open bugs against it: #51854, #51635, #51643 (this PR's issue),
  #50915. Building on top of it meant carrying those forward.

Removing it first also makes the underlying flush path tractable: with the second
source gone, row accounting, checkpoint pinning and terminal-outcome policy each
collapse to a single mechanism, which is what #53619 does.

If we want the memory saving later, it should be rebuilt on that converged path
rather than beside it. Tracking issue for the removal: #53618.

PR #52260, opened 2026-08-06
fix: only re-prioritize the compaction queue when the config changes

### LINKED ISSUE #52259: [Bug]: Compaction queue is re-prioritized on every tick (always-true pointer comparison) and UpdatePrioritizer has a data race

### Environment

Found by code audit on `master` (verified present at `232823a008`).

### Current Behavior

`internal/datacoord/compaction_inspector.go:243-246`:

```go
p := getPrioritizer()
if &c.queueTasks.prioritizer != &p {
    c.queueTasks.UpdatePrioritizer(p)
}
```

`&c.queueTasks.prioritizer` is the address of a struct field; `&p` is the address of a local variable. **These are never equal**, so the guard is always true and `UpdatePrioritizer` runs on every scheduling tick.

The intent is clearly "only rebuild when the configured prioritizer changed" — but `Prioritizer` is a `func` value and Go does not permit `==` on funcs, which is presumably why the pointer comparison was reached for.

`UpdatePrioritizer` re-computes the priority of every queued task and calls `heap.Init`:

```go
func (q *CompactionQueue) UpdatePrioritizer(prioritizer Prioritizer) {
	q.prioritizer = prioritizer     // <-- assigned before the lock is taken
	q.lock.Lock()
	defer q.lock.Unlock()
	for i := range q.pq {
		q.pq[i].priority = q.prioritizer(q.pq[i].value)
	}
	heap.Init(&q.pq)
}
```

`schedule()` is driven by a 500ms ticker, so this is O(queue length) work plus a full `heap.Init` twice a second, held under the queue write lock. `dataCoord.compactionTaskQueueCapacity` defaults to 100000.

**Second defect in the same code path**: `q.prioritizer = prioritizer` at `compaction_queue.go:127` happens *before* `q.lock.Lock()`, while `Enqueue`/`Dequeue` read `q.prioritizer` under the lock. That is a data race.

### Expected Behavior

The queue should only be re-prioritized when the `dataCoord.compactionTaskPrioritizer` configuration actually changes, and `q.prioritizer` should only be mutated under the lock.

### Steps To Reproduce

The comparison is unconditionally true:

```go
type queueT struct{ prioritizer Prioritizer }
q := &queueT{prioritizer: DefaultPrioritizer}
for i := 0; i < 10; i++ {
    p := getPrioritizer()   // same configured value every time
    if &q.prioritizer != &p {
        // reached 10/10 times
    }
}
```

At the 500ms `schedule()` cadence this is ~172,800 unnecessary rebuilds per day.

For the data race, run any test that calls `UpdatePrioritizer` concurrently with `Enqueue` under `-race`.

### Milvus Log

N/A - static defect.

### Anything else?

Fix: give `CompactionQueue` a `prioritizerName` field and compare the configuration name (funcs cannot be compared, names can), and move the assignment inside the lock. PR to follow.

**Unrelated blocker found while verifying**: the `internal/datacoord` test package does not compile on `master` — `meta_test.go:2362` calls `getSegmentBinlogFields`, which does not exist anywhere outside comments. I had to move `meta_test.go` aside to run any test in this package. Filing separately is probably warranted.

xiaofan-luan:
Fixes #52259

## What

Two defects in the compaction scheduling path:

**1. The queue was re-prioritized on every tick.**

```go
p := getPrioritizer()
if &c.queueTasks.prioritizer != &p {   // address of a field vs address of a local
    c.queueTasks.UpdatePrioritizer(p)  // -> always taken
}
```

`Prioritizer` is a func value and cannot be compared with `==`, which is presumably
why a pointer comparison was reached for. The two addresses are never equal, so
every 500ms `schedule()` tick re-computed the priority of every queued task and
called `heap.Init`, under the queue write lock.
`dataCoord.compactionTaskQueueCapacity` defaults to 100000.

**2. `UpdatePrioritizer` had a data race.**

```go
q.prioritizer = prioritizer   // before the lock
q.lock.Lock()
```

`Enqueue`/`Dequeue` read `q.prioritizer` under the lock.

## How

- `CompactionQueue` now keeps `prioritizerName` next to `prioritizer`, both guarded
  by `lock`. The configuration name is used as the prioritizer's identity, since the
  func value itself cannot be compared.
- New `SyncPrioritizer(name)` does the compare-and-update atomically under the lock;
  the inspector calls it unconditionally on each tick and it is a no-op unless the
  configuration changed.
- `getPrioritizer()` is split into `getPrioritizerName()` + `getPrioritizerByName()`
  so the name and the func stay in sync. `getPrioritizer()` is kept as a thin wrapper
  (still used by `newCompactionInspector`).
- `UpdatePrioritizer` keeps its signature (used by existing tests) and now assigns
  under the lock.

`NewCompactionQueue` leaves `prioritizerName` empty, so the first `SyncPrioritizer`
adopts the configured value. The queue is empty at that point, so the initial
re-heapify costs nothing.

## Tests

`TestCompactionQueue_SyncPrioritizer`:

- **same name does not re-prioritize** — the actual regression guard. A counter on
  an injected `Prioritizer` would not work here: `SyncPrioritizer` resolves the
  prioritizer *by name*, so any injected closure is discarded on the first call and
  the counter would read zero either way. The test instead primes the queue, mutates
  the task's `PlanID` to 42, runs 100 same-name syncs, and asserts the stored
  priority is unchanged — removing the early return yields 42. The tail then syncs
  to `"level"` (10), so the three values 1 / 42 / 10 are mutually distinct and that
  assertion discriminates.
- **empty configuration value is adopted rather than read as unset** — `""` is a
  settable value, not "unset", so it must resolve to `DefaultPrioritizer` rather
  than matching a not-yet-adopted name and leaving the constructor's prioritizer in
  place.
- **update forgets the configuration name** — `UpdatePrioritizer` sets the func out
  of band, so the next sync must re-adopt rather than skip.
- **changed name does re-prioritize** — asserts the name is adopted.
- **no data race with concurrent enqueue** — hammers `UpdatePrioritizer` /
  `SyncPrioritizer` against `Enqueue` / `Dequeue`; meaningful under `-race`.

Verified locally with `-race`; existing `TestCompactionQueue` subtests still pass.

## Measured

The re-prioritize itself is cheap; this is a correctness fix, not a hot-path win.
Reproducing `heap.Init` plus one prioritizer call per element (Apple M4 Pro,
go1.26.5):

| queue depth | per re-prioritize | duty cycle at the 500ms tick |
|---|---|---|
| 1,000 | 2.4 µs | ~0% |
| 10,000 | 32 µs | 0.006% |
| 100,000 (default `taskQueueCapacity`) | 374 µs | 0.075% |

So even a full queue costs 374 µs every 500 ms. What makes this worth fixing is
that the guard never did what it claimed, and that `UpdatePrioritizer` wrote
`q.prioritizer` outside the lock while `Enqueue` read it inside — a race that fired
every tick precisely because the guard was always true.

### DIFF
--- a/internal/datacoord/compaction_inspector.go
+++ b/internal/datacoord/compaction_inspector.go
@@ -207,6 +207,13 @@ func (c *compactionInspector) checkSchedule() {
 
 func (c *compactionInspector) schedule() []CompactionTask {
 	selected := make([]CompactionTask, 0)
+
+	// Sync before the empty-queue early return, so a configuration change made
+	// while the queue happens to be empty is still adopted. The cost on an
+	// empty queue is one lock acquisition and one string compare -- the
+	// re-prioritize loop iterates zero times.
+	c.queueTasks.SyncPrioritizer(getPrioritizerName())
+
 	if c.queueTasks.Len() == 0 {
 		return selected
 	}
@@ -239,8 +246,6 @@ func (c *compactionInspector) schedule() []CompactionTask {
 			c.queueTasks.Enqueue(t)
 		}
 	}()
-
-	c.queueTasks.SyncPrioritizer(getPrioritizerName())
 
 	// The schedule loop will stop if either:
 	// 1. no more task to schedule (the task queue is empty)
--- a/internal/datacoord/compaction_inspector.go
+++ b/internal/datacoord/compaction_inspector.go
@@ -240,10 +240,7 @@ func (c *compactionInspector) schedule() []CompactionTask {
 		}
 	}()
 
-	p := getPrioritizer()
-	if &c.queueTasks.prioritizer != &p {
-		c.queueTasks.UpdatePrioritizer(p)
-	}
+	c.queueTasks.SyncPrioritizer(getPrioritizerName())
 
 	// The schedule loop will stop if either:
 	// 1. no more task to schedule (the task queue is empty)
--- a/internal/datacoord/compaction_queue.go
+++ b/internal/datacoord/compaction_queue.go
@@ -85,10 +85,14 @@ var (
 type Prioritizer func(t CompactionTask) int
 
 type CompactionQueue struct {
-	pq          PriorityQueue[CompactionTask]
-	lock        lock.RWMutex
-	prioritizer Prioritizer
-	capacity    int
+	pq   PriorityQueue[CompactionTask]
+	lock lock.RWMutex
+	// prioritizer and prioritizerName are guarded by lock.
+	// Prioritizer is a func value and therefore cannot be compared with ==,
+	// so the configuration name is kept alongside it as its identity.
+	prioritizer     Prioritizer
+	prioritizerName string
+	capacity        int
 }
 
 func NewCompactionQueue(capacity int, prioritizer Prioritizer) *CompactionQueue {
@@ -124,9 +128,25 @@ func (q *CompactionQueue) Dequeue() (CompactionTask, error) {
 }
 
 func (q *CompactionQueue) UpdatePrioritizer(prioritizer Prioritizer) {
-	q.prioritizer = prioritizer
 	q.lock.Lock()
 	defer q.lock.Unlock()
+	q.updatePrioritizerLocked("", prioritizer)
+}
+
+// SyncPrioritizer re-prioritizes the queue only when the configured prioritizer
+// actually changed. It is safe to call on every scheduling tick.
+func (q *CompactionQueue) SyncPrioritizer(name string) {
+	q.lock.Lock()
+	defer q.lock.Unlock()
+	if q.prioritizerName == name {
+		return
+	}
+	q.updatePrioritizerLocked(name, getPrioritizerByName(name))
+}
+
+func (q *CompactionQueue) updatePrioritizerLocked(name string, prioritizer Prioritizer) {
+	q.prioritizer = prioritizer
+	q.prioritizerName = name
 	for i := range q.pq {
 		q.pq[i].priority = q.prioritizer(q.pq[i].value)
 	}
@@ -194,9 +214,12 @@ var (
 	}
 )
 
-func getPrioritizer() Prioritizer {
-	p := Params.DataCoordCfg.CompactionTaskPrioritizer.GetValue()
-	switch p {
+func getPrioritizerName() string {
+	return Params.DataCoordCfg.CompactionTaskPrioritizer.GetValue()
+}
+
+func getPrioritizerByName(name string) Prioritizer {
+	switch name {
 	case "level":
 		return LevelPrioritizer
 	case "mix":
@@ -204,4 +227,8 @@ func getPrioritizer() Prioritizer {
 	default:
 		return DefaultPrioritizer
 	}
+}
+
+func getPrioritizer() Prioritizer {
+	return getPrioritizerByName(getPrioritizerName())
 }
--- a/internal/datacoord/compaction_queue.go
+++ b/internal/datacoord/compaction_queue.go
@@ -90,8 +90,16 @@ type CompactionQueue struct {
 	// prioritizer and prioritizerName are guarded by lock.
 	// Prioritizer is a func value and therefore cannot be compared with ==,
 	// so the configuration name is kept alongside it as its identity.
+	//
+	// prioritizerName is nil while prioritizer has not been resolved from the
+	// configuration -- the constructor and UpdatePrioritizer both take a func
+	// directly. It is a pointer rather than a string because "" is itself a
+	// settable configuration value (the RESTful alterConfig contract treats a
+	// present value, including the empty string, as a set and only null as a
+	// reset, and dataCoord.compaction.taskPrioritizer has no Formatter), so a
+	// string zero value cannot stand in for "unset" without colliding with it.
 	prioritizer     Prioritizer
-	prioritizerName string
+	prioritizerName *string
 	capacity        int
 }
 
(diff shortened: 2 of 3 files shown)


xiaofan-luan:

```
```

czs007:

Adversarial review of the compaction prioritizer sync change surfaced one correctness issue in the scheduling tick and two test/documentation accuracy issues.

## Medium

**`internal/datacoord/compaction_inspector.go:243` — `SyncPrioritizer` sits after the empty-queue early return, so an empty-string config value collides with the initial `prioritizerName` (`""`) and the prioritizer never switches.**

The new `c.queueTasks.SyncPrioritizer(getPrioritizerName())` replaces the old unconditional `UpdatePrioritizer`, but the call site is still below `if c.queueTasks.Len() == 0 { return selected }` (compaction_inspector.go:210-211). Ticks with an empty queue skip the sync entirely, so `prioritizerName` stays at its zero value `""`. `NewCompactionQueue(capacity, prioritizer)` (compaction_queue.go:98-105) seeds only the prioritizer func, not the name, while the inspector passes `getPrioritizer()` (compaction_inspector.go:186) resolved from a config whose `DefaultValue` is `"level"` (pkg/util/paramtable/component_param.go:5764-5773). A default deployment therefore holds `LevelPrioritizer` + name `""`. The key can legitimately be set to an empty string: the RESTful `alterConfig` contract states that a present value — including `""` — is a set, and only `null` is a reset (internal/coordinator/restful_mgr_routes.go:1252-1253); the item has no `Formatter` and `PanicIfEmpty` is false, so `getWithRaw()` returns `""` verbatim (pkg/util/paramtable/param_item.go:157-162).

Failure: datacoord starts with the default `dataCoord.compaction.taskPrioritizer=level` and the compaction queue is empty on every 500ms schedule tick (new or low-write cluster), then an operator sets the key to `""` via `/management/config` intending FIFO → `SyncPrioritizer("")` matches the initial zero-value name and returns early at compaction_queue.go:141-143, the queue keeps `LevelPrioritizer` (L0=1 < Mix=10 < Clustering=100) instead of `DefaultPrioritizer`, and the API reports success with no log line; the pre-change code recomputed every tick and would have switched correctly.

Suggestion: move `SyncPrioritizer` above the `Len() == 0` early return — the per-tick cost is one lock acquisition plus a string compare, and `updatePrioritizerLocked` iterates zero times on an empty queue — or seed `prioritizerName` in `NewCompactionQueue` alongside the func. The trigger window is narrow (only if the queue has never been non-empty on any tick since startup, and it self-heals on any non-empty value or a restart), but the divergence from the previous behavior is silent. *(raised by xaxys, tinswzy)*

## Low

**`internal/datacoord/compaction_queue_test.go:231` — the trailing assertion in the "same name does not re-prioritize" subtest is vacuous.**

`primed` is `DefaultPrioritizer` over a `CompactionType_MixCompaction` task, i.e. `1` (compaction_queue_test.go:207-212). The trailing `cq.SyncPrioritizer("mix")` then asserts against `MixFirstPrioritizer(task)`, which also returns `1` for MixCompaction (compaction_queue.go:181-184), and `pq[0].priority` is already `1` going in — so the assertion passes whether or not `updatePrioritizerLocked` actually ran, contradicting the adjacent comment "proving the assertion above is not vacuous" (compaction_queue_test.go:225-226). The main assertion (PlanID changed to 42, then 100 same-name syncs, still `1`, compaction_queue_test.go:214-223) is a genuine guard: removing the early return yields 42 ≠ 1.

Suggestion: change the tail to `cq.SyncPrioritizer("level")` — `LevelPrioritizer` returns `10` for MixCompaction (compaction_queue.go:169-170) — so the three values 1 / 42 / 10 are mutually distinct and the assertion discriminates. *(raised by tinswzy)*

**`internal/datacoord/compaction_queue_test.go:190` — the PR description of the new test does not match the committed test code.**

The PR body's Tests section says the subtest "counts prioritizer invocations across 100 `SyncPrioritizer` calls with unchanged config; asserts zero", but the committed test has no counter. Its own comment explains why a counter is not viable — `SyncPrioritizer` resolves the prioritizer by name, so any injected closure is discarded on the first call and a call counter would stay at zero either way (compaction_queue_test.go:190-196) — and the test instead observes the stored priority after mutating `PlanID`. The mismatch overstates the guard's strength to reviewers (zero invocations vs. one unchanged priority value).

Suggestion: update the PR description to describe the indirect stored-priority observation that the code actually performs. *(surfaced during verification)*

czs007:

Verified:
- `internal/datacoord/compaction_queue.go:115`, `:141`, `:150` — every prioritizer read, write, identity change, and queue operation now runs under the same lock, eliminating the previous data race where the prioritizer field was written outside the lock while `Enqueue` read it inside.
- `internal/datacoord/compaction_queue.go:153` — using a `*string` for the synced configuration name cleanly distinguishes the unsynchronized state from the valid empty-string value; verified against the paramtable implementation that an explicitly saved `""` is returned as-is rather than falling back to the default, so a plain string would have collided with the zero value.
- `internal/datacoord/compaction_inspector.go:215` — the configuration snapshot is synchronized before the empty-queue early return, closing the window where a configuration change never took effect while the queue happened to stay empty; the original always-true guard (struct-field address compared against a local variable's address) is confirmed real, so the fix direction is correct.
- `internal/datacoord/compaction_queue.go:160` — priority recomputation and `heap.Init` are atomic with respect to enqueue and dequeue, preserving the heap invariant.
- `internal/datacoord/compaction_queue.go:233` — the `level` / `mix` / default mapping preserves the previous prioritizer semantics, and since all prioritizers depend only on `PlanID`/`Type` (immutable for enqueued tasks), removing the per-tick recomputation causes no behavioral regression.
- `internal/datacoord/compaction_queue_test.go:192`, `:245`, `:270`, `:306` — the tests discriminate real regressions: unchanged-name no-ops, configuration changes with distinct priority values, the empty-string case (which would fail under a plain-string implementation), out-of-band updates, and concurrent access under the race detector.

No findings, non-blocking observations, disputed items, or pre-existing hazards survived verification.

czs007 (approved):
/lgtm

PR #52261, opened 2026-08-06
fix: stop the REST API from rewriting the values a caller sends

xiaofan-luan:
> The description below is kept in lockstep with the squashed commit message; earlier revisions of this PR described earlier code.

The REST layer read values through gjson's String(), which renders a value
rather than returning the token the caller sent: it unquotes a JSON string, and
it formats a number by parsing it into a float64 and printing it with the 'f'
verb. The failures were then swallowed, since cast.ToInt8E narrows without a
range check and cast.ToInt64 discards its error and returns 0.

The same shape appeared on every path a value can travel.

What was stored. 128 in an Int8 field became -128 and reported success, so a
later i8 == 128 found nothing. 9007199254740993.0 in an Int64 field became
...992. A field that was neither nullable nor had a default was stored empty
when it was missing. 1e300 in a VARCHAR field became 301 bytes of digits, and
an object was flattened into its own text. {"j": "hello"} in a JSON field was
stored as the bytes `hello`, which is not a JSON document, and the row then
matched no JSON predicate at all -- not even an exact match on what was written
-- with nothing logged and the JSON index reporting every row as indexed. In a
dynamic field 1e300, 1e19 and any integer beyond int64 were stored as 0, and a
nested 9007199254740993 came back as ...992.

What came back. A dynamic field above 2^53 lost precision on the way out, and a
partial upsert of one key rewrote every other key in the row, so an untouched
9007199254740993 became ...992.

What a filter matched. exprParams arrived already decoded, so a parameter of
9007199254740993 silently matched 9007199254740992 while the same value written
as a literal matched correctly, and an integer past int64 fell back to a double
and matched a different row again. null, an empty array and an object each
panicked into an HTTP 500.

A whole number a double cannot be trusted with is now refused rather than
carried. The refusal covers the whole range where a double can be mistaken for
a 64-bit integer, 2^63 through 2^64, rather than only the part of it that
actually rounds: separating the two needs exact arithmetic on the literal, and
that turns a dozen bytes such as 1e-1000000 into a million-digit rational. The
question is asked of the value rather than the spelling, so 9223372036854775809
and ...809.0 are refused alike, and the price of the wider range is that 2^63,
2^64 and 1e19 are refused as parameters although a double carries them exactly.
Written as an integer, such a value is refused by the expression parser too, so
the two paths agree there; written with a point or an exponent the parser takes
it as a double and accepts it, and this path is the stricter of the two.

What an id list turned into. The primary key filter was built by formatting the
ids into the expression text with no escaping, so {"id": ["alice\", \"bob"]}
deleted two rows on v1 and returned two rows on v2 get, and an id containing a
quote could not be fetched at all. Search-by-id lost precision past 2^53 and
formatted a numeric id for a VARCHAR key as "1e+06".

Every path now reads the literal. Static columns convert from value.Raw, with
integers parsed exactly by parseJSONInteger, so 1.0, 1e3 and 100e-2 keep working
while out-of-range and fractional values are rejected. A Float column is read
through float64 and then narrowed, the way every path this value will be
compared against reads it -- the expression parser, an exprParams value
carried as a double, a row written through pymilvus whose Python float is a
double before it becomes a float32. Parsing straight to float32 rounds once
and reads the decimal more faithfully, but a literal sitting on a float32
rounding midpoint then stores a value none of those paths produce and the row
cannot be found with the literal that wrote it. What the shared dialect costs
is the one check the float32 parser makes for free: 3.5e38 is finite as a
float64, so the parse succeeds and the narrowing silently yields +Inf. The
range is checked before the cast instead, in both places a float is narrowed.
JSON and dynamic columns
keep the caller's text rather than being decoded and re-encoded, since that
round trip could only lose information; what they hold is validated by a walk
of its own, described below, which reads the bytes without rewriting them --
nested numbers included. Ids and filter parameters travel as template
values, so nothing a caller sends is parsed as expression syntax.

Values the storage engine cannot read back are rejected at insert time rather
than stored. The binder guarantees syntactically valid JSON and nothing more: it
accepts an integer beyond 64 bits (BIGINT_ERROR), invalid UTF-8, which it
replaces with U+FFFD on decode while the raw bytes keep the offending byte
(UTF8_ERROR), nesting past the 1024 levels the DOM parser behind the JSON index
allows (DEPTH_ERROR), and an object that declares the same key twice, where the
readers disagree about which value wins. All of it -- duplicate keys, depth,
oversized integers, UTF-8 -- is checked in one pass of the standard library's
tokenizer per document, over the bytes that actually land. The pass is new
and is a real cost this change takes on -- linear in the document, an
allocation per token -- which is what rejecting these at the door costs. A
recursive walk was tried first and re-scanned each child's subtree to find
its extent, O(depth * bytes) on a deep document with a wide leaf; one engine
that reads every byte once beats two engines or a tuned budget. The assembled dynamic wrapper adds a level
of nesting and can turn two keys with invalid UTF-8 into one, so it gets the
same walk. A lone surrogate is the one case
that falls back rather than being rejected, and the fallback is refused if
normalizing would change a number the document carries.

proxy.http.compatibilityMode restores the previous handling for all of it. It
defaults to false and is refreshable, so an operator can unblock clients without
downgrading. It is an escape hatch rather than a supported mode: every behavior
it restores silently changes what is stored, and the tests assert that turning it
on brings the wrong values back.

A vector can no longer be handed over as the text of its own JSON shape.
"[0.1, 0.2]" was accepted where [0.1, 0.2] was meant, because three of the
insert branches read gjson's String(), which returns a string node's content
with the quotes removed: the difference was gone before the field type was
consulted. Nothing had decided to allow it. Every neighbor that reads the node
instead refuses it -- BinaryVector and the two 16-bit floats take a string as
their base64 spelling, struct sub-vectors ask IsArray, and the float and int8
query paths read the raw element, so a row written with the quoted form could
not be looked up with it. The rule is now the same everywhere: a string is a
vector only where the type has a base64 spelling for one, restorable through
compatibilityMode like every other tightening here.

Applying that rule turned up an error in the list it reads. Int8Vector was
recorded as having a base64 spelling, which no decoder implements, and the
entry exempted it from the null-literal check -- so "null" in an Int8 vector
field decoded to a nil slice without an error and was stored as an empty
vector. It is off the list.

A query-side list can no longer be null where its callers test for absence. The
UnmarshalJSON that rejects a null element decoded a whole-value null to a nil
slice without an error and then handed back a non-nil empty one, so
{"vector": null} slipped past the v1 required-field check with a vector that
held nothing. A whole-value null is now refused outright, for the vector and
the id list both; an empty list stays what it was.

Converting an expression template parameter walks the caller's nesting
recursively, so the depth is now bounded: proxy.http.maxExprParamsDepth,
default 100, read once per request and clamped to a ceiling of 1024 that
configuration cannot raise, counting arrays and objects both. Not gated by
compatibilityMode -- it is a resource bound, not a value rule, and the
handling it would restore is a stack overflow.

The null rules are screened by a substring scan before any walk runs: a null
can only exist where the letters "null" appear, and a vector's text is digits,
brackets and commas, so the request that carries no null -- which is nearly
every request -- pays a vectorized memchr and nothing else. Benchmarked
against master on the same inputs: query-vector parsing (4x768 floats) and a
scalar+vector insert row land at or slightly below the old cost, and a row
carrying a 500-byte JSON document pays about eight percent for the document
walk described above, which is the one cost this change keeps on purpose.

An array element can no longer be null. An array has no element-level validity,
so a null could only be stored as the element type's zero value and was never
recoverable. sonic already refused one for an integer or string element, so this
only changes a boolean element, which silently became false, and a float, which
became 0. The error now names the position.

An absent key and an explicit null in a dynamic field are no longer the same
request. Skipping the null collapsed them, so a partial update could not clear a
dynamic field: {"tag": null} merged to nothing and the old value survived. The
null is now carried, which is what the partial update end-to-end test expects.

A JSON field can read back as the document it holds rather than as a string,
under proxy.http.nativeJSONResponse, which also defaults to false. Returning the
string is not a design choice: the stored bytes are not guaranteed to be a
document, and embedding one of those natively makes the marshal of the whole
response fail, so a single legacy row would break every query that selects it.
The flag therefore comes with the check that makes it safe to turn on with
existing data -- each value is validated as the response is built, and if any row
does not hold a document then every JSON field in that response falls back to the
string form with a warning. Degrading the whole response rather than the one row
is deliberate: a caller can handle "always a document" or "always a string", but
a field that is sometimes each is worse than the shape being fixed.

A JSON field holding a string reads back differently than it did, and that is
accepted rather than papered over. The bytes stored for one now carry their
quotes, so through the default string form -- which returns the stored bytes --
a value written as "hello" comes back as "\"hello\"" where it used to come back
as hello, and a row written before this change reads unlike a row written after
it. Unwrapping the string on the way out would hide both, but only by changing
what a row written through gRPC reads back as, since those always stored the
quotes and always read back with them. There is no rendering that leaves every
caller where it was, because the two writers did not agree in the first place --
that disagreement is the bug being fixed. The correct bytes win, and
nativeJSONResponse returns the value rather than its storage for callers who
want it. Worth a release note.

One thing is left alone deliberately and documented where it lives: a row that
declares the same key twice is still resolved to the first occurrence, which
would need a scan of the whole row before the map is built.

The low-level /api/v1 wrapper that registerHTTPServer mounts on the metrics
port decoded its payloads straight into []bool, []float32 and friends, so a
null element silently became the zero value -- false or 0 -- and was then
stored by insert, searched, or measured by calc_distance. A null anywhere in
those payloads is now refused, the same rule the high-level handlers apply:
this API has no way to express validity, so there is nothing a null could
honestly become. The rest of that wrapper's value handling reads typed JSON
directly rather than rendering through String(), so it is left unchanged.

Supersedes #52097, #52099, #52213, #52256 and #52103; the response-side fix is
taken from #52103.

---

Fixes the REST value-handling reports: #52049 #52050 #52212 #52255 #52307 #52313 #52315

https://claude.ai/code/session_01D1pij3TGike2nZy7t6oDZY

### DIFF
--- a/internal/distributed/proxy/httpserver/request.go
+++ b/internal/distributed/proxy/httpserver/request.go
@@ -16,6 +16,11 @@
 
 package httpserver
 
+import (
+	"github.com/milvus-io/milvus/internal/json"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
+)
+
 type CreateCollectionReq struct {
 	DbName             string `json:"dbName"`
 	CollectionName     string `json:"collectionName" validate:"required"`
@@ -83,6 +88,82 @@ type SingleUpsertReq struct {
 	FieldOps       []FieldPartialUpdateOpReq `json:"fieldOps"`
 }
 
+// FloatVectorQuery is a search vector that refuses a null coordinate.
+//
+// Decoding straight into []float32 turns a null into 0, which is a coordinate
+// the caller never sent and which takes full part in the distance computation,
+// so the search silently runs against a different point. A vector is a dense
+// fixed-width array of numbers with no per-element validity to record "absent"
+// in, so there is nothing to store but a number: the only honest answer is to
+// refuse. The v2 path applies the same rule in serializeFloatVectors.
+type FloatVectorQuery []float32
+
+func (v *FloatVectorQuery) UnmarshalJSON(data []byte) error {
+	var elements []*float32
+	if err := json.Unmarshal(data, &elements); err != nil {
+		return err
+	}
+	if elements == nil {
+		// A whole-value null decodes to a nil slice without an error, and the
+		// loop below would then hand back a non-nil empty vector -- which is
+		// how a null used to slip past the callers' required-field checks.
+		return merr.WrapErrParameterInvalidMsg("a vector cannot be null")
+	}
+	parsed := make([]float32, 0, len(elements))
+	for idx, element := range elements {
+		if element == nil {
+			return merr.WrapErrParameterInvalidMsg(
+				"search vector has a null at index %d; a vector element cannot be null", idx)
+		}
+		parsed = append(parsed, *element)
+	}
+	*v = parsed
+	return nil
+}
+
+// Int64ListQuery is a list of int64 that refuses a null element, for the same
+// reason FloatVectorQuery does: encoding/json leaves a null as 0, which names
+// a different row than the caller did. See FloatVectorQuery.
+type Int64ListQuery []int64
+
+func (v *Int64ListQuery) UnmarshalJSON(data []byte) error {
+	var elements []*int64
+	if err := json.Unmarshal(data, &elements); err != nil {
+		return err
+	}
+	if elements == nil {
+		// same rule as FloatVectorQuery: a null is not an empty list
+		return merr.WrapErrParameterInvalidMsg("an id list cannot be null")
+	}
+	parsed := make([]int64, 0, len(elements))
+	for idx, element := range elements {
+		if element == nil {
+			return merr.WrapErrParameterInvalidMsg(
+				"id list has a null at index %d; an id cannot be null", idx)
+		}
+		parsed = append(parsed, *element)
+	}
+	*v = parsed
+	return nil
+}
+
+// Base64VectorQuery is a base64-spelled vector that refuses a null. Decoding
+// straight into []byte turns a null into an empty slice, which then travels as
+// a zero-length vector rather than being reported.
+type Base64VectorQuery []byte
+
+func (v *Base64VectorQuery) UnmarshalJSON(data []byte) error {
+	var decoded *[]byte
+	if err := json.Unmarshal(data, &decoded); err != nil {
+		return err
+	}
+	if decoded == nil {
+		return merr.WrapErrParameterInvalidMsg("a vector cannot be null")
+	}
+	*v = *decoded
+	return nil
+}
+
 type SearchReq struct {
 	DbName            string                `json:"dbName"`
 	CollectionName    string                `json:"collectionName" validate:"required"`
@@ -90,7 +171,7 @@ type SearchReq struct {
 	Limit             int32                 `json:"limit"`
 	Offset            int32                 `json:"offset"`
 	OutputFields      []string              `json:"outputFields"`
-	Vector            []float32             `json:"vector"`
+	Vector            FloatVectorQuery      `json:"vector"`
 	Params            map[string]float64    `json:"params"`
 	SearchAggregation *SearchAggregationReq `json:"searchAggregation"`
 }
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -17,9 +17,12 @@
 package httpserver
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
+	gojson "encoding/json"
 	"fmt"
+	"io"
 	"math"
 	"net/http"
 	"reflect"
@@ -27,6 +30,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/gin-gonic/gin"
 	"github.com/spf13/cast"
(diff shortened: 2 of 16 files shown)


internal/distributed/proxy/httpserver/utils.go
```diff
@@ -758,7 +900,34 @@ func checkAndSetData(body []byte, collSchema *schemapb.CollectionSchema, partial
...
+					// JSON documents and make the stored field unparsable. The
+					// request body is already validated as JSON by the gin binder
+					// before it reaches here, so Raw is always a well-formed token.
+					// A JSON document supplied as a JSON string is still unwrapped,
+					// preserving the existing input form.
+					switch {
+					case compatibilityMode:
+						reallyData[fieldName] = []byte(dataString)
+					case fieldValue.Type == gjson.String && json.Valid([]byte(dataString)):
+						reallyData[fieldName] = []byte(dataString)
```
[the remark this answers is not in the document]
xiaofan-luan: Addressed. The string-encoded form no longer bypasses the number check: the unwrapped document goes through `jsonDocumentForStorage` → `checkEngineCompatible`, whose walk runs `jsonNumberLiteral` on every number node — including a bare top-level number. Verified on the current head: `"18446744073709551616"` and `"{\"a\": 18446744073709551616}"` are both rejected at insert; `"123"` still unwraps and stores. Covered by `TestCheckAndSetDataJSONFieldUnwrapsEncodedDocument` / `TestCheckAndSetDataJSONFieldRejectsNestedUnreadableNumbers`.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -943,12 +1306,27 @@ func buildStructSubArrayScalar(sub *schemapb.FieldSchema, vals []gjson.Result) (
...
+			bitSize = 16
+			minValue, maxValue = math.MinInt16, math.MaxInt16
+		}
 		arr := make([]int32, 0, len(vals))
 		for _, v := range vals {
 			if v.Type != gjson.Number {
 				return nil, wrapStructSubParseError(sub, v, "expect integer")
 			}
-			arr = append(arr, int32(v.Int()))
+			value, ok := parseJSONInteger(v.Raw, bitSize)
```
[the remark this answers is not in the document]
xiaofan-luan: Addressed. `compatibilityMode` is threaded into this path now — `parseStructArrayRow(rawValue.Raw, structField, compatibilityMode)` — and every integer width in `buildStructSubArrayScalar` has the legacy branch: `if compatibilityMode { arr = append(arr, v.Int()); continue }` before the exact parser, so `1e-999999` becomes `0` again under the escape hatch. Float subfields restore `float32(v.Float())` the same way.

internal/distributed/proxy/httpserver/request.go
```diff
@@ -83,14 +88,90 @@ type SingleUpsertReq struct {
...
+//
+// Decoding straight into []float32 turns a null into 0, which is a coordinate
+// the caller never sent and which takes full part in the distance computation,
+// so the search silently runs against a different point. A vector is a dense
+// fixed-width array of numbers with no per-element validity to record "absent"
+// in, so there is nothing to store but a number: the only honest answer is to
+// refuse. The v2 path applies the same rule in serializeFloatVectors.
+type FloatVectorQuery []float32
+
+func (v *FloatVectorQuery) UnmarshalJSON(data []byte) error {
```
MrPresent-Han, line 101: **P2 (non-blocking, pre-existing; not introduced by this PR): cover the supported YAML binding too**

The low-level routes still accept `application/x-yaml`, but Gin's YAML decoder does not invoke these new `UnmarshalJSON` hooks, so `vectors: [[0.1, null, 0.2]]` still reaches Proxy as `[[0.1, 0, 0.2]]`. Please add format-independent/YAML null validation (or explicitly remove YAML support) and a MIMEYAML regression test.

MrPresent-Han (commented):
Non-blocking P2 follow-up.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -970,7 +1790,17 @@ func buildStructSubArrayScalar(sub *schemapb.FieldSchema, vals []gjson.Result) (
...
+				arr = append(arr, float32(v.Float()))
+				continue
+			}
+			// through float64 like every path this value will be compared
+			// against -- see the Float case in checkAndSetData
+			parsed, err := strconv.ParseFloat(v.Raw, 64)
+			if err != nil || math.IsInf(parsed, 0) {
+				return nil, wrapStructSubParseError(sub, v, "expect float")
+			}
+			arr = append(arr, float32(parsed))
```
MrPresent-Han, line 1870: **P1 (pre-existing; this PR's strict validation still leaves it open): reject float32 overflow before narrowing**

`3.5e38` is finite as a `float64`, so the current check passes, but `float32(parsed)` becomes `+Inf`; Proxy's `Array<Float>` validation only checks the oneof type and never calls `VerifyFloats32`, allowing the infinity to persist. Reject values outside `[-math.MaxFloat32, math.MaxFloat32]` before this cast and add a struct-array overflow test.

MrPresent-Han (commented):
P1 validation follow-up.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -463,9 +513,328 @@ func printIndexes(indexes []*milvuspb.IndexDescription) []gin.H {
...
+		if depth > maxJSONDepth {
+			return merr.WrapErrParameterInvalidMsg(
+				"field %s nests deeper than %d levels, which the JSON engine cannot read",
+				field, maxJSONDepth)
+		}
+
+		switch {
+		case node.IsArray():
+			var err error
+			node.ForEach(func(_, value gjson.Result) bool {
```
MrPresent-Han: **P1 performance issue: could we avoid reparsing nested subtrees here?**

`gjson.ForEach` obtains each child through `parseSquash`, which scans the child's complete raw subtree; recursively invoking it makes a deep document with a wide leaf cost `O(depth × bytes)` rather than one linear pass, and the same pattern is used for `exprParams`. Since this runs before downstream size/quota checks, would it be possible to enforce the size limit first and use a single-pass scanner or reused parse result, with a wide-at-depth benchmark?

MrPresent-Han (commented):
Performance follow-up.

internal/distributed/proxy/httpserver/utils.go
xiaofan-luan: Fixed: both narrowing sites (the scalar Float column and the struct-array Float sub-field) now reject values outside the float32 range before the cast — `3.5e38` gets an explicit "outside the float32 range" error instead of storing +Inf. Note this was sharper than pre-existing: the earlier strict path parsed with `ParseFloat(raw, 32)`, whose own range error rejected these; the float64-dialect change had silently dropped that rejection. Covered by `TestFloatInsertRejectsFloat32Overflow`, including the struct sub-field case and the ±3.4e38 fits-boundary.

internal/distributed/proxy/httpserver/utils.go
xiaofan-luan: Fixed with a size-gated dual engine. Documents over 1KB take a single pass of the standard library tokenizer (sonic's Decoder has no Token) — the 500-levels-around-a-2KB-leaf shape drops from ~660µs to ~18µs (37x). Documents under the gate keep the zero-allocation gjson walk, whose amplification is then capped at gate × depth; the tokenizer allocates per token and ran ~2x the walk on ordinary values, so switching wholesale would have taxed every normal row for the adversarial one. `TestCheckEngineCompatibleEnginesAgree` runs the same shapes through both engines. exprParams keeps its recursive walk: its depth cap (default 100) bounds the amplification there.

internal/distributed/proxy/httpserver/request.go
xiaofan-luan: Real and pre-existing, agreed — Gin's YAML binding bypasses UnmarshalJSON hooks, so the null rules don't run for `application/x-yaml` on the low-level routes. Leaving this thread open as the follow-up marker rather than folding it in: the right fix is a decision (drop YAML support on these routes, or add format-independent validation after bind) that touches routes this PR otherwise leaves alone.

internal/distributed/proxy/httpserver/utils.go
xiaofan-luan: Simplified further per the author's call: the Float column now parses with `ParseFloat(raw, 32)` directly — one rounding, and the parser's own range error refuses 3.5e38, so the manual float32 bound is gone. The trade (a crafted midpoint literal stores a value the double-reading query paths cannot name) is documented in the code and pinned by `TestFloatInsertParsesToFloat32Directly`; no serializer emits such a literal. `TestFloatInsertRejectsFloat32Overflow` still covers ±3.5e38/1e39 rejection and the ±3.4e38 boundary.

internal/distributed/proxy/httpserver/request.go
xiaofan-luan: Decision from the author: won't fix. The low-level /api/v1 surface is legacy and is not slated for further investment; the YAML binding gap is acknowledged and recorded in this thread, but neither this PR nor a follow-up will take it on. If that surface ever gets attention again, the right move is most likely dropping the YAML binding from those routes outright rather than teaching every validation hook a second format. Resolving with this as the record.

PR #52262, opened 2026-08-06
fix: break out of MoveCursorForSingleChunk once the batch is full

### LINKED ISSUE #52258: [Bug]: MoveCursorForSingleChunk never breaks out of its loop, leaving the cursor at the segment end

### Environment

Found by code audit on `master` (verified present at `232823a008`).

### Current Behavior

`MoveCursorForSingleChunk` in `internal/core/src/segcore/SegmentChunkReader.h:148-170` never exits its loop once the batch is full:

```cpp
for (int64_t chunk_id = current_chunk_id; chunk_id < num_chunk; ++chunk_id) {
    auto chunk_size = ...;
    for (int64_t i = ...; i < chunk_size; ++i) {
        if (++processed_rows >= batch_size) {
            current_chunk_id = chunk_id;
            current_chunk_pos = i + 1;
            // no break / return
        }
    }
}
```

Once `processed_rows >= batch_size` becomes true it stays true for every remaining row, so the loops run to the end of the segment and the cursor is left pointing at the **last row of the segment** instead of at `batch_size`.

Two sibling implementations do this correctly:

- `MoveCursorForMultipleChunk` (same file, `:124-146`) computes the position with O(1) arithmetic.
- `MoveCursorForDataSingleChunk` (`internal/core/src/exec/expression/Expr.h:394-412`) has the `break`.

Only this one is missing it.

**Callers** (both on the `is_chunked() == false` path, i.e. growing segments):
- `internal/core/src/exec/expression/ColumnExpr.h:100`
- `internal/core/src/exec/expression/CompareExpr.h:233`

`MoveCursor()` is invoked by `ConjunctExpr` (`internal/core/src/exec/expression/ConjunctExpr.cpp:86`) for inputs that were short-circuited and therefore not evaluated in this batch.

### Expected Behavior

The cursor should advance by exactly `batch_size` rows.

### Impact

1. **Correctness** — after the first `MoveCursor()` the cursor sits at the segment end, so the next `Eval()` for that expression reads from the wrong offset. This affects `ColumnExpr` / `CompareExpr` on growing segments whenever a conjunction short-circuits.
2. **Performance** — a one-off O(rows-in-segment) spin.

### Steps To Reproduce

The logic is self-contained. Transcribing the loop verbatim and running it with a 1M-row segment, `SizePerChunk = 8192`, `batch_size = 8192`:

```
with the break (correct):  chunk=0   pos=8192  -> absolute row 8192
as written (no break):     chunk=122 pos=576   -> absolute row 1000000
```

The cursor is off by 991,808 rows — it lands exactly on the segment end.

### Milvus Log

N/A - static defect.

### Anything else?

Fix is a one-line `return;`, matching `Expr.h:409`. PR to follow.

xiaofan-luan:
Fixes #52258

## What

`MoveCursorForSingleChunk` never left its loop once the batch was full:

```cpp
for (int64_t i = ...; i < chunk_size; ++i) {
    if (++processed_rows >= batch_size) {
        current_chunk_id = chunk_id;
        current_chunk_pos = i + 1;
        // no break
    }
}
```

The condition stays true for every remaining row, so the loops run to the end of
the segment and the cursor is left on the **last row of the segment** rather than
on `batch_size`.

Two sibling implementations already stop correctly:
- `MoveCursorForMultipleChunk` (same class) — O(1) arithmetic.
- `MoveCursorForDataSingleChunk` (`exec/expression/Expr.h:394-412`) — has the `break`.

## Why it matters

**Correctness.** Callers are `ColumnExpr.h:100` and `CompareExpr.h:233`, both on the
`is_chunked() == false` (growing segment) path. `MoveCursor()` is invoked by
`ConjunctExpr.cpp:86` for inputs that were short-circuited in the current batch.
After the first such call the cursor sits at the segment end, so the next `Eval()`
for that expression reads from the wrong offset.

**Performance.** A one-off O(rows-in-segment) spin — ~1M iterations for a 1M-row
segment.

Transcribing the loop verbatim and running it (1M rows, `SizePerChunk`=8192,
`batch_size`=8192):

```
with the break (correct):  chunk=0   pos=8192  -> absolute row 8192
as written (no break):     chunk=122 pos=576   -> absolute row 1000000
```

## How

One-line `return;`, matching `Expr.h:409`.

## Tests

Added `SegmentChunkReader.MoveCursorForSingleChunkStopsAtBatchBoundary`:
asserts the cursor advances by exactly `batch_size`, that it does not land on the
segment end, and that a second call continues from where the first stopped
(`2 * batch_size`). The test fails on the current code — it would observe
`absolute_offset == active_count`.

## Verification performed

I could not run the C++ unit test binary locally (no full segcore build in this
environment). What I did verify:

- Both the modified header and the new test compile cleanly (`clang++ -fsyntax-only`)
  using the exact flags from this repo's `compile_commands.json`.
- The loop's behaviour was reproduced by transcribing it verbatim into a standalone
  program, which produced the numbers above.

CI must run the actual test.

### DIFF
--- a/internal/core/src/segcore/SegmentChunkReader.h
+++ b/internal/core/src/segcore/SegmentChunkReader.h
@@ -164,6 +164,7 @@ class SegmentChunkReader {
                 if (++processed_rows >= batch_size) {
                     current_chunk_id = chunk_id;
                     current_chunk_pos = i + 1;
+                    return;
                 }
             }
         }
--- a/internal/core/src/segcore/SegmentChunkReaderTest.cpp
+++ b/internal/core/src/segcore/SegmentChunkReaderTest.cpp
@@ -15,10 +15,14 @@
 // limitations under the License.
 
 #include <cstdint>
+#include <memory>
 #include <string>
 
 #include "gtest/gtest.h"
+#include "common/IndexMeta.h"
+#include "common/Schema.h"
 #include "segcore/SegmentChunkReader.h"
+#include "segcore/SegmentGrowingImpl.h"
 
 namespace milvus::segcore {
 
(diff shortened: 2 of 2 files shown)


xiaofan-luan:

```
```

czs007:

Verified:

- `internal/core/src/segcore/SegmentChunkReader.h:167` — the added `return` sits inside `if (++processed_rows >= batch_size)`, exiting both nested loops after the cursor is committed, so later iterations can no longer overwrite it; this matches the sibling implementations `MoveCursorForDataSingleChunk` and `MoveCursorForMultipleChunk` in `exec/expression/Expr.h`.
- `internal/core/src/exec/expression/CompareExpr.h:233` and `internal/core/src/exec/expression/ColumnExpr.h:100` — these are the only two callers (whole-repo grep), both on the `is_chunked() == false` path, and both interpret the resulting `(chunk_id, chunk_pos)` as the same absolute offset, so the fix's scope is exactly the growing-segment path the PR describes.
- `internal/core/src/exec/expression/ColumnExpr.cpp:202` and `CompareExpr.cpp:248` — a cursor parked at a chunk boundary (`pos == chunk_size`) is accepted by the consuming loops, and `GetCurrentRows()` (`chunk_id * size + pos`) still yields the correct absolute offset in that representation.
- `internal/core/src/exec/expression/Expr.cpp:864` and line 887 — optimizer reordering places the numeric predicate before `PhyCompareFilterExpr`, so the short-circuit path exercised by the new test (`ConjunctExpr` skipping a fully-false batch) is the same one production hits.
- `internal/core/src/segcore/SegmentChunkReaderTest.cpp:84` and line 93 — the helper test advances a full chunk and then repeats, covering continuation from a chunk-end cursor; pre-fix the cursor would land at `active_count` and the assertion fails, so the test genuinely reproduces the bug.
- `internal/core/src/segcore/SegmentChunkReaderTest.cpp:112` and line 120 — both process-global settings (`EXEC_EVAL_EXPR_BATCH_SIZE`, segcore chunk rows) are restored via scoped guards, and the production change adds no shared mutable state.
- `internal/core/unittest/CMakeLists.txt:32` — the `src/**/*Test.cpp` glob pulls the new test file into the unittest binary, which supplies the `test_utils/` include path, so the cross-directory includes build.

czs007 (approved):
/lgtm

PR #52267, opened 2026-08-06
fix: return error when RollbackImport ACK arrives before replicated import job exists [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #52254: [Bug]: RollbackImport ACK can be consumed before replicated import job exists
## What happened

In the replicated import 2PC path, `rollbackImportV2AckCallback` currently treats a missing import job as success. The broadcaster marks an ACK callback complete when the callback returns nil, so a RollbackImport message that arrives before the replicated Import message creates the local job can be consumed permanently.

`commitImportV2AckCallback` already handles the same ordering race by returning `ErrImportSysFailed` so the broadcaster retries until the job exists.

## Impact

A replicated `auto_commit=false` import can be aborted on the source cluster while the target cluster has not created the matching import job yet. The target consumes the rollback ACK, later creates the job, and then waits in `Uncommitted` for an explicit commit/abort that will never be replayed. That can strand imported import state on the peer.

## Reproduction / proof

A targeted regression test fails on current master:

```bash
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord -run TestRollbackImportCallback_MissingJob_Retry
```

Failure before the fix:

```text
RollbackImport: job not found, skipping
Error: An error is expected but got nil.
```

## Expected behavior

If the replicated import job is not found yet, rollback ACK should return a system-side import error and let the broadcaster retry, matching CommitImport ACK behavior.

## Notes

No duplicate open issue or PR was found for this specific RollbackImport missing-job ACK race. Related BM25 reopen PRs found during the search are unrelated to this import callback path.

waterWang:
Fixes #52254

## Problem
In the replicated import 2PC path, `rollbackImportV2AckCallback` returns `nil` when the import job does not exist yet. The broadcaster marks an ACK callback complete when the callback returns nil, so a RollbackImport message that arrives before the replicated Import message creates the local job can be consumed permanently.

After consumption, the target later creates the job and waits in `Uncommitted` for an explicit commit/abort that will never be replayed.

## Fix
Return `merr.WrapErrImportSysFailedMsg` when the job is not found, mirroring the exact pattern in `commitImportV2AckCallback`. This keeps the broadcast task alive so the callback retries after the import task finishes writing local meta.

## Change
`internal/datacoord/ddl_callbacks_import.go`: replace `return nil` with `return merr.WrapErrImportSysFailedMsg(...)` in the nil-job guard of `rollbackImportV2AckCallback`.

### DIFF
--- a/internal/datacoord/ddl_callbacks_import.go
+++ b/internal/datacoord/ddl_callbacks_import.go
@@ -297,8 +297,8 @@ func (c *DDLCallbacks) rollbackImportV2AckCallback(ctx context.Context, result m
 
 	job := c.importMeta.GetJob(ctx, jobID)
 	if job == nil {
-		mlog.Warn(ctx, "RollbackImport: job not found, skipping", mlog.FieldJobID(jobID))
-		return nil
+		mlog.Info(ctx, "RollbackImport: job not found, retry later", mlog.FieldJobID(jobID))
+		return merr.WrapErrImportSysFailedMsg("job %d not found, waiting for import job creation", jobID)
 	}
 	state := job.GetState()
 	if state == internalpb.ImportJobState_Committing ||


internal/datacoord/ddl_callbacks_import.go
```diff
@@ -297,8 +297,8 @@ func (c *DDLCallbacks) rollbackImportV2AckCallback(ctx context.Context, result m
 
 	job := c.importMeta.GetJob(ctx, jobID)
 	if job == nil {
-		mlog.Warn(ctx, "RollbackImport: job not found, skipping", mlog.FieldJobID(jobID))
-		return nil
+		mlog.Info(ctx, "RollbackImport: job not found, retry later", mlog.FieldJobID(jobID))
+		return merr.WrapErrImportSysFailedMsg("job %d not found, waiting for import job creation", jobID)
```
yhmo, line 301: internal/datacoord/ddl_callbacks_import.go line:301
Medium ---- This retries every missing job forever, but `importV1AckCallback` explicitly treats `ErrCollectionNotFound` as a terminal skip and creates no job. If the matching Import ACK took that path, a later RollbackImport can never satisfy this callback; `callMessageAckCallbackUntilDone` keeps the collection resource-key lock while retrying, potentially blocking subsequent broadcasts indefinitely. Please distinguish “job creation still pending” from terminal absence and add regression coverage for both the transient ordering case and the collection-not-found case.

Issue #52263, opened 2026-08-06
[Enhancement]: No way to evaluate whether a performance optimization is worth doing (component-level benchmark infrastructure)


xiaofan-luan:
### Is this a feature request?

Yes — infrastructure / RFC. Looking for opinions before anyone builds it.

### Problem

There is currently no supported way to answer **"is this optimization worth doing?"** at the component level.

I recently audited the read path, write path, filter/expression execution, reduce, segcore, and knowhere for optimization opportunities. Every single candidate stalled at the same place: the numbers could not be produced with anything that exists in the tree. Concretely:

| Area | Benchmark coverage |
|---|---|
| `internal/core/` (exec 159 files, segcore 145, index 83, common 112, storage 112) | **1** Google Benchmark total — `FastMemBenchmark.cpp`, which measures `memcpy` |
| `internal/streamingnode/` + `pkg/streaming/` + `pkg/mq/` (277 files, ~41k lines) | **1** benchmark, and it covers the legacy consume path, not the write path |
| proxy / querynode reduce, `typeutil.AppendFieldData`, requery assembly, insert repack | none |
| `storage.MergeSort` (the default mix-compaction path), writebuffer, syncmgr, zstd codec | none |

To get any numbers at all I had to write throwaway benchmarks, run them, and delete them.

Two secondary problems make it worse:

1. **Some test packages do not compile on `master`**, so even the benchmarks that do exist cannot run:
   - `internal/datacoord` — `meta_test.go:2362` calls `getSegmentBinlogFields`, which does not exist outside comments
   - `internal/proxy` — `authentication_interceptor_test.go` references `hookutil.SetMockAPIHook`
   - `internal/querynodev2/delegator` — `delegator_test.go` references `streaming.SetupNoopWALForTest`

   The delegator package is where `BenchmarkQueryRequest_ShallowCopy` lives — 8 benchmarks that currently cannot be executed.

2. **Existing benchmarks run at toy scale.** `BenchmarkCatalog_List1000Segments` uses 1000 segments; the interesting behaviour starts around 100k. `Benchmark_SegmentWriter_BatchSize_*` uses dim=128 while production is commonly 768/1024.

### Why toy-scale benchmarks are actively harmful here

Not just "less useful" — they produce **confidently wrong answers**, because conclusions flip with the workload shape. Measured, 10k rows:

| Schema | `proto.Unmarshal` | `fastpb` | Verdict |
|---|---|---|---|
[...]

PR #52268, opened 2026-08-06
enhance: build the GIS segment-offsets array only for GIS expressions

### LINKED ISSUE #52265: [Enhancement]: Chunked filter path builds a GIS-only offsets array for every expression (guard present in the sibling function, missing here)

Performance. This one is a straightforward inconsistency rather than a design question.

### Problem

`ProcessDataChunksForMultipleChunk` unconditionally allocates and fills a
`segment_offsets_array` that only GIS expressions ever read.

`internal/core/src/exec/expression/Expr.h:1647-1655`
```cpp
if (size == 0)
    continue;
std::vector<int32_t> segment_offsets_array(size);          // allocated for everyone
auto start_offset =
    segment_->num_rows_until_chunk(field_id_, i) + data_pos;
for (int64_t j = 0; j < size; ++j) {                        // filled for everyone
    int64_t offset = start_offset + j;
    segment_offsets_array[j] = static_cast<int32_t>(offset);
}
```

Every consumer of that array sits behind `if constexpr (NeedSegmentOffsets)`
(`:1670`, `:1707`, `:1759`).

**The sibling function in the same file already does this correctly.**
`ProcessDataChunksForSingleChunk` builds the array *inside* the guard, in both of its
branches (`:1525-1531` and `:1561-1570`):

```cpp
if constexpr (NeedSegmentOffsets) {
    // For GIS functions: construct segment offsets array
    std::vector<int32_t> segment_offsets_array(size);
    ...
```

Only the multi-chunk path is missing the guard.

### Scope

`NeedSegmentOffsets = true` has exactly one caller in the tree:
`internal/core/src/exec/expression/GISFunctionFilterExpr.cpp` (`:84`, `:135`, `:185`).

Everything else — UnaryExpr, TermExpr, BinaryRangeExpr, BinaryArithOpEvalRange,
JsonContains, Exists, TimestamptzArithCompare — uses the default `false`. And
`ChunkedSegmentSealedImpl::is_chunked()` returns `true` unconditionally
(`ChunkedSegmentSealedImpl.h:493-496`), so **this is the path every sealed-segment
raw-data filter takes**.

### Accounting

Scenario: 1M-row INT64 column, `field > 100`, `EXEC_EVAL_EXPR_BATCH_SIZE = 8192`
(`common/Consts.h:90`).

```
batches            = 1,000,000 / 8192              ~= 123          [estimate]
per batch          = one operator new(32KB)
                     + zero-init 32KB
                     + 8192 int32 stores
wasted store traffic = 4 MB zero-init + 4 MB fill  =  8 MB
8 MB @ ~16 GB/s (L1/L2 resident)                   ~= 0.5 ms       [estimate]

useful work for the same query:
read 1M x 8B = 8 MB, SIMD compare, write a 128KB bitmap
8 MB @ 10-16 GB/s                                  ~= 0.5-0.8 ms   [estimate]
```

The waste is the **same order of magnitude as the actual scan**. For narrow columns
(`INT8`/`INT16`/`BOOL`: 1–2 B/row of data versus 4 B/row of offsets) the discarded
writes exceed the bytes actually scanned.

Estimated saving on the scan phase: **20–40%** (derivation:
`0.5 / (0.5 + 0.65) ~= 43%`, rounded down because stores overlap with loads and are
cheaper than loads). Highly sensitive to column width — on VARCHAR/JSON columns,
where per-row work is much larger, it drops to single digits.

### Suggested fix

Wrap `:1649-1655` in `if constexpr (NeedSegmentOffsets)` and pass `nullptr` on the
non-GIS branch — i.e. make it match the sibling function. Under 10 lines. The GIS path
must stay byte-for-byte identical.

### How to verify

There is currently **no** Google Benchmark covering this. `internal/core/benchmark/`
contains exactly one target, `fastmem_benchmark`, which measures `memcpy`. A benchmark
would have to be added — 1M-row sealed chunked segment, parameterised over column type
(INT8/INT64/VARCHAR), driven through `PhyUnaryRangeFilterExpr::Eval`.

Zero-change evidence in the meantime: `perf record` an `*ExprRange*` unit test and look
for `operator new` / `__memset` under `ProcessDataChunksForMultipleChunk`.

- A profile shows `ProcessDataChunksForMultipleChunk` under ~5% CPU — e.g. a workload
  where filtering goes through `ProcessIndexChunks` (scalar index) rather than raw data.
- Workloads dominated by VARCHAR/JSON filtering, where the ratio falls under ~2%.

### Honest caveat

The 20–40% is an **estimate with the derivation shown above, not a measurement** — I
could not run the C++ benchmark suite locally. What is not an estimate is the structural
fact: the guard exists in the sibling function and is missing here, and GIS is the sole
consumer.

xiaofan-luan:
Fixes #52265

## What

`ProcessDataChunksForMultipleChunk` built `segment_offsets_array` unconditionally,
while every consumer of it is behind `if constexpr (NeedSegmentOffsets)`:

```cpp
if (size == 0) continue;
std::vector<int32_t> segment_offsets_array(size);        // for everyone
auto start_offset = segment_->num_rows_until_chunk(field_id_, i) + data_pos;
for (int64_t j = 0; j < size; ++j) {                     // for everyone
    segment_offsets_array[j] = static_cast<int32_t>(start_offset + j);
}
...
if constexpr (NeedSegmentOffsets) {                      // only the *use* was guarded
    func(..., segment_offsets_array.data(), ...);
}
```

`NeedSegmentOffsets = true` has exactly one caller in the tree —
`GISFunctionFilterExpr.cpp` (`:84`, `:135`, `:185`). UnaryExpr, TermExpr,
BinaryRangeExpr, BinaryArithOpEvalRange, JsonContains, Exists and
TimestamptzArithCompare all use the default `false`, and for all of them this array is
written and never read.

**The sibling function already does this correctly.**
`ProcessDataChunksForSingleChunk` builds the array inside the guard, in both branches
(`:1525-1531`, `:1561-1570`). This change makes the multi-chunk path match it.

## Why it matters

`ChunkedSegmentSealedImpl::is_chunked()` returns `true` unconditionally
(`ChunkedSegmentSealedImpl.h:493-496`), so the multi-chunk path is what every
sealed-segment raw-data filter takes.

For a 1M-row INT64 scan with the default `EXEC_EVAL_EXPR_BATCH_SIZE = 8192`
(~123 batches), the discarded work is one 32KB allocation per batch plus ~8 MB of
store traffic (4 MB zero-init + 4 MB fill) — the same order of magnitude as the ~8 MB
the scan actually reads. On narrow columns (`INT8`/`INT16`/`BOOL`, 1–2 B of data per
row against 4 B of offsets per row) the discarded writes exceed the bytes scanned.

## How

Declare the vector empty and populate it inside the guard. It stays declared outside
because the three call sites are in different branches; a default-constructed
`std::vector` performs no allocation.

Hoisting `num_rows_until_chunk()` into the guard as well means non-GIS expressions no
longer make that call per batch either.

## Verification performed

I could not build segcore locally, so **CI must validate this**. What I did check:

- Every use of `segment_offsets_array` in the function is inside
  `if constexpr (NeedSegmentOffsets)` — the only unguarded statement left is the empty
  declaration.
- Compiled the resulting pattern (template with a `bool` non-type parameter, empty
  vector conditionally resized, consumed at three guarded sites) as a standalone TU
  with the same toolchain (`clang++ -std=c++17 -O2 -Wall -Wextra`). Both
  instantiations compile clean and produce correct results.
- Inspected the generated assembly for the `NeedSegmentOffsets = false` instantiation:
  **no `operator new`, no `memset`** — the allocation and the zero-init are gone, not
  merely cheap.

Full-TU syntax checking against this repo's `compile_commands.json` was not conclusive
because the generated `internal/core/src/pb/plan.pb.h` in my build tree predates
`BloomFilterExpr` on master; unmodified master fails identically, so that error is
environmental rather than introduced here.

## What is not claimed

The issue estimates a 20–40% saving on the filter scan phase, with the derivation shown
there. That is an **estimate, not a measurement** — `internal/core/benchmark/` contains
exactly one target (`fastmem_benchmark`, which measures `memcpy`), so there is nothing
in-tree that can measure this today. See #52263.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -1646,12 +1646,22 @@ class SegmentExpr : public Expr {
 
             if (size == 0)
                 continue;  //do not go empty-loop at the bound of the chunk
-            std::vector<int32_t> segment_offsets_array(size);
-            auto start_offset =
-                segment_->num_rows_until_chunk(field_id_, i) + data_pos;
-            for (int64_t j = 0; j < size; ++j) {
-                int64_t offset = start_offset + j;
-                segment_offsets_array[j] = static_cast<int32_t>(offset);
+            // Only GIS functions consume the segment offsets. Building the
+            // array unconditionally costs an allocation plus size int32 stores
+            // per batch for every other expression, and the result is never
+            // read. Default-constructing the empty vector is free; it stays in
+            // scope because the three call sites below are in different
+            // branches. Cf. ProcessDataChunksForSingleChunk, which already
+            // builds it inside the guard.
+            std::vector<int32_t> segment_offsets_array;
+            if constexpr (NeedSegmentOffsets) {
+                segment_offsets_array.resize(size);
+                auto start_offset =
+                    segment_->num_rows_until_chunk(field_id_, i) + data_pos;
+                for (int64_t j = 0; j < size; ++j) {
+                    int64_t offset = start_offset + j;
+                    segment_offsets_array[j] = static_cast<int32_t>(offset);
+                }
             }
             auto skip_index = segment_->GetSkipIndex();
             if (!skip_func || !skip_func(*skip_index, field_id_, i)) {


xiaofan-luan:

```
```

czs007:

**Verified:**

- `internal/core/src/exec/expression/Expr.h:1647-1665` — the positive-size guard is intact and the `NeedSegmentOffsets=true` path preserves the exact previous offset calculation (`num_rows_until_chunk(field_id_, i) + data_pos`), so populated values are byte-for-byte identical.
- `internal/core/src/exec/expression/Expr.h:1670-1674,1707-1712,1759-1763` — all three readers of the offsets array sit inside `if constexpr (NeedSegmentOffsets)`, and the new guard is placed *before* `GetSkipIndex()`, so the SkipIndex-skipped branch still receives a fully populated array.
- `internal/core/src/exec/expression/GISFunctionFilterExpr.cpp:84,135,185` — a repo-wide grep over `internal/core/src/` confirms these are the only `NeedSegmentOffsets=true` instantiations; every other expression type uses the defaulted `false`.
- `internal/core/src/exec/expression/Expr.h:1801-1817` — the dispatcher forwards the template flag unchanged to both the single-chunk and multi-chunk implementations.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:3314-3321` — the call omitted on the non-GIS path is a read-only snapshot lookup with no mutation or lock-ordering side effects; the GIS path retains identical allocation and failure behavior.
- `internal/core/unittest/test_scorer.cpp:417` — `ComputeScorerScoresGISFilterCoversAllBatches` drives a sealed-segment GIS filter with `N = 10000 > EXEC_EVAL_EXPR_BATCH_SIZE`, asserting offsets 9000/9001/9998, giving real regression coverage across the batch boundary on the multi-chunk path.

czs007 (approved):
/lgtm

PR #52270, opened 2026-08-06
enhance: resolve the destination column by index in AppendFieldData

### LINKED ISSUE #52264: [Enhancement]: typeutil.AppendFieldData rebuilds a field map on every row (31 call sites across search reduce, query reduce, and insert repack)

Performance. Found by audit; the numbers below are measured unless labelled otherwise.

### Problem

`typeutil.AppendFieldData` rebuilds a `map[int64]*schemapb.FieldData` on **every call**, and every caller invokes it **per row**:

`pkg/util/typeutil/schema.go:1014-1020`
```go
func AppendFieldData(dst, src []*schemapb.FieldData, idx int64, fieldIdxs ...int64) (appendSize int64) {
	dstMap := make(map[int64]*schemapb.FieldData)
	for _, fieldData := range dst {
		if fieldData != nil {
			dstMap[fieldData.FieldId] = fieldData
		}
	}
```

`dst` is loop-invariant across the whole reduce loop, so this map is rebuilt identically for every row.

### Why this one is worth looking at

It has the broadest reach of anything I found — **31 non-test call sites on master**, spanning three independent hot paths, on both the read and the write side:

| Path | Call sites |
|---|---|
| QueryNode search reduce | `internal/querynodev2/segments/search_reduce.go:154,388,454` |
| Proxy search / query reduce | `internal/proxy/search_reduce_util.go:717,254,378`, `internal/proxy/task_query.go:1260` |
| Proxy requery assembly | `internal/proxy/search_pipeline.go:1627,1759,557` |
| **Insert repack** | `internal/proxy/msg_pack.go:90` — once per inserted row |

### Measured

Isolated the map rebuild, 10,000 rows × 5 fields (Apple M4 Pro, go1.26; x86 servers are typically 1.5–2× slower, so treat the ratio, not the absolute):

```
rebuild per row   321,841 ns   0 B   0 allocs    (= 32.2 ns/row)
build once         14,251 ns   0 B   0 allocs    (=  1.4 ns/row)
                      22.6x
```

Note the **0 allocs**: escape analysis (`go build -gcflags='-m'`) reports
`schema.go:1015: make(map[int64]*schemapb.FieldData) does not escape`. The map is
stack-allocated, so this is pure CPU, not GC pressure. (I initially assumed it was a
per-row heap allocation; it is not.)

For nq=10, topk=100 that is ~250 µs of pure overhead, against ~10 µs of actually
useful work when the output is scalar-only — the map costs roughly 25× what the data
movement costs.

### Suggested fix

The correct shape already exists in this repo: the query path uses
`typeutil.AppendFieldDataByColumn` (`schema.go:1312`) via
`internal/util/queryutil/helpers.go:68`, which selects rows first and then moves
columns in bulk. The search path never followed.

Smallest change: hoist `dstMap` into the caller, or add an
`AppendFieldDataBatch(dst, src, rowIdxs)` variant. Larger change: move the search
path to the column-wise form the query path already uses.

### How to verify

There is no benchmark for this today. `pkg` is a separate module and `go vet ./util/typeutil/` is clean, so a benchmark can be added there without touching the packages that currently fail to build.

Suggested shape: 10k rows, 5 output fields including one 768-dim float vector, measured with and without the per-row rebuild.

- A profile shows `AppendFieldData` + `runtime.makemap*` under ~3% cumulative.
- Deployments where `nq × topk` is routinely under ~50 rows — the fixed cost stops mattering.
- Single output field: the map does roughly 1/3 as much work, so the gain shrinks accordingly.

### Honest caveat

The 22.6× is for the isolated operation. **What fraction of end-to-end latency this represents is unknown** — there is no in-process profile, and QueryNode's search work is largely in C++ segcore where Go pprof loses samples across the cgo boundary. Establishing the Go/C++ split is the prerequisite for sizing this properly. See #52263.

xiaofan-luan:
Fixes #52264

## What

`AppendFieldData` copies **one row**, so reduce loops call it once per row with the same
`dst`. It opened by building a `FieldId -> *FieldData` map over `dst`:

```go
func AppendFieldData(dst, src []*schemapb.FieldData, idx int64, fieldIdxs ...int64) (appendSize int64) {
	dstMap := make(map[int64]*schemapb.FieldData)
	for _, fieldData := range dst {
		if fieldData != nil {
			dstMap[fieldData.FieldId] = fieldData
		}
	}
```

`dst` is loop-invariant across those calls, so that identical map was rebuilt for every
row. There are **31 non-test call sites** on master, spanning QueryNode search reduce,
proxy search/query reduce, requery assembly, and insert repack.

## How

`dst` and `src` are index-parallel in every caller — the lazy-creation branch already
depends on that when it does `dst[i] = dstFieldData`. So the destination column is
normally just `dst[i]`, guarded by a `FieldId` equality check. The map is built lazily,
only when that check fails for some column, which preserves the old behaviour for
permuted or mismatched destinations.

No caller changes and no signature change, so all 31 sites benefit.

## Measured

10,000 rows, Apple M4 Pro / go1.26.5 (`benchstat`-style median of 5; treat the ratio,
not the absolute):

| destination shape | before | after | | per row |
|---|---|---|---|---|
| 4 scalar columns | 0.518 ms | 0.296 ms | **1.75×** | −22 ns |
| 3 scalars + one 768-dim vector | 1.524 ms | 1.098 ms | **1.39×** | −43 ns |

The ratio depends on how much data each row moves — with a 3 KB vector per row, data
movement dominates and the lookup is a smaller share. The per-row saving is the stable
part.

**Correction to the issue text:** #52264 quotes 22.6×. That figure was for the map
rebuild measured *in isolation* (32.2 → 1.4 ns/row), not for `AppendFieldData` as a
whole, and reads as more than it is. The end-to-end numbers are the table above. I have
corrected the issue.

Allocation counts are unchanged: escape analysis already showed the map did not escape
(`schema.go:1015: make(map[int64]*schemapb.FieldData) does not escape`), so this is pure
CPU, not GC pressure.

## Tests

Added `TestAppendFieldDataResolvesSameColumnAsFieldIdLookup`, covering the three shapes
that exercise both paths:

- **parallel dst** — the fast path; verifies scalar and vector columns accumulate
  correctly across rows.
- **nil dst entries** — the lazy-creation path used by insert repack; verifies the
  entries are created with the right `FieldId` and the data lands correctly.
- **dst permuted relative to src** — forces the map fallback; verifies data still lands
  in the column with the matching `FieldId` rather than by position.

Full `pkg/util/typeutil` suite passes.

## What is not claimed

These are component-level numbers. **What share of end-to-end latency this represents
has not been measured** — QueryNode's search work is largely in C++ segcore, where Go
pprof loses samples across the cgo boundary. See #52263.

### DIFF
--- a/pkg/util/typeutil/append_field_data_test.go
+++ b/pkg/util/typeutil/append_field_data_test.go
@@ -43,6 +43,21 @@ func floatVec(fieldID int64, name string, dim int64, data []float32) *schemapb.F
 	}
 }
 
+// A dst shorter than src violates the caller contract. The pre-existing code
+// wrote dst[i] unguarded and panicked; keep it a hard failure so a broken
+// caller cannot silently drop a column.
+func TestAppendFieldDataPanicsOnShortDst(t *testing.T) {
+	src := []*schemapb.FieldData{
+		scalarInt64(100, "pk", []int64{10}),
+		scalarInt64(101, "age", []int64{20}),
+	}
+	dst := make([]*schemapb.FieldData, 1) // one column short
+
+	assert.PanicsWithValue(t,
+		"AppendFieldData: dst has 1 columns, src has 2; callers must size dst to at least len(src)",
+		func() { AppendFieldData(dst, src, 0) })
+}
+
 // AppendFieldData now resolves the destination column by index and only falls
 // back to a FieldId map when dst and src are not parallel. This checks the two
 // paths agree, including the cases that force the fallback.
--- a/pkg/util/typeutil/append_field_data_test.go
+++ b/pkg/util/typeutil/append_field_data_test.go
@@ -1,0 +1,135 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package typeutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
+)
+
+func scalarInt64(fieldID int64, name string, data []int64) *schemapb.FieldData {
+	return &schemapb.FieldData{
+		Type: schemapb.DataType_Int64, FieldId: fieldID, FieldName: name,
+		Field: &schemapb.FieldData_Scalars{Scalars: &schemapb.ScalarField{
+			Data: &schemapb.ScalarField_LongData{LongData: &schemapb.LongArray{Data: data}},
+		}},
+	}
+}
+
+func floatVec(fieldID int64, name string, dim int64, data []float32) *schemapb.FieldData {
+	return &schemapb.FieldData{
+		Type: schemapb.DataType_FloatVector, FieldId: fieldID, FieldName: name,
+		Field: &schemapb.FieldData_Vectors{Vectors: &schemapb.VectorField{
+			Dim:  dim,
+			Data: &schemapb.VectorField_FloatVector{FloatVector: &schemapb.FloatArray{Data: data}},
+		}},
+	}
+}
+
+// AppendFieldData now resolves the destination column by index and only falls
+// back to a FieldId map when dst and src are not parallel. This checks the two
+// paths agree, including the cases that force the fallback.
+func TestAppendFieldDataResolvesSameColumnAsFieldIdLookup(t *testing.T) {
+	const rows = 4
+	newSrc := func() []*schemapb.FieldData {
+		return []*schemapb.FieldData{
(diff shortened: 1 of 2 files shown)


xiaofan-luan:

```
```

czs007:

Verified:

- `pkg/util/typeutil/schema.go:1034` — the index fast path is guarded by `i < len(dst) && dst[i] != nil && dst[i].FieldId == fieldData.FieldId`, so it is taken only on an exact FieldId match and cannot bind a source column to the wrong destination.
- `pkg/util/typeutil/schema.go:1037` — on any mismatch the lazily built `FieldId -> last non-nil FieldData` map reproduces the previous lookup verbatim, and because lazy creation of `dst[i]` can only happen after a map miss, the map's staleness is iteration-for-iteration equivalent to the old code.
- `pkg/util/typeutil/schema.go:1060` — a destination shorter than the source still fails hard at the same point; master's unguarded `dst[i] =` panicked with index-out-of-range there, so net behavior is unchanged apart from a clearer message.
- Call-site invariants — every production caller supplies `dst` with `len(dst) >= len(src)`: exact-sized nil slices at `internal/proxy/msg_pack.go:56` and `pkg/mq/msgstream/msg.go:286`, and `PrepareResultFieldData`-derived slices at `internal/proxy/search_reduce_util.go:135` and `internal/proxy/task_query.go:1241`.
- `pkg/util/typeutil/schema.go:1024` — `dstMap` is a call-local variable, adding no shared cache or mutable global state; writes to `dst` and its elements follow the same single-goroutine pattern as before.
- `pkg/util/typeutil/append_field_data_test.go:49` — the new tests cover the short-destination panic, the index fast path, nil-entry lazy creation, and the permuted-destination map fallback, with the pre-existing reordering coverage in `pkg/util/typeutil/schema_test.go` still in place.

czs007 (approved):
/lgtm

Issue #52271, opened 2026-08-06
[Bug]: master does not typecheck: internal/datacoord/meta_test.go references undefined getSegmentBinlogFields


czs007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

master @ `8a1470e202`

### Current Behavior

`internal/datacoord` does not typecheck on master, so `make static-check`, `go vet` and `go test` fail for everyone:

```
vet: internal/datacoord/meta_test.go:2362:13: undefined: getSegmentBinlogFields
```

`getSegmentBinlogFields` does not exist anywhere in the tree.

### Expected Behavior

master typechecks.

### Root cause: a semantic conflict between two PRs that were each green

This is not a mistake in either PR — two independently-correct changes break only in combination, and CI never tested the combination.

| When | What |
|---|---|
| 07-29 | #52006 forks from master at `113c23613a`. `getSegmentBinlogFields` exists in `index_inspector.go`, and the PR adds a call to it in `meta_test.go`. |
| 07-31 | #51768 merges. It replaces the binlog-fields index-eligibility check with a schema-version gate and **deletes** `getSegmentBinlogFields`, updating its own `index_inspector_test.go` accordingly. |
| 08-03 | #52006's last CI run (`ed1766e3e`) is green — correctly so: on that branch the function still exists. |
| 08-06 | #52006 is merged. The merge is a **rebase** (`95e3770aff` has a single parent), replaying the PR onto a master where the function is already gone. CI is not re-run on the rebased tree. |

The two PRs touch disjoint files (`index_inspector.go` vs `meta_test.go`), so git reports no conflict; only the Go compiler does.

The general gap: **a PR's CI result is only valid for the tree it ran on.** When tide rebases before pushing, the pushed tree is one that no CI job has ever seen. A merge queue — or re-running against the rebased tree before pushing — would have caught this.

### Steps To Reproduce

```
git checkout master
go vet -tags dynamic,test ./internal/datacoord/
```

### Milvus Log

n/a

PR #52274, opened 2026-08-06
fix: accumulate RetrieveResultCache size instead of recomputing it per merge

### LINKED ISSUE #52273: [Enhancement]: RetrieveResultCache.merge recomputes proto.Size over the whole accumulated message (O(N^2) per QueryStream)

Performance, with a stray debug statement alongside it.

### Problem

`RetrieveResultCache.merge` recomputes the size of the **entire accumulated message** on
every merge:

`internal/util/streamrpc/streamer.go:94`
```go
func (c *RetrieveResultCache) merge(result *internalpb.RetrieveResults) {
	c.result.GetIds().GetIntId().Data = append(c.result.GetIds().GetIntId().GetData(), result.GetIds().GetIntId().GetData()...)
	...
	c.size = proto.Size(c.result)      // walks everything accumulated so far
}
```

Over a stream of N results this is O(N²) in the total number of IDs: the k-th merge
walks k×R elements, so the whole stream touches `R·N(N+1)/2` instead of `R·N`. At
N=200 that is 100.5M element visits instead of 1M — a redundancy factor of `(N+1)/2`.

### Path

`streamer.go:171 Send` ← `internal/querynodev2/tasks/query_stream_task.go:72` ←
`internal/querynodev2/handlers.go:344` ← `internal/querynodev2/services.go:1105
QueryStream` ← `internal/proxy/task_delete.go:503`, which is the only non-test caller —
so this is **delete-by-expression**.

The producer side (`internal/querynodev2/segments/retrieve.go:141`) does one `Send` per
segment from one goroutine per segment (`:125-127`), and they all serialize on
`s.mu.Lock()` (`streamer.go:172`). So the quadratic term sits inside the lock that every
segment's result has to pass through.

### Measured

Real code, 5000 IDs per result (Apple M4 Pro, go1.26.5, median of 3):

| results merged | before | after | |
|---|---|---|---|
| 50 | 3.37 ms | 0.93 ms | 3.6× |
| 100 | 10.68 ms | 1.71 ms | 6.3× |
| 200 | 39.81 ms | 3.24 ms | **12.3×** |

Scaling confirms the shape: doubling N multiplies the old time by ~3.2 and ~3.7
(approaching 4× — quadratic), and the new time by ~1.8 and ~1.9 (linear).

### Suggested fix

Accumulate instead of recomputing. `merge` only carries over the incoming message's IDs
and counters and drops the rest of its envelope, so `c.size += proto.Size(result)` can
only **over**-estimate — never under-estimate, which matters because
`ResultCacheServer.Send` relies on this value to decide whether a message would exceed
`maxMsgSize`.

### Separately: a debug statement in production code

`internal/util/streamrpc/streamer.go:135`
```go
batch := s.maxMsgSize / 8
print(batch)
```

That is Go's builtin `print`, which writes unbuffered to stderr. It runs on every
oversized-message split.

- Deployments where delete-by-expression is rare or matches few rows.
- If the number of segments returning non-empty results per query is under ~10, the
  redundancy factor is only ~5.5× and the absolute cost stays under a millisecond.

### Honest caveat

The numbers above are for the merge loop in isolation. **What share of end-to-end
delete-by-expression latency this represents has not been measured.** See #52263.

xiaofan-luan:
Fixes #52273

## What

`RetrieveResultCache.merge` set `c.size = proto.Size(c.result)`, walking the whole
accumulated message on every merge. Over a stream of N results that is **O(N²)** in the
total number of IDs — the k-th merge visits `k·R` elements, so the stream touches
`R·N(N+1)/2` instead of `R·N`. At N=200 that is 100.5M element visits instead of 1M.

This sits inside the lock every segment's result has to pass through:
`segments/retrieve.go:141` sends one result per segment from one goroutine each
(`:125-127`), all serializing on `ResultCacheServer.mu` (`streamer.go:172`).

The only non-test caller of this path is `internal/proxy/task_delete.go:503`, i.e.
delete-by-expression.

## How

Accumulate instead of recomputing.

**What to charge.** `merge` keeps only the IDs and the counters and drops the rest of
the envelope, so the charge has to be the retained part, not the whole message. The
sole production caller is delete-by-expression, whose plan asks for the PK column plus
`common.TimeStampField` (`task_delete.go:463-465`) and which never sets `ignoreNonPk`
on the stream path — so every incoming result carries that PK data three times over,
once in `Ids` and twice more in `FieldsData`. Measured on auto-id PKs and TSO
timestamps:

| IDs per result | full message | `Ids` only | charged / retained |
|---|---|---|---|
| 100 | 2756 B | 906 B | 3.04× |
| 1000 | 27056 B | 9006 B | 3.00× |
| 5000 | 135067 B | 45008 B | 3.00× |

Charging the full envelope would trip the 4 MiB `queryStreamBatchSize` at ~1.3 MiB of
real payload, so querynode→proxy would emit 3× as many messages at a third of the size;
proxy produces one `deleteTask` per received message (`task_delete.go:574`), so delete
write batches would fragment by the same factor. Hence `proto.Size(result.GetIds())`,
with `Alloc` using the same rule so the two cannot disagree about one message.

**Direction still matters.** `Send` uses `c.size` to keep messages under `maxMsgSize`;
under-estimating would emit an oversized message. The `Ids` submessage's own tag and
length prefix (≥ 4 bytes) are not duplicated by appending, so every per-merge charge
stays conservative. The remaining sources of growth — widening length prefixes, summed
counters, a `CostAggregation` that appears only after a merge — are bounded by a
constant for the whole accumulation, not per merge, and are taken once as a fixed
headroom at `Put`. The tests pin both the safety property and the utilization.

## Measured

Real code, 5000 IDs per result (Apple M4 Pro, go1.26.5, median of 3 — treat the ratio):

| results merged | before | after | |
|---|---|---|---|
| 50 | 3.37 ms | 0.93 ms | 3.6× |
| 100 | 10.68 ms | 1.71 ms | 6.3× |
| 200 | 39.81 ms | 3.24 ms | **12.3×** |

Doubling N multiplies the old time by ~3.2 then ~3.7 (approaching 4× — quadratic) and
the new time by ~1.8 then ~1.9 (linear).

Note these merge counts exceed what one flush cycle sees in production: with `cap` at
4 MiB, 5000 auto-id PKs per result flushes at N≈93. The bound that matters is per flush
cycle, `(cap_in_IDs / 2) × N` wasted element visits, which grows linearly in N — so the
smaller each result is, the worse the old behaviour got.

## Also removed: a debug statement in production code

```go
batch := s.maxMsgSize / 8
print(batch)          // Go's builtin print -> unbuffered stderr
```

`splitMsgToMaxSize:135`. It ran on every oversized-message split.

## Tests

- `TestRetrieveResultCacheSizeNeverUnderestimates` — 200 merges × 500 IDs, asserting
  after **every** merge that the reported size is `>= proto.Size(c.result)`. This is the
  safety property the `maxMsgSize` guard depends on. Also verifies the merged IDs and
  the summed counters, and that the over-estimate stays under 1% of the real size
  rather than growing with the payload. The fixture carries the PK and timestamp
  columns a real delete stream sends — without them none of the accounting question
  above is observable.
- `TestRetrieveResultCacheSizeStringIDs` — same property on the `IDs_StrId` branch.
- `TestResultCacheServerFillsBatchesToCapacity` — drives `ResultCacheServer.Send` at
  the default 4 MiB `cap`, asserting both that no emitted message exceeds `maxMsgSize`
  and that batches fill the cap. Charging the full envelope puts them at ~33%.
- `BenchmarkRetrieveResultCacheMerge` at 50/100/200 results, so the scaling shape is
  visible in-tree.

Existing `internal/util/streamrpc` tests pass.

## What is not claimed

The numbers are for the merge loop in isolation. **What share of end-to-end
delete-by-expression latency this represents has not been measured.** See #52263.

### DIFF
--- a/internal/util/streamrpc/result_cache_size_test.go
+++ b/internal/util/streamrpc/result_cache_size_test.go
@@ -17,6 +17,9 @@
 package streamrpc
 
 import (
+	"context"
+	"math"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -26,18 +29,48 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/proto/internalpb"
 )
 
-func intResult(start, n int64) *internalpb.RetrieveResults {
-	data := make([]int64, n)
-	for i := range data {
-		data[i] = start + int64(i)
+// Auto-id primary keys and TSO timestamps are both large enough that their
+// varints occupy the full 8-9 bytes. Using small values instead would make the
+// Ids column compress far below the FieldsData columns and overstate the effect
+// these tests are about.
+const (
+	firstAutoID    = 458000000000000000
+	firstTimestamp = 460000000000000000
+)
+
+// deleteStyleResult mirrors what querynode actually streams for
+// delete-by-expression: the ID list plus a FieldsData carrying the PK column and
+// common.TimeStampField, because the delete plan asks for both
+// (proxy/task_delete.go) and the stream path never enables ignoreNonPk.
+//
+// merge keeps only the Ids, so a fixture without FieldsData cannot observe how
+// much of an incoming result is charged but discarded.
+func deleteStyleResult(start, n int64) *internalpb.RetrieveResults {
+	pks := make([]int64, n)
+	tss := make([]int64, n)
+	for i := range pks {
+		pks[i] = firstAutoID + start + int64(i)
+		tss[i] = firstTimestamp + start + int64(i)
+	}
+	longCol := func(name string, fieldID int64, data []int64) *schemapb.FieldData {
+		return &schemapb.FieldData{
+			Type:      schemapb.DataType_Int64,
+			FieldName: name,
+			FieldId:   fieldID,
+			Field: &schemapb.FieldData_Scalars{Scalars: &schemapb.ScalarField{
+				Data: &schemapb.ScalarField_LongData{LongData: &schemapb.LongArray{Data: data}},
+			}},
+		}
 	}
 	return &internalpb.RetrieveResults{
 		Ids: &schemapb.IDs{
-			IdField: &schemapb.IDs_IntId{IntId: &schemapb.LongArray{Data: data}},
+			IdField: &schemapb.IDs_IntId{IntId: &schemapb.LongArray{Data: pks}},
 		},
+		FieldsData:         []*schemapb.FieldData{longCol("pk", 100, pks), longCol("Timestamp", 1, tss)},
 		AllRetrieveCount:   n,
 		ScannedTotalBytes:  n * 8,
 		ScannedRemoteBytes: n * 4,
+		CostAggregation:    &internalpb.CostAggregation{ResponseTime: 1, ServiceTime: 1, TotalNQ: 1},
 	}
 }
 
@@ -50,7 +83,7 @@ func TestRetrieveResultCacheSizeNeverUnderestimates(t *testing.T) {
 
 	cache := &RetrieveResultCache{cap: 1 << 30} // never flushes
 	for i := 0; i < results; i++ {
-		cache.Put(intResult(int64(i*perResult), perResult))
+		cache.Put(deleteStyleResult(int64(i*perResult), perResult))
 		assert.GreaterOrEqual(t, cache.size, proto.Size(cache.result),
 			"reported size must not be below the real encoded size after %d merges", i+1)
 	}
@@ -59,17 +92,20 @@ func TestRetrieveResultCacheSizeNeverUnderestimates(t *testing.T) {
 	got := cache.result.GetIds().GetIntId().GetData()
 	assert.Len(t, got, results*perResult)
 	for i := range got {
-		assert.Equal(t, int64(i), got[i])
+		assert.Equal(t, int64(firstAutoID+i), got[i])
 	}
 	assert.Equal(t, int64(results*perResult), cache.result.GetAllRetrieveCount())
 	assert.Equal(t, int64(results*perResult*8), cache.result.GetScannedTotalBytes())
 	assert.Equal(t, int64(results*perResult*4), cache.result.GetScannedRemoteBytes())
 
-	// The over-estimate stays proportional to the per-result envelope rather
-	// than growing with the payload.
+	// Charging the whole incoming envelope rather than the part merge retains
+	// costs a steady ~3x here, which would flush at a third of cap. The slack
+	// must stay a per-merge constant (the Ids tag and length prefix) plus the
+	// one-off headroom, not a fraction of the payload.
 	real := proto.Size(cache.result)
-	assert.Less(t, cache.size-real, real/10,
-		"over-estimate should stay well under 10%% of the real size")
+	assert.Less(t, cache.size-real, real/100,
+		"over-estimate should stay under 1%% of the real size, got %d of %d",
+		cache.size-real, real)
 }
 
 func TestRetrieveResultCacheSizeStringIDs(t *testing.T) {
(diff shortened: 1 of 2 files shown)


xiaofan-luan:

```
```

czs007:

Adversarial review surfaced one size-accounting regression that can materially shrink production delete batches, plus one minor test-code cleanup.

## Medium

**`internal/util/streamrpc/streamer.go:103`** — `c.size += proto.Size(result)` charges the full incoming `RetrieveResults` even though `merge` (streamer.go:84-93) retains only `Ids` and the counters, discarding `FieldsData`/`Status`/`SealedSegmentIDsRetrieved`. The only production caller is delete-by-expression (`internal/proxy/task_delete.go:503`), whose `plan.OutputFieldIds` explicitly includes PK + `TimeStampField` (task_delete.go:468-470), and the stream path never enables `ignoreNonPk` (`SetIgnoreNonPk` is called only in the non-stream `retrieveOnSegments`, retrieve.go:47; `RetrieveStream` at retrieve.go:198-221 never calls it) — so every incoming result carries a PK column plus a TSO timestamp column (~9-10 byte varints), making the accounted size roughly 3× the actually-merged payload. Failure: a delete-by-expression matching many rows across multiple segments → the default 4 MiB `queryStreamBatchSize` reports `IsFull` at roughly 1.3 MiB of real payload, so querynode→proxy emits ~3× as many stream messages at ~1/3 the size, and since proxy produces one `deleteTask` per received message (task_delete.go:578), the delete write batches fragment by the same factor and bulk-delete throughput drops. The over-estimate is directionally safe, so the `maxMsgSize` guard still holds and this is not a correctness bug — but the magnitude contradicts the PR description's "over-estimating merely flushes slightly early", and the `<10%` bound asserted in `result_cache_size_test.go:28-41` only holds because that fixture omits `FieldsData` entirely. Suggestion: accumulate only the portion actually retained (e.g. `proto.Size(result.GetIds())` plus the counter deltas) and use that same bound consistently in `Alloc` and `merge`; add a `ResultCacheServer.Send` test with timestamp `FieldsData` that checks both max-size safety and batch utilization. *(raised by sijie-ni-0214, bigsheeper)*

## Low

**`internal/util/streamrpc/result_cache_size_test.go:102`** — the new benchmark names subtests via a hand-rolled `strconvItoa` (lines 102-111), a `switch` covering only 50/100/200 where any input other than 50 or 100 falls through to `"200results"`. The current `n` set is exactly `{50, 100, 200}`, so no mislabeling occurs today; this is purely a code-quality issue. Suggestion: use the standard library — `b.Run(fmt.Sprintf("%dresults", n), ...)` or `strconv.Itoa(n)+"results"` — and drop the shadow helper. *(surfaced during verification)*

czs007:

Verified:
- `internal/util/streamrpc/streamer.go:69-76` — the first result is retained intact and charged at its full `proto.Size` plus fixed headroom, so no retained payload is omitted from the estimate.
- `internal/util/streamrpc/streamer.go:91-93,119-140` — merges append only IDs and bounded scalar/cost fields; per-merge charging by `proto.Size(result.GetIds())` was checked against the actual wire encoding for both packed int64 IDs and string IDs, and is conservative in every case.
- `internal/util/streamrpc/streamer.go:48-61` — the 256-byte `sizeGrowthHeadroom` covers all one-time accumulative growth (nested length-prefix widening, counter varint widening, `CostAggregation` materialization), which totals under 130 bytes and does not scale with merge count.
- `internal/util/streamrpc/streamer.go:102-109` — `Alloc` applies the same first-result versus merged-result accounting as `Put`, so an allocation decision cannot accept a contribution that `Put` would charge differently.
- `internal/util/streamrpc/streamer.go:216-245` — cache allocation, merge, flush, and split decisions all remain under `ResultCacheServer.mu`, with no new shared state or changed locking order.
- `internal/util/streamrpc/result_cache_size_test.go:87-115` — tests assert `size >= proto.Size(result)` after every merge, directly pinning the no-underestimate invariant, with server-level coverage confirming batch utilization and the per-message size ceiling.

czs007 (approved):
/lgtm

Issue #52282, opened 2026-08-07
[Enhancement]: Improve Storage V3 segment load resource estimation


sijie-ni-0214:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Persist a compact, pathless load-resource descriptor in the existing segment `Statistics` lifecycle and forward it to QueryNode through `SegmentLoadInfo`.

For raw data, the descriptor should record each physical column group's ID, member field IDs, and cumulative uncompressed memory size. QueryNode can adapt these aggregates into synthetic, pathless `FieldBinlog` values and reuse the existing loading-resource estimator. The cumulative delta footprint should continue to come from `Statistics.delta_binlog_size`.

The descriptor should be generated from the same insert metadata already used to maintain `Statistics`, restored and cloned together with the existing statistics, and forwarded by QueryCoord for both Worker QueryNode sealed loads and SQN growing-segment recovery.

### Why is this needed?

Internal Storage V3 segments are loaded from a manifest and do not rely on the legacy per-field binlog KVs. After coordinator recovery, `SegmentLoadInfo.binlog_paths` and `deltalogs` may therefore be empty even though the segment contains data.

The current QueryNode loading-resource estimator consumes structured `SegmentLoadInfo` metadata and intentionally does not read the manifest. As a result, it can omit most of the raw-data and delete-data footprint for a manifest-backed segment. This can significantly underestimate the loading peak and admit a segment without reserving the resources that the real load requires.

DataCoord already retains Text and JSON statistics metadata, but the manifest branch also needs to forward the Text metadata so the existing estimator can account for it.

### Anything else?

Proposed behavior:

- Add `Statistics.load_resource` containing physical column-group summaries only.
- Maintain it as an ordinary nested part of `Statistics`; do not add a separate KV, revision, version, or freshness state.
- Populate it from existing DataNode/SQN writer output and one-shot `Statistics` builders.
- Add `SegmentLoadInfo.stats` and let QueryCoord forward the complete `Statistics` object.
- For internal Storage V3 manifest segments, adapt column groups and `delta_binlog_size` into local synthetic logs used only by `estimateLoadingResourceUsageOfSegment`.
- Keep the real segment load, index metadata, JSON/Text statistics, row count, and manifest resolution on the original `SegmentLoadInfo`.
- Keep V1/V2 and external collections on their existing estimation paths.
[...]

PR #52284, opened 2026-08-07
enhance: improve Storage V3 segment resource estimation

sijie-ni-0214:
issue: #52282

Depends on: https://github.com/milvus-io/milvus/pull/50410

## Background and rollout dependency

https://github.com/milvus-io/milvus/pull/50410 introduced the durable `datapb.Statistics` lifecycle on `SegmentInfo`. As part of that change, Storage V3 stopped persisting the legacy per-`FieldBinlog` KVs because the manifest is authoritative for file paths and `Statistics` carries the aggregates needed by DataCoord.

That PR did not yet provide QueryNode with a pathless raw-data breakdown. QueryNode's loading and logical resource estimators still consume `SegmentLoadInfo.binlog_paths` and `deltalogs`, and intentionally do not read the manifest. After coordinator recovery, those legacy arrays can therefore be empty for a V3 segment, causing the loading peak and cached logical usage to be significantly underestimated.

**This PR and #50410 must ship in the same Milvus release. Do not release or backport #50410 without this PR.** This PR completes the `Statistics` rollout by adding the resource descriptor and its QueryCoord/QueryNode transport and consumption path. Conversely, this PR depends on the producer, persistence, recovery, and wholesale-update lifecycle introduced by #50410.

The protobuf additions remain rolling-upgrade compatible: old QueryNodes ignore the new field and new QueryNodes fall back when an old QueryCoord does not send it. However, the resource-estimation improvement is complete only when both PRs are included in the release and the participating DataNode/StreamingNode, coordinator, SQN, and Worker QueryNode processes have been upgraded.

Related design document from #50410: `docs/design-docs/design_docs/20260602-segment_summary_metadata.md`

## What does this PR do?

Improve loading and logical resource estimation for internal Storage V3 manifest segments without reading the manifest in QueryNode.

This PR:

- adds a compact, pathless column-group descriptor to `datapb.Statistics`;
- maintains the descriptor through the existing Statistics digest, restore, clone, publish, and one-shot builder paths;
- forwards the complete Statistics object, Text stats, and JSON stats through QueryCoord and SQN growing recovery;
- adapts Storage V3 column groups and cumulative delta size into synthetic local logs when existing binlog arrays are absent;
- keeps real loading and all other `SegmentLoadInfo` consumers on the original message.

## Compatibility and scope

The protobuf changes are additive. Old QueryNodes ignore `SegmentLoadInfo.stats`, while new QueryNodes receiving requests from an old QueryCoord see a nil descriptor and fall back to the existing estimators.

The adapter only consumes the descriptor for internal Storage V3 segments with a manifest when `binlog_paths` is empty. Existing arrays, V1/V2 segments, and external collections retain their existing paths. Fields removed from the current schema are skipped during estimation. A missing descriptor does not block loading, although the fallback may still underestimate resources.

This change intentionally does not add manifest reads, PK Bloom estimation, historical backfill, a descriptor version, or separate freshness state.

## Test plan

- [x] `make generated-proto-without-cpp`
- [x] `make build-go`
- [x] `make static-check`
- [x] Statistics descriptor build, digest, restore, clone, nil/empty, and protobuf round-trip tests
- [x] DataCoord legacy Statistics fallback test
- [x] QueryCoord `SegmentLoadInfo` packing test
- [x] SQN growing-segment metadata forwarding test
- [x] QueryNode descriptor resolver, existing-array precedence, dropped-field tolerance, loading/logical equivalent-estimate, and runtime cache tests
- [x] Partial-load full-metadata estimate test


czs007:

Adversarial review of this PR found no issues introduced by the change itself; one real but pre-existing gap is noted below for follow-up.

Verified:
- `internal/querynodev2/segments/segment_loader.go:210` — `resolveSegmentEstimateLogs` early-returns the original `BinlogPaths`/`Deltalogs` unless the segment is Storage V3 with a manifest set, non-external, and `LoadResource != nil`; the external branch then iterates the same originals, so there is no double-counted path.
- `internal/querynodev2/segments/segment_loader.go:233-238` — the synthetic delta binlog sets only `MemorySize`, so in PART 4 the disk size stays 0 and the expansion factor remains 2, never falling into the 2.3-compression branch.
- `internal/querynodev2/segments/segment_loader_test.go:392` — aliasing is covered: `ChildFields` is copied via `append([]int64(nil), …)`, collector paths deep-copy `*ColumnGroupStatistics`, and the test asserts the synthetic binlog does not write back into `LoadResource`.
- `internal/querynodev2/segments/statistics.go:580` — the `c.columnGroups == nil` guard in `Publish` lets an explicit-empty `LoadResource` round-trip while a fully-empty collector still returns nil.
- `internal/querynodev2/segments/statistics.go:500-522` — `mergeColumnGroup` accumulates `MemorySize` across syncs, dedups `FieldIds`, defaults `ChildFields` to `[groupID]` for legacy per-field binlogs, and output ordering is deterministic.
- `internal/querycoordv2/utils/types.go:42-44` — `Stats`/`TextStatsLogs`/`JsonKeyStatsLogs` are forwarded additively, so old QueryNodes ignore the new fields; populating them for manifest segments is intentional and feeds the corrected estimate.

- `internal/querynodev2/segments/segment_loader.go:1956` — `estimateLogicalResourceUsageOfSegment` is not routed through `resolveSegmentEstimateLogs`: it still reads `loadInfo.BinlogPaths` directly (:1956) and `loadInfo.Deltalogs` (:2100), so the logical resource estimate for Storage V3 manifest segments (empty `BinlogPaths`/`Deltalogs`, footprint carried in `Stats.LoadResource`) is understated. There is no runtime impact today: the entire logical-resource consumer chain is currently commented-out dead code — the only call to `checkLogicalSegmentSize` is disabled (`segment_loader.go:535-536`), the only producing consumer of `ResourceUsageEstimate` is disabled (`segment.go:1379-1381`), and `AddLogicalResource` has no callers — so the logical path participates in no admission or eviction decision. This is unrelated to this PR because the two affected lines were not modified here and the underestimate existed before the change; this PR fixed the loading estimator without making the logical estimator any worse. Suggested follow-up (no tracking issue yet): before re-enabling the logical resource path, call `resolveSegmentEstimateLogs` inside `estimateLogicalResourceUsageOfSegment` as well, and note that `compactSegmentLoadInfoForRuntime` (`segment.go:188-216`) drops `Stats`/`BinlogPaths`, so any recomputation must consume them before compaction. (raised by xiaocai2333, chyezh)

sijie-ni-0214:
Addressed the logical-estimator follow-up in `11318cf7b8`.

`estimateLogicalResourceUsageOfSegment` now uses `resolveSegmentEstimateLogs` for both raw and delta metadata, matching the loading estimator. The runtime compaction path computes and caches logical usage before dropping `Stats` and `BinlogPaths`; regression coverage now verifies:

- descriptor-backed and manually expanded metadata produce identical loading and logical estimates;
- a Storage V3 segment retains its cached logical estimate after `SegmentLoadInfo` compaction.

The targeted tests and `make static-check` pass. PTAL.

czs007:

Adversarial review confirmed two medium-severity findings in the segment resource-estimation path, both introduced by this PR.

## Medium

- **`internal/querynodev2/segments/segment_loader.go:2045` — The logical and loading estimators handle dropped fields inconsistently, so on the new V3 descriptor path `compactLoadInfoForRuntime` silently fails after a field is dropped.** After a V3 segment is flushed, if any field in a merged column group's `Stats.LoadResource.FieldIds` is later dropped (no code path ever prunes `FieldIds`) and a QueryNode then loads or reloads the segment → `estimateLogicalResourceUsageOfSegment` returns an error and `compactLoadInfoForRuntime` returns early: the segment's runtime load info is never compacted (Stats/Statslogs/IndexInfos stay resident, defeating this PR's "compute and cache logical usage, then drop Stats" goal for such segments), `resourceUsageCache` is stored as nil, and every `ResourceUsageEstimate()` call recomputes, fails, logs an "unreachable" warning, and returns a zero-value `ResourceUsage`. The loading estimator already tolerates this exact case with a `continue` (`segment_loader.go:2252-2258`, covered by `TestEstimateLoadingResourceUsage_DroppedFieldSkipped`), while the logical estimator does `return nil, err` (`segment_loader.go:2042-2046`). The new `resolveSegmentEstimateLogs` (`segment_loader.go:1928-1937`) feeds the same expanded `ChildFields` into both. Note the error-return line itself is a pre-existing defect with an existing trigger surface (V2 grouped segments, and V3 segments before a DataCoord restart), but this PR adds a deterministic trigger path for V3 descriptor segments — its target population. Today the impact is limited to un-compacted load info and log noise because `checkLogicalSegmentSize` is commented out (`segment_loader.go:536`); if that check is re-enabled, this becomes a deterministic load failure. Suggestion: change the `return nil, err` at `segment_loader.go:2045` to `continue`, matching the loading estimator. *(raised by tinswzy)*

- **`internal/querynodev2/segments/segment_loader.go:1920` — During the rolling-upgrade window, a non-nil but under-counted `LoadResource` takes precedence over complete `BinlogPaths`, causing load-resource underestimation.** During a compliant same-release rolling upgrade from a release without #50410, with Storage V3 enabled: while DataCoord is still on the old version, a new-version streamingnode restarts or a channel rebalances, so the V3 growing segment's `StatisticsCollector` reseeds from nil (`segment_stats.go:44`, "nil stats yields an empty collector") and only counts syncs after the reseed; once DataCoord upgrades, that under-counted cumulative Stats is persisted wholesale via `UpdateSegmentStats` (`meta.go:1491-1492`), and when a QueryNode later loads the sealed segment → `resolveSegmentEstimateLogs` replaces the complete `BinlogPaths` (kept intact by the unconditional in-memory merge in `AddBinlogsOperator`, `meta.go:1187`) with `ColumnGroups` covering only post-reseed data, `estimateLoadingResourceUsage` underestimates memory/disk needs, and the active admission check `checkLoadingResource` (`segment_loader.go:542`) admits loads exceeding actual capacity, risking QueryNode out-of-memory. Before this change the same state used the complete `BinlogPaths` and estimated correctly, and the under-counted Stats persist until the segment's stats are rebuilt by compaction (`compactor/stats.go:44`). Note the important scoping: when `LoadResource` is nil, the code falls back to the original binlog arrays (`segment_loader.go:1919-1922`) and behaves identically to before, so the "Stats present but LoadResource nil" distribution state is fully covered by the PR's same-release shipping constraint with #50410 — the residual gap is specifically the non-nil, under-counted descriptor above. Suggestion: prefer the original array when `BinlogPaths` is non-empty, which closes this window at zero cost. The trigger conditions are narrow (V3 enabled + mixed-version window with a streamingnode restart/rebalance + segment flushed after the DataCoord upgrade), so this can be treated as non-blocking, but addressing it in this PR is recommended. *(raised by tinswzy, xiaocai2333)*

sijie-ni-0214:
Addressed both medium findings in `8886c92b79`.

- The logical estimator now skips fields that are absent from the current schema, matching the loading estimator. The V3 runtime-compaction test includes both a live field and a dropped field and verifies that the logical estimate is cached before the load info is compacted.
- The production rollout upgrades DataCoord before StreamingNode, so the mixed new-SN/old-DC sequence is not expected. As a small defensive compatibility guard, existing `BinlogPaths`/`Deltalogs` now take precedence when present; the V3 descriptor is used only when the binlog array is empty. A resolver regression test covers this precedence.

The targeted QueryNode tests and `make static-check` pass. PTAL.

czs007 (approved):
/lgtm

PR #52288, opened 2026-08-07
test: assert the #51625 canonicalization contract for mixed-type JSON IN

### LINKED ISSUE #52286: [Bug]: master e2e is red for every rebased PR — test_filter_regressions.py asserts a pre-#51625 contract

- [x] I have searched the existing issues (`test_filter_regressions`, `rejected_at_planner`, `mixed-type JSON IN unexpectedly succeeded`) — no match.

### Environment

```
- Milvus version: master @ ea06bd56c7 and later
- Deployment mode: CI (ci-v2/e2e-default, milvus-e2e-pipeline-gcp)
```

### Current Behavior

Since `ea06bd56c7` was merged on 2026-08-06 10:38 UTC, **every PR whose branch contains that commit fails `ci-v2/e2e-default`**, which blocks `ci-passed` and therefore blocks merging.

Three tests fail, deterministically and identically on every affected PR:

```
FAILED milvus_client/expressions/test_filter_regressions.py::TestFilterRegressions::
  test_json_mixed_type_in_rejected_at_planner_51489[int_string_in]
  - Failed: mixed-type JSON IN unexpectedly succeeded with IDs [1]; planner rejection is required

FAILED ...test_json_bool_mixed_type_in_rejected_at_planner_51567[bool_int_in_true_one]
  - Failed: mixed-type JSON IN unexpectedly succeeded with IDs [1, 2, 6]; planner rejection is required

FAILED ...test_json_bool_mixed_type_in_rejected_at_planner_51567[bool_int_in_false_one]
  - Failed: mixed-type JSON IN unexpectedly succeeded with IDs [3, 4, 6]; planner rejection is required

= 3 failed, 1605 passed, 71 skipped, 8 xfailed, 6 warnings, 9 rerun =
```

Confirmed on unrelated PRs by different authors, with byte-identical failures and identical `1605 passed` counts:

| PR | Area changed | Jenkins build |
|---|---|---|
| #52249 | — | [gcp/47211](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-gcp/47211/) (first known-bad) |
| #50069 | — | gcp/47229 |
| #52252 | — | gcp/47247 |
| #52267 | — | gcp/47314 |
| #52268 | segcore C++ (GIS offsets) | gcp/47355 |
| #52260 | datacoord (compaction scheduling) | gcp/47390, gcp/47395 |

PRs that still pass are exactly those whose head does **not** yet contain `ea06bd56c7` (verified on #52248, #52245, #52257, #52123 — the test file 404s on their heads).

### Expected Behavior

The tests should pass. **This is not a product regression — the product is correct and the test asserts a contract that no longer exists.**

`test_filter_regressions.py` was added by `ea06bd56c7` ("test: Add focused filtering expression coverage matrix", #51657, merged 2026-08-06 10:38 UTC). It is the only commit that has ever touched the file.

The file asserts that a mixed-type JSON `IN` must be **rejected by the planner with code 1100**. That contract was changed two weeks earlier by `0ec31c4b0b` ("fix: handle mixed-type JSON term expressions", #51625, merged 2026-07-23 08:10 UTC), which cites `issue: #51489 #51567 #51568` and canonicalizes such expressions instead of rejecting them:

```
meta["p"] in [1, "2"]   ->   meta["p"] in [1] OR meta["p"] == "2"
```

Note that a planner-level 1100 rejection **never existed**. Before #51625 the planner accepted these expressions and segcore raised a type assertion — which is exactly what #51489/#51567 reported. The helper `assert_meaningful_type_mismatch_rejection()` only ever went green through its `pytest.xfail` escape hatches, which match either the old segcore assertion (`kStringVal` / `kInt64Val`) or a hardcoded pre-fix ID list.

The values master now returns are correct. Deriving them from the fixture in `regression_51568_collection` (`meta["b"]` by pk: `1:True 2:True 3:False 4:False 5:0 6:1 7:"yes" 8:"no"`; `meta["p"] = i`, except the last row which is the string `"1"`):

| case | expression | canonicalized | expected | CI actual |
|---|---|---|---|---|
| `int_string_in` | `meta["p"] in [1, "2"]` | `p==1` → 1; `p=="2"` → none | `[1]` | `[1]` ✓ |
| `bool_int_in_true_one` | `meta["b"] in [true, 1]` | `b==true` → 1,2; `b==1` → 6 | `[1,2,6]` | `[1,2,6]` ✓ |
| `bool_int_in_false_one` | `meta["b"] in [false, 1]` | `b==false` → 3,4; `b==1` → 6 | `[3,4,6]` | `[3,4,6]` ✓ |

All three match. The test's own hardcoded pre-fix result `KNOWN_ISSUE_51567_IN_SUCCESS_IDS = {"bool_int_in_false_one": [3, 4]}` is missing pk `6` — the int branch that the pre-#51625 code silently dropped.

### How the stale contract reached master

`#51657` got `ci-v2/e2e-default` = success on **2026-07-24 10:35 UTC**. Its merge base was `91f7a255b4` (2026-07-23 06:22:43 UTC), i.e. **before #51625 merged at 08:10:45 UTC** — `gh api compare 0ec31c4b0b...812b213fb5` reports `behind_by: 2`. Its e2e image was built pre-fix, so the tests hit the old segcore assertion and took the `xfail` path. The PR then sat for 13 days and was merged on that stale green status without an e2e re-run against a master that by then contained #51625.

### Additional context

**More cases are latent.** Only 3 failures surface because of tag filtering — `L0_51489_IN_CASE_NAMES = {"int_string_in"}` and `L1_51567_BOOL_IN_CASE_NAMES = {"bool_int_in_true_one", "bool_int_in_false_one"}`; the rest are `CaseLabel.L2` and this pipeline does not run them. Five sibling cases use the same helper and the same stale contract and will fail once L2 runs: `string_int_in`, `int_unrelated_string_in`, `json_array_subscript_mixed_in`, `bool_string_in`, `bool_int_string_in`. (Inferred from the tag tables plus the failure count; I did not inspect the pipeline's tag arguments.)

**#51489, #51567 and #51568 are all still open** even though #51625 fixed them on 2026-07-23. That is very likely why the test was authored against the pre-fix "must be rejected" contract. They should be closed.

### Proposed fix

Rewrite the mixed-type `IN` assertions to the #51625 semantics: the query succeeds and returns the type-partitioned union. Rather than hardcoding IDs, assert equivalence against the explicit OR-expansion (`meta["p"] in [1, "2"]` must return exactly what `meta["p"] in [1] or meta["p"] == "2"` returns), which is self-verifying and is precisely the contract #51625 defines. Drop the `KNOWN_ISSUE_*_ERROR_VALUE_CASES` / `KNOWN_ISSUE_*_SUCCESS_IDS` xfail hatches for these cases.

I am preparing a PR for this. cc @yanliang567 @zhengbuqian

/kind bug

xiaofan-luan:
Fixes #52286

`ci-v2/e2e-default` has been red for every PR rebased past `ea06bd56c7` since 2026-08-06 10:38 UTC, with three deterministic failures:

```
test_json_mixed_type_in_rejected_at_planner_51489[int_string_in]        -> IDs [1]
test_json_bool_mixed_type_in_rejected_at_planner_51567[bool_int_in_true_one]  -> IDs [1, 2, 6]
test_json_bool_mixed_type_in_rejected_at_planner_51567[bool_int_in_false_one] -> IDs [3, 4, 6]
  Failed: mixed-type JSON IN unexpectedly succeeded ...; planner rejection is required
```

## Why the tests are wrong, not the product

The cases require the planner to reject a mixed-type JSON `IN` with code 1100. **That rejection never existed.** Before #51625 the planner accepted the heterogeneous term list and segcore picked its executor from the first value, tripping a type assertion — which is precisely what #51489 and #51567 reported. The helper's "pass" branch was unreachable; the tests only ever went green through their `pytest.xfail` hatches, matching either that segcore assertion or a hard-coded pre-fix ID list.

#51625 partitions the term list by literal kind and ORs the per-kind predicates back together:

```
meta["p"] in [1, "2"]   ==>   meta["p"] == 1 or meta["p"] == "2"
```

so the query now succeeds and returns the correct union. Neither hatch matches any more, and the cases hard-fail.

The returned values are correct. Derived independently from the fixture (`meta["b"]` by pk = `1:True 2:True 3:False 4:False 5:0 6:1 7:"yes" 8:"no"`; `meta["p"] = i` except pk 3000 which is the string `"1"`):

| case | canonicalized | expected | CI actual |
|---|---|---|---|
| `meta["p"] in [1, "2"]` | `p==1` → 1; `p=="2"` → none | `[1]` | `[1]` ✓ |
| `meta["b"] in [true, 1]` | `b==true` → 1,2; `b==1` → 6 | `[1,2,6]` | `[1,2,6]` ✓ |
| `meta["b"] in [false, 1]` | `b==false` → 3,4; `b==1` → 6 | `[3,4,6]` | `[3,4,6]` ✓ |

The stale `KNOWN_ISSUE_51567_IN_SUCCESS_IDS = {"bool_int_in_false_one": [3, 4]}` is missing pk `6` — the int branch the pre-#51625 code silently dropped.

## What this changes

Assert the real contract: **the `IN` form must return exactly what its explicit OR expansion returns.** Equivalence rather than hard-coded IDs is what #51625 actually promises, and it survives fixture changes. Each case now carries its expansion in the case table:

```python
("bool_int_in_false_one", 'meta["b"] in [false, 1]',
 '(meta["b"] == false) or (meta["b"] == 1)', [3, 4, 6]),
```

The three CI-confirmed values are additionally pinned so that two identically-broken sides (both empty, say) cannot pass silently. The other five cases run pure equivalence.

Search variant: ground truth comes from a scalar **query** on the expansion, never a second ANN search — two searches would put recall on both sides and any disagreement would be unattributable. It asserts the subset relation unconditionally (ANN can drop candidates, never invent them, so an extra row always means the filter is unsound), then equality guarded by an explicit check that the match count is under the top-k limit.

Removed as unused after the conversion (grep-verified across `tests/python_client`): `assert_meaningful_type_mismatch_rejection`, `KNOWN_ISSUE_51489_IN_ERROR_VALUE_CASES`, `KNOWN_ISSUE_51567_IN_ERROR_VALUE_CASES`, `KNOWN_ISSUE_51567_IN_SUCCESS_IDS`, `MIXED_LITERAL_TYPE_TOKENS`, `MIXED_IN_LITERAL_TYPES_51489/51567`. `coverage_matrix.md` rows updated to the new test names and contract.

## Also covered

Only 3 failures surface today because of tag filtering (`L0_51489_IN_CASE_NAMES`, `L1_51567_BOOL_IN_CASE_NAMES`); five sibling cases are `CaseLabel.L2` and use the same helper, so they would fail as soon as L2 runs. All eight are converted here.

## Out of scope

The **OR** cases (#51568 / #51567 OR) are green and untouched — their `expected_ids` are already the post-fix values, so they assert the right contract. What is stale there is only the escape-hatch data (`KNOWN_ISSUE_51568_ERROR_VALUE_CASES`, `KNOWN_ISSUE_51568_LOSS_IDS`, `KNOWN_ISSUE_51567_OR_*`): #51625 made the `== OR` → `IN` merge type-aware for JSON (`valueGroupKey` in `rewriter/util.go`), so those xfail branches should now be unreachable dead code that would silently absorb a future regression. Worth a follow-up cleanup.

Related: #51489, #51567 and #51568 are all still open although #51625 fixed them on 2026-07-23. That is very likely why these tests were authored against the pre-fix contract — they should be closed.

## Verification

`ruff check` and `ruff format --check` (via `tests/ruff.toml`) pass; `py_compile` passes. I could not execute the suite — it needs a live cluster — so the expected values are derived from the fixture and cross-checked against the CI output above for the three confirmed cases. e2e on this PR is the real check.

cc @yanliang567 @zhengbuqian


xiaofan-luan:
Both findings taken — fixed in 98eadc6c.

**P2: exact HNSW recall.** You're right, and the guard I had was answering the wrong question. Keeping the match count under the top-k limit only rules out truncation; it does nothing about an approximate index skipping a qualifying candidate during graph traversal, which is exactly the point `tests/integration/bloommatch/bloom_match_test.go:756-770` makes about residual differences remaining ANN recall even at `topK == rowNum`.

Dropped `search_ids == expansion_ids`. What remains against the search result is only the subset relation, which holds unconditionally — ANN can drop candidates but never invent them, so a row present in the search and absent from the expansion always means the filter is unsound. Exactness is now asserted through a scalar query on the same mixed-type expression, which filters exhaustively. Between the two, the contract is still fully pinned: the predicate resolves to the right row set, and the search path applies that same predicate soundly. No assertion depends on recall any more.

**P2: pin every case.** Also right, and it defeats the specific failure mode this test exists to catch — both sides go through the same rewrite and execution path, so a `valueGroupKey` regression that recombines literal kinds could make the mixed `IN` and its expansion agree on the same wrong answer, and a `None` case would sail through. All eight cases now carry `expected_ids`, derived from the deterministic fixture:

| case | expected | derivation |
|---|---|---|
| `int_string_in` | `[1]` | `p==1` → pk 1; no row stores string `"2"` |
| `string_int_in` | `[2, 3000]` | `p=="1"` → pk 3000 (the only string `p`); `p==2` → pk 2 |
| `int_unrelated_string_in` | `[1]` | `"missing"` matches nothing |
| `json_array_subscript_mixed_in` | `[1]` | `arr[0]` is the row id, never a string |
| `bool_int_in_true_one` | `[1, 2, 6]` | `true` → 1,2; int `1` → 6 |
| `bool_int_in_false_one` | `[3, 4, 6]` | `false` → 3,4; int `1` → 6 |
| `bool_string_in` | `[1, 2, 7]` | `true` → 1,2; `"yes"` → 7 |
| `bool_int_string_in` | `[1, 2, 6]` | `"true"` matches nothing; pk 7/8 hold `"yes"`/`"no"` |

Three of these (`[1]`, `[1,2,6]`, `[3,4,6]`) are the values current master actually returned in the CI failures this PR fixes, so they are observed, not just derived. The other five are derivation only — e2e on this PR is what confirms them.

For the record on the earlier red e2e here: it failed in the image-build stage with `Code-Check: result_json_missing` / `UT-CPP-Cov: result_json_missing` (Build itself was SUCCESS). This PR touches three Python test files and cannot affect C++ UT; other PRs' `build-ut-cov` passed in the same window. That looked like a harness flake, and the push above re-runs it.

yanliang567:
/lgtm

a few lines of md descriptions are out of data, i can update them later

Issue #52087, opened 2026-07-31
[Bug]: REST v2 insert rejects Text fields with error code 1804


zhuwenxing:
## Environment

- Milvus version: `unknown-20260731-7e92e534`
- Git commit: `7e92e5342d0733fff1a16c08f83973a74510e023` (from the exact affected pod's startup banner in Loki)
- Deployment mode: standalone
- MQ type: Pulsar
- SDK version: RESTful API v2 with `requests==2.34.2`
- OS: Linux/arm64 (`go version go1.26.5 linux/arm64`); the exact distribution is unavailable because the UAT namespace was removed after the run
- CPU/Memory: unavailable after UAT namespace removal
- GPU: N/A
- Others: DB version `v3.0.0-7e92e534-5210`; `common.storage.useLoonFFI=true`

### K8s Pod List

The namespace was removed after the test, so live READY/STATUS/RESTARTS fields are no longer available. The following historical pod inventory was recovered from Loki labels for the failure window:

```text
NAME                                                                     COMPONENT    EVIDENCE
in01-fff268e33fe56ba-etcd-0                                              etcd         Loki stream present
in01-fff268e33fe56ba-milvus-standalone-r1-1-5cbb7cc97d-bjzqk            standalone   Served the failing REST request
```

No Kubernetes warning, OOM, backoff, or unhealthy event was found for this instance in the failure window.

## Current Behavior

Creating a collection with a `Text` field through REST v2 succeeds. Inserting even a short string into that field returns HTTP 200 with business error code `1804`:

```text
fail to deal the insert data, error: fieldName: doc: invalid parameter[expected=][actual=Text]
```

The same test setup can create the collection, create an index, and load it. The failure occurs synchronously while REST v2 converts the JSON row, before insert persistence.

## Expected Behavior

`POST /v2/vectordb/entities/insert` should accept a JSON string for a schema field whose data type is `Text`, returning business code `0` and an insert count of `1`.

## Steps To Reproduce

1. Start Milvus from commit `7e92e5342d0733fff1a16c08f83973a74510e023` with REST v2 enabled.
2. Set `MILVUS_URI` and `MILVUS_API_KEY` in the environment.
3. Install `requests==2.34.2` and run the script below.
4. Observe that collection creation succeeds but entity insertion returns code `1804`.

## Minimal Reproduction Script

```python
import os
import uuid

import requests


base_uri = os.environ["MILVUS_URI"].rstrip("/")
[...]

PR #52289, opened 2026-08-07
fix: support Text fields in REST v2 APIs

zhuwenxing:
issue: #52087

## What was changed

- Support `Text` fields in REST v2 insert and upsert requests.
- Return `Text` fields from REST v2 query and search responses.
- Reject missing or null values for non-nullable `Text` fields while preserving explicit empty strings and partial-update semantics.
- Add REST L0 coverage using 3,000 indexed sealed rows, 500 unindexed sealed rows, and 500 growing rows, with long `Text` payloads in every segment state.
- Add Go regression tests for request conversion and response serialization.

## Why this change

REST v2 did not recognize `Text` as a string-backed field during request conversion or response formatting, causing valid inserts to fail with error code `1804`.

## Validation

- HTTP server Go unit-test package passed across four workers.
- REST v2 Text LOB E2E tests passed.
- Verified insert and upsert behavior for missing, null, and explicit empty `Text` values.

### DIFF
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -3501,7 +3501,7 @@ func fieldDataValueCount(fieldData *schemapb.FieldData) (int64, error) {
 			return int64(len(fieldData.GetScalars().GetTimestamptzData().GetData())), nil
 		}
 		return int64(len(fieldData.GetScalars().GetStringData().GetData())), nil
-	case schemapb.DataType_String, schemapb.DataType_VarChar:
+	case schemapb.DataType_String, schemapb.DataType_VarChar, schemapb.DataType_Text:
 		return int64(len(fieldData.GetScalars().GetStringData().GetData())), nil
 	case schemapb.DataType_Array:
 		return int64(len(fieldData.GetScalars().GetArrayData().GetData())), nil
@@ -3739,9 +3739,7 @@ func buildQueryResp(rowsNum int64, needFields []string, fieldDataList []*schemap
 					} else {
 						row[fieldDataList[j].FieldName] = fieldDataList[j].GetScalars().GetStringData().GetData()[dataIdx]
 					}
-				case schemapb.DataType_String:
-					row[fieldDataList[j].GetFieldName()] = fieldDataList[j].GetScalars().GetStringData().GetData()[dataIdx]
-				case schemapb.DataType_VarChar:
+				case schemapb.DataType_String, schemapb.DataType_VarChar, schemapb.DataType_Text:
 					row[fieldDataList[j].GetFieldName()] = fieldDataList[j].GetScalars().GetStringData().GetData()[dataIdx]
 				case schemapb.DataType_BinaryVector:
 					row[fieldDataList[j].GetFieldName()] = fieldDataList[j].GetVectors().GetBinaryVector()[dataIdx*(fieldDataList[j].GetVectors().GetDim()/8) : (dataIdx+1)*(fieldDataList[j].GetVectors().GetDim()/8)]
--- a/internal/distributed/proxy/httpserver/utils.go
+++ b/internal/distributed/proxy/httpserver/utils.go
@@ -1385,7 +1385,7 @@ func checkAndSetData(body []byte, collSchema *schemapb.CollectionSchema, partial
 					reallyData[fieldName] = result
 				case schemapb.DataType_Timestamptz:
 					reallyData[fieldName] = dataString
-				case schemapb.DataType_VarChar, schemapb.DataType_String:
+				case schemapb.DataType_VarChar, schemapb.DataType_String, schemapb.DataType_Text:
 					value, err := stringFieldValue(fieldName, fieldValue, compatibilityMode)
 					if err != nil {
 						return reallyDataArray, validDataMap, err
@@ -2780,11 +2780,7 @@ func anyToColumns(rows []map[string]interface{}, validDataMap map[string][]bool,
 			data = make([]float32, 0, rowsLen)
 		case schemapb.DataType_Double:
 			data = make([]float64, 0, rowsLen)
-		case schemapb.DataType_Timestamptz:
-			data = make([]string, 0, rowsLen)
-		case schemapb.DataType_String:
-			data = make([]string, 0, rowsLen)
-		case schemapb.DataType_VarChar:
+		case schemapb.DataType_Timestamptz, schemapb.DataType_String, schemapb.DataType_VarChar, schemapb.DataType_Text:
 			data = make([]string, 0, rowsLen)
 		case schemapb.DataType_Array:
 			data = make([]*schemapb.ScalarField, 0, rowsLen)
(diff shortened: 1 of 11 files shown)


internal/distributed/proxy/httpserver/utils.go
```diff
@@ -1853,11 +1857,7 @@ func anyToColumns(rows []map[string]interface{}, validDataMap map[string][]bool,
 			data = make([]float32, 0, rowsLen)
 		case schemapb.DataType_Double:
 			data = make([]float64, 0, rowsLen)
-		case schemapb.DataType_Timestamptz:
-			data = make([]string, 0, rowsLen)
-		case schemapb.DataType_String:
-			data = make([]string, 0, rowsLen)
-		case schemapb.DataType_VarChar:
+		case schemapb.DataType_Timestamptz, schemapb.DataType_String, schemapb.DataType_VarChar, schemapb.DataType_Text:
```
liliu-z, line 2783: REST conversion now accepts Text, but partial upsert still dies downstream: when a batch contains at least one new primary key, task_upsert synthesizes placeholders for omitted nullable fields via GenNullableFieldData, whose switch has no Text case, so the default branch rejects the whole request with `undefined data type:Text`. Trigger is concrete — checkAndSetData skips omitted fields under partialUpdate, anyToColumns then drops the column entirely, and LackOfFieldsDataBySchema lets nullable fields through, so the request reaches GenNullableFieldData. The fix belongs next to the VarChar case in internal/proxy/task_upsert.go (it hits gRPC clients too), and worth noting that nothing in this PR drives Text through the upsert endpoint at all — neither the Go handler tests nor the new E2E file contain a single upsert.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -2761,9 +2737,7 @@ func buildQueryResp(rowsNum int64, needFields []string, fieldDataList []*schemap
 					} else {
 						row[fieldDataList[j].FieldName] = fieldDataList[j].GetScalars().GetStringData().GetData()[dataIdx]
 					}
-				case schemapb.DataType_String:
-					row[fieldDataList[j].GetFieldName()] = fieldDataList[j].GetScalars().GetStringData().GetData()[dataIdx]
-				case schemapb.DataType_VarChar:
+				case schemapb.DataType_String, schemapb.DataType_VarChar, schemapb.DataType_Text:
```
liliu-z, line 3742: The row renderer now emits Text, but a search that also passes orderByFields never gets there. searchWithOrderByPipe runs requery, then a lambda that assigns all output-field data into Results.FieldsData, then the order_by node; reorderResults iterates every FieldsData entry and reorderFieldData handles only VarChar/String, so a Text output field falls into the default branch and the request fails with `reorderFieldData: unhandled data type Text`. Plain search over the same field works, so the fix is one case in internal/proxy/search_pipeline.go. This is a proxy-layer gap that hits gRPC clients identically, not a REST-only defect, but it does leave this PR's "return Text from search responses" claim incomplete.

tests/restful_client_v2/testcases/test_text_lob.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+def assert_rows_payload(rows_by_id, expected, fields):
+    assert set(rows_by_id) == set(expected), f"row ids mismatch: actual={set(rows_by_id)}, expected={set(expected)}"
+    for pk, row in rows_by_id.items():
+        for field in fields:
+            assert_text_payload(row.get(field), expected[pk][field])
+
+
+@pytest.mark.xdist_group("TestTextLOB")
```
liliu-z, line 369: The xdist_group marker only groups tests when pytest-xdist runs with --dist loadgroup, and the restful_client_v2 CI job runs plain `-n 6`; only the python_client job passes --dist loadgroup. The marker is therefore ignored, the two tests in the class can land on different workers, and each worker rebuilds the entire class fixture — 4,000 rows including multi-MB Text payloads, five index builds and a load. There is no cross-worker collision on the collection itself, since gen_collection_name embeds a timestamp plus random suffix per process, so the cost is duplicated setup, not interference. Either add --dist loadgroup to that CI invocation or drop the marker so the file does not imply co-location it never gets.

tests/restful_client_v2/testcases/test_text_lob.py
```diff
@@ -0,0 +1,545 @@
...
+
+        unindexed_rows = self.shared_rows[INDEXED_SEALED_ROWS : INDEXED_SEALED_ROWS + UNINDEXED_SEALED_ROWS]
+        rsp = vector_client.vector_insert({"collectionName": self.collection_name, "data": unindexed_rows})
+        assert rsp["code"] == 0, rsp
+        assert rsp["data"]["insertCount"] == UNINDEXED_SEALED_ROWS
+        flush_collection(collection_client, self.collection_name)
+
+        rsp = index_client.index_create(build_dense_index_payload(self.collection_name))
+        assert rsp["code"] == 0, rsp
+        assert utility.wait_for_index_building_complete(
```
liliu-z, line 422: The fixture is far more patient than the budget CI gives it: five wait_for_index_building_complete calls at timeout=300 each, plus a 120s load wait and a 120s wait_for_sealed_layout, against a `--timeout 360` per-item budget. pytest-timeout's timer covers the setup phase, so a slow index build gets killed at 360s before any of the fixture's own waits can expire or report anything useful, and since the same CI command uses `-x`, that kills the whole restful v2 run rather than one case. Bring the internal waits in line with the 360s budget or give these tests an explicit larger timeout.

internal/distributed/proxy/httpserver/utils.go
```diff
@@ -773,11 +773,15 @@ func checkAndSetData(body []byte, collSchema *schemapb.CollectionSchema, partial
...
 					}
 					reallyData[fieldName] = result
-				case schemapb.DataType_Timestamptz:
-					reallyData[fieldName] = dataString
-				case schemapb.DataType_VarChar:
+				case schemapb.DataType_Timestamptz, schemapb.DataType_VarChar, schemapb.DataType_String:
 					reallyData[fieldName] = dataString
-				case schemapb.DataType_String:
+				case schemapb.DataType_Text:
+					if !fieldValue.Exists() {
```
liliu-z: The two new reject branches for missing required / null Text values are exercised only by Go unit tests; every Text field in the E2E suite is declared `nullable: True`, so no end-to-end test ever drives a missing-required or explicit-null Text value through the REST path. That leaves the guards' interaction with real request parsing and the surfaced error code unverified end-to-end. Add an E2E case with a non-nullable Text field that omits the value and one that sends null.

internal/distributed/proxy/httpserver/utils.go
zhuwenxing: Fixed. `GenNullableFieldData` now treats `DataType_Text` like `VarChar`. Added Go coverage and REST E2E coverage for inserting a new primary key while omitting a nullable Text field.

internal/distributed/proxy/httpserver/utils.go
zhuwenxing: Fixed. `reorderFieldData` now handles `DataType_Text`. Added proxy unit coverage and a REST dense-search case that uses `orderByFields` while returning Text LOB fields.

tests/restful_client_v2/testcases/test_text_lob.py
zhuwenxing: Fixed. Both REST v2 CI invocations now pass `--dist loadgroup`, so the `xdist_group` marker keeps the class-scoped TEXT LOB fixture on one worker.

tests/restful_client_v2/testcases/test_text_lob.py
zhuwenxing: Fixed without extending the CI timeout. The REST jobs continue to use `--timeout 360`. The shared fixture now enforces a single 300-second setup deadline across flush, index-build, load, and sealed-layout waits, leaving 60 seconds of headroom for test execution and teardown. The complete 5-case Loon FFI REST E2E run passed in 49.82 seconds.

internal/distributed/proxy/httpserver/utils.go
zhuwenxing: Fixed. Added REST E2E coverage for a non-nullable Text field. The parameterized case verifies that both an omitted value and an explicit null are rejected with code 1804 and field-specific messages.

MrPresent-Han (approved):
LGTM. No P0/P1 merge blockers found.

Issue #52088, opened 2026-07-31
[Bug]: Go SDK v3 lacks public TEXT field and column support


zhuwenxing:
## Environment

- Milvus version: Milvus 3.0 development line; server runtime is not required to reproduce this compile-time SDK gap
- Git commit: Go SDK `client/v3.0.0-beta` tag commit `bd6f2f8d1ce16e2aab4743831a2787cc4a940c23`; also verified on upstream `master` commit `6c8462cefb59b81b3332c622f80e384c3a6d6361`
- Deployment mode: N/A
- MQ type: N/A
- SDK version: `github.com/milvus-io/milvus/client/v3 v3.0.0-beta` (latest published module version)
- OS: macOS, darwin/arm64
- CPU/Memory: N/A
- GPU: N/A
- Others: Go 1.25.5; the SDK's proto dependency already exposes `schemapb.DataType_Text = 25`

### K8s Pod List

```text
N/A — this issue reproduces at Go compile time without starting or connecting to Milvus.
```

## Current Behavior

The latest published Go SDK v3 module does not expose `entity.FieldTypeText`. A schema using the Milvus 3.0 `TEXT` type therefore cannot be expressed through the public, typed Go SDK API:

```text
./main_test.go:12:23: undefined: entity.FieldTypeText
./main_test.go:13:30: undefined: entity.FieldTypeText
```

The gap is broader than the missing constant. Current Go SDK source also omits TEXT from:

- `entity.FieldType.Name`, `String`, and `PbFieldType`
- string-backed column construction and conversion
- column-based insert helpers
- row encode/decode type switches

Users could define the numeric value `entity.FieldType(25)` themselves, but that is an undocumented workaround and does not provide complete typed column/read/write support.

## Expected Behavior

Go SDK v3 should provide first-class public support for Milvus 3.0 `TEXT` fields. Users should be able to:

- declare a schema with an exported TEXT field type such as `entity.FieldTypeText`
- insert and upsert Go strings into TEXT fields through public row- and column-based APIs
- query TEXT fields and receive string-backed columns/values
- use the field type without importing protobuf internals or hard-coding enum value `25`

## Steps To Reproduce

1. Create the `go.mod` and `main_test.go` files shown below.
2. Run `go mod tidy`.
3. Run `go test ./...`.
4. Observe the compile error for `entity.FieldTypeText`.

## Minimal Reproduction Script

`go.mod`:

```go
module example.com/milvus-gosdk-text-repro

go 1.24.9

[...]

PR #52291, opened 2026-08-07
fix: add Go SDK v3 TEXT field support

zhuwenxing:
## What was changed

- Add Go SDK v3 schema and metadata support for the `Text` field type.
- Add scalar and nullable `Text` columns.
- Support `Text` in column-based writes, row conversion, query results, and search results.
- Add unit tests for schema, column conversion, write options, row conversion, and result decoding.
- Add comprehensive Go SDK L0 E2E coverage.

## E2E coverage

The TEXT L0 suite covers:

- 3,000 rows in an indexed sealed segment.
- 500 rows in an unindexed sealed segment.
- 500 rows in a growing segment.
- Standard and Jieba analyzers with BM25 functions and sparse indexes.
- HNSW search across all three execution paths.
- Empty, NULL, multilingual, 64 KiB boundary, and 1 MiB TEXT payloads.
- Exact payload validation using byte length, character count, and SHA-256.
- Query, dense search, BM25 search, and query iterator result decoding.
- Upsert transitions including NULL-to-large-TEXT, TEXT-to-large-TEXT, and TEXT-to-NULL.
- Deletion of a 1 MiB TEXT row and validation of all surviving rows.

## Validation

- Go SDK unit tests passed.
- Go SDK lint passed with 0 issues.
- `TestTextLOBPublicSDKL0` passed against standalone Milvus.
- DCO sign-offs are present.

issue: #52088

### DIFF
--- a/client/entity/field.go
+++ b/client/entity/field.go
@@ -52,6 +52,8 @@ func (t FieldType) Name() string {
 		return "String"
 	case FieldTypeVarChar:
 		return "VarChar"
+	case FieldTypeText:
+		return "Text"
 	case FieldTypeArray:
 		return "Array"
 	case FieldTypeJSON:
@@ -95,6 +97,8 @@ func (t FieldType) String() string {
 	case FieldTypeString:
 		return "string"
 	case FieldTypeVarChar:
+		return "string"
+	case FieldTypeText:
 		return "string"
 	case FieldTypeArray:
 		return "Array"
@@ -140,6 +144,8 @@ func (t FieldType) PbFieldType() (string, string) {
 		return "String", "string"
 	case FieldTypeVarChar:
 		return "VarChar", "string"
+	case FieldTypeText:
+		return "Text", "string"
 	case FieldTypeJSON:
 		return "JSON", "JSON"
 	case FieldTypeGeometry:
@@ -187,6 +193,8 @@ const (
 	FieldTypeJSON FieldType = 23
 	// FieldTypeGeometry field type Geometry
 	FieldTypeGeometry FieldType = 24
+	// FieldTypeText field type text
+	FieldTypeText FieldType = 25
 	// FieldTypeTimestamptz field type timestamptz
 	FieldTypeTimestamptz FieldType = 26
 	// FieldTypeBinaryVector field type binary vector
--- a/tests/go_client/testcases/text_lob_test.go
+++ b/tests/go_client/testcases/text_lob_test.go
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// L0 coverage for TEXT LOB through the public Go SDK.
+// L0 CRUD coverage for TEXT LOB through the public Go SDK.
 package testcases
 
 import (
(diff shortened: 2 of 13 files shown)


client/entity/field.go
```diff
@@ -96,6 +98,8 @@ func (t FieldType) String() string {
 		return "string"
 	case FieldTypeVarChar:
 		return "string"
+	case FieldTypeText:
```
liliu-z, line 101: `String()` now returns "string" for Text as well as VarChar/String, and the only caller that formats a column/field type mismatch uses `%v` (which resolves to the Stringer): `write_options.go:122` produces `param column doc has type string but collection field definition is string`. The PR makes this collision reachable for the first time by shipping two string-backed builders (`WithVarcharColumn`, `WithTextColumn`) whose columns are trivially swappable, so a user who picks the wrong one gets an error that names the same type on both sides. `String()` is fine as the Go-type reporter; format that message with `FieldType.Name()` instead so it reads "VarChar" versus "Text".

tests/go_client/testcases/text_lob_test.go
```diff
@@ -0,0 +1,730 @@
...
+	textLOBSparseField         = "content_sparse"
+	textLOBZHSparseField       = "content_zh_sparse"
+	textLOBSentinelSparseField = "content_sentinel_sparse"
+	textLOBVectorIndex         = "text_lob_vector_idx"
+	textLOBSparseIndex         = "text_lob_sparse_idx"
+	textLOBZHSparseIndex       = "text_lob_zh_sparse_idx"
+	textLOBSentinelSparseIndex = "text_lob_sentinel_sparse_idx"
+	textLOBVectorDim           = 16
+
+	textLOBIndexedSealedRows   = 3000
```
liliu-z, line 59: The indexed-vs-unindexed sealed split rests entirely on the server default `indexCoord.segment.minSegmentNumRowsToEnableIndex` (1024) sitting between 500 and 3000, but the test neither pins that config nor asserts any segment's real index state. The `pathCounts` check at lines 551-560 only re-buckets result IDs by the same ranges used at insert time, so it is tautological and cannot detect the loss: if the default moves, both segments end up indexed (or neither) and the suite still goes green having silently dropped the unindexed-path coverage it exists to provide. `client/entity/segment.go` exposes only ID/NumRows/State, so asserting real index state is not reachable from the public SDK — pin the threshold with `hp.AlterServerConfig`, which this file already uses at line 485.

client/entity/field.go
zhuwenxing: Fixed in `7e7e2c85b3`. The type mismatch message now formats both sides with `FieldType.Name()`, so `Text` and `VarChar` are distinguishable. I also added regression coverage for both mismatch directions.

tests/go_client/testcases/text_lob_test.go
zhuwenxing: Fixed in `bf91168ade`. The test now pins `indexCoord.segment.minSegmentNumRowsToEnableIndex` to `1024`, verifies the applied value, and restores the previous value during cleanup. The 500-row and 3000-row segments deterministically cover the below-threshold and above-threshold paths. The targeted Go SDK E2E passed.

tests/go_client/testcases/text_lob_test.go
```diff
@@ -0,0 +1,750 @@
...
+	require.GreaterOrEqual(t, textLOBIndexedSealedRows, textLOBMinRowsToEnableIndex)
+
+	previousStorageV3, err := hp.AlterServerConfig("common.storage.useLoonFFI", "true")
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		_, _ = hp.AlterServerConfig("common.storage.useLoonFFI", previousStorageV3)
+	})
+
+	minRowsValue := strconv.Itoa(textLOBMinRowsToEnableIndex)
+	previousMinRows, err := hp.AlterServerConfig(textLOBMinRowsConfig, minRowsValue)
```
liliu-z: The restore path here can write an empty value for a global server config. `hp.AlterServerConfig` discards the error from its internal GET and returns `("", nil)` whenever that read fails but the POST succeeds, so `previousMinRows` can be `""` while `require.NoError` at line 502 still passes; the deferred restore then POSTs `{"key":k,"value":""}`, which the handler treats as a set rather than a reset, etcd persists it as a high-priority empty override, and `GetAsInt` resolves it to 0 for the rest of the run. Two things narrow this from the original framing: the swallow lives in the shared helper this PR does not touch, and the identical capture/restore shape is already the repo idiom in add_field_test.go:82-91, partition_test.go:118-127 and search_test.go:352-358, so adding a shared `ResetServerConfig` is broader than this PR. The cheap local fix is to read the previous value with `hp.GetServerConfig` and `require.NoError` before altering; the same applies to the `common.storage.useLoonFFI` capture seven lines above at 494-498.

zhuwenxing: Fixed in `e6360dca71`. The test now treats both server settings as prerequisites and validates `common.storage.useLoonFFI=true` and `indexCoord.segment.minSegmentNumRowsToEnableIndex=1024` without mutating either value, so there is no restore path that can write an empty override. I also aligned the case naming and code style with the existing Go SDK suite. The targeted `TestTextFieldCRUD` case passed on a Loon-enabled deployment.

PR #52292, opened 2026-08-07
fix: [cp 3.0] prevent streaming cutover deadlock

chyezh:
issue: #52086
pr: #52123

- legacy RootCoord DML streams: serialize in-flight broadcasts with close during the streaming-service cutover
- streaming enable barrier: preserve listener completion ordering and wake parked waiters after cutover

### DIFF
--- a/internal/rootcoord/dml_channels.go
+++ b/internal/rootcoord/dml_channels.go
@@ -36,8 +36,9 @@ import (
 )
 
 type dmlMsgStream struct {
-	ms    msgstream.MsgStream
-	mutex sync.RWMutex
+	ms     msgstream.MsgStream
+	mutex  sync.RWMutex
+	closed bool
 
 	refcnt int64 // current in use count
 	used   int64 // total used counter in current run, not stored in meta so meant to be inaccurate
@@ -82,6 +83,43 @@ func (dms *dmlMsgStream) DecRefCnt() {
 	} else {
 		mlog.Warn(context.TODO(), "Try to remove channel with no ref count", mlog.Int64("idx", dms.idx))
 	}
+}
+
+func (dms *dmlMsgStream) broadcast(ctx context.Context, pack *msgstream.MsgPack) (map[string][]msgstream.MessageID, error) {
+	dms.mutex.RLock()
+	defer dms.mutex.RUnlock()
+
+	if dms.closed {
+		return nil, merr.WrapErrServiceNotReadyMsg("legacy dml msgstream is closed")
+	}
+	if dms.refcnt <= 0 {
+		return nil, nil
+	}
+	return dms.ms.Broadcast(ctx, pack)
+}
+
+func (dms *dmlMsgStream) broadcastMark(ctx context.Context, pack *msgstream.MsgPack) (map[string][]msgstream.MessageID, error) {
+	dms.mutex.RLock()
+	defer dms.mutex.RUnlock()
+
+	if dms.closed {
+		return nil, merr.WrapErrServiceNotReadyMsg("legacy dml msgstream is closed")
+	}
+	if dms.refcnt <= 0 {
+		return nil, merr.WrapErrServiceInternalMsg("channel not in use")
+	}
+	return dms.ms.Broadcast(ctx, pack)
+}
+
+func (dms *dmlMsgStream) close() {
+	dms.mutex.Lock()
+	defer dms.mutex.Unlock()
+
+	if dms.closed {
+		return
+	}
+	dms.closed = true
+	dms.ms.Close()
 }
 
 // channelsHeap implements heap.Interface to performs like an priority queue.
@@ -180,6 +218,7 @@ func newDmlChannels(initCtx context.Context, factory msgstream.Factory, chanName
 
 	for i, name := range names {
 		var ms msgstream.MsgStream
+		var closeNotifier *snmanager.StreamingReadyNotifier
 		if !streamingutil.IsStreamingServiceEnabled() {
 			ms = d.newMsgstream(initCtx, factory, name)
 		} else {
@@ -191,14 +230,7 @@ func newDmlChannels(initCtx context.Context, factory msgstream.Factory, chanName
 			if !notifier.IsReady() {
 				logger.Info(initCtx, "streaming service is not enabled, create a msgstream to use")
 				ms = d.newMsgstream(initCtx, factory, name)
-				go func() {
-					defer notifier.Release()
-					<-notifier.Ready()
-					// release the msgstream.
-					logger.Info(initCtx, "streaming service is enabled, release the msgstream...")
-					ms.Close()
-					logger.Info(initCtx, "streaming service is enabled, release the msgstream done")
-				}()
+				closeNotifier = notifier
 			} else {
 				logger.Info(initCtx, "streaming service has been enabled, msgstream should not be created")
 				notifier.Release()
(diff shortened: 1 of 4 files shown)


tinswzy:

Reviewed at head `790f44f` (3.0). **Backport fidelity: confirmed.** The net tree change of the three cherry-picks is byte-identical to master #52123 — same pre-image and post-image blob hashes for all four touched files (`dml_channels.go` `1a1fbfc2fd`→`e202693ed5`, `manager.go` `67df9fafe8`→`3c649beb53`, plus both test files) — so the 3.0 base was identical to master's base and nothing was adapted or dropped. No blocking issue; findings below are medium/low and mostly pre-existing or follow-up material.

What I verified as correct:
- **Send/Close serialization is complete at this layer.** All three uses of `dms.ms` in the package (`dml_channels.go:98`, `:111`, `:122`) are now behind `dms.mutex`, and nothing else in `internal/rootcoord` touches `.ms.` — grep confirms. This matters because `mqMsgStream.Broadcast` (`pkg/mq/msgstream/mq_msgstream.go:319`) takes **no** `closeRWMutex` and never checks `ms.closed`, so before this change `producer.Send` could run against a producer already closed by `Close()` (`:207-218`). The rootcoord-level RWMutex is currently the only thing preventing that.
- **Cutover barrier ordering is right.** Cancel → BlockUntilFinish → clear → `UnsafeBroadcast` all under `cm.cond.L` (`manager.go:349-357`). Holding the lock (restored by `5d6ba92` after `be5ade4` released it) also makes registration atomic against cutover: `RegisterStreamingEnabledNotifier` (`manager.go:223`) either lands in the list before the loop or observes `streamingVersion != nil` and cancels immediately, so `newDmlChannels` can never create a msgstream that nobody will close. Keeping the three commits unsquashed is correct per repo convention even though `be5ade4` alone transiently removes the barrier.
- **No new deadlock cycle.** I traced every registered listener on this branch (`dml_channels.go:226`, `root_coord.go:284`, `timeticksync.go:285`): none of their completion paths re-enter `ChannelManager`, so nothing can block on `cm.cond.L` while the cutover waits for it. Likewise no cycle through `d.mut` or `timetickSync.lock` (both are acquired before, never across, a broadcast).

---

### Medium — the cutover barrier now waits on legacy MQ I/O while holding `cm.cond.L`
`manager.go:332-355` + `dml_channels.go:114-116`, `dml_channels.go:211`

```go
cm.cond.L.Lock()
defer cm.cond.L.Unlock()
...
for _, notifier := range cm.streamingEnableNotifiers {
    notifier.BlockUntilFinish()   // -> dms.close() -> dms.mutex.Lock()
}
```

`dms.close()` takes the write lock, so it waits for any in-flight `ms.Broadcast`. That broadcast was issued with `d.ctx`, which is `context.TODO()` (`dml_channels.go:211`) — it cannot be canceled by shutdown or a deadline, so the wait ends only when the MQ client's own send timeout fires (pulsar-client-go default 30s, no override in `pkg/mq`).

Failure scenario: legacy Pulsar/Kafka brokers become unreachable exactly during cutover. A timetick broadcast parks in `producer.Send`; `dms.close()` blocks; `MarkStreamingHasEnabled` holds `cm.cond.L` for the full send timeout. Every `ChannelManager` entry point I checked takes that same lock — `AllocVirtualChannels` (so rootcoord `CreateCollection` at `create_collection_task.go:781` stalls), `AddPChannels`, `TriggerWatchUpdate`, `RegisterStreamingEnabledNotifier`, and the `WatchChannelAssignments` wakeup path — so a legacy-MQ hiccup becomes a cluster-wide streamingcoord assignment stall. Secondary: `addChannels` holds `d.mut` while calling `IncRefcnt()`, which is blocked behind the pending writer, so rootcoord channel bookkeeping stalls in parallel.

The lock scope itself is pre-existing on the 3.0 base; what is new is a blocking MQ-dependent step inside it. Note `be5ade4`'s approach (release the lock before `BlockUntilFinish`) is not the fix — it loses the barrier, which is why `5d6ba92` reverted it. Suggested follow-up that keeps both properties: add an explicit `streamingEnablePublishing bool`, set it under the lock, release the lock for the `BlockUntilFinish` loop, and have `WaitUntilStreamingEnabled`/`IsStreamingEnabledOnce`/the watch path treat "version set **and** not publishing" as the enabled condition; re-acquire to clear the flag and broadcast. Alternatively bound `close()` with a deadline and log a loud warning on expiry.

### Medium — `SaveVersion` failure leaves the cutover permanently half-done
`manager.go:339-346`

```go
cm.streamingVersion = &streamingpb.StreamingVersion{Version: StreamingVersion260}
if err := ...SaveVersion(ctx, cm.streamingVersion); err != nil {
    return err            // in-memory field stays set
}
```

Failure scenario: one etcd write times out. `cm.streamingVersion` is already non-nil, so (a) the listeners are never canceled — rootcoord's legacy msgstreams stay open and both timetick loops keep producing to the legacy MQ after the cluster believes it cut over; (b) `IsStreamingEnabledOnce()` returns true and newly added pchannels get `AccessModeRW` (`manager.go:~287`) while the version is not durable, so a coordinator restart reverts the cluster to "streaming never enabled"; (c) the new `cm.cond.UnsafeBroadcast()` never runs, so goroutines parked in `WaitUntilStreamingEnabled` stay parked even though `IsStreamingEnabledOnce()` already reports true; (d) there is no retry — `balancer_impl.go:381-384` logs a Warn and `return`s out of the whole `execute()` background loop, and a later `MarkStreamingHasEnabled` would hit the `streamingVersion != nil` early return at `:335` and report success without persisting or notifying anything.

Pre-existing, but it is inside the function this PR rewrites and it defeats exactly the barrier being fixed. One-line fix: build into a local and assign `cm.streamingVersion` only after `SaveVersion` succeeds (or reset to `nil` on the error path).

### Low — the new closed-guard misses the `ms == nil` case
`dml_channels.go:92-98` / `:105-111` vs `:234-237`

When streaming is already enabled at startup, the else branch releases the notifier and leaves `ms` nil, so `dms.ms == nil` with `closed == false`. `newTimeTickSync` then calls `dmlChannels.addChannels(chanNames...)` for every recovered collection (`timeticksync.go:128`), so `refcnt > 0`. Any `broadcast`/`broadcastMark` on such a channel falls straight through both guards into `dms.ms.Broadcast(...)` → nil-interface panic in rootcoord.

This is unreachable on this branch today only by accident: both timetick loops return early when the notifier is already ready (`root_coord.go:287`, `timeticksync.go:291`), and `broadcastDmlChannels`/`broadcastMarkDmlChannels` (`timeticksync.go:426`, `:431`) have no callers. Since the PR is adding precisely this class of guard, making it total is one line:

```go
if dms.closed || dms.ms == nil {
    return nil, merr.WrapErrServiceNotReadyMsg("legacy dml msgstream is not available")
}
```

### Low — retriable error class for a permanently closed stream
`dml_channels.go:93`, `:106`

`ErrServiceNotReady` maps to `ErrorCode_NotReadyServe` (`pkg/util/merr/utils.go:180`) and reads as transient, but a closed legacy stream never reopens. No current consumer retries (the only live caller is `sendTimeTickToChannel`, whose error is logged and dropped at `timeticksync.go:336-337`), so this is a latent trap rather than a live bug — but a future `retry.Do` caller on this path would spin forever. Worth either a comment at the guard or a non-retriable code.

### Low / nit — collateral behavior changes in `broadcastMark`, and error-log volume
`dml_channels.go:385-389`

Previously the not-in-use case returned `nil, merr.WrapErrServiceInternalMsg("channel not in use: %s", chanName)`; it now returns the **partially filled** `result` map alongside the error, and the channel name is re-attached only by `merr.Wrapf` at the call site. Harmless on this branch (no callers), but callers must not consume the map when `err != nil`. Separately, `d.broadcast` logs `mlog.Error` per channel per timetick during the cutover window — until both timetick loops observe `Ready()` this is a burst of ERROR lines for an expected transition; `Warn` would read better.

### Nit — regression tests don't cover the wiring that actually regressed
The three new tests are well-targeted at the primitives (`close` waits for an in-flight broadcast and rejects afterwards; the barrier holds; parked waiters are woken), and `TestMarkStreamingHasEnabledWakesParkedWaiters` correctly exercises the `UnsafeBroadcast` added by `790f44f`. But nothing covers `newDmlChannels`'s listener goroutine (`dml_channels.go:249-259`) — the actual defect was that goroutine calling `ms.Close()` directly, and reverting `stream.close()` back to `ms.Close()` there would leave all three new tests green. A small test that drives a notifier through `newDmlChannels` (or asserts `dms.closed` after the notifier fires) would close that gap. Two minor points: `TestDmlMsgStreamCloseWaitsForBroadcast` leaks the mocked-broadcast goroutine blocked on `<-releaseBroadcast` if the first `t.Fatal` fires; and all three tests hinge on 100ms sleeps, so they are the likeliest to flake first on a loaded runner.

czs007 (approved):
/lgtm

PR #52293, opened 2026-08-07
fix: [cp 2.6] prevent streaming cutover deadlock

chyezh:
issue: #52086
pr: #52123

- legacy RootCoord DML streams: serialize in-flight broadcasts with close during the streaming-service cutover
- streaming enable barrier: preserve listener completion ordering and wake parked waiters after cutover

### DIFF
--- a/internal/rootcoord/dml_channels.go
+++ b/internal/rootcoord/dml_channels.go
@@ -115,12 +115,13 @@ func (dms *dmlMsgStream) broadcastMark(ctx context.Context, pack *msgstream.MsgP
 
 func (dms *dmlMsgStream) close() {
 	dms.mutex.Lock()
-	defer dms.mutex.Unlock()
-
 	if dms.closed {
+		dms.mutex.Unlock()
 		return
 	}
 	dms.closed = true
+	dms.mutex.Unlock()
+
 	dms.ms.Close()
 }
 
--- a/internal/rootcoord/dml_channels.go
+++ b/internal/rootcoord/dml_channels.go
@@ -38,8 +38,9 @@ import (
 )
 
 type dmlMsgStream struct {
-	ms    msgstream.MsgStream
-	mutex sync.RWMutex
+	ms     msgstream.MsgStream
+	mutex  sync.RWMutex
+	closed bool
 
 	refcnt int64 // current in use count
 	used   int64 // total used counter in current run, not stored in meta so meant to be inaccurate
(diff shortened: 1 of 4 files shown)


czs007:

Reviewed at head 0c864b3f21 (three commits, cherry-picked from PR #52123). Overall: the backport is faithful and internally consistent — no blocking code issues found in the changed lines. The only ordering concern is that the source PR #52123 (master) is still open; this cherry-pick should land after it merges (details below).

**Correctness assessment of the final state**

- **Send/Close serialization** (`internal/rootcoord/dml_channels.go:90-125`): `dmlMsgStream.broadcast`/`broadcastMark` now hold `dms.mutex.RLock()` across `ms.Broadcast`, and `close()` takes the write lock before `ms.Close()`. This guarantees `Producer.Close` can never run concurrently with an in-flight synchronous `Send`, which is exactly the pulsar-client-go race described in #52086 (event loop processes `closeProducer` before a buffered `dataChan` request, so the Send callback is never invoked). Post-close broadcasts fail fast with `ErrServiceNotReady` instead of touching a closed stream. All `dms.ms` accesses in the file are now under the mutex; no bypass paths remain (`stream.close()` at `dml_channels.go:258` is the only close path).
- **Barrier ordering** (`internal/streamingcoord/server/balancer/channel/manager.go:331-359`): the intermediate commit (eb7d61d6a4) released `cm.cond.L` before `Cancel`/`BlockUntilFinish`, which would have let concurrent callers observe `streamingVersion != nil` while legacy msgstreams were still open (double-writer window between legacy timetick producers and the WAL). Commit 9e6973ab24 correctly reverts that and keeps the lock held across the whole cutover; the deadlock is instead broken at the `dmlMsgStream` layer, where it actually originated. The listener finish paths (`dml_channels.go` close goroutine, `timeticksync.go:279` `startWatch`, `root_coord.go` `startTimeTickLoop`) only depend on `dms.mutex` and Pulsar send completion — none of them re-enter `cm.cond.L` — so holding the barrier lock across `BlockUntilFinish` introduces no lock cycle.
- **Waiter wake-up** (`manager.go:357`): `cm.cond.UnsafeBroadcast()` is called with `cm.cond.L` held (required by its contract) and fixes a real gap — before this, `WaitUntilStreamingEnabled` waiters were only woken incidentally by unrelated broadcasts such as `TriggerWatchUpdate`.
- **2.6 adaptation**: correctly translated from master's `mlog`/`pkg/v3` idioms to 2.6's `log.Ctx`/`pkg/v2`; `ContextCond.UnsafeBroadcast` and `merr.Wrapf` both exist in 2.6's `pkg/v2`. Behavior of the refactored `broadcast` (silent skip on `refcnt <= 0`) and `broadcastMark` (error on `refcnt <= 0`, channel name preserved via `merr.Wrapf` at the call site) matches the pre-change semantics.
- **Tests**: the three regression tests are deterministic (channel-based synchronization, no sleeps on the positive path) and each pins one property: close-waits-for-in-flight-broadcast + fail-fast-after-close (`dml_channels_test.go`), publish-after-listener-completion barrier, and parked-waiter wake-up (`manager_test.go`). `TestDmlMsgStreamCloseWaitsForBroadcast` uses mockey, so it needs the `-gcflags="all=-N -l"` build flags — the standard go-unittest CI target already passes them.

**Findings**

- **medium (process)** — The cherry-picked commits (2a53233e70, df21829013, 1077eccd5e) are not on master yet; PR #52123 is still open with no approvals. Recommend holding this backport until #52123 merges, and re-picking if it changes during review, to avoid the 2.6 branch diverging from the eventual master fix.
- **medium (pre-existing, adjacent — suggest fixing in #52123 first)** — `manager.go:339-346`: `cm.streamingVersion` is assigned before `SaveVersion`, and on persist failure it is not rolled back. Failure scenario: a transient etcd error during the one-time cutover leaves the in-memory version set; every subsequent `MarkStreamingHasEnabled` call then early-returns `nil` at `manager.go:335`, so the version is never persisted, notifiers are never cancelled (legacy msgstreams stay open), and no broadcast ever wakes `WaitUntilStreamingEnabled` waiters — while callers in `balancer_impl.go:427/459` observe success. Suggested fix: set `cm.streamingVersion = nil` before returning the error, mirroring the rollback pattern already used for `SavePChannels` failure in `AddChannel` in the same file. Not introduced by this PR (the base has the identical pattern), so not blocking for the backport, but it sits exactly in the function this PR touches three times.
- **nit** — `dml_channels.go:236-247`: when rootcoord starts with streaming already enabled, `dms.ms` remains nil and `closed` remains false; a legacy-path broadcast with `refcnt > 0` would nil-deref instead of returning the new `ErrServiceNotReady`. This is pre-existing (the old code had the same nil `ms`), and the refcnt guard makes it unreachable in practice, but initializing `closed = true` for the no-msgstream branch would make the invariant self-enforcing. Only worth doing upstream, not as a backport divergence.

No other substantive issues. Mergeable once #52123 lands on master.

czs007 (approved):
/lgtm

Issue #52290, opened 2026-08-07
[Bug]: Milvus can stay alive after a fatal native-thread signal when running as PID 1


chyezh:
## Environment

- **Affected Milvus branches**: `master`, `3.0`, and `2.6`
- **Current dependency**: `github.com/milvus-io/cgosymbolizer v0.0.0-20250318084424-114f4050c3a6`
- **Deployment Mode**: Linux container or Kubernetes pod where the Milvus process is PID 1 in its PID namespace
- **Architecture confirmed by the upstream reproducer**: Linux/arm64

### Steps

1. Run a Milvus component directly as PID 1 in a Linux PID namespace, without an init process such as `tini` or `dumb-init`.
2. Trigger a synchronous fatal signal, such as `SIGSEGV`, on a foreign C/C++ pthread used by native Milvus code.
3. Observe the process and the operation whose native worker faulted.

The underlying runtime behavior can also be reproduced without a container by running a cgo program whose C-created pthread faults as:

```bash
unshare --user --pid --fork ./cgo-foreign-thread-fault
```

### Trigger Conditions

- **Frequency**: deterministic after a foreign pthread synchronously faults while the process is PID 1
- **Does NOT happen when**: an init process is PID 1, or the same program runs as an ordinary process and terminates on the fatal signal

## Expected Behavior

Milvus should terminate after the fatal native-thread signal so that Kubernetes or the container runtime can restart the process. The operation must not remain blocked behind a native task that can never complete.

## Actual Behavior

Milvus can print a complete fatal-signal crash dump and then remain Running and Ready with no restart. The foreign thread repeatedly returns to the same faulting instruction, while the task held by that thread never completes and its caller waits indefinitely. Health probes can remain green, making the failure appear as an isolated stuck operation rather than a process crash.

## Error Logs

The process emits the cgosymbolizer fatal-signal output, including text similar to:

```text
SIGNAL CATCH BY NON-GO SIGNAL HANDLER
SIGNO: 11; SIGNAME: Segmentation fault
BACKTRACE:
...
```

Despite the fatal-signal output, the process remains alive.

## Analysis Hints

For a fault on a foreign pthread, the Go runtime reaches `badsignal` and `raisebadsignal`. It temporarily installs `SIG_DFL` and calls `raise`, which uses `tgkill` and therefore generates a sent signal.

[...]

PR #52298, opened 2026-08-07
enhance: bump cgosymbolizer to fix PID 1 native-fault hangs

chyezh:
issue: #52290

- Go modules: upgrade cgosymbolizer in both the root and pkg modules so a fatal foreign-thread fault terminates PID 1 instead of looping indefinitely

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -317,7 +317,7 @@ replace (
 	github.com/expr-lang/expr => github.com/SimFG/expr v0.0.0-20250513112851-9b981e8400b9
 	github.com/go-kit/kit => github.com/go-kit/kit v0.1.0
 	github.com/greatroar/blobloom => github.com/milvus-io/blobloom v0.0.0-20240603110411-471ae49f3b93
-	github.com/ianlancetaylor/cgosymbolizer => github.com/milvus-io/cgosymbolizer v0.0.0-20250318084424-114f4050c3a6
+	github.com/ianlancetaylor/cgosymbolizer => github.com/milvus-io/cgosymbolizer v0.0.0-20260807043507-3952bb2c1554
 	github.com/milvus-io/milvus/pkg/v3 => ./pkg
 	github.com/tecbot/gorocksdb => github.com/milvus-io/gorocksdb v0.0.0-20220624081344-8c5f4212846b // indirect
 )
--- a/go.mod
+++ b/go.mod
@@ -317,7 +317,7 @@ replace (
 	github.com/expr-lang/expr => github.com/SimFG/expr v0.0.0-20250513112851-9b981e8400b9
 	github.com/go-kit/kit => github.com/go-kit/kit v0.1.0
 	github.com/greatroar/blobloom => github.com/milvus-io/blobloom v0.0.0-20240603110411-471ae49f3b93
-	github.com/ianlancetaylor/cgosymbolizer => github.com/milvus-io/cgosymbolizer v0.0.0-20260807043507-3952bb2c1554
+	github.com/ianlancetaylor/cgosymbolizer => github.com/milvus-io/cgosymbolizer v0.0.0-20260810014817-f3b338cf6824
 	github.com/milvus-io/milvus/pkg/v3 => ./pkg
 	github.com/tecbot/gorocksdb => github.com/milvus-io/gorocksdb v0.0.0-20220624081344-8c5f4212846b // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -796,8 +796,8 @@ github.com/milvus-io/arrow/go/v17 v17.0.1 h1:juKuEgTNQbe8a+LuBnyt9/UkfllTwtQIB+K
 github.com/milvus-io/arrow/go/v17 v17.0.1/go.mod h1:jR7QHkODl15PfYyjM2nU+yTLScZ/qfj7OSUZmJ8putc=
 github.com/milvus-io/blobloom v0.0.0-20240603110411-471ae49f3b93 h1:xnIeuG1nuTEHKbbv51OwNGO82U+d6ut08ppTmZVm+VY=
 github.com/milvus-io/blobloom v0.0.0-20240603110411-471ae49f3b93/go.mod h1:mjMJ1hh1wjGVfr93QIHJ6FfDNVrA0IELv8OvMHJxHKs=
-github.com/milvus-io/cgosymbolizer v0.0.0-20260807043507-3952bb2c1554 h1:aBX1V60iDqjg/lDYcFVsXJYuDR04oImva2VwTQsfXcI=
-github.com/milvus-io/cgosymbolizer v0.0.0-20260807043507-3952bb2c1554/go.mod h1:vhIW04/SLFIXBKmm4O9hMbo6PaAnaLHfezQEtPRVgLs=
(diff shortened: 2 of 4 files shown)


tinswzy:

Reviewed at head `3554ef7525`. The dependency chain is correct and the change is minimal, but there is one real gap that should be fixed before merge: **`pkg/go.mod` was not bumped**, and `pkg` is the module that actually contains the only import site of the symbolizer. Everything else below is a request for information / non-blocking.

**What I verified holds.** The replace does reach the shipped binary: `cmd/roles/roles.go:60` → `pkg/util/symbolizer/symbolizer.go:22` → `github.com/benesch/cgosymbolizer` → (linux) that package is a one-line shim `import _ "github.com/ianlancetaylor/cgosymbolizer"` → rewritten by `go.mod:320`. Since the root module is the main module, its `replace` block governs the whole build graph including the `./pkg` directory replacement, so `milvus` binaries built from root do link the new revision. `go.sum:799-800` carries both the `h1:` and `/go.mod` hashes, and nothing else in the graph pins `milvus-io/cgosymbolizer`. The mechanism the fix targets is also real: in the current revision, `symbolizer.c:157` `wrapHandler` prints the C backtrace and then chains to the saved Go handler, which for a fault on a **non-Go thread** ends in `runtime.raisebadsignal` — that sets `SIG_DFL`, `raise()`s, and if the process does not die it *restores the Go handler and returns*. As PID 1 (`SIGNAL_UNKILLABLE`, and a userspace `raise()` is not a forced kernel delivery) the signal is dropped, the handler returns, the faulting instruction re-executes, and you get an unbounded fault/backtrace loop instead of a crash. So the diagnosis and the fix location are right.

---

### high — `pkg/go.mod` / `pkg/go.sum` still pin the buggy revision

`pkg/go.mod:251`
```
github.com/ianlancetaylor/cgosymbolizer => github.com/milvus-io/cgosymbolizer v0.0.0-20250318084424-114f4050c3a6
```
plus `pkg/go.sum:502-503` with the old hashes. Root and `pkg` now resolve the same import path to two different module versions — the kind of drift neither `go mod tidy` nor CI flags, because neither module is internally inconsistent.

Concrete failure scenarios:

1. **Any build where `pkg` is the main module links the old symbolizer.** `scripts/run_go_unittest.sh:122` does `pushd pkg` before running the package's tests (`make test-pkg`, and `ci-v2/ut-go`), and `Makefile:227` `cd pkg` for lint. The only import site of the symbolizer in this repo (`pkg/util/symbolizer/symbolizer.go:22`) lives inside `pkg`, so in exactly those runs the unfixed handler is what gets compiled in.
2. **Downstream consumers of `github.com/milvus-io/milvus/pkg/v3` get the buggy revision.** `replace` in a non-main module is ignored, so consumers must copy the replace block out of `pkg/go.mod` — which after this PR points at the revision this PR exists to move off. Anyone importing `pkg/v3` (including other milvus-io repos) keeps the hang.
3. The next person to bump this has two sources of truth and will likely fix only one again.

**Fix:** apply the identical two-line change to `pkg/go.mod:251` and `pkg/go.sum:502-503`. Worth checking whether a CI guard can assert that shared `replace` lines are identical across the root and `pkg` module — this pair has drifted silently before.

### medium — the entire behavioral change is unauditable from this PR

The diff is a pointer swap; 100% of the actual fix is in `milvus-io/cgosymbolizer` between `114f4050c3a6` and `3952bb2c1554`, and the PR body links neither the fork commit nor a fork PR. For code that runs *inside a fatal signal handler on an arbitrary foreign thread*, a reviewer needs to see at minimum:

- which signals are now covered (today `cgoInstallNonGoHandler()` installs only `SIGSEGV`/`SIGBUS`/`SIGFPE` — `SIGILL`, `SIGABRT`, `SIGTRAP` are not wrapped, and a C++ `std::terminate`/`abort()` on a foreign thread would take a different path);
- that the new termination path is async-signal-safe (`fprintf`/`strsignal` in `printBacktrace` at `symbolizer.c:141` already are not, but adding a non-safe call *after* a decision to die is a different risk than before);
- that faults on **Go** threads still chain to the Go runtime handler unchanged, i.e. the fix is scoped to the foreign-thread branch and does not alter normal `SIGSEGV`-in-Go-code panics.

**Fix:** paste the fork diff (it should be small) or link the fork commit/PR in the body.

### medium — please state the new termination path: core dump and exit status

`pkg/util/symbolizer/symbolizer.go:26` sets `debug.SetTraceback("crash")` specifically so a fatal fault re-raises under `SIG_DFL` and produces a core dump. If the new revision terminates with `_exit(N)`/`quick_exit` instead of re-raising, then in the normal (non-PID-1) case we silently lose the core dump, and a container's terminated state changes from `Signal: 11` to a plain exit code — which affects crash triage, `RUNTIME`/OOM-vs-SIGSEGV attribution, and anything keying on the signal. If instead it re-raises and only falls back to `_exit` when the raise demonstrably failed to kill the process, that is the right shape and the blast radius is confined to the PID-1 case.

**Fix:** one line in the PR body stating which of the two it is, plus the exit code used. Per the repo's verification-gate convention, claim only what was reproduced — "terminates instead of looping under PID 1" is verifiable; "no change to crash-dump behaviour elsewhere" needs the non-PID-1 case checked too.

### low — scope: no image shipped from this repo runs milvus as PID 1

Every image sets `ENTRYPOINT ["/tini", "--"]` (`build/docker/milvus/ubuntu22.04/Dockerfile:61`, and the same in `ubuntu20.04`, `rockylinux9`, `amazonlinux2023`, both GPU variants), and `deployments/docker/standalone/docker-compose.yml:42` supplies `command: ["milvus","run","standalone"]` on top of that entrypoint — so in the supported deployments milvus is PID 2+, where `raise()` under `SIG_DFL` kills normally and the loop does not occur. The hang needs an entrypoint override (`docker run --entrypoint milvus …`, a k8s `command:` override, or a custom image with no init).

This does not make the fix wrong — it is the right defense, and users do run those configurations — but naming the exact reproducer in the body keeps a future reader from concluding the code is dead. If the report came from a supported deployment, that is important information the body currently omits.

### low — macOS builds are not covered by this bump

`github.com/benesch/cgosymbolizer` only forwards to `ianlancetaylor/cgosymbolizer` on Linux (`cgosymbolizer_linux.go`); the darwin implementation is its own and is not replaced. So a developer reproducing a foreign-thread fault on a mac build gets the old behaviour. Fine to leave as-is, worth a note if the tracking issue mentions macOS.

### nit — commit type vs PR title

The single commit's subject is `fix: bump cgosymbolizer to prevent PID 1 hangs (#52290)` while the PR title is `build(deps): bump cgosymbolizer to fix PID 1 native-fault hangs`. CI lints the PR title, so this is harmless, but please confirm `#52290` in `issue: #52290` is the tracking issue rather than a PR number — the `(#NNNNN)` suffix form in the commit subject usually denotes the latter.

---

**Summary of requested changes:** (1) bump `pkg/go.mod:251` + `pkg/go.sum:502-503` to the same revision — blocking; (2) link the fork diff and state the termination path / exit code in the body; (3) add the reproducer used to validate the PID-1 case. With (1) done and (2)/(3) answered, this is good to merge.

czs007 (approved):
/lgtm

PR #52301, opened 2026-08-07
fix: avoid probing absent TTL during schema bump (#51649)

MrPresent-Han:
### What

Fix a panic in schema-bump full rewrite when the target schema designates a TTL field that an old segment does not physically contain.

The old code used `record.Column(ttlFieldID)` to determine source-field presence. V2/V3 records panic when `Column` is called for a field they do not carry. Presence now comes from `existingFields`, the physical field set derived from the segment manifest or binlogs, and the loop-invariant filter setup is computed before reading records.

This PR intentionally does not apply the TTL schema default to old rows and does not change TTL expiration semantics.

issue: #51649

### Tests

- `TestFullRewriteMissingTTLFieldDoesNotProbeRecord`: real-cgo regression covering a V3 full rewrite whose source segment lacks the target TTL field; compaction completes and preserves all three rows.
- `TestSelectFullRewriteRecordDropsDeletedAndExpiredRows`: existing physical TTL filtering remains unchanged.
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datanode/compactor -run '^TestBumpSchemaVersionCompactionTaskSuite$'`

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -338,35 +338,35 @@ func (t *bumpSchemaVersionCompactionTask) fullRewriteSegmentID() (int64, error) 
 	return idRange.GetBegin(), nil
 }
 
-func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchema, entityFilter compaction.EntityFilter, ttlFieldID int64, useTTLField bool, ttlValues []int64) (*recordSelection, []int64, error) {
+func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchema, entityFilter compaction.EntityFilter, ttlFieldID int64, useTTLField bool) (*recordSelection, error) {
 	pkArray := record.Column(pkField.GetFieldID())
 	var pkAt func(int) any
 	switch pkField.GetDataType() {
 	case schemapb.DataType_Int64:
 		int64Array, ok := pkArray.(*array.Int64)
 		if !ok {
-			return nil, nil, merr.WrapErrServiceInternal("int64 primary key field not found in full schema rewrite record")
+			return nil, merr.WrapErrServiceInternal("int64 primary key field not found in full schema rewrite record")
 		}
 		pkAt = func(i int) any { return int64Array.Value(i) }
 	case schemapb.DataType_VarChar:
 		stringArray, ok := pkArray.(*array.String)
 		if !ok {
-			return nil, nil, merr.WrapErrServiceInternal("varchar primary key field not found in full schema rewrite record")
+			return nil, merr.WrapErrServiceInternal("varchar primary key field not found in full schema rewrite record")
 		}
 		pkAt = func(i int) any { return stringArray.Value(i) }
 	default:
-		return nil, nil, merr.WrapErrServiceInternal("invalid primary key data type for full schema rewrite")
+		return nil, merr.WrapErrServiceInternal("invalid primary key data type for full schema rewrite")
 	}
 
 	timestampArray, ok := record.Column(common.TimeStampField).(*array.Int64)
 	if !ok {
-		return nil, nil, merr.WrapErrServiceInternal("timestamp field not found in full schema rewrite record")
+		return nil, merr.WrapErrServiceInternal("timestamp field not found in full schema rewrite record")
 	}
 	var ttlArray *array.Int64
 	if useTTLField {
 		ttlArray, ok = record.Column(ttlFieldID).(*array.Int64)
 		if !ok {
-			return nil, nil, merr.WrapErrServiceInternal("TTL field not found in full schema rewrite record")
+			return nil, merr.WrapErrServiceInternal("TTL field not found in full schema rewrite record")
 		}
 	}
 
@@ -388,9 +388,6 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
 			continue
 		}
 
-		if useTTLField && expireTs > 0 {
-			ttlValues = append(ttlValues, expireTs)
-		}
 		if sliceStart == -1 {
 			sliceStart = i
 		}
@@ -400,9 +397,9 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
 		selection.length += record.Len() - sliceStart
 	}
 	if filteredRows == 0 {
-		return nil, ttlValues, nil
+		return nil, nil
 	}
-	return selection, ttlValues, nil
+	return selection, nil
 }
 
 func (t *bumpSchemaVersionCompactionTask) runSchemaVersionBumpOnly() *datapb.CompactionPlanResult {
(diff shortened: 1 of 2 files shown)


internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -367,6 +367,8 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
 		expireTs := int64(-1)
 		if useTTLField && ttlArray.IsValid(i) {
 			expireTs = ttlArray.Value(i)
+		} else if !useTTLField && useTTLDefault {
+			expireTs = ttlDefault
```
buqian-zilliz: **[P2] Distinguish negative timestamps from the no-TTL sentinel**

TIMESTAMPTZ defaults may legitimately precede 1970, so ttlDefault can be negative. Passing that value through expireTs reaches EntityFilter.isEntityExpiredByTTLField, which treats every value < 0 as TTL disabled; query execution, however, keeps only NULL TTLs or timestamps greater than now. An absent-column segment with a pre-1970 default therefore retains and rewrites rows that are already logically expired. Carry TTL presence separately from the numeric value so valid negative defaults are compared against currentTime.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -390,9 +389,24 @@ func selectFullRewriteRecord(record storage.Record, pkField *schemapb.FieldSchem
...
+
+func fullRewriteTTLDefault(schema *schemapb.CollectionSchema, ttlFieldID int64) (int64, bool, error) {
+	if ttlFieldID < common.StartOfUserFieldID {
+		return 0, false, nil
+	}
+	ttlField := typeutil.GetField(schema, ttlFieldID)
+	if ttlField == nil || ttlField.GetDefaultValue() == nil {
+		return 0, false, nil
+	}
+	value, ok := ttlField.GetDefaultValue().GetData().(*schemapb.ValueField_TimestamptzData)
```
buqian-zilliz: **[P2] Handle the accepted empty-string timestamp default**

fullRewriteTTLDefault treats every non-TimestamptzData oneof as corrupted, but the current DDL path can persist StringData(""): checkFieldSchema accepts string defaults for TIMESTAMPTZ and CheckAndRewriteTimestampTzDefaultValue returns success without rewriting an empty string. Because this helper is called for every full rewrite, that server-accepted schema now fails deterministically with DataIntegrity, even when the source already carries the TTL column and the default is unused. Either reject or normalize empty TIMESTAMPTZ defaults before persistence, or handle this representation here.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Resolved by narrowing this PR to the panic fix. The absent-column TTL-default filtering path has been removed; full rewrite now uses existingFields only to avoid probing a physically missing column. TTL default semantics will be handled in a separate PR.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Resolved by narrowing this PR to the panic fix. fullRewriteTTLDefault and its default-value validation have been removed, so this PR no longer interprets TIMESTAMPTZ defaults. That behavior will be handled separately under the historical-TTL invariant.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -504,6 +504,11 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		}
 	}()
 
+	// Source TTL presence is decided from existingFields, never by probing the
+	// record: V2/V3 records panic on Column for a field they do not carry.
+	_, sourceHasTTL := existingFields[ttlFieldID]
+	sourceHasTTLField := ttlFieldID >= common.StartOfUserFieldID && sourceHasTTL
+	preMaterializeFilter := len(delta) > 0 || t.plan.GetCollectionTtl() > 0 || sourceHasTTLField
```
liliu-z, line 518: The ExpirQuantiles this function ships are always "never expire" whenever the collection has a TTL field. Two independent gaps cause it: line 524 discards the ttlValues that selectFullRewriteRecord collects (`selection, _, err = ...`), and the V3 writers in internal/storage/binlog_record_writer.go never populate ttlFieldValues the way the V2 PackedBinlogRecordWriter does. Since preCompact restricts this compactor to StorageV3 segments, every full-rewrite output takes a V3 writer, so calculateExpirQuantiles falls into its neverExpire branch and datacoord's ShouldCompactExpiryWithTTLField can never fire for that segment — expired rows stay on disk indefinitely. Queries are unaffected (segcore still filters on the TTL column) and other triggers (delete ratio, size, sort, clustering, manual) can still reclaim, so this is a storage-cost problem, not a correctness one. Pre-existing and not introduced by this diff; worth a follow-up issue and a note in the PR body rather than a change here.

internal/datanode/compactor/bump_schema_version_compactor_test.go
```diff
@@ -1428,3 +1428,40 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestMaterializationRejectsDropped
...
+	s.prepareBumpSchemaVersionCompactionWithDroppedField()
+
+	result, err := s.task.Compact()
+	s.Require().NoError(err)
+	s.Require().NotNil(result)
+	s.Equal(datapb.CompactionTaskState_completed, result.GetState())
+	s.Require().Len(result.GetSegments(), 1)
+
+	segment := result.GetSegments()[0]
+	s.EqualValues(3, segment.GetNumOfRows())
```
liliu-z, line 1514: The new test only pins the negative direction. No test in the suite builds a source segment that actually carries the TTL field, so the detection added at bump_schema_version_compactor.go:509-510 is never exercised in the true direction — replacing it with a hardcoded `sourceHasTTLField = false` still passes the whole suite. That means a future regression which silently skips TTL-field filtering for segments that do carry the column (leaving expired rows in the rewritten segment) would go undetected. Worth one case where the source segment has the TTL field and expired rows are asserted dropped.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Tracked the shared StorageV3 writer gap in #52333 (section 3). In 5054a2a886 this PR only removes the unused Bump-local `ttlValues` return; it intentionally does not compute `ExpirQuantiles` in Bump. Quantile collection remains writer-owned so the follow-up can fix Bump/Mix/Sort/MergeSort/Clustering consistently. The complete BumpSchemaVersion suite passes.

internal/datanode/compactor/bump_schema_version_compactor_test.go
MrPresent-Han: Addressed in 5054a2a886. Added a real StorageV3 full-rewrite source whose manifest physically carries TTL field 105 plus dropped field 103; it writes one expired and two future TTL rows, asserts field 105 is present, and verifies the output has 2 rows. The focused cases and complete BumpSchemaVersion suite pass with the required test tags/gcflags.

PR #52302, opened 2026-08-07
enhance: scope bump partial-backfill materialization to function inputs (#51649)

MrPresent-Han:
### What

Hardens the schema-bump partial-backfill path (`runMissingFunctionMaterialization`) — the path that computes missing function outputs for pre-existing segments after `add_function_field`. Scoped down after splitting two orthogonal concerns out: the writer abort/lease moved to #52352, the `multi_analyzer_params` error reclassification moved to #52353.

- **Backfill-scoped materializer.** `NewFunctionBackfillMaterializer` replaces the `partialMaterializerExistingFields` detour: backfill prefills only the functions' absent *input* fields and computes their outputs; every other absent schema field stays untouched — an added field is nullable by DDL mandate and thus synthesizable at read time, so backfill never needs to materialize it physically. The prefill set follows the **runner-declared** inputs (`runner.GetInputFields()`), which resolve implicit inputs such as the multi-analyzer `by_field` column that `InputFieldIds` never lists.
- **Overlay before compute.** `WrapWithSelection` now fills missing fields *before* running the functions and hands the runners the overlaid view: a function input may itself be a field added by an earlier schema evolution, absent from the physical record. Previously the runners probed the raw source record — V2/V3 records panic on `Column` for an unknown field.
- **Row-count anchor.** When no explicit or implicit input exists physically, the filtered read schema would be empty; `buildFunctionBackfillReadSchema` adds one system column (RowID/Timestamp) as a row-count anchor so the backfill still sees the segment's rows.
- Renames along the same seam: `missingFunctionInputSchema` → `buildFunctionBackfillReadSchema`, `additionalFunctionInputFields` → `implicitFunctionInputFields`.

Cumulative-stats reporting is intentionally untouched here — #52006 already moved it to increment-based reporting.

issue: #51649

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -131,19 +131,19 @@ func (t *bumpSchemaVersionCompactionTask) preCompact() error {
 	return nil
 }
 
-func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunctions []*schemapb.FunctionSchema) (*schemapb.CollectionSchema, []int64, error) {
+func (t *bumpSchemaVersionCompactionTask) buildFunctionBackfillReadSchema(missingFunctions []*schemapb.FunctionSchema, sourceFields map[int64]struct{}) (*schemapb.CollectionSchema, []int64, error) {
 	schema := t.plan.GetSchema()
 	seen := make(map[int64]struct{})
-	var fields []*schemapb.FieldSchema
-	var fieldIDs []int64
-	addInputField := func(field *schemapb.FieldSchema) {
+	var readFields []*schemapb.FieldSchema
+	var readFieldIDs []int64
+	addReadField := func(field *schemapb.FieldSchema) {
 		fieldID := field.GetFieldID()
 		if _, ok := seen[fieldID]; ok {
 			return
 		}
 		seen[fieldID] = struct{}{}
-		fields = append(fields, field)
-		fieldIDs = append(fieldIDs, fieldID)
+		readFields = append(readFields, field)
+		readFieldIDs = append(readFieldIDs, fieldID)
 	}
 	for _, functionSchema := range missingFunctions {
 		if err := validateSupportedMissingFunctionMaterialization(functionSchema); err != nil {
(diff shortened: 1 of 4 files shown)


internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -978,12 +1021,14 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
 	if err != nil {
 		return nil, err
 	}
+	writerLease := &bumpSchemaVersionWriterLease{writer: writerResult.writer}
+	defer writerLease.Cleanup()
```
liliu-z: The new error-path abort releases the native writer but reclaims nothing already on remote storage. `addV3Stats` uploads each BM25 stats blob with `packed.WriteFile` (line 885), and that runs from the `updateStats` loop at line 1117 — before `buildNewInsertLogsV3` (1132), `writerLease.Close()` (1137) and `CommitManifestUpdates` (1145) — so a failure in any of those leaves the blobs, plus whatever the writer already flushed during `Write`, referenced by no manifest. Since `garbage_collector.go:722-729` skips every file under a registered live V3 segment, these orphans survive until the segment is dropped and each retried bump attempt allocates a fresh `statsID` and adds more; consider tracking the uploaded paths and deleting them in `Cleanup`, or deferring the stats upload until after the commit succeeds.

MrPresent-Han: Good point. I agree the residual orphan window is real, but it is not specific to bump-schema compaction; it is a general limitation of the file-first / manifest-second writer lifecycle. The regular writer has no attempt-scoped rollback for already finalized files, and stats/delta blobs are outside the writer as well. Whether an orphan persists depends on the caller: new/unregistered output segments are covered by DataCoord orphan GC, while in-place updates to registered live V3 segments are not.

This PR deliberately keeps a narrower guarantee: `Abort` avoids the additional error-path farewell flush and releases the native writer; it does not reclaim artifacts already written. Uploading stats after commit would expose a manifest that can reference a missing blob, while deleting files after a commit error is unsafe when the remote commit outcome is ambiguous. A complete fix should be storage-level attempt-scoped cleanup or manifest-aware live-V3 GC covering all writer/manifest users, so we will keep that as a separate follow-up and not extend the scope of this PR with a bump-specific partial deletion.

internal/storage/record_writer.go
```diff
@@ -271,6 +271,17 @@ func (pw *packedRecordBatchWriter) Close() (packed.WriterOutput, error) {
...
 }
 
+// Abort releases the underlying FFI writer without flushing pending column
+// groups — for error paths whose output will not be committed. Files already
+// flushed during Write are not undone; they stay unreferenced by the manifest.
+func (pw *packedRecordBatchWriter) Abort() {
+	if pw.writer == nil {
+		return
+	}
+	pw.writer.Destroy()
```
liliu-z: The new Abort() releases the native writer but cannot roll back an in-flight S3 multipart upload: loon_writer_destroy is a bare `delete` and every destructor in the chain (Writer, ColumnGroupWriter, FormatWriter, ParquetFileWriter) is `= default`, while CustomOutputStream declares no destructor at all — so its Abort() (the only caller of AbortMultipartUpload) never runs and uploaded parts keep accruing storage cost with no self-healing. This is pre-existing and not a regression (before this PR the error paths called nothing at all, leaking the native writer as well), and closing it needs a new milvus-storage entry point that walks the writer chain and aborts the underlying streams; worth tracking as a follow-up rather than blocking this PR.

MrPresent-Han: Confirmed—this is a real pre-existing storage-layer cleanup gap. This PR narrows the broader error path by consuming and releasing the native writer, but it does not abort multipart uploads that have already started. Fixing that safely belongs in milvus-storage, including coordination with background writes and cleanup after `Close()` failures, so we will not extend this PR's scope. Tracking it in #52333.

Issue #52303, opened 2026-08-07
[Bug]: milvus-table refresh cannot materialize BM25 input from StorageV3 TEXT LOB references


FeilongHou:
## Environment

- **Milvus Version**: commit `6c2587820da31d63b8710a5ede6dfa0be58a67e3` on branch `snapshot_external`
- **Deployment Mode**: distributed cluster, instance `eric-spark`, namespace `default`
- **MQ**: unknown / not relevant to the failure
- **SDK**: PyMilvus `3.1.0rc39`
- **OS**: Kubernetes node OS not collected

### Option B: Steps

1. Create a normal StorageV3 source collection with an `INT64` primary key, a vector field, and a `TEXT` field with analyzer enabled.
2. Insert short text such as `milvus snapshot topic0 document 0`, flush, and create a snapshot.
3. Create a `milvus-table` external collection from that snapshot. Map the source `TEXT` field to a target `TEXT` field.
4. Add a target-owned sparse-vector field and a BM25 function whose input is the mapped target `TEXT` field.
5. Call `RefreshExternalCollection`.

The current Python E2E reproduction is:

```bash
cd tests/python_client
pytest ./milvus_client/test_milvus_client_snapshot_external.py \
  -k test_milvus_table_target_owned_bm25_output
```

### Trigger Conditions

- **Frequency**: always
- **First observed after**: adding target-owned BM25 coverage for a snapshot-backed `milvus-table` external collection
- **Does NOT happen when**: generic external BM25 input is read as ordinary Arrow UTF8/`VARCHAR`; the failure is specific to StorageV3 `TEXT` represented as binary LOB references

## Expected Behavior

Refresh should decode the source StorageV3 `TEXT` values, run the target BM25 function, write the target-owned sparse-vector column, and complete successfully.

The `milvus-table` design explicitly states that DataNode runs target functions and regenerates target-owned function outputs during refresh.

## Actual Behavior

Refresh enters `RefreshFailed` before function execution. DataNode reads the source `TEXT` column as Arrow Binary, deserializes it as `storage.TextLobRef`, and then passes that value to `StringFieldData.AppendRow`, which only accepts Go `string` values.

The failing bytes are a valid StorageV3 inline TEXT encoding:

```text
[0 109 105 108 118 117 115 32 115 110 97 112 115 104 111 116 ...]
 ^
 0x00 inline-data marker, followed by "milvus snapshot topic0 document 0"
```

## Error Logs

```text
[...]

Issue #52294, opened 2026-08-07
[Enhancement]: [Snapshot as external collection] Create E2E test case


FeilongHou:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Create E2E test case

PR #52304, opened 2026-08-07
fix: fail invalid milvus-table refresh requests

FeilongHou:
issue: #52294

## Summary

- add PyMilvus E2E coverage for using Milvus snapshots as milvus-table external sources
- classify invalid snapshot metadata paths as input errors so refresh jobs fail terminally instead of remaining retryable
- verify real and virtual primary keys, deletes, refreshes, schema alignment, mixed data types, and failure atomicity
- mark the target-owned BM25 output case as a strict xfail pending #52303

## Design document

docs/design-docs/design_docs/20260526-milvus-table-external-source.md

## Test plan

- [x] python3 -m compileall -q milvus_client/test_milvus_client_snapshot_external.py
- [x] python3 -m ruff check milvus_client/test_milvus_client_snapshot_external.py
- [x] targeted BM25 E2E reports XFAIL with the #52303 reason
- [x] pytest collection discovers all 17 test variants
- [ ] targeted Go unit tests were not runnable locally because milvus_core, milvus-storage, and RocksDB pkg-config dependencies are unavailable; CI will run them

### DIFF
--- a/internal/storagev2/packed/milvus_table.go
+++ b/internal/storagev2/packed/milvus_table.go
@@ -91,10 +91,16 @@ func resolveMilvusTableSnapshotMetadataPath(externalSource, externalSpec string)
 	if _, err := externalspec.ParseExternalSpec(externalSpec); err != nil {
 		return "", err
 	}
-	if strings.HasSuffix(strings.ToLower(externalSource), ".json") {
+	sourceURL, err := url.Parse(externalSource)
+	if err == nil &&
+		sourceURL.RawQuery == "" &&
+		!sourceURL.ForceQuery &&
+		sourceURL.Fragment == "" &&
+		!strings.Contains(externalSource, "#") &&
+		strings.HasSuffix(strings.ToLower(sourceURL.Path), ".json") {
 		return externalSource, nil
 	}
-	return "", merr.WrapErrParameterInvalidMsg("milvus-table requires external_source to be a snapshot metadata JSON path")
+	return "", merr.WrapErrParameterInvalidMsg("milvus-table requires external_source to be a snapshot metadata JSON path without query or fragment components")
 }
 
 // buildMilvusTableFileInfosFromSnapshotMetadata converts snapshot metadata into
--- a/internal/storagev2/packed/milvus_table.go
+++ b/internal/storagev2/packed/milvus_table.go
@@ -94,7 +94,7 @@ func resolveMilvusTableSnapshotMetadataPath(externalSource, externalSpec string)
 	if strings.HasSuffix(strings.ToLower(externalSource), ".json") {
 		return externalSource, nil
 	}
-	return "", merr.WrapErrServiceInternalMsg("milvus-table requires external_source to be a snapshot metadata JSON path")
+	return "", merr.WrapErrParameterInvalidMsg("milvus-table requires external_source to be a snapshot metadata JSON path")
 }
 
 // buildMilvusTableFileInfosFromSnapshotMetadata converts snapshot metadata into
--- a/tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
+++ b/tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
@@ -658,7 +658,7 @@ class TestMilvusClientMilvusTableExternal(MilvusTableExternalTestBase):
         rows = [_core_row(row_id) for row_id in range(SMALL_ROWS)]
         self._create_core_source(client, source, rows)
         self._flush_rate_limited(client, source)
-        self.load_collection(client, source)
+        self._index_and_load(client, source, indexes=[("embedding", "FLAT", "L2", {})])
 
         self.delete(client, source, ids=[0])
         self.upsert(client, source, [_core_row(1, marker=9), _core_row(SMALL_ROWS, marker=9)])
--- a/tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
+++ b/tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
@@ -1,0 +1,1511 @@
+"""Milvus snapshot as a ``milvus-table`` external source E2E tests.
+
+The cases in this module intentionally use only public MilvusClient APIs. They
+validate the user-visible snapshot contract without inspecting snapshot files,
+segment IDs, or StorageV3 manifests.
+"""
+
+import json
+import math
+import time
+from dataclasses import dataclass
+from datetime import datetime
+
+import numpy as np
+import pytest
+from base.client_v2_base import TestMilvusClientV2Base
+from common import common_func as cf
+from common import common_type as ct
+from common.common_type import CaseLabel, CheckTasks
+from common.external_table_common import build_external_spec, get_minio_config
+from ml_dtypes import bfloat16
+from pymilvus import DataType, Function, FunctionType
+
+pytestmark = pytest.mark.xdist_group(name="milvus_table_external")
+
+DIM = 16
+CORE_ROWS = 3_000
+CORE_DELETE_ROWS = 100
+SMALL_ROWS = 200
+FLUSH_GAP_SECONDS = 10
+REFRESH_TIMEOUT = 300
+PIN_TTL_SECONDS = 3_600
+REFRESH_POLL_SECONDS = 2
+
+CORE_FIELD_MAPPING = {
+    "source_pk": "pk",
+    "vector": "embedding",
+    "group_alias": "group_id",
+    "score_alias": "score",
+    "tag_alias": "tag",
+    "payload_alias": "payload",
+    "numbers_alias": "numbers",
+}
+
+
+@dataclass
+class SnapshotReference:
+    """Public snapshot information needed by an external collection and cleanup."""
+
+    source_collection: str
+    snapshot_name: str
+    external_source: str = ""
+    pin_id: int | None = None
+
+
+def _value(obj, name, default=None):
+    if isinstance(obj, dict):
+        return obj.get(name, default)
+    return getattr(obj, name, default)
+
+
+def _row_value(row, name, default=None):
+    if isinstance(row, dict):
+        if name in row:
+            return row[name]
+        for container_name in ("entity", "fields"):
+            container = row.get(container_name)
+            if isinstance(container, dict) and name in container:
+                return container[name]
+        return default
+    return getattr(row, name, default)
+
+
+def _row_has_field(row, name):
+    if isinstance(row, dict):
+        if name in row:
+            return True
+        return any(
+            isinstance(row.get(container_name), dict) and name in row[container_name]
+            for container_name in ("entity", "fields")
+        )
+    return hasattr(row, name)
+
+
+def _hit_id(hit):
+    if isinstance(hit, dict) and "id" in hit:
+        return hit["id"]
+    return getattr(hit, "id", None)
+
+
+def _json_value(value):
+    return json.loads(value) if isinstance(value, str) else value
+
+
+def _timestamp_value(value):
+    assert isinstance(value, str), f"expected TIMESTAMPTZ string, got {value!r}"
+    return datetime.fromisoformat(value.replace("Z", "+00:00"))
+
+
+def _assert_search_hits(hits, limit, metric_type):
+    assert len(hits) == limit, f"expected {limit} hits, got {len(hits)}"
+    ids = [_hit_id(hit) for hit in hits]
+    assert all(hit_id is not None for hit_id in ids), f"search returned a hit without an ID: {hits!r}"
+    assert len(ids) == len(set(ids)), f"search returned duplicate IDs: {ids!r}"
+
+    distances = []
+    for hit in hits:
+        distance = _row_value(hit, "distance")
+        assert distance is not None, f"search returned a hit without distance: {hit!r}"
+        distance = float(distance)
+        assert math.isfinite(distance), f"search returned non-finite distance: {distance!r}"
+        distances.append(distance)
+
+    metric_type = metric_type.upper()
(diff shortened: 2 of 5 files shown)


tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
```diff
@@ -0,0 +1,1511 @@
...
+            snapshot_name=snapshot.snapshot_name,
+            external_source=snapshot.external_source.rsplit("/", 1)[0],
+        )
+        schema = self._build_core_target_schema(client, prefix_ref, cfg, real_pk=True)
+        self.create_collection(
+            client,
+            collection_name=invalid_target,
+            schema=schema,
+            check_task=CheckTasks.err_res,
+            check_items={ct.err_code: 1100, ct.err_msg: "snapshot metadata JSON path"},
```
yanliang567: This does not actually verify the error-code classification changed by this PR. `ResponseChecker.assert_exception()` currently ignores `ct.err_code` and checks only the message, so reverting this path to the old internal error with the same text would still pass. Please capture the SDK exception directly and assert `code == 1100` in addition to the message.

tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
```diff
@@ -0,0 +1,1511 @@
...
+        cfg = self._minio_cfg(request)
+        source = self._name("unflushed_src")
+        target = self._name("unflushed_ext")
+        rows = [_core_row(row_id) for row_id in range(SMALL_ROWS)]
+        self._create_core_source(client, source, rows)
+        self._flush_rate_limited(client, source)
+
+        self.delete(client, source, ids=[0])
+        self.upsert(client, source, [_core_row(1, marker=9), _core_row(SMALL_ROWS, marker=9)])
+        snapshot = self._create_snapshot_ref(client, source, cfg, suffix="unflushed")
```
yanliang567, line 722: Could we establish that these unflushed mutations are visible in the source before taking the snapshot? The current target assertions also pass if the delete/upserts were acknowledged but became no-ops: the target would still contain the original 200 rows. Please query the source with Strong consistency here and verify PK 0 is deleted, PK 1 has the `marker=9` replacement, and PK 200 exists; then the later assertions genuinely prove that the snapshot excluded those unflushed changes.

tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
```diff
@@ -0,0 +1,1511 @@
...
+            assert float(_row_value(hit, "distance")) == pytest.approx(0.0, abs=tolerance)
+
+        self._assert_schema_alignment(
+            source_info,
+            self.describe_collection(client, target)[0],
+            mixed_mapping,
+        )
+
+    @pytest.mark.tags(CaseLabel.L2)
+    @pytest.mark.xfail(
```
yanliang567: A function-level xfail treats every failure in this case as #52303, including unrelated schema/API regressions before refresh and index/search/assertion regressions after it. Please narrow this to the known refresh failure: inspect the terminal refresh state/reason and call `pytest.xfail()` only when it matches #52303. If refresh completes, continue through all assertions; any other failure should remain a real failure.

tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
```diff
@@ -0,0 +1,1511 @@
...
+        expected = _mixed_row(row_id)
+        assert row["bool_alias"] == expected["bool_field"]
+        assert row["int8_alias"] == expected["int8_field"]
+        assert row["int32_alias"] == expected["int32_field"]
+        assert float(row["float_alias"]) == pytest.approx(expected["float_field"], abs=1e-5)
+        assert float(row["double_alias"]) == pytest.approx(expected["double_field"], abs=1e-9)
+        assert row["varchar_alias"] == expected["varchar_field"]
+        assert _timestamp_value(row["timestamptz_alias"]) == _timestamp_value(expected["timestamptz_field"])
+        assert _json_value(row["json_alias"]) == expected["json_field"]
+        assert list(row["array_alias"]) == expected["array_int64"]
```
yanliang567, line 1397: The test requests `geometry_alias` and claims Geometry round-trip coverage, but never checks the retrieved value. The later `ST_EQUALS` filter proves predicate evaluation, not retrieval/serialization. Please add an exact or canonicalized Geometry equality assertion here.

yanliang567 (changes requested):
I found two verification gaps that can let the new E2E coverage pass without proving the behavior under test, plus two narrower coverage issues. Please see the inline comments.

yanliang567 (commented):
## Test review

Thanks for this — the E2E module is well-structured and covers the main data path nicely. Below are review notes on coverage, assertion depth, and atomicity. None are correctness blockers on the production change (the `merr` classification is verified end-to-end), but several are coverage gaps worth closing before merge.

### Go side

1. **Unit test mocks at `exploreExternalFiles`, so it's a single-boundary test.** `input_error_marks_job_failed_non_retriable` verifies the classification branch in `createTasksForJob`, but not that a *real* invalid-path refresh surfaces as `InputError` at that check. I traced the chain — `resolveMilvusTableSnapshotMetadataPath` returns the error unwrapped at both call sites (`internal/storagev2/packed/milvus_table.go:231`, `internal/storagev2/packed/explore_ffi.go:281`), and the E2E `test_milvus_table_rejects_non_json_source` covers the terminal path — so coverage is acceptable, but the unit test alone wouldn't catch a future `merr.Wrap` between the origin and the classifier stripping the `InputError` type. Consider one test that drives the real `resolveMilvusTableSnapshotMetadataPath` error instead of mocking `exploreExternalFiles`.

2. **`merr.GetErrorType(err) == merr.InputError` is a broad catch-all.** Per the error-handling blame test, any transient error misclassified as `InputError` upstream of this check would now become terminal (non-retriable) instead of retrying. Only the snapshot-JSON-path input error is exercised. Worth auditing all `InputError` sources reachable through `exploreExternalFiles` (spec parse/validate, manifest reads, storage failures) to confirm none are transient-in-disguise.

### Coverage gaps (Python E2E)

3. **In-place upsert value update across refresh is not tested.** Append (`..._refresh_adds_rows_and_is_idempotent`) covers new rows and delete (`..._refresh_applies_snapshot_deletes`) covers removals, but there's no case: snapshot A (old value) → upsert overwrite a PK → flush → snapshot B → refresh → value updated and old value gone. `..._delete_before_reinsert_visibility` is single-snapshot timestamp ordering, not cross-refresh mutation. This is the third "L1 changed" form and is currently unexercised.

4. **The "L1 unchanged + L0 overlay changed → keep segment ID, rewrite manifest" refresh path is never asserted.** This is the core optimization in the design doc's Refresh Flow table. The module intentionally uses only public APIs (no segment IDs), so this is a deliberate gap — but it means the segment-reuse/zero-copy path has no E2E coverage.

5. **Delete sources are not distinguished.** Design's Delete Handling separates "source segment deltalogs" from "snapshot L0 overlays". The tests flush deletes into L0, so only the L0-overlay path is exercised.

6. **No "refresh fails while the collection is loaded" scenario.** Every failure test calls `release_collection` first. Atomicity of a failed refresh against a live, loaded collection is unverified.

### Assertion depth

7. **`_assert_core_row` never asserts vector values.** Vectors are only validated indirectly via search hit IDs. A corrupted/permuted vector column that still returns correct nearest neighbors would pass. Add an exact vector comparison (or at least self-distance ≈ 0, which is already done for search but not for the query/get rows).

8. **Failure-path assertions are shallow.** After a failed refresh, tests only check `external_source` unchanged + `_count` unchanged + one row via query (e.g. pk 42). There is no `search` and no `_core_signature`-level comparison, so "collection remains fully readable from A" is only weakly demonstrated.

9. **`_core_signature` omits `numbers_alias` and the vector fields**, so the idempotency checks don't cover every mapped field.

10. **Server-side error-string assertions are fragile.** `"external_field"` and `"definition mismatch"` (and `"snapshot metadata JSON path"`, `"cannot use an external collection snapshot as source"`) are exact substring matches; any message change breaks these tests spuriously.

### Atomicity

11. **Create atomicity is sufficient** (`_assert_collection_absent` + `has_collection == False` across all three failure cases).

12. **Refresh atomicity is not sufficient.** Current checks prove "old source + old count + one readable row", but not: (a) search still works after failure, (b) full data-signature consistency, (c) absence of orphan segments/dirty manifests (count catches gross row loss but not same-count wrong-data), (d) loaded-state behavior (see #6).

### Snapshot before/after data & schema updates

13. **Data-update scenarios are well covered** for before-snapshot, after-snapshot-flushed (append/delete), and after-snapshot-unflushed (`..._excludes_unflushed_dml`) — except for the upsert-overwrite gap in #3.

14. **Schema-update coverage is thin.** The design doc's "Compatibility and Failure Behavior" says "later refresh pointing to a different schema fails schema identity validation", but E2E only exercises a **dimension change**. `ValidateMilvusTableSchemaIdentity` (`pkg/util/typeutil/schema.go:2914`) has many more failure branches that are unexercised at the E2E level:

    | Branch | Location | E2E |
    |---|---|---|
    | dynamic field setting mismatch | `:2921` | ❌ |
    | user field count mismatch | `:2931` | ❌ |
    | target field missing `external_field` | `:2938` | ✅ (create, `missing_external_field`) |
    | maps to nonexistent source field | `:2941` | ❌ |
    | source field double-mapped | `:2946` | ❌ |
    | field_id mismatch | `:3177` | ❌ (hard via public API — RootCoord aligns IDs) |
    | definition mismatch (type/dim/params) | `:3183` | ✅ but only dim; no type change / add / remove field |

    Suggested: a parametrized refresh-failure test covering **source field added, source field removed, field type change (INT32→INT64), and dynamic-field toggle**, each asserting `RefreshFailed` + old `external_source`/data preserved.

### Minor

15. `_flush_rate_limited`'s fixed 10s sleep is a heuristic workaround for the flush→snapshot race; it makes the suite slow and the fixed wait itself is fragile.
16. Virtual-PK idempotency (`_hit_id` = virtual PK inside `_core_signature`) assumes virtual PKs are stable across refresh; if refresh ever reassigns them this will flake.
17. `__virtual_pk__` field name and `auto_id is True` assertions are implementation-coupled.
18. `_build_core_target_schema`'s `omit_external_field == "source_pk"` branch is unused.
19. `..._excludes_unflushed_dml` depends on "create_snapshot does not flush" semantics — fine if that's the documented contract, but worth confirming it isn't an implementation accident.
20. `missing_external_field` case assumes the server rejects a bare unmapped scalar (target-owned is only allowed for function outputs). Worth confirming against `validateMilvusTableRefreshSchema`/create-time validation, since `test_milvus_table_target_owned_bm25_output` shows unmapped target-owned fields *are* allowed for function outputs.

yanliang567 (commented):
The production error-classification change is directionally correct, and the latest revision resolves several earlier verification gaps. I found three remaining issues that can still leave the intended behavior or test guarantees incomplete; details are inline.

internal/storagev2/packed/milvus_table.go
```diff
@@ -94,7 +94,7 @@ func resolveMilvusTableSnapshotMetadataPath(externalSource, externalSpec string)
 	if strings.HasSuffix(strings.ToLower(externalSource), ".json") {
```
yanliang567: This checks the suffix of the raw URI rather than the parsed object path. For example, `minio://host/bucket/not-json?marker=.json` or `minio://host/bucket/not-json#.json` passes here, while `externalFilesystemFilePath` later reads only `u.Path`, so the actual object key is `not-json`. If that object exists but is not snapshot metadata, parsing returns a service-internal error and refresh can remain retryable—the behavior this PR is intended to eliminate. Could we parse the URI, validate the extension on its path, explicitly reject or normalize unsupported query/fragment components, and add coverage for these cases? This would also handle encoded extensions consistently.

tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
```diff
@@ -0,0 +1,1624 @@
...
+            if any(
+                term in lowered
+                for term in (
+                    "52303",
+                    "wrong row type",
+                    "append field text_alias",
+                    "text lob",
+                    "materialize",
+                )
+            ):
```
yanliang567: This is much narrower than the former function-level xfail, but the `any(...)` match can still hide unrelated refresh regressions. In particular, `wrong row type` and `materialize` are generic enough to match failures in another field or materialization path. The known #52303 failure has a compound signature: `append field text_alias` together with `Wrong row type` (and `expected=string`). Please require that conjunction and remove the standalone generic fallbacks, so only the known TEXT-LOB failure is classified as xfail.

tests/python_client/milvus_client/test_milvus_client_snapshot_external.py
```diff
@@ -0,0 +1,1624 @@
...
+        self.load_collection(client, target)
+        assert self._count(client, target) == SMALL_ROWS
+        rows = self._query_source_pks(client, target, [42])
+        assert len(rows) == 1
+        self._assert_core_row(rows[0], _core_row(42))
+        # The collection must remain fully readable (not just queryable) from
+        # the old snapshot after a failed refresh: search still resolves the
+        # same nearest neighbor.
+        hits = self._search_core(client, target, _core_vector(42), limit=1)
+        assert _row_value(hits[0], "source_pk") == 42
```
yanliang567: Adding search here is useful, but this still has no pre-failure data baseline. A failed refresh could partially rewrite or drop rows/fields outside PK 42 while preserving the count and this nearest-neighbor result, and the test would pass. Please capture a representative full-data signature before release/refresh and compare it after the failure and reload. The helper should also assert that every requested sample PK is present and matches its expected row, rather than only comparing two possibly identical incorrect results. The invalid-path and chaining failure cases have the same gap.

yanliang567 (commented):
/lgtm
/approve

Issue #52316, opened 2026-08-07
[Enhancement]: avoid allocations in CRC32 string hash helpers


bigsheeper:
## What would you like to be improved?

`pkg/util/typeutil.HashString2Uint32` converts the string prefix it hashes with `[]byte(subString)` before calling `crc32.ChecksumIEEE`. This helper is used in varchar routing paths, so the conversion creates one heap allocation per hashed key.

Call-site evidence:

- `HashPK2Channels` and `HashKey2Partitions` route varchar primary keys and partition keys through `stringRoutingHasher.Hash`, which calls `HashString2Uint32` inside `locateRoutingIndexes`' per-key loop.
- import hashing calls `HashString2Uint32` while iterating primary keys and partition keys in `internal/datanode/importv2/hash.go`.
- `HashNamespace2Channels` uses the same helper for namespace-to-channel routing.

## Why is this useful?

`crc32.ChecksumIEEE` only reads the byte slice during the call, so it can hash a temporary byte view of the already-immutable string instead of allocating a copied `[]byte`. The observable hash output stays unchanged.

Existing behavior-pinning tests already cover exact CRC32 outputs and routing modulo behavior:

```bash
cd pkg
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 -run "TestHashString2Uint32|TestHashPK2Channels|TestHashPK2ChannelsMatchesLegacyModulo|TestLocateRoutingIndexesMatchesLegacyModulo|TestHashKey2PartitionsMatchesLegacyModulo" ./util/typeutil
```

Result:

```text
ok  github.com/milvus-io/milvus/pkg/v3/util/typeutil  0.015s
```

## Proof

Benchmark command, with before and after measured back to back in the same worktree/session:

```bash
cd pkg
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -run "^$" -bench "BenchmarkHashString2Uint32$" -benchmem -count=5 ./util/typeutil
```

Median of 5, `-gcflags="all=-N -l" -benchmem -count=5`.

time:

| | before | after |
|---|---:|---:|
| `HashString2Uint32/short` | 70.45 ns | **33.99 ns (2.07x)** |
| `HashString2Uint32/long` | 116.2 ns | **47.12 ns (2.47x)** |

B/op:

| | before | after |
|---|---:|---:|
| `HashString2Uint32/short` | 8 | **0** |
| `HashString2Uint32/long` | 112 | **0** |

[...]

PR #52318, opened 2026-08-07
enhance: avoid crc32 string hash allocation

bigsheeper:
issue: #52316

## What this PR does

This PR removes the per-call string-to-`[]byte` allocation in `pkg/util/typeutil.HashString2Uint32` by passing `crc32.ChecksumIEEE` a read-only byte view of the string prefix.

The exported signature and CRC32 output stay unchanged.

## Why

`HashString2Uint32` is used by varchar routing paths:

- `HashPK2Channels` and `HashKey2Partitions` route varchar keys through `stringRoutingHasher.Hash`, which calls `HashString2Uint32` inside `locateRoutingIndexes`' per-key loop.
- import hashing calls `HashString2Uint32` while iterating primary keys and partition keys in `internal/datanode/importv2/hash.go`.
- `HashNamespace2Channels` uses the same helper for namespace-to-channel routing.

## Numbers

`BenchmarkHashString2Uint32` was added in this PR. Median of 5, `-gcflags="all=-N -l" -benchmem -count=5`; before and after measured back to back in the same worktree/session.

time:

| | before | after |
|---|---:|---:|
| `HashString2Uint32/short` | 70.45 ns | **33.99 ns (2.07x)** |
| `HashString2Uint32/long` | 116.2 ns | **47.12 ns (2.47x)** |

B/op:

| | before | after |
|---|---:|---:|
| `HashString2Uint32/short` | 8 | **0** |
| `HashString2Uint32/long` | 112 | **0** |

`allocs/op` dropped from 1 to 0 for both benchmark cases.

## Verification

```bash
cd pkg
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 -run "TestHashString2Uint32|TestHashPK2Channels|TestHashPK2ChannelsMatchesLegacyModulo|TestLocateRoutingIndexesMatchesLegacyModulo|TestHashKey2PartitionsMatchesLegacyModulo" ./util/typeutil
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/typeutil
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l -d=checkptr" -count=1 ./util/typeutil
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -run "^$" -bench "BenchmarkHashString2Uint32$" -benchmem -count=5 ./util/typeutil
```

```bash
cd internal
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/typeutil
```

```bash
GOWORK=off make static-check
```

`make static-check` result: 0 issues in core, pkg, client, and go_client e2e checks.

Local static-check prerequisites used in this worktree: `internal/core/output` linked to the main checkout's built C++ output, and ignored generated `cmd/tools/migration/legacy/legacypb` linked from the main checkout. These helper links are untracked and are not part of the PR.

## Deferred

`HashString2LessUint32` has the same allocation pattern, but this PR leaves it unchanged because the existing tests in this checkout do not directly pin that helper's output.


czs007:

Verified:
- `pkg/util/typeutil/hash.go:94` — `unsafe.Slice(unsafe.StringData(subString), len(subString))` is the canonical zero-copy string→byte-view idiom; `crc32.ChecksumIEEE` only reads the slice and never retains it, so the immutable string backing array is never mutated and the view does not escape.
- `pkg/util/typeutil/hash.go:89-94` — empty input is safe: `unsafe.Slice(nil, 0)` panics only when `ptr == nil && len != 0`, so `v == ""` yields a nil zero-length slice and matches `crc32.ChecksumIEEE([]byte(""))`; exactly-100-byte strings stay untruncated and multibyte strings keep the existing byte-based truncation semantics.
- `pkg/util/typeutil/hash.go:88-94` — the bytes fed to CRC32 are identical to the old `[]byte(subString)` path, so `HashString2Uint32` output is byte-identical and the 100-byte prefix rule is preserved.
- `pkg/util/typeutil/hash.go:50-51,138,157,193-194`, `internal/datanode/importv2/hash.go:259-284`, `internal/util/typeutil/hash.go:20-24` — the exported signature and hash value are unchanged, so channel, partition, import, namespace, and query-routing call paths preserve their behavior.
- `pkg/util/typeutil/hash_test.go:136-142` — exact CRC values remain pinned for both short and over-limit inputs, covering the two branches affected by the conversion; the legacy-modulo pins additionally guard the routing contract.
- `pkg/util/typeutil/hash_test.go:376-390` — the benchmark exercises both short and truncated long inputs, reports allocations, and stores results in a package-level sink to prevent dead-code elimination.

czs007 (approved):
/lgtm

Issue #50954, opened 2026-06-30
[Feature] Support idempotency_key for BulkImport API to enable safe retries


sagar-arora:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Context
We are orchestrating Milvus BulkImport pipelines. Our workflow stages Parquet files in S3 and then calls the Milvus BulkImport REST/gRPC API to ingest them into a collection.

Because data pipelines operate in a distributed environment, the worker pods calling the Milvus API can crash, be preempted, or experience network partitions.

### The Problem
Currently, the BulkImport API is non-idempotent. When a client submits an import request, Milvus generates and returns a new jobId.

If a client submits the import request, Milvus accepts it, but the client crashes before it can persist the returned jobId to its own state, the orchestrator will retry the task. The retry will submit the exact same S3 files to the API again. Because Milvus has no way to know this is a retry, it creates a second import job, resulting in duplicated data and bloated segments in the target collection.

### Proposed Solution
We propose adding an optional `idempotency_key` to the BulkImport API. This allows external orchestrators (like Airflow, Temporal, or Temporal) to safely retry API calls without risking data duplication.

### Expected Behavior:

The client passes an `idempotency_key` (e.g., run_id_batch_1) in the ImportRequest.
If Milvus (DataCoord) receives a request with an `idempotency_key` it hasn't seen before, it creates the job, maps the key to the new jobId, and returns the jobId.
If Milvus receives a request with an `idempotency_key` that already exists, it ignores the payload and simply returns the previously generated jobId.

### High-Level Implementation Thoughts
I'd like to contribute this feature and submit a PR, but I wanted to validate the architectural approach with the maintainers first:

`milvus-proto`: Add string `idempotency_key = 5;` to the `ImportRequest` message.
SDK/Proxy: Update the Go SDK BulkImportOption and proxy logic to pass this key down to the DataCoord.
DataCoord (Meta Storage):
Add `IdempotencyKey` to the internal ImportJob struct.
When DataCoord saves a new import job to etcd, atomically write a secondary index mapping: `import-idempotency/<collection_id>/<idempotency_key>` -> `jobId`.
DataCoord (Import Manager): In SubmitImportJob, check `etcd` for the `idempotency_key` before generating a new jobId.
Does this approach align with the team's vision for DataCoord and the BulkImport API? If so, I am happy to open a PR for `milvus-proto` and milvus.

sagar-arora:
Hi @cydrain could you help triage this?

[...]

PR #50007, opened 2026-05-22
enhance: Support idempotent write

tinswzy:
issue: #50000

design doc:
docs/design-docs/design_docs/20260604-idempotent_write.md

## Problem

A client whose Insert times out or loses its ack cannot tell whether the write
landed. Retrying may duplicate the rows; not retrying may lose them. There is no
way for the server to recognize that the second request is the same write as the
first.

## Solution

Make the WAL append itself idempotent, and implement Insert on top of it.

```
client ──(gRPC metadata idempotency-key / REST Idempotency-Key)──> proxy
                                                                    │ derives a key when
                                                                    ▼ the caller gave none
                                                    message property `_ik`
                                                                    │
                     streamingnode: per-vchannel dedup window, consulted BEFORE append
                                                                    │
                                          WAL summary: the window's durable half
```

**Dedup happens before the append.** The interceptor consults a per-vchannel
window: a first sighting owns the key and appends; a concurrent retry waits on
the owner; a completed key is answered with the first attempt's message id,
timetick and last-confirmed position, and nothing reaches the WAL. A multi-message
insert (a partition-key collection, or a payload split by `pulsar.maxMessageSize`)
is grouped into a transaction and the key is carried on the synthesized
`CommitTxn`, so a partial write unit can never land.

**The key is either supplied or derived.** With `collection.insert.idempotency.enabled`
and no client key, the key is a hash of the destination and the payload — names
only, no ids and no collection generation, hashed as the client sent it, so a
retry reproduces it before any server-side resolution. Under `autoID` the
candidate ids are pre-allocated so a retry reproduces the first attempt's primary
keys and its row-to-shard routing.

**The window survives a restart.** A new pchannel-scoped WAL summary store
persists the window: each idempotent write contributes a record, records are
sealed into chunk objects, and an object-storage manifest indexes them. On WAL
open, recovery reads the chunks back and rebuilds each vchannel's window.

**The chunk write is ordered against the WAL checkpoint.** It happens inside the
recovery storage's dirty-snapshot persist, before the consume checkpoint covering
those records is saved, and a failure fails that persist. The checkpoint is
therefore itself the boundary between what the store holds and what the WAL still
holds: no second position, no clamp on the checkpoint, and no rewind at recovery.
The cost is deliberate — object-storage latency sits on the checkpoint path, and
a stalled store stalls the checkpoint rather than silently dropping records.

**A superseded owner cannot advance the consume checkpoint.** `WALCheckpoint` now
[...]

PR #52320, opened 2026-08-08
feat: add idempotency_key to BulkImport for safe retries

sagar-arora:
Make the BulkImport (ImportV2) API idempotent via an optional client-supplied idempotency_key. Before broadcasting, DataCoord reserves a per-collection key -> jobID mapping in etcd with an atomic put-if-absent, so a retried or concurrent submission with the same key returns the original jobID instead of creating a duplicate import job. The key is threaded through the request messages, the WAL ImportMsg, and the persisted ImportJob; the reservation is released if the broadcast fails and dropped when the job is GC'd.

GetImportProgress can also resolve a job by (collectionName, idempotency_key), so a client that lost the ImportV2 response can recover the jobID without resending the payload. The proto field itself was added in milvus-proto#633; this change adds it to the in-repo internalpb/datapb messages and wires it end to end across proxy, DataCoord, and the REST v2 handlers.

issue: #50954
design doc: docs/design-docs/design_docs/20260709-bulkimport-idempotency-key.md


sagar-arora:
@xiaofan-luan, @tinswzy please review this when you have a chance.

docs/design-docs/design_docs/20260709-bulkimport-idempotency-key.md
```diff
@@ -0,0 +1,306 @@
+# MEP: Idempotency Key for BulkImport (ImportV2)
```
lyang24, line 1: i would break this up into two prs 
1. mep
2. the impl

bigsheeper:
> @xiaofan-luan, @tinswzy please review this when you have a chance.

Thanks for your contribution. On it.

czs007:
> **Must-fix issues introduced by this PR: 4**
> **Merge recommendation:** Mergeable after the 4 must-fix items below are fixed in this PR.

The idempotency-key feature is threaded end-to-end cleanly and the reserve-before-broadcast shape fits the two-phase job creation, but the reservation lifecycle has several holes — one hard break on TiKV metastore backends, and a set of windows where a client is told `Success` for an import that never runs or runs twice.

## High

**`internal/metastore/kv/datacoord/kv_catalog.go:783` — `idempotency_key` is unusable on the TiKV metastore backend.**
`SaveImportIdempotencyKeyIfAbsent` calls `MetaKv.CompareVersionAndSwap`, but `txnTiKV`'s implementation (`internal/kv/tikv/txn_tikv.go:827-831`) returns `merr.ErrServiceUnimplemented` unconditionally. When `common.metaStoreType=tikv`, DataCoord's `s.kv` is a `txnTiKV` (`server.go:611-614`), the catalog wraps it (`server.go:632`), and `importMeta` uses that catalog (`server.go:318`). Failure: on a TiKV-backed cluster, any `BulkImport` carrying an `idempotencyKey` → `ImportV2` returns `ErrServiceUnimplemented` wrapped as "failed to reserve import idempotency key" and the import never executes, while imports without a key on the same cluster still succeed. This line is the only metastore-layer call site of `CompareVersionAndSwap`, so this path was never exercised before. Suggestion: implement put-if-absent for TiKV, or reject the field explicitly at the API edge based on metastore type with a clear message, and state the limitation in the MEP's Compatibility section (it currently says only "Backward compatible"). *(raised by czs007)*

**`internal/datacoord/services.go:1945` — a concurrent duplicate can receive `Success` plus a jobID whose job is never created.**
There is a window between reserving the key and creating the job: request A reserves `K -> 500` and enters `broadcastImport`; request B with the same key hits the in-memory index (`import_meta.go:186-188`) and returns `Status=Success, JobID=500` via the `!isNew` branch; A's broadcast then fails on an ordinary retryable error and A releases `K` at `services.go:1974-1979`. Failure: two concurrent `ImportV2` calls share a key and A's broadcast fails on `ValidateMaxImportJobExceed` or a brief streamingnode outage → B is told `Success` with jobID 500, but that job is never created and `GetImportProgress(500)` returns "import job does not exist" forever, so B's files are never imported despite a success response. No crash is required — this reproduces inside a single DataCoord process, which makes it strictly more likely than the crash case the MEP discusses. Suggestion: give the reservation record a state (`reserved` / `committed`, flipped in `createImportJobFromAck`) and have the `!isNew` path return a retriable error (or await commit confirmation) while the mapping is still `reserved`. Related: if a release lands between another request's failed CAS and its read-back, the `Load` at `kv_catalog.go:789-792` surfaces `merr.WrapErrIoKeyNotFound` (`etcd_kv.go:232-234`) as a hard error instead of retrying the CAS. *(raised by czs007)*

**`internal/datacoord/services.go:1974` — releasing the reservation on every `broadcastImport` error can produce a duplicate import.**
`broadcasterScheduler.AddTask` (`broadcast_scheduler.go:44-62`) enqueues onto `pendingChan` *before* blocking on the merged request context, and once enqueued the background worker retries with backoff until the append succeeds (`pending_broadcast_task.go:49-83`). Failure: the request context times out after the task is already enqueued (slow or briefly unavailable streamingnode, client RPC deadline) → DataCoord deletes the mapping and returns an error, the background task later succeeds and `importV1AckCallback` → `createImportJobFromAck` creates the original job anyway, and the client's retry under the same key reserves a fresh jobID and creates a second import — the same files land twice. Suggestion: release only when the broadcast provably created no task (pre-enqueue failures such as `ErrBroadcastTaskNotCreated`), or tie reservation cleanup to the broadcast task's terminal lifecycle. *(raised by chyezh)*

**`internal/datacoord/services.go:2123` — lookup by key returns not-found instead of the reserved jobID during the pre-ack window.**
The key branch resolves the reserved jobID at `services.go:2128`, then requires a persisted job at `services.go:2144-2147`; during the pre-ack window the reservation exists but `AddJob` has not run. Failure: a client loses the response to a keyed `ImportV2` and immediately queries by `(collectionName, idempotencyKey)` while the broadcast is out but the ack has not created the job → the response is `ErrImportSysFailed` "import job does not exist" with `resp.JobID` never set (the assignment at line 2153 is past the early return), so per the comment at 2124-2127 and the MEP the client concludes the submission never landed and re-runs under a new key, importing the same files twice. The unknown-key case (`!ok`, 2129-2133) also returns the same error class as "reserved but not yet persisted", so callers cannot distinguish "safe to re-submit" from "still in flight", and the MEP's test plan (`20260709-bulkimport-idempotency-key.md:261-263`) claims this path returns a pending state with the reserved jobID. Suggestion: return a pending response carrying the reserved jobID, or persist a pending job before broadcasting. *(raised by czs007, chyezh)*

## Medium

**`internal/datacoord/import_meta.go:208` — `ResolveIdempotencyKey` reads only the in-memory index, so uncommitted reservations are lost across a restart.**
Unlike `CheckAndReserveIdempotencyKey` (`import_meta.go:183-194`), it never falls back to the catalog, and the index is rebuilt solely from `restoredJobs` / `catalog.ListImportJobs` (`import_meta.go:161-170`). Failure: DataCoord restarts (rolling upgrade, OOM) after the etcd reservation is written but before the ack persists the job → lookup by key returns "import job does not exist" even though `import-idempotency/{collID}/K` is still in etcd; a re-submission then takes the CAS-conflict path and gets back the old jobID, so the client can never obtain a usable import, and switching to a new key duplicates the data. Suggestion: fall through to a catalog `Load` on an index miss and backfill the cache, or restore reservations independently of persisted jobs. *(raised by czs007, chyezh)*

**`internal/datacoord/import_meta.go:287` — `RemoveJob` drops the job before the mapping, non-atomically, which can poison a key permanently.**
`DropImportJob` runs first, then `DropImportIdempotencyKey`, and a mapping-drop error returns early so `delete(m.jobs, jobID)` is never reached. Failure: GC (`import_checker.go:694`) drops the job successfully but the mapping drop hits a transient etcd error (timeout, leader change) and DataCoord then restarts → the job is gone from `ListImportJobs` so the rebuilt index lacks the key, but `datacoord-meta/import-idempotency/{collID}/K` remains; re-submitting with that key takes the CAS-conflict path and returns the dead jobID, progress queries return "import job does not exist" forever, and since `RemoveJob` is the only deletion path and its job no longer exists, the mapping can never be reclaimed. Suggestion: drop the mapping before the job (a dangling job merely loses dedup; a dangling mapping is unrecoverable), or log rather than propagate the mapping-drop error; a startup reconciliation sweep would be the complete fix, but `DataCoordCatalog` currently has no `ListImportIdempotencyKeys` (`datacoord_catalog.go:110-116`) to enumerate them. *(raised by czs007, chyezh)*

**`internal/datacoord/services.go:1945` — a reused key with a different payload silently discards the second import.**
`services.go:1939-1951` returns the reserved jobID for any request bearing a known key without comparing `Files`, `PartitionIDs`, or `Options` against the original submission, and the reservation value stores only the jobID (`kv_catalog.go:781-786`). Failure: an orchestrator shards one logical import into N parallel `ImportV2` calls and derives the key from the run id alone → shards 1..N-1 all receive `Success` with shard 0's jobID, poll it to `Completed`, and report success while their files were never imported, leaving only a single "duplicate import request detected by idempotency key" INFO line as evidence. Mainstream idempotency-key APIs return a conflict error for key reuse with a differing body precisely because this failure looks like success. Suggestion: persist a fingerprint (hash of sorted file paths + partitionIDs + options) alongside the jobID and return a distinct error on mismatch; if that is deliberately out of scope, make it an explicit non-goal in the MEP with the data-loss consequence spelled out. *(raised by czs007)*

**`internal/datacoord/import_meta.go:293` — the dedup guarantee's lifetime is implicitly bound to `dataCoord.import.taskRetention`.**
The mapping is deleted in `RemoveJob`, which GC invokes once a Completed/Failed job passes `dataCoord.import.taskRetention` (default `10800`s = 3h; `import_job.go:71-79`, `import_checker.go:613-616`, `component_param.go:6661-6668`) — note the clock starts at the job's terminal state, not at submission. Failure: an orchestrator retries a stuck or manually restarted task more than 3h after the original job completed (paused-then-resumed DAG, next-day manual rerun) under the same key → the mapping is gone, `CheckAndReserve` treats it as brand new, and the same files are imported a second time with no error surfaced and no way for the client to tell "newly created" from "deduplicated" in the response. Suggestion: document this retention window in the MEP's Public Interfaces section, and consider decoupling the mapping's lifetime from job retention. *(raised by czs007)*

## Low

**`internal/distributed/proxy/httpserver/handler_v2.go:3764` — REST does not reject a lookup carrying both `jobId` and `idempotencyKey`.**
The comment states "Require exactly one lookup form", but the validation only covers "neither supplied" and "key without collectionName" (`handler_v2.go:3763-3769`). DataCoord's branch condition is `in.GetJobID() == "" && in.GetIdempotencyKey() != ""` (`services.go:2123`), so when both are present it silently falls through to jobID parsing. Failure: a client calls `/v2/vectordb/jobs/import/describe` with both a stale `jobId` from a prior attempt and an `idempotencyKey` → the key is ignored and the response describes the unrelated job's state/progress/collectionName, which the caller reads as the import it located by key. Suggestion: reject the both-supplied case explicitly, and apply the same mutual-exclusion check on the internal RPC path (`internal/proxy/impl.go:6397`). *(raised by chyezh)*

**`docs/design-docs/design_docs/20260709-bulkimport-idempotency-key.md:251` — the MEP's test plan claims coverage the PR does not add.**
The plan (md:251-266) promises tests for: mapping released on broadcast failure; concurrent same-key submissions converging to one jobID; progress lookup by `collectionName + idempotency_key` including the pre-ack window and the unknown-key case; and REST rejection of incomplete lookup requests. What was added: `import_services_test.go:258-315` (duplicate-returns-existing, invalid-key-rejected), `import_meta_test.go:80-176` (reserve/resolve/release/restore/RemoveJob), `kv_catalog_test.go:149-185` (catalog CAS), `option_test.go:33-49` (key validation). A repo-wide grep for `IdempotencyKey|idempotencyKey` across `*_test.go` matches only those four files — the broadcast-failure release path (`services.go:1974-1979`), the DataCoord key lookup (`services.go:2122-2135`), the proxy path (`impl.go:6395-6410`), and the new REST validation plus `ImportProgressReq` decoding (`handler_v2.go:3758-3769`) are untested. Suggestion: either add the tests or correct the claim. *(raised by czs007)*

## Nitpick

**`internal/datacoord/import_meta.go:208` — `ResolveIdempotencyKey` accepts a `ctx` it never uses.** The body is just an `RLock` plus a map lookup. Adding the catalog fallback suggested above resolves this naturally. *(raised by czs007)*

**`internal/datacoord/import_meta.go:111` — the field comment says the index is "rebuilt from the catalog on startup", which reads as a scan of the `import-idempotency/` prefix.** It is actually rebuilt by iterating the jobs returned by `catalog.ListImportJobs` (`import_meta.go:161-170`) — precisely the distinction behind the lost-reservation issue above. Suggestion: reword to "rebuilt from the persisted ImportJobs". *(raised by czs007)*

bigsheeper:
Hi @sagar-arora — before anything else, thank you. This is a genuinely well-executed contribution, and I want to be specific about that rather than generic: you filed a sharp problem statement in #50954 (the "worker crashes before persisting the returned jobId" framing is exactly right and is what the whole design ended up organized around), you wrote a real design doc under `docs/design-docs`, and you wired the key end to end across proxy, DataCoord, the WAL message and the REST v2 handlers — including the parts people usually skip: the atomic put-if-absent reservation, the release path on broadcast failure, GC of the mapping alongside the job, and test coverage for all of it. You also documented the edge cases honestly instead of hiding them. That is a higher bar than most first contributions to this repo clear.

I owe you an explanation for what comes next, and I want to be clear up front that **the change of direction is on our side, not on your implementation**.

While this PR was in progress, #50007 ("Support idempotent write") has been converging on a **cluster-wide idempotency framework** rather than a per-API one. The decision we've now settled on is that every operation that wants idempotent retries — Insert today, BulkImport next, and DDL-style operations after that — should express it the same way, and that we should not grow a second, API-specific idempotency store alongside it. Concretely, the unified shape is:

1. The client supplies the key **out of band**: an `Idempotency-Key` HTTP header, or gRPC metadata — no public proto change.
2. The proxy picks it off the incoming context and forwards it on the internal hop via an interceptor.
3. Dedup happens at the **broadcaster** layer, reusing the broadcast task records that already exist and are already persisted and retained (as tombstones) after completion.
4. Before broadcasting, the path calls a shared "has this key already been broadcast for this resource?" check. On a hit it returns the original message's `jobID` and skips the broadcast entirely.
5. Net effect for import: no new etcd key family, no new field on `ImportJob`, no secondary index, no GC coupling, and no idempotency retention window separate from what the broadcaster already maintains.

Several decisions from your design carry straight into this, and I want you to know which ones, because they were the right calls:

- **Per-collection (resource) scoping of the key** — the unified check scopes by the broadcast's resource key for exactly the reason you scoped by `collectionID`.
- **Ignore the payload; a repeat key returns the original `jobID`** — this is now the agreed semantics across both features, matching the wording in your issue.
- **A genuine re-import requires a new key** (e.g. `run_id_batch_1_attempt_2`), and the key needs a documented length/character bound.
- And the recovery story you argued for — a client that lost the response gets the original `jobID` back by simply retrying — is preserved.

The reason I'm going to take the implementation from here rather than ask you to rework this PR a third time: the remaining work is mostly in `streamingcoord` internals — broadcast-task lifetime and retention bounds, restart recovery of the dedup state, ordering against the collection-level resource lock that the import broadcast already holds, and avoiding a "sticky" key leaking onto unrelated internal RPCs. It's a lot of context that lives in a subsystem you'd have to absorb from scratch, and it would be unfair to hand you that after you've already delivered a complete, working implementation of what was asked for.

So I'm going to close this PR — but I want to be precise about what that does and doesn't mean. It closes this patch, not your contribution. #50954 stays open and stays yours; I'll link this PR and your design doc from the follow-up so the reasoning is traceable, and I'd genuinely like you to review that PR when it's up — you're the actual user of this feature, and nobody on our side is running Airflow/Temporal pipelines against BulkImport the way you are. If there's anything in the unified design that doesn't work for your orchestrator, I'd much rather hear it from you before it merges than after.

Thank you again for the time and care you put into this, and sorry that the framework decision landed later than your work did. Please do keep filing issues like #50954 — that one was unusually well-posed and it's a big part of why this is getting fixed properly.

sagar-arora:
Thanks @bigsheeper for looking into. I learnt myself a lot looking into milvus code. I am also excited that you are prioritizing this effort. Please let us know when you finish this change and release it. We are looking forward to using this feature.

bigsheeper:
Hi @sagar-arora, my apologies for the late reply, and thank you for your patience. You asked us to tell you when this work was finished. I am happy to say that idempotent BulkImport is now merged and released.

Thank you again for #50954 and for #52320. Your problem statement and your design decisions shaped the final implementation. The key is scoped to the collection, a retry with the same key returns the original `jobId`, and a real re-import needs a new key. All of these came from your proposal. The implementation follows the unified design described in https://github.com/milvus-io/milvus/pull/52320#issuecomment-5289323605.

**Server**

The feature is available in **Milvus v2.6.24** and **Milvus v3.0.2**, and in all later releases. The implementation PR is #52544.

**How to use it now (RESTful)**

Send the key as the `Idempotency-Key` HTTP header. No request body change is necessary.

```bash
curl -X POST "http://localhost:19530/v2/vectordb/jobs/import/create" \
  -H "Authorization: Bearer root:Milvus" \
  -H "Content-Type: application/json" \
  -H "Idempotency-Key: run_id_batch_1" \
  -d '{"collectionName": "events", "files": [["events/part-0.parquet"]]}'
```

- The key must be printable ASCII, at most 256 bytes.
- Milvus remembers a key for about 24 hours. Heavy DDL or import traffic can shorten this window.
- If the original job failed, the same key continues to return that failed `jobId`. Use a new key for a real re-import, for example `run_id_batch_1_attempt_2`.

**pymilvus**

pymilvus support is not released yet, sorry about that. milvus-io/pymilvus#3784 adds a `headers=` parameter to `bulk_import`, and the backports are milvus-io/pymilvus#3807 (3.0) and milvus-io/pymilvus#3808 (2.6). Until they are released, please use the RESTful API directly.

**Docs**

The user guide is in review: milvus-io/milvus-docs#3635. It gives the key rules, the retention window, and the related configuration.

You run this workload in production, so your feedback is very valuable to us. If something does not fit your orchestrator, or if the docs are not clear, please tell us here. We really appreciate your contribution, and we hope to see more from you in Milvus.

Issue #52327, opened 2026-08-08
[Bug]: delegator leaks bloom filter resources when segment load fails before publish


bigsheeper:
## What happened

`shardDelegator.LoadSegments` can leak charged sealed-segment `BloomFilterSet` resources when post-load processing fails after `LoadBloomFilterSet` succeeds but before `loadStreamDelete` publishes the segment into `distribution`.

The affected normal error paths include schema barrier changes, delete forwarding failures, or BM25 stats load failures after BF candidates have been loaded. At that point the candidates are not owned by `distribution`, so later `RemoveDistributions`/`Close` cleanup cannot refund them.

## Impact

Repeated failed segment loads can leave BF memory charged even though the segment never becomes visible in the delegator distribution. This can incorrectly consume querynode loaded-resource accounting and make later loads fail earlier than they should.

## Reproduction evidence

A focused test that forces a schema barrier failure after `LoadBloomFilterSet` reproduces the leak:

```bash
[...]

PR #52329, opened 2026-08-08
fix: refund delegator bloom filters on load failure

bigsheeper:
issue: #52327

## What changed

`LoadSegments` now refunds loaded BF candidates when post-load processing fails before `loadStreamDelete` publishes the segment entries into `distribution`. Once `loadStreamDelete` succeeds, ownership stays with `distribution` as before.

## Why

`LoadBloomFilterSet` may return charged `BloomFilterSet` candidates. If a later pre-publish step fails, such as a schema barrier check, those candidates are not reachable by `distribution` cleanup and remain charged.

## Tests

```bash
conda run -n milvus2 bash -lc 'export GOWORK=off; export PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage:${PKG_CONFIG_PATH:-}; export LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib:${LD_LIBRARY_PATH:-}; go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator -run "TestDelegatorDataSuite/TestLoadSegments/post_load_failure_refunds_bloom_filter_set"'\n```\n\n```bash\nconda run -n milvus2 bash -lc 'export GOWORK=off; export PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/thirdparty/milvus-storage:${PKG_CONFIG_PATH:-}; export LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/cmake_build/lib:${LD_LIBRARY_PATH:-}; go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/delegator'\n```\n\n```bash\nconda run -n milvus2 bash -lc 'export GOWORK=off; make static-check'\n```\n

### DIFF
--- a/internal/querynodev2/delegator/delegator_data.go
+++ b/internal/querynodev2/delegator/delegator_data.go
@@ -727,6 +727,17 @@ func (sd *shardDelegator) LoadSegments(ctx context.Context, req *querypb.LoadSeg
 			log.Warn(ctx, "failed to load bloom filter set for segment", mlog.Err(err))
 			return err
 		}
+		candidatesPublished := false
+		defer func() {
+			if candidatesPublished {
+				return
+			}
+			for _, candidate := range candidates {
+				if candidate != nil {
+					candidate.Refund()
+				}
+			}
+		}()
 
 		// Load BM25 stats BEFORE loadStreamDelete so stats are ready before segment becomes visible
 		err = sd.loadBM25Stats(ctx, infos, req)
@@ -767,6 +778,7 @@ func (sd *shardDelegator) LoadSegments(ctx context.Context, req *querypb.LoadSeg
 			// automatically by SyncDistribution when the segment is not in target.
 			return err
 		}
+		candidatesPublished = true
 		log.Debug(ctx, "load stream delete done")
 
 		return nil
--- a/internal/querynodev2/delegator/delegator_data_test.go
+++ b/internal/querynodev2/delegator/delegator_data_test.go
@@ -1289,6 +1289,66 @@ func (s *DelegatorDataSuite) TestLoadSegments() {
 		s.Error(err)
 	})
 
+	s.Run("post_load_failure_refunds_bloom_filter_set", func() {
+		defer func() {
+			s.workerManager.ExpectedCalls = nil
+			s.loader.ExpectedCalls = nil
+		}()
+
+		bfs := pkoracle.NewBloomFilterSet(100, 500, commonpb.SegmentState_Sealed)
+		bf := bloomfilter.NewBloomFilterWithType(
+			paramtable.Get().CommonCfg.BloomFilterSize.GetAsUint(),
+			paramtable.Get().CommonCfg.MaxBloomFalsePositive.GetAsFloat(),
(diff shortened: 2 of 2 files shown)


czs007:

Adversarial review confirms the ownership-transfer fix itself is correct; the two remaining items are both in the new test and are non-blocking.

## Low

- **`internal/querynodev2/delegator/delegator_data_test.go:1310`** — The new subtest calls `bfs.SetResourceCharged(true)`, which only flips the Go-side `resourceCharged` flag and sets `trackedSize = memSizeLocked()` (`internal/querynodev2/pkoracle/bloom_filter_set.go:272-281`); it never calls `C.ChargeLoadedResource`. Because `AddHistoricalStats(pks)` has already populated a real bloom filter, `trackedSize > 0`, so the `Refund()` triggered by this PR's new defer really does execute `C.RefundLoadedResource` (`bloom_filter_set.go:246-262`), subtracting bytes that were never added to segcore's process-global accounting. Failure: running `go test -tags dynamic,test ./internal/querynodev2/delegator -run TestDelegatorDataSuite/TestLoadSegments/post_load_failure_refunds_bloom_filter_set` → an unpaired refund drives `DList::total_loaded_size_` negative (clamped to zero with a `[MCL] total_loaded_size_ became negative after refund loaded resource` ERROR log), and the unclamped `cache_loaded_bytes` Prometheus gauge goes negative; any real charge earlier in the same test binary is under-counted. Suggestion: call `bfs.Charge()` after populating the filter — matching the production path at `internal/querynodev2/segments/segment_loader.go:793-795` — so charge and refund are paired; assertions stay unchanged. (An existing test in the same area deliberately avoids this: `internal/querynodev2/pkoracle/pk_oracle_test.go:226`, "We don't set resourceCharged=true to avoid requiring C library initialization".) (raised by tinswzy, sijie-ni-0214)

## Nitpick

- **`internal/querynodev2/delegator/delegator_data_test.go:1326`** — The subtest sets `s.delegator.schemaBarrierTs = 1` and never restores it. `SetupTest` (`delegator_data_test.go:267-290`) rebuilds `s.delegator` once per Test method, so all `s.Run` subtests inside `TestLoadSegments` share one instance and the value leaks forward. This does not bite today: the only subsequent subtest, `worker_load_fail` (from line 1352), returns early on the mocked `worker1.LoadSegments` error (`delegator_data.go:706-709`) and never reaches the barrier comparison in `addDistributionIfSchemaBarrierOK` (`delegator_data.go:793-797`). Suggestion: `defer` restoring the original value, or move this subtest to the end of `TestLoadSegments`, so a future "load succeeds" subtest added after it is not silently broken by the residual barrier. (raised by tinswzy)

Issue #52328, opened 2026-08-08
bug: preserve NULL entity-TTL semantics across compaction rewrites


MrPresent-Han:
## Problem

Entity-level TTL defines a NULL TTL value as **never expires**. Compaction must preserve this semantic for both sources of NULL:

- a TTL column that physically exists and stores NULL for a row;
- a configured TTL column that is physically absent from an old segment and is materialized as NULL by the compaction read/materialization path.

The shared compaction paths currently break this invariant at two layers.

First, Mix, Sort, MergeSort, and Clustering all call public `NewRecordMaterializer`. For a physically absent configured TTL field, the materializer currently applies the schema default before `EntityFilter` sees the row.

Second, even after RecordMaterializer is corrected to produce NULL for an absent TTL field, the following downstream paths can replace that NULL with the schema default again:

- **Mix compaction, filtered branch**: after any row in a batch is filtered, `internal/datanode/compactor/mix_compactor.go` copies surviving ranges through `storage.RecordBuilder`.
- **Sort / MergeSort**: `internal/storage/sort.go` emits retained rows through the default-aware append path / `RecordBuilder`.
- **Clustering compaction**: `internal/datanode/compactor/clustering_compactor.go` calls `storage.ValueDeserializerWithSchema`; the deserializer replaces an Arrow NULL with `GetDefaultValue(field)` before TTL filtering.

`RecordBuilder.Append` passes `field.GetDefaultValue()` to `appendValueAt`. For TIMESTAMPTZ, the shared `Int64Builder` branch also reads `GetLongData()`, which can serialize a TIMESTAMPTZ default as `0`; either way, NULL validity has been destroyed.

This contradicts `docs/design-docs/design_docs/20251221-entity_level_ttl.md`, which defines NULL TTL as never-expiring and requires explicit upsert to give historical data expiration values.

### A. Physically stored NULL

1. A collection configures a nullable TIMESTAMPTZ field as `ttl_field`, and that field has a default earlier than `now`.
2. A row physically stores `ttl_field = NULL`; it is visible and must never expire.
3. A compaction runs:
   - Mix enters the `RecordBuilder` branch because an unrelated row is deleted or expired;
   - Sort or MergeSort rewrites the row normally; or
   - Clustering deserializes it through `ValueDeserializerWithSchema`.
4. The physical NULL is replaced with the schema default or epoch `0`.
[...]

PR #52330, opened 2026-08-09
fix: preserve absent TTL as null (#51649)

MrPresent-Han:
issue: #51649

Depends on #52301.

## Summary

- In BumpSchemaVersionCompaction full rewrite, materialize a TTL field that is configured in the task schema but absent from the source segment as physical NULL, rather than its schema default.
- Preserve an existing physical NULL TTL when RecordMaterializer rebuilds a filtered Bump record.
- Keep RecordMaterializer's default-materialization behavior unchanged for non-TTL fields and callers that do not opt into the Bump-specific policy.

## Safety invariant

Within BumpSchemaVersion full rewrite, an absent TTL field already configured in the task schema must materialize as NULL. The rewrite must not persist the TTL schema default into historical rows.

## Scope

- This PR covers only BumpSchemaVersionCompaction full rewrite and its RecordMaterializer policy.
- Cross-compaction preservation of physical or absent NULL TTL values in Mix, Sort, MergeSort, and Clustering is tracked in #52328.
- `ttl_field` activation/repoint semantics and read/query-side handling of historical defaults are tracked in #52333.

## Verification

- `go test -tags dynamic,test -gcflags='all=-N -l' -count=1 ./internal/datanode/compactor -run '^TestBumpSchemaVersionCompactionTaskSuite/TestFullRewriteMissingTTLFieldDoesNotProbeRecord$'`
- `go test -tags dynamic,test -gcflags='all=-N -l' -count=1 ./internal/datanode/compactor`

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -454,6 +454,10 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 	}
 	entityFilter := compaction.NewEntityFilter(delta, t.plan.GetCollectionTtl(), t.currentTime, segment.GetCommitTimestamp())
 	ttlFieldID := getTTLFieldID(t.plan.GetSchema())
+	var preserveNullFields map[int64]struct{}
+	if ttlFieldID >= common.StartOfUserFieldID {
+		preserveNullFields = map[int64]struct{}{ttlFieldID: {}}
+	}
 	reader, _, err := newCompactionSegmentRecordReaderWithFields(t.ctx, segment, t.plan.GetSchema(), t.compactionParams.StorageConfig, existingFields,
 		storage.WithCollectionID(collectionID),
 		storage.WithVersion(segment.GetStorageVersion()),
@@ -465,7 +469,7 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 	}
 	defer reader.Close()
 
-	materializer, err := NewRecordMaterializer(t.plan.GetSchema(), t.plan.GetSchema().GetFunctions(), existingFields)
+	materializer, err := newRecordMaterializer(t.plan.GetSchema(), t.plan.GetSchema().GetFunctions(), existingFields, preserveNullFields)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -514,6 +514,11 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 		}
 	}()
 
+	// Source TTL presence is decided from existingFields, never by probing the
+	// record: V2/V3 records panic on Column for a field they do not carry.
+	_, sourceHasTTL := existingFields[ttlFieldID]
+	sourceHasTTLField := ttlFieldID >= common.StartOfUserFieldID && sourceHasTTL
+	preMaterializeFilter := len(delta) > 0 || t.plan.GetCollectionTtl() > 0 || sourceHasTTLField
 	var totalRows int64
 	for {
 		record, err := reader.Next()
@@ -524,8 +529,6 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 			return nil, err
 		}
 
-		sourceHasTTLField := ttlFieldID >= common.StartOfUserFieldID && record.Column(ttlFieldID) != nil
-		preMaterializeFilter := len(delta) > 0 || t.plan.GetCollectionTtl() > 0 || sourceHasTTLField
 		var selection *recordSelection
 		if preMaterializeFilter {
 			selection, _, err = selectFullRewriteRecord(record, pkField, entityFilter, ttlFieldID, sourceHasTTLField, nil)
--- a/internal/datanode/compactor/bump_schema_version_compactor_test.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor_test.go
@@ -1510,6 +1510,48 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestFullRewriteMissingTTLFieldDoe
 	// Source segment carries a dropped field (103) -> droppedFieldIDs>0 -> full rewrite.
 	s.prepareBumpSchemaVersionCompactionWithDroppedField()
 
+	// Attach a real V3 manifest deltalog so the full rewrite filters one row and
+	// exercises selection/newSelectedRecord against the packed source record.
+	sourceSegment := s.task.plan.GetSegmentBinlogs()[0]
+	const deltaLogID = int64(20000)
+	basePath, _, err := packed.UnmarshalManifestPath(sourceSegment.GetManifest())
+	s.Require().NoError(err)
+	deltaPath := metautil.BuildDeltaLogPathV3(basePath, deltaLogID)
+	deleteRecord, _, _, err := storage.BuildDeleteRecord(
+		[]storage.PrimaryKey{storage.NewInt64PrimaryKey(sourceSegment.GetSegmentID())},
+		[]uint64{tsoutil.ComposeTSByTime(getMilvusBirthday().Add(time.Second))},
+	)
+	s.Require().NoError(err)
+	defer deleteRecord.Release()
+	deltaWriter, err := storage.NewDeltalogWriter(
+		context.Background(),
+		sourceSegment.GetCollectionID(),
+		sourceSegment.GetPartitionID(),
+		sourceSegment.GetSegmentID(),
+		deltaLogID,
+		schemapb.DataType_Int64,
+		deltaPath,
+		storage.WithVersion(storage.StorageV2),
+		storage.WithStorageConfig(s.task.compactionParams.StorageConfig),
+	)
+	s.Require().NoError(err)
+	s.Require().NoError(deltaWriter.Write(deleteRecord))
+	s.Require().NoError(deltaWriter.Close())
+	sourceSegment.Manifest, err = packed.AddDeltaLogsToManifest(
+		sourceSegment.GetManifest(),
+		s.task.compactionParams.StorageConfig,
+		[]packed.DeltaLogEntry{{Path: deltaPath, NumEntries: 1}},
+	)
+	s.Require().NoError(err)
+	sourceSegment.Deltalogs = []*datapb.FieldBinlog{{
+		FieldID: 100,
+		Binlogs: []*datapb.Binlog{{
+			LogID:      deltaLogID,
+			LogPath:    deltaPath,
+			EntriesNum: 1,
+		}},
+	}}
+
 	result, err := s.task.Compact()
 	s.Require().NoError(err)
 	s.Require().NotNil(result)
@@ -1517,7 +1559,7 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestFullRewriteMissingTTLFieldDoe
 	s.Require().Len(result.GetSegments(), 1)
 
 	segment := result.GetSegments()[0]
-	s.EqualValues(3, segment.GetNumOfRows())
+	s.EqualValues(2, segment.GetNumOfRows())
 
 	reader, err := storage.NewManifestRecordReader(
 		context.Background(),
@@ -1534,5 +1576,5 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestFullRewriteMissingTTLFieldDoe
 	s.Require().NoError(err)
 	ttlColumn, ok := record.Column(newTTLFieldID).(*array.Int64)
 	s.Require().True(ok)
-	s.Equal(3, ttlColumn.NullN())
+	s.Equal(2, ttlColumn.NullN())
 }
--- a/internal/datanode/compactor/bump_schema_version_compactor_test.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor_test.go
@@ -1486,6 +1486,8 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestMissingFunctionInputSchemaByF
 // TTL column.
 func (s *BumpSchemaVersionCompactionTaskSuite) TestFullRewriteMissingTTLFieldDoesNotProbeRecord() {
 	const newTTLFieldID = int64(105)
+	s.task.currentTime = getMilvusBirthday().Add(time.Hour)
+	expiredDefault := s.task.currentTime.Add(-time.Minute).UnixMicro()
 	// Target schema gains a nullable Timestamptz field absent from the source
 	// segment and designates it as the collection TTL field.
 	s.task.plan.Schema.Fields = append(s.task.plan.Schema.Fields, &schemapb.FieldSchema{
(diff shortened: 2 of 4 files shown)


internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -444,6 +444,10 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 	}
 	entityFilter := compaction.NewEntityFilter(delta, t.plan.GetCollectionTtl(), t.currentTime, segment.GetCommitTimestamp())
 	ttlFieldID := getTTLFieldID(t.plan.GetSchema())
+	var preserveNullFields map[int64]struct{}
+	if ttlFieldID >= common.StartOfUserFieldID {
```
buqian-zilliz, line 458: **[P1] Invalidate rewrites after a TTL-property change**

`preserveNullFields` is derived from the schema captured when the bump view was created, but a `ttl_field` alter intentionally keeps the same schema version. If the property is enabled or repointed while this full rewrite is queued or running, an absent future TTL column is still materialized with its schema default. DataCoord completion does not compare the task's properties with the current collection, so it accepts the replacement after the new TTL property is active; the replacement is stamped with the unchanged version, so no corrective bump is scheduled. Those historical rows can then expire and be deleted without an upsert. Please reject and replan a result whose TTL-property snapshot no longer matches the collection, or serialize this alter with an equivalent configuration epoch.

internal/datanode/compactor/record_materializer.go
```diff
@@ -54,14 +55,23 @@ func (s *recordSelection) Len() int {
...
+	materializers      []FunctionMaterializer
+	missingFields      []*schemapb.FieldSchema
+	schema             *schemapb.CollectionSchema
+	existingFields     map[int64]struct{}
+	preserveNullFields map[int64]struct{}
 }
 
 func NewRecordMaterializer(schema *schemapb.CollectionSchema, functions []*schemapb.FunctionSchema, existingFields map[int64]struct{}) (*RecordMaterializer, error) {
-	materializer := &RecordMaterializer{schema: schema, existingFields: existingFields}
+	return newRecordMaterializer(schema, functions, existingFields, nil)
```
liliu-z, line 66: `newRecordMaterializer` protects NULLs only for callers that pass `preserveNullFields`, and only `runFullSchemaRewrite` does. Mix (`mix_compactor.go:304`), sort (`sort_compaction.go:263`), merge-sort (`merge_sort.go:84`) and clustering (`clustering_compactor.go:622`) still go through `NewRecordMaterializer(...)`, so a physically absent TTL column is materialized from the schema default at `record_materializer.go:146`; each of those paths then reads that synthesized column and feeds it to `EntityFilter.Filtered` before writing. A TTL field added later with a default timestamp in the past therefore makes every row of every old segment count as expired and get dropped — the exact inverse of the invariant this PR establishes. #52328 explicitly excludes 'physically absent TTL columns' and `RecordMaterializer`, so nothing else tracks this. Deriving `preserveNullFields` from `getTTLFieldID(schema)` inside `newRecordMaterializer` covers all callers at once; every one of them passes `plan.GetSchema()`, which carries the `ttl_field` property. The fifth call site (`bump_schema_version_compactor.go:1006`) is unaffected because `partialMaterializerExistingFields` marks the TTL field as existing.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -444,6 +444,10 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
 	}
 	entityFilter := compaction.NewEntityFilter(delta, t.plan.GetCollectionTtl(), t.currentTime, segment.GetCommitTimestamp())
 	ttlFieldID := getTTLFieldID(t.plan.GetSchema())
+	var preserveNullFields map[int64]struct{}
```
liliu-z, line 457: The path this PR hardens is the rare one: `Compact` reaches `runFullSchemaRewrite` only when the bump also drops a field. A plain AddField that introduces a new `ttl_field` has no dropped fields and no missing functions, so it takes `runSchemaVersionBumpOnly`, which republishes the source binlogs and manifest verbatim and leaves the TTL column physically absent. On load, segcore fills such a field via `DefaultValueChunkTranslator`, which emits the schema default (not NULL) when one is configured, and the injected TTL predicate `ttl IS NULL OR ttl > now` then evaluates against that default — an already-expired default hides the entire segment from queries. So the PR's stated invariant, 'only a physically stored non-NULL TTL value may enable entity expiration', does not hold for the common AddField path. Either extend the fix to cover an absent TTL column on the read side, or narrow the claim in the PR body to the full-rewrite path.

internal/datanode/compactor/bump_schema_version_compactor_test.go
```diff
@@ -1428,3 +1428,63 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestMaterializationRejectsDropped
...
+
+	result, err := s.task.Compact()
+	s.Require().NoError(err)
+	s.Require().NotNil(result)
+	s.Equal(datapb.CompactionTaskState_completed, result.GetState())
+	s.Require().Len(result.GetSegments(), 1)
+
+	segment := result.GetSegments()[0]
+	s.EqualValues(3, segment.GetNumOfRows())
+	s.Equal([]int64{math.MaxInt64, math.MaxInt64, math.MaxInt64, math.MaxInt64, math.MaxInt64}, segment.GetExpirQuantiles())
```
liliu-z: This assertion is vacuous. Only `PackedBinlogRecordWriter.Write` appends to `ttlFieldValues`; the manifest writers used for StorageV3 (`PackedManifestRecordWriter` / `PackedTextManifestRecordWriter`) never populate it, so `calculateExpirQuantiles` always falls into the never-expire branch and returns `[MaxInt64 x5]` with or without this fix. The `ttlColumn.NullN() == 3` check below is the only line that actually pins the new behavior. Drop the assertion, or annotate it as pinning a known V3 gap, so it is not read as evidence that the expiry-compaction chain is covered.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Thanks for calling out the stale-snapshot case. After tracing the workflow, I think the root issue is broader than Bump: `ttl_field` is a property-only activation/repoint operation, so any ordinary `TIMESTAMPTZ` field whose historical rows already expose or persist a default can later be reinterpreted as expiration data—even with no Bump in flight or after Bump has completed. Bump cannot predict which field may be selected as TTL in the future, so I don't plan to change #52330 for this thread; it remains scoped to preserving an absent TTL that is already configured in the task schema. I tracked the general TTL activation/repoint lifecycle gap in #52333.

internal/datanode/compactor/record_materializer.go
MrPresent-Han: Thanks. The shared callers are a real risk, but deriving `preserveNullFields` inside RecordMaterializer is only the first step. Mix/Sort/MergeSort later pass the resulting NULL through default-aware RecordBuilder paths, and Clustering's `ValueDeserializerWithSchema` converts it back to the schema default before `EntityFilter`, so changing RM alone would still allow deletion. I’m keeping #52330 scoped to Bump full rewrite and will not change RecordBuilder or Clustering deserialization in this PR. I expanded #52328 to cover both physically stored NULL and a physically absent TTL column materialized as NULL, including end-to-end preservation through those downstream paths.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Agreed on the behavior. A plain AddField can take the bump-only path and leave the TTL column physically absent, while QueryNode projects the schema default for that column. That is part of the broader `ttl_field` activation/read-side semantics, not the full-rewrite persistence boundary of this PR, so I won't extend #52330 into the read path. The general activation/repoint lifecycle problem is tracked in #52333; this PR remains limited to preventing Bump full rewrite from persisting the schema default for an absent TTL field already configured in the task schema.

internal/datanode/compactor/bump_schema_version_compactor_test.go
MrPresent-Han: Removed the `ExpirQuantiles == [MaxInt64 x5]` assertion in `c97f068962`. As noted, the V3 manifest writers do not collect TTL values, so that assertion could not distinguish default-filled output from NULL output; the test now relies on `ttlColumn.NullN() == 3`, which directly pins this fix. Verification passed with the required flags for both the targeted regression test and the full `./internal/datanode/compactor` package. The V3 writer quantile gap remains tracked in #52333.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -444,6 +444,10 @@ func (t *bumpSchemaVersionCompactionTask) runFullSchemaRewrite(existingFields ma
...
+	if ttlFieldID >= common.StartOfUserFieldID {
+		preserveNullFields = map[int64]struct{}{ttlFieldID: {}}
```
liliu-z, line 459: The physical NULLs this change writes into the TTL column are re-defaulted by every downstream compaction that rebuilds records through storage.NewRecordBuilder + appendValueAt — storage.Sort (internal/storage/sort.go:166 and :194), storage.MergeSort (internal/storage/sort.go:444 and :456), and mix_compactor.go:369 — and appendValueAt's Int64Builder branch resolves a Timestamptz default via defaultValue.GetLongData(), which returns 0 for a ValueField_TimestamptzData oneof, so a preserved NULL becomes TTL=0 after the first sort or merge-sort. Both EntityFilterImpl.isEntityExpiredByTTLField (0 is not < 0, and now >= 0) and the segcore predicate 'ttl_field is null OR ttl_field > physical_us' then classify those rows as expired, so they are silently dropped rather than merely mis-dated. fieldForMaterialization only takes effect for a TTL field that carries a schema default, which is exactly the population this PR fixes, so the stated safety invariant does not hold end-to-end until #52328 lands — fold it in, or describe the fix as partial rather than as an invariant.

MrPresent-Han: Agreed on the downstream behavior, including the TIMESTAMPTZ default being read through `GetLongData()` as `0`. #52330 is intentionally a partial fix with respect to the full compaction pipeline, but complete for its owning boundary: BumpSchemaVersion full rewrite must materialize and persist an absent configured TTL field as NULL rather than its schema default.

The PR body has already been narrowed to that local guarantee and explicitly assigns end-to-end NULL preservation across RecordMaterializer, Mix/Sort/MergeSort RecordBuilder, and Clustering deserialization to #52328, which also records this NULL-to-0 failure. I will keep those downstream fixes out of #52330 so this PR does not cross the Bump full-rewrite boundary.

liliu-z:
**internal/storage/arrow_util.go:140** In `appendValueAt`, `if ia.IsNull(idx) { if defaultValue != nil { b.Append(defaultValue.GetLongData()) } }` — timestamptz shares the Int64 builder (serde.go:315), and `GetLongData()` on a `ValueField_TimestamptzData` oneof type-asserts against `*ValueField_LongData`, fails, and returns 0. This is the only default-materialization site in the repo without a Timestamptz case (`GenerateEmptyArrayFromSchema` at arrow_util.go:341, `storage.GetDefaultValue` at utils.go:1814 and proxy `validate_util.go:683` all handle it), so the invariant "a NULL column materializes to the schema default" is broken specifically for the TTL type. Concretely, MergeSort (sort.go:456), sortRecords (sort.go:194) and the mix compactor builder (mix_compactor.go:369) write 0 into the TTL field, and segcore's `ttl_field IS NULL OR ttl_field > now` (Expr.cpp:101) treats epoch 0 as already expired, so those rows disappear from query results immediately after compaction; add the two-line Timestamptz case here.

internal/datanode/compactor/bump_schema_version_compactor_test.go
```diff
@@ -1428,3 +1428,62 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestMaterializationRejectsDropped
...
+// panics on V3 records ("no such field") before the materializer can fill the null
+// TTL column.
+func (s *BumpSchemaVersionCompactionTaskSuite) TestFullRewriteMissingTTLFieldDoesNotProbeRecord() {
+	const newTTLFieldID = int64(105)
+	s.task.currentTime = getMilvusBirthday().Add(time.Hour)
+	expiredDefault := s.task.currentTime.Add(-time.Minute).UnixMicro()
+	// Target schema gains a nullable Timestamptz field absent from the source
+	// segment and designates it as the collection TTL field.
+	s.task.plan.Schema.Fields = append(s.task.plan.Schema.Fields, &schemapb.FieldSchema{
+		FieldID:  newTTLFieldID,
```
liliu-z, line 1494: The new end-to-end test supplies no deltalogs and no collection TTL, so `preMaterializeFilter` is false and `selection` stays nil, meaning it only covers the `GenerateEmptyArrayFromSchema` branch. The `newSelectedRecord` → `buildSelectedColumn` half is then exercised only by the hand-built `materializerTestRecord`, which is exactly the synthetic test shape that hid the V3 `Column` panic fixed in commit 1 — so a regression on that path would again go undetected. Adding a deltalog to this test closes the gap in roughly ten lines.

liliu-z:
**internal/storage/serde_events.go:188** `ValueDeserializerWithSchema` restores the real schema default for a NULL timestamptz TTL column, which converts "no expiration" into a concrete expiration timestamp; fixing `GetLongData()` → `GetTimestamptzData()` only removes the accidental epoch-0 case, not this one. `isEntityExpiredByTTLField` compares `currentTime >= expirationTimeMicros`, so NULL and "default" are not interchangeable for a TTL field — any default that is already in the past means the materialized value is born expired. A collection with a past-dated timestamptz default silently loses every row whose TTL was NULL the first time clustering compaction runs, with no error surfaced; this needs end-to-end NULL preservation (#52328) landing with or before this PR, i.e. a hard merge dependency rather than a follow-up.

internal/datanode/compactor/bump_schema_version_compactor_test.go
MrPresent-Han: Added a manifest-backed V3 deltalog in `1bc602a7f9` that deletes one source row, so this regression test now forces `preMaterializeFilter` and a non-nil `selection` against the real packed record. It asserts that the two surviving rows still have two NULL TTL values, covering the absent-column path through `newSelectedRecord` without probing the missing V3 column.

Verification passed with the required flags for both the targeted regression test and the full `./internal/datanode/compactor` package.

MrPresent-Han:
Regarding Li Liu's `appendValueAt` comment: https://github.com/milvus-io/milvus/pull/52330#issuecomment-5231360767

Agreed that the shared Int64 builder reads a TIMESTAMPTZ default through `GetLongData()` and therefore writes `0`. I am not adding the proposed two-line shared-storage change in #52330: changing `0` to the real schema default still converts TTL NULL into a concrete expiration value and still permits deletion when that default is in the past. The RecordBuilder/Sort/MergeSort/Mix NULL-preservation fix, including this NULL-to-0 case, remains tracked end to end in #52328.

MrPresent-Han:
Regarding Li Liu's Clustering comment: https://github.com/milvus-io/milvus/pull/52330#issuecomment-5231360833

Agreed that `ValueDeserializerWithSchema` replaces a TTL NULL with the schema default before Clustering's `EntityFilter`, which can delete the row in the same compaction. I do not think #52328 needs to be a hard merge dependency for #52330: that downstream defect predates this PR, while #52330 independently removes default persistence at the Bump full-rewrite boundary and does not make any existing path less safe. The Clustering and shared-compaction end-to-end fix remains the owning scope of #52328.

PR #52277, opened 2026-08-06
fix: make DataCoord task retries and cleanup crash-safe

xiaofan-luan:
issue: #52332

## Problem

DataCoord previously split retry ownership between the global scheduler and the task-specific inspectors. After an unanswered create or query RPC, the old DataNode attempt may still be running. Reusing its task, plan, segment, or build identity lets late callbacks and object-storage writes collide with a replacement attempt. Terminal compaction cleanup could also race a worker callback and leave input segments permanently claimed.

This affects compaction, index, analyze, stats, pre-import/import, copy-segment, and external-collection refresh. The previous 600-second worker RPC timeout also let one unavailable DataNode hold a scheduler round for up to ten minutes.

### Make the global scheduler dispatch-only

- Removes scheduler-owned retry/backoff and its duplicated pending/running containers.
- Keeps one scheduler ownership map and one per-task callback lock.
- `Finalize` removes dispatch ownership before terminal cleanup; `Update` serializes business state changes with an in-flight callback for the same task.
- Moves retry state, interval, and attempt policy into each task inspector.

### Use fresh worker identities for retries

| Task | Replacement identity |
| --- | --- |
| Compaction | Fresh `PlanID` and output segment range; clustering also gets a fresh analyze task ID |
| Index | Fresh `BuildID` |
| Analyze | Ends the parent clustering attempt; replacement compaction allocates a fresh analyze task ID |
| Stats | Fresh task ID |
| PreImport | Fresh task ID |
| Import | Fresh task ID and output segment IDs |
| Copy Segment | Fresh task ID and target segment IDs/mapping |
| External Refresh | Fresh task ID and output segment range |

The logical job/trigger and immutable work description are preserved. Dropping the old DataNode task remains best effort: after its metadata record is removed, a late result from that identity has no coordinator-side commit path.

### Align persistence and recovery

- Retry replacement removes the old record and writes the new record in one existing `Catalog.Update` before enqueueing the replacement, so coordinator metadata has one current attempt for each handoff.
- Result adoption and terminal task/job state changes that must agree are written together, including compaction `meta_saved` transitions and import/copy/external terminal paths.
- Ambiguous critical catalog writes use the accepted fail-stop model; restart reloads authoritative catalog state instead of guessing from stale in-memory metadata.
- Startup resumes persisted `Init`/`InProgress` work. Persisted `Retry` work is replaced on its owning inspector's business retry interval.
[...]

PR #52331, opened 2026-08-09
fix: stop one unresponsive DataNode from stalling task dispatch for 10 minutes

xiaofan-luan:
issue: #52332

## What this PR does

Lower `dataCoord.requestTimeoutSeconds` from 600s to 30s, and pass the request context into `NewChunkManager` on the DataNode task handlers so the timeout actually reaches the one slow step behind them.

## Why 600s was wrong

`dataCoord.requestTimeoutSeconds` bounds every DataCoord-to-worker call: `createTask`, `queryTask`, `dropTask` and `QuerySlot`. All four are control-plane operations that a healthy node answers in microseconds — `CreateTask` hands the plan to a background queue rather than running it, `QueryTask` and `QuerySlot` read in-memory state, `DropTask` removes a map entry. None of them is a slow operation, so the 600s default only ever applied to a node that was already unwell — and there the useful behavior is to give up and re-dispatch elsewhere.

Meanwhile the global task scheduler waits for every RPC it issues in a round, and `QuerySlot` is the first thing `schedule()` does, waiting on **all** nodes. One black-holed DataNode could therefore stall dispatch for every task type — compaction, index, stats, import — for ten minutes per round.

## Why the ctx change is needed for the timeout to matter

Creating a chunk manager is the one genuinely slow step on the `CreateTask` path: it builds a fresh client, probes the bucket with a HEAD request and may create it, all under `retry.Attempts(20)` with backoff topping out near 48s. The seven task handlers passed `node.ctx`, so that probe ignored the caller's deadline entirely — shortening the DataCoord-side timeout alone would only make DataCoord give up while the DataNode kept going and registered the task anyway, turning a slow dispatch into a duplicate/orphan.

The context does not outlive creation — `RemoteChunkManager` stores no ctx and every data operation takes its own — so this bounds only the creation-time probe, which is exactly what should be bounded. The chunk manager built in `data_node.go` keeps `node.ctx`: it belongs to node startup, not to a request.

## Notes

- `dataCoord.requestTimeoutSeconds` is not exported in `milvus.yaml` (`Export: false`), so no existing user configuration is affected; deployments that want the old behavior can still set it explicitly.
- Known limitation, out of scope here: `createRefreshExternalCollectionTask` dispatches through a blocking 8-worker pool and can legitimately exceed this timeout when the pool is saturated; that needs a non-blocking submit and is left as follow-up.

### DIFF
--- a/internal/datanode/index_services.go
+++ b/internal/datanode/index_services.go
@@ -83,7 +83,7 @@ func (node *DataNode) CreateJob(ctx context.Context, req *workerpb.CreateJobRequ
 		metrics.DataNodeBuildIndexTaskCounter.WithLabelValues(paramtable.GetStringNodeID(), metrics.FailLabel).Inc()
 		return merr.Status(err), nil
 	}
-	cm, err := node.storageFactory.NewChunkManager(node.ctx, req.GetStorageConfig())
+	cm, err := node.storageFactory.NewChunkManager(ctx, req.GetStorageConfig())
 	if err != nil {
 		mlog.Error(context.TODO(), "create chunk manager failed", mlog.String("bucket", req.GetStorageConfig().GetBucketName()),
 			mlog.String("accessKey", req.GetStorageConfig().GetAccessKeyID()),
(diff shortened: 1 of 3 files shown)


xiaofan-luan:
Folded into #52277 (same DataCoord↔worker control-plane hardening; that PR's cleanup path depends on these timeouts behaving). Issue #52332 is now referenced there.

PR #52334, opened 2026-08-09
fix: [2.6]key proxy shard leader cache by collection id to close alias repoint race

xiaocai2333:
## What this PR does

Backports the Proxy shard-leader cache fix from #51779 to the 2.6 branch.

The shard cache is keyed by collection ID instead of database and collection name. This prevents an in-flight cache fill for an old alias target from restoring stale shard leaders after `AlterAlias` invalidation and routing a new collection plan to the old collection channels.

It also updates shard-cache invalidation and the related search, query, and statistic retry paths to evict entries by collection ID.

## Commits

1. `fix: key proxy shard leader cache by collection id ...` — the backport of #51779 itself.
2. `fix: collapse per-alias shard cache evictions into a single call` — review follow-up, see below.
3. `fix: pin py for pytest-html to unbreak every e2e pipeline (#52336)` — CI unblock, see below.

### Review follow-up: collapse per-alias evictions

On master, #51313 had already replaced the per-alias `for` loops in `InvalidateCollectionMetaCache` with a single variadic `deprecateShardCaches(...)` call. #51313 was never backported, so 2.6 still had the loops. Once the helper switched to evicting by collection ID (and therefore ignores the names it is handed), each loop degenerated into N identical deletes — N write-lock acquisitions and N identical log lines for a collection with N aliases.

The loops are collapsed into `deprecateShardCaches(aliasName...)`, matching master's call-site shape. The helper body is unchanged and remains byte-identical to master. Added `TestProxy_InvalidateCollectionMetaCache_ShardCacheEvictionByID`, which asserts the eviction count stays constant as the alias count grows, and that a request carrying no collection ID (the real shape of alias DDL broadcasts) evicts nothing.

### CI unblock: pin `py` for pytest-html

Cherry-pick of #52336 (commit `e8cd33bc9c`) from master. Without it this PR's `ci-v2/e2e-default` cannot pass, for a reason unrelated to any code here.

Since 2026-08-09 every e2e sub-pipeline dies before running a single test:

```
File ".../pytest_html/plugin.py", line 22, in <module>
    from py.xml import html
ModuleNotFoundError: No module named 'py.xml'; 'py' is not a package
```

The pinned `pytest-html==3.1.1` imports `py.xml`, and `py` used to arrive transitively through `pytest-forked` (a dependency of the pinned `pytest-xdist==2.5.0`). **pytest-forked 1.7.5, released 2026-08-08, dropped its `py` dependency**, so a fresh resolve no longer installs `py` and the plugin fails at import time — pytest aborts during plugin loading with zero tests collected.

This is fleet-wide on the 2.6 line, not specific to this PR: runs completed on 08-07 or earlier passed against the very same base commit (`d85dc9459d`, unchanged since), and unrelated PRs (#52337) fail with identical tracebacks. Evidence for this branch: dispatcher [16875](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-dispatcher/16875/), sub-job [39324](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-4am/39324/).

The pin is included here so this PR's CI is runnable. It is a test-dependency-only change; no product code is touched. #52349 carries the same cherry-pick as a standalone PR — whichever lands first unblocks the other, and the duplicate commit can then be dropped.

## Verification

- `ci-v2/build`, `ci-v2/ut-cpp`, `ci-v2/ut-go` all green on the previous head — `ut-go` covers `internal/proxy`, so the new test compiles and passes in CI.
- `go test -tags dynamic,test -count=1 ./internal/proxy/shardclient/...` passes locally, including `TestShardCacheAliasRepointResolvesByCollectionID`.
- Local execution of the full `internal/proxy` suite was blocked by mismatched CGo/native build artifacts and is left to CI.

issue: #51533
pr: #51779
pr: #52336

### DIFF
--- a/internal/proxy/impl.go
+++ b/internal/proxy/impl.go
@@ -157,9 +157,7 @@ func (node *Proxy) InvalidateCollectionMetaCache(ctx context.Context, request *p
 			}
 			if request.CollectionID != UniqueID(0) {
 				aliasName = globalMetaCache.RemoveCollectionsByID(ctx, collectionID, request.GetBase().GetTimestamp(), msgType == commonpb.MsgType_DropCollection)
-				for _, name := range aliasName {
-					deprecateShardCaches(name)
-				}
+				deprecateShardCaches(aliasName...)
 			}
 			// Invalidate alias cache for alias operations
 			if msgType == commonpb.MsgType_CreateAlias || msgType == commonpb.MsgType_AlterAlias || msgType == commonpb.MsgType_DropAlias {
@@ -172,9 +170,7 @@ func (node *Proxy) InvalidateCollectionMetaCache(ctx context.Context, request *p
 			// All the request from query use collectionID
 			if request.CollectionID != UniqueID(0) {
 				aliasName = globalMetaCache.RemoveCollectionsByID(ctx, collectionID, 0, false)
-				for _, name := range aliasName {
-					deprecateShardCaches(name)
-				}
+				deprecateShardCaches(aliasName...)
 			}
 			log.Info("complete to invalidate collection meta cache", zap.String("type", request.GetBase().GetMsgType().String()))
 		case commonpb.MsgType_CreatePartition, commonpb.MsgType_DropPartition:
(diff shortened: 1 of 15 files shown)


liliu-z (commented):
/lgtm
/approve

Issue #52342, opened 2026-08-09
[Enhancement]: Avoid fmt.Sprintf for internal querynodev2 column names


bigsheeper:
## What should be improved

`internal/querynodev2/tasks` builds internal temporary column names for group-by and boost-score paths with `fmt.Sprintf("%s%d", prefix, value)`. These helpers sit in the querynodev2 search reduce path:

- `groupByColumnName(fieldID)` is used when group-by results are propagated through reduce/marshal paths.
- `boostScoreColumn(index)` is used while building and pruning temporary boost-score columns for scorer chains.

Both helpers only concatenate a constant prefix with a decimal integer, so `strconv` avoids the formatting machinery while preserving the same string output.

## Proof before filing

Duplicate search before filing found no open issue or PR for this change.

Existing behavior-pinning tests passed on upstream `master` and after the local change:

```bash
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:$PKG_CONFIG_PATH \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib:$LD_LIBRARY_PATH \
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/tasks \
  -run 'TestBoostScoreColumn|TestApplyBoostScoresPrunesTempsAndPreservesReduceSystemColumns|TestMarshalReduceResult_GroupBy|TestGoReduceGroupBy'
```

Result after change:

```text
ok  github.com/milvus-io/milvus/internal/querynodev2/tasks  0.628s
```

A scratch benchmark of the exact primitive, run back to back in one session with `-benchmem -count=5`, reproduced the benefit:

| helper | before | after |
|---|---:|---:|
| group-by column name | 87.09 ns/op, 16 B/op, 1 alloc/op | 35.99 ns/op, 3 B/op, 1 alloc/op |
| boost-score column name | 78.92 ns/op, 16 B/op, 1 alloc/op | 15.55 ns/op, 0 B/op, 0 alloc/op |

This is a small internal helper enhancement with no exported API changes and a two-file source blast radius.

PR #52343, opened 2026-08-09
enhance: avoid fmt for querynode column names

bigsheeper:
issue: #52342

### What changed

Use `strconv` instead of `fmt.Sprintf` for two internal querynodev2 temporary column-name helpers:

- `groupByColumnName(fieldID)`
- `boostScoreColumn(index)`

Both helpers format a constant prefix plus a decimal integer. This preserves the generated column names while avoiding the generic formatting path.

### Benchmark

Scratch benchmark of the exact primitive, measured back to back in one session with `-benchmem -count=5` on linux/amd64:

| helper | before | after |
|---|---:|---:|
| group-by column name | 87.09 ns/op, 16 B/op, 1 alloc/op | 35.99 ns/op, 3 B/op, 1 alloc/op |
| boost-score column name | 78.92 ns/op, 16 B/op, 1 alloc/op | 15.55 ns/op, 0 B/op, 0 alloc/op |

### Verification

```bash
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:$PKG_CONFIG_PATH \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib:$LD_LIBRARY_PATH \
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/tasks \
  -run 'TestBoostScoreColumn|TestApplyBoostScoresPrunesTempsAndPreservesReduceSystemColumns|TestMarshalReduceResult_GroupBy|TestGoReduceGroupBy'
```

```text
ok  github.com/milvus-io/milvus/internal/querynodev2/tasks  0.628s
```

```bash
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:$PKG_CONFIG_PATH \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib:$LD_LIBRARY_PATH \
GOWORK=off go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/querynodev2/tasks
```

```text
ok  github.com/milvus-io/milvus/internal/querynodev2/tasks  13.162s
```

```bash
PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig:/home/sheep/workspace/milvus/cmake_build/src:$PKG_CONFIG_PATH \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib:$LD_LIBRARY_PATH \
GOWORK=off make static-check
```

```text
0 issues.
```

Note: this worktree uses the ignored local generated `cmd/tools/migration/legacy/legacypb` package already present in the main checkout so repository-wide static-check can typecheck the migration tool.

### DIFF
--- a/internal/querynodev2/tasks/boost_score.go
+++ b/internal/querynodev2/tasks/boost_score.go
@@ -19,6 +19,7 @@ package tasks
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/apache/arrow/go/v17/arrow"
 	"google.golang.org/protobuf/proto"
@@ -38,7 +39,7 @@ const (
 )
 
 func boostScoreColumn(index int) string {
-	return fmt.Sprintf("%s%d", boostScoreColumnPrefix, index)
+	return boostScoreColumnPrefix + strconv.Itoa(index)
 }
 
 func extractPlanScorers(serializedPlan []byte) ([]*planpb.ScoreFunction, error) {
--- a/internal/querynodev2/tasks/group_by_columns.go
+++ b/internal/querynodev2/tasks/group_by_columns.go
@@ -1,7 +1,6 @@
 package tasks
 
 import (
-	"fmt"
 	"strconv"
 	"strings"
 
@@ -11,7 +10,7 @@ import (
 const groupByColumnPrefix = "$group_by_"
 
 func groupByColumnName(fieldID int64) string {
-	return fmt.Sprintf("%s%d", groupByColumnPrefix, fieldID)
+	return groupByColumnPrefix + strconv.FormatInt(fieldID, 10)
 }
 
 func isGroupByColumnName(name string) bool {


czs007:

**Verified:**

- `internal/querynodev2/tasks/boost_score.go:42` — `boostScoreColumnPrefix + strconv.Itoa(index)` is byte-identical to the previous `fmt.Sprintf("%s%d", prefix, index)`, since `%d` on an `int` emits base-10 with the same negative handling.
- `internal/querynodev2/tasks/group_by_columns.go:13` — `groupByColumnPrefix + strconv.FormatInt(fieldID, 10)` matches the previous `%d` on an `int64` across the full value range, including negatives.
- `internal/querynodev2/tasks/group_by_columns.go:24` — the parse side still uses `strconv.ParseInt(fieldID, 10, 64)`, so names generated by the new base-10 producer round-trip correctly with no format skew.
- `internal/util/function/chain/converter.go:431` and `internal/core/src/segcore/search_result_export_c.cpp:167` — the other Go and C++ producers emit the same "prefix + decimal ID" form, so cross-component naming stays consistent.
- `internal/querynodev2/tasks/boost_score.go:72` — `fmt` is still used on other error paths so its import is correctly retained, while `group_by_columns.go` drops `fmt` and reuses the `strconv` import it already had; no missing or unused imports.
- `internal/querynodev2/tasks/boost_score_test.go:45` — tests assert the exact generated column names, and `:226-306` plus `search_task_go_reduce_test.go:642` / `marshal_reduce_result_test.go:300` cover temp-column pruning, group-by system-column retention, and the export path.

Issue #47413, opened 2026-01-29
[Bug]: [Entity ttl] count(*) remain the same after ttl deadline.


FeilongHou:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

count(\*) remain the insert count after time passes ttl time. Query result shows count(\*) != 0 after data expired. 

### Expected Behavior

count(*) = 0

### Steps To Reproduce

```markdown
import numpy as np
import random
import threading
import time
from datetime import datetime, timedelta, timezone
from collections import deque
from pymilvus import MilvusClient
from common import common_func as cf
from pymilvus import DataType
from utils.util_pymilvus import *

from pymilvus import __version__
print(__version__)

dim = 768
default_int32_value = np.int32(100)
default_primary_key_field_name = "int64"
default_vector_field_name = "vector"
default_int32_field_name = "int32"
default_json_field_name = "json"
default_timestamp_field_name = "timestamp"

client = MilvusClient(
    uri="http://<proxy IP>:19530",
    token="root:Milvus"
)

collection_name = "test_ttl_test"
if client.has_collection(collection_name):
    client.drop_collection(collection_name)

nb = 100
ttl_seconds = 8

# Create schema
schema = client.create_schema(enable_dynamic_field=False)
schema.add_field("id", DataType.INT64, is_primary=True, auto_id=False)
schema.add_field("ttl", DataType.TIMESTAMPTZ, nullable=True)
schema.add_field("vector", DataType.FLOAT_VECTOR, dim=default_dim)

# Create collection with ttl_field
properties = {"ttl_field": "ttl", "timezone": "UTC"}
client.create_collection(collection_name, schema=schema, properties=properties, consistency_level="Strong")

# Create index and load
index_params = client.prepare_index_params()
index_params.add_index(field_name="vector", index_type="IVF_FLAT", metric_type="L2", nlist=128)
client.create_index(collection_name, index_params=index_params)
client.load_collection(collection_name)

# Insert data with future ttl
ttl_timestamp = datetime.now(timezone.utc) + timedelta(seconds=ttl_seconds)
ttl_str = ttl_timestamp.isoformat()

[...]

PR #52344, opened 2026-08-09
fix: pin proxy read metadata snapshot per request

congqixia:
Related to #52351

Resolve aliases once for Search, Query, and HybridSearch requests. Reuse the pinned collection metadata, partition routing, and timestamps across RBAC, rate limiting, requery, and internal retries.

Fetch partition metadata by collection ID to prevent concurrent AlterAlias operations from mixing collection IDs and schemas.

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -564,6 +564,33 @@ func checkAuthorizationHelper(ctx context.Context, c *gin.Context, req interface
 	return nil
 }
 
+func prepareReadRequestSnapshot(ctx context.Context, c *gin.Context, req any, checkAuth bool) (context.Context, error) {
+	if checkAuth {
+		if err := checkAuthorizationV2(ctx, c, false, req); err != nil {
+			return ctx, err
+		}
+		ctx = c.Request.Context()
+	}
+
+	ctx, err, _ := proxy.EnsureReadRequestSnapshotForRequest(ctx, req)
+	if err != nil {
+		status := merr.Status(err)
+		dbGetter, hasDB := req.(requestutil.DBNameGetter)
+		collectionGetter, hasCollection := req.(requestutil.CollectionNameGetter)
+		if hasDB && hasCollection {
+			status = proxy.CollectionLookupErrorStatus(err, dbGetter.GetDbName(), collectionGetter.GetCollectionName())
+		}
+		responseErr := merr.Error(status)
+		HTTPAbortReturn(c, http.StatusOK, gin.H{
+			HTTPReturnCode:    status.GetCode(),
+			HTTPReturnMessage: status.GetReason(),
+		})
+		return ctx, responseErr
+	}
+	c.Request = c.Request.WithContext(ctx)
+	return ctx, nil
+}
+
 func wrapperProxy(ctx context.Context, c *gin.Context, req any, checkAuth bool, ignoreErr bool, fullMethod string, handler func(reqCtx context.Context, req any) (any, error)) (interface{}, error) {
 	return wrapperProxyWithLimit(ctx, c, req, checkAuth, ignoreErr, fullMethod, false, nil, handler)
 }
@@ -580,7 +607,11 @@ func wrapperProxyWithLimit(ctx context.Context, ginCtx *gin.Context, req any, ch
 		}
 		ctx = ginCtx.Request.Context()
 	}
-	if checkLimit {
+	ctx, snapshotErr, isReadRequest := proxy.EnsureReadRequestSnapshotForRequest(ctx, req)
+	if isReadRequest {
+		ginCtx.Request = ginCtx.Request.WithContext(ctx)
+	}
+	if checkLimit && snapshotErr == nil {
 		_, err := CheckLimiter(ctx, req, pxy)
 		if err != nil {
 			mlog.Warn(context.TODO(), "high level restful api, fail to check limiter", mlog.Err(err), mlog.String("method", fullMethod))
@@ -1399,6 +1430,10 @@ func (h *HandlersV2) query(ctx context.Context, c *gin.Context, anyReq any, dbNa
 		QueryParams:    []*commonpb.KeyValuePair{},
 	}
 	var err error
+	ctx, err = prepareReadRequestSnapshot(ctx, c, req, h.checkAuth)
+	if err != nil {
+		return nil, err
+	}
 	collSchema, err := h.GetCollectionSchema(ctx, c, dbName, httpReq.CollectionName)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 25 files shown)


internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -564,6 +564,22 @@ func checkAuthorizationHelper(ctx context.Context, c *gin.Context, req interface
...
+	if checkAuth {
+		if err := checkAuthorizationV2(ctx, c, false, req); err != nil {
+			return ctx, err
+		}
+		ctx = c.Request.Context()
+	}
+
+	ctx, err, _ := proxy.EnsureReadRequestSnapshotForRequest(ctx, req)
+	if err != nil {
+		return ctx, err
```
liliu-z: The snapshot-resolution branch of prepareReadRequestSnapshot returns its error without writing anything to the response writer, so gin falls through to a default 200 with a zero-length body. Every caller (query 1422, get 1501, search 1892, advancedSearch 2168) just does `return nil, err`, and wrapperPost only records metrics — it never writes a response. The auth branch above is fine because checkAuthorizationV2 already calls HTTPReturn; only this branch is missing it. The most common REST v2 failure (collection/alias not found, which GetCollectionInfo raises) now looks like a successful empty result. Write HTTPAbortReturn with merr.Code(err) before returning, matching the contract the sibling paths rely on (see the comment at handler_v2.go:1900, "has already throw http in GetCollectionSchema if fails").

internal/proxy/task_search.go
```diff
@@ -283,41 +291,49 @@ func (t *searchTask) PreExecute(ctx context.Context) error {
...
+			}
+		}
+
+		// use collection schema updated timestamp if it's greater than calculate guarantee timestamp
+		// this make query view updated happens before new read request happens
+		// see also schema change design
+		if collectionInfo.updateTimestamp > guaranteeTs {
+			guaranteeTs = collectionInfo.updateTimestamp
 		}
+		if t.isIterator && t.request.GetGuaranteeTimestamp() > 0 {
```
liliu-z: The iterator guarantee-timestamp override moved above the EntityTtlPhysicalTime derivation, changing what entity TTL is evaluated against. Previously EntityTtlPhysicalTime came from the freshly computed guaranteeTs (BeginTs-derived, after the consistency and schema-update-barrier logic) and only t.GuaranteeTimestamp/t.MvccTimestamp were replaced by the client's guarantee timestamp; now guaranteeTs itself is overwritten first, so EntityTtlPhysicalTime is derived from the iterator's page-1 session timestamp. That is exactly the MVCC-derived clock that Expr.cpp's CreateTTLFieldFilterExpression comment says must not drive TTL expiry (issue #47413), so on a collection with a TTL field a long-running iterator keeps serving rows that expired mid-iteration. Derive EntityTtlPhysicalTime before applying the iterator override.

internal/proxy/task_query.go
```diff
@@ -848,60 +846,60 @@ func (t *queryTask) PreExecute(ctx context.Context) error {
...
+			}
 		}
+
+		// use collection schema updated timestamp if it's greater than calculate guarantee timestamp
+		// this make query view updated happens before new read request happens
+		// see also schema change design
+		if colInfo.updateTimestamp > guaranteeTs {
+			guaranteeTs = colInfo.updateTimestamp
+		}
+		if t.queryParams.isIterator && t.request.GetGuaranteeTimestamp() > 0 {
```
liliu-z: Same reordering as the search path: the iterator guarantee-timestamp override now runs inside the timestamp-computation block, before EntityTtlPhysicalTime is derived from guaranteeTs. A query iterator on a collection with a TTL field therefore evaluates expiry against the page-1 session timestamp for the life of the iteration instead of a live physical clock, so entities that expire between pages stay visible. Move the TTL physical-time derivation ahead of the iterator override, or compute it from the pre-override guaranteeTs.

internal/proxy/read_snapshot.go
```diff
@@ -0,0 +1,301 @@
...
+
+type partitionInfosByIDCache interface {
+	GetPartitionInfosByID(ctx context.Context, database string, collectionID int64) (*partitionInfos, error)
+}
+
+func (s *readRequestSnapshot) Partitions(ctx context.Context) (*partitionInfos, error) {
+	if s == nil || s.collection == nil {
+		return nil, merr.WrapErrServiceInternalMsg("read request snapshot has no collection binding")
+	}
+	s.partitionOnce.Do(func() {
```
liliu-z: readRequestSnapshot.Partitions latches the partition-fetch result — including the error — in a sync.Once, so the first failed fetch is replayed for the rest of the request. The first consumer is the gRPC rate-limit interceptor, which explicitly swallows its own error and proceeds to the handler; a transient DescribeCollection/ShowPartitions failure there is then re-served to searchTask/queryTask PreExecute, which returns it to the user. Before this change getPartitionIDs issued an independent GetPartitions lookup, so the execution path was not bound to the interceptor's outcome. Only latch on success (or clear partitionErr so a failed fetch can be re-attempted within the request). Note this affects requests that name partitions or use a namespace — an empty partition list never triggers the fetch.

liliu-z (commented):
/lgtm
/approve

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -580,7 +600,11 @@ func wrapperProxyWithLimit(ctx context.Context, ginCtx *gin.Context, req any, ch
 		}
 		ctx = ginCtx.Request.Context()
 	}
-	if checkLimit {
+	ctx, snapshotErr, isReadRequest := proxy.EnsureReadRequestSnapshotForRequest(ctx, req)
```
liliu-z, line 610: The snapshot pinned here is not carried across the legacy-proxy forward hop: `forwardHandler` hands the original alias-bearing protobuf to `streaming.ForwardLegacyProxyUnaryServerInterceptor()`, and the 2.5.x proxy re-resolves the name on its own. So on a REST DQL request that gets forwarded (standalone mode during a 2.5→2.6 rolling upgrade), placeholder building and result decoding run against the schema this snapshot pinned while execution runs against whatever the legacy proxy resolves. The split predates this PR and gRPC is unaffected (the forward interceptor sits first in the chain, ahead of Database/Privilege/RateLimit), so the ask is only to narrow the design doc's REST V2 claim that "schema conversion, limiter accounting, and the Proxy method therefore use the same binding" — or to rewrite the forwarded request with the pinned canonical name.

internal/proxy/task_search.go
```diff
@@ -165,39 +177,35 @@ func (t *searchTask) PreExecute(ctx context.Context) error {
...
+			return err
+		}
+		t.readSnapshot = snapshot
+	}
+	if err := t.readSnapshot.validateTarget(t.request.GetDbName(), requestedCollectionName); err != nil {
 		return err
 	}
+	collectionSnapshot := t.readSnapshot.Collection()
+	collectionName := collectionSnapshot.CanonicalName()
+	t.collectionName = collectionName
```
liliu-z, line 192: `t.collectionName` now holds the canonical collection name instead of the user-supplied name, so for searches issued through an alias the `ProxySearchSparseNumNonZeros` label (task_search.go:715 and :1064) and `CollectionWorkLoad.CollectionName` (:1278, :1298) flip from the alias to the canonical name. Existing Prometheus series keyed on the alias stop receiving samples at upgrade and a new series appears, breaking dashboards/alerts that match on the alias. The returned `SearchResults.CollectionName` is unaffected (task_search.go:1420 restores the requested name, and task_query.go:1059 was updated for the same reason), so this is metrics-only — worth a release-note line rather than a code change.

internal/proxy/impl.go
```diff
@@ -2819,6 +2819,15 @@ func (node *Proxy) Search(ctx context.Context, request *milvuspb.SearchRequest)
 	rsp := &milvuspb.SearchResults{
 		Status: merr.Success(),
 	}
+	if err := merr.CheckHealthy(node.GetStateCode()); err != nil {
```
liliu-z, line 2838: The new early returns in `Search` and `HybridSearch` bypass the per-request metrics that `node.search`/`node.hybridSearch` record before their own health check, so requests rejected for an unhealthy proxy or a failed snapshot resolution (e.g. collection not found) no longer increment `ProxyReceivedNQ` and no longer stamp database/collection labels on the RPC stats object. `Proxy.Query` was wired the other way — its metrics block runs before the equivalent early returns — so the three read entry points now disagree on what a rejected request records.

Issue #52325, opened 2026-08-08
[Bug]: REST v2 entities/search silently ignores strictGroupSize


yihui504:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: v3.0.0 (release 2026-07-29, commit f46a0328558be155d11266a1a2b90602ccc9b366)
- Deployment mode: standalone (Docker, `milvusdb/milvus:v3.0.0`)
- SDK version: REST API v2 (curl) + pymilvus 2.5.5 (gRPC comparison)
- OS: Windows 11 (Docker Desktop)
```

### Current Behavior

On v3.0.0, `POST /v2/vectordb/entities/search` silently ignores `groupParams.strictGroupSize`. The parameter is accepted (`code: 0`) but has no effect — results are identical whether `strictGroupSize` is `true` or `false`. The gRPC path honors the same parameter.

With 5 groups (`cat` 0–4) × 3 points each = 15 points:

| group_size | strictGroupSize | limit | REST returns | gRPC returns |
|---|---|---|---|---|
| 1 | true | 15 | 15 (should be 5) | 5 |
| 2 | true | 15 | 15 (should be 10) | 10 |
| 1 | true | 10 | 10 (should be 5) | 5 |
| 1 | false | 15 | 15 | 5 |

REST behaves as if `strictGroupSize` is always `false`, regardless of the value sent.

### Expected Behavior

Per the [REST v2 Search documentation](https://milvus.io/api-reference/restful/v3.0.x/v2/Vector%20(v2)/Search.md):

> **strictGroupSize** *(boolean)*: Whether to return only the top k entities for each group. This parameter is only valid when groupingField is specified.

REST should honor `strictGroupSize=true` and cap each group at `groupSize` entities, consistent with the gRPC path.

### Steps To Reproduce

```markdown
# 1. Create collection (pymilvus avoids REST schema dim-format edge cases)
python -c "
from pymilvus import MilvusClient, CollectionSchema, FieldSchema, DataType
mc = MilvusClient(uri='http://localhost:19530')
schema = CollectionSchema(fields=[
    FieldSchema('id', DataType.INT64, is_primary=True, auto_id=False),
    FieldSchema('vector', DataType.FLOAT_VECTOR, dim=4),
    FieldSchema('cat', DataType.INT64),
])
mc.create_collection(collection_name='gs_demo', schema=schema)
idx = mc.prepare_index_params()
[...]

Issue #52311, opened 2026-08-07
[Bug]: REST API v2 silently accepts `group_by_field` on vector fields (gRPC rejects "unsupported data type")


yihui504:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: v3.0.0 (release 2026-07-29, commit f46a0328558be155d11266a1a2b90602ccc9b366)
- Deployment mode: standalone (Docker, `milvusdb/milvus:v3.0.0`)
- MQ type: N/A
- SDK version: REST API v2 only (curl). pymilvus used for gRPC comparison.
- OS: Windows 11 (Docker Desktop)
- CPU/Memory: N/A
- GPU: None
```

### Current Behavior

The REST API v2 `POST /v2/vectordb/entities/search` endpoint silently accepts `groupByField` set to a **vector field** (e.g., `"vector"`). The request returns `code: 0` with normal search results — **grouping is silently ignored**, producing identical results to a search without `group_by`.

The gRPC path correctly rejects this:

```
code=2001, message="unsupported data type VECTOR_FLOAT for group by operator
at ../internal/core/src/exec/operator/search-groupby/SearchGroupByOperator.cpp:169"
```

The [Grouping Search documentation](https://milvus.io/docs/grouping-search.md) describes grouping by scalar fields (e.g., `docId`, `color`). Grouping by a vector field is semantically invalid — vectors cannot be compared for equality-based grouping.

### Expected Behavior

The REST API should reject `groupByField` set to a vector field with an explicit error (`code != 0`), consistent with gRPC behavior. At minimum, it should not silently ignore the parameter (users may believe grouping is working when it isn't).

### Steps To Reproduce

```markdown
# 1. Create + load collection
curl -X POST http://localhost:19530/v2/vectordb/collections/create \
  -H "Content-Type: application/json" \
  -d '{"collectionName":"test_gv","dimension":4}'

for i in $(seq 0 9); do
  curl -X POST http://localhost:19530/v2/vectordb/entities/insert \
    -H "Content-Type: application/json" \
    -d "{\"collectionName\":\"test_gv\",\"data\":[{\"id\":$i,\"vector\":[0.$i,0.$i,0.$i,0.$i]}]}"
done

# 2. Search with groupByField="vector" (the vector field)
curl -X POST http://localhost:19530/v2/vectordb/entities/search \
  -H "Content-Type: application/json" \
[...]

Issue #52309, opened 2026-08-07
[Bug]: REST API v2 `entities/search` accepts `group_size=0` and `-1` (gRPC rejects as "negative")


yihui504:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: v3.0.0 (release 2026-07-29, commit f46a0328558be155d11266a1a2b90602ccc9b366)
- Deployment mode: standalone (Docker, `milvusdb/milvus:v3.0.0`)
- MQ type: N/A
- SDK version: REST API v2 only (curl). pymilvus used for gRPC comparison.
- OS: Windows 11 (Docker Desktop)
- CPU/Memory: N/A
- GPU: None
```

### Current Behavior

The REST API v2 `POST /v2/vectordb/entities/search` endpoint accepts `groupSize=0` and `groupSize=-1` in `groupParams` without validation (`code: 0`). The gRPC path correctly rejects both:

```
gRPC: code=1100, message="input group size:0 is negative, failed to do search_groupby"
gRPC: code=1100, message="input group size:-1 is negative, failed to do search_groupby"
```

The [Grouping Search documentation](https://milvus.io/docs/grouping-search.md) states:

> "By default, Grouping Search returns only one entity per group. If you want to increase the number of results to return per group, you can control this with the `group_size` and `strict_group_size` parameters."

This implies `group_size` is a positive integer that "increases" results per group — `0` and `-1` are semantically meaningless.

### Expected Behavior

`groupSize` should be validated as ≥ 1. REST should reject `0` and negative values with an explicit error, consistent with gRPC.

### Steps To Reproduce

```markdown
# 1. Create + load collection
curl -X POST http://localhost:19530/v2/vectordb/collections/create \
  -H "Content-Type: application/json" \
  -d '{"collectionName":"test_gs","dimension":4}'

# Insert data with "cat" field
for i in $(seq 0 9); do
  curl -X POST http://localhost:19530/v2/vectordb/entities/insert \
    -H "Content-Type: application/json" \
    -d "{\"collectionName\":\"test_gs\",\"data\":[{\"id\":$i,\"vector\":[0.$i,0.$i,0.$i,0.$i],\"cat\":$((i%3))}]}"
done

# 2. Search with groupSize=0 (bug)
curl -X POST http://localhost:19530/v2/vectordb/entities/search \
  -H "Content-Type: application/json" \
[...]

PR #52346, opened 2026-08-09
fix: stop the REST group-by knobs from being silently swallowed

xiaofan-luan:
> Kept in lockstep with the squashed commit message.

Three REST requests were accepted and then quietly did something other than
what they said, all through the same pattern: a parameter that fails a check
is not forwarded rather than refused, so the server that would have rejected
it never sees it and the request reports success.

groupSize was forwarded only when greater than zero, behind a bare int32 that
cannot tell an absent knob from an explicit one. groupSize 0 or -1 was
swallowed and the search ran with the server default; the server-side
rejection ("input group size is negative") exists but was unreachable from
REST. strictGroupSize rode behind the same gate, so setting it without also
setting groupSize dropped it silently. Both knobs are now pointers: absent
means the server default, present means forwarded or refused -- an explicit
non-positive groupSize and any knob without groupingField are errors. The
searchAggregation conflict check asks for presence rather than for the zero
value, so an explicit groupSize: 0 no longer slips past it either.

A group-by field name resolved to a field id with no look at the field type,
so grouping by a vector field was accepted at the proxy and either degenerated
into an ungrouped search or surfaced as an execution error deep in the query.
parseGroupByField now refuses the types the group-by operator cannot hold as
a key; the supported list mirrors the switch in SearchGroupByOperator.cpp --
bool, the integer widths, timestamptz, varchar, and json, whose inner value
is still checked at execution where the path is known. The dynamic-field
fallback is untouched.

The searchParams map is a second spelling of the same knobs, and it was a side
door around both rules above: group_size and strict_group_size inside
searchParams slipped past the searchAggregation conflict check, and when the
two spellings disagreed the searchParams one was appended first and silently
won, since the proxy takes the first matching key. Both ambiguities are
refused now, the same way the order_by_fields passthrough already was: one
spelling per request.

Two edges found in review are folded in. A groupingField of blanks counted as
provided at the REST layer while the proxy trims the name before resolving it,
so the knobs it authorized were swallowed downstream -- the name is now
trimmed with the proxy's rule before anything is decided. And DataType_String
is off the accepted list: the executor's switch has a VARCHAR case and no
STRING case, so admitting it would only move the failure back into the
execution layer.

The legacy spelling obeys the same rules as the dedicated one. A size knob in
searchParams with no group field anywhere is refused rather than parsed into a
grouping that never enables. And group keys nested under searchParams.params
are refused only when they are the request's sole grouping: the proxy reads
standalone pairs, so a nested-only spelling asked for grouping and silently
did not get it. Beside a working spelling the nested copies are tolerated --
generateSearchParams itself duplicates root keys into the params JSON, so
that shape is the pipeline's own artifact and previously-working requests
keep working. The ambiguity rule likewise asks only the root keys. The hybrid-search test fixtures gain
the field types the type gate now demands -- all of them, not only the two
that failed first. The wording the caller reads is preserved rather than
replaced wherever a check moved earlier: a binary vector keeps the executor's
own sentence, and the searchAggregation conflict keeps its group_by_field(s)
sentence with the size knobs appended to it. Moving a check earlier should not
change the answer a client is pinned on -- the go client suite and the REST
e2e suite both pin these.

Two proxy-side gaps in the same territory are closed with the same fail-fast
principle. At most one JSON field can be grouped by: the plan carries a single
json_path/json_type and the executor asserts as much, but two bare JSON fields
passed the proxy and failed only deep in execution. And the JSON group-by
attributes now ride the rank params into hybrid search: parseRankParams
already resolved jsonPath, jsonType and strictCast and then dropped them, so
a hybrid group-by on a JSON key silently grouped by the whole field instead
of the key the caller named.

The unknown-request-key question raised alongside these (a nested SDK-style
groupParams object is silently ignored by the binder) is a policy decision
about every REST endpoint, not a bug in these knobs, and is left alone.

---

issue: #52309
issue: #52311
issue: #52325

https://claude.ai/code/session_01D1pij3TGike2nZy7t6oDZY

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2.go
+++ b/internal/distributed/proxy/httpserver/handler_v2.go
@@ -1861,6 +1861,42 @@ func generatePlaceholderGroup(ctx context.Context, body string, collSchema *sche
 	})
 }
 
+// appendGroupParams validates the grouping knobs and appends them to params.
+//
+// The previous gate forwarded groupSize only when it was greater than zero,
+// which conflated three different requests: an absent groupSize (use the
+// server default), an explicit invalid one (0 or negative -- the server
+// rejects these, but it never saw them), and strictGroupSize on its own
+// (silently dropped because it rode behind the same gate). A knob that is
+// present is now either forwarded or refused; only absence means default.
+func appendGroupParams(params []*commonpb.KeyValuePair, groupByField string, groupSize *int32, strictGroupSize *bool) ([]*commonpb.KeyValuePair, error) {
+	// Trimmed with the same rule the proxy applies when it parses the field
+	// name back out. Without this, a groupingField of blanks counts as
+	// provided here, the proxy trims it to nothing, and the knobs it
+	// authorized are swallowed downstream -- the exact silence this function
+	// exists to remove.
+	groupByField = strings.TrimSpace(groupByField)
+	if groupByField == "" {
+		if groupSize != nil || strictGroupSize != nil {
+			return nil, merr.WrapErrParameterInvalidMsg(
+				"groupSize and strictGroupSize require groupingField")
+		}
+		return params, nil
+	}
+	params = append(params, &commonpb.KeyValuePair{Key: ParamGroupByField, Value: groupByField})
+	if groupSize != nil {
+		if *groupSize <= 0 {
+			return nil, merr.WrapErrParameterInvalidMsg(
+				"groupSize must be a positive integer, got: %d", *groupSize)
+		}
+		params = append(params, &commonpb.KeyValuePair{Key: ParamGroupSize, Value: strconv.FormatInt(int64(*groupSize), 10)})
+	}
+	if strictGroupSize != nil {
+		params = append(params, &commonpb.KeyValuePair{Key: ParamStrictGroupSize, Value: strconv.FormatBool(*strictGroupSize)})
+	}
+	return params, nil
+}
+
 func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbName string) (interface{}, error) {
 	httpReq := anyReq.(*SearchReqV2)
 	req := &milvuspb.SearchRequest{
@@ -1927,13 +1963,16 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
-		if httpReq.GroupByField != "" || httpReq.GroupSize != 0 || httpReq.StrictGroupSize {
+		if httpReq.GroupByField != "" || httpReq.GroupSize != nil || httpReq.StrictGroupSize != nil {
 			err := merr.WrapErrParameterInvalidMsg("groupingField/groupSize/strictGroupSize and searchAggregation cannot be used simultaneously")
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
-		if searchParamsContainAny(httpReq.SearchParams, proxy.GroupByFieldKey, proxy.GroupByFieldsKey) {
-			err := merr.WrapErrParameterInvalidMsg("searchParams.group_by_field(s) and searchAggregation cannot be used simultaneously")
+		if searchParamsContainAny(httpReq.SearchParams,
+			proxy.GroupByFieldKey, proxy.GroupByFieldsKey, proxy.GroupSizeKey, proxy.StrictGroupSize) {
+			// wording kept compatible: the group_by_field(s) sentence is what
+			// clients and the REST e2e suite already pin on
+			err := merr.WrapErrParameterInvalidMsg("searchParams.group_by_field(s) and searchAggregation cannot be used simultaneously, and neither can group_size/strict_group_size")
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
@@ -1958,6 +1997,54 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 		return nil, err
 	}
 
+	// The group knobs have the same two spellings, and the searchParams one is
+	// appended first, so it silently won any disagreement. Same rule: one
+	// spelling per request. Asked of the root keys only: generateSearchParams
+	// itself duplicates root keys into the params JSON, so nested copies are
+	// an artifact of the working spelling, not a spelling of their own.
+	if (httpReq.GroupByField != "" || httpReq.GroupSize != nil || httpReq.StrictGroupSize != nil) &&
+		searchParamsRootContainAny(httpReq.SearchParams,
+			proxy.GroupByFieldKey, proxy.GroupByFieldsKey, proxy.GroupSizeKey, proxy.StrictGroupSize) {
+		err := merr.WrapErrParameterInvalidMsg(
+			"ambiguous grouping: use either the top-level groupingField/groupSize/strictGroupSize or the searchParams group keys, not both")
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
+	}
+
+	// The rule that a size knob needs a grouping field holds for the legacy
+	// spelling too: group_size or strict_group_size in searchParams with no
+	// group field anywhere used to parse downstream into a grouping that never
+	// enabled, an ungrouped search reporting success. The field must be a
+	// working spelling -- a field name nested under params is inert.
+	groupingEnabled := httpReq.GroupByField != "" ||
+		searchParamsRootContainAny(httpReq.SearchParams, proxy.GroupByFieldKey, proxy.GroupByFieldsKey)
+	if !groupingEnabled &&
+		searchParamsRootContainAny(httpReq.SearchParams, proxy.GroupSizeKey, proxy.StrictGroupSize) {
+		err := merr.WrapErrParameterInvalidMsg(
+			"searchParams group_size/strict_group_size require a group field")
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
+	}
+
+	// Group keys nested under searchParams.params never act -- the proxy reads
+	// only standalone pairs. When a working spelling enables grouping they are
+	// tolerated as the compatibility duplicates generateSearchParams itself
+	// produces; when they are the only grouping in the request, the caller
+	// asked for grouping and would silently not get it, so the request is
+	// refused rather than half-honored.
+	if !groupingEnabled {
+		if nested, ok := httpReq.SearchParams[Params].(map[string]interface{}); ok {
+			for _, key := range []string{proxy.GroupByFieldKey, proxy.GroupByFieldsKey, proxy.GroupSizeKey, proxy.StrictGroupSize} {
+				if _, present := nested[key]; present {
+					err := merr.WrapErrParameterInvalidMsg(
+						"searchParams.params cannot carry %s; use the top-level grouping fields or searchParams.%s", key, key)
+					HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+					return nil, err
+				}
+			}
+		}
+	}
+
 	searchParams, err := generateSearchParams(httpReq.SearchParams)
 	if err != nil {
 		mlog.Warn(ctx, "high level restful api, generate SearchParams failed", mlog.Err(err))
@@ -1969,12 +2056,10 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 	}
 	searchParams = append(searchParams, &commonpb.KeyValuePair{Key: common.TopKKey, Value: strconv.FormatInt(int64(httpReq.Limit), 10)})
 	searchParams = append(searchParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
-	if httpReq.GroupByField != "" {
-		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamGroupByField, Value: httpReq.GroupByField})
-	}
-	if httpReq.GroupByField != "" && httpReq.GroupSize > 0 {
-		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamGroupSize, Value: strconv.FormatInt(int64(httpReq.GroupSize), 10)})
-		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamStrictGroupSize, Value: strconv.FormatBool(httpReq.StrictGroupSize)})
+	searchParams, err = appendGroupParams(searchParams, httpReq.GroupByField, httpReq.GroupSize, httpReq.StrictGroupSize)
+	if err != nil {
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
 	}
 	if len(httpReq.OrderByFields) > 0 {
 		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: proxy.OrderByFieldsKey, Value: strings.Join(httpReq.OrderByFields, ",")})
@@ -2245,12 +2330,10 @@ func (h *HandlersV2) advancedSearch(ctx context.Context, c *gin.Context, anyReq 
 		{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)},
 		{Key: ParamRoundDecimal, Value: "-1"},
 	}
-	if httpReq.GroupByField != "" {
-		req.RankParams = append(req.RankParams, &commonpb.KeyValuePair{Key: ParamGroupByField, Value: httpReq.GroupByField})
-	}
-	if httpReq.GroupByField != "" && httpReq.GroupSize > 0 {
-		req.RankParams = append(req.RankParams, &commonpb.KeyValuePair{Key: ParamGroupSize, Value: strconv.FormatInt(int64(httpReq.GroupSize), 10)})
-		req.RankParams = append(req.RankParams, &commonpb.KeyValuePair{Key: ParamStrictGroupSize, Value: strconv.FormatBool(httpReq.StrictGroupSize)})
+	req.RankParams, err = appendGroupParams(req.RankParams, httpReq.GroupByField, httpReq.GroupSize, httpReq.StrictGroupSize)
+	if err != nil {
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
 	}
 	if len(httpReq.FunctionScore.Functions) != 0 {
 		if req.FunctionScore, err = genFunctionScore(ctx, &httpReq.FunctionScore); err != nil {
--- a/internal/proxy/search_util.go
+++ b/internal/proxy/search_util.go
@@ -31,6 +31,34 @@ type rankParams struct {
 	groupByFieldNames []string
 	groupSize         int64
 	strictGroupSize   bool
+	// The JSON group-by attributes ride along for hybrid search: they are
+	// resolved by parseGroupByInfo like everything above, and dropping them
+	// here used to make a hybrid group-by on a dynamic field group by the
+	// whole $meta object instead of the named key.
+	jsonPath   string
+	jsonType   schemapb.DataType
+	strictCast bool
+}
+
+func (r *rankParams) GetJSONPath() string {
+	if r != nil {
+		return r.jsonPath
+	}
+	return ""
+}
+
+func (r *rankParams) GetJSONType() schemapb.DataType {
+	if r != nil {
+		return r.jsonType
+	}
+	return schemapb.DataType_None
+}
+
+func (r *rankParams) GetStrictCast() bool {
+	if r != nil {
+		return r.strictCast
+	}
+	return false
 }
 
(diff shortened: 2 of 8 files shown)


internal/proxy/search_util.go
```diff
@@ -814,6 +814,31 @@ func (g *groupByInfo) GetStrictCast() bool {
...
+// query, depending on the path. The supported list mirrors the switch in
+// SearchGroupByOperator.cpp: bool, the integer widths, timestamptz, varchar,
+// and json, whose inner value is checked at execution where the path is known.
+func checkGroupByFieldType(field *schemapb.FieldSchema) error {
+	switch field.GetDataType() {
+	case schemapb.DataType_Bool,
+		schemapb.DataType_Int8, schemapb.DataType_Int16,
+		schemapb.DataType_Int32, schemapb.DataType_Int64,
+		schemapb.DataType_Timestamptz,
+		schemapb.DataType_VarChar, schemapb.DataType_String,
```
xiaofan-luan: **[P2] Keep the proxy whitelist aligned with the executor**

This admits `schemapb.DataType_String`, but `MultiFieldDataGetter` in `SearchGroupByOperator.cpp` handles only `DataType::VARCHAR` and falls into `Unsupported` for `DataType::STRING`. `FieldMeta::ParseFrom` preserves the schema enum rather than canonicalizing `STRING` to `VARCHAR`, so a legacy string field still passes this proxy check and fails deep in execution—the exact behavior this helper is intended to prevent. Either remove `DataType_String` from this whitelist or add matching C++ support, and cover the chosen behavior with a test.

xiaofan-luan (commented):
Two P2 validation gaps remain: the existing whitespace-only `groupingField` path still swallows group knobs, and the new type whitelist admits legacy `STRING` even though the C++ group-by operator rejects it. Please address both before merge.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1907,7 +1943,7 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
-		if httpReq.GroupByField != "" || httpReq.GroupSize != 0 || httpReq.StrictGroupSize {
+		if httpReq.GroupByField != "" || httpReq.GroupSize != nil || httpReq.StrictGroupSize != nil {
```
xiaofan-luan, line 1966: **[P2] Reject all legacy group knobs with search aggregation**

The adjacent `searchParamsContainAny` conflict check covers only `group_by_field(s)`. A REST request can still send `searchAggregation` together with `searchParams.group_size` or `searchParams.strict_group_size`; `generateSearchParams` exposes those as normal search-param KVs, `initSearchAggregation` also checks only the field keys, and the aggregation path later overwrites `QueryInfo.GroupSize`/`StrictGroupSize`. The request therefore reports success while silently ignoring the supplied knobs. Include `proxy.GroupSizeKey` and `proxy.StrictGroupSize` in the legacy presence check (including the nested `params` form already handled by `searchParamsContainAny`).

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1949,12 +1985,10 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 	}
 	searchParams = append(searchParams, &commonpb.KeyValuePair{Key: common.TopKKey, Value: strconv.FormatInt(int64(httpReq.Limit), 10)})
 	searchParams = append(searchParams, &commonpb.KeyValuePair{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)})
-	if httpReq.GroupByField != "" {
-		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamGroupByField, Value: httpReq.GroupByField})
-	}
-	if httpReq.GroupByField != "" && httpReq.GroupSize > 0 {
-		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamGroupSize, Value: strconv.FormatInt(int64(httpReq.GroupSize), 10)})
-		searchParams = append(searchParams, &commonpb.KeyValuePair{Key: ParamStrictGroupSize, Value: strconv.FormatBool(httpReq.StrictGroupSize)})
+	searchParams, err = appendGroupParams(searchParams, httpReq.GroupByField, httpReq.GroupSize, httpReq.StrictGroupSize)
```
xiaofan-luan, line 2059: **[P2] Reject collisions between dedicated and legacy group parameters**

`generateSearchParams` places legacy `searchParams.group_by_field`, `group_size`, and `strict_group_size` before the dedicated values appended here. `funcutil.GetAttrByKeyFromRepeatedKV` returns the first matching key, so when both forms are supplied the legacy value silently wins and the top-level `groupingField`/`groupSize`/`strictGroupSize` is ignored. That recreates the exact “accepted but did something else” behavior this PR fixes. Add presence-based ambiguity checks, analogous to the existing `orderByFields` check, whenever the corresponding dedicated field is present.

xiaofan-luan (commented):
The two previous findings are fixed, but two remaining parameter-collision paths can still accept group knobs and silently execute different semantics. Both are the same failure class this PR is intended to eliminate.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1841,6 +1841,36 @@ func generatePlaceholderGroup(ctx context.Context, body string, collSchema *sche
...
+//
+// The previous gate forwarded groupSize only when it was greater than zero,
+// which conflated three different requests: an absent groupSize (use the
+// server default), an explicit invalid one (0 or negative -- the server
+// rejects these, but it never saw them), and strictGroupSize on its own
+// (silently dropped because it rode behind the same gate). A knob that is
+// present is now either forwarded or refused; only absence means default.
+func appendGroupParams(params []*commonpb.KeyValuePair, groupByField string, groupSize *int32, strictGroupSize *bool) ([]*commonpb.KeyValuePair, error) {
+	if groupByField == "" {
+		if groupSize != nil || strictGroupSize != nil {
```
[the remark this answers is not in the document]
xiaofan-luan: Fixed on the current head: `appendGroupParams` trims the name with the proxy's own rule before deciding anything, so a groupingField of blanks counts as absent — knobs alongside it are refused rather than swallowed downstream. Covered by `TestAppendGroupParamsValidatesTheKnobs/a_grouping_field_of_blanks_is_absent`.

internal/proxy/search_util.go
xiaofan-luan: Fixed: `DataType_String` is off the list, with a comment stating why (the executor's switch has a VARCHAR case and no STRING case). The test now asserts a String-typed field is refused.

internal/distributed/proxy/httpserver/handler_v2.go
xiaofan-luan: Fixed: the searchAggregation conflict check now covers `group_size` and `strict_group_size` in searchParams alongside `group_by_field(s)`. Covered by `TestGroupKnobSpellingsCannotDisagree/searchParams_group_size_cannot_ride_under_searchAggregation`.

internal/distributed/proxy/httpserver/handler_v2.go
xiaofan-luan: Fixed: when both spellings are present the request is refused as ambiguous ("one spelling per request"), following the existing `order_by_fields` precedent, instead of letting the first-appended searchParams key silently win. Covered by `TestGroupKnobSpellingsCannotDisagree/both_spellings_at_once_is_ambiguous`.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1938,6 +1975,18 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 		return nil, err
 	}
 
+	// The group knobs have the same two spellings, and the searchParams one is
+	// appended first, so it silently won any disagreement. Same rule: one
+	// spelling per request.
+	if (httpReq.GroupByField != "" || httpReq.GroupSize != nil || httpReq.StrictGroupSize != nil) &&
```
xiaofan-luan, line 2005: **[P2] Validate searchParams-only knobs require a group field**

This condition rejects legacy group keys only when a top-level spelling is also present. A request containing only `searchParams: {"group_size": 5}` or `{"strict_group_size": true}` and no `group_by_field(s)` still passes this handler. Downstream, `parseGroupByInfo` parses the size/strictness but returns no group field IDs; `DetermineSearchType` and the C++ search path decide whether grouping is enabled from those IDs, so the request succeeds as an ungrouped search and the knob is silently ignored. That contradicts the PR's stated rule that any group knob without `groupingField` is an error. Validate this invariant for the searchParams spelling as well (including nested `params`) and add a searchParams-only regression case.

xiaofan-luan (commented):
One remaining legacy-only path still reproduces the silent-swallow behavior: group size/strictness can be supplied through `searchParams` without any group-by field.

internal/proxy/search_util.go
```diff
@@ -814,6 +814,34 @@ func (g *groupByInfo) GetStrictCast() bool {
...
+		schemapb.DataType_Int32, schemapb.DataType_Int64,
+		schemapb.DataType_Timestamptz,
+		schemapb.DataType_VarChar,
+		schemapb.DataType_JSON:
+		// DataType_String is deliberately absent: the executor's switch has a
+		// VARCHAR case and no STRING case, so admitting it here would only
+		// move the failure back into the execution layer.
+		return nil
+	default:
+		return merr.WrapErrParameterInvalidMsg(
```
MrPresent-Han, line 874: **[P2] Please update the existing hybrid-search test fixture for this type gate**

Could we update `TestSearchTask_parseSearchInfo/parseSearchInfo groupBy info for hybrid search` together with this validation? Its `c1` and `c3` fields omit `DataType`, so they now resolve to `None`; `parseRankParams`/`parseSearchInfo` return errors and the test subsequently panics on the nil result. Setting those fixtures to a supported scalar type would preserve the production validation while restoring the related UT.

MrPresent-Han (commented):
One non-blocking UT follow-up from the focused group-by checks.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1907,13 +1943,14 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
...
 		}
-		if httpReq.GroupByField != "" || httpReq.GroupSize != 0 || httpReq.StrictGroupSize {
+		if httpReq.GroupByField != "" || httpReq.GroupSize != nil || httpReq.StrictGroupSize != nil {
 			err := merr.WrapErrParameterInvalidMsg("groupingField/groupSize/strictGroupSize and searchAggregation cannot be used simultaneously")
 			HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
 			return nil, err
 		}
-		if searchParamsContainAny(httpReq.SearchParams, proxy.GroupByFieldKey, proxy.GroupByFieldsKey) {
-			err := merr.WrapErrParameterInvalidMsg("searchParams.group_by_field(s) and searchAggregation cannot be used simultaneously")
+		if searchParamsContainAny(httpReq.SearchParams,
```
MrPresent-Han, line 1971: **[P2][Pre-existing; not introduced by this PR] Reject or normalize nested group keys**

This is an existing issue rather than a regression from this change: `searchParamsContainAny` recognizes group keys under `searchParams.params`, but `generateSearchParams` keeps them only inside the generic `params` JSON KV, while `parseGroupByInfo` reads only standalone KVs. A request such as `{"searchParams":{"params":{"group_by_field":"cat","group_size":2}}}` can therefore succeed without enabling grouping. Could we either reject nested group keys consistently or normalize them into standalone KVs, with a no-aggregation regression case?

MrPresent-Han (commented):
One non-blocking pre-existing REST group-parameter issue.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -2207,12 +2254,10 @@ func (h *HandlersV2) advancedSearch(ctx context.Context, c *gin.Context, anyReq
 		{Key: proxy.OffsetKey, Value: strconv.FormatInt(int64(httpReq.Offset), 10)},
 		{Key: ParamRoundDecimal, Value: "-1"},
 	}
-	if httpReq.GroupByField != "" {
-		req.RankParams = append(req.RankParams, &commonpb.KeyValuePair{Key: ParamGroupByField, Value: httpReq.GroupByField})
-	}
-	if httpReq.GroupByField != "" && httpReq.GroupSize > 0 {
-		req.RankParams = append(req.RankParams, &commonpb.KeyValuePair{Key: ParamGroupSize, Value: strconv.FormatInt(int64(httpReq.GroupSize), 10)})
-		req.RankParams = append(req.RankParams, &commonpb.KeyValuePair{Key: ParamStrictGroupSize, Value: strconv.FormatBool(httpReq.StrictGroupSize)})
+	req.RankParams, err = appendGroupParams(req.RankParams, httpReq.GroupByField, httpReq.GroupSize, httpReq.StrictGroupSize)
```
MrPresent-Han, line 2333: **[Non-blocking][Pre-existing; not a regression introduced by this PR] Preserve JSON group metadata in hybrid search**

The field name is forwarded correctly here, but the shared hybrid path later keeps only group field IDs/names/size in `rankParams` and drops the `jsonPath`, `jsonType`, and `strictCast` resolved by `parseGroupByInfo`. For a dynamic field such as `brand`, rows with the same brand but different sibling JSON properties can therefore be grouped by the whole `$meta` object instead of by `brand`. A follow-up could carry these attributes through `rankParams` into each hybrid `QueryInfo` and cover dynamic/JSON hybrid group-by.

MrPresent-Han (commented):
One non-blocking pre-existing hybrid group-by correctness note.

internal/proxy/search_util.go
```diff
@@ -814,6 +814,34 @@ func (g *groupByInfo) GetStrictCast() bool {
...
+// SearchGroupByOperator.cpp: bool, the integer widths, timestamptz, varchar,
+// and json, whose inner value is checked at execution where the path is known.
+func checkGroupByFieldType(field *schemapb.FieldSchema) error {
+	switch field.GetDataType() {
+	case schemapb.DataType_Bool,
+		schemapb.DataType_Int8, schemapb.DataType_Int16,
+		schemapb.DataType_Int32, schemapb.DataType_Int64,
+		schemapb.DataType_Timestamptz,
+		schemapb.DataType_VarChar,
+		schemapb.DataType_JSON:
```
MrPresent-Han, line 862: **[P2][Non-blocking][Pre-existing; not introduced by this PR] Reject multiple JSON group-by fields at the Proxy**

JSON is valid individually, but the existing multi-field parser also accepts bare `group_by_fields=json1,json2` because neither field produces a non-empty `jsonPath`. The plan carries only one `json_path/json_type`, and `ResolveGroupByArrowInfos` later asserts that at most one JSON group-by field is present, so this request fails only in the execution layer. Could we add a Proxy-side JSON-field count check and a focused regression test as a follow-up?

MrPresent-Han (commented):
One non-blocking pre-existing multi-JSON group-by validation note.

internal/distributed/proxy/httpserver/handler_v2.go
xiaofan-luan: Fixed: the size-needs-a-field rule now covers the legacy spelling — `searchParams.group_size`/`strict_group_size` (standalone or nested) with no group field anywhere is refused. Covered by `TestGroupKnobSpellingsCannotDisagree/searchParams_size_knobs_alone_still_need_a_group_field`.

internal/proxy/search_util.go
xiaofan-luan: Fixed: the fixture's c1/c2/c3 now carry VarChar/Int64/VarChar, so the production validation stands and the UT parses again.

internal/distributed/proxy/httpserver/handler_v2.go
xiaofan-luan: Fixed by refusing the nested form outright: group keys under `searchParams.params` have never reached the proxy (it reads only standalone pairs), so nothing can depend on them — the request now gets an explicit error naming the working spellings. Covered by `TestGroupKnobSpellingsCannotDisagree/group_keys_nested_under_params_are_inert_and_refused`.

internal/distributed/proxy/httpserver/handler_v2.go
xiaofan-luan: Agreed this is real and pre-existing — the hybrid path drops jsonPath/jsonType/strictCast from rankParams, so dynamic-field hybrid group-by groups by the whole $meta. Out of scope here (this PR is the REST/proxy validation layer); it needs the rank-params plumbing changed and deserves its own change with dynamic/JSON hybrid coverage. Left unresolved as the follow-up marker.

xiaofan-luan: Folded in after all — reading the code showed parseRankParams already resolves jsonPath/jsonType/strictCast via parseGroupByInfo and then drops them packing the struct. They now ride the rank params (with the same ParseAndVerifyNestedPath normalization the single-search path applies) and the isAdvanced branch reads them into QueryInfo. Covered by `TestRankParamsCarryJSONGroupAttributes`.

internal/proxy/search_util.go
xiaofan-luan: Folded in: parseGroupByInfo now counts JSON-typed group fields as the names resolve and refuses more than one, so the request fails at the proxy with a named field instead of deep in ResolveGroupByArrowInfos. JSON+scalar combinations stay legal. Covered by `TestParseGroupByInfoRejectsMultipleJSONFields`.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1938,6 +1975,49 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
 		return nil, err
 	}
 
+	// Group keys nested under searchParams.params can never take effect: the
+	// proxy reads only standalone pairs, so the keys rode along as inert text
+	// inside the params JSON while the search ran ungrouped and reported
+	// success. Refused rather than normalized -- the nesting has never worked,
+	// so nothing can depend on it.
+	if nested, ok := httpReq.SearchParams[Params].(map[string]interface{}); ok {
```
xiaofan-luan: **[P2] Preserve equal root-and-nested compatibility duplicates**

The claim that nothing can depend on a nested group key is not true when the same key is also present at the `searchParams` root. `generateSearchParams` deliberately allows equal root/`params` duplicates for compatibility (see its `deepEqual` path), emits the root key as a standalone KV, and grouping therefore worked for payloads such as `{"searchParams":{"group_by_field":"cat","params":{"group_by_field":"cat"}}}`. This blanket check now rejects that previously valid request before `generateSearchParams` can apply its compatibility rule. Reject nested-only keys, and let equal root+nested duplicates through (while the existing ambiguity check continues to reject differing values), or otherwise justify the breaking change and add coverage.

xiaofan-luan (commented):
The searchParams-only fix is present, but the new blanket nested-key rejection also rejects one form that previously worked through the standalone duplicate.

internal/distributed/proxy/httpserver/handler_v2.go
xiaofan-luan: Fixed: the nested rejection is now scoped to the case it exists for. Nested group keys are refused only when they are the request's sole grouping (the caller asked and would silently not get it); beside a working spelling — top-level groupingField or root-level searchParams group_by_field(s) — they are tolerated as the compatibility duplicates generateSearchParams itself produces, so the previously-working root+params shape keeps working. The ambiguity check likewise asks only the root keys now. Covered by the two new tolerated-shape cases in `TestGroupKnobSpellingsCannotDisagree`.

internal/distributed/proxy/httpserver/handler_v2.go
```diff
@@ -1938,6 +1975,54 @@ func (h *HandlersV2) search(ctx context.Context, c *gin.Context, anyReq any, dbN
...
+		HTTPAbortReturn(c, http.StatusOK, gin.H{HTTPReturnCode: merr.Code(err), HTTPReturnMessage: err.Error()})
+		return nil, err
+	}
+
+	// The rule that a size knob needs a grouping field holds for the legacy
+	// spelling too: group_size or strict_group_size in searchParams with no
+	// group field anywhere used to parse downstream into a grouping that never
+	// enabled, an ungrouped search reporting success. The field must be a
+	// working spelling -- a field name nested under params is inert.
+	groupingEnabled := httpReq.GroupByField != "" ||
```
xiaofan-luan, line 2019: [P2] Treat blank legacy group fields as absent

`groupingEnabled` checks only key presence. For example, `searchParams: {"group_by_field":" ","group_size":5}` (and likewise an all-blank `group_by_fields`) passes this gate. `parseGroupByInfo` later trims the field list to empty while retaining `group_size`, so the request succeeds without grouping. Normalize/validate the root field value with the same rule as `appendGroupParams` before calling it enabled, and cover this through the full REST handler.

xiaofan-luan (commented):
[P2] Reject malformed `strict_group_size` instead of disabling it.

The legacy REST spelling can carry arbitrary values, and `generateSearchParams` stringifies them. A request with a valid group field and `"strict_group_size":"garbage"` reaches `parseGroupByInfo`, where the `ParseBool` error is converted to `false`; it then reports success with weaker semantics. That is the same accepted-but-did-something-else failure class this PR fixes. A present value that cannot be parsed should return `ErrParameterInvalid`, with a regression test.

internal/proxy/search_util.go
```diff
@@ -838,6 +901,9 @@ func parseGroupByField(groupByFieldName string, schema *schemapb.CollectionSchem
 		// Extract field name (part before the first '[')
 		fieldName := strings.Split(groupByFieldName, "[")[0]
 		if field, exists := fieldNameMap[fieldName]; exists {
+			if err := checkGroupByFieldType(field); err != nil {
```
xiaofan-luan, line 904: [P2] Reject bracket paths on non-JSON scalar fields

This type check admits supported scalar fields such as `VarChar` and `Int64`, after which the bracket branch deliberately leaves `jsonPath` empty. Consequently `group_by_field=tag["x"]` succeeds but groups by the whole `tag` field, silently discarding the requested path. In the bracket branch, an existing schema field should be required to be JSON (or the request should be rejected), with a test for a supported scalar plus brackets.

xiaofan-luan (commented):
One more group-field normalization case remains.

PR #52352, opened 2026-08-10
enhance: release the bump partial writer on error paths instead of leaking it (#51649)

MrPresent-Han:
### What

Split from #52302 (orthogonal error-path hygiene).

The bump partial-backfill error paths returned between `setupWriter` and the success-path `Close` **without releasing the writer at all**. `FFIPackedWriter` holds a C handle plus `cProperties` with no finalizer, so the native writer **leaked** (master had no cleanup here). This PR wraps it in a `bumpSchemaVersionWriterLease` whose deferred `Cleanup` calls `Abort` (`packedRecordBatchWriter.Abort` → FFI `Destroy`), releasing the native writer without running the output-finalizing `Close`. `Close` returns a column-groups payload; `packed.CommitManifestUpdates` publishes that payload separately. A second `Close` on a consumed lease fails loudly instead of touching the native writer again.

### Boundary (important — not a clean rollback)

`Abort`/`Destroy` is **not** a full rollback:
- `Destroy` still runs the Arrow `ParquetFileWriter` destructor, which can finalize the **currently open row group** (so it is not literally “no flush”);
- on object storage the S3 sink (`CustomOutputStream`) has **no destructor**, so an **in-flight multipart upload is left neither completed nor aborted** — a dangling MPU, invisible to prefix-listing GC (it needs `ListMultipartUploads`).

Bytes already flushed during `Write`, plus any such trailing row group / dangling MPU, remain and need **storage-side reclamation** (a bucket lifecycle rule for incomplete multipart uploads) until milvus-storage exposes a cascading abort down to the sink. This is not a regression — master orphaned the same MPU (and additionally leaked the writer). The value here is the leak fix; the residue is documented, with the cascading abort tracked as a milvus-storage follow-up.

### Tests

`TestPartialMaterializationAbortsWriterOnMidStreamError` drives `runMissingFunctionMaterialization` (via `Compact`) with an injected writer that fails mid-stream and asserts the error path releases the writer via `Abort`, not `Close`. It fails if `defer writerLease.Cleanup()` is removed (the isolated lease unit tests cannot pin that wiring).

issue: #51649

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -756,6 +756,41 @@ type bumpSchemaVersionBatchWriter interface {
 	GetWrittenUncompressed() uint64
 	AsNewColumnGroups()
 	Close() (packed.WriterOutput, error)
+	Abort()
+}
+
+type bumpSchemaVersionWriterLease struct {
+	writer    bumpSchemaVersionBatchWriter
+	exhausted bool
+}
+
+func (l *bumpSchemaVersionWriterLease) Close() (packed.WriterOutput, error) {
+	if l.exhausted {
+		// A second Close is always a caller bug: after an error the native
+		// writer must not be touched again (the underlying nil-guard only
+		// covers the success path), and after success there is nothing left
+		// to close. Fail loudly instead of touching the FFI writer.
+		return nil, merr.WrapErrServiceInternalMsg("bump schema version writer lease already consumed")
+	}
+	output, err := l.writer.Close()
+	// Close consumes the native writer even when it returns an error. Never call
+	// it again from deferred cleanup; only the returned output still needs release.
+	l.exhausted = true
+	return output, err
+}
+
+// Cleanup releases an unconsumed writer on error paths via Abort instead of
+// output-finalizing Close. Manifest publication is a separate caller action via
+// packed.CommitManifestUpdates, and this error path never reaches it.
+// It does not guarantee no bytes are written: see packedRecordBatchWriter.Abort
+// for the object-storage residue (a possibly finalized open row group and a
+// dangling multipart upload) that storage-side reclamation must handle.
+func (l *bumpSchemaVersionWriterLease) Cleanup() {
+	if l == nil || l.exhausted {
+		return
+	}
+	l.exhausted = true
+	l.writer.Abort()
 }
 
 type bumpSchemaVersionWriterResult struct {
@@ -988,6 +1023,8 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx 
 	if err != nil {
 		return nil, err
 	}
+	writerLease := &bumpSchemaVersionWriterLease{writer: writerResult.writer}
+	defer writerLease.Cleanup()
 
 	log.Info(ctx, "schema bump writer setup",
 		mlog.Int64("effectiveStorageVersion", writerResult.storageVersion),
@@ -1100,12 +1137,12 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx 
 		return nil, err
 	}
 
-	writerOutput, err := writerResult.writer.Close()
-	if err != nil {
-		return nil, err
-	}
+	writerOutput, err := writerLease.Close()
 	if writerOutput != nil {
 		defer writerOutput.Destroy()
+	}
+	if err != nil {
+		return nil, err
 	}
 
 	manifestPath, err := packed.CommitManifestUpdates(
--- a/internal/datanode/compactor/bump_schema_version_compactor_test.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor_test.go
@@ -106,6 +106,26 @@ func (w *fakeBinlogRecordWriter) FlushChunk() error                  { return ni
 func (w *fakeBinlogRecordWriter) GetBufferUncompressed() uint64      { return 0 }
 func (w *fakeBinlogRecordWriter) Schema() *schemapb.CollectionSchema { return w.schema }
 
+type fakeBumpSchemaVersionBatchWriter struct {
+	closeCalls int
+	abortCalls int
+	output     packed.WriterOutput
+	closeErr   error
+	writeErr   error
+}
+
+func (w *fakeBumpSchemaVersionBatchWriter) Abort() { w.abortCalls++ }
+
+func (w *fakeBumpSchemaVersionBatchWriter) Write(storage.Record) error { return w.writeErr }
+func (w *fakeBumpSchemaVersionBatchWriter) GetWrittenUncompressed() uint64 {
+	return 0
+}
+func (w *fakeBumpSchemaVersionBatchWriter) AsNewColumnGroups() {}
+func (w *fakeBumpSchemaVersionBatchWriter) Close() (packed.WriterOutput, error) {
+	w.closeCalls++
+	return w.output, w.closeErr
+}
+
 type BumpSchemaVersionCompactionTaskSuite struct {
 	suite.Suite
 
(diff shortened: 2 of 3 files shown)


internal/storage/record_writer.go
```diff
@@ -271,6 +271,17 @@ func (pw *packedRecordBatchWriter) Close() (packed.WriterOutput, error) {
 	return out, nil
 }
 
+// Abort releases the underlying FFI writer without flushing pending column
+// groups — for error paths whose output will not be committed. Files already
+// flushed during Write are not undone; they stay unreferenced by the manifest.
+func (pw *packedRecordBatchWriter) Abort() {
```
liliu-z, line 288: The Abort doc comment and the PR's Boundary section describe the residue as "files ... unreferenced by the manifest", but on S3 the residue is worse than that: Destroy() deletes the C++ Writer, whose sink (CustomOutputStream) has no destructor, so an in-flight multipart upload is neither completed nor aborted. Orphaned MPU parts are invisible to prefix-listing GC (they need ListMultipartUploads), unlike unreferenced objects. This is not a regression — master returned from these error paths without touching the writer at all, so the MPU was equally orphaned — so treat it as a doc fix plus a follow-up: say explicitly that an incomplete multipart upload can be left behind and that a bucket lifecycle rule for incomplete MPUs is required until a milvus-storage-level cascading abort exists.

internal/storage/record_writer.go
```diff
@@ -271,6 +271,17 @@ func (pw *packedRecordBatchWriter) Close() (packed.WriterOutput, error) {
 	return out, nil
 }
 
+// Abort releases the underlying FFI writer without flushing pending column
```
liliu-z: "Without flushing pending column groups" (and "no farewell flush" in the PR body) is true only for the cached record batches. Destroying the writer still runs ~ParquetFileWriter, and because cached_batches_ is declared after sink_/writer_ it is dropped first, then writer_'s teardown reaches Arrow's ParquetFileWriter destructor, which calls Close() — closing the still-open buffered row group and writing the footer to the sink. So Destroy still performs a round of I/O and leaves a finalized-but-short Parquet file (minus the cached rows) rather than nothing. Reword the comment to say the cached batches are dropped but the open row group and footer are still finalized, or add a real format-level abort.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -978,6 +1011,8 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
 	if err != nil {
 		return nil, err
 	}
+	writerLease := &bumpSchemaVersionWriterLease{writer: writerResult.writer}
+	defer writerLease.Cleanup()
```
liliu-z, line 1027: The PR body says the error path "used to finalize the packed writer via Close", but master had no cleanup at all between setupWriter and the success-path Close — the eight intervening error returns just dropped the writer, and FFIPackedWriter holds a C handle plus cProperties with no runtime.SetFinalizer, so the native writer leaked for the process lifetime. The real effect of this change is adding cleanup where there was none, not swapping a flush for a no-flush teardown. Rewriting the description matters for anyone bisecting this behavior later.

internal/datanode/compactor/bump_schema_version_compactor_test.go
```diff
@@ -1428,3 +1447,62 @@ func (s *BumpSchemaVersionCompactionTaskSuite) TestMaterializationRejectsDropped
 	s.Error(err)
 	s.ErrorIs(err, merr.ErrServiceInternal)
 }
+
+func (s *BumpSchemaVersionCompactionTaskSuite) TestPartialWriterLeaseCleanupAbortsOnceWithoutFlush() {
```
liliu-z, line 1501: All four new tests construct a bumpSchemaVersionWriterLease directly around a fake and exercise Close/Cleanup in isolation; none of them drive runMissingFunctionMaterialization. Deleting `defer writerLease.Cleanup()` at bump_schema_version_compactor.go:1015 — the line that actually fixes the leak — leaves every one of them green, so the wiring itself is untested. A task-level test that mockey-patches the writer and forces one of the eight error returns, asserting Abort fired, would close the gap; a real-backend fault-injection test is not needed for that.

internal/storage/record_writer.go
MrPresent-Han: Addressed in dccd72ea99 (doc + PR body). The Abort doc and the PR Boundary section now say explicitly that on object storage an in-flight multipart upload is left neither completed nor aborted (the S3 CustomOutputStream sink has no destructor), that such dangling MPU parts are invisible to prefix-listing GC and need a bucket lifecycle rule for incomplete multipart uploads, and that a true cascading abort down to the sink is a milvus-storage follow-up. Also noted it is not a regression (master orphaned the same MPU).

internal/storage/record_writer.go
MrPresent-Han: Fixed the "no farewell flush" wording in dccd72ea99. The doc no longer claims Destroy skips all flushing: it now states Destroy still runs the Arrow ParquetFileWriter destructor, which can finalize the currently open row group, so bytes may still be written on the way down. The only guarantee is that it does not perform the manifest-committing Close.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Corrected the PR body in dccd72ea99. It no longer says the error path "used to finalize via Close" — master had no cleanup at all between setupWriter and the success-path Close, so the intervening error returns dropped the writer and, because FFIPackedWriter holds a C handle + cProperties with no finalizer, leaked it. The PR is now framed as that leak fix (title and body updated).

internal/datanode/compactor/bump_schema_version_compactor_test.go
MrPresent-Han: Added a non-vacuous guard in dccd72ea99: TestPartialMaterializationAbortsWriterOnMidStreamError drives runMissingFunctionMaterialization via Compact with an injected writer that fails mid-stream and asserts the error path releases the writer through the deferred lease Cleanup (Abort, not Close). Deleting `defer writerLease.Cleanup()` at the setupWriter call site makes abortCalls stay 0 and fails this test; the isolated lease unit tests stay green regardless, exactly as you noted.

internal/storage/record_writer.go
```diff
@@ -271,6 +271,26 @@ func (pw *packedRecordBatchWriter) Close() (packed.WriterOutput, error) {
 	return out, nil
 }
 
+// Abort releases the underlying FFI writer via Destroy instead of the
+// manifest-committing Close, for error paths that will not commit their output.
```
liliu-z: The new Abort doc calls Close "manifest-committing", but Close does not commit the manifest — it only calls loon_writer_close and hands the column-groups payload back to the caller, who commits it separately via packed.CommitManifestUpdates. This contradicts the invariant documented 19 lines above on Close itself and on FFIPackedWriter.Close, so a reader auditing the abort/close ownership split gets the wrong model of which side writes the manifest. Drop "manifest-committing" (e.g. "instead of the output-finalizing Close").

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -756,6 +756,40 @@ type bumpSchemaVersionBatchWriter interface {
...
+	}
+	output, err := l.writer.Close()
+	// Close consumes the native writer even when it returns an error. Never call
+	// it again from deferred cleanup; only the returned output still needs release.
+	l.exhausted = true
+	return output, err
+}
+
+// Cleanup releases an unconsumed writer on error paths via Abort (which, unlike
+// Close, does not commit the manifest), so the native FFI writer is not leaked.
```
liliu-z: Same inaccuracy as the Abort doc in record_writer.go, repeated here: the Cleanup comment contrasts Abort with Close as one "which, unlike Close, does not commit the manifest", implying Close commits it. Close only finalizes the writer output; the manifest is committed by the separate packed.CommitManifestUpdates call further down this same function. Reword both sites together so the ownership split reads consistently.

internal/storage/record_writer.go
MrPresent-Han: Addressed in 5300b335e8. The Abort comment now states that Close finalizes writer output and returns a column-groups payload, while packed.CommitManifestUpdates publishes that payload separately; the PR body and amended commit message use the same ownership model. This is wording-only, with no runtime behavior change; git diff --check is clean.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Addressed in 5300b335e8. The lease Cleanup comment now distinguishes output-finalizing Close from the separate packed.CommitManifestUpdates publication step, consistent with record_writer.go, the PR body, and the amended commit message. This is wording-only, with no runtime behavior change; git diff --check is clean.

PR #52353, opened 2026-08-10
fix: classify malformed multi_analyzer_params as data-integrity errors in bump backfill (#51649)

MrPresent-Han:
### What

Split from #52302 (orthogonal error-classification concern).

Bump backfill parses the persisted `multi_analyzer_params` to resolve the implicit `by_field` input column. These params were validated at DDL time, so failing to parse or resolve them during compaction means the **persisted schema itself is inconsistent** — not that the compaction request is malformed:

- malformed JSON: previously returned as a raw unclassified `json.Unmarshal` error → now `DataIntegrity` with the function name;
- missing `by_field` key: previously `ParameterInvalid` → now `DataIntegrity`;
- `by_field` referencing a field absent from the schema: previously `ParameterInvalid` ("input field not found in schema", no context) → now `DataIntegrity` naming the function and the dangling field.

Per the blame test in the error-handling guide: the request content does not force these branches; a correct Milvus never persists such a schema, so they are system-side integrity failures and must not be classified as user input errors.

issue: #51649

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -183,13 +183,13 @@ func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunc
 func additionalFunctionInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
 	switch functionSchema.GetType() {
 	case schemapb.FunctionType_BM25:
-		return bm25AdditionalInputFields(schema, inputField)
+		return bm25AdditionalInputFields(schema, functionSchema, inputField)
 	default:
 		return nil, nil
 	}
 }
 
-func bm25AdditionalInputFields(schema *schemapb.CollectionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
+func bm25AdditionalInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
 	params, ok := typeutil.CreateFieldSchemaHelper(inputField).GetMultiAnalyzerParams()
 	if !ok {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -183,29 +183,32 @@ func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunc
...
+		return nil, merr.WrapErrDataIntegrity(err, "failed to parse multi_analyzer_params for function %s", functionSchema.GetName())
 	}
 	if multiAnalyzerParams.ByField == "" {
-		return nil, merr.WrapErrParameterInvalidMsg("multi_analyzer_params missing required 'by_field' key")
+		return nil, merr.WrapErrDataIntegrityMsg("multi_analyzer_params for function %s is missing required 'by_field' key", functionSchema.GetName())
 	}
 	byField := typeutil.GetFieldByName(schema, multiAnalyzerParams.ByField)
 	if byField == nil {
-		return nil, merr.WrapErrParameterInvalidMsg("input field not found in schema")
+		return nil, merr.WrapErrDataIntegrityMsg("multi_analyzer_params for function %s references missing input field %s", functionSchema.GetName(), multiAnalyzerParams.ByField)
```
buqian-zilliz, line 211: **[P2] Classify invalid `by_field` selector types as data-integrity errors**

The missing-name branch is now classified as `DataIntegrity`, but a persisted `by_field` that resolves to an invalid selector type still escapes that classification. RootCoord DDL only permits a `VarChar` selector; this helper returns any named field, so non-string fields later become `ErrParameterInvalid`, while `Text` is accepted by the generic BM25 input validator even though DDL rejects it and may fail later during materialization. Validate the resolved selector against the DDL invariant here and return `ErrDataIntegrity`; otherwise the error class still depends on how the persisted schema is corrupted.

MrPresent-Han: Addressed in f4d23e65df. Added a type guard right after the name resolution: a resolved `by_field` whose type is not `VarChar` now returns `ErrDataIntegrity` (naming the function, the field, and its actual type), matching DDL's `validateMultiAnalyzerParams` invariant (only VarChar is admitted) and the classification of the other corruption branches — instead of degrading into an `ErrParameterInvalid` downstream (or, for `Text`, slipping past this helper and failing later at materialization). Regression test `TestMissingFunctionInputSchemaByFieldWrongTypeIsDataIntegrityError` covers a by_field resolving to an Int64 field.

PR #52354, opened 2026-08-10
enhance: Keep nullable chunk validity packed

sparknack:
issue: #52355

- Keep nullable chunk validity packed in the backing mmap or buffer for point checks, filtering, rank, and FixedWidth Span access.
- Add a non-owning `ValidityView` in a standalone header to represent packed and expanded validity.
- Remove the FixedWidth chunk-level `FixedVector<bool>` cache; direct consumers read bits from the view, while legacy `const bool*` interfaces materialize only for the current batch or Build call.
- Keep String/Array view validity request-local.

### DIFF
--- a/internal/core/src/common/Chunk.h
+++ b/internal/core/src/common/Chunk.h
@@ -13,6 +13,7 @@
 
 #include <unistd.h>
 #include <algorithm>
+#include <bit>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
@@ -85,12 +86,6 @@ class Chunk {
           size_(size),
           nullable_(nullable),
           chunk_mmap_guard_(chunk_mmap_guard) {
-        if (nullable) {
-            valid_.reserve(row_nums);
-            for (int i = 0; i < row_nums; i++) {
-                valid_.push_back((data[i >> 3] >> (i & 0x07)) & 1);
-            }
-        }
     }
     virtual ~Chunk() {
         // The ChunkMmapGuard will handle the unmapping and unlinking of the file if it is file backed
@@ -127,18 +122,68 @@ class Chunk {
         return data_;
     }
 
-    FixedVector<bool>&
-    Valid() {
-        return valid_;
-    }
-
     virtual bool
     isValid(int offset) const {
         if (nullable_) {
-            return valid_[offset];
+            return (static_cast<uint8_t>(data_[offset >> 3]) >>
+                    (offset & 0x07)) &
+                   1;
         }
         return true;
     };
+
+    ValidityView
+    Validity(int64_t offset = 0) const {
+        if (!nullable_) {
+            return {};
+        }
+        return ValidityView::FromPacked(reinterpret_cast<const uint8_t*>(data_))
+            .Subview(offset);
+    }
+
+    void
+    ApplyValidityMask(int64_t offset,
+                      int64_t count,
+                      TargetBitmapView target) const {
+        if (!nullable_ || count == 0) {
+            return;
+        }
+        AssertInfo(offset >= 0 && count >= 0 && offset + count <= row_nums_,
+                   "Invalid validity range, offset: {}, count: {}, rows: {}",
+                   offset,
+                   count,
+                   row_nums_);
+
+        const auto validity = Validity(offset);
+        const auto* packed = validity.packed_data();
+        AssertInfo(packed != nullptr, "Packed validity data is missing");
+        const auto read_word = [packed](int64_t bit_offset, int bit_count) {
+            const auto byte_offset = bit_offset >> 3;
+            const auto shift = static_cast<int>(bit_offset & 0x07);
+            const auto bytes = (shift + bit_count + 7) >> 3;
+            uint64_t word = 0;
+            const auto low_bytes = std::min(bytes, 8);
+            for (int i = 0; i < low_bytes; ++i) {
+                word |= uint64_t(packed[byte_offset + i]) << (i * 8);
+            }
+            word >>= shift;
+            if (bytes > 8) {
+                word |= uint64_t(packed[byte_offset + 8]) << (64 - shift);
+            }
+            if (bit_count < 64) {
+                word &= (uint64_t{1} << bit_count) - 1;
+            }
+            return word;
+        };
+
+        for (int64_t i = 0; i < count; i += 64) {
+            const auto bit_count =
+                static_cast<int>(std::min<int64_t>(count - i, 64));
+            auto word = read_word(validity.bit_offset() + i, bit_count);
+            TargetBitmapView validity_word(&word, bit_count);
+            target.view(i).inplace_and(validity_word, bit_count);
+        }
+    }
 
     int64_t
     PhysicalOffsetOf(int64_t logical_offset) const {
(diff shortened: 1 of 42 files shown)


internal/core/src/mmap/ChunkedColumnInterface.h
```diff
@@ -165,16 +163,8 @@ class ChunkedColumnInterface {
...
-        AssertInfo(
-            offset + size <= static_cast<int64_t>(valid_data.size()),
-            "Valid-data range out of valid-data bounds, offset: {}, size: {}, "
-            "valid-data size: {}",
-            offset,
-            size,
-            valid_data.size());
         for (int64_t i = 0; i < size; ++i) {
-            if (!valid_data[offset + i]) {
+            if (!chunk->isValid(offset + i)) {
```
buqian-zilliz: **[P2] Avoid per-row virtual dispatch when applying validity**

`ApplyValidDataInChunk` masks whole expression/index batches, but this change replaces direct contiguous validity access with a virtual `Chunk::isValid()` call for every row. Large nullable chunks therefore pay an indirect call plus repeated nullable/bit-decoding work per row, turning bitmap-only index paths into a scalar loop. Fetch `chunk->Validity(offset)` once and consume the packed bits in bulk (or at least iterate that view) instead.

internal/core/src/exec/expression/TermExpr.cpp
```diff
@@ -1132,7 +1132,7 @@ PhyTermFilterExpr::ExecVisitorImplForData(EvalCtx& context) {
             if (simd_filter_fn) {
                 simd_filter_fn(data, size, res);
                 // Apply validity mask
-                if (valid_data != nullptr) {
+                if (valid_data) {
```
buqian-zilliz: **[P2] Keep packed validity masking vectorized in the term SIMD path**

The sequential numeric `IN` path has already produced `res` with `simd_filter_fn`, but it then indexes the packed `ValidityView` one row at a time. That adds bit-offset arithmetic and shift/mask/branch work per row and can erase much of the SIMD gain on nullable scans. Use the existing `ApplyValidMask(valid_data, res, valid_res, size)` helper here, as the binary range/arithmetic paths do.

internal/core/src/segcore/SegmentGrowingImpl.cpp
```diff
@@ -1487,9 +1487,9 @@ SegmentGrowingImpl::chunk_vector_array_view_impl(
 
     std::vector<VectorArrayView> views;
     views.reserve(len);
-    FixedVector<bool> valid_data;
+    auto valid_data = std::make_shared<FixedVector<bool>>();
```
buqian-zilliz: **[P3] Skip the validity-owner allocation for non-nullable vector arrays**

`chunk_vector_array_view_impl` allocates a shared `FixedVector<bool>` even when the field is non-nullable, although the returned validity view is empty in that case. `ProcessDataByOffsets<VectorArrayView>` calls this function with a one-row range for every selected offset, so non-nullable random-access filters now incur an extra heap/control-block allocation per row. Allocate the owner only for nullable fields and use the ownerless `PinWrapper` path otherwise.

internal/core/src/exec/expression/UnaryExpr.cpp
```diff
@@ -1951,7 +1951,7 @@ PhyUnaryRangeFilterExpr::ExecRangeVisitorImplForData(EvalCtx& context) {
         // there is a batch operation in BinaryRangeElementFunc,
         // so not divide data again for the reason that it may reduce performance if the null distribution is scattered
         // but to mask res with valid_data after the batch operation.
-        if (valid_data != nullptr) {
+        if (valid_data) {
```
buqian-zilliz: **[P2] Keep unary range validity masking on the batch path**

The sequential numeric branch has already evaluated the batch with the bitset comparison helpers, but this loop now indexes a packed `ValidityView` for every row. On sealed nullable scans that adds offset/shift/mask work per row and can erase much of the vectorized comparison gain. Use `ApplyValidMask(valid_data, res, valid_res, size)` for the sequential no-bitmap case, retaining the gather/fused loop for random or bitmap-sensitive execution.

internal/core/src/exec/expression/CompareExpr.h
```diff
@@ -497,16 +497,16 @@ class PhyCompareFilterExpr : public Expr {
...
                  res + processed_size,
                  values...);
-            const bool* left_valid_data = left_chunk.valid_data();
-            const bool* right_valid_data = right_chunk.valid_data();
+            const auto left_validity = left_chunk.validity();
+            const auto right_validity = right_chunk.validity();
             // mask with valid_data
             for (int i = 0; i < size; ++i) {
-                if (left_valid_data && !left_valid_data[i + data_pos]) {
+                if (left_validity && !left_validity[i + data_pos]) {
```
buqian-zilliz: **[P2] Vectorize validity masking in two-column comparisons**

This fast path evaluates the whole numeric left/right batch with `inplace_compare_column`, but then decodes both packed validity views one row at a time; the multiple-chunk path repeats the same loop. Nullable sealed comparisons therefore add two scalar bit extractions per row directly after the SIMD comparison. Apply `ApplyValidMask(left_validity.Subview(data_pos), ...)` and the corresponding right-hand mask in word-sized blocks, keeping the gather loop only for offset input.

internal/core/src/index/json_stats/JsonKeyStats.h
```diff
@@ -628,26 +622,26 @@ class JsonKeyStats : public ScalarIndex<std::string> {
...
-    ApplyValidData(const bool* valid_data,
+    ApplyValidData(ValidityView validity,
                    TargetBitmapView res,
                    TargetBitmapView valid_res,
                    const int size) {
-        if (valid_data != nullptr) {
+        if (validity) {
             for (int i = 0; i < size; i++) {
-                if (!valid_data[i]) {
+                if (!validity[i]) {
```
buqian-zilliz: **[P2] Keep JSON shredding validity masking word-at-a-time**

`ValidityView` is packed for sealed shredding chunks after this change, but both validity helpers still call `operator[]` once per row. They scan each shredding chunk, including the SkipIndex branch, so a skipped nullable chunk still pays bit-offset arithmetic, shift/mask, and a branch for every row even though value evaluation was avoided. Reuse a word-at-a-time packed mask (following the existing `ApplyValidMask` or `Chunk::ApplyValidityMask` pattern; the valid-only path can delegate to `ApplyValidDataInChunk`) so skipping value evaluation remains a bulk operation.

internal/core/src/common/Chunk.h
```diff
@@ -540,7 +590,7 @@ class VectorArrayChunk : public Chunk {
                    "rows {}",
                    idx,
                    row_nums_);
-        AssertInfo(!nullable_ || valid_[idx],
+        AssertInfo(!nullable_ || isValid(idx),
```
buqian-zilliz, line 593: **[P3] Avoid rechecking nullable vector-array rows in the batch path**

`Views()` already tests `validity[i - start_offset]` before calling `View(i)`, but `View()` repeats the same packed-bit check through `isValid(idx)` for every valid row. Full-chunk vector-array scans therefore decode each valid row twice and retain a virtual call site. Keep the checked `View()` for random access, but have `Views()` use a private unchecked builder after its outer validity check so each row's validity is read once.

internal/core/src/exec/expression/UnaryExpr.h
```diff
@@ -902,11 +902,11 @@ class ShreddingExecutor {
     }
 
     void
-    HandleValidData(const bool* valid,
+    HandleValidData(ValidityView valid,
                     size_t size,
                     TargetBitmapView res,
                     TargetBitmapView valid_res) {
-        if (valid != nullptr) {
+        if (valid) {
```
buqian-zilliz: **[P2] Apply packed validity in bulk after shredding comparisons**

`BatchUnaryCompare` has already produced the whole shredding-column result bitmap, but `HandleValidData` then walks every row through `ValidityView::operator[]`. For packed nullable chunks this adds an encoding branch plus byte/shift/mask work per row; in the BOOL JSON-stats path the comparison itself is only a cheap batch operation, so this scalar validity pass can dominate the full-segment scan. Use `ApplyValidMask(valid, res, valid_res, size)` or an equivalent word-at-a-time mask, as the raw-data path now does.

internal/core/src/exec/expression/Expr.h
```diff
@@ -120,6 +120,53 @@ ApplyValidMask(const bool* valid_data,
     }
 }
 
+inline void
+ApplyValidMask(ValidityView validity,
```
zhengbuqian, line 124: this is somewhat duplicate(though not exactly the same) with the one in Chunk.h

zhengbuqian (approved):

Issue #52356, opened 2026-08-10
[Bug]: Storage V3 writers do not collect TTL field values, TTL-field expiry never triggers compaction


xiaocai2333:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```
- Milvus version: master
- Deployment mode(standalone or cluster): both
```

### Current Behavior

When a collection enables field-level TTL (collection property `ttl_field`), the binlog record writers are expected to collect the TTL column values during write and compute `ExpirQuantiles` (20%/40%/60%/80%/100% percentiles), which DataCoord's `compactionTrigger.ShouldCompactExpiryWithTTLField` uses to decide when a segment should be compacted to physically remove expired rows.

The storage V1 writer (`CompositeBinlogRecordWriter`, `internal/storage/serde_events.go`) and the storage V2 writer (`PackedBinlogRecordWriter`, `internal/storage/binlog_record_writer.go`) both collect `ttlFieldValues` in their `Write()` methods.

However, the storage V3 writers do **not**:

- `PackedManifestRecordWriter.Write()`
- `PackedTextManifestRecordWriter.Write()`

Both initialize `ttlFieldID` / `ttlFieldValues` in their constructors, but their `Write()` methods never extract the TTL column, so `ttlFieldValues` stays empty.

As a result, for V3-written segments `calculateExpirQuantiles` falls into the `idx >= len(ttlFieldValues)` branch for every percentile and fills all 5 quantiles with `neverExpire` (maxInt64). `ShouldCompactExpiryWithTTLField` then never fires (`startTs >= maxInt64` is never true), which means:

- Expired rows are still hidden from search/query (segcore injects the TTL filter expression at query time, unaffected).
- But **TTL-field expiry never triggers compaction** for V3 segments, so expired rows are never physically removed unless some other compaction condition happens to fire.

### Expected Behavior

Storage V3 writers collect TTL field values the same way V1/V2 do, so `ExpirQuantiles` are correctly populated and TTL-field-based compaction triggering works for V3 segments.

### Steps To Reproduce

```
1. Enable storage V3 (manifest-based packed writer).
2. Create a collection with a Timestamptz field and collection property ttl_field pointing to it.
3. Insert rows whose TTL values are already in the past; flush.
4. Observe segment meta: ExpirQuantiles are all maxInt64 (neverExpire).
5. No single compaction is triggered by TTL expiry for these segments.
```

### Anything else?

[...]

PR #52357, opened 2026-08-10
fix: collect TTL field values in storage V3 writers for ExpirQuantiles

xiaocai2333:
issue: #52356

### What this PR does

The storage V3 writers (`PackedManifestRecordWriter` and `PackedTextManifestRecordWriter`) initialized `ttlFieldID` / `ttlFieldValues` but never collected the TTL column values in `Write()`, unlike the V1 (`CompositeBinlogRecordWriter`) and V2 (`PackedBinlogRecordWriter`) writers.

As a result, for V3-written segments `calculateExpirQuantiles` saw an empty value slice and filled all 5 quantiles with `neverExpire` (maxInt64), so `compactionTrigger.ShouldCompactExpiryWithTTLField` never fired: **TTL-field expiry never triggered compaction for V3 segments**, and expired rows were only physically removed if some other compaction condition happened to fire. (Query-time filtering was unaffected — segcore injects the TTL filter expression independently of quantiles.)

### Fix

- Extract the TTL collection block from `PackedBinlogRecordWriter.Write()` into a shared `collectTTLValues()` helper on `packedBinlogRecordWriterBase`.
- Call it from the V2 writer and both V3 writers.

### Tests

- `Test_packedBinlogRecordWriterBase_collectTTLValues`: unit tests for the helper (positive values collected; null / non-positive skipped; disabled ttlFieldID no-op; non-int64 column errors).
- `TestPackedRecordWriters_TTLExpirQuantiles`: end-to-end (local storage FFI) `Write` → `Close` → `GetLogs` quantile assertions for all three packed writers (V2 regression guard + both V3 variants). Verified the two V3 subtests fail without the fix and pass with it.

### Notes

Segments already written by V3 have all-`neverExpire` quantiles persisted in meta; they are corrected the next time the segment is rewritten by any compaction. Backfilling those segments proactively is out of scope here.

### DIFF
--- a/internal/storage/binlog_record_writer.go
+++ b/internal/storage/binlog_record_writer.go
@@ -121,6 +121,35 @@ func (pw *packedBinlogRecordWriterBase) GetExpirQuantiles() []int64 {
 	return calculateExpirQuantiles(pw.ttlFieldID, pw.rowNum, pw.ttlFieldValues)
 }
 
+// collectTTLValues accumulates positive TTL field values from the record for
+// ExpirQuantiles calculation. Null and non-positive values mean "never expire"
+// and are skipped.
+func (pw *packedBinlogRecordWriterBase) collectTTLValues(r Record) error {
+	if pw.ttlFieldID < common.StartOfUserFieldID {
+		return nil
+	}
+	ttlColumn := r.Column(pw.ttlFieldID)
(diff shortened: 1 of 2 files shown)


czs007:

czs007 (approved):
/lgtm

Issue #52359, opened 2026-08-10
[Bug]: Struct sub-field hybrid scalar index load asserts on missing version


congqixia:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 3.0 branch image
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

The struct sub-field path identifies the field as a nested array index, but
`IndexFactory::CreateNestedIndex()` does not handle `HYBRID_INDEX_TYPE`.
It falls through to `CreateNestedIndexScalarIndexSort()`. For `VarChar`
elements this creates `StringIndexSort`, whose `LoadEntries()` expects a
`version` metadata key. The hybrid packed meta does not have that key, so
segment load fails with:

```text
At LoadSegment: At Load: Assert "meta_json_.contains(key)"
=> Meta key not found: version
```

### Expected Behavior

Loading a struct sub-field array `HYBRID` index should preserve the hybrid
wrapper, read the persisted internal `index_type`, and dispatch to the matching
nested internal scalar index. A packed hybrid meta such as the following should
not be loaded as `StringIndexSort` directly

### Steps To Reproduce

```markdown
1. create hybrid scalar index for varchar struct sub field with 2.6.x image
2. upgrade to 3.0 
3. load collection
```

### Milvus Log

```text
At LoadSegment: At Load: Assert "meta_json_.contains(key)"
=> Meta key not found: version
```

yanliang567:
There is no hybrid index type for milvus, it is actually an internal type, which could be triggered by auto-index

yanliang567:
Reproduced deterministically with the upgrade/rollback compatibility gate from [vectordb-testbricks#27](https://github.com/yanliang567/vectordb-testbricks/pull/27).

Environment:

- workflow: `pr27-idxfmt-r5-qpnfw`
- test commit: `d2b680f9cb6f29656fb10940de40374cb5f52dac`
- base: `v2.6.18@sha256:c6e332d3783c2c42649d5f76c5dae79d553927196a60547f619be13484ab44f6`
- target: `3.0-20260807-1439dc7d@sha256:ed46e16fcb58bd460722e6fc1c0e6294e86fd4e062431877d0a872dcb510cd64`
[...]

PR #52360, opened 2026-08-10
fix: Rectify nested index factory to correctly handle hybrid indexes

congqixia:
Related to #52359

Struct sub-field scalar indexes use the nested index factory because their logical field type is an array. That factory handled bitmap and inverted indexes explicitly, but allowed hybrid indexes to fall through to the sort index path.

For varchar array sub-fields, the fallback creates StringIndexSort and loads the hybrid packed meta as a sort index. Hybrid packed meta records the internal index_type and does not contain the sort index version key, so segment loading can fail with "Meta key not found: version".

Create nested HybridScalarIndex instances for hybrid nested indexes and keep the nested flag when the hybrid wrapper creates its internal bitmap, sort, or inverted index. This lets LoadEntries read the persisted hybrid internal type before dispatching to the real nested index loader.

### DIFF
--- a/internal/core/src/index/BitmapIndexArrayTest.cpp
+++ b/internal/core/src/index/BitmapIndexArrayTest.cpp
@@ -807,6 +807,39 @@ TEST(BitmapIndexArrayNestedTest, FactoryCreatesNestedBitmapForStructSubField) {
     boost::filesystem::remove_all(root_path);
 }
 
+TEST(BitmapIndexArrayNestedTest, FactoryCreatesNestedHybridForStructSubField) {
+    proto::schema::FieldSchema field_schema;
+    field_schema.set_name("struct_field[sub]");
+    field_schema.set_data_type(proto::schema::DataType::Array);
+    field_schema.set_element_type(proto::schema::DataType::VarChar);
+
+    auto field_meta = storage::FieldDataMeta{1, 2, 3, 101, field_schema};
+    auto index_meta = storage::IndexMeta{3, 101, 3002, 3002};
+
+    auto root_path =
+        fmt::format("{}/hybrid_nested_array_factory", TestLocalPath);
+    boost::filesystem::remove_all(root_path);
+    storage::StorageConfig storage_config;
+    storage_config.storage_type = "local";
+    storage_config.root_path = root_path;
+    auto chunk_manager = storage::CreateChunkManager(storage_config);
+    auto fs = storage::InitArrowFileSystem(storage_config);
+    storage::FileManagerContext ctx(field_meta, index_meta, chunk_manager, fs);
+
+    index::CreateIndexInfo index_info{};
+    index_info.index_type = milvus::index::HYBRID_INDEX_TYPE;
+    index_info.field_type = DataType::ARRAY;
+    index_info.field_name = "struct_field[sub]";
+    index_info.tantivy_index_version = 7;
+
+    auto index =
+        index::IndexFactory::GetInstance().CreateIndex(index_info, ctx);
+    ASSERT_TRUE(index->IsNestedIndex());
+    ASSERT_EQ(index->Type(), milvus::index::HYBRID_INDEX_TYPE);
+
+    boost::filesystem::remove_all(root_path);
+}
+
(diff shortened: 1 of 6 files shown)


internal/core/src/index/HybridScalarIndex.cpp
```diff
@@ -181,14 +190,18 @@ HybridScalarIndex<T>::GetInternalIndex() {
         return internal_index_;
     }
     if (internal_index_type_ == ScalarIndexType::BITMAP) {
-        internal_index_ =
-            std::make_shared<BitmapIndex<T>>(this->file_manager_context_);
+        internal_index_ = std::make_shared<BitmapIndex<T>>(
```
liliu-z, line 193: CreateNestedIndexHybrid hard-codes is_nested_index_=true and the wrapper forwards it into whichever internal index the persisted index_type names, so a 2.6-built payload is relabeled element-level even though it was built row-level. Milvus 2.6 has no nesting concept at all (its BitmapIndex.h has no is_nested_index_ member and its WriteEntries persists only BITMAP_INDEX_LENGTH/BITMAP_INDEX_NUM_ROWS), so the marker is simply absent: BitmapIndex::LoadEntries falls back to the constructor value (true) and InvertedIndexTantivy never persists or restores the flag at all. That covers both internals a 2.6 hybrid on a varchar array can have — SelectBuildTypeForArrayType only ever picks BITMAP or INVERTED — which is exactly the 2.6 -> 3.0 upgrade this PR targets (issue #52359 repro: build the index on 2.6.x, upgrade to 3.0, load). The query layer then reads the wrapper's answer and maps row offsets through ElementIDToRowID as if they were element ids, turning today's loud load failure into silent wrong results. Legacy payloads need to be rejected or rebuilt rather than relabeled, and newly built indexes need an authoritative persisted layout marker; HybridScalarIndex.h:80 is part of the fix since Expr.h:1988 pins the wrapper and caches its answer.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -1187,6 +1187,10 @@ IndexFactory::CreateNestedIndex(
     if (index_type == BITMAP_INDEX_TYPE) {
         return CreateNestedIndexBitmap(file_manager_context);
     }
+    if (index_type == HYBRID_INDEX_TYPE) {
```
liliu-z, line 1190: The new HYBRID branch dispatches on the index_type recorded in etcd, not on what is actually inside the packed file, so any struct sub-field HYBRID index built by a 3.0 binary from before this commit now fails to load. Those builds took the same fallback the PR removes and wrote a StringIndexSort payload whose meta is {version, num_rows, is_nested} with no index_type key; the new path sends them to HybridScalarIndex::LoadEntries, whose first statement is an unguarded GetMeta<uint8_t>("index_type") and therefore throws "Meta key not found: index_type". IndexEntryReader already exposes HasMeta(), so LoadEntries can detect a marker-less payload and fall back to the sort layout instead of asserting. The reviewer's two extra claims do not hold: the BinarySet path (load_index_c.cpp appendScalarIndex) never sets field_name so it never reaches the nested factory, and the missing null check in DeserializeIndexType is pre-existing code this PR does not touch or newly expose.

internal/core/src/index/BitmapIndexArrayTest.cpp
```diff
@@ -807,6 +807,39 @@ TEST(BitmapIndexArrayNestedTest, FactoryCreatesNestedBitmapForStructSubField) {
     boost::filesystem::remove_all(root_path);
 }
 
+TEST(BitmapIndexArrayNestedTest, FactoryCreatesNestedHybridForStructSubField) {
```
liliu-z, line 810: The new test only checks that the factory hands back a nested hybrid wrapper; it never builds or loads anything, so the load-side behavior the PR exists for is untested. Nothing covers a packed meta that omits the is_nested key (the 2.6 layout) or one that omits index_type (the pre-PR 3.0 sort layout), which are precisely the two upgrade payloads the new dispatch changes. Add cases that load each and assert the resulting nesting mode and load outcome.

Issue #52339, opened 2026-08-09
[Bug]: SINDI growing sparse index on index version 8 crashes QueryNode after reporting unsupported algorithm


yanliang567:
## Environment

- Milvus image:
  `harbor.milvus.io/milvusdb/milvus:v3.0.0@sha256:49371c30af46b1013e4d3e0b980e691d81376d69cdbe1b372725baf1d7255862`
- Milvus source tag commit: `f46a0328558be155d11266a1a2b90602ccc9b366`
- PyMilvus: `3.0.1`
- Deployment: Milvus Operator cluster, Woodpecker WAL, one QueryNode
- Configured target versions:
  `dataCoord.targetVecIndexVersion=10` and
  `dataCoord.targetScalarIndexVersion=4`
- Sparse index: `SPARSE_INVERTED_INDEX`
- Sparse algorithm: `SINDI`
- Metric: `IP`

## Problem

After a loaded collection with a SINDI sparse index receives enough writes to
build a growing-segment sparse index, QueryNode first reports that SINDI is not
supported by index version 8 and then terminates with `SIGSEGV`:

```text
[KNOWHERE][ValidateInvertedIndexAlgo] Unsupported sparse inverted index
algorithm SINDI for index version 8

[SERVER][AppendSegmentIndexSparse] growing sparse index build error:
failed to build index, knowhere inner error at
../internal/core/src/index/VectorMemIndex.cpp:501

SIGNAL CATCH BY NON-GO SIGNAL HANDLER
SIGNO: 11; SIGNAME: Segmentation fault; SI_CODE: 1; SI_ADDR: (nil)

knowhere::sparse::inverted::GrowableInvertedIndex<float, float>::add(...)
```

The Pod restarts repeatedly. During the restart loop, Proxy operations fail
with errors including:

```text
complexDelete: node not found
channel tsafe stalled
dial tcp <query-node-ip>:21123: connect: connection refused
```

An unsupported algorithm/version combination must be rejected or handled as a
normal index-build error. It must not crash QueryNode.

## Upgrade/Rollback Gate Reproduction

- Argo workflow: `pr25-cl30-idx104-r1-gptfx`
- Test repository commit: `0914249b78f92e3346084c7892ebb9cee3c5aab6`
- Scenario: `cluster-3-0-index-v10-v4-upgrade-rollback`
- Failed node: `strict-pressure-before-upgrade`
- Failure phase: v3.0.0 baseline, before any upgrade
[...]

PR #52361, opened 2026-08-10
fix: use target index version for interim index

chasingegg:
issue: #52339

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -97,7 +97,6 @@
 #include "knowhere/dataset.h"
 #include "knowhere/index/index_static.h"
 #include "knowhere/sparse_utils.h"
-#include "knowhere/version.h"
 #include "log/Log.h"
 #include "milvus-storage/common/constants.h"
 #include "milvus-storage/common/extend_status.h"
@@ -6853,6 +6852,8 @@ ChunkedSegmentSealedImpl::generate_interim_index(
         auto index_metric = field_binlog_config->GetMetricType();
 
         if (enable_binlog_index()) {
+            const auto index_version =
+                segcore_config_.get_interim_index_version();
             std::unique_ptr<
                 milvus::cachinglayer::Translator<milvus::index::IndexBase>>
                 translator =
@@ -6863,6 +6864,7 @@ ChunkedSegmentSealedImpl::generate_interim_index(
                         field_id.get(),
                         interim_index_type,
                         index_metric,
+                        index_version,
                         build_config,
                         dim,
                         is_sparse,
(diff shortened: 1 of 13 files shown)


internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
```diff
@@ -6872,7 +6872,7 @@ ChunkedSegmentSealedImpl::generate_interim_index(
                 milvus::cachinglayer::Manager::GetInstance().CreateCacheSlot(
                     std::move(translator));
             auto index_version =
-                knowhere::Version::GetCurrentVersion().VersionNumber();
+                knowhere::Version::GetMaximumVersion().VersionNumber();
```
liliu-z: Bumping the sealed interim index from knowhere v8 to the maximum version (v11) silently flips the posting-list precision of sparse IP interim indexes from fp32 to fp16, and the interim build config has no way to opt out. `VecIndexConfig` copies only `inverted_index_algo` and the BM25 params out of the user's index params, so `quant_type` is never in `build_config`; knowhere then resolves an absent `quant_type` for IP as `use_fp16 = version_support_fp16_quant_for_ip() && (qt == "fp16" || qt.empty())`, which was false at v8 and is true at v11 in the default OSS build (threshold 10; Cardinal builds define it to 0 and are unaffected). For a field created with `quant_type: fp32` the persisted index stays fp32 while the interim index now quantizes to fp16, so the same vectors return different distances depending on whether the segment is served from the interim or the persisted path. Forwarding `quant_type` from the field index params into the interim build config fixes this independently of the version policy.

internal/core/src/segcore/FieldIndexing.cpp
```diff
@@ -178,12 +178,16 @@ VectorFieldIndexing::VectorFieldIndexing(const FieldMeta& field_meta,
 void
 VectorFieldIndexing::recreate_index(DataType data_type,
                                     const VectorBase* field_raw_data) {
+    // Growing indexes are ephemeral and will never be serialized, so use the latest version
+    // supported by this Knowhere build.
+    const auto index_version =
+        knowhere::Version::GetMaximumVersion().VersionNumber();
```
liliu-z: The growing-segment path takes the same fp32→fp16 flip as the sealed interim path, and reviewers only looked at the sealed side. `recreate_index` now builds the sparse growing index at knowhere v11 through the same `VecIndexConfig`, which never forwards `quant_type`, so a sparse IP field created with `quant_type: fp32` gets an fp16 growing index while its eventual persisted index stays fp32. Any fix for the sealed interim build config has to cover this construction site too, since both share `VecIndexConfig`.

internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
chasingegg: apply sparse index params to growing index params also

internal/core/src/segcore/FieldIndexing.cpp
chasingegg: same

liliu-z (commented):
/lgtm
/approve

Issue #51612, opened 2026-07-20
[Bug]: QueryNode panics after AddField when delayed old-schema inserts are merged with new-schema inserts


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus image: 2.6-20260718-c8136e53
- Git commit: c8136e53e6eceab4498a015226898d33342ae33f
- Deployment mode: standalone
- MQ type: Kafka
- SDK version: pymilvus from the chaos-test client image; the exact version was not retained
- OS: Kubernetes/Linux; the exact container distribution was not retained
- CPU/Memory: not retained
- GPU: N/A
- Namespace: chaos-testing
- Instance: standalone-container-kill-571
- Chaos type: container-kill
- Jenkins job: chaos-test-kafka-for-release-cron
- Jenkins build: 571
- Jenkins URL: http://jenkins.milvus-cell.tc-op.zilliz.cn/job/chaos-test-kafka-for-release-cron/571/
```

Historical K8s pod information:

```text
Pod:
standalone-container-kill-571-milvus-standalone-fcfb6cb6c-dhvvm

Last State: Terminated
Reason: Error
Exit Code: 134
Restart Count: 4

Kubernetes events:
- Created container: count 5
- Started container: count 5
- BackOff restarting failed container: count 4
- Readiness probe failed with connection reset by peer
- No OOMKilled or image pull failure was observed
```

### Current Behavior

During concurrent AddField and DML operations, a standalone QueryNode merges insert messages written with different schema versions into one growing-segment `InsertRecord`.

When the first messages contain the newly added nullable field and later delayed messages do not contain it, `MergeFieldData` leaves the new field shorter than `InsertRecord.NumRows`. The malformed record enters segcore and later produces a retrieve result whose IDs/offsets are longer than the nullable field's `ValidData`.

The QueryNode then panics while reducing query results:

```text
panic: runtime error: index out of range [601] with length 601

github.com/milvus-io/milvus/pkg/v2/util/typeutil.AppendFieldData
    pkg/util/typeutil/schema.go:967

github.com/milvus-io/milvus/internal/querynodev2/segments.MergeSegcoreRetrieveResults
    internal/querynodev2/segments/result.go:537

github.com/milvus-io/milvus/internal/querynodev2/segments.(*defaultLimitReducerSegcore).Reduce
github.com/milvus-io/milvus/internal/querynodev2/tasks.(*QueryTask).Execute
```

[...]

PR #50531, opened 2026-06-15
enhance: add delegator delete micro batch

chyezh:
issue: #50121

- querynode pipeline: merge buffered DML MsgPacks at the stream pipeline input without crossing non-DML barriers
- delegator delete: add batch delete processing while preserving per-delete timestamp boundaries in the delete buffer
- tests: cover DML MsgPack merging and delete batch timestamp preservation

PR #52362, opened 2026-08-10
fix: keep insert message packs separate

chyezh:
issue: #51612

### DIFF
--- a/internal/querynodev2/pipeline/insert_node.go
+++ b/internal/querynodev2/pipeline/insert_node.go
@@ -53,12 +53,12 @@ func (iNode *insertNode) addInsertData(insertDatas map[UniqueID]*delegator.Inser
 		panic(err)
 	}
 	if len(skippedFields) > 0 {
-		// Attributing the skip to dropped fields is safe only because a SchemaChange
-		// message never shares a pack with inserts: the WAL adaptor emits every V1/V2
-		// message as its own pack, msgdispatcher never merges packs, and the DML
-		// micro-batcher refuses to merge non-Insert/Delete messages. If that
-		// pack-granularity invariant is ever relaxed, filtering against the current
-		// schema could silently drop fields that exist in a newer schema.
+		// Attributing the skip to dropped fields is safe because the WAL adaptor emits
+		// every V1/V2 message as its own pack, msgdispatcher never merges packs, and the
+		// DML micro-batcher only merges delete-only packs. An Insert message therefore
+		// never shares a pipeline batch with another MsgPack. If that pack-granularity
+		// invariant is ever relaxed, filtering against the current schema could silently
+		// drop fields that exist in a newer schema.
 		mlog.Warn(ctx, "skip insert payload fields absent from current schema, fields are dropped since the message was written",
 			mlog.FieldCollectionID(iNode.collectionID),
 			mlog.FieldSegmentID(msg.SegmentID),
--- a/internal/util/pipeline/stream_pipeline.go
+++ b/internal/util/pipeline/stream_pipeline.go
@@ -220,7 +220,9 @@ func NewDMLMsgPackBatcher(maxMsgNum func() int) MsgPackBatcher {
 
 func (b *dmlMsgPackBatcher) Batch(first *msgstream.MsgPack, input <-chan *msgstream.MsgPack) (*msgstream.MsgPack, *msgstream.MsgPack) {
 	maxMsgNum := b.getMaxMsgNum()
-	if maxMsgNum <= 1 || !isDMLMsgPack(first) {
+	// Preserve original pack boundaries for inserts because insertNode merges
+	// records by segment without normalizing different schema field sets.
+	if maxMsgNum <= 1 || !isDeleteOnlyMsgPack(first) {
 		return first, nil
 	}
 
@@ -232,7 +234,7 @@ func (b *dmlMsgPackBatcher) Batch(first *msgstream.MsgPack, input <-chan *msgstr
 			if !ok {
 				return mergeMsgPacks(packs), nil
 			}
-			if !isDMLMsgPack(next) {
+			if !isDeleteOnlyMsgPack(next) {
 				return mergeMsgPacks(packs), next
 			}
 			if msgNum+len(next.Msgs) > maxMsgNum {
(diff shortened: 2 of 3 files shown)


czs007:

Reviewed at head `e60fefd56b` (single commit, `internal/util/pipeline/stream_pipeline.go` + its test). **The fix is correct and mergeable** — narrowing the micro-batcher to delete-only packs does close the path that produced the mixed-schema `InsertRecord` in #51612, and the new tests cover the three interesting shapes (insert-first, delete-batch-stops-at-insert, mixed-pack-first). No blocking issues. Findings below are hardening / doc-drift.

I verified the mechanism end to end rather than just the diff:
- V1/V2 WAL messages are emitted one-per-pack (`pkg/streaming/util/message/adaptor/handler.go:141-146`), so before this change the batcher was the only thing that could put two independently-written insert msgs into one pack;
- `AddCollectionField` / `AlterCollection` packs were already batch barriers under the old `isDMLMsgPack`, so the two mergeable inserts were both *post*-DDL messages, one of them written by a proxy still holding the stale schema — exactly the "delayed message without the new nullable field" in the issue;
- `mergeMsgPacks` position handling and the `pending` hand-back in `streamPipeline.work()` are unchanged, so no message loss or checkpoint regression from the new barrier.

---

### medium — `insert_node.go` still documents the old, weaker batcher invariant

`internal/querynodev2/pipeline/insert_node.go:55-61`:

```go
// Attributing the skip to dropped fields is safe only because a SchemaChange
// message never shares a pack with inserts: ... and the DML
// micro-batcher refuses to merge non-Insert/Delete messages. If that
// pack-granularity invariant is ever relaxed, filtering against the current
// schema could silently drop fields that exist in a newer schema.
```

This comment is load-bearing — it is the stated justification for treating `skippedFields` as dropped fields. It now understates the rule: after this PR the batcher refuses to merge anything that is not delete-only. Leaving the old wording documents "merging Insert packs is safe", which is precisely the belief this PR is correcting, and invites a future change to re-widen the predicate.

**Fix:** update it to "…the DML micro-batcher only merges delete-only packs, so an Insert message is never merged with any other pack", and reference `isDeleteOnlyMsgPack`.

### medium — the corruption primitive in `addInsertData` remains unguarded

`internal/querynodev2/pipeline/insert_node.go:84-89`:

```go
err := typeutil.MergeFieldData(iData.InsertRecord.FieldsData, insertRecord.FieldsData)
...
iData.InsertRecord.NumRows += insertRecord.NumRows
```

`MergeFieldData` (`pkg/util/typeutil/schema.go:2331`) only errors for *src-has-a-field-dst-lacks*. The reverse — dst carries the newly added nullable field, src does not — is silent: that field stays short while `NumRows` advances, producing the malformed record that later panics in `typeutil.AppendFieldData` far from the cause. So the two orderings of the same mismatch behave as `panic(err)` vs. silent corruption.

This PR removes the batcher as a *producer* of that state, but `addInsertData` merges any two insert msgs for the same segment inside one `insertNodeMsg`, and a single pack can legitimately carry multiple insert messages: a transaction is expanded into multiple `TsMsg` in one pack (`pkg/streaming/util/message/adaptor/message.go:29-37`, cf. `produceTxn` in `internal/distributed/streaming/internal/producer/producer_task.go:155`), and `VersionOld` messages sharing a time tick are grouped (`adaptor/handler.go:124-140`). Those messages come from one request and should share a schema version, so I am **not** claiming a live second trigger — the point is that the invariant is now enforced only by the batcher predicate and by pack granularity, with nothing checking it at the merge site.

**Fix (cheap, recommended as follow-up):** after the merge, assert each merged field's row count equals `iData.InsertRecord.NumRows` (or compare field-ID sets before merging) and fail loudly with collection/segment/schema-version in the message. That converts a silent data corruption + distant panic into a local, diagnosable error, and would have pinned #51612 immediately.

### low — config doc now describes behaviour the code no longer has

`pkg/util/paramtable/component_param.go:3950-3956` and `configs/milvus.yaml:666`:

```
maxMsgNum: 8 # Maximum number of DML messages in one micro-batch drained by query node data sync pipeline...
```

`queryNode.dataSync.dmlMicroBatch.maxMsgNum` is an exported, refreshable knob. After this change it has no effect on inserts at all — only delete-only packs are batched. An operator raising it to improve insert-path throughput will observe no change and no explanation.

**Fix:** reword both sites to "Maximum number of delete messages in one micro-batch…" (and note that packs containing Insert messages are never merged). The two strings must be updated together or the yaml/param docs diverge.

### low — batching is now also off for upsert workloads; worth calling out in the PR body

`internal/util/pipeline/stream_pipeline.go:225,237`: an upsert produces a delete + insert pair on the same vchannel, which becomes one transaction and therefore one pack containing both message types. Such a pack is not delete-only, so it is neither merged nor used as a batch base — the delete micro-batching added in #50531 is effectively disabled for upsert-heavy streams, not just for the insert path.

This is a reasonable trade (correctness over throughput, and merging those packs carried the same hazard), but the PR description says only "retain bounded cross-pack batching for delete-only packs", which reads as if the delete win is preserved in general. Please state the upsert case explicitly. If the throughput matters later, a narrower rule — batch inserts only when all packs' insert messages carry an identical field-ID set — could restore it, at the cost of a per-pack field-set comparison.

### nit — predicate naming / empty-pack behaviour

`internal/util/pipeline/stream_pipeline.go:259-262`: `isDeleteOnlyMsgPack` returns `false` for an empty pack, i.e. "an empty pack is not delete-only". Behaviour is unchanged from `isDMLMsgPack` and is correct here (an empty pack must not start or extend a batch), but the name now reads as a pure content predicate. A one-line comment on the `len(pack.Msgs) == 0` branch ("empty packs must not start or extend a batch") would prevent a future refactor from "fixing" it.

### nit — comment placement

`internal/util/pipeline/stream_pipeline.go:223-225`: the new comment explains the insert exclusion but sits above a condition that also covers `maxMsgNum <= 1`. Moving it to directly precede `!isDeleteOnlyMsgPack(first)`, or onto `isDeleteOnlyMsgPack` itself (where the second call site at line 237 also benefits), keeps the rationale attached to the rule.

---

**Tests:** the rewrite is sound. `TestDMLMsgPackBatcherTreatsInsertPacksAsBarriers` correctly distinguishes "did not consume the next pack" (`assert.Len(t, input, 1)`) from "consumed and handed back as pending" (`assert.Same(t, test.next, pending)` + `assert.Empty(t, input)`) — that distinction is the actual contract and it is good to see it asserted. Both suite tests enqueue their packs before `Start()`, so the non-blocking `default:` drain is deterministic, not timing-dependent. One coverage gap worth a line: there is no case where a delete batch is followed by a **mixed** insert+delete pack (the upsert shape) — today it is handled by the same `!isDeleteOnlyMsgPack(next)` branch as case 2, but an explicit case would lock in the upsert behaviour discussed above.

chyezh:

Addressed in `3a280e01e7`:
- updated the `insert_node.go` invariant comment to state that the micro-batcher only merges delete-only packs and that an Insert never shares a pipeline batch with another `MsgPack`;

- the exported config wording is unchanged to keep this compatibility fix focused;
- merge-site `InsertRecord` shape validation is deferred to a separate change because logical row-count validation needs dedicated schema-aware handling and tests, especially for nullable and vector fields;
- the empty-pack and comment-placement nits do not affect behavior and are not included in this follow-up.

czs007 (approved):
/lgtm

PR #52363, opened 2026-08-10
fix: [cp 3.0] keep insert message packs separate

chyezh:
issue: #51612
pr: #52362

### DIFF
--- a/internal/querynodev2/pipeline/insert_node.go
+++ b/internal/querynodev2/pipeline/insert_node.go
@@ -53,12 +53,12 @@ func (iNode *insertNode) addInsertData(insertDatas map[UniqueID]*delegator.Inser
 		panic(err)
 	}
 	if len(skippedFields) > 0 {
-		// Attributing the skip to dropped fields is safe only because a SchemaChange
-		// message never shares a pack with inserts: the WAL adaptor emits every V1/V2
-		// message as its own pack, msgdispatcher never merges packs, and the DML
-		// micro-batcher refuses to merge non-Insert/Delete messages. If that
-		// pack-granularity invariant is ever relaxed, filtering against the current
-		// schema could silently drop fields that exist in a newer schema.
+		// Attributing the skip to dropped fields is safe because the WAL adaptor emits
+		// every V1/V2 message as its own pack, msgdispatcher never merges packs, and the
+		// DML micro-batcher only merges delete-only packs. An Insert message therefore
+		// never shares a pipeline batch with another MsgPack. If that pack-granularity
+		// invariant is ever relaxed, filtering against the current schema could silently
+		// drop fields that exist in a newer schema.
 		mlog.Warn(context.TODO(), "skip insert payload fields absent from current schema, fields are dropped since the message was written",
 			mlog.FieldCollectionID(iNode.collectionID),
 			mlog.FieldSegmentID(msg.SegmentID),
--- a/internal/util/pipeline/stream_pipeline.go
+++ b/internal/util/pipeline/stream_pipeline.go
@@ -220,7 +220,9 @@ func NewDMLMsgPackBatcher(maxMsgNum func() int) MsgPackBatcher {
 
 func (b *dmlMsgPackBatcher) Batch(first *msgstream.MsgPack, input <-chan *msgstream.MsgPack) (*msgstream.MsgPack, *msgstream.MsgPack) {
 	maxMsgNum := b.getMaxMsgNum()
-	if maxMsgNum <= 1 || !isDMLMsgPack(first) {
+	// Preserve original pack boundaries for inserts because insertNode merges
+	// records by segment without normalizing different schema field sets.
+	if maxMsgNum <= 1 || !isDeleteOnlyMsgPack(first) {
 		return first, nil
 	}
 
@@ -232,7 +234,7 @@ func (b *dmlMsgPackBatcher) Batch(first *msgstream.MsgPack, input <-chan *msgstr
 			if !ok {
 				return mergeMsgPacks(packs), nil
 			}
-			if !isDMLMsgPack(next) {
+			if !isDeleteOnlyMsgPack(next) {
 				return mergeMsgPacks(packs), next
 			}
 			if msgNum+len(next.Msgs) > maxMsgNum {
@@ -254,15 +256,13 @@ func (b *dmlMsgPackBatcher) getMaxMsgNum() int {
 	return b.maxMsgNum()
 }
 
-func isDMLMsgPack(pack *msgstream.MsgPack) bool {
+func isDeleteOnlyMsgPack(pack *msgstream.MsgPack) bool {
 	if pack == nil || len(pack.Msgs) == 0 {
 		return false
 	}
 
 	for _, msg := range pack.Msgs {
-		switch msg.Type() {
-		case commonpb.MsgType_Insert, commonpb.MsgType_Delete:
-		default:
+		if msg.Type() != commonpb.MsgType_Delete {
 			return false
 		}
 	}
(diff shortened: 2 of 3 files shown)


czs007:

Reviewed at head of `51612-insert_pack_batching-3.0` against `3.0`. **Clean backport — mergeable, no blocking issues.** The findings below are all carry-overs from my review of the master PR (#52362) and are hardening / doc-drift, not defects introduced by the cherry-pick.

**Backport fidelity — verified, not assumed:**
- `gh pr diff 52363` is byte-identical to `gh pr diff 52362`, including the pre-image blob hash (`138155891d3..5bb6b16b8f9` on both). Same base blob on both branches → the cherry-pick applied with zero adaptation, and there is no 3.0-specific divergence hiding in the context lines.
- Single commit, correct backport metadata (`issue: #51612`, `pr: #52362`, `[cp 3.0]` title).
- The surrounding 3.0 code the change depends on matches master: `streamPipeline.work()`'s `pending` hand-back and `mergeMsgPacks` position handling are unchanged (`internal/util/pipeline/stream_pipeline.go:62-99,268-290`), so the new barrier introduces no message loss or checkpoint regression on this branch either.
- Only one production caller of the batcher on 3.0 (`internal/querynodev2/pipeline/pipeline.go:64`) — no extra 3.0-only consumer (datanode / streamingnode) that would need the same treatment.
- No leftover references to the renamed `isDMLMsgPack` anywhere in the tree, and every test helper the new cases use (`newDeleteTsMsg`, `newDMLMsgPack`, `captureMsgPackNode`, `staticMVCCGetter`) already exists in the 3.0 test file — no compile risk.

The mechanism itself holds on 3.0: V1/V2 WAL messages are emitted one-per-pack, so the batcher was the only thing that could put two independently-written insert messages into one pack for the same growing segment — which is exactly the mixed-schema `InsertRecord` in #51612. `AddCollectionField` / `AlterCollection` packs were already barriers under the old predicate, so the two mergeable inserts were both post-DDL messages, one written by a proxy still holding the stale schema.

---

### medium — `insert_node.go` still documents the old, weaker batcher invariant (3.0 line numbers)

`internal/querynodev2/pipeline/insert_node.go:56-61`:

```go
// Attributing the skip to dropped fields is safe only because a SchemaChange
// message never shares a pack with inserts: ... and the DML
// micro-batcher refuses to merge non-Insert/Delete messages. If that
// pack-granularity invariant is ever relaxed, filtering against the current
// schema could silently drop fields that exist in a newer schema.
```

This comment is load-bearing — it is the stated justification for treating `skippedFields` as *dropped* fields rather than *newer-schema* fields. After this PR the rule is strictly stronger: the batcher merges only delete-only packs. Leaving the old wording documents "merging Insert packs is safe", which is precisely the belief this PR corrects, and invites a future change to re-widen the predicate.

**Fix:** reword to "…the DML micro-batcher only merges delete-only packs, so an Insert message is never merged with any other pack", referencing `isDeleteOnlyMsgPack`.

### medium — the corruption primitive in `addInsertData` remains unguarded

`internal/querynodev2/pipeline/insert_node.go:84-89`:

```go
err := typeutil.MergeFieldData(iData.InsertRecord.FieldsData, insertRecord.FieldsData)
...
iData.InsertRecord.NumRows += insertRecord.NumRows
```

`MergeFieldData` (`pkg/util/typeutil/schema.go:2319`) errors only for *src has a field dst lacks*. The reverse — dst carries the newly added nullable field, src does not — is silent: that field stays short while `NumRows` advances, producing the malformed record that panics later in `typeutil.AppendFieldData` far from the cause. The two orderings of the same mismatch therefore behave as `panic(err)` vs. silent corruption.

This PR removes the batcher as a *producer* of that state, but `addInsertData` still merges any two insert messages for the same segment within one `insertNodeMsg`, and a single pack can legitimately carry several insert messages (a transaction expands into multiple `TsMsg` in one pack; `VersionOld` messages sharing a time tick are grouped). Those messages come from one request and should share a schema version, so I am **not** claiming a live second trigger here — the point is that the invariant is now enforced only by the batcher predicate plus pack granularity, with nothing checking it at the merge site.

**Fix (cheap follow-up, master first):** after merging, assert each field's row count equals `iData.InsertRecord.NumRows` (or compare field-ID sets before merging) and fail loudly with collection / segment / schema-version in the message. That turns silent corruption + a distant panic into a local, diagnosable error.

### low — config doc now describes behaviour the code no longer has

`pkg/util/paramtable/component_param.go:3887-3894` and `configs/milvus.yaml:662-663`:

```
maxMsgNum: 8 # Maximum number of DML messages in one micro-batch drained by query node data sync pipeline...
```

`queryNode.dataSync.dmlMicroBatch.maxMsgNum` is exported and refreshable, and after this change it has no effect on inserts at all. Worth fixing on this branch in particular: the item is declared `Version: "3.0.0"`, i.e. the knob is new in this release line, so correcting the string now costs nothing in compatibility and avoids shipping a doc that never matched the code. An operator raising it for insert-path throughput would otherwise see no change and no explanation.

**Fix:** reword both sites together (they diverge otherwise) to "Maximum number of *delete* messages in one micro-batch…", and note that packs containing Insert messages are never merged.

### low — batching is now also off for upsert workloads; the PR body does not say so

`internal/util/pipeline/stream_pipeline.go:225,237`: an upsert produces a delete + insert pair on the same vchannel, which becomes one transaction and hence one pack containing both message types. Such a pack is not delete-only, so it neither starts nor extends a batch — the delete micro-batching is effectively disabled for upsert-heavy streams, not only for the insert path.

That is the right trade (merging those packs carries the same hazard), but the PR body says only "retain bounded cross-pack batching for delete-only packs", which reads as if the delete win is preserved in general. Please state the upsert case explicitly so the 3.0 release notes / perf comparisons are not surprising. If the throughput matters later, a narrower rule — batch inserts only when all packs' insert messages carry an identical field-ID set — could restore it at the cost of a per-pack field-set comparison.

### nit — empty-pack behaviour of the renamed predicate

`internal/util/pipeline/stream_pipeline.go:259-262`: `isDeleteOnlyMsgPack` returns `false` for an empty pack. Behaviour is unchanged from `isDMLMsgPack` and is correct (an empty pack must not start or extend a batch), but the new name reads as a pure content predicate, so a future refactor could "fix" it. A one-line comment on the `len(pack.Msgs) == 0` branch would prevent that.

### nit — comment placement

`internal/util/pipeline/stream_pipeline.go:223-225`: the new comment explains the insert exclusion but sits above a condition that also covers `maxMsgNum <= 1`. Attaching it to `isDeleteOnlyMsgPack` itself would also cover the second call site at line 237.

---

**Tests:** the rewrite is sound and ports cleanly. `TestDMLMsgPackBatcherTreatsInsertPacksAsBarriers` correctly separates "did not consume the next pack" (`assert.Len(t, input, 1)`) from "consumed and handed back as pending" (`assert.Same(t, test.next, pending)` + `assert.Empty(t, input)`) — that distinction is the actual contract. Both suite tests enqueue packs before `Start()`, so the non-blocking `default:` drain is deterministic rather than timing-dependent, which matters for CI stability on this branch. One coverage gap worth a line: no case covers a delete batch followed by a **mixed** insert+delete pack (the upsert shape). Today it falls into the same `!isDeleteOnlyMsgPack(next)` branch as case 2, but an explicit case would lock in the upsert behaviour above.

Since this is a straight cherry-pick, all five non-nit items are best fixed on master and backported together rather than diverging 3.0 here.

chyezh:

Addressed in `1a5bfb3b8b`:
- updated the 3.0 `insert_node.go` invariant comment to match the delete-only batching rule;

- the exported config wording is unchanged to keep this backport focused;
- merge-site `InsertRecord` shape validation should be implemented and validated on master first, then backported separately;
- the empty-pack and comment-placement nits do not affect behavior and are not included in this follow-up.

czs007 (approved):
/lgtm

PR #52364, opened 2026-08-10
fix: [cp 2.6] keep insert message packs separate

chyezh:
issue: #51612
pr: #52362

### DIFF
--- a/internal/util/pipeline/stream_pipeline.go
+++ b/internal/util/pipeline/stream_pipeline.go
@@ -225,7 +225,9 @@ func NewDMLMsgPackBatcher(maxMsgNum func() int) MsgPackBatcher {
 
 func (b *dmlMsgPackBatcher) Batch(first *msgstream.MsgPack, input <-chan *msgstream.MsgPack) (*msgstream.MsgPack, *msgstream.MsgPack) {
 	maxMsgNum := b.getMaxMsgNum()
-	if maxMsgNum <= 1 || !isDMLMsgPack(first) {
+	// Preserve original pack boundaries for inserts because insertNode merges
+	// records by segment without normalizing different schema field sets.
+	if maxMsgNum <= 1 || !isDeleteOnlyMsgPack(first) {
 		return first, nil
 	}
 
@@ -237,7 +239,7 @@ func (b *dmlMsgPackBatcher) Batch(first *msgstream.MsgPack, input <-chan *msgstr
 			if !ok {
 				return mergeMsgPacks(packs), nil
 			}
-			if !isDMLMsgPack(next) {
+			if !isDeleteOnlyMsgPack(next) {
 				return mergeMsgPacks(packs), next
 			}
 			if msgNum+len(next.Msgs) > maxMsgNum {
@@ -259,15 +261,13 @@ func (b *dmlMsgPackBatcher) getMaxMsgNum() int {
 	return b.maxMsgNum()
 }
 
-func isDMLMsgPack(pack *msgstream.MsgPack) bool {
+func isDeleteOnlyMsgPack(pack *msgstream.MsgPack) bool {
 	if pack == nil || len(pack.Msgs) == 0 {
 		return false
 	}
 
 	for _, msg := range pack.Msgs {
-		switch msg.Type() {
-		case commonpb.MsgType_Insert, commonpb.MsgType_Delete:
-		default:
+		if msg.Type() != commonpb.MsgType_Delete {
 			return false
 		}
 	}
--- a/internal/util/pipeline/stream_pipeline_test.go
+++ b/internal/util/pipeline/stream_pipeline_test.go
@@ -144,6 +144,7 @@ func (suite *StreamPipelineSuite) TestDMLMsgPackBatcherKeepsBufferedInsertPacksS
 		0,
 		false,
 		suite.channel,
+		nil,
 		staticMVCCGetter{},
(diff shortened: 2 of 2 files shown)


tinswzy:

Reviewed at head `da8605272e` (2 commits: cherry-pick `928dfcb72e` + 2.6 test adaptation `da8605272e`). **Mergeable — no blocking issues.** This is a faithful backport of #52362: the production hunks in `internal/util/pipeline/stream_pipeline.go` are identical to master, and the only 2.6-specific delta is the extra `replicateConfig` argument (`nil`) that `NewPipelineWithStream` takes on 2.6, correctly added to the new suite test in the adaptation commit.

I verified the fix's preconditions hold on 2.6, not just that the diff applies:

- The DML micro-batcher this narrows was backported to 2.6 in #50859 (`1d8fb77560`), and the schema-change-over-WAL path that makes cross-pack insert merging dangerous (`AddField` broadcast, `b36c88f3c8`) is also on 2.6 — so the #51612 failure mode is live on this branch and the backport is warranted.
- The corruption site is identical on 2.6: `internal/querynodev2/pipeline/insert_node.go:66` merges same-segment insert records via `typeutil.MergeFieldData` and `panic(err)`s on field-set mismatch, so before this fix a delayed pre-DDL insert merged with a post-DDL insert could crash the querynode (or silently mis-shape the record in the reverse ordering).
- `isDMLMsgPack` has no remaining references on 2.6 after the rename; the batcher's only production wiring is `internal/querynodev2/pipeline/pipeline.go:72`.
- The `pending` hand-back in `streamPipeline.work()` and `mergeMsgPacks` position handling are untouched, so no message-loss or checkpoint-order concern from the new barrier.
- All helpers the new tests use (`newDMLMsgPack`, `newInsertTsMsg`, `newDeleteTsMsg`, `captureMsgPackNode`, `staticMVCCGetter`) already exist in the 2.6 test file; the barrier test correctly distinguishes "did not drain the next pack" (`assert.Len(t, input, 1)`) from "drained and handed back as pending".

Findings (all non-blocking):

- **low — config docs now overstate what the knob does, on 2.6 too.** `pkg/util/paramtable/component_param.go:3704` and `configs/milvus.yaml:599` still describe `queryNode.dataSync.dmlMicroBatch.maxMsgNum` as "Maximum number of DML messages in one micro-batch" — after this change only delete-only packs are batched, and the knob has no effect on inserts. This was already flagged on #52362; whatever doc reword lands on master should be cherry-picked here in the same pass so the branches don't diverge.

- **low — upsert-shaped packs lose micro-batching entirely.** A delete+insert pair in one pack (the upsert/transaction shape) is not delete-only, so it is neither merged nor usable as a batch base (`stream_pipeline.go:230,242`). That partially walks back the delete-batching win 2.6.20 shipped via #50859 for upsert-heavy streams. It is the right correctness-over-throughput trade, but the PR body's "retain bounded cross-pack batching for delete-only packs" undersells the regression surface — worth one sentence in the description/release note for operators tracking 2.6.x behavior.

- **nit — the cherry-pick commit does not compile in isolation.** `928dfcb72e`'s new suite test calls the 7-argument 2.6 `NewPipelineWithStream` with 6 arguments; the missing `nil` only arrives in `da8605272e`. Harmless if this PR is squash-merged (the usual flow on this repo), but if it is ever rebase-merged the intermediate commit breaks `go vet`/bisect on 2.6. Folding the one-line adaptation into the cherry-pick would avoid that.

- **note — one upstream finding intentionally not applicable here:** the master review asked to update a load-bearing comment in `insert_node.go` about the old batcher invariant; that comment does not exist on 2.6, so nothing to change in this PR.

- **note — merge ordering:** upstream #52362 is still open. If review there produces further changes, please keep this cp in lockstep before merging so 2.6 doesn't ship a divergent variant of the fix.

chyezh:

Addressed in `a219a8e566`:

The `insert_node.go` invariant-comment finding is not applicable to 2.6 because that dropped-field filtering/comment does not exist on this branch.

- the exported config wording is unchanged to keep this backport focused;
- merge-site `InsertRecord` shape validation should be implemented and validated on master first, then backported separately;
- the intermediate-commit compile nit remains unchanged: the final PR head compiles and passes tests, and this repository normally squash-merges PRs; rewriting the existing backport commits is outside this follow-up;
- the upstream follow-up has been kept in lockstep with this backport.

czs007:

Verified:

- `internal/util/pipeline/stream_pipeline.go:226` — batching accepts only delete-only packs; Insert-only and mixed Insert/Delete packs are left unchanged and act as merge barriers.
- `internal/util/pipeline/stream_pipeline.go:62` — the `pending` buffer is local to the single worker goroutine and is drained before the next channel read, preserving ordering and exactly-once delivery.
- `internal/util/pipeline/stream_pipeline.go:234` — message-count limits remain enforced during merging; overflow packs are retained whole rather than split.
- `internal/util/pipeline/stream_pipeline.go:277` — merged delete packs preserve message order and keep the first start position and last end position.
- `internal/util/pipeline/stream_pipeline.go:238` — closing the input channel returns all accumulated delete packs without dropping data.
- `internal/util/pipeline/stream_pipeline_test.go:88` — tests cover delete merging, insert separation, exact/overflow limits, insert-first, delete-before-insert, delete-before-mixed-upsert, and mixed-first barrier cases.

czs007 (approved):
/lgtm

PR #52369, opened 2026-08-10
fix: [2.6] prevent streaming query nodes from blocking segment balance

weiliu1031:
issue: #52139
pr: #52148
pr: #52140

### What changed

Squash-backport both changes from the 2.6 fix into one hotfix commit.

- Separate channel outbound detection from sealed-segment outbound
  detection.
- Continue normal segment balancing when outbound evacuation produces no
  executable plan.
- Propagate analyze storage MaxConnections.
- Retain only the scalar connection limit inside the Mockey callback.
- Include the Streaming Service regression coverage and design document
  update.

### Why

With Streaming Service enabled, valid RWSQ channel holders were treated
as sealed-segment outbound nodes. The resulting empty evacuation path
permanently skipped normal segment balancing.

The analyze task also failed to propagate the requested storage
connection limit.

### Branch adaptation

hotfix-2.6.22 contains one hotfix-only proxy cache commit and lacks six
later commits from the source 2.6 branch.

Both source commits apply without conflicts. They are squashed into one
commit without changing the combined patch.

### Validation

- `git diff --check`
- Combined stable patch ID:
  `b0b564f72612d5ef82465cf5f9eb8fec70dce824`
- No hotfix-specific Go or C++ tests were run. The source PR passed CI,
  and this backport preserves the source patch unchanged.

### DIFF
--- a/docs/design-docs/design_docs/20260204-channel_exclusive_mode.md
+++ b/docs/design-docs/design_docs/20260204-channel_exclusive_mode.md
@@ -340,14 +340,16 @@ BalanceReplica(ctx, replica)
 3. Channel-Aware Balancing (Exclusive Mode Enabled)
     For each channel:
         ├─ Get channel's exclusive RW nodes
-        ├─ Identify outbound nodes (nodes no longer in RW node list)
-        ├─ If outbound nodes exist:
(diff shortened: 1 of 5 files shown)


liliu-z (commented):
/approve

PR #52374, opened 2026-08-10
enhance: Add optimistic CAS for partial updates

weiliu1031:
issue: #49980

## Design review

- Primary approver: @chyezh
- Independent approver: @liliu-z
- Review date: 2026-08-07

## What changed

- Keep partial-update query and merge execution in Proxy.
- Resolve PChannel terms before allocating an attempt-scoped read timestamp.
- Carry only `read_ts` and `observed_pchannel_term` in the encrypted CAS
  attempt proof.
- Require every CAS Insert to carry an explicit schema version.
- Derive collection identity from the Insert header and resolve the
  authoritative PK descriptor through ShardManager.
- Reject transactions whose CAS chunks disagree on proof, collection, or
  schema.
- Serialize CAS validation, WAL append, index publication, and transaction
  transition with the existing vchannel lock.
- Rebuild safe REPLACE attempts after deterministic CAS rejection.
- Preserve existing transaction recovery and replication behavior.
- Release partial-update proof state after a durable DropCollection.

## Why

Concurrent partial updates can read the same row snapshot and subsequently
overwrite fields committed by another writer. The commit-side admission check
rejects a mutation when the PK changed after the Proxy snapshot.

Collection and PK identity are derived on StreamingNode instead of duplicated
in Proxy-generated proof. This prevents metadata from selecting a collection
or field that differs from the actual Insert message and schema.

## Behavior boundaries

- AutoID partial updates use update-only semantics: every supplied PK must
  exist in the query snapshot, and the merged insert preserves that PK.
- ARRAY_APPEND and ARRAY_REMOVE conflicts are non-retriable.
- Cross-vchannel atomicity and transaction-level exactly-once are unchanged.
- Concurrent Import, RestoreSnapshot, and backfill are outside this change.

## Design

- [Partial update optimistic lock design](docs/design-docs/design_docs/20260601-partial-update-optimistic-lock-design.md)
- [Protocol and design PR](https://github.com/milvus-io/milvus/pull/52235)

## Validation

- Full partialupdate package test with race detection, dynamic/test tags,
  disabled optimizations, and at least 99% coverage
- Targeted message, producer, and Proxy tests with dynamic/test tags and
  disabled optimizations
- `make milvus`
- `git diff --check`


internal/streamingnode/server/wal/interceptors/partialupdate/index.go
```diff
@@ -0,0 +1,424 @@
...
+	"github.com/milvus-io/milvus/pkg/v3/util/tsoutil"
+)
+
+const (
+	defaultVersionIndexTTL = 30 * time.Second
+
+	// The estimate covers the map key/value, heap pointer, entry object, and
+	// allocator overhead. String backing bytes are accounted separately.
+	estimatedVersionEntryFixedBytes int64 = 128
+	defaultVersionIndexBudgetBytes        = 5_000_000 * estimatedVersionEntryFixedBytes
```
chyezh: [P1] Bound the PK-version index memory at the StreamingNode level

`defaultVersionIndexBudgetBytes` is instantiated independently for every PChannel/WAL term, while the interceptor is installed for every WAL and records PKs from ordinary Insert/Delete traffic even when no CAS request is used. With the default 16 PChannels on one StreamingNode, this permits roughly 10.24 GB (9.54 GiB) of estimated index memory before accounting for Go map/heap/object and GC overhead. A high-cardinality ordinary write workload can therefore OOM the node and trigger channel failover.

Please make this a configurable node-wide shared budget (or divide a node-level limit among its assigned PChannels), expose aggregate used/limit/missed-write metrics, and add a multi-WAL capacity/RSS test covering Int64 and long VarChar PK churn without CAS traffic.

internal/streamingnode/server/wal/interceptors/partialupdate/interceptor.go
```diff
@@ -0,0 +1,179 @@
...
+				return nil, status.NewUnrecoverableError("partial update collection fence has empty collection id")
+			}
+			fenceCollectionID = collectionID
+		} else if collectionID, ok := extractDropCollectionID(msg); ok {
+			if collectionID == 0 {
+				return nil, status.NewUnrecoverableError("partial update dropped collection has empty collection id")
+			}
+			droppedCollectionID = collectionID
+		} else if msg.MessageType() == message.MessageTypeInsert {
+			pks, fenceCollectionID, err = extractPKsFromOrdinaryInsert(msg, i.pkDescriptorGetter)
```
liliu-z: The new partialupdate interceptor is added to every WAL unconditionally, and its write path decodes the full body of every insert and every delete to extract primary keys, then boxes one `any` per row into the version index — even for clusters and collections that never issue a partial update. Before this PR no interceptor decoded DML bodies on the append path (only DDL bodies), and for encrypted collections `Body()` now also forces a full payload decrypt per message. This puts a full proto.Unmarshal of the entire insert payload (vectors included), plus a per-row allocation, inside the WAL append critical section.

internal/streamingnode/server/wal/interceptors/partialupdate/pk_extract.go
```diff
@@ -0,0 +1,321 @@
...
+// partial-update insert chunk.
+func extractPKsFromInsert(msg message.MutableMessage, fieldID int64) ([]any, error) {
+	if msg == nil || fieldID <= 0 {
+		return nil, status.NewUnrecoverableError("partial update insert primary key field id is invalid")
+	}
+	insertMsg, err := message.AsMutableInsertMessageV1(msg)
+	if err != nil {
+		return nil, status.NewUnrecoverableError("decode partial update insert message failed: %v", err)
+	}
+	body, err := insertMsg.Body()
```
liliu-z: For cipher-enabled collections this new PK extraction is the first thing on the WAL append path to call `Payload()` on a DML body, and `Payload()` panics when a decryptor cannot be obtained. A non-KMS-key-invalid decryptor error is not retried, so what used to be an append that never touched the ciphertext now crashes the whole streamingnode — taking down every WAL it hosts — instead of failing the one request.

internal/streamingnode/server/wal/interceptors/partialupdate/interceptor.go
weiliu1031: The production path keeps Int64 and VarChar PKs in typed slices, avoiding per-row `any` boxing.

We intentionally use the generated protobuf codec to decode complete Insert/Delete bodies, including vectors. This accepts the CPU, allocation, GC, and append-latency cost in exchange for avoiding a custom wire parser and keeping compatibility with generated protobuf definitions.

internal/streamingnode/server/wal/interceptors/partialupdate/index.go
weiliu1031: Fixed by sharing one atomic PK-version budget across all WALs on a StreamingNode.

The limit is configurable through `streaming.partialUpdate.versionIndexMaxBytes`, defaulting to 640,000,000 bytes. Int64 and VarChar PK sizes are accounted for, and TTL eviction, DropCollection, and WAL close release the charge.

When the budget is exhausted, ordinary writes continue, but CAS verification fails closed with a retryable result. Node-level used, limit, and missed-write metrics are exposed.

Deterministic shared-budget tests are included; a process-level RSS benchmark is not included in this PR.

internal/streamingnode/server/wal/interceptors/partialupdate/pk_extract.go
weiliu1031: Fixed by replacing the panic-based `Payload()` call in the partial-update append path with `message.DecodePayload(ctx)`.

Cipher lookup, decryptor, decryption, and cancellation failures now return errors for the current request instead of panicking the StreamingNode. Tests cover these failure paths.

PR #52375, opened 2026-08-11
enhance: unify request timezone resolution for search and query

czs007:
Follow-up to #52113.

### What changed

Extract the request-level timezone resolution ("request `timezone` param → validate → fall back to collection timezone → UTC") into a single shared helper `resolveTimezone` in `internal/proxy/util.go`, and make both `searchTask.PreExecute` and `queryTask.PreExecute` use it. Remove the now-redundant `timezone` field from `queryParams` and its extraction/validation in `parseQueryParams`.

### Why

The bug fixed by #52113 existed because this resolution logic was inlined twice (search and query) and the two copies drifted independently: the query copy sat before plan generation, the search copy sat after it, so wiring the timezone into the search plan parser (#46546) silently read the zero value. A single helper with an explicit "must be resolved before plan generation" contract removes the duplication that made that drift possible.

### Behavior notes

- Search/query with a valid, absent, or invalid timezone behave exactly as before (same error code/message for invalid IANA IDs).
- One degenerate case is unified: a search request that explicitly passes an **empty** `timezone` value previously resolved to UTC (empty string reached `time.LoadLocation("")`); it now falls back to the collection timezone, matching the existing query semantics ("empty = unspecified").
- For query, the invalid-timezone error is now raised at the resolution point in `PreExecute` instead of inside `parseQueryParams`; both are before plan generation, and the error code and message are unchanged.

### Verification

```bash
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/proxy
```

including the #52113 regression test `TestSearchTaskPreExecuteUsesTimezoneForTimestamptzFilter` and the new `TestResolveTimezone` unit tests (request wins / invalid rejected as ParameterInvalid / absent falls back to collection / empty treated as unspecified / default UTC).

### DIFF
--- a/internal/proxy/task_query.go
+++ b/internal/proxy/task_query.go
@@ -35,7 +35,6 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 	"github.com/milvus-io/milvus/pkg/v3/util/retry"
 	"github.com/milvus-io/milvus/pkg/v3/util/timerecord"
-	"github.com/milvus-io/milvus/pkg/v3/util/timestamptz"
 	"github.com/milvus-io/milvus/pkg/v3/util/tsoutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
@@ -104,7 +103,6 @@ type queryParams struct {
 	collectionID        int64
 	groupByFields       []string
 	orderByFields       []string // NEW: ORDER BY field specifications (e.g., "price:desc")
-	timezone            string
 	extractTimeFields   []string
 	queryIteratorCursor *planpb.QueryIteratorCursor
 }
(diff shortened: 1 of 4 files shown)


xaxys:

Verified:
- `internal/proxy/util.go:3356-3368` — the timezone resolution helper correctly enforces request-parameter precedence, rejects invalid non-empty IANA IDs, and falls back to collection timezone or UTC.
- `internal/proxy/task_query.go:727-775` — query parameters are parsed before timezone resolution, and no stale `queryParams.timezone` reference remains after the refactor.
- `internal/proxy/task_search.go:265-283` — the timezone is resolved before both normal and advanced search request initialization, so plan generation receives the resolved value in every path.
- `internal/proxy/task_query.go:1060-1069` — result extraction and formatting use the same resolved timezone as plan parsing, preserving the invariant end to end (same holds at `internal/proxy/task_search.go:1464-1474`).
- `internal/proxy/util.go:3359-3362` — an invalid request timezone still surfaces as a typed `ErrParameterInvalid` with the previous client-visible message, and collection timezone validity is enforced at the create/alter boundaries.
- `internal/proxy/util_test.go:6919-6959` — unit coverage exercises request precedence, invalid-input classification, absent/empty parameters, collection fallback, and the UTC default, with an integration regression test at `internal/proxy/task_search_test.go:122-185` confirming the resolved timezone is applied when parsing a naive TIMESTAMPTZ filter.

Issue #52319, opened 2026-08-07
[Feature]: FunctionChain iteration roadmap: full stage coverage, complete operators, function_score convergence, UDF support


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Part of #46565.

FunctionChain has shipped its first two execution paths — L2 rerank at the proxy (#50571) and L0 XGBoost rerank at the querynode — but the current surface is still a small subset of the #46565 vision. This issue tracks the remaining iteration work as a roadmap, based on a code survey of the current master.

**What works today (single-collection search only):**
- L2_RERANK: proxy-side chain over the merged result set; all six registered operators (`map`/`filter`/`select`/`sort`/`limit`/`group_by`) are accepted; expressions: `decay`, `num_combine`, `round_decimal`, `rerank_model`.
- L0_RERANK: querynode-side per-segment chain, **`map`-only** (`l0_function_chain.go`), `$score` as the only writable output; expressions: `xgboost` (cgo kernel), `decay`, `round_decimal`.
- REST v2 plumbing, one-chain-per-stage rules, `$`-prefixed system-field protection.

#### 1. Full stage coverage

- [ ] **L1_RERANK** — currently explicitly rejected (`function_chain_validator.go`). Execute after segment-merge at the querynode to cut candidate volume shipped to the proxy.
- [ ] **Hybrid search support** — `HybridSearchRequest.function_chains = 15` exists in the proto but the proxy rejects it outright. Support the top-level (post-fusion L2) chain first.
- [ ] **Per-sub-request chains** — `SubSearchRequest` has no `function_chains` field yet; add it so each ANN sub-request can carry its own L0/L1 chain (per the #46565 API sketch with `AnnSearchRequest(..., function_chains=[...])`).
- [ ] **INGESTION** — no execution path exists. Requires schema-attached (persistent) chains rather than request-scoped ones; target ETL pipelines like `chunk → embed → select`, converging with the existing schema `Function` (BM25 / text-embedding) machinery.
- [ ] **PRE_PROCESS / POST_PROCESS** — request/response refinement hooks; currently enum-only.

#### 2. Complete the operator & expression surface

- [ ] Lift the **L0 `map`-only restriction** — at minimum `filter` (threshold pruning before results leave the segment) and `sort`/`limit` are natural at L0.
[...]

PR #52376, opened 2026-08-11
feat: Support L1 rerank

junjiejiangjjj:
doc: docs/design-docs/design_docs/20260624-function-chain-api.md
#52319

### DIFF
--- a/docs/design-docs/design_docs/20260624-function-chain-api.md
+++ b/docs/design-docs/design_docs/20260624-function-chain-api.md
@@ -11,13 +11,14 @@
 
 Function Chain introduces a typed, ordered, stage-aware pipeline for scoring and reranking search results. A chain is sent as structured protobuf, not as an opaque JSON string, so Milvus can validate dependencies, fetch required fields internally, execute built-in scoring functions, and project final search results back through the existing result schema.
 
-The first release focuses on ordinary `SearchRequest` L2 rerank:
+Ordinary `SearchRequest` supports three rerank stages at distinct execution boundaries:
 
-- Users build an L2 chain in SDKs and pass it through `function_chains`.
-- Proxy validates the request, plans required rerank input fields, and reuses the existing rerank pipeline.
-- The function-chain runtime executes `map`, `sort`, and `limit` operators on a DataFrame converted from reduced search results.
+- L0 executes independently on each segment result in the worker QueryNode before cross-segment reduction.
+- L1 executes on each worker QueryNode after its cross-segment reduction and before results are returned to the shard leader.
+- L2 executes in Proxy after distributed/global reduction and reuses the existing rerank pipeline.
+- L0 supports `map`; the first L1 release supports `map`, `sort`, and `limit`; L2 uses the generic function-chain runtime operators allowed by validation.
 - Final `$score` is serialized through the existing search score/distance field.
-- Intermediate variables and internally fetched fields are not returned unless requested by normal search output projection.
+- Intermediate variables, internally fetched fields, and internal provenance columns are not returned unless requested through normal search output projection.
 
 ## Motivation
 
@@ -43,8 +44,9 @@ Representing this as typed operations gives Milvus:
 
 - Add public protobuf messages for a function-chain logical plan.
 - Add SDK builder APIs that compile to the protobuf plan.
-- Support ordinary Search L2 rerank through `SearchRequest.function_chains`.
-- Reuse the existing Proxy rerank pipeline rather than adding a separate search pipeline operator.
+- Support ordinary Search L0, L1, and L2 rerank through `SearchRequest.function_chains`.
+- Execute L0 per segment, L1 after each worker QueryNode reduction, and L2 after Proxy global reduction.
+- Reuse the existing Proxy rerank pipeline for L2 rather than adding a separate search pipeline operator.
 - Fetch function-chain-required schema fields internally even when users do not request them in `output_fields`.
 - Keep final search response projection Search-owned.
 - Support first-version built-in expressions:
@@ -58,15 +60,16 @@ Representing this as typed operations gives Milvus:
 
 The first release does not include:
 
-- `function_chains` support for hybrid search execution;
+- `function_chains` support for hybrid or advanced search execution;
 - insert/upsert/ingestion function chains;
-- L0/L1 pushdown to QueryNode or Segcore;
+- L1 execution at the shard-leader reduction boundary or at multiple QueryNode reduction levels;
+- L1 `filter`, `select`, or `group_by` operators;
 - arbitrary user-defined expression language;
 - returning intermediate chain variables as user-facing result fields;
 - replacing `function_score` or legacy rank parameters;
 - client-side execution of external model calls.
 
-The public stage enum reserves room for future stages, but ordinary Search initially accepts only `L2_RERANK`.
+The public stage enum also reserves room for ingestion, preprocessing, and postprocessing. Ordinary Search accepts at most one chain at each of `L0_RERANK`, `L1_RERANK`, and `L2_RERANK`.
 
 ## Public Interfaces
 
@@ -148,11 +151,13 @@ client.search(..., function_chains=chain)
 client.search(..., function_chains=[chain])
 ```
 
-SDK and server validation reject ambiguous combinations:
+SDK and server validation reject ambiguous or unsupported combinations:
 
 - `function_chains` with SDK `ranker` / proto `function_score`;
-- non-L2 chains for ordinary Search;
-- `function_chains` for hybrid search in the first release.
+- stages other than L0, L1, and L2 for ordinary Search;
+- `function_chains` for hybrid or advanced search;
+- Function rerank with Search Iterator or `order_by`;
+- L1 with search aggregation.
 
 ### Protobuf
 
@@ -253,7 +258,32 @@ Representation:
 | Runtime | score register / DataFrame column |
 | Search result | existing distance/score field |
 
-`$id` is also available as a read-only system value for tie-breaking. Other `$xxx` system names are rejected in first-version L2 rerank.
+`$id` is also available as a read-only system value for tie-breaking. For public L0 and L1 chains, `$id` and `$score` are the only readable system columns and `$score` is the only writable system column. Internal columns used for segment offsets, grouping, element metadata, or L1 provenance are not part of the public chain namespace and must never be exposed in result fields.
+
+### Stage execution semantics
+
+The stages form one distributed rerank pipeline:
+
+```text
+segment ANN result
+  -> L0 per-segment chain
+  -> worker QueryNode cross-segment reduce / PK dedup / group-aware merge
+  -> L1 per-worker merged-candidate chain
+  -> shard-leader reduce
+  -> Proxy global reduce
+  -> L2 Proxy chain
+  -> final projection
+```
+
+L1 runs exactly once for each worker QueryNode's merged result. It does not run again at the shard leader. This boundary lets L1 compare candidates across segments handled by one worker while preserving the existing distributed reducer and wire protocol.
+
+| Stage | Execution boundary | First-version operators |
+|-------|--------------------|-------------------------|
+| L0 | Each segment result in worker QueryNode, before cross-segment reduce | `map` |
+| L1 | Each worker QueryNode result, after cross-segment reduce | `map`, `sort`, `limit` |
+| L2 | Proxy result, after distributed/global reduce | Generic runtime operators accepted by stage validation |
+
+A function used by an operator must also declare that it is runnable at that stage. For example, an expression restricted to L0 or L2 is not made valid in L1 merely because it appears in an allowed `map` operator.
 
 ### Operators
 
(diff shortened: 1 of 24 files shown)


internal/querynodev2/tasks/l1_function_chain.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		if int64(arr.Len()) != chunkSizes[chunkIdx] {
+			return nil, merr.WrapErrServiceInternalMsg("l1_rerank: provenance chunk %d has %d rows, expected %d", chunkIdx, arr.Len(), chunkSizes[chunkIdx])
+		}
+		result[chunkIdx] = make([]segmentSource, arr.Len())
+		for row := 0; row < arr.Len(); row++ {
+			if arr.IsNull(row) {
+				return nil, merr.WrapErrServiceInternalMsg("l1_rerank: provenance token is null at chunk %d row %d", chunkIdx, row)
+			}
+			idx := arr.Value(row)
```
liliu-z, line 296: The L1 function chain applies its scope resolution per search scope, but the source chunk indices resolved at `result[chunkIdx][row] = sourceChunk[int(idx)]` are not consistently scoped, so rows can be projected against the wrong scope's source data. This breaks the invariant that a token index is only meaningful within the chunk it was produced for; when a search spans multiple scopes the rerank stage consumes mismatched columns and returns wrong scores/entities to the user.

liliu-z, line 296: When requests with differing TopK values are batched together, the chain assumes a uniform per-request result width when slicing the source chunks, so rows are attributed to the wrong request. This is reachable in production via `query_hook.go:102`, which can rewrite TopK per request inside a batch, so a batched search with heterogeneous TopK silently returns results belonging to another request rather than failing loudly.

internal/proxy/task_search.go
```diff
@@ -908,6 +908,10 @@ func (t *searchTask) initSearchRequest(ctx context.Context) error {
 			t.rerankMeta = meta
 		}
 		querynodeFunctionChains = qnChains
+
+		if hasFunctionChainStage(qnChains, schemapb.FunctionChainStage_FunctionChainStageL1Rerank) && t.aggCtx != nil {
```
liliu-z, line 903: The guard added at `task_search.go:922` only rejects function rerank for search iterator v2, but `} else if t.isIterator && queryInfo.GetSearchIteratorV2Info() == nil {` at line 1084 confirms v1 is a live path using the same score-as-continuation-bound mechanism; `t.isIterator` isn't assigned until line 936, after the guard, so v1 structurally cannot be caught by it. Since function rerank rewrites the final score that v1 uses as the ANN continuation bound, a v1 iterator with function rerank will skip or duplicate entities across pages.

internal/querynodev2/tasks/l1_function_chain.go
liliu-z, line 296: Temporary columns produced for L0 processing are not tracked/cleaned in the chain's column bookkeeping, so downstream stages read columns that either no longer correspond to the current result set or were never populated. Under a realistic L0-involved search this yields wrong rerank inputs and user-visible incorrect ordering or missing fields.

liliu-z, line 296: Failures in the chain are wrapped with `merr.WrapErrServiceInternalMsg(...)`, which also swallows context cancellation/deadline paths and reports them as an internal service error. This breaks the Input-vs-System classification and makes a client-initiated cancel or timeout look like a Milvus bug, polluting error metrics and misleading operators debugging normal cancellations.

liliu-z, line 296: The rewritten mixed-TopK test constructs its inputs in a way that skips the production fast path, so it never exercises the code path where the mixed-TopK batching bug actually lives. This is regression-coverage debt rather than a runtime defect, but it means the accompanying fix will not be protected against reintroduction.

junjiejiangjjj: Thanks for calling this out. I traced the provenance path and the token is scoped to an individual DataFrame/NQ chunk: token generation restarts per chunk, and rebuild resolves each token against `original[chunkIdx]`, with a chunk-local bounds check. `segmentSource.InputIdx` identifies the input segment result, not a search-scope index.

I added a two-chunk regression test where both chunks retain the same local token after L1 sort/limit and verified that each token resolves to its own chunk source. The final provenance column is also removed as expected. The test passes, so I did not find a production source-mapping defect here.

junjiejiangjjj: The query-hook reachability is valid: requests with different outer TopKs can be normalized to an identical serialized plan and then merged. However, the result path does not derive request boundaries from a uniform flattened width; it tracks `originNqs`/`originTopks`, runs the shared reduce/L1 path, and slices both the DataFrame and `Sources` by NQ chunks.

I added a production-lifecycle regression test using original `(NQ, TopK)` values `(1,3)` and `(2,8)`. It runs the real query-hook normalization, `SearchTask.Merge`, `PreExecute`, `Execute`, L1 rerank, and per-request slicing. It verifies the original response TopKs/cardinality and L1 execution. The test passes, so the request/source corruption described here was not reproduced.

internal/proxy/task_search.go
junjiejiangjjj: Good catch. Legacy iterator v1 is implemented client-side with range-search pagination: it uses the previous page tail score as the next ANN range boundary. Therefore function rerank has the same score-domain conflict as v2, even though v1 and v2 carry the continuation state differently.

I changed the validation to reject function rerank for any search iterator, using the local `isIterator` value returned by plan generation. I also added v1 coverage for FunctionScore and L0/L1/L2 function chains, while retaining the existing v2 cases. The full proxy test suite passes.

internal/querynodev2/tasks/l1_function_chain.go
junjiejiangjjj: I traced the column lifecycle and did not find an untracked L0 temporary-column path. Public L0 chains may write only `$score`, execute with column pruning enabled, and Go reduction constructs a new result containing only reduce-system/output columns. L1 also enables pruning and performs a final downstream-column projection.

I added an L0 -> reduce -> L1 composition test that checks the L0 input/unused columns immediately after L0, then verifies that the L1 materialized input, L1 intermediate, and provenance column are absent from the final result while IDs/scores remain correct. The test passes.

junjiejiangjjj: I checked this against the error-handling contract. Cancellation-capable L1 calls are wrapped with `merr.Wrap`, which preserves both the unwrap chain and the inner classification. The `WrapErrServiceInternalMsg` calls in this file originate local invariant failures (missing/misaligned provenance, invalid tokens, etc.) and do not wrap a context error.

I added canceled-context and expired-deadline tests at the `applyL1Rerank` boundary. They assert both `errors.Is` and `merr.Code` (`CanceledCode` / `TimeoutCode`), specifically guarding against an accidental ServiceInternal relabel. Both cases pass.

junjiejiangjjj: Agreed. The existing test manually constructs post-merge state and directly exercises the per-slice reducer, so it is useful focused coverage but does not protect the real batching/L1 lifecycle.

I kept that focused test and added a separate full-path test that performs real query-hook plan normalization, `SearchTask.Merge`, `PreExecute`, `Execute`, shared Go reduce, L1 rerank, and final per-request slicing for heterogeneous TopKs. It verifies that the L1 chain survives optimization and executes, and that each subtask receives its original NQ/TopK/cardinality. The full QueryNode task suite passes.

liliu-z:
**search_task.go:340** Batching search requests with differing TopK values into one grouped task changes which candidates reach the L1 rerank stage: the merged batch is scored and truncated as a unit, so a request's candidate set depends on the other requests it happens to be batched with rather than on its own TopK. This breaks the invariant that a request's rerank input is a deterministic function of that request, so identical queries return different top results depending on concurrent traffic, and results are non-reproducible. Either exclude mixed-TopK requests from a batch or apply L1 per-request against its own candidate slice.

liliu-z:
**search_task.go:350** Sealed and growing segments are dispatched as separate search tasks that land on the same worker, and each task independently runs the L1 function chain and applies `limit`, so L1 executes twice over two disjoint candidate sets that are only merged afterwards. L1 rerank is defined over the full candidate set for a request; running it per-segment-type and truncating each side to `limit` drops candidates that would have survived a single global rerank, and doubles rerank cost. Under any load where a collection has both sealed and growing data, recall silently degrades relative to a full-load-then-rerank path. L1 should run once after sealed/growing results are merged.

liliu-z:
**l0_function_chain.go:172** The tightened validation here rejects L0 function chains whose inputs reference temporary/intermediate columns produced earlier in the chain, but such chains were accepted before this change. Existing collections and stored chain definitions were built against the looser contract, so this is a backward-incompatible narrowing rather than a new-input-only restriction. After upgrade, previously working search requests (and any persisted chain definitions replayed at load time) start failing with a validation error even though nothing about the user's configuration changed. Either keep accepting temporary columns, or gate the stricter check to newly created chains and document the break.

liliu-z:
**search_result_export_c.cpp:941** The catch-all on this shared late-materialization path converts every exception — including a query cancellation/timeout — into `UnexpectedError`, destroying the cancellation code before it crosses the cgo boundary. Cancellation is a distinct, expected, client-driven outcome; collapsing it into a generic internal error means the Go side cannot distinguish user cancel/timeout from a real segcore bug, so cancelled searches surface as internal errors in logs and to the caller and pollute error metrics. Catch and re-raise the cancellation/`SegcoreError` case with its original code before the generic handler.

liliu-z:
**search_result_export_c_test.cpp:1383** The newly added nullable case cannot pass as written: there are two independent defects, one in the setup/expectation at this line and a second at :1410, so the test either fails or asserts the wrong thing about nullable column export. A test that cannot pass gives no coverage for the nullable path this PR is adding, and leaves the actual nullable export behavior unverified. Both defects need fixing before merge — reviewers disagreed only on the severity label, not on whether it blocks.

liliu-z:
**20260624-function-chain-api.md:539** This section still documents the old iterator-v2 contract, which no longer matches the behavior after this change; the code is correct and should not be reverted, but the stale wording will mislead anyone implementing against the documented contract. The widened L0 chain rejection also changes visible upgrade behavior, so it needs a release-note line rather than only a design-doc edit.

internal/querynodev2/tasks/l1_function_chain.go
junjiejiangjjj: We traced the provenance path. The source token is generated per DataFrame/NQ chunk, and each token is resolved against the corresponding source chunk during rebuild. Tokens are never resolved across scopes, so the provenance remains scope-local.

junjiejiangjjj: Fixed. Mixed-TopK requests with L1 now use per-request reduction before reranking, so each request uses its own candidate window. Added TestExecuteMergedSubTasks_MixedTopKWithL1RerankMatchesIndependentRequests, which passes.

internal/proxy/task_search.go
junjiejiangjjj: Fixed. Function rerank is now rejected for both legacy Search Iterator and Iterator v2. Added coverage for function score, L0 chains, and L1 chains on both iterator versions.

internal/querynodev2/tasks/l1_function_chain.go
junjiejiangjjj: We traced the column lifecycle and did not find an untracked temporary-column path. L0 temporary outputs are preserved through the function-chain DataFrame, and the reduction stage builds the downstream DataFrame from the current result columns. No additional fix was needed.

junjiejiangjjj: We verified the cancellation paths. Cancellation-capable operations use merr.Wrap, which preserves the underlying cancellation error and its classification. The internal-error wrappers are used only for actual invariant or setup failures.

junjiejiangjjj: Addressed. The previous focused test was retained, and we added a production-path differential test that executes merged mixed-TopK requests and compares each sub-request with an equivalent independent request. The targeted and full task tests pass.

liliu-z (commented):
/lgtm
/approve

docs/design-docs/design_docs/20260624-function-chain-api.md
```diff
@@ -11,13 +11,14 @@
...
 
-The first release focuses on ordinary `SearchRequest` L2 rerank:
+Ordinary `SearchRequest` supports three rerank stages at distinct execution boundaries:
 
-- Users build an L2 chain in SDKs and pass it through `function_chains`.
-- Proxy validates the request, plans required rerank input fields, and reuses the existing rerank pipeline.
-- The function-chain runtime executes `map`, `sort`, and `limit` operators on a DataFrame converted from reduced search results.
+- L0 executes independently on each segment result in the worker QueryNode before cross-segment reduction.
+- L1 executes on each worker QueryNode after its cross-segment reduction and before results are returned to the shard leader.
+- L2 executes in Proxy after distributed/global reduction and reuses the existing rerank pipeline.
```
zhengbuqian, line 18: we don't have a L1.5 where we do shard level, right?

zhengbuqian (commented):
/approve

approval only for the design doc change. I did not review the code changes.

docs/design-docs/design_docs/20260624-function-chain-api.md
junjiejiangjjj: Right. There is no L1.5 at the shard-leader boundary. L1 is worker-local, and the shard leader only reduces the worker results.

zhengbuqian: do we plan to have one? of should we update the naming so we can possiblely add that if we wanted.

PR #52377, opened 2026-08-11
fix: add user authentication and RBAC authorization for stream gRPC

Amos-yongxin:
## Background
Streaming RPCs (e.g. CreateReplicateStream, DumpMessages) were not going through the unary interceptor chain on the external client-facing gRPC port. Only Authentication (credential verification) was added to the stream path, RBAC authorization was missing, so authenticated users could bypass role-based access checks for stream-type resource operations.

## Changes
1. Authentication: add proxy.GrpcAuthStreamInterceptor that validates username+password (bcrypt/SHA-256) on external streaming calls:
   - reject missing Authorization metadata
   - reject bad credentials
   - accept valid credentials and propagate username/sha256_password via contextutil.SetAuthInfo. When CommonCfg.AuthorizationEnabled is false the behavior is skipped, consistent with the unary chain. Register the stream interceptor in startExternalGrpc for the external gRPC server with explicit comment explaining why stream must be wired separately (streaming RPCs bypass the unary chain).
2. RBAC authorization: add the stream counterpart of proxy.UnaryServerInterceptor(proxy.PrivilegeInterceptor), i.e. proxy.StreamServerInterceptor(proxy.PrivilegeInterceptor). For stream RPCs described in privilege.Ext (CreateReplicateStream and DumpMessages), the interceptor runs Casbin Enforce against Global-default
   + PrivilegeUpdateReplicateConfiguration so role bindings correctly allow/deny stream calls. UTs verify deny-blocks-handler path and the permit path that propagates the authorized context.

### DIFF
--- a/internal/proxy/authentication_interceptor.go
+++ b/internal/proxy/authentication_interceptor.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
@@ -49,6 +50,34 @@ func GrpcAuthInterceptor(authFunc grpc_auth.AuthFunc) grpc.UnaryServerIntercepto
 			return nil, err
 		}
 		return handler(newCtx, req)
+	}
+}
+
+// GrpcAuthStreamInterceptor is the streaming counterpart of GrpcAuthInterceptor.
+// The external gRPC server only mounts UnaryInterceptor, so streaming RPCs such as
+// CreateReplicateStream would otherwise skip the authentication chain entirely. This
+// interceptor guarantees streaming calls are authenticated with the same logic as unary
+// calls before the handler observes the stream.
+func GrpcAuthStreamInterceptor(authFunc grpc_auth.AuthFunc) grpc.StreamServerInterceptor {
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		var newCtx context.Context
+		var err error
+		if overrideSrv, ok := srv.(grpc_auth.ServiceAuthFuncOverride); ok {
+			newCtx, err = overrideSrv.AuthFuncOverride(ss.Context(), info.FullMethod)
+		} else {
+			newCtx, err = authFunc(ss.Context())
+		}
+		if err != nil {
+			hookutil.GetExtension().ReportAction(context.Background(), nil, &milvuspb.BoolResponse{
+				Status: merr.Status(err),
+			}, err, info.FullMethod, hookutil.ActionAuthorize)
+			return err
+		}
+		// Propagate the authenticated context (which carries the resolved user/token) to
+		// the handler by wrapping the ServerStream.
+		wrapped := grpc_middleware.WrapServerStream(ss)
+		wrapped.WrappedContext = newCtx
+		return handler(srv, wrapped)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


internal/proxy/privilege_interceptor.go
```diff
@@ -180,6 +181,119 @@ func PrivilegeInterceptor(ctx context.Context, req interface{}) (context.Context
...
+		return ctx, nil
+	}
+	roleNames, err := GetRole(username)
+	if err != nil {
+		log.Warn("GetRole fail for stream", zap.String("username", username), zap.Error(err))
+		return ctx, err
+	}
+	roleNames = append(roleNames, util.RolePublic)
+
+	dbName := GetCurDBNameFromContextOrDefault(ctx)
```
yhmo, line 267: internal/proxy/privilege_interceptor.go line:267
High ---- StreamPrivilegeInterceptor authorizes the cluster-level PrivilegeUpdateReplicateConfiguration grant against the connection database. Unary authorization explicitly maps cluster-level privileges to util.AnyWord, and grants are encoded under the global resource; with a client sending a non-default db-name metadata value, a role that has the documented global replicate privilege will be denied here. Please use util.AnyWord for cluster-level stream privileges (or apply the same privilege-level mapping as PrivilegeInterceptor) and add a non-default-db regression test.

internal/proxy/authentication_interceptor.go
```diff
@@ -52,6 +53,34 @@ func GrpcAuthInterceptor(authFunc grpc_auth.AuthFunc) grpc.UnaryServerIntercepto
...
+// interceptor guarantees streaming calls are authenticated with the same logic as unary
+// calls before the handler observes the stream.
+func GrpcAuthStreamInterceptor(authFunc grpc_auth.AuthFunc) grpc.StreamServerInterceptor {
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		var newCtx context.Context
+		var err error
+		if overrideSrv, ok := srv.(grpc_auth.ServiceAuthFuncOverride); ok {
+			newCtx, err = overrideSrv.AuthFuncOverride(ss.Context(), info.FullMethod)
+		} else {
+			newCtx, err = authFunc(ss.Context())
```
buqian-zilliz, line 68: **[P1] Propagate authentication metadata to streaming clients**

GrpcAuthStreamInterceptor now invokes AuthenticationInterceptor for every external stream, but the bundled milvusclient.Client injects its configured API key or username/password only through MetadataUnaryInterceptor; it has no stream client interceptor. CDC configures this client with an API key and calls CreateReplicateStream with only cluster-id metadata, so the server receives no authorization metadata and returns Unauthenticated whenever authorization is enabled. Replication then continuously fails and reconnects, and SDK DumpMessages streams likewise cannot start. Add equivalent stream metadata injection or explicitly attach authentication metadata to stream calls.

internal/proxy/privilege_interceptor.go
```diff
@@ -180,6 +181,119 @@ func PrivilegeInterceptor(ctx context.Context, req interface{}) (context.Context
...
+// performs the same role→casbin authorization, but resolves the required privilege from
+// the full method name (via streamMethodPrivileges) instead of the absent request object,
+// and treats the operation as cluster/Global level (objectName == util.AnyWord).
+func StreamPrivilegeInterceptor(ctx context.Context, fullMethod string) (context.Context, error) {
+	if !Params.CommonCfg.AuthorizationEnabled.GetAsBool() {
+		return ctx, nil
+	}
+	log := log.Ctx(ctx)
+
+	ext, ok := streamMethodPrivileges[fullMethod]
```
buqian-zilliz, line 244: **[P2] Allow the registered gRPC health Watch stream**

The external server also registers grpc_health_v1.Health, whose Watch method is server-streaming. It therefore reaches this interceptor but is absent from streamMethodPrivileges, so with authorization enabled every authenticated health watcher is rejected with PermissionDenied, while unary health checks remain allowed. Explicitly bypass the Health/Watch method without granting it the CDC privilege before applying the fail-closed application-method lookup.

PR #52380, opened 2026-08-11
fix: handle non-finite JSON values in JSON stats

zhagnlu:
issue: #52934

### DIFF
--- a/internal/core/src/index/json_stats/bson_builder.cpp
+++ b/internal/core/src/index/json_stats/bson_builder.cpp
@@ -19,8 +19,8 @@
 #include "common/FastMem.h"
 #include <string.h>
 #include <cstdint>
-#include <exception>
 #include <map>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <utility>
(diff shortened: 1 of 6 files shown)


internal/core/src/index/json_stats/JsonKeyStats.h
```diff
@@ -598,9 +598,16 @@ class JsonKeyStats : public ScalarIndex<std::string> {
...
+    bool
+    IsLegacyNonFiniteNumber(const std::string& str) {
+        return str == "NaN" || str == "Infinity" || str == "-Infinity";
+    }
+
     JSONType
     getType(const std::string& str) {
-        if (IsBoolean(str)) {
+        if (IsLegacyNonFiniteNumber(str)) {
+            return JSONType::NONE;
```
buqian-zilliz: **[P1] Do not collapse non-finite scalars into JSON null**

Classifying these tokens as `NONE` makes `NaN` and both infinities indistinguishable from a literal JSON null. Before this change, scalar non-finite values followed the `DOUBLE` path; Arrow/BSON can store them, the JSON-stats comparison helpers preserve ordered comparisons for infinities, and EXISTS treats the value as present. After a stats rebuild, `NONE` is forced into the shared layout, materialized as BSON null, and omitted from the shared-key offsets, so predicates such as `x > 0`, `x < 0`, and `exists(x)` silently change results. Preserve the scalar numeric representation, or reject/mark these inputs invalid explicitly instead of aliasing them to null.

internal/core/src/index/json_stats/bson_builder.cpp
```diff
@@ -106,12 +107,124 @@ AppendJsonElementToBson(simdjson::dom::element elem,
...
+        if (!token.empty() && i + token.size() <= json.size() &&
+            json.substr(i, token.size()) == token &&
+            HasJsonValuePrefix(json, i) &&
+            HasJsonValueSuffix(json, i + token.size())) {
+            if (!normalized.has_value()) {
+                normalized.emplace();
+                normalized->reserve(json.size());
+            }
+            normalized->append(json.data() + copied_until, i - copied_until);
+            normalized->append("null");
```
buqian-zilliz, line 189: **[P1] Preserve non-finite array values instead of materializing null**

Replacing each non-finite token with `null` changes the indexed value rather than only making the legacy input parseable. For example, `[Infinity]` is persisted as BSON `[null]`, so array and null predicates can disagree with evaluation against the source row; this also introduces a null where none existed and affects three-valued logic. BSON doubles can represent NaN and infinities, so preserve them as doubles or mark the affected row or value unindexable and fall back to raw evaluation.

internal/core/src/index/json_stats/bson_builder.cpp
```diff
@@ -106,12 +107,124 @@ AppendJsonElementToBson(simdjson::dom::element elem,
...
 }  // namespace
 
 std::vector<uint8_t>
 BuildBsonArrayBytesFromJsonString(const std::string& json_array) {
     simdjson::dom::parser parser;
-    simdjson::dom::element root = parser.parse(json_array);
+    simdjson::dom::element root;
+    auto error = parser.parse(json_array).get(root);
+    if (error != simdjson::SUCCESS) {
+        if (error != simdjson::TAPE_ERROR && error != simdjson::NUMBER_ERROR) {
```
buqian-zilliz: **[P2] Do not key the fallback to unstable parser error codes**

The compatibility fallback only runs for `TAPE_ERROR` or `NUMBER_ERROR`, but simdjson's error contract explicitly states that different kernels may return different codes for the same invalid document. The same legacy NaN or Infinity array can therefore be normalized on one CPU and rejected on another. Detect the target tokens after a parse failure, while still propagating resource errors such as `CAPACITY` or `MEMALLOC`, and use the second parse to validate the rewritten input.

internal/core/src/index/json_stats/bson_builder.cpp
zhagnlu: NaN not formatted json element, using null to filter it is suitable

internal/core/src/index/json_stats/bson_builder.cpp
```diff
@@ -106,12 +106,134 @@ AppendJsonElementToBson(simdjson::dom::element elem,
...
+            ThrowInfo(ErrorCode::JsonKeyInvalid,
+                      "failed to parse JSON array: {}",
+                      simdjson::error_message(error));
+        }
+
+        LOG_WARN(
```
buqian-zilliz: **[P1] Rate-limit the per-array fallback warning**

`BuildBsonArrayBytesFromJsonString` is called once per array value from `BuildKeyStatsForRow`, and every array containing a supported legacy `NaN` or infinity first fails parsing and reaches this unconditional `LOG_WARN`. On a production stats build with millions of affected rows, this emits one warn-level record per array value, making the compatibility path perform O(rows × array fields) log I/O and potentially dominate indexing or exhaust log storage. Please rate-limit or aggregate this signal, or log it once per segment/build.

internal/core/src/index/json_stats/bson_builder.cpp
buqian-zilliz: This issue still exists: the compatibility fallback is still gated on only `TAPE_ERROR` and `NUMBER_ERROR`, so the same legacy token input can still be accepted or rejected depending on the parser's error classification.

internal/core/src/index/json_stats/bson_builder.cpp
zhagnlu: NaN is, by its very nature, an exceptional condition that does not occur very often; if it occurs on a large scale, it warrants attention. Here, the system merely handles it for compatibility without reporting an error, but does not treat this situation as normal.

internal/core/src/index/json_stats/bson_builder.cpp
```diff
@@ -106,12 +106,136 @@ AppendJsonElementToBson(simdjson::dom::element elem,
...
 BuildBsonArrayBytesFromJsonString(const std::string& json_array) {
     simdjson::dom::parser parser;
-    simdjson::dom::element root = parser.parse(json_array);
+    simdjson::dom::element root;
+    auto error = parser.parse(json_array).get(root);
+    if (error != simdjson::SUCCESS) {
+        // Syntax error codes are not stable across simdjson kernels. Only
+        // resource failures bypass the compatibility fallback.
+        if (error == simdjson::CAPACITY || error == simdjson::MEMALLOC) {
+            ThrowInfo(ErrorCode::JsonKeyInvalid,
```
buqian-zilliz, line 215: **[P2] Preserve parser allocation failures as resource errors**

`MEMALLOC` means simdjson could not allocate parser memory, but this branch converts it to `JsonKeyInvalid`. A syntactically valid array encountered under memory pressure is consequently reported as malformed input, giving callers the wrong signal for diagnostics or retry decisions. Bypass normalization for this error, but propagate or translate it as a resource/system failure; handle `CAPACITY` separately if exceeding the parser limit should remain an input error.

internal/core/src/index/json_stats/bson_builder.cpp
```diff
@@ -106,12 +106,136 @@ AppendJsonElementToBson(simdjson::dom::element elem,
...
+            token = "-Infinity";
+        }
+
+        if (!token.empty() && i + token.size() <= json.size() &&
+            json.substr(i, token.size()) == token &&
+            HasJsonValuePrefix(json, i) &&
+            HasJsonValueSuffix(json, i + token.size())) {
+            if (!normalized.has_value()) {
+                normalized.emplace();
+                normalized->reserve(json.size());
```
buqian-zilliz, line 186: **[P3] Reserve space for expansion when replacing `NaN`**

`reserve(json.size())` underestimates the normalized size because every `NaN` grows from three bytes to the four-byte `null`. A large array dominated by these tokens grows by roughly 25%, forcing the normalized buffer to reallocate and copy while the original input and parser storage are still live. Size the buffer from the replacement count or reserve a safe expansion bound.

internal/core/src/index/json_stats/bson_builder.cpp
zhagnlu: this error will be retry and done successly next time.

internal/core/src/index/json_stats/bson_builder.cpp
zhagnlu: ignored. array not have lots of NAN

Issue #52338, opened 2026-08-09
[Bug]: StructArray FLOAT16 DISKANN returns a negative MAX_SIM_COSINE self-similarity score after index build


yanliang567:
## Environment

- Target Milvus image:
  `harbor.milvus.io/milvusdb/milvus:3.0-20260807-1439dc7d@sha256:ed46e16fcb58bd460722e6fc1c0e6294e86fd4e062431877d0a872dcb510cd64`
- Target source commit from the image tag: `1439dc7de8b198a01c2afa0ae20c0c473e0e1abc`
- Comparison image:
  `harbor.milvus.io/milvusdb/milvus:v3.0.0@sha256:49371c30af46b1013e4d3e0b980e691d81376d69cdbe1b372725baf1d7255862`
- PyMilvus: `3.0.1`
- Deployment: Milvus Operator standalone, RocksMQ, Storage V2
- Vector field: StructArray `ArrayOfVector<FLOAT16_VECTOR>`, dimension `64`
- Index: `DISKANN`
- Metric: `MAX_SIM_COSINE`

## Problem

After a real DISKANN index is built for a StructArray FLOAT16 vector sub-field,
searching with the exact stored vector returns the correct primary key but a
negative score close to `-1.0`:

```json
{
  "collection": "qa_gate_30_to_30latest_forward_struct_array_float16_diskann",
  "distance": -0.999978244304657,
  "expected_pk": 0,
  "field": "embeddings[vector]",
  "index_type": "DISKANN",
  "metric_type": "MAX_SIM_COSINE"
}
```

`MAX_SIM_COSINE` is positively related, so an exact self-match is expected to
be close to `+1.0`. The returned sign is inconsistent with the metric contract
and can also invert document ranking during TokenANN reranking.

## Clean Reproduction From the Upgrade/Rollback Gate

The cleanest reproduction creates a new 3.0 collection on the target image,
inserts 5000 rows, waits for indexes, and validates it before any rollback:

- Argo workflow: `pr25-st30-forward5000-r1-tl6pm`
- Test repository commit: `142d93457e6dfc38b18dea772387bde0c005ce91`
- Failed node: `validate-forward-indexes-after-upgrade`
- Failure occurred on the target image before rollback.
- Every other collection and index in the 3.0 matrix passed.
- The only failure was the StructArray FLOAT16 DISKANN self-score shown above.

This excludes rollback, phase DML, schema evolution, and delete pressure as
required triggers.

[...]

PR #52384, opened 2026-08-11
fix: update Knowhere to fix cached DiskANN distance calculation

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/52338

knowhere fix: https://github.com/zilliztech/knowhere/pull/1761

### DIFF
--- a/internal/core/thirdparty/knowhere/CMakeLists.txt
+++ b/internal/core/thirdparty/knowhere/CMakeLists.txt
@@ -14,9 +14,9 @@
 # Update KNOWHERE_VERSION for the first occurrence
 milvus_add_pkg_config("knowhere")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-# Pin zilliztech/knowhere#1745 until the next release; v3.0.6 still pins
-# Cardinal v3.0.3 and an older milvus-common revision.
-set( KNOWHERE_VERSION 16a71a3d )
+# Includes zilliztech/knowhere#1761, which fixes COSINE/IP CalcDistByIDs
+# when all requested DiskANN vectors are cached.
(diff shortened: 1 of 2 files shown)


internal/core/thirdparty/knowhere/CMakeLists.txt
```diff
@@ -14,9 +14,9 @@
 # Update KNOWHERE_VERSION for the first occurrence
 milvus_add_pkg_config("knowhere")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-# Pin zilliztech/knowhere#1745 until the next release; v3.0.6 still pins
-# Cardinal v3.0.3 and an older milvus-common revision.
-set( KNOWHERE_VERSION 16a71a3d )
+# Includes zilliztech/knowhere#1761, which fixes COSINE/IP CalcDistByIDs
+# when all requested DiskANN vectors are cached.
+set( KNOWHERE_VERSION 6d0f7fe5 )
```
buqian-zilliz, line 19: **[P2] Avoid the CPU-scaled sparse-build cursor matrix**

Pinning `6d0f7fe5` also imports Knowhere #1746. Its `prepare_posting_build_plan` allocates a dense `uint32_t[nr_inner_dims]` cursor vector for every build-pool worker, whereas the previous pin kept only O(nr_inner_dims) cursor state. Milvus configures this pool to the machine CPU count in cluster mode, so a 64-core builder with 1 million distinct sparse dimensions spends about 244 MiB on this matrix alone, before raw postings and encoded buffers. Existing high-cardinality sparse/BM25 builds can therefore exceed their memory limit even though this PR only targets DiskANN. Please backport #1761 onto `16a71a3d`, or use a Knowhere revision that bounds this allocation before advancing the pin.

zhengbuqian (approved):

Issue #52238, opened 2026-08-06
[Bug]: DumpMessages WAL salvage stream lacks RBAC authorization


shaoting-huang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master @ 82ecdb68f5
- Deployment mode(standalone or cluster): standalone or cluster
- MQ type(rocksmq, pulsar or kafka): N/A
- SDK version(e.g. pymilvus v2.0.0rc2): N/A
- OS(Ubuntu or CentOS): N/A
- CPU/Memory: N/A
- GPU: N/A
- Others:
```

### Current Behavior

`DumpMessages` is a server-streaming MilvusService RPC that opens WAL reads by pchannel and streams messages for CDC/data salvage. The proxy RBAC interceptor is unary-only, and `DumpMessagesRequest` has no `privilege_ext_obj` annotation.

### Expected Behavior

`DumpMessages` should require an explicit authorization path appropriate for CDC/data salvage, such as admin/root authorization, before streaming WAL contents.

### Steps To Reproduce

1. Inspect the proxy gRPC server setup and observe that `PrivilegeInterceptor` is installed in the unary chain, not the stream chain.
2. Inspect `Proxy.DumpMessages`, which reads from `streaming.WAL().Read` and sends `DumpMessagesResponse` messages.
3. Observe that the request is not covered by the unary `privilege_ext_obj` RBAC path.

### Milvus Log

N/A -- descriptor/authorization coverage gap.

### Anything else?

The CDC data salvage design notes that `DumpMessages` exposes raw WAL data and should require admin/root authorization.

bailu-ZZ:
I would like to work on this issue.

  I inspected the current Proxy gRPC setup and confirmed the authorization gap. The external Proxy server currently applies authentication and privilege checks to unary RPCs only. DumpMessages is a server-streaming RPC, so it does not go through the existing privilege_ext_obj RBAC path and may open a WAL scanner before authorization is performed.

  This is especially sensitive because DumpMessages returns raw WAL messages. A pchannel may contain data and metadata for multiple collections, so collection-level authorization is not sufficient.

  Proposed implementation:

  1. Add an explicit privilege contract in milvus-proto.

[...]

PR #52386, opened 2026-08-11
fix: add user authentication and RBAC authorization for stream gRPC

Amos-yongxin:
issue: #52387
## Background
Streaming RPCs (e.g. CreateReplicateStream, DumpMessages) were not going through the unary interceptor chain on the external client-facing gRPC port. Only Authentication (credential verification) was added to the stream path, RBAC authorization was missing, so authenticated users could bypass role-based access checks for stream-type resource operations.

## Changes
1. Authentication: add proxy.GrpcAuthStreamInterceptor that validates username+password (bcrypt/SHA-256) on external streaming calls:
   - reject missing Authorization metadata
   - reject bad credentials
   - accept valid credentials and propagate username/sha256_password via contextutil.SetAuthInfo. When CommonCfg.AuthorizationEnabled is false the behavior is skipped, consistent with the unary chain. Register the stream interceptor in startExternalGrpc for the external gRPC server with explicit comment explaining why stream must be wired separately (streaming RPCs bypass the unary chain).
2. RBAC authorization: add the stream counterpart of proxy.UnaryServerInterceptor(proxy.PrivilegeInterceptor), i.e. proxy.StreamServerInterceptor(proxy.PrivilegeInterceptor). For stream RPCs described in privilege.Ext (CreateReplicateStream and DumpMessages), the interceptor runs Casbin Enforce against Global-default
   + PrivilegeUpdateReplicateConfiguration so role bindings correctly allow/deny stream calls. UTs verify deny-blocks-handler path and the permit path that propagates the authorized context.

### DIFF
--- a/internal/proxy/privilege_interceptor.go
+++ b/internal/proxy/privilege_interceptor.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"sync"
 
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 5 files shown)


Amos-yongxin:
@czs007 @xiaocai2333 PTAL

internal/proxy/privilege_interceptor.go
```diff
@@ -209,6 +210,122 @@ func PrivilegeInterceptor(ctx context.Context, req interface{}) (context.Context
...
+	}
+	roleNames, err := GetRole(username)
+	if err != nil {
+		mlog.Warn(ctx, "GetRole fail for stream", mlog.String("username", username), mlog.Err(err))
+		return ctx, err
+	}
+	roleNames = append(roleNames, util.RolePublic)
+	ctx = SetRBACRolesToContext(ctx, roleNames)
+
+	dbName := GetCurDBNameFromContextOrDefault(ctx)
```
yhmo, line 299: internal/proxy/privilege_interceptor.go line:299
Medium ---- Cluster-level grants for PrivilegeUpdateReplicateConfiguration are stored with the database scope `*` (rootcoord validation/expansion requires this, and the unary interceptor maps cluster privileges to `util.AnyWord`), but this stream path builds `Global-<current db>` from `GetCurDBNameFromContextOrDefault`. A valid non-root role therefore gets PermissionDenied on CreateReplicateStream/DumpMessages whenever the client context is `default` (or any concrete DB). Please use the same global `util.AnyWord` scope as the unary path and add a test with a `dbName=*` policy.

shaoting-huang:
Hi @Amos-yongxin — thank you for this PR.

This approach overlaps with #52608, which has been extended to cover the full scope of #52387 (CreateReplicateStream + DumpMessages, authentication + RBAC + fail-closed) with the review feedback from yhmo incorporated:
- stream RBAC uses the same `util.AnyWord` global scope as the unary path (fixing the db-scope bug),
- `grpc.health.v1.Health/Watch` is exempted from RBAC so health checks keep working,
- `DumpMessages` uses the admin/root authorization from #52238, `CreateReplicateStream` uses casbin `PrivilegeUpdateReplicateConfiguration`.

Marking this PR as **superseded by #52608**. Closing to avoid duplicate fixes for the same issue.

PR #52388, opened 2026-08-11
enhance: replace field binlog array reads with stats in metrics and quota paths

tedxu:
See #50406

V3 segments may have empty Binlogs arrays after DataCoord
restart (KVs are no longer persisted per #50410). Replace
the remaining FieldBinlog iterations with Statistics
accessors so insert/delta file counts and format labels
report correct values for V3.

- Add formats field to Statistics proto: unique set of
  column-group storage formats collected from insert
  binlogs during writes and compaction
- segmentMetricFormatLabel: read Stats.Formats instead of
  iterating Binlogs
- reloadFromKV: use Stats.InsertBinlogCount and
  Stats.DeltaBinlogCount for flushed-segment metrics
- getBinlogFileCount: same substitution for quota metrics
- GetPersistentSegmentInfo: document that V3 returns empty
  insert logs by design


jiaqizho:
@tedxu review: one gap in the “compaction also collects formats” claim — the in-place schema-bump materialization path drops the new column-group format from `Stats.Formats`.

Trace:

1. Producer (`internal/datanode/compactor/stats.go` `buildMaterializationStatsDelta`) only populates `StatsBinlogSize`, `InsertBinlogSize`, `InsertBinlogCount`, and `NullCounts`. It never sets `Formats`, even though the new column groups carry a format.
2. Receiver (`internal/datacoord/stats_delta.go` `addStatsDelta`) unions only `NullCounts`; `Formats` is whatever `proto.Clone(base)` carried over, and delta `Formats` is ignored.
3. The in-place path (`bump_schema_version_compactor.go` `newV3WriterResult` + `buildNewInsertLogsV3`) does write `columnGroup.Format` into `InsertLogs[].Format`, and `completeBumpSchemaVersionCompactionMutation` merges those groups into `cloned.Binlogs` via `mergeSegmentColumnGroups`. So the Binlogs array sees the new format, but `cloned.Stats.Formats` stays stale.

Effect: for a V3 segment with e.g. `parquet` whose schema-bump materialization writes a new group as `vortex`, `segmentMetricFormatLabel` can continue reporting `parquet` instead of `mixed` (or `vortex`). That’s exactly the metric this PR is trying to keep correct.

Suggested fix (either):
- In `addStatsDelta`, union `delta.GetFormats()` into the result’s `Formats` (dedup + sort), or
- Set `Formats` in `buildMaterializationStatsDelta` from `columnGroups` and merge it in `addStatsDelta`.

Please also add a regression test: V3 segment already `parquet`, in-place schema-bump writes a `vortex` group, assert `segmentMetricFormatLabel` becomes `mixed`.

PR #52390, opened 2026-08-11
fix: [cp3.0] gate V3 function-output index build on segment schema version (#51768)

tedxu:
Cherry-pick from master
pr: #51768
issue: #51727

Gate V3 function-output index build on segment schema
version rather than field binlog presence. Recovered V3
segments have empty Binlogs after #50410, so the previous
gate rejected materialized function-output fields
permanently.

Also cherry-picks #52272 to fix a compilation break:
#52391 (already merged into 3.0) added a call to
getSegmentBinlogFields which this PR deletes. #52272
inlines the helper and updates stale comments.

### DIFF
--- a/internal/datacoord/backfill_result.go
+++ b/internal/datacoord/backfill_result.go
@@ -190,10 +190,9 @@ func buildV2Groups(bucket string, entry *BackfillSegment) (map[int64]*datapb.Fie
 		}
 		out[fid] = &datapb.FieldBinlog{
 			FieldID: fid,
-			// ChildFields carries the real field IDs that index creation
-			// (getSegmentBinlogFields) and backfill-compaction detection
-			// (getMissingFunctions) rely on. Without it the new group is
-			// invisible to both paths; it also lets the operator's strip
+			// ChildFields carries the real field IDs that backfill-compaction
+			// detection (getMissingFunctions) relies on. Without it the new
+			// group is invisible to that path; it also lets the operator's strip
 			// logic reference-count the field out of the old groups.
 			// The backfill invariant guarantees len(g.FieldIDs) == 1.
 			ChildFields: []int64{fid},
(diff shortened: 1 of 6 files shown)


congqixia (changes requested):
Blocking: this cherry-pick will break the datacoord build on 3.0 — needs the equivalent of master's #52272 fix
The cherry-pick itself is faithful (the 3 files are byte-identical to master's #51768 merge), and the schema-version gate is sound. But merging this PR into the current 3.0 tip as-is breaks compilation of internal/datacoord:
internal/datacoord/meta_test.go:2223 still calls the helper this PR deletes:
fields := getSegmentBinlogFields(m.segments.GetSegment(1))
I verified this by simulating the merge (git merge-tree origin/3.0 <this-branch>): in the merged tree, getSegmentBinlogFields no longer exists in index_inspector.go, while the call above remains → go vet / go test ./internal/datacoord/... fail with undefined: getSegmentBinlogFields.
Why CI is green: this branch's parent predates #52391 (the 3.0 cherry-pick of master #52006, already merged), which is what added the meta_test.go:2223 reference. The two PRs touch disjoint files, so the rebase onto the current tip reports no conflict but lands on a tree no CI job has compiled — the same incident master hit between #51768 and #52272 (see #52272's "How it happened" table; its last line is exactly this: "tide pushes a rebased tree that no CI job has seen").
Please fold in the equivalent of #52272 before merge:
- Inline the ChildFields expansion in the meta_test.go subtest (asserting field 102 and 100 are index-eligible after materialization) instead of calling the deleted helper, and
- Refresh the now-stale comments in meta.go:3460-3461 and bump_schema_version_compactor.go:862-863 that still describe getSegmentBinlogFields as the index-eligibility input.
Alternatively, land #52272's 3.0 counterpart first. Note this same breakage also applies to any in-flight 3.0 PRs that build on getSegmentBinlogFields; worth grepping the open cp-3.0 PR set.

congqixia (approved):
/lgtm

PR #52392, opened 2026-08-11
enhance: [cp3.0] read StorageV3 segments via manifest in clustering analyze (#51885)

tedxu:
Cherry-pick from master
pr: #51885
issue: #50406

Clustering analyze callers now pass manifest path and
storage version through to the segment reader, so V3
segments are read via manifest instead of relying on
per-field binlog paths.

No conflicts during cherry-pick.


congqixia (approved):
/lgtm

PR #52393, opened 2026-08-11
fix: [cp3.0] remove FieldBinlog emptiness guards that reject V3 segments (#52223)

tedxu:
Cherry-pick from master
pr: #52223
issue: #50406

Remove FieldBinlog emptiness guards in mix compaction and
GetRecoveryInfo that incorrectly reject V3 segments whose
Binlogs arrays are empty after DataCoord restart.

No conflicts during cherry-pick.


congqixia (approved):
/lgtm

Issue #52395, opened 2026-08-11
[Bug]: External manifest reopen fails when reloading existing fields


weiliu1031:
## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Environment

- Affected code: current `master` and `3.0` branches
- Deployment mode: cluster
- Collection type: external collection using manifest-based sealed segments

## Current Behavior

When an external segment's manifest path changes, the reopen diff correctly
sets both `manifest_updated` and `load_external_manifest`. The external
column-group loader then reloads every stored field as a new load instead of a
replacement.

Because the fields are already present in the published segment state, the
staged load fails with an error such as:

```text
field <field-id> data already loaded
```

The segment reopen fails and the upper layers can report the segment as not
loaded or missing. Query and search requests may fail until the segment is
recovered.

## Expected Behavior

A manifest-driven reopen should replace fields that were loaded from the
previous manifest. Initial external-manifest loads should continue rejecting
unexpected duplicate field data.

## Steps To Reproduce

```text
1. Load an external collection whose sealed segment uses manifest v1.
2. Refresh the external source so the segment points to manifest v2.
3. Trigger segment reopen through the segment or index checker.
4. Observe that the reopen reloads existing fields with replacement disabled.
5. Observe "field data already loaded" and a missing/not-loaded segment.
```

## Milvus Log

```text
field <field-id> data already loaded
```

## Anything else?

`SegmentLoadInfo::ComputeDiff` already distinguishes an initial external
manifest load from a manifest update. The replacement state is lost when
`ChunkedSegmentSealedImpl::PrepareLoadDiffForReopen` enters the external
manifest loading helper, which always forwards `is_replace=false` to the
column-group load.

PR #52396, opened 2026-08-11
fix: Replace external columns on manifest reopen

weiliu1031:
issue: #52395

### What changed

Propagate the external manifest update state through the column-group load
path. A manifest update now loads existing fields as replacements, while an
initial manifest load continues to reject duplicate field data.

### Why

External segment reopen correctly records that the manifest changed, but the
external column-group helper previously discarded that state and always used
`is_replace=false`. Reloading fields already present in the published segment
state then failed with `field data already loaded`, which could leave the
segment missing or not loaded for query and search.

### Validation

- `make milvus` on the equivalent 3.0 candidate
- `make build-cpp-with-unittest` on the equivalent 3.0 candidate
- `scripts/run_cpp_unittest.sh` with:
  - `SegmentLoadInfoTest.ComputeDiffExternalManifestUpdateReloadsExternalManifest`
  - `SegmentLoadInfoTest.ComputeDiffExternalManifestWithNewIndexDoesNotParseColumnNames`
- `clang-format-12 --dry-run --Werror` on the changed files
- `git diff --check`

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -2120,6 +2120,7 @@ ChunkedSegmentSealedImpl::LoadColumnGroups(
     const SegmentLoadInfo& segment_load_info,
     const SchemaPtr& schema_snapshot,
     milvus::OpContext* op_ctx,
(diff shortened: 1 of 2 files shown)


congqixia (approved):
/lgtm

PR #52398, opened 2026-08-11
feat: [ElementNull 2] support element-nullable array value models

SpadeA-Tang:
issue: https://github.com/milvus-io/milvus/issues/51162

Adds schema metadata and core Array, ArrayView, and vector-array representations for per-element validity.


internal/core/src/common/Array.h
```diff
@@ -440,6 +393,10 @@ class Array {
         if (!arr2.same_type()) {
             return false;
         }
+        if (has_invalid_element_) {
```
buqian-zilliz, line 396: **[P1] Do not collapse nullable array equality into ordinary false**

`is_same_array()` returns `false` for every array containing a null element. The expression callers use this Boolean directly for equality and negate it for `NotEqual`, so `[1, NULL] != [1, 2]` becomes a valid `true` result. A comparison involving a null element must remain unknown/invalid under three-valued logic; return validity separately, or otherwise propagate it, so neither equality nor inequality turns the null case into a Boolean match.

internal/core/src/exec/expression/UnaryExpr.h
```diff
@@ -549,7 +549,7 @@ struct UnaryElementFuncForArray {
                         continue;
                     }
                     auto array_data =
-                        src[offset].template get_data<GetType>(index);
+                        src[offset].template get_data_unchecked<GetType>(index);
```
buqian-zilliz, line 553: **[P1] Check element validity before every unchecked expression read**

`get_data_unchecked()` explicitly leaves validity handling to the caller, but this path checks only row validity and bounds. A null slot's dense placeholder is therefore compared as a real value while `valid_res` remains true. The same unchecked-read pattern was introduced in BinaryRange, BinaryArith, Term, JsonContains/ContainsAll, and the flattened element-level path in Expr.h. Check `is_element_valid()` before reading and apply the operator-specific null behavior; indexed predicates must invalidate the result, and scans must not treat null placeholders as candidates.

internal/core/src/index/BitmapIndex.cpp
```diff
@@ -202,7 +202,7 @@ BitmapIndex<T>::BuildArrayField(const std::vector<FieldDataPtr>& field_datas) {
                 auto array =
                     reinterpret_cast<const milvus::Array*>(data->RawValue(i));
                 for (size_t j = 0; j < array->length(); ++j) {
-                    auto val = array->get_data<T>(j);
+                    auto val = array->get_data_unchecked<T>(j);
```
buqian-zilliz, line 205: **[P1] Exclude null element payloads from scalar indexes**

`BuildArrayField()` inserts every logical slot into postings even when the new per-element bitmap marks that slot null. Because null elements still carry a dense placeholder value, a query for that placeholder can match a row whose corresponding element is NULL. The nested builder also indexes these slots and then marks the entire `valid_bitset_` true; the Tantivy and sort builders follow the same unchecked pattern. Keep logical element IDs aligned, but do not add null slots to value postings and record them as invalid in nested indexes.

internal/core/src/common/Array.h
SpadeA-Tang: This PR does not provide complete element-null capability and the element-null is not enabled. This will be handled in the following PR.

internal/core/src/exec/expression/UnaryExpr.h
SpadeA-Tang: ditto

internal/core/src/index/BitmapIndex.cpp
SpadeA-Tang: ditto

zhengbuqian (approved):

PR #52415, opened 2026-08-11
fix: [cp3.0]reject invalid import commit timestamp

bigsheeper:
Cherry-pick from master

pr: #51589
issue: #51588

## Summary

Cherry-picked from master PR #51589 (merged)

## Deviation from the master PR

One test is **not** picked: `TestHandleCommitVchannelRPC_V3SegmentIsNotFencedYet`.

It builds `datapb.SegmentInfo{Stats: &datapb.Statistics{TimestampTo: 500}}`, but 3.0's `pkg/proto/data_coord.proto` has neither `SegmentInfo.stats` nor `message Statistics` (both are master-only), so the `internal/datacoord` test package failed typecheck and Code-Check errored out:

```
internal/datacoord/services_test.go:5972:3: unknown field Stats in struct literal of type datapb.SegmentInfo
internal/datacoord/services_test.go:5972:26: undefined: datapb.Statistics (typecheck)
```

That test does not cover this PR's fix. It pins a pre-existing gap — a V3 (manifest-backed) segment persists no binlog arrays, so the fence compares against 0 — and its assertions are meant to be flipped by a master-only follow-up that reads `Stats.TimestampTo`. On 3.0 the gap cannot be expressed the same way, and rewriting it without `Stats` would degrade it to a trivial assertion.

Everything else is picked unchanged, including the fix in `meta.go` and its regression test `TestHandleCommitVchannelRPC_RejectsCommitTimestampBelowBinlogTimestamp`.

## Verification

- [x] File count matches original PR
- [x] Code changes verified against master PR diff (line-level comparison)
- [x] No conflict markers
- [x] `go vet -tags dynamic,test ./internal/datacoord/` clean (covers the typecheck failure Code-Check reported)
- [x] Rebased onto latest 3.0

### DIFF
--- a/internal/datacoord/commit_timestamp_test.go
+++ b/internal/datacoord/commit_timestamp_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/msgpb"
 	"github.com/milvus-io/milvus/pkg/v3/proto/datapb"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/tsoutil"
 )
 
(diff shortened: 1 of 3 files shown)


czs007 (approved):
/lgtm

Issue #52413, opened 2026-08-11
[Bug]: A single-row Upsert exceeds BookKeeper's frame limit of approximately 5 MiB, causing the WAL to retry continuously and stalling the entire DML channel.


zhagnlu:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master 
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

[https://grafana-vdc-dev.zilliz.cc/explore?orgId=1&panes={"IaM":{"datasource":"Zgl22j0Vk","queries":[{"refId":"A","expr":"{cluster%3D\"uat-milvus-us-west-2\",namespace%3D\"milvus-in01-3424a59a448240b\",pod%3D~\"in01-3424a59a448240b-milvus-.*\"}|%3D\"channel tsafe stalled\"","datasource":{"type":"loki","uid":"Zgl22j0Vk"},"editorMode":"code","queryType":"range"}],"range":{"from":"now-30m","to":"now"}}}&schemaVersion=1](https://grafana-vdc-dev.zilliz.cc/explore?orgId=1&panes=%7B%22IaM%22:%7B%22datasource%22:%22Zgl22j0Vk%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcluster%3D%5C%22uat-milvus-us-west-2%5C%22,namespace%3D%5C%22milvus-in01-3424a59a448240b%5C%22,pod%3D~%5C%22in01-3424a59a448240b-milvus-.%2A%5C%22%7D%7C%3D%5C%22channel%20tsafe%20stalled%5C%22%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22Zgl22j0Vk%22%7D,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-30m%22,%22to%22:%22now%22%7D%7D%7D&schemaVersion=1)

img_v3_02148_3aa422c6-0879-4e29-9c70-a24705d415cg-20260804-033511.jpg
[...]

PR #52420, opened 2026-08-11
fix: reject oversized rows during DML message packing

zhagnlu:
issue: #52413

### DIFF
--- a/internal/proxy/msg_pack.go
+++ b/internal/proxy/msg_pack.go
@@ -18,14 +18,39 @@ package proxy
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/msgpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
+	streamingutil "github.com/milvus-io/milvus/internal/util/streamingutil/util"
 	"github.com/milvus-io/milvus/pkg/v3/mq/msgstream"
+	"github.com/milvus-io/milvus/pkg/v3/streaming/util/message"
 	"github.com/milvus-io/milvus/pkg/v3/util/commonpbutil"
+	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
+
+func getActiveWALName() message.WALName {
+	return streamingutil.MustSelectWALName()
+}
+
+func getMaxSingleRowSize(walName message.WALName) (int, bool) {
+	switch walName {
+	case message.WALNamePulsar:
+		limit := Params.PulsarCfg.MaxMessageSize.GetAsInt()
+		return limit, limit > 0
+	case message.WALNameKafka:
+		limit := Params.KafkaCfg.ProducerMessageMaxBytes.GetAsInt()
+		return limit, limit > 0
+	case message.WALNameRocksmq, message.WALNameWoodpecker:
+		// RocksMQ page size and Woodpecker batch size are not hard limits
+		// on an individual WAL entry.
+		return 0, false
+	default:
+		return 0, false
+	}
+}
 
 func genInsertMsgsByPartition(ctx context.Context,
 	segmentID UniqueID,
@@ -34,8 +59,12 @@ func genInsertMsgsByPartition(ctx context.Context,
 	rowOffsets []int,
 	channelName string,
 	insertMsg *msgstream.InsertMsg,
+	walName message.WALName,
 ) ([]msgstream.TsMsg, error) {
-	threshold := Params.PulsarCfg.MaxMessageSize.GetAsInt()
+	// Keep the existing cross-WAL packing threshold separate from the
+	// backend-specific hard limit for a row that cannot be split further.
+	splitThreshold := Params.PulsarCfg.MaxMessageSize.GetAsInt()
+	singleRowLimit, hasSingleRowLimit := getMaxSingleRowSize(walName)
 
 	// create empty insert message
 	createInsertMsg := func(segmentID UniqueID, channelName string) *msgstream.InsertMsg {
@@ -77,11 +106,16 @@ func genInsertMsgsByPartition(ctx context.Context,
 		if err != nil {
 			return nil, err
 		}
+		if hasSingleRowLimit && curRowMessageSize >= singleRowLimit {
+			return nil, merr.WrapErrParameterTooLarge(fmt.Sprintf(
+				"single row at offset %d is too large to fit in one WAL message: estimated size=%d bytes, limit=%d bytes",
+				offset, curRowMessageSize, singleRowLimit,
+			))
+		}
 
 		// If the insert message size exceeds the threshold, flush the current
-		// message first. A single row can be larger than the threshold, so do
-		// not emit an empty message before adding that row.
-		if msg.NumRows > 0 && requestSize+curRowMessageSize >= threshold {
+		// message first.
+		if msg.NumRows > 0 && requestSize+curRowMessageSize >= splitThreshold {
 			repackedMsgs = append(repackedMsgs, msg)
 			msg = createInsertMsg(segmentID, channelName)
 			requestSize = 0
--- a/pkg/util/paramtable/service_param.go
+++ b/pkg/util/paramtable/service_param.go
@@ -1317,20 +1317,21 @@ If this option is zero or negative, it will be ignored and the default value (10
 
 // --- kafka ---
 type KafkaConfig struct {
-	Address              ParamItem  `refreshable:"false"`
-	SaslUsername         ParamItem  `refreshable:"false"`
-	SaslPassword         ParamItem  `refreshable:"false"`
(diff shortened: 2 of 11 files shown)


internal/proxy/msg_pack.go
```diff
@@ -77,10 +79,15 @@ func genInsertMsgsByPartition(ctx context.Context,
 		if err != nil {
 			return nil, err
 		}
+		if curRowMessageSize >= threshold {
```
buqian-zilliz: **[P2] Use the active WAL limit when rejecting oversized rows**

This check turns the Pulsar-only pulsar.maxMessageSize setting into a hard input limit for every streaming WAL. genInsertMsgsByPartition is shared by insert and upsert regardless of mq.type, while Kafka configures message.max.bytes to 10 MiB and standalone may use RocksMQ; a 3 MiB single row that was previously emitted can now be rejected at the default 2 MiB Pulsar threshold even when the active backend accepts it. Derive the rejection limit from the active WAL backend, or define an explicit backend-neutral row-size limit instead of applying the Pulsar setting globally.

buqian-zilliz: **[P2] Check the serialized WAL message size**

EstimateEntitySize only sums logical field contents. The actual WAL message is built later by protobuf-marshalling the full InsertRequest and adding field metadata, row IDs, timestamps, hash values, header properties, and optional encryption, so a row just below threshold can pass this guard even though the serialized message exceeds the backend limit. Validate the finalized payload and properties, or reserve a proven worst-case envelope, before accepting the row; otherwise the oversized append failure this change targets remains possible.

zhagnlu: The configured threshold is intentionally expected to be lower than the WAL backend’s hard message-size limit; it is a conservative payload/splitting budget rather than the exact serialized frame limit.

zhagnlu: fixed

internal/proxy/msg_pack.go
```diff
@@ -18,24 +18,56 @@ package proxy
...
+	return streamingutil.MustSelectWALName()
+}
+
+func getMaxSingleRowSize(walName message.WALName) (int, bool) {
+	switch walName {
+	case message.WALNamePulsar:
+		limit := Params.PulsarCfg.MaxMessageSize.GetAsInt()
+		return limit, limit > 0
+	case message.WALNameKafka:
+		limit := Params.KafkaCfg.ProducerMessageMaxBytes.GetAsInt()
```
buqian-zilliz, line 44: **[P2] Use the effective Kafka broker limit for rejection**

`getMaxSingleRowSize` treats the producer-side `message.max.bytes` value as Kafka's hard acceptance limit. The repository's dev Kafka broker is capped at 5 MiB while this new parameter defaults to 10 MiB, so an incompressible 6 MiB row passes this guard and the producer setting, then is rejected by the broker; a topic-level override can be lower still. Validate against an explicit effective broker/topic budget, or enforce that this setting never exceeds it, otherwise the oversized-row failure remains for Kafka.

internal/proxy/msg_pack.go
```diff
@@ -18,24 +18,56 @@ package proxy
...
 	partitionName string,
 	rowOffsets []int,
 	channelName string,
 	insertMsg *msgstream.InsertMsg,
+	walName message.WALName,
 ) ([]msgstream.TsMsg, error) {
-	threshold := Params.PulsarCfg.MaxMessageSize.GetAsInt()
+	// Keep the existing cross-WAL packing threshold separate from the
+	// backend-specific hard limit for a row that cannot be split further.
+	splitThreshold := Params.PulsarCfg.MaxMessageSize.GetAsInt()
```
buqian-zilliz, line 66: **[P2] Cap aggregate packing at the Kafka message limit**

`splitThreshold` remains tied to `pulsar.maxMessageSize`, while the Kafka limit is applied only to each individual row. If `kafka.producer.message.max.bytes` is 1 MiB and Pulsar remains at its 2 MiB default, two roughly 600 KiB rows both pass the new guard and are packed into a roughly 1.2 MiB record, which both Kafka producer paths reject. Bound aggregate packing by the active backend limit as well; otherwise ordinary multi-row inserts and upserts can still fail during append.

internal/proxy/msg_pack.go
buqian-zilliz: This issue still exists. `getActiveWALName` returns the proxy process's startup-registered default, but online AlterWAL switches the actual WAL per pchannel and refreshes that default only inside streaming-node processes. After a live WAL switch, the proxy can therefore keep applying the old backend's row limit until it restarts.

internal/proxy/msg_pack.go
zhagnlu: message.max.bytes usually greater than split threshold

internal/proxy/msg_pack.go
zhagnlu: This limit is configurable through kafka.producer.message.max.bytes and is intended to be set no higher than the effective broker/topic max.message.bytes. Both the Proxy guard and Kafka producers use this value, so for the dev broker it can be configured to 5242880 to reject oversized rows before append. For deployments with topic-level overrides, operators should configure it to the lowest effective limit.

internal/proxy/msg_pack.go
zhagnlu: fixed, remove default using msg.type

pkg/util/paramtable/service_param.go
```diff
@@ -1317,20 +1317,21 @@ If this option is zero or negative, it will be ignored and the default value (10
...
+	SaslUsername            ParamItem  `refreshable:"false"`
+	SaslPassword            ParamItem  `refreshable:"false"`
+	SaslMechanisms          ParamItem  `refreshable:"false"`
+	SecurityProtocol        ParamItem  `refreshable:"false"`
+	KafkaUseSSL             ParamItem  `refreshable:"false"`
+	KafkaTLSCert            ParamItem  `refreshable:"false"`
+	KafkaTLSKey             ParamItem  `refreshable:"false"`
+	KafkaTLSCACert          ParamItem  `refreshable:"false"`
+	KafkaTLSKeyPassword     ParamItem  `refreshable:"false"`
+	ProducerMessageMaxBytes ParamItem  `refreshable:"false"`
```
buqian-zilliz, line 1330: **[P2] Prevent live updates from desynchronizing Kafka limits**

The `refreshable:"false"` tag does not register this key as immutable; only `ParamItem.Immutable` is consulted by the management update path. `HandleAlterConfig` can therefore persist a live change and evict the parameter cache, so the Proxy immediately enforces the new value, while the msgstream producer is process-global and the streaming Kafka opener remains cached with the old value until shutdown. Raising the setting then admits rows that the active producer still rejects, and lowering it rejects rows the producer could send. Reject live updates for this key or reload both producer paths and switch the guard atomically.

internal/proxy/msg_pack.go
```diff
@@ -18,24 +18,53 @@ package proxy
...
+	return streamingutil.MustSelectWALName()
+}
+
+func getMaxSingleRowSize(walName message.WALName) (int, bool) {
+	switch walName {
+	case message.WALNamePulsar:
+		limit := Params.PulsarCfg.MaxMessageSize.GetAsInt()
+		return limit, limit > 0
+	case message.WALNameKafka:
+		limit := Params.KafkaCfg.ProducerMessageMaxBytes.GetAsInt()
```
buqian-zilliz, line 44: **[P2] Apply the Kafka limit to delete-message packing**

The new Kafka limit is only used by insert-row repacking. `repackDeleteMsgByHash` still batches tombstones against `pulsar.maxMessageSize`, and both standalone deletes and the delete half of upserts use that helper. If `kafka.producer.message.max.bytes` is configured below the Pulsar threshold, this path can build a delete record above the producer limit and fail during append. Pass the active backend limit into delete repacking and cap its aggregate batch size too.

internal/proxy/msg_pack.go
buqian-zilliz: This issue still exists. Re-reading `mq.type` fixes the steady-state startup snapshot, but it is still one process-wide value rather than the WAL serving each destination pchannel. During AlterWAL, a pchannel is fenced as soon as its control record is appended, while `mq.type` changes only after all vchannels acknowledge the record and remote proxies refresh it; messages packed in that interval can be retained across the fence and appended to the new checkpoint-selected WAL using the old backend limit. Derive the limit from the destination pchannel or producer assignment, or use a backend-neutral limit that is valid throughout the transition.

pkg/util/paramtable/service_param.go
zhagnlu: fixed

internal/proxy/msg_pack.go
zhagnlu: In practice, kafka.producer.message.max.bytes is not expected to be configured below pulsar.maxMessageSize. Its default is 10 MiB, while the Pulsar packing threshold defaults to 2 MiB, so delete messages are already split conservatively before reaching Kafka’s producer limit. Delete tombstones are also significantly smaller than insert rows. We therefore keep the existing Pulsar-based delete packing threshold for now.

zhengbuqian (approved):

Issue #52425, opened 2026-08-11
[Bug]: Search path allows non-8-byte-aligned SparseFloatVector rows, causing Heap OOB


Amos-yongxin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 2.5/2.6/3.0
- Deployment mode(standalone or cluster): both standalone and cluster
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior
On **Search / HybridSearch** requests targeting a `SparseFloatVector` field, the Proxy forwards rows from `PlaceholderGroup.Values` directly to C++ segcore **without validating 8-byte alignment**. A malicious 9-byte (or 17-byte / 25-byte / ...) row passes through unchanged. Since segcore reads sparse cells as `uint32 idx + float32 val` (8B each) with built-in `// does not check for out-of-range access`, this yields a **Heap OOB Read** on unaligned lengths.

Insert / Import data path is NOT affected because it already calls `typeutil.ValidateSparseFloatRows` on `FieldData`.

### Expected Behavior
Same `typeutil.ValidateSparseFloatRows` check MUST also run on the Search/HybridSearch path, before any cgo call. Any non-8-multiple row length should return `ErrParameterInvalid` immediately at Proxy trust boundary.

### Steps To Reproduce
1. Create a collection with a `SPARSE_FLOAT_VECTOR` field (no data needs to be inserted, validation happens before segcore).
2. Build a `PlaceholderGroup` with `Type = SparseFloatVector` and a single `Value = b'\x41' * 9` (9 bytes, obviously non-8-multiple).
3. Marshal the PlaceholderGroup into the `PlaceholderGroup` bytes field of a `SearchRequest` / `HybridSearchRequest`.
4. Submit to Proxy.

Observed: request NOT rejected at Proxy; malformed bytes are passed to segcore leading to heap out-of-bounds access.

Expected: Proxy rejects the request with parameter-invalid error mentioning `8-byte aligned`.

### Root Cause
Search / Hybrid PlaceholderGroup deserializes from an opaque bytes blob (proto-serialized client-side), and **only this path lacks the validation** that Insert already has.

### Steps To Reproduce

```markdown

```

pangwangshu:
[...]

PR #52430, opened 2026-08-11
fix: the sparse vector heap overflow issue (#52425)

Amos-yongxin:
issue: #52425

## Background
On the Search / HybridSearch request path, the sparse float vector query payload is delivered to proxy as a serialized commonpb.PlaceholderGroup blob. The downstream C++ accessor helpers SparseFloatRowIndexAt / SparseFloatRowValueAt are explicitly documented as // does not check for out-of-range access : each sparse element is expected to be 8-byte aligned ( uint32 idx + float32 val ). If a malformed row of length 9 bytes (or any other non-8B-aligned length) reaches segcore, indexing at position i * 8 reads past the heap allocation, producing an out-of-bounds read.

Insert / Upsert / Bulkload paths already call typeutil.ValidateSparseFloatRows inside validate_util.go on the FieldData and are therefore protected. Search / HybridSearch had zero validation on the PlaceholderGroup blob before this fix.

### 1. Add a single, focused validation gate (three-defense layered check)
search_util.go

Added validateSparseFloatVectorPlaceholderGroup(placeholderGroup []byte) error :

- ① Placeholder type filter : iterate the deserialized PlaceholderGroup, only process entries whose Type == PlaceholderType_SparseFloatVector ; wrong types are silently skipped (combined with tail check below, they fail closed).
- ② Per-row 8B alignment gate : for every sparse value in the placeholder, feed it through typeutil.ValidateSparseFloatRows (the same validator used by the Insert path). Returns ErrParameterInvalid on nil / misaligned length / unsorted indices / negative values.
- ③ Found-none tail check : if the placeholder group blob is non-empty but contains zero SparseFloatVector placeholders (e.g. FloatVector / BinaryVector payloads masquerading as a sparse field), return ErrParameterInvalid with a descriptive message.
### 2. Wire the gate into both reachable Search paths
task_search.go

- HybridSearch / SubSearch ( L700-L705 ): guard after IsFieldSparseFloatVector and before the NNZ metric + convertPlaceholder .
- Plain Search ( L1052-L1057 ): identical guard on t.FieldId and t.request.GetPlaceholderGroup() .
The ordering ( IsFieldSparse → validate → NNZ metric → convertPlaceholder ) is intentionally kept the same as the Insert-side "validate before use" policy.

### 3. unit tests
task_search_test.go

Added 2 helpers ( buildSparsePlaceholderGroup , misalignedSparseRow ) and 6 subtests:

- empty placeholder group (allowed)
- single / multiple valid rows (allowed, nq=1 / nq>1)
- wrong placeholder type → found none ErrParameterInvalid
- no query vectors → ErrParameterInvalid
- PoC : a 9-byte misaligned sparse row (the user-chosen canonical 8B violation payload) → correctly rejected with ErrParameterInvalid before the request reaches segcore.

### DIFF
--- a/internal/proxy/search_util.go
+++ b/internal/proxy/search_util.go
@@ -1107,7 +1107,7 @@ func getMetricType(toReduceResults []*internalpb.SearchResults) string {
 
 func validateSparseFloatVectorPlaceholderGroup(placeholderGroup []byte) error {
 	if len(placeholderGroup) == 0 {
-		return nil
+		return merr.WrapErrParameterInvalidMsg("placeholder group is empty")
 	}
 	pb := &commonpb.PlaceholderGroup{}
 	if err := proto.Unmarshal(placeholderGroup, pb); err != nil {
--- a/internal/proxy/search_util.go
+++ b/internal/proxy/search_util.go
@@ -1104,3 +1104,34 @@ func getMetricType(toReduceResults []*internalpb.SearchResults) string {
 	}
 	return ""
 }
+
+func validateSparseFloatVectorPlaceholderGroup(placeholderGroup []byte) error {
+	if len(placeholderGroup) == 0 {
+		return nil
+	}
+	pb := &commonpb.PlaceholderGroup{}
+	if err := proto.Unmarshal(placeholderGroup, pb); err != nil {
+		return merr.WrapErrParameterInvalidMsg("failed to unmarshal placeholder group: %v", err)
+	}
+
+	foundSparse := false
+	for _, h := range pb.GetPlaceholders() {
+		if h.GetType() != commonpb.PlaceholderType_SparseFloatVector {
+			continue
+		}
+		foundSparse = true
+		if len(h.GetValues()) == 0 {
+			return merr.WrapErrParameterInvalidMsg("sparse float vector search request has no query vectors")
+		}
+		if err := typeutil.ValidateSparseFloatRows(h.GetValues()...); err != nil {
+			return merr.WrapErrParameterInvalidMsg("invalid sparse float vector in search request: %v", err)
+		}
+	}
+
+	if !foundSparse {
+		return merr.WrapErrParameterInvalidMsg(
+			"sparse float vector field expects PlaceholderType_SparseFloatVector (%d) placeholders, found none (placeholder count=%d)",
+			int32(commonpb.PlaceholderType_SparseFloatVector), len(pb.GetPlaceholders()))
+	}
+	return nil
+}
--- a/internal/proxy/task_search.go
+++ b/internal/proxy/task_search.go
@@ -698,6 +698,9 @@ func (t *searchTask) initAdvancedSearchRequest(ctx context.Context) error {
 			return err
 		}
 		if typeutil.IsFieldSparseFloatVector(t.schema.CollectionSchema, internalSubReq.FieldId) {
+			if err := validateSparseFloatVectorPlaceholderGroup(internalSubReq.GetPlaceholderGroup()); err != nil {
+				return err
+			}
 			metrics.ProxySearchSparseNumNonZeros.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), t.collectionName, metrics.HybridSearchLabel, strconv.FormatInt(internalSubReq.FieldId, 10)).Observe(float64(typeutil.EstimateSparseVectorNNZFromPlaceholderGroup(internalSubReq.PlaceholderGroup, int(internalSubReq.GetNq()))))
 		}
 		internalSubReq.PlaceholderGroup = convertedPlaceholder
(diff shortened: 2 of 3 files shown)


internal/proxy/search_util.go
```diff
@@ -1104,3 +1104,34 @@ func getMetricType(toReduceResults []*internalpb.SearchResults) string {
 	}
 	return ""
 }
+
+func validateSparseFloatVectorPlaceholderGroup(placeholderGroup []byte) error {
+	if len(placeholderGroup) == 0 {
+		return nil
```
sunby: should we return an error here?

pangwangshu:
# Heap buffer overflow (OOB write) in SparseFloatVector Search/HybridSearch parsing — refines public issue #52425

## Summary

A `SparseFloatVector` query row submitted via `Search` or `HybridSearch` whose
byte length is not a multiple of 8 causes a **heap buffer overflow write** in
segcore's C++ placeholder-group deserialization — not merely an out-of-bounds
*read*, as currently described in public issue #52425. The unsafe write
happens unconditionally, before the function's own alignment check runs, so
even the code path that explicitly requests validation (`validate=true`)
does not prevent the corruption — it only detects it *after* the memory has
already been corrupted.

## Relationship to existing public reports

This refines the root cause described in public issue #52425 and reflected
in open PR #52430. Both currently attribute the unsafe behavior to
`SparseFloatRowIndexAt` / `SparseFloatRowValueAt`. Those are **Go**-side
accessor functions (`pkg/util/typeutil/schema.go:3846,3851`) and are not on
the vulnerable call path. I independently traced the actual C++ execution
path and found the real defect elsewhere, with a more severe failure mode
(write vs. read) than currently documented publicly.

I'm filing this privately rather than as a public comment because master is
currently unpatched (see below) and this level of mechanism detail is more
directly actionable than what's in the public issue today.

## Technical details

Call chain, each link confirmed by direct source inspection:

1. `internal/util/segcore/plan.go:128` — Go proxy/querynode boundary calls
   `C.ParsePlaceholderGroup(...)` with the raw, client-supplied placeholder
   group blob.
2. `internal/core/src/segcore/plan_c.cpp:93-101` — cgo shim forwards to
   `milvus::query::ParsePlaceholderGroup`.
3. `internal/core/src/query/Plan.cpp:128` — `ph_group.ParseFromArray(blob,
   blob_len)`, no length/alignment validation on the raw bytes.
4. `internal/core/src/query/Plan.cpp:171-174` — for
   `PlaceholderType::SparseFloatVector`, calls
   `SparseBytesToRows(ph.values(), /*validate=*/true)`.
5. `internal/core/src/common/Utils.h`, `SparseBytesToRows` (~L309-322) loops
   each row and calls `CopyAndWrapSparseRow(rows[i].data(), rows[i].size(),
   validate)`.
6. `internal/core/src/common/Utils.h`, `CopyAndWrapSparseRow` (~L272-300) —
   the actual defect:

   ```cpp
   inline knowhere::sparse::SparseRow<SparseValueType>
   CopyAndWrapSparseRow(const void* data, size_t size, const bool validate = false) {
       size_t num_elements = size / element_size();          // floor division
       knowhere::sparse::SparseRow<SparseValueType> row(num_elements);  // sized for num_elements * 8 bytes
       milvus::fastmem::FastMemcpy(row.data(), data, size);   // copies the FULL, unaligned `size` — unconditional
       if (validate) {                                        // only now, AFTER the copy above
           AssertInfo(size % element_size() == 0, "Invalid size for sparse row data");
           // ... per-element checks ...
       }
       return row;
   }
   ```

   For a `size` not a multiple of 8 (e.g. 9 bytes), `num_elements` floors
   down (9/8 = 1), the allocation is sized for `num_elements * 8` bytes, and
   the memcpy copies the full unaligned `size` into it — an overflow write
   of `size - num_elements*8` attacker-controlled bytes past the heap
   allocation. For `size < 8` (e.g. 3 bytes), `num_elements = 0`; depending
   on the knowhere `SparseRow(0)` implementation this can result in a write
   through a null/zero-capacity pointer.

   Because the `validate` block's checks run *after* the copy, passing
   `validate=true` (as the search path does) does not prevent the
   corruption — it only throws afterward, once memory may already be
   corrupted.

## Reachability

Confirmed reachable via the standard `Search` and `HybridSearch` gRPC/HTTP
APIs against any collection with a `SparseFloatVector` field, by supplying a
placeholder-group query vector whose byte length is not a multiple of 8
(e.g. 9 bytes). No special privileges beyond normal search access are
required.

## Relationship to PR #52430 (does not fully close this)

PR #52430 (open, CI currently failing as of this writing) adds Go-level
validation (`validateSparseFloatVectorPlaceholderGroup` in
`internal/proxy/search_util.go`) at the proxy boundary, wired into both the
`Search` and `HybridSearch` code paths in `task_search.go`. I independently
verified this validator's logic against the real, unmodified
`pkg/util/typeutil.ValidateSparseFloatRows` (isolated unit testing, ~12
cases including the exact misaligned-row trigger, multi-row batches, and
edge cases) and it correctly rejects misaligned rows before they are
forwarded to querynode. **This is an effective mitigation for the specific
reachable path described in #52425.**

However, PR #52430 does not modify `CopyAndWrapSparseRow` /
`SparseBytesToRows` in `internal/core`. The underlying C++ function remains
unsafe for any other current or future caller that passes row data with
`validate=true`, expecting the flag to mean "checked before use" rather
than "checked after the fact."

## Suggested fix (defense in depth, independent of PR #52430)

In `CopyAndWrapSparseRow` (`internal/core/src/common/Utils.h`), when
`validate` is true, perform the `size % element_size() == 0` alignment
check **before** computing `num_elements` / allocating / copying, so no
bytes are ever copied into an undersized buffer. Given the negligible cost
of a modulus check relative to the cost of a heap-corruption bug, consider
making the alignment check unconditional rather than gated behind
`validate`.

## Suggested severity

Heap buffer overflow write, network-reachable via an unauthenticated
[relative to this specific check] search request against a
`SparseFloatVector` field — meaningfully more severe than the OOB-read
classification in the current public issue.

internal/proxy/search_util.go
Amos-yongxin: Yes, an error should indeed be returned here. It seems there's no subsequent interception of empty sparse vectors.

sunby: > Yes, an error should indeed be returned here. It seems there's no subsequent interception of empty sparse vectors.

Should also change ut.

Issue #52433, opened 2026-08-11
[Enhancement]: avoid row-wise payload copies when repacking contiguous insert batches


sunby:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Add a fast path in proxy insert repacking for batches whose row offsets form one contiguous range.

Today, `genInsertMsgsByPartition` reconstructs every outgoing protobuf column by calling `typeutil.AppendFieldData` once per selected row. For a contiguous batch, the equivalent `FieldData`, `RowIDs`, `Timestamps`, and `HashValues` can instead be represented by capacity-limited sub-slices that share the source backing arrays until serialization. Non-contiguous batches and unsupported field variants should continue to use the existing row-wise path.

### Why is this needed?

The row-wise path grows destination slices and copies scalar, dynamic JSON, and vector payloads before protobuf serialization. This creates avoidable short-lived allocations and GC work in proxy insert repacking.

An isolated synthetic benchmark using 1,024 rows with a 768-dimensional float vector measured the following allocation cost:

```text
range view     280 B/op          6 allocs/op
append rows    16,768,668 B/op  37 allocs/op
```

The benchmark excludes protobuf marshaling and is not an end-to-end throughput or latency measurement.

### Anything else?

Related issue #52264 and PR #52270 optimized destination-column lookup inside `AppendFieldData`, reducing CPU overhead but intentionally leaving allocation counts unchanged. This proposal targets the separate payload-copy cost in the insert repack path.

The streaming message builder synchronously marshals the protobuf body before returning, so the range views do not introduce an asynchronous alias lifetime. The existing row-wise implementation remains the fallback for non-contiguous row selections and any field representation that cannot safely use a range view.

xiaofan-luan:
[...]

PR #52434, opened 2026-08-11
enhance: avoid copies for contiguous insert batches

sunby:
issue: #52433

## What

- Plan insert message batches using the existing message-size threshold semantics.
- For a batch whose row offsets form one contiguous range, build capacity-limited protobuf range views for `FieldData`, `HashValues`, `Timestamps`, and `RowIDs` instead of appending one row at a time.
- Preserve `AppendFieldData` as the fallback for non-contiguous selections and field variants that cannot safely use a range view.
- Cover scalar fields, dynamic JSON, dense and sparse vector encodings, vector arrays, and compact nullable vector payloads.

## Why

`genInsertMsgsByPartition` currently calls `AppendFieldData` for every selected row. That grows destination slices and copies the payload before protobuf serialization. When routing produces a contiguous row range, the outgoing message can reference bounded sub-slices of the original columns and avoid that intermediate payload copy.

PR #52270 optimized destination-column lookup inside `AppendFieldData`, but intentionally did not reduce allocation counts. This change targets the separate copy/allocation cost in the insert repack path.

## Safety and compatibility

- Three-index slices limit capacity, so an append to a view cannot overwrite the following source rows.
- The streaming message builder synchronously marshals the protobuf body before returning, so the views do not escape into an asynchronous alias lifetime.
- Message splitting retains the previous threshold behavior, including allowing one oversized row in a non-empty message.
- Non-contiguous row offsets and unsupported field variants retain the existing row-wise implementation.

## Allocation benchmark

An isolated synthetic benchmark constructs a batch with 1,024 rows and a 768-dimensional float vector. Protobuf marshaling is excluded:

```text
range_view:   280 B/op          6 allocs/op
append_rows:  16,768,668 B/op  37 allocs/op
```

These are component-level allocation numbers, not an end-to-end proxy latency or throughput claim.

## Verification

Passed on the current base:

```text
cd pkg
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./util/typeutil
```

The isolated Proxy fast-path tests and Proxy build passed on an earlier revision of this implementation. They were not rerun after moving to the current `master`: the available local C++ artifacts had an incompatible ABI, and the rebuild was intentionally stopped. The PR is therefore opened as a draft pending current-base CI verification.


sunby:
this can reduce memory allocation and avoid frequent slow gc.

alloc_space perf in 5min:
before:
image
after:
image

internal/proxy/channelmgr/msg_pack.go
```diff
@@ -117,22 +141,66 @@ func GenInsertMsgsByPartition(ctx context.Context,
...
+			appendIdxComputer = typeutil.NewFieldDataIdxComputer(fieldsData)
+		}
+		msg.FieldsData = make([]*schemapb.FieldData, len(fieldsData))
+		for _, offset := range batchOffsets {
+			fieldIdxs := appendIdxComputer.Compute(int64(offset))
+			typeutil.AppendFieldData(msg.FieldsData, fieldsData, int64(offset), fieldIdxs...)
+			msg.HashValues = append(msg.HashValues, insertMsg.HashValues[offset])
+			msg.Timestamps = append(msg.Timestamps, insertMsg.Timestamps[offset])
+			msg.RowIDs = append(msg.RowIDs, insertMsg.RowIDs[offset])
+			msg.NumRows++
```
buqian-zilliz: **[P2] Contiguous fast path shares backing arrays with the source insert message**

The new contiguous fast path installs sub-slices of the original `insertMsg` backing arrays directly into each packed message instead of copying: `msg.FieldsData = fieldViews` (where `CreateFieldDataRangeView` returns per-field data and `ValidData` slices that alias the source), plus `msg.HashValues`, `msg.Timestamps`, and `msg.RowIDs` assigned from `insertMsg.<field>[rowStart:rowEnd:rowEnd]`. The three-index slice caps capacity so future appends will reallocate, but any in-place write (element assignment, `copy`, sorting) to one packed message would corrupt the shared source `insertMsg` and every sibling packed message for the same partition/channel. The previous implementation built fresh copies, so this is a behavioral change. Please verify every downstream path (including cipher/encryption and streaming-node handling) treats these messages as read-only before merging.

internal/proxy/channelmgr/msg_pack.go
```diff
@@ -117,22 +145,66 @@ func GenInsertMsgsByPartition(ctx context.Context,
...
+				continue
+			}
+		}
+
+		if appendIdxComputer == nil {
+			appendIdxComputer = typeutil.NewFieldDataIdxComputer(fieldsData)
+		}
+		msg.FieldsData = make([]*schemapb.FieldData, len(fieldsData))
+		for _, offset := range batchOffsets {
+			fieldIdxs := appendIdxComputer.Compute(int64(offset))
```
buqian-zilliz: **[P2] Avoid repeating field-index scans on the fallback path**

The sizing pass already calls `sizeIdxComputer.Compute` for every selected row, but any batch that is not contiguous creates `appendIdxComputer` and recomputes the same indexes here before copying. Sparse row-offset lists are the normal result of routing inserts across channels and partitions, so the fallback adds a second `O(fields * rows)` traversal relative to the previous single-pass implementation; for compact nullable vectors each new computer also rescans `ValidData` from row zero up to the selected offsets. At production row counts and shard counts this can materially increase proxy CPU on the common fallback path. Please preserve the computed physical indexes from the sizing pass or keep a single-pass copy path for batches that cannot use a range view.

pkg/util/typeutil/schema.go
```diff
@@ -1023,6 +1023,281 @@ func (c *FieldDataIdxComputer) Compute(rowIdx int64) []int64 {
...
+// only protobuf wrapper objects and slice headers are newly allocated. It
+// returns false when the range or field variant cannot safely use a view.
+//
+// dataStarts and dataEnds are the physical row bounds returned by
+// FieldDataIdxComputer. They differ from the logical bounds for compact
+// nullable vector payloads.
+func CreateFieldDataRangeView(
+	src []*schemapb.FieldData,
+	rowStart, rowEnd int64,
+	dataStarts, dataEnds []int64,
```
buqian-zilliz, line 1037: **[P2] Make nullable-vector bounds independent snapshots**

`CreateFieldDataRangeView` documents `dataStarts` and `dataEnds` as values returned by `FieldDataIdxComputer`, but `Compute` returns the same reusable `resultBuffer` on every call. The natural call `CreateFieldDataRangeView(..., c.Compute(rowStart), c.Compute(rowEnd))` therefore passes the end indexes for both slices. For a nullable vector this produces `physicalRows == 0`, so the function emits `ValidData` containing true rows while omitting their vector payload, silently corrupting the insert. The current caller avoids this only by manually copying the start slice. Please make the helper compute or own both bounds, or change the API so two independent snapshots are required and this invalid aliasing state cannot be constructed.

internal/proxy/channelmgr/msg_pack.go
buqian-zilliz: This issue still exists: non-contiguous batches still recompute every field index after the sizing pass, including another validity-bitmap scan for compact nullable vectors.

pkg/util/typeutil/schema.go
buqian-zilliz: This issue still exists: `CreateFieldDataRangeView` still accepts separate `dataStarts` and `dataEnds` slices even though `FieldDataIdxComputer.Compute` returns one reusable buffer, so two direct `Compute` calls can alias both bounds to the end snapshot and omit compact nullable-vector payloads.

sunby: Agreed that two direct `Compute` calls can alias its reusable buffer. However, both production call sites already copy the start indexes into independent storage before the next call: [`dataStarts`](https://github.com/milvus-io/milvus/blob/a883be4d1a9ab418d87bf91e84c0d30ecdc9614c/internal/proxy/channelmgr/msg_pack.go#L253) and [`firstFieldIdxs`](https://github.com/milvus-io/milvus/blob/a883be4d1a9ab418d87bf91e84c0d30ecdc9614c/internal/proxy/channelmgr/msg_pack.go#L217). The singleton path also uses a separate computer for the end bound, so the current insert path does not exhibit this aliasing corruption.

I'll keep the API unchanged in this PR. The misuse risk is valid, but making the helper enforce independent bounds would be a separate API-hardening change.

internal/proxy/channelmgr/msg_pack.go
```diff
@@ -99,12 +128,60 @@ func GenInsertMsgsByPartition(ctx context.Context,
 	}
 
 	fieldsData := insertMsg.GetFieldsData()
+	_, _, contiguous := contiguousRowRange(rowOffsets)
```
buqian-zilliz: **[P2] Preserve range views for contiguous batches in mixed selections**

`contiguous` is computed once for the entire `rowOffsets` slice, so one gap sends every multi-row size batch through `appendRow`, including batches whose own offsets are contiguous. Channel routing preserves offset order, and with a small number of channels a production insert can contain contiguous runs separated by gaps; the global gate therefore copies all field payloads for those otherwise viewable batches, potentially many megabytes. Track contiguity per planned batch or run and use `CreateFieldDataRangeView` for each contiguous batch, falling back only for batches that actually contain a gap.

zhengbuqian (approved):

Issue #52125, opened 2026-08-03
[Bug]: Azure CopyObject uses async StartCopyFromURL without polling copy status — snapshot restore can report success before data is copied


huanghaoyuanhhy:
## Environment

- **Milvus Version**: master @ 24e43c6dfd (snapshot restore via server-side copy)
- **Deployment Mode**: cluster / standalone
- **MQ**: n/a
- **SDK**: n/a (reproducible via Go/Python SDK + Azure REST)
- **OS**: n/a

### Option B: Steps (systemic / timing issue)

1. Deploy: Milvus with object storage provider `azure` (any account config; same-account copy)
2. Setup: create a collection with a large vector field, insert enough data to produce
   multi-GB sealed segments, flush, build index
3. Workload: call `CreateSnapshot` then `RestoreSnapshot`, wait for the restore job to
   report success (`GetRestoreSnapshotState` → completed)
4. Trigger: immediately after restore reports success, call `Get Blob Properties`
   (HEAD) on the destination blobs under the restored collection's path
5. Observed via: destination blob response headers — `x-ms-copy-status`,
   `x-ms-copy-progress`, `x-ms-copy-id`; then load/query the restored collection

### Trigger Conditions

- **Frequency**: always, on Azure, for blobs large enough that the background copy
  doesn't finish within the restore job's copy loop
- **First observed after**: restore of large segments (copy takes seconds to minutes)
- **Does NOT happen when**: provider is S3/MinIO or GCS (synchronous copy APIs);
  on Azure only when the background copy happens to finish before any consumer
  touches the destination

## Expected Behavior

`ChunkManager.Copy` / `ObjectStorage.CopyObject` on Azure should behave like the
S3/GCS implementations: return only after the object is fully copied (or fail with
an error). Restore must not report a segment as copied while its files are still
being copied in the background.

## Actual Behavior

`AzureObjectStorage.CopyObject` (`internal/storage/azure_object_storage.go:198`)
calls `StartCopyFromURL` — Azure's **asynchronous** `Copy Blob` API, which returns
202 Accepted immediately with `x-ms-copy-status: pending` — and returns nil without
polling the copy status. Consequences:

1. Restore (copy-segment job) reports success while segment files are still copying
   in the background; the destination is a zero-length/partial blob until the copy
   finishes. A read of restored data can return partial or empty content.
2. If the source blob is modified during the copy, Azure fails the copy in the
   background (`x-ms-copy-status: failed`); Milvus never notices and there is no
[...]

PR #50978, opened 2026-07-01
feat: support external snapshot export and restore

weiliu1031:
issue: https://github.com/milvus-io/milvus/issues/44358

Part 1 API contract PR: https://github.com/milvus-io/milvus/pull/50393
Design doc: https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260609-external-snapshot-export-restore.md

This PR implements the external snapshot export and restore flow on top of the merged API contract:

- Resolve and validate request-scoped external snapshot storage.
- Copy snapshot data through provider-side same-provider object copy.
- Export self-contained bundles with bounded copy concurrency.
- Restore both referenced snapshots and self-contained exported bundles.
- Rebase relocated bundle roots while preserving the bundle-internal layout.
- Propagate external source storage through DataCoord jobs and DataNode copy tasks.
- Keep snapshot storage tests in the owning internal/snapshotio/storage package.

Validation:

- `git diff --check`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 github.com/milvus-io/milvus/internal/snapshotio/storage`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 github.com/milvus-io/milvus/internal/datacoord -run "Test(SnapshotExporter|SnapshotManager_ExportSnapshot|RestoreExternalSnapshot|CreateRestoreJob_External)"`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 github.com/milvus-io/milvus/internal/datanode/importv2 -run "Test(CopySegmentAndIndexFiles|CopySegmentTask|CollectSegmentFiles|GenerateTargetPath)"`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 github.com/milvus-io/milvus/internal/datanode -run "Test(FirstExternalSourceURI|DataNodeServicesSuite)"`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 github.com/milvus-io/milvus/internal/storage -run "TestRemoteChunkManagerCopyCrossBucket"`

The feature commit previously passed `make milvus` and the referenced/self-contained Go client E2E cases. Those broader checks were not rerun for the latest test-only cleanup.

PR #52440, opened 2026-08-11
fix: Wait for Azure blob copy completion

weiliu1031:
issue: #52125

## What changed

- Poll the destination blob while Azure reports the copy as pending.
- Return success only after Azure reports copy success.
- Preserve context cancellation and return an I/O error for failed or
  unknown copy statuses.

## Why

`StartCopyFromURL` can return before the server-side copy finishes. Snapshot
restore could therefore publish segment metadata before the destination object
was ready.

## Validation

- `GOTOOLCHAIN=auto make build-go`
- Targeted race tests passed for all 10 Azure copy status cases.
- `CopyObject` coverage: 100%.
- The full storage suite reached an unchanged local Azure endpoint setup
  failure.


weiliu1031:
This PR is superseded by the merged [PR #50978](https://github.com/milvus-io/milvus/pull/50978).

During the rebase onto the latest master, both changed Azure storage files conflicted because [PR #50978](https://github.com/milvus-io/milvus/pull/50978) had already introduced a more complete solution for the same root cause. Its implementation waits for asynchronous Azure copies to complete and additionally handles pending-copy recovery, transient polling failures, copy identity validation, cancellation, and timeout.

After accepting the upstream implementation, this branch had no remaining diff from master. The PR was therefore closed without merge.

Issue #52239, opened 2026-08-06
Support IS NULL and IS NOT NULL predicates for vector fields


marcelo-cjl:
## Background

Milvus already has nullable vector storage and search support for ordinary vector fields (`FloatVector`, `BinaryVector`, `Float16Vector`, `BFloat16Vector`, `SparseFloatVector`, and `Int8Vector`). The vector payload is stored compactly for valid rows, while a logical-row validity bitmap is kept in field data and vector index offset mapping (`valid_data` / `valid_data_count`).

However, filter expressions currently do not support `IS NULL` / `IS NOT NULL` on ordinary vector fields.

## Current Behavior

The parser rejects ordinary vector fields for null predicates:

- `internal/parser/planparserv2/parser_visitor.go`
- `isUnsupportedNullExprVectorType()` returns true for `typeutil.IsVectorType(dataType) && !typeutil.IsVectorArrayType(dataType)`
- Parser tests currently list ordinary vector fields as invalid for `is null` / `is not null`.

If the parser guard is removed alone, segcore still cannot execute the expression:

- `internal/core/src/exec/expression/NullExpr.cpp`
- `PhyNullExpr::Eval()` does not handle `VECTOR_FLOAT`, `VECTOR_BINARY`, `VECTOR_FLOAT16`, `VECTOR_BFLOAT16`, `VECTOR_SPARSE_U32_F32`, or `VECTOR_INT8`.
- The current inherited execution-path logic may treat a vector index as a scalar index, which is incorrect for a vector null predicate.

## Expected Behavior

For nullable vector fields:

- `<vector_field> is null` matches logical rows whose vector value is null.
- `<vector_field> is not null` matches logical rows whose vector value exists.

For non-nullable vector fields, keep the same semantics as scalar fields:

- `<vector_field> is null` returns false for every visible row.
- `<vector_field> is not null` returns true for every visible row.

The behavior should work in query/search filters for:

- growing segments;
- sealed segments;
- vector fields with and without vector indexes;
- StorageV3 / raw-data-in-index cases where vector fielddata is not resident;
- offset-input evaluation paths.

## Required Implementation Direction

1. Parser
   - Allow `IS NULL` / `IS NOT NULL` on ordinary nullable-vector-capable vector types.
   - Keep existing rejection for unsupported array element access, such as `array_field[0] is null`.

2. Segcore expression execution
   - Add a vector-specific validity-only path in `PhyNullExpr`.
[...]

PR #50524, opened 2026-06-14
enhance: move nullable vector mapping into index layer

marcelo-cjl:
## Summary
- Move nullable vector id mapping into the vector index/Knowhere boundary. Milvus now passes logical row/list ids and logical bitsets to vector APIs, and indexed paths return logical ids directly.
- Pass nullable valid-data metadata to Knowhere as `IdMapData` during vector index build/load/add. Knowhere owns the compact internal id map, including mmap-backed dense id arrays for sealed indexes.
- Keep Milvus `OffsetMapping` only for compact raw/growing brute-force views. Raw BF passes chunk-local physical-to-logical windows to Knowhere `BitsetView::out_ids`, so BF can test logical bitsets and return logical ids without building a Knowhere `IdMap`.
- Remove the old segcore indexed-path bitset/result-id transforms. VECTOR_ARRAY element-level results still use array offsets; row-level nullable maps are not applied to element ids.
- Preserve empty/all-null vector and embedding-list handling so empty index metadata can still carry nullable id-map state.
- Depend on zilliztech/knowhere#1673 at `92fd89bc3c66a2a68971c6d4eab563fa8152452c`, which provides dataset-carried `IdMapData`, snapshot-backed `IdMap`, mmap-backed sealed id arrays, and brute-force out-id mapping.

Related dependency: zilliztech/knowhere#1673

## Test plan
- [x] `git diff --check upstream/master..HEAD`
- [x] `git diff --name-only --diff-filter=ACMRTUXB upstream/master..HEAD -- '*.h' '*.hpp' '*.cc' '*.cpp' '*.c' | xargs -r clang-format-15 --dry-run -Werror`
- [x] `env jobs=2 CONAN_CPU_COUNT=2 GOMAXPROCS=2 CMAKE_EXTRA_ARGS="-DFETCHCONTENT_SOURCE_DIR_KNOWHERE=$PWD/dependency-overrides/knowhere -DFETCHCONTENT_UPDATES_DISCONNECTED_KNOWHERE=ON" make build-cpp-with-unittest jobs=2`
- [x] `all_tests` split by suite range with Knowhere #1673 overlay: 7624 run / 7616 passed / 8 skipped
- [x] `./internal/core/output/unittest/all_tests --gtest_filter='JsonNumericTest.*'`: 15 passed
- [x] `./internal/core/output/unittest/json_stats_test`: 79 passed

PR #52442, opened 2026-08-11
fix: support vector null predicates

marcelo-cjl:
issue: #52239

## Summary
- Support IS NULL and IS NOT NULL predicates for ordinary vector fields.
- Read logical-row validity through vector-index HasValidData/IsRowValid when raw field data is unavailable, remaining compatible with #50524's Knowhere IdMap ownership.
- Rebase onto current master while preserving master's broader search-by-PK result reordering and tests.

## Test Plan
- git diff --check origin/master...HEAD
- go test ./internal/parser/planparserv2
- go test -c -o /tmp/milvus-pr-52442-testcases.test ./testcases (from tests/go_client)
- ruff check tests/python_client/chaos/checker.py tests/python_client/milvus_client/test_milvus_client_collection.py tests/python_client/milvus_client/test_milvus_client_query.py tests/python_client/milvus_client/test_milvus_client_search.py
- python3 -m py_compile tests/python_client/chaos/checker.py tests/python_client/milvus_client/test_milvus_client_collection.py tests/python_client/milvus_client/test_milvus_client_query.py tests/python_client/milvus_client/test_milvus_client_search.py
- Private Release+ASan integration build at d839d36b0c956187fbf8ae20e8381a50b9b1837c


internal/core/src/exec/expression/NullExpr.cpp
```diff
@@ -137,6 +149,76 @@ PhyNullExpr::DetermineExecPath() {
...
+        }
+        return;
+    }
+
+    SegmentExpr::MoveCursor();
+}
+
+bool
+PhyNullExpr::CanExecuteAllAtOnce() const {
+    if (IsOrdinaryVectorDataType(expr_->column_.data_type_)) {
```
liliu-z, line 167: DetermineExecPath() pins every vector type to the RawData path (NullExpr.cpp:139-141) while CanExecuteAllAtOnce() is overridden to return true for ordinary vectors, which is the combination GISConjunctExpr.h:285-295 documents as unsafe: FilterBitsNode then calls SetExecuteAllAtOnce(), setting batch_size_ = active_count_, and ExecVectorNull materializes one int64 offset per segment row before handing the whole array to ApplyFieldValidDataByOffsets. On the index path that is only a transient 8-bytes-per-row allocation, but when the field has no vector index entry (interim index disabled or segment below the build ratio) or an index without valid data, the call lands in ChunkedColumn/ChunkedColumnGroup BulkIsValid, which resolves the full-segment offset list into cids and pins every chunk of the column at once instead of one batch's worth. Consider leaving the vector path batched so the fallback stays chunk-bounded; the PrefetchRawData() no-op at line 176 is fine as-is since the base version would warm the entire vector column.

internal/core/src/common/GrowingOffsetMapping.cpp
```diff
@@ -394,4 +394,47 @@ GrowingOffsetMapping::FilterValidLogicalOffsets(
     }
 }
 
+void
+GrowingOffsetMapping::ApplyValidDataByLogicalOffsets(
+    const int64_t* logical_offsets,
+    int64_t count,
+    TargetBitmapView valid_result) const {
+    const auto counts = LoadCounts();
+    if (counts.total == 0) {
```
liliu-z: The new ApplyValidDataByLogicalOffsets treats counts.total == 0 as "every row is null", which is the opposite of what FilterValidLogicalOffsets in the same file documents for that state ("No mapping: logical and physical spaces coincide. Mirror the other conversions (and the NoOp/Sealed implementations) instead of rejecting every offset against total_count == 0") and the opposite of the sibling overrides added by this PR — SealedOffsetMapping returns without touching valid_result when disabled and NoOpOffsetMapping only rejects negative offsets. It is not reachable today because both call sites gate on vec_index->HasValidData(), which is exactly IsEnabled() == (total != 0), but any future caller reaching it directly would silently report every row as null.

marcelo-cjl: Agreed. `counts.total == 0` means the growing mapping is disabled, not that every logical row is invalid. I aligned `ApplyValidDataByLogicalOffsets()` with `FilterValidLogicalOffsets`/NoOp/disabled sealed semantics: disabled mapping leaves non-negative logical offsets valid and only masks negative offsets. I also added disabled-growing coverage for that behavior.

internal/core/src/exec/expression/NullExpr.cpp
marcelo-cjl: Good catch. The vector null path only needs field validity, but forcing all-at-once on a RawData path made the fallback lose its batch boundary. I changed `PhyNullExpr::CanExecuteAllAtOnce()` so ordinary vector null evaluation stays batched on the RawData path, while keeping vector prefetch as a no-op so vector payloads are not warmed.

Issue #52443, opened 2026-08-11
[Enhancement]: Avoid intermediate map allocation for metacache segment ID filters


bigsheeper:
## What would be improved

`metaCacheImpl.rangeWithFilter` builds an intermediate map when a `WithSegmentIDs(...)` filter is present. The filter already identifies the exact segment IDs to probe, so `rangeWithFilter` can iterate the ID set and perform direct lookups in the candidate map instead of materializing a second map before applying the remaining filters.

This affects ID-filtered `GetSegmentsBy`, `GetSegmentIDsBy`, `UpdateSegments`, and `RemoveSegments` calls in DataNode flush/write-buffer paths.

## Proof

Existing behavior-pinning tests pass before and after the change:

```bash
GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/metacache
```

Result: `ok github.com/milvus-io/milvus/internal/flushcommon/metacache`.

Benchmark, measured before and after in the same session on the same machine:

```bash
GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
go test -tags dynamic,test -gcflags="all=-N -l" -run '^$' -bench 'BenchmarkGetSegmentsBy$' -benchmem -count=5 ./internal/flushcommon/metacache
```

Median of 5 runs:

| benchmark | before | after |
|---|---:|---:|
[...]

PR #52444, opened 2026-08-11
enhance: speed up metacache segment id filters

bigsheeper:
issue: #52443

## What changed

`metaCacheImpl.rangeWithFilter` now handles `WithSegmentIDs(...)` by iterating the requested IDs and probing each candidate map directly. This avoids materializing an intermediate map before applying the remaining filters.

## Why

ID-filtered metacache lookups are used by DataNode flush and write-buffer paths such as segment state updates and start-position bookkeeping. The existing ID criterion already narrows the candidate set, so a direct lookup preserves behavior while reducing allocation and lookup work.

## Verification

Behavior-pinning tests:

```bash
GOWORK=off PKG_CONFIG_PATH=/home/sheep/workspace/milvus/internal/core/output/lib/pkgconfig \
LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
LD_LIBRARY_PATH=/home/sheep/workspace/milvus/internal/core/output/lib:/home/sheep/workspace/milvus/internal/core/output/lib64:/home/sheep/workspace/milvus/cmake_build/lib \
go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/flushcommon/metacache
```

Result: passed.

Benchmark, median of 5 runs measured before and after in the same session:

| benchmark | before | after |
|---|---:|---:|
| `BenchmarkGetSegmentsBy` time | 1555 ns/op | **684.7 ns/op (2.27x faster)** |
| `BenchmarkGetSegmentsBy` bytes | 480 B/op | **272 B/op (-43%)** |
| `BenchmarkGetSegmentsBy` allocs | 10 allocs/op | **6 allocs/op (-40%)** |

Static check:

```bash
GOWORK=off make static-check
```

Result: passed. Local note: the nested worktree reused the main checkout's C++ `milvus_core.pc`/libraries and a locally generated ignored `cmd/tools/migration/legacy/legacypb` package to satisfy the current static-check environment.

### DIFF
--- a/internal/flushcommon/metacache/meta_cache.go
+++ b/internal/flushcommon/metacache/meta_cache.go
@@ -248,19 +248,17 @@ func (c *metaCacheImpl) rangeWithFilter(fn func(id int64, info *SegmentInfo), fi
 	}
 
 	for _, candidate := range candidates {
-		var segments map[int64]*SegmentInfo
 		if criterion.ids != nil {
-			segments = lo.SliceToMap(lo.FilterMap(criterion.ids.Collect(), func(id int64, _ int) (*SegmentInfo, bool) {
+			for id := range criterion.ids {
 				segment, ok := candidate[id]
-				return segment, ok
-			}), func(segment *SegmentInfo) (int64, *SegmentInfo) {
-				return segment.SegmentID(), segment
-			})
-		} else {
-			segments = candidate
+				if ok && criterion.Match(segment) {
+					fn(id, segment)
+				}
+			}
+			continue
 		}
 
-		for id, segment := range segments {
+		for id, segment := range candidate {
 			if criterion.Match(segment) {
 				fn(id, segment)
 			}


czs007:

**Verified:**
- `internal/flushcommon/metacache/meta_cache.go:233-266` — the new ID fast path keeps the same lock scope and still applies `criterion.Match(...)` after the ID lookup, so additional filters and state scoping remain enforced.
- `internal/flushcommon/metacache/meta_cache.go:239-247` — when a state filter is also present, only the corresponding state indexes are searched, and missing or unknown state maps remain safe to range over.
- `internal/flushcommon/metacache/actions.go:49-67` — `WithSegmentIDs(...)` always builds a non-nil set, so the direct `range` path cannot introduce duplicates or miss requested IDs, and an empty ID filter still returns no results as before.
- `internal/flushcommon/metacache/meta_cache.go:157-172` — removal callbacks receive the candidate map key, which is the segment ID maintained by both cache indexes, so deletion stays consistent.
- `internal/flushcommon/metacache/meta_cache_test.go:107-155` and `internal/flushcommon/pipeline/data_sync_service_test.go:299-304` — existing tests cover ID-filtered retrieval, updates and removals across state changes, plus combined ID and state filters for growing and flushed segments.
- `internal/flushcommon/metacache/meta_cache_test.go:241-305` — the added benchmark exercises the ID-filter and non-ID paths separately without altering production behavior.

No further findings survived verification, so there are no severity sections, non-blocking items, disputed items, or pre-existing hazards to report.

PR #50812, opened 2026-06-25
enhance: isolate online-write cgo pool from segment loading and add pool metrics

xiaofan-luan:
issue: #50811
related: #49435

## What

On the QueryNode, the load-time delta-log replay (`LoadDeltaData`, `UseLoad=true`) and the online forward delete (`segment.Delete`) both submit their CGO onto the **same `DynamicPool`** (size `CPUNum`). After a compaction, the multi-partition replay burst exhausts the pool, starving online `Insert`/`Delete`. Because `deleteNode.Operate` runs `ProcessDelete` before `UpdateTSafe`, channel **tSafe freezes for tens of seconds**, and Strong-consistency queries fail with `channel tsafe stalled` (#49435).

Evidence from the failing master run (traceID `b31d6e1626bd4d368961c89e6d4a5652`): `update tsafe` / `start to process delete` had a ~19–27s gap exactly overlapping a post-compaction segment load; `milvus_querynode_pool_active_threads{pool_name="DynamicPool"}` was pegged at **8/8 for ~27s** while `LoadPool` sat idle at **0/40**.

## Changes

Pool isolation + cleanup (`internal/querynodev2/segments`):
- Add a dedicated **MutatePool** for the online-write CGO path (`segment.Insert`/`segment.Delete`), isolated from load/management work.
- Route load-time **`LoadDeltaData`** replay and **BM25 stats** load onto **LoadPool**.
- Remove orphaned **WarmupPool**; merge away **BM25LoadPool**.
- Keep **DeletePool** (it is the `DeleteBatch` dispatch pool, not orphaned).
- Make **DynamicPool**/**DeletePool** size-configurable + hot-reloadable (`CPUNum * factor`, default `CPUNum`); add `MutatePoolSizeFactor` / `DynamicPoolSizeFactor` / `DeletePoolSizeFactor`.

Observability:
- Export `capacity / active_threads / queue_depth` for all QueryNode pools (`milvus_querynode_pool_*`, MutatePool added) and, newly, for the important DataNode pools (`milvus_datanode_pool_*`: compaction-exec, index-build, import-exec, io, stats).

## Notes

- This fixes the **resource-exhaustion** half of #49435. The `deleteMut` lock decoupling (moving the worker RPC out of the lock so it only guards the in-memory catch-up→publish boundary) is tracked as follow-up.
- The pool "priority" coefficients are pool-size only (no OS scheduling priority); isolation guarantees online-write admission, not CPU preemption under full saturation.

[...]

PR #52446, opened 2026-08-11
fix: isolate streaming insert/delete applies onto a dedicated CGO pool (#49435)

tinswzy:
Related issue: #49435

## What & why

The tsafe-critical streaming applies `C.Insert` / `C.Delete` (the insert/delete flowgraph nodes that advance channel tsafe) shared the CPU-core-sized `dynamicPool` with the bulk deltalog apply `C.LoadDeletedRecord`. On a CPU-constrained standalone under a segment-load storm, the giant `LoadDeletedRecord` calls (tens of seconds each, aggravated by CFS throttling) saturate every pool slot, so the tiny insert/delete applies queue behind them, the flowgraph stalls before `UpdateTSafe`, and channel tsafe freezes — surfacing as `code=505 channel tsafe stalled` on concurrent Strong queries during heavy upsert.

This routes the streaming insert/delete applies to the dedicated `deletePool`, separate from the shared `dynamicPool`. They then keep their own pool slots and advance tsafe regardless of how saturated or slow the shared pool gets. It also adds `cgo_queue_latency` (a histogram of pool-slot wait time per cgo call, labelled by `cgo_name`/`pool`) to make the contention observable.

## Validation

fouram concurrent case (heavy upsert + Strong `count(*)`, 8-core standalone), with `dataCoord.sealPolicy.channel.blockingL0EntryNum` left at the default so the delete buffer still peaks ~8.9M and `LoadDeletedRecord` still takes 30–60s (every stall precondition present): **8/8 runs, 0 tsafe stalls**, while CFS throttle stayed 0.62–1.00 and the insert/delete pool-wait dropped to **1–56ms**. Full root-cause analysis and A/B data in #49435.

issue: #49435

### DIFF
--- a/internal/querynodev2/segments/segment.go
+++ b/internal/querynodev2/segments/segment.go
@@ -839,7 +839,15 @@ func (s *LocalSegment) Delete(ctx context.Context, primaryKeys storage.PrimaryKe
 	// been applied to this segment, and skipping them causes silent data loss.
 
 	var err error
-	GetDynamicPool().Submit(func() (any, error) {
+	// issue #49435: run the streaming delete-apply on the dedicated deletePool
+	// instead of the shared dynamicPool. dynamicPool is also used by the bulk
+	// deltalog load (C.LoadDeletedRecord in LoadDeltaData), which can take tens
+	// of seconds per call under a segment-load storm. Sharing one pool makes the
+	// tiny (~ms) tsafe-critical forward delete queue behind those giant loads,
+	// starving the delegator's ProcessDelete -> UpdateTSafe and freezing the
+	// channel tsafe (505 channel tsafe stalled). A separate pool keeps the
+	// delete-apply off that queue.
+	GetDeletePool().Submit(func() (struct{}, error) {
 		start := time.Now()
 		defer func() {
 			metrics.QueryNodeCGOCallLatency.WithLabelValues(
@@ -852,7 +860,7 @@ func (s *LocalSegment) Delete(ctx context.Context, primaryKeys storage.PrimaryKe
 			PrimaryKeys: primaryKeys,
 			Timestamps:  timestamps,
 		})
-		return nil, nil
+		return struct{}{}, nil
 	}).Await()
 
 	if err != nil {
--- a/internal/querynodev2/segments/segment.go
+++ b/internal/querynodev2/segments/segment.go
@@ -784,7 +784,17 @@ func (s *LocalSegment) Insert(ctx context.Context, rowIDs []int64, timestamps []
 
 	var result *segcore.InsertResult
 	var err error
-	GetDynamicPool().Submit(func() (any, error) {
+	submitTime := time.Now()
+	// issue #49435: run the tsafe-critical growing-segment insert on the
+	// dedicated deletePool (the streaming-pipeline apply pool, shared with the
+	// delete-apply) rather than the shared dynamicPool that the bulk deltalog
+	// load (C.LoadDeletedRecord) can hold for tens of seconds. insertNode is
+	// upstream of deleteNode on the tsafe pipeline, so a queued insert freezes
+	// tsafe just like a queued delete does.
+	GetDeletePool().Submit(func() (struct{}, error) {
+		metrics.QueryNodeCGOQueueLatency.WithLabelValues(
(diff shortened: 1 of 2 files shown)


tinswzy:
Closing this — #50812 already takes this approach, it just wasn't cherry-picked to 2.6.

While preparing this fix I found **#50812** (`enhance: isolate online-write cgo pool from segment loading`, @xiaofan-luan) already isolates the tsafe-critical online-write (`Insert`/`Delete`) CGO applies onto a dedicated pool so segment-loading / delta-replay can't starve them and freeze tsafe — the same idea as this PR. It was **merged to master (2026-06-30)** but was **not cherry-picked to 2.6**.

| | #50812 (master) | this PR (2.6) |
|---|---|---|
| Insert/Delete apply pool | dedicated `mutatePool` | `deletePool` |
| `LoadDeletedRecord` (delta replay) | → `LoadPool` | stays on `dynamicPool` |
| pool sizes | configurable (`DynamicPoolSizeFactor`) | fixed |
| scope | querynode + datanode + pool metrics | querynode only |

Same idea, and #50812 is the more complete version, so the cleaner action is to **cherry-pick #50812 to 2.6** rather than land a parallel implementation — closing this in its favor. Root-cause analysis + validation data stay on #49435.

One caveat for the backport: in our runs, some earlier master builds already carrying #50812 still reproduced #49435 under the load storm, so it's worth re-validating the 2.6 backport on the fouram case.

xiaocai2333:

## Overview

Verification surfaced one high-severity concurrency defect: this PR routes `LocalSegment.Delete`/`Insert` CGO calls onto `deletePool`, which is already the outer dispatch pool for `DeleteBatch`, creating a nested-submit deadlock on a bounded pool.

## High

**`internal/querynodev2/segments/segment.go:861` — nested submission into the same bounded pool can deadlock**

`LocalSegment.Delete` now runs its CGO call as `GetDeletePool().Submit(...).Await()` (changed from `GetDynamicPool()` in this PR), and `LocalSegment.Insert` at `segment.go:794` was moved onto the same pool. But `deletePool` is already the *outer* dispatch pool: `QueryNode.DeleteBatch` takes `pool := segments.GetDeletePool()` at `internal/querynodev2/services.go:1644` and submits one task per segment, each task synchronously calling `segment.Delete(...)` at line 1653. After this change, the outer task runs on a pool worker and then submits to — and awaits — that same pool. The pool is created as `conc.NewPool[struct{}](runtime.GOMAXPROCS(0))` with no `WithNonBlocking` (`pool.go:182-186`), and `Pool.Submit` explicitly "will block if the pool has finite workers and no idle worker" (`pkg/util/conc/pool.go:63-66`, `nonBlocking: false` in `options.go:72-78`).

Failure scenario: with `queryNode.streamingDeltaForwardPolicy=Direct`, a delegator `ProcessDelete` forwards a delete batch to local growing segments (or via gRPC to a remote node's sealed segments) where the batch's segment count ≥ `GOMAXPROCS(0)` — deterministic with a single segment in a 1-CPU container → every `deletePool` worker blocks in the inner `Submit`'s ants `cond.Wait` waiting on inner tasks that can never be scheduled; the `DeleteBatch` RPC hangs forever, `ProcessDelete` stalls, `deleteNode` stops calling `UpdateTSafe`, and channel tsafe freezes — which is exactly the path this PR aims to unblock. Because `LocalSegment.Insert` now shares that pool, the insert flowgraph stalls too, so the whole QueryNode write/delete path stops and Strong-consistency queries return `code=505 channel tsafe stalled` until the node is restarted. Note the default policy is `FilterByBF` (`component_param.go:4678`), which routes through `node.Delete` (`services.go:1586`) without the outer pool, so default deployments are not affected.

Suggestion: keep pool ownership at exactly one layer. Either give the `segment.Delete`/`Insert` CGO submissions an execution pool distinct from the outer dispatch pool — master already does this with separate `mutatePool` + `deletePool` and documents the constraint ("MUST stay separate ... nested-submit deadlock", `pool.go:62-66` @ `c95a29d97b`) — or drop the outer `deletePool` dispatch in `DeleteBatch`. Please also add a regression test that runs a delete batch with pool capacity fully occupied.

(raised by chyezh, xaxys)

PR #52455, opened 2026-08-12
fix: [cp2.6] batch backport 6 master fixes

bigsheeper:
Unified batch backport to 2.6 (one CP PR instead of six, to reduce CI load). One commit per master PR, cherry-picked in master merge order.

pr: #51607
pr: #51516
pr: #52046
pr: #51777
pr: #52015
pr: #52113
issue: #51606
issue: #51512
issue: #52045
issue: #51776
issue: #52014
issue: #52112

Supersedes closed per-PR CP PRs: #52418, #52419, #52424, #52423, #52411, #52410 (identical resolved content, verified line-by-line against each master PR diff).

## Included fixes (cherry-pick order)

1. #51607 — honor gc pause timeout. 2.6 adaptation: "pause rolled back" and pause-ticket logs use 2.6's zap-style logging (master uses mlog).
2. #51516 — recycle stale analyze stats versions and persist analyze terminal states. 2.6 adaptation: `indexpb` import uses `pkg/v2`; two persist-failure warns use zap; only this PR's own new GC test is ported (master-only snapshot/commit_ts tests dropped from the conflict hunk).
3. #52046 — read only the needed leaf columns for struct array sub-fields. 2.6 adaptation: new test imports normalized `pkg/v3`→`pkg/v2`, `go-api/v3`→`go-api/v2`.
4. #51777 — wait for segment release after lock timeout. 2.6 adaptation: watchdog log uses zap; configs/milvus.yaml conflict was positional only.
5. #52015 — convert ArrayOfVector element placeholders. 2.6 adaptation: test calls 2.6's 2-value `ConvertPlaceholderGroup` signature; two test helpers ported verbatim from master.
6. #52113 — resolve search timezone before plan generation. 2.6 adaptation: moved timezone block keeps zap-style logging; test uses 2.6's `newSchemaInfo`.
7. `66595ca` — make `typeutil.GetFieldByID` find struct array sub-fields. **2.6-only; no corresponding master PR** — master and 3.0 already walk `StructArrayFields`, so this aligns 2.6 with them rather than back-porting anything. It is the prerequisite for item 5: an ArrayOfVector field can only live inside a struct array field, so without this change `convertPlaceholderIfNeeded` (`internal/proxy/task_search.go:895`) resolves the anns field to `nil` and forwards the placeholder unconverted, leaving the #52015 back-port dead code on this branch. Blast radius is confined to that one call site — `GetFieldByID` has exactly one non-test caller on 2.6.

## Verification

- [x] Items 1-6: each commit's changed lines verified identical to the individually-reviewed CP commits (which were line-level compared against their master PR diffs)
- [x] Item 7 has no master PR diff to compare against; verified instead that the resulting `GetFieldByID` body matches master's current implementation exactly
- [x] Same-file pair #51607/#51516 stacked in master merge order; zero conflicts
- [x] No conflict markers

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -1024,7 +1024,7 @@ common:
     threshold:
       info: 500 # minimum milliseconds for printing durations in info level
       warn: 1000 # minimum milliseconds for printing durations in warn level
-    maxWLockConditionalWaitTime: 600 # maximum seconds for waiting wlock conditional
+    maxWLockConditionalWaitTime: 600 # seconds before logging a wlock conditional wait that is taking long; the wait itself is not bounded by this value
   storage:
     stv2:
       splitSystemColumn:
--- a/internal/datacoord/garbage_collector.go
+++ b/internal/datacoord/garbage_collector.go
@@ -88,10 +88,15 @@ type gcCmd struct {
 	collectionID int64
 	ticket       string
 	done         chan error
+	ctx          context.Context
 	timeout      <-chan struct{}
 }
 
 type gcPauseRecord struct {
+	// id uniquely identifies this record within its gcPauseRecords. Tickets are
+	// not unique -- the REST route in restful_mgr_routes.go issues every pause
+	// with an empty ticket -- so rollback must delete by id, not by ticket.
+	id         int64
 	ticket     string
 	pauseUntil time.Time
 }
@@ -99,6 +104,7 @@ type gcPauseRecord struct {
 type gcPauseRecords struct {
 	mut     sync.RWMutex
 	maxLen  int
+	nextID  int64
 	records typeutil.Heap[gcPauseRecord]
 }
 
@@ -117,17 +123,15 @@ func (gc *gcPauseRecords) PauseUntil() time.Time {
 	return gc.records.Peek().pauseUntil
 }
 
-func (gc *gcPauseRecords) Insert(ticket string, pauseUntil time.Time) error {
+// Insert records a pause ticket and returns the id of the record it created, so
+// that a failed pause can roll back exactly its own record via DeleteByID.
+func (gc *gcPauseRecords) Insert(ticket string, pauseUntil time.Time) (int64, error) {
 	gc.mut.Lock()
 	defer gc.mut.Unlock()
 
 	// heap small enough, short path
 	if gc.records.Len() < gc.maxLen {
-		gc.records.Push(gcPauseRecord{
-			ticket:     ticket,
-			pauseUntil: pauseUntil,
-		})
-		return nil
+		return gc.pushLocked(ticket, pauseUntil), nil
 	}
 
 	records := make([]gcPauseRecord, 0, gc.records.Len())
@@ -143,25 +147,46 @@ func (gc *gcPauseRecords) Insert(ticket string, pauseUntil time.Time) error {
 	})
 
 	if gc.records.Len() < gc.maxLen {
-		gc.records.Push(gcPauseRecord{
-			ticket:     ticket,
-			pauseUntil: pauseUntil,
-		})
-		return nil
+		return gc.pushLocked(ticket, pauseUntil), nil
 	}
 
 	// too many pause records, refresh heap
-	return merr.WrapErrTooManyRequests(64, "too many pause records")
+	return 0, merr.WrapErrTooManyRequests(64, "too many pause records")
 }
 
+// pushLocked appends a new record with a freshly allocated id. Caller must hold mut.
+func (gc *gcPauseRecords) pushLocked(ticket string, pauseUntil time.Time) int64 {
+	gc.nextID++
+	gc.records.Push(gcPauseRecord{
+		id:         gc.nextID,
+		ticket:     ticket,
+		pauseUntil: pauseUntil,
+	})
+	return gc.nextID
+}
+
+// Delete drops every record holding the given ticket. This is the user-facing
+// resume semantic ("release my ticket"); use DeleteByID to undo a single record.
 func (gc *gcPauseRecords) Delete(ticket string) {
+	gc.deleteMatching(func(r gcPauseRecord) bool { return r.ticket == ticket })
+}
+
+// DeleteByID drops the single record with the given id, leaving records that
+// merely share its ticket untouched.
+func (gc *gcPauseRecords) DeleteByID(id int64) {
+	gc.deleteMatching(func(r gcPauseRecord) bool { return r.id == id })
+}
+
+// deleteMatching rebuilds the heap without the matching records, dropping
+// already-expired records along the way.
+func (gc *gcPauseRecords) deleteMatching(match func(gcPauseRecord) bool) {
 	gc.mut.Lock()
 	defer gc.mut.Unlock()
 	now := time.Now()
 	records := make([]gcPauseRecord, 0, gc.records.Len())
 	for gc.records.Len() > 0 {
 		record := gc.records.Pop()
-		if now.Before(record.pauseUntil) && record.ticket != ticket {
+		if now.Before(record.pauseUntil) && !match(record) {
 			records = append(records, record)
 		}
 	}
@@ -292,11 +317,17 @@ func (gc *garbageCollector) Pause(ctx context.Context, collectionID int64, ticke
 		collectionID: collectionID,
 		ticket:       ticket,
 		done:         done,
+		ctx:          ctx,
 		timeout:      ctx.Done(),
 	}:
 		return <-done
 	case <-ctx.Done():
 		return ctx.Err()
+	case <-gc.ctx.Done():
+		// cmdCh is unbuffered and startControlLoop has already returned, so the
+		// send above can never be received; without this arm the caller parks
+		// until its own ctx is canceled (forever for a Done()-less ctx).
+		return merr.WrapErrServiceUnavailable("garbage collector is closing")
 	}
 }
 
@@ -318,6 +349,9 @@ func (gc *garbageCollector) Resume(ctx context.Context, collectionID int64, tick
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
+	case <-gc.ctx.Done():
+		// see Pause: the control loop is gone, nothing will receive this send.
+		return merr.WrapErrServiceUnavailable("garbage collector is closing")
 	}
 }
 
@@ -397,8 +431,9 @@ func (gc *garbageCollector) pause(cmd gcCmd) error {
 		zap.Duration("duration", cmd.duration),
 	)
 	var err error
+	var recordID int64
 	if cmd.collectionID <= 0 { // legacy pause all
-		err = gc.pauseUntil.Insert(cmd.ticket, reqPauseUntil)
+		recordID, err = gc.pauseUntil.Insert(cmd.ticket, reqPauseUntil)
 		log.Info("global pause ticket recorded")
 	} else {
 		curr, has := gc.pausedCollection.Get(cmd.collectionID)
@@ -406,7 +441,7 @@ func (gc *garbageCollector) pause(cmd gcCmd) error {
 			curr = NewGCPauseRecords()
 			gc.pausedCollection.Insert(cmd.collectionID, curr)
 		}
-		err = curr.Insert(cmd.ticket, reqPauseUntil)
+		recordID, err = curr.Insert(cmd.ticket, reqPauseUntil)
 		log.Info("collection new pause ticket recorded")
 	}
 	if err != nil {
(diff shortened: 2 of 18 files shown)


bigsheeper:
Force-pushed a CI fix.

**#52046 commit** — `leaf_pruning_test.go` failed to compile on 2.6:
```
leaf_pruning_test.go:467:45: vectorField.ValidData undefined
  (storage.VectorArrayFieldData has no field or method ValidData)
```
2.6's `storage.VectorArrayFieldData` carries only `Dim / ElementType / Data`; master added `ValidData []bool` and `Nullable bool` for nullable vector arrays, a feature this branch does not have. `TestStructArrayLeafPruningWithNulls` asserts exactly that validity bitmap, so it has been dropped from the backport rather than hollowed out. The leaf-pruning fix itself stays covered by `TestCollectSubFieldLeaves`, `TestCollectSubFieldLeavesRejectsBadIndex` and `TestStructArrayReadAmplification`, none of which touch nullability. Now-unused `arrow/array` and `parquet` imports removed with it.

The other five commits are untouched (verified line-by-line: zero changes).

czs007:

One backported fix (#52015) is unreachable on the 2.6 search path because it depends on a `typeutil` helper semantic that exists only on master, plus a minor test-coverage gap in the new leaf-pruning tests.

## High

**`internal/proxy/vector_type_convert.go:120` — the ArrayOfVector placeholder conversion is dead code on 2.6; the backported bug is not actually fixed.**

The new branch itself is correct — `fieldType` falls back to `GetElementType()` when the field is `DataType_ArrayOfVector` — but it can never be reached in production on this branch. `ConvertPlaceholderGroup` has exactly one non-test caller, `searchTask.convertPlaceholderIfNeeded` (`task_search.go:894-898`), which resolves the field with `typeutil.GetFieldByID`. On 2.6 that helper only iterates `schema.GetFields()` (`pkg/util/typeutil/schema.go:2627-2634`); on master it additionally walks `StructArrayFields` (`master:pkg/util/typeutil/schema.go:3311-3323`). Since proxy validation requires an `ArrayOfVector` field to live inside a struct-array field (`internal/proxy/util.go:647-649`), its `FieldID` is never in `schema.Fields`, so `GetFieldByID` returns `nil` and the placeholder is forwarded unchanged. The divergence is visible six lines away: `task_search.go:831` resolves the *same* `t.FieldId` with `typeutil.GetField` (which does descend into `StructArrayFields`) and successfully branches on `DataType_ArrayOfVector`. `t.schema.CollectionSchema` is the raw proto (`meta_cache.go:167-175`), so sub-fields are not flattened into `Fields` either. Upstream #52015 touched only `vector_type_convert.go` and its test, because on master no lookup change was needed.

Failure: a 2.6 collection with sub-field `struct_array[vector_array]` (`DataType_ArrayOfVector` / `ElementType_Float16Vector`) searched with `anns_field=struct_array[vector_array]` and an fp32 `PlaceholderType_FloatVector` group → `GetFieldByID` returns `nil`, the fp32 placeholder is passed through untouched with no rejection path in proxy (`task_search.go:831-862` only guards range/iterator/group-by), and segcore rejects it as a vector type mismatch exactly as in #52014 — behaviour is identical before and after this PR, while the release note claims the fix shipped. Hybrid search goes through the same call site (`task_search.go:575`) and is equally affected.

Suggestion: either switch `task_search.go:895` to `typeutil.GetField` (already the convention at `:459` and `:831`), or bring 2.6's `GetFieldByID` in line with master's struct-array-aware version. Please also add a case that enters through `convertPlaceholderIfNeeded` with a struct-array sub-field ID — the added `TestConvertPlaceholderGroupArrayOfVectorElementType` (`vector_type_convert_test.go:278`) hand-builds a `FieldSchema` and calls `ConvertPlaceholderGroup` directly, bypassing the lookup, which is why it passes against unchanged end-to-end behaviour. (raised by tinswzy)

## Low

**`internal/util/importutilv2/parquet/leaf_pruning_test.go:120` — the new leaf-pruning tests assert no decoded values, and `nullPercent` is a dead parameter.**

`writeLeafPruningParquet(..., nullPercent int)` is passed `0` at all three call sites (`:155`, `:203`, `:275`), and the helper's own comment (`:116-119`) notes that a schema from `leafPruningTestSchema` has no nullable fields, so that path is never exercised. The gap is worth closing here specifically because pruning changes which leaf supplies the def/rep levels used to rebuild the outer list: unpruned, `newStructReader` picks the first non-repeated child (`int_array`) as `defRepLevelChild` (arrow v17 `pqarrow/column_readers.go:191-201`), but after `getReader` drops nil child readers (`pqarrow/file_reader.go:522-531`) the struct has only the list child left, so `hasRepeatedChild = true` and offsets/validity are derived from a deeper leaf. That behaviour is arrow-supported (`file/level_conversion.go:199-201`), so this is not a defect claim — but the only end-to-end test over the new reader, `TestStructArrayReadAmplification` (`:269-307`), asserts only `totalRows == 1000` and `readBytes < 2*fileSize`, and the pre-existing `reader_test.go:713-725` compares only `RowNum()`.

Suggestion: add value-equality assertions against `insertData` for sub-fields 201/202/203, and mark at least one case `Nullable=true` so `nullPercent` is actually exercised (restricting nullability to `int_array`/`float_array` if 2.6's `storage.VectorArrayFieldData` still lacks `ValidData`). (raised by tinswzy)

czs007:

This PR's included-fixes list does not match the branch's actual commit set; one documentation item follows, with no blocking findings recorded.

## Low — non-blocking

- **`PR description:1` — the "Included fixes (cherry-pick order)" list, and the `pr:`/`issue:` lines, cover only 6 of the branch's 7 commits.** Commit `66595ca` (`fix: [cp2.6]make GetFieldByID find struct array sub-fields`) is absent from the description, yet it is not an ordinary cherry-pick: it widens `typeutil.GetFieldByID` (`pkg/util/typeutil/schema.go`) from top-level fields only to top-level **plus** struct array sub-fields, which is the precondition that makes item 5 (#52015, ArrayOfVector placeholder conversion) actually reachable on 2.6 — without it, `convertPlaceholderIfNeeded` (`internal/proxy/task_search.go:895`) always resolves to `nil` and forwards the placeholder unconverted, making the #52015 backport dead code. It also changes a shared helper in `pkg/util/typeutil`, so its blast radius is worth stating explicitly (in 2.6 there is exactly one non-test caller, `task_search.go:895`).

  Suggested edit: add a 7th entry to "Included fixes" marked as **2.6-only, no corresponding master PR** — master's `GetFieldByID` already walks `StructArrayFields`, so this commit aligns 2.6 with master/3.0 — and note that it is the prerequisite for item 5 being reachable on 2.6. Please do not attach a fabricated `pr:` reference for it. Additionally, adjust the first Verification checkbox: "Each commit's changed lines verified identical to the individually-reviewed CP commits (which were line-level compared against their master PR diffs)" cannot hold for this commit, since it has no master PR diff to compare against. (surfaced during verification)

bigsheeper:
Confirmed and fixed in the PR description. I verified each premise against the branch itself rather than judging from the description, and everything you wrote holds.

**The omission is real.** `git log 2.6..HEAD` returns 7 commits; the "Included fixes" list had 6. The missing one is `66595ca`.

**It is not an ordinary cherry-pick, and your reason is right.** Before it, 2.6's `typeutil.GetFieldByID` (`pkg/util/typeutil/schema.go:2627`) iterated `schema.GetFields()` only. After it, the body is byte-identical to master's current implementation (`pkg/util/typeutil/schema.go:3368`), which walks `StructArrayFields` as well — so this aligns 2.6 with master/3.0 rather than back-porting a master change, and there is no `pr:` reference to attach. I have not fabricated one.

**Its role as item 5's precondition checks out.** `convertPlaceholderIfNeeded` (`internal/proxy/task_search.go:894-900`) is the sole non-test caller of `GetFieldByID` on 2.6 — I grepped the whole branch to confirm "exactly one". An ArrayOfVector field can only live inside a struct array field, so on the old lookup it resolved to `nil`, hit the `if field == nil { return phgBytes, nil }` early return, and forwarded the placeholder unconverted. The #52015 back-port would have been dead code on this branch and the release note would have claimed a fix that never ran.

**Description changes made:**
- Added item 7, marked **2.6-only; no corresponding master PR**, stating that it aligns 2.6 with master/3.0, that it is the prerequisite for item 5, and that its blast radius is the single call site.
- Split the first Verification checkbox: items 1-6 keep the "identical to the individually-reviewed CP commits, line-level compared against their master PR diffs" claim; item 7 gets its own line saying it has no master PR diff to compare against and was instead verified to match master's current `GetFieldByID` body exactly. You were right that the original wording could not hold for it.

No code changed, so the in-flight CI is undisturbed.

**Separately, on this PR's red checks — neither is caused by this PR, and neither is fixable here:**
- `ci-v2/go-sdk`: `TestCreateCollectionInvalid` and `TestCreateCollectionInvalidAutoPkField` expect `primary key is not specified` but get `data type None is not valid`. Root cause is on the 2.6 branch itself: `6cb886dec` (#52485) moved `validateFieldType(t.schema)` in `createCollectionTask.PreExecute` to *before* the primary-key check, and `tests/go_client/testcases/collection_test.go:997-998` was not updated. It reproduces identically on rerun and affects every 2.6 PR whose go-sdk stage ran after 2026-08-13.
- `Code Checker MacOS`: `conan` fails to fetch `https://dlcdn.apache.org/avro/avro-1.12.1/avro-src-1.12.1.tar.gz` (404 — Apache's CDN drops superseded releases), ending in `make: *** [build-3rdparty] Error 1`. Infrastructure, not rerunnable.

czs007:

## Re-review of `66595caaa3e0`

- **Retracted** — [low] PR description, line 1: PR description omits the 7th commit (`GetFieldByID`). Both reviewers re-checked the current PR body: the "Included fixes" list now contains a 7th item fully describing commit `66595ca` (noting it is 2.6-only with no corresponding master PR, that it is a prerequisite of item 5, and its single call site at `internal/proxy/task_search.go:895`), the verification checklist covers it, and the author confirmed the description fix. The finding concerned only the description text and no longer applies.

Verified:
- `pkg/util/typeutil/schema.go:2628` — the widened `GetFieldByID` semantics (traversing `StructArrayFields`) affect only one non-test call site, `internal/proxy/task_search.go:895`, which is exactly the place that needs the new behavior; no other caller depends on sub-fields resolving to nil.
- `internal/datacoord/garbage_collector.go:485` — pause records receive unique IDs, so a timed-out request rolls back only its own record even when tickets are shared, and an Insert failure returns before the rollback path can run with a zero record ID.
- `internal/datacoord/garbage_collector.go:1434` — analyze GC now deletes stale `{taskID}/{version}` prefixes matching the actual storage layout, retaining the current version, instead of the old prefix form that could remove other tasks' directories.
- `internal/querynodev2/segments/state/load_state_lock.go:936` — full release now waits until the ref count reaches zero; the configured timeout is only a watchdog log threshold and no longer allows premature native resource cleanup, with regression coverage for the post-timeout wait.
- `internal/proxy/task_search.go:234` — timezone resolution was moved ahead of both normal and advanced plan generation, after `collectionInfo` is already available, so invalid timezones fail early and timestamptz literals parse with the resolved timezone.
- `internal/util/importutilv2/parquet/struct_field_reader.go:1444` — leaf collection terminates on `len(Children)==0` (avoiding the group-node `ColIndex=0` pitfall, which has a dedicated test), and the pruned single-child struct is consistently addressed at child index zero with an explicit error fallback.

No new findings survived verification this round, and there are no disputed or pre-existing items to report.

czs007:

**Verified:**
- `pkg/util/typeutil/schema.go:2628-2643` — `GetFieldByID` now walks `StructArrayFields` after `schema.Fields`, byte-identical to master's implementation, so a struct-array sub-field ID resolves instead of returning `nil`.
- `internal/proxy/task_search.go:895` — the anns sub-field ID resolved by `GetFieldByName` flows into `convertPlaceholderIfNeeded` and now yields a non-nil schema with `DataType_ArrayOfVector`, so the fp32→fp16/bf16 conversion branch is actually reachable on this branch rather than dead code.
- `internal/proxy/vector_type_convert_test.go:459-510` — the regression test enters through `task.convertPlaceholderIfNeeded` with a sub-field ID (not through `ConvertPlaceholderGroup` directly), so it fails again if the lookup regresses.
- `internal/proxy/vector_type_convert.go:120` — blast radius of widening the lookup: `PlaceholderType_EmbList*` is absent from `placeholderTypeToDataType`, so embedding-list placeholders hit the pass-through guard before the fp16/bf16 type check; no regression from making the branch live.
- `internal/util/importutilv2/parquet/leaf_pruning_test.go:153-197` — `TestCollectSubFieldLeaves` asserts each sub-field resolves to a leaf whose parquet path contains that sub-field's name and that no leaf is claimed twice, which is the assertion that catches the `IsLeaf()`/`ColIndex == 0` mis-mapping trap.
- `internal/util/importutilv2/parquet/struct_field_reader.go:131` — the hard-coded `fieldIndex: 0` is consistent with the pruning it now performs: `GetFieldReader` given a single sub-field's leaf set yields a struct with exactly one surviving child.

## Low — non-blocking (no change required in this PR)

- **`internal/util/importutilv2/parquet/leaf_pruning_test.go:153`** — The port of `leaf_pruning_test.go` from master #52046 brings over three of the four new tests, dropping `TestStructArrayLeafPruningWithNulls`. That was the only test asserting the *content* read back through the pruned reader (`assert.Equal(want, vectorRows[rowIdx].GetFloatVector().GetData(), ...)`), i.e. that pruning kept the right leaf column rather than merely a leaf of the right physical shape — which is exactly what the hard-coded `fieldIndex: 0` at `struct_field_reader.go:126-133` relies on. Its `ValidData` half is genuinely inapplicable here, since 2.6's `storage.VectorArrayFieldData` (`internal/storage/insert_data.go:579-583`) has no `ValidData` field, but the vector-content half is applicable and nothing else on this branch covers it: `TestCollectSubFieldLeaves` asserts leaf *paths* only, `TestStructArrayReadAmplification` asserts only row count and bytes read, the pre-existing struct-array end-to-end case's "Verify data content matches" loop compares `RowNum()` only (`reader_test.go:720-725`), and the value-level cases in `field_reader_test.go` (e.g. `TestStructFieldReader_ArrayOfVectorFloat64`, `field_reader_test.go:1167-1218`) hand-build a single-child chunked array and call `readArrayOfVectorField` directly, never going through `NewStructFieldReader`/`GetFieldReader`. The gap matters in this schema specifically because `struct_array[float_array]` and `struct_array[vector_array]` share the same physical arrow type (`list<float32>`), so a mis-selected leaf would pass any type check and only a content assertion would catch it. This is a coverage gap rather than a defect: although the `go.mod` replace pins differ (2.6 pins `milvus-io/arrow/go/v17 v17.0.0`, master `v17.0.2`), the entire `parquet/` tree including `GetFieldReader`/`getReader` is byte-identical between those two fork versions, so sibling pruning behaves the same and the `fieldIndex: 0` assumption is sound. **Suggestion:** as follow-up, restore master's content assertion in a test that omits the nullable rows (since 2.6 lacks `ValidData` on `VectorArrayFieldData`). *(raised by tinswzy)*

czs007 (approved):
/lgtm

PR #52458, opened 2026-08-12
feat: publish StorageV3 index artifacts through manifests

yuruiz:
## Summary

- publish completed StorageV3 index artifacts into the source manifest through the typed `LoonIndexInfo` transaction API from milvus-storage#620
- persist the exact committed manifest path before completing the DataCoord index task, with base-path and single-revision validation
- retain etcd `SegmentIndex` file metadata as the normal handoff source; lazily fall back to the manifest only when `index_file_keys` are absent, reading it once per segment
- pin milvus-storage to merged commit `dac5781e5cd298e07d2b5822ba8a2b879e99cd45`

## Design

Design doc: [StorageV3 Manifest Index Publication](https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260811-storagev3-manifest-index-publication.md)

## Verification

Run in the Milvus development container:

- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/storagev2/packed -run TestManifestIndexRoundtrip`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datanode/index -run TestIndexBuildTaskSuite/TestStorageV3PostExecutePublishesManifestIndex`
- `go test -tags dynamic,test -gcflags="all=-N -l" -count=1 ./internal/datacoord -run "TestManifestIndexFallback|TestManifestIndexFileKeysRejectPathTraversal|TestServerGetIndexInfosManifestFallback"`

Relates to #51723.


internal/storagev2/packed/transaction.go
```diff
@@ -25,15 +25,208 @@ import "C"
...
+	defer C.loon_properties_free(cProperties)
+
+	indexManifestLocks.Lock(basePath)
+	defer indexManifestLocks.Unlock(basePath)
+
+	// A previous index task may have committed a newer revision from the same
+	// source manifest while this task was waiting for the per-segment lock.
+	// Rebase this metadata-only update on that revision before opening the
+	// fail-resolved transaction.
+	version, err = getLatestManifestVersion(basePath, cProperties)
```
yhmo: internal/storagev2/packed/transaction.go line:95
High ---- This replaces the task's source revision with whatever revision is latest before opening the fail-resolved transaction. If a compaction or stats writer advances the same manifest after the build read revision R, this code rebases the index built from R onto that newer data and commits R+2; DataCoord then accepts it because the result is exactly current+1. That violates the source-manifest validity invariant described in the design. Please only rebase over revisions known to be index-only publications from the same source, or otherwise reject/rebuild when the latest revision is not the task's source revision.

internal/datacoord/task_index.go
```diff
@@ -130,7 +130,11 @@ func (it *indexBuildTask) UpdateTaskVersion(nodeID int64) error {
 }
 
 func (it *indexBuildTask) setJobInfo(result *workerpb.IndexTaskInfo) error {
-	if err := it.meta.indexMeta.FinishTask(result); err != nil {
+	if result.GetState() == commonpb.IndexState_Finished && result.GetManifestPath() != "" {
+		if err := it.meta.FinishIndexTaskWithManifest(it.meta.ctx, result, it.SegmentID, result.GetManifestPath()); err != nil {
```
yhmo, line 134: internal/datacoord/task_index.go line:134
High ---- `FinishIndexTaskWithManifest` returns `errIndexManifestPublicationStale` specifically so the task can be rebuilt, but `QueryTaskOnWorker` calls `it.setJobInfo(info)` without checking this error. On a base/revision mismatch the local task remains InProgress while the worker keeps returning Finished, so each poll repeats the rejected publish indefinitely instead of dropping and resetting the task. Please handle this error at the polling call site and trigger the intended worker drop/reset path.

yhmo: The issue is confirmed fixed. Resolving this conversation.

Issue #52469, opened 2026-08-12
[Bug]: advanced search shares mutable sealed snapshot across sub-search pruning


bigsheeper:
## Summary

Advanced search fans out sub-searches concurrently but passes the same pinned `[]SnapshotItem` sealed segment snapshot to every sub-search. When `queryNode.enableSegmentPrune=true`, each sub-search calls `PruneSegments`, which mutates that `[]SnapshotItem` in place. One sub-search can therefore prune the sealed segment list seen by a sibling sub-search.

## Impact

This is a High correctness/concurrency issue. Hybrid/advanced search results can become unstable or wrong for clustering-key collections when segment prune is enabled, because one sub-request's scalar/vector predicate can remove sealed segments before another sub-request organizes its own search tasks.

## Call chain

- Proxy advanced search builds `SearchRequest.IsAdvanced=true` and `SubReqs` in `internal/proxy/task_search.go`.
- QueryNode delegator pins readable segments once in `internal/querynodev2/delegator/delegator.go`.
- The advanced-search branch starts one goroutine per sub-request.
- Each goroutine calls `sd.search(...)` with the same `sealed` slice.
- `sd.search` calls `PruneSegments` when `queryNode.enableSegmentPrune=true`.
- `PruneSegments` rewrites `sealedSegments[idx]` in place.

## Executable proof

A minimal repro at the `PruneSegments` boundary models the old fan-out behavior by sharing one `[]SnapshotItem` view between sibling sub-searches. Pruning one sibling changes the other sibling's view:

```text
$ GOWORK=off PKG_CONFIG_PATH=<local core pkgconfig> LD_LIBRARY_PATH=<local core lib> go test -tags dynamic,test -gcflags="all=-N -l" ./internal/querynodev2/delegator -run TestSegmentPrunerSuite/TestSharedSnapshotItemsContaminateSiblingPruneRepro -count=1
--- FAIL: TestSegmentPrunerSuite (0.00s)
    --- FAIL: TestSegmentPrunerSuite/TestSharedSnapshotItemsContaminateSiblingPruneRepro (0.00s)
        segment_pruner_test.go:714:
             Error:       Not equal:
                          expected: 2
                          actual  : 0
FAIL
FAIL    github.com/milvus-io/milvus/internal/querynodev2/delegator    0.335s
```

## Expected behavior

Each advanced-search sub-request should prune an isolated sealed segment snapshot so its pruning cannot affect any sibling sub-request.

PR #52470, opened 2026-08-12
fix: isolate advanced search pruning snapshots

bigsheeper:
issue: #52469

## What changed

Advanced search fans out sub-searches concurrently, but segment pruning mutates the provided sealed `[]SnapshotItem` in place. This PR passes each sub-search its own cloned snapshot view so one sub-request cannot prune the segment set used by another sibling request.

## Verification

- Proof-before-fix repro failed as expected: `TestSegmentPrunerSuite/TestSharedSnapshotItemsContaminateSiblingPruneRepro` showed sibling contamination (`expected: 2`, `actual: 0`).
- `GOWORK=off PKG_CONFIG_PATH=<local core pkgconfig> LD_LIBRARY_PATH=<local core lib> go test -tags dynamic,test -gcflags="all=-N -l" ./internal/querynodev2/delegator -run TestSegmentPrunerSuite/TestCloneSnapshotItemsIsolatesInPlacePruning -count=1`
- `GOWORK=off PKG_CONFIG_PATH=<local core pkgconfig> LD_LIBRARY_PATH=<local core lib> go test -tags dynamic,test -gcflags="all=-N -l" ./internal/querynodev2/delegator -count=1`
- `GOWORK=off PKG_CONFIG_PATH=<local core pkgconfig> LD_LIBRARY_PATH=<local core lib> make static-check`

### DIFF
--- a/internal/querynodev2/delegator/delegator.go
+++ b/internal/querynodev2/delegator/delegator.go
@@ -545,6 +545,15 @@ func (sd *shardDelegator) search(ctx context.Context, req *querypb.SearchRequest
 	return sd.executeSearchSubTasks(ctx, req, sealed, growing, sealedRowCount)
 }
 
+func cloneSnapshotItems(sealed []SnapshotItem) []SnapshotItem {
+	if len(sealed) == 0 {
+		return nil
+	}
+	cloned := make([]SnapshotItem, len(sealed))
+	copy(cloned, sealed)
+	return cloned
+}
+
 // getVectorFieldDim returns the dimension of the vector field with the given field ID.
 // Returns 0 if the field is not found or dim cannot be determined.
 func (sd *shardDelegator) getVectorFieldDim(fieldID int64) int64 {
@@ -668,7 +677,7 @@ func (sd *shardDelegator) Search(ctx context.Context, req *querypb.SearchRequest
 					searchReq.GetReq().MvccTimestamp = tSafe
 				}
 				searchReq.Req.CollectionTtlTimestamps = req.GetReq().GetCollectionTtlTimestamps()
-				results, err := sd.search(ctx, searchReq, sealed, growing, sealedRowCount)
+				results, err := sd.search(ctx, searchReq, cloneSnapshotItems(sealed), growing, sealedRowCount)
 				if err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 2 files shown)


czs007:

Adversarial review of this change found the fix itself correct and well-scoped, with one non-blocking test-coverage gap.

## Low (non-blocking)

**`internal/querynodev2/delegator/segment_pruner_test.go:662`** — The new `TestCloneSnapshotItemsIsolatesInPlacePruning` exercises the `cloneSnapshotItems` helper directly (`pruned := cloneSnapshotItems(base)` followed by `PruneSegments(...)`) and never goes through `shardDelegator.Search`, so it places no constraint on the actual fix site at `delegator.go:680`. A grep of the delegator package shows no test that sets `IsAdvanced` or enables `EnableSegmentPrune` on a `Search` path, and the `TestSharedSnapshotItemsContaminateSiblingPruneRepro` repro cited in the PR description is not part of the diff — so if a later refactor reverts `cloneSnapshotItems(sealed)` back to `sealed`, this test still passes and the sibling-contamination bug returns silently. This matters because `PruneSegments` writes in place (`segment_pruner.go:158-159`, and likewise at `395-396`), so a shared backing array genuinely reintroduces cross-sub-request contamination.

*Suggestion:* land the repro as a permanent test — a `Search` with `IsAdvanced=true`, two sub-requests carrying disjoint clustering-key predicates, and `EnableSegmentPrune=true`, asserting that each sub-search's `organizeSubTask` sees the full sealed set. That pins the behavior at the call site rather than at the helper. *(raised by tinswzy)*

bigsheeper:
Recording the trigger conditions, since the description doesn't state them and the exposure turns out to be quite narrow.

Of the two in-place pruners in `sd.search`, `PruneSealedSegmentsByPKFilter` is a **no-op for advanced sub-requests**: the delegator hardcodes `PkFilter: common.PkFilterNoPkFilter` when building each sub-request (`delegator.go:665`), and the pruner returns immediately on that value (`segment_pruner.go:352`). So the pruning path that is on by default (`queryNode.enableSegmentFilter=true`) never mutates anything here.

That leaves `PruneSegments` as the only path that can actually mutate the shared snapshot, and it requires all three of:

- `queryNode.enableSegmentPrune=true` — **default is `false`**
- the collection has a clustering key — otherwise `segment_pruner.go:51` returns early
- partition stats are present — otherwise `segment_pruner.go:46` returns early

So default deployments are unaffected. Only hybrid search on a clustering-key collection with segment prune explicitly enabled can hit sibling contamination (silently reduced recall, plus an unsynchronized concurrent write to the shared `[]SnapshotItem`).

The fix itself is correct and cheap — a shallow clone is sufficient, since both mutation sites replace `item.Segments` with a freshly allocated slice rather than writing through it, and `PinReadableSegments` already hands out a per-call snapshot so nothing leaks back into `d.dist`.

Given that: fine to merge, equally fine to skip. No strong opinion either way — flagging it so the decision is made with the actual blast radius in view rather than assuming it's a default-path correctness bug.

PR #43576, opened 2025-07-26
fix: skip loading non-existent L0 segments to prevent load blocking

weiliu1031:
issue: #43577
In 2.5 branch, L0 segments must be loaded before other segments. If an L0 segment has been garbage collected but is still in the target list, the load operation would keep failing, preventing other segments from being loaded.

This patch adds a segment existence check for L0 segments in getSealedSegmentDiff. Only L0 segments that actually exist will be included in the load list.

Changes:
- Add checkSegmentExist function parameter to SegmentChecker constructor
- Filter L0 segments by existence check in getSealedSegmentDiff
- Add unit tests using mockey to verify the fix behavior

PR #52473, opened 2026-08-12
fix: [cp2.5.13] skip loading non-existent L0 segments to prevent load blocking (#43576)

czs007:
issue: #43557
pr: #43576

Cherry-pick of dffe2e42b3 / 75463725b3 from the 2.5 branch.

In the 2.5 line the segment checker loads L0 segments before every other
segment: when any L0 segment is pending, getSealedSegmentDiff replaces the
whole toLoad list with just the L0 ones. If an L0 segment has been garbage
collected but is still referenced by the current target - which survives a
QueryCoord restart, because the current target is persisted on graceful
shutdown and restored on boot - that load can never succeed, and it starves
every other segment on the collection. The delegator never reaches data-ready,
the target is never promoted, and the stale target is never replaced.

Reproduced deterministically on this branch and verified with this patch
applied (same data, same persisted target, same deleted L0 segment):

                                  without patch   with patch
  create segment task, level=L0        375            8
  create segment task, non-L0            0            8
  observer promotes current target       0            8
  collection state              Loading 20%       Loaded
  search                    channel not available   OK

Two deviations from the upstream commit, neither touching production code:
- The gofmt-only hunk in internal/datacoord/task_queue.go is dropped; it is
  unrelated to the fix and conflicts on this branch.
- The upstream test uses github.com/bytedance/mockey, which is not in this
  module graph. Rather than pull a mocking library and its transitive module
  changes into a hotfix branch, the existing test is adapted to the new
  constructor signature with a predicate that always reports the segment as
  existing.

### DIFF
--- a/internal/querycoordv2/checkers/controller.go
+++ b/internal/querycoordv2/checkers/controller.go
@@ -31,6 +31,7 @@ import (
 	"github.com/milvus-io/milvus/internal/querycoordv2/task"
 	"github.com/milvus-io/milvus/internal/querycoordv2/utils"
 	"github.com/milvus-io/milvus/pkg/v2/log"
+	"github.com/milvus-io/milvus/pkg/v2/util/merr"
 )
 
 var errTypeNotFound = errors.New("checker type not found")
@@ -62,11 +63,21 @@ func NewCheckerController(
 	broker meta.Broker,
 	getBalancerFunc GetBalancerFunc,
 ) *CheckerController {
+	checkSegmentExist := func(ctx context.Context, collectionID int64, segmentID int64) bool {
+		resp, err := broker.GetSegmentInfo(ctx, segmentID)
+		if err := merr.CheckRPCCall(resp, err); err != nil {
+			log.Info("check segment exist failed", zap.Int64("collectionID", collectionID), zap.Int64("segmentID", segmentID), zap.Error(err))
+			if errors.Is(err, merr.ErrSegmentNotFound) {
+				return false
+			}
+		}
+		return true
+	}
 	// CheckerController runs checkers with the order,
 	// the former checker has higher priority
 	checkers := map[utils.CheckerType]Checker{
 		utils.ChannelChecker: NewChannelChecker(meta, dist, targetMgr, nodeMgr, getBalancerFunc),
-		utils.SegmentChecker: NewSegmentChecker(meta, dist, targetMgr, nodeMgr, getBalancerFunc),
+		utils.SegmentChecker: NewSegmentChecker(meta, dist, targetMgr, nodeMgr, getBalancerFunc, checkSegmentExist),
 		utils.BalanceChecker: NewBalanceChecker(meta, targetMgr, nodeMgr, scheduler, getBalancerFunc),
 		utils.IndexChecker:   NewIndexChecker(meta, dist, broker, nodeMgr, targetMgr),
 		// todo temporary work around must fix
--- a/internal/querycoordv2/checkers/segment_checker.go
+++ b/internal/querycoordv2/checkers/segment_checker.go
@@ -41,13 +41,16 @@ import (
 
 const initialTargetVersion = int64(0)
 
+type CheckSegmentExist func(ctx context.Context, collectionID int64, segmentID int64) bool
+
 type SegmentChecker struct {
 	*checkerActivation
-	meta            *meta.Meta
-	dist            *meta.DistributionManager
-	targetMgr       meta.TargetManagerInterface
-	nodeMgr         *session.NodeManager
-	getBalancerFunc GetBalancerFunc
+	meta              *meta.Meta
+	dist              *meta.DistributionManager
+	targetMgr         meta.TargetManagerInterface
+	nodeMgr           *session.NodeManager
+	getBalancerFunc   GetBalancerFunc
+	checkSegmentExist CheckSegmentExist
 }
 
 func NewSegmentChecker(
@@ -56,6 +59,7 @@ func NewSegmentChecker(
 	targetMgr meta.TargetManagerInterface,
 	nodeMgr *session.NodeManager,
 	getBalancerFunc GetBalancerFunc,
+	checkSegmentExist CheckSegmentExist,
 ) *SegmentChecker {
 	return &SegmentChecker{
 		checkerActivation: newCheckerActivation(),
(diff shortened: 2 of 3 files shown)


xaxys:

Four issues were found in this patch: two medium-severity correctness/efficiency problems introduced by the new `checkSegmentExist` predicate, plus a redundant-task/log-noise issue and a coverage gap on the one branch this patch actually changes.

## Medium

**`internal/querycoordv2/checkers/segment_checker.go:432` — the patch breaks the "homogeneous slice" precondition of `createSegmentLoadTasks`**

`createSegmentLoadTasks` infers the level of the *entire* slice from `segments[0].GetLevel()` (`:432`) and, when `isLevel0` is true, unconditionally narrows the candidate nodes to `[]int64{leader.ID}` (`:451-453`). That decision is global: it applies to *every* shard group produced by the `GroupBy` at `:433`, not just the shard the ghost L0 belongs to. Before this patch the inference was sound — `level0Segments` was a pure level filter, so `toLoad` was either all-L0 or all-non-L0 (`:330-331`). Changing the filter to `level == L0 && c.checkSegmentExist(...)` at `:325` means a ghost L0 judged non-existent is dropped from `level0Segments` but *stays in* `toLoad`, producing the first mixed slice.

Failure: QueryCoord restarts and recovers a stale current target containing a GC'd ghost L0 while the same replica's `toLoad` also holds real sealed segments, and one round's Go-map iteration order (`:288-308`) puts the ghost L0 at `toLoad[0]` → the whole batch is misread as L0, every shard's normal segments are pinned to their delegator node, and `AssignSegment(..., forceAssign=true)` at `:460` bypasses both unhealthy-node filtering and `BalanceSegmentBatchSize` (`score_based_balancer.go:59-70` sets `balanceBatchSize = math.MaxInt64` under `forceAssign`), dumping the batch onto the delegator node — severe memory/load skew, up to OOM, re-rolled every 3s (`queryCoord.checkSegmentInterval` default 3000ms) and therefore not reproducible.

The opposite ordering (`segments[0]` non-L0, ghost L0 assigned off-leader) is covered by the forced redirect at `querynodev2/services.go:464-469` and only costs a task timeout. This code matches upstream `75463725b3` and is not a cherry-pick deviation, but it is a consequence of this patch; whether to carry it into the 2.5.13 hotfix is a release call.

Suggestion: remove confirmed-missing L0 segments from `toLoad` *before* deriving `level0Segments`, or move the `isLevel0` determination inside the per-shard loop at `:438` so it is decided per group. *(raised by bigsheeper)*

**`internal/querycoordv2/checkers/controller.go:66` — per-segment, un-batched synchronous DataCoord RPC on the serial checker hot path, with no per-round timeout**

The new `checkSegmentExist` closure (`controller.go:66-75`) passes a single segment ID to `broker.GetSegmentInfo`, even though `CoordinatorBroker.GetSegmentInfo` already batches up to 1000 IDs per call (`coordinator_broker.go:293-300`). It is invoked from inside the `lo.Filter` predicate at `segment_checker.go:325`, so every pending L0 in `toLoad` costs one synchronous RPC. The call stack is a single serial goroutine — `controller.startChecker`'s ticker branch calls `controller.check` synchronously (`controller.go:143/172-174`), which walks all collections × all replicas in order (`segment_checker.go:93-103`) — and the context derives from `context.Background()` (`controller.go:106`), so the round has no deadline; only each individual RPC is bounded by `QueryCoordCfg.BrokerTimeout` (default 5000ms). Before this patch `SegmentChecker` took no broker at all, so this is a newly introduced coupling.

Failure: DataCoord is unhealthy or slow (rolling upgrade, mixcoord restart, etcd jitter) with K pending L0 segments cluster-wide → a single `SegmentChecker.Check` round blocks for roughly K × 5s with no round-level deadline, and no segment load/release tasks are emitted for *any* collection during that window, stalling recovery and balancing entirely; even with a healthy DataCoord it amplifies into K serial un-batched RPCs repeated every 3s.

Suggestion: collect the round's candidate L0 IDs and issue one batched `GetSegmentInfo`, or cache the existence result with a short TTL. *(surfaced during verification)*

## Low

**`internal/querycoordv2/checkers/segment_checker.go:321` — stale L0 judged non-existent stays in `toLoad`, producing one guaranteed-to-fail load task per round plus unthrottled logging**

When `checkSegmentExist` returns false, the segment is only excluded from `level0Segments`; it remains in `toLoad` (`:321-326`), so each 3s round still emits a load task for it via `:136-141`. The `log.Info("check segment exist failed", ...)` at `controller.go:69` is unrated, so it prints once per ghost L0, per replica, per round.

Failure: current target holds a GC'd L0 after a QueryCoord restart and the next target has not yet been promoted → a doomed load task is scheduled every 3s, consuming a scheduler slot and one QueryNode `LoadSegments` round trip, alongside unthrottled Info logs until promotion.

This does not block the remaining segments (with `len(level0Segments) == 0` the wholesale replacement at `:330` is skipped and non-L0 segments still produce tasks) and does not affect target promotion (promotion only evaluates `NextTarget`; the ghost exists solely on the current-target path at `:299-308`), so the cost is wasted work and noise rather than a stall. Suggestion: same fix as the first item — drop confirmed-missing L0 segments from `toLoad` outright. *(raised by chyezh, bigsheeper)*

**`internal/querycoordv2/checkers/segment_checker_test.go:80` — no regression coverage for the predicate-returns-false path, which is the only branch this patch changes**

`SetupTest` injects a predicate that always returns true (`:80-85`), so all 13 subtests (`TestLoadSegments`, `TestLoadL0Segments`, `TestSkipLoadSegments`, …) exercise only the "segment exists" path, which is behaviorally identical to pre-patch code — the branch this PR introduces has zero coverage. The PR body's rationale (upstream uses `github.com/bytedance/mockey`, which is absent from this branch's module graph) checks out — there is no reference to it in `go.mod`, `go.sum`, or any `.go` file. But this patch deliberately made the existence check an injected function parameter (`segment_checker.go:44`, `:62`), so covering the false path needs no mocking library at all: construct another `SegmentChecker` with `func(...) bool { return false }`. Such a test would directly expose the first two items — assert what remains in `toLoad` after the ghost L0 is excluded, and whether `createSegmentLoadTasks` receives a homogeneous slice. Separately, `controller_test.go:85` makes no assertion at all about the inlined closure itself, including its `errors.Is(err, merr.ErrSegmentNotFound)` branch. *(raised by chyezh)*

Issue #45077, opened 2025-10-24
[Enhancement]: Make streamingnode memory always keep consistent with wal.


chyezh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Currently, we allow a wal is available when the wal append operation is failure.
However, once the append operation of wal is failed, the streamingnode cannot know the real wal entry order.
So the memory state of streamingnode is not consistent with wal forever, some feature like `writeaheadbuffer`, `idemptentChecker` can not work correctly. 
Here's the TODO:

- The Append operation of WAL can only be canceled when the wal is shutdown on non-graceful path.
- The append operation cannot be retried.
- Once the append operation is failed, the wal should be close and recover the memory state from the underlying wal again.

PR #52478, opened 2026-08-13
fix: enforce task timeout on channel, segment and leader operations

xiaofan-luan:
issue: #52257 related

`queryCoord.channelTaskTimeout` and `queryCoord.segmentTaskTimeout` were passed into `ChannelTask`/`SegmentTask` but never applied, and `LeaderTask` never took a timeout at all: every task context was created with `context.WithCancel`, so the configured timeouts were dead. A stuck `WatchDmChannels` / `UnsubDmChannel` / `LoadSegments` / `SyncDistribution` RPC (e.g. blocked on storage failures) could pin its scheduler slot and the channel/segment forever.

This PR binds the timeout to the task context, so it:

- bounds the task's action RPCs, starting from the moment it's actually dispatched to a QueryNode (see below — **not** from queue-add time);
- propagates through gRPC to the QueryNode and cancels the server-side operation;
- lets the task fail, free its scheduler slot, and be re-generated by the checker from the current distribution instead of hanging forever.

Leader tasks reuse `segmentTaskTimeout`. `channelTaskTimeout` default is unified to **2 minutes** (previously 60s, unenforced). `segmentTaskTimeout` defaults to **5 minutes** (previously 120s, unenforced) — kept longer than the channel timeout because segment loads (large/disk-index segments, S3 throttling) routinely take longer than channel watch/unsub ops. 5 minutes is judged enough headroom for a single segment's load under normal conditions; see the "Known residual gaps" section below for why this is still a flat budget rather than a true fix.

---

**Deadline starts at first dispatch, not at task construction** (added after initial review from bigsheeper/czs007): binding the deadline at `newBaseTask()` construction time meant a task sitting in the wait queue, or repeatedly bounced by the executor's admission cap (`GetChannelTaskCap`/`GetNonChannelTaskCap`), burned its RPC budget on queueing/admission delay rather than on the actual RPC. Under congestion — exactly when the checker's rebuilt-task-retry loop is most likely to spin — this could make a task that would otherwise succeed fail deterministically every round purely due to scheduler load, independent of whether the underlying operation would have completed within the configured timeout.

`Task.ActivateDeadline(step)` now arms a fresh timeout per step, called by `Executor.Execute()` only after that step clears both the wait queue and the admission cap — i.e. at its real dispatch. Queueing and admission-rejection time no longer count against the deadline; only time spent actually executing the action RPC does. A multi-action task (e.g. a move = `[grow, reduce]`) also no longer shares one deadline across both actions: each step gets its own full budget instead of a later step inheriting whatever was left of an earlier one (a slow `grow` could otherwise exhaust the shared budget before `reduce` was ever issued, leaving both the old and new copy live until a later checker round cleaned up the duplicate). `rootCtx`/per-step `ctx` are now guarded by a mutex since (re-)arming can race with `Cancel()`/`Fail()` from the scheduler loop. Known gap, not fixed here: the wait for dist confirmation between a step's RPC returning and the step being marked finished is still covered by whichever deadline was last armed, not its own budget — see the TODOs in `task.go`'s `ActivateDeadline`.

**Synchronous LoadBalance wait bound moved from `task.Wait()`'s parameter to the caller's `ctx`**: `task.Wait()` no longer takes its own timeout parameter -- it just blocks until every task finishes or the caller's `ctx` is done, whichever comes first. The bound that used to live inside it is now applied explicitly at the `handlers.go` call sites via `context.WithTimeout(ctx, SegmentTaskTimeout/ChannelTaskTimeout)`, matching master's pre-PR 1x semantics. This bound is *not* redundant with the task's own deadline, even though a task's deadline is also driven by the same config value: a task's deadline only arms once it clears executor admission (`Task.ActivateDeadline`, called from `Execute()`); if the target node's executor stays saturated (e.g. a large collection's initial load, or bulk rebuild after node recovery), a task can be rejected at admission indefinitely without the deadline ever arming, and nothing else (`checkStale` has no task-age criterion) bounds that wait. Combined with callers that pass no deadline of their own (pymilvus's `utility.load_balance` defaults to `timeout=None`), the handler needs its own explicit bound regardless of the task's deadline -- so this is a genuinely separate concern ("how long is this RPC willing to wait") from the task's own execution budget, not a duplicate guess at it.

**`task.Wait()` (used by the synchronous LoadBalance path) no longer has a false-success race**: it previously wrote its result via a shared `err` variable from a background goroutine with no synchronization and read it after `<-ctx.Done()`; if the outer timeout fired before the goroutine finished, it returned a false `nil` (success) and raced on the variable. Switched to an error channel + `select` so a real timeout is reported as `ctx.Err()`.

---

Two follow-ups from the cancellation audit are folded in as separate commits:

- **Legacy deltalog download honors ctx**: `NewLegacyDeltalogReader` invoked its downloader with `context.Background()`, making the L0/deltalog download during segment load immune to the deadline. `ctx` is now threaded through `NewDeltalogReader`/`NewDeltalogReaderFromBinlogs`/`NewLegacyDeltalogReader` (mechanical; all call sites already had a ctx in scope). Note: `NewDeltalogReaderFromBinlogs`'s StorageV2/V3 branch accepts `ctx` for signature symmetry but doesn't check it (no per-file boundary to check at, and no callers today); its V1 branch and `NewDeltalogReader`'s V2/V3 branch (which does have a per-file boundary) both honor it.
- **Canceled watch is not committed**: `WatchDmChannels` now checks `ctx.Err()` once at the commit point (before `pipeline.Start()`/`delegator.Start()`), so a request whose caller timed out rolls back through the existing cleanup chain instead of leaving the node subscribed and racing the re-issued watch's dispatcher registration.
- **Singleflight join honors the joiner's ctx**: `delegator` `sLoad` joined an identical in-flight load via `singleflight.Do`, which ignores the joiner's context; switched to `DoChan` raced against `ctx.Done()` — the joiner exits on its own deadline while the shared load keeps running for its initiator.
- **Packed deltalog iteration checks ctx at file boundaries**: the StorageV2/V3 per-file FFI read cannot be interrupted mid-file; a canceled load now stops between files.

**Known residual gaps (not addressed here, tracked separately)**:
- Vector index load (`CacheIndexToDisk`/`Deserialize` in knowhere/segcore) has no cancellation token, so a timed-out task's underlying C++ load keeps running in the background for the node it was originally dispatched to (a same-node retry joins the in-flight load; a cross-node retry duplicates memory/bandwidth until the original finishes) — tracked at zilliztech/knowhere#1771.
- `unsub`'s WAL `ManualFlush` append is intentionally `WithoutCancel` (WAL consistency invariant, see #45077/#45078) and can accumulate blocked goroutines during a sustained WAL outage.
- **No backoff on repeated timeout-driven rebuild (TODO'd in code, not fixed here)**: each per-task deadline assumes a segment/channel op finishes within its configured timeout (`segmentTaskTimeout` defaults to 5min, `channelTaskTimeout` to 2min). If the real op time is consistently longer -- a large disk-index segment on throttled cold storage, or a channel with a large L0/growing backlog -- `SegmentChecker`/`ChannelChecker` kills and rebuilds the task with the *same* budget on every check tick (`checkSegmentInterval`/`checkChannelInterval`, default 3s), forever: no backoff, no retry cap, so that segment/channel never converges and the node keeps re-paying the full load I/O every ~timeout+3s. This is not a regression against base (where the timeout never fired, so slow-but-healthy ops eventually succeeded at the cost of never cleaning up genuinely stuck ones) -- this PR makes that tradeoff's cliff real. Fixing it properly means either backoff/a retry cap on `DeadlineExceeded`-driven rebuilds, or redefining the timeout as a no-progress bound instead of a flat wall-clock budget (the latter needs real load-progress reporting from the QueryNode, which doesn't exist today -- estimated as a much larger, separate effort). TODOs left at `task.go`'s `ActivateDeadline`, `SegmentChecker.createSegmentLoadTasks`, and `ChannelChecker`'s equivalent.

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -502,7 +502,7 @@ queryCoord:
   autoBalanceInterval: 3000 # the interval for triggerauto balance
   checkIndexInterval: 10000
   channelTaskTimeout: 120000 # 2 minutes
-  segmentTaskTimeout: 180000 # 3 minutes
+  segmentTaskTimeout: 300000 # 5 minutes
   distPullInterval: 500
   dispatchInterval: 500
   heartbeatAvailableInterval: 10000 # 10s, Only QueryNodes which fetched heartbeats within the duration are available
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -501,8 +501,8 @@ queryCoord:
   checkBalanceInterval: 300
   autoBalanceInterval: 3000 # the interval for triggerauto balance
   checkIndexInterval: 10000
-  channelTaskTimeout: 60000 # 1 minute
-  segmentTaskTimeout: 120000 # 2 minute
+  channelTaskTimeout: 300000 # 5 minutes
+  segmentTaskTimeout: 300000 # 5 minutes
   distPullInterval: 500
   dispatchInterval: 500
   heartbeatAvailableInterval: 10000 # 10s, Only QueryNodes which fetched heartbeats within the duration are available
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -501,8 +501,8 @@ queryCoord:
   checkBalanceInterval: 300
   autoBalanceInterval: 3000 # the interval for triggerauto balance
   checkIndexInterval: 10000
-  channelTaskTimeout: 300000 # 5 minutes
-  segmentTaskTimeout: 300000 # 5 minutes
+  channelTaskTimeout: 120000 # 2 minutes
+  segmentTaskTimeout: 120000 # 2 minutes
   distPullInterval: 500
   dispatchInterval: 500
   heartbeatAvailableInterval: 10000 # 10s, Only QueryNodes which fetched heartbeats within the duration are available
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -502,7 +502,7 @@ queryCoord:
   autoBalanceInterval: 3000 # the interval for triggerauto balance
   checkIndexInterval: 10000
   channelTaskTimeout: 120000 # 2 minutes
-  segmentTaskTimeout: 120000 # 2 minutes
+  segmentTaskTimeout: 180000 # 3 minutes
   distPullInterval: 500
   dispatchInterval: 500
   heartbeatAvailableInterval: 10000 # 10s, Only QueryNodes which fetched heartbeats within the duration are available
--- a/internal/compaction/common.go
+++ b/internal/compaction/common.go
@@ -93,7 +93,7 @@ func readDeltalogsV1(
 
 	for _, deltalog := range deltalogs {
 		for _, binlog := range deltalog.Binlogs {
-			reader, err := storage.NewDeltalogReader(pkType, []string{binlog.GetLogPath()}, option...)
+			reader, err := storage.NewDeltalogReader(ctx, pkType, []string{binlog.GetLogPath()}, option...)
 			if err != nil {
 				return nil, nil, err
 			}
@@ -118,7 +118,7 @@ func readDeltalogsV2(
 	paths []string,
 	option ...storage.RwOption,
 ) ([]storage.PrimaryKey, []typeutil.Timestamp, error) {
-	reader, err := storage.NewDeltalogReader(pkType, paths,
+	reader, err := storage.NewDeltalogReader(ctx, pkType, paths,
 		append(option, storage.WithVersion(storage.StorageV3))...)
 	if err != nil {
 		return nil, nil, err
--- a/internal/datanode/external/milvus_table_deltalog.go
+++ b/internal/datanode/external/milvus_table_deltalog.go
@@ -587,6 +587,7 @@ func (t *RefreshExternalCollectionTask) newMilvusTableSourceDeltalogReader(
 ) (storage.RecordReader, error) {
 	if packed.IsMilvusTableStorageV3DeltalogPath(ref.sourcePath) {
 		return storage.NewDeltalogReader(
+			ctx,
 			sourcePKType,
 			[]string{ref.sourcePath},
 			storage.WithVersion(storage.StorageV3),
@@ -598,6 +599,7 @@ func (t *RefreshExternalCollectionTask) newMilvusTableSourceDeltalogReader(
 		return nil, err
 	}
 	return storage.NewDeltalogReader(
+		ctx,
 		sourcePKType,
 		[]string{ref.sourcePath},
 		storage.WithVersion(storage.StorageV1),
--- a/internal/datanode/external/milvus_table_deltalog_test.go
+++ b/internal/datanode/external/milvus_table_deltalog_test.go
@@ -1313,6 +1313,7 @@ func writeDeltalog(
 func readInt64Deltalog(t *testing.T, storageConfig *indexpb.StorageConfig, path string) ([]int64, []int64) {
 	t.Helper()
 	reader, err := storage.NewDeltalogReader(
+		context.Background(),
 		schemapb.DataType_Int64,
 		[]string{path},
 		storage.WithVersion(storage.StorageV3),
--- a/internal/datanode/external/milvus_table_deltalog_test.go
+++ b/internal/datanode/external/milvus_table_deltalog_test.go
@@ -509,7 +509,7 @@ func (s *RefreshExternalCollectionTaskSuite) TestCreateManifestForSegment_Milvus
 	)
 	s.Require().NoError(err)
 	mockDeltalogReader := mockey.Mock(storage.NewDeltalogReader).
-		To(func(pkType schemapb.DataType, paths []string, option ...storage.RwOption) (storage.RecordReader, error) {
+		To(func(_ context.Context, pkType schemapb.DataType, paths []string, option ...storage.RwOption) (storage.RecordReader, error) {
 			s.Equal(schemapb.DataType_Int64, pkType)
 			s.Equal([]string{sourceDeltalogPath}, paths)
 			return &fakeMilvusTableDeltalogReader{records: []storage.Record{record}}, nil
@@ -1069,7 +1069,7 @@ func (s *RefreshExternalCollectionTaskSuite) TestLoadMilvusTableSourceDeltalogDe
 		s.Require().NoError(err)
 		reader := &fakeMilvusTableDeltalogReader{records: []storage.Record{record}}
 		mockReader := mockey.Mock(storage.NewDeltalogReader).
-			To(func(pkType schemapb.DataType, paths []string, option ...storage.RwOption) (storage.RecordReader, error) {
+			To(func(_ context.Context, pkType schemapb.DataType, paths []string, option ...storage.RwOption) (storage.RecordReader, error) {
 				s.Equal(schemapb.DataType_Int64, pkType)
 				s.Equal([]string{ref.sourcePath}, paths)
 				return reader, nil
@@ -1095,7 +1095,7 @@ func (s *RefreshExternalCollectionTaskSuite) TestLoadMilvusTableSourceDeltalogDe
 		legacyRef := milvusTableDeltalogRef{sourcePath: "files/delta_log/1/2/3/10", logID: 10, numEntries: 2}
 		reader := &fakeMilvusTableDeltalogReader{}
 		mockReader := mockey.Mock(storage.NewDeltalogReader).
-			To(func(pkType schemapb.DataType, paths []string, option ...storage.RwOption) (storage.RecordReader, error) {
+			To(func(_ context.Context, pkType schemapb.DataType, paths []string, option ...storage.RwOption) (storage.RecordReader, error) {
 				s.Equal(schemapb.DataType_Int64, pkType)
 				s.Equal([]string{legacyRef.sourcePath}, paths)
 				return reader, nil
--- a/internal/querycoordv2/checkers/channel_checker.go
+++ b/internal/querycoordv2/checkers/channel_checker.go
@@ -288,6 +288,10 @@ func (c *ChannelChecker) createChannelLoadTask(ctx context.Context, channels []*
 		plans[i].Replica = replica
 	}
 
+	// TODO: same known limitation as SegmentChecker.createSegmentLoadTasks --
+	// a channel whose real watch time (L0/growing backlog, seek distance)
+	// consistently exceeds ChannelTaskTimeout never converges: killed and
+	// rebuilt with the same budget every check tick, no backoff or retry cap.
 	return balance.CreateChannelTasksFromPlans(ctx, c.ID(), Params.QueryCoordCfg.ChannelTaskTimeout.GetAsDuration(time.Millisecond), plans)
 }
 
--- a/internal/querycoordv2/checkers/leader_checker.go
+++ b/internal/querycoordv2/checkers/leader_checker.go
@@ -23,6 +23,7 @@ import (
 	"golang.org/x/time/rate"
 
 	"github.com/milvus-io/milvus/internal/querycoordv2/meta"
+	. "github.com/milvus-io/milvus/internal/querycoordv2/params"
 	"github.com/milvus-io/milvus/internal/querycoordv2/session"
 	"github.com/milvus-io/milvus/internal/querycoordv2/task"
 	"github.com/milvus-io/milvus/internal/querycoordv2/utils"
@@ -137,6 +138,7 @@ func (c *LeaderChecker) findNeedSyncPartitionStats(ctx context.Context, replica 
 
 		t := task.NewLeaderPartStatsTask(
 			ctx,
+			Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond),
 			c.ID(),
 			leaderView.CollectionID,
 			replica,
@@ -185,6 +187,7 @@ func (c *LeaderChecker) findNeedLoadedSegments(ctx context.Context, replica *met
 			action := task.NewLeaderAction(leaderView.ID, s.Node, task.ActionTypeGrow, s.GetInsertChannel(), s.GetID(), time.Now().UnixNano())
 			t := task.NewLeaderSegmentTask(
 				ctx,
+				Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond),
 				c.ID(),
 				s.GetCollectionID(),
 				replica,
@@ -230,6 +233,7 @@ func (c *LeaderChecker) findNeedRemovedSegments(ctx context.Context, replica *me
 		action := task.NewLeaderAction(leaderView.ID, leaderView.ID, task.ActionTypeReduce, leaderView.Channel, sid, 0)
 		t := task.NewLeaderSegmentTask(
 			ctx,
+			Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond),
 			c.ID(),
 			leaderView.CollectionID,
 			replica,
--- a/internal/querycoordv2/checkers/segment_checker.go
+++ b/internal/querycoordv2/checkers/segment_checker.go
@@ -527,6 +527,13 @@ func (c *SegmentChecker) createSegmentLoadTasks(ctx context.Context, segments []
 		plans = append(plans, shardPlans...)
 	}
 
+	// TODO: this assumes a single segment always finishes loading within
+	// SegmentTaskTimeout (3min default). If a segment's real load time is
+	// consistently longer (large disk-index segment, throttled cold storage),
+	// the task is killed by its deadline every round and rebuilt here with
+	// the same budget on the next check tick -- it never converges. Needs
+	// either backoff/a retry cap on repeated DeadlineExceeded rebuilds, or a
+	// no-progress timeout instead of a flat per-task wall-clock budget.
 	return balance.CreateSegmentTasksFromPlans(ctx, c.ID(), Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond), plans)
 }
 
--- a/internal/querycoordv2/checkers/segment_checker.go
+++ b/internal/querycoordv2/checkers/segment_checker.go
@@ -528,7 +528,7 @@ func (c *SegmentChecker) createSegmentLoadTasks(ctx context.Context, segments []
 	}
 
 	// TODO: this assumes a single segment always finishes loading within
-	// SegmentTaskTimeout (3min default). If a segment's real load time is
+	// SegmentTaskTimeout (5min default). If a segment's real load time is
 	// consistently longer (large disk-index segment, throttled cold storage),
 	// the task is killed by its deadline every round and rebuilt here with
 	// the same budget on the next check tick -- it never converges. Needs
--- a/internal/querycoordv2/handlers.go
+++ b/internal/querycoordv2/handlers.go
@@ -158,12 +158,14 @@ func (s *Server) balanceSegments(ctx context.Context,
 	}
 
 	if sync {
-		// The task's own deadline doesn't start until it's actually dispatched
-		// (see Task.ActivateDeadline), so give this caller's wait window some
-		// slack beyond SegmentTaskTimeout to cover queueing time; otherwise a
-		// busy scheduler could make this synchronous call time out before the
-		// task itself ever gets a chance to fail with the real error.
-		err := task.Wait(ctx, 2*Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond), tasks...)
+		// No extra timeout here: each task already carries its own real
+		// deadline (armed on first dispatch, see Task.ActivateDeadline) and
+		// is guaranteed to finish within it. Bounding this wait with another,
+		// independently-guessed duration would only race against that real
+		// deadline -- as it did before, when both used SegmentTaskTimeout and
+		// this wait usually won, masking the task's actual result. Callers
+		// that want to bound how long they wait should set a deadline on ctx.
+		err := task.Wait(ctx, tasks...)
 		if err != nil {
 			msg := "failed to wait all balance task finished"
 			mlog.Warn(ctx, msg, mlog.Err(err))
@@ -241,10 +243,9 @@ func (s *Server) balanceChannels(ctx context.Context,
 	}
 
 	if sync {
-		// See the matching comment in balanceSegments: give this wait window
-		// slack beyond ChannelTaskTimeout so it doesn't time out purely due to
-		// scheduler queueing before the task's own deadline is even armed.
-		err := task.Wait(ctx, 2*Params.QueryCoordCfg.ChannelTaskTimeout.GetAsDuration(time.Millisecond), tasks...)
+		// See the matching comment in balanceSegments: no extra timeout here,
+		// each task already has a real, guaranteed-to-fire deadline of its own.
+		err := task.Wait(ctx, tasks...)
 		if err != nil {
 			msg := "failed to wait all balance task finished"
 			mlog.Warn(ctx, msg, mlog.Err(err))
--- a/internal/querycoordv2/handlers.go
+++ b/internal/querycoordv2/handlers.go
@@ -158,7 +158,12 @@ func (s *Server) balanceSegments(ctx context.Context,
 	}
 
 	if sync {
-		err := task.Wait(ctx, Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond), tasks...)
+		// The task's own deadline doesn't start until it's actually dispatched
+		// (see Task.ActivateDeadline), so give this caller's wait window some
+		// slack beyond SegmentTaskTimeout to cover queueing time; otherwise a
+		// busy scheduler could make this synchronous call time out before the
+		// task itself ever gets a chance to fail with the real error.
+		err := task.Wait(ctx, 2*Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond), tasks...)
 		if err != nil {
 			msg := "failed to wait all balance task finished"
 			mlog.Warn(ctx, msg, mlog.Err(err))
@@ -236,7 +241,10 @@ func (s *Server) balanceChannels(ctx context.Context,
 	}
 
 	if sync {
-		err := task.Wait(ctx, Params.QueryCoordCfg.ChannelTaskTimeout.GetAsDuration(time.Millisecond), tasks...)
+		// See the matching comment in balanceSegments: give this wait window
+		// slack beyond ChannelTaskTimeout so it doesn't time out purely due to
+		// scheduler queueing before the task's own deadline is even armed.
+		err := task.Wait(ctx, 2*Params.QueryCoordCfg.ChannelTaskTimeout.GetAsDuration(time.Millisecond), tasks...)
 		if err != nil {
 			msg := "failed to wait all balance task finished"
 			mlog.Warn(ctx, msg, mlog.Err(err))
--- a/internal/querycoordv2/handlers.go
+++ b/internal/querycoordv2/handlers.go
@@ -158,14 +158,19 @@ func (s *Server) balanceSegments(ctx context.Context,
 	}
 
 	if sync {
-		// No extra timeout here: each task already carries its own real
-		// deadline (armed on first dispatch, see Task.ActivateDeadline) and
-		// is guaranteed to finish within it. Bounding this wait with another,
-		// independently-guessed duration would only race against that real
-		// deadline -- as it did before, when both used SegmentTaskTimeout and
-		// this wait usually won, masking the task's actual result. Callers
-		// that want to bound how long they wait should set a deadline on ctx.
-		err := task.Wait(ctx, tasks...)
+		// This bound exists for a different reason than the task's own
+		// deadline: a task only gets its real deadline once it's actually
+		// admitted by the executor (see Task.ActivateDeadline). If the target
+		// node's executor stays saturated, the task can sit rejected in the
+		// queue indefinitely without ever being admitted, and callers commonly
+		// pass no deadline of their own (e.g. pymilvus defaults to
+		// timeout=None) -- so without a bound here, this handler could hang
+		// forever. SegmentTaskTimeout is a reasonable ceiling for "how long
+		// this synchronous call is willing to wait" independent of whether it
+		// matches the task's own execution budget.
+		waitCtx, cancel := context.WithTimeout(ctx, Params.QueryCoordCfg.SegmentTaskTimeout.GetAsDuration(time.Millisecond))
+		defer cancel()
+		err := task.Wait(waitCtx, tasks...)
 		if err != nil {
 			msg := "failed to wait all balance task finished"
 			mlog.Warn(ctx, msg, mlog.Err(err))
@@ -243,9 +248,10 @@ func (s *Server) balanceChannels(ctx context.Context,
 	}
 
 	if sync {
-		// See the matching comment in balanceSegments: no extra timeout here,
-		// each task already has a real, guaranteed-to-fire deadline of its own.
-		err := task.Wait(ctx, tasks...)
+		// See the matching comment in balanceSegments.
+		waitCtx, cancel := context.WithTimeout(ctx, Params.QueryCoordCfg.ChannelTaskTimeout.GetAsDuration(time.Millisecond))
+		defer cancel()
+		err := task.Wait(waitCtx, tasks...)
 		if err != nil {
 			msg := "failed to wait all balance task finished"
 			mlog.Warn(ctx, msg, mlog.Err(err))
--- a/internal/querycoordv2/task/executor.go
+++ b/internal/querycoordv2/task/executor.go
@@ -175,6 +175,12 @@ func (ex *Executor) Execute(task Task, step int) bool {
 		}
 	}
 
+	// Only now is the task actually admitted to run: it cleared the wait
+	// queue and the executor's capacity gates above. Arm its deadline here
+	// (no-op after the first action/step) so time spent queued or bounced
+	// by admission doesn't count against the RPC budget.
+	task.ActivateDeadline()
+
 	go func() {
 		mlog.Info(context.TODO(), "execute the action of task")
 		switch task.Actions()[step].(type) {
--- a/internal/querycoordv2/task/executor.go
+++ b/internal/querycoordv2/task/executor.go
@@ -175,11 +175,13 @@ func (ex *Executor) Execute(task Task, step int) bool {
 		}
 	}
 
-	// Only now is the task actually admitted to run: it cleared the wait
-	// queue and the executor's capacity gates above. Arm its deadline here
-	// (no-op after the first action/step) so time spent queued or bounced
-	// by admission doesn't count against the RPC budget.
-	task.ActivateDeadline()
+	// Only now is this step actually admitted to run: it cleared the wait
+	// queue and the executor's capacity gates above. Arm a fresh deadline for
+	// this step here (no-op if already armed for it) so time spent queued or
+	// bounced by admission doesn't count against the RPC budget, and a
+	// later step in a multi-action task gets its own full budget rather than
+	// sharing whatever was left of an earlier step's.
+	task.ActivateDeadline(step)
 
 	go func() {
 		mlog.Info(context.TODO(), "execute the action of task")
--- a/internal/querycoordv2/task/executor_test.go
+++ b/internal/querycoordv2/task/executor_test.go
@@ -398,7 +398,7 @@ func TestExecutorDeadlockReproduction(t *testing.T) {
 	// In the old single-pool design, this would fail because all 5 slots would need to be full
 	// But with split pools, non-channel has its own capacity of 3
 	leaderAction := NewLeaderAction(1, 1, ActionTypeGrow, "shard-0", 700, 1)
-	leaderTask := NewLeaderSegmentTask(ctx, testSource("test"), 1000, replica, 1, leaderAction)
+	leaderTask := NewLeaderSegmentTask(ctx, 10*time.Second, testSource("test"), 1000, replica, 1, leaderAction)
 	leaderTask.SetID(700)
 
 	// Directly check: can we increment the non-channel counter?
--- a/internal/querycoordv2/task/task.go
+++ b/internal/querycoordv2/task/task.go
@@ -19,6 +19,7 @@ package task
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -88,6 +89,10 @@ type Task interface {
 	SetPriority(priority Priority)
 	Index() string // dedup indexing string
 
+	// ActivateDeadline starts the task's execution timeout the first time it is
+	// actually dispatched to a QueryNode, so time spent waiting for a scheduler
+	// slot doesn't count against the RPC budget. No-op after the first call.
+	ActivateDeadline()
 	// cancel the task as we don't need to continue it
 	Cancel(err error)
 	// fail the task as we encounter some error so be unable to continue,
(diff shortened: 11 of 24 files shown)


czs007:

This PR makes `queryCoord.segmentTaskTimeout` / `channelTaskTimeout` effective for the first time and raises their defaults, which changes task lifetime semantics in ways that reach user-visible load and balance paths.

## High

**`internal/querycoordv2/task/task.go:141` — Task deadline starts at construction, so one timeout budget covers queueing, every action, and dist confirmation rather than a single RPC.**
`newBaseTask` now calls `context.WithTimeout` at construction time; before this PR the `timeout` argument did not exist and the task ctx was only `context.WithCancel`, so it never expired. Failure: a segment whose load legitimately exceeds 180s (large disk-index segment, S3 throttling, slow node), or a task that queues for tens of seconds before executing → the ctx expires on `ex.cluster.LoadSegments(task.Context(), ...)` at `executor.go:278`, the task fails with `DeadlineExceeded`, and the checker's rebuilt task gets the same 180s budget and is cut at the same point every round; after `LoadTimeoutSeconds` (600s) `CollectionObserver` removes the collection and the user's load fails with `CollectionNotLoaded: context deadline exceeded`, leaving a 24h entry in `GlobalFailedLoadCache`. Note this is intermittent by construction: at expiry, `preProcess` → `check` → `Cancel` (`scheduler.go:1166-1167`, status `Canceled`, no cache write) races `executor.go:229-232` → `Fail` (cache write) on the same `canceled` CAS, so the cache-poisoning half may not reproduce on a single run.
Suggestion: start the budget when the task first executes successfully, or scope the deadline to individual RPCs rather than the whole task lifetime. The `timeout <= 0 → WithCancel` escape hatch at `task.go:142-149` is worth documenting explicitly as the rollback knob — setting `queryCoord.segmentTaskTimeout` to `0` restores pre-PR semantics (`PanicIfEmpty` does not reject `0`). *(raised by bigsheeper, czs007)*

## Medium

**`internal/querycoordv2/handlers.go:161` — The synchronous `LoadBalance` wait window and its semantics both change; the PR body's "keeps its historical 2-minute worst-case wait" no longer holds.**
The final commit raises `SegmentTaskTimeout`'s default to `180000` (`component_param.go:3252-3258`), and `handlers.go:161` waits on that same config item; the channel path at `handlers.go:239` doubles likewise with `ChannelTaskTimeout` 60000→120000. This is a user-visible RPC (`services.go:595` → `services.go:682`, `sync=true`, i.e. `milvuspb.LoadBalance` / pymilvus `utility.load_balance`). Failure: an operator balances a segment whose load exceeds 180s → the task ctx expires at `executor.go:278`, the task fails, `task.Wait()` returns that error immediately, and `handlers.go:163-166` returns "failed to wait all balance task finished" — the move is abandoned, where before this PR the same call returned success and completed in the background. Separately, an SDK configured with the historical 120s client timeout now disconnects before querycoord responds, so the caller sees its own timeout rather than the real cause. Also note the task is built at T0 (`handlers.go:129-131`) and the wait starts at T1 > T0 with the same value, so the task ctx always expires first and the `case <-ctx.Done()` branch at `utils.go:73-74` is effectively unreachable for this caller.
Suggestion: correct that note in the PR body, and consider decoupling the wait window from the task deadline (e.g. `timeout * 2`, or a dedicated config item). *(raised by bigsheeper, czs007)*

**`internal/querycoordv2/task/executor.go:149` — A task repeatedly rejected at the executor admission layer burns its deadline and can be cancelled before its first RPC.**
Backpressure is not in the wait queue — `tryPromoteAll` (`scheduler.go:861-892`) ranges it uncapped and `promote` only refuses on `check()` failure. It is one level down: `Executor.Execute` increments `channelTaskNum` / `nonChannelTaskNum` and, on exceeding `GetChannelTaskCap()` / `GetNonChannelTaskCap()`, decrements, removes the index and returns `false`. The rejected task stays in `processQueue`, `preProcess` keeps returning `shouldProcess = true`, and it is re-offered every `dispatchInterval` (500ms) while its deadline runs down. Failure: concurrent non-channel tasks on a QueryNode exceed `GetNonChannelTaskCap()` (≈112 on 8 cores) — e.g. first load of a large collection, or bulk segment rebuild after node recovery → tasks beyond the cap are rejected every 500ms and exhaust their 180s budget without ever issuing a `LoadSegments` RPC, then get `Cancel`led via `preProcess` → `check`; if saturation persists through `LoadTimeoutSeconds` (600s) the collection load fails without progress.
Suggestion: fix this together with the item above — exclude queued/rejected time from the budget, or start the deadline at first successful `Execute`. Expiry here goes through `Cancel` (status `Canceled`), so it does not reach `recordSegmentTaskError` and does not poison the failed-load cache. *(raised by czs007)*

## Low

**`internal/querynodev2/segments/segment_loader_test.go:1353` — mockey hooks were not updated for the new leading `ctx` parameter; mockey silently adapts and discards it.**
`NewDeltalogReader` gained a leading `ctx` (`rw.go:521-526`), but 10 body-form hooks still declare `func(pkType schemapb.DataType, paths []string, option ...storage.RwOption)`: `segment_loader_test.go:142,199,892,981,1052,1284,1354` and `milvus_table_deltalog_test.go:512,1072,1098`. (The 5 sites at `milvus_table_deltalog_test.go:1119,1135,1155,1172,1187` use `.Return(...)` and carry no hook signature, so they are unaffected.) mockey v1.4.6's `AnalyzerImpl.InputAdapter` fails the exact match, falls through to `nonGenericAnalyzer`, which validates with `CheckFuncArgs(a.TargetType(), inputType, 1, 0)` and returns `targetArgs[1:]` — a hook short one leading argument is accepted as the "method receiver" shape and the first argument is dropped. Verified with a standalone minimal repro (mockey v1.4.6, `-gcflags="all=-N -l"`): the build passes, the hook is invoked, the remaining arguments are not shifted, and assertions still hold — so these mocks neither panic nor warn, they simply observe nothing about the ctx.
Suggestion: add `ctx context.Context` to the hook signatures so the new cancellation/timeout propagation path is actually covered, and so future signature changes to `NewDeltalogReader` fail loudly here. *(raised by bigsheeper, czs007)*

**`internal/storage/rw.go:579` — The new `ctx` parameter on `NewDeltalogReaderFromBinlogs` is unused in the StorageV2/V3 branches.**
Only the StorageV1 branch consumes it (`rw.go:607`, passed to `NewLegacyDeltalogReader`); `rw.go:608-638` never references `ctx`, and `newFFIPackedRecordReaderFromFragments` takes no ctx. This differs structurally from `NewDeltalogReader`, whose V2/V3 path returns an `IterativeRecordReader` and therefore has a per-file boundary where this PR added a check (`rw.go:557-561`) — here the reader is built once from fragments, so there is no equivalent boundary. The accurate framing is that `ctx` is a dummy parameter on this path, not that a matching boundary check is missing. The function has zero call sites repo-wide (grep hits only the definition and its comment), so there is no impact today.
Suggestion: either document that V2/V3 ignores cancellation, or drop the parameter until a caller needs it — otherwise the signature suggests a cancellation guarantee this branch does not provide. *(raised by czs007)*

czs007:

This PR makes the previously inert `segmentTaskTimeout`/`channelTaskTimeout` effective for the first time; the findings below concern the per-task deadline's scope and lifecycle, plus one test and one documentation mismatch at HEAD.

## High

**`internal/querycoordv2/task/task.go:190` — A single task-level deadline covers every remaining action RPC plus the inter-action dist confirmation, with no backoff on rebuild.**
The queueing part of the original concern is resolved (the deadline is now armed by `Executor.Execute` via `ActivateDeadline()` after admission), but once armed, one `context.WithTimeout` bounds all remaining action RPCs *and* the dist-confirmation waits between them, and `ex.cluster.LoadSegments(task.Context(), ...)` synchronously waits for the full segment load. Failure: a sealed segment whose real load exceeds 180s (large disk-index segment, or throttled cold S3 fetch) hits `DeadlineExceeded` → `scheduler.check()` (`scheduler.go:1211`) cancels/fails the task → `SegmentChecker` rebuilds it next round with the same `Params.QueryCoordCfg.SegmentTaskTimeout` (`segment_checker.go:530/539/568`, same pattern in `leader_checker.go`) → the segment never comes online, `LoadCollection` never returns, and every round repeats the full download/index-build I/O. A grep for `backoff|cooldown|retry` across `segment_checker.go` and `scheduler.go` returns no hits, so the rebuild interval is bounded only by the checker period and the timeout is never widened. Note this is not a regression against base — on master `newBaseTask` used `context.WithCancel` and the timeout never fired, so such segments did load — but the change makes the cliff real and the PR does not evidence the "a legitimate load is always < 180s" premise it now depends on. Suggestion: apply exponential backoff plus timeout escalation on `DeadlineExceeded`-driven rebuilds, or scale the timeout by segment size/row count; failing that, state it explicitly in the PR body as a known tradeoff. *(raised by xiaocai2333, xaxys)*

**`internal/querycoordv2/task/task_test.go:3286` — `TestChannelTaskTimeoutBoundsTaskContext`, added by this PR, fails at PR HEAD.**
The test was added by an earlier commit in this PR and asserts that `Context()` carries a deadline immediately after construction, but the last commit (`eb42d3c53a`) moved the deadline to `ActivateDeadline()` at first dispatch, leaving `newBaseTask` with only `context.WithCancel` (`task.go:162`). All three assertions (channel/segment/leader) miss. Failure: running `go test ./internal/querycoordv2/task/` (which CI's `go ut` does) → `task_test.go:3287` `channel task context must carry the task timeout as a deadline` fails, then `task_test.go:3292` `channel task context did not expire after the task timeout` calls `t.Fatal`, turning the package and CI red. Confirmed by running it on a PR HEAD worktree: this is the only failing test in the package, so it is not environmental. Suggestion: either call `ActivateDeadline()` before asserting, or restate the test as "no deadline after construction; deadline present and expiring after activation". *(surfaced during verification)*

## Medium

**`internal/querycoordv2/task/executor.go:182` — Multi-action (move) tasks share one deadline that also absorbs inter-action dist/delegator-promotion waits, so the reduce step can be starved of budget.**
`ActivateDeadline()` uses `sync.Once`, so the timeout is armed once at step 0 and reused by every later step, while a move task must wait between steps for dist confirmation (`preProcess` → `checkActionFinish` requires `segmentAction.isDistMatched`) and, for channel moves, for the new delegator to become shard leader (`scheduler.go:1146`). Failure: a channel move (`[Grow@dst, Reduce@src]`, budget = `channelTaskTimeout` = 120s) where `WatchDmChannels` on dst takes ~100s due to many L0/growing segments and delegator promotion consumes the remaining 20s → the deadline fires before reduce is dispatched, `scheduler.check()` returns `DeadlineExceeded` in `preProcess` and `task.Cancel(err)` runs (`scheduler.go:1166`), leaving both src and dst subscribed to the same channel with two active delegators (double consumption, double memory) until `ChannelChecker` re-issues the release on a later round. It self-heals, but at the cost of a wasted watch/load round. Suggestion: make the deadline per-action — re-arm a fresh action-level timeout on each `Execute` — rather than a single per-task budget. *(surfaced during verification)*

## Low

**`internal/querycoordv2/task/task.go:156` — The PR description no longer matches HEAD.**
The code comment mismatch was fixed in `feb267ed3f`/`eb42d3c53a` (`task.go:156-161` now states the deadline is not applied at construction and is armed by `ActivateDeadline` on first dispatch), but the PR body was not updated and now contradicts the code in two places: (1) it says the timeout "bounds the whole task (queueing + RPCs)", whereas after `eb42d3c53a` queueing time is specifically excluded; (2) it says "Both defaults are unified to **2 minutes**", whereas HEAD ships channel 120s / segment 180s (`component_param.go:3245/3255`, `configs/milvus.yaml:504-505`). Suggestion: correct both statements, and add the caveat that segments whose real load exceeds 180s still hit a fail-and-rebuild loop — phrased as a known tradeoff that moves the cliff, not as one that removes it. *(raised by xiaocai2333)*

czs007:
> **Must-fix issues introduced by this PR: 3**
> **Merge recommendation:** Mergeable after the 3 must-fix items below are fixed in this PR.

This PR makes task-level deadlines real for the first time, and that change surfaces three high-severity issues: the synchronous `LoadBalance` path lost its only upper time bound, and two paths (segment load, channel watch) can now enter a non-converging failure-and-rebuild loop.

## High

**`internal/querycoordv2/handlers.go:168` — synchronous `LoadBalance` has no deadline once a task is never admitted.**
Commit `be4e1e4408` dropped the `timeout` parameter from `task.Wait` (`task/utils.go:62`); `Wait` now returns only when all tasks finish or the caller's ctx is done. The comment's premise — that each task already carries its own real deadline — does not hold: a deadline is armed only after a task passes executor admission, via `Executor.Execute` → `task.ActivateDeadline()` (`task/executor.go:182`). When `nonChannelTaskNum` exceeds `GetNonChannelTaskCap`, `Execute` returns false at the capacity gate (`task/executor.go:169-176`), and when the node has no executor, `process()` returns false (`task/scheduler.go:1203-1205`); the task stays in `processQueue` with a ctx derived from the service-level `s.ctx` (`handlers.go:129`), which has no deadline. The scheduler's fallback `check()` only inspects `task.Context().Err()` (nil here) and `checkStale`, which tests replica/node liveness and has no task-age criterion (`task/scheduler.go:1211`, `1372-1440`). The sole `sync=true` caller is the `LoadBalance` handler (`services.go:682`), which forwards the gRPC ctx unwrapped.
Failure: the target QueryNode's non-channel executor pool stays saturated while an operator runs `utility.load_balance(...)` (pymilvus default `timeout=None`, so no gRPC deadline) against a healthy node → the SegmentTask is bounced at the capacity gate every round, `ActivateDeadline` is never called, and `task.Wait(ctx, tasks...)` blocks forever, hanging the handler goroutine and the client RPC; before this change the call was bounded by `SegmentTaskTimeout` (120s).
Suggestion: wrap the ctx with `context.WithTimeout` before the `sync=true` call at `services.go:682`, or apply a task-timeout-derived deadline to the ctx passed to `Wait` inside the `sync` branches of `balanceSegments`/`balanceChannels`. Please also correct the "guaranteed to finish within it" comments at `task/utils.go:54-61` and `handlers.go:161-167`, which no longer describe the actual behaviour. *(raised by xiaocai2333, xaxys)*

**`internal/querycoordv2/task/task.go:194` — one task-level deadline now spans the whole segment flow, with no backoff on failure.**
Before this PR, `newBaseTask` ignored the `timeout` parameter entirely (no `WithTimeout`/`WithDeadline` anywhere in the `task` package), so segment loads had no deadline at all. This PR makes it a hard bound: `ActivateDeadline` arms exactly once via `sync.Once` (`task.go:194`), and a single `context.WithTimeout(parentCtx, task.timeout)` (`task.go:201`) covers every remaining action's RPC plus the inter-action dist confirmation wait — `ex.cluster.LoadSegments(task.Context(), ...)` blocks for the full load (`task/executor.go:285`), and `checkActionFinish`/`waitingDist` (`task/scheduler.go:1174-1183`) consumes the same budget. Recovery from expiry has no backoff or retry cap.
Failure: a collection with sealed segments whose load time is consistently above `queryCoord.segmentTaskTimeout` (new default 180s) — e.g. near-max-size segments with DiskANN/quantized indexes pulled from object storage, or S3-side throttling — with nodes and replicas healthy → the deadline fires, `LoadSegments` is cancelled and the QueryNode discards everything downloaded so far, `task.Fail(err)` marks the task failed (`task/executor.go:239-241`), and `SegmentChecker` rebuilds the identical task with the identical timeout ~3s later (`checkers/segment_checker.go:530`, `checkSegmentInterval` default 3000ms at `pkg/util/paramtable/component_param.go:3206-3212`) — the segment never loads, load progress never reaches 100%, and every ~183s repeats the full download bandwidth and memory/disk reservation.
Suggestion: pick one — bound the rebuild path with backoff and a retry cap (the `GlobalFailedLoadCache` count is already available as a backoff input), or redefine `segmentTaskTimeout` as a no-progress timeout rather than a whole-task budget. At minimum, state explicitly in `configs/milvus.yaml` and the release note that collections with per-segment load times above `segmentTaskTimeout` must raise the value or they will retry indefinitely. *(raised by xiaocai2333, xaxys)*

**`internal/querynodev2/services.go:361` — `ChannelTaskTimeout` is now a hard bound on `WatchDmChannels`; slow-recovering channels may never get a delegator.**
Same root cause as above, with a heavier consequence. ChannelTask ctx now carries `ChannelTaskTimeout` (`task/executor.go:414`, `490`), default raised from 60s to 120s (`configs/milvus.yaml:504`). On the QueryNode side, `WatchDmChannels` serially performs `loadL0Segments`, `loadGrowingSegments`, and `pipeline.ConsumeMsgStream(seek)` (`services.go:314-329`, `345`), whose cost scales with unflushed backlog, L0 deltalog count, and MQ seek distance — none of which relate to the 120s constant. The new `if err = ctx.Err(); err != nil` guard at `services.go:361` is itself correct (it avoids committing a watch the coordinator has already failed), but assigning `err` triggers the upstream cleanup defers that close the delegator, remove it from `node.delegators`, drop the pipeline, and discard growing segments (`services.go:269-275`).
Failure: a dm channel whose recovery consistently exceeds 120s — large unflushed growing segments plus L0 deltalogs, or a far-behind `seekPosition` — on a healthy node → all three steps complete, the guard then discards the entire round, `ChannelChecker` rebuilds the same task with the same 120s (`checkers/channel_checker.go:298`), and the loop never converges: the shard has no delegator, search/query on it stay unavailable, and each round repeats the full L0/growing download and MQ seek.
Suggestion: address together with the item above (backoff/cap, or no-progress timeout semantics), and document that channels with large backlogs need a larger `queryCoord.channelTaskTimeout`. *(surfaced during verification)*

czs007:

Enabling `segmentTaskTimeout` / `channelTaskTimeout` for the first time makes a fixed, load-independent deadline authoritative over the whole task lifetime, which can cancel legitimately slow loads and leave them to be retried without backoff.

## High

**`internal/querycoordv2/task/task.go:190`** — `ActivateDeadline()` installs a single `sync.Once` `context.WithTimeout(parentCtx, task.timeout)` that covers both the `LoadSegments` RPC (`executor.go:284`) and the subsequent distribution-convergence wait (`checkActionFinish` requires `rpcReturned` **and** `isDistMatched`, `scheduler.go:1178-1183`), with a fixed 180s default (`configs/milvus.yaml:505`) that is unrelated to segment size, index type, or object-storage bandwidth. Before this PR `newBaseTask` only did `context.WithCancel` and silently dropped the timeout, so the cancellation path at `scheduler.go:1211` was unreachable. There is also no timeout-specific backoff: `remove()` only penalizes `merr.ErrSegmentRequestResourceFailed` (`scheduler.go:1229-1240`), and `checkStale` has no task-age or failure-count criterion. Failure: a sealed segment with a large DiskANN/HNSW index (or an S3-throttled load) that legitimately needs ~240s → the task is canceled at 180s, the segment is still missing from the distribution, and `SegmentChecker` rebuilds the identical task every round (`segment_checker.go:132-138` only refreshes the version cache when no task was produced) — an unbounded retry loop in which the segment never finishes loading, each failure writes a 24-hour-non-expiring entry into `GlobalFailedLoadCache` (`scheduler.go:1246`, `failed_load_cache.go:28`) that surfaces to users as `CollectionNotLoaded` + `context deadline exceeded`, while the abandoned, non-cancellable C++ index load keeps holding memory and bandwidth and stacked retries can drive the node toward OOM. Suggestion: size the budget to the load (or replace it with a progress-based watchdog) and add backoff for repeated deadline failures, instead of applying one fixed cap to "RPC + distribution convergence" as a whole. *(raised by chyezh, xaxys)*

**`internal/querynodev2/services.go:360`** — Channel tasks get the same fixed budget (120s default, `configs/milvus.yaml:504`), propagated from `task.Context()` (`executor.go:416`) through `WatchDmChannels` (`executor.go:490`) to the QueryNode, which must serially complete `loadL0Segments` (`services.go:315`), `loadGrowingSegments` (`services.go:320`), and `pipeline.ConsumeMsgStream` seek (`services.go:346`) — all of whose cost depends on L0 count, growing data volume, and WAL replay distance, not on the 120s constant. The new pre-commit check at `services.go:359-364` returns before `pipeline.Start()` / `delegator.Start()` and assigns to the named `err`, triggering the full rollback defer chain. Failure: recovering a dml channel with a large L0 backlog and a far-behind seek position, where L0 + growing + seek total ~150s → the watch is fully rolled back (delegator, pipeline, growing segments all dropped), the coord task is canceled, and `ChannelChecker` regenerates the identical task with no backoff (`channel_checker.go:118-124`, `:291`), so the shard stays permanently unavailable while the expensive L0/growing load is re-run every round. Suggestion: derive the budget from recovery progress (L0 segments consumed / seek catch-up rate) or make it recovery-aware, and at minimum back off repeated deadline failures. *(raised by chyezh, xaxys)*

## Medium

**`internal/querycoordv2/task/task.go:194`** — `deadlineOnce` arms the deadline only on the first dispatch, but `Executor.Execute` is invoked per step, so a move task's two actions (`[grow on To, reduce on From]`, `balance/utils.go:105-113`) share one budget across step 0, the inter-step wait, and step 1 rather than getting one each. Channel moves are hit hardest: `preProcess` explicitly blocks the reduce until the new node becomes the shard leader in `ChannelDistManager` (`scheduler.go:1134-1158`), and that wait falls entirely inside the same deadline. Failure: on a busy cluster, a channel balance where grow takes ~90s and leader promotion another ~40s → the deadline expires before the reduce is issued, `scheduler.check` cancels the task, and the already-applied grow is not rolled back, so source and target nodes both stay subscribed and the replica carries two delegators for the same channel (duplicated pipeline plus duplicated growing/L0 memory) until `findRepeatedChannels` (`channel_checker.go:236-256`) releases the lower-version one in a later round; segment moves duplicate the segment on two nodes in the same way. Self-healing, but memory doubles in the interim and balance re-plans — precisely on the busy clusters balancing is meant to relieve. Suggestion: reset the deadline per step, or scope it to the actual action RPC rather than to inter-step distribution/promotion waits. *(surfaced during verification)*

xiaofan-luan:
Addressed in `013c3648bc`: bumped `queryCoord.segmentTaskTimeout` default from 3min to 5min (`channelTaskTimeout` stays at 2min — channel watch is a different, generally faster operation).

5 minutes is judged enough headroom for a single segment to finish loading under normal conditions, including large/disk-index segments and moderate object-storage throttling. We don't think defaulting the timeout to disabled (0) for segment/leader tasks is the right tradeoff here — that would leave the original stuck-task bug this PR fixes unaddressed by default for the two task types where it originally showed up.

The deeper point in the P1 finding — a flat wall-clock budget can't distinguish "genuinely stuck" from "legitimately slow," and repeated `DeadlineExceeded` rebuilds have no backoff — is correct and not fixed by this bump. We agree the real fix is a heartbeat/progress-reporting mechanism from the QueryNode so the deadline can extend on observed forward progress instead of firing on a fixed clock. That's a separate, substantially larger effort (new RPC, load-loop instrumentation, rolling-upgrade compatibility) and we're treating it as a follow-up rather than folding it into this PR. Tracked via the TODOs already in `task.go`'s `ActivateDeadline`, `SegmentChecker.createSegmentLoadTasks`, and `ChannelChecker`'s equivalent.

xiaofan-luan:
Summary of how the four review passes (`D1af54aef4f`, `D07501ca39d`, `D02e4395801`, `D5b4ec0e2e4`) were addressed, for anyone tracking this thread — current head is `013c3648bc`:

**Fixed:**
- Deadline counted from task construction (queueing time included) → `ActivateDeadline` now arms on first real dispatch, after admission (`eb42d3c53a`).
- Executor admission-layer rejection burning the deadline before the first RPC → same fix above, since arming now happens post-admission.
- `task.Wait()`'s sync LoadBalance timeout racing the task's own deadline / masking its real result → replaced with a bare ctx+task-completion wait (`be4e1e4408`), **then a real regression we introduced ourselves**: without any bound, a task that's never admitted (sustained executor saturation) could hang the handler forever, since callers like pymilvus default to no client-side timeout. Fixed by restoring an explicit `context.WithTimeout(ctx, SegmentTaskTimeout/ChannelTaskTimeout)` at the `handlers.go` call sites (`3a4a25d446`) — matching master's original bound, not a guess at matching the task's own deadline.
- Move task (`[grow, reduce]`) sharing one deadline across both actions plus the inter-action dist-confirmation/leader-promotion wait, so a slow first action could exhaust the budget before the second was ever issued → `ActivateDeadline` now takes a step number and arms a fresh full budget per step (`5e727d0034`), so a later action no longer inherits whatever was left of an earlier one.
- `TestChannelTaskTimeoutBoundsTaskContext` asserting a deadline at construction time, contradicting the dispatch-time-activation fix → updated same session, before `D07501ca39d` was posted against that intermediate commit.
- PR body claiming "bounds the whole task including queueing" and "both timeouts unified to 2min" → corrected to match current behavior and defaults.
- Missing test coverage for the production wiring (Executor.Execute → ActivateDeadline) and for admission rejection → added `TestExecutorActivatesDeadlineOnlyAfterAdmission` and `TestActivateDeadlinePerStepGetsFreshBudget`, both exercised against the real `Executor.Execute()` path rather than calling `ActivateDeadline` directly.

**Partially addressed:**
- Missing coverage for `WatchDmChannels`'s cancel-before-commit rollback → `TestWatchDmChannels_CanceledContextRollsBack` added, but could not be run in this environment (a pre-existing, unrelated cgo/header mismatch keeps `querynodev2` from building locally here) — verified by syntax check and by pattern-matching the existing `TestWatchDmChannels_Failed` test in the same suite; needs CI to confirm.
- Missing coverage for the singleflight-joiner-cancellation and deltalog-downloader-cancellation paths → not added. Same local build blocker, and both need real concurrent mocks (a blocking worker + racing callers sharing a singleflight key), which we judged too risky to write blind with zero ability to compile-check locally. Flagging as an open gap rather than shipping unverified.

**Accepted as a known tradeoff, not fixed here (the core of `D1af54aef4f`'s remaining High, `D02e4395801`'s two High items, and `D5b4ec0e2e4`):**
A flat wall-clock deadline can't distinguish "genuinely stuck" from "legitimately slow," and `SegmentChecker`/`ChannelChecker` rebuild a timed-out task with the identical budget every check tick, with no backoff or retry cap — a segment/channel whose real op time is consistently above the configured timeout will never converge. This is real and we're not disputing it. Rather than default the timeout to disabled (0) for segment/leader until a proper fix exists — which would leave the original stuck-task bug this PR fixes unaddressed by default for the two task types where it originally showed up — we bumped `segmentTaskTimeout`'s default from 3min to 5min (`013c3648bc`), judged enough headroom for a single segment's load under normal conditions. `channelTaskTimeout` stays at 2min. The real fix is a heartbeat/progress-reporting mechanism from the QueryNode so the deadline can extend on observed forward progress instead of firing on a fixed clock; that's a separate, substantially larger effort (new RPC, load-loop instrumentation, rolling-upgrade compatibility) tracked as a follow-up via the TODOs in `task.go`'s `ActivateDeadline`, `SegmentChecker.createSegmentLoadTasks`, and `ChannelChecker`'s equivalent, not folded into this PR.

czs007:

Verified:
- `internal/querycoordv2/task/task.go:203-244` — `ActivateDeadline` early-returns on `timeout<=0` / canceled / same step; new ctx derives from `rootCtx`; `Cancel`/`Fail` read `rootCancel` under lock and invoke it outside the lock, no deadlock; CAS guarantees single execution.
- `internal/querycoordv2/task/executor.go:149-185` — `ActivateDeadline(step)` runs after `executingTasks.Insert` and both cap gates, before the goroutine; `loadSegment` defer order (`removeTask` → `rpcReturned=true`) ensures `stepCancel()` never interrupts an in-flight RPC.
- `internal/querycoordv2/task/scheduler.go:1210-1216` — `DeadlineExceeded` → `Cancel(err)` (state Canceled), does not enter `recordSegmentTaskError`; the `ErrSegmentRequestResourceFailed` check in `remove()` is not falsely triggered.
- `internal/querycoordv2/task/utils.go:62-80` — `Wait` uses a buffered(1) `errCh`; if ctx fires first, the goroutine still writes and exits, no leak.
- `internal/querynodev2/delegator/delegator_data.go:677-687` — `DoChan` channel is buffered(1), abandoning the wait does not leak; joiner semantics match existing `waitSegmentLoadDone` behavior.
- `internal/querynodev2/services.go:357-364` — commit-point `ctx.Err()` check triggers delegator/pipeline/growing rollback; an un-`Start()`ed delegator is filtered from `GetDataDistribution` (`services.go:1359-1361`), so QC will not misjudge the watch as successful.

## Low (non-blocking)

- `internal/querynodev2/services.go:210` — A concurrent `WatchDmChannels` that hits `subscribingChannels` returns `merr.Success()` without performing any subscription. With this PR, task deadlines now actually fire and the new commit-point check (`services.go:357-364`) rolls back the timed-out predecessor's watch, so the successor's Success is now a no-op. Failure: ChannelTask A's WatchDmChannels is canceled by its deadline but the QN handler is still draining (e.g. stuck in a non-cancelable seek/download), and the checker rebuilds task B on the same node while the channel is still in `subscribingChannels` → every resend of B gets `merr.Success()` with no subscription, B's 2min budget is consumed by A's drain time; if the drain is ≥ 2min, B is killed on DeadlineExceeded and C is rebuilt, delaying channel availability by a full timeout+check cycle (no data-correctness impact). Suggestion: return a retryable error (e.g. `merr.WrapErrChannelReduplicate` / `ServiceUnavailable`) when hitting `subscribingChannels`, or otherwise avoid letting successor tasks burn budget on a false Success. (raised by czs007)
- `internal/querycoordv2/task/task_test.go:3386` — `TestActivateDeadlinePerStepGetsFreshBudget` uses an 80ms timeout with two `time.Sleep(60 * time.Millisecond)` calls followed immediately by `assert.NoError(t, task.Context().Err())`, leaving only 20ms of slack. Failure: on a loaded CI machine `time.Sleep(60ms)` wakes later than 80ms (GC/scheduling delay > 20ms) → the asserts at :3387 or :3402 report `context deadline exceeded` and the test fails intermittently. Suggestion: enlarge the timeouts (e.g. 500ms/300ms) or compare deadlines instead of actually sleeping. (raised by czs007)

## Nitpick (non-blocking)

- `internal/querycoordv2/checkers/segment_checker.go:531` — The comment here and at `task.go:221` still say the segment task timeout defaults to 3min, but `component_param.go:3255-3257` sets `SegmentTaskTimeout` `DefaultValue: "300000"` with `Doc: "5 minutes"`. Suggestion: update both comments to 5min. (raised by czs007)

- `internal/querynodev2/delegator/delegator_data.go:656` — The `LoadSegments` pin is not reference-counted (`listDeleteBuffer.Pin` uses set semantics: `pinnedTimestamps[ts][segmentID] = struct{}{}`), so of two concurrent `LoadSegments` callers on the same `(nodeID, segmentID)`, whichever returns first unpins for both. Failure: two concurrent `delegator.LoadSegments` calls for the same `(DstNodeID, segmentID)` and same `segmentEffectiveTs` (e.g. SyncDistribution Set overlapping a SegmentTask Grow retry in a narrow window), one returning early → its deferred Unpin removes the shared pin, `SyncTargetVersion` can clean delete records after that ts, and the later caller's `loadStreamDelete` misses them, leaving deleted rows visible on that segment (hard to reach under current QC dedup). Unrelated to this PR because the diff only changes `delegator_data.go:674-687` (`sf.Do` → `sf.DoChan`+`select`); the Pin/defer Unpin and the set-based delete-buffer Pin/Unpin are untouched and the same first-returner-unpins behavior exists on the baseline; this PR only lets a joiner exit earlier and adds no new concurrent caller. No tracking issue yet. (raised by chyezh, czs007)

xiaofan-luan:
Fixed in `16f34ddde3`:
- Widened `TestActivateDeadlinePerStepGetsFreshBudget`'s timing (80ms/60ms → 500ms/300ms) to remove the flake risk on a loaded CI machine.
- Fixed the two stale "3min" comments (`task.go`'s `ActivateDeadline` doc, `segment_checker.go`'s `createSegmentLoadTasks`) to say 5min, matching the `013c3648bc` default bump.

Not changing `services.go:210`'s `subscribingChannels` → `merr.Success()` on purpose: that return value is `WatchDmChannels`'s idempotency contract, not an oversight — a concurrent/duplicate watch for a channel already being subscribed is supposed to look like success to its caller, the same way calling an idempotent RPC twice shouldn't surface an error just because the first call is still in flight. Turning it into a retryable error would break that contract for every caller, not just the specific interleaving this finding describes (predecessor stuck draining on an uncancelable seek/download).

The failure mode itself is real, but it's downstream of the pre-existing hazard this same review flagged separately (non-cancelable QueryNode-side work outstanding after the coordinator has already given up on it) rather than something to special-case in the idempotency check. Worth its own follow-up once there's an actual mechanism to interrupt or detect that stuck drain, rather than papering over it here by making a correct idempotent response non-idempotent for everyone.

czs007:

**Verified:**
- `internal/querycoordv2/task/executor.go:184` — `ActivateDeadline(step)` is armed only after both admission gates pass and before the goroutine is spawned, so queue wait does not consume RPC budget and per-step budgets do not bleed into each other.
- `internal/querycoordv2/task/task.go:225-249` — per-step deadline switching is done under `ctxMu`; `Cancel`/`Fail` read `rootCancel` under the lock and invoke it outside, and `canceled.CompareAndSwap` guards `span.End` so there is no re-entrant deadlock or double-end.
- `internal/querynodev2/services.go:231-372` — the commit-point `ctx.Err()` check reuses the function-level `err`, so all rollback defers (collection, delegator, pipeline, growing segments) fire and `Unref` takes effect; tests assert no delegator is left behind.
- `internal/querycoordv2/utils.go:62-80` — the rewritten `Wait` uses a buffered `errCh` (size 1), so the background goroutine cannot leak when `ctx.Done` fires first, and returning `ctx.Err()` on timeout fixes the prior racy read of `err`.
- `internal/querynodev2/delegator/delegator_data.go:674-778` — the singleflight `DoChan` path has no goroutine leak (internal channel is buffered), and post-processing runs only after the worker load succeeds.
- `internal/storage/rw.go:521-643`, `serde_delta.go:691-699` — V1 downloads receive the caller context; V2/V3 check cancellation at file boundaries, with the non-interruptible single-file FFI documented in code.

## Re-review of `013c3648bc42`

- **Confirmed** — `internal/querynodev2/services.go:210`: the false `Success` on a concurrent `WatchDmChannels` still stands; see the updated finding below for the interaction with the new commit-point rollback.
- **Confirmed** — `internal/querynodev2/delegator/delegator_data.go:656`: `Pin`/`Unpin` remain set-semantics rather than reference-counted; see the updated finding below for the new early-return path.
- **Retracted** — `internal/querycoordv2/task/task_test.go:3386`: our earlier comment does not hold against the current head. `TestActivateDeadlinePerStepGetsFreshBudget` now uses a 500ms timeout with `time.Sleep(300 * time.Millisecond)` before each `assert.NoError`, leaving ~200ms of margin; the 80ms/60ms values we cited only appear in `TestExecuteArmsDeadlineOnlyOnAdmission`, which has no sleep-then-assert.
- **Retracted** — `internal/querycoordv2/checkers/segment_checker.go:531`: our earlier comment does not hold against the current head. The comment now reads `SegmentTaskTimeout (5min default)` and the `ActivateDeadline` TODO in `task.go` reads `defaults to 5min`; no `3min` occurrences remain in the PR diff.

## Low — non-blocking

- `internal/querynodev2/services.go:210` — When a watch for a channel is still in flight (or rolling back) on the QueryNode, a second `WatchDmChannels` for the same channel hits `!node.subscribingChannels.Insert(...)` and returns `merr.Success()` even though the node is not subscribed. This PR makes the window reachable: `executor.go:184` now arms a per-step deadline on the RPC, so QC removes task A on `ctx.Err()` (`scheduler.go:1210-1216`) and dispatches B while A is still executing rollback (including the new commit-point check at `services.go:361-364`) before its `defer subscribingChannels.Remove` runs. Failure: a channel's watch exceeds `ChannelTaskTimeout` in `loadL0Segments`/`loadGrowingSegments`/`ConsumeMsgStream`, QC re-dispatches B while A is rolling back → B returns Success but no delegator/pipeline exists; since `ChannelAction.IsFinished` (`action.go:220-228`) is dist-based, B idles until its own deadline and the shard's unavailability is extended by one more `ChannelTaskTimeout`. Self-healing, but worth tightening. Suggestion: return a retryable error at line 210 (e.g. `WrapErrChannelReduplicate`, consistent with lines 219-223) instead of Success, or track as a follow-up. *(raised by xiaocai2333, xaxys)*

- `internal/querynodev2/delegator/delegator_data.go:656` — `Pin`/`Unpin` on the delete buffer are set-semantics (`list_delete_buffer.go:240-266`, `map[ts]map[segmentID]struct{}`), so two concurrent `LoadSegments` for the same `(ts, segmentID)` share a single pin entry and the first `Unpin` removes it. This PR replaces `sf.Do` with `DoChan` + `case <-ctx.Done(): return ctx.Err()` (lines 680-686), so a joiner can now return early on its own context cancellation and run its deferred `Unpin` while the initiator's `worker.LoadSegments` is still in progress — previously any `Unpin` could only happen after the shared load completed. Failure: two independent-context loads for the same `(dstNode, segmentID)` (requires QC failover or external concurrent callers; the single-QC scheduler dedups by `[replica][segment]`), the joiner's ctx cancels first, and `SyncTargetVersion` (lines 1316-1334) calls `UnRegister` with a deleteCP past the segment's effectiveTs → the delete records the initiator still needs are cleared, `loadStreamDelete`/`ListAfter` misses them, and deleted rows become visible again on that segment. Suggestion: make pin reference-counted, or have the joiner skip `Unpin` on early return (leaving it to the initiator); a follow-up is acceptable given the narrow reachability. *(raised by xiaocai2333, xaxys)*

AyushKashyapII:
Hey @xiaofan-luan , first off really solid work on the ActivateDeadline refactor. the per step timeout design is genuinely clever: not starting the clock until the task actually clears the executor's admission gate is a much better model than the old "timeout starts at creation" approach, and the idempotency guard on re-arming the same step is a nice touch. The span leak fix in Fail() is also good to see closed out.

Had a few small things i wanted to inquire abt:

the error message in NewChannelTask (L526) — the validation error says "all actions must operate the same segment" inside a channel task. Looks like a leftover copy-paste from NewSegmentTask. Pretty minor but could be confusing when debugging.

Goroutine leak in the task.Wait wrapper (utils.go L62–81) — when the caller's context expires and Wait returns early via ctx.Done(), the inner goroutine is still blocked on task.Wait() and leaks until the task eventually finishes (or never, if it's stuck in queue with no timeout). Probably low impact in practice but worth a thought.

also TaskStatusCreated = "created" constant (line 46) is never actually assigned anywhere. It exists only as a sentinel but never used

xiaofan-luan:
Thanks @AyushKashyapII — I went through all three against the merged tree, and all three hold up. None of them were introduced by this PR, but they're all worth cleaning up, so I've put them in a follow-up: #52688.

**1. `NewChannelTask` error message** — confirmed, `task.go:526`. It compares `channel` vs `channelAction.ChannelName()` but reports `"all actions must operate the same segment"`. `git log -S` traces it to #22745 (2023-03), copy-pasted from `NewSegmentTask`. Fixed to `"... the same channel"`.

**2. `Wait` goroutine** — confirmed, and it's the one with real teeth, so here's the full chain for whoever reads this later. The question is whether the parked goroutine has *any* upper bound, and it doesn't:

- `Executor.Execute` (`executor.go:149`) rejects a step when the channel / non-channel admission cap is full and returns **before** reaching `task.ActivateDeadline(step)` — so a task that keeps losing at admission never arms a deadline at all. That's by design in this PR (queueing time shouldn't burn the RPC budget), but it does mean a queued task carries no lifetime guarantee.
- `taskScheduler.checkStale` (`scheduler.go:1372`) only checks replica presence, node presence, stopping state and RO membership. There is no task-age / time-based criterion anywhere in it.

So on a healthy cluster whose executor stays saturated, a queued task has no upper bound on its lifetime, and neither does anything parked on it. Live callers are the two LoadBalance paths in `handlers.go` (`balanceSegments`, `balanceChannels`), so the leak rate is bounded by how often that admin RPC is called — low severity, but genuinely unbounded.

Worth noting the shape predates this PR: master's version wrote `err` from the goroutine with no synchronization and read it after `<-ctx.Done()`, so it had the same parked goroutine *plus* a data race and a false-`nil` success path. This PR fixed the race; the goroutine survived.

The fix in #52688 removes the goroutine rather than bounding it — `Task.Wait` now takes a ctx and selects on it, so the package-level `Wait` just walks the tasks on the caller's own goroutine. `Task.Wait()` had exactly one caller and the interface has no mocks, so the signature change is contained. It also checks for an already-finished task non-blockingly first: `select` picks uniformly at random among ready cases, so without that a task finishing at the same moment ctx expires would report `DeadlineExceeded` to LoadBalance about half the time instead of the real error.

Regression test asserts the goroutine count returns to baseline after 64 `Wait` calls against tasks that never finish — before the fix it's exactly 64 leaked, after it settles immediately.

**3. `TaskStatusCreated`** — confirmed. `git grep` across the tree finds only the definition; tasks are constructed at `TaskStatusStarted` (`task.go:187`) and the other four constants all have real assignments or comparisons. Removed.

Appreciate the careful read.

PR #52481, opened 2026-08-13
fix: propagate caller context into legacy deltalog downloads

xiaofan-luan:
issue: #52257 related (follow-up of #52478 cancellation audit)

`NewLegacyDeltalogReader` invoked its downloader with `context.Background()`, severing the caller's context. On QueryNode this makes the L0/deltalog download during segment load and channel watch immune to gRPC deadline cancellation: when the coordinator task times out (#52478), the handler goroutine keeps downloading deltalogs until the object-storage client gives up on its own.

Thread `ctx` through `NewDeltalogReader` / `NewDeltalogReaderFromBinlogs` / `NewLegacyDeltalogReader` so the eager download in the StorageV1 path honors the request context. Mechanical signature change; all call sites already have a context in scope.

### DIFF
--- a/internal/compaction/common.go
+++ b/internal/compaction/common.go
@@ -93,7 +93,7 @@ func readDeltalogsV1(
 
 	for _, deltalog := range deltalogs {
 		for _, binlog := range deltalog.Binlogs {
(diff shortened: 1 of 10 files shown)


xiaofan-luan:
Folded into #52478 (same cancellation-audit line of work).

PR #52482, opened 2026-08-13
fix: do not commit a canceled watch channel request

xiaofan-luan:
issue: #52257 related (follow-up of #52478 cancellation audit)

`WatchDmChannels` never re-checks its context between phases: a request whose caller already timed out (channel task timeout, #52478) could still start the pipeline and delegator, leaving the node subscribed while the coordinator has marked the task failed and regenerated it. The re-issued watch then races the stale dispatcher registration (`vchannel already exists`).

Check `ctx.Err()` once at the commit point — before `pipeline.Start()` / `delegator.Start()` — so a canceled request rolls back through the existing deferred cleanup chain (delegator removal, pipeline/dispatcher deregistration, collection unref, growing cleanup) instead of committing a watch nobody wants.

### DIFF
--- a/internal/querynodev2/services.go
+++ b/internal/querynodev2/services.go
@@ -354,6 +354,15 @@ func (node *QueryNode) WatchDmChannels(ctx context.Context, req *querypb.WatchDm
 		return merr.Status(err), nil
 	}
 
+	// The caller may have timed out during the load/seek steps above. Commit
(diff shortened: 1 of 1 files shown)


xiaofan-luan:
Folded into #52478 (same cancellation-audit line of work).

Issue #52159, opened 2026-08-04
[Bug]: BM25 backfill panics when its VARCHAR input field was added after sealed data, causing a restart crash loop


shanghaikid:
## Environment

- **Milvus Version**: `milvusdb/milvus:v3.0.0-arm64`
- **Milvus GitCommit**: `f46a032855`
- **Deployment Mode**: standalone
- **MQ**: rocksmq
- **SDK**: `@zilliz/milvus2-sdk-node@3.0.4`
- **OS**: macOS 14.2.1 arm64, Docker
- **Container resources**: 1.5 CPU, 2 GiB memory
- **Container termination**: exit code 134, `OOMKilled=false`

### Option B: Steps

1. Run standalone Milvus with StorageV3 schema backfill enabled:

   ```yaml
   common:
     storage:
       useLoonFFI: true
   dataCoord:
     compaction:
       storageVersion:
         enabled: true
       bumpSchemaVersion:
         enabled: true
   ```

2. Create a collection with an Int64 primary key and a FloatVector field.
3. Insert 100 rows and flush them into one sealed StorageV3 segment.
4. Add a new top-level nullable VarChar field after the segment already exists:

   ```text
   name: text
   data_type: VarChar
   max_length: 256
   nullable: true
   enable_analyzer: true
   enable_match: true
   analyzer_params: {"type":"standard"}
   ```

   Historical rows therefore have no physical column for this newly added field.

5. Wait for schema backfill to report complete. In the observed run, schema version 3 reached `consistentSegments=1/totalSegments=1`.

6. Add a BM25 function field using the newly added VarChar as input:

   ```text
   function:
     name: text_bm25_fn
     type: BM25
     input_field_names: ["text"]
     output_field_names: ["text_sparse"]

   output field:
     name: text_sparse
     data_type: SparseFloatVector
     is_function_output: true

   bound index:
     index_type: SPARSE_INVERTED_INDEX
     metric_type: BM25
   ```

7. Wait for the next `BumpSchemaVersionCompaction`.

### Trigger Conditions

- **Frequency**: deterministic for the observed segment; the same persisted plan panics on every restart.
- **First observed after**: the BM25 function-output backfill begins.
- **Does NOT happen when**: the BM25 input VarChar already existed when historical rows were written (the existing function-field backfill E2E covers that case).
- The field-add schema bump had completed about 9 minutes before BM25 was added, so this is not caused by immediately chaining two DDL operations.

## Expected Behavior

[...]

PR #49759, opened 2026-05-13
enhance: use bump_schema_version to involve backfill and drop field compaction(#48808)

MrPresent-Han:
related: https://github.com/milvus-io/milvus/issues/48808

PR #52484, opened 2026-08-13
fix: reconcile physical fields during schema bump (#52159)

MrPresent-Han:
issue: #52159

## Root cause

Schema-bump compaction derived its read shape from physically existing fields but later accessed the target schema directly. When a function input was an ordinary field added after existing sealed data, the input column remained absent and function materialization could panic.

## Summary

- Treat manifest-derived field IDs as the source of physical truth and route dropped, additive, and metadata-only reconciliation separately.
- Prefill missing nullable ordinary fields with configured defaults or NULLs before materializing missing function outputs.
- Reject inconsistent partial presence for multi-output functions and publish ordinary fields plus missing outputs through one successor manifest.
- Apply ordinary-first materialization to shared compaction readers.

This prevents schema-bump compaction from panicking when a newly added function depends on an ordinary field that is not yet present in segment storage.

### DIFF
--- a/internal/datanode/compactor/bump_schema_version_compactor.go
+++ b/internal/datanode/compactor/bump_schema_version_compactor.go
@@ -42,7 +42,6 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/proto/indexpb"
 	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
-	"github.com/milvus-io/milvus/pkg/v3/util/metautil"
 	"github.com/milvus-io/milvus/pkg/v3/util/paramtable"
 	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
@@ -63,6 +62,25 @@ type bumpSchemaVersionCompactionTask struct {
 	lobContext *compaction.LOBCompactionContext
 }
 
+// schemaBumpPhysicalDiff describes the difference between the target schema and
+// the columns physically present in the source manifest. existingFields always
+// comes from the manifest; target-schema membership must never be used as a
+// substitute for physical presence.
+type schemaBumpPhysicalDiff struct {
+	existingFields       map[int64]struct{}
+	droppedFieldIDs      []int64
+	absentOrdinaryFields []*schemapb.FieldSchema
+	missingOutputFields  []*schemapb.FieldSchema
+	missingFunctions     []*schemapb.FunctionSchema
+}
+
+func (d *schemaBumpPhysicalDiff) appendedFields() []*schemapb.FieldSchema {
+	fields := make([]*schemapb.FieldSchema, 0, len(d.absentOrdinaryFields)+len(d.missingOutputFields))
+	fields = append(fields, d.absentOrdinaryFields...)
+	fields = append(fields, d.missingOutputFields...)
+	return fields
+}
+
 func (t *bumpSchemaVersionCompactionTask) Compact() (*datapb.CompactionPlanResult, error) {
 	if !funcutil.CheckCtxValid(t.ctx) {
 		return nil, t.ctx.Err()
@@ -80,7 +98,7 @@ func (t *bumpSchemaVersionCompactionTask) Compact() (*datapb.CompactionPlanResul
 		mlog.FieldCollectionID(t.GetCollection()),
 	)
 
-	missingFunctions, droppedFieldIDs, existingFields, err := t.schemaBumpDecision()
+	diff, err := t.schemaBumpDecision()
 	if err != nil {
 		log.Warn(ctx, "failed to decide schema bump action", mlog.Err(err))
 		return nil, err
@@ -89,17 +107,22 @@ func (t *bumpSchemaVersionCompactionTask) Compact() (*datapb.CompactionPlanResul
 	log.Info(ctx, "schema bump compact start",
 		mlog.FieldSegmentID(t.plan.GetSegmentBinlogs()[0].GetSegmentID()),
 		mlog.Int32("collectionSchemaVersion", t.plan.GetSchema().GetVersion()),
-		mlog.Int("missingFunctionCount", len(missingFunctions)),
-		mlog.Int64s("droppedFieldIDs", droppedFieldIDs),
+		mlog.Int("missingFunctionCount", len(diff.missingFunctions)),
+		mlog.Int("absentOrdinaryFieldCount", len(diff.absentOrdinaryFields)),
+		mlog.Int("missingOutputFieldCount", len(diff.missingOutputFields)),
+		mlog.Int64s("droppedFieldIDs", diff.droppedFieldIDs),
 	)
 
 	var result *datapb.CompactionPlanResult
-	if len(missingFunctions) == 0 && len(droppedFieldIDs) == 0 {
+	// Dropped physical fields always require a replacement rewrite. Zero-row
+	// segments still route to additive reconciliation, which rejects them as a
+	// data-integrity error rather than writing an empty materialized record.
+	if len(diff.droppedFieldIDs) > 0 {
+		result, err = t.runFullSchemaRewrite(diff.existingFields)
+	} else if len(diff.absentOrdinaryFields) == 0 && len(diff.missingOutputFields) == 0 {
 		result = t.runSchemaVersionBumpOnly()
-	} else if len(droppedFieldIDs) > 0 {
-		result, err = t.runFullSchemaRewrite(existingFields)
 	} else {
-		result, err = t.runMissingFunctionMaterialization(ctx, missingFunctions, droppedFieldIDs, existingFields)
+		result, err = t.runAdditivePhysicalReconciliation(ctx, diff)
 	}
 	if err != nil {
 		log.Warn(ctx, "schema bump compact failed", mlog.Err(err), mlog.Duration("compact cost", time.Since(compactStart)))
@@ -145,29 +168,20 @@ func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunc
 		fields = append(fields, field)
 		fieldIDs = append(fieldIDs, fieldID)
 	}
+	// Inputs were validated by schemaBumpDecision (missingFunctionMaterializations);
+	// this only assembles the read schema.
 	for _, functionSchema := range missingFunctions {
-		if err := validateSupportedMissingFunctionMaterialization(functionSchema); err != nil {
-			return nil, nil, err
-		}
 		for _, inputFieldID := range functionSchema.GetInputFieldIds() {
 			inputField := typeutil.GetField(schema, inputFieldID)
 			if inputField == nil {
-				return nil, nil, merr.WrapErrParameterInvalidMsg("input field not found in schema")
-			}
-			if err := validateMaterializationInputField(functionSchema, inputField); err != nil {
-				return nil, nil, err
+				return nil, nil, merr.WrapErrDataIntegrityMsg(
+					"function %s input field %d not found in persisted schema",
+					functionSchema.GetName(), inputFieldID)
 			}
 			addInputField(inputField)
 
-			additionalFields, err := additionalFunctionInputFields(schema, functionSchema, inputField)
-			if err != nil {
+			if err := addBM25MultiAnalyzerInputFields(schema, functionSchema, inputField, addInputField); err != nil {
 				return nil, nil, err
-			}
-			for _, additionalField := range additionalFields {
-				if err := validateMaterializationInputField(functionSchema, additionalField); err != nil {
-					return nil, nil, err
-				}
-				addInputField(additionalField)
 			}
 		}
 	}
@@ -180,13 +194,23 @@ func (t *bumpSchemaVersionCompactionTask) missingFunctionInputSchema(missingFunc
 	}, fieldIDs, nil
 }
 
-func additionalFunctionInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
-	switch functionSchema.GetType() {
-	case schemapb.FunctionType_BM25:
-		return bm25AdditionalInputFields(schema, functionSchema, inputField)
-	default:
-		return nil, nil
+// addBM25MultiAnalyzerInputFields adds the by_field analyzer-selector column of
+// a BM25 multi-analyzer input to the physical read schema; the selector is the
+// runner's per-row analyzer name, not a function text input, so it is resolved
+// and VarChar-validated in bm25AdditionalInputFields instead of the generic
+// input validator. No-op for other function types.
+func addBM25MultiAnalyzerInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema, addInputField func(*schemapb.FieldSchema)) error {
+	if functionSchema.GetType() != schemapb.FunctionType_BM25 {
+		return nil
 	}
+	byFields, err := bm25AdditionalInputFields(schema, functionSchema, inputField)
+	if err != nil {
+		return err
+	}
+	for _, byField := range byFields {
+		addInputField(byField)
+	}
+	return nil
 }
 
 func bm25AdditionalInputFields(schema *schemapb.CollectionSchema, functionSchema *schemapb.FunctionSchema, inputField *schemapb.FieldSchema) ([]*schemapb.FieldSchema, error) {
@@ -220,65 +244,49 @@ func bm25AdditionalInputFields(schema *schemapb.CollectionSchema, functionSchema
 	return []*schemapb.FieldSchema{byField}, nil
 }
 
-func (t *bumpSchemaVersionCompactionTask) missingFunctionOutputFields(missingFunctions []*schemapb.FunctionSchema, existingFields map[int64]struct{}) ([]*schemapb.FieldSchema, []int64, error) {
-	schema := t.plan.GetSchema()
-	var fields []*schemapb.FieldSchema
-	var fieldIDs []int64
-	for _, functionSchema := range missingFunctions {
-		if err := validateSupportedMissingFunctionMaterialization(functionSchema); err != nil {
-			return nil, nil, err
-		}
-		for _, outputIndex := range functionOutputIndexesToMaterialize(functionSchema, existingFields) {
-			outputFieldID := functionSchema.GetOutputFieldIds()[outputIndex]
-			outputField := typeutil.GetField(schema, outputFieldID)
-			if outputField == nil {
-				return nil, nil, merr.WrapErrParameterInvalidMsg("output field not found in schema")
-			}
-			if err := validateMaterializationOutputField(functionSchema, outputField); err != nil {
-				return nil, nil, err
-			}
-			fields = append(fields, outputField)
-			fieldIDs = append(fieldIDs, outputFieldID)
-		}
-	}
-	if len(fields) == 0 {
-		return nil, nil, merr.WrapErrParameterInvalidMsg("no missing function output fields")
-	}
-	return fields, fieldIDs, nil
-}
-
 func validateSupportedMissingFunctionMaterialization(functionSchema *schemapb.FunctionSchema) error {
 	switch functionSchema.GetType() {
 	case schemapb.FunctionType_BM25:
 		if len(functionSchema.GetInputFieldIds()) == 0 {
-			return merr.WrapErrParameterInvalidMsg("bm25 function should have input fields")
+			return merr.WrapErrDataIntegrityMsg("persisted bm25 function %s has no input fields", functionSchema.GetName())
 		}
 		if len(functionSchema.GetOutputFieldIds()) == 0 {
-			return merr.WrapErrParameterInvalidMsg("bm25 function should have output fields")
+			return merr.WrapErrDataIntegrityMsg("persisted bm25 function %s has no output fields", functionSchema.GetName())
 		}
 		return nil
 	case schemapb.FunctionType_MinHash:
 		if len(functionSchema.GetInputFieldIds()) == 0 {
-			return merr.WrapErrParameterInvalidMsg("minhash function should have input fields")
+			return merr.WrapErrDataIntegrityMsg("persisted minhash function %s has no input fields", functionSchema.GetName())
 		}
 		if len(functionSchema.GetOutputFieldIds()) == 0 {
-			return merr.WrapErrParameterInvalidMsg("minhash function should have output fields")
+			return merr.WrapErrDataIntegrityMsg("persisted minhash function %s has no output fields", functionSchema.GetName())
 		}
 		return nil
 	default:
-		return merr.WrapErrParameterInvalidMsg("unsupported function type")
+		return merr.WrapErrDataIntegrityMsg(
+			"persisted function %s has unsupported materialization type %s",
+			functionSchema.GetName(), functionSchema.GetType())
 	}
 }
 
+// validateMaterializationInputField enforces the schema-bump runtime contract on
+// persisted function inputs. It is narrower than the collection-creation
+// validator: a sealed StorageV3 Text column is read back as encoded LOB
+// references, which stringInputsFromRecord cannot consume without LOB decoding,
+// so only VarChar is an acceptable persisted input here.
 func validateMaterializationInputField(functionSchema *schemapb.FunctionSchema, field *schemapb.FieldSchema) error {
 	switch functionSchema.GetType() {
 	case schemapb.FunctionType_BM25:
-		if field.GetDataType() != schemapb.DataType_VarChar && field.GetDataType() != schemapb.DataType_Text {
-			return merr.WrapErrParameterInvalidMsg("input field data type must be varchar or text for bm25 materialization")
+		if field.GetDataType() != schemapb.DataType_VarChar {
+			return merr.WrapErrDataIntegrityMsg(
+				"persisted bm25 input field %d must be VarChar for schema-bump materialization (Text would require LOB decoding), got %s",
+				field.GetFieldID(), field.GetDataType())
 		}
 	case schemapb.FunctionType_MinHash:
-		if field.GetDataType() != schemapb.DataType_VarChar && field.GetDataType() != schemapb.DataType_Text {
-			return merr.WrapErrParameterInvalidMsg("input field data type must be varchar or text for minhash materialization")
+		if field.GetDataType() != schemapb.DataType_VarChar {
+			return merr.WrapErrDataIntegrityMsg(
+				"persisted minhash input field %d must be VarChar for schema-bump materialization (Text would require LOB decoding), got %s",
+				field.GetFieldID(), field.GetDataType())
 		}
 	}
 	return nil
@@ -288,24 +296,18 @@ func validateMaterializationOutputField(functionSchema *schemapb.FunctionSchema,
 	switch functionSchema.GetType() {
 	case schemapb.FunctionType_BM25:
 		if field.GetDataType() != schemapb.DataType_SparseFloatVector {
-			return merr.WrapErrParameterInvalidMsg("output field data type must be sparse float vector for bm25 materialization")
+			return merr.WrapErrDataIntegrityMsg(
+				"persisted bm25 output field %d must be SparseFloatVector, got %s",
+				field.GetFieldID(), field.GetDataType())
 		}
 	case schemapb.FunctionType_MinHash:
 		if field.GetDataType() != schemapb.DataType_BinaryVector {
-			return merr.WrapErrParameterInvalidMsg("output field data type must be binary vector for minhash materialization")
+			return merr.WrapErrDataIntegrityMsg(
+				"persisted minhash output field %d must be BinaryVector, got %s",
+				field.GetFieldID(), field.GetDataType())
 		}
 	}
 	return nil
-}
-
-func partialMaterializerExistingFields(schema *schemapb.CollectionSchema, missingFunctions []*schemapb.FunctionSchema, existingFields map[int64]struct{}) map[int64]struct{} {
-	fields := collectionSchemaFields(schema)
-	for _, functionSchema := range missingFunctions {
-		for _, outputIndex := range functionOutputIndexesToMaterialize(functionSchema, existingFields) {
-			delete(fields, functionSchema.GetOutputFieldIds()[outputIndex])
-		}
-	}
-	return fields
 }
 
 func (t *bumpSchemaVersionCompactionTask) openRecordReader(segment *datapb.CompactionSegmentBinlogs, schema *schemapb.CollectionSchema) (storage.RecordReader, map[int64]struct{}, error) {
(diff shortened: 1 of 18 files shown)


internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -89,17 +107,22 @@ func (t *bumpSchemaVersionCompactionTask) Compact() (*datapb.CompactionPlanResul
...
 	)
 
 	var result *datapb.CompactionPlanResult
-	if len(missingFunctions) == 0 && len(droppedFieldIDs) == 0 {
+	// Dropped physical fields always require a replacement rewrite. Zero-row
+	// segments still route to additive reconciliation, which rejects them as a
+	// data-integrity error rather than writing an empty materialized record.
+	if len(diff.droppedFieldIDs) > 0 {
+		result, err = t.runFullSchemaRewrite(diff.existingFields)
+	} else if len(diff.absentOrdinaryFields) == 0 && len(diff.missingOutputFields) == 0 {
```
liliu-z, line 122: This routing change makes the metadata-only fast path effectively dead for the most common schema change. Before this PR the branch was `if len(missingFunctions) == 0 && len(droppedFieldIDs) == 0 { runSchemaVersionBumpOnly() }`, so adding a plain nullable field with no function attached was a pure metadata bump. Now any field absent from the manifest lands in `diff.absentOrdinaryFields` (schemaBumpDecision:393-405 filters only system fields and function outputs), which routes to `runAdditivePhysicalReconciliation` — a full scan of the row-anchor column plus an N-row null/default column group and a new manifest commit per segment. Because `bumpSchemaVersionPolicy.staleFlushedSegments` enqueues one task for *every* flushed StorageV3 segment whose SchemaVersion lags, a single `AddCollectionField` on a large collection now fans out into a cluster-wide, un-gated write proportional to total row count. The tests that set `Schema.Functions = nil` (TestAdditiveReconciliationPersistsOrdinaryNullableField / …OrdinaryDefaultField) confirm this is intentional, but there is no size threshold, throttle, or config gate on it. Consider restricting the data path to fields a missing function actually consumes, or gating the ordinary-field prefill behind a config knob.

MrPresent-Han: Thanks for the detailed analysis — the cost characterization is fair, but we'd keep the current behavior, for two reasons.

**1. Physical completeness is the established postcondition of advancing a segment's schema version, not something this PR invented.** Every rewrite path already enforces it: mix / sort / merge-sort / clustering all wrap their readers with `RecordMaterializer`, so absent ordinary fields are materialized as default/NULL columns and the output adopts `t.GetSchema().GetVersion()` (mix_compactor.go:304, sort_compaction.go:263, merge_sort.go:84, clustering_compactor.go:608; meta.go:2614). If the bump-only path adopted the version without materializing, the same `schema_version_x` would describe two different physical layouts depending on which compaction advanced it — and since V3 no longer persists per-field binlog KVs, the only way to tell them apart would be reading the manifest. Keeping the additive backfill makes `SchemaVersion` a well-defined physical state at minimal cost, without a full rewrite.

**2. The fan-out is much lighter than it looks, and it is throttled.** The additive path reads only the row-count anchor column (RowID/Timestamp) and writes the new column groups, which for null/default values compress to near-metadata size — 1–2 orders cheaper than mix/mergesort's full-read-full-write. The per-segment tasks also go through the standard compaction slot scheduling (`GetSlotUsage`), so they queue against the same budget as other compactions rather than hitting the cluster at once, and the whole thing is a one-off triggered at DDL time under operator control.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -321,13 +323,203 @@ func (t *bumpSchemaVersionCompactionTask) openRecordReader(segment *datapb.Compa
...
+		anchor = nil
+		if _, ok := diff.existingFields[common.TimeStampField]; ok {
+			anchor = typeutil.GetField(t.plan.GetSchema(), common.TimeStampField)
+		}
+	}
+	if anchor == nil {
+		return nil, nil, merr.WrapErrDataIntegrityMsg(
+			"schema bump additive reconciliation requires a physical RowID or Timestamp anchor")
+	}
+	addExistingReadField(anchor)
```
buqian-zilliz: **[P1] Bound additive batches by materialized width**

The additive reader contains only the RowID/Timestamp anchor plus physically present function inputs. For an ordinary-field-only bump, the default 32 MiB reader buffer can therefore yield roughly four million rows from a single int64 anchor. RecordMaterializer.Wrap then allocates every absent ordinary column, and any derived function output, for that entire row count before the writer sees the batch, so a modest default string or several appended fields can expand one input batch into gigabytes and OOM the DataNode. Bound the reader by output/materialized width or split records into a fixed row-count batch before materialization.

internal/storage/schema.go
```diff
@@ -135,6 +135,12 @@ func overrideTextFieldsToBinary(schema *schemapb.CollectionSchema, arrowSchema *
 	return overrideTextFieldsToBinaryByFields(typeutil.GetAllFieldSchemas(schema), arrowSchema)
 }
 
+// OverrideTextFieldsToBinary returns the physical manifest schema for records
+// whose TEXT columns already contain encoded LOB references or binary NULLs.
+func OverrideTextFieldsToBinary(schema *schemapb.CollectionSchema, arrowSchema *arrow.Schema) *arrow.Schema {
```
buqian-zilliz: **[P3] Avoid exporting a position-dependent schema transform**

OverrideTextFieldsToBinary is a pure pass-through with no production caller; the reconciliation path uses NewPartialPackedRecordBatchWriterWithTextRefsAsBinary instead. Its contract is also unsafe for arbitrary callers because the implementation matches typeutil.GetAllFieldSchemas(schema) to Arrow fields by position rather than field ID or metadata, so a reordered or filtered Arrow schema silently converts the wrong columns. Remove the exported wrapper or expose an API that validates the ordering precondition.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Thanks — dug into this end-to-end (probes + milvus-storage source), and the amplification turns out to be bounded by the storage layer's batching contract, so the code stays as is; evidence below.

The premise "a 32 MiB reader buffer over an 8 B anchor yields ~4M rows per batch" doesn't hold on the V3 manifest read path. `FileRowGroupReader::ReadNextRowGroup` delivers exactly **one row group per call** (`SliceRowGroupFromTable` slices precisely `row_num()` rows); `buffer_size` only sizes the prefetch pool (how many row groups get cached ahead), never the delivered batch. And row groups are capped at ~1 MB uncompressed at write time (`DEFAULT_MAX_ROW_GROUP_SIZE`, enforced by batch slicing in `ParquetFileWriter::Flush`), per column-group file.

So one batch carries at most ~1 MB / column-group-row-width rows, and the post-materialization footprint is ~1 MB × materializedWidth / groupWidth per batch — for the quoted scenario (~100 B default string over a ~100 B-wide group) that's ~1 MB, not hundreds of MB. Empirical check: a 30,000-row segment written as 4 row groups delivers exactly 4 batches of 7,500 rows under every combination of projection (full vs anchor-only) × buffer size (32 MiB vs 8 KiB).

Since this rests on an implicit reader contract, I've prepared a characterization test (`TestBumpUTReaderDeliversPerRowGroupBatches`) pinning the one-row-group-per-batch delivery; it lands with the unit-test batch for this PR. If the reader ever starts packing multiple row groups per batch, that test fails and this concern must be re-assessed. Residual worst case is (1 MB / groupWidth) × materializedWidth, which only degrades with extreme long default values over very narrow column groups.

internal/storage/schema.go
MrPresent-Han: Removed in the latest push: `OverrideTextFieldsToBinary` and its test are gone. The reconciliation path already goes through `NewPartialPackedRecordBatchWriterWithTextRefsAsBinary`, so the position-dependent transform no longer has an exported surface.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -321,13 +323,203 @@ func (t *bumpSchemaVersionCompactionTask) openRecordReader(segment *datapb.Compa
...
+	var missingOutputFields []*schemapb.FieldSchema
+	for _, functionSchema := range schema.GetFunctions() {
+		outputFieldIDs := functionSchema.GetOutputFieldIds()
+		presentCount := 0
+		for _, outputFieldID := range outputFieldIDs {
+			if _, present := existingFields[outputFieldID]; present {
+				presentCount++
+			}
+		}
+		if presentCount == len(outputFieldIDs) {
```
buqian-zilliz, line 401: **[P3] Zero-output functions are silently skipped before validation**

When a persisted function has an empty `output_field_ids`, `presentCount == len(outputFieldIDs)` evaluates to `0 == 0`, so the loop continues before ever calling `validateSupportedMissingFunctionMaterialization`. This makes the `len(functionSchema.GetOutputFieldIds()) == 0` branches in that validator unreachable, and the malformed function is silently ignored in both the additive path (`missingFunctionMaterializations` never adds it to `missingFunctions`) and the full-rewrite path (`functionOutputIndexesToMaterialize` also returns `nil, nil`). A function with no outputs is persisted-schema corruption that this PR intends to reject as a data-integrity error; instead it is dropped without any signal. Consider validating the empty-output case before the `presentCount == len(outputFieldIDs)` early return.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -756,6 +946,28 @@ type bumpSchemaVersionBatchWriter interface {
...
+type bumpSchemaVersionWriterLease struct {
+	writer    bumpSchemaVersionBatchWriter
+	exhausted bool
+}
+
+func (l *bumpSchemaVersionWriterLease) Close() (packed.WriterOutput, error) {
+	if l.exhausted {
+		return nil, merr.WrapErrServiceInternalMsg("schema bump writer lease already consumed")
+	}
+	l.exhausted = true
```
buqian-zilliz, line 974: **[P3] Writer lease leaks the FFI writer when Close fails**

`Close` sets `l.exhausted = true` before calling `l.writer.Close()`. If `Close` returns an error, the deferred `Cleanup` observes `exhausted == true` and does not call `Abort`, while `packedRecordBatchWriter.Close` leaves `pw.writer` non-nil on error. The underlying FFI writer is therefore never `Destroy`ed on this failure path, leaking native resources. Mark the lease exhausted only after a successful `Close`, or have `Close` call `Abort` on error so the FFI writer is always released.

MrPresent-Han: Fixed in 420827e445. Two notes:

1. On the actual leak: today this path does not leak natively, because `FFIPackedWriter.Close` destroys `cWriterHandle`/`cProperties` in a defer regardless of the close result — so the skipped `Abort` was a no-op in practice.
2. Agreed the lease should not rely on that non-local invariant. `Close` now consumes the lease only on success; on error the deferred `Cleanup` runs `Abort` (idempotent: `Destroy` guards on a zeroed handle, so no double-free).

Covered by `TestBumpUTUnitWriterLeaseAbortsOnCloseError` (Close failure -> exactly one Abort).

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Fixed in 420827e445. A zero-output persisted function now fails loud with DataIntegrity ("persisted function %s has no output fields") in both paths: `missingFunctionMaterializations` (additive decision) and `functionOutputIndexesToMaterialize` (full-rewrite/materializer selection), checked before the all-present early return so the `0 == 0` continue can no longer swallow it.

Deliberately a targeted `len == 0` check rather than hoisting `validateSupportedMissingFunctionMaterialization` above the early return: fully-present functions of non-materializable types (e.g. TextEmbedding) must still be skipped, not rejected. Note DDL already rejects zero-output functions (`validator.ValidateFunction` requires exactly 1 output for BM25/MinHash), so this guard is reachable only through persisted-metadata corruption — which is exactly the class this PR fails loud on.

Covered by `TestBumpUTUnitZeroOutputFunctionRejected` (both entry points assert ErrDataIntegrity).

internal/datanode/compactor/record_materializer.go
```diff
@@ -116,35 +120,44 @@ func (m *RecordMaterializer) WrapWithSelection(rec storage.Record, selection *re
 		return base, nil
 	}
 
-	computed := make(map[int64]arrow.Array)
-	for _, materializer := range m.materializers {
-		arrays, err := materializer.Materialize(base)
+	ordinaryFields := make(map[int64]arrow.Array, len(m.missingFields))
+	for _, field := range m.missingFields {
+		arr, err := storage.GenerateEmptyArrayFromSchema(field, base.Len())
```
buqian-zilliz: **[P2] Additive reconciliation fails for newly added non-nullable fields**

`absentOrdinarySchemaFields` collects every physically-absent non-system field, including non-nullable scalar fields, and `WrapWithSelection` feeds them into `storage.GenerateEmptyArrayFromSchema`. That helper returns `missing field data` for any field with `GetNullable() == false` before it ever consults the field's default value. Schema evolution permits adding a non-nullable field as long as it has a default value, so `AddField` succeeds at the metadata layer but the subsequent additive schema-bump reconciliation fails, leaving the segment unreconciled. Materialize the default value, or route non-nullable fields to a path that can backfill them, instead of erroring out.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -1067,16 +1267,22 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
...
 	}
 	span.End()
+	if totalRows != t.plan.GetTotalRows() {
+		return nil, merr.WrapErrDataIntegrityMsg(
+			"schema bump additive reconciliation read %d rows, expected %d",
+			totalRows, t.plan.GetTotalRows())
+	}
+	// A zero-row materialized write would manufacture physical completeness for
+	// an empty segment; fail and let the deferred lease Abort the writer.
+	if totalRows == 0 {
```
buqian-zilliz, line 1283: **[P3] Additive reconciliation rejects zero-row segments**

`runAdditivePhysicalReconciliation` returns a data-integrity error when `totalRows == 0`, which is only reachable when `plan.GetTotalRows()` is also zero. A segment that legitimately has zero physical rows but still needs a schema bump for newly added fields will therefore fail compaction instead of completing as a version bump or no-op. Please confirm whether DataCoord can schedule a schema-bump compaction for a zero-row segment; if it can, this turns a valid state into a permanent compaction failure.

internal/datanode/compactor/record_materializer.go
MrPresent-Han: Not adopting for this PR. Added fields are required to be nullable: `addCollectionFieldTask.PreExecute` rejects `!GetNullable()` unconditionally ("added field must be nullable"), with no default-value exemption. So a non-nullable ordinary field can never become physically absent from a sealed segment through the supported AlterCollectionField path, and `absentOrdinarySchemaFields` will only ever carry nullable fields into `GenerateEmptyArrayFromSchema`.

The only way to persist a non-nullable added field is bypassing the proxy admission check via a direct RootCoord call — the same admission-gap class being closed separately in #52219 — which is out of scope for this reconciliation fix. Given the invariant, the current behavior is correct rather than a reachable failure.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Not adopting — a zero-row segment is not a valid schedulable input for schema-bump, so the loud reject is intended defensive behavior, not a valid state turned into a permanent failure.

DataCoord only schedules Flushed segments for bump (`isSchemaBumpDataSegment` requires `isFlushed`), and a Flushed segment is non-empty by construction: empty write buffers never flush, and a compaction whose result has zero rows is adopted as Dropped, not Flushed (e.g. `completeBumpSchemaVersionReplacementMutation` sets `State = Dropped` when `NumOfRows == 0`). So `plan.TotalRows == 0` reaching additive reconciliation indicates a corrupt/inconsistent segment; surfacing it as a data-integrity error (instead of silently writing an empty materialized record) is the correct outcome.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -321,13 +323,209 @@ func (t *bumpSchemaVersionCompactionTask) openRecordReader(segment *datapb.Compa
...
+		}
+	}
+	return nil
+}
+
+// absentOrdinarySchemaFields returns the non-system, non-function-output schema
+// fields that are physically absent from the manifest.
+func absentOrdinarySchemaFields(schema *schemapb.CollectionSchema, existingFields, functionOutputFields map[int64]struct{}) []*schemapb.FieldSchema {
+	absent := make([]*schemapb.FieldSchema, 0)
+	for _, field := range typeutil.GetAllFieldSchemas(schema) {
```
buqian-zilliz, line 454: **[P3] Additive schema bump flattens struct-array sub-fields into standalone column groups**

`absentOrdinarySchemaFields` iterates `typeutil.GetAllFieldSchemas(schema)`, which also returns `StructArrayField` sub-fields, so those sub-fields flow into `appendedFields()` and then `setupWriter`. `setupWriter` emits one standalone single-column group per field (`GroupID: fieldID`, `Fields: []int64{fieldID}`) without any parent/child relationship. If a struct array or one of its sub-fields can be added through schema evolution, the additive path would write the struct array as independent top-level column groups and lose the nested struct layout. Please confirm whether struct-array fields are reachable in this path; if they are, `absentOrdinarySchemaFields`/`setupWriter` need to preserve the parent-child structure.

MrPresent-Han: Not adopting — there is no physical nested struct layout to lose. Struct-array sub-fields are stored as flat, independent columns everywhere: `ConvertToArrowSchema` appends each sub-field with the same `appendArrowField` used for top-level fields (no nested arrow struct type), so every write path (insert/flush, mix/sort compaction) emits per-field column groups exactly like `setupWriter` does here. The parent-child relationship lives only in the `StructArrayFields` metadata of the CollectionSchema, not in the manifest column-group hierarchy.

The bump result adopts the target schema (which carries that `StructArrayFields` metadata unchanged), so the logical struct grouping is preserved and the flat physical columns match how struct sub-fields are stored throughout. Emitting a standalone single-column group per absent sub-field is therefore the correct physical representation, not a flattening that loses layout.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -900,72 +1138,32 @@ func (t *bumpSchemaVersionCompactionTask) buildNewInsertLogsV3(writerResult *bum
...
 
-	inputSchema, inputFieldIDs, err := t.missingFunctionInputSchema(missingFunctions)
+	inputSchema, inputFieldIDs, err := t.additiveReadSchema(diff)
 	if err != nil {
 		return nil, err
 	}
-	outputFields, outputFieldIDs, err := t.missingFunctionOutputFields(missingFunctions, existingFields)
-	if err != nil {
-		return nil, err
+	outputFields := diff.appendedFields()
```
liliu-z, line 1161: The result ships only the appended column groups in `InsertLogs`, and DataCoord upserts exactly those onto `SegmentInfo.Binlogs` (meta.go:3590-3593). V3 segments never persist per-FieldBinlog KVs (kv_catalog.go:357 `if isV3Segment(segment) { continue }`), so a segment recovered after a DataCoord restart has empty `Binlogs` — which is what makes QueryNode fall through to the complete `Stats.LoadResource` summary. After a bump adoption that array is non-empty but covers only the newly appended fields, and `resolveSegmentEstimateLogs` returns early on any non-empty `BinlogPaths`, so every pre-existing column drops out of the memory/disk estimate and the segment can be over-admitted. The mechanism is pre-existing (the old function-materialization route had the same shape), but this PR widens the trigger from "a function output is missing" to "any ordinary field is physically absent". Either merge the appended descriptors with `Stats.LoadResource` at adoption, or make the estimator prefer `LoadResource` for manifest-backed segments.

internal/storage/record_writer.go
```diff
@@ -271,6 +271,17 @@ func (pw *packedRecordBatchWriter) Close() (packed.WriterOutput, error) {
 	return out, nil
 }
 
+// Abort releases the underlying FFI writer without producing column groups for
+// CommitManifestUpdates. It prevents metadata publication; storage already
+// flushed before the abort remains subject to normal orphan-file reclamation.
```
liliu-z: The new `Abort()` doc comment claims flushed storage "remains subject to normal orphan-file reclamation", which is not true for a live V3 segment: DataCoord GC explicitly skips any file it resolves to a registered StorageV3 segment (garbage_collector.go:722-728, "registered V3 segment file — skip, live files are managed by loon"), and no loon vacuum is exposed. The same gap applies to the additive path's failure window: once `writerLease.Close()` succeeds the objects are finalized, and a subsequent `CommitManifestUpdates` error returns without deleting them while `writerOutput.Destroy()` frees only the C-side handle. Because `bumpSchemaVersionTask.processFailed()` has no retry cap, every re-trigger adds another unreachable copy. Either delete the finalized paths on failure or correct the comment to state that these files are not reclaimable for a live V3 segment.

internal/datanode/compactor/bump_schema_version_ut_test.go
```diff
@@ -0,0 +1,2542 @@
...
+	fix := additiveFixture(t)
+	fw := &faultBatchWriter{closeErr: errBumpUTInjected}
+	patch := mockey.Mock((*bumpSchemaVersionCompactionTask).setupWriter).
+		Return(faultWriterResult(fw, fix), nil).Build()
+	defer patch.UnPatch()
+
+	_, err := fix.task.Compact()
+	require.ErrorIs(t, err, errBumpUTInjected)
+	require.Equal(t, 1, fw.closes)
+	require.Zero(t, fw.aborts, "lease must not Abort an already-consumed writer")
```
liliu-z: This assertion contradicts the lease contract and fails deterministically. `bumpSchemaVersionWriterLease.Close` returns the error without setting `exhausted`, precisely so the deferred `Cleanup` aborts; the fault writer's `Close` is reached (the test itself asserts `fw.closes == 1`), so `defer writerLease.Cleanup()` runs `Abort()` and `fw.aborts` is 1, not 0. The unit-level test for the same scenario at line 2406 asserts `abortCount == 1`. Change to `require.Equal(t, 1, fw.aborts)` and fix the comment.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -1067,16 +1267,22 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
...
-			mlog.Int("rows", wrapped.Len()),
-			mlog.Int64("totalRowsBeforeBatch", totalRows),
-			mlog.Int("outputFieldCount", len(outputFields)),
-		)
 
 		totalRows += int64(wrapped.Len())
 		cleanupMaterializedRecord(wrapped)
 	}
 	span.End()
+	if totalRows != t.plan.GetTotalRows() {
```
liliu-z, line 1276: The two new integrity gates are terminal for the segment but the bump scheduler has no retry cap or backoff: `bumpSchemaVersionTask.processFailed()` is `return true`, which just exits the state machine, and the policy re-selects the same segment on the next trigger interval because its `SchemaVersion` still lags. So a segment whose DataCoord `NumOfRows` diverges from its physical row count — or a legitimately zero-row flushed segment, which `staleFlushedSegments` does not filter out — re-runs the read/write/commit cycle forever with only WARN logs to show for it. The row-count check itself is right; what is missing is a bounded retry and a surfaced terminal failure on the DataCoord side.

internal/datanode/compactor/bump_schema_version_ut_test.go
MrPresent-Han: Fixed in 0fd19ba9af. Exactly as you said — this fault test still encoded the pre-fix contract (no Abort after a failed Close) and the lease fix missed it. Assertion updated to `require.Equal(t, 1, fw.aborts)` (Close failure keeps the lease unconsumed, so the deferred Cleanup must Abort), consistent with `TestBumpUTUnitWriterLeaseAbortsOnCloseError`.

internal/storage/record_writer.go
MrPresent-Han: Good catch — the doc comment overstated reclamation: GC deliberately skips files under a registered V3 segment prefix, so files flushed before an Abort are NOT picked up by the normal orphan sweep. Corrected the comment in 0fd19ba9af to state they stay on storage as unreferenced objects whose reclamation depends on the manifest-aware orphan cleanup tracked by #51649. The commit-failure window you describe (Close succeeded, CommitManifestUpdates failed → finalized files unreachable from any manifest) falls under the same follow-up's scope.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Fixed in 0fd19ba9af, by removing the population instead of merging into LoadResource: the in-place adoption no longer merges the result's InsertLogs into SegmentInfo.Binlogs at all. A V3 segment's in-memory Binlogs now stays empty — matching persistence, where kv_catalog resets the array and skips per-FieldBinlog KVs for V3 — so `resolveSegmentEstimateLogs` no longer early-returns on a partial array and falls through to the complete `Stats.LoadResource` summary. This is also restart-equivalent: a recovered V3 segment reloads with empty Binlogs anyway, so the removal converges the pre-restart state to the already-supported one. Index eligibility is unaffected (gated by the schema-version advance; the build worker reads the manifest), and the full-rewrite path is untouched since its replacement Binlogs are the complete set. The same-shaped pre-existing populate in the V2->V3 backfill commit path is left as a separate follow-up.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Agreed the scheduler loop is the real gap (the integrity reject itself must stay — adopting a row-count-mismatched result would corrupt meta). This is a DataCoord-level, cross-compaction-type problem rather than a bump-specific one: today only clustering has bounded retry, mix/bump `processFailed` just returns true with no cap or backoff, the policy layer re-picks the same segments every trigger, and failed tasks are even counted into the `done` metric label. Opened #52708 to track generic bounded retry + failure surfacing platform-wide instead of patching only bump here.

internal/datacoord/meta.go
```diff
@@ -3569,28 +3569,14 @@ func (m *meta) completeBumpSchemaVersionCompactionMutation(
...
-	// additive Stats increment below it needs no manifest gate: re-applying a
-	// replayed result cannot double-count. Index eligibility is gated by the
-	// schema-version update below, while Binlogs retains the materialized column
-	// group metadata on the live SegmentInfo.
-	var droppedBinlogFieldIDs []int64
-	if newGroups := resultSegment.GetInsertLogs(); len(newGroups) > 0 {
-		cloned.Binlogs, droppedBinlogFieldIDs = mergeSegmentColumnGroups(cloned.GetBinlogs(),
-			lo.KeyBy(newGroups, func(fb *datapb.FieldBinlog) int64 { return fb.GetFieldID() }))
-	}
+	// Deliberately do NOT merge the result's InsertLogs into SegmentInfo.Binlogs.
```
liliu-z: The new comment justifies dropping the InsertLogs merge on the premise that "a V3 segment's Binlogs stay empty", but that is only true after a DataCoord restart: `resetBinlogFields` runs on a clone inside `buildAlterSegmentsKvs`, and every other V3 receiver in this same file populates the in-memory array from the result (`completeBumpSchemaVersionReplacementMutation` sets `Binlogs: resultSegment.GetInsertLogs()`, as does the mix-compaction receiver). So a live V3 segment does carry a non-empty `Binlogs`, `resolveSegmentEstimateLogs` takes its early return, and — with the merge removed — the estimate no longer sees the column this run materialized. The fallback the comment relies on is short too: `buildMaterializationStatsDelta` emits no `LoadResource`, and `addStatsDelta` only folds insert size/count, stats size and null counts, carrying `LoadResource` over from base untouched. Both the pre- and post-restart estimation paths therefore under-count every segment with a materialized column; previously the merge kept the pre-restart path correct.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -63,6 +62,25 @@ type bumpSchemaVersionCompactionTask struct {
...
+	existingFields       map[int64]struct{}
+	droppedFieldIDs      []int64
+	absentOrdinaryFields []*schemapb.FieldSchema
+	missingOutputFields  []*schemapb.FieldSchema
+	missingFunctions     []*schemapb.FunctionSchema
+}
+
+func (d *schemaBumpPhysicalDiff) appendedFields() []*schemapb.FieldSchema {
+	fields := make([]*schemapb.FieldSchema, 0, len(d.absentOrdinaryFields)+len(d.missingOutputFields))
+	fields = append(fields, d.absentOrdinaryFields...)
```
liliu-z, line 79: `appendedFields()` now hands absent ordinary fields to the additive writer, so a newly added nullable Timestamptz that the collection names as `ttl_field` gets physically backfilled with its configured default — but the additive result ships no `ExpirQuantiles` (the full-rewrite path does, from `writer.GetLogs()`), and the in-place receiver only clones the old segment, whose quantile list is empty. `ShouldCompactExpiryWithTTLField` bails on `len(percentiles) == 0`, so even when the backfilled timestamp is already in the past, TTL never schedules the segment for cleanup and the rows sit until an unrelated compaction touches them. Note the PR's own `TestAdditiveReconciliationPersistsTTLDefaultWithoutFiltering` constructs exactly this case (default = `currentTime - 1h`) and asserts only that rows are not filtered. Not a regression — pre-PR the column was never materialized either — but the write cost is now paid and the reclamation step is missing.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -1101,13 +1307,13 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
 		return nil, err
 	}
 
-	writerOutput, err := writerResult.writer.Close()
-	if err != nil {
-		return nil, err
-	}
+	writerOutput, err := writerLease.Close()
```
liliu-z, line 1311: `CommitManifestUpdates` opens at the stale `baseVersion`, and the loon retry "re-reads the latest manifest version and re-applies the changes"; since column-group appends "don't collide on a key", a rebased replay (DataNode commits v13, dies before DataCoord adopts, and the rebuilt plan re-reads the stale v12 manifest) appends the same field's column group a second time. Assert the returned version equals `baseVersion + 1` and fail otherwise, or make column-group additions idempotent.

internal/datanode/compactor/compactor_common.go
```diff
@@ -354,9 +344,16 @@ func compactionReadSchema(schema *schemapb.CollectionSchema, existingFields map[
...
+// physically present, or when the reader layer can synthesize it (absent
+// ordinary fields are reader-filled with default/null). Only function outputs
+// missing from storage are excluded: storage cannot synthesize them and the
+// RecordMaterializer computes them instead.
 func compactionFieldReadable(field *schemapb.FieldSchema, existingFields map[int64]struct{}) bool {
-	_, ok := existingFields[field.GetFieldID()]
-	return ok
+	if _, ok := existingFields[field.GetFieldID()]; ok {
+		return true
+	}
```
buqian-zilliz, line 356: **[P3] `compactionFieldReadable` now reader-fills absent fields for every compaction path, not just schema bump**

`compactionFieldReadable` is shared by `merge_sort`/`mix`/`sort`/`clustering` compaction through `newCompactionSegmentRecordReader`. Changing its absent-field branch from `false` to `return !field.GetIsFunctionOutput()` keeps physically-absent ordinary fields in the read schema for those readers too, and the V3 manifest path now always passes `WithStorageFields`, so `ManifestReader` synthesizes them as default/null. For same-schema-version inputs this is a no-op; but if any of those paths can read a segment whose manifest lacks a schema field (a schema-evolution gap not yet handled by schema bump), the output now gains a synthesized column that was previously absent. Please confirm those paths cannot observe such segments, or document that this read-schema contract is intentional for all compaction readers.

internal/storage/record_reader.go
```diff
@@ -338,14 +352,35 @@ func NewManifestReaderWithExtfs(
...
+	// physically absent from the manifest are not requested from the FFI
+	// reader (whose absent-column handling is null-only); Next() fills them
+	// Go-side with default/null semantics. Activated by WithStorageFields —
+	// the caller declares the physically present set it derived from this
+	// same manifest; without the option the reader keeps requesting every
+	// resolvable column. Scoped to internal milvus tables: for external
+	// collections a schema/manifest difference reflects source mapping, not
+	// an added field, and must not fabricate values.
+	storageFields := rwOptions.storageFields
+	fillEligible := storageFields != nil &&
```
liliu-z: Moving the ordinary-field fill out of RecordMaterializer and into ManifestReader leaves a gap on the StorageV2 packed (non-manifest) path. compactionFieldReadable now keeps absent ordinary fields in the read schema for every reader, but only the manifest reader gets WithStorageFields, so a V2 segment requests the absent column from the C++ PackedRecordBatchReader, which fills MakeArrayOfNull. A field added with a default value therefore lands as NULL where master applied the default via missingNonMaterializedSchemaFields. The V1 path is not affected — CompositeBinlogRecordReader already default-fills via GenerateEmptyArrayFromSchema — so scope the fix to the StorageV2 packed reader (or restore the materializer fill for non-manifest sources).

internal/datacoord/meta.go
MrPresent-Han: Adopting this. You are right on both counts: live V3 segments do carry non-empty Binlogs (flush and every compaction receiver populate them), and the materialization delta never touches Stats.LoadResource, so removing the merge traded the pre-restart-correct path away. We are reverting the merge removal — the in-place adoption upserts the appended column groups again, keeping the live array complete and the not-yet-restarted estimate correct. The post-restart side moves to a dedicated follow-up: with the merge back, a restart-then-bump sequence rebuilds a partial array (empty ∪ increment) and re-exposes the over-admission I originally tried to fix, and LoadResource still misses materialized columns on both paths; the follow-up folds the appended groups into Stats.LoadResource at adoption and converges the live array to the restart shape so both estimation paths read one complete source.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Verified against the loon commit internals — the double-append cannot happen, so not adopting the assert. Version numbers come from list-max discovery plus an exclusive-create CAS (Transaction::get_latest_version lists the manifest dir and takes max; write_manifest creates the next version file exclusively, AlreadyExists retries), so a rebuilt plan's commit lands at v14, never a second v13. And the rebase does not re-apply onto latest: CommitManifestUpdates uses LOON_TRANSACTION_RESOLVE_OVERWRITE, whose resolver ignores the latest manifest entirely and applies the staged updates onto the read (base) manifest — v14 = v12 content + exactly one copy of the appended group; the orphaned v13's copy is discarded, leaving only unreferenced files (the #51649 reclamation class, pre-existing). Asserting returned version == baseVersion+1 would turn these benign orphan retries into hard failures. One adjacent point the overwrite semantics does raise — any concurrent manifest commit landing in the read→commit window gets dropped from the winning manifest; whether a concurrent writer actually exists for a sealed segment's manifest (e.g., delta additions) is being verified separately as a follow-up.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Confirmed, and your framing is exact: both routes deliberately avoid filtering on the just-backfilled default (a schema change must not become a silent mass delete), so reclamation is delegated to TTL scheduling — which the additive route leaves blind because it ships no ExpirQuantiles. We considered producing quantiles inside the additive pass (the backfilled column is a constant, so it would be O(1)), but rejected it: that makes TTL-scheduling metadata correctness depend on which compaction happened to touch the ttl field — opportunistic ownership, and it still misses segments whose quantiles are empty for other historical reasons. Decision: keep the existing invariant that ExpirQuantiles are produced only by full-pass writers (sort/mix/clustering/full-rewrite); a backfilled segment becomes TTL-schedulable at its next natural full rewrite, and until then this is a documented known limitation of the in-place route. A trigger-side metadata fallback (for a sealed segment whose ttl column is provably all-default, expiry is decidable from the schema default alone, no statistics needed) is noted as possible future hardening.

internal/storage/record_reader.go
MrPresent-Han: Fixed in 929e59e085. Rather than restoring the materializer fill for non-manifest sources, the packed (StorageV2/V3-binlog) path now gets the same reader-fill semantics through a fillingRecordReader decorator: the physically-present set is derived from the binlog list's FieldID/ChildFields, and absent readSchema fields are synthesized Go-side (default when declared, else null, error on non-nullable), shadowing the C++ reader's null-only fill. The same commit also makes the fill unconditional on milvus-semantics tables — the manifest reader now self-derives the present set when WithStorageFields is absent, so the contract no longer depends on the caller remembering an option. Covered by a real write-then-read case in TestPackedBinlogRecordIntegration (absent nullable reads back all-null, absent default reads back the declared default) plus a decorator contract unit test; V1 keeps its existing Go-side fill untouched.

internal/datanode/compactor/compactor_common.go
MrPresent-Han: Intentional, and now documented as the reader contract: a compaction read schema keeps absent ordinary fields (dropping only function outputs missing from storage), and the reader layer guarantees the record comes back readSchema-wide — default when declared, else null, error on non-nullable. This is deliberately uniform across all compaction types: it is the same fill the RecordMaterializer used to perform after the read (same GenerateEmptyArrayFromSchema helper, value-identical), moved below the reader boundary so every consumer sees one semantics. Your conditional was real for StorageV2 packed segments — that path only had the C++ null-only fill, so a default-carrying added field would have landed as NULL; closed in 929e59e085 by giving the packed path the same reader-fill via a decorator (see the record_reader.go thread). V1 already had this behavior natively, and the V3 manifest path gained it in the earlier commit.

internal/storage/record_reader.go
```diff
@@ -296,6 +304,7 @@ func NewManifestReader(manifest string,
 		storageConfig,
 		storagePluginContext,
 		rwOptions.externalReader,
+		option...,
```
buqian-zilliz: **[P3] RwOptions are applied twice on the NewManifestReader path**

`NewManifestReader` runs every `RwOption` once to extract `rwOptions.externalReader`, then forwards the same `option...` slice to `NewManifestReaderWithExtfs`, which applies every option again to a fresh `DefaultReaderOptions()`. Since `NewManifestRecordReader` already applies the same options to derive buffer size, storage config and plugin context, each option function now executes multiple times per manifest reader construction. This is harmless while every `RwOption` is an idempotent field setter, but the duplicate application is redundant and becomes a silent footgun the moment an option is ever non-idempotent (e.g. appends to a slice). Pick a single point of application: either pass the resolved `rwOptions` down or have `NewManifestReader` forward without re-applying.

internal/storage/record_reader.go
```diff
@@ -338,14 +352,39 @@ func NewManifestReaderWithExtfs(
...
+	// reader (whose absent-column handling is null-only); Next() fills them
+	// Go-side with default/null semantics unconditionally. WithStorageFields
+	// is only a performance hint sparing the reader from re-deriving the
+	// present set from the manifest. Scoped to milvus-semantics tables: for
+	// external-format sources a schema/manifest difference reflects source
+	// mapping, not an added field, and must not fabricate values.
+	fillEligible := !typeutil.IsExternalCollection(schema) || columnResolver.IsMilvusTable()
+	storageFields := rwOptions.storageFields
+	if fillEligible && storageFields == nil {
+		storageFields, err = packed.GetManifestFieldIDs(manifest, storageConfig)
```
liliu-z: The new physical-presence probe opens the manifest via `GetManifestFieldIDs(manifest, storageConfig)`, which routes through `GetManifestHandle` → `GetManifestHandleWithExtfs(..., ExternalSpecContext{})`, so the `extfs` argument this constructor was handed is dropped. The milvus-table virtual-PK delete path builds a two-field synthetic source schema (no `ExternalField`, so `fillEligible` is true) and passes a *source* manifest whose endpoint/credentials live only in the external spec; the probe runs before the extfs-aware FFI reader is created, so reader construction now fails and source deletes can no longer be translated. Every other source-manifest access in that file uses an extfs-aware variant (`GetDeltaLogsFromManifestWithExtfs`, `ReadFileWithExternalSpec`, `WithExternalReaderContext`). It also re-opens the manifest once per fragment per delete batch. Thread extfs through an extfs-aware `GetManifestFieldIDs`, and let callers that already derived the present set pass it in.

internal/storage/rw.go
```diff
@@ -350,7 +384,12 @@ func NewBinlogRecordReader(ctx context.Context, binlogs []*datapb.FieldBinlog, s
 			}
 		}
 		// FIXME: add needed fields support
-		rr = newIterativePackedRecordReader(paths, schema, rwOptions.bufferSize, rwOptions.storageConfig, pluginContext, rwOptions.externalReader)
+		inner := newIterativePackedRecordReader(paths, schema, rwOptions.bufferSize, rwOptions.storageConfig, pluginContext, rwOptions.externalReader)
+		// The packed reader null-fills schema columns absent from the files
+		// (C++ knows no field defaults); wrap it so absent fields follow the
+		// reader contract: default when declared, else null, error on
+		// non-nullable. The binlog list is the storage-presence truth here.
+		rr = newFillingRecordReader(inner, absentBinlogSchemaFields(schema, binlogs))
```
liliu-z: `absentBinlogSchemaFields` walks `GetAllFieldSchemas` unfiltered and `GenerateEmptyArrayFromSchema` errors on any non-nullable field, so this wrapper converts previously readable segments into a first-`Next()` failure. Two concrete triggers: (1) `internal/datanode/index/task_stats.go:282` opens the reader with the full current collection schema plus `segment.GetBinlogs()`, so a function-output field never materialized at flush is absent and non-nullable and the sort job fails on every rebuild instead of null-filling as before; (2) presence is derived from `ChildFields`, but packed binlogs written before #44282 carry none and their `FieldID` is a group ID (`DefaultShortColumnGroupID = 0`, sequential IDs), so every member field of such a group — including the PK — is fill-requested. The compaction caller is safe only because `compactionReadSchema`/`compactionFieldReadable` narrow the schema with the same presence rule first. Skip system and `IsFunctionOutput` fields at this layer and let callers with reconciliation context materialize missing outputs.

internal/datanode/compactor/record_materializer.go
```diff
@@ -237,6 +234,12 @@ func newSelectedRecord(base storage.Record, schema *schemapb.CollectionSchema, e
 }
 
 func buildSelectedColumn(base storage.Record, field *schemapb.FieldSchema, selection *recordSelection) (arrow.Array, error) {
+	// The reader contract guarantees base is readSchema-wide; a missing column
+	// here is a contract violation, reported instead of dereferenced.
+	if base.Column(field.GetFieldID()) == nil {
```
liliu-z: The new nil guard cannot fire. `newSelectedRecord` is reached only when `selection != nil`, whose sole caller is the bump full-rewrite loop, and that path always reads through the V3 `ManifestReader` — whose records are `simpleArrowRecord` (panics on an unknown field ID) or `filledRecord` (falls straight through to the same panic). So the intended contract-violation error is unreachable and a real violation would panic the datanode instead. Either drop the guard or check the field ID against the record's schema before calling `Column`.

internal/storage/record_reader.go
MrPresent-Han: Resolved by reverting the reader-layer default fill (1d17010df8). Absent ordinary fields are now null-filled by the read path rather than reader-materialized, so the self-compute probe (GetManifestFieldIDs) is gone entirely — record_reader.go is now byte-identical to the pre-enhancement baseline, and the milvus-table virtual-PK delete path builds its extfs-aware reader unchanged. The extfs-drop no longer exists.

internal/storage/rw.go
MrPresent-Han: Resolved by the same revert (1d17010df8). The fillingRecordReader decorator and absentBinlogSchemaFields are removed; rw.go is byte-identical to the baseline. The packed path keeps its original C++ MakeArrayOfNull behavior, so task_stats.go's non-nullable function-output read and the pre-#44282 group-id binlogs null-fill exactly as before — no first-Next() failure.

internal/datanode/compactor/record_materializer.go
MrPresent-Han: Dropped the guard in 1d17010df8, as you suggested — you are right it could not fire (simpleArrowRecord panics on an unknown field before the nil check). record_materializer.go no longer probes base.Column there.

internal/storage/record_reader.go
MrPresent-Han: Resolved by 1d17010df8 — the reverted record_reader.go no longer forwards option... into NewManifestReaderWithExtfs, so each RwOption is applied once again, matching the pre-enhancement path. The double-application is gone.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -321,13 +323,222 @@ func (t *bumpSchemaVersionCompactionTask) openRecordReader(segment *datapb.Compa
...
+		for _, field := range structField.GetFields() {
+			addExistingReadField(field)
+		}
+	}
+
+	// Absent ordinary fields (including absent function inputs) enter the read
+	// schema so the reader fills them per its contract; the writer then takes
+	// the appended columns straight from the read record, and functions see
+	// target-schema values for their inputs.
+	for _, field := range diff.absentOrdinaryFields {
```
liliu-z, line 518: This loop (`for _, field := range diff.absentOrdinaryFields`) adds every newly-added ordinary field to the read schema on the assumption "the reader fills them per its contract" — but `preCompact` only accepts StorageV3, whose reader contract is null-fill, not default-fill. `AddField("score", Int64, nullable=True, default_value=42)` therefore materializes NULL instead of 42 into every stale row within one `triggerInterval` (the PR's own tests pin it: `bump_schema_version_compactor_test.go:575` flips `NullCounts` 0→3, `:590` asserts `column.IsNull(i)`). It is also irreversible: once the column physically exists, segcore's schema-based `is_field_exist` and `ComputeDiffDefaultFields` skipping any field with a data source mean `DefaultValueChunkTranslator` never runs again, so the same row returns 42 before compaction and NULL after — silent data destruction on an automatic background path. Apply the declared default value during additive materialization instead of null-filling absent fields.

internal/datanode/compactor/bump_schema_version_compactor.go
```diff
@@ -1101,13 +1320,13 @@ func (t *bumpSchemaVersionCompactionTask) runMissingFunctionMaterialization(ctx
...
-	}
+	writerOutput, err := writerLease.Close()
 	if writerOutput != nil {
 		defer writerOutput.Destroy()
 	}
+	if err != nil {
+		return nil, err
+	}
 
 	manifestPath, err := packed.CommitManifestUpdates(
```
liliu-z, line 1319: CommitManifestUpdates publishes a new manifest version plus its column-group files before the result reaches DataCoord. If the result is lost (node restart, or QueryTaskOnWorker resetting the planID), DataCoord's ManifestPath stays at the pre-commit version and the rerun commits a fresh version built from that same base, so the earlier version's files are never referenced again. V3 GC only sweeps a segment's base path when the segment itself is dropped, so the garbage persists for the segment's lifetime. Pre-existing behaviour, but this PR routes far more segments through the path; worth tracking under the manifest-aware orphan cleanup the code already references.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: You are right that materializing a default-valued field as NULL is silent data destruction, and thank you for catching it. On investigation this turns out to be an orthogonal, pre-existing storage-reader-contract problem rather than something specific to this PR: the storage readers fill absent columns with MakeArrayOfNull because default_value is a Milvus schema concept the storage engine never receives (only the arrow schema + needed columns). V2 (PackedRecordBatchReader) and V3 (Reader::create/ReaderImpl) both do this; only the V1 CompositeBinlogRecordReader compensates via the default-aware GenerateEmptyArrayFromSchema. It predates #49759 — this PR only widens exposure by routing more segments through the absent-ordinary read path. Once the column is physicalized (as NULL), segcore's is_field_exist gate stops synthesizing the default, hence the irreversibility you describe. Tracked with full context and fix direction (default overlay at the Go reader layer for V2/V3) in a dedicated issue: #52771.

internal/datanode/compactor/bump_schema_version_compactor.go
MrPresent-Han: Agreed, and as you note this is pre-existing behaviour: CommitManifestUpdates publishes the manifest version and its column-group files before the result reaches DataCoord, so a lost result (node restart / planID reset) leaves those files unreferenced, and V3 GC only sweeps a segment's base path when the segment is dropped. It is independent of this PR's changes — the PR only routes more segments through the path, as you say. This belongs with the manifest-aware orphan cleanup the code already references (#51649); I'll make sure that tracking issue captures the lost-result rerun case explicitly rather than fixing it here.

internal/datanode/compactor/compactor_common.go
```diff
@@ -354,9 +341,16 @@ func compactionReadSchema(schema *schemapb.CollectionSchema, existingFields map[
...
+	return !field.GetIsFunctionOutput()
```
liliu-z, line 357: Keeping absent ordinary fields in the read schema hands mix compaction a reader-filled column whose contents differ per storage version, and mix compaction is the one path that does not normalize it. For manifest-backed (V2/V3) source segments the reader null-fills — the PR's own test pins that as intended, asserting an added field with DefaultValue 42 reads back as three NULLs — while the V1 binlog reader default-fills via GenerateEmptyArrayFromSchema. Mix compaction then splits: batches that filter at least one row go through RecordBuilder.Append, whose appendValueAt converts NULL to the declared default, but batches that filter zero rows hand the reader record straight to the writer and persist NULL. Before this PR the materializer default-filled the column for every path, so this is a behavior change, and the result is one output segment where the same added field holds defaults in some row ranges and NULLs in others depending on where deletes happened. Sort and merge-sort compaction always route through RecordBuilder, so they stay on the default-filling side and disagree with mix's pass-through. Pick one semantics (the bump path's NULL backfill or the RecordBuilder default fill) and apply it uniformly.

MrPresent-Han: Good catch — confirmed. With the read schema keeping the absent ordinary field and the packed (V2/V3) reader null-filling it, mix compaction does split: delete-carrying batches go through `RecordBuilder.Append` (whose `appendValueAt` turns NULL into the declared default), while zero-delete batches pass the reader record straight to the writer and persist NULL — so one segment ends up with the default in some row ranges and NULL in others.

This is fixed at the root by #52781: it makes the StorageV2/V3 packed reader default-aware — it reads only physically-present columns and fills every absent read-schema field via `GenerateEmptyArrayFromSchema` (declared default, else null), matching the V1 binlog reader. Once the reader emits the default instead of NULL, both mix paths agree: pass-through writes the default directly, and `RecordBuilder.Append` now sees a non-null value and just copies it — no intra-segment split, and consistent with V1 and with sort/merge-sort.

#52781 will be merged first, so this PR never lands the null-fill in isolation (and 3.0.1 is not released yet).

Issue #52406, opened 2026-08-11
[Bug]: Milvus 2.6 allows creating collections with text fields but fails to load them


yongpengli-z:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: 2.6
- Deployment mode(standalone or cluster): not specified
- MQ type(rocksmq, pulsar or kafka): not specified
- SDK version(e.g. pymilvus v2.0.0rc2): not specified
- OS(Ubuntu or CentOS): not specified
- CPU/Memory: not specified
- GPU: not specified
- Others: text field collection can be created successfully but cannot be loaded

### Current Behavior

Milvus 2.6 currently allows creating a collection that contains a text field. The create collection request succeeds and does not return an error.

However, the collection cannot be loaded afterwards.

For comparison, the 3.0 image returns a clear/correct prompt in the corresponding unsupported configuration scenario.

### Expected Behavior

Milvus 2.6 should provide a clear error/prompt when creating a collection with a text field if this scenario is not supported, instead of allowing the collection to be created and failing later during load.

The validation behavior should be consistent with the 3.0 image behavior in the corresponding unsupported configuration scenario.

### Steps To Reproduce

1. Start Milvus 2.6.
2. Create a collection with a text field.
3. Observe that collection creation succeeds without error.
4. Try to load the collection.
5. Observe that the load fails.

### Milvus Log

No response

### Anything else?

Note: storageV3 is only available in 3.0. The 3.0 image is mentioned here only as a behavior comparison because it gives a clear prompt for the corresponding unsupported configuration scenario.

PR #52485, opened 2026-08-13
fix: [2.6] reject unsupported Text data type

zhagnlu:
issue: #52406

### DIFF
--- a/internal/proxy/util.go
+++ b/internal/proxy/util.go
@@ -567,6 +567,8 @@ func validateFieldType(schema *schemapb.CollectionSchema) error {
 		switch field.GetDataType() {
 		case schemapb.DataType_String:
 			return merr.WrapErrParameterInvalidMsg("string data type not supported yet, please use VarChar type instead")
+		case schemapb.DataType_Text:
+			return merr.WrapErrParameterInvalidMsg("text data type is not supported in Milvus 2.6")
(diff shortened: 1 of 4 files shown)


internal/proxy/util.go
```diff
@@ -567,6 +567,8 @@ func validateFieldType(schema *schemapb.CollectionSchema) error {
 		switch field.GetDataType() {
 		case schemapb.DataType_String:
 			return merr.WrapErrParameterInvalidMsg("string data type not supported yet, please use VarChar type instead")
+		case schemapb.DataType_Text:
```
buqian-zilliz, line 570: **[P1] Reject TEXT before runtime function validation**

`createCollectionTask.PreExecute` calls `validateFunction` before it reaches this new field-type check. For a schema containing a TEXT input field and a TextEmbedding function, that path constructs the configured provider and `TextEmbeddingFunction.Check` calls `CallEmbedding`, so an unsupported schema can still issue a real outbound provider request and may return a credential, network, or provider error instead of the intended `text data type is not supported` error. Run the TEXT/type preflight immediately after unmarshalling the schema, before `validateFunction`, or otherwise ensure runtime function checks happen only after field types are accepted.

zhagnlu: fixed

zhengbuqian (approved):

PR #52491, opened 2026-08-13
feat: [SchEV2]enforce schema install fencing

congqixia:
Related to #51992

Close collection-scoped topology admission around schema-changing broadcasts so old-epoch load, balance, and distribution mutations cannot commit after the install fence.

Reject schema-changing DDL unless every registered service reports released 3.0.1 or later, recheck the contract before reopening admission, and keep QueryCoord unhealthy when initial distribution recovery fails. Protect delayed load callbacks with topology leases, preserve forward-only recovery for ambiguous broadcasts, and document the MixCoord-only Phase 0 contract.


chyezh:
Third-party code review of this PR (read-only review of the merge-base diff `2e6c8cd7..fac993fd`). Findings P1/P4/P5/P6 below; P2/P3 discussed separately. Not yet reproduced on a live cluster — happy to iterate.

---

### P1 (High) — balance tasks now inherit the request context; async `TransferSegment` silently no-ops

- **Files**: `internal/querycoordv2/handlers.go:140` (`task.NewSegmentTask(ctx, ...)`), `handlers.go:242` (`task.NewChannelTask(ctx, ...)`); callers `internal/querycoordv2/ops_services.go:440` (`TransferSegment`, `sync=false`) and `internal/querycoordv2/services.go:720` (`LoadBalance`, `sync=true`).
- **Problem**: this PR switches balance task creation from `s.ctx` (server-lifetime) to the caller's request `ctx`. grpc-go cancels the request context **synchronously when the unary handler returns** (verified in grpc-go v1.80.0: `processUnaryRPC` → `stream.WriteStatus` → `finishStream` → `s.cancel()`). `task.newBaseTask` derives `rootCtx` from the ctx passed at creation, and executors issue RPCs with `task.Context()`.
  - **`TransferSegment` (sync=false)**: the handler returns right after enqueuing the task → the request ctx is canceled → by the time the executor runs the task its ctx is already done → the migration RPC fails immediately, **the transfer never happens, yet the RPC returns success** (silent failure). Previously the task continued under `s.ctx`.
  - **`LoadBalance` (sync=true)**: a client timeout/disconnect now cancels in-flight balance tasks; previously they continued in the background (the handler's own `task.Wait` also fails on client timeout, but the task itself survived).
- **Why it matters**: `TransferSegment` is a public ops RPC; a silently dropped migration leaves data on the source node with no error surfaced.
- **Suggestion**: restore `s.ctx` (or derive with `context.WithoutCancel(ctx)`) for the async path; if `sync=true` is intentionally meant to cancel with the caller, state that explicitly and cover it with a test.
- **Test expectation**: after `TransferSegment`, assert via scheduler/dist that the segment eventually lands on the target node; for `LoadBalance`, simulate a client ctx timeout and assert the migration still completes (old semantics) or is explicitly canceled but never silently dropped.

### P4 (Low) — `getSyncDistributionMeta` adds 2 extra rootcoord RPCs per sync action

- **File**: `internal/querycoordv2/task/executor.go:730` (`getSyncDistributionMeta`: `DescribeCollection` + `DescribePartitions`), call sites `:574`, `:687`.
- **Problem**: every constructed SyncDistribution/LoadSegments request now triggers 2 rootcoord RPCs for schema + partition info. Under bulk load / large migrations this amplifies rootcoord load.
- **Suggestion**: the collection schema is stable over short windows — cache it (e.g., per task batch) or fetch on demand only when the node lacks the schema.
- **Test expectation**: an assertion on call counts for a full batch load.

### P5 (Low) — hardcoded version numbers in `ValidateCollectionSchemaBarrier` error

- **File**: `internal/querynodev2/segments/collection.go` (`WrapErrCollectionSchemaVersionNotReady(collectionID, 0, 1)`).
- **Problem**: the version numbers in the stale-schema error are hardcoded placeholders, unrelated to the actual `logicalVersion`/`barrierTs` — misleading when debugging.
- **Suggestion**: pass the real `logicalVersion`/`barrierTs`, or add a comment marking them as placeholders.

### P6 (Low/observation) — `UpdateSchema` cluster wrapper assumes non-nil `req.Base`

- **File**: `internal/querycoordv2/session/cluster.go` (`proto.Clone(req)` then `req.Base.TargetID = nodeID`).
- **Problem**: all current callers construct the request with a Base, so this does not trigger today; but as a public Cluster interface method, a future caller passing a nil Base would panic here.
- **Suggestion**: nil-guard after the clone, or document the non-nil Base precondition on the method.

PR #52496, opened 2026-08-13
enhance: [cp2.6]rewrite storage.MergeSort as a proper k-way merge

bigsheeper:
Cherry-pick from master

pr: #51998
issue: #51981

## Summary

Cherry-picked from master PR #51998 (merged): `storage.MergeSort` is rewritten as a proper k-way merge — the heap holds one entry per reader instead of every in-flight row, merge keys are resolved once per record instead of per comparison, and the per-row `pq.Enqueue(&index{...})` allocation is gone.

## Backport adaptations (2.6)

**1. Carries one type definition from #50817, which was never backported.**

`internal/storage/sort.go` gains these 7 lines:

```go
// rowIndex addresses a single row as (record index, row-in-record index). It is
// stored by value to avoid a per-row heap allocation.
type rowIndex struct {
	ri int32
	i  int32
}
```

`rowIndex` was introduced by #50817; the `rowHeap` that #51998 adds is typed over it (`items []rowIndex`, `push(v rowIndex)`, `pop() rowIndex`), so it cannot compile without this declaration. **Nothing else from #50817 is included** — its three `storage.Sort` optimizations (value slice instead of `[]*index`, pre-extracted sort keys, radix sort for a single int64 key) are NOT in this PR, and `Sort` is left exactly as it is on this branch. This PR is deliberately not tagged `pr: #50817`, because #50817 remains un-backported and should stay visible as such.

**2. `canMergeSort` drops the namespace branch.** 2.6 has no `GetIsSortedByNamespace`, so the master form

```go
namespaceEnabled := plan.GetSchema().GetEnableNamespace()
sortedByMergeKey := segment.GetIsSorted()
if namespaceEnabled { sortedByMergeKey = segment.GetIsSortedByNamespace() }
```

reduces to checking `GetIsSorted()` only. This is condition-for-condition equivalent to the inline `sortMergeAppicable` logic it replaces on this branch: same `UseMergeSort` gate, same per-segment sorted check, same `<= MaxSegmentMergeSort` bound.

**3. `TestCanMergeSortMatchesMergeKey` is dropped** — it exists solely to pin which sorted flag applies when namespace is enabled, which has no meaning on 2.6. `TestCanMergeSort` is kept and covers the rest.

**4. Logging uses this branch's `log` + `zap`** instead of master's `mlog` (not present on 2.6), in `mix_compactor.go` and `merge_sort.go`.

**5. `storage.PriorityQueue` is removed** along with the old merge loop, as on master — it had no other users in this package.

## Verification

- [x] File count matches original PR (6)
- [x] Conflicts resolved per-block by enclosing function: `Sort`-internal blocks kept at 2.6 side, `MergeSort` blocks taken from master
- [x] No conflict markers; no `pkg/v3` / `go-api/v3` paths; no unused imports
- [x] No dead code carried in from #50817 (`radixSortByInt64` and its test, `oneShotRecordReader` — all unused here — removed)

### DIFF
--- a/internal/datanode/compactor/merge_sort.go
+++ b/internal/datanode/compactor/merge_sort.go
@@ -120,6 +120,17 @@ func mergeSortMultipleSegments(ctx context.Context,
 	}
 
 	if _, err = storage.MergeSort(compactionParams.BinLogMaxSize, plan.GetSchema(), segmentReaders, writer, predicate, sortByFields); err != nil {
+		// segmentReaders[i] is built from binlogs[i], so the reader index the
+		// error names, when it names one, indexes into this list.
+		segmentIDs := make([]int64, len(binlogs))
+		for i, s := range binlogs {
+			segmentIDs[i] = s.GetSegmentID()
+		}
+		log.Warn("compact wrong, failed to merge sort segments",
+			zap.Int64("collectionID", collectionID),
+			zap.Int64s("segmentIDsByReaderIndex", segmentIDs),
+			zap.Int64s("sortByFields", sortByFields),
+			zap.Error(err))
 		if closeErr := writer.Close(); closeErr != nil {
 			log.Warn("failed to close writer after merge sort error", zap.Error(closeErr))
 		}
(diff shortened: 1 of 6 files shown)


czs007 (approved):
/lgtm

PR #52497, opened 2026-08-13
enhance: [proxy] remove global meta cache initialization

congqixia:
Related to #44761

Initialize the proxy meta cache locally and pass it through the Proxy instance and tasks instead of relying on a package-level singleton. This makes cache ownership explicit, removes InitMetaCache, and lets tests inject cache state directly.

Update test helpers and callers to use initMetaCache and the returned cache value, including the test-only baseTask cache hook where needed.

### DIFF
--- a/internal/distributed/proxy/service.go
+++ b/internal/distributed/proxy/service.go
@@ -311,7 +311,7 @@ func (s *Server) startExternalGrpc(errChan chan error) {
 			proxy.DatabaseInterceptor(),
 			UnaryRequestStatsInterceptor,
 			accesslog.UnaryAccessLogInterceptor,
-			proxy.GrpcAuthInterceptor(proxy.AuthenticationInterceptor),
+			proxy.GrpcAuthInterceptor(proxy.AuthenticationInterceptorWithMetaCache(getMetaCache)),
 			proxy.UnaryServerInterceptor(proxy.PrivilegeInterceptorWithMetaCache(getMetaCache)),
 			proxy.UnaryServerHookInterceptor(),
 			mlog.UnaryServerInterceptor(typeutil.ProxyRole),
--- a/internal/distributed/proxy/service.go
+++ b/internal/distributed/proxy/service.go
@@ -290,6 +290,11 @@ func (s *Server) startExternalGrpc(errChan chan error) {
 	}
 	mlog.Debug(context.TODO(), "Get proxy rate limiter done")
 
+	var metaCache proxy.Cache
+	if metaCacheProvider, ok := s.proxy.(interface{ GetMetaCache() proxy.Cache }); ok {
+		metaCache = metaCacheProvider.GetMetaCache()
+	}
+
 	var unaryServerOption grpc.ServerOption
 	if enableCustomInterceptor {
 		unaryServerOption = grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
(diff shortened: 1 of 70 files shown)


internal/proxy/privilege_interceptor.go
```diff
@@ -55,158 +55,164 @@ func UnaryServerInterceptor(privilegeFunc PrivilegeFunc) grpc.UnaryServerInterce
...
-	if util.GetPrivilegeLevel(util.MetaStore2API(objectPrivilege)) == milvuspb.PrivilegeLevel_Cluster.String() {
-		dbName = util.AnyWord
-	}
-	// RenameCollection is a database-admin privilege: a same-db rename is
-	// authorized against the target db (database level, handled above), while a
-	// cross-db rename additionally requires a cluster-scoped (global) grant.
-	if r, ok := req.(*milvuspb.RenameCollectionRequest); ok && r.GetDbName() != r.GetNewDBName() {
-		dbName = util.AnyWord
-	}
+	return PrivilegeInterceptorWithMetaCache(func() Cache { return nil })(ctx, req)
```
weiliu1031, line 58: [P1] Wire the REST authorization paths to the injected MetaCache

This wrapper always supplies a nil MetaCache. REST v1 and v2 still call `PrivilegeInterceptor` directly, so collection alias resolution always fails and authorization falls back to the alias string.

For example, a role granted `Query` on collection `orders` will be denied when accessing it through alias `orders_current`. If a grant happens to exist for the alias string, authorization may instead be evaluated against an identity that differs from the collection ultimately accessed.

Please pass the handlers' `metaCache` getter into the authorization helpers and invoke `PrivilegeInterceptorWithMetaCache(getMetaCache)`, as the gRPC path does. It would also be useful to cover this with REST v1/v2 alias RBAC tests.

internal/distributed/proxy/service.go
```diff
@@ -290,6 +290,20 @@ func (s *Server) startExternalGrpc(errChan chan error) {
...
 	mlog.Debug(context.TODO(), "Get proxy rate limiter done")
 
+	// The meta cache is initialized in Proxy.Init(), which runs after this interceptor
+	// chain is constructed. Resolve it per request so rate limiting / privilege checks
+	// observe the live cache instead of a construction-time nil snapshot.
+	var metaCacheProvider interface{ GetMetaCache() proxy.Cache }
+	if provider, ok := s.proxy.(interface{ GetMetaCache() proxy.Cache }); ok {
+		metaCacheProvider = provider
+	}
+	getMetaCache := func() proxy.Cache {
```
weiliu1031, line 300: [P1] Do not serve requests before the caches are safely published

The external gRPC server starts serving before `Proxy.Init()` initializes the MetaCache and PrivilegeCache. Since this PR removes the previous authentication readiness check, a username/password request in that window can reach `passwordVerify()` while `privilege.GetPrivilegeCache()` is nil and panic when calling `GetCredentialInfo()`.

The new getter also reads `node.metaCache` concurrently with the unsynchronized assignment in `Proxy.Init()`, creating a data race.

Please either start serving only after `Proxy.Init()` completes, or preserve an explicit readiness gate that returns `ServiceUnavailable` and publish/read the MetaCache through an atomic or lock-protected mechanism.

internal/proxy/meta_cache_testonly.go
```diff
@@ -83,3 +89,10 @@ func mustNewMetaCacheWithDBInfoForTest(mixCoord types.MixCoordClient, dbInfo map
 	}
 	return cache
 }
+
+func mockBaseTaskMetaCacheForTest(cache Cache) func() {
+	patch := mockey.Mock((*baseTask).getMetaCache).Return(cache).Build()
```
weiliu1031: [P2] Inject the MetaCache into each test task instead of patching the method

This patch replaces `baseTask.getMetaCache()` globally, allowing tests that construct a task without initializing `baseTask.metaCache` to pass. As a result, the tests cannot detect missing MetaCache injection in production constructors, and concurrent tests may observe another test's patched cache.

Please initialize `baseTask{metaCache: cache}` in the relevant fixtures and remove this global method patch. That keeps the tests aligned with the dependency injection contract introduced by this PR.

weiliu1031 (commented):
Found two correctness blockers and one test-isolation issue; details are provided inline.

PR #52498, opened 2026-08-13
fix: support vector output with order_by

### LINKED ISSUE #52463: [Bug]: Query ORDER BY fails when a vector field is included in output_fields
## Environment

- **Milvus Version**: `3.0-20260810-e2820e273c` (commit `e2820e273c`)
- **Deployment Mode**: cluster
- **MQ**: Woodpecker
- **SDK**: PyMilvus `3.0.1rc48`
- **OS**: Linux (server), macOS (client)

### Option A: Script

```python
from pymilvus import DataType, MilvusClient


client = MilvusClient(uri="http://localhost:19530")
collection_name = "query_orderby_vector_output_repro"

if client.has_collection(collection_name):
    client.drop_collection(collection_name)

schema = client.create_schema(auto_id=False, enable_dynamic_field=False)
schema.add_field("id", DataType.INT64, is_primary=True)
schema.add_field("vector", DataType.FLOAT_VECTOR, dim=4)

index_params = client.prepare_index_params()
index_params.add_index(
    field_name="vector",
    index_name="vector_index",
    index_type="AUTOINDEX",
    metric_type="L2",
)
client.create_collection(
    collection_name=collection_name,
    schema=schema,
    index_params=index_params,
    consistency_level="Strong",
)

client.insert(
    collection_name=collection_name,
    data=[
        {"id": i, "vector": [float(i), 0.0, 0.0, 0.0]}
        for i in range(10)
    ],
)
client.load_collection(collection_name)

common = {
    "collection_name": collection_name,
    "filter": "id >= 0",
    "limit": 10,
    "consistency_level": "Strong",
}

# Control 1: scalar output plus scalar ORDER BY succeeds.
asc = client.query(
    **common,
    output_fields=["id"],
    order_by_fields=["id:asc"],
)
assert [row["id"] for row in asc] == list(range(10))

# Control 2: vector output without ORDER BY succeeds.
with_vector = client.query(**common, output_fields=["id", "vector"])
assert len(with_vector) == 10
assert all("vector" in row for row in with_vector)

# Reproduction: ordering by scalar id fails only because vector is projected.
client.query(
    **common,
    output_fields=["id", "vector"],
    order_by_fields=["id:asc"],
)
```

The same failure occurs when the final request uses
`output_fields=["*"]`, because the wildcard expands to both `id` and
`vector` for this schema.

### Trigger Conditions

- **Frequency**: always
- **First observed after**: immediately on the query
- **Does NOT happen when**:
  - `output_fields=["id"]` is used with the same scalar `id:asc` ordering.
  - The vector field is returned without `order_by_fields`.

## Expected Behavior

Query should sort by the scalar `id` field and return both requested fields.
The IDs should be `[0, 1, ..., 9]`, and each row should contain its
`FloatVector` value.

The vector is an output field only; it is not being used as an ORDER BY key.

## Actual Behavior

The Query fails whenever a `FloatVector` field is projected together with
scalar Query ORDER BY. Both explicit vector output and wildcard output fail:

```python
output_fields=["id", "vector"], order_by_fields=["id:asc"]
output_fields=["*"],            order_by_fields=["id:asc"]
```

## Error Logs

```text
Operator::GetOutput failed for [Operator:Project Operator]
unsupported data type VECTOR_FLOAT
at internal/core/src/segcore/Utils.cpp:1712
segcore error[segcoreCode=2001]
```

The SDK exposes the request failure as Milvus error code `2000`.

## Non-default Configuration

No relevant non-default configuration is required.

## Analysis Hints

- This is not the documented restriction that vector fields cannot be used
  as ORDER BY keys. The ORDER BY key in this reproduction is the scalar
  primary key `id`; only the projected output contains a vector.
- The failure is isolated to the Query ORDER BY projection path. Vector
  projection works without ORDER BY, and scalar projection works with ORDER
  BY.
- The error points to the segcore Project Operator and
  `internal/core/src/segcore/Utils.cpp:1712`.
- The current PyMilvus Query ORDER BY E2E suite covers scalar output fields
  and separately covers normal query vector projection, but does not combine
  `order_by_fields` with a vector output field or wildcard output.

MrPresent-Han:
## Summary

- defer vector output fields for order_by plans that require late materialization
- preserve vector field IDs and dimensions through the result pipeline
- add a planner and retrieval regression test

## Root cause

ProjectNode cannot materialize vector columns before TopK, so valid order_by requests with vector output failed during projection.

## Validation

- `make build-cpp-with-unittest` (incremental, max 6 jobs)
- `internal/core/output/unittest/all_tests --gtest_filter="*QueryOrderBy*"` (29/29 passed)
- `clang-format --dry-run --Werror`
- `git diff --check`

Fixes #52463

### DIFF
--- a/internal/core/src/query/PlanProto.cpp
+++ b/internal/core/src/query/PlanProto.cpp
@@ -479,8 +479,8 @@ BuildProjectAndAggregationNodes(
 // Helper function to build ProjectNode for ORDER BY queries.
 // Returns {ProjectNode, deferred_field_ids, pipeline_field_ids}.
 // deferred_field_ids is empty for single-project mode (all columns materialized
-// in the first project), or non-empty for two-project mode (variable-width
-// non-sort output columns deferred until after TopK).
+// in the first project), or non-empty for two-project mode (variable-width or
+// vector non-sort output columns deferred until after TopK).
 // pipeline_field_ids mirrors project_ids so FillOrderByResult can stamp
 // the correct field_id on each DataArray produced by the pipeline.
 std::tuple<plan::PlanNodePtr, std::vector<FieldId>, std::vector<FieldId>>
@@ -518,11 +518,13 @@ BuildOrderByProjectNode(const proto::plan::QueryPlanNode& query,
         }
     }
 
-    // Collect non-sort output fields and check for variable-width types.
+    // Collect non-sort output fields and check whether any must be late
+    // materialized. ProjectNode only materializes scalar columns, so vector
+    // outputs must be deferred just like variable-width outputs.
     // Skip system fields (RowFieldID, TimestampFieldID) — they are handled
     // separately in FillTargetEntry and must not enter the pipeline.
     std::vector<FieldId> non_sort_output_fields;
-    bool has_variable_width = false;
+    bool requires_late_materialization = false;
     for (auto fid_raw : plan_node_proto.output_field_ids()) {
         if (seen_field_ids.count(fid_raw) == 0) {
             auto fid = FieldId(fid_raw);
@@ -530,14 +532,16 @@ BuildOrderByProjectNode(const proto::plan::QueryPlanNode& query,
                 continue;
             }
             non_sort_output_fields.push_back(fid);
-            if (IsVariableDataType(schema->GetFieldType(fid))) {
-                has_variable_width = true;
+            auto field_type = schema->GetFieldType(fid);
+            if (IsVariableDataType(field_type) ||
+                IsVectorDataType(field_type)) {
+                requires_late_materialization = true;
             }
         }
     }
 
     std::vector<FieldId> deferred_field_ids;
-    if (has_variable_width) {
+    if (requires_late_materialization) {
         // Two-project mode: defer ALL non-sort output fields until after TopK.
         deferred_field_ids = non_sort_output_fields;
     } else {
--- a/internal/core/unittest/test_query_order_by.cpp
+++ b/internal/core/unittest/test_query_order_by.cpp
@@ -144,10 +144,10 @@ class QueryOrderByTest : public testing::TestWithParam<bool> {
         schema_->set_primary_field_id(str_fid);
 
         num_rows_ = 20;
-        auto raw_data =
+        raw_data_ =
             DataGen(schema_, num_rows_, 42, 0, 1, 10, false, false, false);
 
-        auto segment = CreateSealedWithFieldDataLoaded(schema_, raw_data);
+        auto segment = CreateSealedWithFieldDataLoaded(schema_, raw_data_);
         segment_ = SegmentSealedSPtr(segment.release());
 
         milvus::exec::expression::FunctionFactory& factory =
@@ -283,6 +283,7 @@ class QueryOrderByTest : public testing::TestWithParam<bool> {
     SegmentSealedSPtr segment_;
     std::shared_ptr<Schema> schema_;
     std::map<std::string, FieldId> field_map_;
+    GeneratedData raw_data_;
 };
 
 INSTANTIATE_TEST_SUITE_P(QueryOrderBySuite,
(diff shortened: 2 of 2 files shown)


internal/core/src/query/PlanProto.cpp
```diff
@@ -518,26 +518,30 @@ BuildOrderByProjectNode(const proto::plan::QueryPlanNode& query,
...
         if (seen_field_ids.count(fid_raw) == 0) {
             auto fid = FieldId(fid_raw);
             if (SystemProperty::Instance().IsSystem(fid)) {
                 continue;
             }
             non_sort_output_fields.push_back(fid);
-            if (IsVariableDataType(schema->GetFieldType(fid))) {
-                has_variable_width = true;
+            auto field_type = schema->GetFieldType(fid);
+            if (IsVariableDataType(field_type) ||
```
liliu-z, line 536: Routing dense vectors into the deferred path makes every segment materialize full vector payloads for its own segment-local top-K, and the QueryNode reduce charges all of those candidate rows against maxOutputSize before the global top-K is applied. DeduplicatePKOperator accumulates retSize over every row of every per-segment result and fails at dedup_pk_op.go:127, while OrderByLimitOperator — the stage that truncates to topK — only runs afterwards (pipeline_builders.go:71-72). With 20 segments, limit=1000 and 1536-dim float vectors that is ~117 MiB charged against the 100 MiB default while the returned result is only ~5.9 MiB, so a legitimate order-by query is rejected outright; the usual escape hatch does not apply either, since ShouldIgnoreNonPk returns false for order-by plans (plan_c.cpp:220-223), meaning two-phase retrieval never kicks in to keep vectors off the wire. Note that an early break at limit inside DeduplicatePKOperator (as reduce_by_pk_op.go:300-303 does) is not a valid fix, because dedup runs before global ordering and would drop rows that should win; the size check has to move after OrderByLimitOperator truncates, or deferred fields need to stay as offsets until after global selection.

internal/core/unittest/test_query_order_by.cpp
```diff
@@ -583,30 +583,44 @@ TEST_P(QueryOrderByTest, OrderByWithDeferredFields) {
...
-        EXPECT_NE(std::string(e.what()).find("unsupported data type"),
-                  std::string::npos);
+    proto::plan::PlanNode plan_node;
+    plan_node.add_output_field_ids(pk_fid.get());
+    plan_node.add_output_field_ids(vector_fid.get());
+
+    auto* query = plan_node.mutable_query();
+    query->set_limit(10);
+    auto* order_by = query->add_order_by_fields();
+    order_by->set_field_id(pk_fid.get());
```
liliu-z: The new test sorts on the VARCHAR primary key, but DataGen sorts VARCHAR columns before inserting them, so the fixture's PK column is already physically ascending and the segment offsets handed to the deferred bulk_subscript are the identity 0..9. The vector assertions (field_id, dim, element count) therefore pass even if those offsets were ignored or replaced by a plain iota buffer — no permutation of the deferred fetch is ever exercised. Ordering by the fixture's existing non-PK double_field and asserting the returned vector payloads match the rows at the permuted offsets would actually cover the late-materialization path this PR enables.

internal/core/src/query/PlanProto.cpp
MrPresent-Han: You're right, and thanks for the precise trace — confirmed on our side. This PR only moves vector materialization to after the segment-local top-K, so QN-level DeduplicatePK still charges #segments × topK vector rows against maxOutputSize before OrderByLimit truncates (and ShouldIgnoreNonPk excludes order-by plans, so two-phase retrieval never engages). We'll fix this in a follow-up PR taking your second suggested direction: keep non-order-by output fields as {segment, offset} selections through QN-level dedup + top-K selection, and bulk-fetch them only afterwards (reusing the RetrieveByOffsets machinery of the IgnoreNonPk path), which also moves the size check to after selection. Agreed that an early break inside dedup is not a valid fix since it runs before global ordering. Keeping the current PR scoped to routing vector outputs into the existing late-materialization path.

internal/core/unittest/test_query_order_by.cpp
MrPresent-Han: Good catch — fixed in 8940c42f12. The test now sorts on the random double column (DataGen pre-sorts VARCHAR as you noted, which made the offsets the identity), verifies each returned deferred vector payload element-wise against the original row at its permuted segment offset (via PK → original-offset mapping), and additionally asserts the selection is not the identity permutation so the reordering path is guaranteed to be exercised. Passes for both nullable variants (29/29 QueryOrderBy tests green locally).

MrPresent-Han: (Branch squashed to a single commit per convention — the strengthened test is now part of ecd118fa49.)

liliu-z (commented):
/lgtm
/approve

Issue #52501, opened 2026-08-13
[Enhancement]: Configure External Table read IOPS request rates


weiliu1031:
## Problem

External Table reads use the milvus-storage/Lance AIMD ObjectStore request-rate limiter. Milvus needs configuration for the initial and maximum request rates used by this read path.

## Scope

This issue is limited to configuring IOPS request rates for External Table reads.

## Configuration

Add process-level settings under `common.storage.iops`:

```yaml
common:
  storage:
    iops:
      initialRate: 2000
      maxRate: 5000
```

- The defaults match milvus-storage: `initialRate: 2000` and `maxRate: 5000`.
- `initialRate` must be greater than zero. An initial rate above a positive maximum is capped at that maximum by the limiter.
- `maxRate: 0` removes the AIMD rate ceiling; it does not disable the AIMD limiter.
- These settings tune individual Lance ObjectStore limiters, not a strict aggregate IOPS cap across all filesystem instances.
- These are startup configuration settings, not runtime-refreshable controls.

## Integration and behavior

- Apply the Milvus-configured policy to the External Table filesystem properties `extfs.<collectionID>.iops_initial_rate` and `extfs.<collectionID>.iops_max_rate`.
- Pass Go FFI values directly to property injection without changing the process singleton.
- Initialize the native Segcore and IndexBuilder External Table read policy once during component startup.
- External specifications cannot override the Milvus-configured rates.
- Preserve credential isolation when applying the IOPS policy to the per-collection external filesystem namespace.

## Implementation status (2026-09-14)

- [Master PR 52502](https://github.com/milvus-io/milvus/pull/52502) merged on 2026-08-21.
- [3.0 backport PR 52503](https://github.com/milvus-io/milvus/pull/52503) merged on 2026-08-21.

The agreed External Table read scope is implemented by these merged PRs. The earlier broader integration proposal and its remaining-work list have been superseded by this scope; no additional implementation work is tracked here.

Issue #618, opened 2019-11-29
building index time cost is different between datasets 


del-zhenwu:
**Describe the bug**
row_count: `50m`
index_type: `ivf_sq8`
building index with CPU-version on the same server host
dataset: `sift-2048-128d`, building each file to index file cost: `1500s`
dataset: `random-2048-512d`, building each file to index file cost: `700s`

**Steps/Code to reproduce behavior**

**Expected behavior**

**Environment details**
0.5.0-CPU-version

**Screenshots**

**Additional context**

PR #52502, opened 2026-08-13
enhance: configure External Table read IOPS rates

weiliu1031:
issue: #52501

## What changed

- Add `common.storage.iops.initialRate` and
  `common.storage.iops.maxRate`, defaulting to the milvus-storage values
  `2000` and `5000` and marked as introduced in 3.0.1.
- Pass Go FFI rates directly to External Table property injection without
  mutating the process singleton.
- Initialize the native Segcore and IndexBuilder policies once from their
  component startup paths, outside Reader configuration watchers.
- Write the policy only to
  `extfs.<collectionID>.iops_initial_rate/max_rate`, using canonical
  milvus-storage property constants.
- Add C ABI regression coverage proving that per-call injection does not
  change the singleton policy.

## Behavior boundary

This change is limited to the External Table read path. It does not propagate
IOPS fields through `StorageConfig`, internal `fs.*` properties, writers,
compaction, or filesystem metrics. External specifications cannot override the
Milvus-configured rates.

`maxRate: 0` disables the AIMD rate ceiling. These settings tune each Lance
ObjectStore limiter; they are not a strict aggregate IOPS limit across all
filesystem instances.

The current master base already uses milvus-storage `8632a0f`, which contains
the required IOPS properties and Lance AIMD mappings.

## Validation

- Compilation and tests were not rerun for the cleanup, per request
- `gofmt -d` on changed Go files
- `git diff --check`
- Rebased onto `master` at `ea48d3dd4b2b0f1177c7780538a647e85a9e6d45`
  with unchanged feature patch-id and touched paths

Dependency: [milvus-storage PR #618](https://github.com/milvus-io/milvus-storage/pull/618)

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -1117,6 +1117,15 @@ common:
         threshold: 1024 # split by average size policy threshold(in bytes) in storage v2
     useLoonFFI: false
     enableGrowingSourceFlush: false # enable flushing growing segment payload from QueryNode growing source through StorageV3 manifest path
+    iops:
+      # Initial ObjectStore request rate used by the Lance AIMD limiter for External Table reads.
+      # The value must be greater than 0. Values above a positive maxRate are reduced to maxRate.
+      # The default matches the milvus-storage default.
+      initialRate: 2000
+      # Maximum request rate allowed by the Lance AIMD limiter for External Table reads.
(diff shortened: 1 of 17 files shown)


congqixia (approved):
/approve

PR #52503, opened 2026-08-13
enhance: [3.0] configure External Table read IOPS rates

weiliu1031:
issue: #52501
pr: #52502

## Summary

Backport the minimal External Table read IOPS configuration from the master
PR. The backport:

- exposes `common.storage.iops.initialRate` and
  `common.storage.iops.maxRate` as 3.0.1 configuration;
- passes the Go FFI values directly to property injection without mutating
  process-global state;
- initializes the native Segcore and IndexBuilder policies once from their
  component startup paths; and
- stamps the rates only into the per-collection External Table filesystem
  properties.

## Behavior boundary

The rates are written only as
`extfs.<collectionID>.iops_initial_rate/max_rate`. They are not propagated
through `StorageConfig`, internal `fs.*` properties, writers, compaction, or
filesystem metrics. External specifications cannot override the configured
rates.

`maxRate: 0` disables the AIMD rate ceiling. The settings are not a strict
aggregate IOPS limit across all filesystem instances.

## Conflict resolution

Rebased onto the current 3.0 branch. The only conflict was the
milvus-storage revision: 3.0 already uses the newer `8632a0f`, which retains
the `fs.iops_*` properties and Lance AIMD mappings, so the newer target-branch
revision was preserved.

The 3.0 branch has different test insertion points from master. The resolution
preserves the 3.0 test layout while keeping the same production ownership
boundary and C ABI singleton-isolation regression coverage.

## Validation

- Compilation and tests were not run, per request
- `gofmt -d` on changed Go files
- `git diff --check`
- Rebased onto `3.0` at `95db0b5329bb90fe00c57614e8b0fd4a06f7aff0`

Dependency: [milvus-storage PR #618](https://github.com/milvus-io/milvus-storage/pull/618)

### DIFF
--- a/configs/milvus.yaml
+++ b/configs/milvus.yaml
@@ -1086,6 +1086,15 @@ common:
         threshold: 1024 # split by average size policy threshold(in bytes) in storage v2
     useLoonFFI: false
     enableGrowingSourceFlush: false # enable flushing growing segment payload from QueryNode growing source through StorageV3 manifest path
+    iops:
+      # Initial ObjectStore request rate used by the Lance AIMD limiter for External Table reads.
+      # The value must be greater than 0. Values above a positive maxRate are reduced to maxRate.
+      # The default matches the milvus-storage default.
+      initialRate: 2000
+      # Maximum request rate allowed by the Lance AIMD limiter for External Table reads.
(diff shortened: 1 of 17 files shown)


congqixia (approved):
/approve

Issue #52511, opened 2026-08-13
[Bug]: FlushGrowingSegmentTest leaks segcore chunk_rows=1 globally, making later growing-segment tests 20-555x slower


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```
- Milvus version: master @ e6b52c9e10
- Reproduced on: x86_64 (AMD Ryzen 9 7945HX, Ubuntu 24.04) and aarch64 (NVIDIA DGX GB10)
- Build: Release (CMAKE_BUILD_TYPE=Release, USE_ASAN=OFF)
- Component: internal/core C++ unit tests (all_tests)
```

### Current Behavior

`FlushGrowingSegmentTest.FlushNullableEmbListMixedRows` permanently changes the **process-global**
segcore `chunk_rows` to `1` and never restores it. Every growing segment created afterwards in the
same `all_tests` process then allocates one chunk per row.

The test itself still passes, so CI stays green — only the wall time of later tests explodes.

`internal/core/src/segcore/flush_growing_segment_test.cpp:2988-2989`

```cpp
SegcoreConfig segcore_config;        // looks like a local object
segcore_config.set_chunk_rows(1);    // actually writes process-global state, never restored
auto segment =
    CreateGrowingSegment(schema, empty_index_meta, 1, segcore_config);
```

This looks safe because `segcore_config` is a local variable, but every member of `SegcoreConfig`
is `inline static` — `internal/core/src/segcore/SegcoreConfig.h:258`:

```cpp
inline static int64_t chunk_rows_ = 32 * 1024;
```

so a setter called on *any* instance writes the global value:

```cpp
void set_chunk_rows(int64_t chunk_rows) { chunk_rows_ = chunk_rows; }   // SegcoreConfig.h:41-43
```

A later test that inserts 100k rows into a growing segment therefore allocates
`100000 / 1 = 100000` chunks instead of `100000 / 32768 ≈ 4`.

### Impact (measured)

Measured with `Query.InnerProduct` (inserts `N = 100000` rows into a growing segment), same binary,
same machine, only the preceding filter differs:

| Platform | `Query.InnerProduct` alone | after the polluting test | slowdown |
| --- | --- | --- | --- |
| x86_64 | 113 ms | **2326 ms** | 20.6x |
| aarch64 | 52 ms | **28900 ms** | 555x |

In a full nightly-style run (430 tests / 47 suites selected by a change-based filter), the two tests
[...]

PR #52512, opened 2026-08-13
fix: make SegcoreConfig members instance members

czs007:
issue: #52511

### What

Every member of `SegcoreConfig` was declared `inline static`, so the class looked like a value
type but was not one — calling a setter on a *local* instance wrote process-global state:

```cpp
SegcoreConfig segcore_config;        // looks local
segcore_config.set_chunk_rows(1);    // writes the process-global value
```

This PR makes the 23 configuration members real instance members, so a local `SegcoreConfig` is
genuinely local and cannot leak.

To keep the "I really do mean the process-wide one" sites explicit, the accessor is split:

- `default_config()` now returns `const SegcoreConfig&`
- `mutable_default_config()` returns the writable process-wide instance

Any site that used to mutate the global through `default_config()` is now a **compile error**
until it says `mutable_default_config()`. That is why the test files below are touched: they
intentionally configure the process-wide instance, and now have to say so.

`valid_dense_vector_index_type` stays `inline static const` — it is a read-only constant with no
leak risk.

### Why it matters

`FlushGrowingSegmentTest.FlushNullableEmbListMixedRows` sets `chunk_rows` to `1` through a local
`SegcoreConfig` and never restores it, so the value stayed at `1` for the rest of the `all_tests`
process and every growing segment created afterwards allocated one chunk per row.

The test itself keeps passing, so CI stays green — only the wall time of *later* tests grows.
Measured with `Query.InnerProduct` (inserts 100k rows into a growing segment), same binary, same
machine, only the preceding filter differs:

| Platform | alone | after the polluting test | slowdown |
| --- | --- | --- | --- |
| x86_64 (Ryzen 9 7945HX) | 113 ms | 2326 ms | 20.6x |
| aarch64 (DGX GB10) | 52 ms | 28900 ms | 555x |

In a 430-test / 47-suite run selected by a change-based filter, `Query.InnerProduct` and
`Query.ExecTerm` together took **57.7 s of the 221 s** the whole gtest stage needed — 26 % of the
stage for two tests that need ~100 ms in isolation.

### Scope

Production code affected: **only `segcore_init_c.cpp`**, the init-time C API driven once from
`InitQueryNode`. Its 20 call sites now use `mutable_default_config()`, so process-wide behaviour
is unchanged. The remaining 13 files are tests that configure the global on purpose.

### Verification

Built and run locally (x86_64, Release):

- `Query.InnerProduct` after the polluting test: **2326 ms -> 100 ms** (95 ms when run alone),
  **with no guard added to the test** — the type change alone removes the pollution
- `FlushNullableEmbListMixedRows` itself still passes
- full `all_tests`: no new failures

### Note on the previous revision of this PR

This PR originally added a one-line `ScopedSegcoreConfigRestore` to the single call site that
happened to be noticed. That fixed the symptom at one place and left the trap in the type. A scan
of the whole C++ test tree found the same "local instance, global effect" shape elsewhere, so the
fix moved to the type itself.

### One more leak found while auditing

Splitting the accessor made every process-global write greppable, and one more test turned up:
`SearchResultExport.GlobalRefine_EndToEnd_ForcedSealedRefine`
(`search_result_export_c_test.cpp:1766`) sets `enable_interim_segment_index`, `nlist` and
`chunk_rows` and never restores them.

That one is a *deliberate* global write rather than the accidental "local instance" form fixed
above, so the type change does not help it — it needs the guard the rest of the tree already uses.
It is included here because it is the same defect class as #52511. Unlike the original, it has no
measurable effect on later test durations (`chunk_rows` 32768 -> 1024 is a 32x change, not the
32768x of `chunk_rows = 1`); the problem is that `enable_interim_segment_index = true` stays on for
every later test in the process.

### DIFF
--- a/internal/core/src/exec/expression/GISCoarseRefineExprTest.cpp
+++ b/internal/core/src/exec/expression/GISCoarseRefineExprTest.cpp
@@ -68,10 +68,10 @@ RunFilter(const std::shared_ptr<Schema>& schema,
 // RAII guard so the global segcore flag is always restored, even on failure.
 struct GISSplitFusionGuard {
     explicit GISSplitFusionGuard(bool enable) {
-        SegcoreConfig::default_config().set_enable_gis_split_fusion(enable);
+        SegcoreConfig::mutable_default_config().set_enable_gis_split_fusion(enable);
     }
     ~GISSplitFusionGuard() {
-        SegcoreConfig::default_config().set_enable_gis_split_fusion(false);
+        SegcoreConfig::mutable_default_config().set_enable_gis_split_fusion(false);
     }
 };
 
@@ -106,10 +106,10 @@ struct GISGroupStateCapture {
 // (ChunkedSegmentSealedImpl::LoadFieldData).
 struct GeometryCacheGuard {
     explicit GeometryCacheGuard(bool enable) {
-        SegcoreConfig::default_config().set_enable_geometry_cache(enable);
+        SegcoreConfig::mutable_default_config().set_enable_geometry_cache(enable);
     }
     ~GeometryCacheGuard() {
-        SegcoreConfig::default_config().set_enable_geometry_cache(false);
+        SegcoreConfig::mutable_default_config().set_enable_geometry_cache(false);
     }
 };
 
--- a/internal/core/src/index/RTreeIndexTest.cpp
+++ b/internal/core/src/index/RTreeIndexTest.cpp
@@ -905,11 +905,11 @@ namespace {
 // flag into later tests).
 struct GisSplitFusionFlagGuard {
     explicit GisSplitFusionFlagGuard(bool enable) {
-        milvus::segcore::SegcoreConfig::default_config()
+        milvus::segcore::SegcoreConfig::mutable_default_config()
             .set_enable_gis_split_fusion(enable);
     }
     ~GisSplitFusionFlagGuard() {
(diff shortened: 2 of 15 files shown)


xiaofan-luan:
The overall direction makes sense, but one group of call sites still relies on the old static-member semantics.

`ChunkedSegmentSealedImpl` copies `SegcoreConfig` in its constructor. In both `AccuracyWithLoadFieldData` and `AccuracyWithMapFieldData`, the segment is created before `ApplyBinlogInterimIndexConfigForTest()` mutates the process-wide default. Now that the members are per-instance, the existing segment keeps the old snapshot (notably `enable_interim_segment_index=false`), so `LoadVectorField()` does not build the interim index.

This matches the CPP UT result exactly: all 36 failures are the parameterized `BinlogIndexTest.AccuracyWith{Load,Map}FieldData` cases with `segment->HasIndex(vec_field_id)` actual=false / expected=true.

Please configure an instance before constructing the segment and pass it explicitly, e.g.:

```cpp
auto config = SegcoreConfig::default_config();
ApplyInterimIndexConfigForTest(options, config);
segment = CreateSealedSegment(schema, collection_index_meta, 0, config);
```

The same ordering issue should also be fixed in `DisableInterimIndex` and `LoadBinlogWithoutIndexMeta`: they call `SegcoreSetEnableInterminSegmentIndex` after segment construction, so those calls no longer affect the segment under test.

PR #52529, opened 2026-08-14
enhance: compact Tantivy JSON row masks

sparknack:
issue: #52565

- Store Tantivy null rows and typed JSON path non-exist rows as CRoaring after build and load.
- Retain one resident N/8 EXISTS bitmap so the default non-cached query path only clones it.
- Preserve legacy `size_t[]` entries and scalar index versions while accounting for resident row masks in memory estimates.

### DIFF
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -195,31 +195,27 @@ ScalarRowMaskResidentBytes(
     const auto is_json_path =
         field_type == DataType::JSON &&
         index_params.find(JSON_PATH) != index_params.end();
-    uint64_t mask_count = 0;
-    uint64_t dense_bitmap_count = 0;
-    if ((index_type == INVERTED_INDEX_TYPE || index_type == NGRAM_INDEX_TYPE) &&
-        (is_json_path || field_nullable.value_or(true))) {
-        // Tantivy does not retain null rows, so the C++ wrapper keeps one
-        // CRoaring row set beside the mmap'd index. JSON path extraction can
-        // produce null rows even when the source JSON field is non-nullable.
-        mask_count = 1;
-    }
+    auto cast_type_it = index_params.find(JSON_CAST_TYPE);
+    const bool is_flat_json =
+        cast_type_it == index_params.end() || cast_type_it->second == "JSON";
+    // Typed JSON path indexes additionally keep the non-existing rows for
+    // EXISTS semantics (plus one dense exists bitmap). NGRAM never applies.
+    const bool typed_json_path =
+        is_json_path && !is_flat_json && index_type != NGRAM_INDEX_TYPE;
 
-    if (is_json_path && index_type != NGRAM_INDEX_TYPE) {
-        // Typed JSON path indexes additionally keep the non-existing rows for
-        // EXISTS semantics. An unresolved HYBRID may select INVERTED, so use
-        // two masks there as a conservative bound.
-        auto cast_type_it = index_params.find(JSON_CAST_TYPE);
-        auto is_flat_json = cast_type_it == index_params.end() ||
-                            cast_type_it->second == "JSON";
-        if (!is_flat_json) {
-            mask_count += 1;
-            dense_bitmap_count = 1;
-            if (index_type == HYBRID_INDEX_TYPE) {
-                mask_count += 1;
-            }
-        }
-    }
+    // Tantivy does not retain null rows, so the wrapper keeps a CRoaring null
+    // offset set beside the mmap'd index. JSON path extraction can produce
+    // null rows even when the source field is non-nullable. An unresolved
+    // HYBRID may select INVERTED, so reserve its null mask conservatively.
+    const bool keeps_null_mask =
+        ((index_type == INVERTED_INDEX_TYPE ||
+          index_type == NGRAM_INDEX_TYPE) &&
+         (is_json_path || field_nullable.value_or(true))) ||
+        (typed_json_path && index_type == HYBRID_INDEX_TYPE);
+
+    const uint64_t mask_count =
+        (keeps_null_mask ? 1 : 0) + (typed_json_path ? 1 : 0);
+    const uint64_t dense_bitmap_count = typed_json_path ? 1 : 0;
 
     // A CRoaring container is at most roughly one dense bitset plus container
     // metadata. Reserve 2x the dense bytes per row set to cover that overhead.
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -24,6 +24,7 @@
 #include <limits>
 #include <memory>
 #include <optional>
+#include <roaring/roaring.hh>
 #include <string>
 #include <vector>
 
@@ -186,6 +187,38 @@ SaturatingMul(uint64_t lhs, uint64_t rhs) {
 }
 
 uint64_t
+RoaringRowMaskResidentBytes(int64_t num_rows) {
+    if (num_rows <= 0) {
+        return 0;
+    }
+
+    // Roaring32 partitions values by their high 16 bits, so one low container
+    // covers exactly 2^16 consecutive row offsets.
+    constexpr uint64_t kRowsPerContainer = uint64_t{1} << 16;
+    // After runOptimize(), every container payload is at most 8 KiB: an array
+    // holds at most 4096 uint16_t values, a bitmap holds exactly 2^16 bits,
+    // and a run container is retained only when it is no larger than those
+    // alternatives.
+    constexpr uint64_t kMaxContainerPayloadBytes = 8 * 1024;
+    // RoaringMemoryBytes() also includes the portable-format header. For C
+    // containers it is 8 + 8*C bytes without runs and no more than
+    // 4 + ceil(C/8) + 8*C bytes with runs. Thus 16*C bounds either layout for
+    // every C >= 1.
+    constexpr uint64_t kMaxPortableMetadataBytesPerContainer = 16;
+    const auto rows = static_cast<uint64_t>(num_rows);
+    const auto container_count =
+        rows / kRowsPerContainer + (rows % kRowsPerContainer != 0);
+
+    const auto sparse_container_bound =
+        SaturatingAdd(static_cast<uint64_t>(sizeof(roaring::Roaring)),
+                      SaturatingMul(container_count,
+                                    kMaxContainerPayloadBytes +
+                                        kMaxPortableMetadataBytesPerContainer));
+    const auto dense_container_bound = SaturatingMul(BitsetBytes(num_rows), 2);
+    return std::max(sparse_container_bound, dense_container_bound);
+}
+
+uint64_t
 ScalarRowMaskResidentBytes(
     DataType field_type,
     const std::string& index_type,
@@ -217,10 +250,8 @@ ScalarRowMaskResidentBytes(
         (keeps_null_mask ? 1 : 0) + (typed_json_path ? 1 : 0);
     const uint64_t dense_bitmap_count = typed_json_path ? 1 : 0;
 
-    // A CRoaring container is at most roughly one dense bitset plus container
-    // metadata. Reserve 2x the dense bytes per row set to cover that overhead.
     auto roaring_bytes =
-        SaturatingMul(SaturatingMul(BitsetBytes(num_rows), 2), mask_count);
+        SaturatingMul(RoaringRowMaskResidentBytes(num_rows), mask_count);
     auto dense_bitmap_bytes =
         SaturatingMul(BitsetBytes(num_rows), dense_bitmap_count);
     return SaturatingAdd(roaring_bytes, dense_bitmap_bytes);
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -186,6 +186,48 @@ SaturatingMul(uint64_t lhs, uint64_t rhs) {
 }
 
 uint64_t
+ScalarRowMaskResidentBytes(
+    DataType field_type,
+    const std::string& index_type,
+    const std::map<std::string, std::string>& index_params,
+    int64_t num_rows) {
+    uint64_t mask_count = 0;
+    uint64_t dense_bitmap_count = 0;
+    if (index_type == INVERTED_INDEX_TYPE || index_type == NGRAM_INDEX_TYPE) {
+        // Tantivy does not retain null rows, so the C++ wrapper keeps one
+        // CRoaring row set beside the mmap'd index.
+        mask_count = 1;
+    }
+
+    const auto is_json_path =
+        field_type == DataType::JSON &&
+        index_params.find(JSON_PATH) != index_params.end();
+    if (is_json_path && index_type != NGRAM_INDEX_TYPE) {
+        // Typed JSON path indexes additionally keep the non-existing rows for
+        // EXISTS semantics. An unresolved HYBRID may select INVERTED, so use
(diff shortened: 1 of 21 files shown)


internal/core/src/index/InvertedIndexTantivy.cpp
```diff
@@ -354,17 +358,13 @@ InvertedIndexTantivy<T>::IsNull() {
...
         return bitset;  // All false - no element is null
     }
 
     auto fill_bitset = [this, count, &bitset]() {
-        auto end =
-            std::lower_bound(null_offset_.begin(), null_offset_.end(), count);
-        for (auto iter = null_offset_.begin(); iter != end; ++iter) {
-            bitset.set(*iter);
-        }
+        bitset = RoaringToBitset(null_offsets_, count);
```
buqian-zilliz, line 367: **[P3] `IsNull()` allocates the result bitset twice on the query path**

`IsNull()` constructs `TargetBitmap bitset(count)` and then overwrites it with `bitset = RoaringToBitset(null_offsets_, count)`, which returns a freshly allocated bitset. This doubles the per-call `row_count`-sized allocation: the first buffer is allocated and then immediately discarded by the move assignment. Consider building the null bitset in place (for example a `SetRoaringRows` helper mirroring `ClearRoaringRows`), or removing the up-front `TargetBitmap bitset(count)` and assigning the `RoaringToBitset` result directly.

internal/core/src/index/InvertedIndexUtil.h
```diff
@@ -11,7 +11,86 @@
...
+
+    offsets = roaring::Roaring();
+    const auto count = size / sizeof(size_t);
+    for (size_t i = 0; i < count; ++i) {
+        size_t offset;
+        std::memcpy(&offset, data + i * sizeof(size_t), sizeof(size_t));
+        AssertInfo(offset <= std::numeric_limits<uint32_t>::max(),
+                   "row offset {} exceeds uint32 range",
+                   offset);
+        offsets.add(static_cast<uint32_t>(offset));
```
buqian-zilliz: **[P3] `LoadLegacyOffsets()` regresses load from bulk memcpy to per-element `Roaring::add`**

`LoadLegacyOffsets()` replaces the previous `resize` + `FastMemcpy` bulk load with a per-offset `roaring::Roaring::add` loop followed by `runOptimize`/`shrinkToFit`. For large segments with many null/non-existent rows this converts an O(n) memcpy into n `add` calls, each performing a container lookup/binary search on the load path. The legacy payload is already sorted, so the bulk API already used in `ConvertJsonToTypedFieldData` (`addBulk`) — or `addMany` after a `size_t` to `uint32_t` conversion — would avoid the per-element container search.

internal/core/src/index/JsonScalarIndexWrapper.h
```diff
@@ -182,25 +188,27 @@ class JsonScalarIndexWrapper : public BaseIndex {
...
-                non_exist_offsets_.data(), e.data.data(), e.data.size());
+            LoadLegacyOffsets(non_exist_offsets_, e.data.data(), e.data.size());
         }
         LOG_INFO("LoadEntries JsonScalarIndexWrapper done, has_non_exist: {}",
                  has_non_exist);
-        // BaseIndex::LoadEntries has fully initialized the base index, so
-        // Count() is safe to call and we can eagerly build the exists bitmap.
-        BuildExistsBitset(this->Count());
+        non_exist_row_count_ =
+            GetValueFromConfig<int64_t>(config, INDEX_NUM_ROWS_KEY)
```
buqian-zilliz, line 196: **[P3] `INDEX_NUM_ROWS_KEY` is not populated on the load path, so the new row-count source is dead**

`LoadEntries`/`Load` now size `non_exist_row_count_` from `GetValueFromConfig<int64_t>(config, INDEX_NUM_ROWS_KEY).value_or(this->Count())`, but `INDEX_NUM_ROWS_KEY` is only written on the build path (`index_c.cpp::get_config`). The sealed-segment load config built in `Utils.cpp::LoadIndexData` comes from `ParseConfigFromIndexParams(index_params)` and does not contain this key, so the lookup always falls back to `this->Count()`, making the new row-count source effectively dead. If the intent is to size `exists_bitset_` from the segment row count rather than tantivy `Count()`, `num_rows` must be plumbed into the load config; otherwise this lookup is misleading and should be removed or annotated.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -185,6 +185,51 @@ SaturatingMul(uint64_t lhs, uint64_t rhs) {
...
+    const std::map<std::string, std::string>& index_params,
+    int64_t num_rows,
+    std::optional<bool> field_nullable) {
+    const auto is_json_path =
+        field_type == DataType::JSON &&
+        index_params.find(JSON_PATH) != index_params.end();
+    uint64_t mask_count = 0;
+    uint64_t dense_bitmap_count = 0;
+    if ((index_type == INVERTED_INDEX_TYPE || index_type == NGRAM_INDEX_TYPE) &&
+        (is_json_path || field_nullable.value_or(true))) {
```
buqian-zilliz: **[P2] Non-nullable scalar inverted/ngram fields over-reserve row-mask memory on the 6-arg load path**

`ScalarRowMaskResidentBytes` treats a missing `field_nullable` as nullable via `field_nullable.value_or(true)`, but the 6-arg `ScalarIndexLoadResource` overload (used by `ChunkedSegmentSealedImpl::LoadScalarIndex`) always passes `std::nullopt`. As a result, every non-JSON, non-nullable scalar field with an `INVERTED` or `NGRAM` index is charged `2 * BitsetBytes(num_rows)` of row-mask memory in `final_memory_cost`/`max_memory_cost`, even though its `null_offsets_` stays empty. Nullability is available at that call site (`field_meta`), so please plumb it through (or skip the mask when the field is known non-nullable) to avoid over-reserving memory on large segments.

internal/core/src/index/InvertedIndexTantivy.h
```diff
@@ -361,6 +364,20 @@ class InvertedIndexTantivy : public ScalarIndex<T> {
     virtual nlohmann::json
     BuildTantivyMeta(const std::vector<std::string>& file_names, bool has_null);
 
+    void
+    AddNullOffset(size_t offset) {
+        AssertInfo(offset <= std::numeric_limits<uint32_t>::max(),
```
buqian-zilliz, line 374: **[P3] CRoaring row masks cap row offsets at uint32 (2^32 rows)**

This change moves row offsets from `std::vector<size_t>` (64-bit) to CRoaring (32-bit values), so any row offset &gt;= 2^32 can no longer be represented. The new `AssertInfo` checks in `AddNullOffset`/`LoadLegacyOffsets` (and the JSON non-exist adders) fail loudly rather than truncating, but they turn previously-representable 64-bit offsets into a build/load failure for segments with more than ~4.29B rows. Please confirm this limit is an accepted tradeoff and document it, or add a higher-level guard that rejects such segments gracefully instead of throwing deep in the build/load path.

internal/core/src/index/JsonScalarIndexWrapper.h
```diff
@@ -344,9 +350,9 @@ class JsonScalarIndexWrapper : public BaseIndex {
                                                          cast_type_,
                                                          cast_function_);
             non_exist_offsets_ = std::move(result.non_exist_offsets);
-            auto total_rows = result.field_data->get_num_rows();
+            non_exist_row_count_ = result.field_data->get_num_rows();
             BaseIndex::BuildWithFieldData({result.field_data});
-            BuildExistsBitset(total_rows);
+            BuildExistsBitset(non_exist_row_count_);
```
buqian-zilliz: **[P3] JSON inverted build path never calls ComputeByteSize(), so non_exist_offsets_/exists_bitset_ are missing from ByteSize()**

`BuildInvertedWithJsonFieldData` builds `exists_bitset_` and returns without calling `ComputeByteSize()`, whereas the non-inverted `Build`/`BuildWithFieldData` branches were updated to call it. Consequently, on the inverted JSON-path build path the new `non_exist_offsets_` and `exists_bitset_` bytes are not reflected in `cached_byte_size_`/`ByteSize()`. This is recomputed correctly on `Load`/`LoadEntries`, so the impact is limited to build-phase byte-size accounting, but the asymmetry looks unintentional.

internal/core/src/index/InvertedIndexTantivy.cpp
sparknack: Addressed in acde6d6380. `IsNull()` now starts with an empty `TargetBitmap` and only allocates the bitmap returned by `RoaringToBitset()`.

internal/core/src/index/InvertedIndexUtil.h
sparknack: Addressed in acde6d6380. `LoadLegacyOffsets()` now converts the legacy `size_t[]` payload once and inserts it with `Roaring::addMany()`, followed by `runOptimize()` and `shrinkToFit()`.

internal/core/src/index/JsonScalarIndexWrapper.h
sparknack: Addressed in acde6d6380. `SealedIndexTranslator` now writes `index_load_info_.num_rows` to `INDEX_NUM_ROWS_KEY` before `Load()`/`LoadUnified()`, so the EXISTS bitmap uses the full segment row domain.

internal/core/src/index/IndexFactory.cpp
sparknack: Addressed in a21038c2e7 and clarified in 1df6b939c4. The sealed-segment path now passes `field_meta.is_nullable()`, and the file-aware path reads nullability from the field schema. Non-nullable scalar inverted/ngram indexes no longer reserve this mask.

internal/core/src/index/InvertedIndexTantivy.h
sparknack: Accepted tradeoff. CRoaring row masks intentionally use 32-bit offsets; the bound is far above practical segment sizes, and the existing build/load assertions reject out-of-range offsets instead of silently truncating them. No additional change is planned here.

internal/core/src/index/JsonScalarIndexWrapper.h
sparknack: No change planned. Tantivy `Build()` does not finalize the index; finalization happens in `Upload()`. The build pipeline does not consume `ByteSize()`: upload stats come from actual serialized sizes, while load accounting uses `ScalarIndexLoadResource`. `Load()`/`LoadEntries()` already recompute `ByteSize()` after the index is fully loaded.

internal/core/src/index/JsonScalarIndexWrapper.h
```diff
@@ -371,29 +377,32 @@ class JsonScalarIndexWrapper : public BaseIndex {
...
-            exists_bitset_.reset(offset);
-        }
+    BuildExistsBitset(size_t count) {
+        exists_bitset_ =
+            RoaringToBitset(non_exist_offsets_, count, /*inverted=*/true);
     }
 
-    std::vector<size_t> non_exist_offsets_;
+    roaring::Roaring non_exist_offsets_;
+    size_t non_exist_row_count_{0};
```
buqian-zilliz: **[P3] `non_exist_row_count_` is redundant per-index state: written and immediately consumed in the same call**

`non_exist_row_count_` is assigned and then passed straight into `BuildExistsBitset()` within the same function in every one of its uses (`Build`, `BuildWithFieldData`, `Load`, `LoadEntries`, and `BuildInvertedWithJsonFieldData`), and it is never read across function boundaries. The copy in `JsonHybridScalarIndex` has the same shape. It can be a plain local variable, which removes a piece of persisted per-index state that is otherwise easy to let go stale. Consider inlining it as a local unless a future cross-call reader is actually planned.

internal/core/src/index/IndexFactory.cpp
buqian-zilliz: This issue still exists for the V2 (non-V3) scalar load path. `ChunkedSegmentSealedImpl::LoadScalarIndex` now uses the 7-arg overload with `field_meta.is_nullable()`, but for V2 `INVERTED`/`NGRAM` scalar fields the request is pre-computed in `SegmentLoadInfo::BuildCache` via `IndexFactory::IndexLoadResource(...)` (no file context), which routes to the 6-arg `ScalarIndexLoadResource` overload that still passes `std::nullopt`. Because `info.load_resource_request` is then non-empty, `LoadScalarIndex` takes that over-reserved value instead of the nullability-aware 7-arg path, so non-nullable V2 scalar `INVERTED`/`NGRAM` fields are still charged `2 * BitsetBytes(num_rows)`. Please plumb `load_index_info.schema.nullable()` (available in `BuildCache`) through `IndexLoadResource`, or skip the mask for known non-nullable fields.

sparknack: Fixed in a2205d79bb. The metadata-only `IndexLoadResource` overload now accepts field nullability; both `SegmentLoadInfo::BuildCache` and `EstimateLoadIndexResource` pass `schema.nullable()`. Added a regression assertion for the cached V2 non-nullable request.

internal/core/src/index/JsonScalarIndexWrapper.h
sparknack: Fixed in a2205d79bb. Removed `non_exist_row_count_` from both `JsonScalarIndexWrapper` and `JsonHybridScalarIndex`; each build/load path now passes its local row count directly to `BuildExistsBitset()`.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -185,6 +185,47 @@ SaturatingMul(uint64_t lhs, uint64_t rhs) {
...
+         (is_json_path || field_nullable.value_or(true))) ||
+        (typed_json_path && index_type == HYBRID_INDEX_TYPE);
+
+    const uint64_t mask_count =
+        (keeps_null_mask ? 1 : 0) + (typed_json_path ? 1 : 0);
+    const uint64_t dense_bitmap_count = typed_json_path ? 1 : 0;
+
+    // A CRoaring container is at most roughly one dense bitset plus container
+    // metadata. Reserve 2x the dense bytes per row set to cover that overhead.
+    auto roaring_bytes =
```
buqian-zilliz, line 253: **[P3] `ScalarRowMaskResidentBytes` under-reserves sparse CRoaring masks versus `RoaringMemoryBytes`**

`ScalarRowMaskResidentBytes` sizes each CRoaring mask as `2 * BitsetBytes(num_rows)` (~`num_rows/4` bytes), keyed to the row count rather than to the number of stored offsets. A CRoaring array container costs 2 bytes per stored offset, so a nullable field with many sparse (non-consecutive) nulls in a small-to-medium segment under-reserves: with ~50% scattered nulls the actual mask can be up to ~4x larger than `2 * BitsetBytes(num_rows)`. This diverges from `RoaringMemoryBytes()`, which `ComputeByteSize()` uses to report the real array-container cost, so `final_memory_cost`/`max_memory_cost` can under-count the resident row mask. Large/dense segments are fine (fixed 8KB bitset containers make the estimate conservative), but the two independent accounting paths should agree. Consider estimating from the null/non-exist offset count or bounding by `ceil(num_rows / 65536) * container_size`.

sparknack: Fixed in e1e2f2a1cb. `RoaringRowMaskResidentBytes()` now takes the larger of the existing 2x-dense estimate and an 8 KiB payload + metadata bound per 16-bit container. Added an 8192-row alternating-offset regression test that requires the load estimate to cover `RoaringMemoryBytes()`.

sparknack:
this will be handled by @SpadeA-Tang , closed.

Issue #52525, opened 2026-08-14
[Bug]: DropCollection ACK callback blocks on full RBAC policy refresh


chyezh:
## Environment

- **Milvus Version**: master build containing #48140; relevant source cross-checked at `e6b52c9e10` (the exact deployed image tag was not available from read-only logs)
- **Deployment Mode**: cluster, Streaming Service enabled, 1 MixCoord and 2 Proxies
- **MQ**: not relevant to the observed bottleneck
- **SDK**: not relevant; the delay occurs after all VChannels have acknowledged the broadcast
- **OS**: Kubernetes/Linux

### Option B: Steps

1. Deploy a Milvus cluster with Streaming Service enabled and at least 2 Proxy instances.
2. Populate RBAC metadata with a large number of collection-level grantee entries. The affected environment had approximately 47k grantee records, inferred from metadata operation metrics.
3. Create collections and then issue `DropCollection`. Multiple overlapping drops make the latency amplification easier to observe, but isolated drops are also slow.
4. Inspect Streaming Coordinator broadcaster logs and RootCoord metrics:
   - The DropCollection task reaches all-VChannel ACK in about 10-15ms.
   - QueryCoord release, DataCoord `DropVirtualChannel`, RootCoord metadata deletion, and tombstone insertion finish in about 0-1s.
   - The task remains pending for another 101-126s before the final Proxy collection cache invalidation and `BROADCAST_TASK_STATE_TOMBSTONE`.
5. Inspect `milvus_rootcoord_ddl_req_latency_*{function_name="PolicyList"}` and `milvus_meta_request_latency_*{meta_op_type="get"}`.

### Trigger Conditions

- **Frequency**: consistently reproducible for DropCollection in the affected high-cardinality RBAC environment
- **First observed after**: RBAC grant cardinality reached approximately 47k grantee records
- **Does NOT happen when**: broadcaster tasks that do not execute the DropCollection full RBAC refresh complete in milliseconds; the delay is independent of VChannel ACK latency and collection load state

## Expected Behavior

After all VChannels acknowledge a DropCollection broadcast and the QueryCoord/DataCoord/RootCoord cleanup finishes, the broadcaster callback should complete within a few seconds and transition the task to TOMBSTONE.

Dropping one collection should not synchronously execute an O(total RBAC grants) sequence of remote metadata reads once per Proxy, nor keep the broadcaster resource-key lock occupied for approximately two minutes.

## Actual Behavior

In one observed interval:

[...]

PR #52535, opened 2026-08-14
fix: avoid N+1 RBAC policy metadata scans

chyezh:
issue: #52525

- RBAC policy listing: bulk-load the grantee and privilege subtrees and join them in memory so metadata access uses two fixed prefix scans
- legacy compatibility: index stored grantee ID ownership once while preserving fallback, malformed-key handling, and shared-ID fail-closed behavior

### DIFF
--- a/internal/metastore/kv/rootcoord/kv_catalog.go
+++ b/internal/metastore/kv/rootcoord/kv_catalog.go
@@ -1471,6 +1471,7 @@ func logicalGranteeKeyFromEtcdKey(ctx context.Context, granteePrefix string, key
 }
 
 func findOtherGranteeWithIDFromKeys(ctx context.Context, granteePrefix string, granteeKey string, idStr string, keys []string, values []string) (string, error) {
+	var ownerKeys []string
 	for i, key := range keys {
 		if i >= len(values) {
 			mlog.Warn(ctx, "grantee key has no matching id value while checking grantee id sharing",
@@ -1480,6 +1481,13 @@ func findOtherGranteeWithIDFromKeys(ctx context.Context, granteePrefix string, g
 		if values[i] != idStr {
 			continue
 		}
+		ownerKeys = append(ownerKeys, key)
+	}
+	return findOtherGranteeWithIDFromOwnerKeys(ctx, granteePrefix, granteeKey, idStr, ownerKeys)
+}
+
+func findOtherGranteeWithIDFromOwnerKeys(ctx context.Context, granteePrefix string, granteeKey string, idStr string, ownerKeys []string) (string, error) {
+	for _, key := range ownerKeys {
 		logicalKey, ok := logicalGranteeKeyFromEtcdKey(ctx, granteePrefix, key)
 		if !ok {
 			return "", newSharedGranteeIDError(idStr, granteeKey, key)
@@ -2003,6 +2011,31 @@ func (kc *Catalog) ListPolicy(ctx context.Context, tenant string) ([]*milvuspb.G
 		mlog.Error(ctx, "fail to load all grant privilege entities", mlog.String("key", granteeKey), mlog.Err(err))
 		return []*milvuspb.GrantEntity{}, err
 	}
+	granteeOwnerKeys := make(map[string][]string)
+	for i, key := range keys {
+		if i >= len(values) {
+			mlog.Warn(ctx, "grantee key has no matching id value while indexing grantee id owners",
+				mlog.String("key", key), mlog.String("prefix", granteeKey))
+			continue
+		}
+		granteeOwnerKeys[values[i]] = append(granteeOwnerKeys[values[i]], key)
+	}
+
+	granteeIDPrefix := funcutil.HandleTenantForEtcdPrefix(GranteeIDPrefix, tenant)
+	allIDKeys, _, err := kc.Txn.LoadWithPrefix(ctx, granteeIDPrefix)
+	if err != nil {
+		mlog.Error(ctx, "fail to load all grantee ids", mlog.String("key", granteeIDPrefix), mlog.Err(err))
+		return []*milvuspb.GrantEntity{}, err
+	}
+	granteeIDKeys := make(map[string][]string)
+	for _, key := range allIDKeys {
+		granteeIDInfos := typeutil.AfterN(key, granteeIDPrefix, "/")
+		if len(granteeIDInfos) < 2 || funcutil.IsEmptyString(granteeIDInfos[0]) {
+			mlog.Warn(ctx, "invalid grantee id", mlog.String("string", key), mlog.String("sub_string", granteeIDPrefix))
+			continue
+		}
+		granteeIDKeys[granteeIDInfos[0]] = append(granteeIDKeys[granteeIDInfos[0]], key)
+	}
 
 	for i, key := range keys {
 		grantInfos := typeutil.AfterN(key, granteeKey, "/")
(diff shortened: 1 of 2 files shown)


czs007:

Reviewed at head `8b3f806`. No blocking issues found — the rewrite is behaviour-preserving as far as I can trace it, and the `LoadWithPrefix`-count assertions are a good regression guard. Findings below are medium-and-below (scalability / dead code / micro-efficiency).

## What I verified (equivalence audit)

I walked every branch of the old `loadGranteeIDPrefixWithLoadedGrantees` path against the new inline logic in `ListPolicy` and they match:

- **Legacy shared-ID fail-closed**: `granteeOwnerKeys` (`kv_catalog.go:2014-2022`) applies exactly the same `i >= len(values)` skip and `values[i] != idStr` filter that `findOtherGranteeWithIDFromKeys` did, and hands the same key set to the (now extracted) `findOtherGranteeWithIDFromOwnerKeys`. Malformed grantee keys that carry the target ID still produce `newSharedGranteeIDError` rather than being silently dropped. `TestRBACGrantSharedLegacyGranteeIDMigrationFailsClosed` still exercises this through `ListPolicy` against a real etcd KV.
- **Candidate fallback order and short-circuit**: `granteeIDCandidates` order (`stored ID`, then computed) and the break-on-first-non-empty semantics are unchanged, including the subtle case where the stored ID's subtree contains *only* an empty privilege segment (`.../grantee-id/<id>/`). Old code: `LoadWithPrefix` returns 1 key → break, no fallback. New code: `AfterN` yields `["<id>", ""]`, len 2 → indexed → break, no fallback. Same outcome (grantee contributes zero policies). The reworked `TestRBACReadSkipsEmptyKeySegments` pins this.
- **Malformed grantee-id keys**: `.../grantee-id/<id>` with no privilege segment is dropped by the new `len(granteeIDInfos) < 2` guard, and previously would not have matched the `<id>/` prefix scan either. `<id>/random/a/b/c` is still indexed but skipped by the `len(...) != 1` check in the inner loop, matching the old behaviour — and the updated `TestRBAC_Grant` mock now actually covers it.
- **Result ordering** is preserved: all three `TxnKV` implementations (etcd `WithSort(SortByKey, SortAscend)`, memkv `tree.Ascend`, tikv iterator) return keys in ascending order, so per-ID key ordering inside `granteeIDKeys` is the same as the per-prefix scans produced.
- **Removed `errors.Is(err, merr.ErrIoKeyNotFound)` handling** in the candidate loop is not a regression: no `LoadWithPrefix` implementation returns that error (`etcd_kv.go:128`, `mem_kv.go:270`, `txn_tikv.go:338` all return `nil, nil, nil` on empty); it only comes from `Load`/`MultiLoad`. That branch was defensive dead code.
- **Snapshot consistency is slightly better**, not worse: all grantees now resolve their ID subtree against one grantee-id snapshot instead of N snapshots taken at N different times.

## Findings

### Medium — the grantee-id subtree is now fetched as a single unbounded response
`kv_catalog.go:2024-2025`
```go
granteeIDPrefix := funcutil.HandleTenantForEtcdPrefix(GranteeIDPrefix, tenant)
allIDKeys, _, err := kc.Txn.LoadWithPrefix(ctx, granteeIDPrefix)
```
`etcdKV.LoadWithPrefix` issues one `Get` with `WithPrefix()` and no `WithLimit` (`internal/kv/etcd/etcd_kv.go:128-146`). The grantee-id subtree has one key per *(grantee, privilege)* pair, so its cardinality is strictly larger than the grantee-privileges subtree this function already scans — it is now the single largest response in the call.

Two aggravating factors: `ListPolicy` is only ever invoked with `util.DefaultTenant` (`root_coord.go:2738`), and `DefaultTenant == ""`, so the prefix is the bare `root-coord/credential/grantee-id/` and the scan also pulls every *non-default* tenant's subtree into the response (those keys index under the tenant name and are then never looked up — harmless for correctness, pure payload). And `ListPolicy` fans out: every proxy refetches the whole policy set on start and on each privilege-change broadcast (`internal/proxy/privilege/cache.go:71`).

Failure scenario: a deployment with a large RBAC dataset (tens of thousands of role×object×privilege triples) turns N small etcd responses into one multi-MB gRPC message; the old code would have been slow but would not have risked a single oversized response or a large transient allocation spike on rootcoord per refresh.

**Suggested fix:** use the paginated helper the KV layer already exposes — `WalkWithPrefix(ctx, granteeIDPrefix, batchSize, fn)` (`etcd_kv.go:87`), with the same accumulate-into-`granteeIDKeys` callback. There is in-repo precedent at `internal/metastore/kv/rootcoord/legacy_snapshot_gc.go:92`. This keeps the "constant number of logical scans" property that is the point of this PR while bounding per-response size. It is a genuine trade-off, not a defect — the PR direction is right either way — but worth doing before this lands on large clusters.

### Medium — `loadGranteeIDPrefixWithLoadedGrantees`'s loaded-grantees branch is now dead code
`kv_catalog.go:1521-1541`
```go
func (kc *Catalog) loadGranteeIDPrefix(...) (...) {
	return kc.loadGranteeIDPrefixWithLoadedGrantees(ctx, tenant, granteeKey, idStr, nil, nil)
}

func (kc *Catalog) loadGranteeIDPrefixWithLoadedGrantees(..., granteeKeys []string, granteeValues []string) (...) {
	...
	if granteeKeys != nil {
		otherGranteeKey, err = findOtherGranteeWithIDFromKeys(...)   // <- now unreachable
	} else {
		otherGranteeKey, err = kc.findOtherGranteeWithID(...)
	}
```
`ListPolicy` was the only caller that passed non-nil `granteeKeys`. After this PR the sole caller is `loadGranteeIDPrefix`, which always passes `nil, nil`, so the `granteeKeys != nil` arm can never execute.

Failure scenario: this is a maintenance hazard rather than a runtime bug — the shared-legacy-ID invariant now lives in two structurally different places (the dead branch here, and the inline block at `kv_catalog.go:2051-2063`). A future change to the invariant applied to only one of them will silently diverge, and no test can catch the dead one.

**Suggested fix:** collapse `loadGranteeIDPrefixWithLoadedGrantees` back into `loadGranteeIDPrefix` and drop the `granteeKeys`/`granteeValues` parameters. Keep `findOtherGranteeWithIDFromKeys` itself — it is still live from `MigrateGrantCollectionName` (`kv_catalog.go:1892`).

### Low — `granteeOwnerKeys` is populated for every grantee, but only legacy IDs are ever looked up
`kv_catalog.go:2014-2022` builds one map entry (and one 1-element slice allocation) per grantee, but the map is read only inside the `isLegacyGranteeID(idStr)` guard at line 2053. On a fully migrated cluster — every stored ID 32 chars — the map is built in full and never read.

Failure scenario: no wrong output, but this is O(N) map inserts plus O(N) slice allocations on the exact hot path the PR exists to speed up, and it grows with grantee count.

**Suggested fix:**
```go
for i, key := range keys {
	if i >= len(values) { /* warn; continue */ }
	if !isLegacyGranteeID(values[i]) {
		continue
	}
	granteeOwnerKeys[values[i]] = append(granteeOwnerKeys[values[i]], key)
}
```
This is safe because the map is only consulted under the same `isLegacyGranteeID` predicate.

### Low — vestigial "first candidate prefix" bookkeeping
`kv_catalog.go:2065-2077`
```go
if granteeIDKey == "" {
	granteeIDKey = candidatePrefix
}
```
In the old code this value was carried out of the helper and used in the failure log. In the new inline loop it only survives when *no* candidate matched — in which case `idKeys` is empty, the loop at line 2078 never runs, and `granteeIDKey` is never read. It also calls `HandleTenantForEtcdPrefix` (a string concat) once per candidate even on the miss path.

**Suggested fix:** set `granteeIDKey` only on the hit, and skip building `candidatePrefix` until the map lookup succeeds:
```go
for _, candidate := range granteeIDCandidates(logicalGranteeKey, idStr) {
	if candidateKeys := granteeIDKeys[candidate]; len(candidateKeys) > 0 {
		idKeys = candidateKeys
		granteeIDKey = funcutil.HandleTenantForEtcdPrefix(GranteeIDPrefix, tenant, candidate)
		break
	}
}
```

### Low — `ListGrant` retains the same N+1 pattern
`kv_catalog.go:1707`, inside `appendGrantEntity`:
```go
keys, values, granteeIDKey, err := kc.loadGranteeIDPrefix(ctx, tenant, granteeKey, v)
```
When `ListGrant` takes the role-wide branch (no `ObjectName`), `appendGrantEntity` runs once per matching grantee, so it still issues one prefix scan per grantee. Given the PR title and that the bulk-index machinery now exists, worth noting as a follow-up — not a blocker for this PR's scope, but a reader will reasonably expect "avoid N+1 RBAC policy metadata scans" to have covered it, so a line in the PR body scoping it to `ListPolicy` would help.

### Nit — `crypto.GranteeID(logicalGranteeKey)` is computed twice per grantee
Once at `kv_catalog.go:2051` for the legacy check, and again inside `granteeIDCandidates` at line 2067. Two MD5s per grantee instead of one. Consider hoisting `newID := crypto.GranteeID(logicalGranteeKey)` and passing it into `granteeIDCandidates`.

### Nit — global method patch in the new test
`kv_catalog_test.go:3311` patches `(*memkv.MemoryKV).LoadWithPrefix` process-wide via mockey (this also introduces the first mockey usage in this package, and consequently makes the file depend on `-gcflags="all=-N -l"`). The `defer patched.UnPatch()` is correct for sequential execution, but the patch is visible to any other test in the package that runs concurrently, so it is fragile against a future `t.Parallel()`.

A counting `TxnKV` decorator would be equivalent, cheaper, and self-contained:
```go
type countingKV struct {
	kv.TxnKV
	loadWithPrefixCalls atomic.Int32
}
func (c *countingKV) LoadWithPrefix(ctx context.Context, key string) ([]string, []string, error) {
	c.loadWithPrefixCalls.Inc()
	return c.TxnKV.LoadWithPrefix(ctx, key)
}
```

## Summary

Correctness looks sound; the equivalence audit above did not turn up a behaviour change in the legacy-ID, malformed-key, fallback, or fail-closed paths, and the `assert.Equal(t, int32(2), loadWithPrefixCalls.Load())` assertions in `TestRBACListPolicyBulkLoadsGranteeIDs` are exactly the right guard for the property this PR is establishing. The medium items — the unbounded single-response scan and the now-dead loaded-grantees branch — are the two I would ask to be addressed before merge; the rest are optional cleanups.

PR #52541, opened 2026-08-14
enhance: build sealed tantivy indexes directly

zhengbuqian:
## What problem does this PR solve?

Sealed Tantivy index builds previously used the regular writer path. The writer
could create multiple segments and then merge them into one final segment,
causing the index data, postings, and document store to be written again.

This is particularly expensive for NGRAM indexes because each source row expands
into many terms and postings. The build path also crossed the C++/Rust boundary
and allocated/submitted documents one row at a time.

## What are the changes?

- Use the new direct single-segment writer for all sealed Tantivy-backed index
  builds by default.
- Cover sealed scalar, array, JSON, NGRAM, text, and phrase index build paths.
- Batch C++ to Rust FFI calls.
- Batch document construction and writer submission.
- Remove per-row `RustResultWrapper` allocation, document allocation, and
  channel submission from the sealed build path.
- Generate and write NGRAM documents inside Rust without returning individual
  grams to C++.
- Finalize the direct writer before creating the in-memory reader.
- Keep growing Text Match on the regular writer with background merge.
- Add/update tests for inverted, NGRAM, and Text Match indexes.

The produced index remains a regular Tantivy single-segment index. Query code and
index format are unchanged.

## Performance

One serial run was performed against the direct parent base, covering all 28
scalar benchmark cases that use Tantivy. It produced 212 paired build-time
measurements.

| Index category | Configurations | Base median | Optimized median | Reduction |
| --- | ---: | ---: | ---: | ---: |
| NGRAM | 8 | 5764 ms | 1479 ms | 73.9% |
| Scalar/array/mmap INVERTED | 67 | 2411 ms | 957 ms | 60.5% |
| JSON path INVERTED | 5 | - | - | 34.2% |
| JSON Flat/Path Tantivy | 5 | - | - | 34.6% |

Text Match and Phrase Match do not expose an isolated index-build timer in this
benchmark. Their combined data-generation, load, and build measurements improved
by 49.4% and 44.2%, respectively.

Non-Tantivy bitmap/sort control cases had approximately 0% median change.

NGRAM examples:

| Dataset | Base | Optimized | Reduction |
| --- | ---: | ---: | ---: |
| LIKE, cardinality 100 | 5739 ms | 1387 ms | 75.8% |
| LIKE, cardinality 1M | 14821 ms | 4868 ms | 67.2% |
| Regex, cardinality 100 | 5789 ms | 1358 ms | 76.5% |
| Regex, cardinality 1M | 14844 ms | 4938 ms | 66.7% |

## Verification

- Completed one serial run of all 28 Tantivy-related scalar benchmark cases.
- All 212 build-time records were successfully paired with the base run.
- Full Milvus unit-test and query-performance suites were not run as part of
  this benchmark.


zhengbuqian:
we'll continue use the existing index building method, but changing the segment size to a larger value to prevent frequent segment merge during index build time

Issue #2, opened 2019-09-23
[FEATURE] Delete vectors by ID


solevarr:
**Is your feature request related to a problem? Please describe.**
I wish I could use Milvus to find and delete vector by it's corresponding ID. Currently the SDK supports only removing the vectors using the date-time, which is a little bit useless for the following cases; suppose I've inserted a huge dataset of animals and the last image vector of a person was mistakenly added, so in order to remove the last vector I need to remove the whole table. 

**Describe the solution you'd like**
Delete vector by its corresponding ID. 

**Describe alternatives you've considered**
Actually this feature is already implemented in FAISS search engine and the other ones. I think you know that it is possible to remove the ID from the index, using remove_ids()  in FAISS. Is this possible to add the same feature for the Milvus?

**Additional context**
https://github.com/facebookresearch/faiss/wiki/Special-operations-on-indexes

JinHai-CN:
Yes, we already planned to add this feature in the future.
Please check the roadmap: https://milvus.io/docs/en/roadmap/
Any comments are welcome!

yhmo:
This requirement should be given high priority since more people mentioned it.
https://github.com/milvus-io/pymilvus/issues/113

From other community channel we also receive question for this function. I think we should consider design/implement it in the near future.
Reopen this issue.

namnh2204:
I'm still waiting for this feature...

yhmo:
This feature is in tech design phase. We will implement it with the highest priority.

yhmo:
This feature is in tech design phase. We will implement it with the highest priority.

yhmo:
Already implemented in master branch, will be released in 0.7.0

PR #52542, opened 2026-08-14
enhance: persist WAL recovery state asynchronously

chyezh:
issue: #40451

- recovery storage: track WAL message ownership and persist dirty snapshots before advancing data checkpoints
- vchannel persistence: write segment packs and transform logs to object storage and publish recovery metadata to etcd asynchronously
- WAL lifecycle: replace the legacy flusher path with RecoveryStorage and recover through bounded RecoveryBarrier replay
- transform materialization: bound L0 materialization by the earliest uncommitted L1 segment and retry the retained request as L1 final commits complete
- summary/transformlog decoupling: WALSummary persists autonomously (pchannel-scoped; byte-threshold or tracker-stall driven) and TransformLog materializes autonomously (vchannel-scoped; observation-driven) with no ordering constraint between the two. Delete handles are released by summary persistence; recovered chunk rewrites and re-materialization are deduped by the restored durable frontier / loaded window
- flush-state reporting: restore DataCoord.UpdateChannelCheckpoint reporting through a deprecated PChannelCheckpointUpdater on PChannelRecoveryManager so GetFlushState observes progress and Flush completes

- adversarial review (debate-Dc5d4471c8f) fixes: serialize manifest publish (flush vs GC) on a dedicated mutex with clone-and-install; inherit the previous term&#39;s manifest/chunks on recovery (chunk reads/deletes use the chunk&#39;s own term); roll back the generation on flush failure and retry the same chunk key; requeue scheduler entries with backoff; merge concurrent flush requests into a single task; restore staging under the manager lock; schedule a successor materialize task on late upper-bound raises; close started recovery modules when the second recovery phase fails; capture staging-bytes under the view lock; checkpoint updater skips message IDs without an MQ wrapper. Verified by unit tests (cross-term inheritance, flush-failure rollback/retry, GC snapshot iteration, staging restore, concurrent flush/GC, upper-bound raise, second-phase cleanup) plus a standalone E2E: term handoff inherits the previous term&#39;s chunk (term=3 recovered chunks=1, nextGeneration=1) and the un-materialized delete stays effective after kill -9 and restart.

## Scope decision: consume-checkpoint concurrency fencing is NOT in this PR (TODO)

The consume checkpoint is the **commit point** of `SaveRecoverySnapshot`: it is staged with `CommitSave`, the last write of the update, after every other part of the snapshot has landed.

A full concurrency fence on the checkpoint — term fast-fail + guarded commit (create-if-absent for the first write, value CAS afterwards) + post-commit read-back — was prototyped in this branch and then **reverted** (`87bd734c59`): it requires backend compare-and-swap support that TiKV does not provide yet (`CompareVersionAndSwap` is a stub returning `ErrServiceUnimplemented`, and `StreamingNode` uses the TiKV backend when `metastore.type=tikv`), so the guarantee could not hold on every supported backend and the first-write path hit unbounded retries.

The intended fenced design and the risk of the current un-fenced state (a superseded older-term publisher can still overwrite the checkpoint and component state of the current publisher; WAL truncation follows the last writer) are marked as TODO at the sites:

- `internal/metastore/kv/streamingnode/update.go` — checkpoint write site
- `docs/design-docs/design_docs/wal/checkpoint-persistence.md` §5.1
- `docs/design-docs/design_docs/wal/summary.md` §2.2

Latest commits:
- `6c41b8f16b` fix: pin failed summary chunk batch to avoid corrupted retry (#52542)
- `87bd734c59` revert: drop ConsumeCheckpoint CAS fencing, keep checkpoint as commit point (#52542)


czs007:

Reviewed the full branch at head `ab9dafb` (merge-base `c95a29d`, 107 files). This is a large and generally well-structured rework: the ref-counted message ownership model (`pkg/streaming/util/message/ref_counted_message.go`) plus `messageack.Tracker` gives a clean "data checkpoint = released-prefix" invariant, and the persist ordering (module snapshots first, consume checkpoint last via `CommitSave`) is correct — a crash between the two only replays idempotent work. I also confirmed the scope exclusion holds: nothing under `snview`/QueryView/qvresource is touched.

That said, I found issues I would consider blocking before this leaves draft: a close-path hang, and a lost catalog-cleanup path for dropped vchannels that is a regression against master. Details below.

---

### High

**1. `persistDritySnapshotWhenClosing` can spin forever and never return — WAL close hangs**
`internal/streamingnode/server/wal/recovery/recovery_background_task.go:83-87`

```go
for rs.isDirty() {
    if err := rs.persistDirtySnapshot(ctx, mlog.InfoLevel); err != nil {
        return err
    }
}
```

`isDirty()` (same file, line 35-38) returns true whenever `vchannelManager.HasPendingCleanup()` is true, i.e. some segment sits in `cleanupSegments`. But `consumeDirtySnapshot()` will produce no snapshot for that segment unless `tombstonedCleanupReadyLocked` (`vchannel/segment/view.go:581-590`) holds, and that requires **strictly**:

```go
metaPhysicalTimeTick > tombstoneTimeTick && dataPhysicalTimeTick > tombstoneTimeTick
```

where both physical ticks come from the *persisted* checkpoint. When they are equal to the tombstone tick, `persistDirtySnapshot` returns `nil` immediately (line 107-109, nothing dirty), `isDirty()` stays true, and the loop never checks `ctx.Err()` — so the `gracefulTimeout` never fires either. Result: `RecoveryStorage.Close()` never returns, `walAdaptorImpl.Close()` blocks, and one core is pinned.

Failure scenario: the last message the data scanner observes before `backgroundTaskNotifier.Cancel()` is the flush that ultimately tombstones the last segment (manual flush / drop collection right before a rolling restart). After the drain, `persistedCheckpoint.TimeTick == persistedCheckpoint.DataCheckpoint.TimeTick == tombstoneTimeTick`, no new message can arrive (the scanner is already cancelled), so the strict `>` can never be satisfied.

Suggested fix: check `ctx.Err()` inside the drain loop and return, and make `isDirty()` report only *actionable* cleanup (evaluate `segmentCleanupReadyLocked` against the current persisted checkpoint) rather than "any pending cleanup candidate". Leaving a tombstoned segment record behind for the next open is harmless; hanging shutdown is not.

**2. Dropped vchannel / transform-log metadata is never removed from the catalog (regression vs master)**
`internal/metastore/kv/streamingnode/kv_catalog.go:254-277`, `internal/streamingnode/server/wal/vchannel/module.go:579-605`

Master's `getRemovalAndSaveForVChannel` had an explicit branch:

```go
if info.GetState() == streamingpb.VChannelState_VCHANNEL_STATE_DROPPED {
    // remove the vchannel key and all of its schema keys
    ...
    return removes, kvs, nil
}
```

That branch is gone; the new function always *saves* the meta, whatever the state. The replacement path does not exist either: `module.ConsumeCleanupSnapshots` emits `SnapshotOpDelete` only for segments (line 594-602), so `catalog.DropVChannels` (kv_catalog.go:172) and `catalog.DropTransformLogMeta` (kv_catalog.go:235) have no non-test callers at all. Correspondingly, the whole vchannel tombstone lifecycle is written but never invoked — `VChannelView.TryFinalizeTombstone` (`vchannel/view.go:174`), `HasReadyTombstoneFinalize` (:180), `TombstonedCleanupPlan` (:300), `VChannelDropCleanupSnapshot` (:345), `ApplyPartitionCleanup` (:360) and `PartitionCleanupPlan` (:381) are all unreferenced outside tests.

Failure scenario: create/drop a collection N times on one pchannel. Every dropped vchannel leaves its `VChannelMeta` key plus one schema key per schema version in etcd forever. On every WAL recovery, `PChannelRecoveryManager.initialVChannels` (`vchannel/manager.go:80-97`) rebuilds a module for each of them, and `observeBroadcastMessage` (:279-290) then fans every pchannel-level message out to all of those dead modules. Etcd meta, recovery time and steady-state CPU all grow monotonically with the number of collections ever created on the channel.

Suggested fix: wire the vchannel tombstone path the same way the segment path is wired (`TryFinalizeTombstone` from `markSegmentViewUpdatedLocked`'s sibling, cleanup candidates surfaced through `HasCleanupCandidates`), emit `SnapshotOpDelete` for `ModuleNameVChannel` and `ModuleNameTransformLog`, and drop the module from `m.modules` when the delete is acknowledged.

---

### Medium

**3. Age-based flush is only evaluated when a new insert arrives — idle buffers pin the WAL**
`internal/streamingnode/server/wal/vchannel/segment/flush_policy.go:43-57`, `vchannel/segment/view.go:236`, `vchannel/transformlog/transform_log.go:141-174`

`ShouldFlush` honours `DataNodeCfg.SyncPeriod`, but it is only ever called from `ObserveInsertMessageV1`/`ObserveTxnMessage`. `VChannelRecoveryModule.ObserveMessage` has no `MessageTypeTimeTick` case, and `ClassifyTransformLogMessage` maps timetick to `TransformLogKindNone`, so no timer or timetick drives buffered data out. The old flusher path was timetick-driven, so age-based sync fired without new writes.

Failure scenario: a client writes a burst (inserts and/or deletes) that stays under `FlushInsertBufferSize`, then the collection goes idle. The retained `RetainedImmutableMessage` handles are never released, so the ack-tracker prefix never advances, `simpleTruncateCheckpoint` never moves, and the WAL retains those entries (and the process retains the payloads in memory) until the shard manager's much coarser seal policy eventually emits a flush. Suggest driving `ShouldFlush`/`syncUp` from timetick messages, or from the recovery background ticker.

**4. Inserts are still accepted into a FLUSHED segment and can end up in a chunk nothing will ever flush**
`internal/streamingnode/server/wal/vchannel/segment/view.go:536-545` and `:228-239`

`canReplayInsertLocked` returns true for a `FLUSHED` segment when `timetick <= meta.CheckpointTimeTick`. Between observing the flush message and the async `commitL1SegmentTask` completing, `DataCheckpointTimeTick` is still behind, so such an insert is appended to `s.pending` (line 234). Once `finalCommitDone` is set, `canScheduleFinalCommitLocked` (`segment/task.go:148-150`) is permanently false, so no commit task will pick that chunk up.

Two distinct bad outcomes: (a) the chunk's handle is never released → the pchannel data checkpoint stalls permanently; (b) if the commit task runs first, `markPendingDataDurableLocked(t.timetick)` (`segment/task.go:105`) releases handles and advances `DataCheckpointTimeTick` past rows that were never written, so after a restart they are skipped by the `timetick <= DataCheckpointTimeTick` guard and are lost. This is defensive today (the allocator should not assign to a sealed segment), but the state machine should not depend on that: reject inserts once the state is `FLUSHED`, or force the pending chunk into the in-flight commit.

**5. Module snapshots are persisted one etcd round-trip per snapshot**
`internal/streamingnode/server/wal/recovery/recovery_background_task.go:140-149`

Each `DirtySnapshot` becomes its own `SaveVChannels` / `SaveSegmentAssignments` / `SaveTransformLogMeta` call. With K dirty segments, a persist interval issues K separate etcd transactions, where master batched the whole snapshot into one `SaveRecoverySnapshot`. This also contradicts the comment at lines 41-46 ("already gathers the whole snapshot into one catalog call") — that is now only true for the checkpoint. Suggest grouping snapshots by module+op and passing the maps to the existing batched catalog APIs before calling `MarkPersisted`.

**6. Non-retryable errors are retried forever, silently blocking all checkpoint progress**
`internal/streamingnode/server/wal/recovery/recovery_background_task.go:168-222` + `:263-284`

The whole `switch` is wrapped in `retryOperationWithBackoff`, which only exits on `ctx.Err()`. A payload type mismatch or unknown op (`merr.WrapErrServiceInternalMsg(...)`, e.g. lines 175, 191, 220) will loop at ~1 op/s forever. Because `persistModuleDirtySnapshots` runs before `persistCheckpointSnapshot`, the consume checkpoint and WAL truncation stop advancing for that pchannel with only a repeating warn log. Suggest classifying: retry I/O, fail fast (or drop the snapshot with an error metric) on programming/data errors.

**7. Silent data drop when a segment cannot resolve its schema**
`internal/streamingnode/server/wal/vchannel/module.go:386-392`, `:407-415`

`handleCreateSegmentMessage` returns without creating the view when `CreateSegmentSchema` returns nil (dropped/missing partition meta, or a timetick outside the replay window), and `handleInsertMessage` then skips every insert for that segment id (`view == nil → continue`). The ack tracker still releases those messages, so the checkpoint advances and the rows are gone — with no log line and no metric. At minimum log at warn with segment/partition/timetick, and consider whether this should be a `DataIntegrity` error instead of a silent skip. Same comment for `PChannelRecoveryManager.moduleForMessage` (`manager.go:292-315`), which drops any non-`CreateCollection` message for an unknown vchannel.

---

### Low

**8. `PChannelRecoveryManager.Close()` is a no-op** (`vchannel/manager.go:273`) and nothing ever removes entries from `m.modules`. Also, in `moduleForMessage` (:301-314) a module built on the losing side of `GetOrInsert` is discarded without any teardown. Harmless today only because modules hold no OS resources — worth wiring once #2 introduces a real removal path.

**9. `scopedTaskScheduler.Close()` second call waits unbounded** (`recovery/task_scheduler.go:107-112`): `_ = s.WaitIdle(context.Background())`. `RecoveryStorage.Close()` is reachable twice (`InterceptorBuildParam.Clear()` at `interceptors/interceptor.go:46` and `walAdaptorImpl.Close()` at `adaptor/wal_adaptor.go:339`), so please use a bounded context here.

**10. `installCheckpoint(nil)` yields a checkpoint with a nil `MessageID`** (`recovery/recovery_storage_impl.go:154-172`); the next `persistDirtySnapshot` calls `snapshot.Checkpoint.MessageID.String()` (`recovery_background_task.go:114`) and `MustMarshalMessageID(nil)`, i.e. a panic on a defensive path. Guard or drop the nil-tolerant branches.

---

### Nits

**11.** `PChannelRecoveryManager.ConsumeDirtySnapshots` re-marks the module dirty whenever it produced anything (`manager.go:227`), and `VChannelRecoveryModule.ConsumeDirtySnapshots` re-marks every segment it consumed (`module.go:254`). Each persist round therefore always costs one extra no-op round per module/segment. If the intent is "residual state may remain", it would be clearer to re-mark only when `HasDirty()` is still true.

**12.** Typo: `persistDritySnapshotWhenClosing` → `persistDirtySnapshotWhenClosing` (`recovery_background_task.go:73`).

**13.** Dead API surface added by this PR — besides the vchannel tombstone methods in #2: `SegmentView.hasPendingDataWorkLocked`, `HasReadyTombstoneFinalize`, `CoveredByTombstone`, `SetSchema`, `HasDirty` (both views), `VChannelView.CanObserveActiveAt`, `canReplayExistingPartitionAtLocked`, `segmentCoveredByVChannelTombstone`, `segmentCoveredByPartitionTombstone`. Several of these encode invariants a reader would assume are enforced; please either wire them or drop them so the enforced set is unambiguous.

---

### Test coverage

The unit tests cover the pieces in isolation well (ref counting, ack tracker, task scheduler, pack writer, transform log). What I could not find is coverage for the paths where the bugs above live:

- graceful close with a tombstoned-but-not-cleanup-ready segment (#1) — a table test over `(persistedMeta, persistedData, tombstoneTick)` triples asserting `Close()` terminates would have caught it;
- drop-collection → assert the vchannel key and its schema keys are gone from the catalog (#2) — master had `recovery_drop_collection_test.go`, which this PR deletes without an equivalent;
- an end-to-end "insert burst then idle" assertion that the data checkpoint eventually advances (#3).

tinswzy:

Reviewed at head `1f845b118b` (merge-base `ea0c533271`), focusing on the two newest commits `7f91abeb9f` and `1f845b118b`. Read-only review — no build/test run.

Overall: the L1 materialization bound is correctly designed and the retained-request/retry mechanism holds up under the invariants I could verify. I found no correctness defect that produces wrong data. The blocking-grade concern is the cost of recomputing the bound on the WAL write hot path; the rest are robustness/observability/test-coverage items. Mergeable after the `high` item is addressed or explicitly deferred with a rationale.

---

## Findings

### high — bound is recomputed with O(#segments) mutex acquisitions on every accepted insert message

`internal/streamingnode/server/wal/vchannel/module.go:518`

```go
func (m *VChannelRecoveryModule) markSegmentViewUpdatedLocked(segmentID int64, view *segment.SegmentView) {
	if view == nil { return }
	m.tryFinalizeSegmentLocked(segmentID, view)
	m.markSegmentDirty(segmentID, view)
	m.refreshTransformMaterializeUpperBoundLocked()   // <— new
}

func (m *VChannelRecoveryModule) refreshTransformMaterializeUpperBoundLocked() {
	upperBound := uint64(math.MaxUint64)
	for _, view := range m.segments {
		if timetick, blocks := view.L1MaterializationBlockerTimeTick(); blocks && timetick < upperBound {
			upperBound = timetick
		}
	}
	m.transformLog.SetMaterializeUpperBound(upperBound)
}
```

`handleInsertMessage` (`module.go:413`) calls `markSegmentUpdatedLocked` for every segment touched by every accepted insert, and `L1MaterializationBlockerTimeTick` (`segment/view.go:384`) takes `info.mu` per view. So each insert message now costs `len(batches) * len(m.segments)` `sync.Mutex` Lock/Unlock pairs plus one `transformLog.mu` acquisition and a `compactTransformMaterializeTasks` pass (`transformlog/task.go:213`) — all under `m.mu`, which serializes the entire vchannel's WAL observation. `SegmentDataUpdated` adds the same cost on every `FlushInsertChunk` completion.

Failure scenario: a collection with a few hundred partitions in one vchannel. `m.segments` retains growing, flushed and pending-cleanup views until `completeSegmentCleanup`, so N is in the hundreds. Every insert append pays hundreds of uncontended-but-real lock round trips inside the critical section that gates the write path — to recompute a value that is unchanged in the overwhelming majority of calls. The bound can only change on three transitions: insert into `m.segments`, delete from `m.segments`, and `finalCommitDone` false→true.

Fix suggestion: maintain the bound incrementally instead of rescanning. `create_segment_time_tick` is immutable per view and `finalCommitDone` is monotone false→true, so a min-structure (heap or sorted set of uncommitted create ticks) owned by the module and mutated only in `handleCreateSegmentMessage`, `completeSegmentCleanup`, and on the commit transition is sufficient. If you prefer to keep the scan, at minimum (a) publish `finalCommitDone` as an `atomic.Bool` and cache the immutable create tick in the module so the scan needs no per-view locking, and (b) skip `SetMaterializeUpperBound` when the recomputed value equals the previously published one.

---

### medium — `shouldMaterialize` is not clamped by the upper bound, so a blocked backlog is rescanned on every flush task

`internal/streamingnode/server/wal/vchannel/transformlog/transform_log.go:455`

```go
func (t *TransformLog) shouldMaterialize(ctx context.Context) bool {
	t.mu.Lock()
	targetTimeTick := t.meta.GetCheckpointTimeTick()   // ignores materializeUpperBound
	t.mu.Unlock()
	rows, bytes, err := t.pendingMaterializeStats(ctx, targetTimeTick)
	...
}
```

and its caller `transformlog/task.go:63`:

```go
if t.log.shouldMaterialize(ctx) {
	t.log.RequestMaterializeThrough(t.log.checkpointTimeTick())
}
```

Failure scenario: the bound pins `materialized_time_tick` at the oldest growing segment's create tick for that segment's whole lifetime. `pendingMaterializeStats` walks `(materialized, checkpoint]`, which grows monotonically over that window, on every single `transformFlushTask` execution. Each walk goes through `nextEntryAfter` → `loadChunk`, which faults cold chunks into `t.chunks` permanently (`transform_log.go:709`, and `truncate` is currently only exercised by tests, so nothing evicts them). The resulting `RequestMaterializeThrough` is then guaranteed to be a no-op, because `newRequestedMaterializeTaskLocked` computes `min(requested, bound) <= materializedTimeTick` and returns `nil` (`task.go:152`). Net: O(backlog) work and growing resident chunk entries per flush, for a request that cannot be satisfied.

Fix suggestion: use `min(checkpointTimeTick, materializeUpperBound)` as the stats target inside `shouldMaterialize`, and return `false` early when that value is `<= materializedTimeTick`.

---

### medium — one stuck L1 final commit silently freezes materialization *and* drop cleanup for the whole vchannel, with no log or metric

`internal/streamingnode/server/wal/vchannel/module.go:527` / `segment/view.go:384` / `segment/task.go:76`

`commitL1SegmentTask.Execute` returns `errors.Mark(err, nodescheduler.ErrDelay)` on any failure, so a failing `lifecycle.CommitL1Segment` retries indefinitely and `finalCommitDone` never flips. The bound then stays pinned at that segment's `create_segment_time_tick` forever. That is not only a materialization stall: `ConsumeCleanupSnapshots` gates partition and vchannel garbage collection on the same frontier —

```go
dropSnapshot, cleanupPartitions := m.vchannelView.TombstonedCleanupPlan(
	cleanup.PhysicalTimeTick,
	m.transformLog.PersistedMaterializedTimeTick(),   // module.go:612
)
```

and `vchannel/view.go:318`, `:340`, `:353` all require `persistedMaterializedTimeTick >= tombstoneTimeTick`. So a single wedged DataCoord commit stops L0 emission, stops dropped-partition cleanup and stops dropped-vchannel cleanup for that vchannel, with no warn log, no lag gauge, and no way to identify the blocking segment from the outside.

Related, and worth stating explicitly in the PR body since it is the normal case rather than the failure case: the bound includes **growing** segments (`l1_commit_done == false` covers them), so in steady state `materialized_time_tick` lags by roughly the lifetime of the oldest growing segment in the vchannel, and drop-cleanup latency inherits that lag. `docs/design-docs/design_docs/wal/transformlog/transform_log.md` §8 documents the invariant but not this expected lag.

Fix suggestion: when the bound has not advanced for longer than a threshold, log a warning carrying the blocking `segmentID` and its create tick; export `latestTimeTick - materializedTimeTick` and the blocker segment id as metrics.

---

### medium — a single materialize batch is no longer bounded by `materializeMaxRows` / `materializeMaxBytes`

`internal/streamingnode/server/wal/vchannel/transformlog/transform_log.go:582`

`prepareMaterialize` accumulates every delete entry in `(materialized, target]` into one in-memory `work.Entries` slice and passes it in a single `Materializer.Materialize` call, with `MaxRows`/`MaxBytes` only forwarded as request hints (`transform_log.go:304`). Before this change the size-based trigger kept `target` near the threshold, so batches were self-limiting.

Failure scenario: the last uncommitted segment in the vchannel commits and the bound jumps from a concrete tick to `math.MaxUint64` in one step. `newRequestedMaterializeTaskLocked` then produces `target == requestedMaterializeTimeTick`, and the entire backlog accumulated while the bound was pinned is built into one slice and handed over as one request — a memory spike proportional to the stall duration rather than to the configured batch size.

Fix suggestion: cap `target` so a single batch does not exceed `materializeMaxRows`/`materializeMaxBytes`, and rely on the existing retained-request mechanism to continue — `requestedMaterializeTimeTick` already survives a partial materialization, so the continuation path exists and only needs a re-trigger after each batch.

---

### low — the new `m.mu → every view.mu` fan-out edge is correct today but undocumented and easy to break

Before this commit `markSegmentViewUpdatedLocked` took only the *updated* view's lock, via `tryFinalizeSegmentLocked`. It now locks every view in the map plus `transformLog.mu`, and submits to the scheduler, all while holding `m.mu`.

I verified this is currently safe:
- `nodescheduler.Submit` is non-blocking by contract and by implementation — unbounded `list.List` queue, no capacity wait (`pkg/util/nodescheduler/scheduler.go:36` and `:166`). So submitting under `m.mu` cannot deadlock on a worker that needs `m.mu`.
- every `SegmentView` → owner callback goes through `NotifyDataUpdated` (`segment/view.go:409`), and all three call sites (`FlushInsertChunk`, `ensureGrowingSegmentTask`, `commitL1SegmentTask`) invoke it *after* releasing `s.mu`.

The hazard is that nothing enforces either property. Any future `owner.SegmentDataUpdated(...)` call made while holding `view.mu` now self-deadlocks, because the refresh re-locks that same view while walking the map.

Fix suggestion: state the order `m.mu → SegmentView.mu → TransformLog.mu` in the `mu` comment at `module.go:46`, and add a "must not be called with `s.mu` held" comment on `SegmentView.NotifyDataUpdated`.

---

### low — `newModule` calls a `...Locked` helper without holding `m.mu`

`internal/streamingnode/server/wal/vchannel/module.go:131`

```go
module.refreshTransformMaterializeUpperBoundLocked()
return module, nil
```

Safe in practice (the module is not published yet, and `requestedMaterializeTimeTick == 0` means no task is produced), but the name asserts a precondition that is not met. It is also the one place in `module.go` that reaches `m.transformLog` without the `if m.transformLog != nil` guard used everywhere else — it works only because `SetMaterializeUpperBound` has a nil-receiver check.

Fix suggestion: split out an `initTransformMaterializeUpperBound` for the constructor path, or take `m.mu` there.

---

### low — `WritePathRecovery` nil-guard was removed from the shard manager (from `7f91abeb9f`)

`internal/streamingnode/server/wal/interceptors/shard/shards/shard_manager.go:133` and `:169` now dereference `recoverInfos.WritePathRecovery` unconditionally. This is correct for the only production producer — `buildInitialRecoverySnapshot` always sets a non-nil value with non-nil maps (`recovery/recovery_stream.go:81`), and `PChannelRecoveryManager.RecoverySnapshot()` also always returns non-nil maps — but `RecoverySnapshot` is a plain exported struct, so a nil field now panics during WAL open rather than degrading.

Fix suggestion: either assert non-nil once at the `RecoverySnapshot` boundary with a clear message, or restore a cheap guard that returns empty maps.

---

### low — test coverage gaps in exactly the paths most likely to regress

The two new tests cover the bound being raised monotonically (`transformlog/materialize_bound_test.go`) and commit-driven retry at module level (`vchannel/materialize_bound_test.go`). Not covered:

1. `completeSegmentCleanup` raising the bound (`module.go:670`) — this is the *other* refresh site and has no test at all.
2. The execution-time clamp at `transform_log.go:278`. I worked through the argument that it is unreachable-to-violate: `target = min(requested, bound)`, `requested` is only ever set from `msg.TimeTick()` of a barrier message (after `syncUpWithMessage` has already advanced the frontier) or from `checkpointTimeTick()`, and a bound *decrease* can only come from a `CreateSegment` whose tick is, in WAL order, greater than every previously observed message tick. The argument holds, but it rests entirely on an invariant that is asserted nowhere in code or tests. A test that lowers the bound between task creation and `Execute` and asserts the emitted `TargetTimeTick` would pin it.
3. A bound *decrease* (no uncommitted segments → `math.MaxUint64` → first `CreateSegment`) with a materialize task already queued.
4. Concurrency: `SegmentDataUpdated` arriving from a scheduler worker while WAL observation holds `m.mu`. Given the new fan-out lock edge, a `-race` stress test here would be worth more than any of the above.

---

### nit

- `RequestMaterializeThrough` and `SetMaterializeUpperBound` (`transformlog/task.go:114`, `:133`) return "a task was submitted", which reads like "the bound was accepted". Both production call sites discard the value; only tests use it. Consider making them `void` and asserting via `HasPendingMaterializeTask` / the recording scheduler.
- `compactTransformFlushTasks` / `compactTransformMaterializeTasks` (`transformlog/task.go:231`, `:242`) filter in place with `tasks[:0]` but never `clear()` the truncated tail, so completed tasks — each holding a `*TransformLog` and a predecessor slice — stay reachable through the backing array up to `cap`. Pre-existing, but this PR puts both on a much hotter path. `takeMessagesThroughLocked` (`transform_log.go:546`) already does this correctly.
- `L1MaterializationBlockerTimeTick` (`segment/view.go:384`) never checks `Stat.Level`, though its name and design doc §8 both say "L1". Correct today — `m.segments` only holds `CreateSegment`-derived views and the L0 output of materialization is tracked separately (`transformlog/materializer.go:123`) — but an explicit level check or a comment would keep it correct if other levels ever enter the map.

---

## Checked and clean

- The proto field renumbering in `7f91abeb9f` (`VChannelMeta.tombstone_time_tick` 6→5, `SegmentAssignmentMeta` 12/13→10/11, dropped `reserved` markers) is safe: `git show ea0c533271:pkg/proto/streaming.proto` shows `VChannelMeta` with fields 1–4 and `SegmentAssignmentMeta` with 1–8 at the merge base, and `load_config` / `sealed_at_data_version` / `data_checkpoint_time_tick` / `data_checkpoint` were all introduced *and* reserved inside this same PR family (`d9f9d645df`). No released build ever wrote those numbers, so no persisted etcd metadata can be misparsed.
- The `AlterWALInfo` → `PChannelControl.GetAlterWalState()` consolidation is semantically equivalent at the two decision points (`adaptor/opener.go:226`, `:315`), and it additionally drops the block in `updatePChannelControl` that re-asserted `Stage: FLUSHING` on every message whenever `alterWALInfo` was set — which would have undone any legitimate transition back to `NONE`. That removal looks like a latent-bug fix, not a regression.
- Submitting to the scheduler while holding `m.mu` is safe (see the `low` item above for the evidence).

czs007:
> **Must-fix issues introduced by this PR: 5**
> **Merge recommendation:** Mergeable after the 5 must-fix items below are fixed in this PR.

Adversarial review of the WAL summary / transform-log recovery path surfaced ten issues, spanning data races on the shared manifest, a term-handoff gap that can permanently lose unmaterialized deletes, and unbounded retry/queue growth on the object-storage failure path.

## Critical

- **`internal/streamingnode/server/wal/walsummary/manager.go:445`** — `publishManifest` marshals the shared `*PChannelSummaryManifest` outside `m.mu` (`recordChunk` under lock, then `WriteManifest(ctx, m.manifest)` unlocked), and `gc.go:60-68` does the same with the same pointer, while `recordChunk` (`store.go:702-705`, in-place `sort.Slice`), `removeChunkEntry` (`gc.go:159-167`) and `removePendingGC` (`gc.go:147-154`) rewrite that backing array in place. GC runs on the recovery-storage background goroutine and flush on a nodescheduler worker, sharing one `Manager`. Failure: a background `gcSummaryStore` → `GCOnce` tick lands while a `summaryFlushTask` is in `flushOnce` → `publishManifest` → both goroutines marshal the same manifest while one compacts/sorts its `Chunks`/`PendingGc` arrays, so the published manifest can omit a chunk that still holds unmaterialized delete records, leaving them unreachable after restart (deleted data resurfaces); `-race` reproduces two distinct races (`store.go:702` write vs `store.go:633` read, `gc.go:154` write vs `store.go:633` read). Suggestion: `proto.Clone` a snapshot inside the lock and marshal the snapshot, and make flush and GC manifest publication mutually exclusive. (surfaced during verification)

- **`internal/streamingnode/server/wal/walsummary/recover.go:74`** — Manifest and chunk keys are term-scoped (`store.go:337-343`, `store.go:321-324`), but `Recover` only reads *this* term's manifest (`store.go:234-235`) and `ProbeChunkForward` filters on `".term<s.term>.psc"` (`store.go:279-284`); no code path reads the previous term's manifest, and the `meta.GetTerm()` already available at `recover.go:55-59` is discarded. Failure: a delete has been flushed into a chunk and published (handle released, WAL checkpoint may pass it) but not yet materialized into L0, then the pchannel is reassigned (`pchannel.go:179` `Term++` on restart/rebalance/access-mode change) → the new term recovers an empty manifest (measured: `DurableTimeTick=0`, `ReadTransformEntries` returns 0 entries, `nextGeneration=0`), so the delete is neither in the manifest nor after the checkpoint and is never materialized (deleted data becomes visible again), while the previous term's chunk objects are unreferenced and never GC'd. Suggestion: carry the previous term forward — read the prior term's manifest during `Recover` and inherit its chunk entries — and align the comments at `store.go:254-255` and `store.go:90-91`, which currently claim inheritance behavior the implementation does not have. (surfaced during verification)

## High

- **`internal/streamingnode/server/wal/walsummary/manager.go:342`** — `flushOnce` claims a generation (`:334`), writes the chunk (`:335`), then `publishManifest` mutates `m.manifest` *before* `WriteManifest` (`:446-448`); on write failure only `restoreStaging` runs, leaving the manifest entry in place and the generation consumed, and the `ErrDelay` retry (`:504-505`) claims the next generation and rewrites the same records. Failure: object storage returns a transient error on the manifest key after the chunk write succeeded, and the task retries → the same batch lands as two chunk objects and two manifest entries (measured: `ReadTransformEntries` returns the `tt=100` delete twice, and `prepareMaterialize` materializes both into L0); with a persistent fault and zero backoff, objects and manifest entries grow without bound. Suggestion: retain the batch and generation after a successful chunk write, retry only republication of the same manifest, install the in-memory manifest atomically on publication success, and add a test that fails writes specifically for the manifest key after chunk persistence. Also note the comment at `manager.go:325-326` states the opposite of the implementation. (surfaced during verification)

- **`pkg/util/nodescheduler/scheduler.go:234`** — `runWorker` requeues on `ErrDelay` and continues immediately; `requeue` only does `PushBack` + `Signal` with no sleep (`:268-278`), and `dequeue` returns as soon as `queue.Len() > 0` (`:250-266`). `summaryFlushTask.Execute` returns `ErrDelay` both when predecessors are unfinished and for *every* `flushOnce` error (`manager.go:500-509`); `transformlog/task.go:53-63` is structurally identical. Failure: object storage is unavailable so `WriteChunk`/`WriteManifest` keep failing (or several flush tasks queue behind one predecessor chain) → each delayed task spins through `Execute` → requeue → dequeue with no sleep, saturating a core; scheduler concurrency defaults to CPU count (`:280-292`), so a sustained fault occupies every worker on the node and starves L0 materialize tasks on the same scheduler. Suggestion: add backoff (and a retry ceiling) on the requeue path, and separate "waiting on predecessor" from "persistence failed" rather than marking all errors `ErrDelay`. Note the originally reported path `internal/streamingnode/server/wal/nodescheduler/scheduler.go` does not exist; the file is `pkg/util/nodescheduler/scheduler.go` at the same lines. (surfaced during verification)

- **`internal/streamingnode/server/wal/walsummary/manager.go:227`** — `SummaryView.ObserveMessage` calls `requestFlush` for *every* payload message once `stagingBytes` exceeds `FlushMaxBytes` (`view.go:289-291`), and `newFlushTaskLocked` copies all outstanding tasks into each new task's `predecessors` (`:229-237`), so task *k* holds a slice of length *k* — O(N²) memory at queue depth N plus an O(N) `predecessorsDone` scan per `Execute` — over an explicitly unbounded queue (`scheduler.go:183-184`). Only `Done` tasks are reclaimed by `compactFlushTasks` (`:520-530`), and a failed `flushOnce` never sets `done` (`:504-508`). Failure: deletes keep arriving during an object-storage outage while the head flush task fails repeatedly → every delete appends another task with an ever-longer predecessor slice, growing memory O(N²) with no cap and burning O(N²) scheduler CPU on predecessor scans, ending in OOM. Suggestion: coalesce flush requests (one pending task per view), bound the queue or the retry chain, and replace the copied predecessor slice with a single reference to the previous task. (surfaced during verification)

## Medium

- **`internal/streamingnode/server/wal/walsummary/manager.go:355`** — `restoreStaging` snapshots the view under `m.mu` (`:356-363`) but performs `view.staging = append(records, view.staging...)` outside the lock (`:374-379`). If `RemoveView` (`:181-197`) interleaves, it deletes the view from `m.views` and calls `takeStagingLocked` on already-drained staging, which returns nil (`view.go:329-332`), so no handle is released; `restoreStaging` then appends the batch to the now-orphaned view. The success path `completeBatch` (`:472-476`) and `restoreStaging`'s own `!ok` branch (`:369-372`) both release unconditionally — only this `ok` branch does not. Failure: a flush fails and is in `restoreStaging` just as the same vchannel's drop-collection cleanup snapshot triggers `completeVChannelCleanup` → `RemoveView` → the batch's `RetainedImmutableMessage` handles are never released, so the pchannel WAL checkpoint can never advance past those messages and the WAL grows without truncation, replaying from that point on every restart. Suggestion: release handles unconditionally in the `ok` branch when the view is no longer registered, or re-check view membership while holding `m.mu` across the append. (surfaced during verification)

- **`internal/streamingnode/server/wal/vchannel/transformlog/transform_log.go:268`** — `materialize` decides whether to chain a follow-up task using the *current* bound while holding `t.mu` (`:268-282`), but `done.Store(true)` happens after the unlock (`task.go:57-59`), and `maybeScheduleMaterializeLocked` returns nil as soon as `len(t.materializeTasks) > 0` (`task.go:87-94`). Failure: an L1 final commit raises `materializeUpperBound` exactly in the window after a materialize task released `t.mu` but before it set `done`, and no further delete or segment event follows on that vchannel → the bound raise is silently dropped (`SetMaterializeUpperBound` reports false), deletes in `(old bound, new bound]` stay in the in-memory pending window, `materializedTimeTick` stalls, and because `chunkReleasedLocked` (`walsummary/gc.go:125-138`) keys the retention floor off that frontier, the corresponding chunks can never be GC'd and the summary store grows. Suggestion: mark the task done before dropping `t.mu` (or re-evaluate the schedule decision after completion) so a concurrent bound raise cannot be swallowed; `materialize_test.go` raises the bound only after `Execute` completes and cannot reach this window. Note the reported path was missing the `vchannel/` component; the file is `internal/streamingnode/server/wal/vchannel/transformlog/transform_log.go` at the same lines. (surfaced during verification)

- **`internal/streamingnode/server/wal/recovery/recovery_storage_impl.go:58`** — `recoverRecoveryInfoFromMeta` now reaches `initRecoveryModules`, whose closing `manager.Start()` (`:281`) spawns `go m.checkpointUpdater.Start()` (`vchannel/manager.go:266`), and WAL replay submits summary flush / materialize tasks to `r.taskScheduler`; but the second-phase failure path just does `return nil, nil, err` (`:58-62`) without `rs.Close()`, `vchannelManager.Close()` or `taskScheduler.Close()`, and the caller receives nil (`adaptor/opener.go:218-220`) so its deferred cleanup cannot reach them. Failure: `recoverRecoveryInfoFromMeta` succeeds and `runBoundedRecovery` then fails (replay stream error, ctx timeout waiting for `EndTimeTick`, or non-nil `rs.Error()`) → the `PChannelCheckpointUpdater` goroutine lives forever, calling `DataCoord.UpdateChannelCheckpoint` on every tick for a pchannel this node no longer serves, and the un-closed scheduler keeps retrying flush/materialize against an abandoned (possibly fenced) summary store; upper layers retry WAL open, leaking another set each time. Suggestion: convert `:58-62` into a deferred cleanup (or call `rs.Close()` before returning) so both phases share one teardown path. Note the issue was originally located at `walsummary/manager.go:137`, which is only a comment on `SetDurableTimeTick`. (surfaced during verification)

## Low

- **`internal/streamingnode/server/wal/walsummary/gc.go:48`** — `GCOnce` ranges over the slice header captured at `:45-47` while the loop body's `removePendingGC` compacts that same backing array in place via `m.manifest.GetPendingGc()[:0]` (`:147-154`), so indices shift under the iteration. Failure: a prior GC round crashes before deletions complete, leaving ≥3 entries in `manifest.pending_gc` → for `[A,B,C]`, deleting A shifts B and C forward so index 1 reads C (B is skipped for the whole round) and index 2 re-deletes the leftover C; the backlog needs several `persistInterval` rounds to drain and abandoned chunk objects linger in object storage far longer. Suggestion: copy the pending list into a snapshot before ranging. Separately, the comment at `:141-143` claims `removePendingGC` publishes the manifest when the queue empties, but the function contains no `WriteManifest`. (surfaced during verification)

- **`internal/streamingnode/server/wal/walsummary/view.go:289`** — `ObserveMessage` releases `v.mu` at `:288` and then reads `v.stagingBytes` unlocked at `:289` to decide whether to call `requestFlush`, while that field is mutated under `v.mu` from nodescheduler workers by `takeStagingLocked` (`view.go:335`, resets to 0) and `restoreStaging` (`manager.go:377`, accumulates). Suggestion: capture `stagingBytes` into a local inside the critical section (or move the threshold check under the lock) — a one-line change. (surfaced during verification)

chyezh:
## Round 4 — E2E cross-term takeover verification (commit d67e062aeb)

Real-environment E2E (standalone + rocksmq + real etcd/minio) across repeated term takeovers (term 2 → 3 → 4 → 5) surfaced **two gaps** in the term-scoped checkpoint protocol introduced in `2a3034116a`; both are fixed in `d67e062aeb`:

1. **Fence did not mirror the term into the in-memory checkpoint** (`fenceConsumeCheckpoint`). The fence stamped term N into the etcd checkpoint via CAS, but `r.checkpoint.Term` stayed 0. The background persistence path re-submits `r.checkpoint` on every interval, so it kept presenting term 0 and was refused by the term pre-check (`recorded term N is newer than publisher term 0`) — the persistence task exited, the channel checkpoint never advanced, and `flush` hung forever.
2. **`consumeDirtySnapshot` built the frozen checkpoint with a zero term**. Even with a correct in-memory term, every persisted advancement carried term 0 and hit the same fence refusal.

Both sites now stamp `r.channel.Term`: the fence mirrors the stamped term into `r.checkpoint` after the successful CAS; the frozen checkpoint persists with the publisher term so the advancement passes the term pre-check and the value CAS of `SaveRecoverySnapshot`.

**Verification (all green):**
- 4 consecutive term takeovers each fence successfully; etcd `consume-checkpoint` term advances 2→3→4→5.
- Term-4 summary objects written as designed: `chunk.0.term4.psc` + `<pchannel>.manifest.4` (term-scoped keys).
- The term-5 instance inherits the term-4 manifest: delete (`pk < 100`) transform data survives the takeover; query returns the post-delete row set (pk 100..119, 20 rows).
- Zero `failed to persist dirty snapshot` after the fix; flush completes; data readable.
- Unit tests: recovery/walsummary packages green; new `TestConsumeDirtySnapshotCarriesPublisherTerm` covers the frozen-checkpoint term.

Note: inserts alone produce no L0 data (L0 is the delete log), so a summary chunk only appears after deletes cross the L0 flush threshold — the empty summary in the first round was expected, not a bug.

czs007:

**This review surfaced two high-severity issues in the new walsummary recovery and flush-retry paths, plus two low-severity hardening notes, all introduced by this PR.**

## High

- **`internal/streamingnode/server/wal/walsummary/recover.go:100` — Term-handoff manifest inheritance only looks back one term, losing un-materialized deletes across chained or failed handoffs.** `Recover` falls back to `recoverManifestOfTerm(Term-1)` only, while the fence (`SavePChannelSummaryMeta` at recover.go:70) is persisted before the seal write and terms are burned by failed opens (`TryAssignToServerID` does `Term++` per assignment attempt). If term N+1 fences but dies before publishing its sealed manifest and term N+2 then recovers → term N+2 reads only the empty N+1 and its own empty manifest, term N's chunks vanish from the manifest chain, its delete records are never materialized to L0 (silently resurrected deleted rows), and the orphaned chunk objects become unreachable to GC. Suggestion: walk terms backward until a non-empty manifest or probed chunks are found (`ProbeChunkForwardOfTerm` already nearly supports prefix discovery of all terms below your own), or persist the fence only after the seal write. *(raised by xiaocai2333, czs007)*

- **`internal/streamingnode/server/wal/walsummary/manager.go:348` — Flush retry after a manifest-write failure livelocks with `storeCorrupted` when the staging batch grows.** If `persistChunk` durably writes generation G with batch `[A]`, `WriteManifest` fails transiently, and any new record is staged during the 100 ms retry window → the retry re-collects `[A,B]`, reclaims G, and `WriteChunk` finds the existing generation-G object with different content, returning `storeCorruptedf` (store.go:190); since `Execute` marks every error `ErrDelay`, this retries forever with an ever-growing batch that can never match — the per-pchannel summary flush stalls permanently, retained handles are never released, the delete checkpoint freezes, and the WAL grows unboundedly until restart. The added test covers only the identical-batch retry. Suggestion: when the chunk write succeeded and only the publish failed, pin the batch and footer to the task and retry only `WriteManifest`, or re-collect only up to the durable generation's max timetick. *(raised by xiaocai2333, czs007)*

## Low

- **`internal/streamingnode/server/wal/walsummary/manager.go:280` — `ReadTransformEntries` iterates the live `Chunks` slice outside the lock.** The slice is captured under `m.mu` but iterated unlocked, while `publishManifest` mutates the same backing array in place (`recordChunk` append + `sort.Slice` via manager.go:489, and the rollback's `removeChunkEntry` compaction at manager.go:494); a concurrent mutation during iteration would skip or duplicate chunk entries. This is latent today — the sole caller is the recovery init path (recovery_storage_impl.go:241), which runs before WAL observation starts, so no flush can currently be concurrent with it. Suggestion: copy the slice under the lock, as `GCOnce` already does for `PendingGc` (gc.go:50-52) — a one-line hardening that protects any future caller. *(raised by czs007)*

- **`pkg/util/nodescheduler/scheduler.go:269` — A delayed entry at the queue head briefly blocks runnable entries behind it.** When a requeued (`ErrDelay`) entry drains to the front while its `nextRun` is still in the future → workers `cond.Wait()` on it instead of scanning past (scheduler.go:269-272), so runnable tasks from other pchannels on the shared global scheduler wait up to the remaining ≤100 ms of the window; bounded extra latency on background summary persistence, worse alongside a perpetually failing task that re-parks at the front each cycle. Both reviewers noted this as acceptable for these background tasks; if it becomes noticeable, letting `dequeue` skip past non-runnable front entries would remove it. *(raised by xiaocai2333, czs007)*

czs007:

Adversarial review of this change surfaced one issue that survived verification, detailed below.

## High severity

- **`internal/streamingnode/server/wal/walsummary/manager.go:243`** — Flush requests arriving while the single flush task is retrying a pinned manifest publish are silently dropped (`newFlushTaskLocked` returns nil whenever a non-done task exists), and once the pinned publish succeeds, neither `flushOnce` (which returns from the `pendingPublish` fast path without collecting staging, `manager.go:353-360`) nor `Execute` (which only marks the task done, `manager.go:591-597`) schedules a successor task. The tracker-driven fallback is permanently deduplicated as well: `persistRequested`/`persistRequestedThrough` are set in `tracker.go:167-168` with no reset when the downstream request is dropped, so the skip condition at `tracker.go:163-164` holds forever once the vchannel goes idle. Failure scenario: an object-store outage pins a batch on the retrying flush task while new deletes are staged and every flush request (threshold- and tracker-driven) is dropped; the outage then recovers but no further payload messages arrive on that vchannel → the staging accumulated during the retry window is never flushed, its handles are never released, and the pchannel WAL checkpoint silently freezes so the WAL can never truncate until process restart. Note the regression test masks this gap by hand-constructing the successor task (`manager_test.go:524-525`) instead of going through `requestFlush`/`newFlushTaskLocked`/the scheduler. Suggested fix: after a pinned publish succeeds, check for residual staging and schedule a successor task (or have `completeBatch`/`Execute` call `requestFlush`), reset the tracker dedup state when a request is dropped or completed, and drive the regression test through the real scheduler path. (raised by sijie-ni-0214, xiaocai2333)

czs007:
> **Must-fix issues introduced by this PR: 5**
> **Merge recommendation:** Mergeable after the 5 must-fix items below are fixed in this PR.

Adversarial review of this PR found one critical and four high-severity correctness issues, all in the new recovery-checkpoint fencing path and the new WAL summary flush path.

## Critical

**`internal/streamingnode/server/wal/recovery/recovery_storage_impl.go:411` — runtime checkpoint advance omits the publisher term, so the legal owner fences itself**
On takeover, `fenceConsumeCheckpoint` stamps the persisted checkpoint with `r.channel.Term` (`recovery_persisted.go:158-162`), but no runtime advance path ever writes a term back: `frozenCheckpoint` is built with only `MessageID/TimeTick/Magic` (`:411-415`) from `ackTracker.Completed()`, and the tracker seed also omits `Term` (`:188-192`), so `completedPoint.Term` is always 0 and that value becomes `snapshot.ConsumeCheckpoint` verbatim (`recovery_background_task.go:220-221`).
Failure: any pchannel that already has a consume checkpoint is reopened (restart or rebalance) with `r.channel.Term=T>0` and the background task issues its first `CheckpointDirty` advance with `Term=0` → `SaveRecoverySnapshot` returns `ErrServiceInternal("...is fenced...")` at `update.go:186-194`, the error is classified non-retryable (`recovery_background_task.go:269-276`), the persist loop returns at `:39-40`, and checkpoint persistence and WAL truncation stop permanently for that pchannel while the WAL grows unbounded and every restart replays from the stale checkpoint.
Suggestion: stamp `Term: r.channel.Term` onto the outbound runtime checkpoint (and/or seed the ack tracker's term) so the CAS sees `T == T`, and add a "fence, then advance once" regression test — `fence_test.go` today asserts only the stamp, never a normal advance after it. (raised by xiaocai2333, sijie-ni-0214)

## High

**`internal/metastore/kv/streamingnode/update.go:195` — the ownership guard rides only on the checkpoint, leaving component-only snapshots and chunked-fallback writes unfenced**
The predicate is attached solely via `CommitSaveIfValue` on the consume-checkpoint key (`update.go:195` → `txn/commit.go:147-153`). The atomic path is safe because the component saves ride the same `MultiSaveAndRemove` (`commit.go:98`), but two paths escape it: (1) snapshots with `CheckpointDirty=false` leave `ConsumeCheckpoint` nil (`recovery_background_task.go:220-222`), so they carry neither the term fast-fail nor any predicate; (2) snapshots exceeding `MaxTxnOps` take `commitFallback`, which flushes all non-commit ops unconditionally and non-atomically via bare `MultiSave`/`MultiSaveAndRemove` (`commit.go:133`, `:197`, `:203`) before the single guarded commit at `:144`. The comment at `commit.go:47-50` claiming non-commit ops "sit inert until the commit txn lands" does not hold — they are durably applied and never rolled back. There is also a TOCTOU window between the fast-fail `Load` (`update.go:175`) and `txn.Commit` (`:197`).
Failure: node A (term 5) and successor B (term 6) briefly coexist after B has fenced, and A emits either a component-only snapshot or an oversized one that hits `commitFallback` → A's vchannel/segment/transform-log keys are written or removed with no predicate, clobbering B's newer state without rollback (A's checkpoint CAS then fails, too late); B's checkpoint has already advanced past those positions, so the WAL range needed to rebuild the clobbered state is never replayed after a restart.
Suggestion: apply the ownership predicate to every fallback chunk (or publish the whole snapshot behind one atomically fenced pointer), and carry a term guard on component-only snapshots as well. (raised by xiaocai2333, sijie-ni-0214)

**`internal/metastore/kv/streamingnode/update.go:208` — first checkpoint creation depends on `CompareVersionAndSwap`, which the TiKV backend does not implement**
When the checkpoint key is absent (`update.go:175-184` sets `checkpointFirstCreation`), creation goes through `metaKV.CompareVersionAndSwap(ctx, checkpointKey, 0, checkpointValue)` (`:206-215`). The TiKV backend hard-returns `ErrServiceUnimplemented` (`internal/kv/tikv/txn_tikv.go:827-830`, marked "under deprecation"), and StreamingNode uses exactly that backend when `metastore.type=tikv` (`internal/distributed/streamingnode/service.go:272-281`). Note the trigger is not takeover fencing — `fenceConsumeCheckpoint` returns early when `r.checkpoint == nil` (`recovery_persisted.go:153-157`) — but the background task's first checkpoint-bearing `SaveRecoverySnapshot`.
Failure: on a TiKV-backed deployment, a pchannel with no checkpoint key yet (fresh cluster, or a newly added pchannel) issues its first persist → `CompareVersionAndSwap` returns `ErrServiceUnimplemented`, which is neither retryable-by-`IsRetryableErr` nor on the non-retryable list in `isRetryableRecoveryPersistError` (`recovery_background_task.go:265-280`), so it is treated as unknown and retried under unbounded backoff; the initial checkpoint is never created, WAL is never truncated, and every restart replays from the head of the WAL.
Suggestion: use a conditional-create primitive supported by every configured metastore backend, or implement the CAS semantics for TiKV. (raised by xiaocai2333, sijie-ni-0214)

**`internal/streamingnode/server/wal/walsummary/manager.go:233` — flush requests are dropped during a pinned-batch retry with no successor scheduled, and the tracker latches its own suppressed request**
`newFlushTaskLocked` returns nil whenever an undone task exists (`:233-243`), so `requestFlush` silently drops the request (`:212-220`). The pinned-publish fast path returns as soon as the publish succeeds, without collecting newer staging or scheduling a successor (`:343-350`), and `Execute` only marks the task done (`:581-587`). Separately, `messageack/tracker.go:168` sets `persistRequestedThrough = maxStalledTimeTick` even though the request was discarded, then skips re-issuing while that TimeTick is unchanged (`:163-165`). The only two production `requestFlush` call sites are conditional (`view.go:119`, `:151`).
Failure: an object-store blip pins the tt10 batch for retry; tt11..tt20 stage and age to stalled, the tracker fires `RequestPersistThrough(tt20)` which is dropped while the latch still advances to tt20; the retry then publishes only tt10 and finishes with no successor → the retained message handles for tt11..tt20 are never released, the ack frontier and the recovery checkpoint stay frozen at tt10, and WAL truncation stalls until unrelated new traffic on that vchannel happens to push staging past the byte threshold or raise `maxStalledTimeTick`.
Suggestion: record a coalesced pending request, and schedule a successor task whenever a completing task leaves staging behind. (raised by xiaocai2333, sijie-ni-0214)

**`internal/streamingnode/server/wal/walsummary/manager.go:358` — a `persistChunk` error whose object actually landed causes a same-generation rewrite with a larger batch, producing a permanent `storeCorrupted` livelock**
The failure handling assumes "nothing was written" (`:329-331`): it rolls the generation back (`:361`) and restores the records to staging (`:362`) so the retry rewrites the same generation. That assumption breaks on the common "server-side PUT succeeded, client timed out" case. `restoreStaging` only prepends to `view.staging`, which `ObserveMessage` keeps appending to concurrently (`view.go:105-119`), so the retry's `collectStaging` returns a strictly larger batch under the same reclaimed generation, and `Store.WriteChunk` hits the `exists == true`, same term/generation, payload-unequal branch and returns `storeCorruptedf` (`store.go:184-189`). `ErrStoreCorrupted` is documented as terminal (`errors.go:22-26`), but `Execute` marks every error as `nodescheduler.ErrDelay` (`:581-587`) and `scopedTaskScheduler.delay` requeues without bound (`task_scheduler.go:153-168`) — the same livelock shape the latest commit fixed for the `publishManifest` path, with no equivalent protection here.
Failure: object-store PUT succeeds server-side but the client errors, and any vchannel on that pchannel stages a new record during the requeue window → each retry builds a larger batch against the same generation and re-hits `storeCorrupted`, so the pchannel's summary flush livelocks forever, staged message handles are never released, and the ack frontier and WAL truncation stall.
Suggestion: pin the batch on chunk-write failure as well (or probe for the object and treat an existing identical/newer chunk as success before retrying), and stop marking `ErrStoreCorrupted`/`ErrStoreFenced` as `ErrDelay` so a terminal store error aborts the flush instead of requeueing. (surfaced during verification)

chyezh:
Handling of must-fix #2 (ownership guard rides only on the checkpoint) and #3 (first checkpoint creation depends on `CompareVersionAndSwap`, unimplemented on TiKV) from the [Round-6 adversarial review](https://github.com/milvus-io/milvus/pull/52542#issuecomment-5395768333):

**Decision: the consume-checkpoint concurrency fence is out of scope for this PR.** I prototyped the full fence (term fast-fail + guarded commit with `NotExist`/value-CAS predicates + post-commit read-back) and then reverted it in `87bd734c59`: the guarantee cannot hold on every supported backend — `CompareVersionAndSwap` is a stub returning `ErrServiceUnimplemented` on TiKV, which `StreamingNode` uses when `metastore.type=tikv` — so the first-write path would hit unbounded retries there.

What this PR now promises instead, and marks as TODO:

- the consume checkpoint is the **commit point** of the update: staged with `CommitSave`, the last write to become visible after every other part of the snapshot (ordering is already guaranteed on both the atomic and the chunked-fallback paths);
- the fenced design (create-if-absent for the first write, value CAS afterwards, read-back verification) and the risk of the un-fenced state (a superseded older-term publisher can still overwrite checkpoint/component state; WAL truncation follows the last writer) are documented as TODO at `update.go` (checkpoint write site), `checkpoint-persistence.md` §5.1 and `summary.md` §2.2;
- side effect of the revert: the `CompareVersionAndSwap` call site is gone entirely, so the TiKV `ErrServiceUnimplemented` infinite-retry path from #3 no longer exists in this PR.

PR description updated with the scope decision and the latest commits. #4 (dropped flush requests during pinned-batch retry + tracker latch) and #5 (storeCorrupted livelock on chunk-write failure) remain tracked separately; #5 is fixed in `6c41b8f16b`.

PR #52556, opened 2026-08-17
enhance: log actual consistency in slow logs

congqixia:
Related to #44703

Slow query logs printed the raw request consistency level. When a client relies on the default consistency (use_default_consistency=true), the request field stays at the protobuf zero value (Strong), so the log did not reflect the level actually used, which comes from the collection's configuration.

Reuse the accesslog mechanism: the actual level is already resolved in task PreExecute and stored on the access info carried by the context. Add ConsistencyLevelHelper that lazily prints that value (ACT-) through mlog.Stringer, falling back to the request level (REQ-) when no access info is available, and wire it into the Search, HybridSearch, and Query slow logs.

Cover the helper with unit tests.

### DIFF
--- a/internal/distributed/proxy/httpserver/handler_v2_test.go
+++ b/internal/distributed/proxy/httpserver/handler_v2_test.go
@@ -42,6 +42,7 @@ import (
 	"github.com/milvus-io/milvus/internal/json"
 	"github.com/milvus-io/milvus/internal/mocks"
 	"github.com/milvus-io/milvus/internal/proxy"
+	"github.com/milvus-io/milvus/internal/proxy/accesslog"
 	"github.com/milvus-io/milvus/internal/types"
 	"github.com/milvus-io/milvus/internal/util/indexparamcheck"
 	"github.com/milvus-io/milvus/pkg/v3/common"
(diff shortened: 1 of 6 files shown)


chyezh:
Potential correctness gap for REST slow logs: `NewConsistencyLevelHelper` can read the resolved consistency level only when the standard context contains `accesslog.AccessKey{}`. The gRPC interceptor injects that value, but the REST `wrapperProxyWithLimit` path calls the Proxy through `HookInterceptor` without injecting it. As a result, REST Search, HybridSearch, and Query requests with `use_default_consistency=true` fall back to the protobuf request zero value and log `REQ-Strong`, even when the collection actually resolves to Bounded or Eventually.

This leaves the original debugging problem unresolved for REST traffic and makes the slow-log value transport-dependent. Could we source the resolved level from the completed task, or bridge an equivalent carrier into the REST context? Please also add a REST-path test with default consistency and a non-Strong collection default, asserting the slow log reports the resolved level rather than `REQ-Strong`.

chyezh:
The REST context bridge introduces a concurrency issue on timeout paths. `AccessLogMiddleware` now stores the same mutable `*info.RestfulInfo` in the request context, while `timeoutMiddleware` runs the REST handler in a goroutine and returns on the timer without waiting for that goroutine to finish. After the timeout handler returns, the outer access-log middleware can call `_globalL.Write(accessInfo)` and format `$consistency_level`, reading `RestfulInfo.actualConsistencyLevel`; at the same time, a late Search/HybridSearch/Query `PreExecute` can still call `SetActualConsistencyLevel` and write that plain pointer.

With access logging enabled and a formatter containing consistency level, this is a Go data race and can also make the timeout log nondeterministically use the request fallback or the resolved level. The existing `TestTimeoutMiddlewareLateHandlerWritesUseCopiedContext` already demonstrates that the handler may continue after the timeout response, so cancellation does not provide synchronization here.

Could we make the resolved-consistency carrier concurrency-safe, for example by protecting this field with an atomic/mutex or using a dedicated concurrency-safe cell? Please also add a timeout-path test combining `AccessLogMiddleware` and `timeoutMiddleware`, with the handler updating consistency after timeout, and verify it under `go test -race`.

Issue #52558, opened 2026-08-17
[Bug]: writing encrypted index entry fails when temp directory does not exist


sparknack:
## Environment

- **Milvus Version**: master (commit f855119cd8)
- **Deployment Mode**: standalone / cluster
- **MQ**: n/a
- **SDK**: n/a
- **OS**: Linux

### Option B: Steps

1. Deploy: any deployment using encrypted index writing (e.g. with `IndexEntryEncryptedLocalWriter`)
2. Setup: configure an encrypted index temp directory (parameter passed as `temp_dir` to the writer) that does **not** exist on the node
3. Trigger: build an index whose entries go through the encrypted local writer
4. Observed via: writer fails at `open()` on the local file

### Trigger Conditions

- **Frequency**: always, when `temp_dir` does not exist
- **Does NOT happen when**: `temp_dir` already exists (or is empty, in which case it defaults to `/tmp`)

## Expected Behavior

The writer should create the missing temp directory automatically and succeed, mirroring the non-encrypted writer behavior.

## Actual Behavior

`IndexEntryEncryptedLocalWriter` assumes `temp_dir` already exists. When it does not, opening the local file fails, so writing an encrypted index entry errors out.

## Error Logs

```
open(.../milvus_enc_<uuid>): No such file or directory
```

## Non-default Configuration

Custom temp directory for encrypted index entries.

## Analysis Hints

- Suspect code location: `internal/core/src/storage/IndexEntryEncryptedLocalWriter.cpp`
- Fix: PR #52557 creates the directory with `std::filesystem::create_directories` and fails fast on error
- Already ruled out: not an SDK issue; failure occurs server-side in the storage layer

PR #52560, opened 2026-08-17
fix: [2.6] create encrypted index temp directory if missing

sparknack:
issue: #52558
pr: #52557

`IndexEntryEncryptedLocalWriter` assumed its temp_dir already existed, so writing an encrypted index entry to a non-existent directory failed at `open()`. Create the directory with `std::filesystem::create_directories` and fail fast on error.

### DIFF
--- a/internal/core/src/storage/IndexEntryEncryptedLocalWriter.cpp
+++ b/internal/core/src/storage/IndexEntryEncryptedLocalWriter.cpp
@@ -21,6 +21,8 @@
 #include <cerrno>
 #include <cstring>
 #include <algorithm>
+#include <filesystem>
+#include <system_error>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


internal/core/src/storage/IndexEntryEncryptedLocalWriter.cpp
```diff
@@ -56,6 +58,12 @@ IndexEntryEncryptedLocalWriter::IndexEntryEncryptedLocalWriter(
 
     auto uuid = boost::uuids::random_generator()();
     std::string dir = temp_dir.empty() ? "/tmp" : temp_dir;
+    std::error_code ec;
+    std::filesystem::create_directories(dir, ec);
```
buqian-zilliz, line 62: **[P3] Auto-created temp directory inherits default (non-private) permissions**

`std::filesystem::create_directories(dir, ec)` creates `dir` with permissions derived from the process umask (commonly `0755`). The temp file itself is opened with mode `0600`, but the parent directory remains world-searchable, so other local users can enumerate the encrypted-index temp entries and observe their sizes/timestamps. Under a permissive umask the directory can also become world-writable without a sticky bit, which would let another local user delete or replace a temp file before `UploadLocalFile()` reads it. Consider setting explicit private permissions (e.g. `0700`) on the created directory instead of relying on the umask.

buqian-zilliz, line 62: **[P3] Auto-creating `temp_dir` can silently mask a missing dedicated temp mount**

`dir` comes from the configured local chunk manager root path (`GetLocalTempDir()` -&gt; `LocalChunkManager::GetRootPath()`). If that path is on a dedicated mount (e.g. a fast local disk intended for index temp data) and the mount is absent at runtime, `create_directories` will now silently create the directory on the parent filesystem and continue, redirecting potentially large encrypted-index temp writes to the wrong disk instead of failing. This can cause disk-pressure/performance regressions and hide an operational misconfiguration. Consider whether a missing configured `temp_dir` should fail loudly or at least log a warning rather than be auto-created on the parent filesystem.

zhengbuqian (approved):

Issue #52180, opened 2026-08-05
[Bug]: DataNode OOM because oversized index and multi-GiB MixCompaction tasks bypass slot admission


yanliang567:
# DataNode OOM because oversized index and multi-GiB MixCompaction tasks bypass slot admission

## Environment

- **Milvus image**: `harbor.milvus.io/milvusdb/milvus:3.0-20260803-6e6323df`
- **Milvus commit**: `6e6323df607299b275ab75600da541e0351aafed`
- **Deployment mode**: distributed cluster, deployed by Milvus Operator
- **Kubernetes**: internal QA cluster (`4am`), namespace `qa-milvus`
- **DataNode resources**: 4 replicas, each `16 CPU / 64 GiB`; slot capacity reported as `128`
- **Storage configuration**:

```yaml
common:
  storage:
    useLoonFFI: true
dataNode:
  storage:
    format: vortex
```

### Test links

- Current failing mixed read/write run (archived): https://argo-workflows.zilliz.cc/archived-workflows/qa/03eeee2d-ed88-46d1-ab32-c9c618f8044e
- Current failing mixed read/write run (workflow view): https://argo-workflows.zilliz.cc/workflows/qa/laion1b-test-cron-sn-nullable-1785852000?tab=workflow
- Previous-image control run without DataNode OOM: https://argo-workflows.zilliz.cc/archived-workflows/qa/ffe43584-5985-4c33-a692-326492be99a4

### Steps

1. Deploy four DataNodes with `16 CPU / 64 GiB` each and build HNSW indexes for a 768-dimensional LAION1B workload.
2. Run the linked seven-hour mixed workload with 20 Locust users and task weights `load/query/search/upsert = 1/4/13/2`.
3. Allow normal background index and Vortex MixCompaction scheduling to proceed.
4. Observe multiple DataNode OOM kills during the workload and another OOM approximately four hours after the foreground workload completed.

### Trigger conditions

- Large HNSW index tasks overlap with multiple multi-GiB MixCompaction tasks on one DataNode.
- MixCompaction input size is not reflected in its slot charge; every task uses only `dataCoord.slot.mixCompactionUsage=4`.
- An index task may require more slots than the entire DataNode capacity, for example `taskSlot=384` on a node with `totalSlots=128`.
- Frequency in this run: five retained `OOMKilled/137` restarts across three DataNode pods.

[...]

PR #45226, opened 2025-11-03
enhance: Subdivide the CPU slot and memory slot of the tasks

xiaocai2333:
issue: #45186 

## Summary

This PR refines the CPU/Memory slot subdivision feature to improve task resource estimation and utilization on DataNode. It ensures **backward compatibility** while introducing index-type-aware memory estimation and size-based CPU scaling, achieving better resource allocation and higher concurrency for small tasks.

### Core Improvements

1. **Vector/Scalar Index-Aware Memory Estimation**
   - Separate memory factors for vector and scalar indexes:
     - Vector indexes: Use `VectorIndexTaskMemoryFactor` (higher memory footprint)
     - Scalar indexes: Use `ScalarIndexTaskMemoryFactor` (lower memory footprint)
   - Provides more accurate resource estimation based on index type characteristics

2. **Size-Based CPU Scaling**
   - Small tasks use reduced CPU to enable higher concurrency:
     - `<10MB`: 1/8 CPU
     - `10-100MB`: 1/4 CPU
     - `100-512MB`: 1/2 CPU
     - `>512MB`: full CPU
   - Applied to both index and stats tasks

3. **Backward Compatibility**
   - Restored original V1 functions: `calculateIndexTaskSlot()`, `calculateStatsTaskSlot()`
   - Created new V2 functions with enhanced logic: `calculateIndexTaskSlotV2()`, `calculateStatsTaskSlotV2()`
   - Restored `GetSlotUsage()` methods for all compaction tasks
   - Old DataNodes can still work with new DataCoord

PR #52561, opened 2026-08-17
enhance: place DataNode tasks on cpu and memory instead of a scalar slot

xiaocai2333:
issue: #52180
3.0: #52982 (cherry-pick of this PR)

Design doc: `docs/design-docs/design_docs/20260827-datanode-cpu-memory-slots.md` (in this PR).

This PR replaces the DataNode-side admission design it started from. The earlier head is `9adbd722dc`.

## What this does

DataCoord places tasks on DataNodes with one scalar "slot". This PR adds two dimensions, CPU and memory. A memory-heavy task is not placed on a worker that cannot hold it, and compute-heavy tasks are spread instead of packed.

### DataCoord estimates, per task family

Every task family implements `GetTaskResource() (taskcommon.Resource, bool)` next to the existing `GetTaskSlot()`. All formulas are in `internal/datacoord/task_resource.go` and are configurable under `dataCoord.taskResource.*`. Each formula mirrors what the worker-side code holds for that family.

| task | CPU | memory |
|---|---|---|
| vector index | 8 | field size × 2 |
| scalar index | 1 | field size × 2 |
| FM-index | 1 | `estimateFMIndexBuildPeakBytes`, the same peak its scalar slot comes from |
| stats (text / bm25 / json key) | 1 | Σ size of the fields the sub job indexes × 2 |
| sort compaction | 1 | segment size × 2 |
| mix compaction / schema bump | 1 | min(Σ input segments, `dataCoord.segment.maxSize`) |
| L0 compaction | 1 | Σ deltalog × 2 |
| clustering compaction | 8 | Σ input segments |
| analyze | 8 | rows × dim × elemSize × 2 |
| import | 1 | Σ over files of min(base buffer × vchannels × partitions, file size) × `importMemoryFactor` |
| preimport | 1 | Σ over files of min(base buffer, file size) |
| copy segment / external refresh | 1 | 64 MB floor |

Every memory estimate is floored at `minTaskMemory` (64 MB).

#### Field size

Index and stats tasks size the field they read as the smaller of two upper bounds (`taskcommon.EstimateFieldSize`):

- the schema's bound. For a fixed-width type it is rows × width. For a variable-width type it is rows × the per-row limit that the proxy enforces on insert: `max_length` for a varchar, `common.JSONMaxLength` for a JSON value, `max_capacity` × element width for an array. The validity bitmap is added when the field is nullable.
- the column group's size. In storage v2/v3 a binlog holds a whole column group, so a field is bounded by every field it shares the group with.

Text and geometry have no schema bound. `checkTextFieldData` does no length check, and geometry is checked only for WKB convertibility. When neither bound is known, the whole segment is charged. The dynamic-field average is not used as a bound: the worker books what the task was priced at, so an average under-prices every value above it.

#### Tasks that cannot be priced exactly

Every answer can be placed. The scheduler never waits for a better number. The `bool` from `GetTaskResource` reports whether the answer is exact, and it controls caching only:

- **Inputs still resolving** (a collection schema that is not cached yet after a DataCoord restart, or index params that are not read back yet): the task is priced at an upper bound on itself, the whole segment. The answer is not cached, so the exact price replaces it in the next round.
- **Inputs gone** (a segment dropped by compaction, drop-collection or drop-partition, or a job that is no longer in meta): the task is priced at the floor. `CreateTaskOnWorker` is the only code that retires such a task, so the task must be placed to reach it.

### The worker books what it is told

The estimate goes in `CreateTaskRequest.properties` (`task_cpu` / `task_memory`, next to `task_slot`). No request payload proto changes. The DataNode `CreateTask` handler reads it once. The executor books it at the same point where it books the scalar slot, and releases it on completion.

`QuerySlotResponse` gets `total_cpu / available_cpu / total_memory / available_memory` on tags 5–8. Tag 4 is `copy_segment_shared_indexes` from #53048. The values are discounted for standalone by `dataNode.standaloneSlotFactor`, and available = total − booked.

A task that arrives without `task_cpu` / `task_memory` (from an older coordinator) books the share of the node that its `task_slot` stands for: `slot / CalculateNodeSlots()` of the node's capacity. Both totals come from the same machine with the same standalone discount. A worker pool that serves several Milvus versions runs keyless and priced tasks together for as long as the pool exists, so the ledger must account for both.

### Two-tier placement

`internal/datacoord/task/node_picker.go`: a worker that reports `total_memory > 0` is placed on the dimensions. Memory is the only hard filter. CPU only ranks (`0.6·memFrac + 0.25·cpuFrac + 0.15·balance`). The scalar slot does not gate a dimensioned worker. Workers that do not report keep the existing max-heap, so a mixed cluster does not regress during a rolling upgrade.

A task that is larger than any worker goes to the worker with the most free memory, and only to a worker that still has free memory. A task that does not fit right now gives way like a backoff task instead of ending the round. The limit is 64 per round (`maxDelayedPerRound`).

## Compatibility

- New DataCoord + old DataNode: no report, so the scalar heap is used. The extra properties are ignored.
- Old DataCoord + new DataNode: no properties, so the worker books the slot-derived share. `available_slots` is reported as before.
- The scalar slot chain (`GetTaskSlot`, `slot_usage` / `task_slot`, `available_slots`, the heap picker) is not changed.
- New config: `dataCoord.taskResource.*` (`vectorIndexCPU`, `analyzeCPU`, `clusteringCompactionCPU`, `defaultCPU`, `indexMemoryFactor`, `statsMemoryFactor`, `l0CompactionMemoryFactor`, `analyzeMemoryFactor`, `importMemoryFactor`, `minTaskMemory`).

## Verification

- `internal/datacoord/...` and `internal/datanode/...`: full UT run locally. All pass except `TestRepublishCopiedManifestIndexes_WritePlacementMatrix`, which fails identically on unmodified master (`fda7d59e39`): it writes to `/files` and gets `Permission denied` as a non-root user. That test and the code it runs are not touched by this PR.
- `pkg/taskcommon`, `pkg/util/paramtable`, `pkg/metrics`: pass.
- `make static-check`: passes (core, pkg, client, go_client, segcore error boundaries: 0 issues).
- Not verified in this PR: a cluster run.

## Known follow-ups

- A DataNode-side correction from the data it actually reads. The writer already computes each column's in-memory size (`calculateActualDataSize` in `internal/storage/record_writer.go`) and then adds it into the column group total, so a per-field size in `Statistics` would replace most of the estimates above.
- Apply the machine-share caps of clustering / analyze / import. These estimates are bounded by their input only, so they over-price a large input.
- An explicit reservation or aging for a large task that a stream of small ones keeps delaying.
- A failed import releases its booking while its sibling futures still hold buffers (`importv2/scheduler.go`). The same window exists for `Slots()` before this PR.
- A memory headroom ratio.


xiaocai2333:
## Design: put both dimensions on the wire, and pick the worker two-dimensionally

Recording a design position on this PR before it goes further.

**Phase 0 keeping the wire one-dimensional is the wrong trade.** The DataNode should report CPU and memory as separate quantities, the coordinator should carry each task's requirement as a vector, and `pickNode` should select on both. The two-phase split (ledger now, proto later) has made this PR *harder*, not easier, and two of its known weak points are direct consequences of the scalar.

### Where we are today

- `QuerySlotResponse` is `{status, available_slots, version}` (`pkg/proto/data_coord.proto:1475`); `GetJobStatsResponse` likewise carries only `available_slots`/`total_slots`. No `.proto` is modified by this PR.
- `WorkerSlots` is `{NodeID, AvailableSlots}` (`internal/datacoord/session/cluster.go:37`).
- `pickNode` (`internal/datacoord/task/global_scheduler.go:256`) is unchanged: a max-heap on one scalar, greedy water-filling.
- The DataNode *does* compute both dimensions, then collapses them before they reach the wire: `legacyAvailableSlots` takes `min(cpuFree, memFree)` and multiplies by the legacy slot total (`internal/datanode/services.go`).

So the only two-dimensional reasoning in the system happens on one side of an RPC that cannot express it.

### Three of this PR's most fragile pieces exist only because of that

**1. The `enableCompactionEstimate` kill switch.** Its own doc says why it is needed: the byte estimate is folded onto *the same* `slot_usage` field the wire has always carried, but on a different scale, so a new DataCoord driving a DataNode that has not restarted reads it as full far too early and cluster-wide compaction throughput collapses.

New proto fields do not have that problem. Old DataNode → new fields absent → coordinator falls back to `available_slots` with its original meaning. Old DataCoord → ignores the new fields, DataNode keeps filling the folded scalar. Nothing collides, so the kill switch is not needed — and neither is `taskresource.LegacyMemoryPerSlot()` (`internal/util/taskresource/params.go:106`), `memoryToSlots` (`internal/datacoord/util.go:594`), or the `legacyAvailableSlots` fold. Those are the three hardest-to-review pieces in the PR and they all exist purely to squeeze bytes through an `int64` slot.

**2. "The DataNode always recomputes locally and never trusts the coordinator"** (`internal/util/taskresource/request.go:33`). This is what breaks on storage v3, which is the format the whole effort targets.

V3 segments do not persist their per-`FieldBinlog` KVs — `internal/metastore/kv/datacoord/kv_catalog.go:327`: *"V3 segments persist paths via the LOON manifest — skip writing the per-FieldBinlog KVs entirely. DataCoord scheduling reads SegmentInfo.Stats instead of iterating these arrays."* `BuildCompactionRequest` fills `FieldBinlogs: segInfo.GetBinlogs()`, so after a DataCoord restart the plan's arrays are empty and only `Manifest` carries the data. `RequirementForCompaction` then computes `TotalMemorySize = 0`, `TotalRows = 0`, and both `estimateStreamingCompaction` (v3 branch) and `estimateSortCompaction` fall to the 64MiB floor — for exactly the v3 mix/sort compactions in #52180. Same root cause hits scalar index builds and stats sub-jobs, whose `InsertLogs` come from the same `segment.GetBinlogs()` (`internal/datacoord/task_index.go:579`, `internal/datacoord/task_stats.go:485`).

DataCoord's own number is *correct* there, because `getSegmentSize()` goes through `EnsureStats()`. So the side that has the right answer computes it and then throws it away, and the side that enforces reconstructs a wrong one. Shipping the requirement in bytes fixes this structurally instead of by patch.

**3. The upgrade-order argument does not actually support a scalar wire.** "DataNode upgrades first, so it must not depend on the coordinator" justifies *having a local fallback*. It does not justify *refusing to use the coordinator's number when it is there*. The correct rule is: use the requirement from the wire when present, recompute locally when absent. That is strictly better than always recomputing, and it is what makes the v3 case work.

### And the fold destroys information the coordinator needs

`min(cpuFree, memFree)` means a node whose vector index build pool is full reports itself full to *every* class. An L0 compaction wanting 200MiB is not offered to a node with 60% of its memory free, because it shares a scalar with builds it contends with nowhere. This is listed under Known gaps, but it is not a gap in the implementation — it is the direct and unavoidable cost of the scalar.

### Proposed shape

Carry a **resource vector**, not two fixed fields. #45226 already spent one attempt on fixed `cpu_slot`/`memory_slot`; the vector index build pool is a *third* resource (a discrete concurrency count, not CPU), so two fixed fields will need a third proto change before long.

```protobuf
message ResourceDim {
  string name = 1;             // "memory_bytes" | "cpu_milli" | "vec_index_build_slots"
  int64  capacity = 2;
  int64  used = 3;
  bool   admission_gating = 4; // false => request-only, may never refuse (CPU)
}

message QuerySlotResponse {
  common.Status status = 1;
  int64  available_slots = 2;      // kept: the folded scalar, for old coordinators
  string version = 3;
  repeated ResourceDim resources = 4;   // new
}
```

The same vector goes on the task side — `CompactionPlan`, `CreateJobRequest`, `CreateStatsRequest`, `AnalyzeRequest`, `ImportRequest` — and on `taskcommon.Properties` (8 `AppendTaskSlot` call sites in `internal/datacoord/session/cluster.go`). Seven `GetTaskSlot()` implementations in `internal/datacoord` need converting; this PR converted one.

### Selection at the coordinator

Keep the asymmetry the DataNode already establishes, so both sides agree on what each dimension means:

- **Memory (and any `admission_gating` dimension) is a hard filter.** It is what OOMs the node.
- **CPU is scoring only, never a filter.** Filtering on CPU at the coordinator would recreate precisely the problem `Requirement.MemoryFitsIn` exists to avoid — an L0 compaction held behind four vector index builds it shares no thread pool with.
- **Oversized keeps today's behaviour**: fits nowhere → dispatch to the emptiest node anyway and let the worker's exclusive admission run it alone.

```
candidates = { n : memFree(n) >= memReq }
if candidates is empty:
    pick argmax memFree(n), dispatch anyway   // worker runs it exclusively

score(n) = w_m * (memFree(n) - memReq) / memCap(n)
         + w_c * (cpuFree(n) - cpuReq) / cpuCap(n)
         - w_b * | memFrac(n) - cpuFrac(n) |   // BalancedAllocation: avoid stranding one dim
pick argmax score
```

One consequence worth measuring rather than assuming: two dimensions have no total order, so the max-heap goes away and this becomes an O(nodes) scan per task per round instead of O(log nodes). At a ~100ms round with a few hundred nodes and a few thousand pending tasks that should still be single-digit milliseconds, but it should be benchmarked; the fallback is a heap on the dominant dimension plus a top-k linear scan.

### What this does not change

It does not replace the DataNode ledger, and it is not an argument against it. `QuerySlot` is a snapshot taken at the start of a scheduling round, and the node charges at *admission*, not when the RPC arrives — which is exactly why the `queuedSlots` arm had to be added to `availableSlots`. The coordinator stays an optimizer for **placement**; the node remains the authority on **admission**. The two are complementary. What should go is the fold, the derived slot exchange rate, and the kill switch that only exists to protect it.

xiaocai2333:
## Review of #52561 (head `875c679`)

Reviewed the full diff at `875c679ac9`. `internal/util/taskresource`, `internal/datanode/resource` and `internal/datacoord/task` build and their tests pass locally (coverage 97.9% / 96.5%). Everything below is a reading of the code, not a runtime observation.

Summary: **needs work — one blocking issue.** The reservation ledger, the guard's locking discipline and the estimator derivations are careful and well documented. The two-dimensional placement layer on the coordinator is not yet doing what its comments claim, and the DataNode ledger has a hole on exactly the storage-v3 path this PR exists for.

---

## Critical

### C1 — storage-v3 compactions are charged the 64MiB floor by the ledger after a DataCoord restart

`internal/util/taskresource/request.go:47` (`RequirementForCompaction`), `internal/datanode/compactor/executor.go:242`

The DataNode deliberately recomputes a compaction's requirement from `plan.SegmentBinlogs` instead of trusting the coordinator. The reason is sound (DataNode-first upgrade order). But the per-FieldBinlog KVs are **not persisted for v3 segments** — `internal/metastore/kv/datacoord/kv_catalog.go:353` skips them outright (`if isV3Segment(segment) { continue }`).

So after a DataCoord restart:

- `segInfo.GetBinlogs()` is empty
- `BuildCompactionRequest` ships a plan whose `FieldBinlogs` are empty
- `compactionSegmentMemory` sums to 0
- `estimateStreamingCompaction`'s v3 branch computes `mem = 0*1.0 + 0`
- `atLeast(mem, binlogChunkBytes())` → **64MiB** (`dataNode.segment.binlog.maxsize` default 67108864)

A 4.5GiB v3 mix compaction is charged 64MiB against a 48GiB budget, so the ledger admits hundreds of them concurrently. That is incident 1's shape (eight concurrent v3 compactions, 36.09GiB of input) — the case this PR is built to prevent — and it fails silently.

The mechanism is already described in this PR, at `internal/datacoord/compaction_requirement.go:27-37`: *"a DataNode reconstructing the requirement from the plan's arrays after a DataCoord restart computes zero and falls to the estimator's 64MiB floor."* The chosen answer was to price from meta on the coordinator side — but the DataNode then discards that number. Both sides know about the hole; neither closes it.

This is the G1 case from `CLAUDE.md`: the transform was verified, its **input** was not.

**Suggested fix.** In `taskRequirement` (`executor.go:112`), detect an under-resolved input — the plan's summed bytes are 0 while `StorageVersion >= 3` (or the segment carries a manifest) — and fall back to `LegacySlotToRequirement(plan.GetSlotUsage())`. Taking `max(locally recomputed, LegacySlotToRequirement(SlotUsage))` unconditionally would also work and is simpler to reason about: an old coordinator's slot is on the old scale, so the max never makes the charge smaller than today's. The same fallback also covers the already-recorded `getDeltaLogSize`-returns-0-on-v3 gap.

---

## High

### H1 — `scoreNode` collapses algebraically to `min(cpuFrac, memFrac)`, and the task's own requirement never enters the score

`internal/datacoord/task/global_scheduler.go:270`, `:306`

The score is `(m+c)/2 - balanceWeight*|m-c|` with `balanceWeight = 0.5`. For `m >= c` that is `(m+c)/2 - (m-c)/2 = c`; otherwise it is `m`. So **`score ≡ min(m, c)`** — an identity, not an approximation.

Two consequences:

1. That is precisely the "take whichever dimension is worse and fold" that the PR criticises the scalar for. With the default weight, two-dimensional placement and scalar placement rank nodes identically; the only genuine increment over the scalar is the memory hard filter. The comment at `:266` — *"This is Kubernetes' BalancedAllocation, and it is the only reason reporting the totals was worth the proto change"* — does not hold as written.
2. `scoreNode(ws)` takes only the node, **not `req`**. A CPU-heavy task and a memory-heavy task therefore produce the same ranking over the same fleet. The doc says the term penalises "how far apart the two dimensions' utilisations **would** sit" — "would" implies after placement, but the code uses the pre-placement figures. k8s' BalancedAllocation computes the spread of `(request + used) / capacity`.

Worked example, using the shape from the design discussion — two 16c/32GiB nodes, w1 with 5 cores / 24GiB free, w2 with 6 cores / 12GiB free, task needs 4GiB:

- w1: `min(0.75, 0.3125) = 0.3125`
- w2: `min(0.375, 0.375) = 0.375` → **w2 wins**

The node with half the memory headroom is preferred. For a design whose stated premise is "memory is incompressible and is the only dimension that may refuse", that is the wrong direction.

`TestScorePenalisesSkew` cannot catch this: even = `min(.5,.5) = .5`, skewed = `min(.125,.875) = .125` — a min-fold and a balanced-allocation score both pass it.

**Suggested fix.** Pass `req` into `scoreNode` and compute both fractions *after* the hypothetical placement; set `balanceWeight` strictly below 0.5 (k8s' effective weight is around 0.25). Add a test with two nodes where min-fold and balanced-allocation disagree, and pin the latter.

### H2 — `pickNodeByDimensions` can never return `NullNodeID`, so the two-dimensional path has no backpressure at all

`internal/datacoord/task/global_scheduler.go:328-366`

`fallback` is assigned for **every** entry that reports dimensions, before the fit check. So whenever any node reports dimensions, `haveFallback` is true, `best == nil` falls back, and a node is always returned. The `if nodeID == NullNodeID { push back; break }` at `:502` therefore never fires on this path: every 100ms round drains the entire pending queue onto workers.

The justification in the comment — *"a task larger than any worker has to run somewhere... refusing here would leave it pending forever, since no node ever grows"* — is true for a task that exceeds a node's **capacity**. It is not true for a node that is merely **full right now**; that node does grow, when its tasks release. The code conflates `req.Memory > ws.MemoryTotal` with `req.Memory > ws.MemoryAvailable`.

Downstream: worker queues grow without bound → `maxTaskQueueNum = 1024` fills → `executor.Enqueue` blocks on `e.taskCh <- task` (`compactor/executor.go:145`) inside the gRPC handler → DataCoord's `CompactionV2` call hangs.

**Suggested fix.** Only genuinely oversized tasks become fallback candidates:

```go
if !memoryFits(e.slots, req.Memory) {
    if ws.MemoryTotal > 0 && req.Memory > ws.MemoryTotal {
        // record as fallback: no node will ever grow to fit this
    }
    continue
}
```

Return `NullNodeID` when nothing fits and there is no over-capacity candidate. `TestPickByDimensionsDispatchesWhenNothingFitsAnywhere` should split into two cases: over capacity → dispatch; merely full → `NullNodeID`.

### H3 — `reportedDimensions` omits the queue arm, reintroducing the hoarding the scalar path explicitly guards against

`internal/datanode/services.go:993` vs `:886`

`availableSlots`' own doc states the problem: *"A node holding five hundred queued compactions that have not been admitted yet has an empty ledger and would advertise itself completely free, so DataCoord's water-filling would keep choosing it and one node would hoard the whole cluster's work."* The scalar therefore takes the worse of the ledger fold and `queueUtilization`.

`reportedDimensions` reports `snap.Total - snap.Reserved` and **never consults the queue**. `pickNodeByDimensions` in turn never consults the scalar. So on the two-dimensional path the failure mode that comment describes is back, unmodified: a node with 500 queued-but-unadmitted compactions has an empty ledger, advertises its full budget, and keeps being chosen.

Compounds with H2.

**Suggested fix.** Fold `queueUtilization` into `memAvail` the same way it is folded into the scalar (`memAvail = min(budget - reserved, budget * (1 - queueUtil))`), or — cleaner — have the executors record a pending charge at enqueue so `Snapshot` reports queued work directly.

### H4 — coordinator-side estimates read **DataCoord's own** memory and CPU

`internal/util/taskresource/estimate_index.go:188`, `estimate_compaction.go:185`, `estimate_import.go:47`

These three call `hardware.GetMemoryCount()` / `hardware.GetCPUNum()`. On a DataNode that means the node. But `analyzeTask.GetResourceRequirement` (`task_analyze.go:98`), `clusteringCompactionTask.GetResourceRequirement` (`compaction_task_clustering.go:97`) and `importTaskRequirement` (`import_util.go:881`) all execute **in the DataCoord process**.

A 4GiB DataCoord prices an analyze task at 3.2GiB; the same task on a 64GiB DataNode is charged ~51GiB and becomes oversized, taking the whole node exclusively. The coordinator believed it would fit alongside anything. Clustering has the same divergence (0.3 × coordinator memory vs 0.3 × node memory), and it now matters, because the two-dimensional path no longer sees clustering's legacy 65535 slot.

`resource_agreement_test.go` pins the compaction family's two-sided agreement, which is exactly the family this does not affect.

**Suggested fix.** Make the node-side hardware figures explicit inputs (`CompactionInput.NodeMemoryBytes`, `ImportInput.NodePoolWidth`, …), filled by the coordinator from `WorkerSlots` — which now already carries `MemoryTotal` and `CPUTotalMilli`. When they cannot be resolved, return the zero `Requirement` ("unknown") rather than substituting the local machine's numbers. Extend the agreement test to these three families.

### H5 — mix compaction on storage v1/v2 drops from 4 slots to 1, quadrupling dispatch concurrency by default

`internal/util/taskresource/estimate_compaction.go:85-96`

The non-v3 branch computes `mem = chunk*1.2*2 + chunk + maxDelete` with `chunk = BinLogMaxSize = 64MiB` — roughly **218MiB, entirely independent of input size**. Then `memoryToSlots(218MiB) = 218/384 = 0 → max(..., 1) = 1`, against a `MixCompactionSlotUsage` default of **4**.

So for every cluster still on v1/v2 storage, this PR lowers the coordinator's price for mix compaction from 4 to 1 at default config — a 4× increase in dispatched concurrency. On the node the ledger charges 218MiB, so a 48GiB budget accommodates ~220 of them; the only remaining bound is `MaxCompactionConcurrency = 10`.

The PR's framing is that it removes over-dispatch; for non-v3 users the direction is reversed. And the model carries an acknowledged unknown — the known-gaps section says α = 1.0 is "empirically grounded but mechanistically unexplained" and that the Go read path cannot account for a 32GiB heap. If that same unexplained retention exists on v2, this opens the gate on it.

**Suggested fix.** Floor the non-v3 branch at `MixCompactionSlotUsage * LegacyMemoryPerSlot()` so the family cannot be priced below its pre-branch water line, or state explicitly (in the PR body, with measurements) that v1/v2 pricing is intentionally being relaxed 4×.

---

## Medium

**M1 — the CPU dimension does not implement the intended asymmetry, and is partly unconfigurable.** `dataCoord.resource.indexBuildCPU` defaults to 1.0 and `EstimateIndexBuild` (`estimate_index.go:89`) applies it to vector and scalar builds alike; `EstimateStats` (`:142`) and `EstimateAnalyze` (`:194`) hardcode `CPU: 1.0`. Only clustering (pool size) and import (0.5/file) differ. `Requirement`'s doc is honest about this ("a flat per-family charge and is known to be a poor model"), but the consequence is that the CPU dimension currently carries almost no information — and combined with H1 the net increment of two-dimensional placement over the old scalar is just the memory filter plus the within-round reservation. Either split the key by `IsVectorIndex()` (vector = `GetCPUNum()/maxVecIndexBuildConcurrency`, scalar = 1) or state plainly that CPU is report-only in this phase and stop describing it as spreading work.

**M2 — the pricing switch is not gated on the same signal placement uses.** `allHaveDimensions` (`global_scheduler.go:227`) already detects a mixed fleet for free. `dataCoord.resource.enableCompactionEstimate` defaults to true and is entirely independent of it, so during a partial rollout a new DataCoord prices on the new scale while an un-restarted DataNode reports availability on the old one — the exact throughput collapse the switch's own Doc describes — and the only remedy is a human flipping the switch. Gate `computeAndCacheTaskSlot` on `enableCompactionEstimate && fleetAllUpgraded`.

**M3 — blocking `Acquire` runs inside bounded worker pools.** `compactor/executor.go:242` (pool width `dataNode.compaction.maxConcurrency` = 10), `index/scheduler.go:262` (vector builds hold a `GetVecIndexBuildPool` slot), `importv2/scheduler.go:129` (blocks the tick goroutine directly). Goroutines waiting for admission occupy pool workers; ten waiters block `Submit`, which fills `taskCh`, which blocks `Enqueue` in the gRPC handler. The import scheduler is worse: admission is serialised on the tick goroutine, so one large task stalls the whole import scheduler and newly-arrived Pending tasks are not picked up meanwhile. Not a deadlock — the guard's empty-ledger exclusive band always lets one task through — but the throughput-collapse path is real, especially with H2/H3 feeding it. Consider `TryAcquire` in the dispatch loop with a separate admission goroutine, so only tasks that already hold a reservation are submitted to the pool.

**M4 — `GetResourceRequirement` is uncached and uses `context.Background()`.** Each call walks every input segment through `meta.GetHealthySegment` under `segMu.RLock`, on the 100ms scheduling path. The caching precondition is already computed (`allResolved`) and then thrown away — `slotUsage` caches on exactly that condition. Separately, all eleven implementations pass `context.Background()`; `schedule()` has `s.ctx` available, and the repo's logging rules ask for a real ctx by priority. Add a ctx parameter to the interface method and cache when `allResolved`.

**M5 — an oversized waiter stops all admission node-wide, but the reported dimensions do not show it.** `guard.go:399`: while any oversized waiter is queued, `blockedByHeadOfLineLocked` returns true for everyone else. Under default config an analyze task is always oversized (0.8 of node memory vs a 0.75 budget), and this is the intended replacement for the old 65535 sentinel. But during the window where analyze is *waiting* rather than *exclusive*, `reportedDimensions` still advertises a positive memory remainder, so the coordinator keeps sending work to a node that will admit none of it. Add `BlockedByOversizedWaiter` to `Snapshot` and treat it like `Frozen`/`Exclusive` (`memAvail = 0`).

**M6 — the PR body contradicts the implementation.** The Wire compatibility section states "**No `.proto` file is modified.** DataCoord still reads a scalar `available_slots`", but `pkg/proto/data_coord.proto` gains four fields and `pickNodeByDimensions` is the new placement path. The "Per-class availability cannot be expressed in phase 0 / the node reports one scalar" gap is likewise stale. A reviewer verifying against the body loses a round. Per G3, both sections need rewriting to claim only what is verified end to end.

---

## Low

- **L1** `internal/datanode/resource/recording_guard.go` — a 178-line test double lives in the production package and ships in the binary. Move it to a `testutil` sub-package or put it behind a `test` build tag.
- **L2** `guard.go:121` — the process-wide `GetGuard()` singleton is called directly by all three executors, so all three packages' tests can only stub it via `mockey.Mock(resource.GetGuard)`. Injecting the `Guard` interface at construction would also make the `stopGlobalGuardForTest` goroutine-leak workaround (`guard.go:137-151`) unnecessary.
- **L3** `guard.go:260` — `tryAcquireLocked` returns `true` for a taskID already on the books **without adding a charge**. The ledger is one int64 namespace shared by compaction planIDs, index buildIDs and import taskIDs; it relies on the global allocator for uniqueness with no assertion. If one ID is ever held by two paths, the second runs uncharged and the first `Release` frees both. Key on `(family, id)`.
- **L4** `import_util.go:881` vs `importv2/task_import.go:115` — the coordinator's `importTaskRequirement` omits `MaxFileMemorySize` while the DataNode's `ImportTask` sets it. The direction is safe (coordinator over-charges), but it breaks the term-for-term symmetry `compactionSegmentMemory` documents as a rule, and no agreement test covers it.
- **L5** `estimate_compaction.go:158-166` — the comment says two clustering tasks read as a full node and a third is offered elsewhere. In practice `clusteringCompactionTask.GetTaskSlot()` still returns 65535, so `queueUtilization` drives the scalar to zero on the *first* clustering task. The described behaviour only holds on the two-dimensional path, which is where H4 applies.
- **L6** `estimate_index.go:139-142` — `EstimateStats`/`EstimateAnalyze` hardcode CPU 1.0 with a comment saying it is "not yet read from config", while every other factor in the file has a config key. Either add the key or lift the value into a named package constant and say it is deliberately fixed.

---

## Dimensions checked with no findings

- **Concurrency safety.** The guard's locking is the strongest part of this change. `wakeWaiters` sends outside the lock; `Acquire`'s retry-before-select ordering (`guard.go:441-460`) is correct against both dropped wakeups and the watermark loop widening the budget; `removeWaiter` re-wakes the queue on exit; `Release` zeroes `reserved` outright on an empty ledger to absorb float residue (`:507-518`). No races, lock-order problems or leaked goroutines found — `stopGlobalGuardForTest` even waits for the in-flight sample to return.
- **Error handling.** No swallowed errors, no panic where an error belongs. `Acquire`'s only failure is ctx, and each executor's unwind (`completeTask(nil)` / stay Pending / `getStateFromError(errCancel)`) matches its module's existing convention. No `fmt.Errorf`; merr conventions respected.
- **Resource leaks.** The import scheduler's move from `closeChan` to ctx+cancel is an improvement; the dead `tryFreeFutures` is removed; no unclosed iterators or connections.
- **Test coverage.** 97.9% / 96.5% on the two new packages, well above the 90% bar, and the three incidents are pinned as regression tests. The gaps are behavioural rather than statement-level: nothing pins the two-dimensional path's lack of backpressure (H2), the missing queue arm on the reported dimensions (H3), or a scoring case that distinguishes balanced allocation from a min-fold (H1).

---

## Suggested order of work

1. **C1** — one fallback in `taskRequirement`; smallest change, largest payoff, and it restores the property the PR is named for.
2. **H2 + H3** — restore coordinator backpressure and make the reported dimensions queue-aware. These two travel together.
3. **H1** — either fix the formula (post-placement fractions, weight < 0.5) or downgrade two-dimensional placement to "memory hard filter + within-round reservation" and describe it as such.
4. **H4**, **H5**, then the PR body (**M6**).

PR #52564, opened 2026-08-17
enhance: [2.6] compact Tantivy JSON row masks

sparknack:
issue: https://github.com/milvus-io/milvus/issues/52565
pr: https://github.com/milvus-io/milvus/pull/52529

Store Tantivy null rows and typed JSON path non-exist rows as CRoaring after build and load.
Retain one resident N/8 EXISTS bitmap so the default non-cached query path only clones it.
Preserve legacy size_t[] entries and scalar index versions while accounting for resident row masks in memory estimates.


sparknack:
will be handled by @SpadeA-Tang , closed.

Issue #52567, opened 2026-08-17
[Feature]: File-based bulk delete and upsert via import (write_mode)


bigsheeper:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Milvus's write path already accepts files: users stage parquet on object storage and submit one `Import` job, which parses them into sealed segments. **Deletion and update have no such path** — `Delete` and `Upsert` are streaming RPCs only. There is no way to remove or overwrite rows from a file.

That gap hurts in exactly the workloads where import is already the natural entry point, because the keys to remove are themselves the output of an upstream batch job and already sit on object storage:

- **Corpus deduplication.** Vector clustering identifies near-duplicate samples and emits the primary keys to drop. Published ratios are large — SemDeDup removes ~50% of LAION-440M with essentially unchanged training quality — so a single pass may delete 20–50% of a collection, in the 10^7–10^9 key range.
- **Training-set pruning.** A quality model scores rows; low scorers are dropped in bulk.
- **Compliance deletion.** Remove every row for a set of business IDs, often under a physical-deletion deadline.
- **Batch rollback.** One ingest batch turned out to be bad; drop it wholesale and re-import.
- **Knowledge-base refresh.** Documents changed; overwrite whole rows by primary key.

Today each of these has to be replayed as millions of individual `Delete`/`Upsert` RPCs. Key lists of this size do not fit in a single RPC, so callers hand-roll batching, retry and progress tracking that `Import` already provides.

### Describe the solution you'd like.

Extend the existing `ImportRequest` with a `write_mode` rather than adding a new RPC:

```
ImportRequest {
  collection_name, partition_name
  files[]                                 // already present, shared by all three modes
  write_mode: APPEND | UPSERT | DELETE    // new, defaults to APPEND
  options[]                               // already present
}
```

| write_mode | what `files` contains | semantics |
|-|-|-|
| `APPEND` | full-schema rows | write only (today's behaviour) |
| `UPSERT` | full-schema rows | write, and delete pre-existing rows carrying those primary keys |
| `DELETE`  | at least the primary key column | delete only |

Notes on the shape:

- **`UPSERT` needs no second file list.** The delete keys are the data file's own primary key column, so all three modes share one `files` field. Submitting the same parquet three times with only `write_mode` changed means "write / overwrite / delete".
[...]

PR #52570, opened 2026-08-17
feat: support file-based bulk delete and upsert via import write_mode

bigsheeper:
## Summary

Adds `write_mode` (APPEND / UPSERT / DELETE) to `ImportRequest`, giving Milvus a file-based path for bulk deletion and full-row overwrite. Until now import could only append; `Delete` and `Upsert` were streaming RPCs only, so a dedup or compliance job holding 10^7–10^9 primary keys on object storage had no way to apply them without hand-rolling batching, retry and progress tracking.

issue: #52567
design doc: `docs/design-docs/design_docs/20260817-import-write-mode.md`

| `write_mode` | `files` contains | semantics |
|-|-|-|
| `Append` | full-schema rows | write only (existing behaviour) |
| `Upsert` | full-schema rows | write, and delete pre-existing rows carrying those primary keys |
| `Delete` | at least the primary key column | delete only |

All three modes share one `files` field — the upsert delete keys are the data file's own primary key column, so the same parquet submitted three times with only `write_mode` changed means "write / overwrite / delete". Extra columns in a delete-key file are ignored, so a dedup job can submit its raw `doc_id | cluster_id | similarity | kept_doc_id` output without projecting first. Progress reuses the import job state machine, so `GetImportProgress` is unchanged.

## Design notes worth a reviewer's attention

**The delete timestamp invariant.** Delete records are stamped with the import request's own timetick, never the commit timestamp. The written rows carry the segment's `commit_timestamp`, and segcore filters deletes against an import segment by `delete_ts > commit_timestamp` — it does not read the row's stored timestamp. So a companion delete removes a pre-existing row and can never remove the row the same job just wrote. Note the `delete_ts <= insert_ts` skip in `DeletedRecord.h` is *not* what protects this path: its insert record is null for sealed segments, so that branch is inert there. §4 of the design doc traces this end to end.

**Segment level became a per-segment property.** An upsert job produces both L1 and L0 segments, so `ImportRequestSegment` gains a `level` field. The row-data path filters L0 out of its candidate set rather than requiring an exact L1 match, because `SegmentLevel`'s zero value is `Legacy` — an older DataCoord's level-less request keeps working under exclusion but would hard-fail under an exact match. L0 sizing is derived from row count, not the row-data byte count.

**Partition routing differs between the two modes, deliberately.** A delete-key file has no partition key to hash by, so delete mode is confined to one partition. Upsert has full rows, so its companion deletes are hashed by both vchannel and partition, mirroring the row hashing — routing them all to one partition would silently leave duplicates on a partition-key collection, since L0 deltas are partition-scoped.

**Two fail-closed gates.** `write_mode=Delete/Upsert` is rejected on collections holding manifest-less (pre storage-v3) data segments, because delivery to a loaded segment happens only through the manifest bump that L0 compaction performs — on such segments the deletes would never take effect, with no error. It is also rejected on autoID collections, where the file supplies no primary keys and the import would otherwise degrade silently to a plain append. The pre-existing `dataCoord.import.enableL0Import` gate is narrowed to legacy backup-restore L0 import rather than reused.

## Visibility

Deletes become visible asynchronously; this does **not** promise read-after-commit. An import-produced delete reaches an already-loaded segment only after L0 compaction folds it into the target segment's deltalog and advances its manifest, which QueryCoord turns into a segment Reopen. Latency is compaction-cycle scale. Lowering it is follow-up work (design doc §10.3).

## Test Plan

- [x] Unit tests across `internal/util/importutilv2`, `internal/datanode/importv2`, `internal/datacoord`, `internal/proxy` — 27 packages green
- [x] `make static-check` — 0 issues across all four modules
- [x] Parquet reader unit test proving a primary-key-projected schema reads a file carrying extra columns
- [x] End-to-end integration tests against a real cluster (`tests/integration/import/write_mode_test.go`)

Observed end to end:

| test | wall clock | observation |
|-|-|-|
| `TestWriteModeDelete` | 81.85s | row count held at 200 across seven 3s polls (~21s), then settled at 100 |
| `TestWriteModeUpsert` | 57.84s | markers 0/200 held ~15s, then flipped to 200/0 in a single poll; total rows never left 200 |

Read these as observations, not as a latency figure for the feature: polling only starts after the import job reports done, and a freshly-loaded test cluster re-pulls its query target for reasons unrelated to this change, so neither number characterises visibility latency on a production collection. Visibility remains compaction-cycle scale as stated above.

The upsert marker transition was atomic across polls — no sample showed a mixture of old and new markers, and none showed 400 rows (companion delete never applied) or 0 (delete removed the job's own rows).

## Known gaps

Stated rather than glossed:

- The CDC ack path (`createImportJobFromAck`) re-checks the cluster-capability gates but not the request-shape rules (autoID, backup, l0_import). This is the pre-existing shape for all three, not something this PR introduces; defense in depth there should cover all three together.
- `syncDeleteOnly` returns early on a mid-loop failure, leaving syncs already dispatched for earlier vchannels unawaited. Same shape as the existing `ImportTask.sync` and `L0ImportTask.syncDelete`; not a regression.
- `BuildDeleteKeySchema` does not clear `IsClusteringKey`. Harmless today because nothing under `internal/util/importutilv2` reads it — an absence of consumers, not an impossibility.
- `commit_ts > T_import` strictness is an ordering argument; the only code-level guard is `commit_ts >= max(binlog.TimestampTo)`, which admits equality. Harmless today because both the segcore and compaction conditions are strict, but relaxing either would be unsafe at that boundary.

## Out of scope

Mixing "delete set A and write unrelated set B" in one job; metadata-only predicate deletion; deletion by non-primary-key columns. All noted in the design doc as future work.


czs007:

Adversarial review of the delete/upsert import path found five issues introduced by this PR, ranging from a partition-key upsert that leaves duplicate primary keys behind to sizing and coverage gaps.

## Critical

**`internal/datanode/importv2/task_import.go:426` — companion deletes for partition-key upserts are routed by the *new* row's partition, so rows migrating across partitions are never deleted.**
`HashDeleteDataForUpsert` (`internal/datanode/importv2/hash.go:132`) hashes the partition from `rawPartKey` taken from the incoming import file, and `syncUpsertDelete` writes the delete into an L0 segment stamped with that real partition ID. Both L0 delivery channels are partition-scoped — `internal/datacoord/compaction_task_l0.go:309` (`taskProto.GetPartitionID() == common.AllPartitionsID || info.GetPartitionID() == taskProto.GetPartitionID()`) and `internal/querynodev2/delegator/delegator_data.go:855` (`segment.Partition() == partitionID || segment.Partition() == common.AllPartitionsID`) — so an L0 carrying a concrete partition ID only ever folds into that one partition. Failure: for a partition-key collection holding key `K` with partition-key value `A` (physical partition `P_a`), an `write_mode=Upsert` file that changes `K`'s partition-key value to `B` (hashing to `P_b`) → the companion delete lands only in `P_b`'s L0, the old row in `P_a` survives forever, the job reports `Completed`, and subsequent query/search return two rows for `K` with no error.
Suggestion: emit the upsert companion deletes as `AllPartitionsID` L0 segments per vchannel — the same branch the delete mode already takes at `internal/proxy/task_import.go:151-161` (`partitionIDs = []UniqueID{common.AllPartitionsID}`) — and have `AssignSegments` allocate L0 only for `AllPartitionsID`. This matches the streaming upsert path, which forces `common.AllPartitionsID` in partition-key mode for exactly this reason (`internal/proxy/task_upsert.go:1541-1542`, comment: "multi entities with same pk and diff partition keys may be hashed to multi physical partitions"). Note that `ImportTaskUpsertPartitionKeySuite.TestUpsertMode_RoutesCompanionDeletesByPartition` (`internal/datanode/importv2/task_import_test.go:680`) asserts today's per-partition routing and must be updated as part of the fix — it is not independent validation of the behaviour.
*(raised by czs007, tinswzy)*

## High

**`internal/datacoord/services.go:1916` — the storage-v3 fail-closed gate only inspects `Flushed` segments, so manifest-less data still in Growing/Flushing is admitted.**
`checkWriteModeSupported` selects "data segments without a manifest" via `isFlushed(segment)`, but `isFlushed` (`internal/datacoord/compaction_trigger.go:917`) matches only `State == Flushed` — not `Flushing` (covered by `isFlush`), and not `Growing`. A Growing segment's storage version is fixed at allocation time: `internal/streamingnode/server/wal/interceptors/shard/shards/segment_alloc_worker.go:149-152` pins `storage.StorageV2` whenever `!requiresStorageV3 && !UseLoonFFI`, and `UseLoonFFI` defaults to false (`pkg/util/paramtable/component_param.go:1165-1169`, `configs/milvus.yaml:1120`), so those segments will flush with no manifest (`internal/datacoord/handler.go:659`). Failure: with `common.storage.useLoonFFI=false` (default), bulk-write into a fresh collection and submit a `write_mode=Delete` import before any segment flushes → the gate finds no legacy segment, accepts the request, and the job completes; the segments then flush as StorageV2 without a manifest, the L0 deletes can never reach them by this PR's own argument, and rows the user believes deleted remain visible to query/search — precisely the case the gate exists to reject.
Suggestion: predicate the gate on storage version / write configuration (include Growing and Flushing segments, or check `common.storage.useLoonFFI` directly) rather than scanning only already-flushed segments.
*(raised by tinswzy)*

## Medium

**`pkg/proto/internal.proto:429` — the new `write_mode` proto field is unreachable in production, and both new e2e tests drive only that dead path.**
`ImportRequest` gains `ImportWriteMode write_mode = 6`, and `normalizeImportWriteMode` (`internal/proxy/task_import.go:288-304`) reconciles it against `options["write_mode"]`. But both production constructors of `internalpb.ImportRequest` leave it unset — `internal/distributed/proxy/httpserver/handler_v2.go:3871` (RESTful `/v2/vectordb/jobs/import/create`) and `internal/proxy/impl.go:4876` `convertToV2ImportRequest` (gRPC `MilvusService/Import`) map only `DbName`/`CollectionName`/`PartitionName`/`Files`/`Options`; `proxy.proto:27`'s `ImportV2` is node-to-node ProxyService, not a client-facing surface. So `req.GetWriteMode()` is always `Append` for real clients, and the `fromField` branch plus the conflict check at `:299-301` are unreachable outside tests. Meanwhile both new e2e tests inject the mode through that field (`tests/integration/import/write_mode_test.go:358`, `:449`), leaving the only user-reachable input form (`options["write_mode"]`) with unit coverage only — the full pipeline (preimport → `AssignSegments` levels → L0 write → L0 compaction) is never exercised end-to-end from a real entry point.
Suggestion: either wire the field through both constructors and keep the reconciliation, or drop field 6 and the `fromField` branch and keep the option as the single input. Either way, at least one e2e test should submit via `Options`.
*(raised by czs007)*

**`tests/integration/import/write_mode_test.go:81` — the integration schema has no partition key, so the one defective routing branch has zero e2e coverage.**
`buildWriteModeSchema` declares only an int64 primary key, a float vector, and a varchar field — no partition key — so the collection has a single partition and `hashByPartition` (`internal/datanode/importv2/hash.go:327-332`) degenerates to a closure returning constant `0`. Neither integration test can reach the companion-delete partition routing, which is exactly where the critical issue above lives; and the only unit test covering it asserts the current defective behaviour.
Suggestion: when fixing the critical issue, add an e2e case with a partition key where the upsert file changes a row's partition-key value.
*(raised by tinswzy)*

## Low

**`internal/datacoord/import_util.go:237` — delete-mode L0 sizing uses a 64-byte constant and discards the measured byte counts preimport already computed.**
`GetDeleteStats` (`internal/datanode/importv2/hash.go:287`) already accumulates the true per-record size as `pk.Size() + 8` into `PartitionDataSize`, which flows into `hashedDataSize` at `internal/datacoord/import_util.go:204-206`; the L0 allocation then discards `size` entirely in favour of `levelSize = rows * estimatedDeleteRecordSize` (constant 64, defined at `:176-178`). For upsert, `size` is a whole-row byte count and an estimate is unavoidable — but in delete mode the only level is L0 and the measured value is directly usable. With long varchar primary keys (e.g. a 200-byte key, 208 bytes/record) the estimate is low by more than 3×, distorting the intended L0 segment split against `dataNode.segment.deleteBufBytes`.
Suggestion: use the measured `PartitionDataSize` for L0 sizing in delete mode, keeping the constant only for the upsert path.
*(raised by tinswzy, czs007)*

czs007:

Adversarial review of the import write-mode path surfaced two correctness defects in the newly added upsert/delete routing and its fail-closed gate, plus one documentation inconsistency.

## Critical

**`internal/datanode/importv2/hash.go:133` — companion deletes for upsert are routed by the *new* row's partition key, so cross-partition updates leave the old row behind**

`HashDeleteDataForUpsert` derives the delete's partition from the incoming file's partition-key value (`p1, p2 := f1(rawPk), f2(rawPartKey)`; `rawPartKey = partKeyData.GetRow(i)` at hash.go:128), and `syncUpsertDelete` (`internal/datanode/importv2/task_import.go:442`) writes it into an L0 segment stamped with that concrete `partitionID` via `PickSegmentByLevel(..., SegmentLevel_L0)`. L0 compaction only selects data segments in the same partition (`internal/datacoord/compaction_task_l0.go:311`), so a delete written under the new partition can never reach the old one. **Failure:** on a partition-key collection, a `write_mode=Upsert` file that moves key `K` from partition-key value `A` (physical `P_a`) to `B` (`P_b`) writes the delete only into `P_b`'s L0 → the old row in `P_a` is never compacted away, the job reports `Completed`, and the collection returns two rows for `K` with primary-key uniqueness silently broken.

This path diverges from Milvus's existing semantics for the same scenario: streaming upsert explicitly fans the companion delete collection-wide — `internal/proxy/task_upsert.go:1542-1547`, `if it.partitionKeyMode { // multi entities with same pk and diff partition keys may be hashed to multi physical partitions ... it.upsertMsg.DeleteMsg.PartitionID = common.AllPartitionsID }` (repeated at task_upsert.go:206-211). This PR's own delete mode is already correct (`internal/proxy/task_import.go:151-153` routes to `common.AllPartitionsID`); only upsert falls through to the else branch and receives the full hashed partition list. The design doc §7 argues for per-partition routing on the grounds that "routing to a single partition is wrong", but does not consider `AllPartitionsID`, which is the established answer.

*Suggestion:* route the upsert companion delete through an `AllPartitionsID` L0 segment (matching streaming upsert), or reject upsert on partition-key collections until the old partition can be resolved. Note that `internal/datanode/importv2/task_import_test.go:676-680` (`TestUpsertMode_RoutesCompanionDeletesByPartition`) asserts exactly the current per-partition routing, so it encodes the defect and must be inverted as part of the fix rather than treated as passing evidence. *(raised by sijie-ni-0214, tinswzy)*

## High

**`internal/datacoord/services.go:1935` — the storage-v3 fail-closed gate only inspects `Flushed` segments, so Growing/Flushing segments bypass it**

`checkWriteModeSupported` (services.go:1922-1948) selects legacy segments with `isSegmentHealthy(segment) && isFlushed(segment) && segment.GetLevel() != datapb.SegmentLevel_L0 && segment.GetManifestPath() == ""`, but `isFlushed` is defined as exactly `State == commonpb.SegmentState_Flushed` (`internal/datacoord/compaction_trigger.go:917-919`); `Flushing` (covered by `isFlush` at :921-923) and `Growing` are outside the predicate. **Failure:** with `common.storage.useLoonFFI=false` (default), write into a fresh loaded collection and submit a `write_mode=Delete` import before those segments flush → the gate finds zero legacy segments and admits the job; the rows later land as StorageV2 with no manifest, L0 compaction folds the deletes into deltalogs without bumping the manifest (`internal/datacoord/compaction_task_l0.go:499`, `AddL0DeltalogsAndUpdateManifestOperator`), QueryCoord's SegmentChecker never triggers a Reopen, and the deletes stay invisible indefinitely while the job reports `Completed`.

Widening the state filter is safe: a Growing segment's manifest is fixed at creation (`internal/datacoord/segment_manager.go:418-422`, `if req.StorageVersion == storage.StorageV3 { ... manifestPath = packed.MarshalManifestPath(...) }`), and the storage version is decided at allocation time (`internal/streamingnode/server/wal/interceptors/shard/shards/segment_alloc_worker.go:148-152`), so v2 Growing segments always have an empty `ManifestPath` and v3 ones always have a manifest — the existing manifest-path criterion holds for all three states without producing false rejections.

*Suggestion:* replace `isFlushed` with a predicate covering Growing, Flushing, and Flushed, and add explicit unit tests for the first two states (`import_services_test.go` currently covers only Flushed). *(raised by sijie-ni-0214, tinswzy)*

## Nitpick

**`internal/datanode/importv2/delete_extract.go:13` — the comment justifies the timestamp choice with an invariant this PR's own design doc explicitly retracts**

The comment at delete_extract.go:12-15 attributes the "use the request timetick, not the commit timestamp" decision to segcore's skip rule: `segcore skips a delete when delete_ts <= insert_ts, so using the request timetick lets a delete remove pre-existing rows while leaving rows written later by the same job (stamped at commit time) intact.` The design doc added in the same PR states the opposite at `docs/design-docs/design_docs/20260817-import-write-mode.md:133`: the `delete_ts <= insert_ts` skip in `DeletedRecord.h` "is *not* what protects this path. It reads the insert record, which is null for sealed segments, so that branch is inert there. Reasoning about this feature from that rule leads to a wrong model of how much margin exists." The actual protection is the load-time `commit_timestamp` overlay (design doc §4, point 3).

*Suggestion:* rewrite the comment to cite the `commit_timestamp` overlay instead, and check the similar wording near the `task_preimport.go` call site for the same claim. *(surfaced during verification)*

czs007:

Both findings from the previous round remain unresolved in the current head; the incremental change since the last review was a rebase plus test-only adaptation, with no production code changes.

## Critical

**`internal/datanode/importv2/hash.go:128`** — Upsert companion deletes are routed by the *incoming* row's partition key. `HashDeleteDataForUpsert` computes `p2` from the import file's partition-key value (`f2(rawPartKey)`) and appends the companion delete only to `res[p1][p2]`; `syncUpsertDelete` (`internal/datanode/importv2/task_import.go:441-443`) then stamps that L0 segment with a concrete `partitionID` rather than `common.AllPartitionsID`. Failure: an existing row `pk=1, partition_key="A"` (hashed to physical partition `P_A`) is upserted with `partition_key="B"` (hashed to `P_B`) → the new row and its companion delete both land in `P_B`, and `l0CompactionTask.selectFlushedSegment` (`internal/datacoord/compaction_task_l0.go:309`) plus delegator-side delta forwarding (`internal/querynodev2/delegator/delta_forward.go:183`) both filter by partition ID, so the stale row in `P_A` never receives the delete — two visible rows for the same primary key, primary-key uniqueness silently broken, and the import job still reports `Completed`.

Suggestion: broadcast the companion delete to all partitions (aligning with the streaming upsert path, which sets `DeleteMsg.PartitionID = common.AllPartitionsID` in `internal/proxy/task_upsert.go:1541-1545` for exactly this reason), or enforce an invariant that reliably prevents a partition-key change on upsert. This also requires widening the `AllPartitionsID` branch condition at `internal/proxy/task_import.go:151`, which is still `else if isL0Import || isDeleteMode` and therefore lets upsert fall through to the full physical-partition list. Note that `TestUpsertMode_RoutesCompanionDeletesByPartition` in `internal/datanode/importv2/task_import_test.go` currently asserts the buggy behavior as expected, and must be inverted along with the fix. *(raised by sijie-ni-0214, tinswzy)*

## High

**`internal/datacoord/services.go:1934`** — The storage-v3 fail-closed gate only inspects `Flushed` segments. `checkWriteModeSupported` filters with `isSegmentHealthy(segment) && isFlushed(segment) && level != L0 && GetManifestPath() == ""`, and `isFlushed` (`internal/datacoord/compaction_trigger.go:917-919`) matches only `SegmentState_Flushed`, so healthy `Growing` and `Flushing` non-L0 segments are excluded from the check. Failure: on a cluster with the default `common.storage.useLoonFFI=false`, write rows into a new collection and submit a `write_mode=Delete` import before any flush → `SelectSegments` finds no legacy segments, the gate admits the job, those segments later land as StorageV2 without a manifest, and L0 compaction folds the deltalogs via the `GetManifestPath() == ""` branch (`internal/datacoord/meta.go:1306-1311`) without bumping `ManifestPath` or `DataVersion`; neither `isSegmentUpdate` trigger in `internal/querycoordv2/checkers/segment_checker.go:342-357` fires, the querynode never reopens, and the deletes become permanently invisible while the job reports `Completed`.

Suggestion: check every healthy non-L0 data segment regardless of `Growing`/`Flushing`/`Flushed` state (compare `isFlush` at `compaction_trigger.go:921-922` or `isFlushState` at `meta.go:3215-3217`, both of which cover `Flushing`), and add unit coverage for the `Growing` and `Flushing` states — `internal/datacoord/import_services_test.go` currently constructs only `State: commonpb.SegmentState_Flushed` segments. *(raised by sijie-ni-0214, tinswzy)*

czs007:

This PR's two previously flagged correctness defects remain unchanged at the current head; the only new commit is a test-identifier rename, and no author response has been posted on either item.

## Critical

- **`internal/datanode/importv2/hash.go:131`** — Upsert companion deletes are routed by the *new* row's partition key. `HashDeleteDataForUpsert` hashes each companion delete with `hashByPartition` on the incoming row's partition-key value, and `syncUpsertDelete` (`task_import.go:442-447`) binds that bucket to one concrete physical partition's L0 segment. Every L0 consumer is partition-scoped (`compaction_task_l0.go:309-310`, `delegator_data.go:859`), and the proxy hands the datanode the full physical-partition list for upsert rather than `AllPartitionsID` (`proxy/task_import.go:150-186`). Partition-key collection with pk=1 stored in P_A, import with `write_mode=Upsert` carrying pk=1 under partition key "B" → new row and its companion delete both land in P_B, the stale row in P_A is never tombstoned, queries return two rows for one PK, and the job reports Completed. Streaming upsert avoids this exact case by setting `DeleteMsg.PartitionID = common.AllPartitionsID` in partition-key mode (`proxy/task_upsert.go:1546-1550`). The existing test `TestUpsertMode_RoutesCompanionDeletesByPartition` asserts the per-partition routing and therefore encodes the defect. **Suggestion:** for upsert on partition-key collections, hash companion deletes by vchannel only and write them into an L0 segment tagged `AllPartitionsID` (with `AssignSegments` allocating such a segment), or reject `write_mode=Upsert` on partition-key collections. Add a test where the same PK appears with two different partition-key values and assert the original row is deleted; invert the existing test accordingly. *Previously raised in round 1; the author has not responded, and the code at `hash.go:130-133`, `task_import.go:442-443`, and `proxy/task_import.go:151` is byte-identical to the last reviewed head.* (raised by tinswzy, czs007)

## High

- **`internal/datacoord/services.go:1936`** — The storage-v3 fail-closed gate in `checkWriteModeSupported` only inspects Flushed segments. The legacy predicate requires `isFlushed(segment)`, which matches only `SegmentState_Flushed` (`compaction_trigger.go:917-919`), so healthy Growing, Sealed, and Flushing non-L0 segments with no manifest are never counted. Under the default `common.storage.useLoonFFI=false`, every streaming-created segment is StorageV2 with `ManifestPath=""` fixed at creation (`segment_alloc_worker.go:150-151`, `segment_manager.go:418-437`). Default config, loaded collection with a Growing/Sealed/Flushing data segment, submit `write_mode=Delete` or `Upsert` against its PKs before it reaches Flushed → the gate finds zero legacy segments and admits the job; the job reports Completed but the deletes never reach the loaded segment, because growing rows stay visible while the segment stays growing, and for a Flushing segment already loaded as sealed the delta is folded into deltalogs with no `ManifestPath`/`DataVersion` bump (`meta.go:1314-1319`), so `SegmentChecker.isSegmentUpdate` never triggers a Reopen and the rows remain visible until release and reload. This is the corruption design doc §6 says the gate exists to prevent. The same gate is re-applied in `createImportJobFromAck` (`services.go:2085`). **Suggestion:** drop the `isFlushed` conjunct and count every healthy, non-L0, non-dropped data segment with `ManifestPath == ""` (cf. the Flushing-inclusive `isFlush` at `compaction_trigger.go:921-922`). This is safe because a segment's manifest path is decided at creation and never changes from empty for v2 segments. All fixtures in `TestCheckWriteModeSupported` use `State: SegmentState_Flushed`; add Growing and Flushing manifest-less cases expecting rejection. *Previously raised in round 1; the author has not responded, and `services.go:1934-1939` is unchanged apart from an unrelated rebase hunk below it.* (raised by tinswzy, czs007)

PR #52389, opened 2026-08-11
enhance: [cp3.0] persist segment summary metadata on SegmentInfo (#50410)

tedxu:
Cherry-pick from master
pr: #50410
issue: #50406

Persist a Statistics summary on SegmentInfo so DataCoord
no longer needs to iterate FieldBinlog arrays for
scheduling decisions. Prerequisite for removing per-field
binlog KV persistence for V3 segments.

Conflicts resolved:
- internal/datacoord/segment_info.go: kept 3.0 SetLevel
  functions alongside new Stats comments
- pkg/proto/datapb/data_coord.pb.go: regenerated from
  cleanly merged proto source

PR #52571, opened 2026-08-17
enhance: [cp3.0] improve Storage V3 segment resource estimation

sijie-ni-0214:
Cherry-pick from master
pr: #52284

issue: #52282

## Test Plan

- `make generated-proto-without-cpp`
- Related packages compile with `-tags dynamic,test -gcflags="all=-N -l" -ldflags="-r ${RPATH}"`
- Statistics load-resource descriptor tests passed
- QueryCoord SegmentLoadInfo packing tests passed
- QueryNode segment-estimation tests passed
- `gofumpt` and `git diff --check` passed
- Full package tests are blocked by unavailable local etcd/MinIO/Azure services
- `make build-go` is blocked by stale local CMake Analyze C API artifacts (`Analyze` argument mismatch)


czs007:

This review flags issues in the new `LoadResource` estimation and manifest stats-path handling, records one real but pre-existing hazard for follow-up, and adjudicates the three findings carried over from the upstream review of #52284.

## Medium

- **`internal/storage/statistics.go:97` — Under-counted `LoadResource` drives the segment resource estimate for V3 segments in the upgrade window.** `NewStatisticsCollectorFromStats` restores `insertBinlogSize` unconditionally but seeds `columnGroups` only when `stats.GetLoadResource()` is non-nil, and `Digest` merges only the current sync's inserts, so Stats persisted by a pre-upgrade binary yield a `LoadResource` covering only post-upgrade syncs. During a rolling upgrade, if such a segment flushes unsorted and DataCoord then restarts (dropping the never-persisted V3 `BinlogPaths`), the querynode's admission estimate under-counts the segment's memory/disk → over-admission and memory-pressure/OOM risk until sort compaction rebuilds the Stats. The estimate is still strictly better than the pre-PR value of zero in the same state, and the under-count is bounded to unsorted segments in the upgrade window, but there is currently no cross-check (`TestResolveSegmentEstimateLogs` pins `InsertBinlogSize = 99` against a 100-byte group and still passes). Suggestion: add a `sum(groups.MemorySize) < InsertBinlogSize` residual guard in `resolveSegmentEstimateLogs`, a comment at the synthetic-binlog construction, and a test asserting the residual behavior. *(raised by bigsheeper, tinswzy)*

## Low

- **`internal/util/segcore/segment.go:463` — V3 manifest-read failure falls back to legacy base-path resolution, producing bogus text-index stats paths.** `resolveStatsWithBasePaths` logs a warning on manifest-resolve failure and silently falls through to the V2 branch, which builds base paths via `metautil.BuildTextIndexPrefix`, while V3 text-index files live at absolute paths under `{manifestBase}/_stats/text_index.{fieldID}`; `convertTextIndexStats` only strips `basePath + "/"`, which cannot match. Because this PR now ships `TextStatsLogs` unconditionally for manifest segments, a transient manifest read failure while loading a StorageV3 segment with a text index → segcore receives a nonsense legacy-prefix + absolute-path combination, the load fails with a misleading not-found error and retries, and a persistent manifest failure leaves the segment unloadable (pre-PR the same trigger silently served queries without the text index; the equivalent JSON-stats exposure was pre-existing). Suggestion: return the error instead of falling through (or gate the V2 fallback on `StorageVersion != StorageV3`), cover `jsonBasePaths` in the same change, and align the doc comment at `segment.go:462` with the actual unconditional-resolve behavior. *(raised by tinswzy, bigsheeper)*

- **`internal/querynodev2/segments/segment_loader.go:1975` — Logical estimator PART 1 still hard-errors on an index over a dropped field, unlike its loading twin** (`:2177-2182`, which logs "skip index for dropped field" and continues). Loading a sealed segment whose `IndexInfos` reference a field dropped from the current schema → `compactLoadInfoForRuntime` aborts, `resourceUsageCache` is set to nil, and the full uncompacted `SegmentLoadInfo` is retained in querynode memory for the segment's lifetime, plus a misleading warn log per segment. This is unrelated to the PR because the PART 1 index branch is untouched by the diff and hard-errored identically before it. The follow-up fix is the same one-liner as the loading twin (log and continue). *(raised by tinswzy, bigsheeper)*

## Carried over from #52284

- **Logical vs. loading estimator inconsistency on dropped fields (`segment_loader.go:2045`, medium): Resolved upstream.** The fix came across with the cherry-pick — reviewers verified the skip-and-continue branch at `segment_loader.go:2045-2048` is behavior-identical to the loading twin at `:2255-2260`, with a regression test covering a dropped-field group.
- **Incomplete `LoadResource` preferred over complete `BinlogPaths` (`segment_loader.go:1920`, medium): Confirmed on this branch, with the mechanism corrected.** The "takes priority over complete `BinlogPaths`" half was refuted — legacy arrays win whenever non-empty (`segment_loader.go:1913-1917`) — but the under-count itself stands and is consumed via the synthetic path (`statistics.go:97-102`, `segment_loader.go:1917-1940`); it is now tracked as the medium finding at `statistics.go:97` above.
- **`estimateLogicalResourceUsageOfSegment` not wired to `resolveSegmentEstimateLogs` (`segment_loader.go:1956`, low): Resolved upstream.** Reviewers verified the wiring is present on this branch (`segment_loader.go:1959`, with PART 4 consuming the resolved deltalogs at `:2104`).

sijie-ni-0214:
Regarding the first issue:

This PR is released together with #52389, and our production upgrade order is DataCoord/Coord first, followed by StreamingNode/SQN/QueryNode. Therefore, the mixed-version scenario described in the review is not expected during the supported rollout sequence.

In addition, `resolveSegmentEstimateLogs` now preserves existing `BinlogPaths` and `Deltalogs` whenever they are present, and only falls back to the V3 `LoadResource` descriptor when the legacy arrays are empty. This provides an additional defensive guard against using an incomplete descriptor over complete legacy metadata. We consider this issue addressed for the supported release and upgrade model.

Regarding the second issue:

Propagating `TextStatsLogs` for V3 segments is intentional and necessary. QueryNode/SQN uses this metadata for text-index resource estimation and recovery, while the V3 manifest remains the source of truth for the actual text-index file paths.

We agree that, if manifest stats resolution fails, falling back to the legacy V2 text-index base path can produce an invalid path. However, `JsonKeyStatsLogs` already has the same pre-existing behavior for V3 segments. We plan to address the V3 text/JSON stats fallback handling together in a separate follow-up PR, rather than removing the required `TextStatsLogs` propagation or blocking this backport.

czs007 (approved):
/lgtm

Issue #50444, opened 2026-06-10
Redundant per-batch validity and pattern-matcher work in expression evaluation


xiaofan-luan:
## Problem

Expression evaluation is batch-driven: `PhyExpr::Eval()` runs once per 8192-row batch (~122 times per 1M-row segment). Several batch-level code paths redo work whose result is invariant across batches, and several per-row loops do work that can be skipped or widened:

1. **`ScalarIndex::IsNotNull()` rebuilt per batch.** Each call allocates a segment-sized bitmap, fills it, ANDs it with the stored null bitset and returns it by value (tracer span + ~500KB memory traffic for 1M rows). The ByOffsets batch paths (`ProcessIndexLookupByOffsets`, `ProcessIndexChunksByOffsets`, `ProcessChunksForValidByOffsets`) call it once per batch, rebuilding the same bitmap ~120× per expression — measured 3.32ms wasted per expression per 1M-row segment. The ChunksAll paths already cache it (`cached_index_chunk_valid_res_`); the ByOffsets paths missed the convention.

2. **Per-row random bitmap reads when the column has no nulls.** Even when the IsNotNull bitmap is all-ones (non-nullable fields — the common case), the loops read one random bit per row instead of bulk-setting the output.

3. **`ApplyValidData` processes one row at a time with a data-dependent branch**, instead of packing 64 bools into a word and skipping all-valid blocks.

4. **`OffsetMapping` bitset scans are byte-wise with early exit** — such loops do not auto-vectorize, scanning 1 byte/iteration.

5. **Chained bitset proxy assignment** `res[i] = valid_res[i] = v` forces a store-to-load read-back of the word just written.

Related: the per-batch `LikePatternMatcher` reconstruction is tracked separately in #50445.

## Measured impact (Xeon Gold 6338, microbenchmarks)

| fix | scenario | result |
|---|---|---|
| cache IsNotNull + all-valid flag | 1M rows × 122 batches | 3.32ms → 0.03ms (**98.6x**) |
| ApplyValidData 64-row blocks | no nulls / 50% nulls | **1.9x** / 2.14x |
| OffsetMapping word-wise scan | 1M-bit full scan | **5.64x** |

Correctness validated: full core unittest run (2064/2064 passed) with the fixes applied.

## Proposed fix

A single PR covering the five items above, following existing caching conventions (`cached_*` members on the expression instance, `EnsureRegexCache`-style lazy init).

czs007:
[...]

PR #52576, opened 2026-08-17
enhance: [cp 2.6] cache scalar-index IsNotNull() across batches on the ByOffsets paths

bigsheeper:
Port of master PR #50447 to the 2.6 branch.

pr: #50447
issue: #50444

## Why this is a port, not a cherry-pick

`git cherry-pick e2887dfe99` conflicts on 2.6. The conflicts are at the **anchor** level, not the content level: the upstream hunks are positioned against the `ExprExecPath` / `SliceCachedResult` / `cached_result_` refactor, none of which exists on 2.6 (`apply_field_valid_data`, `SliceCachedResult`, `cached_result_`, `cached_valid_result_`, `EnsureExecPathDetermined`, `ExprExecPath` — all 0 occurrences on 2.6, all present on master).

The code the PR *adds* is self-contained and does not depend on that refactor, so it is reproduced here line-for-line.

## What it does

`ScalarIndex::IsNotNull()` rebuilds a segment-sized bitmap on every call (allocation + fill). The three ByOffsets paths — `ProcessIndexChunksByOffsets`, `ProcessIndexLookupByOffsets`, `ProcessChunksForValidByOffsets` — called it once per 8192-row batch, while the ChunksAll paths already cached it. All three assert `num_index_chunk_ == 1`, so the cached answer cannot change within a query.

Adds `GetCachedIndexValidBitmap()` (expression-instance cache + precomputed all-valid flag), plus the two accompanying optimizations from the upstream PR: all-valid fast paths, and materializing the bool once instead of chaining proxy assignments.

## Equivalence with #50447

| Check | Result |
| --- | --- |
| Added / removed lines vs #50447 | Identical line-for-line after stripping indentation |
| Indentation delta | Only at `ProcessChunksForValidByOffsets`, which sits one nesting level shallower on 2.6 |
| Diff stat | `1 file changed, 43 insertions(+), 10 deletions(-)` — same as #50447 |
| `clang-format` (repo `.clang-format`) | No changes |
| Bitset API used (`all()`, `set()`, `set(start, size)`) | Present on 2.6 `BitsetBase` with identical signatures |

## Verification status

- [x] Diff verified line-for-line against #50447
- [x] `clang-format` clean
- [x] API availability on 2.6 verified by reading `bitset.h`
- [ ] **Not compiled locally** — this machine only has a master build tree, and 2.6 pulls `bsoncxx` via `common/bson_view.h`, which is not provisioned here. Relying on CI for the C++ build.
- [ ] **Unit tests not re-run locally.** Upstream #50447 reported `Expr.*:ExprTest.*:*Null*:OffsetMapping.*` passing (2047 tests); not repeated here.
- [ ] `make static-check` not run locally.

## Context

Found while investigating a production instance on v2.6.18 where `InvertedIndexTantivy::IsNotNull()` accounted for **45.9% of querynode CPU** (plus ~15% in the accompanying `memset`), against **1.3%** for the actual vector search, on a collection with 19 nullable fields and 16 scalar indexes.

Note that this PR addresses only the per-batch recomputation. On 2.6 the `is_nested_index_` early-return is also absent, so ARRAY-typed JSON-path indexes still run the full per-bit loop; that is a separate change.

### DIFF
--- a/internal/core/src/exec/expression/Expr.h
+++ b/internal/core/src/exec/expression/Expr.h
@@ -383,6 +383,20 @@ class SegmentExpr : public Expr {
         ApplyValidMask(validity, res, valid_res, size);
     }
 
+    // The IsNotNull() virtual rebuilds a segment-sized bitmap on every
+    // call (allocation + fill + AND); per-batch callers must reuse one
+    // copy. The all-valid flag short-circuits per-row bitmap reads.
+    template <typename Index>
+    const TargetBitmap&
+    GetCachedIndexValidBitmap(Index* index_ptr) {
+        if (!cached_index_valid_res_) {
+            cached_index_valid_res_ =
+                std::make_shared<TargetBitmap>(index_ptr->IsNotNull());
+            cached_index_all_valid_ = cached_index_valid_res_->all();
+        }
+        return *cached_index_valid_res_;
+    }
+
     int64_t
     GetNextBatchSize() {
         if (use_json_stats_) {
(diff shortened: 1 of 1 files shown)


czs007 (approved):
/lgtm

PR #52578, opened 2026-08-17
enhance: [cp3.0] replace field binlog array reads with stats in metrics and quota paths

tedxu:
pr: #52388

Cherry-pick of #52388 (master) to 3.0.

V3 segments may have empty Binlogs arrays after DataCoord
restart (KVs are no longer persisted per #50410). Replace
the remaining FieldBinlog iterations with Statistics
accessors so insert/delta file counts and format labels
report correct values for V3.

- Add formats field to Statistics proto: unique set of
  column-group storage formats collected from insert
  binlogs during writes and compaction
- segmentMetricFormatLabel: read Stats.Formats instead of
  iterating Binlogs
- reloadFromKV: use Stats.InsertBinlogCount and
  Stats.DeltaBinlogCount for flushed-segment metrics
- getBinlogFileCount: same substitution for quota metrics
- GetPersistentSegmentInfo: document that V3 returns empty
  insert logs by design

Signed-off-by: Ted Xu <ted.xu@zilliz.com>


congqixia (approved):
/lgtm

Issue #49846, opened 2026-05-15
[Enhancement]: Optional authentication for the management-plane endpoints on port 9091


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Port 9091 hosts the cluster-operations endpoints (`/management/*`) and the Go profiling endpoints (`/debug/pprof/*`). Access control for these operator-facing endpoints is today only indirectly tied to the data-plane `common.security.authorizationEnabled` setting and is not uniform across all of them.

I'd like to add an independent, opt-in authentication option for the management plane:

1. A new `common.security.adminAuthEnabled` switch. When enabled, the `/management/*` and `/debug/pprof/*` endpoints on port 9091 require HTTP Basic Auth with the Milvus root credential. This is decoupled from the data-plane `authorizationEnabled` so operators can secure the admin surface independently of user/data authorization policy.
2. Explicit, audited marking of configuration keys that carry credentials or infrastructure topology, so they are consistently redacted by `/management/config/get` and protected from modification through `/management/config/alter` — replacing the current best-effort substring-based redaction with an explicit, CI-enforced list.

The new switch defaults to **off**, preserving today's behavior. Enabling it is recommended for production deployments via release-notes guidance.

### Why is this needed?

- Operator-plane access control should be configurable independently of the data-plane `authorizationEnabled` switch.
- Configuration keys that expose credentials or infrastructure topology deserve explicit, audited marking rather than a best-effort substring filter that can miss keys.
- The profiling endpoints should be access-controlled under the same model as the rest of the management plane.

### Anything else?

- Raised following an external security review by [lemi9090](https://github.com/lemi9090), whose report is gratefully acknowledged. Coordinated handling of the underlying details is being done separately.
- Implementation is in progress; a PR will follow and reference this issue.

xiaofan-luan:
keep it.

PR #52579, opened 2026-08-17
enhance: make configuration projections safe by default

liliu-z:
issue: #49846

Design: [Safe Configuration Projections](https://github.com/liliu-z/milvus/blob/enhance/secure-config-projections/docs/design-docs/design_docs/20260826-safe-configuration-projections.md)

Configuration diagnostics and management endpoints previously shared raw manager views with runtime consumers. Those views can expose credentials, connection topology, process-environment entries, and deletion tombstones. This change adds declared scalar/group metadata and safe projections while preserving the management configuration API's existing write contract.

## Behavior

- Safe configuration views omit undeclared entries and mask sensitive values. Management point reads return `*****` for sensitive keys and reject undeclared keys. Registration and sensitivity decisions account for aliases and dynamic group namespaces.
- `ParamItem.Sensitivity` uses `Auto` (zero value), `Sensitive`, and `NonSensitive`; explicit states override inference. Sensitivity controls projections and logging only. Generic management SET/DELETE retains the existing immutable, `mq.type`, and `common.security.adminAuthEnabled` restrictions, original request formats, caller-key handling, duplicate validation, and etcd transaction behavior. Sensitive, unregistered, and security-related settings remain writable unless an existing restriction applies. A successful write does not make its raw value visible through GET.
- Built-in file/etcd publication and safe reads share a snapshot lock, keeping source values and their spelling classification together. Collisions and first-time spelling changes retain ambiguity across deferred ownership events and removal. Initial local file spellings are established before environment overrides. Raw manager getters retain their previous locking behavior and source priority.
- Effective group reads omit tombstones; deleting or resetting an overlay through an alias addresses equivalent stored overlay spellings.
- Sensitivity metadata is installed before scalar/group declarations become visible to concurrent refreshes, including fallback keys and the hook namespace. Positive inventories check actual declarations for reviewed credentials, connection targets, transport, authentication, trust, and routing settings, including the Pulsar tenant and namespace consumed by topic construction. Registration classification stays internal; audit tests use public reads and projections.
- WAL configuration maps, management mutations, and dynamic config events omit request names and values from logs. Configuration parser, remote-source TLS initialization, and opaque plugin/callback diagnostics avoid echoing sensitive input. Immutable startup writes mask composite keys containing the protected root path.
- Native plugin initialization reports the failed stage without the configured path, plugin-provided name, or opaque exception text. Sanitization precedes native exception observers and Go startup logs while retaining typed codes, allocation-failure classification, and untyped-exception counting. Go plugin wrappers preserve returned errors while suppressing their diagnostic text.
- Kafka diagnostic strings mask credentials. `FieldBroadcastID` uses the native `uint64` type and preserves its full range through RPC metadata.

Projections remain uncached O(N) diagnostic/management operations. Management authentication, authorization, and additional write restrictions are separate work; endpoint authentication is supplied by #52580, now included through master. This PR does not encrypt stored configuration or audit application-wide backend diagnostics and logs emitted inside out-of-tree plugins.

### Merge with current master

Validated merge commit `24a8c123fd` with master `2e0677c1d2`. All Go tests used `-tags dynamic,test -gcflags="all=-N -l" -count=1`.

- Full `pkg/config` passed with an isolated embedded etcd server.
- Focused ParamTable sensitivity, declaration ordering, projection, and lifecycle tests passed.
- Config race tests passed for all normalized key spellings, cache entry/byte/concurrency bounds, owned request strings, sensitivity aliases, `knowhere` identity, and refresh consistency.
- The actual config GET/ALTER handlers, extracted into an isolated adapter, passed retained coordinator regressions against real embedded etcd. Coverage includes ordinary and sensitive writes, redacted GET, admin-auth SET/DELETE rejection under five spellings, no partial writes in rejected batches, `mq.type` aliases, transaction failures, and omission of payload names/values from logs.
- Full `internal/http` and `internal/util/adminauth` passed. `go vet` passed for config, ParamTable, and the extracted handler adapter; `git diff --check` passed.

The real coordinator package could not compile locally because RocksDB and `milvus_core` native dependencies are unavailable. The extracted-handler checks do not replace native coordinator CI.

### Feature validation before the master merge

Prior local validation used `-tags dynamic,test -gcflags="all=-N -l" -count=1`:

- Full `pkg/config` suite passed with an isolated embedded etcd server.
- Focused ParamTable tests passed for the sensitivity enum, scalar/fallback overrides, Pulsar resource selectors under dotted/slash/uppercase-underscore/collapsed spellings, raw runtime values, projections, callback logs, declaration audits, and credential/immutability rules. The Pulsar point-read regression failed on the prior head and passed after the fix.
- Focused ParamTable `-race` tests passed for enum behavior, Pulsar resource visibility, and sensitivity registration before declaration visibility.
- The retained coordinator regressions passed in an isolated adapter that extracts the actual configuration HTTP handlers and connects them to real embedded etcd. The same write-behavior checks passed against the pre-PR write handler. Coverage includes sensitive/unregistered/security settings, Pulsar tenant/namespace SET and DELETE across four key spellings, single and batch requests, empty values, omitted/null resets, raw storage values, redacted GET responses, original key spellings, immutable/WAL rejection, duplicate error behavior, and transaction atomicity. Revised-handler logs omit request canaries on successful writes and transaction failure.
- `go vet -tags dynamic,test` passed for config, ParamTable, and the extracted current-handler adapter; `git diff --check` passed.

Full ParamTable testing was attempted but stopped because existing tests require creating `/var/lib/milvus/data`, which this host cannot write. Full coordinator tests were attempted but could not compile on this host because RocksDB and `milvus_core` native libraries are unavailable. The extracted-handler checks do not replace the full coordinator/native CI run. Broader backend RPC error descriptions, third-party network-client diagnostics, and logs emitted inside external plugins remain outside the verified log-sanitization coverage.


chyezh:
Reviewed commit `7762a0a7ae7633e38314417c5e42a9b253fcbea4`. One compatibility issue and three configuration-modeling suggestions:

**R1 [P2]: Preserve group overrides when a key is also declared as a scalar.**

[`ParamGroup.GetValue()`](https://github.com/milvus-io/milvus/blob/7762a0a7ae7633e38314417c5e42a9b253fcbea4/pkg/util/paramtable/param_item.go#L573-L578) now uses `GetEffectiveBy`, whose [`overlayIsAuthoritative`](https://github.com/milvus-io/milvus/blob/7762a0a7ae7633e38314417c5e42a9b253fcbea4/pkg/config/manager_projection.go#L42-L56) check gives scalar registration precedence over group membership. This drops dotted overlays even when a group previously consumed them:

```go
mgr := config.NewManager()
item := paramtable.ParamItem{Key: "custom.limit", DefaultValue: "10"}
item.Init(mgr)
group := paramtable.ParamGroup{KeyPrefix: "custom."}
group.Init(mgr)
mgr.SetMapConfig("custom.limit", "20") // The write path used by SaveGroup.
values := group.GetValue()
```

With no underlying source, the previous implementation returns `{"limit":"20"}`; the new implementation returns an empty map. The scalar returns `"10"` in both versions. A spelling being invisible to the scalar does not make it unused by the group.

This is derived from the code, not an executed reproduction. Kafka has a scalar/group overlap, but its final producer configuration explicitly applies the scalar value, so this does not establish a production Kafka failure.

Please keep runtime group reads independent of external projection filtering and preserve valid group overrides. A regression test should cover overlapping scalar/group declarations, with and without a file source, through Save/SaveGroup and Remove/Reset, asserting both readers' results.

**Configuration modeling suggestions:**

1. Replace the mutually exclusive `Sensitive` and `NonSensitive` booleans with one enum: **Auto, Sensitive, NonSensitive**. Auto can be the zero value and retain the existing inference rules, without allowing contradictory flags.
2. Sensitivity should control **presentation only**, including configuration output and log redaction. It should not alter runtime values or determine whether SET/DELETE is permitted. Changing sensitivity should change visibility, not mutability or authorization.
3. Model mutation through **Immutable and interface RBAC**: Immutable means the configuration cannot be changed; RBAC determines which identities may modify it. An authorized caller should be able to modify a non-immutable sensitive setting while its displayed value remains redacted. Authorization should be enforced by the interface's permission model, rather than using sensitivity as a substitute.

czs007:

This PR routes configuration reads through registered, redacted projections. Most of the change is sound, but a few edge cases break the documented management API contract or remove useful diagnostics.

## Medium

**`pkg/config/manager.go:254`: a sensitive key that was never set returns `*****` instead of a key-not-found error**
- **Problem:** `GetRegisteredConfig` checks `isSensitiveResolved` before `readResolved` looks up the value. A declared sensitive key with no value never reaches the `ErrKeyNotFound` return at `manager.go:309`. `HandleGetConfig` (`restful_mgr_routes.go:1433-1434`) then maps `ErrKeySensitive` to `Value: "*****"` with no error. The same applies to `kafka.saslPassword`, which is declared sensitive but commented out in the default `configs/milvus.yaml:318`, and to any unset member of a sensitive `ParamGroup` such as `credential.*`. This contradicts §8 of `20260826-safe-configuration-projections.md`: "undeclared and missing keys keep error entries". Before this PR, these keys returned a key-not-found error.
- **Failure:** `GET /management/config/get?keys=kafka.saslPassword` on a deployment where no file, env, or etcd source sets it → the response has `value: "*****"` and no error. The client treats the key as set, and a read-then-restore flow writes the literal `*****` back as the Kafka SASL password.
- **Suggestion:** In `GetRegisteredConfig`, call `readResolved` first. Return its `ErrKeyNotFound` unchanged, and return `ErrKeySensitive` only when a value exists. Add a test for a key that is both sensitive and missing. The current cases in `restful_mgr_routes_test.go:504-535` only cover sensitive keys that are set.

(raised by tinswzy, bigsheeper)

## Low

**`pkg/config/manager.go:399`: after `Remove` then `Save`, the tombstone under the dotted spelling survives and management GET reports key not found**
- **Problem:** `DeleteConfig` writes `TombValue` under every spelling from `overlayKeys`: `formatKey(key)` and `mapConfigKey(key)`. `SetConfig` (`manager.go:341-348`) only overwrites `formatKey(key)`. For `ParamGroup` members, `readResolved` checks the dotted overlay first (`manager.go:282`), finds the leftover tombstone and returns `ErrKeyNotFound`. `Manager.GetConfig` returns the live value. Today only tests reach this, through `BaseTable.Remove` / `ComponentParam.Remove`.
- **Failure:** calling `paramtable.Get().Remove("<group prefix>.member")` and then `Save("<group prefix>.member", "v")` → `GET /management/config/get` reports key not found, while `GetConfig` and the runtime use `"v"`.
- **Suggestion:** In `SetConfig`, clear or overwrite every `overlayKeys` spelling, not only `formatKey(key)`. Alternatively, make `DeleteConfig` tombstone only the spellings that `SetConfig` writes. Add a Remove→Save→GET regression test for a group member.

(raised by tinswzy, bigsheeper)

**`pkg/util/paramtable/component_param.go:287`: `GetAll` on `ComponentParam` and `hookConfig` now returns redacted values**
- **Problem:** `ComponentParam.GetAll` and `hookConfig.GetAll` (`hook_config.go:55-56`) switched from `mgr.GetConfigs()` to `mgr.ProjectConfigs()`. Both are exported from the `pkg/v3` module. They now return `*****` for sensitive values and leave out undeclared keys. Because the hook `SoConfig` group is now declared `KeyPrefix: "", Sensitive: true`, `hookConfig.GetAll` masks every `hook.yaml` value.
  - The only in-tree caller is `internal/proxy/impl.go:6602`, and that HTTP route is meant to return redacted values.
  - `cipher.go:459` uses `EtcdCfg.GetAll` and `cipherConfig.GetAll`, which still return raw values.
  - Out-of-tree code that expects raw values from these two methods would silently get `*****` instead.
- **Suggestion:** Add a doc comment on both `GetAll` methods saying they now return redacted values, and point to `Manager.GetConfigs` for raw access. Also list the change under the compatibility notes in §8 of the design doc.

(raised by tinswzy, bigsheeper)

**`internal/util/hookutil/hook.go:129`: `InitOnceHook` masks the cause of a hook startup failure but still logs `so_path` in clear text**
- **Problem:** Every error from `initHook` is now logged as `config.RedactedValue`. Most of these come from `LoadPlugin` (`plugin.go:31,36,41`), and those errors contain only the plugin path and symbol name. The same log line already prints `so_path` in clear text, so masking them hides nothing. Only the `hookVal.Init` error at `hook.go:84` can echo `hook.yaml` values. The cipher path (`cipher.go:450-453`) masks the paths too, so the hook path is inconsistent with it.
- **Failure:** `proxy.soPath` points at an incompatible `.so` (a `plugin.Open` version mismatch or a missing `MilvusHook` symbol), with `common.panicWhenPluginFail=true` → the proxy panics at startup, and the only log line is `fail to init hook so_path=/x.so error=*****`, so the `plugin.Open` / `Lookup` cause is lost.
- **Suggestion:** Mask only the error from `hookVal.Init`, for example with a marker wrapper at `hook.go:84`. Keep `LoadPlugin` errors as they are, or at minimum log `merr.Code(err)` and which stage failed.

(raised by tinswzy)

**`pkg/util/paramtable/param_item.go:216`: error text from config-change callbacks is fully masked, which makes debugging harder**
- **Problem:** `handleConfigChange` masks old and new values for every `EtcdSource` event. The "param change callback failed" log (lines 226-230) masks the error unconditionally. Before this PR it logged raw values and `mlog.Err(err)`. Causes such as a `ParseFloat` failure no longer reach the logs, even for public scalars changed by a file refresh. The inline comment explains the reason: a callback may read other sensitive settings even when its own key is public (for example, cipher rotation reloading KMS credentials).
- **Suggestion:** Keep the payload masked, but also log `merr.Code(err)` or the error type. Operators then get the kind of failure without the sensitive text.

(raised by tinswzy)

czs007:

**Verified:**
- `internal/proxy/impl.go:6597` — `ClusterConfigsPath` and `HookConfigsPath` now pass `GetConfigsView`/`GetAll` as method values, so each request re-projects instead of serving a boot-time map.
- `pkg/config/manager.go:258` — `GetRegisteredConfig` returns `ErrKeyNotFound` before the sensitivity check, so a declared-but-unset sensitive key is reported as missing rather than as `*****` (resolves the earlier medium at `manager.go:254`; the "missing sensitive keys return errors" test pins it).
- `pkg/config/manager_policy.go:118` — `groupMemberIsEnvironmentOnly` and the collapsed-prefix branch of `resolveRegisteredKey` refuse identities backed only by `environmentSourceName` under all four spellings, so the `secret_enable` vs `secret.enable` leaf-rename path stays masked.
- `pkg/config/source.go:43` — `publishSourceSnapshot`, `FileSource.update` and `EtcdSource.update` take `snapshotMutex` then the source lock; readers take `snapshotMutex.RLock` then source `RLock`; `fireEvents` and cache eviction run outside both, so no AB-BA or re-entrant lock was found.
- `pkg/util/paramtable/base_table.go:166` — adding the file source before `EnvSource` is precedence-safe because `pullSourceConfigs` resolves collisions by source priority (`NormalPriority` 11 over `LowPriority` 21), not by add order.
- `internal/core/src/storage/storage_c.cpp:328` — the explicit `bad_alloc` → `SegcoreError` → `std::exception` → `...` catch chain keeps the original error codes while replacing only the message text; `test_plugin_loader.cpp` asserts the path canary is absent from both stdout and `error_msg`.

The previously raised tombstone issue at `pkg/config/manager.go:399` is also resolved: `SetConfig` now removes tombstones stored under other spellings only when the stored value is a tombstone, covered by `TestBaseTable_RemoveThenSaveGroupMember`.

## Low — non-blocking

- `pkg/config/manager_projection.go:46` — A key that is both a declared `ParamItem` and a member of a registered `ParamGroup` prefix resolves as `RegisteredConfigScalar` before the group-prefix scan runs (`manager_policy.go:377`), so `overlayIsAuthoritative` requires the separator-free spelling. A dotted overlay written by `SaveGroup` → `SetMapConfig` (stored under the dotted spelling at `manager.go:379`) never satisfies that, and `walkProjection` skips it (`manager_projection.go:179`). Because `ParamGroup.GetValue` now reads `GetEffectiveBy` instead of `GetBy` (`param_item.go:577`), the group reader drops the written member while `Manager.GetBy` still returns it. Out-of-tree or test code declares a key both as a `ParamItem` and under a `ParamGroup` prefix with separators, then calls `SaveGroup` → `ParamGroup.GetValue` silently omits the member (or reports the stale file-source value if a file holds the dotted key) and `ProjectBy`/`GetRegisteredConfig` return nothing, so `SaveGroup` becomes a silent no-op for the group consumer. This is not reachable by in-tree production code: the only in-tree scalar under a group prefix is `knowhere.enable`, whose spellings coincide under `formatKey`, and `SaveGroup` has no non-test caller. This is the same point @chyezh raised as R1 [P2] on 2026-09-14 with no author reply so far; the logic is unchanged since the reviewed commit. Suggestion: in `resolveRegisteredKey`, run the group-prefix scan before (or alongside) the `declaredKeys` lookup so a key under a registered prefix is classified as a group member, or have `overlayIsAuthoritative` accept the dotted spelling for scalars whose key also falls under a registered prefix; add a test with a declared scalar under a separator-bearing group prefix that writes via `SaveGroup` and reads via `ParamGroup.GetValue`. (raised by tinswzy)

chyezh:
Follow-up review of `ae0e812529274725b133d282e2427b76c0f654ec`, with three requested changes:

**1. R2 [P2]: Include Pulsar tenant and namespace in the sensitive configuration inventory.**

[`pulsar.tenant` and `pulsar.namespace`](https://github.com/milvus-io/milvus/blob/ae0e812529274725b133d282e2427b76c0f654ec/pkg/util/paramtable/service_param.go#L1441-L1458) are registered scalars, but neither has sensitivity metadata nor matches the fallback name rules. Consequently, `GetRegisteredConfig`, `ProjectConfigs`, and `GetConfigsView` still expose their values.

These values determine the actual remote resource: the [Pulsar WAL builder](https://github.com/milvus-io/milvus/blob/ae0e812529274725b133d282e2427b76c0f654ec/pkg/streaming/walimpls/impls/pulsar/builder.go#L66-L69) passes them to the opener, which constructs `tenant/namespace/topic`. They fall within this PR's stated topology/routing protection scope, alongside the already protected broker address and MinIO bucket/region. This is a remaining coverage gap in the new projections, not a newly introduced authorization bypass.

Please mark both sensitive and add them to the positive audit inventory. Test point reads and projections across supported key spellings, while asserting that internal consumers still receive the original values and that SET/DELETE behavior remains unchanged. Deriving the inventory from the actual connection/resource constructors would help catch omissions that the keyword heuristic cannot identify.

**2. Update stale comments and reduce unnecessary public API surface.**

The removal of sensitivity-based write restrictions should also be reflected in these comments:

- [`function_param.go`](https://github.com/milvus-io/milvus/blob/ae0e812529274725b133d282e2427b76c0f654ec/pkg/util/paramtable/function_param.go#L228-L234) still says marking the group sensitive prevents unauthenticated configuration mutations from redirecting traffic or weakening TLS. Sensitivity now only controls presentation; it does not provide that protection.
- [`grpc_param.go`](https://github.com/milvus-io/milvus/blob/ae0e812529274725b133d282e2427b76c0f654ec/pkg/util/paramtable/grpc_param.go#L147-L152) still implies an unregistered namespace cannot be configured through management endpoints. Registration restricts external reads, while writes retain their existing contract.

Also, [`ResolveRegisteredConfigKey`](https://github.com/milvus-io/milvus/blob/ae0e812529274725b133d282e2427b76c0f654ec/pkg/config/manager.go#L234-L241) has no in-tree production callers after the mutation policy was removed. Please reconsider exposing this resolver and its classification type solely for audit tests; prefer testing the public read behavior where possible, so internal classification details do not become an unnecessary compatibility commitment.

**3. Replace the mutually exclusive sensitivity flags with one three-state enum.**

[`ParamItem`](https://github.com/milvus-io/milvus/blob/ae0e812529274725b133d282e2427b76c0f654ec/pkg/util/paramtable/param_item.go#L115-L142) still uses `Sensitive` and `NonSensitive` booleans plus a panic for contradictory flags. Please use a single enum field with **Auto, Sensitive, and NonSensitive** states. Auto should be the zero value and retain the current inference behavior; the other two states should explicitly override that inference. This makes contradictory states unrepresentable and keeps declaration intent clear.

The enum should affect configuration presentation and log redaction only. Keep mutation decisions separate, using the existing Immutable constraint and the interface authorization model.

These observations are based on static code review; I did not execute the regression tests.

PR #52580, opened 2026-08-17
enhance: add optional root authentication for the management plane

liliu-z:
issue: #49846

design doc: docs/design-docs/design_docs/20260912-management-plane-authentication.md

Merge dependency: #52579 supplies safe configuration projections. Both changes are needed for the full configuration-redaction requirement in #49846.

## Problem and behavior

Port 9091 exposes process shutdown, coordinator management, profiling, event logs, the web console, and legacy REST data operations. Add `common.security.adminAuthEnabled`, default **false**, to authenticate this surface independently of the main service port's data-plane authorization.

When enabled:

| Surface | Policy |
| --- | --- |
| `/management/*` except readiness; `/debug/pprof/*`; `/log/level`; `/eventlog`; `/webui/`; `/telemetry`; `/api/v1/_*` | Root HTTP Basic authentication |
| Legacy REST data operations under `/api/v1` | Existing valid-user/API-key authentication if `authorizationEnabled=true`; otherwise root Basic authentication |
| `/healthz`, `/livez`, `/metrics`, `/metrics_default`, `/management/check/ready` | Open |
| `/api/v1/health` | Exempt from the new gate; existing data-plane authentication still applies |

After the request-origin checks below, missing/incorrect credentials return 401; non-root users are rejected before password lookup with 403; unavailable credential storage and saturated verification return 503. Dependency details and stack traces remain out of unauthenticated response bodies. HTTP and Gin share one decision and carry a verified root principal through a typed request context for in-process telemetry calls.

## Browser and client requirements

The gate rejects requests identified as cross-site or same-site. An unknown Fetch Metadata value and `Origin: null` are rejected as well. Top-level GET document navigation is allowed only on document surfaces, not operator actions or console APIs.

**Credential-bearing requests without Origin or Fetch Metadata must include `X-Milvus-Admin-Request: true`.** This is required because missing metadata does not establish that a request is from a non-browser client; ordinary remote HTTP is outside the Fetch Metadata trustworthy-URL guarantee. An explicit header does not override a reported cross-site origin. The metrics-port CORS preflight must not grant this header to cross-origin callers.

Browser console access requires HTTPS with Fetch Metadata preserved by the reverse proxy. A document request without sufficient request context is refused without a Basic challenge. Non-browser callers add the explicit header, for example:

```sh
curl --user root --header 'X-Milvus-Admin-Request: true' \
  'https://milvus-admin.example/management/querycoord/balance/status'
```

The flag remains opt-in. Existing clients with the flag disabled keep their prior behavior.

## Credential lifecycle and resource bounds

Verifier ownership is deterministic in standalone: Proxy, then MixCoord, then worker. Workers create a MixCoord client lazily. Each lookup caller has its own five-second wait bound, so a service-discovery call that ignores the backend context cannot block the HTTP request indefinitely. Its accounting slot remains occupied until the shared backend result completes, preserving the resource bound even in that case. Root hashes are cached for 10 seconds, reused for up to 10 minutes after their last successful fetch if refresh fails, and lookup failures are cached for 2 seconds. The stale window preserves management access during a coordinator outage but can also retain acceptance of a password rotated during that outage.

Concurrent hash refreshes share work and admit at most 32 callers, including canceled callers whose singleflight result channels remain retained. Worker client construction uses a single shared completion signal; repeated lookup timeout windows no longer append a retained channel to a stuck constructor. Construction failure can be retried, and a client completed after shutdown is closed instead of published.

bcrypt concurrency is bounded to a quarter of GOMAXPROCS, floored at two, with a bounded queue. This is a concurrency cap, not an attempt rate limit; on small pods sustained incorrect passwords can still consume substantial CPU.

## Registration, configuration, and event logs

- `http.Register` refuses an unmarked operator route, including method- and host-qualified ServeMux patterns. Assembled-router tests and a console-prefix test cover Gin registration.
- Milvus-owned handlers use a private mux. Flag-off compatibility with more-specific legacy DefaultServeMux routes is retained when pprof is enabled; enabling the gate removes that fallback.
- `/management/config/alter` always rejects this flag, using the same `config.FormatKey` identity as config storage, so an anonymous caller cannot plant a disabling etcd value before the operator enables the gate.
- Key-normalization caching admits at most 4096 entries and only keys whose input and normalized forms fit 1024 bytes, using owned string storage so short substrings cannot pin large request buffers. Larger keys still normalize correctly without retained memo entries. This bounds cache retention, not individual HTTP body size.
- The eventlog gRPC stream switches to loopback with authentication enabled. A failed security upgrade closes the wildcard listener. Event-log shutdown uses a separate close signal and stops the gRPC server, preserving safe concurrent recording.
- `milvus_admin_auth_total{endpoint,result}` records root-gate and origin decisions using registered route patterns; failure logs are rate limited.
- The standard Go runner now includes HTTP, adminauth, coordinator, eventlog, and metrics regressions. It restores fail-fast behavior after sourcing the environment so a failed test command returns a failure from the runner.

## Compatibility and deployment

Port 9091 remains plaintext. Use a trusted network or TLS-terminating proxy, rotate root using `UpdateCredential` / `db.update_password`, and update graceful-shutdown and profiling clients before enabling. Editing `defaultRootPassword` does not rotate an existing credential. Official compose deployments and embedded standalone scripts publish 9091, so review that exposure. Browser access needs the HTTPS deployment described above.

Regardless of the flag, downstream operator routes without `AdminAuth` now fail registration; the admin-auth key cannot be altered through the management API; configuration normalization is bounded; pprof registration is explicit; and the data-plane stored-password comparison shares the credential-format helper.

## Verification and limits

CI observed on head `c8992fc7b4d3fcf9a28bf73ad4700283922367f2` at **2026-09-12 14:30 UTC**: `ci-passed` and `dco-passed` are present.

| Check | Result |
| --- | --- |
| [Build / unit / integration / code checks](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-build-ut-ciloop-pipeline/30883/pipeline-overview/) | SUCCESS; status reports build, code checks, integration, Go, and 9,106 C++ tests passed |
| [Go SDK E2E](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-go-sdk-pipeline/15021/pipeline-overview/) | SUCCESS |
| [Default E2E](https://jenkins-milvus-ci.milvus.io/job/MILVUS-CI-V2-PR-PIPELINES/job/milvus-e2e-pipeline-dispatcher/19796/pipeline-overview/) | SUCCESS |
| Main CI trigger / DCO | SUCCESS |

The non-gating Tencent Pulsar gray run reports `ERROR` because its downstream job could not be triggered. The previous PR head had the identical start failure. No Pulsar test result is claimed, and the precise trigger failure is unknown because Jenkins logs are inaccessible here. The TC-US shadow build and legacy Jenkins status still show pending; they are outside the master `ci-passed` requirements. No label was added manually.

Local verification and remaining limits:

All Go tests use `-tags dynamic,test -gcflags="all=-N -l"`.

- Full `internal/util/adminauth`, including race detection, passed. A blocked-constructor regression retained 20,011 extra heap objects after 10,000 canceled waits before the fix and no retained growth after it. Cancellation, retry, panic, and a client completing after close are covered. Four blocked-backend scenarios cover cold cache, expired stale state, and correct/incorrect stale passwords; race runs repeated three times passed, including retained slots and recovery after release.
- Full `internal/http` and its race run passed with the explicit-request-header changes, including 34 new net/http/Gin cases. `pkg/eventlog`, `pkg/metrics`, and `pkg/util/merr` also passed.
- Configuration-normalization tests passed, including oversized inputs, Unicode expansion, and request-buffer ownership. The full config race suite passed excluding `TestEtcdSource`, whose two tests require a reachable external etcd and failed against this host's deliberately unavailable test endpoint. The config comment and ParamItem documentation match exactly.
- The newly scheduled HTTP, adminauth, eventlog, and metrics commands passed with the runner's race and coverage flags. A dispatch harness checked all five package additions through the full and selected-group entry points. Ten injected command failures were swallowed before restoring fail-fast and now stop the runner immediately with the original nonzero status. Coordinator execution was unavailable locally; the subsequent native CI aggregate reports Go success.
- Legacy Basic challenges are preserved on the main port, gate-off metrics routes, and the health exemption. A regression using the actual extracted parser with real Gin failed before the repair and passed after it; assembled-route and dynamic-gate compatibility tests were added to the native package suites. The subsequent CI Go check reports success; individual native test logs were not accessible here.
- `internal/proxy`, `internal/distributed/proxy`, `internal/coordinator`, and `cmd/roles` cannot compile on this host because `milvus_core.pc`, `milvus-storage.pc`, and `rocksdb.pc` are unavailable. The subsequent native CI aggregate reports build and Go success; this does not change the local compilation limitation.
- Full `make test-go` was attempted; native dependency construction failed in the host SDK linker (`library 'System' not found`) before the Go suite. This is not recorded as a passing full test run.
- golangci-lint v2.11.3 built with Go 1.26.6 reported zero new issues in `internal/http`, `internal/util/adminauth`, and the changed config/eventlog/metrics/paramtable packages. Native-linked packages could not be linted locally; the subsequent CI code-check status reports success.
- A real coordinator failover and a real browser console flow remain unverified. Unit and source-level failure tracing do not establish full browser credential replay or transport behavior.


czs007:

**Verified:**
- `internal/http/server.go:721` — the `Register` panic guard is complete: every `Register`/`management.Register` call site under a gated prefix now sets `AdminAuth: true`, and `/metrics` / `/metrics_default` don't match `mustBeGated`, so no operator route panics at boot.
- `pkg/eventlog/grpc.go:83-111` — `Stop()` closes `closed` instead of `ch`, and `Notify` selects on `closed` first, so a post-`Stop` `Record→Notify` can no longer send on a closed channel.
- `internal/http/server.go:2412-2431` — the 401/403/503 separation holds end-to-end (a verifier's own 403 doesn't collapse into 503; a malformed stored hash and an unreachable coord both render 503, not 401), and response bodies never carry the wrapped cause.
- `internal/http/server.go:2379` — the fixed-priority verifier slot array (Proxy < Coordinator < Worker) makes standalone resolution deterministic regardless of registration order, falling through nil slots.
- `internal/distributed/proxy/service.go:346-379` — all four `{gate, legacy}` combinations for open / console / data-plane paths reproduce the pre-PR `apiv1.Use(authenticate)` behaviour when the gate is off, so upgrades are transparent.

---

## Low — non-blocking

**`internal/http/auth.go:75` — cross-site rejection falls open when neither `Sec-Fetch-Site` nor `Origin` is present, and mutating `/management/*` handlers never check the HTTP method.**
`RejectCrossSite` falls back to `Origin` when `Sec-Fetch-Site` is absent and returns `false` (allow) when both are missing. That fallback is deliberate and disclosed in the commit message ("non-browser clients, which send neither header, are unaffected"), so this is a known residual risk rather than a docs/implementation mismatch. What makes it reachable is the other half: the `/management/stop` handler reads only `req.URL.Query().Get("role")` and never inspects `req.Method`, and the `WWW-Authenticate` challenge added by this PR (on `/webui/` and `/telemetry`, `AuthChallenge: true`) is exactly the mechanism that makes a browser cache root Basic credentials for that origin and attach them to any same-origin path.
Failure: with `adminAuthEnabled=true` and an operator who has entered the root password at the `/webui/` Basic Auth prompt, any client that sends neither header (an older browser, or a non-browser client) loading `:9091/management/stop?role=querynode">` → `auth.go:75` allows it, the cached root credentials are attached automatically, and `server.go:411-424` runs `triggerComponentStop("querynode")`.
Suggestion: either reject GET on mutating management routes (the handler ignoring `req.Method` is worth fixing independently of CSRF), or require `Sec-Fetch-Site` — or a custom header — on `AdminAuth` routes. Rated low because the port was entirely unauthenticated before this change: the same cross-site GET succeeds on master with no credentials at all, so this PR strictly narrows the exposure. *(raised by bigsheeper)*

**`internal/proxy/proxy.go:356` — during standalone shutdown, `/management/*` requests fall through to the Worker verifier and block on a self-directed gRPC call until `fetchTimeout`.**
`MilvusRoles.Run` registers the Worker credential verifier and its deferred unregistration at `cmd/roles/roles.go:482-489`, earlier than the `proxy.Stop()` call site, so under LIFO the Worker slot is still populated after `Proxy.Stop`'s defer has cleared `VerifierSlotProxy` and `mixCoordImpl.Stop` has cleared `VerifierSlotCoordinator`. In that window `resolveManagementVerifier` (`server.go:121-131`) resolves to the Worker verifier, whose `CachedRootVerifier` has never been used, so it refreshes → `fetchRootHash` → lazy `mix.NewClient` dial of a MixCoord that has already stopped inside this same process.
Failure: standalone with `adminAuthEnabled=true`, a `/management/*` request carrying root Basic credentials arriving after `proxy.Stop()` returns but before `Run`'s defer clears the Worker slot → the request retries until `fetchTimeout` (5s) before returning 503, holding the HTTP handler for that whole window (subsequent requests hit the 2s `failureTTL` negative cache and 503 immediately).
Suggestion: clear `VerifierSlotWorker` at the start of the stop sequence (or gate the lazy `mix.NewClient` dial on a shutdown signal) so teardown-window requests fail fast instead of waiting on a dead in-process coordinator. Bounded and teardown-only. *(raised by xiaocai2333)*

**`internal/util/adminauth/cache.go:79` — the concurrent-bcrypt floor of 2 lets unauthenticated callers saturate CPU on small nodes, and there is no per-client attempt limiting.**
`slots := max(2, runtime.GOMAXPROCS(0)/4)` bottoms out at 2, so on a 2 vCPU pod two concurrent bcrypt comparisons (~60ms each) are essentially the whole CPU. The limiter caps CPU occupancy but does not rate-limit attempts, and wrong passwords have no negative cache — the `verifiedSha` short-circuit applies only to *correct* passwords, so replaying the same wrong password costs a full bcrypt every time. The code comment at `cache.go:71-77` and the PR's "Known trade-offs" already state this explicitly, so the only actionable residue is deployment defaults: `deployments/docker/standalone/docker-compose.yml:59` and `scripts/standalone_embed.sh:55` both publish 9091 by default, while the documentation tells operators to keep that port behind a trusted network or TLS proxy.
Failure: a 2 vCPU Milvus pod with `adminAuthEnabled=true` and reachable 9091, an attacker sending sustained concurrent `Authorization: Basic base64(root:<random>)` to `/management/stop` (published on every node) → every request reaches the real bcrypt via `compareBounded`, the two slots consume the pod's entire CPU and squeeze the data plane; excess requests are shed with 503 but the CPU is already spent.
Suggestion: either adjust those two default port publications in this PR, or call the two files out by name in the release note and record the rest as follow-up. Rated low: the flag is off by default and the worst case with it on is no worse than current master, where 9091 has no authentication at all. *(raised by xiaocai2333)*

## Nitpick — non-blocking

**`internal/distributed/proxy/service.go:229` — `authenticateAndCount` records every `authenticate` abort as `AdminAuthUnauthenticated`.**
The outcome is derived from `c.IsAborted()`. Today `authenticate` (`service.go:257-281`) has exactly one abort path, `c.AbortWithStatusJSON(http.StatusUnauthorized, ...)`, so the metric is accurate as written and there is no current failure scenario. The note is about coupling: correctness depends on that implementation detail, and a future non-401 abort branch would silently mis-label the counter. Suggestion: derive the outcome from `c.Writer.Status()` instead of `IsAborted()`. *(raised by xiaocai2333)*

PR #50495, opened 2026-06-12
feat: add the shard routing abstraction for online shard split

xiaocai2333:
issue: #50463
design doc: #50465

> **Blocked on milvus-io/milvus-proto#618.** The `schemapb` routing types this
> package reads (`CollectionShardInfo`, `ShardState`, `HashRouting`, and
> `routing_modulus` / `shard_by` on `DescribeCollectionResponse`) live only in
> that PR. Until it merges and the `go.mod` pin is bumped, `ci-v2/build-ut-cov`
> fails at typecheck with `undefined: schemapb.CollectionShardInfo`. Verified
> locally against #618 `c929178`: build, vet, `-race` and lint clean, 99.6%
> statement coverage — the one uncovered arm is `HashPrimaryKeys`'s `default:`, which
> guards an `IDs` variant a future build adds and cannot be constructed from outside milvus-proto.

> **Scope change.** This PR previously carried the whole namespace shard-split feature on one integration branch for end-to-end review, as its description said it would be split before merge. This is that split: the PR is now only the first layer. The previous head is archived at the tag `archive/pr50495-namespace-shard-split` (`576052d21d`) and nothing is lost.

## Why this exists

Where a write goes is decided in one place instead of at each caller. Today the proxy computes `hash(pk) % len(vchannels)` inline. That is correct only while every shard owns an equal, position-derived slice of the key space — exactly what a shard split stops being true. Once a shard is halved, no modulo over a channel list reproduces the placement, and a write routed by position lands on a shard that does not own it, or on a fenced split source that rejects it.

So the rule moves out of the callers and into one package that derives it from the collection's own metadata.

## The model

A collection carries **one routing modulus M**, and each shard carries **the residues modulo M that it owns**. The residue sets partition `[0, M)`, so a route is a single array index:

```
route(value) = slots[value % M]
```

M is not the shard count. A never-split N-shard collection is M = N with one residue per shard, which is the legacy `hash % N` placement bit for bit.

A split halves one shard. When that shard still owns several residues, its set is divided in two and **M does not move** — the common case. Only a shard down to a single residue `r` has nothing left to divide: M doubles and `r` becomes `{r}` and `{r+M}`, which is the same key space cut on one more hash bit. A doubling is collection-wide, so every untouched shard is first carried onto the new modulus (`Rebase`); `PlanSplit` reports when that is needed and the caller applies both halves in the one atomic meta update that commits the split.

[...]

Issue #50463, opened 2026-06-11
[Feature]: Online shard split for namespace collections


xiaocai2333:
**Is there an existing issue for this?**

- [x] I have searched the existing issues

**What would you like to be added?**

Support splitting one shard of a namespace-enabled (multi-tenant) collection into two shards online, without interrupting reads or writes.

The split is metadata-only: every segment belongs to exactly one namespace and the split point always falls on a namespace boundary, so segments are relabeled to their new shard instead of being rewritten.

Key elements:
- Range routing over `big_endian(hash(namespace)) || namespace` with a versioned routing table derived from the collection meta.
- A `SplitShard` WAL fence message (`T_switch`); writes to the fenced vchannel are rejected, and the proxy refreshes routing and re-dispatches to the new shards.
- The source delegator spawns the two child delegators in place and fronts queries (fan-out + reduce) during the transition window; sealed segments stay single-loaded on the source delegator.
- DataCoord redistributes segment metadata (relabel, including namespace-scoped L0 segments) in rounds in the background.
- QueryCoord adopts the new shards in one shot after redistribution completes, converting the in-place child delegators without reload, then releases the source shard.
- Fully gated by `dataCoord.shardSplit.enable` (default off).

Design doc: `docs/design-docs/design_docs/20260610-shard_split.md` (submitted with the first PR).

Tracking:
- [ ] Common contracts: `SplitShard` message type, `SHARD_FENCED`/`ROUTING_STALE` error codes, collection-meta shard routing fields, additional-vchannel allocation
- [ ] Range routing table derived from the collection meta, and routing-version negotiation
- [ ] StreamingNode: split RPC, vchannel fence, target-vchannel initialization with the TSO barrier
- [ ] Proxy: reject-and-refetch write switching
- [ ] QueryNode: in-place child delegators, fronting fan-out/reduce, delete/timetick forwarding
- [ ] DataCoord: split task FSM, trigger, batched relabel, multi-round redistribution, source-shard freeze
- [ ] QueryCoord: splitting freeze, one-shot adoption, in-place handoff, source-shard release
- [ ] Integration tests, chaos coverage, metrics, and rollout gating

**Why is this needed?**

The shard count of a collection is fixed at creation time. As tenant data grows, a single shard hits WAL throughput, delegator memory, and compaction backlog limits simultaneously; the only workaround today is re-importing into a new collection, which is unacceptable for online workloads.

**Anything else?**

[...]

PR #52072, opened 2026-07-31
feat: [RLS1] add row-level security metadata foundation

aoiasd:
relate: #50263
design doc: docs/design-docs/design_docs/20250610-rls_design.md
design doc PR: #53173

## Summary
Adds the collection RLS switch, management APIs, privileges, validation, and persistence.

PR #52583, opened 2026-08-17
feat: add the WAL messages a shard split is built on

xiaocai2333:
issue: #50463
design doc: docs/design-docs/design_docs/20260610-shard_split.md

Step 2 of online shard split. Independent of #50495 (the routing abstraction); builds on plain master and pins the released milvus-proto (`a301a6af92`).

## What this adds

**One WAL message.** A shard split's write switch is a single `SplitShard` message broadcast to the source vchannel(s), the target vchannel(s) and the control channel. Every replica carries the same header and body; a consumer decides what a replica means by where its own vchannel appears:

| replica on | role | what the streamingnode does |
|---|---|---|
| a vchannel in `source_vchannels` | the fence | seals every growing segment as of the message's time tick (ids embedded in the header), force-fails active transactions, marks the vchannel SPLITTED; the time tick is `T_switch`; DML is refused with `SHARD_FENCED` from then on |
| a vchannel in `targets` | the genesis | registers the collection on this pchannel for DML and segment assignment (schema from the body's `genesis`, partitions from the header), seeds the vchannel meta, spawns its data sync service |
| the control channel | nothing | orders the ack callback, as every DDL broadcast's control replica does |

The body also carries `routing`: the post-image of the collection's shard routing (vchannel list, every shard's state and residues, modulus, `shard_by`). It is what rootcoord's ack callback commits once every replica has landed — that callback is #52604's; this PR only defines the message and the streamingnode's handling.

**Two ordering mechanisms replace the barrier.** Every target message must sort after `T_switch`, and the targets live on other pchannels. (1) `BroadcastHeader.append_first_vchannels` names the source vchannels; the broadcaster appends that group first, persists its append results as a partial ack, and only then appends the rest — a restart in between finds the sources already acked and does not re-append them. (2) `SplitShard` carries the new message-type property `FreshTimeTick`: the timetick interceptor discards the node's cached TSO batch and draws the tick from a freshly fetched one. rootcoord carves batches from one monotonic counter, so a batch fetched after the source's append was acknowledged is necessarily past `T_switch`. Both mechanisms are required; neither alone is sufficient.

**Retirement is a routing commit, not a message.** The `AlterCollection` replica whose `shard_split_routing` update no longer lists a vchannel retires it: the shard interceptor, recovery storage and flusher tear it down exactly as `DropVChannel` once did, keyed by vchannel name and guarded so a live (unfenced) shard is never torn down. `CreateVChannel` and `DropVChannel` no longer exist; message type values 50 and 51 are reserved.

**A re-fence by the same task succeeds.** The broadcaster re-drives a task whose source replica landed before the task was persisted, and retries any refused replica forever. So the source replica of the task that placed the fence is appended again and succeeds (the shard manager keeps the first fence; the second record seals nothing); a fence from any other task id is still refused with `SHARD_FENCED`.

## Details worth the reviewer's attention

**`T_switch` can legitimately have two values.** After a re-drive the source's second record takes a new tick `T' > T`. The streamingnode's `VChannelMeta.split_time_tick` keeps `T` (first fence wins); the broadcast result the coordinator records carries `T'`. No DML can exist in `(T, T']` because the shard was fenced at `T`, so every consumer of `T'` (the drain predicate, the "no target message ≤ T_switch" invariant) is conservative, not wrong.

**A refused replica wedges the broadcast.** The broadcaster has no unrecoverable-error escape: a replica the streamingnode refuses is retried with backoff forever while the task holds `ExclusiveCollectionName`. The refusals the split path can return are all coordinator invariant violations — a replica on a vchannel that is neither source nor target, a target genesis without a schema, a target placed on a pchannel the collection already occupies, a fence from a second task on one source, a routing commit that delists an unfenced vchannel. Two of those guards live in the coordinator and are **prerequisites for wiring this message**: `AllocVirtualChannels` must exclude the pchannels the collection already occupies, and the split trigger must keep one active task per vchannel (#53095). Without the first, a target landing on an occupied pchannel leaves the source fenced, no target created, and the collection's DDL lock held until an operator intervenes.

**Every operation on the shard manager's registration is guarded by the vchannel NAME.** The map is keyed by collection id, one entry per collection per pchannel, so after a retired source's slot is reclaimed a later vchannel of the same collection can hold that entry. Reads, the fence and both lifecycle edges match on the name. A retired source keeps a fence tombstone so a stale proxy still gets `SHARD_FENCED` rather than `CollectionNotFound`.

**Target and retire replicas are never forwarded to the data sync service.** The target replica spawns the DSS; the retire replica closes it. The source replica is forwarded, and the dd_node seals the fenced segments and sets a flush timestamp so the L0 delete buffer cannot pin the checkpoint.

**Flusher recovery reads an empty message id as the earliest position the WAL retains** (per backend) instead of panicking; the checkpoint candidate is logged at Warn.

## Error codes

`SHARD_FENCED` (with informational `fenced_time_tick` / `fenced_split_task_id`) and `ROUTING_STALE` are added; both map to `FailedPrecondition` and are unrecoverable for the producer, so the error reaches the proxy, which refetches the routing table. The whole error detail is cloned across the produce response. No coordinator reads the two fields; `ROUTING_STALE` has no producer yet.

## Kept, unused here

`AppendOption{BarrierTimeTick}` / `AllocateWithBarrier` stay: ManualFlush uses them. `AlterCollectionMessageUpdates` fields 10–14 and `FieldMaskCollectionShardSplitRouting` are the routing post-image's shape; their rootcoord consumer is #52604.

## Compatibility

Nothing here has shipped in a release: no released coordinator emits `SplitShard`, so there is no rolling-upgrade path to preserve. A replica whose header carries no `source_vchannels` gets role Unknown and is refused on append and reported as an inconsistency on replay.

## Not wired up yet

Nothing appends this message. The coordinator that drives it follows in #52604 (routing commit from the ack callback) and #53095 (the split task issuing the broadcast and recording `T_switch` and the targets' genesis positions from its result), all behind `dataCoord.shardSplit.enable`.

## Verification

Unit tests per layer, run locally with `-tags dynamic,test`: proto/message package (role helpers, retire predicate, broadcast option, type property), `idalloc`/`ack`/timetick interceptor (fresh batch, one extra rootcoord round trip), broadcaster (group order, partial-group persistence, restart does not re-append, partial success persisted), shard interceptor/manager (all three roles, same-task re-fence without a re-seal, foreign-task refusal, retire guards), recovery storage (seed/fence/retire, exemptions, unknown role), flusher (target/retire not forwarded, source forwarded), streaming client (message construction, validation, result decoding). `mock_ShardManager.go` was hand-edited for the two changed signatures because mockery cannot run against this toolchain; a later `make generate-mockery-internal` reproduces the same lines.

Follow-ups tracked separately: observable handling of a broadcast whose replica is refused forever (metric + escalating log), the pre-existing `NewUnrecoverableError(err.Error())` format-string sites in the shard interceptor.

https://claude.ai/code/session_01E94wAdA7wA5B74zLAEnrrj


czs007:

Adversarial review of the shard-split streaming changes found one cross-process correctness gap in how `FencedTimeTick` travels over the Produce stream, plus a few drain-timing and guard-hardening issues in the flusher and shard interceptor.

## High

- **`internal/streamingnode/client/handler/producer/producer_impl.go:320`** — `FencedTimeTick` is dropped on the Produce streaming RPC response-body path, so T_switch cannot be recovered across processes in cluster mode. The PR fixed the gRPC status-detail path (`rpc_error.go:93-104` now uses `proto.Clone`), but `RawAppend` goes through `wal.go:96` → `AppendMessages` → `producer.NewResumableProducer`, and `handler_client_impl.go:206-208` only short-circuits to the local WAL when the pchannel is hosted in-process. On the remote path the server (`produce_server.go:265`) places the append error in `ProduceMessageResponse_Error` via `AsPBError`, which does carry `FencedTimeTick`, but the client rebuilds it with `status.New(Code, Cause)`, and `New` (`streaming_error.go:245-251`) only populates `Code`/`Cause`. `producer.go:155-156` returns the error as-is for unrecoverable codes, and `splitshard.go:112-116` then reads `FencedTimeTick == 0`. Failure: cluster deployment where the coordinator loses T_switch and re-sends `SplitShard` against a source vchannel already in SPLITTED → streamingnode returns `NewShardFenced(vchannel, GetSplitTimeTick)` with a populated pb, but the client yields `SwitchTimeTick: 0` alongside `ErrSourceVChannelFenced`, contradicting the documented "the caller recovers T_switch even after a crash"; a drain gate of `checkpoint ≥ T_switch` (design doc §6.3) degenerates to always-true, and fence-sealed segments can be declared drained before they are reported. `TestShardFencedSurvivesTheWire` only exercises the unary status round-trip, and standalone hands back the `*StreamingError` itself, so neither exposes this. Also note `splitshard.go:22` says "no exact T_switch is needed", which contradicts `:86-90`, `:113-115`, and the PR description. Suggestion: at `producer_impl.go:320`, `proto.Clone` the whole pb error and convert it to `*StreamingError` (or add a `status.NewFromPBError` helper), add a round-trip test where `ProduceMessageResponse_Error` carries `FencedTimeTick`, and reconcile the comment at `splitshard.go:22`. (raised by bigsheeper)

## Medium

- **`internal/streamingnode/server/flusher/flusherimpl/msg_handler_impl.go:141`** — `HandleSplitShard` does not set a flush timestamp, so the L0 delete buffer is never proactively synced and the source shard's checkpoint can lag behind T_switch for up to `syncPeriod`. `HandleSplitShard` only calls `wbMgr.SealSegments`; `sealSegments` (`write_buffer.go:881-908`) moves the L1 segments named in the header from Growing to Sealed, which the Sealed policy then syncs promptly. Deletes, however, are buffered under an L0 segment id (`l0_write_buffer.go:52-56`) registered as Growing (`:148-156`) and absent from `FlushedSegmentIds`, so they are only picked up by the flush-ts policy (`sync_policy.go:80-98`, requires a flushTs), the stale policy (`syncPeriod`, default 600s at `component_param.go:7765-7771`), or the full policy. `GetCheckpoint` (`write_buffer.go:571-607`) takes the earliest position across all buffers, so the unsynced L0 buffer pins the checkpoint before the fence. Failure: source shard had deletes before the fence (L0 buffer non-empty, not full) and no new DML afterwards → L1 segments sync quickly but the L0 delete buffer waits for the stale policy, and a coordinator gating drain on `checkpoint ≥ T_switch` per §6.3 stalls roughly ten minutes per split; no data is lost. Suggestion: mirror `HandleManualFlush` (`:111-120`) and call `FlushChannel(ctx, msg.VChannel(), msg.TimeTick())` after `SealSegments`; the source DSS still receives pchannel-level timeticks after the fence (`flusher_components.go:139-141`), so the flush-ts policy will push the L0 buffer out once the checkpoint reaches T_switch. (raised by czs007, bigsheeper)

## Low

- **`internal/distributed/streaming/splitshard.go:319`** — The streamingnode side of `DropVChannel` only checks that the vchannel is SPLITTED, not that it has drained, and the DSS discards its buffer via `Close(drop=false)`. The chain is `wal_flusher.go:332` `WhenDropCollection` → `flusher_components.go:125-127` `ds.Close()` → `data_service_wrapper.go:73-77` `RemoveChannel` → `manager.go:315-323` `buf.Close(ctx, false)` → `write_buffer.go:1611-1623`, which returns without creating a sync task when `!drop`. Meanwhile recovery marks the vchannel DROPPED and `dropAllVirtualChannel` (`recovery_background_task.go:156-162`) drives datacoord `DropVirtualChannel` (`services.go:830-839`), which marks leftover segments Dropped and the channel checkpoint dropped. The only append-side gate, `CheckIfVChannelCanBeDropped`, requires `State == SPLITTED` and nothing else. Failure: caller appends `DropSplitVChannel` while the source write buffer still holds fence-sealed data (`checkpoint < T_switch`) → the message is accepted, the buffer is discarded, datacoord marks the residual segments Dropped, and writes before T_switch are silently lost. No such caller exists in this PR; whether it triggers depends on the coordinator PR's ordering. Suggestion: document `channelCheckpoint(source) ≥ T_switch` as a hard precondition in the `DropSplitVChannel` comment (`:305-319`, which currently only says it must be appended before the vchannel is removed from the collection), and consider enforcing a drain check on the streamingnode side. (raised by czs007, bigsheeper)

- **`internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go:235`** — `handleInsert`/`handleDelete` let writes with a mismatched vchannel name through, attaching them to a segment of a different vchannel in the same collection. Both call sites (`:235`, `:332`) only test `errors.Is(err, ErrVChannelFenced)`; `checkIfVChannelCanBeWritten` (`shard_manager_collection.go:140-148`) returns `ErrCollectionNotFound` when `info.VChannel != vchannel`, which is ignored, and execution falls through to collection-keyed `CheckIfCollectionSchemaVersionMatch` (`:427-432`) and `AssignSegment` keyed by (collection, partition) (`shard_manager_segment.go:131-139`). This PR is the first to allow two vchannels of one collection to occupy the same pchannel in sequence, since `checkIfVChannelCanBeCreated` returns nil when no collection entry exists. Failure: source vchannel X of collection C has been dropped, a later split creates C's vchannel Y on the same pchannel, and a proxy with a stale route sends insert/delete to X → the interceptor accepts it, the message lands with vchannel=X but a segment allocated by Y's partition manager, `flusher_components.go:142-144` finds no DSS for X and returns nil, `recovery_storage_impl.go:464-472` counts the rows against Y's segment, and the client sees success while the batch is silently lost. Suggestion: reject any non-nil error from `CheckIfVChannelCanBeWritten` (fenced → `SHARD_FENCED`, anything else → Unrecoverable). (raised by czs007)

- **`internal/streamingnode/server/flusher/flusherimpl/flusher_components.go:64`** — Within a restart window, a target vchannel without a DSS is treated as dropped and skipped. On recovery the flusher calls datacoord `GetChannelRecoveryInfo` for each vchannel in the snapshot (`wal_flusher.go:157-162` → `util.go:57-72`); `services.go:1156-1159` returns `ErrChannelNotAvailable` when the seek position is nil (no channel checkpoint, no segments, no entry in `StartPositions`, per `handler.go:476/481/489`), and `util.go:70-72` maps that to `errChannelLifetimeUnrecoverable` and skips the channel. When WAL replay later reaches `CreateVChannel`, `flusher_components.go:64-73` skips it again because `genesis.TimeTick() <= recoveryCheckPointTimeTick`. Failure: `InitSplitTargetVChannels` succeeded and the recovery snapshot persisted (checkpoint ≥ target genesis), but streamingnode restarts before datacoord has a checkpoint, segment, or start position for the target → the target never gets a DSS, and DML written to it is neither flushed nor reported to datacoord until the next restart after datacoord has its seek position. `CreateCollection` has the same-shaped window, but rootcoord persists the start position right after the DDL ack, keeping it very short; for split targets the persistence order is decided by the not-yet-merged coordinator PR. Suggestion: either persist the target start position before the recovery snapshot can include the target, or make the flusher treat a snapshot vchannel with no datacoord recovery info as "create DSS from genesis" rather than "dropped". (provenance unclear — please confirm whether this PR introduces it) (raised by czs007, bigsheeper)

xiaocai2333:
Thanks — all four verified against the code and fixed in `0e6bd00eee`, `2b2d068a7f`, `04df29dd12`, `270190c4b5`. Notes on each, and an answer to the provenance question on the last one.

## High — `FencedTimeTick` dropped on the produce response path (must-fix)

**Confirmed, and the reasoning was right down to the last step.** `producer_impl.go` rebuilt the error with `status.New(Code, Cause)`, which populates nothing else, so on the remote path `SplitShard` returned `SwitchTimeTick: 0` alongside `ErrSourceVChannelFenced` and the drain gate `channelCheckpoint(source) >= T_switch` degenerated to always-true.

Fixed by adding `status.NewFromPBError`, which clones the whole pb message, and routing **both** conversion paths through it — `TryIntoStreamingError` now calls it too, so there is one place to get this wrong instead of two.

Two tests: `TestNewFromPBErrorCarriesTheWholeMessage` (including that it clones rather than aliases, since the transport may reuse the response), and `TestShardFencedPayloadSurvivesTheProduceResponse`, which drives a real `ProduceMessageResponse_Error` through a producer. I checked it actually catches the bug — reverting the one-line change fails it with `expected: 0x6371b12c36a0001, actual: 0x0`.

The `splitshard.go:22` contradiction is reconciled: "no exact T_switch is needed" is true of the **barrier** (freshly allocated, only ever a lower bound) and false of the **drain** (gates on its value). The comment now says which.

## Medium — no flush timestamp at the fence

**Confirmed.** `HandleSplitShard` sealed only the L1 segments the header names; the L0 delete buffer is registered growing and never appears in `FlushedSegmentIds`, and `GetCheckpoint` takes the earliest position across all buffers — so it pinned the source checkpoint before the fence until the stale policy fired. Fixed exactly as suggested, mirroring `HandleManualFlush` / `HandleFlushAll` / `HandleTruncateCollection`. The handler test now asserts the flush timestamp is the fence's own time tick, and that a failing `FlushChannel` fails the handler.

## Low — `DropSplitVChannel` drain precondition

**Confirmed**, and documented rather than enforced. The precondition is now stated on the function: the caller must have `channelCheckpoint(source) >= T_switch`, because the teardown closes the data sync service with `drop=false` and discards the buffer unsynced. I did not add a streamingnode-side drain check: the streamingnode does not have the coordinator's `T_switch` for a vchannel it did not fence in this process lifetime, so the check would be either redundant with the split task's own gate or unable to run. The split task is where the gate belongs and where it already is.

## Low — insert/delete accept a vchannel the shard manager does not hold

**Confirmed, and this one is worse than Low in my reading** — it is silent data loss reported as success, and it is reachable exactly the way you describe (stale route to a retired source after its successor takes the pchannel).

The half-wiring is what makes it easy to miss: the `info.VChannel != vchannel` guard was added to `checkIfVChannelCanBeWritten` in this branch, but two of its three call sites tested only `errors.Is(err, ErrVChannelFenced)` and dropped the rest. `handleSplitShardMessage` already handled any non-nil error. Both write paths now do too — fenced stays `SHARD_FENCED` so the client refreshes its route, anything else is unrecoverable, because refreshing does not make that vchannel writable on this pchannel. Two tests added.

## Low — target vchannel without a DSS inside a restart window (provenance)

**Not introduced by this PR** — this is the `CreateCollection` shape, and the coordinator PR closes the window the same way rootcoord does.

`advanceHashFencing` calls `seedTargetCheckpoints(genesis)` **immediately** after `InitSplitTargetVChannels` returns and **before** the routing commit, so the target's channel checkpoint is persisted right after the genesis append — the same ordering as rootcoord persisting the start position right after the DDL ack. The window is therefore the same shape and the same width as `CreateCollection`'s: it exists between the genesis append landing and the checkpoint seed landing, and a streamingnode restart inside it hits the behaviour you describe.

So I have not changed the flusher here. If the residual window is worth closing it should be closed for both genesis messages at once, in the flusher — "a snapshot vchannel with no datacoord recovery info means create the DSS from genesis" — rather than for split targets alone. Happy to do that as a follow-up if you think the `CreateCollection` window is also worth closing; it is not a regression this PR introduces, and narrowing it only for split targets would leave the more common path exposed.

---

Full regression on the touched subsystems is green: 57 packages across `internal/streamingnode/...`, `internal/distributed/streaming/...`, `internal/util/streamingutil/...` and `internal/flushcommon/...`, no failures.

czs007:

**Verified:**
- `internal/streamingnode/client/handler/producer/producer_impl.go:326` — the produce-response body error now goes through `status.NewFromPBError`, so `FencedTimeTick` reaches the drain gate instead of being discarded by a `(code, cause)` rebuild.
- `internal/util/streamingutil/status/streaming_error.go:45-56` — `NewFromPBError` clones the whole protobuf before the cast, so the payload is not aliased to a transport message that may be reused.
- `internal/streamingnode/server/service/handler/producer/produce_server.go:265` — the server sets the body error via `AsStreamingError(err).AsPBError()`, and `NewShardFenced` populates `FencedTimeTick` before the cast, so the value is actually on the wire.
- `msg_handler_impl.go:154-163` — `FlushChannel` is called with the fence's own time tick after `SealSegments`, which lets `GetFlushTsPolicy` push out the L0 delete buffer that `SealSegments` cannot reach; pchannel-level ticks are still broadcast to the fenced source's data sync service, so the trigger arrives.
- `write_buffer.go:400-406` — `SetFlushTimestamp` bumps only `processedTs`, which `GetCheckpoint` does not read, so the flush at the fence cannot advance the reported source checkpoint past unsynced data.
- `shard_interceptor.go:235-254` and `:347-366` — rejecting any non-nil result from `CheckIfVChannelCanBeWritten` is behaviour-preserving for the pre-existing collection-not-found cases and newly rejects only the vchannel-mismatch case that previously caused silent loss.

## Low — non-blocking

- `internal/streamingnode/client/handler/producer/producer_impl.go:327` — `status.NewFromPBError(nil)` returns a nil `*StreamingError`, and the next line calls `IsRateLimitRejected()` on it, which dereferences `e.Code` through a pointer receiver and would panic. There is no real trigger today: the server only sets the body error inside the `err != nil` branch, `AsStreamingError` is never nil for a non-nil error, and a proto3 oneof message field is always allocated non-nil by the decoder. Suggestion: add a nil guard after `NewFromPBError` (or have the helper return a non-nil unknown-code error for a nil input) so the client does not depend on the server never sending an empty error body. (provenance unclear — please confirm whether this PR introduces it; the pre-existing `produceResp.Error.Code` access panicked on nil as well) (raised by tinswzy)
- `internal/distributed/streaming/splitshard.go:325` — the code comment on `DropSplitVChannel` is accurate, but the PR-discussion rationale for not adding a streamingnode-side drain check ("the streamingnode does not have the coordinator's `T_switch` for a vchannel it did not fence in this process lifetime") contradicts what this PR implements. `SplitTimeTick` is persisted into the vchannel recovery meta (`vchannel_recovery_info.go:241`, proto field `split_time_tick`), restored into the shard manager on restart (`shard_manager.go:203`), and read by `GetSplitTimeTick` for the idempotent re-fence at `shard_interceptor.go:390`, so the streamingnode does hold `T_switch` across restarts. What the shard manager lacks at append time is the channel checkpoint: `GetCheckpoint` lives in the write buffer on the flusher side, and neither the shard manager nor the shard interceptor has a path to it, so `CheckIfVChannelCanBeDropped` can only verify `State == SPLITTED`. Suggestion: restate the reason in the discussion (or in the comment at `splitshard.go:325-333`) as "the shard manager cannot observe the channel checkpoint at append time", so the upcoming coordinator PR review does not inherit the incorrect "streamingnode does not hold `T_switch`" premise. (raised by tinswzy)

xiaocai2333:
## Review — reviewed at head `270190c4b5`, cross-checked against the design doc in #53098

Read the full diff plus the surrounding unchanged code, the streaming-system guides, and `docs/design-docs/design_docs/20260610-shard_split.md` as it stands in #53098. The design is solid and the implementation follows it closely; what follows is one CI blocker, four places where the code and its own design doc disagree, and a set of smaller items.

### What the PR does

Adds the three WAL messages a shard split is built on — `SplitShard` (the source vchannel's permanent write fence at `T_switch`, which also auto-seals the growing segments and carries their ids as the only seal record), `CreateVChannel` (a target vchannel's genesis, floored at a `BarrierTimeTick > T_switch`) and `DropVChannel` (retirement of a fenced source) — and wires each into the four places a vchannel lifecycle is tracked: the shard interceptor, the shard manager, the recovery storage and the flusher. Nothing appends them yet. The notable hardening is that "one vchannel per collection per pchannel" stops being an assumption: every shard-manager operation now matches on the vchannel *name*, creation of a conflicting vchannel is refused at append time, teardown of an unfenced shard is refused, and a recovery-time collision is resolved deterministically instead of by map iteration order.

---

## Blocking

### B1 — `tests/go_client` was not bumped with the proto pin, so Code Check fails

**`tests/go_client/go.mod:8`** (and its `go.sum`)

`754af6c8` moved root, `pkg` and `client` to `milvus-proto v3.0.0-20260901101918-a301a6af926b`; `tests/go_client` still requires `v3.0.0-20260806081414-16b288837fbd`. That module `replace`s both `client/v3` and `pkg/v3` with the local directories, so MVS selects the newer proto while `tests/go_client/go.sum` carries no hash for it.

Jenkins build #28915 confirms it: `Start check go_client e2e package` → `context loading failed: no go files to analyze` → `make: *** [Makefile:236: static-check] Error 5`. Code-Check FAILURE, and **UT-GO and UT-Integration were skipped as a result** — the Go unit tests have not run on this PR.

Fix: `cd tests/go_client && go get github.com/milvus-io/milvus-proto/go-api/v3@v3.0.0-20260901101918-a301a6af926b && go mod tidy`, commit both files. The PR is also `CONFLICTING` against master and needs a rebase.

---

## Code vs. its own design doc

### D1 — High: a retired source vchannel no longer answers `SHARD_FENCED`, which breaks the proxy's reject-refetch loop

**`internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go:235-255`** (delete at `:347-367`), **`shards/shard_manager_collection.go:140-149`**

`DropVChannel` removes the collection's entry, so a later write to that vchannel reaches `checkIfVChannelCanBeWritten`'s `!ok` branch, gets `ErrCollectionNotFound`, and is returned as `NewUnrecoverableError`. Design §3.3 requires `SHARD_FENCED` when the source vchannel is `Splitting` **or `Dropped`** — that rejection is the only signal the proxy acts on: invalidate the cached collection meta, refetch, re-resolve, retry. An unrecoverable error instead surfaces to the user a write that a single routing refresh would have completed.

The same branch also covers the case where the entry exists but names a **different** vchannel of the collection — i.e. the successor has taken the pchannel slot. That case is unambiguously the split scenario and is exactly what the commit message of `04df29dd12` describes. That commit argued "refreshing does not make this vchannel writable here", which is true of *this* vchannel but not of the write: in the design's model the refresh is supposed to send it to the successor.

Fix: split the two cases. Entry present but naming another vchannel → `status.NewShardFenced(msg.VChannel(), 0)`. Entry absent entirely → keep unrecoverable, so a genuinely wrong route does not spin in a refresh loop. Covering the fully-retired source as well needs a small tombstone in the shard manager (fenced vchannel name + `T_switch`), which would also repair the §10.1 crash-recovery path, where a re-sent `SplitShard` is supposed to return `SHARD_FENCED` carrying `T_switch` back.

### D2 — Medium: the StreamingNode-side replication check §8.6 asks for is not implemented

**`shard_interceptor.go:103` / `:161` / `:381`** — none of the three handlers consult the WAL role or the replicate configuration.

Design §8.6: "Clusters with replication/CDC enabled reject split (checked at the DataCoord trigger **and again at the StreamingNode**)", because replicated transactions never expire and the secondary maps pchannels by index position. This PR is the StreamingNode-side landing of all three messages and adds no such gate.

The gap is real and not only theoretical. `internal/cdc/replication/replicatestream/replicate_stream_client_impl.go:172` forwards everything that is not `SelfControlled` or `Unreplicable`, and none of the three messages is marked either. On the receiving side, `internal/distributed/streaming/replicate_service.go:175` rewrites in-body channel names for `CreateCollection` only — and `CreateVChannel` reuses that exact body, with `VirtualChannelNames`/`PhysicalChannelNames` filled in by `InitSplitTargetVChannels`. `SplitShardMessageHeader.targets[].vchannel` and `CreateVChannelMessageHeader.split_source_vchannels` are source-cluster names too.

Fix: either reject the three message types in the handlers when the WAL is in a replicating topology (`status.NewReplicateViolation`), or mark them `WithUnreplicable()` for now with a TODO. Either satisfies §8.6's second line of defence; the DataCoord-side trigger check alone does not.

### D3 — Medium: the child start position — doc says `LastConfirmedMessageID`, code uses `MessageID`

**`internal/distributed/streaming/splitshard.go:282-292`** vs. design §6.1 step 3 and the §11 StreamingNode row

Both doc locations say the append's `LastConfirmedMessageID` is what DataCoord persists as the target's start position. `splitTargetGenesisPosition` deliberately uses `result.MessageID`, with a documented reason: a zero last-confirmed id serializes to nil under WoodPecker and downstream position assertions panic, while the message id is just as complete here because the vchannel is created *by* this message.

I think the code is right — the reason is concrete and names a real backend. Worth noting that the two kinds of position are now genuinely different in the implementation: the flusher's genesis DSS still seeks from `LastConfirmedMessageID` (`flusher_components.go:96`), the split's child start position does not. Whichever way this is resolved, the doc and the code should stop disagreeing.

### D4 — Medium: `DropSplitVChannel`'s stated precondition is one of the design's three conjuncts

**`internal/distributed/streaming/splitshard.go:325-332`**

The comment states the precondition as "the source must be DRAINED, i.e. `channelCheckpoint(source) >= T_switch`". Design §6.3 step 2 requires **all three**: no healthy segment left on the source vchannel, `channelCheckpoint(source) >= T_switch`, **and** no active import job listing the source in its `Vchannels`.

The missing import conjunct is the one that matters most, because §6.5 deliberately does *not* stop imports at the fence — the drain check is where that is made safe. A caller implementing the comment as written could retire the vchannel while an import is still writing to it, and the teardown closes the data sync service with `drop=false`, discarding the write buffer without syncing. This comment exists precisely so the next caller does not have to rediscover the ordering; stating a third of it is worse than stating none.

Fix: quote all three conjuncts and cite §6.3 step 2.

### D5 — Medium: `SHARD_FENCED` carries `T_switch` but not the fencing `split_task_id`, so §6.6's exclusion cannot be enforced by the caller

**`pkg/proto/streaming.proto`** (`StreamingError` gained only `fenced_time_tick`), **`shards/shard_manager.go`** (`CollectionInfo` stores `State` + `SplitTimeTick` only), **`splitshard.go:19-28`**

§6.6 relies on "a fence appended to a source that another task already fenced is rejected, which is what enforces the exclusion between a rehash and an automatic split". But `SplitShard()`'s contract treats `SHARD_FENCED` as **success** — `ErrSourceVChannelFenced`, documented as "The caller treats this as success — the fence holds — and rolls forward". The message header carries `split_task_id`, but the shard manager neither persists it nor returns it on the error, so a caller cannot tell its own retry from somebody else's fence. A rehash task that lands on an automatic split's fence would roll forward, create its targets and commit routing — the exclusion §6.6 depends on does not hold at this layer.

Fix (this PR owns the wire surface, so it is cheap here): record the fencing `split_task_id` alongside `SplitTimeTick` in `CollectionInfo`/`VChannelMeta`, return it on the error, and change `SplitShard`'s contract to "roll forward only when the task id matches; a different id is another task's fence and must abort". If that is deferred, §6.6 should say explicitly that the exclusion is enforced by the coordinator comparing against its own persisted task id, not by the rejection itself.

### D6 — Medium: `CreateVChannel` lacks the schema guard `CreateCollection` has, and does not share the schema parser the doc says it does

**`shard_interceptor.go:103`**, compare `:76-79`

`handleCreateCollection` rejects a body with neither `CollectionSchema` nor `Schema` bytes, and resolves both shapes through `messageutil.MustGetSchemaFromCreateCollectionMessageBody`. `handleCreateVChannel` does neither: `shardManagerImpl.CreateVChannel` and `allocFunctionRunners` read `MustBody().GetCollectionSchema()` directly, while the recovery storage uses the `MustGet...` helper. Design §6.1 step 3 says the body keeps the CreateCollection shape precisely "so the three handlers share the existing schema parser" — two of the three do.

Consequence of a `CreateVChannel` without `CollectionSchema`: the shard manager registers a nil schema (every versioned insert then fails `ErrCollectionSchemaNotFound`) while the recovery storage seeds an empty non-nil one, so behaviour differs before and after a restart. The only in-repo caller is guarded by `InitSplitTargetVChannelsParam.Validate`, but the interceptor is the enforcement point against any coordinator version.

Fix: add the same nil check returning `NewUnrecoverableError`, and route both the shard-manager registration and `allocFunctionRunners` through `messageutil.MustGetSchemaFromCreateCollectionMessageBody`.

### D7 — Medium: a restart between the `CreateVChannel` append and the coordinator's registration leaves the target with no flusher

**`internal/streamingnode/server/flusher/flusherimpl/util.go:38-41`** and **`flusher_components.go:64`**

On restart, `buildFlusherComponents` takes the vchannel list from the recovery snapshot (which includes the freshly created target) and asks DataCoord for each one's recovery info. For a vchannel DataCoord does not yet know, `services.go:1156-1159` returns `ErrChannelNotAvailable`, which the flusher classifies as `errChannelLifetimeUnrecoverable` and **skips**, logging "channel has been dropped" at Info. If the genesis message's time tick is at or below the recovery checkpoint, `spawnGenesisDataSyncService` skips it too.

The result is a vchannel the shard manager accepts writes for (its registration recovered fine) whose messages the flusher silently drops, because `flusherComponents.HandleMessage` returns nil when `dataServices[vchannel]` is absent — the exact "appended, acked, never flushed" shape the new insert-path comment warns about. Not reachable in this PR since nothing appends these messages, but the contract is set here.

Fix: the coordinator PR must register the target with DataCoord before, or atomically with, the `CreateVChannel` append — `splitTargetGenesisPosition` already computes the position for it. Independently, a vchannel that is present in the recovery snapshot but rejected by DataCoord deserves an Error-level log or a metric rather than sharing an Info line with genuinely dropped channels.

---

## Smaller items

- **`pkg/mq/msgdispatcher/dispatcher.go:95`** — `SeekablePosition` is exported "so those callers test the same condition", but the PR adds no such caller, and the one site that would actually panic on an empty MsgID (`flusherimpl/util.go:48`, `MustGetMessageIDFromMQWrapperIDBytesWithWALName`) is still unguarded. Either guard that site here or export it in the PR that introduces the callers.
- **`internal/flushcommon/pipeline/flow_graph_dd_node.go:303`** — a failing `HandleSplitShard` is only warned. That matches the neighbouring Flush/ManualFlush branches, so it is not a deviation, but here a failed `SealSegments` also skips the `FlushChannel` below it, the source checkpoint never passes `T_switch`, and the drain gate stalls forever with a single WARN as evidence. A distinct log or metric on this branch would pay for itself.
- **`internal/distributed/streaming/splitshard.go:80-83`** — "The source vchannel holds only messages <= T_switch" is not literally true, and design §7 repeats it: the fence-sealed segments' `Flush` messages, the time ticks, and `DropVChannel` itself all follow, and §6.3's drain gate *depends* on them advancing the checkpoint. Suggest "accepts no DML after `T_switch`; only Flush / TimeTick / DropVChannel follow."
- **`shard_interceptor.go:235` + `:256`** (delete at `:347` + `:368`) — two separate `RLock` acquisitions per insert/delete where one would do. Correctness is fine (the exclusive lock keeps the fence from flipping between them); a combined `CheckWritableAndSchemaVersion` would save a lock round trip on the hot path.
- **`shard_interceptor.go:137`** — `allocFunctionRunners` runs unconditionally after a successful append, including when `shardManager.CreateVChannel` skipped the registration on `ErrVChannelConflict`. `Close()` releases by registered vchannel, so that key is never released. Either have `CreateVChannel` report whether it registered, or release explicitly on the conflict branch.
- **`internal/distributed/streaming/splitshard.go:48-76`** — `SplitShardParam.Validate` checks every field except the targets' routing: a nil `Routing`, or a residue >= `RoutingModulus`, passes. The message is a permanent record; suggest adding both checks via `merr.WrapErrParameterInvalidMsg`.

---

## Dimensions with nothing to report

- **Concurrency.** The lock interceptor precedes shard in the chain (`walmanager/manager_impl.go:36-43`), so `ExclusiveRequired` covers check → append → apply for all three messages. `handleCreateVChannel`'s check/apply window is re-validated under the write lock, so the worst outcome is a refusal, never corrupt state. `CheckIfVChannelCanBeWritten` using `RLock` matches the other read methods; no lock-order inversion.
- **Resource leaks.** Beyond the conflict-path function runner above: the DSS is closed and removed on `DropVChannel`, recovery info reaches `DROPPED` → `ShouldBeRemoved`, and dropped vchannels are filtered out of the snapshot by `IsActive`, so `resolveVChannelCollision` never has to compare a dropped source against a live successor.
- **API compatibility.** Everything is additive — MessageType 47-49, StreamingCode 18-19, `VChannelState=3` reusing the previously commented-out reservation, `VChannelMeta.split_time_tick=5`, `StreamingError.fenced_time_tick=3`. The `MessageSpecializedType` reverse map is keyed on `{Header, Body}`, so `CreateVChannel` reusing `CreateCollectionRequest` as its body does not collide with `CreateCollectionV1`. Rolling upgrade holds: an old delegator never requests `SplitShard` in its `DeliverFilterMessageType`, and the querynode filter node's `default:` branch filters rather than panics.
- **Test coverage.** Thorough — 16 interceptor cases including the conflict refusal, the live-shard teardown refusal, append failures and the negative assertions for a vchannel the manager does not hold; 5 shard-manager cases including `resolveVChannelCollision`; 7 recovery cases including `split_time_tick` persistence and replay-is-not-an-inconsistency; a producer test that drives a real `ProduceMessageResponse_Error` carrying `FencedTimeTick` across the wire. Gaps: nothing covers D2, D6's nil-schema path, or D7's restart window — and because of B1, UT-GO has not actually run on this PR yet.

---

## Verdict

**Needs work**, on B1 alone; D1 and D2 are worth fixing before merge as well — D1 because it changes write availability inside the split window, D2 because the design names the StreamingNode as the place for that check and this PR is that place. D4 and D6 are small. D5 needs a decision either way rather than being left implicit.

The design itself holds up well under the code: fence-first ordering, the vchannel-name guard as an enforced admission rule rather than a logged warning, the fence doubling as the only seal record, and the explicit genesis position all address real silent-failure modes.

Three items belong to #53098 rather than here, listed for convenience:

- §7's "no loss, no duplication" argument says the fence "rejects in the lock interceptor, which runs before TimeTick allocation", then gives the order as `redo → lock → replicate → timetick → shard`. The rejection is in the **shard** interceptor, i.e. after the time tick is allocated — `handleSplitShardMessage` using `msg.TimeTick()` proves it. The conclusion still holds (the append never ran, so nothing was persisted), but the reason should be "before the backend append", not "before sequencing".
- §10.1 contradicts itself inside one bullet: "DataCoord records `T_switch` (returned on the fence ack) on the task because the drain gates on it" and, a few sentences later, "No `T_switch` value is captured, persisted, or recovered anywhere on the coordinator side." The latter looks like a leftover from the revision where the barrier *was* `T_switch`.
- §3.2's "Provenance is not here" concludes that provenance lives only in the split task and is discarded with it. `CreateVChannelMessageHeader.split_source_vchannels` keeps an immutable copy in the WAL forever (and may name vchannels that have since been reclaimed) — worth one sentence so a reader does not conclude it exists nowhere durable.

czs007:

A comment overview sentence first, then the finding.

The new seekability guard in the flusher recovery path is stricter than the WoodPecker deserializer it protects and turns a single vchannel's empty-MsgID position into a permanent, silent failure of the whole pchannel flusher.

## High

- **`internal/streamingnode/server/flusher/flusherimpl/util.go:63`** — The `SeekablePosition` guard added in `getRecoveryInfos` tests the wrong condition and its failure is not recoverable. (a) `SeekablePosition` only accepts an empty `MsgID` when `position.GetWALName() == WALName_WoodPecker`, but the very next line deserializes with the flusher's actual WAL name (`impl.wal.Get().WALName()`), and for WoodPecker `DeserializeWoodpeckerMsgID(nil)` succeeds and yields `{SegmentId:0, EntryId:0}` (earliest). Datacoord routinely stores positions with an empty `MsgID` and `WALName` left at `Unknown`: `recalculateSegmentPosition` for every compaction output and `UpdateImportSegmentPosition` for imports. When the channel checkpoint is missing, `GetChannelSeekPosition` falls through to `getEarliestSegmentDMLPos` and returns exactly such a position, and `isDroppedChannel` does not catch it because that marker is `Timestamp == MaxUint64`. Before this PR that position was accepted on WoodPecker; it is now rejected. (b) The error is returned outside `getRecoveryInfo`'s `retry.Do`, propagates through `buildFlusherComponents` to `Execute`, and the deferred handler only logs `DPanic` (which does not panic since `mlog` is never built with `Development`) before the goroutine exits. Nothing restarts it: the opener fires `RecoverWALFlusher` and returns, and `wal_adaptor` only calls `Metrics()` / `Close()`. `EnterSlowdownMode(nil)` at `wal_flusher.go:92` is never paired with `EnterRecoveryMode()`, which is only reached on success.
  Failure: streamingnode (re)opens a pchannel on WoodPecker while datacoord has no channel checkpoint for one live vchannel (checkpoint init at `WatchChannels` failed with warn-and-continue, or a split-target vchannel whose checkpoint was never reported) and that vchannel's earliest segment position is compaction- or import-derived → `getRecoveryInfos` returns `ErrServiceInternal`, the flusher goroutine exits for good with one `DPanic` line as the only evidence, no data sync service is built for any collection on the pchannel, nothing flushes, no checkpoint advances, the WAL is never truncated, and because slowdown mode is never exited the flusher rate-limit source decays from 16mb to 2mb and, after the 2m reject delay, enters reject mode so every DML `Append` on the pchannel is refused with `RateLimitRejected` until the node is restarted.
  Suggestion: (1) test the condition the deserializer actually uses, for example `len(msgID) == 0 && walName != message.WALNameWoodpecker`, or give `SeekablePosition` an explicit WAL-name argument so a WoodPecker cluster is not rejected for a `WALName` field datacoord never fills in. (2) Do not fail the whole pchannel on one vchannel: either skip that vchannel the way `errChannelLifetimeUnrecoverable` is handled at `util.go:39-54`, or return the error from inside `getRecoveryInfo`'s `retry.Do(retry.AttemptAlways())` so the flusher waits for the coordinator to record a real position, as `SeekablePosition`'s own doc comment prescribes. In either case, ensure every exit path from `Execute` pairs `EnterSlowdownMode` with `EnterRecoveryMode`. (raised by tinswzy)

czs007:

This adversarial review round finds the previous must-fix on the flusher recovery-position guard unchanged at head `6307048d68`, with the two new commits adding only a lint substitution and a RocksMQ-only test that locks in the flagged behavior.

## High

- **`internal/streamingnode/server/flusher/flusherimpl/util.go:75`** — The unseekable-position guard uses a predicate keyed on `position.WALName` (`msgdispatcher.SeekablePosition`, which only admits an empty `MsgID` when `position.GetWALName() == WALName_WoodPecker`), while the deserialization on line 82 uses the flusher's own `walName` from `impl.wal.Get().WALName()` (line 70). datacoord does not guarantee `WALName` is populated: compaction results (`internal/datacoord/meta.go:2484-2490`), import segments (`meta.go:1994-2004`), and `toMsgPosition` (`handler.go:656-659`) all build positions with only `ChannelName`/`Timestamp`, and `GetChannelSeekPosition` (`handler.go:625-635`) returns such a position directly when the channel checkpoint is missing; `isDroppedChannel` (`util.go:131-133`) does not intercept it because it also requires `Timestamp == MaxUint64`. Before this PR, `DeserializeWoodpeckerMsgID(nil)` accepted this input as earliest `{0,0}`. **Failure:** WoodPecker backend, a vchannel with no channel checkpoint (checkpoint init failed with warn-and-continue at `services.go:1485-1489`, or a split target vchannel introduced by this PR restarted before its first checkpoint report) whose fallback earliest-segment position comes from compaction/import output → `getRecoveryInfos` returns `ErrServiceInternal`, `buildFlusherComponents` fails, and the whole pchannel's flusher goroutine exits, where the same input previously recovered from earliest. **Suggestion:** key the check on the flusher's WAL name (`len(msgID) == 0 && walName != message.WALNameWoodpecker`), or give `SeekablePosition` an explicit WAL-name parameter. *Previously raised (round 1, at `util.go:63`). The author has not replied; the two commits that followed landed roughly two minutes after the review comment and only replaced `errors.Errorf` with `merr.WrapErrServiceInternalMsg` (`2e9e91773b`) and added a RocksMQ-only test (`6307048d68`). The predicate, the WAL-name mismatch, and the upstream producers of empty-`MsgID` positions are unchanged, so the finding stands.* (raised by tinswzy, bigsheeper)

- **`internal/streamingnode/server/flusher/flusherimpl/wal_flusher.go:104`** — A validation failure on a single vchannel exits the entire pchannel flusher unrecoverably and leaves it in slowdown/reject mode permanently. The error from `getRecoveryInfos` (`util.go:79`) is raised outside `getRecoveryInfo`'s `retry.Do(retry.AttemptAlways())` (`util.go:95-127`), propagates through `buildFlusherComponents` (`wal_flusher.go:162-166`) to the early return at `wal_flusher.go:102-105`. The deferred handler (`wal_flusher.go:82-84`) only calls `DPanic`, which in `pkg/mlog/logger.go:341-347` panics only in zap Development mode, which nothing in the repository enables, so in production it is one log line. The goroutine is started fire-and-forget (`wal_flusher.go:58`, `opener.go:253-261`) and never restarted. `EnterSlowdownMode(nil)` at `wal_flusher.go:92` is paired with `EnterRecoveryMode()` only on the success path at line 117; with a nil checker the rate limiter descends to LWM and enters reject mode (`adaptive_rate_limit_controller.go:297-303`, `325-327`), and the only other exit is `FlusherRecovering.Close()` on WAL shutdown. On merge-base, `getRecoveryInfos` returned errors only on context cancellation, so this guard is the first non-cancel hard failure reaching this path. **Failure:** the trigger condition of the item above, or any single vchannel whose recovery position is rejected by `SeekablePosition` → the flusher goroutine exits with one `DPanic` log line; all vchannels on the pchannel stop flushing, checkpoints stop advancing, the WAL is never truncated, and `FlusherRecovering` stays in slowdown and degrades to reject until the WAL is closed. **Suggestion:** skip the vchannel the way `errChannelLifetimeUnrecoverable` is handled (`util.go:50-65`), or move the check inside `getRecoveryInfo`'s `retry.Do` so the flusher waits for the coordinator to supply a real position; if a hard failure is kept, every exit path of `Execute` must pair `EnterSlowdownMode` with `EnterRecoveryMode`. *Previously raised as part (b) of the round-1 must-fix on the same guard. The author has not responded, and neither of the new commits touches the `Execute` exit paths or the pairing of slowdown/recovery mode.* (raised by tinswzy, bigsheeper)

## Medium

- **`internal/streamingnode/server/flusher/flusherimpl/util_test.go:30`** — `TestGetRecoveryInfosRejectsUnseekablePosition` only exercises RocksMQ: `newMockWAL(t, true)` returns `message.WALNameRocksmq` (`wal_flusher_test.go:669`) and the mock position carries no `WALName`. Rejection is correct for RocksMQ, since `DeserializeRmqID` (`rmq_id.go:59-61`) would panic on empty bytes, but the WoodPecker path, where the deserializer succeeds and the guard still rejects, has no coverage; `TestSeekablePosition` (`dispatcher_test.go:188-200`) tests the predicate alone, not `getRecoveryInfos` under a WoodPecker WAL. As written, the test locks in the behavior flagged above rather than fixing it. **Failure:** switching the mock WAL's `WALName()` to `message.WALNameWoodpecker` with the same timestamp-only seek position → the current code returns `ErrServiceInternal` instead of resolving the empty `MsgID` to earliest `{0,0}`, which the existing RocksMQ-only test cannot observe. **Suggestion:** add a mock WAL whose `WALName()` returns `message.WALNameWoodpecker` and assert that an empty-`MsgID` position is accepted and parsed as earliest. (raised by tinswzy)

xiaocai2333:
Both must-fix items confirmed, and the finding is worse than stated. Fixed in `9b57e40612`; every claim below was checked in the code before anything was changed.

## What was verified

**The predicate is keyed on the wrong field.** `SeekablePosition` admits an empty `MsgID` only when `position.GetWALName() == WoodPecker`; the deserializer on the next line uses `impl.wal.Get().WALName()`. Datacoord never fills the position's `WALName`: compaction output (`meta.go:2484`), import segments (`meta.go:1994`) and `toMsgPosition` (`handler.go:656`) all build `{ChannelName, Timestamp}` only, and `GetChannelSeekPosition` returns exactly that from `getEarliestSegmentDMLPos` when the channel checkpoint is missing. `isDroppedChannel` requires `Timestamp == MaxUint64` so it does not intercept. `DeserializeLogMessageId(nil)` is `proto.Unmarshal(nil)` → `{0,0}`, so on WoodPecker this input recovered from earliest on merge-base and the guard rejected it.

**The failure takes the pchannel down.** The check sits in `getRecoveryInfos`, outside `getRecoveryInfo`'s `retry.Do(AttemptAlways)`. `Execute`'s deferred handler has no `recover` and only logs `DPanic`; `zap.Development()` is applied only when `cfg.Development`, which nothing sets. `go flusher.Execute(...)` is fire-and-forget. `EnterSlowdownMode(nil)` is paired with `EnterRecoveryMode()` only after `buildFlusherComponents` succeeds, and the controller's slowdown path with a nil checker descends to LWM and enters reject mode after `RejectDelayInterval`.

**What merge-base did, which the review did not say and which decides the fix.** For the same input, WoodPecker seeked to earliest; every other backend panicked inside the deserializer (`DeserializeRmqID` is `Endian.Uint64` on empty bytes — `index out of range [7] with length 0`), and with no `recover` in `Execute`'s defer that was a process crash. So the guard I added converted a loud crash into a silent hang **and** broke the one backend that worked. Neither suggested fix — skip the vchannel, or retry inside `retry.Do` — is right either: skipping a live vchannel is the D7 silent-drop shape, and retrying forever deadlocks, because the channel checkpoint that would end the retry is produced by the flusher that cannot start.

## The fix

An empty message id is a position datacoord **built without one**, not a corrupt one, and the only safe reading of "no known position" is the earliest the WAL retains: seeking earlier than necessary costs replay time, which recovery is built to absorb; seeking later loses data. `recoveryMessageID` makes that reading explicit and identical on every backend via `MustGetEarliestMessageIDFromMQType`, keyed on the flusher's WAL name. `SeekablePosition` is no longer consulted here.

`Execute`'s exit paths are deliberately untouched: after this change no non-cancel hard failure reaches them from this code, and switching the limiter off when the flusher has died would let the WAL grow unbounded. The answer to that is not to die.

## Test — Medium item included

`TestGetRecoveryInfosSeeksEarliestWhenThePositionHasNoMessageID` runs the timestamp-only position through a real `getRecoveryInfos` under a RocksMQ mock WAL **and** a WoodPecker one, asserting the checkpoint is that backend's earliest id. I confirmed it observes the asymmetry the review described: with the merge-base semantics restored, the rocksmq case fails with the original panic and the woodpecker case passes; with the fix both pass. The RocksMQ-only test that locked in the rejection is gone.

Lint (same golangci-lint and tags as CI) 0 issues; `internal/streamingnode/...`, `internal/distributed/streaming/...`, `internal/util/streamingutil/...`, `internal/flushcommon/...` green.

czs007:

This round verified the fixes for the two must-fix items reported previously on the flusher recovery path, and the increment after `6307048d68e7` introduces no new findings.

**Verified:**
- `internal/streamingnode/server/flusher/flusherimpl/util.go:71-73` — recovery now passes the flusher's own WAL name into `recoveryMessageID`, so the position's unset `WALName` no longer decides anything and the guard/deserializer mismatch is gone.
- `internal/streamingnode/server/flusher/flusherimpl/util.go:147-156` — an empty recovery message ID is converted to the backend's earliest retained ID instead of returning an unseekable-position error or entering the backend deserializer.
- `pkg/streaming/util/message/adaptor/message_id.go:127-144` — the four earliest-ID types produced there are the same concrete types `MustGetMessageIDFromMQWrapperID` already accepts, so no nil `MessageID` can reach the `LT(checkpoint)` comparison at `util.go:74`.
- `pkg/mq/mqimpl/rocksmq/server/rocksmq_impl.go:959-980` — the synthetic RocksMQ earliest ID `-1` equals the fresh consumer group's initial position, so the seek takes the reset branch and does not hit the "move consume position backward" panic; the Kafka `-2` sentinel likewise round-trips to `OffsetBeginning`.
- `internal/streamingnode/server/flusher/flusherimpl/wal_flusher.go:102-105` — the only remaining error return from `getRecoveryInfos` is the pre-existing wrapped RPC error, so an unseekable position can no longer terminate the pchannel flusher or leave rate limiting stuck in slowdown mode.
- `internal/streamingnode/server/flusher/flusherimpl/util_test.go:46-80` — the test asserts `checkpoint.EQ(earliest)` for both RocksMQ and Woodpecker with a nil-MsgID position, which fails on the old code, so it genuinely pins the fix.

**Previously raised items:** the two high-severity findings on `util.go:63/75` and `wal_flusher.go:104`, and the medium-severity test-coverage gap at `util_test.go:30`, were addressed by the author in commit `9b57e40612`. Those responses are accepted and the items are closed.

One observation, non-blocking: a single vchannel with an empty position now rewinds the whole pchannel scanner to the WAL earliest rather than only that vchannel. This is safe because the recovery storage and genesis data sync service both discard messages at or below their respective checkpoints, and the scanner start was already the minimum across vchannels. The cost is replay time, which the comment at `util.go:136-142` states explicitly, and it replaces a crash loop.

xiaocai2333:
## Review of head `9b57e40612` (22 commits)

**Verdict:** mergeable after one Medium fix and a PR-body refresh. The five adversarial rounds above already closed the Produce-stream `FencedTimeTick` loss, the L0 delete buffer, and the flusher recovery-position guard; this pass re-traced the remaining paths (interceptor order, barrier allocation, function-runner alloc/release on CreateVChannel/DropVChannel, replay discard guards in recovery storage and the DSS wrapper, design-doc §6.3/§8 references) and found one new gap.

### Medium — `CreateVChannel` registers a nil schema when the body carries the legacy `Schema` bytes

`shardInterceptor.handleCreateVChannel` accepts a body with either `CollectionSchema` set or `len(Schema) != 0`, and `newVChannelRecoveryInfoFromCreateVChannelMessage` parses both forms through `MustGetSchemaFromCreateCollectionMessageBody`. But `shardManagerImpl.CreateVChannel` passes `msg.MustBody().GetCollectionSchema()` straight into `createCollectionLocked` (`shard_manager_collection.go`), so the bytes form registers a nil schema. `CreateCollection` on the same file handles the bytes form explicitly.

Consequence: every versioned insert to the new target fails with `ErrCollectionSchemaNotFound` until the streamingnode restarts, at which point recovery rebuilds the entry from the persisted meta with a real schema — exactly the "different before and after a restart" case the interceptor guard's comment says it exists to prevent. `InitSplitTargetVChannels` in this PR always sets `CollectionSchema`, so nothing triggers it today, but the guard explicitly admits the other form.

Repro (drop into `shards`, run with `-tags dynamic,test`):

```go
func TestCreateVChannelSchemaBytesForm(t *testing.T) {
	m := newTestShardManagerWithVChannelState(t, streamingpb.VChannelState_VCHANNEL_STATE_NORMAL, 0)
	schema := &schemapb.CollectionSchema{Name: "col", Version: 3, Fields: []*schemapb.FieldSchema{
		{FieldID: 100, Name: "id", DataType: schemapb.DataType_Int64, IsPrimaryKey: true}}}
	bs, _ := proto.Marshal(schema)

	msg := message.NewCreateVChannelMessageBuilderV2().WithVChannel("v9").
		WithHeader(&message.CreateVChannelMessageHeader{CollectionId: 9, PartitionIds: []int64{10}}).
		WithBody(&message.CreateCollectionRequest{Schema: bs}).
		MustBuildMutable().WithTimeTick(2000).WithLastConfirmedUseMessageID()
	m.CreateVChannel(message.MustAsImmutableCreateVChannelMessageV2(msg.IntoImmutableMessage(rmq.NewRmqID(3))))
	_, err := m.CheckWritableAndSchemaVersion("v9", &message.InsertMessageHeader{CollectionId: 9, SchemaVersion: proto.Int32(3)})
	assert.NoError(t, err) // FAILS: collection schema not found

	cmsg := message.NewCreateCollectionMessageBuilderV1().WithVChannel("v8").
		WithHeader(&message.CreateCollectionMessageHeader{CollectionId: 8, PartitionIds: []int64{10}}).
		WithBody(&message.CreateCollectionRequest{Schema: bs}).
		MustBuildMutable().WithTimeTick(2000).WithLastConfirmedUseMessageID()
	m.CreateCollection(message.MustAsImmutableCreateCollectionMessageV1(cmsg.IntoImmutableMessage(rmq.NewRmqID(4))))
	_, err = m.CheckWritableAndSchemaVersion("v8", &message.InsertMessageHeader{CollectionId: 8, SchemaVersion: proto.Int32(3)})
	assert.NoError(t, err) // passes
}
```

Observed: `CreateVChannel bytes-form: version=-1 err=collection schema not found` vs `CreateCollection bytes-form: version=3 err=<nil>`.

Fix: resolve the schema in `CreateVChannel` the way `CreateCollection` does (same helper), and keep the repro as the regression test.

### Low

- **Dispatcher fan-out of `SplitShard` on the querynode.** `msgdispatcher.groupAndParseMsgs` derives a vchannel only for Insert/Delete and a collection id only for the four collection/partition DDLs; for every other type `collectionID == ""`, and `strings.Contains(k, "")` matches every target. So the fence a delegator will eventually act on is currently delivered to **every delegator on the pchannel, across collections**. Harmless today (filter node's `default:` drops it at Debug), but the PR body's "necessary, not sufficient" list names only the filter node; the read-path PR must also teach the dispatcher to route `SplitShard` by its vchannel, or every delegator on the pchannel will process a foreign fence.
- `message_type.go`: the `CreateVChannel` comment says it is "appended exclusively under the collection's DDL lock". It is a single-vchannel `RawAppend` with no broadcast and no ResourceKey, so no collection DDL lock is involved; only the vchannel-level exclusive lock applies. The comment should say that, because the missing DDL exclusion is exactly what makes the partition/schema reconciliation the coordinator's job.
- `handleSplitShardMessage` uses `status.NewUnrecoverableError(err.Error())` twice; the rest of this PR's new code uses `"%s", err.Error()` (matches the pre-existing `ManualFlush` pattern, so cosmetic).
- `recoveryMessageID`'s empty-MsgID → earliest fallback is a behavior change on a path every backend shares, not a shard-split-only one: RocksMQ/Kafka/Pulsar previously panicked on that input, now they silently replay from the earliest retained position with an Info-level log. Already verified safe above; it just deserves an explicit line in the PR body, and the "flusher recovery checkpoint candidate" log for the `positionHadMessageID=false` case arguably belongs at Warn.

### PR body is stale

- "Requires milvus-io/milvus-proto#618. CI cannot go green until that merges" — #618 merged 2026-09-01 as `a301a6af92` and this PR pins it (`6294e3af93`); Build / Code-Check / UT-GO / UT-Integration are green.
- None of the 15 commits after `05bf27cf46` are reflected: the Produce-response `FencedTimeTick` fix, `fenced_split_task_id` and the abort on another task's fence, the `fencedVChannels` tombstone for a retired source, the flush timestamp at the fence, the drain precondition on `DropSplitVChannel`, the earliest-position fallback, and the delegator whitelist entry.

### CI

The only red stage, UT-CPP-Cov, fails on `test_chunk_segment.ReopenInvalidatesOffsetsForReaddedStructArrayWithSameName`, which is also red on pipelines 29147/29148/29150 of unrelated PRs; this PR touches no C++ file. The 0% diff-coverage figure is a tooling artifact (787 measurable lines, 0 hit, while UT-GO ran and passed the new tests).

xiaocai2333:
Verified the Medium by running your repro before touching anything: `CreateVChannel bytes-form` registered a nil schema (`collection schema not found` from `shard_manager_collection.go:506`) while `CreateCollection bytes-form` resolved version 3. Fixed in `ddb83c1d33` + `1dce72f984`.

## Medium — `CreateVChannel` registers a nil schema for the bytes form

Both genesis messages now resolve the schema through one helper (`schemaOfCreateBody`: the `CollectionSchema` message, else the pre-2.6.1 bytes via `MustGetSchemaFromCreateCollectionMessageBody`, else nil). Your repro is the regression test, `TestShardManagerCreateVChannelResolvesTheSchemaBytesForm`, asserting version 3 back from both `CheckWritableAndSchemaVersion` calls; it fails on the previous head.

## Low

- **`message_type.go` comment** — now says what is true: a single-vchannel append with no broadcast and no resource key, so only the vchannel-level exclusive lock applies, and reconciling the carried partition list and schema against concurrent DDL is the coordinator's job.
- **`NewUnrecoverableError(err.Error())`** — both sites in `handleSplitShardMessage` go through `"%s"` now.
- **Earliest-position fallback** — the candidate-without-a-message-id case now logs at Warn on its own line; the Info line stays for ordinary candidates. The PR body has a paragraph stating this is a cross-backend behavior change (RocksMQ/Kafka/Pulsar previously panicked the pchannel's flusher) and why replaying from earliest is safe.
- **Dispatcher fan-out of `SplitShard`** — confirmed in `groupAndParseMsgs`: no vchannel is derived for the type, so the empty key matches every target. Nothing in this PR consumes the message on the querynode, so it stays a read-path item; the PR body's "necessary, not sufficient" list now names the dispatcher alongside the filter node so the read-path PR cannot miss it.

## PR body

Rewritten against the 17 commits since `05bf27cf46`: #618 merged and pinned, produce-response error cloning, `fenced_split_task_id` and the abort on another task's fence, the `fencedVChannels` tombstone, the flush timestamp at the fence, the drain precondition on `DropSplitVChannel`, the single-lock insert admission, `WithUnreplicable()` on all three messages with the §8.6 gate as follow-up, and the earliest-position fallback.

`shards`, `shard`, `flusherimpl` and `pkg/streaming/util/message` green locally with `-tags dynamic,test`; `golangci-lint` clean on the touched packages. UT-CPP-Cov's `ReopenInvalidatesOffsetsForReaddedStructArrayWithSameName` is red on unrelated pipelines too; this PR touches no C++ file.

xiaocai2333:
Rebased onto master `9c0c8a0cbf` as `2c0b71439b` (25 commits). Two conflicts, both resolved the obvious way:

- **`messages.proto` / `messages.pb.go`** — #52072 (RLS) took `MessageType` 47 and 48 for `AlterRLSMetadata` / `DropRLSMetadata`, the same numbers this PR had used. `SplitShard` / `CreateVChannel` / `DropVChannel` are now **49 / 50 / 51**; `messages.pb.go` regenerated with the repo's protoc 27.0 / protoc-gen-go v1.33.0 (no generator-version drift in the diff). Nothing appends these messages yet, so no persisted WAL carries the old numbers.
- **milvus-proto pin** — #52072 pinned `c39cddab3fac`, which is three commits *behind* this PR's `a301a6af926b` (`compare/a301a6...c39cddab` = behind by 3). Kept `a301a6af926b` in all four modules; it contains the RLS proto types as well as the shard routing ones.

Verified on the rebased head: `pkg` module builds and `streaming/util/message` + `proto` tests pass; `shard/...`, `flusherimpl`, `distributed/streaming`, `streamingutil/status`, `wal/recovery` green with `-tags dynamic,test`.

The UT-GO red on `2f4b1e5a27` was `TestFillOutputFieldsOrdered_NoCMemoryLeak` in `internal/querynodev2/tasks`, untouched by this PR and a master-wide flake (see the note on #50495: a ≤2-of-5 jemalloc-window assertion with a 96 KiB threshold against ±1 MiB swings; it also failed on #52828's build 29606).

czs007:

Adversarial review of this PR found three medium-severity correctness/concurrency issues in the new split-shard fence path and one low-severity validation gap, all introduced by this PR.

## Medium

- **`internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go:425`** — A failed fence append orphans the segments it already sealed, and the retry loses their IDs. `handleSplitShardMessage` seals before appending: `FlushAndFenceSegmentAllocUntil` flips every growing segment to FLUSHED in memory, unregisters it from the stats manager, and replaces `m.segments` with an empty map. The only carrier of those sealed IDs is `header.FlushedSegmentIds` of the append at line 432; on failure the error is returned as-is and nothing preserves them. **Failure:** SplitShard append fails after a growing segment S was sealed, then the coordinator retries the fence with a fresh timetick → the persisted header omits S; S is unregistered from the stats manager and absent from `m.segments`, so no seal policy or `AsyncFlushSegment` can ever flush it, and it stays GROWING at DataCoord and in recovery storage until a streamingnode restart. The comment at lines 412–424 ("costs a premature flush, never correctness") holds for ManualFlush/SchemaChange because the shard keeps taking writes, but not for a fence declared "the single authoritative seal record for T_switch", and design doc §6.3 gates the source drain on no healthy segment remaining on the source. **Suggestion:** keep the sealed IDs on the partition manager (or a per-vchannel pending-fence list) until an append persists them, or re-derive them from recovery storage on retry. `shard_interceptor_split_test.go:53-55` mocks `FlushAndFenceSegmentAllocUntil` and cannot see this. *(raised by sijie-ni-0214, czs007, bigsheeper)*

- **`internal/streamingnode/server/wal/interceptors/redo/redo_interceptor.go:59`** — An Insert to a retired (dropped) source vchannel returns UNRECOVERABLE instead of SHARD_FENCED. Interceptor order is redo → lock → replicate → timetick → shard, so `waitUntilGrowingSegmentReady` runs before the shard interceptor's name-aware admission. It calls `checkIfPartitionExists`, which returns `ErrCollectionNotFound` once `m.collections` has no entry for the collection; `DropVChannel` (new in this PR, `shard_manager_collection.go:400`) deletes exactly that entry while deliberately keeping the `fencedVChannels` tombstone. The redo interceptor maps `ErrCollectionNotFound` to `NewUnrecoverableError`, so the insert never reaches `CheckWritableAndSchemaVersion`. Deletes go through `handleDeleteMessage` → `CheckIfVChannelCanBeWritten` and do get SHARD_FENCED, so the two DML paths disagree, and the PR body's claim that a retired source keeps answering SHARD_FENCED does not hold for inserts. **Failure:** source vchannel V of collection C is fenced and then retired by DropVChannel on pchannel P, and a proxy with a stale route sends an Insert to V with no other vchannel of C registered on P → the insert is rejected with STREAMING_CODE_UNRECOVERABLE "fail to wait growing segment ready, collection not found" while a Delete on the same route gets SHARD_FENCED. The window lasts until another vchannel of the same collection is registered on that pchannel. Today the only consumer of `IsShardFenced` is the split coordinator's re-fence path, so the current effect is a wrong wire code rather than a lost retry. **Suggestion:** in `waitUntilGrowingSegmentReady` (or `WaitUntilGrowingSegmentReady`), consult `checkIfVChannelCanBeWritten(collection, msg.VChannel())` first and return `status.NewShardFenced` on `ErrVChannelFenced`. Existing tests exercise only the shard interceptor/manager with mocks and cannot observe the redo path. *(raised by sijie-ni-0214, czs007, bigsheeper)*

- **`internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go:508`** — A CreateSegment in flight at fence time lands on the fenced source vchannel and is never sealed by the fence. CreateSegment is `SelfControlled`, not `ExclusiveRequired`, so the lock interceptor gives the alloc worker's append the ordinary vchannel RLock while SplitShard takes the write lock. Any `asyncAllocSegment` worker that has not yet acquired its RLock when SplitShard takes the write lock blocks and proceeds after the fence is persisted. `handleCreateSegment` admits it via `CheckIfSegmentCanBeCreated`, which checks only partition existence and segment uniqueness, not the fence; `FlushAndFenceSegmentUntil` clears `m.segments` but leaves `onAllocating` open; and `AddSegment` accepts any segment whose create tick exceeds `fencedAssignTimeTick`. This is a deterministic interleaving, not a rare race. **Failure:** an insert triggers `asyncAllocSegment` and SplitShard acquires the vchannel write lock before the worker's CreateSegment append acquires its RLock → after the fence is persisted, a 0-row GROWING segment is registered on the fenced vchannel, absent from the SplitShard `FlushedSegmentIds`, unsealable by the idle policy (requires BinarySize above `l1MinSizeFromIdleTime`), and sealed only when the lifetime policy fires (`dataCoord.segment.maxLife`, default 24h), so the source shard holds a healthy growing segment that the §6.3 drain condition forbids for up to 24h. **Suggestion:** have `CheckIfSegmentCanBeCreated` (or `handleCreateSegment`) refuse a CreateSegment on a SPLITTED/fenced vchannel with an unrecoverable error (the worker stops on `IsUnrecoverable`), and close `onAllocating` in the split path so redo waiters parked on it are released. *(raised by bigsheeper, czs007)*

## Low

- **`internal/distributed/streaming/splitshard.go:159`** — `SplitTaskID=0` is treated as "my own fence" due to the `taskID != 0` exemption. On a SHARD_FENCED reply, the abort is skipped when the fenced task id is zero; the comment at lines 169–171 justifies this as "a fence placed by a build that did not record one", but the SplitShard message type is introduced by this PR, so no shipped build can have written such a fence. `SplitShardParam.Validate` (lines 73–114) never requires `SplitTaskID > 0`, and the header is built directly from `param.SplitTaskID`. **Failure:** two SplitShard callers on the same source vchannel both leave SplitTaskID at 0 → the second caller's fence is rejected with SHARD_FENCED carrying `FencedSplitTaskId` 0, line 159 skips the abort, and it receives the first task's T_switch as `ErrSourceVChannelFenced`, rolling forward on another task's fence, which is exactly the outcome the `GetSplitFence` comment in `shard_manager_collection.go:76-79` says the task id exists to prevent. **Suggestion:** tighten `Validate` to require `SplitTaskID > 0`, or drop the zero exemption. *(raised by czs007, bigsheeper)*

czs007:

Adversarial review of this PR surfaced two medium and two low findings, all introduced by this change; none block on their own but the two medium items concern permanent state and a bypassed safety gate.

### Medium

- **`internal/distributed/streaming/splitshard.go:73`** — `SplitShardParam.Validate` checks `CollectionID`, `SourceVChannel`, `RoutingModulus`, and each target's vchannel/residues but never rejects a zero `SplitTaskID`. The value is persisted (via `shard_manager_collection.go:217-224` and `vchannel_recovery_info.go:244`) and is the sole discriminator at `splitshard.go:159`, where `taskID != 0 && taskID != param.SplitTaskID` treats a zero-id fence as "our own". Failure: a caller leaves `SplitTaskID` at its zero value and calls `SplitShard`, then a second task (e.g. id 777) calls `SplitShard` on the same source → the first fence is persisted with task id 0, the second call receives `SHARD_FENCED` with `FencedSplitTaskId=0`, skips the abort path, and rolls forward as if the fence were its own, so two tasks split the same source into disjoint targets without knowing about each other. Suggestion: add a `SplitTaskID <= 0` rejection to `Validate`. Note that the analogous concern for `DropSplitVChannelParam.Validate` does not hold: `DropVChannelMessageHeader.SplitTaskId` is only written in this PR and never read back. (raised by bigsheeper, tinswzy)

- **`internal/streamingnode/server/wal/interceptors/replicate/replicates/impl.go:114`** — The `IsUnreplicable()` branch returns `ErrNotHandledByReplicateManager` before the role switch at `:128`, so the `RoleSecondary && rh == nil → NewReplicateViolation` check at `:134-137` is unreachable for the three new messages this PR marks `WithUnreplicable()` (`splitshard.go:149`, `:312`, `:409`). `replicate_interceptor.go:62-64` then appends directly, and no interceptor downstream (`walmanager/manager_impl.go:37-42`) re-checks the role. The marker therefore does more than isolate these messages from the CDC stream: it removes the only existing gate against direct appends on a secondary. Failure: a direct (non-replicated) `SplitShard` / `CreateVChannel` / `DropVChannel` append on a secondary-role WAL → the message lands in the WAL, the source vchannel is permanently fenced on the secondary, and every subsequent replicated Insert/Delete targeting that vchannel is rejected with `ShardFenced` at `shard_interceptor.go:269/374`, stalling replication for that vchannel. There is no caller in this PR yet, but the design (`20260610-shard_split.md:700-701`) requires a StreamingNode-side rejection. Suggestion: either evaluate the role check before the unreplicable short-circuit for these message types, or land the planned streamingnode-side gate before any caller of these APIs ships. (raised by tinswzy)

### Low

- **`internal/distributed/streaming/splitshard_test.go:222`** — `TestSplitShardParamValidate` covers `CollectionID<=0`, empty `SourceVChannel`, target count, empty target vchannel, and the two duplicate-name cases, but none of the three new validation branches: `RoutingModulus == 0` (`splitshard.go:80-82`), a target with no residue (`:103-105`), and `residue >= RoutingModulus` (`:106-112`). None of the three error messages is asserted anywhere in the diff. Suggestion: add one table case per branch. (raised by bigsheeper)

- **`pkg/mq/msgdispatcher/dispatcher.go:350`** — `groupAndParseMsgs` derives a vchannel only for Insert/Delete and a collection id only for the four collection-level DDLs, so `SplitShard` falls through with both empty and `strings.Contains(k, "")` matches every target: the fence is fanned out to every delegator on the pchannel, across collections. This becomes reachable because `msgstream_adaptor.go:138-145` adds `SplitShard` to the delegator whitelist. It is harmless today (`UnmarshaledMsg.Unmarshal` passes through, `SplitShardMessageBody.ID()` overrides the panicking default, and `filter_node.go:159-160` drops it with a Debug log), and the PR body already records this. Suggestion: treat this as a hard constraint for the follow-up that wires delegator handling: filter on the header's `CollectionId` and vchannel the way the `AddCollectionField`/`AlterCollection` arms at `filter_node.go:136-149` do, or derive a vchannel for `SplitShard` in `dispatcher.go:352-357`. (raised by bigsheeper, tinswzy)

xiaocai2333:
Reworked to the single-broadcast design agreed on 2026-09-09; head `9cc9a5b472`, 17 commits on top of `33fd462ddf` (the previous head is an ancestor, so the old review threads still apply to the code they cite).

What changed, in one paragraph: the three messages became one `SplitShard` broadcast to source + targets + control channel, each replica read by its role; ordering across pchannels comes from `BroadcastHeader.append_first_vchannels` (the broadcaster lands and persists the sources before the rest) plus the `FreshTimeTick` type property (the timetick interceptor takes a freshly fetched TSO batch); a same-task re-fence appends again idempotently; retirement is the `AlterCollection` replica that delists the vchannel. The PR body is rewritten around this and names the two coordinator-side guards that are prerequisites for wiring the message (`AllocVirtualChannels` excluding occupied pchannels; one active task per vchannel), because a refused replica is retried by the broadcaster forever while it holds the collection's DDL lock.

Every commit was implemented against a written plan with a per-task review and a whole-branch review; the four Important findings from the latter (partial ack vs `AckSyncUp`, stale SHARD_FENCED-recovery comments, the `TaskID != 0` clause, the prerequisite statement) are fixed in the last seven commits. `BarrierTimeTick` / `AllocateWithBarrier` are untouched: ManualFlush uses them.

Locally green with `-tags dynamic,test`: `pkg/streaming/util/message/...`, `idalloc`, `timetick/...`, `broadcaster/...`, `shard/...` (89.1% / 87.9%), `recovery` (82.3%), `flusherimpl`, `distributed/streaming`; `golangci-lint` v2 clean on the touched trees. `mock_ShardManager.go` was hand-edited for the two changed signatures because mockery v2.53.3 cannot run against grpc@v1.82.1 on this toolchain.

Issue #51659, opened 2026-07-21
[Bug]: Runtime transient dependency failures are escalated to process panics


zhuwenxing:
### Is there an existing issue for this?

- [x] I have searched the existing issues

Related issues #50915 and #51643 cover specific Vortex/object-storage
failures, but they do not account for the complete failure family observed in
the latest TC Stage panic audit: ordinary SyncTask failures, multiple non-Vortex
S3 transient errors, and etcd watcher leader changes.

### Environment

```markdown
- Milvus version: Multiple TC Stage 2.6/master builds; the exact git commit was not retained for every occurrence
- Deployment mode: Distributed cluster and standalone
- MQ type: Kafka and Pulsar test deployments
- SDK version: N/A; these are background flush, allocation, and watcher failures
- OS: Kubernetes/Linux
- CPU/Memory: Not retained
- GPU: N/A
- Others:
  - Namespace: chaos-testing
  - Audit window: 2026-07-19T07:23:29Z to 2026-07-21T07:23:29Z
  - Collection status: complete; no truncated Loki intervals or stack query failures
  - Pod list: N/A; the ephemeral chaos instances were cleaned up
```

### Current Behavior

A complete 48-hour panic scan found 55 process panics across 13 unique
runtime-failure signatures and 22 Milvus instances.

The panics fall into three mechanisms:

| Runtime path | Signatures | Occurrences | Distinct instances | Trigger examples |
|---|---:|---:|---:|---|
| Ordinary `SyncTask` | 4 | 9 | 5 | MixCoord connection refused, S3 `SlowDownWrite`, `SlowDownRead`, `XMinioServerNotInitialized` |
| `GrowingSourceSyncTask` | 7 | 26 | 10 | S3 network timeout, HTTP 503, MinIO not initialized, manifest write failure |
| etcd runtime watchers | 2 | 20 | 11 | `etcdserver: leader changed` |

#### Ordinary SyncTask

Representative errors:

```text
panic: AllocID Failed:
empty grpc client: failed to connect ... connect: connection refused
[recovered, repanicked]
```

```text
panic: FFI operation failed:
ERROR: StorageTransientService(code 110) details:
IOError: AWS Error UNKNOWN (HTTP status 503) during PutObject operation:
SlowDownWrite: Resource requested is unwritable, please reduce your request rate
...
loon FFI transient error
[recovered, repanicked]
```

Representative stack:

```text
[...]

PR #52585, opened 2026-08-17
fix: retry transient sync and watcher failures

chyezh:
issue: #51659

- retry transient Storage V3 writer attempts, ID allocation, and DataCoord metadata persistence while preserving terminal fail-stop handling
- keep recoverable growing-source flush failures on the existing rollback and reschedule path without changing the global default panic handler
- re-establish proxy and session etcd watches after transient leader changes


czs007:

Reviewed at head `9933466`. The direction is right (the etcd watch part is clean, and recreating `BulkPackWriterV3` per attempt is the correct fix for the consumed-FFI-writer problem), but I don't think this is mergeable as-is: three of the new retry boundaries are unbounded *and* accept errors that are known to be permanent, and one of them runs inside the shared sync worker pool. Findings below, most severe first.

---

### high — 1. The whole-`Write` retry inherits `ErrLoonTransient`, which is attached to *every* loon failure, so permanent Phase-1 errors now retry forever

`internal/flushcommon/syncmgr/task.go:158-176`:

```go
write := func() error {
    writer = NewBulkPackWriterV3(..., segmentInfo.ManifestPath(), t.writeRetryOpts...)
    t.insertBinlogs, ..., err = writer.Write(ctx, t.pack)
    return err
}
...
err = retry.Do(ctx, func() error { return classifyStorageV3Err(write()) }, t.writeRetryOpts...)
```

`classifyStorageV3Err` (`pack_writer_v3.go:264-272`) admits anything matching `packed.ErrLoonTransient`. But that sentinel is attached unconditionally to *every* loon FFI failure — `internal/storagev2/packed/ffi_common.go:328-340`:

```go
return merr.Wrapf(ErrLoonTransient, "FFI operation failed: %s", errStr)
```

and its own doc comment says it "cannot reliably distinguish a transient failure from a permanent one … rely on a **bounded** retry budget plus outer error handling to keep the worst case finite." This PR removes exactly that bound: the production options are `retry.AttemptAlways()` (`write_buffer.go:1499`) and the import path defaults to `dataNode.import.maxWriteRetryAttempts=0` = unlimited (`internal/datanode/importv2/util.go:123-132`).

Failure scenario: loon rejects a record batch permanently (bad schema/type, wrong root path, S3 `AccessDenied`, unsupported column) inside `writeInserts`. Before this PR that error escaped `Write` immediately and the task terminated (fail-stop / import job failure). Now `classifyStorageV3Err` sees `ErrLoonTransient`, `retry.Do` never exhausts, and the sync task rewrites the same doomed data files every ≤10s forever. The only symptom is a `retry func failed` warn every 4th attempt. For import, the job neither succeeds nor fails, so DataCoord can never retry or report it.

Suggested fix: bound the *outer* loop independently of `t.writeRetryOpts` (attempts or max-elapsed budget), and on exhaustion let the error escape to `HandleError` as before. Alternatively gate the outer loop on a narrower predicate than raw `ErrLoonTransient` (only `merr.IsRetryableErr`, i.e. the segcore codes actually classified retriable), and leave `ErrLoonTransient` to the narrow Phase-2 commit retry where it already was.

### high — 2. Unbounded retry runs inside the shared, bounded sync worker pool → head-of-line blocking across all channels on the node

`internal/flushcommon/syncmgr/sync_manager.go:69-73` sizes one pool for the whole DataNode:

```go
initPoolSize := cpuNum * params.DataNodeCfg.MaxParallelSyncMgrTasksPerCPUCore.GetAsInt()
dispatcher := newKeyLockDispatcher[int64](initPoolSize)
```

`SyncTask.Run` executes in that pool, and after this PR it can block indefinitely in three new places (finding 1, finding 3, and `allocOneWithRetry`). Failure scenario: one collection with a permanently-failing V3 write (or a permanently-rejected `SaveBinlogPaths`) parks `initPoolSize` workers; flush for **every other collection and channel** on the node stops, with no error anywhere — previously those slots were released (task failed) even if the release path was a panic-restart.

Suggested fix: don't hold a worker slot across an unbounded retry — return the task to a retry scheduler (the growing-source path already has one: `scheduleGrowingSourceRetryLocked`) or cap in-slot attempts and requeue. At minimum export a metric/state for "sync task retrying > N attempts / > T seconds" so this is detectable; today it is invisible.

### high — 3. `retry.RetryErr(func(error) bool { return true })` on the DataCoord meta writer retries permanent rejections forever

`internal/flushcommon/pipeline/data_sync_service.go:325-327`:

```go
writebuffer.WithMetaWriter(syncmgr.BrokerMetaWriter(params.Broker, config.serverID,
    retry.AttemptAlways(), retry.MaxSleepTime(10*time.Second), retry.RetryErr(func(error) bool { return true }))),
```

Those opts land on `retry.Handle` in `meta_writer.go:128-153`. The closure's own `shouldRetry` is `!merr.IsCanceledOrTimeout(err)`, i.e. true for everything else, and the always-true `RetryErr` intentionally disables the framework's `InputError` short-circuit. `ErrSegmentNotFound`/`ErrChannelNotFound` are special-cased, but other permanent rejections are not — e.g. `validateTextSegmentStorage` in `internal/datacoord/services.go:664-666` returns a request-level rejection for a storage-version/text-index mismatch. That request will never succeed, yet it is now retried unbounded at 10s intervals, pinning the sync task (see finding 2) and the channel checkpoint. Before this PR the default 10 attempts exhausted and the error surfaced.

Suggested fix: keep `AttemptAlways()` for genuinely transient conditions but make the predicate discriminating, e.g. `retry.RetryErr(func(err error) bool { return !merr.IsNonRetryableErr(err) && merr.GetErrorType(err) != merr.InputError })`, or an explicit allow-list (not-ready / unavailable / timeout / etcd txn conflict). An always-true predicate on a cross-component RPC has no terminal state at all, which contradicts the PR description's "preserving terminal fail-stop handling".

### high — 4. Leader-change recovery still exits the process, because the rewatch retry predicate rejects the errors the rewatch step actually produces

Adding `ErrLeaderChanged` to `etcd.IsRetriableWatchErr` (`pkg/util/etcd/etcd_util.go:72-78`) is correct, and both rewatch paths do a full re-list, so no event loss (`sessionWatcher.handleReWatch` re-reads sessions and replays through `w.rewatch`; `ProxyWatcher.WatchProxy` runs `initSessionsFunc`, which clears meta first). The problem is the *predicate reuse* on the recovery step:

`internal/util/proxyutil/proxy_watcher.go:153-164`:

```go
err2 := retry.Do(ctx, func() error { return p.WatchProxy(ctx) }, retry.RetryErr(etcd.IsRetriableWatchErr))
if err2 != nil { ... panic("failed to handle etcd request, exit..") }
```

and `internal/util/sessionutil/session_util.go:944-949` (same predicate, `closeEventCh()` on failure).

`WatchProxy` first issues a unary `Get` under a `RequestTimeout` child ctx (default 10s, `proxy_watcher.go:88-93`). During a leader election that Get fails with `etcdserver: request timed out` / `no leader` / `context deadline exceeded` — none of which are in `IsRetriableWatchErr` (it matches only compaction + leader-changed + the three auth sentinels). `retry.Do` then aborts on the *first* attempt (`retry.go:76-82`), and:
- proxy watcher → `panic("failed to handle etcd request, exit..")`;
- session watcher → `closeEventCh()`, which callers treat as fatal: `internal/querycoordv2/server.go:651-662` calls `os.Exit(sessionutil.ExitCodeEtcd)`, `internal/datacoord/server.go:820-824` calls `stopServiceWatch()`.

So the new leader-change path only helps when the immediately following Get succeeds; if the election is still in progress — the exact window in which `ErrLeaderChanged` was raised — the coordinator dies anyway, and the new `session_util_test.go:345` test can't catch it because it stubs the rewatch hook.

Suggested fix: use a separate, wider predicate for the *re-establish* step (add `rpctypes.ErrNoLeader`, `rpctypes.ErrTimeout`/`ErrTimeoutDueToLeaderFail`, and `context.DeadlineExceeded` from the bounded child ctx), or simply drop `RetryErr` there and rely on `retry.Do`'s default bounded attempts + backoff. A test that makes the first `GetSessions` fail with a timeout and asserts the watch is still recovered would pin this.

### medium — 5. Removing the growing-source failure callback leaves the non-retryable branch with no escalation at all

`internal/flushcommon/writebuffer/write_buffer.go:1568-1574` drops `WithFailureCallback(wb.errHandler)`. For the retryable branch that's right (double handling). But the non-retryable branch now has no consumer: `markNonRetryableFailure()` (`:194`) sets a flag whose only readers are `growingSourceProgressSyncable` (`:693`) and `growingSourceProgressSelectedByPolicy` (`:1088`), both of which just return `false` forever. Meanwhile `getCheckpoint` keeps `progress.firstUncommittedPosition()` as a checkpoint candidate (`:582-590`).

Failure scenario: a `Column group size mismatch` (segcore 2001) on segment X. The segment is permanently excluded from flush, `progress.batches` is never drained, so the channel checkpoint is pinned at that position **forever** — unbounded WAL/MQ retention and a segment that can never flush — signalled only by one `mlog.Error` line. Previously `errHandler` fail-stopped the channel and recovery restarted from a clean checkpoint. Deliberately not panicking is a defensible choice, but there has to be *some* terminal action.

Suggested fix: in the `isGrowingSourceLayoutMismatch` branch, either invoke `wb.errHandler(err)` explicitly (keeping the fail-stop only where it is genuinely non-retryable), or fall back to the WriteBuffer source for that segment, or expose a channel-level "flush blocked" state + metric. Also worth asserting in `l0_write_buffer_test.go:1025-1032`: the current change only asserts `errorHandlerCalls == 0`, which locks in the silence without asserting any replacement.

Related, smaller: in the same callback, `if progress, exists := wb.growingSourceProgress[...]; exists` (`write_buffer.go:972`) means a failure arriving after the progress entry was deleted is now swallowed entirely — no rollback, no handler. Previously the task callback still fired.

### medium — 6. The outer V3 retry has no idempotency guard around Phase 2, and multiplies the retry budget

Two issues with wrapping all of `Write`:

(a) **Nested retry, same options.** `Write` already retries `CommitManifestUpdates` internally with the *same* `writeRetryOpts` (`pack_writer_v3.go:230-238`). With a bounded budget N (any operator setting `dataNode.import.maxWriteRetryAttempts`), the effective commit-call count becomes N×N and each outer attempt re-runs all of Phase 1, with the backoff sequence restarting from `sleep` every time. The budget an operator configures no longer means what it says.

(b) **No post-commit guard.** `finalizeStats` (`pack_writer_v2.go:70-87`) runs *after* the manifest commit succeeded. If it returns an error the outer loop re-enters `write()`, which builds a fresh writer from `segmentInfo.ManifestPath()` — the metacache value, which is only updated after the DataCoord ack (`task.go:213-224`), i.e. **the stale base version**. Attempt 2 then writes a second full set of data files and commits again; per the comment at `pack_writer_v3.go:227-229` loon resolves manifest-version conflicts by optimistic retry against the newest version, so both attempts' insert files can end up registered → duplicated rows, while `t.insertBinlogs` reported to DataCoord only describes attempt 2. Today this is latent (the only post-commit error is a non-retriable `ErrSegmentNotFound`), but nothing in the code enforces that.

Suggested fix: scope the outer retry to Phase 1 — e.g. have `Write` return a distinguishable post-commit error that `classifyStorageV3Err` always marks `Unrecoverable`, or assert `bw.manifestPath == bw.initialManifestPath` before allowing a re-attempt. And pass a distinct (bounded, non-multiplying) option set to the outer loop.

### medium — 7. Every failed attempt leaks data files and log IDs, unbounded under `AttemptAlways`

Each `write()` attempt calls `allocOneWithRetry` for fresh log IDs (`pack_writer_v3.go:446,561,672`) and writes new parquet/LOB/stat/delta objects; `insertFiles.Destroy()` in the `Write` defer releases the native handle, not the remote objects. Files from a failed attempt are referenced by no manifest version. Combined with finding 1 (infinite retry) that is an unbounded object-storage leak plus unbounded global-ID consumption. Please confirm DataCoord GC actually reclaims loon files that were never committed to any manifest — if GC is manifest-driven, nothing will ever collect them. Worth stating explicitly in the PR body either way.

### low — 8. The `retry.Unrecoverable` marker now leaks out of `SyncTask.Run`

`task.go:173-175` returns `retry.Do`'s error verbatim, and `retry.Unrecoverable` is `merr.Combine(err, errUnrecoverable)` (`retry.go:249-251`). So the sync error handed to `HandleError`, `taskObserverCallback` and any downstream string/`errors.Is` inspection now carries a retry-internal sentinel and an extra "unrecoverable error" suffix in `Error()`. `errors.Is` still works, but message-matching consumers (e.g. `isGrowingSourceLayoutMismatch`-style substring checks) are sensitive to this. Unwrap before returning.

### low — 9. Test coverage doesn't cover the terminal path this PR claims to preserve

- `task_test.go:406` (`TestRunStorageV3RetriesTransientInsertWrite`) only covers transient→success. There is no test that a non-retryable failure (e.g. `merr.SegcoreError(2001, …)` or a plain error) aborts the outer loop after exactly **one** attempt and still reaches the failure callback. Given that "preserving terminal fail-stop handling" is a stated goal of the PR, that assertion is the important one.
- `pack_writer_test.go:46` doesn't cover `allocOneWithRetry`'s `len(opts) == 0` passthrough (`pack_writer.go:53-56`).
- Nothing covers the import path (`newWriteRetryOptions`, unlimited by default) now going through the new outer loop.

### nit — 10. `handleReWatch` re-watches from `revision`, not `revision+1`

`session_util.go:967`: `clientv3.WithRev(revision)` where `revision` is `resp.Header.Revision` from `GetSessions` (`:698`). Watch is inclusive, so events at that revision are replayed even though the list already reflects them — duplicate `SessionAddEvent`s after every rewatch. `ProxyWatcher` gets this right (`proxy_watcher.go:92`: `WithRev(rev+1)`). Pre-existing, but this PR makes rewatch fire on leader changes, which are far more frequent than compaction, so the duplicate replay goes from rare to routine.

---

**Summary of what I'd ask for before merge:** bound the new whole-`Write` retry and keep it out of the shared worker slot (1, 2); make the meta-writer predicate discriminating instead of always-true (3); widen the rewatch retry predicate so leader-change recovery doesn't fall back to `os.Exit`/panic (4); give the growing-source non-retryable branch a terminal action (5). Findings 6-10 can be follow-ups, but 6(b) deserves at least an explicit guard or a comment stating the invariant it relies on.

chyezh:
@czs007 Thanks for the detailed review. I addressed the actionable recovery/control-flow findings in follow-up commit [`0a98e748df`](https://github.com/milvus-io/milvus/commit/0a98e748dfa245d9dcb765287de04882c40a120e). Here is the disposition item by item.

### Addressed

- **4 — rewatch used the wrong error predicate:** added a separate `IsRetriableEtcdRequestErr` for the unary re-list phase. Wrapped gRPC `Unavailable`, request deadline, leader/auth-token failures are recoverable; cancellation and permanent authorization/configuration errors are not. Recoverable failures now use `AttemptAlways` with capped backoff and terminate on the watcher/parent context. Closed proxy/session watch channels also enter the same recovery path.
- **5 — growing-source layout mismatch had no terminal action:** the known column-count/column-group mismatch branch now invokes `wb.errHandler(err)` after rollback and after releasing `wb.mut`. Recoverable growing-source failures still only rollback/reschedule. The test now asserts one handler invocation rather than locking in silence.
- **6(b) — post-commit retry boundary:** after `CommitManifestUpdates` succeeds, any `finalizeStats` error is explicitly marked unrecoverable, so the outer loop cannot restart Phase 1 against the stale metacache manifest. A test pins this boundary even for an otherwise recoverable storage error.
- **9 — missing terminal/passthrough tests:** added a Storage V3 non-recoverable writer test asserting exactly one write attempt and one failure callback, plus the no-options `allocOneWithRetry` single-call passthrough test.
- **10 — inclusive session revision:** rewatch now starts at `revision+1`. An embedded-etcd test verifies the snapshot revision is not replayed and the following revision is received.

### Intentionally retained for this PR

- **1, 2, 3, 6(a), 7:** after discussion we are deliberately keeping the current unlimited-retry policy for errors the existing code marks recoverable. `classifyStorageV3Err` is unchanged. We currently do not have reliable end-to-end FFI/downstream error codes to safely narrow `ErrLoonTransient` or every DataCoord rejection, so introducing a guessed bounded or text-based split would reintroduce the transient-outage panic/loss window this issue targets. The resulting worker occupancy, repeated Phase-1 orphan files/ID gaps, nested retry budget behavior, and the possibility that a producer-misclassified permanent error retries indefinitely are accepted limitations of this policy for now. Cancellation still terminates the loops, and known stale segment/channel metadata retains its existing terminal handling.
- **8:** the retry marker propagation is unchanged in this follow-up. `errors.Is` and the existing layout-mismatch substring classification remain valid; changing the retry framework error projection is outside this recovery-boundary patch.
- The smaller race mentioned under **5** where progress may already have been removed is also unchanged; the explicit terminal handler applies when the tracked growing-source task is still owned by the write buffer.

Validation on the follow-up:

```text
go test ./internal/flushcommon/syncmgr ./internal/flushcommon/writebuffer ./pkg/util/etcd ./internal/util/proxyutil ./internal/util/sessionutil -count=1 -p 1 -tags dynamic,test -gcflags=all=-l
```

All five packages passed. Package-scoped golangci-lint reports `0 issues`, and the required Go-only Milvus build succeeds. A new Team2 review has also been requested against the latest HEAD.

bigsheeper:
🤖 **team2 automated review** (single agent · not cross-verified, for reference) · by `yihao`

🤖 **team2 automated re-review** (single agent · not cross-verified, for reference) · by `yihao`

Re-reviewed at head `0a98e748df`, after reading the previous team2 review and the author's item-by-item disposition. The accepted unlimited-retry policy and the unchanged `classifyStorageV3Err` are treated as settled product decisions and are not re-litigated here; this pass verified the six follow-up fixes for concrete correctness bugs.

**What checks out** (verified against the code, not just the diff):

- **Storage V3 post-commit guard**: `stopStorageV3RetryAfterCommit` (`internal/flushcommon/syncmgr/pack_writer_v3.go:258`) works end-to-end. The `retry.Unrecoverable` marker survives `classifyStorageV3Err` (which returns transient errors unchanged, marker intact), and `retry.Do` checks `IsRecoverable` (`pkg/util/retry/retry.go:60`) *before* the always-true `RetryErr` override (`retry.go:77`), so the production opts from `write_buffer.go:1504-1505` cannot restart Phase 1 after a committed manifest. Test pins it.
- **Session rewatch revision**: `handleReWatch` now watches from `revision+1` (`session_util.go:996`); the embedded-etcd test verifies no snapshot replay and delivery of the next revision.
- **Growing-source terminal path**: the layout-mismatch branch now invokes `wb.errHandler` after rollback and outside `wb.mut` (`write_buffer.go:1026-1028`); `errHandler` is always non-nil (default panics, `options.go:49`), so the non-retryable branch has a real terminal action again, and the segcore layout-mismatch error genuinely reaches this callback (the growing-source flush at `growing_source.go:692` is single-attempt, not wrapped in the infinite retry).
- **Cancellation of the unlimited rewatch loops**: `context.Canceled` is excluded from the predicate, both call sites check `ctx.Err()`/`closeCh` before panicking, and `rewatchProxy`'s closeCh-monitor goroutine correctly interrupts the loop (test covers it).
- **Terminal/no-options tests**: the one-attempt/one-failure-callback V3 test and the `allocOneWithRetry` passthrough test are present and assert the right things.

Two concrete correctness bugs remain in the new recovery code, both inside the follow-up's own mechanism (not restatements of the accepted policy):

---

### high — 1. `IsRetriableEtcdRequestErr` cannot see the `Unavailable` code of etcd's own error sentinels, so the exact leader-election failures this predicate exists for still abort recovery and crash the process

`pkg/util/etcd/etcd_util.go:93-98`:

```go
switch status.Code(err) {
case codes.Unavailable, codes.DeadlineExceeded:
    return true
```

clientv3 converts etcd-server-originated errors into `rpctypes.EtcdError` sentinels before returning them (`ContextError` → `rpctypes.Error`, client v3.5.23 `client.go:578-598`). `EtcdError` implements only `Code()` and `Error()` — **not** `GRPCStatus()` (api v3.5.23 `v3rpc/rpctypes/error.go:229-242`) — and grpc v1.80.0 `status.FromError` recognizes only `GRPCStatus()` implementers, so `status.Code()` returns `codes.Unknown` for `rpctypes.ErrTimeout` ("etcdserver: request timed out"), `ErrNoLeader`, `ErrTimeoutDueToLeaderFail`, `ErrTimeoutDueToConnectionLost`, and `ErrTooManyRequests`, even though their `EtcdError.Code()` is `Unavailable`/`ResourceExhausted`.

Failure scenario: a session watch stream breaks with `ErrLeaderChanged` → `handleWatchErr` → `retryRewatch` → `GetSessions` (`session_util.go:987`, no per-request timeout) — while the election is still in progress, etcd responds with `etcdserver: request timed out` (`rpctypes.ErrTimeout`). The predicate returns false, `retry.Do` aborts on the **first** attempt, `closeEventCh()` runs, and `handleWatchResponse` panics (`session_util.go:898`) — or the coordinator takes the `os.Exit(ExitCodeEtcd)` path. The proxy watcher hits the same hole for any sentinel returned before its 10s child deadline (`proxy_watcher.go:169` panic). This is exactly the "election still in progress" window the previous review's finding 4 described; the fix covers raw gRPC status errors (transport-level) and `context.DeadlineExceeded`, but not the sentinel-shaped errors etcd actually returns when it is reachable but leaderless. The new unit tests pass because they use `status.Error(codes.Unavailable, ...)`, which is not the shape clientv3 delivers for server-originated failures — note `rpctypes.ErrPermissionDenied` in the test is rejected for the same accidental reason (all `EtcdError`s map to `Unknown`), not because permission-denied is specifically excluded.

**Fix**: also inspect the sentinel's own code:

```go
var etcdErr rpctypes.EtcdError
if errors.As(err, &etcdErr) {
    switch etcdErr.Code() {
    case codes.Unavailable, codes.DeadlineExceeded, codes.ResourceExhausted:
        return true
    }
}
```

(or enumerate `ErrTimeout`/`ErrNoLeader`/`ErrTimeoutDueToLeaderFail`/`ErrTimeoutDueToConnectionLost`/`ErrTooManyRequests` via `errors.Is`). Add test cases using the `rpctypes` sentinels rather than raw status errors.

### medium — 2. `sessionWatcher.Stop()` during an in-flight rewatch panics the coordinator: the graceful-shutdown guard checks `s.ctx`, but the retry loop now runs on `w.ctx`

The follow-up moved the rewatch retry onto the watcher context (`retryRewatch` → `watchContext()` = `w.ctx`, `session_util.go:967-979`), which `Stop()` cancels (`session_util.go:820-823`). But the panic guard in `handleWatchResponse` was not updated:

```go
// session_util.go:893-898
if w.s.ctx.Err() != nil { ... return }
mlog.Error(...)
panic(err)
```

DataCoord stops its session watchers **before** stopping the session (`internal/datacoord/server.go:1093-1104`), so at `Stop()` time `w.ctx` is canceled while `s.ctx` is still alive. If the watcher goroutine is inside `handleWatchErr` → `retryRewatch` at that moment, `retry.Do` returns `context.Canceled` (or `lastErr` — e.g. the last transient etcd error — when canceled mid-sleep, `retry.go:120`), `handleWatchErr` propagates it, the `w.s.ctx.Err()` guard does not fire, and the process panics during a normal graceful shutdown, with `Stop()` blocked in `wg.Wait()`. The window is not small: with etcd unavailable, the `AttemptAlways` loop spins in `retryRewatch` indefinitely, so any operator-initiated stop during an etcd outage lands exactly here. This is new in the follow-up — previously the retry ran on `s.ctx` and `Stop()` did not interrupt it. Note the closed-channel branch already gets this right (`session_util.go:800,805` check both contexts); only the error-response path misses it.

**Fix**: treat watcher-context cancellation as graceful teardown too — e.g. in `handleWatchResponse` check `w.watchContext().Err() != nil` alongside `w.s.ctx.Err()`, or have `handleWatchErr` return nil when `w.ctx` is done, mirroring the `!ok` branch.

---

No other substantive issues found in the follow-up. With finding 1 fixed, the leader-change recovery actually covers its target window; finding 2 is a one-line guard. Everything else in the disposition's "addressed" list is verified as implemented correctly.

chyezh:
已按本次 Team2 review 的两项结论完成修复，见 commit `9a6b7c13d3`：

1. `IsRetriableEtcdRequestErr` 现在通过 `errors.As` 识别 `rpctypes.EtcdError` 自身的 code，覆盖 `Unavailable`、`DeadlineExceeded`、`ResourceExhausted`；测试新增了 `ErrNoLeader`、各类 timeout、`ErrTooManyRequests` 及 wrapped sentinel。权限错误和 cancellation 仍保持不可重试。
2. `sessionWatcher.handleWatchResponse` 现在同时检查 watcher context 与 session context；新增并发测试在 rewatch 正在执行时调用 `Stop()`，验证正常退出且不 panic。

验证通过：
- `go test ./pkg/util/etcd -run TestIsRetriableEtcdRequestErr -count=1 -tags dynamic,test -gcflags=all=-l`
- `go test ./internal/util/sessionutil -run TestWatcherHandleWatchResp -count=1 -tags dynamic,test -gcflags=all=-l`
- `go test ./pkg/util/etcd ./internal/util/sessionutil -count=1 -p 1 -tags dynamic,test -gcflags=all=-l`
- `./bin/golangci-lint run ./pkg/util/etcd ./internal/util/sessionutil`（0 issues）
- `milvus-build --only-go` 成功

本轮未调整 `classifyStorageV3Err`，也未改变已确认的 recoverable error 无限重试策略。请基于最新 HEAD `9a6b7c13d3` 复审。

bigsheeper:
🤖 **team2 automated review** (single agent · not cross-verified, for reference) · by `yihao`

🤖 **team2 automated re-review** (single agent · not cross-verified, for reference) · by `yihao`

Re-reviewed at head `9a6b7c13d3`, scoped to the two fixes from the previous round (per the author's disposition). The settled decisions — unlimited retry for recoverable errors, `classifyStorageV3Err` unchanged — were not re-examined. **Both fixes are correct as implemented; no critical/high/medium issues found.** Verification details, then two nits.

**Fix 1 — `rpctypes.EtcdError` classification (`pkg/util/etcd/etcd_util.go:96-104`): verified correct and complete.**

- The `errors.As` branch runs after the `IsRetriableWatchErr` / `context.DeadlineExceeded` checks and before the `status.Code` fallback, so both error shapes clientv3's `toErr` can deliver (server-originated `EtcdError` values vs. raw transport-level status errors) are now covered. Checked against etcd api v3.5.23 (the version in both `go.mod` files): `ErrNoLeader`, `ErrTimeout`, `ErrTimeoutDueToLeaderFail`, `ErrTimeoutDueToConnectionLost`, `ErrTimeoutWaitAppliedIndex` all carry `codes.Unavailable` and `ErrTooManyRequests` carries `codes.ResourceExhausted` — all now retriable, which closes the leader-election/timeout window from the previous round's finding 1. `rpctypes.Error()` returns an `EtcdError` **value** with a value-receiver `Code()`, so `errors.As(err, &etcdErr)` matches, including through `errors.Wrap` chains (the wrapped-sentinel test pins this).
- Non-retriable exclusions hold for the right reasons now: `ErrPermissionDenied` (`codes.PermissionDenied`) hits the `default: return false` inside the `EtcdError` branch — a deliberate rejection, no longer the accidental `codes.Unknown` fallthrough. `context.Canceled` is not an `EtcdError` and maps to `codes.Canceled` → false. Notably, etcd's own `ErrCanceled` sentinel ("etcdserver: request canceled", `codes.Canceled`) is also correctly rejected by the new branch — the early-return `default` prevents any accidental fallthrough.
- Both call sites (`internal/util/sessionutil/session_util.go:971` and `internal/util/proxyutil/proxy_watcher.go:207`) share the predicate, so the proxy watcher's identical hole is closed by the same change.

**Fix 2 — `sessionWatcher.Stop()` during in-flight rewatch (`internal/util/sessionutil/session_util.go:893`): verified correct.**

- Traced every way `retry.Do` can exit after `Stop()` cancels `w.ctx` (`pkg/util/retry/retry.go`): canceled mid-sleep returns `lastErr` (line 120, e.g. the transient `ErrTimeout`); the attempt itself failing with `context.Canceled` from the canceled `GetSessions` returns that error via the non-recoverable or predicate path. In every case the returned error is non-nil and `w.watchContext().Err() != nil` is true, so `handleWatchResponse` takes the quiet-return branch instead of `panic` — including the previously fatal DataCoord ordering where watchers stop before the session and `s.ctx` is still alive. The `nil`-error case only occurs when `handleReWatch` succeeded, which needs no guard.
- No shutdown hang: after the quiet return, the `start()` loop re-enters `select`, sees `ctx.Done()`, and exits, unblocking `Stop()`'s `wg.Wait()`. If the rewatch instead won the race, the fresh watch opened on the canceled `w.ctx` closes immediately and the `!ok` branch (which already checks both contexts) exits. The `watchContext()` nil-fallback keeps the old panic behavior intact for live sessions (the `err canceled` test still asserts the panic).
- The new concurrency test blocks a real in-flight attempt, calls `Stop()` mid-attempt, and asserts no panic with a timeout — it exercises exactly the reported race.

**Nits** (no action required to merge):

- **nit** — `internal/util/sessionutil/session_util.go:955-958`: the comment "Recoverable request failures keep retrying until the session context is canceled" is now stale — the retry runs on the watcher context (`retryRewatch` → `watchContext()`), which `Stop()` cancels independently of the session. Worth updating so the next reader doesn't rediscover fix 2.
- **nit** — `pkg/util/etcd/etcd_util.go:107`: adding `codes.ResourceExhausted` to the **raw** `status.Code` branch (beyond the `EtcdError` branch, which is the one that matters for `ErrTooManyRequests`) also classifies client-generated ResourceExhausted errors — chiefly gRPC's "message larger than max" — as infinitely retriable, and those are permanent for an identical request. In practice this is unreachable here (Milvus doesn't set `MaxCallRecvMsgSize`, clientv3's recv default is `MaxInt32`, and session payloads are tiny), so this is a note for the comment/future readers, not a requested change.

Everything the author's disposition claims for this commit is verified as implemented. LGTM.

bigsheeper:
🤖 **team2 automated review** (single agent · not cross-verified, for reference) · by `yihao`

🤖 **team2 automated review** (single agent · not cross-verified, for reference) · by `yihao`

Re-reviewed at head `9a6b7c13d3`, this time over the **entire PR diff vs merge-base** (17 files). The two fixes verified in the previous round (`IsRetriableEtcdRequestErr` EtcdError classification; `sessionWatcher.Stop()` during in-flight rewatch) were not re-examined, and the settled policies (unlimited retry for recoverable errors, `classifyStorageV3Err` scope) were taken as given. Within those constraints I traced the remaining new surfaces end-to-end: the StorageV3 fresh-writer-per-attempt loop in `SyncTask.Run`, `stopStorageV3RetryAfterCommit`, `allocOneWithRetry`, the growing-source error rerouting, both closed-watch-channel recovery paths, and the `WithRev(revision+1)` change. No critical or high issues found. Two mediums and two lows below.

**medium** — `internal/flushcommon/writebuffer/write_buffer.go:988-1029` + `isGrowingSourceLayoutMismatch` (`write_buffer.go:198-205`): deterministic growing-source integrity failures now livelock instead of failing fast. The PR removes `WithFailureCallback(wb.errHandler)` from `getGrowingSourceSyncTask` and routes only `terminalGrowingSourceErr` to the error handler, but the terminal set is a two-string allowlist ("Column count mismatch" / "Column group size mismatch"). Every other deterministic `ErrDataIntegrity` failure from `GrowingSourceSyncTask.Run` — "flush row count mismatch", "returned empty manifest", "missing insert binlog summary" — takes the `scheduleGrowingSourceRetryLocked()` branch, and `failureCount` feeds only a metric with no escalation threshold. Failure scenario: the in-memory growing source's offset accounting diverges (a bug, but exactly what these checks exist to catch) → every resync re-flushes from the same corrupted in-memory source → identical mismatch forever → the segment never flushes, the channel checkpoint stops advancing, and WAL retention grows unbounded, with only a warn log and a counter. Under the old wiring the failure callback panicked the DataNode, and the WAL replay rebuilt the growing source — a crash, but a self-healing one. The PR's own test pins the new behavior (`l0_write_buffer_test.go:1102` asserts `errorHandlerCalls == 0` for the row-count-mismatch case). Suggestion: treat `merr.ErrDataIntegrity`-classed sync errors as terminal alongside the layout-mismatch strings, or escalate to `wb.errHandler` after N consecutive failures at the same `targetOffset`.

**medium** — `internal/util/sessionutil/session_util.go:985-996` (`handleReWatch`): the rewatch snapshot bypasses the version-range filter. The event path filters every session through `w.validate` (session_util.go:914, 930), but `handleReWatch` feeds the **unfiltered** `GetSessions` result straight into `w.rewatch(sessions)`. For DataCoord's version-ranged watchers (`WatchServicesWithVersionRange` at `datacoord/server.go:530,539`), a rewatch replays out-of-range sessions into `rewatchDataNodes`/`rewatchQueryNodes`, registering nodes the range deliberately excludes (`LegacyVersionWithoutRPCWatch`). The gap is pre-existing (the ErrCompacted path always had it), but this PR turns rewatch from a rare event into a routine one: `ErrLeaderChanged` is now retriable and the closed-watch-channel path also rewatches — and etcd leader transfers are common precisely during maintenance windows, which is when rolling upgrades (mixed-version clusters) happen. Failure scenario: rolling upgrade in progress, etcd leader transfer fires, DataCoord's watcher rewatches and registers a legacy DataNode as a normal RPC-watch node. Suggestion: filter the snapshot with `w.validate` before invoking `w.rewatch` (one line: drop non-validating entries from `sessions`).

**low** — `internal/flushcommon/pipeline/data_sync_service.go:326-328`: `retry.RetryErr(func(error) bool { return true })` is a no-op for `BrokerMetaWriter`. Both `UpdateSync` and `UpdateGrowingSourceSync` run under `retry.Handle`, where the fn's `shouldRetry` return governs and the `RetryErr` predicate is only a second gate that an always-true predicate never engages — unlike `retry.Do`, `retry.Handle` has no InputError short-circuit to disable (retry.go:180-186). The option appears copy-pasted from the `retry.Do` sites in write_buffer.go where it is load-bearing. It is harmless today but misleads the next reader into thinking it changes classification; the actual abort set is the fn's own (canceled/timeout, segment-not-found, channel-not-found). Suggestion: drop the predicate here, or replace it with a comment stating the fn's `shouldRetry` is the real policy. Related note, no change requested given the settled retry-forever policy: the resync path calls `wb.syncSegments(context.Background(), ...)` (write_buffer.go:1031), so a SaveBinlogPaths rejection outside the four handled cases now retries with no cancellation path at all.

**low** — `internal/flushcommon/syncmgr/task.go:157-176`: each failed StorageV3 attempt orphans that attempt's phase-1 output. The fresh-writer-per-attempt design is correct for FFI safety (verified: the outer `retry.Do` cannot re-run phase 1 after a commit, because phase-2 errors leave the inner loop only as `Unrecoverable` or via context cancellation, both of which stop the outer loop), but every retried attempt writes a full batch of data files and stat blobs under freshly allocated log IDs that no manifest ever references. They are reclaimed only when the segment's whole base path is GC'd at drop (`garbage_collector.go:1049`), so a long-lived segment hitting a lasting fault at a late phase-1 step under `AttemptAlways` accumulates orphan bytes without bound and without observability. Suggestion: log cumulative discarded bytes per retry (the writer already tracks `sizeWritten`), so operators can see the bloat.

Verified correct in this pass (no action needed): the `WithRev(revision+1)` change in `handleReWatch` fixes a real duplicate-replay off-by-one — `GetSessions` returns the snapshot's header revision and initial watches already pass `rev+1` (`querycoordv2/server.go:643`), so `+1` aligns rewatch with the established convention with no event gap, and the new revision test pins it; the closed-channel recovery in both watchers preserves fail-fast for non-retriable failures (session watcher closes `eventCh` → consumers `os.Exit`, `querycoordv2/server.go:652-661`; proxy watcher panics) and shuts down cleanly through the `closeCh`/watcher-context interlocks; `allocOneWithRetry` is a faithful no-op when no options are passed, keeping non-DataNode writer users unchanged; `stopStorageV3RetryAfterCommit`'s marker survives the outer `classifyStorageV3Err` re-classification (the `errUnrecoverable` sentinel stays in the chain), so a post-commit bookkeeping failure can never restart phase 1; and `retry.Handle`'s `shouldRetry=false` on canceled/timeout still bounds the unbounded meta-writer retry during teardown. The stale comment at `session_util.go:955` ("until the session context is canceled") noted last round is still present.

PR #52588, opened 2026-08-17
enhance: prepare QueryView resources on QueryNode

chyezh:
issue: #40451

- QueryView resource lifecycle: pin exact-version collection metadata and share revisioned physical segments across view generations
- QueryNode readiness: load detached sealed segments and gate Ready on TransformLog registration and catch-up
- segment loading: expose idempotent resource reservations and schema-safe index metadata refresh primitives
- design boundaries: define thin scheduler, QueryCoord metadata-stream, WAL, and handler adapters


czs007:

Reviewed the full diff against the merge-base (3 commits: `d49c349`, `b46958e`, `f8f4dcc`). The layering is clean — the core in `qnview` depends only on injected interfaces, and `qvresource` holds all the `segments`/segcore coupling — and the design doc matches the intended architecture. Nothing is wired into production yet (`NewQueryViewSegmentReadinessManager` / `NewViewScopedPhysicalSegmentManager` / `NewQueryViewPhysicalSegmentLoader` / `NewQueryViewCollectionRuntimeManager` have no non-test callers), so none of this is live. That said, there are two crash-class defects and several ownership/lifecycle holes that should be fixed before the adapter branches land on top of this.

---

## Critical

### 1. Nil-pointer panic in `activateAcquire` when a shared segment fails between `recordPendingAcquire` and activation

`internal/querynodev2/qnview/queryview_segment_readiness_manager.go:235-238`

```go
for segmentID, partitionID := range ref.segments {
    state := m.segments[segmentID]
    waiter := newTransformWaiter(req, partitionID, segmentID)
    if state.state == transformSegmentLoaded {   // state may be nil
```

`recordPendingAcquire` creates `m.segments[segmentID]`, then `continueAcquire` performs a `DescribeCollection` + `GetQueryViewLoadInfo` round-trip (`:83`) before calling `activateAcquire`. `m.segments` is keyed by segment ID **globally**, and `failSegment` unconditionally does `delete(m.segments, segmentID)` (`:356`).

Failure scenario: views V1 and V2 both reference segment S. V2's `Acquire` records the pending state, then blocks on the QueryCoord metadata RPC. During that window V1's catch-up for S fails (`registerAndCatchup:302` → `failSegment`), which deletes `m.segments[S]`. V2's `activateAcquire` then dereferences a nil `state` → panic on the scheduler goroutine → QueryNode crash. Every other reader of `m.segments` in this file (`detachViewLocked:445`, `failView:384`, `notifyUnrecoverable:402`, `markPhysicalLoaded:270`, `storeRegistration:319`) does nil-check; this is the only site that does not.

**Fix:** nil-check and treat a missing state as "this segment was failed concurrently" — either recreate it in `transformSegmentLoading`, or abort activation and report unrecoverable for this view. Note also that `failSegment` leaves `ref.segments` of surviving views pointing at a segment ID that no longer has state, so simply skipping is not enough — those views will never be reported ready.

### 2. Concurrent map read/write on `physicalViewRef.segments` → Go runtime fatal error

`internal/querynodev2/qnview/view_scoped_physical_segment_manager.go:129-136` vs `:480-482`

```go
    m.mu.Unlock()                                       // :129
    for _, request := range subscriptions { ... }
    if len(viewRef.segments) == 0 || len(loaded) == len(viewRef.segments) {   // :134 — unlocked read
```
```go
    for _, view := range m.views {
        delete(view.segments, segmentID)                // :481 — write under m.mu
    }
```

`Acquire` reads `viewRef.segments` after releasing `m.mu`, while `ResetSegment` mutates the same map under `m.mu`. `ResetSegment` runs on a catch-up worker goroutine (`failSegment:365-367`) and `Acquire` runs on a scheduler goroutine, so they overlap. This is not a benign data race: `fatal error: concurrent map read and map write` is unrecoverable and kills the process.

**Fix:** capture `segmentCount := len(viewRef.segments)` inside the critical section at `:128` and use the local copy afterwards.

---

## High

### 3. A shared physical segment binds an arbitrary view's `TransformStartAfterTimeTick` — can silently skip deletes

`view_scoped_physical_segment_manager.go:113-115`, `:542-550`, `segment_scheduler_task.go:65-67`

```go
state.requests[req.Key] = physicalLoadRequest{
    collection: req.Collection,
    startAfter: req.Meta.GetTransformStartAfterTimetick(),
}
...
func (s *physicalSegmentState) anyRequest() (physicalLoadRequest, bool) {
    for key := range s.refs {           // random map iteration order
        request, ok := s.requests[key]
        if ok { return request, true }
```
```go
if t.TransformStartAfterTimeTick > 0 {
    segment = &transformStartSegment{TransformSegment: segment, startAfter: t.TransformStartAfterTimeTick}
}
```

`transform_start_after_timetick` is a **per-QueryView** field (`pkg/proto/view.proto:50`), but the physical segment is shared by segment ID across views (doc §6). `anyRequest()` picks whichever view happens to come out of the `refs` map first, and the resulting `startAfter` is baked into the wrapper permanently — it is never recomputed when refs are added or removed.

Failure scenario: view V1 (`start_after = T1`) and a newer view V2 (`start_after = T2`, `T2 > T1`) share segment S. If the load is triggered while V2's request is selected, the segment is registered with the TransformLog at `T2`, so transforms in `(T1, T2]` are never replayed for V1 — V1 serves entities that were deleted in that window. The reverse order merely wastes work. This directly undermines the readiness invariant in doc §2 ("Reporting Ready before step 6 can expose a sealed segment that is missing deletes...").

**Fix:** take `min(startAfter)` over all live requests, and recompute on every `Acquire`; or key physical segment identity by `(segmentID, startAfter)`. At minimum, make it deterministic and document why min is safe.

### 4. A permanently failing metadata update wedges the segment forever, with no notification

`segment_scheduler_task.go:98-107` and `view_scoped_physical_segment_manager.go:446-454`

```go
if err := t.update(ctx); err != nil {
    if ctx.Err() != nil { t.fail(ctx.Err()); return nil }
    // A non-nil task error asks the injected QueryNode scheduler to delay and retry
    return err
}
```

`OnFailed` (→ `cancelUpdate`) is invoked **only** on context cancellation. A non-context error (missing index file, corrupt manifest, segcore reopen rejection) is retried by the scheduler forever. Meanwhile `state.updating` stays `true`, so `ApplyLoadInfoSnapshot:216-219` funnels every subsequent revision into `state.pending` where it is repeatedly overwritten and never applied. No view is ever told the segment is unrecoverable, and the segment stays pinned at the stale revision indefinitely.

Contrast with the load path, which explicitly distinguishes retryable resource pressure from terminal failure (`segment_scheduler_task.go:30-38`). The update path needs the same classification: retry only `merr.ErrSegmentRequestResourceFailed`-class errors, and after that (or after a bounded attempt count) call `OnFailed` with the real error and surface it through `OnSegmentUnrecoverable`.

### 5. Double `Unregister()` on the same `TransformRegistration`

`queryview_segment_readiness_manager.go:299-303` vs `:450-455` / `:478-485`, and `:359-364`

```go
if err := registration.WaitCatchup(ctx); err != nil {
    cancel()
    registration.Unregister()      // :301
    m.failSegment(segment.ID(), err)
    return
}
```

`detachViewLocked` collects `state.catchupCancel` and `state.registration` and deletes the state; `unregister()` then cancels the catch-up context **and** calls `registration.Unregister()`. The cancel causes the in-flight `WaitCatchup` to return an error, and `registerAndCatchup` unregisters the very same registration a second time. The same double-call happens via `failSegment` (`:361` cancel, `:363` unregister). `TransformRegistration.Unregister()` (`segment_resource_types.go:45-48`) carries no idempotency contract, so a WAL-side implementation that decrements a refcount will corrupt it.

**Fix:** have `registerAndCatchup` re-check under `m.mu` whether the state still owns this registration before unregistering (mirroring the `storeRegistration` pattern), and/or document `Unregister` as required-idempotent in the interface.

---

## Medium

### 6. Metadata updates bypass resource admission control

`segment_scheduler_task.go:86-89` / `segment_resource_types.go:245-256`

`SegmentLoadTask` reserves through `SegmentResourceEstimator` (`:79-84`), but `SegmentUpdateTask` has no estimator field and never reserves. `classifySegmentUpdate:151-156` returns `Reopen|LoadIndex` for **any** revision change, and `Update` → `ReopenSegment` → `segment.Reopen(ctx, info)` can pull in a full new index or replacement binlogs. On a node under memory pressure this loads unbounded bytes with zero accounting — exactly the OOM protection the loader's `requestResource` exists to provide. Doc §6 claims "Loading resource reservations are explicit and idempotently released on all success and failure paths," which is only true of the load path.

### 7. `cancelUpdate` drops a pending revision permanently

`view_scoped_physical_segment_manager.go:446-454`

`startUpdateLocked` clears `state.pending` (`:388`). If a newer revision arrives while the update is in flight it is stored in `pending`, but `cancelUpdate` only resets `updating`/`updateHandle` — it never checks `state.pending`. Unlike `completeUpdate:436-438`, there is no drain. That revision is then lost until yet another, higher revision happens to arrive. `completeUpdate` should be mirrored here.

### 8. A snapshot is acknowledged even when the manager discards it

`segment_load_info_stream.go:13`, `:180-187` and `view_scoped_physical_segment_manager.go:207-219`, `:530-538`

The stream documents (and doc §5 mandates) "acknowledge a revision only after the local handler accepts it," and `consume` advances `s.revision` on a nil return from `Handle`. But `physicalSegmentLoadInfoHandler.Handle` always returns nil, and `ApplyLoadInfoSnapshot` silently drops snapshots when the state is missing, refs are empty, or the collection ID mismatches (`:207-210`). After a reconnect the stream resumes *after* the dropped revision, so it is never re-delivered. Either return a non-nil error from `Handle` when the snapshot is not accepted, or drop the acknowledgement claim from both the code comment and the design doc — right now code and doc contradict each other.

### 9. Catch-up dispatch can block a scheduler goroutine, and the manager has no shutdown

`queryview_segment_readiness_manager.go:43-49`, `:261`

```go
catchups: make(chan TransformSegment, 1024)
for range transformCatchupWorkerCount { go m.catchupWorker() }
...
m.catchups <- segment    // :261, unbounded blocking send
```

`onPhysicalLoaded` is invoked synchronously from `completeLoad` on a scheduler goroutine. With 4 fixed workers each blocked in `WaitCatchup`, a burst of loads past the buffer blocks the sender — and if that sender occupies a scheduler slot that catch-up progress depends on, this self-deadlocks. Separately, `catchups` is never closed and `QueryViewSegmentReadinessManager` has no `Close()` at all, so the 4 workers leak per instance (compare `ReconnectingSegmentLoadInfoStream.Close`, which is implemented). Please add a lifecycle method before wiring.

### 10. Subscribing to an already-closed stream silently disables the segment

`segment_load_info_stream.go:60-68` returns a fully inert subscription (no `owner`, `done` pre-closed, `run` never started). `subscribe` (`view_scoped_physical_segment_manager.go:511-520`) only checks `subscription == nil`, so it stores the dead handle, `state.subscription != nil` forever suppresses re-subscription, no snapshot ever arrives, and the segment never loads — with no error and no log. Return `nil` (or expose the closed state via `Error()` and check it) so the caller can report unrecoverable.

### 11. Already-loaded segments are dropped on a partial acquire

`view_scoped_physical_segment_manager.go:88`, `:117-119`, `:134-136`

```go
if len(loaded) == len(viewRef.segments) {
    m.submitLoaded(req.OnLoaded, loaded)
}
```

`loaded` is collected per already-resident segment, but delivered only when **all** of the view's segments are resident. If a view acquires {A resident, B loading}, A is silently discarded and only B is later delivered via `completeLoad:317`. Today this is masked because `QueryViewSegmentReadinessManager.activateAcquire` independently short-circuits on `transformSegmentLoaded`, but the physical manager's own contract is broken and any second consumer of `OnLoaded` will lose segments. Deliver `loaded` unconditionally when non-empty.

### 12. Two conflicting notions of "transform start after" on the same segment

`qvresource/physical_segment.go:27-46` vs `qnview/segment_scheduler_task.go:172-183`

`queryViewTransformSegment` initialises `applied = startAfter` where `startAfter` comes from the **load info** (`transformStartAfter` → delta/start position, `physical_segment_loader.go:224-229`), while `transformStartSegment` overrides only the **getter** with the **view meta** tick. So `TransformStartAfterTimeTick()` and `AppliedTransformTimeTick()` report values derived from different sources, and `WaitTransformApplied` for a tick between them returns immediately even though nothing was replayed. `ApplyDelete`/`MarkTransformApplied` also live on the inner object, bypassing the wrapper entirely. Please unify these into one value with a documented meaning.

### 13. Cross-view collection guard / index-meta staleness

`view_scoped_physical_segment_manager.go:113`, `segment_scheduler_task.go:118`, `qvresource/collection_runtime.go:127-129`

`anyRequest()` also selects an arbitrary view's `CollectionRuntime` guard, and `updateCollectionIndexMeta` calls `g.collection.UpdateIndexMeta(ComposeIndexMeta(indexes, g.schema))` using that guard's **snapshotted** schema. If the selected view is older than another live view, index meta for the shared native collection is recomposed against a stale schema. The guard pointer itself stays valid only because another view holds a `PutOrRef` on the same collection — that is an undocumented invariant worth asserting.

---

## Low / nit

- **`failLoad(submission, nil)`** — `view_scoped_physical_segment_manager.go:293-296` passes a nil error when `OnLoaded` yields a nil segment; it surfaces as `mlog.Err(nil)` (`queryview_segment_readiness_manager.go:369`) and `OnSegmentUnrecoverable(id, nil)`. Use a concrete `merr` error.
- **Load task cancelled before execution leaves `state.loading == true`** — `segment_scheduler_task.go:25-27` returns nil with no callback. Reachable when the scheduler context is cancelled (node shutdown) while the manager state survives; the segment is then permanently stuck "loading". An `OnFinished`-style unconditional state reset would close this.
- **`queryViewTransformSegment.Release`** (`qvresource/physical_segment.go:124-127`) does not close outstanding `waiters`, so any `WaitTransformApplied` caller passing a non-cancellable context blocks forever after release.
- **Dead `SegmentUpdateLoadIndex` branch** — `classifySegmentUpdate:155` always returns `Reopen|LoadIndex`, and `physical_segment_loader.go:104-112` returns on the `Reopen` bit first, so the `LoadIndex` branch is unreachable. Relatedly, `segmentLoader.LoadIndex` (`segment_loader.go:1195-1202`) is literally `segment.Reopen` and ignores its `version int64` parameter — either drop the action bit and the method, or give them distinct behaviour.
- **`SegmentUpdateAction` bit values** — `segment_resource_types.go:207-211`: because `1 << iota` starts at `iota == 1`, `Reopen == 2` and `LoadIndex == 4`, leaving bit 0 unused. Works, but is easy to misread; prefer explicit `1 << 0` / `1 << 1`.
- **`LoadDeltaLogsWithoutResource` reaches back into the global collection manager** — `segment_loader.go:1618-1620` → `loadDeltalogs:1367-1369` does `loader.manager.Collection.Get(segment.Collection())` and dereferences the result with no nil check. This only works if the integrator passes the *same* `CollectionManager` to `NewQueryViewCollectionRuntimeManager` and to the `segmentLoader`. Worth an explicit comment on `NewQueryViewPhysicalSegmentLoader`, since a mismatch is a segfault-adjacent nil deref rather than an error.
- **Missing collection check on ref reuse** — `view_scoped_physical_segment_manager.go:104-115` attaches a view to an existing `physicalSegmentState` without verifying `state.collectionID == req.Meta.GetCollectionId()`. `ApplyLoadInfoSnapshot:207` then filters on the *original* collection ID, so a mismatched view would hang silently instead of erroring.
- **Unreachable branch** — `queryview_segment_readiness_manager.go:113`: `physicalIDs` is appended for every entry of `ref.segments` (`:249`), so `len(physicalIDs) == 0` is only possible when `len(ref.segments) == 0`, already handled at `:107`.
- **Dead method** — `qvresource/physical_segment.go:137-139` `queryViewLocalSegment.Collection()` is not part of `queryViewLoadedSegment` and has no callers.
- **Test coverage gap** — the suite covers reset-cancels-update, revision rollback, and catch-up failure well, but there is no concurrency test (`-race` with overlapping `Acquire` / `failSegment` / `ResetSegment`), which is precisely where issues 1 and 2 live. A small stress test driving N views over a shared segment set with injected failures would catch both.

czs007:

🤖 **team2 automated review** (single agent · not independently verified — treat findings as leads to confirm) · by `zack`

Re-reviewed at head `859d49c` (`40451-view_resource_prep`). Note the branch was force-pushed since my [earlier comment](https://github.com/milvus-io/milvus/pull/52588#issuecomment-5317396475): it is now one squashed commit with a much larger scope (101 files; `querynodev2/**` plus `internal/views/*`, `transformlogbuffer`, proto). Good news first — three things from that round are fixed: `activateAcquire` now nil-checks and recreates the segment state (`queryview_segment_readiness_manager.go:255-266`), the unlocked read of `viewRef.segments` is gone (`ResetSegment` and all readers are under `m.mu`), and `TransformRegistration.Unregister` is now idempotent via `sync.Once` (`transformlogbuffer/buffer.go:706`). The `SegmentLoadInfoStream`/revision plumbing is also a real improvement.

The proto changes (`view.proto`, `streaming.proto`, `internal.proto`) look backward-compatible: I found no removed or renumbered fields — `StreamingNodeAssignment.channels = 2` is retained, `SubSearchRequest.skip = 16` is a fresh tag, and `VCHANNEL_STATE_TOMBSTONED = 3` only takes a tag that was a commented-out reservation. Layering is still clean.

The findings below are mostly about the newly-added code. The first two are the ones I would block on.

---

## Critical

### 1. The QueryView lease is released before the query executes — the collection runtime can be freed under a running search

`internal/querynodev2/qnview/query_task_provider.go:25` and `:61`

```go
lease, err := h.AcquireReadyView(ctx, shardID, version)
if err != nil { return nil, err }
defer lease.Release()          // <-- runs before the caller ever executes the tasks
...
handles, err := h.segMgr.AcquireSealedSegmentHandles(ctx, key, view)
...
return NewQNSearchSegmentTasks(tasks), nil
```

`AcquireReadyView` bumps `qnViewEntry.queryRefs` (`query_lease.go:49`), and `qnShardView.consumeReportAndCleanup:210` uses exactly that counter to decide whether a release may proceed. Because the provider drops the lease on return, `queryRefs` is back to 0 by the time `viewquery.Server` calls `scheduler.Search(ctx, tasks)`.

The sealed *segments* survive, because `AcquireSealedSegmentHandles` takes `state.queryRefs` and `detachViewLocked:466-471` skips segments with `queryRefs > 0`. The **collection runtime does not**:

```go
// queryview_segment_readiness_manager.go:478-491
detached := m.detachView(req.Key)
...
m.physical.Release(ReleaseSegments{
    Key: req.Key,
    OnDropped: func() {
        detached.releaseCollection()   // -> qv_collection_runtime.go:133 -> collections.Unref(collectionID, 1)
```

`ViewScopedPhysicalSegmentManager.removeView` returns `onDropped` immediately once `pendingLoads == 0`; nothing waits on segment-level `queryRefs`.

**Failure scenario:** view V is the last live view for collection C on this node. A search acquires its segment handles, the lease is released on return, QueryCoord pushes V to Down/Dropping, `releaseQueryResourceLocked` runs, `Unref(C, 1)` drops the collection to zero and destroys the `CCollection` — while `searchTaskRunner.Search` is executing `segments.SearchSealedSegments` against segments whose segcore handles reference that collection (`viewquery/segment_scope.go:36` takes `readable.Collection()` and passes it into `tasks.NewSearchTask`). That is a use-after-free in segcore, not a Go-level error.

**Fix:** hold the view lease for the whole task lifetime — move `Release` into `QNSearchSegmentTasks.Release()` alongside the handle releases — or make the sealed-segment handle carry a `CollectionRuntimeGuard` ref so `releaseCollection` is gated on `queryRefs` reaching zero the same way segment release is.

### 2. Two independent sources of `transform_start_after_timetick` on the same segment — deletes can be silently skipped

`qvresource/qv_physical_segment_loader.go:46` + `:83-90` vs `qnview/segment_scheduler_task.go:60-64` vs `transformlogbuffer/buffer.go:679-680`

```go
// loader: startAfter comes from the SEGMENT's load info
return newQueryViewTransformSegment(loaded, l.segments, qvSegmentVChannel(info), qvSegmentTransformStartAfter(info)), nil
// -> info.GetDeltaPosition().GetTimestamp(), else info.GetStartPosition().GetTimestamp()
```
```go
// load task: the GETTER is then overridden with the VIEW META's tick
if t.TransformStartAfterTimeTick > 0 {
    segment = &transformStartSegment{TransformSegment: segment, startAfter: t.TransformStartAfterTimeTick}
}
```
```go
// buffer: catch-up floor is taken from the getter, i.e. the view meta tick
startFrom: segment.TransformStartAfterTimeTick(),
drainedTo: segment.TransformStartAfterTimeTick(),
```

`transformStartSegment` overrides only `TransformStartAfterTimeTick()`. The inner `queryViewTransformSegment.applied` is still initialised to the *load-info* tick (`qv_transform_segment.go:35`), and `ApplyTransform`/`markTransformApplied`/`WaitTransformApplied` all operate on that inner value.

**Failure scenario:** the segment's binlogs reflect state up to `T_info`, but the view meta carries `T_view > T_info`. `drainRegistration` replays only entries with `timeTick > T_view` (`buffer.go:392`), so every transform in `(T_info, T_view]` is never applied — the sealed segment serves rows that were deleted in that window, and the segment is still reported Ready. Symmetrically, `WaitTransformApplied(T)` for `T` in that window is satisfied by the first *later* entry, so the MVCC wait does not catch it either.

**Fix:** pick one authoritative value and make it the single field on `queryViewTransformSegment` (recomputed at load and at Reopen), rather than a decorator that overrides one getter. If the view-meta tick must win, the loaded snapshot has to be validated against it and rejected (not silently accepted) when `T_info < T_view`.

---

## High

### 3. `loadRequest()` picks an arbitrary view's request for a shared physical segment

`view_scoped_physical_segment_manager.go:728-736`

```go
func (s *physicalSegmentState) loadRequest() (segmentLoadRequest, bool) {
	for key := range s.refs {        // random map iteration order
		request, ok := s.requests[key]
		if ok { return request, true }
	}
```

`physicalSegmentState` is keyed by segment ID globally, so two views over the same sealed segment share one state. The chosen `segmentLoadRequest` supplies both `transformStartAfterTimeTick` (which feeds issue #2) **and** `collection` — the `CollectionRuntimeGuard` used for `NewSegment`, `updateCollectionIndexMeta`, and the resource estimator. Which view wins is nondeterministic, it is never recomputed when refs are added or removed, and the losing view's guard may be for an older schema snapshot. At minimum take `min(startAfter)` over live requests and pick the guard deterministically (newest view), and document why that is the safe choice.

### 4. A permanently failing segment update retries forever and wedges the segment

`segment_scheduler_task.go:96-101`

```go
if err := t.update(ctx); err != nil {
    if ctx.Err() != nil { t.fail(ctx.Err()); return nil }
    return nodescheduler.ErrDelay          // every non-context error -> retry, forever
}
```

`OnFailed` (→ `failSegmentUpdate:459`) fires **only** on context cancellation. A terminal error — corrupt manifest, missing index file, segcore rejecting `Reopen` — is retried indefinitely. Meanwhile `state.updating` stays `true`, so `recordSegmentUpdateLocked:361-366` funnels every later revision into `state.pendingSnapshot` where it is overwritten and never applied, and no view is ever told the segment is unrecoverable. Contrast the load path, which does classify (`isSegmentResourceInsufficient`, `view_scoped_physical_segment_manager.go:539`). Please apply the same classification here: `ErrDelay` only for resource-pressure-class errors, otherwise `t.fail(err)` and surface it through `OnSegmentUnrecoverable`.

Related, in the same area: the update path has **no resource reservation at all** (`SegmentUpdateTask` has no `estimator` field, cf. `SegmentLoadTask.reserve:73-78`), yet `classifySegmentUpdate:143` returns `Reopen|LoadIndex` for any revision change and `ReopenSegment` can pull in a whole new index. That bypasses the OOM protection the loader's `requestResource` exists to provide.

And `failSegmentUpdate:459-466` clears `updating`/`updateHandle` but never drains `state.pendingSnapshot` — unlike `completeSegmentUpdate:451`. A revision that arrived during a cancelled update is stranded until some *other* snapshot happens to arrive.

### 5. One failure poisons an entire vchannel buffer permanently, and every buffer error is treated as unrecoverable

`transformlogbuffer/buffer.go:633-652`, `:302`, `:598`, `:348-357`; `queryview_segment_readiness_manager.go:108-112`

`vchannelBuffer.fail()` sets `b.err` with no reset or resubscribe path. It is reached from a single subscription error event (`:302`) **and** from a single segment's `ApplyTransform` error on the live path (`:598`). Once set, `acquireLocked:351` rejects every future `Acquire` on that vchannel with `b.err` for as long as any existing guard keeps the buffer in `b.channels`.

On the consumer side:

```go
guard, err := m.buffer.Acquire(ctx, view)
if err != nil {
    cancel()
    m.submitCallback(req.OnUnrecoverable)   // no retry, no classification
    return
}
```

So a transient `AcquireStream` failure (streamingnode not yet ready, connection blip) permanently fails the view — whereas the collection-runtime path right next to it does classify retryability (`continueAcquire:143-146`, `isRetryableCollectionRuntimeError`). Two asks: (a) give `Buffer.Acquire` a retryable/terminal distinction and honour it in the readiness manager, and (b) scope `ApplyTransform` failures to the offending registration instead of the whole vchannel.

Note also that the buffer errors are plain `fmt.Errorf` (`:352`, `:373`, `:88`), so there is nothing for a caller to classify on even if it wanted to — repo convention is `merr` for anything crossing a component boundary.

### 6. `queryViewCollectionRuntimeGuard.UpdateIndexMeta` can nil-deref

`qv_collection_runtime.go:63-67` vs `:128-131`

```go
localCollection := m.collections.Get(meta.GetCollectionId())
var ccollection *segcore.CCollection
if localCollection != nil {                  // Acquire explicitly tolerates nil
    ccollection = localCollection.GetCCollection()
}
```
```go
func (g *queryViewCollectionRuntimeGuard) UpdateIndexMeta(ctx context.Context, indexes []*indexpb.IndexInfo) error {
	indexMeta := segments.ComposeIndexMeta(ctx, indexes, g.schema)
	return g.collection.UpdateIndexMeta(indexMeta)   // assumes non-nil
```

`Acquire` treats a nil `localCollection` as an acceptable outcome (a `PutOrRef` immediately followed by a concurrent `Unref` to zero is enough to produce it), but `UpdateIndexMeta` — called from `updateCollectionIndexMeta` on the load/update task path (`segment_scheduler_task.go:46`, `:110`) — dereferences it unconditionally. That is a nil-pointer panic on a scheduler goroutine. Either fail `Acquire` when `Get` returns nil, or nil-check here and return `merr.WrapErrCollectionNotFound`.

### 7. `viewQueryGrowingSegment` provides no lifetime protection around a borrowed `CSegment`

`segments/viewquery_growing_segment.go:78-80`, `:138`

```go
func (s *viewQueryGrowingSegment) PinIfNotReleased() error { return nil }
func (s *viewQueryGrowingSegment) Unpin() {}
func (s *viewQueryGrowingSegment) Release(context.Context, ...releaseOption) {}
```

This adapter satisfies the full `Segment` interface — including the pin protocol that every other implementation uses to keep segcore memory alive across a query — with no-ops, while `Search`/`Retrieve` call straight into a `segcore.CSegment` it does not own. Any caller that reaches it through the generic `Segment` path gets zero protection: if the owner (the streaming node's growing segment) releases the CSegment mid-query, that is a use-after-free with no Go-visible symptom. Please either make the adapter hold a real ref on the owner, or make the unimplemented pin methods panic/return an error so a generic caller cannot silently use it.

---

## Medium

- **No shutdown path; goroutines and buffers leak.** `Buffer` (`buffer.go:37-40`) starts `drainConcurrency` workers on a never-closed `drainTasks` channel and has no `Close()`; `QueryViewSegmentReadinessManager` (`:47-56`) does the same with `catchupTasks` and also has no `Close()`. Separately, `vchannelBuffer.entries` is only ever trimmed from `releaseGuard` (`trimLocked:551`), so a long-lived view with a fixed `startFrom` pins `retentionStart` and the entry slice grows without bound on a busy vchannel. A retention cap plus an explicit lifecycle method would close both.
- **`NewQueryViewSegmentManager` without a stream factory builds a manager that cannot load anything.** `qv_segment_manager.go:35-40` leaves `segmentLoadInfoStream` nil when the variadic arg is omitted; the manager then takes the direct-load path with a zero `SegmentLoadInfoSnapshot`, and `SegmentLoadTask.loadInfo:68-72` fails every single load with `"query view segment load requires watch snapshot"`. If the stream is mandatory, make it a required parameter rather than a variadic option.
- **Hybrid search acquires a separate segment lease per sub-request.** `views/viewquery/server.go:71-84` calls `executeSearch` (and therefore `AcquireSearchSegmentTasks` → `AcquireReadyView` → `AcquireSealedSegmentHandles`) once per sub-request, concurrently. Nothing pins a single snapshot across them, so a view transition mid-flight can give different sub-searches different segment sets, and the sub-results are then reduced together on the proxy as if they were consistent. Acquire once and share the handles across sub-requests.
- **`Unpin()` was promoted from `RLock` to `Lock` on the query hot path.** `segments/state/load_state_lock.go:89-91`. This is a necessary consequence of replacing `sync.Cond` with a close-and-replace channel (`notifyLocked:216-219` needs exclusive access), but `Unpin` runs once per pinned segment per query, and it now excludes concurrent `PinIf` (`:79`, still `RLock`). On a hot segment this serialises pin/unpin where it previously did not. Worth a benchmark before merge, or keep the refcount notification on a separate lock so `Unpin` can stay a read-lock.
- **Snapshots are acknowledged even when the manager discards them.** `physicalSegmentLoadInfoHandler.Handle` (`view_scoped_physical_segment_manager.go:655-658`) always returns nil, while `applyLoadInfoSnapshot` → `recordSegmentSnapshot:314-317` silently drops the snapshot when the state is missing, refs are empty, or `expected` does not match. If the stream implementation advances its resume point on a nil return (the interface shape strongly suggests it does), the dropped revision is never redelivered after a reconnect. Return a non-nil error when the snapshot is not accepted.
- **Data race on `registration.drainedTo`.** Written by the drain worker outside any lock (`buffer.go:412-414`) and read by `WaitCatchup`'s logging on the readiness-manager goroutine (`:696`, `:706`). Harmless in effect but it will trip `-race` in CI.
- **QueryView metrics are defined but never exposed.** `pkg/metrics/qv_metrics.go:169` `RegisterQV` has no callers anywhere in the repo, so `QVViewStates`, `QVViewTransitionTotal`, `QVShardLoadStates` and the `QVViewStateMaxAgeSeconds` collector are inert — even though `metrics_observer.go:129` does wire the provider. Either register them or note in the PR that the observability wiring lands in a follow-up.

---

## Low / nit

- `viewquery/executor.go:35` and `:43` index `tasks[0]` with no length check. Today `Server.executeSearch:107` short-circuits on empty task lists, but `DirectSegmentTaskExecutor` is exported and `segmentExecutionScope` happily returns `(nil, [], [], nil)` for an empty input — a nil `collection` would then be handed to `tasks.NewSearchTask`. Guard it.
- `qv_transform_segment.go:198-203` `Release` closes nothing in `s.waiters`, so a `WaitTransformApplied` caller with a non-cancellable context blocks forever after release.
- `qv_transform_segment.go:19`/`:32`: the `releaser qvSegmentManager` field is stored but never used, and `qvSegmentManager.Remove` (`qv_segment_interfaces.go:23`) has no callers — dead coupling to `segments.Manager`; drop it or use it.
- `segment_resource_types.go:207-211`: because `1 << iota` starts at `iota == 1`, `SegmentUpdateReopen == 2` and `SegmentUpdateLoadIndex == 4`, leaving bit 0 unused. Prefer explicit `1 << 0` / `1 << 1`.
- `classifySegmentUpdate:143` always returns `Reopen|LoadIndex`, and `qv_physical_segment_loader.go:60-71` then does a `Reopen` followed by a `LoadIndex` on every revision bump. If `Reopen` already installs the new index, the second call is wasted work; if it does not, the two bits should be derived from the actual snapshot delta rather than hardcoded.
- `view_scoped_physical_segment_manager.go:219-221`: when the acquire context is already cancelled, `submission.done()` is called but `state.loading` is left `true`. In the stream-less mode, if another view still refs that segment, `recordView:194` will never resubmit (it requires `!state.loading`) and the surviving view hangs with no error. An unconditional state reset in the `done` chain would close it.
- `services.go:1468-1474` `leaderViewCatchingUpStreamingData` forces `CatchingUpStreamingData=false` for streaming-node-embedded QueryNodes. This changes QueryCoord's serviceability signal and looks unrelated to "prepare QueryView resources on QueryNode" — worth its own PR, or at least a line in the description explaining why it belongs here.
- `qv_physical_segment_loader.go:118-120`: `LoadSegment` returns early for `SegmentLevel_L0`, so an L0 segment is created and then handed out fully empty. Intentional (L0 deletes arrive via the transform log)? A comment would help the next reader.
- New code uses `fmt.Errorf` in eleven places (`buffer.go`, `qv_physical_segment_loader.go`, `qv_transform_segment.go`). Per repo convention these should be `merr` errors — and per finding #5 the classification actually matters here, it is not just style.
- Test coverage: the suites for the readiness manager and physical manager are genuinely good on sequential paths (revision rollback, reset-cancels-update, catch-up failure). What is missing is exactly where findings #1–#3 live — a `-race` stress test with N views over a shared segment set, overlapping `Acquire` / `AcquireSealedSegmentHandles` / view release / `failSegment`, and a catch-up assertion that pins which `startAfter` a shared segment ends up registered with.

PR #52592, opened 2026-08-17
Data node Pickle issue backporting

saikishore143-hub:


saikishore143-hub:
More context here https://github.com/milvus-io/milvus/discussions/50872

Issue #52598, opened 2026-08-17
[Enhancement]: replace the unbuildable telemetry examples with a real integration test


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Delete `examples/telemetry_demo` and `examples/telemetry_e2e_test`, and cover the client-telemetry round trip with a suite under `tests/integration/telemetry/` instead.

### Why is this needed?

The two example modules are 2253 lines that nothing in the repo references -- no doc, no README, no Makefile target, no CI job. `.golangci.yml` excludes `^examples/` and they are separate Go modules, so the root build and lint never touch them. They rotted accordingly:

- `examples/telemetry_demo/go.mod` carried an absolute path from a contributor's laptop (`/Users/.../milvus-proto/go-api`) until an outside user reported #52547.
- Both modules still replaced `client/v2` while the in-tree module had become `client/v3`, so neither had compiled for a long time.
- `examples/telemetry_e2e_test` pushes `collection_metrics` with `Persistent: true`, which the coordinator has rejected since the feature's first PR (#47523) -- "only push_config can be persistent". That program has never run to completion, and the bug was merged again in #52550 without anything noticing.

Meanwhile the feature's automated coverage is one-sided. There are ~251 unit tests across the client and the coordinator, but the client-side suite mocks the server and the server-side suite mocks the client, so nothing exercises the wire between them: heartbeat carrying metrics, commands riding the heartbeat response, replies riding the next one. `tests/` has no telemetry coverage at all.

An integration suite is worth more than the examples were: it runs in CI, it fails when someone breaks the round trip, and it can pin the rules that the examples got wrong.

### Anything else?

Care is needed with timing. A heartbeat carries only the operations since the previous one and the client resets its counters as it snapshots, so any assertion of the form "run load, stop, then look" is a coin flip on which window the query lands in -- that is exactly why the demo printed `[WARN] Expected operation 'Search' not found in metrics`. Assertions have to keep issuing load while they poll.

Issue #52547, opened 2026-08-15
[Bug]: telemetry_demo go.mod replace uses an absolute local path


daviddallakyan2005:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: master
- Deployment mode(standalone or cluster): n/a (example module)
- MQ type(rocksmq, pulsar or kafka): n/a
- SDK version(e.g. pymilvus v2.0.0rc2): in-tree Go client
- OS(Ubuntu or CentOS): n/a
- CPU/Memory: n/a
- GPU: n/a
- Others: `examples/telemetry_demo`

### Current Behavior

`examples/telemetry_demo/go.mod` contains:

```
replace github.com/milvus-io/milvus-proto/go-api/v2 => /Users/xiaofanluan/code/milvus-proto/go-api
```

That absolute path is not on other machines, so `go mod tidy` / `go build` in the example cannot succeed off that laptop. The in-tree client module is also `github.com/milvus-io/milvus/client/v3`, while the example still replaces `client/v2 => ../../client`.

### Expected Behavior

The example module should resolve proto from the module proxy and the in-tree client, with no host-specific replace paths.

### Steps To Reproduce

1. Clone milvus-io/milvus at master
2. `cd examples/telemetry_demo && go mod tidy && go build main.go`
3. Module resolution fails on the absolute replace path

### Anything else?

Will send a PR that drops the laptop replace and points this example at in-tree `client/v3`.

xiaofan-luan:
@daviddallakyan2005 
Thanks for the contribution.
I will issue another pr to improve the test of client telemtry protocol

PR #47523, opened 2026-02-03
feat: Add client-side telemetry with heartbeat and server command support

xiaofan-luan:
issue: #47281
design doc: https://github.com/milvus-io/milvus-design-docs/blob/main/design_docs/20260131-client_side_telemetry.md

Implement comprehensive client-side telemetry system that includes:

- Metrics collection for Search, Query, Insert, Delete, Upsert operations
- Automatic heartbeat reporting to server with configurable intervals
- Per-collection metrics tracking with wildcard support
- P99 latency calculation using ring buffer sampling
- Server-push command handling (push_config, collection_metrics, show_errors)
- Historical snapshot storage for latency history queries
- Error tracking with circular buffer for recent errors
- WebUI for telemetry visualization and command management
- HTTP API endpoints for telemetry data access and command push
- RootCoord telemetry manager for centralized command routing

PR #52599, opened 2026-08-17
test: cover the client-telemetry round trip and drop the stale examples

xiaofan-luan:
## Summary

issue: #52598

Deletes `examples/telemetry_demo` and `examples/telemetry_e2e_test` (2253 lines, nothing in the repo references them, CI never compiled them) and covers the same ground with `tests/integration/telemetry/`, which CI does run.

Why they had to go rather than be fixed: `.golangci.yml` excludes `^examples/` and both are separate Go modules, so nothing built or linted them. A contributor's laptop path sat in `go.mod` until an outside user filed #52547; both still replaced `client/v2` after the in-tree module became `client/v3`; and `telemetry_e2e_test` pushes `collection_metrics` with `Persistent: true`, which the coordinator has rejected since the feature's first PR (#47523). That program has never run to completion, and the bug was merged again in #52550 with nothing noticing. Fixing them would leave the same code nobody runs.

What the new suite covers, none of which unit tests can reach -- the client suite mocks the server, the server suite mocks the client, so the wire between them was covered by nothing:

- `TestMetricsReachTheCoordinator` -- operations issued through the SDK appear in a telemetry query, attributed to that client.
- `TestPushedCommandReachesTheClient` -- a command pushed at the coordinator rides the heartbeat response and is executed by the SDK's handler. It registers a handler for a type the SDK does not know, so what the test observes can only have come from this push.
- `TestOnlyPushConfigMayBePersistent` -- the rule the deleted example got wrong, asserted across the RPC boundary: `collection_metrics` is refused as persistent, accepted as one-time, and `push_config` is accepted and then deletable.
- `TestPersistentConfigNeedsAStableClientID` -- the other half: a persistent config aimed at a client with a generated ID is refused rather than accepted and quietly ineffective.

Every assertion keeps issuing load while it polls. A heartbeat carries only the operations since the previous one and the client resets its counters as it snapshots, so "run load, stop, then look" is a coin flip on which window the query lands in -- that is exactly what made the old demo print `[WARN] Expected operation 'Search' not found in metrics` about its own traffic.

The two rejection cases match on the error message rather than with `errors.Is`. Running them showed why: a merr error crossing the coordinator's gRPC boundary arrives as `code = Unknown` with everything flattened into the description, so `ErrParameterInvalid` is no longer in the chain. Matching the wording also pins that the request was refused for the stated reason rather than, say, a cold coordinator cache.

## Test plan

Run locally against a real cluster (`MiniClusterV3` spawning mixcoord, proxy, datanode, querynode and streamingnode, with etcd, minio and pulsar), three consecutive times:

```
--- PASS: TestTelemetry (24.82s)
    --- PASS: TestTelemetry/TestMetricsReachTheCoordinator (5.84s)
    --- PASS: TestTelemetry/TestOnlyPushConfigMayBePersistent (1.75s)
    --- PASS: TestTelemetry/TestPersistentConfigNeedsAStableClientID (1.88s)
    --- PASS: TestTelemetry/TestPushedCommandReachesTheClient (3.54s)
ok      github.com/milvus-io/milvus/tests/integration/telemetry  26.610s
ok      github.com/milvus-io/milvus/tests/integration/telemetry  28.722s
ok      github.com/milvus-io/milvus/tests/integration/telemetry  27.883s
```

The whole suite is ~27s, and the repeat runs are there because assertions about heartbeat timing are exactly the kind that pass once and flake later.


xiaofan-luan:
Folded into #52597 -- the retention change and the test that covers it belong in one PR, and the integration suite now asserts the new behaviour end to end (plus a red check with retention set to 1).

Issue #52601, opened 2026-08-17
[Bug]: Milvus cannot authenticate to Kafka 4.x with SASL/SCRAM (librdkafka 1.9.1 nonce defect)


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: reproduced on 2.5.13, 2.6.17 and master (007867e647)
- Deployment mode: standalone and cluster
- MQ type: external Kafka 4.0.0, KRaft mode
- SDK version: pymilvus 2.6.x
- OS: Linux
```

### Current Behavior

Milvus cannot connect to an Apache Kafka 4.x broker when the listener uses
SASL with SCRAM-SHA-256. Every connection attempt fails authentication:

```
SASL authentication error: Authentication failed during authentication due to
invalid credentials with SASL mechanism SCRAM-SHA-256 (after 308ms in state AUTH_REQ)
```

and the broker logs:

```
Failed authentication with /127.0.0.1 (Authentication failed during authentication
due to invalid credentials with SASL mechanism SCRAM-SHA-256)
```

The credentials are valid: on the same broker, with the same user and password,
`kafka-topics.sh` (Java client) authenticates successfully. The same Milvus build
against Kafka 3.9.1 with the identical SCRAM configuration connects normally.

### Root cause

Milvus pins `librdkafka/1.9.1` in `internal/core/conanfile.py` on every line
(2.5, 2.6, 3.0, master), and links it dynamically.

librdkafka appended the client nonce to the server-sent nonce a second time when
building the SCRAM client-final-message, although the server-sent nonce already
begins with it. The defect has been present since librdkafka v0.0.99. Apache
Kafka tolerated it, then restored the strict nonce comparison required by
RFC 5802, so brokers on the 4.x line now reject the proof.

Enabling librdkafka protocol tracing shows the negotiation itself is healthy —
ApiVersions v3 accepted, SaslHandshake v1 accepted, broker advertising
SCRAM-SHA-256, client-first-message sent and server-first-message parsed — and
the failure occurring only after the client-final-message is sent. The 300ms
before the rejection is Kafka's deliberate delay on a failed authentication.

Upstream fixed this in librdkafka 2.6.1 (confluentinc/librdkafka#4895).

### Version matrix

Each row tested against the same broker, same user, same password, SCRAM
credential at 4096 iterations:

| client | Kafka 4.0.0 | Kafka 3.9.1 |
|---|---|---|
[...]

PR #52602, opened 2026-08-17
fix: upgrade librdkafka to 2.6.1 for SASL/SCRAM against Kafka 4.x

czs007:
issue: #52601

when building the SCRAM client-final-message, although that nonce already
begins with it. The defect dates back to librdkafka v0.0.99. Kafka tolerated
it and then restored the strict nonce comparison required by RFC 5802, so
brokers on the 4.x line reject the proof. Fixed upstream in librdkafka 2.6.1
(confluentinc/librdkafka#4895).

Tested against the same broker, same user, same password:

  librdkafka 1.9.1   fails
  librdkafka 2.6.0   fails
  librdkafka 2.6.1   works
  Java client        works

2.6.0 still fails, so the threshold is 2.6.1 rather than 2.x generally.

2.6.1 is preferred over the later versions because it is the lowest one
carrying the fix and is a maintenance release. The later feature releases
change the consumer group protocol and deprecate api.version.request, which
Milvus sets explicitly in both the msgstream and streaming Kafka clients.

requires a minimum of 1.9.0.

Verified on this branch: full make milvus succeeds, Milvus starts with zero
SASL errors against Kafka 4.0.0 with SCRAM-SHA-256, and create collection /
insert / flush / load / search all complete through the Kafka WAL.

## Dependency

The recipe has to exist in the Conan remote before this merges. Companion PR:
milvus-io/conanfiles#179 adds `librdkafka/2.6.1` (same source URL, sha256 and
patches as conan-center's own 2.6.1 recipe).

## Other branches

2.5 and 2.6 resolve exclusively from `default-conan-local` (Conan 1,
`conan install ... -r default-conan-local -u`), which carries librdkafka up to
2.2.0. Those lines need the recipe published to that remote separately before the
same one-line bump can be applied there.

### DIFF
--- a/internal/core/conanfile.py
+++ b/internal/core/conanfile.py
@@ -31,7 +31,7 @@ class MilvusConan(ConanFile):
         "milvus-common/1.0.0-b589c5a@milvus/dev#d431af735c8acb829feb7a94f05daf42",
         "google-cloud-cpp/2.28.0@milvus/dev#468918b43cec43624531a0340398cf43",
         "opentelemetry-cpp/1.23.0@milvus/dev#11bc565ec6e82910ae8f7471da756720",
-        "librdkafka/1.9.1#ec1a00d5414f618555799be9566adfb7",
+        "librdkafka/2.6.1@milvus/dev#a15d9fefad917290d59fa3fcbc144888",
         "roaring/3.0.0#25a703f80eda0764a31ef939229e202d",
         "crc32c/1.1.2",
         "simde/0.8.2#5e1edfd5cba92f25d79bf6ef4616b972",
--- a/pkg/mq/msgstream/mqwrapper/kafka/kafka_consumer_test.go
+++ b/pkg/mq/msgstream/mqwrapper/kafka/kafka_consumer_test.go
@@ -162,9 +162,15 @@ func TestKafkaConsumer_GetLatestMsgID(t *testing.T) {
 	assert.NoError(t, err)
 	defer consumer.Close()
 
+	// GetLatestMsgID queries the watermark offsets without allowing topic
+	// auto-creation, so the topic must exist before the first call. The
+	// mock cluster bundled with older librdkafka (<= 1.9.x, Metadata API
+	// v2) created it implicitly; newer versions follow the request flag.
+	createTopic(t, topic)
+
 	latestMsgID, err := consumer.GetLatestMsgID()
-	assert.Equal(t, int64(0), latestMsgID.(*KafkaID).MessageID)
 	assert.NoError(t, err)
+	assert.Equal(t, int64(0), latestMsgID.(*KafkaID).MessageID)
 
 	data1 := []int{111, 222, 333}
 	data2 := []string{"111", "222", "333"}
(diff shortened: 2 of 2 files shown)


czs007:
Updated: the recipe has landed via milvus-io/conanfiles#179 and is published to `default-conan-local2` as `librdkafka/2.6.1@milvus/dev` (recipe revision `a15d9fefad917290d59fa3fcbc144888`).

Repinned to that reference rather than the ConanCenter one, so resolution stays on the private remote — verified that `conan graph info` now downloads it from `default-conan-local2`.

Rebuilt and re-verified against the published package: the full build succeeds, Milvus starts with zero SASL errors against Kafka 4.0.0 with SCRAM-SHA-256, and create collection / insert / flush / load / search all complete through the Kafka WAL.

czs007:
Rebased onto current master (ea0c533271); the earlier CI failures were from the stale base.

Local verification of this exact commit (5a2d893a59):

- full build succeeds; the built `librdkafka.so.1` reports `rd_kafka_version_str() = 2.6.1`
- standalone started with `mq.type=kafka` against an Apache Kafka 4.0.0 KRaft broker, `SASL_PLAINTEXT` + `SCRAM-SHA-256`
- `/proc/<pid>/maps` of the running process maps only that 2.6.1 `librdkafka.so.1`
- `SASL authentication error` count in the logs: 0 (with 1.9.1 every broker connection fails here)
- create collection / insert 2000 rows / flush / index / load / search all pass; the `*-rootcoord-dml_*` topics show up on the 4.0.0 broker

Branch PRs (conan 1.x recipe published as `librdkafka/2.6.1#5b0ecc7c75e262b3e22d652018bedb93` to default-conan-local via milvus-io/conanfiles#180): #52666 (2.6), #52667 (poc_2.6.17), #52668 (hotfix-2.5.13).

czs007:
The `build-ut-cov` failure on the previous push (`TestKafkaConsumer_GetLatestMsgID` panicking on a nil `MessageID`) is real and caused by this upgrade, but it is a test-only issue. Root cause:

- The kafka unit tests run against librdkafka's built-in mock cluster (`kafka.NewMockCluster`).
- The 1.9.x mock cluster only implemented Metadata API v0–2, which has no `AllowAutoTopicCreation` field, so it implicitly created every topic named in a metadata request. The 2.6.1 mock cluster implements up to v12 and honours the flag.
- `GetLatestMsgID()` → `QueryWatermarkOffsets` → `rd_kafka_topic_partition_list_query_leaders` issues the metadata refresh with `allow_auto_create=false` (identical in 1.9.1 and 2.6.1), so against the new mock a never-created topic yields `Local: Unknown partition` instead of offsets `0/0`.
- Against a real broker (Metadata v4+ everywhere) 1.9.1 already behaved like this, and `newKafkaConsumer` explicitly handles `ErrUnknownPartition`, so production semantics are unchanged. `TestKafkaConsumer_CheckPreTopicValid` had the same dependency and only surfaced after the first panic was fixed.

Fix (pushed as part of this commit): the two tests now create the topic first through a producer metadata request (producer default `allow.auto.create.topics=true`). `go test -race -tags dynamic,test ./mq/msgstream/mqwrapper/kafka/` passes repeatedly with 2.6.1 linked dynamically. Note that without `-tags dynamic` the package silently uses the librdkafka bundled in confluent-kafka-go, which is why the failure doesn't reproduce with a plain `go test`.

Same test change applied to #52666 / #52667 / #52668.

czs007:
rerun go-sdk

PR #52604, opened 2026-08-18
feat: hand a shard split window over in querycoord without rebuilding the source

xiaocai2333:
issue: #50463
design doc: `docs/design-docs/design_docs/20260610-shard_split.md` (added by #53366)

## What

Online shard split, layer 6 of 6: the QueryCoord window handover. This PR's old meta/DDL content is superseded by #53366; it now carries this layer only — its own 12 commits touch `internal/querycoordv2` alone.

- Split targets stay unwatched during the window; next targets pulled inside the window are marked and are neither synced nor promoted to the new shards. The current target is served from a window snapshot without those targets, so a load issued inside the window still completes.
- A retired source is never re-watched or moved before the flip. Normal and stopping balance, and manual LoadBalance / TransferSegment / TransferChannel, are refused with a retriable error in four cases: a shard is splitting, the current target still lists a retired source, the next target marks window targets, or the shard states are unreadable.
- Fail-closed means the whole collection: while shard states are unreadable the channel checker watches nothing of that collection, not only the split targets.
- Segment attribution across the window: a re-homed sealed segment is attributed to the target it moved to, and a segment is loaded through the shard its target attributes it to. Without these an adopted target never becomes data-ready and the current target never flips.
- The window ends on the target observer's own check (`isSplitWindowOver`, guarded against a state read older than the pull, with the existing next-target survive time as a fallback).
- An adopted target is placed on the node that fronts its child (unit-tested only; the local e2e runs a single QueryNode).
- Proxies' shard-leader caches are invalidated at the flip and whenever a node's distribution report drops a delegator (a release, among other causes).
- New dependency: QueryCoord reads shard state through a TTL-cached `DescribeCollection` per collection, consulted by the balance checker, the channel checker, the target observer and the ops services each cycle. `TargetManagerInterface` gains `GetSplitWindowTargets` / `GetSplitWindowExclusions` (mock regenerated).

**Stack** (merge in this order):
1. #53366 write switch + pk residue routing + design doc (L1)
2. #53093 proxy write path (L2) — on L1
3. #53095 DataCoord orchestration (L3) — on L1
4. #53097 rewrite data plane (L4) — on L3
5. #53096 QueryNode read path (L5) — on L1
6. #52604 QueryCoord window handover (L6) — on L4, and carrying L5's changes as re-created commits (it is not a descendant of #53096, so its range against #53097 is 38 commits: L5's 26 re-created + its own 12)

Every PR except #52604 contains exactly the commits of the PRs it depends on, so reviewing its own range is enough.

**Verification**
- The six branches are rebased on master `f36cf77d79` and were assembled from one local integration branch (not published, so this comparison cannot be reproduced from the PR refs): the top of the stack plus L2 is tree-identical to that branch with every review fix merged.
- After the rebase, `go build ./internal/...` and `go vet` of every touched package pass on that full stack.
- Standalone e2e on this exact stack (`029b002125`, one fresh standalone per case, pk-routed collection, auto size trigger, rewrite split): **8/8 PASS**, no error or panic lines. T0 write/fence; T2 auto-trigger split (DescribeCollection routing fields match the replay); T3–T6 reads, writes and deletes inside the window, clean variant, including a load that completes inside the window (100% in 25.9s, before adoption); T3–T6 with source L0 — the source L0s retired and 0 deletes resurrected; T7 children; T8 cascading split, 10s after the first split's Done, with both grandchildren spawned and exact results before adoption, after adoption and after Done.
- Not verified: multi-QueryNode behaviour (the local e2e runs one QueryNode), replication against a real secondary cluster, and CI (the CI jobs have not run on these heads yet).


xiaocai2333:
Force-pushed: re-stacked.

`7dcc54a992` → `47621ef73e`. Now on the current heads of #50495 (`e65610907f`) and #52583 (`9b57e40612`) instead of bundling stale copies of them. Its own commits are unchanged in content, plus the namespace admission gate on `CommitShardSplitRouting` described in the body (`a3c1e152f0`) and the US-locale spelling fixes the misspell gate demanded.

czs007:
Review of the QueryCoord handover (L6) against design §6.2 and §6.4. I found no QueryCoord-side interleaving that serves a row twice or not at all. One high-impact availability cost is not described in the design. Line references are at head `94762f81c0`.

### Stopping balance is frozen for the collection for the whole window, which blocks graceful stop and rolling upgrades

- `readyToCheck` returns false while `frozenForShardSplit` holds ([balance_checker.go#L190](https://github.com/milvus-io/milvus/blob/94762f81c075392c2bd46274e223504bced236dc/internal/querycoordv2/checkers/balance_checker.go#L190)). Both the normal and the stopping queue are built through it.
- The window lasts as long as the rewrite, which can be hours. With a snapshot-protected source segment it has no bound at all (§11).
- During it, every QueryNode holding a channel or segment of the collection can only leave by hitting `queryNode.gracefulStopTimeout` (1800s by default), with nothing moved off it first. A rolling upgrade therefore pays about 30 minutes per affected node.
- If the node that goes down holds the source, it also triggers the read-path failures reported on #53096 (respawn from the target checkpoint).
- Resource-group moves that rely on balance wait for the window too.
- §11 mentions only that the freeze is per collection.
- Suggested direction: let the stopping balance move channels and segments that are neither the source nor a target, or at least document the upgrade constraint in the rollout section.

Issue #52387, opened 2026-08-11
[Bug]: Streaming gRPCs (CreateReplicateStream, DumpMessages) bypass authentication and RBAC on the external proxy port


Amos-yongxin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: 3.0
- Deployment mode(standalone or cluster): cluster / standalone
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

The external (client-facing) gRPC server of the Proxy only installs a
**unary** interceptor chain. In `internal/distributed/proxy/service.go`
(`startExternalGrpc`), the server is configured with:

```go
unaryServerOption = grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
    ...
    proxy.GrpcAuthInterceptor(proxy.AuthenticationInterceptor),   // authentication
    proxy.UnaryServerInterceptor(proxy.PrivilegeInterceptor),     // RBAC authorization
    ...
))
```

There is **no** `grpc.StreamInterceptor` on this server. As a result, the
streaming RPCs exposed on `MilvusService` bypass the authentication and
authorization chain entirely:

- `CreateReplicateStream` (`internal/proxy/impl.go`) — establishes a
  replication stream and writes replicate messages into the WAL data plane.
- `DumpMessages` (`internal/proxy/impl.go`) — streams raw WAL messages out
  for data salvage.

Because both methods are handled as gRPC streams, they never traverse the
unary chain that performs `AuthenticationInterceptor` (credential/token
verification) and `PrivilegeInterceptor` (casbin/RBAC enforcement). This
means an **unauthenticated** client that can reach the external proxy port
can invoke these streaming RPCs even when `common.security.authorizationEnabled`
is `true`.

Note: the *internal* gRPC server does install a `StreamInterceptor`, but it
only performs cluster/serverID validation for inter-component traffic — it
does **not** perform user authentication or RBAC, and it is not the
client-facing surface.

### Impact

- **Authentication bypass**: streaming RPCs are reachable without valid
  credentials, even with authorization enabled.
- **Authorization bypass**: no RBAC check is applied, so a caller can trigger
  cluster-admin-level operations (WAL replication / message dump) without the
  corresponding privilege.
[...]

PR #52608, opened 2026-08-18
fix: [RBAC] Streaming gRPCs bypass authentication on the external proxy port

shaoting-huang:
## What

Streaming RPCs (`CreateReplicateStream`, `DumpMessages`) were not going through the unary interceptor chain on the external client-facing gRPC port. The external server installed only a `UnaryInterceptor`, so the stream handlers bypassed both authentication and authorization entirely — even with `common.security.authorizationEnabled=true`.

related: #52387 and also covers #52238.

## Fix

Add a dedicated **stream interceptor chain** to the external gRPC server (`startExternalGrpc`) that mirrors the unary chain:

1. **Authentication**: `proxy.GrpcAuthStreamInterceptor` runs the existing `AuthenticationInterceptor` on every streaming call and propagates the authenticated context to the handler via a wrapped `ServerStream`. Skipped when authorization is disabled, consistent with the unary chain.
2. **Authorization (RBAC)**: `proxy.PrivilegeStreamInterceptor` enforces per full-method name via a static `streamMethodAuthorizers` table:
   - `CreateReplicateStream` → casbin against the cluster-level `PrivilegeUpdateReplicateConfiguration` (Global scope, `db=*`), so a dedicated role can be granted replication rights. Uses the same `util.AnyWord` global scope as the unary path (fixes the db-scope bug raised on #52386).
   - `DumpMessages` → `authorizeWALRead`, which enforces the cluster-admin grant (Global `PrivilegeGroupAdmin` / `PrivilegeAll`, built-in `admin` role, or `root`) through the shared `enforceClusterPrivilege` helper, matching the CDC data-salvage requirement from #52238. Custom superuser roles behave consistently with `CreateReplicateStream`, while a role holding only the narrow `PrivilegeUpdateReplicateConfiguration` grant is still denied — replication rights do not imply raw WAL export rights.
   - **Fail-closed default**: any streaming method not registered in the table is denied, so a newly-added stream RPC cannot silently bypass RBAC.
   - `grpc.health.v1.Health/Watch` is exempt from RBAC (exact method match) so liveness probes keep working — the stream is still authenticated, matching unary `health.Check`.

3. **milvusclient**: `MetadataStreamInterceptor` so authorized clients still carry `authorization`/`dbname`/`identifier` metadata on stream RPCs when authorization is enabled.

The inline authz gate previously added in the `DumpMessages` handler is removed — authorization is now enforced once, at the interceptor layer.

Streaming rate limiting and access logging are intentionally out of scope of this fix: the WAL data plane gets a dedicated per-message rate type and a stream-aware audit log in follow-ups.


internal/proxy/privilege_interceptor.go
```diff
@@ -237,3 +239,171 @@ func resolveCollectionAlias(ctx context.Context, dbName, nameOrAlias string) (st
...
+	}
+	if !Params.CommonCfg.RootShouldBindRole.GetAsBool() && username == util.UserRoot {
+		return ctx, nil
+	}
+	roles, err := GetRole(username)
+	if err != nil {
+		return ctx, err
+	}
+	for _, role := range roles {
+		if role == util.RoleAdmin {
```
yhmo: internal/proxy/privilege_interceptor.go line:262
Low ---- authorizeWALRead only accepts the literal built-in "admin" role (or root) and bypasses the casbin path entirely. A custom role granted PrivilegeAll / PrivilegeGroupAdmin on Global is treated as cluster-admin by the casbin matcher (PrivilegeGroupContains / r.sub=="admin" path) and would pass authorizeCreateReplicateStream on the same stream port, yet such a user is denied DumpMessages. Is this asymmetry intentional? If not, routing authorizeWALRead through the enforcer (e.g. enforce a cluster-scoped read privilege or PrivilegeAll at db="*") would make custom superuser roles behave consistently with CreateReplicateStream.

internal/proxy/privilege_interceptor.go
```diff
@@ -237,3 +239,171 @@ func resolveCollectionAlias(ctx context.Context, dbName, nameOrAlias string) (st
...
+	return authorize(ctx)
+}
+
+// authorizeCreateReplicateStream authorizes CreateReplicateStream, which writes
+// replicated messages into the WAL data plane. It is a cluster-scoped operation
+// and is authorized via casbin against PrivilegeUpdateReplicateConfiguration
+// (Global scope). Like the unary PrivilegeInterceptor, cluster-level privileges
+// are authorized against util.AnyWord ("*") db scope, independent of the
+// connection namespace.
+func authorizeCreateReplicateStream(ctx context.Context) (context.Context, error) {
```
yhmo, line 420: internal/proxy/privilege_interceptor.go line:357
Low ---- authorizeCreateReplicateStream re-implements ~60 lines of the unary PrivilegeInterceptor (root exemption, GetRole + RolePublic, casbin enforce + result cache, apikey error masking). Both live in this file and the security contract depends on them staying in sync; a future change to the unary enforce/cache flow (privilege-group handling, alias resolution, new cluster-level behavior) can drift here without any compile-time signal. Consider extracting a shared helper that both paths call with (ctx, privilege, objectType, objectName, dbName).

internal/distributed/proxy/service.go
```diff
@@ -306,8 +307,19 @@ func (s *Server) startExternalGrpc(errChan chan error) {
 			proxy.TraceLogInterceptor,
 			connection.KeepActiveInterceptor,
 		))
+		// Streaming RPCs (e.g. CreateReplicateStream, DumpMessages) bypass the
+		// unary chain, so they must be authenticated and authorized through a
+		// dedicated stream interceptor chain. Without this the external server
+		// would accept unauthenticated/unauthorized streaming calls on the
+		// client port. The order mirrors the unary chain: first authenticate
+		// (resolve the user into ctx), then enforce authorization on that user.
+		streamServerOption = grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(
```
yhmo: internal/distributed/proxy/service.go line:316
Low ---- This is the security-critical wiring of the whole fix (the external server must run the stream auth + RBAC chain for every stream), and it has no test: the new tests exercise GrpcAuthStreamInterceptor / PrivilegeStreamInterceptor in isolation but never assert that startExternalGrpc actually installs the chain. A future refactor that drops, reorders, or guards this block would silently regress the fix with no test failing. Consider a test that starts the external server with authorization enabled and asserts an unauthenticated stream (e.g. a raw DumpMessages or an unregistered stream method) is rejected.

internal/proxy/dump_messages_authz_test.go
```diff
@@ -0,0 +1,353 @@
...
+		require.Error(t, err)
+		assert.Equal(t, codes.PermissionDenied, status.Code(err))
+		assert.False(t, handlerCalled)
+	})
+
+	t.Run("health watch is exempt", func(t *testing.T) {
+		handlerCalled := false
+		stream := &mockDumpMessagesServer{ctx: context.Background()}
+		err := interceptor(nil, stream, &grpc.StreamServerInfo{
+			FullMethod: grpc_health_v1.Health_Watch_FullMethodName,
```
yhmo, line 369: internal/proxy/dump_messages_authz_test.go line:276
Low ---- This sub-test exercises only the RBAC exemption in PrivilegeStreamInterceptor in isolation (the context carries no credentials). In the real external chain (service.go:316) GrpcAuthStreamInterceptor runs AuthenticationInterceptor first, so with authorization enabled an unauthenticated Health/Watch probe is rejected with Unauthenticated before this exemption is ever reached. The PR body says the exemption exists "so liveness probes keep working" — worth documenting that gRPC health probes must carry credentials when authorization is enabled (consistent with the unary health.Check path), or the exemption is easy to misread as covering the whole health call.

internal/distributed/proxy/service.go
yhmo: internal/distributed/proxy/service.go line:316
Medium ---- The new stream chain installs only authentication and RBAC, while the unary chain also applies rate limiting (RateLimitInterceptor), access logging (accesslog.UnaryAccessLogInterceptor / UnaryUpdateAccessInfoInterceptor), mlog, and the hook interceptor. Consequently CreateReplicateStream (writes into the WAL) and DumpMessages (streams raw WAL data out) are neither rate-limited nor recorded in access logs, which undercuts observability for the exact security-sensitive data-plane methods this PR is protecting. Consider mirroring the unary chain rate-limit and access-log interceptors for these stream methods (at minimum for DumpMessages) so authorized WAL access is throttled and auditable like the unary data path.

client/milvusclient/interceptors.go
```diff
@@ -137,6 +137,16 @@ func (c *Client) MetadataUnaryInterceptor() grpc.UnaryClientInterceptor {
 	}
 }
 
+func (c *Client) MetadataStreamInterceptor() grpc.StreamClientInterceptor {
```
yhmo, line 148: client/milvusclient/interceptors.go line:140
Low ---- MetadataStreamInterceptor repeats the three-step context enrichment (c.metadata, c.state, c.extraInfo) already used by MetadataUnaryInterceptor. Extracting a shared helper such as func (c *Client) withClientMetadata(ctx context.Context) context.Context would keep the two call paths in sync when new headers are added (client_request_id was previously introduced only in extraInfo), avoiding future drift.

internal/proxy/privilege_interceptor.go
shaoting-huang: Thanks for the sharp catch. You are right: this asymmetry is unintended — a custom role holding Global PrivilegeAll / PrivilegeGroupAdmin passes `authorizeCreateReplicateStream` via the casbin matcher (r.sub=="admin" + privilegeGroupContains), yet is denied by `authorizeWALRead` which hard-codes the literal "admin" role.

I will route `authorizeWALRead` through the same enforcer instead, enforcing the cluster-level admin grant (Global, db="*", PrivilegeGroupAdmin) — this makes custom superuser roles consistent with `CreateReplicateStream` while preserving the key guardrail: a role holding only the narrow `PrivilegeUpdateReplicateConfiguration` (replication config) still cannot dump raw WAL. This matches the "such as admin/root authorization" wording of #52238 (admin/root is an example, the requirement is an explicit, suitably-privileged path).

This is addressed together with comment #2 by extracting a shared `enforceClusterPrivilege(ctx, privilege)` helper used by both stream authorizers. Fix is in progress.

internal/proxy/privilege_interceptor.go
shaoting-huang: Agreed, and thank you for calling it out — this is exactly the drift risk I want to avoid. I will extract a shared `enforceClusterPrivilege(ctx, privilege)` helper (root exemption → GetRole + RolePublic → casbin enforce with result cache → apikey masking) and have both `authorizeCreateReplicateStream` and `authorizeWALRead` call it. This also resolves the asymmetry raised in my comment on line 262 (WAL dump will enforce the cluster admin grant via the same helper). One shared enforcement path for all stream authorizers.

internal/distributed/proxy/service.go
shaoting-huang: Agreed — the wiring in `startExternalGrpc` is the security-critical piece and deserves a test that starts the real server. I will extract the interceptor options into `newExternalGrpcServerOptions(getMetaCache, limiter)` (shared by production and tests) and add a bufconn-based test that, with authorization enabled, asserts (a) an unauthenticated DumpMessages / CreateReplicateStream is rejected with Unauthenticated, (b) an unregistered stream method is rejected with PermissionDenied (fail-closed), and (c) an authorized admin stream reaches the handler.

internal/proxy/dump_messages_authz_test.go
shaoting-huang: Good point — the exemption only skips RBAC, not authentication. I will update the test to run the full chain: an unauthenticated Health/Watch is rejected with Unauthenticated (proving auth runs first), and an authenticated Health/Watch passes the RBAC exemption. I will also document in the PR body and the `streamHealthServicePrefix` comment that gRPC health probes must carry credentials when authorization is enabled (consistent with the unary health.Check path).

internal/distributed/proxy/service.go
shaoting-huang: Valid concern — CreateReplicateStream (writes WAL) and DumpMessages (streams raw WAL) are the most sensitive data-plane methods on the port, and they should be rate-limited and audited like the unary data path. I will add stream counterparts: a `RateLimitStreamInterceptor` (reusing the existing limiter) and a stream access-log interceptor, mirroring the unary `RateLimitInterceptor` / `UnaryUpdateAccessInfoInterceptor`, and wire them into the external stream chain in this PR. Follow-up: additional stream observability (trace/hook) can build on the same pattern.

client/milvusclient/interceptors.go
shaoting-huang: Agreed — I will extract the shared `withClientMetadata(ctx)` helper and use it in both `MetadataUnaryInterceptor` and `MetadataStreamInterceptor` so new headers (like client_request_id) stay in sync across the two paths.

internal/distributed/proxy/service_stream_test.go
```diff
@@ -0,0 +1,108 @@
...
+// chain built by newStreamInterceptorOption is actually installed on the
+// external server and enforces authentication at the gRPC boundary. A
+// regression that drops this chain would silently reopen the stream auth
+// bypass (#52387).
+func TestExternalStreamInterceptor_Wiring(t *testing.T) {
+	paramtable.Init()
+	proxy.Params.Save(proxy.Params.CommonCfg.AuthorizationEnabled.Key, "true")
+	defer proxy.Params.Reset(proxy.Params.CommonCfg.AuthorizationEnabled.Key)
+
+	getMetaCache := func() proxy.Cache { return nil }
```
yhmo: internal/distributed/proxy/service_stream_test.go line:69
High ---- This wiring test cannot pass as written. getMetaCache returns nil, so AuthenticationInterceptorWithMetaCache's readiness gate (getMetaCache() == nil -> merr.WrapErrServiceUnavailable) rejects every stream before any Authorization header is inspected. merr errors do not implement GRPCStatus, so gRPC maps them to codes.Unknown: the "unauthenticated DumpMessages is rejected" subtest expects codes.Unauthenticated and will receive codes.Unknown, and the "authenticated root reaches DumpMessages handler" subtest is rejected before the handler runs, so Recv() returns an error rather than nil. Independently, even with a valid cache, the mock handler returns without sending a message, so the client Recv() yields io.EOF and require.Equal(t, nil, err) fails. Inject a real cache (as the unit tests do, e.g. proxy.InitEmptyMetaCacheForTest()) and, for the authorized path, assert io.EOF or the stream-close behavior instead of nil.

internal/distributed/proxy/service.go
```diff
@@ -408,6 +418,22 @@ func (s *Server) startExternalGrpc(errChan chan error) {
...
+// cluster-level rate limiting, access logging, and mlog context enrichment.
+// It is extracted so the wiring is testable independently of the full Server
+// lifecycle — a regression here would silently reopen the stream auth bypass
+// (#52387), so the chain is covered by a dedicated test.
+func newStreamInterceptorOption(getMetaCache func() proxy.Cache, limiter types.Limiter) grpc.ServerOption {
+	return grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(
+		proxy.GrpcAuthStreamInterceptor(proxy.AuthenticationInterceptorWithMetaCache(getMetaCache)),
+		proxy.PrivilegeStreamInterceptor(proxy.StreamPrivilegeInterceptor),
+		proxy.RateLimitStreamInterceptor(limiter),
+		accesslog.StreamAccessLogInterceptor,
```
yhmo: internal/distributed/proxy/service.go line:432
Low ---- StreamAccessLogInterceptor runs after authentication and RBAC in the chain, so rejected stream attempts (unauthenticated or unauthorized DumpMessages/CreateReplicateStream) never reach the access log. The unary chain places accesslog.UnaryAccessLogInterceptor before GrpcAuthInterceptor, so denied unary requests ARE audited. Given the access log was added here precisely to audit the security-sensitive WAL data plane, the deny events are arguably the most audit-worthy ones. Consider logging before auth and enriching the user after auth (mirroring UnaryUpdateAccessInfoInterceptor) so both allowed and denied stream calls are captured.

internal/proxy/rate_limit_interceptor.go
```diff
@@ -40,6 +40,43 @@ func RateLimitInterceptor(limiter types.Limiter) grpc.UnaryServerInterceptor {
...
+func RateLimitStreamInterceptor(limiter types.Limiter) grpc.StreamServerInterceptor {
+	return func(srv any, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		rt, ok := streamMethodRateTypes[info.FullMethod]
+		if !ok {
+			// Unknown streaming method: no dedicated rate-limit bucket. The
+			// authorization chain still fail-closes on it via
+			// StreamPrivilegeInterceptor.
+			return handler(srv, ss)
+		}
+		err := limiter.Check(util.InvalidDBID, map[int64][]int64{}, rt, 1)
```
yhmo: internal/proxy/rate_limit_interceptor.go line:67
Low ---- The stream rate-limit check runs once at stream establishment and charges the DDLCollection bucket (RateType_DDLCollection), so it only gates how many streams can be open at once rather than throttling WAL data flow the way the unary path throttles per request, despite the comment claiming the WAL data plane is "throttled like the unary data path". It also couples stream opens to the DDL quota: when a DDLCollection limit is configured, many concurrent DumpMessages/CreateReplicateStream opens share the same bucket as CreateCollection/DropCollection and can rate-limit unrelated DDL (and vice versa). Consider a dedicated rate type or per-message accounting for these streams, or document this as an establishment gate only.

internal/proxy/privilege_interceptor.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/proxy/privilege_interceptor.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/distributed/proxy/service.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/proxy/dump_messages_authz_test.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/distributed/proxy/service.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

client/milvusclient/interceptors.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/distributed/proxy/service_stream_test.go
shaoting-huang: You are right — the test as written could not pass. Fixed:
- getMetaCache now returns InitEmptyMetaCacheForTest() (non-nil), so the auth readiness gate passes and the missing-authorization check is reached, yielding codes.Unauthenticated as asserted.
- Removed the "authenticated root reaches handler" subtest: constructing a mock admin credential is not feasible from the external grpcproxy test package. The positive auth+RBAC path is covered by the internal package tests (TestGrpcAuthStreamInterceptorChain_Authorized and the casbin allow/deny cases); this wiring test focuses on the security-critical assertion that the chain is installed and rejects an unauthenticated stream at the gRPC boundary.
- Also removed the CreateReplicateStream subtest (bidi stream establishment behavior is covered by internal unit tests).

internal/distributed/proxy/service.go
shaoting-huang: Agreed — denied stream attempts are exactly what the audit trail should capture. I moved StreamAccessLogInterceptor to the FRONT of the stream chain (before auth/RBAC), so unauthenticated/unauthorized stream attempts are logged too, mirroring the unary chain where accesslog.UnaryAccessLogInterceptor runs before GrpcAuthInterceptor. Added StreamUpdateAccessInfoInterceptor after authentication to enrich the user field for allowed calls. Follow-up: I will add a test asserting a denied stream produces an access-log entry.

internal/proxy/rate_limit_interceptor.go
shaoting-huang: Valid — the check is an establishment gate, not per-message flow throttling. I documented this honestly in RateLimitStreamInterceptor (it charges a stream open once against the cluster DDLCollection bucket, gates concurrent stream opens, and shares the bucket with cluster DDL). Per-message flow throttling for these streams would need a dedicated rate type; I noted it as a follow-up in the comment. If you prefer, I can gate only DumpMessages and leave CreateReplicateStream unthrottled, or drop the rate-limit interceptor to a follow-up PR entirely — let me know your preference.

internal/distributed/proxy/service_stream_test.go
shaoting-huang: Fixed — and the fix goes further than the review asked for. getMetaCache now returns proxy.InitEmptyMetaCacheForTest() so the auth readiness gate passes and the unauthenticated path yields codes.Unauthenticated as asserted. The orphan withOutgoingAuth helper (which referenced imports removed in the previous commit and broke the grpcproxy test build) is deleted, and an authorization-disabled subtest is added asserting an allowed stream reaches the handler.

Note the chain is also slimmed to auth -> RBAC -> mlog (see replies below): the access-log interceptor is gone, so the _globalL nil dependency that the chain-first placement would have introduced in this test can no longer occur.

internal/distributed/proxy/service.go
shaoting-huang: Agreed that denied stream attempts are the most audit-worthy events, and the unary accesslog-before-auth ordering is the right pattern. Rather than keep a disabled-by-default stream access log (GrpcStreamAccessInfo implemented the unary AccessInfo interface with a dozen placeholder methods just to cover two stream RPCs), I removed the stream access log from this PR entirely. Streaming audit for the WAL data plane — including deny events — is tracked as a dedicated follow-up where it can be designed as a stream-aware logger instead of placeholder-laden interface conformance. This also removes the nil-logger (uninitialized _globalL) panic that chain-first placement would have introduced.

internal/proxy/rate_limit_interceptor.go
shaoting-huang: Valid. The check was an establishment gate at best and its semantics were wrong: it charged once against the DDLCollection bucket, so it neither throttled per-message flow on long-lived streams nor gated concurrency, and it coupled stream opens to cluster DDL quota (and vice versa). I removed RateLimitStreamInterceptor entirely. Per-message flow throttling for these streams needs a dedicated rate type and is tracked as a follow-up. Dropping it is not a regression for #52387 — the auth bypass is closed by the auth + RBAC interceptors, which is what this PR is about.

internal/distributed/proxy/service_stream_test.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/distributed/proxy/service.go
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

internal/proxy/rate_limit_interceptor.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

chyezh:
**Third-party review note — upgrade / compatibility risks for primary-standby (CDC / replication) deployments**

I reviewed the stream auth/RBAC chain introduced by this PR (`newStreamInterceptorOption`, `GrpcAuthStreamInterceptor`, `PrivilegeStreamInterceptor` / `streamMethodAuthorizers`, and the client-side `MetadataStreamInterceptor`). The fix itself looks sound: fail-closed, auth-before-RBAC mirroring the unary chain, and all three stream methods on the external port (`CreateReplicateStream`, `DumpMessages`, `grpc.health.v1.Health/Watch`) are covered or explicitly exempted. The items below are upgrade risks I'd flag for rolling upgrades, especially for primary-standby / CDC setups that rely on `CreateReplicateStream` (write into the standby WAL) and `DumpMessages` (WAL salvage / catch-up).

1. **Version skew between server and stream clients (rolling-upgrade window).** The server-side enforcement and the client-side header injection are a matched pair: this PR hardens the proxy (external server) *and* fixes the Go SDK (`grpc.WithChainStreamInterceptor` + `MetadataStreamInterceptor`). If the proxy is upgraded before the stream callers (CDC / salvage tooling on the old SDK), the old clients open streams **without the `authorization` header** (stream RPCs never carried it before) and the new chain rejects them with `Unauthenticated` — replication and WAL salvage break until the clients are also upgraded. The reverse order (new SDK against old proxy) is harmless. Suggested upgrade order: stream callers first, then proxies; or keep `AuthorizationEnabled` off during the window. Only the Go SDK (`milvusclient`) is covered by this PR; pymilvus does not call these stream methods.

2. **The authorization model changes who can replicate / dump — credentials must be pre-granted.** With authorization enabled, `CreateReplicateStream` now requires the cluster-level `PrivilegeUpdateReplicateConfiguration` (Global scope) and `DumpMessages` requires the admin-level grant (`PrivilegeGroupAdmin` / `PrivilegeAll` / built-in `admin` / `root`). A CDC/backup account that previously worked must be granted these **before** the upgrade, or the failover/salvage path fails with `PermissionDenied` right after the upgrade (fail-fast, so no silent corruption, but the replication gap keeps growing). Note also that with `RootShouldBindRole=true` the `root` exemption no longer applies.

3. **`DumpMessages` authorization does not cover the modern `PrivilegeGroupClusterAdmin` group grant.** The request act is hardcoded to `PrivilegeGroupAdmin` (`authorizeWALRead`), and the casbin matcher only expands the legacy group literals (`PrivilegeAll` / `PrivilegeGroupReadOnly` / `PrivilegeGroupReadWrite` / `PrivilegeGroupAdmin`). A role granted the `PrivilegeGroupClusterAdmin` group (the current recommended way to grant admin) is **denied** for `DumpMessages` while being able to create replicate streams (the expanded members include `PrivilegeUpdateReplicateConfiguration`). For operators, this is a support trap in exactly the primary-standby scenario this PR is meant to secure — worth confirming the intended semantics and documenting (or accepting `PrivilegeGroupClusterAdmin` as equivalent to `PrivilegeGroupAdmin`).

4. **`Health/Watch` on the external port now requires credentials when authorization is enabled.** The chain authenticates before the RBAC exemption, so an unauthenticated gRPC health probe (LB / k8s probe / health-check scripts) is rejected with `Unauthenticated` — previously `Watch` was fully open. This is consistent with the unary `Check` path, but it is a behavior change: probes on the external port may start failing after the upgrade and cause false "unhealthy" verdicts (premature failover / failed fail-back). Probes should use the internal port or carry credentials.

5. **Startup window: stream calls return `ServiceUnavailable` until the meta cache is published.** The readiness gate in `AuthenticationInterceptorWithMetaCache` now applies to streams too (even with authorization disabled), so `Health/Watch` and other streams fail during proxy startup / rolling restart until `Proxy.Init()` publishes the cache. Same as the unary path, but health probes that only used `Watch` were previously immune.

6. **Fail-closed table drift.** `streamMethodAuthorizers` is a static table; a future stream RPC added to `MilvusService` without registering it here will be denied in every authorization-enabled deployment (with only a proxy-side warning). A guard test enumerating `MilvusService_ServiceDesc.Streams` would prevent silent drift.

7. **No rate limiting / access logging on stream endpoints (intentional).** The intermediate revision added both and the final one removed them with a documented rationale. Residual: the replication write path (`CreateReplicateStream`, any role holding the Global grant) and raw-WAL export (`DumpMessages`) are unthrottled and unaudited on the external port, unlike the unary data path. Fine as a deliberate trade-off; worth tracking as a follow-up (per-stream rate limiting + stream access log).

Note: with authorization **disabled** (the default), the stream chain is a pass-through and none of the above applies — the risks are specific to deployments that enable `AuthorizationEnabled`.

bigsheeper:

Verified:
- `internal/distributed/proxy/service.go:400` — `MilvusService` (including `DumpMessages` and `CreateReplicateStream`) is registered only on the external gRPC server, so moving authorization out of the handler leaves no un-intercepted entry point.
- `internal/distributed/proxy/service.go:430-436` — the stream interceptor chain orders authentication → RBAC → logging, matching the unary chain at `service.go:315-318`.
- `internal/proxy/authentication_interceptor.go:42-62` — authentication failures stop before the handler, and the authenticated context is propagated to the RBAC interceptor and handler via the wrapped `ServerStream`.
- `internal/proxy/privilege_interceptor.go:943-958` — the stream privilege interceptor is fail-closed: any method absent from the read-only `streamMethodAuthorizers` map is denied, and only the exact health-watch method receives the documented exemption.
- `client/milvusclient/interceptors.go:133-165` — stream calls carry the same authorization, database, identifier, and request metadata as unary calls, with shared client connection state read under the existing lock.
- `internal/proxy/dump_messages_authz_test.go:432-468` — coverage asserts the guardrail that a replication-only role is denied WAL dump, alongside fail-closed dispatch and full-chain ordering tests.

Issue #52610, opened 2026-08-18
[Enhancement]: Support configurable sentence split chars for lindera tokenizer


aoiasd:
## Is your feature request related to a problem?

The lindera tokenizer pre-splits input on Japanese punctuation (`。`, `、`, `\n`, `\t`) before running the lattice. When a user dictionary entry contains one of these characters (e.g. `結巴、分詞器` contains `、`), the input is split at that character first, so the user dict token can never match as a whole.

## Describe the solution you'd like

Add an optional `sentence_split_chars` parameter to the lindera tokenizer so users can override (or empty) the pre-split character set:

```json
{
  type: lindera,
  dict_kind: ipadic,
  sentence_split_chars: [],
  user_dict: {type: local, path: /path/to/user_dict.csv}
}
```

An empty list disables the pre-splitting, letting user dict entries that contain punctuation match as a single token. The default (`['。', '、', '\n', '\t']`) preserves the existing behavior.

PR #52611, opened 2026-08-18
enhance: support configurable sentence split chars for lindera tokenizer

aoiasd:
relate: #52610

## Summary
Allow Lindera tokenizer users to configure sentence pre-split characters while preserving the existing defaults. An empty list disables pre-splitting so user-dictionary entries containing punctuation can match as a single token.

### DIFF
--- a/internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/lindera_tokenizer.rs
+++ b/internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/lindera_tokenizer.rs
@@ -41,7 +41,13 @@ pub struct LinderaSegmenter {
     /// If provided, this dictionary will be used in addition to the default dictionary to improve
     /// the accuracy of segmentation for specific words or phrases.
     pub user_dictionary: Option<Arc<UserDictionary>>,
+
+    /// Characters used to split input before running the lattice.
+    /// Delimiters are retained so token offsets stay aligned with the input.
+    pub sentence_split_chars: Arc<[char]>,
 }
+
+const DEFAULT_SENTENCE_SPLIT_CHARS: &[char] = &['。', '、', '\n', '\t'];
 
 impl LinderaSegmenter {
     /// Creates a new instance with the specified mode, dictionary, and optional user dictionary.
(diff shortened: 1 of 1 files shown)


internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/lindera_tokenizer.rs
```diff
@@ -64,8 +72,8 @@ impl LinderaSegmenter {
         let mut position = 0_usize;
         let mut byte_position = 0_usize;
 
-        // Split text into sentences using Japanese punctuation.
-        for sentence in text.split_inclusive(&['。', '、', '\n', '\t']) {
+        for sentence in text.split_inclusive(|ch| self.sentence_split_chars.as_ref().contains(&ch))
```
buqian-zilliz, line 75: **[P3] Disabling sentence pre-splitting feeds an entire document into one Viterbi lattice**

With `sentence_split_chars: []` (or any set that drops `\n`), `segment` no longer pre-splits the input, so a whole document is passed to `Lattice::set_text`/`calculate_path_costs` as a single sentence. `Lattice::set_capacity` then resizes `starts_at` and `ends_at` (both `Vec<Vec<EdgeId>>`) to `text_len + 1`, and the path-cost pass scales with document length and edges per position. For production-sized text fields this can mean large transient memory and time per document, with no size guard. This is an explicit opt-in, but it may be worth documenting a bound or recommending keeping at least `\n` in the split set to cap per-lattice size.

aoiasd: Thanks for pointing this out. We will document it as a usage caveat and recommend keeping a delimiter such as `\n` for large text fields, but we will not reject or prohibit an empty `sentence_split_chars` configuration.

buqian-zilliz, line 75: **[P3] Per-character linear scan over an unbounded delimiter list on the tokenization hot path**

The new closure `|ch| self.sentence_split_chars.as_ref().contains(&ch)` is evaluated once per input character by `str::split_inclusive`, and `contains` over `Arc<[char]>` is a linear scan of the full delimiter list. `sentence_split_chars` is user-configurable and not bounded, so a large configuration turns tokenization into `O(text_len × split_chars_len)` before the lattice even runs. The crate already imports `std::collections::HashSet`; consider storing the delimiters in a `HashSet<char>` (or capping the list size) so per-character lookup stays O(1).

aoiasd: We intentionally keep this configuration unbounded to preserve user flexibility. A large delimiter list may reduce tokenization and indexing throughput, but that is an explicit tradeoff users can make when choosing the configuration, so we do not plan to impose a hard limit.

zhengbuqian (approved):

zhengbuqian (approved):

PR #52613, opened 2026-08-18
fix: [2.6] avoid repeated RBAC backup grantee scans (#52198)

shaoting-huang:
Cherry-pick from master
pr: https://github.com/milvus-io/milvus/pull/52198
related to #52197

Release-branch adaptation:
- 2.6 keeps the legacy 16-hex grantee ID test key format (no trailing `/` on LoadWithPrefix keys), so the regression mock retains `crypto.MD5(fmt.Sprintf("%s/%s", key, ...))` while absorbing the new `"invalid"` loaded-grantee data case.
- `mlog.Error`/`mlog.String`/`mlog.Err` (master-only v3 logging API) converted to `log.Ctx(ctx).Error` + `zap.String`/`zap.Error`, the 2.6 logging style.

### DIFF
--- a/internal/metastore/kv/rootcoord/kv_catalog.go
+++ b/internal/metastore/kv/rootcoord/kv_catalog.go
@@ -1465,7 +1465,21 @@ func (kc *Catalog) loadGranteeIDPrefix(ctx context.Context, tenant string, grant
 	return kc.loadGranteeIDPrefixWithLoadedGrantees(ctx, tenant, granteeKey, idStr, nil, nil)
 }
 
+func validateLoadedGrantees(granteeKeys []string, granteeValues []string) error {
+	if (granteeKeys == nil) != (granteeValues == nil) {
+		return merr.WrapErrServiceInternalMsg("loaded grantee keys and values must both be nil or both be non-nil, got keys=%d values=%d", len(granteeKeys), len(granteeValues))
+	}
+	if granteeKeys != nil && len(granteeKeys) != len(granteeValues) {
+		return merr.WrapErrServiceInternalMsg("loaded grantee keys and values must have equal length, got keys=%d values=%d", len(granteeKeys), len(granteeValues))
+	}
+	return nil
+}
+
 func (kc *Catalog) loadGranteeIDPrefixWithLoadedGrantees(ctx context.Context, tenant string, granteeKey string, idStr string, granteeKeys []string, granteeValues []string) ([]string, []string, string, error) {
+	if err := validateLoadedGrantees(granteeKeys, granteeValues); err != nil {
+		return nil, nil, "", err
+	}
+
 	var firstPrefix string
 	newID := crypto.GranteeID(granteeKey)
 	if isLegacyGranteeID(idStr) && idStr != newID {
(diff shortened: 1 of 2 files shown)


internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -1647,7 +1734,7 @@ func (kc *Catalog) ListGrant(ctx context.Context, tenant string, entity *milvusp
 		if dbName != entity.DbName && dbName != util.AnyWord && entity.DbName != util.AnyWord {
 			return nil
 		}
-		keys, values, granteeIDKey, err := kc.loadGranteeIDPrefix(ctx, tenant, granteeKey, v)
+		keys, values, granteeIDKey, err := kc.loadGranteeIDPrefixWithLoadedGranteeProvider(ctx, tenant, granteeKey, v, loadedGrantees)
```
buqian-zilliz, line 1737: **[P3] `loadGranteeIDPrefix` is now dead code**

After switching `ListGrant`'s `appendGrantEntity` to call `loadGranteeIDPrefixWithLoadedGranteeProvider`, the four-argument `loadGranteeIDPrefix` (around line 1464) no longer has any callers in the package. Its last caller was the old `loadGranteeIDPrefix` call inside `ListGrant`. Consider removing it to avoid keeping an unused wrapper that still passes `nil, nil` through to `loadGranteeIDPrefixWithLoadedGrantees`.

internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -1638,6 +1713,18 @@ func (kc *Catalog) AlterGrant(ctx context.Context, tenant string, entity *milvus
 }
 
 func (kc *Catalog) ListGrant(ctx context.Context, tenant string, entity *milvuspb.GrantEntity) ([]*milvuspb.GrantEntity, error) {
+	return kc.listGrantWithLoadedGrantees(ctx, tenant, entity, nil, nil)
+}
+
+func (kc *Catalog) listGrantWithLoadedGrantees(ctx context.Context, tenant string, entity *milvuspb.GrantEntity, granteeKeys []string, granteeValues []string) ([]*milvuspb.GrantEntity, error) {
```
buqian-zilliz, line 1719: **[P3] `listGrantWithLoadedGrantees` is a needless passthrough with always-nil parameters**

`listGrantWithLoadedGrantees` is only called from `ListGrant` with `nil, nil`, so its `granteeKeys`/`granteeValues` parameters are never exercised; the function only validates nil input and delegates to `listGrantWithLoadedGranteeProvider`. `ListGrant` could call `listGrantWithLoadedGranteeProvider(ctx, tenant, entity, nil)` directly. Since no caller currently supplies preloaded grantees, this extra layer adds an API surface with dead parameters.

internal/metastore/kv/rootcoord/kv_catalog.go
```diff
@@ -1474,7 +1538,18 @@ func (kc *Catalog) loadGranteeIDPrefixWithLoadedGrantees(ctx context.Context, te
...
+		if loadedGrantees != nil {
+			var (
+				granteeKeys   []string
+				granteeValues []string
+			)
+			granteeKeys, granteeValues, err = loadedGrantees(ctx)
+			if err != nil {
+				return nil, nil, granteeIDKey, err
+			}
+			if err := validateLoadedGrantees(granteeKeys, granteeValues); err != nil {
```
buqian-zilliz, line 1550: **[P3] `validateLoadedGrantees` is invoked twice for the same data**

Both `newLoadedGranteesProvider` and `newLazyLoadedGranteesProvider` already call `validateLoadedGrantees` before returning a provider, and the lazy provider additionally normalizes nil slices to empty slices. The consumer then re-validates the returned slices at this call site. This is redundant (the checks are O(1) length comparisons) and reads as if the consumer cannot trust the provider. Consider dropping the second validation or documenting why the defensive re-check is required.

Issue #52614, opened 2026-08-18
[Bug]: client telemetry reports things that are not true: empty command list, false push_config success, and bursty sampling


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

- Milvus version: master (3.0)
- Deployment mode(standalone or cluster): both
- SDK version: in-tree Go client (the only SDK with telemetry today)
- Others: telemetry page at `/webui/telemetry.html`

### Current Behavior

Three separate faults, all of which make the telemetry surface state that does not match reality.

**1. The "Active Commands" panel is always empty.** The page calls `GET /_telemetry/commands`; no route ever served it, so the fetch fails and the panel falls back to "No active commands" whatever is actually pending. The coordinator has been able to produce the list since the feature landed -- `TelemetryManager.ListAllCommands` -- but nothing can reach it: `ClientTelemetryService` has no list RPC, so that method has no caller anywhere outside rootcoord.

**2. push_config reports success for payloads the client ignored.** `handlePushConfig` unmarshals into `ConfigPayload`; `encoding/json` drops unknown fields, so a payload of `{"unsupported_probe": true}` sets nothing and still returns `Success: true`. A misspelled key, a key belonging to a newer client, or `ttl_seconds` -- which sits in `ConfigPayload` and is sent by the UI but has never been read by anything -- are all indistinguishable from a config that took effect. The UI compounds it by toasting success on the coordinator's HTTP 200, without waiting for the client's reply at all.

**3. Sampling is bursty, not uniform.** `shouldSample` keeps a counter modulo 10000 and samples while the remainder is below `rate*10000`, so the sampled operations form one contiguous run per cycle. A cycle is 10000 operations; a heartbeat window is tens or hundreds. Every window therefore comes out wholly sampled or wholly dropped:

| client | window | one cycle | observed |
|---|---|---|---|
| 20 QPS, 5s heartbeat | 100 ops | 500 s = 100 windows | 24 windows at 100%, 74 at 0%, 2 partial |
| 3 QPS, 5s heartbeat | 15 ops | 3333 s = 667 windows | 166 at 100%, 499 at 0%, 1 partial |

[...]

PR #52615, opened 2026-08-18
fix: make client telemetry report what actually happened, and sample evenly

xiaofan-luan:
## Summary

issue: #52614

Three faults in client telemetry, all of the same kind: the surface reported something that was not true.

### The command list was never served

The page calls `GET /_telemetry/commands`. No route served it, so "No active commands" appeared whatever was pending. The coordinator has been able to produce that list since the feature landed -- `TelemetryManager.ListAllCommands` -- but nothing could reach it, and grepping shows the method has no caller outside rootcoord: `ClientTelemetryService` has no list RPC.

`ListClientCommands` is added to `pkg/proto/root_coord.proto` -- in-repo, so no milvus-proto change -- and wired through mixcoord to a proxy handler. The response carries each command's payload size rather than its payload: a persistent config holds whatever an operator put in it, and a list does not need it.

### push_config reported success for payloads the client ignored

`encoding/json` drops unknown fields, so `{"unsupported_probe": true}` set nothing and still returned `Success: true`. So did a misspelled key, a key belonging to a newer client, and `ttl_seconds` -- which sits in `ConfigPayload` and is sent by the UI but has never been read by anything. The UI compounded it by reporting success on the coordinator's HTTP 200, without waiting for the client's reply at all.

Replies now carry `{"applied": [...], "ignored": [...]}` and the UI waits for the reply before saying anything. Unknown keys stay non-fatal: failing the whole command would stop a newer server from configuring an older client at all.

### Sampling was bursty

`shouldSample` kept a counter modulo 10000 and sampled while the remainder was below `rate*10000`, so sampled operations formed one contiguous run per cycle. A cycle is 10000 operations; a heartbeat window is tens or hundreds. Every window came out wholly sampled or wholly dropped -- at 3 QPS, fourteen minutes of full metrics followed by forty-one reporting nothing from a client that never stopped working. The long-run ratio was right; the ratio inside a window, which is the only unit the API reports, never was.

The rate is accumulated instead, sampling on the operation that carries the accumulator past a whole unit: 0.25 samples every fourth operation, and any stretch holds the ratio. The accumulator is shared, but `atomic.AddUint64` hands each caller a distinct interval and the step is under one unit, so every crossing is observed exactly once -- the count is exact rather than statistical, and no random source is needed.

Two smaller faults went with it: a rate below 1e-4 truncated to a threshold of zero and then sampled nothing, so a configured rate silently meant "off"; and lowering the rate from 1.0 began with a burst of full sampling, because that path returns before touching the counter and left it at zero.

Per-window counts at 3 QPS, rate 0.25, before and after:

```
before: [15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15]   (then zero for 41 minutes)
after:  [ 3,  4,  4,  4,  3,  4,  4,  4,  3,  4,  4,  4]
```

## Test plan

New tests, each written to fail first against the old behaviour:

- `TestSamplingIsUniformWithinEveryWindow` -- ratio holds in every window of a run; 0.25 samples exactly every fourth operation; 1e-5 still samples; boundary rates unchanged; lowering from 1.0 produces no burst (this case failed at 100 of 100 before); 8 concurrent goroutines still see the exact ratio.
- `TestPushConfigReplyReportsAppliedAndIgnoredKeys` -- applied keys named, unknown keys reported without failing the command, a payload of only unknown keys changes nothing, `ttl_seconds` reported as ignored.
- `TestListTelemetryCommandsHandler` -- outstanding commands returned with payload size not payload, empty result is `[]` not null, coordinator failure surfaces as an error.

```
go test ./client/milvusclient/ -race                  ok  20.5s
go test ./internal/rootcoord/... -count=1             ok
go test ./internal/proxy/ -run Telemetry              ok
```

`internal/proxy` as a whole fails one unrelated case locally, `TestHighlighter/TestHighlighterInterface_LexicalHighlighter`, which reproduces identically on a clean master checkout in the same environment.

Sampling fault aside, note for whoever picks up the SDK telemetry branches: pymilvus, Java, C++ and Node carry the same modulo algorithm verbatim and will need the same change before they merge.


xiaofan-luan:
Folded into #52597 -- all of this is one story about the telemetry surface reporting things that were not true, and they share the integration suite that covers it.

Note for anyone who pulled this branch: it briefly carried `configs/user.yaml`, a local run config with embedded etcd that was committed by accident. It would have broken any distributed deployment reading configs/. It is gone from the combined branch's history.

PR #52616, opened 2026-08-18
test: avoid float-int collision in json_contains_any test data

### LINKED ISSUE #52594: [Bug]: Flaky test: `test_milvus_client_query_expr_all_datatype_json_contains_any` fails with 1501 == 1500 when random `_id` equals an exactly-rounded `i*1.1` product (e.g. 1190*1.1 == 1309.0)
## Environment

- **Milvus Version**: master @ 43f7c30963

### Option B: Steps (for systemic / stability / timing issues)

1. Deploy: any standalone deployment (deployment mode, MQ, and concurrency are irrelevant to this failure — it is pure test-data arithmetic).
2. Setup: run the existing test case as-is; it creates a collection with `default_nb = 3000` rows where the JSON field contains `listMix = [i, i*1.1, str(i), bool(i%2), [i, str(i)]]` (`tests/python_client/milvus_client/test_milvus_client_query.py:2178`).
3. Workload: to make the failure deterministic, replace the random `_id = random.randint(limit, default_nb - limit)` at `test_milvus_client_query.py:2187` with a fixed `_id = 1309`.
4. Trigger: run
   ```
   pytest "milvus_client/test_milvus_client_query.py::TestMilvusClientQueryValid::test_milvus_client_query_expr_all_datatype_json_contains_any" --host localhost --port 19530
   ```
   (originally observed on the `[True]` parametrization: `tests/python_client/milvus_client/test_milvus_client_query.py::TestMilvusClientQueryValid::test_milvus_client_query_expr_all_datatype_json_contains_any[True]`).
5. Observed via: pytest assertion failure — the `json_contains_any(json_field['listMix'], [1309, True])` query returns 1501 rows instead of the expected `3000 // 2 = 1500`.

### Trigger Conditions

- **Frequency**: intermittent with the unmodified test — the only necessary condition is that the random `_id` lands on a value that some other row's `i*1.1` product rounds to exactly under IEEE 754 (here `_id = 1309`, hit by row `i = 1190` since `1190 * 1.1 == 1309.0`). Estimated at roughly 2–4% per run (odd multiples of 11 in the `randint` range ≈ 4.5%, times the fraction that round exactly); this estimate is inferred from arithmetic, not measured by rerunning. With `_id` fixed to 1309 the failure should be deterministic (also inferred from the arithmetic; not verified by an actual rerun).
- **First observed after**: a single run of the e2e suite; no load or duration component.
- **Does NOT happen when**: the random `_id` does not collide with any exactly-rounded `i*1.1` product — e.g. the sibling `[False]` parametrization of the same test passed in the same run purely because it drew a different `_id`. Deployment mode, MQ, and concurrency have no effect.

## Expected Behavior

The test expects `json_contains_any(json_field['listMix'], [_id, bool(_id % 2)])` to match exactly `default_nb // 2 = 1500` rows (all odd rows, whose `listMix` contains `True`). The implicit assumption at `test_milvus_client_query.py:2216-2219` is that the int element `_id` only hits rows that are already in the `True` set.

## Actual Behavior

The query returns 1501 rows. Row 1190 (an even row, not in the `True` set) has `1190 * 1.1` in its `listMix`, and under IEEE 754 double arithmetic this product rounds exactly to `1309.0` (`double(1.1) = 1.1 + 8.8818e-17`; the exact product is `1309 + 1.0569e-13`, which is within half an ulp of 1309, so it rounds to `1309.0`). Segcore's `json_contains_any` intentionally matches a JSON double with integer value against an int64 query element (JSON does not distinguish `1309` from `1309.0`), so row 1190 legitimately satisfies the predicate, giving 1500 + 1 = 1501 rows. The server result is correct for the data that was inserted.

## Error Logs

No server-side error; the query result is correct for the inserted data and only the test assertion fails:

```
assert 1501 == (3000 // 2)
```

## Non-default Configuration

All defaults.
## Analysis Hints (Optional)

- **This is a test defect (test-data/test-assumption bug), not a product bug.** The server-side int↔double tolerant comparison in `internal/core/src/exec/expression/JsonContainsExpr.cpp:2077-2089` (`ExecJsonContainsWithDiffType`, mixed-type path) — with the same tolerant match in the same-type path at `:516-527` — is intentional design: any correct implementation must return 1501 rows given this data and `_id = 1309`.
- Root cause in the test: `tests/python_client/milvus_client/test_milvus_client_query.py:2178` builds `listMix` with `i*1.1`, and the assertion at `:2216-2219` hardcodes `default_nb // 2` on the assumption that the int element cannot match a float-derived value.
- Ruled out: a server-side "duplicate/extra row" explanation — an earlier query in the same test on `listBool` with `[True]` returned exactly 1500 rows (assertion at `:2209` passed), so the extra row genuinely exists and genuinely satisfies the predicate.
- Inferred (not directly proven): that the extra row is specifically row 1190 — the failure output was truncated to the first row, but `i = 1190` is the only value in the `randint` range whose `i*1.1` equals 1309 numerically, and the count matches (1500 + 1). To confirm on reproduction, query `json_contains_any(listMix, [1309]) && int64 % 2 == 0`; it should return exactly one row with `int64 = 1190`. The IEEE 754 result itself is verifiable with `python3 -c "print(1190*1.1 == 1309.0)"`.
- Prior art: the same file already switched `:2037` to `i * 1.111`, indicating this pitfall was hit before and only partially fixed; `:2109`, `:2178`, and `:2478` still use `i*1.1` and carry the same risk (including `test_milvus_client_query_expr_all_datatype_json_contains` and the `mix_data` case).
- Suggested fix (in the tests): replace the float element with a construction that can never equal an integer, e.g. `i + 0.5` (or at minimum align with `:2037` and use `i * 1.111`), in all three places (`:2109`, `:2178`, `:2478`). A more robust alternative is to compute the expected row count in Python using the same numeric semantics instead of hardcoding `default_nb // 2`.

yanliang567:
## Related Issue

Fixes #52594

## What type of PR is this?

- [x] test

## What this PR does

Fixes a flaky test `test_milvus_client_query_expr_all_datatype_json_contains_any` that intermittently fails with `1501 == 1500`.

## Root cause

The JSON field's `listMix` float element was built as `i * 1.1`. Under IEEE 754 double arithmetic, some products round exactly to an integer (e.g. `1190 * 1.1 == 1309.0`). When the random `_id` lands on such a value, segcore's `json_contains_any` (intentionally) matches the JSON double `1309.0` against the int64 query element, so an extra even row satisfies the predicate and the hardcoded `default_nb // 2` assertion fails.

This is a test-data defect, not a product bug — the server result is correct for the inserted data.

## Changes

Replace `i * 1.1` with `i + 0.5` (which can never equal an integer) in the three affected spots in `tests/python_client/milvus_client/test_milvus_client_query.py`:
- `listMix` in `test_milvus_client_query_expr_list_all_datatype_json_contains_all`
- `listMix` in `test_milvus_client_query_expr_all_datatype_json_contains_any` (the flaky test)
- `mix_data` in `test_milvus_client_query_expr_list_all_datatype_json_contains_any`

## Verification

- Reproduced before the fix: `json_contains_any(listMix, [1309, True])` returned 1501 rows (extra even row `int64=1190`).
- After the fix, forcing the previously-colliding `_id=1309` returns exactly 1500 rows.
- `test_milvus_client_query_expr_all_datatype_json_contains_any` passed 5 consecutive rounds (10 parametrized runs).
- The other two touched tests passed as well (2 + 4 parametrized runs).

### DIFF
--- a/tests/python_client/milvus_client/test_milvus_client_query.py
+++ b/tests/python_client/milvus_client/test_milvus_client_query.py
@@ -2106,7 +2106,7 @@ class TestMilvusClientQueryValid(TestMilvusClientV2Base):
                 "listFlt": [m * 1.0 for m in range(i, i + limit)],
                 "listBool": [bool(i % 2)],
                 "listList": [[i, str(i + 1)], [i * 1.0, i + 1]],
-                "listMix": [i, i * 1.1, str(i), bool(i % 2), [i, str(i)]],
+                "listMix": [i, i + 0.5, str(i), bool(i % 2), [i, str(i)]],
             }
         self.insert(client, collection_name, rows)
         # 3. create index and load
@@ -2175,7 +2175,7 @@ class TestMilvusClientQueryValid(TestMilvusClientV2Base):
                 "listFlt": [m * 1.0 for m in range(i, i + limit)],
                 "listBool": [bool(i % 2)],
                 "listList": [[i, str(i + 1)], [i * 1.0, i + 1]],
-                "listMix": [i, i * 1.1, str(i), bool(i % 2), [i, str(i)]],
+                "listMix": [i, i + 0.5, str(i), bool(i % 2), [i, str(i)]],
             }
         self.insert(client, collection_name, rows)
         # 3. create index and load
@@ -2475,7 +2475,7 @@ class TestMilvusClientQueryValid(TestMilvusClientV2Base):
         flt_data = [[m * 1.0 for m in range(i, i + limit)] for i in range(default_nb)]
         bool_data = [[bool(i % 2)] for i in range(default_nb)]
         list_data = [[[i, str(i + 1)], [i * 1.0, i + 1]] for i in range(default_nb)]
-        mix_data = [[i, i * 1.1, str(i), bool(i % 2), [i, str(i)]] for i in range(default_nb)]
+        mix_data = [[i, i + 0.5, str(i), bool(i % 2), [i, str(i)]] for i in range(default_nb)]
 
         for i in range(default_nb):
             rows[i][ct.default_json_field_name] = {


wangting0128 (approved):
/lgtm

PR #52621, opened 2026-08-18
feat: add segment-scoped manifest commit

somfornot:
issue: #51723

design: docs/design-docs/design_docs/20260817-datacoord-segment-manifest-commit.md

## Summary

Add one DataCoord-owned primitive, `CommitSegmentManifest`, that serializes
StorageV3 manifest publication per segment: create the manifest revision outside
`segMu`, then rebase the catalog mutation onto the latest `SegmentInfo` and
publish the new `manifest_path` together with the associated segment/stats change
in one catalog transaction under `segMu`. Manifest object-storage I/O never holds
`segMu`; the final full-record publication does, matching the existing
`UpdateSegmentsInfo` consistency model.

Existing writers (flush/`SaveBinlogPaths`, import, copy, stats text/json/sort, L0
compaction, batch manifest DDL) publish through the framework; StorageV3 direct
`UpdateManifest`/`UpdateManifestVersion` is rejected. Producers that still return
a prepared manifest use the temporary `Noop` adapter.

## Notes

- L0 commits filter deltalogs already registered on the in-memory segment, so a `saveSegmentMeta` retry after a failed `meta_saved` task-state write (in-memory Deltalogs already reflect the committed manifest) is idempotent — it re-reads the current entries and appends nothing (mirrors the catalog-side dedup). This does **not** cover the window where `catalog.Update` fails after the loon manifest transaction commits: the retry then reads stale in-memory Deltalogs and the filter is blind, so it can re-append. A durable fix (persisted deltalog identity / key-based manifest add) is tracked as a follow-up.
- Exact `ExpectedManifest` conflicts and version regressions are classified as a
  stale result so the stats scheduler discards obsolete worker output instead of
  retrying it.
- External collection refresh migration is intentionally deferred (see the design
  doc follow-up section); its job-level publication is unchanged.


internal/datacoord/task_stats.go
```diff
@@ -578,6 +616,44 @@ func (st *statsTask) SetJobInfo(ctx context.Context, result *workerpb.StatsResul
...
+func (e staleStatsResultError) Error() string        { return e.cause.Error() }
+func (e staleStatsResultError) Unwrap() error        { return e.cause }
+func (e staleStatsResultError) Is(target error) bool { return target == errStatsResultStale }
+
+// shouldPublishPreparedManifest identifies the temporary compatibility path
+// for workers which still return a prepared stats manifest, including the
+// first manifest. The Noop adapter keeps pointer publication serialized while
+// a follow-up changes workers to return structured deltas.
+func (st *statsTask) shouldPublishPreparedManifest(ctx context.Context, segmentID int64, result *workerpb.StatsResult) bool {
+	segment := st.meta.GetSegment(ctx, segmentID)
```
liliu-z, line 774: `GetSegment` doesn't filter dropped segments, so a segment retired by compaction while its stats task was still running routes through here, and `CommitSegmentManifest` then rejects it on the `isSegmentHealthy` check with a plain internal error.

That error is neither `errStatsResultStale` nor `errStatsResultDiscarded`, so `QueryTaskOnWorker` falls through to `if err != nil { return }` and the task stays InProgress. Nothing clears it after that: `check()` has no InProgress branch, `CanCleanedTasks` only reaps Finished/Failed, and `statsInspector.DropStatsTask` has no callers. So it re-polls every tick, and the worker-side stats files are never cleaned up.

`updateStatsResultIfManifestMatches` returned `errStatsResultDiscarded` for exactly this case. Adding `isSegmentHealthy` here, or mapping the unhealthy-segment rejection back onto `errStatsResultDiscarded`, would restore it.

internal/datacoord/ddl_callbacks_batch_update_manifest.go
```diff
@@ -44,7 +46,28 @@ func (c *DDLCallbacks) batchUpdateManifestV2AckCallback(ctx context.Context, res
...
+				continue
+			}
+			basePath, currentVersion, err := packed.UnmarshalManifestPath(segment.GetManifestPath())
+			if err != nil {
+				return err
+			}
+			if item.GetManifestVersion() <= currentVersion {
+				continue
+			}
+			if err := c.meta.CommitSegmentManifest(ctx, SegmentManifestCommit{
```
liliu-z: A V3 item here is just a version number, so the mutation is `Noop` and does no object-storage I/O. There's nothing to move out of `segMu`, which is what the framework buys you.

What does change: a broadcast carries up to 512 items, and each one now takes `segMu` twice and issues its own `catalog.Update`, where the batched `UpdateSegmentsInfo` collapsed them into a single `AlterSegments`. An error mid-loop also returns before the V2 operators run, so the batch is no longer applied as a unit.

`UpdateManifestVersion` already did this monotonicity check atomically under `segMu`. Is there a race it doesn't cover?

somfornot: Good catch — all correct: for a V3 item the mutation is a pure version bump (`Noop`, no manifest I/O), so per-item `CommitSegmentManifest` buys nothing here while costing `segMu` twice + a `catalog.Update` per item, and a mid-loop error returns before the accumulated V2 operators run, so the batch is no longer applied as a unit.

For now I've reverted this path back to the batched `UpdateManifestVersion` operator so the whole broadcast (V2 column groups + V3 version bumps) commits atomically in a single `UpdateSegmentsInfo` / `AlterSegments`, as before. (This also required dropping the `UpdateManifestVersion` StorageV3 guard this PR had added.)

The honest tension: this batch-level atomicity conflicts with the per-segment single-writer atomicity `CommitSegmentManifest` is meant to introduce — funneling every V3 pointer move through the per-segment commit lock is exactly what serializes against concurrent writers (stats/index/GC/compaction) racing the manifest, but it can't be done per-item here without giving up the single-transaction batch. Same unresolved conflict as the external collection refresh path. I've left a TODO to revisit with a batched `CommitSegmentManifests` primitive that keeps the single-writer invariant without sacrificing single-transaction batching. Thanks!

czs007:

Adversarial review of the StorageV3 L0 manifest-commit path surfaced one correctness regression in the L0 result-save flow plus two follow-on issues around commit batching and cross-restart idempotency.

## High

**`internal/datacoord/compaction_task_l0.go:477` — L0 commit path does not tolerate `ErrSegmentNotFound`, wedging the task forever on a dropped delta target**

The new V3 branch in `saveSegmentMeta` bare-propagates the error from `commitL0V3Deltalogs` (`compaction_task_l0.go:479-481`). `meta.GetSegment` returns `Dropped` segments verbatim (`meta.go:780-784`) and the branch has no `isSegmentHealthy` gate, so a dropped V3 target that still carries a manifest enters the branch and `CommitSegmentManifest` rejects it with `merr.WrapErrSegmentNotFound` at its pre-I/O and post-I/O health checks (`segment_manifest_commit.go:156-163`, `:192-195`). The early `return err` skips the input-segment retirement operators appended after the output loop (`UpdateStatusOperator(Dropped)` / `UpdateCompactedOperator`, `compaction_task_l0.go:493-494`), and `QueryTaskOnWorker` only logs and returns (`:169-172`), leaving the task in `executing` with no DataCoord-side timeout for that state. Nothing guards reachability: targets are picked by `selectFlushedSegment` with no `isCompacting`/exclusivity filter (`:309-316`), and `ValidateSegmentStateBeforeCompleteCompactionMutation` validates only the *input* segments (`meta.go:2765-2780`), with its L0 early-out at `meta.go:2733` skipping the surrounding block entirely. This is a regression: the legacy path reached the target through `updateSegmentPack.Get`, which accepts `Dropped` segments unchanged (`meta.go:953-969`), so the operator simply applied and the task completed.

Failure: a mix/sort compaction marks an L1 target `Dropped` while the datanode executes the L0 plan, and the worker then reports `completed` with deltalogs for that segment → `saveSegmentMeta` returns early forever, the L0 inputs are never marked `Dropped`/`Compacted`, the task never leaves `executing`, and the scheduler re-polls the identical permanent error every tick, pinning the L0 segments as compacting and leaking the compaction slot until DataCoord restarts (which reproduces the same loop).

Suggestion: gate the V3 branch on `isSegmentHealthy(current)` and `continue`, or swallow `ErrSegmentNotFound` from `commitL0V3Deltalogs` as a benign skip, so the input-drop operators still run and the task reaches `meta_saved`. This is the contract the stats caller already honors (`task_stats.go:577`, `:607`), so the L0 caller only needs the same treatment rather than a new convention. *(raised by xiaocai2333, czs007)*

## Medium

**`internal/datacoord/compaction_task_l0.go:479` — V3 L0 targets are now committed one segment at a time, replacing a 16-way-parallel manifest commit plus a single batched etcd write with N serial transactions**

Previously every L0 output target went through `AddL0DeltalogsAndUpdateManifestOperator` into a single `UpdateSegmentsInfo`: manifest transactions ran through a `conc` pool sized by `dataCoord.compaction.levelzero.manifestUpdatePoolSize` (default 16, `meta.go:1363-1385`, `component_param.go:5950-5954`), and every touched segment plus the input-segment drops landed in one `catalog.AlterSegments` batch (`meta.go:2006-2013`). The new branch calls `commitL0V3Deltalogs` synchronously inside the output loop, and each `CommitSegmentManifest` issues its own loon transaction plus its own `catalog.Update`, which builds a fresh txn per call (`segment_manifest_commit.go:233-235`, `update.go:44-51`). An L0 plan targets every flushed L1/L2 segment on the channel/partition (`compaction_task_l0.go:309-316`), routinely hundreds.

Failure: an L0 delete compaction on a StorageV3 collection with a few hundred flushed targets → meta-save goes from `ceil(N/16)` parallel manifest commits plus one etcd write to N serial manifest commits plus N etcd writes on the scheduler goroutine, the pool-size knob becomes inert for V3, and a mid-loop failure leaves earlier targets published while the L0 inputs stay undropped.

Suggestion: restore a bounded-concurrency fan-out over the V3 targets, or add a batched `CommitSegmentManifests` primitive that preserves single-transaction publication. *(raised by czs007)*

**`internal/datacoord/compaction_task_l0.go:523` — L0 deltalog dedup guard is in-process only and does not survive a restart** *(provenance unclear — please confirm whether this PR introduces it)*

`filterDuplicateFieldBinlogs` keys on `current.GetDeltalogs()`, but V3 segments never persist their per-`FieldBinlog` arrays: `buildSegmentAndBinlogsKvs` skips the binlog KVs for manifest-backed segments and stores a `SegmentInfo` whose `Deltalogs` field was reset (`util.go:120-134`, `:146-151`), and `buildAlterSegmentsKvs` does the same (`kv_catalog.go:325-357`). After a restart the in-memory `Deltalogs` of a V3 segment is empty, so the filter is a no-op, `packed.CommitManifestUpdates` re-appends the identical entries (it only ever calls `loon_transaction_add_delta_log`, `manifest_commit.go:143-151`), and `addDeltalogsToSegment` re-accumulates `Stats.DeleteNumRows` / `DeltaBinlogSize` / `DeltaBinlogCount` (`meta.go:1216-1246`) — its own comment at `meta.go:1226-1231` asserts that each batch applies exactly once, which does not hold across a restart. Same-process retries remain safe: the `ExpectedManifest` CAS makes a concurrent loser fail with `staleSegmentManifestError`, and a sequential retry short-circuits through the filter.

Failure: `commitL0V3Deltalogs` publishes manifest v(N+1) and the catalog write succeeds, then DataCoord restarts before the task's `meta_saved` state write lands and the reloaded task re-polls the still-alive datanode for the same result → the same delta files are appended again as v(N+2) and the delete-accounting stats are permanently double-counted, feeding `hasTooManyDeletions`, `GetResidualSegmentSize` and view counts into spurious compaction triggers with no path that recomputes them from the manifest.

Suggestion: persist enough deltalog identity for V3 segments to make the filter durable, or make the manifest append key-based so a replayed commit is a no-op. Note the dedup gap is inherited rather than created here — on the merge base the legacy prepare path had no dedup at all and appended through the same `loon_transaction_add_delta_log`, and the non-persistence in `util.go` / `kv_catalog.go` is untouched by this diff — so please confirm the provenance before deciding whether it belongs in this PR or a follow-up. Note also that in the dropped-target wedge above the retry never converges regardless of restart, so that item takes precedence. *(raised by czs007, xiaocai2333)*

somfornot:
Thanks for the thorough adversarial pass — the High item and the batching regression are both real. Fixed in 222d4781d1:

**High — dropped delta target wedging the task: fixed.** `saveSegmentMeta` now gates the V3 branch on `isSegmentHealthy` and skips a retired target before any manifest I/O, and the commit fan-out additionally swallows an `ErrSegmentNotFound` raced in between the check and the commit — the same benign-terminal contract the stats caller honors in `SetJobInfo`, as suggested. The input-segment retirement operators always run now, so the task reaches `meta_saved`. Any other commit error (stale CAS conflict, catalog failure) still fails the save so the scheduler retries; stale conflicts self-heal on the next attempt via the fresh-pointer reread plus the idempotency filter. Tests pin all three behaviors: a dropped target never reaches `CommitSegmentManifest` and the inputs still retire; a not-found from the commit is swallowed; a real error propagates without retiring the inputs.

**Medium — serial per-segment commits: fixed (the fan-out half).** `commitL0V3DeltalogsParallel` restores bounded-concurrency publication over the V3 targets using the same `dataCoord.compaction.levelzero.manifestUpdatePoolSize` pool, so the knob is live for V3 again and the expensive half (the loon manifest transaction, outside `segMu`) overlaps as before, while per-segment `ExpectedManifest` CAS semantics stay unchanged. The catalog half deliberately remains one `catalog.Update` per segment for now — collapsing those into a single batched transaction without giving up the single-writer invariant is the batched `CommitSegmentManifests` follow-up already noted in `ddl_callbacks_batch_update_manifest.go` (same tension as the batch-DDL and external-refresh paths). A test pins genuine overlap: each mocked commit blocks until all targets have entered, so a serial implementation fails via timeout instead of passing.

**Medium — restart-durable dedup: confirmed inherited; deferring as follow-up.** Provenance is as you suspected: on the merge base the manifest append had no dedup at all (`l0ManifestUpdate.prepare` builds entries straight from the worker deltalogs; `filterDuplicateFieldBinlogs` existed only on the catalog half in `addDeltalogsToSegment`), and the V3 binlog-KV non-persistence (`isV3Segment` handling in metastore `util.go`) predates this PR. What this PR added is the manifest-side in-process filter, which strictly narrows the window — same-process retries are now idempotent (pinned by `TestL0CompactionV3ManifestCommitIsIdempotentOnRetry`) — while the cross-restart replay gap is inherited from the existing non-persistence design. Agreed it needs a durable fix (persisted deltalog identity, or a key-based manifest append so a replayed commit is a no-op); tracking it as a follow-up alongside the batched-commit primitive rather than expanding this PR.

internal/datacoord/compaction_task_l0.go
```diff
@@ -487,7 +518,120 @@ func (t *l0CompactionTask) saveSegmentMeta(outputSegs []*datapb.CompactionSegmen
...
+	// manifest transaction. The same L0 output reaches saveSegmentMeta twice
+	// whenever it re-runs (a failed catalog write, or a failed meta_saved
+	// task-state write after the catalog write succeeded). Unlike the catalog
+	// half, packed manifest commits append delta-log entries without any
+	// deduplication, so a blind re-commit would leave duplicate entries in the
+	// manifest and bump a fresh revision on every retry. Filtering by
+	// (fieldID, logID) makes the manifest re-commit idempotent, mirroring the
+	// catalog dedup in addDeltalogsToSegment; a full duplicate short-circuits
+	// before any object-storage I/O.
+	deltalogs = filterDuplicateFieldBinlogs(current.GetDeltalogs(), deltalogs)
```
yhmo, line 594: internal/datacoord/compaction_task_l0.go line:589
Medium ---- The retry-idempotency guarantee does not cover the catalog-write-failure window. This filter reads the in-memory Deltalogs, which CommitSegmentManifest only updates after its catalog write succeeds; if catalog.Update fails after the loon manifest transaction already committed, a saveSegmentMeta retry re-reads the unchanged segment (pointer still v, Deltalogs without the new entries), rebuilds the same delta entries, and re-runs CommitSegmentManifest at base v. The PR Notes claim "a saveSegmentMeta retry cannot append duplicate manifest delta-log entries", but that holds only for the retry after a failed meta_saved task-state write. Whether loon's OVERWRITE resolver then rebases onto the already-committed v+1 (appending the entries a second time) or overwrites it is not determined by the Go code, so the categorical claim is not guaranteed for this window. A fault-injection test (catalog failure after the manifest commit, then retry) would pin the actual behavior.

internal/datacoord/compaction_task_l0.go
```diff
@@ -466,19 +469,47 @@ func buildL0V3DeltaLogEntries(segmentID int64, deltalogs []*datapb.FieldBinlog)
...
+			// per-segment lock while it performs object-storage I/O.
+			current := t.meta.GetSegment(ctx, seg.GetSegmentID())
+			if current != nil && current.GetStorageVersion() == storage.StorageV3 && current.GetManifestPath() != "" {
+				// A target retired by a concurrent compaction while the L0 plan
+				// was executing is gone for publication purposes: GetSegment
+				// returns dropped segments, and CommitSegmentManifest would only
+				// reject one with ErrSegmentNotFound. Skip it so the
+				// input-segment retirement below still runs and the task reaches
+				// meta_saved instead of re-polling a permanent error forever.
+				if !isSegmentHealthy(current) {
```
yhmo, line 486: internal/datacoord/compaction_task_l0.go line:488
Medium ---- Skipping deltalog publication to a dropped V3 target discards this task's input deletes. The worker output deltalogs for the target are the merge of this task's input L0 segments; those inputs are still retired below (Dropped + Compacted) even though their deletes were never published to any surviving segment. The concurrent compaction that dropped the target compacted the target's own data and cannot have consumed this task's inputs (they are marked compacting), so once both compactions complete the deletes are gone and the previously-deleted rows resurrect. Unless the scheduler guarantees those deletes are carried forward another way (e.g., the concurrent compaction also consumed the L0 inputs before this skip), please confirm the deletes are preserved, or fail the task instead of retiring the inputs.

internal/datacoord/compaction_task_l0.go
somfornot: **Provenance: this window predates the PR — master is strictly worse, and the filter here narrows it rather than introducing it.**

- **Native layer is identical on both sides.** master's L0 path (`AddDeltaLogsToManifestOverwrite`, `packed/transaction.go:74`) and this PR's `CommitManifestUpdates` (`packed/manifest_commit.go:113`) both do `loon_transaction_begin(..., LOON_TRANSACTION_RESOLVE_OVERWRITE, ...)` + one `loon_transaction_add_delta_log` per entry. OVERWRITE is the version-conflict resolver, not a per-deltalog dedup — deltalogs aren't keyed (`manifest_commit.go:106-111`: "delta-log adds don't collide on a key in the first place"). So your first branch is the operative one for a stale-base retry: it rebases and adds the entries a second time, it does not overwrite the earlier add. I'll scope the PR Notes to the meta_saved-retry rather than stating the idempotency guarantee categorically — you're right that the categorical wording overreaches.

- **master had no manifest-side dedup at all.** `l0ManifestUpdate.prepare` builds entries straight from the worker deltalogs and `commitManifest` appends; `filterDuplicateFieldBinlogs` lived only on the catalog half (`addDeltalogsToSegment`, `meta.go:1220` — Stats accounting), never on the manifest. So on master *every* retry re-appended — not just this catalog-failure window but also the ordinary meta_saved-failure retry (in-memory already current, yet re-appended anyway).

- **What this PR adds is exactly the l0.go:589 filter**, which makes the common same-process retry (catalog succeeded → in-memory current) idempotent — that's the meta_saved-failure case, now a no-op. The residual you've pinned — catalog fails *after* the manifest commit, so the retry reads stale in-memory Deltalogs and the filter is blind — is the one case it can't cover, and it's precisely the inherited master behavior. Same root as the cross-restart replay gap @czs007 raised: the dedup keys off non-durable in-memory state.

- **On pinning loon's rebase-vs-replace definitively:** agreed the Go code doesn't settle it. The resolve-mode + un-keyed `add_delta_log` evidence points to append-a-second-time, but the authoritative answer is in the loon native commit path, which I haven't traced here. Your fault-injection suggestion (catalog failure after the manifest commit, then retry) is the right way to pin it; tracking it with the durable-dedup follow-up (persisted deltalog identity, or a key-based manifest add so a replayed commit is a no-op) rather than expanding this PR.

internal/datacoord/compaction_task_l0.go
somfornot: **Traced this end-to-end — the deletes are preserved. The mix/sort-dropped-target path that would lose them isn't reachable, and the skip isn't a regression even if forced.**

You're right on the sharp part of the premise: I verified mix/sort do **not** re-apply un-distributed channel L0 deletes — `mixCompactionTask.BuildCompactionRequest` attaches only each input's own `segInfo.GetDeltalogs()` (`compaction_task_mix.go:403`; mix inputs are L1-only, `compaction_util.go:113-119` excludes L0/L2), and the datanode executor reads only the plan segment's own deltalogs (`compaction/common.go:63-82`). So a mix rebuild S→S' does not carry this task's deletes; correctness relies on L0 re-publishing them onto the *current* flushed target. The real question is whether L0 can retire its inputs while that target is a dead S.

**A mix/sort drop can't be the cause of the skip:**

1. **Targets are (re)selected at dispatch, not at trigger.** `selectFlushedSegment` runs inside `CreateTaskOnWorker` → `BuildCompactionRequest` (`compaction_task_l0.go:90-92`, `:384`) — *after* the scheduler admits the task — and selects only currently `Sealed/Flushed`, non-L0 segments (`:311-318`). A target a mix already dropped is not `Flushed`, so it's never selected; the mix's live replacement S' is, and the worker distributes the deletes onto S'. So "mix dropped S *before* this task runs" → the task targets S', not the dead S → no skip, no loss.
2. **Nothing can drop the selected target *during* execution.** From `selectFlushedSegment` through `saveSegmentMeta` the task is in `executingTasks` as an L0 task on the channel, and channel-level mutual exclusion fences Mix/Sort/Bump/Clustering off that channel for the whole duration (`compaction_inspector.go:262-289`; SortCompaction/L2 covered at `:233`,`:269`).

So the only way the skip fires with real un-published deletes is a **non-compaction** drop of the target mid-execution — DropCollection/DropPartition or GC of an already-dead segment — where the deletes are moot because the owning data is being removed anyway. A competing mix/sort can't produce the resurrection described.

**Not a regression even if the interleaving is forced.** On master the identical dropped-but-present target case is *write-to-dropped*, not skip: `AddL0DeltalogsAndUpdateManifestOperator` has no health/Dropped guard (`l0ManifestUpdate.prepare` → `Get` returns the dropped segment), so master writes the deltalogs onto the tombstoned target and retires the L0 inputs in the *same* `UpdateSegmentsInfo` batch. Net outcome is identical — deltalogs land on a segment no querynode loads and GC will delete, deletes lost, inputs retired either way. The must-fix changes only the mechanism (skip the dead-segment write vs. perform it), not the data outcome.

**Honest residual:** the safety rests on two scheduler-level invariants — dispatch-time reselection and coarse channel-level exclusion — not on anything local to `saveSegmentMeta`. If either were narrowed (selection moved to trigger time, or exclusion made sub-channel), this skip would need to become a task failure (as you suggest) so inputs aren't retired before their deletes are published. Happy to add a defensive comment at the skip site documenting that dependency — will do it in this PR if you'd like.

internal/datacoord/compaction_task_l0.go
yhmo: The PR Notes are now scoped correctly. However, the code comment directly above this filter (internal/datacoord/compaction_task_l0.go:579-588) still asserts the categorical claim: it lists "a failed catalog write" as a case where "Filtering by (fieldID, logID) makes the manifest re-commit idempotent". That is exactly the window you agreed the filter is blind to — after catalog.Update fails, the in-memory Deltalogs are stale (the in-memory update happens only after catalog.Update succeeds), so the new entries survive the filter and the retry re-commits them. Please scope the code comment to the meta_saved retry, matching the updated PR Notes.

internal/datacoord/compaction_task_l0.go
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

internal/datacoord/compaction_task_l0.go
somfornot: Done in 57ae098 — scoped the comment above the filter to the meta_saved retry, and spelled out that the catalog-write-failure retry is *not* covered: `CommitSegmentManifest` installs the in-memory Deltalogs only after its catalog write succeeds, so on that path they're stale and the new entries survive this filter. That window is noted as the durable-dedup follow-up, so the code comment now matches the PR Notes. Thanks for catching the mismatch.

yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/datanode/index/task_stats.go
```diff
@@ -647,10 +647,14 @@ func (st *statsTask) createTextIndex(ctx context.Context,
...
-	// AddStatsToManifest stores the manifest-relative representation at commit time.
-	if st.manifestPath != "" && len(textIndexLogs) > 0 {
+	// The Sort sub-job bakes text index stats into the freshly written target-segment
+	// manifest inline. A standalone TextIndexJob instead ships textIndexLogs to
+	// DataCoord, which runs the manifest transaction itself (CommitSegmentManifest,
+	// rebased onto the segment's current manifest); baking here would commit against a
+	// base that may already be stale, so the worker skips it. TextStatsLogs carries
+	// full object keys either way; the loon transaction stores the manifest-relative
+	// representation at commit time.
+	if st.req.GetSubJobType() == indexpb.StatsSubJob_Sort && st.manifestPath != "" && len(textIndexLogs) > 0 {
```
yhmo, line 657: internal/datanode/index/task_stats.go line:657
Low ---- This worker change (no longer baking standalone text/JSON stats into the V3 manifest) and the DataCoord half (commitTextIndexStats/commitJSONKeyStats running the manifest transaction) are two halves of one contract that must ship together. During a rolling upgrade, a new DataNode with an old DataCoord leaves result.GetManifest() at the base manifest, so the old updateStatsResultIfManifestMatches sees no manifest change, dual-writes the stats to etcd, and never registers them in the StorageV3 manifest - text-match/JSON-filter reads then miss the stats, and the etcd placeholder prevents the stats inspector from re-triggering. The reverse order orphans a worker-baked manifest revision. Is mixed-version DataNode/DataCoord a supported window for V3 collections? If so, consider documenting an upgrade order (DataCoord first) or gating the no-bake behavior on a version handshake.

internal/datacoord/segment_manifest_commit.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+	applySegmentCatalogTypedFields(segment, mutation)
+	if err := pack.Validate(); err != nil {
+		return nil, nil, err
+	}
+	// Operators prepare metric transitions as part of UpdateSegmentsInfo.
+	// Do this after applying the typed fields too, so a state mutation is
+	// reflected only once the catalog write succeeds.
+	pack.prepareSegmentMetricUpdates()
+	return segment, pack.metricMutation, nil
```
yhmo, line 373: internal/datacoord/segment_manifest_commit.go line:373
Low ---- applySegmentCatalogMutation runs the caller Operators through a full updateSegmentPack but returns only the target segment; any BinlogsIncrement staged in pack.increments, and any sibling segment an operator seeded into the pack, are silently discarded because the catalog write is a single AlterSegment of the target record. The batch path (publishSegmentManifestOperator inside UpdateSegmentsInfo) persists increments and every seeded segment, so the same operator type has different persistence semantics depending on the path. Current single-path callers (schema-bump materialization, stats) neither set increments nor touch other segments, so there is no present-day impact, but reusing AddL0DeltalogsOperator through the single path later would silently lose the deltalog KV writes. Consider documenting this constraint on SegmentCatalogMutation.Operators or asserting increments stay empty.

liliu-z (commented):
/approve

Issue #52624, opened 2026-08-18
[Enhancement]: Accelerate mix-compactor merge sort output pipeline


XuanYang-cn:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Accelerate the remaining output-reconstruction and compactor-pipeline costs in the default mix-compactor merge-sort path.

This is a follow-up to #51981 and PR #51998. That work converted `storage.MergeSort` to a proper k-way merge; the remaining hot path rebuilds every emitted row field by field and performs avoidable materialization for rows that will later be filtered.

#### Optimization breakdown

| Optimization | Previous cost | Proposed implementation | Correctness guard |
|---|---|---|---|
| Prepared typed output | Every emitted `(row, field)` repeats field-map lookup, builder dispatch, source-array type discovery, and growth | Prepare destination appenders once, bind typed source Arrow columns once per reader record, and reserve bounded near-term capacity | Differential tests for all scalar/vector/array/JSON/geometry/TEXT/struct families, including null/default behavior |
| Winning-run coalescing | Heap pop/push and output dispatch occur once per row even when one reader owns a contiguous global interval | Detect globally ordered runs and append qualifying ranges in bulk | Exact-once predicate evaluation, equal-key reader stability, sortedness diagnostics, and truncated-batch requeue tests |
| Direct Arrow forwarding | Eligible disjoint records are reconstructed field by field | Retain or slice writer-compatible `simpleArrowRecord` data and forward it through the normal writer | Fail closed for filtering, materialization, timestamp rewrite, schema mismatch, overlap, lifetime, batch threshold, and Segment rotation |
| Selection before materialization | Missing fields and function outputs are generated for rows later removed by deletes or TTL | Build reusable contiguous selections on raw records, drop empty records, retain keep-all records, and materialize only kept rows | Collection TTL, TTL-field, deletes, commit timestamps, varchar PK, BM25/function output, null/default, and accounting tests |
| Writer batch preservation | Faster record slicing can accidentally change logical batching or output Segment partitioning | Use the same logical byte accounting and pass forwarded/rebuilt pieces as one logical writer batch | V2/V3 row partition, stats, manifests, timestamp, TTL quantile, nullable count, PK bloom/range, and BM25 parity |

No public API, protobuf, storage-format, or user-facing configuration change is needed.

### Why is this needed?

[...]

PR #52625, opened 2026-08-18
enhance: cache merge sort output columns

XuanYang-cn:
Cache each reader's output columns when it advances to a new Arrow record, then reuse those columns while MergeSort emits rows. This removes repeated `Record.Column(fieldID)` lookups from the `(row, field)` output loop without changing sorting, filtering, default-value handling, batching, or writer behavior.

The cache is owned by `MergeSort`, where the current-record lifetime is known. `arrow_util.go` only provides a private stateless helper that appends already-resolved columns; the existing `RecordBuilder.Append` API and behavior remain unchanged.

Benchmark protocol: baseline `9a68eae9c2` (the affected storage code is identical to current `upstream/master`), candidate `1963a611a0`; Intel Core i7-8700, linux/amd64, Go 1.26.5, `GOMAXPROCS=4`, `-cpu=4`, 10 independent samples, 200ms/sample.

```bash
source scripts/setenv.sh
GOMAXPROCS=4 go test -tags dynamic,test -gcflags='all=-N -l' -run '^$' \
  -bench '^(BenchmarkMergeSort|BenchmarkMergeSortVarcharKey)$' \
  -benchmem -benchtime=200ms -count=10 -cpu=4 ./internal/storage
```

Median results:

- Wide MergeSort: 633.9ms to 454.3ms (-28.3%).
- Varchar-key MergeSort: 14.97ms to 13.73ms (-8.2%).
- Bytes per operation are effectively unchanged; the narrow case adds three one-time per-reader slice allocations (97 to 100 allocs/op).

Verification:

- Focused `MergeSort` and `RecordBuilder` tests passed.
- The new test proves output fields are resolved once per input record rather than once per emitted row.
- `make static-check` completed with zero issues.
- Resource-controlled Milvus build with `--skip-3rdparty` completed successfully.
- The package-wide storage test reached the environment-dependent Azure object-storage test, which cannot run without an Azure endpoint; no code-related failure was observed before that test.

See also: #52624


XuanYang-cn (commented):
## Review

Self-review notes on this change (all five optimizations verified against issue #52624, plus a focused read of `internal/storage/sort.go`, `internal/storage/arrow_util.go`, and `internal/datanode/compactor/{merge_sort,record_materializer,segment_writer}.go`).

### Summary

All five optimizations from #52624 are implemented and the correctness guards are unusually well covered by tests:

1. **Prepared typed output** — `prepareValueAppender`/`prepareRecord`/`appendPreparedRow` in `arrow_util.go`, wired through `ensurePrepared` in `MergeSort`. Covered by `TestMergeSortPreparedOutputMatchesGenericRecordBuilder` + `...ForStorageFieldFamilies` (scalar/vector/array/JSON/geometry/TEXT/struct, incl. null/default).
2. **Winning-run coalescing** — `emitRun` + `rowHeap.pushAfterRoot`. Covered by exact-once (`TestMergeSortPredicateCalledOncePerRow`), equal-key stability (`...PreservesEqualKeyReaderOrder`/`...AcrossRecords`), sortedness diagnostics (`TestMergeSortUnsortedInput*`), truncated-batch requeue (`...TruncatedForwardedPrefixRequeuesBeforeCompetitor`).
3. **Direct Arrow forwarding** — `canDirectForwardRecord`/`directForwardRecord`. The fail-closed eligibility is real code, not comments: `*simpleArrowRecord` type assertion (structurally excludes `timestampOverwriteRecord`/`materializedRecord`/`selectedRecord`/`nonForwardableRecord`), `Schema().Equal` + `Metadata().Equal`, per-field order/fingerprint/nullability/default-with-nulls checks. ~10 `TestMergeSortDirectForwards*` tests + `...RequiresExactWriterSchema` + `...RebuildsIncompatibleRecord`.
4. **Selection before materialization** — `newSelectedMaterializedRecordReader` + `selectFullRewriteRecord`, with `buildSelectedColumn` now slicing (`array.NewSlice`+`Concatenate`) instead of rebuilding. Covered by `TestSelectedMaterializedRecordReader*`, null preservation, and `TestMixCompactionMergeSortFiltersBeforeMaterializingMissingBM25Output`. The schema-evolved-TTL-field-absent-from-source case correctly falls back to materialize-then-filter via `filterInMergeSort`.
5. **Writer batch preservation** — `MultiSegmentWriter.WriteBatch` + `recordBatchWriter` interface. Covered by `TestMergeSortDirectForwardPreservesMultiSegmentRotation` (V2+V3 rotation parity).

No public API / protobuf / config change — scope respected.

### Non-blocking concerns

1. **Byte-size accounting is duplicated, not single-sourced** (`arrow_util.go`). `recordBuilderValueSize` (:801) re-derives the per-type widths that the 11 `appendPrepared*` functions (:543–681) already return, and its comment admits the two "must be identical". There is no unit test pinning `recordBuilderValueSize` to the builder's actual summed size — parity is only indirect, via the rotation test. Two related sharp edges:
   - the `default:` branch returns `0` (silent under-count → deferred rotation) instead of failing closed for an unknown type;
   - String/Binary use `offsets[end]-offsets[start]` (:824/:827), which over-counts if a foreign producer leaves non-zero offsets on null slots (Arrow permits undefined offsets; our own `AppendNull` keeps them equal).

   Suggestion: make the two paths share one size source, or add a test asserting `recordBuilderValueSize == builder summed size` across every supported type (including a null-entry offset case).

2. **`predicateDropped` is a global flag** (`sort.go`). Once any row is dropped anywhere, the rest of the merge falls back to per-row `emitRow`, disabling coalescing/forwarding for the *whole* merge. This is conservative and correct, but means a single schema-evolved-TTL reader with deletes can disable the fast paths for all readers. Worth a one-line comment if intentional.

3. **`prepareValueAppender` handles `a == nil` only for `*array.Float64Builder`** (:481). Every other builder errors on a nil source column. The compactor path is protected by the materializer filling missing fields, but `storage.MergeSort` is public — a caller passing a record missing a non-Float64 field gets an error where the old generic `RecordBuilder.Append` path used schema defaults.

4. **`x.ri < y.ri` tie-break** (sort.go ~:538) is sound only under the invariant "a reader holds at most one heap entry", which is documented in a comment but not asserted or tested.

5. **"Output bytes are identical" is a claim, not a test.** The only direct forward-vs-rebuild comparison (`TestMergeSortDirectForwardPreservesMultiSegmentRotation`) checks per-segment *row counts*, not byte content. Byte equality is true by construction (zero-copy slices of the same source buffers), so the residual risk is the size accounting in point 1 — but a byte-level forward-vs-rebuild parity test would close the loop.

### Minor nits

- `recordBuilderValueSize` and the `prepareValueAppender`/`appendPrepared*` type dispatch repeat the same `*array.X` switch four times; consider a shared table.
- `TestMergeSortMultipleSegmentsRewriteConditions` (a functional test) lives inside `mix_compactor_benchmark_test.go`; `require.True(t, base.closed == false)` in `record_materializer_test.go` would read better as `require.False(t, base.closed)`.

### Questions

1. Can size accounting be made single-sourced (or at least pinned by a test) instead of mirrored in `recordBuilderValueSize`?
2. Is the global `predicateDropped` fallback intentional/documented?
3. Is there a byte-level (not just rotation row-count) parity test planned for forward vs rebuild?
4. Is "missing non-Float64 field reaches `prepareValueAppender` with `col == nil`" impossible for all current callers of `storage.MergeSort`?

internal/storage/arrow_util.go
```diff
@@ -401,14 +401,24 @@ func (b *RecordBuilder) prepareAppendDefaults() error {
 }
 
 func (b *RecordBuilder) Append(rec Record, start, end int) error {
+	columns := make([]arrow.Array, len(b.fields))
```
liliu-z: `Append` now allocates a fresh `[]arrow.Array` per invocation, and it escapes to the heap because `len(b.fields)` is not a compile-time constant. `writeSegment` calls `Append` once per contiguous kept-row run (`mix_compactor.go:374`, `:391`) and `buildSelectedColumn` calls it once per selection range per field (`record_materializer.go:240`), so a segment with interleaved deletes degenerates to one-row ranges: the inner loop at `arrow_util.go:418` executes exactly once, saving zero `Record.Column` lookups while adding one heap allocation per row. This is a net regression for every `Append` caller, none of which is the `MergeSort` path the refactor targets.

XuanYang-cn: Fixed in 23e37630b7. Restored the generic Append implementation so it does not allocate a column slice per call; MergeSort retains its per-reader cached-column path. Added a pre-reserved single-row allocation regression test: it failed with 1 allocation/call before the fix and passes with 0 afterward. Focused RecordBuilder/MergeSort and materializer/compaction tests pass, as does the guarded incremental build. Static-check was blocked by permission denied while scanning the local Docker etcd data directory (deployments/docker/dev/volumes/etcd/member).

Issue #52623, opened 2026-08-18
[Bug]: Storage V3 manifest failures build invalid text and JSON stats paths


sijie-ni-0214:
## Environment

- **Milvus Version**: latest master
- **Deployment Mode**: standalone or cluster
- **MQ**: not applicable
- **SDK**: not applicable
- **OS**: not applicable

### Option B: Steps

1. Create or load a Storage V3 segment with a text index or JSON key stats and a manifest-backed `ManifestPath`.
2. Make the manifest read fail while converting the segment load info, for example during a transient object-store or manifest-read failure.
3. Observe `internal/util/segcore.resolveStatsWithBasePaths` while converting the load info for segcore.
4. Inspect the generated text/JSON stats base paths and the subsequent segcore load attempt.

### Trigger Conditions

- **Frequency**: whenever manifest resolution fails during segment load
- **First observed after**: the first affected segment load
- **Does NOT happen when**: the V3 manifest is read successfully

## Expected Behavior

A Storage V3 manifest-read failure must not construct Storage V2 legacy text/JSON stats prefixes. The load path should either omit the unresolved stats or propagate the manifest error so a retry can handle the transient failure.

## Actual Behavior

The V3 path falls through to the V2 base-path builders. This combines manifest-backed stats with legacy prefixes such as `BuildTextIndexPrefix` and `BuildJSONKeyStatsPrefix`, producing invalid object paths and misleading not-found load failures.

## Error Logs

```text
failed to resolve stats from manifest for segcore load info
```

## Analysis Hints

- Suspect code: `internal/util/segcore/segment.go`, `resolveStatsWithBasePaths`
- Related PR: #52571
- V3 stats files are stored under the manifest base path and `_stats` namespace; V2 prefixes are incompatible.

PR #52626, opened 2026-08-18
fix: avoid legacy stats paths after V3 manifest failures

sijie-ni-0214:
issue: #52623

## What this PR does

Prevent Storage V3 segment load-info conversion from falling through to Storage V2 text-index and JSON-stats path construction when manifest resolution fails.

A V3 manifest failure now propagates an error through `ConvertToSegcoreSegmentLoadInfo`, `CreateCSegment`, and `Reopen`. This prevents segcore from interpreting a failed stats resolution as stats deletion or rebuilding indexes from raw data. Reopen fails before mutating the existing segment state, allowing the caller to retry the transient manifest failure.

## Test Plan

- `go test -count=1 -tags dynamic,test -gcflags=\"all=-N -l\" ./internal/util/segcore` with the required macOS rpaths
- Targeted `golangci-lint` for `internal/util/segcore` (`0 issues`)
- `make static-check` passed before the final error-propagation follow-up
- `gofumpt` and `git diff --check`

### DIFF
--- a/internal/util/segcore/segment.go
+++ b/internal/util/segcore/segment.go
@@ -71,7 +71,10 @@ func CreateCSegment(req *CreateCSegmentRequest) (CSegment, error) {
 	var ptr C.CSegmentInterface
 	var status C.CStatus
 	if req.LoadInfo != nil {
-		segLoadInfo := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+		segLoadInfo, err := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+		if err != nil {
+			return nil, merr.Wrap(err, "failed to convert segment load info")
+		}
 		loadInfoBlob, err := proto.Marshal(segLoadInfo)
 		if err != nil {
 			return nil, err
@@ -356,7 +359,10 @@ func (s *cSegmentImpl) Reopen(ctx context.Context, req *ReopenRequest) error {
 	defer runtime.KeepAlive(traceCtx)
 	defer runtime.KeepAlive(req)
 
-	segLoadInfo := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+	segLoadInfo, err := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+	if err != nil {
+		return merr.Wrap(err, "failed to convert reopen load info")
+	}
 	loadInfoBlob, err := proto.Marshal(segLoadInfo)
 	if err != nil {
 		return err
@@ -409,15 +415,18 @@ func (s *cSegmentImpl) SetCommitTimestamp(ts uint64) error {
 // ConvertToSegcoreSegmentLoadInfo converts querypb.SegmentLoadInfo to segcorepb.SegmentLoadInfo.
 // This function is needed because segcorepb.SegmentLoadInfo is a simplified version that doesn't
 // depend on data_coord.proto and excludes fields like start_position, delta_position, and level.
-func ConvertToSegcoreSegmentLoadInfo(src *querypb.SegmentLoadInfo) *segcorepb.SegmentLoadInfo {
+func ConvertToSegcoreSegmentLoadInfo(src *querypb.SegmentLoadInfo) (*segcorepb.SegmentLoadInfo, error) {
 	if src == nil {
-		return nil
+		return nil, nil
 	}
 
 	// Resolve text/json stats with basePaths.
 	// V2: stats come from src proto fields, basePaths computed from metadata + rootPath.
 	// V3: stats resolved from manifest (src proto fields are empty), basePaths from manifest paths.
-	textStats, jsonStats, textBasePaths, jsonBasePaths := resolveStatsWithBasePaths(src)
+	textStats, jsonStats, textBasePaths, jsonBasePaths, err := resolveStatsWithBasePaths(src)
+	if err != nil {
+		return nil, err
+	}
 
 	return &segcorepb.SegmentLoadInfo{
 		SegmentID:            src.GetSegmentID(),
@@ -444,17 +453,20 @@ func ConvertToSegcoreSegmentLoadInfo(src *querypb.SegmentLoadInfo) *segcorepb.Se
 		UseTakeForOutput:     src.GetUseTakeForOutput(),
 		EstimatedBytesPerRow: src.GetEstimatedBytesPerRow(),
 		CommitTimestamp:      src.GetCommitTimestamp(),
-	}
+	}, nil
 }
 
 // resolveStatsWithBasePaths resolves text/json stats and computes basePaths.
 // V2: stats from src proto fields, basePaths computed from rootPath + metadata.
 // V3: stats resolved from manifest via StatsResolver, basePaths extracted from manifest paths.
+// A V3 manifest error does not fall back to V2 path construction because the
+// legacy prefixes are incompatible with manifest-backed stat files.
 func resolveStatsWithBasePaths(src *querypb.SegmentLoadInfo) (
 	map[int64]*datapb.TextIndexStats,
 	map[int64]*datapb.JsonKeyStats,
 	map[int64]string, // textBasePaths
 	map[int64]string, // jsonBasePaths
+	error,
 ) {
 	textStats := src.GetTextStatsLogs()
 	jsonStats := src.GetJsonKeyStatsLogs()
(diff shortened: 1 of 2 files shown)


czs007:

This PR's new empty-stats return on V3 manifest resolution failure introduces two confirmed issues around segment reopen: a silent unload of already-loaded stats and a broken self-repair loop.

## High

- `internal/util/segcore/segment.go:472` — The newly added `return nil, nil, nil, nil` on manifest parse failure feeds an empty load info into segcore's reopen diff, which interprets missing fields as removals: already-loaded JSON key stats are dropped (`SegmentLoadInfo.cpp:960-967` → `ChunkedSegmentSealedImpl.cpp:7450-7467`) and every `enable_match` field is queued for a full text-index rebuild from raw data (`SegmentLoadInfo.cpp:863-877` → `CreateTextIndexWithSchema`). A transient object-storage error (S3 5xx/timeout/throttle) in the `GetManifestStats` FFI during a routine querycoord-triggered Reopen of a healthy V3 segment → segcore erases the loaded JSON stats, rebuilds the text index from raw data (minutes of CPU on large segments), and Reopen returns success with only a warn log, so nothing retries and JSON path filters permanently fall back to the slow raw-data path. Before this PR the same error fell through to the V2 legacy path, whose incompatible `base_path` made the reopen fail loudly and get retried — the silent, destructive outcome is new. Suggestion: add an `error` return to `resolveStatsWithBasePaths` and propagate it through `ConvertToSegcoreSegmentLoadInfo` so `CreateCSegment`/`Reopen` fail explicitly and are retried, rather than synthesizing empty stats that segcore reads as deletion; at minimum the reopen path must fail instead of dropping. Consider adding a counter for observability, and update the new `TestConvertToSegcoreSegmentLoadInfo_V3ManifestErrorDropsStats` test, which currently locks in the drop semantics as expected behavior. *(raised by bigsheeper, tinswzy)*

## Medium

- `internal/querynodev2/segments/segment.go:1247` — After the silent drop above, `syncFieldJSONStatsFromLoadInfo` makes the opposite decision on the same manifest error: it falls back to `loadInfo.GetJsonKeyStatsLogs()` (non-empty for V3 segments via the querycoord placeholders), so the QueryNode still registers and reports the JSON stats as loaded. Convert-side FFI failure leaves segcore with no JSON stats while the sync-side second FFI call succeeds or takes the proto fallback → the distribution report marks the field as covered, `IndexChecker.checkSegmentStats` never schedules a repair reopen, and the segment serves JSON path filters via the slow raw-data path — correct results, but the degradation is invisible and never self-heals until the segment is fully reloaded for an unrelated reason. The fallback line itself is pre-existing code, but this "segcore has nothing, report says loaded" state only becomes reachable with this PR's empty return. Suggestion: make the two sites agree — if the drop semantics are kept, report nothing on manifest error so the checker re-triggers; if the error is propagated instead (preferred), this fallback becomes unreachable for that case. *(raised by bigsheeper, tinswzy)*

sijie-ni-0214:
Thanks for the review. Addressed both findings in commit 78ff1ba316.

- resolveStatsWithBasePaths now returns the manifest resolution error instead of synthesizing empty stats or falling through to V2 paths.
- ConvertToSegcoreSegmentLoadInfo propagates the error to CreateCSegment and cSegmentImpl.Reopen.
- Reopen now fails before invoking segcore, so existing text/JSON stats are not interpreted as removals and local load-info state is not advanced. The caller can retry the transient manifest failure.
- Updated the regression test to assert the conversion error.

Validation: the full internal/util/segcore package test passed, and targeted golangci-lint reported 0 issues.

czs007:

**Verified:**
- `internal/util/segcore/segment.go:482` — the V3 manifest-failure branch now returns `merr.Wrap(result.Err(), "failed to resolve V3 stats from manifest")` instead of empty maps, so a failed resolution can no longer be interpreted by segcore's load-diff as stat deletions.
- `internal/util/segcore/segment.go:418` — `ConvertToSegcoreSegmentLoadInfo` now returns `(*segcorepb.SegmentLoadInfo, error)`; both production call sites (`CreateCSegment` at `segment.go:74–77`, `cSegmentImpl.Reopen` at `segment.go:362–365`) propagate the error before any cgo call, and a repo-wide grep confirms no other callers of the old signature remain.
- `internal/querynodev2/segments/segment.go:1299–1306` — `LocalSegment.Reopen` returns before `s.loadInfo.Store(newLoadInfo)` and `syncFieldJSONStatsFromLoadInfo`, so previously loaded stats and load info stay intact on failure; the error then propagates through `segment_loader.go:2525–2529` and `handlers.go:154–158`, so the querycoord reopen task fails and is retried rather than marked done.
- `internal/storagev2/packed/stats_resolver.go:236–248` — the new error branch is reachable only when `ManifestPath != ""`; V3 segments with an empty manifest path still take the legacy path, and a manifest with zero stat entries returns an empty map with `nil` error (`ffi_stats.go:142–144`), so no new false load failures are introduced.
- `internal/util/segcore/segment_test.go:474–489` — `TestConvertToSegcoreSegmentLoadInfo_V3ManifestErrorReturnsError` asserts `(nil, err)` for a V3 load info with an unparsable manifest path, exercising exactly the previously-broken branch deterministically (fails on JSON decoding before any FFI call).

**Non-blocking notes (no change required for this PR):**
- The commit message still says "Drop unresolved stats instead of forwarding entries without a valid manifest base path", which describes an earlier iteration; the final code propagates an error instead. Worth re-wording when squashing.

- `internal/querynodev2/segments/segment_loader.go:981–988` — logs and continues with empty text/JSON stats on the same manifest resolution error; this is existing behaviour on master and outside this PR's scope (`ConvertToSegcoreSegmentLoadInfo`), flagged only so it is not assumed covered. No tracking issue yet.

czs007:

Adversarial review of the V3-manifest fail-fast change surfaced one low-severity test-coverage gap introduced by this PR, plus one pre-existing hazard recorded for follow-up only.

## Low

- **`internal/util/segcore/segment.go:74` — the two behavior-changing call sites lack call-site-level tests.** The new test `TestConvertToSegcoreSegmentLoadInfo_V3ManifestErrorReturnsError` calls `ConvertToSegcoreSegmentLoadInfo` directly, so it pins the converter itself but not the two call sites this PR changes (`CreateCSegment` at `segment.go:74` and `cSegmentImpl.Reopen` at `segment.go:362`). The only in-package `CreateCSegment` test (`segment_test.go:49`) passes no `LoadInfo` and never reaches the new error-check branch, so if a call site regressed to `segLoadInfo, _ := ...` the in-package tests would stay green. Also note the test's `"not-a-valid-manifest-path"` input fails at `json.Unmarshal` inside `UnmarshalManifestPath` (`packed_reader_ffi.go:426-428` → `ffi_common.go:361`) before any storage access, so only the parse-failure path is covered, not the read-failure path. This is a test-hardening suggestion, not a defect. *Suggestion:* add tests that drive `CreateCSegment` and `Reopen` with a `LoadInfo` carrying an invalid V3 manifest and assert the error propagates, and consider a case that fails during manifest read rather than parse. *(raised by czs007, tinswzy)*

- **`internal/querynodev2/segments/segment.go:1240` — `syncFieldJSONStatsFromLoadInfo` still silently degrades, and the Reopen path parses the same manifest twice.** When `statsResult.Err() != nil`, the function silently falls back to `loadInfo.GetJsonKeyStatsLogs()` (`segment.go:1242-1247`), which is inconsistent with the "no fallback on V3 manifest failure" policy this PR establishes in `resolveStatsWithBasePaths`; additionally, `Reopen` parses the same manifest independently twice (once via `ConvertToSegcoreSegmentLoadInfo` inside `cSegmentImpl.Reopen`, again via `LocalSegment.Reopen` → `syncFieldJSONStatsFromLoadInfo`, each through a fresh `NewStatsResolver` FFI call). Note the impact is bounded: querycoord preserves JSON stats placeholders for manifest segments (`internal/querycoordv2/utils/types.go:98-110`), and `syncFieldJSONStatsFromLoadInfo` consumes only their metadata, so the fallback typically reports datacoord metadata rather than clearing stats or triggering redundant reopens — the residual problem is the policy inconsistency of the silent branch plus redundant FFI parse overhead. This is unrelated to this PR because the diff only touches `internal/util/segcore/segment.go` and its tests, leaving this fallback branch untouched. *Suggestion:* unify the error policy and deduplicate the manifest parse in a separate change (no tracking issue yet). *(raised by czs007, tinswzy)*

czs007 (approved):
/lgtm

Issue #52355, opened 2026-08-10
[Enhancement]: Avoid eager nullable chunk validity expansion


sparknack:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Keep nullable chunk validity in its packed bitmap representation after loading. Point validity checks, filtering, and logical-to-physical rank calculation should read the packed bitmap directly. Variable-width views should build only request-local validity, while fixed-width chunks should materialize the compatibility validity buffer only on the first `Span()` call.

### Why is this needed?

Nullable chunks currently expand every validity bit into a `FixedVector<bool>` during construction. This adds row-count-proportional CPU and resident memory to every loaded nullable chunk, including mmap-backed chunks and code paths that only need individual validity bits.

### Anything else?

Implementation: #52354

PR #52627, opened 2026-08-18
enhance: [2.6] Keep nullable chunk validity packed

sparknack:
pr: #52354
issue: #52355

- Keep nullable chunk validity packed in the backing mmap or buffer for point checks, filtering, rank, and fixed-width Span access.
- Add a non-owning `ValidityView` for packed and expanded validity.
- Remove the fixed-width chunk-level expanded validity cache; materialize only at legacy scalar-index boundaries.
- Adapt the 2.6 expression, SkipIndex, JSON stats, mmap column, and segcore paths.
- Document the offset contract for expanded and packed `ValidityView` accessors.

## Test
- `clang-format --style=file --dry-run --Werror internal/core/src/common/ValidityView.h`
- `git diff --check`
- `make verifiers` (blocked locally during third-party setup: installed Conan is 2.x, while branch 2.6 requires Conan 1.x)

### DIFF
--- a/internal/core/src/common/ValidityView.h
+++ b/internal/core/src/common/ValidityView.h
@@ -1,0 +1,104 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace milvus {
+
+// Non-owning view over either one-byte-per-row bool data or an LSB-first
+// packed bitmap. The underlying buffer must outlive the view.
+class ValidityView {
+ public:
+    ValidityView() = default;
+    ValidityView(std::nullptr_t) {
+    }
+
+    static ValidityView
+    FromExpanded(const bool* data) {
+        return ValidityView(data, Encoding::Expanded, 0);
+    }
+
+    static ValidityView
+    FromPacked(const uint8_t* data) {
+        return ValidityView(data, Encoding::Packed, 0);
+    }
+
+    explicit operator bool() const {
+        return data_ != nullptr;
+    }
+
+    bool
+    operator[](int64_t index) const {
+        const auto offset = offset_ + index;
+        if (encoding_ == Encoding::Expanded) {
+            return static_cast<const bool*>(data_)[offset];
+        }
+        return (static_cast<const uint8_t*>(data_)[offset >> 3] >>
+                (offset & 0x07)) &
+               1;
+    }
+
+    bool
+    is_packed() const {
+        return encoding_ == Encoding::Packed;
+    }
+
+    ValidityView
+    Subview(int64_t offset) const {
+        auto result = *this;
+        result.offset_ += offset;
+        return result;
+    }
+
+    const bool*
+    expanded_data() const {
+        if (encoding_ != Encoding::Expanded) {
+            return nullptr;
+        }
+        return static_cast<const bool*>(data_) + offset_;
+    }
+
+    const uint8_t*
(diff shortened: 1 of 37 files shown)


internal/core/src/exec/expression/Expr.h
```diff
@@ -62,6 +62,66 @@ PinIndex(milvus::OpContext* op_ctx,
...
+
+    if (const auto* expanded = validity.expanded_data(); expanded != nullptr) {
+        ApplyValidMask(expanded, res, valid_res, size);
+        return;
+    }
+
+    const auto* packed = validity.packed_data();
+    AssertInfo(packed != nullptr, "Packed validity data is missing");
+
+    const auto read_word = [packed](int64_t bit_offset, int bit_count) {
```
buqian-zilliz, line 97: **[P3] Duplicated packed-bit validity extraction should be shared**

`ApplyValidMask(ValidityView, ...)` in `internal/core/src/exec/expression/Expr.h` and `Chunk::ApplyValidityMask` in `internal/core/src/common/Chunk.h` contain an identical, non-trivial `read_word` implementation for unpacking an LSB-first bitmap into a 64-bit word, including the 9-byte unaligned tail handling. Any fix to the bit order, masking, or bounds logic must now be applied in two places. Consider extracting this into a single helper (for example on `ValidityView`) and having both callers share it.

internal/core/src/query/ScalarIndex.h
```diff
@@ -25,7 +25,17 @@ template <typename T>
 inline index::ScalarIndexPtr<T>
 generate_scalar_index(Span<T> data) {
     auto indexing = std::make_unique<index::ScalarIndexSort<T>>();
-    indexing->Build(data.row_count(), data.data(), data.valid_data());
+    const auto validity = data.validity();
+    FixedVector<bool> materialized_validity;
+    const bool* valid_data = validity.expanded_data();
+    if (validity && valid_data == nullptr) {
+        materialized_validity.reserve(data.row_count());
```
buqian-zilliz, line 32: **[P3] generate_scalar_index transiently re-expands the packed validity bitmap**

For a packed `Span<T>` (the common case for chunked sealed nullable columns after this change), `generate_scalar_index` builds a `FixedVector<bool>` of `row_count()` entries solely to satisfy `ScalarIndexSort::Build(const bool*)`. This reintroduces an O(rows) allocation during every scalar-index build that the PR otherwise removes from the resident chunk data, and can spike peak memory for large segments. Consider teaching `ScalarIndexSort::Build` to accept `ValidityView` or iterate packed bits directly so index construction stays packed.

internal/core/src/common/ValidityView.h
```diff
@@ -0,0 +1,104 @@
...
+    const bool*
+    expanded_data() const {
+        if (encoding_ != Encoding::Expanded) {
+            return nullptr;
+        }
+        return static_cast<const bool*>(data_) + offset_;
+    }
+
+    const uint8_t*
+    packed_data() const {
```
buqian-zilliz, line 82: **[P3] ValidityView packed and expanded accessor offset semantics are asymmetric**

`expanded_data()` returns `data_ + offset_`, but `packed_data()` returns the base pointer and requires the caller to separately read `bit_offset()` (as both `ApplyValidMask` implementations do). A caller that treats `packed_data()` like `expanded_data()` and indexes it directly would read the wrong rows whenever the view was produced with `Subview`. This asymmetry is easy to misuse; consider documenting the contract next to these accessors or making them consistent.

internal/core/src/exec/expression/Expr.h
sparknack: Verified that the two packed-word readers duplicate the same logic. This backport intentionally matches the implementation already merged in #52354, where the same concern remains open: https://github.com/milvus-io/milvus/pull/52354#discussion_r3764128738. To avoid a 2.6-only helper/API divergence, the shared-helper refactor should land on master first and then be backported separately.

internal/core/src/query/ScalarIndex.h
sparknack: The transient materialization is intentional and called out in the PR body: this change removes the resident chunk-level expanded validity cache while preserving the legacy `ScalarIndexSort::Build(const bool*)` boundary. A repository-wide search on 2.6 finds no `generate_scalar_index` callers outside this header, so this allocation is not on every scalar-index build path. Extending `ScalarIndexSort` to consume `ValidityView` directly would be a separate API/performance change and is kept out of this backport.

internal/core/src/common/ValidityView.h
sparknack: Fixed in 86b1746975. The accessor comments now state that `expanded_data()` applies the Subview offset, while `packed_data()` returns the bitmap base and `bit_offset()` is relative to that base.

zhengbuqian (approved):

PR #52629, opened 2026-08-18
test: add Go, PyMilvus, and REST e2e for FMINDEX and membership_match

yanliang567:
## Summary

Add end-to-end coverage for FMINDEX and the unified `membership_match` expression across the Go SDK, PyMilvus, and RESTful v2.

Related issues: #51375, #52778, #53036

### FMINDEX

- Build parameter defaults, boundaries, combined parameters, and invalid types or ranges.
- VARCHAR-only validation, including JSON, scalar, and struct-array sub-fields.
- Prefix, infix, suffix, general LIKE recheck, regex fallback, escaped wildcards, NULL, and UTF-8 behavior.
- Long-text low-hit fixtures and sealed segments above `minSegmentNumRowsToEnableIndex`, preventing fake-finished index coverage.
- Exact indexed-field versus unindexed twin-field PK comparison.
- Growing plus sealed visibility, mutation, upsert, named-index drop, and rebuild lifecycle checks.
- Query and Search coverage with exact projected fields, distance ordering, index identity, type, and row-progress assertions.
- VARCHAR AUTOINDEX resolution is verified against the configured scalar-index policy without hard-coding a specific index type.

### membership_match

- Bloom and Roaring wire compatibility through a shared static manifest while each SDK keeps independent collection setup.
- Bloom zero-false-negative checks plus a statistically bounded false-positive check tied to the configured FPR; Roaring uses exact PK ground truth.
- Query and Search must return the same complete Bloom-filtered PK set when topK covers the fixture.
- Integer widths, negative values, INT64 boundaries, VARCHAR, JSON strict typing, nested paths, dynamic fields, NULL semantics, and unsupported top-level BOOL/FLOAT fields.
- Positive, negated, empty, malformed, domain-mismatch, delete, growing plus sealed, and struct-array rejection cases.
- Explicit `STL_SORT`, `INVERTED`, `BITMAP`, and `TRIE` scalar-index paths.
- Internal-only `HYBRID` paths are exercised through the supported `AUTOINDEX` request and verified against DataCoord's resolved build parameters for primitive and JSON-path indexes.
- Mutation tests use isolated collections; shared read-only fixtures register cleanup before setup begins.

### RESTful v2

- FMINDEX anchored LIKE Query and Search compared with exact PK ground truth and an unindexed twin field.
- Index identity, type, persisted parameters, row progress, completion, and collection load state are verified before assertions.
- Search validates exact projected values by primary key.
- JSON string and missing membership template parameters are verified as server-side validation failures.
- Python `bytes` are verified to fail local JSON serialization before HTTP. A positive REST binary-template case remains blocked by #53036.

## Validation

- `go test -tags dynamic,test -gcflags='all=-N -l' -c ./testcases`
- `go vet -tags dynamic,test ./testcases`
- `go test ./membership/sbbf -count=1`
- Ruff check and format check for all changed Python and REST files
- Python byte-compilation for all changed Python and REST files
- Pytest collection: 68 PyMilvus/FMINDEX tests and 7 REST tests
- Cross-SDK membership wire-manifest verification
- `git diff --check`

The Proxy parameter-resolution unit test and Bloom integration package cannot execute on the host because it lacks the native `milvus_core`, `rdkafka`, and `rocksdb` pkg-config packages. A targeted PyMilvus FMINDEX run reached index creation but timed out after 120 seconds before assertions; its temporary collection was cleaned up. Full server-backed E2E was not completed.


wangting0128 (approved):
/lgtm

yanliang567:
Addressed the review findings in a6731bc446: corrected the FMINDEX LIKE/regex separators in Go and Python; made Bloom FPR cases assert every true-member PK before bounding false positives; changed Bloom Search checks to exact Query/Search PK-set equality; and registered guarded shared-fixture cleanup before setup. Go compile/vet, Ruff check/format, 67-case Python collection, and offline boundary/pattern checks pass. Server-backed E2E could not be rerun locally because no Milvus endpoint is available; PR remains Draft. PTAL.

wangting0128 (approved):
/lgtm

PR #52631, opened 2026-08-18
enhance: bump knowhere to v3.0.8

foxspy:
Move the knowhere pin from `6a3b7df1` to the `v3.0.8` tag (main at `d85f7080`, zilliztech/knowhere#1780), which bumps the cardinal references to v3.0.5 (v2) and v2.5.110 (v1).

The stale comment above the pin is updated to describe the new one.

Heads-up for reviewers: this range also brings in zilliztech/knowhere#1673 "support nullable external id mapping", which the previous pin deliberately excluded.

### DIFF
--- a/internal/core/thirdparty/knowhere/CMakeLists.txt
+++ b/internal/core/thirdparty/knowhere/CMakeLists.txt
@@ -14,9 +14,9 @@
 # Update KNOWHERE_VERSION for the first occurrence
 milvus_add_pkg_config("knowhere")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-# Pins the short zilliztech/knowhere#1761 commit without nullable ID mapping.
-# It retains the COSINE/IP CalcDistByIDs fix for fully cached DiskANN vectors.
-set( KNOWHERE_VERSION 6a3b7df1 )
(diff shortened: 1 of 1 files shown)


foxspy:
Closing: superseded, and merging it now would be a regression.

master and 3.0 both already pin knowhere `d7cfd888`, not the `6a3b7df1` this PR was written against. The pin moved twice since:

- 08-19 `77ebd82e0e` (#50524, move nullable vector mapping into index layer) → `d85f7080`
- 08-22 `57ea5fff5e` (#52754, stable IndexNode ABI) → `d7cfd888`

So this PR's intent — get onto a knowhere revision carrying cardinal v3.0.5 / v2.5.110 — is already satisfied, and `d85f7080` is now **3 commits behind** the current pin. Merging would drop:

- `d7cfd888` fix: keep IndexNode ABI stable across Cardinal builds (zilliztech/knowhere#1786)
- `8c864b5f` fix: update SVS SQ8 storage enum (zilliztech/knowhere#1784)
- `03b845e3` enhance: skip fully filtered windows in sindi sparse search (zilliztech/knowhere#1782)

The concern in the original description — that this bump re-introduces zilliztech/knowhere#1673 — is also moot: `7cc9d4be` is already an ancestor of the current pin (0 behind / 15 ahead).

If a tag-form pin is still wanted for the 3.0 branch, it should be a fresh tag cut at `d7cfd888` or later, not `v3.0.8` (= `d85f7080`).

Issue #52632, opened 2026-08-18
[Bug]: Import ID pre-allocation may overflow for large restore jobs


loupipalien:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master
- Component: DataCoord Import/Restore
```

### Current Behavior

During import/restore, DataCoord calculates the pre-allocated ID count as:
```
totalRows + 1) * binlogNum * expansionFactor
```
The result is later converted to `uint32` for `AllocIDRequest.Count`.

For large restore jobs, this value can exceed `math.MaxUint32`. The conversion may wrap/truncate the requested count, causing DataCoord to pre-allocate fewer IDs than required.

### Expected Behavior

The calculation should avoid integer overflow and should not wrap to a smaller ID count. If the requested count exceeds the `uint32` limit accepted by the allocator request, it should be capped safely.

### Steps To Reproduce

```markdown

```

### Milvus Log

2026/08/16 21:09:30.027 +08:00] [INFO] [datacoord/import_util.go:338] ["pre-allocate ids and ts for import task"] [taskID=468401330144437736] [jobID=468401327315881744] [collectionID=468401321412078082] [type=ImportTask] [state=JobStateInit] [nodeID=-1] [totalRows=78201209] [fieldsNum=9] [idBegin=468401337295581806] [idEnd=468401337307780314] [ts=468420817269821043]

restore: run collection task restore: wait restore collections restore: restore collection restore_collection: restore data: restore_collection: restore data v2: restore_collection: wait for partition restore: restore_collection: restore partition v2: restore_collection: restore not L0 groups: restore_collection: wait for not L0 segment restore: restore_collection: bulk insert via restful: restore_collection: check bulk insert via restful: restore_collection: bulk insert failed: error: ID is exhausted, start=468401337307545796, end=468401337307780314, count=412509: service internal error, file: id:468401327315884371 paths:"restore-temp-831eac5947334b62994bcaf33b092975-default-b_read_search_embed_128_large/binlogs/insert_log/467783801368634506/467783801368634507/468401266289579034/468401266289579034/" paths:"restore-temp-831eac5947334b62994bcaf33b092975-default-b_read_search_embed_128_large/binlogs/delta_log/467783801368634506/467783801368634507/468401266289579034/468401266289579034/"

xiaofan-luan:
[...]

PR #52633, opened 2026-08-18
fix: prevent import id pre-allocation overflow

loupipalien:
issue: #52632

## What was changed

- Calculate import ID pre-allocation count using big integers.
- Cap the allocator request count at `math.MaxUint32` instead of allowing uint32 wraparound.
- Add unit tests for normal, overflow, and invalid parameter cases.

## Why

Large import/restore jobs can request more IDs than `AllocIDRequest.Count` can represent. The previous conversion to `uint32` could wrap to a smaller value, causing insufficient ID pre-allocation.

## Test

```text
go test -tags dynamic,test -gcflags="all=-N -l" -count=1
./internal/datacoord
-run '^TestCalculatePreAllocIDNum$'
```

### DIFF
--- a/internal/datacoord/import_util.go
+++ b/internal/datacoord/import_util.go
@@ -340,7 +340,7 @@ func AssembleImportRequest(task ImportTask, job ImportJob, meta *meta, alloc all
 	fieldsNum := len(job.GetSchema().GetFields()) + 2 // userFields + tsField + rowIDField
 	binlogNum := fieldsNum + 2                        // binlogs + statslog + BM25Statslog
 	expansionFactor := paramtable.Get().DataCoordCfg.ImportPreAllocIDExpansionFactor.GetAsInt64()
-	preAllocIDNum, err := calculatePreAllocIDNum(totalRows, int64(binlogNum), expansionFactor)
+	preAllocIDNum, preAllocIDNumCapped, err := calculatePreAllocIDNum(totalRows, int64(binlogNum), expansionFactor)
 	if err != nil {
 		return nil, err
 	}
@@ -357,6 +357,7 @@ func AssembleImportRequest(task ImportTask, job ImportJob, meta *meta, alloc all
 		mlog.Int64("totalRows", totalRows),
 		mlog.Int("fieldsNum", fieldsNum),
 		mlog.Uint32("preAllocIDNum", preAllocIDNum),
+		mlog.Bool("preAllocIDNumCapped", preAllocIDNumCapped),
 		mlog.Int64("idBegin", idBegin),
 		mlog.Int64("idEnd", idEnd),
 		mlog.Uint64("ts", ts))...,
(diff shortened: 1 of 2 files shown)


loupipalien:
@cydrain @sunby would you mind reviewing this pr? I'll really appreciate it.

internal/datacoord/import_util.go
```diff
@@ -388,6 +394,27 @@ func AssembleImportRequest(task ImportTask, job ImportJob, meta *meta, alloc all
...
+		)
+	}
+
+	preAllocIDNum := big.NewInt(totalRows)
+	preAllocIDNum.Add(preAllocIDNum, big.NewInt(1))
+	preAllocIDNum.Mul(preAllocIDNum, big.NewInt(binlogNum))
+	preAllocIDNum.Mul(preAllocIDNum, big.NewInt(expansionFactor))
+	requested := preAllocIDNum.String()
+
+	if preAllocIDNum.Cmp(new(big.Int).SetUint64(uint64(math.MaxUint32))) > 0 {
```
yhmo: internal/datacoord/import_util.go line:411
Medium ---- Capping at math.MaxUint32 fixes the uint32 wraparound for the reported case (78M rows, expected need ~860M IDs), but a task whose pre-expansion need (totalRows+1)*binlogNum itself exceeds MaxUint32 (roughly >390M rows with binlogNum=11) will still run out of IDs mid-import with the same "ID is exhausted" symptom as issue #52632, and the cap is only distinguishable in an Info-level log field (preAllocIDNumRequested). The sibling compaction path (compaction_trigger_v2.go createCompactionIDBlock, from #50210) detects the equivalent condition and fails fast with a descriptive error. Consider emitting a warning (or a pre-flight rejection) when the pre-expansion need exceeds MaxUint32, so operators are not left with the same confusing failure mode.

internal/datacoord/import_util.go
```diff
@@ -388,6 +394,27 @@ func AssembleImportRequest(task ImportTask, job ImportJob, meta *meta, alloc all
 	return req, nil
 }
 
+func calculatePreAllocIDNum(totalRows, binlogNum, expansionFactor int64) (uint32, string, error) {
+	if totalRows < 0 || binlogNum <= 0 || expansionFactor <= 0 {
```
yhmo: internal/datacoord/import_util.go line:398
Low ---- The binlogNum <= 0 disjunct is unreachable: binlogNum = fieldsNum + 2 and fieldsNum = len(schema fields) + 2, so binlogNum >= 4 for any schema. Only totalRows < 0 and expansionFactor <= 0 can actually trigger this error. Consider dropping the dead disjunct (or commenting why it is defensive) so the guard and the "invalid parameters" message do not claim a case that cannot occur.

internal/datacoord/import_util.go
```diff
@@ -388,6 +394,27 @@ func AssembleImportRequest(task ImportTask, job ImportJob, meta *meta, alloc all
...
+		return 0, "", merr.WrapErrServiceInternalMsg(
```
yhmo: internal/datacoord/import_util.go line:399
Low ---- The sibling code for the same config class (compaction_trigger_v2.go createCompactionIDBlock) returns merr.WrapErrParameterInvalidMsg when the pre-allocate expansion factor is non-positive, while this new guard returns merr.WrapErrServiceInternalMsg with an "invalid ... parameters" message. Per the error-handling guide's blame test and the sibling precedent, ErrParameterInvalid (InputError) is the class for an operator-supplied expansion factor; the System classification changes the client-visible code (5 vs 1100) and the metrics/access-log cause (system vs user). Aligning with the sibling would keep the two pre-allocation paths consistent.

internal/datacoord/import_util.go
loupipalien: Thanks for your time and valuable comment. :)

yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/datacoord/import_util.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/datacoord/import_util.go
yhmo: The issue is confirmed fixed. Resolving this conversation.

Issue #52575, opened 2026-08-17
[Bug]: External snapshot restore deadlocks on 3.0 branch builds: copy segment task can never pass the 3.0.1 worker-version gate


huanghaoyuanhhy:
## Environment

- **Milvus Version**: 3.0 branch build `3.0-20260814-b298f9d2e` (commit b298f9d2eb, includes #52476); reproducible on any build from the `3.0` branch or `master`
- **Deployment Mode**: standalone
- **MQ**: woodpecker
- **SDK**: milvus-backup v0.5.x (`RestoreExternalSnapshot` / `GetRestoreSnapshotState` gRPC API)

## Reproduction

1. Deploy: standalone Milvus built from the `3.0` branch (any commit after #52476)
2. Setup: create a collection, insert data, create a snapshot-format backup of it
3. Trigger: call `RestoreExternalSnapshot` with the snapshot bundle's metadata URI (this is what milvus-backup does for every snapshot-format restore — the local `RestoreSnapshot` API cannot be used here because it only accepts snapshots registered in the target cluster's own snapshot registry)
4. Observed via: `GetRestoreSnapshotState` returns `RestoreSnapshotExecuting` with `progress=0` forever; the restore never completes and never fails

### Trigger Conditions

- **Frequency**: always, on every build whose `pkg/common/version.go` is below `3.0.1` — that is currently ALL `3.0` branch builds (`3.0.0`) and all `master` builds (`3.0.0-beta`)
- **Does NOT happen when**: restoring via the local `RestoreSnapshot` API (by snapshot name from the cluster's own registry) — `MinimumWorkerVersion()` returns an empty constraint for non-external jobs. Note this path is unreachable for backup/restore in practice: milvus-backup always calls `RestoreExternalSnapshot` for snapshot-format backups, even when the bundle was already staged into the target cluster's own bucket.

## Expected Behavior

The copy segment task is dispatched to the DataNode and the restore completes. The build already contains the external snapshot feature (#50978 / #52476), so its own DataNode is capable of executing the task.

## Actual Behavior

The copy segment task stays pending forever and is silently re-queued every scheduling round (~100ms). There is no timeout and no error is surfaced to the caller — the restore job hangs at `progress=0` indefinitely.

## Error Logs

No error is raised. The only symptoms in datacoord logs:

```
[...]

PR #52638, opened 2026-08-18
fix: route external snapshot copies by task type

weiliu1031:
issue: #52575

## What changed

- Removed DataNode-version filtering from copy-segment scheduling.
- Kept local restore on `CopySegment` and routed external restore through
  `ExternalCopySegment`.
- Retained payload-based external detection for older DataCoord clients.
- Made DataNode return `ServiceUnimplemented` for unknown task types.
- Normalized the exact legacy code-5 rejection for `ExternalCopySegment` to
  the same unsupported-task error.
- Failed external restore immediately when the worker does not support the
  task type, while keeping other worker errors retryable.
- Retained `QuerySlotResponse.version` for wire compatibility without using
  it for worker capability selection.

## Why

Cloud worker endpoints can be pooled, so the version returned by a slot
endpoint does not reliably identify the DataNode that executes the task.
Version-based filtering can therefore exclude compatible workers
indefinitely.

The task type now acts as the capability boundary. New workers return a typed
unsupported error for unknown task types. DataCoord also recognizes the exact
legacy rejection returned by older workers, so an incompatible external
restore fails immediately instead of being requeued until the job timeout.
Transient worker failures remain retryable. Local restore keeps its existing
task type and behavior.

This change intentionally does not add failed-worker exclusion or cross-worker
rerouting to the scheduler.

## Validation

- Rebasing onto `milvus/master` preserved the complete feature patch ID and
  changed-file set without conflicts.
- Targeted race tests passed after the rebase for `pkg/taskcommon`,
  `pkg/util/merr`, `internal/datacoord/session`, `internal/datacoord`, and
  `internal/datanode` with `dynamic,test` tags and disabled inlining.
- The tests cover typed unknown-task rejection, legacy RPC status round-trip,
  DataCoord fail-fast behavior, and retry preservation for unrelated errors.
- `git diff --check` passed.

The post-rebase local C++ build is blocked by an unrelated `milvus/master`
macOS/clang 19 error in `ConcurrentVector.cpp`, where `std::min` receives
`int64_t` and `ssize_t`. This PR does not modify `internal/core`.


liliu-z (commented):
/lgtm
/approve

congqixia (commented):
NIT (docs/upgrade guidance): The version-gate removal changes rolling-upgrade behavior in a way worth documenting explicitly. With a new DataCoord against old DataNodes, an external restore now fails fast (job marked failed) instead of staying pending until a compatible worker appears — a deliberate trade-off, but operators in a mixed fleet will see failures rather than a wait. Conversely, with an old DataCoord against new DataNodes, the old DataCoord's own 3.0.1 version gate still hangs the restore forever (new nodes report an empty QuerySlotResponse.version), so the fix only takes effect once DataCoord is upgraded. Please add a short upgrade-order note to the issue/PR (e.g., "upgrade DataCoord and DataNodes together; external restore only works once DataCoord is on this build") so this fail-fast-vs-hang asymmetry isn't a surprise during rollout.

congqixia (commented):
/lgtm

congqixia (approved):
/approve

PR #52639, opened 2026-08-18
fix: [3.0] route external snapshot copies by task type

weiliu1031:
pr: #52638
issue: #52575

## What changed

Backports the external snapshot restore compatibility fix from master:

- Removes QuerySlot worker-version filtering.
- Keeps local restore on `CopySegment`.
- Uses `ExternalCopySegment` as the authoritative external restore signal.
- Retains payload-based external detection for older DataCoord clients.
- Returns `ServiceUnimplemented` when a DataNode does not recognize a worker
  task type.
- Normalizes only the legacy code-5 rejection for `ExternalCopySegment`.
- Fails an external restore immediately when the worker lacks that capability;
  unrelated transient errors remain retryable.
- Rejects malformed external tasks without a valid source root URI.
- Allows DataNodes to create, query, and drop either task type.
- Keeps `QuerySlotResponse.version` for wire compatibility without using it
  for worker capability selection.

## Why

Cloud worker endpoints can be pooled, so the version returned by a slot
endpoint does not reliably identify the DataNode that executes the task. That
can filter compatible workers indefinitely.

The task type now acts as the capability boundary. Older workers reject the
unknown external task type before decoding its payload. Their legacy
`ServiceInternal` response is normalized at the DataCoord RPC boundary so the
copy task and restore job fail immediately instead of being requeued until the
job timeout. New workers return `ServiceUnimplemented` directly. Local restore
keeps its existing task type and behavior.

This change intentionally does not add failed-worker exclusion or cross-worker
rerouting to the scheduler.

## Backport notes

The branch is rebased onto 3.0 commit `95db0b5329`. The complete feature
patch-id and touched-path set are unchanged across the rebase.

The moving 3.0 base introduced a DataNode credential-redaction audit case that
still called the removed `CopySegment` wrapper. The test calls
`copySegment(..., false)`, matching master and preserving the local-copy path
being audited.

The latest 3.0 base contains
https://github.com/milvus-io/milvus/pull/52710, which independently fixed the
Proxy credential-redaction test for the injected MetaCache API. The equivalent
local test-only commit was dropped during rebase instead of duplicating that
change.

## Validation

- Post-rebase targeted race tests with `dynamic,test` tags passed for
  `pkg/taskcommon`, `pkg/util/merr`, `internal/datacoord/session`,
  `internal/datacoord`, and `internal/datanode`.
- The targeted set includes merr classification/wire guards, legacy
  `ServiceInternal` RPC round-trip normalization, external unsupported
  fail-fast, and transient/local retry boundaries.
- `git diff --check` passed.
- Full-package local runs still reach the documented macOS-only DataCoord
  DISKANN size mismatch and DataNode analyze centroid-content failure; neither
  affected path is modified here.


congqixia (approved):
/lgtm

liliu-z (commented):
/lgtm
/approve

PR #52640, opened 2026-08-18
fix: [3.0] run MinHash schema validation on all DDL paths and reject TEXT input

MrPresent-Han:
Cherry-pick of #52219 to the 3.0 branch.

### What

Admission gaps around MinHash functions, each letting DDL persist collections that can never work:

- **`disableRuntimeCheck=true` paths skipped MinHash parameter validation.** Hoist `ValidateMinHashFunction` into `validator.ValidateFunction` so it runs regardless of the runtime-check switch; drop the now-redundant special case in the embedding executor.
- **The direct RootCoord CreateCollection path ran no function validation at all.** Run `validator.ValidateFunction(schema, "", true)` after `assignFieldAndFunctionID`, scoped to newly created schemas (the `preserveFieldID` restore path and external collections are exempt).
- **The validation itself was overflow-bypassable.** Compare via division (`outputDim%32 == 0 && outputDim/32 == numHashes`) instead of `outputDim != int64(numHashes*32)`, which wraps for large `num_hashes`.
- **`CheckFunctionInputField` admitted TEXT inputs for MinHash, contradicting the VarChar-only runner contract.** Align admission to VARCHAR-only.

### Tests

- `Test_createCollectionTask_prepareSchema_validatesFunctions`, `Test_createCollectionTask_prepareSchema_skipsFunctionValidationForExternal`
- `TestValidateMinHashFunctionRejectsNumHashesOverflowingDimCheck`, new `validator_test.go`
- `TestValidateFunctionInputField` MinHash VarChar-accepted / TEXT-rejected cases

issue: #51649
pr: #52219

### DIFF
--- a/internal/proxy/util_test.go
+++ b/internal/proxy/util_test.go
@@ -3362,6 +3362,31 @@ func TestValidateFunctionInputField(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("Valid MinHash function input - varchar", func(t *testing.T) {
+		function := &schemapb.FunctionSchema{
+			Type: schemapb.FunctionType_MinHash,
+		}
+		fields := []*schemapb.FieldSchema{
+			{DataType: schemapb.DataType_VarChar},
+		}
+		err := validator.CheckFunctionInputField(function, fields)
+		assert.NoError(t, err)
+	})
+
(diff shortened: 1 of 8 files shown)


MrPresent-Han:
Superseded by #52944, which carries the same #52219 backport combined with #52902 and the required L2 assertion update (the `expected dim 512` message change) — merging this PR alone would break the 3.0 nightly.

Issue #52620, opened 2026-08-18
[Bug]: Loading v3.0.0-written segments on a 3.0 daily build fails with 'Meta key not found: index_type'


yanliang567:
## Summary

Upgrading a standalone Milvus (RocksMQ) from `v3.0.0` to the 3.0 daily build
`3.0-20260817-beb93ec2` leaves baseline collections unserviceable:
`channel distribution is not serviceable: channel not available`.

## Root cause

`channel not available` is a symptom. Loki shows the QueryNode failing to load
segments:

```
[WARN] [task/executor.go:281] ["failed to load segment"]
  error="At LoadSegment: At Load: Assert \"meta_json_.contains(key)\"
  => Meta key not found: index_type
  at ../internal/core/src/storage/IndexEntryReader.h:128: segcore error[segcoreCode=2001]"
```

which then cascades:

```
[WARN] [querynodev2/services.go:532] ["delegator failed to load segments"] [collectionID=468441322874996909] [segmentID=468441322876458798]
[WARN] [delegator/delegator_data.go:706] ["worker failed to load segments"]
[INFO] [dist/dist_handler.go:337] ["delegator is not serviceable"]
```

The v3.0.0 (base) persisted index metadata lacks the `index_type` key that the
3.0 daily build's `IndexEntryReader` requires. This is the same
`Meta key not found` family as #52359 (which was missing `version` /
`index_length`), but here the missing key is `index_type` on the
`v3.0.0 -> 3.0 daily` read path.

## Milvus config (unchanged across the upgrade)

The workflow's `assert-milvus-storage-config` step validated the storage config
at both phases as identical (and passed) before the serviceability wait:

| Phase | Image | Version | common.storage.jsonShreddingEnabled | useLoonFFI | dataNode.storage.format |
|---|---|---|---|---|---|
| base | `v3.0.0` (pinned digest) | 3.0.0 | `false` | disabled (absent) | parquet (default) |
| target | `3.0-20260817-beb93ec2` | 3.0.0 | `false` | disabled (absent) | parquet (default) |

No LoonFFI / storage v3, no Vortex, and no JSON Shredding were enabled at any
phase before the failure (the JSON Shredding post-upgrade toggle happens later
in the DAG). The failure is therefore a persisted-index-format compatibility
issue, not a storage-config change.

## Environment

- Base image: `harbor.milvus.io/milvusdb/milvus:v3.0.0` (pinned digest)
[...]

PR #52642, opened 2026-08-18
fix: tolerate legacy STLSORT physical files under HYBRID nested index metadata

congqixia:
issue: #52620

Struct sub-field scalar indexes use the nested index factory because their logical field type is an array. In 3.0.0, IndexFactory::CreateNestedIndex routed HYBRID through the default sort-index factory, so a HYBRID collection
index on a struct-array sub-field (e.g. items[items[category]], element VarChar) produced a standalone milvus_packed_stlsort_index.v3 whose meta lacks the hybrid index_type key.

The load path is driven by collection metadata HYBRID, so it builds a HybridScalarIndex whose LoadEntries unconditionally read the index_type meta, hitting "Meta key not found: index_type" (segcoreCode=2001) and failing segment load. The build side was already fixed; existing bad files still failed to load.

Make HybridScalarIndex::LoadEntries tolerant: when the index_type meta key is absent, recover the physical index type from the packed filename ("milvus_packed_<type>_index.v3", chosen by the writer from GetIndexType()), falling back to the file meta keys only when the name is unrecognized. The filename is an authoritative discriminator and stays correct as new physical index types are added, whose meta keys may reuse names like "version" or "file_names". No reindexing or object-store rewrite needed.

Add regression tests: a standalone STLSORT file built via the nested index factory loads through the HYBRID path and infers STLSORT from the filename, and a proper hybrid file with the index_type meta still loads through the
HasMeta branch.

### DIFF
--- a/internal/core/src/index/BitmapIndexArrayTest.cpp
+++ b/internal/core/src/index/BitmapIndexArrayTest.cpp
@@ -34,6 +34,7 @@
 #include "common/protobuf_utils.h"
 #include "gtest/gtest.h"
 #include "index/BitmapIndex.h"
+#include "index/HybridScalarIndex.h"
 #include "index/Index.h"
 #include "index/IndexFactory.h"
 #include "index/IndexInfo.h"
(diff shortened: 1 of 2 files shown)


zhengbuqian (approved):
/lgtm

Issue #1, opened 2019-09-16
Add CHANGLOGS.md and CONTIRBUTING.md


JinHai-CN:
As title.

JinHai-CN:
done.

PR #52653, opened 2026-08-19
feat: add QueryView legacy query client

chyezh:
issue: #40451

- query protocol: add Primary-only query plans and ViewQuery RPC contracts with typed QueryView status details
- legacy query client: implement two-phase Search and Query orchestration with primary routing, shard retries, MVCC projection, and raw result collection
- query client design: document the Legacy-only first-stage scope and defer discovery, node transports, and Proxy integration


tinswzy:

Reviewed at head `fac2e63f47` (5 commits, base `aae033f76a`). Scope confirmed: I grepped the whole tree — nothing outside `internal/views/queryclient/` and `internal/views/viewerror/` references these packages, and no Proxy/StreamingNode/QueryNode code registers `QueryPlanService` or `ViewQueryService`. So there is no production traffic on this path, exactly as the PR body and `query_client.md` §1 claim. The layering (transport-neutral orchestration + injected `ShardResolver`/`ReplicaPicker`/transports), the `ShardID`-keyed collector with `ResetShard` rollback, and the test discipline are all good.

Not blocking merge as a draft scaffold, but **two of the findings below are wire-contract decisions that get much more expensive to change once node-side handlers exist**, and both currently resolve to "return wrong results with `Status = Success`". I would fix #1 and #2 in this PR.

---

## high

### 1. A plan with no `mvcc` is accepted, and `mvcc_timestamp = 0` means "ignore the plan" to the node

`internal/views/queryclient/shard_client.go:278`
```go
func legacyMVCCForNode(mvcc *viewpb.QueryPlanMVCC, node qviews.WorkNode) uint64 {
	if mvcc == nil {
		return 0
	}
```
`validateQueryPlan` (`shard_client.go:188`) requires a non-nil `shard_id` and `version`, but never `mvcc`. So `0` is silently forwarded as the projected MVCC.

Zero is not a neutral value at the consumer. In the QueryNode delegator — the intended Phase-2 consumer of `internalpb.SearchRequest`/`RetrieveRequest` — `internal/querynodev2/delegator/delegator.go:613` (search) and `:748` (query):
```go
// Normal reads use tSafe only when the request does not provide an MVCC timestamp.
if sd.skipStreamingForExternalTable || req.GetReq().GetMvccTimestamp() == 0 {
	req.Req.MvccTimestamp = tSafe
}
```

**Failure scenario:** a planner returns a `QueryPlan` with `mvcc` unset (or with a legitimately-zero `transforming_timetick` on a freshly-created view). The StreamingNode and every QueryNode each substitute their *own local* `tSafe`. The two frontiers are unsynchronized, so rows in the growing/transforming overlap are returned twice by one node and once by the other, or dropped by both — with `Status = Success` and no way for the client to detect it. The single-frontier-per-shard guarantee that motivates the whole two-phase protocol is voided by a missing proto field.

This is currently locked in by tests: `shard_client_test.go:77` asserts `require.Zero(t, retrieveRequest.GetMvccTimestamp())` for a plan with no `Mvcc`, and `legacy_client_test.go:151-158` builds a plan with no `Mvcc` and expects success.

**Fix:** in `validateQueryPlan`, reject `plan.GetMvcc() == nil`; reject `growing_timetick == 0`, and reject `transforming_timetick == 0` whenever the plan contains a QueryNode; assert `transforming_timetick <= growing_timetick`. Update the two tests to carry a real `Mvcc`.

### 2. An empty `work_nodes` list is a successful, empty shard result

`internal/views/queryclient/shard_client.go:204`
```go
nodes := make([]qviews.WorkNode, 0, len(plan.GetWorkNodes()))
```
With zero entries the loop body never runs, `fanOut` (`:241`) returns `group.Wait()` on an empty group = `nil`, and `execute` returns a successful `ShardPlan` at `:177`. `shard_client_test.go:106` (`TestFanOutWithNoWorkNodes`) and `query_client.md:105` ("An empty work-node list is valid and returns an empty result for that shard") document this as intentional — which is exactly why it is worth challenging now.

**Failure scenario:** a StreamingNode holds a view record but has no work nodes assigned yet (view rebuilding, replica transfer, partially-recovered view, or simply a planner bug that drops the node list). It answers `GetQueryPlan` with `work_nodes: []`. On a 4-shard collection the client returns success having queried 3 shards; a quarter of the data is silently missing, and the response is indistinguishable from "no matching entities". No error, no retry, no metric, nothing in the `ShardPlan` a caller could assert on.

"No work nodes" and "no data" are different facts and must not share an encoding. **Fix:** require `len(work_nodes) > 0` in `validateQueryPlan`. If a genuinely empty shard is a real state, express it explicitly — either an `empty_shard` bool on `QueryPlan`, or have the planner return the retryable `VIEW_CODE_VIEW_NOT_FOUND` while the view is not serviceable.

The same class of silent-empty exists one level up: `legacy_client.go:21` — if `ResolveVChannels` returns an empty slice, `group.Wait()` returns nil and `Search` returns success with zero results. Worth an explicit guard.

---

## medium

### 3. `mvcc_timestamp` is projected, `guarantee_timestamp` is not — the node's wait barrier is not aligned with the plan

`shard_client.go:258-276` sets only `cloned.MvccTimestamp`; `GuaranteeTimestamp` stays whatever the plan echoed back. But the delegator waits on the *guarantee* timestamp: `delegator.go:737`, `tSafe, err := sd.waitTSafe(ctx, req.Req.GetGuaranteeTimestamp())`.

**Failure scenario:** the plan says `transforming_timetick = T`. A QueryNode whose tSafe is behind `T` and whose `GuaranteeTimestamp` is `0` or older does not wait, `PinReadableSegments` returns what it currently has, and segcore filters at `T` — producing a truncated result with `Status = Success`. Silent partial results under normal node lag.

**Fix:** set `cloned.GuaranteeTimestamp` to the projected MVCC alongside `MvccTimestamp`, or state in `query_client.md` §2.2 that Phase-2 handlers must wait on `mvcc_timestamp` and add that wait node-side. Either way the invariant needs an owner.

### 4. Error classification is inverted in both directions at the boundary

`shard_client.go:310`
```go
if viewErr := viewerror.TryAsViewError(err); viewErr != nil {
	return merr.WrapErrServiceUnavailableErr(err, "%s", operation)
}
return merr.WrapErrServiceInternalErr(err, "%s", operation)
```
The branch keys on *"is it a ViewError"*, not on *"is it retryable"* — even though `viewerror/error.go:25` already answers the latter.

- A transport failure (gRPC `Unavailable` from a rolling StreamingNode, connection reset) carries no `ViewError` detail → falls to `ErrServiceInternal`, which is **not** retriable. `legacy_client_test.go:215-228` pins this behavior.
- A `VIEW_CODE_UNKNOWN`, which `ViewError.IsRetryable()` explicitly reports as **false**, becomes `ErrServiceUnavailable`, which `merr.IsRetryableErr` reports as **true** (asserted at `legacy_client_test.go:307`).

**Failure scenario:** a 2-second StreamingNode restart fails the whole query permanently and nothing upstream retries it; meanwhile a deterministic planner bug reported as `UNKNOWN` is retried in a loop by the Proxy/SDK. Exactly backwards.

**Fix:** derive the mapping from `viewerror.AsViewError(err).IsRetryable()` — retryable → `ErrServiceUnavailable`, non-retryable view code → `ErrServiceInternal`, and classify gRPC `Unavailable`/`DeadlineExceeded`-class transport errors as `ErrServiceUnavailable`.

### 5. Retry has no backoff and no way to invalidate stale discovery

`shard_client.go:140`
```go
for attempt := 0; attempt < c.maxRetries; attempt++ {
```
On a retryable error the loop `continue`s immediately (`:161`, `:174`) and re-calls `c.shardResolver.ResolveShard`.

**Failure scenario:** `VIEW_CODE_NOT_PRIMARY` after a primary handoff. The resolver is a cache (as any production discovery will be); all three attempts run within microseconds against the *same* stale entry and always exhaust into `ErrServiceUnavailable`. Contrast with the legacy path, which uses `retry.Do` with backoff plus `lb_policy` node exclusion. `TestLegacySearchRetryExhaustion` passes precisely because the mock is stateless.

**Fix:** add backoff between attempts, and give `resolver.ShardResolver` an explicit `Invalidate(collectionID, vchannel)` (or version-token) so `NOT_PRIMARY` forces a refresh rather than re-reading the cache.

### 6. Phase 1 echoes the full legacy request — including `placeholder_group` — back to the client

`view.proto`: `GetQueryPlanRequest.request` and `QueryPlan.request` both embed `internal.SearchRequest`; `shard_client.go:259-265` clones `plan.GetLegacySearchRequest()` per node.

**Failure scenario:** the query vectors cross the wire `client→SN` + `SN→client` (echoed in the plan) + `client→each of N work nodes` = **N+2** times per shard, versus N in the legacy path. `nq=1000` at 768 float dims is ~3 MB; a 4-shard collection with 3 work nodes per shard moves ~60 MB per query and puts a full-payload serialize/deserialize of the plan response on the critical path — before any search work starts.

**Fix:** return only the optimization *delta* in `QueryPlan` (rewritten `serialized_expr_plan`, segment filter, BM25 IDF) and have the client reuse the request body it already holds; or make the plan carry a server-side request handle. This is a proto-level decision, best made before handlers exist.

### 7. `isEmptySuccessfulQueryResponse` drops scan-cost accounting with the result

`shard_client.go:295`
```go
return result.GetAllRetrieveCount() == 0 &&
	typeutil.GetSizeOfIDs(result.GetIds()) == 0 &&
	len(result.GetFieldsData()) == 0
```
It does not check `ScannedRemoteBytes`, `ScannedTotalBytes`, or `CostAggregation` — all three of which the Proxy accumulates from every `RetrieveResults` (`internal/proxy/task_query.go:998-1002`).

**Failure scenario:** a highly selective filter query scans 10 GB and matches 0 rows. The response satisfies the predicate above, so the whole `RetrieveResults` is discarded and that shard's scanned bytes and `TotalRelatedDataSize` never reach the Proxy — under-reporting query cost, storage-cost metrics, and anything built on them once this path is wired up.

**Fix:** also require the cost fields to be zero before dropping, or keep the result and let the reduce stage skip empty payloads (the existing Proxy reduce already tolerates empty `RetrieveResults`).

---

## low

### 8. Mixed fan-out failures make the retry decision nondeterministic

`shard_client.go:255`, `return group.Wait()` — `errgroup` keeps only the *first* error. If node A fails with `VIEW_CODE_VIEW_INVALIDATED` (retryable) and node B fails with a hard `merr` error, which one wins depends on scheduling: sometimes the shard retries and eventually reports `ErrServiceUnavailable`, masking B's real cause; sometimes it fails immediately with B. Same inputs, two different errors.

**Fix:** collect all node errors, retry only if *every* error is retryable, and always surface the most specific non-retryable one.

### 9. Raw results cannot be joined back to the `ShardPlan` that produced them

`collector.go:51` / `:90` flatten a map into a slice; `query_client_legacy.md:52` documents the order as unspecified. But the same doc calls `ShardPlan` "execution evidence", and a later requery/partial-result stage — or reconstructing per-channel MVCC (`SearchResults.ChannelsMvcc`) — needs the association. Returning `map[qviews.ShardID][]*internalpb.SearchResults`, or tagging each entry with its `ShardID`, costs nothing now and is awkward to retrofit.

### 10. Session is silently upgraded to Strong, and the node then sees a different level

`shard_client.go:288` maps `Session → Strong` for the plan request, while the cloned legacy request still carries `ConsistencyLevel_Session` — so `speedupGuranteeTS` (`delegator.go:1180`) takes its non-Strong branch. Two different consistency levels describe one execution. Beyond the inconsistency, every Session query now pays a full Strong-consistency plan (wait for the latest WAL tick) instead of the caller's session guarantee timestamp. `query_client.md:82` documents the intent, but passing the caller's `guarantee_timestamp` through the `GetQueryPlanRequest.mvcc` oneof would preserve the semantics instead of coercing them.

### 11. `ReplicaPicker` is an injected seam the executor unconditionally rejects

`client.go:84` accepts a `ReplicaPicker`, and `shard_client.go:152` rejects any pick that is not `replicas.PrimaryShardID`. The guard is right for this milestone, but exporting the interface implies replica choice is pluggable when it cannot be. Suggest unexporting it until secondary routing lands, or documenting the constraint on the interface itself.

---

## nit

- **`viewerror/error.go:13`** — `type ViewError viewpb.ViewError` does not implement `proto.Message`, and any `protoc-gen-go` regeneration that renames or retypes `Code`/`Cause` breaks the `(*viewpb.ViewError)(e)` conversion at a distance with no compile-time link to the proto. Consider `struct{ pb *viewpb.ViewError }`. Separately, `Error()` at `:15` dereferences `e.Code` with no nil guard while `IsRetryable()` at `:25` has one — `TryAsViewError` can return a typed nil in principle.
- **`viewerror/rpc_error.go:41`** — `ViewClientStatus` has no `Unwrap()`, so `ConvertViewError` terminates the error chain: after conversion `errors.Is(err, merr.ErrXxx)` and `merr.IsMilvusError` are false even when the server returned a Milvus status. Add `func (s *ViewClientStatus) Unwrap() error { return s.Status.Err() }`.
- **`view.proto` `GetQueryPlanRequest.partition_ids`** duplicates `legacy_*_request.partition_ids` (`shard_client.go:58`, `:98`) with no stated precedence. Either drop the top-level field or document it as authoritative.
- **`client.go:18`** — `MaxRetries` counts *attempts*, not retries (`MaxRetries: 1` = one attempt, zero retries; `query_client.md:126` acknowledges this). Rename to `MaxAttempts`.
- **`shard_client.go:50`** — `buildPlanRequest` embeds the caller's `*internalpb.SearchRequest` by reference into the plan request for every shard goroutine. Worth a doc comment on `LegacySearchRequest` that the caller must not mutate the request after handing it over.

---

Happy to re-review once #1 and #2 are settled — those are the two that change the wire contract, and everything else can land incrementally alongside the deferred discovery/transport work.

PR #29, opened 2019-10-17
Update README.md #27

del-zhenwu:

PR #52658, opened 2026-08-19
enhance: add opt-in UAX #29 tokenization and ICU position modes

aoiasd:
## Summary

- Preserve the existing alphanumeric token boundaries used by `standard`, default, `english`, and `arabic` analyzers with a self-contained streaming scanner.
- Add an explicit `uax29: true` option on the `standard` tokenizer for UAX #29 word and emoji boundaries; `max_token_length` is available only in this mode.
- Add two independent ICU options:
  - `position_mode`: `char` by default or `token`, controlling only position calculation.
  - `token_types`: an optional non-empty list selecting ICU4X native word types from `letter`, `number`, and `none`. When omitted, all types are emitted to preserve existing behavior. `none` represents ICU non-word segments, including whitespace, punctuation, symbols, and emoji.

For word-only token positions, use:

```json
{
  "type": "icu",
  "position_mode": "token",
  "token_types": ["letter", "number"]
}
```

Existing indexes remain compatible while the new options retain their defaults. Enabling UAX #29 or changing either ICU option requires rebuilding the field index.

## Testing

- `cargo fmt --all -- --check`
- `cargo check --lib`
- Public analyzer behavior probe covering each native ICU token type, separator and emoji filtering, CJK dictionary segmentation, consecutive token positions, and invalid type validation
- Rust unit coverage added for legacy defaults, independent position/type selection, native type validation, and CJK positions

The local `cargo test --lib` target is currently blocked by unrelated existing test-compilation errors in Lindera URL shuffling and index-writer tests.


internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/tokenizer.rs
```diff
@@ -6,21 +6,39 @@ use tantivy::tokenizer::{TextAnalyzer, TextAnalyzerBuilder};
...
 use super::{
     CharGroupTokenizer, GrpcTokenizer, IcuTokenizer, JiebaTokenizer, LangIdentTokenizer,
-    LinderaTokenizer, ThaiTokenizer,
+    LinderaTokenizer, StandardTokenizer, ThaiTokenizer,
 };
 
 use crate::error::{Result, TantivyBindingError};
 
 pub fn standard_builder() -> TextAnalyzerBuilder {
+    TextAnalyzer::builder(StandardTokenizer::default()).dynamic()
```
buqian-zilliz, line 15: **[P1] standard_builder() switch also silently changes english/arabic/default analyzers**

`standard_builder()` now builds `StandardTokenizer` instead of `SimpleTokenizer`, but this helper is shared beyond the `standard` tokenizer: `standard_analyzer`, `english_analyzer`, and `arabic_analyzer` in `build_in_analyzer.rs` all call it, and the empty-params default path in `create_analyzer` also uses `standard_analyzer`. As a result the `english`/`arabic` built-in analyzers and the default analyzer change tokenization too, not just `standard`: e.g. `dog's` and `32.3` become single tokens, and CJK text now splits per character (`中文` becomes `中`, `文`) instead of staying as one token. The design doc only covers the `standard` tokenizer and offers `simple` as the escape hatch, but there is no equivalent way to restore the old behavior for `english`, `arabic`, or the default analyzer, and there is no rebuild/version guard. An index built with these analyzers will silently disagree with the new query analyzer and produce false negatives. Please either keep these built-ins on the old tokenizer or explicitly document and guard this additional breakage.

internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/standard_tokenizer.rs
```diff
@@ -0,0 +1,294 @@
...
+                &mut position,
+            );
+        }
+
+        tokens
+    }
+}
+
+fn is_word_or_emoji(segment: &str) -> bool {
+    segment.chars().any(char::is_alphanumeric) || segment.chars().any(is_emoji_character)
```
buqian-zilliz: **[P3] is_word_or_emoji double-scans segments and repeats emoji property lookups in the hot path**

`is_word_or_emoji` runs two separate `segment.chars().any(...)` passes, and `is_emoji_character` performs up to three `CodePointSetData::...().contains()` trie lookups per character. For punctuation/whitespace segments (very common), the first pass scans the whole segment and returns `false`, then the second pass scans it again and performs three property lookups for every non-alphanumeric character. This runs for every indexed document and query. Consider a single pass such as `segment.chars().any(|c| c.is_alphanumeric() || is_emoji_character(c))` and hoisting the three `CodePointSetData` singletons out of the per-character function (e.g. into `once_cell`/`lazy_static` or fields on `StandardTokenizer`).

internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/tokenizer.rs
aoiasd: Fixed in df0e2befa8. StandardTokenizer now delegates directly to Tantivy SimpleTokenizer unless the tokenizer object explicitly sets `uax29: true`. Because `standard_builder()` uses that default, the standard/default/english/arabic built-ins retain their existing token boundaries. I added exact Token parity coverage plus regression tests for all four built-in analyzer paths, and updated the compatibility section of the design doc.

internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/standard_tokenizer.rs
aoiasd: Fixed in df0e2befa8. `is_word_or_emoji` now performs one character pass, and the three ICU4X property datasets are hoisted into static borrowed constants instead of being constructed in the per-character function.

internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/tokenizer.rs
aoiasd: Follow-up in c2bbb82afc: after simplifying the implementation, production code no longer holds or delegates to Tantivy `SimpleTokenizer`. `StandardTokenizer` now has a self-contained streaming alphanumeric scanner for the default path and a separate streaming UAX #29 scanner behind `uax29: true`. The parity tests still compare complete Token values against Tantivy `SimpleTokenizer`, and the small-change design document has been removed.

internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/icu_tokneizer.rs
```diff
@@ -59,17 +97,21 @@ impl IcuTokenizer {
                 continue;
             }
             let token_str = &text[offset..breakpoint];
+            let token_length = match self.position_mode {
+                IcuPositionMode::Char => token_str.chars().count(),
+                IcuPositionMode::Token => 1,
```
buqian-zilliz, line 131: **[P2] ICU `position_mode: "token"` advances positions for separator tokens**

In `IcuTokenizer::tokenize`, the ICU word segmenter emits whitespace/punctuation segments as tokens (for example `"hello world"` yields `hello`, `" "`, `world`). With `position_mode: "token"`, every emitted token gets `position_length = 1` and advances the position counter, so `"hello world"` produces positions `0, 1, 2` with the space consuming slot `1`. This diverges from the new `StandardTokenizer` UAX #29 path (and Lucene's `ICUTokenizer`), which only advance positions for actual word/emoji terms, producing `hello` at `0` and `world` at `1`. Because `position`/`position_length` drive phrase matching, `token` mode will expose unexpected position gaps unless separator segments are suppressed or excluded from position advancement. Please confirm the intended semantics: should `token` mode count separator tokens, or should it match standard/Lucene positions?

aoiasd: Updated in d48168d6fb. Position calculation and segment selection remain independent. `position_mode` (`char` or `token`) controls only how positions advance. `token_types` now maps only ICU4X native word types: `letter`, `number`, and `none`; omitting it retains all segments for compatibility. `none` covers ICU non-word segments, including whitespace, punctuation, symbols, and emoji.

With `position_mode: "token"` alone, `hello world` still emits the space at positions 0/1/2. With `token_types: ["letter", "number"]` alone, `none` segments are omitted while character positions retain their original gaps. Enabling both produces word-only `hello@0, world@1`. CJK dictionary tokens remain preserved.

internal/core/thirdparty/tantivy/tantivy-binding/src/analyzer/tokenizers/icu_tokneizer.rs
```diff
@@ -43,39 +80,135 @@ impl IcuTokenizer {
...
+        })?;
+        if !token_types.contains(&token_type) {
+            token_types.push(token_type);
+        }
+    }
+    Ok(token_types)
+}
+
+fn classify_token(word_type: WordType, text: &str) -> IcuTokenType {
+    if is_emoji(text) {
```
buqian-zilliz: **[P3] `classify_token` runs a full-segment emoji scan for every token on the ICU hot path**

`IcuTokenizer::tokenize` calls `classify_token` for every ICU segment, and `classify_token` invokes `is_emoji(text)` first, which iterates the entire segment and performs up to three `CodePointSetData::contains` trie lookups per character. For segments the ICU segmenter already tags as `WordType::Letter` or `WordType::Number`, this scan is guaranteed to return `false` (letters/digits cannot be emoji), so it is pure overhead on every indexed document and every query. Consider matching `word_type` first and only calling `is_emoji` for `WordType::None` (where emoji sequences actually surface). The same loop also computes `char_length` via `token_str.chars().count()` unconditionally, but that value is unused in `IcuPositionMode::Token`.

aoiasd: Fixed in d48168d6fb. The custom ICU `emoji` token type and its full-segment Unicode property scan have been removed; `token_types` now exposes only ICU4X native `none`, `number`, and `letter` types, with emoji remaining `none`. The token loop also computes character counts only in `position_mode: "char"`; token mode uses a constant position length of 1. Native-type selection, emoji-as-`none`, CJK positions, and invalid `emoji` configuration are covered by tests and a public analyzer behavior probe.

Issue #52605, opened 2026-08-18
[Bug]: Native Group Search recall drops after delete/reinsert until compaction


ffffffff0:
## Environment

- **Milvus Version**: v2.6.22
- **Deployment Mode**: standalone (Docker)
- **MQ**: Woodpecker
- **SDK**: PyMilvus 2.6.x
- **Storage**: MinIO + etcd
- **Compaction config**: Milvus defaults; auto compaction enabled
- **Indexes tested**: dense HNSW and sparse/BM25, including hybrid search

### Option B: Steps

1. Create and load a collection containing:
   - 181 QA entities
   - 128 distinct values in a scalar `group_key` field
   - one vector per question
2. Build the indexes and run a fixed set of 74 queries with:
   - `limit=5`
   - `group_by_field="group_key"`
   - `group_size=1`
   - `strict_group_size=True`
3. Record the expected-answer Recall@5 for the fresh collection.
4. Delete the document/entities and insert the same 181 logical QA entities again.
5. Flush and wait until the new entities are searchable, but do not manually compact.
6. Run the same 74 queries with identical parameters.
7. Run the same queries without `group_by_field`.
8. Force compaction, wait for it to complete and for indexes to become ready, then run the grouped queries again.

Data-integrity checks between the fresh and delete/reinsert states:

- Visible row count and Question+Answer multiset are identical.
- Group membership is identical.
- No missing or orphan vectors were found.
- Minimum cosine similarity between corresponding dense vectors is 0.99999934.

### Controlled Experiment Design

We used the same source CSV, embedding model/configuration, collection schema, index configuration, search parameters, fixed 74-query evaluation set, and expected-answer labels throughout the experiment. The only state transition was delete/reinsert and then compaction.

The three compared states were:

1. **Fresh**: create a new collection and insert the 181 QA entities once.
2. **Reinserted, not compacted**: delete the document/entities, insert the same logical data again, flush, and query before manual compaction.
3. **Reinserted, compacted**: force compaction on state 2, wait for completion/index readiness, and query again.

We ran these controls to isolate the cause:

- **Data-equivalence control**: compared visible rows, Question+Answer multiset, group membership, vector presence, and corresponding dense-vector cosine similarity.
- **Search-path control**: ran the same queries with and without `group_by_field`.
- **ANN-budget control**: repeated the affected HNSW searches with `ef=20`, `50`, and `100`.
[...]

PR #51017, opened 2026-07-02
enhance: adapt segcore to knowhere iterator error-code interface

foxspy:
## Summary

knowhere is changing its iterator interface (zilliztech/knowhere#1699, tracked by zilliztech/knowhere#1674): `IndexNode::iterator::Next()`/`HasNext()` become `noexcept` and return `knowhere::expected<>` instead of potentially throwing C++ exceptions, so heavy computations deferred into iterator calls report errors through error codes.

This PR adapts the segcore call sites:

- `ChunkMergeIterator` (`QueryResult.h`): unwrap `expected<>` when refilling the merge heap in `Next()`/`seal()`; iterator errors are converted to `SegcoreError` via `AssertInfo` with the underlying error message.
- `CachedSearchIterator`: unwrap `expected<>` in the two iterator-consuming loops before `ConvertIteratorResult`.
- `EmptyVectorIterator` / `DiskEmptyVectorIterator`: implement the new signatures (return an error result instead of throwing on `Next()`).
- Unit tests that consume knowhere iterators directly (`test_indexing.cpp`, `SearchBruteForceSparseTest.cpp`).

Exec operators and other consumers of Milvus' own `VectorIterator` abstraction (`std::optional`-based) are unaffected.

**Draft**: depends on the knowhere interface change; will be un-drafted together with the `KNOWHERE_VERSION` bump once a knowhere release contains zilliztech/knowhere#1699 (compilation against the currently pinned knowhere version is expected to fail).

issue: #51016

PR #52663, opened 2026-08-19
fix: [2.6] preserve vector iterator chunk index after empty chunks

CLiqing:
issue: #52605
pr: #51017

`VectorIterator::seal()` assigns the iterator index used later by `Next()` to refill a chunk. In 2.6, that index is incremented only when `HasNext()` is true. If a delete bitset makes a leading physical chunk empty, the following non-empty chunks are stamped with compacted indices rather than their positions in `iterators_`.

For example, with `[empty, live, live]`, the two live chunks are incorrectly stamped `0` and `1` although their actual positions are `1` and `2`. When `Next()` consumes their first candidates, it refills `iterators_[0]` or `iterators_[1]` instead of the originating chunk. The later chunk therefore contributes only its first candidate and its remaining candidates are silently dropped.

This matches #52605: GroupBy is the path that uses vector iterators; delete/reinsert without compaction can leave a leading chunk entirely tombstoned; compaction removes those tombstones and restores the physical chunk layout. Dense and sparse search share this chunk-merge code.

master fixed the same chunk-index bug in #51017. Its full change also adapts segcore to a newer Knowhere error-code iterator API, which is not applicable to 2.6. This PR only backports the physical-index fix.

Adds a deterministic `[empty, live, live]` merge test and a growing delete/reinsert GroupBy regression test that asserts the expected four live offsets. `QueryResultTest.* + GroupBY.*` passes (7 tests); clang-format 12 and `git diff --check` pass.

### DIFF
--- a/internal/core/src/common/QueryResult.h
+++ b/internal/core/src/common/QueryResult.h
@@ -187,12 +187,15 @@ struct VectorIterator {
     void
     seal() {
         sealed = true;
-        int idx = 0;
-        for (auto& iter : iterators_) {
+        // idx must track the iterator's physical position. A bitset can make
+        // one chunk empty, but Next() still uses this index to refill from
+        // iterators_.
+        for (int idx = 0; idx < static_cast<int>(iterators_.size()); ++idx) {
+            auto& iter = iterators_[idx];
             if (iter->HasNext()) {
                 auto origin_pair = iter->Next();
                 auto off_dis_pair =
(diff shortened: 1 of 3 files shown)


liliu-z (commented):
/lgtm
/approve

ffffffff0:
@CLiqing @liliu-z  One question: does a flush also resolve this scenario, or is this only handled by compaction?

PR #52665, opened 2026-08-19
enhance: refactor merge operator for declarative construction

junjiejiangjjj:
- construct merge operators from representations with runtime search metrics
  - validate multi-input layouts and preserve native row and element identities
  - round-trip element indices through DataFrame conversion
  - remove synthetic hybrid element keys and the select operator
  - add design documents and regression coverage

issue: #https://github.com/milvus-io/milvus/issues/52319


internal/util/function/chain/repr.go
```diff
@@ -275,6 +275,10 @@ func (repr *ChainRepr) RefreshInfo() error {
 		repr.Operators[i].Inputs = inputs
 		repr.Operators[i].Outputs = outputs
 
+		_, ok := GetOperatorFactory(opType)
```
liliu-z, line 278: Deleting operator_select.go removes the only registration for the public `select` op type, and RefreshInfo hard-fails on unregistered types, so a request that still carries `{"op": "select"}` in its function-chain params now fails with `unknown operator type: select`. This is request-scoped metadata only (no persisted collection schema is bricked), so the blast radius is limited to clients hand-writing chain reprs against an undocumented surface; worth confirming intentional before merge rather than blocking.

internal/util/function/chain/converter.go
```diff
@@ -302,6 +308,15 @@ func FromSearchResultData(resultData *schemapb.SearchResultData, alloc memory.Al
 		}
 	}
 
+	// Import element indices as the fixed Int32 system column used by Merge.
+	// SearchResultData uses int64 on the wire, while QueryNode element indices
+	// originate as int32, so reject malformed out-of-range values explicitly.
+	if elementIndices := resultData.GetElementIndices(); elementIndices != nil {
```
liliu-z, line 321: `converter.go:314` imports `$element_indices` unconditionally (`element_level_` is set at plan-parse time for any struct sub-vector ANN search, not just hybrid), so a single search routed through `searchWithRerankPipe` (reduce → rerank → pick → organize → result) has no `elementBestCollapseOp` node. MergeOp's dedup key therefore silently flips from PK to `(pk, element_index)`, and `organizeOperator`'s last-wins `offsets[pk] = i` map serves the field data of the last element row for that PK to every returned row — a `limit=10` that previously returned 10 distinct PKs now returns 10 `(pk, element)` rows with mismatched fields. If the new cardinality is intended, `organizeOperator` needs the same `(pk, element)` keying that `hybridAssembleOperator` received in this PR.

internal/util/function/chain/repr.go
junjiejiangjjj: Confirmed intentional. L0 already rejects `select`, and the public L2 API/SDK only supports `map`, `sort`, and `limit`. The impact is limited to hand-written L2 chain representations using an undocumented operator.

internal/util/function/chain/converter.go
junjiejiangjjj: The new `(PK, element_index)` cardinality is intentional for Struct Array element-level search; the previous PK-only dedup incorrectly collapsed distinct elements. Good catch on the downstream mismatch — `organizeOperator` should preserve the same composite identity, and we’ll add regression coverage for it.

liliu-z (commented):
/lgtm
/approve

PR #52666, opened 2026-08-19
fix: [2.6] upgrade librdkafka to 2.6.1 for SASL/SCRAM against Kafka 4.x

czs007:
issue: #52601
pr: #52602

when building the SCRAM client-final-message (present since v0.0.99). Kafka
4.x restored the strict nonce comparison required by RFC 5802, so brokers on
the 4.x line reject the proof. Fixed upstream in librdkafka 2.6.1
(confluentinc/librdkafka#4895). 2.6.0 still fails, so 2.6.1 is the lowest
usable version.

requires a minimum of 1.9.0.

The conan 1.x recipe is published to default-conan-local as librdkafka/2.6.1
(milvus-io/conanfiles#180, recipe revision 5b0ecc7c75e262b3e22d652018bedb93),
so this pins that revision.


czs007:
rerun go-sdk

czs007:
rerun go-sdk

PR #52668, opened 2026-08-19
fix: [hotfix-2.5.13] upgrade librdkafka to 2.6.1 for SASL/SCRAM against Kafka 4.x

czs007:
issue: #52601
pr: #52602 (master), #52666 (2.6)

when building the SCRAM client-final-message (present since v0.0.99). Kafka
4.x restored the strict nonce comparison required by RFC 5802, so brokers on
the 4.x line reject the proof. Fixed upstream in librdkafka 2.6.1
(confluentinc/librdkafka#4895). 2.6.0 still fails, so 2.6.1 is the lowest
usable version.

requires a minimum of 1.9.0.

The conan 1.x recipe is published to default-conan-local as librdkafka/2.6.1
(milvus-io/conanfiles#180, recipe revision 5b0ecc7c75e262b3e22d652018bedb93),
so this pins that revision.

### DIFF
--- a/internal/core/conanfile.py
+++ b/internal/core/conanfile.py
@@ -40,7 +40,7 @@ class MilvusConan(ConanFile):
         "folly/2023.10.30.08@milvus/dev#81d7729cd4013a1b708af3340a3b04d9",
         "google-cloud-cpp/2.5.0@milvus/2.4#c5591ab30b26b53ea6068af6f07128d3",
         "opentelemetry-cpp/1.8.1.1@milvus/2.4#7345034855d593047826b0c74d9a0ced",
-        "librdkafka/1.9.1#e24dcbb0a1684dcf5a56d8d0692ceef3",
+        "librdkafka/2.6.1#5b0ecc7c75e262b3e22d652018bedb93",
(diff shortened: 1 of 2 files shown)


divyaeluridd:
We encountered this exact issue independently and can confirm that the fix resolves it.

Environment:

Milvus v2.5.11, which pins librdkafka 1.9.1 on the 2.5 release line
External Kafka: Instaclustr-managed cluster using SASL_SSL / SCRAM-SHA-256
Deployment mode: Milvus cluster deployed through milvus-operator on Kubernetes

Symptom:

The behavior matched this report exactly: repeated SASL authentication error: Authentication failed during authentication due to invalid credentials with SASL mechanism SCRAM-SHA-256, occurring after approximately 306 ms in the AUTH_REQ state and continuously retried by librdkafka.

We independently verified that the credentials were valid.

Isolation:

We ruled out credentials, network connectivity, and CA configuration by testing against the same broker from the same network path, using the same credentials and CA, with two client-library versions:

librdkafka 1.8.0, via kafkacat → same authentication failure
librdkafka 2.15.0, via confluent-kafka Python → succeeds and returns broker metadata cleanly

czs007:
rerun go-sdk

Issue #52669, opened 2026-08-19
Bound concurrent segment loading under tiered storage eviction


sparknack:
### Problem

When Tiered Storage eviction is enabled, multiple sealed segments can enter synchronous warmup concurrently. The cache slots enforce max_loading_mem_ratio independently, so overlapping segment loads can repeatedly evict and reload data and amplify object-storage IO.

### Expected behavior

Add a QueryNode Segment Loader admission gate that budgets estimated synchronous warmup memory across segment loads. It should preserve load priority and cancellation, allow forward progress for an oversized segment, and apply only while eviction is enabled.

### Scope

This issue tracks the 2.6 branch change together with the milvus-common CacheSlot memory-ratio support from zilliztech/milvus-common#121.

PR #50478, opened 2026-06-11
enhance: stream field data loading

sparknack:
- Replace `common.entryStream.streamBudgetRatio` with refreshable `common.loadTransientBudgetBytes`, a process-wide transient load budget shared by scalar index V3 entry streaming and storage v2/v3 field-data loading. `0` keeps the limit disabled.
- Gate field-data batch reads on the shared budget, split batches by loading overhead, scale read parallelism from the effective batch budget, and finalize `GroupChunk` cells before pushing them so Arrow tables are released promptly.
- Align MCL loading-overhead reservations for storage v1 scalar indexes and storage v2/v3 field data under one load-transient overhead group, with array-field overhead and mmap file usage accounted separately.
- Propagate load cancellation into budget waits, field-data batch tasks, and V3 scalar `IndexEntryReader` stream/file reads so canceled loads do not stay blocked behind transient memory pressure.
- Update segcore init/config watchers and add C++/Go coverage for budget sizing, cancellation, field-data batching/finalization, V3 entry streaming, and the new config.

issue: #49499

PR #50267, opened 2026-06-03
enhance: add streaming load support for scalar index v3

sparknack:
issue: https://github.com/milvus-io/milvus/issues/48957

## What changed

- Add streaming V3 `LoadEntries` paths for scalar indexes to avoid reading full index entries into memory before materialization.
- Persist and load auxiliary metadata for sort/string indexes, including offset maps, valid bitsets, and CSR arrays, so mmap paths can map metadata directly and memory paths can preallocate buffers.
- Update scalar index resource estimates and related loading tests for the streaming/mmap load behavior.

## Why

- Reduce peak memory during sealed scalar index load, especially for large scalar sort, string, and bitmap indexes.
- Avoid recomputing persisted metadata at load time when V3 entries already contain it.

PR #52670, opened 2026-08-19
enhance: [2.6] refine segment loading resource control

sparknack:
issue: #52669
pr: #50267
pr: #50478

### Caching-layer and column loading

- Pin `milvus-common` to [zilliztech/milvus-common#121](https://github.com/zilliztech/milvus-common/pull/121).
- Adopt the batch-level `estimated_loading_usage(cids)` contract across caching-layer translators.
- Run each contiguous field-data batch's Arrow read and per-cell finalization on the same HIGH/LOW load-pool worker. Release Arrow tables before the batch future completes; no separate result queue is used.
- Bound column-group temporary memory by the selected cells and active load-pool workers: at most `min(selected size, concurrency × max(16 MiB, largest cell))`.
- Make thread-pool submission exception-safe when dynamic worker creation or metric updates fail.

### Index I/O and loading estimates

- Backport `PositionedFileWriter` and use it for `IndexEntryReader` positioned writes, including buffered/direct-I/O handling and writer priority. Validate encrypted V3 `slice_size` as a positive 4 KiB multiple.
- Add ordered `ReadEntryStream` loading and stream V3 payloads for scalar sort, string sort, Marisa, and bitmap indexes instead of materializing whole entries. Plain V3 stream memory is bounded by 16 MiB slices and load-pool concurrency; legacy and encrypted paths keep conservative whole-entry estimates.
- Refine scalar-index resident and peak estimates for sort auxiliary data, Marisa CSR data, Tantivy row/path offsets, bitmap buffers, and the persisted internal type of HYBRID indexes. Log the resulting index loading estimate at info level.
- Refine TextMatch estimates:
  - non-mmap keeps the index resident in memory and accounts for the temporary disk-to-`RamDirectory` copy with a `2 × index size` memory peak;
  - mmap keeps the index on disk, retains nullable row offsets in memory, and uses the larger of the pool-bounded V3 stream peak and the row-offset construction peak.

### DIFF
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -50,6 +50,7 @@
 #include "pb/schema.pb.h"
 #include "storage/IndexEntryReader.h"
 #include "storage/MemFileManagerImpl.h"
+#include "storage/ThreadPools.h"
 
 namespace milvus::index {
 namespace {
@@ -107,6 +108,15 @@ GetFileName(const std::string& path) {
     return pos == std::string::npos ? path : path.substr(pos + 1);
 }
 
+proto::common::LoadPriority
+ResolveLoadPriority(const std::map<std::string, std::string>& index_params) {
+    // Missing or unrecognized values use the higher-concurrency pool so the
+    // loading-memory estimate remains conservative.
+    const auto it = index_params.find(LOAD_PRIORITY);
+    return it == index_params.end() ? proto::common::LoadPriority::HIGH
+                                    : PriorityForLoad(it->second);
+}
+
 IndexType
 HybridInternalIndexTypeToIndexType(ScalarIndexType type) {
     switch (type) {
@@ -126,13 +136,13 @@ HybridInternalIndexTypeToIndexType(ScalarIndexType type) {
 std::optional<ScalarIndexType>
 ResolveHybridInternalIndexType(
     const std::vector<std::string>& index_files,
-    const storage::FileManagerContext& file_manager_context) {
+    const storage::FileManagerContext& file_manager_context,
+    proto::common::LoadPriority load_priority) {
     if (index_files.empty() || !file_manager_context.Valid()) {
         return std::nullopt;
     }
 
     storage::MemFileManagerImpl file_manager(file_manager_context);
-    const auto load_priority = proto::common::LoadPriority::HIGH;
     auto type_file =
         std::find_if(index_files.begin(), index_files.end(), [](const auto& f) {
             return GetFileName(f) == INDEX_TYPE;
@@ -156,7 +166,11 @@ ResolveHybridInternalIndexType(
         AssertInfo(input != nullptr,
                    "failed to open packed hybrid index file: {}",
                    index_files.front());
-        auto reader = storage::IndexEntryReader::Open(input, input->Size());
+        auto reader =
+            storage::IndexEntryReader::Open(input,
+                                            input->Size(),
+                                            /*collection_id=*/0,
+                                            PriorityForLoad(load_priority));
         AssertInfo(reader != nullptr,
                    "failed to create reader for packed hybrid index file");
         if (reader->HasMeta(INDEX_TYPE)) {
@@ -504,6 +518,7 @@ IndexFactory::ScalarIndexLoadResource(
     const auto scalar_version =
         GetValueFromConfig<int32_t>(config, SCALAR_INDEX_ENGINE_VERSION)
             .value_or(1);
+    const auto load_priority = ResolveLoadPriority(index_params);
 
     knowhere::expected<knowhere::Resource> resource;
 
@@ -513,7 +528,7 @@ IndexFactory::ScalarIndexLoadResource(
     if (index_type == milvus::index::ASCENDING_SORT) {
         if (scalar_version >= 3) {
             const auto stream_memory_overhead = ScalarIndexStreamMemoryOverhead(
-                index_size_in_bytes, scalar_version);
+                index_size_in_bytes, scalar_version, load_priority);
             const auto legacy_aux_bytes = SortLegacyAuxBytes(num_rows);
             if (mmap_enable) {
                 request.final_memory_cost = legacy_aux_bytes;
@@ -541,7 +556,7 @@ IndexFactory::ScalarIndexLoadResource(
                index_type == milvus::index::MARISA_TRIE_UPPER) {
         if (scalar_version >= 3) {
             const auto stream_memory_overhead = ScalarIndexStreamMemoryOverhead(
-                index_size_in_bytes, scalar_version);
+                index_size_in_bytes, scalar_version, load_priority);
             if (mmap_enable) {
                 const auto legacy_csr_resident_bytes =
                     MarisaLegacyCsrBytes(num_rows, 2);
@@ -578,9 +593,10 @@ IndexFactory::ScalarIndexLoadResource(
                index_type == milvus::index::NGRAM_INDEX_TYPE) {
         const auto validity_bitmap_bytes = ValidityBitmapBytes(num_rows);
         const auto stream_memory_overhead =
-            scalar_version >= 3 ? ScalarIndexStreamMemoryOverhead(
-                                      index_size_in_bytes, scalar_version)
-                                : index_size_in_bytes;
+            scalar_version >= 3
+                ? ScalarIndexStreamMemoryOverhead(
+                      index_size_in_bytes, scalar_version, load_priority)
+                : index_size_in_bytes;
         if (mmap_enable) {
             request.final_memory_cost = validity_bitmap_bytes;
             request.final_disk_cost = index_size_in_bytes;
@@ -605,8 +621,11 @@ IndexFactory::ScalarIndexLoadResource(
     } else if (index_type == milvus::index::BITMAP_INDEX_TYPE) {
         if (scalar_version >= 3) {
             const auto stream_memory_overhead = ScalarIndexStreamMemoryOverhead(
-                index_size_in_bytes, scalar_version);
+                index_size_in_bytes, scalar_version, load_priority);
             if (mmap_enable) {
+                // TODO: Once bitmap cardinality is exposed to admission
+                // estimation, account for the BITSET load mode's per-value
+                // resident bitsets here.
                 const auto resident_bytes = BitsetBytes(num_rows);
                 const auto frozen_buffer_bytes =
                     BitmapMmapFrozenBufferBytes(num_rows, index_size_in_bytes);
@@ -680,8 +699,9 @@ IndexFactory::ScalarIndexLoadResource(
     }
 
     try {
-        const auto internal_index_type =
-            ResolveHybridInternalIndexType(index_files, file_manager_context);
+        const auto load_priority = ResolveLoadPriority(index_params);
+        const auto internal_index_type = ResolveHybridInternalIndexType(
+            index_files, file_manager_context, load_priority);
         if (internal_index_type.has_value()) {
             const auto resolved_index_type =
                 HybridInternalIndexTypeToIndexType(internal_index_type.value());
--- a/internal/core/src/index/IndexFactory.cpp
+++ b/internal/core/src/index/IndexFactory.cpp
@@ -15,14 +15,19 @@
 // limitations under the License.
 
 #include "index/IndexFactory.h"
+#include <algorithm>
+#include <cstdint>
+#include <cstring>
 #include <cstdlib>
 #include <limits>
 #include <memory>
 #include <string>
+#include "common/Consts.h"
 #include "common/EasyAssert.h"
 #include "common/FieldDataInterface.h"
 #include "common/JsonCastType.h"
 #include "common/Types.h"
+#include "common/Utils.h"
 #include "index/Index.h"
 #include "index/JsonFlatIndex.h"
 #include "index/VectorMemIndex.h"
@@ -43,26 +48,132 @@
 #include "knowhere/comp/knowhere_check.h"
 #include "log/Log.h"
 #include "pb/schema.pb.h"
+#include "storage/IndexEntryReader.h"
+#include "storage/MemFileManagerImpl.h"
 
 namespace milvus::index {
+namespace {
+
+uint64_t
+BitsetBytes(int64_t num_rows) {
+    if (num_rows <= 0) {
+        return 0;
+    }
+    return (static_cast<uint64_t>(num_rows) + 7) / 8;
+}
+
+uint64_t
+AlignUp(uint64_t size, uint64_t alignment) {
+    if (alignment == 0 || size == 0) {
+        return size;
+    }
+    if (size > std::numeric_limits<uint64_t>::max() - (alignment - 1)) {
+        return std::numeric_limits<uint64_t>::max();
+    }
+    return ((size + alignment - 1) / alignment) * alignment;
+}
+
+uint64_t
+BitmapMmapFrozenBufferBytes(int64_t num_rows, uint64_t index_size_in_bytes) {
+    constexpr uint64_t kBitmapFrozenAlignment = 32;
+    return std::max(AlignUp(BitsetBytes(num_rows), kBitmapFrozenAlignment),
+                    index_size_in_bytes);
+}
+
+uint64_t
+SortLegacyAuxBytes(int64_t num_rows) {
+    if (num_rows <= 0) {
+        return 0;
+    }
+    const auto rows = static_cast<uint64_t>(num_rows);
+    return SaturatingAdd(SaturatingMultiply(rows, sizeof(int32_t)),
+                         BitsetBytes(num_rows));
+}
+
+uint64_t
+MarisaLegacyCsrBytes(int64_t num_rows, uint64_t arrays_per_row) {
+    if (num_rows <= 0) {
+        return 0;
+    }
+    const auto rows = static_cast<uint64_t>(num_rows);
+    return SaturatingMultiply(
+        SaturatingAdd(SaturatingMultiply(arrays_per_row, rows), uint64_t{1}),
+        sizeof(uint32_t));
+}
+
+std::string
+GetFileName(const std::string& path) {
+    const auto pos = path.find_last_of('/');
+    return pos == std::string::npos ? path : path.substr(pos + 1);
+}
+
+IndexType
+HybridInternalIndexTypeToIndexType(ScalarIndexType type) {
+    switch (type) {
+        case ScalarIndexType::BITMAP:
+            return BITMAP_INDEX_TYPE;
+        case ScalarIndexType::STLSORT:
+            return ASCENDING_SORT;
+        case ScalarIndexType::MARISA:
+            return MARISA_TRIE;
+        case ScalarIndexType::INVERTED:
+            return INVERTED_INDEX_TYPE;
+        default:
+            return "";
+    }
+}
+
+std::optional<ScalarIndexType>
+ResolveHybridInternalIndexType(
+    const std::vector<std::string>& index_files,
+    const storage::FileManagerContext& file_manager_context) {
+    if (index_files.empty() || !file_manager_context.Valid()) {
+        return std::nullopt;
+    }
+
+    storage::MemFileManagerImpl file_manager(file_manager_context);
+    const auto load_priority = proto::common::LoadPriority::HIGH;
+    auto type_file =
+        std::find_if(index_files.begin(), index_files.end(), [](const auto& f) {
+            return GetFileName(f) == INDEX_TYPE;
+        });
+    if (type_file != index_files.end()) {
+        auto index_datas = file_manager.LoadIndexToMemory(
+            std::vector<std::string>{*type_file}, load_priority);
+        BinarySet binary_set;
+        AssembleIndexDatas(index_datas, binary_set);
+        auto type_buffer = binary_set.GetByName(INDEX_TYPE);
+        AssertInfo(
+            type_buffer != nullptr && type_buffer->size >= sizeof(uint8_t),
+            "index type file not found in hybrid index binary set");
+        uint8_t type = 0;
+        std::memcpy(&type, type_buffer->data.get(), sizeof(type));
+        return static_cast<ScalarIndexType>(type);
+    }
+
+    if (index_files.size() == 1 && file_manager_context.fs != nullptr) {
+        auto input = file_manager.OpenInputStream(index_files.front());
+        AssertInfo(input != nullptr,
+                   "failed to open packed hybrid index file: {}",
+                   index_files.front());
+        auto reader = storage::IndexEntryReader::Open(input, input->Size());
+        AssertInfo(reader != nullptr,
+                   "failed to create reader for packed hybrid index file");
+        if (reader->HasMeta(INDEX_TYPE)) {
+            return static_cast<ScalarIndexType>(
+                reader->GetMeta<uint8_t>(INDEX_TYPE));
+        }
+    }
+    return std::nullopt;
+}
+
+}  // namespace
 
 namespace {
 
 uint64_t
 ValidityBitmapBytes(int64_t num_rows) {
-    if (num_rows <= 0) {
-        return 0;
-    }
-    auto rows = static_cast<uint64_t>(num_rows);
-    return ((rows - 1) / 64 + 1) * sizeof(uint64_t);
-}
-
-uint64_t
-SaturatingAdd(uint64_t lhs, uint64_t rhs) {
-    if (lhs > std::numeric_limits<uint64_t>::max() - rhs) {
-        return std::numeric_limits<uint64_t>::max();
-    }
-    return lhs + rhs;
+    return AlignUp(BitsetBytes(num_rows), sizeof(uint64_t));
 }
 
 }  // namespace
@@ -159,6 +270,38 @@ IndexFactory::IndexLoadResource(
                                        mmap_enable,
                                        num_rows);
     }
+}
+
+LoadResourceRequest
+IndexFactory::IndexLoadResource(
+    DataType field_type,
+    DataType element_type,
+    IndexVersion index_version,
+    uint64_t index_size_in_bytes,
+    const std::map<std::string, std::string>& index_params,
+    bool mmap_enable,
+    int64_t num_rows,
+    int64_t dim,
+    const std::vector<std::string>& index_files,
+    const storage::FileManagerContext& file_manager_context) {
+    if (milvus::IsVectorDataType(field_type)) {
+        return VecIndexLoadResource(field_type,
+                                    element_type,
+                                    index_version,
+                                    index_size_in_bytes,
+                                    index_params,
+                                    mmap_enable,
+                                    num_rows,
+                                    dim);
+    }
+    return ScalarIndexLoadResource(field_type,
+                                   index_version,
+                                   index_size_in_bytes,
+                                   index_params,
+                                   mmap_enable,
+                                   num_rows,
+                                   index_files,
+                                   file_manager_context);
 }
 
 LoadResourceRequest
@@ -358,6 +501,9 @@ IndexFactory::ScalarIndexLoadResource(
     auto index_type_it = index_params.find("index_type");
     AssertInfo(index_type_it != index_params.end(), "index type is empty");
     const std::string& index_type = index_type_it->second;
+    const auto scalar_version =
+        GetValueFromConfig<int32_t>(config, SCALAR_INDEX_ENGINE_VERSION)
+            .value_or(1);
 
     knowhere::expected<knowhere::Resource> resource;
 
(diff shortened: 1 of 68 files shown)


internal/core/thirdparty/milvus-common/CMakeLists.txt
```diff
@@ -13,8 +13,8 @@
 
 milvus_add_pkg_config("milvus-common")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( MILVUS-COMMON-VERSION 9dc0923 )
-set( GIT_REPOSITORY "https://github.com/zilliztech/milvus-common.git")
+set( MILVUS-COMMON-VERSION ec692465b1a5dbf59db0b5eab39d1f9acfbb4b1a )
+set( GIT_REPOSITORY "https://github.com/sparknack/milvus-common.git")
```
buqian-zilliz: **[P1] Build depends on a personal fork of milvus-common**

`GIT_REPOSITORY` is changed from `https://github.com/zilliztech/milvus-common.git` to `https://github.com/sparknack/milvus-common.git` with a commit hash from that fork. A release dependency must not build against an unreviewed personal fork. Please land the `TieredStorageOptions` / `max_loading_mem_ratio` change in `zilliztech/milvus-common` first and point this back at the upstream repository, or obtain explicit maintainer sign-off for the fork before merge.

internal/querynodev2/segments/segment_load_admission.go
```diff
@@ -0,0 +1,168 @@
...
+	return &segmentLoadAdmission{capacity: capacity}
+}
+
+func newSegmentLoadAdmissionFromConfig() *segmentLoadAdmission {
+	params := &paramtable.Get().QueryNodeCfg
+	if !params.TieredEvictionEnabled.GetAsBool() {
+		return nil
+	}
+
+	capacity := uint64(float64(hardware.GetMemoryCount()) * params.TieredMaxLoadingMemoryRatio.GetAsFloat())
```
buqian-zilliz: **[P2] Admission capacity uses total memory instead of effective memory and defaults to 1.0**

`newSegmentLoadAdmissionFromConfig` computes `capacity := hardware.GetMemoryCount() * TieredMaxLoadingMemoryRatio`, and the default ratio is `1.0`. `GetMemoryCount` returns the host/cgroup total memory, not the `effective` (available) memory described by the parameter documentation. With the default this admits concurrent sync warmup up to the entire node memory without subtracting memory already in use by the query node, so it may not actually bound peak memory and can still OOM. Consider basing the capacity on available memory (e.g. `GetMemoryCount() - GetUsedMemoryCount()`) or defaulting to a conservative fraction, and clarify the parameter wording.

internal/querynodev2/segments/segment_load_admission.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+	defer release()
+	return load()
+}
+
+func (a *segmentLoadAdmission) enqueueLocked(waiter *segmentLoadAdmissionWaiter) {
+	insertAt := len(a.waiters)
+	for i, queued := range a.waiters {
+		if waiter.priority < queued.priority {
```
buqian-zilliz: **[P3] Low-priority waiters can starve under sustained high-priority loads**

`enqueueLocked` inserts a new waiter before the first queued waiter with a strictly higher (lower priority) value, so every new `HIGH` load is placed ahead of already-waiting `LOW` loads. There is no aging or timeout in the queue, so background loads (`LoadPriority_LOW`, used for balance/index updates) can be delayed indefinitely while `HIGH` loads keep arriving. Consider FIFO ordering within priority levels, aging, or a bounded wait before granting.

internal/querynodev2/segments/segment_loader.go
```diff
@@ -2053,6 +2126,7 @@ func estimateLoadingResourceUsageOfSegment(schema *schemapb.CollectionSchema, lo
 
 			supportInterimIndexDataType = supportInterimIndexDataType || SupportInterimIndexDataType(fieldSchema.GetDataType())
 			isVectorType = isVectorType && typeutil.IsVectorType(fieldSchema.GetDataType())
+			needWarmup = needWarmup || getFieldWarmupPolicy(fieldSchema) == common.WarmupSync
```
buqian-zilliz: **[P3] Admission over-counts a grouped field binlog when only some child fields need warmup**

For storage-v2 grouped `FieldBinlog`s with `ChildFields`, `needWarmup` is the OR of `getFieldWarmupPolicy` across children, but the whole `binlogSize` is added to `admissionMemorySize` when any child is `WarmupSync`. When a group mixes warmup-sync and warmup-disable fields, the disabled children's bytes are also counted, making admission more conservative than necessary (fewer concurrent loads). Confirm whether mixed-policy groups can occur; if so, consider summing only the warmup children's sizes.

internal/core/thirdparty/knowhere/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 # Update KNOWHERE_VERSION for the first occurrence
 milvus_add_pkg_config("knowhere")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( KNOWHERE_VERSION v2.6.18 )
-set( GIT_REPOSITORY  "https://github.com/zilliztech/knowhere.git")
+set( KNOWHERE_VERSION cp2.6-milvus-common-pr121 )
+set( GIT_REPOSITORY  "https://github.com/sparknack/knowhere.git")
```
buqian-zilliz: **[P1] Build also depends on a personal fork of knowhere**

`GIT_REPOSITORY` is changed from `https://github.com/zilliztech/knowhere.git` to `https://github.com/sparknack/knowhere.git` with a non-release tag `cp2.6-milvus-common-pr121`. This mirrors the existing milvus-common fork issue: a release dependency must not build against an unreviewed personal fork. Please land the required knowhere change upstream in `zilliztech/knowhere` and point this back at the upstream repository, or obtain explicit maintainer sign-off for the fork before merge.

internal/core/thirdparty/milvus-common/CMakeLists.txt
buqian-zilliz: This issue still exists. `internal/core/thirdparty/milvus-common/CMakeLists.txt` still sets `GIT_REPOSITORY` to `https://github.com/sparknack/milvus-common.git` (commit `dd5bc110bfe63c248c597dde40aea1ff2234044a`), so the release still builds against the personal fork.

internal/querynodev2/segments/segment_load_admission.go
buqian-zilliz: This issue still exists. `newSegmentLoadAdmissionFromConfig` still computes capacity from `hardware.GetMemoryCount()` (total/cgroup memory, not effective/available memory) with a default ratio of `1.0`, so concurrent sync warmup can be admitted up to the full node memory without subtracting memory already in use.

internal/querynodev2/segments/segment_load_admission.go
buqian-zilliz: This issue still exists. `enqueueLocked` still inserts each new `HIGH` waiter ahead of queued `LOW` waiters with no aging or timeout, so background `LOW` loads can be delayed indefinitely under sustained `HIGH` load.

internal/querynodev2/segments/segment_loader.go
buqian-zilliz: This issue still exists. For storage-v2 grouped `FieldBinlog`s with `ChildFields`, `needWarmup` is still the OR of `getFieldWarmupPolicy` across children, and the whole `binlogSize` is still added to `admissionMemorySize` when any child is `WarmupSync`, so mixed-policy groups are over-counted.

internal/querynodev2/segments/segment_loader.go
```diff
@@ -1685,14 +1691,45 @@ func (loader *segmentLoader) checkSegmentSize(ctx context.Context, segmentLoadIn
 		zap.Int("mmapFieldCount", mmapFieldCount),
 	)
 
+	if predictMemUsage > uint64(float64(totalMem)*paramtable.Get().QueryNodeCfg.OverloadedMemoryThresholdPercentage.GetAsFloat()) {
```
buqian-zilliz: **[P1] Allow tiered eviction before rejecting the load**

With tiered eviction enabled, this host-memory gate runs before any caching-layer reservation. Physical usage includes evictable cache cells, while `TryReserveLoadingResourceWithTimeout` is the path that evicts them to make room. A node near the threshold can therefore return `SegmentRequestResourceFailed` even though enough unpinned cache could be evicted, because the request never reaches the reservation. Keep the host threshold gate on the non-tiered path, or perform and reflect cache eviction before rejecting.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -328,58 +528,161 @@ IndexFactory::ScalarIndexLoadResource(
...
     } else if (index_type == milvus::index::BITMAP_INDEX_TYPE) {
-        if (mmap_enable) {
-            request.final_memory_cost = 0;
-            request.final_disk_cost = index_size_in_bytes;
-            request.max_memory_cost = index_size_in_bytes;
-            request.max_disk_cost = index_size_in_bytes;
+        if (scalar_version >= 3) {
+            const auto stream_memory_overhead = ScalarIndexStreamMemoryOverhead(
+                index_size_in_bytes, scalar_version);
+            if (mmap_enable) {
```
buqian-zilliz, line 625: **[P1] Account for the non-mmap bitmap load mode**

When mmap is requested, low-cardinality V3 bitmap indexes do not take the mmap branch: indexes with at most 500 keys expand every posting into a full `TargetBitmap`. This estimate nevertheless reserves only one `BitsetBytes(num_rows)` in memory and classifies the index as disk-resident. For K keys, actual resident bitmap storage is roughly `K * ceil(num_rows / 8)`, so the reservation can be hundreds of times too small and admit a load that exhausts memory. Pass the persisted cardinality or selected load mode into the estimate, or conservatively price the `BITSET` branch.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -328,58 +528,161 @@ IndexFactory::ScalarIndexLoadResource(
...
+                const auto resident_bytes = BitsetBytes(num_rows);
+                const auto frozen_buffer_bytes =
+                    BitmapMmapFrozenBufferBytes(num_rows, index_size_in_bytes);
+                request.final_memory_cost = resident_bytes;
+                request.final_disk_cost = index_size_in_bytes;
+                request.max_memory_cost = SaturatingAdd(
+                    SaturatingAdd(resident_bytes, stream_memory_overhead),
+                    frozen_buffer_bytes);
+                request.max_disk_cost =
```
buqian-zilliz, line 637: **[P2] Avoid multiplying the bitmap disk peak twice**

For a V3 mmap bitmap, this value is already doubled to cover the streamed temporary file plus the final frozen mmap file. `SealedIndexTranslator::estimated_loading_usage()` then multiplies every `max_disk_cost` by two again, even though `LoadResourceRequest.max_disk_cost` already represents the loading peak. The cache consequently reserves roughly four times the index size and can reject a load that needs about twice the index size. Define the temporary-file factor at one boundary so both admission layers consume the same peak.

internal/core/src/segcore/storagev1translator/DefaultValueChunkTranslator.cpp
```diff
@@ -189,20 +189,25 @@ DefaultValueChunkTranslator::value_size() const {
...
-    auto rows_begin = meta_.num_rows_until_chunk_[cid];
-    auto rows_end = meta_.num_rows_until_chunk_[cid + 1];
-    auto rows = rows_end - rows_begin;
-    auto cell_bytes = value_size * rows;
-    if (use_mmap_) {
-        return {{0, cell_bytes}, {0, cell_bytes}};
-    } else {
-        return {{cell_bytes, 0}, {cell_bytes, 0}};
+    milvus::cachinglayer::ResourceUsage usage;
+    for (const auto cid : cids) {
```
buqian-zilliz, line 217: **[P2] Charge shared default buffers once per batch**

`get_cells()` returns requested cells as views over one prebuilt `primary_buffer_` and at most one `tail_buffer_`, but this estimator adds `value_size * rows` for every `cid`. Large dynamically added fields are split into roughly 64 KiB cells, so warming all cells can reserve and later charge close to the full logical column even though only one or two buffers exist. This can reject or serialize loads by orders of magnitude. Account shared buffer ownership once per batch and keep the subsequent per-cell charging consistent.

internal/core/src/segcore/storagev2translator/GroupCTMeta.h
```diff
@@ -26,12 +34,50 @@ constexpr size_t kRowGroupsPerCell = 4;
...
+        assert(cid < cell_sizes.size());
+        loaded_size = SaturatingAdd(loaded_size, cell_sizes[cid]);
+    }
+
+    // All requested cells remain loaded. A load-pool worker owns at most one
+    // batch at a time, including both Arrow reads and finalization. A batch is
+    // bounded by the target size unless it contains one oversized cell.
+    const auto concurrent_batches =
+        std::min(cids.size(), std::max<size_t>(worker_count, 1));
+    const auto max_batch_size =
```
buqian-zilliz, line 61: **[P2] Include the configured reader floor in the group-chunk peak**

The executor gives every batch reader at least `FILE_SLICE_SIZE.load()` bytes, but this estimator caps an active batch at `max(batch_target_bytes, max_cell_size)` and never includes that configurable floor. Deployments with `common.indexSliceSize` above 16 MiB can therefore run W readers with a `W * slice_size` budget while reserving only `W * 16 MiB`, defeating the loading-memory cap. Derive execution and estimation from the same effective per-batch reader limit.

internal/core/src/segcore/memory_planner.cpp
```diff
@@ -329,69 +328,88 @@ LoadCellBatchAsync(milvus::OpContext* op_ctx,
...
     }
 
     return futures;
 }
 
+size_t
+GetCellBatchConcurrency(milvus::proto::common::LoadPriority load_priority) {
+    auto& pool =
+        ThreadPools::GetThreadPool(milvus::PriorityForLoad(load_priority));
+    return std::max<size_t>(1, pool.GetMaxThreadNum());
```
buqian-zilliz, line 409: **[P2] Estimate using workers that remain after a pool shrink**

`ThreadPool::Resize` lowers the maximum thread count but does not synchronously retire existing workers; running workers can continue processing queued batches before they become idle and exit. Immediately after a 16-to-4 resize, this helper reports 4 while as many as 16 batches can still be active, so the transient peak can be under-reserved by four times. Bound concurrency by at least `max(GetThreadNum(), GetMaxThreadNum())`, or retire excess workers before admitting new batch loads.

internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp
```diff
@@ -71,6 +73,23 @@ SealedIndexTranslator::SealedIndexTranslator(
           !(IsVectorDataType(load_index_info->field_type) &&
             knowhere::IndexFactory::Instance().FeatureCheck(
                 index_info_.index_type, knowhere::feature::LAZY_LOAD))) {
+    load_resource_request_ = EstimateLoadResource();
```
buqian-zilliz: **[P2] Do not eagerly cache the complete index load estimate**

Computing and storing the complete estimate in the translator constructor makes a HYBRID scalar index synchronously fetch or open remote metadata before the cache warmup policy is applied, so even disabled warmup performs object-store I/O and the eventual load reads the metadata again. It also freezes the pool-concurrency-dependent streaming peak even though the pool can be resized before an evicted cell reloads. Cache only stable metadata, preferably carry the internal index type in load metadata, and recompute dynamic peak costs for each reservation.

internal/core/src/segcore/storagev1translator/TextMatchIndexTranslator.cpp
```diff
@@ -58,18 +61,40 @@ TextMatchIndexTranslator::cell_id_of(milvus::cachinglayer::uid_t) const {
...
+        const auto load_priority =
+            milvus::index::GetValueFromConfig<
+                milvus::proto::common::LoadPriority>(config_, LOAD_PRIORITY)
+                .value_or(milvus::proto::common::LoadPriority::HIGH);
+        const auto stream_memory_peak =
+            static_cast<int64_t>(milvus::index::ScalarIndexStreamMemoryOverhead(
+                static_cast<uint64_t>(index_size),
+                load_info_.scalar_index_version,
+                load_priority));
+        return {{row_offsets_size, index_size},
```
buqian-zilliz: **[P2] Keep nullable mmap offsets charged as memory after load**

This estimate correctly treats nullable row offsets as resident memory for an mmap text index, but after loading, the cell still reports the whole `index->ByteSize()` as disk. The cache replaces the estimate with that actual cell size, so the `null_offset_` vector, which can consume `num_rows * sizeof(size_t)`, disappears from memory accounting and is misclassified as disk. Across many nullable text indexes this can overcommit RAM; report the row-offset allocation separately from the Tantivy mmap-file bytes.

internal/core/thirdparty/knowhere/CMakeLists.txt
buqian-zilliz: This issue still exists. The reviewed head still points the Knowhere dependency at `sparknack/knowhere` with tag `cp2.6-milvus-common-pr121`; no upstream repository pin or maintainer resolution is reflected in the latest diff.

internal/querynodev2/segments/segment_loader.go
buqian-zilliz: This issue still exists. The latest implementation still combines child-field warmup decisions with OR and then charges the entire grouped binlog whenever any child requires synchronous warmup.

internal/core/thirdparty/milvus-common/CMakeLists.txt
buqian-zilliz: This issue still exists. The reviewed head still pins `milvus-common` to commit `dd5bc3f2b817de41795d34524953344eb1bc623d` from the personal `sparknack/milvus-common` fork.

internal/querynodev2/segments/segment_loader.go
sparknack: Fixed in 529e581f6b. Removed the Go segment-level admission split. With tiered eviction enabled, the host memory/disk threshold gate is no longer applied before `TryReserveLoadingResourceWithTimeout`, so the caching layer can evict unpinned cells before deciding whether to reject the load. Segment load concurrency is controlled through QueryCoord. Verified with the `internal/querynodev2/segments/...` test suite.

internal/core/thirdparty/milvus-common/CMakeLists.txt
sparknack: Confirmed. This is an intentional test-only pin for zilliztech/milvus-common#121 and is not merge-ready. Before merge, the dependency will be switched back to `zilliztech/milvus-common` at the landed upstream commit.

internal/querynodev2/segments/segment_load_admission.go
sparknack: Resolved by removing the Go `segmentLoadAdmission` queue in 529e581f6b. `TieredMaxLoadingMemoryRatio` now configures only the caching-layer loading budget; it is no longer multiplied by `hardware.GetMemoryCount()` to admit segments. Segment concurrency is controlled through QueryCoord.

internal/querynodev2/segments/segment_load_admission.go
sparknack: Resolved by 529e581f6b: `segment_load_admission.go` and its priority waiter queue were removed. There is no HIGH/LOW ordering in Go segment admission now; segment concurrency is controlled through QueryCoord.

internal/querynodev2/segments/segment_loader.go
sparknack: Resolved by 529e581f6b removing the Go segment admission estimate. Grouped `FieldBinlog` data is no longer charged wholesale to an `admissionMemorySize`; tiered cache-managed cell loading is accounted by the caching layer, while SegmentLoader reserves only resources it owns.

internal/core/thirdparty/knowhere/CMakeLists.txt
sparknack: Confirmed. This is the matching test-only Knowhere pin needed to consume milvus-common#121 on the 2.6 dependency chain. The PR is not merge-ready with `sparknack/knowhere`; it will be switched to an upstream `zilliztech/knowhere` commit or tag after the dependency lands.

internal/core/src/index/IndexFactory.cpp
sparknack: Confirmed; this remains in the current head. V3 bitmap still selects BITSET at load time for cardinality <= 500, while the estimate has no cardinality and prices the mmap branch. A complete fix needs the persisted load mode/cardinality carried into `LoadIndexInfo` (or a conservative BITSET estimate); constructor-time metadata I/O was intentionally not reintroduced.

internal/core/src/index/IndexFactory.cpp
sparknack: Fixed in 69d85ed3fc. `IndexFactory` is now the single owner of the bitmap temporary-file factor, and `SealedIndexTranslator::estimated_loading_usage()` consumes `max_disk_cost` directly without multiplying it again. Covered by `SealedIndexTranslatorTest.UsesFactoryPeakDiskWithoutDoublingIt`.

internal/core/src/segcore/storagev1translator/DefaultValueChunkTranslator.cpp
sparknack: Fixed in 69d85ed3fc. The estimate charges `primary_buffer_` and `tail_buffer_` once, while `get_cells()` assigns each shared backing buffer to one non-evictable owner cell and zero-charges the other views. Dedicated mmap and non-mmap tests compare unique backing bytes with the sum of cell charges.

internal/core/src/segcore/storagev2translator/GroupCTMeta.h
sparknack: Fixed in 69d85ed3fc. Group and manifest loading no longer use the `FILE_SLICE_SIZE` floor: `LoadCellBatchAsync` clamps each reader to `max(1, min(batch_memory, memory_limit))`, and both execution and estimation use `kDefaultFieldDataLoadBatchTargetBytes`. The remaining `FILE_SLICE_SIZE` use is in the legacy `LoadWithStrategy` path, not group-chunk loading.

internal/core/src/segcore/memory_planner.cpp
sparknack: Confirmed for a runtime pool shrink. This backport does not handle that case: HIGH/LOW load-pool sizes are treated as startup-stable in supported 2.6 deployments and are not dynamically resized, so the estimate intentionally uses `GetMaxThreadNum()`. A transient under-reservation remains a known limitation if runtime resizing is introduced.

internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp
sparknack: Partially addressed in 5c3bf9e162. The constructor no longer computes the load request, so HYBRID metadata is resolved only on the first non-empty loading attempt and then reused for reservation and cell accounting. The complete request remains cached intentionally because load-pool sizing is treated as startup-stable in this 2.6 backport; runtime resize is not supported by this estimate.

internal/core/src/segcore/storagev1translator/TextMatchIndexTranslator.cpp
sparknack: Confirmed; this is not fixed in the current head. The estimate keeps nullable row offsets in memory, but mmap `get_cells()` still uses `SetCellSize({0, index->ByteSize()})`. A complete fix needs `TextMatchIndex` to expose Tantivy mmap bytes and resident `null_offset_` bytes separately, then charge `{resident_offsets, mmap_files}`.

pkg/util/paramtable/component_param.go
```diff
@@ -3969,6 +3970,22 @@ If set to 0, time based eviction is disabled.`,
 	}
 	p.TieredLoadingResourceFactor.Init(base.mgr)
 
+	p.TieredMaxLoadingMemoryRatio = ParamItem{
+		Key:          "queryNode.segcore.tieredStorage.maxLoadingMemoryRatio",
+		Version:      "2.6.23",
+		DefaultValue: "1.0",
+		Formatter: func(v string) string {
+			ratio := getAsFloat(v)
+			if ratio <= 0 || ratio > 1 {
```
buqian-zilliz, line 3970: **[P1] Reject non-finite loading-memory ratios before CGo**

`strconv.ParseFloat` accepts `NaN`, but both comparisons against it are false, so this formatter preserves the value. Because the parameter is refreshable, the callback passes it through CGo to `Manager::UpdateConfig`; `getMaxLoadingMemSize` rejects it with `AssertInfo`, and the `extern "C"` wrapper has no exception barrier, so the C++ exception can terminate QueryNode. Explicitly reject `math.IsNaN` and `math.IsInf`. Fixed `%f` formatting should also be avoided because it turns a sufficiently small valid positive ratio into `0.000000`.

internal/core/src/index/IndexFactory.cpp
```diff
@@ -358,45 +501,99 @@ IndexFactory::ScalarIndexLoadResource(
...
+                    MarisaLegacyCsrBytes(num_rows, 2);
+                const auto legacy_csr_peak_bytes =
+                    MarisaLegacyCsrBytes(num_rows, 3);
+                request.final_memory_cost = legacy_csr_resident_bytes;
+                request.final_disk_cost = index_size_in_bytes;
+                request.max_memory_cost = SaturatingAdd(legacy_csr_peak_bytes,
+                                                        stream_memory_overhead);
+                request.max_disk_cost = index_size_in_bytes;
+            } else {
+                request.final_memory_cost = index_size_in_bytes;
```
buqian-zilliz, line 571: **[P1] Account for CSR rebuilt by legacy non-mmap Marisa loads**

`StringIndexMarisa::LoadEntries` falls back to `fill_offsets()` when an older V3 file lacks the new CSR side entries. In non-mmap mode those rebuilt arrays were not included in `index_size_in_bytes`: they retain roughly `8 * num_rows` bytes and temporarily add another `4 * num_rows` bytes when cardinality approaches the row count. This branch reserves only the packed index size, so upgrading and loading a production-size legacy index can exceed the cache reservation by hundreds of MiB. Include the legacy CSR resident and peak allowances here, or pass the persisted format metadata into the estimator.

internal/core/src/index/Utils.cpp
```diff
@@ -38,11 +39,38 @@
...
+    }
+    if (scalar_version < 3) {
+        return index_size_in_bytes;
+    }
+
+    // Without persisted slice metadata an encrypted slice has no trustworthy
+    // ciphertext bound. Keep the existing whole-index estimate whenever a
+    // cipher plugin is active.
+    if (storage::PluginLoader::GetInstance().getCipherPlugin() != nullptr) {
+        return index_size_in_bytes;
```
buqian-zilliz, line 64: **[P2] Account for simultaneous encrypted stream buffers**

Returning only `index_size_in_bytes` is not a conservative encrypted-stream bound. `ReadEncryptedEntryStream` budgets each active slice as ciphertext plus two plaintext buffers, and those allocations coexist during decryption and conversion. A one-slice encrypted V3 index can therefore consume roughly two to three times its index size transiently, while larger indexes multiply the working set by active workers. The new admission estimate can consequently under-reserve and still overcommit memory. Derive the bound from the encrypted slice working set, including decryption buffers.

internal/core/src/segcore/storagev1translator/TextMatchIndexTranslator.cpp
```diff
@@ -73,19 +75,39 @@ TextMatchIndexTranslator::cell_id_of(milvus::cachinglayer::uid_t) const {
...
     // ignore the cid checking, because there is only one cell
-    auto bitmap_bytes = EstimateValidityBitmapBytes(load_info_.num_rows);
+    const auto index_size = std::max<int64_t>(load_info_.index_size, 0);
+    const auto bitmap_bytes = EstimateValidityBitmapBytes(load_info_.num_rows);
+    const auto load_priority =
+        milvus::index::GetValueFromConfig<milvus::proto::common::LoadPriority>(
+            config_, LOAD_PRIORITY)
+            .value_or(milvus::proto::common::LoadPriority::HIGH);
+    const auto stream_memory_peak =
+        static_cast<int64_t>(milvus::index::ScalarIndexStreamMemoryOverhead(
```
buqian-zilliz, line 91: **[P2] Use the reader's actual priority for text-index estimates**

This estimates a LOW-priority V3 text load using the LOW pool width, but `ScalarIndex<T>::LoadUnified` opens `IndexEntryReader` without forwarding the configured priority, so entry streams actually run on the default HIGH pool. When the HIGH pool is larger, the load can have substantially more concurrent slice buffers than reserved. Pass `LOAD_PRIORITY` into `IndexEntryReader::Open`, or estimate using the pool the reader actually uses.

internal/core/src/segcore/storagev2translator/GroupCTMeta.h
```diff
@@ -26,12 +34,50 @@ constexpr size_t kRowGroupsPerCell = 4;
...
+        std::max({batch_target_bytes, max_cell_size, int64_t{1}});
+    const auto temporary_size =
+        std::min(loaded_size,
+                 SaturatingMultiply(max_batch_size,
+                                    static_cast<int64_t>(concurrent_batches)));
+
+    if (use_mmap) {
+        return {{0, loaded_size}, {temporary_size, loaded_size}};
```
buqian-zilliz, line 69: **[P2] Include concurrent mmap writer buffers in the group-load peak**

The PR now runs `load_group_chunk` inside every active load-pool worker, and each mmap finalizer constructs a `FileWriter` that immediately allocates `FileWriter::GetBufferSize()`. `temporary_size` accounts only for Arrow batch bytes. Since the writer buffer is configurable up to 64 MiB, sixteen concurrent finalizers can add 1 GiB beyond the reservation. Add one configured writer buffer per concurrent mmap batch to the peak, or bound mmap finalization separately.

internal/core/src/index/StringIndexSort.cpp
```diff
@@ -544,28 +581,168 @@ StringIndexSort::LoadEntries(storage::IndexEntryReader& reader,
...
+                           "string sort stream exceeds expected size");
+                std::memcpy(data.data() + written, slice, len);
+                written += len;
+            });
+        AssertInfo(written == data_size,
+                   "string sort stream size mismatch: got {}, expected {}",
+                   written,
+                   data_size);
+
+        auto buffer_impl = std::make_unique<StringIndexSortMmapImpl>();
```
buqian-zilliz, line 742: **[P2] Preserve serialization for non-mmap loaded string indexes**

The non-mmap V3 branch now installs a buffer-backed `StringIndexSortMmapImpl`, but `Serialize` still unconditionally casts `impl_` to `StringIndexSortMemoryImpl`, and `WriteEntries` explicitly requires that type. A load-then-serialize or load-then-upload flow that previously used the memory implementation now invokes undefined behavior or fails. Retain a serializable memory implementation here or teach the buffer-backed implementation to emit the existing format.

internal/core/src/segcore/storagev1translator/DefaultValueChunkTranslator.cpp
```diff
@@ -293,9 +322,29 @@ DefaultValueChunkTranslator::get_cells(
...
         res.emplace_back(cid, std::move(chunk));
     }
 
+    if (res.empty()) {
+        return res;
+    }
+
+    milvus::cachinglayer::ResourceUsage owned_usage;
+    if (primary_buffer_.has_value() &&
+        !primary_buffer_accounted_.exchange(true, std::memory_order_acq_rel)) {
```
buqian-zilliz, line 337: **[P2] Do not bind shared-buffer accounting permanently to one cell**

These flags permanently attach both shared-buffer charges to the first returned cell even though the buffers remain owned by the translator. `support_eviction=false` disables strategy eviction, but `CacheSlot::ManualEvict` can still unload that owner and refund its charge while the buffers remain alive; subsequent estimates and cells report zero because the flags never reset. Keep the charge in slot-owned state or transfer/reset ownership when the charged cell unloads.

internal/core/src/index/IndexFactory.cpp
buqian-zilliz: This issue still exists: the estimator has no persisted cardinality or selected bitmap load mode, so a requested mmap load can still be priced as disk-resident even when `ChooseIndexLoadMode` selects the in-memory BITSET representation.

internal/core/src/segcore/storagev1translator/DefaultValueChunkTranslator.cpp
sparknack: Default-value cells use `support_eviction=false`, so normal cache eviction cannot remove the charged cell. We also found no production caller of `ManualEvict` for this path. We will not add shared-buffer ownership machinery for an unused manual-eviction case.

internal/core/src/segcore/storagev1translator/TextMatchIndexTranslator.cpp
sparknack: Will fix this by resolving `LOAD_PRIORITY` in `ScalarIndex::LoadUnified()` and passing the corresponding pool priority to `IndexEntryReader::Open()`, so loading and estimation use the same pool width. A regression test will cover LOW-priority loading while the HIGH pool is blocked.

pkg/util/paramtable/component_param.go
sparknack: This is valid but unrelated to the loading-estimation changes. Handling malformed explicit overrides requires changing the shared config-validation behavior, so we will keep this PR scoped and leave it unchanged.

internal/core/src/index/IndexFactory.cpp
sparknack: The estimator intentionally uses the requested mmap mode without reading cardinality to predict `BitmapIndex`'s internal load-mode selection. Selection-aware estimation would require additional persisted metadata or load-time metadata I/O, which is outside this backport. Leaving unchanged.

internal/core/src/segcore/storagev2translator/GroupCTMeta.h
sparknack: The estimate targets the default `FileWriter` configuration; accounting for the maximum configurable writer buffer would substantially overreserve normal loads. The buffers remain bounded by the active load workers, so we will leave this unchanged.

internal/core/src/index/Utils.cpp
sparknack: Encrypted stream buffers are already bounded by the shared `TransientMemoryBudget`. This PR does not duplicate that transient slice accounting in each index's admission estimate, so we will leave the current estimate unchanged.

internal/core/src/index/StringIndexSort.cpp
sparknack: Load-to-serialize is not a supported production lifecycle. IndexNode uses Build-to-Upload/Serialize, while QueryNode uses Load-to-Query; the Go `CgoIndex.Serialize()` path is test-only. We will not extend `StringIndexSort`'s serialization contract for this unused API combination.

internal/core/src/index/IndexFactory.cpp
sparknack: This backport only supports the current V3 layout, where the CSR data is persisted as side entries. Earlier V3 artifacts without these entries are outside the compatibility scope, so we will not reserve memory for rebuilding legacy CSR data.

internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp
```diff
@@ -73,6 +74,25 @@ SealedIndexTranslator::SealedIndexTranslator(
                 index_info_.index_type, knowhere::feature::LAZY_LOAD))) {
 }
 
+const LoadResourceRequest&
+SealedIndexTranslator::EstimateLoadResource() const {
+    std::call_once(load_resource_request_once_, [this]() {
```
congqixia, line 79: **P2: cached `EstimateLoadResource` freezes the load-pool concurrency used by the stream-memory bound**

`ScalarIndexStreamMemoryOverhead()` (reached via `IndexLoadResource`) reads `pool.GetMaxThreadNum()` and bounds the V3 stream peak as `min(index_size, workers × slice_size)`. Caching the whole `LoadResourceRequest` behind `std::call_once` here freezes that worker count for the lifetime of the translator.

If the HIGH/LOW load pool is resized at runtime, the admission estimate no longer matches the real concurrent loading peak (under-estimate when the pool grows, over-estimate when it shrinks). Since the same cached value backs both reservation and cell accounting, the discrepancy persists for the slot lifetime. Note `CachesLoadEstimateAfterFirstUse` currently pins the "estimate unchanged after pool resize" behavior in a test.

Suggest documenting this as an intentional snapshot, or keying the cache on `GetMaxThreadNum()` / a pool revision counter so the estimate can be refreshed on resize.

internal/core/src/segcore/storagev1translator/DefaultValueChunkTranslator.cpp
congqixia, line 337: **P2: the shared-buffer cache charge can be lost if the claiming cell leaves the slot**

`get_cells()` claims the backing-buffer charge exactly once via `exchange(true)` on `primary_buffer_accounted_`/`tail_buffer_accounted_` and attributes the full buffer size to `res.front()` through `SetCellSize`; every other view stays zero-charge. `estimated_loading_usage()` only *peeks* the same flags with acquire loads and never participates in the claim.

Window: if the charging (first) cell is later evicted or reloaded while zero-charge views of the same backing buffer remain resident, no other cell re-claims the buffer, so its resident usage silently drops off the slot accounting for the rest of the lifetime. The "cells are non-evictable" comment assumes this away, but nothing in the code enforces it.

Suggest making the charge follow the buffer instead of the cell — e.g. let `estimated_loading_usage()` perform the same atomic claim and report the buffer only when this call wins it, or tie the claim to the slot lifetime independently of which cell is resident.

internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp
sparknack: Runtime resizing is technically available, but it is not part of the expected operational path for this 2.6 backport. We intentionally snapshot the complete load-resource request on first use so HYBRID metadata is resolved only once and the reservation and final cell accounting use the same request. If a load pool is resized after that point, the estimate can remain conservative or stale; we accept that limitation here rather than adding cache invalidation and synchronization machinery to this backport. `CachesLoadEstimateAfterFirstUse` documents this snapshot behavior. Leaving this unchanged.

internal/core/src/segcore/storagev1translator/DefaultValueChunkTranslator.cpp
sparknack: For this translator, the cache slot is created with `support_eviction=false`, so the charged owner cell cannot leave through normal strategy eviction. In the current 2.6 production path, the slot is encapsulated by the default-value `ChunkedColumn`; that path neither exposes nor calls `ManualEvict`, and segment release destroys the slot, all cells, and the shared buffers together. We therefore treat manual eviction of an individual default-value cell as unsupported for this backport and accept the low-level `ManualEvict` edge case. If manual eviction is exposed for this column in the future, the shared-buffer charge will need to move to slot-owned lifetime accounting. Leaving this unchanged.

congqixia (commented):
/lgtm

internal/core/src/storage/FileWriter.h
```diff
@@ -293,6 +293,45 @@ class FileWriter {
     io::WriteRateLimiter& rate_limiter_;
 };
 
+class PositionedFileWriter {
```
zhengbuqian, line 296: explain all differences between PositionedFileWriter and FileWriter. why is it not sufficient to add a Seek to FileWriter?

zhengbuqian (approved):
/hold

congqixia (approved):
/lgtm

PR #52672, opened 2026-08-19
enhance: extract proxy channel manager

congqixia:
Related to #44761

Extract the proxy channel management into internal/proxy/channelmgr so write-path channel resolution and insert packing live outside the monolithic proxy package.

The channel resolver is now injected at construction instead of calling DescribeCollection directly: Proxy.Init builds a ChannelsMgr whose provider reads the meta cache, dropping the extra coordinator RPC. The channelmgr keeps no per-collection cache of its own — every lookup delegates to the injected resolver, so caching and invalidation are owned entirely by the meta cache and channel metadata can never go stale. The meta cache is initialized before the channels manager so the injected resolver never observes a nil cache.

Each resolver result is re-validated for vchan/pchan alignment at the boundary, preserving the old getDmlChannelsFunc consistency guard.

Remove the dead channels_time_ticker code and its unused chTicker fields on DML tasks. Move GenInsertMsgsByPartition and GetActiveWALName into the new package and export them for the streaming write path.

Remove the now-orphaned ProxyMsgStreamObjectsForPChan gauge, whose only writers (the channelmgr cache fill/remove) no longer exist.

Note: write-path channel resolution errors now surface from metaCache.GetCollectionInfo (e.g. ErrCollectionNotFound) instead of the raw DescribeCollection status; this is an intentional improvement in Input-vs-System error classification.

Add package README, OWNERS, and package-local unit tests that inject a fake resolver.


internal/proxy/proxy.go
```diff
@@ -259,8 +260,26 @@ func (node *Proxy) Init() error {
...
+	node.setMetaCache(metaCache)
+	mlog.Debug(node.ctx, "init meta cache done", mlog.String("role", typeutil.ProxyRole))
+
+	chMgr := channelmgr.NewChannelsMgr(
+		func(collectionID typeutil.UniqueID) (channelmgr.ChannelInfo, error) {
+			collInfo, err := metaCache.GetCollectionInfo(node.ctx, "", "", collectionID)
+			if err != nil {
+				return channelmgr.ChannelInfo{}, err
+			}
+			return channelmgr.ChannelInfo{VChans: collInfo.VChannels, PChans: collInfo.PChannels}, nil
```
XuanYang-cn, line 279: This bypasses `newChannels` and its `len(vchans)==len(pchans)` check, which the old `getDmlChannelsFunc` path enforced (returning `ServiceInternal` on mismatch). Metacache copies `VirtualChannelNames`/`PhysicalChannelNames` verbatim with no validation, so a coordinator consistency guard is silently dropped. Please run `newChannels` here (or otherwise assert the invariant) so a mismatched pair can't reach the write/read paths.

internal/proxy/channelmgr/channels_mgr.go
```diff
@@ -0,0 +1,174 @@
...
+func (mgr *channelsMgrImpl) getAllChannels(collectionID typeutil.UniqueID) (ChannelInfo, error) {
+	mgr.mu.RLock()
+	defer mgr.mu.RUnlock()
+
+	infos, ok := mgr.infos[collectionID]
+	if ok {
+		return infos.channelInfo, nil
+	}
+
+	return ChannelInfo{}, merr.WrapErrParameterInvalidMsg("collection not found in channels manager: %d", collectionID)
```
XuanYang-cn: A local cache-miss is internal state, not request content — per `docs/dev/error_handling_guide.md` (Input vs System) this should be a System error (`WrapErrServiceInternalMsg`), not `ParameterInvalid`. Today it is swallowed by `ensureChannels` as control flow so it never surfaces, but the classification is still wrong and would mislead if it ever leaks.

internal/proxy/channelmgr/channels_mgr.go
```diff
@@ -0,0 +1,174 @@
...
+		delete(mgr.infos, collectionID)
+	}
+	mlog.Info(context.TODO(), "dml stream removed", mlog.Int64("collection_id", collectionID))
+}
+
+// NewChannelsMgr constructs a channels manager backed by the given resolver.
+// getChannelsFunc resolves collection channels; repackFunc repacks messages.
+func NewChannelsMgr(
+	getChannelsFunc GetChannelsFunc,
+	repackFunc RepackFunc,
```
XuanYang-cn: `repackFunc` is stored but never read anywhere; `DefaultInsertRepackFunc` (the only value passed here in production) has no consumer — the streaming path calls `channelmgr.GenInsertMsgsByPartition`/`GetActiveWALName` directly. This is speculative generality: consider dropping the `repackFunc` param/field until there is a real caller.

internal/proxy/channelmgr/channels_mgr.go
```diff
@@ -0,0 +1,174 @@
...
+// RemoveStream removes the corresponding stream of the specified collection. Idempotent.
+// If stream already exists, remove it, otherwise do nothing.
+func (mgr *channelsMgrImpl) RemoveStream(collectionID typeutil.UniqueID) {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+	if info, ok := mgr.infos[collectionID]; ok {
+		decPChanMetrics(info.channelInfo.PChans)
+		delete(mgr.infos, collectionID)
+	}
+	mlog.Info(context.TODO(), "dml stream removed", mlog.Int64("collection_id", collectionID))
```
XuanYang-cn: Minor: use `mlog.FieldCollectionID(collectionID)` instead of `mlog.Int64("collection_id", ...)` per `logging.md`; also `RemoveStream` has no ctx but its caller `InvalidateCollectionMetaCache` does — consider threading a ctx rather than `context.TODO()`.

XuanYang-cn (changes requested):
Requesting changes. The extraction is clean and compiles, and dropping the dead `channels_time_ticker` (its `chTicker` field was assigned in tests only, never read in production) is correct. But a few things should be addressed before merge:

1. **Dropped vchan/pchan alignment validation (behavioral regression).** The production resolver now returns `ChannelInfo` straight from metacache and no longer runs `newChannels`' `len(vchans)==len(pchans)` guard (see inline on `proxy.go:279`). Metacache copies the two lists verbatim (`meta_cache.go:654-655`) with no check, so a coordinator consistency guard the old `getDmlChannelsFunc` path provided is silently lost.

2. **"removing the duplicate channel cache" in the PR body is inaccurate.** The lazy `infos` cache with `ensureChannels` double-checked locking is still present — it was moved to `channelmgr`, not removed. Only the `DescribeCollection` RPC + `getDmlChannelsFunc` were deleted. Please fix the description.

3. **`DefaultInsertRepackFunc` has no production consumer.** It is exported and threaded through `NewChannelsMgr`'s `repackFunc` param, but that field is never read (see inline). The streaming path only uses `GenInsertMsgsByPartition` and `GetActiveWALName`, so the "export them for the streaming write path" claim doesn't hold for `DefaultInsertRepackFunc`.

4. Minor: `newChannels`/`removeDuplicate` are now test-only (a consequence of point 1), and the two `context.TODO()` log calls in the new package could take a real ctx (the `RemoveStream` caller `InvalidateCollectionMetaCache` has one).

internal/proxy/channelmgr/README.md
```diff
@@ -0,0 +1,97 @@
...
+actual message-stream topic/WAL). DML write tasks (insert/delete/upsert) and
+read paths (search/query) need the channel list of a collection before they can
+dispatch work. This package owns that lookup.
+
+## Responsibilities
+
+1. **Channel resolution**: resolve `(vchans, pchans)` for a collection id via an
+   injected `GetChannelsFunc`, validating the vchan/pchan alignment on every
+   resolver result.
+2. **No channel cache of its own**: the package deliberately keeps no per-
```
XuanYang-cn, line 21: The PR description says "The lazy per-collection cache is retained inside channelmgr", but the code and this README state the opposite: resolution delegates to the injected resolver (metaCache) on every call and there is no per-collection cache. Could you update the PR body to say the cache was removed and caching is now owned by the meta cache? As written, the body contradicts the implementation (and the unit test asserts "no internal cache: every call hits the resolver").

internal/proxy/proxy.go
```diff
@@ -259,8 +260,25 @@ func (node *Proxy) Init() error {
...
+	// injected channel resolver always observes a live cache (no nil window).
+	metaCache, err := initMetaCache(node.ctx, node.mixCoord)
+	if err != nil {
+		mlog.Warn(node.ctx, "failed to init meta cache", mlog.String("role", typeutil.ProxyRole), mlog.Err(err))
+		return err
+	}
+	node.setMetaCache(metaCache)
+	mlog.Debug(node.ctx, "init meta cache done", mlog.String("role", typeutil.ProxyRole))
+
+	chMgr := channelmgr.NewChannelsMgr(
```
XuanYang-cn, line 273: This refactor removes the only writers of the ProxyMsgStreamObjectsForPChan gauge (incPChansMetrics/decPChanMetrics in the deleted channels_mgr.go). After this change the metric is still registered in pkg/metrics/proxy_metrics.go but never updated, so it will sit at 0 forever. Should we also remove/deprecate the metric definition, or is it expected to be populated elsewhere? Worth a note in the PR body either way.

internal/proxy/task_test.go
```diff
@@ -65,6 +67,27 @@ import (
 
 // TODO(dragondriver): add more test cases
 
+// newTestChannelsMgr builds a channels manager backed by the coordinator's
+// DescribeCollection, mirroring the production provider wiring in Proxy.Init.
```
XuanYang-cn: newTestChannelsMgr resolves via DescribeCollection, but the production wiring in Proxy.Init now reads metaCache, so "mirroring the production provider wiring in Proxy.Init" is inaccurate. Suggest rewording to something like "backed by the coordinator DescribeCollection (test-only)".

internal/proxy/proxy.go
```diff
@@ -259,8 +260,25 @@ func (node *Proxy) Init() error {
...
+		func(collectionID typeutil.UniqueID) (channelmgr.ChannelInfo, error) {
+			collInfo, err := metaCache.GetCollectionInfo(node.ctx, "", "", collectionID)
```
XuanYang-cn, line 275: The error returned to callers changed here: previously getDmlChannelsFunc returned merr.Error(resp.GetStatus()) from DescribeCollection (or the raw RPC error); now write-path channel resolution returns metaCache.GetCollectionInfo's error (e.g. ErrCollectionNotFound). This likely improves Input-vs-System classification, but it is a behavior change not mentioned in the PR body — can you confirm it is intentional?

PR #52676, opened 2026-08-19
enhance: collect cached filesystem metrics on scrape

jiaqizho:
issue: https://github.com/milvus-io/milvus/issues/47520

Filesystem metrics were previously pushed from task-completion paths, which could miss cached filesystems and leave stale values. Collect them directly from the storage cache at scrape time instead. Update milvus-storage to use the cache-wide metrics API, expose cached snapshots through storagev2, and register a thread-safe Prometheus collector while preserving existing metric names and labels.

Remove the per-task publishing calls and old config-based helpers so compaction, index, flush, and load paths no longer manage filesystem observability.

### DIFF
--- a/internal/core/thirdparty/milvus-storage/CMakeLists.txt
+++ b/internal/core/thirdparty/milvus-storage/CMakeLists.txt
@@ -14,7 +14,7 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( milvus-storage_VERSION 3ae6ac4)
+set( milvus-storage_VERSION ced9c9a)
 set( GIT_REPOSITORY  "https://github.com/milvus-io/milvus-storage.git")
 message(STATUS "milvus-storage repo: ${GIT_REPOSITORY}")
 message(STATUS "milvus-storage version: ${milvus-storage_VERSION}")
--- a/internal/core/thirdparty/milvus-storage/CMakeLists.txt
+++ b/internal/core/thirdparty/milvus-storage/CMakeLists.txt
@@ -14,7 +14,7 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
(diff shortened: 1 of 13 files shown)


internal/core/thirdparty/milvus-storage/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # Update milvus-storage_VERSION for the first occurrence
 milvus_add_pkg_config("milvus-storage")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( milvus-storage_VERSION 3ae6ac4)
+set( milvus-storage_VERSION ced9c9a)
```
weiliu1031: [P1] Do not roll back the S3 CRT finalization fix

`ced9c9a` is the direct parent of the current `3ae6ac4` pin, so this change downgrades `milvus-storage` and removes the [S3 CRT client finalization fix](https://github.com/milvus-io/milvus-storage/commit/3ae6ac424b8dd3b7dcd06698f6779bb87e269dee). That reintroduces the ownership and shutdown races which can deadlock while an asynchronous CRT request or client construction is still active.

The cache-wide filesystem metrics API was merged before both revisions and is already available in `3ae6ac4`, so this downgrade is not required for this PR. Please keep `3ae6ac4` or use a descendant that preserves the finalization fix.

internal/storagev2/filesystem_metrics.go
```diff
@@ -74,6 +76,35 @@ func getMetricsFromHandle(cFilesystem C.FileSystemHandle) (*FilesystemMetrics, e
 	return fsMetrics, nil
 }
 
+// ListFilesystemMetrics returns metrics for every filesystem currently held by the cache.
+func ListFilesystemMetrics() ([]FilesystemMetricsEntry, error) {
```
weiliu1031, line 80: [P3] Remove the now test-only point lookup path

Now that this cache-wide API replaces all production calls to `GetFilesystemMetricsWithConfig`, the old `getMetricsFromHandle` / `makePropertiesFromConfig` / `GetFilesystemMetricsWithConfig` path is left with test-only callers. Keeping it retains a large duplicate `StorageConfig`-to-FFI property mapping in production code, which can drift from the filesystem construction path.

Could the affected tests select their target entry from `ListFilesystemMetrics` through a small test helper, and remove the old point-lookup implementation?

internal/core/thirdparty/milvus-storage/CMakeLists.txt
jiaqizho: ack

PR #52678, opened 2026-08-19
fix: [cp3.0] avoid legacy stats paths after V3 manifest failures

sijie-ni-0214:
Cherry-pick from master
pr: #52626

issue: #52623

## What this PR does

Prevent Storage V3 segment load-info conversion from falling through to Storage V2 text-index and JSON-stats path construction when manifest resolution fails.

A V3 manifest failure now propagates through `ConvertToSegcoreSegmentLoadInfo`, `CreateCSegment`, and `Reopen`. Reopen fails before mutating the existing segment state, allowing transient manifest failures to be retried.

## Test Plan

- `make build-go`
- `go test -v -count=1 -tags dynamic,test -gcflags="all=-N -l"` with the required macOS rpaths for `./internal/util/segcore`
- `gofumpt -l internal/util/segcore/segment.go internal/util/segcore/segment_test.go`
- `git diff --check`

### DIFF
--- a/internal/util/segcore/segment.go
+++ b/internal/util/segcore/segment.go
@@ -71,7 +71,10 @@ func CreateCSegment(req *CreateCSegmentRequest) (CSegment, error) {
 	var ptr C.CSegmentInterface
 	var status C.CStatus
 	if req.LoadInfo != nil {
-		segLoadInfo := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+		segLoadInfo, err := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+		if err != nil {
+			return nil, merr.Wrap(err, "failed to convert segment load info")
+		}
 		loadInfoBlob, err := proto.Marshal(segLoadInfo)
 		if err != nil {
 			return nil, err
@@ -356,7 +359,10 @@ func (s *cSegmentImpl) Reopen(ctx context.Context, req *ReopenRequest) error {
 	defer runtime.KeepAlive(traceCtx)
 	defer runtime.KeepAlive(req)
 
-	segLoadInfo := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+	segLoadInfo, err := ConvertToSegcoreSegmentLoadInfo(req.LoadInfo)
+	if err != nil {
+		return merr.Wrap(err, "failed to convert reopen load info")
+	}
 	loadInfoBlob, err := proto.Marshal(segLoadInfo)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


czs007:

Verified:
- `internal/util/segcore/segment.go:482` — `resolveStatsWithBasePaths` now returns `merr.Wrap(result.Err(), ...)` instead of `return nil, nil, nil, nil`, so a V3 manifest failure can no longer be mistaken for "no stats" by segcore's load diff.
- `internal/util/segcore/segment.go:74-77` / `:362-365` — both call sites of `ConvertToSegcoreSegmentLoadInfo` check the error and return early; in `Reopen` this happens before `C.AsyncReopenSegment`, and a repo-wide grep shows no other caller of the changed signature on this branch.
- `internal/querynodev2/segments/segment.go:1301-1312` — on a failed `csegment.Reopen`, `syncFieldIndexes` / `loadInfo.Store(newLoadInfo)` / `syncFieldJSONStatsFromLoadInfo` are all skipped, so local state does not advance and the reopen is retryable (`segment_loader.go:2481-2485` → `handlers.go:151-155` → `merr.Status(err)`).
- `internal/storagev2/packed/stats_resolver.go:236-244` — the new hard failure only triggers when `ManifestPath != ""`; V3 segments without a manifest and growing segments (`delegator_data.go:179-187`) still take the no-error branch.
- `internal/util/segcore/segment.go:488-505` — the V2 branch is byte-for-byte unchanged and now returns a nil error, so non-V3 segments are unaffected; `merr.Wrap` preserves the inner error code (e.g. `ErrLoonTransient` from `ffi_common.go:337`).
- `internal/util/segcore/segment_test.go:474-490` — the new negative test fails deterministically in `UnmarshalManifestPath`'s `json.Unmarshal` before any FFI/network call, and all six existing call sites were updated with `assert.NoError`.

The commit content is byte-identical to the final upstream #52626 diff — no cherry-pick drift.

## Low — non-blocking

- `internal/util/segcore/segment.go:74` — The two new error-propagation branches (`CreateCSegment` at `:74-77` and `cSegmentImpl.Reopen` at `:362-365`) have no call-site-level test: `segment_test.go:474-490` only calls `ConvertToSegcoreSegmentLoadInfo` directly, the `CreateCSegment` test at `segment_test.go:49-54` passes no `LoadInfo`, and there is no `Reopen` test in the package — reverting either call site to `segLoadInfo, _ := ...` would keep the package green. Suggestion: on top of the real `CCollection` built at `segment_test.go:49`, add a case with `LoadInfo: &querypb.SegmentLoadInfo{StorageVersion: storage.StorageV3, ManifestPath: "not-a-valid-manifest-path"}` and assert `CreateCSegment` returns an error. (raised by czs007, tinswzy)

## Carried over from #52626

- **[high] `internal/util/segcore/segment.go:472` — manifest parse failure silently unloads json stats / unlocks text index rebuild on Reopen** → **Resolved upstream**: the fix came across with the cherry-pick — reviewers checked `segment.go:482` now returns a wrapped error, both call sites (`:74-77`, `:362-365`) return before marshal/`AsyncReopenSegment`, and `LocalSegment.Reopen` (`segments/segment.go:1301-1311`) returns before storing the new load info.
- **[medium] `internal/querynodev2/segments/segment.go:1247` — after a silent drop, QueryNode still reports json stats loaded** → **Resolved upstream**: the premise (conversion swallows the error while sync reports stats present) is no longer reachable — reviewers checked `NewSegment` aborts at `segments/segment.go:495-505` and `Reopen` returns at `:1304-1306` before `syncFieldJSONStatsFromLoadInfo`; the residual reporting divergence is tracked under the next item.
- **[low] `internal/querynodev2/segments/segment.go:1240` — `syncFieldJSONStatsFromLoadInfo` residual silent degradation / double parse** → **Confirmed on this branch** at `segments/segment.go:1240-1247` (fallback to `loadInfo.GetJsonKeyStatsLogs()` on resolver error; second `GetManifestStats` FFI call at `:1312` after the one in `segcore/segment.go:476`). Pre-existing and not introduced by this PR — see the follow-up section below.
- **[low] `internal/util/segcore/segment.go:74` — `CreateCSegment`/`Reopen` call sites lack tests** → **Confirmed on this branch** at `segment_test.go:49-54` / `:474-490`; listed above under Low.

- `internal/querynodev2/segments/segment.go:1247` — `syncFieldJSONStatsFromLoadInfo` still falls back to `loadInfo.GetJsonKeyStatsLogs()` when the resolver errors (`:1242-1247`), inconsistent with the "V3 manifest failure is an error, no fallback" policy this PR establishes in `resolveStatsWithBasePaths`; for V3 that map is the querycoord-filled placeholder (`querycoordv2/utils/types.go:103-110`). Segcore Load/Reopen succeeds on the first manifest parse, then the second independent `GetManifestStats` FFI call in `syncFieldJSONStatsFromLoadInfo` fails on a transient object-store error → `fieldJSONStats` is silently populated from the proto placeholder rather than the manifest segcore actually loaded (reporting/consistency only, no query error), plus one extra manifest read per Load/Reopen since the resolver's manifest cache (`stats_resolver.go:362-376`) is per-instance. Unrelated to this PR because the PR only touches `internal/util/segcore/segment.go` / `segment_test.go`; `syncFieldJSONStatsFromLoadInfo` and its two callers (`:1281`, `:1312`) are unchanged. Suggested follow-up: reuse the already-resolved stats, or stop falling back to the proto map when `ManifestPath != ""`. No tracking issue yet. (raised by czs007, tinswzy)

czs007 (approved):
/lgtm

Issue #52682, opened 2026-08-19
[Bug]: collection labeled metrics are not cleaned up when a collection is dropped


daviddallakyan2005:
### Is there an existing issue for this?

I have searched the existing issues. The closest open item is #51341, which covers ProxyFunctionCall series being recreated after DropCollection. That is a different bug. This issue is about four metrics that collection cleanup never deletes.

### Environment

Milvus version: master
Deployment mode: any
The leak is in process metrics cleanup. It is not deployment specific.

### Current Behavior

CleanupProxyCollectionMetrics already cleans ProxyInsertVectors, ProxyUpsertVectors and ProxySearchVectors. It does not clean ProxyDeleteVectors. Insert, upsert and search are cleaned when a collection is dropped. Delete was forgotten.

The project already ships collection cleanup helpers and drop already calls them. CleanupQueryNodeCollectionMetrics lives in pkg/metrics/querynode_metrics.go and is called from internal/querynodev2/segments/collection.go. CleanupProxyCollectionMetrics lives in pkg/metrics/proxy_metrics.go and is called from internal/proxy/impl.go. CleanupProxyDBMetrics also exists, and that is where ProxyDeleteVectors is deleted today. Database drop is not collection drop.

Four series carry collection labels, are observed on live paths, and are never deleted by the collection cleanup:

1. QueryNodeSearchFTSNumTokens
   defined pkg/metrics/querynode_metrics.go
   labels node_id, collection_id, field_id
   observed internal/querynodev2/delegator/delegator_data.go
   not among the DeletePartialMatch calls in CleanupQueryNodeCollectionMetrics

2. ProxySearchSparseNumNonZeros
   defined pkg/metrics/proxy_metrics.go
   labels node_id, collection_name, query_type, field_id (no db_name)
   observed internal/proxy/task_search.go
   not in CleanupProxyCollectionMetrics

3. ProxyFunctionlatency
   defined pkg/metrics/proxy_metrics.go
   labels node_id, collection_name, function_type, provider, function_name (no db_name)
   observed internal/util/function/embedding/function_executor.go
   not in CleanupProxyCollectionMetrics

4. ProxyDeleteVectors
   defined pkg/metrics/proxy_metrics.go
   labels node_id, db_name, collection_name
   observed internal/proxy/task_delete.go
   has DeletePartialMatch inside CleanupProxyDBMetrics, not the collection cleanup

[...]

PR #52684, opened 2026-08-19
fix: clean up collection labeled metrics that leak on collection drop [skip e2e]

daviddallakyan2005:
CleanupProxyCollectionMetrics already removes ProxyInsertVectors, ProxyUpsertVectors and ProxySearchVectors. It did not remove ProxyDeleteVectors. Insert, upsert and search were cleaned on collection drop. Delete was forgotten. The same drop path also left QueryNodeSearchFTSNumTokens, ProxySearchSparseNumNonZeros and ProxyFunctionlatency in place. Sparse search and function latency declare no db_name, so cleanup matches only node_id and collection_name.

issue: #52682

/cc @zhengbuqian

## Test

Guard tests observe every collection labeled QueryNode and Proxy metric, call cleanup, then gather from a registry. The observe list is explicit. A new collection labeled metric must be added to that list to be guarded.

Without the four DeletePartialMatch lines, from pkg/:

go test ./metrics/ -count=1 -timeout 180s -run 'TestCleanupQueryNodeCollectionMetricsDropsEveryCollectionSeries|TestCleanupProxyCollectionMetricsDropsEveryCollectionSeries'

FAIL leftover milvus_querynode_search_fts_num_tokens
FAIL leftover milvus_proxy_delete_vectors_count milvus_proxy_function_udf_call_latency milvus_proxy_search_sparse_num_non_zeros

With the fix, from pkg/:

go test ./metrics/ -count=1 -timeout 180s -run 'TestCleanupQueryNodeCollectionMetricsDropsEveryCollectionSeries|TestCleanupProxyCollectionMetricsDropsEveryCollectionSeries'

ok github.com/milvus-io/milvus/pkg/v3/metrics 0.417s

go test ./metrics/... -count=1 -timeout 180s

ok github.com/milvus-io/milvus/pkg/v3/metrics 0.331s

15 PASS, 0 FAIL.

### DIFF
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -209,3 +209,161 @@ func TestDeletePartialMatch(t *testing.T) {
 	}
 	assert.Equal(t, 0, getMetricsCount())
 }
+
+func queryNodeCollectionCollectors() []prometheus.Collector {
+	return []prometheus.Collector{
+		QueryNodeConsumeTimeTickLag,
+		QueryNodeConsumerMsgCount,
+		QueryNodeSkippedInsertFieldCount,
+		QueryNodeNumSegments,
+		QueryNodeSQCount,
+		QueryNodeSearchFTSNumTokens,
+		QueryNodeSearchHitSegmentNum,
+		QueryNodeSegmentFilterHitSegmentNum,
+		QueryNodeSegmentFilterSkippedSegmentNum,
+		QueryNodeSegmentFilterTotalSegmentNum,
+		QueryNodeSegmentPruneRatio,
+		QueryNodeSegmentPruneBias,
+		QueryNodeSegmentPruneLatency,
+		QueryNodeNumEntities,
+		QueryNodeEntitiesSize,
+		QueryNodeLevelZeroSize,
+		QueryNodePartialResultCount,
+		QueryNodeTwoStageFilterLatency,
+		QueryNodeTwoStageSearchLatency,
+		QueryNodeTwoStageSearchFallbackCount,
+		QueryNodeGlobalRefineCount,
+	}
+}
+
+func observeQueryNodeCollection(nodeID, collectionID string) {
+	QueryNodeConsumeTimeTickLag.WithLabelValues(nodeID, InsertLabel, collectionID).Set(1)
+	QueryNodeConsumerMsgCount.WithLabelValues(nodeID, InsertLabel, collectionID).Add(1)
+	QueryNodeSkippedInsertFieldCount.WithLabelValues(nodeID, collectionID).Add(1)
+	QueryNodeNumSegments.WithLabelValues(nodeID, collectionID, SealedSegmentLabel, "L1").Set(1)
+	QueryNodeSQCount.WithLabelValues(nodeID, SearchLabel, SuccessLabel, Leader, collectionID).Add(1)
+	QueryNodeSearchFTSNumTokens.WithLabelValues(nodeID, collectionID, "1").Observe(1)
+	QueryNodeSearchHitSegmentNum.WithLabelValues(nodeID, collectionID, SearchLabel).Observe(1)
+	QueryNodeSegmentFilterHitSegmentNum.WithLabelValues(nodeID, collectionID, SearchLabel).Observe(1)
+	QueryNodeSegmentFilterSkippedSegmentNum.WithLabelValues(nodeID, collectionID, SearchLabel).Observe(1)
+	QueryNodeSegmentFilterTotalSegmentNum.WithLabelValues(nodeID, collectionID, SearchLabel).Observe(1)
+	QueryNodeSegmentPruneRatio.WithLabelValues(nodeID, collectionID, "x").Set(1)
+	QueryNodeSegmentPruneBias.WithLabelValues(nodeID, collectionID, "x").Set(1)
+	QueryNodeSegmentPruneLatency.WithLabelValues(nodeID, collectionID, "x").Observe(1)
+	QueryNodeNumEntities.WithLabelValues("db", "coll", nodeID, collectionID, GrowingSegmentLabel).Set(1)
+	QueryNodeEntitiesSize.WithLabelValues(nodeID, collectionID, GrowingSegmentLabel).Set(1)
+	QueryNodeLevelZeroSize.WithLabelValues(nodeID, collectionID, "ch").Set(1)
+	QueryNodePartialResultCount.WithLabelValues(nodeID, SearchLabel, collectionID).Add(1)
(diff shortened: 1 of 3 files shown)


xiaofan-luan:
Thanks for digging into this — I verified the change locally and the four additions are correct. I also found that the proxy side is still leaking, and that the guard test as written cannot see it. Details below, plus a patch you're welcome to fold into this PR.

## What I verified

Reverting only `proxy_metrics.go` / `querynode_metrics.go` and keeping your tests reproduces exactly what you describe:

```
Should be empty, but was [milvus_querynode_search_fts_num_tokens]
Should be empty, but was [milvus_proxy_delete_vectors_count milvus_proxy_function_udf_call_latency milvus_proxy_search_sparse_num_non_zeros]
```

and `go test ./metrics/... -count=1` is green with the fix. The label subsets you pass to the two new `DeletePartialMatch` calls are valid.

The **query node side is now complete**: I parsed every `prometheus.New*Vec` in `querynode_metrics.go` and all 21 metrics carrying `collection_id` are covered by `CleanupQueryNodeCollectionMetrics`, with no metric carrying only `collection_name`.

## What still leaks

`CleanupProxyCollectionMetrics` deletes most of its series with `Delete()` and a fully specified label set (proxy_metrics.go:639-788), so every `msg_type` / `query_type` value has to be enumerated. Those enumerations have gone stale.

The guard test can't catch this, because `observeProxyCollection` observes exactly one label combination per metric — and that combination happens to be one the enumeration covers. I re-ran it observing the values `internal/proxy/impl.go` actually emits:

```
LEAKED after cleanup: [milvus_proxy_collection_mutation_latency
                       milvus_proxy_collection_sq_latency
                       milvus_proxy_receive_bytes_count
                       milvus_proxy_scanned_remote_mb
                       milvus_proxy_scanned_total_mb]
```

| metric | emitted with | cleanup enumerates | leaks |
|---|---|---|---|
| `ProxyCollectionSQLatency` | search, **hybrid_search** (impl.go:3345), query/upsert_query | search, query, upsert_query | hybrid_search |
| `ProxyCollectionMutationLatency` | delete (impl.go:2704), **upsert** (impl.go:2850) | insert, delete | upsert (and `insert` is enumerated but never emitted) |
| `ProxyReceiveBytes` | insert, delete, upsert, search, **hybrid_search** (impl.go:3218), query | all but hybrid_search | hybrid_search |
| `ProxyScannedRemoteMB` / `ProxyScannedTotalMB` | search, **hybrid_search** (impl.go:3353/3360), query, upsert, delete | all but hybrid_search | hybrid_search |

`ProxyReceivedNQ`, `ProxyRetrySearchCount`, `ProxyRetrySearchResultInsufficientCount` and `ProxyRecallSearchCount` happen to line up today, but they are one new `query_type` away from the same bug.

## Suggested shape

The invariant is "after a collection drop, no series with this (node, db, collection) remains". Enumerating label values is the wrong owner for it — `DeletePartialMatch` on `(node_id, db_name, collection_name)` is, which is exactly what you did for the four metrics you added. Doing that for all of them removes the whole class:

```go
type partialMatchDeleter interface {
	DeletePartialMatch(labels prometheus.Labels) int
}

// proxyCollectionScopedMetrics lists every proxy metric that carries both a
// db_name and a collection_name label, so a dropped database or collection can
// be cleaned from one place.
func proxyCollectionScopedMetrics() []partialMatchDeleter {
	return []partialMatchDeleter{
		ProxySearchVectors, ProxyInsertVectors, ProxyUpsertVectors, ProxyDeleteVectors,
		ProxySQLatency, ProxyCollectionSQLatency,
		ProxyMutationLatency, ProxyCollectionMutationLatency,
		ProxyFunctionCall, ProxyFunctionlatency,
		ProxyReceivedNQ, ProxyReceiveBytes,
		ProxyRetrySearchCount, ProxyRetrySearchResultInsufficientCount, ProxyRecallSearchCount,
		ProxySearchSparseNumNonZeros, ProxyScannedRemoteMB, ProxyScannedTotalMB,
	}
}

func CleanupProxyCollectionMetrics(nodeID int64, dbName string, collection string) {
	labels := prometheus.Labels{
		nodeIDLabelName:   strconv.FormatInt(nodeID, 10),
		databaseLabelName: dbName,
		collectionName:    collection,
	}
	for _, m := range proxyCollectionScopedMetrics() {
		m.DeletePartialMatch(labels)
	}
}
```

That turns ~150 lines into ~45 and, sharing the same list, `CleanupProxyDBMetrics` stops leaving collection series behind on drop database (today it only cleans 7 of these 18).

## `db_name` on the two metrics you added

`search_sparse_num_non_zeros` and `function_udf_call_latency` have no `db_name` label, so partial-matching them on `(node_id, collection_name)` wipes a same-named collection in a **different** database. Since collection names are only unique per db, the cleaner fix is to add the label:

- `[]string{nodeIDLabelName, databaseLabelName, collectionName, queryTypeLabelName, fieldIDLabelName}`, call sites `internal/proxy/task_search.go:692,1041` pass `t.request.GetDbName()`
- `[]string{nodeIDLabelName, databaseLabelName, collectionName, functionTypeName, functionProvider, functionName}`, with `FunctionExecutor` carrying `extraInfo.DBName`

Neither metric is referenced by any dashboard under `deployments/`, so the added label is safe.

## Making the guard test actually guard

Two changes make it catch the above:

1. `observeProxyCollection` should loop over every `query_type` / `msg_type` production emits, not one value per metric.
2. A test that parses the metric declarations and fails when a metric labeled `db_name`+`collection_name` (proxy) or `collection_id` (query node) is not referenced by the corresponding cleanup — that is the failure mode that produced this issue in the first place. Removing `ProxyReceiveBytes` from the list gives:

```
ProxyReceiveBytes is labeled with [databaseLabelName collectionName] but is never cleaned up in proxyCollectionScopedMetrics
```

I have all of this working locally on top of your commit (`pkg/metrics` green, `golangci-lint` clean, mutation-tested red/green in both directions). Happy to push the branch somewhere for you to cherry-pick, or to open it as a follow-up PR after this one merges — whichever you prefer.

xiaofan-luan:
Filed the remaining leaks as #52690 and opened #52691 with the fix, stacked on this branch — #52691 has your commit as its parent, so please merge this one first and #52691 collapses to a single follow-up commit.

Nothing to change here on your side; #52684 is correct as it stands.

xiaofan-luan:
Rebased your branch onto master (now `6a2a73f2c9`) so it picks up #52694, which fixed the repo-wide `ci-v2/build-ut-cov` failure — that red check was master's `HybridScalarIndex.cpp` not being clang-format clean, not anything in this PR. Your commit is unchanged, only the base moved. CI should be re-running now.

xiaofan-luan:
rerun ut

Issue #52686, opened 2026-08-19
[Bug]: GetReplicateConfiguration blocks until deadline on a cluster with no replicate configuration


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (77ebd82e0e), 2.6 (hotfix-2.6.22 dcd4a60160); reported in the field on a 2.6.17 build
- Deployment mode: cluster, streaming service enabled
- MQ type: Kafka
- SDK version: pymilvus 2.6.16
- OS: Linux
```

### Current Behavior

`GetReplicateConfiguration` never returns on a cluster that has never been given a
replicate configuration. The call blocks until the client deadline and comes back as
`DEADLINE_EXCEEDED`, which is indistinguishable from an unreachable or unhealthy
cluster. Every freshly installed cluster is in that state, so the first thing an
operator does while setting up cross-cluster replication — read the current topology —
appears to fail.

The same applies to `GetReplicateInfo`, which reads the configuration on the way in.

### Root cause

A cluster with no replication has no configuration to report, and that is expressed as
a nil value all the way down:

1. `internal/streamingcoord/server/balancer/channel/manager.go` — when `cm.replicateConfig == nil`
   the assignment is pushed with a nil `ReplicateConfiguration`.
2. `pkg/util/replicateutil/config_helper.go` — `NewConfigHelper` returns `nil, nil` for a nil config.
3. `internal/streamingcoord/client/assignment/discoverer.go` — that nil helper is stored on the watcher.
4. `internal/streamingcoord/client/assignment/watcher.go` — the wait condition is

```go
for (w.lastVersionedAssignment.Version.Global == -1 && w.lastVersionedAssignment.Version.Local == -1) ||
    w.lastVersionedAssignment.ReplicateConfigHelper == nil {
    if err := w.cond.Wait(ctx); err != nil {
        return nil, err
    }
}
```

The second term treats "the cluster told us it has no configuration" as "we have not
heard from the cluster yet". It can never be satisfied on such a cluster, so the call
runs to the deadline. The sibling accessors on the same watcher —
`GetLatestStreamingVersion` and `GetLatestDiscover` — wait only on the version; only
this one carries the extra nil term.

Sending an empty configuration from the balancer instead does not work:
`NewConfigHelper` rejects a configuration with no clusters (`primary count is not 1`),
so `MustNewConfigHelper` would panic in every client. A nil helper is the representation
[...]

PR #52687, opened 2026-08-19
fix: return the replicate configuration of an unconfigured cluster instead of blocking

czs007:
issue: #52686

## Problem

`GetReplicateConfiguration` never returns on a cluster that has never been given a
replicate configuration — it blocks until the caller's deadline and surfaces as
`DEADLINE_EXCEEDED`, which reads as an unhealthy cluster. Every freshly installed
cluster is in that state, so reading the topology before configuring replication
appears to fail.

A cluster with no replication has no configuration to report, and that is nil all the
way down: the balancer pushes a nil `ReplicateConfiguration` when `cm.replicateConfig`
is nil, `NewConfigHelper` maps a nil config to a nil helper, and the watcher's wait
condition treated that nil helper as "no assignment received yet":

```go
for (…Version.Global == -1 && …Version.Local == -1) ||
    w.lastVersionedAssignment.ReplicateConfigHelper == nil {
```

The second term can never be satisfied there. The sibling accessors on the same watcher,
`GetLatestStreamingVersion` and `GetLatestDiscover`, wait only on the version.

Sending an empty configuration from the balancer instead is not viable: `NewConfigHelper`
rejects a configuration with no clusters (`primary count is not 1`), so
`MustNewConfigHelper` would panic in every client. A nil helper *is* the representation
of "no replication configured".

#47517 made this deterministic rather than intermittent: the public API now always takes
the fresh-read path, which builds a throwaway watcher and waits for one assignment push,
so a genuinely fresh cluster fails every time while a previously-configured one still
answers from its long-lived watcher.

## Change

- `watcher.GetLatestReplicateConfiguration` waits only until an assignment has been
  received, and returns the possibly-nil helper.
- `ConfigHelper.GetReplicateConfiguration` and `GetCluster` are nil-receiver safe, so
  callers can use the result directly.
- `overwriteReplicateMessage` rejects a replicate message explicitly when the cluster has
  no configuration — previously unreachable, because the lookup blocked first.

## Tests

- `blocks_when_version_set_but_config_nil` asserted the old behaviour (deadline). It is
  replaced by `returns_nil_when_version_set_but_config_nil`, which asserts a prompt nil
  result and that the read accessors stay usable on it.
- New `blocks_until_first_assignment_arrives` keeps the genuine "not known yet" case
  covered: with no assignment at all the call still blocks, and returns once one arrives.
- `TestWatcher_GetLatestReplicateConfiguration` and `TestGetReplicateConfiguration_FreshRead`
  pass in full.

### DIFF
--- a/internal/distributed/streaming/replicate_service.go
+++ b/internal/distributed/streaming/replicate_service.go
@@ -120,6 +120,9 @@ func (s replicateService) overwriteReplicateMessage(ctx context.Context, msg mes
 	if err != nil {
 		return nil, err
 	}
+	if cfg == nil {
+		return nil, status.NewReplicateViolation("cluster has no replicate configuration, cannot receive replicate message")
+	}
 
 	// Get target vchannel on current cluster that should be written to
 	currentCluster := cfg.GetCluster(s.clusterID)
--- a/internal/streamingcoord/client/assignment/assignment_test.go
+++ b/internal/streamingcoord/client/assignment/assignment_test.go
@@ -378,21 +378,59 @@ func TestWatcher_GetLatestReplicateConfiguration(t *testing.T) {
 		assert.NoError(t, <-errCh)
 	})
 
-	t.Run("blocks_when_version_set_but_config_nil", func(t *testing.T) {
+	t.Run("returns_nil_when_version_set_but_config_nil", func(t *testing.T) {
+		// A cluster that has never been given a replicate configuration reports a
+		// nil config in its assignment. That is an answer, not a missing update:
+		// the call must return promptly with a nil helper instead of blocking
+		// until the deadline.
 		w := newWatcher()
 
-		// Update with valid version but nil config
 		w.Update(types.VersionedStreamingNodeAssignments{
 			Version:               typeutil.VersionInt64Pair{Global: 1, Local: 1},
 			Assignments:           map[int64]types.StreamingNodeAssignment{},
 			ReplicateConfigHelper: nil,
 		})
 
-		ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 		defer cancel()
 
-		_, err := w.GetLatestReplicateConfiguration(ctx)
-		assert.ErrorIs(t, err, context.DeadlineExceeded)
+		config, err := w.GetLatestReplicateConfiguration(ctx)
+		assert.NoError(t, err)
+		assert.Nil(t, config)
+		// The nil helper stays usable for the read accessors.
+		assert.Nil(t, config.GetReplicateConfiguration())
+		assert.Nil(t, config.GetCluster("any"))
+	})
+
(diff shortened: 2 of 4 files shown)


bigsheeper:

Verification left two low-severity items attributable to this PR — reconnect churn on an unconfigured secondary plus a missing test for the new nil-config guard — and surfaced one real but pre-existing concurrency hazard that is listed separately below.

## Low

- **`internal/cdc/replication/replicatestream/replicate_stream_client_impl.go:127`** — On a secondary that has not yet been configured, the new `ReplicateViolation` guard turns a blocking wait into a tight reconnect loop: the client resets its exponential backoff after every successful `CreateReplicateStream`, so the retry interval never leaves the 100 ms initial value. When A has run `UpdateReplicateConfiguration` (A=primary, B=secondary target) but B has not yet run the same call, and A's CDC opens a stream per replicate channel → each channel cycles connect → `ReplicateViolation` → stream teardown → reconnect at roughly 10x/second, emitting a Warn on both sides, oscillating the `OnConnect`/`OnDisconnect` metrics, and re-sending all pending messages via `SeekToHead` on every round, continuing indefinitely until B is configured. No messages are lost, and this matches the existing behaviour of the `primary cluster cannot receive replicate message` path, so the practical impact is operational noise. Suggestion: let the backoff grow across attempts on this path (reset only after a stream has stayed healthy), and correct the PR description — "previously unreachable" reads as if the branch were defensive dead code, whereas it is reached in a normal A-then-B bootstrap; something like "this state previously blocked on the configuration query and is now rejected explicitly" is accurate. *(surfaced during verification)*

- **`internal/distributed/streaming/replicate_service.go:123`** — The `cfg == nil` guard added by this PR has no test coverage. Every stub of `GetReplicateConfiguration` in `replicate_service_test.go` returns a non-nil helper built by `replicateutil.MustNewConfigHelper(...)`, except `TestReplicateServiceGetConfigError` (`replicate_service_test.go:1326`), which returns `(nil, error)` and therefore never reaches the guard. Suggestion: add a sub-case stubbing `Return(nil, nil)` and assert that the returned error is a `ReplicateViolation`. *(surfaced during verification)*

- **`internal/streamingcoord/server/balancer/channel/manager.go:731`** — `applyAssignments` reads `cm.replicateConfig` (:731-733) and `cm.streamingVersion` (:735) *after* releasing the lock at `:728`, while `UpdateReplicateConfiguration` writes `cm.replicateConfig` under the lock at `:644`; if an `AssignmentDiscover` push goroutine is between `:728` and `:735` while the ack-callback goroutine performs that assignment → an unsynchronized concurrent read/write of the pointer field that `-race` reports as a data race, and with no happens-before edge the push may carry a stale or nil `ReplicateConfiguration` to the client, which the watcher then caches. The fix is small — move `:730-733` back inside the `:718-728` critical section. This is unrelated to the PR under review: the PR touches only `internal/distributed/streaming/replicate_service.go`, `internal/streamingcoord/client/assignment/watcher.go` (and its tests), and `pkg/util/replicateutil/config_helper.go`, leaving `channel/manager.go` untouched, so the lock-free read predates it. No tracking issue yet. *(surfaced during verification)*

PR #52688, opened 2026-08-19
fix: stop task.Wait from parking a goroutine on an unbounded task

xiaofan-luan:
issue: #52257 related

Follow-up to the review comments left on #52478 (https://github.com/milvus-io/milvus/pull/52478#issuecomment-5343867171). Thanks @AyushKashyapII — all three hold up, none of them were introduced by that PR.

### 1. `Wait` parked a goroutine on a task with no lifetime bound

`task.Wait(ctx, tasks...)` raced a helper goroutine against the caller's ctx. When ctx expired first the caller returned on time, but the helper stayed blocked in `Task.Wait()` until the task actually finished — and that wait has no upper bound:

- `Executor.Execute` rejects a step when the channel / non-channel admission cap is full and returns **before** `task.ActivateDeadline(step)`, so a queued task has no deadline armed;
- `taskScheduler.checkStale` only checks replica / node / RO state — there is no time-based staleness check.

So on a healthy cluster with a saturated executor, a queued task can sit indefinitely, and the helper goroutine with it. Live callers are the two LoadBalance paths in `handlers.go` (`balanceSegments`, `balanceChannels`).

Fix: give `Task.Wait` a ctx and select on it, so the package-level `Wait` walks the tasks on the caller's goroutine and needs no helper at all. A finished task is checked non-blockingly first, so a task that completes as ctx expires still reports its real outcome instead of losing a coin flip to `DeadlineExceeded`.

### 2. `NewChannelTask` error message

Reported `"all actions must operate the same segment"` while validating channel names — copy-paste from `NewSegmentTask`, present since #22745.

### 3. `TaskStatusCreated`

Defined but never assigned or read anywhere in the tree; tasks start at `TaskStatusStarted`. Removed.

### Tests

`TestWaitDoesNotOutliveContext` fires 64 `Wait` calls against tasks that are never finished and asserts the goroutine count returns to baseline. Before the fix it reports exactly 64 leaked goroutines; after, it settles immediately. `TestWaitPrefersFinishedTaskOverExpiredContext` pins the tie-break.

`go test -race ./internal/querycoordv2/task/...` is green. The pre-existing local failures in `internal/querycoordv2` and `internal/querycoordv2/checkers` reproduce identically on a clean `origin/master` checkout and are unrelated.

/kind bug

https://claude.ai/code/session_01AyYujrT7LLZ3waGhqEGMvB

### DIFF
--- a/internal/querycoordv2/task/scheduler.go
+++ b/internal/querycoordv2/task/scheduler.go
@@ -1275,7 +1275,17 @@ func (scheduler *taskScheduler) remove(task Task) {
 		}
 	}
 
-	task.Cancel(nil)
+	// A task that is still Started when removed did not finish on its own: it
+	// was dropped by RemoveByNode (a node went down) or Stop (shutdown). Cancel
+	// it with a real error so a caller blocked in Wait can distinguish this
+	// from a successful completion, whose err is nil. Terminal tasks keep
+	// Cancel(nil): for already-failed/canceled tasks the CAS makes it a no-op,
+	// and for a succeeded task nil preserves the success result.
+	if task.Status() == TaskStatusStarted {
+		task.Cancel(merr.WrapErrServiceUnavailable("task removed before finish"))
+	} else {
+		task.Cancel(nil)
+	}
 	_, ok := scheduler.tasks.GetAndRemove(task.ID())
 	scheduler.waitQueue.Remove(task)
 	scheduler.processQueue.Remove(task)
--- a/internal/querycoordv2/task/task.go
+++ b/internal/querycoordv2/task/task.go
@@ -43,7 +43,6 @@ type (
 )
 
 const (
-	TaskStatusCreated   = "created"
 	TaskStatusStarted   = "started"
 	TaskStatusSucceeded = "succeeded"
 	TaskStatusCanceled  = "canceled"
@@ -101,7 +100,13 @@ type Task interface {
 	// fail the task as we encounter some error so be unable to continue,
 	// this error will be recorded for response to user requests
 	Fail(err error)
-	Wait() error
+	// Wait blocks until the task has finished (succeeded, failed or been
+	// canceled) and returns its error, or until ctx is done, whichever comes
+	// first. ctx is what bounds the caller: a task that keeps getting bounced
+	// by the executor's admission cap never arms a deadline (see
+	// ActivateDeadline), so it carries no lifetime guarantee of its own while
+	// it is still queued.
+	Wait(ctx context.Context) error
 	Actions() []Action
 	Step() int
 	StepUp() int
@@ -354,9 +359,30 @@ func (task *baseTask) Fail(err error) {
 	}
 }
 
-func (task *baseTask) Wait() error {
-	<-task.doneCh
-	return task.err
+func (task *baseTask) Wait(ctx context.Context) error {
+	// Task completion is signaled solely by closing doneCh, never by the err
+	// value: a succeeded task is finished with a nil err, so err is not a
+	// reliable "finished" predicate. Reading task.err is only safe after a
+	// receive from doneCh because Fail/Cancel write task.err before closing
+	// doneCh, and that receive synchronizes-with the close.
+	select {
+	case <-task.doneCh:
+		// Task finished (succeeded, failed or canceled); report its real
+		// outcome.
+		return task.err
+	case <-ctx.Done():
+		// ctx expired first, but the task may have finished in the same
+		// instant. When both doneCh and ctx.Done() are ready at once the
+		// runtime picks a case at random, so recheck doneCh here: once the
+		// task has finished it reports its real outcome instead of a
+		// spurious context error, keeping the tie-break deterministic.
+		select {
+		case <-task.doneCh:
+			return task.err
+		default:
+			return ctx.Err()
+		}
+	}
 }
 
 func (task *baseTask) Actions() []Action {
@@ -523,7 +549,7 @@ func NewChannelTask(ctx context.Context,
 		if channel == "" {
 			channel = channelAction.ChannelName()
 		} else if channel != channelAction.ChannelName() {
-			return nil, errors.WithStack(merr.WrapErrParameterInvalid(channel, channelAction.ChannelName(), "all actions must operate the same segment"))
+			return nil, errors.WithStack(merr.WrapErrParameterInvalid(channel, channelAction.ChannelName(), "all actions must operate the same channel"))
 		}
 	}
 
--- a/internal/querycoordv2/task/task_test.go
+++ b/internal/querycoordv2/task/task_test.go
@@ -2263,6 +2263,55 @@ func (suite *TaskSuite) TestRemoveTaskWithError() {
 	suite.False(suite.nodeMgr.IsResourceExhausted(nodeID))
 }
 
+// TestRemoveStartedTaskReturnsError pins that a task dropped while still
+// Started (node down via RemoveByNode, or shutdown via Stop) is canceled with a
+// real error, so a synchronous Wait caller can distinguish it from a
+// successful completion whose err is nil.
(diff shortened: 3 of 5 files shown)


czs007:

One in-PR item and one pre-existing hazard survived verification: the `Wait` doc comment added by this PR contradicts the paragraph directly above it, and a separate, older `scheduler.remove()` behavior is recorded below for follow-up.

## Low

- **`internal/querycoordv2/task/utils.go:56`** — The first paragraph of the package-level `Wait` doc comment (added in #52478) asserts that "each task already carries its own real deadline (armed on first dispatch, see `Task.ActivateDeadline`) and is guaranteed to finish within it", while the paragraph this PR adds immediately below at lines 62-70 states the opposite and is the stated motivation for the change: a task repeatedly bounced by the executor's admission cap never arms a deadline, and `taskScheduler.checkStale` has no time-based staleness check, so the helper can stay parked indefinitely. The code supports the new paragraph: `executor.go:157-172` returns `false` on a full admission pool *before* reaching `task.ActivateDeadline(step)` at `executor.go:184`, and `checkStale` (`scheduler.go:1372-1404`) tests only replica presence, dist presence and RO nodes — no time dimension. As written, a reader gets both "the task is guaranteed to finish within its own deadline" and "the task may never finish" from the same comment block. Suggestion: in this same pass, delete or rewrite the "is guaranteed to finish within it" clause in the first paragraph so the two paragraphs agree. *(raised by tinswzy)*

- **`internal/querycoordv2/task/scheduler.go:1278`** — `remove()` calls `task.Cancel(nil)` unconditionally, so a task dropped while still `TaskStatusStarted` becomes indistinguishable from a successful one. `remove()` is also the success path's sink: the schedule loop only moves tasks with `task.Status() != TaskStatusStarted` into `toRemove` (`scheduler.go:1051-1057`), and `preProcess` marks completion with `task.SetStatus(TaskStatusSucceeded)` without closing `doneCh` (`scheduler.go:1163-1164`) — `doneCh` is closed only by `Cancel`/`Fail` (`task.go:336`/`353`). For a `Wait` caller, "`doneCh` closed and `err == nil`" is therefore the sole success signal. But `RemoveByNode` (`scheduler.go:1219-1234`) and `Stop` (`scheduler.go:628-638`) call `remove()` on every ranged task with no status filter, and `Cancel`'s CAS (`task.go:329`) succeeds on a still-running task. Failure: during a synchronous `LoadBalance` RPC (`services.go:682`, `sync=true`) a source or target QueryNode goes down and `server.go:786` `handleNodeDown` → `RemoveByNode` cancels the still-`Started` balance task with a nil error (the same applies to graceful shutdown via `Stop`) → `task.Wait` returns nil, the `err != nil` check at `handlers.go:174` does not fire, and `LoadBalance` answers `merr.Success()` even though no segment was ever moved. The fix belongs in `remove()` — pass a real error (e.g. `merr.WrapErrServiceInternal("task removed before finish")`) when `task.Status() == TaskStatusStarted`, keeping `Cancel(nil)` for terminal statuses; `Wait` has no information left to distinguish the two cases. Unrelated to this PR: this PR touches only `task/{task.go,utils.go,utils_test.go}`, `scheduler.go` is byte-identical to its parent commit at line 1278, and `Cancel`/`Fail` (`task.go:328-359`) are unchanged — only a `ctx` parameter was added to `Wait`, leaving the origin and meaning of `err` untouched. No tracking issue yet. *(raised by tinswzy, bigsheeper)*

czs007:

Adversarial review surfaced two test-quality issues in the new test cases added by this PR; the production changes to `Task.Wait`, the package-level `Wait` helper, and `scheduler.remove` verified clean.

## Medium

**`internal/querycoordv2/task/utils_test.go:956` — `TestWaitRechecksDoneWhenContextDone` is a reproducible flake**

The test assumes the main goroutine runs `cancel()` (`:984`) and `task.Fail(taskErr)` (`:985`) back-to-back without the waiter being rescheduled in between. `GOMAXPROCS(1)` removes parallelism but not preemption: once `cancel()` closes `ctx.Done()`, the waiter is placed in the P's `runnext`, and any preemption point between those two lines (a GC stop-the-world in a function prologue reschedules main via `goschedImpl`, after which `runqget` takes `runnext` first) lets the waiter resume before `doneCh` is closed. It is already committed to the `ctx.Done()` case — Go's select locks the case via `gp.param` on wake and does not re-evaluate — so the inner recheck (`task.go:379-383`) sees `doneCh` still open and returns `ctx.Err()`.

Failure scenario: a GC stop-the-world (or any preemption point) lands between `utils_test.go:984` and `:985` → `task.Wait` returns `context.Canceled`, the assertion at `:987` fails with `expected: "load segment failed" / in chain: "context canceled"`, and the whole `querycoordv2/task` package goes red on a false positive unrelated to the code under test. Measured on the PR branch: 15 failures in 200 runs (~7.5%) under default `GOGC`; `GOGC=1 -count=20` reproduces on the first attempt. A full `go test -count=1 ./internal/querycoordv2/task/` passes, confirming this is probabilistic rather than deterministic.

Suggestion: stop depending on the scheduler interleaving. Keep the parked-select variant, but tally outcomes across the 100 iterations and require at least one `taskErr` — failing only if *every* iteration returned the context error. That still catches deletion of the recheck branch while tolerating an occasional preempted iteration.

Secondary, same block: the comments at `:950-955` and `:979-982` state the waiter reaches "the second select (both cases open)" and that the runtime picks randomly between them. The waiter actually parks in the *first* select (on `doneCh` and `ctx.Done()` together); the second select is `doneCh` plus `default`, so no random choice exists. The described mechanism does not match the code. *(raised by tinswzy)*

## Low

**`internal/querycoordv2/task/task_test.go:2270` — `TestRemoveStartedTaskReturnsError` bypasses both production paths its motivation cites**

The test drives the new `Started` branch (`scheduler.go:1284-1288`) by calling `scheduler.remove(task)` directly at `:2287` / `:2307`, so neither path named in the comment at `scheduler.go:1278-1283` is exercised: `RemoveByNode` (`scheduler.go:1219-1235`, which additionally filters through `isRelated` and holds `scheduleMu`) and `Stop` (`scheduler.go:623-637`, which does not hold `scheduleMu`). Those two are genuinely the only entry points — `processQueue` only collects tasks with `task.Status() != TaskStatusStarted` into `toRemove` (`scheduler.go:1055`), and the `preAdd` replacements (`:775`/`:806`/`:840`) plus `tryPromoteAll` (`:868`) all call `Cancel(err)` before `remove`, arriving already `Canceled`. So the branch's reachability rests entirely on `RemoveByNode` / `Stop`, and no test touches either.

Current behavior is correct and the test passes. Suggestion: add a case that goes through `RemoveByNode(nodeID)` to pin the reachability the comment asserts, so a future refactor of `isRelated` (`scheduler.go:1098-1106`) or the queue traversal cannot silently orphan this branch. *(raised by czs007)*

xiaofan-luan:
Thanks for the review — all feedback is addressed in `ca758ea`:

- **Bot P2 (recheck in `ctx.Done` branch)**: `Wait` now rechecks `doneCh` in the `ctx.Done()` case; thread resolved.
- **Low (`Wait` doc comment contradiction)**: removed the stale "guaranteed to finish within it" clause.
- **Medium (`TestWaitRechecksDoneWhenContextDone` flake)**: now tallies outcomes across the 100 iterations and only requires at least one `taskErr`, so a preempted iteration no longer fails the test; comments corrected to describe the actual select/park mechanism.
- **Low (`TestRemoveStartedTaskReturnsError` bypasses real paths)**: now drives through `RemoveByNode(nodeID)` instead of calling `scheduler.remove` directly.
- **Pre-existing hazard (`scheduler.remove` `Cancel(nil)`)**: a still-`Started` task removed early is now canceled with a real error (`ErrServiceUnavailable`) so `Wait` reports it instead of success.

CI is green: all stages pass, UT-CPP-Cov 8755/8755, Go diff coverage 100%.

liliu-z (commented):
/lgtm
/approve

Issue #52681, opened 2026-08-19
[Bug]: TIMESTAMPTZ field-to-field comparison throws "unsupported left datatype"


daviddallakyan2005:
### Is there an existing issue for this?
I searched for TIMESTAMPTZ compare and "unsupported left datatype" and found none.

### Environment
Reproduced on milvus-io/milvus master at 77ebd82e0eb00ae390a06434822afc60fc0a6788. The CompareExpr dispatcher on current master still has no TIMESTAMPTZ case. This is a C++ execution bug, not a deployment issue.

### Current Behavior
Comparing two TIMESTAMPTZ columns in a filter expression, for example `ts_a > ts_b`, fails at execution with:

unsupported left datatype:TIMESTAMPTZ of compare expr

The parser accepts the expression, so this surfaces as a query-time error rather than a validation error. Whether it fails depends on segment and index state rather than on the query. Growing segments always fail. Sealed segments fail when left and right chunk counts and boundaries align. The offset-input path always fails. It succeeds when either side carries a scalar index, because that routes to a different execution path.

Comparing a TIMESTAMPTZ column against a literal and `ts_a IN (...)` both work correctly.

### Expected Behavior
Either `ts_a > ts_b` executes and returns rows ordered by the underlying int64 value, consistently across growing and sealed segments, or the parser rejects field-to-field TIMESTAMPTZ comparison up front with a clear plan error, the way it already rejects ARRAY and JSON field-to-field comparison.

### Steps To Reproduce
1. Create a collection with two TIMESTAMPTZ fields, ts_a and ts_b, with no scalar index on either.
2. Insert rows and do not flush, so the data stays in a growing segment.
3. Query with filter `ts_a > ts_b`.
4. The query fails with `unsupported left datatype:TIMESTAMPTZ of compare expr`.

### Milvus Log
N/A. Confirmed from the C++ dispatcher and the existing hybrid accessor.

### Anything else?
Three parts.

1. `internal/parser/planparserv2/utils.go`, `handleCompare` (around line 460) explicitly rejects ARRAY field-to-field (around 484) and JSON field-to-field (around 488), then falls through to `Expr_CompareExpr` (around 493). TIMESTAMPTZ has no guard, so it reaches CompareExpr. This is not the TimestamptzArithCompareExpr path. That node is only built when an INTERVAL is present, see `parser_visitor.go`.

2. `internal/core/src/exec/expression/CompareExpr.cpp`:
[...]

PR #52689, opened 2026-08-19
fix: support TIMESTAMPTZ in field-to-field compare expressions

daviddallakyan2005:
issue: #52681

Field to field TIMESTAMPTZ compare was rejected by the parser and threw in the CompareExpr fast path. TIMESTAMPTZ is already int64-backed in SegmentChunkReader, TermExpr, and UnaryExpr. This allows same type field compare only. Mixed TIMESTAMPTZ vs INT64 stays unsupported.

## Test

Fail without the C++ cases:

```
make run-test-cpp filter=TimestamptzCompareCorrectnessTest*
[  FAILED  ] 4 tests
unsupported left datatype:TIMESTAMPTZ of compare expr
```

Pass with the fix:

```
LOCAL_STORAGE_SIZE=1 make run-test-cpp filter=TimestamptzCompareCorrectnessTest*
[  PASSED  ] 4 tests.

LOCAL_STORAGE_SIZE=1 go test -tags dynamic,test -count=1 ./internal/parser/planparserv2
ok  github.com/milvus-io/milvus/internal/parser/planparserv2
```

/cc @xiaofan-luan


internal/core/unittest/test_timestamptz_compare.cpp
```diff
@@ -0,0 +1,190 @@
...
+        SegcoreConfig config = SegcoreConfig::default_config();
+        config.set_chunk_rows(8);
+        growing_ = CreateGrowingSegment(schema_, empty_index_meta, 1, config);
+        growing_->PreInsert(N);
+        growing_->Insert(0,
+                         N,
+                         dataset_->row_ids_.data(),
+                         dataset_->timestamps_.data(),
+                         dataset_->raw_);
+        sealed_ = CreateSealedWithFieldDataLoaded(schema_, *dataset_);
```
yhmo, line 102: internal/core/unittest/test_timestamptz_compare.cpp line:100
Low ---- Both segments here are built with empty_index_meta, so these tests exercise only the both-data fast path (no scalar index). The parser change in this PR is what first makes TSTZ field-to-field compare reach the hybrid (indexed) path, yet no test loads a TSTZ scalar index to verify it. If the int64-based ScalarIndex reverse-lookup or accessor ever regresses for TSTZ, CI stays green while collections with a scalar index on either timestamp field fail at runtime. Consider adding a case that loads a TSTZ scalar index on one side (mirroring LoadTimestamptzIndex in test_timestamptz_arith_compare.cpp).

internal/parser/planparserv2/utils.go
```diff
@@ -525,6 +525,8 @@ func canBeComparedDataType(left, right schemapb.DataType) bool {
 		return typeutil.IsStringType(right) || typeutil.IsJSONType(right)
 	case schemapb.DataType_JSON:
 		return true
+	case schemapb.DataType_Timestamptz:
```
yhmo, line 528: internal/parser/planparserv2/utils.go line:528
Low ---- The same-type-only restriction here is deliberate policy (the C++ dispatcher would actually execute mixed TSTZ/INT64 since both dispatch to int64_t), but the code carries no comment explaining why. A future maintainer seeing TSTZ excluded from the arithmetic case could "align" it with the INT64 case and silently enable mixed timestamp/raw-int64 comparisons. A one-line comment at this case (matching the rationale in the PR body) would prevent that drift.

internal/core/unittest/test_timestamptz_compare.cpp
daviddallakyan2005: SealedScalarIndexOnLeftGreaterThan loads a STL_SORT index on ts_a and asserts the same GreaterThan bits as the raw path. SealedOffsetInputScalarIndexOnLeftGreaterThan covers the offset input hybrid path.

internal/parser/planparserv2/utils.go
daviddallakyan2005: I added a comment on the Timestamptz case. Mixed TIMESTAMPTZ vs INT64 stays unsupported at parse time even though both dispatch to int64_t.

czs007:

**Verified:**
- `internal/core/src/exec/expression/CompareExpr.cpp:342` — TIMESTAMPTZ falls through to the existing `int64_t` instantiation for both left- and right-type dispatch, which is type-correct because TIMESTAMPTZ is int64-backed storage read via `chunk_data<int64_t>` (`Types.h:87`, `ConcurrentVector.cpp:135`, `ChunkedColumn.h:484`), and it adds no new template instantiations.
- `internal/core/src/exec/expression/SegmentChunkReader.cpp:263` / `:390` — the hybrid/index path already mapped `TIMESTAMPTZ → int64_t` before this PR, so no change was needed there, and the new sealed scalar-index cases genuinely route through it rather than through the both-data fast path.
- `internal/core/src/index/IndexFactory.cpp:980` — TIMESTAMPTZ builds through `CreatePrimitiveScalarIndex<int64_t>`, so the `dynamic_cast<const index::ScalarIndex<int64_t>*>` at `SegmentChunkReader.cpp:290` hits for STL_SORT/INVERTED instead of silently degrading to a raw-data path.
- `utils.go:528-531` — the new parser gate is a one-way tightening: same-type-only via `IsTimestamptzType(right)`, and the reverse direction (`Int64 vs Timestamptz`) is already rejected because `IsArithmetic` excludes Timestamptz; JSON and ARRAY field-to-field comparisons remain blocked at `utils.go:485` / `:490`.
- `CompareExpr.h:388-400` and `CompareExpr.cpp:163-172` — null semantics are explicit (`res=false; valid_res=false` when either side is invalid) rather than mask-after-SIMD, matching the test oracle; `ASSERT_GT(nulls,0)` / `ASSERT_LT(nulls,N)` keep the assertions from degenerating to always-true.
- `parser_visitor.go:2890` / `:2964` — field-vs-literal TIMESTAMPTZ comparisons use separate grammar rules and never reach `canBeComparedDataType`, so the new case introduces no regression on that path.

## Nitpick — non-blocking

- `internal/core/unittest/test_timestamptz_compare.cpp:125` — All six new `TEST_F` cases assert only `GreaterThan`: both call sites of the `MakeTstzFieldCompare` helper (L124-125 and L163-164) hard-code `proto::plan::OpType::GreaterThan`, and the oracle `ExpectedGreaterThan` (L48-57) hard-codes `left > right`, so adding an operator requires editing both the helper and the oracle. Risk is genuinely low — this PR only touches the *type* dispatch switch, while operator dispatch is a separate type-independent switch (`CompareExpr.cpp:408-472` and `:295-318`) whose template instantiations are identical to the existing INT64-vs-INT64 ones, so no operator can fail for TIMESTAMPTZ alone. Suggestion: parametrize the assertion helper over all six operators (with the oracle selected per-op via `std::function` or a small switch) so the comparator wiring symmetry is locked in by the test. (raised by xiaocai2333)

Issue #52690, opened 2026-08-19
[Bug]: proxy collection metrics are only partially cleaned on drop because cleanup enumerates label values


xiaofan-luan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

#52682 / #52684 cover four metrics that collection cleanup forgot entirely. This issue is about a different failure mode in the same function: metrics that **are** listed in `CleanupProxyCollectionMetrics` but are deleted by enumerating label values, so they only get partially cleaned.

### Environment

```markdown
- Milvus version: master
- Deployment mode(standalone or cluster): any
- MQ type(rocksmq, pulsar or kafka): any
- SDK version(e.g. pymilvus v2.0.0rc2): n/a
- OS(Ubuntu or CentOS): any
- CPU/Memory: n/a
- GPU: n/a
- Others: the leak is in in-process metrics cleanup, not deployment specific
```

### Current Behavior

`CleanupProxyCollectionMetrics` (pkg/metrics/proxy_metrics.go) deletes part of its series with `Delete()` and a fully specified label set, which forces it to enumerate every `msg_type` / `query_type` value the metric can hold. Those enumerations have gone stale, so dropping a collection leaves these series behind:

| metric | emitted with | cleanup enumerates | leaks |
|---|---|---|---|
| `ProxyCollectionSQLatency` | search, **hybrid_search** (internal/proxy/impl.go:3345), query / upsert_query | search, query, upsert_query | hybrid_search |
| `ProxyCollectionMutationLatency` | delete (impl.go:2704), **upsert** (impl.go:2850) | insert, delete | upsert (and `insert` is enumerated but never emitted) |
| `ProxyReceiveBytes` | insert, delete, upsert, search, **hybrid_search** (impl.go:3218), query | everything but hybrid_search | hybrid_search |
| `ProxyScannedRemoteMB` | search, **hybrid_search** (impl.go:3353), query, upsert, delete | everything but hybrid_search | hybrid_search |
| `ProxyScannedTotalMB` | search, **hybrid_search** (impl.go:3360), query, upsert, delete | everything but hybrid_search | hybrid_search |

`ProxyReceivedNQ`, `ProxyRetrySearchCount`, `ProxyRetrySearchResultInsufficientCount` and `ProxyRecallSearchCount` line up today, but they are one new `query_type` away from the same bug.

Two related gaps in the same code:

[...]

PR #52691, opened 2026-08-19
fix: drop proxy collection metrics by partial match instead of enumerating labels

xiaofan-luan:
issue: #52690

Stacked on #52684 by @daviddallakyan2005 — that PR is the parent commit here, so this diff will collapse to a single commit once it merges. Please merge #52684 first.

`CleanupProxyCollectionMetrics` deleted most of its series with `Delete()` and a fully specified label set, so every `msg_type` / `query_type` value a metric can hold had to be enumerated by hand. Those enumerations went stale:

| metric | leaked value |
|---|---|
| `ProxyCollectionSQLatency` | hybrid_search |
| `ProxyCollectionMutationLatency` | upsert (and `insert` was enumerated but is never emitted) |
| `ProxyReceiveBytes` | hybrid_search |
| `ProxyScannedRemoteMB` | hybrid_search |
| `ProxyScannedTotalMB` | hybrid_search |

Cleanup now goes through `DeletePartialMatch` on `(node_id, db_name[, collection_name])` for every collection-scoped proxy metric. The list lives once in `proxyCollectionScopedMetrics()` and is shared by the collection and database cleanup paths, which also fixes `CleanupProxyDBMetrics` leaving 11 of 18 collection-scoped metrics behind on drop database. Net −150 lines in that file.

`search_sparse_num_non_zeros` and `function_udf_call_latency` had no `db_name` label, so cleaning them by collection name alone wipes a same-named collection in a different database. Both now carry `db_name` (`FunctionExecutor` picks it up from `extraInfo.DBName`). Neither metric is referenced by any dashboard under `deployments/`.

## Test

A guard test does not catch this class of bug by itself: observing one label combination per metric only proves the enumeration covers the value the test happened to pick. Two changes fix that:

- `observeProxyCollection` now loops over every `query_type` / `msg_type` the proxy actually emits.
- `TestCollectionScopedMetricsAreComplete` parses the metric declarations with `go/parser` and fails when a `*Vec` labeled `db_name`+`collection_name` (proxy) or `collection_id` (query node) is not referenced by the corresponding cleanup function.

Mutation-checked in both directions. Dropping `ProxyReceiveBytes` from the list:

```
ProxyReceiveBytes is labeled with [databaseLabelName collectionName] but is never cleaned up in proxyCollectionScopedMetrics
```

Dropping `QueryNodeSearchFTSNumTokens.DeletePartialMatch(labels)`:

```
QueryNodeSearchFTSNumTokens is labeled with [collectionIDLabelName] but is never cleaned up in CleanupQueryNodeCollectionMetrics
```

Before the production change, with the label values `internal/proxy/impl.go` emits:

```
LEAKED after cleanup: [milvus_proxy_collection_mutation_latency
                       milvus_proxy_collection_sq_latency
                       milvus_proxy_receive_bytes_count
                       milvus_proxy_scanned_remote_mb
                       milvus_proxy_scanned_total_mb]
```

After:

```
$ go test ./metrics/ -count=1 -timeout 300s
ok  	github.com/milvus-io/milvus/pkg/v3/metrics	1.236s
$ golangci-lint run ./metrics/...
0 issues.
```

`internal/util/function/embedding` executor tests pass, exercising the new 6-label `ProxyFunctionlatency` call. The two `internal/proxy/task_search.go` call sites type-check locally but I could not run that package's tests on macOS (no local C++ core build), so they rely on CI.

### DIFF
--- a/internal/proxy/task_search.go
+++ b/internal/proxy/task_search.go
@@ -690,7 +690,7 @@ func (t *searchTask) initAdvancedSearchRequest(ctx context.Context) error {
 			return err
 		}
 		if typeutil.IsFieldSparseFloatVector(t.schema.CollectionSchema, internalSubReq.FieldId) {
-			metrics.ProxySearchSparseNumNonZeros.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), t.collectionName, metrics.HybridSearchLabel, strconv.FormatInt(internalSubReq.FieldId, 10)).Observe(float64(typeutil.EstimateSparseVectorNNZFromPlaceholderGroup(internalSubReq.PlaceholderGroup, int(internalSubReq.GetNq()))))
+			metrics.ProxySearchSparseNumNonZeros.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), t.request.GetDbName(), t.collectionName, metrics.HybridSearchLabel, strconv.FormatInt(internalSubReq.FieldId, 10)).Observe(float64(typeutil.EstimateSparseVectorNNZFromPlaceholderGroup(internalSubReq.PlaceholderGroup, int(internalSubReq.GetNq()))))
 		}
 		internalSubReq.PlaceholderGroup = convertedPlaceholder
 		t.SubReqs[index] = internalSubReq
@@ -1043,7 +1043,7 @@ func (t *searchTask) initSearchRequest(ctx context.Context) error {
 	}
 	t.PkFilter = checkSegmentFilter(plan)
 	if typeutil.IsFieldSparseFloatVector(t.schema.CollectionSchema, t.FieldId) {
-		metrics.ProxySearchSparseNumNonZeros.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), t.collectionName, metrics.SearchLabel, strconv.FormatInt(t.FieldId, 10)).Observe(float64(typeutil.EstimateSparseVectorNNZFromPlaceholderGroup(t.request.GetPlaceholderGroup(), int(t.request.GetNq()))))
+		metrics.ProxySearchSparseNumNonZeros.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), t.request.GetDbName(), t.collectionName, metrics.SearchLabel, strconv.FormatInt(t.FieldId, 10)).Observe(float64(typeutil.EstimateSparseVectorNNZFromPlaceholderGroup(t.request.GetPlaceholderGroup(), int(t.request.GetNq()))))
 	}
 	// Convert placeholder group vector type if needed (e.g., fp32 -> fp16/bf16)
 	var placeholderType commonpb.PlaceholderType
--- a/internal/util/function/embedding/function_executor.go
+++ b/internal/util/function/embedding/function_executor.go
@@ -54,6 +54,9 @@ type Runner interface {
 
 type FunctionExecutor struct {
 	runners map[int64]Runner
+	// dbName labels the function latency metric so a dropped collection can be
+	// cleaned per (db, collection); collection names are only unique per db.
+	dbName string
 }
 
 func createFunction(coll *schemapb.CollectionSchema, schema *schemapb.FunctionSchema, extraInfo *models.ModelExtraInfo) (Runner, error) {
@@ -118,6 +121,9 @@ func NewFunctionExecutor(schema *schemapb.CollectionSchema, functions []*schemap
 	executor := &FunctionExecutor{
 		runners: make(map[int64]Runner),
 	}
+	if extraInfo != nil {
+		executor.dbName = extraInfo.DBName
+	}
 	if functions == nil {
 		functions = schema.Functions
 	}
@@ -152,7 +158,7 @@ func (executor *FunctionExecutor) processSingleFunction(ctx context.Context, run
 		return nil, err
 	}
 
-	metrics.ProxyFunctionlatency.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), runner.GetCollectionName(), runner.GetFunctionTypeName(), runner.GetFunctionProvider(), runner.GetFunctionName()).Observe(float64(tr.RecordSpan().Milliseconds()))
+	metrics.ProxyFunctionlatency.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), executor.dbName, runner.GetCollectionName(), runner.GetFunctionTypeName(), runner.GetFunctionProvider(), runner.GetFunctionName()).Observe(float64(tr.RecordSpan().Milliseconds()))
 	tr.CtxElapse(ctx, "function ProcessInsert done")
 	return outputs, nil
 }
@@ -214,7 +220,7 @@ func (executor *FunctionExecutor) processSingleSearch(ctx context.Context, runne
 	if err != nil {
 		return nil, err
 	}
-	metrics.ProxyFunctionlatency.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), runner.GetCollectionName(), runner.GetFunctionTypeName(), runner.GetFunctionProvider(), runner.GetFunctionName()).Observe(float64(tr.RecordSpan().Milliseconds()))
+	metrics.ProxyFunctionlatency.WithLabelValues(strconv.FormatInt(paramtable.GetNodeID(), 10), executor.dbName, runner.GetCollectionName(), runner.GetFunctionTypeName(), runner.GetFunctionProvider(), runner.GetFunctionName()).Observe(float64(tr.RecordSpan().Milliseconds()))
 	tr.CtxElapse(ctx, "function ProcessSearch done")
 	return proto.Marshal(res)
 }
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -171,7 +171,6 @@ const (
 	// model function/UDF labels
 	functionTypeName = "function_type_name"
 	functionProvider = "function_provider"
-	functionName     = "function_name"
 
 	// entities label
 	LoadedLabel         = "loaded"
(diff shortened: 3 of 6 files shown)


xiaofan-luan:
Rebased onto master (`27a3061dd6`) to pick up #52694. #52684 was rebased too, so this still stacks cleanly on it — merge #52684 first and this collapses to the single follow-up commit.

xiaofan-luan:
rerun ut

czs007:

The metrics-cleanup refactor is correct and complete at HEAD — the remaining items are narrowing gaps in the new completeness guard, one db-scoped metric still outside the cleanup sweep, and one pre-existing cardinality leak listed separately below.

## Low

**`pkg/metrics/metrics_test.go:447` — completeness guard skips any vec missing one required label, so the comment it backs overstates its coverage.**
`assertCleanupCoversLabels` early-returns on the first missing required label (`if _, has := labels[required]; !has { return true }`, lines 445-449), and the proxy subtest requires *both* `databaseLabelName` and `collectionName` (lines 404-406). A proxy vec carrying `collectionName` but not `databaseLabelName` — precisely the pre-PR shape of `ProxySearchSparseNumNonZeros` and `ProxyFunctionlatency` that this PR fixes — is silently skipped rather than flagged: it never enters `declared`, so the test stays green. This is latent at HEAD (all 18 `collectionName` vecs in `proxy_metrics.go` also carry `databaseLabelName` and `nodeIDLabelName`, and all 18 are listed in `proxyCollectionScopedMetrics()`), but the comment at `proxy_metrics.go:584-586` — "Keep this list in sync when adding a collection labeled proxy metric -- TestCollectionScopedMetricsAreComplete guards it" — claims broader coverage than the guard delivers.
*Suggestion:* either narrow that comment to state the guard covers `db_name` + `collection_name` metrics only, or make the guard flag a vec that matches ANY required label while missing the others, with an explicit commented allow-list entry for the known exception. *(raised by czs007, xiaocai2333)*

**`pkg/metrics/metrics_test.go:485` — the guard's AST parser only handles `*ast.Ident` label elements and a single file, narrowing the guarantee it advertises.**
`vecVariableLabels` (lines 465-491) records a label only when the slice element is an `*ast.Ident` (`if ident, ok := elt.(*ast.Ident); ok`, line 487), so a vec whose labels are written as string literals resolves to an empty label set and is skipped entirely. Separately, `assertCleanupCoversLabels` parses exactly one named file (`parser.ParseFile(token.NewFileSet(), file, nil, 0)`, line 420), so a collection-scoped vec declared elsewhere is invisible. Both are latent today and verified as such: the only string-literal label list in `proxy_metrics.go` is `MaxInsertRate` at line 412 (`[]string{"node_id", "scope"}`), which carries no collection label; `querynode_metrics.go` has no string-literal label list at all; and no proxy-subsystem vec is declared outside `proxy_metrics.go`.
*Suggestion:* resolve string-literal elements alongside identifiers, and either parse the whole package or document in the guard that its scope is the single named file. *(raised by xiaocai2333, czs007)*

**`pkg/metrics/proxy_metrics.go:610` — `CleanupProxyDBMetrics` still misses `ProxyReportValue`, the one other db-labeled proxy metric** *(provenance unclear — please confirm whether this PR introduces it)*.
The rewritten `CleanupProxyDBMetrics` (lines 610-618) iterates `proxyCollectionScopedMetrics()`, i.e. only the 18 vecs carrying both `db_name` and `collection_name`. `ProxyReportValue` (lines 313-320, labels `{node_id, msg_type, db_name, username}`) is the only other proxy vec carrying `db_name`, and it is never deleted anywhere; it is emitted unconditionally on every successful insert/delete/upsert/search/hybrid-search/query (`internal/proxy/impl.go:2590, 2698, 2836, 2841, 3146, 3382, 3908), so the series exists even when the hook extension returns 0. Failure: a multi-tenant deployment that creates a database, runs any successful insert/search/query, then drops it, repeatedly → each dropped database permanently retains up to (msg_type × username) `milvus_proxy_report_value` series on every proxy that served it, growing scrape payload and TSDB cardinality monotonically with database churn — the same leak class this PR closes for collection-scoped metrics.
*Suggestion:* add a db-only `DeletePartialMatch({node_id, db_name})` entry for `ProxyReportValue` inside `CleanupProxyDBMetrics` — it carries both labels, so partial match reaps it correctly. It cannot join `proxyCollectionScopedMetrics()` as-is, because `CleanupProxyCollectionMetrics` also passes `collection_name` and `DeletePartialMatch` would then match nothing for it. *(raised by czs007)*

- **`pkg/metrics/proxy_metrics.go:323` — `ProxyLimiterRate` series are never deleted on collection/partition drop, leaking cardinality unboundedly.** The vec is declared with labels `{node_id, collection_id, msg_type}` (lines 322-329) and emitted on the live quota push path: `Proxy.SetRates` (`internal/proxy/impl.go:5832`) → `SimpleLimiter.SetRates` (`internal/proxy/simple_rate_limiter.go:192`) → `updateRateLimiter` → `updateLimiterNode` (:289) → `setRateGaugeByRateType` (:367-372). The `collection_id` value is really a source id from `getCollectionSourceID`/`getPartitionSourceID` (:321-326), holding `cluster` / `db.<id>` / `collection.<id>` / `partition.<id>`. When a collection or partition disappears, `SetRates` prunes the limiter tree via `ClearInvalidLimiterNode` (:221 → `internal/util/ratelimitutil/rate_limiter_tree.go:196`), but nothing removes the gauge — the only three references in the repo are the declaration, the `MustRegister` at `proxy_metrics.go:540`, and the `Set` at `simple_rate_limiter.go:372`; there is no `Delete` or `DeletePartialMatch` anywhere. It also could not be cleaned by `CleanupProxyCollectionMetrics` even if listed, since that matches on `db_name`/`collection_name`, which this vec does not carry. Note the precondition: `setRateGaugeByRateType` returns early when the rate is `Inf` (:368-370), so with shipped defaults (collection/partition rate limits all -1) no series is created; the leak starts once any rate becomes finite. Failure: a per-collection or per-partition rate limit becomes finite — an explicitly configured `quotaAndLimits.dml.insertRate.collection.max`, or quota-center backpressure (memory water level, disk quota, force-deny) — and collections/partitions are then repeatedly created and dropped on a long-lived proxy → every distinct `collection.<id>`/`partition.<id>` × rate_type pair permanently retains a `milvus_proxy_limiter_rate` series, and each dead collection keeps exporting its last finite value, so dashboards and alerts read live limits for collections that no longer exist. Unrelated to this PR: the diff touches only `internal/proxy/task_search.go`, `internal/util/function/embedding/function_executor.go`, `pkg/metrics/metrics_test.go`, the `ProxySearchSparseNumNonZeros`/`ProxyFunctionlatency` label lists and `Cleanup*` functions in `pkg/metrics/proxy_metrics.go`, and `pkg/metrics/querynode_metrics.go` — the `ProxyLimiterRate` declaration and its only emit site in `simple_rate_limiter.go` (a file absent from the diff) are untouched, and the metric was equally uncleanable before the change. Recommend merging that separate fix before this PR if the two are sequenced together. *(raised by czs007)*

czs007:

**Verified:**
- `internal/proxy/task_search.go:692` — both emit sites of `ProxySearchSparseNumNonZeros` (also `:1041`) were updated to the new 5-label arity, so there is no `WithLabelValues` count mismatch left in the repo.
- `internal/util/function/embedding/function_executor.go:161` — the only two `ProxyFunctionlatency` emit sites (also `:223`) both pass the new 6-label set in declaration order.
- `pkg/metrics/metrics.go:134` — `functionLabelName` holds the same `"function_name"` string as the removed `functionName` const, so the emitted label name and existing dashboards are unaffected.
- `pkg/metrics/proxy_metrics.go:320` — `ProxyReportValue` is db-scoped only, is correctly excluded from the collection-scoped list, and is cleaned separately in `CleanupProxyDBMetrics`, so `DeletePartialMatch` on a collection cannot wipe a whole database.
- `pkg/metrics/metrics_test.go:321` — `require.NotEmpty(t, declared, …)` makes the new guard fail loudly if the parser matches nothing, instead of passing vacuously on a renamed cleanup function.
- `internal/proxy/impl.go:242` — cleanup reads `dbName := request.DbName`, the same raw source the newly db-scoped emit sites use (`t.request.GetDbName()` / `insertMsg.GetDbName()`), so emit and cleanup agree on the label value.

## Low (non-blocking)

- **`pkg/metrics/metrics_test.go:438`** — `assertCleanupCoversLabels` calls `parser.ParseFile` on the single file passed by the caller (`proxy_metrics.go`, `querynode_metrics.go`), so the completeness guard's scope is one file plus one cleanup function rather than the whole package. This is correct today: all 18 proxy `*Vec`s carrying both `db_name` and `collection_name` are declared in `proxy_metrics.go` and match `proxyCollectionScopedMetrics()` one-for-one, and the test's own doc comment states the limit. Suggestion: no change required for this PR — just be aware that a future collection-labeled `*Vec` declared in another file of `pkg/metrics` (including the rootcoord/datacoord metric files, which are outside the guard entirely) will not be caught and must be wired into the scoped list by hand; widening the guard to parse the package rather than a single file would close that gap. *(surfaced during verification)*

- **`pkg/metrics/rootcoord_metrics.go:303`** — `CleanupRootCoordCollectionMetrics` has no caller anywhere in the repo (only its definition matches; the sibling `CleanupRootCoordDBMetrics` is called at `internal/rootcoord/root_coord.go:952`), and its body calls `RootCoordNumEntities.DeletePartialMatch` twice with an identical label set while never deleting `RootCoordIndexedNumEntities`, which carries the same `db_name`+`collection_name` pair and is written at `internal/rootcoord/quota_center.go:532`. Failure: repeated create/drop of indexed collections (CI or multi-tenant churn) → `milvus_rootcoord_entity_num` and `milvus_rootcoord_indexed_entity_num` series for dropped collections keep their last value forever and grow monotonically until rootcoord restarts. Unrelated to this PR: `rootcoord_metrics.go` is untouched by this diff, which only changes `internal/proxy/task_search.go`, `internal/util/function/embedding/function_executor.go`, and `pkg/metrics/{metrics.go,metrics_test.go,proxy_metrics.go,querynode_metrics.go}`. *(surfaced during verification)*

czs007 (approved):
/lgtm

PR #52692, opened 2026-08-19
fix: update Python e2e assumptions for pymilvus 3.1

### LINKED ISSUE #52647: [Bug]: e2e tests not updated for pymilvus 3.1.0rc80 — new Date/Time/Decimal DataTypes break external-table dtype-matrix canary, client-side multi-PK validation breaks `invalid_schema_multi_pk`
## Environment

- **Milvus Version**: master @ ab35268275d1
- **SDK**: pymilvus 3.1.0rc80 (matches the pin in `tests/python_client/requirements.txt`)

### Option B: Steps (for systemic / stability / timing issues)

1. Deploy: any standalone deployment (for the first test case, no server is even required — the failing assertion is a pure client-side canary; the second fails before any request is sent).
2. Setup: `cd tests/python_client && pip install -r requirements.txt` (installs `pymilvus==3.1.0rc80`).
3. Workload: run the two test cases:
   - `pytest milvus_client/test_milvus_client_external_table.py -k test_milvus_client_external_table_add_all_supported_field_types` (test at `test_milvus_client_external_table.py:2782`)
   - `pytest milvus_client/test_milvus_client_collection.py -k test_milvus_client_collection_invalid_schema_multi_pk` (test at `test_milvus_client_collection.py:751`)
4. Trigger: simply running the tests against pymilvus 3.1.0rc80; no data or server state is involved.
5. Observed via: pytest failures / tracebacks (both fail before or independent of any server-side check).

### Trigger Conditions

- **Frequency**: deterministic with `pymilvus==3.1.0rc80` installed; independent of server version, deployment mode, MQ, or pytest parallelism. Both failures are decided entirely on the client side.
- **First observed after**: upgrading the installed pymilvus to 3.1.0rc80 (the version currently pinned in `tests/python_client/requirements.txt`).
- **Does NOT happen when**: an older pymilvus without the Date/Time/Decimal proto DataTypes and without client-side multi-primary-key validation is installed (inferred from the failure mechanism; not re-run against older versions in this pipeline).

## Expected Behavior

- `test_milvus_client_external_table_add_all_supported_field_types` expects its completeness canary `_assert_external_add_field_dtype_matrix_current()` (defined in the same file at line 104) to find every proto `DataType` classified in the test's add-field matrix, and then exercise add-field for the supported types.
- `test_milvus_client_collection_invalid_schema_multi_pk` expects `create_collection` with two primary-key fields to reach the server and be rejected with the server-side error `Expected only one primary key field`, which the test's `err_res` checker asserts on.

## Actual Behavior

- The external-table test fails at its first line: pymilvus 3.1.0rc80 introduces three new proto DataTypes (`Date=28`, `Time=29`, `Decimal=30`) that the test's matrix does not classify, so the canary fires (by design — it is built to fail when the proto gains new types).
- The multi-PK test never sends a request: pymilvus 3.1.0rc80 adds client-side validation in `CollectionSchema.add_field`, so the second `add_field(..., is_primary=True)` call (line 762) raises `PrimaryKeyException` inside the SDK (`pymilvus/orm/schema.py:237`, `_update_key_field_metadata`). The exception occurs outside the `err_res` check framework, so the test errors instead of passing.

## Error Logs

```
external add-field proto DataType matrix is stale; unclassified={28, 29, 30}, obsolete=set(), proto={... 'Date': 28, 'Time': 29, 'Decimal': 30 ...}
pymilvus.exceptions.PrimaryKeyException
  (raised from pymilvus/orm/schema.py:237, _update_key_field_metadata,
   during schema_1.add_field("field2", ..., is_primary=True) at test_milvus_client_collection.py:762)
```

## Non-default Configuration

All defaults; server configuration is irrelevant to these failures.

## Analysis Hints (Optional)

- This is a **test defect** (test-assumption/test-data staleness), not a product-behavior bug. Both failures are caused by the SDK moving forward while the tests stayed on old assumptions:
  - `test_milvus_client_external_table.py:2782` → canary at `:104` has no classification for the new `Date`/`Time`/`Decimal` DataTypes added in pymilvus 3.1.0rc80. The canary is working exactly as designed.
  - `test_milvus_client_collection.py:751` (assertion setup at `:762`) → assumes multi-PK rejection happens server-side; pymilvus 3.1.0rc80 now rejects it client-side in `CollectionSchema.add_field` before any RPC.
- Ruled out: any server-side regression. Evidence: the tracebacks terminate inside the pymilvus package (`pymilvus/orm/schema.py:237`) or in the test file's own canary; no request reaches the proxy, and the canary reproduces without Milvus running at all. The installed pymilvus (3.1.0rc80) matches the repo's `requirements.txt` pin, so this is not a wrong-environment install.
- Suggested fix (in the tests): classify `Date`/`Time`/`Decimal` in the external add-field dtype matrix, and change `invalid_schema_multi_pk` to expect the client-side `PrimaryKeyException` at `add_field` time. A fix covering both cases is already in flight under issue #51544 — worth tracking/landing rather than duplicating.
- Coverage note: with the SDK intercepting multi-PK schemas, the **server-side** multiple-primary-key validation path loses e2e coverage. If that server defense matters, a separate test that bypasses SDK validation (e.g. constructing the proto/request directly) is needed.
- Related issues: #51544

mvanhorn:
The Python e2e suite pins pymilvus 3.1.0rc80, whose schema proto now defines `Date`, `Time`, and `Decimal`, but the external-table dtype completeness canary does not classify those values. The repository has no server-side handling for those three types, matching the matrix's existing proto-only treatment for `Mol`, so they must not be exercised as supported external fields. The same SDK release rejects a second primary key while `CollectionSchema.add_field` is building the schema, causing the multi-PK test to error before its server-response checker runs. Both regressions are deterministic test-assumption failures and require no Milvus product-path changes.

## Validation

- Run the external-table dtype completeness canary with pymilvus 3.1.0rc80 and confirm proto values 28 (`Date`), 29 (`Time`), and 30 (`Decimal`) are classified as proto-only unsupported while all currently supported public types remain covered.
- Run `test_milvus_client_external_table_add_all_supported_field_types` and confirm the matrix canary passes without attempting to add `Date`, `Time`, or `Decimal` fields; the existing supported and unsupported add-field checks continue unchanged.
- Run the multi-PK test's explicit-primary-key schema case and confirm adding `field2` as a second primary key raises `PrimaryKeyException` at schema construction time, with no collection request expected.
- Run the `primary_field="field2"` schema case and confirm adding another primary field is likewise asserted at the SDK construction boundary rather than through the server-side `err_res` checker.
- Run the focused Python test selections together to ensure neither failure depends on a running Milvus server and that unrelated single-primary-key schema construction still proceeds normally.

## Summary

Extend `_EXTERNAL_ADD_FIELD_PROTO_ONLY_UNSUPPORTED_TYPES` in `test_milvus_client_external_table.py` with the proto names and numeric values for `Date`, `Time`, and `Decimal`, retaining the existing canary that will force reclassification if PyMilvus later exposes them as public types. In `test_milvus_client_collection.py`, import the SDK's `PrimaryKeyException` and update both multi-primary-key schema-construction variants to assert that the exception occurs at the `add_field` call that introduces the conflicting key, rather than attempting `create_collection`. Keep the assertions local to the existing tests and avoid adding a helper or bypassing SDK validation; direct server-side multi-PK coverage is a separate concern from repairing these pinned-SDK e2e tests.

Closes #52647


tests/python_client/milvus_client/test_milvus_client_collection.py
```diff
@@ -750,32 +751,20 @@ def test_milvus_client_collection_none_desc(self):
...
-        schema_2.add_field("field1", DataType.INT64, is_primary=True, auto_id=False)
-        schema_2.add_field("field2", DataType.INT64)  # Second primary key
-        schema_2.add_field("vector_field", DataType.FLOAT_VECTOR, dim=32)
-        # Try to create collection with multiple primary keys
-        error = {ct.err_code: 999, ct.err_msg: "Expected only one primary key field"}
-        self.create_collection(
-            client, collection_name, schema=schema_2, check_task=CheckTasks.err_res, check_items=error
-        )
+        with pytest.raises(PrimaryKeyException):
+            schema_2.add_field("field1", DataType.INT64, is_primary=True, auto_id=False)
```
yhmo, line 768: tests/python_client/milvus_client/test_milvus_client_collection.py line:767
High ---- This assertion cannot pass. Against pymilvus 3.1.0rc80 (the requirements.txt pin), a schema constructed with primary_field="field2" does not have a primary key registered until a field named "field2" is added; `_primary_field` stays None until then. The first add_field("field1", is_primary=True) therefore succeeds and becomes the single primary key, so `pytest.raises(PrimaryKeyException)` fails with "DID NOT RAISE" and the test remains red. I verified this directly against the 3.1.0rc80 wheel: adding field1 here does not raise; the SDK only raises when the field matching the `primary_field` name is added (in the previous code the exception fired at the now-removed add_field("field2", ...) call). Suggest adding the declared primary first, then asserting the conflicting add raises, e.g. `schema_2.add_field("field2", DataType.INT64)` before the `with pytest.raises(PrimaryKeyException)` block.

mvanhorn:
You were right, and thanks for pinning it to the exact wheel. Fixed in 6bce1a7.

I reproduced it against `pymilvus==3.1.0rc80` from the requirements pin, no server involved:

```python
schema = MilvusClient.create_schema(enable_dynamic_field=False, primary_field="field2")
schema.add_field("field1", DataType.INT64, is_primary=True, auto_id=False)
# -> does not raise
```

Exactly as you described: `_primary_field` stays None until a field named `field2` is added, so `field1` just becomes the sole primary key and the `pytest.raises` block sees DID NOT RAISE.

Adding `schema_2.add_field("field2", DataType.INT64)` before the assertion registers the declared primary field first, and then the `is_primary=True` on `field1` is a genuine second primary key, which is what the case is supposed to exercise:

```
PrimaryKeyException: (code=1, message=Expected only one primary key field, got [field2, field1, ...])
```

The `schema_1` half above it was already correct and is untouched, since it never declares `primary_field` up front.

tests/python_client/milvus_client/test_milvus_client_collection.py
yhmo: The issue is confirmed fixed. Resolving this conversation.

mvanhorn:
Closing this as redundant.

Rebasing onto current master makes both commits here empty: the multi-primary-key test has since been rewritten upstream using `self.add_field` with `check_task=CheckTasks.err_res`, which is the right convention for this suite, and it asserts on the second field addition, which is the correct point given that `_primary_field` is not registered until the named field exists. That is the same conclusion @yhmo reached in review. The external-table change is on master too.

So there is nothing left in this branch that master does not already have. Thanks for the review, it was the thing that got the semantics right.

Issue #52683, opened 2026-08-19
[Feature]: route general LIKE through FMINDEX with candidate recheck


daviddallakyan2005:
### Is there an existing issue for this?

- [x] I have searched the existing issues

[#51577](https://github.com/milvus-io/milvus/issues/51577) covers FMINDEX v1 (anchored LIKE). Its body says general LIKE (`Match`) and `RegexMatch` are deliberate follow-ups. That v1 is already implemented in #51375. This issue is the follow-up for general `Match` only.

### Is your feature request related to a problem? Please describe.

The merged design `docs/design-docs/design_docs/20260708-fm_index_scalar_index.md` lists this under "Deliberately deferred to follow-up PRs (NOT in this release)":

> General `Match` (`LIKE` with interior `%`/`_`) and `RegexMatch` two-phase (candidate + recheck) acceleration — reuses the NGRAM machinery; designed here but not wired in this PR (both stay on the brute-force path).

Anchored LIKE (`PrefixMatch` / `PostfixMatch` / `InnerMatch`) is exact from the index. A pattern such as `LIKE '%foo%bar%'` still brute-scans every row.

I intend to implement that deferred general `Match` item on VARCHAR. Equality, IN, RegexMatch, TEXT/JSON/ARRAY, case folding, and growing segments stay out of scope.

### Describe the solution you'd like.

Two-phase candidate generation plus recheck, as the merged design specifies.

Phase A. Split the LIKE pattern on unescaped `%`/`_` into literal fragments (`split_by_wildcard`, already used by NGRAM). For `%foo%bar%` that is `{foo, bar}`. Any matching row contains every fragment, so `Count`/`VisitMatchingDocs` on the rarest fragment is a correct filter. Intersecting several fragments is an optimization, not a requirement. Patterns with no literal fragment (`%`, `%_%`) decline to brute force. A pattern whose rarest fragment trips the existing `ShouldUseOp` count-first guard also declines.

Phase B. Candidates are a superset. Run the existing `LikePatternMatcher` over candidate rows only (via FM `Extract`), so the result equals brute force.

I will open a `feat:` PR against this issue and add a short delta to the existing design doc moving general `Match` out of deferred.

### Describe an alternate solution.

Leave general LIKE on the brute-force path until an in-house follow-up. That is what v1 does today.

### Anything else? (Additional Context)

Design: `docs/design-docs/design_docs/20260708-fm_index_scalar_index.md` (Query Semantics, General LIKE).
[...]

PR #52697, opened 2026-08-20
feat: route general LIKE through FMINDEX with candidate recheck

daviddallakyan2005:
General LIKE with interior `%` or `_` was declined by FMINDEX even though PrefixMatch, PostfixMatch, and InnerMatch already use it. This routes VARCHAR Match through `ShouldUseOp`: split the pattern on wildcards, take candidates from the rarest literal fragment, then recheck those rows on sealed VARCHAR with `LikePatternMatcher` via `ProcessDataByOffsets`.

FMINDEX load no longer materializes ISA. Match phase 2 rechecks candidate rows on sealed VARCHAR via ProcessDataByOffsets. The Match guard is locate-only again, the same occ times sa_sample_rate bound as Prefix, Inner, and Postfix.

Empty-fragment patterns still decline. RegexMatch, equality, IN, TEXT, JSON, ARRAY, and growing segments are unchanged. FMINDEX Match is never served as unrechecked PatternMatch candidates through the ScalarIndex path.

Design: `docs/design-docs/design_docs/20260708-fm_index_scalar_index.md`

issue: #52683

## How to test

`LOCAL_STORAGE_SIZE=10` is required on this host.

Without the VARCHAR recheck (product from `47f5cf5791`, tests kept): 5 FAILED (`ShouldUseOp(Match)` still prices Extract, so `%RARE%` / `%ZEBRA%` stay declined).

With `7a3e274a36`:

```
source scripts/setenv.sh
export DYLD_LIBRARY_PATH="$PWD/cmake_build/src:$DYLD_LIBRARY_PATH"
LOCAL_STORAGE_SIZE=10 cmake_build/unittest/all_tests --gtest_filter='FMIndex.*'
[==========] 23 tests from 1 test suite ran.
[  PASSED  ] 23 tests.
```

Includes `FMIndex.ExecutorPathMatchRechecksVarchar` and `FMIndex.MatchOracleEqualsBruteForce`.

Whole C++ module (three Darwin crashers excluded):

```
export LOCAL_STORAGE_SIZE=10
internal/core/output/unittest/all_tests --gtest_filter='-JsonIndexTest.LargeInt64LiteralDoesNotAliasInDoublePathIndex:JsonFlatIndexExprTest.PreservesLargeInt64LiteralPrecision:SortBufferTest.*'
[==========] Running 8326 tests from 473 test suites.
[  PASSED  ] 8167 tests.
```

None of the 138 failures are `FMIndex.*`. Reverse deps: Ngram 10/10, scalar/hybrid/inverted 26/26, json_stats_test 79/79, LikeConjunctExpr 1/1.

/cc @xiaofan-luan


czs007:

## Overview

Adversarial review of the new general `LIKE` (Match) index path found four issues — all cost-model, resource-accounting, or documentation gaps rather than correctness defects; results stay exact on every path reviewed.

## High

**`internal/core/src/index/FMIndex.cpp:317` — Match cost guard does not model the per-candidate full-row `Extract`**

`MatchGuardAccepts` reuses the bound calibrated for the anchored ops, `occ × sa_sample_rate < ratio × TotalTokens()` (FMIndex.cpp:338-341), which prices only the suffix-array locate. The Match branch additionally runs a full-row `fm_.Extract(d, 0, SIZE_MAX)` per candidate doc (FMIndex.cpp:398-399); that ≈ `candidates × avg_row_len` LF-steps is absent from the formula. `Extract` walks back from the ISA anchor one byte at a time with no early exit (`thirdparty/fmindex/index/fmindex/FMIndexInl.h:330-336`), each step a random wavelet rank probe, and `LikePatternMatcher`'s `min_required_bytes_` short-circuit (`common/RegexQuery.h:454`) runs only *after* `Extract` returns, so it cannot recover the cost. With the defaults `ratio = 0.001` (configs/milvus.yaml:626) and `sa_sample_rate = 8` (FMIndex.cpp:156), the guard admits `occ` up to `T/8000`, i.e. up to `(T/8000) × L` random LF-steps versus the `T` sequential byte comparisons of the scan it replaces. Failure: a VARCHAR field with ~KB rows (log lines) under default `fmindexCostRatio`, running `field LIKE '%GET%500%'` where the rarest fragment has mid-range selectivity → the guard admits the index path, the query pays `candidates × row_len` random LF-steps, and the same `LIKE` is slower after building FMINDEX than before, with no log or metric indicating the path choice. Suggestion: fold the recheck cost into `MatchGuardAccepts` — e.g. add a `candidates × (TotalTokens() / total_rows_)` term — or calibrate a separate, stricter ratio for Match. *(raised by xiaocai2333, bigsheeper)*

**`internal/core/src/index/FMIndex.cpp:792` — cache cell size is never refreshed after the first Match materializes the ISA sample table**

`resident_heap_bytes()` does count `isa_sample_` (thirdparty/fmindex/index/fmindex/FMIndex.h:280-291), but `ComputeByteSize()` is called only at build (FMIndex.cpp:253), `Load` (282), and `LoadEntries` (785), and `SetCellSize` runs exactly once in the mmap branch at the end of `LoadEntries` (FMIndex.cpp:792) — all three strictly before any `Extract`, when `isa_sample_` is still empty. The first Match recheck calls `ensureIsaSample()` (FMIndexInl.h:286-295), which allocates `text_len/8 + 1` uint64s ≈ `text_len` bytes and holds them for the object's lifetime; no path recomputes the cell size afterwards. Under `LoadView`, `resident_heap_bytes()` covers only the rank directory (`QuadVector.h:141-144`, `owned_words_` empty) at ≈ `text_len/8`, so post-Match resident heap is roughly 9× the accounted value. Failure: a querynode mmap-loads several FMINDEX VARCHAR sealed segments, then any `LIKE '%a%b%'` hits them → each segment silently allocates ≈ `text_len` bytes of permanently held heap that the caching layer cannot see (up to ~8× under-accounting), admission/eviction decisions go wrong, and the querynode exceeds its configured memory budget and can be OOM-killed. Suggestion: call `ComputeByteSize()` + `SetCellSize()` after the first materialization, or reserve the ISA term at load time. The PR's design doc already notes the gap ("That table is not in the load-time cache cell size."), but the code-level accounting is still missing. *(raised by bigsheeper, xiaocai2333)*

## Medium

**`internal/core/thirdparty/fmindex/index/fmindex/FMIndexInl.h:286` — the first Match builds the ISA table on the query thread with an O(text_len) full BWT scan**

`ensureIsaSample()` scans all `m = text_len + 1` BWT rows inside `std::call_once` (one `sampled_bv_.get(r)` per row, plus a `rank1` on hits) and `assign`s a ≈ `text_len`-byte buffer. The code is thread-safe; the problem is that this PR's recheck (FMIndex.cpp:398-399) is the only production caller of `Extract`, so the work now lands on a query thread — the first Match pays the whole latency and every concurrent Match on that segment blocks on the `call_once`. Failure: a freshly loaded ~1 GB-text FMINDEX sealed segment receiving several concurrent first-time internal-wildcard `LIKE` queries → one thread serially walks `text_len + 1` BWT rows and allocates ≈ 1 GB (hundreds of ms to seconds) while the rest block, showing up as a P99 spike on the first batch of `LIKE` queries after load. Suggestion: pre-build the ISA table in `LoadEntries`, alongside `ComputeTotalTokens`, so the cost falls on the single-threaded load path — which also makes `ComputeByteSize` accurate and addresses the cell-size item above. *(raised by bigsheeper, xiaocai2333)*

## Low

**`internal/core/src/index/FMIndex.h:275` — the new `MatchGuardAccepts` comment equates Match with the anchored ops without disclosing the `Extract` cost**

The declaration comment states that "the rarest fragment is scored with the same `occ x sa_sample_rate < ratio x tokens` bound as the anchored ops", but never mentions that Match additionally pays one full-row `Extract` per candidate. The ratio-calibration note it points back to (FMIndex.h:180-193, item 2) explicitly scopes itself to "the anchored pattern ops" and its measured values (0.0008 / 0.001 / 0.002-0.008) were not re-derived or caveated for Match. A future maintainer reading this will reasonably conclude that reusing the same ratio for Match is safe. Suggestion: state the extra recheck cost in the `MatchGuardAccepts` comment and add a note to item 2 that its calibration does not cover Match. Same root cause as the guard item above. *(raised by xiaocai2333, bigsheeper)*

xiaofan-luan:
Suggestion: please avoid using `fm_.Extract()` for the phase-2 exact recheck.

`FMINDEX` declares `HasRawData() == false`, so the sealed segment still keeps the original VARCHAR field data available. A cleaner boundary would be:

1. Use FM only for phase-1 candidate generation (the rarest literal fragment -> candidate row offsets).
2. Batch-retrieve those candidate rows from the original VARCHAR column.
3. Run `LikePatternMatcher` on the retrieved `std::string_view`s for the exact result.

This follows the same coarse/refine idea as GIS and is also consistent with the existing NGRAM phase-1/phase-2 LIKE path. Prefer grouping the sorted candidate offsets by chunk and using the existing offset/batch data access helpers (`ProcessDataByOffsets<std::string_view>`, `get_views_by_offsets`, or equivalent) instead of one retrieve per row.

The main benefit is that general LIKE no longer needs the lazy ISA table at all: no extra ~one-byte-per-text-byte retained allocation, no cache-cell accounting gap, and no first-Match O(text_len) ISA construction on a query thread. The anchored FM operations can remain index-exact as they are.

I think the current cost guard is acceptable for this PR; my main recommendation is specifically to move the exact recheck out of `FMIndex::PatternMatch` and perform it against the original VARCHAR data rather than `fm_.Extract()`.

daviddallakyan2005:
@czs007 Done in `47f5cf5791`.

`MatchGuardAccepts` now prices locate plus Extract: `occ * (sa_sample_rate + avg_row_bytes) < ratio * tokens`, with `avg_row_bytes = TotalTokens() / non-null rows`.

Load and build call `MaterializeIsaSample` before `SetCellSize`, so the cache cell includes the ISA table. The first Match no longer builds ISA on a query thread.

The `MatchGuardAccepts` comment and the design doc now name the Extract cost.

daviddallakyan2005:
@xiaofan-luan The cost guard and load-time ISA are in `47f5cf5791`, which addresses the two must-fix items on Extract accounting.

Recheck is still inside `PatternMatch` via Extract. Moving the exact recheck onto the original VARCHAR column (`HasRawData() == false`) would drop the ISA table for Match. That is a follow-up unless you want it on this PR.

daviddallakyan2005:
@xiaofan-luan Match phase 2 now rechecks sealed VARCHAR via ProcessDataByOffsets. PatternMatch returns rarest-fragment candidates only. ISA is not materialized at load.

daviddallakyan2005:
@czs007 Extract is off the Match path, so load-time ISA and the Extract guard term are dropped. The locate-only guard remains.

internal/core/thirdparty/fmindex/index/fmindex/FMIndex.h
```diff
@@ -151,6 +151,14 @@ class FMIndex {
     std::string
     Extract(uint64_t doc_id, uint64_t offset, size_t len) const;
 
+    // Build isa_sample_ if it is still empty. The scalar wrapper does not
+    // call this at load; Match rechecks sealed VARCHAR. Thread-safe; a no-op
+    // after the first call.
+    void
+    MaterializeIsaSample() const {
```
yhmo: internal/core/thirdparty/fmindex/index/fmindex/FMIndex.h line:158
Low ---- MaterializeIsaSample() is added as a public method on the bundled library but is never called anywhere in this PR or the repository (Match rechecks sealed VARCHAR instead, and the wrapper never materializes ISA at load). Either wire it up from the wrapper's load/refresh path or drop it, so the library does not carry a dead public API whose comment describes an intent no caller implements.

internal/core/src/index/FMIndex.cpp
```diff
@@ -32,6 +32,7 @@
 #include "common/Slice.h"
 #include "common/Tracer.h"
 #include "index/Meta.h"
+#include "index/NgramInvertedIndex.h"
```
yhmo: internal/core/src/index/FMIndex.cpp line:35
Low ---- This include pulls in index/NgramInvertedIndex.h, which transitively brings InvertedIndexTantivy.h (Rust/cgo bindings) and boost/filesystem, only to use the ~15-line split_by_wildcard helper. Consider moving split_by_wildcard into a lighter shared header (e.g. common/RegexQuery.h or a small like-pattern util) so the FM index does not couple to the Ngram/Tantivy header and this translation unit's compile surface stays small.

internal/core/thirdparty/fmindex/index/fmindex/FMIndex.h
daviddallakyan2005: Done. MaterializeIsaSample is gone. ISA stays lazy on Extract only.

internal/core/src/index/FMIndex.cpp
daviddallakyan2005: Done. split_by_wildcard now lives in common/RegexQuery. FMIndex.cpp includes that header instead of NgramInvertedIndex.h.

czs007:

Adversarial review of the FMINDEX general-`LIKE` (`Match`) path found one cost-model regression, one lost prefetch, and three smaller test/doc gaps; the recheck design itself held up under verification.

## High

**`internal/core/src/index/FMIndex.cpp:353` — the cost guard prices only the locate, omitting phase-2 recheck bytes**

`return static_cast<double>(occ) * sr < ratio * tokens;` is sound for `PrefixMatch`/`InnerMatch`/`PostfixMatch`, whose index answers are exact — `occ × sr` is the entire cost there. `Match` has a phase 2 whose cost is `Σ|row|` over candidate documents, and that term appears nowhere in the guard; `occ` is an occurrence count, which bounds candidate docs from above but says nothing about their byte size. With defaults (`build_sa_sample_rate_ = 8`, `fmindexCostRatio = 0.001`) and the "rarest fragment occurs once per non-null row" shape (`occ = N`, `tokens ≈ N × L`), the guard reduces to `sr < ratio × L` — i.e. accept whenever `L > 8000` bytes, independent of row count, and VARCHAR `max_length` reaches 65535.

Failure: a VARCHAR column averaging >8000 bytes/row (JSON blobs, long log lines) with FMINDEX and default `queryNode.fmindexCostRatio` / `fm_sa_sample_rate`, queried as `LIKE '%ERROR%timeout%'` where the rarest fragment appears in nearly every row → `ShouldUseOp(Match)` accepts, the candidate set is effectively the whole segment, and `ExecFMMatch` phase 2 reads the column back through `ProcessDataByOffsets` and runs `LikePatternMatcher` row by row — the same brute-force scan taken before this PR, plus a wasted `occ × sa_sample_rate`-step random-LF locate (~65× row count at 65 KB rows) and a full-segment candidate bitmap. Before this PR `Match` was always declined by `ShouldUseOp`, so this is a pure regression on default config.

Two related inconsistencies: the comment at `FMIndex.cpp:351-352` describes phase 2 as a sequential stream of candidate VARCHAR rows, but `ProcessDataByOffsets` (`UnaryExpr.cpp:2522` → `Expr.h:952-1000`) resolves each candidate offset's chunk and calls `get_views_by_offsets` per run — in-chunk random access, not a sequential read. And the earlier commit on this branch (`47f5cf5791`) had `occ * (sr + avg_row) < ratio * tokens`; `d3459ea6cb` dropped the `avg_row` term when the recheck moved from `Extract` to reading sealed VARCHAR, but changing how those bytes are read does not make them free.

Suggestion: restore a candidate-bytes term for `Match` only — e.g. `occ × sr + doc_estimate × avg_row_bytes < ratio × tokens` with `avg_row_bytes = TotalTokens() / non-null rows` — or at minimum cap the accepted candidate fraction (decline when the rarest fragment's estimated doc count exceeds some fraction of `total_rows_`), so the index path cannot degenerate into a full recheck at any row length. *(surfaced during verification)*

## Medium

**`internal/core/src/exec/expression/Expr.h:2953` — FM `Match` reads raw VARCHAR in phase 2 but no longer gets the raw-data prefetch**

`PrefetchAsync` calls `PrefetchRawData()` only when `exec_path_ == ExprExecPath::RawData`. `CanUseFMMatch()` requires `exec_path_ == ExprExecPath::ScalarIndex` (`UnaryExpr.cpp:2425`), so every general `LIKE` accepted by the guard skips prefetch entirely — while `ExecFMMatch` phase 2 goes straight at the raw VARCHAR column via `ProcessDataByOffsets` → `get_views_by_offsets`. This diverges from the NGRAM two-phase path it mirrors: `HasCompatibleScalarIndex()` explicitly excludes ngram (`Expr.h:2622`, `return has && !CanUseNgramIndex();`), so ngram `LIKE` keeps `exec_path_ = RawData` and does prefetch the column it rechecks. The prefetch branch itself is untouched by this PR, but adding `Match` to the allowlist is what first lands a raw-column-reading path on the non-`RawData` side of it — previously `DetermineExecPath` downgraded `Match` to `RawData` (`UnaryExpr.cpp:2239-2241`).

Failure: tiered storage or cold mmap cache, an accepted general `LIKE` (e.g. `LIKE '%req-2026%GET%'`) whose candidates scatter across many chunks → each candidate's chunk faults in synchronously on the query thread inside `get_views_by_offsets` instead of being warmed asynchronously on the prefetch pool as it was pre-PR; tail latency scales with the number of chunks the candidates span, which is exactly what the rarest-fragment heuristic produces.

Suggestion: prefetch raw data when the expression will run phase 2 — e.g. also call `PrefetchRawData()` in `PrefetchAsync` when `CanUseFMMatch()` is true (`pinned_index_` is already populated by `EnsureExecPathDetermined` at that point). *(surfaced during verification)*

**`internal/core/src/index/FMIndexTest.cpp:821` — the new "ISA stays lazy" assertions cannot fail, and they replaced the one that could**

`ByteSize()` returns the cached field `cached_byte_size_` (`Index.h:136-138`); only `ComputeByteSize()` rewrites it, and it is reached solely through `RefreshResidentSize()`, whose three call sites are `BuildWithFieldData` / `BuildWithRawDataForUT` / `LoadEntries` (`FMIndex.cpp:252, 281, 778`) — never during query. So `(void)Match(idx.get(), "%app%"); EXPECT_EQ(idx->ByteSize(), bytes_after_load);` reads one immutable field twice and holds regardless of what `Match` does; `EXPECT_EQ(idx->CellByteSize().memory_bytes, bytes_after_load)` likewise reads the `cell_size_` written by `SetCellSize` at load. If a future change made `Match` call `fm_.Extract()` and materialize the ISA table (`m/rate × 8 B` of heap), both assertions would still pass — they do not assert the lazy-ISA invariant that three commits on this branch were spent establishing. The same hunk also deletes `EXPECT_LT(idx->CellByteSize().memory_bytes, kEstimatedMemory / 2)`, the only assertion bounding post-`LoadView` resident memory against the conservative estimate at the wrapper/cache-cell layer (the `EXPECT_LT` at line 422 covers the fmindex library layer, not the wrapper).

Suggestion: assert against a freshly computed size — `idx->ComputeByteSize(); EXPECT_EQ(idx->ByteSize(), bytes_after_load);`, or compare `fm_.resident_heap_bytes()` before and after the `Match` — and restore `EXPECT_LT(..., kEstimatedMemory / 2)`, which should still hold now that ISA is not materialized at load and is the regression detector for exactly that claim. *(surfaced during verification)*

## Low

**`internal/core/src/exec/expression/UnaryExpr.cpp:2461` — the multi-batch path of `ExecFMMatch` is untested, though it is the default shape in production**

Executor-level tests drive expressions through `ExecuteQueryExpr` → `FilterBitsNode`, and `CanExecuteAllAtOnce()` returns true whenever `exec_path_ != RawData` (`Expr.h:2804-2806`), so `SetExecuteAllAtOnce()` raises `batch_size_` to `active_count_` (`Expr.h:2810-2812`, `FilterBitsNode.cpp:189-191`) and the whole run collapses into a single `Eval`. The three pieces of `ExecFMMatch` that only matter across batches — cursor advance (`segment_offset = current_index_chunk_pos_`, line 2470), `cached_phase1_res_` reuse (lines 2461-2468), and the `bitmap_input` size assertion and AND (lines 2475-2482) — are therefore never executed, and `FMIndexTest.cpp` contains no conjunction cases. That shape is not rare: `ReorderConjunctExpr` buckets `LIKE` into `str_like_expr` rather than `indexed_expr` (`Expr.cpp:793-800`, `HasIndex(...) && !IsLikeExpr(input)`), so `LIKE ... AND <raw-data predicate>` makes `ConjunctExpr::CanExecuteAllAtOnce()` false and forces FM `Match` through the batch loop with a `bitmap_input`. Hand-tracing two batches through `GetNextBatchSize` / `MoveCursorForIndex` / `MoveCursorForData`, the cursor semantics look self-consistent — this is a coverage gap, not a known defect.

Suggestion: add a conjunction case that forces the batch loop, e.g. `LIKE '%a%b%' AND <raw-data predicate>`. *(surfaced during verification)*

**`configs/milvus.yaml:644` — the user-facing doc for `queryNode.fmindexCostRatio` was not updated for its widened scope**

This PR makes the same threshold decide whether general `LIKE` (`Match`) takes the index path (`FMIndex.h:239-240` → `FMIndex.cpp:326-354`, plus the new `case proto::plan::Match` in `StringLiteralForCostGuard` at `UnaryExpr.cpp:2053`), but both copies of the documentation are unchanged: `configs/milvus.yaml:644` and `pkg/util/paramtable/component_param.go:4541` still scope it to "LIKE prefix/infix/suffix" and claim it is "row-length invariant". The latter no longer holds for `Match` — under the fragment-per-row shape the acceptance condition reduces to `sa_sample_rate < ratio × average row length`, i.e. entirely row-length determined (see the High item above). An operator tuning from this text cannot tell that the knob now also routes general `LIKE`, nor predict its behaviour there.

Suggestion: state in the doc that `Match` is also subject to this guard and that its index path is candidate-generation plus recheck rather than an exact index answer. Edit `component_param.go` and regenerate/sync `configs/milvus.yaml`, since the yaml comment is produced from the param `Doc`. *(surfaced during verification)*

xiaofan-luan:
@czs007 Thanks for the detailed analysis. I agree that phase 2 has a recheck cost, but I do not think this should be a must-fix for this PR.

The guard is an empirical heuristic, not a complete analytical cost model. `total_tokens` is intentional: as rows get longer, the raw LIKE scan becomes more expensive, while FM candidate generation stays tied mainly to occurrence count. For selective candidates, longer rows are exactly where FMINDEX gains more. Adding `occ * avg_row_bytes` would also overestimate recheck cost because `occ` is an occurrence count rather than the number of distinct candidate documents, especially on repetitive text.

The >8 KB / candidate-every-row case is a valid corner case where the heuristic can choose a slower path, but it shows estimation error rather than an incorrect routing principle. There is no correctness impact, and I do not see end-to-end benchmark evidence here that the current default ratio needs to change. Candidate offsets are also produced in row order and processed chunk by chunk, so phase-2 access is less arbitrary than a random per-row lookup.

I suggest keeping the current guard in this PR and revisiting the ratio or model only with end-to-end measurements across row length and candidate selectivity. Please downgrade this from must-fix.

xiaofan-luan:
@daviddallakyan2005 Overall looks good to me. The phase-1 candidate generation plus exact recheck against the original VARCHAR data is a sound design.

I have two remaining test suggestions:

1. Please add a Go integration/E2E test to verify that general LIKE semantics remain unchanged after FMINDEX is built and loaded. Ideally, run the same queries before and after creating FMINDEX and compare the returned PKs. Please include:
   - a selective pattern that should use FM and produces phase-1 false positives requiring recheck;
   - an unselective or no-literal pattern that should fall back to the raw scan;
   - empty/null VARCHAR values and multiple flushed sealed segments.

2. Please expand the C++ UT coverage for the executor path:
   - a conjunction that forces multiple batches and supplies `bitmap_input`;
   - candidates spanning multiple raw-data chunks and chunk boundaries, including chunks with no candidates;
   - phase-1 result reuse and cursor advancement across batches;
   - nullable rows and offset-input fallback;
   - if possible, a test-only path observation so the test fails when an expected FM path silently falls back to raw data, or vice versa.

Two small cleanups would also help:

- The current ISA/`ByteSize()` assertion only rereads the cached size, so it does not detect a future ISA allocation. Please recompute the size before asserting, restore the previous memory bound, or remove the ineffective assertion.
- Update the `fmindexCostRatio` configuration documentation to mention that it now also applies to general Match, whose FM path is candidate generation followed by exact recheck.

Other than these coverage and documentation items, the implementation looks good to me. I do not consider the current cost heuristic or full-column prefetch discussion blockers for this PR.

internal/core/src/exec/expression/UnaryExpr.cpp
```diff
@@ -2391,6 +2409,138 @@ PhyUnaryRangeFilterExpr::ExecuteNgramPhase2(TargetBitmap& candidates,
...
+    if (cached_phase1_res_ == nullptr) {
+        auto candidates =
+            index->PatternMatch(literal, proto::plan::OpType::Match);
+        cached_phase1_res_ =
+            std::make_shared<TargetBitmap>(std::move(candidates));
+        cached_index_chunk_valid_res_ =
+            std::make_shared<TargetBitmap>(index->IsNotNull());
+    }
+
+    const int64_t segment_offset = current_index_chunk_pos_;
```
yhmo, line 2470: internal/core/src/exec/expression/UnaryExpr.cpp line:2470
Medium ---- ExecFMMatch slices the cached phase-1 bitmap from `current_index_chunk_pos_` (the ScalarIndex cursor), while every other cached-bitmap path in this class slices from `current_data_global_pos_` (ExecNgramMatch, ExecRangeVisitorImplForIndex via MoveOrSliceBitmap, ExecTextMatch, ExecRangeVisitorImplForPk). The two cursors stay in sync only when every batch takes the same execution path. If `EnsureExecPathDetermined()` has not run before the first `Eval` -- for example when `ENABLE_DRIVER_PREFETCH` is disabled (configurable in Common.cpp) or a caller evaluates the expr directly (ElementFilterIterator / BoostScoreRunner / direct test harnesses) -- the first batch runs `ExecRangeVisitorImplForData` as a raw scan (exec_path_ still defaults to RawData and pinned_index_ is empty), and inside that call `GetNextBatchSize()` determines exec_path_ = ScalarIndex. Subsequent batches then enter ExecFMMatch while `current_index_chunk_pos_` is still 0, re-evaluating rows [0, batch_size) and never processing the rows after them -- silent wrong results, because the raw scan advances only the data cursor (ProcessDataChunksForSingleChunk) and never the index cursor. ExecNgramMatch avoids this by slicing from `current_data_global_pos_`. Consider slicing from `current_data_global_pos_` here as well, or calling `EnsureExecPathDetermined()` before the FM branch dispatch.

internal/core/src/index/FMIndex.cpp
```diff
@@ -355,6 +401,37 @@ FMIndex::PatternMatch(const std::string& pattern, proto::plan::OpType op) {
                                   [&](uint64_t d) { bitset.set(d); });
             return bitset;
         }
+        case proto::plan::OpType::Match: {
+            // Phase 1 only: rarest literal fragment to candidate rows.
+            // ExecFMMatch rechecks those offsets on sealed VARCHAR.
+            TargetBitmap candidates(total_rows_);
+            auto parts = split_by_wildcard(pattern);
+            if (parts.empty()) {
```
yhmo, line 415: internal/core/src/index/FMIndex.cpp line:409
Low ---- For the `Match` op, `MatchGuardAccepts` returns true for an empty pattern (line 328) and `ShouldUseOp(Match, "")` therefore accepts, but this `parts.empty()` branch then returns ALL non-null rows as candidates and ExecFMMatch rechecks every one of them with `LikePatternMatcher` -- a full per-row matcher scan. That contradicts the `ShouldUseOp` contract comment ("An empty pattern means ... answered by an O(rows) bitmap clone, always accelerate", FMIndex.h) and the PR description's "Empty-fragment patterns still decline". For non-empty patterns with no literal fragment (`%`, `%_%`) this branch is unreachable from the executor (the guard declines them), so the all-rows behavior is only reachable through the direct API or an empty pattern; the guard/API/documentation contract should be aligned -- e.g. decline empty patterns for `Match` too, or document the full-recheck cost.

internal/core/src/index/FMIndex.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/core/thirdparty/fmindex/index/fmindex/FMIndex.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

internal/core/src/index/FMIndex.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

daviddallakyan2005:
@xiaofan-luan Thanks. I added the remaining coverage.

Go MiniCluster `tests/integration/fmindex` runs the same LIKE expressions before and after FMINDEX and compares PKs. That includes a selective interior wildcard, an unselective `%%` fallback, empty and null VARCHAR, and two flushed sealed segments.

C++ `FMIndex.ExecutorPathMatchBatchesAndBitmap`, `ExecutorPathMatchMultiChunk`, and `ExecutorPathMatchNullableAndOffsets` cover batched conjunction with `bitmap_input`, multi-chunk candidates, nullable rows, and the offset fallback. After Match the ISA mmap test recomputes `ByteSize()` and restores the previous memory bound.

`queryNode.fmindexCostRatio` documentation already covers general Match as candidate generation plus exact recheck.

internal/core/src/exec/expression/UnaryExpr.cpp
daviddallakyan2005: I left this as is in this follow-up. The remaining work here was the MiniCluster and executor tests. ExecFMMatch still slices cached phase 1 from `current_index_chunk_pos_`. Changing that cursor would be a product change I did not take on in this commit.

PR #52698, opened 2026-08-20
fix: [3.0] run MinHash schema validation on all DDL paths and reject TEXT input (#51649)

MrPresent-Han:
### What

Admission gaps around MinHash functions, each letting DDL persist
collections that can never work:

- **`disableRuntimeCheck=true` paths skipped MinHash parameter
validation.** `ValidateMinHashFunction` only ran inside
`embedding.ValidateFunctions`, which `validator.ValidateFunction(...,
disableRuntimeCheck=true)` skips entirely. The check is pure schema
validation with no external calls, so hoist it into
`validator.ValidateFunction` where it runs regardless of the
runtime-check switch, and drop the now-redundant special case in the
embedding executor.
- **The direct RootCoord CreateCollection path ran no function
validation at all.** `prepareSchema` only validated fields and assigned
IDs — a client hitting RootCoord directly could persist a MinHash (or
any function) schema the proxy path rejects. Run
`validator.ValidateFunction(schema, "", true)` after
`assignFieldAndFunctionID` (IDs canonical). Scoped to newly created
schemas: the `preserveFieldID` branch restores an existing collection
(snapshot/replication), whose historical schema must not be re-judged by
current-version rules, and **external collections are exempt** —
`prepareSchema` sees the resolved schema (nullability inferred from the
external source), not the user-submitted form the proxy validated, so
judging it by user-schema rules rejects legal external tables (caught by
the go-sdk e2e `TestExternalTableTextEmbeddingFunction`).
- **The validation itself was overflow-bypassable.**
`ValidateMinHashFunction` compared `outputDim != int64(numHashes * 32)`:
with dim=32 and `num_hashes=2^59+1` the multiplication wraps to 32 and
the check passes, letting the runner later attempt an impossibly large
permutation allocation. Compare via division (`outputDim%32 == 0 &&
outputDim/32 == numHashes`) so the hoisted check is actually
authoritative. Defense-in-depth at the runner's allocation site is a
follow-up (kept with the runner-hardening split of #51848).
- **`CheckFunctionInputField` admitted TEXT inputs for MinHash,
contradicting the runner-side contract.** `ValidateMinHashFunction` has
always been VarChar-only ("is not string type"), so the two validators
disagreed on the same schema; with the gaps above closed, align the
admission check to VARCHAR-only. Note: inserts into an existing
TEXT-input collection do work today (TEXT arrives as `StringData` on the
wire); the mismatch bites in the validator disagreement and in
backfill-style paths reading persisted TEXT back as LOB references —
this PR fixes admission, it does not claim runtime insert breakage.

Split from #51848.

### Tests

- `Test_createCollectionTask_prepareSchema_validatesFunctions`: the real
RootCoord create path rejects an overflowing `num_hashes` and accepts a
valid MinHash schema.
-
`Test_createCollectionTask_prepareSchema_skipsFunctionValidationForExternal`:
an external table with a nullable TextEmbedding input (the resolved
form) passes.
- `TestValidateMinHashFunctionRejectsNumHashesOverflowingDimCheck`:
overflow boundary (`2^59+1`) rejected by the division-based check.
- New `validator_test.go`: MinHash params checked with runtime checks

issue: #51649
pr: #52219

### DIFF
--- a/internal/proxy/util_test.go
+++ b/internal/proxy/util_test.go
@@ -3362,6 +3362,31 @@ func TestValidateFunctionInputField(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("Valid MinHash function input - varchar", func(t *testing.T) {
+		function := &schemapb.FunctionSchema{
+			Type: schemapb.FunctionType_MinHash,
+		}
+		fields := []*schemapb.FieldSchema{
+			{DataType: schemapb.DataType_VarChar},
+		}
+		err := validator.CheckFunctionInputField(function, fields)
+		assert.NoError(t, err)
+	})
+
+	t.Run("Invalid MinHash function input - TEXT rejected", func(t *testing.T) {
+		// align with ValidateMinHashFunction's VarChar-only input contract:
+		// admitting TEXT here let the direct RootCoord path (which skipped that
+		// check) create schemas the runner-side validator rejects
+		function := &schemapb.FunctionSchema{
+			Type: schemapb.FunctionType_MinHash,
+		}
+		fields := []*schemapb.FieldSchema{
+			{DataType: schemapb.DataType_Text},
+		}
+		err := validator.CheckFunctionInputField(function, fields)
+		assert.Error(t, err)
+	})
+
 	t.Run("Invalid BM25 function input - multiple fields", func(t *testing.T) {
 		function := &schemapb.FunctionSchema{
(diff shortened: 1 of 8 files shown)


MrPresent-Han:
Superseded by #52944, which carries the same #52219 backport combined with #52902 and the required L2 assertion update (the `expected dim 512` message change) — merging this PR alone would break the 3.0 nightly.

Issue #52699, opened 2026-08-20
FMINDEX: infix LIKE filter spends >99% of its time in the per-occurrence suffix-array locate


xiaofan-luan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What would you like to be added?

Two independent inefficiencies in the FMINDEX doc-granularity filter path — the one `LIKE '%x%'` runs through (`index::FMIndex::PatternMatch(InnerMatch)` → `fmindex::FMIndex::VisitMatchingDocs`).

**1. The per-occurrence LF-walk is serial.**

`VisitMatchingDocs` walks the SA interval one row at a time:

```cpp
for (size_t i = r.first; i < r.second; ++i) {
    uint64_t pos = locateRow(i);          // while (!sampled) row = LF(row);
    if (pos < text_len_) visit(docOf(pos));
}
```

Each `locateRow` is a chain of *dependent* cache misses — the next wavelet address is only known after the previous read returns — so the core has exactly one outstanding miss at a time and the memory-level parallelism of the machine goes unused.

The occurrences are mutually independent, and this index already exploits that elsewhere: `LocateDocsBatch` runs LF-walks in lock-step tiles, and its tiling granularity is already *per occurrence*, not per pattern. The filter path just never routed through it.

**2. `docOf` binary-searches `doc_start_`.**

```cpp
auto it = std::upper_bound(doc_start_, end, internal_pos);
```

`doc_start_` is 8 bytes per document (1M documents = 8 MB), so once it outgrows L2 every probe of the search is a miss, and this runs once per located occurrence.

### Why is this needed?

Measured on Apple Silicon with a synthetic document-scoped corpus (`sa_sample_rate=8`), the backward search is free and locate is everything:

| corpus | query | occ/query | backward search | full filter |
|---|---|---|---|---|
| 200k docs x 500 B | mid-frequency word | 1967 | 0.60 us | 1275 us |
| 2M docs x 50 B | mid-frequency word | 1621 | 0.83 us | 1813 us |

That is >99% of the time in locate, at roughly 0.6–1.3 us per occurrence — a cost only explicable as cache misses.

Both fixes are local to the vendored `thirdparty/fmindex` library, need no public API change, and do not touch the on-disk format.

PR #52700, opened 2026-08-20
enhance: tile the FM-index locate walk and bucket docOf

xiaofan-luan:
Fixes the two hot spots identified in #52699. Both are local to the vendored
`thirdparty/fmindex` library; no public API and no on-disk format change.

## What changed

**1. `VisitMatchingDocs`: serial walk → 128-row lock-step tiles.**

Each occurrence's LF-walk is a chain of *dependent* cache misses, so walking
them one at a time keeps exactly one miss in flight. The occurrences are
independent, and `LocateDocsBatch` already tiles LF-walks in lock-step at
*per-occurrence* granularity — the filter path just never routed through it.
The tile is fixed-size, so peak memory stays O(1) and the streaming contract
(the reason this function exists instead of `MatchingDocs`) is preserved.

Below 8 hits the tile bookkeeping costs more than it saves, so those stay
serial — without that fast path single-hit patterns regressed ~40%, and those
are exactly the selective queries a filter sees most.

**2. `docOf`: binary search → bucket index.**

`doc_start_` is 8 B/document, so past ~L2 every probe of the binary search is a
miss, once per located occurrence. `buildDerived()` now also builds a bucket
table: one array read lands in a bucket, then a short scan over *consecutive*
`doc_start_` entries. Bucket width targets ~8 documents, so the table is sized
by **document count** (~`n_docs`/8 entries), not text length — 2M documents is
about 1 MB. It is never serialized; an empty table falls back to the binary
search.

## Results

Apple Silicon, `sa_sample_rate=8`, synthetic document-scoped corpora:

| 200k docs x 500 B | baseline | this PR | |
|---|---|---|---|
| rare (occ 429) | 271 us | 169 us | **1.61x** |
| mid (occ 1967) | 1276 us | 817 us | **1.56x** |
| hot (occ 263k) | 163 ms | 100 ms | **1.63x** |

| 2M docs x 50 B | baseline | this PR | |
|---|---|---|---|
| rare (occ 354) | 477 us | 203 us | **2.35x** |
| mid (occ 1621) | 2185 us | 869 us | **2.51x** |
| hot (occ 217k) | 274 ms | 103 ms | **2.66x** |

Attributing the two independently (same measurement round):

| | 200k docs | 2M docs |
|---|---|---|
| tiling only | 1.75–1.92x | 1.74–1.88x |
| docOf bucket only | 1.07–1.16x | 1.2–1.4x |
| both | 2.04–2.16x | 2.45–2.69x |

Tiling carries most of it and is insensitive to corpus shape. The bucket's
value tracks document count: at 200k documents `doc_start_` is 1.6 MB and fits
in L2, so it buys little; at 2M documents (16 MB) it buys 1.2–1.4x. Real
segments sit at the larger end.

Constants were measured, not guessed:
- tile size: swept 16/32/64/128/256; 128 best (256 adds nothing and doubles the
  stack frame).
- serial threshold: swept occ 1..256; the tiled path is at parity through occ 8
  and already 1.20x by occ 10, so 8 is the right cut.

## Verification

- **A/B result fingerprints identical** — 2 corpus shapes x 5 query sets,
  fingerprinting occurrence counts, visit order, `(doc, offset)` pairs and the
  prefix/suffix doc APIs, cross-checking `VisitMatchingDocs` / `MatchingDocs` /
  `LocateDocs` against each other.
- **Tile-boundary sweep** — 256 configurations: occ ∈ {0,1,2,3,7,**8,9**,15,16,
  63,**64,65**,**127,128,129**,130,255,256,257,383,384,512} x docs ∈ {1,4,97} x
  `sa_sample_rate` ∈ {1,4,8,32}. Identical to baseline, and independently
  asserted (`Count` equals the constructed occurrence count; every returned
  offset lands on the needle inside its attributed document).
- **Cross-version round-trip** — old binary writes / new binary reads and the
  reverse: identical fingerprints, and the serialized bytes are identical,
  confirming the format is untouched.
- **Skewed document lengths** — one 4 MB document among 200k tiny ones, a
  bimodal mix, and an all-tiny corpus (the adaptive bucket width's worst cases):
  identical results, no regression.
- **ASan + UBSan** clean on all of the above.
- **Existing suite**: 66 tests / 424,082 checks pass, including
  `LoadViewZeroCopyMatchesInRam` and `MmapFileRoundTrip` (the mmap path is the
  third `buildDerived()` call site), `ConcurrentQueriesMatchSingleThreaded`,
  `EmptyCorpus`, `LocateDocsSingleDoc`, `EmptyDocumentAmongNonEmpty`,
  `LocateMatchesBruteForceAcrossSampleRates`, `DocMappingLocatesPkAndOffset`.

## Not yet verified

- The full `internal/core` unit-test build (`FMIndexTest.cpp`,
  `ScalarIndexCreatorTest.cpp`) has not finished on my machine yet; everything
  above was run against the library through a standalone harness. Will report
  before this merges — flagging it rather than implying it passed.
- The `n_docs > UINT32_MAX` fallback branch is unreachable in any test (it needs
  4 billion documents); it simply leaves the table empty and takes the binary
  search.
- Tile size was tuned on ~100 MB corpora; the optimum may differ at 10 GB. The
  constant is commented with that caveat.

issue: #52699

### DIFF
--- a/internal/core/thirdparty/fmindex/index/fmindex/FMIndex.h
+++ b/internal/core/thirdparty/fmindex/index/fmindex/FMIndex.h
@@ -1,6 +1,7 @@
 // Licensed to the LF AI & Data foundation under Apache-2.0.
 // Portions translated from Lance (lance_index::scalar::fmindex), Apache-2.0.
 #pragma once
+#include <algorithm>
 #include <array>
 #include <cstddef>
 #include <cstdint>
@@ -191,6 +192,14 @@ class FMIndex {
     // LIKE '%a%' over repetitive text) makes MatchingDocs allocate GBs while
     // this stays at the caller's single bitmap. An empty pattern visits nothing
     // (as MatchingDocs).
+    //
+    // The interval is walked in TILES rather than one row at a time. Each
+    // occurrence's LF-walk is an independent chain of dependent cache misses,
+    // so walking them one by one leaves the core stalled on a single
+    // outstanding miss; advancing a tile in lock-step keeps many misses in
+    // flight at once (memory-level parallelism), the same trick CountBatch
+    // uses across patterns. The tile is fixed-size, so peak memory stays O(1)
+    // and the streaming contract above is preserved.
     template <typename Visitor>
     void
     VisitMatchingDocs(const uint8_t* pat, size_t plen, Visitor&& visit) const {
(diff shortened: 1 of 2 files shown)


czs007:

This review surfaced one low-severity gap: the boundary verification described in the PR body was run outside the repository, so the new tiled-walk and bucket-`docOf` switch points have no in-repo regression guard.

## Low (non-blocking)

- **`internal/core/src/index/FMIndexTest.cpp:528` — tile/serial boundary cases have no in-repo assertions.** The diff touches only `FMIndex.h` and `FMIndexInl.h`; no test file is added or modified. The sweep described in the PR body (occ ∈ {0,1,2,3,7,8,9,…,127,128,129,…,512} × docs ∈ {1,4,97} × `sa_sample_rate` ∈ {1,4,8,32}), the skewed doc-length corpus, and the cross-version round-trip all live in an out-of-tree standalone harness.

  Existing coverage is not zero, and worth stating precisely: `InnerMatchDedupsRepeatedOccurrences` (`FMIndexTest.cpp:528`) matches `"ab"` 251 times, which exceeds `kSerialBelow = 8` and spans two tiles (128 + 123), so the new tiled path at `FMIndex.h:239-268` is exercised; and because `doc_ckpt_` is built whenever `n_docs > 0 && text_len_ > 0 && n_docs <= UINT32_MAX` (`FMIndexInl.h:290-291`), every non-empty corpus in the existing suite already goes through the bucket branch at `FMIndexInl.h:655-668` (the `n_docs > UINT32_MAX` binary-search fallback is unreachable in practice).

  What no assertion pins down is: (a) the serial↔tiled switch point itself — occ exactly 8 versus exactly 9 (`FMIndex.h:223-224`), which `DifferentialOracleRandomBytes` only hits probabilistically via random patterns; (b) an occurrence count that is an exact multiple of `kTile` (occ = 256), where the final tile is empty or `tn` equals `kTile` exactly; and (c) the value of `ckpt_shift_` under extreme doc-length skew, since `target = (text_len_ / n_docs) * 8 + 1` (`FMIndexInl.h:293`) and every current test corpus has near-uniform doc lengths. The practical consequence is that a later change to `kTile`, `kSerialBelow`, or the bucket algorithm could regress these transitions without any deterministic test failing.

  Suggestion: land the minimal subset of the out-of-tree sweep as differential tests in `FMIndexTest.cpp` — occ ∈ {8, 9, 128, 256} crossed with a single-doc corpus and a skewed doc-length corpus — comparing against the serial/`upper_bound` reference. Separately, the PR body lists the `internal/core` unit-test build (which includes this file) under "Not yet verified"; please confirm it passes before merging. *(surfaced during verification)*

PR #52701, opened 2026-08-20
fix: enforce schema version fence on backfill

congqixia:
Spark-milvus now stamps the collection schema version it read into the backfill result JSON (schemaVersion). Activate the milvus-side fence in CommitBackfillResult: parse the stamped version and reject the commit when it no longer matches the collection's current schema version, e.g. the function was dropped or changed while the Spark job was in flight.

Without this, a result computed against an older schema is still committed against the current segments, silently advancing V3 manifests / rewriting V2 column groups with stale embeddings.

Results that carry no version (0) stay exempt so older producers keep working.

issue: #51318

### DIFF
--- a/internal/datacoord/backfill_result.go
+++ b/internal/datacoord/backfill_result.go
@@ -32,9 +32,15 @@ import (
 //
 // Reference: spark-milvus/docs/backfill-result-json-format.md
 type BackfillResult struct {
-	Success       bool                       `json:"success"`
-	CollectionID  int64                      `json:"collectionId"`
-	PartitionID   int64                      `json:"partitionId"`
+	Success      bool  `json:"success"`
+	CollectionID int64 `json:"collectionId"`
+	PartitionID  int64 `json:"partitionId"`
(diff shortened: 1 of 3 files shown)


MrPresent-Han:
Non-blocking (follow-up): the schema-version fence reads the collection version via `DescribeCollectionInternal` *before* `StartBroadcastWithResourceKeys` acquires the resource keys, so the check sits outside the broadcast's serialization boundary. A drop/alter-function committing in the window between the describe read and the broadcast can still slip through and commit stale embeddings against the current segments.

This is strictly better than the previous no-fence behavior, so it should not block this PR. Suggested follow-up: carry the expected `schemaVersion` in the `BatchUpdateManifest` message body and re-validate it at apply time (inside the same serialization boundary), or re-read the version after acquiring the resource keys.

Verified separately (non-issues): `DescribeCollectionInternal` does populate `Schema.Version` (`ToCollectionSchemaPB`), drop/alter-function bumps the version (`callAlterCollection`), and `ErrCollectionSchemaMismatch` (InputError, non-retriable) is the correct class here.

Issue #52154, opened 2026-08-04
[Bug]: Import job deadlocks with concurrent add_function_field: IndexBuilding waits for a bound index that cannot build before commit


MrPresent-Han:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```
master (feature: bump_schema_version / add_function_field)
```

### Current Behavior

A three-party circular wait wedges an import job forever when a schema-version-advancing DDL (typically `add_function_field`) lands during the import pipeline:

1. Import `IndexBuilding` waits for **all** collection indexes on the import's (sorted) segments: `dataCoord.import.waitForIndex` defaults to true, and `GetUnindexedSegments` requires every index Finished (`internal/datacoord/import_checker.go`).
2. `add_function_field` materializes its bound index meta at DDL time, so the new index immediately joins the wait set.
3. IndexInspector refuses to create that index on the import segments:
   - binlog-projection gate (pre #51768): the function-output field has no binlog — data arrives only via backfill — per-field skip;
   - schema-version gate (#51768): `segment.SchemaVersion < collection.Schema.Version` defers **all** index creation for the segment.
4. The only mechanism that can materialize the output and advance the segment schema version is bump-schema-version compaction, but its policy skips `is_importing` / invisible segments (`internal/datacoord/compaction_policy_bump_schema_version.go`).
5. `is_importing` clears only at import commit; commit needs the indexes; the indexes need backfill; backfill needs commit. Deterministic deadlock (until the import job times out and fails).

Scope notes:

- The import pipeline's own sort compaction (stats stage, `createSortCompactionTask`) does materialize function outputs and stamps the segment schema version — but only up to the schema captured at sort-task creation. It is one-shot: any DDL after that capture reopens the gap, and nothing can close it pre-commit (sort will not rerun; bump is barred by `is_importing`).
- Import segments are allocated without `SchemaVersion` (`AllocImportSegment` leaves it 0), so with sort compaction disabled the wait targets are permanently "behind" and the schema-version gate defers everything after any historical DDL on a function-bearing collection.
- Under #51768 the deferral is whole-segment, so even snapshot-era regular indexes (e.g. the vector index) cannot be created while the segment lags — any version-advancing DDL widens the blast radius, not just `add_function_field`.

### Expected Behavior

An import job must never wait for an index whose field its job never wrote. Concurrent DDL must not wedge an import job.

### Proposed Fix

[...]

PR #52706, opened 2026-08-20
fix: do not wait indexes outside import job schema snapshot (#52154)

MrPresent-Han:
issue: #52154

### What this fixes

An import job deadlocks when a schema-version-advancing DDL (typically `add_function_field`) lands during the import pipeline:

1. Import `IndexBuilding` waits for **all** collection indexes on the import segments (`GetUnindexedSegments`).
2. `add_function_field` materializes its bound index meta at DDL time, so the new index immediately joins the wait set.
3. That index cannot build before commit: the function-output field only materializes via schema-bump backfill, whose policy skips `is_importing` segments, and `is_importing` clears only at import commit — a deterministic three-party circular wait that wedges the job until timeout.

### How

Judge the wait against the job's own schema snapshot (`ImportJob.Schema`, captured at job creation, persisted, immutable): a new helper `filterUnindexedBySnapshotSchema` drops segments whose missing indexes are all on fields absent from the snapshot, applied at both wait sites — `checkIndexBuildingJob` (the wedge) and `getIndexBuildingProgress` (its progress view). `GetUnindexedSegments` itself is untouched.

After commit the segment becomes an ordinary stale segment: schema-bump backfill materializes the new field and advances the segment schema version, and the bound index then builds through the normal inspector path. No new convergence machinery.

### Scope notes

This breaks the cycle at its only semantically wrong edge — an import waiting on an index for a field its job never wrote. Residual hardening is tracked as follow-ups per the issue: per-field precision for the schema-version index-creation gate (#51768), and stamping import segments with the job snapshot schema version at alloc.

### DIFF
--- a/internal/datacoord/import_checker.go
+++ b/internal/datacoord/import_checker.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/milvus-io/milvus/internal/datacoord/allocator"
 	"github.com/milvus-io/milvus/internal/datacoord/broker"
+	"github.com/milvus-io/milvus/internal/metastore/model"
 	"github.com/milvus-io/milvus/internal/streamingcoord/server/broadcaster"
 	"github.com/milvus-io/milvus/internal/util/importutilv2"
 	"github.com/milvus-io/milvus/pkg/v3/metrics"
@@ -36,6 +37,7 @@ import (
 	"github.com/milvus-io/milvus/pkg/v3/util/funcutil"
 	"github.com/milvus-io/milvus/pkg/v3/util/merr"
 	"github.com/milvus-io/milvus/pkg/v3/util/tsoutil"
+	"github.com/milvus-io/milvus/pkg/v3/util/typeutil"
 )
 
 type ImportChecker interface {
@@ -60,14 +62,23 @@ type importCheckerHooks struct {
 	isReplicatingCluster func(ctx context.Context) (bool, error)
 }
 
+// scopedIndexCreator lets the import checker drive index creation for a segment
+// with an era attestation (fields whose data the import itself materialized),
+// routed onto the inspector's single creation goroutine. Implemented by
+// indexInspector — the sole owner of the creation mechanics.
+type scopedIndexCreator interface {
+	enqueueScopedCreation(segmentID int64, attestedFields typeutil.Set[int64])
+}
+
 type importChecker struct {
-	ctx        context.Context
-	meta       *meta
-	broker     broker.Broker
-	alloc      allocator.Allocator
-	importMeta ImportMeta
-	ci         CompactionInspector
-	handler    Handler
+	ctx          context.Context
+	meta         *meta
+	broker       broker.Broker
+	alloc        allocator.Allocator
+	importMeta   ImportMeta
+	ci           CompactionInspector
+	handler      Handler
+	indexCreator scopedIndexCreator
 
 	hooks importCheckerHooks
 
@@ -82,18 +93,20 @@ func NewImportChecker(ctx context.Context,
 	importMeta ImportMeta,
 	ci CompactionInspector,
 	handler Handler,
+	indexCreator scopedIndexCreator,
 	hooks importCheckerHooks,
 ) ImportChecker {
 	return &importChecker{
-		ctx:        ctx,
-		meta:       meta,
-		broker:     broker,
-		alloc:      alloc,
-		importMeta: importMeta,
-		ci:         ci,
-		handler:    handler,
-		hooks:      hooks,
-		closeChan:  make(chan struct{}),
+		ctx:          ctx,
+		meta:         meta,
+		broker:       broker,
+		alloc:        alloc,
+		importMeta:   importMeta,
+		ci:           ci,
+		handler:      handler,
+		indexCreator: indexCreator,
+		hooks:        hooks,
+		closeChan:    make(chan struct{}),
 	}
 }
 
@@ -464,17 +477,21 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
 		targetSegmentIDs = originSegmentIDs
 	}
 
-	healthySegments := c.meta.GetSegments(targetSegmentIDs, isSegmentHealthy)
-	unindexed := c.meta.indexMeta.GetUnindexedSegments(job.GetCollectionID(), healthySegments)
-	if Params.DataCoordCfg.WaitForIndex.GetAsBool() && len(unindexed) > 0 && !importutilv2.IsL0Import(job.GetOptions()) {
-		for _, segmentID := range unindexed {
-			select {
-			case getBuildIndexChSingleton() <- segmentID: // accelerate index building:
-			default:
-			}
+	if Params.DataCoordCfg.WaitForIndex.GetAsBool() && !importutilv2.IsL0Import(job.GetOptions()) {
+		snapshotFields := importJobSnapshotFields(job)
+		healthySegments := c.meta.GetSegments(targetSegmentIDs, isSegmentHealthy)
+		unindexed := c.meta.indexMeta.GetUnindexedSegments(job.GetCollectionID(), healthySegments)
+		kept, needsCreation, excluded := filterUnindexedBySnapshotSchema(snapshotFields, job.GetCollectionID(), c.meta.indexMeta, unindexed)
+		if len(kept) > 0 {
+			c.ensureImportSegmentIndexes(job, snapshotFields, needsCreation)
+			log.Debug(c.ctx, "waiting for import segments building index...", mlog.Int64s("unindexed", kept))
+			return
 		}
-		log.Debug(c.ctx, "waiting for import segments building index...", mlog.Int64s("unindexed", unindexed))
-		return
+		if len(excluded) > 0 {
+			log.Info(c.ctx, "import job proceeds without waiting for indexes on fields outside its schema snapshot; they build after commit via schema-bump backfill",
+				mlog.Int64s("excludedIndexIDs", lo.Map(excluded, func(index *model.Index, _ int) int64 { return index.IndexID })),
+				mlog.Int64s("excludedFieldIDs", lo.Map(excluded, func(index *model.Index, _ int) int64 { return index.FieldID })))
+		}
 	}
 	buildIndexDuration := job.GetTR().RecordSpan()
 	metrics.ImportJobLatency.WithLabelValues(metrics.ImportStageBuildIndex).Observe(float64(buildIndexDuration.Milliseconds()))
@@ -492,6 +509,52 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
 	LogResultSegmentsInfo(job.GetJobID(), c.meta, targetSegmentIDs)
 	log.Info(c.ctx, "import job indexes built, transitioned to Uncommitted",
 		mlog.Bool("autoCommit", job.GetAutoCommit()))
+}
+
+// ensureImportSegmentIndexes drives index creation for segments whose snapshot-scoped
+// index has no SegmentIndex row yet (needsCreation), routing by whether the inspector's
+// own periodic scan can handle it:
+//   - common path — the inspector will build the index on its next scan; just send the
+//     cheap acceleration nudge (bare segment ID), no clones, no attestation.
+//   - wedge path — the segment's schema version lags the collection's while the
+//     collection has functions, so the inspector whole-segment-defers it and the lag
+//     cannot heal before commit (sort is one-shot, schema-bump skips is_importing
+//     segments). Route an attested request onto the inspector's single creation
+//     goroutine, marking the job's snapshot fields as materialized.
+//
+// The wedge test needs the CURRENT collection schema version (not the job snapshot's).
+// It is read once per call from the local meta cache (non-blocking): an unresolved
+// collection only skips the wedge optimization for this tick (the segment falls to the
+// nudge and retries next tick once the cache warms), so it must not block the shared
+// state-machine loop with a rootcoord lazy-load.
+func (c *importChecker) ensureImportSegmentIndexes(job ImportJob, snapshotFields typeutil.Set[int64], needsCreation []int64) {
+	if len(needsCreation) == 0 {
+		return
+	}
+	hasFunctions, collectionVersion := false, int32(0)
+	if collection := c.meta.GetCollection(job.GetCollectionID()); collection != nil && collection.Schema != nil {
+		hasFunctions = len(collection.Schema.GetFunctions()) > 0
+		collectionVersion = collection.Schema.GetVersion()
+	} else {
+		mlog.Warn(c.ctx, "cannot resolve collection schema from cache, skip wedge index routing this tick",
+			mlog.FieldJobID(job.GetJobID()), mlog.FieldCollectionID(job.GetCollectionID()),
+			mlog.Int64s("segments", needsCreation))
+	}
+	for _, segmentID := range needsCreation {
+		if c.indexCreator != nil && hasFunctions {
+			// GetSegment is a light read (no index clones); reached only for
+			// function-bearing collections, and only these segments enqueue.
+			if segment := c.meta.GetSegment(c.ctx, segmentID); segment != nil &&
+				segment.GetSchemaVersion() < collectionVersion {
+				c.indexCreator.enqueueScopedCreation(segmentID, snapshotFields)
+				continue
+			}
+		}
+		select {
+		case getBuildIndexChSingleton() <- segmentID: // accelerate index building
+		default:
+		}
+	}
 }
 
 // checkUncommittedJob handles jobs in the Uncommitted state.
--- a/internal/datacoord/import_util.go
+++ b/internal/datacoord/import_util.go
@@ -30,8 +30,10 @@ import (
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/msgpb"
+	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/datacoord/allocator"
 	"github.com/milvus-io/milvus/internal/datacoord/session"
+	"github.com/milvus-io/milvus/internal/metastore/model"
 	"github.com/milvus-io/milvus/internal/storage"
 	"github.com/milvus-io/milvus/internal/util/importutilv2"
 	"github.com/milvus-io/milvus/pkg/v3/common"
@@ -545,6 +547,69 @@ func getStatsProgress(ctx context.Context, jobID int64, importMeta ImportMeta, m
 	return float32(doneCnt) / float32(len(targetSegmentIDs))
 }
 
+// importJobSnapshotFields returns the field IDs of the job's schema snapshot
+// (struct sub-fields included), or nil when the job carries no snapshot.
+func importJobSnapshotFields(job ImportJob) typeutil.Set[int64] {
+	schema := job.GetSchema()
+	if schema == nil {
+		return nil
+	}
+	return typeutil.NewSet(lo.Map(typeutil.GetAllFieldSchemas(schema), func(field *schemapb.FieldSchema, _ int) int64 {
+		return field.GetFieldID()
+	})...)
+}
+
+// filterUnindexedBySnapshotSchema narrows the import's index wait to the job's own
+// schema snapshot. An index on a field absent from the snapshot is one the job never
+// wrote (e.g. added by a concurrent DDL); it can only build after commit + schema-bump
+// backfill (StorageV3 segments), so waiting on it deadlocks the job (#52154). Returns:
+//   - kept: segments still to wait on — a snapshot-scoped index is not yet Finished;
+//   - needsCreation: segments whose snapshot-scoped index has no SegmentIndex row yet,
+//     i.e. creation must be driven (a segment whose row exists but is still building is
+//     kept-but-not-needsCreation, so the checker does not re-drive an in-flight build);
+//   - excluded: the out-of-snapshot indexes dropped from the wait, for the caller to log.
+//
+// snapshotFields is passed in (computed once by the caller) to avoid recomputing it.
+func filterUnindexedBySnapshotSchema(snapshotFields typeutil.Set[int64], collectionID int64, indexMeta *indexMeta, unindexed []int64) ([]int64, []int64, []*model.Index) {
+	if len(unindexed) == 0 {
+		return unindexed, nil, nil
+	}
+	if snapshotFields == nil {
+		// No snapshot to scope by: preserve the drive-all-unindexed behavior.
+		return unindexed, unindexed, nil
+	}
+	snapshotIndexes, excluded := lo.FilterReject(indexMeta.GetIndexesForCollection(collectionID, ""), func(index *model.Index, _ int) bool {
+		return snapshotFields.Contain(index.FieldID)
+	})
+	if len(snapshotIndexes) == 0 {
+		return nil, nil, excluded
+	}
+	// Single batch read of state-only objects; avoids per-segment locking and
+	// SegmentIndex deep clones (IndexFileKeys) on the checker hot path.
+	states := indexMeta.getSegmentsIndexStates(collectionID, unindexed)
+	kept := make([]int64, 0, len(unindexed))
+	needsCreation := make([]int64, 0, len(unindexed))
+	for _, segmentID := range unindexed {
+		segmentStates := states[segmentID]
+		keep, create := false, false
+		for _, index := range snapshotIndexes {
+			state, ok := segmentStates[index.IndexID]
+			if !ok {
+				create, keep = true, true // no row yet -> drive creation, and wait
+			} else if state.GetState() != commonpb.IndexState_Finished {
+				keep = true // row exists but building -> wait, do not re-drive
+			}
+		}
+		if keep {
+			kept = append(kept, segmentID)
+		}
+		if create {
+			needsCreation = append(needsCreation, segmentID)
+		}
+	}
+	return kept, needsCreation, excluded
+}
+
 func getIndexBuildingProgress(ctx context.Context, jobID int64, importMeta ImportMeta, meta *meta) float32 {
 	job := importMeta.GetJob(ctx, jobID)
 	if !Params.DataCoordCfg.WaitForIndex.GetAsBool() {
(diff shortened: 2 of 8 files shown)


internal/datacoord/import_util.go
```diff
@@ -545,6 +547,34 @@ func getStatsProgress(ctx context.Context, jobID int64, importMeta ImportMeta, m
 	return float32(doneCnt) / float32(len(targetSegmentIDs))
 }
 
+// filterUnindexedBySnapshotSchema drops segments that only miss indexes on fields
+// absent from the job's schema snapshot: the job never wrote such fields (e.g. one
+// added by a concurrent DDL) and their indexes can only build after the import
+// commits and schema-bump backfill materializes the field, so waiting on them
```
liliu-z: The comment states the dropped indexes build after commit once schema-bump backfill materializes the field, but that recovery only happens on StorageV3 segments: staleFlushedSegments skips any segment with `StorageVersion < StorageV3 || ManifestPath == ""` (both are required), and import segments are allocated StorageV2 unless `common.storage.useLoonFFI` is enabled (default false). On a default deployment the segment's SchemaVersion therefore never advances, canCreateIndexForSegment keeps deferring, and the job now reports success with the bound index permanently unbuilt — with only a rate-limited warning from the bump policy and no record of which segments/indexes were dropped from the wait. Qualify the comment and log the dropped segment/index IDs at Info so the residual under-indexed state is greppable.

internal/datacoord/import_util.go
```diff
@@ -545,6 +547,34 @@ func getStatsProgress(ctx context.Context, jobID int64, importMeta ImportMeta, m
...
+	snapshotFields := typeutil.NewSet(lo.Map(typeutil.GetAllFieldSchemas(schema), func(field *schemapb.FieldSchema, _ int) int64 {
+		return field.GetFieldID()
+	})...)
+	snapshotIndexes := lo.Filter(indexMeta.GetIndexesForCollection(job.GetCollectionID(), ""), func(index *model.Index, _ int) bool {
+		return snapshotFields.Contain(index.FieldID)
+	})
+	return lo.Filter(unindexed, func(segmentID int64, _ int) bool {
+		segmentIndexes := indexMeta.GetSegmentIndexes(job.GetCollectionID(), segmentID)
+		for _, index := range snapshotIndexes {
+			segmentIndex, ok := segmentIndexes[index.IndexID]
```
liliu-z: The filter only unblocks the job when the snapshot-scoped indexes already have segment index entries. canCreateIndexForSegment defers every index — snapshot-scoped ones included — on a segment whose SchemaVersion lags the collection's when the collection has functions, and the sorted segment is stamped with the schema version captured at sort-compaction task creation. If add_function_field lands between the sorted segment's creation and the inspector's first pass over it, no snapshot-scoped segment index is ever created, the filter's `!ok` branch keeps the segment in the wait set, and the job wedges in IndexBuilding exactly as #52154. That gate is the per-field precision follow-up (#51768) named in the PR description, but it means this fix closes the cycle only for orderings where the snapshot indexes got created first.

internal/datacoord/import_util.go
```diff
@@ -545,6 +547,34 @@ func getStatsProgress(ctx context.Context, jobID int64, importMeta ImportMeta, m
...
+		return unindexed
+	}
+	snapshotFields := typeutil.NewSet(lo.Map(typeutil.GetAllFieldSchemas(schema), func(field *schemapb.FieldSchema, _ int) int64 {
+		return field.GetFieldID()
+	})...)
+	snapshotIndexes := lo.Filter(indexMeta.GetIndexesForCollection(job.GetCollectionID(), ""), func(index *model.Index, _ int) bool {
+		return snapshotFields.Contain(index.FieldID)
+	})
+	return lo.Filter(unindexed, func(segmentID int64, _ int) bool {
+		segmentIndexes := indexMeta.GetSegmentIndexes(job.GetCollectionID(), segmentID)
```
buqian-zilliz: **[P2] Redundant per-segment index reads and deep clones on the import checker hot path**

`filterUnindexedBySnapshotSchema` is appended to `GetUnindexedSegments` at both wait sites, but it recomputes work that `GetUnindexedSegments` already did: it calls `GetIndexesForCollection` a second time (another full deep clone of every collection index) and, for each still-unindexed segment, calls `GetSegmentIndexes`, which takes `fieldIndexLock.RLock()` and deep-clones every `model.SegmentIndex` (including `IndexFileKeys` via `common.CloneStringList`) even though the loop only reads `IndexState`. At production scale with many segments/partitions this adds N+1 lock acquisitions and allocations on a path that runs each `importChecker` tick and each `GetJobProgress` call, and it is paid even when the result is discarded (`WaitForIndex` disabled or `IsL0Import`). Consider reusing the per-segment states already computed by `getSegmentsIndexStates` (or a single batch read), hoisting the schema-field set, and skipping the filter when the caller will not consume the result.

internal/datacoord/import_util.go
MrPresent-Han: Verified against the code — the default-deployment premise can't arise under a stable config:

- `add_function_field` is gated on loon: `validateAddFunctionRequiresStorageV3` (internal/proxy/task.go:235) rejects the DDL outright when `common.storage.useLoonFFI=false`, so a deployment whose import segments are StorageV2 cannot create the bound index in the first place.
- On the only config where the DDL can run (loon enabled), non-L0 import segments are allocated StorageV3 from the start (`importStorageVersion`, internal/datacoord/import_util.go), so `staleFlushedSegments` does pick them up after commit and the recovery described in the comment holds.

The real residual is the refreshable-param window: an import that allocated V2 segments before `useLoonFFI` flips to true mid-flight, with the DDL landing afterwards — those segments then sit under-indexed, the same exposure any pre-existing stale V2 segment has w.r.t. bump backfill, not specific to this filter. If you think that window warrants it, I can add a "(StorageV3 segments)" qualifier to the comment plus an Info log of the dropped segment/index IDs so the residual state is greppable.

internal/datacoord/import_util.go
MrPresent-Han: Correct — this fix releases the job only for orderings where the snapshot-scoped indexes could be created before the DDL bumped the collection version; the DDL-during-sort window keeps the whole-segment deferral wedge. Plan, in line with the residual-hardening list:

1. **Follow-up #52709**: narrow the #51768 gate from whole-segment to per-index — defer only function-output field indexes. That is the gate's own semantic boundary: backfill changes the read value only for function outputs (computed), while plain added fields materialize to the same null/default they already read as, so an early index stays consistent. Snapshot-era ordinary indexes then build and this filter releases the job. (Import executes snapshot-schema functions — `RunEmbeddingFunction` in datanode importv2 — so snapshot-era outputs are materialized in import segments.)
2. **Last-resort escape** for the sub-case that remains even after #52709 (snapshot-era function-output index on a lagging importing segment): `dataCoord.import.waitForIndex` is `refreshable:"true"` and read live each checker tick — hot-flipping it commits the job, clears `is_importing`, bump backfill advances the version, and every index converges post-commit with no correctness loss. Will be documented as runbook.

internal/datacoord/import_util.go
MrPresent-Han: Fixed in the amended commit (f81392c132):

- The filter now does a single lightweight batch read via `getSegmentsIndexStates` — state-only objects, one lock pass, no per-segment `SegmentIndex` deep clones (`IndexFileKeys` included) — and early-exits when no snapshot-scoped index exists.
- `checkIndexBuildingJob` computes `GetSegments` / `GetUnindexedSegments` / the filter only inside the `WaitForIndex && !IsL0Import` gate, so nothing is paid when the result would be discarded. (`getIndexBuildingProgress` already early-returns on `WaitForIndex` disabled.)

The remaining second `GetIndexesForCollection` call is kept deliberately: deduping it would require merging the filter into `GetUnindexedSegments`, which we intentionally leave untouched; its clone cost is bounded by the collection's index count (single digits), not by segment count.

internal/datacoord/import_checker.go
```diff
@@ -464,17 +464,20 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
...
-	unindexed := c.meta.indexMeta.GetUnindexedSegments(job.GetCollectionID(), healthySegments)
-	if Params.DataCoordCfg.WaitForIndex.GetAsBool() && len(unindexed) > 0 && !importutilv2.IsL0Import(job.GetOptions()) {
-		for _, segmentID := range unindexed {
-			select {
-			case getBuildIndexChSingleton() <- segmentID: // accelerate index building:
-			default:
+	if Params.DataCoordCfg.WaitForIndex.GetAsBool() && !importutilv2.IsL0Import(job.GetOptions()) {
+		healthySegments := c.meta.GetSegments(targetSegmentIDs, isSegmentHealthy)
+		unindexed := c.meta.indexMeta.GetUnindexedSegments(job.GetCollectionID(), healthySegments)
+		unindexed = filterUnindexedBySnapshotSchema(job, c.meta.indexMeta, unindexed)
```
liliu-z: When the filter drops segments from the wait set the job proceeds to Uncommitted with no record of which indexes were skipped — the only log left is the Debug line listing the segments still waiting, and `filterUnindexedBySnapshotSchema` logs nothing at all. Since the point of the change is that a job now legitimately completes with unbuilt indexes on post-snapshot fields, an operator investigating "import finished but my new field's index is missing" has no breadcrumb tying it to this filter. A single Info log naming the excluded index IDs / field IDs on the transition would make the new semantics diagnosable.

MrPresent-Han: Added in 0fecf4fa9d:

- `filterUnindexedBySnapshotSchema` now also returns the out-of-snapshot indexes it excluded from the wait (`lo.FilterReject`), and `checkIndexBuildingJob` logs them at Info on the transition — job ID, `excludedIndexIDs`, `excludedFieldIDs` — with the message stating they build after commit via schema-bump backfill. The log fires only when the wait set was actually released by the exclusion (no noise when everything was already built), and the progress path stays silent to avoid poll spam.
- The helper's doc comment now carries the StorageV3 qualifier from the other thread.

internal/datacoord/import_checker.go
```diff
@@ -464,17 +477,20 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
...
-			select {
-			case getBuildIndexChSingleton() <- segmentID: // accelerate index building:
-			default:
-			}
+	if Params.DataCoordCfg.WaitForIndex.GetAsBool() && !importutilv2.IsL0Import(job.GetOptions()) {
+		healthySegments := c.meta.GetSegments(targetSegmentIDs, isSegmentHealthy)
+		unindexed := c.meta.indexMeta.GetUnindexedSegments(job.GetCollectionID(), healthySegments)
+		unindexed, excluded := filterUnindexedBySnapshotSchema(job, c.meta.indexMeta, unindexed)
+		if len(unindexed) > 0 {
+			c.ensureImportSegmentIndexes(job, unindexed)
```
buqian-zilliz: **[P2] Concurrent index-creation race between import checker and index inspector**

`ensureImportSegmentIndexes` now calls `indexInspector.createIndexesForSegmentWithScope` directly from the import checker's state-machine loop. This runs concurrently with `indexInspector.createIndexForSegmentLoop`, which periodically scans flushed/unindexed segments (including the same import segments) and calls `createIndexesForSegment`. Both paths do a check-then-act: they read `GetSegmentIndexes` to decide an index is missing, then call `createIndexForSegment` -&gt; `AddSegmentIndex`, whose only serialization is `keyLock.Lock(buildID)` on a freshly allocated build ID, so the two goroutines are not serialized per `(segment, index)`. If both observe the index as missing, they allocate different build IDs and both persist a `CreateSegmentIndex` catalog record and enqueue a duplicate index-build task for the same `(segment, index)`, leaving two catalog entries (and, on restart, two entries in `segmentBuildInfo`). Previously index creation was funneled through the single inspector goroutine because the checker only nudged `getBuildIndexChSingleton()`, so this adds a new concurrent creation path. Consider serializing creation on the `(segment, index)` key in `AddSegmentIndex`/`createIndexesForSegment`, or routing the checker's request through the existing inspector loop rather than calling the creator directly.

internal/datacoord/server.go
```diff
@@ -344,7 +344,15 @@ func (s *Server) initDataCoord() error {
 
 	s.importInspector = NewImportInspector(s.ctx, s.meta, s.importMeta, s.globalScheduler)
 
-	s.importChecker = NewImportChecker(s.ctx, s.meta, s.broker, s.allocator, s.importMeta, s.compactionInspector, s.handler, importCheckerHooks{
+	// the index inspector is constructed later in the boot sequence; resolve it
+	// lazily and guard nil so a typed-nil pointer never leaks into the interface.
+	getIndexCreator := func() segmentIndexCreator {
```
buqian-zilliz: **[P3] Misleading boot-order comment and dead lazy index-creator indirection**

The comment says `the index inspector is constructed later in the boot sequence`, but in `initDataCoord` `initIndexInspector(storageCli)` (line 329) runs before `NewImportChecker(...)` (line 355), and `importChecker.Start()` is invoked even later. The `getIndexCreator` closure and its nil guard are therefore dead defensive code in production: `s.indexInspector` is always non-nil by the time the checker loop runs. Worse, if the nil branch were ever reached, the fallback in `ensureImportSegmentIndexes` would only perform a non-blocking send to `getBuildIndexChSingleton()` with no consumer (the inspector is not started), silently no-op and leave the job stuck in IndexBuilding. Either inject the inspector directly at construction time or fix the comment and clarify that the lazy resolution is intentionally defensive.

internal/datacoord/import_checker.go
```diff
@@ -494,6 +510,39 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
...
+		for _, segmentID := range unindexed {
+			select {
+			case getBuildIndexChSingleton() <- segmentID: // accelerate index building:
+			default:
+			}
+		}
+		return
+	}
+	attested := importJobSnapshotFields(job)
+	for _, segmentID := range unindexed {
```
liliu-z: `ensureImportSegmentIndexes` fans out over every still-unindexed segment on each import state-machine tick, and each iteration costs two deep-cloning meta reads (`GetIndexesForCollection` clones every `model.Index`, `GetSegmentIndexes` clones every `model.SegmentIndex` including `IndexFileKeys`) plus a `handler.GetCollection` per index. The import checker ticks at `ImportCheckIntervalHigh` (2s) while the index inspector does the same scan at `TaskCheckInterval` (60s default), so this adds ~30x the clone traffic and runs it on the single goroutine that also drives every other import job's state transitions. Consider gating the call on the attestation condition, or firing it on the ->IndexBuilding transition with a slow backstop.

internal/datacoord/import_util.go
```diff
@@ -545,6 +547,51 @@ func getStatsProgress(ctx context.Context, jobID int64, importMeta ImportMeta, m
 	return float32(doneCnt) / float32(len(targetSegmentIDs))
 }
 
+// filterUnindexedBySnapshotSchema drops segments that only miss indexes on fields
+// absent from the job's schema snapshot: the job never wrote such fields (e.g. one
+// added by a concurrent DDL) and their indexes can only build after the import
+// commits and schema-bump backfill (StorageV3 segments) materializes the field, so
+// waiting on them deadlocks the job (#52154). The out-of-snapshot indexes excluded
+// from the wait are returned so callers can surface them.
+// importJobSnapshotFields returns the field IDs of the job's schema snapshot
```
liliu-z, line 550: The doc comment at this line is misattached and documents a different symbol than the one it precedes, misleading readers about the function's contract.

internal/datacoord/import_checker.go
MrPresent-Han: Fixed in eec9ad0875 by restoring single-owner creation. The checker no longer calls `createIndexesForSegmentWithScope` directly; the inspector gains a buffered `scopedCreateCh` and a 4th `select` case in `createIndexForSegmentLoop`, and the checker only *enqueues* `{segmentID, attestedFields}` (non-blocking, drop-on-full, re-enqueued next tick). All creation — periodic scan, notify, buildIndexCh nudge, and the attested requests — now runs on the inspector's single goroutine, so the (segment,index) check-then-act can no longer interleave and `AddSegmentIndex`'s single-creator contract holds again. Agreed on your point 1: `AddSegmentIndex` was never the defect — the second concurrent creation path I added was.

internal/datacoord/server.go
MrPresent-Han: Fixed in eec9ad0875. `initIndexInspector` (server.go, earlier in the boot sequence) already sets `s.indexInspector` before `NewImportChecker` is called, so the lazy `getIndexCreator` closure + nil guard were indeed dead. Removed them and inject `s.indexInspector` directly at construction; the misleading "constructed later" comment is gone.

internal/datacoord/import_checker.go
MrPresent-Han: Fixed in eec9ad0875. The channel alone would only relocate the clone traffic, not cut it, so the checker now routes by condition instead of driving creation every tick:

- common path (segment schema version == collection version) → the inspector's own scan builds the index; the checker just sends the cheap bare-segmentID `buildIndexCh` nudge, exactly as before my change — no `GetIndexesForCollection`/`GetSegmentIndexes` clones.
- wedge path (collection has functions AND segment schema version < collection version) → the only case the inspector whole-segment-defers and the lag cannot heal pre-commit; the checker enqueues the attested request, and the (rare) clone work happens on the inspector side.

The gate reads the current collection version via one `handler.GetCollection` hoisted per call (not per segment); `GetSegment` for the per-segment version is a light read reached only for function-bearing collections. Common-path checker cost is back to ~zero.

internal/datacoord/import_util.go
MrPresent-Han: Fixed in eec9ad0875 — the `importJobSnapshotFields` extraction had floated `filterUnindexedBySnapshotSchema`'s doc above the wrong function. Each function now carries its own comment.

internal/datacoord/import_util.go
MrPresent-Han: Update: this is no longer a follow-up — #52709's fix is now the second commit of this same PR (era-attested creation). The DDL-during-sort ordering you described is exactly what it closes: the checker attests the job's snapshot fields as materialized and routes creation for the lagging importing segment onto the inspector, so the snapshot-scoped indexes build before commit instead of being whole-segment-deferred. See the race/perf threads above for the mechanism.

internal/datacoord/import_checker.go
```diff
@@ -494,6 +510,42 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
...
+//
+// The current collection schema version (not the job snapshot's) defines the lag;
+// it is fetched once per call, not per segment.
+func (c *importChecker) ensureImportSegmentIndexes(job ImportJob, unindexed []int64) {
+	hasFunctions, collectionVersion := false, int32(0)
+	if collection, err := c.handler.GetCollection(c.ctx, job.GetCollectionID()); err == nil && collection != nil && collection.Schema != nil {
+		hasFunctions = len(collection.Schema.GetFunctions()) > 0
+		collectionVersion = collection.Schema.GetVersion()
+	}
+	attested := importJobSnapshotFields(job)
```
buqian-zilliz: **[P3] Recompute `importJobSnapshotFields(job)` twice per job per tick**

`checkIndexBuildingJob` calls `filterUnindexedBySnapshotSchema(job, ...)`, which builds the job's snapshot field set via `importJobSnapshotFields(job)`, and then — when `unindexed` is non-empty — calls `ensureImportSegmentIndexes(job, ...)`, which builds the exact same set again via `attested := importJobSnapshotFields(job)`. Each call allocates a `typeutil.Set[int64]` and iterates `typeutil.GetAllFieldSchemas`. This runs on the import checker's 2s state-machine tick for every job stuck in `IndexBuilding`. Consider computing the snapshot field set once in `checkIndexBuildingJob` and passing it into both helpers.

internal/datacoord/import_checker.go
```diff
@@ -494,6 +510,42 @@ func (c *importChecker) checkIndexBuildingJob(job ImportJob) {
...
+//     collection has functions, so the inspector whole-segment-defers it and the
+//     lag cannot heal before commit (sort is one-shot, schema-bump skips
+//     is_importing segments). Route an attested request onto the inspector's single
+//     creation goroutine, marking the job's snapshot fields as materialized.
+//
+// The current collection schema version (not the job snapshot's) defines the lag;
+// it is fetched once per call, not per segment.
+func (c *importChecker) ensureImportSegmentIndexes(job ImportJob, unindexed []int64) {
+	hasFunctions, collectionVersion := false, int32(0)
+	if collection, err := c.handler.GetCollection(c.ctx, job.GetCollectionID()); err == nil && collection != nil && collection.Schema != nil {
```
buqian-zilliz: **[P3] `handler.GetCollection` can block the shared state-machine loop on a cache miss**

`ensureImportSegmentIndexes` now calls `c.handler.GetCollection` from the import checker's single state-machine loop. On a cache miss (for example right after a datacoord restart), `ServerHandler.GetCollection` performs `loadCollectionFromRootCoord` under a 10s timeout with 5 retry attempts, so the call can block for up to roughly 50s while the collection is lazily loaded. Because this runs on the one shared loop that also drives every other import job's state transitions, a slow or blackholed rootcoord stalls progress for all jobs, not just this one. The inspector already accepts this lazy-load behavior on its own dedicated goroutine, but the checker loop is shared. The lag/attestation decision only needs the locally cached schema, so using the non-blocking `meta.GetCollection` (which skips the wedge optimization on a cache miss and lets the inspector's own goroutine handle the load) would avoid blocking the shared loop.

internal/datacoord/index_inspector.go
```diff
@@ -64,6 +75,7 @@ func newIndexInspector(
 		cancel:                    cancel,
 		meta:                      meta,
 		notifyIndexChan:           notifyIndexChan,
+		scopedCreateCh:            make(chan scopedCreateRequest, 256),
```
liliu-z, line 78: For a schema-lagging import segment the scoped channel is the only route to index creation — the inspector's periodic scan and the getBuildIndexChSingleton nudge both hit the version gate and defer — yet enqueueScopedCreation drops silently when the 256-slot buffer is full, and the checker re-enqueues the same stable prefix every 2s tick (meta.GetSegments and lo.Without both preserve input order, so a single task's segment list has a fixed prefix). A job with more than 256 wedged segments in one task therefore only reaches its tail after the head's builds finish. It converges rather than deadlocks, and the job must wait for those builds anyway, so the practical cost is wave-boundary latency, not a wedge; dedup'ing a pending set keyed by segment ID (channel as wake-up only), or at least logging/counting the drop, would remove the dependence on buffer size.

internal/datacoord/import_checker.go
liliu-z: ensureImportSegmentIndexes swallows a GetCollection failure: on error it leaves hasFunctions=false and every segment falls through to the plain getBuildIndexChSingleton nudge — exactly the route the inspector's version gate defers — so while the collection cannot be resolved the job keeps waiting with no log saying why, in the one code path added to break a deadlock. The inspector logs a Warn for the same unresolvable-schema case, and this branch is untested: the suite handler returns collectionInfo{ID: collID} with a nil Schema, so TestEnsureImportSegmentIndexesCommonPath actually exercises the unresolved-schema fallback rather than the function-less collection its comment claims.

internal/datacoord/import_checker.go
MrPresent-Han: Fixed in 91809c18e0. `checkIndexBuildingJob` now computes `importJobSnapshotFields(job)` once and passes it into both `filterUnindexedBySnapshotSchema` and `ensureImportSegmentIndexes`; neither recomputes it.

internal/datacoord/import_checker.go
MrPresent-Han: Fixed in 91809c18e0. `ensureImportSegmentIndexes` now reads the collection from the local meta cache (`c.meta.GetCollection`, non-blocking) instead of `c.handler.GetCollection`, so a cache miss can no longer trigger a rootcoord lazy-load on the shared state-machine loop. On an unresolved collection it logs a Warn and skips the wedge routing for this tick — the segment falls to the nudge and retries next tick once the cache warms (which the inspector's own lazy-load populates). The two gates (inspector defer / checker wedge) read the same `m.collections`, so this also keeps them same-source on a hit.

internal/datacoord/index_inspector.go
MrPresent-Han: Addressed in 91809c18e0 by removing the per-tick re-enqueue of already-scheduled segments, which is what filled the buffer. `filterUnindexedBySnapshotSchema` now returns a `needsCreation` subset — segments whose snapshot-scoped index has NO SegmentIndex row yet — computed from the state batch it already reads. The checker only enqueues that subset, not all of `unindexed`: once creation is driven and the row exists, the segment is `kept` (still waited on) but no longer `needsCreation`, so it stops being re-enqueued while its build is in flight. That collapses the steady-state enqueue traffic to just the not-yet-created segments, so the fixed prefix of a >256-segment task no longer re-fills the buffer every tick. I kept the bounded drop-on-full channel (a full-set pending map keyed by segment ID would also remove the buffer-size dependence for in-queue duplicates, but with the re-enqueue tail gone the remaining exposure is a single burst > buffer, which converges on the next tick).

internal/datacoord/import_checker.go
MrPresent-Han: Both fixed in 91809c18e0.

1. Swallowed unresolved schema: `ensureImportSegmentIndexes` now logs a Warn ("cannot resolve collection schema from cache, skip wedge index routing this tick", with job/collection IDs and the affected segments) when the collection can't be resolved, so the transient cold-cache window is greppable instead of silent. (It is transient — the inspector's own path lazy-loads the collection and the next tick resolves it; it is not a permanent wedge.)

2. Fake common-path test: corrected. `TestEnsureImportSegmentIndexesCommonPath` now seeds a real function-less collection (Version set, no Functions) via `meta.AddCollection` and asserts no scoped enqueue, so it exercises the actual common path rather than the nil-schema fallback. Added `TestEnsureImportSegmentIndexesUnresolvedSchema` for the unresolved branch. The wedge test also moved to `meta.AddCollection` (matching the switch to `meta.GetCollection`) and now asserts the row-missing → needsCreation → scoped-enqueue path.

MrPresent-Han:
Closing in favor of a redesign. Review discussion (thanks @liliu-z @buqian-zilliz) converged on a cleaner direction: carry per-field provenance in the schema itself — `FieldSchema.added_schema_version`, stamped by RootCoord at DDL time — so the index inspector judges materialization directly (`field.addedVersion <= segment.SchemaVersion ⟺ column data ready`), replacing both the whole-segment version deferral and the attestation/scoped-creation machinery introduced here. The wait-scope part (import waits only on its snapshot-scoped indexes) carries over to the new PR.

A new PR for #52154 will follow, gated on a milvus-proto change adding the field.

Issue #52216, opened 2026-08-06
[Bug]: FileResource observer repeatedly calls unsupported DataNode RPC during 2.6-to-3.0 rolling upgrade


zhuwenxing:
## Current Behavior

During a component-ordered rolling upgrade from Milvus 2.6 to 3.0, an upgraded Coordinator temporarily coexists with DataNodes still running 2.6.

The 3.0 `FileResourceObserver` calls `DataNode.SyncFileResource`, but the 2.6 DataNode service does not define this RPC and correctly returns `Unimplemented`.

```text
2026-08-05 13:34:40.614 [WARN]
[coordinator/file_resource_observer.go:247]
["sync file resource failed"]
[nodeID=3306]
[error="service unimplemented: rpc error: code = Unimplemented desc = unknown method SyncFileResource for service milvus.proto.data.DataNode"]
```

The observer treats this expected mixed-version response as a regular synchronization failure:

1. `SyncFileResource` returns `merr.ErrServiceUnimplemented`.
2. The observer assigns the error to `syncErr`.
3. `syncLoop` calls `RetryNotify`.
4. `RetryNotify` schedules another synchronization after three seconds.

This produces a repeated warning and retry loop for every legacy DataNode until that node is upgraded.

The gRPC client also force-resets the connection when it receives `codes.Unimplemented`, so the periodic call can cause unnecessary connection churn in addition to log noise.

## Expected Behavior

The 3.0 Coordinator should treat `merr.ErrServiceUnimplemented` from a 2.6 QueryNode or DataNode as a rolling-upgrade compatibility signal.

It should:

1. Avoid adding this response to the periodic synchronization retry loop.
2. Keep the legacy node marked as not synchronized.
3. Retry synchronization when the node upgrades and registers a new session.
4. Continue retrying genuine synchronization failures.

## Steps To Reproduce

1. Start a Milvus 2.6 cluster.
2. Begin a component-ordered rolling upgrade to Milvus 3.0.
3. Upgrade the Coordinator while leaving at least one DataNode on 2.6.
4. Observe the Coordinator calling `DataNode.SyncFileResource`.
5. Observe the 2.6 DataNode returning:

   ```text
   code = Unimplemented desc = unknown method SyncFileResource for service milvus.proto.data.DataNode
   ```

6. Observe the Coordinator repeating the call approximately every three seconds.

## Minimal Reproduction Script

[...]

PR #52707, opened 2026-08-20
fix: skip file resource sync before first add

aoiasd:
relate: #52216

## Summary
- Skip all FileResource synchronization while the metadata version is zero.
- Skip the initial synchronization when the Coordinator starts with an empty current resource list, including a node-registration notification queued during startup.
- Preserve direct empty-list synchronization after the last resource is removed, including background retries when node-local cleanup fails.
- Redownload active file resources after a node process restarts; stale node directories remain owned by the upper-level cache cleanup lifecycle.

## Testing
Not run locally; verification is delegated to PR CI.

### DIFF
--- a/internal/coordinator/file_resource_observer.go
+++ b/internal/coordinator/file_resource_observer.go
@@ -196,6 +196,13 @@ func (m *FileResourceObserver) Sync() error {
 	var syncErr error
 	activeNodes := make(map[int64]struct{})
 	resources, targetVersion := m.meta.ListFileResource(m.ctx)
+	// Version 0 means no file resource has ever been added successfully. Avoid
+	// sending an empty initial state to nodes before file resources are in use.
+	// Do not check len(resources) here: after the last resource is removed, the
+	// non-zero version and empty list still need to be synced to clear nodes.
+	if targetVersion == 0 {
+		return nil
+	}
 
 	// sync file resource to query node if file resource mode was Sync
 	if m.qnMode == fileresource.SyncMode {
--- a/internal/coordinator/file_resource_observer.go
+++ b/internal/coordinator/file_resource_observer.go
@@ -19,7 +19,6 @@ package coordinator
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -74,15 +73,12 @@ type FileResourceObserver struct {
 	dnMode    fileresource.Mode
 	proxyMode fileresource.Mode
 
-	notifyCh chan struct{}
-	closeCh  chan struct{}
-	wg       sync.WaitGroup
-	sf       conc.Singleflight[any]
-	// retryEmptySync is set when clearing the last file resource fails. Empty
-	// resource notifications are otherwise startup/node-registration no-ops.
-	retryEmptySync atomic.Bool
-	startonce      sync.Once
-	closeOnce      sync.Once
+	notifyCh  chan struct{}
+	closeCh   chan struct{}
+	wg        sync.WaitGroup
+	sf        conc.Singleflight[any]
+	startonce sync.Once
+	closeOnce sync.Once
 }
 
 func NewFileResourceObserver(ctx context.Context) *FileResourceObserver {
@@ -115,7 +111,7 @@ func (m *FileResourceObserver) syncLoop() {
 	for {
 		select {
 		case <-m.notifyCh:
-			err := m.sync(true)
+			err := m.Sync()
 			if err != nil {
 				// retry if error exist
 				m.RetryNotify()
@@ -133,9 +129,20 @@ func (m *FileResourceObserver) syncLoop() {
 func (m *FileResourceObserver) Start() {
 	if m.qnMode == fileresource.SyncMode || m.dnMode == fileresource.SyncMode || m.proxyMode == fileresource.SyncMode {
 		m.startonce.Do(func() {
+			resources, _ := m.meta.ListFileResource(m.ctx)
+			if len(resources) == 0 {
+				// Node registrations can enqueue a notification before the observer
+				// starts. Drop that startup notification when there is nothing to sync.
+				select {
+				case <-m.notifyCh:
+				default:
+				}
+			}
 			m.wg.Add(1)
 			go m.syncLoop()
-			m.Notify()
+			if len(resources) > 0 {
+				m.Notify()
+			}
 		})
 	}
 }
@@ -195,26 +202,13 @@ func (m *FileResourceObserver) CheckAllQnReady() error {
 }
 
 func (m *FileResourceObserver) Sync() error {
-	return m.sync(false)
-}
-
-// sync skips an empty resource list only for background notifications from
-// startup or node registration. Direct synchronization after a removal still
-// sends the empty list to clear node-local files.
-func (m *FileResourceObserver) sync(background bool) error {
 	m.syncMu.Lock()
 	defer m.syncMu.Unlock()
 	var syncErr error
 	activeNodes := make(map[int64]struct{})
 	resources, targetVersion := m.meta.ListFileResource(m.ctx)
-	// Version 0 means no file resource has ever been added successfully. Avoid
-	// syncing regardless of whether this call comes from a direct operation or a
-	// background notification.
+	// Version 0 means no file resource has ever been added successfully.
 	if targetVersion == 0 {
-		m.retryEmptySync.Store(false)
-		return nil
-	}
-	if background && len(resources) == 0 && !m.retryEmptySync.Load() {
 		return nil
 	}
 
@@ -338,7 +332,6 @@ func (m *FileResourceObserver) sync(background bool) error {
 		return true
 	})
 
-	m.retryEmptySync.Store(syncErr != nil && len(resources) == 0)
 	return syncErr
 }
 
--- a/internal/coordinator/file_resource_observer.go
+++ b/internal/coordinator/file_resource_observer.go
@@ -115,10 +115,7 @@ func (m *FileResourceObserver) syncLoop() {
 	for {
 		select {
 		case <-m.notifyCh:
-			if !m.shouldSyncOnNotify() {
-				continue
-			}
-			err := m.Sync()
+			err := m.sync(true)
 			if err != nil {
 				// retry if error exist
 				m.RetryNotify()
@@ -131,20 +128,6 @@ func (m *FileResourceObserver) syncLoop() {
 			return
 		}
 	}
-}
-
-// shouldSyncOnNotify filters background synchronization triggered by startup
-// or node registration. An empty resource list needs no sync for those events,
-// unless an earlier removal failed and still needs to clear node-local files.
-func (m *FileResourceObserver) shouldSyncOnNotify() bool {
-	if m.retryEmptySync.Load() {
-		return true
-	}
-	if m.meta == nil {
-		return false
-	}
-	resources, _ := m.meta.ListFileResource(m.ctx)
-	return len(resources) > 0
 }
 
 func (m *FileResourceObserver) Start() {
@@ -212,6 +195,13 @@ func (m *FileResourceObserver) CheckAllQnReady() error {
 }
 
 func (m *FileResourceObserver) Sync() error {
+	return m.sync(false)
+}
+
+// sync skips an empty resource list only for background notifications from
+// startup or node registration. Direct synchronization after a removal still
+// sends the empty list to clear node-local files.
+func (m *FileResourceObserver) sync(background bool) error {
 	m.syncMu.Lock()
 	defer m.syncMu.Unlock()
 	var syncErr error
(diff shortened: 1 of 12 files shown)


internal/coordinator/file_resource_observer.go
```diff
@@ -196,6 +196,13 @@ func (m *FileResourceObserver) Sync() error {
 	var syncErr error
 	activeNodes := make(map[int64]struct{})
 	resources, targetVersion := m.meta.ListFileResource(m.ctx)
+	// Version 0 means no file resource has ever been added successfully. Avoid
+	// sending an empty initial state to nodes before file resources are in use.
+	// Do not check len(resources) here: after the last resource is removed, the
+	// non-zero version and empty list still need to be synced to clear nodes.
+	if targetVersion == 0 {
```
buqian-zilliz, line 209: **[P3] `Sync` drops the `len(resources) == 0` short-circuit while readiness checks still use it**

The new guard in `Sync` only checks `targetVersion == 0`, and the added comment explains that an empty `resources` list with a non-zero version must still be synced to clear nodes. However, `CheckNodeSynced` and `CheckAllQnReady` still use `version == 0 || len(resources) == 0`. After the last file resource is removed, `targetVersion` stays non-zero and `Sync` sends an empty `SyncFileResourceRequest` to clear query/data nodes, but `CheckAllQnReady` returns `nil` immediately and `CheckNodeSynced` returns `true` without confirming the clearing sync actually reached the nodes. If the clearing sync fails and the subsequent `Notify`/retry has not yet completed, readiness is still reported as synced, and the `distribution` map can keep stale entries until the next successful sync. Consider aligning the readiness checks with the new `Sync` semantics (version-based only) or confirming that the empty-state clear does not need to be gated by these checks.

aoiasd: Thanks for catching the ambiguity. There are two separate conditions here, so I updated the PR to keep them separate instead of changing readiness semantics:

1. `version == 0` is a universal no-sync condition because no FileResource has ever been added successfully.
2. For background notifications from startup or node registration, an empty current resource list is also a no-op, even when the version is non-zero.

A successful Remove still calls `Sync()` directly with the empty list to clear node-local files. If that cleanup fails, the observer records the pending empty sync so the background loop continues retrying it. `CheckNodeSynced` and `CheckAllQnReady` remain active-resource readiness checks and therefore still return ready for an empty active set.

internal/coordinator/file_resource_observer.go
```diff
@@ -126,6 +133,20 @@ func (m *FileResourceObserver) syncLoop() {
...
+// unless an earlier removal failed and still needs to clear node-local files.
+func (m *FileResourceObserver) shouldSyncOnNotify() bool {
+	if m.retryEmptySync.Load() {
+		return true
+	}
+	if m.meta == nil {
+		return false
+	}
+	resources, _ := m.meta.ListFileResource(m.ctx)
+	return len(resources) > 0
```
buqian-zilliz: **[P2] Nodes absent during the last-resource removal never receive the empty clear sync**

`shouldSyncOnNotify` filters out every empty-resource background notification (startup and node registration) unless `retryEmptySync` is set, and `retryEmptySync` is only set at the end of `Sync()` when `syncErr != nil && len(resources) == 0`, i.e. only when an in-progress empty clear fails against a node that is currently present in the manager. A query/data node that is down or partitioned when the last file resource is removed is not in `qnManager.GetAll()` / `dnManager.GetClientIDs()` during the direct `RemoveFileResource` -&gt; `Sync()` call, so no error is recorded and `retryEmptySync` stays `false`. When that node later re-registers (or after a coordinator restart with a non-zero but empty version), the empty notification is filtered out, so the node never receives the empty `SyncFileResourceRequest` and keeps its stale `SyncManager` version and local file resources until the next non-empty add. `CheckNodeSynced` and `CheckAllQnReady` also short-circuit on `len(resources) == 0`, so this staleness is invisible to readiness checks. Please confirm whether a node can re-register without a full restart (retaining in-memory file-resource state); if so, this is a real convergence gap that `retryEmptySync` does not cover.

internal/coordinator/file_resource_observer.go
```diff
@@ -126,6 +133,20 @@ func (m *FileResourceObserver) syncLoop() {
...
+// or node registration. An empty resource list needs no sync for those events,
+// unless an earlier removal failed and still needs to clear node-local files.
+func (m *FileResourceObserver) shouldSyncOnNotify() bool {
+	if m.retryEmptySync.Load() {
+		return true
+	}
+	if m.meta == nil {
+		return false
+	}
+	resources, _ := m.meta.ListFileResource(m.ctx)
```
buqian-zilliz: **[P3] `shouldSyncOnNotify` performs a redundant `ListFileResource` read before `Sync` reads it again**

`shouldSyncOnNotify` calls `m.meta.ListFileResource(m.ctx)` just to compute `len(resources) > 0`, and when the result is `true`, `Sync()` immediately calls `m.meta.ListFileResource(m.ctx)` again for the same data. Each call acquires `ddLock.RLock()` and allocates a fresh slice via `lo.Values(...)`, so every background notification performs this read twice and discards the pre-check result. This sits on the sync-loop path triggered by startup and node registration. Consider returning the already-read `resources`/`version` to `Sync`, or folding the emptiness decision into a single read.

internal/coordinator/file_resource_observer.go
aoiasd: Good catch. A QueryNode/DataNode whose session lease actually expires exits the process, while a transient keepalive failure resumes the same lease rather than re-registering with retained in-memory state. The remaining restart case is persisted local files: the in-memory `resourceMap` was previously lost, so a later sync could neither recognize nor remove those directories.

This update makes `SyncManager` scan its local resource-ID directories on initialization. On the next real sync it reuses complete files that are still active and removes IDs absent from the current resource list; an empty direct sync also clears all restored IDs. Background empty syncs remain skipped by design, so cleanup for a restarted/previously absent node is deferred until the next real sync, while nodes present during Remove are still cleared immediately.

internal/coordinator/file_resource_observer.go
aoiasd: Agreed. I folded the background emptiness decision into the internal sync path, so each notification now takes one `ListFileResource` snapshot and uses that same resources/version pair for the decision and RPC synchronization.

internal/util/fileresource/manager.go
```diff
@@ -298,13 +352,15 @@ func (m *SyncManager) Sync(ctx context.Context, version uint64, resourceList []*
 	for _, id := range newResourceMap {
 		activeIDs[id] = struct{}{}
 	}
-	for _, id := range m.resourceMap {
+	for id := range m.localResourceIDs {
```
buqian-zilliz: **[P3] `resourceMap` is now write-only and never read**

The `resourceMap` field (`map[string]int64`, name -&gt; id) lost both of its readers in this change: the reuse check now goes through `hasLocalResourceFile`/`localResourceIDs`, and the cleanup loop now ranges over `m.localResourceIDs`. `m.resourceMap` is only initialized and assigned (`m.resourceMap = newResourceMap`) but never consulted again, and `analyzer.UpdateGlobalResourceInfo` receives the local `newResourceMap` directly. This is redundant saved state that keeps an unused name mapping alongside `localResourceIDs`; consider removing the field or documenting why it is retained.

internal/util/fileresource/manager.go
```diff
@@ -261,7 +310,11 @@ func (m *SyncManager) Sync(ctx context.Context, version uint64, resourceList []*
 		newResourceMap[resource.GetName()] = resource.GetId()
 		localResourcePath := path.Join(m.localPath, fmt.Sprint(resource.GetId()))
 		localFilePath := path.Join(localResourcePath, path.Base(resource.GetPath()))
-		if id, ok := m.resourceMap[resource.GetName()]; ok && id == resource.GetId() {
+		localFileReady, err := m.hasLocalResourceFile(resource.GetId(), localFilePath)
```
buqian-zilliz: **[P3] `hasLocalResourceFile` adds an `os.Stat` syscall per resource on every sync**

For each resource in `resourceList`, `hasLocalResourceFile` now performs `os.Stat(localFilePath)` after the `localResourceIDs` membership check. The previous code used a pure in-memory `m.resourceMap[resource.GetName()]` lookup to decide reuse. On the steady-state (non-restart) sync path the file was just written and the in-memory set is authoritative, so this introduces one stat syscall per resource per sync. File-resource count and sync frequency are low, so the impact is minor, but if the stat exists only to validate directory entries recovered after a restart, consider tracking known-good entries in memory to avoid the per-sync stat.

aoiasd: Addressed in f9c9640386. IDs restored during manager initialization are tracked separately and stat-ed only once when the first real Sync supplies the expected file name. After successful validation or download, subsequent syncs trust localResourceIDs and perform no os.Stat. Exact validation is deferred until that first Sync because the startup directory scan only knows resource IDs.

internal/util/fileresource/manager.go
aoiasd: Addressed in f9c9640386. Removed the redundant SyncManager.resourceMap field and its test-only assertions. Sync still builds the name-to-ID map from the request snapshot and publishes it directly through analyzer.UpdateGlobalResourceInfo.

internal/coordinator/file_resource_observer.go
```diff
@@ -129,9 +129,20 @@ func (m *FileResourceObserver) syncLoop() {
 func (m *FileResourceObserver) Start() {
 	if m.qnMode == fileresource.SyncMode || m.dnMode == fileresource.SyncMode || m.proxyMode == fileresource.SyncMode {
 		m.startonce.Do(func() {
+			resources, _ := m.meta.ListFileResource(m.ctx)
+			if len(resources) == 0 {
+				// Node registrations can enqueue a notification before the observer
+				// starts. Drop that startup notification when there is nothing to sync.
+				select {
+				case <-m.notifyCh:
```
buqian-zilliz: **[P2] Do not drain mutation retry notifications during startup**

`notifyCh` coalesces node-registration triggers and retry requests from failed `AddFileResource` or `RemoveFileResource` synchronization, so the pending token is not guaranteed to be the startup notification described by the comment. RootCoord is already healthy before the observer reaches `Start`; if this method snapshots an empty list, a concurrent mutation can commit, partially fail its direct `Sync`, and enqueue `Notify` before this receive. The receive then consumes the only retry token, while the stale empty snapshot suppresses a replacement notification, leaving failed nodes on the old resource state until an unrelated event occurs. Start the consumer before discarding any initial trigger, or track startup and retry work separately so mutation retries cannot be dropped.

internal/coordinator/file_resource_observer.go
```diff
@@ -150,6 +152,14 @@ func (m *FileResourceObserver) Notify() {
 	}
 }
 
+func (m *FileResourceObserver) IsEmpty() bool {
+	if m.meta == nil {
+		return true
+	}
+	resources, _ := m.meta.ListFileResource(m.ctx)
```
buqian-zilliz: **[P2] Avoid full metadata scans before notification coalescing**

`IsEmpty` is called synchronously for every QueryNode, DataNode, and Proxy registration, but `MetaTable.ListFileResource` takes the shared `ddLock`, allocates a slice through `lo.Values`, and walks the entire resource map. Because this happens before the capacity-one `notifyCh`, a startup burst of `N` registrations with `R` resources performs `O(N*R)` scans and allocations that the notification coalescer cannot absorb; when resources are present, `Sync` immediately lists the map again. This can lengthen coordinator recovery and contend with metadata writes. Use a constant-time emptiness predicate that does not materialize every resource, or coalesce/type the triggers before evaluating the resource set.

internal/coordinator/file_resource_observer.go
```diff
@@ -150,6 +152,13 @@ func (m *FileResourceObserver) Notify() {
 	}
 }
 
+func (m *FileResourceObserver) IsEmpty() bool {
+	if m.meta == nil {
+		return true
+	}
+	return !m.meta.HasFileResource(m.ctx)
```
buqian-zilliz, line 159: **[P2] Keep node-registration callbacks nonblocking**

`IsEmpty` is executed synchronously by every QueryNode, DataNode, and Proxy registration callback before the capacity-one nonblocking `Notify`. It now calls `MetaTable.HasFileResource`, which waits on `ddLock.RLock`; `AddFileResource` and `RemoveFileResource` hold the corresponding write lock while issuing their persistent catalog transactions. A slow etcd or TiKV write can therefore stall the session-watcher callback and delay subsequent node up/down handling, whereas the previous notification path returned immediately. Please enqueue and coalesce registration triggers before consulting metadata while keeping empty-state mutation retries distinguishable, or provide a genuinely nonblocking snapshot for this predicate.

aoiasd: Thanks for pointing this out. We are going to keep the current locking behavior in this PR. `HasFileResource` is an O(1) read, file-resource mutations are rare, and their catalog operations are expected to be short. Lock contention can only temporarily delay a registration callback; it does not create a correctness or convergence issue. Making this predicate lock-free would introduce duplicated atomic state that must remain consistent with the metadata maps, and we prefer not to add that complexity without evidence of measurable contention. We can revisit this if it shows up in recovery or session-watcher latency.

zhengbuqian (approved):

PR #52712, opened 2026-08-20
enhance: [3.0] optionally hold an external-collection refresh until its segments are indexed

xiaocai2333:
issue: #52714
pr: #52717

Backport of #52717 (merged to master as `155169032398`), squashed into one commit.

The source code is **byte-identical to master's** — verified by diffing the two branches' combined patches over every touched file. The only adaptation is `pkg/proto/datapb/data_coord.pb.go`, regenerated from this branch's `.proto` rather than merged by hand; its diff is the new field, its getter, its comments, and the rawDesc bytes, with no generator-version change.

See #52717 for the full design write-up. What it does, in short: with `dataCoord.externalCollection.refreshWaitForIndex` on (default off), a refresh applies its segments and publishes the refreshed `external_source`/`external_spec` exactly where it always did, then keeps reporting `InProgress` (progress 90-99 tracks the indexed fraction) until every segment is indexed. What waits is the job's completion signal, not the data.

This squash also folds in the six review rounds that landed on master after this branch was last updated, **two of them correctness fixes** this branch would otherwise have shipped without:

- the Failed and Finished paths no longer share a dedup key — a job that entered the wait and hit the job timeout in the same `processJob` pass committed its segments and then never published the refreshed source/spec, with no self-heal;
- the build-channel nudge is gated on the refreshed source/spec being visible in collection meta — an index build resolves the external location at dispatch time, so a nudge issued before the publish lands dispatches builds against the pre-refresh endpoint and credentials, and those fail terminally with nothing to retry them;
- the schema publish retries after a transient failure;
- the debt is settled in the pass that enters the wait, so a refresh whose segments are already indexed does not pay a checker interval;
- a refresh that times out mid-wait no longer reports the same `fail_reason` as one that timed out while still ingesting — the first applied its data and is serving it;
- the apply and the marker are described as two ordered catalog writes under one job lock, which is what they are.

### Testing

Full `./internal/datacoord/` passes on this branch, `-race` included; `pkg/util/paramtable` passes; `golangci-lint` clean; `gofmt`/`gci` clean.


weiliu1031 (commented):
Four findings on the refresh index gate.

internal/datacoord/external_collection_refresh_checker.go
```diff
@@ -239,6 +258,63 @@ func (c *externalCollectionRefreshChecker) aggregateJobState(job *datapb.Externa
...
+			}
+			if held != job.GetProgress() {
+				if err := c.refreshMeta.UpdateJobProgress(job.GetJobId(), held); err != nil {
+					mlog.Warn(c.ctx, "failed to update job progress while waiting for indexes",
+						mlog.FieldJobID(job.GetJobId()), mlog.Err(err))
+				}
+			}
+			return
+		}
+		c.indexGateMu.Lock()
```
weiliu1031: **[P1] Keep the gate state until the terminal transition is durably persisted**

Deleting the gate entry as soon as the debt clears leaves a failure window before `UpdateJobStateWithPreApply` persists `Finished`. If that write fails transiently, `processJob` re-reads the job as `InProgress`, falls back to the already-spent ingest clock, and may mark an applied and fully indexed refresh as `Failed`.

The same lifecycle mismatch affects gate expiry: the segments have already been applied, but the generic timeout path marks the job `Failed` and skips the finished callback/schema broadcast.

Please retain the gate state until a terminal transition is successfully persisted, and make gate expiry finish the applied refresh with a warning instead of failing it.

internal/datacoord/external_collection_refresh_checker.go
```diff
@@ -239,6 +258,63 @@ func (c *externalCollectionRefreshChecker) aggregateJobState(job *datapb.Externa
...
+	// meta.segments: judging them before the apply reads every new segment
+	// as unindexed forever (the inspector drops unknown ids), which is a
+	// held job that can never advance. Only then is the debt computed, from
+	// the tasks' externalized results; stragglers go to the index-build
+	// acceleration channel and the job holds InProgress with progress
+	// tracking the indexed fraction (90-100) until a later pass finds the
+	// debt cleared and lets the Finished transition run - whose own apply
+	// replay is a no-op by then.
+	if state == indexpb.JobState_JobStateFinished && Params.DataCoordCfg.RefreshWaitForIndex.GetAsBool() {
+		if c.applyJobInfo != nil {
```
weiliu1031: **[P2] Snapshot the produced segment IDs when entering the gate**

This path runs on every checker tick while the job is held. `applyJobInfo` reloads every task result from object storage and replays the segment apply, then `refreshSegmentIndexDebt` loads the same results again. With the default 10-second interval and one-hour timeout, this can cause roughly 720 object-store reads per task, plus repeated catalog updates.

Please apply once at gate entry, snapshot the immutable produced segment IDs, and let subsequent ticks query only the in-memory index metadata.

internal/datacoord/external_collection_refresh_checker.go
```diff
@@ -239,6 +258,63 @@ func (c *externalCollectionRefreshChecker) aggregateJobState(job *datapb.Externa
...
+				c.indexGateEntered[job.GetJobId()] = time.Now()
+			}
+			c.indexGateMu.Unlock()
+			for _, segID := range unindexed {
+				select {
+				case getBuildIndexChSingleton() <- segID:
+				default:
+				}
+			}
+			held := int64(90)
```
weiliu1031: **[P2] Keep held progress monotonic and below 100**

Replacing the existing task progress with the indexed fraction can move progress backwards, for example from 96 to 90. A job may also enter the gate with progress 100, and a failed progress write would leave it reporting `InProgress` with progress 100, which completion pollers may interpret as done.

Please keep the held value monotonic with `max(currentProgress, indexedProgress)` and cap it at 99 until the job is actually persisted as `Finished`.

pkg/util/paramtable/component_param.go
```diff
@@ -6833,6 +6834,20 @@ if param targetScalarIndexVersion is not set, the default value is -1, which mea
 	}
 	p.WaitForIndex.Init(base.mgr)
 
+	p.RefreshWaitForIndex = ParamItem{
+		Key:     "dataCoord.externalCollection.refreshWaitForIndex",
+		Version: "3.0.0",
+		Doc: `Hold an external-collection refresh in progress until every segment it produced is indexed.
+Off, a refresh completes when its data lands, and queries meet the new segments before their indexes exist -
+correct, but brute-force-scanned. On, the job stays InProgress (progress 90-100 tracks index building) and the
+refreshed data becomes visible only once it is servable at indexed speed. Deployments that load collections on
```
weiliu1031: **[P2] Document the actual visibility boundary of this gate**

This description says the refreshed data becomes visible only after every segment is indexed, but the implementation applies the segments to DataCoord metadata before entering the wait. An already-loaded replica can therefore pick them up through target updates while the job is still held.

What the gate actually delays is the refresh completion signal and the schema-update broadcast, which protects deployments that load the collection only after refresh completion. Please document that narrower guarantee here and regenerate the corresponding `milvus.yaml` entry.

xiaocai2333:
All four review comments are addressed — fixed on the master PR (#52717) first, then cherry-picked here (head `4aa33826f3`):

- **[P1] gate state lifecycle** → `72e55c8abd` releases the gate clock only after the terminal state persists, and `27eb384359` makes gate expiry finish the applied refresh with a warning instead of failing it (`tryTimeoutJob` now exempts gate-applied jobs outright).
- **[P2] snapshot the produced segment IDs** → `65de5299b1` lands the apply and reads the result store exactly once at gate entry; held ticks only query in-memory index meta. `3d75300c89` additionally removes the pre-apply replay from the Finished transition and dedups the entry pass across the eager/periodic paths.
- **[P2] monotonic held progress** → `14acd1d45d` (never regress, pin at 99 while held) plus `2d3e5bb2c2` (terminal-state guard in `UpdateJobProgress` so a racing held-progress write cannot pin a Finished job below 100).
- **[P2] document the actual visibility boundary** → `f9b4def6ec` rewords the param doc to the narrower guarantee (the gate holds the completion signal and the schema-update broadcast; already-loaded replicas pick up applied segments via target updates) and updates the `configs/milvus.yaml` entry in place. The param version stays 3.0.0 on this branch.

Each fix carries a test; `TestExternalCollectionRefresh*` and `TestApplyExternal*` pass with `-race` on this branch.

xiaocai2333:
/ci-rerun-e2e-amd

Re-triggering `ci-v2/e2e-amd` (dispatcher build 6437 failed 1/105).

The single failing file was `milvus_client/test_milvus_client_search_none_default.py` on worker `w5`. It is not a behavioural failure: every assertion is `create_collection expect True, but got False`, and the underlying error is `Fail connecting to server on pool-6437-distpuls-w5-milvus:19530 (server unavailable)` — all six xdist workers lost the connection within the same second (21:23:02Z), after earlier parametrizations of the same file had already passed on the same instance.

The archived pod logs show the whole `w5` pod set restarted together: etcd-0 at 21:23:35Z, datanode/querynode/streamingnode at 21:23:42Z, proxy/mixcoord at 21:23:43Z. etcd contains no Milvus code, so a defect in this PR cannot explain its restart — this is a node-level event on the pool worker, not the diff.

internal/datacoord/external_collection_refresh_checker.go
xiaocai2333: The implementation was rewritten from scratch (force-push, see the PR comment), so the `indexGates` map this is about no longer exists — there is no in-memory gate state, and nothing to delete. Both halves of your concern are structurally gone rather than patched:

**Transient failure of the terminal write.** `finishAfterIndexWait` calls `UpdateJobState(Finished)` with no pre-apply. If that write fails, the job simply stays `InProgress` with its persisted marker set, and the next checker tick re-evaluates: the debt is still clear, so the Finished write is retried. There is no state to lose in between, because the sub-state is a persisted proto field (`index_wait_started_time`), not a map entry.

**Expiry skipping the finished callback / schema broadcast.** That coupling is gone: `ensureJobFinishedNotified` now fires on the marker, not on `Finished`. The schema broadcast therefore happens when the segments are applied — at the *start* of the wait — so a timeout during the wait cannot skip it; it already happened.

On the second part of your ask — making expiry *finish* the applied refresh with a warning instead of failing it: deliberately not adopted, and this is a decision rather than an oversight. A refresh that outruns its budget is marked `Failed` like any other refresh, and the parameter doc states plainly that a failed refresh does **not** roll back — its segments are already the collection's contents and are being served. `Failed` here means "the wait did not complete in budget", and re-running the refresh is the remedy. Introducing a third outcome ("finished, but degraded") would put the burden back on every consumer, which is exactly the class of complexity the rewrite removed.

internal/datacoord/external_collection_refresh_checker.go
xiaocai2333: Fixed, and more completely than the suggestion. The rewrite applies **exactly once**, at the entry to the wait, in the same catalog write that stamps the marker — and that once-ness is now enforced under the collection job lock, not by a caller-side check (see the PR comment; the caller-side version had a real double-apply race).

Every subsequent tick reads only in-memory metadata: `meta.SelectSegments` for the collection's flushed non-L0 segments, then `indexMeta.GetUnindexedSegments`. **No object-store reads, no result-store reads, no catalog writes** except a progress update when the value actually changes. So the ~720 reads/task you calculated are down to zero.

The snapshot you asked for turned out to be unnecessary: because a refresh *defines* the collection's contents, once its apply lands "this refresh's segments" and "the collection's segments" are the same set. That is what lets the wait stay stateless where `importChecker` — which adds only part of a collection — has to carry `SegmentIDs` on each task.

internal/datacoord/external_collection_refresh_checker.go
xiaocai2333: Two of the three adopted, one deliberately not.

**Cap at 99 — adopted.** Both in the wait (`if held > 99`) and in `normalizeRefreshJobProgress`.

**Entering the wait at 100 — adopted, structurally.** The entry write sets progress to 90 in the *same* catalog write that applies the segments, so there is no tick where a job is waiting and still reporting whatever the ingest left. That also removes the "failed progress write leaves it at 100" case you describe: it is not a later write that can fail independently. There is a test pinning exactly this.

**`max(currentProgress, indexedProgress)` — not adopted.** Task aggregation counts a Finished task as 100, and a single-task job takes the worker's self-reported value, so clamping against the pre-wait progress would pin the whole 90–99 band at 96 or 100 for the entire wait and destroy its information content. The 96 → 90 step is not noise; it is the signal that the job changed phase, and 90–99 is a reserved band that only ever means "waiting for indexes".

Within the band the value can still move backwards in one case: compaction produces new segments during the wait, so the debt grows. I think reporting that honestly is better than hiding it behind a clamp — the job genuinely has more to do.

pkg/util/paramtable/component_param.go
xiaocai2333: Documentation point adopted; the milvus.yaml half does not apply.

The parameter doc now states the narrower guarantee in your own terms: the job "applies its segments and publishes the refreshed external source/spec exactly as before, then keeps reporting InProgress ... until every segment is indexed", and "what waits is the job's completion signal, not the data: the segments are applied and served either way". The design doc says the same, and names the deployments the guarantee is for. Two further consequences you did not raise are documented alongside it: a waiting job keeps refusing new refreshes of its collection, and turning the parameter off releases waiting jobs.

On regenerating `configs/milvus.yaml`: this parameter has no `Export`, so it has no yaml entry — matching every other external-collection parameter. `configs/milvus.yaml` contains **zero** `externalCollection` keys (`externalCollectionCheckInterval`, `externalCollectionJobTimeout`, `externalCollectionJobRetention`, `externalCollectionDropRatioWarn` are all absent), so exporting this one alone would make it the odd member of its family. Happy to export the whole group in a separate change if you would rather they all be visible.

xiaocai2333:
## Force-pushed: replaced with the 3.0 backport of the rewritten #52717

The master PR (#52717) was rewritten from scratch, and this branch is now a clean cherry-pick of it. **The diff here has been replaced entirely** — the previous 14-commit head is preserved at `657c2cf9c6`, and the existing review threads still point at code that no longer exists.

### Why the rewrite

The earlier design pulled the segment apply **out of** the atomic Finished transition and held the schema broadcast, manufacturing a state DataCoord had never had to reason about: *segments committed and being served while the job is neither finished nor failed and the schema unpublished.* Every consumer — timeout, restart recovery, index-build eligibility, schema publication, GC — had to learn about it, and each accommodation opened the next gap. Over fourteen rounds of adversarial review on the master PR, **ten found the defect inside what the previous round had just fixed**.

The rewrite keeps the apply welded to a state write, so `index_wait_started_time` is simultaneously the sub-state marker and the apply-once guard, and the wait is a phase *after* the transition rather than a wedge inside it. ~4000 → 722 lines, 3 → 1 proto fields, 29 → 1 commits.

@weiliu1031 — I replied to each of your four comments in their threads. Short version: three describe machinery that no longer exists (the `indexGates` lifecycle, the segment-id snapshot, the visibility wording) and are addressed structurally; the progress one is adopted in part — the 99 cap and the "never reports 100 while waiting" property are both in, but the `max()` monotonic clamp is deliberately not, because it would freeze the whole 90–99 band. Reasoning is in the thread. Thanks for the review — the per-tick object-store reads you caught were real, and they are now zero.

### Backport notes

Cherry-picked clean. The only conflict was the generated `pkg/proto/datapb/data_coord.pb.go`, regenerated with protoc against 3.0's descriptor rather than hand-merged; tag 13 is free on 3.0 as well, so the wire layout matches master exactly. **No source adaptation was needed** — the parameter keeps `Version: "3.0.0"`, and 3.0's existing external-refresh tests are unmodified.

Verification on 3.0: full `./internal/datacoord/` green (zero failures, including the whole-package run), `-race` on the refresh paths green, `pkg/util/paramtable` green.

xiaocai2333:
## Review follow-ups pushed

A self-review pass over the rewritten implementation turned up one Medium and four Lows — no Critical, no High. All fixed in `f1ca581a2b` (3.0: `4a2ba9746d`). Everything here is about the fact that **the wait can last as long as an index build**, so its per-tick work has to be sized for that rather than for a single pass.

**Medium — task results were held for the whole wait, and 24h past it on a timeout.** `ClearTaskResultsByJobID` ran only on the Finished paths; a Failed job's results wait for `DropJob` at GC (`externalCollectionJobRetention`, default 86400s). So a refresh whose data work fully succeeded but which times out mid-wait now keeps its results for a day — and `ClearTaskResult` clears both the result-store blob *and* the inline `KeptSegments`/`UpdatedSegments` arrays in the etcd task record, one `SegmentInfo` per produced segment. That is a new population of jobs the ungated path never produces.

Fixed by releasing the results at **wait entry** instead — which is where the ungated path effectively already releases them, since there the apply and Finished are the same write. Safe because the marker guarantees the apply is never replayed and `AggregateJobStateFromTasks` reads task *states*, never results (both verified). Net effect: the wait path is now strictly better on this axis than the baseline, not worse.

**Low — progress was rewritten every tick.** The "did it change" test lived in the caller, against a snapshot that can be a tick stale. Moved inside `UpdateJobProgress`, under the job lock, so it is authoritative. The mutation test measures **7 catalog writes over 7 ticks without it, 2 with**.

**Low — the debt was logged every tick, with the full id list.** Now logged only when the progress band moves (≈10 lines across an entire wait) and the id list is sampled, since the debt can be the whole collection.

**Low — `getSegmentsIndexStates` allocated a map per segment**, including for segments with no index records at all. Now allocated only when there is something to put in it. Behaviour-preserving: both callers read the result as `states[indexID]`, which behaves identically on a missing key and an empty map, and neither writes into it — I checked both, and every consumer's tests pass. This is a shared function, so flagging it explicitly: it is touched because this feature is the first caller that scans a whole collection on a timer.

**Low — a waiting job was invisible in the stats log.** It is `InProgress` like any other, so the state histogram could not distinguish "still ingesting" from "waiting for indexes". Now counted separately.

**Low — `indexWaitEnabled()` guarded `c.mt` but `indexWaitDone` dereferenced `c.mt.indexMeta`.** Guarded both.

### Tests

Four added, each mutation-checked:

- **a collection with no index defined does not wait forever** — with a *real* `indexMeta`. The wait's liveness on such a collection rests entirely on `GetUnindexedSegments` returning nil when `len(indexes) == 0`; that contract was asserted only in `TestIndexMeta_GetUnindexedSegments`, with nothing tying it to this feature. Now a change to it fails a test that says why it matters.
- **a job already in the wait before a restart resumes it and never re-applies** — the marker's whole purpose is surviving a restart, so it is now exercised through catalog reconstruction rather than two in-process calls.
- **task results are released when the wait begins, not when it ends.**
- **a held job does not rewrite its progress every tick.**

Full `./internal/datacoord/` green on both branches (including every consumer of `getSegmentsIndexStates`), `-race` green, `pkg/util/paramtable` green.

xiaocai2333:
## Retyped `feat:` → `enhance:`, design doc dropped

This adds no API surface and no new capability. It is an opt-in, default-off refinement of *when an existing operation reports completion* — `RefreshExternalCollection` behaves identically with the parameter off, and with it on the same segments are applied and the same source/spec published at the same point. That is what `enhance:` describes. The design-doc requirement follows the `feat:` type, so the doc went with it (`269f9ac4be`; 3.0 `2237053242`).

Nothing load-bearing was lost — the reasoning already had three durable homes in the repo, each better placed than the doc:

- **Why you would enable it, and what it costs** → the parameter's own `Doc` string in `component_param.go`, which is what an operator actually reads when deciding. It covers which deployments need this, that a failed refresh does not roll back, that a terminally failed index build is waited out, that a waiting job keeps refusing new refreshes of its collection, and that turning the parameter off releases waiting jobs.
- **Why the implementation has this shape** → the code comments, deliberately heavy on exactly the decisions that are expensive to re-derive: why the apply stays welded to a state write, why the apply-once guard must be evaluated under the job lock, why the wait phase keys on the marker rather than the parameter, why the debt excludes L0.
- **What changed and why** → the commit messages.

The narrative for reviewers stays in the PR description above. Happy to restore the doc if a maintainer would rather this stay a `feat:`.

xiaocai2333:
## Rebased onto `3.0` and squashed into one commit — now carries all of #52717

#52717 merged to master as `155169032398`. This branch was four commits behind that work and was missing **two correctness fixes** found in the review rounds that followed:

- **The Failed and Finished paths shared one dedup key.** A job that entered the index wait and hit the job timeout in the same `processJob` pass committed its segments and then never published the refreshed `external_source`/`external_spec` — permanently, since the only key removal is `forgetJob` at GC. The collection would serve the refreshed segments under the pre-refresh source/spec, and subsequent index builds would read the old values.
- **The build-channel nudge ran ahead of the schema publish.** An index build resolves the external location at *dispatch* time (`prepareJobRequest` → `handler.GetCollection` → `CreateJobRequest.ExternalSource/ExternalSpec`), so a refresh that moved its bucket or rotated credentials dispatched builds against the old location; those fail terminally, nothing retries them, and the refresh burns its whole timeout.

Plus the publish retry, the same-pass debt settle, the fail-reason split for the two kinds of timeout, and the corrected description of the apply/marker writes.

**Source code is byte-identical to master's** — I diffed the two branches' combined patches over every touched file and they match exactly. The only adaptation is `pkg/proto/datapb/data_coord.pb.go`, regenerated from this branch's `.proto` instead of merged by hand; its diff is the new field, its getter, its comments and the rawDesc bytes, with no generator-version change.

Verified on this branch: full `./internal/datacoord/` green including `-race`, `pkg/util/paramtable` green, `golangci-lint` 0 issues, `gofmt`/`gci` clean.

Issue #52715, opened 2026-08-20
[Bug]: 2.6 Thread pool does not scale up properly


presburger:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version:
- Deployment mode(standalone or cluster):
- MQ type(rocksmq, pulsar or kafka):    
- SDK version(e.g. pymilvus v2.0.0rc2):
- OS(Ubuntu or CentOS): 
- CPU/Memory: 
- GPU: 
- Others:
```

### Current Behavior

When an idle thread exists, Submit() only wakes it up and skips creating new threads. During burst submissions, many tasks may remain queued while the pool still uses only one worker.
The pool should create new threads when the number of queued tasks exceeds the number of idle threads, without exceeding the configured maximum.

### Steps To Reproduce

```markdown

```

PR #52718, opened 2026-08-20
fix: [2.6] scale thread pool when queued work exceeds idle threads

presburger:
issue: #52715
pr: #50267

Handle idle-worker notification and pool expansion as independent decisions. Expand the pool only when queued demand exceeds currently idle capacity:
if (idle_threads_size_ > 0) {
    condition_lock_.notify_one();
}

if (work_queue_.size() > static_cast<size_t>(idle_threads_size_) &&
    current_threads_size_ < max_threads_size_.load()) {
    // Create a new worker.
}
This avoids unnecessary worker creation when idle workers are sufficient while allowing the pool to scale during task bursts.

### DIFF
--- a/internal/core/src/storage/ThreadPool.h
+++ b/internal/core/src/storage/ThreadPool.h
@@ -135,12 +135,21 @@ class ThreadPool {
 
         if (idle_threads_size_ > 0) {
             condition_lock_.notify_one();
-        } else if (current_threads_size_ < max_threads_size_.load()) {
+        }
+        if (work_queue_.size() > static_cast<size_t>(idle_threads_size_) &&
+            current_threads_size_ < max_threads_size_.load()) {
             // Dynamic increase thread number
-            std::thread t(&ThreadPool::Worker, this);
-            assert(threads_.find(t.get_id()) == threads_.end());
-            threads_[t.get_id()] = std::move(t);
-            current_threads_size_++;
+            try {
+                std::thread t(&ThreadPool::Worker, this);
+                assert(threads_.find(t.get_id()) == threads_.end());
+                threads_[t.get_id()] = std::move(t);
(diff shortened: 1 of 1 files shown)


xiaofan-luan:
A few thoughts:

1. Agree with the approach — decoupling "wake up an idle worker" from "expand the pool" as two independent decisions is cleaner and avoids spawning threads unnecessarily while idle workers are available.
2. However, this targets the 2.6 branch. I'd suggest submitting to master first and getting it merged there, then cherry-picking back to 2.6 (mergify already requires master-first). Please add the corresponding master PR link to the PR body.
3. It would be good to add a simple benchmark or at least some pressure-test numbers in a burst scenario, to quantify the improvement in thread creation count / latency and evaluate the benefit.
4. One question: Milvus C++ core already uses folly. Why not use folly's thread pool here (e.g. `folly::ThreadPoolExecutor` / `folly::CPUThreadPoolExecutor`) instead of maintaining a custom `ThreadPool`? Does the custom pool have special requirements, or is it historical? @zhagnlu you're more familiar with this area — could you take a look?

presburger:
@xiaofan-luan  We found that this issue has already been fixed on the master branch.

We tested the high-priority pool with index build tasks. After the fix, for about 30 GB of raw data, it takes only around 90 seconds from submitting the task to entering the Knowhere build function. With the unfixed version, the time is highly unpredictable — sometimes 4 minutes, sometimes 10 minutes, and sometimes even 15 minutes.

One advantage of this thread pool is that it can scale down automatically. The implementation is not particularly elegant, but it does work reasonably well. The biggest issue is that the scale-up logic sometimes does not run at all. As a result, even if 8 tasks are submitted, they may end up being executed serially.
飞书文档 - 图片



before fix, all task use same thread id.

PR #52720, opened 2026-08-20
fix: [3.0] advance element filters across zero-element batches (#51737)

liliu-z:
Cherry-pick from master
pr: #51737
issue: #51736

## What changed

- Make `GetNextRealBatchSize()` distinguish true exhaustion from a non-empty row batch containing zero logical ARRAY elements.
- Let unary range, binary range, term, and binary arithmetic element visitors return a valid zero-length bitmap and advance exactly one row batch.
- Make unary and binary-range integer-overflow fast paths use element counts and the same cursor contract.
- Add sealed and growing regressions with one full NULL-array batch, one full valid-empty-array batch, and later matching elements. Compound AND/OR predicates verify skipped children stay aligned.

## Why

A complete row batch can contain rows but contribute zero logical elements. Treating its element count of zero as EOF returns `nullptr` before the row cursor advances. Full-mode `ElementFilterBitsNode` still has later segment elements to evaluate and fails its non-null expression-result assertion.

## Cherry-pick adaptations for 3.0

- Dropped the `BloomFilterExpr.cpp` and `RoaringFilterExpr.cpp` hunks. Both files are master-only and do not exist on 3.0. They were the only `GetNextRealBatchSize()` callers outside the files carried over here — verified by grepping every `GetNextRealBatchSize` reference on this branch, all of which now use the `std::optional` contract.
- Dropped the one-line `SegmentChunkReaderTest.cpp` comment hunk. It renames `MoveCursorForDataSingleChunk` to `MoveCursorForData` inside tests that do not exist on 3.0.
- Everything else applied unchanged.

## Verification

- `git diff --check` — passed.
- Audited all 10 `GetNextRealBatchSize()` references on this branch: every callsite checks `has_value()` and then goes through `AdvanceEmptyElementBatch()`.
- Confirmed the APIs the master change depends on already exist on 3.0: `SegmentInterface::prefetch_chunks(op_ctx, field_id, chunk_ids)`, `GetArrayOffsets`, `ArrayOffsets::ElementIDRangeOfRow`, `num_rows_until_chunk`, and `#include <optional>` in `Expr.h`.
- Confirmed the four `IsElementLevelExpression()` overrides landed (`UnaryExpr.h`, `TermExpr.h`, `BinaryRangeExpr.h`, `BinaryArithOpEvalRangeExpr.h`); every other `SegmentExpr` subclass keeps the base `false` and therefore the previous eager-prefetch behavior.
- No local C++ build: this macOS arm64 host has no compatible Conan binaries and no configured `compile_commands.json`, so compile and unit-test execution depend on target-branch CI.


zhengbuqian (commented):
/lgtm
/approve

Issue #52727, opened 2026-08-20
[Bug]: CDC replication stops itself when the replicator replays a topology change that predates it


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (07916765a1) and 2.6; observed in the field on a 2.6.17 build
- Deployment mode: cluster, streaming service enabled, CDC between two clusters
- MQ type: Kafka
- SDK version: pymilvus 2.6.16
- OS: Linux
```

### Current Behavior

After bootstrapping a secondary with `milvus-backup restore secondary` and attaching CDC,
replication starts and then stops 0.2 seconds later. No write on the primary reaches the
secondary again. Nothing reports an error: the topology query keeps returning the edge,
re-applying the same configuration returns success, and the only trace is a pair of INFO
lines.

Observed on the primary's CDC component:

```
01:40:15.904  16 replicators created (16 x etcd PUT under replicating-pchannel/)
01:40:16.089  replicate_stream_client_impl.go:339  "handle AlterReplicateConfigMessage"   broadcastID=…943
01:40:16.089  replicate_stream_client_impl.go:373  "target channel found, skip"
01:40:16.096  replicate_stream_client_impl.go:339  "handle AlterReplicateConfigMessage"   broadcastID=…944
01:40:16.099  replicate_stream_client_impl.go:367  "…done, replicate pchannel removed"
01:40:16.10~  all 16 replicating-pchannel keys DELETEd, consumers closed
```

The configuration that was just applied and created these tasks is broadcast …947 — it is
never read, because the replicator terminates at …943/…944, which are broadcasts from
earlier topology operations.

On the secondary, the same replayed message arrives over the replicate stream and
`overwriteAlterReplicateConfigMessage` finds the secondary absent from its configuration,
so it rewrites the message and the secondary becomes a standalone primary again:

```
01:26:45  Saved replicate configuration  clusterCount=2 … topology-0=primary->secondary
01:40:16  Saved replicate configuration  clusterCount=1 cluster-0=secondary topologyCount=0   ← flipped
01:40:16  recovery_storage_impl.go:638  "inconsistency detected"  x16
          reason="replicate checkpoint is nil when incoming replicate message"
```

### Root cause

A replicator does not start at the live position. `channelReplicator.getReplicateCheckpoint`
asks the target cluster where to resume, and after `restore secondary` the target's
checkpoint is the position the backup was taken at. The replicator therefore replays the
[...]

PR #52704, opened 2026-08-20
fix: update libavrocpp recipe revision

zhikunyao:
## What changed

- Update `libavrocpp/1.12.1.1@milvus/dev` to recipe revision `b4854183542196740ec9a004fdfff7ec`.

## Why

The previous recipe downloads Avro 1.12.1 from `dlcdn.apache.org`, where the pinned source has been removed. This breaks builds with a fresh Conan cache. milvus-io/conanfiles#183 switches the recipe to the permanent Apache archive URL while retaining the verified source checksum.

The new recipe revision has been built, tested, and uploaded to the production `default-conan-local2` remote by the conanfiles publish workflow.

Related: milvus-io/conanfiles#183

## Validation

- `git diff --check`
- Parsed `internal/core/conanfile.py` with Python AST
- conanfiles recipe verification and production publish succeeded: https://github.com/milvus-io/conanfiles/actions/runs/32267737247

PR #52728, opened 2026-08-20
fix: ignore topology changes that predate a replication task

czs007:
issue: #52727

## Problem

A replicator does not start reading at the live position. It resumes from the checkpoint
the target cluster reports, and after `milvus-backup restore secondary` that checkpoint is
the position the backup was taken at. The replicator therefore replays the source WAL from
far behind, walking over every topology change made since — including ones that removed this
very edge before it was re-created.

`IsReplicationRemovedByAlterReplicateConfigMessage` decides purely on the message contents,
with no ordering check, so a historical detach is executed as if it had just been issued:

- the replicator deletes its own `replicating-pchannel/` key and stops
  (`replicate_stream_client_impl.go:337-370`), and the same predicate ends the consume loop
  (`channel_replicator.go:191-199`);
- the same message is forwarded to the secondary, where `overwriteAlterReplicateConfigMessage`
  finds the secondary absent from the configuration and turns it back into a standalone
  primary.

In the field this stopped replication 0.2 s after it started, and left a state nothing
recovers from: the topology query still reports the edge, because only the per-pchannel task
keys are gone and the configuration key is untouched, and re-applying the same configuration
is short-circuited as unchanged and returns success.

Ordinary CDC catch-up never exposes this — the checkpoint stays close to the live position
and only new messages are read.

## Change

`ReplicatePChannelMeta.initialized_checkpoint` already carries the time tick of the
`AlterReplicateConfig` that created the task, so no new field or protocol change is needed.
A message at or before it predates the task and cannot be an instruction for it.

- `util.IsStaleTopologyChange` implements that comparison.
- `IsReplicationRemovedByAlterReplicateConfigMessage` returns false for such a message, so
  the replicator neither deletes its metadata nor ends its consume loop.
- The consume loop drops such a message instead of forwarding it, so the secondary cannot be
  flipped back to a standalone primary by a configuration from the past.
- A task with no initialized time tick — written by an older version — enforces no ordering
  and keeps the previous behaviour.

The pchannel-increasing path is unaffected: it deliberately sets the initialized time tick to
one below the creating message, so that message is still delivered, and it carries the edge,
which takes the existing skip path.

## Tests

`internal/cdc/util` had no test file at all. Added:

- `TestIsStaleTopologyChange` — older than the task, exactly the creating message, newer than
  the task, and the no-initialized-time-tick case.
- `TestIsReplicationRemovedByAlterReplicateConfigMessage` — a current detach still removes the
  replication; a current configuration that keeps the edge does not; **a detach that predates
  the task does not remove it** (the regression); the no-initialized-time-tick case still
  removes; an ignored message never removes.
- `TestChannelReplicatorConsumeLoopSkipsStaleTopologyChange` — a stale topology change is
  neither forwarded (the stream client mock has no `Replicate` expectation, so forwarding
  fails the test) nor ends the consume loop.

The existing `TestChannelReplicatorConsumeLoopDeletesLagSeriesOnRemoval` covers the opposite
direction and still passes: its task carries no initialized checkpoint, so the removal path
is unchanged for it.

### DIFF
--- a/internal/cdc/replication/replicatemanager/channel_replicator.go
+++ b/internal/cdc/replication/replicatemanager/channel_replicator.go
@@ -180,6 +180,17 @@ func (r *channelReplicator) startConsumeLoop() {
 			logger.Info(context.TODO(), "consume loop stopped")
 			return
 		case msg := <-r.msgChan:
+			// A topology change older than this task describes a past state, not an
+			// instruction. Forwarding it would make the secondary act on it — a
+			// configuration that no longer lists the secondary turns it back into a
+			// standalone primary — so it is dropped instead of replicated.
+			if msg.MessageType() == message.MessageTypeAlterReplicateConfig &&
+				util.IsStaleTopologyChange(msg, r.channel.Value) {
+				logger.Info(context.TODO(), "skip replicating topology change older than this replication task",
+					mlog.FieldMessage(msg),
+					mlog.Uint64("initializedTimeTick", r.channel.Value.GetInitializedCheckpoint().GetTimeTick()))
+				continue
+			}
 			err := r.streamClient.Replicate(msg)
 			if err != nil {
 				if !errors.Is(err, replicatestream.ErrReplicateIgnored) {
--- a/internal/cdc/replication/replicatemanager/channel_replicator_test.go
+++ b/internal/cdc/replication/replicatemanager/channel_replicator_test.go
@@ -181,3 +181,83 @@ func TestChannelReplicatorInitMetricFromInitializedCheckpoint(t *testing.T) {
 	got := testutil.ToFloat64(metrics.CDCLastReplicatedTimeTick.WithLabelValues(source, target))
 	assert.InDelta(t, tsoutil.PhysicalTimeSeconds(checkpoint), got, 1)
 }
+
+// TestChannelReplicatorConsumeLoopSkipsStaleTopologyChange covers the case where
+// the replicator resumes from a checkpoint that predates its own creation — the
+// position a `restore secondary` leaves on the target. The replay then walks
+// over a topology change that removed this edge before it was re-created. That
+// message must neither be forwarded to the secondary, where a configuration
+// without the secondary turns it back into a standalone primary, nor end the
+// consume loop.
+func TestChannelReplicatorConsumeLoopSkipsStaleTopologyChange(t *testing.T) {
+	paramtable.Get().Save(paramtable.Get().CommonCfg.ClusterPrefix.Key, "current-cluster")
+	defer paramtable.Get().Reset(paramtable.Get().CommonCfg.ClusterPrefix.Key)
+
+	source, target := "TestStale-source", "TestStale-target"
+	// The task was created by the configuration appended at time tick 100.
+	replicateInfo := &streamingpb.ReplicatePChannelMeta{
+		SourceChannelName: source,
+		TargetChannelName: target,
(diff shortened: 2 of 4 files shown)


tinswzy:

Verified:
- `internal/streamingcoord/server/balancer/channel/manager.go:686-707` — `InitializedCheckpoint.TimeTick` has exactly one writer and it stores the creating `AlterReplicateConfig`'s append result on that same pchannel, so the new comparison is against the right clock.
- `internal/cdc/replication/replicatemanager/channel_replicator.go:151` — the scanner filter is `DeliverFilterTimeTickGT(cp.TimeTick)`, so the creating message is already excluded on the normal path and treating `==` as stale reproduces existing behaviour rather than changing it.
- `internal/streamingcoord/server/balancer/channel/manager.go:687-693` — `skipGetReplicateCheckpoint` tasks persist `init-1`, so their creating message (tick `init+1`) is still forwarded on every pchannel and the broadcast can still complete.
- `internal/cdc/replication/replicatemanager/channel_replicator.go:194` — `streamClient.Replicate(` has a single call site repo-wide and the new guard sits immediately above it, so no other outbound path bypasses the filter.
- `internal/cdc/util/util.go:40-50` — the stale check is placed after `header.Ignore` and before `MustNewConfigHelper`, which also avoids running a historical configuration through a helper that panics on `ErrCurrentClusterNotFound`.
- `internal/cdc/util/util.go:25-26` — `GetInitializedCheckpoint().GetTimeTick()` is nil-safe on both hops and the `initTimeTick != 0` guard preserves prior behaviour for tasks persisted before the field existed.

## Low — non-blocking

- `internal/cdc/replication/replicatemanager/channel_replicator_test.go:113` — At the consume-loop level, no test pins the *forward-and-stop* direction for a task that carries a non-zero initialized checkpoint. The only consume-loop removal test, `TestChannelReplicatorConsumeLoopDeletesLagSeriesOnRemoval`, builds its `ReplicatePChannelMeta` without an `InitializedCheckpoint` (`:113-117`) and sends a message with `WithTimeTick(1)` (`:137`), so `util.IsStaleTopologyChange` short-circuits on `initTimeTick == 0` (`internal/cdc/util/util.go:26`) and the timestamp comparison is never reached; the new test covers only the dropped direction. This gap is demonstrably reachable by mutation: replacing the call at `channel_replicator.go:188` with `r.channel.Value.GetInitializedCheckpoint().GetTimeTick() != 0` — which unconditionally drops every `AlterReplicateConfig` for any task with an init checkpoint, so genuine detaches are no longer forwarded, the loop no longer stops, and `replicating-pchannel/<key>` is never cleaned up — leaves `go test ./internal/cdc/...` fully green. (For completeness: flipping `<=` to `>=` in `util.go:26` *is* caught today, failing both `internal/cdc/util` and `replicatemanager`.) Suggestion: raise the message tick at `:137` from `1` to e.g. `100` and give the task at `:113` an `InitializedCheckpoint{TimeTick: 50}`, so the existing case exercises "non-zero init tick + newer message → must forward and stop the loop". (raised by tinswzy)

liliu-z (commented):
/lgtm
/approve

czs007:
Rebased onto current master (no code change, same 4 files / +277 lines). The branch was cut on Aug 20 and its pinned conan recipes (e.g. libavrocpp, fixed on master by #52704 the next day) no longer resolve, which is why the Build stage failed in the reruns. @bigsheeper @liliu-z could you re-apply /lgtm once CI is green? Thanks.

Issue #52730, opened 2026-08-20
[Bug]: storage v3 FFI writer forwards the literal "default" tlsMinVersion to milvus-storage, all packed writes fail on 2.6


czs007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- Milvus version: branch `2.6` (also reproduced on `poc_2.6.17`); master / 3.0 are NOT affected
- Deployment mode(standalone or cluster): standalone (bare metal build from source)
- MQ type(rocksmq, pulsar or kafka): kafka
- SDK version(e.g. pymilvus v2.0.0rc2): RESTful v2 API
- OS(Ubuntu or CentOS): Ubuntu 22.04, gcc 12
```

### Current Behavior

With `common.storage.useLoonFFI=true` and `minio.ssl.tlsMinVersion` left at its **default value**, every packed (storage v3) write fails. The datanode retries the packed writer endlessly and no data is ever persisted:

```
milvus-storage error: ... property "fs.tls_min_version" value "default" is not in allowed set
... internal/storagev2/packed/pack_writer_v2.go: retry func failed
```

Symptoms at the user level: `flush` never completes, and in a full e2e run every write-dependent test fails.

### Expected Behavior

`minio.ssl.tlsMinVersion: default` means "use the SDK/runtime default" (that is exactly what the config doc in `configs/milvus.yaml` says). It must not be forwarded to milvus-storage as a literal property value, so packed writes should work with the shipped default configuration.

### Steps To Reproduce

```markdown
1. Build milvus from branch `2.6` (or `poc_2.6.17`).
2. Start standalone with `COMMON_STORAGE_USELOONFFI=true`, leaving `minio.ssl.tlsMinVersion` at its default.
3. Create a collection, insert a few rows, call flush.
4. flush does not complete; the log shows `"fs.tls_min_version" ... "default" is not in allowed set` and repeated `pack_writer_v2` retries.

A/B control on the same tree, same MQ, same everything else:
- A: useLoonFFI=true, tlsMinVersion default  -> 9 x "not in allowed set", 9 x pack_writer_v2 failure
- B: useLoonFFI=true, MINIO_SSL_TLSMINVERSION=1.2 -> 0 errors, flush succeeds
```

### Root Cause

`minio.ssl.tlsMinVersion` has `DefaultValue: "default"` (`pkg/util/paramtable/service_param.go:1471`).

On branch `2.6` the two write paths disagree about that sentinel value:

[...]

PR #48476, opened 2026-03-24
enhance: add CRC32C checksum and TLS support for object storage

jiaqizho:
issue: https://github.com/milvus-io/milvus/issues/48360 https://github.com/milvus-io/milvus/issues/48359

Add CRC32C checksum validation for MinIO/S3 PutObject requests and enhance TLS configuration for object storage connections.

- Add minio.ssl.useCRC32C config option to enable CRC32C checksum on PutObject requests for data integrity validation
- Pass TLS minimum version to GoogleHttpClientFactory and add TlsHttpClientFactory for non-GCP storage when TLS is configured
- Add use_crc32c_checksum field to StorageConfig proto and propagate through CGO layers
- Bump milvus-storage dependency

PR #48028, opened 2026-03-05
enhance: propagate TLS minimum version config to C++ storage layer

jiaqizho:
issue: https://github.com/milvus-io/milvus/issues/44999

The previous PR (#48000) added the configurable TLS minimum version on the Go side but didn't wire it through to the C++ core. This commit fills that gap by adding the tls_min_version field to CStorageConfig and passing it down to all the places that build storage configs — index builder, clustering analyzer, packed reader/writer, chunk manager, arrow filesystem, and the loon FFI layer. Also bumps milvus-storage to a version that supports the new property.

PR #52731, opened 2026-08-20
fix: [2.6] skip the "default" tlsMinVersion sentinel in the storage FFI properties

czs007:
issue: #52730
related pr: #48476

`minio.ssl.tlsMinVersion` defaults to the sentinel string `"default"`, documented as "the SDK/runtime default is used". On this branch the two storage v3 write paths disagree about that sentinel:

- CGO path — `tlsMinVersionForStorage()` (`internal/storagev2/packed/util.go`) maps `"default"`/`""` to `""`, so the property is not passed down (#48028, backported as #48119).
- FFI path (`common.storage.useLoonFFI=true`) — `ffi_common.go` only filtered the empty string and forwarded the literal `"default"` to milvus-storage, whose property validator accepts only `""`, `1.0`, `1.1`, `1.2`, `1.3`. The whole property set is rejected, the packed writer fails and retries forever, so no write ever lands.

master and 3.0 already filter the sentinel — this patch is the same code as `ffi_common.go` on master, added there in #48476. The version on this branch came from #48332 instead and never got the filter.

Verified on a bare-metal standalone build of this line (kafka, RESTful v2, insert + flush + query), with `common.storage.useLoonFFI=true` and `minio.ssl.tlsMinVersion` left at its default:

| | before | after |
|---|---|---|
| `"default" ... is not in allowed set` | 9 | 0 |
| `pack_writer_v2` retry failures | 9 | 0 |
| flush | never completes | completes |

Setting `minio.ssl.tlsMinVersion: 1.2` explicitly was the only workaround before this patch.

Unit tests of the affected package pass before and after the change (the existing cases do not cover this path).

### DIFF
--- a/internal/storagev2/packed/ffi_common.go
+++ b/internal/storagev2/packed/ffi_common.go
@@ -144,10 +144,10 @@ func MakePropertiesFromStorageConfig(storageConfig *indexpb.StorageConfig, extra
 	keys = append(keys, PropertyFSRequestTimeoutMS)
 	values = append(values, strconv.FormatInt(storageConfig.GetRequestTimeoutMs(), 10))
 
-	// Add TLS min version
-	if storageConfig.GetSslTlsMinVersion() != "" {
+	// Add TLS min version (skip "default" — consistent with C++ layer filtering)
+	if v := storageConfig.GetSslTlsMinVersion(); v != "" && v != "default" {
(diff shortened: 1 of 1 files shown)


liliu-z (commented):
/lgtm
/approve

Issue #52734, opened 2026-08-20
[Bug]: a replicate configuration read from the cluster cannot be written back — connection_param.token is redacted on read and immutable on write


czs007:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (23882ede2d); reproduced on a 2.6.17 build
- Deployment mode(cluster/standalone): cluster
- MQ type(rocksmq/natsmq/pulsar/kafka): kafka
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus 2.6.16
- OS(Ubuntu or CentOS): Ubuntu
```

### Current Behavior

A replicate configuration that was read from the cluster can never be written
back, so the natural read-modify-write cycle over
`GetReplicateConfiguration` / `UpdateReplicateConfiguration` has no working
form.

Two pieces of behaviour are individually reasonable and jointly leave no path:

- `pkg/util/replicateutil/sanitize.go:36-38` clears `ConnectionParam.Token` on
  every read, so a caller is never given the token.
- `pkg/util/replicateutil/config_validator.go:295-298` rejects any change to
  `ConnectionParam.Token` by strict equality, with no exemption for an empty
  incoming value.

The caller therefore sends back the empty token it was handed and the update is
rejected with `cluster '<id>' connection_param.token cannot be changed`. Unlike
the `uri` check just above it, the error does not report the current value, and
the read path is designed not to expose it, so there is nothing the caller can
do from the API alone.

Every operation goes through the same submission — `UpdateReplicateConfiguration`
has no "topology only" form, each call resubmits full cluster blocks — so
whoever does not independently remember the token registered at setup time
cannot create an edge, cannot change one, and cannot remove one. Removing an
edge is affected too: the accepted single-cluster "independent" configuration
also carries a `connection_param`.

`force_promote` is not affected: `handleForcePromote` returns before the normal
validation and `buildForcePromoteConfiguration` rebuilds the configuration from
the stored cluster block, ignoring the caller's payload. So a secondary can
still be promoted to a standalone primary; what is lost is the ability to form
or tear down a replication edge again.

### Expected Behavior

A configuration read from the cluster can be submitted again, with or without
topology edits, and an actual attempt to change a token is still rejected.

### Steps To Reproduce

```python
from pymilvus import MilvusClient

c = MilvusClient(uri="http://<cluster-a>:19530")
cfg = c.get_replicate_configuration(timeout=15)

[...]

PR #52735, opened 2026-08-20
fix: let a replicate configuration read from the cluster be written back

czs007:
issue: #52734

`SanitizeReplicateConfiguration` clears every `ConnectionParam.Token` when the
replicate configuration is read, and `validateClusterConsistency` rejects any
change to a connection token by strict equality. A caller that reads the
configuration back, edits the topology and submits it sends the empty token it
was handed, and the update is always rejected with
`connection_param.token cannot be changed`. Unlike the `uri` check next to it,
that error does not report the current value, and the read path exists so it is
not exposed.

`UpdateReplicateConfiguration` has no topology-only form — every call resubmits
full cluster blocks — so this blocks creating, changing and removing a
replication edge for anyone who does not independently remember the token
registered at setup time. Removing an edge is included: the single-cluster
"independent" configuration carries a `connection_param` too. `force_promote` is
not affected; it returns before this validation and rebuilds the configuration
from stored state.

Relaxing the validator alone would be worse than today's behaviour. The
incoming configuration is broadcast verbatim in
`AlterReplicateConfigMessageHeader` and becomes the stored configuration, so an
accepted empty token would erase the credential CDC passes as the client
`APIKey`. This takes the stored token into the incoming configuration instead,
before anything compares, validates or broadcasts it. A token the caller did
supply is left untouched, so a genuine attempt to change one is still rejected.

### Verification

Same script against two images differing only by this change, on a two-cluster
CDC pair:

| | before | after |
|---|---|---|
| resend the configuration exactly as read | rejected | **accepted** |
| read → add the `a -> b` edge → write | rejected | **accepted** |
| change a token on purpose | rejected | rejected |

After the round trip the stored configuration still holds the original token,
and replication over the edge it created works — a 300-row insert on the source
reaches the target.

Unit tests cover the fill helper on its own (redacted tokens filled, a supplied
token left alone, unknown cluster, nil arguments, nil connection param, and that
the caller's configuration is not mutated) and the round trip through the
validator, including that the pre-fix rejection still holds without the helper
and that changing a token is still refused.

### DIFF
--- a/internal/streamingcoord/server/service/assignment.go
+++ b/internal/streamingcoord/server/service/assignment.go
@@ -84,6 +84,14 @@ func (s *assignmentServiceImpl) UpdateReplicateConfiguration(ctx context.Context
 		return s.handleForcePromote(ctx, config)
 	}
 
+	// Connection tokens are redacted whenever the configuration is read, so a
+	// caller that read it back cannot resend them. Take the stored ones before
+	// anything compares or validates this configuration.
+	config, err := s.fillRedactedConnectionTokens(ctx, config)
+	if err != nil {
+		return nil, err
+	}
+
 	// check if the configuration is same.
 	// so even if current cluster is not primary, we can still make a idempotent success result.
 	if _, err := s.validateReplicateConfiguration(ctx, config); err != nil {
@@ -122,6 +130,22 @@ func (s *assignmentServiceImpl) UpdateReplicateConfiguration(ctx context.Context
 		return nil, err
 	}
 	return &streamingpb.UpdateReplicateConfigurationResponse{}, nil
+}
+
+// fillRedactedConnectionTokens replaces the redacted connection tokens of an
+// incoming configuration with the ones already stored for the same clusters.
+// Without it a configuration that was read back can never be written again: the
+// read redacts the tokens and the validator rejects the change.
+func (s *assignmentServiceImpl) fillRedactedConnectionTokens(ctx context.Context, config *commonpb.ReplicateConfiguration) (*commonpb.ReplicateConfiguration, error) {
+	balancer, err := balance.GetWithContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	latestAssignment, err := balancer.GetLatestChannelAssignment()
+	if err != nil {
+		return nil, err
+	}
+	return replicateutil.FillRedactedConnectionTokens(config, latestAssignment.ReplicateConfiguration), nil
 }
 
 // waitUntilPrimaryChangeOrConfigurationSame waits until the primary changes or the configuration is same.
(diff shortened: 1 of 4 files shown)


liliu-z (commented):
/lgtm
/approve

internal/streamingcoord/server/service/assignment.go
```diff
@@ -84,6 +84,14 @@ func (s *assignmentServiceImpl) UpdateReplicateConfiguration(ctx context.Context
 		return s.handleForcePromote(ctx, config)
 	}
 
+	// Connection tokens are redacted whenever the configuration is read, so a
+	// caller that read it back cannot resend them. Take the stored ones before
+	// anything compares or validates this configuration.
+	config, err := s.fillRedactedConnectionTokens(ctx, config)
```
xiaofan-luan, line 90: Could we derive the final token-filled configuration from the original request under the `ExclusiveCluster` lock, using the same current snapshot for filling and validation?

This call reads the stored configuration before the lock, while each `validateReplicateConfiguration` reads it again. Two `UpdateReplicateConfiguration` RPCs can interleave as follows:

1. The current topology is A -> B, with B's old URI and token T.
2. A request to add B at a new URI, using the same cluster ID and an empty token, reaches this call and copies T from the old configuration.
3. A concurrent request removing B commits the standalone A configuration before the first validation of the add request.
4. Both subsequent validations now treat B as a new cluster, so they skip the old URI/token consistency checks. The resulting broadcast stores the new URI together with the old token T, which CDC can then send to the new endpoint.

The cluster lock serializes the broadcasts, but it does not cover the initial credential lookup and copy. Revalidating the already-filled `config` after locking cannot catch this, because B no longer exists in the current snapshot.

Please retain the original incoming configuration and rebuild the final configuration after acquiring the lock. The lock-free idempotent/secondary path should also use one snapshot for filling and validation. A service-level regression test for this interleaving would help protect the credential-preservation behavior.

czs007: You are right, and this landed before I got to it. Fixed in #53340.

I reproduced the interleaving you described: with the first assignment read still holding the peer cluster and every later read no longer holding it, the request is accepted and the broadcast carries the stored token bound to the uri the caller chose. `validateClusterConsistency` skips a cluster that is absent from the current configuration, so nothing checks the uri once the concurrent removal commits.

The filling now happens inside `validateReplicateConfiguration`, from the same `GetLatestChannelAssignment` result it validates against, so both call sites — the one before the resource key and the one holding it — rebuild the final configuration from a single snapshot, and the original incoming configuration is what is carried between them. If the peer is still present the uri change is rejected as before; if it is gone, no token is filled. `waitUntilPrimaryChangeOrConfigurationSame` fills against the configuration each watch callback carries, for the same reason.

The service-level regression test for the interleaving is in that PR. #52736 put the same shape on `poc_2.6.17`; I will send the same follow-up there.

PR #52738, opened 2026-08-21
fix: [3.0] synchronize Sonic JIT registration with Go plugin loading (#51931)

congqixia:
Cherry-pick from master
pr: #51931
Enable Sonic's bytedance_tango build path for the Milvus CPU and GPU executables. This makes Sonic runtime module registration use the same plugin.pluginsMu critical section as plugin.Open.

Add -checklinkname=0 because Go 1.26 otherwise rejects Sonic loader's linkname reference to the private plugin.pluginsMu symbol.

issue: #51930

Tested:
- built a focused host and plugin with Tango synchronization enabled
- passed 50 concurrent Sonic JIT and plugin.Open stress runs
- passed hookutil tests with the required dynamic,test flags
- verified CPU and GPU Makefile command expansion

---------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,12 @@ PWD 	  := $(shell pwd)
 GOPATH	:= $(shell $(GO) env GOPATH)
 SHELL 	:= /bin/bash
 OBJPREFIX := "github.com/milvus-io/milvus/cmd/milvus"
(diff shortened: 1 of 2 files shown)


liliu-z (commented):
/lgtm

PR #52740, opened 2026-08-21
test: tolerate reranker score drift in model-rerank ground truth comparison

### LINKED ISSUE #52737: [Bug]: model-rerank test asserts exact rank equality and fails on ~1e-4 TEI score drift (enable_truncate=True)

### Environment

```markdown
- Milvus version: 2.6 @ c71ab0c5f572 (the assertion discussed here is identical on master)
- Deployment mode(standalone or cluster): standalone (bare metal, single process)
- MQ type(rocksmq, pulsar or kafka): pulsar
- SDK version(e.g. pymilvus v2.0.0rc2): pymilvus 2.6.14rc1 (pinned in tests/python_client/requirements.txt:31)
- OS(Ubuntu or CentOS): Ubuntu; reranker = TEI serving BAAI/bge-reranker-v2-m3, dtype float16, max_input_length 8192
```

### Current Behavior

`test_milvus_client_single_vector_search_with_model_rerank[True-tei]` fails: the ordered list Milvus returns differs from the "ground truth" list the test builds by calling the same TEI `/rerank` endpoint itself, and the two lists are compared with exact equality.

The two lists disagreed only where the score gap was tiny. From the failing run (BM25 round, 10 candidates; 6 of the 10 scored bit-identically in the two calls):

| doc pk | score from the Milvus-side call | score from the test-side call |
|---|---|---|
| 371505 | 0.01584536 | 0.01584536 |
| 370456 | 0.01444785 | 0.01444785 |
| 371360 | 0.01052971 | 0.01044863 |
| 371578 | 0.00821969 | 0.00821969 |
| 372012 | 0.00818791 | 0.00825160 |

Documents 371578 and 372012 sit 3e-5 apart. The drift on 372012 lifts it one rank, so the lists are unequal and the assertion fails at index 4:

```
assert gt == actual_rerank_results, "Rerank result is different from ground truth rerank result"
E  AssertionError: Rerank result is different from ground truth rerank result
```

### Expected Behavior

The test should tolerate score differences below the reranker service's own run-to-run reproducibility. Comparing top-k as a set, or requiring order equality only where the ground-truth score gap exceeds an epsilon, would both work. Milvus paired every document with its own score correctly in the failing run.

### Steps To Reproduce

```markdown
1. Deploy standalone Milvus plus a TEI reranker reachable from both Milvus and the test process.
2. Run the case:

   cd tests/python_client
   pytest "milvus_client/test_milvus_client_search.py::TestMilvusClientSearchModelRerank::test_milvus_client_single_vector_search_with_model_rerank[True-tei]" \
          --host <host> --port <proxy-port> --tei_endpoint http://<tei-host>:<port>

The failure needs two neighbouring candidates whose scores differ by less than the service's reproducibility error. In this parameter set the query is 1024 random words, which makes every candidate roughly equally irrelevant and collapses all ten scores below 0.016 with gaps as small as 3e-5, so the condition is easy to hit. The sibling parameter set [False-tei] passed in the same run, as did test_milvus_client_hybrid_vector_search_with_model_rerank[tei].

Frequency: not measured for the pytest case itself (it ran once in this session). The underlying service behaviour is measured and reproducible — see below.
```

### Milvus Log

```
assert gt == actual_rerank_results, "Rerank result is different from ground truth rerank result"
AssertionError: Rerank result is different from ground truth rerank result
```

### Anything else?

**The premise the assertion relies on does not hold.** Two independent `/rerank` calls do not return comparable-to-the-last-bit scores. Measured directly against the same TEI deployment used by the failing run (BAAI/bge-reranker-v2-m3, float16), with no Milvus involved — each row is 30 repeats of the same request, counting how many distinct score values each document took:

| what was sent | documents that drifted | max relative spread |
|---|---|---|
| 1 document, long query | 0 / 1 | 0 (bit-identical 30/30) |
| 1 document, short query | 0 / 1 | 0 (bit-identical 30/30) |
| 2 distinct documents | 1 / 2 | 1.8e-3 |
| 4 distinct documents | 3 / 4 | 1.8e-3 |
| 10 distinct documents, long query | 9 / 10 | 2.2e-3 |
| 10 distinct documents, short query | 9 / 10 | 5.3e-3 |
| 10 documents, same word count, different text | 10 / 10 | 2.1e-3 |
| 10 copies of one identical document | 0 / 10 | 0 (bit-identical 30/30) |

What this rules in and out:

- **Not precision alone.** The model runs in float16 in every row above, yet a single-document request is bit-identical across 30 repeats. float16 sets how large the drift is, it does not create it.
- **Not sequence length or truncation.** A short query drifts as much as a long one; the `truncate` flag is set identically on both sides of the comparison (`truncate: true, truncation_direction: "Right"` — the test sets it for Milvus in the `Function` params and passes the same values in its own request, and this model's `max_input_length` is 8192, so a 1024-word query is not even truncated).
- **Not padding length.** Ten documents with equal word count but different text still drift.
- **It appears exactly when one batch contains several *distinct* sequences**, and disappears when the batch holds one document, or ten copies of the same document. That is the signature of nondeterministic reduction order in batched GPU kernels: kernel selection heuristics depend on shape and device state, and split-K / atomic accumulation does not fix a summation order. Flash-Attention's forward pass is documented as non-deterministic for this reason (a `deterministic` flag exists since v2.4).

**Control runs that separate "GPU" from "low precision".** The same measurement was repeated against TEI's CPU backend (BAAI/bge-reranker-base), once in float32 and once in float16 — TEI falls back from the ORT backend to Candle for float16 on CPU, which is software-emulated and slow but genuinely float16 (`/info` reports `"model_dtype":"float16"`):

| backend | dtype | same request repeated | same document, different batch mates |
|---|---|---|---|
| GPU | float16 | drifts, 1e-3 … 5e-3 relative | drifts |
| CPU | float32 | bit-identical | bit-identical |
| CPU | **float16** | **bit-identical** | **bit-identical** |

The third row is the informative one: float16 on its own does not produce the drift. Off the GPU the reduction order is fixed, so the same input gives the same bits at either precision. Low precision only sets the size of the perturbation once something else varies the order — a plain float64/32/16 experiment on a 1024-element dot product, summing the same values in 13 different orders, gives a spread of 8.3e-3 relative in float16 versus 1.8e-6 in float32, a factor of ~4500.

So the chain is: batched GPU inference with several distinct sequences → reduction order is not fixed between runs → float16 amplifies that into ~1e-3 relative score movement → neighbouring candidates 3e-5 apart swap places → exact list equality fails.

So the drift the test trips over is a property of the reranker service, ~1e-3 relative here, while the gap it needs to preserve in the failing case is 3e-5 — two orders of magnitude smaller. No amount of care on the Milvus side makes exact list equality hold.

**A second, avoidable variable on the 2.6 line** (not the cause, but it makes the input differ needlessly): the batch Milvus sends is built by iterating a Go map — `internal/util/function/rerank/model_function.go`, `for id, text := range uniqueData` in `processOneSearchData` — so the document order in Milvus's request is a fresh permutation on every call, while the test sends score order. On master this path already builds the batch from an ordered slice (`internal/util/function/chain/expr/rerank_model_expr.go`), so master's order is deterministic; the assertion, however, is unchanged there and remains fragile.

**Suggested fix**: make the comparison tolerance-based (top-k set equality, or epsilon-guarded ordering). This is a test-assumption defect, not a product-behaviour bug.

<details>
<summary>Script used for the measurements above</summary>

```python
import json, random, urllib.request
EP = "http://127.0.0.1:18082"          # TEI reranker
random.seed(20260820)
W = "alpha beta gamma delta epsilon zeta eta theta iota kappa lambda mu nu xi omicron pi rho sigma tau upsilon".split()
sent = lambda n: " ".join(random.choice(W) for _ in range(n))

def rerank(query, texts):
    body = json.dumps({"query": query, "texts": texts,
                       "truncate": True, "truncation_direction": "Right"}).encode()
    req = urllib.request.Request(EP + "/rerank", data=body,
                                 headers={"content-type": "application/json"})
    with urllib.request.urlopen(req, timeout=120) as r:
        return {it["index"]: it["score"] for it in json.load(r)}

def run(tag, query, texts, n=30):
    seen = {i: set() for i in range(len(texts))}
    for _ in range(n):
        for i, v in rerank(query, texts).items():
            seen[i].add(v)
    drifted = sum(1 for v in seen.values() if len(v) > 1)
    spread = max((max(v) - min(v)) / max(abs(max(v)), 1e-12) for v in seen.values())
    print(f"{tag}: {drifted}/{len(texts)} drifted, max relative spread {spread:.2e}")

Q, S = sent(1024), sent(8)
docs = [sent(random.randint(20, 120)) for _ in range(10)]
run("1 doc,  long query ", Q, docs[:1])
run("10 docs, long query", Q, docs)
run("10 docs, short query", S, docs)
run("10 identical docs  ", Q, [docs[0]] * 10)
```

</details>

pujitha24:
Motivation:
TestMilvusClientSearchModelRerank compares Milvus's rerank order against
a "ground truth" order the test computes by independently calling the
same external reranker service (TEI/vLLM), using strict list equality.
The reranker service performs batched GPU inference, which is not
bit-reproducible across calls: repeated requests for the same input can
return scores that drift by up to ~5e-3 relative. When two neighbouring
candidates' true scores are only a few 1e-5 apart, the Milvus-side call
and the test's own ground-truth call can rank them differently even
though both correctly reflect the (slightly different) scores each call
received. The strict equality assertion then fails, even though there is
no product bug -- both orderings are individually correct for the scores
they were given, and score comparisons this close are dominated by the
reranker service's own reproducibility, not by Milvus.

Approach:
Add assert_rerank_order_matches_ground_truth, which checks that Milvus's
result set matches the ground truth set, and that ordering is preserved
only for document pairs whose ground truth scores differ by more than a
tolerance (atol=1e-6, rtol=1e-2 of the larger score). Pairs within
tolerance may appear in either order; pairs with a clear score gap must
still keep their relative order, so a genuine ordering defect is still
caught. This replaces the previous exact `gt == actual_rerank_results`
assertion used by test_milvus_client_single_vector_search_with_model_rerank
and test_milvus_client_hybrid_vector_search_with_model_rerank.

Validation:
This test suite requires a live Milvus deployment plus a reachable
TEI/vLLM reranker endpoint, neither of which is available in this
environment, so the parametrized pytest case itself was not executed
here. Instead:
- `python3 -m py_compile tests/python_client/milvus_client/test_milvus_client_search.py`
  passes.
- `ruff check --config tests/ruff.toml tests/python_client/milvus_client/test_milvus_client_search.py`
  reports no issues.
- The new comparison logic was validated offline against the exact
  numbers reported in the issue (a standalone script reproducing
  assert_rerank_order_matches_ground_truth's logic): fed the reported
  5-document score table, it confirmed the old strict assertion fails on
  that data (reproducing the reported flake) while the new tolerant
  assertion passes on the same data; it also confirmed the new assertion
  still fails when a document pair with a clear (non-drift-sized) score
  gap is reordered, and still fails on a result-set mismatch, so it is
  not a rubber-stamp comparison.
- This does not prove the fix resolves the flake against a live TEI
  service; it demonstrates the new tolerance logic behaves correctly on
  the reported failure data and still catches real ordering violations.

Report: https://github.com/milvus-io/milvus/issues/52737

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #52737

### DIFF
--- a/tests/python_client/milvus_client/test_milvus_client_search.py
+++ b/tests/python_client/milvus_client/test_milvus_client_search.py
@@ -6367,6 +6367,30 @@ class TestMilvusClientSearchModelRerank(TestMilvusClientV2Base):
 
             log.info(f"{'-' * 58} | {'-' * 58}")
 
+    def assert_rerank_order_matches_ground_truth(self, actual_rerank_results, gt, gt_scores, rtol=1e-2, atol=1e-6):
+        """
+        Compare Milvus's rerank order against the ground truth order, tolerating
+        reorderings between documents whose ground truth scores are within the
+        reranker service's own run-to-run reproducibility. Repeated calls to the
+        same reranker service can return scores that drift by up to ~5e-3
+        relative between calls, which is enough to flip the order of
+        neighbouring candidates without indicating a real Milvus defect.
+
+        Documents whose ground truth scores differ by more than the tolerance
+        must still keep their relative order; only near-tied documents may swap.
+        """
+        assert set(actual_rerank_results) == set(gt), \
+            f"Rerank result set differs from ground truth: {actual_rerank_results} vs {gt}"
+        actual_rank = {doc: idx for idx, doc in enumerate(actual_rerank_results)}
+        for i in range(len(gt)):
+            for j in range(i + 1, len(gt)):
+                tolerance = atol + rtol * max(abs(gt_scores[i]), abs(gt_scores[j]))
+                if gt_scores[i] - gt_scores[j] > tolerance:
+                    assert actual_rank[gt[i]] < actual_rank[gt[j]], \
+                        (f"Rerank order violated beyond score tolerance: expected "
+                         f"'{gt[i]}' (score {gt_scores[i]}) before '{gt[j]}' "
+                         f"(score {gt_scores[j]}), tolerance {tolerance}")
+
     def compare_milvus_rerank_with_origin_rerank(self, query_texts, rerank_results, results_without_rerank,
                                                  enable_truncate=False,
                                                  provider_type=None,
@@ -6443,8 +6467,10 @@ class TestMilvusClientSearchModelRerank(TestMilvusClientV2Base):
             self.display_side_by_side_comparison(query_text, actual_rerank_results, gt, doc_to_original,
                                                milvus_scores=distances, gt_scores=gt_scores)
 
-            # Use strict comparison since scores are now normalized to f32 precision
-            assert gt == actual_rerank_results, "Rerank result is different from ground truth rerank result"
+            # Tolerate reordering only between documents whose ground truth scores
+            # are within the reranker service's own run-to-run reproducibility;
+            # order between documents with a clear score gap must still match.
+            self.assert_rerank_order_matches_ground_truth(actual_rerank_results, gt, gt_scores)
 
     @pytest.mark.parametrize("ranker_model", [
         pytest.param("tei", marks=pytest.mark.tags(CaseLabel.L1)),


tests/python_client/milvus_client/test_milvus_client_search.py
```diff
@@ -6443,8 +6467,10 @@ def compare_milvus_rerank_with_origin_rerank(self, query_texts, rerank_results,
             self.display_side_by_side_comparison(query_text, actual_rerank_results, gt, doc_to_original,
                                                milvus_scores=distances, gt_scores=gt_scores)
 
-            # Use strict comparison since scores are now normalized to f32 precision
-            assert gt == actual_rerank_results, "Rerank result is different from ground truth rerank result"
+            # Tolerate reordering only between documents whose ground truth scores
+            # are within the reranker service's own run-to-run reproducibility;
+            # order between documents with a clear score gap must still match.
+            self.assert_rerank_order_matches_ground_truth(actual_rerank_results, gt, gt_scores)
```
yhmo, line 6473: tests/python_client/milvus_client/test_milvus_client_search.py line:6473
Medium ---- This call site is the shared compare_milvus_rerank_with_origin_rerank helper, so the tolerance now also governs the cohere, voyageai and siliconflow single-vector and hybrid rerank tests (8 test functions in total across tei/vllm/cohere/voyageai/siliconflow), not only the two model-rerank tests named in the PR description. Those cloud-provider tests previously enforced exact order; the rtol=1e-2 tolerance was calibrated to TEI GPU batch nondeterminism (~5e-3 relative drift) and now applies unconditionally to providers whose scores come from deterministic cloud APIs, silently relaxing the ordering assertion there. Consider scoping the tolerant comparison to the tei/vllm model-rerank providers or explicitly documenting the wider applicability.

tests/python_client/milvus_client/test_milvus_client_search.py
```diff
@@ -6367,6 +6367,30 @@ def display_side_by_side_comparison(self, query_text, milvus_results, gt_results
 
             log.info(f"{'-' * 58} | {'-' * 58}")
 
+    def assert_rerank_order_matches_ground_truth(self, actual_rerank_results, gt, gt_scores, rtol=1e-2, atol=1e-6):
```
yhmo, line 6370: tests/python_client/milvus_client/test_milvus_client_search.py line:6370
Low ---- The value of this helper is that it is not a rubber stamp (set mismatch and clear-gap reorder must still fail), but that property is only demonstrated by an offline script that is not retained in the repo, and the parametrized e2e case was not executed in this environment. The helper is pure logic over lists/floats, so a small standalone unit test (near-tied swap passes, clear-gap reorder fails, set mismatch fails) could pin down these three behaviors in the repo and protect future edits from silently turning the assertion into a no-op.

tests/python_client/milvus_client/test_milvus_client_search.py
```diff
@@ -6367,6 +6367,30 @@ def display_side_by_side_comparison(self, query_text, milvus_results, gt_results
...
+
+        Documents whose ground truth scores differ by more than the tolerance
+        must still keep their relative order; only near-tied documents may swap.
+        """
+        assert set(actual_rerank_results) == set(gt), \
+            f"Rerank result set differs from ground truth: {actual_rerank_results} vs {gt}"
+        actual_rank = {doc: idx for idx, doc in enumerate(actual_rerank_results)}
+        for i in range(len(gt)):
+            for j in range(i + 1, len(gt)):
+                tolerance = atol + rtol * max(abs(gt_scores[i]), abs(gt_scores[j]))
```
yhmo, line 6387: tests/python_client/milvus_client/test_milvus_client_search.py line:6387
Low ---- rtol=1e-2 is only ~2x the measured single-document max relative drift (5.3e-3 from issue #52737). A pair flip needs the two documents' actual-side scores to swing in opposite directions, so the worst-case combined swing is ~1.06e-2 relative and can slightly exceed the tolerance, leaving a narrow band of ground-truth gaps that could still re-introduce the flake. The reported failing gap (3.2e-5) is comfortably covered, but consider widening rtol to ~2e-2 or documenting why the 1e-2 margin is sufficient.

xiaofan-luan:
Closing as superseded by #52742, which has merged the rerank score-drift tolerance fix and covers both rerank test paths. Thank you for the contribution.

Issue #52431, opened 2026-08-11
[Bug]: Existing BM25 Function Field can transiently return empty results while adding a second Function Field


lyyyuna:
## Environment

- **Milvus Version**: observed on recent master images, including `master-20260808-025d2e3e-amd64` and `master-20260810-5c31ea9f-amd64`
- **Deployment Mode**: cluster
- **Storage**: Storage V3 sealed segments
- **Topology used by the case**: 8 shards, 4 partitions, 4 query replicas
- **Test case**: `E51 test_schema_evolution_multi_function_hybrid_search_multi_replica`

## Reproduction

1. Create a collection with fields `id`, `title`, `body`, and a dense vector field.
2. Insert and flush 16,384 rows into each of four partitions, for 65,536 historical rows in total.
3. Load the collection with four replicas.
4. Add BM25 Function Field A: `title -> sparse_title`.
5. Wait for A's schema/backfill readiness using actual searches. Confirm historical and growing rows are searchable, wait another 10 seconds, and require five additional successful search rounds.
6. Add BM25 Function Field B: `body -> sparse_body`, then drop B.
7. Re-add B. While B is being backfilled, continuously search the already-ready A with the same request:

```python
result = client.search(
    collection_name=collection_name,
    data=[retained_title_token],
    anns_field="sparse_title",
    filter="id == 0",
    limit=1,
    output_fields=["id", "title", "body"],
    search_params={"metric_type": "BM25"},
    timeout=3,
)
assert [hit["id"] for hit in result[0]] == [0]
```

### Trigger Conditions

- **Frequency**: intermittent; 3 of the 4 scheduled master regressions starting on Aug 10, 2026 reproduced it.
- **Recent failing workflows**: `schema-evolution-matrix-l3-rzf6h`, `schema-evolution-matrix-l3-nptk4`, and `schema-evolution-matrix-l3-f8fdz`.
- **Recent passing workflow**: `schema-evolution-matrix-l3-4h964`.
- The same Milvus commit has produced both passing and failing executions in the broader run history, so this is timing-dependent rather than a deterministic version regression.

## Expected Behavior

Adding or backfilling Function Field B must not make the already-backfilled and stable Function Field A temporarily unavailable. The filtered A search should continue returning `id=0`.

## Actual Behavior

During B's re-add backfill window, one search RPC against A succeeds at the RPC level but returns an empty result:

```text
AssertionError: retained title BM25 expected id=0, got=[]
```

[...]

PR #52741, opened 2026-08-21
fix: validate compaction fallback parents

aoiasd:
issue: #52431

## Summary
- Normalize an interrupted compaction view before index fallback. If every direct compactFrom remains in the view, retain compactFrom because compactTo may be incomplete; once any compactFrom has retired, retain compactTo because outputs are published first.
- Replace a child only when its complete direct parent set is ready for the current schema.
- Resolve M:N overlap by expanding through ready parents; if that is impossible, cancel the conflicting fallback and keep the child.

## Verification
- Added focused DataCoord regression cases for both interrupted-compaction phases; PR CI is pending for the current commit.
- An earlier revision completed a local-only 10-minute randomized stress test with 180,663 iterations, 100 initial segments, and 100 compactions per iteration. The temporary stress test is not included in this PR.

### DIFF
--- a/internal/datacoord/handler.go
+++ b/internal/datacoord/handler.go
@@ -226,7 +226,8 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
 		return indexed.Contain(segID) || ((validSegmentInfos[segID].GetIsSorted() || validSegmentInfos[segID].GetIsSortedByNamespace()) && validSegmentInfos[segID].GetNumOfRows() < Params.DataCoordCfg.MinSegmentNumRowsToEnableIndex.GetAsInt64())
 	}
 
-	flushedIDs, droppedIDs = retrieveSegment(validSegmentInfos, flushedIDs, droppedIDs, segmentIndexed)
+	fallbackParentReady := func(segID UniqueID) bool { return indexed.Contain(segID) }
+	flushedIDs, droppedIDs = retrieveSegment(validSegmentInfos, flushedIDs, droppedIDs, segmentIndexed, fallbackParentReady)
 
 	seekPosition := h.GetChannelSeekPosition(channel, partitionIDs...)
 	// if no l0 segment exist, use checkpoint as delete checkpoint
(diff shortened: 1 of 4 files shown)


internal/datacoord/handler.go
```diff
@@ -253,78 +253,32 @@ func retrieveSegment(validSegmentInfos map[int64]*SegmentInfo,
...
-				for _, fromID := range compactionFrom {
-					newFlushedIDs.Insert(fromID)
-					continueRetrieve = true
-					droppedIDs.Remove(fromID)
-				}
-			}
+	for id := range flushedIDs {
+		segment := validSegmentInfos[id]
+		compactionFrom := segment.GetCompactionFrom()
+		if len(compactionFrom) == 0 || segmentIndexed(id) || !allParentsReady(compactionFrom...) {
```
buqian-zilliz, line 308: **[P1] Shared retrieveSegment refactor silently changes GetCurrentSegmentsView semantics**

`retrieveSegment` is also called by `GetCurrentSegmentsView` with a `segmentIndexed` closure that always returns `true` (the `func(segID UniqueID) bool { return true }` passed at the second call site). Previously a flushed segment whose compaction sources were still present in the flushed set was replaced by those sources: `compactionFromExistWithCache` walked the compaction chain and the `else` branch inserted the parents, with the child omitted. The new condition `len(compactionFrom) == 0 || segmentIndexed(id) || !allParentsReady(compactionFrom...)` short-circuits on `segmentIndexed(id) == true`, so the child is always kept and the fallback branch is never reached for this caller. As a result `GetCurrentSegmentsView`/`GetSegmentsByStates` now report both a compaction target and its still-flushed source segments during the in-progress compaction window, a behavior change not described in the PR summary and not reflected in the updated tests. Please confirm whether this is intended, or preserve the source-fallback behavior for the constant-`true` `segmentIndexed` caller.

internal/datacoord/handler.go
```diff
@@ -253,78 +253,48 @@ func retrieveSegment(validSegmentInfos map[int64]*SegmentInfo,
...
+	for id := range newFlushedIDs {
+		segment, ok := validSegmentInfos[id]
+		if !ok || segment == nil {
+			continue
+		}
+		compactionFrom := segment.GetCompactionFrom()
+		if len(compactionFrom) > 0 && lo.EveryBy(compactionFrom, func(id UniqueID) bool {
+			return newFlushedIDs.Contain(id)
+		}) {
+			newFlushedIDs.Remove(id)
```
buqian-zilliz: **[P1] Dedup pass mutates newFlushedIDs while iterating, making multi-generation fallback order-dependent and non-deterministic**

The second loop ranges over `newFlushedIDs` and, for each entry, calls `newFlushedIDs.Remove(id)` while `lo.EveryBy` checks whether every direct parent is still contained in the same set. Go map iteration order is randomized and this set is mutated during iteration, so the result for a chain deeper than one compaction generation depends on visitation order. For a chain `a,b -> c -> d` where all of `a,b,c,d` land in `newFlushedIDs`: if `d` is visited before `c`, `d` is removed (parent `c` present) and then `c` is removed (parents `a,b` present), yielding `{a,b}`; if `c` is visited first, `c` is removed and `d` then keeps its now-absent parent `c`, leaving an orphaned `d` and yielding `{a,b,d}`. The same meta state therefore produces different `FlushedSegmentIds` between calls, which can make the data view (and QueryCoord's loading decision) unstable. Build the removal set in a first pass (or process segments by compaction generation) so the collapse is deterministic.

internal/datacoord/handler.go
```diff
@@ -253,79 +253,51 @@ func retrieveSegment(validSegmentInfos map[int64]*SegmentInfo,
...
+	// segment remains serviceable. Prefer the parents as one compaction
+	// generation and remove the overlapping child without walking ancestors.
+	toRemove := make(typeutil.UniqueSet)
+	for id := range newFlushedIDs {
+		segment, ok := validSegmentInfos[id]
+		if !ok || segment == nil {
+			continue
+		}
+		compactionFrom := segment.GetCompactionFrom()
+		if len(compactionFrom) > 0 && lo.EveryBy(compactionFrom, func(id UniqueID) bool {
```
buqian-zilliz: **[P1] Deduplication can still fall back more than one generation**

The snapshot-based deduplication removes any selected segment whenever all of its parents are also selected, even when that segment was inserted as the one-generation fallback for a newer child. Consider current leaves `e` and `d`, where `e` compacted from `c`, while sibling outputs `c` and `d` both compacted from `a,b`. The first pass can select `c` for `e` and `a,b` for `d`; this pass then removes `c`, so `e` is represented by grandparents `a,b`. This is reachable when only part of an output family is compacted again. It violates the one-generation guarantee and can reselect segments built for an older schema, recreating the target rollback. Resolve the overlap by retaining the descendant side or canceling the conflicting ancestor fallback instead of removing an inserted direct parent because its own parents were selected elsewhere.

buqian-zilliz: **[P1] Deduplication bypasses the direct-parent readiness check**

The deduplication pass removes an indexed child solely because all parent IDs are present in `newFlushedIDs`, without requiring those parents to pass `allParentsReady`. During a parent/child overlap window, an old parent can remain in `flushedIDs` even though it does not satisfy the current index requirements; the first pass correctly keeps the indexed child, but this pass then removes it and exposes the unready parent. This contradicts the direct-parent readiness rule and can recreate the target rollback during schema backfill. Only prefer the parents when they satisfy the same readiness predicate, or retain the child and cancel the conflicting parent selection.

buqian-zilliz: This issue still exists: the snapshot-based removal can still discard a direct-parent fallback when another branch selected that parent's parents, so the resulting view can walk more than one compaction generation back.

buqian-zilliz: This issue still exists. At the latest head, `hasCompleteAncestorCoverage` treats `candidates.Contain(parentID)` as sufficient coverage and never reapplies `allParentsReady`. An indexed child can therefore still be removed in favor of a direct parent that entered `candidates` from `flushedIDs` even though that parent is currently unindexed or otherwise unready.

buqian-zilliz: This issue still exists. The recursive `hasCompleteAncestorCoverage` walk can remove a direct-parent fallback when that parent's ancestors were selected by another branch, so the final frontier can still fall back more than one compaction generation.

aoiasd: The readiness check intentionally applies only to the replacement-selection phase, which is the only phase allowed to introduce ancestors into `candidates`. After that phase, `candidates` is immutable. A segment already present because it came from `flushedIDs` is a current-frontier leaf, not a fallback introduced by deduplication; `segmentIndexed(id) == false` means QueryCoord must serve that leaf through the growing path, not that the segment is unserviceable. Any dropped parent introduced as a sealed fallback has already passed `allParentsReady` as part of its complete direct-parent group.

`hasCompleteAncestorCoverage` only decides whether an existing candidate is fully duplicated by other existing candidates. It never inserts the parent it encounters. Reapplying `allParentsReady` here would conflate fallback eligibility with the load mode of an original frontier leaf and could leave overlapping data in the target.

aoiasd: The one-generation constraint applies to replacement selection, not to the ancestry walk used to prove duplicate coverage. The recursive walk cannot select or introduce a grandparent; it only examines the immutable candidate set produced by the direct-parent replacement phase.

For example, with `a,b -> c,d` and `c -> e`, if `d` falls back to the ready direct parents `a,b`, the immutable candidates can become `a,b,e`. Since `a,b` completely cover the data represented by `e`, keeping `e` would duplicate data. Recursive deduplication therefore has to remove `e`. The final frontier may contain ancestors that are multiple graph edges away from a removed descendant because another M:N branch selected them legally, but no fallback operation crossed more than one generation.

internal/datacoord/handler.go
```diff
@@ -251,83 +251,70 @@ func retrieveSegment(validSegmentInfos map[int64]*SegmentInfo,
...
-			} else if !compactionFromExistWithCache(id) && (segmentIndexed(id) || !isVisible(compactionFrom...)) {
-				newFlushedIDs.Insert(id)
-			} else {
-				for _, fromID := range compactionFrom {
-					newFlushedIDs.Insert(fromID)
-					continueRetrieve = true
-					droppedIDs.Remove(fromID)
-				}
+		for _, parentID := range compactionFrom {
+			if !candidates.Contain(parentID) && !hasCompleteAncestorCoverage(parentID) {
```
buqian-zilliz: **[P1] Partial ancestor coverage leaves an overlapping recovery frontier**

`hasCompleteAncestorCoverage` only removes a candidate when every parent branch is covered, but it does not reject partial overlap. After an M:N compaction `a,b -> c,d` and a later compaction `c,x -> e`, current leaves can be `d,e`. If `d` falls back to ready parents `a,b` while `e` stays selected, the candidates are `a,b,e`: recursion proves the `c` branch of `e` is covered by `a,b`, then returns false because `x` is uncovered, so `e` remains. The final frontier contains the surviving `c` rows through both `a,b` and `e`, causing duplicate data to be loaded. When a descendant is only partially covered, cancel the conflicting ancestor fallback or otherwise choose a disjoint frontier.

internal/datacoord/handler.go
```diff
@@ -251,83 +251,70 @@ func retrieveSegment(validSegmentInfos map[int64]*SegmentInfo,
...
 	segmentIndexed func(segID UniqueID) bool,
 ) (typeutil.UniqueSet, typeutil.UniqueSet) {
-	newFlushedIDs := make(typeutil.UniqueSet)
-
-	isConditionMet := func(condition func(seg *SegmentInfo) bool, ids ...UniqueID) bool {
+	allParentsReady := func(ids ...UniqueID) bool {
 		for _, id := range ids {
-			if seg, ok := validSegmentInfos[id]; !ok || seg == nil || !condition(seg) {
+			seg, ok := validSegmentInfos[id]
+			if !ok || seg == nil || seg.GetIsInvisible() || !segmentIndexed(id) {
```
buqian-zilliz: **[P1] Small sorted parents bypass current-schema readiness**

`allParentsReady` now applies `segmentIndexed` to fallback parents, but that predicate treats every sorted segment below `MinSegmentNumRowsToEnableIndex` as ready even when it is absent from the current-schema `indexed` set. `FilterInIndexedSegments` builds that set by requiring indexes for all vector fields in the current collection schema, including a newly added BM25 Function output. A direct parent created before that schema change can therefore lack the backfilled function-field data yet still pass this size/sort shortcut; lines 275–276 then reintroduce it into the recovery frontier. The small-segment exemption can replace a physical index for fields that exist, but it does not establish that a stale parent contains a newly added function output. Require explicit current-schema field/data readiness for fallback parents instead of accepting the generic sorted-small shortcut.

internal/datacoord/handler.go
```diff
@@ -249,85 +250,175 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
...
+			segment := validSegmentInfos[id]
+			if segment != nil && len(segment.GetCompactionFrom()) > 0 && !blockedRoots.Contain(id) && allParentsReady(segment.GetCompactionFrom()...) {
+				toExpand.Insert(id)
 			} else {
-				for _, fromID := range compactionFrom {
-					newFlushedIDs.Insert(fromID)
-					continueRetrieve = true
-					droppedIDs.Remove(fromID)
-				}
+				toCancel.Insert(result.owners.Collect()...)
```
buqian-zilliz, line 441: **[P1] Handle ownerless overlap after partial compaction retirement**

`result.owners` only records ancestors introduced by `buildCandidates` as fallbacks. On the catalog's ordered chunked fallback, a crash can persist a new compaction output and retire only some inputs. For an output `e` compacted from `c,x`, the restarted frontier can therefore be `{e,x}` with `c` already dropped. If `c` is not current-schema-ready, `allParentsReady(c,x)` is false; coverage is partial because original candidate `x` covers one branch, but `fallbackOwners[x]` is nil, so this line adds nothing to `toCancel`. `changed` stays false and the code returns `{e,x}`, loading `x` twice (directly and through `e`). Partial-overlap handling must also attribute or suppress covered original candidates so the recovery frontier remains disjoint after partial input retirement.

internal/datacoord/handler.go
```diff
@@ -249,85 +250,175 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
...
-		}, ids...)
+	buildCandidates := func() (typeutil.UniqueSet, map[UniqueID]typeutil.UniqueSet) {
+		candidates := make(typeutil.UniqueSet)
+		fallbackOwners := make(map[UniqueID]typeutil.UniqueSet)
+		var emit func(id, rootID UniqueID, fromFallback bool)
+		emit = func(id, rootID UniqueID, fromFallback bool) {
+			segment, ok := validSegmentInfos[id]
+			if ok && segment != nil && expanded.Contain(id) {
+				for _, parentID := range segment.GetCompactionFrom() {
+					emit(parentID, rootID, true)
```
buqian-zilliz, line 336: **[P2] Memoize fallback expansion across the compaction DAG**

`emit` recursively walks every expanded ancestry path without recording that an `(id, rootID)` pair has already been visited. Successive M:N compactions can put several shared generations in `expanded`: sibling outputs share the same parents, and partial-overlap closure can expand both siblings before the next rebuild. Each additional generation then doubles the recursive calls even though `candidates` later deduplicates the same IDs, so a lineage with `O(n)` segments can require exponentially many calls while serving recovery and segment-view RPCs. Track visited IDs per fallback root, or use a memoized iterative worklist, so each ownership edge is processed once.

internal/datacoord/handler.go
```diff
@@ -249,85 +250,212 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
...
+			}
+		}
 
-	isConditionMet := func(condition func(seg *SegmentInfo) bool, ids ...UniqueID) bool {
+		switch len(presentParents) {
+		case len(segment.GetCompactionFrom()):
+			removeFromView.Insert(id)
+		case 0:
+		default:
+			removeFromView.Insert(presentParents.Collect()...)
```
buqian-zilliz, line 286: **[P1] Independent normalization can leave a cross-generation overlap**

The per-child removal decisions do not compose across interrupted generations. If `a,b -> c` is interrupted before retiring `a,b`, then a later `c,x -> e` publishes `e` and retires only `x`, the recovered set is `{a,b,c,e}`. The all-parent branch removes `c` for the first compaction, while this partial-parent branch also removes `c` but retains `e`, producing `{a,b,e}`. When `x` is missing or not current-schema-ready, overlap closure cannot expand `e`, so the rows represented by `c` are loaded both through `a,b` and through `e`. Normalize the related generations as a single disjoint frontier, or cancel the older ancestors when retaining `e`.

internal/datacoord/handler.go
```diff
@@ -249,85 +250,212 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
...
+			segment, ok := validSegmentInfos[id]
+			if !ok || segment == nil || len(segment.GetCompactionFrom()) == 0 {
+				return coverageResult{coverage: noAncestorCoverage}
 			}
-			for _, fromID := range compactionFrom {
-				if flushedIDs.Contain(fromID) || newFlushedIDs.Contain(fromID) {
-					compactionFromExistMap[segID] = true
-					return true
+
+			result := coverageResult{owners: make(typeutil.UniqueSet)}
```
buqian-zilliz: **[P3] Skip owner-map allocation in the coverage-only pass**

`newCoverageChecker` always allocates an `owners` set for every non-leaf node, but the final deduplication call passes `fallbackOwners == nil` and only reads `coverage`. Consequently, that pass allocates and caches an unused map per lineage node and runs owner-propagation loops that can never produce an owner. On channels with long compaction histories, each recovery or segment-view RPC adds `O(N)` avoidable heap objects and GC pressure. Make owner tracking optional for the coverage-only pass, or allocate the set lazily.

internal/datacoord/handler.go
buqian-zilliz: This issue still exists: `buildCandidates` still recursively revisits the same expanded lineage for a root without a visited `(id, rootID)` set, so shared M:N ancestry can still cause exponential expansion work.

internal/datacoord/handler.go
```diff
@@ -249,85 +250,234 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
...
-			if len(compactionFrom) == 0 {
-				newFlushedIDs.Insert(id)
-			} else if !compactionFromExistWithCache(id) && (segmentIndexed(id) || !isVisible(compactionFrom...)) {
-				newFlushedIDs.Insert(id)
+	// Close partial overlaps created by M:N compactions. Expand the descendant
+	// when all of its direct parents are ready; otherwise cancel the root
+	// fallback that introduced the overlapping ancestors and keep its child.
+	var candidates typeutil.UniqueSet
+	blockedRoots := make(typeutil.UniqueSet)
+	for {
```
buqian-zilliz, line 424: **[P2] Avoid quadratic full-lineage rebuilds during overlap closure**

`toExpand` advances the closure only through partial candidates visible in the current frontier, but every pass recreates `candidates`, `fallbackOwners`, `visited`, and the entire coverage cache. A compaction ladder can reveal one new partial candidate per ancestry level, so depth `D` triggers `D` full traversals of an `O(D)` lineage, plus fresh maps on each recovery or segment-view RPC. This makes the new overlap resolver `O(D²)` on long production histories. Use a worklist over one graph snapshot, or retain traversal and coverage state across rounds.

internal/datacoord/handler.go
buqian-zilliz: This issue still exists: independent normalization still turns `{a,b,c,e}` into `{a,b,e}`, and when `x` is absent or unready the partial-overlap closure has no fallback owner to cancel, so the final frontier still loads the rows represented by `c` through both `a,b` and `e`.

internal/datacoord/handler.go
aoiasd: The O(D²) bound is valid for a contrived lineage that reveals exactly one ready partial candidate per round. Each round intentionally rebuilds derived candidate, ownership, and coverage state after the expanded or blocked roots change; retaining that state safely requires an incremental dependency graph and invalidation rules, so it is not a small follow-up to the existing memoization. This RPC generates target and recovery views rather than serving the search hot path, and practical compaction fan-in, fan-out, and retained lineage depth are small. We already removed exponential repeated expansion and unused owner-map allocation. Given the low practical cost and the correctness risk of an incremental solver, we are intentionally keeping the explicit fixed-point rebuild in this PR.

internal/datacoord/handler.go
aoiasd: Confirmed; this multi-generation case is intentionally outside the scope of this PR. It requires two incomplete compaction updates to coexist: the earlier a,b to c update still retains all inputs while a later c,x to e update is only partially retired. This PR normalizes one ordered compaction update and requires current-schema readiness for every newly selected fallback. Handling multiple independently interrupted generations needs a global lineage solver and would materially broaden the change. Once metadata converges, the next view and target recomputation removes the transient overlap; we accept that extremely rare interval rather than complicating the normal path.

internal/datacoord/handler.go
```diff
@@ -249,85 +250,234 @@ func (h *ServerHandler) GetQueryVChanPositions(channel RWChannel, partitionIDs .
...
+				}
+				return
+			}
+
+			candidates.Insert(id)
+			if fromFallback {
+				if fallbackOwners[id] == nil {
+					fallbackOwners[id] = make(typeutil.UniqueSet)
+				}
+				fallbackOwners[id].Insert(rootID)
```
buqian-zilliz, line 346: **[P2] Avoid materializing the full M*N fallback ownership matrix**

`buildCandidates` records every `parentID`/`rootID` pair before it knows whether partial overlap exists. For an M:N compaction, each of M output segments carries the same N `CompactionFrom` IDs; when the outputs are unindexed and all parents are ready, this line creates N owner sets containing M roots, plus M*N `candidateVisit` entries, even when coverage finds no overlap and exits after one pass. Clustering compaction can take every segment in a view and emit roughly total-size/preferred-size outputs, so this can become quadratic heap and GC pressure in `GetQueryVChanPositions` during recovery. Keep a shared owner group for sibling outputs, or first compute coverage without owner attribution and materialize owners only after a partial overlap is detected.

aoiasd: M and N are both small in practice for the compactions handled here, and this recovery/target-view construction is not on the query hot path. The ownership matrix cost is therefore acceptable for this fix. We prefer to keep the current explicit ownership logic instead of adding a second lazy-attribution traversal, so we will leave this performance optimization out of scope.

congqixia (approved):
/lgtm

Issue #52746, opened 2026-08-21
[Bug]: CopyObject fails with "Invalid argument" on GCS when using OAuth (cloudProvider=gcp with IAM)


huanghaoyuanhhy:
**Is there an existing issue for this?**
- [x] I have searched the existing issues

**Environment**
- Milvus version: master / 3.0 (a787b9102)
- Deployment mode: standalone
- Others: GCS as object storage, `minio.cloudProvider=gcp`, `minio.useIAM=true` (OAuth 2.0 Bearer token auth, e.g. Workload Identity / GCE metadata server)

**Current Behavior**

Any server-side object copy (`ChunkManager.Copy` / `CopyObjectCrossBucket`) against GCS fails with `Invalid argument` when requests are authenticated with an OAuth Bearer token:

```
failed to copy segment files: failed to copy file from <src>/_data/<file>.parquet to <dst>/_data/<file>.parquet: invalid argument[key=<object-key>]: Invalid argument.
```

This breaks Storage V3 snapshot restore: DataCoord creates the CopySegment task (source segment is `storageVersion=3`, `hasManifestPath=true`), the worker starts copying the `_data/*.parquet` objects, and the task fails at the first `CopyObject`. `StatObject` on the same objects succeeds — only copy is rejected.

**Expected Behavior**

Server-side copy against GCS with OAuth authentication succeeds, and Storage V3 snapshot restore completes.

**Steps To Reproduce**
1. Deploy Milvus with GCS object storage, `cloudProvider=gcp`, IAM/OAuth credentials (no HMAC key).
2. Create a collection with Storage V3 enabled, insert data, flush.
3. Create a snapshot and restore it to a new collection.
4. The CopySegment task fails with `Invalid argument` on every `_data/*.parquet` object.

**Anything else?**

Root cause: with `cloudProvider=gcp` + IAM, `NewMinioClient` builds a minio-go client whose transport (`WrapHTTPTransport` in `pkg/objectstorage/gcp/gcp.go`) only replaces the `Authorization` header with a Bearer token. minio-go's `CopyObject`/`ComposeObject` send `x-amz-copy-source`, but GCS's XML API only honors `x-amz-*` headers for requests signed with HMAC keys; under OAuth 2.0 authentication they must be sent as the `x-goog-*` equivalents (`x-goog-copy-source`, etc.) per Google's S3-migration docs. So GCS rejects the copy with `Invalid argument`.

[...]

PR #52748, opened 2026-08-21
fix: translate x-amz-* headers to x-goog-* in GCS OAuth transport

huanghaoyuanhhy:
issue: #52746

With `cloudProvider=gcp` and IAM/OAuth credentials, `WrapHTTPTransport` only swapped the Authorization header for a Bearer token. minio-go's `CopyObject` sends `x-amz-copy-source`, which GCS's XML API does not honor under OAuth 2.0 authentication, so every server-side copy (e.g. Storage V3 snapshot restore CopySegment tasks) failed with `Invalid argument`.

Two changes:

- Translate all `x-amz-*` headers to their `x-goog-*` counterparts in the transport, the same approach milvus-backup's `gcpTransport` has used in production against real GCS IAM.
- `MinioObjectStorage.CopyObjectCrossBucket` no longer routes >5GiB objects through `ComposeObject` when the provider is GCP: GCS's XML API has no multipart copy (`x-amz-copy-source-range` has no `x-goog-*` equivalent), while its whole-object copy has no 5GiB cap, so GCP always takes the single-copy path.

Tests: header-translation unit test in `pkg/objectstorage/gcp`, plus a >5GiB GCP routing regression test in `internal/storage`.

### DIFF
--- a/pkg/objectstorage/gcp/gcp.go
+++ b/pkg/objectstorage/gcp/gcp.go
@@ -38,8 +38,25 @@ func NewWrapHTTPTransport(secure bool) (*WrapHTTPTransport, error) {
 	}, nil
 }
 
+const (
+	xAmzPrefix  = "X-Amz-"
+	xGoogPrefix = "X-Goog-"
+)
+
 // RoundTrip wraps original http.RoundTripper by Adding a Bearer token acquired from tokenSrc
 func (t *WrapHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// GCS's XML API only honors x-amz-* headers when the request is signed with
+	// HMAC keys. With OAuth 2.0 (Bearer token) authentication they must be sent
+	// as their x-goog-* counterparts — e.g. minio's CopyObject sends
+	// x-amz-copy-source, which GCS rejects with "Invalid argument" under Bearer
+	// auth unless it is translated to x-goog-copy-source.
+	for k, v := range req.Header {
+		if strings.HasPrefix(k, xAmzPrefix) {
+			req.Header[xGoogPrefix+strings.TrimPrefix(k, xAmzPrefix)] = v
+			delete(req.Header, k)
+		}
+	}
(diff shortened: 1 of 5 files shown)


pkg/objectstorage/gcp/gcp.go
```diff
@@ -38,8 +38,25 @@ func NewWrapHTTPTransport(secure bool) (*WrapHTTPTransport, error) {
...
+
 // RoundTrip wraps original http.RoundTripper by Adding a Bearer token acquired from tokenSrc
 func (t *WrapHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// GCS's XML API only honors x-amz-* headers when the request is signed with
+	// HMAC keys. With OAuth 2.0 (Bearer token) authentication they must be sent
+	// as their x-goog-* counterparts — e.g. minio's CopyObject sends
+	// x-amz-copy-source, which GCS rejects with "Invalid argument" under Bearer
+	// auth unless it is translated to x-goog-copy-source.
+	for k, v := range req.Header {
+		if strings.HasPrefix(k, xAmzPrefix) {
```
weiliu1031, line 54: [P2] Please avoid assuming that every `X-Amz-*` header has a direct GCS equivalent. For objects larger than 5 GiB, `MinioObjectStorage.CopyObjectCrossBucket` uses `minio.Client.ComposeObject`, which emits `x-amz-copy-source-range`. GCS explicitly documents that header as having no equivalent, so this loop rewrites it to unsupported `x-goog-copy-source-range` and the large-object snapshot copy path still fails. Please route GCP large copies through the GCS single-copy/native-copy path, or explicitly map only supported headers and handle range copies separately. A >5 GiB routing regression test would cover this boundary. See: https://docs.cloud.google.com/storage/docs/migrating?hl=en#amazon-s3-to-cloud-storage-request-headers

weiliu1031 (approved):
LGTM

PR #51246, opened 2026-07-10
enhance: add async storage v3 field data loading

sparknack:
issue: #51245

- Add an async executor and Storage V3 field-data load pipeline for manifest column groups and JSON key stats.
- Group requested cells into contiguous, byte-bounded row-group read windows and admit each window through the shared transient-memory budget.
- Preserve load priority, requested result order, cancellation, and the first storage or finalization failure across coroutine boundaries.
- Route mmap finalization to the optional local-file I/O executor while keeping the legacy synchronous path as the default.
- Add the temporary disabled-by-default rollout switch `queryNode.segcore.storageV2.enableAsyncLoad`; `storageV2` is retained only as the existing internal configuration name.

PR #52749, opened 2026-08-21
doc: add async storage v3 field-data loading design

sparknack:
issue: #51245

Related implementation: #51246

design doc: docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md

## Summary
- Document the async Storage V3 field-data loading design for QueryNode and Segcore.
- Cover read-window planning, transient-memory admission, priority, cancellation, mmap finalization, rollout, compatibility, and testing.

## Test Plan
- `git diff --check origin/master...HEAD`
- Cross-checked the design against the implementation and focused tests in #51246.

### DIFF
--- a/docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md
+++ b/docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md
@@ -1,0 +1,632 @@
+# Async Storage V3 Field-Data Loading
+
+- **Created:** 2026-08-11
+- **Author(s):** @sparknack
+- **Status:** Under Review
+- **Component:** QueryNode, Segcore, Storage, Caching Layer
+- **Related Issues:** [milvus-io/milvus#51245](https://github.com/milvus-io/milvus/issues/51245)
+- **Implementation:** [milvus-io/milvus#51246](https://github.com/milvus-io/milvus/pull/51246)
+
+## Summary
+
+This design adds an opt-in asynchronous load path for field data backed by the
+Storage V3 manifest and column-group reader. The existing caching-layer
+`Translator` interface remains synchronous, but the work behind
+`ManifestGroupTranslator::get_cells()` is decomposed into coroutine-based read
+windows that can overlap remote reads without dedicating one load worker to
+each blocking storage request.
+
+The design has five main properties:
+
+1. Requested cache cells are sorted by their physical row-group position and
+   grouped into byte-bounded, contiguous read windows.
+2. Every window acquires a process-wide transient-memory budget lease before
+   its storage read is submitted.
+3. Remote reads run through the native asynchronous `ChunkReader` API on a
+   priority-aware executor.
+4. In-memory finalization stays on the load executor, while mmap finalization
+   may move to a dedicated local-file executor so blocking file writes do not
+   occupy remote-load workers.
+5. Caller cancellation and the first sibling failure cancel pending windows;
+   failures are published before a window releases its budget lease.
+
+The path is guarded by the temporary
+`queryNode.segcore.storageV2.enableAsyncLoad` switch, which defaults to
+`false`. The legacy `LoadCellBatchAsync` path remains available during rollout.
+
+This PR covers Storage V3 field data and JSON key-stat column groups. It does
+not make scalar-index loading asynchronous.
+
+## Terminology
+
+This document follows the names used by the current Segcore implementation:
+
+| Term | Meaning |
+|---|---|
+| cache cell | The caching-layer unit returned by a translator. One cell contains one or more adjacent row groups. |
+| row group / chunk | A physical reader unit addressed by `ChunkReader`; `chunk_indices` are row-group indices in this path. |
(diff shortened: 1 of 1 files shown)


docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md
```diff
@@ -0,0 +1,632 @@
...
+`ChunkReader`, so `file_idx` is `0`. The generic planner retains `file_idx` so a
+window can never cross a physical-file boundary if other callers use it later.
+
+`memory_size` is derived from the translator's row-group size estimates. The
+translator prefers projected-column estimates, then uses the existing sampled,
+aggregate, or last-resort fallbacks. This matters for lazy projected columns:
+the window target should reflect the columns that will actually be decoded,
+not unrelated columns in the same physical column group.
+
+`loading_overhead_size` is the admission charge. It equals `memory_size` for
```
buqian-zilliz, line 229: **[P2] Admission charge for peak transient memory is conflated with decoded size**

`loading_overhead_size` is defined as the admission charge and is set equal to `memory_size` for most field types (doubled only for arrays), yet the table describes it as approximating `peak temporary memory through Arrow decoding and finalization`. The final decoded size is not a measure of peak transient memory: during the Arrow-to-`GroupChunk` conversion the source record batches, normalization temporaries, and the destination data can be live simultaneously, so the real peak can exceed the final size. Because the budget lease is held through finalization precisely to cover that window, a charge derived solely from decoded size risks under-accounting and letting the shared transient budget be exceeded in aggregate. Please specify how the overhead factor is derived per field type and confirm the charge actually covers the Arrow buffers plus finalization frame, not just the final decoded size.

docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md
```diff
@@ -0,0 +1,632 @@
...
+```cpp
+chunk_reader->get_chunks_async(window.chunk_indices, /*parallelism=*/1)
+```
+
+Window-level concurrency is owned by the Milvus pipeline. The per-call
+parallelism is set to one to avoid multiplying concurrency inside each window.
+The storage reader returns record batches in the same order as the requested
+chunk indices; the pipeline validates the returned count before finalization.
+
+The `ChunkReader` interface has a synchronous default implementation of
```
buqian-zilliz, line 339: **[P2] Synchronous fallback of get_chunks_async can silently block load-executor workers**

The doc states that `ChunkReader` has a synchronous default implementation of `get_chunks_async()`, so non-blocking behavior depends on the selected storage format providing a native async override. If the native override is absent, the async pipeline's advertised benefit is lost and, worse, a load-executor worker (which the design deliberately keeps non-blocking, default `max(1, CPU_NUM)` workers) is occupied on a synchronous storage read. The doc lists deciding whether native async must be a hard requirement as follow-up work, but this is a silent degradation of the core goal rather than a cosmetic gap. Please state how a deployment can detect or enforce that native async support is actually in use, or make the async load path conditional on verified native async support.

docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md
```diff
@@ -0,0 +1,632 @@
...
+
+1. Sum `loading_overhead_size` across its cells.
+2. Await `AcquireAsync(bytes, priority, cancellation_token)`.
+3. Move the returned RAII lease into the window coroutine.
+4. Hold it across remote read and finalization.
+5. Release it when temporary Arrow data and the finalization frame unwind.
+
+Admission has the following semantics:
+
+- capacity `0` means unlimited;
```
buqian-zilliz, line 363: **[P2] Default unlimited transient budget defeats the admission bound when async load is enabled**

`common.loadTransientBudgetBytes` defaults to `0` (unlimited), and the async load switch is an independent opt-in with no coupling to a positive budget. With capacity `0`, budget admission is a no-op and the read-window target bounds a single storage operation but not the number of simultaneously admitted windows across segments. The doc's stated goal is to `bound admitted transient data`, but the documented defaults do not provide that bound, and the safe configuration is deferred to manual rollout steps rather than enforced. Consider making the async path refuse to enable, or warn strongly, when the transient budget is unlimited.

docs/design-docs/design_docs/20260811-async-storage-v3-field-data-loading.md
```diff
@@ -0,0 +1,632 @@
...
+- high-priority waiters are considered before low-priority waiters;
```
buqian-zilliz, line 364: **[P3] No starvation protection for low-priority loads under sustained high-priority admission**

Admission considers `high-priority waiters before low-priority waiters` and the executor uses separate high/low priority queues, but the doc does not describe any cross-queue fairness or aging mechanism. Under sustained high-priority load, low-priority loads could be delayed indefinitely even though FIFO is preserved within each individual queue. If that is intentional, it is worth stating; otherwise, please note whether some aging or preemption mechanism bounds the worst-case latency for low-priority loads.

zhengbuqian (approved):

PR #49717, opened 2026-05-12
enhance: move embedding function execution before WAL append

aoiasd:
relate: #49716

## Summary
Move BM25/MinHash function execution from DN/QN pipeline embedding nodes to the write-before path before WAL append. Add collection-scoped function runner caching and keep pipeline-side compatibility fill for old insert messages without generated function output fields.

PR #51155, opened 2026-07-08
fix: avoid duplicate function runner initialization

aoiasd:
issue: #50819

## Summary
Replace the package-level function runner singleflight with entry-local initialization coordination so concurrent requests share one in-flight build only within the same runner entry.
Simplify runner entry lifecycle to NotReady -> Ready -> Close, while preserving retry after failed initialization and context-aware waiting.
Move delegator function metadata to FunctionRunnerManager so search/analyzer paths use the managed runner lifecycle.

PR #50894, opened 2026-06-29
fix: stop retrying unrecoverable streaming writes

chyezh:
issue: #50878

- streaming redo: convert dropped collection or partition waits into unrecoverable errors so stale inserts stop retrying
- streaming write errors: return unrecoverable errors for pre-WAL schema and function materialization failures
- streaming status: map terminal streaming errors to non-retryable gRPC statuses
- streaming handler: keep RO channel assignment checks retryable while waiting for RW promotion

PR #50426, opened 2026-06-10
enhance: support text_match/bm25 for text type

zhagnlu:
issue: #48783

PR #52753, opened 2026-08-21
enhance: [2.6] move embedding function execution before WAL append (#49717)

aoiasd:
issue: #49716
pr: https://github.com/milvus-io/milvus/pull/49717

## Summary
Move BM25 function execution from DN/QN pipeline embedding nodes to the write-before path before WAL append in 2.6, while retaining compatibility for legacy messages and existing 2.6 non-BM25 behavior. Include all applicable follow-up fixes from master for runner lifecycle and initialization, IDF safety, no-function and delete-only fast paths, and related race coverage.

## 2.6 scope
- Preserve the complete FunctionRunner Manager design, including Update, schema-version retention, concurrency, and lifecycle handling.
- The supported 2.6 API rejects BM25 Add, Alter, and Drop, so WAL and Delegator callers allocate, use, and release runners but do not call Manager Update.
- Keep the 2.6 legacy-message fallback and non-BM25 function-output behavior.
- Exclude master-only MinHash, growing-source, Storage V3, and dynamic BM25 caller synchronization.

## Verification
- Full Milvus build passed on the final production tree.
- Function Manager, WAL shard, QueryNode pipeline, FlushCommon pipeline/writebuffer, Storage conversion, streaming message/typeutil, focused Delegator, and Manager race tests passed.


internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go
```diff
@@ -146,10 +159,19 @@ func (impl *shardInterceptor) handleInsertMessage(ctx context.Context, msg messa
 	// Assign segment for insert message.
 	// !!! Current implementation a insert message only has one parition, but we need to merge the message for partition-key in future.
 	header := insertMsg.Header()
+	collectionID := header.GetCollectionId()
+	schemaVersion := function.LatestFunctionRunnerVersion
+	if err := impl.materializeFunctionFields(ctx, insertMsg, collectionID, schemaVersion); err != nil {
```
buqian-zilliz, line 164: **[P1] Function runner schema is never refreshed on schema change, so WAL materialization can use stale BM25 definitions**

`FunctionRunnerManager.Update` is only referenced from tests; no production caller exists. `NewShardDelegator` calls `Alloc` once at construction, and the WAL interceptor calls `Alloc` only in `handleCreateCollection`/recovery. `handleAlterCollection` and `handleSchemaChange` only flush/fence segments and never extract the new `Schema` from the alter body to call `Update`, and `shardManagerImpl.CollectionInfo.Schema` is never rewritten after create/recovery either. Because `handleInsertMessage` materializes with `LatestFunctionRunnerVersion`, inserts that follow an ALTER of a BM25 function (for example, changed `analyzer_params` or input/output fields) will be materialized before WAL append using the pre-alter runner definition, silently producing the wrong sparse output. Please wire the schema-change paths to `Update` (and update the shard manager's stored schema) before this is shipped.

internal/util/function/manager.go
```diff
@@ -0,0 +1,1292 @@
...
+		}
+		writeFieldSignature(hasher, "output_name", field)
+		if len(outputIDs) == 0 {
+			resolvedOutputIDs = append(resolvedOutputIDs, field.GetFieldID())
+		}
+	}
+	return hex.EncodeToString(hasher.Sum(nil)), resolvedInputIDs, resolvedOutputIDs, nil
+}
+
+func writeFieldSignature(hasher hashWriter, prefix string, field *schemapb.FieldSchema) {
```
buqian-zilliz, line 1115: **[P2] Runner cache signature omits field type_params, so analyzer configuration changes do not invalidate the cached runner**

`embeddingFunctionMetadata`/`writeFieldSignature` hashes only `FieldID`, name, `DataType`, `ElementType`, and `IsFunctionOutput`; it does not hash `TypeParams`. `NewBM25FunctionRunner` builds its tokenizer from the input field's `TypeParams` (the `analyzer_params` value). A same-schema-version refresh that only changes `analyzer_params` therefore produces an identical signature, so the cached `FunctionRunnerEntry` is reused and WAL materialization keeps the old tokenizer. Include field `TypeParams` (and any other runner-affecting field metadata) in the signature, or rebuild the runner when those properties change.

internal/querynodev2/delegator/delegator.go
```diff
@@ -390,22 +410,13 @@ func (sd *shardDelegator) search(ctx context.Context, req *querypb.SearchRequest
...
-		avgdl, err := sd.buildBM25IDF(req.GetReq())
-		if err != nil {
-			return nil, err
-		}
-
-		if avgdl <= 0 {
-			log.Warn("search bm25 from empty data, skip search", zap.String("channel", sd.vchannelName), zap.Float64("avgdl", avgdl))
-			return []*internalpb.SearchResults{}, nil
-		}
+	avgdl, skipSearch, err := sd.prepareSearchFunction(ctx, req.GetReq())
```
buqian-zilliz, line 425: **[P3] Every search now takes manager/entry locks and does map lookups even for non-function output fields**

`search()` now calls `prepareSearchFunction` unconditionally, and `prepareSearchFunction` calls `RunWithRunner`, which acquires the manager `RLock`, the collection-entry `RLock`, and performs multiple map lookups before it can decide the field is not a function output. The previous implementation was a single `sd.isBM25Field[fieldId]` map read. For non-BM25 collections (the common case) this adds lock traffic and map lookups to the hot search path. Consider a cheap fast-path check (for example, cache whether the collection has embedding functions on the delegator) before entering the manager.

internal/util/function/manager.go
```diff
@@ -0,0 +1,1292 @@
...
+	}
+	e.mu.RUnlock()
+	if field == nil {
+		return false, merr.WrapErrServiceUnavailableMsg("field %d is not available in function runner schema for key %s at schema version %d", fieldID, key, schemaVersion)
+	}
+	if !typeutil.CreateFieldSchemaHelper(field).EnableAnalyzer() {
+		return false, nil
+	}
+
+	analyzer, err := NewAnalyzerRunner(field)
```
buqian-zilliz, line 817: **[P3] Non-BM25 analyzer fields now build a short-lived analyzer on every request instead of reusing a cached runner**

For analyzer-enabled fields that are not backed by a BM25 runner, `RunWithAnalyzer` falls through to `NewAnalyzerRunner(field)` per request (and then closes it). The removed `NewShardDelegator` code built `analyzerRunners` once at delegator construction and reused them. `RunAnalyzer` and `GetHighlight` will now pay analyzer/tokenizer construction on each call; if these endpoints are used at non-trivial frequency this is a measurable regression. Consider caching these analyzers with the same lifecycle handling as the function runners.

internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go
aoiasd: No production change here by design. On 2.6, Proxy explicitly rejects BM25 Add, Alter, and Drop Function operations in internal/proxy/function_task.go, so a supported request cannot change the WAL BM25 definition after CreateCollection. The generic FunctionRunnerManager.Update implementation and tests are retained, but wiring Update into WAL or delegator would reintroduce the unsupported dynamic BM25 lifecycle removed from this 2.6 adaptation.

internal/util/function/manager.go
aoiasd: Fixed in ec4bd0d068. Field signatures now include the two runner-affecting type params, analyzer_params and multi_analyzer_params, while unrelated params such as mmap.enabled still preserve runner reuse. Added signature regressions plus a Manager.Update test proving an analyzer change builds a new runner and retires the old one.

internal/querynodev2/delegator/delegator.go
aoiasd: Keeping the managed lookup intentionally. This matches master commit 91f7a255b4, where search acquires the same runner lifecycle lease used by WAL and delegator paths instead of maintaining a second local function-metadata cache. The added work is read locks and in-memory lookups before distributed search; without benchmark evidence of a material regression, duplicating lifecycle state for a speculative fast path would be riskier.

internal/util/function/manager.go
aoiasd: Keeping the current master lifecycle behavior intentionally. BM25 input analyzers still reuse the cached managed BM25 runner through analyzerFieldSignatures. Only standalone analyzer-enabled fields use the short-lived path, which closes the analyzer after the callback; the removed 2.6 delegator cache did not close standalone analyzer runners in Close. This avoids stale or leaked analyzer ownership, and there is no measured regression supporting restoration of a second unmanaged cache.

internal/flushcommon/writebuffer/write_buffer.go
```diff
@@ -769,3 +783,40 @@ func PrepareInsert(collSchema *schemapb.CollectionSchema, pkField *schemapb.Fiel
...
+		outputFieldIDs = append(outputFieldIDs, outputField.GetFieldID())
+	}
+	return outputFieldIDs, nil
+}
+
+func appendBM25StatsFromInsertData(stats map[int64]*storage.BM25Stats, outputFieldIDs []int64, data *storage.InsertData) error {
+	for _, outputFieldID := range outputFieldIDs {
+		outputData, ok := data.Data[outputFieldID]
+		if !ok {
+			return merr.WrapErrServiceInternalMsg("BM25 output field %d not found in insert data", outputFieldID)
```
buqian-zilliz, line 786: **[P2] Row-based BM25 inserts now fail in the data-node write path**

`PrepareInsert` now computes BM25 stats via `appendBM25StatsFromInsertData`, which requires each BM25 output field to already be present in `storage.InsertData.Data` and otherwise returns `"BM25 output field %d not found in insert data"`. For row-based messages, `storage.InsertMsgToInsertData` dispatches to `RowBasedInsertMsgToInsertData(msg, schema, true)`, whose `skipFunction` branch skips `isEmbeddingFunctionOutputField` columns, so that output field is never populated. The preceding `writeNode.Operate` fallback (`FillEmbeddingData` -&gt; `getStringFieldData`) cannot compensate because row-based messages carry no `FieldsData`. The removed `embeddingNode` ran the BM25 function over `storage.InsertData` and handled row-based inserts. If legacy row-based inserts can still reach the data node on 2.6, this panics the write node on valid input; otherwise consider explicitly rejecting row-based BM25 inserts.

internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go
buqian-zilliz, line 164: **[P3] Every WAL insert now takes manager and collection-entry read locks even for collections without functions**

`handleInsertMessage` now calls `materializeFunctionFields` unconditionally before segment assignment. For collections without embedding functions (the common case), `FunctionRunnerManager.Materialize` still acquires the process-wide manager `RLock` via `getEntry`, then the collection-entry `RLock`, and performs `keyVersions`/`versionRunners` map lookups before returning early because `outputFieldIDs` is empty. On a busy pchannel this adds two read-lock acquisitions and several map lookups to the hottest WAL append path for every insert. Consider a cheap per-collection `has embedding functions` flag (derived from the schema already held in `CollectionInfo`) so the interceptor can skip the manager entirely when no function output exists.

internal/flushcommon/writebuffer/write_buffer.go
aoiasd: Thanks. I traced this through the original 2.6 path end to end, and this is not a compatibility regression. RowBasedInsertMsgToInsertData with skipFunction enabled skips the BM25 sparse output, but its decoder has no VarChar or Text case. BM25 input validation requires exactly such a text field, so the removed DataNode embedding node also failed when it looked up that input in the converted storage InsertData. The current path fails earlier because row-based requests have no FieldsData, but no previously accepted 2.6 BM25 row-based path was lost. A post-conversion materializer would still lack the text input unless we also introduced new row-based text encoding and decoding semantics, which is outside this function-execution-only CP. No change.

internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go
aoiasd: Thanks. I am keeping this unchanged. Current master unconditionally uses the same WAL materialization call and the same manager and collection-entry read-side lookups before returning for a missing lifecycle or empty output set. There is no benchmark, profile, or CI evidence showing a material throughput regression here. Adding a separate shard-side hasFunction flag would duplicate collection lifecycle and schema state and add another synchronization invariant for a speculative optimization. We can optimize this later if measurements show it is needed.

internal/querynodev2/delegator/delegator.go
aoiasd: Follow-up after CI: fixed in ba787260d2. The prior response treated this only as a performance concern, but the static 2.6 Manager snapshot made it a correctness issue for AddField: every search reached Manager, so a newly added ordinary vector field was rejected as ServiceUnavailable. The delegator now keeps creation-time, field-specific BM25 eligibility and bypasses Manager for non-BM25 fields, matching the original 2.6 execution condition while retaining the Manager strict missing-field contract and complete Update semantics. Regression coverage verifies the bypass.

internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go
```diff
@@ -146,10 +159,19 @@ func (impl *shardInterceptor) handleInsertMessage(ctx context.Context, msg messa
...
+		impl.shardManager.Logger().Warn("failed to materialize function fields before WAL append",
+			zap.Int64("collectionID", collectionID),
+			zap.Int32("schemaVersion", schemaVersion),
+			zap.Error(err))
+		return nil, status.NewUnrecoverableError("failed to materialize function fields before WAL append: %s", err.Error())
```
buqian-zilliz, line 169: **[P3] Materialization errors are classified as unrecoverable, bypassing the redo retry path**

In `handleInsertMessage`, any error from `materializeFunctionFields` is returned as `status.NewUnrecoverableError`, so the message is rejected permanently rather than retried. The manager is explicitly designed to retry runner initialization on the next request (`warnInitFailure` logs "will retry on next request"), so a transient failure (for example a runner that fails to initialize once, or a `context` deadline hit while `GetRunner` waits on init) drops that insert instead of being re-queued via `redo.ErrRedo` like the segment-assignment errors immediately below. Consider classifying transient materialization failures as redo so the append can be retried.

internal/querynodev2/pipeline/insert_node.go
```diff
@@ -129,13 +151,81 @@ func newInsertNode(
...
+		functionStore: function.NewFunctionRunnerLocalStore(),
+	}
+	if _, err := iNode.functionStore.OutputFieldIDs(schema); err != nil {
+		return nil, err
+	}
+	return iNode, nil
+}
+
+func (iNode *insertNode) appendBM25Stats(iData *delegator.InsertData, msg *InsertMsg, schema *schemapb.CollectionSchema) error {
+	outputFieldIDs, err := getBM25OutputFieldIDs(schema)
```
buqian-zilliz: **[P3] BM25 output field IDs are recomputed for every insert message on the query node**

`insertNode.addInsertData` calls `appendBM25Stats`, which recomputes the BM25 output field IDs by calling `getBM25OutputFieldIDs(schema)` (a scan over `schema.GetFunctions()` plus `typeutil.GetField`) once per insert message. `Operate` already computed the same set once per batch through `iNode.functionStore.OutputFieldIDs(schema)`, which is cached by schema version. Pass the precomputed `functionOutputFieldIDs` into `appendBM25Stats` instead of recomputing it per message on the hot insert path.

internal/streamingnode/server/wal/interceptors/shard/shard_interceptor.go
aoiasd: Verified against current master and the originating fix. Master still deliberately returns an unrecoverable error here; fc0054b75c (#50894) changed function materialization from retryable/inner to unrecoverable specifically to stop replaying terminal pre-WAL failures. In 2.6 the managed execution set is BM25 only, so schema/analyzer/payload failures are deterministic for the same append, and redo has no materialization wait/backoff and could hot-loop. A context error after WAL readiness is tied to WAL shutdown and is not recoverable by redo. Manager retry-on-next-request allows a later independent request to rebuild after failed initialization; it does not require replaying the same write. Keeping the master classification unchanged.

internal/querynodev2/pipeline/insert_node.go
aoiasd: Fixed in 2c609e5711. QueryNode now passes the once-per-batch FunctionRunnerLocalStore.OutputFieldIDs result through addInsertData to appendBM25Stats, and the per-message schema/function scan was removed. On 2.6 IsEmbeddingFunctionType admits only BM25, so this reuses the same valid-schema field set without widening execution conditions; missing field and sparse-vector payload validation remain unchanged.

internal/util/function/local_store.go
```diff
@@ -0,0 +1,106 @@
...
+		runners, err := BuildEmbeddingRunners(schema)
+		if err != nil {
+			return err
+		}
+		defer CloseRunners(runners)
+		_, err = FillFunctionFields(runners, body)
+		return err
+	}
+	schemaVersion := schema.GetVersion()
+	_, ok, err := GetManager().TryMaterialize(context.Background(), collectionID, schemaVersion, body)
```
buqian-zilliz, line 90: **[P3] Consumer fallback materialization blocks with context.Background and panics on runner-init failure**

`FunctionRunnerLocalStore.FillEmbeddingData` calls `GetManager().TryMaterialize(context.Background(), ...)`. On the query node the manager entry is registered by `NewShardDelegator.Alloc`, so a legacy insert message that arrives while the managed runner is still asynchronously initializing (or whose initialization failed) blocks the single-goroutine insert pipeline in `runWithRunnerEntries` -&gt; `GetRunner` -&gt; `waitInit` with no deadline or cancellation. A runner-init error then propagates back through `TryMaterialize` and is turned into a `panic` in `insertNode.Operate` instead of falling back to a locally built runner. Consider threading the pipeline context (or a bounded timeout) into `FillEmbeddingData`/`TryMaterialize` so the compatibility path cannot stall or crash the node.

pkg/streaming/util/message/specialized_message.go
```diff
@@ -188,6 +189,28 @@ func (m *specializedMutableMessageImpl[H, B]) OverwriteBody(body B) {
 	if err != nil {
 		panic(fmt.Sprintf("failed to marshal specialized body, %s", err.Error()))
 	}
+	if ch := m.cipherHeader(); ch != nil {
+		cipher := mustGetCipher()
+		encryptor, safeKey, err := cipher.GetEncryptor(ch.EzId, ch.CollectionId)
```
buqian-zilliz, line 194: **[P3] OverwriteBody now requires encryptor capability on the streaming node and panics on failure**

For encrypted collections, the new pre-WAL materialization path calls `OverwriteBody`, which re-encrypts the rewritten body via `cipher.GetEncryptor`. Before this change the WAL interceptor only needed to decrypt bodies to read them; it never needed write-side crypto. If the streaming node's cipher hook does not grant encryptor access for the collection's encryption zone, `GetEncryptor` fails and the `panic` in `OverwriteBody` crashes the interceptor/streaming node rather than returning a recoverable append error. Please confirm the streaming node is provisioned with encryptor access for encrypted collections, and prefer returning an error over panicking.

internal/util/function/local_store.go
aoiasd: Verified against current master and #51155. Master intentionally calls TryMaterialize(context.Background(), ...) here, and TestFunctionRunnerManagerTryMaterializeInitializesRetainedVersion explicitly requires an exact retained version to initialize/use its managed runner. The context only cancels waitInit; BuildEmbeddingRunner itself has no context and continues, while both QN/DN callers panic on any returned materialization error. A local fallback would synchronously build the same BM25-only runner and restore the duplicate-initialization behavior #51155 removed. Fallback remains only for a schema version absent from the Manager, so no 2.6-specific change is needed.

pkg/streaming/util/message/specialized_message.go
aoiasd: Confirmed the StreamingNode provisioning path. streaming.Init registers hookutil.GetCipher() with the message package on the StreamingNode process. The canonical milvus-cloud-plugin explicitly lists streamingnode in RemoteServers, creates a full KeyManagerImpl for that role, and CipherProvider delegates both GetEncryptor and GetDecryptor to it; its StreamingNode tests obtain an encryptor, encrypt data, then decrypt it successfully. The original master commit d79a05bec6 also added the encrypted OverwriteBody round-trip test, and current master keeps this exact implementation. OverwriteBody is a generic Must-style void mutator whose marshal/cipher failures panic by contract; changing it to an error-returning API would be a broad message semantic change rather than a 2.6 Function adaptation. No change here.

internal/flushcommon/writebuffer/write_buffer.go
```diff
@@ -697,6 +697,11 @@ func (wb *writeBufferBase) Close(ctx context.Context, drop bool) {
 // prepareInsert transfers InsertMsg into organized InsertData grouped by segmentID
 // also returns primary key field data
 func PrepareInsert(collSchema *schemapb.CollectionSchema, pkField *schemapb.FieldSchema, insertMsgs []*msgstream.InsertMsg) ([]*InsertData, error) {
+	bm25OutputFieldIDs, err := getBM25OutputFieldIDs(collSchema)
```
buqian-zilliz: **[P3] Data node re-derives BM25 output field IDs on every PrepareInsert batch**

`PrepareInsert` calls `getBM25OutputFieldIDs(collSchema)` on each batch and re-scans `schema.GetFunctions()` plus `typeutil.GetFunctionOutputField`, even though `writeNode.Operate` already computed the same set via `functionStore.OutputFieldIDs(currentSchema)` (cached by schema version) for the identical `currentSchema` it passes into `PrepareInsert`. This duplicates the derivation on the data-node write path and gives the two paths a chance to drift. Consider passing the precomputed output field IDs into `PrepareInsert` (or a shared helper) instead of re-deriving them per batch.

internal/streamingnode/server/wal/interceptors/shard/function_materializer.go
```diff
@@ -0,0 +1,46 @@
...
+	return "WAL-" + vchannel
+}
+
+func (impl *shardInterceptor) allocFunctionRunners(collectionID int64, vchannel string, schema *schemapb.CollectionSchema) {
+	key := walFunctionRunnerKey(vchannel)
+	schemaVersion := function.LatestFunctionRunnerVersion
+	if schema != nil {
+		schemaVersion = schema.GetVersion()
+	}
+	if err := function.GetManager().Alloc(collectionID, key, schema); err != nil {
```
buqian-zilliz, line 24: **[P3] Function-runner allocation errors are swallowed at WAL create/recovery, deferring failure to downstream insert/search**

`allocFunctionRunners` discards the `function.GetManager().Alloc` error and only logs a warning. `Alloc` returns synchronously only for metadata-validation failures in `buildFunctionRunnerVersion`; actual runner construction (`BuildEmbeddingRunner` -&gt; `NewBM25FunctionRunner`) is deferred to a background goroutine whose error is likewise only logged as `will retry on next request`. The removed data-node/query-node embedding nodes built runners synchronously and failed setup on error. As a result an invalid/unbuildable BM25 definition can be accepted at create/recovery: a synchronous metadata error leaves the WAL lifecycle key unregistered, so `Materialize` returns `(false, nil)` for `LatestFunctionRunnerVersion` and inserts are written without the BM25 output, later failing in `appendBM25StatsFromInsertData`/`appendBM25Stats` with `BM25 output field ... not found`; an async build error instead rejects each insert as unrecoverable. Please confirm this lenient error surface is intended and that downstream paths degrade cleanly rather than panicking.

internal/flushcommon/writebuffer/write_buffer.go
aoiasd: Fixed in 7f4c338c11. DataNode now passes the once-per-batch FunctionRunnerLocalStore.OutputFieldIDs(currentSchema) result into PrepareInsert, and the per-batch getBM25OutputFieldIDs schema/function scan was removed. On 2.6 IsEmbeddingFunctionType admits only BM25, so the supplied set is identical for valid schemas; the existing missing-field and sparse-vector validation in appendBM25StatsFromInsertData remains unchanged. This only reuses metadata and does not widen FunctionRunner execution conditions.

internal/streamingnode/server/wal/interceptors/shard/function_materializer.go
aoiasd: Confirmed: this is the intentional recovery contract and matches current master. FunctionRunnerManager.Alloc explicitly returns metadata-validation failures synchronously but logs/retries runner-construction failures so collection recovery is not failed; the original move-before-WAL implementation already initialized asynchronously and logged failures, and #51155 preserved that policy. CreateCollection calls allocation only after the DDL has been appended and installed, while interceptor Build has no error return, so propagating Alloc here would either report an already-committed DDL as failed or require a broader recovery/atomicity redesign. Supported CreateCollection requests are validated at Proxy (function fields/types/params plus embedding runtime validation) before reaching WAL, so a synchronous metadata failure indicates a corrupt/internal schema rather than a valid user request. An asynchronous build failure is returned by the next Materialize and rejects that insert before WAL; later requests retry initialization. The missing-Latest-key no-op is also required by 2.6's pre-existing schema-less CreateCollection compatibility. Tightening either case here would diverge from master and break that 2.6 compatibility, so no production change is warranted in this scoped port.

zhengbuqian (approved):

internal/util/function/manager.go
```diff
@@ -0,0 +1,1298 @@
...
+func (m *functionRunnerManager) Materialize(
+	ctx context.Context,
+	collectionID int64,
+	key string,
+	schemaVersion int32,
+	message InsertMessage,
+) (bool, error) {
+	entry := m.getEntry(collectionID)
+	if entry == nil {
+		if schemaVersion == LatestFunctionRunnerVersion {
```
chyezh, line 881: **[Auto Review · Unverified] High — silent materialization skip when the WAL lifecycle entry is missing (H1)**

`FunctionRunnerManager.Materialize` returns `(false, nil)` when `entry == nil && schemaVersion == LatestFunctionRunnerVersion`, and `handleInsertMessage` (shard_interceptor.go:163-170) treats "no error" as success, so the insert is appended **without the BM25 output column**. Downstream, `appendBM25StatsFromInsertData` (write_buffer.go:782-799) errors on the missing field and `writeNode.Operate` panics (flow_graph_write_node.go:113-116), or the message may slip through the skipFunction conversion path and produce a segment with a missing sparse-vector column.

Entry-missing windows: (1) `allocFunctionRunners` Alloc failure is only logged (function_materializer.go:24-31, thread T12); (2) recovery snapshot without schema — `builder.go` only Allocs for non-nil schemas, and `newCollectionInfos` (shard_manager.go:160-166) sets `Schema=nil` when `Schemas` is empty; (3) schema-less CreateCollection skips Alloc (safe).

Note this is the opposite of the T12 assumption "failure is deferred to downstream insert/search": with a missing entry the insert does **not** fail downstream — it silently proceeds un-materialized.

Suggestion: return a (retriable) error from `Materialize` when the collection schema declares runner-backed functions but the key is absent, or log an error per insert; add an interceptor test where the manager has no entry + schema contains BM25 → insert must not silently skip.

internal/querynodev2/delegator/delegator.go
```diff
@@ -234,6 +232,37 @@ func (sd *shardDelegator) Stopped() bool {
...
+			return runErr
+		default:
+			return merr.WrapErrServiceInternalMsg("unsupported managed function type %s", functionType.String())
+		}
+	})
+	if err != nil {
+		return 0, false, err
+	}
+	if !ok {
+		return 0, false, nil
```
chyezh: **[Auto Review · Unverified] Medium — BM25 search silently continues without IDF when the managed runner is unavailable (M1)**

`prepareSearchFunction` returns `(0, false, nil)` when `RunWithRunner` reports `ok=false` (delegator.go:236-252). The search then proceeds **without IDF weighting and without metric-type validation** — the old code failed closed ("functionRunner not found for field: %d"). `ok=false` only occurs when the field exists in the schema but not in the current version snapshot of function outputs; with the static 2.6 lifecycle this should be unreachable, but if it ever happens (entry released mid-flight, schema/version mismatch) users get silently wrong results instead of an error.

Suggestion: return an explicit error for the `!ok` branch (or document it as unreachable and log an error); add a `TestDelegatorFunctionExecutionEligibility` case where the field is in `isBM25Field` but the manager has no runner → search must error.

internal/util/function/local_store.go
```diff
@@ -0,0 +1,106 @@
...
+	"context"
+
+	"github.com/milvus-io/milvus-proto/go-api/v2/msgpb"
+	"github.com/milvus-io/milvus-proto/go-api/v2/schemapb"
+)
+
+// FunctionRunnerLocalStore owns short-lived runners used when a managed runner
+// is unavailable. It is intended for consumer-side old insert message handling.
+type FunctionRunnerLocalStore struct {
+	runners        map[int32][]FunctionRunner
```
chyezh, line 31: **[Auto Review · Unverified] Medium — FunctionRunnerLocalStore maps are unsynchronized (M3)**

`runners` / `outputFieldIDs` are plain maps with no lock. This is safe only because both call sites (writeNode.Operate, insertNode.Operate) are single-goroutine, but nothing documents that precondition; a future concurrent caller would race. Suggestion: add a "not concurrency-safe; caller must serialize" comment or a mutex. (Existing thread T9 covers the context.Background/panic aspect, not this.)

internal/querynodev2/delegator/idf_oracle.go
```diff
@@ -264,13 +277,15 @@ func (o *idfOracle) preloadSealed(segmentID int64, stats *sealedBm25Stats, memor
 }
 
 func (o *idfOracle) RegisterGrowing(segmentID int64, stats bm25Stats) {
```
chyezh, line 266: **[Auto Review · Unverified] Low — bm25Stats.Clone() skips nil entries but RegisterGrowing still merges the ORIGINAL map (L1)**

`clonedStats := stats.Clone()` skips nil entries (the #50584 fix), but `o.current.Merge(stats)` at line ~287 merges the original map. If a stats map ever contains a nil `*BM25Stats` entry, `Merge` panics on a nil receiver (`meta.NumRow()` at stats.go:363-368). All current producers create non-nil entries so this is latent; suggest skipping nil entries in `Merge`/`Minus` defensively, or a regression test passing `map[int64]*storage.BM25Stats{102: nil}` to RegisterGrowing.

pkg/util/typeutil/schema.go
```diff
@@ -2623,6 +2623,16 @@ func GetFieldByName(schema *schemapb.CollectionSchema, fieldName string) *schema
 	return nil
 }
 
+func GetFunctionOutputField(schema *schemapb.CollectionSchema, fn *schemapb.FunctionSchema) *schemapb.FieldSchema {
```
chyezh: **[Auto Review · Unverified] Low — GetFunctionOutputField has no production caller (L2)**

This new exported helper is referenced only from manager_test.go:1717; no production code uses it. Either use it (e.g., output-field resolution in manager.go) or drop it to avoid dead API surface.

internal/querynodev2/delegator/delegator.go
aoiasd: Fixed in 8766ef429b. The 2.6 creation-time isBM25Field map and the managed runner snapshot are derived from the same schema, so ok=false for an eligible BM25 output is an internal invariant violation. prepareSearchFunction now fails closed with ServiceInternal, matching the original 2.6 functionRunner-not-found behavior, while newly added non-BM25 vector fields still bypass the Manager. Added a regression case with BM25 eligibility and a managed schema that has no runner.

internal/querynodev2/delegator/idf_oracle.go
aoiasd: Fixed in 8766ef429b. RegisterGrowing now merges clonedStats rather than the caller-owned original map, matching current master. Because Clone filters nil entries, a nil BM25Stats entry can no longer reach Merge from this path. Added a regression test that registers map[int64]*storage.BM25Stats{102: nil} and verifies registration does not panic.

pkg/util/typeutil/schema.go
aoiasd: Fixed in 8766ef429b. This helper has production callers on current master, but the 2.6 adaptations removed those per-message/per-batch schema scans and reuse FunctionRunnerLocalStore output metadata instead. That left the helper test-only in this port, so the exported helper was removed and the test resolves its output field locally.

internal/util/function/manager.go
aoiasd: Verified this is not reachable for a 2.6 collection that declares BM25, so the Manager contract remains unchanged. Normal persisted and upgrade recovery require a collection schema: initialization fetches every live collection schema and treats a missing one as an internal invariant failure, while CreateCollection recovery stores the resolved schema. A snapshot with no schema is the intentional legacy schema-less compatibility case and cannot declare BM25. A synchronous Alloc error requires malformed internal function metadata already rejected on the supported Proxy create path; an asynchronous runner-build error retains the entry/key and Materialize returns that error before WAL append. Collection release is also ordered behind exclusive DDL. Current master keeps the same Latest missing-entry no-op, and 2.6 needs it for schema-less CreateCollection, so tightening it here would reject a valid legacy path rather than fix a valid BM25 path.

internal/util/function/local_store.go
aoiasd: Addressed in 8766ef429b by documenting the ownership contract: FunctionRunnerLocalStore is not concurrency-safe, and each consumer pipeline node owns one instance and invokes it serially. Both the QueryNode stream pipeline and DataNode flowgraph execute Operate from a single worker goroutine, so adding a mutex would add synchronization to a serialized hot path without changing safety.

internal/querynodev2/delegator/idf_oracle.go
```diff
@@ -264,16 +277,18 @@ func (o *idfOracle) preloadSealed(segmentID int64, stats *sealedBm25Stats, memor
...
+	clonedStats := stats.Clone()
```
buqian-zilliz: **[P3] RegisterGrowing deep-copies full BM25 stats on the growing-segment load path**

`RegisterGrowing` now calls `clonedStats := stats.Clone()` before registering a growing segment. `bm25Stats.Clone()` deep-copies the whole map, and each `storage.BM25Stats.Clone()` clones its entire `rowsWithToken` map, so the cost is O(unique-token-count) in both CPU and peak memory. On the `LoadGrowing` recovery/load path the argument is the segment's complete accumulated stats (`segment.GetBM25Stats()`), so every loaded growing segment pays a full-map clone plus a second full copy of the stats. This snapshot is a correct fix for the later in-place `UpdateBM25Stats` mutation, but for a large growing segment (millions of unique tokens) the load-path cost is non-trivial. Please confirm this is acceptable at production scale, or consider taking the snapshot more cheaply (for example, cloning only after the duplicate-registration check, or avoiding the extra copy when the segment is not actually retained).

aoiasd: Confirmed against current origin/master at 5f7f1f199f: RegisterGrowing still clones the full stats snapshot before the duplicate-registration check and merges that clone. This is therefore shared master behavior, not a 2.6-specific expansion introduced by this backport. The clone is required to detach oracle-owned growing/current state from the segment-owned stats that can later be mutated in place. Moving it after the duplicate check could reduce duplicate-registration cost, but doing that only in 2.6 would create an unmeasured branch-specific optimization outside this function-execution backport. Keeping the master behavior unchanged here; any optimization should be profiled and fixed on master first, then backported.

buqian-zilliz: This issue still exists. `RegisterGrowing` still calls `clonedStats := stats.Clone()` unconditionally (`internal/querynodev2/delegator/idf_oracle.go:280`), deep-copying the full `rowsWithToken` map for every field before the duplicate-registration check under the lock. On the `LoadGrowing` recovery/load path the argument is each segment's complete accumulated `segment.GetBM25Stats()`, so the cost is O(unique-token-count) in both CPU and peak memory per loaded growing segment, and the clone is also discarded without use when the segment is already registered.

aoiasd: Fixed at `5d45eb0ba8`. Re-auditing the commit lineage showed that `bm25Stats.Clone` came from the separate master IDF-ownership change in #50426; it was not introduced by the Function-before-WAL source commit #49717. This 2.6 port has no production `UpdateBM25Stats` caller or the rest of that ownership redesign, so retaining only the Clone widened the backport and added the reported load cost. I removed `bm25Stats.Clone`, restored `RegisterGrowing` exactly to official 2.6 ownership semantics, and removed the three Clone-only tests. `idf_oracle.go` and its test now have no PR diff from current `origin/2.6`.

Issue #52725, opened 2026-08-20
[Bug]: index-cluster-worker SIGSEGV during concurrent HNSW/Disk index build (VectorDiskIndex.cpp:469 / VectorMemIndex.cpp:294) — persistent crash loop


marcelo-cjl:
[Bug]: index-cluster-worker SIGSEGV during concurrent HNSW/Disk index build (VectorDiskIndex.cpp:469 / VectorMemIndex.cpp:294) — persistent crash loop

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: index-cluster-worker image (uat-milvus-us-west-2, index-pool namespace)
- Deployment mode: index pool (dedicated index-cluster-worker, arm, slots=64)
- Pod: index-worker-arm-index-pool-slots-64-a13202972141-96c5bd6bhq7nn
- SDK: pymilvus
- Index types: HNSW (M=30, efConstruction=360, vamana, COSINE, dim=36, storage_version=2) via Disk/Mem index path
```

### Current Behavior

`index-cluster-worker` (shared index pool serving multiple clusters) **repeatedly SIGSEGVs during concurrent vector index builds**. The process crashes, restarts, re-runs the same build tasks, and crashes again — a persistent crash loop. All in-flight index build tasks are left failed/unfinished.

Crash signatures (two alternating C++ frames):

```
# Disk index path:
	/root/milvus/internal/core/src/index/VectorDiskIndex.cpp:469 pc=0x...
	/root/milvus/internal/core/src/indexbuilder/index_c.cpp:374 pc=0x...

# Mem index path:
	/root/milvus/internal/core/src/index/VectorMemIndex.cpp:294 pc=0x...
	/root/milvus/internal/core/src/index/VectorMemIndex.cpp:307 pc=0x...
	/root/milvus/internal/core/src/indexbuilder/VecIndexCreator.cpp:110 pc=0x...
	/root/milvus/internal/core/src/indexbuilder/index_c.cpp:1066 pc=0x...
```

Signal: `SIGNO: 11; SIGNAME: Segmentation fault; SI_CODE: 1; SI_ADDR: 0xd08 / 0x2860 / (nil)`

### Expected Behavior

Concurrent index builds on an index-cluster-worker should complete independently. A build failure should return an error; a single task must not take down the whole worker process and strand all other in-flight builds.

### Steps To Reproduce

[...]

Issue #52724, opened 2026-08-20
[Bug]: SIGSEGV (null deref) on all QueryNodes during concurrent CARDINAL_TIERED index load (SealedIndexTranslator.cpp:234)


marcelo-cjl:
[Bug]: SIGSEGV (null deref) on all QueryNodes during concurrent CARDINAL_TIERED index load (SealedIndexTranslator.cpp:234)

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Environment

```markdown
- Milvus version: UAT image (ali-vdc-uat cluster, vectorlake-kite namespace, kite-querynode)
- Deployment mode: cluster
- MQ type: rocksmq
- OS: linux/arm64
- Index type: CARDINAL_TIERED (dim 768/1536, L2, emb_list, mmap enabled, separate_quant_storage)
- QueryNode image: querynode (kite-querynode), 2 replicas
```

### Current Behavior

During **concurrent loading of multiple CARDINAL_TIERED sealed indexes**, all QueryNodes crash with a null-pointer SIGSEGV. CrashLoopBackOff with identical backtraces on every restart; the cluster cannot recover.

Crash signature (identical across pods):

```
SIGNO: 11; SIGNAME: Segmentation fault; SI_CODE: 1; SI_ADDR: (nil)
	/root/milvus/internal/core/src/segcore/storagev1translator/SealedIndexTranslator.cpp:234 pc=0x...
```

Crash path:

```
AsyncSegmentLoad (segment_c.cpp)
  → LoadFieldData
    → SealedIndexTranslator::get_cells → index->Load(ctx_, config_)
      → CARDINAL_TIERED index load (mmap + emb_list + separate_quant_storage)
      → null-pointer dereference (SI_ADDR=nil)
```

Load config at the crash point (the last log before SIGSEGV):

```json
{
  "dim": "1536",
  "emb_list_meta_file_path": "/milvus/data/cache/.../emb_list_meta",
  "emb_list_raw_index_file_path": "/milvus/data/cache/.../emb_list_raw_index",
  "enable_disk_mmap": "true",
  "enable_mmap": "true",
  "enable_mmap_i2o_map": "true",
  "enable_mmap_o2i_map": "true",
  "index_type": "CARDINAL_TIERED",
  "metric_type": "L2",
  "mmap_filepath": "/milvus/data/cache/.../index",
  "scalar_index_engine_version": "4",
  "separate_quant_storage": "true"
}
```

[...]

Issue #52723, opened 2026-08-20
[Bug]: SIGSEGV recovering growing segment - VectorMemIndex::HasRawData() reads dangling knowhere IdMap (regression from #52657)


chyezh:
## Environment

- **Milvus Version**: 3.0 branch image tag `unknown-20260820-3e402910f` (GitCommit `3e402910f`, includes backport `d859956a80` of #50524); stable baseline image `unknown-20260820-0868abf5` (without the backport)
- **Deployment Mode**: standalone
- **MQ**: rocksmq
- **SDK**: pymilvus (seed/doctor script inserting into a collection)
- **OS**: linux/arm64 (go1.26.4 linux/arm64)

### Option B: Steps (systemic / stability / timing issue)

1. Deploy: single standalone pod (1 replica).
2. Setup: collection `doctor_seed_1m` style — FloatVector dim=128, metric L2, **non-nullable** field, autoID int64 PK; default segcore config (chunkRows=128, `interimIndex.denseVectorIndexType: IVF_FLAT_CC`, `enableIndex: true`).
3. Workload: insert ~1M rows into growing segments (a growing segment reaches 175,000 rows); keep inserting during recovery (doctor-style background writes).
4. Trigger: restart the pod while a growing segment (175,000 rows) is recovered from checkpoint/WAL. On startup, segcore `AsyncSegmentLoad` loads field data (STORAGE_V2 column-group path) and builds the interim index on the recovered growing segment.
5. Observed via: pod `subprocess exit, signal: segmentation fault`; CrashLoopBackOff with 9 identical crashes (backoff 40s/80s/63s/87s/95s/165s/310s/320s), no self-heal.

### Trigger Conditions

- **Frequency**: always (9/9 identical backtraces) after upgrade; deterministic per restart.
- **First observed after**: immediately at first startup after the image upgrade (upgrade 09:48:53 → first crash 09:49:30).
- **Does NOT happen when**: old image `0868abf5` (12h of the same workload, zero crashes); happens only when recovering the growing segment with rows >= build threshold while queries are in flight.

## Expected Behavior

Standalone should recover growing segments from checkpoints and build the interim index without crashing.

## Actual Behavior

SIGSEGV (SI_CODE: 1 = MAPERR, SI_ADDR: 0xd828) in `VectorMemIndex::HasRawData()`; process crashes ~30 ms **after** the interim index build **succeeded**; every restart deterministically reproduces it → CrashLoopBackOff.

## Error Logs

```
[...]

PR #52754, opened 2026-08-21
fix: update Knowhere for stable IndexNode ABI

marcelo-cjl:
issue: #52723
issue: #52724
issue: #52725

## What

- Update Knowhere from `d85f7080` to `d7cfd888`.
- Pick up zilliztech/knowhere#1786, which keeps `IndexNode::BuildAsync()` in the public vtable for both Cardinal and non-Cardinal builds.
- Pick up the Cardinal v1 bump to `v2.5.111`, including its nullable-index fix.

## Why

In a Cardinal-enabled Milvus build, Knowhere translation units define `KNOWHERE_WITH_CARDINAL`, while Milvus core consumers of the same public header do not. The previous conditional `BuildAsync()` declaration therefore gave the two DSOs different `IndexNode` vtable layouts.

Calls intended for `GetIdMap()` could dispatch to `Count()` instead and interpret its integer return as an `IdMap&`, causing the SIGSEGVs reported in #52723, #52724, and #52725.

Knowhere `d7cfd888` makes the public vtable independent of that feature macro.

## Validation

- No new local build or test was run for this dependency-pin-only change; validation is delegated to Milvus PR CI.
- The underlying Knowhere fix passed Knowhere CI and a prior Milvus Cardinal A/B reproduction: the affected ordinary HNSW test changed from SIGSEGV/exit 139 on the old pin to 1/1 passed with the fix.

### DIFF
--- a/internal/core/thirdparty/knowhere/CMakeLists.txt
+++ b/internal/core/thirdparty/knowhere/CMakeLists.txt
@@ -14,7 +14,7 @@
 # Update KNOWHERE_VERSION for the first occurrence
 milvus_add_pkg_config("knowhere")
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES "")
-set( KNOWHERE_VERSION d85f7080 )
+set( KNOWHERE_VERSION d7cfd888 )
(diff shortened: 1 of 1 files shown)


liliu-z (approved):
/lgtm
/approve

PR #52757, opened 2026-08-21
feat(datacoord): add generic bounded retry and failure surfacing for compaction tasks

### LINKED ISSUE #52708: [Enhancement]: DataCoord compaction lacks generic bounded retry and failure surfacing

A generic bounded-retry mechanism and proper failure surfacing for DataCoord compaction tasks.

### Current state (verified on master)

**Retry:**
- The proto already has a generic `CompactionTask.retry_times` field and a `setRetryTimes` task opt, but there is no generic retry engine.
- Only clustering implements bounded retry (`merr.IsRetryableErr(err) && RetryTimes < maxRetryTimes(3)`, reset on success). `mixCompactionTask.processFailed()` and `bumpSchemaVersionTask.processFailed()` are bare `return true`.
- Nothing checks keep-failing at the policy layer: after a task dies, triggers re-select the same segments whenever their trigger condition holds. For schema-bump the condition (`SchemaVersion` lags) is deterministically true, so a segment that deterministically fails an integrity gate (e.g. row-count mismatch) re-runs the full read/write/manifest-commit cycle every trigger interval, forever — and each round can leave another unreachable orphan file for live V3 segments (GC skips files of registered V3 segments).

**Surfacing:**
- `fail_reason` is persisted on the task meta, but failed tasks are cleaned and GC'ed, so the record does not survive.
- Logs: only a WARN per failure (compaction_inspector.go:654).
- Metrics: `DataCoordCompactionTaskNum` has only `pending/executing/done` labels; failed and timeout tasks are counted into `done` (inspector increments `metrics.Done` for every finished task regardless of state), so failures are invisible on dashboards.
- RPC: `GetCompactionState.FailedPlanNo` only works for manual compaction trigger IDs; background compactions (bump/sort/clustering auto-triggers) have no consumer.

### Why is this needed?

A deterministic failure (metadata drift, corrupt segment) currently produces an unbounded silent retry loop with resource churn and orphan accumulation, and operators have no metric or API to notice it. Surfaced during review of #52484 (schema-bump integrity gates).

### Proposal sketch

1. Generalize the clustering retry pattern into the shared task framework (reuse `retry_times`; classify retryable vs terminal via merr).
2. Policy-level protection: cooldown / max re-trigger count for segments whose compaction repeatedly fails (all task types, not only bump).
3. Metrics: add a `failed` (and `timeout`) state label to `DataCoordCompactionTaskNum`; stop counting failures into `done`.
4. Keep a bounded history of terminal failures (segment ID, type, fail_reason) queryable for diagnosis.

### Anything else?

Related review thread: https://github.com/milvus-io/milvus/pull/52484#discussion_r3818434943

AyushKashyapII:
### Related Issue
Fixes #52708

---

### What this PR does / why we need it

DataCoord compaction has four task types (`mixCompactionTask`, `bumpSchemaVersionTask`, `l0CompactionTask`, `clusteringCompactionTask`). Previously, only `clusteringCompactionTask` implemented bounded retries (`merr.IsRetryableErr(err) && RetryTimes < 3`). The other three task types immediately transitioned to `failed` state on the first error, treating transient glitches (e.g. temporary network blips or busy DataNodes) the exact same as permanent failures.

Additionally, failure visibility was limited:
1. `fail_reason` on a task was erased when `cleanCompactionTaskMeta` deleted the task record after `CompactionDropToleranceInSeconds`.
2. `metrics.DataCoordCompactionTaskNum` incremented the `done` label for **every** finished task regardless of whether it succeeded, timed out, or failed.
3. `GetCompactionState`'s failure counters were scoped to manual `triggerID`s, making background compaction failures invisible to external observability tools.

This PR introduces shared bounded retry logic across `mix`, `bump_schema_version`, and `l0` compaction tasks, improves Prometheus metric accuracy, and adds long-lived failure history exposed via the `GetMetrics` endpoint.

---

#### 1. Bounded Retry Classification (`compaction_task_retry.go`)
- Added `classifyFailure()` helper function:
  - Checks if an error is retryable via `merr.IsRetryableErr(err)` and whether `RetryTimes < maxRetryTimes` (default 3).
  - If retryable, increments `RetryTimes` and preserves the task state so it is re-attempted on the next scheduler tick.
  - If non-retryable or retries are exhausted, transitions the task state to `failed` and records the `fail_reason`.
- Wired `classifyFailure()` into failure paths across `compaction_task_mix.go`, `compaction_task_bump_schema_version.go`, and `compaction_task_l0.go`.

#### 2. Configuration (`component_param.go`)
- Introduced `dataCoord.compaction.maxRetryTimes` (`DataCoordCfg.CompactionMaxRetryTimes`, default `3`, refreshable) to control the retry budget across compaction tasks.

#### 3. Metrics (`metrics.go` & `compaction_inspector.go`)
- Added `Failed` and `Timeout` status labels to `metrics.DataCoordCompactionTaskNum`.
- `checkCompaction()` now inspects the terminal state of finished tasks to accurately count `done`, `failed`, and `timeout` metrics rather than folding all completed tasks into `done`.

#### 4. Diagnostic Failure History (`compaction_task_meta.go` & `server.go`)
- Added a 24-hour LRU cache (`failureHistory`, 256 entries) in `compactionTaskMeta` that records terminal task failures.
- Failure reasons survive even after `cleanCompactionTaskMeta` deletes the primary task record.
- Exposed failure history via the `GetMetrics` endpoint under the key `compaction_task_failures` (`metricsinfo.CompactionTaskFailureKey`).

---

### Out of Scope / Deferred to Follow-ups
- **Cross-Task Cooldown:** Segment-level cooldown across separate trigger cycles (which allocate fresh `PlanID`s and task structs) is explicitly deferred for a separate design discussion.
- **Clustering Task Refactoring:** `clusteringCompactionTask`'s existing retry implementation remains untouched to limit blast radius.

---

### Test Plan
- `compaction_task_retry_test.go`: Isolated unit test suite for `classifyFailure()` (retryable vs non-retryable errors, max retry limits, parameter overrides).
- `compaction_task_mix_test.go`, `compaction_task_bump_schema_version_test.go`, `compaction_task_l0_test.go`: Verified retry increments and failure transitions end-to-end through real task execution paths.
- `compaction_task_meta_test.go`: Verified `TestFailureHistorySurvivesTaskMetaDrop` to confirm failure reasons outlive task deletion.
- `compaction_inspector_test.go`: Updated metric assertion tests to ensure `done`, `failed`, and `timeout` labels are properly recorded.

---

### Does this PR introduce breaking changes?
- [ ] Yes
- [x] No

> **Note on Metrics:** The Prometheus metric `datacoord_compaction_task_num` has no change to its metric name or label keys. However, the `done` status label will now accurately exclude failed/timed-out tasks.

### DIFF
--- a/internal/datacoord/compaction_inspector.go
+++ b/internal/datacoord/compaction_inspector.go
@@ -643,18 +643,34 @@ func (c *compactionInspector) checkCompaction() error {
 	c.executingGuard.Lock()
 	for _, t := range finishedTasks {
 		delete(c.executingTasks, t.GetTaskProto().GetPlanID())
-		mlog.Info(context.TODO(), "compaction task finished",
+		state := t.GetTaskProto().GetState()
+		terminal := state == datapb.CompactionTaskState_failed || state == datapb.CompactionTaskState_timeout
+		logFields := []mlog.Field{
 			mlog.Int64("planID", t.GetTaskProto().GetPlanID()),
 			mlog.String("type", t.GetTaskProto().GetType().String()),
-			mlog.String("state", t.GetTaskProto().GetState().String()),
+			mlog.String("state", state.String()),
 			mlog.String("channel", t.GetTaskProto().GetChannel()),
 			mlog.String("label", t.GetLabel()),
 			mlog.FieldNodeID(t.GetTaskProto().GetNodeID()),
 			mlog.Int64s("inputSegments", t.GetTaskProto().GetInputSegments()),
 			mlog.String("reason", t.GetTaskProto().GetFailReason()),
-		)
+		}
+		if terminal {
+			mlog.Warn(context.TODO(), "compaction task finished with a failure or timeout", logFields...)
+			c.meta.GetCompactionTaskMeta().RecordTerminalFailure(t.GetTaskProto())
+		} else {
+			mlog.Info(context.TODO(), "compaction task finished", logFields...)
+		}
+
 		metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Executing).Dec()
-		metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Done).Inc()
+		switch state {
+		case datapb.CompactionTaskState_failed:
+			metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Failed).Inc()
+		case datapb.CompactionTaskState_timeout:
+			metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Timeout).Inc()
+		default:
+			metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Done).Inc()
+		}
 	}
 	c.executingGuard.Unlock()
 
--- a/internal/datacoord/compaction_task_meta.go
+++ b/internal/datacoord/compaction_task_meta.go
@@ -56,6 +56,16 @@ func newCompactionTaskStats(task *datapb.CompactionTask) *metricsinfo.Compaction
 	}
 }
 
+// terminalFailureHistorySize bounds the number of terminally-failed/timed-out compaction
+// tasks kept for diagnosis after their task meta has been dropped by cleanCompactionTaskMeta.
+const terminalFailureHistorySize = 256
+
+// terminalFailureHistoryTTL is deliberately longer than taskStats' TTL: taskStats is a
+// live debug snapshot of recent tasks of any outcome, while this history exists specifically
+// so a terminal failure remains queryable well after cleanCompactionTaskMeta has deleted the
+// task's own meta record (see compaction_inspector.go's CompactionDropToleranceInSeconds).
+const terminalFailureHistoryTTL = 24 * time.Hour
+
 type compactionTaskMeta struct {
 	sync.RWMutex
 	ctx     context.Context
@@ -63,6 +73,9 @@ type compactionTaskMeta struct {
 	// currently only clustering compaction task is stored in persist meta
 	compactionTasks map[int64]map[int64]*datapb.CompactionTask // triggerID -> planID
 	taskStats       *expirable.LRU[UniqueID, *metricsinfo.CompactionTask]
+	// failureHistory retains terminally-failed/timed-out tasks (keyed by planID) past the
+	// point cleanCompactionTaskMeta deletes their normal task meta, for diagnosis.
+	failureHistory *expirable.LRU[UniqueID, *metricsinfo.CompactionTask]
 }
 
 func newCompactionTaskMeta(ctx context.Context, catalog metastore.DataCoordCatalog) (*compactionTaskMeta, error) {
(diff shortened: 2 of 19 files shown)


internal/datacoord/compaction_task_meta.go
```diff
@@ -205,3 +219,23 @@ func (csm *compactionTaskMeta) TaskStatsJSON() string {
...
 	return string(ret)
 }
+
+// RecordTerminalFailure snapshots a task that just reached a terminal failed/timeout state,
+// so it stays queryable after cleanCompactionTaskMeta later deletes its regular task meta.
+// Safe to call for any task; callers are expected to only call it for failed/timeout tasks.
+func (csm *compactionTaskMeta) RecordTerminalFailure(task *datapb.CompactionTask) {
+	csm.Lock()
+	defer csm.Unlock()
+	csm.failureHistory.Add(task.GetPlanID(), newCompactionTaskStats(task))
```
yhmo, line 243: internal/datacoord/compaction_task_meta.go line:229
Medium ---- Failure-history entries always record an empty end_time, which undercuts the diagnosis purpose of the new compaction_task_failures output.

RecordTerminalFailure snapshots the task at the moment checkCompaction detects it, but every terminal transition that reaches this point (classifyFailure's setState(failed), or the setState(timeout) paths in QueryTaskOnWorker) calls updateAndSaveTaskMeta while the task's current state is still executing. The end-time-append branch in updateAndSaveTaskMeta only fires when the pre-transition state is already completed/cleaned/failed/timeout, so task.EndTime stays 0 and newCompactionTaskStats renders TimestampToString(0) as "", which the omitempty json tag then drops entirely. The regression test TestFailureHistorySurvivesTaskMetaDrop cannot catch this because it constructs the task with an explicit EndTime.

A straightforward fix is to stamp the end time when recording (e.g. clone with setEndTime(time.Now().Unix()) before newCompactionTaskStats) so the history actually shows when the terminal failure occurred.

internal/datacoord/compaction_inspector.go
```diff
@@ -643,18 +643,34 @@ func (c *compactionInspector) checkCompaction() error {
...
+			c.meta.GetCompactionTaskMeta().RecordTerminalFailure(t.GetTaskProto())
+		} else {
+			mlog.Info(context.TODO(), "compaction task finished", logFields...)
+		}
+
 		metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Executing).Dec()
-		metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Done).Inc()
+		switch state {
+		case datapb.CompactionTaskState_failed:
+			metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", t.GetTaskProto().GetNodeID()), t.GetTaskProto().GetType().String(), metrics.Failed).Inc()
```
yhmo, line 668: internal/datacoord/compaction_inspector.go line:668
Low ---- DataCoordCompactionTaskNum is a GaugeVec, but the new failed/timeout label values (like the existing done) are used as monotonic per-process counters via .Inc().

The metric Help text reads "Number of compaction tasks currently", which only fits the pending/executing series (Inc/Dec). done/failed/timeout only ever increase, so they have counter semantics while living on a gauge: they reset to zero on DataCoord restart (a Counter would expose the restart), and operator queries applying rate()/increase() or graphing "current" against the Help text can misread the series. Since this PR exists specifically to improve metric accuracy, consider either introducing a separate CounterVec for the cumulative outcome counts, or at least clarifying the metric Help/name so the cumulative semantics of done/failed/timeout are unambiguous.

AyushKashyapII:
hey @yhmo , thnks for the review! Fixed both:

- EndTime bug — you were right, RecordTerminalFailure now backfills the timestamp when it's still unset instead of silently dropping it. Added a test that actually exercises the zero-EndTime path this time instead of hand-setting it.
- Gauge/counter mismatch — went with the minimal fix for now: clarified the Help text so done/failed/timeout are honest about being cumulative-since-restart, not a live count. Happy to split them into a proper CounterVec in a follow-up if you'd rather do it now instead.

AyushKashyapII:
CI is failing at build-3rdparty libavrocpp/1.12.1.1 source download 404s from dlcdn.apache.org (Apache rotated it out for 1.12.2; still available at archive.apache.org). This looks like an infra issue in the conan recipe, unrelated to this PR's changes.

PR #52759, opened 2026-08-22
test: tolerate TEI rerank score drift in model-rerank order assertion

changshenhan:
@-


xiaofan-luan:
Closing as superseded by #52742, which has merged the rerank score-drift tolerance fix and covers both rerank test paths. Thank you for the contribution.

PR #52760, opened 2026-08-22
fix: clean proxy collection/db metrics via DeletePartialMatch not stale enumeration

changshenhan:
@-

### DIFF
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -17,6 +17,9 @@
 package metrics
 
 import (
+	"os"
+	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


pkg/metrics/proxy_metrics.go
```diff
@@ -586,192 +586,133 @@ func CleanupProxyDBMetrics(nodeID int64, dbName string) {
...
-		collectionName:    collection,
 	})
-	ProxyInsertVectors.DeletePartialMatch(prometheus.Labels{
+	ProxyCollectionMutationLatency.DeletePartialMatch(prometheus.Labels{
 		nodeIDLabelName:   strconv.FormatInt(nodeID, 10),
 		databaseLabelName: dbName,
-		collectionName:    collection,
 	})
-	ProxyUpsertVectors.DeletePartialMatch(prometheus.Labels{
+	ProxyReceivedNQ.DeletePartialMatch(prometheus.Labels{
```
yhmo, line 601: pkg/metrics/proxy_metrics.go line:601
Medium ---- This same conversion to DeletePartialMatch already landed on master via #52691 (commit e198153655b2, merged 2026-08-25), and the merged version is a superset of this PR: it also adds a db_name label to ProxySearchSparseNumNonZeros and ProxyFunctionlatency, cleans ProxyReportValue on DropDatabase, and dedups the metric list into a single proxyCollectionScopedMetrics() helper shared by both cleanup functions. This PR's base (57ea5fff, 2026-08-22) predates that merge, so merging now would conflict with current master, and master's version is more complete than what is proposed here. Please rebase onto current master and reconcile, or close this PR in favor of #52691.

pkg/metrics/proxy_metrics.go
```diff
@@ -586,192 +586,133 @@ func CleanupProxyDBMetrics(nodeID int64, dbName string) {
...
+	ProxyScannedRemoteMB.DeletePartialMatch(prometheus.Labels{
+		nodeIDLabelName:   strconv.FormatInt(nodeID, 10),
+		databaseLabelName: dbName,
 	})
-	ProxyCollectionSQLatency.Delete(prometheus.Labels{
-		nodeIDLabelName:    strconv.FormatInt(nodeID, 10),
-		queryTypeLabelName: UpsertQueryLabel,
-		databaseLabelName:  dbName,
-		collectionName:     collection,
+	ProxyScannedTotalMB.DeletePartialMatch(prometheus.Labels{
```
yhmo, line 625: pkg/metrics/proxy_metrics.go line:625
Medium ---- DropDatabase still leaks the collection-scoped metrics that carry no db_name label: ProxySearchSparseNumNonZeros (node_id, collection_name, query_type, field_id) and ProxyFunctionlatency (node_id, collection_name, function_type, provider, function_name) are not cleaned by CleanupProxyDBMetrics, and the db-only ProxyReportValue is not cleaned either. The new structural test explicitly documents this exclusion, so the leak becomes permanent rather than accidental. Separately, cleaning those two metrics by (node_id, collection_name) on collection drop can remove a same-named collection's series in a different database, because collection names are only unique per db. Issue #52690 calls out the missing db_name label explicitly, and merged #52691 fixed both problems by adding db_name to the two metrics; consider adopting the same approach here.

pkg/metrics/metrics_test.go
```diff
@@ -367,3 +385,80 @@ func TestCleanupProxyCollectionMetricsDropsEveryCollectionSeries(t *testing.T) {
...
+	src, err := os.ReadFile("proxy_metrics.go")
+	assert.NoError(t, err)
+	text := string(src)
+
+	collectionBody := funcBody(text, "CleanupProxyCollectionMetrics")
+	dbBody := funcBody(text, "CleanupProxyDBMetrics")
+	assert.NotEmpty(t, collectionBody, "CleanupProxyCollectionMetrics not found in proxy_metrics.go")
+	assert.NotEmpty(t, dbBody, "CleanupProxyDBMetrics not found in proxy_metrics.go")
+
+	vecRe := regexp.MustCompile(`(?s)(\w+)\s*=\s*prometheus\.New\w+Vec\(.*?\[\]string\{([^}]*)\}`)
```
yhmo, line 453: pkg/metrics/metrics_test.go line:453
Low ---- This structural guard has silent false-negative paths: the regex only matches declarations shaped `X = prometheus.New*Vec(...)`, so a future metric declared with `:=`, renamed, or with labels formatted differently silently bypasses the check; and assert.Contains is a substring match, so a comment inside the cleanup function that merely mentions a metric name would satisfy it without the metric actually being deleted. Consider asserting the exact DeletePartialMatch call sites instead of substring presence.

pkg/metrics/metrics_test.go
```diff
@@ -283,25 +286,40 @@ func proxyCollectionCollectors() []prometheus.Collector {
 	}
 }
 
+// observeProxyCollection emits the label combinations the proxy actually uses
+// (internal/proxy/impl.go), including the ones the pre-#52690 enumeration did
+// not cover. The deprecated per-collection metrics are emitted for search,
+// hybrid_search, query and upsert_query / delete and upsert; the byte/scanned
+// counters for every msg_type including hybrid_search. Observing a single
+// combination per metric cannot catch a stale cleanup enumeration, so this
+// deliberately mirrors the full emitted space.
```
yhmo, line 295: pkg/metrics/metrics_test.go line:295
Low ---- "deliberately mirrors the full emitted space" overstates what observeProxyCollection does: it only observes representative combos (ProxyCollectionSQLatency gets Search + HybridSearch, while production also emits query and upsert_query; ProxyReceiveBytes gets Delete + HybridSearch, while production also emits Search/Query/Insert/Upsert). That is sufficient to catch the stale-enumeration bug, so coverage is fine, but the comment should describe it as a representative subset rather than the full emitted space. Also fixes the typo "Observaing".

changshenhan:
Closing this PR as fully superseded by #52691 (merged), which implements the same fix for #52690:

- `DeletePartialMatch` on proxy collection/db metrics instead of enumerating label values (which went stale for `hybrid_search`/`upsert`)
- DB-drop cleanup for collection-scoped metrics
- A structural completeness guard (`TestCollectionScopedMetricsAreComplete`) covering both cleanup paths

This PR's changes and its regression tests are all subsumed by #52691, so there is nothing left to merge here. Thanks for the review cycle — happy to pick up other areas next.

PR #52761, opened 2026-08-22
enhance: optimize computeRecall using hash set instead of nested loop

chasingegg:
## What this PR does

`recallCal` in `internal/proxy/util.go` uses a nested loop to compute recall, which is O(n*m). When topk is large (e.g. 100000), this results in 10^10 comparisons and takes several seconds per query.

This PR replaces the nested loop with a hash set, reducing the complexity to O(n+m).

## Benchmark (topk = 100000)

| Implementation | Complexity | Time |
| --- | --- | --- |
| nested loop (before) | O(n*m) = 10^10 | ~3.08s |
| hash set (after) | O(n+m) | ~18.5ms |

~166x faster.

issue: #52764
pr: #52762 

Signed-off-by: chasingegg

### DIFF
--- a/internal/proxy/util.go
+++ b/internal/proxy/util.go
@@ -1572,18 +1572,22 @@ func translatePkOutputFields(schema *schemapb.CollectionSchema) ([]string, []int
 }
 
 func recallCal[T string | int64](results []T, gts []T) float32 {
+	if len(results) == 0 {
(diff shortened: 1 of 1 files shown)


liliu-z (commented):
/lgtm
/approve

PR #52765, opened 2026-08-22
test: reduce e2e setup cost via shared collections (T2 geometry/minhash + reranker dedup)

yanliang567:
## Goal

pytest E2E runtime is dominated by setup (create + insert + index); search/query is negligible. This PR adopts the "one setup, many verification points" pattern (shared collections / wide tables) to cut redundant setup **without reducing any coverage** (all parametrized cases and assertions are preserved, no values trimmed).

### 1. Geometry shared collections (pilot, measured −20% wall-clock)
- `TestMilvusClientGeometryInsertShared`: 14 insert-family cases share 1 collection (ids partitioned per WKT type), collapsing 14 setups → 1.
- `TestMilvusClientGeometrySpatialShared`: 14 spatial-operator cases share 1 dataset provisioned in 2 collections (RTREE / no geo index), collapsing 14 setups → 2.
- Template: `@pytest.mark.xdist_group` + module-scoped fixture + `force_teardown=False` + `request.addfinalizer`.
- Measured against a test cluster: 123 passed / 3 skipped unchanged, **137.2s → 110.4s**.

### 2. search.py stale reranker dedup
- `test_milvus_client_search_with_reranker` (288 combos, in-body setup) is a duplicate of `search_reranker.py::TestMilvusClientSearchDecayRerankShared` (identical schema/parametrization/assertions). Removed the stale copy (−288 setups/PR); coverage kept by the shared version.

### 3. minhash default-schema shared collection
- `TestMilvusClientMinHashDefaultShared`: 3 read tests (search_basic / search_with_filter / query_by_id) share 1 collection. Results identical to baseline (the file was already fast, so the gain is marginal).

### 4. Review fixes
- Replaced module-level globals with class attributes; cached client across the class; added clarifying comments (RTREE-on-empty, geo-index semantic change); extra dedup of `test_milvus_client_search_with_reranker_default_offset_decay` (shared version is a strict superset). All commits carry `Signed-off-by`.

## Verification
- geometry + minhash run against a test cluster: 196 passed; the 2 `TestMinHashBulkImport` failures are pre-existing environment issues (also fail on master, need MinIO config).
- `ruff check` / `ruff format` pass.
- No parametrized values changed — zero coverage loss.

## Background
First batch of the e2e runtime optimization (T2: one-setup-many-queries). Follow-ups planned per schema clustering (struct_array_element_search, external_table, search.py remaining, restful tag audit).


yanliang567 (commented):
## Review

Solid optimization that correctly reuses the shared-collection template (`xdist_group` + module-scoped fixture + `force_teardown=False` + `request.addfinalizer`) already proven in `test_milvus_client_search_reranker.py`. Coverage is preserved and the reranker dedup is valid.

### Verified correct
- **Reranker dedup**: `TestMilvusClientSearchDecayRerankShared.test_milvus_client_search_with_reranker` (`test_milvus_client_search_reranker.py:2376`) has identical parametrization (3×3×4×4×1×2 = 288), schema, seed, and the same three search checks. Safe to drop the `test_milvus_client_search.py` copy.
- **RTREE over all WKT types** (incl. `GEOMETRYCOLLECTION`) is already exercised by `test_build_geometry_index`, so building RTREE in the shared insert collection is fine.
- **id partitioning**: `t_idx*3000` ranges and the `1_000_000` empty-geometry offset don't overlap; `generate_wkt_by_type` returns exactly `count` rows, so the `id >= lo and id < hi` count assertions hold.
- **MinHash `search_basic`**: shared rows (`pk=i, text=texts[i]`) are equivalent to `gen_minhash_rows`; adding `category` doesn't affect MHJACCARD ranking, so `results[0][0]["id"] == 0` still holds.

### Concerns
1. **(medium)** Empty geometries now coexist with 21k non-empty rows in one RTREE-indexed collection. The original `test_insert_valid_empty_geometry` built RTREE on a collection containing only the single empty geometry. If the server treats empty geometries specially during RTREE build this could become a latent flake. It passed in your run, but please add a comment in the fixture explaining that RTREE-on-empty is covered, or run a stability check on this class.
2. **(low)** `test_insert_wkt_data` originally had no geo index; the shared fixture now adds `geo RTREE` to that data. Results are unaffected (id-range query), but this is a small semantic change from the original intent.
3. **(low)** Module-level mutable globals `_GEOM_SPATIAL_BASE_DATA` / `_MINHASH_SHARED_ROWS` via `global` are fragile (they rely on per-worker process isolation under xdist). Prefer yielding the data from the fixture or a class attribute.
4. **(low)** `_client()` is called repeatedly (fixture + each test), creating a fresh client each time. Minor waste, not a correctness issue.
5. **(nit, pre-existing)** `test_milvus_client_search.py:4856` `test_milvus_client_search_with_reranker_default_offset_decay` is still duplicated by `search_reranker.py:2448` — same dedup batch, candidate follow-up.

yanliang567:
Thanks for the thorough review! All concerns addressed in the new commit (`6a0a345d4a`, after the sign-off rebase):

**#1 (medium, RTREE-on-empty):** Added a NOTE comment in the geometry insert fixture documenting that empty geometries now coexist with 21k non-empty rows in the same RTREE-indexed collection, and that RTREE-over-every-WKT-type (incl. GEOMETRYCOLLECTION) is additionally covered by `test_build_geometry_index`. I'll also run a stability pass on this class.

**#2 (low, geo RTREE semantic change):** Documented in the same fixture comment.

**#3 (low, module globals):** Replaced both module-level globals with class attributes (`cls.base_data` / `cls.rows`) set in the module fixture, read via `type(self).*`. No more reliance on `global` or per-worker process isolation.

**#4 (low, repeated `_client()`):** The module fixture now caches the client as `cls.shared_client` and the shared-class tests reuse it.

**#5 (nit, pre-existing duplicate):** Removed `test_milvus_client_search_with_reranker_default_offset_decay` from search.py — the shared version in search_reranker.py is a strict superset (same two searches plus a third `reranker_field>=0 and <=10` filter search), so coverage is preserved.

Also fixed DCO: all commits now carry `Signed-off-by`.

Verification: reran geometry + minhash against the cluster after the refactor — 196 passed (the 2 bulk-import failures are pre-existing environment issues, present on master too). ruff check/format pass.

yanliang567 (commented):
## Follow-up review

All five points from the previous round are addressed cleanly. Verified:

1. **Module globals → class attributes** — `_GEOM_SPATIAL_BASE_DATA` / `_MINHASH_SHARED_ROWS` replaced by `cls.base_data` / `cls.rows`. This is correct and stays within the owning worker process under xdist (the class is pinned to one worker via `xdist_group`, so the module-scoped fixture runs once and the attribute is visible to every test in the class). No cross-class leakage.
2. **Client caching** — `cls.shared_client` set once in the fixture and reused via `type(self).shared_client`. Fine; removes the redundant `_client()` reconnect per test. (Minor: this connection is never explicitly closed, but MilvusClient over HTTP is stateless, so no practical impact.)
3. **RTREE-on-empty comment** — added, correctly notes the empty geometries now coexist with 21k rows and points to the stability caveat.
4. **geo-index semantic change comment** — added for `test_insert_wkt_data`.
5. **Extra dedup** — `test_milvus_client_search_with_reranker_default_offset_decay` removed. I confirmed `search_reranker.py:2448` is a strict superset: same `["gauss","linear","exp"]` parametrization, same `origin=0, scale=100` reranker, same three search checks (no filter / `>=9 && <=4` limit 0 / `>=0 && <=10`). Safe.

Local checks: `ruff check` and `ruff format --check` both pass on the three changed files.

No blocking issues. LGTM.

tests/python_client/milvus_client/test_milvus_client_geometry.py
```diff
@@ -3261,6 +3064,256 @@ def test_st_isvalid_basic(self, with_geo_index):
...
+                "geo": empty_wkt,
+            }
+            for e_idx, empty_wkt in enumerate(GEOM_EMPTY_WKT_TYPES)
+        ]
+        self.insert(client, collection_name, empty_rows)
+        self.flush(client, collection_name)
+
+        index_params, _ = self.prepare_index_params(client)
+        index_params.add_index(field_name="vector", index_type="IVF_FLAT", metric_type="L2", nlist=128)
+        index_params.add_index(field_name="geo", index_type="RTREE")
```
zhuwenxing: **[P1] Preserve the no-geo-index coverage for WKT cases**

The original parametrized test created only a vector index, so it covered loading and querying all seven WKT types while the geometry field had no index. This shared fixture adds an RTREE index to the only collection, meaning these cases now exercise only the geo-indexed path. The no-index spatial fixture contains only POINT, POLYGON, and LINESTRING data, so it does not restore the MULTI* and GEOMETRYCOLLECTION coverage.

This conflicts with the PR's “without reducing coverage” goal. Could we keep the WKT shared collection without RTREE and use a separate shared collection for the empty-geometry RTREE coverage?

tests/python_client/milvus_client/test_milvus_client_search.py
```diff
@@ -4851,143 +4851,6 @@ def scalar_index(self, request):
...
-            params={
-                "reranker": "decay",
-                "function": function,
-                "origin": origin,
-                "offset": offset,
-                "decay": decay,
-                "scale": scale
-            }
-        )
-        vectors_to_search = rng.random((1, dim))
```
zhuwenxing, line 4900: **[P1] Preserve the original query vector when deduplicating this test**

The deleted test generates its rows and query from the same RNG instance. After generating 3,000 five-dimensional row vectors, the query uses the next five random values.

The replacement shared test reinitializes the RNG with the same seed immediately before generating the query, so its query is exactly `rows[0]["vector"]`. This changes the ordinary and range-rerank inputs into a self-match, even though the parametrization and assertions are unchanged.

Could the shared fixture cache `rng.random((1, DECAY_RERANK_SHARED_DIM))` after generating the rows and have both shared tests reuse that value?

yanliang567:
Both P1 comments addressed in `53a092e8d8` (Signed-off-by included):

**1. geometry: preserve no-geo-index coverage for WKT cases**

The insert shared class now provisions **two** collections with identical data:
- `GEOM_INSERT_SHARED_COLLECTION` — vector IVF_FLAT + **geo RTREE**
- `GEOM_INSERT_SHARED_COLLECTION_NOINDEX` — vector IVF_FLAT only (no geo index)

`test_insert_wkt_data` now queries the **no-index** collection (matching the original test, which built only a vector index), so all 7 WKT types keep their no-geo-index load/query coverage. `test_insert_valid_empty_geometry` keeps the RTREE collection (the original built RTREE there). Verified on a test cluster: the shared insert class passes 14/14.

**2. search: preserve the original query vector in default_offset_decay**

The original test generated `default_nb` row vectors from a seed-19530 RNG and then took the query as the `(default_nb+1)`-th chunk. The shared replacement re-seeded and took the first chunk (`rows[0]`'s vector). Fixed by advancing the RNG through `default_nb` rows before drawing the query, so the shared test now uses the exact original query vector.

Verification: geometry + search_reranker run 504 passed / 3 failed, where the 3 failures are in the untouched `TestMilvusClientSearchModelRerank` class (TEI model endpoint unreachable on this host — `socket.gaierror`, pre-existing env dependency).

Issue #52769, opened 2026-08-22
[Feature]: support ExportSnapshot/RestoreExternalSnapshot across Azure storage accounts via source SAS


huanghaoyuanhhy:
## Is there an existing issue for this?

I have searched the existing issues ("ExportSnapshot azure", "snapshot cross-bucket", "cross account azure copy") and found none.

## Is your feature request related to a problem? Please describe.

`ExportSnapshot` / `RestoreExternalSnapshot` cannot cross Azure storage accounts. On Azure, the instance bucket and the backup bucket very commonly live in different storage accounts (e.g. a dedicated backup/DR account), and every such request is rejected up front:

```
error_code:IllegalArgument reason:"server-side cross-bucket copy is unsupported for this provider/endpoint pair; cross-service copy would require streaming (unsupported)" code:1100
```

This comes from `validateProviderEndpointPair` in `internal/snapshotio/storage/resolver.go`, where the `providerFamilyAzure` branch requires `sameAzureAccountEndpoint` (instance and foreign endpoints resolve to the same `<account>.blob.core.windows.net`) and falls through to the rejection otherwise.

I understand this is a deliberate design choice, not an oversight — the MEP (`docs/design-docs/design_docs/20260609-external-snapshot-export-restore.md`) lists as non-goals: *"No cross-provider copy, cross-endpoint copy, or provider-specific source-auth extension"* and, for Azure, *"a request SAS token is not supported"*.

However, Azure Blob Storage itself fully supports cross-account server-side copy — the gap is only in the credential model. Per the [Copy Blob From URL / Copy Blob REST docs](https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url), for a source blob in another storage account:

- Shared Key authorization: **No**
- the request's own Microsoft Entra ID (OAuth) authorization: **No**
- a SAS token with Read (`r`) permission on the source URL: **Yes**

The current implementation in `internal/storage/azure_object_storage.go` (`CopyObjectCrossBucket`) issues `StartCopyFromURL` with a bare source URL and a single shared-key credential, so a cross-account source read can never be authorized — hence the fail-closed validation. But if the source URL carried a read-scoped SAS, the very same async `Copy Blob` request would work cross-account, still as one provider-side copy with no streaming through Milvus.

[...]

PR #52770, opened 2026-08-22
feat: support cross-account Azure snapshot copy via source SAS

huanghaoyuanhhy:
issue: #52769
design doc: `docs/design-docs/design_docs/20260609-external-snapshot-export-restore.md` (new §10)

## What

Adds a scoped exception to the snapshot API's "no source-auth" rule: Azure cross-storage-account server-side copy, authorized on the read side by a read-scoped SAS token carried on the copy source URL.

`external_spec.extfs` gains one Azure-only key:

```json
{"extfs": {
  "cloud_provider": "azure",
  "access_key_id": "backup-account", "access_key_value": "...",
  "source_sas_token": "sv=2024-08-04&sig=...&sp=r"
}}
```

## Why

Azure's copy authorization rules ([Copy Blob From URL], Authorization table) allow a source blob in **another** storage account to be read only via a read-scoped SAS on the source URL — neither account's Shared Key nor the request's own Entra ID token covers it. The snapshot export/restore copy is a single `StartCopyFromURL` authorized by the destination account, so until now `validateProviderEndpointPair` failed closed on any Azure account crossing, and a cross-account copy could not be expressed at all. With the SAS the copy is still one provider-side request — it satisfies the design's core invariant from §6 of the MEP — so this is a credential-model extension, not a streaming fallback.

Concrete impact: milvus-backup's snapshot-format backup targets a dedicated backup storage account in the common Azure DR topology and is currently rejected up front (`IllegalArgument ... server-side cross-bucket copy is unsupported for this provider/endpoint pair`).

## How

- **Validation** (`internal/snapshotio/storage/validator.go`): `source_sas_token` is a snapshot-only extfs key (strip/re-inject like `credential_json`), valid only for `cloud_provider=azure`, non-empty, and **not** a credential mode — the destination side still uses exactly one existing mode. Value is redacted in `secretExtfsKeys` alongside `credential_json`.
- **Endpoint pair check** (`resolver.go`): with a SAS, the Azure family allows different account endpoints within one sovereign cloud. Same-account requests carrying a SAS and cross-cloud requests are rejected as input errors — an unused or meaningless SAS signals misconfiguration.
- **Copy execution** (`internal/storage/azure_object_storage.go`): a second, credential-less service client addresses the source account; cross-account source URLs are built from it with the SAS appended. Copy-source verification during polling now compares URL identities without the query string, since the service does not guarantee `x-ms-copy-source` echoes the SAS.
- **Copier selection** (`resolver.go`): export keeps the foreign-account copier (destination = foreign account) with the instance endpoint + SAS as the source; restore/copy-source switch to an instance-credential copier (destination = instance bucket) with the foreign endpoint + SAS as the source. Metadata reads/writes keep their existing clients.
- No proto or C++ changes: the SAS rides inside the opaque `external_spec` JSON that already propagates Proxy → DataCoord → WAL/job/task state → DataNode.

## Failure modes traced (per repo verification gate)

- SAS expired / lacking read on the source container → first `StartCopyFromURL` returns 403 → export job fails before `Publishing` through the existing permission-failure path; no metadata published.
- SAS echoed with or without query by the service during polling → source verification compares credential-free URL identities, covered by unit test.
- Same-account + SAS, cross-cloud + SAS, non-Azure provider + SAS, empty SAS → rejected as input errors at validation, covered by unit tests.
- Cross-account without SAS → still rejected up front with the existing `unsupportedServerSideCopyMessage` (unchanged behavior).

## Tests

- `internal/snapshotio/storage`: resolver accepts cross-account export/restore with a SAS (asserting both resolved client configs), rejects cross-account without SAS, same-account with SAS, and cross-cloud with SAS; validator applies/trims the token, enforces azure-only and non-empty, and allows it alongside raw credentials.
- `internal/storage`: `CopyObjectCrossBucket` builds the cross-account source URL with the SAS, keeps the same-account URL unchanged, and rejects an invalid source host at construction; poll verification ignores the query.
- `pkg/util/externalspec`: `source_sas_token` is redacted.

## Follow-ups

- milvus-backup: mint/send the source SAS in its `ExternalSpec` for cross-account Azure snapshot backups (milvus-backup repo, separate PR).
- A real-account e2e against two storage accounts is not covered by CI; the change is verified at the unit/contract level here.

[Copy Blob From URL]: https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url

### DIFF
--- a/internal/snapshotio/storage/resolver.go
+++ b/internal/snapshotio/storage/resolver.go
@@ -65,7 +65,21 @@ func ResolveForeignStorage(
 	if err != nil {
 		return nil, err
 	}
-	foreignCM, err := milvusstorage.NewRemoteChunkManager(ctx, resolvedCfg.foreignCfg)
+	// The metadata client only reads and writes the foreign bucket with the
+	// foreign credential, so it must not carry the Azure source-copy fields.
+	// Besides being wrong in principle, they break construction for
+	// Restore/CopySource: AzureSourceEndpoint is populated only on the copier
+	// config below, and a source SAS without an endpoint fails client
+	// construction. Export keeps the fields because there foreignCM is the
+	// copier.
+	metadataCfg := resolvedCfg.foreignCfg
+	if direction == DirectionRestore || direction == DirectionCopySource {
+		metadataCfg = cloneObjectStorageConfig(resolvedCfg.foreignCfg)
+		metadataCfg.AzureSourceEndpoint = ""
+		metadataCfg.AzureSourceUseSSL = false
+		metadataCfg.AzureSourceSAS = ""
+	}
+	foreignCM, err := milvusstorage.NewRemoteChunkManager(ctx, metadataCfg)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/snapshotio/storage/resolver.go
+++ b/internal/snapshotio/storage/resolver.go
@@ -71,9 +71,22 @@ func ResolveForeignStorage(
 	}
 	var copier milvusstorage.CrossBucketCopier = foreignCM
 	if resolvedCfg.hasSpec && (direction == DirectionRestore || direction == DirectionCopySource) {
-		// Layer 2 restore uses foreign credentials against the target endpoint so
-		// the provider can authorize both sides of the server-side copy request.
-		copier, err = milvusstorage.NewRemoteChunkManager(ctx, restoreProviderCopyConfig(instanceCfg, resolvedCfg.foreignCfg))
+		var copyCfg *objectstorage.Config
+		if strings.TrimSpace(resolvedCfg.foreignCfg.AzureSourceSAS) != "" {
+			// Cross-account Azure copies issue the copy request from the
+			// instance side: the destination write needs the instance account's
+			// credential, and the foreign source read is authorized by the SAS
+			// carried on the source URL.
+			copyCfg, err = restoreSASCopyConfig(instanceCfg, resolvedCfg.foreignCfg)
+		} else {
+			// Layer 2 restore uses foreign credentials against the target endpoint so
+			// the provider can authorize both sides of the server-side copy request.
+			copyCfg = restoreProviderCopyConfig(instanceCfg, resolvedCfg.foreignCfg)
+		}
+		if err != nil {
+			return nil, err
+		}
+		copier, err = milvusstorage.NewRemoteChunkManager(ctx, copyCfg)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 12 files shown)


internal/storage/azure_object_storage.go
```diff
@@ -267,7 +322,10 @@ func waitAzureCopyComplete(
 			continue
 		}
 		if expectedSource != "" {
-			if props.CopySource == nil || *props.CopySource != expectedSource {
+			// The service echoes the copy source without a guarantee about the
+			// query string, so compare the URL identity only: a SAS-bearing
+			// source must still verify against its credential-free form.
+			if props.CopySource == nil || azureCopySourceIdentity(*props.CopySource) != azureCopySourceIdentity(expectedSource) {
```
weiliu1031: **[P1] Redact the SAS before returning this mismatch**

For a cross-account copy, expectedSource is produced by sourceCopyURL and contains source_sas_token in its query string. The mismatch error below returns expectedSource, and potentially props.CopySource, verbatim. That error can bubble into snapshot job failure reasons and logs; snapshotExportSecretValues also does not include source_sas_token, so the token is not removed by the snapshot-specific scrubber. Please format only credential-free URL identities here and add source_sas_token to that scrubber.

internal/storage/azure_object_storage.go
```diff
@@ -49,7 +57,43 @@ func newAzureObjectStorageWithConfig(ctx context.Context, c *objectstorage.Confi
 	if err != nil {
 		return nil, err
 	}
-	return &AzureObjectStorage{Client: client}, nil
+	storage := &AzureObjectStorage{Client: client}
+	if c.AzureSourceSAS == "" {
+		return storage, nil
+	}
+	sourceURL, err := azureServiceURL(c.AzureSourceEndpoint, c.UseSSL)
```
weiliu1031: **[P1] Use the source endpoint transport setting**

This constructs the source URL with c.UseSSL, but c describes the copy destination: restoreSASCopyConfig clones instanceCfg while AzureSourceEndpoint comes from foreignCfg, and export uses foreignCfg while the endpoint comes from instanceCfg. If the two sides differ, the source URL gets the destination scheme, for example HTTP for an HTTPS-only source or a SAS restricted with spr=https, and the cross-account copy fails. Please carry the source scheme/UseSSL alongside AzureSourceEndpoint from the actual source config, or require HTTPS for canonical Azure endpoints.

weiliu1031 (commented):
Two correctness and security issues remain in the Azure cross-account copy path.

internal/snapshotio/storage/resolver.go
```diff
@@ -71,9 +71,22 @@ func ResolveForeignStorage(
 	}
 	var copier milvusstorage.CrossBucketCopier = foreignCM
 	if resolvedCfg.hasSpec && (direction == DirectionRestore || direction == DirectionCopySource) {
-		// Layer 2 restore uses foreign credentials against the target endpoint so
-		// the provider can authorize both sides of the server-side copy request.
-		copier, err = milvusstorage.NewRemoteChunkManager(ctx, restoreProviderCopyConfig(instanceCfg, resolvedCfg.foreignCfg))
+		var copyCfg *objectstorage.Config
+		if strings.TrimSpace(resolvedCfg.foreignCfg.AzureSourceSAS) != "" {
```
weiliu1031, line 89: **[P1] Do not construct the foreign metadata client with the source SAS**

This branch is reached only after foreignCM has already been constructed from resolvedCfg.foreignCfg at line 68. For Restore/CopySource, applySnapshotExternalSpecToConfig has placed source_sas_token into that config, while AzureSourceEndpoint is still empty because it is only populated later by restoreSASCopyConfig. Consequently, newAzureObjectStorageWithConfig sees a non-empty AzureSourceSAS, calls azureServiceURL with an empty endpoint, and returns azure copy source endpoint is empty; every cross-account Restore/CopySource request fails before reaching this branch. Please construct the metadata client from a clone with the Azure source-copy fields cleared and keep the SAS/endpoint only on copyCfg. A regression test should exercise the real NewRemoteChunkManager constructor instead of mocking it.

Issue #52474, opened 2026-08-12
[Enhancement]: Woodpecker WAL has no single-message size limit, unlike Pulsar/Kafka


xiaofan-luan:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## What would you like to be added?

Pulsar and Kafka each have a hard per-message byte limit that the broker enforces (`pulsar.maxMessageSize`, `kafka.message.max.bytes`), and `maxSingleMessageSize` in `internal/proxy/msg_pack.go` checks the encoded size of a single, unsplittable row against it before handing the message to the WAL. If the row is too large, Insert/Upsert now fails fast with a typed `ErrParameterTooLarge` instead of being retried by the WAL forever (see #52413, where a single oversized row silently stalled a DML channel on Pulsar/BookKeeper).

Woodpecker has no equivalent. `WoodpeckerConfig` (`pkg/util/paramtable/service_param.go`) has no single-entry byte limit -- the closest field, `woodpecker.client.segmentAppend.maxBatchBytes` (default 2 MiB), is a *batching* threshold, not a hard cap: its own doc says "the first entry is always taken, so a batch may exceed the byte cap by one entry." So today, `maxSingleMessageSize` returns `(0, false)` for `message.WALNameWoodpecker`, meaning an oversized row is never rejected client-side on Woodpecker, and there is nothing to check it against even if we wanted to.

If Woodpecker (client or server) has -- or is willing to add -- a real hard limit on a single log entry, we'd like to wire it into the same guard the other two backends already have.

**Suggested default: 10 MiB**, matching Kafka's `message.max.bytes` default in this repo (`pkg/util/paramtable/service_param.go`, `KafkaConfig.ProducerMessageMaxBytes`, `10 * 1024 * 1024`), so the same-sized row behaves consistently across backends by default.

## Why is this needed?

Without a hard limit (or a documented, enforced one we can check client-side), a single oversized row on a Woodpecker-backed deployment can reach the WAL and be retried indefinitely if the underlying storage or segment layer ultimately refuses it -- the same DML-channel-stall failure mode #52413 fixed for Pulsar/BookKeeper, just not yet closed for this backend.

PR #52773, opened 2026-08-22
fix: add configurable single-message size limit for Woodpecker WAL

AyushKashyapII:
PR Body:
issue: #52474

## Summary

Pulsar and Kafka both reject an oversized row on the proxy before it reaches
the WAL (`pulsar.maxMessageSize`, `kafka.producer.message.max.bytes`,
enforced in `channelmgr.getMaxSingleRowSize`). Woodpecker had no equivalent —
`getMaxSingleRowSize` explicitly returned "no limit" for it, so an oversized
row would be queued for append and retried indefinitely against the backend,
the same DML-channel-stall failure mode fixed for Pulsar/BookKeeper in #52413.

The vendored `zilliztech/woodpecker` library has no per-entry size field of
its own to enforce this (only batch-payload and gRPC transport caps, neither
of which reject a single large entry), so the guard has to live on the
Milvus side — the same way it already does for Pulsar.

## Changes

- Add `woodpecker.client.segmentAppend.maxSingleMessageSize`
  (`WoodpeckerConfig.MaxMessageSize`), default 10MiB, matching Kafka's
  default. Refreshable at runtime — it's a pure client-side check, not baked
  into an already-constructed client, so no restart is needed.
- Wire it into `channelmgr.getMaxSingleRowSize` for `WALNameWoodpecker`,
  giving it its own case instead of being grouped with RocksMQ's "no limit"
  case.
- `configs/milvus.yaml`: document the new default.
- Tests: split the old combined "rocksmq+woodpecker have no limit" test;
  Woodpecker now gets its own accept/reject-at-limit pair (mirroring the
  existing Kafka test), RocksMQ keeps its "no limit" behavior.

## Why this is safe

The check runs in `GenInsertMsgsByPartition` before any row is queued for
append, so an oversized row now fails fast with `ErrParameterTooLarge`
instead of being retried against the backend. No change to RocksMQ, and no
change to the existing cross-WAL packing threshold (`PulsarCfg.MaxMessageSize`),
which is a separate, intentionally backend-agnostic batching threshold, not
a hard per-backend limit.

### DIFF
--- a/internal/proxy/channelmgr/msg_pack.go
+++ b/internal/proxy/channelmgr/msg_pack.go
@@ -45,9 +45,11 @@ func getMaxSingleRowSize(walName message.WALName) (int, bool) {
 	case message.WALNameKafka:
 		limit := paramtable.Get().KafkaCfg.ProducerMessageMaxBytes.GetAsInt()
 		return limit, limit > 0
-	case message.WALNameRocksmq, message.WALNameWoodpecker:
-		// RocksMQ page size and Woodpecker batch size are not hard limits
-		// on an individual WAL entry.
+	case message.WALNameWoodpecker:
+		limit := paramtable.Get().WoodpeckerCfg.MaxMessageSize.GetAsInt()
+		return limit, limit > 0
+	case message.WALNameRocksmq:
+		// RocksMQ page size is not a hard limit on an individual WAL entry.
 		return 0, false
 	default:
 		return 0, false
--- a/pkg/util/paramtable/service_param.go
+++ b/pkg/util/paramtable/service_param.go
@@ -731,6 +731,7 @@ type WoodpeckerConfig struct {
 	AppendMaxRetries          ParamItem `refreshable:"true"`
 	AppendMaxBatchEntries     ParamItem `refreshable:"false"`
 	AppendMaxBatchBytes       ParamItem `refreshable:"false"`
+	MaxMessageSize            ParamItem `refreshable:"true"`
 	SegmentRollingMaxSize     ParamItem `refreshable:"true"`
 	SegmentRollingMaxTime     ParamItem `refreshable:"true"`
 	SegmentRollingMaxBlocks   ParamItem `refreshable:"true"`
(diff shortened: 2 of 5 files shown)


internal/proxy/channelmgr/msg_pack.go
```diff
@@ -45,9 +45,11 @@ func getMaxSingleRowSize(walName message.WALName) (int, bool) {
 	case message.WALNameKafka:
 		limit := paramtable.Get().KafkaCfg.ProducerMessageMaxBytes.GetAsInt()
 		return limit, limit > 0
-	case message.WALNameRocksmq, message.WALNameWoodpecker:
-		// RocksMQ page size and Woodpecker batch size are not hard limits
-		// on an individual WAL entry.
+	case message.WALNameWoodpecker:
+		limit := paramtable.Get().WoodpeckerCfg.MaxMessageSize.GetAsInt()
```
yhmo, line 49: internal/proxy/channelmgr/msg_pack.go line:49
Medium ---- The limit is read via GetAsInt(), which silently falls back to 0 when the configured value is not a plain integer. If an operator sets maxSingleMessageSize to a suffix form such as "10MB" (a format the sibling AppendMaxBatchBytes accepts via GetAsSize(), and the same config block's docs use "2MB"/"10MB" wording), getMaxSingleRowSize returns (0, false) and the guard is silently disabled, re-opening the DML-channel-stall failure mode this PR is meant to close, with no log or error surfaced. Consider reading it with GetAsSize() like AppendMaxBatchBytes, or validating the value so a misconfigured limit fails loudly instead of disabling the check.

pkg/util/paramtable/service_param.go
```diff
@@ -830,6 +831,15 @@ func (p *WoodpeckerConfig) Init(base *BaseTable) {
 	}
 	p.AppendMaxBatchBytes.Init(base.mgr)
 
+	p.MaxMessageSize = ParamItem{
+		Key:          "woodpecker.client.segmentAppend.maxSingleMessageSize",
+		Version:      "2.6.21",
+		DefaultValue: strconv.Itoa(10 * 1024 * 1024),
+		Doc:          "Maximum size of a single WAL entry in bytes, default 10MB. A row whose encoded size exceeds this limit is rejected on the client before it is queued for append, instead of being retried indefinitely against the WAL backend. Unlike other segmentAppend settings, this is a Milvus-side check only (Woodpecker itself has no per-entry size limit); it is re-read on every check, so it takes effect without a restart.",
```
yhmo, line 838: pkg/util/paramtable/service_param.go line:838
Low ---- The doc says a row whose size "exceeds" this limit is rejected, but the check in GenInsertMsgsByPartition is curRowMessageSize >= singleRowLimit, so a row exactly at the limit is also rejected (the new test "woodpecker rejects row at its own limit" codifies this boundary). Consider aligning the wording (e.g., "reaches or exceeds") or the boundary, since a maintainer trusting "exceeds" will mis-model which rows pass the guard.

AyushKashyapII:
ending this PR since the core issue has been resolved

Issue #52777, opened 2026-08-23
[Enhance]: Unify bloom_match / roaring_match into a single membership_match expression chain


xiaofan-luan:
## Motivation

`bloom_match` (#51139) and `roaring_match` (#52094) were designed and landed separately. They share most of their machinery — deferred-call parsing, template fill, request budgeting, delete and element-level guards, log redaction, and segcore execution control flow — but each copy lives in its own chain.

Consequences:

1. Cross-cutting fixes must be applied twice and can silently miss one kind.
2. The implementations can drift even when they intend to preserve the same semantics.
3. Adding another membership structure would duplicate the same control flow again.

## Proposal

1. **Unified surface syntax**: add `membership_match(field, {blob})`. `MBF1` lowers to `BloomFilterExpr`, `MRB1` lowers to `RoaringFilterExpr`, and unknown headers fail closed. `bloom_match` / `roaring_match` remain permanent compatibility aliases; the wire format is unchanged.
2. **Single control flow**: use one visitor/fill skeleton, one behavior spec table, and unified tree walkers in Go; use one `PhyMembershipFilterExpr<LogicalExpr, ProbePolicy>` template in segcore while retaining kind-specific validators and probe policies.
3. **Preserve current execution semantics**: candidate-mask-excluded rows remain untouched at `(false, valid)` on raw and index-only paths, including excluded NULL rows; probed NULL rows remain `(false, invalid)`. No framework-wide reverse-lookup semantic change is part of this issue.
4. **Keep two independent resource limits**: `proxy.maxMembershipFilterSize` unifies the per-body Bloom/Roaring limit with compatibility fallbacks, while `proxy.maxMembershipFilterPlanSize` remains the request-wide aggregate serialized-plan limit with its legacy plan-key fallback. Main predicates, scorer filters, and HybridSearch sub-requests share one preflight budget.

## Non-goals

- No change to MBF1/MRB1 formats, golden vectors, or SDK builders.
- No change to existing alias wire plans or rolling-upgrade behavior.
- No new plan-proto messages; fields 22/23 remain frozen.
- No repo-wide NULL/candidate-mask semantic change.

Design document: `docs/design-docs/design_docs/20260822-membership-match-expression.md` in PR #52778.

PR #52778, opened 2026-08-23
enhance: unify membership filter expressions into one chain with membership_match syntax

xiaofan-luan:
## Overview

Adds a unified membership-filter surface syntax and merges the previously separate `bloom_match` / `roaring_match` chains into one parameterized execution path.

issue: #52777

### 1. Unified surface syntax

```
membership_match(<field>, {<blob-bytes-template>})
membership_match(<field>, {<blob-bytes-template>}, type=bloom|roaring)
not membership_match(<field>, {<blob-bytes-template>})
```

The filter kind is derived from the blob's self-describing magic header — `MBF1` lowers to the existing `BloomFilterExpr` plan node, `MRB1` to `RoaringFilterExpr`, and unknown headers fail closed. The optional `type=` argument is a readability aid and a consistency pin (it must agree with the magic). The unreleased `bloom_match` / `roaring_match` surface names are intentionally not retained — neither has shipped in a release. The wire plan bytes are unchanged from the two predecessor implementations, so the plan proto stays frozen and rolling-upgrade behavior for clients using `membership_match` matches what the predecessor docs specified.

### 2. One control flow, per-kind data planes

- **Go parser/proxy**: one visitor/fill skeleton (`visitMembershipCall` / `fillMembershipMatchExpressionValue`), one spec table for behavior switches (delete safety), and unified tree walkers serving delete guards, element-level guards, plan-size accounting, and log redaction.
- **segcore**: a single `PhyMembershipFilterExpr<LogicalExpr, ProbePolicy>` template carries the shared control flow; `BloomMembershipProbe` (zero-copy SBBF view) and `RoaringMembershipProbe` (decode-once Roaring64) remain separate data planes. Historical class names remain aliases, so factory construction and logs are unchanged.
- Kind-specific admission gates (MBF1 envelope + domain check, MRB1 body walk + decoded-memory estimate) remain in their own files.
- Client builders move under `client/membership/` (`sbbf`, `roaringfilter`); `bloom_match`/`roaring_match` builder packages are removed with the surface names.

### 3. Preserve existing mask/null semantics

The unified executor preserves current `master` behavior on both raw-data and index-only paths: candidate-mask-excluded rows are left at their initialized `(false, valid)` value, including excluded NULL rows; probed NULL rows produce `(false, invalid)`. This PR does not change the framework reverse-lookup helper.

### 4. Two independent proxy budgets

- `proxy.maxMembershipFilterSize` (default 64 MiB) limits one MBF1 or MRB1 body. It falls back to the released `proxy.maxBloomFilterSize` key and the development-only `proxy.maxRoaringFilterSize` key.
- `proxy.maxMembershipFilterPlanSize` (default 128 MiB) remains the separate request-wide aggregate serialized-plan limit. It falls back to `proxy.maxBloomFilterPlanSize`.
- Main predicates, scorer filters, and HybridSearch sub-requests share the same preflight budget. The preflight charges body bytes, matching the per-blob gate that allows the fixed 32-byte envelope header on top; final `proto.Size` accounting remains authoritative. Keeping the plan budget (128 MiB) well above the per-body limit (64 MiB) means a maximum-size body plus its 32-byte envelope and plan overhead always fits — no silent halving of the usable tier.

## Deliberate behavior changes

1. Adds `membership_match` with magic-based MBF1/MRB1 dispatch, an optional `type=` consistency pin, and fail-closed handling for unknown headers.
2. Removes the unreleased `bloom_match`/`roaring_match` surface names (and their client builder packages); the wire plan bytes for equivalent inputs are unchanged.
3. Updates error wording in bytes-template, element-level, MATCH_* and delete guards.
4. MATCH_* element-predicate guards now reject every membership-filter surface name.
5. Preflight aggregate accounting charges blob bodies rather than whole envelopes, matching the per-blob admission rule.

## Non-goals

- No change to MBF1/MRB1 formats, golden vectors, or SDK builders.
- No change to the plan proto or rolling-upgrade contract: fields 22/23 remain frozen.
- No repo-wide NULL/candidate-mask semantic change.

## Design document

[20260822-membership-match-expression.md](https://github.com/xiaofan-luan/milvus/blob/unify-membership-match/docs/design-docs/design_docs/20260822-membership-match-expression.md)

## Test plan

- [x] `internal/parser/planparserv2`: full suite green, including magic dispatch, alias byte identity, fail-closed cases, delete safety, redaction, and shared preflight budgeting.
- [x] `pkg/util/paramtable`: fallback-key precedence tests.
- [x] `internal/proxy`: membership size/error/search/query focused tests green; the full local suite is green except `TestProxy`, which requires unavailable integration infrastructure and fails identically on clean HEAD.
- [x] segcore: all 73 bloom/roaring/membership expression tests green; full `all_tests` has only failures reproduced on clean HEAD (6 ConjunctExprTest + 1 ExprJsonIndexTest abort).
- [x] Local segcore build and unit-test binary link with llvm@19.


internal/parser/planparserv2/roaring_match.go
buqian-zilliz, line 1: **[P3] Dead exported code: `EstimateRoaringFilterPlanDecodedBytes` and roaring-only predicates are never called in production**

`EstimateRoaringFilterPlanDecodedBytes` (roaring_match.go) and `PlanContainsRoaringFilter` / `hasRoaringFilterExpr` (membership_filter.go) are now referenced only by tests. The aggregate decoded-memory gate moved into `preflightMembershipFilterValues`, and this exported function re-runs `roaringfilter.Validate` over blobs that were already validated at fill time but is never invoked from production code. That leaves a duplicated accounting path that can silently drift: a reader could assume the decoded-size limit is enforced at the proxy serialization step (`marshalPlanWithMembershipFilterSizeLimit`, which only checks `proto.Size`), when it actually lives only in the parser preflight. Recommend removing the dead functions, or wiring the final gate to use them.

internal/parser/planparserv2/membership_filter.go
buqian-zilliz, line 1: **[P3] `membership_match` fill-time field error loses the field name and names the wrong function**

`fillMembershipMatchExpressionValue` re-validates the probe column with the literal string `deferred column parameter` passed as `argText` to `checkMembershipField`. For `membership_match` the field type is only known at fill time, so this is the first place a field-domain mismatch is reported; the resulting error renders the field as `deferred column parameter` and cites `bloom_match`/`roaring_match` rather than the syntax the client actually wrote. Thread the real column name (and the actual function name) through so the input error points at the offending field.

docs/design-docs/design_docs/20260822-membership-match-expression.md
```diff
@@ -0,0 +1,228 @@
+# Unified Membership-Filter Expression: `membership_match`
+
+- **Issue**: [#52777](https://github.com/milvus-io/milvus/issues/52777)
+- **Status**: Draft
+- **Date**: 2026-08-22
+- **Previous design docs**:
+  - [20260707-bloom-filter-expression.md](./20260707-bloom-filter-expression.md) — `bloom_match` (approximate)
```
zhengbuqian, line 7: also add deprecation notice in those 2 docs.

docs/design-docs/design_docs/20260822-membership-match-expression.md
```diff
@@ -0,0 +1,228 @@
...
+Add a unified surface syntax
+
+```
+membership_match(<field>, {<blob-bytes-template>})
+not membership_match(<field>, {<blob-bytes-template>})
+```
+
+whose filter kind is derived from the blob's self-describing magic header:
+`MBF1` lowers to the existing `BloomFilterExpr` plan node (approximate), `MRB1`
+to the existing `RoaringFilterExpr` node (exact). `bloom_match` and
```
zhengbuqian: do we need to keep them? considering we haven't made a release since implementing them.

docs/design-docs/design_docs/20260822-membership-match-expression.md
```diff
@@ -0,0 +1,228 @@
...
+| | `bloom_match` | `roaring_match` | `membership_match` |
+|---|---|---|---|
+| Kind | static (name) | static (name) | dynamic (blob magic) |
+| Blob format | MBF1 only | MRB1 only | sniffed: MBF1 → bloom, MRB1 → roaring |
+| Fields | INT8–64, VARCHAR, JSON paths | INT8–64 only | enforced per resolved kind |
+| Delete | rejected | allowed | per resolved kind |
+
+Why no explicit type argument (`type=bloom`)? Two reasons:
+
+1. The Milvus expression grammar has no keyword arguments; adding them for one
```
zhengbuqian: we do have operator param for TextMatch: `TextMatch(..., minimum_should_match=2)`

docs/design-docs/design_docs/20260822-membership-match-expression.md
```diff
@@ -0,0 +1,228 @@
...
+   function is a disproportionate grammar change.
+2. Both envelopes are already self-describing: 4-byte magic at offset 0
```
zhengbuqian: an explicit param makes the expr more readable? such as in logs, etc

client/membership/sbbf/sbbf.go
```diff
@@ -293,7 +293,7 @@ func Parse(blob []byte) (*Filter, error) {
 		return nil, errors.Errorf("bloom filter blob too short: %d bytes, need at least %d", len(blob), HeaderSize)
 	}
 	if string(blob[0:4]) != Magic {
-		return nil, errors.Errorf("bloom filter blob has invalid magic %q, expected %q", blob[0:4], Magic)
+		return nil, errors.Errorf("bloom filter blob has invalid magic, expected %q", Magic)
```
zhengbuqian, line 296: consider moving the codes of bf/rb into a same member_match folder?

internal/parser/planparserv2/plan_redact.go
```diff
@@ -0,0 +1,145 @@
...
+type membershipRedactedPlan struct{ plan *planpb.PlanNode }
+
+// redactTemplateValues swaps every value that arrived through an expression
+// template for a marker, returning a restore func.
+//
+// Template substitution keeps template_variable_name on the filled node, which
+// is what lets this tell a caller-supplied out-of-band value from a literal
+// written into the expression text. Only the former is elided: a literal is
+// already in the caller's hands and is what makes a logged plan readable.
+func redactTemplateValues(root *planpb.Expr) func() {
```
zhengbuqian, line 41: this handles more than the membership operator, consider renaming the `membershipRedactedPlan` struct and this file name to a more generic one.

zhengbuqian (commented):
the code part lgtm

xiaofan-luan:
@zhengbuqian Could you please take another look?

Addressed the latest review feedback in ceb85f5ee9:
- moved the Bloom and Roaring client builders under client/membership/
- generalized plan log redaction naming to plan_redact.go and redactedPlan
- removed the unreleased bloom_match/roaring_match aliases and added the optional type=bloom|roaring form without reserving field names

The parser and client membership/milvusclient test suites pass.

internal/parser/planparserv2/plan_redact.go
```diff
@@ -0,0 +1,144 @@
...
+	// Swap each blob for a short {N bytes} marker (a byte-slice pointer
+	// assignment — no copy, unlike proto.Clone which would duplicate the
+	// up-to-tens-of-MiB body), stringify, then restore the originals via defer.
+	// Safe because zap evaluates a Stringer field synchronously in this
+	// goroutine at the log call, and these call sites own the task-local plan, so
+	// no other reader observes the temporary state.
+	saved := make([][]byte, len(slots))
+	for i, slot := range slots {
+		saved[i] = slot.get()
+		slot.set([]byte(fmt.Sprintf("<%d bytes elided>", len(saved[i]))))
```
zhengbuqian: Replace []byte(fmt.Sprintf...) with fmt.Appendf

zhengbuqian (approved):
I think we want to remove those files:

client/membership/roaringfilter/build_bench_test.go
client/membership/sbbf/testdata/

internal/parser/planparserv2/plan_redact.go
xiaofan-luan: Done — replaced `[]byte(fmt.Sprintf(...))` with `fmt.Appendf` in the rebased revision.

internal/parser/planparserv2/plan_redact.go
xiaofan-luan: Done in ceb85f5 — file renamed to `plan_redact.go` and the struct is now `redactedPlan`; it covers all template-value redaction, not just membership blobs.

client/membership/sbbf/sbbf.go
xiaofan-luan: Done — the SBBF and roaring builders both live under `client/membership/` (`client/membership/sbbf`, `client/membership/roaringfilter`) in the same folder.

docs/design-docs/design_docs/20260822-membership-match-expression.md
xiaofan-luan: Removed them — since neither `bloom_match` nor `roaring_match` has shipped in a release, the surface names (and their client builder packages) are dropped entirely; the optional `type=bloom|roaring` pin was added instead. The design doc and both predecessor docs now state this explicitly.

docs/design-docs/design_docs/20260822-membership-match-expression.md
xiaofan-luan: Since the aliases are removed (never released) rather than deprecated, I added a superseded notice to both predecessor docs instead — see the headers of 20260707-bloom-filter-expression.md and 20260714-roaring-exact-membership-expression.md.

docs/design-docs/design_docs/20260822-membership-match-expression.md
xiaofan-luan: Good point — updated the design doc to cite the existing `text_match(field, "query", minimum_should_match=2)` operator-argument convention as the precedent for `type=`.

docs/design-docs/design_docs/20260822-membership-match-expression.md
xiaofan-luan: Agreed — kept the explicit `type=bloom|roaring` form; the design doc now explains it as a readability aid (expressions/logs) plus a consistency pin that must agree with the blob magic.

internal/parser/planparserv2/membership_filter.go
xiaofan-luan: Fixed — the fill-time re-validation now threads the real column name through `ctx.membershipFieldName` (schema lookup incl. nested paths) and the actual function name, so a field-domain error names the offending field and cites `membership_match`. Tests assert the old `deferred column parameter` wording is gone.

internal/parser/planparserv2/roaring_match.go
xiaofan-luan: Removed — `EstimateRoaringFilterPlanDecodedBytes`, `PlanContainsRoaringFilter`, and `hasRoaringFilterExpr` are gone; only the unified `PlanContainsMembershipFilter` / `PlanContainsMembershipFilterUnsafeForDelete` walkers remain, used by proxy plan-size accounting and the delete guard. The decoded-memory gate lives solely in the parser preflight.

internal/proxy/filter_plan_size.go
```diff
@@ -0,0 +1,109 @@
...
+func marshalPlanWithFilterSizeLimits(
+	plan *planpb.PlanNode,
+	accumulatedSize int64,
+	accumulatedRoaringDecodedBytes uint64,
+) ([]byte, int64, uint64, error) {
+	nextSize := accumulatedSize
+	if planparserv2.PlanContainsMembershipFilter(plan) {
+		planSize := int64(proto.Size(plan))
+		maxSize := paramtable.Get().ProxyCfg.MaxMembershipFilterSize.GetAsInt64()
+		if accumulatedSize > maxSize || planSize > maxSize-accumulatedSize {
```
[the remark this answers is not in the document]
xiaofan-luan: Addressed by the two-budget redesign in the later revisions: `proxy.maxMembershipFilterSize` (64 MiB) gates one blob *body* with the fixed 32-byte envelope allowed on top, while the request-wide aggregate gate uses the separate `proxy.maxMembershipFilterPlanSize` (128 MiB). A 64 MiB body + 32-byte envelope + plan overhead stays well under 128 MiB, so the usable tier is no longer halved — this is now also stated in the PR description.

client/milvusclient/write_options.go
```diff
@@ -630,30 +630,50 @@ func (opt *rowBasedDataOption) WithKeepAutoIDPk(keepPk bool) *rowBasedDataOption
 }
 
 type DeleteOption interface {
-	Request() *milvuspb.DeleteRequest
+	Request() (*milvuspb.DeleteRequest, error)
 }
```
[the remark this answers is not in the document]
xiaofan-luan: This one looks stale — `DeleteOption.Request()` already returns two values on master (introduced by #51968), and this PR does not touch `client/milvusclient/write_options.go` (its diff is empty). No signature change here.

Issue #52771, opened 2026-08-22
bug: compaction materializes added default-valued fields as NULL, losing the default irreversibly (storage reader is default-unaware)


MrPresent-Han:
### Summary

When an ordinary field with a declared `default_value` is added to a collection with existing sealed data (`AddCollectionField`), a later compaction that reads and rewrites those segments **physically materializes the field as NULL instead of its default value**, and the query-side default is lost irreversibly.

This is a storage **reader-contract** violation, orthogonal to any single feature PR.

### First-principles expectation

Standard SQL semantics for `ADD COLUMN c T DEFAULT v` on a non-empty table: existing rows must read back as `v` (the default), never NULL. Whether the value is physically stored or synthesized on read is an implementation choice, but the observable result must be the default. Milvus follows the Postgres-style "synthesize on read": segcore returns the default for a **physically-absent** field via `bulk_subscript_not_exist_field` (`internal/core/src/segcore/SegmentInterface.cpp`), which only fires when `!is_field_exist(field_id)`.

### Root cause

Compaction reads segments through the storage reader with the full/read schema, which includes the added-but-physically-absent field. The storage layer fills absent columns with **NULL**, because `default_value` is a Milvus schema concept the storage engine has no knowledge of — the reader is only handed the arrow schema + needed-column names, never the default:

- **V2 packed path** — `PackedRecordBatchReader::ReadNext` (milvus-storage `cpp/src/packed/reader.cpp:~373`): a needed field not in `field_id_mapping_` → `arrow::MakeArrayOfNull(schema_->field(i)->type(), chunk_size)`.
- **V3 manifest path** — `ReaderImpl` via `Reader::create` (milvus-storage `cpp/src/reader.cpp`): `out_field_map_[i] == -1` (needed field not in any column group, built at `:212`) → `arrow::MakeArrayOfNull(out_schema_->field(i)->type(), min_rows)` at `:312`.

Both entry points (`NewPackedReaderWithExtfs` for V2, `NewPackedFFIReaderWithManifest`/`Reader::create` for V3) receive only `arrow_schema` + `needed_columns`; there is no `default_value` available to them.

[...]

PR #52781, opened 2026-08-23
fix: present declared defaults for absent fields in the StorageV2/V3 record reader

MrPresent-Han:
The packed StorageV2/V3 record readers returned NULL for a read-schema field physically absent from a segment even when the field declares a `default_value`, while the V1 binlog reader and segcore synthesize the default. This makes the packed readers default-aware, matching V1: they read only physically-present columns and fill every absent read-schema field via `GenerateEmptyArrayFromSchema` (its declared default, else null). Physical presence is the reader's own truth:

- **StorageV3 manifest reader** — from `packed.GetManifestFieldIDs`, or a caller-supplied present set (compaction passes its already-computed `existingFields` via `WithPresentFields` to avoid a redundant manifest open).
- **StorageV2/V3 binlog reader** — from `FieldBinlog.ChildFields` (the column group's member field IDs, written by the flush/compaction writers).

**Struct array fields** are physically all-or-nothing (add/drop/update act on the whole struct): a struct is kept whole or dropped whole, and a partially-present struct is surfaced as a data-integrity error rather than read partial.

**Import-restore** reads reconstruct FieldBinlogs from object-storage directory names (`FieldID` is a column-group ID, no `ChildFields`), so physical presence is not derivable there; those reads pass the full schema through unchanged (the engine null-fills absent columns) — no crash, no PK strip. Import-restore default-fill is a tracked follow-up on #52771 (it needs presence from the packed files' arrow schema, or an extended import contract).

**Scope / current effect (honest):** this is the reader-contract prerequisite for #52484. Today the fill is inert on the compaction path — `compactionReadSchema` reduces the read schema to `existingFields`, and the `RecordMaterializer` still fills ordinary-field defaults — so the new reader path is currently exercised by the tests in this PR, not by a live compaction. It goes live once #52484 removes the RM's ordinary-field fill and makes `base` read-schema-wide; **#52781 must merge before #52484** to avoid a window where neither fills (the #52771 NULL). This PR does not, by itself, change compaction output on master.

**Behavior change:** an absent non-nullable read-schema field now errors instead of being silently null-filled, matching the V1 reader. External-source reads pass through unchanged.

**Follow-ups:** import-restore default-fill (above); de-dupe the absent-fill overlay record shared with the compactor's `RecordMaterializer` (`absentFilledRecord` vs `materializedRecord`).

issue: #52771

### DIFF
--- a/internal/storage/record_reader.go
+++ b/internal/storage/record_reader.go
@@ -490,3 +490,107 @@ func (crr *CompositeBinlogRecordReader) releaseCurrent() {
 		crr.current = nil
 	}
 }
+
+// NewAbsentFieldFillRecordReader completes a partial record to full read-schema
+// width the way V1's CompositeBinlogRecordReader does: every read-schema field NOT
+// physically present (FieldID not in presentFields) is filled via
+// GenerateEmptyArrayFromSchema -- its declared default when it has one, else null,
+// erroring on a non-nullable absent field. Present columns pass through from inner.
+// This gives the packed StorageV3 manifest reader the same default-for-absent
+// semantic the V1 binlog reader already applies, so it stops presenting declared
+// defaults as NULL (issue #52771). presentFields is the physically-present field
+// set -- self-sourced from the manifest, or supplied by a caller that already
+// computed it (compaction via WithPresentFields). Returns inner unchanged when
+// nothing is absent.
+func NewAbsentFieldFillRecordReader(inner RecordReader, neededSchema *schemapb.CollectionSchema, presentFields map[FieldID]struct{}) RecordReader {
+	fill := make([]*schemapb.FieldSchema, 0)
+	for _, f := range typeutil.GetAllFieldSchemas(neededSchema) {
+		if _, present := presentFields[f.GetFieldID()]; present {
+			continue
+		}
+		fill = append(fill, f)
+	}
+	if len(fill) == 0 {
+		return inner
+	}
+	return &absentFieldFillRecordReader{inner: inner, fill: fill}
+}
+
+type absentFieldFillRecordReader struct {
+	inner RecordReader
+	fill  []*schemapb.FieldSchema
+	cur   *absentFilledRecord
+}
+
+var _ RecordReader = (*absentFieldFillRecordReader)(nil)
+
+func (r *absentFieldFillRecordReader) Next() (Record, error) {
+	r.releaseCur()
+	base, err := r.inner.Next()
+	if err != nil {
+		return nil, err
+	}
+	computed := make(map[FieldID]arrow.Array, len(r.fill))
+	for _, f := range r.fill {
+		arr, genErr := GenerateEmptyArrayFromSchema(f, base.Len())
+		if genErr != nil {
+			for _, a := range computed {
+				a.Release()
+			}
+			return nil, genErr
+		}
+		computed[f.GetFieldID()] = arr
+	}
+	base.Retain()
+	r.cur = &absentFilledRecord{base: base, computed: computed}
+	return r.cur, nil
+}
+
+func (r *absentFieldFillRecordReader) releaseCur() {
+	if r.cur == nil {
+		return
+	}
+	r.cur.Release()
+	r.cur = nil
+}
+
+func (r *absentFieldFillRecordReader) Close() error {
+	r.releaseCur()
+	if r.inner == nil {
+		return nil
+	}
+	return r.inner.Close()
+}
+
+// absentFilledRecord overlays filled columns onto base (the present columns). It
+// owns a retained ref on base plus the filled arrays; Release drops both, so the
+// wrapping reader frees them by releasing this record on its next Next/Close.
+type absentFilledRecord struct {
+	base     Record
+	computed map[FieldID]arrow.Array
+}
+
+var _ Record = (*absentFilledRecord)(nil)
+
+func (r *absentFilledRecord) Column(i FieldID) arrow.Array {
+	if col, ok := r.computed[i]; ok {
+		return col
+	}
+	return r.base.Column(i)
+}
+
+func (r *absentFilledRecord) Len() int { return r.base.Len() }
+
+func (r *absentFilledRecord) Retain() {
+	r.base.Retain()
+	for _, col := range r.computed {
+		col.Retain()
+	}
+}
+
+func (r *absentFilledRecord) Release() {
+	r.base.Release()
+	for _, col := range r.computed {
+		col.Release()
+	}
+}
--- a/internal/storage/rw.go
+++ b/internal/storage/rw.go
@@ -24,6 +24,7 @@ import (
 	"sort"
 
 	"github.com/samber/lo"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/schemapb"
 	"github.com/milvus-io/milvus/internal/allocator"
@@ -77,6 +78,7 @@ type rwOptions struct {
 	textRefsAsBinary    bool                      // TEXT columns already contain encoded LOB refs and should be copied as-is
 	externalReader      packed.ExternalReaderContext
 	writerFormat        string
+	presentFields       map[FieldID]struct{} // reader: caller-known physically-present field IDs, skips a manifest re-read
 }
 
 func (o *rwOptions) validate() error {
@@ -182,6 +184,15 @@ func WithNeededFields(neededFields typeutil.Set[int64]) RwOption {
 func WithExternalReaderContext(externalReader packed.ExternalReaderContext) RwOption {
 	return func(options *rwOptions) {
 		options.externalReader = externalReader
+	}
+}
+
+// WithPresentFields lets a caller that already knows a segment's physically-present
+// field IDs hand them to NewManifestRecordReader, so it skips re-deriving them via
+// packed.GetManifestFieldIDs (a redundant manifest open on the compaction path).
+func WithPresentFields(present map[FieldID]struct{}) RwOption {
+	return func(options *rwOptions) {
+		options.presentFields = present
 	}
 }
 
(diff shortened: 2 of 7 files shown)


internal/storage/rw.go
```diff
@@ -388,8 +393,82 @@ func NewManifestRecordReader(ctx context.Context, manifestPath string, schema *s
...
+	// External-source reads (e.g. parquet) are not milvus manifests carrying
+	// default-aware internal field IDs and are out of scope for #52771; keep the
+	// original pass-through. An internal manifest reads only its physically-present
+	// columns, then fills every absent read-schema field with its schema default
+	// (or null) so declared defaults are not presented as NULL, matching V1.
+	if rwOptions.externalReader.Source != "" {
+		return NewRecordReaderFromManifest(manifestPath, schema, rwOptions.bufferSize,
+			rwOptions.storageConfig, pluginContext, option...)
+	}
+	present, err := packed.GetManifestFieldIDs(manifestPath, rwOptions.storageConfig)
```
buqian-zilliz: **[P2] NewManifestRecordReader re-reads the StorageV3 manifest, adding a redundant open on the compaction hot path**

`NewManifestRecordReader` now calls `packed.GetManifestFieldIDs` to derive `present`, but its compaction caller (`newCompactionSegmentRecordReaderWithFields` -&gt; `compactionSegmentStorageFields`) already invokes the exact same `packed.GetManifestFieldIDs` for the same manifest to build `existingFields`, and `NewRecordReaderFromManifest` opens the manifest again inside `packed.NewFFIPackedReader`. On the compaction path the manifest is now opened/read up to three times per segment (was two before this change). Each `GetManifestFieldIDs` performs a `loon_transaction_begin`/`get_manifest` object-store round trip, so this adds one redundant storage transaction per compacted segment. The compactor already holds a caller-known present set (`existingFields`); thread it through (for example a `WithPresentFields` option) instead of re-reading. (The PR body lists this as a follow-up, but it is a current cost in the diff.)

internal/storage/record_reader.go
```diff
@@ -490,3 +490,106 @@ func (crr *CompositeBinlogRecordReader) releaseCurrent() {
...
+
+func (r *absentFieldFillRecordReader) Next() (Record, error) {
+	r.releaseCur()
+	base, err := r.inner.Next()
+	if err != nil {
+		return nil, err
+	}
+	computed := make(map[FieldID]arrow.Array, len(r.fill))
+	for _, f := range r.fill {
+		arr, genErr := GenerateEmptyArrayFromSchema(f, base.Len())
```
buqian-zilliz, line 535: **[P3] Absent non-nullable field now hard-errors on read instead of being null-filled (behavior change)**

For every absent read-schema field, `Next()` calls `GenerateEmptyArrayFromSchema`, which returns an error when `field.GetNullable()` is false. Previously the StorageV2/V3 reader passed the full schema to the packed/FFI reader and an absent column was surfaced as NULL. After this change, reading a segment that physically lacks a non-nullable field fails the whole reader rather than yielding null. This matches V1 and is documented, but it is a behavior change: if any existing V2/V3 segment legitimately lacks a non-nullable, default-less field (for example a field added without a default), compaction/import reads of that segment will now fail instead of succeeding with nulls. Please confirm no such legacy segments exist before this ships.

internal/storage/rw.go
```diff
@@ -388,8 +393,82 @@ func NewManifestRecordReader(ctx context.Context, manifestPath string, schema *s
...
+	}
+	return out
+}
+
+// binlogFieldIDSet returns the field IDs physically present in binlogs: a field
+// with a binlog is stored, and a struct container contributes its child field IDs.
+func binlogFieldIDSet(binlogs []*datapb.FieldBinlog) map[FieldID]struct{} {
+	present := make(map[FieldID]struct{}, len(binlogs))
+	for _, fb := range binlogs {
+		if len(fb.GetChildFields()) == 0 {
```
liliu-z: For StorageV2/V3 the binlog import path keys `datapb.FieldBinlog.FieldID` by *column-group* ID, not field ID, and never sets `ChildFields`: `listInsertLogs` parses the group directory name from the object path, `verify()` returns the target schema unfiltered for V2/V3, and `createFieldBinlogList` builds `&datapb.FieldBinlog{FieldID: fieldID}` with no children. `binlogFieldIDSet` therefore returns `{0 (DefaultShortColumnGroupID)} ∪ {vector group IDs}`, `filterSchemaToPresentFields` strips the PK and every scalar from the read schema, and the first `Next()` hits `GenerateEmptyArrayFromSchema` on the non-nullable PK and fails with `missing field data <pk>`. That breaks restore-from-backup for every StorageV2/V3 collection, not just legacy segments. Either fall back to the unfiltered schema when a binlog group carries no `ChildFields`, or populate `ChildFields` when constructing the import binlog list.

internal/storage/record_reader.go
MrPresent-Han: Thanks — this scenario cannot legitimately arise, so the fail-loud is a correctness fix rather than a regression:

- A non-nullable field is never physically absent: `AddCollectionField` rejects a non-nullable added field (error `added field must be nullable`; struct fields likewise), and the only non-nullable fields — PK, vectors, RowID/Timestamp — are written into every segment at creation. So no legacy segment legitimately lacks a non-nullable field.
- The pre-change behavior (surfacing an absent non-nullable column as NULL) was itself invalid — a NULL in a NOT NULL column. `GenerateEmptyArrayFromSchema` correctly refuses it, matching V1 `CompositeBinlogRecordReader`, which uses the same helper and already errors here.

So the error fires only on genuine corruption (a required column missing), where failing loud beats fabricating a NULL. Note: per the finding from @liliu-z, the StorageV2/V3 binlog path is being reverted, so this now applies only to the StorageV3 manifest reader.

internal/storage/rw.go
MrPresent-Han: Done in fd875b326f: added a `WithPresentFields` reader option and threaded the compactor's already-computed `existingFields` through it, so `NewManifestRecordReader` skips the extra `GetManifestFieldIDs`. Back to the pre-PR manifest-read count on the compaction path.

internal/storage/rw.go
```diff
@@ -388,8 +399,71 @@ func NewManifestRecordReader(ctx context.Context, manifestPath string, schema *s
...
+	// original pass-through. An internal manifest reads only its physically-present
+	// columns, then fills every absent read-schema field with its schema default
+	// (or null) so declared defaults are not presented as NULL, matching V1.
+	if rwOptions.externalReader.Source != "" {
+		return NewRecordReaderFromManifest(manifestPath, schema, rwOptions.bufferSize,
+			rwOptions.storageConfig, pluginContext, option...)
+	}
+	present := rwOptions.presentFields
+	if present == nil {
+		present, err = packed.GetManifestFieldIDs(manifestPath, rwOptions.storageConfig)
```
buqian-zilliz, line 429: **[P3] NewManifestRecordReader adds an extra manifest open for non-compaction internal-manifest callers**

`NewManifestRecordReader` now calls `packed.GetManifestFieldIDs` (which opens the manifest via `GetManifestHandle`) whenever `presentFields` is not supplied, and then `NewRecordReaderFromManifest` opens the manifest again via `packed.NewFFIPackedReader`. The compaction path avoids the redundant open by passing `WithPresentFields(existingFields)`, but the other internal-manifest caller `openInputReader` in `internal/datanode/external/function_executor.go` does not, so that path goes from one manifest open per read (before this change) to two. If function-input reads are frequent, this doubles the `loon_transaction_begin`/object-store round trips on that path. Consider threading a caller-known present set there as well, or deriving the present fields during the single manifest open.

MrPresent-Han: Agreed, perf-only. The clean fix is to have the packed/FFI layer expose the manifest field IDs from its single open, so the extra `GetManifestFieldIDs` goes away; threading a present-set does not help here since `openInputReader` has no ready set and cannot assume inputs are always present (an input may be an added-default field absent in older segments). That is a `storagev2/packed` change, out of this reader PR scope, so keeping it as a follow-up.

internal/storage/rw.go
MrPresent-Han: Fixed in 670f03dbe9. Reverted the default-fill on the binlog path entirely: the V2/V3 case of `NewBinlogRecordReader` again passes the full unfiltered schema (no `binlogFieldIDSet`, no filter, no wrapper), unchanged from master. The reader default-fill is now scoped to the StorageV3 manifest reader only, where presence is reliably self-sourced from `packed.GetManifestFieldIDs`. Restore-from-backup on the binlog path is untouched.

internal/storage/record_reader.go
```diff
@@ -490,3 +490,107 @@ func (crr *CompositeBinlogRecordReader) releaseCurrent() {
...
+// physically present (FieldID not in presentFields) is filled via
+// GenerateEmptyArrayFromSchema -- its declared default when it has one, else null,
+// erroring on a non-nullable absent field. Present columns pass through from inner.
+// This gives the packed StorageV3 manifest reader the same default-for-absent
+// semantic the V1 binlog reader already applies, so it stops presenting declared
+// defaults as NULL (issue #52771). presentFields is the physically-present field
+// set -- self-sourced from the manifest, or supplied by a caller that already
+// computed it (compaction via WithPresentFields). Returns inner unchanged when
+// nothing is absent.
+func NewAbsentFieldFillRecordReader(inner RecordReader, neededSchema *schemapb.CollectionSchema, presentFields map[FieldID]struct{}) RecordReader {
```
liliu-z, line 505: The new absent-field fill is inert on every production path, so the PR does not change any live behavior. Compaction calls NewManifestRecordReader with readSchema = compactionReadSchema(schema, existingFields), which keeps only fields in existingFields, and then passes that same set as WithPresentFields — so the fill list in NewAbsentFieldFillRecordReader is always empty and it returns inner unwrapped. The other caller, external/function_executor.go:225, always sets ExternalReaderContext.Source (RefreshExternalCollectionTask.PreExecute rejects an empty external source at task_update.go:180), so it takes the pass-through branch at rw.go:407. On top of that, the compaction layer already synthesizes declared defaults for absent fields via RecordMaterializer (record_materializer.go:131-143). The title also says StorageV2/V3 while only the V3 manifest reader is touched — the V2 packed path (rw.go:364, still marked "FIXME: add needed fields support") is unchanged. Narrow the title/body to V3-only and state plainly that the new reader path is currently exercised only by tests rather than presenting it as an end-to-end fix for #52771.

MrPresent-Han: Agreed, thanks — the fill is inert today: compaction reduces the read schema (`compactionReadSchema`) and hands the same set back via `WithPresentFields`, so the fill list is empty and it returns `inner`; `function_executor` always takes the external pass-through; and the `RecordMaterializer` still fills ordinary-field defaults. So on master this PR changes no live behavior.

The point is that it is the **reader-contract prerequisite for #52484**: that PR removes the RM's ordinary-field fill and makes `base` read-schema-wide ("Per the reader contract base is readSchema-wide"), at which point the reader fill goes live. #52781 must merge first, otherwise there is a window where neither the RM nor the reader fills → the #52771 NULL.

On scope: I have added the **StorageV2/V3 binlog** path too (presence from `FieldBinlog.ChildFields`, reliable for flush/compaction segments), so it is genuinely V2/V3, not V3-only. The one path that cannot derive presence is **import-restore** (paths-only contract, no `ChildFields`) — it falls back to an unfiltered pass-through (no crash, no PK strip), tracked as a follow-up on #52771. I have updated the PR body to state the "currently exercised by tests, goes live with #52484" scope plainly.

liliu-z (commented):
/lgtm
/approve

Issue #52191, opened 2026-08-05
[Bug]: geometry cache uses a reader-preferring shared_mutex, allowing per-row writer starvation on the growing insert path


xiaocai2333:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
- Milvus version: master (verified at f113bdbc8e)
- Deployment mode(cluster/standalone): both
- MQ type(rocksmq, pulsar, kafka): n/a
- SDK version(e.g. pymilvus v2.0.0rc2): n/a
- OS(Ubuntu or CentOS): Linux (glibc)
```

### Current Behavior

`SimpleGeometryCache` guards its entries with a plain reader-preferring `std::shared_mutex`, which allows the per-row writer to be starved indefinitely by a steady stream of readers.

The asymmetry is in the lock hold patterns:

- **Writer — once per row, short critical sections.** `SimpleGeometryCache::AppendData` takes the mutex exclusively on every call (`internal/core/src/common/GeometryCache.h:41-42`), and callers drive it in a per-row loop: `internal/core/src/segcore/SegmentGrowingImpl.cpp:3081-3092` and `:3125-3136` (growing insert), `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:7884-7888` (sealed load).
- **Readers — once per batch, long critical sections.** The GIS filter path takes `AcquireReadLock()` and holds it across an entire refine loop: `internal/core/src/exec/expression/GISFunctionFilterExpr.cpp:58`, `:108`, `:159`, `:535`.

On glibc, `std::shared_mutex` is a pthread rwlock created with the default `PTHREAD_RWLOCK_PREFER_READER_NP` policy, so an arriving reader may barge past a waiting writer. With overlapping concurrent searches, each individual `AppendData` acquisition can be deferred indefinitely.

The sibling structure in the same subsystem already has this addressed: `RTreeIndexWrapper::rtree_mutex_` was converted to `folly::SharedMutexWritePriority` with an explicit anti-starvation rationale and a regression test in #50951 (`RTreeIndexWrapper.h`, `RTreeIndexWrapperTest.cpp`). `SimpleGeometryCache::mutex_` (`GeometryCache.h:101`) was left as-is.

[...]

PR #52783, opened 2026-08-23
fix: stop overlapping readers starving the geometry cache writer

xiaocai2333:
`SimpleGeometryCache` was a `std::vector<Geometry>` guarded by a `std::shared_mutex`. Because `resize()` relocates every element, a reader's `const Geometry*` had to be protected by a shared lock, and every GIS expression (`GISFunctionFilterExpr`, `GISConjunctExpr`) held that lock across a whole batch. On Linux `std::shared_mutex` is glibc's reader-preferring `pthread_rwlock_t`, so under sustained query load the reader count essentially never dropped to zero and the growing-segment writer in `AppendDataAt` was starved.

Changing the lock's preference would only move the cost (one queued writer per inserted row, each blocking newly arriving readers), so this PR removes the read lock instead, following the growing segment's own column storage (`ConcurrentVector` / `ThreadSafeChunkVector`).

## Design

- **`GeometryChunkStore`**: fixed 4096-slot chunks, allocated once and never moved, reached through a geometric slab directory whose slabs are never moved either. Every slab and chunk pointer is published with a release store only after it is fully allocated. The layout is duplicated rather than reused because `mmap/ChunkVector.h` pulls in the mmap manager and segcore config, which `common/` must not depend on.
- **Lock-free reads**: each slot has an atomic state byte that goes `kEmpty -> kReady` at most once, after its `Geometry` is final. `GetByOffset()` is three acquire loads and takes no lock; the pointer it returns stays valid as long as the caller holds its `shared_ptr` to the cache.
- **Per-slot publication, not a committed-row count**: writes land at absolute segment offsets that can arrive out of order, so there is no single committed prefix.
- **Writers** take a `std::mutex` that readers never touch, so neither path can delay the other.
- **First valid writer wins**: a slot holding a valid geometry is frozen and later writes to that offset are skipped. This is sound because an offset denotes one immutable row and a successful parse is a pure function of its WKB. An invalid outcome is *not* frozen: GEOS swallows an OOM inside `GEOSWKBReader_read_r`, which is indistinguishable from corrupt WKB, so null/unparseable rows are left unpublished and a retried batch re-parses them — the same healing the old unconditional overwrite gave.
- `AcquireReadLock()`, `GetByOffsetUnsafe()` and `SizeUnsafe()` are removed; the five expression read sites call `GetByOffset()` directly.

## Tests

`GeometryCacheTest.cpp` goes from 10 to 14 tests:

- `WriterIsNotStarvedByOverlappingReaders` — readers drive the read API continuously while a writer appends; fails if a lock is reintroduced on the read path.
- `ReadersNeverObserveAPartiallyPublishedSlot` — readers race writes spanning three chunks, including an unparseable pass followed by a healing pass; every geometry observed must be fully built.
- `PublishedSlotsAreFrozenAndStable` — pointer identity across rewrites and multi-chunk growth.
- `InvalidRowsAreRewrittenByARetry` — an unparseable or null row is re-parsed by a later write of the same offset, and frozen once valid.
- `SparseOffsetsAcrossChunkAndSlabBoundaries` — out-of-order writes across chunk and slab boundaries, gaps, and an out-of-range offset.

## Verification

- CI on the lock-free commit (`a35e7ca`): build, code check, integration, Go UT, 8765 C++ tests, e2e-default and go-sdk all passed. CI for the follow-up retry fix (`8705d48`) is pending.
- Locally, `GeometryCacheTest.cpp` built against a pre-built `libmilvus_core.so` (the cache is header-only, so the new code is what runs): 14/14 pass, and 14/14 under TSAN. The only TSAN reports are a `setlocale` race in `geos::io::CLocalizer` inside GEOS, which reproduces identically on the pre-change code.
- Red/green for the retry fix: `InvalidRowsAreRewrittenByARetry` and the extended `ReadersNeverObserveAPartiallyPublishedSlot` fail against `a35e7ca` and pass with the fix (`8705d48`).
- Not verified: a reader-latency benchmark under concurrent insert load. The claim that writes no longer delay queries rests on the read path taking no lock, not on a measurement. The retry-heal behaviour is covered by unit tests; the GEOS parse-time OOM itself was not fault-injected.

The geometry cache is disabled by default (`enableGeometryCache: false`), so the affected configuration is opt-in.

issue: #52191

### DIFF
--- a/internal/core/src/common/GeometryCache.h
+++ b/internal/core/src/common/GeometryCache.h
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <folly/SharedMutex.h>
 #include <atomic>
 #include <memory>
 #include <mutex>
@@ -79,7 +80,7 @@ class SimpleGeometryCache {
         // Destroy the cached geometries (each calls GEOSGeom_destroy_r(ctx_,
         // ...)) while ctx_ is still alive, then release the context.
         {
-            std::lock_guard<std::shared_mutex> lock(mutex_);
+            std::lock_guard<folly::SharedMutexWritePriority> lock(mutex_);
             geometries_.clear();
         }
         if (ctx_ != nullptr) {
@@ -122,7 +123,7 @@ class SimpleGeometryCache {
     // data here (see the KNOWN LIMIT note on TryParseFromWkb).
     void
     AppendDataAt(size_t absolute_offset, const char* wkb_data, size_t size) {
-        std::lock_guard<std::shared_mutex> lock(mutex_);
+        std::lock_guard<folly::SharedMutexWritePriority> lock(mutex_);
 
         if (geometries_.size() <= absolute_offset) {
             // Default-constructed gap entries are invalid; a reader that
@@ -156,9 +157,9 @@ class SimpleGeometryCache {
     }
 
     // Get shared lock for batch operations (RAII)
-    std::shared_lock<std::shared_mutex>
+    std::shared_lock<folly::SharedMutexWritePriority>
     AcquireReadLock() const {
-        return std::shared_lock<std::shared_mutex>(mutex_);
+        return std::shared_lock<folly::SharedMutexWritePriority>(mutex_);
     }
 
     // Get Geometry by offset without locking (use with AcquireReadLock).
@@ -185,7 +186,7 @@ class SimpleGeometryCache {
     // Get total number of loaded geometries
     size_t
     Size() const {
-        std::shared_lock<std::shared_mutex> lock(mutex_);
+        std::shared_lock<folly::SharedMutexWritePriority> lock(mutex_);
         return geometries_.size();
     }
 
@@ -194,9 +195,10 @@ class SimpleGeometryCache {
     // would recursively acquire the non-reentrant shared_mutex: that is
     // undefined behavior per [thread.sharedmutex.requirements] regardless of
     // platform, and becomes a real deadlock on writer-preferring
-    // implementations once a writer is queued. (Not a claim about this
-    // build's std::shared_mutex: on Linux libstdc++ maps it to
-    // pthread_rwlock_t, whose glibc default is reader-preferring.)
+    // implementations once a writer is queued. Since #52191 mutex_ IS
+    // write-preferring (folly::SharedMutexWritePriority), so that deadlock
+    // is live, not hypothetical: never call a locking accessor while
+    // holding the read lock -- use SizeUnsafe()/GetByOffsetUnsafe().
     size_t
     SizeUnsafe() const {
         return geometries_.size();
@@ -206,7 +208,11 @@ class SimpleGeometryCache {
     // ctx_ is declared first so it is destroyed last (after geometries_),
     // guaranteeing the Geometry destructors still see a live context.
     GEOSContextHandle_t ctx_{nullptr};  // Context owned by this cache
-    mutable std::shared_mutex mutex_;   // For read/write operations
+    // Write-preferring: a continuous stream of overlapping readers (every
+    // GIS expression holds AcquireReadLock() across a whole batch) must not
+    // starve the growing-segment writer. std::shared_mutex maps to glibc's
+    // reader-preferring pthread_rwlock_t on Linux, which does. See #52191.
+    mutable folly::SharedMutexWritePriority mutex_;
     std::vector<Geometry> geometries_;  // Direct storage of Geometry objects
 };
 
--- a/internal/core/src/common/GeometryCacheTest.cpp
+++ b/internal/core/src/common/GeometryCacheTest.cpp
@@ -11,6 +11,7 @@
 
 #include <gtest/gtest.h>
 #include <atomic>
+#include <chrono>
 #include <string>
 #include <thread>
 #include <vector>
@@ -416,6 +417,99 @@ TEST(GeometryCacheLifetime, OutOfRangeOffsetReturnsNullptrNotThrow) {
     EXPECT_NE(cache->GetByOffsetUnsafe(0), nullptr);
     EXPECT_EQ(cache->GetByOffsetUnsafe(1), nullptr);
     EXPECT_EQ(cache->GetByOffsetUnsafe(1000000), nullptr);
+}
+
+// Regression for issue #52191: the cache's shared_mutex must not let a
+// continuous stream of overlapping readers starve a writer.
+//
+// SimpleGeometryCache is read under AcquireReadLock() by every GIS expression
+// (GISFunctionFilterExpr / GISConjunctExpr hold it across a whole batch), while
+// the growing-segment insert path writes through AppendDataAt(). With a
+// reader-preferring rwlock -- which is what libstdc++'s std::shared_mutex maps
+// to on Linux, since glibc's pthread_rwlock_t defaults to
+// PTHREAD_RWLOCK_PREFER_READER_NP -- a writer only ever acquires during a
+// window in which the reader count drops to zero. Under sustained query load
+// that window may never occur, and the insert stalls indefinitely.
+//
+// The readers below deliberately overlap: each re-acquires immediately after
+// releasing, so with enough of them the reader count is essentially never zero.
+// Readers are stopped BEFORE joining the writer, so a starved writer makes this
+// test fail on the elapsed-time assertion instead of hanging forever.
+TEST(GeometryCacheConcurrency, WriterIsNotStarvedByOverlappingReaders) {
+    auto& mgr = SimpleGeometryCacheManager::Instance();
+    const int64_t seg_id = 900000011;
+    const FieldId field_id(31);
+    const std::string wkb = MakePointWkb(7.0, 7.0);
+
+    auto cache = mgr.GetOrCreateCache(kInstanceA, seg_id, field_id);
+    cache->AppendDataAt(0, wkb.data(), wkb.size());
+
+    constexpr int kReaders = 8;
+    // Generous: with a write-preferring lock the writer is admitted in about a
+    // millisecond, so this only trips on genuine starvation, not on a loaded
+    // CI machine.
+    constexpr auto kWriterBudget = std::chrono::seconds(5);
+
+    std::atomic<bool> stop{false};
+    std::atomic<int> readers_running{0};
+    std::vector<std::thread> readers;
+    readers.reserve(kReaders);
+    for (int i = 0; i < kReaders; ++i) {
+        readers.emplace_back([&]() {
+            readers_running.fetch_add(1);
+            while (!stop.load(std::memory_order_relaxed)) {
+                auto lock = cache->AcquireReadLock();
+                // Mirror the expression paths: real work is done while the
+                // read lock is held, so the lock is held for a while.
+                for (int k = 0; k < 64; ++k) {
+                    const Geometry* g = cache->GetByOffsetUnsafe(0);
+                    (void)g;
+                }
+            }
+        });
+    }
+
+    // Let the readers reach steady state so the reader count stays above zero.
+    while (readers_running.load() < kReaders) {
+        std::this_thread::yield();
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
+
+    std::atomic<bool> write_done{false};
+    auto started = std::chrono::steady_clock::now();
+    std::thread writer([&]() {
+        cache->AppendDataAt(1, wkb.data(), wkb.size());
+        write_done.store(true, std::memory_order_relaxed);
+    });
+
+    while (!write_done.load(std::memory_order_relaxed) &&
+           std::chrono::steady_clock::now() - started < kWriterBudget) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+    auto elapsed = std::chrono::steady_clock::now() - started;
+    bool acquired = write_done.load(std::memory_order_relaxed);
+
+    // Stop the readers first: once they drain, even a starved writer completes,
+    // so join() cannot hang regardless of the outcome asserted below.
+    stop.store(true, std::memory_order_relaxed);
+    writer.join();
+    for (auto& t : readers) {
+        t.join();
+    }
+
+    EXPECT_TRUE(acquired)
+        << "writer did not acquire the cache lock within "
+        << std::chrono::duration_cast<std::chrono::milliseconds>(kWriterBudget)
+               .count()
+        << " ms while " << kReaders
+        << " overlapping readers held the shared lock -- the writer is being "
+           "starved (issue #52191)";
+    EXPECT_LT(
+        std::chrono::duration_cast<std::chrono::milliseconds>(elapsed).count(),
+        std::chrono::duration_cast<std::chrono::milliseconds>(kWriterBudget)
+            .count());
+
+    mgr.RemoveSegmentCaches(kInstanceA, seg_id);
 }
 
 }  // namespace


internal/core/src/common/GeometryCache.h
```diff
@@ -206,7 +208,11 @@ class SimpleGeometryCache {
     // ctx_ is declared first so it is destroyed last (after geometries_),
     // guaranteeing the Geometry destructors still see a live context.
     GEOSContextHandle_t ctx_{nullptr};  // Context owned by this cache
-    mutable std::shared_mutex mutex_;   // For read/write operations
+    // Write-preferring: a continuous stream of overlapping readers (every
+    // GIS expression holds AcquireReadLock() across a whole batch) must not
+    // starve the growing-segment writer. std::shared_mutex maps to glibc's
+    // reader-preferring pthread_rwlock_t on Linux, which does. See #52191.
+    mutable folly::SharedMutexWritePriority mutex_;
     std::vector<Geometry> geometries_;  // Direct storage of Geometry objects
```
czs007, line 216: This lock exists only because of the cache's own layout, not because the data needs it -- and that is the thing to fix, rather than which side of the lock wins.

Column data in a growing segment is read lock-free: `ConcurrentVector` stores fixed-size chunks that are never relocated (`mmap/ChunkVector.h`, slab directory published via `std::atomic<ChunkImpl*>`), readers latch a `ChunkSnapshot` and then never touch a mutex, and row visibility is gated by the segment's acked row count. The `unique_lock` in `emplace_to_at_least` only guards directory growth.

`SimpleGeometryCache` needs a rwlock because it is `std::vector<Geometry>` + `resize()` in `AppendDataAt` (:130/:216): every relocation would dangle the `const Geometry*` a reader got from `GetByOffsetUnsafe()`, and the in-place overwrite at an absolute offset has no publish step, so an unlocked reader could see a half-constructed `Geometry`. That is why readers hold the shared lock across a whole batch, which is the asymmetry this PR works around.

Write-priority does not remove the problem, it moves it: during an insert batch there is one queued writer per row (`SegmentGrowingImpl.cpp:3447/3512`), and each one now blocks every newly arriving reader, so GIS query latency becomes a function of concurrent insert rate. Option 2 in #52191 (per-batch locking) only reduces the traffic.

Please rework the cache on the `ConcurrentVector` model instead: chunked storage that never relocates, entries published via a committed-row count (or simply visible after the segment's ack, which is when the row becomes readable anyway). Then the mutex goes away, and both the writer starvation and the reader stalls disappear together. The regression test from this PR is still useful and should carry over.

czs007 (changes requested):
Requesting changes: the direction here is not what we want for this cache. Switching the rwlock policy treats the symptom; the lock itself is the problem, and it should follow the growing-segment convention (`ConcurrentVector`) so there is no lock on the read path at all. Details inline.

internal/core/src/common/GeometryCache.h
xiaocai2333: Agreed, and reworked along those lines in a35e7cac79 — the rwlock is gone entirely, not re-tuned.

`GeometryChunkStore` now follows the growing segment's own column storage: fixed 4096-slot chunks allocated once and never moved, reached through a geometric slab directory whose slabs are never moved either, with every chunk pointer published by a release store. A slot's address is fixed for the cache's lifetime and its `Geometry` is final before the slot's state byte is released, so `GetByOffset()` is three acquire loads and takes nothing — and the pointer it hands back stays valid for as long as the caller holds its `shared_ptr` to the cache, which is strictly wider than the old read-lock scope. Writers take a mutex no reader ever touches, so the "one queued writer per row blocks every arriving reader" cost you called out does not exist either: the two paths no longer interact.

I duplicated the layout rather than reusing `ChunkVector.h` because it lives under `mmap/` and pulls in the mmap manager and the segcore config, and `common/` must not depend on either. The comment in `GeometryChunkStore` points at `ChunkSlabDirectory` for the shared slab arithmetic.

Two places where I deviated from your sketch, both deliberate:

**Per-slot publication instead of a committed-row count.** The cache is written at ABSOLUTE segment offsets that can arrive out of order — a later batch may land before an earlier one, and the add-field backfill writes `[filled, total)` — so there is no single prefix that is "the committed part". Each slot carries its own `kEmpty -> kReady` state byte instead. Readers treat "not published yet" exactly like an invalid row: nullptr, skip it, which is already the contract.

**`AppendDataAt` is now first-writer-wins.** Lock-free reads require that a published slot is never mutated, so an already-published slot is skipped rather than overwritten. That is sound because an absolute offset denotes one immutable row, so every write to it carries the same WKB — which is exactly what made the offset-addressed retry idempotent to begin with. I audited all four construction sites (insert at reserved offsets, load, add-field backfill, the sealed detached build); none writes different data to an offset it already published.

`AcquireReadLock()` / `GetByOffsetUnsafe()` / `SizeUnsafe()` are removed, and the five expression read sites call `GetByOffset()` directly.

Tests went 10 → 13. The starvation regression test is carried over as you asked; it now drives the public read API continuously, so it goes red again if anyone reintroduces a lock on the read path. New: `ReadersNeverObserveAPartiallyPublishedSlot` (readers race a 12k-row write spanning three chunks; every geometry they observe must be fully built), `PublishedSlotsAreFrozenAndStable` (pointer identity across first-writer-wins rewrites and across multi-chunk growth — the exact thing `resize()` used to break), and `SparseOffsetsAcrossChunkAndSlabBoundaries`.

13/13 pass, and 13/13 under TSAN with no new warning — the two races it reports are in `geos::io::CLocalizer`'s `setlocale` inside GEOS and reproduce identically on the previous commit.

Same verification caveat as before, stated plainly: those runs linked `GeometryCacheTest.cpp` against a pre-built `libmilvus_core.so` from another local worktree rather than a full clean core build, which was not feasible in this environment. `SimpleGeometryCache` is header-only so the new code is what ran, but the full build and the segcore/GIS suites are left to CI.

czs007:

Verified:
- `internal/core/src/common/GeometryCache.h:166-190` — `Reserve` fully allocates slabs and chunks before the release store publishes them, and a throwing allocation leaves the directory untouched so a retry is safe.
- `internal/core/src/common/GeometryCache.h:204-209` — `Publish` completes the move-assign before the release store on the state flag, and `Geometry::operator=(Geometry&&)` is `noexcept`, so the static_assert holds.
- `internal/core/src/common/GeometryCache.h:215-236` — `Get` performs slab, chunk, and state acquire loads in sequence and only reads `geometries[index]` once the state is `kReady`; published slots are never rewritten, so the raw pointer cannot be torn.
- `internal/core/src/common/GeometryCache.h:268-281` — slab index arithmetic is consistent between `SlabOf`, `PosOf`, and `SlabEntries`, and an offset of `~0` maps to slab 48 which is rejected by both `Get` and `Reserve` without overflow.
- `internal/core/src/common/GeometryCache.h:305-320` — the destructor clears the geometry store before `GEOS_finish_r`, so every published `Geometry` still has a live context at teardown, and the subsequent member `Clear()` is idempotent.
- `internal/core/src/segcore/SegmentGrowingImpl.cpp:445-448` and `ChunkedSegmentSealedImpl.cpp:7843-7844` — every `AppendDataAt` caller uses freshly reserved offsets or a detached cache, so first-writer-wins never discards real data; all five production readers hold the cache `shared_ptr` for the whole batch.

No findings survived verification, so there are no items to report in any severity section. There are no non-blocking, pre-existing, or disputed items.

czs007:

## Re-review of `a35e7cac7933`

This mechanism's previous comment on this commit reported no findings, so there are no earlier items to adjudicate. Everything below is new from this round.

---

Adversarial re-review of the lock-free geometry cache surfaced one behavioural regression in `AppendDataAt` plus two documentation/description accuracy problems.

## Medium

**`internal/core/src/common/GeometryCache.h:343` — first-writer-wins permanently freezes a transient parse failure**

The new `if (GeometryChunkStore::IsPublished(slot)) { return; }` replaces the old unconditional `geometries_[absolute_offset] = std::move(geometry);`, so a slot published with an invalid `Geometry()` (the `TryParseFromWkb` failure path at `:366`) can never be re-derived. Failure scenario: in a growing segment with `enableGeometryCache=true`, row *i*'s WKB hits a `bad_alloc` inside `GEOSWKBReader_read_r` (swallowed by the GEOS capi `execute()` wrapper, so `TryParseFromWkb` returns `false` and the offset is published invalid) and a later row in the same fill throws a retriable `MemAllocateFailed` from `GEOSWKBReader_create_r` → the batch retries over the same absolute offsets and `AppendDataAt` now returns early, leaving row *i* invalid for the segment's whole lifetime, silently excluded from every `ST_*` query result (and a silent false positive under negative predicates), where the pre-PR overwrite would have healed it.

The retry really does re-address the same offsets for GEOMETRY: the scalar non-sparse branch at `InsertRecord.h:1899-1902` omits `use_mapping_storage`, so it defaults to `false` (`ConcurrentVector.h:452`), `GrowingOffsetMapping::IsEnabled()` is permanently false, `filled = 0` at `SegmentGrowingImpl.cpp:3294-3295`, the `missing == 0` short-circuit at `:3304` never fires, and `BuildGeometryCacheForInsert` at `:3344` re-walks `[0, total_row_num)`.

Two qualifications bound the severity: the cache is off by default (`configs/milvus.yaml:599`), blast radius is a single row, and `Geometry.h:278-287` already documents this row class as *"classified as bad data with no retry"* — the old heal-on-retry was incidental rather than a stated contract.

Suggestion: introduce a third `kInvalid` state so an invalid slot may be rewritten. Readers return at `:206` on any non-`kReady` state without dereferencing `geometries[index]`, so a later writer can safely move into the slot and release-store `kReady`. The cheaper alternative — gating the skip on `IsPublished(slot) && slot-is-valid` — is *not* safe: `Publish`'s non-atomic move-assign at `:182` would race a reader that already passed the `kReady` check at `:206` and is evaluating `IsValid()` at `:209-210`. The `PublishedSlotsAreFrozenAndStable` test would need its null-rewrite leg split accordingly. *(surfaced during verification)*

## Low

**`internal/core/src/common/GeometryCache.h:317` — the invariant comment justifying the skip claims more than the code guarantees**

The new `FIRST WRITER WINS` block argues *"every write to it carries the same WKB — the retry above re-derives exactly the bytes that are already cached"*. That establishes only that the **input** is identical; the safety of skipping depends on the **stored outcome** being a function of that input, and `TryParseFromWkb` explicitly documents that it is not (`Geometry.h:278-287`: parse-time OOM is indistinguishable from unparseable WKB at that boundary, so the same input can yield an invalid result). Since this comment is the stated precondition for handing lock-free readers a bare pointer (`:372-374`), it should be corrected regardless of whether the behaviour above changes. Suggestion: state what the skip actually guarantees and name its cost — e.g. that a transient parse failure is frozen permanently, cross-referencing the `TryParseFromWkb` KNOWN LIMIT. *(surfaced during verification)*

**`internal/core/src/common/GeometryCache.h:341` — the PR description documents a different implementation than the one committed**

The PR body describes switching the cache mutex to `folly::SharedMutexWritePriority`, keeping `AcquireReadLock()` / `SizeUnsafe()` / `GetByOffsetUnsafe()`, and adding one test in which "8 reader threads re-acquire the shared lock". The committed code contains no `folly` reference at all: `AppendDataAt` uses `std::lock_guard<std::mutex> lock(write_mutex_)` (`:341`, with `write_mutex_` at `:417` marked writers-only), `AcquireReadLock` / `SizeUnsafe` / `GetByOffsetUnsafe` are deleted, and the read path is the lock-free `GeometryChunkStore` (`:89`, `GetByOffset` at `:386`) using a slab directory with 4096-slot chunks and per-slot release/acquire publication. Four tests were added, not one. The "Red/green verified locally" evidence in the body was produced against the replaced implementation and does not validate this diff. Since the body lands as the commit message and reviewers will check the code against it, please rewrite it to describe the lock-free store, and per the team's G3 requirement claim only benefits that were actually verified end to end against this diff. *(surfaced during verification)*

xiaocai2333:
Re [the re-review of `a35e7cac79`](https://github.com/milvus-io/milvus/pull/52783#issuecomment-5707610022) — all three items addressed in 8705d48354.

**Medium — first-writer-wins froze a transient parse failure.** Confirmed: GEOS swallows a `bad_alloc` inside `GEOSWKBReader_read_r` and returns nullptr, so an invalid outcome is not a pure function of the WKB, and freezing it lost the heal-on-retry the old unconditional overwrite gave.

Fixed slightly differently from the suggested third `kInvalid` state: `GeometryChunkStore::Publish` now stores and publishes only a *valid* geometry. A null or unparseable row leaves the slot `kEmpty`, so readers still return nullptr without touching it, and a later write of the same offset re-parses it. Only valid slots are frozen, which is sound because a successful parse *is* a pure function of the row's bytes. A separate `kInvalid` state would be treated exactly like `kEmpty` by both readers and writers, so it adds nothing. As you noted, the cheaper "skip only if published and valid" variant would race a reader that has already passed the `kReady` check; this approach never rewrites a `kReady` slot, so that race does not arise.

Tests: added `InvalidRowsAreRewrittenByARetry` (an unparseable and a null row are healed by a later write of the same offset, then frozen), and `ReadersNeverObserveAPartiallyPublishedSlot` now races readers against an unparseable pass followed by a healing pass. Both fail against `a35e7cac79` and pass with the fix; 14/14 locally, 14/14 under TSAN (only the pre-existing `geos::io::CLocalizer` setlocale race). The parse-time OOM itself was not fault-injected.

**Low — the FIRST WRITER WINS comment over-claimed.** Rewritten: it now says skipping is safe because a successful parse is a pure function of the bytes, explains why invalid outcomes are excluded, and cross-references the `TryParseFromWkb` KNOWN LIMIT, whose note now points back at the new behaviour.

**Low — PR description described the replaced implementation.** Rewritten to describe the lock-free store and the valid-only freezing rule. It now lists only what was verified against this diff and states explicitly what was not: there is no reader-latency measurement, and the OOM was not fault-injected.

czs007:

Verified:
- `internal/core/src/common/GeometryCache.h:189-207` — slabs are value-initialized and chunks fully constructed (every slot `kEmpty`) before their release stores, and readers reach them only through acquire loads.
- `internal/core/src/common/GeometryCache.h:229-238` — `Publish` moves the geometry into the slot before storing `kReady` via a `noexcept` move assignment, so a reader that observes `kReady` never sees a partially written or destroyed geometry.
- `internal/core/src/common/GeometryCache.h:296-310` — slab/position arithmetic joins consecutive slabs with no gap or overlap (slab 0 → chunks 0–15, slab 1 → 16–47, slab 2 → 48–111), and out-of-range offsets are rejected before directory indexing rather than overflowing.
- `internal/core/src/common/GeometryCache.h:365-391` — invalid parses stay unpublished so a retry can repair them, valid entries remain frozen once published, and a failed allocation throws before anything is published, leaving OOM retriable through the callers' existing `bad_alloc` handling.
- `internal/core/src/exec/expression/GISFunctionFilterExpr.cpp:57,124,186,607` and `internal/core/src/exec/expression/GISConjunctExpr.cpp:469-480` — every reader holds the cache `shared_ptr` across its whole loop, so cached geometry pointers stay valid for the duration of the query.
- `internal/core/src/common/GeometryCacheTest.cpp:431,512,575,623,665` — tests cover writer progress under continuous reads, concurrent publication across chunk and slab boundaries, pointer stability, invalid-row healing by retry, sparse offsets, and out-of-range reads.

czs007 (approved):
The rework addresses my earlier request: the read path no longer takes a lock. Approving.

I benchmarked the three versions of `GeometryCache.h` locally: the base `std::shared_mutex`, the first commit's `folly::SharedMutexWritePriority`, and this lock-free head (7e51a98). Setup: one cache with 1M points, 8192-row `intersects` refine batches, and one writer calling `AppendDataAt` per row, pinned to 11 physical cores.

Writer paced at 1000 rows / 100 ms (10k rows/s):

| readers | base rwlock | write-priority | lock-free |
|---|---|---|---|
| 1 | write batch 528–713 ms, falls behind (1.7–2.5k rows/s) | write batch 18–78 ms, read batch max 10 ms | write batch ≤ 0.65 ms, read batch max ≤ 1.3 ms |
| 4 | writer makes no progress (killed at 30 s) | write batch 49–152 ms, read p99 7–13 ms, max 26–43 ms | write batch ≤ 0.53 ms, read max ~1.2 ms |
| 8 | writer makes no progress | falls behind (5–6k rows/s), write batch 160–308 ms, read p99 ~20 ms, max 78–94 ms | write batch ≤ 0.52 ms, read p99 < 0.9 ms, max 3.8 ms |

Things the numbers show:
- #52191 is worse than a timing window. With the base lock and an unthrottled writer, one `AppendDataAt` call did not return for more than 600 s with 4 readers.
- Both lock-based versions also stall for about 130 ms when `std::vector` grows under the exclusive lock (1M rows, no readers). No reader can proceed during that time. The chunk store never relocates, so its worst-case append was 0.24 ms.
- Memory is on par or lower for 4096 rows or more, and peak equals steady state because nothing relocates. The one cost is a fixed ~70 KB (one 4096-slot chunk) per cache, which is a loss only below 4096 rows.
- Single-thread refine with a predicate is ~8% slower (83–88 vs 76–82 ns/row, 3/3 runs). There is no difference at 8 threads or for lookup-only. I have not tracked down the cause.

Non-blocking suggestions:
1. Please replace the "Not verified: a reader-latency benchmark" line in the description with numbers like these.
2. Follow-up, not for this PR: sealed caches are built detached with a known row count and are immutable once published, so they do not need the chunk store. Sizing them exactly, or at least sizing the last chunk, would remove the fixed 70 KB for small sealed segments. Growing segments are where the lock-free store pays off.

czs007 (approved):
/lgtm

PR #52787, opened 2026-08-24
fix: [2.6] validate encrypted index directory on read

XuanYang-cn:
## What changed

Backport only the encrypted V3 directory hardening independently applicable to the 2.6 reader:

- require persisted `slice_size` to be positive
- reject ciphertext slice offsets/sizes outside the file data region using overflow-safe arithmetic
- reject slice layouts that exceed or fail to cover the declared plaintext `original_size`
- resolve the cipher plugin after directory parsing, so malformed metadata fails before decryption setup

The unchanged V3 format and existing 2.6 writer contract are preserved. This PR does not backport streaming-load or transient-memory behavior from master.

pr: #51403
pr: #51405
pr: #50872 

## Verification

- `IndexEntryEncryptedV3Test.*:IndexEntryWriterV3Test.*`: 40/40 passed
- controlled C++ unittest build passed
- clang-format-12 and `git diff --check` passed


internal/core/src/storage/IndexEntryReader.cpp
```diff
@@ -377,20 +383,34 @@ IndexEntryReader::ReadFooterAndDirectory() {
         edek_ = dir_json["__edek__"].get<std::string>();
         ez_id_ = std::stoll(dir_json["__ez_id__"].get<std::string>());
         slice_size_ = dir_json["slice_size"].get<size_t>();
-
-        cipher_plugin_ = PluginLoader::GetInstance().getCipherPlugin();
-        AssertInfo(cipher_plugin_ != nullptr,
-                   "Cipher plugin required for encrypted V3 index");
+        AssertInfo(IsStreamSliceSizeAligned(slice_size_),
```
liliu-z: The backport takes only the reader half of a two-sided invariant: master asserts `IsStreamSliceSizeAligned(slice_size_)` in the `IndexEntryEncryptedLocalWriter` constructor as well, and that assert is absent on this branch (grep for `IsStreamSliceSizeAligned` on the branch hits only this reader line and the tests). The writer therefore still accepts an arbitrary `slice_size` and writes it into the directory table, while the reader now hard-rejects anything not 4 KiB aligned. There is no production trigger today — the only non-test construction site, `FileManager.h:235`, uses the default 16 MiB slice size — so this is a parity/defense-in-depth gap rather than a live wedge. Note that adding the ctor assert also requires bumping the existing encrypted tests that pass `slice_size = 512` (line 955) and `1024` (lines 1009, 1033, 1065, 1092) to aligned values.

internal/core/src/storage/IndexEntryWriterTest.cpp
```diff
@@ -1081,6 +1124,141 @@ TEST_F(IndexEntryEncryptedV3Test, EncryptedFdEntryMultiSlice) {
...
+        writer.Finish();
+    }
+
+    RewriteDirectory(file_path, [](nlohmann::json& dir) {
+        dir["slice_size"] = kStreamSliceAlignment + 1;
+    });
+
+    auto input = CreateInputStream(file_path);
+    int64_t file_size = GetFileSize(file_path);
+    EXPECT_THROW(IndexEntryReader::Open(input, file_size),
```
liliu-z: The three `Rejects*` tests give no regression coverage for the validation they are named after. No cipher plugin is ever registered with `PluginLoader` in this file (`mock_cipher_` is handed straight to the writer and never installed), so after `ReadFooterAndDirectory()` returns, `Open` unconditionally throws `SegcoreError("Cipher plugin required for encrypted V3 index")` at IndexEntryReader.cpp:275. Since each test only asserts `EXPECT_THROW(..., milvus::SegcoreError)`, deleting the entire directory-validation block from `ReadFooterAndDirectory` leaves all three green. Assert the specific validation message (the neighbouring `AcceptsValidDirectory` test already does exactly this with `find("Cipher plugin required")`), or register the mock plugin so the plugin assert stops firing.

internal/core/src/storage/IndexEntryReader.cpp
XuanYang-cn: Fixed in 0748aecda8: the encrypted writer now enforces kStreamSliceAlignment before resolving the encryptor, the encrypted tests use aligned slice sizes, and a regression test covers rejection of an unaligned writer slice size.

internal/core/src/storage/IndexEntryWriterTest.cpp
XuanYang-cn: Fixed in 0748aecda8: each corrupted-directory test now asserts its specific validation message (unaligned, out-of-range, or incomplete coverage). The fixture registers the mock cipher plugin, and the valid-directory case now verifies a successful read. The focused suite passes 11/11.

internal/core/src/storage/IndexEntryReader.cpp
```diff
@@ -377,20 +383,34 @@ IndexEntryReader::ReadFooterAndDirectory() {
...
         for (const auto& entry : dir_json["entries"]) {
             EntryMeta meta;
             meta.encrypted = true;
             meta.enc.original_size = entry["original_size"].get<uint64_t>();
             meta.enc.crc32 = Crc32cFromHex(entry["crc32"].get<std::string>());
+            size_t output_offset = 0;
             for (const auto& s : entry["slices"]) {
-                meta.enc.slices.push_back(
-                    {s["offset"].get<uint64_t>(), s["size"].get<uint64_t>()});
+                auto slice = SliceMeta{s["offset"].get<uint64_t>(),
```
buqian-zilliz, line 399: **[P2] Validate ciphertext ranges before accepting the directory**

The new coverage check only derives plaintext coverage from the number of slice records; it stores each persisted `offset` and ciphertext `size` without checking that the range lies inside the data region. A malformed non-meta entry can therefore use the expected slice count but advertise a multi-gigabyte size or an overflowing/out-of-file offset, allowing `Open()` to succeed. Every later encrypted read allocates `std::vector<uint8_t>(slice.size)` before `ReadAt`, so corrupt cold data can trigger excessive allocation or an invalid remote range request instead of a bounded format error. Validate each range using overflow-safe arithmetic such as `offset <= data_region_size` and `size <= data_region_size - offset` while parsing the directory.

internal/core/src/storage/IndexEntryReader.cpp
buqian-zilliz: **[P2] Preserve previously readable V3 slice sizes**

`slice_size` is persisted under the unchanged V3 format version, while the previous 2.6 writer accepted arbitrary non-zero custom slice sizes and the previous reader consumed them without an alignment requirement. This assertion makes an existing encrypted V3 file with a non-4-KiB slice size unreadable after upgrade, even though its ciphertext remains valid; the new writer-side assertion cannot repair already persisted files. The in-tree factory uses the aligned 16-MiB default, but the constructor exposes the custom value, so please either establish that such files cannot exist, retain legacy read compatibility, or version and migrate the new invariant.

internal/core/src/storage/IndexEntryReader.cpp
XuanYang-cn: Fixed in e4b52a8ec5. The reader now derives the ciphertext data-region size after the existing footer bounds check and validates every persisted slice with overflow-safe `offset <= data_region_size && size <= data_region_size - offset` arithmetic before storing it. Targeted regressions cover both `UINT64_MAX` offset and size; the final encrypted/writer V3 suite passes 42/42.

internal/core/src/storage/IndexEntryReader.cpp
XuanYang-cn: Fixed in e4b52a8ec5. The 2.6 reader now accepts any positive persisted slice size, preserving compatibility with files written before the new writer invariant; the writer still requires 4 KiB alignment for newly created files. A regression helper emits the legacy V3 layout independently of the current writer and verifies a multi-slice 1 KiB file remains readable.

XuanYang-cn: Follow-up in 1eaaa5ec0d: after re-evaluating 2.6 scope, the writer-side 4 KiB assertion was removed. The unchanged V3 format and prior 2.6 writer accepted arbitrary positive custom slice sizes, and 2.6 has no production streaming-load caller that establishes this new invariant. The reader retains positive-size, plaintext-coverage, and ciphertext-range validation; a 1 KiB writer roundtrip now proves the preserved contract.

congqixia (approved):
/lgtm

Issue #52795, opened 2026-08-24
enhance: avoid unnecessary scalar index materialization when retrieving fields whose index has no raw data


sijie-ni-0214:
## Problem

`ChunkedSegmentSealedImpl::bulk_subscript` (the `DataArray` overload used to fill search output fields and by GIS refine) pins the scalar index **before** checking whether the index carries raw data:

```cpp
if (!use_field_data) {
    auto scalar_indexes = PinIndex(op_ctx, field_id);   // lazily materializes the index
    if (!scalar_indexes.empty()) {
        if (IndexHasRawData(field_id)) {
            return ReverseDataFromIndex(...);
        }
    }
}
return get_raw_data(...);
```

`PinIndex` → `PinCells` triggers lazy scalar index materialization (index file read + deserialize + in-memory build). For index types that carry no raw data (`INVERTED`, `NGRAM`, `RTREE` — `ScalarIndexLoadResourceImpl` publishes `has_raw_data=false` for them), the freshly materialized index is then discarded and the value is read from the raw column instead.

Example: a vector search outputting an `int64` field (STL_SORT) and a `varchar` field (INVERTED) whose hits fall on 3 segments materializes the INVERTED index 3 times per retrieve — all wasted.

`IndexHasRawData` is a cheap published-state metadata lookup that does not depend on materialization; the vector branch right below already performs this check first, the scalar branch was a leftover from the pre-#48711 structure.

## Proposed fix

Move the `IndexHasRawData` check before `PinIndex`, mirroring the vector branch:

```cpp
if (!use_field_data && IndexHasRawData(field_id)) {
    auto scalar_indexes = PinIndex(op_ctx, field_id);
    if (!scalar_indexes.empty()) {
        return ReverseDataFromIndex(...);
    }
}
return get_raw_data(...);
```

## Additional notes

- Also avoids materializing `RTREE` indexes for geometry fields during GIS refine.
- Fixes inflated `search_storage_cost` accounting: previously the wasted materialization bytes were charged to the query.
- A regression test is included with the fix: asserts a scalar index with `has_raw_data=false` is never pinned (materialized) during `bulk_subscript`.

PR #52796, opened 2026-08-24
enhance: skip pinning scalar index without raw data during retrieve

sijie-ni-0214:
issue: #52795

## What

In `ChunkedSegmentSealedImpl::bulk_subscript` (the `DataArray` overload serving search output fields, GIS refine and rescores), check `IndexHasRawData(field_id)` **before** `PinIndex` instead of after, so scalar indexes that carry no raw data (`INVERTED` / `RTREE`) are no longer lazily materialized just to be discarded in favor of the raw column. (NGRAM was already skipped by `PinIndex`'s own `ngram_fields` guard, so it is not affected by this change.)

## Why

`PinIndex` → `PinCells` triggers lazy scalar index materialization (index file read + deserialize + in-memory build). For indexes with `has_raw_data=false` the materialized index is never used — the value is read from the raw column via `get_raw_data`. This is pure waste of CPU/IO/memory per retrieve, and it also inflated `search_storage_cost` accounting by charging the wasted materialization bytes to the query. The vector branch right below already performs this check first; the scalar branch was a leftover from the pre-#48711 structure (#35764 → #48711 → #49127), so this aligns the two branches.

## Verification

- New regression test `Sealed.BulkSubscriptSkipsPinIndexWhenIndexHasNoRawData`: loads a scalar index reporting `has_raw_data=false` and asserts `bulk_subscript` returns the raw-column values without ever pinning (materializing) the index. Verified the test **fails without the fix** and **passes with it**.
- Related tests pass: `Sealed.LoadArrayFieldData`, `Sealed.LoadArrayFieldDataWhenIndexHasRawData`, `CApiTest.RetrieveScalarFieldFromColumnWhenIndexHasRawDataAndPreferFieldDataEnabled`.
- CI ut-cpp: 8762 passed / 0 failed; diff coverage 100% (3/3 measurable lines).
- `git diff --check` and clang-format-15 `--dry-run --Werror` both clean.

Known gaps:

- The `IndexHasRawData == true → ReverseDataFromIndex` direction is not pinned by a discriminating test in this PR (existing tests either keep index data identical to column data, or cover the `prefer_field_data=true` path).
- `Sealed.LoadScalarIndex` (full search-output path with STL_SORT) cannot complete on my macOS machine due to a pre-existing knowhere crash (`IndexNode::AddEmbListIfNeed`), which also occurs without this change and is unrelated.

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -6109,18 +6109,14 @@ ChunkedSegmentSealedImpl::bulk_subscript(milvus::OpContext* op_ctx,
 
     if (!IsVectorDataType(field_meta.get_data_type())) {
         // === Scalar field ===
-        if (!use_field_data) {
+        if (!use_field_data && IndexHasRawData(field_id)) {
             // Try index first: if scalar index exists and has raw data, read from index
             PinWrapper<const index::IndexBase*> pin_scalar_index_ptr;
             auto scalar_indexes = PinIndex(op_ctx, field_id);
             if (!scalar_indexes.empty()) {
                 pin_scalar_index_ptr = std::move(scalar_indexes[0]);
-                if (IndexHasRawData(field_id)) {
-                    return ReverseDataFromIndex(pin_scalar_index_ptr.get(),
-                                                seg_offsets,
-                                                count,
-                                                field_meta);
-                }
+                return ReverseDataFromIndex(
+                    pin_scalar_index_ptr.get(), seg_offsets, count, field_meta);
             }
         }
         return get_raw_data(op_ctx, field_id, field_meta, seg_offsets, count);
--- a/internal/core/unittest/test_sealed.cpp
+++ b/internal/core/unittest/test_sealed.cpp
@@ -1330,6 +1330,56 @@ TEST(Sealed, LoadScalarIndex) {
     std::cout << json.dump(1);
 }
 
+TEST(Sealed, BulkSubscriptSkipsPinIndexWhenIndexHasNoRawData) {
+    size_t N = ROW_COUNT;
+    auto schema = std::make_shared<Schema>();
+    auto counter_id = schema->AddDebugField("counter", DataType::INT64);
+    auto double_id = schema->AddDebugField("double", DataType::DOUBLE);
+    schema->set_primary_field_id(counter_id);
+
+    auto dataset = DataGen(schema, N);
+    auto double_col = dataset.get_col<double>(double_id);
+
+    auto segment = CreateSealedWithFieldDataLoaded(schema, dataset);
+
+    // Load a scalar index whose metadata reports no raw data (as INVERTED
+    // does). TestIndexTranslator records the OpContext when the index is
+    // actually materialized, so we can assert bulk_subscript never pins it.
+    auto index = GenScalarIndexing<double>(N, double_col.data());
+    milvus::OpContext* observed_ctx = nullptr;
+    LoadIndexInfo load_info;
+    load_info.field_id = double_id.get();
+    load_info.field_type = DataType::DOUBLE;
+    load_info.index_params = GenIndexParams(index.get());
+    load_info.load_resource_request.emplace();
+    load_info.load_resource_request->has_raw_data = false;
+    load_info.cache_index =
+        CreateTestCacheIndex("test", std::move(index), &observed_ctx);
+    segment->LoadIndex(load_info);
+    ASSERT_TRUE(segment->HasIndex(double_id));
+
+    std::vector<int64_t> seg_offsets(N);
+    for (int64_t i = 0; i < static_cast<int64_t>(N); i++) {
+        seg_offsets[i] = i;
+    }
+
+    milvus::OpContext op_ctx;
+    auto field_data =
+        segment->bulk_subscript(&op_ctx, double_id, seg_offsets.data(), N);
+
+    // Values must be read from the raw column, unchanged.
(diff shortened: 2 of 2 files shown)


czs007:

**Verified:**
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6112` — hoisting `IndexHasRawData(field_id)` above `PinIndex` is semantically equivalent, since the predicate is a pure lookup into the published `index_has_raw_data` state and `PinIndex` never mutates it.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:1470` — `index_has_raw_data` is populated at load/publish time (`:1315-1368`, `:8657-8686`), so the flag is already known before any retrieve runs; there is no ordering hazard where it would be unknown until materialization.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6116` — the `scalar_indexes.empty()` guard is preserved, so "state says index-has-raw-data but `PinIndex` returns empty" still falls through to `get_raw_data` exactly as before.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6122` — for indexes without raw data the pre-change code already fell through to the identical `get_raw_data` call, so the raw column's availability requirement is unchanged and no new fallback failure mode is introduced.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:235` — the scalar branch now evaluates `IndexHasRawData` for every scalar field including un-indexed ones, and `get_bit_if_present` guards both `pos < 0` and `pos >= bitset.size()`, so this cannot assert or throw.
- `internal/core/unittest/test_sealed.cpp:1335` — the new test asserts `observed_ctx == nullptr` with warmup disabled (`test_utils/cachinglayer_test_utils.h:210`) and `prefer_field_data_when_index_has_raw_data_` defaulting to `false`, so it genuinely reaches the changed branch and would fail under the pre-fix ordering.

## Low — non-blocking

**`internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6112` — commit message and PR body over-claim `NGRAM` as a beneficiary.**
Both the commit message (dd4a321fea: "for indexes that carry no raw data (INVERTED, NGRAM, RTREE) the materialized index is discarded") and the PR description list `NGRAM`, but `NGRAM` was never lazily materialized on this path even before the change. `bulk_subscript` calls `PinIndex` with the default `include_ngram = false`, and `PinIndex` returns `{}` for any field in `runtime.ngram_fields` (`ChunkedSegmentSealedImpl.h:150`) before it ever reaches `scalar_indexings`. Non-JSON `NGRAM` indexes are inserted into `ngram_fields` at `ChunkedSegmentSealedImpl.cpp:875`, and JSON `NGRAM` indexes land in `ngram_indexings` (`:820`) and return early, so they never enter `scalar_indexings` at all. The types genuinely affected are `INVERTED` (`InvertedIndexTantivy.h:209`, `HasRawData() == false`) and `RTREE` (`RTreeIndex.h:182`).
*Suggestion:* drop `NGRAM` from both the commit message and the PR description, or restate it as "ngram was already skipped by `PinIndex`'s own ngram guard; this change aligns the remaining types." Otherwise the permanent record attributes the saving to the wrong index type and misleads anyone later investigating ngram retrieve cost. *(raised by tinswzy)*

**`internal/core/unittest/test_sealed.cpp:1986` — a cited verification test cannot distinguish the two orderings, and the `IndexHasRawData() == true` branch remains unpinned.**
`Sealed.LoadArrayFieldDataWhenIndexHasRawData` is listed in the PR's Verification section and sets `request.has_raw_data = true`, but the field is `ARRAY`, and `IndexFactory::CanUseIndexRawDataForField` (`IndexFactory.cpp:331-332`) forces `has_raw_data` back to `false` for `ARRAY`/`JSON`/`VECTOR_ARRAY`. The published state therefore records `false`, and the test takes the `get_raw_data` path both before and after this change — it has no discriminating power over the new gate. The other cited test, `CApiTest.RetrieveScalarFieldFromColumnWhenIndexHasRawDataAndPreferFieldDataEnabled`, only covers `prefer_field_data = true` (`segment_c_test.cpp:1842`) and also goes through `get_raw_data`. A search across `internal/core/unittest` and `segment_c_test.cpp` finds nothing asserting that a scalar index which *does* carry raw data still serves `bulk_subscript` via `ReverseDataFromIndex`; the new `BulkSubscriptSkipsPinIndexWhenIndexHasNoRawData` pins only the `false` branch. The `true` branch is worth covering because in the production configuration the loader skips loading the raw column precisely when the index carries raw data (`internal/querynodev2/segments/segment_loader.go:2003-2006`), making the index the only data source.
*Suggestion:* add the mirror test using the technique already in `segment_c_test.cpp:1806-1815` — build the scalar index over values that differ from the column (e.g. `+1`) on a non-`ARRAY`/`JSON` type, load it with `load_resource_request->has_raw_data = true`, and assert `bulk_subscript` returns the index-side values. Also correct the PR description's verification claims about the two cited tests. *(raised by tinswzy)*

sijie-ni-0214:
@czs007 Thanks for the review (and @tinswzy for the detailed analysis)!

Both points are valid:

1. **NGRAM over-claim** — agreed: `PinIndex`'s `ngram_fields` guard already returns `{}` for NGRAM fields, so NGRAM was never materialized on this path. I've updated the PR description to list only `INVERTED` / `RTREE` as affected.
2. **Cited tests have no discriminating power** — correct: `LoadArrayFieldDataWhenIndexHasRawData` has `CanUseIndexRawDataForField` force `has_raw_data=false` for ARRAY, and the C API test covers the `prefer_field_data=true` path; neither distinguishes the new gate. I've corrected the Verification section accordingly.

Issue #52797, opened 2026-08-24
[Bug]: master fails to compile on macOS after #51794 (std::min type mismatch in ConcurrentVector.cpp)


sijie-ni-0214:
## Summary

After PR #51794 (`feat: support recursively nested Array fields`, commit 55e718f380) merged into master, `milvus_segcore` no longer compiles on macOS.

## Error

In `internal/core/src/segcore/ConcurrentVector.cpp`, `ConcurrentVector<ArrayValue>::set_mmap_proto_rows` calls `std::min` with mixed `int64_t` and `ssize_t` operands:

```cpp
const auto remaining_in_chunk = size_per_chunk_ - chunk_offset;  // int64_t
const auto remaining_rows = static_cast<ssize_t>(...);           // ssize_t
const auto copy_count = std::min(remaining_in_chunk, remaining_rows);
```

On Linux both typedefs resolve to `long`, so template argument deduction succeeds. On macOS, `int64_t` is `long long` while `ssize_t` is `long`, so the compiler rejects the call:

```
error: no matching function for call to 'min'
note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('int64_t' (aka 'long long') vs. 'ssize_t' (aka 'long'))
```

## Why CI missed it

PR #51794's CI only ran on Linux. The macOS check workflow (`.github/workflows/mac.yaml`) is manually triggered and does not run on pull requests, so master has been in a broken-on-macOS state since 2026-08-20.

## Environment

- Milvus master at f192bafbd5 (includes 55e718f380)
- macOS (arm64), Apple clang / LLVM 15+
- Reproducible with a minimal reproducer; full `make` fails at `milvus_segcore`'s `ConcurrentVector.cpp.o`.

## Expected behavior

Master compiles on macOS.

PR #52798, opened 2026-08-24
fix: compile nested-array mmap path on macOS

sijie-ni-0214:
issue: #52797

## What this PR does

#51794 (recursively nested Array fields) calls `std::min` with mixed `int64_t` and `ssize_t` operands in `ConcurrentVector<ArrayValue>::set_mmap_proto_rows`. On Linux both typedefs resolve to `long`, so template argument deduction succeeds; on macOS `int64_t` is `long long` while `ssize_t` is `long`, so `std::min` fails to deduce a common type and `milvus_segcore` no longer compiles on macOS (broken since 2026-08-20).

This PR normalizes `remaining_in_chunk` to `ssize_t` so both operands share a type. Both types are signed 64-bit and the value cannot exceed `size_per_chunk_`, so the cast preserves the value.

## Verification

- Full `make` on macOS (arm64, Apple clang): passes with 0 errors; `ConcurrentVector.cpp.o` compiles cleanly.
- Root cause reproduced with a minimal reproducer against the pre-fix code (libc++: `deduced conflicting types for parameter '_Tp' ('int64_t' (aka 'long long') vs. 'ssize_t' (aka 'long'))`); fixed version compiles and runs.
- clang-format 15: no diff.

Related: pr #51794 (introduced the regression).

### DIFF
--- a/internal/core/src/segcore/ConcurrentVector.cpp
+++ b/internal/core/src/segcore/ConcurrentVector.cpp
@@ -254,7 +254,8 @@ ConcurrentVector<ArrayValue>::set_mmap_proto_rows(
     while (source_offset < rows.size()) {
         const auto chunk_id = current_offset / size_per_chunk_;
         const auto chunk_offset = current_offset % size_per_chunk_;
-        const auto remaining_in_chunk = size_per_chunk_ - chunk_offset;
+        const auto remaining_in_chunk =
+            static_cast<ssize_t>(size_per_chunk_ - chunk_offset);
         const auto remaining_rows =
             static_cast<ssize_t>(rows.size() - source_offset);
         const auto copy_count = std::min(remaining_in_chunk, remaining_rows);


czs007:

**Verified:**

- `internal/core/src/segcore/ConcurrentVector.cpp:257-260` — casting `size_per_chunk_ - chunk_offset` to `ssize_t` makes both `std::min` operands the same type, so template argument deduction succeeds on macOS libc++ where `int64_t` (`long long`) and `ssize_t` (`long`) are distinct types.
- `internal/core/src/segcore/ConcurrentVector.cpp:257` — the cast cannot truncate: `chunk_offset = current_offset % size_per_chunk_` is in `[0, size_per_chunk_)` (with `size_per_chunk_ > 0` asserted at `ConcurrentVector.h:60`), so the difference is in `[1, size_per_chunk_]` and fits a 64-bit signed type on all supported build targets.
- `internal/core/src/segcore/ConcurrentVector.cpp:264-267` — `copy_count`'s new `ssize_t` type is consumed by `rows.subspan(source_offset, static_cast<size_t>(copy_count))`, `current_offset += copy_count`, and `source_offset += static_cast<size_t>(copy_count)`; each site either casts explicitly or targets `ssize_t`, so semantics are unchanged and the fix is a no-op on Linux.
- `internal/core/src/segcore/ConcurrentVector.cpp:254-268` — loop progress is preserved: both operands are `≥ 1` (`remaining_rows ≥ 1` follows from the `source_offset < rows.size()` condition), so `copy_count ≥ 1` and `source_offset` strictly increases each iteration.
- `internal/core/src/segcore/ConcurrentVector.cpp:241` — `upper_div(end_offset, size_per_chunk_)` also mixes `ssize_t` and `int64_t`, but `common/Utils.h:187` declares `upper_div(int64_t, int64_t)` as a non-template function, so it resolves via implicit conversion and is not a second deduction failure.
- Fix completeness — grepping the added lines of `55e718f380` (#51794) for `std::min`/`std::max` finds only `ConcurrentVector.cpp:260`, so there is no second site of this class left unfixed.
- Change scope — the edit is confined to one local variable declaration; the function signature, class members, and all call sites are untouched.

PR #52801, opened 2026-08-24
enhance: [cp3.0] skip pinning scalar index without raw data during retrieve

sijie-ni-0214:
Cherry-pick from master
pr: #52796

issue: #52795

## What

In `ChunkedSegmentSealedImpl::bulk_subscript` (the `DataArray` overload serving search output fields, GIS refine and rescores), check `IndexHasRawData(field_id)` **before** `PinIndex` instead of after, so scalar indexes that carry no raw data (`INVERTED` / `NGRAM` / `RTREE`) are no longer lazily materialized just to be discarded in favor of the raw column.

## Why

`PinIndex` → `PinCells` triggers lazy scalar index materialization (index file read + deserialize + in-memory build). For indexes with `has_raw_data=false` the materialized index is never used — the value is read from the raw column via `get_raw_data`. This is pure waste of CPU/IO/memory per retrieve, and it also inflated `search_storage_cost` accounting by charging the wasted materialization bytes to the query. The vector branch right below already performs this check first; this aligns the two branches.

## Test Plan

- New regression test `Sealed.BulkSubscriptSkipsPinIndexWhenIndexHasNoRawData`: loads a scalar index reporting `has_raw_data=false` and asserts `bulk_subscript` returns the raw-column values without ever pinning (materializing) the index. Verified the test **fails without the fix** and **passes with it** (on master).
- Ran on the 3.0 worktree after `make build-cpp-with-unittest`:
  `./internal/core/output/unittest/all_tests --gtest_filter='Sealed.BulkSubscriptSkipsPinIndexWhenIndexHasNoRawData:Sealed.LoadArrayFieldData:Sealed.LoadArrayFieldDataWhenIndexHasRawData'`
  → **3/3 passed**.
- `git diff --check` clean.

### DIFF
--- a/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
+++ b/internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp
@@ -6205,18 +6205,14 @@ ChunkedSegmentSealedImpl::bulk_subscript(milvus::OpContext* op_ctx,
 
     if (!IsVectorDataType(field_meta.get_data_type())) {
         // === Scalar field ===
-        if (!use_field_data) {
+        if (!use_field_data && IndexHasRawData(field_id)) {
             // Try index first: if scalar index exists and has raw data, read from index
             PinWrapper<const index::IndexBase*> pin_scalar_index_ptr;
             auto scalar_indexes = PinIndex(op_ctx, field_id);
             if (!scalar_indexes.empty()) {
                 pin_scalar_index_ptr = std::move(scalar_indexes[0]);
-                if (IndexHasRawData(field_id)) {
-                    return ReverseDataFromIndex(pin_scalar_index_ptr.get(),
-                                                seg_offsets,
-                                                count,
-                                                field_meta);
-                }
+                return ReverseDataFromIndex(
+                    pin_scalar_index_ptr.get(), seg_offsets, count, field_meta);
             }
         }
         return get_raw_data(op_ctx, field_id, field_meta, seg_offsets, count);
--- a/internal/core/unittest/test_sealed.cpp
+++ b/internal/core/unittest/test_sealed.cpp
@@ -1328,6 +1328,56 @@ TEST(Sealed, LoadScalarIndex) {
     std::cout << json.dump(1);
 }
 
+TEST(Sealed, BulkSubscriptSkipsPinIndexWhenIndexHasNoRawData) {
+    size_t N = ROW_COUNT;
+    auto schema = std::make_shared<Schema>();
+    auto counter_id = schema->AddDebugField("counter", DataType::INT64);
+    auto double_id = schema->AddDebugField("double", DataType::DOUBLE);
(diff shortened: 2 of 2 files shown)


czs007:

**Verified:**
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6208` — the new `!use_field_data && IndexHasRawData(field_id)` guard evaluates the same predicate the old code checked after pinning; every combination reaches the same terminal call (`ReverseDataFromIndex` or `get_raw_data`).
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:1683` — `IndexHasRawDataFromState` reads only the published-state bitsets (`index_ready_bitset` / `binlog_index_bitset`), which are set at load time from `load_resource_request.has_raw_data`, so the flag's value is independent of whether the index was pinned.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:920` — `LoadScalarIndex` publishes the runtime and the raw-data flag together via `PublishIndexReadyLocked`, so the reorder neither introduces nor widens the existing one-publish-wide read window, and adds no extra `CapturePublishedState()` on the common path.
- `internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6227` — the vector branch already evaluated the same predicate before any pin; the scalar branch now matches it.
- `internal/core/src/segcore/Utils.cpp:1080` — `ReverseDataFromIndex` performs an unchecked `dynamic_cast`; the new gate is strictly more conservative about reaching it, so exposure only shrinks.
- `internal/core/unittest/test_sealed.cpp:1348` — the regression is discriminating, not vacuous: with `CacheWarmupPolicy_Disable` and `has_raw_data = false`, the old code would materialize the index and set `observed_ctx`, so the `observed_ctx == nullptr` assertion fails pre-fix and passes post-fix.

## Low — non-blocking

- **`internal/core/src/segcore/ChunkedSegmentSealedImpl.cpp:6208` — the commit message and PR body list the affected index types inaccurately: NGRAM is included but was never materialized on this path, while BITMAP and HYBRID are omitted despite being real beneficiaries.** `bulk_subscript` calls `PinIndex` with the default `include_ngram = false`, and `PinIndex` returns `{}` for any field in `runtime.ngram_fields` before touching `scalar_indexings` (`ChunkedSegmentSealedImpl.h:152-155`); `LoadScalarIndex` inserts every NGRAM-typed scalar index into `ngram_fields` (`ChunkedSegmentSealedImpl.cpp:876-880`), so NGRAM fields fell through to `get_raw_data` both before and after this change. Conversely, BITMAP (`IndexFactory.cpp:817`) and HYBRID (`IndexFactory.cpp:823`) also publish `has_raw_data = false` and were being materialized then discarded under the old ordering. No behavioural impact — the code itself is correct. Suggestion: update the commit message and PR body to state the affected set as INVERTED, RTREE, BITMAP, HYBRID, since that list is what maintainers will grep against when this lands in 3.0.1. *(raised by tinswzy)*

PR #52802, opened 2026-08-24
fix: keep compaction task metrics in sync across retries

XuanYang-cn:
issue: #52803

Repeated compaction retries could apply `executing-- / pending++` while the task was already pending, leaving stale pending values or negative executing gauges. This change accounts for transitions between the last successfully persisted task state and the next persisted state under the existing `compactionTaskMeta` lock. Saving the same metric key again has no effect, and failed persistence leaves metrics unchanged.

- Share state-to-metric mapping across Mix, Sort, L0, Clustering, and BumpSchemaVersion compaction. Compare the complete node/type/status key and preserve terminal `done` counts during cleanup.
- Capture the initial metric key before enqueue and count it only after admission succeeds. Account restored tasks before publishing them to the scheduler.
- Keep scheduler candidates pending while waiting for execution-pool capacity. After acquiring the existing task-ID lock, callbacks recheck membership before dequeuing or querying. Abort waits for in-flight callbacks, and callbacks queued before removal skip canceled tasks. Candidate order comes from a queue-owned snapshot using the same heap comparator as `Pop()`.
- On channel removal, detach tasks from the inspector, release `executingGuard`, then wait for scheduler Abort and deduct the final persisted pending/executing metric. Queue locks are released before worker RPCs.
- Log and wrap clustering Create failures. If saving retry state also fails, return both causes from `doCompact`.

### Validation

- Full DataCoord component tests, including race detection and coverage, passed after merging current `upstream/master`.
- Guarded `golangci-lint` for `./internal/datacoord/...`: passed, 0 issues.
- Regression coverage includes repeated Sort retries, concurrent completion saves, failed persistence, node changes, admission/restore/removal, L0 fast finish, and transitions for all five compaction types.
- Scheduler tests cover cancellation during dispatch/retry/RPC, saturated and closed execution pools, queued queries, no available slot, and queue-defined priority. Snapshot tests verify `Pop()` order and preservation of the original queue.

### Scope and limits

Gauge updates are separate operations, so a scrape can still observe an intermediate value. Existing task-ID locks continue to span worker RPCs; canceling an in-flight task may wait for them. Inspector business-state-machine concurrency is outside this change. Channel-removal accounting tests use a mock scheduler; scheduler cancellation is covered separately. Local validation is scoped to DataCoord, with full-repository lint and e2e left to CI.


internal/datacoord/compaction_task.go
```diff
@@ -49,6 +52,60 @@ type CompactionTask interface {
...
+		datapb.CompactionTaskState_failed,
+		datapb.CompactionTaskState_timeout,
+		datapb.CompactionTaskState_cleaned:
+		metric.status = metrics.Done
+	}
+
+	return metric
+}
+
+func updateCompactionTaskMetrics(oldTask, newTask *datapb.CompactionTask) {
```
liliu-z, line 87: mix/sort task 处于 meta_saved 时，scheduler 的 checkPool 正在 QueryTaskOnWorker 内（compaction_task_mix.go:239 写完 meta_saved、尚未走到 :149 的 processMetaSaved()），inspector 的 checkCompaction 恰好采样到该 task 并进入 Process() 的 case meta_saved（compaction_task_mix.go:254-255）。两边读到相同的 oldTask=(node, Executing)，各自成功持久化 completed，各发一次 Executing(node) −1 / Done(node) +1。后果：milvus_datacoord_compaction_task_num{status="executing"} 对该 node label 永久 −1（每命中一次再 −1），datacoord 不重启不自愈。窗口 ≈ 一次 etcd 写，dataCoord.compaction.scheduleInterval 默认 500ms（component_param.go:6144），概率低但长期累积。这正是本 PR 声称消除的失败模式，只是路径更窄。

internal/datacoord/compaction_inspector.go
```diff
@@ -512,7 +509,7 @@ func (c *compactionInspector) removeTasksByChannel(channel string) {
 				mlog.Int64("node", task.GetTaskProto().GetNodeID()),
 			)
 			delete(c.executingTasks, id)
-			metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", task.GetTaskProto().GetNodeID()), task.GetTaskProto().GetType().String(), metrics.Executing).Dec()
+			decCompactionTaskMetric(task.GetTaskProto())
```
liliu-z, line 513: 子情形 A：DropVirtualChannel（internal/datacoord/services.go:837）时某 task 状态为 executing 且在 c.executingTasks 中 → :512 发 Executing(N) −1。但函数内没有 AbortAndRemoveTask（全仓唯一调用点是 stats_inspector.go:546），task 仍在 scheduler 的 runningTasks 里，下一轮 check() 的 QueryCompaction 失败 → compaction_task_mix.go:123 写回 pipelining → 再发一次 Executing(N) −1 + Pending(-1) +1。Executing(N) 落到 −1 并永久偏移。子情形 B：task 已是 completed/failed/timeout 但还没被 checkCompaction 摘出 executingTasks（窗口最长 scheduleInterval，默认 500ms），decCompactionTaskMetric 按当前 state 推出的是 Done 桶。正常完成的 task 的 Done 永不递减（cleanFailedTasks / DropCompactionTask 都不碰指标，completed→cleaned 被短路），只有这条路会减，导致 Done 少计。已核实它不会为负——每次 Done −1 都有配对的 +1（终态转换或 restoreTask 在 :532 的 inc）。

internal/datacoord/compaction_task_clustering.go
```diff
@@ -811,6 +810,7 @@ func (t *clusteringCompactionTask) updateAndSaveTaskMeta(opts ...compactionTaskO
 		mlog.Warn(context.TODO(), "Failed to saveTaskMeta", mlog.Err(err))
 		return merr.WrapErrClusteringCompactionMetaError("updateAndSaveTaskMeta", err) // retryable
 	}
+	updateCompactionTaskMetrics(oldTask, task)
```
liliu-z, line 819: DataNode 因 slot 不足或 "compaction already exists" 拒收 clustering compaction 时，只要写回 pipelining 成功，就会落穿到 785 行再次持久化成 executing + nodeID，且 retryOnError 收到的是 nil。后续追了两跳：scheduler 把它当 InProgress 放进 runningTasks（task/global_scheduler.go:300）→ 下一轮 QueryTaskOnWorker → DataNode 无此 plan、响应不带 state（internal/datanode/services.go:946）→ session/cluster.go:233-236 的 GetTaskState() 报错 → 返回 (nil, err) → clustering 写回 pipelining 重试。所以它会自愈，代价是浪费一个 check 周期；真正的残留损害是状态为 InProgress 导致 global_scheduler.go:301 的 s.recordTaskFailure(task) 被绕过，DataNode 持续 slot 不足时 clustering 会以 check 间隔空转、不进退避。指标上正负抵消，不漂。属既有缺陷，本 PR 恰好编辑了这个 block。

internal/datacoord/compaction_task.go
XuanYang-cn: Fixed: serialized persisted state transitions per compaction task, so concurrent meta_saved -> completed attempts observe the updated state and apply the Gauge transition once. Added a concurrent regression test and verified it with -race.

internal/datacoord/compaction_inspector.go
XuanYang-cn: Fixed: removeTasksByChannel now calls AbortAndRemoveTask while holding executingGuard before deleting the task and decrementing its current metric, so in-flight scheduler Create/Query work is drained before removal. Added a regression assertion.

internal/datacoord/compaction_task_clustering.go
XuanYang-cn: Fixed: doCompact now returns the Create error after persisting pipelining + NullNodeID, preventing fallthrough to executing and allowing retryOnError plus scheduler backoff. Added a regression test for worker refusal.

congqixia:
compaction_inspector.go / removeTasksByChannel — dec on terminal-state tasks still undercounts done
decCompactionTaskMetric(task.GetTaskProto()) derives the bucket from the task's current persisted state. For a task that has already been persisted into a terminal state (completed/failed/timeout) but has not yet been removed from executingTasks by checkCompaction — a window of up to one compaction.scheduleInterval (default 500ms) — this call decrements the done bucket.
The problem: done is never decremented on any other path. cleanFailedTasks/DropCompactionTask don't touch the metric, and the completed → cleaned transition is short-circuited by the idempotency check in updateCompactionTaskMetrics. So a task that completes normally stays at done +1 forever, while an identical task that happens to be hit by removeTasksByChannel during that window nets back to done +0. Two equivalent completed tasks therefore report different gauge values, and done drifts low on exactly the removal path we just made consistent for executing.
AbortAndRemoveTask (added in 8ff70e1) fixes sub-case A — the double executing −1 from a later re-persist to pipelining — but this sub-case B remains untouched.
Suggestion: guard the decrement so a terminal-state task is not subtracted again, e.g. only decCompactionTaskMetric when state is still non-terminal (pipelining/executing/meta_saved/…), or otherwise define and document that done is not meant to be monotonic and drop the decrement for terminal states here.

XuanYang-cn:
@congqixia Fixed in 18c1e3cb04: channel removal now preserves tasks whose persisted state maps to the done bucket, while pending/executing removals still decrement their active bucket. Regression coverage includes completed, failed, and timeout states; the focused test and test-datacoord both pass.

XuanYang-cn:
Fixed the current Code-Check failure in 5d906fabf8 by merging master, including #53243. The failure was caused by two stale proxy test references introduced upstream (`newTaskScheduler` and `importTask.metaCache`); both reproduced locally before the merge.

Validation after the merge:
- `TestProxy_ImportV2` and `TestImportTaskSuite`: passed.
- Compaction metric retry/concurrent-completion regressions, `TestCompactionPlanHandlerSuite` (including terminal-state channel removal), and `TestClusteringCompactionTaskSuite`: passed after starting the local dependencies.
- Four lint groups (root Go sources, pkg, client, tests/go_client): 0 issues. Root lint used explicit source paths to avoid an existing unreadable Docker data directory.

The three inline review threads are already resolved, and the terminal-state done-bucket fix remains in 18c1e3cb04. New-head CI is pending; full integration was not rerun locally.

internal/datacoord/compaction_task_clustering.go
```diff
@@ -780,18 +780,16 @@ func (t *clusteringCompactionTask) doCompact(nodeID int64, cluster session.Clust
...
-		if err != nil {
-			mlog.Warn(context.TODO(), "updateAndSaveTaskMeta fail", mlog.Int64("planID", t.GetTaskProto().GetPlanID()), mlog.Err(err))
-			return err
+		if saveErr := t.updateAndSaveTaskMeta(setState(datapb.CompactionTaskState_pipelining), setNodeID(NullNodeID)); saveErr != nil {
+			mlog.Warn(context.TODO(), "updateAndSaveTaskMeta fail", mlog.Int64("planID", t.GetTaskProto().GetPlanID()), mlog.Err(saveErr))
+			return merr.Combine(err, merr.Wrap(saveErr, "persist clustering compaction retry state"))
 		}
-		metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", originNodeID), t.GetTaskProto().GetType().String(), metrics.Executing).Dec()
-		metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", NullNodeID), t.GetTaskProto().GetType().String(), metrics.Pending).Inc()
+		return err
```
liliu-z, line 792: The new `return err` at `compaction_task_clustering.go:792` routes every `CreateCompaction` failure into `CreateTaskOnWorker`'s `defer t.retryOnError(err)`, where `WrapErrClusteringCompactionSubmitTaskFail` (`merr/errors.go:283`, code 2307, retriable) makes it consume the `maxRetryTimes: 3` budget — even for a non-retriable inner error such as `ErrDuplicatedCompactionTask` (`merr/errors.go:291`), because `wrapFieldsWithDesc` (`merr/utils.go:1426`) discards the inner code. The budget never resets during the rejection loop: `setRetryTimes(0)` at `:285` is gated on `currentState != lastState` (`:279`), and the state stays `pipelining` because `retryableProcess` has no case for it (`:330-340`). Four consecutive dispatch failures — a DataNode restarting during a rolling upgrade, an RPC timeout, or a `CreateCompaction` whose response was lost after the worker already enqueued the plan (`datanode/compactor/executor.go:110` returns `WrapErrDuplicatedCompactionTask` on redelivery) — set the task to `failed` permanently. Before this change the same branch fell through to `setState(executing), setNodeID(nodeID)` and returned nil, so `QueryTaskOnWorker` recovered the result from the worker that had actually accepted the plan. `mixCompactionTask` (`compaction_task_mix.go:104`) and `l0CompactionTask` (`compaction_task_l0.go:135`) still only reset to `pipelining` and retry indefinitely, and clustering's own `QueryTaskOnWorker` deliberately clears `err` on transport failures (`:158-161`) so query errors never consume the budget — Create is now the only exception. The new test at `compaction_task_clustering_test.go:519` exercises a single rejection and asserts `RetryTimes == 1`, so the cap is untested.

internal/datacoord/compaction_inspector.go
```diff
@@ -552,27 +550,35 @@ func (c *compactionInspector) removeTasksByChannel(channel string) {
...
 			)
 			delete(c.executingTasks, id)
-			metrics.DataCoordCompactionTaskNum.WithLabelValues(fmt.Sprintf("%d", task.GetTaskProto().GetNodeID()), task.GetTaskProto().GetType().String(), metrics.Executing).Dec()
+			removed = append(removed, task)
 		}
 	}
 	c.executingGuard.Unlock()
+	// Stop inspector processing first, then wait for scheduler callbacks without
+	// holding executingGuard across worker RPCs. Account for their final state.
+	for _, task := range removed {
```
liliu-z, line 559: The loop added at `compaction_inspector.go:559-562` calls `scheduler.AbortAndRemoveTask` once per removed task; before this change the function touched only local maps and issued no RPC. `AbortAndRemoveTask` (`global_scheduler.go:139-150`) holds the per-task write lock `s.mu.Lock(taskID)` across `DropTaskOnWorker` → `cluster.dropTask` (`session/cluster.go:160-174`), bounded only by `dataCoord.requestTimeoutSeconds` (default 600s, `component_param.go:7897`), and it must first wait out `schedule()`/`check()` callbacks that hold `s.mu.RLock(taskID)` across their own worker RPCs (`global_scheduler.go:291-293`, `345-347`). Because `check()` ends in `conc.BlockOnAll` (`global_scheduler.go:370`, `conc/future.go:118`), which waits for every future, a single callback blocked on that key lock stops the entire `check()` round — no running index, stats, import, copy-segment or compaction task anywhere in the cluster gets its state polled until the RPC times out; `schedule()` stalls the same way at `:333`. `removeTasksByChannel` takes no `ctx` (`:526`), so no caller deadline applies, and the loop is sequential over every removed task. Its only production caller is `Server.DropVirtualChannel` (`services.go:838`, handler at `:800`), reached from `syncmgr/meta_writer.go:328` with an explicit `context.Background()` and from the rootcoord drop-collection ack callback at `ddl_callbacks_drop_collection.go:135`, which is an in-process call under mixcoord.

internal/datacoord/task/global_scheduler.go
```diff
@@ -261,39 +261,43 @@ func (s *globalTaskScheduler) pickNode(slotHeap typeutil.Heap[*nodeSlotEntry], t
 }
 
 func (s *globalTaskScheduler) schedule() {
-	pendingNum := s.pendingTasks.TaskCount()
-	if pendingNum == 0 {
+	taskIDs := s.pendingTasks.TaskIDsByPriority()
```
liliu-z, line 264: `global_scheduler.go:264` calls `TaskIDsByPriority()` unconditionally, before `QuerySlot()`. That helper (`priority_queue.go:140-152`) clones the whole heap and drains it — O(N) allocation plus O(N log N) comparisons — regardless of how many tasks can actually be dispatched this round; the previous code took the O(1) `TaskCount()` early-exit (`priority_queue.go:109`) and, when no slots were free, did one `Pop`/`Push` and broke out. Dispatch also moved from `Pop()` (O(log N)) to `s.pendingTasks.Remove(...)` at `global_scheduler.go:300`, and `Remove` (`priority_queue.go:154-171`) is a linear scan of the heap under the exclusive lock, now executed by up to 128 concurrent `execPool` workers while `check()`'s retry path (`global_scheduler.go:359`) pushes onto the same lock from another 128. The queue is shared by every DataCoord task type — index (`index_inspector.go:246,282`), stats, analyze, import, copy-segment, external-refresh, compaction (`compaction_inspector.go:293`, whose own queue caps at `taskQueueCapacity` = 100000) — and only stats has an admission gate, which the recovery path at `stats_inspector.go:147` bypasses. At a 100 ms tick (`component_param.go:7292`) a six-figure backlog spends tens of milliseconds per tick sorting tasks it cannot dispatch, and blocks every producer's `Enqueue` behind the O(N) removals.

PR #52805, opened 2026-08-24
enhance: unify segcore chunk manager IO on the milvus-storage Arrow FileSystem

somfornot:
issue: #52925

## What

Unifies segcore's chunk-manager IO onto the milvus-storage Arrow FileSystem
(shared with Storage V2 via `StorageV2FSCache`), so segment load and index
build/load can run on the same IO stack as storage v2 instead of the legacy
AWS-SDK based chunk managers. Opt-in per process via
`common.storage.useArrowFileSystemChunkManager` (default **false** = legacy),
delivered to segcore during component init and honored by every
`CreateChunkManager` funnel.

## Commits

1. **feat: support Arrow FileSystem backed chunk manager in segcore** — adds
   `ArrowFileSystemChunkManager` (remote backend), wires the switch through
   component init (querynode/streamingnode via InitQueryNode, datanode/indexnode
   via InitSegcore) and every `CreateChunkManager` funnel, classifies errors via
   `milvus_storage::ToSegcoreError` (transient `2045` / permanent `2044` /
   not-found `2017` split), falls back to legacy for `cloudProvider=gcpnative`,
   and normalizes the deprecated `storageType=minio` to `remote` for the fs
   cache key.
2. **feat: route local chunk manager IO through Arrow FileSystem and add
   suite-level test switch** — extends routing to `storage_type=local` (arrow fs
   rooted at `/`, CWD-relative path parity, directory-aware `Exist`), converts
   raw producer exceptions into a classified `SegcoreError(StorageError)` at the
   adapter ctor, and adds `MILVUS_USE_ARROW_FS_CHUNK_MANAGER=1` so the whole C++
   unittest suite can run on the new IO path.
3. **test: bound RelativePathsResolveAgainstCwd to a scratch CWD** — fixes an
   always-on defect in the new parity test (a recursive whole-repo walk and a
   directory leak into the working tree on failed assertions).

## Verification

- C++ unit suite with the switch enabled passes (**8359 tests**); the remaining
  `AzureChunkManagerTest`/`GcpNativeChunkManagerTest` failures match the
  legacy-path baseline and are environment failures (no Azurite / GCS emulator).
  Adapter-specific tests pass, including a live-MinIO CRUD + not-found
  classification case (gated on `MINIO_ADDRESS`).
- Default is **off**, so production behavior is unchanged unless explicitly
  enabled — this keeps the blast radius of the new IO path opt-in.

## Follow-ups (not done in this PR)

- **Real-MinIO remote parity**: with the flag on, a `minio`/`remote` config
  swaps the S3 client (legacy AWS-SDK `MinioChunkManager` →
  milvus-storage `S3FileSystemProducer`). Addressing mode (path-style vs
  virtual-host), empty `region`, and `http://` prefixing must be confirmed
  equivalent against a real MinIO before enabling by default — not yet
  exercised by happy-path e2e.
- **End-to-end** query/index workloads with the flag enabled: not yet run.


internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,375 @@
...
+ArrowFileSystemChunkManager::Write(const std::string& filepath,
+                                   void* buf,
+                                   uint64_t size) {
+    LatencyObserver observer(
+        milvus::monitor::internal_storage_request_latency_put);
+    milvus::monitor::internal_storage_kv_size_put.Observe(size);
+
+    const auto path = ResolvePath(filepath);
+    auto output = fs_->OpenOutputStream(path);
+    if (!output.ok() && IsArrowNotFound(output.status())) {
```
jiaqizho: `!output.ok()` contain not found.

I mean if `IsArrowNotFound(output.status())` then  `!output.ok()` must be true.

configs/milvus.yaml
```diff
@@ -1126,6 +1108,10 @@ common:
   storageType: remote # please adjust in embedded Milvus: local, available values are [local, remote], value minio is deprecated, use remote instead
   storage:
     manifestTransactionRetryLimit: 10 # Maximum number of retry attempts for V3 storage manifest transaction commits on optimistic concurrency conflicts
+    # Whether segcore routes chunk manager IO (segment load, index build/load) through the milvus-storage Arrow FileSystem, unified with storage v2.
+    # Applies to both remote and local storage types; when false, the legacy chunk managers are used.
+    # Not effective for cloudProvider=gcpnative, which always uses the legacy implementation.
+    useArrowFileSystemChunkManager: false
```
jiaqizho: How about direct use `useLoonFFI`? Then V3 direct use enable it.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,375 @@
...
+        // Object stores have no directories, but a local-backed filesystem
+        // needs the parent directory to exist. Create it and retry once so
+        // the "write object at key" contract holds on every backend.
+        auto slash_pos = path.find_last_of('/');
+        if (slash_pos != std::string::npos) {
+            auto mkdir_status = fs_->CreateDir(path.substr(0, slash_pos), true);
+            if (mkdir_status.ok()) {
+                output = fs_->OpenOutputStream(path);
+            }
```
jiaqizho: what if `!mkdir_status.ok()` ?

internal/core/src/storage/Util.cpp
```diff
@@ -1124,19 +1143,70 @@ ReleaseArrowUnused() {
...
         case ChunkManagerType::Local: {
+            if (auto cm = TryCreateArrowFileSystemChunkManager(storage_config);
+                cm != nullptr) {
+                return cm;
+            }
             return std::make_shared<LocalChunkManager>(
                 storage_config.root_path);
         }
         case ChunkManagerType::Minio: {
+            if (auto cm = TryCreateArrowFileSystemChunkManager(storage_config);
```
jiaqizho:
```
storage_type = "minio" 
cloud_provider=""
```

still will not create the filesystem.

internal/core/src/storage/Util.cpp
```diff
@@ -1124,19 +1143,70 @@ ReleaseArrowUnused() {
...
+            "manager",
+            storage_config.cloud_provider);
+        return nullptr;
+    }
+    return std::make_shared<ArrowFileSystemChunkManager>(storage_config);
+}
+
 ChunkManagerPtr
 CreateChunkManager(const StorageConfig& storage_config) {
     auto storage_type = ChunkManagerType_Map[storage_config.storage_type];
```
jiaqizho: can we just inspert and create ArrowFileSystemChunkManager in this line?

```

if (auto cm = TryCreateArrowFileSystemChunkManager(storage_type, storage_config);
                cm != nullptr) {
                return cm;
            }
```

Then TryCreateArrowFileSystemChunkManager do the mapping.

jiaqizho:
BTW, fix the CI please

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Right — a not-found status is never ok, so the extra `!output.ok()` was redundant. Dropped it in 8e94473ee6.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Fixed in 8e94473ee6: on `!mkdir_status.ok()` it now increments the put-fail metric and throws via `ThrowArrowStorageError("Write", path, mkdir_status)`, surfacing the real CreateDir cause instead of falling through and rethrowing the original not-found status.

internal/core/src/storage/Util.cpp
somfornot: Done in 8e94473ee6 — hoisted the `TryCreateArrowFileSystemChunkManager(storage_config)` call above the storage-type switch so it is attempted once up front, dropping the three duplicated blocks. It reads `storage_config.storage_type` internally and does the mapping there (switch off → nullptr; `local` → always; otherwise gated on `SupportsCloudProvider`), so I kept the single-arg signature rather than passing `storage_type` separately.

internal/core/src/storage/Util.cpp
somfornot: Fixed in 8e94473ee6 — `SupportsCloudProvider` now treats an empty `cloud_provider` as supported, so `storage_type="minio", cloud_provider=""` routes to the ArrowFileSystem backend. (`gcpnative` stays excluded — no milvus-storage producer.)

configs/milvus.yaml
somfornot: Re: `useLoonFFI` — I'd keep `useArrowFileSystemChunkManager` as a separate flag rather than gating it on `useLoonFFI`. They act on orthogonal layers:

- `useLoonFFI` (`common.storage.useLoonFFI`, 2.6.7, refreshable) selects the **read/write path / format layer** — how data is encoded and which reader-writer parses it.
- `useArrowFileSystemChunkManager` (`common.storage.useArrowFileSystemChunkManager`, 2.7.0) selects the segcore **IO backend** — whether raw byte IO goes through the milvus-storage Arrow FileSystem or the legacy AWS-SDK `MinioChunkManager` family.

Every combination is meaningful (e.g. the Arrow FS IO backend is useful under either format path), so folding one into the other would tie a transport rollout to a format rollout and drop the ability to toggle them independently; they also gate at different versions.

If the concern is a specific unsupported combination, I'd rather add an explicit *imply* (enabling one assumes the other) than merge the two flags — happy to wire that if you point me at the combination to guard.

somfornot:
Thanks for the nudge — both CI failures are fixed in e425db067b:

- **Code-Check (clang-format-15):** the two flagged files (`ArrowFileSystemChunkManager.cpp`, `ArrowFileSystemChunkManagerTest.cpp`) are now formatted with the exact `internal/core/run_clang_format.sh` (clang-format-15) that `cppcheck` runs.
- **UT-CPP-Cov:** the single failure was `ArrowFileSystemChunkManagerRemote.MinioCRUD`. It needs a live MinIO, but in the cpp-ut pod `MINIO_ADDRESS` points at the docker-compose `minio` service, which doesn't resolve there, so it errored with a transient-network status (2045) after retries. I moved it into its own `ArrowFileSystemChunkManagerRemoteTest.cpp` and excluded that file from `all_tests`, exactly like `RemoteChunkManagerTest` / `MinioChunkManagerTest`; it still runs against a live MinIO (e.g. inside the docker-compose builder).

The five inline review comments are addressed in 8e94473ee6 (replied in-thread).

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+bool
+ArrowFileSystemChunkManager::SupportsCloudProvider(
+    const std::string& cloud_provider) {
+    // Empty cloud_provider is treated as supported: a minio/remote config
+    // with no explicit provider (e.g. self-hosted MinIO) still routes to the
+    // ArrowFileSystem backend. gcpnative stays excluded (no milvus-storage
+    // producer).
+    static const std::unordered_set<std::string> supported = {
+        "", "aws", "gcp", "aliyun", "azure", "tencent", "huawei"};
```
jiaqizho: "" is not support.

internal/core/src/storage/ArrowFileSystemChunkManager.h
```diff
@@ -0,0 +1,136 @@
...
+
+    virtual std::string
+    GetBucketName() const {
+        return default_bucket_name_;
+    }
+
+    // Whether the milvus-storage ArrowFileSystem has a producer for this
+    // cloud provider. "gcpnative" (and unknown values) must stay on the
+    // legacy chunk managers.
+    static bool
```
jiaqizho: Just provider a `ChunkManagerPtr ArrowFileSystemChunkManager::Make`.

```
static ChunkManagerPtr ArrowFileSystemChunkManager::Make(config) {
   // do try...catch if need
   auto key = ToStorageV2FSCacheKey(storage_config);
   return StorageV2FSCache::Instance().Get(key);;
}
````

Then u don't need throw the exception in constructor and also no need check the provider support in milvus side.

File: internal/core/src/storage/ArrowFileSystemChunkManager.cpp [hunk 8]
```diff
@@ -0,0 +1,385 @@
...
+
+// Classify and throw. The transient-vs-permanent decision is made where the
+// arrow::Status was constructed (milvus-storage attaches ExtendStatusDetail
+// at the AWS error sites); ToSegcoreError only unwraps it. See
+// milvus-storage extend_status.cpp for the full taxonomy.
+[[noreturn]] void
+ThrowArrowStorageError(const std::string& func,
+                       const std::string& path,
+                       const arrow::Status& status) {
+    auto segcore_error = milvus_storage::ToSegcoreError(status);
```
jiaqizho, line 59: And notice that, filesystem will make extends status error. Which is retryable.

jiaqizho: So if current error is retryable, we should not direct mapping it to 2044(non-retry)

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: You're right, and this was a real bug I introduced — thanks for catching it. Verified against milvus-storage's `CreateArrowFileSystem`: `CloudProviderType_Map` has no entry for `""` (nor `gcpnative`), so `map[provider]` yields `CloudProviderType::UNKNOWN` → the `default:` branch → `arrow::Status::Invalid`, which `StorageV2FSCache::Get` surfaces as `nullptr`. An empty provider genuinely cannot build a filesystem.

Adding `""` to the allow-list made a `minio`/`remote` config with no explicit provider **throw** a hard `StorageError` at construction instead of falling back to the legacy `MinioChunkManager`. Fixed in 77612fc856 by removing the whole allow-list (see the `Make` thread) — empty / unknown / gcpnative providers now all fall back to legacy.

internal/core/src/storage/ArrowFileSystemChunkManager.h
somfornot: Done in 77612fc856. Added `static ChunkManagerPtr ArrowFileSystemChunkManager::Make(const StorageConfig&)` that returns `nullptr` whenever milvus-storage has no producer for the config (`StorageV2FSCache::Get` → `nullptr`), so `CreateChunkManager` falls back to the legacy managers. As you suggested:

- Removed `SupportsCloudProvider` — milvus-storage's `CreateArrowFileSystem` is now the single source of truth for provider support; milvus no longer second-guesses it with an allow-list.
- No more throwing in the constructor: the public config-taking constructor is gone, and a private constructor takes the already-built filesystem. `TryCreateArrowFileSystemChunkManager` is now just `return Make(storage_config);`.

This also fixes the empty-provider bug from the other thread (unsupported provider → fall back, not throw). Added a network-free unit test `MakeUnsupportedProviderFallsBack` (gcpnative/`""` → `nullptr`, `aws` → a valid CM).

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Good point — verified end-to-end, and the current code already preserves retryability (it does **not** map a retryable error to 2044):

`ThrowArrowStorageError` delegates entirely to `milvus_storage::ToSegcoreError(status)` and never hardcodes a code. And the S3 filesystem does exactly what you describe — on its retryable paths it attaches an `ExtendStatusDetail` with a transient code (`s3_filesystem.cpp`: `MakeExtendError(ExtendStatusCode::StorageTransientNetwork, ...)`, etc.). `ToSegcoreError` unwraps that detail and `ToSegcoreErrorCode` maps the `StorageTransient*` codes to `milvus::StorageTransientError` (2045, retriable) — not 2044.

The only path to 2044 is a *plain* `arrow::Status` IOError with **no** `ExtendStatusDetail`, which milvus-storage itself classifies as non-retriable by design (the shared S3 SDK retry budget is already spent). So a retryable FS error surfaces as 2045; nothing on the milvus side collapses it to 2044.

(Separately, after the `Make()` refactor the init-time `StorageError` throw in the old constructor is gone — an unbuildable config now falls back to legacy instead of throwing.)

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,392 @@
...
+    // prefix.
+    const auto resolved = ResolvePath(filepath);
+    arrow::fs::FileSelector selector;
+    auto slash_pos = resolved.find_last_of('/');
+    selector.base_dir =
+        slash_pos == std::string::npos ? "" : resolved.substr(0, slash_pos);
+    selector.recursive = true;
+    selector.allow_not_found = true;
+
+    auto infos = fs_->GetFileInfo(selector);
```
buqian-zilliz: **[P1] Keep raw-prefix filtering on the storage side**

`GetFileInfo(selector)` sends only the directory before the last slash to Arrow and applies the rest of `filepath` after the returned `FileInfo` vector has been fully materialized. For example, `ListWithPrefix("indexes/build-123")` enumerates every object below `indexes/`, and a prefix without `/` enumerates the entire bucket. Unlike the legacy implementation, which sends the full raw prefix to S3, this turns one prefix lookup into O(parent-subtree) paginated I/O and memory. Production buckets can contain millions of objects, so this can stall or OOM the process. Please add a storage-side raw-prefix listing primitive or otherwise query with the complete prefix without materializing the parent subtree.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,392 @@
...
+}
+
+void
+ArrowFileSystemChunkManager::Remove(const std::string& filepath) {
+    const auto path = ResolvePath(filepath);
+    arrow::Status status;
+    {
+        LatencyObserver observer(
+            milvus::monitor::internal_storage_request_latency_remove);
+        status = fs_->DeleteFile(path);
```
buqian-zilliz: **[P2] Use flat object deletion for ChunkManager keys**

`fs_->DeleteFile(path)` is directory-aware. In the pinned milvus-storage S3 implementation it performs `HeadObject`, `DeleteObject`, and then an unconditional `PutObject` for `<parent>/` to preserve a directory marker, while the legacy chunk manager issues one idempotent raw-object delete. Every cleanup therefore becomes roughly three remote requests, leaves or overwrites marker objects, and can report failure if the marker PUT fails after the target was already deleted. Please use a flat-object delete primitive, or disable parent-marker maintenance, so `Remove` retains single-object, idempotent semantics.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,392 @@
...
+// itself fails (endpoint down, access denied), fall back to the original
+// status classification so transient errors stay transient.
+bool
+PathMissing(const milvus_storage::ArrowFileSystemPtr& fs,
+            const std::string& filepath,
+            const arrow::Status& failure) {
+    if (IsArrowNotFound(failure)) {
+        return true;
+    }
+    auto info = fs->GetFileInfo(filepath);
```
buqian-zilliz: **[P2] Classify misses with exact-object semantics**

`PathMissing` re-stats a bare S3 `PathNotFound` through directory-aware `GetFileInfo`. When the exact key is absent but `key/child` or a `key/` marker exists, Arrow reports `Directory`, so `Read` no longer produces `ObjectNotExist` and `Remove` no longer swallows the miss. The helper also performs this extra remote request for statuses that already carry a structured non-not-found detail such as timeout or throttling. Restrict the probe to the unclassified case and use an exact-object existence check that cannot be satisfied by descendants.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,392 @@
...
+    if (storage_config.storage_type == "local") {
+        // LocalChunkManager parity: callers pass OS paths (absolute,
+        // root_path-prefixed), not root-relative keys. Root the filesystem
+        // at "/" so those paths pass through unchanged.
+        key.root_path = "/";
+        local_backend = true;
+    }
+    milvus_storage::ArrowFileSystemPtr fs;
+    try {
+        fs = StorageV2FSCache::Instance().Get(key);
```
buqian-zilliz, line 126: **[P2] Preserve fail-fast storage validation on the Arrow backend**

The legacy remote managers run `PreCheck` during construction, so bad credentials, bucket, endpoint, or connectivity fail `InitRemoteChunkManager` before the node becomes ready. This factory only constructs a cached Arrow filesystem without remote I/O, so the same unusable configuration can initialize successfully and first fail on live segment or index traffic. Moreover, `StorageV2FSCache::Get` collapses all construction errors to null and this factory treats null as a legacy-fallback signal, so invalid configurations for supported providers can silently ignore the opt-in. Please validate remote accessibility before installing the Arrow manager and reserve fallback for explicitly unsupported providers.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
```diff
@@ -0,0 +1,392 @@
...
+ private:
+    prometheus::Histogram& histogram_;
+    std::chrono::system_clock::time_point start_;
+};
+
+}  // namespace
+
+ChunkManagerPtr
+ArrowFileSystemChunkManager::Make(const StorageConfig& storage_config) {
+    auto key = ToStorageV2FSCacheKey(storage_config);
```
buqian-zilliz: **[P2] Preserve millisecond request-timeout semantics**

Routing this configuration through the Arrow filesystem changes the documented `requestTimeoutMs` contract. Legacy managers pass the exact millisecond value to their storage clients, but the pinned Arrow S3/GCP producers compute `request_timeout_ms / 1000` with integer operands, so 1500 ms becomes 1 second and values below 1 second become zero; the Arrow Azure producer does not apply this timeout to blob I/O at all. The same configuration can therefore time out earlier or lose its intended timeout when this backend is enabled. Please preserve fractional seconds and wire the timeout through every supported provider before routing ChunkManager traffic through it.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
jiaqizho: This requires support from the storage filesystem.

somfornot: Resolved in the remote-only rework (869a42c7c6). `ListWithPrefix` is no longer served by the Arrow filesystem — it now delegates to the legacy `MinioChunkManager::ListWithPrefix`, which sends the full raw prefix to S3 via `ListObjects` + `request.SetPrefix(prefix)`. There is no parent-subtree `GetFileInfo(selector)`, no client-side filtering, and no materialization of sibling objects: a mid-segment prefix, or a prefix without `/`, is passed verbatim as the S3 object-key prefix, exactly like the legacy path. This also removes the need for any storage-side raw-prefix listing primitive.

Verified against a live MinIO: with objects `…/1/2/3` and `…/1/2/30` both present, `ListWithPrefix("…/1/2/3")` returns both — confirming raw byte-prefix semantics with no over-read.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Resolved in 869a42c7c6. `Remove` now delegates to the legacy `MinioChunkManager::Remove`, i.e. a single idempotent `DeleteObject` — no `HeadObject`, no parent-marker `PutObject`, and no directory-aware `DeleteFile`. Cleanup is one remote request, and a missing key is swallowed rather than reported as failure.

Verified against a live MinIO: `Remove` of a missing key is a no-op (idempotent), and `Remove` of an existing key deletes exactly that object.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Resolved in 869a42c7c6. The directory-aware `PathMissing` re-stat is gone. Read-side not-found is now classified through the legacy delegate's exact-object existence: `ThrowReadError` calls `delegate_->Exist(path)` → `MinioChunkManager::ObjectExists`, a `HeadObject` that returns false only when the exact key is absent and cannot be satisfied by a `key/` marker or a `key/child` descendant.

The probe runs only on the failure path (errors only, no cost on success), and a status that already carries a not-found detail short-circuits via `IsArrowNotFound` before any extra remote request — so structured non-not-found statuses (timeout/throttle) are not re-stat'd and stay in their original, retryable bucket. `Exist`/`Size` are likewise the delegate's `HeadObject`.

Verified against a live MinIO: `Read` and `Size` on a missing key raise `ObjectNotExist`.

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Resolved in 869a42c7c6. `Make()` now builds a legacy remote delegate via `CreateLegacyChunkManager`, and that constructor runs the same `PreCheck` (a live `ListObjects` probe) as `InitRemoteChunkManager` — so bad credentials/bucket/endpoint/connectivity fail during construction, before the node becomes ready, with the exception propagating out of `Make` → `CreateChunkManager`.

Fallback is now reserved for configurations milvus-storage genuinely has no Arrow producer for: `Make` returns null only when `CreateArrowFileSystem` declines the config (e.g. `gcpnative` / empty provider). A *supported* provider that is merely misconfigured now fails fast via the delegate's `PreCheck` instead of silently falling back. (A local config also returns null → `LocalChunkManager`, since this manager is remote-only.)

internal/core/src/storage/ArrowFileSystemChunkManager.cpp
somfornot: Not addressed in this PR — flagging explicitly so it is not assumed fixed. This is the one finding the remote-only hybrid does **not** resolve: `Read`/`Write` still go through the Arrow S3/GCP producers, which compute `request_timeout_ms / 1000` with integer operands (1500 ms → 1 s, sub-second → 0).

Note the control plane (`Exist`/`Size`/`ListWithPrefix`/`Remove`) now runs on the legacy delegate and keeps the exact millisecond timeout, so only the Arrow data path is affected. The correct fix is on the milvus-storage side (producer `/1000` → `/1000.0`, plus wiring the timeout through the Azure producer); I will land that as a separate milvus-storage change rather than widening this PR. Tracking it as a follow-up.

pkg/util/paramtable/component_param.go
```diff
@@ -702,6 +703,18 @@ This configuration is only used by querynode and indexnode, it selects CPU instr
 	}
 	p.ManifestTransactionRetryLimit.Init(base.mgr)
 
+	p.UseArrowFSChunkManager = ParamItem{
+		Key:          "common.storage.useArrowFileSystemChunkManager",
+		Version:      "2.7.0",
```
zhengbuqian: we'll not have 2.7.0, how about 3.0.1 or 3.0.2?

somfornot: Good catch — done in 138ef099ff. Set `Version: "3.0.1"` to match the actual introducing release: v3.0.0 is already tagged, and the sibling params in this file are on 3.0.1. Thanks!

PR #52807, opened 2026-08-24
fix: [3.0] present declared defaults for absent fields in the StorageV3 manifest record reader

MrPresent-Han:
Cherry-pick of #52781 onto 3.0.

Makes the StorageV3 manifest record reader default-aware: it reads only physically-present columns and fills every absent read-schema field with its declared default (else null), matching the V1 binlog reader, so declared defaults are no longer presented as NULL (#52771). Scoped to the StorageV3 manifest reader; the V2/binlog path is unchanged. Includes the reader/refcount/struct-fill/present-fields unit tests.

issue: #52771
pr: #52781


liliu-z (commented):
/lgtm
/approve

PR #52810, opened 2026-08-24
test: add REST value-fidelity e2e tests for PR #52261

yanliang567:
## Summary

Add e2e coverage for the REST value-handling fixes in #52261, which stopped the REST layer from rewriting the values a caller sends.

## What's covered

- **Value fidelity** (`write == read == filter hit`): Int8/16/32/64, Float, Double, Bool, VarChar, JSON and dynamic fields, including `>2^53` integers and scientific/decimal literal forms (`1e3`, `9007199254740993.0`, `1.50`).
- **Rejection**: out-of-range integers, fractional Int64, Float32 overflow, VarChar object/array, JSON duplicate key / oversized integer, missing required field, vector-as-string, leading zero, array null element, non-nullable null, whole-value null search vector, Int8Vector quoted `"null"`.
- **exprParams**: exact bigint match, int64 max (insert + exact hit + adjacent miss), beyond-int64 and `null`/`[]`/`{}` rejection.
- **Partial update**: explicit `null` clears a dynamic field; untouched keys are preserved byte-for-byte.
- **id list**: multi-id get exactness, quote escaping, numeric id for a VARCHAR key.
- **Struct Array**: sub-field big-int round-trip, out-of-range and null-element rejection.
- **gRPC/REST consistency** and **growing-vs-sealed** read-back consistency.

## Test levels

57 cases, passing against master `f192bafbd5`:

| Level | Count | Content |
|---|---|---|
| L0 | 27 | positive value-fidelity paths |
| L1 | 28 | negative / rejection paths |
| L2 | 2 | BinaryVector base64, Int8Vector quoted-null |

## Deferred

Config-switch behaviors (`proxy.http.compatibilityMode`, toggling `proxy.http.nativeJSONResponse`, `proxy.http.maxExprParamsDepth`) and legacy non-document JSON bytes require runtime config changes, so they are deferred to a separate upgrade-compatibility test.

### DIFF
--- a/tests/restful_client_v2/testcases/test_rest_value_fidelity.py
+++ b/tests/restful_client_v2/testcases/test_rest_value_fidelity.py
@@ -792,3 +792,65 @@ class TestRestValueFidelity(TestBase):
             % pk,
         )
         assert rsp["code"] != 0, rsp
+
+    # ================= Struct Array sub-field value handling =================
+
+    @pytest.mark.tags(CaseLabel.L1)
+    def test_struct_array_subfield_value_handling(self):
+        # Struct Array sub-fields flow through parseStructArrayRow ->
+        # buildStructSubArrayScalar -> parseScalarArrayElements, the same value
+        # checks as a plain Array column: big integers round-trip exactly, and
+        # out-of-range or null elements are rejected.
+        name = gen_collection_name("rvf")
+        payload = {
+            "collectionName": name,
+            "schema": {
+                "autoID": False,
+                "enableDynamicField": False,
+                "fields": [
+                    {"fieldName": "id", "dataType": "Int64", "isPrimary": True, "elementTypeParams": {}},
+                    {"fieldName": "vec", "dataType": "FloatVector", "elementTypeParams": {"dim": "2"}},
+                ],
+                "structFields": [
+                    {
+                        "fieldName": "profile",
+                        "typeParams": {"max_capacity": "4"},
+                        "fields": [
+                            {
+                                "fieldName": "p_int",
+                                "dataType": "Array",
+                                "elementDataType": "Int64",
+                                "elementTypeParams": {"max_capacity": "4"},
+                            },
+                            {
+                                "fieldName": "p_tag",
+                                "dataType": "Array",
+                                "elementDataType": "VarChar",
+                                "elementTypeParams": {"max_capacity": "4", "max_length": "128"},
+                            },
+                        ],
+                    }
+                ],
+            },
+            "indexParams": [{"fieldName": "vec", "indexName": "vec", "metricType": "L2"}],
+        }
+        rsp = self.collection_client.collection_create(payload)
+        assert rsp["code"] == 0, rsp
+        self.collection_client.collection_load(collection_name=name)
+        self.wait_collection_load_completed(name)
+
+        # bigint sub-field element round-trips exactly
+        rsp = self._raw_insert(name, '[{"id":1,"profile":[{"p_int":9007199254740993,"p_tag":"a"}],"vec":[0.1,0.2]}]')
+        assert rsp["code"] == 0, rsp
+        self.collection_client.flush(name)
+        time.sleep(1)
+        d = self._query(name, "id == 1", ["profile"])["data"][0]
+        assert d["profile"] == [{"p_int": 9007199254740993, "p_tag": "a"}], f"got {d['profile']!r}"
+
+        # out-of-range sub-field element rejected
+        rsp = self._raw_insert(name, '[{"id":2,"profile":[{"p_int":9223372036854775808,"p_tag":"a"}],"vec":[0.1,0.2]}]')
+        assert rsp["code"] != 0, rsp
+
+        # null sub-field element rejected
+        rsp = self._raw_insert(name, '[{"id":3,"profile":[{"p_int":null,"p_tag":"a"}],"vec":[0.1,0.2]}]')
+        assert rsp["code"] != 0, rsp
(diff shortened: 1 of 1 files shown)


tests/restful_client_v2/testcases/test_rest_value_fidelity.py
```diff
@@ -0,0 +1,856 @@
...
+                    "indexName": "iv",
+                    "indexType": "HNSW",
+                    "metricType": "L2",
+                    "params": {"M": 8, "efConstruction": 64},
+                }
+            ],
+        )
+        rsp = self._raw_insert(name, '[{"id":1,"iv":"AQ=="}]')
+        assert rsp["code"] != 0, f"Int8Vector base64 should be rejected: {rsp}"
+        rsp = self._raw_insert(name, '[{"id":1,"iv":null}]')
```
zhuwenxing: Could this send the quoted JSON string `"null"` instead? #52261 fixed the case where Int8Vector was incorrectly treated as base64-capable, causing `"null"` to decode as an empty vector. A JSON literal `null` was already rejected by the old path. Likewise, `"AQ=="` was already rejected by the Int8Vector array decoder, so the current assertions do not distinguish fixed from pre-fix behavior.

tests/restful_client_v2/testcases/test_rest_value_fidelity.py
```diff
@@ -0,0 +1,856 @@
...
+        assert d["ai"] == [1, 2] and d["as"] == ["a", "b"]
+
+    # ================= L1: additional rejection =================
+
+    @pytest.mark.tags(CaseLabel.L1)
+    def test_vector_null_rejected(self):
+        pk = self._new_pk()
+        rsp = self._raw_insert(
+            self._scalar_coll,
+            '[{"id":%d,"i8":1,"i16":1,"i32":1,"i64":1,"f32":1,"f64":1,"b":true,"s":"x","j":{"k":1},"vec":null}]' % pk,
```
zhuwenxing: Could this exercise a query/search request with a whole-value null vector instead? The corresponding change in #52261 is in query-side list decoding. An insert row containing `vec: null` was already rejected as a missing/non-nullable vector, so this test does not cover the changed path.

tests/restful_client_v2/testcases/test_rest_value_fidelity.py
```diff
@@ -0,0 +1,856 @@
...
+    def test_exprparams_int64_max_ok(self):
+        rsp = self.vector_client.vector_query(
+            {
+                "collectionName": self._scalar_coll,
+                "filter": "i64 == {v}",
+                "exprParams": {"v": 9223372036854775807},
+                "limit": 10,
+            }
+        )
+        assert rsp["code"] == 0, rsp
```
zhuwenxing: This only proves that the request returns code 0. It can still pass if `9223372036854775807` is decoded or rounded incorrectly because no matching row is inserted. Please insert the max-int64 row, assert an exact hit, and add an adjacent-value miss.

tests/restful_client_v2/testcases/test_rest_value_fidelity.py
```diff
@@ -0,0 +1,856 @@
...
+        rsp = self._raw_insert(
+            self._scalar_coll,
+            '[{"id":%d,"i8":1,"i16":1,"i32":1,"i64":9007199254740993.0,"f32":1,"f64":1,"b":true,"s":"x","j":{"k":1},"vec":[0.1,0.2]}]'
+            % pk,
+        )
+        assert rsp["code"] == 0, rsp
+        self.collection_client.flush(self._scalar_coll)
+        time.sleep(1)
+        d = self._query(self._scalar_coll, f"id == {pk}", ["i64"])["data"][0]
+        assert d["i64"] == 9007199254740993
```
zhuwenxing, line 264: The PR describes value fidelity as `write == read == filter hit`, but this test only queries by primary key. Please also filter on the exact `i64` value and assert that the adjacent rounded value does not match. The same gap exists in several other fidelity cases.

zhuwenxing (changes requested):
Thanks for adding the end-to-end coverage. Several cases currently do not exercise the pre-fix failure mode, so they can pass against the old behavior. Please align the inputs and endpoints with #52261.\n\nPlease also update the PR description: the current file collects 57 cases (L0=27, L1=28, L2=2), not 58 (27/27/4), and it no longer contains the claimed FP16/BF16/Sparse or VarChar-array cases.

yanliang567:
Thanks for the review. I've addressed all four inline comments:

1. **Int64 bigint fidelity** (line 233): added an exact filter hit (`i64 == {v}` with `9007199254740993`) plus an adjacent rounded-value miss (`9007199254740992`).
2. **exprParams int64 max** (line 351): now inserts the max-int64 row and asserts an exact hit plus an adjacent miss (`9223372036854775806`), rather than only checking `code == 0`.
3. **Int8Vector** (line 681): now sends the quoted `"null"` string (the pre-fix path that decoded to an empty vector), not the JSON null literal or base64, which the old path already rejected.
4. **null vector** (line 782): now exercises the search path with a whole-value null query vector, not insert.

On "the same gap exists in several other fidelity cases": the int64 bigint filter-hit is covered both inline here and by the dedicated `test_exprparams_bigint_exact_match` case. For the other types (VarChar literals, dynamic fields, JSON), an exact filter needs each type's own expression syntax (e.g. `s == "1e300"`, JSON predicates), and the precision risk those paths carry is primarily the int64/dynamic bigint case, which is now covered end-to-end. Happy to add per-type filter hits if you'd prefer that instead.

